PR #72, opened 2012-03-22
Fixed canceling the Logging function through the button on the main toolbar

Susurrus:


Strattoon:
Merged directly to origin/master in d3ec111f9. Implemented SprigMainWindow as the default root shell with SprigTopBar, SprigStatusBar, FLIGHT/PLANNING/RECORDS hotkeys, and Advanced access to stock QGC. Verification: qmllint, git diff --check, cmake debug build, Sprig unit subset 9/9, qt-orchestration verify-slice.

PR #73, opened 2012-03-22
Linechart UI improvements

Susurrus:
Modified the Linechart widget so that the "Show units" button is under the "Curve Selection" frame, which is what it actually modifies. This also involved me moving it into the Linechart.ui file instead of having it procedurally generated. Is there a reason that the Linechart widget is half described in the .ui file and half generated in `LinechartWidget::createLayout()`?

Also moved all of the UI that was under the "Curve Selection" to inside of it at the bottom so that the grouping of UI elements is more coherent.

I also wanted to add 2px of padding to both of the frames within this widget, but I couldn't figure out how to do it. The UI elements are a bit too cramped without this.

Finally this actually seems to have affected the performance of the initial loading of the UI once a MAVLink input source has been selected. I've been testing it on the demo_log simulated data. Maybe someone can confirm this is occurring and offer insight on how to fix it.

### DIFF
--- a/src/ui/Linechart.ui
+++ b/src/ui/Linechart.ui
@@ -25,8 +25,11 @@
   <property name="windowTitle">
    <string>Form</string>
   </property>
-  <layout class="QGridLayout" name="gridLayout" columnstretch="5,5,5,5,100">
-   <item row="0" column="0" colspan="4">
+  <property name="styleSheet">
+   <string notr="true"/>
+  </property>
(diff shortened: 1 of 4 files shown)


Strattoon:
Merged directly to origin/master in d3ec111f9. Implemented SprigTopBar and registered it in custom/custom.qrc. Verification: qmllint, cmake debug build, Sprig unit subset 9/9, qt-orchestration verify-slice.

PR #1100, opened 2014-12-09
Joystickinput settargets2

LorenzMeier:
@thomasgubler Can you give some recommendations how to best integrate this branch? Is this the better solution compared to joystickinput_settargets?

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -2552,60 +2552,31 @@ void UAS::setExternalControlSetpoint(float roll, float pitch, float yaw, float t
                     yawrate
                     );
         } else if (joystickMode == JoystickInput::JOYSTICK_MODE_MANUAL) {
-            if(((base_mode & MAV_MODE_FLAG_DECODE_POSITION_MANUAL) && (base_mode & MAV_MODE_FLAG_DECODE_POSITION_SAFETY)) || (base_mode & MAV_MODE_FLAG_HIL_ENABLED)) {
 
-                // Transmit the manual commands only if they've changed OR if it's been a little bit since they were last transmit. To make sure there aren't issues with
-                // response rate, we make sure that a message is transmit when the commands have changed, then one more time, and then switch to the lower transmission rate
-                // if no command inputs have changed.
-                // The default transmission rate is 50Hz, but when no inputs have changed it drops down to 5Hz.
-                bool sendCommand = false;
-                if (countSinceLastTransmission++ >= 10) {
-                    sendCommand = true;
-                    countSinceLastTransmission = 0;
-                } else if ((!isnan(roll) && roll != manualRollAngle) || (!isnan(pitch) && pitch != manualPitchAngle) ||
-                        (!isnan(yaw) && yaw != manualYawAngle) || (!isnan(thrust) && thrust != manualThrust) ||
-                        buttons != manualButtons) {
-                    sendCommand = true;
+            // Save the new manual control inputs
+            manualRollAngle = roll;
+            manualPitchAngle = pitch;
+            manualYawAngle = yaw;
+            manualThrust = thrust;
+            manualButtons = buttons;
 
-                    // Ensure that another message will be sent the next time this function is called
-                    countSinceLastTransmission = 10;
-                }
+            // Store scaling values for all 3 axes
+            const float axesScaling = 1.0 * 1000.0;
 
-                // Now if we should trigger an update, let's do that
-                if (sendCommand) {
-                    // Save the new manual control inputs
-                    manualRollAngle = roll;
-                    manualPitchAngle = pitch;
-                    manualYawAngle = yaw;
-                    manualThrust = thrust;
-                    manualButtons = buttons;
+            // Calculate the new commands for roll, pitch, yaw, and thrust
+            const float newRollCommand = roll * axesScaling;
+            // negate pitch value because pitch is negative for pitching forward but mavlink message argument is positive for forward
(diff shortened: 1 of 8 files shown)


Susurrus:
Sounds like this branch needs some cleanup/more work before a PR. Should we close this until then?

thomasgubler:
I think this is still only branch which enables joystick offboard control with px4. Hence I would prefer to leave it here. Someone who needs joysytick control can clean it up eventually.

DonLakeFlyer:
Leave the branch but close the pull?

thomasgubler:
Isn't the open PR more obvious for someone who wants to pick up this work? As an open source project we need to accept that sometimes works is done at a slower pace (for whatever reason).

BTW we have 287 open issues and only 8 PR. If someone feels like doing house cleaning I would start with the issues. ;-)

LorenzMeier:
@Susurrus I have seen that some open source projects do not allow feature requests to be filed as issues. To me that makes some sense. I know that you're trying to help by giving a lot of good feedback via issues, but in the current workflow the older ones or the feature request ones just consume time to read and handle. It would be great if you could sort through them and kill all which are not direct bug reports or a short-term low hanging fruit you're planning to attack yourself or someone volunteers to fix.

DonLakeFlyer:
If it still needs work, then might as well leave it open. You get better tracking with a pull than just a separate branch and an issue. Could use more details as to what exactly it was and why it's here. It was a bit mysterious to anyone other than Lorenz and Thomas as to what exactly this was until a few comments ago.

thomasgubler:
See the comments here https://github.com/mavlink/qgroundcontrol/pull/813 which got removed again by revert on master.
This PR is basically the revert of the revert of #813

DonLakeFlyer:
Thanks Thomas.

thomasgubler:
* rebased
  * re-introduced manual control, is now a joystick mode like attitude or position.
  * quickly tested manual control mode with PX4 HIL.
  * some small fixes and style improvements

LorenzMeier:
I've rebased this again and squashed it into a single commit. I want to get this onto master as part of an effort to make Joysticks / Gamepads a 100% replacement for the remote control for SITL environments.

src/uas/UAS.cc
```diff
@@ -2410,56 +2410,198 @@ void UAS::setManualControlCommands(float roll, float pitch, float yaw, float thr
...
+                    0
+                    );
+        } else if (joystickMode == JoystickInput::JOYSTICK_MODE_FORCE) {
+            // Send the the force setpoint (local pos sp external message)
+            float dcm[3][3];
+            mavlink_euler_to_dcm(roll, pitch, yaw, dcm);
+            const float fx = -dcm[0][2] * thrust;
+            const float fy = -dcm[1][2] * thrust;
+            const float fz = -dcm[2][2] * thrust;
+            uint16_t typeMask = (3<<10)|(7<<3)|(7<<0)|(1<<9); // select only FORCE control (disable everything else)
```
Susurrus: This is a super-confusing way to write out a typemask. It'd be better to just write it out as hex if all the parts don't just look like `1<<X`, especially since these bits aren't assigned in a logical order.

src/uas/UAS.cc
```diff
@@ -2410,56 +2410,198 @@ void UAS::setManualControlCommands(float roll, float pitch, float yaw, float thr
...
+                        (!isnan(yaw) && yaw != manualYawAngle) || (!isnan(thrust) && thrust != manualThrust) ||
+                        buttons != manualButtons) {
+                    sendCommand = true;
+
+                    // Ensure that another message will be sent the next time this function is called
+                    countSinceLastTransmission = 10;
+                }
+
+                // Now if we should trigger an update, let's do that
+                if (sendCommand) {
```
Susurrus: Shouldn't be this be done for all joystick modes?

src/uas/UAS.cc
```diff
@@ -2410,56 +2410,198 @@ void UAS::setManualControlCommands(float roll, float pitch, float yaw, float thr
...
+                    0,
+                    0
+                    );
+        } else if (joystickMode == JoystickInput::JOYSTICK_MODE_VELOCITY) {
+            // Send the the local velocity setpoint (local pos sp external message)
+            static float vx = 0;
+            static float vy = 0;
+            static float vz = 0;
+            static float yawrate = 0;
+            //XXX: find decent scaling
```
Susurrus: Can determining these scalings only be done once this commit is merged or are they actually not necessary? Ideally this and the other XXX comments are resolved one way or another before its merged.

Susurrus:
Have manual control been tested with this PR? It's really important to my project, but I'm defending next week and my dissertation is due after that so I don't have time to test this PR on my own within the next couple of weeks.

src/uas/UAS.cc
LorenzMeier: No, because the other modes send direct set points to the vehicle, which has an internal timeout. The joystick is acting like the high-level companion computer in these modes, and much higher updates rates are expected by the safety timeout logic in those cases.

LorenzMeier:
@Susurrus Testing now, but I don't recommend to use master for thesis defenses in general ;-).

LorenzMeier:
Sweet. Works as advertised!

LorenzMeier:
@Susurrus Thanks for the review! I went for the current type mask since I still found it easier to interpret than plain hex. I can transform small hex numbers in my head, but not a full mask.

thomasgubler:
:clap: thanks for getting this merged!

PR #1147, opened 2014-12-25
Load/Save connection settings - incomplete review

crashmatt:
Status
The comm ports can now be stored and recalled in any quantity and order.  Code is based on a relatively recent master using Qt5.3.2.  Changes are mostly the minimum to make it work.

Auto-connection does not work yet.  I am not sure how to decide which ports should re-connect.  It can either remember shich ports were open on shutdown or add a menu option.  Open to suggestions.

Settings are stored on shutdown so crashed settings are not remembered.  It is possible to serialize a connection when it is created but that is more complex to manage.

Mavlink protocol has not been tested, only opening a link.  Only UDP and serial are tested.
  

Significant changes:
Both LinkManager and LinkInterface are a QGCSettingsGroup.  The group serializes itself and uses a link to its parent to get the full settings path.  This means groups can be moved around without a problem.  I have not considered how multipe settings files will work.

The link ID number is used as an identifier for serialization and is persistent. New Link IDs are now set as the next avaialble link number, not the auto-increment it used to be.

There is now a LinkInterface.cc.  There was quite a bit of repeated functionality in the children that should be in LinkInterface.  For some reason it would not allow me to put this in the header.  Re-re-re-comping the header was getting very tiresome so LinkInterface.cc was born.  Not commented yet.


crashmatt:
Too many changes too quickly.  By the time I finish test there is the same amount of merging and testing to do.  Please delete the merge request.  I will do a small patch to hardcode the behaviour I need.

DonLakeFlyer:
I’ve been trying to push a lot of things through before I headed out of town for a while. Now that I’m gone, the codebase should stay static for a while. So good time to wring things out without getting smacked around every time I check in.

DonLakeFlyer:
@crashmatt I'm back in town after being gone forever. Where are things at with this?

crashmatt:
Don,
I have stopped work on QGC to work on a PX4 variant for a new board.  Will
also be moving house in a few weeks so I don't expect to have any time for
QGC soon.

The difficulty making the merge indicates that I made the modification
unnecessarily complex.  Probably best to go with something simpler which is
easier to test.

/Matt

DonLakeFlyer:
"The difficulty making the merge indicates that I made the modification unnecessarily complex.  Probably best to go with something simpler which is easier to test."

Not really. I run into this all the time with seemingly simply changes. If I find some time I'll try to clean it up for you.

DonLakeFlyer:
@dogmaphobic This can be closed now since your change saves settings?

dogmaphobic:
> This can be closed now since your change saves settings?

Yes. The changes in #1234 take care of this, so this PR should be discarded.

PR #1155, opened 2014-12-28
Rc param tune

thomasgubler:
A first test version of the UI for configuring the RC to Parameter mapping: The process is initiated by the user by opening the new context menu in the parameter view. The dialog window then makes sure that the param value which the user sees is up to date and lets the user configure the options before sending the mapping via mavlink.

screenshot from 2014-12-28 23 46 52

screenshot from 2014-12-28 23 35 15

More docs will follow.

mavlink PR: https://github.com/mavlink/mavlink/pull/309
px4 PR: https://github.com/PX4/Firmware/pull/1543

This will fail teamcity because of the mavlink update.  As soon as the mavlink PR is in I will update the submodule here.

### DIFF
--- a/src/qgcunittest/MockUAS.h
+++ b/src/qgcunittest/MockUAS.h
@@ -1,24 +1,24 @@
 /*=====================================================================
- 
+
  QGroundControl Open Source Ground Control Station
- 
+
  (c) 2009 - 2014 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
- 
+
  This file is part of the QGROUNDCONTROL project
- 
+
  QGROUNDCONTROL is free software: you can redistribute it and/or modify
  it under the terms of the GNU General Public License as published by
  the Free Software Foundation, either version 3 of the License, or
  (at your option) any later version.
- 
+
  QGROUNDCONTROL is distributed in the hope that it will be useful,
  but WITHOUT ANY WARRANTY; without even the implied warranty of
  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  GNU General Public License for more details.
- 
+
  You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


src/uas/UAS.cc
```diff
@@ -3467,3 +3467,45 @@ void UAS::stopLowBattAlarm()
...
+{
+    qDebug() << "sendMapRCToParam" << param_id << "scale" << scale << "value0" << value0 << "param rc chan index" << param_rc_channel_index;
+    mavlink_message_t message;
+
+    mavlink_msg_param_map_rc_pack(mavlink->getSystemId(),
+                                  mavlink->getComponentId(),
+                                  &message,
+                                  this->uasId,
+                                  0,
+                                  param_id.toStdString().c_str(),
```
thomasgubler: fixed in f48f62cf50e26191917b76d34cb5b7c7ac79dd94

src/uas/UAS.cc
```diff
@@ -3467,3 +3467,45 @@ void UAS::stopLowBattAlarm()
...
+{
+    qDebug() << "unsetRCToParameterMap";
+    for (int i = 0; i < 3; i++) {
+        mavlink_message_t message;
+        mavlink_msg_param_map_rc_pack(mavlink->getSystemId(),
+                                      mavlink->getComponentId(),
+                                      &message,
+                                      this->uasId,
+                                      0,
+                                      "",
```
thomasgubler: fixed as well in f48f62cf50e26191917b76d34cb5b7c7ac79dd94

thomasgubler:
I see now that teamcity reported some reorder errors which didn't show when building locally.

thomasgubler:
Release build pass TC. Debug builds fail because the MockUAS class needs to implement some of the new functions in the base class.

thomasgubler:
ef7e777 and 29d0fa4 fixed the remaining issues. This passes now TC, ok to merge?

PR #1159, opened 2014-12-30
Main window central view and dock widget re-architecture

DonLakeFlyer:
This is a pretty major change to both how the central view in MainWindow works as well as DockWidgets. The initial driver was to fix issue with QML not working as a MainWindow central widget. Although not fully proved, I believe the issue is associated with the central view being a stacked widget and the contents of the stacked widgets being child QMainWindow's. I think these nested QMainWindow's were confusing Qt. In general the previous setup seemed like a dangerous hack usage of QMainWindow.

Here is what is different:
* The central view is no longer a stacked widget. A single view is added to the central widget layout and removed as needed.
* Central Views (Setup, Plan, Fly, ...) are only created as they are needed instead of creating all of them at boot time.
* There is now a single set of Dock Widgets which are used by all views and available to all views.
* DockWidgets are only created when they are needed, not all at boot.
* Each view still has it's own seperate saved window state and set of visible dock widgets just like it did before.
* The change was so major I had to bump the settings version which is going to cause previous settings to be tossed. I tried to avoid this, but got too paranoid with such a major change that it would just lead to subtle bugs which were just saved settings issues.
* Almost forgot the thing that may annoy people (other than having to trash previous settings): I got rid of the "Advanced Mode" and "Show Titlebars" settings. It now always shows title bars and you can always drag to re-order widgets. I would like to try it this way to simplify random settings and menu items. If it makes people crazy I'll put it back the way it was before. Keep in mind that I have a plan to completely re-do the main window widget infrastructure to be able to have a more 21st Century ui where the widgets float above the main view instead of taking up space from it. This will lead to a much more usable UI for both PC and tablet. I removed these settings to start working my way towards that new ui.

I ran memory usage numbers on OS X by booting to Fly view and connecting:
* Before change: 131M
* After change: 82M

So as you can see a really significant reduction in memory usage. CPU usage show go down as well since there are fewer duplicate widgets chewing up CPU. Fingers crossed that this will fix the QML main window problems as well.


thomasgubler:
One thing I noticed while using this version: the configuration views open now as a pop-up instead of embedded in the setup tab. This doesn't bother me for one time things like the RC config. However it was always quite handy to have the setup tab open with the full parameter list so that I could quickly switch to it from the fly view. With this version I always have to open and close that pop up and navigate again through the parameter tree.

LorenzMeier:
+1 - I don't think Pop-ups are a great idea from a general UI design perspective.

DonLakeFlyer:
Agree. Majority of this is temporary setup to begin the switch to qml. The key is that a number of the setup screens need to be modal. In that change I pulled out the parameter view stuff. I'll figure out a better way to do it. Take me a while because I'm out of town for a month

bansiesta:
+1, makes QGC usable for me again!

DonLakeFlyer:
Here's my plan: Fix up the merge so I can push this through. Then I'm going to switch back to the left hand button bar, instead of using the summary view to get to config screens. This will get rid of pop-ups. This will also put back Firmware upgrade and parameters top level config buttons so you can leave setup sitting on the parameters screen. 

I'll figure out some other way to handle the modality issues. My specific concern is to not allow user to go do something else in the middle of doing a calibration for example.

DonLakeFlyer:
Ran into a nasty merge in SettingsDialog.ui. I'm pushing through a bogus fix for that merge in this pull which will break master. I'm still going to merge so I can get the rest through. I will then submit a separate pull to fix that.

chzw517:
it ran into another problem, when i click the setup button, it shutdown by itself.

here is some message showed in the terminal

Orphaned log file count 0
qrc:/qml/SetupViewButtons.qml:26:13: Type SetupButton unavailable
qrc:///qml/QGroundControl/Controls/SetupButton.qml:4:1: module "QtGraphicalEffects" is not installed
QObject::connect: Cannot connect (null)::setupButtonClicked(QVariant) to SetupView::_setupButtonClicked(QVariant)
QObject::connect: Cannot connect (null)::firmwareButtonClicked() to SetupView::_firmwareButtonClicked()
QObject::connect: Cannot connect (null)::parametersButtonClicked() to SetupView::_parametersButtonClicked()
QObject::connect: Cannot connect (null)::summaryButtonClicked() to SetupView::_summaryButtonClicked()

chzw517:
chzw517@XPS13:~/uav/qgc/qgroundcontrol$ git pull
Already up-to-date.
chzw517@XPS13:~/uav/qgc/qgroundcontrol$ date
Thu Feb  5 22:43:24 HKT 2015

DonLakeFlyer:
What OS and what Qt version?

chzw517:
Os: ubuntu15.04 alpha2
Qt:  5.3.2

DonLakeFlyer:
@chzw517 Can you look in your qt install directory. On my ubuntu its: /usr/lib/x86_64-linux-qnu/qt5. There should be a qml directory under there. Look in that directory to see if there is a QtGraphicalEffects subdirectory and let me know.

chzw517:
chzw517@XPS13:/usr/lib/x86_64-linux-gnu/qt5/qml$ ls -l
total 8
drwxr-xr-x 5 root root 4096 Jan 30 20:59 QtQuick
drwxr-xr-x 2 root root 4096 Jan 30 20:59 QtQuick.2

so sad``` i don't have it... 
then what lib should i install?

thank you very much...

chzw517:
ok fix it... after install this package "qml-module-qtgraphicaleffects"

DonLakeFlyer:
Awesome thanks for figuring this out. I'll update readme.

chzw517:
screenshot from 2015-02-10 17 32 26


black again...any ideal?

DonLakeFlyer:
@chzw517 No strange debug output this time? I'll have a pull in a few minute you can try.

DonLakeFlyer:
@chzw517 Can you try pull #1220?

PR #1178, opened 2015-01-21
Made Q3D emit updateWidget so that local view updates correctly.

jgoppert:
Local 3D view was completely unusable because updateWidget was never being called after it was renamed from update. The signal was not being emitted by Q3D.


jgoppert:
Does anyone have an issue with merging this, it makes the local 3d view unusable as is.

DonLakeFlyer:
Resubmitting this through TeamCity now that it is up again.

DonLakeFlyer:
Seems reasonable. I'll merge after it passes TC.

PR #1184, opened 2015-02-02
Remove erroneous use of QFileDialog::selectNameFilter()

dogmaphobic:
No filters are to be selected. The argument selectedFilter is a pointer to a QString to receive the filter selected/used by the user.

### DIFF
--- a/src/QGCFileDialog.cc
+++ b/src/QGCFileDialog.cc
@@ -101,9 +101,6 @@ QString QGCFileDialog::getSaveFileName(QWidget* parent,
     {
         QFileDialog dlg(parent, caption, dir, filter);
         dlg.setAcceptMode(QFileDialog::AcceptSave);
-        if (selectedFilter) {
-            dlg.selectNameFilter(*selectedFilter);
-        }
         if (options) {
             dlg.setOptions(options);
         }


DonLakeFlyer:
Don't you need to pull the selected filter using selectedNameFilter after the exec? Doesn't look like selectedFilter is returned in the save case.

dogmaphobic:
Yes, that is what should be done if the *selectedFilter* argument was being used. As it is, it is not used. In fact, there is code in there asserting if it is ever defined.

dogmaphobic:
> Yes, that is what should be done if the selectedFilter argument was being used. As it is, it is not used. In fact, there is code in there asserting if it is ever defined.

Replying to myself... With that said, should I go ahead and create another issue to fix this *selectedFilter* argument never allowed? It's mostly a function of the caller to make use of it. Based on the comments on the code, it seems it is not allowed to be used for the time being because of the lack of unit testing. If that's all there is to it, I can fix it all in one swoop...

DonLakeFlyer:
I was originally trying to mirror the Qt open file apis which the QGC versions. Given the fact that nobody is using that parameter how about removing it completely from the APIs. Less change for buggy implementation without testing then.

Susurrus:
Let's remove it completely. It's a minor change to do that and I would consider it worthwhile.

dogmaphobic:
> Let's remove it completely. It's a minor change to do that and I would consider it worthwhile.

Will do. Creating an issue for it...

Issue #1182, opened 2015-01-31
File extensions need consistency.


dogmaphobic:
Unlike something such as a text editor, where you want full freedom to define file extensions, most, if not all extensions used by files written and read by QGroundControl are very specific. The extensions used to read and write these files should be consistent and enforced where applicable.

This is primarily an issue when creating files. The function used to request a filename from the user should have an option to define if the extension is mandatory (strict), in which case the function will make sure only valid extension(s) are used.

Susurrus:
@thomasgubler I'd like to pull you into this issue. #1173 brought up the important point of whether we should allow users to rename file extensions, as most of them are very specific.

We deal with the following extensions for user-created files:
 * .mavlink - MAVLink log files. Binary & timestamped.
 * .qgw - Custom tool widget created by the user.
 * .log - Log files saved the LineChartWidget.
 * .txt - Waypoint file -  Text, possibly tab-delimited.
 * .svg, .pdf - Image files from the QGCDataPlot2D **unsure what this is**.
 * .csv, .txt - Data from the QGCDataPlot2D widget
 * .txt - Parameter file

I'd like to make the proposal to use only a single file extension per feature above (except the image saving format) and make it mandatory to use these proper extensions, unlike how it's done now.

I'd like to propose the following strict file extension naming scheme:
 * .mavlink - No change
 * .qgw - No change
 * .log - Log files saved the LineChartWidget **possibly rename this to CSV or TSV depending**
 * .txt (Waypoint file) - **Rename to .waypoints**
 * .svg, .pdf - No change
 * .csv, .txt (QGCDataPlot2D output) - **Make only .csv**
 * .txt (Parameter file) - **Rename to .params**

I'd really prefer a .csv format for the waypoints and parameters files, but I don't think that's worth doing at this stage. At least doing it this way will make things more consistent and easier to differentiate the different output files QGC can generate.

Susurrus:
@LorenzMeier Since this is a big user-facing change, can you weigh in here?

dogmaphobic:
I will start working with the plumbing. Adding a flag to the openFileSave telling it the extension is strict and within the function, enforcing the extension. Then it’s just a matter of figuring out the places where this should be enforced and what extensions to use in those cases.

thomasgubler:
@Susurrus I am fine with the file ending changes. They seem reasonable. I suggest that for compatibility we still allow all file endings in the file open dialogs, agree?

Susurrus:
[...]

PR #1188, opened 2015-02-04
File extensions need consistency. Issue #1182.

dogmaphobic:
Fixes issue #1182 

Added a new flag to the QGCFileDialog::getSaveFileName() telling it to enforce a given file type (suffix/extension). It works in tandem with the default suffix argument I added earlier.

If a default suffix is given, it will be appended to the filename if the user does not enter one themselves. That is, if the user simply enters *foo* and the default suffix is set to *bar*, the returned filename will be *foo.bar*. However, if the user specifies a suffix, the *strict* flag will determine what is done. If the user enters *foo.txt* and *strict* is false, the function returns *foo.txt* (no suffix enforced). If strict is true however, the default suffix is appended no matter what. In the case above, the function will return *foo.txt.bar* (suffix enforced).
 
If *strict* is set and the file name given by the user is renamed (the *foo.txt.bar* example above), the function will check and see if the file already exists. If that's the case, it will ask the user if they want to overwrite it.

The current file types handled by this:

* .mavlink 	- Flight Data Log - Enforced
* .qgw 		- Widgets - Enforced
* .log 		- Log files saved by LineChartWidget - Enforced
* .waypoints	- Enforced
* .csv		- QGCDataPlot2D output - Enforced
* .params		- Parameter files - Enforced
*  .svg .pdf	- Plot Files - Not enforced. The default is pdf.

I left the extension used in LinechartWidget::startLogging() above set to “log” (as it was) because I felt saving it as a “csv” would confuse it with the files saved by QGCDataPlot2D. These files are not consumed (not loaded) anywhere in QGC.

In MainWindow.cc (~ 839) there is a function to “startVideoCapture()”, which is using “bmp” as the default extension. I don’t know yet what this function is saving and what type of “video” could be saved into a bmp.

**Note**: Not tested under Windows. I’m curious to see how “All Files (*)” behave there.


Susurrus:
Are you sure that QGCDataPlot2D outputs valid CSV data? I never looked into it, so I'm relying on your inspection here.

And does the LineChartWidget output valid CSV? If it does, then it should be using the .csv file extension. .csv is a standard file format, so it doesn't matter what generates it in that regard.

dogmaphobic:
Ha! It would help if I knew what the heck those Widgets are doing. That’s why I started with mundane tasks so I could get an understanding of it all

QGCDataPlot2D has close to no documentation. It seems it is only visible before you connect. I assume you are able to load something and analyze it. It handles both binary (.raw or .imu) files as well as .csv files. Note that .csv here is a broad term. The “c” is one of many possible separators it understands. The loader will smartly identify and handle tabs, commas, semi-colons, spaces, tildes, and pipe symbols as separators.

LineChartWidget saves a “space” separated file.

With that said, I’m more than happy to have it do and save whatever makes sense. For now I’m relying on people here who actually have a full grasp of the code to tell me what needs to be done. As I said, I started with mundane tasks so I could get going and begin to understand the code. I’m working on it. It’s only been less than a week. The general UI is quite complex from both an engineering as well as from an user point of view. A great part, if not the biggest part in my interest in this project is to make it easy to use and understand (and reliable). I have over 25 years of experience designing very complex systems with a rather large user base. I’m volunteering my time but obviously with a self interest in wanting something that works. From my research, I felt this to be the most viable project to get somewhere before DJI completely dominates this market. Eventually I will be able to contribute more meaningfully.

DonLakeFlyer:
I wouldn't get too hung up on going with strict extensions for everything. If there is confusion around actual usage or format then leave it as non-strict for now and people can do what they want.

On your usability comment: This is our major focus right now. We are in the process of working with a UI designer to make this better. First area we are working on are Setup screens. If you would like to join in on this that would be awesome. It's going a bit slow right now since I'm the only one working on it.

Susurrus:
> I wouldn't get too hung up on going with strict extensions for everything. If there is confusion around actual usage or format then leave it as non-strict for now and people can do what they want.

The problem with leaving it not well specified is that this is what **causes** the confusion. QGC is bad about specifying the format of the files that it handles, both from a saving and a loading perspective. I suggest we actually work to tackle this and just make sure loading works for all the old formats we support but enforce strict rules for saving things. The point is that we want to have a single set of files that everyone understands, but allowing people to save file extensions however we want we make it harder for people who don't. It's the classic strict-but-consistent versus flexible-but-vague dilemma. And since I see post after post to the QGC mailing list by newbies trying to understand the system, I suggest we lean more towards being strict and consistent. It works best for new users until they understand the system. And right now QGC should be targeted towards those users. I'm sick of hearing posts on the mailing list saying how APM works and QGC doesn't. Let's change that!

Now onto the widgets. QGCDataPlot2D just reads files, correct? If it does, then there's no issues with nailing that down in this PR.

But for the LineChartWidget, I'd actually suggest we move it over to using comma-separated files for saving things. This has the widest support among 3rd-party tools (Excel, MATLAB, Python, etc.). I've never heard of a program saving space-separated values. **But** let's leave this as a separate PR so it can be resolved independently. Let's just leave the LineChartWidget as saving .log files and **not** be strict about enforcing that line ending.

Can the LineChartWidget load files too? If it does, can it only handle it's own .log format or is it as flexible as the QGCDataPlot2D tool? (man I wish this was a separate tool).

DonLakeFlyer:
I don't disagree with any of that, but from a priority standpoint, changing code like LineChartWidget to csv is way down the list. That's power use functionality already. I'd rather see code changes in basic things like setup, connection and flight first.

Susurrus:
> I don't disagree with any of that, but from a priority standpoint, changing code like LineChartWidget to csv is way down the list. That's power use functionality already. I'd rather see code changes in basic things like setup, connection and flight first.

I agree with you and that's why it's going to be left to a separate issue/PR as I mentioned in my post. I think it's worthwhile getting this stuff nailed down nicely and deferring on the LinechartWidget. I'd also like to point out #1189 to follow up on my comment on the QGCDataPlot2D tool.

dogmaphobic:
> On your usability comment: This is our major focus right now. We are in the process of working with a UI designer to make this better. First area we are working on are Setup screens. If you would like to join in on this that would be awesome. It's going a bit slow right now since I'm the only one working on it.

Usability issues implies something is usable but has issues 8-P To be honest, I have not been able to fully setup my Pixhawk yet (let alone fly it). I managed to go through the firmware upload and *some* radio setup but not much more than that. My main issue is that I don’t want to lug the aircraft (a hexacopter) to my workstation. Working on my laptop is a pain. I have not yet been able to connect the 3DR Radio Modems. QGC simply ignores it, but that’s for another thread (the whole serial comm code needs a lot of TLC).

I really want this to work! I have a bit of time now and I want to get as much I can done before I get busy again. Yes, obviously I want to join ( haven’t I already? :)

dogmaphobic:
QGCDataPlot2D will load either binary files (.imu or .raw) or ascii files (.csv). When you load a binary file, it converts to csv internally (it seems to be the case at least). You can mess with the data and save it back out. It always saves it as .csv. In addition, it also allows you to save the plot itself (graph). It saves it as either .pdf or .svg but that’s a separate story.

LinechartWidget itself doesn’t load anything. It only saves. I do not know if something else loads its files though.

I’m changing LinechartWidget save **not** to force .log and hopefully we can close this with this one change.

While at it, allow me to rant a bit :) There seems to be no coding guidelines concerning class member variable names. That makes it really confusing to follow the code as I can’t tell what is local scope or what is a class member variable. These member variables should be made easy to identify (*m_variable* instead of *variable* as an example). If it’s ok, I will start changing these in the files I work on.

DonLakeFlyer:
“I have not been able to fully setup my Pixhawk yet (let alone fly it).”

Exactly what we are working on right now in a separate thread. Many people are failing on basic setup. If you’d like to contribute coding in that space send me your real email address and I’ll include you on the threads. I can help you with 3DR radios on a separate thread as well. Then we can collect that information as feedback on how to fix the radio usability.

dogmaphobic:
> send me your real email address

No wanting to really put it on a public space but if you get my name, put an @ in between and a .com at the end, you will have it :)

DonLakeFlyer:
I’ve been writing all wholly new code with a consistent style. The style for member variables is _variable for example. If you look at any brand new files, like everything under src/FactSystem you can see what the style looks like and should be able to follow along. I need to update the wiki coding guidelines. I’ll try to get to that in the next couple days. For now you can use the examples as guidelines.

DonLakeFlyer:
I gave it a try. You can also email me at don@thegagnes.com.

dogmaphobic:
> I gave it a try.

I got it and replied to it.

DonLakeFlyer:
Linux builds are failing. You can click on the Details link to download the build logs. Looks like you've got a #include "QFoo.h" instead of a #include <QFoo> in there.

dogmaphobic:
```
src/QGCFileDialog.cc:26:32: fatal error: QRegularExpression.h: No such file or directory
#include "QRegularExpression.h"
```

No idea. I'll make a blind attempt by using

```
#include <QRegularExpression>
```

instead but I won't be able to really check until later (leaving for dinner).

DonLakeFlyer:
That's it. Qt headers should always be #include <QFoo>. If you use the QFoo.h it may or may not work correctly.

DonLakeFlyer:
@dogmaphobic Now that this passes TC this is good to go right?

dogmaphobic:
Yes, I haven't touch it in a week. It should be fine.

PR #1191, opened 2015-02-04
Back to non-modal, no pop-up Setup View

DonLakeFlyer:
- Setup screens now open in place
- Left hand setup view buttons are now implemented in QML and are closer to the UI spec
- Setup buttons show setup state with red/green indicators

There is still a usability issue with trying to click a setup button while armed, or when one setup step must be performed before another. The user is warned correctly and setup is disallowed, but that clicked setup button still stays selected. Currently the setup view is still a hybrid QML and QWidget implementation. Once all setup screens are cut over to QML I will switch this to a full QML implementation and fix this problem.

@thomasgubler Could you check this pull to make sure it hasn't brought back the QML "Black screen of death".

Here is what the left hand button bar looks like now:
screen shot 2015-02-04 at 10 24 47 am


thomasgubler:
@DonLakeFlyer cool, this works on arch linux.


I noticed that you set `NAV_DLL_N` in the safety screen. This parameter only has a incluence if `NAV_DLL_OBC` = 1. I suggest configuring only `COM_DL_EN` and `COM_DL_LOSS_T` in this screen.

DonLakeFlyer:
Thanks @thomasgubler. I'll take a lookout safety again to see what I did there. I'm going to move that into a separate issue so I don't forget.

PR #1197, opened 2015-02-04
Add Light/Dark theme support to QML

DonLakeFlyer:
Adds support for QML ui to respect a Light/Dark style palette. Initial palette colors are basic light/dark style similar to exist Widget based light/dark style. Colors will be better defined as we work with UI designers. This may also require adding additional color properties to the palette at some point.


DonLakeFlyer:
After this goes through I think we are ready to release another Stable before we move into the Setup UI redesign.

PR #1199, opened 2015-02-05
Coding style examples

DonLakeFlyer:
These are example header and code files for QGroundControl coding style. They basically match the style that I've been using for the last year for all the new code I've been writing. They are not meant to be comprehensive, just cover the high level stuff. Don't really want to enforce a really detailed coding style.

Comments on whether something important is missing.

### DIFF
--- a/CodingStyle.cc
+++ b/CodingStyle.cc
@@ -1,0 +1,96 @@
+/*=====================================================================
+ 
+ QGroundControl Open Source Ground Control Station
+ 
+ (c) 2009 - 2014 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ 
+ This file is part of the QGROUNDCONTROL project
+ 
+ QGROUNDCONTROL is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+ 
+ QGROUNDCONTROL is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+ 
+ You should have received a copy of the GNU General Public License
+ along with QGROUNDCONTROL. If not, see <http://www.gnu.org/licenses/>.
+ 
+ ======================================================================*/
+
+// The above copyright block should be at the top of every file.
+
+/// @file
+///     @brief This is an example class c++ file which is used to describe the QGroundControl
+///             coding style. In general almost everything in here has some coding style meaning.
+///             For example, note how this doxygen block uses the /// variant as the preferred style
+///             of doxygen marker. Not all style choices are explained.
+///
+///     @author Don Gagne <don@thegagnes.com>
+
+#include "CodingStyle.h"
+#include "QGCApplication.h"
+
+#include <QFile>
+#include <QDebug>
+
+// Note how the Qt headers and the QGroundControl headers above are kept seperate
+
+Q_LOGGING_CATEGORY(CodingStyleLog, "CodingStyleLog")
+
+CodingStyle::CodingStyle(QObject* parent) :
+    QObject(parent),
+    _protectedVariable(1),
+    _privateVariable1(2),
+    _privateVariable2(3)
+{
+
+}
+
+/// Document non-obvious private methods in the code file.
+void CodingStyle::_privateMethod(void)
+{
+    // Always include braces even for single line if/for/...
+    if (uas != _uasId) {
+        return;
+    }
+    
+    // Note the brace placement
+    if (_lastSeenComponent == -1) {
+        _lastSeenComponent = component;
+    } else {
+        Q_ASSERT(component == _lastSeenComponent);  // Asserts are your friend
+    }
+}
+
+void CodingStyle::_privateMethod2(void)
+{
+    Fact* fact = qobject_cast<Fact*>(sender());
+    Q_ASSERT(fact);
+    
+    QVariant typedValue;
+    switch (fact->type()) {
+        case FactMetaData::valueTypeInt8:
+        case FactMetaData::valueTypeInt16:
+        case FactMetaData::valueTypeInt32:
+            typedValue.setValue(QVariant(value.toInt()));
+            break;
+            
+        case FactMetaData::valueTypeUint8:
+        case FactMetaData::valueTypeUint16:
+        case FactMetaData::valueTypeUint32:
+            typedValue.setValue(value.toUInt());
+            break;
+            
+        case FactMetaData::valueTypeFloat:
+            typedValue.setValue(value.toFloat());
+            break;
+            
+        case FactMetaData::valueTypeDouble:
+            typedValue.setValue(value.toDouble());
+            break;
+    }
+}
--- a/CodingStyle.cc
+++ b/CodingStyle.cc
@@ -2,7 +2,7 @@
  
  QGroundControl Open Source Ground Control Station
  
- (c) 2009 - 2014 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ (c) 2009 - 2015 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
  
  This file is part of the QGROUNDCONTROL project
  
@@ -37,9 +37,13 @@
 #include <QFile>
 #include <QDebug>
 
+#include <math.h>
+
 // Note how the Qt headers and the QGroundControl headers above are kept seperate
 
 Q_LOGGING_CATEGORY(CodingStyleLog, "CodingStyleLog")
+
+const int CodingStyle::_privateStaticVariable = 0;
 
 CodingStyle::CodingStyle(QObject* parent) :
     QObject(parent),
@@ -85,12 +89,24 @@ void CodingStyle::_privateMethod2(void)
             typedValue.setValue(value.toUInt());
             break;
             
-        case FactMetaData::valueTypeFloat:
+        case FactMetaData::valueTypeFloat: {
+            int localScopedVar = 1;
             typedValue.setValue(value.toFloat());
+        }
             break;
             
         case FactMetaData::valueTypeDouble:
             typedValue.setValue(value.toDouble());
             break;
     }
+}
+
+void CodingStyle::_methodWithManyArguments(QWidget*         parent,
+                                           const QString&   caption,
+                                           const QString&   dir,
+                                           Options          options1,
+                                           Options          options2,
+                                           Options          options3)
+{
+    // Implementataion here...
 }
--- a/CodingStyle.h
+++ b/CodingStyle.h
@@ -39,16 +39,22 @@
 #include <QXmlStreamReader>
 #include <QLoggingCategory>
 
+#include <limits.h>
+
 #include "Fact.h"
 #include "UASInterface.h"
 
-// Note how the Qt headers and the QGroundControl headers above are kept seperate
+// Note how the Qt headers, Syste, headers and the QGroundControl headers above are kept in seperate groups
 
 // If you are going to use a logging category for a class it should have the same name as the class
 // with a suffix of Log.
 Q_DECLARE_LOGGING_CATEGORY(CodingStyleLog)
 
-/// Here is the class documentation. Class names are PascalCase.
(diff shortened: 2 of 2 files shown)


DonLakeFlyer:
My plan is to then link the wiki to these for coding style examples.

Susurrus:
I'd prefer if we didn't commit useless code into the repo and leave it purely on the wiki or the website.

CodingStyle.cc
```diff
@@ -0,0 +1,96 @@
+/*=====================================================================
+ 
+ QGroundControl Open Source Ground Control Station
+ 
+ (c) 2009 - 2014 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
```
thomasgubler: 2015 :)

CodingStyle.h
```diff
@@ -0,0 +1,93 @@
...
+
+/// @file
+///     @brief This is an example class header file which is used to describe the QGroundControl
+///             coding style. In general almost everything in here has some coding style meaning.
+///             For example, note how this doxygen block uses the /// variant as the preferred style
+///             of doxygen marker. Not all style choices are explained.
+///
+///     @author Don Gagne <don@thegagnes.com>
+
+#ifndef CodingStyle_h
```
Susurrus: Why aren't these upper-cased? They're macro constants which should always be upper-cased. Additionally underscores should separate words in this instance, so please change to `CODING_STYLE_H`.

CodingStyle.h
```diff
@@ -0,0 +1,93 @@
...
+private slots:
+    void _privateSlot(void);
+    
+private:
+    // Private methods and variable names begin with an "_". Documentation for
+    // non-obvious private methods goes in the code file, not the header.
+    void _privateMethod(void);
+    
+    /// Document non-obvious variables in the header file. Long descriptions go here.
+    int _privateVariable1;
```
Susurrus: Don't use ints. They should almost never be used. Change this to uint16_t or something like that.

CodingStyle.cc
```diff
@@ -0,0 +1,96 @@
...
+#include "QGCApplication.h"
+
+#include <QFile>
+#include <QDebug>
+
+// Note how the Qt headers and the QGroundControl headers above are kept seperate
+
+Q_LOGGING_CATEGORY(CodingStyleLog, "CodingStyleLog")
+
+CodingStyle::CodingStyle(QObject* parent) :
```
Susurrus: Also add a function that takes a ton of arguments that follows this same formatting for the function arguments. There was a question about that in a recent PR.

CodingStyle.h
```diff
@@ -0,0 +1,93 @@
...
+
+// Note how the Qt headers and the QGroundControl headers above are kept seperate
+
+// If you are going to use a logging category for a class it should have the same name as the class
+// with a suffix of Log.
+Q_DECLARE_LOGGING_CATEGORY(CodingStyleLog)
+
+/// Here is the class documentation. Class names are PascalCase.
+
+class CodingStyle : public QObject
```
Susurrus: Add a class static constant. It's been unclear to me how those should be dealt with.

CodingStyle.h
```diff
@@ -0,0 +1,93 @@
...
+#include <QObject>
+#include <QMap>
+#include <QXmlStreamReader>
+#include <QLoggingCategory>
+
+#include "Fact.h"
+#include "UASInterface.h"
+
+// Note how the Qt headers and the QGroundControl headers above are kept seperate
+
```
Susurrus: Add some macro constants and functions to give some guidance there.

CodingStyle.h
```diff
@@ -0,0 +1,93 @@
...
+Q_DECLARE_LOGGING_CATEGORY(CodingStyleLog)
+
+/// Here is the class documentation. Class names are PascalCase.
+
+class CodingStyle : public QObject
+{
+    Q_OBJECT
+    
+public:
+    CodingStyle(QObject* parent = NULL);
```
Susurrus: Needs a Doxygen comment. There's none of that in this entire file and we really should be trying to follow that standard. Once we do we can auto-generate documentation to put on the website!

CodingStyle.h
```diff
@@ -0,0 +1,93 @@
...
+    ~CodingStyle();
+    
+    /// Document public methods which are non-obvious in the header file
+    bool publicMethod1(void);
```
Susurrus: Needs a Doxygen comment.

CodingStyle.cc
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+void CodingStyle::_privateMethod2(void)
+{
+    Fact* fact = qobject_cast<Fact*>(sender());
+    Q_ASSERT(fact);
+    
+    QVariant typedValue;
+    switch (fact->type()) {
```
Susurrus: Add a `default:` specifier. **All** switch statements should include a default specifier, even if it does nothing. This helps alleviate potential errors (same reason to **always** use curly-braces around if-statements).

CodingStyle.cc
```diff
@@ -0,0 +1,96 @@
...
+        case FactMetaData::valueTypeInt8:
+        case FactMetaData::valueTypeInt16:
+        case FactMetaData::valueTypeInt32:
```
Susurrus: Add styling for situations where you need to create a local variable inside a case statement? In C you need curly-brackets around it to give an inner scope to link the variable to, don't know if that's required for C++.

CodingStyle.cc
```diff
@@ -0,0 +1,96 @@
...
+ along with QGROUNDCONTROL. If not, see <http://www.gnu.org/licenses/>.
+ 
+ ======================================================================*/
+
+// The above copyright block should be at the top of every file.
+
+/// @file
+///     @brief This is an example class c++ file which is used to describe the QGroundControl
+///             coding style. In general almost everything in here has some coding style meaning.
+///             For example, note how this doxygen block uses the /// variant as the preferred style
```
Susurrus: Why use the single-line comment? Most programs auto-generate the following:
```
/**
 * Comment here
 * @param ...
 */
```

CodingStyle.cc
```diff
@@ -0,0 +1,96 @@
...
+///
+///     @author Don Gagne <don@thegagnes.com>
+
+#include "CodingStyle.h"
+#include "QGCApplication.h"
+
+#include <QFile>
+#include <QDebug>
+
+// Note how the Qt headers and the QGroundControl headers above are kept seperate
```
Susurrus: You should also include standard library headers in a 3rd group. That keeps things rather clear.

I also like to alphabetize the list, but that can't always be done. I would add that as a recommendation.

CodingStyle.cc
```diff
@@ -0,0 +1,96 @@
...
+
+Q_LOGGING_CATEGORY(CodingStyleLog, "CodingStyleLog")
```
Susurrus: Might want to note that macro functions should be implemented such that they need a semi-colon at the end.

CodingStyle.cc
dogmaphobic: Yes, it's required just as well:

```c++
switch (foo) {
    case 1: {
        uint_16_t bar = 0;
        _call_private_method(bar);
        }
        break;
    default:
        _bollocks();
        break;
}
```

DonLakeFlyer:
@dogmaphobic Just adding Gus to make sure he is in on this conversation.

CodingStyle.h
DonLakeFlyer: Yup good point.

CodingStyle.cc
DonLakeFlyer: Will do

CodingStyle.h
DonLakeFlyer: Will do

CodingStyle.h
DonLakeFlyer: Macro constants make sense, but I don't the details associated with macro functions it think will get too involved too quickly.

CodingStyle.cc
DonLakeFlyer: Given our compiler warning settings always using default is not a good requirement. If you are using an enum as the switch the compiler will warn of missing cases. In these cases it is not known whether a default makes sense or not. You want to be able to catch the case where the switch fully exhausts the enum set with cases and then a new enum is added but the switch is not updated. As it currently stands this will fail compile. You don't want to always add a default for those cases. That later introduced bug tends to be harder to catch than the missing default one. Given the explanation of when to use and when to not is very long, I'd rather not have it in the coding style and catch it in code review.

DonLakeFlyer:
Some more explanation:
- My approach to the style document is to keep it at a high level, simple and not too long. I don't want to scare contributors away with too many requirements, or have people not pay attention because it's just too much information. The second step of code review can catch the more detailed stuff.
- I'm using the '///' version of doxygen commenting as opposed to the '/**'. Since it closer to most C++ style commenting as well as most of the commenting in the existing codebase. I'm also leaving out the optional '@brief' tag in most cases. So using that, the file is fully documented.

As far as where to put these documents, I figured alongside the code would be more useful. But I'm fine either way. Curious as to what other folks thing.

CodingStyle.cc
DonLakeFlyer: Yes to 3rd group. Not sure about alpabetize. I see very little of that in the existing code and I'd guess it would get out of sync fast since it will be non-obvious. Curious as to what other people have to say? I could go either way.

CodingStyle.cc
DonLakeFlyer: Given the compiler requires it do you think it worthwhile to add anyway? If not, would rather keep things shorter.

dogmaphobic:
> @dogmaphobic Just adding Gus to make sure he is in on this conversation.

:smile: Only the very egregious ones bother me (class member variables for instance :) As long as it follows some basic common sense and it's all consistent I'm fine. Talking about coding guideline is a bit like telling the NRA guns are bad.

I'm fine with all of it and I do like the idea of having a sample file right in the code base so I don't have to hunt down some wiki. The readme at the root should mention it and we're all happy.

DonLakeFlyer:
I'll clean this up tonight taking into account all the comments. Then we can re-visit.

CodingStyle.cc
Susurrus: Yes because the question is where do the curly brackets go and what's the indentation within them. You don't need to explain anything, just insert dogmaphobic's example verbatim into the code somewhere (assuming that's the bracket style we want).

DonLakeFlyer: Will do

Susurrus:
I'm onboard with adding it to the repo as long as we put it under the `/doc` folder.

CodingStyle.h
DonLakeFlyer: I'm matching the file naming convention here for this one #ifdef since I think it is more readable.

CodingStyle.h
DonLakeFlyer: Leaving these as ints since they are simpler and due to almost all Qt APIs using straight int you tend to use int a lot in the code to match that.

DonLakeFlyer:
Keeping the CodingStyle files at the upper level of the source tree since I think they will be more visible there.

DonLakeFlyer:
This should be good to go now. I'll merge and then update the wiki to reference it.

DonLakeFlyer:
I've updated a number of the wiki pages to provide better developer instruction as well as to link to these files.

CodingStyle.h
Susurrus: Then we have exceptions for macros. It'll be more consistent if we don't do this. It's also common in every other C/C++ project I've seen to capitalize the header guard.

DonLakeFlyer: If that's just too strange for folks I can switch to more normal.

CodingStyle.cc
```diff
@@ -0,0 +1,112 @@
+/*=====================================================================
+ 
+ QGroundControl Open Source Ground Control Station
+ 
+ (c) 2009 - 2015 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+
```
dogmaphobic: Just because we are discussing this, most files have some old date and a lot of them don't have this header at all.

DonLakeFlyer: I usually add the header if it is missing to a file I'm working in. Sometime's I'll update the copyright, but mostly I forget.

CodingStyle.h
Susurrus: I think it is, but I'm just one man. @LorenzMeier Care to chime in?

DonLakeFlyer: Already changed to CODINGSTYLE_H (which matches majority of existing code pattern) and pushed to master

Issue #1192, opened 2015-02-04
Incorrect Safety Config


DonLakeFlyer:
I noticed that you set NAV_DLL_N in the safety screen. This parameter only has a incluence if NAV_DLL_OBC = 1. I suggest configuring only COM_DL_EN and COM_DL_LOSS_T in this screen.

DonLakeFlyer:
Fixed

PR #1202, opened 2015-02-05
Reworking the layout of the "Return Home" Safety Dialog.

dogmaphobic:
Getting to know QLM.

I've reworked the layout of the Safety Dialog. I moved the **Triggers** up on top and the **Options** down below. I felt it followed a better order of priorities. I've also made the "Loiter Altitude" to be visible only if the loiter check box is checked. The idea being that of showing less is better. However, I'm open to change this to *disabled* instead of *invisible*. For that, I would also need to know what palette to use to gray it out...

Note that there seems to be a discrepancy with the font style used for the *CheckBox* and the regular *Label*.

screen shot 2015-02-05 at 2 46 26 pm


thomasgubler:
Relevant: https://github.com/mavlink/qgroundcontrol/issues/1192

dogmaphobic:
> Relevant: #1192

Yes, when either one of those (or both as shown below) are set, there are extra warnings that explain it. In this screen grab, I also have *Loiter* unchecked to show that you don't see the loiter altitude fields in this case.

screen shot 2015-02-05 at 3 10 41 pm

DonLakeFlyer:
With my work on Firmware setup I've started creating QGC* variants of the basic controls. For example QGCButton is a button which follows the styling and palette for QGC. It uses Button as it's base class. This will automatically adjust to theme palette changes as well as enabled/disabled state changes. Seems like a QGCText and QGCLabel which automatically respects the QGC palette and adds enabled/disabled color change as well would be useful as well? Then you could just use these and greatly reduce a lot of the QML you have to do yourself.

DonLakeFlyer:
Should have also said that this looks so much better already. Awesome!

DonLakeFlyer:
Another general comment on doing commits: While I'm working on something I will tend do all sorts of commits until I'm finished. Then I create a diff for my branch, create a new branch, apply the diff and then re-commit with a new set of commit(s) and comments, then I create a Pull from that new branch. This gets rid of the interim commit comments which may not make sense to anyone but myself.

dogmaphobic:
First, my git foo ain't that great yet :) Second, these were things I found during the conversation. You mean I should cancel this PR and start over with a new, cleaned up branch?

dogmaphobic:
For instance, by just giving the dialog above another look, I realized that the RC Transmitter Signal Lost should be the very first thing in the list. I would normally fix that and push the change to this branch...

DonLakeFlyer:
Nope. Just a comment for later. Fine to add things after the fact. If it gets too out of hand there are times where I will bail and re-submit. But this is fine.

dogmaphobic:
> Seems like a QGCText and QGCLabel which automatically respects the QGC palette and adds enabled/disabled color change as well would be useful as well?

Absolutely! :)

DonLakeFlyer:
I'll pull out the new controls into a separate pull then to get them to you faster.

dogmaphobic:
> I'll pull out the new controls into a separate pull then to get them to you faster.

Perfect!

Here is the newly ordered result:

screen shot 2015-02-05 at 4 12 22 pm

dogmaphobic:
On a somewhat related note, is there a place where you can set the units between metric and Imperial hogshead, furlong and dry peck?

DonLakeFlyer:
The units come directly from specification of the parameter mavlink xml which is then loaded into FactMetaData. At this point they are hardwired to the parameter spec. But if FactControls are used everywhere to hook to parameters then the FactSystem provides a central point to add that support and the UI QML would not need to change at all to make it work.

dogmaphobic:
> The units come directly from specification of the parameter mavlink xml which is then loaded into FactMetaData.

Hmmm... Do you know what else would be helpful? To be able to pull *longDescription* off of *Fact* and use it as a tooltip.

DonLakeFlyer:
Seems like tooltip would not be on for all controls though so it may not make sense to be a standard part of the base control. You can add it case by case as needed. For example in your Safety Config do you really want to tooltip all controls with the parameter descriptions? Seems like that may add more noise than be helpful. More of a power user thing?

DonLakeFlyer:
@thomasgubler Regarding #1992. Are you saying that the COM_DL_LOSS_EN checkbox should be connected to the COM_DL_LOSS_T value not the NAV_DLL_N value?

dogmaphobic:
The idea was to have an attribute. Something like *factTootipEnabled*, which could default to false and you just enabled it when needed. The point being, it would be a single line, just the attribute, next to the control you want it.

On the other hand, now that I look at the actual long descriptions within that XML file, they are not exactly user friendly :) So yes, indeed we might need to add something ourselves on a case by case basis.

DonLakeFlyer:
I'd rather not chew the memory to have the tooltip bound even if it is not displayed. Regarding long descriptions, these need to be cleaned up on the firmware side anyway, since my plan for the parameter editor is to use these. I've already put through a proposal there which will make the parameter metadata much more useful and trustworthy (https://github.com/PX4/Firmware/issues/1485). Just haven't gotten around to doing it yet. Also if longDescription really isn't appropriate we could add a specific tooltip value to the metadata as well.

DonLakeFlyer:
@thomasgubler Can you comment on the possible parameter issues? Or is that correct and good to go?

thomasgubler:
> @thomasgubler Regarding #1992. Are you saying that the COM_DL_LOSS_EN checkbox should be connected to the COM_DL_LOSS_T value not the NAV_DLL_N value?

yes, NAV_DLL_N is an advanced feature only activated with NAV_DLL_OBC

DonLakeFlyer:
Thanks Thomas. @dogmaphobic Looks like one last change and this should be good to go.

dogmaphobic:
> Looks like one last change and this should be good to go.

No worries. I've been buried in comm configuration for the past few days. Trying to bring order to chaos... :)

DonLakeFlyer:
@dogmaphobic Now that LinkManager is through can you make this one last change and then I'll merge?

dogmaphobic:
> Now that LinkManager is through can you make this one last change and then I'll merge?

Sure! Let me wrap up a few things here first and I will take a look at that.

dogmaphobic:
All right. I think this is all done. Let me know what you think...

DonLakeFlyer:
Waiting for TC pass then I'll merge.

Can you take a look and modifying your editor setting to not remove extra spaces on lines that you haven't changed? Or I wonder if that is a git setting? It's just generating clutter in the diffs.

dogmaphobic:
> Can you take a look and modifying your editor setting to not remove extra spaces on lines that you haven't changed? Or I wonder if that is a git setting? It's just generating clutter in the diffs.

I've already done that it should be ok for future commits. With that said, there ought to be a way to set this end to permanently ignore these changes. That would allow us to keep cleaning these issues without cluttering. Or, I could run a sed scripts in the entire source tree and make one jumbo commit to clean it all  white spaces in one go :)

DonLakeFlyer:
If you are using Xcode (like me) I don’t think there is any way to prevent these spaces from happening unless you turn off it’s auto-formatting option which is quite useful. It will only add the spaces to lines you changed and not modify places in existing files you didn’t touch. I’d rather not touch all files since the existing spaces are harmless. It’s not super easy to find a requirement that will work with all the editors in use.

dogmaphobic:
No, I don't use Xcode. If it's a lot of coding, I use Sublime Text. If it's just something quick, I use Qt Creator itself. No big deal. I will just leave it alone.

PR #1204, opened 2015-02-07
Qml controls

DonLakeFlyer:
New set of base Qml Controls which respect QGCPalette and theme changes:
- QGCLabel: Just like Label except it support enable/disable visuals.
- QGCButton: Just like Button. Still basic support at this time. No image or menu support.
- QGCRadioButton: Just like RadioButton
- QGCCheckBox: Just like CheckBox

These are available through:
    import QGroundControl.Controls 1.0

Also updated QGCPalette with better color set (final set will come from designers). You now need to import QGCPalette like this:
   import QGroundControl.Palette 1.0

Here is that the palette controls look like in both themes.
screen shot 2015-02-07 at 12 29 30 pm
screen shot 2015-02-07 at 12 29 43 pm
You can get to this window from the Tool Widgets menu in debug mode.

Work left to to:
- Convert FactControls to use these as base controls so palette is fully respect
- Unit Test which brings up Qml control test window
- Unit Test which clicks through all Setup Views


DonLakeFlyer:
@dogmaphobic Gus, once this goes through you'll have a label control in the correct palette that supports enable/disable.

PR #1205, opened 2015-02-08
Qt readme 5.4

jgoppert:
I didn't have any luck with Qt 5.3 on ubuntu 14.04, so I tried 5.4 and it works.I got the black box issue again on 5.3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,12 +65,12 @@ Supported builds for Linux are 32 or 64-bit, built using gcc.
 * For Fedora: `sudo yum install qt-creator qt5-qtbase-devel qt5-qtdeclarative-devel qt5-qtserialport-devel qt5-qtsvg-devel qt5-qtwebkit-devel SDL-devel SDL-static systemd-devel`
 * For Arch Linux: `pacman -Sy qtcreator qt5-base qt5-declarative qt5-serialport qt5-svg qt5-webkit`
 
-##### Install Qt5.3+ from PPA
-Note: Please be aware that the time of writing, Qt5.3 is unavailable in the official repositories Ubuntu 14.04/Mint 17.*. If it has become available since, please follow the instructions for installing Qt5.3 on Ubuntu.
-* Add this PPA to your sources.list: `ppa:beineri/opt-qt532-trusty`
-* Run the following in your terminal: `sudo apt-get update && sudo apt-get install qt53tools qt53base qt53declarative qt53serialport qt53svg qt53webkit`
-* Next, set the environment variables by executing in the terminal: `source /opt/qt53/bin/qt53-env.sh` or copy and paste the contents to your `~/.profile` file to set them on login.
-* Verify that the variables have been set: `echo $PATH && echo $QTDIR`. The output should read `/opt/qt53/bin:...` and `/opt/qt53`.
+##### Install Qt5.4 from PPA
+Note: Please be aware that the time of writing, Qt5.5 is unavailable in the official repositories Ubuntu 14.04/Mint 17.*. If it has become available since, please follow the instructions for installing Qt5.4 on Ubuntu.
+* Add this PPA to your sources.list: `ppa:beineri/opt-qt54-trusty`
+* Run the following in your terminal: `sudo apt-get update && sudo apt-get install qt54tools qt54base qt54declarative qt54serialport qt54svg qt54webkit qt54quickcontrols qt54graphicaleffects`
+* Next, set the environment variables by executing in the terminal: `source /opt/qt54/bin/qt54-env.sh` or copy and paste the contents to your `~/.profile` file to set them on login.
+* Verify that the variables have been set: `echo $PATH && echo $QTDIR`. The output should read `/opt/qt54/bin:...` and `/opt/qt54`.
 
 #### [Optional] Install additional libraries
 * For text-to-speech (espeak)


LorenzMeier:
@DonLakeFlyer Should we just require 5.4 on all platforms?

DonLakeFlyer:
This is off of latest master?

jgoppert:
Yes, off of master.

DonLakeFlyer:
Ok. Let me try myself on Ubuntu to see if I can repo. There was a specific fix to make this work. It may have broken with the last setup pull. I'll try moving to 5.4 as well to see what happens.

jgoppert:
qquickwidget was reporting an error about an empty screen graph.

DonLakeFlyer:
It happens when the main QML item has a size of 0, 0. The black screen was clicking on "Setup" or black screens after clicking a setup button like Firmare?

jgoppert:
black buttons on right menu of setup, mains screens worked correctly

DonLakeFlyer:
Can you attach a screen shot?

DonLakeFlyer:
Also if you are building yourself can you try modifying src/VehicleSetup/SetupViewButtons.qml to include:
    width: 100
    height: 600

Right after:
    id: topLevel

And see what happens?

jgoppert:
I uninstalled qt53 for now, so it would take me awhile to set up.

DonLakeFlyer:
Lets try this. In the shot below was the Firmware button invisible? How about the right hand side with instructions and Next button and so forth? Or both invisible?
screen shot 2015-02-07 at 5 34 38 pm

jgoppert:
I'm installing qt5.3 again to test it out. But the firmware button was invisible. The rest was visible, if I remember.

DonLakeFlyer:
Ok, that's what I thought. On Ubuntu 14.10 with Qt 5.3.0 it working fine. So not super helpful.

jgoppert:
After suggested fix of 

     width: 100
     height: 600

     Right after:
     id: topLevel

No change in behavior as described below:

Firmware button is fine, goes black after connection:

screenshot from 2015-02-07 20 47 10
screenshot from 2015-02-07 20 48 30

jgoppert:
I'm tempted to just require qt5.4 for ubuntu 14.04. It is just as easy to install the 5.4 ppa as it is to install the 5.2 ppa.

DonLakeFlyer:
Goes black after connection. Didn’t realize that. That’s different. Let me look a little. 5.4 may still be the right move anyway since it’s the latest release. If so, will want to move all OS forward likely.

dogmaphobic:
I can confirm the problem. I have both 5.3 and 5.4 installed with Ubuntu 14.04. With 5.3, you get black boxes where the Setup buttons (left pane) should be. It all works when built with 5.4. I noticed that yesterday but didn't give much thought until I saw this...

dogmaphobic:
Here is a screen grab showing it (Ubuntu 14.04, Qt 5.3):

screen shot 2015-02-08 at 9 55 45 am

Susurrus:
I'd like to avoid another Qt jump if we can. What version of 5.3 are people seeing the problem on? Because there has been a 5.3.1 bugfix release.

jgoppert:
I tried this on my laptop and I got the black screen again after params were transmitted and this error, using Qt5.4 this time, which is scarry:

UASParameterCommsMgrLog: emitting parameterListUpToDate
Default compId:  50
qrc:/qml/SensorsComponentSummary.qml:18: TypeError: Cannot read property 'value' of undefined
qrc:/qml/SensorsComponentSummary.qml:29: TypeError: Cannot read property 'value' of undefined
qrc:/qml/SensorsComponentSummary.qml:40: TypeError: Cannot read property 'value' of undefined
^C


It seems these new qml components are very unstable. Maybe we should consider sticking with normal Qt if they cause so many problems.

jgoppert:
Screenshot on my laptop on Qt 5.4 ubuntu 14.04 after parameter update.

screenshot from 2015-02-08 14 06 58

DonLakeFlyer:
It's just a sizing issue. There is a path where the size of the top level item is 0,0 and the resizing seems to work differently on different OS. You can fix it by manually sizing the item yourself even Qt is supposed to do it for you.

Susurrus:
From what @DonLakeFlyer has said, it sounds like this should not be merged and we should figure out a better way to deal with resizing the widgets.

jgoppert:
Agreed. We need to figure out all the quirks of the qml.

DonLakeFlyer:
I’ll be working on this tomorrow. My hope is to fix the black as well as target 1024x768 min resolution and get that all fixed up.

DonLakeFlyer:
Wait I didn’t see this down the thread. This is just a regular old bug in the qgc qml code not an os specific quirk.

jgoppert:
@DonLakeFlyer ah, good to know it isn't just my system.

DonLakeFlyer:
@jgoppert I just need to figure out the sequence which causes it. Looks like a possible connect/disconnect/connect bug. I tried something simple but it didn't happen. Can you give me exact steps to repro? Then I can fix and add unit test to make sure it doesn't happen again.

Susurrus:
From looking at the code quickly it looks like the `autopilot.parameters[]` array is not initialized properly.

DonLakeFlyer:
Yup, that's the symptom. But need to find the code path that causes it.

DonLakeFlyer:
Tracking whether we need this pull in Issue #1207

DonLakeFlyer:
This is replaced by Pull #1249.

PR #1206, opened 2015-02-08
Deleted image from setup buttons, to help with low res screens.

jgoppert:
It was impossible to select the parameters tab in the setup menu for reasonable resolutions (e.g. 1027x768). This fixes that by eliminating the gear picture that wasn't serving much purpose.


DonLakeFlyer:
The gear is a placeholder for button specific images coming soon from the ui designers. Give me a couple days to rethink the layout for a 1024x768 minimum resolution instead of removing it. I'll fix it up.

DonLakeFlyer:
This should be fixed by Pull #1216. So this change is no longer needed.

PR #1213, opened 2015-02-08
Fix setting the baud rate.

### LINKED ISSUE #1212: QGC cannot connect at 115200 baud
Been wasting the last week trying to connect to my board, but it turns out QGC won't connect at any baud rate besides the default 57600. This also applies to the databits and stopbits settings.

Susurrus:
There's also a partial fix for the databits and stopbits. I'm unsure how to deal with the logic
around the 'is_cdc' variable, as I can't really follow what it's for.

Closes #1212

### DIFF
--- a/src/comm/SerialLink.cc
+++ b/src/comm/SerialLink.cc
@@ -770,17 +770,8 @@ bool SerialLink::setBaudRateType(int rateIndex)
     return result;
 }
 
-bool SerialLink::setBaudRateString(const QString& rate)
-{
-    bool ok;
-    int intrate = rate.toInt(&ok);
-    if (!ok) return false;
-    return setBaudRate(intrate);
-}
-
 bool SerialLink::setBaudRate(int rate)
 {
-
     bool accepted = false;
     if (rate != m_baud) {
         m_baud = rate;
@@ -835,37 +826,6 @@ bool SerialLink::setParityType(int parity)
                 }
             emit updateLink(this);
         }
-    }
-    return accepted;
-}
-
-
-bool SerialLink::setDataBits(int dataBits)
-{
-
-    qDebug("SET DATA BITS");
-    bool accepted = false;
-    if (dataBits != m_dataBits) {
-        m_dataBits = static_cast<QSerialPort::DataBits>(dataBits);
-        accepted = true;
-        if (m_port && !m_is_cdc)
-            accepted = m_port->setDataBits(static_cast<QSerialPort::DataBits>(dataBits));
-        emit updateLink(this);
-    }
-    return accepted;
-}
-
-bool SerialLink::setStopBits(int stopBits)
-{
-
-    // Note 3 is OneAndAHalf stopbits.
-    bool accepted = false;
-    if (stopBits != m_stopBits) {
-        m_stopBits = static_cast<QSerialPort::StopBits>(stopBits);
-        accepted = true;
-        if (m_port && !m_is_cdc)
-            accepted = m_port->setStopBits(static_cast<QSerialPort::StopBits>(stopBits));
-        emit updateLink(this);
     }
     return accepted;
 }
--- a/src/comm/SerialLink.h
+++ b/src/comm/SerialLink.h
@@ -123,11 +123,6 @@ signals: //[TODO] Refactor to Linkinterface
 public slots:
     bool setPortName(QString portName);
     bool setBaudRate(int rate);
-    bool setDataBits(int dataBits);
-    bool setStopBits(int stopBits);
-
-    // Set string rate
-    bool setBaudRateString(const QString& rate);
 
     // Set ENUM values
     bool setBaudRateType(int rateIndex);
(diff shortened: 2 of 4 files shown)


Susurrus:
@LorenzMeier @DonLakeFlyer This breaks things pretty hard for non-57600 modem use, which I think is a lot of systems besides the PX4. Can you guys help me out here?

LorenzMeier:
Sure! Good to merge? @DonLakeFlyer do you just merge if you're happy?

DonLakeFlyer:
Haven't had time to look at it yet. Give me a couple minutes.

Susurrus:
Well, the problem is that this doesn't work for the stop bits and data bits from my testing and I don't know why. It has something to do with the `is_cdc` member variable, but without documentation I haven't really figured out how that functionality is supposed to work. Looks like you put it in @LorenzMeier so I was hoping you could help out with fixing that code path.

LorenzMeier:
I don't think I put that in, and I would guess it can be removed. It probably tries to imply that setting the baud rate is pointless for true CDC devices, and/or it might have been a workaround for something else. If you remove it, I'd like to cross-check with Pixhawk, but otherwise I would assume it can go.

DonLakeFlyer:
Right. It's checking for px4 in the device description in which case it assumes it's a cdc device and in turn doesn't really set baud rate, stop bits etc. It just stores the values and makes believe they were set. 

@Susurrus This cdc thing shouldn't be affecting your code path since your device should not have px4 in the description. Are you hitting this path? Also what was the original bug? I don't see from this pull what was previously wrong?

Susurrus:
What is a CDC device first of all?

@DonLakeFlyer The problem was that the UI for the baud rate didn't actually trigger changing the settings in the SerialLink object. This is because the signals/slots are incorrect. You can't correct a SerialLinkInterface object to slots in a subclass, so they were failing at runtime and never being triggered. I fixed that by moving the slot to the same class as the UI and having that call into the proper SerialLinkInterface functions.

So this fixes the baud rate. But in the commit you'll see I tried to do the same fix for the data and stop bits, but it doesn't work. I stepped through the code until I ran into the SerialLink code that checks its `is_cdc` member and uses defaults if it detects it. I got lost about here so I decided to turn to you two for help.

LorenzMeier:
http://en.wikipedia.org/wiki/USB_communications_device_class

Most likely the CDC check doesn't account for your corner case. We could try to disable the special case handling and see what happens in that case.

Susurrus:
It sets it to false unless it's a PX4 device. So it's false for me. The thing is I don't see why the CDC code should even be kicking in, this is before I've even finished configuring the SerialLink! At this point there is no link, so it couldn't detect if it's a CDC device or not.

LorenzMeier:
Is the CDC init value maybe just wrong and set to true?

Susurrus:
I think it is true by default.

DonLakeFlyer:
It's called in a number of places: constructor, setportname, .... If the port name is available it will do it's thing. If not it ends up a no-op.

It defaults to true but as soon as you name the port it will get reset to the right value.

DonLakeFlyer:
The fact that it hits the cdc case for stop bits shouldn't matter should it? If you are not connected it just saves the values and uses it at hardware connect time. At hardware connect time cdc should be set correctly since port name is set correctly at that time and will make the right setDataBits calls. Did you find that setDataBits was not being set in the hardwareConnect call?

Susurrus:
I watched the console debugging output which says what connections are used for the SerialLink object. When I hit connect it never changed the stop or data bits.

DonLakeFlyer:
It won't specifically out the set data bits thing. If it is working correctly you should have gotten:
        qDebug() << "Configuring port";
At connect time.

Susurrus:
It looks like removing the CDC logic as done in #1215 fixed the data/stop bits problem. So this PR appears to do its job so I'd like to merge it. And I'd like to propose pushing this PR and cleaning up the stop/databits in #1215.

DonLakeFlyer:
Need to talk with Lorenz on removing the retry stuff.

Susurrus:
Can we merge this in the meantime? This will at least fix the baudrate for now and is technically correct.

DonLakeFlyer:
@dogmaphobic Up to Gus since he has already root-tilled this code. I would probably just make sure he has your connect fixes in his pull.

dogmaphobic:
> Up to Gus since he has already root-tilled this code. I would probably just make sure he has your connect fixes in his pull.

I'm fine given that it will be a couple of days before I can have something properly cleaned up and documented to push. I can manually deal with the conflicts after this goes through.

Susurrus:
Since this passes and it fixes this issue, I'd like to merge it then. @LorenzMeier @DonLakeFlyer Just want one final confirmation as things are pretty crazy with the comms code right now.

LorenzMeier:
I think we can merge it in. I will try a bit with it merged and see how it does.

DonLakeFlyer:
Good to go. Likely blow up with a merge conflict with my other pull but I'll deal with that.

Issue #1214, opened 2015-02-09
Disconnecting a UAS from a live link triggers log file save dialog


Susurrus:
I think this has to do with the disconnect/reconnect behavior, as I wouldn't imagine this is intentional. If you get to the limit of your radio range you might go in and out of reception. Having this **modal** dialog pop up several times would definitely limit QGC's usefulness then.

DonLakeFlyer:
You didn't disconnect using QGC right? You just broke the connection from a hardware standpoint? If so, the broken link code needs to be modified to look different than a user-driven disconnect/reconnect. Should be pretty straightforward. Dialog needs to be modal so it doesn't lose the system state out from under it. For example a re-connect would cause loss of the temp log file.

Susurrus:
Why doesn't the temp log system just save temp logs everytime it connects using an incrementing style. Then when QGC is launched again it can clean any extra ones up. This would be more complicated, but I'm very wary of modal dialogs in a system that's supposed to control robotic systems.

DonLakeFlyer:
I thought there was code which automatically disconnected/reconnected when it lost a heartbeat for some length. There doesn't seem to be. The only way the link would be disconnected would be if the user specifically disconnected the link. If you just lose communication over the link when you go out of range for example the actual link will stay connected as far as I can see. The toolbar might say connection lost but that's just because it lost heartbeat. What's the problem here? Or is there some other repro?

Susurrus:
Oh there definitely is. I've been debugging stuff and it's the most annoying thing ever because I have to check whether QGC is connected or not. It continually connects/disconnects and never gives up.

But I don't understand what you're asking about.

DonLakeFlyer:
Ok. I need to dig more to find it. The repro case would be to connect QGC over radios and then pull the radio on the non-QGC side?

Susurrus:
Yeah, easiest is to stop the autopilot transmitting: hold it in reset, power it off, disconnect the UART, etc.

DonLakeFlyer:
Found it. It's a "feature" of the SeriaLink code only. Will need to distinguish between user disconnected and SerialLink code trying to do it's thing.

Susurrus:
Well, should this be something that's pushed up into the `LinkInterface` code, or possibly the SerialInterface code? Though I don't know what problem this code is trying to solve. Does disconnecting and reconnecting ever fix anything?

DonLakeFlyer:
[...]

PR #1215, opened 2015-02-09
Remove link error retry logic and cdc checking

DonLakeFlyer:
- Added retry logic to port open to handle case where Pixhawk is past bootloader but still waiting for USB driver to start

This fixes part of Issue #1214 assuming it is the right thing to do. Tested cdc removal on Pixhawk with USB connection at various baud rates and it still worked fine.

### DIFF
--- a/src/comm/SerialLink.cc
+++ b/src/comm/SerialLink.cc
@@ -23,7 +23,6 @@ SerialLink::SerialLink(QString portname, int baudRate, bool hardwareFlowControl,
     m_bytesRead(0),
     m_port(Q_NULLPTR),
     type(""),
-    m_is_cdc(true),
     m_stopp(false),
     m_reqReset(false)
 {
@@ -38,8 +37,6 @@ SerialLink::SerialLink(QString portname, int baudRate, bool hardwareFlowControl,
     {
         m_portName = ports.first().trimmed();
     }
-
-    checkIfCDC();
 
     // Set unique ID and add link to the list of links
     m_id = getNextLinkId();
@@ -139,7 +136,6 @@ void SerialLink::loadSettings()
     if (settings.contains("SERIALLINK_COMM_PORT"))
     {
         m_portName = settings.value("SERIALLINK_COMM_PORT").toString();
-        checkIfCDC();
 
         m_baud = settings.value("SERIALLINK_COMM_BAUD").toInt();
         m_parity = settings.value("SERIALLINK_COMM_PARITY").toInt();
@@ -161,46 +157,12 @@ void SerialLink::writeSettings()
     settings.setValue("SERIALLINK_COMM_FLOW_CONTROL", getFlowType());
 }
 
-void SerialLink::checkIfCDC()
-{
-    QString description = "X";
-    foreach (QSerialPortInfo info,QSerialPortInfo::availablePorts())
-    {
-        if (m_portName == info.portName())
-        {
-            description = info.description();
-            break;
-        }
-    }
-    if (description.toLower().contains("mega") && description.contains("2560"))
-    {
-        type = "apm";
-        m_is_cdc = false;
-        qDebug() << "Attempting connection to an APM, with description:" << description;
-    }
-    else if (description.toLower().contains("px4"))
-    {
-        type = "px4";
-        m_is_cdc = true;
-        qDebug() << "Attempting connection to a PX4 unit with description:" << description;
-    }
-    else
-    {
-        type = "other";
-        m_is_cdc = false;
-        qDebug() << "Attempting connection to something unknown with description:" << description;
-    }
-}
-
-
 /**
  * @brief Runs the thread
  *
  **/
 void SerialLink::run()
 {
-    checkIfCDC();
-
     // Initialize the connection
     if (!hardwareConnect(type)) {
         //Need to error out here.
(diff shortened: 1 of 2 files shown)


Susurrus:
Yep, just had a deadline yesterday, but now that that's done I'd like to resolve this.

So I gave it a spin, but it does have a downside where making a connection and then disconnecting within a few seconds brings down all of QGC. Looks like it has to do with all the `msleep()` calls. So 2 questions: does it make sense to do this for 6 seconds, does it really take that long for the bootloader. And finally, I think it makes more sense to only sleep for .1s at a time in order to leave the UI much more responsive.

Also the code you added could really use more comments. I don't want someone else to have to come back in 6 months and ask these same questions all over again. This should also include making the fixed sleep timeout a constant static member of SerialLink.cc. Same with the number of retries. Magic numbers make me wary.

Susurrus:
What do you think of making there be 20 x .1s retries for the bootloader and 10 * .1s retries for opening the port?

Susurrus:
But even with those settings the problem persists. If you try to disconnect a link you made before it gives you the MAVLink error/baud rate error the UI just freezes until the MAVLink error triggers. And then you get 2 MAVLink error popups and the thing disconnects properly. This is far from ideal behavior.

Susurrus:
On the plus side this appears to fix #1212 when applied on top of #1213.

DonLakeFlyer:
"But even with those settings the problem persists. If you try to disconnect a link you made before it gives you the MAVLink error/baud rate error the UI just freezes until the MAVLink error triggers. And then you get 2 MAVLink error popups and the thing disconnects properly. This is far from ideal behavior."

Are you sure the has anything to do with this pull and wasn't already a problem in master?

dogmaphobic:
I've been working on this for the past few days as part of my reorg of the comm links. I already have all this working in my work branch...

DonLakeFlyer:
@dogmaphobic You've changed the cdc stuff and the connect/disconnect retry logic embedded in the SerialLink code? Didn't think your stuff was related.

dogmaphobic:
Yes and no. I've changed how the configuration is stored (port, baud rate, flow control, etc.) and as a result, a lot of that source file changed. I removed a whole lot of unused cruft added over time that was not being used and causing a lot of confusion. The main issue being that configuration was not be respected and overwritten (the problem of not being able to change baud rate for instance). By cleaning that up and loading a sane set of parameters associated with an individual connection, all these problems went away. In other words, once I get my end done, a lot of this will be replaced. I would rather wait before pushing this in as it will cause merge issues for me.

I'm almost at a point where you could try the branch yourself. Serial Link is complete but I need to work on the other types.

DonLakeFlyer:
Take a look at this pull and see if it overlaps with your stuff. If yours is almost ready we can wait and @Susurrus can work from private builds temporarily. Also FYI the concept of removing the auto-link connect/disconnect is not yet approved. I need to have a conversation with Lorenz on that.

Point me to your branch when you are ready and I can start playing around with it.

dogmaphobic:
> Also FYI the concept of removing the auto-link connect/disconnect is not yet approved. I need to have a conversation with Lorenz on that.

I didn't remove it. I just made it work :) I'll let you know when you can test my branch. It should be testable shortly.

As to the changes in this pull, the main overlap is around the fact all those set/get parameters are now in a different interface. With that said, the change is not so great that I can't manually fix once I merge.

DonLakeFlyer:
This pull can wait till after you are done and then we can figure out if it is still needed.

DonLakeFlyer:
@LorenzMeier Are you ok with removing the link disconnect/re-connect retry logic in serial link? See Issue 1214 for more comments as well.

LorenzMeier:
Yes, remove it if that helps. We'll see then what the user experience with the bootloader is like and what to do to fix it.

DonLakeFlyer:
This pull has a change to get past the bootloader as well as the USB driver startup delays. I think it actually seems to work better in more cases than the previous code. I'll fix up the merge conflict, pass TC again and then merge in.

Susurrus:
This code works for me except in the case where all the MAVLink error logic kicks in, so it's not a problem here. I still would prefer that all the magic numbers get set as static consts member variables and that there be a few more comments explaining what the new code is doing and why it's doing it. Not everyone's familiar with the PX4 startup/connection process.

DonLakeFlyer:
@Susurrus Can you explain the MAVLink issue again. I'm not quite getting it. Want to make sure I don't screw something up.

Susurrus:
To reproduce:
 1) Open a serial port with the wrong settings.
 2) Re-open the serial port connections window
 3) Click Disconnect

Now what will happen is the UI will freeze. This freeze ends when 2 dialogs pop up both describing problems with the MAVLink data stream. But the freeze lasts a while, I'm thinking it has something to do with the msleeps that have been added (one is 6s long). Once the dialogs pop up the UI is then usable again but you have 2 dialog windows to deal with.

src/comm/SerialLink.cc
```diff
@@ -449,6 +383,7 @@ bool SerialLink::hardwareConnect(QString &type)
 
         for (retries = 0; retries < retry_limit; retries++) {
             if (!isBootloader()) {
+                QGC::SLEEP::msleep(500);
```
DonLakeFlyer: I need to remove this. I left it in by mistake while I was testing things.

DonLakeFlyer:
If you are getting errors from mavlink then the port is open and has nothing to do with these changes. The only time you could hit a six second delay would be if it found a bootloader and the bootloader was hung for 6 seconds. Then it would fail port open which would be correct. If you have no bootloader the code is a no-op.

The rebase on this just isn't going to work. I'm going to close this out and submit a new one.

src/comm/SerialLink.cc
```diff
@@ -475,11 +411,17 @@ bool SerialLink::hardwareConnect(QString &type)
...
     //    port->setCommTimeouts(QSerialPort::CtScheme_NonBlockingRead);
 
-    if (!m_port->open(QIODevice::ReadWrite)) {
+    // After the bootloader times out, it still can take a second or so for the USB driver to come up and make
+    // the port available for open. So we retry a few times to wait for it.
+    for (int openRetries=0; openRetries<4; openRetries++) {
+        if (!m_port->open(QIODevice::ReadWrite)) {
+            qDebug() << "Port open failed, retrying";
+            QGC::SLEEP::msleep(500);
+        }
```
dogmaphobic: I know this PR is closed but because it got me confused, I figured it wouldn't hurt to point these out.
This loop doesn't get out when the connection succeeds. It just keeps trying to connect again...

DonLakeFlyer: Yup. Thanks Gus. I'm going to put through a new PR for all of this once your stuff gets through.

dogmaphobic:
I think this is all fixed in my linkManager branch. Isn't it?

DonLakeFlyer:
Don't know. I'll have to take a look at a Pull so I can see a diff.

PR #1216, opened 2015-02-09
Rework setup buttons to fit in lower resolution displays

DonLakeFlyer:


DonLakeFlyer:
@jgoppert Can you give this a try?

PR #1221, opened 2015-02-10
Switch to new-style sensor calibration

LorenzMeier:
@DonLakeFlyer This is the change required on the parameter side. We could as well keep the old code in parallel and switch based on the existence of the outdated param. No other changes are needed.

### DIFF
--- a/src/ui/px4_configuration/QGCPX4SensorCalibration.cc
+++ b/src/ui/px4_configuration/QGCPX4SensorCalibration.cc
@@ -142,10 +142,10 @@ void QGCPX4SensorCalibration::parameterChanged(int uas, int component, QString p
     }
 
     // Check mag calibration naively
-    if (parameterName.contains("SENS_MAG_XOFF")) {
-        float offset = value.toFloat();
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Ok. I'll use this as a starting point. Bunch of other stuff needs changing as well.

DonLakeFlyer:
You can have 3 MAGS but there is still only one SENS_EXT_MAG_ROT? Should there be one each of this as well?

LorenzMeier:
Yes there should. I didn't get around to handle that though yet.

LorenzMeier:
Screw it, I just hit this on another thing. I'll do it now.

DonLakeFlyer:
What will the param names be so I can update QGC ahead of time.

LorenzMeier:
I will keep the param name, so it'll continue to work. I'll phase in an alternate name after when its done. Probably CAL_MAG0_ROT

DonLakeFlyer:
Subsumed by Pull #1223

Issue #1208, opened 2015-02-08
Calibration indicators not turning green


jgoppert:
The color of the mag/gyro/accel calibration indicators now don't change to green after calibration on master.

DonLakeFlyer:
Looks like in latest px4fmu firmware master the parameters have changed for calibration. This is also causing the problems with the black screens you are getting because parameters are missing that are expected.

@LorenzMeier Did something change with parameters associated with calibration?

DonLakeFlyer:
Looks like they may have changed to CAL_* parameters.

LorenzMeier:
Hi Don,

I didn't realize that the black screens can be associated to that. Yes, the param names had to be adjusted once to the CAL_ set.

DonLakeFlyer:
@LorenzMeier Should the calibration code just be changed the CAL_...1_ version of these new params?

LorenzMeier:
I have a PR somewhere. Let me dig.

DonLakeFlyer:
I can fix it. I'm going to need to change the VehicleComponent code as well to implement an interim parameter mapping capability. So it's going to get pretty involved to support both Stable and master since they use different parameter sets.

LorenzMeier:
Here you go.
https://github.com/mavlink/qgroundcontrol/pull/1221

LorenzMeier:
Its not too involved actually. If you can query if a parameter exists in the set its pretty simple. Since QGC only uses them to check if something has been set up, its not very involved in their actual values.

DonLakeFlyer:
It will end up being easier to leave to code connected to the old params for now and re-route in the param manager. Long term it will need to get fancier due to the FactSystem stuff.

LorenzMeier:
Fixed in #1223.

PR #1223, opened 2015-02-10
Support two param sets for calibration

DonLakeFlyer:
The parameters for calibration were changed in the latest master. QGC
needs to work with both. Hacked in for now, waiting on a true parameter
re-mapping implementation.

This fixed Issue #1208.


DonLakeFlyer:
@LorenzMeier Can you double check this?

LorenzMeier:
I will test this tomorrow early morning.

DonLakeFlyer:
Still needs additional fix for re-requesting parameter set after cal. Working on it.

Issue #1227, opened 2015-02-12
Lastest master broken by latest mavlink module update


DonLakeFlyer:
Unit tests will fail as well Setup View. Working through with Lorenz.

DonLakeFlyer:
Fixed

PR #1228, opened 2015-02-12
Fix SetupView to work with latest mavlink headers

DonLakeFlyer:
- Updated MainWindowTest to catch this problem
- Also implemented initial pass at SetupView unit test

This is a temp fix for Issue #1227


LorenzMeier:
Sorry about the merge fail - I put in a new PR which densifies that enum again.

DonLakeFlyer:
This is the real fix now, not temp.

PR #1230, opened 2015-02-12
Fix small logic error with NULL serial port.

dogmaphobic:
I had fixed this in my branch but when running today with master, it crashed on me again (it happens for me when switching USB connections between VMs). It's a small fix but if the port doesn't create, it crashes when moving threads...

P.S. It seems my editor is cleaning up blank spaces too and git is taking those as changes as well...

### DIFF
--- a/src/comm/SerialLink.cc
+++ b/src/comm/SerialLink.cc
@@ -329,7 +329,7 @@ void SerialLink::run()
         }
         QGC::SLEEP::msleep(SerialLink::poll_interval);
     } // end of forever
-    
+
     if (m_port) {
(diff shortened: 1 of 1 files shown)


src/comm/SerialLink.cc
```diff
@@ -463,12 +463,11 @@ bool SerialLink::hardwareConnect(QString &type)
     }
 
     m_port = new QSerialPort(m_portName);
-    m_port->moveToThread(this);
-
     if (!m_port) {
```
dogmaphobic: this *m_port->moveToThread(this);* was being called before the test for a valid pointer was made.

PR #1232, opened 2015-02-13
Convert Firmware Upgrade to QML

DonLakeFlyer:
- Much simpler UI, select firmware, click upgrade your done
- MVC style qml programming model
- Since upgrade still tends to be flaky brought log back

screen shot 2015-02-12 at 3 57 32 pm

Will now work with UI designer to pretty up a bit.


DonLakeFlyer:
Note this does not yet support "rebooting" plugged in board. Still need to unlug. That will come in another pull.

DonLakeFlyer:
Should have also said the the guts of the upgrade code are the same as they were before. No changes there. I did move the files to VehicleSetup though. So really just ui changes, internals are the same.

No progress bar in Qml yet. Just found out that progress bar didn't show up till 5.4! We'll get that once the conversion to 5.4 is complete. Day or so...

PR #1233, opened 2015-02-13
Fix Eigen includes

DonLakeFlyer:
Using "..." on header includes not only is nasty but can end up doing different things on different compiler versions. This fixes a problem on Fedora.

### DIFF
--- a/src/uas/UASManager.h
+++ b/src/uas/UASManager.h
@@ -32,10 +32,13 @@ This file is part of the QGROUNDCONTROL project
 #define _UASMANAGER_H_
 
 #include "UASManagerInterface.h"
+#include "UASInterface.h"
+
 #include <QList>
 #include <QMutex>
-#include <UASInterface.h>
-#include "../../libs/eigen/Eigen/Eigen"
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@Susurrus Can you try building this on Fedora. Just generic build. Should fix the problem.

Susurrus:
I'm building right now, but this looks sane to me.

Just to note that searching for other instances of this (`grep "\"\.\." --exclude=*.ui src -r`) shows the problem in other files:
```
src/ui/map/MAV2DIcon.h:#include "../../libs/opmapcontrol/opmapcontrol.h"
src/ui/map/QGCMapWidget.h:#include "../../../libs/opmapcontrol/opmapcontrol.h"
src/ui/map3D/Pixhawk3DWidget.cc:#include "../MainWindow.h"
src/ui/XbeeConfigurationWindow.h:#include "../comm/HexSpinBox.h"
```

DonLakeFlyer:
Can you open an issue on that and assign to me?

PR #1234, opened 2015-02-13
Link Configuration Management

dogmaphobic:
The ultimate goal is to plug a device, set it up and go fly it (or whatever it does) without any fuss.

The goal of this first installment is to organize the code a bit so it's more readable, clean up a bit of left over cruft, and manage link configurations (and links in general).

* Added a concept of Link Configuration where a link configuration (of a given certain type) is created and maintained through the new Manage Links preferences dialog.
* Newly discovered (Serial) PX4 devices are automatically created and configured (if not already configured).
* Link types that now use link configuration are created through the link factory in LinkManager. Unfortunately, 3DR Radio Modems cannot be easily identified as they use run-of-the-mill FTDI UART/USB chips and look no different than any other device using these chips. Eventually I want to *listen* to these unidentified serial ports and look for something speaking MAVLink.
* Links are not mandated to use link configuration. You can still create link classes in the old fashion and handle them as usual. This is the case for link types I have yet to migrate, such as TCPLink, XBeeLink, etc.
* Link configuration are saved and restored through QSettings.
* Fixed several issues with the Serial Link. In addition, now all settings are manipulated through the link configuration instance, instead of directly interacting with the link instance. With that said, the Serial Link still needs quite a bit of TLC.
* I made the call to remove ancient baud rates from the serial configuration. I just assumed this would be pretty useless on a 50 baud connection. HF packet radio link?
* With UPDLink, you can now maintain the target hosts through its settings' UI as these have been moved to the link configuration. Its QGCUDPLinkConfiguration settings dialog has been re-written from scratch.
* Added a MockConfiguration to the MockLink class. As the name implies, it’s a mock with no settings other than the name (base class settings).
* Added code to properly save and restore last used connection. This is used for when the auto connect flag is set and it will now load and connect a real configuration, instead of some random Serial port.
* Remove the serial port “Baud Rate” drop down box from the main tool bar. It no longer makes sense.
* Replaced the Link drop down box (the one that only had serial ports and *UDP*) with a Link Configuration drop down box. You select the configuration you want to connect and hit the connect button. You can also invoke the *Manage Links* from it. Note that when you first run QGC, it will automatically create and populate this with any PX4 it finds.
* Added a method to the settings dialog, which allows to invoke it and go straight into one of its panes. 
* Added a size spacer to QGCMessageBox to enlarge the message dialogs a bit under Mac OS. By default, they were too narrow and there was no way to change the width.
* Minor coding style guideline fixes wherever I touched.
* Forcing no proxy for UDP connections. This came through google groups and as the change is right where I'd been working, I added it here to avoid collision.
* I noticed that some times, the disconnect signal would get to the updateConnect button (tool bar) before the link was physically disconnected. That would cause it to count it as a still connected link and get the connect button logic out of sync.

Note that this supersedes PR #1147, which should be discarded if this is accepted.


dogmaphobic:
The failure is only while running unit test on Linux. Both Mac and Windows passed.
```
[23:23:20] :	 [Step 11/17] QDEBUG : FactSystemTestGeneric::qmlUpdate_test() SENT COMMAND 245
[23:23:20] :	 [Step 11/17] QDEBUG : FactSystemTestGeneric::qmlUpdate_test() MockLink: Unhandled mavlink message, id: 76
[23:23:21] :	 [Step 11/17] QWARN  : FactSystemTestGeneric::qmlUpdate_test() QThread: Destroyed while thread is still running
[23:23:21]W:	 [Step 11/17] Aborted (core dumped)
[23:23:21]W:	 [Step 11/17] Process exited with code 134
```

Crap... Was this me? Running --unittest in the debugger I get a crash in a different place. It's within the new operator in MAVLinkProtocol::_createSingleton():

```
QDEBUG : FactSystemTestGeneric::parameter_test() Deleting UAS object:  "MAV 128"
PASS   : FactSystemTestGeneric::parameter_test()
*** Error in `<longpath>Debug/debug/qgroundcontrol': corrupted double-linked list: 0x00000000018c7a10 ***
```

DonLakeFlyer:
The singleton startup and takedown is really touchy. I fixed it up to be way better than before but there are likely still holes in there. There a few intermittent issue I've seen but these look to be new. They may not specifically be in your code. But you may have moved around creation/destruction sequence which are causing an existing bug to show up where it didn't before.

Susurrus:
So, this as a massive commit is really hard to review by sight. Would it be possible to get this patch broken up to make this easier to review? I know this is a pain in the ass, but considering the size of this patch and how it working is critical to QGC, I think it'd be helpful to have more eyes on it.

It'd be nice to at least have the main stuff in one commit and the unit testing stuff in another. And to separate out the whitespace stuff from real changes. This is really easy to do with `git add -p` where you can select the specific parts of the file to commit.

DonLakeFlyer:
@dogmaphobic You're probably going to have to figure out some way to turn off the whitespace thing in your editor. Ok for this pull, but it just generates too many changes. I'll take a look through the code (and re-run) over the weekend.

Susurrus:
I don't think the whitespace thing is a problem, just leave them as separate patches. It's really easy using `git add -p` as I mentioned. We really should be cleaning up the whitespace as we go anyways.

dogmaphobic:
> Would it be possible to get this patch broken up to make this easier to review?

It would help me a lot if you gave me specific details and examples on how to go about it. As it is, it feels I'm spending a great deal of time getting things here as opposed to using my time to contribute to the actual code.

Susurrus:
That's the problem with a good development process, it's slow, but it generally makes more mistakes. Watch any of the big open source projects (Mesa, Linux kernel, etc.) and you'll see they are very picky about their patches. Part of "contributing to the actual code" is letting us vet your code and make sure it's sound. Writing code isn't the hard part in contributing to open source projects, it's making sure it's good code, which requires a lot of review. Really your goal in writing patches isn't to reduce the amount of time you spent making the code changes, it's to reduce the amount of time reviewers need to spend vetting it. 

> It would help me a lot if you gave me specific details and examples on how to go about it

[This](http://gitready.com/advanced/2009/01/15/piecemeal-staging.html) kind of goes over the basics. After this it's really just how you want to organize your commits. Big commits are **really** bad because they make it hard to review. Sometimes this can't be avoided, but usually it can. For example, with your code it would have been good to have at least 3 commits:
 * One that does all the LinkConfiguration changes
 * One that updates the unit tests
 * One that handles whitespace.

Just that alone would have make things simpler. And this would be easy to do after-the-fact by choosing exactly what you want to commit using `git add -p`, as the link above demonstrates. Given that you already have a commit, [this process](https://stackoverflow.com/questions/1526044/partly-cherry-picking-a-commit-with-git) is what you want to follow.

But the ideal process is to never break building master, because that makes tracking down regressions difficult when going back in history, so each of your changes should be self-contained. This is a little harder to do and I don't worry about it too much as, like you've said, it's too much work for a project this small (but note that this is a requirement for patches to large projects, their large changes end up being 20+ commits to help reviewing them).

dogmaphobic:
I wouldn’t argue the need for code review. I actually agree with all you’ve said. Separating blank space changes from code changes makes all the sense in the world. I’m just getting a hard time dealing with git, which I find it to be quite destructive if you don’t know what you are doing (most of the time for me). That’s what I meant about getting stuff here. I can certainly appreciate git for what it is but it has a very steep learning curve when you go beyond the mundane.

Also note that I didn’t go anywhere near unit testing within this commit :)

dogmaphobic:
Ha! I learned a new trick. Add a `?w=1` to the end of the URL and it will ignore white space changes:

https://github.com/mavlink/qgroundcontrol/pull/1234/files?w=1

DonLakeFlyer:
Awesome, that will make things easier.

dogmaphobic:
I've been trying to track down this Linux issue with the unit testing. I have not even attempted to understand what that is doing but the crash has something to do with parameter testing. What I noticed was that after pulling this branch, when I did a `git status`, this showed up:

```modified:   libs/mavlink/include/mavlink/v1.0 (new commits)```

After I did a `git submodule update` and rebuilt it, the unit tests pass consistently.

Is this something I missed in my commit? Why is it only an issue on Linux?

**Comment by dogmaphobic** [comment]
While at it, on an unrelated note, I noticed that during the tests, the MainWindow is created 3 times and several of the objects it creates are never deleted. Of notice is `MAVLinkDecoder`, which keeps running. At the end of the test, there are 3 instances running.

**Comment by DonLakeFlyer** [comment]
Seems like MAVLinkDecoder should be a QGCSingleton?

**Comment by dogmaphobic** [comment]
I found a few issues with the connect/disconnect signals from LinkManager, which I plan on working next. In the mean time, I’ve pushed an important fix in this PR as this could end up as a crash were it to happen.

**Comment by Susurrus** [comment]
@DonLakeFlyer I'm pretty certain the MavlinkDecoder should be a singleton. Only problem there is that I think the datastream is decoded twice and 2 different decoders are used. Could be thinking of an old version of the code however.

**Comment by DonLakeFlyer** [comment]
Bummer the ?w=1 thing shows you a better view but then you can't seem to comment on things any longer.

File: src/QGCMessageBox.h [hunk 1]
```diff
@@ -115,6 +115,10 @@ private slots:
             QMessageBox box(icon, emptyTitle, title, buttons, parent);
             box.setDefaultButton(defaultButton);
             box.setInformativeText(text);
+            // Get this thing off a proper size
+            QSpacerItem* horizontalSpacer = new QSpacerItem(500, 0, QSizePolicy::Minimum, QSizePolicy::Expanding);
+            QGridLayout* layout = (QGridLayout*)box.layout();
+            layout->addItem(horizontalSpacer, layout->rowCount(), 0, 1, layout->columnCount());
```
DonLakeFlyer: Curious as to what this does?

src/comm/LinkInterface.h
```diff
@@ -59,27 +60,33 @@ class LinkInterface : public QThread
...
 
+    /**
+     * @brief Destructor
+     * LinkManager take ownership of Links once they are added to it. Once added to LinkManager
+     * use LinkManager::deleteLink to remove if necessary.
+     **/
     virtual ~LinkInterface() {
-        // LinkManager take ownership of Links once they are added to it. Once added to LinkManager
-        // user LinkManager::deleteLink to remove if necessary/
         Q_ASSERT(!_ownedByLinkManager || _deletedByLinkManager);
```
DonLakeFlyer: Is the comment above no longer true or different now? Reason I put it there is because the assert (and the whole link ownership thing) is a bit strange and wanted to provide context if someone hit it.

DonLakeFlyer:
Do any of your changes invalidate the current saved settings format? If so the settings version number will need to be bumped. Also just FYI: At a minimum TCP and XBee will need to be added back soon after this goes through since I think both of those are still actively used.

Next I'm going to do a little more testing. Then after that I think we can call it good.

src/QGCMessageBox.h
dogmaphobic: QMessageBox sizing, particularly the width, under Mac OS is awful. It's always too narrow and breaks the text down to many lines. What I did above was to force to be at least 500 pixels wide.

src/comm/LinkInterface.h
dogmaphobic: I have not done anything yet to LinkManager pertaining to link management. That's next in my todo list. In summary, you cannot ever *delete* a link as it breaks MAVLinkProtocol. The links are just disconnected but remain loaded and inactive. In other words, forced memory leak as they are for ever lost and unused. Once I get that re-visited, I will deal with this who-owns-what and how to handle object deletion.

dogmaphobic:
> Do any of your changes invalidate the current saved settings format? If so the settings version number will need to be bumped.

Technically yes. Link configuration is now stored in a different place (different keys). I'm also now saving the last used connection to be restored if the *auto connect* flag is set. This is a new setting. With that said, the old settings are simply ignored and unused. It won't affect the new code.

> At a minimum TCP and XBee will need to be added back soon after this goes through since I think both of those are still actively used.

Those are also next in my todo list. I actually have quite a few XBees around and wanted to test them for telemetry, something I can bring over to Mac OS, which currently lacks support for it.

DonLakeFlyer:
No need to bump settings rev. Been playing around with it this morning. So nice to have you connection settings saved. Awesome. merging now.

src/comm/LinkInterface.h
DonLakeFlyer: Due to the cross threading issue I would guess that the link object is going to need to be reference counted in order to get disconnect/delete to work correctly.

Issue #1207, opened 2015-02-08
Black setup window on QGC


TSC21:
Hi there,

I'm trying to download the latest firmware to Pixhawk using the QGC GUI, but strangely I can't get the "Setup" window to show anything:
![screenshot from 2015-02-08 14 31 39](https://cloud.githubusercontent.com/assets/5048656/6096891/695d598c-af9f-11e4-88c8-05caefaa21ab.png)
What can be the problem? One thing I notice is that when I change to the Setup window, I get this in the terminal:
```bash
QQuickWidget: render scenegraph with no context
```
May this be Qt related?

Thanks in advance!

Note:
QGroundControl v2.3.0
Ubuntu 14.04.1 LTS
Qt v5.3.2

TSC21:
Update:

I tried to clean all (which included clone the repo again and compile).
Now, pressing the setup button gives me the following:
```bash
QObject::connect: Cannot connect (null)::summaryButtonClicked() to SetupView::_summaryButtonClicked()
Segmentation fault (core dumped)
```
Any idea?

DonLakeFlyer:
If you switch to Qt 5.4 as show in this Pull: https://github.com/mavlink/qgroundcontrol/pull/1205 <https://github.com/mavlink/qgroundcontrol/pull/1205>.  It will all work.

TSC21:
Yes solved the problem. For the record:
1. add `ppa:beineri/opt-qt54-trusty` to source list:
```bash
sudo add-apt-repository ppa:beineri/opt-qt54-trusty
2. install `qt54` and its apps (p.e.: `sudo apt-get update && sudo apt-get install qt54tools qt54base qt54declarative qt54serialport qt54svg qt54webkit qt54quickcontrols qt54xmlpatterns qt54x11extras qt54websockets qt54sensors qt54script qt54quick1 qt54qbs qt54multimedia qt54location qt54imageformats qt54graphicaleffects qt54creator qt54connectivity qt54canvas3d`)
3. add `source /opt/qt54/bin/qt54-env.sh` to `~/.profile`.

Probably better to add this to https://github.com/mavlink/qgroundcontrol readme, right?

DonLakeFlyer:
I"m re-opening this because there is still an issue on Qt5.3 with a black setup button list after connect. I can finally repro myself and will try to fix on 5.3 before we just move to 5.4.

DonLakeFlyer:
[...]

PR #1242, opened 2015-02-17
Repaired overwriting current files for flight data log (fix #1241)

### LINKED ISSUE #1241: Saving flight data log: When choosing existing file, file is not overwritten
When choosing an existing file as destination to save the flight data log, a warning appears, that the existing file will be overwritten. However, even if I click on 'yes' , the file is not overwritten and the log not saved.

Tested on Ubuntu 14.04.01 LTS

basil-huber:
changed QGCApplication::saveTempFlightDataLogOnMainThread(QString tempLogfile) to enable overwriting files
existing file is removed before saving file (QFile::copy cannot overwrite files)

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -498,12 +498,20 @@ void QGCApplication::saveTempFlightDataLogOnMainThread(QString tempLogfile)
             if(QFile::exists(saveFilename) && !QFile::remove(saveFilename)){
                 // if the file cannot be removed, prompt user and ask new path
                 saveError = true;
+<<<<<<< Updated upstream
                 QMessageBox::warning (MainWindow::instance(), "File Error","Could not overwrite existing file.\nPlease provide a different file name to save to.");
                 QFile::copy(tempLogfile, saveFilename);
             } else if(!QFile::copy(tempLogfile, saveFilename)) {
                 // if file could not be copied, prompt user and ask new path
                 saveError = true;
                 QMessageBox::warning (MainWindow::instance(), "File Error","Could not create file.\nPlease provide a different file name to save to.");
+=======
+                QGCApplication::criticalMessageBoxOnMainThread("File Error","Could not overwrite existing file.\nPlease provide a different file name to save to.");
+            } else if(!QFile::copy(tempLogfile, saveFilename)) {
+                // if file could not be copied, prompt user and ask new path
+                saveError = true;
+                QGCApplication::criticalMessageBoxOnMainThread("File Error","Could not create file.\nPlease provide a different file name to save to.");
+>>>>>>> Stashed changes
             }
         }
     } while(saveError); // if the file could not be overwritten, ask for new file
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Thanks again Basil.

src/QGCApplication.cc
```diff
@@ -483,15 +483,28 @@ void QGCApplication::criticalMessageBoxOnMainThread(const QString& title, const
...
+            qgcApp()->mavlinkLogFilesLocation(),
+            tr("Flight Data Log Files (*.mavlink)"),
+            "mavlink");
+    
+        if (!saveFilename.isEmpty()) {
+            // if file exsits already, try to remove it first to overwrite it
+            if(QFile::exists(saveFilename) && !QFile::remove(saveFilename)){
+                saveError = true;
+                QMessageBox::warning (MainWindow::instance(), "File Error","Could not overwrite existing file");
+                QFile::copy(tempLogfile, saveFilename);
```
DonLakeFlyer: Not sure why this is here. Isn't it just going to fail?

src/QGCApplication.cc
```diff
@@ -483,15 +483,28 @@ void QGCApplication::criticalMessageBoxOnMainThread(const QString& title, const
...
+            } else if(!QFile::copy(tempLogfile, saveFilename)) {
+                saveError = true;
+                QMessageBox::warning (MainWindow::instance(), "File Error","Could not create file");
```
DonLakeFlyer: Can you change both error strings to include ". Please provide a different file name to save to." at the end?

dogmaphobic: Also, use `QGCMessageBox` instead when you can so these message boxes are standardized.

src/QGCApplication.cc
basil-huber: That was a relict from copy/pasting stuff around. I remove it.  Well spotted.

DonLakeFlyer:
Just FYI to folks: This afternoon (PST time) I'm going to be taking down TeamCity to move it to Qt 5.4. I'll try to get the latest set of pulls through first.

Susurrus:
Wait, why is QGC moving to Qt5.4 then?

DonLakeFlyer:
Issue #1207

src/QGCApplication.cc
```diff
@@ -483,15 +483,29 @@ void QGCApplication::criticalMessageBoxOnMainThread(const QString& title, const
...
+            qgcApp()->mavlinkLogFilesLocation(),
+            tr("Flight Data Log Files (*.mavlink)"),
+            "mavlink");
+    
+        if (!saveFilename.isEmpty()) {
+            // if file exsits already, try to remove it first to overwrite it
+            if(QFile::exists(saveFilename) && !QFile::remove(saveFilename)){
+                // if the file cannot be removed, prompt user and ask new path
+                saveError = true;
+                QGCApplication::criticalMessageBoxOnMainThread("File Error","Could not overwrite existing file.\nPlease provide a different file name to save to.");
```
DonLakeFlyer: You're already on the main thread so you just need QGCApplication::warning. Warning instead of critical. Critical is more for you're about to crash type messages.

DonLakeFlyer:
Thanks again Basil. Sorry for all the hassle with something that seems so simple!

basil-huber:
No problem. It's great that someone takes care that the code stays clean.

Issue #1237, opened 2015-02-14
System messages to users are not visible in a meaningful fashion


LorenzMeier:
Both PX4 and APM report status via MAVLink text messages to the user. The implementation is here:
https://github.com/mavlink/qgroundcontrol/blob/master/src/ui/uas/QGCMessageView.cc

It is displayed on Fly -> Info View -> Messages tab:
![schnappschuss 2015-02-14 21 40 46](https://cloud.githubusercontent.com/assets/1208119/6200967/28d240e2-b492-11e4-9ab9-244784ce8fd4.png)

These messages are really important in particular if the system configuration is off. It would be great to have this reworked - I wouldn't advise any modal dialogs, but very visible notifications e.g. in the status bar showing the last message and allowing to expand to see the full log.

Susurrus:
They are shown in the top statusbar. What more were you thinking?

PR #1245, opened 2015-02-17
Message Handler

dogmaphobic:
In response to #1237, I've moved the handling of messages to a new, singleton class responsible for receiving, collecting and handling these messages. The existing **Info View** Widget, which contains a **Messages** tab, now collects messages from this message handler.

In addition, and the reason for all this, I've added a mouse "hover" event on the message label in the tool bar. Once the mouse enters the label area, a drop down list shows up showing the list of messages. Once the mouse moves away from the list, it automatically closes.

This is an attempt to find a best solution to display to the users these messages. Now that the mechanics are in place, we can present these messages in any way you want.

screen shot 2015-02-17 at 12 44 23 pm

### DIFF
--- a/src/ui/QGCToolBar.cc
+++ b/src/ui/QGCToolBar.cc
@@ -31,6 +31,26 @@ This file is part of the QGROUNDCONTROL project
 #include "UASManager.h"
 #include "MainWindow.h"
 #include "QGCApplication.h"
+#include "QGCMessageView.h"
+
+
+// Label class that sends mouse hover events
+class QCGHoverLabel : public QLabel
+{
+public:
+    QCGHoverLabel(QGCToolBar* toolBar) : QLabel(toolBar)
+    {
+        _toolBar = toolBar;
+    }
+protected:
+    void enterEvent(QEvent* event)
+    {
+        Q_UNUSED(event);
+        _toolBar->enterMessageLabel();
+    }
+private:
+    QGCToolBar* _toolBar;
+};
 
 QGCToolBar::QGCToolBar(QWidget *parent) :
     QToolBar(parent),
@@ -47,6 +67,7 @@ QGCToolBar::QGCToolBar(QWidget *parent) :
     _linkMgr(LinkManager::instance()),
     _linkCombo(NULL),
     _linkComboAction(NULL),
+    _rollDownMessages(NULL),
     _linksConnected(false),
     _linkSelected(false)
 {
@@ -148,7 +169,7 @@ void QGCToolBar::createUI()
     toolBarWpLabel->setAlignment(Qt::AlignCenter);
     toolBarWpAction = addWidget(toolBarWpLabel);
 
-    toolBarMessageLabel = new QLabel(this);
+    toolBarMessageLabel = new QCGHoverLabel(this);
     toolBarMessageLabel->setToolTip(tr("Most recent system message"));
     toolBarMessageLabel->setObjectName("toolBarMessageLabel");
     toolBarMessageAction = addWidget(toolBarMessageLabel);
@@ -413,7 +434,7 @@ void QGCToolBar::updateView()
         if (QGC::groundTimeMilliseconds() - lastSystemMessageTimeMs < 15000) {
             toolBarMessageLabel->setText(QString("%1").arg(lastSystemMessage));
         } else {
-            toolBarMessageLabel->clear();
+            toolBarMessageLabel->setText(tr("Messages"));
         }
     }
 
@@ -677,15 +698,6 @@ void QGCToolBar::_disconnectFromMenu(bool checked)
     _linkMgr->disconnectLink(link);
 }
 
-void QGCToolBar::clearStatusString()
-{
-    if (toolBarMessageLabel->text().length() > 0)
-    {
-        lastSystemMessage = "";
-        changed = true;
-    }
-}
-
 void QGCToolBar::_linkComboActivated(int index)
 {
     int type = _linkCombo->itemData(index).toInt();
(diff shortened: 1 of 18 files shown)


src/ui/QGCToolBar.cc
```diff
@@ -721,3 +733,27 @@ void QGCToolBar::_updateConfigurations()
...
+ */
+void QGCToolBar::enterMessageLabel()
+{
+    if(!_rollDownMessages)
+    {
+        QPoint p = toolBarMessageLabel->mapToGlobal(QPoint(0,0));
+        _rollDownMessages = new QGCMessageViewRollDown(MainWindow::instance(),this);
+        _rollDownMessages->setAttribute(Qt::WA_DeleteOnClose);
+        _rollDownMessages->move(mapFromGlobal(p));
+        _rollDownMessages->setMinimumSize(400,300);
```
Susurrus: Can we not set a minimum size and instead set it to autoexpand to display it's contents? I'm not certain if that's possible, but it'd be nice to not cover up so much UI when trying to read messages.

DonLakeFlyer:
FYI on class naming scheme I've been using. I only prepend QGC if the class overrides an existing Qt class. Otherwise the QGC prefix just ends up being noise. 

Also ui might look better if the background of the dropdown was slightly transparent as well as @Susurrus' comments.

LorenzMeier:
Awesome! I think this will save a lot of drama. Since you progress so fast, would it be possible to tease you to a follow-up PR with a "missed messages" icon in the toolbar?

dogmaphobic:
> FYI on class naming scheme I've been using. I only prepend QGC if the class overrides an existing Qt class. Otherwise the QGC prefix just ends up being noise.

Agree as it makes total sense. That would be the case for `QGCMessageHandler` and `QGCUasMessage`, which I've just created. The rest however, was already QGCxxx to begin with. Something to add to the guidelines.

Regarding sizing, transparency and so on, keep these coming and I can combine them all in one commit.

dogmaphobic:
> follow-up PR with a "missed messages" icon in the toolbar?

It's sort of already there. Before my changes, the last message would be displayed for a few seconds and then the text would be cleared. I now, instead of clearing the text, replace it with "Messages". The idea was just a place holder for at some point, replace it with an icon or something. If you hover the mouse over the "Messages" label, the message list is shown.

screen shot 2015-02-17 at 1 03 16 pm

DonLakeFlyer:
Yes, the old code is inconsistent on QGC* naming scheme. I've just kind of developed that scheme over time.

I'm going to try to get the coding style stuff up tonight. Bogged down with great weather which means I'm spending most of my time flying. (Sorry, US East Coasters!)

dogmaphobic:
> Sorry, US East Coasters!

Sure, thanks. That helps 8-P

Snowing and the high my reach 20 today, -2 tomorrow. Balmy...

dogmaphobic:
The sizing of the dialog is a bit of an issue. The embedded text widget does not change its size dynamically based on its content. I would have to iterate the strings, find the longest one, map that size to the context and size the dialog accordingly. Doable, but not now.

For the time being, I changed the opacity a bit so you can see some of the background.

screen shot 2015-02-17 at 3 46 37 pm

DonLakeFlyer:
Sounds good. We can do a rework when we do a full toolbar pass.

src/ui/uas/UASMessageView.cc
```diff
@@ -0,0 +1,159 @@
...
+
+UASMessageViewRollDown::UASMessageViewRollDown(QWidget *parent, QGCToolBar *toolBar)
+    : UASMessageView(parent)
+{
+    _toolBar = toolBar;
+    setAttribute(Qt::WA_TranslucentBackground);
+    setStyleSheet("background-color: rgba(0%,0%,0%,80%); border: 2px;");
+    QPlainTextEdit *msgWidget = ui()->plainTextEdit;
+    QAction* clearAction = new QAction(tr("Clear Text"), this);
+    connect(clearAction, SIGNAL(triggered()), msgWidget, SLOT(clear()));
```
DonLakeFlyer: Is there no way to use the direct class reference way of connect'ing so you get a compile failure on the SIGNAL or SLOT missing? Sometimes there isn't without all sorts of add casting. Same thing in other connect calls.

dogmaphobic: I just copied and pasted what was there to begin with. I can revisit these though I'm not 100% sure what you mean. That QAction adds a context menu, which currently has one single option "Clear" the message list. I left it alone for the time being. Eventually, I wanted to add a way to save the contents out to a file but figured copying the contents and pasting out to a new file would do fine for now.

dogmaphobic: Speaking of which, it doesn't even work! As I said, I had just copied and pasted. Now that I actually tried the "Clear Text", it has no effect. Fixing it...

DonLakeFlyer: Got it. FYI for now, or try to fix if you want: I've been trying to converting as many of these over as I hit them. There are two ways to connect:

Method 1: connect(signalObject, SIGNAL(...), slotObject, SLOT(...);
Using this syntax things will still compile if SIGNAL or SLOT are missing or incompatible. Instead you'll get a runtime warning.

Method 2: connect(signalObject, &SignalObjectClass::signalMethod, slotObject, &SlotObjectClass::slotMethod);
Using this syntax the compile will fail if either signal or slot are incorrect.

What tends to happen is that later on, someone changes the method for a signal or a slot without fixing all the method 1 type connect calls. This leads to errors that only show up at runtime. If you use method 2 this will automatically break at compile time. It is not always possible to use method 2. For example some signals have multiple overriden methods. It gets nasty in those cases.

DonLakeFlyer: Probably because the signal connection is broken. Look in your debug output. I always run with QT_FATAL_WARNINGS=1 in my config. This turns all qWarnings into asserts which will stop the run.

DonLakeFlyer:
@dogmaphobic Let me know when this is good to go an I'll merge. I've finished looking through the code.

dogmaphobic:
Sure, please go ahead and merge it. We can continue later on if needed. I'm kind of stuck waiting for it any way.

Issue #1247, opened 2015-02-17
UASView slow to update


Susurrus:
I've recently implemented the throttle output for my vehicle with the VFR_HUD message. It's really useful to know the throttle value when underway. But the UASView is very slow to update, so it gives a much slower-than-realtime status, looks to be about every 3s. This is the same regardless of the low-power mode.

Would people be amenable to moving this to a 1s update rate? That'd give much better feedback to the user.

Susurrus:
Looking at the code it looks like the update rate should be .8s normally and be 3x that during low-power mode. So I'm not certain how 3s is occurring in the non-powersave mode.

Susurrus:
Digging through this a bit more, it looks like the update logic has gotten quite complex because it's trying to account for the following:
  * Animating a color fade for the heartbeat messages
  * Flashing red at 5Hz when the system is disconnected
  * Responding to Qt widget events, for a reason I don't understand.

The reason I say it's currently too complex is because it doesn't seem to update the whole UI for a UAS at the 800ms intervals, instead updating the color fade quickly but not otherwise updating things like the uptime timer and other stuff.

My proposal here is to remove the color fade completely as well as the other widget event timer setting (which doesn't seem to do anything anyways). We'd keep the 5Hz flashing that is per-UAS but ditch the rest to keep the logic simple.

So, do people like the heartbeat fade is I guess the pending question for this change.

Susurrus:
Or maybe it's just worth changing it so that the throttle value is updated everytime a VFR_HUD message is received but otherwise leaving the logic as it stands.

LorenzMeier:
The issue is that a lot of other messages arrive at 5-10 Hz, making a widget redraw unacceptable at that rate.

Susurrus:
That's fair. Since I won't have time to fix this and it's a feature request, closing for now.

PR #1248, opened 2015-02-17
Uasview throttle

Susurrus:
Fix for #1247. I opted for updating the throttle value whenever a new thrust value is received.

Note that the code I removed didn't seem to do anything. I don't see a different mouse cursor when hovering over an inactive UAS.

Also added comments where appropriate.

### DIFF
--- a/src/ui/uas/UASView.cc
+++ b/src/ui/uas/UASView.cc
@@ -229,49 +229,6 @@ void UASView::mouseDoubleClickEvent (QMouseEvent * event)
 {
     Q_UNUSED(event);
     UASManager::instance()->setActiveUAS(uas);
-    // qDebug() << __FILE__ << __LINE__ << "DOUBLECLICKED";
-}
-
-void UASView::enterEvent(QEvent* event)
-{
-    if (event->type() == QEvent::MouseMove)
-    {
-        emit uasInFocus(uas);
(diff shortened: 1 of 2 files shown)


LorenzMeier:
The throttle shouldn't be just drawn, it should be rate limited. It makes no sense to re-render the widget at 10 Hz, and there is nothing in the code that would prevent that unhealthy pattern. I fully agree that less than 3 seconds response time is desirable, but the widget should control the rate, not the incoming messages.

Re-rendering based on data rate makes us very susceptible to slight changes in telemetry rate, and would mean that QGC sucks on systems connected via Wifi (and we'll see most drones using Wifi or Bluetooth or 4G in the near future) and drain batteries of laptops and tablets in no time.

We've been there before, and we should not go back.

Susurrus:
I just wanted to get something started here. So how can I proceed with this? The rate update of this widget is complicated and needlessly complex. Is there a way to simplify it somehow? We can continue this discussion in #1247.

Also, I think the first commit of mine is fine. Are you okay with me cherry-picking that commit into master and then closing this PR?

LorenzMeier:
Cherry picked and closing - I still think your proposed change makes sense, we just need to work out how to continue to rate-limit the UI.

PR #1249, opened 2015-02-17
Qt5.4 upgrade

DonLakeFlyer:
This moves (and validates) base Qt version required to 5.4. Fixes Issue #1207.


DonLakeFlyer:
First pass of this through TC is expected to fail. I'm validating Qt version check on all OS. Then this afternoon I'll add support to TC for 5.4 builds.

DonLakeFlyer:
QGC google group has been notified of change. If any insurmountable technical issues that come up there and I'll hold. If not, I'll move forward with the full switch in a day or two.

DonLakeFlyer:
Working through a Qt5.4 issue that Bill Bonney is seeing with APM Planner on OSX. So far I'm not seeing it though in QGC. Still looking.

DonLakeFlyer:
I managed to screw this pull up. I'm closing and re-creating.

PR #1254, opened 2015-02-18
Adding TCP to the Link Configuration Manager

dogmaphobic:
Adding TCP to the list of connection types handled by the link configuration manager.
This push also fixes an issue where changing a name of an UDP link was not working.
Updating copyright dates as I go through these files.


DonLakeFlyer:
I don't have anything I can connect to via TCP so I just did a quick review through the code.

Also FYI: If you are feeling adventurous the majority of the SerialLink implementation is bogus as far as how it uses a threaded run loop. QSerialPort is already threaded and supports a signal based programming model which would make the SerialLink code much much simpler and similar to how TCPLink is implemented using signals instead of a threaded run loop.

dogmaphobic:
> I don't have anything I can connect to via TCP so I just did a quick review through the code.

Me either. I had to rely on the (modified) unit tests.

Regarding the threading in SerialLink, coincidentally I was just thinking about it when I saw how the XBee link is implemented (XBee is another can of worms, which I will leave for later--too much work of dubious value).

I've been trying to make sense of how links are created, connected, disconnected and never deleted. Trying to come up with a clean way to *clean* these dangling, unused links.

DonLakeFlyer:
Ok. If you worked through unit test then I'll merge. Best we can do.  Did you turn the unit test back on? I think I turned it off a while back because of intermittent failures.

dogmaphobic:
> Did you turn the unit test back on? I think I turned it off a while back because of intermittent failures.

No, I didn't! I simply *fixed* what I needed and *assumed* it was running. It wasn't until now that you mentioned it that I noticed it's disabled. I run it as it is and it obviously all passed with flying colors. No wonder :)

dogmaphobic:
Enabling Unit Test and checking how it is behaving...

dogmaphobic:
Hmmm... I enabled it and found a crash. It's crashing because `TCPLinkUnitTest::init()` is being called twice in a row (the `TCPLinkUnitTest` constructor is only called once). I won't be able to look further into it until later though.

DonLakeFlyer:
If it's just a UT crash I'd consider just leaving it off. It's been off for a while so it has likely degraded even further. There is already an Issue in to fix it and turn it back on.

dogmaphobic:
> If it's just a UT crash I'd consider just leaving it off.

All right. In that case, it's all there and done.

PR #1256, opened 2015-02-18
Qt 5.4 Upgrade

DonLakeFlyer:
Replacement for Pull #1249.

Waiting to hear back from Lorenz on one last sanity check. I'll let the Google Group email notice sit for today. Then if Lorenz comes back positive and no technical issues from Google Group still start the cutover tomorrow.

### DIFF
--- a/QGCInstaller.pri
+++ b/QGCInstaller.pri
@@ -19,10 +19,16 @@
 
 installer {
     MacBuild {
-        QMAKE_POST_LINK += && $$dirname(QMAKE_QMAKE)/macdeployqt $${DESTDIR}/qgroundcontrol.app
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
TC is switched over. Doing test tuns now.

DonLakeFlyer:
Only thing not working now is Windows Release build due to problem with speech api compile error. @LorenzMeier Did you test the new build I posted on OS X? I should be ready to bump to Qt 5.4 soon. Then I'll merge, kick off daily builds and test those.

DonLakeFlyer:
Re-ran through TC just to be safe. I'm going to merge now and tonight I'll test installers.

PR #1258, opened 2015-02-20
Cleaning dangling links

dogmaphobic:
Links were being disconnected but never deleted (and never re-used). These links were just accumulating and bloating resources. I did not have much success trying to simply disconnecting and immediately deleting them because it takes a thread cycle for the link disconnect signal to reach the MAVLinkProtocol instance, which keeps talking to the link. It would be a boatload of work to change that signaling and much more prone to errors.

Instead, I've implemented a simple delayed link delete mechanism to clean up these dangling, unused links.

### DIFF
--- a/src/comm/LinkInterface.h
+++ b/src/comm/LinkInterface.h
@@ -57,7 +57,8 @@ public:
     LinkInterface() :
         QThread(0),
         _ownedByLinkManager(false),
-        _deletedByLinkManager(false)
+        _deletedByLinkManager(false),
+        _flaggedForDeletion(false)
     {
         // Initialize everything for the data rate calculation buffers.
         inDataIndex  = 0;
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Seems like reference counting would be a better way to handle this. I think QSharedPointer does that but I've never used that specific class. If you reference count the link, then delete in LinkManager and then also when it gets the signal in mavlink the actual delete will happen at the right time. The timer thing seems a bit error prone.

DonLakeFlyer:
TC Windows Release build failure is being caused by a Windows Update to the windows sdks. I have a workaround for it in my Qt5.4 pull which I'll push through today. So until that goes through this will fail TC.

dogmaphobic:
I'm not too thrilled with that timer either but it does what it is supposed to do and it works. I made it so it is wired within the Link Manager and not accessible from the outside, which minimizes misuse. Reference count is the proper way to go but it is also a much longer term proposition as it implies a much broader architecture change.

DonLakeFlyer:
Sounds good. Once I get 5.4 switched over in TC I re-run this through and merge once it passes.

DonLakeFlyer:
Resubmitted to TC.

PR #1262, opened 2015-02-24
Added progress bar support back

DonLakeFlyer:


DonLakeFlyer:
To firmware upgrade...

Issue #1255, opened 2015-02-18
Deleting a UAS in the UASView after disconnecting crashes QGC


Susurrus:
Maybe remove the right-click "Delete" option should be removed? I know there has been some action to not allow removing UASes.

Setting this to important since it's a crasher, at least on x64 Linux.

DonLakeFlyer:
This is more cross-thread object deleted signaling problems.

DonLakeFlyer:
Fixed

PR #1266, opened 2015-02-25
Fix object de-reference crash on deletion

DonLakeFlyer:
Fix for Issue #1255

### DIFF
--- a/src/ui/QGCWaypointListMulti.cc
+++ b/src/ui/QGCWaypointListMulti.cc
@@ -2,65 +2,63 @@
 #include "ui_QGCWaypointListMulti.h"
 #include "UASManager.h"
 
+void* QGCWaypointListMulti::_offlineUAS = NULL;
+
 QGCWaypointListMulti::QGCWaypointListMulti(QWidget *parent) :
     QWidget(parent),
-    offline_uas_id(0),
-    ui(new Ui::QGCWaypointListMulti)
+    _ui(new Ui::QGCWaypointListMulti)
 {
-    ui->setupUi(this);
+    _ui->setupUi(this);
     setMinimumSize(600, 80);
-    connect(UASManager::instance(), &UASManager::UASCreated, this, &QGCWaypointListMulti::systemCreated);
-    connect(UASManager::instance(), &UASManager::activeUASSet, this, &QGCWaypointListMulti::systemSetActive);
+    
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
@Susurrus Can you give this a try?

Susurrus:
I'll start testing it now. In the future can you leave your style changes for a different commit, makes it really hard to see the real changes to the code. And that makes it easier to make sure the style changes are purely cosmetic.

Susurrus:
Doesn't crash anymore, so that's fixed. But two follow-up issues:
  * After deleting the UAS the Unmanned Systems widget now shows the "Connect Link" button, which doesn't do anything when pressed.
  * Closing still pops up the "active connections to vehicles" dialog.

DonLakeFlyer:
Can you enter an Issue assigned to me on the Connect thing. I need to pull that out of there and replace it with some ui to say no current UAS or something like that.

For the other problem, here is what I tried:
- Boot
- Connect to Pixhawk over USB
- Switch to Plan where I have the widget
- Disconnect
- Got save flight log dialog, cancel
- Right click Delete UAS
- Quit app
- No connection dialog, just shuts down

Do you have different steps to repro this?

Susurrus:
I don't disconnect in QGC, I just stop the serial stream on the UAS end:

  1. Boot
  2. Connect to UAS
  2. Switch to Plan where I have the widget
  2. Stop the serial stream on the UAS end
  2. Right click Delete UAS
  2. Quit QGC
  2. Got save flight log dialog, cancel
  2. App quits

DonLakeFlyer:
Delete is incorrectly made available when the heartbeat times out. It should only be there when the UAS is disconnected.

Susurrus:
So is fixing the remove action not going to be part of this PR? I'm firing up my test station to check this out.

DonLakeFlyer:
No it won't. That ends up being way more involved. It's related the the rest of the UI which falss on it's face when it tries to work against a disconnected UAS. I removed the delete action and added Issue #1275 to fix it and put it back at some point.

Also FYI: With the new serial port code step #4 will no longer disconnect the link which is why you will still get #7.

Susurrus:
So is deleting a UAS is not supported at all anymore? If that's the case, then I imagine this commit is r+ by me.

DonLakeFlyer:
Nope. I think if you delete the link after you disconnect it may go away, but can't remember.

DonLakeFlyer:
Thanks @Susurrus

PR #1271, opened 2015-02-25
Firmware Upgrade ui re-work, plus QGCComboBox work

DonLakeFlyer:
- QGCComboBox now follows QGCPalette and ui design
- Firmware Upgrade uses combo box for firmware selection
- Firmware Upgrade warns with message box if you select beta or developer firmware versions

screen shot 2015-02-25 at 12 15 54 pm


DonLakeFlyer:
Also made Upgrade button a Primary button.

PR #1281, opened 2015-02-28
Sensor config in Qml

DonLakeFlyer:
- First iteration of Sensor Config in Qml. Fully functional, just not pretty. Works correctly with latest firmware parameter set (Stable and Master). UI is text log based only. Nice visuals come later. Includes support for multiple internal/external mag rotation settings, as well as correct fixed wing or non-fixed win calibration differences.
- Autopilot plugin detects older version firmware parameter set. Warns user that QGC doesn't support setup on this old firmware. Suggests firmware upgrade. Setup components not displayed in this case.
- Added new Qml FactComboBox control in QGroundControl.FactControls
- Added new Qml IndicatorButton control in QGroundControl.Controls which is a push button with the little red/green indicator support

screen shot 2015-02-27 at 3 33 25 pm


LorenzMeier:
@DonLakeFlyer This is the fix you needed: https://github.com/PX4/Firmware/pull/1850. I can't test tonight, but maybe you want to give it a quick go.

DonLakeFlyer:
I'll give it a try. Probably tomorrow my time.

PR #1290, opened 2015-03-01
Radio Cal no longer selects flight mode switches

DonLakeFlyer:
You have to do that from Flight Modes config. Changing in prep for new Flight Modes implementation.


DonLakeFlyer:
This pull does not check for radio cal needed correctly. I'll update that in a separate pull.

PR #1293, opened 2015-03-01
Update Radio Cal trigger

DonLakeFlyer:
Using new logic to signal need for radio cal:
- Needed if any attitude control mapping is missing
- Needed if min, max and trim for still set to defaults for channel 1-4


DonLakeFlyer:
- Removed mode switch mapping from radio component summary
- Added Flaps, Aux1, Aux2 mapping to radio component summary

PR #1299, opened 2015-03-02
Flight Mode config trial test

DonLakeFlyer:
@LorenzMeier Can you give this a try and comment?

I spent a bunch of time thinking about ui for flight mode config given the various switch mappings that are possible:
rc_mode_switch_flow

Tried to do it with dropdowns but it gets complicated really fast due to previously mapped channels and so forth. Then thought maybe a single channel setup screen and an all switches on individual channels screen. Not happy with any of that. Then I came up with the idea of allowing the user to drag/drop the switches onto channels and then show them visually what they end up for a Flight Mode configuration. This pull is an initial implementation of this. You start out  like this, with no switches assigned.
screen shot 2015-03-01 at 8 27 16 pm

You can then drag/drop the unassigned switches to channels. Including multiple switches to the same channel. Then down below you get a little visual of what you are building live as you build it. Here are a few examples, include everything on one channel, everything on individual channels and a hybrid example. As far as I can tell these are all legal.

screen shot 2015-03-01 at 8 27 40 pm
screen shot 2015-03-01 at 8 28 25 pm
screen shot 2015-03-01 at 8 26 39 pm

Note that the current implementation does not allow you to drag something once you assign it to a channel. That is still NYI. Wanted to get a proof of concept type thing out there for comments. The actual settings of the RC_MAP_* parameters to correct values is implemented though. So while you are moving things around params are being set appropriately. All implemented in Qml connecting to the FactSystem with no C++ code. Pretty freakin' cool stuff.

Remaining work:
- Allow dragging out from a channel to another channel or back to unassigned
- Automatically generate reasonable threshold range values for RC_*_TH
- Advanced mode: Allow user to tweak threshold ranges
- Live display of mode selection connected to transmitter like rc cal


DonLakeFlyer:
Question: The diagram above says nothing of RC_MAP_ACRO_SW. It seems like that is a variant of Manual but can't really find doc on it. Does that need to be included? Can it only be a standalone switch?

DonLakeFlyer:
Switches can now be moved between channels to re-assign. You can also drag back to Unassigned to remove a switch. Also added initial help text.

DonLakeFlyer:
@LorenzMeier @thomasgubler @dogmaphobic Hey guys. This is in a pretty workable state now. It will correctly map all RC_MAP_*_SW parameters according to what you do. It should also show you a correct visual of how this setup will work. Or at least I think it is close. It does not yet deal with setting the threshold ranges RC_*_TH values. Could you give it a try and see if it makes any sense to you?

dogmaphobic:
> Could you give it a try and see if it makes any sense to you?

I'm preparing my QML Tool Bar commit, which is quite involved. I will take a look at it once I'm done.

DonLakeFlyer:
Better tile visuals:
screen shot 2015-03-02 at 12 48 14 pm

LorenzMeier:
Awesome! With the improved tiles the readability is much better. I do have an operational question: If you assign two switches to one channel, does this work as expected? We never tried this before, and so I wasn't sure. We only ever tried one switch per channel and all switches on one channel.

LorenzMeier:
Ah and one nitty gritty detail: QGC should already know about the channel count, if not, the MAVLink message contains the relevant data. Its probably less confusing if we only show the channels which are present - but I'm sure you did already have this on your list.

DonLakeFlyer:
As far as I know you only know channel count if your transmitter is turned on.

LorenzMeier:
Fair enough. Would you like an RC count param which we set after calibration?

DonLakeFlyer:
Now supports setting of thresholds and visual display of switch ranges. The threshold will automatically be evenly distributed across the switch range. You should now be able to create any crazy switch configuration you like and you'll automatically get it setup for you.

The visual uses a vertical progress bar which will eventually be live tied to rc input. The modes are shown on the bar in their correct threshold positions. Need a better visual for this that shows the ranges better.

Everything updates on the fly as you drag modes around to channels.

screen shot 2015-03-02 at 4 13 51 pm

DonLakeFlyer:
I'm going to need a new parameter which is used to signal turning off this automatic range distribution. With a checkbox tied to that param you can then tweak thresholds yourself if you like. Or maybe I implement dragging the thresholds around in the switch display. All pretty doable.

DonLakeFlyer:
Closing this PR since it looks like the proof of concept will work. I'll finish this up and put together a real pull.

PR #1301, opened 2015-03-02
Correct order of rotations to get roll, pitch and yaw corresponding to c...

baptistelabat:
...lassical aeronautics convention and coherent with Fly view


LorenzMeier:
This changes to NED?

baptistelabat:
I think the intention was already to be in NED (that is where the minus sign on the yaw came from, and why x and y were already inverted). When heading north, roll and pitch were all right, but when heading south, roll was inverted for example.

LorenzMeier:
Ok, understood. Many thanks!

PR #1302, opened 2015-03-03
Qml Tool Bar

dogmaphobic:
### New, QML based Main Tool Bar

This is a first cut into the new QML based tool bar. It has around 90% of the functionality of the original, Qt based tool bar and some new additional features not found there. The main things still to be done:

* MAV name is not yet displayed
* Current waypoint and distance is not yet displayed
* Ability to define (in a new section of the preferences dialog) what you want to show or not.

In additions, I’m sure there will be a bit of tweaking as people start looking at it and come up with ideas and/or suggestions.

#### Things unrelated to the tool bar (but reflected whereupon) that need fixing:

* Battery capacity is not currently being reported by the system. The voltage is correct but the battery is always empty (0%).

* The messages UI component shows the current number of messages for a specific message type. In order, if there are critical/error messages, those are displayed (red background). If not, it looks for warning types (orange background). If not, it will look for “normal” messages (blue background). If there are no messages the control is not displayed at all. My doubt is what to do once the user looks at the messages (hover the mouse over the control). Once a critical message arrives, the control changes to red and stays that way. Should it go back to blue?

* An RC loss is certainly a critical message. Should an RC signal regained message also be “critical”?

* The “Connection Lost” state is out of whack. It works fine and it makes sense when I put a lead bag around the 3DR Radio Modem (cutting off the signal). When I remove the bag, connection is regained and everything is fine. But when I hit the “Disconnect” button and the link is closed, the system keeps looking for heartbeats and issue a signal when it can’t find one. It should not do that.

The old QGCToolBar code is still there (inactive). Once we’re happy with this one, I can remove it.

#### HACK ALERT

I’ve added a really ugly hack. For whatever reason, by having a QQuickWidget inside a QDockWidget (MainToolBar), the main menu is not shown when the app first starts on Mac OS. I looked everywhere, and though I found a few reference to this bug, I could not find a solution. What I did notice was that if any other window gets focus, the menu comes up when you come back to QGC. That is, if you were to click on another window and then back to QGC, the menus would appear. The hack creates a 0x0 dialog (immediately after the main UI is built) and immediately closes it. That works around the issue and it will do until I find the root of the problem.

#### Unrelated changes that came with it:

* Added a *Disconnect* modality to the *Connect* button in the Link Configuration dialog. If you select a connected link, the button changes to *Disconnect*.

* Cleaned up a bit of code debris in MainWindow along with conforming some methods and variables. This was in preparation for a subsequent PR where I will be doing a cleanup of the Main Menu.

screen shot 2015-03-02 at 11 38 59 pm


dogmaphobic:
Hmmm TC failed because of this warning:
```
cl : Command line warning D9025: overriding '/Zc:strictStrings' with '/Zc:strictStrings-'
```
This comes from qgrounfcontrol.pro:

```
WindowsBuild {
	QMAKE_CXXFLAGS_WARN_ON += /W3 \
        /wd4996 \   # silence warnings about deprecated strcpy and whatnot
        /wd4005 \   # silence warnings about macro redefinition
        /wd4290 \   # ignore exception specifications
        /Zc:strictStrings-  # work around win 8.1 sdk sapi.h problem
    WarningsAsErrorsOn {
        QMAKE_CXXFLAGS_WARN_ON += /WX
    }
}
```

Note that the other PRs have the same warning, yet they seemed to have passed.

DonLakeFlyer:
Looks like it's because the MainWindow unit test failed:

[20:56:37][Step 9/17] ********* Start testing of MainWindowTest *********
[20:56:37][Step 9/17] Config: Using QtTest library 5.4.0, Qt 5.4.0 (i386-little_endian-ilp32 shared (dynamic) debug build; by MSVC 2013)
[20:56:37][Step 9/17] PASS   : MainWindowTest::initTestCase()
[20:56:39][Step 9/17] Process exited with code -1073741819
[20:56:39][Step 9/17] Step Windows - Test (Command Line) failed

LorenzMeier:
Awesome! What level of feedback are you looking for? Do you want cross testing or are you already open to feedback to the visuals?

dogmaphobic:
> What level of feedback are you looking for? Do you want cross testing or are you already open to feedback to the visuals?

Any level and everything goes. I did test under Mac, Windows and Ubuntu but the more the merrier. At the moment, I'm trying to find out why the unit test is failing under Windows.

DonLakeFlyer:
My suggestion would be that unless something is broken let this go through as is. Then collect feedback for another pull.

LorenzMeier:
I'm building right now. My main feedback would be to address the overall height immediately, as its eating some relevant aspect of my Notebook screen real estate.

LorenzMeier:
It seems like just the mode label is consuming that?

DonLakeFlyer:
For messages ui I would suggest that dropping the dropdown clears the count to 0 and state to blue. I'd probably also change it to a real dropdown instead of a hover so that the action is more explicit.

dogmaphobic:
> It seems like just the mode label is consuming that?

Yes, but it's more a case of side effect. I had the height already set and later decided to group those 3 vertically just because I could fit them there. Once thing I didn't test was on small screens so that did not occur to me until now. I can split them up into two columns to narrow the height at the expense of a wider area.

dogmaphobic:
> For messages ui I would suggest that dropping the dropdown clears the count to 0 and state to blue. I'd probably also change it to a real dropdown instead of a hover so that the action is more explicit.

I agree with both. If it's a consensus, I can change that.

DonLakeFlyer:
Target should be 1024x768 resolution minimum. Then you need to subtract off from 768 width the Ubuntu sidebar thing to get the real minimum width. Not sure what that actually is. The toolbar should be able to fit into that width.

DonLakeFlyer:
For message dropdown if you want to get fancy, I'd only show the arrow for dropdown on hover to remove clutter and use less real estate.

LorenzMeier:
Don's suggestion for the message approach sounds great.

dogmaphobic:
> For message dropdown if you want to get fancy, I'd only show the arrow for dropdown on hover to remove clutter and use less real estate.

Only show the arrow as opposed to what? The icon? The count?

DonLakeFlyer:
I'd just jam the arrow on top of things on hover and leave everything else there to keep it simple.

dogmaphobic:
> I'd just jam the arrow on top of things on hover and leave everything else there to keep it simple.

Got it. I will see what I can do.

Installing "Debugging Tools" for Windows to find out the unit test problem...

dogmaphobic:
I've rearranged things so it should fit smaller windows. The new minimum width is now 910 pixels.
I've also changed the mouse hover behavior over the messages icon. On hover, it now shows a drop down icon. You now need to click the icon for the message window to show up.

dogmaphobic:
Oh well, after all the work installing the debug tools for Windows, I run the unit test and it passed all tests.

dogmaphobic:
How can the unit test be failing in MainWindowTest if I disabled it? That would explain why it fails there and not here. The original MainWindowTest made direct calls into the tool bar, which are no longer there. I had changed that. If the code running is the original one, it would crash as it tries to access a non existing QGCToolBar.

LorenzMeier:
Teamcity merges with master and tests the result - if you rebase this branch on master, you might get better results, as the merge happens on your end and under your control. https://help.github.com/articles/about-git-rebase/

dogmaphobic:
Ha! That's where I get messed up. I've never used *Rebase*. Primarily because of all the menacing warnings about its use. That page above has this:

> Warning: Because changing your commit history can make things difficult for everyone else using the repository, it's considered bad practice to rebase commits when you've already pushed to a repository.

That was enough to keep me away from it. What I've been doing so far was to pull mavlink/master (to my master) and *Merge* that into my work branch (*qmlToolBar* in this case). Now I'm completely lost and have no idea what to do.

LorenzMeier:
How about pulling upstream master in then? I don't see a merge commit with current master in your commit list here. It doesn't have the warnings associated like rebase has.

dogmaphobic:
That didn't seem to change much. Why is the Mac build passing? It's the same code.

Susurrus:
If you're just trying to figure out rebase, you need to use a branch that doesn't exist remotely. As in you haven't already pushed it to ANY remote repository. rebase works best when you mail patches to a mailing list, not using pull requests on Github. I suggest branching for your current branch, then doing a rebase, and opening a new PR for it.

dogmaphobic:
Yep. That's how I understood it. I do keep pushing to my remote branch as this is how I keep all 3 machines (Mac, Linux and Windows) in sync. I will try what you suggested and see in how much trouble I get :)

dogmaphobic:
Killed this and created a new one (#1307) after suggestions from @Susurrus.

DonLakeFlyer:
I rebase my branches all the time even after the pull request is sitting on github. That's how you fix merge issues as well as make sure your changes work with the latest master.

DonLakeFlyer:
Just to be more specific. I switch to my master on my fork. Git pull upstream master to bring it up to date. Then switch to the branch I'm working on and git rebase master. Now my branch is up to date with the real master

PR #1305, opened 2015-03-03
Remove old battery estimation cruft

LorenzMeier:
@dogmaphobic This should fix the battery estimation for you - please cross test. If not, the diff should light up the relevant locations to fix - it should be a quick fix, if any.

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -34,6 +34,8 @@
 
 Q_LOGGING_CATEGORY(UASLog, "UASLog")
 
+#define UAS_DEFAULT_BATTERY_WARNLEVEL 20
+
 /**
 * Gets the settings from the previous UAS (name, airframe, autopilot, battery specs)
 * by calling readSettings. This means the new UAS will have the same settings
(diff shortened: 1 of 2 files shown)


dogmaphobic:
Good. It also helped me to see that the color I had for the battery bars didn't work :) I did not look into the accuracy of the returned value but it now shows something at least. I've already fixed the colors and will be adding that to #1302.

Thanks!

snap 2015-03-03 at 16 41 37 909x60

LorenzMeier:
Cool! Looks like this is all settled then. I get to a guess-estimate of 35% charge level for a 4S machine at 15.28V, which is what the log contained as well.

LorenzMeier:
Do you have the button visuals / PNGs for the 1. / 2. / 3. buttons? That seems visually the only major thing remaining. It does look awesome already. What a difference to what it had before!

LorenzMeier:
I've merged this, so if you rebase on master you should have the fixes in the right place.

dogmaphobic:
> Do you have the button visuals / PNGs for the 1. / 2. / 3. buttons?

I assume you are talking about these chevron looking buttons. That would not work as they overlap. I'm open to suggestions though...

buttons

LorenzMeier:
If you use a Z level and put the leftmost "Pro" button on the highest level, couldn't you implement it easily? The left button will mask parts of the right button, but that should clip the area correctly and nicely:
http://qt-project.org/wiki/QML-Application-Structuring-Approaches

dogmaphobic:
I had talked about that with @DonLakeFlyer. I will see what I can do. Do we really need the **Pro** button? Part of my plan is to reorganize the main menu and add an *Advanced* menu. I would move all the **Pro** features there.

The idea is to remove the *Communication* menu (move its only *Manage Links* to the *File* menu), Add an *Advanced* menu, move the *Tools Widgets* menu as a submenu of the *Advanced* menu, as well as moving all **Pro** features.

dogmaphobic:
On the other hand, I also want to add a section to the *Preferences* dialog where you can set what is displayed in the tool bar. I guess the **Pro** button could be one of the things you can add.

LorenzMeier:
Feel free to completely hide the pro entries. I think that was more of a visual thing than a suggestion for the logical menu structure.

PR #1307, opened 2015-03-04
New, QML based Tool Bar.

dogmaphobic:
### New, QML based Main Tool Bar

This is a ~~first~~ second cut into the new QML based tool bar. It has around 90% of the functionality of the original, Qt based tool bar and some new additional features not found there. This PR supersedes #1302 (which I will be closing soon after I'm done here).

The main things still to be done:

* MAV name is not yet displayed
* Current waypoint and distance is not yet displayed
* Ability to define (in a new section of the preferences dialog) what you want to show or not.

In additions, I’m sure there will be a bit of tweaking as people start looking at it and come up with ideas and/or suggestions.

#### Things unrelated to the tool bar (but reflected whereupon) that need fixing:

* An RC loss is certainly a critical message. Should an RC signal regained message also be “critical”?

* The “Connection Lost” state is out of whack. It works fine and it makes sense when I put a lead bag around the 3DR Radio Modem (cutting off the signal). When I remove the bag, connection is regained and everything is fine. But when I hit the “Disconnect” button and the link is closed, the system keeps looking for heartbeats and issue a signal when it can’t find one. It should not do that.

The old QGCToolBar code is still there (inactive). Once we’re happy with this one, I can remove it.

#### HACK ALERT

I’ve added a really ugly hack. For whatever reason, by having a QQuickWidget inside a QDockWidget (MainToolBar), the main menu is not shown when the app first starts on Mac OS. I looked everywhere, and though I found a few reference to this bug, I could not find a solution. What I did notice was that if any other window gets focus, the menu comes up when you come back to QGC. That is, if you were to click on another window and then back to QGC, the menus would appear. The hack creates a 0x0 dialog (immediately after the main UI is built) and immediately closes it. That works around the issue and it will do until I find the root of the problem.

#### Unrelated changes that came with it:

* Added a *Disconnect* modality to the *Connect* button in the Link Configuration dialog. If you select a connected link, the button changes to *Disconnect*.

* Cleaned up a bit of code debris in MainWindow along with conforming some methods and variables. This was in preparation for a subsequent PR where I will be doing a cleanup of the Main Menu.

screen shot 2015-03-03 at 8 51 42 pm


DonLakeFlyer:
Awesome

DonLakeFlyer:
@LorenzMeier @dogmaphobic Hey guys, was TeamCity not reporting correctly on this one? TC logs show windows unit test failing yet this was merged.

LorenzMeier:
I must have missed the windows unit test - it seemed to me it was fixed eventually.

DonLakeFlyer:
I think maybe MainWindow UT was fixed, but then it just broke on SetupView UT.

PR #1309, opened 2015-03-04
Text to Speech Tweaks

dogmaphobic:
Simple cosmetic change that was annoying me. When the RC signal is lost or regained, the reported time is given in milliseconds and the text to speech engine was spelling out this large number that is mostly incomprehensible. I changed the logic so when it detects a given time in milliseconds ([0-9]*ms), it converts into seconds, or minutes and seconds when warranted. It makes for a more normal spoken message. Now, instead of QGC speaking “RC signal regained after 372695 milliseconds” it says “RC signal regained after 6 minutes and 12 seconds”.

Note that when the RC signal is lost, the time reported is the elapsed time since the PX4 was started. However, when the RC signal is regained, the reported elapsed time is the time since the signal was lost.

### DIFF
--- a/src/audio/QGCAudioWorker.cpp
+++ b/src/audio/QGCAudioWorker.cpp
@@ -2,6 +2,7 @@
 #include <QDebug>
 #include <QCoreApplication>
 #include <QFile>
+#include <QRegularExpression>
 
 #include "QGC.h"
 #include "QGCAudioWorker.h"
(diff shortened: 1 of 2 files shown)


LorenzMeier:
Cool! Makes a lot of sense! I can't imagine anyone would oppose to this change, so merging straight in.

Issue #1312, opened 2015-03-04
[new toolbar] Voltage value overlaps the battery icon


Susurrus:
While "10.48 V" is being displayed for the battery voltage I'm seeing the 1 halfway overlapping the battery icon.

LorenzMeier:
@dogmaphobic How about limiting it to the first decimal digit and human-rounding the 2nd? 10.48V is much too detailed, 10.5V is good enough. There is also enough jitter on voltage that those numbers would oscillate quite a bit in a real flight.

Susurrus:
That sounds good to me. If you needed more detail you could look directly at the MAVLink message data anyways, so no loss there.

dogmaphobic:
> While "10.48 V" is being displayed for the battery voltage I'm seeing the 1 halfway overlapping the battery icon.

That shouldn't happen. The code looks for screen density (dpi) to make sure it's consistent across platforms so things like this don't happen. What are you running and what are the physical screen dimensions?

With that said, I will round to a single decimal point.

Susurrus:
I'm seeing this problem with the satellite view as well. But I have a myriad of display problems on Fedora 21 XFCE, so I haven't really reported a lot of them as I think it's a Qt/XFCE problem and not necessary QGC (Like most text is displayed way too dark).

LorenzMeier:
@Susurrus I still think you should implement a charge level estimator. You'll confuse the heck out of anybody trying to use your boat with that ancient battery technology. Its a 3-line change and save all the argument about the icon ;-).

Susurrus:
@LorenzMeier True, but I'm writing my Ph.D. thesis, not working on making this more usable for others unfortunately. We always launched with it fully charged and it can run for longer than there's daylight. This is a waste of time for me at this stage in the project.

Susurrus:
Indeed, note that I just upgraded to Fedora 21 and this is no longer an issue. The number is close to the satellite icon for that button, but I'm closing this issue anyways.

Issue #1311, opened 2015-03-04
[new toolbar] Downarrow for message list is shown even when no messages are there


Susurrus:
I didn't realize what was going on until I realized it had no messages to display. I suggest not showing the downarrow or showing just a simple list saying "No messages" to clarify this for the user.

dogmaphobic:
If the message count is 0, the entire control should not be visible. Are you saying it is visible with 0 messages?

Susurrus:
Yep. It just shows the loudspeaker icon and nothing else in that control and the Info View's Messages tab shows nothing (just the Connect Link button as per #1276).

Issue #1310, opened 2015-03-04
Number of satellites shows 99 in new toolbar


Susurrus:
According to `common.xml`, specifying 255 for the number of satellites visible should indicate that there is no value for this field. The toolbar shouldn't show this field if that's the case.

dogmaphobic:
Good catch. The value sent containing the number of satellites is a *double*. I found it strange and even stranger than once in a while I would get -1.0 satellites. Should the Satellite control be hidden altogether in that case?

Susurrus:
I think it should be. It could also just be shown with an X. But, like Lorenz, I'm using a small screen for QGC so I'd just prefer it to be removed entirely.

LorenzMeier:
I would propose to keep the GPS icon at all times as long as we're receiving some GPS data - its a quite important status of the system. I would further propose to turn the satellite icon red if no lock is present and keep it blue if lock is present. If the number of satellites is unknown just don't show a count, else present the count.

@Susurrus would that work with you? If you have no GPS, it would also not show up. I keep myself looking in the field often for signs of GPS lock and good reception, and it would be nice to have this easily available.

dogmaphobic:
I would rather show it with an X or something to that effect. As to the screen size, I'm adding settings to the preferences dialog allowing you to select what gets to the tool bar or not. That way, you can disable it as a preference instead.

LorenzMeier:
@dogmaphobic X might indicate missing, so if the count is negative (unknown) I think leaving it out might be the better choice. You still have the lock indicator you can use to style the icon according to no lock, 2D lock and 3D lock.

Susurrus:
The coloring sounds good, but I suggest green when there's a good fix and red when it gets bad, the current blue I'm seeing doesn't help me understand what's going on. Additionally, if there's no satellite number information, just leave out the number but put it in the tooltip so people understand that the UAS isn't reporting that information.

PR #1315, opened 2015-03-04
Tool Bar Tweaks

dogmaphobic:
Tool Bar Tweaks

Per #1310, the satellite icon now has 4 color behaviors:

* No GPS: Gray
* GPS present but no lock: Red
* 2D Lock: Blue (rare)
* 3D Lock: Green

The count is only shown if more than 0 satellites are seen. The logic behind the gray icon when there is no GPS is that you (will be able to) can chose to have the icon in the tool bar as a preference. In that way, the presence of the icon is based on a preference and not automatic. It makes for a more predictable behavior.

Per #1311, I’ve fixed the message icon so the drop down triangle is not shown if there are no messages to be displayed. Clicking on the gray icon has no effect if there are no messages. The icon is always visible, even if no messages are present, following the same logic as in above.

Per #1312, I’ve changed the voltage to 1 decimal point. The battery icon has 3 color behaviors:

* No battery data: Blue
* Battery level > 40%: Green
* Battery level <= 40% Red

The reason I didn’t do gray for no battery data is because of the voltage. My next task will be to get voltage data (values for low and high) and use those to determine the icon color for when no battery level data exists. Only if there is neither battery level nor voltage data would the icon be gray.

### DIFF
--- a/src/ui/toolbar/MainToolBar.cc
+++ b/src/ui/toolbar/MainToolBar.cc
@@ -48,12 +48,14 @@ MainToolBar::MainToolBar()
     , _waypointDistance(0.0)
     , _currentWaypoint(0)
     , _currentMessageCount(0)
+    , _messageCount(0)
     , _currentErrorCount(0)
     , _currentWarningCount(0)
     , _currentNormalCount(0)
     , _currentMessageType(MessageNone)
     , _satelliteCount(-1)
-    , _dotsPerInch(72.0)
+    , _dotsPerInch(96.0)    // Default to Windows as it's more likely not to report below
+    , _satelliteLock(0)
     , _rollDownMessages(0)
 {
     setSizePolicy(QSizePolicy::MinimumExpanding, QSizePolicy::MinimumExpanding);
@@ -68,8 +70,12 @@ MainToolBar::MainToolBar()
     }
     // Get screen DPI to manage font sizes on different platforms
     QScreen *srn = QGuiApplication::screens().at(0); // TODO: Find current monitor as opposed to picking first one
-    _dotsPerInch = (qreal)srn->logicalDotsPerInch(); // Font point sizes are based on Mac 72dpi
-
+    if(srn && srn->logicalDotsPerInch() > 50.0) {
+        _dotsPerInch = (qreal)srn->logicalDotsPerInch(); // Font point sizes are based on Mac 72dpi
+    } else {
+        qWarning() << "System not reporting logical DPI.";
+    }
+    // Give the QML code a way to reach us
     setContextPropertyObject("mainToolBar", this);
     setSource(QUrl::fromUserInput("qrc:/qml/MainToolBar.qml"));
     setVisible(true);
@@ -187,7 +193,7 @@ void MainToolBar::onEnterMessageArea(int x, int y)
         _currentMessageCount = 0;
         _currentMessageType = MessageNone;
         if(count != _currentMessageCount) {
-            emit messageCountChanged(0);
+            emit newMessageCountChanged(0);
         }
         if(type != _currentMessageType) {
             emit messageTypeChanged(MessageNone);
@@ -258,14 +264,15 @@ void MainToolBar::_setActiveUAS(UASInterface* active)
     // If switching the UAS, disconnect the existing one.
     if (_mav)
     {
-        disconnect(UASMessageHandler::instance(), &UASMessageHandler::textMessageReceived,  this, &MainToolBar::_handleTextMessage);
-        disconnect(_mav, &UASInterface::heartbeatTimeout,                                   this, &MainToolBar::_heartbeatTimeout);
-        disconnect(_mav, &UASInterface::batteryChanged,                                     this, &MainToolBar::_updateBatteryRemaining);
-        disconnect(_mav, &UASInterface::modeChanged,                                        this, &MainToolBar::_updateMode);
-        disconnect(_mav, &UASInterface::nameChanged,                                        this, &MainToolBar::_updateName);
-        disconnect(_mav, &UASInterface::systemTypeSet,                                      this, &MainToolBar::_setSystemType);
-        disconnect(_mav, SIGNAL(statusChanged(UASInterface*,QString,QString)),              this, SLOT(_updateState(UASInterface*,QString,QString)));
-        disconnect(_mav, SIGNAL(armingChanged(bool)),                                       this, SLOT(_updateArmingState(bool)));
+        disconnect(UASMessageHandler::instance(), &UASMessageHandler::textMessageCountChanged,  this, &MainToolBar::_handleTextMessage);
+        disconnect(_mav, &UASInterface::heartbeatTimeout,                                       this, &MainToolBar::_heartbeatTimeout);
+        disconnect(_mav, &UASInterface::batteryChanged,                                         this, &MainToolBar::_updateBatteryRemaining);
+        disconnect(_mav, &UASInterface::modeChanged,                                            this, &MainToolBar::_updateMode);
+        disconnect(_mav, &UASInterface::nameChanged,                                            this, &MainToolBar::_updateName);
+        disconnect(_mav, &UASInterface::systemTypeSet,                                          this, &MainToolBar::_setSystemType);
+        disconnect(_mav, &UASInterface::localizationChanged,                                    this, &MainToolBar::_setSatLoc);
+        disconnect(_mav, SIGNAL(statusChanged(UASInterface*,QString,QString)),                  this, SLOT(_updateState(UASInterface*,QString,QString)));
+        disconnect(_mav, SIGNAL(armingChanged(bool)),                                           this, SLOT(_updateArmingState(bool)));
         if (_mav->getWaypointManager())
         {
             disconnect(_mav->getWaypointManager(), &UASWaypointManager::currentWaypointChanged,  this, &MainToolBar::_updateCurrentWaypoint);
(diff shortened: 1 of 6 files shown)


src/ui/toolbar/MainToolBar.cc
```diff
@@ -258,14 +264,15 @@ void MainToolBar::_setActiveUAS(UASInterface* active)
...
+        disconnect(UASMessageHandler::instance(), &UASMessageHandler::textMessageCountChanged,  this, &MainToolBar::_handleTextMessage);
+        disconnect(_mav, &UASInterface::heartbeatTimeout,                                       this, &MainToolBar::_heartbeatTimeout);
+        disconnect(_mav, &UASInterface::batteryChanged,                                         this, &MainToolBar::_updateBatteryRemaining);
+        disconnect(_mav, &UASInterface::modeChanged,                                            this, &MainToolBar::_updateMode);
+        disconnect(_mav, &UASInterface::nameChanged,                                            this, &MainToolBar::_updateName);
+        disconnect(_mav, &UASInterface::systemTypeSet,                                          this, &MainToolBar::_setSystemType);
+        disconnect(_mav, &UASInterface::localizationChanged,                                    this, &MainToolBar::_setSatLoc);
+        disconnect(_mav, SIGNAL(statusChanged(UASInterface*,QString,QString)),                  this, SLOT(_updateState(UASInterface*,QString,QString)));
```
Susurrus: Can you change this to use the new-style signal/slot definitions like above?

LorenzMeier:
@dogmaphobic I'm not sure I would recommend doing charge level estimation on the ground. There is no standard way to do it and its trivial for the autopilot to do it onboard using its system's knowledge.
Its exactly the code that led to tremendous bugs, complex logic and which I removed yesterday. Really, if the autopilot doesn't provide the data, QGC should not try to estimate it and present the user with potentially bogus data.

src/ui/toolbar/MainToolBar.cc
dogmaphobic: If you mean ```statusChanged()```, I didn't because it is an overloaded signal (really don't like those).

DonLakeFlyer: This will all go away with my next upgrade of the FactSystem. The FactSystem is meant to support multiple providers. Currently the only provider is parameters. Next provider is mavlink messages. This will replace the myriad of UASInterface signals. So for now I’d leave this alone because it will al go away at some point.

dogmaphobic:
The idea in this particular case is not trying to estimate charge. In case, and only in case there is no battery level data, I can make the icon red if the voltage level with nearing the "low mark".

LorenzMeier:
The topic is too complex to explain in writing in any efficient fashion, but to put it simple: You can't estimate the low mark in a useful way without making a whole bunch of assumption and complex user setup.
All of the mainstream systems estimate onboard and QGC should not mess with it for the remaining few that don't. You're messing with flight safety if you get it wrong. Its much more useful to show the real data: That we don't know what the battery level is.

dogmaphobic:
Fair enough. Specially if it means I don't have to do anything :)

dogmaphobic:
>  PASS   : SetupViewTest::initTestCase()
[13:37:07]W:	 [Step 9/17] Process exited with code -1073741819

Looking into it...

DonLakeFlyer:
Can you make sure to fix/turn back on the unit tests you turned off as soon as possible? The top level ones like main window and setup view catch a lot of nasty DOA type cross platform problems.

dogmaphobic:
> Can you make sure to fix/turn back on the unit tests you turned off as soon as possible?

Yes. I'm fully aware. I'm not even sure what the issue is though. It seems to run fine here (when I tried yesterday). It's just that Windows development is a major PITA given that it takes 20 minutes to build QGC compared with 2 on Mac OS and Linux. Yesterday the issue was that I had screwed up my rebasing and TC was using the wrong code. Today I have no idea yet.

DonLakeFlyer:
Windows development is a major PITA: You got that right!

I’ll take a quick look. The break is in master.

DonLakeFlyer:
I can repro the crash, but I think it's going to require a debug qt to figure it out. I don't have a Windows debug Qt 5.4 yet. Getting that going. Just FYI: That's a multi-hour build so It's going to take a while.

DonLakeFlyer:
As far as I can tell so far the break may be only unit test related as opposed to something the end user will see. Windows Qt has always been flakier than the other OS on ui based unit tests. I've fought that battle before. So for now I think it's ok to turn off SetupView UT if that will get master into a working state again.

DonLakeFlyer:
@dogmaphobic Gus. Is this ok to merge now?

DonLakeFlyer:
Also Gus in case you don't already know this. The SetupView UT crash is not related to changes in this pull. It was introduced by the new toolbar pull. It's easier to for me to see stuff like that since I can poke through the full TeamCity logs.

src/ui/toolbar/MainToolBar.cc
```diff
@@ -68,8 +70,12 @@ MainToolBar::MainToolBar()
     }
     // Get screen DPI to manage font sizes on different platforms
     QScreen *srn = QGuiApplication::screens().at(0); // TODO: Find current monitor as opposed to picking first one
-    _dotsPerInch = (qreal)srn->logicalDotsPerInch(); // Font point sizes are based on Mac 72dpi
-
+    if(srn && srn->logicalDotsPerInch() > 50.0) {
+        _dotsPerInch = (qreal)srn->logicalDotsPerInch(); // Font point sizes are based on Mac 72dpi
+    } else {
+        qWarning() << "System not reporting logical DPI.";
```
Susurrus: Might be worth stating that a specific default would be used. If we're spitting out warnings, we should give users a way to understand what was done anyways and possibly what they should do to fix this. If I saw this warning my first thought would be, okay, what kind of screen resolution is it going to use if it can't find a valid value?

dogmaphobic:
> Also Gus in case you don't already know this. The SetupView UT crash is not related to changes in this pull. It was introduced by the new toolbar pull.

I figured it later. It seems my last pull for the new tool bar never made to TC. It sat there all night and it never run. The SetupViewTest is somewhat identical to the earlier MainWindowTest I had previously disable for this same reason. It makes sense that it would crash as they both do exactly the same thing. With that said, this only happens on Windows and only with TC (it did run here through the end just fine).

> Gus. Is this ok to merge now?

Let me at least try re-enabling both for non Windows so at least we have something being tested. Once I make the commit and TC is happy, we can merge it.

src/ui/toolbar/MainToolBar.cc
DonLakeFlyer: qWarning is more like an assert with better text. Not really an end user thing.

DonLakeFlyer:
Makes sense to disable only for Windows for now. After you push that I'll let it go through TC again and merge if pass. 

My debug windows Qt build just finished. Now building master from it. I'll see what I can find.

DonLakeFlyer:
Aha, that's may be the issue with TC. If you add a new commit github may still be showing the results from the last run. Then if the TC run gets hung up somehow you my merge due to old results. I can see from the logs that the last TC run did hang up and just went through a few hours ago.

src/ui/toolbar/MainToolBar.cc
Susurrus: It is if they launch it from the command line. Let me put it this way, can you give me a good reason not to give more detail here? For developers, who see this message, it'd be nice to debug. And I do watch the console to see what QGC's doing, so I appreciate more detail. I don't see what the harm is to add another 5 words here to describe what it did since it couldn't find a logical DPI like: "System not reporting logical DPI. Defaulting to 120ppi." Is that really that bad?

DonLakeFlyer:
@dogmaphobic Ok I have the crash in debug Qt. It's the same problem I ran into before with Windows ui unit tests. Let me figure out how to hack it to make it work.

dogmaphobic:
> Ok I have the crash in debug Qt. It's the same problem I ran into before with Windows ui unit tests. Let me figure out how to hack it to make it work.

Allright. I will wait. If you find the issue, I can simply re-enabled it globally. Keep in mind that the inner portion, dealing with the toolbar directly, still needs to be disabled until we find a way to interact with it.

DonLakeFlyer:
No, go ahead and turn it off for now. This is going to take some time to track down the way to fix. I'll also figure out how to fix up MainWindow and SetupView to  work again and do the toolbar clicks while I'm at it so you don't have to deal with that.

dogmaphobic:
Well, as it is, they are both turned off. If that's the case, this PR is ready to merge. We can deal with the tests on a different PR.

LorenzMeier:
So all clear for the chevrolet icons? ;-)

DonLakeFlyer:
Chevrolet! :) Cool.

dogmaphobic:
Working on it (among 20 different things at the same time). In my early experiments with Canvas and Context2D, it would be hellish to mimic the behavior of a button. My next try would be to use transparent PNG or SVG and see what kind of trouble I can get with those...

dogmaphobic:
Which is to say, not for this PR :)

DonLakeFlyer:
@dogmaphobic Want me to make you a multi-button chevron style control? Should just be custom MouseArea hit testing. I've already stolen most of the button code to make the QGC versions so I'm pretty familiar with it. The key is one big control that holds multiple buttons instead of trying to make a single chevron button control that you use multiple times.

DonLakeFlyer:
This just passed TC. But I'm holding off merge until my trial fix goes through. Once I get some data from that I'll merge.

dogmaphobic:
How would you draw the Chevrolets? :) Just some dead area in between or part of the button? What I had done so far was using Canvas and Context2D to draw the arrow looking thing. Using Z ordering I was able to overlap them. But because I was literally painting everything from scratch, handling hover and highlight behavior turned into a nightmare. (by the way, Lorenz just beat me to the Chevrolet joke while writing this). Please feel free to deal with that as I can certainly use the time to get other things done.

Thanks!

DonLakeFlyer:
Steal the Button.qml and ButtonStyle.qml code from Qt. Make one big custom control which has all four buttons in it that is loosely based on Button. Draw them using whatever qml thing lets you do that. Write custom MouseArea hit testing code to handle the highlighting/clicking. I've been in the private Qt source for Button quite a bit if you want me to take a whack at it.

DonLakeFlyer:
By all four buttons I mean just draw four buttons in one custom control. Not actually use four really button controls.

DonLakeFlyer:
FYI: I have the fix for SetupView. I'll fix up the unit tests and then turn them back on in a separate pull. Merging this now.

PR #1316, opened 2015-03-04
Text to speech tweaks

dogmaphobic:
Changes based on suggestion from @Susurrus. The ```([0-9]*ms)``` versus ```([0-9]+ms)``` was a lapse of mine. The use of a non static rule was based on the fact this runs in a separate thread and I didn't know how Qt handled it. It's safe.


DonLakeFlyer:
Resubmitting to TC

DonLakeFlyer:
Resubmitting just to double check after UT fix went in.

PR #1318, opened 2015-03-04
Fix OS X Installer yet again

DonLakeFlyer:
macdeployqt can not be reliably used to create a dmg. Sometimes it just fails silently. Going back to using hdiutil to create the dmg.


DonLakeFlyer:
Master is broken. Hence the TC failure.

Issue #1317, opened 2015-03-04
QGC refuses to completely quit


Susurrus:
On 64-bit Linux, Fedora 21 (actually 20 as well), qgroundcontrol never quits when I tell it to. The UI closes, but when running it from the command line, the process never dies. It looks like there are still 7 processes running at this point but taking 0% CPU and holding 10% of my memory. I imagine there's a deadlock going on somewhere between threads? I don't know how the threading model for QGC works.

So to completely close QGC I have to force-kill it.

DonLakeFlyer:
Same on Ubuntu.

kd0aij:
I see only one process left running on Ubuntu

tail of console output:
Link deleted:  "ttyACM0"
SerialLinkLog: Already disconnected
UASLog: removeLink: "3f83b90"
UASLog: link count: 1
Deleting UAS object:  "MAV 001"
^C

DonLakeFlyer:
@dogmaphobic Gus, do you think you could look at this one. I'm not much of a Linux person. We may need to back port any fix to Stable V2.3.

dogmaphobic:
I'll take a look...

dogmaphobic:
I could not reproduce the problem. It exits just fine. I started a session from the command line, made a connection, messed with it for a bit and exited. It exited promptly:

```
Default compId:  50
Deleting UAS object:  "MAV 001"
After app delete
ubuntu:~/$
```

DonLakeFlyer:
This is being caused by audio output. It's hung up in a call to speak and the thread won't exit. I can repro. But I also can't get speak to say anything to me. Is espeak audio output working for folks?

kd0aij:
espeak works for me on 14.10, doesn't hang
Sound works in QGC also, and the voice sounds much nicer. Which one is it?

Susurrus:
espeak doesn't work for me on Fedora 21, but It works on Windows 8.1 for me.

DonLakeFlyer:
> espeak works for me on 14.10, doesn't hang

@kd0aij Can you look at a stack trace for when shutdown hangs? If it's in the GAudioOutput destructor on one of the threads than that is why it's hanging.

kd0aij:
#0  pthread_cond_wait@@GLIBC_2.3.2 () at
../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
[...]

PR #1320, opened 2015-03-05
Unit Test fixes

DonLakeFlyer:
- Work around strange Window Qt bug/feature/something strange
- SetupView and MainWindow tests modified for new toolbar and turned back on.

### DIFF
--- a/src/AutoPilotPlugins/PX4/Tests/FlightModeConfigTest.cc
+++ b/src/AutoPilotPlugins/PX4/Tests/FlightModeConfigTest.cc
@@ -31,7 +31,9 @@
 ///
 ///     @author Don Gagne <don@thegagnes.com>
 
-UT_REGISTER_TEST(FlightModeConfigTest)
+// Turned off due to signletonn shutdown sequencing problem. Also since FlightMode config is
+// being re-written this test will need to be re-written as well.
+//UT_REGISTER_TEST(FlightModeConfigTest)
 
 FlightModeConfigTest::FlightModeConfigTest(void) :
     _mockUASManager(NULL),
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
@dogmaphobic If you look at this pull you'll see the fix for the crash in MainWindow.cc. Theoretically doing a setVisible(true) on widget should be a no-op since widgets automatically start out visible if you parent them to a visible parent. Some thing for doing a separate setParent call instead of doing it in the constructor. Unfortunately with Qt, theoretical doesn't cut it. Both setParent and setVisible set off a shitstorm of activity. Hence it's a good idea to always provide a parent in the constructor for QWidgets an QObjects. Specifically setVisible(true) will cause Qt to crash when running unit tests due a bug or something in revoking drop targets deep inside Qt. So we run unit tests with invisible windows for testing. Plus by doing that I don't have windows randomly pop up all over my computer when TC runs!

DonLakeFlyer:
I should have mentioned that I ran into another singleton shutdown problems in FlightModeConfigTest. The shutdown mechanism just doesn't work cross threads. There needs to be more coordination between the main thread on the singletons which are on their own threads to make it works correctly. This is likely related to Issue #1317 as well. I'm going to need to dive back in here and fix this for real this time.

dogmaphobic:
Funny you mentioned it. I noticed just this morning the *setVisible(true)* and wondered what the need of it was. It's detritus from copying/pasting code and not really paying attention to see if these calls are really necessary for children widgets. And yes, it bothers me to create widgets with no parents. I had it at some point when I was using QQuickView and removed it when switching to QGCQmlWidgetHolder.

Thanks for cleaning that up.

PR #1321, opened 2015-03-05
Completed Tool Bar settings in Preferences Dialog

dogmaphobic:
I've added a Tool Bar section to the general section of the preferences dialog where you can set what to display in the tool bar. For now you can set if these items are shown:

* GPS
* Battery Status
* Messages
* Mav Icon

I've cleaned up a bit the dialog while at it though it seems the preferences dialog is also in need of quite a bit of TLC.

snap 2015-03-04 at 22 15 24 500x604

### DIFF
--- a/src/ui/MainWindow.h
+++ b/src/ui/MainWindow.h
@@ -126,12 +126,8 @@ public:
     /// @brief Restore (and connects) the last used connection (if any)
     void restoreLastUsedConnection();
     
-#ifdef UNITTEST_BUILD
-    // Returns a pointer to the MainToolBar so that unit tests can change views.
+    /// @brief Gets a pointer to the Main Tool Bar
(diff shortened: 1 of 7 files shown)


DonLakeFlyer:
We generally don't use the on_showGPS_clicked style syntax for signal connection. The reason is that if the UI changes the signal connection can be too easily broken without anyone realizing it. Hence all the direct connect with an &reference and not SIGNAL/SLOT style calls. This way compile breaks when things get out of sync. We had piles of these problems before we switched to the more anal syntax.

dogmaphobic:
Do you want me to change or let it be for the time being? The entire preferences dialog need a revamp at some point any way. Are we keeping it Qt or switching it Qt Quick as well?

DonLakeFlyer:
Not the end of the world. Should go Qml at some point down the road.

PR #1322, opened 2015-03-05
Remove unused code

DonLakeFlyer:
Not even being compiled.


dogmaphobic:
Have you looked at the files under files/images? Only a fraction of them are used :) It's so many I thought about writing a python script to scan the tree and see if they are referenced anywhere...

DonLakeFlyer:
I have an issue on it somewhere in here related to creating installers. The reason being installer creation copies the full file tree to the installer since it can't tell what is used and what isn't. In reality the installer should not need to copy anything from files, since all files should get compiled into the resource. But there are places which still reference the file path directly instead of using resources. The python script would need to:
- Grovel the qgroundcontrol.qrc looking for references
- Grovel the source looking for direct file system references to the "files" true instead of using the resource path (then these need to be fixed)

I would like to get rid of the whole "files" location completely. Lately I've just been putting the resources right next to the code, or maybe in an images subdirectory below. This makes it way easier to see all the pieces in one place.

PR #1325, opened 2015-03-06
Use FactBinder to access facts from Qml

DonLakeFlyer:
Passing parameters QVariantMap across C++ boundary was causing crashes in Qt due to Qml array bugs. Referencing same Fact* across Qml boundary in multiple QQuickWidgets caused jScript GC to crash.

You now create a Fact object in your Qml and connect it to the appropriate fact by setting the "name" property: "property Fact fact:  Fact { name: "RC_MAP_THROTTLE }". The autopilot.parameters mechanism has been removed.

This way the behind the scenes FactBinder C++ object is created in the Qml context side of things thus preventing cross-widget GC references which cause things to crash.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -774,6 +774,7 @@ INCLUDEPATH += \
 HEADERS += \
     src/FactSystem/FactSystem.h \
     src/FactSystem/Fact.h \
+    src/FactSystem/FactBinder.h \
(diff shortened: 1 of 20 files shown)


DonLakeFlyer:
@dogmaphobic Gus: Note the new mechanism for accessing Facts from Qml.

PR #1329, opened 2015-03-06
New Flight Mode config

DonLakeFlyer:
This is a fairly complex piece of C++ and Qml. So it could use some testing from others prior to merge. The new Flight Mode config:
- Allows you to drag flight modes to channels and re-arrange them as you like
- You can assign more than one flight mode to the same channel
- You can have the config automatically set up thresholds for the switches
- You can turn on live RC display to test you switches and thresholds
- It will validate your current flight mode config params to make sure that they are in a valid state prior to letting you use the config. If they are invalid you will need to correct them by changing params directly.

Here is what it looks like when you have no switches assigned. Note the warning about mode switch being required:
screen shot 2015-03-06 at 11 43 16 am

Here is what it looks like with just the mode switch assigned:
screen shot 2015-03-06 at 11 43 29 am

All switches on seperate channels with auto setup of thresholds on. Note how the single function switches do not have their thresholds set to mid-point. This is on purpose. By not setting at mid-point the threshold will work with either a two or three position switch.
screen shot 2015-03-06 at 11 43 42 am

Here are all channels on the same switch with automatic threshold set:
screen shot 2015-03-06 at 11 43 55 am

Here are the two types of validation errors you can get:
screen shot 2015-03-06 at 11 30 30 am
screen shot 2015-03-06 at 11 32 07 am


DonLakeFlyer:
FYI: These visuals are meant to be functional and not necessarily pretty. That will come once this gets merged in and reviewed by the UI designers.

DonLakeFlyer:
I"ve done some more testing with this. I'm going to merge after I fix the conflict and repass TC. That way we can't daily builds with it and Damiano can also give it a try.

LorenzMeier:
Sounds good

DonLakeFlyer:
I'm going to resubmit this since the rebase went strange.

PR #1332, opened 2015-03-06
Chevron buttons in Main Tool Bar

dogmaphobic:
Added custom "chevron" buttons to the main tool bar.

screen shot 2015-03-06 at 3 48 06 pm

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -247,13 +247,11 @@
     <qresource prefix="/qml">
         <file alias="test.qml">src/test.qml</file>
         <file alias="QmlTest.qml">src/QmlControls/QmlTest.qml</file>
-
         <file alias="QGroundControl/FactControls/qmldir">src/FactSystem/FactControls/qmldir</file>
         <file alias="QGroundControl/FactControls/FactLabel.qml">src/FactSystem/FactControls/FactLabel.qml</file>
         <file alias="QGroundControl/FactControls/FactTextField.qml">src/FactSystem/FactControls/FactTextField.qml</file>
         <file alias="QGroundControl/FactControls/FactCheckBox.qml">src/FactSystem/FactControls/FactCheckBox.qml</file>
         <file alias="QGroundControl/FactControls/FactComboBox.qml">src/FactSystem/FactControls/FactComboBox.qml</file>
-
         <file alias="QGroundControl/Controls/qmldir">src/QmlControls/qmldir</file>
         <file alias="QGroundControl/Controls/SubMenuButton.qml">src/QmlControls/SubMenuButton.qml</file>
         <file alias="QGroundControl/Controls/IndicatorButton.qml">src/QmlControls/IndicatorButton.qml</file>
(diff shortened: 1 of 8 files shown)


LorenzMeier:
Awesomeness! Should we remove the numbers in the labels, as these are now kind of implicitly clear? Would also free some space 8)

DonLakeFlyer:
Make sense to me. Hey Gus how do you do the thing where you quote text from a previous comment?

dogmaphobic:
> Gus how do you do the thing where you quote text from a previous comment?

You select what you want to quote and press the "R" key in the keyboard :)

dogmaphobic:
Before connection:

snap 2015-03-06 at 16 18 13 927x40

After:

snap 2015-03-06 at 16 18 53 926x41

dogmaphobic:
Note that the "hover forever" bug is still there. The bug is in QtQuick where the mouse "in button" state is left set to true when you move the mouse away fast enough for it not to notice.

DonLakeFlyer:
> Note that the "hover forever" bug is still there.

I'll get working on that. It's starting to get really annoying!

dogmaphobic:
I didn't spend too much time with it so I didn't see any clean solution. Note that for the message icon, which has a hover state, I simply added a timer to make the drop down triangle go away. That work around would not be good enough for button hovers because it makes it weird (the button going back to normal while you still have the mouse over it). The timer would have to see where the mouse really is but because it's outside the control's bounding rectangle, it won't be simple.

DonLakeFlyer:
Going to have to see what happens with MouseArea.mouseX|Y does when the mouse is outside the area.

dogmaphobic:
> Going to have to see what happens with MouseArea.mouseX|Y does when the mouse is outside the area.

It returns the last coordinates before you moved the mouse away from the control. That is, the coordinates are always within the bounding rectangle. That's what I meant about it not being simple. You may need to get the current mouse location directly from the main window.

DonLakeFlyer:
Yup. It's going to need a cheater C++ object which gives you real mouse x/y. Then some sort of timer compare type thingy.

qgroundcontrol.qrc
```diff
@@ -247,13 +247,11 @@
     <qresource prefix="/qml">
         <file alias="test.qml">src/test.qml</file>
         <file alias="QmlTest.qml">src/QmlControls/QmlTest.qml</file>
-
```
DonLakeFlyer: Any reason to pull out these spaces? I put them in there to help with grouping. Makes it easier to read.

DonLakeFlyer:
@LorenzMeier What do you think about text localization? I've kind of given up on trying to do this correctly (using tr in C++ and qsTr in Qml). Reason is that it is missing from so many places already. Also I just don't see QGC getting localized ever. That's a huge amount of work. Not just string translation but you have to build UI that will react to strings being longer/shorter. Too me that is out of scope for QGC.

Reason I bring it up here is I noticed the qsTr usage in Gus' Qml.

qgroundcontrol.qrc
dogmaphobic: I didn't remove any of these on purpose. It must have been from not this merge but the last one where I had a lot of problems and had to do a lot these by hand. But come to think of it, if that was the case it would not show the differences now. I have no idea... I didn't even look at this file this time around. It was a clean merge and I didn't edit it by hand.

dogmaphobic:
I do it as a force of habit. I agree that localizing QGC would be a monumental task, specially considering that a lot of messages come from Mavlink.

dogmaphobic:
Speaking of which (pun intended), what happens with text to speech when you run QGC on a computer with a different locale? I would imagine the text to speech would follow the locale and get quite confused trying to speak English sentences.

qgroundcontrol.qrc
DonLakeFlyer: Got it. Fine.

DonLakeFlyer:
Ok. Any reason not to merge this now?

dogmaphobic:
None that I can think of. I want to tweak the colors of the various elements but I want to do that on a different PR and I need this in before starting (too much hassle to merge it later).

Issue #1270, opened 2015-02-25
QGCButton hover state doesn't always clear correctly


DonLakeFlyer:
If the button is right next to the edge of a QQuickWidget you can move the mouse out of the widget and hover will still be set. Also in Firmware Upgrade when you click Upgrade it gets stuck in hover even after you answer Yes/No to the warning dialog. Will need to implement some sort of timer check to fix this since Qt Hover support is so broken.

LorenzMeier:
Fixed.

PR #1333, opened 2015-03-06
Fix hover state of Button

DonLakeFlyer:
Fixes Issue #1270

### DIFF
--- a/src/QmlControls/QGCButton.qml
+++ b/src/QmlControls/QGCButton.qml
@@ -4,6 +4,7 @@ import QtQuick.Controls.Styles 1.2
 import QtQuick.Controls.Private 1.0
 
 import QGroundControl.Palette 1.0
+import QGroundControl.MousePosition 1.0
 
 Button {
     // primary: true - this is the primary button for this group of buttons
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
@dogmaphobic If you change QGCToolBarButton to be based on QGCButton instead of Button you'll get this for free.

src/QmlControls/QGCButton.qml
```diff
@@ -11,32 +12,50 @@ Button {
...
+        id:         hoverTimer
+        interval:   250
+        repeat:     true
+
+        onTriggered: {
+            if (__lastCursorX != __mousePosition.mouseX || __lastCursorY != __mousePosition.mouseY) {
+                __forceHoverOff = true
+            } else {
+                __forceHoverOff = false
             }
```
dogmaphobic: I haven't tried it, but I don't see where you map both mouse coordinates to world. Are you comparing the MouseArea coordinates (where its 0,0 is the 0,0 within the control) and the cursor position returned by the main window (which has its 0,0 based on the main window 0,0)?

dogmaphobic: As far as I can tell, you cannot get a world coordinate from within QML. What are available are methods for mapping coordinates between elements (Items) and their parents/children. Never outside. The coordinates you get from mouseX/Y within the MouseArea are based on its bounding rectangle (0,0 to width,height). The coordinates you get from QPoint::pos() are global coordinates (the entire screen). To compare mouseX to QPoint::pos().x you would need to map from the Item's coordinate space to world space. The best I can think of is passing the QObject as an argument to the c++ side and map it there.

dogmaphobic: Didn't I mention this wasn't simple? :)

DonLakeFlyer:
Sorry the variable names are crappy so it's not very obvious what is going on here. I am not trying to map and compare between local and global coordinate system. Here is what it does:
- MousePosition.mouseXY returns the current mouse position in global coordinates
- lastMouseXY are only bound so I can get triggers for onLastMouseXYChanged. I don't really care about the values.
- lastMouseXY will only update and trigger onLastMouseXYChanged while the mouse is within the bounding rectangle
- if the mouse moves outside the bounding rectangle, outside of the QQuickWidget causing the bug, then lastMouseXY will no longer update
- each time lastMouseXY changes I keep track of the last global mouse position in lastCursorXY
- when lastMouseXY stops tracking because of bug, lastCursorXY will be left at last global cursor position we saw
- then the timer kicks off and checks whether the current global mouse positions differs from the last registered global position. If those differ it means that the mouse have kept moving past the bounds of the bounding rectangle. Hence hover should be force off. If they are equal it means than the mouse is standing still within the bounding rectangle, hence hover should stay one.

Phew! Hopefully I explained that in a way that makes sense to someone other than me! It's a bit strange, but it works.

DonLakeFlyer:
Fixed the variable names, and realize I don't need the lastMouseXY variables. I can still connect to the signals.

dogmaphobic:
Classic case of not seeing the forest for the tree. I was so deep into the details I missed the genius of the simplicity of it :)

Issue #1326, opened 2015-03-06
[new toolbar] ARMED and EMERGENCY are easy to confuse


Susurrus:
Since the colors of ARMED and EMERGENCY are both the same color of red, they're easy to confuse. I would expect red to denote an error condition, not necessarily one that is just a warning. Maybe yellow can be used instead for ARMED, like it had been for the old toolbar?

dogmaphobic:
Certainly. The colors in the tool bar are generally experimental as they are not part of the main QGC palette yet (other then buttons and window). When I first started, I had it orange. We need to figure out these colors and make it part of the main QGC palette. @DonLakeFlyer Can you comment on these? Should they be part of QGCPalette?

DonLakeFlyer:
I would say no, these shouldn't be in the palette. Reason: They should stay the same color in both the light and dark themes. That's kind of the rule for now. Kind of the same as the little red/green indicator dots that are all over setup. Those aren't in the palette either. A reason to break that rule would be if you have a useful color that you want to keep same across multiple Qml files, even if it doesn't change in light/dark.

Issue #1327, opened 2015-03-06
[new toolbar] EMERGENCY/ARMED are too low contrast on the dark theme


Susurrus:
The red used for these labels is too dark and low contrast for the dark theme. They look good on the light theme, but the dark they're hard to read. Maybe brighten up the red a bit?

dogmaphobic:
Same general issue as #1326. We just need a properly setup palette.

PR #1335, opened 2015-03-07
Tool bar tweaks

dogmaphobic:
Tweaked text colors in main tool bar.
Tool bar buttons now use QGCButton as base to inherit hover mode fix.
Made QGCButton.showHightLight into a public property so I could use it as a trigger for repainting the chevron buttons.


dogmaphobic:
These tweaks fix issues #1326 and #1327

PR #1337, opened 2015-03-07
New Flight Mode config

DonLakeFlyer:
Replacement for Pull #1329


dogmaphobic:
For what's worth, I'm getting an assert when switching from Radio to Flight Modes. The assert is in FactBinder.cc line 64 (```FactBinder::setName()```). The name being set is **RC_CHAN_CNT**.

LorenzMeier:
I can see how that happens. You need to update the Firmware to make it go away. But we need to handle this in general. @DonLakeFlyer Can we change the assert to something less intrusive?

I have introduced the SYS_PARAM_VER param for you to check parameters versions, and I hot fixed it in 2e1efdc. I'm not sure that's the right way to fix it, but I wanted to unblock things - so please feel free to rework the "fix".

DonLakeFlyer:
My screw up. That parameters should be optional.

PR #1339, opened 2015-03-08
Initial Battery Config

dogmaphobic:
I wanted to get this in so the files are all in place and it doesn't become a merge nightmare later as other things get added in parallel. In other words, because it will take a while to complete these, I would like to get this merged into master sooner than later so the accumulated changes don't get out of control.

This is a very basic *Battery Configuration* (part of *Power Configuration*), without any fancy UI elements yet.

The draft calls for *Propeller Function* and *Magnetometer Distortion* to be in the same panel. May I suggest adding instead a *Show Advanced Settings* check box and only then showing these last two items?

P.S. @DonLakeFlyer Now I noticed your blank lines were again eaten. There must be something on my end "cleaning" these blank lines. I just need to find what is doing it.


LorenzMeier:
I think hiding the advanced two in an advanced config widget makes sense, yes, and getting in a basic widget also makes a lot of sense so the structure is in place.

dogmaphobic:
> I think hiding the advanced two in an advanced config widget makes sense, yes

Perfect. The *Voltage Drop* would also be part of these *Advanced Settings* (along with other battery settings).

When do you use *Battery Capacity*? Is this something that should be exposed? I ask primarily because I have several batteries of different capacities and it would be cumbersome to change this setting every time I change batteries.

pfg:
You could use voltage drop, rather than capacity. The presumption would be a constant number of cells means a constant voltage drop from full charged to empty. In practice would have to be heavily averaged to give meaning full results during flight but its still better than nothing. 

Date: Sun, 8 Mar 2015 13:32:21 -0700

dogmaphobic:
I've created new ScreenTools registered type for QML code to obtain the *Screen DPI Factor*. This is used for computing font point sizes across platforms. This code used to be in MainToolBar.cc/.h providing only the tool bar with its functionality. Now it is available to all QML widgets.
Moved the global mouse position tool into ScreenTools as I felt it would be a place to collect these sort of things.
Took the liberty of replacing the tree used in the Safety Component from a PNG file to a SVG file as the PNG was not scaling up properly.

dogmaphobic:
P.S. Looking at the *Changed Files*, it looks as if I "took ownership" of the ScreenTools.* files. That was not my intention! :) I had created the ScreenTools and only later I decided it would make more sense to combine both into one. That's when I moved the two mouse functions into ScreenTools and removed MousePosition. Git is making it look as if I took it over! :)

dogmaphobic:
Started working on some battery graphics. The Canvas bug is there (if you switch from Setup to something else and back, the arrows are not repainted). I will add the repaint code I added to the tool bar to ScreenTools tomorrow so it can be used globally.

screen shot 2015-03-09 at 2 01 13 am

dogmaphobic:
Note to self:

Add *(in Series)* to *Number of Cells*
Add *S* as a unit for the *Fact*

dogmaphobic:
Any objections to merge this once TC clears it?

PR #1341, opened 2015-03-09
Add assert back

DonLakeFlyer:
Also changed to qWarning. This usually indicates a bad problem so putting assert back.

### DIFF
--- a/src/FactSystem/FactBinder.cc
+++ b/src/FactSystem/FactBinder.cc
@@ -65,7 +65,8 @@ void FactBinder::setName(const QString& name)
             emit valueChanged();
             emit nameChanged();
         } else {
-            qDebug() << "FAILED BINDING PARAM" << name << ": PARAM DOES NOT EXIST ON SYSTEM!";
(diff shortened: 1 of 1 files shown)


LorenzMeier:
I totally disagree with this change. A non-existent parameter will continue in the future (e.g. users upgrading either QGC or the Firmware with a several month step).

It is ridiculous to force an app crash via an assert in that situation. It will leave a really bad impression with users. There has to be a better way to do this.

DonLakeFlyer:
Asserts turn into nothing in release builds. They'll just get incorrect behavior. For debug builds I want developers to come to a dead stop right away. Any code which references new parameters needs to check for existence before it is used. If it doesn't, it is  really bad bug which basically will cause QGC to not work correctly, hence the assert.

LorenzMeier:
I was trying to read up on it, but got very confused - are we setting the define to disable them?
http://qt-project.org/doc/qt-4.8/qtglobal.html#Q_ASSERT

DonLakeFlyer:
Release build automatically does that. All installers are release build

LorenzMeier:
Ok, nevermind. That makes of course more sense than using qDebug().

PR #1344, opened 2015-03-09
Support for bootloading uNode board

crashmatt:


crashmatt:
Added board ID for uNode board.  Simple.

LorenzMeier:
I had already merged your earlier commit. Please compare against what is in master.

crashmatt:
How?  I only did pull request on this issue once.
Unless this was from the outstanding earlier pull request from the previous stuff.

The only difference is the comment on the board ID

PR #1346, opened 2015-03-10
Added code to handle point size fonts in all QML widgets.

dogmaphobic:
When defining fonts in point size, you need to take into account the display pixel density. Otherwise you end up with varying font sizes depending on the platform used (i.e. 72dpi for Mac OS, 96dpi for Windows, etc.) When I created the new tool bar (as a QML widget), I added code to handle this. I've now replicated it in all QML widgets.

While at it I've added missing licensing headers to the files I touched and documented *ScreenTools*.

### DIFF
--- a/src/AutoPilotPlugins/PX4/FlightModesComponent.qml
+++ b/src/AutoPilotPlugins/PX4/FlightModesComponent.qml
@@ -1,3 +1,26 @@
+/*=====================================================================
+
+ QGroundControl Open Source Ground Control Station
+
+ (c) 2009 - 2015 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+
+ This file is part of the QGROUNDCONTROL project
+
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
How about changing this to be something like screenTools.dpiAdjustedPointSize(20) which would just return the correct point size. Would remove some of the noise of all the multiplies in the code. Then you just

font.pointSize: screenTools.dpiAdjustedPointSize(20)

dogmaphobic:
I debated that with myself. It would certainly make for a cleaner looking code. My doubt was about the time it would take for the argument to make the round trip as opposed to doing the multiplication within JavaScript. Come to think of it, this is so rarely done that it's negligible one way or the other. I will change it as suggested.

dogmaphobic:
> FactSystemTestPX4::qml_test()

Looking into it...

dogmaphobic:
It passed here and as far as I can tell, I'm fully up-to-date with master. Any clues?

PR #1348, opened 2015-03-10
Cleanup remnants OpalRT and Kinect

Susurrus:


Susurrus:
Forgot to check the post-install commands. That last commit should bet just the ticket.

PR #1352, opened 2015-03-12
Crosstrack error

Susurrus:
Fix some dead code in both the HSI widget and the PFD widget. I think everything's working here, though the CDI is not like the pictures I've seen on the internet. Hopefully someone else has the `NAV_CONTROLLER_DATA` message being broadcast with correct data and can test this out and confirm it's correct.

The `NAV_CONTROLLER_DATA` message is lacking a few descriptions, so I'm just guessing that the `target_bearing` field is the subtraction of the bearing that the current mission segments has from the current bearing (course over ground).

And should the CDI have an arrow? It's not obvious what it's saying when you're angled 90deg to the current mission track.

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -1342,6 +1342,7 @@ void UAS::receiveMessage(LinkInterface* link, mavlink_message_t message)
             setDistToWaypoint(p.wp_dist);
             setBearingToWaypoint(p.nav_bearing);
             emit navigationControllerErrorsChanged(this, p.alt_error, p.aspd_error, p.xtrack_error);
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
I'm not familiar with any of this stuff, so maybe @LorenzMeier  can help out here.

LorenzMeier:
@Susurrus Could you add a few screenshots? Its quite hard to comment on the changes without seeing what they actually do, and its adding significant overhead for me for the review to recreate a data flow I believe shows a good test case.

Susurrus:
Sorry I hadn't gotten around to this, but thanks for merging anyways. It's mostly existing code that I just enabled.

PR #1354, opened 2015-03-12
Output audio for NOTICE severity STATUSTEXT messages.

Susurrus:
NOTICE messages are "unusual" events that shouldn't be seen during normal operations. In this context I would like to have them output as audio automatically.

Also fixed a bug where setting "#" at the start of the string triggers the audio reading of a message, but what was removed was always exactly "#audio:", which could result in odd behavior. I'd actually propose just using the '#' for indicating that it should be read over audio, but no need to follow through on that here.

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -1253,11 +1253,11 @@ void UAS::receiveMessage(LinkInterface* link, mavlink_message_t message)
             QString text = QString(b);
             int severity = mavlink_msg_statustext_get_severity(&message);
 
-            // Now if the status message starts with "#audio:", or it's at least a
-            // severity level of NOTICE or higher, output it as an audio message.
-            if (text.startsWith("#audio:") || severity <= MAV_SEVERITY_NOTICE)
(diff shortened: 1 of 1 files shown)


Susurrus:
@DonLakeFlyer @LorenzMeier Any comments on this? Does the PIXHAWK output a lot of notices that this would be a problem?

DonLakeFlyer:
Don't know. Lorenz will have to weigh in.

LorenzMeier:
That's fine. We can use the hashtag-only approach. Most of the messages requiring audio have been bumped to the appropriate priority and the audio tag has been removed.
Thanks!

LorenzMeier:
On your proposal: " I'd actually propose just using the '#' for indicating that it should be read over audio, but no need to follow through on that here."

Wy don't we include this just now? It will ensure that audio strings are still read then in case they use the "#a" syntax we had been using intermittently.

Susurrus:
Done. Once it passes TC I'll merge it.

PR #1357, opened 2015-03-12
Serial port fix

dogmaphobic:
For whatever reason, QSerialPort stopped working when you provide the “short” name for the port you want to open. As far as I can tell, this is how it’s always beed done (the short version, I mean). All my 3 platforms, Mac OS, Linux and Windows stopped working. I switched to use the full port name and that fixed the problem.

I do not know the reason for it all to stop working and it is specially weird given that I have not heard of anybody else having issues.

The error I was getting when attempting to open the port was *Not Found*. Opening the port from the command line worked just fine so I knew it had to be QGC (QSerialPort).

When debugging, I switched from *usbserial-AH02QFUO* to */dev/cu.usbserial-AH02QFUO* and it then found the port just fine. From there I switched the port name initialization for using the full name (*SerialConfiguration.systemLocation*) instead of the short version (*SerialConfiguration.portName*).

Because of the port name change, I had to bump the settings version number so it wouldn’t load the old port names.

### DIFF
--- a/src/QGCConfig.h
+++ b/src/QGCConfig.h
@@ -12,7 +12,7 @@
 
 // If you need to make an incompatible changes to stored settings, bump this version number
 // up by 1. This will caused store settings to be cleared on next boot.
-#define QGC_SETTINGS_VERSION 3
+#define QGC_SETTINGS_VERSION 4
 
 #define QGC_APPLICATION_NAME "QGroundControl"
 #define QGC_ORG_NAME "QGroundControl.org"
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
You tested this on Windows right? I seem to remember something strange with port names on Windows.

dogmaphobic:
Yes, I made sure to test in all 3 platforms. On Windows, the full port name is "\\.\COMx".

DonLakeFlyer:
No I remember. At some point it would ne nice to figure out what is wrong here. The reason being that the Windows port names just end up being bizarre for user display.

dogmaphobic:
You're right. I do rename the weird looking port name when I do the automatic renaming of "Unnamed" ports but I forgot about the port drop down list shown to the user. That still shows "\\\\.\COM4". Let me fix that in a bit..

PR #1358, opened 2015-03-13
PrimaryFlightDisplay: Clear up speed label (now showing airspeed and gro...

philipoe:
...undspeed separately to avoid confusion) and add vertical climb label.

Explanation: The current display of speed (either airspeed or groundspeed) in the Primary Flight Display (PFD) can be highly confusing, because QGC internally switches between groundspeed and airspeed. The user can thus never be sure what he actually sees. In heavy wind, there is however a significant difference between ground and airspeed.

This Pull Request thus adds a small modification to the PFD:
 - modifies the velocity meter to show both Airspeed (AS) and Groundspeed (GS) at all times. When we are a a fixed-wing Airplane, AS is shown large and GS shown small, when airspeed is NAN or we are not an airplane then GS is shown large.
 - adds the display of the vertical climb rate (vZ, as a label) to the altimeter.
 - Harmonizes the code and the layout of altimeter and velocitymeter

Let me know if any modifications (e.g. label names for compatibility) are desired.

Preview:
primaryflightdisplay

### DIFF
--- a/src/ui/PrimaryFlightDisplay.cc
+++ b/src/ui/PrimaryFlightDisplay.cc
@@ -999,15 +999,15 @@ void PrimaryFlightDisplay::drawAltimeter(
 		QString s_climb;
 		s_climb.sprintf("vZ:%2.1f", climbRate);
 		drawTextRightCenter(painter, s_climb, smallTextSize, rightEdge - 4 * lineWidth, 1.5*mediumTextSize + markerOffset);
-		
+
 		// Draw climb rate indicator as an arrow
 		float vvPixHeight = -climbRate/ALTIMETER_VVI_SPAN * effectiveHalfHeight;
-        if (abs (vvPixHeight) < markerHalfHeight)
+		if (vvPixHeight > -markerHalfHeight && vvPixHeight < 2 * markerHalfHeight + markerOffset)
             return; // hidden behind marker.
 
         float vvSign = vvPixHeight>0 ? 1 : -1; // reverse y sign
 
-		QPointF vvArrowBegin(rightEdge - w*ALTIMETER_VVI_WIDTH / 2, (vvSign ? 2*markerHalfHeight+markerOffset : markerHalfHeight));
+		QPointF vvArrowBegin(rightEdge - w*ALTIMETER_VVI_WIDTH / 2, (vvSign>0 ? 2*markerHalfHeight+markerOffset : -markerHalfHeight));
         QPointF vvArrowEnd(rightEdge - w*ALTIMETER_VVI_WIDTH/2, vvPixHeight);
         painter.drawLine(vvArrowBegin, vvArrowEnd);
 
--- a/src/ui/PrimaryFlightDisplay.cc
+++ b/src/ui/PrimaryFlightDisplay.cc
@@ -921,7 +921,7 @@ void PrimaryFlightDisplay::drawAltimeter(
         effectiveHalfHeight -= secondaryAltitudeBoxHeight;
     }
 
-    float markerHalfHeight = mediumTextSize*0.8;
+    float markerHalfHeight = mediumTextSize;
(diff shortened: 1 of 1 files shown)


Susurrus:
Airspeedmay also be desired for boats, where I'm at least using it to denote the true water speed.

philipoe:
So this should be OK for your use case because even then, the airspeed would still be shown (just in small-size text below the medium-size text for groundspeed) ?

Susurrus:
Sorry, I should be more clear. It might be desirable to make airspeed the primary display for boats as well as they'd likely put water or hull speed there. Not certain which captains find most useful. I think I care more about groundspeed.

DonLakeFlyer:
Is it valuable to show both airspeed and groundspeed at the same time? Or should this really be configurable like this:
-Default: fixed wing: only AS, non fixed-wing: only GS
- You pick only AS or only GS

Susurrus:
Why not fixedwing get only airspeed where all others get both unless they're not reporting one or another (i.e. they're NaNs)? Then other systems can "choose” by just modifying what speeds they report.

DonLakeFlyer:
> where all others get both

I'm not sure if that is actually useful or not. Hence the question. Don't won't to clutter the display with useless information.

Susurrus:
My reasoning is that QGC can be used with tons of different vehicles (see `MAV_TYPE` enum), and while I know you guys are focused on fixed-wing and rotor vehicles, I think my suggestion works for everyone pretty well. The rotors don't output airspeed, correct? So they just see groundspeed. And the fixed-wing vehicles, we've already stated should be a special case. Then for every other vehicle, they can make their own choice as to what they report, and if they report both, it's assumed that both values are "useful" to the operator.

DonLakeFlyer:
> Why not fixedwing get only airspeed where all others get both unless they're not reporting one or another (i.e. they're NaNs)?

@philipoe Can you update to this?

philipoe:
> Why not fixedwing get only airspeed where all others get both unless they're not reporting one or another (i.e. they're NaNs)?

The question here is whether you consider the two variables (AS+GS) as too much cluttering of the PFD. If yes, then why would we want to "save" fixed-wings from cluttering, while all other systems have a "cluttered" PFD? 

> Is it valuable to show both airspeed and groundspeed at the same time?

Purely talking about fixed-wing, I would say yes, as this can differ significantly but both have a specific meaning: Airspeed concerns aircraft behavior & safety (aircraft dynamics and stall speed), ground speed tells you how fast you can cover a certain distance / mission. With our fixed-wings, we're often flying at 10m/s airspeed, and have <5m/s or >15m/s ground speed, so the difference is very important there. 

IFF you think this is too much information for the general QGC user community, then we could indeed leave only one speed and the altitude as default (but leave the labels "AS", "GS", and "h") and put a small check box to enable the "extended" PFD with the second speed and the climb rate. I think that should cover all use cases. Let me know if that is necessary/better (and please also tell me where to add this check box exactly).

Best,
Philipp

DonLakeFlyer:
Thanks Phillip. Then I think if both are available show it as you have it now. Then I think the one change (if it doesn't already do this) is to not show it if NAN instead of "---".

Big thanks for the pull.

philipoe:
Done, and the label background is also slightly adapted such that it is only drawn when the variable (secondarySpeed) is actually valid. See below.

primaryflightdisplay2

DonLakeFlyer:
I'll double check TeamCity tonight for the run and then I'll merge.

philipoe:
@DonLakeFlyer Don't merge yet, I saw some inconsistencies (seem to not even be related to the PFD) here, will check that tomorrow or the day after that latest and let you know!

LorenzMeier:
Good to merge now?

philipoe:
Yes, you can merge!

PR #1363, opened 2015-03-15
Disable calibration buttons while calibrating

DonLakeFlyer:
Also changed to pull specific small parameter set after cal instead of full refresh. This way, the red/green indicators update much more quickly.


LorenzMeier:
Good to merge?

PR #1364, opened 2015-03-16
Line chart widget: Auto-force ground time if offset is more than 3 seconds

LorenzMeier:
This change enforces ground time if the system is not synced well to the GCS clock. I think this greatly improves usability. If we replace the checkboxes visuals to make them easier to find people should have less issues with the live plot. It also ticks the ground time checkbox, but it does not leave it ticked on the next boot unless the user ticked it manually.

schnappschuss 2015-03-16 10 12 15


LorenzMeier:
@DonLakeFlyer I don't see "pending" messages from TeamCity - is this expected?

DonLakeFlyer:
TC doesn't set pending status.

DonLakeFlyer:
TC apss. Good to merge?

PR #1368, opened 2015-03-18
Update Ubuntu PPA instructions

dagar:


DonLakeFlyer:
Thanks dagar.

dagar:
I'm working on getting a qgroundcontrol ubuntu ppa together, but it's surprisingly difficult and poorly documented. You can't just upload a deb because it has to be built by launchpad.

DonLakeFlyer:
That would be awesome. I think like I said before. If you can figure it out. Then let me know instructions and I might be able to automate TeamCity installers builds to do it for us on each build.

dogmaphobic:
There is always the option of downloading the Qt Online installer, which allows you to have an installation on your home directory and not mess up the system. I find that a lot simpler and cleaner as it doesn't pollute the (root) system: http://www.qt.io/download-open-source

DonLakeFlyer:
Gus: I'm a Linux newbie so I don't know what make sense or not. If you think this is a good option, could you update the readme to include that as well and why you would use one over the other?

dogmaphobic:
How do you *preview* changes to README.md?

DonLakeFlyer:
You hope for the best!

dogmaphobic:
Thanks! :smirk:

dagar:
If you edit it through the github web interface there's a preview tab.

dogmaphobic:
> If you edit it through the github web interface there's a preview tab.

Brilliant! Thanks...

PR #1372, opened 2015-03-18
Advanced Power Config

dogmaphobic:
Cleanup of Power Settings

Advanced power settings are now in their own area and no longer appear within the basic power settings area when enabled. While at it, I've added "Battery Capacity" to the advanced section.

screen shot 2015-03-18 at 3 21 06 pm

### DIFF
--- a/src/AutoPilotPlugins/PX4/ParameterFactMetaData.xml
+++ b/src/AutoPilotPlugins/PX4/ParameterFactMetaData.xml
@@ -30,6 +30,7 @@
             <short_desc>Battery capacity</short_desc>
             <long_desc>Defines the capacity of the attached battery.</long_desc>
             <default>-1.0</default>
+            <unit>mA</unit>
         </parameter>
         <parameter name="BAT_V_SCALE_IO" type="INT32">
             <short_desc>Scaling factor for battery voltage sensor on PX4IO</short_desc>
--- a/src/AutoPilotPlugins/PX4/PowerComponent.qml
+++ b/src/AutoPilotPlugins/PX4/PowerComponent.qml
@@ -273,7 +273,7 @@ Rectangle {
         }
         Rectangle {
             width: parent.width
-            height: 80
+            height: 40
             color: palette.windowShade
             visible: showAdvanced.checked
             Column {
@@ -288,16 +288,6 @@ Rectangle {
                         id: battDropField
                         width: textEditWidth
                         fact: Fact { name: "BAT_V_LOAD_DROP" }
-                        showUnits: true
-                    }
-                }
-                Row {
-                    spacing: 10
-                    QGCLabel { text: "Battery Capacity"; width: firstColumnWidth; anchors.baseline: battCapacityField.baseline}
-                    FactTextField {
-                        id: battCapacityField
-                        width: textEditWidth
-                        fact: Fact { name: "BAT_CAPACITY" }
                         showUnits: true
                     }
                 }
--- a/src/AutoPilotPlugins/PX4/PowerComponent.qml
+++ b/src/AutoPilotPlugins/PX4/PowerComponent.qml
@@ -44,7 +44,7 @@ Rectangle {
     color:  palette.window
 
     property int firstColumnWidth: 220
-    property int textEditWidth:    60
+    property int textEditWidth:    80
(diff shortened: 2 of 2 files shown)


Susurrus:
Any way to apply this Power Config to more than just the PX4?

dogmaphobic:
As long as it's part of the Fact System, anything can be added. What specifically do you have in mind?

Susurrus:
I guess my question was is this a QGC-only configuration? If so, can it be displayed for all autopilots so they can all use it versus just the PX4?

LorenzMeier:
Looks great!
@Susurrus The issue is that you would need to implement the same parameters plus since the autopilot codebases are constantly evolving you would need to deal with changes of those. Given this is all bound to the fact system and is QML, it might just be easier to create your own config plugin and steal the bits you're interested in from the PX4 one.

DonLakeFlyer:
> Given this is all bound to the fact system and is QML, it might just be easier to create your own config plugin and steal the bits you're interested in from the PX4 one.

Yup, what Lorenz said. Custom AutoPilotPlugin is the way to go and then steal what you want. Otherwise you're autopilot would need to masquerade as a px4fmu stack and duplicate the full parameter set. px4fmu specific plugin (src/AutoPilotPlugins/PX4/...) won't be able to handle a partial parameter set.

At some point in the future it would be nice to have a parameter redirector type of thing which would allow some of this to be even more generic. That may actually happen when we start to try to add back apm stack support but not yet. Requires quite a bit of more thought.

DonLakeFlyer:
@dogmaphobic Gus: Curious about battery capacity being part of advanced. That would seem like basic knowledge for anyone that does any sort of electric RC type stuff. That part itself doesn't seem very "Advanced". Is it not required to be set? If it's required and if you don't set it correctly thing go haywire then I would argue to not be advanced.

dogmaphobic:
Battery capacity is *basic* depending on the context. What is the value used for? Is the firmware accumulating power consumption and comparing that to the total capacity? How does it know if a battery was partially drained before it started (as in restarting the PX4)? For it to know that accurately, wouldn't it have to be able to know the voltage drop under load? If not, is it computing battery internal resistance? I have not looked into that part of the firmware enough to know these answers. Not to mention the fact that it would be a bit of a pain to have to come here every time you change a battery in the field if the sizes are different. With all that said, it would be no problem at all moving it up to "basic".

DonLakeFlyer:
@LorenzMeier Let's see what Lorenz says. I don't know how the battery values are used in the firmware.

LorenzMeier:
The battery capacity is not very useful. Users get it wrong constantly. Until we have smart batteries with an in-built ID and charge level chip voltage is the only robust estimate available.

DonLakeFlyer:
Maybe leave it out of here then? And just let it be in the full parameter set .

dogmaphobic:
> Maybe leave it out of here then? And just let it be in the full parameter set .

Makes sense. I will get it out.

While in this subject, why is there three instances of BAT_V_SCALING in ParameterFactMetaData.xml? How can you tell which one is which?

DonLakeFlyer:
Pasting in from conversation in separate thread:

It’s generated from the firmware source using a python code scraper. That scraper also builds this page: https://pixhawk.org/firmware/parameters. The multiple’s are due to ifdef’s in the code which the scraper doesn’t handle. For now I hand carry to file over from the firmware side. I have two things in the works:

1) I”ve modified the firmware build process to generate the file on each build an include it as part of the firmware file. Not connected end to end yet. Need to modify QGC firmware upgrade to pull the file out and use it.
2) I’m going to change the firmware build to build the c++ parameter code from the xml file instead of the other way around. This should lead to a higher quality xml file as well as make it easier to handle the ifdef type cases.

So for now, hand editing(as well as adding new parameters if needed) the xml file on the QGC side is fine. But if you do that you should also modify the C++ code on the firmware side as well. Otherwise once I finish #1 and #2 the missing stuff will come back. For duplicates you should look at the firmware source, find the parameter and find the ifdef for the PX4 board and use that one.

Now I'm off to another fun-filled day in court!

dogmaphobic:
> Makes sense. I will get it out.

Done. I think we are done with this one now.

Issue #1371, opened 2015-03-18
Problem with making RN-XV work with QGC


nimaaghli:
Hi,
I have configured RN-XV  with pixhawk the boud rate is 57600 ,
as soon as im connecting to port in udp i see the UAV in my list so i think im, getting MAV messages correct but after 2 second the program freezes and when i check the terminal im getting this message repeating
 UDP:Adding HOST 169.254.1.25:2000"
169.254.1.25 is the ip adress of RN-XV
Port 2000 is the port of RN-XV.

DonLakeFlyer:
Exact sequence of steps for me to reproduce this please, plus what OS? I'm not familiar with UDP so you will need to walk me through all of it.

dogmaphobic:
I'm adding UDP to my "slave" MavLink. I will be able to test this in the next couple of days...

nimaaghli:
Problem solved buy change UDPlink.cc and UDPlink.h

nimaaghli:
in void UDPLink::readBytes() inside the while its trying to add _config->addHost(host); each time and it is causing program to stop working i changed the code this way 
if (!_hosts.contains(sender.toString())){
            QString host(sender.toString() + ":" + QString("%1").arg((int)senderPort));
            _config->addHost(host);
            _hosts[sender.toString()] = senderPort;
        }
so it only adds the host to list once

dogmaphobic:
The fix in in PR #1374. Thanks for spotting it (and giving me the incentive to finish my own side project).

DonLakeFlyer:
@nimaaghli Could you double-check the latest releases or master build to make sure this is fixed for you now?

nimaaghli:
@DonLakeFlyer Yes i bulid the last releases and tested with single pixhawk it works great . Will test on multiple pixhawk tomorrow.

DonLakeFlyer:
Thanks @nimaaghli. I'm closing this now and you can re-open if you still find a problem.

PR #1374, opened 2015-03-19
UPD Link Fix

dogmaphobic:
Fixes to the UDP link.

This is in response to issue #1371. I got my MavLink *slave* done (the most convoluted pile of wires with a Raspberry PI in the middle of it all) and was able to finally test UDP links. In the process, I found not only the issue mentioned above but also a couple of thread racing conditions. UPD links are functional but as I only have one MavLink device, I have not tested it with multiple connections.

### DIFF
--- a/src/comm/UDPLink.cc
+++ b/src/comm/UDPLink.cc
@@ -94,7 +94,6 @@ QString UDPLink::getName() const
 
 void UDPLink::addHost(const QString& host)
 {
-    qDebug() << "UDP:" << "ADDING HOST:" << host;
     _config->addHost(host);
 }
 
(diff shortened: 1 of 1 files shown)


src/comm/UDPLink.cc
```diff
@@ -317,32 +312,43 @@ void UDPConfiguration::addHost(const QString& host)
             for (int i = 0; i < hostAddresses.size(); i++)
             {
                 // Exclude loopback IPv4 and all IPv6 addresses
-                if (!hostAddresses.at(i).toString().contains(":"))
+                if (!hostAddresses.at(i).toString().contains(":") && !!hostAddresses.at(i).toString().startsWith("127"))
```
LorenzMeier: I'm not too sure about this one. We have quite often running the autopilot code and the GCS on the same machine. Could you think of any reason to disallow localhost here?

LorenzMeier:
Looks good, except for the line excluding localhost.

dogmaphobic:
> Could you think of any reason to disallow localhost here?

Not really. The only reason I did that was because of the original comment in the code ( which I ended up removing in the process).

dogmaphobic:
Loopback re-enabled.

PR #1376, opened 2015-03-19
Cleaning up Tile Cache Queue

dogmaphobic:
Cleaning up the Tile Cache Queue code based on an issue brought up in the px4users  list:

https://groups.google.com/forum/#!topic/px4users/i1vAFNO_KLk

### DIFF
--- a/libs/opmapcontrol/src/core/tilecachequeue.cpp
+++ b/libs/opmapcontrol/src/core/tilecachequeue.cpp
@@ -28,45 +28,39 @@
 
 
 //#define DEBUG_TILECACHEQUEUE
- 
+
+#ifndef DEBUG_TILECACHEQUEUE
+#undef qDebug
+#define qDebug QT_NO_QDEBUG_MACRO
+#endif
+
 namespace core {
 TileCacheQueue::TileCacheQueue()
 {
 
 }
+
 TileCacheQueue::~TileCacheQueue()
 {
-   // QThread::wait(10000);
+    quit();
+    _waitc.wakeAll();
+    wait();
+    this->deleteLater();
 }
 
 void TileCacheQueue::EnqueueCacheTask(CacheItemQueue *task)
 {
-#ifdef DEBUG_TILECACHEQUEUE
     qDebug()<<"DB Do I EnqueueCacheTask"<<task->GetPosition().X()<<","<<task->GetPosition().Y();
-#endif //DEBUG_TILECACHEQUEUE
-    if(!tileCacheQueue.contains(task))
-    {
-#ifdef DEBUG_TILECACHEQUEUE
+    if(!_tileCacheQueue.contains(task)) {
         qDebug()<<"EnqueueCacheTask"<<task->GetPosition().X()<<","<<task->GetPosition().Y();
-#endif //DEBUG_TILECACHEQUEUE
-        mutex.lock();
-        tileCacheQueue.enqueue(task);
-        mutex.unlock();
-        if(this->isRunning())
-        {
-#ifdef DEBUG_TILECACHEQUEUE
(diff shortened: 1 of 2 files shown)


LorenzMeier:
What is the testing level of this? Do you feel confident about the change or want cross-testing?

DonLakeFlyer:
@dogmaphobic You should brush up on QLoggingCategory and in QGC the QGCLoggingCategory overrides. I've been trying not to add new usage of qDebug, but instead to logging categories. This allows you to turn on/off logging from an ini file instead of having to recompile.

DonLakeFlyer:
If Gus says this is good to go then this will be the last thing through into the Stable RC.

dogmaphobic:
> You should brush up on QLoggingCategory and in QGC

... All I wanted was to get of the zillion #ifdefs spread all over the file. As far as I can tell, all those qDebug() calls could go away.

> What is the testing level of this? 

There isn't much to it. There is only one client and one consumer. I run a few times caching a bunch of tiles but that's all I did. Speaking of cache file, it would be a good idea to consolidate the locations where all these files are saved. The tile cache database is stored in its own directory in you home folder, which is not what the rest of the program does. I didn't want to go that deep into the code today though.

dogmaphobic:
> qDebug() calls could go away.

Meaning, I didn't add any of those. They were already there and each one was surrounded by a #ifdef/#endif pair.

DonLakeFlyer:
I'd say good to go then. 

The tile cache does use one of the Qt specific locations using those APIs? Can you point me to that so I can take a look?

DonLakeFlyer:
> Meaning, I didn't add any of those. They were already there and each one was surrounded by a #ifdef/#endif pair.

Ick :)

dogmaphobic:
> The tile cache does use one of the Qt specific locations using those APIs? Can you point me to that so I can take a look?

No it doesn't. It just plops it in you home directory. Note that the Qt specific locations don't work on Android as I found out :) The "home" dir is actually a good place to put it. Just not as a visible directory.

It's defined here:

https://github.com/mavlink/qgroundcontrol/blob/master/libs/opmapcontrol/src/core/cache.cpp#L56

dogmaphobic:
> QGCLoggingCategory overrides

Speaking of Android, not that it's important now, the overrides don't work there. You don't get any output. The regular qDebug() ones do work.

DonLakeFlyer:
> the overrides don't work there.

Not sure what you mean by that?

dogmaphobic:
> Not sure what you mean by that?

Meaning that when you use these types:

```c++
qCDebug(UASLog) << "removeLink:" << QString("%1").arg((ulong)object, 0, 16);
qCDebug(UASLog) << "link count:" << links.count();
```

It produces no output on Android. The regular ```qDebug()``` works fine though.

DonLakeFlyer:
And you turned on UASLog by setting UASLog=true in the qtlogging.ini file? Maybe a location problem with the ini file? Hmm.

dogmaphobic:
> And you turned on UASLog by setting UASLog=true in the qtlogging.ini file? Maybe a location problem with the ini file? Hmm.

Ha! That explains it. Yes, the default location for the ini file is read only on Android (a Qt problem). That's what I meant earlier to sticking to the *home* directory for everything.

DonLakeFlyer:
The Qt APIs specifically state that the location is supposed to be writable!

dogmaphobic:
> The Qt APIs specifically state that the location is supposed to be writable!

Sure, and politicians keep telling us they are honest :)

What I noticed was that there is little "standards" when it comes to Android. Each vendor decides to do whatever they want. The *home* directory at least seems to be always writable. With that said, my Android experience is only one month old!

Issue #1366, opened 2015-03-17
Can't add most message values to the Info View's Quick tab


Susurrus:
This used to be possible, I think maybe adding the Fact system broke this? @DonLakeFlyer Can you confirm this?

I'd really like to show for example: crosstrack error, roll, and yaw and none of these are options. I know it *should* be possible.

DonLakeFlyer:
No changes there that I know of.

Susurrus:
I used to get a huge list of values, now I only see 2 different messages. Does this work for you?

DonLakeFlyer:
Nope. List should come from mavlinkdecoder.

dogmaphobic:
Is this what you're talking about?

![screen shot 2015-03-16 at 11 21 37 pm](https://cloud.githubusercontent.com/assets/749243/6680850/534253e0-cc33-11e4-8558-25cef896f2c1.png)

DonLakeFlyer:
Yup. It should have values in there for each thing that mavlnkdecoder decodes and signals a value change on.

Susurrus:
Exactly. I only see the HEARTBEAT and SYS_STATUS messages. I know this used to work.

DonLakeFlyer:
Don't have GitHub right now. Can you mark this for stable v2.3? So it gets fixed before next stable

dogmaphobic:
It's all there now (#1377)

DonLakeFlyer:
Fixed

PR #1377, opened 2015-03-19
Fixing list of items in Quick Info View

dogmaphobic:
This fixes issue #1366.

Not sure when this changed as it seems it predates me but the Quick Info View was missing a call to let it know where to get data from.

Now you get the full gory, unwieldy list of options to choose from (meaning, we really need to rethink this dialog):

screen shot 2015-03-19 at 3 10 55 pm


Susurrus:
Looks good to me, though I won't be able to test it til next week. It passed TC, so if it's working for you, let's merge it.

PR #1378, opened 2015-03-19
Switch to Stable V2.3 Release Candidates

DonLakeFlyer:
This switches master to kicking out new release candidates. Then we can pump out daily RCs using TC for a day or so that people can try. Once we are comfortable I'll create the Stable V2.3 branch in main. And that will be the official branch.


DonLakeFlyer:
Just FYI: I'm going to merge this as soon as it passes TC so I can start taking some screen shots for an RC email to the google group.

PR #1382, opened 2015-03-22
Use QSharedPointer for cross-thread Link references

DonLakeFlyer:
Also many LinkManager API changes to further isolate all link management inside LinkManager.

This fixes the shutdown issues where MAVLinkProtocol was still holding onto a deleted link. The issue is that you can't keep non-reference counted pointers to LinkInterface objects on another thread. If you do you run into all sorts of cross-thread signaling issues during shutdown. Which in turn leads the non-ui thread the reference a LinkInterface* after it has been deleted.

This change uses QSharedPointers for LinkInterface* objects kept in LinkManager. It also uses copies of these same QSharedPointers in MAVLinkProtocol and UAS we run on different threads. This way they are correctly reference counted and will shutdown in an orderly fashion when they are disconnected.

Another important point to note is that in general LinkManager is not needed to be thread-safe with respect to creating new links. The only thing in LinkManager that is thread-safe is _addLink and _deleteLink. They still need to be thread-safe due to some hack usage with MAVLinkSimulationLink which I still haven't fully gotten rid of.

So now when you disconnect a link, it is also automatically deleted without the need for the delayed delete thing.

@dogmaphobic Can you give this a thorough code review as well as some testing. I've hammered on it and it looks good. But needs extra eyes.

### DIFF
--- a/src/comm/LinkManager.h
+++ b/src/comm/LinkManager.h
@@ -103,21 +103,18 @@ public:
     /// Sets the flag to allow new connections to be made
     void setConnectionsAllowed(void) { _connectionsSuspended = false; }
 
-    /// Creates (and adds) a link  based on the given configuration instance. LinkManager takes ownership of this object. To delete
-    /// it, call LinkManager::deleteLink.
-    LinkInterface* createLink(LinkConfiguration* config);
+    /// Creates, connects (and adds) a link  based on the given configuration instance.
+    LinkInterface* createConnectedLink(LinkConfiguration* config);
 
-    /// Creates (and adds) a link  based on the given configuration name. LinkManager takes ownership of this object. To delete
-    /// it, call LinkManager::deleteLink.
-    LinkInterface* createLink(const QString& name);
+    /// Creates, connects (and adds) a link  based on the given configuration name.
+    LinkInterface* createConnectedLink(const QString& name);
 
-    /// Adds the link to the LinkManager. LinkManager takes ownership of this object. To delete
-    /// it, call LinkManager::deleteLink.
-    void addLink(LinkInterface* link);
-
-    /// Deletes the specified link. Will disconnect if connected.
-    // TODO Will also crash if called. MAVLink protocol is not handling the disconnect properly.
-    void deleteLink(LinkInterface* link);
+    /// Returns true if the link manager is holding this link
+    bool containsLink(LinkInterface* link);
+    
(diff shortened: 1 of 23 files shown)


DonLakeFlyer:
Also FYI: This is my first step towards finally fixing the whole disconnect leaving zombied UAS's hanging around problem. As well as the disconnect, throws a Connection Lost error problem.

dogmaphobic:
> Also FYI: This is my first step towards finally fixing the whole disconnect leaving zombied UAS's hanging around problem. As well as the disconnect, throws a Connection Lost error problem.

I figured that much :) I will take a look at it tomorrow. Thanks!

src/comm/LinkManager.h
```diff
@@ -130,6 +127,14 @@ class LinkManager : public QGCSingleton
...
     /// Disconnect the specified link
     bool disconnectLink(LinkInterface* link);
+    
+    /// Returns true if there are any connected links
+    bool anyConnectedLinks(void);
+    
+    // The following APIs are public but should not be called in normal use. The are mainly exposed
+    // here for unit test code.
+    void _deleteLink(LinkInterface* link);
+    void _addLink(LinkInterface* link);
```
dogmaphobic: Couldn't you leave these private and make the test classes "friend"?

src/ui/QGCToolBar.cc
```diff
@@ -676,10 +676,8 @@ void QGCToolBar::_connectButtonClicked(bool checked)
             // Get the configuration name
             QString confName = _linkCombo->currentText();
             // Create a link for it
-            LinkInterface* link = _linkMgr->createLink(confName);
+            LinkInterface* link = _linkMgr->createConnectedLink(confName);
             if(link) {
-                // Connect it
-                _linkMgr->connectLink(link);
                 // Save last used connection
```
dogmaphobic: This class (old Tool Bar) is not being used. I left the code in for the time being while making sure the new tool bar was fully functional. I will be removing it soon.

dogmaphobic:
At first glance it all looks fine. Note that the serial link code is quite weird with its thread handling. I have in my to do list to remove all that and use QSerialPort signals instead (like both UDP and TCP links).

DonLakeFlyer:
> Note that the serial link code is quite weird with its thread handling.

Yup. It's all a waste of time.

Issue #1386, opened 2015-03-24
GUI changes while changing MAV in multi drone 


nimaaghli:
Hello, 
While having multi drone and clicking on them to control the GUI has some changes on menu the arrow shapes dissapear and show up again when mouse goes over them

nimaaghli:
![screenshot from 2015-03-24 11 59 52](https://cloud.githubusercontent.com/assets/3833065/6806420/fd4c986e-d21e-11e4-9f5d-39183834adc0.png)
![screenshot from 2015-03-24 12 01 22](https://cloud.githubusercontent.com/assets/3833065/6806428/035c4c2c-d21f-11e4-80be-016c2ec13ac4.png)

DonLakeFlyer:
@dogmaphobic Something for you Gus

DonLakeFlyer:
@dogmaphobic I was just talking about the toolbar button redraw problems for you to look at. Can you pull that out into a separate issue?

dogmaphobic:
Duh... This entire time I was assuming it was an issue within the map view. It didn't even occur to me to look at the tool bar and see the chevron buttons are not there. It's the OpenGL Canvas issue (QtQuick inside a QWidget). I know how to get around it, I just need to be able to reproduce it. I will wait for your #1392 to get merged and I will take a look.

PR #1392, opened 2015-03-26
Disconnect really

DonLakeFlyer:
When last link to UAS is disconnected the UAS is deleted as well. This removes the infamous zombie UAS case where a disconnected UAS would cause the UI do to all sorts of unexpected things.

- Also restructured the pluginReady status for plugins
- Cleaned up places which were handling the UAS going away incorrectly


DonLakeFlyer:
@dogmaphobic If you use this pull and Connect to PX4 and then Disconnect you'll get the same button draw problem as Issue #1386

PR #1393, opened 2015-03-26
Hack in QTLocation for map

DonLakeFlyer:
This is just for Gus to look at.


dogmaphobic:
You can close this when you want.

PR #1394, opened 2015-03-27
Fix redraw issues with OpenGL canvases when link states change.

dogmaphobic:
This should fix issue #1386 and the newly-brought-to-life issue when a link is (truly) disconnected (PR #1392). A link state change also causes the UI to fully redraw. Whenever there is a full redraw, the canvases used within QML elements must be forcefully redrawn as they don't get the hint on their own...


DonLakeFlyer:
This is OK for a hack. But we're going to need to figure out the real reason these don't redraw correctly. Possibly even submit Qt fix for it. We'll likely hit the same problem again as we convert UI to Qml in a piecemeal fashion.

Issue #1402, opened 2015-03-30
QGC segfaults on disconnect: Ubuntu 14.10


kd0aij:
latest master
```
SerialLinkLog: Closing Port # 194 "ttyACM0"
MAVLinkProtocolLog: _linkStatusChanged "44a2030" false
UASLog: _linkDisconnected: "/dev/ttyACM0"
UASLog: link count: 1
SerialLinkLog: Already disconnected
Deleting UAS object:  "MAV 001"
Segmentation fault (core dumped)

#0  0x00007f3754347370 in ?? () from /opt/qt54/qml/QtQuick/Controls/libqtquickcontrolsplugin.so
#1  0x00007f378b5ba8eb in QObject::disconnect(QObject const*, char const*, QObject const*, char const*) () from /opt/qt54/lib/libQt5Core.so.5
#2  0x000000000060d50f in DebugConsole::linkSelected(int) ()
#3  0x00000000007a2dcb in DebugConsole::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) ()
#4  0x00007f378b5b81a1 in QMetaObject::activate(QObject*, int, int, void**) ()
   from /opt/qt54/lib/libQt5Core.so.5
#5  0x00007f378cba4951 in QComboBox::currentIndexChanged(int) ()
   from /opt/qt54/lib/libQt5Widgets.so.5
#6  0x00007f378cba6b61 in ?? () from /opt/qt54/lib/libQt5Widgets.so.5
#7  0x00007f378cbab601 in ?? () from /opt/qt54/lib/libQt5Widgets.so.5
#8  0x00007f378b5b81a1 in QMetaObject::activate(QObject*, int, int, void**) ()
   from /opt/qt54/lib/libQt5Core.so.5
#9  0x00007f378b6334e4 in QAbstractItemModel::rowsRemoved(QModelIndex const&, int, int, QAbstractItemModel::QPrivateSignal) () from /opt/qt54/lib/libQt5Core.so.5
#10 0x00007f378b536592 in QAbstractItemModel::endRemoveRows() ()
   from /opt/qt54/lib/libQt5Core.so.5
#11 0x00007f378c31caf2 in QStandardItem::removeRows(int, int) ()
   from /opt/qt54/lib/libQt5Gui.so.5
#12 0x00007f378c31cbd4 in QStandardItemModel::removeRows(int, int, QModelIndex const&) ()
   from /opt/qt54/lib/libQt5Gui.so.5
[...]

Issue #1399, opened 2015-03-28
Crash in DebugConsole


dogmaphobic:
Putting it down here before I forget.

To reproduce:

Start QGC and then plug in PX4 (USB). Click connect while the PX4 is booting. Nothing will happen as QGC is waiting for the bootloader to go away. Hit disconnect and then connect again. DebugConsole crashes in ```DebugConsole::updateTrafficMeasurements()``` as it tries to use a link pointer that is no longer valid.

That whole DebugConsole may need some rethinking as the UI is all broken and its functionality is dubious at best.

LorenzMeier:
We need a text terminal for the system with basic VT100 capabilities. It's the main interface during development. The current implementation is indeed messed up, but it would be nice to get a proper console setup going.

dogmaphobic:
I’m not dismissing the need, rather, the implementation. Not to mention the redundancy with the Terminal widget. It’s on my todo list...

g

DonLakeFlyer:
We need one terminal to rule them all!

DonLakeFlyer:
This is fixed now

PR #1403, opened 2015-03-30
Track links using shared pointed

DonLakeFlyer:
Since debug console maintains a list of links it needs to use shared pointers so it doesn’t run into the link delete ordering problem. This should be the fix for Issue #1402


dogmaphobic:
Isn't this the same as #1399?

DonLakeFlyer:
Yes it looks like it.

Issue #1157, opened 2014-12-29
Re-enable Google Earth on Windows 32 bit builds


DonLakeFlyer:
Google Earth is currently disabled to work around a problem with the view in combination with QML Setup View. Debugging this requires a debug QT with WebKit which is very difficult to get to build.

DonLakeFlyer:
@LorenzMeier This has been turned off for a while. I'd like to toss it completely instead of trying to resurrect it.

Susurrus:
There haven't been any complaints about it being gone and there are plenty of other priorities to focus on, so I'd second that notion.

DonLakeFlyer:
I'm going to toss this. Reason: I'm going to cut over the Windows build to static qt libs. It's the only way to reasonably debug crash logs. Trying to build webkit in Qt on windows for Google Earth is next to impossible to get set up.

PR #1404, opened 2015-03-30
Remove GoogleEarth support

DonLakeFlyer:
See Issue #1157


Susurrus:
Does the README need an update about this as well?

DonLakeFlyer:
Yup thanks.

PR #1405, opened 2015-03-30
Debug console

DonLakeFlyer:
Sorry I was trying to get away with testing on OS X and not have to crank up Ubuntu. The problem was the link deletion ordering problem. But in this case there were two part that needed fixing.


DonLakeFlyer:
This should get #1402 for real this time.

PR #1406, opened 2015-03-31
Terminate audio thread if it won't shutdown nicely

DonLakeFlyer:
Attempting fix for Issue #1317


DonLakeFlyer:
@kd0aij @Susurrus Hey guys can you give this a try?

DonLakeFlyer:
This works in that it doesn't hang but it causes a crash instead later on.

PR #1408, opened 2015-03-31
Make it able to be compiled on a 32-bit windows machine

oberion:
I had a problem of building the QGC on windows visual studio 2012 on a 32-bit machine. When trying to generate the resource object file, the machine was out of memory. 
I have split the resource file into several smaller ones to be able to compile again.


DonLakeFlyer:
How much memory do you have on your machine and how much free disk space?

DonLakeFlyer:
The reason I ask is that I really don't like this change. It makes usage of the resource file confusing. Seems like a hack to fix a problem that isn't quite understood yet. Can you provide build output where the error occurs?

dogmaphobic:
A good side effect of this is that it would speed up builds for development. In theory at least, only the resources you are working with would get rebuilt. With that said, we still need to do a clean up pass on these resources. There are quite a bunch of them not used and just adding bloat.

DonLakeFlyer:
Maybe that's the better answer. Do the cleanup pass which should greatly reduce the resources. I can do that. 

@oberian I'm going to do a pass through the resources which should make them way smaller and most likely fix your problem as well.

dogmaphobic:
I started writing a python script and soon realized it would take me a lot longer to write it than do it by hand. But then, I found something more fun to do and never went back to it :) I guess the easiest thing to do would be to remove all files, run a build, see what fails and pull them back one by one.

Realize that the resulting .cpp files with the resources are rather humongous:

```
-rw-r--r--  1 gus  staff   1.6M Mar 31 12:10 qrc_mapresources.cpp
-rw-r--r--  1 gus  staff    41M Mar 31 23:26 qrc_qgroundcontrol.cpp
```

DonLakeFlyer:
@oberion I believe Pull #1414 should have fixed this problem. So closing this pull. Can you let us know?

PR #1412, opened 2015-04-01
Font's are now correctly and equally size across all OS

DonLakeFlyer:
- Added new dpiAdjustedDefaultFontPointSize to QGCPallete for use in controls
- All Qml controls default to the above point size
- Widget style sheets font sizes are dpi adjusted at boot time
- Fixed some incorrect pixel font size usage

The above changes should limit the need for specifying your own font size. You only need to do that when you want something other than default. In that case you still must always use ScreenTools::dpiAdjustPointSize in your Qml.

### DIFF
--- a/files/styles/style-dark.css
+++ b/files/styles/style-dark.css
@@ -7,7 +7,7 @@
 * {
     background-color: #222;
     color: #FFF;
-    font-size: 11px;
+    font-size: 12pt;
 }
 
 QWidget#viewModeWidget {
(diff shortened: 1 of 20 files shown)


DonLakeFlyer:
@dogmaphobic Gus can you take a look at this to make sure it all makes sense?

dogmaphobic:
It all looks fine. When I first created ScreenTools, I was trying to create it as a singleton as I knew there was a way to register a singleton type for QML. I couldn't get it working and didn't spend the time to find out why. What you did with the static methods solved that :)

DonLakeFlyer:
Hmm, singleton that would be a good idea. Not sure how to do that. I just kind of hacked what I did in. Not pretty but it works.

PR #1414, opened 2015-04-02
Resource cleanup

DonLakeFlyer:
- All resources are either under src or under resources dirs
- Removed files top level directory
- Move flightgear files to top level dir
- Removed unused resources


DonLakeFlyer:
@oberion Hoping that this will fix the out of heap space problem as well.

DonLakeFlyer:
Also files directory is no longer copied to installer images, only flightier directory. This should cut down on installer size quite a bit.

dogmaphobic:
Cool! It was taking forever to cycle through builds because of the time it took to build and compile these resources.

PR #1416, opened 2015-04-04
Mag Cal UI

DonLakeFlyer:
Mag cal now has UI like the rest of sensor cal. Visuals are just like accel cal, except it says to Rotate instead of Hold Still when an orientation is detected.

Coming soon: Visuals on how you need to rotate.


LorenzMeier:
Awesome! I merged the Firmware side, so we can merge this once TC is happy again.

PR #1420, opened 2015-04-05
Status bar

dogmaphobic:
I renamed the *Perspectives* menu to a more standard *View* menu. In addition, I've added an option to show and hide the Status Bar, giving more screen real estate when you don't need the log playback.

screen shot 2015-04-04 at 11 00 41 pm


LorenzMeier:
Are you defaulting the status bar to off? I would assume that defaulting it to off would suit most users best.

dogmaphobic:
> Are you defaulting the status bar to off?

In theory yes, though the commit that fixes that is not showing up here. It made it to my upstream branch but it did not get reflected in this PR.

https://github.com/dogmaphobic/qgroundcontrol/commit/ccb8bf62821bc4aaeb219609d6062ef107354ab4

PR #1429, opened 2015-04-08
Restructure location plugin build

DonLakeFlyer:
Restructured to better fit existing build structure:
- qgroundcontrol.pro is still main .pro file
- QGCApplication.pro builds the app
- QGCLocationPlugin.pro builds the plugin
- QGCCommon.pri is common to both the app and the plugin. This allows build setup and compiler settings to be common across both.
- Moved plugin source to src/QtLocationPlugin to keep libs directory 3rd party only
- Plugin places built artifacts (including lib) into same build-debug and build-release directories as application does
- Removed internal Qt build system usage from plugin .pro. These were the load(...) calls which are only for building inside of Qt itself.
- Cleaned up plugin import/export mechanism to match non Qt internal usage


DonLakeFlyer:
There are a couple things left to do:
- I think I can get rid of .qmake.conf now
- make clean does not clean up lib

Wanted to get a pass through TC first since it will require TC changes as well to synch with this.

dogmaphobic:
> I think I can get rid of .qmake.conf now

Yes, that's used by the *load(...)* call you reference above. If you changed it so it can be built without the Qt internal setup, that's no longer needed. That's just to make sure we are using the (private) header files from the right version of Qt.

DonLakeFlyer:
The Qt plugin docs are a nightmare of lack of or incorrect/old information.

dogmaphobic:
> The Qt plugin docs are a nightmare of lack of or incorrect/old information.

You actually found documentation? :) I just went through all the existing code and derived it from what they did.

DonLakeFlyer:
TC updated and passes. Merge now and need to very installer builds.

dogmaphobic:
Hmm... Something got messed up. Mapping is no longer working. Investigating...

dogmaphobic:
I have to leave and I run out of time to fully debug it. I will take a look at it tomorrow. When you use the mouse wheel to zoom in and out, the map goes blank. If you keep zooming (in or out), sometimes the maps are fetched other times they aren't (and the map window stays blank). This was all working though I can't think what could cause that. The plugin is being properly created. It's just that it is not being called when it should.

DonLakeFlyer:
I tested on all OS and it was working. Likely the last change to remove $$qtlibrarytarget screwed it up. I'll take a look. Probably not till tomorrow.

dogmaphobic:
I don't think so. That would more likely just cause a link time issue, not a runtime issue. In fact, nothing you changed jumps out as something that would change the code produced. The only suspect is the removing that call function from within the project file and I only say that because it was a black box. That is, I don't know all that call did (it was way too many nested calls to track down). In addition, I didn't have a lot of time to do any sort of extensive testing. I just built it and run it. That's when I noticed the strange behavior when zooming in and out. No worries, I'll take a look tomorrow...

DonLakeFlyer:
Had to be one of the last things I did because it had it working fine before those last changes.

DonLakeFlyer:
Strange, it's working fine for me on OS X.

DonLakeFlyer:
Wait a minute. I guess the zoom is a bit flaky. Sometimes it works, sometimes it hangs.

DonLakeFlyer:
Would be good to have some sort of unit test for this stuff.

PR #1431, opened 2015-04-08
Support LOCAL_OFFSET_NED waypoints.

Susurrus:
Useful for when testing and you want to do a specific set of waypoints but aren't sure where you will be able to test them. By using the vehicle's position as the reference point, this is really easy.


Susurrus:
There's a few other areas that might need to be modified, but I don't know, for example, what the `src/ui/WaypointViewOnlyView.cc` is.

Susurrus:
And this works in all of my testing using LOCAL_OFFSET_NED waypoints with external .waypoints files as well as with communicating to the UAS.

DonLakeFlyer:
Haven't dug into this code. By just adding the combo item it automatically selects the new combo item when loading waypoints?

Susurrus:
Loading waypoints from a file does work correctly, yes. Additionally, creating waypoints with the UI by hand works as well. Either of those can be sent to the UAS as well. So from that aspect, yes, this all works.

DonLakeFlyer:
Got it thanks.

Susurrus:
Thanks for merging.

Please also see #1432, which arises from how QGC supports local coordinate frames.

PR #1439, opened 2015-04-09
Fixed thread race condition.

dogmaphobic:
It was introduced by the Open Pilot URL Factory code I added the other day.

### DIFF
--- a/src/QtLocationPlugin/OpenPilotMaps.cc
+++ b/src/QtLocationPlugin/OpenPilotMaps.cc
@@ -37,7 +37,7 @@ This file is part of the QGROUNDCONTROL project
 
 namespace OpenPilot {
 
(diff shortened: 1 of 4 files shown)


Nyaruko:
Hi, are you willing give a hint where is the race codition in the original
code? Will it also affect the original openpilot map widget? I have met
some crash when accessing the online map.

dogmaphobic:
I doubt the problem was in the original code. I basically took the *gist* of what it was doing and re-wrote for my own purpose. It's a completely different code. The original Open Pilot mapping works quite differently in all aspects. The way to reproduce the problem would be to open a large screen and zoom in and out repeatedly. The issue I run into was due to Qt core location engine threading, which is not the case when using Open Pilot's mapping.

PR #1445, opened 2015-04-11
Map Display Tweaks

dogmaphobic:
Disabled map follows heading for the time being. That requires a much larger map (to be able to rotate on its center) and it was causing problems.
Added a scale and a zoom slider.
Changed the minimum zoom level reported by the plugin to a more reasonable value.

screen shot 2015-04-11 at 12 59 54 am


dogmaphobic:
For posterity's sake, Qt Mapping has issues when the vertical resolution is 1536 or over (0x600 or 2 x 256, which is the size of each map tile). The problem manifests when the zoom level is between x.95 and x.20. That is, if the zoom level is between say 8.95 and 9.20 or 9.95 and 10.20, the maps are not rendered (the screen is left blank). If the vertical size is less than 1536, everything works fine.

The issue is related to the number of tiles. If I set the tile size to 128x128, the problem shows up with a smaller height (though around 900 pixels and not 768 as logic would dictate), and if I set it to 512x512 it doesn't show up at all (I can only go as high as 2560 on my monitor).

Issue #1421, opened 2015-04-05
Turn QT_FATAL_WARNINGS back on for OSX


DonLakeFlyer:
Turned off due to this: http://www.qtcentre.org/threads/62133-QNetworkAccessManager-get-generates-SSL-qWarning-output-when-no-SSL-is-used?p=275327#post275327

DonLakeFlyer:
From Qt 5.4.1 onwards you can disable the warning by e.g. setting
    QT_LOGGING_RULES=qt.network.ssl.warning=false
in the environment, or calling
    QLoggingCategory::setFilterRules("qt.network.ssl.warning=false");
in your main().

So need to wait or a bump to 5.4.1 to turn back on.

DonLakeFlyer:
The above doesn't work.

dagar:
travis-ci currently has QT_FATAL_WARNINGS=1 for osx debug builds

DonLakeFlyer:
Thanks Daniel

PR #1449, opened 2015-04-13
Turn off Qt SSL qWarning

DonLakeFlyer:
Fixed Issue #1421


DonLakeFlyer:
This isn't working. Turning QT_FATAL_WARNINGS back off for OSX.

PR #1450, opened 2015-04-14
Experimental mapping/planning tool.

dogmaphobic:
I'm adding this now because I've accumulated too many changes and I don't want to stay too far behind. In addition, I want to work on some issues with the Flight View, which could not be done concurrently (meaning, while this was not committed).

This is the very basic, initial implementation of the new QtQuick Map/Planning tool. Access is provided through a menu entry in the Advanced Menu. The original Map tool is unchanged.

This will also allow @DonLakeFlyer to take a look at it and get acquainted with it as we will most likely be working together on it.

Of notice:

- Refactored several function and variable names to reflect the current terminology used for the various *Views*. Things like *Engineering View* is now properly named *Analyze View* as this is how it is shown to the user.

- Moved the QML Map Tool (along with several other QML controls), into a common directory, which will be used by the various *Primary* views.

- Removed the default margin in the main view, which was eating up valuable screen real estate.


DonLakeFlyer:
TC is a bit wonky now. I'm fighting with QT_FATAL_WARNINGS. If you see aborts from SSL stuff in unit tests, or aborts that say "could not connect to google maps" then I'm working on it and they are not your problem.

dogmaphobic:
It's never my problem. I'm perfect after all 8-P

No worries :)

PR #1451, opened 2015-04-14
Parameter Editor in Qml

DonLakeFlyer:
Functional differences:
- All parameter changes are immediately written to ROM. At this point the FactSystem can only support working this way. There is no option to do it the old way as well. If there is enough grief about this I can modify the FactSystem to support both writing modes but at this point I'd rather not since it will get somewhat complicated.

FactSystem Changes:
- Facts now have the concept of "Group". Multiple Facts can belong to the same group. You can see how groups work in the new editor. The group is part of the Fact meta data.
- Fact binding using the Qml Fact control has changed for non-default component ids. If you need to bind to a parameter which is not in the default component then you must construct the name like this: "PARAM_NAME:id", where id is the component id.

Here is what the new editor looks like. It uses more of the FactSystem meta data to provide better usability. There is much more meta data available for things like range checking and even doing things like using check boxes or combo boxes for editing appropriate parameters. But until I clean up the firmware side of things this is the best I can do for now. You can see already how the meta data is not very good in the limited usage it is getting here.
screen shot 2015-04-13 at 10 10 26 pm


DonLakeFlyer:
@LorenzMeier If you run this you can see the somewhat recent state of parameter meta data which I pulled across manually from the firmware. I would be great if the people responsible for these parameters could at a minimum clean up the "units" and "short description" fields which is what I am using here. 

Then I will complete the following work:
1) Parameter meta data comes packaged in firmware file and QGC is gets updated meta data each time you flash. I already modified firmware to put the .xml file in the .px4 file but it's not quite setup right for usage in QGC. Needs more firmware change and then QGC changes to use it.
2) Change firmware such that the paramater.xml file is the "truth" and C++ code gets generated from it. This will lead for much cleaner meta data. We discussed this a while back.

LorenzMeier:
@tumbili Please cross test

PR #1455, opened 2015-04-14
Qml based dock widget support

DonLakeFlyer:
Lots of stuff in this pull:
- New support for Qml Based dock widgets by using the ViewWidget control as a base control for the dock widget Qml with is then hosted in QGCQmlWidgetHolder. The ViewWidget implementation automatically handles autopilot connect/disconnect thus fixing the zombie disconnected uas problem for new dock widgets. When disconnected they automatically show "no vehicle connected". You can override this with your own qml if you like. You specify your own Qml for the connected state. See ParameterEditorWidget for example usage. This is not the Qml tool widget on top of qml only view implementation, but it is a step towards that.
- Restructured the dock widget functionality such that dock widget save state no longer is tied to specific UAS or whether the UAS is connected or not. This caused all sorts of confusing behavior of widgets coming/going at unexpected times. Now each view saves the states of what widgets are shown and that is all.
- Removing a widget by clicking the X in the titlebar now works correctly.
- Floating widgets are no longer supported. Widgets must remain within the confines of the main window. Thus is in prep for tablet support, where the concept of real estate beyond the confines of the main window does not exist.
- Converted ParameterEditor to a Qml control. It supports two modes: fullMode=true shows all functionality, fullMode=false shows limited functionality: no load/save, no short descriptions.
- Converted the Onboard Parameters to Qml using the new ParameterEditor control in fullMode=false mode.
- Removed the stream rate setting capability which was part of the old Onboard Parameters widget


DonLakeFlyer:
Forgot to say that I had to bump the settings version number due to all the changes.

LorenzMeier:
This sounds awesome!

DonLakeFlyer:
New widget showing ParameterEditor running in fullMode=false:
screen shot 2015-04-14 at 2 52 42 pm

Issue #1456, opened 2015-04-15
Qml Error / module Qt Location is not installed


severinleuenberger:
I use the newest developer build 2.4.0 from 14.04.2015. 
And i get everytime the message:

[E] at qrc:///qml/QGroundControl/HUDControls/QGCMapBackground.qml:32 - "qrc:///q
ml/QGroundControl/HUDControls/QGCMapBackground.qml:32:1: module "QtLocation" is
not installed"
[E] at qrc:///qml/QGroundControl/HUDControls/QGCMapBackground.qml:31 - "qrc:///q
ml/QGroundControl/HUDControls/QGCMapBackground.qml:31:1: module "QtPositioning"
is not installed"
[E] at qrc:///qml/QGroundControl/HUDControls/QGCMapBackground.qml:32 - "qrc:///q
ml/QGroundControl/HUDControls/QGCMapBackground.qml:32:1: module "QtLocation" is
not installed"
[E] at qrc:///qml/QGroundControl/HUDControls/QGCMapBackground.qml:31 - "qrc:///q
ml/QGroundControl/HUDControls/QGCMapBackground.qml:31:1: module "QtPositioning"
is not installed"

What could be the cause here?

I am using the QGC on 2 computers, on a Win7 and the other a Win8 64bit. 
Same message on both. 

Best 
Severin

DonLakeFlyer:
If you get errors like this. Please post them to the google group instead of creating a github issue.

DonLakeFlyer:
We want bugs to start on the email group and then when verified they are real an Issue is created.

DonLakeFlyer:
Reopening since I think this is real.
@dogmaphobic I took a quick look at this. I think QGCSetup.pri needs to be updated for the QtLocation modules. Can you look?

dogmaphobic:
Sure!

dogmaphobic:
Before I go there, when was this changed? If it was a build issue, wouldn't TC fail as well?

DonLakeFlyer:
It's an issue for installer versions where Qt is not already on the machine. Since Qt is already on the TC VMs there isn't any real way to test the installer. So even though QGCSetup.pri runs on all builds, if there is something missing there it will only sometimes show up as a problem when you create an installer from it. And then install on a virgin machine.

DonLakeFlyer:
Most likely here:
    # Copy Qml libraries
    Q_DIR = $$[QT_INSTALL_QML]
[...]

PR #1459, opened 2015-04-15
Add QtLocation stuff to Windows

DonLakeFlyer:
I think the will fix Issue #1456


dogmaphobic:
You beat me to it. I was about to start a new build on Windows to see what the installer produces. I've never created a Windows installer for QGC before...

What I couldn't understand is how it runs from the Release directory. I don't have Qt installed system wide. It's nowhere in the path, yet it all runs.

DonLakeFlyer:
It pulls the location from wherever you did the qmake from.  If you renamed your Qt install directory it would break.

Issue #1458, opened 2015-04-15
Setup is refused even though only one system is connected


LorenzMeier:
This happens if only one system is connected. The issue is real, but I don't understand the mechanism behind it yet.

![airframe_choice_problem](https://cloud.githubusercontent.com/assets/1208119/7156047/da124654-e363-11e4-9b4e-3f78e6bcd7a5.png)

LorenzMeier:
@tumbili For your reference.

DonLakeFlyer:
It's even worse than that. Something recent is causing it to crash on OS X when you do this!

DonLakeFlyer:
Looks like that crash is just when you are testing with MockLink. I can't repro on OSX. Have to try Ubuntu. Is this with a USB connection?

tumbili:
@DonLakeFlyer Hi Don, yes this is with USB connection on Ubuntu 14.04. Even if I restart everything I always get the device files "ttyACM0" and "ttyACM1".

DonLakeFlyer:
Fixed

PR #1460, opened 2015-04-15
Fix connected link check

DonLakeFlyer:
Fix for Issue #1458 (I think)

### DIFF
--- a/src/AutoPilotPlugins/PX4/AirframeComponentController.cc
+++ b/src/AutoPilotPlugins/PX4/AirframeComponentController.cc
@@ -93,12 +93,20 @@ AirframeComponentController::~AirframeComponentController()
 void AirframeComponentController::changeAutostart(void)
 {
     LinkManager* linkManager = LinkManager::instance();
-    
-    if (linkManager->getLinks().count() > 1) {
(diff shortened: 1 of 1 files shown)


LorenzMeier:
Hm, I'm not sure if the general logic works - you can be connected via multiple links to the same system, which would be a valid state. I think its ok to refuse config with multiple links, but then the error message should say that it doesn't like the multiple links. In general I'd prefer to check if you have multiple UAS in the UASManager instead.

DonLakeFlyer:
> you can be connected via multiple links to the same system

Will UAS id be the same for the links if multiple links connected to same system? 

The reason for this check is that it does a disconnect all. I don't want to disconnect some other vehicle the person may be flying.

LorenzMeier:
The UAS ID is globally unique, and if your list count() is 1 then there is not a 2nd system talking to QGC.

DonLakeFlyer:
Ok, let me fix it.

DonLakeFlyer:
I wasn't initially getting what your were saying. Look now. This is what you were talking about right?

LorenzMeier:
I can't see a pushed commit?

DonLakeFlyer:
Oops

LorenzMeier:
yes, looks good!

PR #1462, opened 2015-04-15
Finalizing Primary Flight Control

dogmaphobic:
When I started working with the new Qt mapping system, I added the map as an optional background to the Primary Flight Control. That was a good test bed to see how it all worked. Now that the mapping is all done and working, I decided to instead of removing it from the Primary Flight Control, to fully integrate it properly. That way you have the option of a standard, artificial horizon attitude indicator type, as it has always been and a new map mode, with basic flight instruments.

The normal view is basically what it has always been, now using QML controls instead:

screen shot 2015-04-15 at 1 49 03 am

There is now a button on the top left corner where you can select the view options. The idea of a button is to help with touch interfaces.

screen shot 2015-04-15 at 1 49 05 am

If you select a *Map Background*, the interface changes, adding a current location map on the background and moving the indicators out of the way so you can have a better view.

screen shot 2015-04-15 at 1 47 27 am


DonLakeFlyer:
Awesome stuff. With my Qml widget direction idea I think the Fly view should just change to be the map. Then you can overlap whatever controls over it you want. So in the last screen the PFD bubble thing and the airplane compass thing would just be new Qml dock widgets. You can also add say the parameter editor on top if you want. Does that make sense? I think I could use this view to start that implementation, then it would be available for Plan as well.

LorenzMeier:
The last image actually looks like a pretty useful fly view to me. It has all the info I care about during a flight and has the map context, which is great.

src/ui/qmlcommon/QGCAltitudeWidget.qml
```diff
@@ -74,34 +72,42 @@ Rectangle {
...
+            spacing: _reticleSpacing
+            Repeater {
+                model: _speedArray
+                anchors.left: parent.left
+                Rectangle {
+                    property int _alt: modelData
+                    width:  (_alt % 10 === 0) ? 10 : 15
+                    height: _reticleHeight
+                    color:  Qt.rgba(1,1,1,0.35)
+                    Text {
```
DonLakeFlyer: QGCText so you get the correct default font?

src/ui/flightdisplay/FlightDisplay.qml
```diff
@@ -48,65 +52,141 @@ Rectangle {
...
-        id: windowBackground
-        anchors.fill: parent
-        anchors.centerIn: parent
-        visible: !attitudeWidget.visible && !mapBackground.visible
-        color:   Qt.hsla(0.25, 0.5, 0.45)
-        z:       0
+        adjustSizes();
     }
 
     Menu {
```
DonLakeFlyer: Is there a font for Menus? If so we probably need a QGCMenu and friends to pick up the default font.

src/ui/flightdisplay/QGCFlightDisplay.h
```diff
@@ -69,6 +70,7 @@ class QGCFlightDisplay : public QGCQmlWidgetHolder
     float   latitude            () { return _latitude; }
```
DonLakeFlyer: Do you know that for these you can just do:
'Q_PROPERTY(float latitude           MEMBER _latitude           NOTIFY latitudeChanged)'
As long as you don't need to reference a latitude() call from C++.

src/ui/qmlcommon/QGCAltitudeWidget.qml
```diff
@@ -74,34 +72,42 @@ Rectangle {
...
-        spacing: _reticleSpacing
-        Repeater {
-            model: _speedArray
-            anchors.left: parent.left
-            Rectangle {
-                property int _alt: modelData
-                width:  (_alt % 10 === 0) ? 10 : 15
-                height: _reticleHeight
-                color:  Qt.rgba(1,1,1,0.35)
-                Text {
```
DonLakeFlyer: QGCText?

DonLakeFlyer:
I marked a few usages of Text instead of QGCText but not all. Also there are a number of hardcoded colors in here. Will all of those work ok with light/dark theme or does there need to be new palette entires.

src/ui/qmlcommon/QGCAltitudeWidget.qml
dogmaphobic: I had completely forgotten about the font work you did. I will revisit these and see where it's applicable. Some of these controls are scalable and the font uses sub-point sizes accordingly.

src/ui/flightdisplay/FlightDisplay.qml
dogmaphobic: If you don't enter the font size, Qt will use the appropriate size within the context. Menu is one of these cases. Yes, if you want globally consistent menu fonts, we should probably switch to a QGCMenu type.

src/ui/flightdisplay/QGCFlightDisplay.h
dogmaphobic: I had only glanced at it and I had the impression it would allow the QLM side to write to these variables, which I don't usually want to allow. To be honest I've never paid much attention beyond the initial look up to see how to use these bindings. Only the other day I noticed what you did with the Main Tool bar. I thought about it last night but I didn't want to change that then. Something else for me to revisit.

dogmaphobic:
> Also there are a number of hardcoded colors in here. Will all of those work ok with light/dark theme or does there need to be new palette entires.

Yes, these things use so many different colors that I found very quickly I could not use the system palette. I have not tried, for the most part at least, using these new controls under a *Light* palette. Yet again, something to revisit.

src/ui/flightdisplay/QGCFlightDisplay.h
DonLakeFlyer: Yup. You're right, MEMBER is read/write.

src/ui/flightdisplay/FlightDisplay.qml
DonLakeFlyer: Probably ok for now.

DonLakeFlyer:
Let me know when this is good to merge.

dogmaphobic:
> Let me know when this is good to merge.

I'm not planing on doing anything soon, so yes, it can be merged. Out of all things, I'm having *Mouse* issues. I'm having a case of 3 blind mice. Going out to Best Buy or something to get something that works. Oh how I miss Fry's, with all its idiosyncrasies. East coast is not at all geek friendly. It sucks... (but there is plenty of water though... :)

src/ui/qmlcommon/QGCPitchWidget.qml
```diff
@@ -28,15 +28,21 @@ This file is part of the QGROUNDCONTROL project
...
+    property ScreenTools screenTools: ScreenTools { }
+    property real pitchAngle:       0
+    property real rollAngle:        0
+    property real size:             120
+    property real _reticleHeight:   1
+    property real _reticleSpacing:  size * 0.15
+    property real _reticleSlot:     _reticleSpacing + _reticleHeight
+    property real _longDash:        size * 0.40
+    property real _shortDash:       size * 0.25
+    property real _fontSize:        screenTools.dpiAdjustedPointSize(size * 11 / 120);
```
DonLakeFlyer: Do you need a custom font size here? The default QGC font is 12 pt. If I need larger of smaller fonts I've been using the default point size in the palette and then increment/decrement from there. That way everything follows from the default palette setting.

DonLakeFlyer:
In looking at this more I still think the fonts are not quite correct. Text needs to change to QGCText so you pull the correct app font which could be changed down the road. Then if you want to specify a point size that differs from the default point size I increment/decrement from the default point size in the palette to get something smaller or larger. That way we can change the font info in the QGCPalette in one place and everything changes along with it. For now it's ok that we don't have a QGCMenu but we will need one eventually.

dogmaphobic:
> In looking at this more I still think the fonts are not quite correct. Text needs to change to QGCText so you pull the correct app font which could be changed down the road.

I'm sure what that means. I did a global search and could not find any definition of *QGCText* anywhere. What am I missing?

DonLakeFlyer:
Sorry, QGCLabel

dogmaphobic:
Ha! Let me check...

DonLakeFlyer:
It doesn't do much just sets point size correctly and color. Any place you would use Label you would use the QGC Version instead. Generally Text is not used since that is formatted text whereas Label is plain text. I always get confused by the two. By using the QGC version of everything it provide a global place to override any settings we may want to tweak further down the road.

dogmaphobic:
Yes, it makes perfect sense. Specially in  places when I don't define a size and expect the system to provide a proper one automatically. I'm adding them now...

DonLakeFlyer:
If you look at the new vehicle summary qml I just pushed you can see how I use the point size from the palette as the base and then add points to it to gets something incrementally larger than the default font.

dogmaphobic:
While going through all the QML files, there seems to be a lot of inconsistencies on the versions of QtQuick imported. Things like:
```
import QtQuick 2.3
import QtQuick.Controls 1.2
import QtQuick.Controls.Styles 1.2
```

All sorts of different versions are being used. For QtQuick for instance, some times it's 2.1, others are 2.2, 2.3 and even 2.4. Should we consolidate all to one single standard version?

dogmaphobic:
Hold on on merging. I found some weird font handling in the map slider (I took that from the Qt Location example code)...

DonLakeFlyer:
> there seems to be a lot of inconsistencies on the versions of QtQuick imported

I started noticing that as well. I tend to copy stuff from the docs when I'm stealing examples. And those are all over the map as well so my stuff ends up being all over the map! I tried to think of a way to make this be more automatic. I couldn't figure out way to make this be more automatic like command include file.

dogmaphobic:
> Hold on on merging. I found some weird font handling in the map slider (I took that from the Qt Location example code)

Over and done with... Ready to go.

DonLakeFlyer:
I'll merge after it clears TC one last time. Thanks this is a great pull. Really nice visuals.

src/ui/qmlcommon/QGCPitchWidget.qml
dogmaphobic: Just for posterity's sake, in this particular case above, as well as others like it, the entire gadget is scalable. All internal sizes are computed based on the overall widget size, and that includes the font point size. The base size for the gadget is 120. All internal sizes are computed by a factor derived from the actual size. So, if the gadget is set to 120, the font point size (before DPI adjustment) is 11 points. If you shrink the gadget to 60, the resulting font point size is computed to 5.5 and so on with all internal dimensions. Once this is merged, you can see the result when you resize the flight display window to its minimum size. The gadgets shrink along with it.

DonLakeFlyer: Makes sense

dogmaphobic:
Well, I'm now made a new type of "draggable" items, which both the Attitude and Compass gadgets derive from. That allows you to position those wherever you want within the view. Hold on to the merge :)

LorenzMeier:
Gus, can the round instruments have a transparent background?

dogmaphobic:
> Gus, can the round instruments have a transparent background?

Yes, you can change the overall opacity. Let me see how it looks...

LorenzMeier:
I meant more the square fields covering the map but not holding any info, not the foreground.

dogmaphobic:
> I meant more the square fields covering the map but not holding any info, not the foreground.

Ha! Well, the compass would be fine but the attitude gadget would not work. The artificial horizon (the blue sky and green ground) is a much larger rectangle and I move it and rotate it around based on roll and pitch. The outside rectangle clips it. If I make it transparent, you would see through to the artificial horizon. Qt only clips a rectangle. Because of that, I made the compass square and with the same gray background to match.

dogmaphobic:
> Well, I've now made a new type of "draggable" items, which both the Attitude and Compass gadgets derive from. That allows you to position those wherever you want within the view. Hold on to the merge :)

That will take longer than I expected to make it right. Ignore my comment and please, go ahead and just merge this as is. I will work on the movable, floating gadget in a new PR.

PR #1466, opened 2015-04-16
Link Management Work

dogmaphobic:
I've reworked how links are managed. Gone is the combo box with the links. Instead you will find one single *Connect* button. If nothing is physically connected to the computer and you don't have any connection configured, the only option it will provide is to *Add a Connection*. As you plug in a device, either a Pixhawk or a 3DR Radio Modem, these will be automatically added to the list. These automatically found devices are *Dynamic Devices*. They are not persistent and are also automatically removed from the list once you physically disconnect them from the computer.

Link configurations that are manually entered are *Persistent Links*. These will be saved and always displayed on the connection list. If you edit a *Dynamic* link, it becomes persistent and it will also be saved.

The *Connect* button has three different states. When nothing is connected, it shows *Connect*. If you have one single connection, it shows *Disconnect*. If you have more than one connection, it will also say *Disconnect* but once clicked, a menu will be displayed showing the connected links so you can choose which one to disconnect.

screen shot 2015-04-15 at 10 30 58 pm

### DIFF
--- a/src/QGCConfig.h
+++ b/src/QGCConfig.h
@@ -12,7 +12,7 @@
 
 // If you need to make an incompatible changes to stored settings, bump this version number
 // up by 1. This will caused store settings to be cleared on next boot.
-#define QGC_SETTINGS_VERSION 5
+#define QGC_SETTINGS_VERSION 6
 
 #define QGC_APPLICATION_NAME "QGroundControl"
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
Very nice, makes a lot of sense

PR #1468, opened 2015-04-16
Removing old Primary Flight Display.

dogmaphobic:
I guess we won't be using this anymore...


DonLakeFlyer:
Nothing I like more than throwing code away!

dogmaphobic:
I actually felt bad because someone spent an awful lot of time writing all that :)

PR #1471, opened 2015-04-16
Full meta data support in fact system

DonLakeFlyer:
Fact System now support fully supports all meta data. If meta data xml is not available generic meta data is created. Here is what is available:
- group: This is the parameter group. Defaults to "Default Group" if not available.
- type: type is always available
- defaultValue: Default value for param. This is not always available, depending on meta data. Check defaultValueAvailable for correctness. If not available it default to 0, but that value should not be used.
- defaultValueAvailable: true: defaultValue is valid, false: you should not used defaultValue
- shortDescription: Short description
- longDescription: Long winded description
- units: Units for value (user readable string)
- min/max: Min and max for value. If meta data is not available default is set using the min/max for the type.

Other stuff:
- Converted parameter meta data loader to V2 spec firmware xml file format
- Parameter meta data loader now more resilient to bad xml
- Parameter editor now shows non-default values in orange and has right click menu:
screen shot 2015-04-16 at 10 27 58 am


DonLakeFlyer:
Still fixing and adding some things to this

DonLakeFlyer:
There are still some param dups which I can't get rid of given the python parser on the firmware side, but there are much fewer of them now. So in the case I hit a dup I toss the meta data and fall back to defaults since I can't trust the meta data. And while typing this I realized I didn't implement it correctly so I need another commit!

DonLakeFlyer:
Last commit. Since I was spending so much time on getting clean meta data firmware to QGC I decided to add a Properties.. dialog to the right-click menu to show it all off:
screen shot 2015-04-16 at 1 48 57 pm

LorenzMeier:
Awesome! Looking really great!

DonLakeFlyer:
Performance is kind of crappy because I'm putting too many things in the repeater. I'll merge this when clean, then I'll fix perf later.

severinleuenberger:
Congrats. Awesome! Looks nice and clean. And will be very helpful. Until now, I was standing with the parameter reference on paper on the field. Thank you.

PR #1472, opened 2015-04-17
Created new movable/scalable Item type.

dogmaphobic:
An early experiment where the idea is to have arbitrary (QML) widgets on the *Primary Flight Display*, which you can move them around as well as change their sizes.

Double-clicking on the gadget causes it to restore its original position and size.

screen shot 2015-04-17 at 12 35 20 am


DonLakeFlyer:
Awesome!

PR #1476, opened 2015-04-18
add STM32F4Discovery to supported board by adding BoardID to bootloader

nopeppermint:
I just tryed to connect the stm32f4discoveryboard with qgroundcontrol.
Before doing this a installed the px4bootloader on the discovery.
If I manual upload firmware with the px4bootloader its working without any problems.
If I connect with qgroundcontrol I get:

    Board found: Port: \.\COM30
    Description: \.\COM30
    Attemping to communicate with bootloader...
    Error: Get Board Info: Unsupported board: 99

in the Qgroundcontrol terminal it's:
Found Board:
port name: "com30"
description: PX4 FMU
vendor ID 9900
product id 1
Bootloader error: "Get Board Info: Unsupported board: 99""""

=> according to https://github.com/PX4/Firmware/blob/master/Images/px4-stm32f4discovery.prototype
the Board ID is 99

=> not tested

### DIFF
--- a/src/VehicleSetup/PX4Bootloader.cc
+++ b/src/VehicleSetup/PX4Bootloader.cc
@@ -439,7 +439,7 @@ bool PX4Bootloader::getBoardInfo(QextSerialPort* port, uint32_t& bootloaderVersi
     if (!getBoardInfo(port, INFO_BOARD_ID, _boardID)) {
         goto Error;
     }
-    if (_boardID != _boardIDPX4Flow && _boardID != _boardIDPX4FMUV1 && _boardID != _boardIDPX4FMUV2 && _boardID != _boardIDuNode) {
+    if (_boardID != _boardIDPX4Flow && _boardID != _boardIDPX4FMUV1 && _boardID != _boardIDPX4FMUV2 && _boardID != _boardIDuNode != _boardIDSTM32F4Discovery && _boardID ) {
         _errorString = tr("Unsupported board: %1").arg(_boardID);
         goto Error;
     }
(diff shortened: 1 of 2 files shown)


nopeppermint:
as there is no mavlink support for stm32f4discovery in the current master branch, this will not work at the moment.

DonLakeFlyer:
Supported boards are boards for which there is posted firmware up on Travis: PX4FMU_V2, PX4FMU_V1 and PX4FLOW. Since this board doesn't report itself as one of those then I assume it runs it's own custom firmware. If that is the case then this pull is correct. When you go to flash, it should warn you that you need to select custom firmware if you select one of the other options. Can you test that?

> as there is no mavlink support for stm32f4discovery in the current master branch, this will not work at the moment.

Do you mean before or after this change. If after this change what is still wrong?

DonLakeFlyer:
Discussed with Lorenz. We're going to change the way unknown board ids work. If we can find a bootloader we'll allow custom firmware flash with a matching .px4 file board id. So anyone that needs a quick fix for discovery can use this pull. I'll leave it sitting here. I'll put up the full fix in a day.

DonLakeFlyer:
@nopeppermint Replacement pull is Pull #1477. Can you test this?

nopeppermint:
@DonLakeFlyer 
firmware_update_qgroundcontrol_discovery
Custom Firmware update with Qgroundcontrol and stm32f4discoveryboard is working :+1: 
standard_version_discovery
Standard Version, Beta Testing and Developer Build is not working, but gives you a hint how to do it :+1:

DonLakeFlyer:
You have to do what it says which is to select "Custom Firmware". Standard/Beta/Developer are for Pixhawk boards only.

DonLakeFlyer:
Oops, sorry missed the end part. So working now, thanks.

PR #1478, opened 2015-04-18
First android commit

dogmaphobic:
This is the start of an Android build. I'm making this commit now because it's becoming harder and harder keeping it up-to-date having to do major merge work each time the master branch changes.

It's in no way usable yet. It builds and runs but it still needs a lot of work:

- We will need to work on all aspects of UI dealing with hardcoded pixel sizes. We will need to use proportional dimensions for everything as you can see below. Note that I've temporarily increased the tool bar height (in pixels) for Android builds so I could actually click it :)
- The whole concept of *Main Menu* up on top doesn't work for mobile devices. I already have some ideas how to solve that.
- There is still a lot of code being built, which will be excluded for mobile targets.
- Serial communications is not quite working. At least not reliably. I'm getting crashes in some places where the code is trying to write to files in locations it is not supposed to.

The Java USB Serial driver is the same used by Tower, though I've adapted it for use from c++.

Other than that, it has come a long way since I started. Note that there should have been no changes to the normal (non Android) build, other than android specific exclusions here and there (lots more to come).

s


DonLakeFlyer:
Can I compile this in TC somehow? If so I could at least make sure it doesn’t build break.

LorenzMeier:
Travis offers Android and iOS builds - would that be an option? I find the readability / visibility and performance of Travis really good, and its relatively easy to set up and maintain.
http://docs.travis-ci.com/user/languages/android/

dogmaphobic:
It's not truly an Android build. It's a Qt build targeting Android. You should be able to run it on the same machine you run the Mac OS build as that is your *real* machine (no need for a VM). For now I only have setup to run from within Qt Creator. Let me set it up so it can be built the same way you build the Mac OS build. Then you just need to install 3 things:

- The Android SDK
- The Android NDK
- The Android ARM kit for Qt (Android for armeabi v7 - Part of the Qt package)

Once I have it setup I can add the instructions to README.

dogmaphobic:
For completeness' sake, it's a *qmake* build. It generates a proper makefile that builds everything, including the Java wrapper for Android. The iOS build works the same way. The same qgroundcontrol.pro project file generates makefiles for all targets in the same way. In theory, you could build all platforms on the same computer (using cross-compilation). You just wouldn't be able to run unit tests.

DonLakeFlyer:
Ok. Let me know when it's ready and I'll add a TC build.

DonLakeFlyer:
What's the issue with MockLink? That's really handy for testing things without needing a real connection.

dogmaphobic:
> What's the issue with MockLink?

Nothing with it in particular. I just had to disable unit testing in general until I get something meaningful and MockLink was part of it.

dogmaphobic:
> Ok. Let me know when it's ready and I'll add a TC build.

It will take me a bit to get that going. I won't really have time to do much until Monday.

DonLakeFlyer:
I don' think there's anything in unit tests that wouldn't run on Android. If there is we can disable individual unit tests. Would be handy to have around then we could do something like on hardware testing line the firmware does. Also seems like we could test in the simulator as well?

dogmaphobic:
I took it out firstly because they were causing build issues. Nothing that can't be fixed but when I thought about it, unit testing this would be a major ordeal. You would need to have the target device connected, deploy the application into the device and execute it for unit testing while capturing the output remotely. Doable but way over the top complicated for an automated build scenario.

DonLakeFlyer:
I'd like to still at least keep it building. Because if it gets too far away it will be hard to ever bring it back. The firmware does exactly what you say. Even snaps a picture of the board so you can see the led after it's done. Also, you can't tun tests in the simulator?

QGCApplication.pro
```diff
@@ -78,6 +81,11 @@ QT += testlib
 # OS Specific settings
 #
 
+AndroidBuild {
+    DEFINES += __android__
+    DEFINES += __STDC_LIMIT_MACROS
```
DonLakeFlyer: The compiler settings are all in QGCCommon.pri so that they are common to all projects. You'll find a __STC_LIMIT_MACROS section already in there.

dogmaphobic:
Makes sense. I will re-enable it. The Android simulator is crap and useless. The iOS simulator works well but it's a different code base (you actually generate x86 code instead of Arm). It's better than nothing but you aren't exactly testing the real thing.

QGCApplication.pro
dogmaphobic: Got it. This was a merge issue for when there was only one project file. I will move it.

QGCSetup.pri
```diff
@@ -40,8 +44,10 @@ WindowsBuild {
     BASEDIR_COPY_RESOURCE_LIST = $$replace(BASEDIR,"/","\\")
     QMAKE_POST_LINK += $$escape_expand(\\n) $$QMAKE_COPY_DIR \"$$BASEDIR_COPY_RESOURCE_LIST\\flightgear\" \"$$DESTDIR_COPY_RESOURCE_LIST\\flightgear\"
 } else {
-    # Make sure to keep both side of this if using the same set of directories
-    QMAKE_POST_LINK += && $$QMAKE_COPY_DIR $$BASEDIR/flightgear $$DESTDIR_COPY_RESOURCE_LIST
+    !AndroidBuild {
+        # Make sure to keep both sides of this if using the same set of directories
+        QMAKE_POST_LINK += && $$QMAKE_COPY_DIR $$BASEDIR/flightgear $$DESTDIR_COPY_RESOURCE_LIST
```
DonLakeFlyer: Reminds me that I don't see the FlightGear of HIL simulation code pulled out. Something for later most likely.

DonLakeFlyer:
> Makes sense. I will re-enable it.

Don't bother right now. let's get this in and deal with there rest laster. That would we won't get into merge hell. I'm fine with merge if you are.

QGCApplication.pro
DonLakeFlyer: > I will move it.

I think it still needs to be there because it isn't on for all OS.

DonLakeFlyer: Oops sorry, you said "move" not "remove"

DonLakeFlyer:
@dogmaphobic I'm heading out the door. You can merge this whenever you'r happy with it.

QGCSetup.pri
dogmaphobic: Yeah, I noticed a few of these things too. I figured at some point we need to go through a dead code hunt. This will come when we get everything to QML. For instance, SerialSettingsDialog (.h, .cc, .ui) are only used by TerminalConsole, and even there, it doesn't do the right thing.

PR #1485, opened 2015-04-20
Added RSSI Indicator to ToolBar.

dogmaphobic:
I've added an RSSI indicator to the tool bar. It shows *Telemetry* RSSI and *Remote Control* RSSI. It will display both if both are available or only the one that is available. If neither is available the indicator is not shown at all. I've also added a setting in the main preferences dialog to enable/disable showing RSSI in the tool bar (like all other indicators before it).

In the process, I've also added some logic to look for spoken text messages for things like POSCTL and ALTCTL and replaced the spoken text with a full sentence like *Position Control*, or *Altitude Control*. I'm a bit confused by the messages for MANUAL and ALTCTL. They both say the same thing: *System now is in stabilize mode*. Is that correct?

Questions about RSSI:

- What is the range for these values? I would like to show a normalized value between 0 and 100.
- What is the proper value for telemetry RSSI? There are two: *rssi* and *remote rssi*. Which one makes more sense to be there? What you see there now is *rssi* (not *remote rssi*)
- Remote Control RSSI seems out of whack. First, it doesn't seem to change in relationship to how far the radio is. I even wrapped the entire radio in aluminum foil and it didn't seem to make any difference. Is the PX4 actually reading RSSI off my receiver? (FrSky X8R)
- Before I turn the radio on, only the telemetry RSSI is available as expected. I then turn the radio on and the display now shows both values as expected. However, if I turn the radio off, I get all sorts of audio alerts but the RSSI value doesn't change. Disconnecting and reconnecting to the PX4 makes no difference (kind of follows the logic). I could not find any message/signal I could use to show that RC signal is lost.

Of minor note:

- Fixed a small typo in the description of actionRetransmissionTimeoutChanged in MAVLinkProtocol.h
- Added code to the flight display to only emit coordinate updates to the QML UI if the longitude and/or latitude actually changed.
- Tweaked the various dimensions of the tool bar elements to accommodate the new RSSI indicator.
- I've tweaked the various batteries icons to a single solid charge amount as the way it was before was not scaling properly.

Screen grab showing both RSSI values:

screen shot 2015-04-20 at 12 02 40 am

This one shows when only one is available:

screen shot 2015-04-20 at 12 19 29 am


LorenzMeier:
> Questions about RSSI:
> 
> What is the range for these values? I would like to show a normalized value between 0 and 100.
> What is the proper value for telemetry RSSI? There are two: rssi and remote rssi. Which one makes more sense to be there? What you see there now is rssi (not remote rssi)
Actually you need both: RSSI: How well you receive telemetry on the ground. Remote RSSI: How well *the plane* receives your commands. Both are vital. As for the values: The message is poorly documented - what range are you getting on the bench? 0-100 or rather 0-255-ish? I didn't do the spec for the message, but I'm happy to fix the specs there. http://mavlink.org/messages/common#RADIO_STATUS
> Remote Control RSSI seems out of whack. First, it doesn't seem to change in relationship to how far the radio is. I even wrapped the entire radio in aluminum foil and it didn't seem to make any difference. Is the PX4 actually reading RSSI off my receiver? (FrSky X8R)
Only if you connect the radio's RSSI port to the SBUS-labeled port on Pixhawk, which will then try to sample this as analog value.
> Before I turn the radio on, only the telemetry RSSI is available as expected. I then turn the radio on and the display now shows both values as expected. However, if I turn the radio off, I get all sorts of audio alerts but the RSSI value doesn't change. Disconnecting and reconnecting to the PX4 makes no difference (kind of follows the logic). I could not find any message/signal I could use to show that RC signal is lost.
You might have to dive into the UAS class. The MAVLink message you are looking for is this one: http://mavlink.org/messages/common#RC_CHANNELS. Before it stops updating you should get one with RSSI being either 0 or 255. If not, let me know and we'll look into it on the FW side.

dogmaphobic:
> Actually you need both: RSSI: How well you receive telemetry on the ground. Remote RSSI: How well the plane receives your commands. Both are vital. As for the values: The message is poorly documented - what range are you getting on the bench? 0-100 or rather 0-255-ish?

Got it. The value comes in as uint8_t so I will normalize it from 0-255 based on the fact I'm getting values over 100. What got me confused at first was that the highest number I observed was in the 130s even though the radios are about 20 meters apart. I will add separate indicators for telemetry (with both values) and RC.

> Only if you connect the radio's RSSI port to the SBUS-labeled port on Pixhawk, which will then try to sample this as analog value.

That explains it. No, I don't have it connected properly. I only have SBUS from the receiver into the SBUS IN port into the PX4 (labeled RC). I always wondered how would the RSSI value magically get to the PX4. I just didn't think about it and assumed SBUS was bi-directional and somehow included RSSI. The values for RC RSSI are easy to interpret because the code differentiates between MAVLINK_MSG_ID_RC_CHANNELS  and MAVLINK_MSG_ID_RC_CHANNELS_SCALED and both mavlink_rc_channels_t and mavlink_rc_channels_scaled_t are properly documented.

Speaking of which, is it possible to edit the pages at pixhawk.org to add a better description of these things? I've also noticed a few discrepancies such as the picture showing to connect the telemetry radio on Port 2 while I could only make it work if I connected to Port 1.

> You might have to dive into the UAS class. The MAVLink message you are looking for is this one: http://mavlink.org/messages/common#RC_CHANNELS. Before it stops updating you should get one with RSSI being either 0 or 255. If not, let me know and we'll look into it on the FW side.
> 
> Use the MAVLink inspector tool widget to look at the raw message to judge wether you're getting valid data.

Will do...

Thanks!

LorenzMeier:
It's a wiki, you can register and edit yourself.

dogmaphobic:
Spoke too soon. The struct mavlink_rc_channels_scaled_t has this:
```
uint8_t rssi; ///< Receive signal strength indicator, 0: 0%, 100: 100%, 255: invalid/unknown.
```

While the code receiving it does this:

```
emit remoteControlRSSIChanged(channels.rssi/255.0f);
```

If you don't know off the top of your head, I can dig into the firmware and sort it out...

dogmaphobic:
> It's a wiki, you can register and edit yourself.

I've looked at that page a million times, yet I managed to not ever see the *Register* button...

Done. Thanks!

dogmaphobic:
Regarding the telemetry RSSI, based on the Silabs data sheet, the RSSI register is a log scale:
```
Received signal strength indicator (RSSI) is an estimate of the signal strength in the
channel to which the receiver is tuned. The RSSI value can be read from an 8-bit
register with 0.5 dB resolution per bit.
```
I could not find in the firmware code where the value is collected. Is there any processing done to the value or is it raw from the radio?

LorenzMeier:
Looking at the code I'd say it's raw
https://github.com/tridge/SiK/search?p=2&q=rssi&type=Code&utf8=✓

dogmaphobic:
Perfect. I just show the value in db then (rssi >> 1 == db).

dogmaphobic:
All right... I'm somewhat confident, but not certain, that the telemetry RSSI is now properly computed and displayed. The values are in dB.

The RC telemetry is anybody's guess. I'm about to bail now but when I get back, I will try to physically measure the RSSI output of my receiver and try to make some sense about the numbers I'm getting.

screen shot 2015-04-20 at 11 14 44 am

DonLakeFlyer:
Another thing to keep in mind is that I don't think you can get RSSI from Spektrum? I'll give it a try to see what I get back with mine. Take a while. Chores all day today.

DonLakeFlyer:
Does everything still fit into 1024x768 on Ubuntu with the left hand tool thingy?

LorenzMeier:
On the units side: We now have db next to a linear scale for RC (or maybe that one is logarithmic as well, don't know). How do they relate? Is there a way to simplify it without lying to the user?

LorenzMeier:
And sorry for being so technical - it looks awesome, and seeing a new screenshot with new functionality in the morning already makes my day! Thanks!

dogmaphobic:
> On the units side: We now have db next to a linear scale for RC (or maybe that one is logarithmic as well, don't know). How do they relate? Is there a way to simplify it without lying to the user?

If the telemetry radio data truly is the raw data, it's logarithmic, and that's what is being shown. The numbers also seem right, though I have not measured them. I've used the information I found on the (Silabs) data sheet to convert the values into dB.

https://www.silabs.com/Support%20Documents/TechnicalDocs/Si1000.pdf

What I don't know is what the numbers for RC RSSI mean.

dogmaphobic:
> Does everything still fit into 1024x768 on Ubuntu with the left hand tool thingy?

I've been mindful of it. It's still sub 100 but I'm not quite done with the layout.

dogmaphobic:
> t's still sub 100 but I'm not quite done with the layout.

1000! :)

LorenzMeier:
@dogmaphobic We've discussed the RSSI display and users think that a percent indication is key. So I guess it might make sense to work out the correct conversion based on a physical range check and either do it in the radio or in QGC. You and Don are most welcome on the PX4 weekly call if you're interested:
https://groups.google.com/forum/#!topic/px4users/trSd9ngFwJg

dogmaphobic:
I tried mumble.3drobotics.com but I think it's already done. Nobody around...

dogmaphobic:
> We've discussed the RSSI display and users think that a percent indication is key.

It will be interesting normalizing a log scale to percentage... :)

dogmaphobic:
I think this is done for now. The elements are in place, all sizes adjusted and ready. I've made the icon colors a bit more consistent while at it. The pending issue is what to do about the RSSI values shown to the user. If you all agree, I would merge this as is and work with the values on a different PR (unless we come up with a solution now). Otherwise it will merge rot.

P.S. The absurd amount of error messages was on purpose to see the maximum width of all elements :) I kept flicking the mode control to auto. The side effect I noticed was that the spoken messages kept going for a good two minutes after I let it go.

screen shot 2015-04-20 at 5 04 08 pm copy

LorenzMeier:
Agreed - then lets get this merged once Don is happy.

src/audio/QGCAudioWorker.cpp
```diff
@@ -187,10 +187,18 @@ bool QGCAudioWorker::_getMillisecondString(const QString& string, QString& match
     return false;
 }
 
-QString QGCAudioWorker::_fixMillisecondString(const QString& string) {
+QString QGCAudioWorker::_fixTextMessageForAudio(const QString& string) {
     QString match;
     QString newNumber;
     QString result = string;
+    //-- Look for modes
+    if(result.contains("POSCTL", Qt::CaseInsensitive)) {
```
DonLakeFlyer: Seems like we should do this sort of thing in the firmware instead of fighting a never ending battle in QGC.

LorenzMeier: Yes, agreed.

DonLakeFlyer:
@dogmaphobic Didn't I see you with a firmware pull already? Time to go to the dark side :)

dogmaphobic:
> Didn't I see you with a firmware pull already? Time to go to the dark side :)

Yep, if you consider two PRs that only changed the comments, certainly :)

While at it, it seems GitHub is under siege again by the dark side. It took a while to load a page after various *Server Not Available* errors...

DonLakeFlyer:
@dogmaphobic Good to merge this through, then you can play on the firmware side and tweak this audio back when you get your stuff through there as well.

dogmaphobic:
> firmware side and tweak this audio

I'm wondering about that. The issue is the size of the messages and the time it takes to transmit on a narrow bandwidth. We have to think about this. I would prefer short, known keywords from the firmware and do the *translation* on the QGC side where we have a lot more memory and resources to deal with it.

DonLakeFlyer:
The packet size for text message is a fixed buffer I believe. Fill it or waste it. Same bandwidth. @LorenzMeier can confirm.

DonLakeFlyer:
Yup fixed size 50 char buffer, use it or lose it:

static inline uint16_t mavlink_msg_statustext_pack(uint8_t system_id, uint8_t component_id, mavlink_message_t* msg,
						       uint8_t severity, const char *text)
{
#if MAVLINK_NEED_BYTE_SWAP || !MAVLINK_ALIGNED_FIELDS
	char buf[MAVLINK_MSG_ID_STATUSTEXT_LEN];
	_mav_put_uint8_t(buf, 0, severity);
	_mav_put_char_array(buf, 1, text, 50);
        memcpy(_MAV_PAYLOAD_NON_CONST(msg), buf, MAVLINK_MSG_ID_STATUSTEXT_LEN);
#else
	mavlink_statustext_t packet;
	packet.severity = severity;
	mav_array_memcpy(packet.text, text, sizeof(char)*50);
        memcpy(_MAV_PAYLOAD_NON_CONST(msg), &packet, MAVLINK_MSG_ID_STATUSTEXT_LEN);
#endif

	msg->msgid = MAVLINK_MSG_ID_STATUSTEXT;
#if MAVLINK_CRC_EXTRA
    return mavlink_finalize_message(msg, system_id, component_id, MAVLINK_MSG_ID_STATUSTEXT_LEN, MAVLINK_MSG_ID_STATUSTEXT_CRC);
#else
    return mavlink_finalize_message(msg, system_id, component_id, MAVLINK_MSG_ID_STATUSTEXT_LEN);
#endif
}

dogmaphobic:
I've changed the RC RSSI code to treat the incoming value as per the original spec. It expects a value from 0 to 100 as a valid value and assumes anything over that to be N/A. The RC RSSI indicator is not shown when RSSI == N/A.

Issue #1480, opened 2015-04-19
Airframe config link re-connect crashes application


LorenzMeier:
This is a fresh (completely clean) build off current master:

```
Process:         qgroundcontrol [32449]
Path:            /Users/USER/*/qgroundcontrol.app/Contents/MacOS/qgroundcontrol
Identifier:      org.qgroundcontrol.qgroundcontrol
Version:         2.0.0 (227)
Code Type:       X86-64 (Native)
Parent Process:  Qt Creator [30464]
Responsible:     Qt Creator [30464]
User ID:         501

Date/Time:       2015-04-19 11:44:29.269 +0200
OS Version:      Mac OS X 10.9.5 (13F1066)
Report Version:  11
Anonymous UUID:  0BE00B61-2D1F-543C-38F9-2BD7ACFA6ECA

Sleep/Wake UUID: D16301AB-98EE-45F8-9231-FB885B717103

Crashed Thread:  0  Dispatch queue: com.apple.main-thread

Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Codes: EXC_I386_GPFLT

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   org.qgroundcontrol.qgroundcontrol	0x00000001000fc426 LinkManager::createConnectedLink(LinkConfiguration*) + 22
1   org.qt-project.QtCore         	0x0000000101813114 QSingleShotTimer::timerEvent(QTimerEvent*) + 132
2   org.qt-project.QtCore         	0x0000000101805300 QObject::event(QEvent*) + 48
3   org.qt-project.QtWidgets      	0x0000000102bcedbb QApplicationPrivate::notify_helper(QObject*, QEvent*) + 251
4   org.qt-project.QtWidgets      	0x0000000102bd2110 QApplication::notify(QObject*, QEvent*) + 8192
5   org.qt-project.QtCore         	0x00000001017d9fb3 QCoreApplication::notifyInternal(QObject*, QEvent*) + 115
6   org.qt-project.QtCore         	0x000000010182e3f6 QTimerInfoList::activateTimers() + 1270
7   libqcocoa.dylib               	0x0000000104505d12 QCocoaEventDispatcherPrivate::activateTimersSourceCallback(void*) + 18
8   com.apple.CoreFoundation      	0x00007fff8493f5b1 __CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 17
[...]

PR #1486, opened 2015-04-20
Fix crash on reconnect

DonLakeFlyer:
Fixes Issue #1480


LorenzMeier:
Thanks for fixing this so quick, much appreciated!

PR #1487, opened 2015-04-20
Correctly handle UAS going away

DonLakeFlyer:
Fixes crash on disconnect/connect.


DonLakeFlyer:
TC is barfing on GitHub being a pain.

PR #1490, opened 2015-04-21
Return and Offboard must be on separate channel

DonLakeFlyer:
Updated FlightMode setup to new spec which requires Return and Offboard switches to be on their own channels.


LorenzMeier:
Now I really feel bad. More UI changes than flight code changes for this!

DonLakeFlyer:
This way makes much more sense though.

PR #1493, opened 2015-04-22
Handling multi-platform sizing (fonts and pixels)

dogmaphobic:
Revamping global sizes for both font metrics as well as hard coded pixel dimensions.

This came about when realizing that there was no constant available to compute appropriate dimensions for the various platforms. There is no relationship between the system returned DPI and reality. Instead of returning a hard, physical value, it seems to return an abstract concept based on someone's philosophical rants.

This new system uses instead hard-coded *factors*, pre-computed for each platform and used accordingly. These factors can be changed (font and pixel size independently) and they are saved in the settings. The method for changing it though is somewhat hidden for the time being. Given that this mostly affects QML UI only (some style sheet based fonts used by QWidgets are handled as well), it should be in some proper QML preferences dialog, which is not yet done. For now, you can use the options button in the main flight display (right-click).

That dialog is an experiment in creating a preferences dialog within QML. It is not working (a row of buttons is supposed to be at the bottom, and on Android it's 2/3 off the screen). It looks like QML Dialogs, among other things are a casualty of using QML within QQuickWidget. These things depend on the application being run by QQmlApplicationEngine.

Now the UI looks the same everywhere:

screencap copy

### DIFF
--- a/src/AutoPilotPlugins/PX4/FlightModesComponent.qml
+++ b/src/AutoPilotPlugins/PX4/FlightModesComponent.qml
@@ -364,7 +364,7 @@ Item {
 
                 QGCLabel {
                     text: "FLIGHT MODES CONFIG"
-                    font.pointSize: screenTools.dpiAdjustedPointSize(20);
+                    font.pointSize: screenTools.fontPointFactor * (20);
                 }
 
                 Item { height: 20; width: 10 } // spacer
@@ -894,7 +894,7 @@ Item {
 
                 QGCLabel {
                     text: "FLIGHT MODES CONFIG"
-                    font.pointSize: screenTools.dpiAdjustedPointSize(20);
+                    font.pointSize: screenTools.fontPointFactor * (20);
                 }
 
                 QGCLabel {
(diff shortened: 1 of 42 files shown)


DonLakeFlyer:
I think this is all fine, but I'd like to stare at it for a little bit.

dogmaphobic:
Here is the dialog with the options to change the global *factors*. It would be cool if other people tried this in various scenarios so we can come up with an optimum set of values. I'm particularly interested in how this works in high density displays (Retina, etc.)

fudge

DonLakeFlyer:
I don't see where you are tweaking QGCButton sizes according to OS to make them larger on Android.

DonLakeFlyer:
You can merge when you want. And I"ll look at it after the fact.

dogmaphobic:
> I don't see where you are tweaking QGCButton sizes according to OS to make them larger on Android.

No, I have not done anything yet with *regular* UI elements as they are all used only in the Setup screens for now. I didn't want to change that code and this one in one batch (it would take a lot longer and we would inevitably collide as you are working there). All these changes affect mostly general UI layouts for the toolbar and the primary flight display.

> You can merge when you want. And I"ll look at it after the fact.

I will..

DonLakeFlyer:
My tablet is on the way so I’ll play around with things on that.

PR #1497, opened 2015-04-23
Android SerialPort work.

dogmaphobic:
Still not working right...

The issues are beyond what I understand about Java and/or Android. These changes, besides some cosmetic ones to match the current Qt file headers, take it a bit further but in general, this whole Qt SerialPort layer for Android is not working. Too many crashes (it doesn't at all like being called from a different thread for instance). I'm trying to track down Mike Goza, who originally wrote this and submitted as a patch to Qt (which was never integrated). In the mean time, the only reliable way to connect to Android is UPD/TCP and even that needs to be hardcoded for the moment, as the UI for setting up connections (Qt Widgets) os not usable within Android/iOS.

### DIFF
--- a/android/src/com/hoho/android/usbserial/driver/UsbSerialProber.java
+++ b/android/src/com/hoho/android/usbserial/driver/UsbSerialProber.java
@@ -162,10 +162,10 @@ public enum UsbSerialProber {
      */
     public static List<UsbSerialDriver> findAllDevices(final UsbManager usbManager) {
         final List<UsbSerialDriver> result = new ArrayList<UsbSerialDriver>();
-        Log.i("QGC_UsbSerialProber", "Looking for USB devices");
+        //Log.i("QGC_UsbSerialProber", "Looking for USB devices");
         // For each UsbDevice, call probe() for each prober.
         for (final UsbDevice usbDevice : usbManager.getDeviceList().values()) {
-            Log.i("QGC_UsbSerialProber", "Probing device: " + usbDevice.getDeviceName() + " mid: " + usbDevice.getVendorId() + " pid: " + usbDevice.getDeviceId());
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
Is Tower/DroidPlanner open source. You could see what they are using?

dogmaphobic:
Tower is fully Java (Android). We cannot access the underlying serial port (USB) because the app doesn't have permission. You would only be able to do that on a rooted phone/tablet. I am using the same Java USB/Serial libraries as Tower but with some glue code to talk in between (Qt) c++ and Java. That's what's not working (I think) :)

DonLakeFlyer:
Let me know if you want me to do any poking around. My tablet showed up yesterday but haven’t pulled it out of the box yet.

dogmaphobic:
The code currently calls the Java side and gets return values just fine. The issue is when the Java side calls into Qt. One place to start looking at is here:

https://github.com/dogmaphobic/qgroundcontrol/blob/master/libs/qtandroidserialport/src/qserialport_android.cpp#L145-L170

dogmaphobic:
I would go ahead and merge this. It does fixes some real errors, which led me further ahead to others yet to be solved.

I finished my own, real c++ MavBridge today (using the MavLink C library), which runs on Arm based SBCs (I tried both Udoo and Raspberry PI b+). Up to now I was running a hodgepodge of Python serial/UDP redirectors. I still need to setup a server for these systems to connect and exchange (external) IP addresses so they can communicate using Cellular data.

At any rate, here is what the Setup looks like. Should I spend some time trying to fix it or wait?

screen1 copy

P.S. I'm calling it MavBridge for the moment as there are way to many *MavProxy* programs out there. I also want to integrate video within the same code (so I can throttle traffic better, as opposed to two separate processes competing for UDP traffic).

DonLakeFlyer:
I'll deal with setup. Will help me understand how all the font stuff works.

PR #1501, opened 2015-04-25
Add images for better in-progress mag visuals

LorenzMeier:
@DonLakeFlyer I think it would fix the current dialog if you could replace the image in the "rotate" step with these. Then it will be crystal clear how to rotate. I will try to add gyro integration onboard to ensure people are actually rotating the vehicle before starting the rotate timer. @dogmaphobic @tumbili Feedback welcome.


LorenzMeier:
Together with this PR this should improve things quite a bit:
https://github.com/PX4/Firmware/pull/2083

LorenzMeier:
mag_rotate_down

Here is one of the images as example.

DonLakeFlyer:
I have a massive pull in the works on both the QGC and Firmware sides which fixes almost all the usability complaints. Might be ready some time today:
- Better help text
- Cancel support
- Looser rest position for mag cal
- Better rotation images
- Prompt for board rotation prior to calibration
- Prompt for mag rotations after mag cal
- Other things I can't remember

vehicledownrotate

PR #1503, opened 2015-04-25
Sensor cal rework

DonLakeFlyer:
This is a major rework of sensor cal to provide better usability. It requires a companion pull top firmware to make it all work:
- Calibration can now be cancelled
- Much better help text
- Images which show you how to rotate the vehicle for compass cal
- Prompted to set board rotation before calibrating
- Prompted to set compass rotation(s) after calibrating compass (if needed)
- Other things I forgot


DonLakeFlyer:
Firmware pull: https://github.com/PX4/Firmware/pull/2088

DonLakeFlyer:
screen shot 2015-04-27 at 10 40 48 am

DonLakeFlyer:
Also firmware now has less stringent position detection for mag cal which should make it easier to hit a pose.

PR #1509, opened 2015-04-27
Re-write of parameter system to move all control to FactSystem

DonLakeFlyer:
FactSystem now controls all aspects of loading/reading/writing parameters. All parameter access is through the AutoPilotPlugin object. While making this change i had to remove support for some features instead of putting in a huge effort to rework them to the new architecture. More details below.

Additional Changes:
- Support for Local 3d view was removed. All future mapping support must be on top of QtLocation. With the upcoming mission planning and map display re-write supporting a non-QtLocation based map is just not viable.
- Removed support for Custom Tool Widgets. This will be replaced with the ability to write your own Qml based custom tool widgets using Qml + FactSystem controls.
- The parameter load progress indicator now shows progress whenever there is any set of parameters we are waiting to hear back from the vehicle on.
- The AutoPilotPlugin::refreshParamPrefix now works Example usage is sensor cal re-request all CAL_* parameters back from shield after cal completes.
- Previous param manager implementation wasn't handling high load of requests quite correct. It would drop things on the floor in some cases. Hence the reason why param prefix refresh was not always working correctly.
- MockUAS/MockUASManager no longer supported for unit tests, use MockLink instead
- FileManager unit test temporarily removed. Reason is that I need to convert to MockLink but already have another FileMan pull I'm working on that will merge conflict if I make any changes here.
- Fixed qDebug spam output from meta data loading


mattmccrink:
Is there an ETA for adding the custom widgets (or new framework) back in? We used them extensively for control of our system.... Eeeek.

DonLakeFlyer:
Give me some examples of what you want and I can work on that first.

mattmccrink:
We were using the MAV_CMD enums for things like geo-fencing, payload
deploy, and several custom commands through the MISSION_ITEM message. I
think what I am after is a flexible way to configure and issue commands as
per the mavlink spec. We are not using a PX4 which might put us in right
field on this one. Thanks!

DonLakeFlyer:
So you were using the CommandButton widgets?

mattmccrink:
Yes.

DonLakeFlyer:
Give me a couple days and I'll have it back for you.

mattmccrink:
Thank you, I really appreciate it!

DonLakeFlyer:
@mattmccrink I'm looking at the old command button code and it only seems to send COMMAND_LONG messages. Where you sending MISSION_ITEM messages as well with it somehow?

DonLakeFlyer:
Do you want to just specify your own integer command id or use a pick list of the standard MAV enum ones?

mattmccrink:
Don, we were using the COMMAND_LONG messages. Apologies for the mix up. The
standard MAV enum would be fine. I can use that to add my own as necessary.
Or at this point, would it be easier to design a more flexible container?

DonLakeFlyer:
Pull #1524

mattmccrink:
Taking a look now. Looks amazing. Unrelated (maybe) but I am having build
issues with the release version. TypeMock seems to be undefined in the
LinkManager. I commented this out and it works. It also works if I use the
CONFIG-=debug_and_release CONFIG+=release in the QT call. I can dig further
if this is not something you have come across.
The new parameter system is exactly what I was hoping for.

DonLakeFlyer:
I think I may have screwed up the basing of the pull in relationship to Pull #1520 which is broken because of exactly what you describe. I should have it all cleaned up tomorrow. Once this all gets through and merged in let me know if you need help settings up your ui in qml. It's fairly straightforward but you can make it as complicated as you want.

mattmccrink:
This is really great. I'll take a bit and sink my teeth into it. Really
appreciate it.

mattmccrink:
I am digging into the example now and am getting the "-1 File not found"
return statement. I am using the example posted to qgroundcontrol.org. I
will continue debugging and get back to you. Looks like exactly what I was
after though.
-M

mattmccrink:
Hello Don. The custom command looks great. What do you do most of your QML
editing in? QT creator does not seem to like the linking. Do you happen to
know if Visual Studios will parse them correctly? Thanks!

DonLakeFlyer:
Just a text editor. Qt Creator is horrible at editing Qml in the designer thing.

PR #1511, opened 2015-04-27
Put back Fly View default widgets

DonLakeFlyer:
@dogmaphobic I noticed that these were removed by one of your commits. Was there a reason or just something missed?

### DIFF
--- a/src/ui/MainWindow.cc
+++ b/src/ui/MainWindow.cc
@@ -1086,8 +1086,7 @@ void MainWindow::_loadCurrentViewState(void)
         case VIEW_FLIGHT:
             _buildFlightView();
             centerView = _flightView;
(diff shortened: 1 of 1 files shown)


dogmaphobic:
> I noticed that these were removed by one of your commits. Was there a reason or just something missed?

Cosmetic. I didn't like them and didn't want them there by default. A *first impression* issue, as their UI clashes with QML. I figured those who want them there would have no problem opening them up themselves. Whatever widgets you place there will be remembered in future launches any way. Eventually these widgets will get properly written in QML and we can rethink this whole default side widgets thing.

DonLakeFlyer:
I don't disagree. But users think something is broken since it doesn't work like it did before. Not all of them know that you can add these things. I think keeping the defaults until we convert to Qml will keep most people from being confused. Ok with that?

dogmaphobic:
Function always beats form in these cases. No issues here. I can disable them myself locally so I don't have to look at them :)

PR #1514, opened 2015-04-28
More android

dogmaphobic:
Some further Android SerialPort fixes for @DonLakeFlyer.


libs/qtandroidserialport/src/qserialport_android.cpp
```diff
@@ -109,57 +118,75 @@ QSerialPortPrivate::QSerialPortPrivate(QSerialPort *q)
...
-        envL->DeleteLocalRef(objectClassL);
+        QAndroidJniEnvironment jniEnv;
+        if (jniEnv->ExceptionCheck()) {
+            jniEnv->ExceptionDescribe();
+            jniEnv->ExceptionClear();
+        }
+
+        QAndroidJniObject javaClass(kJniClassName);
+        if(!javaClass.isValid()) {
+            __android_log_print(ANDROID_LOG_ERROR, kJTag, "Java class %s not valid", kJniClassName);
```
dogmaphobic: This is the main Issue (for now) with the SerialPort glue code. Here is where we register the c++ function for the Java side to call. For whatever reason, this part of the code is not finding the * kJniClassName* class (and I'm not allowing the code to proceed). This same class name is used for all outgoing calls (c++ to Java) and working just fine.

DonLakeFlyer:
FYI: TC should be doing a compile only on android builds now. This should at least keep it compile clean.

libs/qtandroidserialport/src/qserialport_android.cpp
```diff
@@ -109,57 +118,75 @@ QSerialPortPrivate::QSerialPortPrivate(QSerialPort *q)
...
+            return false;
+        }
```
dogmaphobic: When returning false here, the port *fails* to open. The main code would normally show an error dialog telling the user there was an error opening the serial port. That is crashing as well, though it has nothing to do with serial port. That's something else we need to deal with later. The crash occurs in *QGCMessageBox::critical()*. Let's fist concentrate on this Java registration.

dogmaphobic:
By the way, another change I had done (and completely removed for this PR) was moving the SerialLink away from threads. I was just using Qt signals the way both UDP and TCP links work. I was wondering how all this Java glue would deal with calls coming from all over the place and different threads. As that did not do anything, I reverted those changes before making this PR.

src/QGCFileDialog.cc
```diff
@@ -217,10 +217,10 @@ void QGCFileDialog::_validate(Options& options)
     Q_ASSERT(qgcApp());
     
     Q_ASSERT_X(QThread::currentThread() == qgcApp()->thread(), "Threading issue", "QGCFileDialog can only be called from main thread");
-    
+#ifndef __android__
     // On OSX native dialog can hang so we always use Qt dialogs
     options |= DontUseNativeDialog;
-    
+#endif
```
dogmaphobic: Ha! I had forgotten about these. This was an attempt to see if this had anything to do with the crash in the Dialogs. It didn't...

src/ui/MainWindow.cc
```diff
@@ -183,8 +183,10 @@ MainWindow::MainWindow(QSplashScreen* splashScreen)
     // Qt 4 on Ubuntu does place the native menubar correctly so on Linux we revert back to in-window menu bar.
     // TODO: Check that this is still necessary on Qt5 on Ubuntu
 #ifdef Q_OS_LINUX
+#ifndef __android__
     menuBar()->setNativeMenuBar(false);
 #endif
+#endif
```
dogmaphobic: This is another experiment. By setting this, or not setting as it is the case, the main menu is shown natively. I personally don't like it as it wastes a lot of space. I want to integrate the main menu with the main tool bar. The idea is to have a little button that you *slide* and the menu slides into view from the side (the standard with mobile devices). I tried that and the logic works. What doesn't work is using QML dialogs as this don't seem to like not having a real QML Application is the main root.

DonLakeFlyer: There isn't really much in the menu. I think we can get rid of it completely and just include a hamburger with a few things on it.

DonLakeFlyer:
TC was mostly done. Android build not quite working yet. Just need to set android environment variables. I'll deal with that tomorrow.

src/ui/MainWindow.cc
dogmaphobic: Well, for the time being, it's the only way to access the rest of the UI. Otherwise you can only access the four things the four main tool bar buttons allow you to access. But yes, it should go away once we have a proper, mobile style menu.

dogmaphobic:
Unbelievable. They actually have an emoji for hamburger: :hamburger:

Issue #1507, opened 2015-04-27
Ability to cancel parameter change


AndreasAntener:
(QGC master from today)
ATM it doesn't seem to be possible to abort a parameter change. Since the changes are now transmitted instantly, it should at least be possible to hit [ESC] to revert an accidental/unwanted change.

LorenzMeier:
Actually that's a very important point - the send safeguard is missing now.

@DonLakeFlyer it would be nice if the UI would show very boldly it's going to send the values on hitting enter or leaving the field. You can easily crash the vehicle with a wrong param, and at the level of the param editor we can't prevent that, unless we introduce a special mode:

One example: the param editor has to allow a roll P gain of zero, because that's how you can best tune the rate gains. Doing that in hover will lead to a crash.

Now we have two options: do the field validation feedback-only, so the field turns red if you enter zero. Or add a checkbox to disengage field validation and only accept safe values if validation is on. What did you have in mind?

At any rate, field reset on ESC sounds like a real need.

DonLakeFlyer:
ESC is doable. Validation against the meta data is going to be added. But in the case you describe if 0 is valid as far as the meta data is concerned then it will be allowed even once validation is implemented. If the meta data says 0 is invalid it won't be allowed unless we add some sort of mode which lets you override the validation. That last option would seem to be safer.

DonLakeFlyer:
Another option: Be much more annoying when changing parameters while armed. In this case add an "Are you sure?" with warning about crashing vehicle. I kind of like that option.

AndreasAntener:
@DonLakeFlyer : no please don't make it more annoying ;) especially not with more dialogs that ask stuff all the times..
If the GUI stays like it is now, ESC ability is a must.

Regarding validation: I'm not sure why the old "changed before send" view disappeared, but validation messages or tips/hints could have easily been displayed there. So the user gets a chance to review his changes, and maybe override validation errors.

DonLakeFlyer:
Old style is gone because I hand't figured out how to implement it with the new FactSystem architecture. I think I have an idea on how to bring it back.

DonLakeFlyer:
I can't seem to get Escape to work, still trying. But Undo does work for me at least on OSX. Command+Z. Probably works on other OS as well.

AndreasAntener:
Hm.. if you can get the old pre-send view back, ESC doesn't need to work there anymore of course ;)
[...]

PR #1519, opened 2015-04-28
New parameter editor

DonLakeFlyer:
- Fixes Issue #1507 providing the ability to cancel a parameter change
- Much better perf since not all params are shown at once
- Much more tablet friendly UI

Screen shots from Galaxy Tab...

Groups are on the left. Select a group to show params for the group:
screenshot_2015-04-28-16-00-08

Click on a parameter opens a panel for editing with all meta data shown:
screenshot_2015-04-28-16-00-19

Note 1: This Qml generates a binding loop error debug output which I believe is incorrect. I'm still looking at it. It doesn't prevent it from working.
Note 2: As I was typing this up I realized that I didn't put in reset to default and the rc map stuff in the new ui. I'll do that in a separate pull.

### DIFF
--- a/src/FactSystem/ParameterLoader.cc
+++ b/src/FactSystem/ParameterLoader.cc
@@ -87,14 +87,6 @@ void ParameterLoader::_parameterUpdate(int uasId, int componentId, QString param
                                     "value:" << value <<
                                     ")";
     
-#if 1
-    // This code is handy for testing re-request logic. It drops ever other param update.
-    static int increment = 0;
-    if (increment++ & 1) {
-        return;
-    }
-#endif
-    
     _dataMutex.lock();
     
     // Restart our waiting for param timer
--- a/src/QmlControls/ParameterEditor.qml
+++ b/src/QmlControls/ParameterEditor.qml
@@ -43,7 +43,7 @@ Rectangle {
 	QGCPalette { id: __qgcPal; colorGroupEnabled: true }
 	ScreenTools { id: __screenTools }
 	ParameterEditorController { id: __controller }
-	QGCLabel { id: __charWidth; text: "X"; visible: false }
+	QGCLabel { id: __textControl; text: "X"; visible: false }
 
 	readonly property real __leftMargin: 10
 	readonly property real __rightMargin: 20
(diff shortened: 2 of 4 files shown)


dogmaphobic:
It looks like the pixel size factor is way off for the Galaxy. The tool bar height is way too high. We do need to come up with some way to figure this out at runtime. I will work on looking at the default font point size and working the pixel size from there (the Connect/Disconnect button, which has its height based on the font size looks right). Once this gets merged in I will see how it compares with the Nexus.

DonLakeFlyer:
Forgot to mention that. The pixel factor thing isn't working for me. I tried using it but it was out of whack. I think the way to make things work is to base everything off of text width/heights. I do:

QGCLabel { id: __textControl; text: "X"; visible: false }
property real __textHeight: __textControl.contentHeight
property real __textWidth:  __textControl.contentWidth

Then base all my widths/heights on multiples of these values as well as using anchors as mush as possible to key everything flowing dynamically. Then I get proportional sizing on all OS.

dogmaphobic:
Yep, it looks like the way to go. The problem is figuring it out globally. That is, take the flight display for example. On its own, it doesn't have any text, and therefore, nothing to measure or a dimension to start with. That's why I needed an interface providing a starting point (ScreenTools).

I'm building a parallel *Application* completely based on QtQuick. There is one singleton class that gets registered and provides all the bindings. That will make it easy for the toolbar to *see* all the different views. This would also provide this *starting point* as the toolbar gets built first and it does have text to measure.

By parallel I mean a different project file that collects the same source code base but avoiding Qt Widgets. It will start rather limited as only a few things are full QML but we can grow from there. Once we get it all done, we can make it the main app and retire the old Qt Widget MainWindow.

DonLakeFlyer:
- Fixed binding loop problem
- Added Reset to Default
- Added Set RC to Param..
- Added Warning:
screen shot 2015-04-28 at 6 32 57 pm

DonLakeFlyer:
> I'm building a parallel Application completely based on QtQuick.

I'm not yet convinced that is a good approach. I'd rather whittle away until things are all gone. I'm not entirely convinced we will actually get there. We may still end up with a small hybrid set of things.

dogmaphobic:
Well, it could be the case. I want to at least take a look and see how things behave away from QtQuickWidget, which seems to be causing some problems I'm not sure we can solve without fixing Qt itself.

DonLakeFlyer:
What else is wrong other than the Chevron drawing hack?

dogmaphobic:
The main issue for me is that the *Dialog* QML type doesn't work. We have tons of dialogs. Also, the issue with Canvas 2D will most likely be there for Canvas 3D as well. In summary, the system was not designed to work this way. QtQuickWidget is a hack to allows us to do what we are trying to do but it doesn't fully work. A lot of these things expect the root to be a *ApplicationWindow* QML type to be the there.

DonLakeFlyer:
> The main issue for me is that the Dialog QML type doesn't work.

I was using the Dialog control right here in parameter editor and it was working fine for me. I just tossed it though with this pull to move to a more tablet style ui. I think in the long run Dialog's are not going to fit our ui style.

I'd bet the Canvas 2d thing is just a Qt bug and not something inherent to using QQuickWidget. We currently have a massive amount of nested QQuickWidget stuff in Setup views and it's working fine there with everything I've thrown at it.

dogmaphobic:
> I was using the Dialog control right here in parameter editor

Did you try it on Android? That's where I was having issues with. Either way, it's an experiment. The idea behind it all is valid for either implementation. Right now we have a MainToolbar.cc, a FlightDisplay.cc, etc, all providing bindings to their particular QML children but there is nothing that links them all. They are all very insulated from one another. I would like to have one main singleton interface, with all the signals, slots, etc. for mavs, system messages and all that. One source for all elements to bind to and communicate with one another.

DonLakeFlyer:
Nope, didn't try on Android. Isn't the fact that things like MainToolbar and FlightDisplay are loosely coupled a good thing? Long term the idea is that there are three main providers of information:
- AutoPilotPlugin
- Parameter Facts
- Mavlink Facts (NYI)
Those are the central switchboard for all information about the vehicle. Any example where cross-talk is valuable?

dogmaphobic:
It's not about been loosely coupled. It's the redundant work in their respective c++ interfaces. They all connect to the signals from those providers above and provide their own binding to the QML side. A lot of redundant work. Think about *ScreenTools*. It's like having a different one for each QML interface. We can combine these into one main *QGCEngine* interface and have all UI elements binding to it just like they do with *ScreenTools*.

Speaking of *ScreenTools*, we are creating new instances of it each each time we do an *import*. That's not very efficient. I've done a bit of reading on ```qmlRegisterSingletonType()``` and that's what we should be using and that's how this *QGCEngine* (or whatever we call it) should be implemented.

DonLakeFlyer:
> They all connect to the signals from those providers above and provide their own binding to the QML side.

Can you show me some examples? I'm not getting it.

dogmaphobic:
For instance, look at these two blobs of code from the toolbar and the flight display:

https://github.com/mavlink/qgroundcontrol/blob/master/src/ui/toolbar/MainToolBar.cc#L298

https://github.com/mavlink/qgroundcontrol/blob/master/src/ui/flightdisplay/QGCFlightDisplay.cc#L121

These signals, and their bound variables could be in one place and all QML UI elements could have access to them. That way, if you ever want to have access to one of these variables/states, you don't need to replicate it.

DonLakeFlyer:
Ok. Got it now. With Mavlink provider for FactSystem you'll be able to bind directly to most of these values just like you do parameters without needing a controller. The UAS signals and methods are mostly just values coming from mavlink messages. Then the rest of the crap in UAS that isn't covered by the Mavlink provider needs to move to AutoPilot plugin and you bind directly to that without a controller. There likely won't be a lot of those.

Maybe I should do Mavlink provider next instead of settings?

DonLakeFlyer:
In the long run I haven't figured out if UAS goes away completely as on object or not and is fully replaced by AutoPilotPlugin. Maybe.

dogmaphobic:
Let me work on trying to get ScreenTools to be a true singleton and use a proper ```qmlRegisterSingletonType```. Let's see how that works first and then we can tackle the others. Go ahead with the *Settings* as I have my head deep around this binding issues at the moment.

dogmaphobic:
> In the long run I haven't figured out if UAS goes away completely as on object or not and is fully replaced by AutoPilotPlugin. Maybe

That's another story, somewhat in parallel to the discussion above. Indeed, the current *state* of things is rather convoluted. Specially when it comes to link connections. I keep getting *Link Regained* messages even though nothing is going on...

DonLakeFlyer:
If you look at the internals of the qml controls they have a Text singleton that they use to get text heights just like we've been hacking in to size things. Would be nice to have a QGCText singleton as well that we could pull default font text width/height off of.

dogmaphobic:
It doesn't work. I will bet it has to do with the fact we're not truly a QML application:

I followed the instructions from here:

http://doc.qt.io/qt-5/qqmlengine.html#qmlRegisterSingletonType

The code I have is:

First, I removed the current registration:

```
qmlRegisterType<ScreenTools>("QGroundControl.ScreenTools", 1, 0, "ScreenTools");
```

Then I added this:

```
static QObject* screenToolsSingletonTypeProvider(QQmlEngine*, QJSEngine*)
{
    ScreenTools* pScTools = new ScreenTools();
    return pScTools;
}
```

And finally, I added the registration in ```void QGCApplication::_createSingletons(void)``` after all the other *regular* singletons are created:

```
//-- Create QML Singleton Interfaces
qmlRegisterSingletonType<ScreenTools>("QGroundControl.ScreenTools", 1, 0, "ScreenTools", screenToolsSingletonTypeProvider);
```

The ```screenToolsSingletonTypeProvider()``` function above is never called and the UI fails to create:

```
qrc:/qml/MainToolBar.qml:338:21: Type QGCLabel unavailable
qrc:///qml/QGroundControl/Controls/QGCLabel.qml:8:1: Element is not creatable.
QQmlInspector: No plugin available for view ' QGCQuickWidget '.
qrc:/qml/FlightDisplay.qml:488:5: Type QGCMapBackground unavailable
qrc:///qml/QGroundControl/FlightControls/QGCMapBackground.qml:159:5: Type QGCSlider unavailable
qrc:///qml/QGroundControl/FlightControls/QGCSlider.qml:83:9: Type QGCLabel unavailable
qrc:///qml/QGroundControl/Controls/QGCLabel.qml:8:1: Element is not creatable.
```

Note the lack of errors about a missing ScreenTool type. It's there. Just not created properly for some reason.

As I said, let me experiment with creating a true QML application and see how all these odd things behave there. At least I will know how it's supposed to work (or not).

DonLakeFlyer:
Something here: http://stackoverflow.com/questions/25403363/how-to-implement-a-singleton-provider-for-qmlregistersingletontype? Looks like registering a single is a bit funky.

dogmaphobic:
> If you look at the internals of the qml controls they have a Text singleton that they use to get text heights just like we've been hacking in to size things. Would be nice to have a QGCText singleton as well that we could pull default font text width/height off of.

There is a singleton type for that too! :) It would be perfect for this but it will probably fail the same way the one above did:

Instead of a *QObject* c++ class, you expose a QML type:
```
pragma Singleton
import QtQuick 2.0
Item {
    property int someComputedFontSize: 12
}
```
From the c++ side you register it like any other QObject interface:
```
qmlRegisterSingletonType(QUrl("qrc:/qml/GetMeTheFontSize.qml"), "QGCControls.GetMeTheFontSize", 1, 0, "ComputedFontSize");
```
From the QML code, you import just like the other ones:
```
import QtQuick 2.0
import QGCControls.GetMeTheFontSize 1.0
Item {
    id: root
    property int someValue: ComputedFontSize.someComputedFontSize
}
```

dogmaphobic:
> Something here: http://stackoverflow.com/questions/25403363/how-to-implement-a-singleton-provider-for-qmlregistersingletontype? Looks like registering a single is a bit funky.

That's a different problem. In the case above, the QQmlEngine never even calls the object creator. The way I see it, we don't have one QQmlEngine engine, but several, for each QuickWdget.

DonLakeFlyer:
Right but we can use the Qt singletons just fine from our Qml without problem. Text.WordWrap or example shows up in all our QQuickWidget instances. So there has to be a way to do it.

DonLakeFlyer:
If you search the Qt source you'll find a couple examples of C++ singletons. There is one in QtPositioning, positioning.cpp Do those not work in our Qml?

 qmlRegisterSingletonType<LocationSingleton  >(uri, major, minor, "QtPositioning", singleton_type_factory);

static QObject *singleton_type_factory(QQmlEngine *engine, QJSEngine *jsEngine)
{
    Q_UNUSED(engine)
    Q_UNUSED(jsEngine)

    return new LocationSingleton;
}

dogmaphobic:
I'm done for the day. I will look into it tomorrow. My hunch is that we do indeed have several QML engines running instead of one global one. That's why these things get out of sync and don't work. The Qt singletons exist within the QtCore environment and are not related to these problems. The QML singletons above exist within the context of one, single QQmlEngine. That's what the documentation talks about:

**NOTE: A QObject singleton type instance returned from a singleton type provider is owned by the QML engine.**

dogmaphobic:
> If you search the Qt source you'll find a couple examples of C++ singletons. There is one in QtPositioning, positioning.cpp Do those not work in our Qml?

I will look at it tomorrow...

DonLakeFlyer:
> My hunch is that we do indeed have several QML engines running instead of one global one.

We have one for each QQuickWidget. You probably need to jam something through here:

QGCQuickWidget::QGCQuickWidget(QWidget* parent) :
    QQuickWidget(parent)
{
    rootContext()->engine()->addImportPath("qrc:/qml");
}

rootContext()->engine() is the QMLEngine for that specific QGCQuickWidget.

dogmaphobic:
> rootContext()->engine() is the QMLEngine for that specific QGCQuickWidget.

Exactly. I think we should have one single one. That's why a Canvas in one doesn't get repainted if an event occurs on a different QMLEngine environment.

dogmaphobic:
Now I'm truly going to bed... :)

dogmaphobic:
Back to the sizing issue across multiple platforms, in looking at the Qt source code, I found this for when they decide a button size:
```
implicitWidth: Math.round(TextSingleton.implicitHeight * 4.5)
implicitHeight: Math.max(25, Math.round(TextSingleton.implicitHeight * 1.2))
```
So, somewhere, which I haven't yet found, there is a Text singleton providing a base size...

DonLakeFlyer:
qtbase/qml/QtQuick/Controls/Private/TextSingleton.qml

dogmaphobic:
> qtbase/qml/QtQuick/Controls/Private/TextSingleton.qml

That only defines an *updateSelectionItem()* function. I was looking for where the size is defined. On Mac OS, the height below returns **16** while on Android (Nexus 7) it returns **38**.
```
import QtQuick.Controls 1.2
import QtQuick.Controls.Private 1.0
...
    Component.onCompleted: {
        console.log('Text height:', TextSingleton.implicitHeight)
    }
...
```

dogmaphobic:
Come to think of it, with that number, we would be better off defining the font size in pixels and using that as a basis (for creating a factor). I just need to find a reliable (and cheap) way to find it without resorting to including private stuff.

DonLakeFlyer:
If you replicate what they did and create a singleton based off of QGCLabel then you should be able to pull those implicit values out of it or our default font.

PR #1520, opened 2015-04-29
Remove the "Simulate" feature

DonLakeFlyer:
This would simulate three vehicles flying around. A while ago it was useful for testing things. Now with the unit tests and MockLink it doesn't really add much value for testing. I'm removing it so I don' have to keep dragging it along as a roto-till various parts of the low-level QGC architecture.


DonLakeFlyer:
FYI: I've copied the simulated vehicle and mission code into MockLink in an ifdef'ed out section. I did that a while ago. I plan to implement Vehicles and Mission support in there.

DonLakeFlyer:
Need to keep MockLink class in release builds as well since it used by log player. Class only, not available in release build for connecting to.

PR #1521, opened 2015-04-29
Hacked version of ScreenToolsSingleton

DonLakeFlyer:
Just a hack to show it working for Gus. The issue is that since we have multiple engines and the singleton is coming from the C++ side of things you need to make sure the object ownership stays on the C++ side. Otherwise the Java GC will take ownership of the object and free it in one QmlEngine and it will get trashed in all other QmlEngine instances.

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -37,6 +37,7 @@
 #include <QPainter>
 #include <QStyleFactory>
 #include <QAction>
+#include <QQmlEngine>
 
 #include <QDebug>
 
@@ -208,6 +209,17 @@ QGCApplication::~QGCApplication()
     _destroySingletons();
 }
 
+static QObject* ScreenToolsSingletonFactory(QQmlEngine *engine, QJSEngine *jsEngine)
+{
+    Q_UNUSED(engine)
+    Q_UNUSED(jsEngine)
+    
+    ScreenTools* screenTools = new ScreenTools;
+    QQmlEngine::setObjectOwnership(screenTools, QQmlEngine::CppOwnership);
+    
+    return screenTools;
+}
+
 void QGCApplication::_initCommon(void)
 {
     QSettings settings;
@@ -273,6 +285,7 @@ void QGCApplication::_initCommon(void)
     // Register our Qml objects
     qmlRegisterType<QGCPalette>("QGroundControl.Palette", 1, 0, "QGCPalette");
     qmlRegisterType<ScreenTools>("QGroundControl.ScreenTools", 1, 0, "ScreenTools");
+    qmlRegisterSingletonType<ScreenTools>("QGroundControl.ScreenTools", 1, 0, "ScreenToolsSingleton", ScreenToolsSingletonFactory);
 	qmlRegisterType<ViewWidgetController>("QGroundControl.Controllers", 1, 0, "ViewWidgetController");
 	qmlRegisterType<ParameterEditorController>("QGroundControl.Controllers", 1, 0, "ParameterEditorController");
 }
--- a/src/VehicleSetup/VehicleSummary.qml
+++ b/src/VehicleSetup/VehicleSummary.qml
@@ -42,6 +42,11 @@ Rectangle {
 
     color: qgcPal.window
 
+Component.onCompleted: {
+    console.log("ScreenToolSingleton");
+    console.log(ScreenToolsSingleton.defaultFontPointSize);
+}
+
     Column {
         anchors.fill: parent
 


dogmaphobic:
Got it. I found what I was (not) doing that caused the UI to fail. Even though I had registered it as a singleton, it was still declared as a new instance in every QML file:
```
    property ScreenTools __screenTools: ScreenTools { }
```
I removed every one of those property declarations and renamed all __screenTools, or whatever those properties were named to *ScreenTools* and it all works. However, a new instance of *ScreenTools* is still created for each *QQuickWidget* we create (ToolBar, Flight Display, and the new QML Map View).

I've added a debug output to the *ScreenTools* constructor:
```
ScreenTools::ScreenTools()
{
    <...>
    qDebug() << "ScreenTools Instance Created";
}
```
And this is the resulting runtime output:
```
Starting /Users/gus/tmp/build-qgroundcontrol-Desktop_Qt_5_4_1_clang_64bit-Debug/debug/qgroundcontrol.app/Contents/MacOS/qgroundcontrol...
QML debugging is enabled. Only use this in a safe environment.
 QDir( "/Users/gus/Library/Preferences/QtProject" , nameFilters = { * },  QDir::SortFlags( Name | IgnoreCase ) , QDir::Filters( Dirs|Files|Drives|AllEntries ) ) 
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
After app delete
```
However, that's better than what we have now (in master) where we do create a zillion instances of the same thing:
```
Starting /Users/gus/tmp/build-qgroundcontrol-Desktop_Qt_5_4_1_clang_64bit-Debug/debug/qgroundcontrol.app/Contents/MacOS/qgroundcontrol...
QML debugging is enabled. Only use this in a safe environment.
 QDir( "/Users/gus/Library/Preferences/QtProject" , nameFilters = { * },  QDir::SortFlags( Name | IgnoreCase ) , QDir::Filters( Dirs|Files|Drives|AllEntries ) ) 
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
ScreenTools Instance Created
Orphaned log file count 0
After app delete
```

I still don't like having more than one instance. It defeats the purpose of a *Singleton* and does not allow us to have one *QGCEngine* type providing bindings to all UI elements. The same would be true with the FactSystem. Right now you only use it in the *Setup* UI. If you were to use it on a different UI (a new QQuickWidget), it would be a new, separate instance of it.

Not sure what the best solution would be. We can create a true, QtCore level singleton (like *MainWindow::instance()*) and have each QML *singleton* simply map to that or create just one QML Engine. The first one would solve the problem of having multiple instances but would not solve the sync issues (or lack thereof) between different QML engines (Canvas and such).

DonLakeFlyer:
It's creating more than once because the factory code I wrote is bad. Save the first one created in a static. Then if that's already created just hand the same one out again.

DonLakeFlyer:
I think creating a single Qml Engine would really confuse Qt. It don't think it's meant to run that way. I found the QQmlEngine::setObjectOwnership thing by searching for someone using C++ singletons in multiple QQuickWidgets. I found someone not using multiple QQuickWidgets, but multiple top-level Qml windows. This ends up being exactly the same usage pattern as multiple QQuickWidgets. Each top level Qml window has it's own distinct QmlEngine.

There is only a single set of FactSystem objects associated with each connected UAS no matter how many times you use it in how many QQuickWidgets. The only duplication is the thin FactBinder wrapper object which is the object behind the Fact Qml item. This connects to the real single instance of the Fact in the system. I created the FactBinder thing because I was having problems passing Fact objects directly into the Qml context. Now I know the reason for that was the same QQmlEngine::setObjectOwnership thing. I was passing the same Fact into multiple QmlEngine contexts and one of them was free'ing and then the other one would crash. Not sure if I'll switch the mechanism back or not. I kind of like the new way where you define Fact items in the Qml instead of referencing them directly from an autopilot instance.

The Setup view is multiple QQuickWidgets. The button bar is a separate one as is each individual setup screen. The parameter editor dock widget is also a QQuickWidget. All connecting to the single instance of the FactSystem stuff without problems.

dogmaphobic:
I'm not in front of the computer right now so I cannot test this. If you pass an existing object pointer to subsequent calls, how do you deal with object destruction? When any of those QQuickWidget *parents* is destroyed, it would be deleting the object, invalidating all other references, no?

DonLakeFlyer:
There are two cases where FactSystem objects are passed across the Qml/C++ boundary:

First:
The FactBinder object is created on the Qml side of things and the Qml Engine owns it and destroys it. It is a thin wrapper which re-routes all property binding to the real Fact object on the C++ side. FactBinder finds the right Fact object from the C++ side and references it internally. The FactBinder destructor is in charge of deciding what it wants to destroy. It just has a pointer to the Fact. When Qml destroys the FactBinder object the FactBinder destructor (their actually isn't one since it doesn't need to do anything special) doesn't touch the C++ side Fact object.

Second:
The single AutoPilotPlugin instance is pushed into the QmlEngine context as a context property. You then reference it using the autopilot.foo in your Qml. As a context property, Qml does not take ownership of this object. Hence when that specific QmlEngine shuts down it doesn't do anything strange with the AutoPilotPlugin object. It just drops the reference.

The only detriment to using multiple QQuickWidget's is the memory overhead of multiple QmlEngines running. You also can't create ui which crosses the widget boundaries easily. Say to overlay some sort of panel over the whole window with opacity. Hence the hybrid conversion approach is to whittle that down to as few QQuickWidgets as possible. Possibly only one in the end but I don't know. 

Remaining hybrid conversion approach:
- For Setup convert RC Cal to Qml. Then I can collapse that down to a single QQuickWidget for the entire view.
- All remaining top level views converted to Qml running inside QQuickWidgets
- All dock widgets converted to Qml running inside QQuickWidget. ParameterEditor is already an example of this. I already have the framework in place for this in the ViewWidget* code and qml under src/ViewWidgets.
- Create new dock widget system which can run dock widget things natively within a Qml view instead of to the side like now.
- Now you have everything you need to convert the top level window fully to Qml.

dogmaphobic:
It makes sense as we need to work one element a time before everything can be switched to one single QML application. That's what I meant by creating an experimental QML project to see how it all behaves, which is also what I meant by one single QML Engine. An application where main.cc creates one QML Engine, points its root QML file and goes from there. No QQuickWidget. In other words, the dynamics of the application will be quite different.

As far as the singleton saga goes, it won't work until we have this one single QML Engine. For the time being, the singleton object will have to be at a lower level and only a wrapper is passed to QML (such as you did with the Fact System).

I tried this just for kicks:
```
static QObject* screenToolsSingletonFactory(QQmlEngine*, QJSEngine*)
{
    static ScreenTools* screenTools = NULL;
    if(!screenTools) {
        screenTools = new ScreenTools;
        QQmlEngine::setObjectOwnership(screenTools, QQmlEngine::CppOwnership);
    }
    return screenTools;
}
```
And it obviously crashes once one of the QQuickWidget objects is deleted, as the singleton ScreenTools is no longer valid for the remaining QQuickWidgets.

One of the reasons for all this is that we are getting rather obese and I'm looking at inefficiencies. I have ScreenTools fixed and I will be pushing a PR soon. The new *ScreenTools* gets everything off one *ScreenToolsSingleton::instance()*. We then have one true instance and *n* wrapper instances where *n* is the number of running QQuickWidget objects. That's a lot better than the current case where each QML reference to *ScreenTools* has one full instance of it.

This is an exercise in looking into these redundant instances issue. *ScreenTools* is a rather simple interface. Once it's in place, I can look at other, heavier ones.

dogmaphobic:
Actually, now that I look at it, it's pointless to go through all the hassle of a true singleton for ScreenTools. It won't save much. One *singleton* per QQuickWidget is just fine :)

DonLakeFlyer:
> pointless to go through all the hassle of a true singleton for ScreenTools

Just a few bytes. The wrappers are probably just as big as the real object.

PR #1523, opened 2015-04-29
Singleton screen tools

dogmaphobic:
This is a full implementation of Don's experiment in #1521.

I've turned ScreenTools into a *singleton* where only one instance of it exists for each instance of QQuickWidget. There will still be a same number of instances as there are instances of QQuickWidget objects but they are a few compared to the zillion they were before.

You still import it the way it's always been done:
```
import QGroundControl.ScreenTools 1.0
```
Where you would normally declare an instance:
```
        property ScreenTools __screenTools: ScreenTools { }
```
and reference it by this property name:
```
        font.pointSize: __screenTools.fontPointFactor * (20);
```
Now you don't need to create the instance at all. You just use it by name:
```
        font.pointSize: ScreenTools.fontPointFactor * (20);
```

### DIFF
--- a/src/AutoPilotPlugins/PX4/FlightModesComponent.qml
+++ b/src/AutoPilotPlugins/PX4/FlightModesComponent.qml
@@ -36,7 +36,6 @@ Item {
     Loader {
         property FlightModesComponentController controller: FlightModesComponentController { }
         property QGCPalette qgcPal: QGCPalette { colorGroupEnabled: true }
-        property ScreenTools screenTools: ScreenTools { }
         property bool loading: true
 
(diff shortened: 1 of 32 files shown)


dogmaphobic:
Is it OK for me to merge this? I'm working on mapping (waypoints) and I've consolidated all Mav management into one *Singleton* class (like ScreenTools) as this is shared among the toolbar, the flight display and now the map display. I'm accumulating way too many changes and merging will be gnarly.

DonLakeFlyer:
Sure. I'm not on much tonight. Sorry. If you really need something just merge it in.

PR #1524, opened 2015-04-29
New Custom Command Widget

DonLakeFlyer:
This allows you to build your own custom command widgets using Qml. You can access the new widgets from the menu.

You must be connected to a vehicle to get this:
screen shot 2015-04-29 at 12 25 33 pm

Here is the example file running:
screen shot 2015-04-29 at 12 26 04 pm

If youre Qml has errors, it shows up in the widget like this:
screen shot 2015-04-29 at 12 26 53 pm

Here is the example Qml file from the wiki:
```javascript
// This is an example Custom Command Qml file. You have full access to the entire Qml language
// for creating any user interface you like. From the ui you can affect the following changes
// with respect to your vehicle:
//    1) Sending COMMAND_LONG commands out over mavlink using QGCButton control
//    2) Modifying parameters
//
// When developing custom Qml file implementations. You must restart QGroundControl to pick up
// the changes. You need to do this even if you select Clear Qml file. Not sure what at the this
// point. Qt must be caching the files somewhere.

import QtQuick 2.2

import QGroundControl.Controls 1.0
import QGroundControl.FactSystem 1.0
import QGroundControl.FactControls 1.0

Item {

    // Your own custom changes start here - everything else above is always required

    Column {
        // The QGCButton control is provided by QGroundControl.Controls. It is a wrapper around
        // the standard Qml Button element which using the default QGC font and color palette.

        QGCButton {
            text: "Set Home to current position"
            onClicked: controller.sendCommand(179, 50, 0, 1, 0, 0, 0, 0, 0, 0)
        }

        // The FactTextField control is provides by GroundControl.FactControls. It is a wrapper
        // around the Qml TextField element which allows you to bind it directly to any parameter.
        // The parameter is changed automatically when you click enter or click away from the field.
        // Understand that there is currently no value validation. So you may crash your vehicle by
        // setting a parameter to an incorrect value. Validation will come in the future.

        // Be very careful when referencing parameters. If you specify a parameter which does not exist
        // QGroundControl will warn and shutdown.

        FactTextField {
            fact: Fact { name: "MAV_SYS_ID" }
        }
    }

    // Your own custom changes end here - everything else below is always required
}
```

### DIFF
--- a/QGCApplication.pro
+++ b/QGCApplication.pro
@@ -354,6 +354,8 @@ HEADERS += \
     src/ui/WaypointList.h \
     src/ui/WaypointViewOnlyView.h \
     src/ViewWidgets/ParameterEditorWidget.h \
+    src/ViewWidgets/CustomCommandWidget.h \
+    src/ViewWidgets/CustomCommandWidgetController.h \
     src/ViewWidgets/ViewWidgetController.h \
     src/Waypoint.h \
 
@@ -489,6 +491,8 @@ SOURCES += \
     src/ui/WaypointList.cc \
     src/ui/WaypointViewOnlyView.cc \
     src/ViewWidgets/ParameterEditorWidget.cc \
+    src/ViewWidgets/CustomCommandWidget.cc \
+    src/ViewWidgets/CustomCommandWidgetController.cc \
     src/ViewWidgets/ViewWidgetController.cc \
     src/Waypoint.cc \
 
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -34,6 +34,7 @@
         <file alias="QGroundControl/Controls/ParameterEditor.qml">src/QmlControls/ParameterEditor.qml</file>
 
         <file alias="ParameterEditorWidget.qml">src/ViewWidgets/ParameterEditorWidget.qml</file>
+        <file alias="CustomCommandWidget.qml">src/ViewWidgets/CustomCommandWidget.qml</file>
 
         <file alias="SetupViewButtonsConnected.qml">src/VehicleSetup/SetupViewButtonsConnected.qml</file>
         <file alias="SetupViewButtonsDisconnected.qml">src/VehicleSetup/SetupViewButtonsDisconnected.qml</file>
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -63,6 +63,7 @@
 #include "QGCLoggingCategory.h"
 #include "ViewWidgetController.h"
 #include "ParameterEditorController.h"
+#include "CustomCommandWidgetController.h"
 
 #ifdef QGC_RTLAB_ENABLED
 #include "OpalLink.h"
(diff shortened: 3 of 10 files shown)


DonLakeFlyer:
There is actually quite a bit more you can do with QGC provided controls but at this point I don't want to document/expose too much of that until it is more baked and bullet proof. But if you're adventurous it's all there for you.

DonLakeFlyer:
Important to note this strangeness:
// When developing custom Qml file implementations. You must restart QGroundControl to pick up
// the changes. You need to do this even if you select Clear Qml file. Not sure what at the this
// point. Qt must be caching the files somewhere.

At some point I'll try to figure out what's going on there. Makes developing a new control a little bit of a pain to keep going back and forth to fix errors.

DonLakeFlyer:
@dogmaphobic This is another example of needing a controller that will eventually go away once the Mavlink Fact provider is done.

DonLakeFlyer:
And here's the answer to the file loading thing:
http://stackoverflow.com/questions/19604552/qml-loader-not-shows-changes-on-qml-file

I'll fix it.

dogmaphobic:
This is great (and nuts). Can you imagine life without stackoverflow? :)

DonLakeFlyer:
@mattmccrink Let's move the discussion to here. When you say your getting file not found that is showing up as an error right in the ui (like the visual above) right after you pick the file? What OS are you using?

mattmccrink:
It shows up as in your example. I have the "test.qml" file located in the release directory. I am using Windows 8.1 and MSVS 2013. It looks like the path may not be concatenated correctly (?). 
qgc_error

DonLakeFlyer:
It's a problem with the drive letter on Windows. Let me figure out how to fix that.

DonLakeFlyer:
Found another bug while working on this. If you get an error, then clear the qml the error panel is still showing being the warning panel.

DonLakeFlyer:
File load problem is fixed

DonLakeFlyer:
Fixing the quit/restart problem require access to QQmlEngine from the controller. That would require some strange reaching around things to make work. So I'm going to leave it as is for now.

DonLakeFlyer:
Waning panel problem is fixed now.

@mattmccrink Let me know is this is working for you. And I'll merge when you give me the clear.

DonLakeFlyer:
Pretty sure this is good now.

mattmccrink:
Apologies, just got to it. Looks good.

LorenzMeier:
The docs here are pretty good, but the ones on the QGC website are not. Does this mean we should change the wiki system to something you would be willing to edit more casually?

DonLakeFlyer:
You mean like the detailed images and things in this pull versus the wiki? I specifically haven't put much much on the wiki because at this point I'm trying to keep this capability low key. Reason being is that the FactSystem isn't full hardened to end users hacking away at it. Until I finish that up with more work on the parameter missing problem and friends I don't want to go broadly with this. Hence the on purpose lack of detail.

NiqueIt:
I need to set MAV_CMD_NAV_GUIDED_ENABLE. As the custom widget is gone, I assume to do it with the custom command and a qml-file. Can you help with syntax or do an example?
Reason is: I want to control pixhawk only by mavlink and no longer by RC.

NiqueIt:
Ey, seams not to be that hard. Should make this sense:

    import QtQuick 2.2

    import QGroundControl.Controls 1.0
    import QGroundControl.FactSystem 1.0
    import QGroundControl.FactControls 1.0

    Item {
        Column {
            QGCButton {
                text: "Enable external controller"
                onClicked: controller.sendCommand(92, 1, 0, 0, 0, 0, 0, 0, 0, 0)
            }
        }
    }

I'm unsure with Param 1, it is described as: 	On / Off (> 0.5f on) (see here: https://pixhawk.ethz.ch/mavlink/)

mattmccrink:
An example was posted at:
http://www.qgroundcontrol.org/custom_command_qml_widgets

You would need to replace

QGCButton {
            text: "Set Home to current position"
            onClicked: controller.sendCommand(179, 50, 0, 1, 0, 0, 0, 0, 0, 0)
        }

with

QGCButton {
            text: "Enable Nav guidance"
            onClicked: controller.sendCommand(92, 1, 0, 0, 0, 0, 0, 0, 0, 0)
        }

The exact enumerations are listed under MAV_CMD
https://pixhawk.ethz.ch/mavlink/

mattmccrink:
Beat me to it. Might make two buttons for enable/disable or a persistent
variable to toggle the executed command.

NiqueIt:
Thanks, glad to be some minutes faster with same result ;-)

inta495:
Is it possible to use custom command to show any value such as attitude or other custom MAVlink message?

DonLakeFlyer:
No it is not.

PR #1529, opened 2015-05-01
Working basic mavlink ftp file upload

crashmatt:
Simple file upload button added.  It does not yet report success or failure.  The file list has to be manually refreshed.

There are some potential minor changes to the PX4 code.  These changes have been discussed with the dev.  This uses file create.  Files can't be overwritten.  There is no file delete function in qgc yet.

Overall low risk on this change.


DonLakeFlyer:
I'll get to look at this, this weekend. A bit swamped with other things. Also TeamCity errors need to be cleaned up.

crashmatt:
I have tried a few ways to fix that error but I'm just not getting it.  It would help if my compiler settings reported the same errors as the teamcity ones.

crashmatt:
TeamCity is now checking the wrong code.  It is reading this line:

if( *((uint32_t*)writeAck->data) !=_writeSize){

but my latest is:

+    const uint32_t* p_writtenSize = (uint32_t*) &writeAck->data[0];
     if( *((uint32_t*)writeAck->data) !=_writeSize){

but I still don't know if that will keep the compiler happy.

DonLakeFlyer:
You're going to need to do the same thing the OpenFile command does to pull information out of &data[0]. Which is to update the Request union to include your usage. This tells the compiler what you are doing and hence it won't get confused with aliasing.

    struct Request
    {
        struct RequestHeader hdr;

        // We use a union here instead of just casting (uint32_t)&payload[0] to not break strict aliasing rules
        union {
            // The entire Request must fit into the payload member of the mavlink_file_transfer_protocol_t structure. We use as many leftover bytes
            // after we use up space for the RequestHeader for the data portion of the Request.
            uint8_t data[sizeof(((mavlink_file_transfer_protocol_t*)0)->payload) - sizeof(RequestHeader)];
			
            // File length returned by Open command
            uint32_t openFileLength;
        };
    };

src/uas/QGCUASFileManager.cc
```diff
@@ -199,6 +199,99 @@ void QGCUASFileManager::_listAckResponse(Request* listAck)
...
+    request.hdr.offset = _writeOffset;
+
+    if(_writeFileSize -_writeOffset > sizeof(request.data) )
+        _writeSize = sizeof(request.data);
+    else
+        _writeSize = _writeFileSize - _writeOffset;
+
+    request.hdr.size = _writeSize;
+
+    // memcpy this?   _writeFileAccumulator.mid(_writeOffset, _writeSize), _writeSize);
```
DonLakeFlyer: memcpy does make more sense here

DonLakeFlyer:
Could you add something like _closeReadSession that will free the memory associated with he write accumulator when the write completes or errors out? Would like to get rid of that memory hanging around when not in use.

crashmatt:
I fixed that lot and gave it a quick functional test.  I did not get to check the failure modes so I can't be sure they work properly.

crashmatt:
TeamCity doesn't seem to be tracking the change yet.  Lets see what happens.

DonLakeFlyer:
Just finished. Failing windows build. Not sure how "not" compiles in other OS. Maybe someone has it defined somewhere as "!".

dogmaphobic:
*not* is new (c++11 or so) and it seems MS doesn't implement it.

DonLakeFlyer:
I'm still on C++ 1985!

LorenzMeier:
I don't think we should get carried away with C++11 right now. Its a cross-platform nightmare still at this point, so all not calls should be replaced with !.

crashmatt:
Well I just replaced not with ! so lets see how that goes.

DonLakeFlyer:
> Could you add something like _closeReadSession that will free the memory associated with he write accumulator when the write completes or errors out? Would like to get rid of that memory hanging around when not in use.

Can you do this as well?

crashmatt:
Don, I didn't add a specific escape function but it does clear the badly named _writeFileAccumulator on all of the exit points I could find.

DonLakeFlyer:
Perfect. Let me take one more look through this and I'll merge.

DonLakeFlyer:
Thanks a ton for the pull Matt.

PR #1530, opened 2015-05-01
Added MavManager QML Interface

dogmaphobic:
The *Main Flight Display*, the *ToolBar*, and the still experimental *Plan View*, all collected and exposed real time Mav data to their respective QML environment. This meant a lot of redundant code. They each made connections to the various interfaces (UASManager, UASInterface, UASWaypointManager, etc.) and reflected these signals as properties to the QML side.

To reduce the amount of code, I've create one singleton *MavManager* that any QML code can import and use. It exposes all real time data to any QML that imports it (it will be interesting to see what can be done from the custom QML code added by Don's #1524 PR).

This was done while I was in the process of beginning to figure out how to render waypoints in the new Plan View. This PR also has the initial work of collecting the waypoint list from the PX4 (all done by the MavManager and available to anyone) and rendering them on the Map Display. As this is inherent to the Map Display, you can optionally enable *Show Waypoints" in the Main Flight Display as well (provided you have the Map Background enabled, of course).

For now, only the very basics are shown. I've been working primarily on the underlying code for collecting data and making it available to the map display. The waypoint markers are still very simplistic, hard coded little squares showing the waypoint ID and that's all.

Note that for now, the only things that are shown are the waypoints loaded from the PX4. You can't add or edit anything yet.

screen shot 2015-05-01 at 12 07 13 am


dogmaphobic:
I've added the ability to move a waypoint around. This is only updating the path on the screen. It is not being updated internally.

If you have *Show Waypoints* enabled on the main flight display, they are always read only. You can only edit them within the Plan View.

DonLakeFlyer:
FYI: This brings up an interesting point which is that Missions should be a Fact provider as well. That would provide a common Fact api to access parameters, mission items, mavlink telemetry. I'll get working on those.

DonLakeFlyer:
Also QGC prefix on classes is only needed for classes which override and existing Qt class. Otherwise it just adds noise.

dogmaphobic:
> Also QGC prefix on classes is only needed for classes which override and existing Qt class. Otherwise it just adds noise.

Yes, I still get confused when naming these things, not to mention the inconsistencies (MavManager and QGCFlightDisplay). I will fix it...

dogmaphobic:
> FYI: This brings up an interesting point which is that Missions should be a Fact provider as well. That would provide a common Fact api to access parameters, mission items, mavlink telemetry. I'll get working on those.

That would make things a bit more consistent with the existing setup interface. In the mean time, I made the *Waypoint* class accessible from QML along with the interfaces in MavManager to access it. How do signals work from the Fact system? Is the value binding automatic?

DonLakeFlyer:
A Fact connected to a Fact control will automatically update when the value changes out from under it through the valueChanged signal. Same thing if not using fact control and your are bound to the value property of a Fact.

    Q_PROPERTY(QVariant value READ value WRITE setValue NOTIFY valueChanged USER true)
    Q_PROPERTY(QVariant valueString READ valueString NOTIFY valueChanged)

You can also connect to the valueChanged signal in C++ if you need to do things there as well. With all these things being Facts you get to use all the FactControls plus have meta data access as well. Also automatic validation against the metadata once I implement that.

PR #1532, opened 2015-05-03
New burst download for file transfer

DonLakeFlyer:
This implements a burst mechanism for file download. The firmware (in a new pull) will stuff as many packets across the pipe as available space will allow in response to each read request. Ends up being a 5x speed improvement for download over usb. Over radio speed stays the same.

Also note that ftp over radio is very flat for reason which I haven't determine. If you turn off mavlink framing on the radio you end up getting much better results. But at this point I don't know why.

I plan to get @crashmatt pull #1529 in first. Since this pull will conflict with that one. Then I"ll clean this one up again.


DonLakeFlyer:
Companion firmware pull: https://github.com/PX4/Firmware/pull/2120

DonLakeFlyer:
Now that's Matt pull is through. I'll work on getting this one through. I can't believe it will actually merge! I need to double check everything.

DonLakeFlyer:
Retesting: List/Download working. Upload not working. Looking at it.

crashmatt:
Don,
I did a modification to the PX4 end. I am having a bit of trouble with my
PX4 builds right now.  When I get that fixed I can retest upstream.

/Matt

LorenzMeier:
Hi Matt,

To not block integration: Could you share the diff in the MAVLink app? We're trying to wrap up file transfer support and the missing diff is either blocking this or you will end up in the same spot as previously: Your changes are not maintained and you get into trouble keeping up with the mainline development. We're trying to prevent that and keep your development branch synced in.

What kind of build issue are you facing right now?

DonLakeFlyer:
Master QGC + Master Firmware: Upload doesn't work.

QGCUASFileManager: Error "Nak received creating file, error: command failed"
QGCUASFileManager: Error "Nak received, error: invalid session"

There is also a bug somewhere which is leaking sessions. Not sure when that came in. It may have always been there. I've verified that download is working. I'm going to merge this in. 

@crashmatt: Can you do your testing/further changes against master after this latest commit passes TC. I'm backed up behind these significant changes in FileManager so don't want this to get into a massive merge conflict again with this.

DonLakeFlyer:
I'm in a massive merge from hell situation. So I'm going to slam this through which could likely break both download and upload even more. Then I'll fix it after the fact as needed. Matt: So we don't step on each other's toes. Don't worry about upload. I'll get it working with whatever is needed on both QGC and firmware side. It's highly likely that I just broke it even more.

DonLakeFlyer:
Ok. Passed TC. Now I'm merging and I'll clean up the mess I've potentially made in download/upload.

crashmatt:
Lorenz,
I was having a moment where nothing but upstream would build and run.  I
seem to have that fixed now.

When I did my last mavlink-ftp test it was on a AUAV-X2 with unknown
firmware.  It might have been my branch and it might have been upstream.
Right now I have lots of branches which spontaneously broke for unknown
reasons.  I will retest this and put in a pull request (if needed) as soon
as it is safe to do so.

/Matt

crashmatt:
I retested and it works with the PX4 upstream mavlink_ftp.cpp
There are no gui notifications yet.  There is no automatic file list update yet. You need to  refresh the file list afterwards.  I will fix the gui stuff later.

PR #1537, opened 2015-05-05
Much better handling of missing parameters

DonLakeFlyer:
This is a much better (and somewhat more complicated) approach to handling missing facts in QGC without having to shutdown the app immediately. You still end up with unusable portions of the ui but it should at least provide you with the ability to orderly retrieve your vehicle and disconnect.

The main concept is to isolate the havoc that a missing parameter causes to one or more subsections of the ui. In order to do this I introduce the concept of a FactPanel in qml. A FactPanel will catch missing parameters, notify the user and disable it's ui. Here is an example of using a FactPanel when you do not need a C++ controller to go along with it:

```javascript
FactPanel {
    Fact { 
        id: sysIdFact
        name: "MAV_SYS_ID"
        onFactMissing: showMissingFactOverlay(name) 
    }
    ...
}
```

So instead of using a top level Rectangle or something like that you use a FactPanel. Then when you declare Facts, you connect the onFactMissing signal to the showMissingFactOverlay method which is from FactPanel.

In this pull I have converted all vehicle summary widgets to using FactPanel. Take a look at [Airframe|Power|...]ComponentSummary.qml for examples. This is what happens when you hack the code to return that all requested parameters are missing:

screen shot 2015-05-05 at 10 43 20 am

You'll get a message box telling you all the params that are missing. And warning that you are in a bad state and you should do something about it as soon as you can. Also if you look behind the message box you will see that the summary items have their ui disabled and show which facts they are missing as well. After you click Ok on the message box the remainder of QGC continues to be operational.

The above example handles missing parameters from Qml code. You also need to be able to handle missing parameters from C++ code which is implemented as a controller for Qml code. In that case you need to derive your controller from the new FactPanelController class. The FactPanelController base class work in conjunction with a FactPanel Qml object to handle missing facts on the C++ side of things.

In this case your Qml would look like this:

```javascript
FactPanel {
    id: panel

    MyController {
        id: controller
        factPanel: panel
    }

    Fact { 
        id: sysIdFact
        name: "MAV_SYS_ID"
        onFactMissing: showMissingFactOverlay(name) 
    }
    ...
}
```

MyController is derived from FactPanelController. Notice how you need to pass in the FactPanel to it using the factPanel property.

Then in your C++ code you can use the following methods:

```javascript
    /// Checks for existence of the specified facts
    /// @return true: all facts exists, false: facts missing and reported
    bool FactPanelController::_allFactsExists(QStringList factList);
    
    /// Report a missing fact to the FactPanel Qml element
    void FactPanelController::_reportMissingFact(const QString& missingFact);
```

You have to write your controller code in a way that it doesn't blow up if a fact is missing. It should report a missing fact and then fail gracefully. By reporting a missing fact it will cause the FactPanel to disable as shown in the first example. I have converted the Airframe component setup screen over to this mechanism. See AirframeComponent.qml and AirframeComponentController.h/cc.

The code will also detect if you use a FactPanel but do not set the factPanel property and show you this:
screen shot 2015-05-05 at 11 30 43 am

If you want to report missing parameters from standalone C++ code not associated with ui. You can use:

```javascript
    /// Used to report a missing Fact. Warning will be displayed to user. Method may be called
    /// multiple times.
    void QGCApplication::reportMissingFact(const QString& name);

    // Usage
    qgcApp()->reportMissingFact("BOGUS_FACT");
```

### DIFF
--- a/QGCApplication.pro
+++ b/QGCApplication.pro
@@ -635,7 +635,8 @@ SOURCES += \
 # Fact System code
 
 INCLUDEPATH += \
-    src/FactSystem
+    src/FactSystem \
+    src/FactSystem/FactControls \
 
 HEADERS += \
     src/FactSystem/Fact.h \
@@ -644,6 +645,7 @@ HEADERS += \
     src/FactSystem/FactSystem.h \
     src/FactSystem/FactValidator.h \
     src/FactSystem/ParameterLoader.h \
+    src/FactSystem/FactControls/FactPanelController.h \
 
 SOURCES += \
     src/FactSystem/Fact.cc \
@@ -652,6 +654,7 @@ SOURCES += \
     src/FactSystem/FactSystem.cc \
     src/FactSystem/FactValidator.cc \
     src/FactSystem/ParameterLoader.cc \
+    src/FactSystem/FactControls/FactPanelController.cc \
 
 # Android
 
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -9,6 +9,7 @@
         <file alias="QmlTest.qml">src/QmlControls/QmlTest.qml</file>
 
         <file alias="QGroundControl/FactControls/qmldir">src/FactSystem/FactControls/qmldir</file>
+        <file alias="QGroundControl/FactControls/FactPanel.qml">src/FactSystem/FactControls/FactPanel.qml</file>
         <file alias="QGroundControl/FactControls/FactLabel.qml">src/FactSystem/FactControls/FactLabel.qml</file>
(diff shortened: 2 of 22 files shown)


DonLakeFlyer:
@dogmaphobic I'm going to convert over all the Setup stuff. I'd wait to use this until I've finished that so I can see what sort of holes still need to be plugged in the apis.

DonLakeFlyer:
Missing a couple things:
- The C++ apis no longer assert if you request a missing fact. They will just return NULL.
- There is no longer a "Panic Shutdown" concept.

Issue #1540, opened 2015-05-06
Prevent reverse lookup for IP UDP hosts


AndreasAntener:
@DonLakeFlyer 
In https://github.com/mavlink/qgroundcontrol/blob/master/src/comm/UDPLink.cc#L299 QHostInfo implicitly does a reverse lookup for IP hosts (or a normal lookup for name hosts). Most of the time we don't have DNS servers accessible if we're testing in private networks, which means QGC always blocks a certain amount of time per host on startup and on changing the host addresses.

I looked very briefly in the API for QHostInfo but couldn't spot a way to prevent the reverse lookup for the IP-address case (which is completely unnecessary..), maybe you have an idea?

DonLakeFlyer:
I know next to nothing about TCP/IP and DNS. So can't help much.

@dogmaphobic Gus: Any ideas?

dogmaphobic:
Yes, the idea is to get an IP address if a host name is passed. At the lower level, if you pass an IP address (instead of a host name), the function simply returns it. Qt is trying to be too smart and doing a reverse lookup. That's what's causing the DNS lookup:

*"If you pass a literal IP address to name instead of a host name, QHostInfo will search for the domain name for the IP (i.e., QHostInfo will perform a reverse lookup). On success, the returned QHostInfo will contain both the resolved domain name and IP addresses for the host name."*

I will change the code and check for an IP address myself instead of asking Qt.

dogmaphobic:
@AndreasAntener, can you please try this? (#1541) My UDP testing ability is limited to one single connection using a redirector.

PR #1541, opened 2015-05-06
Skipping DNS lookup if not needed.

dogmaphobic:
This is a fix for issue #1540.
I'm trying to detect an ip address (x.x.x.x) and using it directly instead of asking Qt, which will do a reverse DNS lookup each time.

### DIFF
--- a/src/comm/UDPLink.cc
+++ b/src/comm/UDPLink.cc
@@ -39,6 +39,36 @@ This file is part of the QGROUNDCONTROL project
 #include "QGC.h"
 #include <QHostInfo>
 
+static bool is_ip(const QString& address)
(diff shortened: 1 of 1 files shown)


AndreasAntener:
Cool, thanks :). I can confirm doesn't block anymore (if you'd like to test it, just set your hosts DNS server to some unreachable IP address). However, I couldn't try UDP connection yet, need to do that tomorrow if you think it's necessary.

dogmaphobic:
I already tested with a UDP connection. I just wanted to make sure it solved what you were seeing. I'll merge it in and we can see if it needs anything else later.
Thanks!

Issue #1544, opened 2015-05-07
The selected parameter changes if you click somewhere in the info panel


AndreasAntener:
To clarify, if you select a parameter, the info panel opens. If you then click somewhere below or above the text field, it might select a different parameter. Seems that the click event is propagated to the list in the background.
It's actually cool that you can select a different parameter from the at this point greyed-out list ;), but if you miss the text field and don't notice it you might change the wrong parameter.

DonLakeFlyer:
Fixed

Issue #1543, opened 2015-05-07
Cannot select airframe


tumbili:
@DonLakeFlyer Maybe I'm just too stupid but I cannot select the airframes at the bottom because I cannot scroll down in the window. Have tried everything but I can't seem to manage.

dogmaphobic:
Yes, I have the same issues on my tablet. Don is working on consolidating all those setup tools into one, consistent widget. Once that's done, you will be able to scroll down.

tumbili:
@dogmaphobic Cool, sounds good.

DonLakeFlyer:
Fixed

tumbili:
@DonLakeFlyer Thanks Don, will try it tomorrow!

Issue #1539, opened 2015-05-06
Remove assertion for unknown airframe type


AndreasAntener:
The assertion for finding a SYS_AUTOSTART match makes debugging QGC hard if we're using a custom airframe config.

https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/PX4/AirframeComponentController.cc#L88
IMHO you could remove the assertions there :)

DonLakeFlyer:
The reason for that is the airframe config code right now doesn't really know how to handle a custom airframe config. It's likely to do strange things, hence the assert. The bug is to support customer airframe config and remove the assert.

LorenzMeier:
I know I'm stating the obvious, but just a text box / spin box with the current number (and accepting a number) would do. It could show instead of the normal UI, along with a short text stating that custom configs can be set, but are not supported in the UI.
Or even simpler: Hide the complete pane, let the user know that he has to use the param editor and can re-enable the GUI config by setting the param to 0 / default.

AndreasAntener:
From a firmware API perspective there isn't much to handle in the airframe configuration, maybe 2 parameters? Should be fairly easy to just show the airframe as not-configured.

But I get your point, I just disagree. I know you are making a huge effort to bring QGC towards the end user, but the developers should at least still be "allowed" to use it ;). This is why it would be nice to have single functions (views) fail fast instead of the complete QGC quitting (as when there are missing parameters).

AndreasAntener:
Oh, sorry Lorenz, didn't see your post before, yes, I was thinking too, if you open that panel you'd just get a text that tells you to look in the parameters.

LorenzMeier:
Don already did that 8)
https://github.com/mavlink/qgroundcontrol/pull/1537#event-298369182

AndreasAntener:
Holy ** that looks awesome ;) will try

DonLakeFlyer:
Fixed

PR #1549, opened 2015-05-09
Airframe config

DonLakeFlyer:
This fixes a number of bugs in airframe config:
Issue #1543 - You can now scroll the config screen if it doesn't fit
Issue #1539  - Non-standard airframe configs are now supported. It will detect this and show you:
screen shot 2015-05-09 at 10 08 52 am
It also allows you to reset you airframe config if you want to switch from non-standard to standard using the Reset button.

The right hand dialog overlay is implemented using a new standard control I implemented: ViewWithDialog. This control allows you to specify the view you want to show and then pop up those right hand dialog overlay thingies as needed. Both the view an dialog are FactPanel's. It's not quite ready for general use. I'm going to convert more stuff over to it like Sensor Cal to get rotations before/after. Also the Parameter Editor which already uses this style of dialog for editing, but has the the bug Issue #1544, which ViewWithDialog fixes.

Once I convert those two over I should has used ViewWithDialog enough to know it's has enough support for generic usage. Then it can become a standard part of our ui toolbox.

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -32,6 +32,8 @@
         <file alias="QGroundControl/Controls/VehicleSummaryRow.qml">src/QmlControls/VehicleSummaryRow.qml</file>
         <file alias="QGroundControl/Controls/arrow-down.png">src/QmlControls/arrow-down.png</file>
         <file alias="QGroundControl/Controls/ViewWidget.qml">src/ViewWidgets/ViewWidget.qml</file>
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
FYI: I'll give more details on how ViewWithDialog works once it's ready.

LorenzMeier:
@DonLakeFlyer Good to merge?

DonLakeFlyer:
Yup

Issue #1548, opened 2015-05-08
Crash after clicking on "apply and restart" in airframe config


LorenzMeier:
@DonLakeFlyer Is this the rare SYS_AUTOSTART param missing case?

```
Process:               qgroundcontrol [97135]
Path:                  /Users/USER/*/qgroundcontrol.app/Contents/MacOS/qgroundcontrol
Identifier:            org.qgroundcontrol.qgroundcontrol
Version:               2.0.0 (227)
Code Type:             X86-64 (Native)
Parent Process:        Qt Creator [94993]
Responsible:           qgroundcontrol [97135]
User ID:               502

Date/Time:             2015-05-08 10:01:16.668 +0200
OS Version:            Mac OS X 10.10.2 (14C2055)
Report Version:        11
Anonymous UUID:        FFA63A60-490C-41EB-AFE8-E82F56014EB6


Time Awake Since Boot: 11000 seconds

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Application Specific Information:
abort() called
Pure virtual function called!

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff91637286 __pthread_kill + 10
1   libsystem_c.dylib             0x00007fff9a516b53 abort + 129
2   libc++abi.dylib               0x00007fff91d7aa21 abort_message + 257
3   libc++abi.dylib               0x00007fff91da142f __cxa_pure_virtual + 18
4   org.qgroundcontrol.qgroundcontrol	0x00000001000cb470 LinkConfiguration::duplicateSettings(LinkConfiguration*) + 16
5   org.qgroundcontrol.qgroundcontrol	0x000000010010b504 QGCApplication::reconnectAfterWait(int) + 132
6   org.qgroundcontrol.qgroundcontrol	0x0000000100289b6f AirframeComponentController::changeAutostart() + 703
7   org.qgroundcontrol.qgroundcontrol	0x000000010030382c AirframeComponentController::qt_metacall(QMetaObject::Call, int, void**) + 108
8   org.qt-project.QtQml          	0x0000000101d8aa4a CallMethod(QObject*, int, int, int, int*, QV8Engine*, QV4::CallData*) + 330
[...]

PR #1553, opened 2015-05-10
Don't delete link configuration if there is still an active link.

dogmaphobic:
The LinkManager keeps checking to see when new (serial) interfaces show up. When that happens, it checks if it's something that can be used (either a PX4 or a 3DR Radio Modem) and automatically creates a link configuration for it. At the same time, if it sees that a previously automatic link is no longer present, it removes it from the list.

This fix prevents removing a link configuration if an active link is still using it. Once the active link goes down, the configuration will then be properly disposed of.

This fixes the crash reported in #1548 as the code was trying to copy the active link's link configuration and, depending on timing, it could have already been deleted.

Note that this fixes the crash but just moves the problem some place else. Once you reboot the PX4, the link is physically disconnected (while the PX4 is rebooting the physical link disappears). If the serial link catches it early on, it emits a *link disconnected* signal and ```UAS::_linkDisconnected(LinkInterface* link)``` removes it causing an assert crash in:

https://github.com/mavlink/qgroundcontrol/blob/master/src/QGCApplication.cc#L663

If the timing is right and QGC actually manages to reconnect, I'm getting a crash in:

https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/PX4/AirframeComponent.cc#L145

At that point, restarting QGC, it can no longer connect to the PX4. When you hit *Connect*, it looks like it's connected but no data is coming from the serial port. The only way to fix this is to power cycle the PX4.

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -506,9 +506,12 @@ void LinkManager::_updateConfigurationList(void)
         // We only care about dynamic links
         if(pLink->isDynamic()) {
             if(pLink->type() == LinkConfiguration::TypeSerial) {
-                SerialConfiguration* pSerial = dynamic_cast<SerialConfiguration*>(pLink);
-                if(!currentPorts.contains(pSerial->portName())) {
-                    _confToDelete.append(pSerial);
+                // Don't mess with connected link. Let it deal with the disapearing device.
+                if(pLink->getLink() == NULL) {
+                    SerialConfiguration* pSerial = dynamic_cast<SerialConfiguration*>(pLink);
+                    if(!currentPorts.contains(pSerial->portName())) {
+                        _confToDelete.append(pSerial);
+                    }
                 }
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
The AirframeComponent.cc crash is the mystery missing SYS_AUTOSTART crash which I can't quite track down what is causing it. Seems like some firmware strangeness, but not sure yet.

"while the PX4 is rebooting the physical link disappears" - This will cause the link to possibly re-assign names, like you said right? Didn't realize that could happen. So looks like auto-reconnect isn't really possible as you stated. I'll work on it.

LorenzMeier:
I have forced the SYS_AUTOSTART parameter to be referenced now in commit PX4/Firmware@a124bc0. That should put that one to rest. The rest of the link management crash looks like it'll be fixed once your changes are done, will re-test then.

Issue #1555, opened 2015-05-11
Param saved 5x


tumbili:

tumbili:
On initial connection via telemetry (using latest master on a quadrocopter) the "param saved" message appears about 5 times.

DonLakeFlyer:
The initial parameter download over telemetry isn't working quite correctly. It ends up re-requesting missed parameters too many times. In the end everything will be fine, since it will all settle down. But it can slow down connection quite a bit.

DonLakeFlyer:
You can still get some extra "settings autosaved" messages. But they should be more infrequent. The way QGC param loader is implemented it doesn't distinguish between parameters missing from initial load complete and waiting for a write param to come back. Because of that during initial param load it can end up generating extra param writes to storage. This greatly simplifies the QGC code. Don't want to change that since we had a number of problems with previous code trying to get to fancy with handling various states.

PR #1564, opened 2015-05-12
Param read timeout too short

DonLakeFlyer:
Was causing unnecessary read re-requests to get sent out. Fix for Issue #1555


LorenzMeier:
Is the TC fail a fail of TC or a fail of the change?

LorenzMeier:
Merging since I figured this must be a TC fail.

PR #1565, opened 2015-05-12
Firmware Upgrade usability rework

DonLakeFlyer:


DonLakeFlyer:
Hit enter too soon!

If you upgrade with QGC connections you get this:
screen shot 2015-05-12 at 10 51 43 am

If you have no QGC connection but are still connected via USB you get this:
screen shot 2015-05-12 at 10 51 55 am

If you make it this far and you are flashing Beta or Developer builds you get something like this:
screen shot 2015-05-12 at 10 57 15 am

Then finally you are asked to plug your board back in:
screen shot 2015-05-12 at 10 57 37 am

LorenzMeier:
So much better! Good to go once TC is happy.

PR #1570, opened 2015-05-13
Increase timeouts for unit test runs in Team City

DonLakeFlyer:


DonLakeFlyer:
Unit Test are suddenly running twice as slow as they use to run. Last Mac OS update seems to have changed something with respect to file permissions and seems to have hosed something else as well with UT speed.

PR #1572, opened 2015-05-14
Sensor cal usability

DonLakeFlyer:
There are quite a few changes in here which have been discussed in various forums to improve usability. Instead of including a massive amount of screens shots I'll let folks just run it and send comments.

While testing this I found a number of things which used to work but which are now broken with the latest firmware:
- I calibrated a board with an external mag but CAL_MAG1_ROT always came back as -1 not allowing me to set external mag rotation.
- Cancel is broken during mag cal. If you cancel during the rotate detection step, it will emit the calibration cancelled message and continue on and still do the calibration step. 
- The mag cal step which tries to detect rotation before calibrating a side seems flaky. I could get it to detect rotation maybe 50% of the time. For example if my board is flat on the table. And I'm mag calibrating the right side up position, I rotate the board in the plane on the table and it would not detect rotation. I would have to tip the board up (which isn't what you are supposed to do) to get it to detect rotation and move on to calibrate that orientation. 
- Gyro cal on the firmware side no longer uses the detect orientation routine which confuses the QGC state machine. QGC expects orientation detection for all gyro/mag/accel calibrations since it used the same UI to show visuals for all of these. Without this QGC never shows you the Hold Still message.

### DIFF
--- a/QGCApplication.pro
+++ b/QGCApplication.pro
@@ -548,7 +548,8 @@ SOURCES += \
 #
 
 INCLUDEPATH += \
-    src/VehicleSetup
+    src/VehicleSetup \
+    src/AutoPilotPlugins/PX4 \
 
 FORMS += \
     src/VehicleSetup/SetupView.ui \
--- a/src/AutoPilotPlugins/PX4/AirframeComponent.qml
+++ b/src/AutoPilotPlugins/PX4/AirframeComponent.qml
@@ -39,12 +39,8 @@ QGCView {
     Component {
         id: view
 
-        FactPanel {
+        QGCViewPanel {
             id:             panel
-            anchors.fill:   parent
-
-            signal showDialog(Component component, string title, int charWidth, int buttons)
-            signal hideDialog
 
             Connections {
                 target: rootQGCView
--- a/src/AutoPilotPlugins/PX4/AirframeComponentController.cc
+++ b/src/AutoPilotPlugins/PX4/AirframeComponentController.cc
@@ -38,14 +38,10 @@
 bool AirframeComponentController::_typesRegistered = false;
 
 AirframeComponentController::AirframeComponentController(void) :
-    _uas(NULL),
     _currentVehicleIndex(0),
     _autostartId(0),
     _showCustomConfigPanel(false)
 {
-    _uas = UASManager::instance()->getActiveUAS();
-    Q_ASSERT(_uas);
-
     if (!_typesRegistered) {
         _typesRegistered = true;
         qmlRegisterUncreatableType<AirframeType>("QGroundControl.Controllers", 1, 0, "AiframeType", "Can only reference AirframeType");
--- a/src/AutoPilotPlugins/PX4/AirframeComponentController.h
+++ b/src/AutoPilotPlugins/PX4/AirframeComponentController.h
@@ -66,7 +66,6 @@ signals:
 private:
     static bool _typesRegistered;
     
-    UASInterface*       _uas;
     QVariantList        _airframeTypes;
     QString             _currentAirframeType;
     QString             _currentVehicleName;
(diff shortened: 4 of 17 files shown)


DonLakeFlyer:
Oops, sorry about the crappy commit history.

DonLakeFlyer:
FYI: I started from a "param reset" to make sure everything was cleared before I did my testing.

DonLakeFlyer:
@LorenzMeier Can you take a look at the firmware side to things?

LorenzMeier:
> I calibrated a board with an external mag but CAL_MAG1_ROT always came back as -1 not allowing me to set external mag rotation.

Once you connect the external mag it will become MAG0. This is intentional as it ensures that someone with a flimsy cable to the external mag runs into a pre-flight fail if the configuration changes.

> Cancel is broken during mag cal. If you cancel during the rotate detection step, it will emit the calibration cancelled message and continue on and still do the calibration step.

It didn't return immediately. Fixed now, thanks!

> The mag cal step which tries to detect rotation before calibrating a side seems flaky. I could get it to detect rotation maybe 50% of the time. For example if my board is flat on the table. And I'm mag calibrating the right side up position, I rotate the board in the plane on the table and it would not detect rotation. I would have to tip the board up (which isn't what you are supposed to do) to get it to detect rotation and move on to calibrate that orientation.

It works as it should. The reason you felt its flaky is that it wants to see 60 degrees before it believes you that you are rotating, and you probably didn't go through full 60. I cut it to 30 as a compromise between being sure and accuracy.

> Gyro cal on the firmware side no longer uses the detect orientation routine which confuses the QGC state machine. QGC expects orientation detection for all gyro/mag/accel calibrations since it used the same UI to show visuals for all of these. Without this QGC never shows you the Hold Still message.

this probably needs the two of us to resolve, but doesn't seem to be a major deal given that it instructs the user to hold still.

DonLakeFlyer:
> Once you connect the external mag it will become MAG0. This is intentional as it ensures that someone with a flimsy cable to the external mag runs into a pre-flight fail if the configuration changes.

I'm getting -1 on both CAL_MAG_ROT params.

> it wants to see 60 degrees

Yup. When I'm testing I don't really rotate all the way around.

LorenzMeier:
Until you re-calibrate you will get -1. It will only change after you calibrated.

DonLakeFlyer:
> It will only change after you calibrated.

It's not changing after I calibrate. I'll see if I have time to try again today. I'm running a board with a 3dr extrnal mag/compass not in a vehicle.

LorenzMeier:
Ok, what do you get on the shell for this?
```
nsh> hmc5883 status
hmc5883: running on bus: 2 (/dev/hmc5883_ext)

hmc5883_read: 101554 events, 0 overruns, 105166048us elapsed, 1035us avg, min 871us max 33095us 303.362us rms
hmc5883_comms_errors: 0 events
hmc5883_buffer_overflows: 101512 events
poll interval:  20 ticks
output  (0.10 0.33 0.47)
offsets (0.09 0.04 0.09)
scaling (1.13 1.01 1.05) 1/range_scale 1090.000 range_ga 1.30
temperature 36.75
report queue    2/96 (0/2 @ 10003f20)
```

LorenzMeier:
In particular this should show up:
```
hmc5883: running on bus: 2 (/dev/hmc5883_ext)
```

If it doesn't, make sure you have the serial cable for the GPS connected. That's the one powering the sensor, the I2C cable only is for the sensor signal.

DonLakeFlyer:
nsh> .[Khmc5883 status
hmc5883: bus 0 not started

LorenzMeier:
Ok, so your wiring is probably bad and the sensor is not actually running.

DonLakeFlyer:
> If it doesn't, make sure you have the serial cable for the GPS connected.

Duh. My desk is such a tangled pile of wires I never noticed it!

DonLakeFlyer:
Pulled it out when I was switching around airspeed sensors.

PR #1583, opened 2015-05-18
Adding Battery Consumption to the tool bar.

dogmaphobic:
Adding battery consumption to the toolbar.

It doesn't seem to be working though. I haven't had the time to dig too deep but even with full throttle, the rate of increase didn't change from that shown at idle (disarmed and motors off). After a minute or so of full throttle (theoretically a good 2A discharge), it still shows around 100mA.

P.S. Throttle response seems to be completely out of whack. Again, I didn't have time to dig further but at first only three motors (out of six) started and moving the throttle all the way up did nothing. After quite a while, all of the sudden, all motors started at full throttle scaring the crap out of me. All this in *Manual* mode.

P.P.S The error messages are related to the radio as I was messing with it and it kept connecting and disconnecting before I started the tests.

screen shot 2015-05-18 at 3 49 49 pm

### DIFF
--- a/src/QmlControls/MavManager.cc
+++ b/src/QmlControls/MavManager.cc
@@ -57,6 +57,7 @@ MavManager::MavManager(QObject *parent)
     , _refreshTimer(new QTimer(this))
     , _batteryVoltage(0.0)
     , _batteryPercent(0.0)
+    , _batteryConsumed(0.0)
     , _systemArmed(false)
     , _currentHeartbeatTimeout(0)
     , _waypointDistance(0.0)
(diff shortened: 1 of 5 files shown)


LorenzMeier:
I presume that's with current stable or master firmware?

dogmaphobic:
Master firmware (flashed from QGC). I will build it locally later tonight and see how it goes...

dogmaphobic:
Oh, yes, QGC master (*Stable*) too...

LorenzMeier:
I briefly checked the code and it looks correct. I need to set up a test with a real vehicle to verify. But UI-wise it looks good and could be merged.

LorenzMeier:
Hah! I can repro with the Hex config - digging..

LorenzMeier:
My screwup during a rework of the changed storage this morning. Fixed in: https://github.com/PX4/Firmware/commit/c06ba047e2ded236b8ce5e17694dfca8b8d6f067

LorenzMeier:
I'm merging this since I think its a setup issue or Firmware issue which is preventing the correct display.

PR #1585, opened 2015-05-18
Fix list EOF handling

DonLakeFlyer:
@LorenzMeier Can you give this a try? Should prevent it from looping.


LorenzMeier:
This fixed it, thanks!

PR #1587, opened 2015-05-19
Horizon level cal

LorenzMeier:
@DonLakeFlyer This implements the horizon levelling and there is a matching Firmware PR coming.


LorenzMeier:
Tested, works with current Firmware master.

DonLakeFlyer:
Could this just be the first step of a standard accel cal? When you do a right-side up accel cal you are 99% of the way to a level cal. Why not combine the two to simplify?

LorenzMeier:
Accel cal is the input to level. It does scale and offsets, level does rotation

LorenzMeier:
This means the level step will often have the autopilot at an angle, which is the one we're compensating in the final step

DonLakeFlyer:
Add it at the end then? Just trying to remove confusion over additional steps/buttons.

DonLakeFlyer:
Also from looking at this code it appears that leveling is non-optional since it is using a red/green indicator and testing a param for being modified. Is that correct? If so, I'll need to add that param to the sensor component as a trigger for the component setup being complete. Otherwise sensor component will show overall green when level has not been completed.

Issue #1582, opened 2015-05-18
Calibration view stays greyed out after compass calibration


AndreasAntener:
Minor issue: after compass calibration the view stays greyed out, you need to switch into another view and back to continue with other calibrations

DonLakeFlyer:
Should be fixed now

PR #1589, opened 2015-05-19
Animation fix

DonLakeFlyer:
Showing a new dialog while the animation from a previous hide was still playing would lead to the panel still being overlaid.

Also fixed some typos, and changed some naming.


DonLakeFlyer:
I believe this fixes Issue #1582 as well.

LorenzMeier:
Does it also fix the aborted calibration making the other calibration buttons inaccessible?

DonLakeFlyer:
I didn't see that bug. Is Cancel not working? I saw #1582 which I tested and it seems to fix. But that is a positive calibration.

PR #1593, opened 2015-05-21
Text cleanup

LorenzMeier:
@DonLakeFlyer This addresses the low hanging fruits in terms of usability.

### DIFF
--- a/src/AutoPilotPlugins/PX4/SensorsComponent.qml
+++ b/src/AutoPilotPlugins/PX4/SensorsComponent.qml
@@ -425,11 +425,19 @@ QGCView {
                             id:             orientationCalAreaHelpText
                             width:          parent.width
                             wrapMode:       Text.WordWrap
-                            font.pointSize: ScreenTools.fontPointFactor * (17);
+                            font.pointSize: ScreenTools.fontPointFactor * (22);
+                            anchors.top: orientationCalArea.top
+                            anchors.left: orientationCalArea.left
+                            anchors.topMargin: 15
+                            anchors.leftMargin: 15
+                            anchors.rightMargin: 15
+                            anchors.bottomMargin: 15
                         }
 
                         Flow {
-                            y:          orientationCalAreaHelpText.height
+                            anchors.top: orientationCalAreaHelpText.bottom
+                            anchors.left: orientationCalAreaHelpText.left
+                            anchors.topMargin: 15
                             width:      parent.width
                             height:     parent.height - orientationCalAreaHelpText.implicitHeight
                             spacing:    5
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
I'm doing a run through on this:
- The Level button doesn't disable when all the other buttons do. Needs to be added to _startVisualCalibration and _stopCalibration
- The set rotation dialog which comes up after mag cal is out of whack.
screen shot 2015-05-21 at 10 26 50 am

DonLakeFlyer:
The second interior Column element you added isn't working correctly. I'm not exactly sure why. I'd just use the outer Column even if the spacing isn't as nice.

DonLakeFlyer:
Last thing: I can't get "Level" to work. It keeps telling me excess roll or excess pitch and fails. I tried re-cal of accel and level again. Same thing.

LorenzMeier:
Then you don't have the autopilot level?

LorenzMeier:
Its not designed to compensate for large rotations. Its fine-tuning.

DonLakeFlyer:
> Then you don't have the autopilot level?

As level as I can get it. It maybe be off by tiny bit.

Just tried it again with a new board I pulled out of a box. But I could level because it requires gyro cal to be complete first? enabled: checkAcc.value != 0 && checkGyro.value != 0. Why do I need a gyro connected to do level cal?

LorenzMeier:
Because levelling the attitude estimate requires the complete attitude estimator inputs to be calibrated.

DonLakeFlyer:
Tried a different board which I could lay flat on my desk and it worked. Tried the other board which is mounted in 3dr bixler mounting thing. Can't get it to work. I'm holding it as level as I can. Noticed more error text:
not enough measurements taken
excess roll angle

LorenzMeier:
ah, now it makes a lot more sense. I can fix that.

LorenzMeier:
I fixed the mag layout I think. Trying to fix the level button led to crashes - please have a look at: The last commit is the offending one.
https://github.com/mavlink/qgroundcontrol/tree/text_cleanup_crash

DonLakeFlyer:
I don't seem to be seeing that commit.

LorenzMeier:
Direct link:
https://github.com/mavlink/qgroundcontrol/commit/c6bce7510b9a1f85aca6ad631fdfcdf061b43313

DonLakeFlyer:
Just making sure you see this:
Need to push level button into controller here:

            onCompleted: {
                controller.factPanel = viewPanel
                controller.statusLog = statusTextArea
                controller.progressBar = progressBar
                controller.compassButton = compassButton
                controller.gyroButton = gyroButton
                controller.accelButton = accelButton
                controller.airspeedButton = airspeedButton
                controller.cancelButton = cancelButton
                controller.orientationCalAreaHelpText = orientationCalAreaHelpText
            }

LorenzMeier:
Fixed, should behave now.

LorenzMeier:
To be more precise: I tested it and it works. It should work for you now, too.

LorenzMeier:
@DonLakeFlyer Calibration should work with HEAD of release_v1.0.0 also in your Bixler now. And it should provide better error messages and abort early on the first fatal error.

PR #1594, opened 2015-05-22
PX4 flight modes: Cleanup acro, introduce stabilize

LorenzMeier:
This just adds labels / string conversions


LorenzMeier:
@DonLakeFlyer This is good to go.

PR #1595, opened 2015-05-22
ESC Calibration

DonLakeFlyer:
Added to the Power Component. Requires firmware pull: https://github.com/PX4/Firmware/pull/2216/files#diff-136fc0576591fc6267a38badf810aec5R142.

Also added new medium and large default font sizes to ScreenTools. So for example the top Setup screen labels are large font and the gray shaded section headers in the setup screens are medium.


LorenzMeier:
Works according to Roman.

LorenzMeier:
So in other words: Good to merge. Firmware part is also going in now. Thanks for making this happen in time for the release!

DonLakeFlyer:
Waiting for TC then it will go in. Let me know if there is anything else for Stable, otherwise I'm going to cut a new Stable tonight.

LorenzMeier:
Maybe in the analyse view hide all widgets except for the plot and file manager?

PR #1601, opened 2015-05-24
UDP fix sim links

LorenzMeier:
@dogmaphobic This is your fix from #1600 applied to the simulation links.

@tumbili This will likely improve your experience when running HIL.

### DIFF
--- a/src/comm/QGCFlightGearLink.cc
+++ b/src/comm/QGCFlightGearLink.cc
@@ -90,7 +90,7 @@ void QGCFlightGearLink::run()
     _udpCommSocket = new QUdpSocket(this);
     Q_CHECK_PTR(_udpCommSocket);
     _udpCommSocket->moveToThread(this);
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
@dogmaphobic Did you take a look at this?

dogmaphobic:
> @dogmaphobic Did you take a look at this?

Yes, those fixes all set the proper socket option. The original bug however, also lacked error handling, which is not part of this batch. There were two problems. The first was related to the socket option. Subsequent connections would fail to bind because the port was *busy* (it takes a while for the kernel to release the port). The second problem was that there was no error handling. The link would not connect, yet, it would issue a *link disconnected* signal on the way out. MavLinkProtocol would get this signal, look for a connected link, find none and then assert as a result.

DonLakeFlyer:
I'll merge this in then. Can you the last set of changes to  you list of stuff?

Issue #1598, opened 2015-05-23
Add RC_MAP_ACRO to mode switch UI


LorenzMeier:
@DonLakeFlyer This one I completely forgot: We reconfigured the RC_MAP_ACRO switch into becoming part of the normal mode selection. If configured it will be evaluated once in manual mode.

Right now its a 2-position switch since we have no acro mode in fixed wing yet, but that would be trivial to add.

@tumbili FYI

Fixed wing:
 - off: manual
 - center: acro (NOT YET IMPLEMENTED IN FIRMWARE, but could be easily changed in sensors.cpp)
 - on: stabilized

Multicopter:
  - off: manual (= stabilized)
  - center: acro
  - on: acro

DonLakeFlyer:
So it needs to be on it's own separate channel?

DonLakeFlyer:
Confused as to how this switch can work with one RC_ACRO_TH defaulting to 0.5? That will only work with a two position switch. But it seems like you are describing this as a 3 position switch: off/center/on?

DonLakeFlyer:
Is there an update this this: https://pixhawk.org/users/system_modes?s[]=flight&s[]=mode, which would explain it?

DonLakeFlyer:
@LorenzMeier Any more info on this?

LorenzMeier:
Fixed with recent master, thanks!

PR #1602, opened 2015-05-25
Validate against Acro switch setting

DonLakeFlyer:
This is the first phase of Acro switch support. It just prevents QGC from doing the wrong thing if the acro switch is configured. Support for setting macro switch in Flight Mode setup is not yet supported.


DonLakeFlyer:
Phase 1 for Issue #1598

PR #1606, opened 2015-05-26
FactPanel V2 (much better missing parameter handling)

DonLakeFlyer:
This is a much better implementation of FactPanel and friends in so far as handling missing parameters. It requires less setup and is more automatic as far as detecting and reporting missing parameters. With this new implementation and pull all of the Setup Qml and C++ code is converted over to use it and is now hardened to missing parameters.

I'll update in a separate comment the new way of doing things.


DonLakeFlyer:
You must now use FactPanel for creating any Qml ui which references Facts. Also a FactPanelController is required in order to access Facts. You can use the provided FactPanelController if all you need to do is access Facts. You can also subclass FactPanelController in your C++ code to create your own controller if needed. In your controller C++ code you also use FactPanelController::getParameterFact to reference facts. If you do both these things, your Qml and C++ code will be hardened to missing parameters and will automatically show the right popups and missing parameter ui overlays.

```javascript
FactPanel {
    id: panel

    // controller is required
    FactPanelController { id: controller; factPanel: panel } 

    // -1 signals default component id
    property Fact myFact: controller.getParameterFact(-1, "RC_MAP_THROTTLE")

    FactLabel {
        fact: controller.getParameterFact(-1 /* signals default component id */, "RC_MAP_THROTTLE")
    }
}
```

Also important to note that the previous Fact { name: "RC_MAP_THROTTLE" } will no longer work. Fact.name and Fact.componentId are ready-only properties now since you must call FactPanelController::getParameterFact to get a fact. You can create an empty fact like this Fact { } which can be handy in some special cases.

Also, the inner working of ParameterLoader::getParameterFact will no longer assert if you try to reference a missing fact from some random piece of C++ code. It will instead return a pointer to a default fact which is initialized to 0. It will also automatically pop the missing parameter dialog to the user.

The combination of all of this should now fully harden QGC against missing facts and let it keep running at least long enough for you to disconnect, shut down and figure out why your firmware is out of whack,

DonLakeFlyer:
Not ready for merge yet. Need to fix destruction sequencing problem between autopilot plugin and Qml references to Facts.

PR #1607, opened 2015-05-27
basic ubuntu ppa instructions

dagar:
I updated the Ubuntu ppa to Stable 2.4 by doing this. It could be turned into a proper script if we get the version out somehow and properly handle the paths.

I still can't register for the wiki (the captcha doesn't work?). Can you readd the instructions?


DonLakeFlyer:
Wiki updated.

PR #1610, opened 2015-05-27
completed signaled too early on QGCView

DonLakeFlyer:


DonLakeFlyer:
When you used QGCView inside QGCQmlWidgetHolder the Component.onCompleted signals comes through before the width and height are set. This causes the dialog panel to work incorrectly if they are displayed from that signal. Need to wait for width/height to be set before signaling.

PR #1611, opened 2015-05-28
Convert RadioCal to Qml

DonLakeFlyer:
This is the last Setup panel to go to Qml. Finishing this up will allow me to convert the entire Setup view to Qml only, instead of the hybrid widget+Qml that it is now. This will then allow me to fix some other setup usability problems.

At this stage RadioCal unit test is turned off. I'm pushing up now to see if I can get some help testing. RadioCal is fairly involved. So extra eyes are appreciated.

RadioCal works just like it did before as far as the sequence of steps.


DonLakeFlyer:
I'll continue on with getting the unit test working before I merge this. Changing from QWidget unit testing to Qml unit test is somewhat involved.

DonLakeFlyer:
Turned out to be easier to covert RadioCal unit test than I thought.

src/AutoPilotPlugins/PX4/RadioComponent.cc
```diff
@@ -140,7 +140,11 @@ QStringList RadioComponent::paramFilterList(void) const
 
 QWidget* RadioComponent::setupWidget(void) const
 {
-    return new PX4RCCalibration;
+    QGCQmlWidgetHolder* holder = new QGCQmlWidgetHolder();
```
vooon: @DonLakeFlyer may i ask you, why you don't use C++11 `auto`? I think that is very handy, and i widely use it in mavros.
Only one thing, seems that compiling is slower, over one minute for such a small project.

DonLakeFlyer: I think in general auto hides too much of the details of the code. So I stay away from it.

DonLakeFlyer: > Only one thing, seems that compiling is slower, over one minute for such a small project.

I haven't noticed much compile time increase. TeamCity is reporting the same builds time for the last couple weeks.

PR #1621, opened 2015-05-30
First stab at iOS builds

dogmaphobic:
It builds clean, deploys and launches. It doesn't (yet) work properly however as QQuickWidget is not supported on iOS. It is supposed to get fixed in Qt 5.5. In the mean time, I'm making this commit now because it's been a pain merging changes back into it every time.

iOS has no Serial link support. Only UDP (and TCP) links are available. I've added Zeroconf (Bonjour/Avahi) to QGC (different PR), which allows an UDP client to automatically find QGC on a local network. The way I run this is having a proxy running on a Raspberry PI. The RPI connects to the PX4 using a serial connection and *finds* a QGC on the local network. It then routes traffic from the serial port (PX4) to the UDP connection (QGC).

While at it, I removed all the simulation code from the *mobile* builds.


DonLakeFlyer:
I trashed this with some the earlier pulls. Looks like it's going to need a rebase first.

dogmaphobic:
And I thought I was done rebasing this thing for the umpteenth time :)

dogmaphobic:
This got too messed up. I will close it and try again.

PR #1623, opened 2015-05-31
First stab at iOS builds

dogmaphobic:
It builds clean, deploys and launches. It doesn't (yet) work properly however as QQuickWidget is not supported on iOS. It is supposed to get fixed in Qt 5.5. In the mean time, I'm making this commit now because it's been a pain merging changes back into it every time.

iOS has no Serial link support. Only UDP (and TCP) links are available. I've added Zeroconf (Bonjour/Avahi) to QGC (different PR), which allows an UDP client to automatically find QGC on a local network. The way I run this is having a proxy running on a Raspberry PI. The RPI connects to the PX4 using a serial connection and *finds* a QGC on the local network. It then routes traffic from the serial port (PX4) to the UDP connection (QGC) and vice versa.

While at it, I removed all the simulation code from the *mobile* builds.


dogmaphobic:
Technically, it's the *second* stab at it but who is counting... :P

dogmaphobic:
Qt 5.5 did indeed fix the issues with QQuickWidget. They've also fixed the mapping bug I reported, changing the private mapping API in the process. I've already fixed those changes as well in preparation for some eventual move to Qt 5.5 once it's released.

screen shot 2015-05-31 at 2 01 57 am

dogmaphobic:
TC failure seems to be unrelated to my commit. Rather, it looks like it barfed while fetching the tree from GitHub (5 out of 6 succeeded):
```
[23:31:45] :	 [Step 5/20] C:\BuildAgent\work\428058b0c4032c78>git  submodule update 
[23:31:46]W:	 [Step 5/20] Cloning into 'libs/mavlink/include/mavlink/v1.0'...
[23:32:07]W:	 [Step 5/20] fatal: unable to access 'https://github.com/mavlink/c_library.git/': Failed connect to github.com:443; No error
[23:32:07]W:	 [Step 5/20] Clone of 'https://github.com/mavlink/c_library.git' into submodule path 'libs/mavlink/include/mavlink/v1.0' failed
[23:32:07]W:	 [Step 5/20] Process exited with code 1
```

LorenzMeier:
This looks awesome! So much looking forward to it!

PR #1629, opened 2015-06-01
Add AeroCore support to firmware upgrade utility and link manager.

smithandrewc:
This will allow users to easily update their AeroCore boards rather than building the code themselves.

The link manager update just avoids some confusion.  When an AeroCore board is plugged in, it will show "AeroCore on /dev/..." rather than "Pixhawk on /dev/...".

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -481,7 +481,10 @@ void LinkManager::_updateConfigurationList(void)
                 }
             } else {
                 // Lets create a new Serial configuration automatically
-                pSerial = new SerialConfiguration(QString("Pixhawk on %1").arg(portInfo.portName().trimmed()));
+		if (portInfo.description() == "AeroCore")
+	                pSerial = new SerialConfiguration(QString("AeroCore on %1").arg(portInfo.portName().trimmed()));
+		else
+	                pSerial = new SerialConfiguration(QString("Pixhawk on %1").arg(portInfo.portName().trimmed()));
                 pSerial->setDynamic(true);
                 pSerial->setPreferred(true);
                 pSerial->setBaud(115200);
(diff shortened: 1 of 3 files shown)


LorenzMeier:
@DonLakeFlyer Looks good to me.

src/comm/LinkManager.cc
```diff
@@ -481,7 +481,10 @@ void LinkManager::_updateConfigurationList(void)
                 }
             } else {
                 // Lets create a new Serial configuration automatically
-                pSerial = new SerialConfiguration(QString("Pixhawk on %1").arg(portInfo.portName().trimmed()));
+		if (portInfo.description() == "AeroCore")
```
DonLakeFlyer: @smithandrewc @dogmaphobic Can we get other boards that could come across with vendor id = 9900 but not be a Pixhawk? Wondering if there should be a description string check for Pixhawk (or other things) here on the other side of the if?

DonLakeFlyer:
Merging since this doesn't need to wait for the answer to my question.

src/comm/LinkManager.cc
smithandrewc: Why not use the portInfo.description() for the name of the device?  The Pixhawk has "PX4 v2 ..." or something like that but could just be parsed out.

dogmaphobic: > Can we get other boards that could come across with vendor id = 9900 but not be a Pixhawk? Wondering if there should be a description string check for Pixhawk (or other things) here on the other side of the if?

Good question. The vendor id 9900 (0x26AC) doesn't seem to be a real, registered id. At least it's not in the official database. Provided all PX4 controllers out there return the same *description*, we could certainly use the same logic used here to identify the AeroCore.

PR #1633, opened 2015-06-02
Adding check for a PX4.

dogmaphobic:
Following #1629, adding code to make sure the newly found *Vendor ID 9900* is really a PX4.

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -481,10 +481,13 @@ void LinkManager::_updateConfigurationList(void)
                 }
             } else {
                 // Lets create a new Serial configuration automatically
-		if (portInfo.description() == "AeroCore")
-	                pSerial = new SerialConfiguration(QString("AeroCore on %1").arg(portInfo.portName().trimmed()));
-		else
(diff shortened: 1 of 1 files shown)


nopeppermint:
this only checks if there is the px4 bootloader?
So there could be a px4-stm32f4discoveryboard, px4fmu-v1, px4fmu-v2 or a pixhawk behind, as they all use the px4 bootloader, right?

LorenzMeier:
In theory, yes

dogmaphobic:
> this only checks if there is the px4 bootloader?
> So there could be a px4-stm32f4discoveryboard, px4fmu-v1, px4fmu-v2 or a pixhawk behind, as they all use the px4 bootloader, right?

Well, it went from accepting anything that announced itself as manufacturer id 9900 to a bit more specific. Any suggestion as to what *not* to accept as Pixhawk?

DonLakeFlyer:
This has nothing to do with the bootloader. This is just for AutoDetect of a Pixhawk board to automatically be added to the connections list. If the string PX4 isn't in the port description it won't get automatically added as a "Pixhawk". I believe the code is correct. Everyone ok with that?

LorenzMeier:
Yes, I think that's sufficient level of detail.

PR #1637, opened 2015-06-04
Added PX4Flow identification

DonLakeFlyer:
Also added logging category


DonLakeFlyer:
Nope. Just shows bad value in summary for airspeed setup.

PR #1647, opened 2015-06-06
Hil fixes

LorenzMeier:
This fixes all current crashes on HIL setup. The HIL perspective as a separate developer-facing view currently makes sense, but it doesn't look like it would be very hard to collapse it into the flight / mission perspectives in the future.

### DIFF
--- a/src/uas/UASManagerInterface.h
+++ b/src/uas/UASManagerInterface.h
@@ -98,6 +98,8 @@ signals:
     void UASCreated(UASInterface* UAS);
     /** A system was deleted */
     void UASDeleted(UASInterface* UAS);
+    /** A system was deleted */
+    void UASDeleted(int systemId);
     /** @brief The UAS currently under main operator control changed */
     void activeUASSet(UASInterface* UAS);
     /** @brief The UAS currently under main operator control changed */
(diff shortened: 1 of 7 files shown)


src/uas/UASManagerInterface.h
```diff
@@ -98,6 +98,8 @@ public slots:
     void UASCreated(UASInterface* UAS);
     /** A system was deleted */
     void UASDeleted(UASInterface* UAS);
+    /** A system was deleted */
+    void UASDeleted(int systemId);
```
DonLakeFlyer: I've been trying to get rid of all the overridden signals with same name different params. The reason is that it the forces you to use the SIGNAL/SLOT connect mechanism instead of the direct method reference . The SIGNAL/SLOT connect mechanism doesn't provide compile time checking for incorrect usage. There were a pile of problems caused by not having this compile time check with broken signals. Can't the HIL has map be changed to index to pointer instead of id?

LorenzMeier: Since its for an instance of the class rather than for a particular UAS ID we could change it, yes.

DonLakeFlyer:
I'll merge in as is anyway so I can start stable and clean this up after.

PR #1648, opened 2015-06-06
Updated parameter list metadata to new position estimator

TSC21:
This is an update to the parameter list metadata of QGC so to allow parameter changing of an underdevelopment position estimator - [LPE](https://github.com/PX4/Firmware/pull/2096#commitcomment-11528996). As the this PR also adds `vision_speed_estimate` topic, which positively contributes to `position_estimator_inav`, also `INAV_W_Z_VIS_V` was added on the list.


LorenzMeier:
Two notes:

  - Parameters get cached. So flashing your binary once with QGC would have loaded that XML file for you
  - Parameters are always visible, but are mapped to the default group when no group is selected. Do you have suggestions how to rename the default group so people have a higher chance of finding it?

TSC21:
> * Parameters get cached. So flashing your binary once with QGC would have loaded that XML file for you

Thanks @LorenzMeier I wasn't aware of it. Here's the parameters being shown on QGC:
screenshot from 2015-06-06 14 27 32

image

> * Parameters are always visible, but are mapped to the default group when no group is selected. Do you have suggestions how to rename the default group so people have a higher chance of finding it?

The group is already created as you can see. Or at least, for my understanding, it is already created here: https://github.com/jgoppert/Firmware/blob/cart_pos_estimator/src/modules/local_position_estimator/params.c

TSC21:
Now I was using default `ParameterFactMetaData.xml` before the changes. But, after uploading the bin using QGC, the parameters are loaded as above, but `ParameterFactMetaData.xml` code is not updated. Is not updated. Is that supposed to be done manually?

DonLakeFlyer:
> But, after uploading the bin using QGC, the parameters are loaded as above

If you flash a bin file parameters will not be updated. Parameters are only included in .px4 files. The meta data file itself is placed in a cache location which varies from OS to OS. The version which is in the repository is not updated. To see if the .px4 flash with parameters is working, flash, reboot and check the parameter editor for updates.

TSC21:
> To see if the .px4 flash with parameters is working, flash, reboot and check the parameter editor for updates.

Yes I confirmed it is working thank you!

> The version which is in the repository is not updated.

So does this PR still applies?

DonLakeFlyer:
> So does this PR still applies?

Can't hurt. Needs to be fixed though. There are typos in the xml file. Details link will show you TeamCity build log files.

TSC21:
I'm getting:
`[05:06:35] :	 [Step 14/20] QWARN  : FactSystemTestPX4::parameter_default_component_id_test() Parameter meta data with bad type: "INT"  name: "LPE_NO_VISION"` in line 971. Can't understand why as https://github.com/PX4/Firmware/pull/2096/files#diff-b346c943d4196c6f075309a366b405a3R11.

DonLakeFlyer:
INT is not a valid type. INT32 is.

TSC21:
But it is set as `INT32`:https://github.com/mavlink/qgroundcontrol/pull/1648/files#diff-ada4637acb0b6a1f18afab8cb36cd5e8R1501 and here also: https://github.com/PX4/Firmware/pull/2096/files#diff-b346c943d4196c6f075309a366b405a3R11.

TSC21:
Do I have to setup `MockLink` also?

DonLakeFlyer:
How did update this .xml file by hand or did you use the output from the firmware build system? If output from the build system then there is something wrong with the parameter parser since it is outputting the wrong thing.

LorenzMeier:
@TSC21 You need to look at this one, which is wrong: https://github.com/mavlink/qgroundcontrol/pull/1648/files#diff-ada4637acb0b6a1f18afab8cb36cd5e8R1597

DonLakeFlyer:
If you updated the xml file by hand then there is a typo as Lorenz says above.

TSC21:
So strings must match what is in the `params.c` file right?

TSC21:
OK checked! Missed that one!

DonLakeFlyer:
There is no need to update MockLink param set. Although that doesn't hurt either.

TSC21:
Well we don't lose nothing :)

DonLakeFlyer:
Still broken. MockLink file must be tab separated. TC results will show up in a few minutes. You can keep fighting with this if you like. But you can also just toss it, since it doesn't really fix anything.

TSC21:
Why I'm not getting any TC results?

DonLakeFlyer:
Not sure. But the error is because the rows you added don't look to be tab separated.

TSC21:
I have the columns separated only by tab so it doesn't make any sense. I have doubled check already.

TSC21:
I understood the mechanics now: only a single tab between `paramData`. The strange thing is between `PARAM NAME` and `VALUE` in some cases, which have a tab, following by multiple whitespaces just for cosmetic enhance.

LorenzMeier:
@TSC21 I presume you're aware that the param file is generated on each build and placed into the Build folder? It sounds like you're editing it manually.

TSC21:
@LorenzMeier I wasn't aware that the parameters file had the same structure as `ParameterFactMetaData.xml`. I will upload this new one then.

TSC21:
Probably this is good to merge, since it doesn't directly conflict with QGC param loader.

DonLakeFlyer:
I'm going to merge after Stable just to be safe

TSC21:
Ok sure.

TSC21:
Is this mergeble already @DonLakeFlyer? Just to keep the workflow on this.

DonLakeFlyer:
Master is sticking on Stable until I get back from out town. So all pulls are on hold.

PR #1655, opened 2015-06-08
Usability fixes:

DonLakeFlyer:
- Only show clear trim dialog once
- Remove animation from bars
- Move additional setup down with bar

@LorenzMeier Can you try this one out?

### DIFF
--- a/src/AutoPilotPlugins/PX4/RadioComponent.qml
+++ b/src/AutoPilotPlugins/PX4/RadioComponent.qml
@@ -236,6 +236,8 @@ QGCView {
                     duration:   1500
                 }
 
+                /*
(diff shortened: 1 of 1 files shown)


LorenzMeier:
Will do and report back. Thanks!

LorenzMeier:
@DonLakeFlyer BTW: I liked the bar animation - just not during calibration 8).

DonLakeFlyer:
Yup, my idea as well. Not enough time right now to make it work correctly. So off for now.

DonLakeFlyer:
Merging in so it will show up in tonights dailies. We can fix any issues after the fact.

LorenzMeier:
looking good, thanks!

PR #1657, opened 2015-06-09
Font to pixels

dogmaphobic:
Font point size simply does not work. Not only it differs from platform to platform, it even differs from QWidget to QtQuick within a same platform. We were using *factors* to accommodate the differences but the problem is that Qt does not support floating point for font sizes (point or pixels). As these factors were used, Qt would round down the sizes creating even more confusion.

Font **pixel** sizes however, are very close in all platforms. From what I measured, the greatest difference (not counting Android), was 0.97. Enough to justify replacing all font size references from points to pixels. On all platforms I tested, Android was the only one that required a doubling in size.

Gone are all "font size factors", and all the runtime computation. In its place, there are now constants for all used font sizes (8 to 22). These factors are defined at compile time per platform as needed (as I mentioned above, only Android needs a different size).

I've replaced all font sizes in the style sheet to *tokens*, which are then parsed and replaced appropriately at runtime.

I've also fixed (somewhat) the QuickView gadget as it was grossly messing up in all platforms. It is now at least readable, provided you have a reasonable number of items to display (it's still possible to make it horrible by adding a zillion items and/or columns). This will need a complete re-write, which is forthcoming anyway with the change over to QtQuick.

With that said, most *gadgets* don't fit docked next to the main view in small screens (smaller than 1280 wide). That's a function of the UI layout in each of them. Making the screen larger (provided you have a larger monitor), makes them normal and usable. On small screens, I see no point in running multiple gadgets side by side as the layout breaks apart.

Tested on:
* Mac OS (MacPro 30" 4k monitor and 27" monitor)
* Mac OS (MacBook Pro 13")
* Windows (24" monitor)
* Ubuntu (24" monitor)
* Android (7" Nexus tablet)
* iOS (iPad Retina)

### DIFF
--- a/src/AutoPilotPlugins/PX4/SensorsComponent.qml
+++ b/src/AutoPilotPlugins/PX4/SensorsComponent.qml
@@ -57,8 +57,8 @@ QGCView {
     // Used to pass help text to the preCalibrationDialog dialog
     property string preCalibrationDialogHelp
 
-    readonly property int sideBarH1PointSize: (ScreenTools.defaultFontPointSize * 1.3 + 0.5)
-    readonly property int mainTextH1PointSize: (ScreenTools.defaultFontPointSize * 1.5 + 0.5)
+    readonly property int sideBarH1PointSize: (ScreenTools.defaultFontPizelSize * 1.3 + 0.5)
+    readonly property int mainTextH1PointSize: (ScreenTools.defaultFontPizelSize * 1.5 + 0.5)
 
     readonly property int rotationColumnWidth: 250
     readonly property var rotations: [
@@ -213,7 +213,7 @@ QGCView {
                     id: compass0ComponentLabel
 
                     QGCLabel {
-                        font.pointSize: sideBarH1PointSize
+                        font.pixelSize: sideBarH1PointSize
                         text: "External Compass Orientation"
                     }
 
@@ -280,7 +280,7 @@ QGCView {
 
             QGCLabel {
                 text: "SENSORS CONFIG"
-                font.pointSize: ScreenTools.largeFontPointSize
+                font.pixelSize: ScreenTools.largeFontPixelSize
             }
 
             Item { height: 20; width: 10 } // spacer
@@ -408,7 +408,7 @@ QGCView {
                         id:             orientationCalAreaHelpText
                         width:          parent.width
                         wrapMode:       Text.WordWrap
-                        font.pointSize: ScreenTools.fontPointFactor * (22);
+                        font.pixelSize: ScreenTools.font22;
                         anchors.top: orientationCalArea.top
                         anchors.left: orientationCalArea.left
                         anchors.topMargin: 15
@@ -480,7 +480,7 @@ QGCView {
                         spacing:            5
 
                         QGCLabel {
-                            font.pointSize: sideBarH1PointSize
+                            font.pixelSize: sideBarH1PointSize
                             text: "Autopilot Orientation"
                         }
 
@@ -506,7 +506,7 @@ QGCView {
                             id: compass0ComponentLabel2
 
                             QGCLabel {
-                                font.pointSize: sideBarH1PointSize
+                                font.pixelSize: sideBarH1PointSize
                                 text: "External Compass Orientation"
                             }
                         }
@@ -531,7 +531,7 @@ QGCView {
                             id: compass1ComponentLabel2
 
                             QGCLabel {
-                                font.pointSize: sideBarH1PointSize
+                                font.pixelSize: sideBarH1PointSize
                                 text: "External Compass 1 Orientation"
                             }
                         }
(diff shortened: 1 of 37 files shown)


src/AutoPilotPlugins/PX4/SensorsComponent.qml
```diff
@@ -57,8 +57,8 @@ QGCView {
     // Used to pass help text to the preCalibrationDialog dialog
     property string preCalibrationDialogHelp
 
-    readonly property int sideBarH1PointSize: (ScreenTools.defaultFontPointSize * 1.3 + 0.5)
-    readonly property int mainTextH1PointSize: (ScreenTools.defaultFontPointSize * 1.5 + 0.5)
+    readonly property int sideBarH1PointSize: (ScreenTools.defaultFontPizelSize * 1.3 + 0.5)
+    readonly property int mainTextH1PointSize: (ScreenTools.defaultFontPizelSize * 1.5 + 0.5)
```
dogmaphobic: I didn't know how to deal with these. I will let @DonLakeFlyer decide if this is OK or not.

dogmaphobic:
TC failed... I will take a look in the morning.

dogmaphobic:
Hmm, it passes here:
```
<...>
QDEBUG : RadioConfigTest::_fullCalibration_test() MockLink: Unhandled mavlink message, id: 76
QDEBUG : RadioConfigTest::_fullCalibration_test() MockLink: Unhandled mavlink message, id: 76
QDEBUG : RadioConfigTest::_fullCalibration_test() MockLink: Unhandled mavlink message, id: 76
QDEBUG : RadioConfigTest::_fullCalibration_test() MockLink: Unhandled mavlink message, id: 76
QDEBUG : RadioConfigTest::_fullCalibration_test() Deleting UAS object:  "MAV 128"
PASS   : RadioConfigTest::_fullCalibration_test()
PASS   : RadioConfigTest::cleanupTestCase()
Totals: 4 passed, 0 failed, 0 skipped, 0 blacklisted
********* Finished testing of RadioConfigTest *********
ALL TESTS PASSED
After app delete
```

dogmaphobic:
Out of ideas...

src/AutoPilotPlugins/PX4/SensorsComponent.qml
DonLakeFlyer: These can just go to medium font

DonLakeFlyer:
I'll look at TC. As far as all the specifically numbered font sizes all of those should go away and we should stick with large/medium/default. Most of the left over ones are just ones that I haven't gotten around to converting yet. I'll fix that up after Stable.

dogmaphobic:
> As far as all the specifically numbered font sizes all of those should go away and we should stick with large/medium/default.

There are a few places that need specific sizes. The toolbar, most HUD gadgets. They all use *specialized* sizes. Yes, in most everywhere else, we should stick to large/medium/default, but we can't get rid of the numbered sizes.

DonLakeFlyer:
> They all use specialized sizes.

I'd use specifically named sizes for those as well. ToolbarTextFontPixelSize or something like that so they are easy to find and can be changed all in one place without affecting other code that may be using the same pixel size.

dogmaphobic:
> I'd use specifically named sizes for those as well.

Hmm... Let me take a look. There shouldn't be that many cases and this might be possible.

DonLakeFlyer:
I can repro the Ubuntu TC problem. May need to build a debug Qt on Ubuntu to figure it out. Sigh!

dogmaphobic:
> I can repro the Ubuntu TC problem. May need to build a debug Qt on Ubuntu to figure it out. Sigh!

The only difference I can see is the GCC version:

TC:
```
********* Start testing of FlightGearUnitTest *********
Config: Using QtTest library 5.4.1, Qt 5.4.1 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 4.8.2)
```

Mine:
```
********* Start testing of FlightGearUnitTest *********
Config: Using QtTest library 5.4.1, Qt 5.4.1 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 4.6.1)
```

dogmaphobic:
That's the gcc used to build Qt. Let me update Qt and see what happens.

DonLakeFlyer:
I think there is something wrong with the way the toolbar qml sets it height. When using a widget holder the size of the widget is determined by the size of the inner qml. If you force the height of the widget holder using setMinimumHeight in MainToolBar constructor is seems to work. So I think the qml inside it ending up with no height set.

DonLakeFlyer:
Or maybe not. How is the height of the toolbar set? The Qml itself doesn't seem to have any dimensions on the outer rectangle.

dogmaphobic:
> How is the height of the toolbar set?

Hmm... It used to be in the constructor of MainToolBar.cc. Now that I look at it, I can't seem to find it. It's supposed to be 40 pixels high.

DonLakeFlyer:
If you set the minimum width to 40 it'll probably all work

dogmaphobic:
Got it. It got removed in the clean up. Putting back on...

PR #1658, opened 2015-06-09
[Do Not Merge, testing only] Firmware Upgrade: APM Flight Stack, 3DR Radio

DonLakeFlyer:
Still a WIP, but about 90% complete at this point. Putting up here to get some testing from folks and gather comments on usability.

This is a major rework of Firmware Upgrade:
- Added support for APM Flight Stack and 3DR Radio
- Simpler workflow for upgrade, just plug the board in and follow the prompts
- For both PX4Flow and 3DR Radio, there are no additional prompts after you plug in
- Non-Stable PX4 Flight Stack builds are hidden behind "Advanced" checkbox
- Better prompting to disconnect from QGC and unplug from USB prior to upgrade

@LorenzMeier Trying to get upgrade to work without needing to disconnect first got too complicated across all the board types. It is possible, and I'll still give it a shot at some point.

Known problems:
- 3DR Radio is still a bit flaky. Timing issue cause it to not work maybe 10% of the time.
- If you select Pixhawk board firmware too quickly you can run into timing issues.


LorenzMeier:
Ok, will test, great stuff!

DonLakeFlyer:
Closing this to make a real PR

PR #1660, opened 2015-06-11
Update firmware upgrade controller links for AeroCore firmware

smithandrewc:
I originally didn't realize we were able to use the cleaner versions of the AWS links.  This pull just brings the links in line with the format of other AWS links.


DonLakeFlyer:
Hold for after stable.

PR #1661, opened 2015-06-11
A few more gibberish into English translations.

dogmaphobic:
A few more gibberish translated to English for the speech-to-text engine.

Plugged a new Pixhawk in for the first time today and got a barrage of gibberish voice alerts.


LorenzMeier:
Looks great! Also happy to fix more things on the FW side. I did fix a whole bunch already, so I hope it led to a noticeable improvement already. My goal is to have a completely TTS compatible message system (with maybe some string replacements like the ones you did on the GCS side).

PR #1662, opened 2015-06-11
XPlane HIL interface: Fix link deletion logic so no QGC reboots are r…

LorenzMeier:
…equired between runs.

This should go into the release. @kd0aij You will want this branch, as it eliminates a lot of the pain with HIL setups. Assuming you have an XPlane CD or dongle this means you only need to reset the autopilot between runs.

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -242,6 +242,8 @@ UAS::UAS(MAVLinkProtocol* protocol, int id) : UASInterface(),
 */
 UAS::~UAS()
 {
+    stopHil();
+    simulation->deleteLater();
     writeSettings();
 }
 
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Rerun through TC to figure out how to fix and get in release

src/uas/UAS.cc
```diff
@@ -242,6 +242,10 @@ UAS::UAS(MAVLinkProtocol* protocol, int id) : UASInterface(),
 */
 UAS::~UAS()
 {
+    stopHil();
+    // wait for the simulator to exit
+    simulation->wait();
```
DonLakeFlyer: @LorenzMeier This crashes when you are not doing a simulation since simulation is NULL. Also not sure about the wait here not waiting forever. What is causing the simulation thread to close such that wait will return if I just shutdown QGC without first stopping the simulation from the widget.

DonLakeFlyer:
Never mind I see it. There is a stopHil above it.

Issue #1634, opened 2015-06-03
Editing UDP communication links doesn't persist modification


AndreasAntener:
If I try to edit an IP address in one of my UDP links, the change doesn't persist. (as soon as the dialog is closed, it has the old IP address again)

dogmaphobic:
It's in #1666

AndreasAntener:
Nice, thanks!

PR #1666, opened 2015-06-12
Fixed a few issues with target UDP hosts.

dogmaphobic:
This fixes issue #1634 .
Not only editing was not persistent, deleting a target was not working either.


LorenzMeier:
merging for release

PR #1669, opened 2015-06-17
Send MAVLink message for all joystick modes

AndreasAntener:
I was looking through the handling here and found that joystick updates only happen if manual mode is set. Might this be a bug or are the other modes unsupported/untested ATM?

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -2578,7 +2578,6 @@ void UAS::setExternalControlSetpoint(float roll, float pitch, float yaw, float t
             const float newThrustCommand = thrust * axesScaling;
 
             // Send the MANUAL_COMMAND message
-            mavlink_message_t message;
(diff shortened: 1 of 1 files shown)


AndreasAntener:
Works at least in HIL for manual.
(But the axis mapping doesn't really work for me)

DonLakeFlyer:
@LorenzMeier @thomasgubler I"ve never used joystick. No not sure if this is correct change or not?

thomasgubler:
I guess this makes sense. This was like your change already before https://github.com/UAVenture/qgroundcontrol/commit/e59b14586a98446dc1050ddb17819b13ccda2554

AndreasAntener:
I was able to use it in manual and attitude mode with this. Shouldn't break anything.

severinleuenberger:
@AndreasAntener Am I right, that this commit is possibly the one who makes Dominik and me very happy? ;-)

AndreasAntener:
@severinleuenberger No this won't make you happy.
As Lorenz wrote to the user list, he implemented the message your system is sending on the beta branch.

PR #1671, opened 2015-06-20
Adding note about USB connection for ESC calibration.

dogmaphobic:
Also adjusting text layout while at it.


LorenzMeier:
Thanks!

PR #1683, opened 2015-06-26
Fixing Qt event loop issue where UDP readReady signal would not work.

dogmaphobic:
For whatever reason, UDP connections would some times stop working. What I've found was that the ```readReady()``` signal would stop being called (outgoing datagrams were being sent just fine though). When the Pixhawk/Companion computer link was set to 57600, this would happen rarely but when the speed was set to 900k, the problem would show up quite often. In the case of the Android build, the issue was consistent at any speed and quite soon. As far as I could tell, Qt was getting stuck on a ```select()``` call and could not recover.

The fix was to stop using the ```readReady()``` signal and running my own loop. In the process, I also moved the *writes* to the link thread (it used to be in the MavLinkProtocol thread).

In addition, I've changed the addressing to ```AnyIPv4``` instead of ```Any```. On Qt 5.5, even though the connection was using IPv4, the *peer address* was being returned using IPv6 scheme and messing everything up. As nothing else in the pipeline supports IPv6, I saw no reason to allow binding to IPv6 interfaces.

### DIFF
--- a/src/comm/UDPLink.cc
+++ b/src/comm/UDPLink.cc
@@ -80,6 +80,7 @@ UDPLink::UDPLink(UDPConfiguration* config)
     #if defined(QGC_ZEROCONF_ENABLED)
     , _dnssServiceRef(NULL)
     #endif
+    , _running(false)
 {
     Q_ASSERT(config != NULL);
     _config = config;
@@ -98,9 +99,12 @@ UDPLink::~UDPLink()
     _config->setLink(NULL);
     _disconnect();
     // Tell the thread to exit
-    quit();
+    _running = false;
     // Wait for it to exit
     wait();
+    while(_outQueue.count() > 0) {
+        delete _outQueue.dequeue();
+    }
     this->deleteLater();
 }
 
@@ -110,8 +114,16 @@ UDPLink::~UDPLink()
  **/
 void UDPLink::run()
 {
-    _hardwareConnect();
-    exec();
+    if(_hardwareConnect()) {
+        while(true) {
+            readBytes();
+            if(_writeBytes() && _running)
+                continue;
+            if(!_running)
+                break;
+            this->msleep(50);
+        }
+    }
     if (_socket) {
         _deregisterZeroconf();
         _socket->close();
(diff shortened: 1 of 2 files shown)


Nyaruko:
May I ask on what system you meet this problem? Since I have used udplink
under windows for quite a long time and never see this problem?

dogmaphobic:
Mac OS, Android and iOS. Someone else (@mhkabir) also had it under Windows.

Nyaruko:
Sorry to bother, but what version of Qt are you using and is this bug also
affecting other sockets, like serialsocket,tcpsocket?

DonLakeFlyer:
Do you think it's ok to spin a loop like this on the tablet? Wondering how much cpu it ends up chewing?

Nyaruko:
I think it should be ok as long as it is on another thread, and the
read/write frequency is not that high (like 50Hz in this case)?

LorenzMeier:
@dogmaphobic We need a better approach unfortunately. We rely on UDP for high-rate telemetry (200-500 Hz). Could you think of a way not requiring an usleep?

LorenzMeier:
Since the data is timestamped we can merge this as stopgap, its just not scaling well to what's actually happening in the system. I presume all incoming packets are emptied from the buffer, not just the last one, correct?

Nyaruko:
There is a bug report here: https://bugreports.qt.io/browse/QTBUG-46552

Seems 5.4.1 is not affected.

LorenzMeier:
@dogmaphobic Could you condition your change on Qt 5.4.1 and merge it? I think we need desperately a stopgap measure.

dogmaphobic:
> Could you condition your change on Qt 5.4.1 and merge it? I think we need desperately a stopgap measure.

I did just that. Note that this bug is also present in Qt 5.5.0, which is frozen. It might be Qt 5.5.1 before we see a real fix. With that said, the loop penalty on data rate is negligible. The loop only sleeps if there is nothing to do. When it does sleep, when it wakes up it proceeds unbuffering everything that might have arrived during the sleep. It would only be an issue if you are relying on realtime data to make decisions on the QGC side. That seems unrealistic given that network latency itself is unpredictable, specially when using cellular network.

dogmaphobic:
Interesting, my loop is actually more CPU efficient than using Qt's signal.

My loop: 9.4% CPU
Qt Signal: 15.7% CPU

screen shot 2015-06-27 at 8 58 48 am

screen shot 2015-06-27 at 9 00 38 am

LorenzMeier:
Merging as stopgap, hope that's fine with you @DonLakeFlyer.

Nyaruko:
So for the daily building, which qt version should I use now?

Could you condition your change on Qt 5.4.1 and merge it? I think we need
desperately a stopgap measure.

I did just that. Note that this bug is also present in Qt 5.5.0, which is
frozen. It might be Qt 5.5.1 before we see a real fix. With that said, the
loop penalty on data rate is negligible. The loop only sleeps if there is
nothing to do. When it does sleep, when it wakes up it proceeds unbuffering
everything that might have arrived during the sleep. It would only be an
issue if you are relying on realtime data to make decisions on the QGC
side. That seems unrealistic given that network latency itself is
unpredictable, specially when using cellular network.

—
Reply to this email directly or view it on GitHub
<https://github.com/mavlink/qgroundcontrol/pull/1683#issuecomment-116028571>
.

dogmaphobic:
> So for the daily building, which qt version should I use now?

For you own local build, that's up to you. The official build (TC) uses 5.4.1 for Mac OS and Ubuntu and 5.4.0 for Windows. I've been using 5.4.2 for desktop builds and 5.5.0 for mobile builds (5.4.x is broken for iOS)

Here is a list of fixes for 5.4.2: https://wiki.qt.io/Change-files-in-Qt-5.4.2

dogmaphobic:
> There is a bug report here: https://bugreports.qt.io/browse/QTBUG-46552

Oh, and thanks for finding this. I spent close to 10 hours on this and I did not find that bug. It would have saved me a bunch of time had I found it earlier :)

DonLakeFlyer:
Bummer on 5.5

Nyaruko:
Sorry to bother, one more little question.
I have read your modified code of UDP link, I noticed that you didn't use
any waitForReadyRead/waitForBytesWritten in the worker thread's while loop.
Is this Ok for UDP link?
In the past I have tried Qt's TCP links by using the qtcpsocket in sync
mode, and they are not working at all without these two calls.

dogmaphobic:
@Nyaruko ```waitForReadyRead()``` is for *buffered* IO. QtUdpSocket is either not buffered or those calls don't do what they are supposed to (they don't work).

On the bright side, the bug report you found above has been elevated to *Critical*, and there is a chance it will make it to 5.5.0 as a result.

dogmaphobic:
Well, 5.5.0 is officially out and QTBUG-46552 didn't make it. I will stick to the current fix until this is resolved.

DonLakeFlyer:
Bummer. I'll cut over master at some point soon.

dogmaphobic:
Crap... Gotta love Qt. Just installed the "official 5.5.0 version" and it's broken for Mac OS (all I tested with so far). Well, sort of. It builds but with thousands upon thousands of warnings because of a wrong check for clang version. Let's hold on with 5.4.x until this thing gets more stable. Master 5.5 is fine though. I thought 5.5.0 was supposed to be frozen over a month ago. This just crept in and it's assigned for 5.5.1.

https://bugreports.qt.io/browse/QTBUG-46833

DonLakeFlyer:
Nice. Can you you follow this and then let me know when I should pull the trigger?

Issue #1681, opened 2015-06-26
We really shouldn't ask our users to manage UDP ports themselves


LorenzMeier:
QGroundControl opened for the longest time a default connection on UDP port 14550. There is literally not a single reason to not do this and require the user to set it up manually, and it did not only break the AR.Drone support, but its also adding back an unnecessary bit of complexity for tablet users.

@dogmaphobic I'm assigning this to you since you did UDP hackery and auto-discovery. I think what we want is to have the link up and running by default, and some method of finding drones on the local network.

dogmaphobic:
I'm not sure I follow. The default port (14550) has not changed. How was the AR.Drone support broken?

As to the auto discovery, the ultimate intent is to have an option for UPD to be always active and published. That way, whenever a *client* supporting auto discovery is started, it automatically connects to QGC.

dogmaphobic:
Tablets will have UDP up and running automatically. Whenever a *client* (a Pixhawk with Zeroconf software) is activated, a pop up will show up asking if you want to connect. I have the same plans in place for Cellular connection. That one will require *registering* with a broker service, which I've also already started, for both sides finding each other in the wide (Internet) network. The goal is to not to have to configure anything. Just start and it all works.

LorenzMeier:
We need the same for the desktop - that's all. I'll try to show you SITL in the next days, which should help a bunch with testing.

DonLakeFlyer:
The key is to do it in such a way that there is not a connection hanging around on the connect dropdown that users will get confused with. Needs to work like plugging in pixhawk to USB: Only there when plugged in. Not there when not plugged in. That way people don't try picking it in the wrong circumstances. Probably need to sniff connections for mavlink.

dogmaphobic:
Besides MavProxy and mavros, both of which requires a great deal of understanding to get it even working, what else connects to QGC using UDP? How is that setup (on the UAV side)?

DonLakeFlyer:
> what else connects to QGC using UDP

Other than AR.Drone I don't know, and I don't really know how that works from the vehicle side of things.

dogmaphobic:
I can't say I had never heard of AR.Drone but until now, I had never paid any attention to it. I went to their web site and I'm not sure I understand what it is. I did notice they use QGC for controlling it (news to me). Does it have a Pixhawk in it? What exactly connects to QGC and how? Is it a normal QGC or some customized version?

dogmaphobic:
[...]

PR #1684, opened 2015-06-27
Automatic UDP Link Configuration

dogmaphobic:
This is in response to the ongoing discussion in #1681. Until we come up with a solution on how to deal with the *Connect* and *Disconnect* buttons in the main tool bar. In the mean time, I've added code that automatically creates a *Default UDP Link*. QGC will not (yet) automatically connect this link. It just automatically adds to the list. This way you don't need to configure a new link in order to use UDP. It will be already added for you to the list. When you click the *Connect* button, it will be there. This is specially useful for mobile devices as the link management UI is still QtWidget based and it sucks on mobile devices.

The next step will be a more fundamental change in the *Connect*/*Disconnect* semantics, which we need to further discuss and come up with a proper solution.

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -357,6 +357,8 @@ void LinkManager::saveLinkConfigurationList()
 
 void LinkManager::loadLinkConfigurationList()
 {
+    bool udpExists = false;
+    bool linksChanged = false;
     QSettings settings;
     // Is the group even there?
     if(settings.contains(LinkConfiguration::settingsRoot() + "/count")) {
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
Once I get my other backed up pulls through: Firmware, Serial Link, Parameter Validation I'll take a look at changing the Connect mechanism to be automatic. This is basically how the new Firmware Upgrade I have works now, you just plug something in and it does the right thing whether it's a Pixhawk, 3DR Radio or Px4 Flow without asking too many questions.

PR #1686, opened 2015-06-28
Moving to Dev 2.6

DonLakeFlyer:


dagar:
If you wanted to work towards automating Ubuntu packaging you could also update these files when the version number is changed

debian/changelog - https://github.com/mavlink/qgroundcontrol/blob/master/debian/changelog
deploy/ubuntu_PPA - https://github.com/mavlink/qgroundcontrol/blob/master/deploy/ubuntu_ppa#L3

DonLakeFlyer:
Yeah. I need to get organized around branching and tagging strategy as well to make it all work. I have it in my head but haven't gotten around to it yet.

dagar:
There are ways to autofill the version/changelog from git with appropriate tagging, but the documentation is somewhat lacking.

PR #1687, opened 2015-06-28
Major re-write of Firmware Upgrade

DonLakeFlyer:
This changes the ui flow of firmware upgrade to hopefully improve usability and ask the least amount of questions possible. Also include new support for:
- Flashing 3DR Radios (3DR Radio setup coming soon)
- Flashing APM Flight Stack as starting point for full APM Flight Stack support in QGC

Tested on Mac and Windows. Not tested on Ubuntu since firmware upgrade through my Ubuntu VM with writable device working correctly through VM requires too many gallons of chicken blood and secret incantations for me to figure out. Should be ok though. Windows tends to be the OS that gives the most problems with firmware.


dogmaphobic:
> requires too many gallons of chicken blood and secret incantations

If you should use ostrich blood instead. Just a few ounces is enough...

PR #1694, opened 2015-07-01
Remove QSerialPortInfo from mobile build

DonLakeFlyer:
This is needed for firmware upgrade which is not part of mobile

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -30,7 +30,9 @@ This file is part of the QGROUNDCONTROL project
 
 #include <QApplication>
 #include <QSslSocket>
+#ifndef __mobile__
 #include <QSerialPortInfo>
+#endif
 
 #include "QGCApplication.h"
 #include "MainWindow.h"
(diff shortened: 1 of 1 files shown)


LorenzMeier:
I would be careful about de-featuring the mobile build too much too soon. I think end users will not want to go back to their desktop for firmware updates. Its very likely not going to happen via USB though, but rather via Wifi (which e.g. Solo already supports, as it does Wifi -> USB internally).

dogmaphobic:
The stuff taken out primarily has to do with serial connections. Android has some limited serial capabilities and iOS has none. Once we have a way to do firmware uploads through TCP/UDP, it would be simple to *turn on* the appropriate parts for mobile.

DonLakeFlyer:
Right. USB firmware upgrade is not technically feasible on mobile. A WiFi based upgrade will be completely different from the client side. The existing code would end up being what you would run on the companion WiFi computer, then the client side becomes more of an FTP server for the image across the connection.

Issue #1701, opened 2015-07-03
Adding a Warning to BAT_V_LOAD_DROP


orangelynx:
Hi,

with the latest fix to the remaining battery estimation algorithm (see https://github.com/PX4/Firmware/issues/2523) one can now accidentally cause the battery to be discharged to critical levels without being warned about it by setting BAT_V_LOAD_DROP incorrectly.

I therefore suggest adding an appropriate warning to the parameter to prevent unexperienced users from igniting their batteries.

PR #1703, opened 2015-07-04
Warning and explanation for voltage drop.

dogmaphobic:
Added some explanation and a warning on the use of *Voltage Drop* as discussed on issue #1701. I'm also now showing the computed compensated minimum voltage under load.

screen shot 2015-07-04 at 12 24 36 pm

### DIFF
--- a/src/AutoPilotPlugins/PX4/PowerComponent.qml
+++ b/src/AutoPilotPlugins/PX4/PowerComponent.qml
@@ -310,8 +310,9 @@ QGCView {
                 visible:        showAdvanced.checked
             }
             Rectangle {
+                id: batteryRectangle
                 width: parent.width
-                height: 40
+                height: 140
                 color: palette.windowShade
                 visible: showAdvanced.checked
                 Column {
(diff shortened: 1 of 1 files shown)


LorenzMeier:
The printing looks quite broken in your screenshot - can we do something about only showing the relevant digits?

dogmaphobic:
> The printing looks quite broken in your screenshot - can we do something about only showing the relevant digits?

:) See #1704

LorenzMeier:
The explanation is great, but its exceeding what an average user will actually read, nor is series resistance helping average Joe - either he already knows how a battery works or he doesn't, and if he doesn't a one-sentence explanation is not going to cut it for him. How about:

"Batteries show less voltage at high throttle. Enter the difference in Volts between idle throttle and full throttle, divided by the number of battery cells. Leave at the default if unsure. If this value is set too high, the battery might be deep discharged and damaged."

We do have in general too much text in QGC. Of course it takes extra time to come up with short text, but  less user support compensates for it 8).

dogmaphobic:
No problem :) I actually struggled to come up with something that wasn't a lecture in EE...

dogmaphobic:
Also updated the screen shot. This was built with Qt 5.4.2 as opposed to 5.5.0 as I had earlier.

dogmaphobic:
While at it, shouldn't the battery capacity also be entered here? How is the remaining battery computed? I think I already asked this before but I can't remember.

LorenzMeier:
The system support coulomb counting. You have to set BAT_CAPACITY to the full capacity.

I think we had the current measurement issue pending - happy to help to fix that.

```
Symbols: x = used, + = saved, * = unsaved
x   BAT_V_EMPTY [7,22] : 3.4000
x   BAT_V_CHARGED [8,23] : 4.2000
x   BAT_V_LOAD_DROP [9,24] : 0.0700
x   BAT_N_CELLS [10,25] : 3
x   BAT_CAPACITY [11,26] : -1.0000
    BAT_V_SCALE_IO [-1,351] : 10000
x   BAT_V_SCALING [257,352] : 0.0046
x   BAT_C_SCALING [258,353] : 0.0124
```

orangelynx:
@dogmaphobic looking good - battery capacity setting should definitely be implemented as well, as coulomb counting can enhance remaining battery estimation very well.

To answer your question: Right now remaining battery life is calculated in two way: once by checking the voltage (taking voltage drop into account) and once by integrating the current that has been measured (coulomb-counting). The integrated current will return the amount of  capacity drawn out of the  battery.

The two percentage values  are compared and the lower one is returned and checked against a threshold (18% -> warning, 9% -> critical). Just need to make sure that both calculations work reliably to avoid early transition to battery failsafe due to wrong result.

PR #1706, opened 2015-07-06
Remove unused files

DonLakeFlyer:


LorenzMeier:
The avrdude folder can also go I think.

PR #1707, opened 2015-07-06
Video streaming

dogmaphobic:
Initial Video Streaming support.

* Mac OS and Linux is fully functional.
* iOS is supported but currently, as iOS is only supported with Qt 5.5.x and the current Qt 5.5.0 is broken, the only way to run is building Qt 5.5 master yourself. In other words, wait...
* Windows is supported ~~in theory but there are still issues with statically linking the gstreamer libraries~~. If someone feels inclined to deal with Windows, that would be great. Otherwise, I will take care of it when I have a chance.
* Android has no support yet.

I've started a Video Streaming *README* but I need this PR to go through first before I can edit it online.

P.S. There were no problems per se with Windows. It's always dynamic linked even though the libraries are in a directory named *static*. You just need to have the gstreamer runtime installed.


dogmaphobic:
Interesting timing. GStreamer now has its own domain (http://gstreamer.com) and a new website with packaged SDKs for the various platforms.

dogmaphobic:
Never mind. They only provide the (old) 0.10 version.

dogmaphobic:
I got the Windows build going and running BUT, there is something wrong with the GL painter in Windows. It would need someone who knows Qt in Windows more than I do, though I plan on taking a look at this at some point.

In the mean time, we should figure out how we want to handle video streaming. For this to work, either the user will have to install the gstreamer runtime or we have to pack it along. For Ubuntu, it's a simple apt-get and those folks should have no problem. For Mac OS, you just install the runtime package and everything works. Windows however, in addition to installing the runtime, it requires setting the PATH to point to the binaries. The installer doesn't do that.

The problem is that *if* QGC is built with video streaming support, it requires these binaries to be installed. If they are not instilled, QGC is not too graceful about it. It simply crashes because it can't find the dynamic libraries.

I don't like the idea of having a program that crashes if you don't have something it depends on installed. This could be resolved by the installer, but it would require quite a bit of work on that end. I'm open to suggestions.

dogmaphobic:
By the way, it's safe to merge as long as the build machine doesn't have gstreamer installed, which most certainly doesn't. If the build doesn't find it, it skips video streaming and it builds without it. For the time being, those who want to use video streaming can build locally.

Ubuntu is a special case because it's very easy to do it. When building the ppa, it would only be a matter of defining gstreamer-1.0 as a dependent and apt-get would do it all for you. It's only an issue for Mac OS and specially Windows.

When I get Android and iOS fully functional, it will include everything in the package so this doesn't apply to mobile.

LorenzMeier:
@dogmaphobic @DonLakeFlyer I think we should package and install the binaries along with the app. User-facing apps should never require a parallel installation of some sort and we're already dragging e.g. the SDL library along - we can do the same with GStreamer. Of course on Linux we use a dependency (we need to properly add that) for apt-get.

dogmaphobic:
For Mac OS, we don’t have an installer, just the app bundle. We could include the gstreamer framework in the same manner other frameworks are bundled with QGC (and fix the internal path to point to *@executable_path*, etc.) What I would do however, is to strip down the framework during the copy. We only need about 10% of it and that would bring down the size considerably. The same is true for the SDL framework. Not only we are copying everything, including the header files, we are not creating the proper symlinks. Instead, everything is duplicated, bloating the final bundle.

For Windows, they provide a merge module we could add to our own installer but we would also need to add a path to it as the installer doesn’t do it. The merge module is also immensely large as it includes everything and the kitchen sink. We might do the same as Mac OS and just “bundle” the DLLs along with QGC.

DonLakeFlyer:
I'd just bundle it in so you don't need to deal with paths.

DonLakeFlyer:
What's the state of this pull? Does it work without requiring any additional installs? Looks like TC is happy, but that doesn't hit everything.

dogmaphobic:
> What's the state of this pull? Does it work without requiring any additional installs? Looks like TC is happy, but that doesn't hit everything.

Yes, it's fine as is. If the build can't find the gstreamer framework/libraries, it ignores video streaming and builds without it. It's safe to merge as it will allow people who want to start playing to build locally. Only when we decide to create official builds will we worry about it. I need to get the installers/bundles done before that can happen.

PR #1710, opened 2015-07-08
Initial setup for Mac Installer (Video Steaming)

dogmaphobic:
This PR (attempts) to bundle the necessary gstreamer binaries into the (Mac OS) app bundle. I've created an installer helper script that handles copying the framework, pruning unused stuff, relocating the dylibs so they can be loaded from within the bundle, etc.

I'm not adding the gstreamer distribution to the GitHub repo. I didn't think it would make sense. Instead, it relies on a system installed framework. This is only relevant to the build machine producing a release bundle. Local builds are not affected by this (they all use the system installed framework). Again, if the build system doesn't find the (system installed) gstreamer libraries, it simply ignores it and builds normally with no video streaming support.

@DonLakeFlyer Note that I could not, for whatever reason build an "installer" build. Even using master, before any of these changes, it produces an app bundle that doesn't find quite a great deal of dependencies. All tests I did for this PR was by manually creating everything. I'm running out of time for figuring out why I can't have an *installer* build done. I will look at it next week.

Here are a few example of dozens of error dialogs that pop up when I try to run the resulting QGC:

screen shot 2015-07-08 at 11 32 50 am

screen shot 2015-07-08 at 11 33 12 am

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+libs/lib/Frameworks/GStreamer.framework
 *.swp
 *.nfs
 CMakeFiles
--- a/QGCApplication.pro
+++ b/QGCApplication.pro
@@ -123,18 +123,6 @@ include(libs/qextserialport/src/qextserialport.pri)
 include(QGCExternalLibs.pri)
 
 #
-# Post link configuration
-#
-
-include(QGCSetup.pri)
-
-#
-# Installer targets
-#
-
-include(QGCInstaller.pri)
-
-#
 # Main QGroundControl portion of project file
 #
 
@@ -702,3 +690,16 @@ AndroidBuild {
         $$PWD/android/src/org/qgroundcontrol/qgchelper/UsbDeviceJNI.java \
         $$PWD/android/src/org/qgroundcontrol/qgchelper/UsbIoManager.java
 }
+
+#-------------------------------------------------------------------------------------
+#
+# Post link configuration
+#
+
+include(QGCSetup.pri)
+
+#
+# Installer targets
+#
+
+include(QGCInstaller.pri)
--- a/QGCInstaller.pri
+++ b/QGCInstaller.pri
@@ -18,6 +18,7 @@
 # -------------------------------------------------
 
 installer {
+    DEFINES += __installer_version__
     MacBuild {
         VideoEnabled {
             # Install the gstreamer framework
--- a/QGCInstaller.pri
+++ b/QGCInstaller.pri
@@ -19,6 +19,19 @@
 
 installer {
     MacBuild {
+        VideoEnabled {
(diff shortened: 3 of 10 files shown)


DonLakeFlyer:
Did you build installer from QtCreator or command line. I think there may some issue with doing it from QtCreator. Also could you try installing a daily build to see if you end up with the same Qml errors?

dogmaphobic:
> Did you build installer from QtCreator or command line. I think there may some issue with doing it from QtCreator. Also could you try installing a daily build to see if you end up with the same Qml errors?

Yep, I did everything from within Qt Creator. I have about an hour left to do all this. I will try :)

dogmaphobic:
Ha! TC has GStreamer installed for Ubuntu :)

```
[09:57:25]W:	 [Step 6/20] Project WARNING: No MAVLink dialect specified, only common messages supported.
[09:57:25]W:	 [Step 6/20] Project MESSAGE: Skipping support for XBee API (64-bit Linux builds not supported)
[09:57:25]W:	 [Step 6/20] Project WARNING: Skipping support for 3DConnexion mice (missing libraries, see README)
[09:57:25]W:	 [Step 6/20] Project MESSAGE: Skipping support for RT-LAB (unsupported platform)
[09:57:25]W:	 [Step 6/20] Project MESSAGE: Including support for speech output
[09:57:25]W:	 [Step 6/20] Project MESSAGE: Skipping support for Zeroconf (unsupported platform)
[09:57:25]W:	 [Step 6/20] Project MESSAGE: Including support for video streaming
```

Fix coming up...

dogmaphobic:
> Did you build installer from QtCreator or command line. I think there may some issue with doing it from QtCreator. Also could you try installing a daily build to see if you end up with the same Qml errors?

I did a build from the command line and had the same errors. I have not had the time to dig further. On the other hand, the downloaded build works fine (http://qgroundcontrol.s3.amazonaws.com/QGroundControl-Stable-V2.5.2.dmg)

dogmaphobic:
> I did a build from the command line and had the same errors. I have not had the time to dig further.

What is the exact command line you use for qmake?

dogmaphobic:
I run:
```
mkdir foo
cd foo
~/local/Qt/5.4/clang_64/bin/qmake -r -spec macx-clang CONFIG+=installer ../qgroundcontrol/qgroundcontrol.pro
```

dogmaphobic:
For a release build, which includes the GStreamer framework within the app bundle, the QGC sets up a series of environment variables for GStreamer to find things. This should only happen for such builds. I've added an ```__installer_version__``` define, which is only set when an *installer* build is run. Regular builds don't set any environment variables and rely on GStreamer being installed on the system.

PR #1712, opened 2015-07-08
Add Gumstix AeroCore board support to firmware flashing.

smithandrewc:
Add in AeroCore support to the new firmware upgrade utility.


DonLakeFlyer:
Sorry, forgot to ping you with my latest firmware re-work. I knew I lost something along the way.

PR #1724, opened 2015-07-20
Update to Flight Mode descriptions

DonLakeFlyer:
@LorenzMeier Here it is for your review. One more question: Should fixed wing use Roll/Pitch/Yaw terminology for sticks or Aileron/Elevator/Rudder?


DonLakeFlyer:
Oops, forgot one more thing. On AltCtl and PosCtl, Roll stick controls "turn" rate, not "yaw' rate? Correct?

LorenzMeier:
If you talk about fixed wing, then yes, its the turn rate.

PR #1737, opened 2015-07-24
Update GST Framework path for 1.5.2

LorenzMeier:
@dogmaphobic I needed this to get it to compile.

### DIFF
--- a/src/VideoStreaming/VideoStreaming.pri
+++ b/src/VideoStreaming/VideoStreaming.pri
@@ -32,7 +32,7 @@ LinuxBuild {
     }
 } else:MacBuild {
     #- gstreamer framework installed by the gstreamer devel installer
-    GST_ROOT = /Library/Frameworks/GStreamer.framework
+    GST_ROOT = /Library/Frameworks/Versions/1.0/GStreamer.framework
     exists($$GST_ROOT) {
         CONFIG      += VideoEnabled
         INCLUDEPATH += $$GST_ROOT/Headers


dogmaphobic:
Hmm... That's not a standard path. It seems everything got installed in the root of /Library/Frameworks. This is what the installation should look like:
```
berlin:~$ ls -l /Library/Frameworks/GStreamer.framework/
total 40
lrwxr-xr-x  1 root  wheel   25 Feb 23 02:44 Commands -> Versions/Current/Commands
lrwxr-xr-x  1 root  wheel   26 Feb 23 02:44 GStreamer -> Versions/Current/GStreamer
lrwxr-xr-x  1 root  wheel   24 Feb 23 02:44 Headers -> Versions/Current/Headers
lrwxr-xr-x  1 root  wheel   26 Feb 23 02:44 Libraries -> Versions/Current/Libraries
lrwxr-xr-x  1 root  wheel   26 Feb 23 02:44 Resources -> Versions/Current/Resources
drwxr-xr-x  4 root  wheel  136 Feb 23 02:44 Versions
```
I assume you installed the package http://gstreamer.freedesktop.org/data/pkg/osx/1.5.2/gstreamer-1.0-devel-1.5.2-x86_64.pkg, correct? I will remove mine and install again to make sure I had not done something weird in between...

dogmaphobic:
Yes, installing that alone gave a different directory structure. I'm downloading the other packages to see if I can figure out what I did. Don't merge this yet as the directory you used is wrong one way or the other.
```
/Library/Frameworks/Versions/1.0/GStreamer.framework
```
This does not exist (and it should not exist).

dogmaphobic:
I've updated the readme showing you actually need to install two packages. With that said, there is something wrong with your installation. In /Library/Frameworks, there should only be *Frameworks*, which are basically, directories with the extension *.framework*. You need to remove the "Versions" you currently have there and reinstall both packages. Before doing it though, paste here the results of:
```
ls -l /Library/Frameworks/
```
so I can take a look. The directory structure should be as I pasted above. I also noticed that the *Headers* symlink is not created by the installer. I did it by hand. I will fix that in a different PR.

dogmaphobic:
> I also noticed that the Headers symlink is not created by the installer. I did it by hand. I will fix that in a different PR.

That's not true. The *Headers* directory **IS** properly created by the installer. I just missed it when scanning the directory structure. No need to change anything. Just cleanup your current installation, install both packages, remove this PR changes and it should all work.

- http://gstreamer.freedesktop.org/data/pkg/osx/1.5.2/gstreamer-1.0-1.5.2-x86_64.pkg
- http://gstreamer.freedesktop.org/data/pkg/osx/1.5.2/gstreamer-1.0-devel-1.5.2-x86_64.pkg

DonLakeFlyer:
Two question:
1) Is this ready to merge?
2) Does this need to get into Stable 2.6 branch?

LorenzMeier:
I'm cross testing now, but looks like I installed the wrong package earlier, so please hold off.

LorenzMeier:
Ok, the README fix with the correct packages did all I needed. Closing.

Issue #1718, opened 2015-07-15
Firmware upgrade sync error


AndreasAntener:
With master from just now, I get the following error on the firmware upgrade:
```
Plug in your device via USB to start firmware upgrade
Found device: Pixhawk
Error: Sync: Send Command: Invalid sync response: 0x21 0x20
Upgrade cancelled
```

LorenzMeier:
@DonLakeFlyer I get the same error.

DonLakeFlyer:
Why would the bootloader be sending back 21/20? That's PROTO_GET_SYNC/PROTO_EOC! Are you running normal bootloaders?

AndreasAntener:
Nothing special there.

DonLakeFlyer:
100% repro? Everything ok when I try.

DonLakeFlyer:
@LorenzMeier I can't even see how the bootloader could have sent these bytes back. There is no code that does that in bootloader. You may need to help out here and step through it in the debugger to see what is going on since I can't repro on any of my boards.

AndreasAntener:
I just tried with a fresh Pixhawk from 3DR and it works there. So my other one must have an older bootloader?

DonLakeFlyer:
Possibly, but it should still work. You have 100% failure on your other board? Do you build firmware source? Wondering is px4-fmuv2_default upload works ok, but QGC fails?

LorenzMeier:
Default definitely works in that state

DonLakeFlyer:
@LorenzMeier Ok. You're going to have to help with debugging here. Without a repro on my side I can't do much. I look at the bootloader and the qgc upgrade code to look for possibilities but couldn't find anything that jumped out from code inspection.

DonLakeFlyer:
Can you run QGC with the --full-logging command line option and email me the log? May get lucky. But still think I won't be able to fix this on my end since I can't repro.

AndreasAntener:
Should I find a log file somewhere or just the console out?

I just noticed a difference between the 2 Pixhawks: the "old one" doesn't stay at the bootloader but boots up normally, then QGC reports the sync error.

LorenzMeier:
Then it doesn't know about the reboot into continuous bootload mode (reboot -b) flag.
@DonLakeFlyer This suggests that the timing broke and the QGC code is now too slow to catch the boot loader.

DonLakeFlyer:
[...]

PR #1744, opened 2015-07-27
Handle older bootloaders for Firmware Upgrade

DonLakeFlyer:


DonLakeFlyer:
This is hopefully the real fix for Issue #1718

PR #1746, opened 2015-07-29
MainToolBar blank on Ubunto 15.04

beigenmann:
MainToolBar black


LorenzMeier:
Looking good. Waiting for TeamCity build test.

LorenzMeier:
@DonLakeFlyer @dagar Looks like we will want to have this in the stable as well, as it fixes the broken Ubuntu experience.

PR #1750, opened 2015-07-30
Airframe meta information

LorenzMeier:
This adds airframe meta info handling from Firmware, which auto-exports all available configurations:

screen shot 2015-07-31 at 11 30 11

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -137,9 +137,6 @@
         <file alias="QGroundControl/FlightControls/qmldir">src/ui/qmlcommon/qmldir</file>
         <file alias="QGroundControl/FlightControls/QGCWaypoint.qml">src/ui/qmlcommon/QGCWaypoint.qml</file>
     </qresource>
-    <qresource prefix="/AutoPilotPlugins/PX4">
-        <file alias="ParameterFactMetaData.xml">src/AutoPilotPlugins/PX4/ParameterFactMetaData.xml</file>
-    </qresource>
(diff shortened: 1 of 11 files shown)


LorenzMeier:
This is ready to go.

LorenzMeier:
@DonLakeFlyer Not sure if this is not a TC fail (triggered by Qt bugs in the QRC file handling). I had similar issues when building locally which "resolved" by deleting the entire build directory. Are you building out of tree and is TC doing that before each build?

LorenzMeier:
Here is the associated Qt bug:
https://bugreports.qt.io/browse/QTBUG-13334

Could you script TC to really wipe out the complete build dir and not just run some sort of clean operation?

DonLakeFlyer:
TC builds from new clean directory. Looking...

qgroundcontrol.qrc
```diff
@@ -247,4 +244,8 @@
     <qresource prefix="/res/audio">
         <file alias="Alert">resources/audio/alert.wav</file>
     </qresource>
+    <qresource prefix="/AutopilotPlugins/PX4">
```
DonLakeFlyer: Unit test is failing because of change in case for AutoPilotPlugins->AutopilotPlugins.

LorenzMeier:
How stupid - I've fixed it and hope it goes through. I've tasked a graphics designer to re-do the airframe images btw. and label all rotors and control surfaces so that we can start using the @output info from the XML to create a mapping table from autopilot to vehicle.

PR #1754, opened 2015-07-31
UAS: Connect message sending via buffered signals

### LINKED ISSUE #1753: Serial port threading broken
When the wrong thread (I need to investigate which one) sends a MAVLink message through the serial link Qt emits this message and stops *completely* sending any data: QSocketNotifier: Socket notifiers cannot be enabled or disabled from another thread

From that point on USB doesn't receive any bytes, although the serial link API reports writing data successfully.

LorenzMeier:
This is an attempt to fix #1753. @kd0aij If you like you can give this a go, I tested it as far as in that it enables HIL mode, but I haven't flown HIL with it yet.

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -165,6 +165,8 @@ UAS::UAS(MAVLinkProtocol* protocol, int id) : UASInterface(),
         componentMulti[i] = false;
     }
 
+    connect(this, &UAS::_sendMessageOnThread, this, &UAS::_sendMessage, Qt::QueuedConnection);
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
The signal on the same thread thing should be on writeBytes in SerialLink.

LorenzMeier:
The problem with that approach is that the signal is buffered and we are using a pointer to a local variable on the stack (the bytes to be sent are on the stack and go out of scope immediately after the call). The buffered signal with blocking wait is not allowed to be used within the same thread, so its not a solution either. We would need to change the API to copy the data instead.

I'll merge this as stopgap so we can fix it for good. Right now at least HIL and probably quite some more code paths are totally broken as QGC stops communicating.

PR #1759, opened 2015-08-04
Airframeconfig usability fix

LorenzMeier:
This fixes a couple of usability issues with airframe config where things had to be done in a certain order and things that intuitively are expected (like clicking on an icon to select) didn't work.

I also fixed timing / race issues with param handling in the process and tested the result quite intensively successfully. When this PR is in its probably a good time to cut a new stable.

### DIFF
--- a/src/AutoPilotPlugins/PX4/AirframeComponentController.cc
+++ b/src/AutoPilotPlugins/PX4/AirframeComponentController.cc
@@ -127,7 +127,7 @@ void AirframeComponentController::_waitParamWriteSignal(QVariant value)
         // Now that both params have made it to the vehicle we can reboot it. All these signals are flying
         // around on the main thread, so we need to allow the stack to unwind back to the event loop before
         // we reboot.
-        QTimer::singleShot(100, this, &AirframeComponentController::_rebootAfterStackUnwind);
+        QTimer::singleShot(800, this, &AirframeComponentController::_rebootAfterStackUnwind);
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/AutoPilotPlugins/PX4/AirframeComponentController.cc
```diff
@@ -127,7 +127,7 @@ void AirframeComponentController::_waitParamWriteSignal(QVariant value)
         // Now that both params have made it to the vehicle we can reboot it. All these signals are flying
         // around on the main thread, so we need to allow the stack to unwind back to the event loop before
         // we reboot.
-        QTimer::singleShot(100, this, &AirframeComponentController::_rebootAfterStackUnwind);
+        QTimer::singleShot(800, this, &AirframeComponentController::_rebootAfterStackUnwind);
```
DonLakeFlyer: Was this not working? Reason I wonder is that the amount of time here shouldn't really matter. It's just to unwind the stack. If it wasn't working then it's likely an indicator that this whole new mechanism I added isn't really working correctly.

src/AutoPilotPlugins/PX4/AirframeComponentController.cc
```diff
@@ -135,8 +135,10 @@ void AirframeComponentController::_rebootAfterStackUnwind(void)
 {    
     _uas->executeCommand(MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN, 1, 1.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0.0f, 0);
     qgcApp()->processEvents(QEventLoop::ExcludeUserInputEvents);
-    QGC::SLEEP::sleep(1);
-    qgcApp()->processEvents(QEventLoop::ExcludeUserInputEvents);
+    for (unsigned i = 0; i < 2000; i++) {
+        QGC::SLEEP::usleep(500);
+        qgcApp()->processEvents(QEventLoop::ExcludeUserInputEvents);
```
DonLakeFlyer: Isn't this a huge amount of wait time? 2000 * (500 usecs)? Maybe I'm reading something wrong.

LorenzMeier: Its 1'000'000 usec which is 1000 ms which is 1 s 8)

src/AutoPilotPlugins/PX4/AirframeComponentController.cc
LorenzMeier: It was definitely not, but because of the test complexity involved I might have screwed up testing. I will re-test again after my other fixes with the original (or lower) number.

src/AutoPilotPlugins/PX4/AirframeComponentController.cc
DonLakeFlyer: Sorry brain not working u == micro, u != milli. Like the new UI changes.

PR #1761, opened 2015-08-05
set QGC version from git

dagar:
This PR uses `git describe` to set the QGC version.

For example the current version displayed for this branch is
v2.6.0-50-gb4f941e

Which means v2.6.0 + 50 commits and git sha b4f941e. A stable version (directly on a tag) would only display the version (eg v2.6.1)

When it's time for a new stable release just drop a new tag. No need to update QGCConfig.h

I'll likely need to tweak this to work on windows.

### DIFF
--- a/src/comm/MAVLinkProtocol.cc
+++ b/src/comm/MAVLinkProtocol.cc
@@ -191,7 +191,7 @@ void MAVLinkProtocol::_linkStatusChanged(LinkInterface* link, bool connected)
         // Send command to start MAVLink
         // XXX hacky but safe
         // Start NSH
-        const char init[] = {0x0d, 0x0d, 0x0d};
+        const char init[] = {0x0d, 0x0d, 0x0d, 0x0d};
         link->writeBytes(init, sizeof(init));
         const char* cmd = "sh /etc/init.d/rc.usb\n";
         link->writeBytes(cmd, strlen(cmd));
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,5 @@ ui_*
 *.so*
 *.moc
 *.prl
+git_version.cpp
+git_version.obj
--- a/QGCApplication.pro
+++ b/QGCApplication.pro
@@ -18,9 +18,10 @@
 # -------------------------------------------------
 
 include(QGCCommon.pri)
+include(git_version.pri)
 
 TARGET = qgroundcontrol
-TEMPLATE =  app
+TEMPLATE = app
 
 # Load additional config flags from user_config.pri
 exists(user_config.pri):infile(user_config.pri, CONFIG) {
--- a/deploy/ubuntu_ppa
+++ b/deploy/ubuntu_ppa
@@ -1,17 +1,28 @@
 #! /bin/bash
 
-version="2.4"
-rsync -av --delete --exclude=.git ~/git/qgroundcontrol/ ~/tmp/qgroundcontrol-${version}/
+tmpdir=`mktemp -d`
+long_version=`git describe --always --tags`
+short_version=`git describe --always --tags | cut -d "-" -f 1,2`
 
-cd ~/tmp
-tar pczf qgroundcontrol_${version}.orig.tar.gz qgroundcontrol-${version}
+version=${long_version:1}
+dir_version=${short_version:1}
+echo $version
 
-cd ~/tmp/qgroundcontrol-${version}/
+debchange -v ${version} --distribution trusty
+
+qmake -r qgroundcontrol.pro
+make git_ver
+
+rsync -av --delete --exclude={.git,build-release,release,Makefile\*} . ${tmpdir}/qgroundcontrol-${dir_version}/
+
+cd ${tmpdir}
+tar pczf qgroundcontrol_${dir_version}.orig.tar.gz qgroundcontrol-${dir_version}
+
+cd ${tmpdir}/qgroundcontrol-${dir_version}/
 debuild -S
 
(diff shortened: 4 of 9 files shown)


LorenzMeier:
Looking forward to this!

DonLakeFlyer:
Can you try to get the Development/Stable label back into this mechanism? I think without it's going to lead to some confusion when people report bugs. The git commit thing is nice but I"m worry it's a bit too techie for most people.

DonLakeFlyer:
I'm going to play around with this concept so I can get both Stable and Development to work automatically

dagar:
@DonLakeFlyer It mainly just needs the windows equivalent of this line.
https://github.com/mavlink/qgroundcontrol/pull/1761/files#diff-bed4e591fd7d9ec94cec0a5331e3266cR37

I don't have a windows machine at the moment so I haven't been able to make any progress.

dagar:
As for the Development/Stable label I honestly don't think we need it, but it's your call. In general shouldn't we steer users away from the development version unless they know what they're doing?

Perhaps the download page should have the stable links, then to get to the daily builds they have to follow a link to an entirely different page.

dagar:
I'll have to rethink this slightly because it will break the Ubuntu PPA where the .git directory isn't present during compile.

DonLakeFlyer:
Wiki already warns people about Daily. Stable versus Development needs to be visibly marked (like in the title-bar as they are now) so that I know what people are running when they report problems.

dagar:
Ready for review.

dagar:
Strange, this works locally on windows 8 and qtcreator, but it's breaking in teamcity.

DonLakeFlyer:
I've been looking a this a bit and I"m not sure it will work. But I'm not very educated on the whole tagging stuff. The problem is that Stable releases are built and tagged on a branch not on master. Whereas Daily is built from master. Will this work in that case?

dagar:
If we don't need to edit a header to set the version then do we still need release branches?

DonLakeFlyer:
You need to move Stable to a branch so you can do point releases on it. Unless there is some other way.

dagar:
You can branch from the tag if you need to do a point release like that.

DonLakeFlyer:
And then the new point release tag needs to be on the branch. It just delays things not working doesn't it?

dagar:
So let's say you tag v2.8.0 on master and then you want to do a v2.8.1. Either you tag v2.8.1 on master again (if appropriate) or you branch from the v2.8.0 tag, apply your fixes and tag v2.8.1. In this specific case your development version on master would be numbered v2.8.0 + #commmits + gGITSHA or you could merge the v2.8.1 branch back into master (where you'd probably want those commits anyway) and then it would be v2.8.1 + #commmits + gGITSHA.

In practice I don't think this will be a problem for this project. You still get stable versioning without any extra work, plus uniquely identifiable development build numbers that tell you roughly how much has happened since a major release and the exact commit.

dagar:
Do you understand why the windows debug build failed this time? I don't see the error.

DonLakeFlyer:
> Do you understand why the windows debug build failed this time? I don't see the error.

Not sure. Tests crashes for some strange reason. I'll re-submit. Also need to look at you other tagging explanation a little more. I think we're good, but not 100% sure yet.

DonLakeFlyer:
I need to debug the windows test to see what's going on.

DonLakeFlyer:
Still fighting with this on Windows. Current guess is that it doesn't like the way that git_version.cpp is compiled, which then causes strange problems with the .exe when linked in.

dagar:
It should be working now. I had to manually use the right CXXFLAGS.

dagar:
To get the right version in teamcity I think you'll need to also grab the tags. Might be easiest to just do a `git fetch --all`.

DonLakeFlyer:
Ok. I want to run this through some tests to see what I get.

src/comm/MAVLinkProtocol.cc
```diff
@@ -191,7 +191,7 @@ void MAVLinkProtocol::_linkStatusChanged(LinkInterface* link, bool connected)
         // Send command to start MAVLink
         // XXX hacky but safe
         // Start NSH
-        const char init[] = {0x0d, 0x0d, 0x0d};
+        const char init[] = {0x0d, 0x0d, 0x0d, 0x0d};
```
DonLakeFlyer: What is this for?

dagar: It's just something small address sanitizer caught.
A few lines after that

    link->writeBytes(init, 4);

I just slipped it in here because it seemed trivial.

DonLakeFlyer: @LorenzMeier This should be 3 not 4 right?

LorenzMeier: It doesn't really matter. It should be *at least* 3.

DonLakeFlyer:
How come on current master there is a v2.7.1 tag, but git describe only returns the v2.6.0 tag?

dagar:
The v2.7.1 isn't a full annotated tag. Try again with --tags to include lightweight tags.

     % git describe --tags
    v2.7.1-71-g16a64ab

If we make sure to use full annotated tags for releases then just plain `git describe` will work.

DonLakeFlyer:
Just double checking: Even if the latest tag is on a branch it will return that one. git describe is branch agnostic right?

dagar:
I wouldn't say branch agnostic, but if you're on a branch and tag a release if will work fine. It just starts at HEAD and follows the commit parent(s) until it finds a tag.

PR #1762, opened 2015-08-06
Notify user of vehicle boot errors

DonLakeFlyer:
This will notify the user of vehicle startup errors. It is triggered once parameter load is completed, success or fail. It shows the user the any of the following severity errors in a modal dialog: MAV_SEVERITY_EMERGENCY, MAV_SEVERITY_ALERT, MAV_SEVERITY_CRITICAL, MAV_SEVERITY_ERROR. Warning them that they should correct these prior to flight.

It also rewords the parameter load error message to try to head the user in the right direction of possibly resolving vehicle startup issues as the root cause, or upgrading to newer firmware.

I don't currently have a board to run this against so I'm just putting this up here now if you want to give it a try against various board boot error case situations. It may not yet be working correctly. Also the wording could likely use some work as well.

### DIFF
--- a/src/FactSystem/ParameterLoader.cc
+++ b/src/FactSystem/ParameterLoader.cc
@@ -29,6 +29,7 @@
 #include "QGCLoggingCategory.h"
 #include "QGCApplication.h"
 #include "QGCMessageBox.h"
+#include "UASMessageHandler.h"
 
 #include <QFile>
 #include <QDebug>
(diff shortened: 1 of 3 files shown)


LorenzMeier:
Thanks, will do!

LorenzMeier:
It didn't show the warnings yet:
screen shot 2015-08-06 at 22 23 03

LorenzMeier:
Closing for PR in this repo.

PR #1763, opened 2015-08-06
Boot report

LorenzMeier:
This shows the appropriate message on boot fails.


LorenzMeier:
This can go in once TC passes. Modal dialog works for param load fail and preflight check fail. This should already catch most users.

DonLakeFlyer:
At some point I'll make some sort of real semi-modal ui for this sort of stuff. I have some ideas. The message pop ups are kind of nasty but easy to do for now

PR #1765, opened 2015-08-07
UASWaypointManager: AutoUpdate new Waypoints when updated from other …

oberion:
…mavlink instances.
This is a feature to automatically update the WP in QGC when they are changed on the autopilot possibly over a different mavlink link.
If PX4/Firmware#2653 is going to be integrated into the px4 code, the autopilot will notify all receiver about WP changes over the waypoint_count message. Currently QGC ignores this message if it has not asked for the WP before. Motivated from MAVROS, QGC could automatically update the WPs.

### DIFF
--- a/src/uas/UASWaypointManager.cc
+++ b/src/uas/UASWaypointManager.cc
@@ -48,7 +48,8 @@ UASWaypointManager::UASWaypointManager(UAS* _uas)
       current_partner_systemid(0),
       current_partner_compid(0),
       currentWaypointEditable(),
-      protocol_timer(this)
+      protocol_timer(this),
+	  _updateWPlist_timer(this)
 {
     _offlineEditingModeMessage = tr("You are in offline editing mode. Make sure to save your mission to a file before connecting to a system - you will need to load the file into the system, the offline list will be cleared on connect.");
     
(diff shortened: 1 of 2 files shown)


src/uas/UASWaypointManager.cc
```diff
@@ -1197,3 +1213,13 @@ void UASWaypointManager::_stopProtocolTimerOnThisThread(void)
 {
     protocol_timer.stop();
 }
+
+
+void UASWaypointManager::_updateWPonTimer()
+{
+	while (current_state != WP_IDLE)
+	{
+		Sleep(100);
```
LorenzMeier: This is a windows only thing. You need to use the Qt equivalent.

src/uas/UASWaypointManager.h
```diff
@@ -168,6 +168,7 @@ public slots:
 private slots:
     void _startProtocolTimerOnThisThread(void);                 ///< Starts the protocol timer
     void _stopProtocolTimerOnThisThread(void);                 ///< Starts the protocol timer
+	void _updateWPonTimer(void);
```
LorenzMeier: Code style / indendation.

LorenzMeier:
@DonLakeFlyer TC down or why didn't it pick up?

src/uas/UASWaypointManager.cc
LorenzMeier: Use: QGC::SLEEP::msleep(100);, assuming the intent were 100 ms.

PR #1768, opened 2015-08-08
[WIP] Follow moving targets / waypoints, such as a moving ground vehicle (f…

philipoe:
…irst implementation).

 --- WIP DO NOT MERGE YET ---

This is the PR belonging to https://groups.google.com/forum/?fromgroups=#!topic/px4users/aUZMJmkIBGs . It implements a very basic way to have "moving waypoints" that can represent a ground vehicle to follow. This PR shall serve as a basis for discussion.

<strong>Current implementation:</strong>
- User Interface: A "NAV:Follow" Action / Waypoint type is added. The user can then select a vehicle to follow from a drop-down list (e.g. "GNDVehicle" shall be followed). The use rcan also set a relative North, East and Altitude distance in meters to allow offsets from that vehicle and a loitering radius about the vehicle. He can also set the update frequency, and can select whether the current GNDVehicle-position shall be set as the home position for the UAV that follows. See image below.
- Functionality: Every time a GLOBAL_POSITION message arrives from the GNDVehicle, this (plus the relative offset) will be set as a new waypoint in the GLOBAL frame for the MAV that is following. This also means that the full waypoint or list of waypoints is transmitted every time.

navfollow

<strong>TODOs:</strong>
- [x] Mavlink would need to be adapted (MAV_CMD_NAV_FOLLOW added to MAV_CMD enum). This is also why the current PR does not compile correctly.
- [ ] Pixhawk code should at some point be changed (currently, this PR just uploads a LOITER_UNLIM waypoint, which works as a baseline solution)
- [ ] Issue below needs to be solved to have clean and user-friendly solution.

<strong>Issues:</strong>
Biggest issue is that the waypoint or the full list of waypoints is uploaded every time a new GLOBAL_POSITION arrives. This starts the tedious process of Send->Ack->Send ... of waypoints between QGC and Pixhawk. Also, QGC violently switches between the "edit waypoint" and "view waypoint" windows. Question to @DonLakeFlyer and @LorenzMeier and others: What is the best way to avoid these two issues? Should this be avoided? An idea is that we should introduce a mavlink message /command such as "UpdateWaypoint", that would only update one specific waypoint without all the ACKs and without starting the full waypoint transmit procedure. However, how do we make sure that we update the correct waypoint (with the correct sequence number) in that case?

Any thoughts?

Best,
Philipp


LorenzMeier:
Can you try to fix the errors TC uncovered?

philipoe:
@LorenzMeier This was mainly due to the non-existing MAV_CMD_NAV_FOLLOW definition in the MAV_CMD enum, which I did not want to change in the common.xml mavlink message definition before yours and DonGagne's approval. But its fixed now.

LorenzMeier:
Thanks! I need to have a fresh look tomorrow morning to understand the change. But I'll promise to provide feedback.

DonLakeFlyer:
I haven't had time to look through this in detail. But it looks like you are implementing a bunch of new ui using QWidget based code. All new ui should be using QML.

philipoe:
I'm implementing the slight change in UI in the QGCMissionNavFollow.XXX files exactly as all other files of this type (see https://github.com/mavlink/qgroundcontrol/tree/master/src/ui/mission ) are implemented also in the current master. They don't seem to be implemented in QML yet, right?

LorenzMeier:
@philipoe I see a lot of code style issues with indendation. Could you set your editor to keep the indentation the same or to use tabs / spaces as in the existing files?

philipoe:
@LorenzMeier Done, finally got the hang of this in MS Visual C++...

LorenzMeier:
@philipoe Code-style wise its looking good, but I think I need to understand your use case a bit better. In particular wether going through the GCS is the right approach or if merely the two vehicles should be talking directly.

philipoe:
@LorenzMeier Exactly, this is one of the central questions:

Our use case: PX4-equipped UAV follows a boat that carries a) the ground station and b) (to have some GPS/GPOS position data) a PX4 that is directly connected to the QGC GS via FTDI cable. Therefore, in our case it is no issue to go through the GroundStation.

Having a direct Vehicle<->Vehcile communication w.r.t. waypoint following was not implemented because
a) The route via the GS offers a more direct way for the user to intervene and check the waypoints that are being uploaded (i.e. what the operator sees on the screen is almost surely what the UAV is following).
b) PX4 code changes would be required (see below), and the same would apply to other autopilots, while the current implementation would allow every autopilot to do vehicle-following
c) We have (personally) seen comm issue with the 3DR-Radios when three radios (here: 2x vehicles, 1xGS) talk. Comments on this would be appreciated.

However, direct vehicle<->vehicle comms would solve the above described TODO #3 (i.e. all the Waypoint handling ACK/etc issues and the switching in QGC), and would be possible by:
 - Uploading a NAV:FOLLOW WP from the GS to the UAV (as implemented already)
 - Have a handler in mavlink that responds to a received GLOBAL_POSITION message from one vehicle by checking whether a NAV:FOLLOW waypoint exists for that vehicle, and updates the waypoint position for that vehicle accordingly. That would not be hard, but the question is: How do we CLEARLY identify these vehicles? Can we use the MAVLINK System ID for this (is it unique)? The user could then specify this system ID to follow in QGC.

Best,
Philipp

philipoe:
@LorenzMeier @DonLakeFlyer Any feedback on the last question, such that this PR can be finished up?

DonLakeFlyer:
How does the APM stack Follow Me support work? Seems like we should mirror that so I don't need to implement two different things on the ground control side of thinigs.

LorenzMeier:
@philipoe To me going through the user GUI seems like a weird architectural choice. It seems more like the GCS should tell one vehicle to follow the other, and then an onboard component should take over. That will also work better because you can factor in the relative velocity of the two and you can operate at a much higher rate.

philipoe:
I had some time to look into this again.

@DonLakeFlyer On the APM Stack, it is implemented through the ground station, too. Mission planner (http://copter.ardupilot.com/wiki/ac2_followme/) and DroidPlanner (https://3drobotics.com/follow-me-mode/) support this. The follow-me functionality ONLY works with the Arducopter stack (see first link above), at least the Arduplane documentation does not list a "follow me" mode although I in general see no reason why this should not work and I have seen videos of this with fixed-wings on youtube. The whole functionality in Mission planner is implemented at https://github.com/diydrones/MissionPlanner/blob/master/FollowMe.cs#L163 . Every two seconds, it sends an updated waypoint position to the copter. This is thus very very similar to what is implemented in the above pull request.

I have done some further testing here in HIL, and the approach definitely works. There will be more time to test this on actual planes in the next week.

@LorenzMeier One central issue I discovered on a PX4FMU v1.7 in HIL with this approach is however that the SD-card can get messed up: I am sending a new waypoint location every 1-2seconds, and (after several hundred or thousand successful writes) I have seen that saving this WP to the SD-card fails! sdlog2 stops at the same time then, so I expect some kind of collisoin there. After this, one can of course not send ANY waypoints anymore, which is a huge problem! Any thoughts on this?

DonLakeFlyer:
From what I can tell here is how APM supports this (from looking at Mission Planner code):
- They take an NMEA feed coming in through a com port for follow me gps positions
- The firmware is in "Follow Me" mode
- The send a single MAV_CMD_NAV_WAYPOINT on a timer to the vehicle using non-standard mission protocol sequence

In general I don't like much of how this works. If we want to do follow me, my suggestion is to get together on both the PX4 and APM Flight Stacks and implement a real protocol for this. To me that is:
- A standard follow me flight mode
- A new MAV_CMD_FOLLOW_ME
- A new mavlink message for transmitting follow me coordinates, let's say SET_FOLLOW_ME_POSITION
- A new MAV_PROTOCOL_CAPABILITY bit to signal follow me support

Then vehicles (or whatever) could talk to each other directly using the SET_FOLLOW_ME_POSITION message, or the QGC could implement it as well from a NMEA feed.

LorenzMeier:
@DonLakeFlyer Yes, that's the right way to do it and what we'll do on PX4 and push for on DroneCode. The hackery via the waypoint protocol makes absolutely no sense.

LorenzMeier:
And when it comes to follow-me the only involvement of the GCS (except for mode switching) should be that it can stream its own position (http://doc.qt.io/qt-5/location-positioning-qml.html).

DonLakeFlyer:
I"m going to close this out given the discussion

PR #1770, opened 2015-08-11
Added port name matching fallback for detecting the Pixhawk FMUv2 on USB

Sapd:
Unfortunately I couldn't add the other boards as I don't know their serial port description.


LorenzMeier:
Awesome, thanks!

PR #1773, opened 2015-08-13
Hil add sensor noise

scott-eddy:
Added a function to generate zero mean noise with arbitrary variance based on Box-Muller algorithm.  Added class variables for sensor variance and setting functions for these variances. 

Diff looks large as I add my cpp logic in a seperate file and wrote a matlab test harness to ensure statistical properties of my code.


LorenzMeier:
Really nice!
http://dash.oznet.ch/view/wCPzadpyJKeAUcKS2Kcusb

PR #1774, opened 2015-08-13
Catch boot loader on fallback

DonLakeFlyer:


LorenzMeier:
Verified by code review. Looks good, also for FMUv1

PR #1775, opened 2015-08-13
Remove MIN/MAX/TRIM as cal trigger

DonLakeFlyer:
This trigger turned out to not work all the time.

@LorenzMeier Could you give this a test. I don't have a radio. Also still think it would be worth considering setting attitude control mapping parameter defaults to 0 to force a radio cal on new flash or param reset.

### DIFF
--- a/src/AutoPilotPlugins/PX4/RadioComponent.cc
+++ b/src/AutoPilotPlugins/PX4/RadioComponent.cc
@@ -58,7 +58,8 @@ bool RadioComponent::requiresSetup(void) const
 
 bool RadioComponent::setupComplete(void) const
 {
-    // Check for mapped attitude controls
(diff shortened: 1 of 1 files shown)


LorenzMeier:
I have a branch for you here. As the comment indicates it will screw up over some people on the first update. Could you please start checking the mappings for RC cal and we'll phase this in once QGC is checking the mappings?
https://github.com/PX4/Firmware/pull/2696

LorenzMeier:
Note the TC fail - might have been due to other unchecked commits which happened in the meantime.

DonLakeFlyer:
> Could you please start checking the mappings for RC cal

QGC already did that. I just removed the ones that weren't working.

Issue #1772, opened 2015-08-12
Allow the use of the HUD when video overlay is enabled


LorenzMeier:
@dogmaphobic It would be nice if I could choose between the block instruments and retaining the normal HUD / flight instrument when video background is enabled. It doesn't make sense for a gimbal, but its great for forward-looking fixed wing.

dogmaphobic:
Yes, I thought about it too. I will see what I can do...

dogmaphobic:
I assume this can be closed? #1777 dealt with it.

DonLakeFlyer:
Assigning to me since I've changed this all around

jaxxzer:
Is this still an issue?

dogmaphobic:
I don't even remember what this was all about. With that said, being almost 2 years later, I would assume it is no longer valid as just about everything about the Fly view has changed since.

LorenzMeier:
Closing as stale.

PR #1777, opened 2015-08-14
Enabling the option of using Compass and Attitude Widgets or HUD.

dogmaphobic:
Following issue #1772, I've added a way for the user to define the use of *Widgets* or *HUD* for both the Compass and the Attitude Indicator. In the process, I've also cleaned up the naming of those controls to avoid confusion.

*Widget*: A floating UI element that can be moved around and resized
*HUD*: A fixed, overlay in the center of the screen

Here is a screen grab (sorry for the mess in my office) showing with the HUD enabled. Note that HUD and Widgets are exclusive. You can have one or the other but not both. You can however, disable both so nothing is shown.

screen shot 2015-08-14 at 9 16 37 am

The (right-click) context menu now shows the options to enable/disable these UI elements:

screen shot 2015-08-14 at 9 25 21 am


dogmaphobic:
TC failure is on a file I did not change: UAS.cc (and it's Windows only). It would appear master was already messed up before I added these changes. I can take a look at it but it won't be until much later.

dogmaphobic:
The issue was introduced with #1773. These *double* to *float* warnings are typical Windows. You need to add the *f* suffix to the values assigned to a *float* type variables. Otherwise it assumes is a double and it spits out this truncation warning. Hopefully this fixes it (didn't try a Windows build).

DonLakeFlyer:
TC is running again? Not sure how that happened! From this run it looks like master is ok though since this passed?

dogmaphobic:
It did run, it failed (as shown in the red X in a656c2467c93eeac9cad81598dde280d2742036c above) and I fixed it :)

DonLakeFlyer:
@dogmaphobic So you have master fixed up with this right?

dogmaphobic:
> So you have master fixed up with this right?

No, the fix is in this PR. It's in d06c1906eaf9db497c213277230f88006cda174d above.

DonLakeFlyer:
Right that's what I meant

PR #1778, opened 2015-08-15
Show vehicle errors after parameters load

DonLakeFlyer:
This is step 1 of the following changes:
1-Completed in this pull) After parameters finish loading (whether success or fail) show all mavlink error messages to the user in a modeless ui.
2-Coming soon) After that has been done, show subsequent mavlink error messages to the user in the same modeless ui. Do this on some sort of delay in between each message to collect as many messages together into a single display to the user as possible.

The goal of this is two-fold:
- Make mavlink error messages more visible to the user since these severity errors need to be acted upon prior to or during flight.
- Mavlink error messages at boot tend to indicate why parameters are not fully loading as well. So this provides a better indicator as to why that might be happening.

This is what the modeless ui looks like:
screen shot 2015-08-15 at 2 55 25 pm

### DIFF
--- a/src/FactSystem/ParameterLoader.cc
+++ b/src/FactSystem/ParameterLoader.cc
@@ -758,20 +758,25 @@ void ParameterLoader::_checkInitialLoadComplete(void)
     UASMessageHandler* msgHandler = UASMessageHandler::instance();
     if (msgHandler->getErrorCountTotal()) {
         QString errors;
+        bool firstError = true;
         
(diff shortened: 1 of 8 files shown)


LorenzMeier:
Nice!

DonLakeFlyer:
We can use this modeless ui mechanism for other message display as well instead of the modal we have now. I'll change things over that make sense over time.

PR #1785, opened 2015-08-22
UAVCAN config button

LorenzMeier:
This adds a button in the power config to allow triggering a UAVCAN configuration run.


DonLakeFlyer:
Looks like you're getting the hang of the whole Qml+Controller thing :)

LorenzMeier:
Looks like it, yeah.

PR #1787, opened 2015-08-23
Bump font size on Windows

DonLakeFlyer:
Windows builds should have a font size about the same as OSX and Ubuntu now.


DonLakeFlyer:
Still more font work to do. Fonts look like crap on Windows since Qt isn't using native font rendering. This is due to renderType not being set to native rendering.

Issue #1781, opened 2015-08-20
Report a list of vehicle startup errors


Sapd:
When there is an error on startup QGC just shows: "Errors were detected on vehicle startup [...]".

This alert was shown to me although all Firmware options like Calibrating/Radio/Power/Safety etc. were green & configured.

I don't really know what was the trigger for the message, as it doesn't show up in the alert box, also the message was gone after the ~5th start of the pixhawk.

So QGC should list a blank list of vehicle errors (like @DonLakeFlyer said on gitter)

Edit: Okay i got those error message today again. But everything works fine...

Issue #1739, opened 2015-07-25
Critical message text unreadable outdoors (red on black)


kd0aij:
The message text popup window displays a very small font in red on a black background in both indoor and outdoor view configurations. This is impossible to read outdoors on a 1920x1080 Chromebook. The Chromebook display is very bright; similar to a Nexus 7. Dark text on a light background would be OK.

The text in the Control widget text panel is also unreadable in outdoor  ("light") mode; appears to be black on black.

DonLakeFlyer:
Different now

kd0aij:
still dark red on black here:
![qgc_crit_msg](https://cloud.githubusercontent.com/assets/2300221/11608422/0ae8a2f8-9b26-11e5-9db6-96a1585ff4cf.png)

info view widget timestamp is now light grey on white, very low contrast:
![info_view_widget](https://cloud.githubusercontent.com/assets/2300221/11608442/5c18b6c2-9b26-11e5-9f10-d927ff88b6c7.png)

LorenzMeier:
@kd0aij Your state is again quite outdated. Please update.

dogmaphobic:
It may be outdated but a lot of those colors are still not optimal. I've been meaning to work on those palettes for a while but have not gotten around to it. The *Light* palette is specially hard to distinguish elements (buttons from background for example).

It will take some work as it's not just a matter of fixing the individual colors but also using the proper colors here and there. I will take a look at it soon...

kd0aij:
oops, still haven't figured out how to do a "shadow build" correctly. Thought I had rebuilt latest master, but the second try is still disappointing. @LorenzMeier are you *not* seeing the same red on black?
![qgc_critical_msg](https://cloud.githubusercontent.com/assets/2300221/11609200/5836f5fc-9b3d-11e5-8d5f-204e6c881df5.png)

DonLakeFlyer:
Thought this was different now. It doesn't seem to be using the palette for the background color which is incorrect as well. Same thing for the toolbar messages at the bottom.

kd0aij:
#2356

PR #1790, opened 2015-08-23
Modeless messages

DonLakeFlyer:
- New modeless mechanism now queues messages if a message is alread being displayed
- Messages are batched in .5 second increments
- Once initial error message list is displayed after parameter load any further mavlink error messages are displayed using modeless mechanism

Also fixed Issue #1781


DonLakeFlyer:
This also partially address Issue #1739 since critical messages now show up in the modeless dialog in big text.

PR #1799, opened 2015-08-28
QGC bugfixes 

NaterGator:
We are primarily trying to harden QGC for Windows systems and catch major crashes that have lingered in the win32 environment for a while. 

This first set of patches mostly addresses initial issues I encountered while building with MSVC 2013 as well as the following bugs:
 - QGCFlightGearLink: fix crash when clicking Stop in HIL simulator panel due to thread-unsafe object access
 - LinechartWidget: fix unable to turn off "Ground Time" checkbox in analysis view
 - TimeSeriesData: decrease busy memory management of plotted data in analysis view

We're using builds of QGC with debug symbols in the release configuration to catch and fix crashing bugs as they pop up, so expect additional patches in the future.

### DIFF
--- a/src/VehicleSetup/SetupView.cc
+++ b/src/VehicleSetup/SetupView.cc
@@ -106,6 +106,7 @@ void SetupView::showFirmware(void)
                                              "showFirmwarePanel",
                                              Q_RETURN_ARG(QVariant, returnedValue));
     Q_ASSERT(success);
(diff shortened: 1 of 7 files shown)


DonLakeFlyer:
Huge thanks for the windows fixes. I'll do a code review on these and then merge in.

DonLakeFlyer:
Thanks again for the pull

PR #1801, opened 2015-08-29
Revert "set QGC version from git"

DonLakeFlyer:
Reverts mavlink/qgroundcontrol#1761


DonLakeFlyer:
I needed to revert this because it won't work with the current TeamCity git steps.

PR #1806, opened 2015-08-30
Add vehicle indicator to flight display

DonLakeFlyer:
It's still a proof of concept at this point but what is there does work. Including support for displaying multiple vehicles, with vehicles coming and going.

I got deep enough into to it to familiarize myself with Gus's mapping stuff and now have a plan as to how to integrate Vehicle, MavManager and Mapping all together.

screen shot 2015-08-29 at 11 00 01 pm

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -136,6 +136,7 @@
         <file alias="QGroundControl/FlightControls/QGCWaypointEditor.qml">src/ui/qmlcommon/QGCWaypointEditor.qml</file>
         <file alias="QGroundControl/FlightControls/qmldir">src/ui/qmlcommon/qmldir</file>
         <file alias="QGroundControl/FlightControls/QGCWaypoint.qml">src/ui/qmlcommon/QGCWaypoint.qml</file>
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
@LorenzMeier First stage of what you were asking for...

LorenzMeier:
very cool! Could we switch colors so the plane is white with a coloured outline? Its impossible to see in the field like that.

DonLakeFlyer:
Not as it stands now with the svg I'm using for the indicator. I'll figure something out.

dogmaphobic:
That's something I can do. Let me create a proper indicator vector art (as opposed to the existing png)...

DonLakeFlyer:
@dogmaphobic We need something that will work both over the satellite maps as well as street maps. Also I need two versions, once for active vehicle and one for in-active vehicles.

DonLakeFlyer:
@dogmaphobic Just so we don't step on each other. I'm  moving lots of stuff around in the flight display stuff. So just the icons would be great and then I"ll integrate them in.

dogmaphobic:
> Just so we don't step on each other. I'm moving lots of stuff around in the flight display stuff. So just the icons would be great and then I"ll integrate them in.

I've noticed it :)

PR #1812, opened 2015-08-31
appveyor support (windows ci)

dagar:


dagar:
The debug build is compiling, but seems to just stop and then timeout after 40 minutes.

dagar:
Appveyor is a little slow, but nearly working.

The windows debug build is currently failing in MavlinkLogTest, but I don't see an obvious error.
https://ci.appveyor.com/project/dagar/qgroundcontrol/build/master-1/job/4e88aiswl5ffqn2g

DonLakeFlyer:
I"ve never seen that crash before, even on an intermittent basis. How can you debug something that only happens on Travis? For TC it has been a non-zero occurence that I"ve had to debug in theTC environment to figure out how to get something fixed.

dagar:
If crashes only on travis become a problem we can add some gdb scripting to print the backtrace and upload the core to s3. I'm sure we can do something similar on windows.

For now I'll disable the windows unit test from running so we get get appveyor building on a regular basis. Then I'll work on getting the tests running separately.

DonLakeFlyer:
FYI - How to get windows stacks is here: http://www.qgroundcontrol.org/bugreporting

dagar:
http://www.latestfiasco.com/qgccrash.reg is 404

DonLakeFlyer:
Crap, something has gone berzerk on my domain.

dagar:
How about throwing qgccrash.reg into test/ or something?

DonLakeFlyer:
I can do that as well. Of course now I have to figure out what that was again. My domain seems to have easten it.

DonLakeFlyer:
I can't repro that mavlink problem. I wonder if it was related to #1834 which I fixed.

PR #1816, opened 2015-08-31
Add Mission Items to Fly View

DonLakeFlyer:
Just limited information/functionality at this point but it's a start. Thanks to @dogmaphobic MavManager implementation going all the way down to Waypoints it's fairly easy to add the new Qml to the map.

Directionally the Plan View is going to be for editing only, Not editing and showing the active flight. The Fly View is for showing active flights.

screen shot 2015-08-31 at 4 02 31 pm


DonLakeFlyer:
My plan is to finish Fly View, then re-implement Plan View.

LorenzMeier:
Nice! When you want to draw lines between the points I would recommend a spline, since that's what the vehicles will actually do.

dogmaphobic:
Hey, those waypoint views on the bottom look great! Question though: How do I get my Video back? :)

DonLakeFlyer:
Video is only the Primary Flight Display. Trying to keep things simple. Let me know if you think that is bad. I think I may have broken the menu there. Nee dot re-hook to the Options button.

dogmaphobic:
OK, I think I've been away for too long. So, what is the *Fly* panel now? I can see that I can go to the *Primary Flight Display* using the *Advanced Menu*, which still has the video option. I'm assuming the current *Fly* panel is in transition?

DonLakeFlyer:
Give me a couple hours to get video back in the Options menu. Working on finishing another pull.

dogmaphobic:
No worries, and no rush. I was just trying to understand it. I'm trying to get the current vehicle position so I can figure out the graphics for it. Waiting for a GPS signal down in my office... :)

DonLakeFlyer:
I was also thinking about using a see through version of the vehicle with just the outline to put around the waypoint indicator on the map to show yaw setting.

PR #1817, opened 2015-08-31
WIP: reduce sensor noise variances

kd0aij:
This PR changes the default sensor noise variances to values measured when stationary. 

I was unable to follow the code past UAS::qt_static_metacall... Is that intentional obfuscation? :)

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -147,22 +147,20 @@ UAS::UAS(MAVLinkProtocol* protocol, Vehicle* vehicle) : UASInterface(),
     blockHomePositionChanges(false),
     receivedMode(false),
 
-    // Initialize HIL sensor noise variances to 0.  If user wants corrupted sensor values they will need to set them
-    // Note variances calculated from flight case from this log: http://dash.oznet.ch/view/MRjW8NUNYQSuSZkbn8dEjY
-    // TODO: calibrate stand-still pixhawk variances
-    xacc_var(1.2914f),
-    yacc_var(0.7048f),
-    zacc_var(1.9577f),
-    rollspeed_var(0.8126f),
-    pitchspeed_var(0.6145f),
-    yawspeed_var(0.5852f),
-    xmag_var(0.4786f),
-    ymag_var(0.4566f),
-    zmag_var(0.3333f),
-    abs_pressure_var(1.1604f),
-    diff_pressure_var(1.1604f),
-    pressure_alt_var(1.1604f),
-    temperature_var(1.4290f),
+    // measured pixhawk sensor noise variances
+    xacc_var(1e-4),
+    yacc_var(1e-4),
+    zacc_var(2e-4),
+    rollspeed_var(2.5e-7),	// mpu6000 gyro total RMS noise spec: 2.5e-3 deg/sec = 8.7e-4 rad/sec => variance 7.6e-7
+    pitchspeed_var(2.5e-7),
+    yawspeed_var(2.5e-7),
+    xmag_var(4e-6),
+    ymag_var(4e-6),
+    zmag_var(3e-6),
+    abs_pressure_var(5e-4),
+    diff_pressure_var(1.0f),
+    pressure_alt_var(5e-2),
+    temperature_var(1e-4),
 
 #ifndef __mobile__
     simulation(0),
@@ -2827,21 +2825,6 @@ void UAS::enableHilXPlane(bool enable)
         }
         qDebug() << "CREATED NEW XPLANE LINK";
         simulation = new QGCXPlaneLink(this);
-
-        float noise_scaler = 0.1f;
-        xacc_var = noise_scaler * 1.2914f;
-        yacc_var = noise_scaler * 0.7048f;
-        zacc_var = noise_scaler * 1.9577f;
-        rollspeed_var = noise_scaler * 0.8126f;
-        pitchspeed_var = noise_scaler * 0.6145f;
-        yawspeed_var = noise_scaler * 0.5852f;
-        xmag_var = noise_scaler * 0.4786f;
-        ymag_var = noise_scaler * 0.4566f;
-        zmag_var = noise_scaler * 0.3333f;
-        abs_pressure_var = noise_scaler * 1.1604f;
-        diff_pressure_var = noise_scaler * 0.6604f;
-        pressure_alt_var = noise_scaler * 1.1604f;
-        temperature_var = noise_scaler * 2.4290f;
     }
     // Connect X-Plane Link
     if (enable)
@@ -2986,9 +2969,9 @@ float UAS::addZeroMeanNoise(float truth_meas, float noise_var)
     
     //TODO add bias term that changes randomly to simulate accelerometer and gyro bias the exf should handle these
     //as well
-    float noise = z0 * (noise_var*noise_var); //calculate normally distributed variable with mu = 0, std = var^2  
+    float noise = z0 * sqrtf(noise_var); //calculate normally distributed variable with mu = 0, std = sqrt(var)
     
-    //Finally gaurd against any case where the noise is not real
+    //Finally guard against any case where the noise is not real
     if(std::isfinite(noise)){
             return truth_meas + noise;
     }


src/uas/UAS.cc
```diff
@@ -147,22 +147,20 @@ UAS::UAS(MAVLinkProtocol* protocol, Vehicle* vehicle) : UASInterface(),
...
+    yacc_var(1e-4),
+    zacc_var(2e-4),
+    rollspeed_var(2.5e-7),	// mpu6000 gyro total RMS noise spec: 2.5e-3 deg/sec = 8.7e-4 rad/sec => variance 7.6e-7
+    pitchspeed_var(2.5e-7),
+    yawspeed_var(2.5e-7),
+    xmag_var(4e-6),
+    ymag_var(4e-6),
+    zmag_var(3e-6),
+    abs_pressure_var(5e-4),
+    diff_pressure_var(1.0f),
```
LorenzMeier: You haven't really reduced the differential pressure variance, its non-surprising that you end up with terrible results.

src/uas/UAS.cc
```diff
@@ -147,22 +147,20 @@ UAS::UAS(MAVLinkProtocol* protocol, Vehicle* vehicle) : UASInterface(),
     blockHomePositionChanges(false),
     receivedMode(false),
 
-    // Initialize HIL sensor noise variances to 0.  If user wants corrupted sensor values they will need to set them
```
LorenzMeier: This is the better approach. The noise we want includes vibration.

src/uas/UAS.cc
```diff
@@ -147,22 +147,20 @@ UAS::UAS(MAVLinkProtocol* protocol, Vehicle* vehicle) : UASInterface(),
...
-    pitchspeed_var(0.6145f),
-    yawspeed_var(0.5852f),
-    xmag_var(0.4786f),
-    ymag_var(0.4566f),
-    zmag_var(0.3333f),
-    abs_pressure_var(1.1604f),
-    diff_pressure_var(1.1604f),
-    pressure_alt_var(1.1604f),
-    temperature_var(1.4290f),
+    // measured pixhawk sensor noise variances
```
LorenzMeier: The noise level on the bench is not really relevant, as what we tune the system for is what we see in flight including vibration.

src/uas/UAS.cc
```diff
@@ -2827,21 +2825,6 @@ void UAS::enableHilXPlane(bool enable)
         }
         qDebug() << "CREATED NEW XPLANE LINK";
         simulation = new QGCXPlaneLink(this);
-
-        float noise_scaler = 0.1f;
```
LorenzMeier: The reason I had this in the simulator specific API is that the horrible jitter is a function of noise level * update rate. If you connect a simulator running at 400 Hz the flight-measured noise levels would be perfectly fine, as each peak would be smoothed out before making it back to QGC. Because of only 50 Hz update rate we kind of have to "pre-filter" the noise level to match the bandwidth we have. Consequently the noise scaling is better being done on the per-simulator interface, factoring in the specific limitations.

The only point to not do this is would be if we think that we will never get a simulator capable of doing 200-400 Hz interfaced through QGC. Right now all faster simulators talk directly to the autopilot and bypass QGC.

src/uas/UAS.cc
```diff
@@ -2986,9 +2969,9 @@ float UAS::addZeroMeanNoise(float truth_meas, float noise_var)
     
     //TODO add bias term that changes randomly to simulate accelerometer and gyro bias the exf should handle these
     //as well
-    float noise = z0 * (noise_var*noise_var); //calculate normally distributed variable with mu = 0, std = var^2  
+    float noise = z0 * sqrtf(noise_var); //calculate normally distributed variable with mu = 0, std = sqrt(var)
```
LorenzMeier: Is the comment wrong or the code?

LorenzMeier:
Thanks! As you will have seen I had a ton of feedback, mostly architectural considerations. I'm not sure if my feedback is relevant or if we should merge as-is. Also note that TeamCity failed, so please look at the log and fix the errors it brought up.

src/uas/UAS.cc
kd0aij: Note I said that the observed diff_pressure variance is _independent_ of diff_pressure_var

src/uas/UAS.cc
kd0aij: If you want to simulate vibration, then the levels must be airframe specific, not global defaults. At the least, they should be modeled based on the number, location and type of motors on the vehicle. Are there any other sources of vibration which this was supposed to model?

src/uas/UAS.cc
kd0aij: I'm completely unaware of any "faster simulators" (does that include FlightGear?). Please explain the architecture or point me to adequate documentation on the system design.

src/uas/UAS.cc
LorenzMeier: I would want something in a KISS sense that gets close to real flight conditions. Since we use one model in simulation anyway you can't claim accuracy, so its better to have one "typical" vibration level rather than trying to fit every possible airframe. We want to model only qualitatively, not quantitatively. No need to complicate things we can't model properly anyway.

src/uas/UAS.cc
kd0aij: both were wrong. What is the physical significance of variance*variance?

src/uas/UAS.cc
LorenzMeier: It is and was always all here:
https://pixhawk.org/users/hil

src/uas/UAS.cc
LorenzMeier: @scott-eddy Will have the details on how the noise is modelled and calculated.

src/uas/UAS.cc
kd0aij: 1) What's running at 50Hz? I show Xplane9 at 120Hz.
2) Does FlightGear talk to QGC at 400Hz?  (and how do I reproduce your 400Hz HIL test?)
3) Where does the sensor noise get injected when QGC is bypassed?
4) How do I run HIL without QGC?

LorenzMeier: > 1) What's running at 50Hz? I show Xplane9 at 120Hz.
Most people get only 50 Hz

> 2) Does FlightGear talk to QGC at 400Hz? (and how do I reproduce your 400Hz HIL test?)
JSBSim via UDP could, if you were to implement it. Also Eddy seems to be able to run much faster.

> 3) Where does the sensor noise get injected when QGC is bypassed?
It doesn't. It needs a central simulation hub, which is why I wanted to break it out from QGC.

> 4) How do I run HIL without QGC?
Read the JSBSim HIL docs I linked above.

src/uas/UAS.cc
kd0aij: Aren't the choices for EasyStar, Rascal, etc. meant to simulate different airframes and motor combinations? Vibration modeling should definitely be specific to each aircraft if we want it to be of any value. There is no "typical" magnitude which would be correct for electric vs. I.C. engines, or fixed-wing vs. rotorcraft. So the correct location for this is in the flight simulator, not QGC. 
Perhaps we should back up and examine whether there is a need to corrupt the sensor signals modeled by the flight simulators at this level.

src/uas/UAS.cc
scott-eddy: This is indeed a mistake on my part.  It should be sqrt(variance) not variance*variance.  

This change should also be reflected in tools/RandomNumberVerification testing harness.  The typo made it through the test because variance was misdefined in both the matlab script and the c++ funtion.  

My mistake

src/uas/UAS.cc
LorenzMeier: It seems like the throttle control issue you ran into with the EasyStar sounds like we really have only the manpower to maintain one default config. I think we will do better by doing just one thing and do that well. When someone comes along with different vibration simulation requirements he can add a switch to select different ones. I fear we end up over-designing and offering choice, but then in fact only test and maintain one config.

src/uas/UAS.cc
kd0aij: @scott-eddy What is your HIL configuration for testing sensor noise addition? Is it one of the standard QGC options? I'm getting the impression that FlightGear and Xplane9 behave differently with high levels of noise on diff_pressure. I've also noticed that the QGC code which calculates airspeed from differential pressure is different for the two simulators, and that's probably not right.

scott-eddy: I use the standard QGC FlightGear configuration and I would agree that the default noise levels are a little high.  I've halved all the values, except for roll/pitch/yaw rate, on my local build environment and they now give results that mirror real flight data quite well.  

I agree with @LorenzMeier that a KISS approach for noise values is what most users will want.  Though I'd really like to see a ui widget for being able to set these values using the setter functions in UAS.h:  https://github.com/mavlink/qgroundcontrol/blob/master/src/uas/UAS.h#L330-L381

kd0aij: Which aircraft are you flying in FlightGear? Only the Rascal works for me, and I find it's impossible to stall it. Doesn't seem at all realistic.

scott-eddy: All of the predefined aircraft work for me though of late I have been using a custom aircraft fie that we generated.  

The Rascal does experience stalls on my end.  Will post a log showing the behaviour when I get the bandwidth to run a sim.  Can you do the same showing the Rascal approaching stall and never achieving it.

LorenzMeier:
@kd0aij How about setting noise levels which are about half the original values from eddy but tuned down where necessary? You can use my tuning as a starting point for airspeed. We can take it then from there.

src/uas/UAS.cc
LorenzMeier: Does that mean we can settle for the Rascal, hide all other configs and tune the noise levels for it?

kd0aij: @scott-eddy Here's a log showing an attempted stall sequence with the Rascal110 in FlightGear. Although I did see a stall break this time, it doesn't behave at all like my real Rascal...
https://copy.com/DfK4d90BNu0p3UTk

I'm looking forward to seeing your simulation log.

LorenzMeier:
@kd0aij What is the final conclusion here? What is the default airframe we should focus on for FlightGear?

src/uas/UAS.cc
scott-eddy: @kd0aij Sorry for the delay, long weekend over here and I managed to get away.  I get similar performance from my simulation. 
http://logs.uaventure.com/view/RSXUbQ8V6toGXk3Jg3wdHY

While I can't speak to how the performance is or is not like your rascal the simulator is running correctly.  If you wanted the sim to be more representative of the flight dynamics of your aircraft you would have to change the values of the JSBSim xml file.  These are your aero coefficients and if you look in the file you will see that they are very basic.  High fidelity simulation would require running aero analysis tools like AVL on your aircraft configuration and generating a new JSBSim xml.

LorenzMeier: I think we're over-optimising here. We neither need completely realistic flight dynamics nor a choice of different aircraft. We need a single model whose flight dynamics, including noise, get "close" for the common flight situations *and* which is easy enough to use so average users are able to benefit from it.
Right now we have a single config (the Rascal) which works, and we need to tune the noise distribution so that it matches the in-flight situation as close as possible without throwing the estimator too much off because it runs slower.

We have to be realistic. We neither have the resources to maintain N configs nor do we have the resources to get the noise distribution perfect. By just getting the Rascal "close" we would however do substantially better than what we have now.

scott-eddy:
@kd0aij @LorenzMeier 
I think that the final conclusion here is to continue to focus on the Rascal110 for flight gear.  An examination of the JSBSim aerodynamic tables in the xml file would be welcome to make the model match the vehicle better.

Additionally I think the default noise variances should be halved and the fix of sqrt should be implemented.  A UI widget to set the variances would be a great help for users to, people will likely want sims that remain uncorrupted by noise and then add noise to ensure their firmware will work in a real world case.  

My vote for implementations to be made now: 
1) half variances to keep effect of vibration
2) fix sqrt error
3) keep noise scalar  included in code for different sim setups.

Future WIPs:
1) new JSBSim xml files and documentation of where values are generated from
2) UI widget for noise setting

LorenzMeier:
Sounds like an excellent summary. I would also comment out the non-Rascal airframes on the QGC and Firmware side so we do not let run people into unmaintained configs.

kd0aij:
I think we need to make sure that the final settings accurately reflect the vibration levels in a properly set up electric aircraft, if they are not to be adjustable. I also doubt that adding zero mean white noise to the simulated sensor outputs is the right way to do that anyway. We also need more justification for the amounts of noise added to the pressure sensors, since my Ninox test flight is clearly not a typical case.

scott-eddy:
This conversation is moving in circles and points to the fact that we need to have these values user adjustable.  

I am of the same mindset of Lorenz in that we need something that we can quickly and easily test new firmware on that approximates system noise to some degree (instead of giving you perfect state estimates).  Once your firmware has passed that initial testing you take actual flight data which is when your real analysis can occur.  

The sensor noise that you will see in flight comes from a host of sources and no matter how  "properly set-up" people's aircraft are there will always be variation in noise between airframes.  Indeed the flight environment itself will lead to different noise levels from log to log on the same airframe.   

A final note, additive Gaussian white noise is a convenient noise model as the Bayesian estimation routines being employed here all make this assumption about noise.  By adding GWN we know that the estimation techniques should still operate fine.

With all this in mind I think that my summary above still stands with a high priority on the UI widget for setting the noise.  @kd0aij if you disagree can you answer the following:
1) What is a better model for noise than additive gaussian white noise, and why?
2) How do you suggest getting and analysing data to validate that the final noise values "accurately reflect vibration levels" on an aircraft? 
3) If adding the noise to the simulator outputs is not the correct way to go about injecting noise into the system how do you suggest it be done?

src/uas/UAS.cc
scott-eddy: I agree with this.  It would be nice to re-examine where the aerodynamics came from for the Rascal aircraft.  Ideally we could generate a wiki page about what each of aero parameters are currently, why they are that value, and how to change them for your individual setup if you need more realistic flight dynamics

kd0aij:
@scott-eddy Perhaps you and I mean different things by "noise". Vibration is certainly not AWGN. There is much merit to testing with no AWGN added to IMU sensors, and I'm not sure why you contend that adding white noise to sensor readings at unspecified sampling frequency is important. But I have no information on the methodology you used to extract noise variances from my flight log; maybe I'll understand your motivation better once I've seen that. BTW, did the lack of individual timestamps on sensor readings cause any concern in your analysis?

Currently, I turn on wind and turbulence in the simulation if I want "noisy" roll/pitch/yaw behavior. That at least has a physical basis which is clearly defined.

scott-eddy:
@kd0aij Lets step back a minute and consider this problem from a high level.
The sources of noise I can think of on an aircraft flying with the px4 are:
1) Environmental noise - this noise includes wind gusts and wind turbulence
2) Internal noise

Internal system noise is coming from:
2a) Vibratory noise - this is coming from the motors, propeller flapping, propeller imbalances, etc.
2b) Sensor noise - generally broken into random noise and sensor bias

Environmental noise is what is being modelled by Flightgear when winds and turbulence are enabled.  By default Flightgear's turbulence model is adhering to MIL-STD-1797A, which essentially is describing the Dryden Wind Turbulence model.  The PX4 project does not want to do any work here as it is already implemented.

This leaves us with the task of modelling the internal system noise which is where you, @LorenzMeier , and myself have varying opinions of how to best go about this.  Feel free to correct me if I'm wrong here but my understanding is that you would have us injecting two (or three) separate noise terms, one for vibratory noise, one for random sensor noise, and (possibly) one for sensor bias.  This approach allows you to apply colored noise to account for vibrations, white noise for sensor random noise, and perhaps a random walk for sensor bias.  It would also require the generation of new functions to create colored noise and random walks.  Additionally it would require users to characterize 3 noise values, per sensor, for their individual set-up.

I believe @LorenzMeier  and I are of the mindset that best is the enemy of good enough.  I think that for the sake of simulation lumping all of the internal system noise into a single noise value on each sensor is a 'good enough' approximation.  Almost any user can comprehend a Gaussian distribution and intuitively will understand as they increase the variance on the noise value for a sensor that sensor will perform worse.  This allows people to pretty easily 'tune' the variances of the simulation to their specific system.  

I agree that adding the noise as it stands is not rigorously modelling the physical system, however I think for most users it will be good enough.  If they don't like how it is being done they are free to turn the variances to zero and add noise through the environment in flight gear as you yourself have done.  While I think there is merit in more rigorously modelling the system I think it would fall under a different WIP.  Perhaps adding an 'Advanced noise configuration' is a route to pursue but THIS WIP "reduce sensor noise variances" should do exactly what it set out to do.  We should fix the sqrt issue and reduce the sensor noise variances.  I also agree with Lorenz that we should remove all configurations but the Rascal as we clearly don't have the bandwidth to support multiple configs.  

If you don't want to implement the changes that I'm purposing I can put in a separate PR to have the minor issues fixed now.  This will also allow new PR's for sim features we've discussed like a widget for setting the values and perhaps and Advanced noise config.

kd0aij:
@scott-eddy I think we disagree only on the value of simulating vibration effects by increasing the sensor white noise level. And I have no problem with applying this PR with the reduced noise levels I measured on the bench. If you and @LorenzMeier insist on variances which are 4 to 7 orders of magnitude greater than that, I'd suggest that we first merge this PR, then add the widget you suggest.

kd0aij:
Unable to access the TeamCity build log: 404 error on "Details" link

DonLakeFlyer:
I'm not seeing a run for 1817. I'll resubmit. I'll fix the details link as well.

LorenzMeier:
I've fixed this now in 0d2f331 by addressing the variance / standard deviation bug and by introducing a scaling based off real flight data which makes the remaining variance survivable. I have also enabled it by default, because the PX4 Flight Stack will detect a channel without variance now and declare the sensor invalid, so we need some noise for HIL. Thanks Mark!

PR #1827, opened 2015-09-03
New Joystick config, new Joystick control classes

DonLakeFlyer:
- New Qml based Joystick Config setup screen, works just like Radio Config
- Removed old Settings dialog screens
- Config allows for throttle stick zero throttle position to be at center stick or at full bottom
- Joystick mode selection is hidden behind Advanced check box
- Joystick mode setting is stored per Vehicle (keyed by mavlink system id)
- Joystick calibration settings are stored per Joystick (not Vehicle)
- Reimplemented Joystick internals
- Radio Config is disabled if Joystick is enabled
- FirmwarePlugin interface has new manualControlReservedButtonCount so each firmware type can reserve the number of buttons it uses for the MANUAL_CONTROL mavlink message. The remainder of the buttons are available to set Vehicle actions against.
- All implemented using Qml friendly objects/methods/properties: JoystickManager, Joystick

screen shot 2015-09-02 at 5 36 45 pm


DonLakeFlyer:
Two things:
1) @LorenzMeier This is going to need some real testing. I tested by looking at RC_CHANNELS output since I couldn't get jmmavsim to work for me.
2) This is guaranteed to break the mobile. It's not ifdef'ed correctly for that yet. This pull got so big I wanted to close it out before it go any bigger. I'll fix mobile as soon as this goes through.

DonLakeFlyer:
Mobile was easy so I fixed it in this pull. Built and ran Android.

PR #1829, opened 2015-09-03
Joystick fixes

DonLakeFlyer:
Changes to follow these rules:
COM_RC_IN_MODE = 2
- All config screen enabled and run as if there was normal RC connected
COM_RC_IN_MODE = 1
- Radio Config disabled
- Flight Mode Config disabled

As well as these changes:
- Change Simulate RC joystick mode to Normal
- Disable “Enabled” check box if COM_RC_IN_MODE > 1
- Move enabled to first setting
- Fixed bugs with active joystick changing in config
- Moved Joystick enabled save setting to Vehicle instead of globally on joystick

### DIFF
--- a/src/AutoPilotPlugins/PX4/FlightModesComponent.cc
+++ b/src/AutoPilotPlugins/PX4/FlightModesComponent.cc
@@ -68,12 +68,13 @@ QString FlightModesComponent::iconResource(void) const
 
 bool FlightModesComponent::requiresSetup(void) const
 {
-    return true;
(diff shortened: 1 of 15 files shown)


DonLakeFlyer:
Also forgot to say that Enabling a joystick sets COM_RC_IN_MODE=1. The COM_RC_IN_MODE=2 setting is not exposed through joystick config.

LorenzMeier:
It should *not* modify the value if the current value is already 2.

DonLakeFlyer:
Ah, wasn't thinking about that. For people that have old joystick setup.

DonLakeFlyer:
You should be able to give this pull a try without the last fix. I"m going to bail out for a few hours and get back to this.

Issue #1782, opened 2015-08-20
QGroundControl Doesn't Work, Again.


stevec5000:
QGC wouldn't work at all when I first tried it the middle of 2014 but eventually it was working fine in Win 7 after about 6 months of bug fixes and updates with ver.2.2.0.  However, I just tried updating to the latest ver. 2.7.1 and all I get is a black window with nothing in it and no way to do anything!
It looks like all the features are gone again, I can't see any way to program the PX4, calibrate the sensors or the R/C transmitter or connect to the aircraft!
Where can I find an older version that still works?  Thanks.

Sapd:
Looks like there is a problem on your system, because the latest version should work without any problems. Try the 2.5.2 stable http://qgroundcontrol.org/oldstable
Does this version work for you? And if not, do other QT applications work for you? (E.g. teamspeak)

LorenzMeier:
This is because Qt moved to OpenGL hardware acceleration and your graphics driver is not recent enough to support it. There will be support for old hardware in the next Qt version, which is scheduled to be released in September. Meanwhile upgrading your graphics drivers might help.

stevec5000:
So QT has screwed up everything again?   You've got to be kidding!  Win 7 has the latest updates and thankfully there is nothing else using QT.  QGC just depresses me every time I've tried to use it!

stevec5000:
Sapd, ver. 2.5.2 didn't work, ver. 2.5.0 and 2.4.0 didn't work either.  All I get is a black screen with a green button in the upper left that says Activate Engine but doesn't do anything.  Ver 2.3.0 seemed to work and I was able to update firmware and check the R/C even though I can't read most of the fuzzy orange text on the screen.  When I tried to calibrate the compass it went crazy though.

dogmaphobic:
I had not booted Windows in quite a while. I just booted it, built the current master and had no issues. I would assume if there was a real issue, a lot more people would have complained by now. There is nothing inherently wrong with Qt. I can only assume the issue is with your computer.

![windows](https://cloud.githubusercontent.com/assets/749243/9414200/9eb814a2-4804-11e5-939a-1f1518e1277e.png)

NaterGator:
@stevec5000 you could install the latest stable qgc and open a command prompt in the application directory then run:
```
set QT_LOGGING_RULES=qt.qpa.gl=true
qgroundcontrol.exe
```

[...]

PR #1835, opened 2015-09-04
Update NSIS script to include ANGLE dlls for windows, add NSIS script…

NaterGator:
… to otherfiles on windows

We have some machines with buggy / non-compliant OpenGL implementations from Intel (HD3000, for example). They'll run for a short (variable) time and then crash inside the Intel OpenGL dlls. 

I have been manually overriding these machines with QT_OPENGL=angle set in the environment, but there are also explicitly [blacklisted devices in mainline Qt](https://github.com/qtproject/qtbase/blob/89cb92f838b43123f51bfddba433bfe54c0e855d/src/plugins/platforms/windows/openglblacklists/default.json). If we don't ship the DLLs added in this PR then we can't fall back to ANGLE or software OpenGL and we'll likely crash every time. Including these does increase the size of the installer a bit but I think it is a worthwhile tradeoff. This may help cases like #1782 as well.

### DIFF
--- a/QGCInstaller.pri
+++ b/QGCInstaller.pri
@@ -49,5 +49,6 @@ installer {
         QMAKE_POST_LINK += $$escape_expand(\\n) $$quote("\"C:\\Program Files \(x86\)\\NSIS\\makensis.exe\"" /NOCD "\"/XOutFile $${DESTDIR_WIN}\\qgroundcontrol-installer-win32.exe\"" "$$BASEDIR_WIN\\deploy\\qgroundcontrol_installer.nsi")
 		#QMAKE_POST_LINK += $$escape_expand(\\n) $$QMAKE_COPY qgroundcontrol.pdb $${DESTDIR_WIN}
 		#QMAKE_POST_LINK += $$escape_expand(\\n) del qgroundcontrol.pdb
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
I didn't think  QT_OPENGL=angle would work till we switched to Qt 5.5? Is it working for you with Qt 5.4?

NaterGator:
Don: per [this](https://blog.qt.io/blog/2014/11/27/qt-weekly-21-dynamic-opengl-implementation-loading-in-qt-5-4/) blog post Qt 5.4 suffixless packages should support dynamic OpenGL. I am building using the latest stable Qt release (5.5.0) but the aforementioned blog post led me to believe this would be supported using Qt 5.4 as well.

DonLakeFlyer:
Ahh, thanks. Didn't realize this ws in 5.4. I just remembered there is still a problem though. There is code in QGC which uses OpenGl directly (a couple of the lesser used widgets). This code will fail without a real OpenGl implementation and needs to be converted away from OpenGL. I'll need to convert those over and then we can start using this mechanism. I'm going to leave this pull open for now, until I get to fixing the widgets. And then I'll change everything over to the suffixless packages for windows. Thanks for the pull and the information.

PR #1838, opened 2015-09-05
Decreasing the angle jitter threshold for row, pitch and yaw.

dogmaphobic:
The threshold was set to 1 degree, which was causing the display to jump a bit.
While at it, I've also changed the airspeed jitter threshold to make it consistent with ground speed.


LorenzMeier:
Nice!

PR #1841, opened 2015-09-06
fix linux unittests failure

dagar:
On travis-ci linux the unit tests assert because QStandardPaths::DocumentsLocation is `/home/travis/Documents` which doesn't exist. Can we just remove the assert?
`documentsDir.mkpath(_defaultSavedFileDirectoryName)` will create the parent directory if it's missing.

    QDir documentsDir(documentsLocation);
    //Q_ASSERT(documentsDir.exists());

    bool pathCreated = documentsDir.mkpath(_defaultSavedFileDirectoryName);

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ before_install:
 install:
   - if [[ "${TRAVIS_OS_NAME}" = "linux" || "${TRAVIS_OS_NAME}" = "android" ]]; then
          sudo apt-add-repository -y ppa:ubuntu-toolchain-r/test
-      && sudo apt-add-repository -y ppa:beineri/opt-qt541
+      && sudo apt-add-repository -y ppa:beineri/opt-qt542
       && sudo apt-get -qq update
       && sudo apt-get -qq install g++-4.8 libc6-i386 qt54tools qt54base qt54declarative qt54serialport qt54svg qt54webkit qt54quickcontrols qt54xmlpatterns qt54x11extras qt54websockets qt54sensors qt54script qt54quick1 qt54multimedia qt54location qt54imageformats qt54graphicaleffects qt54connectivity espeak libespeak-dev libopenscenegraph-dev libsdl1.2-dev libudev-dev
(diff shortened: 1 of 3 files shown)


dagar:
I have no explanation for this, but even if I create the directories manually travis-ci still hits this assert. If I remove the assert the tests run and pass.

    ASSERT: "documentsDir.exists()" in file src/QGCApplication.cc, line 393
    /home/travis/build.sh: line 41: 14186 Aborted                 ./debug/qgroundcontrol --unittest

We could just remove the assert because mkpath will create the parent directory (documentsDir).

        QDir documentsDir(documentsLocation);
        //Q_ASSERT(documentsDir.exists());

        bool pathCreated = documentsDir.mkpath(_defaultSavedFileDirectoryName);

DonLakeFlyer:
I think this can happen when you have a permission problem. Some folks have hit it when the ran QGC under sudo and then ran it without sudo and it fails. That said it should be a qWarning not an assert (in general I"m moving almost all ASSERTs to qWarning).. But it probable still needs to be figure out. Tests likely run because we don't have any that use the value yet. But if at some point we do, they will begin to fail as well.

dagar:
I replaced the assert with qwarning and the linux debug build unit tests are running properly.

DonLakeFlyer:
That's only a temp fix. The unit tests shoudl be running with QT_FATAL_WARNINGS as an evironment variable set. Sorry forgot to tell you about that. I think for now we could leave Linux and maybe Windows with QT_FATAL_WARNINGS off, and only it turn it on for OSX. OSX tends to be the most stable and we'll caught most issue with just one OS.

dagar:
Added QT_FATAL_WARNINGS for osx.

DonLakeFlyer:
I think I stumbled across the mavlink thing by accident. It may have been memory corruption. I might have fixed it in here: https://github.com/mavlink/qgroundcontrol/pull/1847

dagar:
QT_FATAL_WARNINGS removed

PR #1849, opened 2015-09-08
Android Video Streaming

dogmaphobic:
Added libraries for building video streaming for Android. Not tested as Video Streaming has been removed from the Flight Display and there is currently no way to enable Video Background on mobile devices.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ thirdParty/qserialport/bin/
 thirdParty/qserialport/lib/
 libs/thirdParty/libxbee/lib/
 GeneratedFiles/
+gstreamer-1.0-android*
 
 *.autosave
 .settings/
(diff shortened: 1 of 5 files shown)


LorenzMeier:
Is there a specific reason it got removed? Its probable *the* major feature people want, so even if its just a temporary removal I would want to have it back immediately.

dogmaphobic:
Don's been working on the Flight Display (showing waypoints and whatnot). The original Flight Display is now only available through the *Advanced* menu, which is not available in mobile builds. I didn't have the time to deal with that. Not sure how this *Flight Display* and *HUD* as separate UIs will work.

LorenzMeier:
Then lets work out how to arrange things so we can have a consistent UI experience for those core features.

LorenzMeier:
How about doing it kind of picture in picture and you can toggle the main view by clicking the smaller one?

flight_display
flight_display_video

DonLakeFlyer:
Video streaming is available in the current master using the Primary Flight Display widget.

dogmaphobic:
Closing as it got messed up beyond repair. Opening a new PR for this.

PR #1852, opened 2015-09-08
Add video background support

DonLakeFlyer:
On the Fly view:
- You can switch between video and map background through the Options menu
- Map background shows compass and attitude widgets
- Video background shows compass and attitude HUDs
- Both show waypoint list


DonLakeFlyer:
@dogmaphobic Can you give this a try?

Also should have said that settings persistence is still not worked out across all the various widgets and map selections.

DonLakeFlyer:
FYI: I was going to do the PIP thing but I don't think a small block of video, or small block of map is going to be readable. On a tablet, definitely not. I'd like to try this for now. And we can come back to the PIP thing.

dogmaphobic:
The option is working but it should be disabled (greyed out) if video streaming is not enabled. PR #1855 is now complete and Android has video fully supported. Well, not counting the automated test failures, which utterly confuse me.

PR #1853, opened 2015-09-08
More mission editing work

DonLakeFlyer:
All waypoint type editors are now supported. Getting closer...

screen shot 2015-09-08 at 1 21 15 pm

### DIFF
--- a/src/FactSystem/Fact.h
+++ b/src/FactSystem/Fact.h
@@ -92,6 +92,9 @@ public:
     
     void _containerSetValue(const QVariant& value);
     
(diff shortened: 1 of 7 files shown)


LorenzMeier:
Nice!

DonLakeFlyer:
I found a really cool way to make these little editors generic. I turn the mission values into Fact's with FactMetaData then I just populate an array with those Fact and give it to Qml. Then the Qml just needs to make a Repeater like a mini parameter editor. When I'm finished you'll be able to pull full meta data information on the mission values as well as validation just like the param editor.

PR #1855, opened 2015-09-09
Enable Android Video Streaming

dogmaphobic:
Video on Android is now fully functional.

It took quite a bit of trial an error to figure out the used plugins so I could link them statically but the work was worthwhile as it will be the same for iOS.

Sorry for the messy commit but there were so many changes in between that I got way too out of sync and I completely botched the merge/rebase in #1849. Started it all from scratch with this one.

screen1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ thirdParty/qserialport/bin/
 thirdParty/qserialport/lib/
 libs/thirdParty/libxbee/lib/
 GeneratedFiles/
+gstreamer-1.0-android*
 
 *.autosave
 .settings/
--- a/QGCApplication.pro
+++ b/QGCApplication.pro
@@ -668,12 +668,14 @@ INCLUDEPATH += \
 HEADERS += \
     src/VideoStreaming/VideoItem.h \
     src/VideoStreaming/VideoReceiver.h \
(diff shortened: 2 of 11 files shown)


dogmaphobic:
Should I trust the Travis error? It appears the unittest had an error in FactSystemTestPX4. I'm pretty sure I was nowhere near it. The TC test gives me a 404 when I try to look at the *Details*.

dagar:
I'll restart travis linux debug.

dagar:
https://travis-ci.org/mavlink/qgroundcontrol/jobs/79397074

LorenzMeier:
Nice! Looks like I need to get an Android tablet!

DonLakeFlyer:
I think the TC failure is an intermittent unit test crash I introduced in the last couple days. I"ve resubmitted to see what happens.

dogmaphobic:
> but the work was worthwhile as it will be the same for iOS.

Not true. That would be the case were I to build GStreamer on my own, which I will not. The current distribution for iOS is in the form of a Framework, with one single, monolithic binary. I just tried building iOS with the *current* Qt 5.5.1 and had some build path issues ( but no warnings! :) Waiting for this to merge so I can see what's going on.

PR #1862, opened 2015-09-10
Fixing Android screen orientation.

dogmaphobic:
It now uses the sensor to maintain proper landscape mode.


dogmaphobic:
I don't think I could have caused a problem with the Windows build given that all I changed was a configuration entry for the Android build...

dagar:
The windows build started failing due to a commit just before your build. I'll look into it.
https://github.com/mavlink/qgroundcontrol/commit/c9e685554b284148659b0f272b46bf19382fac76

Issue #1864, opened 2015-09-10
Sizing issues with Android (Nexus 7)


dogmaphobic:
There are still sizing issues with mobile devices.

This is the *Setup* view on a Nexus 7

![screen](https://cloud.githubusercontent.com/assets/749243/9800107/1b2ce1fa-57d9-11e5-8b77-fe8e028bfac7.png)

The *Tilt Indicator* is also too small

![screen1](https://cloud.githubusercontent.com/assets/749243/9800190/96b4f34e-57d9-11e5-845c-42d60115cfd5.png)

PR #1866, opened 2015-09-11
Stop using fixed pixel count spacing

DonLakeFlyer:
Fix for the Vehicle Summary part of Issue #1864.

@dogmaphobic Can you give this a try?

### DIFF
--- a/src/VehicleSetup/VehicleSummary.qml
+++ b/src/VehicleSetup/VehicleSummary.qml
@@ -75,8 +75,8 @@ Rectangle {
         }
 
         Flow {
-            width: parent.width
-            spacing: 10
+            width:      parent.width
+            spacing:    ScreenTools.defaultFontPixelWidth
 
             Repeater {
                 model: multiVehicleManager.activeVehicle.autopilot.vehicleComponents
@@ -84,10 +84,10 @@ Rectangle {
 
                 // Outer summary item rectangle
                 Rectangle {
-                    readonly property real titleHeight: 30
+                    readonly property real titleHeight: ScreenTools.defaultFontPixelHeight * 2
 
-                    width:  250
-                    height: 200
+                    width:  ScreenTools.defaultFontPixelWidth * 28
+                    height: ScreenTools.defaultFontPixelHeight * 13
                     color:  qgcPal.windowShade
 
                     // Title bar


dogmaphobic:
There are still quite a few issues in those screens. I can take a look at them if you want me to. It would help to see what it looks like on your tablet so I can see what changes from one to another. Any one of these screens will do. In general, I would rather have the buttons larger so you don't hit something you don't want by mistake.

The *Summary* screen is better, though I would have a bit more space for the buttons in the left. The icons are way too small:

screen1

The *Airframe* screen is almost OK. The text up on top is a bit piled up and the text in the *Apply and Restart* button almost doesn't fit (I'm assuming the generic plane icon is a place holder?):

screen2

The *Radio Config* screen looks mostly fine too. Again, just the text within the buttons being a bit too large (or the buttons too small).

screen3

The *Flight Modes* screen has the same issue with the text up on top and buttons in general. It doesn't fit.

screen4

Same with the *Sensor Config* screen:

screen5

The *Power* screen is quite messed up:

screen6

Same for *Safety Config*

screen7

The *Parameters* screen is mostly OK (other than the "text too big" issue). The one things that feels weird is that the parameter list not only scrolls up and down (the desired effect), it also scrolls side to side (undesired effect):

screen8

DonLakeFlyer:
Thanks for the pictures Gus. The issue is that my galaxy tab is the same dpi as my Mac. So everything shows up fine. Nexus is double dpi than that so anything using fixed pixel size is out of whack. I'll take care of these. Good mindless work for the road.

DonLakeFlyer:
Should be way better now with the above pull. Still much work to do though. The high dpi test command line works great for wringing these things out on a normal box.

DonLakeFlyer:
@dogmaphobic Another run through on your Nexus would bw great.

dagar:
FYI the current play store beta has these changes.
https://play.google.com/store/apps/details?id=org.mavlink.qgroundcontrol

DonLakeFlyer:
Should be much better now on high-dpi Nexus tablets.

dogmaphobic:
Size wise, it all looks fine now. The only main issue is that the button column on the left doesn't scroll, so you can't reach anything below *Power*. I would also make the button default size to about 1.5 X as high. The *Connect* button up on the top right corner for instance. It's too thin for it to be easily "pressed".

screen

PR #1868, opened 2015-09-11
Remove HDD, HUD, HSI

DonLakeFlyer:
These use OpenGL implementation which prevented moving to ANGLE
support. Since they are rarely used they are being remove instead of
being upgrade.


DonLakeFlyer:
@LorenzMeier Can you check this to make sure you are ok with it?

PR #1869, opened 2015-09-12
Add --test-high-dpi command line option

DonLakeFlyer:
This allows for testing issues found on high dpi screen without needing a high dpi screen. it does this by doubling the font sizes:

screen shot 2015-09-12 at 3 51 55 pm


DonLakeFlyer:
Once we clean up all the problems found by this. We'll have a full resolution independant UI which will works on all tablets. But it will also allow us to support picking you own font size for the app.

PR #1878, opened 2015-09-19
QGC Widget for tuning desired parameters/gains

fizzaly:
Sliders allow quick parameter tuning in flight. Changing parameter value in Setup tab currently doesn't show in widget.

### DIFF
--- a/src/FlightMap/Widgets/QGCParameterSliderWidget.qml
+++ b/src/FlightMap/Widgets/QGCParameterSliderWidget.qml
@@ -1,0 +1,241 @@
+/////////////////////////////////////////////////////////////////////////
+////// QGCParameterSliderWidget.qml -- Trieste Devlin, Sept 2015   //////
+////// QGroundControl widget -- sliding parameter tuner	           //////
+////// Note: Functional EXCEPT: slider view doesn't update         //////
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
This should go into the wiki in a section for user contributed custom command widgets not in the source tree. I'll get that set up when I get back home next week and put it there.

DonLakeFlyer:
Can you give me the exact sequence of steps you use to show when the slide does "NOT" get updated correctly? I'll try to fix it.

fizzaly:
Cool, thanks!

Open QGC, go to Plan tab with slider widget - parameter values load. Go to
Setup tab and change a parameter (by typing in a number and hitting enter,
no sliders involved). Then on returning to Plan, the slider doesn't update
to match the new change.

DonLakeFlyer:
I finally got around to looking at this. I can't load the file though because there is a binding loop. If you could get a version that will load I can figure out how to make it work. Also FYI: Now that we are on Qt 5.5.1 I believe there is a supported Qml Slider control you can use instead of making your own.

DonLakeFlyer:
When you have a new file you can just send it to me and I'll look at it. It shouldn't be a pull since this will go on the wiki not in the repo.

PR #1881, opened 2015-09-25
Joystick fix

DonLakeFlyer:
Fix for Issue #1850.

### DIFF
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -157,7 +157,7 @@ MacBuild {
     CONFIG += x86_64
     CONFIG -= x86
     QMAKE_MACOSX_DEPLOYMENT_TARGET = 10.6
-    QMAKE_MAC_SDK = macosx10.9
+    QMAKE_MAC_SDK = macosx10.11
 }
 
 LinuxBuild {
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
@dagar I had to bump to 10.11 sdk on OSX for latest XCode install

QGCCommon.pri
```diff
@@ -157,7 +157,7 @@ MacBuild {
     CONFIG += x86_64
     CONFIG -= x86
     QMAKE_MACOSX_DEPLOYMENT_TARGET = 10.6
-    QMAKE_MAC_SDK = macosx10.9
+    QMAKE_MAC_SDK = macosx10.11
```
LorenzMeier: I just removed this line and it did build fine. Can you test without it?

tumbili:
@DonLakeFlyer @LorenzMeier I tested this. The stick deflections are captured again but there are some new problems. (I'm using the Taranis over USB). When all stick deflections are captured QGC does not ask to move the switches anymore. Furthermore the "save calibration" button does not seem to have an effect as no RC parameter is set after.

DonLakeFlyer:
There is no move the switches step for Joystick calibration. Save calibration only saves the calibration settings. If you want to enable the joystick you need to check the enable checkbox which will in turn set parameters as well.

QGCCommon.pri
DonLakeFlyer: Doesn't work. If I remove the line it wants to use 10.10 and Xcode 7.0 only has 10.11.

PR #1883, opened 2015-09-25
Paramedter Editor search support

DonLakeFlyer:
- Added search support. Select "Search..." from the Tools menu
- Removed parameter editor as a widget. Getting too complex to support both big and small modes with new functionality.

screen shot 2015-09-25 at 10 28 39 am

screen shot 2015-09-25 at 10 21 14 am


LorenzMeier:
Nice! Could we please remove th checkboxes and just search everything? It's uncommon a part name shows up in a param description which is unrelated, and the selection is introducing failure modes for the search which are unnecessary.

PR #1884, opened 2015-09-25
Added APM Firmware Plugin

DonLakeFlyer:
- Added new new FirmwarePlugin::adjustMavlinkMessage support. This allows the plugin to take a whack at modifying the mavlink message on the way in and out. This is used to to adjust for protocol implementation differences aross firmware stacks.
- APM firmware plugin using the above to adjust for the way it handles parameter values
- Remove APM parameter hack code from base code since firmware plugin handles this now
- adjustMavlinkMessage will also be the way we handle the issues with APM stack mavlink severity changes: https://github.com/diydrones/apm_planner/issues/788

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -1,0 +1,167 @@
+/*=====================================================================
+ 
+ QGroundControl Open Source Ground Control Station
+ 
+ (c) 2009 - 2015 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ 
+ This file is part of the QGROUNDCONTROL project
+ 
+ QGROUNDCONTROL is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+ 
+ QGROUNDCONTROL is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+ 
+ You should have received a copy of the GNU General Public License
+ along with QGROUNDCONTROL. If not, see <http://www.gnu.org/licenses/>.
+ 
+ ======================================================================*/
+
+/// @file
+///     @author Don Gagne <don@thegagnes.com>
+
+#include "APMFirmwarePlugin.h"
+#include "Generic/GenericFirmwarePlugin.h"
+
+#include <QDebug>
+
+IMPLEMENT_QGC_SINGLETON(APMFirmwarePlugin, FirmwarePlugin)
+
+APMFirmwarePlugin::APMFirmwarePlugin(QObject* parent) :
+    FirmwarePlugin(parent)
+{
+    
+}
+
+bool APMFirmwarePlugin::isCapable(FirmwareCapabilities capabilities)
+{
+    Q_UNUSED(capabilities);
+    
+    // FIXME: No capabilitis yet supported
+    
+    return false;
+}
+
+QList<VehicleComponent*> APMFirmwarePlugin::componentsForVehicle(AutoPilotPlugin* vehicle)
+{
+    Q_UNUSED(vehicle);
+    
+    return QList<VehicleComponent*>();
+}
+
+QStringList APMFirmwarePlugin::flightModes(void)
+{
+    // FIXME: NYI
+    
+    qWarning() << "APMFirmwarePlugin::flightModes not supported";
+    
+    return QStringList();
+}
+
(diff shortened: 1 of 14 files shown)


src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
```diff
@@ -0,0 +1,167 @@
...
+                break;
+            case MAV_PARAM_TYPE_REAL32:
+                // Already in param_float
+                break;
+            default:
+                qCritical() << "Invalid/Unsupported data type used in parameter:" << paramValue.param_type;
+        }
+        
+        paramValue.param_value = paramUnion.param_float;
+
```
vooon: How it works? Shouldn't encoding step be here?

src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
```diff
@@ -0,0 +1,167 @@
...
+                paramUnion.param_int16 = (int16_t)paramValue.param_value;
+                break;
+            case MAV_PARAM_TYPE_UINT32:
+                paramUnion.param_uint32 = (uint32_t)paramValue.param_value;
+                break;
+            case MAV_PARAM_TYPE_INT32:
+                paramUnion.param_int32 = (int32_t)paramValue.param_value;
+                break;
+            case MAV_PARAM_TYPE_REAL32:
+                // Already in param_float
```
vooon: You should copy that value anyway, else some uninit value of paramUnion will be copied.

src/Vehicle/Vehicle.cc
```diff
@@ -263,6 +266,9 @@ void Vehicle::_sendMessage(mavlink_message_t message)
             static uint8_t messageKeys[256] = MAVLINK_MESSAGE_CRCS;
             mavlink_finalize_message_chan(&message, mavlink->getSystemId(), mavlink->getComponentId(), link->getMavlinkChannel(), message.len, messageKeys[message.msgid]);
             
+            // Give the plugin a chance to adjust
+            _firmwarePlugin->adjustMavlinkMessage(&message);
+
```
vooon: I think better to call adjust before finalize step.

src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
DonLakeFlyer: Correct, thanks.

src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
DonLakeFlyer: This is just modifying the message on the fly right when they come in from the communication link, or right before they got out the communication link. The encoding/decoding is done by the caller.

src/Vehicle/Vehicle.cc
vooon: But whole if body looks little strange.

1. get buffer
2. finalize
3. adjust
4. write

I think that it should be:

1. adjust
2. finalize (if really needed)
3. get buffer
4. write

src/Vehicle/Vehicle.cc
```diff
@@ -263,6 +266,9 @@ void Vehicle::_sendMessage(mavlink_message_t message)
             static uint8_t messageKeys[256] = MAVLINK_MESSAGE_CRCS;
             mavlink_finalize_message_chan(&message, mavlink->getSystemId(), mavlink->getComponentId(), link->getMavlinkChannel(), message.len, messageKeys[message.msgid]);
             
+            // Give the plugin a chance to adjust
+            _firmwarePlugin->adjustMavlinkMessage(&message);
+            
             if (link->isConnected()) {
                 link->writeBytes((const char*)buffer, len);
             } else {
```
vooon: Almost impossible - you already checked connection status.

src/Vehicle/Vehicle.cc
DonLakeFlyer: Correct, this is completely out of whack. Wonder why it worked against APM stack. I'll have to figure that out as well as fix.

src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
vooon: But you modify copies, not mavlink_message_t passed into function.

DonLakeFlyer: I see what you're saying. This is some pretty horrible code. I guess I should never write code in my free time while I'm on vacation again! Let me start over on it.

src/Vehicle/Vehicle.cc
vooon: In libmavconn i done that way:
https://github.com/mavlink/mavros/blob/master/libmavconn/src/interface.cpp#L79
https://github.com/mavlink/mavros/blob/master/libmavconn/include/mavconn/msgbuffer.h#L43

Normally finalize done by pack/encode step.

src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
vooon: :)

PR #1885, opened 2015-09-25
Fix APM adjustMavlinkMessage

DonLakeFlyer:
Previous code was horribly wrong.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -263,7 +263,7 @@ void Vehicle::_sendMessage(mavlink_message_t message)
             // Give the plugin a chance to adjust
             _firmwarePlugin->adjustMavlinkMessage(&message);
             
-            static uint8_t messageKeys[256] = MAVLINK_MESSAGE_CRCS;
+            static const uint8_t messageKeys[256] = MAVLINK_MESSAGE_CRCS;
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
@vooon Can you take a look now?

vooon:
Looks better, but `mavlink_msg_to_send_buffer()` should be after `mavlink_finalize_message_chan()`.
Else there no effect from finalize.

DonLakeFlyer:
Interesting, That's the way it was in the previous implementation as well. I'll change.

src/Vehicle/Vehicle.cc
```diff
@@ -260,20 +260,17 @@ void Vehicle::_sendMessage(mavlink_message_t message)
         if (link->isConnected()) {
             MAVLinkProtocol* mavlink = MAVLinkProtocol::instance();
             
-            // Write message into buffer, prepending start sign
-            uint8_t buffer[MAVLINK_MAX_PACKET_LEN];
-            int len = mavlink_msg_to_send_buffer(buffer, &message);
+            // Give the plugin a chance to adjust
+            _firmwarePlugin->adjustMavlinkMessage(&message);
+            
             static uint8_t messageKeys[256] = MAVLINK_MESSAGE_CRCS;
```
vooon: Also perhaps here should be `static const uint8_t`.

DonLakeFlyer: Sure

PR #1889, opened 2015-09-26
travis-ci android add gstreamer

dagar:


LorenzMeier:
Nice! Dailies with streaming support!

PR #1898, opened 2015-09-27
Icons for Map Display

dogmaphobic:
@DonLakeFlyer 

These are the icons for the Map Display. Once you have them integrated (so I can see them within their context), I can tweak them.

MapCenter is to select where to center the map (Home or Vehicle)
MapType is to select the type of map display (Satellite, road, terrain, etc.)
MapHomeTearDrop is to denote the home position.

For the vehicle position, I'm assuming we can use the existing airplane icon.

This is a start. I assume there will be quite a few new icons to be created. When you find yourself in need of a new icon for something, just make a purple rectangle (or anything) as a placeholder and I can later replace it.


DonLakeFlyer:
Drop down thingy almost working:
screen shot 2015-09-27 at 10 22 40 am

PR #1899, opened 2015-09-28
appveyor s3 deploy

dagar:
Hacks to figure out why s3 deploy doesn't work.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,23 +21,25 @@ test_script:
   - if "%CONFIG%" EQU "debug" ( debug\qgroundcontrol --unittest )
 
 after_build:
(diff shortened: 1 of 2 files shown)


dagar:
Appveyor s3 deploy is not working, but doesn't give any error.
It just says -

    Deploying using S3 provider

 https://ci.appveyor.com/project/dagar/qgroundcontrol/build/appveyor-223/job/o6an4vd1suigstv2

dagar:
@DonLakeFlyer this is finally working. I'll let it run through on master then get you the new continuous build url, then you can disable the last teamcity build.
FYI don't ever break builtin appveyor functionality because it usually fails silently.

PR #1901, opened 2015-09-28
Show home position on flight map

DonLakeFlyer:
@LorenzMeier I flashed master this morning. Booted HIL and armed. Didn't get MAVLINK_MSG_ID_HOME_POSITION across the wire. Double-checked mavlink inspector as well, not there either.

Can you try?


LorenzMeier:
Does the issue still persist?

DonLakeFlyer:
Just flashed master, ran QGC master, HIL+jmavsim, arm, message doesn't come across. Maybe a HIL only problem?

PR #1909, opened 2015-09-29
Add MissionManager unit test

DonLakeFlyer:
- fix bugs found by test
- Implemented Mock mission item protocol in MockLink

I would now deem the new MissionManager about 90% verified against normal operation. Need to add a few more boundary test cases and then also implement Mission Item protocol failure unit testing and we'll be done.

### DIFF
--- a/QGCApplication.pro
+++ b/QGCApplication.pro
@@ -515,6 +515,7 @@ HEADERS += \
     src/FactSystem/FactSystemTestGeneric.h \
     src/FactSystem/FactSystemTestPX4.h \
     src/MissionItemTest.h \
+    src/MissionManager/MissionManagerTest.h \
     src/qgcunittest/FileDialogTest.h \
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
I'm being extremely paranoid with this new Mission code since the consequences of getting it wrong seem pretty bad!

DonLakeFlyer:
@dagar This new unit test found the bug that was causing AppVeyor MavlinkLogTest to fail! Or at least one of them. I'll clean that up once this goes through.

LorenzMeier:
Pretty cool stuff! Unfortunately still failing Appveyor. Still focused on IROS, which is highly relevant and a lot of fun.

DonLakeFlyer:
It's a timing bug in the unit test. It's not waiting for the MissionManager to finish it's protocol of getting mission items. Which leads to a dangling ack timeout signal and then a crash. I have to fix that in a seperate pull.

PR #1937, opened 2015-10-10
MissionEditor usability changes

DonLakeFlyer:
Hopefully a picture is worth a thousand words. This is what you see when you have no mission items:
screen shot 2015-10-09 at 7 29 20 pm


DonLakeFlyer:
FYI: Some of the tool icons are a little funky. I'll get some better ones from @dogmaphobic when he gets back on town.

@LorenzMeier This should address some of your editor feedback from today. Let me know.

Left to do:
- Implement a dirty flag with some sort of visual indicator to let the user know they have not saved/sent their mission since the last set of changes.
- Waypoint height chart. Probably show distance between wayppoints here as well.

DonLakeFlyer:
@LorenzMeier I think with this, it just leaves Flight trajectory lines on the Fly map and then we are good to switch to the new Mission Editor and toss the old stuff. Let me know.

LorenzMeier:
Looks very nice! Certainly also workable. I'm still wondering if the sync tool should be enforced as last step of the mission item editing (e.g. if you make any change and a system is connected you either have to upload or reset your changes, and it will automatically download the onboard mission when you connect to the vehicle).

Right now you could edit the mission, forget to sync and then the system flies something entirely different and you won't notice / understand for a while.

DonLakeFlyer:
> it will automatically download the onboard mission when you connect to the vehicle

It already does that.

> sync tool should be enforced as last step of the mission item editing

That is this. There is no way to know automatically when the user is done editing a mission.

> Left to do:
> 
> Implement a dirty flag with some sort of visual indicator to let the user know they have not saved/sent their mission since the last set of changes.

LorenzMeier:
Do you have a bold "this is not uploaded yet" indicator somewhere so the user knows its not uploaded yet?

DonLakeFlyer:
Hence the terms "Left to do" and "visual indicator"!

> Left to do:
> 
> Implement a dirty flag with some sort of visual indicator to let the user know they have not saved/sent their mission since the last set of changes.

PR #1939, opened 2015-10-10
Enabled _connectLogArm_test

DonLakeFlyer:


DonLakeFlyer:
Still not working. Might be stuck in a loop somewhere.

PR #1940, opened 2015-10-10
Add support for MAV_TYPE specific FirmwarePlugin

DonLakeFlyer:
APM Flight Stack has different firmware depending on MAV_TYPE. This pull adds support for FirmwarePlugin's to be keyed off of both MAV_AUTOPILOT and MAV_TYPE.

### DIFF
--- a/src/FirmwarePlugin/FirmwarePluginManager.cc
+++ b/src/FirmwarePlugin/FirmwarePluginManager.cc
@@ -26,7 +26,7 @@
 
 #include "FirmwarePluginManager.h"
 #include "Generic/GenericFirmwarePlugin.h"
-#include "APM/APMFirmwarePlugin.h"
+#include "APM/ArduCopterFirmwarePlugin.h"
 #include "PX4/PX4FirmwarePlugin.h"
 
 IMPLEMENT_QGC_SINGLETON(FirmwarePluginManager, FirmwarePluginManager)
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
@pritamghanghas This is the pull you need to implement flight modes. It is only partially implemented. See line comment.

src/FirmwarePlugin/FirmwarePluginManager.cc
```diff
@@ -42,11 +42,46 @@ FirmwarePluginManager::~FirmwarePluginManager()
...
         case MAV_AUTOPILOT_ARDUPILOTMEGA:
-            return APMFirmwarePlugin::instance();
+            switch (vehicleType) {
+                case MAV_TYPE_QUADROTOR:
+                case MAV_TYPE_HEXAROTOR:
+                case MAV_TYPE_OCTOROTOR:
+                case MAV_TYPE_TRICOPTER:
+                    return ArduCopterFirmwarePlugin::instance();
+                    break;
+                // FIXME: The remainder of these need to be correctly assigned and new plugin classes created as needed.
```
DonLakeFlyer: This needs to be completed by assigning the correct set to ArduCopter (I may have missed some). And also creating the correct sets for the other firmware types.

pritamghanghas: I don't know whether I understood it correctly. Looks like you want me to
have a baseAPM plugin and special implementations for MAV_TYPES. Is that so?

On Sat, Oct 10, 2015 at 10:19 PM, Don Gagne <notifications@github.com>
wrote:

> In src/FirmwarePlugin/FirmwarePluginManager.cc
> <https://github.com/mavlink/qgroundcontrol/pull/1940#discussion_r41698946>
> :
>
> >  {
> >      switch (autopilotType) {
> >          case MAV_AUTOPILOT_ARDUPILOTMEGA:
> > -            return APMFirmwarePlugin::instance();
> > +            switch (vehicleType) {
> > +                case MAV_TYPE_QUADROTOR:
> > +                case MAV_TYPE_HEXAROTOR:
> > +                case MAV_TYPE_OCTOROTOR:
> > +                case MAV_TYPE_TRICOPTER:
> > +                    return ArduCopterFirmwarePlugin::instance();
> > +                    break;
> > +                // FIXME: The remainder of these need to be correctly assigned and new plugin classes created as needed.
>
> This needs to be completed by assigning the correct set to ArduCopter (I
> may have missed some). And also creating the correct sets for the other
> firmware types.
>
> —
> <https://github.com/mavlink/qgroundcontrol/pull/1940/files#r41698946>.
>



-- 
Regards,
pritam

DonLakeFlyer: APMFirmwarePlugin is now the base class for all APM vehicle type (MAV_TYPE) specific firmware plugins. It is no longer a standalone firmware plugin. You need to implement ArduPlaneFirmwarePlugin, RoverFirmwarePlugin, ... plugins following the model of ArduCopterFirmwarePlugin. Then you are going to need to modify this switch to create the appropriate plugin according to MAV_TYPE. I've already hooked up ArduCopterFirmwarePlugin but I'm not sure I have all the MAV_TYPE's that should be associated with it. For example does MAV_TYPE_COAXIAL and MAV_TYPE_HELICOPTER go with it as well?

pritamghanghas:
One small offtopic query. What is the state of joystick. Will it be
directly usable with APM or that also needs work.

DonLakeFlyer:
Depends on if they support the various SET_POSITION and SET_ATTITUDE mavlink message. My guess would be not, but it should be doable by them.

pritamghanghas:
I don't think they support SET_ATTITUDE. I think that is something on the
todo list and may or may not get included in 3.4.

On Sun, Oct 11, 2015 at 12:35 AM, Don Gagne <notifications@github.com>
wrote:

> Depends on if they support the various SET_POSITION and SET_ATTITUDE
> mavlink message. My guess would be not, but it should be doable by them.
>
> —
> <https://github.com/mavlink/qgroundcontrol/pull/1940#issuecomment-147117487>
> .
>



-- 
Regards,
pritam

PR #1944, opened 2015-10-10
Add dirty bit support to mission editing

DonLakeFlyer:
- Another nice internal feature is dirty bit support in QmlObjectListModel which is now capable of managing a list of QObjects which have dirty bits and expose the combination of all of those to the caller. All you need is for the QObjects which have a "dirty" property and a "dirtyChanged(bool)" signal.
- Also made the sync item fit a littel better

screen shot 2015-10-10 at 1 27 39 pm


LorenzMeier:
Awesome!

PR #1948, opened 2015-10-11
Cut over 100% to new Map/Mission code

DonLakeFlyer:
- Delete all old map/mission code


DonLakeFlyer:
@LorenzMeier All ready to go, if you're ok with current feature set

LorenzMeier:
There are build errors, but with these resolved, sure!

DonLakeFlyer:
@dogmaphobic Huge thanks for all your work on the Qml Map that got us to this point.

LorenzMeier:
Wohoo! Thanks everybody!

dagar:
As a small bonus this also makes the appveyor build faster.

PR #1951, opened 2015-10-11
Log fixes, MavlinkLogTest::_connectLogArm_test

DonLakeFlyer:
Was failing due to incorrectly named MockLinkLogVerbose category which would get past the *Log.debug=false rule. Then it would blow out the log files size limit on AppVeyor.

I hope :)


DonLakeFlyer:
Yes!

PR #1960, opened 2015-10-12
quick fix for yaw axis calibration issue

tumbili:


src/Joystick/Joystick.cc
```diff
@@ -278,8 +278,12 @@ void Joystick::run(void)
             float   pitch = _adjustRange(_rgAxisValues[axis], _rgCalibration[axis]);
             
                     axis = _rgFunctionAxis[yawFunction];
+
             float   yaw = _adjustRange(_rgAxisValues[axis], _rgCalibration[axis]);
-            
+                    _rgCalibration[axis].min = -32768;
```
LorenzMeier: Could you please only do this for axis 0, assuming its the only one which gets destroyed?

LorenzMeier:
@DonLakeFlyer We will need a hotfix since the user base leveraging the joystick in QGC is out of business right now. We've sunk already days into debugging this. I will next try to get you going with Quad HIL so you have a better view on the operation.

DonLakeFlyer:
Wow, really! No reason to calibrate then. If you want a temp workaround just modify the saved .ini file settings directly after you calibrate once. Need to find the real bug. 

Can you change this mod to only hack values in when the values are wrong?

Quad HIL is working fine for me. As is joystick cal and flying with it.

LorenzMeier:
Our main user base is using Linux to develop, and they can't fly any SITL missions right now. So messing with inis is not cutting it. We can probably try to be smarter about overriding the config, yes. But I was hoping we can nail it down rather than coming up with a fancier workaround 8).

PR #1961, opened 2015-10-12
Get mavlink_message_info_t from FirmwarePlugin

DonLakeFlyer:
This allows MavlinkInspector to show APM stack messages

### DIFF
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -7,53 +7,8 @@ WindowsBuild {
 
 #
 # [REQUIRED] Add support for the MAVLink communications protocol.
-# Some logic is involved here in selecting the proper dialect for
-# the selected autopilot system.
 #
-# If the user config file exists, it will be included. If this file
-# specifies the MAVLINK_CONF variable with a MAVLink dialect, support 
-# for it will be compiled in to QGC. It will also create a 
-# QGC_USE_{AUTOPILOT_NAME}_MESSAGES macro for use within the actual code.
-#
-MAVLINKPATH_REL = libs/mavlink/include/mavlink/v1.0
-MAVLINKPATH = $$BASEDIR/$$MAVLINKPATH_REL
-DEFINES += MAVLINK_NO_DATA
-
-# First we select the dialect, checking for valid user selection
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
@pritamghanghas Mavlink Inspector fix for firmware stack specific messages. Can you test with APM. It should all be working, but I'm not sure how to get APM to spit out one of those APM specific messages. Also note that the APM firmware plugin now uses the APM stack specific mavlink headers.

LorenzMeier:
That won't be enough. You also need to use the specific decode routine, since the message spec is used to seed the message payload (so two messages with the same ID can't be accidentally decoded wrongly).

DonLakeFlyer:
Not sure what you mean. Mavlink inspector doesn't use any decode routines. It does this: https://github.com/mavlink/qgroundcontrol/blob/master/src/ui/QGCMAVLinkInspector.cc#L547

LorenzMeier:
It means the message decoder will trash the message because it seeds the CRC with 0 instead of the proper seed. You will not receive any APM messages as a consequence.

DonLakeFlyer:
> message decoder will trash the message because it seeds the CRC with 0 instead of the proper seed

Where in the code?

DonLakeFlyer:
Ok. So this won't work yet until we get a newer version of mavlink that can support multiple interfaces. Once that shows up this should start working. I don't believe this pull specifically needs any additional changes.

pritamghanghas:
Yes, I tested it doesn't work yet.

On Mon, Oct 12, 2015 at 11:05 PM, Don Gagne <notifications@github.com>
wrote:

> Ok. So this won't work yet until we get a newer version of mavlink that
> can support multiple interfaces. Once that shows up this should start
> working. I don't believe this pull specifically needs any additional
> changes.
>
> —
> <https://github.com/mavlink/qgroundcontrol/pull/1961#issuecomment-147470885>
> .
>



-- 
Regards,
pritam

DonLakeFlyer:
@LorenzMeier Any reason to not merge this. It won't work now. But it should start to work when the new mavlink stuff isin.

PR #1965, opened 2015-10-12
Enable quick cached load of parameters when connecting to vehicle

NaterGator:
This branch enables a locally stored copy of the last vehicle's parameters to be requesting from newly connected vehicles. We request the "magic" parameter _HASH_CHECK which triggers the autopilot to compute the CRC32 of all used parameter names and values.

This hash value is computed on the local copy of parameters and checked against the vehicle's hash value to verify the locally cached params can be loaded directly skipping the lengthy request of all params over the radio link. 

I have tested this on [PX4](https://github.com/PX4/Firmware/pull/2982) but have not yet delved into the APM side of things.


LorenzMeier:
Awesome! Have you seen that the CI system failed this? A quick fix would be appreciated!

DonLakeFlyer:
I should have time to look at this tonight.

NaterGator:
I believe the issue was the change from param_request_list to param_request_read in the ParameterLoader ctor. MockLink was asserting because _HASH_CHECK wasn't defined in MockLink.params (and I don't believe it should be included there since it isn't really a param). 

I added some logic to return a value of 0 from MockLink when param_request_read comes in with _HASH_CHECK. I believe it will fix both builds and perhaps we can develop a test-case for loading from cache (valid _HASH_CHECK) and a failure continuing to the existing param_request_list (what we currently do in the test suite).

src/FactSystem/ParameterLoader.cc
```diff
@@ -552,6 +583,106 @@ void ParameterLoader::_writeParameterRaw(int componentId, const QString& paramNa
...
+    }
+
+    QFile cache_file(QFileInfo(QSettings().fileName()).path() + QDir::separator() + "param_cache");
+    cache_file.open(QIODevice::WriteOnly | QIODevice::Truncate);
+
+    QDataStream ds(&cache_file);
+    ds << cache_map;
+}
+
+static MAV_PARAM_TYPE factType2MavParamType(int factType)
```
DonLakeFlyer: Doesn't this already exist in _factTypeToMavType?

src/FactSystem/ParameterLoader.cc
```diff
@@ -552,6 +583,106 @@ void ParameterLoader::_writeParameterRaw(int componentId, const QString& paramNa
...
+    return paramType;
+}
+
+void ParameterLoader::_tryCacheHashLoad(int uasId, QVariant hash_value)
+{
+    uint32_t crc32_value = 0;
+    /* The datastructure of the cache table */
+    QMap<int, MapID2NamedParam> cache_map;
+
+    QFile cache_file(QFileInfo(QSettings().fileName()).path() + QDir::separator() + "param_cache");
```
DonLakeFlyer: I think this will work on all OS. But might be safer to use QDir to build the path.

NaterGator: I went with this based on what I found [here](https://github.com/mavlink/qgroundcontrol/blob/2ba8a0f3323bd91826873d58d4a6f610f6ae49e6/src/AutoPilotPlugins/PX4/PX4AirframeLoader.cc#L68) but I am open to modifying this.

src/FactSystem/ParameterLoader.cc
NaterGator: Yes it absolutely does and I missed it. Sorry, will remove.

src/FactSystem/ParameterLoader.cc
DonLakeFlyer: The directory is correct, but building full paths by hand tends to lead to strange problems on multi-platform. You should use the same pattern of using a QDir, then getting the full path by using QDir::filePath which is safer:
```
        QDir parameterDir = QFileInfo(settings.fileName()).dir();
        airframeFilename = parameterDir.filePath("PX4AirframeFactMetaData.xml");
```

src/FactSystem/ParameterLoader.cc
DonLakeFlyer: No problem, there's a hell of a lot code in QGC to get your head around.

NaterGator:
Something to consider is storing a parameter cache per some combination of system and component ID (or perhaps just system ID?). Users that frequently switch between multiple vehicles would benefit, otherwise they'll incur the param sync cost every time they swap out vehicles.

src/FactSystem/ParameterLoader.cc
NaterGator: Good feedback, I'll patch shortly.

LorenzMeier:
We have system and components (the CAN ESCs expose their params via MAVLink). So keying by system and component makes a lot of sense.

LorenzMeier:
Also please make sure to try against PX4 stable firmware so that the non-existing parameter is not breaking stuff.

NaterGator:
@LorenzMeier I ran against the PX4 stable firmware and did not have issues. [This](https://github.com/NaterGator/qgroundcontrol/blob/9862e074b93ce2e794227fb65c1c360c3d6168e6/src/FactSystem/ParameterLoader.cc#L508) timer runs for 2.5 seconds before [triggering](https://github.com/NaterGator/qgroundcontrol/blob/9862e074b93ce2e794227fb65c1c360c3d6168e6/src/FactSystem/ParameterLoader.cc#L72) a normal param_request_list based sync of all parameters. 

I'll have a look at the CAN ESC stuff to make sure I'm barking up the right tree with respect to grouping caches logically.

LorenzMeier:
Could you shorten that? Loading a single param should fit within 500-800 ms or so and since we have this capability now it would be great to make it instantaneous and not just faster - or how long does the onboard param computation take?

NaterGator:
@LorenzMeier I could shorten it; really the 2.5s time is incredibly generous. As mentioned in the PX4 pull request hash generation is in the order of low 10s of milliseconds (405 parameters was averaging 16ms) so the remaining time is to forgive any losses or retransmits over the data link. It is certainly worth weighing retrying the hash check message & timeout interval before falling back to the request everything approach... I'm not sure where that balance lies though. 

For now I'll drop the initial timeout to 750ms; QGC unlocks fully the instant _HASH_CHECK comes in and matches the cache's hash. In practice if the radios are just falling behind and the request_list message has already been sent before the _HASH_CHECK value is returned the remainder of the params *are* still sent over the data link, but they do not impede use of QGC other than consuming some bandwidth in the data channel (which is transparent to the user).

DonLakeFlyer:
I'd probably just use the same constant as the timeout time for PARAM_REQUEST_LIST.

LorenzMeier:
No, because a high timeout here will affect all existing systems (they will only load the list after timeout).

DonLakeFlyer:
Yup, good point.

DonLakeFlyer:
Change the protocol? Like this:
- Ground: PARAM_REQUEST_LIST
- Vehicle: return new PARAM_CACHE_CRC message
- Vehicle: return normal PARAM_VALUE response as well
- Ground (understands new PARAM_CACHE_CRC): sees PARAM_CACHE_CRC, if match, stops index based individual requests
- Ground (knows nothing about new PARAM_CACHE_CRC): continues on as normal using existing protocol

Would work with any old/new firmware old/new ground station combination. Would have no delays at boot, for crc match.

DonLakeFlyer:
Didn't state it quite right. New ground station connect to old firmware would have the exact same timing as before.

LorenzMeier:
The param emission is stream based, the GCS does not request them. So this wouldn't work. It also seems quite complex for the value add of avoiding a 750 ms timeout on a 15 seconds + transfer.

NaterGator:
On PX4 we could [theoretically](https://github.com/PX4/Firmware/blob/a11ea8744606d03f5c8a7b881bc6b43ee8c1027d/src/modules/mavlink/mavlink_messages.cpp#L312) re-purpose custom_mode in the [heartbeat](http://mavlink.org/messages/common#HEARTBEAT) message.

DonLakeFlyer:
> The param emission is stream based
Crap, that's right I forgot.

LorenzMeier:
@DonLakeFlyer I'm happy with this, are you? I've merged the onboard part:
https://github.com/PX4/Firmware/pull/2982

PR #1971, opened 2015-10-13
Made map buttons toggle between black and orange.

dogmaphobic:
Adjusted icons so they no longer contain a background.
Added *Exclusive* functionality to the *RoundButton* element.
Added OnClicked signal for the drop down buttons so I could turn off *Add Mission Item* and *Home Mode*.

screen shot 2015-10-12 at 8 58 49 pm


DonLakeFlyer:
I messed round with the AppVeyor queue a little. For some reason it runs once for each commit, which is odd. Instead of just the last one.

DonLakeFlyer:
Crap, my editor merge conflicted with yours. Likely the change from _homePositionCoordinate to homePositionCoordinate.

dogmaphobic:
> Crap, my editor merge conflicted with yours. Likely the change from _homePositionCoordinate to homePositionCoordinate.

That was bound to happen given that I was messing around in your code at the same time you were :) It's all merged. Waiting for CI...

Issue #1963, opened 2015-10-12
Pinch Zoom not working on android map


DonLakeFlyer:

dogmaphobic:
Found an existing (reported) bug in Qt related to this. I will try the suggested work around:

https://bugreports.qt.io/browse/QTBUG-47170

dogmaphobic:
Fixed a long time ago.

PR #1980, opened 2015-10-13
Work around for pinch zoom bug in Qt (Qt Location)

dogmaphobic:
This is the workaround found in the *Places* example code in Qt (as explained in #1963). It should be fixed for 5.5.1 but in the mean time, this works.

### DIFF
--- a/src/FlightMap/FlightMap.qml
+++ b/src/FlightMap/FlightMap.qml
@@ -289,4 +289,11 @@ Map {
         }
     }
 */
+
+    MouseArea {
+        //-- TODO: Check if this is still needed when we switch to 5.5.1
+        //-- Workaround for QTBUG-46388 (Pinch zoom doesn't work without it on mobile)
+        anchors.fill: parent
+    }
+
 } // Map


dogmaphobic:
Weird… Some new PR was merged while this PR was running CI and it became **"branch is out-of-date"**. I updated it, which caused CI to relaunch a full build. While that was going, yet another new PR got merged and it again became **"branch is out-of-date"**. There is no option to merge. I can only update it, which will cause a whole new build to be launched again, which is no guarantee a new PR will be merged in and the cycle will repeat indefinitely. The same seems to be happening to #1982. Was this ever like this?

DonLakeFlyer:
@dagar You'll have to talk with Daniel about that.

LorenzMeier:
Gus, this is branch protection, which means you can only merge a branch at a time. It makes a lot of sense though to avoid breakage.

dogmaphobic:
> Gus, this is branch protection, which means you can only merge a branch at a time. It makes a lot of sense though to avoid breakage.

Yep, I get the idea behind it. I just don't remember ever running into this before, unless it was a PR that had been opened for days. Specially when the change is one block on a single file that has no relationship with anything the other PR changed.

Submitting a new merge and let's hope it doesn't become outdated again while running CI.

dogmaphobic:
Yep! #1965 just got merged and this is yet again *out-of-date*. This is now completely synchronous, why wasn't the case before. At this rate, these PRs will never get *up-to-date*.

dagar:
Would it still be a problem if only travis-ci was required?

DonLakeFlyer:
We need Windows verification though.

DonLakeFlyer:
TC didn't re-run pulls when/if master changed out from under them. I started it out that way, but it was a disaster so I switched it to only build once. Or when a new commit was added.

dagar:
So should we disable this feature until we find a way to speed up appveyor by an order of magnitude?

DonLakeFlyer:
I think Lorenz just did

dogmaphobic:
> I think Lorenz just did

Yup!

screen shot 2015-10-13 at 6 37 49 pm

PR #1982, opened 2015-10-13
Remove unused CameraView widget using legacy OpenGL

NaterGator:
This unused widget is causing the build to fail on newer versions of Qt with [dynamically resolved](http://doc.qt.io/qt-5/windows-requirements.html#dynamically-loading-graphics-drivers) OpenGL implementations:
> The dynamic loading has a significant impact on applications that contain native OpenGL calls: they may fail to link since qmake no longer automatically adds opengl32.lib or libglesv2.lib. Instead, applications are expected to use the OpenGL functions via the QOpenGLFunctions class. Thus the direct dependency on the OpenGL library is removed and all calls will be routed during runtime to the implementation chosen by Qt.

There are a few other native OpenGL calls that aren't built by default but could cause problems down the line (ex: VideoStreaming surfaces)


DonLakeFlyer:
This is needed for PX4 Flow camera focusing. It got incorrectly disconnected somewhere along the way. I've had it on my work list to add it back and convert it away from OpenGL.

DonLakeFlyer:
@dogmaphobic Can you look at the comment above with respect to openGL calls? We'll need to convert video streaming. When we move the Qt 5.5 we'll want to use the dynamic thing on Windows since OpenGL drivers there are such a mess.

DonLakeFlyer:
Anything else left to do here @NaterGator? I'll test it when I finally get around to putting it back.

NaterGator:
This now builds for me in Windows with Qt 5.5 so as far as I can tell the
default build is unbroken for new versions. I can't comment if the widget
still works but I am pretty confident it should :)

PR #1983, opened 2015-10-13
Precompile a handful of Qt headers to speed up compilation time

NaterGator:
In my experience this is going to dramatically improve compile times. I tested on windows and saw ~6x speedup using MSVC 2013. Curious to see if the CI builds all complete and what their timing looks like.


DonLakeFlyer:
Ooh, hope it works!

DonLakeFlyer:
4-5 minutes faster on Windows. But something strange going on with OS X.

NaterGator:
A bit disappointing but we can certainly work on improving the included headers for better performance and fixing other OSes. I don't have a Mac to compile on but I'll work on fixing that build.

DonLakeFlyer:
It's a good win on Windows. I'd consider just turning it on for Windows builds.

NaterGator:
@DonLakeFlyer for some reason AppVeyor was glitching out today and didn't pick up the new commit to trigger a build, but things should be good to merge if you are okay with the default behavior being precompiled headers only on windows.

PR #1984, opened 2015-10-13
Support live home position in Mission Editor

DonLakeFlyer:
If you are not connected to a Vehicle you get the Offline Home Position Manager which allows you to set up fake home positions. If you are connected to a Vehicle and it has home position lock then it will use that for home position.


DonLakeFlyer:
This still isn't quite right. But I have a better idea to fix it.

PR #1990, opened 2015-10-14
Changed Home Position Manager to Flying Field Manager

DonLakeFlyer:
This was done to reduce confusion around the fact that you don't really have a home position when you are editing offline. You are only simulating.

- naming changes
- Flying Field positions show up with "F" instead of "H". This is quick hack to differtiate between a live home position and a simulated flying field on.
- Fixed a bunch of button checked/right panel visible state problems. I think they all do the right thing now at the right time.

Just to reiterate what happen once you are connected to vehicle:
- the "F" indicator will go away and change to a real "H" home position indicator
- The Flying Field Manager will no longer be visible, instead you get static display of home position lat/lon/alt

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -117,6 +117,7 @@
         <file alias="QGroundControl/Controls/DropButton.qml">src/QmlControls/DropButton.qml</file>
         <file alias="QGroundControl/Controls/RoundButton.qml">src/QmlControls/RoundButton.qml</file>
         <file alias="QGroundControl/Controls/QGCCanvas.qml">src/QmlControls/QGCCanvas.qml</file>
+        <file alias="QGroundControl/Controls/ExclusiveGroupItem.qml">src/QmlControls/ExclusiveGroupItem.qml</file>
 
         <!-- Vehicle Setup -->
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
@LorenzMeier Take a look at this. Try editing missions connected and disconnected, as well as starting disconnected, editing and then transition to connected to see what iy does. Then send me the feedback for anything to make it better.

LorenzMeier:
Just building on tablet to test this.

DonLakeFlyer:
Looking for feedback on the concept. The editing of Flying Areas sucks, I know that.

LorenzMeier:
Well, it generally worked. I just found these two:
https://github.com/mavlink/qgroundcontrol/issues/1991
https://github.com/mavlink/qgroundcontrol/issues/1992

DonLakeFlyer:
Did it sort of make sense?

LorenzMeier:
Yes, it sort of made sense. I think its easier if we polish the waypoints first. I look more at the workflow and not so much at details, and so getting it 80% right overall first will help me to give better recommendations.

DonLakeFlyer:
Agree. Flying Field is a bit power user anyway. last on the list

PR #2013, opened 2015-10-16
Adding Qt Location headers for Qt 5.5.1

dogmaphobic:
This only affects the build when using Qt 5.5.1. Other versions are unchanged.
So far only tested with Mac OS.


DonLakeFlyer:
There seems to be a new problem with the Windows unit test. But that isn't related to this pull. So I'm merging. I'll figure out what is going on with Windows UT later on.

PR #2014, opened 2015-10-17
Fixing map cache for Qt 5.5.1

dogmaphobic:
Map cacheing for < 5.5.0 was broken and we were using whatever default was provided by QtLocation. Now we're properly creating and configuring a cache manager.

This also solves the weird message output by QtLocation:

```
Plugin uses uninitialized directory for QGeoTileCache which will was deleted during startup
```


DonLakeFlyer:
Guess I'm going to have to figure out the windows unit test break

PR #2016, opened 2015-10-17
travis-ci linux and android use Qt 5.5.1 hosted on s3

dagar:


DonLakeFlyer:
Let's see what sort of hell that breaks loose!

dagar:
2 unit test failures

**Linux**

    FAIL!  : MainWindowTest::_connectWindowCloseGeneric_test() Received a fatal error.
https://travis-ci.org/mavlink/qgroundcontrol/jobs/85862364

**OS X**

    QFATAL : FactSystemTestPX4::parameter_specific_component_id_test() ASSERT failure in QGCSingleton: "Request for singleton that is NULL. If you hit this, then you have likely run into a startup or shutdown sequence bug (possibly intermittent).", file src/uas/UASMessageHandler.cc, line 55
    FAIL!  : FactSystemTestPX4::parameter_specific_component_id_test() Received a fatal error.

https://travis-ci.org/mavlink/qgroundcontrol/jobs/85862366

LorenzMeier:
@DonLakeFlyer By design a singleton should be instantiated on the first access, not in a centralised location, hence the error message makes little sense for that. Did you intent to do a singleton pattern (which would be buggy in this case) or did you in fact not mean to do a real singleton?

DonLakeFlyer:
I understand what a true singleton pattern is. In order to work around a myriad of shutdown sequence bugs which I spent a ton of time trying to figure out I finally gave up and a true singleton pattern. The shutdown problems are caused by singletons referring to each other. Then when you shutdown you get crashes if singletons are not taken down in the right order. I tried for days to make this happen automatically and correctly. But with Qt's concept of things like deleteLater and signals passing between objects and threads I couldn't get it to work. It should be possible, but I just haven't figured out exactly how to do it yet. Likely requires converting all access to shared pointers which is a bunch of work.

So instead of beating my head against the wall for any longer for the sake of purity. I locked down the singleton create and delete sequence ordering which in turn removed almost all of the issues.

DonLakeFlyer:
I haven't made it here yet to look at this. I will.

dagar:
Linux, OS X, Android are successfully building on Qt 5.5.1.
Appveyor is still on 5.5.0

DonLakeFlyer:
Can you rebase and try again

dagar:
Windows is passing. Linux and OSX are now failing the same way.

    FAIL!  : MainWindowTest::_connectWindowClosePX4_test() Received a fatal error.

DonLakeFlyer:
I'm going to have the turn that unit test off. Qt just can't handle it and I don't think the issues are real.

DonLakeFlyer:
Ahhh! Windows is such a pain in the ass. It always seems to run unit tests differently than all other OS. Causing failures which I"m pretty sure just have to do with the way we run unit tests. I think I'm going to cauterize windows UTs back to a workable set and depend on the other OS for full UT testing.

DonLakeFlyer:
You know what, I'm going to give up on Windows unit tests all together. I spent a ton of time fighting with them in TC as well. And in reality it was all not worth the effort since the bugs were just in the mechanism we use for unit tests, all of which works fine on the other os. I've wasted enough time on fixing no real user visible problems. If we find that not to be the case, we can turn them back on at some point and figure it out for real.

Can you turn off UT for windows, and rebase and we should have clean runs.

PR #2018, opened 2015-10-17
Mission Editor usability changes

DonLakeFlyer:
Too many to remember:
screenshot_2015-10-16-20-10-06

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -70,6 +70,7 @@
         <file alias="Help.svg">src/FlightMap/Images/Help.svg</file>
         <file alias="MapCenter.svg">src/FlightMap/Images/MapCenter.svg</file>
         <file alias="MapSync.svg">src/FlightMap/Images/MapSync.svg</file>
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
@dagar Could you show me how to debug the fact that the mac builds are just sitting there with no builds started on them? Can't seem to find any info on travis.

DonLakeFlyer:
Maybe just a backlog? https://www.traviscistatus.com

dagar:
Yes it's a backlog. I was seeing the same thing with PX4/Firmware builds.

DonLakeFlyer:
If I find some sort of problem is there a place to report issues?

dagar:
> If I find some sort of problem is there a place to report issues?

twitter :wink: 

In this case I'm fairly certain they already know. The best thing we can do is reduce load on their infrastructure. Cancel builds we don't need or that have already failed on linux, revisit precompiled headers, etc.

DonLakeFlyer:
Yeah I now. Just wanted to know for future. Twitter!

dagar:
Partially joking. There's also support@travis-ci.com.

PR #2029, opened 2015-10-18
Remove splash screen. Fixes #2028

### LINKED ISSUE #2028: Do away with the splash screen
Really, we shouldn't be writing software that needs a splash screen. If we want to show off our logo we should do so in the idle screen of the app. And we need a logo refresh anyway. @dogmaphobic Are you into logo design as well?

LorenzMeier:
@dogmaphobic @DonLakeFlyer Who knows if the splash screen messes with tablets. For sure easier without.

More code removed! Isn't that beautiful?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
     secure: RGovyUnMw3fp/bHZi058JvANT1rYmNqrsuSYew0cIgirO6YbMHr/rsjwCm1FTYpBl8s1zgr+u2b8ftYnfnCz2YT+Aip4NWrVYpVU0FEmfytGILrnUS0pjlt8m7fU9AKR1ElOSll7yw7e1kftynN39Q321etvwbLZcXon6zz0suE=
 
 matrix:
+    fast_finish: true
     include:
         - os: linux
           env: SPEC=linux-g++-64 CONFIG=debug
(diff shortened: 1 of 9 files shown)


dagar:
:+1:

DonLakeFlyer:
Nothing I like more than tossing code.

.travis.yml
```diff
@@ -9,6 +9,7 @@ env:
     secure: RGovyUnMw3fp/bHZi058JvANT1rYmNqrsuSYew0cIgirO6YbMHr/rsjwCm1FTYpBl8s1zgr+u2b8ftYnfnCz2YT+Aip4NWrVYpVU0FEmfytGILrnUS0pjlt8m7fU9AKR1ElOSll7yw7e1kftynN39Q321etvwbLZcXon6zz0suE=
 
 matrix:
+    fast_finish: true
```
DonLakeFlyer: This isn't a good change. You want to be able to see problems for each OS or debug/release combination. They are not always the same. They tend to be different more often than not. If you stop on first break. Then you won't catch the other different problems.

LorenzMeier: The documentation says that it will fail the build early, not abort the other VMs. It just means it turns the attention of the developer earlier to the failure, which hopefully reaches them before they have switched the context.

LorenzMeier:
@DonLakeFlyer Merging, feel free to revert the Travis change if you are not happy with my explanation.

.travis.yml
DonLakeFlyer: Sounds good

PR #2044, opened 2015-10-19
Remove font query from QWidget side of code

DonLakeFlyer:
For QWidget we will just use the Qt default font sizes. This removes the extra QQuickWidget creation which breaks boot Qml Profiling


DonLakeFlyer:
@dogmaphobic Can you look at this? The only other place besides style sheets that was using font sizes was PFD minimum size which I just set to a fixed size. Not necessarily correct on say a high dpi display. But nothing will go wrong. You'll just have to resize it and you'll be fine. I tested the other widgets and they all look fine.

PR #2046, opened 2015-10-19
Tablet boot performance work

DonLakeFlyer:
Mostly accomplished by delay loading components which are not needed immediately until after the initial display. In reality you can't visually tell that there is a delay.

Cold boot times:
Before changes: 6.5 seconds
After changes: 3.0 seconds

Some notes on performant Qml. Delay load as much as you can using a Loader that has it's source property set once you hit Component.onCompleted. Must be Loader.source, not Loader.sourceComponent in order to fully delay all processing of that component.

The main tool still has some pixel spacing issue which weren't there before. I'll fix that up in a separate pull.


DonLakeFlyer:
I have some changes to improve perceived performance as well. But I'll put those through after this gets in.

Issue #2045, opened 2015-10-19
QGC crashes while switching to plan screen with APM firmware on vehicle


pritamghanghas:
As soon as I click on plan button, it crashes with following messages

MissionManagerLog: writeMissionItems count: 0
MissionManagerLog: _handleMissionRequest sequenceNumber: 0
ASSERT failure in QList<T>::operator[]: "index out of range", file ../../Qt5.5.0/5.5/gcc_64/include/QtCore/qlist.h, line 518
The program has unexpectedly finished.

I am using Qt5.5.0 on Ubuntu 15.04 with AP 3.4-dev

DonLakeFlyer:
I can't repro this. Do you think you could get me a stack trace?

PR #2052, opened 2015-10-20
Share item indicator display code between Plan/Fly views

DonLakeFlyer:
Plan/Fly views now show mission item indicators exactly the same way. Working my way towards lines in Fly. And also figuring my way through restructuring to consider Plan/Fly merge at some point,


DonLakeFlyer:
I believe I have a fix for Issue #2045 in there as well.

PR #2058, opened 2015-10-21
Better light palette for mobile

DonLakeFlyer:
Plus lots of other tweaks to ui for visuals.

screenshot_2015-10-20-19-28-16

screenshot_2015-10-20-19-33-01

screenshot_2015-10-20-19-28-09


DonLakeFlyer:
Work in progress. But about 75% of the way there.

DonLakeFlyer:
Mobile now uses Light palette by default

dogmaphobic:
We need *dark* icons to go with the lighter background. I will look into it.

DonLakeFlyer:
Is it possible to color the svg's using QGCColoredImage?

dogmaphobic:
Nope. I just tried it and it doesn't work. In this case (fixed, two-state case), it's a lot faster for me to generate separate icons than having code to change it. It's 20 minutes of my time compared to a zillion, gazzilion computer cycles to change the color for each icon, each time it's used, for every instance, for every user. Analog to how I feel about Java/JavaScript in general. Makes the life of one person easier (the lazy programmer) while bogging down a billion users...

DonLakeFlyer:
:)

PR #2061, opened 2015-10-21
Joystick fixes

tumbili:


tumbili:
@LorenzMeier @DonLakeFlyer This makes usage of Joystick work for me. The only problem left is that QGC tells me that it couldn't find the parameter "COM_RC_IN_MODE" although it is clearly in the firmware param list.

src/Joystick/Joystick.h
```diff
@@ -152,7 +152,7 @@ class Joystick : public QThread
     
     CalibrationMode_t   _calibrationMode;
     
-    static const int    _cAxes = 4;
+    static const int    _cAxes = 6;
```
LorenzMeier: This is not the real fix. The axes need to come from the axis count detection of the joystick.

src/VehicleSetup/JoystickConfigController.h
```diff
@@ -220,7 +220,7 @@ private slots:
     static const int _attitudeControls = 5;
     
     int _axisCount;                     ///< Number of actual joystick axes available
-    static const int _axisMax = 4;      ///< Maximum number of supported joystick axes
+    static const int _axisMax = 6;      ///< Maximum number of supported joystick axes
```
LorenzMeier: Same issue here: If the user has yaw on axis 8 he still will run into the same issue.

LorenzMeier:
@DonLakeFlyer I think this still needs a proper architectural fix. For testing I propose you swap two axis on the SDL_ return value in in the Joystick::run() method (so that e.g. yaw becomes axis 5).

LorenzMeier:
We should also ensure that the top-level UI shows all axes the joystick has, as its confusing if you move a stick but no indicator is moving.

LorenzMeier:
@DonLakeFlyer Here is how you can repro on Mac OS:
https://github.com/mavlink/qgroundcontrol/pull/2063

Note that with this the fix from @tumbili Won't work, as he replaced one wrongly hardcoded value with another wrongly hardcoded value which is just containing his axis count. All code dealing with raw inputs needs to start using the dynamic axis count value from the SDL library, including the UI which has to show all present axes.

We will replace the MAVLink message with something that has roll / pitch / yaw / throttle and gimbal roll / pitch / yaw:
http://mavlink.org/messages/common#MANUAL_CONTROL

So please do not hardcode the assumption that there are only four axis into everything, or if you do please call them out roll / pitch / yaw / throttle so its obvious. We will soon need to extend those with more axes.

To wrap up: The number of mapped axes is not the same as the number of raw input axes, and the code does not properly tell them apart yet.

tumbili:
@LorenzMeier I guess this goes into the direction, right?

LorenzMeier:
Well, yes and no. Your patch seems to confuse the _cAxes (which probably means calibrated axes?) with _axisCount, which is the raw axis count. These two need to be very carefully used throughout the whole system. You should *not* need to increase the _cAxes value to make your Joystick work. Instead, you should use _axisCount in the correct locations.

DonLakeFlyer:
How about I make the change to use dynamic axis count from the joystick manager? I can test it on my Taranis.

PR #2063, opened 2015-10-21
Emulate a Linux SDL environment on Mac OS

LorenzMeier:
@DonLakeFlyer This is how the world looks like on Linux. It will break the existing joystick code. If you can calibrate and fly with this (note that the joystick needs to be enabled still in the code properly, right now only calibration window is enabling it, so you have to "visit" the joystick window before you can use it), then Linux will work as well.

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -57,7 +57,7 @@ Joystick::Joystick(const QString& name, int axisCount, int buttonCount, int sdlI
     : _sdlIndex(sdlIndex)
     , _exitThread(false)
     , _name(name)
-    , _axisCount(axisCount)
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
No idea what this is supposed to be doing? Are you saying that JoystickManager reports an incorrect axis count from SDK on Linux?

LorenzMeier:
I'm saying that the first N axis on Linux are *NOT* the controls that you're expecting. Your code only works if the first 4 axes are roll, pitch, yaw and throttle, and this change breaks that assumption so we can fix it for all platforms and joysticks which do not by chance follow this assumption.

DonLakeFlyer:
These are not the droids your looking for...

PR #2064, opened 2015-10-21
Fixing OpenGL issues for Windows (Video Streaming)

dogmaphobic:
I finally tried a Windows build using Qt 5.5.1 and found it to be broken if Video Streaming was enabled. This follows @NaterGator OpenGL fixes in #1982.


dogmaphobic:
Crap. This breaks the Android build. Looking into it...

PR #2070, opened 2015-10-21
Text to Speech Work

dogmaphobic:
A bit of work with Text To Speech

* Re-enabled speech for debug builds (only avoids unittest)
* Replaced deprecated API for Mac OS
* Added TTS support for Android
* Fixed some phrases here and there

### DIFF
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -11,8 +11,8 @@ WindowsBuild {
 # the selected autopilot system.
 #
 # If the user config file exists, it will be included. If this file
-# specifies the MAVLINK_CONF variable with a MAVLink dialect, support 
-# for it will be compiled in to QGC. It will also create a 
+# specifies the MAVLINK_CONF variable with a MAVLink dialect, support
+# for it will be compiled in to QGC. It will also create a
 # QGC_USE_{AUTOPILOT_NAME}_MESSAGES macro for use within the actual code.
 #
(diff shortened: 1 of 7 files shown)


QGCExternalLibs.pri
```diff
@@ -79,20 +79,20 @@ INCLUDEPATH += libs/qwt
 # Uninstalling from Linux can be done with `sudo make uninstall`.
 #
 XBEE_DEPENDENT_HEADERS += \
-	src/comm/XbeeLinkInterface.h \
-	src/comm/XbeeLink.h \
-	src/comm/HexSpinBox.h \
-	src/ui/XbeeConfigurationWindow.h \
-	src/comm/CallConv.h
+    src/comm/XbeeLinkInterface.h \
```
DonLakeFlyer: Have you tabs gone wonky? Not sure if that is a tab stop or your tabs are set to 8 instead of 4?

dogmaphobic: > Have you tabs gone wonky? Not sure if that is a tab stop or your tabs are set to 8 instead of 4?

Hmm... It's the other way around. I deleted my Qt Creator settings the other day and it reverted to *Fix Whole Document*. When I edited that file and saved it, it *Fixed* all the original tabs by replacing it with spaces (or the proper number of spaces).

screen shot 2015-10-22 at 3 40 32 pm

Issue #2074, opened 2015-10-22
Flight Mode - Channel Selection


ChristophTobler:
When selecting a new Stabilized/Main channel, the automatic channel selection for Altitude Control and Mission do not adapt (see picture).
![screenshot from 2015-10-21 15 34 41](https://cloud.githubusercontent.com/assets/14769636/10659024/b674bc08-789c-11e5-9484-68cba617d8c6.png)

DonLakeFlyer:
Thanks for the bug report

DonLakeFlyer:
Fixed

PR #2083, opened 2015-10-22
Fix channel index reordering notification

DonLakeFlyer:
Fix for Issue #2074


DonLakeFlyer:
Merging without windows run

PR #2087, opened 2015-10-22
Fix z order

dogmaphobic:
Moved Z order constants to QGroundControlQmlGlobal.

### DIFF
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -48,10 +48,6 @@ Item {
     // Prevents z order drawing problems.
     property bool hideWidgets: false
 
-    readonly property alias zOrderTopMost:   flightMap.zOrderTopMost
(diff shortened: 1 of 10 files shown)


dogmaphobic:
@DonLakeFlyer Right off the top of your head, any clue what could be causing this?
```
QFATAL : FactSystemTestPX4::parameter_specific_component_id_test() ASSERT failure in QGCSingleton: "Request for singleton that is NULL. If you hit this, then you have likely run into a startup or shutdown sequence bug (possibly intermittent).", file src/uas/UASMessageHandler.cc, line 55
```

The line in question has this:

```
IMPLEMENT_QGC_SINGLETON(UASMessageHandler, UASMessageHandler)
```

DonLakeFlyer:
Thathas been an intermittent problem over the last couple weeks. I just found and fixed it tonight. Probably in a pull tomorrow.

PR #2093, opened 2015-10-23
Some work for iOS

dogmaphobic:
The code generation for *git_version* was too much of a pain to deal with as each platform required specific tweaks for the compiler and flags. To generate a proper command for compiling for iOS took an absurd amount of complexity. In the end, I just ditched the whole thing and decided to use a simple define instead. No more code generation.

While at it, fixed some unused variable warning for Android and some OpenGL issues for iOS.


NaterGator:
@DonLakeFlyer this is going to break updating the git hash on rebuilds where either qmake was not rerun *or* the object file for QGCApplication.cc does not need to be updated. 

I don't know if the intent was to have a version string embedded in custom builds that were not built by a CI system, but if that was the intent this will undo the usefulness there.

It seems like the issue here is with manually invoking the compiler... if that is the case we can probably maintain the original spirit of the per-compile updated hash without resorting to one-off cases for each compiler. Let me know if that is the case and I'll work on it.

DonLakeFlyer:
@NaterGator @dogmaphobic @dagar I'll let you guys hash it out amoungt yourselves. I haven't apid much attention to this thing so I don't know much about it.

dogmaphobic:
I assumed that was only used for automated builds? Those start from nothing so everything is built from scratch. Why else would this be important?

dagar:
The version in the title bar is set from git. It's only a possible problem for people that compile qgroundcontrol themselves. If they're continuously pulling in new changes and building at some point they'll likely see the wrong version number. Maybe that's not a problem we need to worry about?

dagar:
I forgot about this until compiling just now, but passing the git sha as a define ruins ccache.

PR #2099, opened 2015-10-24
Add Rattitude mode for PX4

LorenzMeier:
@scott-eddy this adds rattitude in QGC for your PR:
https://github.com/PX4/Firmware/pull/3038

### DIFF
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -38,6 +38,7 @@ enum PX4_CUSTOM_MAIN_MODE {
     PX4_CUSTOM_MAIN_MODE_ACRO,
     PX4_CUSTOM_MAIN_MODE_OFFBOARD,
     PX4_CUSTOM_MAIN_MODE_STABILIZED,
+    PX4_CUSTOM_MAIN_MODE_RATTITUDE
 };
 
 enum PX4_CUSTOM_SUB_MODE_AUTO {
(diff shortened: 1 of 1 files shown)


scott-eddy:
Awesome!

DonLakeFlyer:
How does this affect flight mode switches?

LorenzMeier:
Its a new, common flight mode. I think I messed the addition up, I can select it in the dropdown but its not sticky.

The config part is missing. It should be added the same way as the others. @scott-eddy Has the details.

scott-eddy:
Don, this is mapped to a 2 pos switch and should be configurable from flight modes.  It's kind of a mix between attitude control and acro mode.  I'd  be happy to do dev work on the flight mode cal if you point me in the right directions.

DonLakeFlyer:
So this needs to be on a completely separate switch like Offboard? Does it take priority over all other modes except for Offboard?

DonLakeFlyer:
@scott-eddy Any answers to my questions. Want to finish this up.

LorenzMeier:
Hey Don. I just looked at the mode switching code:
- It depends on the main switch being in manual, like for ACRO
- If ACRO is on, it takes precedence. If its off or unconfigured, then RATTITUDE is evaluated
- If RATTITUDE is on, it switches to it, else it defaults to MANUAL

scott-eddy:
Sorry about that Don, this fell through the cracks on my end.

Lorenz answered succinctly.

DonLakeFlyer:
@dagar Any ideas on this one? Looks like a travis failure.

DonLakeFlyer:
> I think I messed the addition up, I can select it in the dropdown but its not sticky.

Selecting a mode from the dropdown doesn't control what flight mode is selected. It just sends that flight mode to the Vehicle and then if the Vehicle actually changes teh flight mode, the dropdown selection will update then. You could consider the dropdown just a "suggestion" to the vehicle :)

DonLakeFlyer:
@dagar this is fine to merge if you want to just bail on whatever is going wrong on the travis side with this.

kd0aij:
do we have a regression here? I thought Rattitude mode was in the flight modes setup page at one time, but I can't find it now. Also, SITL doesn't offer a rattitude mode.

DonLakeFlyer:
Not implemented

kd0aij:
OK, I must have dreamed that :8ball:

kd0aij:
But why is the status "merged"? Was this replaced by another PR? Seems odd to have a flight mode which QGC doesn't recognize.

DonLakeFlyer:
Hmm, forgot Lorenz added it

PR #2101, opened 2015-10-24
Fixed compass bindings

LorenzMeier:
Untested.


DonLakeFlyer:
I don' think this is the fix. I'll look

PR #2116, opened 2015-10-26
Geodetic math

dagoodma:
Wrote convertEnuToGeo() and convertGeoToEnu() to convert between geodetic and local coordinates. Code mostly taken from here: https://github.com/PX4/Firmware/blob/master/src/lib/geo/geo.c

Also wrote unit tests that tests conversion between geodetic and local coordinates, on or near the origin. Could probably use a failure test where the coordinate to convert deviates 100+ miles from the origin. This would prove the error from distortion due to the Earth's curvature.

All unit tests are passing.


DonLakeFlyer:
FYI: After you clean up build problems you may see unit test failures on Linux. Looks like master may be broken there.

dagoodma:
Passes and seems to work. I was just missing <limits>

LorenzMeier:
We're not using ENU. We're using NED across the board. Please rename and make sure the math is right.

src/QGCGeo.cc
```diff
@@ -0,0 +1,96 @@
...
+    double ref_sin_lat = sin(ref_lat_rad);
+    double ref_cos_lat = cos(ref_lat_rad);
+
+    double c = acos(ref_sin_lat * sin_lat + ref_cos_lat * cos_lat * cos_d_lon);
+    double k = (fabs(c) < epsilon) ? 1.0 : (c / sin(c));
+
+    *x = k * (ref_cos_lat * sin_lat - ref_sin_lat * cos_lat * cos_d_lon) * CONSTANTS_RADIUS_OF_EARTH;
+    *y = k * cos_lat * sin(lon_rad - ref_lon_rad) * CONSTANTS_RADIUS_OF_EARTH;
+
+    *z = coord.altitude() - origin.altitude();
```
LorenzMeier: This should be NED

dagoodma:
I changed the code to use NED. So just to clarify, if your home position is at 10 altitude and you convert a point at 100 altitude, the down component will be -90.

dagoodma:
Unittest fails for MissionManagerTest in SPEC=linux-g++-64 CONFIG=debug, but not on my account I don't think.

DonLakeFlyer:
Unit test failure was fixed last night on master. Not related to this pull.

PR #2121, opened 2015-10-26
Small fixes

tcanabrava:


src/ui/MainWindow.cc
```diff
@@ -651,7 +644,7 @@ void MainWindow::_loadVisibleWidgetsSettings(void)
     if (!widgets.isEmpty()) {
         QStringList nameList = widgets.split(",");
         
-        foreach (QString name, nameList) {
+        foreach (const QString &name, nameList) {
```
DonLakeFlyer: Ah. Interesting. Didn't realize you could save an allocation with these.

DonLakeFlyer:
Thanks

PR #2122, opened 2015-10-27
Calculate/Display distance between waypoints

DonLakeFlyer:


DonLakeFlyer:
I need to tweak this a little to correctly handle waypoints with mixed relative alt settings.

DonLakeFlyer:
Here is how it works:
- If home position is valid (from live vehicle) then all distances can be calculated
- If home position is not valid then only distances between relative altitude waypoints are shown

PR #2124, opened 2015-10-27
Flight view continue

dogmaphobic:
I made the *Instrument Panel* vertical to optimize screen real estate.
I've added some basic telemetry data to the panel. For now we have:

* Altitude (WGS84)
* Ground Speed
* Air Speed (Only shows if there is one to show)
* Vertical Speed (Climb Rate)

Once we figure out what else should go there I can add it. These are some possibilities:

* Altitude Above Ground
* Some kind of a timer, possibly started when armed
* Distance from Ground Station (Ground Station needs its own GPS coordinates for this)

screen shot 2015-10-27 at 1 47 46 pm

While at it, I've added a save/restore setting for *Map Is Main* for the fly view. I've also added a global settings set of functions to QGroundControlQmlGlobal to handle it.

### DIFF
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -56,8 +56,9 @@ Item {
 
     readonly property string _mapName:                  "FlightDisplayView"
     readonly property string _showMapBackgroundKey:     "/showMapBackground"
+    readonly property string _mainIsMapKey:             "MainFlyWindowIsMap"
 
-    property bool _mainIsMap:           !_controller.hasVideo
+    property bool _mainIsMap:           QGroundControl.loadBoolGlobalSetting(_mainIsMapKey, true)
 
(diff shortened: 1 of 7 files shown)


DonLakeFlyer:
Something to think about...

LorenzMeier:
Hey Gus,

Great! I think the altitude is the most relevant bit, so I would want those digits significantly larger.

-Lorenz

dogmaphobic:
> I think the altitude is the most relevant bit, so I would want those digits significantly larger.

There:
screen shot 2015-10-27 at 2 13 01 pm

LorenzMeier:
Looks *awesome*. But what I meant is even more aggressive: *in-action readable*. That means 000.0 digits spanning the full width (with readings going from 0.0 to 999.9) of the element and about the font size of GS calling out "Altitude (m)". You might consider doing something similar for GS and VS as well.

I don't think we should be displaying anything we consider "optional", so I would spend real screen estate for those we care about and try to keep them as few as possible.

PR #2126, opened 2015-10-27
Removing unused Main Flight Display widget.

dogmaphobic:
It used to be a *Docking* window that is no longer used or needed.
Also removed all its resources.
Bumped settings version in case you had it saved with it open.


dogmaphobic:
Unittest failed but I have no idea how I could have caused it:

`MissionManagerTest::_testReadFailureHandlingAPM()`

Was it really something I did?

DonLakeFlyer:
Rebase to master

dogmaphobic:
That didn't seem to help...

DonLakeFlyer:
Wait that's a different failure than I expected. That's probably an autoconnect break.

PR #2130, opened 2015-10-27
Adjustments to Telemetry Data

dogmaphobic:
* Made altitude quite larger
* Changed to columnar label/data
* Removed *Climb Rate*
* Made so if *Air Speed* is present, it replaces *Ground Speed*
* Made tool buttons on the Fly View to be anchored to the top left of the screen (it didn't fit on Android if vertically centered)

s


dogmaphobic:
Time elapsed based on when the vehicle was armed will be a bit more involved than I first thought. It will come next.

LorenzMeier:
Maybe leave this to the autopilot side. Handling the toolbar first is much more worthwhile.

dogmaphobic:
I'm assuming these UT failures are bogus as well?
```FAIL!  : MissionManagerTest::_testWriteFailureHandlingPX4() Compared values are not the same
   Actual   (_multiSpyMissionManager->checkSignalByMask(inProgressChangedSignalMask | newMissionItemsAvailableSignalMask)): 0
   Expected (true)                                                                                                        : 1
   Loc: [src/MissionManager/MissionManagerTest.cc(93)]```

dogmaphobic:
This has been running for around 4 hours now. Can I assume CI is broken?

DonLakeFlyer:
I just restarted all the jobs after I reverted the last autoconnect pull which broke master. But that won't work since CI doesn't merge against latest master.

dogmaphobic:
Well, it passed 552bdfba2bebacbf42795cbc6ddfe84b300ee261 and it's sitting on the last commit, which only removed a spurious qDebug(). I will wait a bit...

PR #2131, opened 2015-10-27
Doc fix for getting gstreamer in to android build

pritamghanghas:
Only Doc change. No need for any build or functionality testing.


dogmaphobic:
I now realized what I did. Instead of untarring from the command line, I just double-clicked the tar ball from the Finder, which created the directory named after the archive name.

The version is for now fixed to 1.5.2. I want to eventually look for *gstreamer-1.0-android** but for that to work, I need to add quite a bit more intelligence to figure out valid versions. For instance, 1.6.0, the current one, is incompatible with our code.

dogmaphobic:
And by the way, thanks for the graphical layout of the directory structure. That will make life a lot easier for folks trying this in the future.

PR #2133, opened 2015-10-28
Detect video stream.

dogmaphobic:
I'm now detecting when a video stream is present. If there is no video coming in (or it stopped), the screen goes black and a notice is presented telling you so.

screen shot 2015-10-28 at 12 44 52 am


dogmaphobic:
It looks like MissionManagerTest is having a bad day :)

DonLakeFlyer:
It always has a bad day!

DonLakeFlyer:
This is rebases to latest master? Those problems look like the one I already fixed.

dogmaphobic:
> This is rebases to latest master? Those problems look like the one I already fixed.

I'm pretty sure it was. I will try again...

PR #2136, opened 2015-10-28
Fix artificial horizon (it was stuck)

### LINKED ISSUE #2135: New instruments: Horizon / ground indicator is not moving
@dogmaphobic In the new instruments in the fly view the horizon (blue / green) indicator is not moving.

dogmaphobic:
This fixes #2135 

The trick to get a circular mask relies on the root entity to be hidden (*visible: false*). The mask element makes only the masked out parts visible. The artificial horizon code had a check to see if it was *visible* and it wouldn't do anything if *hidden*.

screen shot 2015-10-28 at 11 54 13 am

### DIFF
--- a/src/FlightMap/Widgets/QGCArtificialHorizon.qml
+++ b/src/FlightMap/Widgets/QGCArtificialHorizon.qml
@@ -72,7 +72,7 @@ Item {
             Rotation {
                 origin.x: artificialHorizon.width  / 2
                 origin.y: artificialHorizon.height / 2
-                angle: root.visible ? -rollAngle : 0
+                angle:    -rollAngle
             }]
     }
 }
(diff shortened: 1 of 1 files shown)


LorenzMeier:
Hmm, the white lines should be aligned with the artificial horizon. In your screenshot it still looks not right.

dogmaphobic:
Duh, handled translation and forgot rotation. Lack of coffee...

screen shot 2015-10-28 at 12 12 16 pm

dogmaphobic:
While looking at it I found a scaling issue that has been there since the beginning. Hopefully it's now all fixed.

dogmaphobic:
This was fully updated and the change was a single QML file. Yet, CI broken...

dagar:
Only the linux debug build failed during unit tests. Restarting to see if it's consistent.

PR #2139, opened 2015-10-28
Start logging on heartbeat instead of connect

DonLakeFlyer:
Putting this back. UT failure was an intermittent problem already in master.


DonLakeFlyer:
Waiting for all my Singleton, ParmaterLoader, UT fixes to go through. Then I'll re-run this against master to make sure it doesn't cause another wild and strange UT problems.

PR #2140, opened 2015-10-28
Debug window to help define device sizing issues.

dogmaphobic:
This is an attempt to determine some sort of constant to be used to compute dimensions for the various platforms and devices. This window is only reachable on *Debug Builds*.

The idea is to load it, take a screen grab and send it to me so I can gather the necessary data for coming up with a *size right everywhere* solution.

screen shot 2015-10-28 at 3 01 44 pm


DonLakeFlyer:
Killing builds for now to get my UT off pull through. Then I'll rerun. Otherwise it's just going to fail.

PR #2145, opened 2015-10-29
gather linux libs

dagar:
Gathering a list of all linux deps to throw into the package. If anyone knows a better way to do this, or the "right" way please let me know.

### DIFF
--- a/QGCSetup.pri
+++ b/QGCSetup.pri
@@ -98,3 +98,55 @@ WindowsBuild {
 
 }
 
+LinuxBuild {
+    QMAKE_POST_LINK += && mkdir -p $$DESTDIR/libs
+
+    # QT_INSTALL_LIBS
+    QT_LIB_LIST = \
+        libicudata.so.54 \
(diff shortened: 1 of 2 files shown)


LorenzMeier:
Also need libsdl-dev and libudev-dev.

dagar:
I'm not particularly thrilled with this solution, but it does allow running qgroundcontrol on a fresh ubuntu 14.04 without tracking down Qt. You have to start it with `qgroundcontrol-start.sh` to get all the paths set correctly.

DonLakeFlyer:
If those are the exports can use those same onces to point to an online installer Qt?

DonLakeFlyer:
The direction I'm heading is to just include a shell screen in the build you can use to run against the supported online installer version and install locations. Then users can also use this as an example if they want to do something strange like use a different version.

DonLakeFlyer:
Should this be merged so we can cut a new stable? README needs to be udpated as well.

PR #2156, opened 2015-10-29
Fix state of buttons (menu, toolbar)

tcanabrava:
When you selected via the Menu to go to Fly Mode, the QML
toolbar didn't followed the option, and if you selected
via the menu to go to all the three possible views (setup,
fly and plan) the menu would be all-selected.

This tries to keep consistency of things.


dogmaphobic:
Thanks for that but be aware I will be completely redoing the tool bar UI in the next few days.

DonLakeFlyer:
Hoping to have those bogus unit test failurs fixed by tonight.

PR #2157, opened 2015-10-29
Turn off fatal warnings yet again

DonLakeFlyer:
I've restricted unit tests and singletons in another pull that is coming which creates issues this fatal warnings. I'll fix that up separately after the other pull goes through.


DonLakeFlyer:
Slamming this through

PR #2161, opened 2015-10-30
Fix the corner case of a system running on localhost using the networ…

LorenzMeier:
…k and the loopback address in parallel. Force it onto the loopback interface.

@tumbili This fixes the issue you were seeing.


LorenzMeier:
@tumbili I pushed a likely fix for the remaining partner ID case to master. If that doesn't fix it we'll need to debug that one as well.

PR #2173, opened 2015-10-31
Remove Flightgear and JSBSim and only leave options which work

LorenzMeier:
@DonLakeFlyer This is not a complete cleanup, but I'd like to do that once we figure out HIL in general. Also need to think about simulation on tablets (so users can learn how to use QGC). Right now just making sure the options which are displayed actually work.


DonLakeFlyer:
> Also need to think about simulation on tablets

I'd really like to bolt a jmavsim like thing directly into QGC.

LorenzMeier:
Yes, that would be easy enough to integrate once you have a basic 3D view going. Working on a simulation backend here, but it'll take some time: https://github.com/LorenzMeier/SimulationKit

DonLakeFlyer:
If you can give me the simulation engine I can give you the visuals. I did work on Microsoft Flight Simulator for god sakes! So I must have learnt something!

PR #2175, opened 2015-11-01
Update path to executable for FlightGear on OSX

AndreasAntener:
I'm not sure if this changed between FlightGear versions. With the latest one, this is now the path to the executable.


LorenzMeier:
Yes, I think stuff changed.

AndreasAntener:
Hm, maybe it would need a detection of some sort instead of this PR..

AndreasAntener:
This checks for the new and old path now

PR #2178, opened 2015-11-02
Toolbar work

dogmaphobic:
Some work on the Main *Toolbar*.

The layout of the toolbar was quite broken overall and basically unusable on mobile devices. I’ve completely reworked it so it now uses proportional fonts (as opposed to hardcoded) as well as proportional dimensions in order to look the same across all platforms. While at it, I’ve reworked the graphics and layout.

The Toolbar is now exactly the same for both Desktop and Mobile platforms. Because Mobile platforms do not have the *Main Menu*, you won’t be able to access the Settings/Preferences. I’m in the process of moving those settings into the *Setup View* where they belong.

From left to right:

* Setup View

I’ve fixed the dimensions of the main setup view so it no longer *underlaps* the main tool bar. I also replaced the *Flickable* item from the button list with a proper *ScrollView* as that is what should be used there. With that said, the rest of the Setup View with its many panels are still far from usable on mobile platforms.

* Mission View

Allowed map to *underlap* the main tool bar (which is slightly translucent) but limited the waypoint list to the usable screen area below.

* Fly View

Same as above, the map/video *underlaps* the toolbar while the widgets within are constrained to the usable area below it.

* Message Icon

It's only displayed if there are messages to be displayed. The icons shows as a megaphone if only non critical messages are present. If a critical message is present, the icon shown is a pulsating warning icon. Once clicked, the (old) message dialog is displayed where you can see all the system messages. I have yet to rewrite that so it uses a proper, consistent QtQuick widget.

* Satellite Indicator

Shows how reliable satellite coordinate are along with the number of satellite currently visible.

* RC RSSI Indicator

Simple, signal strength icon showing the RSSI as measured by the PX4.

* Battery Indicator

Shows the percentage remaining along with the current (total) voltage. I want to show the voltage per cell but I could not readily find the data source for it.

* Vehicle Selector

Only shown if more than one vehicle is connected. This is still the original, unchanged button as it was there. Only properly aligned. I still need to work on it. Took lower priority because I have never run more than one vehicle at a time.

* Flight Mode Selector

This needs some work. It currently shows a menu with all *Flight Modes*. The issue is that some of those *modes* are not modes but rather, *actions*. They should be removed from the list and added as action buttons elsewhere in the UI (Loiter, Return to Land, and Mission).

The icon next to it is, for now, hardcoded to a quadcopter. The internal list no longer reflects reality and I will need to change it accordingly. At minimum, the icon shown should be representative of the type of vehicle the PX4 is programmed for.

* Arm / Disarm

Self explanatory

* Connect / Disconnect

In theory, this will go away once we have automatic connection functioning. The idea is that you setup (in the *Setup View*) what kinds of connections you have and QGC will automatically start these connections once it starts. There should be no reason to manually connect or disconnect from a vehicle. Once that's done, this button will go away (That's why I didn't spend a lot of time on the icons themselves).

Most of the work was done from the *Fly View* perspective. The main toolbar and the Fly View should be usable in all supported platforms. Even small screen devices (some phones) should be at least visible. The other two views (Setup and Mission) still need some work for them to be usable on mobile platforms.

screen shot 2015-11-02 at 9 01 11 am


LorenzMeier:
Re-running because of this breakage - if it doesn't pass we'll have to run brew update twice: https://github.com/Homebrew/homebrew/issues/45616

LorenzMeier:
Closing in favour of the updated PR.

PR #2185, opened 2015-11-04
Ui tweaks

dogmaphobic:
This is mostly related to the *Settings View*. The main *Settings View* and its various panels were not designed with mobile devices in mind. They use hardcoded dimensions, which don’t scale well. Even on desktop, some of the panels were too large and there was no way to scroll around. I’ve done quite a bit of tweaking in some panels so they can function on mobile devices.

I made them dynamic so they adapt to the screen proportionally. There are still some problem areas but a lot of those are related to the source of the data, rather than then UI itself. Inconsistent mixed case or some times, too long descriptions, improper icons, etc. (I also want to eventually replace all those Airframe icons with something more consistent with the rest of the UI.)

So far I've worked on these:

* Vehicle Summary
* Airframe
* Power
* Safety
* Parameters

No work done for the following panels. It will require full attention and quite a bit of time as they are all quite complex, especially the *Sensors* panel as Don is planning on adding 3D elements to it.

* Firmware
* Joystick
* Radio
* Flight Modes
* Sensors

Todo:

Add the existing QtWidget preferences as new panels to the *Settings View*


dogmaphobic:
@dagar Did I cause this or is it just having a bad hair day?
```
if "%CONFIG%" EQU "installer" ( appveyor PushArtifact C:\projects\qgroundcontrol\release\qgroundcontrol.exe )
Error pushing artifact: The request was aborted: The request was canceled.
Command exited with code 2
```

dagar:
This is just appveyor uploading the installer to s3 randomly failing with a completely unhelpful error message. I'll restart it.

dogmaphobic:
Alright, 3 in a row makes it rather suspicious that it's a fluke of appveyor. Out of ideas and I can't do much else while this is in limbo...

PR #2187, opened 2015-11-04
Fix checkbox aligment.

dogmaphobic:
The original code was *centering* the text within the parent. That caused the text to be away from the *checkbox* if the control was sized larger than its contents. I had removed that and now I noticed that on Android, the *checkbox* was out of alignment. Now I realized what this was trying to do.

This is the proper fix. Instead of centering, it's using vertical alignment.


dogmaphobic:
I guess appveyor just doesn't like me...

dagar:
I'll take a look at appveyor tonight.

PR #2188, opened 2015-11-05
Started preferences panel

dogmaphobic:
Started some basic preferences panel. I placed it as one of the buttons in the *Setup View*. I used the original QtWidgets preference panel as a guide. It's far from complete but all the settings there are functional. That is, they do get set and saved.

While at it, I tweaked the palette a bit. The *Warning* text was bright yellow for both themes, which made it almost impossible to read on a full white background (the window color for the *Light* theme). I also replaced a few hardcoded HTML color tags (all setting the text to bright yellow for warning), with the proper warning color set in the system palette.

### DIFF
--- a/QGCApplication.pro
+++ b/QGCApplication.pro
@@ -270,7 +270,8 @@ HEADERS += \
     src/ui/uas/QGCUnconnectedInfoWidget.h \
     src/ui/uas/UASMessageView.h \
     src/MissionItem.h \
-    src/AutoPilotPlugins/PX4/PX4AirframeLoader.h
+    src/AutoPilotPlugins/PX4/PX4AirframeLoader.h \
+    src/QGCSettings.h
 
 WindowsBuild {
     PRECOMPILED_HEADER += src/stable_headers.h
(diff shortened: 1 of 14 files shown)


LorenzMeier:
Could you add a screenshot? Do you feel this is ready to go for master?

LorenzMeier:
Found the screenshot. There is another dialog which doesn't belong in vehicle setup: The joystick settings. Similarly the debug dialog and the Firmware upgrade dialog don't belong there.

How about we use this opportunity to clearly split QGC "setup" things from vehicle settings?

I see two approaches: One is to entirely split them and turn the menu button for settings into one of these custom buttons where you can pick QGC or vehicle #1 preferences. From a user perspective that makes sense, but the extra click will make accessing the param editor in flight painful. Or we have the vehicle settings in the main UI and QGC settings within the same dialog, but spaced apart so that you get that you have two settings lists, one for the GCS and one for the vehicle. That would be right now only a layout change and probably a better first step.

dogmaphobic:
I guess I put the comment in the wrong place 8P

The debug dialog is temporary and it only shows in debug builds. I was trying to collect data on screen size and density for various devices. I might leave it there a little longer (still only for debug builds) until we're comfortable with the UI across all devices.

Yes, we need to group these settings a little better. I think it's just a matter of shuffling things around. One *Root* setting window that branches off into *General*, *Vehicle*, *Joystick*, etc. I will wait for @DonLakeFlyer  to be back so we can figure this out.

In the mean time, I will merge this in as is as it provides a service regardless.

PR #2191, opened 2015-11-06
Separate Settings Panel

dogmaphobic:
Some more work towards some kind of *Preferences* panel.

Instead of shoving everything into the *Setup* view, I created a new button that brings up a *Preferences* view. I moved the *General* and *MavLink* settings from the *Setup* view over there (along with the *Debug* dialog, which is only present for Debug builds.)

The UI is quite spartan (no decorations or fancy graphics) until we figure out the best way to deal with this. The idea is to keep *Setup* exclusive for Vehicle Parameters you want to access frequently. The *Preferences* panel is the place for settings and initial configurations, which once done, you should have little reason to keep going back.

screen shot 2015-11-06 at 12 57 09 am

While at it, I cleaned up some unused graphics and move a couple of files around into places where it makes more sense (such as moving the general settings into a new *preferences* directory, created just for such files).


LorenzMeier:
Great! I think it would make a lot of sense to try to move the joystick settings over there as well.

PR #2194, opened 2015-11-06
Yet more cleaning

dogmaphobic:


dogmaphobic:
I guess @dagar added some special code checking to see if these are my commits and then pukes on them if it finds out it's mine :) No worries, I've added some code to the video display that if it knows it's your setup, it will show Youtube cat videos instead...

dagar:
Finding a better appveyor solution. https://github.com/mavlink/qgroundcontrol/pull/2196
I like cat videos...

PR #2197, opened 2015-11-08
appveyor cleanup s3 deploy

dagar:
-qmake rename installer output


dagar:
The qgroundcontrol executable was being clobbered by the installer. 

@DonLakeFlyer this changes the windows continuous build url. I'll send you the new URL once the build goes through.

PR #2199, opened 2015-11-09
Adding TTS to iOS

dogmaphobic:
In addition:
* Fixing code exclusion issues where Mac OS and iOS were both included when they shouldn't.
* Hardcoding warning text color on Fly View as that view does not respect the current palette.

### DIFF
--- a/QGCApplication.pro
+++ b/QGCApplication.pro
@@ -87,9 +87,11 @@ MacBuild {
 }
 
 iOSBuild {
-    QMAKE_INFO_PLIST = $${BASEDIR}/ios/iOS-Info.plist
-    ICON = $${BASEDIR}/resources/icons/macx.icns
-    OTHER_FILES += $${BASEDIR}/iOS-Info.plist
+    QMAKE_INFO_PLIST    = $${BASEDIR}/ios/iOS-Info.plist
+    ICON                = $${BASEDIR}/resources/icons/macx.icns
+    OTHER_FILES        += $${BASEDIR}/iOS-Info.plist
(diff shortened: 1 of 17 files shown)


dogmaphobic:
Hmm...
Twice in a roll makes me doubt it as being a fluke but my local unittest (with QT_FATAL_WARNINGS=1) runs fine.

```
********* Finished testing of SetupViewTest *********
ALL TESTS PASSED
After app delete
Debugging has finished
```

dogmaphobic:
@DonLakeFlyer Any ideas on what to do here? I can't reproduce this locally.

```
QWARN  : FactSystemTestPX4::parameter_default_component_id_test() PX4ParameterMetaDataLog: Invalid decimals value, name: "ctl_start_rate"  type: 5  decimals: ""  error: invalid number
/Users/travis/build.sh: line 41: 11831 Abort trap: 6           ./debug/qgroundcontrol.app/Contents/MacOS/qgroundcontrol --unittest
The command "if [[ "${TRAVIS_OS_NAME}" = "osx" && "${CONFIG}" = "debug" ]]; then ./debug/qgroundcontrol.app/Contents/MacOS/qgroundcontrol --unittest; fi" exited with 134.
Done. Your build exited with 1.
```

DonLakeFlyer:
The parameter meta data is wrong. Likely that the version compiled into QGC has the problem. But when you run you're using the meta data from a previous flash. Clear you configure folder of the cached meta data and you'll likely see the same problem

LorenzMeier:
I stuffed up. PR with fixed file coming.

PR #2200, opened 2015-11-10
Correctly updated default param meta data file

LorenzMeier:
@DonLakeFlyer @dogmaphobic This should fix CI.


dogmaphobic:
> This should fix CI.

Or not 8P

LorenzMeier:
Ts.. fix already pushed!

LorenzMeier:
Fix of the fix!

DonLakeFlyer:
I tried fixing this so we could get CI working again. It seems pretty out of whack. mot_v_accel is set as INT_32 but it has default values and min/max which are floats. I tried building firmware to get new meta data from there, but master firmware doesn't build for me either! So I'm not sure what to do at this point. It seems like these changes were committed to master without a pull request thus bypassing all CI testing!

PR #2202, opened 2015-11-12
iOS Build Work (and other stuff)

dogmaphobic:
This is a bit of a large commit. Sorry for the multiple commits within but I had a merge from master in between and I could not rebase/colapse them.

The main purpose of this is to get a proper, App Store ready iOS build (not quite there yet but close).

We use a *subdirs* template to handle our own QtLocation plugin along with the main QGC application. This causes a problem when creating an Xcode project. Qmake fails to properly create an Xcode project for projects with *subdirs* templates. It creates an Xcode project just for the application but it misses the plugin dependency. To work around this, instead of creating a separate (static) QtLocation plugin and later linking it, I'm now just compiling and building it all as one project. ~~This is for iOS only as it requires Xcode to build~~.

~~Note that we currently generate a make file for Mac OS builds. If you ever want to create an Xcode project for whatever reason, this above will apply as well.~~

* Proper iOS build with icons, bundle ID, launch page, etc.
* Forcing the use of shadow builds (too much mess allowing in-place builds)
* Cleaning up some old, unused libraries for Mac OS (OSG and OpenThreads)
* Removing these old libraries from the installer as well
* Found and fixed an issue when building an *Install* build on Mac OS with video streaming enabled (I guess no one ever tried that :)
* Changed the code that prepares the GStreamer framework (Mac OS) so it uses the build directory as a work place rather than the source directory (the build should be able to run with the source on a read only file system).
* Changed the theme on mobile builds from *forced* to light to *optionally* light. It defaults to light scheme but you can change it through the settings. It was hardcoded to light only.
* Gone is the *subdirs* template. We're now back to one single *project*. The ```qgroundcontrol.pro``` is the one and only project once again.


dogmaphobic:
:)

@DonLakeFlyer @dagar 

It looks like all CI builds run *in-place*. Question: Are these builds run from scratch each time? That is, do they create a work directory, clone the repository and run a build? Can I be sure nothing is persistent?

dogmaphobic:
Nah... That still would not work. Looking at the CI build scripts...

dogmaphobic:
Working on it. Bad case of search & replace in the case of Travis and some more hardcoded paths in the Windows installer script...

dogmaphobic:
@dagar Is there a way to check the resulting installers before merging to master?

dagar:
Do this in another branch off of the the main repo and they'll all be uploaded to s3 just like master.

dogmaphobic:
> Do this in another branch off of the the main repo and they'll all be uploaded to s3 just like master.

Nah. Given the number of commits and branches already here, that's prone to be a disaster. I'm running out of time for the day any way...

dogmaphobic:
I ditched the *subdirs* template for all platforms. Tested in all platforms locally and it's all working fine. A heck lot simpler project to boot...

dogmaphobic:
> This branch has conflicts that must be resolved

Oh great... let me deal with that now.

PR #2206, opened 2015-11-13
Add MockLink Start/Stop to Preferences

DonLakeFlyer:
Preparation for auto-connect having no connect/disconnect buttons. Added debug only MockLink button to Qml Preferences to support mock link start/stop when there is no connect/disconnect button.

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -24,6 +24,7 @@
         <file alias="MavlinkSettings.qml">src/ui/preferences/MavlinkSettings.qml</file>
         <file alias="MissionEditor.qml">src/MissionEditor/MissionEditor.qml</file>
         <file alias="MissionEditorHelp.qml">src/MissionEditor/MissionEditorHelp.qml</file>
+        <file alias="MockLink.qml">src/ui/preferences/MockLink.qml</file>
         <file alias="PowerComponent.qml">src/AutoPilotPlugins/PX4/PowerComponent.qml</file>
         <file alias="PowerComponentSummary.qml">src/AutoPilotPlugins/PX4/PowerComponentSummary.qml</file>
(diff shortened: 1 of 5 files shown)


dogmaphobic:
Are we dropping support for TCP connections? Is there any other connection that is *initiated* by QGC? In those cases, we may need a way to activate the connection. It wouldn't just start on its own. What I had thought was that you would configure the connections (in settings) and if one of those types of connection is set, the *Connect* button would be there. In all other (normal) cases, there would be no connect button.

In summary, yes. For 99% of connections, it would be automated and there is no need for a connect buttons. But there could be cases where it could still be needed.

DonLakeFlyer:
As far as I know TCP was only used by APM SITL support. I added it years ago for that. But they support UDP as well. So not sure there is any other use for TCP.

DonLakeFlyer:
The one that is likely more of a  problem is xbee. Not sure how that will work yet. Also no idea how to test, since I don't have one of those.

dogmaphobic:
> The one that is likely more of a problem is xbee. Not sure how that will work yet. Also no idea how to test, since I don't have one of those.

I actually have quite a few but they are used for something completely different (home automation). My guess is that there will always be odd ball cases where you initiate the connection from QGC and for those cases, you would need a *Connect* button. There is no reason to delete the button. Just don't show it unless you have a connection type that needs it. Let's not worry about that for now. Once you are done, I can take a look at the configuration (the comm settings) and based on that, I will figure it out. The MockLink is a good example for when you want the Connect button there. These are non-ordinary connection types for advanced users.

DonLakeFlyer:
There is some strange code in XBeeConfigurationWindow.cpp. It looks like it has custom serial port enumeration code in there for each OS! Looks like it works through a serial port so maybe we can find it automatically like a radio?

PR #2214, opened 2015-11-14
Consistent buttons in Fly and Mission views.

dogmaphobic:
Make the tool buttons in both views consistent.


dogmaphobic:
Looks like a DNS error in appveyor or something else screwy?
```
Attempt to get headers for http://sourceforge.net/projects/nsis/files/NSIS%202/2.46/nsis-2.46-setup.exe/download failed.
Exception calling "GetResponse" with "0" argument(s):
"The remote name could not be resolved: 'sourceforge.net'"
```

PR #2218, opened 2015-11-15
Auto-Connect port stuff

DonLakeFlyer:
- Centralize board detection in QGCSerialPortInfo class. This also hides the android/non-android differences from consumers.
- Change code to use new QGCSerialPortInfo board detection logic.
- Add new requiresUSBMavlinkStart method to LinkInterface. This signals whether the mavlink start shell command is needed.
- Update MAVLinkProtocol to use requiresUSBMavlinkStart appropriately.

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -34,18 +34,13 @@ This file is part of the QGROUNDCONTROL project
 #include <QDebug>
 
 #ifndef __ios__
-#ifdef __android__
-#include "qserialportinfo.h"
-#else
-#include <QSerialPortInfo>
-#endif
+#include "QGCSerialPortInfo.h"
 #endif
 
 #include "LinkManager.h"
 #include "MainWindow.h"
 #include "QGCMessageBox.h"
 #include "QGCApplication.h"
-#include "SerialPortIds.h"
 #include "QGCApplication.h"
 
 QGC_LOGGING_CATEGORY(LinkManagerLog, "LinkManagerLog")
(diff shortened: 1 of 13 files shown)


DonLakeFlyer:
@dogmaphobic Can you check the QGCSerialPortInfo and LinkManager changes? This is in the middle of everything and needs more eyes. I tested by flashing and connecting to all the types of stuff I own.

DonLakeFlyer:
Another step into the maelstrom of Auto-Connect...

DonLakeFlyer:
Also FYI: The current state of this code sets preferred on all auto-detected board types. So they will all show up at the top. Usually not a problem since you normally only have on connection. If you happen to have more than one thing connected, they will all go to the top above your persisted connections. I think thats probably ok.

DonLakeFlyer:
I'm going to merge this so I can make forward progress. @dogmaphobic Can you still double check it after the fact?

src/comm/LinkManager.cc
```diff
@@ -504,8 +499,15 @@ void LinkManager::_updateConfigurationList(void)
 #endif
         // Save port name
         currentPorts << portInfo.systemLocation();
-        // Is this a PX4 and NOT in bootloader mode?
-        if (portInfo.vendorIdentifier() == SerialPortIds::px4VendorId && !portInfo.description().contains("BL")) {
+
+        QGCSerialPortInfo::BoardType_t boardType = portInfo.boardType();
+
+        if (boardType != QGCSerialPortInfo::BoardTypeUnknown) {
+            if (portInfo.isBootloader()) {
```
thomasgubler: I noticed this when verifying #2221 but I am not sure if it's intended: isBootloader() calls boardTypePixhawk() which calls boardType() hence the whole board type logic is done twice in quick succession (see line 503).

DonLakeFlyer: I could probably make boardType return boot loader status to remove the double call. Let me see.

DonLakeFlyer:
Thanks for verifying @thomasgubler

PR #2226, opened 2015-11-16
Full Auto-Connect [WIP do no merge]

DonLakeFlyer:
@LorenzMeier Putting this up so you can give it a try. I still have a lot of testing to do, and also trying to figure out how to write a unit test for it. Links now have both Connected/Disconnected and Active/Inactive states. A link is Connected/Inactive if it is physically connected by no vehicle has been found on the link yet. There is also now the concept of persistent links which are automatically Connected when available. UDP is the first example of a persistent link. I think I'm pretty close to being able to make all link type work with Auto-Connect. Once I wring out UDP, I'll add the rest. The Connect button on the toolbar is currently still there and can still be used for non-persisted links. On tablet this will eventually go away, but on desktop will remain.


LorenzMeier:
Pretty cool! Looking forward to play with this once Appveyor passes.

DonLakeFlyer:
No need to wait for AppVeyor that is just a problem with xbee.

DonLakeFlyer:
@LorenzMeier @dogmaphobic Should should now be in an Alpha sort of state with respect to Auto-Connect for all know board types. I'm sure there are still a ton of edge case issues, given the number I already hit and fixed. But it you just start from boot QGC, do a single connect and shutdown it works. tested with 3DR Radio and Flow. Auto-Connect + El Capitan + Direct UDB puts you in hell you can't get out of, so not much testing there yet.

DonLakeFlyer:
Ok guys this should be much closer now. Can you give it a try to see if it works for you. Connect/Disconnect button is still not the way I want it. But other than that it should be mostly working:
- Auto-connect is supported for all known types: UDP, USB direct Pixhawk, 3DR Radio, Flow
- Connect/Disconnect button is not shown on tablet
- You should be able to plug in or pull the usb and the connection will come/go correctly
- Pull the battery on your vehicle while over radio or udp and the vehicle should go away

Missing:
- Desktop Connect/Disconnect button semantics
- Config options to turn off/on auto-connect for various board types

DonLakeFlyer:
Also the Radio/Tememtry is hacked out from the toolbar for now. I'll fix that.

DonLakeFlyer:
Closing in prep for the real pull

PR #2228, opened 2015-11-16
Apm stuff

tcanabrava:
This is a work in progress dialog to set the correct airframe type on an APM device.
I based this code on the APM Planner airframe configuration but code is 98% mine.
Caveats:
 - slow, a few requests to ardupilot git repository to get the parameters file, probably I should cache them
 - when you click on Airframes config, screen will take around 10s to display because of that.
 - when you click on 'Apply and reestart', 'Yes', the parameters will appear on the debug output and not apply. I was not confident that the code to apply was correct so I commented it out.
 - I had to create a QAbstractListModel and remove the QVariantList for the AirframeTypes because they are created after the download finishes.
 - XML file for the APM drones are handmade and probably with the wrong configuration for the Type (just need a finetuning )

### DIFF
--- a/src/AutoPilotPlugins/APM/APMAirframeComponentController.h
+++ b/src/AutoPilotPlugins/APM/APMAirframeComponentController.h
@@ -54,7 +54,10 @@ public:
     Q_PROPERTY(QString currentVehicleName MEMBER _currentVehicleName CONSTANT)
     Q_PROPERTY(int currentVehicleIndex MEMBER _currentVehicleIndex CONSTANT)
     Q_PROPERTY(int autostartId MEMBER _autostartId NOTIFY autostartIdChanged)
+    Q_PROPERTY(QString fileParams MEMBER _currentFileParams)
+
     Q_INVOKABLE void changeAutostart(void);
+
     
     int currentAirframeIndex(void);
     void setCurrentAirframeIndex(int newIndex);
@@ -74,6 +77,7 @@ private:
     QString         _currentVehicleName;
     int             _currentVehicleIndex;
     int             _autostartId;
+    QString         _currentFileParams;
     bool            _showCustomConfigPanel;
     int             _waitParamWriteSignalCount;
     APMAirframeModel    *_airframeTypesModel;
@@ -128,7 +132,7 @@ public:
     QVariant data(const QModelIndex &index, int role) const;
     int rowCount(const QModelIndex &parent = QModelIndex()) const;
     void setAirframeTypes(const QList<APMAirframeType*>& airframeTypes);
-
+    APMAirframeType *getAirframeType(const QString& airframeTypeName) const;
 private:
      QList<APMAirframeType*> _airframeTypes;
 };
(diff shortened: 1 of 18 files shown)


src/AutoPilotPlugins/APM/APMAirframeComponentController.h
```diff
@@ -0,0 +1,140 @@
...
+    QString name() const;
+    QString imageResource() const;
+
+private:
+    QString         _name;
+    QString         _imageResource;
+    QVariantList    _airframes;
+};
+
+class APMAirframeModel : public QAbstractListModel {
```
DonLakeFlyer: You don't need to build one of these yourself. QmlObjectListModel.cc/h alraedy exists which supports creating a QmlObjectListModel from any QObject.

DonLakeFlyer:
I don't think the APM stack needs a reboot after you apply. Mission Planner doesn't do that does it?

DonLakeFlyer:
How does MissionPlanner or APMPlanner deal with the param file download thing? Does it download it every single time or does it do some sort of check. You could ask Bill on bitter how APM Planner handles it. In general this should be handled like param meta data. There is a version compiled into QGC. And then newer versions are downloaded on the fly and cached.

src/AutoPilotPlugins/APM/AirframeFactMetaData.xml
```diff
@@ -0,0 +1,70 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<airframes>
+  <version>1</version>
+  <airframe_group image="AirframeQuadRotorX.png" name="Quadrotor x">
+      <airframe id="10020" maintainer="Lorenz Meier &lt;lorenz@px4.io&gt;" name="3DR Aero M" file="3DR_AERO_M.param">
```
DonLakeFlyer: Is the maintainer field required? I'd just leave it out.

src/AutoPilotPlugins/APM/APMAirframeComponentController.h
tcanabrava: I discovered that when I had already finished this one, I'll update the code accordingly.

src/AutoPilotPlugins/APM/AirframeFactMetaData.xml
tcanabrava: okay.

tcanabrava:
@DonLakeFlyer it doesn't needs a reboot, I'm removing the check for the multiple devices. about the parameters file, it downloads it everytime.

LorenzMeier:
Can you please add a few screenshots of the thing in operation? We commonly not only review the code, but also the visual quality and usability and it would be great to understand if we're already there with this. Thanks!

tcanabrava:
The visual is the same as the px4 plugin, I'll add one on the new pr that
I'm working on
Em 18 de nov de 2015 06:40, "Lorenz Meier" <notifications@github.com>
escreveu:

DonLakeFlyer:
Let me know when you have the new changes. Then I'll build it myself and take a deeper look.

tcanabrava:
closed because of a new version

PR #2234, opened 2015-11-18
Auto connect support [merge after Stable 2.8]

DonLakeFlyer:
- Known board types are now connected to automatically
- You can turn autoconnect types on/off through settings
- There is no connect/disconnect button any longer. For manual connections you need to go through File/Manage Communication Links.
- Auto-Connections are detected as soon as you plug in a device. When you unplug a device they will go away.

From a testing standpoint this is a pretty massive change. It affects all the more nasty parts of QGC:
- Connect/Disconnect: Anything that relies on being smart about connections is affected. This includes:
  - Firmware upgrade: Since it only works when there are no active connections. Also suspends connections while running.
  - Airframe config: Only works with a single connect
  - Log Replay: Only start if no connections, suspends connection while replay
- Multi-vehicle: Shutting down QGC will multiple auto connect or non autoconnect links is like driving a bulldozer through a jewelry store. Fought with this for days. Likely still don't have all corners handled.

The main change to accomplish this was to add an additional Active/Inactive state to a link. So now you can have an Inactive/Connected link. So you must be careful about coding to a connected or an active link as approrpriate.


dogmaphobic:
I only did simple tests, using one single, direct USB connection. The only weirdness I found so far:

I plugged the Pixhawk directly (USB) and everything came up fine. I went to the Firmware window and noticed the screen tells us: *Click Disconnect in the toolbar above.*. I ignored it, disconnected the USB cable, reconnected and the Firmware upload process came up fine. I was able to upload the firmware just fine but QGC never reconnected. The Firmware window recycled and again displayed the notice above (*Click Disconnect in the toolbar above.*) even though it is not exactly *Connected*. From that point on, I could not connect it again. Unplugging it within the Firmware window restarts the firmware upgrade (expected but weird). If I switch over to the Fly view, it just never connects.

When I try to close QGC, the dialog asking if I want to disconnect the vehicle comes up (it is not *connected*).

Relaunching QGC makes everything go back to normal.

This will be a weird one. I think when the firmware has been successfully uploaded, it should show a message telling it was successful and move away from the Firmware window to something else.

dogmaphobic:
Quick test with a Chinese 3DR Radio clone never connected. I will debug it tomorrow.

DonLakeFlyer:
Firmware upgrade should lock out Autoconnect. As soon as you move it, it should connect again. Looks like it didn't reset maybe. Check the ids and text name for radio.

dogmaphobic:
This is the info on the radio:
```
LinkManagerVerboseLog: portName:           "cu.usbserial-AH02QFUO"
LinkManagerVerboseLog: systemLocation:     "/dev/cu.usbserial-AH02QFUO"
LinkManagerVerboseLog: description:        "FT232R USB UART"
LinkManagerVerboseLog: manufacturer:       "FTDI"
LinkManagerVerboseLog: serialNumber:       "AH02QFUO"
LinkManagerVerboseLog: vendorIdentifier:   1027
LinkManagerVerboseLog: productIdentifier:  24577
```

dogmaphobic:
In QGCSerialPortInfo.cc

This:
```
        } else if (description() == "FT231X USB UART") {
```
Could be changed to:
```
        } else if (description().endsWith("USB UART")) {
```

dogmaphobic:
Multiple vehicles is not working. I started QGC with one already connected through the USB and another through UDP. QGC went nuts with an endless string of (spoken) error messages. I unplugged the USB one but it kept emitting the messages for over a minute before it finally ended. At that point, it disconnected everything (the UI went back to *no active vehicle* state).

I never liked the way spoken messages are handled. The buffer cannot be unlimited. It should contain no more than a couple of messages. New ones should bump old ones out of the queue. In cases like above, it takes minutes for the buffer to clear.

```
LinkManagerLog: New auto-connect port added:  "Pixhawk on cu.usbmodem1" "/dev/cu.usbmodem1"
LinkManagerLog: New heartbeat on link:vehicleId:vehicleMavlinkVersion:vehicleFirmwareType:vehicleType  "/dev/cu.usbmodem1" 1 3 12 13
LinkManagerLog: New heartbeat on link:vehicleId:vehicleMavlinkVersion:vehicleFirmwareType:vehicleType  "Default UDP Link" 1 3 12 13
```

dogmaphobic:
Oh, forgot to mention. When multiple vehicles are connected, the flight widgets got nuts :) The compass and the attitude indicator keep updating for both vehicles, jumping from place to place.

dogmaphobic:
Wrong button...

DonLakeFlyer:
> I was able to upload the firmware just fine but QGC never reconnected.
I should probably allow it to reconnect right after upload is done since that is going to be what you want anyway.

> Multiple vehicles is not working.
Both your vehicles have the same id. So that end up being the same vehicle. Hence they will fight with each other for updatng the ui. This is working correctly. You can have one vehicle id over two links.

> New ones should bump old ones out of the queue.
Probably depending on priority. High sev, never get bumped, low sev do. Can you enter an issue?

DonLakeFlyer:
> Quick test with a Chinese 3DR Radio clone never connected. I will debug it tomorrow.

I"m was already a bit concerned with the radio string fallback since it will catch a lot of stuff. Changing it will make it even more generic. What do you think? Make it more broad and just live with it connecting to random things?

DonLakeFlyer:
@dogmaphobic Can you try your firmware flash case again. I changed it so that it will autoconnect right after a firmware upgrade.

Fojtik:
My device is not connected through USB but through UDP ethernet. I can report whatever I wish, but I really do not know what to report.

DonLakeFlyer:
UDP is listening on port 14550 for a heartbeat. If you are talking through that port then no changes are needed on your side. It will all work automatically.

Fojtik:
But it does not. It already worked, but now it is broken.

DonLakeFlyer:
You have to wait for this pull to get merged in!

Fojtik:
OK so, I will install a new daily build when it  will be ready.

dogmaphobic:
> @dogmaphobic Can you try your firmware flash case again. I changed it so that it will autoconnect right after a firmware upgrade.

That's working now. The message still needs to be fixed though:

screen shot 2015-11-18 at 9 26 18 pm

Funky scenario where it doesn't work: Unplug USB cable to initiate a Firmware upgrade and then *Cancel* the upgrade. It goes into a loop where there is no way out of it. You need to exit QGC and start over.

screen shot 2015-11-18 at 9 30 45 pm

DonLakeFlyer:
> It goes into a loop

Not sure what you mean. Are you expecting it to re-autoconnect after cancel?

DonLakeFlyer:
> The message still needs to be fixed though:

Since the way you disconnect depends on how you connected. I just removed the second sentence about the disconnect button.

DonLakeFlyer:
> I"m was already a bit concerned with the radio string fallback since it will catch a lot of stuff. Changing it will make it even more generic.

Guys, any comments on changing this to be more generic?

dogmaphobic:
> Guys, any comments on changing this to be more generic?

You know my opinion :) If you have an some other FTDI device connected, you ought to know what you're doing. If they have something and it's in use, you won't be able to open the port any way.

dogmaphobic:
> I just removed the second sentence about the disconnect button.

That's what I meant. The button not being there and all.

dogmaphobic:
> Not sure what you mean. Are you expecting it to re-autoconnect after cancel?

Well, if I cancel, QGC is stuck in that mode (asking me to unplug the USB cable). If I switch views, it never connects. If there is no way out of it, I would would shut down QGC once canceled as it is useless at that point.

DonLakeFlyer:
> Well, if I cancel, QGC is stuck in that mode (asking me to unplug the USB cable). If I switch views, it never connects.

Ok, got it. Suspend connections isn't getting reset on a cancelled flash. There is a fallback that it will get reset when the firmware controller destructs. When you say switching views, I assume you mean top level views. Not switching off of the firmware setup view to some other setup page. That should have turned auto connect back on.

dogmaphobic:
> When you say switching views, I assume you mean top level views.

Right! I went to the Fly View.

DonLakeFlyer:
Just outlining work left to do so I don't forget:
- Change radio fallback string
- Fix cancel/failed flash to turn off suspend auto connect

DonLakeFlyer:
Gus, can you give this a try again with the latest changes.

Remaining work:
- Add exclude list
- Allow user created connections to take part in auto-connect (maybe V2, we'll see how hard it is)

DonLakeFlyer:
Since the desktop does not open the radiop fallback broadly I don't think and exclude list is needed for V1. That can wait for later.

So as far as I'm concerned from my end this pull request is done. If you guys are happy with the testing you've done with it. Then I'll merge this after Stable 2.8 goes out the door.

dogmaphobic:
Good:

* After firmware upgrade, QGC reconnects and is all hunky dory.

Bad:

* Selecting the firmware dialog, it should tell me to unplug the USB cable if I want to upload the firmware. It just tells me all vehicles should be disconnected. If I were looking at this for the first time, I wouldn't know what to do.
* If I cancel the firmware upgrade, the dialog returns to its default state (waiting for me to unplug the USB cable). It doesn't *reconnect* and it doesn't allow me to do anything (other than unplugging the USB cable and trying again). If that's the case, it should tell the user to shutdown QGC and start over.

Other than that, it all works. There are some other problems but they are not exactly related to this PR so I will file new issues for those.

Fojtik:
I have not found latest nightly build.
The installer marked as stable now automatically reconnects - it is fine.

PR #2249, opened 2015-11-20
Linux shell script run command

DonLakeFlyer:
This shell script is used to run a installed version of QGC which goes against a Qt installed using the online installer.

### DIFF
--- a/deploy/qgroundcontrol-start.sh
+++ b/deploy/qgroundcontrol-start.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+export LD_LIBRARY_PATH=$HOME/Qt/5.5/gcc_64/lib
+export QML_IMPORT_PATH=$HOME/Qt/5.5/gcc_64/qml/
+export QML2_IMPORT_PATH=$HOME/Qt/5.5/gcc_64/qml/
+export QT_QPA_PLATFORM_PLUGIN_PATH=$HOME/Qt/5.5/gcc_64/plugins/platforms/
+./qgroundcontrol
--- a/deploy/qgroundcontrol-start.sh
+++ b/deploy/qgroundcontrol-start.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
-export LD_LIBRARY_PATH=$HOME/Qt/5.5/gcc_64/lib
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$HOME/Qt/5.5/gcc_64/lib
 export QML_IMPORT_PATH=$HOME/Qt/5.5/gcc_64/qml/
 export QML2_IMPORT_PATH=$HOME/Qt/5.5/gcc_64/qml/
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@dagar This still leaves the issue that the builds generated by travis are linking to libudev.so.0. This version is not available on Ubuntu. Whereas they shoud be linking to libudev.so.1 which is available. If I build it myself on my machine it correctly links to s0.1. So it seems like maybe there is something wrong on the travis machines.

dagar:
Which version of ubuntu are you using?

deploy/qgroundcontrol-start.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+export LD_LIBRARY_PATH=$HOME/Qt/5.5/gcc_64/lib
```
dogmaphobic: The user may have some other system libraries in the ld path. The proper way to define this would be:
```export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$HOME/Qt/5.5/gcc_64/lib```
That way you are *adding* a path rather than replacing it all.

**Comment by DonLakeFlyer** [comment]
> Which version of ubuntu are you using?

14.04 LTS. Multiple people have reported the libudev.s0.0 problem. Seems to be a new problem that popped up with the new installers.

File: deploy/qgroundcontrol-start.sh (hunk 1 above)
Comment (DonLakeFlyer) [inline deploy/qgroundcontrol-start.sh]: Thanks got it. Also, how can I get command line options to get through this?

File: deploy/qgroundcontrol-start.sh [hunk 2]
```diff
@@ -0,0 +1,6 @@
...
+export QML_IMPORT_PATH=$HOME/Qt/5.5/gcc_64/qml/
+export QML2_IMPORT_PATH=$HOME/Qt/5.5/gcc_64/qml/
+export QT_QPA_PLATFORM_PLUGIN_PATH=$HOME/Qt/5.5/gcc_64/plugins/platforms/
+./qgroundcontrol
```
dogmaphobic: To pass along all command line arguments you can do this:
```
./qgroundcontrol "$@"
```

DonLakeFlyer:
Do folks think this pull is good to go now? I've testing it and it works with online installer.

dagar:
I'll test this when I get home.

DonLakeFlyer:
Thanks a ton Daniel. I know I push pretty hard on you guys who have real jobs as well.

dagar:
Okay I'm not seeing the problem with the latest qgroundcontrol.tar.bz2 and my 14.04 machine.
I'll try a fresh vm.

    % ldd ./qgroundcontrol | grep udev
        libudev.so.0 => /lib/x86_64-linux-gnu/libudev.so.0 (0x00007f5ba8b78000)

Travis-ci is currently ubuntu 12.04, but I believe it's possible to get access to newer ubuntu images now. I'll look into it.

PR #2254, opened 2015-11-21
travis-ci ubuntu trusty

dagar:


dagar:
This brings travis-ci up to Ubuntu 14.04.3, but deploy appears to be broken.

DonLakeFlyer:
It passed, woohoo!

dagar:
It's quite a bit slower, but it's on a working build on less ancient ubuntu. I'll undo it for android and the doxygen build to save some time.

PR #2256, opened 2015-11-21
Add support for v4 boards

LorenzMeier:
@davids5 @kd0aij This adds support to flash v4 boards through QGC.

### DIFF
--- a/src/VehicleSetup/Bootloader.h
+++ b/src/VehicleSetup/Bootloader.h
@@ -129,7 +129,7 @@ private:
         
         INFO_BL_REV         =   1,    ///< bootloader protocol revision
         BL_REV_MIN          =   2,    ///< Minimum supported bootlader protocol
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
Can you wait on this until I get auto connect through? This is all different there.

LorenzMeier:
Rebased on master.

DonLakeFlyer:
Are only the bootloader ids different? The serial port ids are the same?

LorenzMeier:
Hm, in the current version yes. But we probably should fix it up in the string as well.

DonLakeFlyer:
Can you make sure QGCSerialPortIds.h/cc is correct as well for the new boards.

LorenzMeier:
Merging as change was trivial and previous runs finished. Works fine with v4 boards.

PR #2257, opened 2015-11-22
Joystick fix: Enable joystick right after calibration & do not send anything to vehicle if not configured

LorenzMeier:
The joystick API had two corner cases:

  * It did not send anything to the vehicle right after calibration if not in the calibration view
  * It did start to send manual control commands to the vehicle during the calibration, even if the joystick was not enabled

There was a third issue which was just a bad user model: The calibration did not enable the joystick by default, which left users stumble over the non-obvious checkbox in the section below. Now its all correct and easy.


LorenzMeier:
Merging since merged state passed.

PR #2261, opened 2015-11-22
Mission Item raw/friendly edit support, plus more [merge after Stable 2.8]

DonLakeFlyer:
Mission Items now support friendly editing and raw editing. Friendly editing, is where we have a user friendly ui for the MAV_CMD. The friendly ui is driven from a json file, so defining friendly ui for new MAV_CMDs is trivial:
screen shot 2015-11-22 at 11 09 56 am
Friendly editing only shows the list of commands supported by the firmware.

By clicking the cog icon you can switch to raw editing which provides full access to all mission item settings:
screen shot 2015-11-22 at 10 57 56 am
Raw editing provides access to all possible MAV_CMDs. Features like waypoint dragging still work while raw editing.

FactSystem changes:
- You can now set a translator into fact meta data. This is used to translate a value from a user visible value to the internally stored value. With MissionItem this is used to show degrees to the user for friendly edit, but store radians internally. Not quite ready to support parameter translation say from meters to feet, but close.
  - void FactMetaData::setTranslators(Translator rawTranslator, Translator cookedTranslator)
  - typedef QVariant (*Translator)(const QVariant& from)
  - Fact::value returns translated value
  - Fact::rawValue returns untranslated value
- You can now set a list of enum strings and values into FactMetaData. FactComboBox then uses this data to show a user friendly list of options to pick from. This new feature will also eventually support enums in the parameter meta data. Just need to define the xml format for it. Coming soon.
  -     void FactMetaData::setEnumInfo(const QStringList& strings, const QVariantList& values);

FirmwarePlugin changes:
- virtual QList<MAV_CMD> FirwarePlugin::supportedMissionCommands(void) returns the list of MAV_CMD's that are supported by the firmware.


DonLakeFlyer:
> virtual QList FirwarePlugin::supportedMissionCommands(void) returns the list of MAV_CMD's that are supported by the firmware.

@LorenzMeier There is where I could use some mechanism from the firmware side to determine this list instead of hardcoding into QGC.

Fojtik:
Yes, nice.

I have in LAT displayed a number 137190049871108.  Please could you fix double number formating to the left edge.

When LAT contains 51.137190049871108, the number seen in editbox seems crazy.

LorenzMeier:
You need 9-11 digits after the dot to get full accuracy. So you can realistically only strip about three of them.

Fojtik:
Yes, it is a good idea to feed full number digits to editbox.

But when editbox gets automatically scrolled left, you do not see the beginning of the number. It is better to display beginning and let less significant digits not to be displayed unless you scroll them.

dogmaphobic:
@Fojtik Can you upload a screen grab with the Lat/Lon showing?

Fojtik:
wpedit

dogmaphobic:
I see. The width of the control is too narrow so it doesn't show what it should. The UI also seems to be not current master. What OS is that? This is what it should look like:

screen shot 2015-11-24 at 6 49 19 am

Fojtik:
I am using Windows 7 and Windows XP. It has something to do with current font width and control width.

dogmaphobic:
Right. Let me find a Windows machine and try it out. Windows can have system wide fonts *Larger/Smaller*. Are you using its default state or have you changed it? I'm just asking so I can see how to work around it. The control width should be proportional to the font (so it looks like the screen grab I sent above).

Thanks!

Fojtik:
I do not remember whether I have changed a font. But tomorrow I could test another on Windows machine.

But the waypoint editor seems to me still pretty unfinished. When I double click to area it does mad unwanted zooming. Is there any way how to turn off "smart" autocentering?

dogmaphobic:
> But the waypoint editor seems to me still pretty unfinished. When I double click to area it does mad unwanted zooming. Is there any way how to turn off "smart" autocentering?

This is a first cut on an entirely new mission editor. Don wrote it all from scratch. It has a lot of rough edges but the basics are there. The idea to get it out now is exactly to get feedback from users and make it better. The *smart* auto centering is a bit disconcerting to me at times as well. It would just be a matter of making it an option.

Fojtik:
It provides not only automatic centering, but unwanted automatic zooming.

DonLakeFlyer:
@Fojtik Can you move this to a separate issue please? The problem is that the max decimal places are not being set correctly.

PR #2269, opened 2015-11-23
Reflow flight mode layout even when joystick control is enabled

NaterGator:
With COM_RC_IN_MODE the flight mode tab doesn't update QML layout and clips/overlaps in a cluttered fashion. This allows layout updates even when blocked by the modal joystick mode warning pane.

image


LorenzMeier:
Thanks! Looking good!

DonLakeFlyer:
Thanks Nate

PR #2270, opened 2015-11-24
README.md: Explain why Qt distro packages can't be used

erikd:


DonLakeFlyer:
@dogmaphobic How do you want to word this? This reads more like QGC is doing someting wrong which it isn't. But still good idea to not have people fight with packages that may or may not work since we are at the whim of the maintainer.

DonLakeFlyer:
Merging in an Gus can change if he wants.

PR #2276, opened 2015-11-24
Fix telemetry radio RSSI calculation for SI1000 based modems

NaterGator:
The old RSSI value was computed by halving the RSSI figure injected in the mavlink stream. Originally we were halving the injected value, likely based on this statement in the datasheet about RSSI resolution:

image

Actual input power follows the following curve:
image

This PR implements accurate input power calculations for telemetry radios that identify themselves matching the [DIYDrones Si1k firmware](https://github.com/Dronecode/SiK/blob/c264c6766ddf778e75fe0b901635ace90b0836f7/Firmware/radio/mavlink.c#L133).

### DIFF
--- a/src/comm/MAVLinkProtocol.cc
+++ b/src/comm/MAVLinkProtocol.cc
@@ -291,8 +291,25 @@ void MAVLinkProtocol::receiveBytes(LinkInterface* link, QByteArray b)
                 // process telemetry status message
                 mavlink_radio_status_t rstatus;
                 mavlink_msg_radio_status_decode(&message, &rstatus);
+                int rssi = rstatus.rssi,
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
@LorenzMeier @dogmaphobic Can you folks who know something about RSSI look at this and merge if ready.

dogmaphobic:
At a glance, it all looks fine to me. I had mentioned this eons ago when I first started messing with telemetry RSSI. For now, we don't even have a way to display it so merging won't affect anything. Once I get the Telemetry RSSI indicators up, I will worry about how to interpret the three states (green, orange, red). The indicator pop up will show the dBm values directly along with other pertinent radio status.

PR #2281, opened 2015-11-25
Add virtual joystick setting.

DonLakeFlyer:
Also reflow settings pane to take less space

### DIFF
--- a/src/QmlControls/QGroundControlQmlGlobal.cc
+++ b/src/QmlControls/QGroundControlQmlGlobal.cc
@@ -29,7 +29,9 @@
 
 #include <QSettings>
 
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
@dogmaphobic Can you help me with this. Need to use QGroundControl.virtualTabletJoystick value to lay out the Fly view to make room. Need to keep bottom edges for thumbs. Current WIP is using 1/4 of available screen height.

screenshot_2015-11-24-16-43-43

dogmaphobic:
I'll take a look tomorrow. I actually had already made the graphics for the *joystick*.

DonLakeFlyer:
At this point just looking to move the instruments out of the way.

PR #2282, opened 2015-11-25
Android Virtual joystick [WIP for debugging only]

DonLakeFlyer:


DonLakeFlyer:
I've turned this on for desktop builds as well. With that I can now fly HIL in POSCTL with just a mouse on desktop. It should theorertically work in SITL as well but I can't get that to work.

DonLakeFlyer:
Closing this pull since the real one is ready now

PR #2285, opened 2015-11-25
Toolbar work

dogmaphobic:
* Added a Telemetry RSSI indicator to the toolbar
* Added a preference panel for setting what to show on the toolbar

By default, Telemetry RSSI and Accumulated Battery Consumption (mA) are disabled.

Screen grab showing everything enabled:

screen shot 2015-11-25 at 11 31 32 am

### DIFF
--- a/qgcresources.qrc
+++ b/qgcresources.qrc
@@ -75,6 +75,8 @@
 
         <!-- Toolbar -->
         <file alias="Armed.svg">src/ui/toolbar/Images/Armed.svg</file>
+        <file alias="ArrowRight.svg">src/ui/toolbar/Images/ArrowRight.svg</file>
+        <file alias="ArrowLeft.svg">src/ui/toolbar/Images/ArrowLeft.svg</file>
         <file alias="Battery.svg">src/ui/toolbar/Images/Battery.svg</file>
         <file alias="Connect.svg">src/ui/toolbar/Images/Connect.svg</file>
         <file alias="Disarmed.svg">src/ui/toolbar/Images/Disarmed.svg</file>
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
My view on ui settings:
- Get the UI right so you don’t need them at all
- They just add clutter and are not used by 99% of the user base. Whatever the defaults are, is what is going to be used.
- It also ends up as a support nightmare. Users turn things off, forget and then ask where the feature they need is.

I just don’t think this is a good idea. to do this.

dogmaphobic:
> Get the UI right so you don’t need them at all

The problem is what can you define as *right*. A novice user would be overwhelmed with too much information while an advanced user wants more information. This whole thing started when you said you wanted an indicator for battery consumption. For most people, knowing the battery percentage is enough. Others want to know the accumulated mA as this is how they keep track of the battery usage.

This is a way to have a *friendly* UI while allowing advanced users to add advanced options.

dogmaphobic:
For instance, @NaterGator fixed the telemetry RSSI in #2276 exactly because he needs it in the UI. I mostly don't use radio so I don't need it.

DonLakeFlyer:
He needs to see it all the time, or he needs to see it when it crosses some important bad thresholds?

dogmaphobic:
> when it crosses some important bad thresholds?

And how would the UI tell the user the telemetry is going bad? I think an alert message box to be way more intrusive unless it's about something that will cause the whole thing to blow up imminently.

dogmaphobic:
It's like my car that has an indicator telling me my *spare* tire is low on air pressure. I couldn't care less. if I get a flat, I call AAA. However, if I am to cross the Serengeti, I would want it to show me.

DonLakeFlyer:
And he falls directly into my trap! You just described how the toolbar should work with some of its indicators. There is no setting in your car to turn off the display of showing you the little flat tire indicator in the dash. Instead it only shows up when you need it.

dogmaphobic:
:P

It shows the air pressure in real time for all tires, including the spare, all the time :)

NaterGator:
I don't really have too much of a dog in this fight but I do have experience with UAS pilots who find information like telemetry RSSI to be critical for flight operations. I think in certain circumstances it can be ignored safely but absolute absence of that information is a bad thing :tm: for users. On the subject of infinite configurability and the ability to turn indicators off... I get what you are going for but I think it is the wrong solution to the right problem (we need to minimize space consumed by UI elements that are absolutely always going to be on screen). 

One compromise I like is to dedicate a space for heads up information like this and cycle through the set of indicators inside that same space. Using a weighting function to add additional "dwell" time on indicators that are approaching a fault condition and upon a fault pop them out into their own permanently displayed UI element until the fault condition is cleared. The trick is in tweaking the default cycle time, thresholds, and weighting functions for indicators to maintain quick-glance awareness of important details without cluttering the UI and taking up precious space on mobile devices.

PR #2289, opened 2015-11-26

tcanabrava:

There are still a few rougth edges ( that I blame for testing it on an ardupilot emulator than a real device ) but I think it's an mergeable work.

apm-groundcontrol

### DIFF
--- a/src/AutoPilotPlugins/APM/APMAirframeComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMAirframeComponent.qml
@@ -21,7 +21,7 @@
 
  ======================================================================*/
 
-import QtQuick 2.2
+import QtQuick 2.5
 import QtQuick.Controls 1.2
 import QtQuick.Controls.Styles 1.2
 import QtQuick.Dialogs 1.2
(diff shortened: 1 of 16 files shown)


src/AutoPilotPlugins/APM/APMAirframeComponent.qml
```diff
@@ -39,13 +39,225 @@ QGCView {
...
+            _boxWidth = rw / idx
+        }
+    }
+
+    APMAirframeComponentController {
+        id:         controller
+        factPanel:  panel
+
+        Component.onCompleted: {
+            if (controller.showCustomConfigPanel) {
```
DonLakeFlyer: This should only happen if the value of the FRAME parameter is set incorrectly.

src/AutoPilotPlugins/APM/APMAirframeComponent.qml
```diff
@@ -39,13 +39,225 @@ QGCView {
...
+        id: customConfigDialogComponent
+
+        QGCViewMessage {
+            id:             customConfigDialog
+
+            message:        "Your vehicle is using a custom airframe configuration. " +
+                                "This configuration can only be modified through the Parameter Editor.\n\n" +
+                                "If you want to Reset your airframe configuration and select a standard configuration, click 'Reset' above."
+
+            property Fact sys_autostart: controller.getParameterFact(-1, "SYS_AUTOSTART")
```
DonLakeFlyer: No such parameter on APM Stack

DonLakeFlyer:
Airframe Config for APM Stack does two things:
1) It sets the FRAME parameter
2) You can optionally copy in a set of default params for you vehicle

In Mission Planner it looks like this:
screen shot 2015-11-29 at 7 18 46 pm

You're missing the concept of setting the FRAME parameter. Also each different "FRAME" should be a separate airframe type ui box. Then the combo for each box should only include the vehicles which are associated with that FRAME type. For example I believe the Iris is a V Tail FRAME type. So the Iris should only be in the combo for the V Tail FRAME type.

src/AutoPilotPlugins/APM/APMAirframeComponent.qml
tcanabrava: Fixed.

src/AutoPilotPlugins/APM/APMAirframeComponent.qml
tcanabrava: That was a forgotten part from your code-stub, removed.

tcanabrava:
Don, all comments are fixed (including the misssing FRAME parameter), do you want me to create another pull request or send on top of this one?

DonLakeFlyer:
Just put it on top of this one for review. Then if the commit list looks too messy you can re-organize after the review is done.

My one last comment would be that when I tried using Mission Planner to set frame and load param set I've never noticed any sort of download delay. I think you said that MP was downloading the set as well. I think you need to restructure the code such that the initial frame selection comes up right away, but then the downloading happens in the background and fills in when available with some sort of indicator or something. The first time I ran your pull I just thought it was broken because the screen was just black. I initially gave up. Then I finally realized I needed to wait. The sequence that happens here needs to be much more user friendly.

tcanabrava:
closed because of a new PR. I didn't managed to send in top of this one because of a git rebase on top of master to make it mergeable.

PR #2297, opened 2015-11-28
Added extended GPS Info

dogmaphobic:
Also made toolbar background consistent between street and satellite maps. The *light* version was too light and hard to see certain colors.

Now making sure to close any open *pop up* item when switching views.

screen shot 2015-11-28 at 1 13 46 am


DonLakeFlyer:
How many times a day do you fly that same mission :)

dogmaphobic:
> How many times a day do you fly that same mission :)

That's the F450, which is the only one I have left with a 3DR Radio. It has the same *mission* loaded for months now and the last time I flew it was probably last spring :) I prop it against my office window so it can *see* the satellites. That's why the attitude indicator shows it bomb diving :) For power supply, I've been using an old UPS lead acid battery that weights about 3 times the entire drone.

PR #2300, opened 2015-11-28
small fix to recognize clone px4flow units

pritamghanghas:


DonLakeFlyer:
There is no good product and vendor ids for these as well?

pritamghanghas:
I actually have one from hobbyking. My friend has from somewhere else. Both
by some manufacturer in china I guess.

Here is the debug out of both, master doesn't recognize them. I just
followed the fix that is already there for clone radios. I guess they are
not modifying the vendor id or product id. I will assume these are whatever
comes with the virtual serial port on these devices.

Serial Port --------------
FirmwareUpgradeLog: port name: "ttyACM0"
FirmwareUpgradeLog: description: "PX4Flow v1.3"
FirmwareUpgradeLog: system location: "/dev/ttyACM0"
FirmwareUpgradeLog: vendor ID: 32902
FirmwareUpgradeLog: product ID: 7725


FirmwareUpgradeLog: Serial Port --------------
FirmwareUpgradeLog: board type 2
FirmwareUpgradeLog: port name: "ttyACM0"
FirmwareUpgradeLog: description: "PX4 FLOW v1.3"
FirmwareUpgradeLog: system location: "/dev/ttyACM0"
FirmwareUpgradeLog: vendor ID: 32902
FirmwareUpgradeLog: product ID: 7725



On Sat, Nov 28, 2015 at 11:24 PM, Don Gagne <notifications@github.com>
wrote:

> There is no good product and vendor ids for these as well?
>
> —
> <https://github.com/mavlink/qgroundcontrol/pull/2300#issuecomment-160324347>
> .
>



-- 
Regards,
pritam

DonLakeFlyer:
So I assume those IDs are just generic and not unique to a radio. So string fallback only way to detect.

pritamghanghas:
I don't know whether they are generic or not but I assumed just like clone
radios they will be. I will look them up.

PR #2304, opened 2015-11-28
Virtual thumb joystick

DonLakeFlyer:
- Turn on virtual joystick from QML General settings. May at some point move to Joystick config as new joystick type for dropdown. Still considering.
- Virtual joystick is available on both desktop (use with mouse) and mobile (use with thumbs)
- Add new QGCMapPalette which is used for light/dark color palette for different map types

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -258,6 +258,7 @@ HEADERS += \
     src/QGCFileDialog.h \
     src/QGCGeo.h \
     src/QGCLoggingCategory.h \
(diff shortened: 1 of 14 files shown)


dagar:
Could we make the virtual joystick work with WASD and arrow keys?

dogmaphobic:
> Could we make the virtual joystick work with WASD and arrow keys?

I'll take a look...

DonLakeFlyer:
Trying to get correct keyboard focus in Qt is a nightmare. It tends to screw up other things so be careful.

PR #2307, opened 2015-11-29
Thumb stick

dogmaphobic:
Experimental positioning of the virtual joystick.
When enabled, the main flight widget is replaced by a simple (temporary) display in the lower right corner.

A few notes:
* If you click/touch somewhere within the circle, the *joystick* jumps to that point. In real life, this would cause a radical change of control position. The position should only start to change once you hit the center of the joystick. Whatever point you started should be 0,0 (as the joystick is always auto centered when let go). At minimum, instead of jumping to the touched/clicked position, there should be a smooth ramp.
* There is a fixed high frequency timer (100Hz) sending joystick updates through the comm pipe. That timer should only be in high throttle when the stick is actually active (is 100Hz really necessary?). When idle (centered position), the timer could be a lot less frequent.
* Aesthetically, the joystick should not move beyond the circle. It is now moving all the way to the bounding rectangle. ~~The position function should convert to polar coordinates and operate within the boundaries of the circle~~.

screen shot 2015-11-29 at 1 31 39 am


dogmaphobic:
Come to think of it, on a real RC joystick, the movement is indeed within a square. The fix then would be to just restrict movement to within the square inscribed within the outer circle:

screen shot 2015-11-29 at 1 31 39 am copy

DonLakeFlyer:
> There is a fixed high frequency timer (100Hz) sending joystick updates through the comm pipe.
Crap I lost changes when I trashed my mac machine. Thought I had recreated all of them. Should be 25Hz which matches real joystick.

> Aesthetically, the joystick should not move beyond the circle.
If the joystick is going outside the rectangle (not circle) thev there is a bug. It should be bounded. I'll test. That said this still does not work correctly. If the finger goes outside the joystick rect pad it should continue to track, but peg to the appropriate boundary.  For example thumb goes out the top, stick should peg to top, but still track left/right position changes.

> The position should only start to change once you hit the center of the joystick.
Not so sure how this would really work. It's easy enough to hit close to the stick position when you are putting your fingers down. My HIL flying has been ok without it. Have to fly it to see if it's a real problem. Needs more HIL/SITL testing though before I'd be willing to fly something real with it.

DonLakeFlyer:
Like the new thumb graphics, cool! Still lots of space in the middle of the screen.

dogmaphobic:
The *joystick* is currently bound to the outer bounding rectangle instead in the rectangle inscribed within. Also, if I touch/click there, the joystick jumps to that position, though not always, which is even weirder.

screen shot 2015-11-29 at 2 32 08 am 3

LorenzMeier:
I think it looks good, but I wouldn't want to keep the approach of changing to a different main flight display. From a user perspective way too many things are jumping around based on what you have activated. Even more important, its a maintenance nightmare and the special widget will be always outdated and broken.

I can't seem to see a reason why you can't move the main indicator widget up? Why should this be a problem?

dogmaphobic:
>I can't seem to see a reason why you can't move the main indicator widget up? Why should this be a problem?

I explained in the discussion in gitter. I don't think the joystick on the lower edge would work, ergonomically speaking. Your natural thumb rest position is around the middle of the screen. At the bottom you would be in a *pinching* position fighting between keeping the tablet straight and thumbing around the joystick.

dogmaphobic:
The whole idea here was to get the thumbsticks on the screen so it could be tested. Once we figure out the best placement, I will spend the time redesigning the UI.

LorenzMeier:
Maybe its just my tablet (11") but on it there would be no space conflict.

dogmaphobic:
Yup. On every other (normal :) ) one, you have no space other than the very edge.

PR #2310, opened 2015-11-29
Change disconnect model for vehicles

DonLakeFlyer:
On some OS when you pull the cable on a usb device the port will close, on other OS it will not. Autoconnect was originally implemented assuming the former. Since that assumption is incorrect I needed to restructure auto connect as well as change how disconnecting an inactive vehicle works.

- Port is no longer automatically closed when you pull a usb cable
- Vehicles no longer automatically disconnect after a certain amount of inactivity time
- When QGC detects inactivity on a Vehicle it will show Connection Lost in the toolbar as well as a Disconnect button.
- QGC delays connecting to a discovered port in order to let the boot loader go by. On some OS, boot loader detection is flaky. If you mistakenly connect to the boot loader then auto connect just gets confused. You will now notice that auto connect is slightly slower. 6+ seconds instead of 1+ seconds.

The above changes fix the following problem on windows builds:
- You can now disconnect correctly
- You can usb connect to a board while QGC is already running.

### DIFF
--- a/src/AutoPilotPlugins/PX4/AirframeComponentController.cc
+++ b/src/AutoPilotPlugins/PX4/AirframeComponentController.cc
@@ -138,7 +138,7 @@ void AirframeComponentController::_rebootAfterStackUnwind(void)
         QGC::SLEEP::usleep(500);
         qgcApp()->processEvents(QEventLoop::ExcludeUserInputEvents);
     }
-    qgcApp()->toolbox()->linkManager()->disconnectAll(false /* disconnectAutoconnectLink */);
+    qgcApp()->toolbox()->linkManager()->disconnectAll();
     qgcApp()->restoreOverrideCursor();
 }
 
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -228,8 +228,6 @@ Item {
 
         QGCMapPalette { id: mapPal; lightColors: !isBackgroundDark }
 
-        Component.onCompleted: console.log("test", mapPal)
-
         MultiPointTouchArea {
             anchors.fill:       parent
             touchPoints: [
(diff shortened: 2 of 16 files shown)


LorenzMeier:
Why do you need to back off for the boot loader? It has a distinct device string. 6 seconds is not great from a user perspective and I had it solid for all OS based on the string.

DonLakeFlyer:
> It has a distinct device string.

Not on Windows it doesn't! The following would not work without waiting:
- Boot QGC
- Plug in direct usb to Pixhawk

I made the time slightly longer the necessary so it didn't end up being flaky.

LorenzMeier:
Why does it have to perform worse then on all OS? I understand you're striving for consistency, but Windows users are the fringe case of the fringe case.

DonLakeFlyer:
> Windows users are the fringe case of the fringe case.

That won't be the case if we start pulling APM users. If we succeed at getting more consumer oriented users, Windows will likely become the majority. That said I can change it to be OS specific. Generally I don't like OS specific differences since it's explodes the text matrix. But the timeout aspect of this is a simple enough area to not cause to many problems if it's OS specific.

DonLakeFlyer:
I'm going to see if I can tune it a little first before I make something OS specific. You still have to wait for the board to boot up before it will respond anyway. If you connect right away a bunch of the connect time is already inactive. So I may be able to tune it such that the actual time from plug board in to active vehicle may end up being the same.

LorenzMeier:
fair enough. BL is 5 seconds, so 6 makes sense

DonLakeFlyer:
> fair enough. BL is 5 seconds, so 6 makes sense

Maybe. I need to measure it. We'll see.

DonLakeFlyer:
I"m going to merge now since auto connect is pretty broken without it. I'll tune in a separate pull.

LorenzMeier:
I designed it. Its is 5 seconds.

DonLakeFlyer:
No I mean how much time it takes to connect using QGC. Depending on how you hit the timer it could be just under 12 seconds max. I should be able to change it to be just a little over the board boot time though.

PR #2329, opened 2015-12-02
Serial settings (QML)

dogmaphobic:
More work done with the link config in QML.
Serial configuration is complete.

Note that I've added an *Auto Connect* settings for link configurations but this is not yet used. More on that once I finish all link configurations.

Fixed an issue with the configuration save function. It was using the total count of configs but it should not count *dynamic* links. You would end up with bogus entries in the settings file.


dogmaphobic:
@dagar @DonLakeFlyer 

Huh?! Any ideas? No project files were changed:

```Project ERROR: Could not resolve SDK path for 'macosx10.11'```

dagar:
Temporary travis-ci issue. https://twitter.com/traviscistatus/status/671908340159541248
Re-running the OSX builds.

PR #2330, opened 2015-12-02

tcanabrava:

Now it correctly handles the FRAME variable and it's faster
regarding the download of the parameters for each type of Frame.

### DIFF
--- a/qgcresources.qrc
+++ b/qgcresources.qrc
@@ -207,5 +207,8 @@
         <file alias="AirframeFactMetaData.xml">src/AutoPilotPlugins/PX4/AirframeFactMetaData.xml</file>
         <file alias="ParameterFactMetaData.xml">src/AutoPilotPlugins/PX4/ParameterFactMetaData.xml</file>
     </qresource>
+    <qresource prefix="/AutoPilotPlugins/APM">
+        <file alias="AirframeFactMetaData.xml">src/AutoPilotPlugins/APM/AirframeFactMetaData.xml</file>
+    </qresource>
 
 </RCC>
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -293,7 +293,9 @@ HEADERS += \
     src/ui/uas/QGCUnconnectedInfoWidget.h \
     src/ui/uas/UASMessageView.h \
     src/AutoPilotPlugins/PX4/PX4AirframeLoader.h \
+    src/AutoPilotPlugins/APM/APMAirframeLoader.h \
     src/QmlControls/QGCImageProvider.h \
+    src/AutoPilotPlugins/APM/APMRemoteParamsDownloader.h
 
 WindowsBuild {
     PRECOMPILED_HEADER += src/stable_headers.h
@@ -405,7 +407,9 @@ SOURCES += \
     src/ui/uas/QGCUnconnectedInfoWidget.cc \
     src/ui/uas/UASMessageView.cc \
     src/AutoPilotPlugins/PX4/PX4AirframeLoader.cc \
+    src/AutoPilotPlugins/APM/APMAirframeLoader.cc \
     src/QmlControls/QGCImageProvider.cc \
+    src/AutoPilotPlugins/APM/APMRemoteParamsDownloader.cc
 
 !iOSBuild {
 SOURCES += \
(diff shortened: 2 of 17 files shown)


DonLakeFlyer:
Tried APM ArduCopter MockLink as well as APM flashed board. I get this when I go to airframe config:

screen shot 2015-12-03 at 1 11 31 pm

tcanabrava:
Don, there was check to see if the app was build in unitTestMode that loaded a PX4 file, I overloocked that, but if your code wans't build in unitTestMode, then I have no idea what could have been wrong it seems that it's loading the PX4 plugin (the XML that I use to populate the APM is very different from this one that is being displayed)
apm-planner-don
. I'm testing the code using ardupilot ArduCopter + my branch and this is what I get.

DonLakeFlyer:
I running a debug build which is always unit test mode.

tcanabrava:
@DonLakeFlyer - ah, this explains it. my last push fixed that. sorry, I was running and testing in release mode.

DonLakeFlyer:
Ok, running against an Iris setup using Mission Planner with latest firmware. Titles are strange. Iris should say V-Tail instead it says V 2. I'm not sure using the same visuals as PX4 will work. Two reasons:

1) Those quad images, show prop rotation. I'm not sure those rotations are the same on APM.
2) The param set you chose is not saved anywhere. For example my Iris has a Tarot. When I used Mission Planner is used the Iris + Tarot gimbal parameter set to setup. But in QGC, Iris+Tarot is not selected which is confusing. The first item in the combo is selected, which is not necessarily correct from a standpoint of the previously loaded param set. I think replicating the Mission Planner UI would be better than trying to use the PX4 visuals. The user model is different so trying to make them the same doesn't really work.

screen shot 2015-12-04 at 6 27 47 pm

LorenzMeier:
@DonLakeFlyer The prop rotations are exactly the same. You also can't just steal APM images as they might have a certain license attached to them.

DonLakeFlyer:
Ok, so you can use the Images but you should design the ui and user model to be the same as Mission Planner.

tcanabrava:
Don,

I don't think two different configuration views for airframes in the same
software is a good thing, (I made something similar to apm planner but
using qgroundcontrol visuals), I'll take a look on how mission planner is
doing it, but staying similar to what qgroundcontrol has is a good thing
for discoverability.

Tomaz
Em 5 de dez de 2015 14:11, "Don Gagne" <notifications@github.com> escreveu:

DonLakeFlyer:
> I don't think two different configuration views for airframes in the same
> software is a good thing,

Not for APM users, they are only family with APM Planner and Mission Planner. So in reality trying to stay consistent only makes sense if the user model really is the same. The issue is that PX4 Stack keep track of a very specific vehicle that you have selecting using the SYS_AUTOSTART parameter. The values essentially stores both frame type and a default parameter set in a single value. The APM stack only store Frame type. It does not store the specific vehicle type. APM users are used to a two step process to work around this. They select frame type and then they optionally select a parameter set they can load.

Your implementation will end up being confusing. More detailed example:
- Select Frame type X
- Select Parameter set 3DR QUAD X4 RTF
- Click apply
- Move away from Airframe Config
- Go back, X Frame will still be selected but 3DR X8-M RTF will be selected in the dropdown.
- User is confused, thinks something is wrong. Selects 3DR QUAD X4 RTF yet again.
- Repeat...

There is no way to fix this problem since the selected parameter set is not saved with the vehicle. The UI you are using gives the illusion that this value should be saved, but it is not. It also differs from the PX4 implementation which does save the setting. That just doesn't pass my usability bar.

Similarity is good, but not when it is only slightly similar and the actual differences will cause confusion. In that case you are much better off with something different.

This needs to be a two step process like it is in APMP and MP. This makes it clear to the user what is actually happening.

tcanabrava:
Okay,

Will redo the configuration screen tomorrow and send a v4. Thanks for your
time on this as I have never worked with drones before.

I think I figured out a good solution that can use qgroundcontrol visuals
I'm a two step way.

Tomaz
Em 5 de dez de 2015 20:46, "Don Gagne" <notifications@github.com> escreveu:

DonLakeFlyer:
> I think I figured out a good solution that can use qgroundcontrol visuals
> I'm a two step way.

Awesome, thanks.

DonLakeFlyer:
I can't pull down down into my master because is has merge conflicts. You'll need to rebase to latest master.

PR #2332, opened 2015-12-02
Enable binding mode commands to joystick buttons

NaterGator:
This is the corresponding QGC support for https://github.com/PX4/Firmware/pull/3307 in PX4. It enables binding gamepad buttons to discrete mode switch commands.

### DIFF
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -174,7 +174,7 @@ bool PX4FirmwarePlugin::setFlightMode(const QString& flightMode, uint8_t* base_m
 
 int PX4FirmwarePlugin::manualControlReservedButtonCount(void)
 {
-    return 8;   // 8 buttons reserved for rc switch simulation
+    return 0;   // 0 buttons reserved for rc switch simulation
 }
 
 void PX4FirmwarePlugin::adjustMavlinkMessage(mavlink_message_t* message)
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -347,9 +347,7 @@ void Joystick::run(void)
                     }
 
                     // Mark the button as pressed as long as its pressed
-                    if (buttonIndex < reservedButtonCount) {
-                        buttonPressedBits |= buttonBit;
-                    }
+                    buttonPressedBits |= buttonBit;
                 }
             }
             
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -347,7 +347,9 @@ void Joystick::run(void)
                     }
 
                     // Mark the button as pressed as long as its pressed
-                    buttonPressedBits |= buttonBit;
+                    if (buttonIndex < reservedButtonCount) {
+                        buttonPressedBits |= buttonBit;
+                    }
                 }
             }
             
@@ -461,6 +463,10 @@ QStringList Joystick::actions(void)
     QStringList list;
 
     list << "Arm" << "Disarm";
+
+    if (_activeVehicle) {
+        list << _activeVehicle->flightModes();
+    }
     
     return list;
 }
@@ -558,6 +564,8 @@ void Joystick::_buttonAction(const QString& action)
         _activeVehicle->setArmed(true);
     } else if (action == "Disarm") {
         _activeVehicle->setArmed(false);
+    } else if (_activeVehicle->flightModes().contains(action)) {
+        _activeVehicle->setFlightMode(action);
     } else {
         qCDebug(JoystickLog) << "_buttonAction unknown action:" << action;
     }


src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
```diff
@@ -174,7 +174,7 @@ bool PX4FirmwarePlugin::setFlightMode(const QString& flightMode, uint8_t* base_m
 
 int PX4FirmwarePlugin::manualControlReservedButtonCount(void)
 {
-    return 8;   // 8 buttons reserved for rc switch simulation
+    return 0;   // 0 buttons reserved for rc switch simulation
```
DonLakeFlyer: If you look here: https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L1089. You'll see that the button bits are turned into RC channels. This is the reason why this is 8. The possible solution is to stop sending button bits in the MANUAL_CONTROL message. Not sure of the value of turning single position joystick switches into rc channels. The results can be unexpected depending on how you have flight modes set up. @LorenzMeier ?

DonLakeFlyer:
Also, where are the UI changes?

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
NaterGator: That's why I include https://github.com/mavlink/qgroundcontrol/pull/2332/files#diff-7179de35ff534916a12fd2e757566fe8R350. It doesn't set the button bits for unreserved buttons anymore. Effective PX4 stops trying to turn gamepad button presses into RC controller switch positions.

We have flight time on this and have not had issues with inadvertently getting an RC switch action from button presses.

NaterGator:
There are not any UI changes necessary. Under the joystick configuration panel you have all the available buttons listed and can select any flight mode to bind to them.

It looks like this (forgive the quirky font rendering):
image

LorenzMeier:
@DonLakeFlyer This looks good to me. Trust the people using the joystick 8).

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
DonLakeFlyer: > We have flight time on this and have not had issues with inadvertently getting an RC switch action from button presses.

It will turn buttons in rc channels if COM_RC_IN_MODE=2 if I remember correctly. Then whether strange things happen or not, depends on if you have Flight Mode config setup up for these channels. This is pretty easy to hit if you just switched your normal config to a HIL config.

You've changed the semantics of what manualControlReservedButtonCount is supposed to mean. It should match what the firmware supports. Firmware supports converting 8 buttons to rc channels as I pointed to. So for PX4 flight stack, those first 8 buttons get sent to the firmware. This is the same way the old joystick code worked. By setting reserved count to 0, you've just turned off that feature.

@LorenzMeier That said I don't quite get the usefulness of sending buttons to firmware. My suggestion is to remove this whole reversed count concept completely. And stop sending any button bits to MANUAL_CONTROL no matter what. But that is up to Lorenz since this was existing functionality that I"m maintaining.

DonLakeFlyer:
So your OK with completely removing the feature which sent button bits to MANUAL_CONTROL and the firmware turned them into RC channel values?

src/Joystick/Joystick.cc
```diff
@@ -461,6 +463,10 @@ QStringList Joystick::actions(void)
     QStringList list;
 
     list << "Arm" << "Disarm";
+
+    if (_activeVehicle) {
+        list << _activeVehicle->flightModes();
```
DonLakeFlyer: Ah, I missed this part to update ui. Sorry

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
LorenzMeier: @DonLakeFlyer Don't judge debug interfaces you're not using. I have defaulted the simulation to pure joystick, so it should be none of QGC's concern if someone enables RC emulation. And if they do and have buttons dual-assigned its their problem. Lets focus here on the plain joystick support.

LorenzMeier:
@DonLakeFlyer Send the bits. They will get ignored if the system is configured as joystick and if its set to RC emulation they are wanted. Pretty please don't try to solve this on the QGC side. This is an advanced feature that should never be used by an average developer or user.

DonLakeFlyer:
> They will get ignored if the system is configured as joystick and if its set to RC emulation they are wanted.

Ahhh! This is what exactly what I was trying to fix with reserved count.

Example:
- Run with COM_RC_IN_MODE=2 (rc emulation)
- Flight Mode config says channel 6 is Return to home
- Joystick config has button 2 configured to set flight mode Loiter
- Press joystick button 2
- Boom!
- QGC will send a set flight mode command for Loiter
- And Firmware will try to set flight mode to Return to home

If I remember correctly when I was testing this. Firmware wins.

LorenzMeier:
Yeah, don't use RC mode 2 if you don't understand the implications of the emulation. It was a mistake to default the system to it.

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
NaterGator: That's true but COM_RC_IN_MODE=2 is not a viable flight mode in my experience. IIRC that disable all RC checks which is a no-go for actual flying vehicles. 

I agree that I've changed the semantics of what it means; ultimately that is Lorenz's call but in my experience having to deal with treating momentary buttons like position toggle switches was a royal pain and prone to problems. In testing I found that older version of PX4 firmware were only marginally comfortable or trustworthy to fly; there were certain mode transitions that were not possible atomically because the user had to press a button to toggle one mode before the other.

I suppose we could still send the reserved button bit states for older PX4 firmware revisions that do not support the complete atomic mode switch behavior. The point of confusion there will be both conflicting sources of mode switch (decoded button press -> switch position vs. bound button SET_MODE commands), and older users wondering why all the auto modes do not properly transition if they have them bound.

LorenzMeier:
I think this commit I just pushed fixes the corner case you were concerned of: https://github.com/PX4/Firmware/commit/c4ed598dff0ff3f54a803f708b1cb4b9be98a9bb

NaterGator:
@LorenzMeier IIRC (sorry, my memory isn't great since we've been using this method of gamepad operation for about a month and a half now) COM_RC_IN_MODE=1 still decodes button bits and causes problems.

COM_RC_IN_MODE=2 buttons are interpreted [here](https://github.com/PX4/Firmware/blob/fa49d27d09670540d002fe17ed06572ea50247f9/src/modules/mavlink/mavlink_receiver.cpp#L1124)
COM_RC_IN_MODE=1 buttons are interpreted [here](https://github.com/PX4/Firmware/blob/fa49d27d09670540d002fe17ed06572ea50247f9/src/modules/mavlink/mavlink_receiver.cpp#L1152)

LorenzMeier:
Per my comment from a few seconds ago I removed the button decoding to fix exactly that case. Now Don doesn't have to reserve bits.

DonLakeFlyer:
I'll let you guys hash out what the firmware is going to change to or not.

@NaterGator A better solution is to remove the whole reserved count thing completely. Then only send button bits to firmware which are not already assigned by the user to something else.

LorenzMeier:
I think the wrap-up of this is that we need re-architect the manual control message to send:

- State of the input device
- Input method (joystick, RC emulation)
- The flight mode it wants (like in CMD_DO_SET_MODE)
- Maybe the failsafe mode if it can't get that flight mode?

LorenzMeier:
@DonLakeFlyer Indeed - locking out bits which have been consumed by QGC would deconflict this nicely.

DonLakeFlyer:
I think MANUAL_CONTROL message is actually fine as is. I would't mix in flight modes and what not into there.

NaterGator:
@LorenzMeier does it make sense for the manual control message to pass the desired flight mode? 

In cases of closed loop packet delivery it make sense for manual control packets to be (almost?) always open loop like best effort UDP traffic and mode switch commands to be closed loop (SET_MODE). 

It makes sense to me having separate messages for the state of the control device versus desired craft mode.

NaterGator:
@DonLakeFlyer per what buttons to send as bits, etc:

Long term I think it makes sense to have a "Custom..." choice in the drop down. From there perhaps one option is to bind a particular button to a particular bit. Another may be to send a COMMAND_LONG with a certain set of configured parameters (ex: to trigger a shutter mechanism on a custom payload). 

Just thinking out loud... we have a few gamepad devices and no two manufacturers are exactly alike regarding where each button is mapped. If a user is advanced enough to want to pass in button bits I would hope they can swing manipulating a custom configuration dialog.

Thoughts?

DonLakeFlyer:
Easily doable. The Mission Item editor code in raw edit mode essentially already provides an editor for this since it support editing all MAV_CMDS. Would be easy enough to use that for editing.

DonLakeFlyer:
> Easily doable.

FYI: I'm talking about the command_long part. But the bit thing could be done as well. Super high power user stuff though.

dagar:
We should update the COM_RC_IN_MODE param description while we're on the subject. It's rather confusing.

RC control input mode	0	2	0
The default value of 0 requires a valid RC transmitter setup. Setting this to 1 disables RC input handling and the associated checks. A value of 2 will generate RC control data from manual input received via MAVLink instead of directly forwarding the manual input data.

LorenzMeier:
Seems like we sorted out at leat the main corner cases. Since COM_RC_IN_MODE = 2 is super power user stuff and actively *disabled* in Firmware master I think this is good to go.

DonLakeFlyer:
> A better solution is to remove the whole reserved count thing completely. Then only send button bits to firmware which are not already assigned by the user to something else.

Needs to be changed to do this.

PR #2333, opened 2015-12-02
Completed Link Configuration Settings

dogmaphobic:
It's all there.

I still need to deal with *Auto Connect* configurations. Links added on the fly should be added to the list of configurations (they have a *dynamic* attribute and are not persistent). Once there, you should be able to Disconnect and Edit them. Once edited, it becomes persistent. None of this is happening yet. Waiting for the *Auto Connect* to settle down before adding this behavior.


DonLakeFlyer:
> Once there, you should be able to Disconnect and Edit them.

You can't really disconnect an auto connect link since it will just come back on the next pass. Disconnect only really works if the link is inactive and the cable is pulled. Hence the toolbar disconnect button when it loses heartbeat.

dogmaphobic:
The idea is that you should be able to disconnect it, which would take it off the "auto" list. Either until it's re-enabled or the the next run. You could then edit it and fine tune it. No different then a manually created link which you set its auto connect flag. There should be no real distinctions between manually created and auto created links. The *auto* part is just a convenience.

DonLakeFlyer:
The current model is the way it is after figuring out how to deal with a nightmare of edge case handling. If you can figure out a better way to do it and keep it all working: Firmware Upgrade, Airframe Config Restart, connect/disconnect/connect again, multi-vehicle connect/disconnect, vehicle going inactive, pulling the USB cable on Windows which does different things than other OS and so forth I'm fine with that. I have no issue with some completely different way to do it either. But as they say "You break it you own it!". It's quite the nightmare. Maybe my design is the cause for that. Don't know. I could believe it. Couldn't come up with anything better after multiple tries and failures of various edge cases.

My suggestion would be to lock down editing on a link which is set to Allow AutoConnect. If you want to change it you need to first make sure there are no active vehicles on it. If not, then turn off Allow AutoConnect, which will automatically disconnect it and remove it from the AC rotation. Now you can edit, then turn Allow AutoConnect back on. And it will go back into the AC rotation with the new values. It's confusing, but it will work with the existing model.

The reason I"m very paranoid of any change to the existing model is that I'm multiple weeks in on fighting with this thing. I know it you breathe on it the wrong way it will break.

PR #2361, opened 2015-12-06
Convert QGCMessageBox to QGCApplication::showMessage

DonLakeFlyer:
- Changed QGCApplication::showToolbarMessage to showMessage since it isn't in the toolbar any more
- Converted a pile of QGCMessageBox usage to QGCApplication::showMessage

A baby step towards native QML top level window for mobile


dogmaphobic:
Ugh... I was just doing this same thing. It started with the complaint about the bad colors. It turned out the text is Rich Text (HTML tags) generated within UASMessageHandler.cc. I had moved the *Crytical Message* display off the toolbar and into the main window. I will now wait for this to get merged. At least you did all the work I on the other end. I was concentrating on UASMessageHandler.cc (the source of the bad colors).

PR #2372, opened 2015-12-08
Mobile runs Native QML

DonLakeFlyer:
This was pretty pervasive so likely I missed some connection somewhere. But as far as I can tell it works.

Still not sure how you close an android app and get a window close event. So the concept of prompting for close when active connections is coded in their, but no idea how you trigger that.

@dogmaphobic Can you up level you Nexus 7 os and check that were are good there now. Then I'll update the play store listing.


kd0aij:
does this address any open issues?

DonLakeFlyer:
None that are in github that I know of.

PR #2376, opened 2015-12-11

tcanabrava:

Now it correctly handles the FRAME variable and it's faster
regarding the download of the parameters for each type of Frame.

apm-planner-final-2
apm-planner-final-4

### DIFF
--- a/qgcresources.qrc
+++ b/qgcresources.qrc
@@ -208,5 +208,8 @@
         <file alias="AirframeFactMetaData.xml">src/AutoPilotPlugins/PX4/AirframeFactMetaData.xml</file>
         <file alias="ParameterFactMetaData.xml">src/AutoPilotPlugins/PX4/ParameterFactMetaData.xml</file>
     </qresource>
+    <qresource prefix="/AutoPilotPlugins/APM">
+        <file alias="AirframeFactMetaData.xml">src/AutoPilotPlugins/APM/AirframeFactMetaData.xml</file>
+    </qresource>
 
 </RCC>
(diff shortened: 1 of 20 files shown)


DonLakeFlyer:
- The frame type names are wrong. They should match the names in screen shot below. "Plus style", "Y6B" and so forth. That is what APM folks are used to.
- The list of available parameter sets to load is shorter than what you get in APMP or MP. See the image below.
- If you run this against something other than ArduCopter (ArduPlane, ...) it will crash because only ArduCopter has the FRAME parameter. All other vehicle types should not show the Airframe config screen. You can test Copter versus Plane using MockLink connections.

screen shot 2015-12-11 at 2 42 52 pm

tcanabrava:
Don,

Thanks for the feedback,

I'll change the names to match the ones on apm planner / mission planner.

About the list of parameters, it's not loading a different set, the only
difference is that it will show only the vehicles of that particular type (
if you select V style then click on the parameters, it will filter for
vehicles that have a v style frame)

The last point I wasn't aware of, fixing it now.
Em 11 de dez de 2015 19:54, "Don Gagne" <notifications@github.com> escreveu:

DonLakeFlyer:
> The last point I wasn't aware of, fixing it now.

I hadn't thought about that either until I got the crash.

DonLakeFlyer:
Also there a travis build breaks that need fixing.

tcanabrava:
Will do, as soon as I get back home. 30h on a bus right now.
Em 11 de dez de 2015 23:19, "Don Gagne" <notifications@github.com> escreveu:

PR #2391, opened 2015-12-14
linux include required libraries

dagar:


DonLakeFlyer:
We're going to need this for our Stable 3.0 release

dagar:
This worked as of a few weeks ago. I'm just dragging my feet on testing.

Do you have any opinion on which distros and versions we should support? I'd say Ubuntu 14.04 - current at a minimum, but what's the next most popular distro?

Once this is tested and working well I'll add dpkg creation to the builds so that Ubuntu users will have a slightly nicer experience (menu entry with icon, no tar.bz2 extraction, etc).

DonLakeFlyer:
Aren't there some Qml source files that need to go in as well? Here is what's in OSX package:
screen shot 2015-12-25 at 11 25 29 am

DonLakeFlyer:
@dagar Any news on this. I think it's still missing the QML stuff.

dagar:
The QML stuff is under libs. It's pulled in using the qmake variable $$[QT_INSTALL_QML].

https://github.com/dagar/qgroundcontrol/blob/linux_libs/QGCSetup.pri#L140

DonLakeFlyer:
Ah, ok. So is this ready to merge then?

dagar:
This works on my main Ubuntu 14.04 machine, but I just tried it in a clean Ubuntu Virtualbox image and it doesn't work.

    QXcbIntegration: Cannot create platform OpenGL context, neither GLX nor EGL are enabled
    Failed to create OpenGL context for format QSurfaceFormat(version 2.0, options QFlags(), depthBufferSize 24, redBufferSize -1, greenBufferSize -1, blueBufferSize -1, alphaBufferSize -1, stencilBufferSize 8, samples -1, swapBehavior 2, swapInterval 1, profile  0) 
    Aborted (core dumped)

Everything in `ldd ./qgroundcontrol` but I'm still looking.

dagar:
This appears to be a virtualbox issues. 
Ready to merge.

DonLakeFlyer:
Anything in the README need to be updated?

dagar:
For now I'll remove the section about installing qt to a default location. On the download page (http://qgroundcontrol.org/downloads) we can remove the part about installing qt and just list the packages to install.

dagar:
On second thought I think the README is fine and we should just update http://qgroundcontrol.org/downloads

DonLakeFlyer:
Done

PR #2407, opened 2015-12-15
APM Copter Basic tuning

DonLakeFlyer:
screen shot 2015-12-15 at 10 38 23 am


DonLakeFlyer:
If you using Qml Slider control, beware:
```
    Component.onCompleted: {
        // Qml Sliders have a strange behavior in which they first set Slider::value to some internal
        // setting and then set Slider::value to the bound properties value. If you have an onValueChanged
        // handler which updates your property with the new value, this first value change will trash
        // your bound values. In order to work around this we don't set the values into the Sliders until
        // after Qml load is done. We also don't track value changes until Qml load completes.
        throttleHover.value = _throttleMid.value
        rollPitch.value = _rateRollP.value
        climb.value = _rateClimbP.value
        rcFeel.value = _rcFeel.value
        _loadComplete = true
    }
```

PR #2417, opened 2015-12-16
add rotation for Stryker VTOL

kd0aij:
pending PR #3354 on PX4/Firmware


DonLakeFlyer:
Thanks Mark. I'll add enum support to the parameter meta data soon. It's mostly plumbed through QGC side already. Then we can just drive from the firmware side instead of needing pulls on both sides.

kd0aij:
Thanks Don.

Issue #2420, opened 2015-12-17
SubMenu buttons for setup need to scale width to largest button


DonLakeFlyer:
Fixed width sizing isn't working on all display dpis.

PR #2422, opened 2015-12-18
Handle dynamic sizing of setup buttons

DonLakeFlyer:
The setup button column width was a fixed size. Not only didn't it work on various dpis, it just can't be a fixed size. The plugin determines the text in the component buttons. So you don't really know till runtime how big they need to be.


DonLakeFlyer:
Fix for #2420

PR #2425, opened 2015-12-18
Esc key fix

pritamghanghas:
Fixes a small irritation. Mainwidnow had a shortcut action on escape that was connected to showNormal, which always led to a smaller minimized window irrespective of whether the application was full screen or not. Now it checks and always returns to maximized.


DonLakeFlyer:
Ah, that was the strange behavior in the dialog. Does Escape key work in the dialog now?

pritamghanghas:
No, It doesn't. But while debugging I noticed that the window getting
smaller has nothing to do with that qml.
The esc key works but not the first time. If I give the field focus
manually once by clicking on it. After that it works no matter how many
time I go to that dialog. So somehow valuefield is not getting focus
programmaticly the very first time and I couldn't find a solution to that.

On Sat, Dec 19, 2015 at 12:49 AM, Don Gagne <notifications@github.com>
wrote:

> Ah, that was the strange behavior in the dialog. Does Escape key work in
> the dialog now?
>
> —
> <https://github.com/mavlink/qgroundcontrol/pull/2425#issuecomment-165875294>
> .
>



-- 
Regards,
pritam

PR #2432, opened 2015-12-20
APM  plane fixes

DonLakeFlyer:
- Airframe component is always created. Just doesn't show up if not ArduCopter
- Flight Modes: Channel Options not available in all firmwares


DonLakeFlyer:
Fix for Issue #2430

Issue #2433, opened 2015-12-20
Do not connect via Radio if Autopilot is found on USB


LorenzMeier:
@DonLakeFlyer QGC currently connects via radio and USB if you connect them both (which is kind of common during setup). While in theory we want to support that on the autopilot side during operation, the setup / calibration code is not designed for this and all hell breaks loose.

The auto-connect code should only talk to the autopilot if it finds one connected directly via USB.

DonLakeFlyer:
This is not going to be easy to do. You don't know whether a link is to the same vehicle until you start talking to it and get a vehicle id back. You can't really do it from the auto-connect side of things. A way to do it would be if a vehicle has multiple links associated with it and one of those links is USB then when QGC sends mavlink messages it only sends to the USB link instead of all links. Would that work? Would it break some other use case? Is USB + Radio the only case of this problem?

LorenzMeier:
Don't connect to anything else if you find an autopilot connected. Its not normal that it is connected, so the user intended to do something specific with it.

DonLakeFlyer:
Hmm, not super understandable special case for multi-vehicle but will work and simpler change.

DonLakeFlyer:
This isn't going to work. If you have a radio hooked to your computer it will have already auto-connected to it. It's just sitting there waiting for a vehicle to show up. It's already too late to not connect to it. Radio will always be first, then USB.

DonLakeFlyer:
Back to not talking to additional links if USB is connected as possibly only option.

LorenzMeier:
Yes, and that seems to make a lot of sense as user model. By plugging the autopilot in the user told you he wants to talk to it through a fast link, and not through multiple.

DonLakeFlyer:
I'll get this done tomorrow.

DonLakeFlyer:
[...]

PR #2438, opened 2015-12-23
If USB directly auto-connected to board, only send/receive on that link

DonLakeFlyer:
This is a fix for #2433. It allows for Setup screens to not get confused by multiple messages from multiple links. But it specifically does not fix FTP for reasons which I don't understand. There does appear to be some unpredictable behavior on the firmware side if you connect usb, which also powers up your radio. Sometimes the radio will be seen by QGC first and then from the on nothing comes through the USB link side of things. Don't understand that behavior either.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -207,11 +207,15 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
         _addLink(link);
     }
 
+#if 0
+    // Unsure if this is the correct thing to do at this point. Trying without it.
+
     LinkInterface* usbDirectLink = _usbDirectLink();
     if (usbDirectLink && usbDirectLink != link) {
         // If we have a direct connection only listen to messages from that link
         return;
     }
+#endif
 
     // Give the plugin a change to adjust the message contents
     _firmwarePlugin->adjustMavlinkMessage(&message);
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -207,6 +207,12 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
         _addLink(link);
     }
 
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
@LorenzMeier This could use some testing.

DonLakeFlyer:
I wonder whether we should allow receive on all links and only force send to the one usb link?

LorenzMeier:
Yeah, maybe. And do that dynamically: Once the same system comes up on USB, only use that link consecutively.

LorenzMeier:
It kind of works: It comes up and connects, but it says it couldn't get params and retries. Does that help or do you need more detail?

DonLakeFlyer:
Does the retry succeed? That happens every once in a while. I think the firwmare boot time crossd some sort of threshold where it misses the first param request.

DonLakeFlyer:
> Once the same system comes up on USB, only use that link consecutively.

That's how it works.

DonLakeFlyer:
Changed it so that receive is allowedon all links, but outgoing send is restricted to Direct USB connection if any.

@LorenzMeier Try again?

LorenzMeier:
It doesn't work yet. It fails to load all params, screwing up the UI. If you refresh the param list manually it all works after.

If you connect a USB hub and plug your radio and Pixhawk to it you should be able to repro this easily.


screen shot 2015-12-26 at 12 58 05
screen shot 2015-12-26 at 12 57 24

DonLakeFlyer:
> If you connect a USB hub and plug your radio and Pixhawk to it you should be able to repro this easily.

I did that many, many times. I'll try it again. Through all the bugs I never had a case where you got partial parameters. That is very odd.

LorenzMeier:
I connected the hardware after QGC was running.

DonLakeFlyer:
In what specific order?

LorenzMeier:
Radio does connect first, although I connect them all at once (on purpose) through the hub.

DonLakeFlyer:
Closing this, since the approach is incorrect

PR #2439, opened 2015-12-23
Review iterations over QStrings

anselmolsm:
There are many ways of iterating over lists in C++/Qt.
In the specific case of QStrings, it is known [1] that using
foreach with const references avoids the creation of new
QString objects in each iteration, reducing the time consumed
by the loops.

[1] https://blog.qt.io/blog/2009/01/23/iterating-efficiently/

Signed-off-by: Anselmo L. S. Melo <anselmo.melo@intel.com>


DonLakeFlyer:
Looks good thanks. I'll have to keep in mind you can get a const out of the iterator.

Issue #2440, opened 2015-12-23
Airframe Config Usability


DonLakeFlyer:
[QGroundControl - Airframe Setup.pdf](https://github.com/mavlink/qgroundcontrol/files/71284/QGroundControl.-.Airframe.Setup.pdf)

We can discuss this document below and adjust as needed.

LorenzMeier:
We need these icons for PX4 (APM will use a subset):

Multirotors:
Quadrotor, Hexrotor, Octarotor: In X and + orientation (one arm to the front vs two arms symmetrical to the front). Preferably all in similar style with similar rotors (two blades).

Planes:

Standard plane:
https://store.3drobotics.com/products/3DR-Aero

Flying Wing:
https://pixhawk.org/platforms/planes/z-84_wing_wing

Inverted V-Tail (Albatross):
http://appliedaeronautics.com/Albatross-Airframe-MAX_p_1.html

Vertical takeoff and landing:
https://pixhawk.org/platforms/vtol/start#aircraft
  * Tailsitter (Duorotor): https://pixhawk.org/platforms/vtol/tbs_caipirinha_vtol
  * Tailsitter (Quadrotor): http://transition-robotics.com/collections/quadshots/products/the-mocha
  * Tiltrotor (Firefly Y6): https://pixhawk.org/platforms/vtol/birdseyeview_firefly
  * Standard VTOL: https://pixhawk.org/platforms/vtol/fun_cub_quad_vtol

Rover: Simple vehicle with four wheels
http://www.hobbyking.com/hobbyking/store/__39020__1_10_Quanum_Skull_Crusher_2WD_Brushless_Monster_Truck_ARR_EU_Warehouse_.html

Boat: Simple surface RC boat
http://www.hobbyking.com/hobbyking/store/__69214__FT009_High_Speed_V_Hull_Racing_Boat_460mm_Green_RTR_.html

dogmaphobic:
And *axial* multirotors:

![axial](https://cloud.githubusercontent.com/assets/749243/11995262/a9326ee0-aa1b-11e5-8002-7d1ca2622ee5.jpg)

For graphics, when applicable, the best would be vector files. I work with Illustrator but SVG files work just fine too.

One thing that should be brought up is the need for two color palettes. One for "Outside" (black text over white background) and another for "Inside" (white text over black background).

DonLakeFlyer:
> We need these icons for PX4 (APM will use a subset):

[...]

PR #2472, opened 2015-12-27
PX4 Airframe config closer the UI spec

DonLakeFlyer:
Working towards Issue #2440

screen shot 2015-12-26 at 4 57 30 pm


DonLakeFlyer:
I'm going to toss this pull in favor of future work which redesigns airframe config to also include wiring and motor details.

Issue #2461, opened 2015-12-25
Shutdown QGC while on Power Component will crash


DonLakeFlyer:
This bug is caused by the fact that the controller destructor tries to talk to the vehicle. But at that point the Vehicle is gone. There are a couple cases like this in Setup. Power and Radio I believe try to close down calibration states in the destructor. So seemingly simple bug, but tricky to handle.

DonLakeFlyer:
If you disregard multi-vehicle switching which needs to be reworked to one window per vehicle...

By then time you get to the destructor in a FactPanelController the vehicle is disconnected. So no need to try to communicate with it to adjust any state.

PR #2485, opened 2015-12-28
Power component shutdown

DonLakeFlyer:
Fix for Issue #2461


DonLakeFlyer:
AppVeyor seems to be down. It won't get past the nsis install. This happens every once in a while. Not it's 100% fatal.

PR #2487, opened 2015-12-29
Allow 23+9 bit Linux builds

DonLakeFlyer:


LorenzMeier:
23 bit! That sounds like a fancy HW architecture!

DonLakeFlyer:
Now you know why it hasn't been supported till now!

PR #2500, opened 2015-12-29
Using HDOP for GPS signal strength indicator.

dogmaphobic:
I'm now using HDOP for determining the GPS signal strength (the graphical indicator). The drop down message box for GPS now shows these additional *raw* data items (HDOP, VDOP and COG).

I could only test with SITL as none of my boards are in a state I can test GPS at the moment.

screen shot 2015-12-29 at 2 30 36 pm

P.S. While at it, that UAS.cc file is in desperate need for a make over...

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -90,6 +90,9 @@ Vehicle::Vehicle(LinkInterface*             link,
     , _batteryConsumed(-1.0)
     , _currentHeartbeatTimeout(0)
     , _satelliteCount(-1)
+    , _satRawHDOP(1e10f)
+    , _satRawVDOP(1e10f)
+    , _satRawCOG(0.0)
     , _satelliteLock(0)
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
> P.S. While at it, that UAS.cc file is in desperate need for a make over...

I'm working my way towards getting rid of it entirely. Moving everything to Vehicle. So if you need to add new mavlink message handling that should go in Vehicle.

dogmaphobic:
> I'm working my way towards getting rid of it entirely. Moving everything to Vehicle. So if you need to add new mavlink message handling that should go in Vehicle.

Do you mean you are moving ```void UAS::receiveMessage(mavlink_message_t message)``` to ```Vehicle```? It is indeed quite a bit of signal redirection as it is right now.

dogmaphobic:
> continuous-integration/appveyor/pr — AppVeyor build failed

Some Windows installer package issue?

PR #2510, opened 2016-01-02
Mission commands

LorenzMeier:
This adds some generic config for VTOL and enables these for PX4.

### DIFF
--- a/src/MissionManager/MavCmdInfo.json
+++ b/src/MissionManager/MavCmdInfo.json
@@ -553,7 +553,7 @@
         {
             "id":           222,
             "rawName":      "MAV_CMD_DO_GUIDED_LIMITS",
-            "friendlyName": "Exetrnal control limits",
+            "friendlyName": "External control limits",
             "description":  "Set limits for external control",
             "param1": {
                 "label":            "Timeout:",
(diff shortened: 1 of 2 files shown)


src/MissionManager/MavCmdInfo.json
```diff
@@ -588,7 +588,20 @@
...
-        { "id": 3000, "rawName": "MAV_CMD_DO_VTOL_TRANSITION", "friendlyName": "MAV_CMD_DO_VTOL_TRANSITION" },
+        {
+            "id": 3000,
+            "rawName": "MAV_CMD_DO_VTOL_TRANSITION",
+            "friendlyName": "VTOL Transition",
+            "description":  "Perform flight mode transition",
+            "category":     "Basic",
+            "param7": {
+                "label":            "Mode:",
+                "default":          0,
```
DonLakeFlyer: This should be either 3 or 4. Sets the default mode when a user adds a new one of these items.

LorenzMeier: I wanted to make sure the user picks the right mode and just doesn't leave the default. This is delicate, so "do nothing" is better than to slam in a transition - I think. I could also see arguments for hover being a good default mode.

DonLakeFlyer: You need to start out with a valid mission command though. 0 is undefined  in MAV_VTOL_STATE which is likely going to cause problem for other firmwares. If "Do nothing is an option" then there needs to be a "Do nothing" enum value. The fact that MAV_VTOL_STATE enum seems to have enums for both the transition command as well as I'm assuming another command which queries state is not great. Those two things should really be in separate enums.

LorenzMeier:
Applied.

PR #2527, opened 2016-01-04
qmake only copy linux libs for installer

dagar:
Otherwise the installer is created for every debug and release build, which often fails outside of Ubuntu with Qt from the online installer.


dagar:
Fixed merge conflict.

PR #2531, opened 2016-01-04
Modernize

tcanabrava:
Test of adding c++11 to be able to change some signals/slots to the new syntax.


DonLakeFlyer:
Did you test HIL to make sure it works with new Eigen?

tcanabrava:
The eigen update is to a bug fix release, no new code added, so my guess is
that it works. I opened qgc and connected a px4 mocklink and a apm mocklink
to make sure thi gs didn't crash.
Em 5 de jan de 2016 18:41, "Don Gagne" <notifications@github.com> escreveu:

DonLakeFlyer:
It's a huge diff though.

tcanabrava:
Yes it is, since it's the seventh bug fix release from the 3.2 series. We
where in 3.2.0,  this is 3.2.7 =)
Em 5 de jan de 2016 20:10, "Don Gagne" <notifications@github.com> escreveu:

DonLakeFlyer:
@LorenzMeier Do you think you could do an XPlane test with this pull?

DonLakeFlyer:
I should be able to get to this in a day or two. Just getting back from being out of town.

DonLakeFlyer:
Tested HIL, looks good.

PR #2547, opened 2016-01-06
APM Log Download "Fix"

dogmaphobic:
APM log entries are 1-based. Some times, the first log entry (entry 0) is sent with an uninitialized ID, zeroed size and date and the log entry count. It's not consistent. Most of the time the first packet is not sent at all. All other (valid) entries are 1-based (even though they all contain the total number of logs).

While "fixing" the log download to deal with this, I've also changed the date threshold to determine an *Unknown Date*. Those are logs recorded with no GPS time. The PX4 just have it zeroed (1 Jan 1970) while the APM returns something around the year 2000.

Renamed the saved log file from ".txt" to ".mavlink". It's not a text file and ".mavlink" is the extension used by the log replay widget.

### DIFF
--- a/src/ViewWidgets/LogDownload.qml
+++ b/src/ViewWidgets/LogDownload.qml
@@ -94,7 +94,7 @@ QGCView {
                             //-- Have we received this entry already?
                             if(controller.model.get(styleData.row).received) {
                                 var d = controller.model.get(styleData.row).time
-                                if(d.getUTCFullYear() < 1980)
+                                if(d.getUTCFullYear() < 2010)
                                     return "Date Unknown"
                                 else
                                     return d.toLocaleString()
--- a/src/ViewWidgets/LogDownloadController.cc
+++ b/src/ViewWidgets/LogDownloadController.cc
@@ -502,7 +502,12 @@ LogDownloadController::_prepareLogDownload()
         ftime = entry->time().toString("yyyy-M-d-hh-mm-ss");
     }
     _downloadData = new LogDownloadData(entry);
-    _downloadData->filename = QString("log_") + QString::number(entry->id()) + "_" + ftime + ".mavlink";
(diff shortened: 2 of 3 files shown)


LorenzMeier:
MAVLink is not the right log format. These are binar, self describing logs

dogmaphobic:
Are these different than the logs loaded by the log replay widget? What extension should be used?

LorenzMeier:
There are telemetry and flash logs. Flash logs use a different format. Px4 uses .px4log, APM probably .bin

LorenzMeier:
You can use .bin for all.

dogmaphobic:
Changed it. It's ".px4log" if PX4, ".bin" otherwise.

DonLakeFlyer:
> Some times, the first log entry (entry 0) is sent with an uninitialized ID, zeroed size and date and the log entry count. It's not consistent. Most of the time the first packet is not sent at all.

Could you put in Issue into ardupilot Github for that so they fix it?

dogmaphobic:
> Could you put in Issue into ardupilot Github for that so they fix it?

"Fixing" it is a subjective term. They would have to change they way they do things. If they fix their firmware, it would break APM Planner and possibly Mission Planner (assuming it loads logs). The protocol does not specify how to interpret the packets. They came up with their own, albeit weird way of handling it. Given that they were the first to implement and use this set of messages, technically, we are the ones doing it differently. If you call zero-based array index *different* that is.

DonLakeFlyer:
Ok. I thought you were saying it was flaky. Seems like if it's inconcistent then it would be a bug. If it's inconsistenly consistent then I guess that Ok :)

dogmaphobic:
The inconsistency is that you don't always get *entry 0*, which is supposed to only tell you how many logs there are. Some times you do, some times you don't. It doesn't matter because all entries contain the total log count.

PR #2564, opened 2016-01-10
Prioritize real joystick over virtual joystick.

rjehangir:
Added conditional statement to disable the virtualTabletJoystick output if a real joystick is connected and enabled. Without this, two different sets of values would be sent and would appear highly sporadic.


DonLakeFlyer:
Make sense, thanks.

PR #2565, opened 2016-01-10
correct spelling mistakes

nopeppermint:
-correct spelling mistakes found with https://github.com/lucasdemarchi/codespell


DonLakeFlyer:
This needs to be more targeted:
- xbee libs are third party so we shouldn't be modifying those
- same issue for qwt changes

DonLakeFlyer:
@nopeppermint You should just run this in the src directory so you don't hit thirdy party stuff. Any updates?

nopeppermint:
@DonLakeFlyer now only corrections in the src directory and the file QGCCommon.pri
are there.

DonLakeFlyer:
Awesome, thanks I'll go through it.

DonLakeFlyer:
Great thanks.

DonLakeFlyer:
Oops, wrong button.

Issue #2568, opened 2016-01-10
GPS indicator confusing


LorenzMeier:
@dogmaphobic The GPS indicator says no GPS available when the GPS module is actually found and sends data. It would be good if the wording would clearly say it found the GPS receiver but no lock or it didn't find a GPS receiver at all.

<img width="361" alt="screen shot 2016-01-10 at 16 24 38" src="https://cloud.githubusercontent.com/assets/1208119/12222246/11afeb84-b7b7-11e5-90e6-470472bc0079.png">

Issue #2566, opened 2016-01-10
QGC OS X latest with APM Copter GPS Status incomplete


lvale:
Hi,

I was checking for other issue and found that GPS Status is not being correctly reported.
APM reports these: 

https://github.com/diydrones/ardupilot/blob/6bc19cd44c6b538bff2d6c089c2a706c171b35a9/libraries/AP_GPS/AP_GPS.h#L75

   /// GPS status codes
    enum GPS_Status {
        NO_GPS = 0,             ///< No GPS connected/detected
        NO_FIX = 1,             ///< Receiving valid GPS messages but no lock
        GPS_OK_FIX_2D = 2,      ///< Receiving valid messages and 2D lock
        GPS_OK_FIX_3D = 3,      ///< Receiving valid messages and 3D lock
        GPS_OK_FIX_3D_DGPS = 4, ///< Receiving valid messages and 3D lock with differential improvements
        GPS_OK_FIX_3D_RTK = 5,  ///< Receiving valid messages and 3D lock, with relative-positioning improvements

but it's clear on this screen that it is still reporting status 3 while receiving status 4

<img width="771" alt="screen shot 2016-01-10 at 13 23 11" src="https://cloud.githubusercontent.com/assets/3130363/12221721/bcf86bba-b79e-11e5-99e0-db280dd9a431.png">

PR #2570, opened 2016-01-10
GPS Indicator fixes.

dogmaphobic:
This is in response to #2568 and #2566.

The indicator will only be enabled if GPS data is actually present. If there is no GPS receiver, you will see this:
screen shot 2016-01-10 at 2 00 36 pm

If there is a receiver but no signal (indoors for example), you get this:
screen shot 2016-01-10 at 1 46 33 pm

With proper reception, you get the full data (though even with 13 satellites, I didn't get beyond "3D Lock"):
screen shot 2016-01-10 at 2 11 34 pm

### DIFF
--- a/src/ui/toolbar/MainToolBar.qml
+++ b/src/ui/toolbar/MainToolBar.qml
@@ -190,16 +190,22 @@ Rectangle {
 
     function getGpsLockStatus() {
         if(activeVehicle) {
-            if(activeVehicle.satelliteLock == 0) {
-                return "No Satellite Link"
-            }
-            if(activeVehicle.satelliteLock == 1) {
-                return "No GPS Lock"
+            if(activeVehicle.satelliteLock < 2) {
+                return "No Satellite Lock"
             }
             if(activeVehicle.satelliteLock == 2) {
                 return "2D Lock"
             }
-            return "3D Lock"
+            if(activeVehicle.satelliteLock == 3) {
+                return "3D Lock"
+            }
+            if(activeVehicle.satelliteLock == 4) {
+                return "3D Lock with Differential"
+            }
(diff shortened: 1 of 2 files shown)


lvale:
@dogmaphobic Hi the 3D Status is not correlated with the number of satellites received but 4D would be DGPS lock and 5 it's RTK lock. While 5 requires a special RTK GPS unit, the DGPS requires a GPS that also listens to WAAS/EGNOS, etc which most of the recent uBlox can receive, but.... only in some parts of the world. We get it in Europe :)

dogmaphobic:
Interesting. I've been completely out of the loop regarding GPS technologies since the first consumer units came out in the 90s. I still have a Sony receiver that is probably stuck in 1916 (Y2K issue) :)

Out of curiosity, aren't these all low orbit? Do they disable services while over certain parts of the world? How would a same GPS receiver work in one place but not in another?

lvale:
I believe it has to do with stationary orbits, so US, Europe, Japan each have their own system :( and they are "somewhat" regional.

I still have a Garmin eMap :)

dogmaphobic:
> I believe it has to do with stationary orbits,

Stationary and orbit don't go together :) The only way to *seem* stationary is to position within the Clarke Belt in geo-synchronous orbit some 40,000+ km up. Wouldn't that be way too far for these tiny receivers to work?

I will take a look at those strings later today. Thanks!

lvale:
LoL. Stationary is obviously wrong in this context :) 

I went out and looked at the Galileo stats here : http://www.esa.int/Our_Activities/Navigation/Facts_and_figures

DonLakeFlyer:
GPS nerds

PR #2575, opened 2016-01-11
Android limit to tablets for now

### LINKED ISSUE #2551: Limit Play store installs to 7" tablets or larger
Hopefully there is some way to do this.

dagar:
- fixes #2551


dagar:
600dp is generally a 7"-or-greater screen - http://developer.android.com/guide/practices/screens-distribution.html

PR #2577, opened 2016-01-11
include qt plugins in linux package

### LINKED ISSUE #2542: Linux Daily Build (qgroundcontrol.tar.bz2) built by travis-ci doesn't work

dagar:
- fixes #2542


dagar:
It looks like we can replace the awkward qgroundcontrol-start.sh script with a qt.conf next to the qgroundcontrol binary.
http://doc.qt.io/qt-5/qt-conf.html

dagar:
Rebased to resolve merge conflict.

Using qt.conf isn't enough because the xcb plugin still can't find a qt lib without updating LD_PRELOAD_PATH, so we're still stuck with qgroundcontrol-start.sh for now.

PR #2596, opened 2016-01-13
"Fix" for hanging on exit when Video Streaming build is enabled.

dogmaphobic:
The destruction order changed and I could not figure why the thread was hanging on a select(). The solution was simply: *don't do it*.


dogmaphobic:
Something is flaky with Travis. The master build just failed with:
```
/home/travis/build.sh: line 45: doxygen: command not found
```

A previous commit (1 out of 3 and not this one) had this error too. I restarted the build and it passed. I don't know how to restart a master build (other than pushing something).

DonLakeFlyer:
I'm not sure how to re-kick master either. I've got some others going through. we'll see what happens

Issue #2465, opened 2015-12-26
Add support for RebootRequired parameter meta data


DonLakeFlyer:
If one if these i modified do a showMessage with the reboot required message

DonLakeFlyer:
When done, remove text from Camera Config

PR #2598, opened 2016-01-13
Add Reboot Required fact meta data

DonLakeFlyer:
Fix for Issue #2465. Parameters which are marked as requiring reboot to take affect in meta data will now warn user with message if param is changed.

@LorenzMeier If you update the firmware parser to support a @RebootRequired param tag which outputs:
```
<reboot_required>true</reboot_required>
``` 
to the meta data xml this will work with PX4 as well.

@dogmaphobic This should help with your Radio parameters. Working to plumb component meta data through the system.

screen shot 2016-01-13 at 10 33 18 am

### DIFF
--- a/src/FactSystem/ParameterLoader.cc
+++ b/src/FactSystem/ParameterLoader.cc
@@ -296,8 +296,8 @@ void ParameterLoader::_valueUpdated(const QVariant& value)
     _writeParameterRaw(componentId, fact->name(), value);
     qCDebug(ParameterLoaderLog) << "Set parameter (componentId:" << componentId << "name:" << name << value << ")";
 
-    if (fact->rebootRequired()) {
-        qgcApp()->showMessage(QStringLiteral("Change of parameter %1 requires a Vehicle reboot to take affect").arg(name));
+    if (fact->rebootRequired() && !qgcApp()->runningUnitTests()) {
+        qgcApp()->showMessage(QStringLiteral("Change of parameter %1 requires a Vehicle reboot to take effect").arg(name));
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/FactSystem/ParameterLoader.cc
```diff
@@ -295,6 +295,10 @@ void ParameterLoader::_valueUpdated(const QVariant& value)
 
     _writeParameterRaw(componentId, fact->name(), value);
     qCDebug(ParameterLoaderLog) << "Set parameter (componentId:" << componentId << "name:" << name << value << ")";
+
+    if (fact->rebootRequired()) {
+        qgcApp()->showMessage(QStringLiteral("Change of parameter %1 requires a Vehicle reboot to take affect").arg(name));
```
LorenzMeier: Isn't it effect?

dogmaphobic: Don's special English interpretation :) \<runs /\>

LorenzMeier:
@DonLakeFlyer Can I just have this?

```
<reboot_required />
```

DonLakeFlyer:
<reboot_required>true</reboot_required>

APM stack had a true/false in there so I matched that. Kind of odd. I could change to just a tag for PX4. Easy.

LorenzMeier:
I think you could just support both. I have matched it for now

LorenzMeier:
Its already on master. The PWM Output params require a reboot and now will trigger it.

src/FactSystem/ParameterLoader.cc
DonLakeFlyer: English snobs

Issue #2599, opened 2016-01-13
Airframe Config Icons


DonLakeFlyer:
First let's list out the needed icons and  consistent filenaming:
* Helicopter
  * Single rotor - Helicopter
  * Coaxial rotor - HelicopterCoaxial
* Flying Wing
  * Simple version (no rotors) - FlyingWing
  * VTOL Duo Tailsitter - VTOLDuoRotorTailSitter
  * VTOL Quad Tailsitter - VTOLQuadRotorTailSitter
* VTOL tiltrotor - VTOLTiltRotor
* Multi Rotor
  * Quad Rotor
    * Plus - QuadRotorPlus
    * X - QuadRotorX
    * V-Tail - QuadRotorVTail
    * A-Tail - QuadRotorATail
    * Wide - QuadRotorWide
  * Quad Rotor H Frame - QuadRotorH
  * Hexa Rotor
    * Plus - HexaRotorPlus
    * X - HexaRotorX
  * Octo Rotor
    * Plus - OctoRotorPlus
    * X - OctoRotorX
    * Coaxial Plus - OctoRotorPlusCoaxial
    * Coaxial X - OctoRotorXCoaxial
  * Tri Copter
    * Plus - YPlus
    * Minus - YMinus
    * Y6A - Y6A
    * Y6B - Y6B
* Rover - Rover
* Boat - Boat
* Plane
  * Standard - Plane
  * Inverted V Tail - PlaneATail
  * VTOL - VTOLPlane
* Simulation - AirframeSimulation
* Unknown - AirframeUnknown

The concept here is that the top most level of bullets specifiy a base "chassis" visual for the icon. This creates a consistent base visual. Then the sub-bullets vary that base chassis by adding rotors. I assume the extension for all is .svg.

We'll also use these names as the available set for airframe meta data xml image field. I'll include back-compat for old names in old meta data files. Note that for Aiframe MetaData image field use there is no file extension.

cfelipesouza:
Great @DonLakeFlyer! Just to clarify, I split into two sets just to organize myself and the file I sent you. I totally agree that both should use the same icons. :)

DonLakeFlyer:
Issues with current set:
- There is only one set of airframe icons for QGC that encompass both PX4 and APM. No need to split them out into two sets. A QuadRotor Plus icon for example is the same for both PX4 and APM.
- I think the Helicopter should use the same color circle rotor visualization as the others (instead of showing individual blades).
[...]

PR #2602, opened 2016-01-13
Use new Airframe Icons + Naming

DonLakeFlyer:
We're waiting on a new icon set to fix the draw problems. But even with that I think it's still better so I'm pushing through. Relates to issue #2599.

screen shot 2016-01-13 at 1 22 08 pm


DonLakeFlyer:
@LorenzMeier Firmware needs to sync up as well: https://github.com/PX4/Firmware/issues/3505

DonLakeFlyer:
I've also created placeholder dups for a few of the icons which are missing. This will also get fixed when we get the new icon set.

Issue #2609, opened 2016-01-14
Show Linux permission setup docs on port open failure


LorenzMeier:
We should show the instructions to add the user to dialout and remove modem manager to users. Then a bold hint to reboot (not just logout). The reason being that you need to re-plug the USB device after logout / login and its already too complex for most people. They continue to not read any documentation and every Linux user results in support overhead.

Issue #2608, opened 2016-01-14
Do not allow people to start QGC with sudo


LorenzMeier:
Too many stupid people run QGC with sudo to work around not having configured their system properly. Here is one way to work out if someone is running with root: http://stackoverflow.com/questions/26552517/get-system-username-in-qt

DonLakeFlyer:
Do you mean that the username will come back as "root" in these cases?

LorenzMeier:
I haven't tested it, but that was my thinking.

dogmaphobic:
Hmm, not sure about that. For Linux/Mac you would use uid. It returns 0 for root. This is irrelevant for Windows any way.

dogmaphobic:
You want to call ```getuid()```. If it returns 0, it's root (or sudo).
http://linux.die.net/man/2/getuid

PR #2612, opened 2016-01-15
Stupid proof Linux serial port use

DonLakeFlyer:
Fix for Issue #2609, #2608

@LorenzMeier @dogmaphobic Can you guys look this over. Especially for wording.

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -188,6 +188,45 @@ QGCApplication::QGCApplication(int &argc, char* argv[], bool unitTesting)
     setAttribute(Qt::AA_DontCreateNativeWidgetSiblings);
 #endif
 
+#ifdef Q_OS_LINUX
+#ifndef __mobile__
+    if (getuid() == 0) {
+        QMessageBox msgBox;
+        msgBox.setInformativeText("You are runnning QGroundControl as root. "
+                                  "You should not do this since it will cause other issues with QGroundControl. "
+                                  "QGroundControl will now exit. "
+                                  "If you are having serial port issues, execute the following commands to fix most serial port issues:\n"
+                                  "sudo usermod -a -G dialout $USER\n"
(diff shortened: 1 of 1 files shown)


dogmaphobic:
Missing:
```
#include <unistd.h>
#include <sys/types.h>
```

dagar:
I like the root and dialout group check, but I'm not sure about including Ubuntu specific instructions. Can that part at least be hidden behind an Ubuntu check? Even if Ubuntu is the only distro you want to officially support I know there are current users on Fedora, Arch, etc. 

You can determine if it's Ubuntu by looking at /etc/os-release

    NAME="Ubuntu"
    VERSION="14.04.3 LTS, Trusty Tahr"
    ID=ubuntu
    ID_LIKE=debian
    PRETTY_NAME="Ubuntu 14.04.3 LTS"
    VERSION_ID="14.04"
    HOME_URL="http://www.ubuntu.com/"
    SUPPORT_URL="http://help.ubuntu.com/"
    BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"

DonLakeFlyer:
What is Ubuntu specific?

dagar:
This is only valid on Ubuntu and Debian

    sudo usermod -a -G dialout $USER
    sudo apt-get remove modemmanager

DonLakeFlyer:
Ah, didn't realize that. We use those same instructions on both the QGroundControl and PX4 web site without mention of Ubuntu. I'd rather not get too fancy since that stuff just tends to break in weird ways. How about just modify the text to mention Ubuntu and/or Debian?

dagar:
Yes that should be fine.

LorenzMeier:
Mentioning Ubuntu should be fine. You only get the union of ```stupid user ∪ Linux``` for Ubuntu and not so much for other Linux systems.

Issue #2636, opened 2016-01-16
Joystick should not be in APM


DonLakeFlyer:
Same thing for thumbstick

DonLakeFlyer:
@LorenzMeier Any ideas where given #2642 which has MANUAL_CONTROL support on custom APM stack. The issue is that Joystick shows a lot of UI that may not work correctly on all firmwares. But whether joystick support is there or not can't really be trigggered on firmware type. Needs to be triggered some other way.

DonLakeFlyer:
New values for MAV_PROTOCOL_CAPABILITY?

DonLakeFlyer:
@rjehangir Adding you, so you can follow along on this one.

rjehangir:
Don, thanks. I want to chime in on this issue because it's important for our work. We're developing ArduSub (https://github.com/bluerobotics/ardupilot) and relying primarily on real and virtual joystick in QGC for user control input. 

We are happy to contribute to and test the joystick thoroughly to make sure that it functions correctly.

PR #2642, opened 2016-01-16
Joystick: Fix throttle center point.

rjehangir:
The lines removed set the center point of the throttle axis to one extreme of the joystick range. This causes an offset value when the stick is centered if the joystick does not return exactly zero. This is important for vehicles with reversible throttle, such as ground robots or boats.

Note: I've tested this code and the joystick behavior is almost identical except that the center point is correct for throttle. The "center stick is zero throttle" and "full down stick is zero throttle" works correctly and produces the same output as the original.

### DIFF
--- a/src/VehicleSetup/JoystickConfigController.cc
+++ b/src/VehicleSetup/JoystickConfigController.cc
@@ -386,11 +386,6 @@ void JoystickConfigController::_inputStickMin(Joystick::AxisFunction_t function,
             } else {
                 _rgAxisInfo[axis].axisMin = value;
             }
-
-            // Check if this is throttle and set trim accordingly
-            if (function == Joystick::throttleFunction) {
-                _rgAxisInfo[axis].axisTrim = value;
-            }
             
             qCDebug(JoystickConfigControllerLog) << "_inputStickMin saving values, function:axis:value:reversed" << function << axis << value << info->reversed;
             


DonLakeFlyer:
If I remember correctly the current joystick code does not support reversible throttle so you are likely to run into other problems. I'll have to look through this to remeber where/why.

rjehangir:
Correct, it does not support reversible throttle. We've got code set up to consider 500-1000 as forward, 500-0 as reverse. It breaks with the current version since the center throttle is not 500. This patch fixes that without affecting operation for any other modes.

rjehangir:
By the way, I'd be thrilled if QGC supported true reversible throttle.

DonLakeFlyer:
Ah, you're running APM Stack. What joystick mode are you using (which mavlink message? MANUAL_CONTROL?). This is going to run into problems with this: https://github.com/mavlink/qgroundcontrol/issues/2636. Since normal APM stack doesn't support MANUAL_CONTROL I had planned to remove it. We likely need to find some other way to determine if firmware has joystick support ot not.

rjehangir:
Yes, running APM stack and using MANUAL_CONTROL since it is supported by QGC. I'm open to changing that. Any suggestions on how we should avoid this issue then?

DonLakeFlyer:
You're doing the right thing by supporting MANUAL_CONTROL on your own. I just hadn't thought about custom firmwares based on APM with added MANUAL_CONTROL support. We'll figure out another way to handle this. Moving the joystick support by firmware or not discussion to #2636.

I have to look through the joystick code see how this throttle trim stuff work agagin so I can make sure there aren't any other problems with this change. Joystick has been very flaky, so I'm pretty praranoid about changing it.

DonLakeFlyer:
The throttle trim value is only part of the problem. Your change sort of fixes it but it isn't the right fix. It just happens to work. At least that is what I think at this point. I need to look at it some more to find the right fix.

rjehangir:
What else is wrong? I can look into it as well if you'd like.

DonLakeFlyer:
Nope, you're right I'm wrong. This is perfect. Sorry about that. The joystick values go through some many transformations it makes my head spin.

rjehangir:
Great. Thanks!

rjehangir:
Speaking of transformations, what's the motivation for mapping the values from a circular shape to a square box? In my opinion that creates a few issues:
- Limits range, which is fairly dramatic on small joysticks on gamepads
- Creates a very non-linear curve for the user, which is non-intuitive and goes against overall ease-of-use
Thoughts?

DonLakeFlyer:
> what's the motivation for mapping the values from a circular shape to a square box?

I'm not sure. @LorenzMeier You added that. Can you explain?

LorenzMeier:
@rjehangir It emulates the square shape of a real RC transmitter. If you don't do the mapping you're getting a different response compared to a RC stick (stick in bottom right corner is the same with the mapping, its not the same without, because most gamepads have a circular, not a rectangular aperture).

rjehangir:
@LorenzMeier Okay, I understand that and why it might be desirable. Are there any situation with the primary vehicle types where a circular shape would cause issues?

Right now, the square created is inscribed within the joystick circle. Would it be an issue to instead have the joystick circle inscribed in a larger square and map to that? For example, you could still arm a vehicle by going to the bottom right corner on throttle/yaw, but there would be less non-linearity (and no sharp non-linearity) in the joystick. That would retain full joystick range and achieve what you want.

PR #2576, opened 2016-01-11
[WIP] travis-ci create Ubuntu dpkg and ppa

dagar:
TODO:

- [ ] travis-ci gpg key for package signing and upload
- [ ] only deploy tagged builds to launchpad
- [ ] upload deb to s3 and github releases
- [ ] copy qt libs into deb
- [ ] deb dependencies (see #2540)
- [ ] shortcut icon for deb and linux tar.bz2

PR #2647, opened 2016-01-16
Update .travis.yml

nopeppermint:
as the travis build https://travis-ci.org/mavlink/qgroundcontrol/jobs/101515894 tells you 
``Cannot add PPA: 'ppa:beineri/opt-qt552-trusty'.``
=> Therefore
`` sudo add-apt-repository -y ppa:beineri/opt-qt552-trusty`` is the problem !
Possible alternative would be 
https://launchpad.net/~beineri/+archive/ubuntu/opt-qt521
or 
https://launchpad.net/~beineri/+archive/ubuntu/opt-qt532-trusty

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ before_install:
 
 install:
   - if [[ "${TRAVIS_OS_NAME}" = "linux" && "${CONFIG}" != "doxygen" ]]; then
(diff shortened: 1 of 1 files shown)


nopeppermint:
solves two problems of  #2576
but there are more to solve

Issue #2308, opened 2015-11-29
Virtual Joystick issues


DonLakeFlyer:
- If you click/touch somewhere within the circle, the joystick jumps to that point. In real life, this would cause a radical change of control position. The position should only start to change once you hit the center of the joystick. Whatever point you started should be 0,0 (as the joystick is always auto centered when let go). At minimum, instead of jumping to the touched/clicked position, there should be a smooth ramp.
- There is a fixed high frequency timer (100Hz) sending joystick updates through the comm pipe. That timer should only be in high throttle when the stick is actually active (is 100Hz really necessary?). When idle (centered position), the timer could be a lot less frequent.
- Aesthetically, the joystick should not move beyond the circle. It is now moving all the way to the bounding rectangle.

DonLakeFlyer:
Also see #2307 for some discussion

pritamghanghas:
I would like to add one more. I need to try joystick first in actual flight. But I think this should be added.
Right now the stick dot moves back to zero if you thumb moves outside the circle. If I have correctly picked the joystick once as mentioned in point 1 of @DonLakeFlyer comment, it should get stuck at max position if the thumb is moving outside of bounding rectangle. Reason, we wouldn't be looking at screen while flying and thumb may move outside. Right now, I have to put my attention again on joystick and reacquire it.

PR #2660, opened 2016-01-18
Virtual joystick improvments

DonLakeFlyer:
- Joystick position is bounded to rectangle
- No longer re-centers if thumb goes outside boundary. Stick stays at edge of boundary
- Thumb Pad moves to center itself at first thumb press location. This way as you put your thumb down for the first time it has no affect on joystick until you actually start dragging your thumb around

Fix for Issue #2308

@rjehangir This should be much better I think


rjehangir:
Don, nice. I think these are great changes for the virtual sticks. I will test this out tomorrow on our vehicle.

PR #2672, opened 2016-01-18
Apm stuff v2

tcanabrava:
A cleaned version of the staled branch: 
- Don't check for null pointers after a new, they won't be null, ever.
- use QStringLiteral / QLatin1Literal when they should be used
- fix some const correctness
- don't create virtual methods on overrides, use 'override' or 'final' keywords when needed.


DonLakeFlyer:
Great, thanks.

PR #2673, opened 2016-01-19
Massive fixes from Clazy

tcanabrava:
Clazy is a Static code analisys tool that knows Qt internals
and issues warnings when misused. This patch is the work of
me fixing tons of clazy warnings regarding only to QStringLiterals

More patches to come.
 
Tested both Android and Linux Desktop builds (debug/release) and tested SITL simulator.


DonLakeFlyer:
The analyzer runs are interesting. But the fact is that I don't trust them to not break something. I'd much rather do things like this right after the 3.0 release goes out the door (next month). Then it gets a lot of bake time in Daily Builds. It's too close to a Stable release to accept such a big change.

tcanabrava:
Sure, no worries. I'll continue fixing the analyzer issues (I'm sure that they don't break the build on the devices I'm testing) and updating this branch, when 3.0 hits it will be rebased against master and safe to merge.

PR #2675, opened 2016-01-20
Adding full path to windeployqt.

dogmaphobic:
When you generate a Visual Studio project, the build fails as it doesn't not know where to find windeployqt.


dogmaphobic:
Just to force a rebuild. Seems like CI was having a bad hair day.

PR #2680, opened 2016-01-21
deploy release and daily build pdbs

### LINKED ISSUE #2679: Need to capture pdbs for Windows daily and stable builds
QGC Windows tends to crash and hang more than other OS versions. I have instructions for collecting crash and hang dumps here: http://www.qgroundcontrol.org/bugreporting. But without a PDB it's not very useful. This is why I used to have the PDBs installed with QGC. Setting  up PDBs in such a way that they could be in a symbol server would be fantastic. But just a way to download the right one that matches the build is good enough to start. The other need is to be able to track back from a daily build commit, to the pdb for it as well as the source commit so everything matches up. Then you have real debugging for a user problem.

dagar:
closes #2679 

- [x] release builds
- [x] daily builds
- [x] daily build rotate
- [ ] developer howto

Questions
- how to rotate daily build symbols
  - store by YYYY-MM-DD/buildId and delete after 1 month?
  - or buildId % 100 and let s3 upload keep overwriting

- estimate s3 storage requirments
- ec2 micro instance for symbol server

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -38,6 +38,7 @@ artifacts:
   - path: qgroundcontrol-installer.exe
     name: qgcinstaller
 
+# TODO: dagar deploy pdbs
 deploy:
   - provider: S3
     access_key_id:
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -30,6 +30,14 @@ build_script:
   - mkdir %SHADOW_BUILD_DIR% && cd %SHADOW_BUILD_DIR% && C:\Qt\5.5\msvc2013\bin\qmake -r CONFIG-=debug_and_release CONFIG+=%CONFIG% CONFIG+=WarningsAsErrorsOn %APPVEYOR_BUILD_FOLDER%\qgroundcontrol.pro
   - cd %SHADOW_BUILD_DIR% && jom
   - if "%CONFIG%" EQU "installer" ( copy %SHADOW_BUILD_DIR%\release\qgroundcontrol-installer.exe %APPVEYOR_BUILD_FOLDER%\qgroundcontrol-installer.exe )
+# Generate the source server information to embed in the PDB
+  - '"C:\Program Files (x86)\Windows Kits\8.1\Debuggers\x64\srcsrv\srctool.exe" -r -u "%SHADOW_BUILD_DIR%\release\qgroundcontrol.pdb" | grep qgroundcontrol | grep -v moc_ | grep -v libs\\mavlink | grep -v build_windows_install | python %APPVEYOR_BUILD_FOLDER%\deploy\genPDBsrcsrv.py > %SHADOW_BUILD_DIR%\release\qgroundcontrol.pdb.srcsrv'
+# write the source server info
+  - '"C:\Program Files (x86)\Windows Kits\8.1\Debuggers\x64\srcsrv\pdbstr.exe" -w -i:%SHADOW_BUILD_DIR%\release\qgroundcontrol.pdb.srcsrv -p:%SHADOW_BUILD_DIR%\release\qgroundcontrol.pdb -s:srcsrv'
+# build the symbol / PE store
+  - 'cd %APPVEYOR_BUILD_FOLDER% && "C:\Program Files (x86)\Windows Kits\8.1\Debuggers\x64\symstore.exe" add /compress /s symbols /f %SHADOW_BUILD_DIR%\release\qgroundcontrol.pdb /t qgroundcontrol'
+  - 'cd %APPVEYOR_BUILD_FOLDER% && "C:\Program Files (x86)\Windows Kits\8.1\Debuggers\x64\symstore.exe" add /compress /s symbols /f %SHADOW_BUILD_DIR%\release\qgroundcontrol.exe /t qgroundcontrol'
+  - ps: 'cd $env:appveyor_build_folder; .\deploy\lowercaseify_symbolstore.ps1'
 
 test_script:
   - if "%CONFIG%" EQU "debug" ( %SHADOW_BUILD_DIR%\debug\qgroundcontrol --unittest )
@@ -38,7 +46,9 @@ artifacts:
   - path: qgroundcontrol-installer.exe
     name: qgcinstaller
 
-# TODO: dagar deploy pdbs
+  - path: symbols\**\*.*_
+    name: symbols
+
 deploy:
   - provider: S3
     access_key_id:
@@ -49,6 +59,17 @@ deploy:
     set_public: true
     folder: $(APPVEYOR_REPO_BRANCH)
     artifact: qgcinstaller
+    on:
+      CONFIG: installer
+  - provider: S3
+    access_key_id:
+      secure: IGAojLMqokL+76DbdulmWDA3MTsxEBBi3ReVVSqTy9c=
+    secret_access_key:
+      secure: RiYqaR+3T2PMNz2j5ur8LCA6H/Zfd4jTX33CZE5iBxm+zaz4QLs25p0B7prpaoNN
+    bucket: qgroundcontrol
+    set_public: true
+    folder: debugging
+    artifact: symbols
     on:
       CONFIG: installer
   - provider: GitHub
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -68,10 +68,23 @@ deploy:
       secure: RiYqaR+3T2PMNz2j5ur8LCA6H/Zfd4jTX33CZE5iBxm+zaz4QLs25p0B7prpaoNN
     bucket: qgroundcontrol
     set_public: true
-    folder: debugging
+    folder: cisyms
     artifact: symbols
     on:
       CONFIG: installer
+      appveyor_repo_tag: false
+  - provider: S3
+    access_key_id:
+      secure: IGAojLMqokL+76DbdulmWDA3MTsxEBBi3ReVVSqTy9c=
+    secret_access_key:
+      secure: RiYqaR+3T2PMNz2j5ur8LCA6H/Zfd4jTX33CZE5iBxm+zaz4QLs25p0B7prpaoNN
+    bucket: qgroundcontrol
+    set_public: true
+    folder: releasesyms
+    artifact: symbols
+    on:
+      CONFIG: installer
+      appveyor_repo_tag: true
   - provider: GitHub
     auth_token:
       secure: dzWLqd0eTKjrAWC5LgqVnwLemmrNdddGA2ZZn/OthAP37mwCLkP2C1zil7ivmEE8
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -199,7 +199,9 @@ WindowsBuild {
 ReleaseBuild {
     DEFINES += QT_NO_DEBUG
     CONFIG += force_debug_info  # Enable debugging symbols on release builds
-    CONFIG += ltcg              # Turn on link time code generation
+    !iOSBuild {
+        CONFIG += ltcg              # Turn on link time code generation
+    }
 
     WindowsBuild {
         # Enable function level linking and enhanced optimized debugging
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -178,11 +178,15 @@ MacBuild | LinuxBuild {
 }
 
 WindowsBuild {
+    QMAKE_CFLAGS_RELEASE -= -Zc:strictStrings
+    QMAKE_CFLAGS_RELEASE_WITH_DEBUGINFO -= -Zc:strictStrings
+    QMAKE_CXXFLAGS_RELEASE -= -Zc:strictStrings
+    QMAKE_CXXFLAGS_RELEASE_WITH_DEBUGINFO -= -Zc:strictStrings
     QMAKE_CXXFLAGS_WARN_ON += /W3 \
         /wd4996 \   # silence warnings about deprecated strcpy and whatnot
         /wd4005 \   # silence warnings about macro redefinition
-        /wd4290 \   # ignore exception specifications
-        /Zc:strictStrings-  # work around win 8.1 sdk sapi.h problem
+        /wd4290     # ignore exception specifications
+
     WarningsAsErrorsOn {
         QMAKE_CXXFLAGS_WARN_ON += /WX
     }
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -198,14 +198,19 @@ WindowsBuild {
 
 ReleaseBuild {
     DEFINES += QT_NO_DEBUG
-    WindowsBuild {
-        # Use link time code generation for better optimization (I believe this is supported in MSVC Express, but not 100% sure)
-        QMAKE_LFLAGS_LTCG = /LTCG
-        QMAKE_CFLAGS_LTCG = -GL
+    CONFIG += force_debug_info  # Enable debugging symbols on release builds
+    CONFIG += ltcg              # Turn on link time code generation
 
-        # Turn on debugging information so we can collect good crash dumps from release builds
-        QMAKE_CXXFLAGS_RELEASE += /Zi
-        QMAKE_LFLAGS_RELEASE += /DEBUG
+    WindowsBuild {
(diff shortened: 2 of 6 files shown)


NaterGator:
If hosting symbols on S3 we'll need something to make S3 behave like a symbol server MS expects: https://github.com/aroben/pepto-symbol

DonLakeFlyer:
So we have a user running a daily build, they will have a commit number in the title bar which tells me which build they are running. Can I connect back from that commit number to that actual .exe and .pdb in S3?

NaterGator:
I would suggest using a crash handler; [breakpad](https://chromium.googlesource.com/breakpad/breakpad/) is a decent choice. The user would submit a crash dump (minidump hopefully, full dump if necessary) which you load into Qt Creator / Visual Studio debugger with the symbol server added to the symbol path.

CDB will resolve the GUID of the compiled executable ([source](http://www.wintellect.com/devcenter/jrobbins/pdb-files-what-every-developer-must-know)) via the crash dump and send it to the symbol server which will then search for the corresponding PDB and load it for you. 

You should not need to embed any specific build information in the application as the VC compiler does this for you. (Though that will likely still be useful for ascertaining exactly what revision a particular crashdump belongs to. In that case I have used a solution akin to [this](http://www.zachburlingame.com/2011/02/versioning-a-native-cc-binary-with-visual-studio/) with cmake generating the necessary compiler flags to setup the defines appropriately. This fills in the application information in the properties shell dialog).

DonLakeFlyer:
Make sense. Do you think you two guys have time/experience to get this working end to end? Would be a huge help.

NaterGator:
Sure. It looks like (this)[http://www.symbolsource.org/Public] may also be an option depending on how we can integrate Appveyor.

dagar:
We're looking at creating a nuget package which would give us easy symbol server appveyor support and provide an alternate windows installation method that could auto update.

DonLakeFlyer:
Nice, thanks

NaterGator:
Right, so ... getting there.

The installer for this branch [here](https://s3-us-west-2.amazonaws.com/qgroundcontrol/pdb/qgroundcontrol-installer.exe) does not include the pdb. If you experience a crash with this build and generate a minidump, you _should_ be able to run pepto-symbol pointed at the S3 symbol repository here: https://s3-us-west-2.amazonaws.com/qgroundcontrol/debugging/symbols/ 

With the pepto-symbol proxy added to your [symbol path](https://msdn.microsoft.com/en-us/library/ms241613.aspx) you would be able to load the minidump in the visual studio debugger and it _should_:
- automatically download the matching executable
- automatically download the correct pdb
- automatically populate source files directly from raw github URLs 


Probably a few more kinks to iron out, and we may well want to get a machine reserved for running the pepto-symbol proxy. Getting there though; this should make troubleshooting crashes from users a bit less painful.

NaterGator:
Well, it does indeed work!

image

That's all for today, I'll polish this up a bit more tomorrow.

NaterGator:
Right, instructions for folks who want to use this.

## Run Your Own Symbol Proxy
Owing to S3 urls being case sensitive and IIS being case insensitive, you need to run a proxy which adjusts all symbol requests from the debugger to lowercase. Clone [pepto-symbol](https://github.com/aroben/pepto-symbol), download a Windows binary of [forego](https://github.com/ddollar/forego) (Go rewrite of foreman), and install [node.js](https://nodejs.org/en/download/). 

Open a command prompt and run: 
`set S3_BUCKET=qgroundcontrol`
`forego start`
(You may have to run an administrator command prompt, approve firewall rules, etc) 

With the proxy server running on your local machine (default port is 5000, change via `set PORT=whatever`) you need to add http://localhost:5000/debugging/symbols to your [symbol path](https://msdn.microsoft.com/en-us/library/ms241613(v=vs.120).aspx#BKMK_Specify_symbol_locations_and_loading_behavior). It should look like this when you are done:
image


## Enable Source Code Resolution
You need to enable source server support in Visual Studio. Instructions to do so are [here](https://msdn.microsoft.com/en-us/library/ms241613(v=vs.120).aspx#BKMK_Use_source_servers). Yes, that's it... if you step through or get a dump with version control annotated PDB it will populate automatically from github. 


With those two steps done you can do your usual Visual Studio debugging... load a core dump, attach to a running process, start the process from the debugger, etc.

NaterGator:
@LorenzMeier & @dogmaphobic where do we want to go from here? I don't have much visibility into the financial impact of this; the combined size of the compressed PDB and exe stored in the S3 bucket is about 9MB per build. According to my rough math that means it is about $12 to store 3800 builds for a year. Maybe we only store PDBs when we're building master, not for every PR/branch? Maybe we cull everything older than 1 year? Maybe it is just lost in the noise?

Users can run their own pepto proxy or we can spin it up on server somewhere. 

We can leave the existing guidance for crash dumps as-is (let windows handle it, ask users to submit them to mailing list) or we can implement a crash handler and reporting system. For now having users submit them may be a good idea until we build some confidence in how to store and triage windows reports. 

Finally: Qt does not currently ship any debug information with their release DLLs. There is some hope that 5.6 (or future) versions may supply release PDBs ([source](https://bugreports.qt.io/browse/QTBUG-29668)) but if it becomes an issue we may want to consider rolling our own infrequent build of Qt and hosting the release PDBs.

LorenzMeier:
@NaterGator @dagar I'd like to just set up up and ignore costs for now. We're paying already about $100 / month for AWS total. I'll try to set up an AWS account for you over the weekend so you can get going yourself.

NaterGator:
@LorenzMeier that sounds good, if that is the case this should definitely get lost in the noise.

Do you mind granting me write access to the repo so I can stop throwing things at @dagar every time I want appveyor to build this branch? I think he's getting tired of it... :unamused:

DonLakeFlyer:
> Do you mind granting me write access to the repo so I can stop throwing things

I'm not quite sure how I do that. Hopefully @LorenzMeier can set it up.

dagar:
What exactly do we need to finish this off? Just hosting?

NaterGator:
Right now it is going into the QGroundControl S3 bucket. If Lorenz is okay with that it is ready to merge.

dagar:
Could we separate releases (tags) and all other builds? That way if the storage ever gets out of control we can just wipe the continuous build symbols and keep the releases forever.

DonLakeFlyer:
Also use the buildid rotation scheme for daily builds so they don't pill up?

NaterGator:
@DonLakeFlyer are the daily builds already rotated? We could just include the PDB files with the daily build installer.

DonLakeFlyer:
There is just a single daily build. But it would be great to have more than one pdb for daily builds. So if someone is running a slightly older one we can still debug to some extent. Is that possible? How does an .exe connect to the correct .pdb through the symbol server.

NaterGator:
That is getting more complicated since Appveyor is managing uploading the artifacts to S3. What likely makes the most sense and is the least painful is to have Lorenz configure lifecycle policies for the respective build types. My S3-fu is a bit weak but I believe that means daily build symbols and release build symbols will have to live in different folders in the bucket, and consequently visual studio will have to be configured with separate symbol paths for daily vs. release builds.

DonLakeFlyer:
Just upload the latest pdb then. Don't worry about keeping more than the last one. We can always ask someone to update to latest daily first. Then try again for the crash.

NaterGator:
That may be a little tricky since the PDB is generated with a unique hash each time and I don't think Appveyor has any API for removing objects from S3. That is why I am advocating using S3's [lifecycle configuration](http://docs.aws.amazon.com/AmazonS3/latest/UG/lifecycle-configuration-bucket-no-versioning.html) to enforce removing older build debugging information.

DonLakeFlyer:
Quick look at the the lifecycle thing would seem to make sense to use.

NaterGator:
@LorenzMeier I think this is waiting on you to confirm you're okay with and have flipped the switch on S3 lifecycle configuration for release vs CI builds.

dagar:
I'll merge this and setup S3 lifecycle management for qgroundcontrol/cisyms

dagar:
I've configured S3 to delete continuous build symbols older than 7 days. Once I verify it's working I'll increase it to a few months.

QGCCommon.pri
```diff
@@ -194,14 +198,19 @@ WindowsBuild {
 
 ReleaseBuild {
     DEFINES += QT_NO_DEBUG
-    WindowsBuild {
-        # Use link time code generation for better optimization (I believe this is supported in MSVC Express, but not 100% sure)
-        QMAKE_LFLAGS_LTCG = /LTCG
-        QMAKE_CFLAGS_LTCG = -GL
+    CONFIG += force_debug_info  # Enable debugging symbols on release builds
+    CONFIG += ltcg              # Turn on link time code generation
```
dagar: @NaterGator the iOS build doesn't like this

NaterGator: The error suggests this fix. You can apply or I will when I get out of this meeting.

.appveyor.yml
```diff
@@ -38,6 +46,9 @@ artifacts:
   - path: qgroundcontrol-installer.exe
     name: qgcinstaller
 
+  - path: symbols\**\*.*_
```
jaxxzer: @NaterGator why is the pdb file called `qgroundcontrol.pd_` and the exe `qgroundcontrol.ex_`?

DonLakeFlyer: Because they are compressed for use with the Pepto symbol server (which I'm not really sure how to set up).

DonLakeFlyer: Seems like it might be easier to store the artifacts uncompressed and then we could just pull individual versions from appveyor as needed?

jaxxzer:
@NaterGator, I get an access denied error when I visit localhost:5000/debugging/symbols and localhost:5000. Any ideas?

capture

Also, shouldn't the address be localhost:5000/cisyms or localhost:5000/releasesyms according to the folders specified in the deployment? https://github.com/mavlink/qgroundcontrol/pull/2680/files#diff-11c909939117928998b102a1fff7d363R71

NaterGator:
Sorry guys, I have been pretty out of the loop. I have an EC2 instance I can set this up on tonight.

jaxxzer:
@NaterGator can you please help me understand how to make use of the .dump file created per these instructions. http://qgroundcontrol.org/bugreporting#reporting_crashes_from_windows_builds

NaterGator:
@jaxxzer you should be able to load the dmp file in visual studio. You are right we separated the symbols into two buckets, cisyms which are cleared periodically, and releasesyms which persist longer. Where you got the build will determine which symbol path to use when configuring Visual Studio (or alternatively add both and the wrong one will 404).

NaterGator:
@jaxxzer can you try adding http://qgc.aws.nate-online.com/releasesyms/symbols/ and http://qgc.aws.nate-online.com/cisyms/symbols/ to your symbol search path? This proxy server will check for symbols on the S3 bucket and send an HTTP 301 redirect directly to S3 if they are found.

jaxxzer:
@NaterGator I will try that. In the meantime...

I've got an installer here:
https://s3.amazonaws.com/downloads.bluerobotics.com/QGC/builds/intentional-crash/QGroundControl-installer.exe

And a corresponding (compressed) pdb here:
https://s3.amazonaws.com/downloads.bluerobotics.com/QGC/cisyms/symbols/qgroundcontrol.pdb/b7d5bb172249456e8874af2cdad283161/qgroundcontrol.pd_

I've added `https://s3.amazonaws.com/downloads.bluerobotics.com/QGC/cisyms/symbols` to my search path. I get this result:
`SYMSRV:  https://s3.amazonaws.com/downloads.bluerobotics.com/QGC/cisyms/symbols/QGroundControl.pdb/B7D5BB172249456E8874AF2CDAD283161/QGroundControl.pdb not found`

The casing and file name (.pdb instead of .pd_) of the search are obviously not correct. How do I resolve the correct path, and decompress the file?

NaterGator:
@jaxxzer real quick, for builds out-of-tree from the primary qgc build server my AWS links will not work (they specifically proxy the CI builds from _this_ repository only).

In your case the path where the symbols are being placed is logged [here](https://ci.appveyor.com/project/blueroboticsappveyor/qgroundcontrol/build/intentional-crash-191#L2105). The reason why you're seeing the behavior that you are is because of [this nuance](https://github.com/aroben/pepto-symbol/blob/master/README.md#that-doesnt-sound-so-hard-why-do-i-need-this) of IIS vs S3. The build scripts convert everything to lowercase before uploading to S3 and then the proxy server (local pepto-symbol for private symbol buckets, my aws link for in-tree builds) takes any incoming requests from symsrv.exe and converts them to lowercase before checking S3. In the case of my EC2 box though, I don't actually serve them from the EC2 instance, I only generate a HEAD request and check for a HTTP 200, if I get it I return an HTTP 301 with the correct S3 URL. 

In either case if the link is directly fed to S3 without fixing the case and S3 returns HTTP 403, symsrv blacklists that server and stops searching. In the normal case of HTTP 404 it will continue by looking for the compressed versions of the symbols:

`Oct 18 15:03:52 ip-172-31-20-52 systemd[1]: Started symproxy-web-1.service.
Oct 18 15:04:53 ip-172-31-20-52 symproxy-web-1.service[4756]: Proxy request to /releasesyms/symbols/QGroundControl.pdb/DF04654ED602473898535949CCD09A811/QGroundControl.pdb
Oct 18 15:04:54 ip-172-31-20-52 symproxy-web-1.service[4756]: Proxy request to /releasesyms/symbols/QGroundControl.pdb/DF04654ED602473898535949CCD09A811/QGroundControl.pd_
Oct 18 15:06:49 ip-172-31-20-52 symproxy-web-1.service[4756]: Proxy request to /releasesyms/symbols/qdds.pdb/8F19E1FADADD44EEA4085D78021EBE324/qdds.pdb
Oct 18 15:06:50 ip-172-31-20-52 symproxy-web-1.service[4756]: Proxy request to /releasesyms/symbols/qdds.pdb/8F19E1FADADD44EEA4085D78021EBE324/qdds.pd_
Oct 18 15:06:50 ip-172-31-20-52 symproxy-web-1.service[4756]: Proxy request to /releasesyms/symbols/qdds.pdb/8F19E1FADADD44EEA4085D78021EBE324/file.ptr
Oct 18 15:06:51 ip-172-31-20-52 symproxy-web-1.service[4756]: Proxy request to /cisyms/symbols/qdds.pdb/8F19E1FADADD44EEA4085D78021EBE324/qdds.pdb
Oct 18 15:06:51 ip-172-31-20-52 symproxy-web-1.service[4756]: Proxy request to /cisyms/symbols/qdds.pdb/8F19E1FADADD44EEA4085D78021EBE324/qdds.pd_
Oct 18 15:06:52 ip-172-31-20-52 symproxy-web-1.service[4756]: Proxy request to /cisyms/symbols/qdds.pdb/8F19E1FADADD44EEA4085D78021EBE324/file.ptr
`

jaxxzer:
> real quick, for builds out-of-tree from the primary qgc build server my AWS links will not work (they specifically proxy the CI builds from this repository only).

I understand this. Still digesting the rest of it.

NaterGator:
Also note when we switch over to builds utilizing Qt 5.9 the PDB files for release versions of Qt are also [available](https://download.qt.io/official_releases/qt/5.9/) and will greatly help in stack unwinding and generating comprehensible debugging output. It will be up to the developer investigating the crash report to have the correct version of those PDBs unzipped and in their symbol path for debugging.

jaxxzer:
I've had some success finally. Thank you for elaborating on the pepto-symbol bit. I had to change the folder on our S3 bucket QGC/->qgc/.

jaxxzer:
Hi @NaterGator, I've got to debug a windows dump again and I'm trying to get going again with a local proxy because it is not an in-tree build. It looks like you edited your excellent instructions about how to set up locally with pepto-symbol and forego. I'm having the following issue when I try to access the proxy (eg. just typing localhost:5000 in a browser). I think I'm forgetting a step.
capture

jaxxzer:
NVM, nailed it:
`S3_BUCKET=downloads.bluerobotics.com`

PR #2682, opened 2016-01-21
bitwise and, not boolean and, to check for bitmasks.

tcanabrava:
Seems that a typo is preventing a bitwise operation.


DonLakeFlyer:
@dogmaphobic This woul dhave been screwing up multi channel sequencing.

Issue #2678, opened 2016-01-20
Flash dialog is super confusing


LorenzMeier:
@DonLakeFlyer @dogmaphobic What QGC is trying to say is here is:

"UNPLUG THE USB PLUG TO START FIRMWARE UPGRADING"

I think that is what it should say. In the center. BIG. Right now users think "it doesn't work".

![screen shot 2016-01-21 at 9 16 43 am](https://cloud.githubusercontent.com/assets/1208119/12465111/c572e244-bfcc-11e5-8233-1cc376f0fa21.png)

Then on to the 2nd issue: If you unplug, the UI resets and moves back to the overview. If you now re-plug without toggling back to Firmware upgrade you're back to square 1. Super annoying.

I have quite a few user reports on this. We need to fix it this time. I've raised it for a very long time now.

  - [ ] Fix flash instructions. Be SHORT
  - [ ] Do not reset UI but stay in Firmware upgrade view

DonLakeFlyer:
Tell me exactly word for word what you want it to say and I'll change it!

LorenzMeier:
"UNPLUG THE AUTOPILOT USB CABLE TO START FIRMWARE UPGRADING"

In center. no other text. In a big font.

DonLakeFlyer:
What about Radio?

LorenzMeier:
Right. Change it to board or something.

DonLakeFlyer:
> I've raised it for a very long time now.

This is the first I've heard of the UI reset issue. Althought now that I know about it I can certainly see it as completetly screwed up. If things just fly by in Gitter I lose them. A detailed GitHub Issue and I won't miss things.

PR #2684, opened 2016-01-21
Firmware upgrade usability

DonLakeFlyer:
Fix for Issue #2678 

screen shot 2016-01-21 at 10 08 52 am


LorenzMeier:
It looks better, but can we center the text more and make it a lot larger? On a large screen it's still in the top left corner.

DonLakeFlyer:
> can we center the text more

This screen doesn't work that way. This is a scrolling text field that has text appended to it. In 3.1 I'll put a real UI on it instead of this logging style ui. That is the real fix. I think any larger is just going to drown out the other information in there which is appended after this text telling you to plug in the board an so forth. You'll likely just move the problem to a different place.

PR #2693, opened 2016-01-22
Add a Vagrantfile

peterbarker:
A "vagrant up" on a host machine of Ubuntu 15.10 will provide binaries and a Qt environment in shadow-build/release which can be run on the host machine.


DonLakeFlyer:
@peterbarker Do you think you could update the README on how to use this for building?

DonLakeFlyer:
Thanks Peter

Issue #2560, opened 2016-01-09
QGC default map location


lvale:
I believe that the default location shown on the map could be saved on the last location received by a vehicle instead of just jumping back to SF.

Example: I disconnected the vehicle while in Europe, and as soon I disconnected the location map just jumped back to the pic.....

Also, some display of coordinates would be useful. Perhaps a crosshair in the middle of the view, and the location would be of the crosshair position.

<img width="1440" alt="screen shot 2016-01-09 at 13 47 28" src="https://cloud.githubusercontent.com/assets/3130363/12216410/253a70f0-b6d8-11e5-9f34-3a768cecff0a.png">

lvale:
@DonLakeFlyer 
Sorted, thanks.
But :)

Vehicle had GPS lock and then was disconnected and this screen is correct, i.e. remained at the last known GPS Pos.

<img width="1436" alt="screen shot 2016-01-15 at 01 01 27" src="https://cloud.githubusercontent.com/assets/3130363/12342606/dbb00e0e-bb23-11e5-92d0-9d92455f6ada.png">


but if I move to the planning tab it goes back to SF

Shouldn't it inherit the GPS pos ??

<img width="1438" alt="screen shot 2016-01-15 at 01 01 38" src="https://cloud.githubusercontent.com/assets/3130363/12342615/ffdef04c-bb23-11e5-93cf-b6b586f4d10e.png">

DonLakeFlyer:
I can make it do that.

DonLakeFlyer:
It's doesn't inherit the GPS position, because that ui is only created when you click the tab. If you never clicked the tab, then it never saw the gps position go by because it wasn't running.

lvale:
I believe it makes sense to "inherit" the last known good position, in this case from the Fly tab.

DonLakeFlyer:
This is going to require some restructuring of the Fly View. The Map and Video items need to be persistent and just switch size and z-order. Otherwise you can't depend on the Fly view being around. By re-loading the items, the Fly view also also loses state like position I think, and re-creating all the mission and flight lines is expensive on the switch.

Other option is to leave things along and just push the map position out of the inner Map control up to the parent. This is doable as well. Likely easier, but still not sure if it's better the re-structure map/video setup.

@dogmaphobic Did you happen to try this structuring first and run into problems? Not sure yet if it's worth the effort to restructure map/video switching mechanism.

dogmaphobic:
[...]

PR #2701, opened 2016-01-22
Making Map and Video view persistent.

dogmaphobic:
They are no longer created/destroyed each time you toggle PIP.
This is response to #2560


DonLakeFlyer:
Pipable! New word.

dogmaphobic:
> Pipable! New word.

I knew you would enjoy it :) I didn't know what to call it and then I saw *Flickable* and thought, what the heck...

PR #2708, opened 2016-01-25
Raise min Qt version to 5.5+

LorenzMeier:


LorenzMeier:
Appveyor restarted. Looks like a CI failure, not a failure of the actual build.

DonLakeFlyer:
@dagar Does this one make any sense to you?

NaterGator:
@DonLakeFlyer if you are referring to the appveyor build failure, that is when sourceforge is down or doing maintenance and we can't install NSIS.

DonLakeFlyer:
I tried re-running it and it failed again. Hence I was confused since that usually works. Then close/reopen seemed to do the trick.

PR #2710, opened 2016-01-25
Fix iOS build.

dogmaphobic:


dagar:
Time for an iOS travis-ci build? I'm somewhat limited in what I can do here without access to a Mac, but I could help get most of the boiler plate in place.

dogmaphobic:
> Time for an iOS travis-ci build? I'm somewhat limited in what I can do here without access to a Mac, but I could help get most of the boiler plate in place.

I would wait until after 3.0 is out. There is no way iOS will be ready for 3.0 so we might as well save a zillion computers cycles building iOS each time a PR comes along. I can easily fix these things as they show up for the time being.

Thanks!

PR #2712, opened 2016-01-27
Fix acelerometer setup

tcanabrava:
some issues regarding APM setup on a local drone.


DonLakeFlyer:
I should have time to look at this sometime today

tcanabrava:
Thanks.

DonLakeFlyer:
You should just switch it to only check the offsets for the first accel. No need to leave the old stuff in an ifdef. Also no need for insUse variable at all or rgOffsets to have two dimensions any longer. Also APMSensorsComponent::setupCompleteChangedTriggerList should match the set of params that are now being used.

DonLakeFlyer:
@tcanabrava Can you clean this up?

tcanabrava:
seems broken because of a forced push.

PR #2718, opened 2016-01-28
Read currently installed USB serial drivers and skip reinstallation

NaterGator:
Nobody asked for this but it drives me crazy on windows so I figured I would address it.

This commit adds a registry scan pass to detect if the USB serial driver package included is already installed (it checks for an exact version match). If the drivers are unnecessary we skip executing the USB driver installer.


DonLakeFlyer:
Nice, thanks Nate.

PR #2730, opened 2016-02-01
Making the ESP8266 component "common" instead of APM/PX4 only.

dogmaphobic:
@DonLakeFlyer Please take a look at this and see if my assumption is correct.

APMComponent and PX4Component were identical. I removed them and made the various components derive from VehicleComponent directly (moving the few methods there along with it). The reason was for a *common* component, such as the ESP8266, you would have to create 2 identical components. One for APM and another for PX4 only because of the class hierarchy.

If at some point a need arises that would validate a need for a specialized component subclass, it can always be created. Everything in VehicleComponent is virtual.

This PR makes the *WiFi Bridge* component common to both APM and PX4.

With all that said, there is something weird going on with the parameters sent by the WiFi Bridge (Component 240). They are not loaded properly when connected to APM (PX4 works fine as before).

screen shot 2016-02-01 at 3 35 33 pm


dogmaphobic:
Whatever the problem is, it's within the WiFi Bridge firmware (I think). The parameters sent by the WiFi Bridge firmware should be, in theory, identical. This is the raw data received by QGC when connected to a PX4:

```
"SW_VER"            01000002
"DEBUG_ENABLED"     00000000
"WIFI_CHANNEL"      0000000B
"WIFI_UDP_HPORT"    000038D6
"WIFI_UDP_CPORT"    000038DB
"WIFI_SSID1"        52786950
"WIFI_SSID2"        72656361
"WIFI_SSID3"        00000000
"WIFI_SSID4"        00000000
"WIFI_PASSWORD1"    72786970
"WIFI_PASSWORD2"    72656361
"WIFI_PASSWORD3"    00000000
"WIFI_PASSWORD4"    00000000
"UART_BAUDRATE"     000E1000
```

And this is the same set when connected to APM:

```
"SW_VER"            00000000
"DEBUG_ENABLED"     58E7A600
"WIFI_CHANNEL"      00000000
"WIFI_UDP_HPORT"    58E70000
"WIFI_UDP_CPORT"    58E70000
"WIFI_SSID1"        1A540000
"WIFI_SSID2"        00000000
"WIFI_SSID3"        00000000
"WIFI_SSID4"        00000000
"WIFI_PASSWORD1"    00000000
"WIFI_PASSWORD2"    00000000
"WIFI_PASSWORD3"    00000000
"WIFI_PASSWORD4"    00000000
"UART_BAUDRATE"     00000000
```

This is the result of:

```
mavlink_param_value_t rawValue;
mavlink_msg_param_value_decode(&message, &rawValue);
mavlink_param_union_t paramVal;
paramVal.param_float = rawValue.param_value;
char foo[16];
sprintf(foo,"%08X", paramVal.param_uint32);
qDebug() << parameterName << " " << foo;
``` 

Something is messed up in the packing and/or unpacking of ```mavlink_param_value_t```.

DonLakeFlyer:
Only a few seconds to reply. Parameters are sent differently in px4 and apm stack. See the mavlink message handler in firmware plugin for the different encodings

PR #2733, opened 2016-02-04
Adding (optional) link status data to the WiFi Bridge panel.

dogmaphobic:
This is optional. You need to click the *Show Status* button to enable it. Data is collected from the WiFi bridge using dynamic json requests as well as directly from QGC's link. This way we can identify where the errors are occurring. In the case below, the errors are at the QGC's UDP receiving end.

*Reset Counters* allows you to synchronize the counters. Each new session of either end will start with 0s, making the end that continues to run to continue *counting*.

screen shot 2016-02-03 at 11 19 31 pm
P.S. The message counts don't match because the QGC data is in real time while the Bridge data comes in 1Hz Query/Response interval.


dogmaphobic:
Forcing rebuild. Linux failed to pull SDL.

PR #2736, opened 2016-02-04
General cleanup

dogmaphobic:
Cleaning up noise and dead code.

I was about to create a new "Custom QML Command" and in the process, decided to clean up a zillion spurious errors, qDebug() calls, etc.

There is a fundamental design issue with the way the *Facts* disappear before the QML file accessing them goes away. This causes a flood of errors as these files try to access these *Facts*. I cleaned a bunch of them but the real fix is to have it so the QML code is down before deleting the underlying controllers and the Fact system.

Things like:
```
qrc:/qml/SafetyComponent.qml:179: TypeError: Cannot read property 'value' of null
qrc:/qml/SafetyComponent.qml:311: TypeError: Cannot read property 'value' of null
qrc:/qml/SafetyComponent.qml:297: TypeError: Cannot read property 'value' of null
```

Widgets load their saved states on boot and save them in their destruction. This is not working. More often than not, something happens that causes the entire application to crash on exit. The call stack is a mile long and I could not find the real source. I've disabled the save/restore until this is found and fixed.

These two calls are disabled:

```
void QGCDockWidget::loadSettings(void)
void QGCDockWidget::saveSettings(void)
```

All links keep track of data rate. At the moment, nobody uses the accumulated data. As this takes time, I've added a method to ```LinkInterface``` to enable/disable data rate collection. It defaults to *false*.

This was an attempt at fighting the data loss that insists on happening with the UDP link.

Reformatted the *CustomCommandWidget* as I'm preparing to create a few for testing various aspects of data flow within QGC.


DonLakeFlyer:
Can you make sure to put in issue in for anything disabled that still needs to be looked at?

NaterGator:
Is the UDP data loss still a problem?

dogmaphobic:
> Can you make sure to put in issue in for anything disabled that still needs to be looked at?

As to the Fact disappearing before the QML files consuming them, we talked about it before. I'm not even sure it's worth the effort given the magnitude of the change. It's just annoying and I clean up (by testing the Fact property) as I bump into them.

The crash on exit, I was almost certain I had already put an issue about it. We talked about that too (you wished me *good luck with that* at the time :) I may have only thought I had. I will look...

dogmaphobic:
> Is the UDP data loss still a problem?

Yes, it is quite a bother. The easiest way to *watch* it happening is downloading a large log file. You can see the download count steady and every now and then it *pauses*. Those are lost packets. It is as if the reads are stalled every now and then (losing packets in the process). I tried everything I could think of. My next test will be writing a full UDP stack from scratch using low level functions (bypassing Qt) just to see if it is related to Qt. I actually started that a couple of times and gave up because the amount of work to deal with platform idiosyncrasies. It happens with any UDP connection. Not just with the ESP8266 bridge.

dogmaphobic:
Oh, and by they way, as I mentioned on gitter, we're up for a boat load of fun when we switch to Qt 5.6. The entire Fact system breaks for some reason. I didn't spend any time looking into it. I was just curious to see how badly the build would be broken.

NaterGator:
Ok. I have concerns about the threading design patterns and think that may be part of the issue. As we know Qt got the threading API badly wrong and it is not at all intuitive to do it properly. I've fought this before so I'll have a look tomorrow and see if I can turn some stones. If this was happening on only one platform I would blame an esoteric OS issue but that doesn't seem to be the case. My best guess is that the socket event loop is running in the UI thread and dropping incoming packets during something like the qml garbage collection lock.

dogmaphobic:
> My best guess is that the socket event loop is running in the UI thread and dropping incoming packets during something like the qml garbage collection lock.

Note that I'm not using the socket signals. That got broken in Qt 5.5 and they haven't fixed yet. You get one bytesReady() call and never again.

NaterGator:
Wow, that's a pretty huge bug! Is it in the Qt bug tracker??

dogmaphobic:
Yep, it's been there since 5.5.0 came out. I haven't kept track of it though. If I remember correctly, it was a Window issue.

dogmaphobic:
Looks like it's more than one problem. The one I had seen is marked as closed:
https://bugreports.qt.io/browse/QTBUG-46552

dogmaphobic:
My comment on that bug back in June 2014:
*No, as far as I can tell, 5.4.1 is not affected. It started with 5.4.2. Also note that 5.5.0 inherited the same problem.*

NaterGator:
Thanks, I just came across the same bug report. I don't have UDP enabled hardware yet but I can get something off the ground... if I have to I'll simulate a UDP client by writing a quick UDP <> serial proxy app. I'm sure we can get to the bottom of this.

dogmaphobic:
You can use MavProxy or SITL.

laurenweinstein1:
Catching up here, I assume there's no longer any need for me to collect diags from my testing of the ESP bridge? Noting that I'm seeing similar loss in both QGC and Tower. I compiled the diag QGC, but was unable to see the "show status" button in the wifi bridge section.

dogmaphobic:
These are separate issues. The dropping UDP packets is an old one and it happens with any UDP connection (not just with the ESP). These drops are consistent but relatively small (20 packets in 1000). The issues you are having are unique to your setup. The *Show Status* was merged "19 hours ago" (according to the PR). I don't know when the daily build is run.
https://github.com/mavlink/qgroundcontrol/pull/2733

laurenweinstein1:
OK, I'll check tomorrow. Thanks.

DonLakeFlyer:
> There is a fundamental design issue with the way the Facts disappear before the QML file accessing them goes away. This causes a flood of errors as these files try to access these Facts. I cleaned a bunch of them

This should only happen during shutdown. The MultiVehicleManager has a phased deletion and vehicle change mechanism which should prevent it from happening during normal operation. Are you seeing it somewhere other than shutdown. At shutdown it's not really worth trying to get rid of the warnings since they don't actually cause a real problem. It can "possibly" be fixed correctly, but would take a lof of effort in changing the object ownership when going between C++ and Qml.

Issue #2735, opened 2016-02-04
Ardupilot firmware flashing from QGC master disabled


lvale:
Hi. Any reason why flashing with QGC of Ardupilot firmware has been disabled ?? Just followed all the steps with Wireshark active and I don’t see any attempt of QGC to query any servers, while trying to flash PX4 runs ok from amazonaws ?
running latest qgc from master on OS X

https://youtu.be/2LvQlmq12Tc

dogmaphobic:
You were trying to load APM Quad V1. I don't see any *V1* for APM in the list:
https://github.com/mavlink/qgroundcontrol/blob/master/src/VehicleSetup/FirmwareUpgradeController.cc#L206-L282

PR #2737, opened 2016-02-04
APM file locations revised

lvale:
To address issue https://github.com/mavlink/qgroundcontrol/issues/2735

### DIFF
--- a/src/VehicleSetup/FirmwareUpgradeController.cc
+++ b/src/VehicleSetup/FirmwareUpgradeController.cc
@@ -208,8 +208,8 @@ void FirmwareUpgradeController::_initFirmwareHash()
         { AutoPilotStackPX4, StableFirmware,    DefaultVehicleFirmware, "http://px4-travis.s3.amazonaws.com/Firmware/stable/px4fmu-v4_default.px4"},
         { AutoPilotStackPX4, BetaFirmware,      DefaultVehicleFirmware, "http://px4-travis.s3.amazonaws.com/Firmware/beta/px4fmu-v4_default.px4"},
         { AutoPilotStackPX4, DeveloperFirmware, DefaultVehicleFirmware, "http://px4-travis.s3.amazonaws.com/Firmware/master/px4fmu-v4_default.px4"},
-        { AutoPilotStackAPM, StableFirmware,    QuadFirmware,           "http://firmware.diydrones.com/Copter/stable/PX4-quad/ArduCopter-v4.px4"},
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Why are v4 listings commented out?

dogmaphobic:
Because they don't exist?
```
:~$ wget http://firmware.diydrones.com/Copter/stable/PX4-quad/ArduCopter-v4.px4
--2016-02-05 12:49:21--  http://firmware.diydrones.com/Copter/stable/PX4-quad/ArduCopter-v4.px4
Resolving firmware.diydrones.com... 54.235.19.222
Connecting to firmware.diydrones.com|54.235.19.222|:80... connected.
HTTP request sent, awaiting response... 404 Not Found
2016-02-05 12:49:21 ERROR 404: Not Found.
```

DonLakeFlyer:
If so they should be removed instead of commented out. Can't remember if Ui will work correctly with the other ones missing?

lvale:
@DonLakeFlyer I commented them because there is an effort to make PX4 versions for stable and beta in a short time.

All the locations have been verified, although I can't test the UI with V4, I did test with V1 and V2

billbonney:
@DonLakeFlyer They don't exist since they haven't been built yet. They will be the valid URLs when a beta release and a stable release are made. At the moment only 'latest' master build will download but its the bleeding edge releases. 
I say leave it as it is, maybe fix the warning message to say no stable/beat version available.

The problem with this link  https://github.com/diydrones/binary
https://github.com/diydrones/binary/blob/master/Firmware/firmware2.xml is that its updated when MO updates the build list, not when an ArduPilot release is made. it should be more formally done by the ardupilot team

lvale:
@billbonney I believe that is an issue we can raise at the next dev meeting ?

lvale:
Removed the commented lines for Beta and Release Quad for v4 boards, which don't exist yet.

billbonney:
@DonLakeFlyer Are you going to add these removed lines when the images become available?

lvale:
I plan to keep the locations updated, when changes occur at the ArduPilot side, if that's ok with you.

Oh, and until a better method gets devised ;)

billbonney:
That doesn't work as its a chicken and egg problem. What you are saying is that all users of APM pxv4 have to update to a daily build of QGC to get a stable build on PX4v4 APM installed. 

It just much better to have a message that says the stable beta build is not available yet. Removing the functionality doesn't make sense 

My advice is out the lines back and fix the warning message of failure to download

billbonney:
Ok, taking my own advice, I've created a pull request here that fixes the issue and indicates when a file is not found and corrective action. 
https://github.com/mavlink/qgroundcontrol/pull/2750
Hope that helps :)

PR #2738, opened 2016-02-05
Update NSIS installer to modern UI

NaterGator:
Another one nobody asked for! The old NSIS installer GUI looks really dated... so let's update to the V2 UI. 

The bitmap looks a bit rough because I just did a cubic rescale in gimp. If there is a vector resource somewhere I can use to create a proper header I'll do so. 

image


LorenzMeier:
I don't think I still have the logo file. Created this almost a decade ago! However, Intel promised to do a new logo, so lets see what they come up with and lets merge this in the mean time.

DonLakeFlyer:
Ask @dogmaphobic to make one for you

PR #2750, opened 2016-02-08
Add Error Message when FW is not found on Server.

billbonney:
This change will display a message to a the user when a FW image is not found as

`Error: File Not Found. Please check beta firmware version is available`
`Error: File Not Found. Please check developer firmware version is available`
`Error: File Not Found. Please check stable firmware version is available`
etc..

ie.

```
QGroundControl can upgrade the firmware on Pixhawk devices, 3DR Radios and PX4 Flow Smart Cameras.
All QGroundControl connections to vehicles must be disconnected prior to firmware upgrade.
Please unplug your Pixhawk and/or Radio from USB.
Plug in your device via USB to start firmware upgrade. 
Found device: Pixhawk
Connected to bootloader:
Version: 5
Board ID: 11
Flash size: 2080768
Downloading firmware...
From: http://firmware.diydrones.com/Copter/beta/PX4-quad/ArduCopter-v4.px4
Error: File Not Found. Please check beta firmware version is available.
Upgrade cancelled
------------------------------------------
If upgrade failed, make sure to connect directly to a powered USB port on your computer, not through a USB hub. Also make sure you are only powered via USB not battery.
Download complete
```

### DIFF
--- a/src/QGCFileDownload.cc
+++ b/src/QGCFileDownload.cc
@@ -121,6 +121,10 @@ void QGCFileDownload::_downloadError(QNetworkReply::NetworkError code)
     
     if (code == QNetworkReply::OperationCanceledError) {
         errorMsg = "Download cancelled";
+
+    } else if (code == QNetworkReply::ContentNotFoundError) {
+        errorMsg = "Error: File Not Found";
+
     } else {
         errorMsg = QString("Error during download. Error: %1").arg(code);
     }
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Thanks @billbonney. I didn't understand that these could be missing at times. Just thought they were wrong.

billbonney:
The images for stable/beta are only going to be missing until the first release of each, and then it should be good always. (Chicken and Ege problem ;-) )

In any case, the message changes are now more meaningful if something goes wrong with any supported platform and firmware type..
:+1:

DonLakeFlyer:
> the message changes are now more meaningful if something goes wrong with any supported platform and firmware type..

Absolutely. Great change, thanks.

Issue #2779, opened 2016-02-11
Really large waypoint file causes Plan View mission item ListView to be very slow to scroll


DonLakeFlyer:
[large_test.waypoints.txt](https://github.com/mavlink/qgroundcontrol/files/126797/large_test.waypoints.txt)

DonLakeFlyer:
Also need to check that the altitude displays is a flickable so you can scroll it.

DonLakeFlyer:
Better implementation is to not show the list when count get high. But instead show a single mission editor. Then the user can click waypoints in the map to change that single editor.

PR #2784, opened 2016-02-12
Handle very large numbers of waypoints

DonLakeFlyer:
- If the item count goes above 100 don't show the editor list
- Move Remove All to sync button so it is available even when editors are not show
- Various change to improve perf when item count is high

Fix for Issue #2779

screen shot 2016-02-12 at 11 04 31 am


DonLakeFlyer:
I figured out the real way to do this.

PR #2787, opened 2016-02-12
Correctly handle large numbers of waypoints

DonLakeFlyer:
Loads 800 waypoints in under a second. Full UI available without lag.


LorenzMeier:
Wohoo!

Issue #2803, opened 2016-02-14
Implement direct control commands: Takeoff, Land, ...


DonLakeFlyer:
Takeoff, Land, Go here, Kill Switch?

DonLakeFlyer:
@dogmaphobic Hey icon man: I think this will likely be a DropButton. Any ideas for icons?

dagar:
I'm looking forward to "go here". Does PX4 need anything added for this to work?

DonLakeFlyer:
Lorenz is specing it all out right now.

LorenzMeier:
Here is what I was thinking so far:

Go to location command (MAV_CMD_DO_REPOSITION). This command should lead to a mode switch to POSCTL / GUIDED. Stick activity should override the position command and let the user take over.

param1: velocity: -1: default. > 0: ground speed
param2: Mode: -1: Hold current position, < -1: Loiter radius, 1: go to position (multicopter mode), 2: go to position: Fixed wing mode (for VTOL capable airframes)
param4: Yaw, NaN to indicate no yaw angle change
param5: Lat
param6: Lon
Param7: Alt

DonLakeFlyer:
* Using NaN to indicate no yaw change is going to be tricky to handle from a ui side. Will require both a check box and a text field. That's a one-off from anything else we have auto-generated in the ui right now. Would be better if it was something the user could type in as well.
* How do you change altitude only? Maybe speed only as well?
* Same issue as NaN with Mode being a combination value and Enum style field. Hard to represent with a good ui. Would be better to have Mode enum values in one param so you can show combo. Then Loiter by itself shown in a text field.

DonLakeFlyer:
Also as far as direct control with PX4 firmware: RETURN_TO_LAUNCH doesn't appear to be implemented in missions just NAV_LAND. Maybe it's implemented only as a direct command? Really need both command in both command/mission mode.

DonLakeFlyer:
Also for MAV_CMD_NAV_TAKEOFF you would want options there to go just straight up and also not affect yaw.

DonLakeFlyer:
Did this discussion move to somewhere else?

LorenzMeier:
No, it just didn't progress yet.

DonLakeFlyer:
Include other things like trad heli motor stop from ArduPilot.

DonLakeFlyer:
Relate to Issue #2669 Kill switch

LorenzMeier:
[...]

PR #2812, opened 2016-02-14
Takeoff by clicking Vehicle

DonLakeFlyer:
This should let us get started on wringing out user model and firmware issues:
screen shot 2016-02-14 at 1 43 28 pm
Related to Issue #2803

QGC sends a MAV_CMD_NAV_TAKEOFF ins response to this ui with alt set and lot/lon=0.

Tried it with master SITL in Manual and POSTL with/without virtual thumbs. It sort of reacted to the command but never went up to the full height.


DonLakeFlyer:
@LorenzMeier Not sure if I should merge this. Wondering if it is appropriate use of TAKEOFF command for direct control.

PR #2818, opened 2016-02-15
Click on Vehicle to takeoff

DonLakeFlyer:
Replacement for #2812


DonLakeFlyer:
I"m going to pull this one back until the firware side is ready. Reason is that there are some other widget changes in here for thumbstick that I need in a separate pull.

Issue #2586, opened 2016-01-12
ParameterlLoader handling of multiple components and hash


DonLakeFlyer:
- Request parameter list as well as hash values at start
- The hash cache must be separated out by component
- hash values will come back from each component
- If hash comes back from component and the hash matches write a value back to that components hash value, which will in turn stop the param list. Then use the cache for that component

dogmaphobic:
While at it, this:

```
 /* compute the crc of the local cache to check against the remote */
    foreach(int component, cache_map.keys()) {
        foreach(int id, cache_map[component].keys()) {
            const QString name(cache_map[component][id].first);
            const void *vdat = cache_map[component][id].second.second.constData();
            crc32_value = QGC::crc32((const uint8_t *)qPrintable(name), name.length(),  crc32_value);
            crc32_value = QGC::crc32((const uint8_t *)vdat, sizeof(uint32_t), crc32_value);
        }
    }
```

makes an assumption all values are sizeof(uint32_t), which assumes the memory location where they were first set was initialized (zeroed) before setting. I had the CRC running only for the length of the actual **type** and was getting different values than those computed by QGC.

DonLakeFlyer:
> write a value back to that components hash value, which will in turn stop the param list.

@LorenzMeier Has this change been made on the firmware side?

LorenzMeier:
not yet

NaterGator:
> While at it, this: [...] makes an assumption all values are sizeof(uint32_t), which assumes the memory location where they were first set was initialized (zeroed) before setting. I had the CRC running only for the length of the actual type and was getting different values than those computed by QGC.

To be fair I made that assumption knowing the only two union initializers used in the python script are float and 32-bit int:
https://github.com/PX4/Firmware/blob/56d67c6f6c8f19d5362936257ed4e6a76513c0dc/Tools/px_generate_params.py#L76

DonLakeFlyer:
@NaterGator You familiar with this code, wondering if you could update it:
- To follow the new spec
- Also need to handle type size appropriately so that we don't need to rely an extra space being zeroed. ArduPilot uses additional type sizes smaller than uiint32_t, so we'll need that to bring them on board. Also to guard against PX4 adding additional smaller types further down the road.

NaterGator:
[...]

PR #2821, opened 2016-02-15
New parameter cache loading spec

NaterGator:
- Request list of all parameters, expecting device to lead each component with _HASH_CHECK cache value
 - Test for hash collision, if so our cache is valid and respond with _HASH_CHECK to stop listing
 - Else: let listing continue as normal to get updated param values
 - Store each cache by systemID + componentID

For #2586 and PX4/Firmware#3764

### DIFF
--- a/src/FactSystem/FactMetaData.h
+++ b/src/FactSystem/FactMetaData.h
@@ -121,7 +121,7 @@ public:
     static const int defaultDecimalPlaces = 3;
 
     static ValueType_t stringToType(const QString& typeString, bool& unknownType);
-    static size_t typeToSize(ValueType_t);
+    static size_t typeToSize(ValueType_t type);
 
 private:
     QVariant _minForType(void) const;
--- a/src/FactSystem/FactMetaData.h
+++ b/src/FactSystem/FactMetaData.h
@@ -121,6 +121,7 @@ public:
     static const int defaultDecimalPlaces = 3;
 
     static ValueType_t stringToType(const QString& typeString, bool& unknownType);
+    static size_t typeToSize(ValueType_t);
 
 private:
     QVariant _minForType(void) const;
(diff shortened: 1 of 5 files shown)


src/FactSystem/FactMetaData.h
```diff
@@ -121,6 +121,7 @@ class FactMetaData : public QObject
     static const int defaultDecimalPlaces = 3;
 
     static ValueType_t stringToType(const QString& typeString, bool& unknownType);
+    static size_t typeToSize(ValueType_t);
```
DonLakeFlyer: Can you add type here? I know nitpick, but sometimes people use this as an indication of an unused param. So it could be confusing.

src/FactSystem/ParameterLoader.cc
```diff
@@ -601,40 +602,42 @@ void ParameterLoader::_writeParameterRaw(int componentId, const QString& paramNa
...
+{
+    const QString spath(QFileInfo(QSettings().fileName()).dir().absolutePath());
+    return spath + QDir::separator() + "ParamCache";
+}
+
+QString ParameterLoader::parameterCacheFile(int uasId, int componentId)
 {
-    const QDir settingsDir(QFileInfo(QSettings().fileName()).dir());
-    return settingsDir.filePath("param_cache");
+    return parameterCacheDir().filePath(QString("%1_%2").arg(uasId).arg(componentId));
```
DonLakeFlyer: Does this just leave the file name as #_#? Can't quite tell fron just the diff. If so, would still be good to have a user readable file name to identify file plus the vehicle id/compid.

NaterGator: The structure looks like this:
image

image

Would veh<ID>comp<ID>.bin be preferable?

DonLakeFlyer: I think that's ok then.

src/FactSystem/FactMetaData.h
DonLakeFlyer: @NaterGator Can you make this one change and this should be good to go right?

NaterGator: Yup, sorry I will make it momentarily!

NaterGator:
This has been rebased onto master and is ready to go.

DonLakeFlyer:
Thanks Nate.

Issue #2837, opened 2016-02-17
Failure to save mission file


WickedShell:
Reproduction steps QGC v2.92.2-123-g4f0562b
Connect to APM:Plane (on pixhawk)
Click to add new waypoints (I had 6 in this but others still seem to error)
Select save to file from the side. Nothing happens, an error does appear in the console.

Full console text below:

[I] at :0 - "Logging ini file directory "C:/Users/WickedShell/AppData/Local/QtPr
oject""
[I] at :0 - "Filter rules "qt.network.ssl.warning=false\n""
[I] at :0 - "Settings location "C:/Users/WickedShell/AppData/Roaming/QGroundCont
rol.org/QGroundControl.ini" Is writable?: true"
[I] at :0 - "Mapping cache directory: "C:/Users/WickedShell/AppData/Local/cache/
QGCMapCache55""
[I] at :0 - "home item QGeoCoordinate(37.8038, -122.462, 0)"
[I] at :0 - "Orphaned log file count 0"
[I] at :0 - "Mapping cache directory: "C:/Users/WickedShell/AppData/Local/cache/
QGCMapCache55""
[E] at qrc:qml/MissionEditor.qml:447 - "qrc:qml/MissionEditor.qml:447:21: QML Li
stView: Binding loop detected for property "height""
[E] at qrc:qml/MissionEditor.qml:447 - "qrc:qml/MissionEditor.qml:447:21: QML Li
stView: Cannot set a negative cache buffer"
[I] at :0 - "home item QGeoCoordinate(37.8038, -122.462, 0)"
[I] at :0 - "Temp log "D:/Scratch/FlightData171152.mavlink" false"
[I] at :0 - "home item QGeoCoordinate(34.5683, -112.479, 1.674)"
[I] at :0 - "home item QGeoCoordinate(34.5683, -112.479, 1.678)"
[E] at qrc:qml/MissionEditor.qml:106 - "qrc:qml/MissionEditor.qml:106: TypeError
: Property 'saveToFile' of object MissionController(0x62382e8) is not a function
"

PR #2839, opened 2016-02-18
Fix incorrect method name

DonLakeFlyer:
Fix for Issue #2837. Screwed it up when I added mission load/save to tablet.


WickedShell:
Should I be opening issues for any similar not a function/type errors I see while testing? In general unless it broke something for me I've been ignoring the console/one offs that periodically happen.

DonLakeFlyer:
Other than at shutdown you should put in Issues. There are some known ones like:
```
qrc:qml/MissionEditor.qml:447:21: QML ListView: Binding loop detected for property "height"
qrc:qml/MissionEditor.qml:447:21: QML ListView: Cannot set a negative cache buffer
```
Which is just a control startup problem, but it stills work correctly. But other than than one the rest should be clean.

Issue #2836, opened 2016-02-17
Land improvement suggestions


jcooper21:
Hi,

After multiple missions on APM Plane 3.5 with QGC I found the following areas that could be improved. 
1. Land waypoint altitude should default to 0m instead of 25m
2. When creating the landing sequence with do start land command the plane glide slope needs to be shown as is comes in for landing. The current setup can result in too steep of autoland. 
3. A confirmation message would be useful when waypoints are uploaded to the craft. 
   4.the waypoint lines are orange and the plane line as it flys is orange. Is would be very helpful to have these as different colors to make it easier to see.

LorenzMeier:
All these issues apply to PX4 as well, so I renamed this land improvement suggestions.

LorenzMeier:
In particular the land altitude not being 0 m is really something we should fix immediately.

DonLakeFlyer:
> In particular the land altitude not being 0 m is really something we should fix immediately.

Did I ever tell you how much I like that mission command editors are defined in json? :)

LorenzMeier:
Are you now bragging or complaining that I didn't sent a PR yet but are lamenting in issues? 8). I should have a long time ago!

DonLakeFlyer:
I was trying to brag and ended up putting my foot into my mouth because altitude default isn't in the default. But it will be soon!

jcooper21:
Gents, I forgot to mention the last issue I noticed which is that terrian follow does not appear to work. At least I couldn't figure out how to get it to work.

jcooper21:
For normal waypoints shouldn't they have a default altitude of 30m relative or something so you don't rush crashing in the ground from the current defaults.

jcooper21:
It would be easiest if the user could set defaults

DonLakeFlyer:
> I noticed which is that terrian follow does not appear to work

Terrain following is not yet supported.

DonLakeFlyer:
> For normal waypoints shouldn't they have a default altitude of 30m relative

Default altitude for generic mission item which acccepts an altitude is 25 meters.

DonLakeFlyer:
@jcooper21 For #2: Do you have a mission file you can attach that I can use to test against with Mission Planner. I assume Mission Planner shows something special here that I need to understand?

jcooper21:
Which feature are you looking to test against? What file would you like? Would you like me to screenshot something?

DonLakeFlyer:
[...]

PR #2841, opened 2016-02-18
Plan: Add Gradient to bottom item status

DonLakeFlyer:
Also fixed bug where last item altitude was incorrectly pulled from home position.

screen shot 2016-02-17 at 6 32 25 pm


DonLakeFlyer:
More work for #2836

PR #2846, opened 2016-02-18
Offline Map Caching

dogmaphobic:
This implements Offline Map caching. It also handles normal disk caching as a byproduct (memory caching is still handle by Qt itself).

Map tiles are now stored in an SQL database. You can create "sets" of tiles, defining regions and zoom levels to capture tiles for offline use. Once in the database, all tiles (from all sets) are available for QGC. The grouping into sets is primarily for easier management of tiles to delete when no longer needed.

screen shot 2016-02-17 at 10 46 58 pm

When adding a new Map Tile Set, you define the region of interest (by panning and zooming the map) and defining the minimum and maximum zoom levels to download and cache. The display will show the exact number of tiles in the current defined set and an *estimate* of the amount of space it will take.

screen shot 2016-02-17 at 10 45 52 pm

In addition, I've enabled support for MapBox maps. To use them however, you need your own account. Once you create an account (which starts at free), you enter your own token. Once the token is entered, you will be able to select MapBox maps.

Memory cache limit defines how much RAM is allocated for caching map tiles. Disk cache defines the maximum amount of space the *Default Set* will use. That's the set that constantly caches whatever it is on the Fly/Mission map. As the limit is reached, older tiles are dropped off. Custom tile sets are not included in this limit. You have full control of the space they use when defining it and once created, they never grow.
screen shot 2016-02-17 at 10 47 30 pm

With all that said, this is a rather large amount of code and I still have some tweaks to make. For instance, I wanted to take a snapshot of the map region for a given set but that is not working until we're fully QtQuick.


NaterGator:
This is awesome Gus!

LorenzMeier:
Awesome! Cache party!

DonLakeFlyer:
Killer feature!

LorenzMeier:
Closing and opening to see what Travis says about this.

Issue #2726, opened 2016-02-01
Mission planning user experience


RomanBapst:
Users are confused with how a plane flies along a path specified by the mission items. They expect the plane to fly by the waypoint as close as specified by the acceptance radius in the mission item.
However, the plane will rather fly along the waypoints in a band which is related to the L1 radius of the controller. It would be useful to indicate this band in the planning view of QGC. This would tell the user that the plane will be flying somewhere inside that band.
- [ ] Enforce minimum waypoint distance (PX4: multicopter: 2 \* NAV_ACC_RAD = min WP distance, fixed wing: 1.0f / M_PI \* FW_L1_DAMPING \* FW_L1_PERIOD \* FW_AIRSPD_TRIM = min WP distance)
- [ ] Draw spline instead of lines
- [ ] Draw spline as transparent band, not as line, to indicate uncertainty
- [ ] Parametrize spline based on onboard parameters (PX4: multicopter: 2 \* NAV_ACC_RAD = min WP distance, fixed wing: 1.0f / M_PI \* FW_L1_DAMPING \* FW_L1_PERIOD \* FW_AIRSPD_TRIM = min WP distance)

This should all be driven by a MAVLink status message, however, to craft the message we first need to learn what it actually needs to contain, so lets start with params we already have.

LorenzMeier:
@DonLakeFlyer @dogmaphobic We could start with a simple spline 200 pixels wide and semi-transparent. We could then go on to refine it based on the actual mission math so it reproduces the path as close as possible.

LorenzMeier:
@NaterGator @dagar Would you be interested to look into this? It seems like this is the primary user experience issue left, because users can plan missions that are absolutely not doable by the aircraft and the best-effort based result is confusing them a lot. We need to expose the result on the UI and enforce some basic parameters like waypoint distances.

LorenzMeier:
This issue is a result of missing visibility from the user side: https://github.com/PX4/Firmware/issues/3639

DonLakeFlyer:
[...]

PR #2849, opened 2016-02-19
Initial Complex mission item support

DonLakeFlyer:
Complex Mission Item described:
- Encapsulates a set of individual mission items into a single entity
- While Mission Editing a complex mission item is represented as a single item in the list
- Has it's own visual representation on the map
- Has it's own mini-editor in the right hand item list
- Has an entry and exit waypoint
- A mission can include multiple complex mission items
- Full load/save round tripping for all of it's information beyond just the mission items themselves

This is the initial work towards the above goal. The first usage of a Complex mission item will be a Survey Grid. Another example to come later would be Structure Scan.

Initial Survey support is hidden behind a "Experimental Survey [WIP - no bugs reports please]" checkbox in General Settings. Still a very long way to go but it is showing signs of life.

screen shot 2016-02-18 at 7 26 38 pm

Related to Issue #2726


LorenzMeier:
Nice!

LorenzMeier:
@birchera FYI. We will also want to support a mission command to set up the trigger.

Issue #2850, opened 2016-02-19
Map cache related crash on boot


LorenzMeier:
@dogmaphobic Just logging it here - will do a clean rebuild now:

```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   org.qt-project.QtLocation     	0x00000001081ef4b1 QCache3Q<QGeoTileSpec, QGeoCachedTileDisk, QCache3QTileEvictionPolicy>::rebalance() + 1009
1   org.qgroundcontrol.qgroundcontrol	0x0000000107877a71 QGeoTiledMappingManagerEngineQGC::_setCache(QMap<QString, QVariant> const&) + 2497
2   org.qgroundcontrol.qgroundcontrol	0x0000000107875ef3 QGeoTiledMappingManagerEngineQGC::QGeoTiledMappingManagerEngineQGC(QMap<QString, QVariant> const&, QGeoServiceProvider::Error*, QString*) + 5907
3   org.qgroundcontrol.qgroundcontrol	0x0000000107874582 non-virtual thunk to QGeoServiceProviderFactoryQGC::createMappingManagerEngine(QMap<QString, QVariant> const&, QGeoServiceProvider::Error*, QString*) const + 50
4   org.qt-project.QtLocation     	0x00000001081e76cc QGeoMappingManager* QGeoServiceProviderPrivate::manager<QGeoMappingManager, QGeoMappingManagerEngine>(QGeoServiceProvider::Error*, QString*, QGeoMappingManager**) + 108
5   libdeclarative_location.dylib 	0x0000000112eadbc7 QDeclarativeGeoMap::pluginReady() + 39
6   libdeclarative_location.dylib 	0x0000000112eaec96 QDeclarativeGeoMap::setPlugin(QDeclarativeGeoServiceProvider*) + 326
7   libdeclarative_location.dylib 	0x0000000112eb15ab QDeclarativeGeoMap::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) + 203
8   libdeclarative_location.dylib 	0x0000000112eb2177 QDeclarativeGeoMap::qt_metacall(QMetaObject::Call, int, void**) + 71
9   org.qt-project.QtQml          	0x000000010947ab6d QQmlVMEMetaObject::metaCall(QMetaObject::Call, int, void**) + 573
[...]

PR #2852, opened 2016-02-19
Hopefully fixes crash Lorenz is seeing.

dogmaphobic:
I can't reproduce this here but the stack trace in #2850 is in the same place my Windows build had issues with. I'm basically trying to get Qt NOT to use disk cache (redundancy). There is no way to disable it altogether so I'm telling it to use 1kb of disk space. This worked for all but Windows, which crashed unless I set it to 100kb. Apparently I was being lucky.

I'm now setting it to 100kb for all platforms. This will all change when we switch to 5.6.


DonLakeFlyer:
Fixed it for me

PR #2854, opened 2016-02-19
Flickable in Offline Maps

dogmaphobic:
The Nexus 7 (Android) has a very wide but short window. Some parts of the Offline maps UI did not fit the height. I've added Flickable where appropriate.

While at it, I've reordered the header files as requested by @DonLakeFlyer.

screen


DonLakeFlyer:
Thanks Gus, I know I'm kinda annoying on that stuff :)

PR #2858, opened 2016-02-19
Fix qt inconsistencies v5

tcanabrava:
Last work in keeping clazy warnings clean.
we should use clazy plugin as default on new commits. :)


dagar:
You could run clazy on travis-ci if you wanted to check every commit.

DonLakeFlyer:
Yes, we should turn that on once we get clean. @tcanabrava Can you coordinate with Daniel on that?

PR #2859, opened 2016-02-20
Android stay on

dogmaphobic:
I tried all possible permutations of the desired `getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);` with no effect whatsoever (as reported by many people at stakoverflow). The issue, I think, is that it needs to be called at a point in the Activity creation process we have no access (that's within the Qt bootstrap code).

I'm using the PowerManager class to force Android from dimming and turning off the screen *while* a vehicle is connected. If there are no vehicles connected, I restore to normal (to save the battery).


dogmaphobic:
AppVeyor is barfing trying to download the installer tool.

dogmaphobic:
@dagar Is there a way to restart AppVeyor without restarting Travis as well in the process? The only way I know is to close and re open the PR.

dagar:
You should already have sufficient access to restart individual travis-ci builds if logged in through github. I'll get you access to do the same on appveyor.

Issue #2863, opened 2016-02-22
Sync up PX4 supported mission command to firmware


DonLakeFlyer:
Current firmware:
```
        // check if we find unsupported items and reject mission if so
        if (missionitem.nav_cmd != NAV_CMD_IDLE &&
            missionitem.nav_cmd != NAV_CMD_WAYPOINT &&
            missionitem.nav_cmd != NAV_CMD_LOITER_UNLIMITED &&
            /* not yet supported: missionitem.nav_cmd != NAV_CMD_LOITER_TURN_COUNT && */
            missionitem.nav_cmd != NAV_CMD_LOITER_TIME_LIMIT &&
            missionitem.nav_cmd != NAV_CMD_LAND &&
            missionitem.nav_cmd != NAV_CMD_TAKEOFF &&
            missionitem.nav_cmd != NAV_CMD_PATHPLANNING &&
            missionitem.nav_cmd != NAV_CMD_DO_JUMP &&
            missionitem.nav_cmd != NAV_CMD_DO_SET_SERVO &&
            missionitem.nav_cmd != NAV_CMD_DO_DIGICAM_CONTROL &&
            missionitem.nav_cmd != NAV_CMD_DO_VTOL_TRANSITION) {

            mavlink_log_critical(_mavlink_fd, "Rejecting mission item %i: unsupported action.", (int)(i+1));
            return false;
        }
```

Current QGC:
```
   list << MAV_CMD_NAV_WAYPOINT
         << MAV_CMD_NAV_LOITER_UNLIM << MAV_CMD_NAV_LOITER_TIME
         << MAV_CMD_NAV_LAND << MAV_CMD_NAV_TAKEOFF
         << MAV_CMD_NAV_ROI
         << MAV_CMD_DO_JUMP
         << MAV_CMD_CONDITION_DELAY
         << MAV_CMD_DO_VTOL_TRANSITION << MAV_CMD_NAV_VTOL_TAKEOFF << MAV_CMD_NAV_VTOL_LAND
         << MAV_CMD_DO_DIGICAM_CONTROL
         << MAV_CMD_DO_SET_SERVO
         << MAV_CMD_DO_CHANGE_SPEED
         << MAV_CMD_NAV_PATHPLANNING;
```

Leave in VTOL takeoff/land since those are coming soon.

DonLakeFlyer:
So that means remove:
MAV_CMD_NAV_ROI
MAV_CMD_CONDITION_DELAY
[...]

PR #2865, opened 2016-02-22
Remove unsupported commands

DonLakeFlyer:
Fix for Issue #2863


LorenzMeier:
Sorry, DO_SPEED needs to stay in because of impending support (which is already tested): https://github.com/PX4/Firmware/pull/3567

PR #2867, opened 2016-02-23
Fix crash when load from Vehicle

DonLakeFlyer:
- Vehicle running mission
- Plan: Load from Vehicle
- Boom!


DonLakeFlyer:
Closing this pull since I have this included in another pull

PR #2869, opened 2016-02-23
Fix presentation of param dialog to avoid having users click on botto…

LorenzMeier:
…m buttons

screen shot 2016-02-23 at 19 00 56


DonLakeFlyer:
This isn't how I want to do it. I'll handle it.

DonLakeFlyer:
Closing this one for the other

PR #2885, opened 2016-02-25
Allow map zoom to go to 20

DonLakeFlyer:


dogmaphobic:
It should go to 11! One louder...

DonLakeFlyer:
A long time ago we lived in an apartment that had a wall mounted AC unit that went to 11. We joked about it all the time :)

PR #2892, opened 2016-02-25
More work on complex mission items

DonLakeFlyer:
- Load/Save working. You can now round trip a complex mission item (like a survey) without having it devolve into just a huge pile of waypoints.
- Entry point working. A complex mission item has both and entry and exit waypoint.

Here is an example after I reloaded from a mission file. You can see the entry into the survey area:
screen shot 2016-02-25 at 3 20 33 pm


DonLakeFlyer:
Next up, in order:
* Working exit point points
* Better polygon creation
* Grid creation

PR #2898, opened 2016-02-26
Boot perf

DonLakeFlyer:
App boot was getting a little pokey on Tablets so I put a little time into Qml profiling. Improved boot by maybe 20%. Some of it it perceived due to using asynchronous loading of map widgets which display after map comes up. Still lots move possible. But enough for now.


DonLakeFlyer:
Pretty significan't improvement on my dog slow 7" Asus

PR #2924, opened 2016-03-03
Domain changes for ArduPilot

lvale:
Domain changes for the ArduPilot locations from DIYDrones to
ArduPilot.org


DonLakeFlyer:
Thanks @lvale

tridge:
note that the SRTM server should also change (does QGC use the ardupilot SRTM server?)

lvale:
@tridge I'm not sure that QGC retrieves SRTM data, yet :)

Issue #2767, opened 2016-02-10
Parameter/Airframe meta data needs to be versioned


DonLakeFlyer:
This will allow the latest version to take precedence over older versions. Whether it is cached or internal to QGC.

PR #2930, opened 2016-03-05
Versioned parameter meta data cache

DonLakeFlyer:
Parameter meta data handling in QGC was not quite right. It wasn't versionsed correctly and the meta data cache also wasn't verioned correctly. One major problem was that if you flashed firmware, the meta data was in the cache. If you then updated QGC with newer parameter meta data the older cached version would still take precedence which was incorrect. Also the cache did not support meta data from both PX4 and ArduPilot. Now both internal QGC and the meta data cache is correctly versioned and includes support for caching PX4 and ArduPilot param meta data during flash.

There is a corresponding Firmware pull which updates the meta data generator to include the right verion numbering.


DonLakeFlyer:
Fix for Issue #2767

PR #2953, opened 2016-03-06
Use clearer languagee to differentiate between flight mode selection,…

LorenzMeier:
… remove a lot of the text in advanced config


LorenzMeier:
screen shot 2016-03-06 at 17 42 39

screen shot 2016-03-06 at 17 36 55

LorenzMeier:
screen_single

LorenzMeier:
screen_multi

DonLakeFlyer:
Yeah better. I wonder about pulling out PWM values. I guess they are evenly distributed so it may be ok.

LorenzMeier:
@DonLakeFlyer screens are up. I fixed the typos that you partially still find in them.

LorenzMeier:
I pulled the PWM values out in simple. I don't think we should touch advanced for now.

LorenzMeier:
@DonLakeFlyer Could you take over and add the channel monitor here? Then it would be good to go I guess.

DonLakeFlyer:
I'm taking this one over

Issue #2275, opened 2015-11-24
Mission restart button


RomanBapst:
What are the alternatives for restarting the mission at the moment?

DonLakeFlyer:
Not currently supported. May a custom command widget and send you own command. Assuming there is a MAV_CMD that will restart mission, don't know.

dagar:
What if in the fly view clicking a waypoint sends MISSION_SET_CURRENT?

LorenzMeier:
@DonLakeFlyer @dagar Yes, QGC should send that command on clicking a waypoint.

LorenzMeier:
Closing as stale.

PR #2986, opened 2016-03-10
Implement click-to-retask via MISSION_SET_CURRENT in flight view for #2275

NaterGator:
Initial implementation. Would love feedback from mobile testers. 

Also raises the issue: in my opinion on desktop we should be changing the cursor to clue the user that an action will be taken if they click on something. Currently the cursor never changes and it is a wild guess. 

Related issue: #2275

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewMap.qml
+++ b/src/FlightDisplay/FlightDisplayViewMap.qml
@@ -48,6 +48,7 @@ FlightMap {
     property bool   _activeVehicleCoordinateValid:  _activeVehicle ? _activeVehicle.coordinateValid : false
     property var    activeVehicleCoordinate:        _activeVehicle ? _activeVehicle.coordinate : QtPositioning.coordinate()
     property var    _gotoHereCoordinate:            QtPositioning.coordinate()
+    property int    _retaskSequence:                0
 
     Component.onCompleted: {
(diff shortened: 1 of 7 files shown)


DonLakeFlyer:
I can't remember is MissionItemView only used in Fly and not Plan? Plan click behavior needs to stay the same.

NaterGator:
Correct, MissionItemView is only used for fly. Plan uses MissionItemIndicator directly.

DonLakeFlyer:
I think this needs confirmation. If you look at how I do guided mode goto you should be able to use the same mechanism. I tried testing it in SITL, but master SITL seems to be a basket case right now.

NaterGator:
@DonLakeFlyer by "confirmation" do you mean mean closed loop packet delivery?

DonLakeFlyer:
No I mean that you should ask the user if this is what they really wanted to do. Example below for change altitude in guided mode. Same prompt happens in guided mode if you click on the map to move the vehicle to a new position.
screen shot 2016-03-09 at 3 13 58 pm

DonLakeFlyer:
@NaterGator If you want I can merge this then I'll add the confirmation check.

DonLakeFlyer:
Awesome, thanks Nate

PR #2988, opened 2016-03-10
Testing precompiled headers on OSX to speed up clang builds

NaterGator:
It would be nice if they didn't stall travis-ci.


NaterGator:
Well that's disappointing :-1:

DonLakeFlyer:
I've tried this before with crappy results as well

PR #2989, opened 2016-03-10
Fix #2987: QGroundControlQmlGlobal.h: flightMapZoom should be a doubl…

### LINKED ISSUE #2987: QGroundControlQmlGlobal.h: flightMapZoom should be a double, not an integer
Zooming in and out in the flight map and minimizing it (by maximizing the video PIP) yields to `Unable to handle unregistered datatype 'FactMetaData::ValueType_t' for property 'Fact::type'`.

That's because `flightMapZoom` is defined as follows:
```
    Q_PROPERTY(int         flightMapZoom         MEMBER _flightMapZoom       NOTIFY flightMapZoomChanged)
``` 

This leads to very weird debugging behaviors in QML/JS where e.g. both branches in an if/else are executed.

shadow1runner:
…e, not an integer

Note that a change in zoom signals a `zoomLevelChanged` event; this in turn triggers `FlightDisplayViewMap` which sets the property changed in this commit to
```
onZoomLevelChanged: QGroundControl.flightMapZoom = zoomLevel
```

This triggers the following message in Application Output: `onZoomLevelChanged: QGroundControl.flightMapZoom = zoomLevel`.

This was the case before and after this fix, I don't know if that's a real issue though

### DIFF
--- a/src/QmlControls/QGroundControlQmlGlobal.h
+++ b/src/QmlControls/QGroundControlQmlGlobal.h
@@ -78,7 +78,7 @@ public:
 
     Q_PROPERTY(QGeoCoordinate lastKnownHomePosition READ lastKnownHomePosition  CONSTANT)
     Q_PROPERTY(QGeoCoordinate flightMapPosition     MEMBER _flightMapPosition   NOTIFY flightMapPositionChanged)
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Yup, that's wrong. Thanks.

PR #3009, opened 2016-03-13
Px4 mode cleanup

LorenzMeier:
@DonLakeFlyer These are the mode switch renames.


DonLakeFlyer:
@LorenzMeier Submitting pulls from something other than a fork doesn't work.

dagar:
It should though, I'll fix it.

DonLakeFlyer:
Closing this to create a new pull. There are other things which needs to change when the names change. That's not quite clear fromthe current code. Need to fix that.

Issue #2948, opened 2016-03-06
We need to wipe the map cache on upgrades


LorenzMeier:
@dogmaphobic I upgraded two machines to new QGC and got these map cache issues on both, which only resolve if you chase up the map cache dir and delete it manually. QGC should wipe the existing map cache on the next update.

![screen shot 2016-03-06 at 11 46 26](https://cloud.githubusercontent.com/assets/1208119/13553778/3be78d18-e394-11e5-99af-e73cb8519c27.png)

DonLakeFlyer:
Isn't that a cache bug. Why clear the cache? The missing tiles should just fill in.

LorenzMeier:
They never do if you run QGC off an "old" directory. We need to wipe the cache when deploying the next stable, because all users will run into this.

LorenzMeier:
Just to be very clear: This never resolves if you don't delete the directory manually. And it never shows after. Its a transition issue from Qt's engine to our own engine.

DonLakeFlyer:
The old a new cache are completely different. The new one is SQL, I think the old one was just internet cache? I really don't think they overlap.

dogmaphobic:
I haven't tried it but I have a clue. Qt will look for its own cache before I ever get asked for a tile. That means it would find its old tile and never request it from our plugin. It doesn't explain the missing tile though.

One way or the other, keeping the old tile cache around is just wasted disk space. I will take a look and see about cleaning out an old tile cache if I see one...

dogmaphobic:
Done with #3027.

PR #3027, opened 2016-03-14
Deleting old map tile disk cache if one is found.

dogmaphobic:
This is in response to #2948


dogmaphobic:
Not sure what's up with Travis. My local Ubuntu unittest build finished fine. Trying again...
```
********* Finished testing of RadioConfigTest *********
ALL TESTS PASSED
After app delete
Debugging has finished
```

Issue #2996, opened 2016-03-11
Reverse the order to Windows start menu item addition


DonLakeFlyer:
The issue is on Windows 10, there is a recently added item in the start menu. When you install QGC on Windows. This will show you one of the compatability mode shortcuts. Worried that users will just click on that for running QGC.

@NaterGator Do you think you could look at this? Not sure if reversing will cause other problems. Or if there is some better solution to prvent this.

NaterGator:
Sure, I'll look at how to hint the primary shortcut as the one that should be displayed after install.

DonLakeFlyer:
Thanks Nate. Hope you had a nice vacation.

NaterGator:
We actually need to set [System.AppUserModel.ExcludeFromShowInNewInstall](https://msdn.microsoft.com/en-us/library/windows/desktop/dd391568(v=vs.85).aspx). Unfortunately that looks like in requires WinAPI and we may be stuck using a grotesque script using the NSIS System plugin or compiling our own NSIS plugin DLL to manipulate the property system. Still investigating the best course of action...

DonLakeFlyer:
Nasty

DonLakeFlyer:
Something here: https://sourceforge.net/p/nsis/mailman/message/30592466/?

NaterGator:
Regrettably I'm not sure I can reproduce this issue. I have installed on Win10 and verified the property store reflects the correct values for the compatibility shortcuts, but I would be nice to get a data point from at least a Windows 7 ([reasoning](https://support.microsoft.com/en-us/kb/2745126)) and 10 user to verify this works.

PR #3032, opened 2016-03-14
Demote safe mode links so they don't confuse users of a fresh install

NaterGator:
Hopefully this will address #2996


DonLakeFlyer:
Awesome thanks Nate.

Issue #3000, opened 2016-03-12
Identify Takeoff/Land waypoints 


dagar:
Could you make takeoff/land waypoints display differently (not numbered)? Something like T for takeoff and L for land?

PR #3038, opened 2016-03-16
Add support for customizing the abbreviated representation of mission…

NaterGator:
… items for #3000


DonLakeFlyer:
This is cool. I like the way you plumbed it through.

Issue #2999, opened 2016-03-12
Waypoints display/edit latitude and longitude


dagar:
It would also be nice if there was an easy way to get lat/lon on the map either by clicking/long press or lat/lon displayed in the corner google earth style.

PR #3039, opened 2016-03-16
[WIP] Initial support for text-input of lat/lon position on mission items

NaterGator:
for #2999

I'm going to be adding support for other coordinate systems (UTM, MGRS) momentarily.

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -61,6 +61,7 @@
         <file alias="QGroundControl/Controls/ModeSwitchDisplay.qml">src/QmlControls/ModeSwitchDisplay.qml</file>
         <file alias="QGroundControl/Controls/ParameterEditor.qml">src/QmlControls/ParameterEditor.qml</file>
         <file alias="QGroundControl/Controls/ParameterEditorDialog.qml">src/QmlControls/ParameterEditorDialog.qml</file>
+        <file alias="QGroundControl/Controls/CoordinateEditor.qml">src/QmlControls/CoordinateEditor.qml</file>
         <file alias="QGroundControl/Controls/QGCButton.qml">src/QmlControls/QGCButton.qml</file>
         <file alias="QGroundControl/Controls/QGCCheckBox.qml">src/QmlControls/QGCCheckBox.qml</file>
         <file alias="QGroundControl/Controls/QGCColoredImage.qml">src/QmlControls/QGCColoredImage.qml</file>
--- a/src/MissionEditor/SimpleItemEditor.qml
+++ b/src/MissionEditor/SimpleItemEditor.qml
@@ -47,6 +47,13 @@ Rectangle {
                                      missionItem.commandDescription)
             }
 
+            CoordinateEditor {
+                visible:   missionItem.specifiesCoordinate
+                coordItem: missionItem
+                width:     valuesColumn.width
+                spacing:   _margin
+            }
+
             Repeater {
                 model: missionItem.comboboxFacts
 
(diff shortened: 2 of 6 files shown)


DonLakeFlyer:
Do you have some images of how this works as well as details on what you want to accomplish? I don't want to make normal mission item editing more complicated for this. 

Just looking at the code, If you are using Facts and meta data you should be able to handle validation that way since lat/lon has a normal min/max doesn't it? The FactTextField will validate automatically for you. But I'm not quite sure how this whole thing works since I can't really give a try while I"m out of town. You also may be able to use a translator to handle the different units. But I'm not quite sure. Need to understand more about the what is needed.

NaterGator:
Sure, I can grab some useful screenshots when I get back to my desktop tomorrow. For now maybe this will help:

image

The gist is basically to take that functionality that currently exists for the Home point plan view (editable text fields for Lat/Lon coordinates) and add that functionality to all mission items that designate position in the plan view. Notionally I think it is important for a user to be able to enter in coordinates communicated to them without resorting to opening another application to find out where the points should lie.

The reason I chose to write a custom CoordinateEditor was because similar to units for individual Facts, I would like to support user-configurable coordinate formats. If the USGS says they'd like crops monitored at a specific UTM coordinate or a coast guard wants search assets dispatched to MGRS coordinate 4QFJ12345678 I would like to support entering theme natively.

Since coordinate systems may not conform as nicely as simple unit conversions for facts I thought it appropriate to wrap their editing in a custom component. Admittedly it is contrived for simple lat/lon, but I believe the utility will become apparent as I flesh this out... another decent example is user preference between decimal degrees and DMS.

DonLakeFlyer:
I consider entering a lat/lon into the main editor to be a power user feature. I don't want that in the normal editor that everyone sees. Lat/Lon is already available in the Show All Values raw editor. Right now, that is kind of like jumping off a cliff though. Maybe a new menu item on the hamburger to get to it?

NaterGator:
Sure, and it likely doesn't make sense for screens without much real estate. Perhaps make it a toggle in the user preferences defaulted to off, ex: "Enable manual entry of mission coordinates"?

DonLakeFlyer:
Also for transformation as long as the transformation is from a single value to a lat/lon it will work But it won't work for things like UTM which required two values (zone+UTM coord) to define a position. I get what you are saying that the conversion may be too complex for the simple raw/cooked translator support in FactSystem.

DonLakeFlyer:
> Perhaps make it a toggle in the user preferences defaulted to off, ex: "Enable manual entry of mission coordinates"?

I think a toggle could work. Especially if you just put in in the hamburger menu instead of settings. Then it's still close for power users, but doesn't get in the way of majority use case.

DonLakeFlyer:
I'm also not so thrilled with the Validator support in Qml. It can be very confusing to users when for some strange reason you can't type various characters into a text field. I tried using it ages ago but it lead to very unexpected behavior. The issue is it wants to keep the value in the field correct at all times. But while you are editing numbers, that isn't always the case. You may add a new numeric character thinking you will then delete the other numeric character you no longer need. But the validator won't let you do it. I originally tried validator for FactSystem FactTextField validation but it just created unexplainable annoying behavior. So instead I went the route of post-validation on accept. Instead of on the fly validation as you type.

NaterGator:
I have similar experience. I don't know about QML side but in C++ land I implemented the validator so that text turns red while the field is invalid, but it never prevents you from entering characters. If, upon finishing editing, the value is still invalid it reverted to the correct valid value (last valid or current state, where appropriate). 

I'd like to accomplish the same here if possible, but I know what you mean. There is nothing more aggravating.

DonLakeFlyer:
Do you have screen shots? I'm ok with this if you turn it on/off via a setting. You could control that setting from the hamburger dropdown. I don't want to make the normal user editors any more complicated.

LorenzMeier:
@NaterGator  Please note that all new contributions to QGC have to be made under a dual license including Apache 2.0 + GPLv3. Just so you're aware that this PR now has to fulfil this requirement.

PR #3043, opened 2016-03-17
Initial Qt 5.6 build support.

dogmaphobic:
First shot at a Qt 5.6 build. Mostly, an update on the location plugin to conform with the new interfaces. For Mac OS, I had to bump the target platform to 10.7 (from 10.6) and include a framework for TTS, which apparently is no longer added automatically by qmake.

Tested on Mac OS, iOS and Ubuntu.

Map elements (like vehicle position, waypoints and lines in between them) are not showing up on Mac OS or Ubuntu. They do show up fine on iOS.

### DIFF
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -60,7 +60,11 @@ linux {
         DEFINES += __macos__
         CONFIG += x86_64
         CONFIG -= x86
+equals(QT_MAJOR_VERSION, 5) | greaterThan(QT_MINOR_VERSION, 5) {
+        QMAKE_MACOSX_DEPLOYMENT_TARGET = 10.7
+} else {
         QMAKE_MACOSX_DEPLOYMENT_TARGET = 10.6
+}
         QMAKE_MAC_SDK = macosx10.11
         QMAKE_CXXFLAGS += -fvisibility=hidden
     } else {
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -103,6 +103,9 @@ MacBuild {
     QMAKE_INFO_PLIST    = Custom-Info.plist
     ICON                = $${BASEDIR}/resources/icons/macx.icns
     OTHER_FILES        += Custom-Info.plist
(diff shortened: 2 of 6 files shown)


dagar:
Ping me when you're ready for 5.6 packages on travis-ci.

dogmaphobic:
> Ping me when you're ready for 5.6 packages on travis-ci.

We better wait for @DonLakeFlyer to get back before doing that. Some of the mission elements are not working with Qt 5.6.

dagar:
I wasn't thinking we'd upgrade within hours of the release :smirk:

dogmaphobic:
> I wasn't thinking we'd upgrade within hours of the release

What do I know... :P

DonLakeFlyer:
Is there something we need in 5.6? Reason I ask is because 5.7 for me is the interesting one which has charting. Would rather wait to put the effort there if it's going to be a hassle to move.

dogmaphobic:
For me, the main reason is to be able to capture a snapshot of the map region selected for a map tile set. This is so I can show it when you go through the sets. It's broken in 5.5.

I'm not sure there is any other important reason though.

NaterGator:
There are a few features added that may be worthwhile, not the least of which is better High-DPI support across all platforms. Windows 10 support also improved, and now have MSVC 2015 compiler support which appveyor is [primarily switching over to](http://www.appveyor.com/updates/2016/03/08). I could also see the Qt Positioning support added for Windows GPS receivers being useful.

That said it also makes the inevitable transition to 5.7 less painful if we don't skip minor versions leading up to the release.

DonLakeFlyer:
Sounds worthwhile

DonLakeFlyer:
@NaterGator Do you think you could try to move us to MSVC 2015?

NaterGator:
I already build on MSVC 2015 on my desktop, though we're going to have to wait for appveyor to integrate Qt 5.6 (which has an MSVC 2015 build). The transition should be completely painless (get ready to say I told you so?)

PR #3051, opened 2016-03-21
Fix compilation regarding isnan and isinf

tcanabrava:
isnan() is defined in math.h, and if we include the cmath
instead of math.h, they are defined in std namespace.


tcanabrava:
fixes the following issue on Arch linux:

./src/uas/UAS.cc:1312:28: error: ‘isnan’ was not declared in this scope
     } else if ((!isnan(roll) && roll != manualRollAngle) || (!isnan(pitch) && pitch != manualPitchAngle) ||

tcanabrava:
closed because didn't worked on travis - already updated a patch that should work fine.

Issue #3048, opened 2016-03-19
Add u-Blox RTK messaging support


LorenzMeier:
@DonLakeFlyer The pattern to do this would be equivalent to this PR: Add a new class which parses the GPS module and sends the RTCM messages via the vehicle object. So I'm _NOT_ saying it is related to follow-me, I'm just saying that it needs a similar structure where an object parses an external data source and talks to the vehicle object: https://github.com/mavlink/qgroundcontrol/pull/2817

Issue #2795, opened 2016-02-13
Implement Follow Me using new mavlink message to vehicle


DonLakeFlyer:
@catch-twenty-two Adding this here so I remember who is dong what

PR #3052, opened 2016-03-22
added u-blox USB device IDs

mazgch:
This enable qgroundcontrol to open a u-blox GPS connected directly by USB.


DonLakeFlyer:
What is this for?

NaterGator:
It seems like it would be related to #3048 and #2795

LorenzMeier:
looks good to me.

DonLakeFlyer:
Right, but it's not going to do anything. Shouldn't this wait until the RTK support is in? Otherwise  QGC will allow connect, but do nothing correctly with it?

LorenzMeier:
It should go in. If you read the patch it will not auto-connect, just be able to see the modules as serial ports in manual configuration, which is the same way harmless as seeing random FTDI devices to which we also do not auto-connect.

No reason to block this on RTK. But with this in someone working just on the interface side can actually "see" the device from C++. Right now it won't show up.

PR #3053, opened 2016-03-22
Fix compilation on some linux boxes

tcanabrava:
This shouldn't break a thing on other linux boxes that compiles
fine.


DonLakeFlyer:
Shouldn't all of this just be converted to is qIsNaN instead? Then you don't have to jump through hoops with OS header differences.

tcanabrava:
Okay.

DonLakeFlyer:
Merging this in anyway, so we fix things for now. Then we can switch to qIsNaN

PR #3060, opened 2016-03-23
Update flight mode naming

DonLakeFlyer:
Replacement for #3009


LorenzMeier:
Looks like its good to go.

DonLakeFlyer:
Yup

DonLakeFlyer:
Hmm, it won't compile without that. I don't see where was it getting the defines in your version?

DonLakeFlyer:
I'm gonna merge and we can figure out the other part later.

PR #2817, opened 2016-02-15
Adding follow me option to qground control

catch-twenty-two:
Adds new follow me class that sends motion information to follow me enabled vehicles over mavlink

PR #3070, opened 2016-03-24
Qgc.enhancement.add follow me option

catch-twenty-two:
> DonLakeFlyer 10:13
 any idea why my build is failing?
I think it's a quirk of the build system caused by the fact that you submitted a pull from your master branch. Move it to a named branch and I think it will work.

this is a replacement for the pull request here:
https://github.com/mavlink/qgroundcontrol/pull/2817


DonLakeFlyer:
The main issue here is making this work across both firmwares. The biggest problem I would say is the concept of a Follow Me flight mode. In ArduPilot there is the concept of a "Guided" flight mode, which I think you must be in for Follow Me to work. But then you turn on follow me as a GCS feature specifically on top of that. There is no specific "Follow Me" flight mode.

As it currently stands the check for whether the vehicle is in "Follow Me" mode is not mavlink generic.

DonLakeFlyer:
Can you ifdef the simulator code into debug builds only?

DonLakeFlyer:
Why does the positioning source run at a faster rate than the output to the vehicle? Doesn't seem like the intermediate values which you get are used. Just the last one that came across. Also a bit concerned that the positioning source is always running even if you aren't doing follow me. That's chewing cycles for no reason.

src/FollowMe/FollowMe.cc
```diff
@@ -0,0 +1,178 @@
...
+    }
+}
+
+void FollowMe::_sendGCSMotionReport(void)
+{
+    QmlObjectListModel & vehicles = *_toolbox->multiVehicleManager()->vehicles();
+    MAVLinkProtocol* mavlinkProtocol = _toolbox->mavlinkProtocol();
+    mavlink_follow_target_t follow_target;
+
+    memset(&follow_target, 0, sizeof(mavlink_follow_target_t));
```
DonLakeFlyer: Aren't there additional fields that need to be set on follow_target like timestamp and est_capabilities?

DonLakeFlyer:
Also don't understand how you offset the follow me position from the current GCS position? How does the vehicle not come right up to you?

DonLakeFlyer:
Related to the follow me flight mode discusion. How are MAV_CMD_DO_FOLLOW and MAV_CMD_DO_FOLLOW_REPOSITION supposed to be used? Are those deprecated?

catch-twenty-two:
> The main issue here is making this work across both firmwares. 
I guess I'm not understanding the issue.  Isn't guided mode just another mode where the uav is "guided" by clicking a map rather than using rc?

http://copter.ardupilot.com/wiki/ac2_guidedmode/

Or is it some type of hybrid mixed mode? Unfortunately I have very little experience with Arducopter...  Or are you saying that you must be in a certain mode before you switch into follow me mode?  Like a prerequisite mode? Could we grey out the "follow me" option until the uav is known to be in one of the prerequisite modes?

catch-twenty-two:
> Can you ifdef the simulator code into debug builds only?

will do

catch-twenty-two:
> Why does the positioning source run at a faster rate than the output to the vehicle?

They are actually the same on my phone (1 second)  I can change the update rate to match the position rate for now.  Eventually though the timer will most likely aggregate all of the phone motion data (acc, rotation, ect..) and send it out as fast as possible so the update rate will probably be a lot faster.

catch-twenty-two:
> Also a bit concerned that the positioning source is always running even if you aren't doing follow me. That's chewing cycles for no reason.

Good point, I can turn this off and on really easily with my enable/disable functions.

src/FollowMe/FollowMe.cc
catch-twenty-two: Yeah, I will go and add these. There are a bunch that I didn't add because I wasn't using them. I won't be able to add all of them yet, but hopefully I'll eventually I get around to all of them!

catch-twenty-two:
> Also don't understand how you offset the follow me position from the current GCS position?

Yeah, I don't it. Generally flies slightly offset because the gps on a phone isn't very accurate I also have the min height limit set to 10 meters for safety on the px4 side.  I can create an offset for now to make it always slightly behind the target.

catch-twenty-two:
> Given the mavlink message I don't quite see how you would implement all of these follow me options: https://www.youtube.com/watch?v=LM1klzuC6Xw


Yeah, I figured this was a bit of a proof of concept just to get something working (baby steps).  We can offset to the right or behind pretty easily with the current messages.  It will just require more math.  Circle mode is probably a little different though.

catch-twenty-two:
> Related to the follow me flight mode discusion. How are MAV_CMD_DO_FOLLOW and MAV_CMD_DO_FOLLOW_REPOSITION supposed to be used? Are those deprecated?

I had the initial code working before these were defined by defining custom px4 messages.  These can definitely be used I thinking in the next iteration though.

catch-twenty-two:
@DonLakeFlyer My over all thoughts are I think getting something working for now is a good step.  But I also realize there are a ton of long term goals that need to be met.  I didn't know that QGC was shooting to be Arducopter compatible as well so I wasn't thinking about that. If that's the case as the feature matures I'll probably need some guidance so we can make everything match up correctly.

LorenzMeier:
@DonLakeFlyer This message is not controlling the vehicle. Its just sending the target data. It's the vehicle's job to do the offset and flight mode control. This is just a data source and entirely MAVLink generic.

DonLakeFlyer:
> It's the vehicle's job to do the offset and flight mode control.

So in order to implement something like the "Right" follow me mode from Tower (it follow the user onthe right side, pointed at the user) the firmware needs some additional set of parameters or something? How is that part going to be mavlink generic? Use mavlink message instead of parameters?

Also, how is QGC supposed to know when to send follow me targets to a Vehicle in a mavlink generic way?

DonLakeFlyer:
Hmm, linux unit test is passing for me locally as well.

DonLakeFlyer:
Looks like it is crashing right away when it tries to run the unit test. Other pulls are going through ok, so it's something to do with your changes. These sorts of things are a pain to debug since they ony repro in travis. One way to do it is to turn on full logging for the builds. You can hack this in by changed QGCApplication. Then you can see from the logs how far you are getting into the app. Then you add other qDebug to binary search your way to the problem. Primitive, time suck, but it works.

catch-twenty-two:
> Looks like it is crashing right away when it tries to run the unit test. Other pulls are going through ok, so it's something to do with your changes. These sorts of things are a pain to debug since they ony repro in travis. One way to do it is to turn on full logging for the builds. You can hack this in by changed QGCApplication. Then you can see from the logs how far you are getting into the app. Then you add other qDebug to binary search your way to the problem. Primitive, time suck, but it works.

Okay thanks @DonLakeFlyer  ill see what I can do.

catch-twenty-two:
@DonLakeFlyer looks like I found the issue.  This pr should be good to go now.

DonLakeFlyer:
Can you squash this down into a single commit? Then I'll merge it in and the I'll figure out how to change this to be supported through some sort of plugin mechanism so this implementation will only work for PX4 side of things.

@LorenzMeier Still not sure how this is going to be mavlink generic. Is there some discussion on how to finish out that concept? Checking a flight mode breaks the mavlink generic implementation.

catch-twenty-two:
> Can you squash this down into a single commit? Then I'll merge it in and the I'll figure out how to change this to be supported through some sort of plugin mechanism so this implementation will only work for PX4 side of things.

Done

PR #3079, opened 2016-03-25
[WIP] Link thread safety enforcement

NaterGator:
We are having issues with various types of link behaving improperly when they are threaded.

This is an attempt to clean up some of the rough spots where the link interfaces are interacted with in a thread unsafe manner.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -733,7 +733,7 @@ void Vehicle::_sendMessageOnLink(LinkInterface* link, mavlink_message_t message)
     uint8_t buffer[MAVLINK_MAX_PACKET_LEN];
     int len = mavlink_msg_to_send_buffer(buffer, &message);
 
-    link->writeBytes((const char*)buffer, len);
+    link->writeBytesSafe((const char*)buffer, len);
     _messagesSent++;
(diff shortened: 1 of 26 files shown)


DonLakeFlyer:
A little better restructuring would be to leave public writeBytes the same. But add new virtual _writeBytesOnThread private which is a replacement for the existing public virtual writeBytes. Then writeBytes signals to itself, with slot calling _writeBytesOnThread virtual. That way nobody calls wrong public writeBytes by accident. This is how I've structured this pattern in other places.

NaterGator:
I've used that pattern too, but wanted to make sure I caught all the existing instances by marking the writeBytes symbol private. Another caveat is it obfuscates that there are potential threading considerations here; this interface is unlikely to see high throughput or high bandwidth use cases, but you wouldn't want users oblivious to the fact that they are potentially invoking queuing or memcopy behavior if they're planning on streaming video or some other particularly intensive operation.

Your call, if you prefer the public method to remain writeBytes with proxying to the signal/slot via new methods I am happy to refactor the names (once I see CI finish / indicate I've caught all the writeBytes users)

DonLakeFlyer:
Ah, I missed writeBytes going private. Can you switch to _writeBytes? That’s good enough.

PR #3089, opened 2016-03-27
adding follow me

catch-twenty-two:
replaces https://github.com/mavlink/qgroundcontrol/pull/3070


LorenzMeier:
@DonLakeFlyer This is ready to go. Firmware support is in and the changes here should not affect existing systems, while still being MAVLink-generic.

src/FollowMe/FollowMe.cc
```diff
@@ -0,0 +1,218 @@
...
+
+#ifdef QT_QML_DEBUG
+    if(_simulate_motion == true) {
+        _createSimulatedMotion(follow_target);
+    }
+#endif
+
+    for (int i=0; i< vehicles.count(); i++) {
+        Vehicle* vehicle = qobject_cast<Vehicle*>(vehicles[i]);
+        if(vehicle->flightMode().compare(_followMeStr, Qt::CaseInsensitive) == 0) {
```
DonLakeFlyer: @LorenzMeier Just to be clear, this is the line which makes this not mavlink generic. Flight modes are firmware specific. I've added this to my list for discussion and we can talk about then. I'm ok with merging and I'll figure out a way to fix it up so it doesn't affect APM side of things.

DonLakeFlyer:
Also FYI: I'm waiting to merge this until after my travis OSX changes go through. I've been fighting with that for days now (OSX daily builds are broken). Don't want anything to hold it up.

src/FollowMe/FollowMe.cc
catch-twenty-two: hey @DonLakeFlyer I'm still confused as to why this is not a generic message.  Let me know when you plan to discuss and Ill try and be there.  I can help integrate this too if you want to give me an outline of what needs to be done I can make the changes.  I'm planning on making a lot more improvements to the tracking ability of the uav through qgc so the more I understand the direction of the architecture towards APM and PX4 the better.

DonLakeFlyer: Flight modes and flight mode names are firmware specific. The set of flight modes, their names and exactly what they do are opaque to QGC and differ between firmware stacks. This means you can't check for a specific flight mode like you are doing without breaking the mavlink generic rule. This needs to be moved to a Firmware plugin mechanism. I'm working on this right now. I'll include you in the pull so you can see what it ends up looking like.

DonLakeFlyer: Also just to be clear: Moving to a FirmwarePlugin mechanism is also a workaround which just clean up QGC central code from a mavlink generic standpoint. There needs to be some sort of mavlink message support to put a vehicle in and out of follow me mode. That is the real solution. Then the whole thing is mavlink generic end to end.

Issue #3015, opened 2016-03-13
(Linux) AudioOutput: espeak_Synth not triggered


shadow1runner:
Even though [emit textToSpeak(inText);](https://github.com/mavlink/qgroundcontrol/blob/dcd55081b4a5fbd5af3d6d03d77aee1394ce6009/src/GAudioOutput.cc#L107) is triggered perfectly fine (e.g. on mode change), [espeak_Synth](https://github.com/mavlink/qgroundcontrol/blob/dcd55081b4a5fbd5af3d6d03d77aee1394ce6009/src/audio/QGCAudioWorker.cpp#L154) is not triggered after the first audio output which using `sim_vehicle.sh -w` is "stabilize flight mode"). I haven't digged any further (yet), but the signal/slot connection should be there (as it's done in the ctor), but no breakpoint on `espeak_Synth` is hit after the first time.

Note that it's not only master, but also the latest release (2.9.4), here it says the audio output is "System one is now in stabilize flight mode and standing by".

DonLakeFlyer:
I changed voice output recently with respect to flight modes. I may no longer trigger a flight mode announce on first vehicle connect. Although it should.

shadow1runner:
Thanks for the fast response. I can confirm that it announces the flight mode on initial connect, but that's the first and _last_ message ever heard from QGC.

DonLakeFlyer:
@dagar You're a Linux person right? I've never been able to get speech working at all on my Ubuntu VM. Do you have time to look at this?

dagar:
I can confirm it stops for some reason. At this point it definitely works everywhere else?

DonLakeFlyer:
Just tested on OSX with a couple flight mode changes. All were announced.

shadow1runner:
And I've just tried under Windows (2.9.3 though), all worked fine.

Note that the application output after each `textToSpeak` emitting show the following line:
```
ALSA lib pcm.c:7905:(snd_pcm_recover) underrun occurred
```

NaterGator:
Sounds like espeak_Synth is never returning. Can you get a backtrace on the speech thread after it has blocked and stopped working?

shadow1runner:
Indeed, step over `espeak_Synth` never returns, here's the whised backtrace of the espeak thread:

```
Thread 9 (Thread 0x7fffc179c700 (LWP 17893)):
[...]

PR #3099, opened 2016-03-28
Fix for espeak / alsa errors for #3015

NaterGator:
Initialize espeak in async playback mode and explicitly synchronize after each synth operation in the worker thread.


DonLakeFlyer:
Thanks Nate

PR #3109, opened 2016-03-29
Change Search Field for Parameter List

tcanabrava:
This way the search field is always visible, there's no
need to add two buttons ( search / cancel ) to it, it has
live filtering (it will filter while you type) instead of
going to the tools menu (1 click), clicking on the search
(1 click), typing what you whant to filter, clicking on
accept(1 click) for each new letter that you wanna add
to the filter.

### DIFF
--- a/src/QmlControls/ParameterEditor.qml
+++ b/src/QmlControls/ParameterEditor.qml
@@ -101,11 +101,36 @@ QGCView {
                     anchors.right:  parent.right
                     height: ScreenTools.defaultFontPixelHeight * 1.75
                     onClicked: {
-                        _searchFilter = false
-                        hideDialog()
+                        searchFor.text = ""
                     }
                 }
+
+                QGCLabel {
+                    font.weight: Font.DemiBold
+                    text:   "Filter By:"
+                    anchors.right: searchFor.left
+                    height: ScreenTools.defaultFontPixelHeight * 1.75
+                }
+
+                QGCTextField {
+                    id:                 searchFor
+                    anchors.topMargin:  defaultTextHeight / 3
+                    anchors.right:      toolsButton.left
+                    width:              ScreenTools.defaultFontPixelWidth * 20
+
+                    onTextChanged: {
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Thanks @tcanabrava. I want to run this and take look before I merge. Probably not till tomorrow since I'm heading of town for the day.

DonLakeFlyer:
* Can you add spacing between the controls: ScreenTools.defaultFontPixelWidth
* Set the baseline for all three such that the text lines up

screen shot 2016-03-29 at 11 57 33 am

Do you know how/if this works on mobile if you have a keyboard text input thingy?

tcanabrava:
@DonLakeFlyer  - as you requested, spacing and baseline. :)

DonLakeFlyer:
Working on testing this on mobile. Take me a little while. XCode just updated with a new compiler which is spitting out new errors. Need to fix those first.

DonLakeFlyer:
I like the feature a lot but it doesn't work on mobile or small screens. The list generation is just barely fast enough for desktop builds. On mobile it's way too slow. Click a letter, wait two seconds, click another letter wait another two seconds. Also it doesn't fit onto small screens like a phone:
img_0988
img_0989

So the possibility here is to make this be desktop only and leave mobile the old way. Having them both be the same won't work without some major performance improvements as well as a different ui.

tcanabrava:
I'll change it to desktop only for now, but I'll take a look on how to
improve speed on mobile too. This shouldn't take too long on mobile, it's
just a list.

Tomaz
Em 31 de mar de 2016 21:44, "Don Gagne" <notifications@github.com> escreveu:

tcanabrava:
Don, I created another branch where I tried to fix the speed of the parameter list (on dekstop it improved like 2x, it can be faster on slower devices), I still didn't fixed the positioning of the things on mobile, because I wanted you to test if the speed is okay for it (I don't have a tablet to test). I'm closing this one.

PR #3119, opened 2016-03-30
Parameter Edit: Handle bitfields as CheckBoxes

tcanabrava:
spectacle lh6179

This makes it possible to select more than one value
to be used in the parameter configuration, and since
the parameter is actually a bitfield - this is the
correct thing to do.

one of the drawbacks is that in the parameter listing
the value of the Fact will not be displayed as string
anymore, but with 'undefined: numerical-value', that is
the OR combination of the applied values.

### DIFF
--- a/src/QmlControls/ParameterEditorDialog.qml
+++ b/src/QmlControls/ParameterEditorDialog.qml
@@ -47,7 +47,18 @@ QGCViewDialog {
     QGCPalette { id: qgcPal; colorGroupEnabled: true }
 
     function accept() {
-        if (factCombo.visible) {
+        if (bitmaskEditor.visible) {
+            var value = 0;
+            for (var i = 0; i < fact.bitmaskValues.length; ++i) {
+                var checkbox = bitmaskEditor.itemAt(i)
+                if (checkbox.checked) {
+                    value |= fact.bitmaskValues[i];
+                }
+            }
+            fact.value = value;
+            fact.valueChanged(fact.value)
+        }
+        else if (factCombo.visible) {
             fact.enumIndex = factCombo.currentIndex
             hideDialog()
         } else {
@@ -127,13 +138,24 @@ QGCViewDialog {
                 visible:        _showCombo
                 model:          fact.enumStrings
 
-                property bool _showCombo: fact.enumStrings.length != 0 && !validate
+                property bool _showCombo: fact.enumStrings.length != 0 && fact.bitmaskStrings.length == 0 && !validate
 
                 Component.onCompleted: {
                     // We can't bind directly to fact.enumIndex since that would add an unknown value
                     // if there are no enum strings.
                     if (_showCombo) {
                         currentIndex = fact.enumIndex
+                    }
+                }
+            }
+
+            Column {
+                Repeater {
+                    id: bitmaskEditor
+                    model: fact.bitmaskStrings
+                    delegate : QGCCheckBox {
+                        text : modelData
+                        checked : fact.value & fact.bitmaskValues[index]
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
The screenshot shows a bad example though... "All" would select all.

tcanabrava:
On Thu, Mar 31, 2016 at 3:11 PM, Lucas De Marchi <notifications@github.com>
wrote:

> The screenshot shows a bad example though... "All" would select all.
>

Done, pushed.

src/QmlControls/ParameterEditorDialog.qml
```diff
@@ -156,6 +156,15 @@ QGCViewDialog {
                     delegate : QGCCheckBox {
                         text : modelData
                         checked : fact.value & fact.bitmaskValues[index]
+                        function check_uncheck_all() {
+                            if (text.toLowerCase() == "all") {
```
DonLakeFlyer: This is a bit too much of a special case to do without knowing this is correct from the meta data. "All" may not always mean this.

src/QmlControls/ParameterEditorDialog.qml
```diff
@@ -138,6 +149,26 @@ QGCViewDialog {
...
 
+            Column {
+                Repeater {
+                    id: bitmaskEditor
+                    model: fact.bitmaskStrings
+                    delegate : QGCCheckBox {
+                        text : modelData
+                        checked : fact.value & fact.bitmaskValues[index]
+                        function check_uncheck_all() {
+                            if (text.toLowerCase() == "all") {
```
DonLakeFlyer: This is a bit too much of a special case to do without knowing this is correct from the meta data. "All" may not always mean this.

DonLakeFlyer:
Why can't this work like FactBitmask.Qml? Is there something wrong with that mechanism? You could make a FactBitmaskRow and a FactBitmaskColumn. Only difference would need to be outer Row/Column holder.

tcanabrava:
FactBitmask.qml applies the valie when the user clicks on it, not when the user clicks on 'Save', thus making it behave differently from all of the other boxes. If you want it to apply as the user clicks, I can change it to use FactBitmask. what you think?

src/QmlControls/ParameterEditorDialog.qml
tcanabrava: okay, will remove this option. @lucasdemarchi  - comments, since you requested this?

lucasdemarchi: What I asked is that when you select "all" it actually selects all the values. I suggested a strcmp, but @DonLakeFlyer is right that it's too generic. Maybe you can do the same but looking at the value instead?

tcanabrava: Nope. The value is "1", that's actually more generic.
Em 4 de abr de 2016 21:05, "Lucas De Marchi" <notifications@github.com>
escreveu:

DonLakeFlyer:
> FactBitmask.qml applies the valie when the user clicks on it,

Ah, right. Forgot about that. This is fine then.

src/QmlControls/ParameterEditorDialog.qml
lucasdemarchi: is the strcmp() done only for this param?  I don't think we would could do anything smarter; right thing IMO would be to fix the value in ardupilot so "all" is actually all. However this means we would be changing the API

DonLakeFlyer: I don't think there is going to be a way to do this. There isn't enough informatio in the meta data.

tcanabrava:
Created a new PR with the new code.

PR #3124, opened 2016-03-31
WIP: Try to fix positioning of items

tcanabrava:
qml-error

This is actually a test - I didn't managed to properly fix it
because it seems I'v hit a Qt bug. I'd appreciate any help.

The first row of items are correctly aligned, but the others, help me god.


DonLakeFlyer:
Since it's recursive this control is a real pain. Also I don't think the sliders instead of check boxes add any value. We don't use the control anywhere else. It should stick to a checkbox. Much better to stay consistent.

DonLakeFlyer:
I tried doing exactly what you are doing (alignment) a while back with the same results. In reality this whole thing needs a re-write bot in the Qml and the Controller. The current implementation is jsut a quick hack which isn't very changable.

PR #3126, opened 2016-04-01
Add search v2

tcanabrava:
Change the Way Parameter Filtering is done
    
    The old ParameterFiltering worked by retrieving a list of
    QStrings that where the name of the Facts of the vehicle,
    but to search for the names we actually got the Fact*, since
    we needed to check it's name() and other internal methods.
    
    Then we passed the StringList to QML, to again retrieve the
    Facts* that it represented.
    
    getting a Fact* from the controller is a slow operation that
    involves quite a few function calls and conversions, QML
    doesn't really like that, so the minimum calls, the better.
    
    This new way we return a QList<QObject*> instead of a QString
    List, that's filled with the Facts* from the call to the
    search function.
    
    On my tests this speeded up the result into 2, 2.5x because
    it: does less comparissons, create less temporaries and reuse
    the Facts searched.

Didn't tested this on mobile because I don't have a Tablet, but I think this should speed up things on iPad

### DIFF
--- a/src/FactSystem/ParameterLoader.cc
+++ b/src/FactSystem/ParameterLoader.cc
@@ -461,13 +461,7 @@ Fact* ParameterLoader::getFact(int componentId, const QString& name)
 
 QStringList ParameterLoader::parameterNames(int componentId)
 {
-    QStringList names;
-
-    foreach(const QString &paramName, _mapParameterName2Variant[_actualComponentId(componentId)].keys()) {
-        names << paramName;
-    }
-
-    return names;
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
Bummer, nope. Still really slow. Actually seems slower than before. This is going to need to be desktop only change.

DonLakeFlyer:
I think switching this whole thing to QmlObjectListMode might be needed to make it fast enough. Doable, but somewhat involved.

PR #3129, opened 2016-04-02
Make it possible to run qgroundcontrol-start.sh from outside its directory

probonopd:
This patch makes it possible to run qgroundcontrol-start.sh from outside its directory, e.g.,
`/qgroundcontrol/qgroundcontrol-start.sh` instead of having to `cd qgroundcontrol/ && ./qgroundcontrol-start.sh`.


dagar:
Tested and works for me. I still hate that we need this script. I really need to devote some time to getting AppImage working and in the builds.
Thanks @probonopd

PR #3138, opened 2016-04-04
Enhancement.position manager.wip

catch-twenty-two:
WIP do not merge, adds an abstraction to the GPS information read by the system.  Any type inheriting from QGeoPositionInfoSource can inject it's GPS data into the system via this new class(simulate qgc movement, add log play back, use an external gps, ect...) . This also removes all simulated position methods from the Follow me class and puts them in their own class.


mhkabir:
Nice architectural cleanup! 
Do you have any suggestions as to where we can best implement a IMU aided state estimator?

catch-twenty-two:
@mhkabir thanks, I figured it would go in it's own class somewhere, and could create a connection for updating whatever is interested in it. (the follow me class for instance) @LorenzMeier said he had tried this before and it was difficult because hardware varies from device to device so much, it makes it difficult to tune it for every phone.

Issue #3121, opened 2016-03-31
remove windows cmd prompt


dagar:
Can we get rid of launching with the cmd prompt?

![screenshot 2016-03-31 09 46 56](https://cloud.githubusercontent.com/assets/84712/14177752/c3d82e2c-f725-11e5-9db6-85804acd4827.png)

DonLakeFlyer:
If we can figure out some other way to get logging information when needed

NaterGator:
This drives me crazy. I'll happily fix it.

DonLakeFlyer:
Thanks Nate.

PR #3139, opened 2016-04-04
Add optional qml-based console to view debugging messages. For #3121

NaterGator:
As a side effect this also fixes those annoying messages during close such as:
`default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:85 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:85: TypeError: Cannot read property 'name' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:101 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:101: TypeError: Cannot read property 'valueString' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:92 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:92: TypeError: Cannot read property 'shortDescription' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:85 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:85: TypeError: Cannot read property 'name' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:101 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:101: TypeError: Cannot read property 'valueString' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:92 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:92: TypeError: Cannot read property 'shortDescription' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:85 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:85: TypeError: Cannot read property 'name' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:101 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:101: TypeError: Cannot read property 'valueString' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:92 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:92: TypeError: Cannot read property 'shortDescription' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:85 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:85: TypeError: Cannot read property 'name' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:101 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:101: TypeError: Cannot read property 'valueString' of null"
default [E] at qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:92 - "qrc:///qml/QGroundControl/FlightMap/ValuesWidget.qml:92: TypeError: Cannot read property 'shortDescription' of null"`
and "External WM_DESTROY received for QWidgetWindow". 

The fix for those is [here](https://github.com/mavlink/qgroundcontrol/commit/c51a1a6ec98519b0c1548fdd491a3ad2980876de#diff-cf66561df07c165223a1ccfd096e2406L424), MainWindow was never properly destroyed due to the line.


NaterGator:
This is how the debug console dialog looks:
image

qgroundcontrol.pro
```diff
@@ -294,7 +297,7 @@ HEADERS += \
     src/AutoPilotPlugins/APM/APMAirframeLoader.h \
     src/QmlControls/QGCImageProvider.h \
     src/AutoPilotPlugins/APM/APMRemoteParamsDownloader.h \
-    src/QtLocationPlugin/QMLControl/QGCMapEngineManager.h \
+    src/QtLocationPlugin/QMLControl/QGCMapEngineManager.h
```
DonLakeFlyer: Is there a reason to remove these? It makes adding new entries more of a pain.

DonLakeFlyer:
@NaterGator Anyway to to tweak this a little bit such that the debug window is a Settings page instead. Also if it could be in release builds. Then we could use this to get console output on Mobile builds as well. It's a problem right now that you can't get this.

NaterGator:
Yes, I'll go ahead and make those tweaks. I also think it would be useful to have a "Save to file" button that saves the current message buffer to a text file. It is already in release builds, that was my primary goal.

DonLakeFlyer:
> I also think it would be useful to have a "Save to file" button that saves the current message buffer to a text file. It is already in release builds, that was my primary goal.

That would be great as well for getting reports from users.

NaterGator:
The android build is force closing when trying to navigate to the console. Not ready for primetime yet. This is how things look on Windows: 
image

DonLakeFlyer:
Let me know if you need me to help with anything on android.

NaterGator:
Two things about the force close...

1. The actual message is `QQmlEngine: Illegal attempt to connect to AppLogModel(0xdfde7a80) that is in a different thread than the QML engine QQmlApplicationEngine(0xdcdd3fe0)`
2. The force close is triggered because we're installing the message handler on all platforms, and that message is considered Fatal, [which we abort on](https://github.com/NaterGator/qgroundcontrol/blob/bc19757e86b8643bc7164ea0eb6355a9b003a7ab/src/QmlControls/AppMessages.cc#L49)

Not sure if we should abort or let Qt's default handler manage the fatal message, especially on mobile.

NaterGator:
This is how things currently look on Android...
image

If I'm nitpicking maybe could a bit more padding on the buttons...

DonLakeFlyer:
Yeah the tolerance is a bit small on the edges. There is a lot a variability in font information from device to device. On my Samsung it tends to look fine. Makes it crazy to come up with something that is great everywhere.

DonLakeFlyer:
Is this all good to go now Nate?

NaterGator:
Do you want me to add some padding for the buttons? If not it should be good to go now.

DonLakeFlyer:
No. Then it will just get screwed up if I change the default padding in the control. You're not setting the width, so the default is good for now. We may need to double check the QGCButton implementation for padding.

DonLakeFlyer:
Thanks Nate. This is a good add. I can now add a log selection ui to it and we have a nice debugging console.

NaterGator:
No problem. I thought about adding category / severity filters (would be fairly trivial with a QSortFilterProxyModel) but figured that deserves a feature ticket after this gets some testing. The signaling mostly takes care of it, but we need to be aware of potential issues with reentrancy / recursion in the message handler when logging is cranked way way up (ex: messages from QStringListModel/QML updates could cause stack overflow). In practice I haven't had it happen but if users notice issue with absurd log levels enabled it is something to keep in mind.

Issue #3150, opened 2016-04-05
Gear icon in instrument panel needs to follow palette


DonLakeFlyer:

DonLakeFlyer:
Same problem with Vibe panel

DonLakeFlyer:
Sizing isn't working in value widget when you add values.

PR #3155, opened 2016-04-06
Fix gearThingy

dogmaphobic:
Fix for #3150. Make sure gear icon follows the palette in use.


DonLakeFlyer:
This is the key to my entire demo!

Issue #3114, opened 2016-03-29
Always show feedback on param refresh


julianoes:
Currently, when you click on "Refresh params", the feedback can be unclear:

If the params really have been updated, you see a nice progress bar while it downloads them. However, if the hash is up-to-date, and no refresh is needed, nothing happens and the user doesn't know if everything is up-to-date or if it just didn't work. 

What about adding a fake progress bar for half a second to imitate that something is happening?

PR #3161, opened 2016-04-07
Show the user some positive feedback when parameters load from cache.…

NaterGator:
… For #3114.

### DIFF
--- a/src/FactSystem/ParameterLoader.cc
+++ b/src/FactSystem/ParameterLoader.cc
@@ -32,8 +32,10 @@
 #include "FirmwarePlugin.h"
 #include "UAS.h"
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Awesome Nate, thanks.

julianoes:
Yay, thanks!

NaterGator:
A quick note on this: we'll need to test with multiple components reporting their params... right now I suspect the progress bar may spaz out / flicker for a second if multiple devices / components sync params at the same time.

DonLakeFlyer:
> I suspect the progress bar may spaz out / flicker for a second if multiple devices / components sync params

I think it did that before as well

PR #3173, opened 2016-04-10
Test if connection type on signal fixes this

LorenzMeier:
@DonLakeFlyer Educated / frantic guess. Would be awesome if you could test!


DonLakeFlyer:
No improvement.

PR #3176, opened 2016-04-10
Reenable writing to stderr on *nix platforms for useful debugging

NaterGator:


src/main.cc
```diff
@@ -114,6 +114,12 @@ jint JNI_OnLoad(JavaVM* vm, void* reserved)
 
 int main(int argc, char *argv[])
 {
+#ifdef Q_OS_UNIX
+    //FOrce writing to the console on UNIX/BSD devices
```
NaterGator: FAcepalm. @DonLakeFlyer  if this passes CI do you mind fixing this comment before merging to master? I hate triggering another CI build to fix one incorrectly cased letter.

DonLakeFlyer:
That would probably take some git gymnastics which I'm not capabable of! I'll throw it into of my other pulls down the road.

DonLakeFlyer:
That would probably take some git gymnastics which I'm not capabable of! I'll throw it into of my other pulls down the road.

PR #3196, opened 2016-04-11
Parameter Edit: Handle bitfields as ComboBoxes

tcanabrava:
This makes it possible to select more than one value
to be used in the parameter configuration, and since
the parameter is actually a bitfield - this is the
correct thing to do.

one of the drawbacks is that in the parameter listing
the value of the Fact will not be displayed as string
anymore, but with 'undefined: numerical-value', that is
the OR combination of the applied values.

### DIFF
--- a/src/QmlControls/ParameterEditorDialog.qml
+++ b/src/QmlControls/ParameterEditorDialog.qml
@@ -47,7 +47,18 @@ QGCViewDialog {
     QGCPalette { id: qgcPal; colorGroupEnabled: true }
 
     function accept() {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
screen shot 2016-04-13 at 4 19 50 pm

Can you add ScreenTools.defaultFontPixelHeight / 2 spacing in between the check boxes.

tcanabrava:
Sure.
Em 13 de abr de 2016 20:21, "Don Gagne" <notifications@github.com> escreveu:

PR #3203, opened 2016-04-12
Skip "non compliant" Qt libraries (for Mac OS App Store)

dogmaphobic:
These are libraries added by `macdeployqt`, which use non public Mac OS API and are therefore *Non Compliant*. They are not used by QGC any way.

Note that you will still see the warning during the build accompanied with the message these libraries are being skipped:
```
WARNING: Plugin "libqsqlodbc.dylib" uses private API and is not Mac App store compliant.
WARNING: Skip plugin "libqsqlodbc.dylib"
WARNING: Plugin "libqsqlpsql.dylib" uses private API and is not Mac App store compliant.
WARNING: Skip plugin "libqsqlpsql.dylib"
```

### DIFF
--- a/QGCInstaller.pri
+++ b/QGCInstaller.pri
@@ -37,7 +37,7 @@ installer {
         # qgroundcontrol.app file. If you specify a path to the .app file the symbolic
         # links to plugins will not be created correctly.
         QMAKE_POST_LINK += && mkdir -p $${DESTDIR}/package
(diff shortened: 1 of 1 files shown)


dogmaphobic:
AppVeyor must be having a bad hair day. Or it just doesn't like the Windows installer tool thingy and is giving you a hint to use something else :P

DonLakeFlyer:
No it's just an appveyor thing. Daniel is going to change it so this stops ahppening.

PR #3209, opened 2016-04-12
Fix windows build bug, about MissionItemTest. Fix Survey delete bug.

haigithub:
My development environment is Windows 7, Qt5.5.1(x86), VS2013. There are always building error about MissionItemTest, after I git clone QGC. When the source code assignes values to the objects of QJsonAray or QJsonObject, building errors appear. After I study the implementation of the class QJsonArray and QJsonObject, I find the solution.

### DIFF
--- a/src/MissionEditor/MissionEditor.qml
+++ b/src/MissionEditor/MissionEditor.qml
@@ -485,7 +485,7 @@ QGCView {
 
                             onRemove: {
                                 itemDragger.clearItem()
-                                controller.removeMissionItem(object.sequenceNumber)
+                                controller.removeMissionItem(index)
                             }
 
                             onInsert: {
--- a/src/MissionManager/MissionItemTest.cc
+++ b/src/MissionManager/MissionItemTest.cc
@@ -281,20 +281,20 @@ void MissionItemTest::_testLoadFromJson(void)
 {
     MissionItem missionItem;
     QString     errorString;
-    QJsonArray  coordinateArray = { -10.0, -20.0, -30.0 };
-    QJsonObject jsonObject
-    {
-        { "autoContinue",   true },
-        { "command",        80 },
-        { "frame",          3 },
-        { "id",             10 },
-        { "param1",         10 },
-        { "param2",         20 },
-        { "param3",         30 },
-        { "param4",         40 },
-        { "type",           "missionItem" },
-        { "coordinate",     coordinateArray },
-    };
+    QJsonArray  coordinateArray;
+    coordinateArray << -10.0 << -20.0 <<-30.0;
+    QJsonObject jsonObject;
+    jsonObject.insert("autoContinue", true);
+    jsonObject.insert("command", 80);
+    jsonObject.insert("frame", 3);
+    jsonObject.insert("id", 10);
+    jsonObject.insert("param1", 10);
+    jsonObject.insert("param2", 20);
+    jsonObject.insert("param3", 30);
+    jsonObject.insert("param4", 40);
+    jsonObject.insert("type", "missionItem");
+    jsonObject.insert("coordinate", coordinateArray);
+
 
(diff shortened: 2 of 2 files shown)


DonLakeFlyer:
What's the issue with Survey delete?

src/MissionEditor/MissionEditor.qml
```diff
@@ -484,8 +501,7 @@ QGCView {
                             onClicked:  setCurrentItem(object.sequenceNumber)
 
                             onRemove: {
-                                itemDragger.clearItem()
-                                controller.removeMissionItem(object.sequenceNumber)
```
DonLakeFlyer: The fix for this bug is just to change this line to:
```
controller.removeMissionItem(index)
```

src/MissionEditor/SimpleItemEditor.qml
```diff
@@ -119,6 +119,13 @@ Rectangle {
                 onClicked:  editorRoot.moveHomeToMapCenter()
                 anchors.horizontalCenter: parent.horizontalCenter
             }
+
+            QGCButton {
```
DonLakeFlyer: There already is a center map tool on the Plan view. This show go there not in the item editors. The name should be "Current item".

src/MissionEditor/MissionEditor.qml
haigithub: I have tested. After changing the code from "controller.removeMissionItem(object.sequenceNumber)" to "controller.removeMissionItem(index)", deleting multi-mission of survey works well

src/MissionEditor/SimpleItemEditor.qml
haigithub: After changing latitude and longitude home position, the map center also stays in the old view. But I want change view to the new home position. So I add this Button for speeding up view changing.

DonLakeFlyer: Oh, I thought this was centering around current item. Sorry not looking close enough.  If you want to center around home just use this:
screen shot 2016-04-13 at 8 44 23 am

haigithub: Aha, it has been finished. That is what I want. Thank you very much.

src/MissionEditor/MissionEditor.qml
haigithub: The issues with Survey delete is that there is an out of range error when I delete one of some surveys which have been edited out of order. I'm sure that it is crashing due to incorrect index.

DonLakeFlyer:
Great, so update the pull to the comments. I'll review again and the merge when the changes pass. Thanks.

haigithub:
@DonLakeFlyer Do you notice the "Fix windows build bug, about MissionItemTest". After I download the source with changing nothing. I build the source code on Windows 7 platform and then encounter build errors. Directly assigning values to the object of classes QJsonArray and QJsonObject fails. Then I pull and request "Fix windows build bug, about MissionItemTest".

DonLakeFlyer:
Not quite sure what you mean? The changes to Json object creation are fine. But you need to update your pull to only include the changes for the json creation and the modified index delete problem. Also can you squash the commits down to fix up the commit messages.

haigithub:
@DonLakeFlyer Do you mean that I combine the commits "Fix windows build bug, about MissionItemTest" and "Fix Survey delete bug." to a commit ?

DonLakeFlyer:
Sorry, my comments are incorrect. The "fix windows build bug" commmit is fine. The "Fix survey delete" commmit needs to be modified to the correct one line change as described in the comments. The "move map center to home" should be remove since it is already implement. May be easier to start over with a new pull/commits than modify this one.

haigithub:
@DonLakeFlyer  OK. I understand.

Issue #2922, opened 2016-03-02
Planning map screen


lvale:
I would suggest to make it easier to plan and follow a mission that either a small text label would be placed near the map point. If taken to the latest consequences we could deprecate the right side mission items bar and clicking on the mission item on the map would open the "configuration" for the point.

<img width="1438" alt="screen shot 2016-03-02 at 23 20 11" src="https://cloud.githubusercontent.com/assets/3130363/13479568/c2824790-e0cf-11e5-8e1b-c5f212399b13.png">

DonLakeFlyer:
Maybe something along these lines fro 3.0

tubeme:
Very nice idea i must say. The current implementation does not work well for missions with more than 10-15 points depending on the screen size. If you have a mission with 50 points as I did it it nightmare, scrolling down the waypoint list, the same time scrolling the map screen... BTW I did try to edit a mission with 30 points in the OS X version... with just the mouse pad of the macbook and it was a disaster!!! 

If this suggestion is accepted we will no longer need to have the waypoint list, giving us more space for mission planning concentrating on landscape and mission objectives.

There is another problem editing the mission. When I change the altitude of a point it does not apply the same alt to the next point. So I have to manually change the altitude of most of the waypoints. Sometimes it remembers my last alt... I dont understand what is the principal...

We would love to have a kind of additional overlay view, like in the game SimCity, that has the same mission spots but somehow indicates the altitude of the mission and its point to point distance and cumulative distance beside every waypoint. This will give the operator a 3D image of the flight. If lets say by coloring from green to red the waypoints on the map and the links between them we can indicate the current altitude of the planned flight. 

What do you think?

DonLakeFlyer:
> When I change the altitude of a point it does not apply the same alt to the next point. So I have to manually change the altitude of most of the waypoints. Sometimes it remembers my last alt... I dont understand what is the principal...

Can you add a separate issue for this? It should be working.

DonLakeFlyer:
> If you have a mission with 50 points as I did it it nightmare, scrolling down the waypoint list, the same time scrolling the map screen

This should have been fixed as well. I'll put in issue for both these things (and above one).

tubeme:
[...]

PR #3213, opened 2016-04-12
Plan: Scroll list to current item

DonLakeFlyer:
* When you select a waypoint in the map the list scrolss to it. it not 100% correct, but it's better than before
* Scroll the item list on the right no longer scrolls the map underneath it

Related to #2922

### DIFF
--- a/src/MissionEditor/MissionEditor.qml
+++ b/src/MissionEditor/MissionEditor.qml
@@ -287,6 +287,7 @@ QGCView {
                         if (addMissionItemsButton.checked) {
                             var sequenceNumber = controller.insertSimpleMissionItem(coordinate, controller.visualItems.count)
                             setCurrentItem(sequenceNumber)
+                            editorListView.positionViewAtIndex(editorListView.count - 1, ListView.Contain)
                         } else {
                             editorMap.mapClicked(coordinate)
                         }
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Need to get jcooper to test on Surface once this goes through.

NaterGator:
I have a SP3, what needs testing?

DonLakeFlyer:
Need to validate that you can still scroll the waypoint item list on the right by dragging when the list grows off the screen.

NaterGator:
You can but it is kinda frustrating. The transparent portions are clipped and consequently sometimes you end up scrolling the basemap when you want to flick the waypoint list. 

I would suggest catching touches between the transparent gaps in the delegates to prevent that from happening. Otherwise this PR seems to work as expected, though it is worth noting that ListView.Contained is unreliable for the last item in the list.

DonLakeFlyer:
> ListView.Contained is unreliable for the last item in the list.

Yeah I know. That's the better than before comment, but not 100%.

> I would suggest catching touches between the transparent gaps in the delegates to prevent that from happening.

I used to have this: https://github.com/mavlink/qgroundcontrol/commit/4d5508c8281c9dc4d68a41cced557d03454076ec. It works with the mosuse. But took it out scrolling on the Surface all together. Do you think you could play around with this part alittle after this pull goes through. To find something that will work on the Surface?

DonLakeFlyer:
@NaterGator You ok with trying to figure out how to get scrolling to work without scrolling the map and working on Surface touch drag as well.

NaterGator:
I'll try to work on it in my spare time given I have one of the affected devices. Seems like this could be pretty nebulous though so I don't want to promise I'll crack it before 3.0

PR #3214, opened 2016-04-12
[WIP] add PX4 airspeed mode

dagar:
-waiting for https://github.com/PX4/Firmware/pull/3861

-verify PX4_CUSTOM_MAIN_MODE in sync before merging


DonLakeFlyer:
What about flight mode setup? Does that need to chagne as well?

dagar:
I was thinking about making this only part of the new mode switch config.

dagar:
This should probably wait until PX4/Firmware#3861 is accepted.

DonLakeFlyer:
Should I revert it? Also just looking at it. Why isn't the new mode added to the end? By adding in the middle doesn't it cause old/new version problems? New QGC running against old firmware is going to set RATTITUDE instead of the new flight mode.

dagar:
Yes it should be at the end, but ended up like this because PX4_CUSTOM_MAIN_MODE was missing RATTITUDE on one side.

PR #3215, opened 2016-04-12
Show GPS error message only on map view

rjehangir:
This hides the "No GPS Lock on Vehicle" message when the video stream is full screen.

I'm doing this for ArduSub, since ROVs don't have GPS but do use the video stream. This would also be helpful for people flying FPV with no GPS. 

If you disagree that the error should be hidden, then I would ask that we put it in a less conspicuous place.


DonLakeFlyer:
I think this is ok.

Issue #3056, opened 2016-03-22
CPU load 100% on disconnect


julianoes:
The CPU load seems to go to 100% for one core when a serial connection gets disconnected. It is independent of auto-connect settings. A restart resolves the problem, and gets it back down to 2.4%.

This is on 2.9.4 on Ubuntu.

DonLakeFlyer:
Disconnected by pulling USB?

julianoes:
Correct, using a Pixhawk.

DonLakeFlyer:
For me, when you click disconnect the CPU goes back to normal. Still not sure what is going on.

LorenzMeier:
People don't click disconnect. They pull the cable and then get angry because 15 mins later their laptop battery has drained like crazy. The disconnect button has a terrible user model. A user never connects anything, so why would he think of disconnecting something he never connected?

DonLakeFlyer:
I explained the reason for disconnect to you already in another issue.

LorenzMeier:
I understand and appreciate it. However, the various user issues resulting from it won't go away if I understand the design choice. They don't understand it and just see that the app pulls 100% CPU load.

One option would be to make the disconnect button graphically much more visible (yellow / orange) and to bring the application to the front (equivalent of "make it jump in the dock") if a disconnect happens.

If you expect users to click disconnect, then let's help them to get there. Right now its extremely easy to miss.

DonLakeFlyer:
You seem to think I'm not going to fix the CPU problem? You don't want to make Disconnect too annoying. It can come and go when you are running over WiFi with a weak signal.

LorenzMeier:
I guess that's the miscommunication, yes.

julianoes:
@DonLakeFlyer Can you just make the reconnect tries slower so it pulls less CPU? Or is there more to it?

DonLakeFlyer:
There is no reconnect trying going on. No such thing with autoconnect. Port is still open.

julianoes:
I see, do you know what it is that is cycling?

DonLakeFlyer:
Nope, not yet. Wondering if its in the serial port code somewhere.

DonLakeFlyer:
So far all things pointing to Qt SerialPort code which is bad since we don't ahve control over that.

LorenzMeier:
[...]

Issue #3072, opened 2016-03-24
Look into auto-disconnect when link goes away


DonLakeFlyer:
Need to figure out how to deal with flaky UDP connections as well as OS differences on usb cable pull signalling across OS.

DonLakeFlyer:
Some discussion here as well: #3056

NaterGator:
@DonLakeFlyer on Windows QSerialPort::linkError is emitted with ResourceError: http://code.qt.io/cgit/qt/qtserialport.git/tree/src/serialport/qserialport_win.cpp?h=v5.5.1#n740

DonLakeFlyer:
If I remember correctly the issue is that the signalling differs between OS. OSX sends aboutToClose whereas Windows does not. It was a while ago I looked at this though. Maybe OSX signals that as well.

DonLakeFlyer:
Still not quite sure what to do with UDP though. WiFi connections come and go quite easily.

DonLakeFlyer:
But in thinking about it communication lost is just missing heartbeat, whereas these are device level signals. So I can distinguish between the two: plug pull, lost communication.

NaterGator:
Correct, semantically and in terms of user experience they are different things. Unreliable Wifi / network drops is just an indication of unreliable link. USB cord pulled (or OS level IO failure) is a more severe indication of a link fault that will definitely not resolve itself without user intervention. 

My OS X-fu is a little weak but I believe it returns ENXIO if the character device disappears, in which case the same error code will fire: http://code.qt.io/cgit/qt/qtserialport.git/tree/src/serialport/qserialport_unix.cpp?h=v5.5.1#n869

I'd say you're on the right track.

PR #3219, opened 2016-04-13
Auto-Disconnect on cable pull

DonLakeFlyer:
* This will automatically disconnect the Vehicle is you pull the cable on the connection for it.
* It does not automatically disconnect if the link i still there but the heartbeat is lost. For that you still get "Communication Lost" and Disconnect button.

Fix for Issues #3056, #3072

### DIFF
--- a/src/comm/LinkInterface.h
+++ b/src/comm/LinkInterface.h
@@ -170,6 +170,9 @@ signals:
     void activeChanged(bool active);
     void _invokeWriteBytes(QByteArray);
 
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
Ok, folks you asked for it. Now I need some serious testing for this with:
* Direct USB connection to Pixhawk
* SiK Radio connection

@LorenzMeier Can you try this on OSX?
@dagar Linux?
@NaterGator Windows?

dagar:
Repeatedly plugging/unplugging a pixhawk in Ubuntu 14.04 works without CPU going to 100%.
I'll close #3056

Issue #3152, opened 2016-04-05
ArduPilot Copter Motor refactoring


lvale:
The new motor refactoring on ArduPilot Copter has deprecated some parameters and added a few. 
Also motor library input ranges from the odd -4500~+4500 range to a more reasonable -1 ~ +1 range
There is a conversion table here:
https://github.com/ArduPilot/ardupilot/commit/a9cda8b38459ca004d0704e46b1a84ea9e7ae845#diff-3bff77ba7f43877f55b1815090e19da3L1019

Current behavior of QGC
<img width="1427" alt="screen-shot-2016-04-05-at-23 14 19" src="https://cloud.githubusercontent.com/assets/3130363/14301038/6b2d7fd2-fb8c-11e5-945b-2138f2272fb6.png">

DonLakeFlyer:
The differences in parameters between various versions of ArduPilot is brutal to maintain for a Ground Station!

billbonney:
I already have two Extended tuning screens is AP2. pre AC3.3, AC3.3 and now AC3.4.

DonLakeFlyer:
I need to create some better architecture to handle all of this. One-off each time is just too crazy.

rmackay9:
Yeah, sorry for the hassles.  We try to make it easy for the GCSs but sometimes we have to break some eggs to bake a cake.  In this case we moved the PIDs into the attitude control library which makes it more modular so it can be more easily swapped out in favour of another (perhaps non-PID) controller.  At the same time we fixed two legacy scaling issues.  So really, it's 3 changes for the price of one!

Old	New	Scaling change (i.e. multiply old value by below value to arrive at new value)
STB_RLL_P	ATC_ANG_RLL_P	none
STB_PIT_P	ATC_ANG_PIT_P	none
STB_YAW_P	ATC_ANG_YAW_P	none
		
RATE_RLL_P	ATC_RAT_RLL_P	0.9 (for x-frame multicopters) or 1.27
RATE_RLL_I	ATC_RAT_RLL_I	1/4500
RATE_RLL_IMAX	ATC_RAT_RLL_IMAX	1/4500
RATE_RLL_D	ATC_RAT_RLL_D	0.9 (for x-frame multicopters) or 1.27
RATE_RLL_FILT_HZ	ATC_RAT_RLL_FILT	none
		
RATE_PIT_P	ATC_RAT_PIT_P	0.9 (for x-frame multicopters) or 1.27
[...]

PR #3235, opened 2016-04-15
Version based parameter name remapping

DonLakeFlyer:
* Add FirmwarePlugin::paramNameRemapMajorVersionMap
* Add FirmwarePlugin::remapParamNameHigestMinorVersionNumber
* Use new routines in ParameterLoader to support remapping of parameters names based on firmware version
* New parameter remapping in APM Tuning screen to handle Copter 3.3->3.4 parameter name changes

The above adds support to the FactSystem/FirmwarePlugin such that a plugin can return parameter name remapping information in order to remap from one parameter name to another across versions. This allows Qml code to be written once and the correct mapping happens behind the scenes. Example:

```
    property Fact _rateRollP:   controller.getParameterFact(-1, "r.ATC_RAT_RLL_P")
    property Fact _rateRollI:   controller.getParameterFact(-1, "r.ATC_RAT_RLL_I")
    property Fact _ratePitchP:  controller.getParameterFact(-1, "r.ATC_RAT_PIT_P")
    property Fact _ratePitchI:  controller.getParameterFact(-1, "r.ATC_RAT_PIT_I")
```

Note the "r." prefix which identifies that this parameter should be remapped as needed depending on firmware version.

Fix for Issue #3152


DonLakeFlyer:
Wait to merge to here back from Randy about param semantics.

PR #3246, opened 2016-04-19
Add GPS drivers and RTCM output for RTK GPS

bkueng:
This adds the drivers and structure used for RTK GPS. The gps driver code is added as a submodule since it is shared with the PX4 firmware.

It provides a working version, but is still under development and not finalized yet. The whole thing is disabled so it does not interfere with the existing code. Dependencies should be Qt and STL only so it should compile for all platforms. Testing was done under Linux.

@DonLakeFlyer can you review this?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "libs/mavlink/include/mavlink/v1.0"]
 	path = libs/mavlink/include/mavlink/v1.0
 	url = https://github.com/mavlink/c_library.git
+[submodule "src/GPS/Drivers"]
+	path = src/GPS/Drivers
+	url = https://github.com/PX4/GpsDrivers.git
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -200,6 +200,7 @@ INCLUDEPATH += \
     src/input \
     src/Joystick \
     src/FollowMe \
+    src/GPS \
     src/lib/qmapcontrol \
     src/MissionEditor \
     src/MissionManager \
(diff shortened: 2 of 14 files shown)


LorenzMeier:
@bkueng Have you seen the build failure?

LorenzMeier:
@DonLakeFlyer I'd like to get this in ASAP, even if suboptimal so we can progress with RTK integration. This works fine in bench tests, pushing hard towards flight tests.

bkueng:
Yes it's `'QSerialPort' file not found` which I assume comes from the use of a too old Qt version in this specific build. QSerialPort was introduced in Qt 5.1. The current is 5.6, so the build is using a really old version. Question is whether we need to be compatible with such old versions. Also I see QSerialPort used elsewhere, so maybe you or @DonLakeFlyer can decide how to best handle this.

LorenzMeier:
I doubt the build system is using Qt 5.1. Can you re-check wether there is another potential cause?

LorenzMeier:
@bkueng I just checked and its the iOS build which is failing. Can you make your module conditional and not include it in the iOS build? We're not supporting serial for iOS anyway.

bkueng:
The relevant option why it fails is this:
https://github.com/mavlink/qgroundcontrol/blob/master/qgroundcontrol.pro#L93
`serialport` is only added for non-mobile builds. If we assume Qt >= 5.1 it should be safe to enable serialport for all platforms and it will build.

LorenzMeier:
Can you please build for Android locally and verify that?

LorenzMeier:
Actually the android build went through fine. I'm not sure you've fully grasped the issue yet. Please go back to the iOS build and work out why just that fails. The Travis output is pretty concise. And please forget the Qt version - there is no system out there with anything smaller than 5.5 installed.

bkueng:
It's clear why it fails. It's just not clear why it succeeds on android.

DonLakeFlyer:
This is fine as it is now to make forward progress since it doesn't really affect the rest of QGC. If there's anything that needs moving around I'll work with Beat to deal with that. This will also allow me to figure out how to work into link connect.

catch-twenty-two:
@LorenzMeier @DonLakeFlyer @bkueng this is awesome, combined with my position manger class it makes it super easy to connect an external gps and have it be used internally by qground control for positioning. (much better follow me accuracy and the ability for qground control to use a real gps on a laptop) I have tested my position manager addition pretty throughly and would like to get it merged in asap before it gets out of sync any further. Let me know what you think

PR is here:
https://github.com/mavlink/qgroundcontrol/pull/3267

PR #3247, opened 2016-04-19
Offline map

dogmaphobic:
Work on the *Offline Tile Cache* UI

* Made the map a lot larger during the definition of a set, making better use of the screen real estate.
* When viewing an existing tile set, the UI now shows the map and delineates the cached region.

screen shot 2016-04-19 at 2 11 17 pm

screen shot 2016-04-19 at 2 09 58 pm

### DIFF
--- a/android/gradle/wrapper/gradle-wrapper.properties
+++ b/android/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=http\://services.gradle.org/distributions/gradle-1.12-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.2.1-all.zip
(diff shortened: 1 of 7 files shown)


DonLakeFlyer:
Have you tried this on an iPhone? Was unusable before.

dogmaphobic:
> Have you tried this on an iPhone?

Nope, avoiding El Capitan for as long as I can :P

dogmaphobic:
> Have you tried this on an iPhone? Was unusable before.

Alright, finally upgraded to El Capitan, built and installed on both iPhone and iPad. As you said, iPhone is unusable. Not sure how to deal with it. Even the iPad has issues. Do we even support "little screens" (iPhone)?

DonLakeFlyer:
I have most stuff workable on my iPhone. I can look at it and make some suggestions with relation to what I did with other stuff.

dogmaphobic:
Wrapping up iOS.

* Adjusted default for size of iOS (it was still a bit large)
* Fixed issue with QGCTextField where it was using the system default font instead of our "scaled" font.
* Adjusted Offline Map Definition UI to fit on an iPad (iOS change only)
* Disable Offline Map cache for iPhone altogether until a new, specialized UI is written. As it is, it doesn't fit and it is unusable.

Tested on Android (Nexus 9) and it all work as is. No change was required.

Committing it now and making the other changes on a different PR.

PR #3248, opened 2016-04-19
update fixed wing cmd default values

dagar:


DonLakeFlyer:
@dagar Just checking that I understand. These are just in order to get a different default for loiter radius for Fixed Wing right? Kind of crappy to have to dupe all of it just for a default, but no other way right now. Need to consider partial override capability at some point.

dagar:
Yes, everything in src/FirmwarePlugin/PX4/MavCmdInfoFixedWing.json is just for better fixed wing defaults. I believe all of it could be shared with APM if there's a way.

DonLakeFlyer:
No it can't shared now. But you should dup this stuff to the APM side as well for now. I've added an issue for a better way to do this in 3.1

dagar:
PX4 fixed wing metadata duplicated for APM.

Issue #3237, opened 2016-04-16
qml file not found 


ivand58:
When tuning button is pressed it reports "qrc:/qml/PX4TuningComponentVTOL.qml: File not found".
The qgroundcontrol version is 2.9.4

DonLakeFlyer:
Tuning config isn't supported for VTOL. Button no longer shows up

LorenzMeier:
Tuning config is supported for VTOL. Where did the file go?

DonLakeFlyer:
Tuning config has never been supported for VTOL. The file never existed.

LorenzMeier:
Well, you're mistaken:
https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/PX4/PX4TuningComponentVTOL.qml

DonLakeFlyer:
Fixed now

PR #3250, opened 2016-04-20
Tuning not supported on VTOL

DonLakeFlyer:
Fix for #3237


LorenzMeier:
Why are you removing this? The VTOL needs its own tuning component.

DonLakeFlyer:
It doesn't exist. Tuning for VTOL is NYI.

LorenzMeier:
Can you please revert this? I've implemented it and the matching QML tuning file is here:
https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/PX4/PX4TuningComponentVTOL.qml

Issue #3251, opened 2016-04-20
Explicit frametype on summary screen


biapalmeiro:
On summary screen, for ArduPilot setup, the frametype is represented by a letter (X, V, H,etc.) what could be interpreted by users as an error.
![frametype](https://cloud.githubusercontent.com/assets/5597786/14639724/7f9029a6-0614-11e6-9aa5-feddf5dd4e06.png)

Instead showing just the main style which is represented by the letter, should be described the type also as on Airframe screen.
![frametypeproposal](https://cloud.githubusercontent.com/assets/5597786/14639768/ab1e3f54-0614-11e6-9369-83d695d48450.jpg)

Frametype styles based on Airframe screen information:
A Tail

H: X and H

Plus style: Quad
Plus style: Hexa
Plus style: Octo
Plus style: Heli

V (3DR Iris)

V Tail

X style: Quad
X style: Hexa
X style: Octo
X style: X8
X style: Tri
Y6A style: Y6A

Y6B (3DR TriCopter)

Version verified: v2.9.4-449-g7d77e22 (Development)

@tcanabrava @anselmolsm @cfelipesouza

biapalmeiro:
@dogmaphobic Do you think it is helpful to create a label for this kind of issue? UI or Design for example?

dogmaphobic:
Just added *UX* :)

PR #3254, opened 2016-04-20
Show a more educative airframe configuration

tcanabrava:
During tests, if the airframe was configured to be an 'X' type
we showed just 'X' on the airframe type. It's correct but a
single letter - specially an X - made users think that it was
a error code, not the airframe name. So, because of that, also
show the vehicle type plus the airframe type - a wrong value, but elucidative for users.

screenshot_20160420_151317


anselmolsm:
Related to #3251

Issue #3190, opened 2016-04-11
Figure out how to make toolbar transparent again with correct colors


DonLakeFlyer:

dogmaphobic:
Assigning to me as I had already done some of the work and have it stashed.

DonLakeFlyer:
Can you put the child controls parented to an Item which is sized correctly, so we don't have to pass around the available height thing? This will also make the right side dialogs work better.

dogmaphobic:
The was my first approach/attempt. The issue is that you need a QGCView way down below and FlightDisplayView is the QGCView. The map/video display has to be "fill parent". The Flight Widget within FlightDisplayView is properly sized and it should be the parent from which you get the height but ValuesWidget uses parent.parent.parent.qgcView to get the size. At that point my head was spinning and I gave up :)

DonLakeFlyer:
This whole parent/child chain needs a roto-tilling. But not till after 3.0. Let me see what I can do.

dogmaphobic:
Done.

PR #3257, opened 2016-04-22
Restoring toolbar transparency.

dogmaphobic:
This is in response to issue #3190 

I've restored the main Toolbar transparency. For that to work, the main map/video view has to fill the entire screen. The gadgets, widgets and dialogs have to be aware of the available height (as opposed to using the full height). That is accomplished by querying *ScreenTools*.

screen shot 2016-04-22 at 5 08 00 am

While at it, I've fixed the layout for the *ValueWidget*, making the tabulation work correctly.

screen shot 2016-04-22 at 5 08 05 am

I still think that has a bit of a *Mission Planner* feel to it. All that's missing is an option to display cubits per lunar cycle.

### DIFF
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -45,9 +45,7 @@ QGCView {
 
     QGCPalette { id: qgcPal; colorGroupEnabled: enabled }
 
-    property var _activeVehicle:    QGroundControl.multiVehicleManager.activeVehicle
-
-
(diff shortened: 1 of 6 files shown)


LorenzMeier:
@dogmaphobic Nice! And I agree that throwing our hands up and giving up at the user model like here is something that we should almost never do. I think having this creep in is acceptable, but it should be crystal clear that this is something that developers think they wanted and not something that is inherently part of the application.

Doing things like this also messes big time with the user model: Because we take friction away from flight operators and developers to actually go into a discussion sorting out what is important and what is not, we probably have by default not the things enabled that operators actually need.

Issue #3262, opened 2016-04-22
QGC Crash while entering Flight Mode Tab if RC is on!


sn0west:
Daily QGC crash if I try to enter the Flight Modes tab while my RC is on. If the RC is off then I can enter this tab. 
If I flash the apm stack QGC dosen't crash.

I have a Mac osx 10.10.5 and flashed the latest px4 master.

App Logg over Terminal
[160422_QGC_Crash_FM.txt](https://github.com/mavlink/qgroundcontrol/files/232520/160422_QGC_Crash_FM.txt)

Syslog
[160422_QGC_Crash_SystemLog.txt](https://github.com/mavlink/qgroundcontrol/files/232521/160422_QGC_Crash_SystemLog.txt)

Do not work
![bildschirmfoto 2016-04-22 um 22 17 44](https://cloud.githubusercontent.com/assets/9918649/14754383/562b198a-08db-11e6-8744-be14b3f5e4ed.png)

jcooper21:
I have the same issue and was just about to post it. I have PX4 beta build from today on the pixhawk.

jcooper21:
The issue is if you click on the flight mode tab before all the sensors are calibrated, then it crashes.  It works if you calibrate everything first.  Hopefully they can fix the bugs.

dogmaphobic:
I could not reproduce it. It took a bit because I had no system using a "normal" radio setup. To make sure, I put together a frame using an X8R Receiver and a Taranis Radio. Flashed the Pixhawk with PX4 Master, went through all the frame setup and calibration and proceeded to go into the flight modes panel and set a couple of channels just fine.

![screen shot 2016-04-22 at 8 59 21 pm](https://cloud.githubusercontent.com/assets/749243/14758338/937fb740-08cd-11e6-9b11-4b401e332065.png)

What exactly is your setup like?

@jcooper21  *Windows 10*, *PX4 beta build* and *you cannot setup an airframe* are not exactly what this issue is about. If you have a better description of your problem, could you please enter a new issue for it?

dogmaphobic:
Ha! missed you by less than a second it seems. Yes, I did go through the calibration first.

dogmaphobic:
Alright, let me reset the Pixhawk to nothing and start over.

dogmaphobic:
The crash is in this line:

[...]

PR #3264, opened 2016-04-23
Fixed crash when switching to the Flight Modes panel.

dogmaphobic:
This was reported by @sn0west and @jcooper21 in #3262. If you were to select the *Flight Modes* panel before running the calibrations (*Sensors*), QGC would crash while plowing through non existing *Facts*.

This "*fix*" prevents you from attempting to run the Flight Modes before running the calibrations. I've also added some NULL pointer checks as the original code was making indirect calls without any checks.

To be consistent with the above, I've moved *Flight Modes* below *Sensors* in the setup list.

@DonLakeFlyer If indeed you need to run the calibration before assigning Flight Modes (weird), I guess this is all that was needed. If not, this is the code that was causing the crash:

https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/PX4/PX4SimpleFlightModesController.cc#L48

The original line was:

`int flightModeReversed = getParameterFact(1, QString("RC%1_REV").arg(flightModeChannel + 1))->rawValue().toInt();`

Where `getParameter()` was returning NULL and proceeding to call `rawValue()` on a NULL pointer. The parameter in question above was `RC0_REV`.

### DIFF
--- a/src/AutoPilotPlugins/PX4/FlightModesComponent.cc
+++ b/src/AutoPilotPlugins/PX4/FlightModesComponent.cc
@@ -1,24 +1,24 @@
 /*=====================================================================
- 
+
  QGroundControl Open Source Ground Control Station
- 
+
  (c) 2009 - 2014 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
- 
+
  This file is part of the QGROUNDCONTROL project
- 
+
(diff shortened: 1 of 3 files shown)


sn0west:
Just to make clear, I did the RC calibration before entering the FM Tab.

PR #3267, opened 2016-04-24
Enhancement.position manager

catch-twenty-two:
Adds an abstraction layer to the positional information read by the qgc system. Any type inheriting from QGeoPositionInfoSource can easily inject it's GPS data into the system via this new class(some examples would be, simulate qgc movement, add log play back, use an external gps, ect...) . This also removes all simulated position methods from the Follow me class and puts them in their own class.


LorenzMeier:
@catch-twenty-two Travis failed

catch-twenty-two:
@LorenzMeier Should be good to go now

PR #3269, opened 2016-04-24
Support REPOSITION MAVLink command in PX4

LorenzMeier:
@DonLakeFlyer @dogmaphobic This goes with: https://github.com/PX4/Firmware/pull/4347

One thing I couldn't figure out yet is how to click a point on the map. How does this work?

Since this is not a major UI thing I'd like to get this into 3.0 so that we can offer basic tablet operation with the release. I found already a few feedback items that we should address in general.

### DIFF
--- a/src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.cc
@@ -134,7 +134,7 @@ void ArduCopterFirmwarePlugin::guidedModeTakeoff(Vehicle* vehicle, double altitu
     }
 
     mavlink_message_t msg;
-    mavlink_command_long_t cmd = {};
+    mavlink_command_long_t cmd;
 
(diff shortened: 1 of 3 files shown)


LorenzMeier:
screen shot 2016-04-24 at 17 29 40

dogmaphobic:
@LorenzMeier 

> One thing I couldn't figure out yet is how to click a point on the map. How does this work?

Don has done all that work. This is how the Mission Editor deals with points on a map:

https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionEditor/MissionEditor.qml#L285-L306

For the *Flight View*, there is a *MouseArea* handler at the bottom already doing something else. You could add the code in there:

https://github.com/mavlink/qgroundcontrol/blob/master/src/FlightDisplay/FlightDisplayViewMap.qml#L127-L140

DonLakeFlyer:
Reposition is already fully supported by ardupilot. You just need to implement the FirmwarePlugin methods on the px4 side. Should require changes to map code.  I'm off the grid at track till tonight/tomorrow. I'll have more time then.

PR #3284, opened 2016-04-26
Fix fact text field 2

tcanabrava:
This commit fixes the QGCViewDialog with non Bitmask values.

### DIFF
--- a/src/FlightMap/Widgets/ValuesWidget.qml
+++ b/src/FlightMap/Widgets/ValuesWidget.qml
@@ -53,7 +53,7 @@ QGCFlickable {
     }
 
     function showPicker() {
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
@dogmaphobic I'm not quite sure what's going on with the new available height thing. Can you review and merge if ok.

dogmaphobic:
> @dogmaphobic I'm not quite sure what's going on with the new available height thing. Can you review and merge if ok.

That was an earlier problem I missed when dealing with toolbar transparency. I had not noticed the dialog presented when entering an invalid format for a Fact (`FactTextField` calls showDialog). That's all said and done. While working on that, I noticed the new bitmask edit was not working (taking over and not dismissing the dialog, giving the impression that everything was frozen). This is what this PR is all about. Tomaz found it and fixed it.

Merging...

PR #3285, opened 2016-04-26
Fix Checkbox Visibility

tcanabrava:
For some reason, the Checkbox fails to appear in a few
systems. with this patch it appears.

With the patch:

screenshot_20160426_151837

Without the patch on Archlinux:

screenshot_20160426_152002

### DIFF
--- a/src/QmlControls/QGCCheckBox.qml
+++ b/src/QmlControls/QGCCheckBox.qml
@@ -9,6 +9,21 @@ CheckBox {
     property var __qgcPal: QGCPalette { colorGroupEnabled: enabled }
 
     style: CheckBoxStyle {
+        indicator: Rectangle {
+            implicitWidth: 16
+            implicitHeight: 16
+            radius: 3
+            border.color: control.activeFocus ? "darkblue" : "gray"
+            border.width: 1
+            Rectangle {
+                visible: control.checked
+                color: "#555"
+                border.color: "#333"
+                radius: 1
+                anchors.margins: 4
+                anchors.fill: parent
+            }
+        }
         label: Item {
             implicitWidth: text.implicitWidth + 2
             implicitHeight: text.implicitHeight


src/QmlControls/QGCCheckBox.qml
```diff
@@ -9,6 +9,21 @@ CheckBox {
...
+            border.width: 1
+            Rectangle {
+                visible: control.checked
+                color: "#555"
+                border.color: "#333"
+                radius: 1
+                anchors.margins: 4
+                anchors.fill: parent
+            }
+        }
```
DonLakeFlyer: Is this copied from the internal Qml style code? Wondering what happens across all OS. Since I think this style code may differ OS to OS. Not quite sure, but maybe.

DonLakeFlyer: I looked at the OSX implementation of Base for indicator and it has way more stuff than just this. For example the checkmark image plus lots of other complexity. How is this code setting the checkmark bitmap which for OS X is part of the indicator style.

DonLakeFlyer:
I tried running this on OS X and it changes the style of all checkboxes which I'm not sure is a good thing. Not sure if other things will be out of whack or not. I'd much rather figure out what is really going wrong here then not use the Qt provided OS specific indicator style implementation. It may just be a color problem or something like that.

tcanabrava:
On Tue, Apr 26, 2016 at 11:36 PM, Don Gagne <notifications@github.com>
wrote:

> I tried running this on OS X and it changes the style of all checkboxes
> which I'm not sure is a good thing. Not sure if other things will be out of
> whack or not. I'd much rather figure out what is really going wrong here
> then not use the Qt provided OS specific indicator style implementation. It
> may just be a color problem or something like that.
>

It's not a good thing to change all checkboxes and I'm aware of that, it
was a fix just to make you guys aware of what's happening here on some
linux boxes. It's not a color problem - the rectangle of the Checkbox still
exists but it's *really* thin, it you look at my broken screenshoot you can
see a line where a checkbox should have been.

DonLakeFlyer:
@tcanabrava Do you know if this is only a problem with checkboxes in QGC? Or does it also happen in other Qt Qml apps on archlinux?

LorenzMeier:
@tcanabrava Please note that all new contributions to QGC have to be made under a dual license including Apache 2.0 + GPLv3. Just so you're aware that this PR now has to fulfil this requirement.

tcanabrava:
Lorenz,

Thanks for letting me know.
I don't have a computer right now, that's why I disappeared. I'll buy a new
computer when I go to Switzerland in two weeks.
Em 31 de mai de 2016 17:38, "Lorenz Meier" <notifications@github.com>
escreveu:

@tcanabrava <https://github.com/tcanabrava> Please note that all new
contributions to QGC have to be made under a dual license including Apache
2.0 + GPLv3. Just so you're aware that this PR now has to fulfil this
requirement.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/mavlink/qgroundcontrol/pull/3285#issuecomment-222812594>,
or mute the thread
<https://github.com/notifications/unsubscribe/AD1zUIO9oP8C4XDBHkWSRhEMcx2xw8uQks5qHJwsgaJpZM4IQVnt>
.

DonLakeFlyer:
Closing this since it needs a different fix

PR #3287, opened 2016-04-27
AutoConnect for RTK GPS

DonLakeFlyer:
First cut at auto connect for RTK GPS:
* It will auto connect to UBlox RTK GPS
* RTK GPS auto-connect can be turned on/off from Settings

Not yet working:
* Auto-disconnect NYI
* Support for connecting to RTK GPS using a manually created comm link is NYI
* Mobile NYI

@LorenzMeier @bkueng FYI

### DIFF
--- a/src/GPS/GPSManager.cc
+++ b/src/GPS/GPSManager.cc
@@ -36,7 +36,7 @@ GPSManager::~GPSManager()
     cleanup();
 }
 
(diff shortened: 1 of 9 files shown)


LorenzMeier:
@DonLakeFlyer Nice! What we need for mobile is to change from QSerialLink to our own SerialLink abstraction (as QSerialLink is broken on Android and our abstraction works).

PR #3291, opened 2016-04-27
Show firmware version if the vehicle is connected

tcanabrava:
screenshot_20160427_163523
Display the firmware version if the vehicle is already connected
and the user clicks on Firmware Upgrade.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMAirframeComponentSummary.qml
+++ b/src/AutoPilotPlugins/APM/APMAirframeComponentSummary.qml
@@ -35,5 +35,10 @@ FactPanel {
                      : /* Fact.value == 10 */  "New Y6");
 
         }
+
+        VehicleSummaryRow {
+            labelText: qsTr("Firmware Version:")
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
The firmware version info on Vehicle should be exposed as properties (if not already there, can't remember) not invokable.

dogmaphobic:
> The firmware version info on Vehicle should be exposed as properties (if not already there, can't remember) not invokable.

It seems it's a response to `MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES`

http://mavlink.org/messages/common#AUTOPILOT_VERSION

tcanabrava:
The firmware version is a text message from mavlink, not transformed into a
fact. There was already a method to parse and store the text on the vehicle
class, I just added a invokable to be able to show it on qml.
Em 28 de abr de 2016 21:34, "Gus Grubba" <notifications@github.com>
escreveu:

tcanabrava:
adding it as property...

tcanabrava:
closing because of a new one opened;

PR #3294, opened 2016-04-28
Camera component

dogmaphobic:
Adding PX4 Camera Component (Settings)

screen shot 2016-04-28 at 8 54 31 am


DonLakeFlyer:
Need to make sure anything new works on iPhone size screen as well.

dogmaphobic:
> Need to make sure anything new works on iPhone size screen as well.

There is quite a bit of redesign needed for mobile in general.

PR #3296, opened 2016-04-29
Show firmware version v2

tcanabrava:
rebased on master, and updated to properties.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMAirframeComponentSummary.qml
+++ b/src/AutoPilotPlugins/APM/APMAirframeComponentSummary.qml
@@ -38,7 +38,7 @@ FactPanel {
 
         VehicleSummaryRow {
             labelText: qsTr("Firmware Version:")
-            valueText: activeVehicle.firmwareMajorVersion() + "." + activeVehicle.firmwareMinorVersion() + "." + activeVehicle.firmwarePatchVersion()
+            valueText: activeVehicle.firmwareMajorVersion + "." + activeVehicle.firmwareMinorVersion + "." + activeVehicle.firmwarePatchVersion
         }
     }
 }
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
The issue with this change is that as it stands now only ArduPilot vehicles know their firmware version numbers. PX4 Vehicle support for firmware version numbering is not yet supported. So for now I'd make ArduPilot only. Put in an issue to support PX4 versioning and update to include the version number user output to both sides.

DonLakeFlyer:
@LorenzMeier Do you think you could change PX4 firmware to send out PX4 version message at first connect? Then I could catch that in the PX4 plugin and set version numbers. This would be easier than having to have QGC ask for it from a timing standpoint,

dagar:
I could add this to PX4 pretty quickly if we could agree on the best way. Would it be better to add a string to AUTOPILOT_VERSION instead of just send it as STATUSTEXT? Then QGC could request MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES on first connect to get the version.

PX4 versioning could be similar to QGC
v1.3.0 (stable)
v1.3.0-12-gGITSHA (development)

http://mavlink.org/messages/common#AUTOPILOT_VERSION

DonLakeFlyer:
The only thing Vehicle has is major.minor.patch as three numbers. No string support at this point.

LorenzMeier:
@DonLakeFlyer A text string is obviously not sustainable. PX4 will respond to a MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES with an AUTOPILOT_VERSION message containing everything you asked for.

DonLakeFlyer:
> A text string is obviously not sustainable.

Umm, no kidding. I never asked for a text string. What I'm asking is for the firmware to just send out AUTOPILOT_VERSION message over a new mavlink connection without QGC having to ask for it with a MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES. That would be simpler. If not, I'll do the request thing but that will take a bit more work.

dagar:
I was suggesting a string in addition to an actual number. Then we could show release candidates, etc.
eg `v1.3.0rc1`

DonLakeFlyer:
Closing this since I'm taking it over and will fix up the PX4 version query side of things.

PR #3297, opened 2016-04-29
Add search v4

tcanabrava:
screenshot_20160429_151909

Add a parameter search field, only on Desktop mode.

### DIFF
--- a/src/FactSystem/ParameterLoader.cc
+++ b/src/FactSystem/ParameterLoader.cc
@@ -467,13 +467,7 @@ Fact* ParameterLoader::getFact(int componentId, const QString& name)
 
 QStringList ParameterLoader::parameterNames(int componentId)
 {
-    QStringList names;
-
-    foreach(const QString &paramName, _mapParameterName2Variant[_actualComponentId(componentId)].keys()) {
-        names << paramName;
-    }
-
-    return names;
+    return _mapParameterName2Variant[_actualComponentId(componentId)].keys();
 }
 
 void ParameterLoader::_setupGroupMap(void)
--- a/src/QmlControls/ParameterEditor.qml
+++ b/src/QmlControls/ParameterEditor.qml
@@ -70,6 +70,7 @@ QGCView {
                 id:     header
                 width:  parent.width
                 height: ScreenTools.defaultFontPixelHeight * 1.75
+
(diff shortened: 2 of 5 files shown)


LorenzMeier:
@tcanabrava Needs work on the Travis build - but the feature looks great. Did you remove it from the tools menu?

tcanabrava:
Only on desktop mode. In mobile it should still be on the menu.
Em 29 de abr de 2016 18:59, "Lorenz Meier" <notifications@github.com>
escreveu:

LorenzMeier:
I think the filter field should be present in all builds and the entry in the tools menu should go. This search field is going to get a lot of use and we want to have it consistently.

tcanabrava:
I agree, but currently the filter is just too slow on mobile builds, so I'm
researching a way to speed it up, until then, unfortunately it will need to
have both options, one for desktop and one for mobile.
Em 30 de abr de 2016 03:49, "Lorenz Meier" <notifications@github.com>
escreveu:

DonLakeFlyer:
Even desktop can still end up being a bit sluggish with high param counts. Parameter editor needs a rewrite to use QmlObjectListMode in order to speed things up. Then it will be possible to do speedy typedown-style search. I'd rather wait until that is done. My concern is that this will just generate additional confusion when people switch between desktop and mobile. I think for 3.0 the solution is to move search out of the tools menu, to solve usability problem, but still use the old mechanism such that it works on all platforms. It will require checks on an iPhone to determine how to fit.

tcanabrava:
Okay, I'll rewrite to use the model.
Em 2 de mai de 2016 20:49, "Don Gagne" <notifications@github.com> escreveu:

DonLakeFlyer:
No need. I'll do it tonight and put in a pull.

DonLakeFlyer:
I under-estimated the amount of work to convert to QmlObjectListModel. I worked my way through what it is going to take and all of this will have to wait till 3.1: https://github.com/mavlink/qgroundcontrol/issues/3309.

In the mean time I'm moving Search out of tools using the same search model that exists now to fix the discoverability issues.

DonLakeFlyer:
I've moved Search to the top level ui. We'll work on performant typedown support for the 3.1 release.

PR #3299, opened 2016-04-30
Support for new esp firmware

dogmaphobic:
The new MavEsp8266 Firmware (1.0.7) allows you to set it to either *AP Mode* (the default, and what it's been up to now) or *Station Mode*. When set to Station Mode, it will connect to the AP defined by the SSID (using the defined password). This allows for multiple vehicles to connect to QGC provided they all belong to the same AP network.

The firmware once booted, will attempt to connect to the defined Access Point. If after a minute it cannot connect, it will revert to AP mode.

screen shot 2016-04-30 at 3 38 06 am


dogmaphobic:
I've added a separate set of SSID and Password for Station Mode to the ESP Firmware (now version 1.0.8). That allows you to have the Station Mode with the credentials to log into your (external) AP while allowing the ESP to use its own credentials when and if it falls back to AP mode.

While at it, I've updated the UI as well.

screen shot 2016-04-30 at 5 05 22 pm

Issue #3301, opened 2016-04-30
Optional params for PX4 config: TRIG_MODE and UAVCAN_ENABLE


LorenzMeier:
@dogmaphobic Could you help out to redesign the UAVCAN config in the power config section and the camera trigger config in the summary and in the camera section?

Not every PX4 drone has this functionality, and it should not lead to error messages in QGC.

LorenzMeier:
Addressed in #3302.

PR #3302, opened 2016-04-30
Px4 usability

LorenzMeier:
This addresses #3301.

### DIFF
--- a/src/AutoPilotPlugins/PX4/CameraComponentSummary.qml
+++ b/src/AutoPilotPlugins/PX4/CameraComponentSummary.qml
@@ -14,7 +14,7 @@ FactPanel {
     QGCPalette { id: qgcPal; colorGroupEnabled: enabled }
     FactPanelController { id: controller; factPanel: panel }
 
-    property Fact _camTriggerMode:  controller.getParameterFact(-1, "TRIG_MODE")
+    property Fact _camTriggerMode:  controller.getParameterFact(-1, "TRIG_MODE", false)
     property Fact _camTriggerPol:   controller.getParameterFact(-1, "TRIG_POLARITY", false) // Don't bitch about missing as these only exist if trigger mode is enabled
     property Fact _auxPins:         controller.getParameterFact(-1, "TRIG_PINS",     false) // Ditto
     property Fact _timeInterval:    controller.getParameterFact(-1, "TRIG_INTERVAL", false) // Ditto
--- a/src/AutoPilotPlugins/Common/RadioComponent.qml
+++ b/src/AutoPilotPlugins/Common/RadioComponent.qml
@@ -469,7 +469,7 @@ QGCView {
                 }
 
                 Repeater {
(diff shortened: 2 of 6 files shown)


src/AutoPilotPlugins/PX4/CameraComponentSummary.qml
```diff
@@ -14,7 +14,7 @@ FactPanel {
     QGCPalette { id: qgcPal; colorGroupEnabled: enabled }
     FactPanelController { id: controller; factPanel: panel }
 
-    property Fact _camTriggerMode:  controller.getParameterFact(-1, "TRIG_MODE")
+    property Fact _camTriggerMode:  controller.getParameterFact(-1, "TRIG_MODE", false)
```
dogmaphobic: @LorenzMeier 
If there is a case where TRIG_MODE is not available at all, the entire Camera Setting panel should be hidden. If that's the case this code should be changed:
https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/PX4/PX4AutoPilotPlugin.cc#L129-L131
And use the same logic as the one used right below it for the WiFi Module. Let me know if this is the case and I can change it.
I added the last boolean to `controller.getParameterFact()` to deal with the rest of the parameters that only exist if TRIG_MODE is disabled. If the entire functionality is missing, we might as well don't show the panel at all.

LorenzMeier: Yes, that would indeed be the better change. If the param is not present no trigger controls should be in place.

dogmaphobic: Working on it. I'm going through a do-over on general UI problems. Mostly the very egregious one first, especially when it comes to the iPhone. I will add this in.

DonLakeFlyer:
I'm slightly concerned with the new false = don't trigger missing parameter thing. The expected pattern was for parameters which could possibly be missing you would explicitly call parameterExists check and then if exist you can get the param. That way it is very obvious what you are doing. My concern is that the "don't trigger" flag is a bit magic/unclear from code readability and may end up being misused. Then if it is used incorrectly, the parameter is missing strange things will happen and the error path is kind of unknown. Whereas if you use getParameterFact on a missing param the UI goes bonkers in a user visible way which should generate an undertstandable bug report. The missing param trigger has saved my ass many times.

dogmaphobic:
That logic fits better with the Javascript property paradigm and those are only used in very few exceptional places. Otherwise you end up having to create a function to make a call, test the result and then make another call to get the Fact/value. No signals, no automatic updates.

DonLakeFlyer:
I use this pattern for missing param handling/code logic forking in Qml:
property Fact _nullFact
property bool _fooExists: parameterExists("foo") 
property Fact _fooFact: _fooExists ? getParameterFact("foo") : _nullFact

QGCTextField { visible: _fooExists; fact: _fooFact }

This way you can still bind to it even when it doesn't exist and you then hide or fork code path based on existence without having to jump through too many hoops in the control code.

That said we just have to be careful code reviewing usage of this new argument. But without it you need to be careful code reviewing craziness like above. Six of one, half a dozen of the other. I think maintaining a similar pattern usage makes sense though it possible/reasonable.

PR #3286, opened 2016-04-26
Android joystick

gregd72002:
Joystick support for Android devices

https://github.com/mavlink/qgroundcontrol/issues/3049

- only one joystick is supported on Android

PR #3305, opened 2016-05-01
Android joystick support

gregd72002:


gregd72002:
Rebased based on https://github.com/mavlink/qgroundcontrol/pull/3286

DonLakeFlyer:
Sorry for the delay on this. I'm out of town for a while with only sporadic spots of time to look at things. I'll try to get to this for a code review.

@dogmaphobic I'm not set up for Android on the road. Do you think you could run this?

DonLakeFlyer:
Just getting around to look at this now. Looks like a rebase has gone haywire with it?

gregd72002:
Not sure what's going on in here. In any case here is a merged version (3 days old) that can be pulled https://github.com/gregd72002/qgroundcontrol/tree/37e662287c584eac23bb752a6dc8ce3cdea51963

DonLakeFlyer:
Can you submit a new pull which is up to date with master?

gregd72002:
enclosed. Not sure why it tries to merge 'libs/mavlink/include/mavlink/v1.0', it should be discarded

DonLakeFlyer:
> enclosed. Not sure why it tries to merge 'libs/mavlink/include/mavlink/v1.0', it should be discarded

Make sure you're up to date on submodules? git submodule init, git submodule update

DonLakeFlyer:
Anyway, thanks. I'll should have some time to look at this over the weekend.

qgroundcontrol.pro
```diff
@@ -266,6 +266,7 @@ HEADERS += \
     src/HomePositionManager.h \
     src/Joystick/Joystick.h \
     src/Joystick/JoystickManager.h \
+    src/VehicleSetup/JoystickConfigController.h \
```
DonLakeFlyer: This add joystick support to all builds which will include iOS which is incorrect. Joystick support should now be everything but ios build.

src/Joystick/Joystick.cc
```diff
@@ -28,14 +28,6 @@
 
 #include <QSettings>
 
-#ifndef __mobile__
```
DonLakeFlyer: I think this file has the same no iOS problem

src/Joystick/JoystickAndroid.h
```diff
@@ -0,0 +1,49 @@
...
+    JoystickAndroid(const QString& name, int axisCount, int buttonCount, int id, MultiVehicleManager* multiVehicleManager);
+    ~JoystickAndroid();
+
+    static QMap<QString, Joystick*> discover(MultiVehicleManager* _multiVehicleManager); 
+
+private:
+    bool handleKeyEvent(jobject event);
+    bool handleGenericMotionEvent(jobject event);
+
+    virtual bool open();
```
DonLakeFlyer: We're staring to use final on virtual function endpoints.

src/Joystick/JoystickAndroid.h
```diff
@@ -0,0 +1,49 @@
...
+class JoystickAndroid : public Joystick, public QtAndroidPrivate::GenericMotionEventListener, public QtAndroidPrivate::KeyEventListener
+{
+public:
+    JoystickAndroid(const QString& name, int axisCount, int buttonCount, int id, MultiVehicleManager* multiVehicleManager);
+    ~JoystickAndroid();
+
+    static QMap<QString, Joystick*> discover(MultiVehicleManager* _multiVehicleManager); 
+
+private:
+    bool handleKeyEvent(jobject event);
```
DonLakeFlyer: Coding style is _fnname and _membername for all private/protected function and member names.

DonLakeFlyer:
I added a few comments inline, but here are some across the board things:
- The removal of all the ifdefs and moving code to mainline sources means that all this code is in the iOS build as well which is incorrect.
- Coding style for protected/private member variables and functions are to prefix with _

Next I'll give it some run testing.

gregd72002:
thanks for looking into this. Re iOS build, correct me if I'm wrong but this should continue to work as before. There is ifdefs in https://github.com/gregd72002/qgroundcontrol/blob/37e662287c584eac23bb752a6dc8ce3cdea51963/src/Joystick/JoystickManager.cc#L67

what basically should activate joystick for all builds but iOS

DonLakeFlyer:
Ok, I'll ok again. It was hard to tell from just the diffs. I saw removal of #ifdef __mobile__ would seem to indicate issues with iOs. But ifyou restructured to not need those then that is a different story.

libs/mavlink/include/mavlink/v1.0
```diff
@@ -1 +1 @@
-Subproject commit 9605f8087d3734ed89fa0aba6f343ced041fb3cb
+Subproject commit 39c14e81edf2eac7b5bfe6771eb56614f46ef098
```
DonLakeFlyer: You need to update your mavlink version:
git submodule init
git submodule update

DonLakeFlyer:
Look slike iOS built fine in Travis so that is ok.

DonLakeFlyer:
I can't build Android on my OSX box. It can't find: #include <QtCore/private/qjni_p.h>. What OS are you building android on and in you Qt install what is the path to that file?

gregd72002:
i'm using QtCreator (qt 5.6) on my OSX (Yosemite). It builds armeabi-v7a for android (SDK: android-22).

What #include is it that it complains about?

DonLakeFlyer:
Ah, 5.6. We don't support 5.6 yet. Needs to be 5.5.

gregd72002:
Right, does it mean you need a new pull request? As long as it's straightforward I will do it next couple of days or so...

DonLakeFlyer:
Yes I'll need a new/updated pull. QGC doesn't work correcttly with 5.6 so it needs to be 5.5 for now.

gregd72002:
this will not work on 5.5 (at lease I do not see a way of implementing it). Qt 5.5 does not seem to implement GenericMotionEventListener nor KeyEventListener for Android. These are the key events that are used to implement Android joystick support.

What are the options, when do you expect to migrate to 5.6?

LorenzMeier:
@gregd72002  Please note that all new contributions to QGC have to be made under a dual license including Apache 2.0 + GPLv3.

NaterGator:
> What are the options, when do you expect to migrate to 5.6?

The Qt 5.7 release is [imminent](https://wiki.qt.io/Qt-5.7-release) and @DonLakeFlyer [expressed](https://github.com/mavlink/qgroundcontrol/issues/3049#issuecomment-202486059) [interest](https://github.com/mavlink/qgroundcontrol/pull/3043#issuecomment-198184319) in skipping the 5.6 release in favor of going directly to 5.7. Since Qt will release an RC imminently and the final release should be a month or less behind that I would agree. 5.7 is going to have a bunch of interesting improvements. Sadly it doesn't look like we'll get release debug symbols in 5.7 :disappointed:

DonLakeFlyer:
> What are the options, when do you expect to migrate to 5.6?

We'll look at moving to a new Qt after 3.0 gets out the door. 3.0 release is in a horizon of weeks not months.

gregd72002:
sounds like this might take some time, I'm not too keen to maintain the Android Joystick support code on regular basis as it is pretty big. Can I suggest to split the pull request into 2 parts?
- restructure (to allow different joystick implementations) - to be merged asap
- Android Joystick support (requires new Qt version) - to be merged at later stage

dagar:
If you push the 2nd piece to another branch we can try to keep it in sync with master until we're ready for a QT upgrade.

DonLakeFlyer:
Yes, the restructure as a separate pull makes a ton of sense. After that gets through put the second part in a separate pull and we'll figure out how to keep it mostly up to date.

gregd72002:
pull request for JoystickRestructure created: https://github.com/mavlink/qgroundcontrol/pull/3459

DonLakeFlyer:
Closing since this is being split into two pulls.

PR #3307, opened 2016-05-02
Ui tweaks

dogmaphobic:
While working on fixing several UI problems for the iPhone, I found an error in ScreenTools that affected all UI in all platforms other than the one the original UI item was developed.

QGC uses two system wide variables that define the default character width and height. This is used throughout the program to compute dimensions in a way that scales as close as possible to all platforms. The system default font is measured (width and height in pixels) and a constant *factor* is used to fudge the value to visually approximate the desired result in the various platforms QGC supports.

The error was that while `ScreenTools.defaultFontPixelHeight` was properly computed, `ScreenTools.defaultFontPixelWidth` was not. That meant that every UI component using `ScreenTools.defaultFontPixelWidth` would only look right within the platform it was developed.

This PR addresses the most egregious errors I could find but many still exists.

Changes:

* Fixed `ScreenTools.defaultFontPixelWidth` as described above
* Fixed both Camera Component and WiFi Component panels for mobile. Especially for smaller screens such as the iPhone (they both relied heavily on `ScreenTools.defaultFontPixelWidth` for computing elements width).
* Made the Setup View buttons (the vertical list of buttons on the left side) slightly bigger on smaller screens.
* Added a contrasting border color to the Critical Message dialog that pops up at the bottom of the screen. That dialog has the same color as the background and it just looked wrong. I went ahead and also added the border to the system message dialog (the one shown when you hit the Megaphone icon on the toolbar).
* Fixed a clipping error in the Setup Menu where were you to scroll it up, it would scroll over the menu bar. This was only visible on small screens.
* Fixed some issues in the Vehicle Summary panel where some text (Label/Value pairs) were being overlapped.
* Added code that if Camera Support is not present, the Camera Component panel is not shown at all.


DonLakeFlyer:
> ScreenTools.defaultFontPixelHeight was properly computed, ScreenTools.defaultFontPixelWidth was not.

Ahhh! I've hit weirdness with this one across OS and just put it up to crazy font stuff. Could never figure out what was going on. Nice.

dogmaphobic:
> Could never figure out what was going on. Nice.

Yep, it was one of those things where I kept looking and not seeing it. It took some debugging on the iPhone to finally track it down.

PR #3314, opened 2016-05-03
Adding a few new parameters to MockLink.

dogmaphobic:
That is one PITA file to edit. I didn't want to change that finicky, persnickety, fussy "tab" delimiter to a "blank" delimiter as I was not sure how to deal with all those APM files in there.


NaterGator:
> I didn't want to change that finicky, persnickety, fussy "tab" delimiter to a "blank" delimiter as I was not sure how to deal with all those APM files in there.
_Sigh_... tab is an unambiguous unit of indent and spaces are used for alignment. And nobody will agree on that point for decades.

dogmaphobic:
No, that's not what I meant :) The parser is looking for an exact count of "\t" charters, instead of delimiting the lines by "blanks" (any number of anything that is not a character). My editor got confused by the tabs/spaces and made a mess out of it. QGC simply crashes to the ground if any line in there has an extra tab (or is missing one. )

dogmaphobic:
>Sigh... tab is an unambiguous unit of indent and spaces are used for alignment.

Exactly. My editor tried to be smart and replaced all non indenting tabs by spaces, causing the whole thing to break.

DonLakeFlyer:
Can you put in an issue about QGC crapping out if the file format is bad. Param save/load code pattern is ancient. Should be bullet proofed.

PR #3333, opened 2016-05-05
UI Layout Reorg

dogmaphobic:
I've changed the way the UI is laid out so it is now consistent across all platforms and devices.

The primary change is moving away from pixel size fonts to point size fonts. And instead of relying on the system default point size, which vary wildly between platforms, QGC now sets its own point size on a platform basis.

The end result, the UI is now consistent and usable from a small phone screen to a large Desktop screen.

I also tweaked several UI items, such as the *Summary Panel*, to conform to the design guidelines, even though these are not quite finalized yet.

I have **NOT** touched any of these and they have yet to go through a make up session:

* General Settings
* Comm Link Settings
* Firmware update
* Joystick settings
* Radio Settings
* Airframe configuration
* Sensors settings
* Tunning
* Parameters
* Anything to do with APM
* Anything to do with Missions

Nexus 4 (Tiny Android Screen):
nexus4

Nexus 7 (Wide and short Android tablet):
nexus7

Nexus 9 (Regular sized Android Tablet):
nexus9

iPhone 6:
iphone6

iPad:
ipad

Mac OS:
mac os

Ubuntu:
ubuntu

I got lazy and didn't want to boot Windows again to get a screen grab off of it. Font rendering in Windows sucks so it all looks like something from the 1980s any way. Maybe it has something to do with the fact I only have one VM running Windows 7.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMAirframeComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMAirframeComponent.qml
@@ -119,7 +119,7 @@ QGCView {
                 anchors.left:           parent.left
                 anchors.right:          applyButton.right
                 text:                   qsTr("Please select your airframe type")
-                font.pixelSize:         ScreenTools.mediumFontPixelSize
+                font.pointSize:         ScreenTools.mediumFontPointSize
                 wrapMode:               Text.WordWrap
             }
 
--- a/src/AutoPilotPlugins/APM/APMAirframeComponentSummary.qml
+++ b/src/AutoPilotPlugins/APM/APMAirframeComponentSummary.qml
@@ -20,11 +20,9 @@ FactPanel {
 
     property Fact sysIdFact:        controller.getParameterFact(-1, "FRAME")
 
-
     Column {
-        anchors.fill: parent
-        anchors.margins: 8
-
+        anchors.fill:       parent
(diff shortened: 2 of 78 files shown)


shadow1runner:
Thank you for this. Looking at your Ubuntu screenshot your UI looks great. I, however, have the problem that since this PR the UI is almost unusable:

screenshot from 2016-05-06 09-51-01

I've tried building master and installed 2.9.5 just now - the above screenshot shows the latest release and as you can see it's hardly readable (resolution 1920x1200, but I've also tried smaller ones).
I haven't changed any type-related settings and also deleted my .ini file, without any changes though.

Should I file an issue for that?
Thank you very much.

PS: Windows looks fine, maybe it's really your VM ;)

dogmaphobic:
Font point size is fixed in all platforms with the Exception of Linux. Surprise, surprise. As you've found, different desktop managers have different ideas about what a point size mean. The screen shot above shows a fully maximized window, which exaggerates the effect. Yes, if I maximize a window on a large monitor, things might get out of proportion. In your case, the font size used by QGC is smaller than the system font (when you compare with the menu bar, which is using the system default font size, but not dramatically smaller.

Nonetheless, look at #3341. You can now fine tune the base font point size used by QGC.

shadow1runner:
As if life wouldn't be complicated enough already^^

Thank you

Issue #3255, opened 2016-04-20
The message box in the bottom screen is over important information or with no-contrast


biapalmeiro:
The message box position hides information on the main tabs.

**Plan**
It is over the box with detailed information about the misison.
![screen shot 2016-04-20 at 2 41 08 pm 2](https://cloud.githubusercontent.com/assets/5597786/14687544/8ee9c636-0714-11e6-8c2b-86b5b19f1182.png)

**Fly**
It is over the main call to action of the screen, the ARM button.
![screen shot 2016-04-20 at 2 40 33 pm 2](https://cloud.githubusercontent.com/assets/5597786/14687578/cfdf5ffc-0714-11e6-8cda-b900b43452ea.png)

**Setup > Summary**
Besides it is not hiding any information, the color box match with the screen background makes it difficulty to be identified as a different element.
![screen shot 2016-04-20 at 2 40 53 pm 2](https://cloud.githubusercontent.com/assets/5597786/14687494/456a5eda-0714-11e6-9ec5-e27fba37fc2d.png)

It could be placed just some pixels above, it would also help user to identify it easily. The background color could be changed, so the contrast on summary screen would be improved, like this:
![summary](https://cloud.githubusercontent.com/assets/5597786/14688671/29fd4742-071a-11e6-9ce2-cea79b667fb7.png)

![plan](https://cloud.githubusercontent.com/assets/5597786/14688675/2c9ab6b0-071a-11e6-8bc2-e5b524dfa36f.png)

![fly](https://cloud.githubusercontent.com/assets/5597786/14688681/32847eb2-071a-11e6-993a-68e5b5bf0082.png)

Version verified: v2.9.4-449-g7d77e22 (Development), Mac OS X El Capitan v.11.10.14

@tcanabrava @anselmolsm @cfelipesouza

DonLakeFlyer:
These are supposed to be critical errors so the fact that they overlay other parts of the UI is kind of ok. The user is expected to respond to them by dismissing the popup. Moving it up into the map will continue to overlap different text indicators like "No GPS Lock" and pre-arm messages. Those show up in the center of the map. So no matter where you put it it is going to overlap something. Especially on smaller displays like phones.

The coloring/style does need to be fixed (see the Setup screen visual). But it needs to follow palette colors (we can add news ones if needed). Happy to take suggestions for new drawing visuals/coloring.

tcanabrava:
It's actually not OK when you needs to quickly disarm the drone but there's
[...]

PR #3343, opened 2016-05-06
Change background of the Critical Message Window

dogmaphobic:
This is in response to #3255 

Changed the background color to offer greater contrast.

screen shot 2016-05-06 at 2 20 49 pm
screen shot 2016-05-06 at 2 21 25 pm


dogmaphobic:
Oh, I also changed the border for the *Message Window*, though I still have quite a bit of work on that whole message thing...

screen shot 2016-05-06 at 2 26 17 pm

PR #3346, opened 2016-05-08
Scale increment and decimal fields according to cooked unit, so that …

LorenzMeier:
…the decimal and increments provided by meta data still work in the cooked display.

@dagar This is what you need. Note that a lot of the fixed wing meta for angles doesn't contain increment or meta and thus the angles here only allow natural numbers. We should re-test this against updated meta data.

screen shot 2016-05-08 at 11 26 59

Slider increment also makes sense:

screen shot 2016-05-08 at 11 30 12

### DIFF
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -596,7 +596,8 @@ int FactMetaData::decimalPlaces(void) const
     int incrementDecimalPlaces = unknownDecimalPlaces;
 
     // First determine decimal places from increment
-    double increment = this->increment();
+    double increment = _rawTranslator(this->increment()).toDouble();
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@LorenzMeier Not quite sure what this change does with just a quick look. I assume it's fine. But there are qDebug's that are still in here which are likely to end up being very noisy. Can you remove them?

LorenzMeier:
I stripped them already. It just syncs the increment and decimals to what the scaling from raw -> cooked looks like. So if 0-1 is the raw range and increment is 0.01 with 2 decimals you will get 0-100 % and 1% increment with 0 decimals.

DonLakeFlyer:
Ok, thanks Lorenz.

Issue #3352, opened 2016-05-08
QML QGCButton with Menu rendering incorrectly


DonLakeFlyer:
<img width="495" alt="screen shot 2016-05-08 at 7 17 27 pm" src="https://cloud.githubusercontent.com/assets/5876851/15100998/929d80ba-1551-11e6-9d46-e0508cace873.png">

Probably some bug in there all along exacerbated by the recent font sizing changes.

dogmaphobic:
Fixed in #3354

Issue #3349, opened 2016-05-08
Button sizes on Android 10" (Galaxy Tab 4) very small in flight view


LorenzMeier:
@DonLakeFlyer @dogmaphobic The buttons are small and almost impossible to activate. The slider is much bigger and ok to use, but could be still bigger (still finicky).

dogmaphobic:
Hopefully it's all fixed in #3354. Tested on Nexus (4, 7 and 9), iPhone and iPad (in addition to make sure nothing changed on non mobile platforms).

LorenzMeier:
Will report back once my tablet updates.

LorenzMeier:
Fixed.

PR #3354, opened 2016-05-09
More size adjustments.

dogmaphobic:
In response to #3349 

Made quite a few UI adjustments:

* Larger button sizes for all mobile devices. It used to be only for *Tiny Screens*
* Made touch sensitive area for the "X" to close both the Critical Message Area and the Message Window larger for mobile devices. Normally, only the "X" is touch sensitive, which works fine on desktop platforms with a mouse but quite tricky to hit it using touch.
* Made the DropDown buttons (and the RoundButtons) slightly larger on all mobile devices.
* Fixed the Down Arrow icon on buttons with menus (#3352). The padding and overall width were being computed incorrectly. In addition, several dimensions were hardcoded with small numbers, which become irrelevant on high density devices.
* Made all QGCButtons, QGCRadioButton and QGCTextField slightly larger on all mobile devices. It used to be larger only on *Tiny Screens*

More cosmetics:
* Capitalized several buttons in both Plan View and Fly View to make them all consistent. Some where first word only, others, such as the map types, which come from the map engine, were fully capitalized.
* Adjusted the width of the buttons within the DropDown buttons so they all match instead of each having a different size based on its content. It makes for a cleaner look. Go ahead, just call me OCD :P

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewWidgets.qml
+++ b/src/FlightDisplay/FlightDisplayViewWidgets.qml
@@ -218,9 +218,9 @@ Item {
 
                             QGCButton {
                                 checkable:  true
-                                checked:    _flightMap ? _flightMap.mapType == text : false
+                                checked:    _flightMap ? _flightMap.mapType === text : false
                                 text:       modelData
-
+                                width:      clearButton.width
                                 onClicked: {
                                     _flightMap.mapType = text
                                     _dropButtonsExclusiveGroup.current = null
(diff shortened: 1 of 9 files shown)


dogmaphobic:
Forgot to mention. The "default font point sizes" are defined arbitrarily. I used the Nexus phones/tablets to come up with those numbers (with the assumption those are "reference" Android units). Buttons (and UI in general) are sized based off the default font point size. If things are still small/big, you can change the default size in *General Settings*.

PR #3382, opened 2016-05-20
support MindPX v2 from Airmind.

iZhangHui:


LorenzMeier:
This looks correct, merging.

PR #3383, opened 2016-05-21
PX4: Do mode switch mapping as done onboard

LorenzMeier:
This should help with confused users who complain that the boundaries do not match.


LorenzMeier:
Tested to work as intended. Can be merged once build passes.

DonLakeFlyer:
Looks like this also handles reverse correctly now I think?

LorenzMeier:
Yes, its all 1:1 the same now.

PR #3386, opened 2016-05-22
try QT_FATAL_WARNINGS everywhere

dagar:


DonLakeFlyer:
Nice that this is working again.

dagar:
Unit tests seem to be working consistently on windows, but appveyor is just too slow to add an additional build that takes twice as long.

PR #3390, opened 2016-05-23
Fix FactPanelController not exposing fixedwing property (for safety p…

NaterGator:
…anel icon)

@dagar can you test and confirm this fixes it? If so I'll merge

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSensorsComponentController.h
+++ b/src/AutoPilotPlugins/APM/APMSensorsComponentController.h
@@ -42,7 +42,7 @@ class APMSensorsComponentController : public FactPanelController
 public:
     APMSensorsComponentController(void);
 
-    Q_PROPERTY(bool fixedWing READ fixedWing CONSTANT)
+    Q_PROPERTY(bool fixedWing READ fixedWing)
     
     Q_PROPERTY(QQuickItem* statusLog MEMBER _statusLog)
     Q_PROPERTY(QQuickItem* progressBar MEMBER _progressBar)
--- a/src/AutoPilotPlugins/PX4/SafetyComponent.qml
+++ b/src/AutoPilotPlugins/PX4/SafetyComponent.qml
@@ -490,7 +490,7 @@ QGCView {
                             spacing:                    _margins * 0.5
                             anchors.verticalCenter:     parent.verticalCenter
                             Row {
-                                visible:                _landSpeedMC !== -1
+                                visible:                !controller.fixedWing && (_landSpeedMC !== -1)
                                 QGCLabel {
                                     anchors.baseline:   landVelField.baseline
                                     width:              _middleRowWidth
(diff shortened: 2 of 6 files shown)


DonLakeFlyer:
This is very odd exposure of a single value. Any explanation for it?

NaterGator:
https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/PX4/SafetyComponent.qml#L367 and associated are broken. Similarly done here: https://github.com/mavlink/qgroundcontrol/blob/b8d25964fd74f78713ff795547f12f8289890829/src/AutoPilotPlugins/PX4/SensorsComponentController.h#L47

dagar:
@DonLakeFlyer with a pixhawk configured as a fixed wing the PX4 safety screen is showing multicopter icons and settings.

screenshot from 2016-05-23 13 03 45

NaterGator:
It is probably an error to specify CONSTANT in the Q_PROPERTY decl per [exposing properties](http://doc.qt.io/qt-5/qtqml-cppintegration-exposecppattributes.html#exposing-properties)

DonLakeFlyer:
```
bool SensorsComponentController::fixedWing(void)
{
    switch (_vehicle->vehicleType()) {
        case MAV_TYPE_FIXED_WING:
        case MAV_TYPE_VTOL_DUOROTOR:
        case MAV_TYPE_VTOL_QUADROTOR:
        case MAV_TYPE_VTOL_TILTROTOR:
        case MAV_TYPE_VTOL_RESERVED2:
        case MAV_TYPE_VTOL_RESERVED3:
        case MAV_TYPE_VTOL_RESERVED4:
        case MAV_TYPE_VTOL_RESERVED5:
            return true;
        default:
            return false;
    }
}
```

SensorComponentController shouldn't have it's own version of this code. It should just:
```
bool SensorsComponentController::fixedWing(void)
{
    return _vehicle->fixedWing();
}
```

This code may be old, prior to Vehicle::fixedWing being available.

I think the CONSTANT should be ok, since it works fine on Vehicle. At least it used to be working ok to not show airspeed cal for example on multirotor. Look for other places that use Vehicle::fixedWing to see if they are broken as well.

NaterGator:
Constant is not ok and should not be on vehicle either since it is derived from a parameter which takes time to sync. Agreed SensorController should not duplicate the functionality offered by Vehicle.

DonLakeFlyer:
Vehicle type comes in from the Vehicle constructor not a parameter. So it is always correct and fixed.

NaterGator:
Fair. Then constant is only not okay on controllers that resolve the type from a member vehicle pointer that can change at any point.

DonLakeFlyer:
> Then constant is only not okay on controllers that resolve the type from a member vehicle pointer that can change at any point.

It's ok on SensorsComponentController. FactPanelController's never have the vehicle change out from under them. That's not how they work. Vehicle is set in the constructor of a FactPanelController. Something else is wrong here. Did you step through to make sure the vehicle type was coming through correctly from the firmware?

DonLakeFlyer:
Is there an Issue in for the problem so I can try to repro?

dagar:
@DonLakeFlyer 
Just connect PX4 fixed wing and look at the safety screen. You'll see the multicopter icons and the MC only setting landing velocity.

DonLakeFlyer:
Fixed: https://github.com/mavlink/qgroundcontrol/pull/3392

PR #3397, opened 2016-05-24
PX4 misc

dagar:
- safety setup hide land speed setting (MC only)
- only show appropriate modes for FW/MC (is there a better way to do this?)
- don't require calibrated sensors in HIL mode


DonLakeFlyer:
The better way to handle HIL would be to not add the Sensors component to the list when HIL by modifying PX4AutoPilotPlugin since calibration in HIL doesn't make any sense any way.

dagar:
Moved the hil check to PX4AutoPilotPlugin as suggested.

FYI the safety screen still shows the multicopter stuff if you upgrade the firmware, set a fixed wing config, auto restart then auto reconnect.

DonLakeFlyer:
Does this cause problems with VTOL? See #3413 for discussion.

dagar:
VTOL will get all the modes, which is correct. Hiding MC land speed (if a
VTOL is a fixedWing) is not. Once you decide what to do in #3413 I'll fix
it.

Issue #3400, opened 2016-05-25
Change Settings pane to work like Setup/Plan/Fly


DonLakeFlyer:
- I already remove the X to close. 
- No transparent section to close.
- Just switch between panels by clicking top level buttons

DonLakeFlyer:
Also General page should be automatically select on first load.

dogmaphobic:
Done in #3406

PR #3406, opened 2016-05-26
Preferences panel

dogmaphobic:
In response to #3400.

The preferences panel is now promoted to a top level panel.


DonLakeFlyer:
Thanks Gus

dogmaphobic:
I still need to rename `MainWindowLeftPanel.qml`. I didn't do it now because there were too many pending things that, by changing its name I could potentially create merge conflicts.

PR #3427, opened 2016-05-28
Update to MAVLink 2.0

LorenzMeier:
@DonLakeFlyer This flicks the switch for MAVLink 2.0 and likely will need compile and runtime fixes. Most notably it needs to be configured to default to MAVLink 1. Probably not for v3.0, but I kind of start to need this to keep development going.


DonLakeFlyer:
What exactly does this do? Is master mavlnk 1.0 now set to run 2.0 instead or something?

LorenzMeier:
Yep. You are getting the 2.0 headers which also support 1.0. I already switched PX4 to this.

DonLakeFlyer:
So can this be merged pre-3.1 or not? Or does this pull need to wait?

LorenzMeier:
Once it builds and is tested it can go in any time. It seems to work fine and is a decent update

DonLakeFlyer:
I assume ok to close?

PR #3436, opened 2016-05-29
Add RC-radio-like exponential to joystick RPY axes

rjehangir:
This adds an exponential response to the joystick roll, pitch, yaw axes for better usability and better similarity to an actual RC radio.

The exponential equation is:

```
float expo = -0.35;

roll = -expo*powf(roll,3) + (1+expo)*roll;
[etc..]
```
The exponential factor is very similar to RC radio expo settings, which usually go from 0% to -50% or so. The output remains normalized regardless of the exponential factor. The default is hardcoded at -35%, which gives a good response for joystick axes.

I made a quick illustration of what this looks like from 0% (linear) to -50% (strong exponential):

https://www.dropbox.com/s/86l66vb2tl7syjv/expo-qgc-20160528.mov?dl=0

Note: This only applies to the physical joysticks and not to the virtual joysticks. I can fix that if you'd like.

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -286,7 +286,7 @@ void Joystick::run(void)
             // Exponential (0% to -50% range like most RC radios)
             // 0 for no exponential
             // -0.5 for strong exponential
-            float expo = -0.35;
+            float expo = -0.35f;
 
             // Calculate new RPY with exponential applied
             roll =      -expo*powf(roll,3) + (1+expo)*roll;
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -22,6 +22,7 @@ const char* Joystick::_settingsGroup =              "Joysticks";
 const char* Joystick::_calibratedSettingsKey =      "Calibrated";
 const char* Joystick::_buttonActionSettingsKey =    "ButtonActionName%1";
 const char* Joystick::_throttleModeSettingsKey =    "ThrottleMode";
+const char* Joystick::_exponentialSettingsKey =     "Exponential";
 
 const char* Joystick::_rgFunctionSettingsKey[Joystick::maxFunction] = {
     "RollAxis",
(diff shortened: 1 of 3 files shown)


LorenzMeier:
@rjehangir  Please note that all new contributions to QGC have to be made under a dual license including Apache 2.0 + GPLv3. Just so you're aware that this PR now has to fulfil this requirement.

rjehangir:
Lorenz, that's completely fine and I confirmed to your email for my past contributions. Do I need to do anything to comply for this PR?

LorenzMeier:
No, just a note for the record. You proposed it before we switched the codebase today and I didn't want to have PRs open from before switching which were merged after without having made this very transparent.

rjehangir:
@DonLakeFlyer I think that would be a good idea. In fact, I didn't think about how this would affect the "Advanced" joystick output options that don't use MANUAL_CONTROL. This may not be appropriate in those situations.

What do you think about this overall? I think it makes sense to have this feature in QGC since the joystick is supposed to replicate the RC radio, but I also understand if you want to keep the output as simple as possible.

I'm happy to implement this on the ArduSub side instead of in QGC if you are hesitant about it.

NaterGator:
Why not just expose the exponential factor as a configurable option in the joystick setup page or general settings?

DonLakeFlyer:
> Why not just expose the exponential factor as a configurable option in the joystick setup page or general settings?

Yup. Just another Joystick setting in there.

rjehangir:
If we do that, should it be a single setting that applies to roll, pitch, and yaw, or should it be on a per-axis basis, like an RC radio?

DonLakeFlyer:
I'd keep a single setting. If we want fancy we can do that later.

rjehangir:
Okay. I'll work on adding a setting on the joystick page.

DonLakeFlyer:
@rjehangir Rusty, this is still here waiting for updates as well.

rjehangir:
Hello, just added the checkbox for enabling and disabling the exponential on the joystick. The checkbox is right above the "Advanced Settings" checkbox.

Issue #3391, opened 2016-05-23
Arm button should not be hidden if no GPS


DonLakeFlyer:
This isn't something that QGC should be handling. For example on ArduPilot you can turn off no-GPS arm failure with a circuit breaker. If I have that turned off I should be able to arm from the button. If firmware wants to fail arm with no GPS it should fail the arm and return a user friendly error. The current QGC implementation is very firmware specific to a specific stack/vehicle type. Another example is that ArduSub has no GPS I would guess. Now you can't arm that either.

DonLakeFlyer:
Other option is mavlink generic mechanism to determine availability of arming. But I think fail with message is likely better since you can give user friendly text back like "Cannot arm without GPS lock" or something like that.

SamChenzx:
@DonLakeFlyer  I favor that GQC allows users to arm motors without GPS lock. Besides, can you also check these two issues, [#3385](https://github.com/mavlink/qgroundcontrol/issues/3385) and [#3384](https://github.com/mavlink/qgroundcontrol/issues/3384) thanks!

DonLakeFlyer:
@LorenzMeier Are you ok with changing this back to the way it was before?

LorenzMeier:
In general allowing the user to arm the system without allowing him to take off (because no position estimate is present) seems like a terrible idea. We can clean up the check, but I would like to avoid having users arming the system before they are really able to fly.

DonLakeFlyer:
Discussion in #3460

PR #3444, opened 2016-05-29
Allow arm with no GPS lock

DonLakeFlyer:
Vehicle should fail arm command if not allowed as opposed to having QGC make the rules for when arm is allowed/disallowed. Otherwise it gets much too complicated as to whether the vehicle even requires a gps. Or if it does whether it has a circuit breaker set for it to allow arm without gps lock.

@LorenzMeier This reverses your change.

Fix for #3391

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewWidgets.qml
+++ b/src/FlightDisplay/FlightDisplayViewWidgets.qml
@@ -459,7 +459,7 @@ Item {
 
                 QGCButton {
                     text:       (_activeVehicle && _activeVehicle.armed) ? (_activeVehicle.flying ? qsTr("Emergency Stop") : qsTr("Disarm")) :  qsTr("Arm")
-                    visible: (_activeVehicle && _activeVehicle.coordinateValid)
+                    visible:    _activeVehicle
                     onClicked:  _guidedModeBar.confirmAction(_activeVehicle.armed ? (_activeVehicle.flying ? _guidedModeBar.confirmEmergencyStop : _guidedModeBar.confirmDisarm) : _guidedModeBar.confirmArm)
                 }
 


LorenzMeier:
I can see where you're coming from but the user model of this is terrible. Its the UI's responsibility to set user expectations, not let him run into then denied actions.

This should not be merged without filing an issue to track the user model failure case.

LorenzMeier:
@DonLakeFlyer Can we resolve this instead to give the vehicle an OK to arm flag and an OK to fly tablet flag and default both to true? I would like to extend the custom status in the PX4 heartbeat to include the flag.

What I would like to have from the user model perspective is a button which says:

If no position info available, just show a label:

```[No position lock: Only manual RC flying]```

If position lock available but in manual flight mode and landed (no arming button):

```[Takeoff]```

Right now its way too manual and has no user model at all. It just mimics the RC functionality, which for my standards is not good enough for a tablet based UI. Too many steps. @dogmaphobic FYI.

DonLakeFlyer:
> Can we resolve this instead to give the vehicle an OK to arm flag and an OK to fly tablet flag and default both to true? I would like to extend the custom status in the PX4 heartbeat to include the flag.

Need way more than that if you want a good user model. We need to clean up the concept of pre-arm check failures as a standard mavlink mechanism. Just saying you can't arm, but not being able to tell the user why (GPS lock is just a single case of possible arm failure) is bad. I think we should discuss offline and I can also you take you through how ArduPilot+Solo handles this. Which is a special case in QGC, but it could be replicated to the PX4 side. I've had this pre-arm problem in the back of my head for a while now, because it an issue with PX4/ArduPilot handling disparity.

DonLakeFlyer:
I'll enter an Issue to capture the per-arm thing.

PR #3455, opened 2016-06-01
Recognize joystick hats and append to the button list

rjehangir:
This PR addresses the issue that joystick/gamepad hats are considered to be "hats" in Windows but "buttons" on Mac and Linux. This causes inconsistent numbers of joystick buttons between OSes. 

This PR causes joystick hats to be detected and appended to the normal button list. Only the up, down, left, right buttons are recognized and appear in the same order that they are detect on Mac/Linux. 

A few things worth mentioning:
- The other buttons are still detected in different orders on different OSes so this doesn't provide complete consistency
- The center button (like the "X" on the Xbox controller) is not detected on Windows. This will be fixed in SDL2
- This does not affect the indices of any buttons so it should not affect any existing joystick performance or setup
- This has been tested on Mac and Windows with Xbox 360 and Logitech 310 controllers
- My motivation for doing this was to provide additional buttons to ArduSub software but it should be broadly useful

Let me know if you have any thoughts or questions. I'm happy to discuss.

Also, I committed this before the repo was switched to the new dual licenses and I'm fine with the new license.


DonLakeFlyer:
Rusty: I want to get the JOystick refactor pull in before this. That may affect this code and may require a little cleanup.

rjehangir:
@DonLakeFlyer Okay, understandable. Any idea on the timeline for that refactor? I've got several people bugging me about this feature.

DonLakeFlyer:
The joystick restructure is in now: https://github.com/mavlink/qgroundcontrol/pull/3459

rjehangir:
Okay. I'll rebase and test/edit based on these changes.

gregd72002:
@rjehangir it should be reasonably easy to fix. The main change is the the SDL code has been moved from Joystick.cc into JoystickSDL.cc

rjehangir:
This PR got all messed up after rebasing. I'm closing it and I'll make a new one.

PR #3457, opened 2016-06-01
GPS RTK: use GPS_RTCM_DATA mavlink message

bkueng:
Switch to GPS_RTCM_DATA to support fragmentation of long RTCM messages, and some other RTK-related stability improvements.

@DonLakeFlyer @LorenzMeier FYI


DonLakeFlyer:
@bkueng Just double checking that the submodule update is correct. That ends up by error in pulls quite often.

bkueng:
yes it's correct

Issue #3456, opened 2016-06-01
QGC: Mission Edit problems.


tubeme:
Hello everybody,

Just to report some problems when programming large missions.

In the bottom bar, where I get info on the alt and distance of waypoint, there is no way I could see after point 58. Probably this number depends on the resolution. I'm fullHD so 58 points :)

Also when the waypoints on the right, fill the height of the window, each new waypoint is hidden, under the window so you cannot see rely the heidht of the waypoint except when you move the mouse and slide the waypoints.

You can see:

![mission sapareva](https://cloud.githubusercontent.com/assets/12617619/15709813/686c4684-280f-11e6-8d4b-3bf000224678.png)

Also when you make new waypoints it did not remember the height from the previous one if you made some corrections. When you make corrections to a waypoint you have to explicitly hit TAB in order to move the focus on another box. Then it remembers the previous waypoint and places its altitude as height value.

DonLakeFlyer:
Height map athe the bottom should be left/right scrollable with the mouse.

NaterGator:
@DonLakeFlyer I think maybe @tubeme is suggesting we scroll that flickable to the inserted item's position too.

DonLakeFlyer:
Ah, good point.

tubeme:
@DonLakeFlyer It is good when you have mouse, but on the field I don't use one... Also I want to see at this bar and on the right waypoints ribbon the current waypoint that I'm editing now. It was hell to set up this 78 point mission...

NaterGator:
@DonLakeFlyer is there a reason we can add onValueChanged handling [here](https://github.com/mavlink/qgroundcontrol/blob/c29de21aa53fb733ef995af8f0e944757934b2a5/src/FactSystem/FactControls/FactTextField.qml#L25)? Obviously we should use the validator in a non-intrusive way so that we don't set invalid values to the fact and don't prevent the user from continuing to edit the field to fix the value. 

That would address @tubeme's other outstanding concern.

DonLakeFlyer:
> It is good when you have mouse, but on the field I don't use one

How do you use QGC without a mouse? How are you selecting things? You can scroll with the a trackpad or touch as well.

DonLakeFlyer:
> Also when you make new waypoints it did not remember the height from the previous one if you made some corrections. When you make corrections to a waypoint height you have to explicitly hit TAB in order to move the focus on another box. Then it remembers the previous waypoint and places its altitude as height value.

[...]

PR #3458, opened 2016-06-01
Tweaks to mission editor model views. Helps to address #3456

NaterGator:
Let Qt manage scrolling the respective views to the right spot. It actually seems to do a decent job. There is still a bit of raciness with the MissionItemStatus view positioning when adding new items, but it is infrequent.

### DIFF
--- a/src/MissionEditor/MissionItemStatus.qml
+++ b/src/MissionEditor/MissionItemStatus.qml
@@ -74,53 +74,43 @@ Rectangle {
             QGCLabel { text: _azimuthText }
         }
 
-        QGCFlickable {
+        ListView {
+            id:    statusListView
+            model: missionItems
+            highlightMoveDuration:  250
             anchors.leftMargin:     _margins
             anchors.rightMargin:    _margins
             anchors.top:            parent.top
             anchors.bottom:         parent.bottom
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
Let me do some testing with this. I've been fighting with the item list scrolling forever. Each way seemed to have it's own different set of problems!

DonLakeFlyer:
For #3456

DonLakeFlyer:
Not sure if ListView can handle invisible items within it. See here:
screen shot 2016-06-02 at 10 01 23 am
The takeoff item is still taking up empty space in the height list.

src/MissionEditor/MissionItemStatus.qml
```diff
@@ -74,53 +74,43 @@ Rectangle {
             QGCLabel { text: _azimuthText }
         }
 
-        QGCFlickable {
+        ListView {
+            id:    statusListView
+            model: missionItems
```
DonLakeFlyer: Can you clean up the spacing alignment here. I know I'm a pain.

NaterGator: Not a pain, if it isn't consistent it will become chaos. I'll open a new PR with fewer typos in the commit message too.

PR #3459, opened 2016-06-01
Joystick restructure

gregd72002:
This is based on discussion in here: https://github.com/mavlink/qgroundcontrol/pull/3305

- introducing Joystick base class (inherited by specific joystick implementations)
- introducing JoystickSDL (sdl implementation for Joystick based on previous Joystick code)
- activating JoystickManager and JoystickConfigManager for all builds (for iOS joystick discovery is not defined and as such will return no joysticks) - see setToolbox function


gregd72002:
ac4ad3b is fixed in 6e33df1

PR #3464, opened 2016-06-02
Obligate LinkInterfaces to implement getLinkConfiguration. …

NaterGator:
11ab9aa0311fc0efbb03d3e387480f390f903ea7 introduced a bug where TCP links would cause a crash loading the sensors page.

Found by @dagar


NaterGator:
Ah, didn't see that image before. I'll fix that now...

PR #3485, opened 2016-06-06
appveyor install gstreamer

dagar:
TODO
 - [x] copy gstreamer libs to installer


dagar:
@DonLakeFlyer windows gstreamer is working.

DonLakeFlyer:
Is this good to merge now?

dagar:
Yes, I've tested the installer on windows. Just be aware that we're copying a hand picked list of gstreamer dlls into the installer. I don't know how brittle this is going to be longer term. 
Also please check that I got rid of the proper flightgear directory.

DonLakeFlyer:
I don't know much about flightgear so not sure I can verify that.

DonLakeFlyer:
Seems correct from the .pri changes to resources.

dagar:
It was the exact same directory of flightgear stuff duplicated at the top level and in resources. If something broke it'll be a trivial fix.

Issue #2641, opened 2016-01-16
Crash on Disconnect from Vehicle


DonLakeFlyer:
- Direct connect Vehicle via USB to Android tablet
- Wait for everything to crank up
- Power down Vehicle
- Click Disconnect
- Crash

DonLakeFlyer:
Caused by:
```
SerialLink::~SerialLink()
{
    // Disconnect link from configuration
    _config->setLink(NULL);
    _disconnect();
    if(_port) delete _port;
    _port = NULL;
}
```
The setLink call emits a signal while in the destructor which confuses Qt since some objects are in the middle of going away.

@dogmaphobic Is it necessary to clear the link here since the whole object is going away anyway?

dogmaphobic:
"Fixed" in #3500

PR #3500, opened 2016-06-07
Removing unused signal.

dogmaphobic:
According to #2641, the link change signal was causing a crash on shutdown. Since this signal is not used anywhere, I removed it.


DonLakeFlyer:
I think it's angry there is a WRITE but no NOTIFY which doesn't really make sense. I was testing on OSX.

PR #3511, opened 2016-06-09
Recognize joystick hats and append to the button list (rev 2)

rjehangir:
(Note: This is a new version of PR #3455 that has been updated for the new joystick code.)

This PR addresses the issue that joystick/gamepad hats are considered to be "hats" in Windows but "buttons" on Mac and Linux. This causes inconsistent numbers of joystick buttons between OSes.

This PR causes joystick hats to be detected and appended to the normal button list. Only the up, down, left, right buttons are recognized and appear in the same order that they are detect on Mac/Linux.

A few things worth mentioning:

- The other buttons are still detected in different orders on different OSes so this doesn't provide complete consistency
- The center button (like the "X" on the Xbox controller) is not detected on Windows. This will be fixed in SDL2
- This does not affect the indices of any buttons so it should not affect any existing joystick performance or setup
- This has been tested on Mac and Windows with Xbox 360 and Logitech 310 controllers
- My motivation for doing this was to provide additional buttons to ArduSub software but it should be broadly useful

Let me know if you have any thoughts or questions. I'm happy to discuss.

Also, I committed this before the repo was switched to the new dual licenses and I'm fine with the new license.

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -30,11 +30,14 @@ const char* Joystick::_rgFunctionSettingsKey[Joystick::maxFunction] = {
     "ThrottleAxis"
 };
 
-Joystick::Joystick(const QString& name, int axisCount, int buttonCount, MultiVehicleManager* multiVehicleManager)
+Joystick::Joystick(const QString& name, int axisCount, int buttonCount, int hatCount, MultiVehicleManager* multiVehicleManager)
     : _exitThread(false)
     , _name(name)
     , _axisCount(axisCount)
     , _buttonCount(buttonCount)
+    , _hatCount(hatCount)
+    , _hatButtonCount(4*hatCount)
+    , _totalButtonCount(_buttonCount+_hatButtonCount)
     , _calibrationMode(CalibrationModeOff)
     , _rgAxisValues(NULL)
     , _rgCalibration(NULL)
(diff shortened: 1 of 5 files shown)


TC3:
Hi Don,
I would greatly appreciate it if you would expedite Rusty's gamepad/joystick PR so I can get on with the testing of my ROV which is built with Blue Robotics hardware and uses QGC as my GCS.
Regards,
Tom Coyle AVD

src/Joystick/Joystick.cc
```diff
@@ -30,11 +30,14 @@ const char* Joystick::_rgFunctionSettingsKey[Joystick::maxFunction] = {
...
 };
 
-Joystick::Joystick(const QString& name, int axisCount, int buttonCount, MultiVehicleManager* multiVehicleManager)
+Joystick::Joystick(const QString& name, int axisCount, int buttonCount, int hatCount, MultiVehicleManager* multiVehicleManager)
     : _exitThread(false)
     , _name(name)
     , _axisCount(axisCount)
     , _buttonCount(buttonCount)
+    , _hatCount(hatCount)
+    , _hatButtonCount(4*hatCount)
```
DonLakeFlyer: Where does this come from? Seems a bit magic. Hats can have buttons, but the actual number is not reported?

DonLakeFlyer: Oh, wait. I was thinking hats were additional axes, but their just additional buttons. This is ok.

PR #3513, opened 2016-06-09
Making tile cache description not mandatory.

dogmaphobic:
For the time being, I'm making sure the description is not empty (so it will continue to work with existing databases).

### DIFF
--- a/src/QtLocationPlugin/QGCTileCacheWorker.cpp
+++ b/src/QtLocationPlugin/QGCTileCacheWorker.cpp
@@ -302,6 +302,9 @@ QGCCacheWorker::_getTileSets(QGCMapTask* mtask)
         while(query.next()) {
             QString name = query.value("name").toString();
             QString desc = query.value("description").toString();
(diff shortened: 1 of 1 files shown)


dogmaphobic:
Build timed out because it took forever to download GStreamer. Restarting...

dagar:
If that keeps happening I'll move all the dependencies to s3.

dogmaphobic:
> If that keeps happening I'll move all the dependencies to s3.

It's the first time I see it. It must have been a network issue.

DonLakeFlyer:
OSX streamer package download just went in last week. So not much time to watch it screw up.

DonLakeFlyer:
@dagar I think it make sense to move it to s3 no matter what though. It will likely speed up builds.

dogmaphobic:
Weird, timed out again?

dagar:
Wow, these are some slow downloads. They'll be on s3 shortly.

PR #3540, opened 2016-06-12
travis-ci install gstreamer for linux

dagar:
This was lost on travis-ci when switching to their newer infrastructure with the apt add-on instead of just using the command line to install packages.


dagar:
TODO:
 - [ ] update AppImage
 - [ ] dependencies for users (or copy to package)

PR #3555, opened 2016-06-13
Save vehicle type for APM mock link

makrebs:
Settings of mock link take now effect (Arducopter / Arduplane)


DonLakeFlyer:
Thanks

Issue #3378, opened 2016-05-16
Virtual Joystick can't control PX4 flight stack.


SamChenzx:
I'm using Qgroundcontrol 2.9.7b as my ground station and I want to use the virtual joystick without a real RC or joystick radio by setting COM_RC_IN_MODE to 0(Joystick/No RC Checks). I can arm motors by command line. The problem is when I attempt to takeoff my quadrotor in manual flight mode which runs PX4 flight code, no matter how I move the virtual joystick, the PWM output seems always be 0, I want to know that do I need extra setup before using virtual joystick, or virtual joystick can not control a real quadrotor?

dagar:
Try COM_RC_IN_MODE 1

SamChenzx:
When I change COM_RC_IN_MODE to 1, as I don't have a real RC radio
connected to quadrotor, QGC warns that:
Detected 0 radio channels. To operate PX4, you need at least 5channels.
And the Radio menu turn into red color. Does it works?

andrejpan:
Hey,

I set up pixracer and QGC 2.9.7 with psp3 gamepad. COM_RC_IN_MODE has settings Joystick/No RC Checks (I do not see any number with this parameter any more). There is also settings for a joystick, QGC should recognize type of gamepad and calibration should be also made. 
![gamepad_settings](https://cloud.githubusercontent.com/assets/1015265/15367618/3e770e62-1d2a-11e6-85a1-4dcbc45e6c81.png)

And I think this is it. In few days I will also try QGC 2.9.7beta and I can report if gamepad control still works.

LorenzMeier:
@SamChenzx You need to arm via QGC to control the system. You cannot arm it with the virtual joysticks. It sounds like you're trying to do that.

andrejpan:
Currently I am arming/disarming motors with virtual joystick with no problem, should we make another issue?

SamChenzx:
@LorenzMeier Yes, that's the problem, I can't arm it with the virtual joysticks. Besides, seems can't arm it via QGC, as in the screen shot, QGC just start with a white circle with no arm/disarm option.
<img width="1440" alt="screen shot 2016-05-18 at 12 51 16 pm" src="https://cloud.githubusercontent.com/assets/10758544/15373016/2f7d4e9e-1cf8-11e6-88f9-fdcdd1987940.png">

LorenzMeier:
[...]

PR #3570, opened 2016-06-14
Virtual joystick: Throttle no longer spring loaded to center

DonLakeFlyer:
Fir for #3378


SamChenzx:
@DonLakeFlyer Thanks I'll try.

PR #3574, opened 2016-06-14
PX4 mode cleanup

dagar:
I made the PX4 mode naming consistent with the COM_FLTMODE* params. I also pulled in the px4_custom_mode.h header straight from PX4 as a crude way to stay in sync.
https://github.com/PX4/Firmware/blob/master/src/modules/commander/px4_custom_mode.h

https://github.com/PX4/Firmware/issues/4796


DonLakeFlyer:
Crap. Mac builds are exceeding the job time limit.

PR #3582, opened 2016-06-16
Fix memory leak during dragging waypoint

NaterGator:
This function is creating the objects on the heap... hence it is owned by the MissionController. If we don't delete them they are leaked.


NaterGator:
Going to be replaced by an overall faster model

PR #3584, opened 2016-06-16
Rework mission lines + various mission-related bug fixes

NaterGator:
This goes a long way towards making waypoint dragging smooth on my machine. Even still with all the QML bindings and aggressive updating I still manage to peg a core while dragging around a waypoint in a large mission, but the update rate is much much higher and things feel significantly more responsive. 

Also included:

- Fixes the memory leak in the waypoint line model 
- Fixes waypoints not being inserted properly from the hamburger menu
- Fixes abbreviations not updating properly when sequence numbers were recomputed 
- Other performance optimizations / tweaks to mission related items

### DIFF
--- a/src/MissionEditor/MissionEditor.qml
+++ b/src/MissionEditor/MissionEditor.qml
@@ -81,6 +81,7 @@ QGCView {
         } else {
             controller.loadMissionFromFilePicker()
             fitViewportToMissionItems()
+            _currentMissionItem = _visualItems.get(0)
         }
     }
 
@@ -185,6 +186,7 @@ QGCView {
             onFilenameReturned: {
                 controller.loadMissionFromFile(filename)
                 fitViewportToMissionItems()
+                _currentMissionItem = _visualItems.get(0)
             }
         }
     }
--- a/src/MissionEditor/MissionEditor.qml
+++ b/src/MissionEditor/MissionEditor.qml
@@ -500,6 +500,11 @@ QGCView {
                                 controller.removeMissionItem(index)
                             }
 
+                            onInsert: {
+                                var sequenceNumber = controller.insertSimpleMissionItem(editorMap.center, i)
+                                setCurrentItem(sequenceNumber)
+                            }
+
                             onMoveHomeToMapCenter: controller.visualItems.get(0).coordinate = editorMap.center
 
                             Connections {
--- a/src/MissionEditor/MissionEditor.qml
+++ b/src/MissionEditor/MissionEditor.qml
@@ -501,7 +501,7 @@ QGCView {
                             }
 
                             onInsert: {
-                                var sequenceNumber = controller.insertSimpleMissionItem(editorMap.center, i)
+                                var sequenceNumber = controller.insertSimpleMissionItem(editorMap.center, insertAfterIndex)
                                 setCurrentItem(sequenceNumber)
                             }
 
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -629,7 +629,7 @@ void MissionController::_recalcWaypointLines(void)
     double homePositionAltitude = homeItem->coordinate().altitude();
     minAltSeen = maxAltSeen = homeItem->coordinate().altitude();
 
-    _waypointLines.clear();
+    _waypointLines.clearAndDeleteContents();
 
     bool linkBackToHome = false;
     for (int i=1; i<_visualItems->count(); i++) {
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -616,8 +616,8 @@ void MissionController::_recalcWaypointLines(void)
 
     qCDebug(MissionControllerLog) << "_recalcWaypointLines";
 
-    CoordVectHashTable old_table = _lines_table;
-    _lines_table.clear();
+    CoordVectHashTable old_table = _linesTable;
+    _linesTable.clear();
     _waypointLines.clear();
 
     bool linkBackToHome = false;
@@ -639,7 +639,7 @@ void MissionController::_recalcWaypointLines(void)
                 if (lastCoordinateItem != homeItem || (showHomePosition && linkBackToHome)) {
                     if (old_table.contains(pair)) {
                         // Do nothing, this segment already exists and is wired up
-                        _lines_table[pair] = old_table.take(pair);
+                        _linesTable[pair] = old_table.take(pair);
                     } else {
                         // Create a new segment and wire update notifiers
                         auto linevect       = new CoordinateVector(lastCoordinateItem->isSimpleItem() ? lastCoordinateItem->coordinate() : lastCoordinateItem->exitCoordinate(), item->coordinate(), this);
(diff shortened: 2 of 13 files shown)


DonLakeFlyer:
Doing some testing on this

src/MissionEditor/MissionEditor.qml
```diff
@@ -498,6 +500,11 @@ QGCView {
                                 controller.removeMissionItem(index)
                             }
 
+                            onInsert: {
+                                var sequenceNumber = controller.insertSimpleMissionItem(editorMap.center, i)
+                                setCurrentItem(sequenceNumber)
+                            }
```
DonLakeFlyer: Was the insert command just lost along the way somewhere? Strange. While you're in here, could you change the signal for insert to have a more meaningful parameter than "i". Something like insertAfterIndex or other.

src/MissionManager/MissionController.h
```diff
@@ -112,6 +117,7 @@ private slots:
     QmlObjectListModel* _visualItems;
     QmlObjectListModel* _complexItems;
     QmlObjectListModel  _waypointLines;
+    CoordVectHashTable  _lines_table;
```
DonLakeFlyer: CamelCase for members

src/MissionManager/MissionController.cc
```diff
@@ -848,12 +927,6 @@ void MissionController::_deinitVisualItem(VisualMissionItem* visualItem)
     disconnect(visualItem, 0, 0, 0);
 }
 
-void MissionController::_itemCoordinateChanged(const QGeoCoordinate& coordinate)
-{
-    Q_UNUSED(coordinate);
-    _recalcWaypointLines();
-}
-
```
DonLakeFlyer: You're just changing the coordinate in the model now right when the coordinate changes as opposed to rebuilding? Is that correct? This was specifically what didn't work before. I needed to do a full rebuild and model change to get coordinates to update.

src/MissionManager/MissionController.h
NaterGator: Old habits... will fix

DonLakeFlyer:
The abbeviations don't seem to be working. See takeoff listed as "1" on the map:
screen shot 2016-06-17 at 9 52 34 am

NaterGator:
@DonLakeFlyer
> You're just changing the coordinate in the model now right when the coordinate changes as opposed to rebuilding? Is that correct? This was specifically what didn't work before. I needed to do a full rebuild and model change to get coordinates to update.

That's part of it. I'm creating a hash-table of all pairs of VisualMissionItems that require a line segment between them that connect to a CoordinateVector.  Previously CoordinateVectors were created/destroyed on the heap exhaustively for each line update operation. Now I reuse any CoordinateVector I can, and bind minimally so the fewest possible QML properties have to be modified for each update. I also batch all of the QmlObjectListModel changes into one operation (modelReset). 

I am not seeing any issues with property updates failing with this approach on Windows, but look forward to hearing if there are issues on other platforms.

NaterGator:
> The abbeviations don't seem to be working. See takeoff listed as "1" on the map:

This bug is unrelated and is because we're not using abbreviation [here](https://github.com/mavlink/qgroundcontrol/blob/fdb13156c0caa95a587775f9c60e4173e13b7a27/src/MissionEditor/MissionEditor.qml#L429)

DonLakeFlyer:
Tested on OSX, Android and iOS. Looks good.

DonLakeFlyer:
@NaterGator How does this statement on MapPolyLine:
"The coordinates on the path cannot be directly changed after being added to the Polyline. Instead, copy the path into a var, modify the copy and reassign the copy back to the path."

Relate to your implemenation?

DonLakeFlyer:
Wait you're re-using the coordinate objects, but still rebuilding the full object list pointing to the same coordinate objects right? So this fits the paradigm for what you can and can't do with MapPolyLine. Ok, now I get it correctly.

NaterGator:
It is quite subtle. Because the waypointLines QmlObjectListModel is connected to the MissionLineView  and I reset the model [here](https://github.com/mavlink/qgroundcontrol/pull/3584/files#diff-b4069298d9baf9c733de5b85f4d647a6R673) the old delegated MapPolyLines are destroyed and recreated. 

For dragging though we don't reset the entire model; the VisualMissionItem [signals](https://github.com/mavlink/qgroundcontrol/pull/3584/files#diff-b4069298d9baf9c733de5b85f4d647a6R649) an update to the CoordinateVector which is an item in the MissionControll:waypointLines model. I am actually not positive this implementation adheres to that warning... let me see if I can address that concern.

DonLakeFlyer:
> I am actually not positive this implementation adheres to that warning... let me see if I can address that concern.

It works though everywhere I tested. This is specifically what didn't work in 5.4. I wonder if we just go with it.

DonLakeFlyer:
Maybe you can update individual coordinates from Qml. But maybe since you are signalling correctly in C++ it all works?

NaterGator:
Maybe, but we may be skirting the rules there. 

[Here's](http://code.qt.io/cgit/qt/qtlocation.git/tree/src/imports/location/qdeclarativepolylinemapitem.cpp?h=v5.5.1#n548) what is updated when we replace the path. Interestingly we also have [this](http://code.qt.io/cgit/qt/qtlocation.git/tree/src/imports/location/qdeclarativepolylinemapitem.cpp?h=v5.5.1#n471) internal slot that does the same thing. Still, probably splitting hairs to save a little tiny bit of performance. Instead of creating the path array in QML [here](https://github.com/mavlink/qgroundcontrol/pull/3584/files#diff-031ab30ad8f5bf45927744f119764c9aL69) I'm going to try exposing the path for MapPolyline  as a member on CoordinateVector so hopefully the QML binding will take care of calling setPath for us. The trick is going to be getting automatic type conversion correct.

DonLakeFlyer:
> Instead of creating the path array in QML here I'm going to try exposing the path for MapPolyline as a member on CoordinateVector so hopefully the QML binding will take care of calling setPath for us.

This is probably why it works for you and I couldn't get it to work.

DonLakeFlyer:
I say we leave it as you have it. It works everywhere I tested.

DonLakeFlyer:
There is so much crazy stuff going on inside MissionManager/Controller as well as UI interaction we really need a comprehensive functional test as well as ui test for all of this. Otherwise, it's impossible to manually test for all the boundary cases.

NaterGator:
> This is probably why it works for you and I couldn't get it to work.

But I haven't done that yet! :worried:

DonLakeFlyer:
The difference in your code is that I think the way you do it creates a different binding/signalling path that works when you change the coordinate out from under it.

NaterGator:
I know I know, you said not to do it but I'm really uncomfortable with the QDeclarativePolylineMapItem::geometry_ and QDeclarativePolylineMapItem::updateMapItem calls getting skipped. I ran this in the debugger and things are working properly. Here's the callstack:

image


Edit: I decided to rerun the old version and you were right, the bindings are managed by a QQmlBinding. Here's the stack trace... I'll revert the last commit
image

NaterGator:
@DonLakeFlyer this is ready to merge, sorry for all the noise.

DonLakeFlyer:
No problem. This is a great change. I just had so many problems with this area of the code being paranoid is a good thing.

Issue #3576, opened 2016-06-16
Connect Solo to iOS QGC causes perf issues


DonLakeFlyer:
QGC will go unresponsive after some amount of time. Doesn't seem to happen with PX4 stack and my PixRacer.

DonLakeFlyer:
ArduPilot SITL seems fine

DonLakeFlyer:
Still a problem even after the yield animation changes.

DonLakeFlyer:
This is because the Solo sends parameter updates for thing like gimbal info and so forth at a very high rate. Very strange way to do things. Then this all gets processed on the main thread and chokes things out.

DonLakeFlyer:
The perf problem is caused by the local param cache being update on every param update from the solo.  For now I'm going to disable param cache for the Solo.

DonLakeFlyer:
The waiting indices are also not quite used correctly. Too much progress bar updating back to 0 for no reason.

DonLakeFlyer:
Need to test fly again to validate fix.

DonLakeFlyer:
Works fine now

PR #3585, opened 2016-06-17
Remove animation to save perf

DonLakeFlyer:
This is the Yield button for messages on the toolbar. When the animation for that starts it chews a ridiculous amount of CPU on iOS. I get it more often on my Solo since it's always complaining about the GPS pre-arm state. Which cause the animation to go off.

Related to #3576

### DIFF
--- a/src/ui/toolbar/MainToolBarIndicators.qml
+++ b/src/ui/toolbar/MainToolBarIndicators.qml
@@ -102,19 +102,6 @@ Row {
                 anchors.verticalCenter:   parent.verticalCenter
                 anchors.horizontalCenter: parent.horizontalCenter
             }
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
It causes the refresh rate to jump to 60 FPS which then causes thing to freak out.

NaterGator:
The image was not set to be cached. I wonder if Qt Quick was reloading it from storage on every opacity change?

DonLakeFlyer:
I'll see if it makes a difference. But I think the real problem is the 60 fps thing.

DonLakeFlyer:
No difference. The animation bumps up the frame rate no matter what.

Issue #2638, opened 2016-01-16
Tame the CPU hog


LorenzMeier:
@DonLakeFlyer I know this is not fun and not so much a consideration during desktop development, but its the No. 1 complaint in the field: QGC eating your battery alive. This is what it currently looks like for me.

It ramps up my fan quite a bit and I'm relatively sure that my field battery life would not be great.

<img width="672" alt="screen shot 2016-01-16 at 12 18 29" src="https://cloud.githubusercontent.com/assets/1208119/12372092/6de4341c-bc4b-11e5-93c8-36b3d62d6f2b.png">

LorenzMeier:
Note that I included WindowServer because its load goes from single-digit percent to 35 once QGC starts.

DonLakeFlyer:
Unless there is something trival here I don't think this will fit into 3.0 release.

LorenzMeier:
It might make sense to run a function profiler to see if the load is somewhere it can be easily fixed, or if its across the board. I had QGC around 20-25% back in the day when I regularly profiled it.

LorenzMeier:
Note that native QML main Window might also help because it uses different rendering - didn't you work out recently how to do this?

DonLakeFlyer:
Hmm, maybe. I'll still have a branch for that. I'll try it to see if it makes a difference. Another interesting thing is that in the commercial version of Qt there is a Qml compiler.

DonLakeFlyer:
What sort of connections were you running when you did this test? We can't get anything higher than 20% CPU to show up.

NaterGator:
@DonLakeFlyer The Qt Quick compiler should only help with initial load times; once the JIT has compiled at runtime things should theoretically be the same. The Qt Quick [performance guide](http://doc.qt.io/qt-5/qtquick-performance.html#bindings) is worth looking over, but you've probably already seen it. 

@LorenzMeier any chance you were streaming video via QGC/Gstreamer?

DonLakeFlyer:
I'm planning to move this to 3.1 release. Speak up now if you disagree.

DonLakeFlyer:
Current builds seem to be ok

PR #3595, opened 2016-06-18
Defer loading and binding of mission editing QML components

NaterGator:
Also should finally fix mission editor view sync when changing current waypoint.

During creation of a very large mission with 100s of points I noticed the mission editor was getting really laggy. QML profiler showed the editor bindings for the simple mission item delegate were extremely expensive and slow.

Instead, load them only for each point when it is selected. This makes the ListView scroll smoothly with a ton of points too.

### DIFF
--- a/src/MissionEditor/MissionEditor.qml
+++ b/src/MissionEditor/MissionEditor.qml
@@ -46,8 +46,9 @@ QGCView {
     readonly property int       _addMissionItemsButtonAutoOffTimeout:   10000
     readonly property var       _defaultVehicleCoordinate:   QtPositioning.coordinate(37.803784, -122.462276)
 
-    property var    _visualItems:          controller.visualItems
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
I should have some time to take a look a this later today.

DonLakeFlyer:
Thanks Nate

NaterGator:
No problem. This ties in tightly with #2638. **Especially** for anything that is a delegate of a model view we should adopt this (or a similar) approach to reduce the number and complexity of bindings. I don't think I fully appreciated how expensive some of these bindings are. Glad we're hacking away at the CPU hog though.

DonLakeFlyer:
We'll have to look at how expensive the smaller lists are. This is one of the few places where you can end up with a large list. That said in 3.1 this big list will likely all go away and I'll move to something like click in the map to show a single item to edit model.

PR #3596, opened 2016-06-18
Introducing Android Joystick

gregd72002:
This introduces basic Android Joystick support. Only one instance of a joystick is supported.
It requires Qt 5.6.


DonLakeFlyer:
Two things:
1) I thought there was a part of this that didn't compile without 5.6? I don't see that now.
2) You have a submodue which isn't up to date in this pull. You need to for a git submodule update. And to fix this pull.

gregd72002:
The submodule version has been fixed.

Re QT version, I've checked again and it will not compile on 5.5.1 on my Mac but runs fine on 5.6

DonLakeFlyer:
All builds are passing on travis. And I just tried building on my OSX box running 5.5 without any problems?

gregd72002:
This is strange... Just to confirm, I'm compiling on OSX with QtCreator (3.6.1): 
Android for armeabi-v7a (GCC 4.9, Qt 5.5.1), Release

And the error I'm getting:

> /Users/kiwi/AndroidStudioProjects/qgroundcontrol/src/Joystick/JoystickAndroid.h:9: error:  QtCore/private/qjni_p.h: No such file or directory


It works fine when compiling for: Android for armeabi-v7a (GCC 4.9, Qt 5.6.0), Release

DonLakeFlyer:
@dagar Something out of whack with travis here. The pull shows success but if you look a the log for the android build, you'll see that it failed for the exact reason greg specified above. Any ideas?

Also I built the wrong thing on my OS X box. So that was user error on my part.

dagar:
Compile failure reported as a successful build. Thanks travis...

dagar:
@DonLakeFlyer @gregd72002 build is now fixed/properly broken

dagar:
@gregd72002 I'm just curious, which joystick or gamepad are you using on android? I'm really interested in getting this stuff to the point I can go flying without any traditional RC gear.

DonLakeFlyer:
I really wanted to get this merged in without having to wait for our change to Qt 5.7. Mainly to prevent breakage caused by master moving forward. I gave it a try to hack it up a bit to at least get it to compile but it's too involved and not much better than just leaving it here to wait until we upgrade Qt version. So this will have to wait.

gregd72002:
@dagar, I'm just using standard PS3 controller. I had to introduce one more code change to make it usable - an accumulator on throttle stick. (PS3 has a spring on throttle stick so it is difficult to fly without the accumulator). Any movement above the mid position adds to throttle, while movement below the mid position - subtracts from the throttle. 
You are not going to throw amazing acrobatics with it as it is too slow but it is good enough for gentle flying.

dagar:
I've used a logitech gamepad through the desktop app, and it was okay if I stuck to higher level stabilized modes. My solution for the throttle was only using PX4 modes where a centered spring loaded throttle might actually be superior (altitude or position control). Is the accumulator optional?

gregd72002:
The accumulator should be optional, but for my needs I've just hardcoded it.

gregd72002:
Well, note that the accumulator is not part of this pull request. It is something that I've added only to my local version.
This pull request will add joystick support to android so it behaves the same way as on PC

dagar:
I was working on allowing the throttle to move to the top two trigger buttons which would be throttle up/down. This frees up the left stick for camera control. For both this type of throttle control and your accumulator I think we'd want a throttle indicator in QGC. Someday I'll finish it...

gregd72002:
@dagar i've created the pull request for the accumulator on throttle (https://github.com/mavlink/qgroundcontrol/pull/4123). Once accepted it should be reasonably easy to add 2 additional actions on buttons - throttle up & down. Question would be - how much should these affect the throttle - i..e +/-10 or more of +-1000? I guess it should be configurable. 
At the same time I am thinking of adding trim options on button press but this is a different story.

dagar:
I'd start with 10% unless you also want to make it configurable right away.

PR #3614, opened 2016-06-22
travis-ci try ccache on OSX and android

dagar:


dagar:
@DonLakeFlyer do you have any experience with ccache on OSX? If we can override QMAKE_CXX I think it would really help build times.

DonLakeFlyer:
Nope, don't know anything about it

NaterGator:
@dagar just add:
`QMAKE_CXX = ccache $$QMAKE_CXX` 

[here](https://github.com/mavlink/qgroundcontrol/blob/53a49fdcd28fbe74a6fdce5d47779bb1541891f7/QGCCommon.pri#L53)

dagar:
Replaced by #3631

Issue #3611, opened 2016-06-21
Look into battery monitor calculator from MP


DonLakeFlyer:

DonLakeFlyer:
ArduPilot side completed. Need to look into same thing for PX4.

DonLakeFlyer:
@LorenzMeier As far as I can tell from looking at params it doesn't quite seem like the calculators are possible. I'm assuming the two params which need tweaking are BAT_A_PER_V and BAT_V_DIV. Let's use ArduPilot as example. It has a voltage divider parameter which always has a value in it. If the vehicle measured voltage is off you can update it using the formula:

(externally measured voltage * voltage divider) / vehicle reported voltage

But with BAT_V_DIV which I assume is the corresponding PX voltage divider param it defaults to -1. So there is no starting value to work from that you can adjust.

dagar:
In the code if the param is -1 the internal value is set based on hardware define.
https://github.com/PX4/Firmware/blob/571798c318a6736f748495378907cf76fae3c884/src/modules/sensors/sensors.cpp#L925

What if we change it to actually set the param?

DonLakeFlyer:
Yes, then the little calculator thingies would work. On the ArduPilot side Power settings has a set of built in settings which you can pick from a Combo. Then the Other selection shows you the dividers so you can tweak. Doesn't that sort of thing make sense with PX4 as well? You can run QGC + ArduCopter MockLink to see what it look like.

DonLakeFlyer:
Link to APM Power options: https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/APM/APMPowerComponent.qml#L54

dagar:
See https://github.com/PX4/Firmware/pull/4912

PR #3629, opened 2016-06-23
Add voltage/current calculators

DonLakeFlyer:
Fix for #3611

screen shot 2016-06-23 at 4 14 17 pm
screen shot 2016-06-23 at 4 14 30 pm


DonLakeFlyer:
@LorenzMeier Is the same thing doable on the PX4 side?

LorenzMeier:
Yes, it would be good if you port that right away. If you flash PX4 master and look at the params you'll see that you'll find the equivalent values there. https://github.com/mavlink/qgroundcontrol/issues/3637

PR #3631, opened 2016-06-24
Ccache test for #3614

NaterGator:
Just seeing if this does what we want in #3614


.travis.yml
```diff
@@ -175,17 +189,20 @@ before_script:
 script:
   - if [ "${TRAVIS_OS_NAME}" = "android" ]; then
         cd ${TRAVIS_BUILD_DIR} &&
-        ./tools/update_android_version.sh
+        ./tools/update_android_version.sh &&
+        export JOBS=4
```
dagar: drop this?

NaterGator: Drop the jobs export or version update? gcc was crashing due to running out of memory when letting the system manage the number of jobs based on number of available cores.

Issue #3608, opened 2016-06-21
Add compassmot support


DonLakeFlyer:

DonLakeFlyer:
http://ardupilot.org/copter/docs/common-compass-setup-advanced.html

davidbitton:
Is there a terminal mode in QGC where I could just call the command?

DonLakeFlyer:
@davidbitton I should have compassmot support done by tomorrow. When I put that in, would you be able to test it?

<img width="1011" alt="screen shot 2016-06-23 at 8 05 11 pm" src="https://cloud.githubusercontent.com/assets/5876851/16326714/e6d2d040-397d-11e6-8a1e-e9f17816c16c.png">
<img width="1014" alt="screen shot 2016-06-23 at 8 05 20 pm" src="https://cloud.githubusercontent.com/assets/5876851/16326715/e6f1c9dc-397d-11e6-83dc-f694d4bde3c0.png">
<img width="1014" alt="screen shot 2016-06-23 at 8 05 28 pm" src="https://cloud.githubusercontent.com/assets/5876851/16326716/e6f1d22e-397d-11e6-8108-5880fcda9192.png">

davidbitton:
Sure.

On Thu, Jun 23, 2016 at 11:06 PM Don Gagne <notifications@github.com> wrote:

> @davidbitton <https://github.com/davidbitton> I should have compassmot
> support done by tomorrow. When I put that in, would you be able to test it?
>
> [image: screen shot 2016-06-23 at 8 05 11 pm]
> <https://cloud.githubusercontent.com/assets/5876851/16326714/e6d2d040-397d-11e6-8a1e-e9f17816c16c.png>
> [image: screen shot 2016-06-23 at 8 05 20 pm]
> <https://cloud.githubusercontent.com/assets/5876851/16326715/e6f1c9dc-397d-11e6-83dc-f694d4bde3c0.png>
> [image: screen shot 2016-06-23 at 8 05 28 pm]
> <https://cloud.githubusercontent.com/assets/5876851/16326716/e6f1d22e-397d-11e6-8108-5880fcda9192.png>
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/mavlink/qgroundcontrol/issues/3608#issuecomment-228247343>,
> or mute the thread
[...]

PR #3642, opened 2016-06-24
Add CompassMot support

DonLakeFlyer:
Fix for #3608


DonLakeFlyer:
Need to recreate to deal with submodule stuff

Issue #3640, opened 2016-06-24
RC Channel Monitor widget range too narrow


DonLakeFlyer:
Currently set to 1000<->2000 but radios can easily go outside that range.

DonLakeFlyer:
MP uses 800<->2200 range so I'll switch to that.

PR #3644, opened 2016-06-24
Radio Cal: PWM range is now 800<->2200

DonLakeFlyer:
Fix for #3640


kd0aij:
that looks good to me

PR #3649, opened 2016-06-25
travis-ci add more caching

dagar:


dagar:
Yet another reason the builds are slow is that travis-ci will at most give us 5 concurrent jobs even if the resources are available. I've moved the doxygen build and linux debug build + unit tests to semaphoreci. 
The Linux, OSX, iOS, Android built and uploaded from travis-ci.

Issue #3641, opened 2016-06-24
Fix unit test framework static delays


NaterGator:
The static delays to allow event loops and self-managing objects to init / cleanup is expensive on CI. We should systematically work through and close the loop on all of them so we're not waiting needlessly.

DonLakeFlyer:
The one you pointed out is not related to unit testing. It's a general shutdown problem.

NaterGator:
I've been looking at this and a huge portion of (in fact, almost all) the overhead comes from synchronizing the start of MockLink and waiting for the parameter load to finish (based on QTimers triggering). I think I'll start there.

DonLakeFlyer:
The parameter load under unit test is specifically timed out by MockLink to simulate a real connection. This allows other signals to get out in between parameters coming down from the vehicle. It’s that way on purpose to replicate the normal order of signals when connected to a real vehicle.

DonLakeFlyer:
I think it could be sped up, but the actual parameter load unit test itself would need to remain as is with the slower timing. Since it's the only way to test the various failure conditions.

NaterGator:
Right, I understand sequencing of signals is critical. One of the things I'm also working on is ensuring deterministic connection of signal/slots before threaded objects are allowed to run their event loop and start generating signals.

As a small example though, every single test has to wait on [this](https://github.com/mavlink/qgroundcontrol/blob/11bcffb0970c34aaff3dc358c975387d29c4f0bc/src/FactSystem/ParameterLoader.cc#L66) timer to expire before the unit test can proceed. The MockLink blasts out its full set of components almost instantly, so this is just wasted 1s per every single individual test run.

DonLakeFlyer:
That shouldn't be the case. If all the parameters from the default component are found, the timer is stopped here: https://github.com/mavlink/qgroundcontrol/blob/master/src/FactSystem/ParameterLoader.cc#L226. And the completed signals should go out immdediately in the call to _checkInitialLoadComplete at the bottom.

NaterGator:
[...]

PR #3653, opened 2016-06-25
Unit test performance evaluation and improvement

NaterGator:


NaterGator:
See #3641

src/FirmwarePlugin/FirmwarePlugin.h
```diff
@@ -147,7 +147,7 @@ class FirmwarePlugin : public QObject
     virtual bool sendHomePositionToVehicle(void);
     
     /// Returns the parameter that is used to identify the default component
-    virtual QString getDefaultComponentIdParam(void) const { return QString(); }
+    virtual QString getDefaultComponentIdParam(void) const { return QString("SYS_AUTOSTART"); }
```
DonLakeFlyer: This is wrong. The generic case knows nothing about the parameter set it is dealing with.

DonLakeFlyer:
I'd rather see the threading changes and deleteLater changes in a separate pull. I'm not so sure I want that for 3.0. That whole space has been so flaky I don't want to introduce any new problems. Current code may not be quite right, but I'm pretty sure it works ok (we may leak a link, but I can live with that). Then in 3.1 we can revisit the whole threading model, deleteLater shutdown stuff.

Does that make sense?

NaterGator:
Sure, I'm not in a position to fix this PR at the moment but I should be able to do it tonight.

DonLakeFlyer:
Decided to shut things down for 3.0. So this will go to 3.1

DonLakeFlyer:
@NaterGator Can you rebase this and I'll take a look at it again

dagar:
Closing as stale.

Issue #3657, opened 2016-06-26
GPS signal strength indicator not meaningful


DonLakeFlyer:
Trying to show hdop with the signal bars just doesn't work. I've been flying with it and it just doens't give useful information. What I really want to see is just sat count and hdop.

Issue #3658, opened 2016-06-26
Toolbar Battery percent text coloring doesn't work outdoors


DonLakeFlyer:
It's green when good percentage and  unreadable outdoors.

PR #3660, opened 2016-06-26
Toolbar: Show sat count and hdop instead of signal strength

DonLakeFlyer:
Fix for #3657

screen shot 2016-06-25 at 6 03 27 pm

### DIFF
--- a/src/ui/toolbar/MainToolBar.qml
+++ b/src/ui/toolbar/MainToolBar.qml
@@ -67,7 +67,7 @@ Rectangle {
     function getBatteryColor() {
         if(activeVehicle) {
             if(activeVehicle.battery.percentRemaining.value > 75) {
-                return colorGreen
+                return qgcPal.text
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@dogmaphobic I've been a lot of test flying lately to verify 3.0 release. I noticed that the signal strength indicator for gps just isn't useful. I tended to instead have to gps tool dropped down pre-flight so I could watch sat count and hdop go up until I got good enough values to fly. A while back I tried to get a discussion going on translating hdop to signal strength bars, which led to the number changes a while back that are in their now for the bar values. But in reality I could never get a concensus from anyone on hdop good/bad. Seems to be a personal thing, also depends somewhat on where/how you are flying. So that also points to trying to convert the a signal strength graphic not being good.

DonLakeFlyer:
This also has a fix for: #3658

PR #3699, opened 2016-06-30
DO NOT MERGE: Fix coordinate scaling bug regarding MAVLink specs

LorenzMeier:
@DonLakeFlyer If merge now this will lead to fly-aways to Africa, so this can only go in once we have a minimum Firmware version check in and enforced.

### DIFF
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -310,8 +310,8 @@ void PX4FirmwarePlugin::guidedModeGotoLocation(Vehicle* vehicle, const QGeoCoord
     cmd.param2 = MAV_DO_REPOSITION_FLAGS_CHANGE_MODE;
     cmd.param3 = 0.0f;
     cmd.param4 = NAN;
-    cmd.param5 = gotoCoord.latitude() * 1e7;
-    cmd.param6 = gotoCoord.longitude() * 1e7;
+    cmd.param5 = gotoCoord.latitude();
+    cmd.param6 = gotoCoord.longitude();
     cmd.param7 = vehicle->altitudeAMSL()->rawValue().toDouble();
     cmd.target_system = vehicle->id();
     cmd.target_component = vehicle->defaultComponentId();


DonLakeFlyer:
@LorenzMeier You need to tell me what the min version is

LorenzMeier:
Thanks for the reminder. I get the wrong version displayed right now in QGC and need to work out why the decoding in QGC is different from mine. And I probably also need to put that particular version on stable then.

LorenzMeier:
@DonLakeFlyer Firmware version is not showing in SITL, but is implemented in PX4 SITL. Any reason QGC is not pulling the version in that case?

screen shot 2016-07-01 at 09 53 58

DonLakeFlyer:
Check that it’s getting the VERSION mavlink message in mavlink inspector. If so, something wrong with QGC. If not, something wrong with SITL.

LorenzMeier:
I never receive MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES on the autopilot side. This is because PX4 does not allow commands to be handled until it has finished the network discovery, which requires a heartbeat from the GCS. But QGC is sending the command right after it got the first heartbeat.

The proper approach would be to retry on the QGC side. The slightly less involved version of this would be to send a GCS heartbeat immediately on creating a new vehicle instance, then backing off 50 ms and only then starting to send commands.

LorenzMeier:
@DonLakeFlyer Minimum version required should be 1.4.1rc1 (its ok if you just check for 1.4.1, any build using that number has the fix)
https://github.com/PX4/Firmware/releases/tag/v1.4.1rc1

It probably makes sense to put this onto this branch and merge it, just implies that anyone using the QGC 3.0.0 RC needs beta or master of PX4, which seems fair.

DonLakeFlyer:
@LorenzMeier I can't do the version number check without the firmware version number mavlink change.

LorenzMeier:
@DonLakeFlyer Can you please re-test with master? It all works for me here.

DonLakeFlyer:
Yup, my screwup.

DonLakeFlyer:
@LorenzMeier How come when I run SITL master against QGC **without** this change, goto goes to the right spot. Shouldn't it go to the wrong place since QGC isn't sending the right format?

LorenzMeier:
Px4 supports old and new spec, since I could tell them apart.

DonLakeFlyer:
Ok. Got it. I'm going to take control of this pull and make a new one to merge into both Stable 3.0 and master.

DonLakeFlyer:
@LorenzMeier Wait: If firmware can detect old/new why the version check then?

LorenzMeier:
It is safe to use new firmware with old and new QGC. It is *NOT* safe to use old firmware which does not contain the check. So when QGC ships with this fix, it needs to make sure it talks to *NEW* Firmware. Makes sense?

DonLakeFlyer:
Duh!

PR #3702, opened 2016-06-30
Turning off cache since it's breaking release builds (maybe!)

DonLakeFlyer:
Or at least testing to see if that is the problem

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ language: cpp
 
 env:
   global:
-    - CCACHE_CPP2=1
     - JOBS=4
     - QT_FATAL_WARNINGS=1
     - SHADOW_BUILD_DIR=/tmp/shadow_build_dir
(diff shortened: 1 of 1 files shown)


NaterGator:
https://www.traviscistatus.com

We can also try nuking the cache from the project's Travis config page.

DonLakeFlyer:
I'll try doing that if this passes

DonLakeFlyer:
Clearing cache seemed to work.

dagar:
You can temporarily disable ccache with `export CCACHE_DISABLE=1`

DonLakeFlyer:
Good to know. I wonder if the cache will just blow up every once in a while and need clearing. Guess we'll see. Now I know how to clear it though.

dagar:
We definitely should be able to trust it all the times. It's limited to 500MB, but travis-ci keeps different caches for branches, etc

I wasn't following things today, but it looks like travis-ci in general has been in pretty bad shape.

Issue #3033, opened 2016-03-14
Wind direction issue


jcooper21:
I don't think the wind direction in QGC is displaying correct data. It does not seem to be actual wind direction.

jcooper21:
@DonLakeFlyer 

PX4 never displays wind direction and speed but APM does. Is this a bug?

DonLakeFlyer:
As far as I know PX4 stack doesn't support wind direction yet. Is APM showing incorrect value?

jcooper21:
I have not tested APM recently so I'm not sure if it's working on the new version. Does px4 support wind speed because that does not show up in QGC either

DonLakeFlyer:
No wind information aat all in PX4 stack,

dagar:
We'd need to implement WIND_COV  (http://mavlink.org/messages/common#WIND_COV) on the PX4 side. Internally the PX4 estimators publish a wind estimate (https://github.com/PX4/Firmware/blob/master/msg/wind_estimate.msg).

dagar:
I'll implement the PX4 side - https://github.com/PX4/Firmware/issues/4678

DonLakeFlyer:
QGC usess MAVLINK_MSG_ID_WIND

dagar:
I saw that, but common only has WIND_COV.

http://mavlink.org/messages/ardupilotmega#WIND

DonLakeFlyer:
Yeah, just realized that as well. Looks like WIND_COV is a superset of ardupilot WIND message. I can convert ardupilot WIND message to WIND_COV in the plugin and keep central QGC to WIND_COV. WIND_COV will need spec changed to allow nan in missing fields though since WIND is subset.

jcooper21:
@dagar @DonLakeFlyer  was this supposed to be in today's beta update of PX4?  I just tested the latest beta and nothing published to QGC in Android.

dagar:
You should be able to see WIND_COV in the mavlink inspector, but the rest of the QGC support hasn't been done yet.

jcooper21:
OK thanks. I was looking in the instrument panel for a reading.

booo:
As far as I can see this is implemented.

PR #3721, opened 2016-07-03
mavlink WIND_COV -> windFactGroup

dagar:
Needs testing. No wind in SITL.


dagar:
Sigh, no wind in HIL either. Can someone just check my math?

dagar:
Fixes #3033

NaterGator:
Math checks out

DonLakeFlyer:
Daniel could you dump an issue on the firmware side for fake wind. At least to send this message, not really simulated? Would be handy for testing.

dagar:
It looks like it was just the switch to EKF2 where it's not enabled quite yet.
https://github.com/PX4/ecl/blob/eaae95fdc4e48e3946a463941dfb9d5e495e975f/EKF/control.cpp#L744

PR #3723, opened 2016-07-03
PX4 add bitmask for param metadata

dagar:
Look at src/FirmwarePlugin/PX4/PX4ParameterMetaData.cc for the main change. I included updated param and airframe metadata. 

https://github.com/mavlink/qgroundcontrol/pull/3723/files#diff-96a673062773c2b8d313746933b7ac71R315

### DIFF
--- a/src/AutoPilotPlugins/PX4/AirframeFactMetaData.xml
+++ b/src/AutoPilotPlugins/PX4/AirframeFactMetaData.xml
@@ -14,6 +14,21 @@
     </airframe>
   </airframe_group>
   <airframe_group image="FlyingWing" name="Flying Wing">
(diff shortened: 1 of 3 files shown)


dagar:
screenshot from 2016-07-03 19-46-54

DonLakeFlyer:
Wondering about that param meta data. Should the doc generator just use the bitmask meta data to generate the description for web? The massive long description is kind of annoying.

dagar:
Once this reaches most people on both sides (PX4 and QGC) we can strip the detailed description, like what's happened with many of the enums. The downside is that the doc generator doesn't currently support any of this.

PR #3730, opened 2016-07-05
PX4 MAV_CMD_NAV_LOITER_TO_ALT and cleanup

dagar:


src/FirmwarePlugin/PX4/MavCmdInfoFixedWing.json
```diff
@@ -59,9 +59,16 @@
             "param3": {
                 "label": "Radius:",
                 "units": "m",
-                "default": 100,
+                "default": 50,
                 "decimalPlaces": 0
             },
+            "param4": {
+                "label":            "Next waypoint start:",
```
DonLakeFlyer: A bit concerned about such a long label fitting on all form factors. I'm going to have to merge and try on a phone to see.

dagar:
This is about my third try at naming this field. 

    Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location

DonLakeFlyer:
Oh, it's that one! I messed around with it for a while and decided to give up until 3.1 :)

DonLakeFlyer:
Do you know how to get PX4 SITL to be fixed wing?

dagar:
You have to use master.

`make posix_sitl_default gazebo_plane`

dagar:
I responded without thinking. This hasn't been merged in PX4 yet.
https://github.com/PX4/Firmware/pull/4992

DonLakeFlyer:
No problem. I just want to test the ui layout.

DonLakeFlyer:
Isn't this for any fixed wing? Seems like it should be on both PX4 and APM. Also isn't there more than one message affected: https://github.com/mavlink/mavlink/pull/531#event-628415121

dagar:
Yes, I suppose I should do that. Not sure about the heading hold.

Could we structure the metadata like this? I think this makes sense with PX4 only using the common mavlink messages.

- Common
 - Fixedwing (PX4)
    - APM_FixedWing
 - Multicopter (PX4)
    - APM_Multicopter

DonLakeFlyer:
The structure should be:
* Common
 * Fixed Wing Common (missing in 3.0)
   * PX4
    * APM
 * Multi-Rotor Common (missing in 3.0)
   * PX4
    * APM

It's not about where the command comes from. That doesn't really matter. It's about whether the command works differently and hence need different ui given vehicle type and/or firmware.

dagar:
Ok, I figured PX4 and common would be close enough that they could just be the same. Maybe not.

DonLakeFlyer:
Common is mavlink generic so it has to have all the fields defined that are available in the mavlink spec. But both PX4 and ArduPilot have many cases where they have not implemented the full set of fields. Because of that you can't use PX4 as Common since it is not to mavlink spec in various places.

DonLakeFlyer:
Moving this to 3.1

dagar:
I've synced up PX4 and APM fixed wing.
Is "Next waypoint start" an acceptable name?

dagar:
Merge now or wait for the better hierarchy?

DonLakeFlyer:
Good for now

Issue #3698, opened 2016-06-30
Memory leak during flight


DonLakeFlyer:
- Setup a looping mission
- Crank it up in SITL against an iOS build
- Wait

Watch the memory usage, go up and up. Doesn't happen really fast, but still a bad leak.

DonLakeFlyer:
The memory increase is from the trajectory lines. Without trajectory lines memory is very stable.

NaterGator:
I was looking at this today. Is there a particularly good reason to store these as coordinate vectors? Simply storing a QVariantList of QGeoCoordinates can be bound directly to a MapPolyLine without any of the model/view/qobject overhead. 

Also, it may not be a bad idea to have a setting for the maximum "age" of the trajectory line. 5 minutes sampled at 1 second could be reserved up front and would only require 5*60*sizeof(QGeoCoordinate) bytes.

DonLakeFlyer:
Yeah, I don't know. This trajectory line stuff is pretty old. So it may be out of what with current understanding of what you can do with MapPolyLine. I was planning at looking at Mission Planner to get a sense for how it handles maximum age of lines.

DonLakeFlyer:
> Simply storing a QVariantList of QGeoCoordinates can be bound directly to a MapPolyLine without any of the model/view/qobject overhead.

That means the full list needs to be signalled for change each time a point is added though. Whereas using a ListModel you get incremental update. Correct?

DonLakeFlyer:
Also MapPolyLine doesn't support a model, it only supports a list. So you can't have a single ListModel of QGeoCoordiantes that updates passed to MapPolyLine.

DonLakeFlyer:
Which I think means there isn't really a better way to do it?

DonLakeFlyer:
@NaterGator Please let me know if you still think there is a different way to do this. Had to look and write things down to re-create in my head why I did it this way.

DonLakeFlyer:
MP has a track length settings that defaults to 200. I assume that means 200 segments.

NaterGator:
[...]

PR #3739, opened 2016-07-07
Trajectory updates

NaterGator:
For #3698, @DonLakeFlyer I'm not getting leaks on Linux but would be grateful if you could test too.

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewMap.qml
+++ b/src/FlightDisplay/FlightDisplayViewMap.qml
@@ -59,18 +59,12 @@ FlightMap {
     }
 
     // Add trajectory points to the map
-    MapItemView {
-        model: _mainIsMap ? _activeVehicle ? _activeVehicle.trajectoryPoints : 0 : 0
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Hmm, I'll try it this afternoon. That's exactly what I did and I couldn't get the lines to draw right.

NaterGator:
The only quirk I've seen is there is a segment to 0,0 if the path list only contains 1 element. Easy fix if that's the only issue.

DonLakeFlyer:
This has the same line drawing problem as when I tried it. Seems like it may be an OSX only bug. iOS seems to be fine. And you did Linux. I tried the 1 element change on my fix and it didn't work. I'm going to try fixed allocation of CoordinateVectors and still use QmlObjectListModel with multiple polylines like the old code. Hopefully that will fix leak and work around line drawing problems.

DonLakeFlyer:
Closing since it doesn't work and there isn't actually a memory leak

PR #3750, opened 2016-07-08
Travis: Pulling out disabled ccache setup

DonLakeFlyer:
Hoping to save a little time from skipping unused install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ language: cpp
 
 env:
   global:
-    - CCACHE_CPP2=1
-    - CCACHE_DISABLE=1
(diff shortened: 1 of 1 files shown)


dagar:
I know it's hard to trust it after a non obvious failure, but this is almost certainly fixable and just some little problem with xcode. The install itself doesn't take anytime, but you could blow away the existing travis-ci cache to prevent it from downloading.

DonLakeFlyer:
Looking at install times from previous pulls and this one it doesn't seem like any sort of build time win to remove. If you think you can get it working at some point I'll leave in.

Issue #3364, opened 2016-05-12
Log download crashes with many entries


LorenzMeier:
If you have 120+ logs QGC will crash on trying to display them:
http://discuss.px4.io/uploads/default/original/1X/875efb3202959946ef26b7e1ab28795e16207e5e.pdf

andrejpan:
It is also happening when there is less logs. I had around 90 logs and I used qgc 2.9.7 at ubuntu 14.04. Here is the print screen: ![logs3](https://cloud.githubusercontent.com/assets/1015265/15363361/cce8cd6e-1d18-11e6-9a5d-12ccdfe3c23c.png)

DonLakeFlyer:
This is the Log Download widget crashing?

andrejpan:
Yeas, but it is also killing QGC 297b.

dogmaphobic:
I was able to reproduce this and I *think* I know where the crash is. The (bulk) change from model to view is running into some race condition in the QtQuickWidget boundary. Given that it is quite unlikely Qt will ever fix these QtQuickWidgets issues, I have to come up with an alternate way of handling this.

DonLakeFlyer:
@dogmaphobic Any update on this one?

dogmaphobic:
Nope. Until we switch away from QtQuickWidget, I suggest not accumulating an inordinate amount of log files. In the mean time, it is best to take the SD card out and load it directly onto the computer. Log transfer through MavLink sucks any way due to the tiny packet size.

DonLakeFlyer:
@dogmaphobic Is there a reason this doesn't just use QmlObjectListModel?

DonLakeFlyer:
There's more than just a refresh crash here. I just crashed booting QGC with the log download widget active. It crashed trying to clear the list on the set active vehicle.

DonLakeFlyer:
Of course it was a one time thing.

DonLakeFlyer:
It's not crashing in the list clear. It's crashing in the disconnect from the previous vehicle. Something is out of whack since there is no previous vehicle.

DonLakeFlyer:
@dogmaphobic The repro is connect/disconnect/connect with widget up. It's crashing because setActiveVehicle is not handling setActiveVehicle(NULL) corrrectly when a vehicle goes away. It still keeps old pointers from a deleted vehicle. The second connect then tries to disconnect signals on a bad pointer.

NaterGator:
[...]

PR #3753, opened 2016-07-08
Use stored QMetaObject::Connections to disconnect deactivated vehicles

NaterGator:
@DonLakeFlyer this is what I was talking about [here](https://github.com/mavlink/qgroundcontrol/issues/3364#issuecomment-231251274)

### DIFF
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -938,16 +938,10 @@ void MissionController::_activeVehicleChanged(Vehicle* activeVehicle)
 {
     qCDebug(MissionControllerLog) << "_activeVehicleChanged activeVehicle" << activeVehicle;
 
-    if (_activeVehicle) {
-        MissionManager* missionManager = _activeVehicle->missionManager();
-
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
Ah, I get it now. Much better pattern. Didn't know  you could do that.

src/MissionManager/MissionController.cc
```diff
@@ -938,16 +938,10 @@ void MissionController::_activeVehicleChanged(Vehicle* activeVehicle)
 {
     qCDebug(MissionControllerLog) << "_activeVehicleChanged activeVehicle" << activeVehicle;
 
-    if (_activeVehicle) {
-        MissionManager* missionManager = _activeVehicle->missionManager();
-
-        disconnect(missionManager, &MissionManager::newMissionItemsAvailable,   this, &MissionController::_newMissionItemsAvailableFromVehicle);
```
DonLakeFlyer: Doesn't this work as well? Seems more understandable:
```
missionManager->disconnect(this);
_activeVehicle->disconnect(this);
```

DonLakeFlyer:
Closing this out. I think the keeping a list mechanism hides things a bit too much. I think the foo->disconnect(this) is less error prone than current stuff. But still shows you a bit about what you are doing.

PR #3759, opened 2016-07-09
Build misc

dagar:
- all tagged builds to s3 latest/
 - appveyor tagged builds to vX.Y.Z
 - name qgroundcontrol -> QGroundControl (fixes android app name)
 - travis-ci ccache OSX error

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -66,6 +66,7 @@ artifacts:
     name: symbols
 
 deploy:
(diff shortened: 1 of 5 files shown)


dagar:
This changes the binary name to QGroundControl so that the Android app was named properly. Not sure if there are other build system/installer ramifications I haven't found yet.

DonLakeFlyer:
Merge?

PR #3762, opened 2016-07-09
Update development version from git

dagar:
branch:sha (Development) date

Too much?

screenshot from 2016-07-09 19-32-43


DonLakeFlyer:
I figure on daily builds there isn't really anything such as too much.

PR #3771, opened 2016-07-10
Warn user on requiring 1.4.1 or above version

DonLakeFlyer:


DonLakeFlyer:
Hmm, travis seems to be confused.

Issue #3778, opened 2016-07-11
Some offline map size is 0


glody:
Test Environment
QGroundControl: Daily Build: v2.9.7b-411-g0fefa4f
Windows: Windows 8.1 64 Bit Laptop
Purpose:
Check if the offline map size is correct
Step:
1. Lanuch QGC
2. Click "Add new set" in Offline Maps item Settings page
3. Download the set
Excepted result:
The map size should not 0
Actual result:
Some offline map size is 0
![map size is 0](https://cloud.githubusercontent.com/assets/1768868/16724716/b6b2a19e-4786-11e6-862b-db69bb0e3d68.JPG)
![map size is 0_02](https://cloud.githubusercontent.com/assets/1768868/16724725/baab63da-4786-11e6-9ef0-8d113f47ab9a.JPG)

dogmaphobic:
I not only was able to reproduce this, I also found a few other oddities with the display of sizes and download status. Not sure what happened as this was all working at some point. Looking...

glody:
Checked on daily build HEAD:e6e0429 2016-07-20 18:58:20 -0400, this issue unable to reproduce. close it. thanks.

PR #3791, opened 2016-07-12
Fixes to Offline Map Cache

dogmaphobic:
This started with something @glody found and posted in #3778.

The original issue was that the Max Disk Cache was not being properly computed. Some changes from a few weeks ago caused it to be computed wrongly and always returning 0. This would cause the engine to keep pruning the database. That in turn triggered an original bug in the pruning code.

Pruning only occurs in the *Default Tile Cache*. The one where tiles are saved when you just move about. Tile sets you create are not pruned. The bug was that when pruning the Default Tile Cache, the logic looking for "common" tiles (to avoid deleting), that is, tiles used in more than one set, was not working. This in turn caused these tiles to be deleted and saved sets that had them got out of sync, gumming the entire work.

I fixed that whole pruning code, along with keeping better track of "common" tiles (and unique tiles for that matter).

The side effect is that trying to use the older database turned out to be way too much work. Therefore, when this new code is used, it will automatically WIPE the existing cache and start over from scratch.

As I was doing that any way, I removed several fields that ended up never being used.

 * New tile cache database engine and schema
 * Maintain not only a total count of tiles on a given set but also the count of *unique* tiles (the difference is the *common* tile set).
 * Fixed the function that format numbers for display. It was using uint32 and if you zoom out a lot, the total tile count far exceeds a 32-Bit number. It would be completely impractical to download that many tiles but we might as well show the proper count.
 * Fixed an issue that prevented resuming stopped downloads after restarting QGC.
 * Changed a bit the Offline Map UI. Don had done some work to make it fit on tiny screens but the cost on larger screens was too high. Things just didn't fit right.
    * Removed that "X" to close as it wasn't working at all (added a proper "Close" or "Cancel" button where appropriate)
    * Fixed the values shown in the info screen. The Default Tile Set was showing wrong numbers.
    * Added a count for *Unique* tiles for sets that have them.
    * Added the number of tiles (along with the size) in the main tile set list
 
With that said, I will revisit the UI for 3.1 as I still don't like what we have. At least the core architecture should now all be working properly.

@glody, this is an area to test as not only these changes were substantial, there hasn't been much testing with offline maps. This pruning issue was only caught because of the "always zero" bug. Thanks for all the testing you've been doing!

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -116,6 +116,8 @@
 #endif
 #endif
 
+#include "QGCMapEngine.h"
+
 QGCApplication* QGCApplication::_app = NULL;
 
 const char* QGCApplication::parameterFileExtension =    "params";
@@ -443,6 +445,11 @@ bool QGCApplication::_initForNormalAppBoot(void)
         settings.setValue(_settingsVersionKey, QGC_SETTINGS_VERSION);
         showMessage("The format for QGroundControl saved settings has been modified. "
                     "Your saved settings have been reset to defaults.");
+    }
+
+    if (getQGCMapEngine()->wasCacheReset()) {
+        showMessage("The Offline Map Cache database has been upgraded. "
+                    "Your old map cache sets have been reset.");
     }
 
     settings.sync();
--- a/src/QmlControls/OfflineMapButton.qml
+++ b/src/QmlControls/OfflineMapButton.qml
@@ -20,7 +20,8 @@ Rectangle
     property bool   checked:    false
     property bool   complete:   false
     property alias  text:       nameLabel.text
-    property alias  size:       sizeLabel.text
+    property int    tiles:      0
+    property string size:       ""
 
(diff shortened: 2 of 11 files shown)


DonLakeFlyer:
Did you test changes on iPhone 5s for sizing? Also if is going to dump the old cache do you do a showMessage to tell the user that map cache has been reset?

dogmaphobic:
> Did you test changes on iPhone 5s for sizing? Also if is going to dump the old cache do you do a showMessage to tell the user that map cache has been reset?

I tested against the 4s and the Nexus 4 (the smallest of the bunch).

> Do you do a showMessage to tell the user that map cache has been reset?

Nope. That happens early on in the boot process though. Let me see what the "preferences reset" dialog does. I could do what it does...

dogmaphobic:
Added dialog. The fun part will be figuring out how to get this commit over to master 8P

PR #3804, opened 2016-07-14
Adding Orbit Mode to Guided Bar

dogmaphobic:
WIP (in conjunction with https://github.com/PX4/Firmware/pull/4792)

Until we figure out a UX to handle these "smart modes", I just added a button to the *Guided Bar* to initiate *Orbit Mode*.

This needs a corresponding change in the firmware. If used against a different firmware version, it will be simply ignored.

@tumbili Your newest commit (the current pr-circleMode branch) is not working for some reason. I didn't have a lot of time to debug on the firmware side. The previous commit (691e664f48818844232757b8eb6899579e5f6a0e) is working fine (with the vehicle moving to the assigned radius distance though).

I also noticed you are not reading in (on the firmware side) parameters other than radius and center coordinates. Note the `PX4FirmwarePlugin::guidedModeOrbit()` function below to see how I setup the parameters. 

The current UI implementation (pressing the "Orbit" button) uses defaults (NAN) to all parameters. There is a commented off version, which sets the orbit center to the map center and a radius of 50m.


dogmaphobic:
@DonLakeFlyer I created a Json blob with the metadata but I could only make it show up in the Plan View if it was added to MavCmdInfoCommon.json, which didn't sound right.

This is what I have so far:

```
        {
           "id":                   4001,
           "rawName":              "MAV_CMD_SET_GUIDED_SUBMODE_CIRCLE",
           "friendlyName":         "Orbit",
           "description":          "Orbit the vehicle around a set point",
           "specifiesCoordinate":  true,
           "category":             "Basic",
           "param1": {
               "label":            "Radius:",
               "default":          3,
               "units":            "m",
               "decimalPlaces":    1
           },
            "param2": {
                "label":            "Velocity:",
                "default":          0,
                "units":            "m/s",
                "decimalPlaces":    1
            },
            "param3": {
                "label":            "Flight Altitude:",
                "default":          25,
                "units":            "m",
                "decimalPlaces":    1
            }
       }
```

@LorenzMeier 

I'm also a bit confused with all these modes/commands. For instance, there is a counter to this `MAV_CMD_SET_GUIDED_SUBMODE_CIRCLE` *submode* called `MAV_CMD_SET_GUIDED_SUBMODE_STANDARD`. The docs state:

"*This command sets the submode to standard guided when vehicle is in guided mode. The vehicle holds position and altitude and the user can input the desired velocites along all three axes.*"

Isn't this just plain *Position Control*? When would ever set this `MAV_CMD_SET_GUIDED_SUBMODE_STANDARD` mode?

DonLakeFlyer:
> I created a Json blob with the metadata but I could only make it show up in the Plan View if it was added to MavCmdInfoCommon.json, which didn't sound right.

Not sure why that wouldn't be right. What do you think is the problem with that? The JSON is just the definition for the command. Whether the firmware supports it or not is controlled by FirmwarePlugin::supportedCommands.

DonLakeFlyer:
Also I noticed now that I"m getting back into Survey that ComplexMissionItem is hacked to just be a survey item. I need to break that out to a sublcass so it's reusable.

dogmaphobic:
> Not sure why that wouldn't be right.

It's just that I assumed *common* meant that only metadata present in all implementations would go there while more specialized stuff would go into their own files.

This doesn't apply for fixed wing for instance.

dagar:
We've talked about having a FW and MC common.
https://github.com/mavlink/qgroundcontrol/pull/3730#issuecomment-231257921

DonLakeFlyer:
Ok. I get it now. The split out of FW/MR into separate JSON files is to support the implementation of the command being "different" between vehicle types and/or firmware types. The differences are in supported params and editing interface only. They do not define whether the command is available or not. 

The concept of a command not being "available" per vehicle type isn't part of the architecture right now. The FirmwarePlugin::supportedMissionCommands doesn't break out by vehicle type. That whole concept gets quite complicated when you bring in offline editing and vtol.

The bottomost baseline JSON file is meant to define the mavlink generic editor for the command. So if the JSON you specified includes info for all the possible params for the command then what you have done is correct. If the PX4 implementation of the command does not support all the params and the JSON has some of them missing. Then there should be both the baseline mavlink generic definition with full params defines. And a PX4 version with just the supported params defined.

I hope that makes this all clear as mud!

DonLakeFlyer:
Also given the FW/MR support difference there should be an isGuidedModeOrbitAvailable in the plugin or something like that. Then this is used to show in the guided bar or not. As it is now Orbit is going to show up for all vehicles and all firmwares which is incorrect.

dogmaphobic:
> Also given the FW/MR support difference there should be an isGuidedModeOrbitAvailable in the plugin or something like that.

Got it, though this seems to be a *for the time being* solution. `isGuidedModeOrbitAvailable` should be a flag somewhere up the tree otherwise we will have to have all sorts of hard-coded (software) dependencies, which will be quite hard to maintain.

> That whole concept gets quite complicated when you bring in offline editing and vtol.

QGC could *save* the last connected vehicle. On subsequent launches, it could reload that metadata until a new vehicle connection replaces it. On desktops, it could even be a File/Open function. Complicated... Though I would guess the vast majority of the cases, the vehicle tends to be the same. Just thinking out loud I guess. Either way, this is a different topic. Still waiting for @tumbili to see what's going on with the firmware. I will add the `isGuidedModeOrbitAvailable` for the time being so the button only shows up when appropriate.

DonLakeFlyer:
> we will have to have all sorts of hard-coded (software) dependencies, which will be quite hard to maintain.

Welcome the the world of supporting multi-firmware versions! Thinking about this a little more the isGuidedModeAvailable could be implemented in the FirmwarePlugin base class and test against supportedMissionCommands to determine availability.

src/Vehicle/Vehicle.cc
```diff
@@ -1637,6 +1637,15 @@ void Vehicle::guidedModeChangeAltitude(double altitudeRel)
     _firmwarePlugin->guidedModeChangeAltitude(this, altitudeRel);
 }
 
+void Vehicle::guidedModeOrbit(const QGeoCoordinate& centerCoord, double radius, double velocity, double altitude)
+{
+    if (!guidedModeSupported()) {
+        qgcApp()->showMessage(guided_mode_not_supported_by_vehicle);
```
DonLakeFlyer: camelCase?

tumbili:
@dogmaphobic I updated the Firmware side, circle mode should now work again.
I've seen that you have the desired altitude as a parameter. Isn't that a separate command already?

src/Vehicle/Vehicle.cc
dogmaphobic: > camelCase?

Where? What do you mean?

dogmaphobic:
> I've seen that you have the desired altitude as a parameter. Isn't that a separate command already?

First, there are two altitudes: The "center" (POI) altitude, which is part of the center XYZ coordinate and the flight altitude. Second, why send two commands when it can be all handle by one single command?

As to the POI altitude, this would imply the ability to move the gimbal. Not sure how that's going to work on the Firmware side.

*Vehicle*
```
cmd.param1 = radius;
cmd.param2 = velocity;
cmd.param3 = altitude;
```
*Center* (POI)
```
cmd.param5 = centerCoord.isValid() ? centerCoord.latitude()  : NAN;
cmd.param6 = centerCoord.isValid() ? centerCoord.longitude() : NAN;
cmd.param7 = centerCoord.isValid() ? centerCoord.altitude()  : NAN;
```

tumbili:
@dogmaphobic Ok got it, thanks!

tumbili:
@dogmaphobic I'm now pulling altitude, velocity, circle radius and circle origin from the command.
Tested with this branch of QGC.
So what's still missing is support for a gimbal based on the desired altitude of the circle origin.

tumbili:
>I'm also a bit confused with all these modes/commands. For instance, there is a counter to this MAV_CMD_SET_GUIDED_SUBMODE_CIRCLE submode called MAV_CMD_SET_GUIDED_SUBMODE_STANDARD. The docs state:

The intention of this was the following:
Since position control modes now has two submodes (standard position control and circle mode) I wanted to have commands for both of them. E.g. if you are in circle mode, how are you going to get out of there without having to switch modes (out and into position control)?

dogmaphobic:
I've added logic so the "firmware" will tell if it supports orbit mode or not. That in turn will determine if the "Orbit" button shows up in the *Guided Bar*. With that said, I don't like this. It's fine for testing (orbit is working) but it's way too much of a hack.

For instance, the UI below shows a situation where I've just hit the *Land* button. The mode up in the tool bar reflects the change and the vehicle is descending. Nonetheless, all buttons in the *Guided Bar* are still available and you can switch "modes" at will. The same is the case if you select *Orbit*. There is no clear way to tell the vehicle is currently "orbiting". Yes, we can dig it up but only using hacks. Is there a way to consolidate these things into logical groups that don't depend on deep introspection? I'm still utterly confused by all these different modes and submodes.

screen shot 2016-07-19 at 4 24 37 pm

dagar:
Would this make sense as a POSCTL submode? https://github.com/PX4/Firmware/blob/master/src/modules/commander/px4_custom_mode.h

DonLakeFlyer:
Even if I'm currently in Land mode, why can't I change my mind and do an RTL instead or Pause? Something may be going wrong ( a cat just jumped out into my landing spot) and I change my mind. Not sure why you would turn off the other modes in that case?

If you do Orbit again, while you are already orbiting does something bad happen?

DonLakeFlyer:
I get what you are saying but I just don't think it needs to be that restrictive.

dogmaphobic:
It's just that from an UX perspective, nothing changed. The only difference is that the mode up on top the toolbar switched to *Land*. If you even notice it as you would to busy doing other things to notice.

Again, this is all fine for just testing all these modes. I just think we should have some idea of a roadmap for an UI to shoot for. It feels like we're just throwing things at random, the code base underneath it is becoming unwieldy, and users are only going to get even more confused.

With that said, this PR is ready to merge. Any other change would be related to UI and it should be in its own PR.

PR #3808, opened 2016-07-16
land don't autoContinue

dagar:
Ignoring the friendlyEdit check is a bit hacky.

### DIFF
--- a/src/MissionManager/SimpleMissionItem.cc
+++ b/src/MissionManager/SimpleMissionItem.cc
@@ -451,9 +451,11 @@ QmlObjectListModel* SimpleMissionItem::comboboxFacts(void)
 bool SimpleMissionItem::friendlyEditAllowed(void) const
 {
     if (_missionCommands->contains((MAV_CMD)command()) && _missionCommands->getMavCmdInfo((MAV_CMD)command(), _vehicle)->friendlyEdit()) {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Why not autocontinue after land? There could be other things past there like turning off the camera or what not.

I"m not sure what I think about turning friendly edit on when autoContinue is not set. The user has no way of knowning consistently now whether this is set or not. At a minimum put in an Issue for fixing that somehow.

dagar:
You should still be able to autocontinue after land, but doesn't it seem like the appropriate default the vast majority of the time?

This seemed like the most appropriate way to remove the "mission finished, loitering" message you get in PX4 after landing.

DonLakeFlyer:
Seems like a strange way to fix firmware problem. Also what does ArduPilot do when you land? Given the fact that the average user is never exposed to autocontinue with missions I don't think it's right to set it to off. Suddenly things aren't going to work if they have things after land and they will have no idea why.

dagar:
Maybe only autocontinue=false if it's land without anything after. I can do something on the PX4 side if you don't want to mess with autocontinue.

DonLakeFlyer:
I don't think setting it to false is the right thing. We don't really know what the suer wants.

PR #3811, opened 2016-07-17
Use exact conversion factor for feetToMeters translation

NaterGator:
Found this when converting between feet/meters frequently on an experimental draggable waypoint fact editor. Doing so causes the value to tend to grow continuously.


DonLakeFlyer:
Internet search fails me again!

DonLakeFlyer:
Can you make a pull against Stable_V3.0 for this as well?

DonLakeFlyer:
Did you happen to look at the other conversion as well for correctnes?

NaterGator:
Closing to implement what we discussed on gitter.

PR #3813, opened 2016-07-17
Switch to exact conversion constants

NaterGator:
Use constant's inverse to ensure we don't accumulate error when converting back and forth

### DIFF
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -17,9 +17,21 @@
 #include "QGroundControlQmlGlobal.h"
 
 #include <QDebug>
+#include <QtMath>
 
 #include <limits>
 #include <cmath>
+
+// Conversion Constants
+// Time
+const qreal FactMetaData::UnitConsts_s::secondsPerHour = 3600.0;
+
+// Velocity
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
I'm not sure toReal is the way to go here since params are 32bit floats?

NaterGator:
Using toReal should just store the platform's preferred floating point representation. As long as the facts use [QVariant::toFloat](http://doc.qt.io/qt-5/qvariant.html#toFloat) they'll be converted automatically. I believe that all parameter writes happen [here](https://github.com/mavlink/qgroundcontrol/blob/e91b6713b021ef736ce4c8929c889814763b5753/src/FactSystem/ParameterLoader.cc#L627-L659) so we should be fine.

DonLakeFlyer:
My concern is round tripping through 64<->32 will also cause values to change when going back/forth through qml, user change, out to vehicle and back? Maybe not.

NaterGator:
Round is converting degrees to integral centidegrees, so that conversion should be precise regardless of the platform and never get close to the numeric limits of int. My thought is that the 100.0 literal for centidegree to degree will be the native platform floating point type, and the toReal() value will be native floating point representation of an integer which _should_ be precise regardless of the number of bits. I don't think there is any issue here...

DonLakeFlyer:
All good to go then. Thanks Nate.

PR #3814, opened 2016-07-17
Switch to exact conversion constants

NaterGator:
Use constant's inverse to ensure we don't accumulate error when converting back and forth


NaterGator:
@DonLakeFlyer is this just a CI quirk?

DonLakeFlyer:
Think so. Travis OSX had been a mess for days now. It keeps hanging up and/or running really slow.

Issue #3802, opened 2016-07-13
Survey: Mission stats


DonLakeFlyer:
Although not solely a feature for Survey generating a set of stats associated with the mission such as flight time and distance, battery usage and so forth is important. Let's use this issue to discuss what information should be collected about a mission.

DonLakeFlyer:
I'm in the process of collecting up the information from the various issues already entered on this...

birchera:
Related to https://github.com/mavlink/qgroundcontrol/issues/3772

In my experience the values that are interesting to know when planning missions are:
* Total flight distance
* Total time
* Max telemetry distance

For VTOLs:
* Splits hover/cruise distance
* Splits hover/cruise time

For Survey missions:
* Number of images that will be taken
* Covered area

@kellysteich FYI

DonLakeFlyer:
@andke In #3772 You have a speed value which I don't quite understand. Can you explain more?

DonLakeFlyer:
FYI to folks looking: @birchera and friends are actively working on support for this.

tubeme:
Another important parameter set is the Altitude information:

Higest altitude
Lowest altitude
Avarage altitude
Cumulative ascent
Cumulative descent

This could orient the pilot about the performance needed for the mission in terms of energy requirements.

DonLakeFlyer:
This is in a good state for 3.2

PR #3818, opened 2016-07-18
[DO NOT MERGE, WIP] Mission stats

birchera:
This PR will eventually solve https://github.com/mavlink/qgroundcontrol/issues/3802.

The idea is to display the mission specific values on the right, while waypoint specific values are listed on the left, where they used to be. It always shows total mission distance and time, as well as max telemetry distance. In case of a vtol connected, it also shows the splits for cruise and hover.

If the active waypoint element is a survey item, the left side contains the expected number of camera shots and the covered area. To this end, the unit of area can be chosen in general settings.


birchera:
screenshot from 2016-07-18 08 28 13

DonLakeFlyer:
Fantastic work! Give me some time today to look at it.

DonLakeFlyer:
I'd like to see this pull split up into multiple pulls. This will make things easier to review and then we can merge in a part at a time. I'd propose the grouping:
1) Settings changes for area units and fact meta data changes for the new translations.
2) Plumb the properties for the calculated values through MissionController and ComplexMissionItem. Initially with just test values. Expose in the status display as you have it. I think that is good enough for now. 
3) Then a separate pull for the calculation of each value. Except for ones that are trivial enough to put together.

FYI: I'll be refactoring ComplexMissionItem into a base class and SurveyItem. I'd like to get that done soon, but we'll have to coordinate the timing since you will be in there.

DonLakeFlyer:
Also should have mentioned that this all looks to be heading int he correct direction. So I see no reason for those multiple pulls to be merged in once they pass CI.

birchera:
This PR gets replaced by multiple smaller ones.

PR #3828, opened 2016-07-20
[Mission stats] Added area translation

birchera:

It adds a choice for area unit, which will then later be used to display the area of a survey mission.

It supports:
* Square meters
* Square kilometers
* Hectares
* Square feet
* Acres
* Square miles


DonLakeFlyer:
All looks great. Can you squash down to a single commit and then I'll merge.

PR #3829, opened 2016-07-20
[Mission stats] Dummy mission stats

birchera:
This PR is part of the new [mission stats](https://github.com/mavlink/qgroundcontrol/issues/3802) feature.

This PR exposes the mission stats values, displaying some dummy values. The actual calculations will follow in separate PRs.

Additional values are:
* Total mission distance
* Total mission time
* For VTOL vehicles both values split in cruise and hover
* If a survey item is selected:
 * Number of shots
 * Covered area

PR #3830, opened 2016-07-20
[Mission stats] Computing number of shots and covered area

birchera:

This PR requires merging of https://github.com/mavlink/qgroundcontrol/pull/3828 and https://github.com/mavlink/qgroundcontrol/pull/3829 first.

* Computes the number of shots of a survey item, based on the distance and the selected triggering distance.
* Computes the area of the drawn polygon in the survey item and displays it as Covered area.

### DIFF
--- a/src/MissionEditor/MissionItemStatus.qml
+++ b/src/MissionEditor/MissionItemStatus.qml
@@ -34,25 +34,30 @@ Rectangle {
 
     property real   _collapsedWidth:    valueGrid.width + (margins * 2)
     property bool   _expanded:          true
+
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
This is going to need a rebase once the other pulls go in. Then I'll take another look at  it.

DonLakeFlyer:
@birchera Ok, other pulls are merged. Thanks. Can you rebase and squash and then I'll take a look.

DonLakeFlyer:
Looks good. In looking at the UI some more I think the Survey specific data should possibly be in the survery item editor itself instead of in the mission stats area. Fine for now. But I think the ui needs some thinking.

PR #3841, opened 2016-07-23
ArduSub: Disable radio setup page and component

rjehangir:
This PR disables the radio setup page when an ArduSub vehicle is connected. The vehicle is detected with MAV_TYPE_SUBMARINE.

The radio page is not needed for ArduSub because it is controlled exclusively through the use of the QGC joystick feature. The radio setup page is confusing for ArduSub users and the calibration process results in settings that are changed to undesirable values.

Please let me know if you have any questions or require any changes. I really appreciate your support!

Thanks!


DonLakeFlyer:
I would instead modify the autopilot plugin such that it doesn't create/add these components to the list if the vehicle is a sub. This way the buttons don't show up at all.

rjehangir:
Hi Don, that's a good idea. I played around with that originally but this version required less changes. I'll close this and do it the other way.

PR #3852, opened 2016-07-25
[Mission stats] Offline planning settings

birchera:

It gives the user the option to select in general settings:
* Vehicle type: {Fixedwing, Multicopter, VTOL}
* Nominal cruise speed (only visible if not selected multicopter)
* Nominal hover speed (only visible if not selected fixedwing)

These values will be used to calculated a flight duration estimate while planning missions.


DonLakeFlyer:
Can you add Rover and Sub to this? Those are active vehicle types as well which can have special differentiation for features when planning missions. For example Rover can have a different implementation/editing of various mission commands just like MR/FW currently does.

I'm not quite sure how to handle the cruise/hover speed concept with respect to other vehicle types like rover/sub. Should each vehicle type have it's own offline speed?

@birchera Maybe a skype discussion on this would be useful since it gets complex pretty fast. I can also talk through what I was thinking here.

DonLakeFlyer:
I'm ok with this going through this way for now to make forward progress. We can figure our additional behavior after further discussion.

Issue #3798, opened 2016-07-12
Fix grid generation line bug


DonLakeFlyer:
You can see in the image the bad line which is not parallel. Not quite sure what polygon shape cause it to happen:
![screen shot 2016-07-12 at 3 40 11 pm](https://cloud.githubusercontent.com/assets/5876851/16786069/ffb98144-4846-11e6-9563-d772154cc1ea.png)

NaterGator:
@DonLakeFlyer it is because the short segment running to the west fell outside  of the polygon clip region.

DonLakeFlyer:
I sort of get what you are saying but I of course can get it to repro any more. Should have saved the mission!

DonLakeFlyer:
Here is a mission that shows the bug: https://gist.github.com/DonLakeFlyer/dd4e2f5923732d493e3d957f91ff6668

PR #3856, opened 2016-07-26
Survey: Adjust grid line direction after polygon clip

DonLakeFlyer:
Fix for #3798

### DIFF
--- a/src/MissionManager/ComplexMissionItem.cc
+++ b/src/MissionManager/ComplexMissionItem.cc
@@ -396,6 +396,24 @@ void ComplexMissionItem::_intersectLinesWithPolygon(const QList<QLineF>& lineLis
     }
 }
 
+/// Adjust the line segments such that they are all going the same direction with respect to going from P1->P2
+void ComplexMissionItem::_adjustLineDirection(const QList<QLineF>& lineList, QList<QLineF>& resultLines)
+{
+    for (int i=0; i<lineList.count(); i++) {
+        const QLineF& line = lineList[i];
+        QLineF adjustedLine;
+
+        if (line.angle() > 180.0) {
+            adjustedLine.setP1(line.p2());
+            adjustedLine.setP2(line.p1());
+        } else {
+            adjustedLine = line;
+        }
+
+        resultLines += adjustedLine;
+    }
+}
+
 void ComplexMissionItem::_gridGenerator(const QList<QPointF>& polygonPoints,  QList<QPointF>& gridPoints)
 {
     double gridAngle = _gridAngleFact.rawValue().toDouble();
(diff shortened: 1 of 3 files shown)


src/MissionManager/ComplexMissionItem.cc
```diff
@@ -396,6 +396,24 @@ void ComplexMissionItem::_intersectLinesWithPolygon(const QList<QLineF>& lineLis
...
 }
 
+/// Adjust the line segments such that they are all going the same direction with respect to going from P1->P2
+void ComplexMissionItem::_adjustLineDirection(const QList<QLineF>& lineList, QList<QLineF>& resultLines)
+{
+    for (int i=0; i<lineList.count(); i++) {
+        const QLineF& line = lineList[i];
+        QLineF adjustedLine;
+
+        if (line.angle() > 180.0) {
```
birchera: I would expect that the flipped lines would have an angle of 180.0°, which means they wouldn't be flipped back.

And if you change it to >=, this statement restricts the choice of grid angles to [0° - 180°), as otherwise the segments are flipped. Instead I would propose to alter the check to something like
```cpp
if (fmod(fabs(line.angle() - gridAngle), 180.0) >= 90)
```

DonLakeFlyer: That's not really what this code does.  It's just ordering the endpoints of the parallel grid lines (P1, P2) such that the vectors of the lines are all pointing in the same direction. The lines themselves stay the same, only the endpoints are flipped. This way when you connect the turn around from the end of one parallel line to the beginning of the next the connection is always on the edge of the polygon instead of possibly traversing the center of the polygon as shown in the image.

I think it's correct, unless I'm misunderstanding your point.

DonLakeFlyer: One more piece of information. The entry point for the survey is still not yet implemented and at this point is arbitrary based on polygon intersection. This endpoint flipping code will affect it as well. But the user themselves really needs control over it. That is NYI at this point. I'm not expecting them to set grid angle to adjust that. Or at least not at this point, since I haven't thought that much about it.

NaterGator: I've found a good approach is to let the user set an initial heading via a radial control. That affords them the ability to intuitively optimize fixed wing survey plans for prevailing winds.

DonLakeFlyer: But that also says that grid angle also informs initial entry point. I was thinking that wasn't the case, but I can see now how it may be. What's the concensus on that?

DonLakeFlyer: I think it's more complex than my simple statement. I'm going to craete a new discussion for this so we an continue there.

DonLakeFlyer:
Survey area grid entry discussion is here: https://github.com/mavlink/qgroundcontrol/issues/3864

PR #3858, opened 2016-07-26
Remove motor test

DonLakeFlyer:
Numbering on motor command is wrong (or strange), so wrong motors spin when using sliders. Fix for #3850


DonLakeFlyer:
This needs to go into 3.0 as well.

DonLakeFlyer:
No need for 3.0. Motor Test is not in 3.0

PR #3868, opened 2016-07-27
Fix download custom firm file err: no file name specified

JcZou:
down bin file to pixhawk fail,because it do not set value for _binFilename in _binLoad

### DIFF
--- a/src/VehicleSetup/FirmwareImage.cc
+++ b/src/VehicleSetup/FirmwareImage.cc
@@ -436,5 +436,7 @@ bool FirmwareImage::_binLoad(const QString& imageFilename)
     
     binFile.close();
     
+    _binFilename = imageFilename;
+    
     return true;
 }


DonLakeFlyer:
Need to look at the repro for this. Possible candidate for 3.0 point release.

DonLakeFlyer:
What are the steps to reproduce this? I just tried installing from the custom firmware option and it worked fine for me.

DonLakeFlyer:
@JcZou ?

JcZou:
Sorry to reply late. I use Stable V3.0,and download my own .bin file via custom firmware download.I think you can reproduce it while download a .bin file

DonLakeFlyer:
Ah, bin file. That's why I couldn't repro. Thanks.

DonLakeFlyer:
Forcing CI to run again.

DonLakeFlyer:
Tested fine. Thanks or the pull.

PR #3871, opened 2016-07-27
Fixes to iOS App Store build. [ci skip]

dogmaphobic:
Fixing the .plist file for the App Store.
Commenting off some automated steps for the build script. I've been building it manually for the time being.


dogmaphobic:
Oops, I guess it didn't skip CI after all...

dagar:
Should we try to do travis-ci master builds deploy directly to testflight?

https://gist.github.com/johanneswuerbach/5559514

DonLakeFlyer:
Would be awesome if doable. Then the other question would be can itunesconnect handle all the uploads? Android seems to be ok with it. Also I think there may be an issue with TestFlight forcing users to install a new build when one becomes available. If true that could get really annoying.

dagar:
If that's the only problem we can hack something on travis-ci to limit the push to once/day.

dogmaphobic:
The only issue I see is that QGC needs to be built with the commercial licensed Qt for App Store deployment. You can't use the generic, open source version.

DonLakeFlyer:
Is that just pulling a license file into travis?

dogmaphobic:
I assume they have a version with a license file. The one I have checks the license online every time it runs.

PR #3879, opened 2016-07-28
Joystick button function selector for ArduSub

rjehangir:
This PR contains changes for a joystick button function selector specifically to support ArduSub. It is integrated so that it does not have any affect whatsoever on existing features and appearance. The button function selector allows selection of the button "function" and the "shift function" directly from the joystick page, greatly increasing usability for ArduSub users.

Additionally, there are some minor changes along the way:

- The `throttleMode` radio buttons on the joystick page are now hidden for ArduSub and forced to use "full down stick is zero throttle" to avoid issues during use.
- The FactComboBox implementation was fixed to allow for null Facts
- The `manualControlReservedButtonCount` function in `APMFirmwarePlugin` was made non-final so it could be overridden by the ArduSub plugin.

I've tested this with a Pixhawk with ArduSub and with the ArduCopter mock link and it works well. Please feel free to test and let me know if there are any issues, comments, changes, etc. 

Thanks!

### DIFF
--- a/src/FactSystem/FactControls/FactComboBox.qml
+++ b/src/FactSystem/FactControls/FactComboBox.qml
@@ -10,7 +10,7 @@ QGCComboBox {
     property Fact fact: Fact { }
     property bool indexModel: true  ///< true: model must be specifed, selected index is fact value, false: use enum meta data
 
-    model: fact.enumStrings
+    model: fact ? fact.enumStrings : null
 
     currentIndex: indexModel ? fact.value : fact.enumIndex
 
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
I need to think about this a little. It should be handled more from a plugin style of things as opposed to checking for sub. But need to stare at it some more.

DonLakeFlyer:
@rjehangir Another thing that would be nice is if you can add ArduSub as a MockLink that we can test with.

rjehangir:
@DonLakeFlyer Okay. I completely agree. I searched for a good way to do this with the ArduSub plugin but didn't come up with much. I'm happy to change it if you have ideas. 

I've got a MockLink in the works as well. I'll try to get that done soon to help out with testing.

rjehangir:
From my research so far, it looks like the best solution might be to put the button function selector into a separate qml file and then conditionally include it with a qml `Loader`. Thoughts?

DonLakeFlyer:
Rusty, I'm going to merge this in. But can you create an Issue to track this so we can figure out the right way to do it. There are a number of issues:
* Joystick config is a global settings, so modifying the setup based on vehicle doesn't quite make sense
* The if (sub) testing should be done more in a firmware plugin style. Like plugin->supportThrottleCenterZero of something like that.
* A mechanism to expose the button ui in  more generic way

rjehangir:
Closing after merge. Fixes to @DonLakeFlyer's last comment are addressed in a new PR, #3896, which fixes this and other semantic issues.

Issue #3894, opened 2016-08-01
Make the ArduSub joystick function selector a plugin feature


rjehangir:
Per Don's comments on PR #3879, there are a number of issues:

- Joystick config is a global setting, so modifying the setup based on vehicle doesn't quite make sense
- The `if (sub)` testing should be done more in a firmware plugin style. Like `plugin->supportThrottleCenterZero` of something like that.
- A mechanism to expose the button ui in more generic way

rjehangir:
These issues are addressed in PR #3896.

@DonLakeFlyer, I'm not 100% sure what you meant by "mechanism to expose the button ui in a more generic way", but I think I've accomplished that. To do this in firmware in ArduSub, we added a "AP_JSButton" library. The interface is now visible when the plugin `supportsJSButton`, which could be done on any vehicle type.

rjehangir:
Changes have been pulled from #3896. Closing.

PR #3896, opened 2016-08-01
Fix plugin semantics for ArduSub additions

rjehangir:
Per Issue #3894, this fixes the way various UI options are detected using a plugin based method instead of direct detection by vehicle type. This fixes this in a number of places beyond those mentioned in #3894.

Specifically, the following methods are now part of `Vehicle` and `FirmwarePlugin`:

- `supportsManualControl`
- `supportsThrottleModeCenterZero`
- `supportsRadio`
- `supportsJSButton`

Please let me know if there are any questions or required changes.

-Rusty


DonLakeFlyer:
Perfect. Still thinking about some way to plugin'ize the js button stuff.

PR #3914, opened 2016-08-03
Use human appropriate values for bytes in log downloader

NaterGator:
Also report the current transfer rate in the currently downloading status to help expose regressions in performance.


NaterGator:
image
This is the new look

PR #3863, opened 2016-07-26
[Mission stats] Compute distance & time

birchera:
This PR is part of the new [mission stats](https://github.com/mavlink/qgroundcontrol/issues/3802) feature.

This PR will require a rebase once https://github.com/mavlink/qgroundcontrol/pull/3852 and https://github.com/mavlink/qgroundcontrol/pull/3830 are merged.

The following is computed:
* Mission distance
* Mission time
* Maximum telemetry distance

For the case of a VTOL the code added also computes
* distance estimate for hover and cruise individually
* time estimate for hover and cruise individually

and outputs them all in the previously established mission statistics display.

PR #3915, opened 2016-08-04
[Mission stats in Survey Feature]: check when survey item is first mission item

AdyashaDash:
As pointed out [here](https://github.com/mavlink/qgroundcontrol/pull/3863#discussion_r73190742) in PR #3863, while calculating mission distance, we fail to take into account the case when survey mission item is the first waypoint after home. This PR takes care of that.


AdyashaDash:
@DonLakeFlyer Could you please have a look at this? :)

PR #3935, opened 2016-08-12
Add support for a custom board

LorenzMeier:
This adds support for another board. Something is not quite right with the boot loader setup as its not showing the next step to flash.


LorenzMeier:
@DonLakeFlyer Good to go now.

PR #3942, opened 2016-08-14
VTOL transition switch UI

LorenzMeier:
This adds the transition switch to the simple flight mode UI. The switch is only displayed if the airframe is a VTOL. Also updates airframe and param meta so QGC shows these new parameters correctly in their correct group.

@tubeme @kd0aij

### DIFF
--- a/src/AutoPilotPlugins/PX4/PX4SimpleFlightModes.qml
+++ b/src/AutoPilotPlugins/PX4/PX4SimpleFlightModes.qml
@@ -197,7 +197,7 @@ Item {
                             Row {
                                 spacing: ScreenTools.defaultFontPixelWidth
 
-                                property Fact fact: controller.getParameterFact(-1, "RC_MAP_TRANS_SW")
+                                property Fact fact: controller.getParameterFact(-1, "RC_MAP_TRANS_SW", false)
                                 visible: (controller.vehicle.vtol && controller.parameterExists(-1, "RC_MAP_TRANS_SW"))
 
                                 QGCLabel {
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Is the VTOL transition parameter only in newer firmwares. If so, the code need to guard against the parameter not being available.

LorenzMeier:
Good point, done.

tubeme:
Not a bad idea to have some sort of indication in the interface in which mode do we fly during flight. Especially valuable if we are outside LOS flying FPV. In mission we could want manual override and make transition and then back. In all this cases we need indication of which mode we are in. Of course we see the camera video and can recognize by it. But often there are cases where we don't have video in the mission.

src/AutoPilotPlugins/PX4/PX4SimpleFlightModes.qml
```diff
@@ -193,6 +193,26 @@ Item {
                                     indexModel: false
                                 }
                             }
+
+                            Row {
+                                spacing: ScreenTools.defaultFontPixelWidth
+
+                                property Fact fact: controller.getParameterFact(-1, "RC_MAP_TRANS_SW")
```
DonLakeFlyer: This will still pop a missing parameter error. Take a look at APMSensorsComponent.qml for examples of how to structure it to prevent that.

LorenzMeier:
@DonLakeFlyer Done, good to go.

src/AutoPilotPlugins/PX4/PX4SimpleFlightModes.qml
```diff
@@ -197,7 +197,7 @@ Item {
                             Row {
                                 spacing: ScreenTools.defaultFontPixelWidth
 
-                                property Fact fact: controller.getParameterFact(-1, "RC_MAP_TRANS_SW")
+                                property Fact fact: controller.getParameterFact(-1, "RC_MAP_TRANS_SW", false)
```
DonLakeFlyer: I forgot all about that support!

PR #3955, opened 2016-08-19
Break out SurveyMissionItem->ComplexMissionItem class hierarchy

DonLakeFlyer:
Previously ComplexMissionItem was really a Survey mission item which is incorrect. Now the new SurveyMissionItem class used ComplexMissionItem as a base class. ComplexMissionItem was restructured to be a base class for all complex mission item types. This will allow for the creation of new Complex mission item types like structure scan.

Still work to do here. The loading process from json need to be changed to create the correct complex mission item type dynamically. Code at this point still hardwired to create survey item.


DonLakeFlyer:
Fix for #3870

PR #3960, opened 2016-08-20
Survey update

DonLakeFlyer:
* Facts have meta data so you now gets units and unit conversion
* Restructured polygon buttons
* Moved survey stats from bottom mission item stats area up into survey editor
* Corrected camera shots to be 0 when no camera is triggered

screen shot 2016-08-19 at 9 03 58 pm


NaterGator:
This is just an observation from the screenshot, but I always thought the abbreviations were for "Start" and "End" so I was surprised to see both S (I guess for survey?) 

Can we do something like S/E so the user is cued into where the craft will fly, and maybe texture the polygon fill with a camera vector graphic?

DonLakeFlyer:
The two "S"es (for Survey) are just NYI stuff. Item abbreviations don't quite work for complex items which have an entry and an exit waypoint. We'll figure something out.

Issue #3797, opened 2016-07-12
Survey Phase 1 Implementation


DonLakeFlyer:
This is the discussion area for Phase 1 Implementation: **Flight Path**

**The first feedback should be on how I'm trying to structure the survey discussion and work using this issue. Then the question is what features are we missing for Phase 1? Within the features I've listed below there is much detail missing which needs to be filled in and/or correction/additions to the initial information I've entered.**

**My hope is that this provides not only a framework for discussion but also the framework for the creation and dispersal of a detailed set of work items to individual devs.**

We need to start with the basics, so Phase 1 Implementation will not be very sexy and will deal solely with the flight path of the vehicle. We will also not worry too much about overly pretty ui at first. If the base capability is architected correctly, we can put whatever ui we want on it. I want to take things one step at a time to focus the discussion and implementation.

An example of base flight path functionality for phase 1 versus something which will be dealt with during future phase work:
Phase 1 will include support to specify the distance between parallel lines of the grid (grid spacing). Phase 1 will not include such things as image overlap or camera sensor information. Although image overlap and camera information affects grid distance, they are simply survey meta data to simplify the process of setting the grid distance (amoung other things). The base capability needed to support them is still just grid spacing.

The process will be that the discussion here leads to updates/adds to the feature list and feature details as well as set of individual works items.

Flight Path feature set (details for each below):
- Polygonal area which defines the survey area
- Survey is treated as a single entity within the overall mission plan
- Ability to have multiple survey areas in a single mission
- Ability to re-edit survey after save/re-load from file
- Values needed to generate a simple grid flight pattern over the polygon
- Values needed to fly a correct turnaround to the next grid line
- Ability to modify the above values at any time and have the flight path dynamically update
- Conversion of above information into a set of waypoints
- Load/Save to file of a mission which includes a survey
- Mission stats: Flight time, battery usage, etc.
- Ability for user to specify information which determines entry into survey area

DonLakeFlyer:
* Polygonal area which defines the survey area

DonLakeFlyer:
* Ability to have multiple survey areas in a single mission

DonLakeFlyer:
* Ability to re-edit survey after save/re-load from file (doesn't degenerate into big pile of waypoints)

DonLakeFlyer:
Values needed to generate a simple grid flight pattern over the polygon

DonLakeFlyer:
Values needed to fly a correct turnaround to the next grid line

[...]

PR #3963, opened 2016-08-20
Survey camera

DonLakeFlyer:
Entering camera values will recalc grid spacing and camera trigger distance:
screen shot 2016-08-20 at 12 39 41 pm
Camera values are not yet saved as meta data in saved mission file.

Related to #3797


src/MissionEditor/SurveyItemEditor.qml
```diff
@@ -115,6 +152,79 @@ Rectangle {
...
+            QGCTextField {
+                id:         sensorHeightField
+                unitsLabel: "mm"
+                showUnits:  true
+                text:       _cameraInfoCanonSX260.sensorHeight.toString()
+
+                onEditingFinished: recalcFromCameraValues()
+            }
+
+            QGCLabel { text: qsTr("Image overlap:") }
```
birchera: This is a cool feature! I would suggest to have separate front and side overlap.

DonLakeFlyer: Yeah I was wondering about that.

PR #3967, opened 2016-08-22
Mavlink 2.0 support

DonLakeFlyer:
* Centralize all mavlink message sending in MAVLinkProtocol to provide single point for mavlink version setting
* Add Mavlink Version options to Settings/Mavlink. Default is "Default to 1, switch to 2 if Vehicle sends version 2"

### DIFF
--- a/src/comm/MAVLinkProtocol.cc
+++ b/src/comm/MAVLinkProtocol.cc
@@ -35,6 +35,7 @@
 #include "QGCApplication.h"
 #include "QGCLoggingCategory.h"
 #include "MultiVehicleManager.h"
+#include "QGroundControlQmlGlobal.h"
 
 Q_DECLARE_METATYPE(mavlink_message_t)
 
@@ -252,7 +253,7 @@ void MAVLinkProtocol::receiveBytes(LinkInterface* link, QByteArray b)
                 {
                     mavlink_message_t msg;
                     mavlink_msg_ping_pack(getSystemId(), getComponentId(), &msg, ping.time_usec, ping.seq, message.sysid, message.compid);
-                    _sendMessage(msg);
+                    sendMessage(link, msg, getSystemId(), getComponentId());
                 }
             }
 
@@ -393,7 +394,7 @@ void MAVLinkProtocol::receiveBytes(LinkInterface* link, QByteArray b)
 
                     // Only forward this message to the other links,
                     // not the link the message was received on
-                    if (currLink && currLink != link) _sendMessage(currLink, message, message.sysid, message.compid);
+                    if (currLink && currLink != link) sendMessage(currLink, message, message.sysid, message.compid);
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
FYI: No signing support yet, coming.

@LorenzMeier Can you review this to make sure it is correct?

src/comm/MAVLinkProtocol.cc
```diff
@@ -426,50 +427,34 @@ int MAVLinkProtocol::getComponentId()
...
+        break;
+    }
+
     // Create buffer
     static uint8_t buffer[MAVLINK_MAX_PACKET_LEN];
-    // Rewriting header to ensure correct link ID is set
-    static uint8_t messageKeys[256] = MAVLINK_MESSAGE_CRCS;
-    mavlink_finalize_message_chan(&message, systemid, componentid, link->getMavlinkChannel(), message.len, message.len, messageKeys[message.msgid]);
+    // Question: What is min_length
+    mavlink_finalize_message_chan(&message, systemid, componentid, link->getMavlinkChannel(), message.len, message.len, mavlink_get_crc_extra(&message));
```
vooon: Use `mavlink_get_msg_entry()` to get `const mavlink_msg_entry_t *` which store `crc_extra`, `length`, `min_length`.

`min_length` size of the struct before `<extensions />` tag, so that sizes are different only on MAVLink 2 messages.

vooon: *Do not forget to check pointer against `nullptr`.*

DonLakeFlyer: Awesome, thanks.

DonLakeFlyer: @vooon I don't see min_length in mavlink_msg_entry_t:
```
typedef struct __mavlink_msg_entry {
	uint32_t msgid;
	uint8_t crc_extra;
	uint8_t msg_len;
	uint8_t flags;             // MAV_MSG_ENTRY_FLAG_*
	uint8_t target_system_ofs; // payload offset to target_system, or 0
	uint8_t target_component_ofs; // payload offset to target_component, or 0
} mavlink_msg_entry_t;
```

vooon: @DonLakeFlyer sorry, i was messed up with my C++11 implementation. But that good proposal to extend `mavlink_msg_entry_t`.

One example where min_length != length is ardupilotmega MEMINFO.

LorenzMeier:
@DonLakeFlyer Do you need the extension for the MAVLink struct for this to work?

DonLakeFlyer:
@LorenzMeier I'm not currently using any extended 2.0 mavlink messages. I just don't quite understand the right way to get/send the min length and length params to the mavlink_finalize_message_chan. Should I call mavlink_get_msg_entry and use mavlink_msg_entry_t.msg_len as min_length? And then use message.len filled in by encode call as the actual length?

vooon:
@DonLakeFlyer why you even need finalize? Each `xxx_pack_chan()` do finalization and passing right length & min_length (`#define`).

@LorenzMeier it may be more logical to have both lengths in msg_entry.

LorenzMeier:
@DonLakeFlyer Let me know if you want cross-testing. We also need to work out the handshaking - if both components understand 2.0 the system should switch automatically.

DonLakeFlyer:
@LorenzMeier Depends on setting on Mavlink Settings page:      
```
enumStrings << "Always use version 1" << "Default to 1, switch to 2 if Vehicle sends version 2" << "Always use version 2";
 ```

LorenzMeier:
What's really missing here is "Discover if vehicle supports 2.0, then switch to it". That's the only user-friendly version and what should be the default.

LorenzMeier:
@DonLakeFlyer Here is a test branch implementing what I suggested. I can update the MAVLink submodule in QGC if you like. https://github.com/PX4/Firmware/pull/5396

DonLakeFlyer:
Ok, got it.

DonLakeFlyer:
Mavlink 2 bit is here: https://github.com/PX4/Firmware/pull/5403

DonLakeFlyer:
I"ll do the capabilities thing in a separate pull.

Issue #3801, opened 2016-07-13
Survey flight path turn around


DonLakeFlyer:
Let's discuss exactly what information is needed and how to geneate the turn-around waypoints between grid lines.

DonLakeFlyer:
From @Andke
 Hi, I have long experience in mapping missions, often in narrow valleys, and over very varying altitude.
The planes in use, are very aggressively tuned (high roll angle) to turn efficiently in terms of space.
Here are a few tips for better planning, that I would love to see a tool use:
![img_20160710_104802_337](https://cloud.githubusercontent.com/assets/1468842/16714560/5f2da2b2-46c7-11e6-9af0-f2dc13028071.jpg)

Please notice the small(normal) acceptance radius in WP1,2,3  - observe the way plane turns, then flies towards WP3, then turn again.

Now see that when it reaches WP4, WP5's acceptance radius is big (actually touch WP4 - resulting in a smooth single turn between and WP5)   -   resulting in better turn, and nice aligning with next path. 
The trick is, basically, that the acceptance radius of WP5,7 should be so big, it actually touches WP4,6  - This way the nav controller will align ASAP, not turn-level-turn.


Also, please allow to "inspect" the survey - see each WP , and set/modify each WP individually.

DonLakeFlyer:
Some question:
* I would expect that turn-around needed for fixed wing and turn-around for multi-rotor is different. Is that correct and necessary?
* Is the @AndKe example above only one possible way of doing a good turn-around? Or are there multiple mechanisns needs that the user choices from.
* In implementing @AndKe style turn-around what variables does the user need to provide in order to generate?

AndKe:
The only "variable" here is the UAV's turn radius, (but it does not need to be entered manually) 
- if for example the grid spacing is 50meters, and a big UAV of 6m wingspan requires 150m to turn, - then the operator already knows that - and adds another >100 to the area being mapped.

In related experience; when we are using bigger machines with bigger turn radius, then we are usually flying higher anyway (have wider space between grid lines.)

I think we all always draw bigger plans than the actual area that have been ordered, that too solves the "problem" with bigger or longer turns (usually due to strong wind - in cases where turning downwind cannot be avoided). 

On the second hand - it could be nice to let the user adjust the "extension" of WP 4,5 and so on , it's not necessary to do it for both.  Maybe WP4 type of wp could be defined as "extend by 20m"  - then WP5-type would move relatively to that. ?

[...]

PR #3970, opened 2016-08-23
Turnaround waypoints for fixed-wing surveying

birchera:
This is an initial solution to https://github.com/mavlink/qgroundcontrol/issues/3801.

By setting the turnaround distance to >0 additional waypoints are inserted. For now they are simply offset from the polygon edge by the given distance.

For turnaround distance <=0 everything remains as before.

screenshot from 2016-08-23 15 20 26

We will test the feature during flights this week. If we find we have to adapt the the shape of the offset, this patch should allow easy prototyping.

### DIFF
--- a/src/MissionManager/SurveyMissionItem.cc
+++ b/src/MissionManager/SurveyMissionItem.cc
@@ -25,6 +25,7 @@ const char* SurveyMissionItem::_jsonGridAltitudeKey =          "gridAltitude";
 const char* SurveyMissionItem::_jsonGridAltitudeRelativeKey =  "gridAltitudeRelative";
 const char* SurveyMissionItem::_jsonGridAngleKey =             "gridAngle";
 const char* SurveyMissionItem::_jsonGridSpacingKey =           "gridSpacing";
+const char* SurveyMissionItem::_jsonTurnaroundDistKey =       "turnaroundDist";
 const char* SurveyMissionItem::_jsonCameraTriggerKey =         "cameraTrigger";
 const char* SurveyMissionItem::_jsonCameraTriggerDistanceKey = "cameraTriggerDistance";
 
@@ -43,6 +44,7 @@ SurveyMissionItem::SurveyMissionItem(Vehicle* vehicle, QObject* parent)
     , _gridAltitudeFact         (0, "Altitude:",                FactMetaData::valueTypeDouble)
     , _gridAngleFact            (0, "Grid angle:",              FactMetaData::valueTypeDouble)
     , _gridSpacingFact          (0, "Grid spacing:",            FactMetaData::valueTypeDouble)
+    , _turnaroundDistFact       (0, "Turnaround dist.:",        FactMetaData::valueTypeDouble)
     , _cameraTriggerDistanceFact(0, "Camera trigger distance",  FactMetaData::valueTypeDouble)
 
     , _gridAltitudeMetaData         (FactMetaData::valueTypeDouble)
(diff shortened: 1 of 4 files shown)


NaterGator:
@birchera for turnaround behavior you may find it more beneficial to "lead in" the subsequent leg rather than "lead out" the previous leg. Doing so gives your craft a greater chance of having a settled attitude as it starts the cross track navigation relevant for taking samples in the ROI.

src/MissionManager/SurveyMissionItem.cc
```diff
@@ -25,6 +25,7 @@ const char* SurveyMissionItem::_jsonGridAltitudeKey =          "gridAltitude";
 const char* SurveyMissionItem::_jsonGridAltitudeRelativeKey =  "gridAltitudeRelative";
 const char* SurveyMissionItem::_jsonGridAngleKey =             "gridAngle";
 const char* SurveyMissionItem::_jsonGridSpacingKey =           "gridSpacing";
+const char* SurveyMissionItem::_jsonTurnaroundDistKey =       "turnaroundDist";
```
NaterGator: 1 space missing here

birchera:
Thanks for the review. I agree with importance of the leading in. Our idea was that you need both in and out, because you also want to avoid attitude changes when reaching the acceptance radius but not the edge of the polygon.

For the leading in I am not sure whether this really fixes it robustly. But if you chose the distance large enough it certainly will.

DonLakeFlyer:
Can you update the unit test to test this as well? Should be able to test the correct addition of extra points when turn around distance is set.

birchera:
@DonLakeFlyer can you check if the implemented unit test is what you meant?

DonLakeFlyer:
Will do. I should be able to get to this by tonight.

DonLakeFlyer:
I thought the idea was to has a short lead out, and a longer lead in as described here: https://github.com/mavlink/qgroundcontrol/issues/3801

birchera:
Short lead out, and a longer lead in certainly could make sense for large grid spacing. When this is not the case compared to the turn radius, I see equal distance as a better option. I hope the following image can visualize this best:



Note the larger overshoot to the right side of the unequal lead out and in distance.

If you still think we should have two parameters for lead in and lead out, I can easily add this.

src/MissionManager/SurveyMissionItem.cc
```diff
@@ -276,6 +281,7 @@ bool SurveyMissionItem::load(const QJsonObject& complexObject, QString& errorStr
     _gridAltitudeFact.setRawValue           (complexObject[_jsonGridAltitudeKey].toDouble());
     _gridAngleFact.setRawValue              (complexObject[_jsonGridAngleKey].toDouble());
     _gridSpacingFact.setRawValue            (complexObject[_jsonGridSpacingKey].toDouble());
+    _turnaroundDistFact.setRawValue         (complexObject[_jsonTurnaroundDistKey].toDouble());
```
DonLakeFlyer: How does Qt react when this key is missing from the json? Does it pop any qWarnings? Does it return 0.0 for the value?

DonLakeFlyer:
Ok, got it. Can you add FactMetaData for the new Fact so we can get units, range validation, decimal places?

src/MissionManager/SurveyMissionItem.cc
kellysteich: No qWarnings and yes it just returns 0.0 for the value.

kellysteich:
FactMetaData for the new turnaroundDistFact has been added.

birchera:
Thanks Kelly, must have missed that.

PR #3978, opened 2016-08-24
Bump timeout to get log download to work on Solo over WiFi

DonLakeFlyer:
Solo seems to be slow in responding to requests. This was causing timeout to fire and then request re-send. The Solo eventually responded to both requests which then cause the download code to get very confused.

### DIFF
--- a/src/ViewWidgets/LogDownloadController.cc
+++ b/src/ViewWidgets/LogDownloadController.cc
@@ -25,7 +25,7 @@
 #include <QBitArray>
 #include <QtCore/qmath.h>
 
-#define kTimeOutMilliseconds 500
+#define kTimeOutMilliseconds 2000
 #define kGUIRateMilliseconds 17
 #define kTableBins           512
 #define kChunkSize           (kTableBins * MAVLINK_MSG_LOG_DATA_FIELD_DATA_LEN)
@@ -122,6 +122,7 @@ LogDownloadController::LogDownloadController(void)
 void
 LogDownloadController::_processDownload()
 {
+    qCDebug(LogDownloadLog) << "Timeout";
     if(_requestingLogEntries) {
         _findMissingEntries();
     } else if(_downloadingLogs) {
@@ -297,6 +298,9 @@ LogDownloadController::_findMissingEntries()
 void
 LogDownloadController::_logData(UASInterface* uas, uint32_t ofs, uint16_t id, uint8_t count, const uint8_t* data)
 {
+    // Useful for debugging but noisy
+    //qCDebug(LogDownloadLog) << "_logData id:ofs:count" << id << ofs << count;
+
     if(!_uas || uas != _uas || !_downloadData) {
         return;
     }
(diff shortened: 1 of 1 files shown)


src/ViewWidgets/LogDownloadController.cc
```diff
@@ -361,11 +365,13 @@ LogDownloadController::_logData(UASInterface* uas, uint32_t ofs, uint16_t id, ui
                 //-- Check for more
                 _receivedAllData();
             } else if (_chunkComplete()) {
+                qCDebug(LogDownloadLog) << "Chunk complete";
                 _downloadData->advanceChunk();
                 _requestLogData(_downloadData->ID,
                                 _downloadData->current_chunk*kChunkSize,
                                 _downloadData->chunk_table.size()*MAVLINK_MSG_LOG_DATA_FIELD_DATA_LEN);
             } else if (bin < _downloadData->chunk_table.size() - 1 && _downloadData->chunk_table.at(bin+1)) {
+                qCDebug(LogDownloadLog) << "Missing data";
```
NaterGator: What is meant by this message? This scope is actually when a fragment of previously missing data has been successfully redownloaded.

NaterGator:
Can we scale this dynamically based on mean response time? 2 seconds is going to be agonizingly slow on systems that are actually fast.

DonLakeFlyer:
This is an error timeout. Things will only timeout when something goes wrong. Like the request message didn't make it to the vehicle, so the vehicle did not respond. It has no affect on download speed. It only slows down telling the user if things didn't work correctly.

NaterGator:
Not exactly. The timer is [bound](https://github.com/DonLakeFlyer/qgroundcontrol/blob/734ddffc4c441509bcadee322c8b5af6d7525901/src/ViewWidgets/LogDownloadController.cc#L117) to the _processDownload function which calls [_findMissingData](https://github.com/DonLakeFlyer/qgroundcontrol/blob/734ddffc4c441509bcadee322c8b5af6d7525901/src/ViewWidgets/LogDownloadController.cc#L421) for the cases where a packet gets dropped.

In practice packets still drop frequently. Even over USB. Increasing this timeout is going to slow things down.

DonLakeFlyer:
That would be a boundary case I believe. It would have to lose the last packet of a chunk request in order to hit that. Otherwise it would realize it got the last bit of a chunk and the call _findMissingData without needing a timeout.

DonLakeFlyer:
I'll let this sit for now in case there is some better way while I'm trying to figure out what is wrong with the main download code.

NaterGator:
Some thoughts before I sit down for dinner:

I would seed the initial latency estimate very high, 2 seconds is appropriate. Start a QElapsedTimer each time we get [here](https://github.com/DonLakeFlyer/qgroundcontrol/blob/734ddffc4c441509bcadee322c8b5af6d7525901/src/ViewWidgets/LogDownloadController.cc#L473) and check for the beginning of a new chunk around [here](https://github.com/DonLakeFlyer/qgroundcontrol/blob/734ddffc4c441509bcadee322c8b5af6d7525901/src/ViewWidgets/LogDownloadController.cc#L322). Use a simple IIR filter to update the latency (since there is a unique 1:1 correlation between explicit request for the start of each long chunk) and set the new timeout to something like 5x the latency measurement. 


I get what you're saying about the uniqueness of the packet that has to drop, but in the case of filling gaps/fragments it can be as small as 1 bin and then a single drop causes the timeout. It isn't the majority of the download but I'm seeing this in practice still over USB.

DonLakeFlyer:
The issue is that it needs two timeouts. A longer one for no response at all to a data request and a shorter one for the stream of data stopping with missing packets along the way.

DonLakeFlyer:
I'll figure this out with my other work on log download

PR #3981, opened 2016-08-26
Camera survey enhancement with additional parameters

kellysteich:
This extends the camera survey feature by:
* adding parameters like ground sampling distance (gsd) and frontal/side overlap
* allowing the calculation of gsd and overlap from altitude and grid spacing or the reverse. This allows you to set up a mission either from desired height and gridspacing or desired gsd and overlap
* adding more camera parameters, like image size (needed for enhanced calculations) and moving their edit fields to a popout to save space
* adding a way to add default camera models and choosing between them from a dropdown menu

### DIFF
--- a/src/MissionEditor/SurveyItemEditor.qml
+++ b/src/MissionEditor/SurveyItemEditor.qml
@@ -1,5 +1,6 @@
 import QtQuick                  2.2
 import QtQuick.Controls         1.2
+import QtQuick.Dialogs          1.2
 
 import QGroundControl               1.0
 import QGroundControl.ScreenTools   1.0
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Could I get some screenshots from this?

kellysteich:
Sure, here they are:

screenshot from 2016-08-26 14-53-54  |  screenshot from 2016-08-26 14-54-07

The left one shows the survey item with additional parameters and the right one the popup that opens when clicking on "Change", allowing you to choose a different camera model.

DonLakeFlyer:
Could there just be a camera combo box on the main survey editor page? Then no need for the dialog. For editable camera list I think that will eventually go into Settings.

kellysteich:
The camera combo box could be there, but then all the editor fields for sensor width/height etc. would also have to be moved there, which makes the survey editor page extremely long. In my opinion it just becomes too long then and usability is reduced, especially on mobile devices. Even as it is now, it just fits on the screen of our tablet, so that you can edit f.ex grid spacing and still see # shots at the same time.
 
Or else you would not be able to enter custom camera values, which I feel would not be very user friendly, especially with the very few camera models currently available.

Which solution did you have in mind? I can still change it though if you feel that that is still the best way to go about it.

DonLakeFlyer:
Ok, let's leave it like this for now and I'll look at restructuring the ui since it's getting to big onthe first screen.

DonLakeFlyer:
@kellysteich Can you rebase, then it will go back through CI and I'll merge once passing.

birchera:
Rebase done.

jcooper21:
I think one key feature was missed here and that is the ability to rotate the survey grid so we can ensure we are always flying cross wind with a fixed wing.

jcooper21:
The other field is mission speed that I cannot seem to figure out how to set.

DonLakeFlyer:
> I think one key feature was missed here and that is the ability to rotate the survey grid so we can ensure we are always flying cross wind with a fixed wing.

That is Grid Angle.

DonLakeFlyer:
> The other field is mission speed that I cannot seem to figure out how to set.

Mission Speed is a vehicle parameter. You should be able to find it in the parameter list. Differs based on firmware. It is not pulled out to Mission Planning yet.

jcooper21:
Cant believe I missed the grid angle setting.  Thanks Don!

Issue #3982, opened 2016-08-27
UVC Video support


cda725:
In an effort to expand the capabilities of Qgroundcontrol I am attempting to add UVC video support.

The purpose for this is to use the system as  ground control station and portable ground control station supporting the myriad of wireless video transmitter/receiver systems that are ubiquitous on remote controlled vehicles today. Most are composite or HDMI on the receiver and UVC devices are available with various input options to meet the users needs.

In addition, UVC does not have the processor demand that devices such as the EZCAP and similar devices do nor does it require drivers. Regardless of OS is just "works" wit whatever system you plug it into.

Most DIY (and many commercial) systems make use of the following components;

1. A remote for direct human control
2. A wireless video link (in varying frequencies) to send the video feed from a camera to a receiver
3. Optionally, often  a tablet or portable PC to run software to display telemetry data such as a map overlay, GPS position, speed and so on. On most commercial systems, in addition to the map data and other telemetry information, the APP usually allows a view of the camera feed for FPV operation.

DJI is a great example of a system that uses a remote in conjunction with a tablet. You can use the tablet to display your location on a map overlay, get data about the vehicle, set waypoints and such and you can view the video feed. This simplifies the overall setup and allows a lightweight portable ground control station.

The DJI system is of course proprietary but basically works the same as a DIY model in that there is a transmitter, receiver and computing device on the ground side. 

Qgroundcontrol is a very nice system and would allow someone to easily clone the commercial systems on a budget. Adding UVC video support would allow several configurations. I will give an example of one below:

Assuming a UAV ( quad drone) build for our example the following would be installed:

1. Gimbal(s) 
2. Camera(s)
3. Wireless telemetry module
4. Pixhawk
5. Wireless video transmitter (which are small, light, powerful and will most will run from 3S to 7S
6. Remote receiver such as an X8R

On the user end the following in our example is used:

1. Taranis X9D remote
2. Android tablet
3. USB hub
4 UVC capture device for the video receiver
5. USB Wireless telemetry receiver.
6. Video receiver with composite or HDMI out

I have tested the USB wireless telemetry modules and can fly several miles with my remote and Qgroundcontrol running from my Android and Windows tablets.  I currently have to run a separate monitor to view my cameras. If one did not need the separate monitor for this configuration they could mount the components to the back of the tablet stand that is attached to the remote and power everything from the Taranis (or whichever brand remote they choose).

[...]

PR #3983, opened 2016-08-28
Video Source Options

dogmaphobic:
Based on discussion #3982
Adding code to allow the user to specify a USB video source. This uses the QML *Camera* item to display video. Any source that shows itself to the OS as a Camera can be used.
http://doc.qt.io/qt-5/qml-qtmultimedia-camera.html

* Renamed `FlightDisplayViewController` to `VideoManager`
* Made `VideoManager` into a singleton (accessible through `QGroundControlQmlGlobal`)
* Addes a combobox to the General Settings allowing the user to chose a video source.

Known Issue:

If you have more than one USB device, for some reason the QML code is not respecting the change. That is, it keeps showing the first used USB *camera*. Switch from USB to UDP Streaming is working fine. UDP Streaming is the default and it did not change.

This is still experimental. I don't have any USB Video Capture device to test with. I tested with a USB camera and it worked. On mobile, the UI shows both front and back cameras but switching between them is not working (discussed above). It requires a restart.


cda725:
Bug being caused by video pipeline not being closed before selecting a new one. I am attempting to patch the bug now.

Notes: This addition allows for a broad range of USB video devices - Good work my friend!

cda725:
I Fixed the bug but am a noob to github. The Fix for selecting video is as follows:

Starting at line 105 of VideoManager.cc Pasts this...

//-----------------------------------------------------------------------------
void
VideoManager::setVideoSource(QString vSource)
{
    _videoSource = vSource;
    QSettings settings;
    settings.setValue(kVideoSourceKey, vSource);
    emit videoSourceChanged();
    QList<QCameraInfo> cameras = QCameraInfo::availableCameras();
    foreach (const QCameraInfo &cameraInfo, cameras) {
        if(cameraInfo.description() == vSource) {
            _videoSourceID = cameraInfo.deviceName();
            VideoManager::disconnect();

            emit videoSourceIDChanged();
            qCDebug(VideoManagerLog) << "Found USB source:" << _videoSourceID << " Name:" << _videoSource;
            break;
        }
    }
    emit isGStreamerChanged();
    qCDebug(VideoManagerLog) << "New Video Source:" << vSource;
}

dogmaphobic:
@cda725 I fixed the issue within the QML code. See #3986. It handles both turning off the camera when not in use and stop/start the camera during changes.

GlidLov:
Hi guys,

i think the bug is still there: i'm trying a UVC device with my win10 laptop, i can see image thru VLC, then when i go into QGC, i select the device, and it shows my front webcam instead of the UVC video stream.

news about it?

DonLakeFlyer:
Can you put in a new Issue for that? Also make sure you are testing with a daily build

Issue #3985, opened 2016-08-29
QGC - OSX activates always FaceTime camera


lvale:
Hi

Even if the Facetime camera is not selected the camera is active (Green LED). With all the privacy mania, I believe the Facetime camera should only be active if explicitly selected

<img width="476" alt="screen shot 2016-08-29 at 02 25 02" src="https://cloud.githubusercontent.com/assets/3130363/18038441/1e64ef66-6d90-11e6-8c73-281ba650b0ad.png">

![img_20160829_022110](https://cloud.githubusercontent.com/assets/3130363/18038445/24ab333a-6d90-11e6-9fde-a967e0520979.jpg)

dogmaphobic:
Yes, the camera should be "stopped" when not using. Still working on it...

lvale:
btw: love the camera features on the Android app :) Works great with OnePlus ;)

lvale:
@dogmaphobic @DonLakeFlyer As a bonus feature of this camera issue a record video stream function would be nice. Next would be to have data overlayed on the video, and recordable :)

dogmaphobic:
> As a bonus feature of this camera issue a record video stream

Record the stream is not simple either but not impossible either. Recording the composite (video and overlay) is not simple at all. On desktop, you're better off using QuickTime or something like it. On mobile, no clue. I like the idea though...

dogmaphobic:
Fixed in #3986

PR #3986, opened 2016-08-29
Turn camera off when not in use.

dogmaphobic:
In response to #3985.


dogmaphobic:
While at it, found the bug that prevented a switch between multiple UVC devices.

PR #3989, opened 2016-08-30
Initial GeoFence implementation

DonLakeFlyer:
Still very crude at this point. Major work in progress. 

screen shot 2016-08-29 at 7 27 34 pm

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -273,12 +273,16 @@ HEADERS += \
     src/LogCompressor.h \
     src/MG.h \
     src/MissionManager/ComplexMissionItem.h \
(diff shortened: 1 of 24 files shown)


DonLakeFlyer:
Back end support is ArduPilot only at this point. But once it's plugin-ized whatever PX4 support shows up can be plumbed through as well.

mayanez:
Hey,

I'm currently working on a geofence database that automatically updates from FAA and other sources. Would this be a feature that you'd be interested in adding?

DonLakeFlyer:
@mayanez That's not quite how GeoFence currently. GeoFence keeps the vehicle inside an area like a polygon. It's doesn't prevent it from going into an area likea no-fly zone.

PR #3991, opened 2016-08-30
RTSP Video, Android x86 and other fixes.

dogmaphobic:
* Added support for Android x86 (including GStreamer)
 * Added support for RTSP Video Streams
  * Switching between UDP and RTSP, or changing the options for either one requires restarting the application. I still need to see what is not being reset for a seamless switch to occur.
 * Made a bit of a reorg in the General Settings panel. The idea is to make all panels consistent.
 * Added build flag to skip UVC support if desired (QGC_DISABLE_UVC)
 * Made Camera view to fit screen, cropping if necessary (as opposed to stretch it)
 * Fixed crash bug on Linux when no UVC device was available
  * Moved QML Camera item into a loader, which is only loaded if UVC is available (it crashes if it is not)

### DIFF
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -30,12 +30,18 @@ linux {
         CONFIG += LinuxBuild
         DEFINES += __STDC_LIMIT_MACROS __rasp_pi2__
     } else : android-g++ {
-        message("Android build")
         CONFIG += AndroidBuild MobileBuild
         DEFINES += __android__
         DEFINES += __STDC_LIMIT_MACROS
         DEFINES += QGC_ENABLE_BLUETOOTH
         target.path = $$DESTDIR
+        equals(ANDROID_TARGET_ARCH, x86)  {
+            CONFIG += Androidx86Build
+            DEFINES += __androidx86__
+            message("Android x86 build")
+        } else {
+            message("Android Arm build")
+        }
     } else {
         error("Unsuported Linux toolchain, only GCC 32- or 64-bit is supported")
     }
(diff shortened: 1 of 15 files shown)


dogmaphobic:
CI for iOS is not using GStreamer. No big deal. At least it provides a test for "non GStreamer" builds.

NaterGator:
For modifying the pipeline  without restarting are you setting it to NULL state and ensuring it finishes all transitions before making changes?

dogmaphobic:
I didn't restart it. It finished with one failure (iOS). I fixed the issues and submitted a new commit. Then the entire thing run again as it should.

dogmaphobic:
Duh, and here I am thinking you were talking about the build. Yes, I'm going through a full delete, which shuts it all down, and creating all new instances again. Something is still not being released.

dagar:
I need to look at getting both arm and x86 binaries on google play.

NaterGator:
@dogmaphobic that's not necessarily true in gstreamer's async / threaded world. If VideoReceiver::stop's call to [gst_element_set_state](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstElement.html#gst-element-set-state) returns GST_STATE_CHANGE_ASYNC the pipeline is unlikely to be stopped by the time you get around to creating the new one. In that case we should call [gst_element_get_state](https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstElement.html#gst-element-get-state) to ensure that the pipeline is fully null'd before we proceed.

dogmaphobic:
RTSP is still work in progress for Android. The "official" build links the needed libraries statically. I'm still collecting the needed libraries so I can update the build.

PR #3992, opened 2016-08-30
HIL add support for aux controls

bartslinger:
The HIL_CONTROLS message in mavlink has four aux values which are not used. This PR adds support for aux1-4 so that they can be used for future developments.


DonLakeFlyer:
I'd rather not have unused code paths. Since this code doesn't do anything right now it just adds noise to the codebase. Happy to take a pull for fully plumbed aux support.

bartslinger:
I'll be back ;-)

DonLakeFlyer:
Great, thanks Bart.

PR #3996, opened 2016-08-30
Plugin-ize geofence for multiple firmware support

DonLakeFlyer:
This should be in a working state for ArduCopter and ArduPlane now.


DonLakeFlyer:
@lvale This should take it to a workable state if you want to play around with it.

DonLakeFlyer:
UI is still crappy but slightly better.

lvale:
Thanks @DonLakeFlyer I'll report back :)

DonLakeFlyer:
Sorry @lvale this pull is wrong. It will work for Plane but won't work for Copter. Closing to fix it.

PR #4000, opened 2016-08-31
Adding missing library to Linux build.

dogmaphobic:
I had added `libQt5Multimedia.so.5` but I missed `libQt5MultimediaQuick_p.so.5`. It will be a trial-and-error until we get it all there as I don't have a way to fully test what libraries are needed. Let's see if this is all it takes.


dogmaphobic:
4,000!

Issue #3999, opened 2016-08-31
QGC complains that firmware v1.5.x is older than 1.4.1


kd0aij:
see https://github.com/PX4/Firmware/issues/5419

DonLakeFlyer:
Firmware problem: https://github.com/PX4/Firmware/issues/5308

PR #4001, opened 2016-08-31
Compare version numbers using the generator function from firmware

NaterGator:
This is for #3999 (and others). Pull req for FW fix is here: https://github.com/PX4/Firmware/pull/5422

This uses the same version generator from firmware to build the comparison value.


DonLakeFlyer:
Huh? autopilot_version mavlink message is also handled in Vehicle to show on summary page. That is going to be wrong as well and that is generic mavlink support. Don't understand why this would go into QGC. Either the mavlink message sends correct data or the firmware needs to be fixed so that it sends correct data.

NaterGator:
Both are happening. I figured importing the version generator and specifying versions a a char array: "v1.4.1" was a better reality check than what we had. Alternatively we can just use a macro (or new fancy C++ constexpr) to generate the firmware version literal directly as an uint32_t. 

I wouldn't have caught the bug if I had not gone through this process though.

DonLakeFlyer:
I still don't see why this is needed in QGC. Also the less firmware specific code in QGC the better. So I'd rather not have this in there. Is there some reason it is needed beyond the generic mavlink code?

DonLakeFlyer:
Closing this since I think we're ok without it.

PR #4009, opened 2016-09-01
PX4 fence plugin

DonLakeFlyer:
* Implemented PX4 GeoFence plugin
* Changed Plan Element (Mission, GeoFence) ui selection mechanism
screen shot 2016-09-01 at 11 24 07 am


DonLakeFlyer:
@LorenzMeier PX4 GeoFence editing is now plumbed fully through QGC. If you give me an api to get/set polygon fence points it would take me only a couple hours to support on the QGC side since polygon fence is already supported by QGC plugin code.

Issue #3815, opened 2016-07-17
Finish GStreamer readme


DonLakeFlyer:
Windows and iOS are listing as TODO.

dogmaphobic:
Done in #4012

PR #4012, opened 2016-09-02
Update README.md [ci skip]

dogmaphobic:
Closing #3815


dogmaphobic:
Crap, I keep putting [ci skip] in the wrong place...

PR #4013, opened 2016-09-03
pull proper gstreamer xcopy from master

dagar:


DonLakeFlyer:
Ah, forgot something. The windows install needs a full install of gstreamer to work. That should be in the appveyor.qml on the daily side I believe.

PR #4025, opened 2016-09-06
Add support for HIL_ACTUATOR_CONTROLS with X-Plane HITL

bartslinger:
This supports the new mavlink message HIL_ACTUATOR_CONTROLS for use with X-Plane HITL. The coolest feature is that it can now control individual actuators in X-Plane, instead of emulating the joystick. The old message HIL_CONTROLS is still working. A checkbox is added in the ui to switch between the two.

I also ran fix_code_style on the files but apparently a lot was still wrong. To get a better idea of the changes I made, only look at the first commit.

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -429,6 +429,30 @@ void UAS::receiveMessage(mavlink_message_t message)
             emit hilControlsChanged(hil.time_usec, hil.roll_ailerons, hil.pitch_elevator, hil.yaw_rudder, hil.throttle, hil.mode, hil.nav_mode);
         }
             break;
+        case MAVLINK_MSG_ID_HIL_ACTUATOR_CONTROLS:
+        {
+            mavlink_hil_actuator_controls_t hil;
+            mavlink_msg_hil_actuator_controls_decode(&message, &hil);
+            emit hilActuatorControlsChanged(hil.time_usec, hil.flags,
+                                            hil.controls[0],
+                                            hil.controls[1],
+                                            hil.controls[2],
+                                            hil.controls[3],
+                                            hil.controls[4],
+                                            hil.controls[5],
+                                            hil.controls[6],
+                                            hil.controls[7],
+                                            hil.controls[8],
+                                            hil.controls[9],
+                                            hil.controls[10],
+                                            hil.controls[11],
+                                            hil.controls[12],
+                                            hil.controls[13],
+                                            hil.controls[14],
+                                            hil.controls[15],
+                                            hil.mode);
+        }
+            break;
         case MAVLINK_MSG_ID_VFR_HUD:
         {
             mavlink_vfr_hud_t hud;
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
Thanks @bartslinger. I'm out of town till next Tuesday. I'll take a look when I get back.

DonLakeFlyer:
@bartslinger Starting to look through this. Can you back out the code style changes. They don't match the QGC coding style:
* https://github.com/mavlink/qgroundcontrol/blob/master/CodingStyle.h
* https://github.com/mavlink/qgroundcontrol/blob/master/CodingStyle.cc

Also your mavlink submodule is out of date so you are incorrectly pushing a mavlink update. Can you bring your mavlink up to date and fix the pull?

src/uas/UAS.cc
```diff
@@ -429,6 +429,30 @@ void UAS::receiveMessage(mavlink_message_t message)
             emit hilControlsChanged(hil.time_usec, hil.roll_ailerons, hil.pitch_elevator, hil.yaw_rudder, hil.throttle, hil.mode, hil.nav_mode);
         }
             break;
+        case MAVLINK_MSG_ID_HIL_ACTUATOR_CONTROLS:
```
DonLakeFlyer: Can you put this into Vehicle.h/cc instead? UAS.cc/h is being deprecated.

bartslinger:
Will do this in the weekend

bartslinger:
Should be good to go now

libs/mavlink/include/mavlink/v1.0
```diff
@@ -1 +1 @@
-Subproject commit 28ea42ccb6ab1d764cacf3f0f55358dc3c66faa9
+Subproject commit e93ac62981a338a7c823364e7c4ff1077e3f8fc1
```
DonLakeFlyer: Old Mavlink submodule is still in the pull

bartslinger:
I don't get it. If I don't include the mavlink update, it does not have the new HIL_ACTUATOR_CONTROLS message and fails to build on my local machine at least. Lets see what appveyor/travis does.

DonLakeFlyer:
Oh, wait crap! I didn't realize you meant to include a mavlink update. I thought you added that by mistake which most people do. Duh! This is fine then.

bartslinger:
No problem, re-reverted it :-p

bartslinger:
The AppVeyor fail does not make sense to me.

NaterGator:
AppVeyor failure was caused by https://github.com/appveyor/ci/issues/1062

DonLakeFlyer:
@dagar Do you see any way to work around the appveyor bug? For now I'm going to merge without appveyor making it all the way through.

dagar:
@DonLakeFlyer we can just use the path for now. https://github.com/mavlink/qgroundcontrol/pull/4071

dagar:
@DonLakeFlyer we can just use the path for now. #4071

PR #4026, opened 2016-09-06
Add StreamingSource Fact

ricardotk:
Since this is a configuration option, it will become clearer this way.

### DIFF
--- a/src/QmlControls/QGroundControlQmlGlobal.cc
+++ b/src/QmlControls/QGroundControlQmlGlobal.cc
@@ -33,6 +33,8 @@ SettingsFact* QGroundControlQmlGlobal::_speedUnitsFact =                        
 FactMetaData* QGroundControlQmlGlobal::_speedUnitsMetaData =                        NULL;
 SettingsFact* QGroundControlQmlGlobal::_batteryPercentRemainingAnnounceFact =       NULL;
 FactMetaData* QGroundControlQmlGlobal::_batteryPercentRemainingAnnounceMetaData =   NULL;
+SettingsFact* QGroundControlQmlGlobal::_streamingSourcesFact =                      NULL;
+FactMetaData* QGroundControlQmlGlobal::_streamingSourcesMetaData =                  NULL;
 
 const char* QGroundControlQmlGlobal::_virtualTabletJoystickKey  = "VirtualTabletJoystick";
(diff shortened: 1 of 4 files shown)


dogmaphobic:
Don't merge this yet. It removes the UVC video option. Now you only have UDP and RTSP. UVC needs to be added back in.

dogmaphobic:
Also, I was under the impression *Facts* corresponded to parameters originating from the flight controller. Is it being used arbitrarily (I don't see an issue with it per se)? Or is it truly expected for the flight controller to provide this *Fact*?

ricardotk:
> Don't merge this yet. It removes the UVC video option. Now you only have UDP and RTSP. UVC needs to be added back in.

Mmm, in branch master only two options appear for me: UDP and RTSP. But anyway, should I add UVC as a Fact here:?

ricardotk:
> Also, I was under the impression Facts corresponded to parameters originating from the flight controller. Is it being used arbitrarily (I don't see an issue with it per se)? Or is it truly expected for the flight controller to provide this Fact?

Its being used arbitrarily, its not provided by a flight controller

dogmaphobic:
> Mmm, in branch master only two options appear for me: UDP and RTSP. But anyway, should I add UVC as a Fact here:?

Yes, it only shows on the list when you have an UVC device connected. There is no *Fact*. It's a hardware query. This is where the ComboBox *model* is built:

https://github.com/mavlink/qgroundcontrol/blob/master/src/FlightDisplay/VideoManager.cc#L179

You added a new *model* (from your *Facts*) but did nothing about the existing code.

dogmaphobic:
Basically there are two very different types of video source. One is digitally streamed from the vehicle and we use GStreamer to handle it. The other is some video capture device physically connected to the device QGC is running on. Usually an analog to digital video converter that takes an analog video input (coming from an analog video receiver) and piping that video through USB.

ricardotk:
Mmm now that you showed me how the UVC part is implemented, I not sure if this patch is useful anymore. I would have to create a list containing the Facts and the cameras from UVC right?

dogmaphobic:
It's not just that. How would you enter a manually configured video source (UDP or RTSP), which has no parameter coming from anywhere? As it is, you can only pick a video source for which a stream definition was provided through a parameter. That implies it only supports vehicles/systems which advertise their video capabilities in a standard fashion.

PR #4027, opened 2016-09-06
[DO NOT MERGE] WIP - Add RTSP video streaming widget

ricardotk:
The idea of this PR is to give you guys a brief of the solution we're planning to video streaming, which includes some patches in qgroundcontrol as well.

This commit is still work in progress but adds an RTSP video streaming widget which will be used to set RTSP streaming configurations such as query streams URIs, formats, capabilities, framesizes and so on.

In order to control these configurations we need the support for a custom MAVLink message which is still being reviewed by the Ardupilot team.

http://discuss.ardupilot.org/t/review-camera-streaming-custom-mavlink-message/11132/5

I would like to ask you guys for a review of what you think of it as well as review the commit itself.

Thanks!

This commit is above #4026, which can be merged to upstream.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -414,6 +414,8 @@ HEADERS += \
     src/ViewWidgets/CustomCommandWidgetController.h \
     src/ViewWidgets/LogDownload.h \
     src/ViewWidgets/LogDownloadController.h \
+    src/ViewWidgets/VideoStreamingWidget.h \
+    src/ViewWidgets/VideoStreamingWidgetController.h \
     src/ViewWidgets/ViewWidgetController.h \
 }
 
@@ -559,6 +561,8 @@ SOURCES += \
     src/ViewWidgets/CustomCommandWidgetController.cc \
     src/ViewWidgets/LogDownload.cc \
     src/ViewWidgets/LogDownloadController.cc \
+    src/ViewWidgets/VideoStreamingWidget.cc \
+    src/ViewWidgets/VideoStreamingWidgetController.cc \
     src/ViewWidgets/ViewWidgetController.cc
 }
 
@@ -809,6 +813,7 @@ INCLUDEPATH += \
 HEADERS += \
     src/VideoStreaming/VideoItem.h \
     src/VideoStreaming/VideoReceiver.h \
+    src/VideoStreaming/VideoUtils.h \
     src/VideoStreaming/VideoStreaming.h \
     src/VideoStreaming/VideoSurface.h \
     src/VideoStreaming/VideoSurface_p.h \
@@ -816,6 +821,7 @@ HEADERS += \
 SOURCES += \
     src/VideoStreaming/VideoItem.cc \
     src/VideoStreaming/VideoReceiver.cc \
+    src/VideoStreaming/VideoUtils.cc \
     src/VideoStreaming/VideoStreaming.cc \
     src/VideoStreaming/VideoSurface.cc \
 
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -12,6 +12,7 @@
         <file alias="APMAirframeComponent.qml">src/AutoPilotPlugins/APM/APMAirframeComponent.qml</file>
         <file alias="APMAirframeComponentSummary.qml">src/AutoPilotPlugins/APM/APMAirframeComponentSummary.qml</file>
         <file alias="CustomCommandWidget.qml">src/ViewWidgets/CustomCommandWidget.qml</file>
+        <file alias="VideoStreamingWidget.qml">src/ViewWidgets/VideoStreamingWidget.qml</file>
         <file alias="LogDownload.qml">src/ViewWidgets/LogDownload.qml</file>
         <file alias="FirmwareUpgrade.qml">src/VehicleSetup/FirmwareUpgrade.qml</file>
         <file alias="FlightDisplayView.qml">src/FlightDisplay/FlightDisplayView.qml</file>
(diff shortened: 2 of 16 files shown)


dogmaphobic:
This takes what I've been doing to a whole new level. I was trying to come up with a better way to allow arbitrary pipelines but got stuck with the problem of having to link *all* of GStreamer in order to support any combination of options.

It's a great way forward but it's using a dead-end method. Those widgets are not going to stay around for long. They are legacy (desktop) code and will have to be replaced. Widgets don't work on mobile platforms for instance.

It also assumes the stream information is coming from the flight controller. Though it's perfect when it does, there needs to be a way to configure it all without any information from the flight controller. An example is the 3DR Solo. We support by configuring the stream ourselves. The Solo does not send us any stream configuration.

This branch has the same (inherited) issue as the one in #4026. It kills the UVC video option.

ricardotk:
> This takes what I've been doing to a whole new level. I was trying to come up with a better way to allow arbitrary pipelines but got stuck with the problem of having to link all of GStreamer in order to support any combination of options.

Thats good to know, maybe we can work together on this after all.

> It's a great way forward but it's using a dead-end method. Those widgets are not going to stay around for long. They are legacy (desktop) code and will have to be replaced. Widgets don't work on mobile platforms for instance.

Oh, what would you recommend then?

> It also assumes the stream information is coming from the flight controller. Though it's perfect when it does, there needs to be a way to configure it all without any information from the flight controller. An example is the 3DR Solo. We support it by configuring the stream ourselves. The Solo does not send us any stream configuration.

I agree, nice catch

lucasdemarchi:
> It also assumes the stream information is coming from the flight controller. Though it's perfect when it does, there needs to be a way to configure it all without any information from the flight controller. An example is the 3DR Solo. We support it by configuring the stream ourselves. The Solo does not send us any stream configuration.

Humn... no. The idea is you have an external process running on the drone that can tell the GCS what are the protocols and everything it needs to consume the video stream. The intention (if we are not there yet) is to be plug and play: they negotiate the parameters and then have the video stream.  We can allow "manual configuration", but ideally the drone would be smart enough to inform the GCS how to setup its end of the connection.

For Solo, it would mean to run the additional daemon there and everything should work. If it doesn't work it's just because it's still a work in progress.

dogmaphobic:
> Thats good to know, maybe we can work together on this after all.

That would be great, especially given that this is the first time I hear something is being done to something I'm actively working on.

> Oh, what would you recommend then?

That depends. Normally, a simple *Video Preferences* panel would suffice. This is where the current preference panels reside:

https://github.com/mavlink/qgroundcontrol/tree/master/src/ui/preferences

> Humn... no. The idea is you have an external process running on the drone that can tell the GCS what are the protocols and everything it needs to consume the video stream. The intention (if we are not there yet) is to be plug and play: they negotiate the parameters and then have the video stream. We can allow "manual configuration", but ideally the drone would be smart enough to inform the GCS how to setup its end of the connection.

That might be your idea but it's not how things work. You can't expect to retrofit this into an unknowable number of video sources. The way you propose would work great for flight controllers (or anything in between) that support this but we also need to support video coming with no other information other than what QGC (its user actually) knows about.
 
> For Solo, it would mean to run the additional daemon there and everything should work. If it doesn't work it's just because it's still a work in progress.

Are you proposing that the only way to get video from the Solo is to reprogram it so it provides what we already know and have working in the current code? Don't get me wrong. The direction you are going is the *ideal* scenario but we cannot stop supporting manually defined video pipelines.

lucasdemarchi:
> Are you proposing that the only way to get video from the Solo is to reprogram it so it provides what we already know and have working in the

If by reprogram you mean running a daemon on its "companion computer", then yes.

> current code? Don't get me wrong. The direction you are going is the ideal scenario but we cannot stop supporting manually defined video pipelines. 

Note that I said there could be an option for manual pipeline. However if we can run the server in the other end there's no reason we shouldn't aim that... It's a matter of negotiating the video sources and parameters between the drone and the GCS. The manual step from my pov can be an option but not necessarily the rule.

lucasdemarchi:
> If by reprogram you mean running a daemon on its "companion computer", then yes.

Just rephrasing what I said: if you can not run something there to advertise the video sources, then you resort to the manual method which I do think is necessary to stay. I just wouldn't like to rule out the autoconfig option.

dogmaphobic:
> then you resort to the manual method which I do think is necessary to stay.

Yep, that's what I meant and it seems fine to me (once we move the settings into a proper *Preferences Panel* and restore UVC video).

 What we need now is a proper protocol for the video streaming definition.

@LorenzMeier This is quite a bit to add to MavLink. Any ideas? Note that the "flight controller" more than likely would have no idea about any of this. It would mostly be up to something in between. I have not seen anything that does what is being proposed here so I have zero basis for analysis and/or comparison.

dogmaphobic:
By the way, what I meant earlier at the top of this thread is the addition of libraries that are not currently shipped with some of the builds. I see for instance the use of `rtpjpegdepay`. That will require additional GStreamer libraries to be bundled with QGC. When adding these things, care should be taken to make sure it works in all platforms. The fastest way to test is to build and run it on Android, where each library must be hand picked (as opposed to linking all 300+ MB of GStreamer libraries).

DonLakeFlyer:
Should I close this?

dogmaphobic:
Either close or it needs to be fixed.

ricardotk:
> Should I close this?

Yes, I'm now dealing with the other part of this solution. I'll close it and send a fixed version soon.

PR #4032, opened 2016-09-13
Improve USB Interface detection reliability and support for LibrePilot compatible USB IDs

AlessioMorale:
This PR aim at adding support for several boards compatible with LibrePilot firmware for its upcoming MAVLink interface Implementation.
This is done with the following additions: 
- More robust Interface detection. It is not tied to a specific layout and support composite interfaces;
- USB id for currently supported boards are supplied;
- DTR is asserted when the port is open;

### DIFF
--- a/src/comm/QGCSerialPortInfo.h
+++ b/src/comm/QGCSerialPortInfo.h
@@ -37,6 +37,7 @@ public:
         BoardTypeMINDPXFMUV2,
         BoardTypeTAPV1,
         BoardTypeASCV1,
+        BoardTypeLibrePilot,
         BoardTypeUnknown
     } BoardType_t;
 
@@ -65,6 +66,12 @@ public:
 
     static const int ubloxRTKVendorId =                     5446;   ///< Vendor ID for ublox RTK
     static const int ubloxRTKProductId =                    424;    ///< Product ID for ublox RTK
+
+    static const int openpilotVendorId =                    0x20A0; ///< Vendor ID for OpenPilot
+    static const int revolutionProductId =                  0x415E; ///< Product ID for OpenPilot Revolution
+    static const int oplinkProductId =                      0x415C; ///< Product ID for OpenPilot OPLink
+    static const int sparky2ProductId =                     0x41D0; ///< Product ID for Taulabs Sparky2
+    static const int CC3DProductId =                        0x415D; ///< Product ID for OpenPilot CC3D
 
     QGCSerialPortInfo(void);
     QGCSerialPortInfo(const QSerialPort & port);
--- a/android/src/com/hoho/android/usbserial/driver/CdcAcmSerialDriver.java
+++ b/android/src/com/hoho/android/usbserial/driver/CdcAcmSerialDriver.java
@@ -297,6 +297,13 @@ public class CdcAcmSerialDriver extends CommonUsbSerialDriver {
                     UsbId.DEVICE_UBLOX_7,
                     UsbId.DEVICE_UBLOX_8,
                 });
+        supportedDevices.put(Integer.valueOf(UsbId.VENDOR_OPENPILOT),
+                new int[] {
+                    UsbId.DEVICE_CC3D,
+                    UsbId.DEVICE_REVOLUTION,
+                    UsbId.DEVICE_SPARKY2,
+                    UsbId.DEVICE_OPLINK,
+                });
         return supportedDevices;
     }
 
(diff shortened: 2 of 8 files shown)


DonLakeFlyer:
@dogmaphobic Can you look? 

Also I think there should be corresponding changes to the non-android auto-connect code/ids as well?

DonLakeFlyer:
@AlessioMorale Looks like I'm going to need to start flying these things for QGC testing. Also going to be very important to get out in front of the mavlink implementation so that it conforms to spec. There are lots of gotchas that can end up requiring a GCS to wrote firmware specific code. Would be good to avoid those from the start. Is there better place we could kick off a discussion of full QGC support for you guys?

AlessioMorale:
Hi @DonLakeFlyer, our MAVLink implementation is not complete yet, so still subject to big changes. This PR is strictly for the "physical interface", to allow a working serial connection to our hardware/firmware using the CDC interface.
 This is needed as our Firmware uses a composite USB interface that expose both an HID and a CDC interfaces that the current QGC usb interface implementation was not able to handle.

On the other hand if there is some test matrix or guidelines that I can use during the development I'll be very happy. 
Right now i used the information available at http://qgroundcontrol.org/mavlink
Thanks!

AlessioMorale:
I was not yet able to look at the autoconnect part. I think i should be able to spend some time with that at the beginning of next week.

DonLakeFlyer:
If you look here: https://github.com/mavlink/qgroundcontrol/blob/master/src/comm/QGCSerialPortInfo.h#L26. You should see similar board ids and so forth. This is for the non-android auto-connect code path.

DonLakeFlyer:
I'd like to see this change cover both Android and desktop OS before it goes through.

AlessioMorale:
@DonLakeFlyer  I'm implementing the QT part.

src/comm/QGCSerialPortInfo.h
```diff
@@ -66,6 +67,11 @@ class QGCSerialPortInfo : public QSerialPortInfo
     static const int ubloxRTKVendorId =                     5446;   ///< Vendor ID for ublox RTK
     static const int ubloxRTKProductId =                    424;    ///< Product ID for ublox RTK
 
+    static const int openpilotVendorId =                    0x20A0; ///< Vendor ID for OpenPilot
+    static const int revolutionProductId =                  0x415E; ///< Product ID for OpenPilot Revolution
+    static const int oplinkProductId =                      0x415C; ///< Product ID for OpenPilot OPLink
+    static const int sparky2ProductId =                     0x41D0; ///< Product ID for Taulabs Sparky2
+    static const int CC3DProductId =                        0x415D; ///< Product ID for OpenPilot CC3D
```
DonLakeFlyer: Can you add a blank line after this?

DonLakeFlyer:
Desktop side looks perfect @AlessioMorale except for my one anal spacing thing. @dogmaphobic Can you look at the android side? I think it's ok, but I'm not that familiar with that code.

AlessioMorale:
Don't worry i'm acquainted to pretty strict source formatting :)

dogmaphobic:
> @dogmaphobic Can you look at the android side? I think it's ok, but I'm not that familiar with that code.

Yeah, right :) I know zilch about Java and have no desire to change that. I copied and pasted that entire Android USB Serial from an Android NDK sample code. With that said, this code is only adding support for new USB interfaces. It is not changing (much) the original code.

AlessioMorale:
I made the following changes to java code:
previously it was just blindly taking interface 0/Endpoint 0 as control endpoint and interface 1/EP0 and 1 as data Write/read endpoints without checking anything.
Now I search within the interfaces for the ones with the relevant interface class (USB_CLASS_COMM for control and USB_CLASS_CDC_DATA for data). 
In a similar way I also pick endpoints based on their direction (USB_DIR_IN /USB_DIR_OUT).
In case something goes wrong I fail back to the previous blind assingment mechanism

DonLakeFlyer:
@AlessioMorale What sort of testing did you do with this? Just the LibrePilot boards or did you also verify Pixhawk connection is still working? Want to know so I can do some testing myself on additional boards if needed.

AlessioMorale:
@DonLakeFlyer sorry, missed your question: Unfortunately I have no pixhawk. Beside LP hardware I tried a common ftdi serial and its layout was recognized correctly

DonLakeFlyer:
Ok, I'll do some testing on this today.

AlessioMorale:
Thanks!

DonLakeFlyer:
Tested: AutoConnect working fine on Pixhawk.

PR #4044, opened 2016-09-16
Maps:Added Statkart Topo2 (very useful norwegian topographic map)

AndKe:
This map is an absolute necessary for navigating the northern part of the country, steep mountains and deep valleys.

### DIFF
--- a/src/FlightMap/FlightMapSettings.cc
+++ b/src/FlightMap/FlightMapSettings.cc
@@ -32,6 +32,7 @@ void FlightMapSettings::setToolbox(QGCToolbox *toolbox)
 #ifndef QGC_NO_GOOGLE_MAPS
     _supportedMapProviders << "Google";
 #endif
+    _supportedMapProviders << "Statkart";
     _loadSettings();
 }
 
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
@dogmaphobic Can you take a look at this?

AndKe:
BTW: due to groundcontrol not saving last map position&zoom (and even if it did)   - when you open the application with this map, you need to zoom out a *lot*  - until Norway can be seen, then zoom in.

dogmaphobic:
Implementation looks correct. I understand the usefulness for someone in Norway but I wish there was a way for this not to be *global*. That is, it's an option for 0.06% of the population. If we start adding specialized maps for everything, we will end up with a rather large list of map sources. With that said, I can't think of a way to narrow the options so it is what it is.

AndKe:
I don't think others will be bothered by it, after all, it's just something in a droplist in settings.

DonLakeFlyer:
I'll do some testing of this tomorrow.

DonLakeFlyer:
Tested and looks good. Thanks @AndKe

PR #4053, opened 2016-09-18
Switching to XCode 8

DonLakeFlyer:


DonLakeFlyer:
XCode 8 breaks Qt qmake. Great!

PR #4058, opened 2016-09-19
Another attempt at Xcode 8

dogmaphobic:
Patching Qt qmake files for Xcode 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -125,7 +125,7 @@ install:
         export QT_DIR=Qt5.5-mac-clang/5.5/clang_64 &&
         export QT_QPA_PLATFORM_PLUGIN_PATH=/tmp/$QT_DIR/plugins &&
         export QML2_IMPORT_PATH=/tmp/$QT_DIR/qml &&
-        export PATH=/tmp/$QT_DIR/bin:$PATH
+        export PATH=/tmp/$QT_DIR/bin:$PATH &&
         tools/patch_qt_for_xcode8.sh
         ;
     fi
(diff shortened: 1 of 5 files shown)


AdyashaDash:
Does that mean that support for xcode 7.3.1 is being withdrawn? I can't compile with xcode 7.3.1 on macosx10.11 now.

The error message I receive says that it could not resolve the SDK path any more.

dogmaphobic:
The SDK is hardcoded in as qmake isn't picking it up automatically. This is done here:
https://github.com/mavlink/qgroundcontrol/blob/master/QGCCommon.pri#L69
It wouldn't be too hard to make it user configurable. I would not attempt at automation because the output of `xcodebuild -showsdks` seems to change all the time.

DonLakeFlyer:
We always have to move to the latest XCode when it comes out since it's the only way to push to the store. That said you can always modify the sdk verion back youself in your own build. Although if you run into problems with a QGC built on older SDK it generally would not be supported. But I think that rarely happens.

dogmaphobic:
Hmm... I just removed that line (`QMAKE_MAC_SDK = macosx10.12`) from QGCCommon.pri and it built it all properly. I only have 10.12 installed (even though I'm still running 10.11) so I can't say for sure this is back working. @AdyashaDash, can you trying commenting that line off (or removing altogether) and see if your build works? If that's the case, we can remove (or leave it commented off).

`QMAKE_MAC_SDK` is set to `macosx` by default (qmake). It used to be number specific. I guess they fixed qmake after all.

AdyashaDash:
@dogmaphobic Thank you so much for your quick response! Removing the said line works for me now.

Also, to clarify, earlier I could still compile master if I didn't include the commits from this PR. After commenting out `QMAKE_MAC_SDK = macosx10.12` now though, I can build master smoothly, without having to removing anything else.

AdyashaDash:
short update: I also just realized while trying to run Software-in-the-loop on PX4 master today that they have moved to xcode 8 as well, so I had to upgrade to xcode 8 sooner rather than later anyway :)

PR #4062, opened 2016-09-20
Show help button on FactTextFields

DonLakeFlyer:
Any place you use a FactTextField you now automatically get detailed help associated with the value. Here is an example using it on a settings (offline edit cruise speed) that may be unclear to the user. Also an example for a param on a mission item.
screen shot 2016-09-19 at 1 52 21 pm
screen shot 2016-09-19 at 1 52 58 pm
screen shot 2016-09-19 at 1 53 03 pm
screen shot 2016-09-19 at 1 53 20 pm

Note: Mission Items params don't have good meta data yet. I'm working on that.

### DIFF
--- a/src/FactSystem/FactControls/FactTextField.qml
+++ b/src/FactSystem/FactControls/FactTextField.qml
@@ -14,6 +14,7 @@ QGCTextField {
     text:       fact ? fact.valueString : ""
     unitsLabel: fact ? fact.units : ""
     showUnits:  true
+    showHelp:   true
 
     property Fact   fact:           null
     property string _validateString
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
This means it is very important to fill out meta data for non-parameter Fact usage. I've created additional tooling in FactMetaData to load meta data from json file to make that easier.

birchera:
This is a very cool feature that will drastically improve the usability for inexperienced users.

We had ideas about also having a link added to the short explanation that links to further information available somewhere in the net, e.g. the user guide. Not sure though how easily this can be implemented..

DonLakeFlyer:
Yeah. That's a great idea. Very easy to implement.

DonLakeFlyer:
https://github.com/mavlink/qgroundcontrol/issues/4069

PR #4075, opened 2016-09-24
Linux add icon with setWindowIcon

### LINKED ISSUE #4073: No application icon (Linux)
There is no application icon, so the generic gray question mark shows up instead.

dagar:
- fixes #4073

### DIFF
--- a/qgcresources.qrc
+++ b/qgcresources.qrc
@@ -3,11 +3,8 @@
         <file alias="opensans">resources/fonts/OpenSans-Regular.ttf</file>
         <file alias="opensans-demibold">resources/fonts/OpenSans-Semibold.ttf</file>
     </qresource>
-
     <qresource prefix="/qmlimages">
         <file alias="AirframeComponentIcon.png">src/AutoPilotPlugins/Common/Images/AirframeComponentIcon.png</file>
-
-        <!-- These are here for back compat with previous airframe meta data -->
         <file alias="AirframeFlyingWing.png">src/AutoPilotPlugins/Common/Images/FlyingWing.svg</file>
         <file alias="AirframeHexaRotorPlus.png">src/AutoPilotPlugins/Common/Images/HexaRotorPlus.svg</file>
         <file alias="AirframeHexaRotorX.png">src/AutoPilotPlugins/Common/Images/HexaRotorX.svg</file>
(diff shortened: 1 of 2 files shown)


src/main.cc
```diff
@@ -211,6 +212,8 @@ int main(int argc, char *argv[])
     QGCApplication* app = new QGCApplication(argc, argv, runUnitTests);
     Q_CHECK_PTR(app);
 
+    QApplication::setWindowIcon(QIcon(":/res/resources/icons/qgroundcontrol.ico"));
```
DonLakeFlyer: Shouldn't this be Linux only?

qgcresources.qrc
```diff
@@ -3,11 +3,8 @@
         <file alias="opensans">resources/fonts/OpenSans-Regular.ttf</file>
         <file alias="opensans-demibold">resources/fonts/OpenSans-Semibold.ttf</file>
     </qresource>
-
```
DonLakeFlyer: Your changes are pulling out blank lines and comments.

src/main.cc
dagar: Shouldn't it be fine everywhere? The other options for windows and osx set the icon for the binary which is also used for the app if this hasn't been set.

qgcresources.qrc
dagar: This is what happens when you use the qtcreator editor.

src/main.cc
DonLakeFlyer: Famous last words :) I'd rather not take the chance. Otherwise I'll need to test all OS.

qgcresources.qrc
DonLakeFlyer: Hence the reason I never use qtcreator to edit resources files. It's terrible.

src/main.cc
dagar: Fair enough.

qgcresources.qrc
dagar: Doesn't really matter to me. I've done this several times before, so this time I just kept the change. Your call.

DonLakeFlyer: It guess it's gonna get trashed at some point. The other file is alreaedy trounced anyway.

PR #4076, opened 2016-09-24
remove mavlink and gps driver submodules

### LINKED ISSUE #4031: absorb mavlink and gpsdrivers submodules
@LorenzMeier @DonLakeFlyer - If I made it trivial (and automated) to pull in new versions of mavlink and gpsdrivers would either of you be opposed to getting rid of git submodules in this project? 

My primary reason is that launchpad isn't able to handle them, but I think it would also reduce developer pain. Launchpad can automatically build QGC on each of the supported Ubuntu versions and create both stable and nightly development PPAs. It would also be a step towards having QGC officially in Ubuntu if desired.

QGC doesn't have the PX4 submodule helper scripts that make working with submodules less painful.

https://github.com/mavlink/qgroundcontrol/issues/3886
https://github.com/mavlink/qgroundcontrol/issues/665

dagar:
- closes #4031

Big initial import, but each incremental change should be manageable and might actually be nice to see in the history.

Update mavlink - `./tools/update_mavlink.sh`
Update GPSDrivers - `./tools/update_gpsdrivers.sh`


DonLakeFlyer:
@dagar Will a new git pull upstream master with this do the right thing to update old repos?

dagar:
I'll verify.

dagar:
Sigh, a regular pull needs to be forced if the contents are in the same location as the deleted submodule (in the same pull). I'll just need to change the paths.

DonLakeFlyer:
@dagar ?

dagar:
Still planned. I'll only proceed if I can make the transition painless.

dagar:
Submodules are terrible. Might revisit later.

PR #4086, opened 2016-09-29
px4firmwareplugin: fix minimum required version comparison

bkueng:
before, with a PX4 version of eg. 1.5.0, the user was wrongly notified.


DonLakeFlyer:
Thanks.

PR #4096, opened 2016-10-03
Must use pack_chan and encode_chan due to mav 2.0 problem

DonLakeFlyer:
@LorenzMeier This is what I had to do since with mav 2.0 I don't see how you can use the non *_chan _pack or _encode apis any longer. Unless you have some other ideas I think I'm stuck with this?

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSensorsComponentController.cc
+++ b/src/AutoPilotPlugins/APM/APMSensorsComponentController.cc
@@ -468,9 +468,13 @@ void APMSensorsComponentController::nextClicked(void)
 
     ack.command = 0;
     ack.result = 1;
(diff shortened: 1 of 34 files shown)


LorenzMeier:
I'm not so sure if you needed to really do that or just re-pack properly, but it looks like useful cleanup anyway?

LorenzMeier:
@DonLakeFlyer I should have been more clear about things: The pack function increases the outgoing message counter. So it should always be `_pack_chan` because otherwise the message counter for the default channel is messed up if you re-run the packing and end up sending via a different link.

The messed up counter will look like packet drop on the receiver side. So it is pretty significant
to get this right. I suggest to fix up CI and then I'm more than happy to run this through MAVLink 1 and 2 testing.

@bkueng Can you assist don with testing once he calls it ready?

PR #4102, opened 2016-10-04
Initial GeoTag implementation

DonLakeFlyer:
* Top level Analyze toolbar button is back. Things like Log Download, Mavlink Inspector and so forth will move there.
* New GeoTag page is current only support in there.
* Icons TBD

screen shot 2016-10-03 at 7 16 34 pm


DonLakeFlyer:
@birchera Your code goes into GeoTagWorker::run

birchera:
Thanks a lot! We'll get working on the tagging asap.

PR #4106, opened 2016-10-04
Fix saving and loading mission file on tablet

AdyashaDash:
Since [here](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/MissionController.cc#L481) the save function requires a filename as input, but we weren't providing this [here](https://github.com/mavlink/qgroundcontrol/compare/master...wingtra:fix/load_save_on_tablet?expand=1#diff-476fe2248af8c57c641f0128ede640a1R259), we couldn't save mission files on tablet.

Further, because of a typo in `loadFromFile`, couldn't load mission files, either.

Fixing these two typos here.


DonLakeFlyer:
Thanks

PR #4108, opened 2016-10-05
Fix insert functionality after Survey Item

AdyashaDash:
What is happening -

screen shot 2016-10-05 at 15 57 09

Currently, the insert functionality (as shown in the screenshot) after a Survey Item, doesn't work and if trying to use thus, qgc stops working in the planning view.

This is because, on using the insert functionality, the sequence number set [here](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionEditor/MissionItemEditor.qml#L77) is the actual sequence number allotted to the waypoint after the SurveyItem has been converted to waypoints, i.e., a large number such as 156 as seen above.

This is then passed to function `insertSimpleMissionItem` defined [here](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/MissionController.cc#L159). However, this function sets indices based on "visual items", [(see example)](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/MissionController.cc#L183), which counts the entire survey item as one big item. Thus, there is a mismatch between indices after SurveyItem when using the `insert()` signal via qml  and then you get the message `Index is not valid`.

This PR fixes the issue.

### DIFF
--- a/src/MissionEditor/MissionEditor.qml
+++ b/src/MissionEditor/MissionEditor.qml
@@ -672,7 +672,7 @@ QGCView {
                             }
 
                             onInsert: {
-                                var sequenceNumber = missionController.insertSimpleMissionItem(editorMap.center, insertAfterIndex)
+                                var sequenceNumber = missionController.insertSimpleMissionItem(editorMap.center, index)
                                 setCurrentItem(sequenceNumber)
                             }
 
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Perfect, thanks

PR #4109, opened 2016-10-05
Mavlink 2.0 support

DonLakeFlyer:
No signing support yet.

This pull still has these two unresolved issues:
* https://github.com/mavlink/mavlink/issues/630
* https://github.com/mavlink/mavlink/issues/625

But I think it's still ok without those.

### DIFF
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -17,7 +17,7 @@ WindowsBuild {
 # a single compiled codebase this hardwiring of dialect can go away. But until then
 # this "workaround" is needed.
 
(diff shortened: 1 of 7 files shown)


DonLakeFlyer:
@LorenzMeier Testing help would be great.

DonLakeFlyer:
QGroundControl - current implementation
* QGC assumes vehicle has control of mavlink 2 determination. I think this is how MP works as well since I can find any settings in MP to turn on/off mavlink 2 support.
* If QGC sees mavlink 2 coming inbound from vehicle on channel it will switch to mavlink 2 for outbound as well.
* There is no setting in QGC to control mavlink 2 outbound to vehicle.

PR #4110, opened 2016-10-05
Update to SDL2 for Joystick Handling

rjehangir:
This PR updates the Simple DirectMedia Layer (SDL) library from 1.2 to 2.0.4. SDL is used for cross-platform joystick handling in QGroundControl and upgrading to SDL2 adds new and improved functionality.

It enables the following features to be implemented:
- Stick/button consistency across OSes (already implemented)
- With the stick/button consistency, we can eventually get rid of the need for a calibration process entirely and you can just choose Mode 1-4 instead
- Hotplugging of joystick devices so you can plug in after starting QGC or if your Xbox controller disconnects you don't have to restart QGC entirely. This will also stop the MANUAL_CONTROL messages from continuing when the joystick is unplugged.
- Battery monitoring for wireless game controllers

There are a few important consequences of this update as well:

- On all OSes, the joystick and button mappings will change, requiring the user to recalibrate. I've included code to check and require a recalibration if the user has upgraded from an earlier SDL version.

This is a relatively simple PR that shifts over the current functionality, and we will implement the other features in the future.

This has been tested on Mac OSX, Ubuntu 14.04, and Windows 10. 

Also note that this update uses the [SDL_GameControllerDB](https://github.com/gabomdq/SDL_GameControllerDB) source, which is copied into the `resources` folder. This code uses the exact same license as SDL itself and the license text is included in that folder, per license requirements.

Please let me know if you have any questions, comments, or concerns about this PR.

-Rusty


DonLakeFlyer:
Thanks a ton Rusty.

rjehangir:
Thanks, Don!

PR #4122, opened 2016-10-11
linux package always include libQt5Multimedia

### LINKED ISSUE #4105: QGC is not opening
Hi i am working ubuntu 14.04 64 bit system i downloaded the daily buid of qgc from its site and isntalled it and also install the additional packages given in github site. But the problem is whenever i start my QGc application is starts with full white background 
screenshot from 2016-10-04 12 39 31

dagar:
- fixes #4105 

My default qmake build had QGC_DISABLE_UVC defined, but still needed these libs. Any objection to just leaving them in all the time?


DonLakeFlyer:
Seems fine to me

Issue #4117, opened 2016-10-07
Add support for accumulator on Joystick throttle


gregd72002:
Some gamepads/joysticks have a spring on all the axis (including throttle). It is pretty impossible to keep the throttle in a given position so flying is difficult. Adding an option in Joystick config to accumulate throttle could be considered to assist.

I personally tried flying with PS3 controller and all works fine but it is difficult to keep the throttle in one position. I have tested an accumulator on the throttle to a great success in the following way:
- when throttle stick is in idle - the accumulator is 0
- when throttle stick is positive - the accumulator gets increased accordingly in a loop
- when throttle stick is negative - the accumulator gets decreased accordingly
- the accumulator is fed as a throttle for processing

As a bonus there could be also a rate option at which the accumulator gets increased/decreased.

PR #4123, opened 2016-10-11
Implementing throttle axis accumulator with deadband support

gregd72002:
Ref: https://github.com/mavlink/qgroundcontrol/issues/4117

This relates only to Joysticks/Gamepads

- adds option for throttle accumulator
- deadband support on all axis when in accumulator mode (requires re-calibration)

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -23,6 +23,7 @@ const char* Joystick::_calibratedSettingsKey =      "Calibrated1"; // Increment 
 const char* Joystick::_buttonActionSettingsKey =    "ButtonActionName%1";
 const char* Joystick::_throttleModeSettingsKey =    "ThrottleMode";
 const char* Joystick::_exponentialSettingsKey =     "Exponential";
+const char* Joystick::_accumulatorSettingsKey =     "Accumulator";
 
 const char* Joystick::_rgFunctionSettingsKey[Joystick::maxFunction] = {
     "RollAxis",
@@ -46,6 +47,7 @@ Joystick::Joystick(const QString& name, int axisCount, int buttonCount, int hatC
     , _lastButtonBits(0)
     , _throttleMode(ThrottleModeCenterZero)
     , _exponential(false)
+    , _accumulator(false)
     , _activeVehicle(NULL)
     , _pollingStartedForCalibration(false)
     , _multiVehicleManager(multiVehicleManager)
@@ -86,6 +88,7 @@ void Joystick::_loadSettings(void)
 
     _calibrated = settings.value(_calibratedSettingsKey, false).toBool();
     _exponential = settings.value(_exponentialSettingsKey, false).toBool();
+    _accumulator = settings.value(_accumulatorSettingsKey, false).toBool();
 
     _throttleMode = (ThrottleMode_t)settings.value(_throttleModeSettingsKey, ThrottleModeCenterZero).toInt(&convertOk);
     badSettings |= !convertOk;
@@ -96,6 +99,7 @@ void Joystick::_loadSettings(void)
     QString maxTpl  ("Axis%1Max");
     QString trimTpl ("Axis%1Trim");
     QString revTpl  ("Axis%1Rev");
+    QString deadbndTpl  ("Axis%1Deadbnd");
 
     for (int axis=0; axis<_axisCount; axis++) {
         Calibration_t* calibration = &_rgCalibration[axis];
@@ -109,9 +113,13 @@ void Joystick::_loadSettings(void)
         calibration->max = settings.value(maxTpl.arg(axis), 32767).toInt(&convertOk);
         badSettings |= !convertOk;
 
+        calibration->deadband = settings.value(deadbndTpl.arg(axis), 0).toInt(&convertOk);
+        badSettings |= !convertOk;
+
         calibration->reversed = settings.value(revTpl.arg(axis), false).toBool();
 
-        qCDebug(JoystickLog) << "_loadSettings axis:min:max:trim:reversed:badsettings" << axis << calibration->min << calibration->max << calibration->center << calibration->reversed << badSettings;
+
+        qCDebug(JoystickLog) << "_loadSettings axis:min:max:trim:reversed:badsettings" << axis << calibration->min << calibration->max << calibration->center << calibration->reversed << calibration->deadband << badSettings;
     }
 
     for (int function=0; function<maxFunction; function++) {
@@ -145,6 +153,7 @@ void Joystick::_saveSettings(void)
 
     settings.setValue(_calibratedSettingsKey, _calibrated);
     settings.setValue(_exponentialSettingsKey, _exponential);
+    settings.setValue(_accumulatorSettingsKey, _accumulator);
     settings.setValue(_throttleModeSettingsKey, _throttleMode);
 
     qCDebug(JoystickLog) << "_saveSettings calibrated:throttlemode" << _calibrated << _throttleMode;
@@ -153,6 +162,7 @@ void Joystick::_saveSettings(void)
     QString maxTpl  ("Axis%1Max");
     QString trimTpl ("Axis%1Trim");
     QString revTpl  ("Axis%1Rev");
+    QString deadbndTpl  ("Axis%1Deadbnd");
 
     for (int axis=0; axis<_axisCount; axis++) {
         Calibration_t* calibration = &_rgCalibration[axis];
@@ -161,6 +171,7 @@ void Joystick::_saveSettings(void)
         settings.setValue(minTpl.arg(axis), calibration->min);
         settings.setValue(maxTpl.arg(axis), calibration->max);
         settings.setValue(revTpl.arg(axis), calibration->reversed);
+        settings.setValue(deadbndTpl.arg(axis), calibration->deadband);
 
         qCDebug(JoystickLog) << "_saveSettings name:axis:min:max:trim:reversed"
                                 << _name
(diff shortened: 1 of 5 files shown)


rjehangir:
@gregd72002 Are you actually using the deadband? It looks like it's hardcoded to zero?

I've been working on a deadband for the joystick code as well so I'd like to coordinate efforts.

gregd72002:
@rjehangir sure, my deadband implementation is very rudimental. It was just implemented into to satisfy the accumulator on throttle so it does not change without user action.

In this implementation:
- deadband is set to 0 initially
- during 1st stage of calibration where sticks are supposed to be centred, it probes the deviations to get the highest for each of the axis; this is done in JoystickConfigController.cc (_inputCenterWaitBegin)
- this is then used in Joystick.cc (_adjustRange) 
- this will only work when axis are centred around 0, hence I apply the deadband only when in throttle accumulator mode (what implies 0 centre)

I'm sure there are better way of doing it. What's your take on it?

rjehangir:
@gregd72002 Okay. Thanks for the explanation.

The deadband I implemented is really simple. It's just a fixed value around the center. I think the main joystick that needs the deadband is the Xbox controller so I experimentally chose the deadband value to work well with that. Most other controllers that I've tried have very clearly defined deadbands by default. 

Since I'm working mostly with ArduSub, my throttle is also centered around zero.

I added a checkbox to the joystick page to turn the deadband on/off.

Thoughts? Yours is a little more sophisticated as long as it works reliably.

-Rusty

gregd72002:
It would be nice to be able to specify custom deadband as opposed to 'sensed' or 'pre-defined'.

Possibly it should be specified as % as opposed to axis units - Thoughts?
Maybe introduce new fields next to 'Axis Monitor' that allows user to define deadband for each axis. I could program it but QML is not my strength...

I have been testing this with PS3 controller.

rjehangir:
@gregd72002 I personally think that adding the ability to specify a custom deadband is too much control to the user and would probably end up confusing them. I'd prefer to keep things as simple and clear as possible. I imagine we could take care of 95% of joysticks with a fixed deadband width.

Thoughts?

gregd72002:
From my experience joystick do change deadband due to wear and tear. While hardcoding the deadband would work for me maybe the sensing option is more versatile?

I put together a quick mockup:
- the deadband is now an option
- it is sensed as described before
- when active, the deadbands are shown red on each of the axis

screen shot 2016-10-12 at 23 55 38

DonLakeFlyer:
Deadband:
* I'm with Rusty. Too many options and you scare people off. 
* The auto-detect is nice, but I would also include a fixed deadband value. Use the max of both.
* I would also warn if auto-detect deadband is too high. This could cause mysterious behavior to the end user.
* Any reason why deadband shouldn't be on all the time? Fewer options the better.

Accumulator:
* If like to find a different name for this. Maybe something alongs the line of "smoothing"?

rjehangir:
@DonLakeFlyer If a fixed deadband is always applied then it will effectively increase the deadband for joysticks that already have large deadbands, like the Logitech 310. Might not be a huge deal though.

DonLakeFlyer:
Ok, you guys know way more about this than I do. Just trying to remove an option for simplification.

gregd72002:
I've added the warning message and renamed a couple of things.

What do you think about the labels? 

screen shot 2016-10-13 at 23 31 10

rjehangir:
@gregd72002 What if we just limit the auto-detected deadband to 0-10% so it can never be set to a crazy number? Then we could get rid of the warning.

Also, as a side note, with the switch to SDL2, I'm hoping to eventually eliminate the calibration process entirely. SDL2 has consistent stick and button numbering so there is no need to detect axes anymore. The deadband could not be detected from that, unfortunately.

Last, is "Spring on throttle support" the new name for accumulator?

gregd72002:
@rjehangir I do not see particular reason for the warning message. The axis clearly show the deadband that is being applied. Also, deadbands are only in use when user explicitly activates them. The implication in here is that they know what deadbands are before activating them. 
For all the accidental deadbands we have the red marker on the axis.

Re SDL2, once the calibration is removed we could introduce user defined deadbands?

"Spring on throttle support" is the new name I proposed for the accumulator.

src/Joystick/Joystick.cc
```diff
@@ -109,9 +116,13 @@ void Joystick::_loadSettings(void)
...
         badSettings |= !convertOk;
 
+        calibration->deadband = settings.value(deadbndTpl.arg(axis), 0).toInt(&convertOk);
+        badSettings |= !convertOk;
+
         calibration->reversed = settings.value(revTpl.arg(axis), false).toBool();
 
-        qCDebug(JoystickLog) << "_loadSettings axis:min:max:trim:reversed:badsettings" << axis << calibration->min << calibration->max << calibration->center << calibration->reversed << badSettings;
+
+        qCDebug(JoystickLog) << "_loadSettings axis:min:max:trim:reversed:badsettings" << axis << calibration->min << calibration->max << calibration->center << calibration->reversed << calibration->deadband << badSettings;
```
DonLakeFlyer: Can you update the text for the new values?

src/Joystick/Joystick.cc
```diff
@@ -161,14 +175,16 @@ void Joystick::_saveSettings(void)
         settings.setValue(minTpl.arg(axis), calibration->min);
         settings.setValue(maxTpl.arg(axis), calibration->max);
         settings.setValue(revTpl.arg(axis), calibration->reversed);
+        settings.setValue(deadbndTpl.arg(axis), calibration->deadband);
 
         qCDebug(JoystickLog) << "_saveSettings name:axis:min:max:trim:reversed"
```
DonLakeFlyer: Same text update here?

src/Joystick/Joystick.cc
```diff
@@ -214,7 +238,7 @@ float Joystick::_adjustRange(int value, Calibration_t calibration)
                             << calibration.min
                             << calibration.max
                             << calibration.center
-                            << calibration.center
+                            << calibration.deadband
```
DonLakeFlyer: Remove center? Shouldn't this be like the others with an add of deadband? Also not sure if there is text update needed here.

src/Joystick/Joystick.h
```diff
@@ -66,6 +67,8 @@ class Joystick : public QThread
 
     Q_PROPERTY(int throttleMode READ throttleMode WRITE setThrottleMode NOTIFY throttleModeChanged)
     Q_PROPERTY(int exponential READ exponential WRITE setExponential NOTIFY exponentialChanged)
```
DonLakeFlyer: Shouldn't all of these be bool (except throttle mode)?

src/VehicleSetup/JoystickConfig.qml
```diff
@@ -87,6 +87,20 @@ SetupPage {
                         color:                  __barColor
                     }
 
+                    // Deadband
```
DonLakeFlyer: It's sort of interesting to display this, but I really wonder whether it will just confuse users as to what it is?

src/VehicleSetup/JoystickConfig.qml
```diff
@@ -382,6 +407,36 @@ SetupPage {
...
+                                Row {
+                                    x:          20
+                                    width:      parent.width
+                                    spacing:    ScreenTools.defaultFontPixelWidth
+                                    visible:    _activeJoystick.throttleMode == 0
+
+                                    QGCCheckBox {
+                                        id:         deadband
+                                        checked:    _activeJoystick.deadband
+                                        text:       qsTr("Deadband (requires re-calibration)")
```
DonLakeFlyer: I'd make this stupid-proof and invalid the calibration if deadband is turned on. Either that or always calc deadband and save so it's always there, but just don't use if not turned on.

src/Joystick/Joystick.cc
gregd72002: center is printed 2 times, I replaced one of them

src/VehicleSetup/JoystickConfig.qml
gregd72002: By default deadbands will be disabled as such there should be no difference to users. It is only when they active them.

src/Joystick/Joystick.h
gregd72002: updated.

gregd72002:
PR refreshed with the comments.

src/VehicleSetup/JoystickConfig.qml
DonLakeFlyer: Sound fine then.

DonLakeFlyer:
With the headband display I can live without the warning. I still don't like the accumulator naming though. I think you want to describe what it does. My logitech F430 (name?) seems to work fine with it's spring loaded throttle. So I wouldn't use this. But if you call it just "spring loaded throttle" then it would push people to use this when they may not need to. Maybe add "smoothing" to the end up of so you hit both: "Spring loaded throttle smoothing"?

gregd72002:
Done

DonLakeFlyer:
Ok, this is good from review. I'm going to do some testing with it then I'll merge.

DonLakeFlyer:
Ran this today. Shouldn't deadbands be supported for throttle down zero as well?

gregd72002:
Well, this is not what I was trying to achieve. But I guess I get your point. Should we:
- move the deadband toggle to Advanced Options on the Joystick Setup (always available)
- when throttle centre 0 mode - the behaviour is as per current implementation
- when throttle down 0 mode - this deadband is not applied for throttle but applied for other axis
?

DonLakeFlyer:
Yeah, I think that all makes sense.

gregd72002:
This has been updated. Not sure why the check is failing - looking into the details shows some mavlink related stuff which should be irrelevant to this PR?

PR #4130, opened 2016-10-14
Switch ArduPilot setup pages to new SetupPage usage, plus vehicle branding

DonLakeFlyer:
SetupPage base control looks like this:
screen shot 2016-10-13 at 6 35 12 pm
The title and help text at the top come to SetupPage base control using VehicleComponent data. If the screen is short help text doesn't show.

Vehicle branding is now supported through FirmwarePlugin:
```
    /// Return the resource file which contains the brand image for the vehicle.
    virtual QString brandImage(const Vehicle* vehicle) const { Q_UNUSED(vehicle) return QString(); }
```
Branding examples:
screen shot 2016-10-13 at 6 35 39 pm
screen shot 2016-10-13 at 6 35 54 pm
screen shot 2016-10-13 at 6 36 08 pm


DonLakeFlyer:
@LorenzMeier I could use a slightly different variation of the PX4 Pro brand image. It works in outdoor style, but not in indoor style. If the black was outlined with white it would work while in indoor style.

rjehangir:
This looks great, Don! Very nice changes.

PR #4131, opened 2016-10-14
Geotagging backend

birchera:
Needs more documentation, testing and some tweaks in robustness.

### DIFF
--- a/src/AnalyzeView/GeoTagController.cc
+++ b/src/AnalyzeView/GeoTagController.cc
@@ -9,6 +9,9 @@
 
 #include "GeoTagController.h"
 #include "QGCFileDialog.h"
+#include "ExifParser.h"
+#include <QtEndian>
+#include <cfloat>
 
 GeoTagController::GeoTagController(void)
     : _progress(0)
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
I working my way through this now. May take me a little while.

src/AnalyzeView/GeoTagController.cc
```diff
@@ -73,15 +76,191 @@ void GeoTagWorker::run(void)
...
+    for (int i = 0; i < imageList.size(); ++i) {
+        QFile file(imageList.at(i).absoluteFilePath());
+        if (!file.open(QIODevice::ReadOnly)) {
+            continue;
+        }
+        _imageBuffers.append(QByteArray(file.readAll()));
+        file.close();
+    }
+
+    qWarning() << "Image loading time elapsed: " << timerLoadImages.elapsed() << " milliseconds";
```
DonLakeFlyer: These shouldn't be warnings. They should be qCDebug with a new logging category for geotagging.

src/AnalyzeView/GeoTagController.cc
```diff
@@ -73,15 +76,191 @@ void GeoTagWorker::run(void)
...
+
+    qWarning() << "Parse exif data time elapsed: " << timerParseExif.elapsed() << " milliseconds";
+
+    ////////// Load PX4 log
+    timerLoadLogFile.start();
+
+    _geoRef.clear();
+    _triggerTime.clear();
+    if (!parsePX4Log()) {
+        qWarning() << "Geotagging failed";
```
DonLakeFlyer: Shouldn't there be some sort of emit error here?

src/AnalyzeView/GeoTagController.cc
```diff
@@ -73,15 +76,191 @@ void GeoTagWorker::run(void)
...
+        return;
+    }
+    qWarning() << "Found " << _geoRef.count() << " trigger logs.";
+
+    qWarning() << "Log loading time elapsed: " << timerLoadLogFile.elapsed() << " milliseconds";
+
+    ////////// Filter Trigger
```
DonLakeFlyer: Just use regular ```//``` comments

src/AnalyzeView/GeoTagController.cc
```diff
@@ -73,15 +76,191 @@ void GeoTagWorker::run(void)
...
+    char triggerHeader[] = {0xA3, 0x95, 0x37, 0x00};
+    int triggerOffsets[2] = {3, 11};
+    int triggerLengths[2] = {8, 4};
+    // load log
+    QFile file(_logFile);
+    if (!file.open(QIODevice::ReadOnly)) {
+        qWarning() << "Could not open log file";
+        return false;
+    }
+    QByteArray log = file.readAll();
```
DonLakeFlyer: These log file can get pretty big can't they? readAll is going to chew a ton of memory.

src/AnalyzeView/GeoTagController.cc
```diff
@@ -73,15 +76,191 @@ void GeoTagWorker::run(void)
...
+    file.close();
+
+    // extract trigger data
+    int index = 1;
+    int sequence = -1;
+    QGeoCoordinate lastCoordinate;
+    while(index < log.count() - 1) {
+        int gposIndex = log.indexOf(gposHeader, index + 1);
```
DonLakeFlyer: Not sure this mechanism will work. You could have random bytes in the log file which could match these header bytes, but they are just in the midlle of some other log message. Seems like you are going to need to parse through the log sequentially to avoid incorrect detection.

birchera: Yes, I am aware of that shortcoming. For the case of the location, the simple fact that only a fraction of the messages are used to tag an image makes it unlikely that a wrong message could be used.

For the trigger logs I inserted the test on line 248 to check that the sequence number makes sense. The same could also be done for location to also catch the majority of the already very rare false positives.

The reason why it is implemented that way: Decoding the whole log, starting with the headers to get the formats of the messages seems quite tedious. Considering that px4log format will not be used for so much longer, I thought we could live with this version until it will be switched to ulog.

birchera:
Thanks for the review. Going to change according to your comments. I also already prepared some improvements and am going to pull them into the PR all together. Unfortunately moving a bit slow at the moment.

src/AnalyzeView/GeoTagController.cc
DonLakeFlyer: I think if you add the sequence check then I could live with it for now.

src/AnalyzeView/GeoTagController.cc
kellysteich: Do you have an idea what would be a good way to do it differently?

DonLakeFlyer:
@birchera Thinking abour wrapping up a 3.1 release to push out all the Survey work into a Stable. Would be great to get this in there.

birchera:
Sounds good. Current version is working well for us, but there is one more patch I would like to bring in, related with the log message format. This will currently prevent others from running it successfully.

birchera:
I updated with the changes I mentioned above. This is now ready from my side.

DonLakeFlyer:
Looks like you have a windows build problem to fix.

DonLakeFlyer:
Windows builds now, but all other OS break. Isn't multi-OS support fun :)

birchera:
It certainly comes at a price.. I will try to fix it.

birchera:
Seems to work out now.

PR #4135, opened 2016-10-14
Add ArduSub firmware branding

rjehangir:
This PR adds ArduSub branding with the new firmware branding feature. The ArduSub logo is only displayed when connected to an ArduSub vehicle.

Example:

ardusub-branding

Let me know if everything is alright with this.

-Rusty


DonLakeFlyer:
Cool! Check it in outdoor theme as well.

rjehangir:
I did! It looks good.

PR #4142, opened 2016-10-16
Fixed minor polygon drawing bug

nanthony21:
Fixed a bug where a black line would be rendered when starting to draw a polygon.

### DIFF
--- a/src/FlightMap/FlightMap.qml
+++ b/src/FlightMap/FlightMap.qml
@@ -367,7 +367,9 @@ Map {
                 // Update drag coordinate
                 var polygonPath = polygonDrawerPolygon.path
                 polygonPath[polygonDrawerPolygon.path.length - 1] = dragCoordinate
-                polygonDrawerPolygon.path = polygonPath
+                if (polygonDrawerPolygon.path.length>2){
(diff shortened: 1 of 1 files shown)


src/FlightMap/FlightMap.qml
```diff
@@ -367,7 +367,9 @@ Map {
                 // Update drag coordinate
                 var polygonPath = polygonDrawerPolygon.path
                 polygonPath[polygonDrawerPolygon.path.length - 1] = dragCoordinate
-                polygonDrawerPolygon.path = polygonPath
+                if (polygonDrawerPolygon.path.length>2){
```
DonLakeFlyer: Can you fix up the coding style spacing here. Should be:
```
if (polygonDrawerPolygon.path.length > 2) {
```

DonLakeFlyer:
Thanks.

This fixes one of the black line problems, but you still get one when you click to add the second point of the polygon and don't move the mouse right away. Do you have time to look at that. I think it happen when the second point equals the third point. Once you start to move the mouse after the second click it goes away.

nanthony21:
I'll take a look. I've noticed a few other issues with the polygon drawing but haven't been able to reproduce any of them.

Issue #3861, opened 2016-07-26
Clarify VTOL takeoff/land waypoints


AndreasAntener:
@sanderux we realized the VTOL takeoff/land are not clear next to the VTOL transition waypoint (for a new user). We should change the title and description:

**VTOL takeoff and transition**
Takeoff from ground using VTOL mode and transition to forwards flight.

**VTOL transition and land**
Transition to VTOL mode and land.

Thoughts? Suggestions?

sanderux:
I agree on the land text, for takeoff we should add that the vehicle will takeoff and transition, then fly to the position.

So: Takeoff in VTOL mode, then transition to forward flight and fly to this location.

AndreasAntener:
Sounds good

PR #4147, opened 2016-10-18
Update VTOL command text

DonLakeFlyer:
Fix for #3861

This is what you wanted right?
screen shot 2016-10-17 at 8 58 41 pm


DonLakeFlyer:
@sanderux, @AndreasAntener This is what you wanted right?

AndreasAntener:
Thanks 👍

sanderux:
for the VTOL takeoff i think it would clarify better to use this text:
Takeoff in VTOL mode, then transition to forward flight and fly to this location.

PR #4150, opened 2016-10-18
Stable v3.0

AlwaysFly1989:
Mr DonLakeFlyer:

         I found that the master had some analyze data error  when connected with pixhawk, such as "M31:attitude.roll",always have very big data. I thought that might be the “MAVLINKDecoder.cc” had some wrong.


DonLakeFlyer:
Not sure what you are doing here? Are you trying to put through a pull request for a fix you made?

PR #4153, opened 2016-10-19
Multicolored polygon

nanthony21:
I originally set out to fix a few minor issues with unwanted black lines being drawn while drawing polygons as mentioned in PR #4142 . I couldn't find a quick and simple way to do this without creating issues so I ended up making a few bigger changes.

In addition to fixing the original issues, this pull request adds a second polygon which indicates what the polygon will look like if you end the drawing, while the original polygon shows what the polygon will look like if you click to add a new point. If you wanted to go back to the original look you could simply set the new polygon to be invisible.


DonLakeFlyer:
I like it!

DonLakeFlyer:
Thanks for the pull.

PR #4161, opened 2016-10-20
Only run unit tests on pull requests

DonLakeFlyer:


DonLakeFlyer:
@dagar Does this look right?

dagar:
I think you want to use TRAVIS_PULL_REQUEST

    TRAVIS_PULL_REQUEST: The pull request number if the current job is a pull request, “false” if it’s not a pull request.

dagar:
I could have sworn we already did something like this already.

DonLakeFlyer:
Thanks

PR #4168, opened 2016-10-22
Syncing SYS_COMPANION definition with Firmware

gregd72002:
SYS_COMPANION message has more telemetry options in the firmware than exposed in QGC.

This brings them in sync.


gregd72002:
https://github.com/PX4/Firmware/blob/master/ROMFS/px4fmu_common/init.d/rcS#L554

dagar:
I'll import a current copy of the entire param and airframe metadata. The build system grabs the latest each time, so this is really only an issue for developers and other people that compile.

dagar:
Full import here - https://github.com/mavlink/qgroundcontrol/pull/4169

Issue #4156, opened 2016-10-19
Power: If UAVCAN is turned on, show UAVCAN settings by default


DonLakeFlyer:

richerlariviere:
I'm working on this issue.

PR #4176, opened 2016-10-24
Fixed 4156 Power: If UAVCAN is turned on, show UAVCAN settings by def…

richerlariviere:


DonLakeFlyer:
In looking at the implementation of the uavcan ui again as well as the parameter definition for UAVCAN_ENABLE it currently seems wrong. I think it should be changed to this:
* Show uavcan settings checkbox should always be visible
* Uavcan settings should be a dropdown with all options not a checkbox with just one option

@LorenzMeier You know more about this stuff. Agree?

DonLakeFlyer:
Fix for #4156

LorenzMeier:
Makes sense

DonLakeFlyer:
@richerlariviere Could you update this pull to match the above discussion?

DonLakeFlyer:
@richerlariviere Any update on this?

richerlariviere:
I'm short in time but I can do it. Just give me a few days.

DonLakeFlyer:
Closing since no activity. Will do myself

Issue #4178, opened 2016-10-25
MAVLink 2.0 detection doesn't work


LorenzMeier:
QGC should look at the capabilities bitmap in this message: http://mavlink.org/messages/common#AUTOPILOT_VERSION

and switch to 2.0 once it reads MAV_PROTOCOL_CAPABILITY_MAVLINK2

dogmaphobic:
Done. It will come when I push my ULog download code.

dogmaphobic:
Fixed in #4180

PR #4180, opened 2016-10-26
Mavlink log uploader

dogmaphobic:
Support for downloading/streaming Ulog files and subsequent upload to logs.px4.io.

Ulog download can be either started/stopped manually or automatically. When set to automatic, logging starts when a vehicle is armed and stops when disarmed.

Logs are collected and can be uploaded to logs.px4.io. You can also set the upload to be automatic, which will happen once the *session* ends (either an automatic session when the vehicle is disarmed or when you manually hit the *Stop Logging* button.

If automatic log upload is enabled, you also have the option to automatically delete the local copy once the log file is successfully uploaded.

No attempt at uploading will occur if either the email or URL fields are empty as these fields are necessary for the uploading form. The description field is optional to the extent that if none is given, a default *QGroundControl Session* text is used.

ULog streaming requires a new message exposed by MAVLink 2.0 While at it, I've fixed an issue in QGC that prevented it from switching to MAVLink 2.0 even if the vehicle supported it (issue #4178).

With all that said, testing was limited and in an otherwise ideal, non real-world scenario.

screen shot 2016-10-25 at 10 16 39 pm


LorenzMeier:
Great! @dogmaphobic Can you please highlight the mandatory input fields so people realise that they need to fill them? It might even take a banner at the top of the menu to highlight that.

PR #4183, opened 2016-10-27
Replacing a few placeholder icons with real ones.

dogmaphobic:
screen shot 2016-10-26 at 10 24 53 pm


DonLakeFlyer:
Thanks Gus.

PR #4186, opened 2016-10-31
Make display of the mission manager optional via firmware plugin

rjehangir:
This PR makes it optional to display the Mission Manager (planning map) via a firmware plugin feature. This is primarily intended for ArduSub, which does not support planned missions yet. It could also be used for FPV-only drones.

This is implemented by hiding the icon from the top bar if the Mission Manager is not supported. It does not remove any of the backend and is therefore a minimal change.

Since support for auto missions is slowly being developed in ArduSub, there is an added option to allow the mission manager to be displayed anyways. That's on the general settings page and only appears if offline mission editing is set to "Sub".

screen shot 2016-10-30 at 7 49 45 pm

There's also a "bug" in this, where you can select the mission manager page, then connect a vehicle that does not support it, but it will not kick you off of the page, however the icon will disappear from the top bar.

This PR should result in better usability and less confusion for ArduSub users. Please let me know if you have any questions or comments.

-Rusty

### DIFF
--- a/src/ui/preferences/GeneralSettings.qml
+++ b/src/ui/preferences/GeneralSettings.qml
@@ -198,10 +198,19 @@ QGCView {
                                 enabled:            true
                             }
                         }
+                        Row {
+                            spacing: ScreenTools.defaultFontPixelWidth
+                            visible: offlineVehicleCombo.currentText == "Sub"
+                            QGCCheckBox {
+                                text:       qsTr("Show mission manager even if not supported by firmware")
+                                checked:    QGroundControl.forceShowMissionManager
+                                onClicked:  QGroundControl.forceShowMissionManager = checked
+                            }
+                        }
                     }
                 }
                 //-----------------------------------------------------------------
-                //-- Miscelanous
+                //-- Miscellanous
                 Item {
                     width:              qgcView.width * 0.8
                     height:             miscLabel.height
@@ -209,7 +218,7 @@ QGCView {
                     anchors.horizontalCenter: parent.horizontalCenter
                     QGCLabel {
                         id:             miscLabel
-                        text:           qsTr("Miscelaneous")
+                        text:           qsTr("Miscellaneous")
                         font.family:    ScreenTools.demiboldFontFamily
                     }
                 }
--- a/src/ui/toolbar/MainToolBar.qml
+++ b/src/ui/toolbar/MainToolBar.qml
@@ -356,6 +356,7 @@ Rectangle {
             anchors.top:        parent.top
             anchors.bottom:     parent.bottom
             exclusiveGroup:     mainActionGroup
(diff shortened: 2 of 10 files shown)


src/ui/preferences/GeneralSettings.qml
```diff
@@ -198,18 +198,27 @@ QGCView {
...
+                                text:       qsTr("Show mission manager even if not supported by firmware")
+                                checked:    QGroundControl.forceShowMissionManager
+                                onClicked:  QGroundControl.forceShowMissionManager = checked
+                            }
+                        }
                     }
                 }
                 //-----------------------------------------------------------------
-                //-- Miscelanous
+                //-- Miscellanous
```
DonLakeFlyer: Closer, but still not quite: Miscellaneous :)

src/ui/preferences/GeneralSettings.qml
```diff
@@ -198,18 +198,27 @@ QGCView {
                                 enabled:            true
                             }
                         }
+                        Row {
+                            spacing: ScreenTools.defaultFontPixelWidth
+                            visible: offlineVehicleCombo.currentText == "Sub"
```
DonLakeFlyer: Why is this only for "offline" and not "online" ArduSub as well?

src/ui/toolbar/MainToolBar.qml
```diff
@@ -356,6 +356,7 @@ Rectangle {
             anchors.top:        parent.top
             anchors.bottom:     parent.bottom
             exclusiveGroup:     mainActionGroup
+            visible:            activeVehicle ? activeVehicle.supportsMissionManager || QGroundControl.forceShowMissionManager : true
```
DonLakeFlyer: This ends up being a bit wierd since force setting is global, but in reality it's only for ArduSub. Problem down the road:
* New plugin turns off supportsMissionManager
* User turns on force mission for ArduSub, leaves it on but then switches away from ArduSub as offline selection.  Force setting is still on, but not visible and will affect all plugins which return supportsMissionManager == false

src/ui/preferences/GeneralSettings.qml
rjehangir: Haha okay.

src/ui/preferences/GeneralSettings.qml
rjehangir: I chose to put the checkbox in the offline section because then it could appear based on which vehicle is selected so that it doesn't affect anything other than Sub. I could also put this in the main settings section, but it seemed inappropriate. Thoughts?

src/ui/toolbar/MainToolBar.qml
rjehangir: Ah, true. Okay. Would you prefer that I do this entirely differently? Or just make this current version more robust?

DonLakeFlyer:
@rjehangir You want to just close this right?

rjehangir:
Yes, closing.

PR #4188, opened 2016-11-01
Update MissionEditor.qml

yl5006:
do not remove before accept


yl5006:
do not remove before accept

src/MissionEditor/MissionEditor.qml
```diff
@@ -1093,7 +1093,7 @@ QGCView {
                     Layout.fillWidth:   true
                     onClicked:  {
                         syncButton.hideDropDown()
-                        _syncDropDownController.removeAll()
+//                        _syncDropDownController.removeAll()
```
DonLakeFlyer: Can you just remove it instead of commenting out?

PR #4189, opened 2016-11-01
Update DropButton.qml

yl5006:
from left to right


DonLakeFlyer:
Simple testing will show you that this change is incorrect:
screen shot 2016-11-09 at 11 05 03 am

Yes there is a bug, but it is on the dropLeft implementation not the dropRight implementation.

DonLakeFlyer:
Can you create a new pull with the dropLeft fix?

PR #4192, opened 2016-11-02
Merge pull request #4104 from DonLakeFlyer/Stable_V3.0

hcx20com:
Handle redirects in file downloads


DonLakeFlyer:
@hcx20com This change is already in Stable? Is there some problem here?

PR #4194, opened 2016-11-04
Checking for MAVLink 2.0 when first message is received.

dogmaphobic:
Following a comment form @bkueng, enabling a test for MAVLink version test when a message is first decoded.

Note that I feel this to be a temporary stop-gap fix until @DonLakeFlyer is back. That `decodedFirstPacket` variable is static and it will "stick" on subsequent connections. This will certainly not work if you have two vehicles connected or if you disconnect and connect again.

### DIFF
--- a/src/comm/MAVLinkProtocol.cc
+++ b/src/comm/MAVLinkProtocol.cc
@@ -212,19 +212,17 @@ void MAVLinkProtocol::receiveBytes(LinkInterface* link, QByteArray b)
         }
         if (decodeState == 1)
         {
-            decodedFirstPacket = true;
-
-            /*
-             * Handled in Vehicle.cc now.
-            mavlink_status_t* mavlinkStatus = mavlink_get_channel_status(mavlinkChannel);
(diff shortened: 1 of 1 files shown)


bkueng:
Yep that works for now, thanks

DonLakeFlyer:
Not sure why this change was made. This prevents reacting to mavlink 1<->2 changes on the fly, which is valid and works as far as I can tell?

bkueng:
@DonLakeFlyer let me try to explain: there are 2 mechanisms:
- on-the-fly switching: in MAVLinkProtocol.cc: this switches when the vehicle uses a different mavlink version, needed when initializing the connection (did you see that this code was previously commented?)
- capability-based switching: this is done in Vehicle.cc. Both endpoints speak mavlink 1, but the vehicle signals mavlink 2 capability, so QGC switches.

IMO both mechanisms are necessary. This PR resolves the case where a vehicle can only send/receive mavlink 2, so QGC cannot first request the capabilities and then switch.

Issue #4198, opened 2016-11-08
Correct Mavlink 2.0 switching semantics


DonLakeFlyer:
QGC will switch to sending Mavlink 2.0 packets if:
1) Vehicle specifies support for Mav 2 in AUTOPILOT_VERSION.capabilities
2) QGC receives a Mav 2 message from the Vehicle (at any time, not just first message)
3) Once QGC switches to Mav 2 for a Vehicle it will stay on Mav 2 for that Vehicle even if vehicle starts to send Mav 1 again.
4) Support for multiple vehicles on same link using different protocol version (1/2) will be allowed

Still not so sure about 4 but I'll give it a try. Not sure if possible when signing messages since I haven't looked at that part of the protocol much. Will require some channel status gymnastics which I think is possible (but maybe not).

DonLakeFlyer:
#1 is not supported

PR #4199, opened 2016-11-09
Better Mavlink 2.0 handshake code

DonLakeFlyer:
Fix for #4198:
* State of decode for first mavlink packet is kept on link by link basis instead of single value for all of QGC.
* QGC will switch to outbound Mav 2 if it gets inbound Mav 2
* QGC will swich to outbound Mav 2 if Vehicle reports Mav 2 capability
* QGC will **not** switch back to outbound Mav 1 from Mav 2 if it gets inbound Mav 1

Note: Multi-vehicle Mav 2 issues are still not fixed / figured out.


DonLakeFlyer:
@LorenzMeier @bkueng I could use review / testing on this.

bkueng:
This looks good, I'll test it tomorrow.

NaterGator:
> Multi-vehicle Mav 2 issues are still not fixed / figured out.

Is keeping a hash table of system_id (plus maybe component_id) as a key for mavlink channels that are independently spec'd to 1.0 vs 2.0 viable? Technically we should be able to support up to 256 channels through defining MAVLINK_COMM_NUM_BUFFERS.

DonLakeFlyer:
@NaterGator It's a bit more complicated than that. The issue is that on the inbound direction you don't know system id until after you parse the message. You need the mavlink channel before you parse the message, hence for inbound it can only be associated a link and not a specific vehicle.

NaterGator:
Ok, could you not run 2 parallel receive channels per link (both mavlink versions), then store the preferred outbound channel version in a hash table? Sorry I know I'm kinda hit-and-running on this...

DonLakeFlyer:
It's more that you need to do a dance with mavlink_status_t structures replacing ones back and forth, copying stuff in/out for inbound/outbound channels/vehicles. Right now things like the message sequence numbers and so forth are all out of whack with the channel == link semantic and multiple vehicles on same link. It needs to move to channel == vehicle, but that requires not using the standard channel/mavlink_status_t support that comes from the headers. I think I know how to do it, just haven't done it yet. Or gotten far enough through it identify additional gotchas.

NaterGator:
Understood; I'll be looking for distractions this weekend so I may see if I can come up with a clean PR as well.

DonLakeFlyer:
It's also a little complicated since you need to take into account / support FirmwarePlugin::adjust[Incoming|Outgoing]MavlinkMessage as well. You need to be able to do that without screwing channel status info.

DonLakeFlyer:
@LorenzMeier Given the SiK Radio Mav 2 problem it seems like this change will break a lot of things since new PX4 firmware announces mav 2 availaility in all cases. Doesn't seem like a good idea to push this through at this point?

LorenzMeier:
Yeah, this is indeed a problem. Since we can't detect the radio's Firmware version the best we can do is probably to introduce a param to not send the flag if the link is on a 3DR radio.

DonLakeFlyer:
Ok. So then I say QGC does this:
* QGC will switch to outbound Mav 2 if it gets inbound Mav 2
* QGC will **not** switch to outbound Mav 2 if Vehicle reports Mav 2 capability
* QGC will not switch back to outbound Mav 1 from Mav 2 if it gets inbound Mav 1

Issue #4051, opened 2016-09-18
Fence status indication to user


lvale:
I really like it that it downloads the fence automagically on connection. Only nagging is that there’s no indication of that. Example. I had setup a fence at my company test flying location. I now connect the vehicle but the location is different, so I can’t see that there’s a polygon fence active at my previous location. I’m not sure on how to alert the user for the existance of a fence. Perhaps an indication on the top line near the GPS status?

I believe the same happens with missions, but haven't tested yet.

PR #4205, opened 2016-11-12
Better Plan view map viewport handling

DonLakeFlyer:
* When the Plan view first comes up it will adjust the viewport on the map to include all items: Mission, Fence, Rally
* When you load items back from the vehicle (Mission, Fence, Rally) the map viewport will adjust to show those items
* New Center map: All Items support to fit all items (mission/fence/rally) into view
* Also fixed a number of fence and rally point bugs

Fix for #4051


DonLakeFlyer:
@dagar Still can't get anything to get past travis setup

dagar:
https://github.com/mavlink/qgroundcontrol/pull/4207

DonLakeFlyer:
Thanks Daniel.

PR #4208, opened 2016-11-13
Vagrant: QGC now requires SDL2

peterbarker:


dagar:
Thanks for keeping this up to date.

Issue #2937, opened 2016-03-05
Don't let messages go to toolbar dropdown during calibration


DonLakeFlyer:
You just end up with a pile of crap in there.

PR #4213, opened 2016-11-14
Filter out calibration messages

DonLakeFlyer:
Fix for #2937


dagar:
This is certainly a usability improvement, but have you or @LorenzMeier thought about what we'd need in mavlink to do it without using STATUSTEXT?

DonLakeFlyer:
There is an issue on that, related to calibration not working over WiFi. Haven't had time for it yet though.

Issue #4171, opened 2016-10-23
Altitude change UI confusing/scary


julianoes:
First of all, cool. I like the ease of use of the commands at the bottom like takeoff, change in altitude, etc. I only just discovered this yesterday but it's very convenient!

Now, to my issue: I find it scary that the default for an altitude change is 100m.

Also, the slider goes from 0 to 100m. I would propose to have a slider from e.g. -50 to 50m and default to 0m. This makes a user realize that it is a relative change and not an absolute one.

dagar:
We could also limit it from going under home.

DonLakeFlyer:
Altitude change is not relative you are specifying and absolute altitude above home if I remember correctly.

NaterGator:
> Altitude change is not relative you are specifying an absolute altitude **above home** if I remember correctly.

I don't know, that sounds relative to me. Am I misunderstanding? When I think of absolute altitudes I think of something more akin to height above MSL or the WGS84 ellipsoid.

DonLakeFlyer:
Talking about two different things:
1) The altitude you specify in the slider is home relative
2) It is not a "relative" adjustment to the existing altitude. 

For example: In Julian's example of setting slider default to 0. That would cause the vehicle to go back to home altitude. It would not cause the vehicle to just stay at the same height.

julianoes:
> 1) The altitude you specify in the slider is home relative

Well, that's what I would have guessed but that's not what's happening.

dagar:
There are 2 cases that use the altitude slider.

1. `Takeoff` and enter altitude above home (MAV_CMD_NAV_TAKEOFF)
2. `Change Altitude` in flight and enter altitude relative to current (MAV_CMD_DO_REPOSITION)

julianoes:
@dagar: Right, I'm referring to 2.

DonLakeFlyer:
Ok. I get get it now. It's a bug in change altitude. Initial implementation was for ArduPilot which is **not** "relative to current" change. Then was used for PX4 as well which **is** "realtive to current."

DonLakeFlyer:
Altitude is now correctly set to home relative altitude.

julianoes:
Thanks @DonLakeFlyer!

PR #4216, opened 2016-11-15
Correct altitude usage for Change Altitude

DonLakeFlyer:
Fix for #4171

### DIFF
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -365,7 +365,7 @@ void PX4FirmwarePlugin::guidedModeTakeoff(Vehicle* vehicle, double altitudeRel)
 
 void PX4FirmwarePlugin::guidedModeGotoLocation(Vehicle* vehicle, const QGeoCoordinate& gotoCoord)
 {
-    if (qIsNaN(vehicle->altitudeRelative()->rawValue().toDouble())) {
+    if (qIsNaN(vehicle->altitudeAMSL()->rawValue().toDouble())) {
         qgcApp()->showMessage(QStringLiteral("Unable to go to location, vehicle position not known."));
         return;
     }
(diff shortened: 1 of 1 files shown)


dagar:
@julianoes correct me if I'm wrong here, but for reposition (guided mode change altitude) I think you were asking to make the slider range from -50 - +50 meters and start centered, instead of 0 - 100 starting at 0.

@DonLakeFlyer FYI your git email address doesn't seem to be associated with your github account anymore.

DonLakeFlyer:
> I think you were asking to make the slider range from -50 - +50 meters and start centered, instead of 0 - 100 starting at 0.

That would be inconsistent with the way ArduPilot works.

dagar:
How so? I'm talking about changing altitude while already in the air and being able to decrease it. It looks like Ardupilot is just being sent an offset. Why can't it be negative?

DonLakeFlyer:
Trying to match tower/mission planner feature so users don’t get confused when moving to new GCS.

On Nov 15, 2016, at 10:09 AM, Daniel Agar <notifications@github.com<mailto:notifications@github.com>> wrote:


How so? I'm talking about changing altitude while already in the air and being able to decrease it. It looks like Ardupilot is just being sent an offset. Why can't it be negative?

dagar:
I think we're talking about slightly different things and confusing relative to home and relative to vehicle (offset).

There are basically 2 options, and I'm happy with either.

1. The slider changes altitude relative to the vehicle's home position (0 = home). If we go this route the slider should be placed at the current altitude and add an additional 100m or so on top.

2. The slider changes altitude relative to the vehicle's position (0 = current altitude). Positive or negative, and if we wanted to be clever the negative could be limited to not allow commanding something below home altitude.

Right now the slider appears as 0-100m and changes the altitude relative to the vehicle's current altitude. It only allows you to increase the altitude.

julianoes:
@dagar: I like both 1 or 2 but 1 seems to be most intuitive.

DonLakeFlyer:
@dagar Can you add an Issue for #1?

PR #4233, opened 2016-11-22
Open file and directory dialogs in foreground

birchera:
On ubuntu and mac some of the file dialogs open behind the QGC window. This is fixed in this PR by giving it the main window as parent.


DonLakeFlyer:
@birchera Can you change QGCFileDialog.cc code as well to reset parent to mainwindow if it is null?  Then this will get caught automatically. I thought that code already did that, but looks like it doesn't. Maybe lost along the way somehow.

PR #4242, opened 2016-11-24
fix for MAVLinkLogManager

bkueng:
The early return lead to data being dropped. Some of this data could have been
part of the ulog header, which lead to corrupt files.

@dogmaphobic with this ulog streaming seems to work well now. Thanks for your efforts.


DonLakeFlyer:
@dogmaphobic Can you review this. Don't know much about this code.

dogmaphobic:
> Can you review this. Don't know much about this code.

@bkueng wrote the original function. I trust him more than me :)

PR #4256, opened 2016-11-27
Plugin build refactor

DonLakeFlyer:
This allows you to create custom builds of QGC which may or may not contain all of the plugins. It also allows you to create a custom build where you can re-use the PX4 or APM plugin code as the base class for you own custom vehicle plugin and then expose only that to the system.

More docs coming on https://donlakeflyer.gitbooks.io/qgroundcontrol-user-guide/content/


DonLakeFlyer:
I'm going to redo this commit once it can do a clean run. Just realized I screwed up the commits.

dogmaphobic:
Replaced by #4257

PR #4259, opened 2016-11-27
Plugin support

dogmaphobic:
Continuing work on Don's PR #4257 
Separating APM/PX4 resource files into their own qrc files
Allowing defining what to include/exclude without having to change QGC's main project file (configuration passed through `CONFIG` defined either from the command line or through your `user_config.pri` file)
Started working on a "custom" configuration so you can add your own plugin, again without having to change QGC's main project file (or maintain it).


DonLakeFlyer:
I would switch the core app option model to match the Factory/Plugin model from FirmwarePlugin and make it work the same way in the qgroundcontrol.pro file. That way we don't have two different ways to do plugins. Something like this: QGCUIPlugin and QGCUIPluginFactory (no "I" prefix, we don't use that anywhere else). Then you use/override it exactly the same way as the Firmware plugins.

Then when we finally get to dynamic, non-statically link plugins we'll switch everything over to a new thing at the same time.

DonLakeFlyer:
Here is my proposal (still typing not done yet):
* Change to use the same model (code and .pro support) as FirmwarePlugin for the options plugin. I think QGCAppPlugin, QGCAppPluginFactory naming should work. It all works with respect to override/replace just like FirmwarePlugin.
* Change to an isCapable bit style approach to turn things on/off to make FirmwarePlugin style
* Changed collapseSetup to a "hide settings" bit.
* Modify SetupView such that the first buttons/pages from from QGCAppPlugin::setupPages method which works just like AutoPilotPlugin::vehicleComponents.
* Default implementation of QGCAppPlugin::setupPages returns standard Vehicle Summary and Firmware Upgrade. You can then override to add/remove pages.

dogmaphobic:
Some observations:

>Change to use the same model (code and .pro support) as FirmwarePlugin for the options plugin. I think QGCAppPlugin, QGCAppPluginFactory naming should work. It all works with respect to override/replace just like FirmwarePlugin.

I’m not sure that’s worth the effort. We’re not replacing an existing architecture with another. We’re adding one where none exists unless you add something. The FirmwarePlugin and its *factory* make sense because not only they do exist already (and you might want to enhance or replace it), they also provide multiple, different instances depending on something (the connecting firmware in this case). For the case of a generic, singleton *custom* plugin that exists only to change some runtime behavior, it’s a whole lot simpler to simply create the singleton, exposed custom instance if found or use the base class defaults if it doesn’t. This avoids the whole *factory* part and eliminates an unnecessary level of complexity. I don’t really see a need to exclude existing files from the build as it is the case with the Firmware Plugin.

>Change to an isCapable bit style approach to turn things on/off to make FirmwarePlugin style

Any reason other than making them look alike? It’s a lot of extra code, enums with bitmap masks, function arguments, testing bits, etc. where a simple `isThereThis()` or `isThereThat()` would do with a simple, no argument call and a simple register answer. It actually takes more bytes to pass an argument, encode/decode and respond than it is saved by the bitmap storage. An argument could be made that it would make adding new features dynamically easier but as either way will require a rebuild of the code any way, I’m not seeing a benefit. With all that said, if you think it should look alike for the sake of consistency, I’m fine with that too.

>Default implementation of QGCAppPlugin::setupPages returns standard Vehicle Summary and Firmware Upgrade. You can then override to add/remove pages.

The idea is to provide settings, which exist before anything is connected. These settings go in the Settings Menu. I haven’t added the code in the settings panel yet (that’s why you have only seen them under the Vehicle Setup panel). They only show up under Vehicle Setup if and when you “hide settings” and collapse them all into one menu. QGCAppPlugin works at the main application level, not at the vehicle level. The overall idea is correct. The plugin should return a list of settings it provides (0 or more) but these are “settings” not “vehicle setup”. Additional *Vehicle Setup* pages, if any, would be added using the existing methods exposed by the `vehicleComponents` method.

So far I’ve been mostly experimenting with this. Ultimately, the settings should come from a same type of list, provided by some interface, This interface can be overridden so it can exclude existing settings and/or add new ones. Another interface determines if you want them in a separate settings menu as it is the case today, or combine them with the vehicle setup panel if that makes more sense for your particular (custom) application.

> I wouldn't add a mainViewIsMap option. I would add support for a Plan view setting to save what view it starts with.

That’s what it does. It defines the default option when you first run the app. Once you change it, it’s saved just like it is today. That is, the change is from:

`QGroundControl.loadBoolGlobalSetting(_mainIsMapKey,  true)`

To:

`QGroundControl.loadBoolGlobalSetting(_mainIsMapKey,  QGroundControl.mainViewIsMap)`

The `true` default is now defined by a method which can be overridden.

>Add support to QGCApplication to call into a QGCAppPlugin method which adjusts settings at boot time (right after QGCApplication loads them). This allows the app plugin to change the defaults on anything it want like like Plan view startup view. Use a similar pattern for other places like this where you could add a useful feature behind a setting instead of a bit twiddle off thingy. Not quite sure but you may be able to setup video this was as well.

That’s exactly what it is doing right now.

DonLakeFlyer:
For QGCAppPlugin::setupPages:
I was just trying to simplify things so you don't really need to do the collapse thing. Just hide settings, then expose settings using QGCAppPlugin::setupPages. If you want to support collapse then you will need two methods, QGCAppPlugin::settingsPages and QGCAppPlugin::setupPages that expose the list of pages.

With respect to Plugin and PluginFactory mechanism:
It creates the mechanism to override and re-use the normal QGC code in the same manner as the other plugins. For example: The standard build in implementation of QGCAppPlugin/Factory sends out the normal settings pages from QGCAppPlugin::settingsPages and static setup pages (Vehicle Summary, Firmware Upgrade) from QGCAppPlugin::settingsPages methods. To create a custom app plugin you create your own QGCAppPluginFactory and your own QGCAppPlugin using the standard version as your base class.  Then decide what you want to modify by overriding the methods. This allows you to add/remove pages from both settings and setup views. QGCApplication then calls the factory to get it. Exactly the same override/replace machanism as FirmwarePlugin. The fact that there aren't multiple variants doesn't really matter.

Settings:
This issue with the way you are doing it is that it requires code changes in mainline code to support. You showed on example above, here is another:
```
+    //-- Check if core plugin defines its own video requirements
 +    if(qgcApp()->qgcOptions()->definesVideo()) {
 +        if(qgcApp()->qgcOptions()->videoUDPPort()) {
 +            setUdpPort(qgcApp()->qgcOptions()->videoUDPPort());
 +            setVideoSource(kUDPStream);
 +        } else {
 +            setVideoSource(kRTSPStream);
 +            setRtspURL(qgcApp()->qgcOptions()->videoRSTPUrl());
 +        }
 +    } else {
 +        setUdpPort(settings.value(kVideoUDPPortKey, 5600).toUInt());
 +        setRtspURL(settings.value(kVideoRTSPUrlKey, "rtsp://192.168.42.1:554/live").toString()); //-- Example RTSP URL
 +    }
```

A mechanism to allow tweaking of settings through a plugin hook in the boot sequence provides access to tweak all settings with no additional mainline code changes. Both from a standpoint of setting a new default, or forcing to a specific value always. It would allow you to do the video settings changes above for example all in the plugin with no change to mainline code. Same thing for the mainIsMap support. No changes to mainline code. A 3rd party QGCAppPlugin author is not going to have access to changing the mainline code like you have. It needs to be able to work all from inside the plugin.

isCapable:
Not really important. I found that I needed to test multiple bits in many places and all the multiple function calls isThis, isThat ended up just as bad. It also matched the mavlink capability support bit thingy.

dogmaphobic:
In the end, there was a single instance of something using QSettings (`mainViewIsMap`), which I changed it, dealing with QSettings directly when initializing the app.

I also removed the methods that determine which setting is available and instead, completely replaced the setting list from scratch within the plugin.

I have yet to figure out how to deal with the Vehicle Setup pages. They won't fit the same model as settings.

Removed references to dynamic plugins. I created a separate issue for that (#4263) and closed it.

I will merge that when it passes CI and continue on separate PRs.

PR #4264, opened 2016-11-29
AutoConnect for UDP links

AlexeyBulatov:
This PR allows to set autoconnect for selected UDP link  that helps if you use QGC for outgoing connection.
Button indicator shows status of autoconnect configuration.

### DIFF
--- a/src/comm/LinkConfiguration.h
+++ b/src/comm/LinkConfiguration.h
@@ -79,7 +79,7 @@ public:
     /*!
      *
      * Is this an Auto Connect configuration?
-     * @return True if this is an Auto Connect configuration (connects automatically at boot time).
+     * @return True if this is an Auto Connect configuration (connects automatically).
      */
     bool isAutoConnect() { return _autoConnect; }
 
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -469,6 +469,13 @@ void LinkManager::_updateAutoConnectLinks(void)
         return;
     }
 
+    for (int i=0; i<_linkConfigurations.count(); i++) {
+        LinkConfiguration* conf = _linkConfigurations.value<LinkConfiguration*>(i);
+        if (!conf->link() && conf->isAutoConnect()) {
+            createConnectedLink(conf);
+        }
+    }
+
     // Re-add UDP if we need to
     bool foundUDP = false;
     for (int i=0; i<_links.count(); i++) {
--- a/src/comm/LinkManager.h
+++ b/src/comm/LinkManager.h
@@ -117,8 +117,8 @@ public:
     /// Load list of link configurations from disk
     void loadLinkConfigurationList();
 
-    /// Save list of link configurations from disk
-    void saveLinkConfigurationList();
+    /// Save list of link configurations to disk
+    Q_INVOKABLE void saveLinkConfigurationList();
 
     /// Suspend automatic confguration updates (during link maintenance for instance)
     void suspendConfigurationUpdates(bool suspend);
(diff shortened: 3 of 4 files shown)


DonLakeFlyer:
Thanks. A useful change. But there are problems:
* You should set autoconnect when you add/edit the config using a checkbox. It shouldn't be on the outside on the link list with the add/edit buttons.
* Although autoconnect links are picked up dynamically if the user turns them on. They are not disconnected dynamically if the user turns them off. They should disconnect the socket if there is no vehicle and the user has disconnected the vehicle from it.
* Autoconnect should only be allowed for UDP links since other link types will need more sophisticated code to check whether the port exists before it tries to connect. Otherwise it will through continuous errors when the device is not connected.

DonLakeFlyer:
@dogmaphobic Thinking about this a little more. Can't you edit the default autoconnect udp link? If so, can't you just add target hosts to the default udp link and then this all works for free?

dogmaphobic:
The only reason the *active* connection is disabled from editing is to avoid changing something that would disrupt the connection. There is no real technical reason other than preventing users from getting in trouble. For instance, if you were to change the connecting port in the middle of a connection, all hell would break loose.

If you know what you are doing though, it's perfectly safe to edit it.

DonLakeFlyer:
Ok, then I think we go with this PR with fixes.

AlexeyBulatov:
Hello.
- If set autoconnect in add/edit config then you can't change its state, because you can't edit connected links.
- It will work as existing "UDPAutoconnect", but it reconnects a link instead of creating a new copy.
- Yeah, button for "autoconnect" enabled only for UDP links.

DonLakeFlyer:
> If set autoconnect in add/edit config then you can't change its state, because you can't edit connected links.

Hmm. That is an issue :) I would change the control to a checkbox then. IndicatorButton tends to only be used to setup buttons on the setup view.

> It will work as existing "UDPAutoconnect", but it reconnects a link instead of creating a new copy.

Not sure what you mean here?

AlexeyBulatov:
Something like that?
screenshot from 2016-12-05 17-52-47
When I said about existing "Default UDP Link" I meant that if one pushed “Disconnect”, it created a new copy of link and therefore a new button. 
It seems unnecessary or at least counterintuitive, because it creates a pile:
screenshot from 2016-12-05 17-43-33
My PR takes another approach: instead of creating a new copy it simply reconnect to existing one (if autoconnect button is on).

DonLakeFlyer:
Ok. you're running into bugs in QGC. 

@dogmaphobic Is there any reason to show autoconnect links here? Seems like they just confuse everyone. And then if people try to disconnect them, strange things happen. I think we should hide autoconnect links.

Now back to this pull:
* I think this should be called something other than "Autoconnect" since that term is already used for a different QGC feature.
* I think it should be something like "Connect at QGC start"
* I think the setting should be inside the Edit panel, not on the main panel.
* If you disconnect a "Connect at QGC start" setting it does not reconnect immediately. It just disconnects and stays disconnected. This allows you to edit the settings again and adjust "Connect at start".
* It should only be supported by UDP links.

dogmaphobic:
> Is there any reason to show autoconnect links here?

Not that I can think of. The list was there before auto connect was in place. I guess the issue never came up. On the other hand, the example above shows that some times it can be useful :)

DonLakeFlyer:
> On the other hand, the example above shows that some times it can be useful :)

Huh, how?

dogmaphobic:
I was being sarcastic. The multiple buttons exacerbated the issue above.

DonLakeFlyer:
Ah, ok. Do you think my proposal for the pull above makes sense? 

I'm also going to clean up the comm links pages to make it more clear what you can/can't do and why.

AlexeyBulatov:
It seems to me that reconnection comes in handy in a scenario when one disconnects in QGC and reboots a vehicle. Otherwise you will have to constantly press the “Connect” button if you select “Connect at QGC start” beforehand. I do not find the latter convenient, either.
So, for example it’s not necessary to make an option for serial connections in Comm Links, because it already always tries to connect.
If you want I may create another indicator button.

PR #4266, opened 2016-11-30
Add Intel Aero RTF in PX4 airframe meta

lucasdemarchi:


dagar:
Was this generated by PX4 or added manually? If manual, could you instead generate a copy of airframes.xml in PX4 (top level of any build folder) and commit it here. Otherwise it's going to be overwritten in the future and each QGC build.

lucasdemarchi:
@dagar done, thanks.   Lot's of changes there that I have no idea... but the Intel Aero RTF is there as well.

dagar:
Thanks, merging.

PR #4276, opened 2016-12-02
Handle delayed RTSP video connection.

dogmaphobic:
RTSP will try to connect to the server. If it cannot connect,  it will simply give up and never try again. I added code to keep trying to make a connection to the server first. Only when I get a connection, will I actually start the stream. 

This solves the problem where if QGC was started before it had a network connection, or before the video source had started, it would not connect to the stream.


dogmaphobic:
It looks like all OS X hosts at travis are down. I gave up waiting. I needed this in to continue but I will leave it to tomorrow.

PR #4277, opened 2016-12-02
when send missions ， rotate Image

yl5006:


DonLakeFlyer:
Thanks for pointing out.  Not quite right, but I think i found the right fix.

PR #4286, opened 2016-12-04
Switch Bluetooth connection from a generic Rfcomm protocol...

dogmaphobic:
... to a specfic SerialPort service.

This is a fix for Android Bluetooth links.

When the Bluetooth link was first implemented, the only way I got a comm link to work with the TBS Crossfire was setting the socket as a generic *Rfcomm* protocol. Something changed after Android 6.x that made this no longer work. Switching to a specific *SerialPort* protocol now enables a socket connection.

Note that I could only test this with Android 6.x and 7.x. Android 5.x no longer pairs with the Crossfire neither does Mac OS. I do not have anything else with Bluetooth so it would be good to know how it behaves with other Bluetooth devices (other than the Crossfire) and other platforms.

Testing was done using Crossfire firmware version 1.44.

### DIFF
--- a/src/comm/BluetoothLink.cc
+++ b/src/comm/BluetoothLink.cc
@@ -150,7 +150,7 @@ bool BluetoothLink::_hardwareConnect()
     _discoveryAgent->start();
 #else
     _createSocket();
-    _targetSocket->connectToService(QBluetoothAddress(_config->device().address), QBluetoothUuid(QBluetoothUuid::Rfcomm));
+    _targetSocket->connectToService(QBluetoothAddress(_config->device().address), QBluetoothUuid(QBluetoothUuid::SerialPort));
 #endif
     return true;
 }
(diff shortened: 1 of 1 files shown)


dogmaphobic:
> I do not have anything else with Bluetooth

A note about iOS devices: The Bluetooth comm layer in QGC follows the Bluetooth 2.x specifications, which are only supported by iOS to Apple *approved* devices, which the Crossfire isn't. Therefore, it will not work with iOS. The Crossfire does support the 4.x protocol (Low Energy), which could in theory be used by iOS but it will require a completely different comm layer to handle it. QGC does not support BLE communications.

DonLakeFlyer:
I've got a bluetooth thingy. I'll give it try.

dogmaphobic:
Are you grabbing this branch for the test or should I merge it?

DonLakeFlyer:
I won't be able to test till tomorrow

dogmaphobic:
Ok, merging and closing then.

PR #4294, opened 2016-12-07
RTSP video crash，use  old  function

yl5006:


dogmaphobic:
What crashed? This was working fine.

yl5006:
software  crash ，May my software version not correct。 but when I start connect rtsp video .software  crash

dogmaphobic:
Do you have the error output when it crashes? Are you using GStreamer 1.5.2?

yl5006:
gstreamer-1.0-x86-1.8.1 ,no out put

dogmaphobic:
QGC only supports 1.5.2. There should be some output unless you are running Windows. The output would indicate if some library was missing or some pipeline error, which could indicate what to do to make it work.

PR #4301, opened 2016-12-07
Toolbar: Remove global property usage

DonLakeFlyer:
* Make MainToolbarIndicators reusable by other views
* Change to font based ratio sizing
* Remove all the up/down/sideways references to global properties


DonLakeFlyer:
* Tested on all iOS simulators
* Tested on my iPhone 5s, and small Asus Android tablet

PR #4302, opened 2016-12-07
Adding a crash guard for when a command resend timeout occurs with an…

dogmaphobic:
… empty command queue.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -1838,6 +1838,12 @@ void Vehicle::sendMavCommand(int component, MAV_CMD command, bool showError, flo
 
 void Vehicle::_sendMavCommandAgain(void)
 {
+    if(!_mavCommandQueue.size()) {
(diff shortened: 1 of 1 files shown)


dogmaphobic:
@DonLakeFlyer I'm pushing this in as I need it elsewhere. I'm not sure what is going on. I didn't have time to take a thorough look at the code that starts the timer. In situations like this, I make sure to set the timer to single shot and only start on demand so it doesn't repeat when you don't expect.

Something like (off my own code):

```
connect(&_timer, &QTimer::timeout, this, &MyClass::_stateManager);
_timer.setSingleShot(true);
```

That way, only when I need it to go off I execute `_timer.start()` and it runs only once.

DonLakeFlyer:
This is already fixed in master. Can't remember which pull.

DonLakeFlyer:
Can you pull this out now?

dogmaphobic:
Technically it doesn't hurt to have a crash guard in there just in case.

PR #4310, opened 2016-12-08
Fix most (all?) QML font rendering glitches on Windows

NaterGator:
This has always driven me crazy. This seems to be a general-case app-wide fix of the font rendering problems no matter what base font size is selected. Presumably this is because the defaultFontPixelWidth/defaultFontPixelHeight are used throughout the QML codebase to define margins (usually divided by 2). Just ensuring that those numbers are even and hence divide by two into an integer prevents the transforms applied to QML views from corrupting Windows font rendering. 

Before:
image

After:
image

### DIFF
--- a/src/QmlControls/ScreenTools.qml
+++ b/src/QmlControls/ScreenTools.qml
@@ -87,8 +87,8 @@ Item {
     function _setBasePointSize(pointSize) {
         _textMeasure.font.pointSize = pointSize
         defaultFontPointSize    = pointSize
-        defaultFontPixelHeight  = _textMeasure.fontHeight
-        defaultFontPixelWidth   = _textMeasure.fontWidth
+        defaultFontPixelHeight  = Math.round(_textMeasure.fontHeight/2.0)*2
+        defaultFontPixelWidth   = Math.round(_textMeasure.fontWidth/2.0)*2
         smallFontPointSize      = defaultFontPointSize  * _screenTools.smallFontPointRatio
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Is the problem when a label ends up positioned on non integer positions? Say 3.5 instead of 3? Changing ScreenTools won't really catch that. Only some of them. Lots of places that multiply font sizes by ratios other than 0.5.

NaterGator:
Don, I had the same thought but it seems like margins disproportionately affect the font rendering quality, and I struggled to find a single instance that still rendered poorly with this fix in. I tried a variety of base point sizes and genuinely couldn't find a single font glitch anywhere. 

Would be interested in seeing feedback from other users. I thought about applying this to only windows, but it strikes me as brittle to apply potential layout altering changes to only one OS.

DonLakeFlyer:
I'd rather clean up the positioning in the places it is used than make a change like this. It's too mysterious. I did a search and cleaning up positioning to fixed integers is not that massive. Then we can stay fully clean moving forward.

Let's do this:
1) We'll let this go through for 3.1
2) Put in an issue to clean up the positioning on 3.2 for real referencing this pull once to back out the change once that is done

NaterGator:
Sounds good. 

I wonder if there is a QMLish way to walk the QML tree and asserts/warns on non-integer positioning properties. I'll look into it.

DonLakeFlyer:
It's kind of stupid that Qml positioning values are "real" but if you don't use integers things blow up!

PR #4315, opened 2016-12-10
Null pointer issue

rjehangir:
This PR fixes an issue that causes QGC to crash when exiting when a joystick is connected. I'm not confident on the series of events when QGC shuts down, but somehow the function to send a MANUAL_CONTROL message is being called after the vehicle's MAVlink connection ceases to exist. This causes a null point exception and a crash.

This code adds one line to check to see if the pointer in question is null.

I'm open to other fixes if there's a better way to do this.

This addresses ArduSub Issue: https://github.com/bluerobotics/ardusub/issues/56

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -979,6 +979,10 @@ void UAS::setExternalControlSetpoint(float roll, float pitch, float yaw, float t
         return;
     }
 
+    if (!_vehicle->priorityLink()) {
+        return;
+    }
+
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@rjehangir This doesn't seem right. A Vehicle should always have a priority link available. This smells like a vehicle shutdown bug which is still lurking in there. This fix masks the problem, but it is likely to pop up someplace else.

PR #4316, opened 2016-12-11
CF2 Support

dennisss:
- Adding support for flashing the Crazyflie 2
- Also adds a basic GUI for reconfiguring the radio on it
- So far supported via reflashing with PX4 (http://dev.px4.io/hardware-crazyflie2.html)

@dagar  I set up the firmware paths to point to px4-travis. Can you set it up to upload the build for the `crazyflie_default` target?

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -806,6 +806,8 @@ HEADERS+= \
     src/AutoPilotPlugins/Common/MixersComponent.h \
     src/AutoPilotPlugins/Common/MotorComponent.h \
     src/AutoPilotPlugins/Common/RadioComponentController.h \
+    src/AutoPilotPlugins/Common/SyslinkComponent.h \
+    src/AutoPilotPlugins/Common/SyslinkComponentController.h \
     src/AutoPilotPlugins/Generic/GenericAutoPilotPlugin.h \
     src/FirmwarePlugin/CameraMetaData.h \
     src/FirmwarePlugin/FirmwarePlugin.h \
@@ -829,6 +831,8 @@ SOURCES += \
     src/AutoPilotPlugins/Common/MixersComponent.cc \
     src/AutoPilotPlugins/Common/MotorComponent.cc \
     src/AutoPilotPlugins/Common/RadioComponentController.cc \
+    src/AutoPilotPlugins/Common/SyslinkComponent.cc \
+    src/AutoPilotPlugins/Common/SyslinkComponentController.cc \
     src/AutoPilotPlugins/Generic/GenericAutoPilotPlugin.cc \
     src/FirmwarePlugin/CameraMetaData.cc \
     src/FirmwarePlugin/FirmwarePlugin.cc \
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -24,6 +24,7 @@
         <file alias="DebugWindow.qml">src/ui/preferences/DebugWindow.qml</file>
         <file alias="ESP8266Component.qml">src/AutoPilotPlugins/Common/ESP8266Component.qml</file>
         <file alias="ESP8266ComponentSummary.qml">src/AutoPilotPlugins/Common/ESP8266ComponentSummary.qml</file>
+        <file alias="SyslinkComponent.qml">src/AutoPilotPlugins/Common/SyslinkComponent.qml</file>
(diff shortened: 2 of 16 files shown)


dagar:
I'll merge once the build passes - https://github.com/PX4/Firmware/pull/6030

LorenzMeier:
@dennisss Can you share a screenshot?

dennisss:
@LorenzMeier 

screen shot 2016-12-11 at 1 18 01 pm

DonLakeFlyer:
I'm going to have to look at this. I think there were some other points which haven't been cleaned up which require Qml changes if you add a new board type.

Also what is a "Syslink"? The UI is this pull is pretty far away from my standards of "goodness".

dennisss:
I tried to change everything that I saw for adding the new board, and it is functional in my testing for  online Firmware upgrades. 

The Crazyflie 2.0 has a co-processor (NRF51822) that handles wireless connectivity among some other things. "Syslink" is the official name of the protocol used to talk to and configure it. I've kept that name in the PX4 Firmware port for this device. That UI is only applicable when used with a Crazyflie drone.

I by no means want to pretend that I'm a good UI designer :)

Suggested changes would be much appreciated.

DonLakeFlyer:
@dogmaphobic Could you talk with @dennisss about the WiFi thing. Seems similar to PixRacer WiFi setup page thingy, so should probably be setup/handled the same way to only show up when available?

DonLakeFlyer:
@dennisss This might be another place depending on how you hooked up the board type:
https://github.com/mavlink/qgroundcontrol/blob/master/src/VehicleSetup/FirmwareUpgrade.qml#L100

Haven't had time to go through the changes yet to truly determine.

dogmaphobic:
> Could you talk with @dennisss about the WiFi thing. Seems similar to PixRacer WiFi setup page thingy, so should probably be setup/handled the same way to only show up when available?

It looks like that's exactly how he did it. It behaves the same as the WiFi Bridge module. It only shows up if you have it, using the same code the WiFi Bridge module uses.

dennisss:
@DonLakeFlyer I think I covered that line here:  https://github.com/mavlink/qgroundcontrol/pull/4316/files#diff-e84c60c54dfc39dd7e7b3c0dc0bc109eR100. I used this old commit as a reference for adding the board support: https://github.com/mavlink/qgroundcontrol/commit/227f1c576b849da97946f407cd772276355032f4

DonLakeFlyer:
Ok, other than the UI things look good. Can you match the style of the PixRacer WiFi page:
screen shot 2016-12-23 at 4 07 51 am

DonLakeFlyer:
@dennisss The space this pull was in went through a bunch of re-work. Sorry about that. Can you rebase and fix it up?

LorenzMeier:
@dennisss Any chance for a rebase?

dennisss:
Rebased and UI updated. Should look more QGC-like now.

screen shot 2017-02-26 at 9 55 14 am

LorenzMeier:
Thanks! @DonLakeFlyer Given this is for very particular hardware I think this looks reasonable - or do you see further improvements missing?

DonLakeFlyer:
@dennisss Sorry I missed you updates till now! Can you rebase and then I'll take a look.

LorenzMeier:
@dennisss Do you have a chance to rebase?

dennisss:
@LorenzMeier sorry for the delay. Rebased now and retested. Everything seems to still be working.

DonLakeFlyer:
Doesn't USBBoardInfo.json need to be updated as well for auto connect to these new boards?

DonLakeFlyer:
Did you try whether autoconnect was working with the new board?

dennisss:
@DonLakeFlyer , it was working without editing the json file as I didn't know that existed until now. I've added the proper config for it there as well.

It looks like it worked for me because the the latest PX4/Bootloader assigns 22 as the product id for the crazyflie bootloader, but apparently that was also the product id of the old FMU v2 bootloader which is still supported in QGC. From what I can tell this conflict doesn't immediately effect either board as there is no conflict in the board id used for firmware updates.

DonLakeFlyer:
Thanks. I haven't been in this part of the code for a while. Couldn't remember if the .json was absolutely needed or not.   I'll have to double-check a little about the duplicate ids in the json to make sure that doesn't create some strangness.

DonLakeFlyer:
Ok, checked that duplicate ids in json would be ok. It's fine. Thanks @dennisss

PR #4327, opened 2016-12-19
Change airplane compass icon to generic arrow icon

rjehangir:
This PR changes the compass icon to an arrow instead of an airplane. This shape better suites the other firmware types besides plane including Copter, Rover, and Sub. The arrow design is loosely based on the compass icon from other digital compasses, such as Garmin watches.

Here are a few screenshots of how it looks at various angles:

screen shot 2016-12-19 at 8 12 42 am

screen shot 2016-12-19 at 8 13 05 am

This is currently implemented across all firmware types, however we can make it a firmware plugin feature if desired. Please let me know if you'd like me to do that!

### DIFF
--- a/qgcresources.qrc
+++ b/qgcresources.qrc
@@ -109,6 +109,8 @@
         <file alias="scale_endLight.png">src/FlightMap/Images/scale_endLight.png</file>
         <file alias="airplaneOutline.svg">src/FlightMap/Images/airplaneOutline.svg</file>
         <file alias="airplaneOpaque.svg">src/FlightMap/Images/airplaneOpaque.svg</file>
+        <file alias="vehicleArrowOutline.svg">src/FlightMap/Images/vehicleArrowOutline.svg</file>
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
Can you change the vehicle icon in the map as well?

DonLakeFlyer:
Maybe make this be VehicleArrow instead of CompassInstrumentArrow and use it in Map as well?

rjehangir:
Okay, I have made this the vehicle icon on the map as well. Note that I created two new icons to match the colors and the opaque/outline versions that are currently used. Here's what it looks like on the map:

screen shot 2016-12-22 at 8 51 59 am

dogmaphobic:
What is this? :)

screen shot 2016-12-22 at 12 10 26 pm

rjehangir:
I think the mock link has a very poor performing GPS ;-)

DonLakeFlyer:
> I think the mock link has a very poor performing GPS ;-)

Yup. I haven't fixed that yet. Noticed it before I left. MockLink is sending a no hdop value (UINT16_MAX or something like that) but the toolbar doesn't quite repect that yet.  I'll fix it soon.

Issue #4320, opened 2016-12-15
Tile request error


yzzir:
On Kubuntu 16.04, when I launch QGC i get errors in terminal:

tile request error   "problem with tile image"

One thing to note, I am behind a company proxy, but I have the ca imported to trust the proxy. And i can access the internet with firefox, curl, git, wget.

DonLakeFlyer:
@dogmaphobic Is the tile pulling code set up to work from behind a proxy. It doesn't come for free. See QGCFileDownload for proxy setup stuff.

PR #4331, opened 2016-12-22
Handling proxies when fetching tiles.

dogmaphobic:
In response to #4320. I added proxy handling elsewhere but I missed one spot. Note that I can't test this as I don't have any proxy to test with.

### DIFF
--- a/src/QtLocationPlugin/QGeoMapReplyQGC.cpp
+++ b/src/QtLocationPlugin/QGeoMapReplyQGC.cpp
@@ -152,11 +152,16 @@ QGeoTiledMapReplyQGC::cacheError(QGCMapTask::TaskType type, QString /*errorStrin
         qWarning() << "QGeoTiledMapReplyQGC::cacheError() for wrong task";
     }
     //-- Tile not in cache. Get it off the Internet.
+    QNetworkProxy proxy = _networkManager->proxy();
+    QNetworkProxy tProxy;
+    tProxy.setType(QNetworkProxy::DefaultProxy);
+    _networkManager->setProxy(tProxy);
     _reply = _networkManager->get(_request);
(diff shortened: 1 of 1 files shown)


fiji3119:
@dogmaphobic  I tried your fix but did not work. I am behind proxy but can access internet via browser, git ,etc. I get "tile request error "Problem with tile image". My OS is ubuntu 16.04.

DonLakeFlyer:
> I get "tile request error "Problem with tile image"

I can't find that text anywhere in the source. Is that exactly what comes out on the console?

DonLakeFlyer:
Can you flash firmware to your board? That also does a download using proxy code. Can you see if that works?

fiji3119:
The message "tile request ..." is produced by Qt Location lib. I can  flash released firmware with no problem. I also installed QGC in Windows to test  - works fine even though both computers are on same network. Confused...

NaterGator:
@fiji3119 Can you clear your tile cache? That error is only present in the QGeoFileTileCache class.

fiji3119:
I cleared .cache/QGCMapCache300 but issue still exists.

DonLakeFlyer:
So you're saying on a Windows build tiles load. But on a Linux build they don't. And these are both the same version of QGC?

NaterGator:
Can you check these locations: https://code.qt.io/cgit/qt/qtlocation.git/tree/src/location/maps/qabstractgeotilecache.cpp?h=v5.6.2#n115

fiji3119:
@DonLakeFlyer Yes. On linux, it may have to do with proxy or certificates. My co-worker claims it works when no proxy is used on network (example from home). I am building QGC v3.1.0 with Qt5.6.1.

PR #4337, opened 2016-12-23
the message handleAltitude

yl5006:
altitudeAMSL and altitudeRelative do not need divide 1000，the unit is meter and float format。 can you use _handleGlobalPositionInt message for altitudeAMSL and altitudeRelative？ because _handleGlobalPositionInt is default 5Hz and _handleAltitude is 1Hz.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -560,8 +560,8 @@ void Vehicle::_handleAltitude(mavlink_message_t& message)
 
     _useAltitudeForAltitude = true;
     _useGpsRawIntForAltitude = false;
-    _altitudeRelativeFact.setRawValue(altitude.altitude_relative / 1000.0);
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Thanks for the fix.

@LorenzMeier This is per our conversation on how to get altitude values when you have no GPS. The current priority of messages where altitude is pulled from is (from low to high):
* GPS_RAW_INT
* GLOBAL_POSITION_INT
* ALTITUDE

Or in other words:
* If you get an ALTITUDE message it will always use altitude from that message and stop using the values from either GLOBAL_POSITION_INT and GPS_RAW_INT.
* If you get GLOBAL_POSITION_INT messages but no ALTITUDE messaages it will use those values for altitude instead of GPS_RAW_INT.

Given the update rates, should the priority instead be:
* GPS_RAW_INT
* ALTITUDE
* GLOBAL_POSITION_INT

Issue #4343, opened 2016-12-23
QGC crashes when sending Vehicle MAVLink message on a link! (at unexpected times)


samacumen:
I would like to bring to your notice that QGC crashes at unexpected times when using Serial/or multiple links. I recently figured out why and where this happens and the root cause. The culprit is the way we have written the function priorityLink() and its usage while sending mavlink messages.

If AutoConnect UDP is used, there will be at least one link in the **Vehicle::_links** Qlist. When the serial USB is connected, there is this serial link. What this logic of priorityLink() does is it gives Serial the priority and if it is not found, the first link in the _links list is returned.

```
/// @return Direct usb connection link to board if one, NULL if none
LinkInterface* Vehicle::priorityLink(void)
{
    foreach (LinkInterface* link, _links) {
        if (link->isConnected()) {
            SerialLink* pSerialLink = qobject_cast<SerialLink*>(link);
	    if (pSerialLink) {
                return link;
            }
        }
    }

    return _links.count() ? _links[0] : NULL;
}
```

**Two problems:**

1) return _links.count() ? _links[0] : NULL; 
When no vehicle is connected, we disable the QGC GUI - so we are safe. But during certain heavy processing, like sending a big mission to the vehicle, if the link (serial/other) is disconnected - the above could/is certainly NULL. Consequences? Crash of QGC, as we use prioritylink()->mavlinkChannel() while sending each message. Access on NULL instance!

2) If there are more than 1 links (apart from serial, say a UDP and a TCP i.e. Serial, UDP, TCP in the list) for the same vehicle (which is allowed) - which link must be chosen if Serial is not available? The first in the list after Serial, i.e. UDP -> does not make much sense. This indicates the TCP link will never be used.

Somehow it feels to me that the implementation of PriorityLink() is incomplete. Suggest to at least place a NULL check apart from link not being connected while sending ANY Mavlink message to the vehicle. 

PS: Also, notice that isConnected() for UDP link will never be 'false' (unless user disconnects manually) and isConnected() for Serial will usually not occur (as the link object is lost immediately on a disconnect).

@LorenzMeier , @DonLakeFlyer

LorenzMeier:
@dogmaphobic Are you able to come up with a patch? This might fix a plethora of issues / crashes.

dogmaphobic:
> Are you able to come up with a patch? This might fix a plethora of issues / crashes.

[...]

PR #4349, opened 2016-12-26
LinkInterface/LinkConfiguration now reference counter using QSharedPointer

DonLakeFlyer:
This prevents crashes where code was referencing deleted links. Actual crash is not easy to repro. There is 20 millisecond window between MultiVehicleManager _deleteVehiclePhase1 and _deleteVehiclePhase2 where the link is deleted but Vehicle may still get a timer fired in between there from things like MissionManager or ParameterManager which will then reference the deleted link. 

Verified fix by expanding the timeout between phase1 and phase2 and then pulling plug on usb during parameter load. Before change: boom! After change: good, link reference counting keeps link alive until Vehicle destructor.

Fix for #4343


DonLakeFlyer:
Needless to say this was a pretty big PITA change (solution: get the architecture right the first time!). Won't be surprised if there are still some holes somewhere.

samacumen:
Thanks @DonLakeFlyer for addressing the issue quickly. Do you have an idea when will the stable build be released for QGC with the mentioned changes (also in #4343)?

DonLakeFlyer:
Takes a couple hours to turn around.

On Dec 27, 2016, at 7:20 AM, Shyam Balasubramanian <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks @DonLakeFlyer<https://github.com/DonLakeFlyer> for addressing the issue quickly. Do you have an idea when will the stable build be released for QGC with the mentioned changes (also in #4343<https://github.com/mavlink/qgroundcontrol/issues/4343>)?

PR #4377, opened 2016-12-31
Add frame configuration page for ArduSub

jaxxzer:
This adds support for ArduSub frame configuration. The frame setup is implemented as a simplified version of the PX4 frame selection page. The page will only appear for vehicles identifying as ArduSub v3.5+

screenshot 2016-12-31 10 08 17
screenshot 2016-12-31 10 13 14
screenshot 2016-12-31 10 09 45


src/AutoPilotPlugins/APM/APMSubFrameComponent.cc
```diff
@@ -0,0 +1,78 @@
...
+{
+    return QUrl::fromUserInput(QStringLiteral("qrc:/qml/APMSubFrameComponentSummary.qml"));
+}
+
+QString APMSubFrameComponent::prerequisiteSetup(void) const
+{
+    APMAutoPilotPlugin* plugin = dynamic_cast<APMAutoPilotPlugin*>(_autopilot);
+    Q_ASSERT(plugin);
+
+    if (!plugin->airframeComponent()->setupComplete()) {
```
DonLakeFlyer: I didn't think ArduSub had an airframe setup?

jaxxzer: It does, but it has been hidden from view.
https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc#L64

It looks like Copter is the only vehicle that needs the Airframe component.
https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/APM/APMAirframeComponent.cc#L19

I have removed the airframe checks in prerequisiteSetup from the lights and sub-frame components.

Issue #3974, opened 2016-08-24
Artificial horizon is half black when in vertical climb or dive


kd0aij:
![art_horiz_90](https://cloud.githubusercontent.com/assets/2300221/17916971/3dd05a08-6975-11e6-91f3-267d985dbdaf.png)

DonLakeFlyer:
Mark, thinking about what this should show is making my head hurt a little. The black part should be blue right?

NaterGator:
It should be whatever color the other half is. If you're (unfortunately) pointed straight at the ground the two hemispheres in your view are both ground.

DonLakeFlyer:
Ah, yeah that's it thanks. Couldn't wrap my head around it as I watched it spin around :)

kd0aij:
yes, it's as though you're at the center of a globe with the top hemisphere
blue, and the bottom one green.

DonLakeFlyer:
@dogmaphobic Took a quick look at code. But I bet you could fix this in about a minute, whereas it would take me a while to figure out how the code works.

PR #4379, opened 2016-12-31
Fixed issue with artificial horizon

dogmaphobic:
In response to #3974

It was not working right when pointing straight up or down. Important for rocketry or ICBMs :P

In reality, something changed that caused the angle to end at 270/-270 instead of 180/-180. That in turn would make the display go past its end.


DonLakeFlyer:
I don't think mavlink spec says anything about the range of the values for attitude. Shouldn't the ui code somewhere normalize the values to a known range instead?

Issue #3800, opened 2016-07-13
Mission: JSON file format


DonLakeFlyer:
The goal of the new JSON file format for missions is to create a format which is both more human readable as well as a format which allows for easier placement of additional metadata. The additional meta data will allow storage for higher level concepts than just waypoints such as Survey or Structure Scan information.

Following is the structure of the JSON.

```
{
    "MAV_AUTOPILOT": 12,
    "complexItems": [ { complex mission item }, { complex mission item } ],
    "groundStation": "QGroundControl",
    "items": [ { mission item }, { mission item } ],
    "plannedHomePosition": { mission item }, 
    "version": "1.0"
}
```
- **MAV_AUTOPILOT** - This field holds the MAV_AUTOPILOT value for the firmware (MAV_AUTOPILOT_PX4=Px4 Pro, MAV_AUTOPILOT_ARDUPILOTMEGA=ArduPilot) this mission file was created for. The reason it is here is that there are differences in implementations of MAV_CMDs between the two firmwares. So we record it to be able to handle any inconsistencies in the right way.
- **complexItems** - This field is meant to hold complex mission items such as survey or structure scan. We'll discuss this more in a separate post.
- **groundStation** - This field specifies the ground station which wrote this file. The reason for this is that although the simple mission items are interchangeable between ground stations. The specification of complex mission items are specific to a ground station.
- **items** - The items field holds the full list of all mission item objects. It contains all mission items including any items which may be references by a higher level complex mission item. This way all ground stations can interchange at a minimum the full list of mission items. The ids which indicate sequence are 1-based.
- **plannedHomePosition** - This is required since the PX4 Pro and ArduPilot firmware handle mission item 0 as sent to a vehicle differently. For ArduPilot this item is sent as item 0. For PX4 Pro this item is not sent to the vehicle. Ground stations can use this value to show a planned home position. The planned home position should be a waypoint item.
- **version** - The version field is the version number for the file format.

A single mission item within the file looks like this:

```
{
    "autoContinue": true,
    "command": 16,
    "coordinate": [ 47.633425920000001, -122.08828776999999, 25 ],
    "frame": 3,
    "id": 2,
    "param1": 0,
    "param2": 0,
    "param3": 0,
    "param4": 0,
    "type": "missionItem"
},
```

The fields should be pretty self-explanatory. The match the values in the mavlink MISSION_ITEM message. 
- **coordinate** - latitude, longitude, altitude
[...]

PR #4383, opened 2017-01-02
New V2 mission file format

DonLakeFlyer:
As I was documenting the mission file format I realized that there was no way to generate the v1 format with a survey in it manually. You would need to know the QGC survey creation internals to do it. The new V2 format fixes this problem. QGC now supports both V1 and V2 formats, but only outputs V2 format.

### DIFF
--- a/src/JsonHelper.cc
+++ b/src/JsonHelper.cc
@@ -97,10 +97,12 @@ void JsonHelper::saveGeoCoordinate(const QGeoCoordinate&    coordinate,
 
 bool JsonHelper::validateKeyTypes(const QJsonObject& jsonObject, const QStringList& keys, const QList<QJsonValue::Type>& types, QString& errorString)
 {
-    for (int i=0; i<keys.count(); i++) {
-        if (jsonObject.contains(keys[i])) {
-            if (jsonObject.value(keys[i]).type() != types[i]) {
-                errorString  = QObject::tr("Incorrect type key:type:expected %1 %2 %3").arg(keys[i]).arg(jsonObject.value(keys[i]).type()).arg(types[i]);
+    for (int i=0; i<types.count(); i++) {
+        QString valueKey = keys[i];
+        if (jsonObject.contains(valueKey)) {
+            const QJsonValue& jsonValue = jsonObject[valueKey];
+            if (jsonValue.type() != types[i]) {
+                errorString  = QObject::tr("Incorrect value type - key:type:expected %1:%2:%3").arg(valueKey).arg(_jsonValueTypeToString(jsonValue.type())).arg(_jsonValueTypeToString(types[i]));
                 return false;
             }
         }
@@ -141,14 +143,13 @@ bool JsonHelper::isJsonFile(const QByteArray& bytes, QJsonDocument& jsonDoc)
 
 bool JsonHelper::validateQGCJsonFile(const QJsonObject& jsonObject,
                                      const QString&     expectedFileType,
-                                     int                supportedMajorVersion,
-                                     int                supportedMinorVersion,
-                                     int&               fileMajorVersion,
-                                     int&               fileMinorVersion,
+                                     int                minSupportedVersion,
+                                     int                maxSupportedVersion,
+                                     int&               version,
                                      QString&           errorString)
 {
     // Check for required keys
-    QStringList requiredKeys = { jsonVersionKey, jsonFileTypeKey };
+    QStringList requiredKeys = { jsonFileTypeKey, jsonGroundStationKey, jsonVersionKey };
     if (!validateRequiredKeys(jsonObject, requiredKeys, errorString)) {
         return false;
     }
@@ -166,25 +167,24 @@ bool JsonHelper::validateQGCJsonFile(const QJsonObject& jsonObject,
         return false;
     }
 
-    // Parse and validate version
+    // Check version - support both old style v1 string and new style integer
 
-    QString incorrectVersionFormatErrorString = QObject::tr("Incorrectly formatted version value");
-    QRegularExpression versionRegExp("(\\d+).(\\d+)");
-    QRegularExpressionMatch match = versionRegExp.match(jsonObject[jsonVersionKey].toString());
-    if (!match.hasMatch()) {
-        errorString = incorrectVersionFormatErrorString;
+    QJsonValue versionValue = jsonObject[jsonVersionKey];
+    if (versionValue.type() == QJsonValue::String && versionValue.toString() == QStringLiteral("1.0")) {
+        version = 1;
+    } else {
+        if (versionValue.type() != QJsonValue::Double) {
+            errorString = QObject::tr("Incorrect type for version value, must be integer");
+            return false;
+        }
+        version = versionValue.toInt();
+    }
+    if (version < minSupportedVersion) {
+        errorString = QObject::tr("File version %1 is no longer supported").arg(version);
         return false;
     }
-    QStringList versionParts = match.capturedTexts();
-    if (versionParts.count() != 3) {
-        errorString = incorrectVersionFormatErrorString;
-        return false;
-    }
-
-    fileMajorVersion = versionParts[0].toInt();
-    fileMinorVersion = versionParts[0].toInt();
-    if (fileMajorVersion > supportedMajorVersion || (fileMajorVersion == supportedMajorVersion && fileMinorVersion > supportedMinorVersion)) {
-        errorString = QObject::tr("File version (%1.%2) is larger than current supported version (%3.%4)").arg(fileMajorVersion).arg(fileMinorVersion).arg(supportedMajorVersion).arg(supportedMinorVersion);
+    if (version > maxSupportedVersion) {
+        errorString = QObject::tr("File version %1 is newer than current supported version %2").arg(version).arg(maxSupportedVersion);
         return false;
     }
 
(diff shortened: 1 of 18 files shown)


DonLakeFlyer:
Added fix for #4384

DonLakeFlyer:
File formatis documents in dev guide: https://donlakeflyer.gitbooks.io/qgroundcontrol-developers-guide/content/MissionFileFormat.html

AndKe:
By "v1" mission file (I assume you mean QGC WPL http://qgroundcontrol.org/mavlink/waypoint_protocol)
Yes, those have no definition of a "survey" the proper way to save them would be to expand the data to a pile of waypoints,  just as you do while uploading to AP.  loading such QGC WPL mission would be... a pile of waypoints.  or - outmost of those waypoints could automatically be used to create a polygon (with a survey)  - but while surely impressive, it's not an essencial feature.

To me, it's more important to be able to load an 2013 mission, (pile of WP's)  - and if I need to edit it, I can draw a survey using it for reference.

Also; is Tridge /mavlink community aware of this new format?  - the last thing we need is for somebody to invent yet another format just because this one is not known. A link to your dev guide should maybe be in the http://qgroundcontrol.org/mavlink/waypoint_protocol page.

DonLakeFlyer:
By V1 I mean the V1 JSON format. The old original text file format is still supported. Discussion with other GCS folks is happening here: #3800

AndKe:
great :+1:

zorandragan:
Hi.
I am new at UAV busineess, recently started using QGC.

Is this "pile of waypoints" within a survey plan export planned to be added as a feature?
It would be very useful to have that list of WPs, so you could edit them, remove unnecessary, etc.

AndKe:
@zorandragan it should be, for proper interchangeability with other GCS's and future-proof.

Until it is implemented, as a workaround:  you can connect to a autipilot (or SITL) and upload a mission, then download it, it will then be at lest presented as "pile of waypoints" , if you download it using APM Planner2 , or mavproxy or other tools, you'll be able to save it as .txt (QGC WPL format)   -

DonLakeFlyer:
> It would be very useful to have that list of WPs, so you could edit them, remove unnecessary, etc.

Even if Survey exported the waypoints in the file format it would not support manipulating them and then reading them back into QGC from the Survey file format. If you want to manipulate them individually you need to "degrade" the Survey back to a pile of waypoints by round tripping through the vehicles and @AndKe describes.

zorandragan:
@AndKe  @DonLakeFlyer  I wasn't very clear, sorry, I will try to explain my idea better. One thing would be to edit those exported waypoints manually. 
The other thing, probably more demanding in terms of software development, is to "breakdown" a survey into manual waypoints and then allow editing in GUI. In the following picture I have tried to depict my idea. 

qgc

1. Beside the edge points of the survey area, I have started drawing white circles within the area representing individual waypoints (these white circles should cover the whole path, but you get the idea ). Then I have selected some points in the middle of the area, rounded by the black rectangle, which I would like to exclude from surveying (there is a structure I would like to avoid).

2. As the survey area gets broken down to individual waypoints, I would like to have a possibility to manipulate each of them if I want. Like with manual waypoints, to change gimbal position, the mode of camera, etc. 

3. When this plan gets exported, a "mixed" format of polygon and waypoint .plan is generated. A ComplexItem like "polygon" contains a list of "SimpleItems". So basically, a plan format which contains  computed flightlines/waypoints and sensor information and polygon coordinates.

The main idea is to let QGC generate waypoints (hundreds of them) instead of me doing it. Most of these points I wouldn't touch but some I might want to manipulate. Then, when exporting the plan, save the sensor information and all the computed waypoints, in exchangeable format. JSON is fine, just to extend the structure to have it all inside.

DonLakeFlyer:
> The main idea is to let QGC generate waypoints (hundreds of them) instead of me doing it. Most of these points I wouldn't touch but some I might want to manipulate.

Start with a Survey. Round trip it through a vehicle to get it broken down into points. Go crazy editing it.

Supporting what you want from within an actual Survey complex item and maintaining it as a Survey complex item after editing is unlikely to happen any time soon.

zorandragan:
> Supporting what you want from within an actual Survey complex item and maintaining it as a Survey complex item after editing is unlikely to happen any time soon.
I understand that would be too complex to handle.

@DonLakeFlyer Thank you very much for the time you've spent reading and answering my long posts, I really appreciate it.

Would it be possible to have an option of saving a more detailed survey .plan file, where camera trigger actions would be visible as well? Nothing in UI, no UI editing, or any kind of manipulating them, just saving them.

By more detailed I mean following:
QGC uses MAV_CMD_DO_SET_CAM_TRIGG_DIST  to define scan lines. What I find usefull
is to have these scan lines, within the survey between two waypoints that define a scan line, broken down to items containing MAV_CMD_DO_DIGICAM_CONTROL commands (or still using MAV_CMD_DO_SET_CAM_TRIGG_DIST  and setting empty parameters to contains coordinates and trigger action). 
To recalculate the scan lines, and instead of scan lines, to have individual camera events, containing coordinates where the camera has to be triggered. All that saved in .plan file. I hope I am being clear, something like

             {
                "autoContinue": true,
                "command": 206, (203)
                "doJumpId": 8,
                "frame": 3,
                "params": [
                    0,
                    0,
                    0,
                    1, (camera action)
                    -35.3635552,
                    149.1648896,
                    50
                ],
                "type": "SimpleItem"
            },

zorandragan:
Hi.

Since there was no response to my suggestion, I think I could start working on this issue on my own, implementing what I need. 

Or this still could be still put on a feature request list?

hamishwillee:
@DonLakeFlyer is on holiday, so you may not get a response quickly.

zorandragan:
@hamishwillee Thanks, let's wait then.

hamishwillee:
That's not a reason not to ping regularlys - here and/or slack. :-)

zorandragan:
Hello everyone, as @hamishwillee suggested, to ping regulary :) I wanted to know if someone still follows this topic?

DonLakeFlyer:
If you start with a Survey complex item. Turn off "take images in turnarounds". Then round trip it through the vehicle you will end up with a mission which has camera command at the start/end of each transect. Then you edit it with more detail as you like.

zorandragan:
When you say “transact” do you mean survey line edge?
I was thinking more about having each of the camera actions being individually available visible. I have already done a round trip through a vehicle, and I got commands at the start/end of each line. But those commands are still triggering camera at a certain distance. Those actions at each of the triggered distances I would like to have as individual waypoints.

DonLakeFlyer:
"transect" = survey line edge

> Those actions at each of the triggered distances I would like to have as individual waypoints.

That is not supported and generally ends up not being that useful except for small special purpose surveys. The reason being it tends to generate missions which have more mission items than the firmware can handle.

zorandragan:
I know it is not supported at the moment by QGC, but how big effort woyld it be to support it? I need exactly those small special survey missions, at least for QGC to be able to do mission planning, I would take care of flight execution and firmware performance.

AndKe:
@zorandragan why would you wish an WP for every camera item ? What would be the exact use scenario ?

zorandragan:
@AndKe One use case would be to have each of these individual waypoints at least saved in a .plan or some other file format, as i have described it previously. In this case QGC would be used as a flught planning software. Then those WPs I could manipulate if I want in firmware or flight planning software. Ideally, QGC could have been displaying each of these WPs and allow its moving/deletion/editing in order to avoid a structure during a flight, a custom camera zoom at certain WP for a closer inspection.

DonLakeFlyer:
> but how big effort woyld it be to support it?

We are talking about two things here:
1) Exclusion zones in a survey. This is a good feature but very difficult to implment.
2) Generating waypoints and camera command at each image location. This is not a huge amount of work. But it's also a feature that I wouldn't add to QGC since it's a special use case that would just make things more confusing than they already are for 99% of the users.

zorandragan:
Probably these things are two features, but related. The way I see it, if once there would be waypoints at each image location, user could manually delete them if there is an object to avoid during a survey mission.
1. I agree it is a very good and useful feature. Which aspect of its implementation would be complex, UI or logic behind?
2. That use case would be very very useful. There could be an “expert” or “advanced” mode for survey mission editing, accessible with a button or some option in the menu. That way a “standard” user mode would still remain, and “advanced” mode would be available if nedded. What do you think about it? Plus saving that survey mission in a file with individual waypoints at each camera action.

AndKe:
"user could manually delete them if there is an object to avoid during a survey mission."
if the object to avoid is in the middle, removing individual camera events won't change the flightpath. 

If the object is on the outer perimeter, it's easy, just reduce the polygon to exclude it.

zorandragan:
> if the object to avoid is in the middle, removing individual camera events won't change the flightpath

You are right, that won't change the path. I should have described it better. I was reffering to waypoints in whole, removing the entire waypoint where the drone should flight to, not only the camera action (command). If a whole waypoint is removed (when I say waypoint I reffer to a position where the drone should flight to), then a "corrected" or "recalculated" flight path should be displayed, with exclusion zone. That's why I have asked which aspect of this feature to implement would be difficult, UI or logic behind.

AndKe:
I am afraid exclusion zones and things like that are way ahead.
I do fly a lot in valleys, and manual planning is the solution for all complex missions, or else we could as well ask for elevation changes curing the survey.
Personally, I hope the devs spend some time on better flight monitoring features.

zorandragan:
Since QGC has a very nice survey planning feature, an extension or improvement to it could be zone exclusions. There could be a step by step approach in getting to this feature, i will try to elaborate on that.

I would go back now to a second feature we have discussed.

@DonLakeFlyer  stated:
> 2. Generating waypoints and camera command at each image location. This is not a huge amount of work. But it's also a feature that I wouldn't add to QGC since it's a special use case that would just make things more confusing than they already are for 99% of the users.

My answer to that was:
> That use case would be very very useful. There could be an “expert” or “advanced” mode for survey mission editing, accessible with a button or some option in the menu. That way a “standard” user mode would still remain, and “advanced” mode would be available if nedded. What do you think about it? Plus saving that survey mission in a file with individual waypoints at each camera action.

IMHO getting to a complete feature can be broken down in implementation steps:
1. Generating waypoints and camera command at each image location
2. Add "Save To File Expert" (or similar) button in the Sync menu to save all waypoints in a file.
3. Adding a new radio button next to Mission/Fence/Rally to select, which would then show all available waypoints at each image location (just show at this step, editing already leads to that other feature of exclusion zones). By already just showing locations where images will be taken, would mean much to a surveyor, and later on would be useful for flight path optimization. Also, it gives the professional look to a flight planning software.

AndKe:
number 2 is covered by my good old request to offer saving in QGC WPL format, the good old .txt file format that all GCS's read:  http://qgroundcontrol.org/mavlink/waypoint_protocol#waypoint_file_format

QGC can already read them, just needs to save them.

zorandragan:
I agree. Plus I would extend that request with saving all waypoints to a file including sensor data. Combination of the existing .plan file with QGC WPL format. It would be great to have these two things contained at one place, camera settings and waypoints.

zorandragan:
@DonLakeFlyer Would you consider adding requests we have discussed to the feature list? 

1. Generating waypoints and camera command at each image location
2. Saving all waypoints to a file including sensor data

DonLakeFlyer:
1) This was discussed with fixed wing survey users prior to the Survey implementation. The conclusion what that fixed wing surveys are not done that way. So this ends up as a special case which would get little use. Survey setup is already complex enough I don't want to add another option which will get little use.
2) Already covered by round tripping through the vehicle and saving. At some point I may implement @AndKe downlevel file format save suggestion. But not in the near future.

PR #4399, opened 2017-01-06
Qgroundcontroll

sunyco:


DonLakeFlyer:
Assuming this is some sort of mistake.

Issue #4414, opened 2017-01-10
Survey functionality, create auto take off and landing/RTL options


leondwyer:
One of the great thing MS does is it handles the auto take of land / rtl waypoint for you to in the survey planner.

It saves having to set them up yourself. there just simple check boxes on the options screen.

DonLakeFlyer:
Also need to consider flight speed since I didn't add that in my referenced pull. The reason is that with QGC a Survey item is slightly different than in Mission Planner. In Mission Planner the Survey widget sets up an entire mission. Whereas in QGC a Survey is only considered a part of a survey and you can have multiple surveys in a mission for example. Need to rationalize that difference with this request.

DonLakeFlyer:
RTL end action and flight speed changes are now handled in first mission settings item. So this is all automatic now.

PR #4420, opened 2017-01-10
Mission Settings replaces Planned Home Position

DonLakeFlyer:
- Shot Interval in Survey
- Much change to time/distance calc
- Time calculation now takes into account DO_CHANGE_SPEED in mission

Mission Settings:
screen shot 2017-01-10 at 3 35 49 pm

Survey - Shot Interval
screen shot 2017-01-10 at 3 36 02 pm


DonLakeFlyer:
This moves the offline Plan settings from Settings/General to this new Mission Settings editor. This way it's closer to where you really need it. Also you can change them directly from the Plan view. If you are offline editing, after you have added items to a mission you can no longer change firmware or vehicle type. You can change the speeds any time you want.

DonLakeFlyer:
I did not place the ability to change the flight speed directly into the Survey item. I'll revisit that at a future time with the other options requested.

DonLakeFlyer:
Related to #4414

PR #4424, opened 2017-01-11
UDP join multicast group for windows client

sanderux:
Explicitly join the default multicast address seems to be required on windows when multicasting telemetry over vpn.

### DIFF
--- a/src/comm/UDPLink.cc
+++ b/src/comm/UDPLink.cc
@@ -245,6 +245,7 @@ bool UDPLink::_hardwareConnect()
     _socket->setProxy(QNetworkProxy::NoProxy);
     _connectState = _socket->bind(host, _udpConfig->localPort(), QAbstractSocket::ReuseAddressHint | QUdpSocket::ShareAddress);
     if (_connectState) {
+        _socket->joinMulticastGroup(QHostAddress("224.0.0.1"));
         //-- Make sure we have a large enough IO buffers
 #ifdef __mobile__
         _socket->setSocketOption(QAbstractSocket::SendBufferSizeSocketOption,     64 * 1024);


DonLakeFlyer:
@dogmaphobic Can you look? I know nothing about networking.

sanderux:
First results are that this works on windows 7 but not on windows 10
That might however be related to the specific computer settings, we'll test some more

sanderux:
Tested on a different machine, it was indeed a computer specific problem and this patch allows QGC to receive multicast telemetry. This is useful when drones use cellular connection, it means they only have to stream the telemetry once multiple groundstations can connect over a vpn.

Perhaps this should be tested on other platforms too but i see no reason for it to fail on those.

dogmaphobic:
> Can you look? I know nothing about networking.

In theory, it just enabled multicast. I would put it as an option to be on the safe side though. This is a fringe use case any way. What I'm curious is what is the use case? One QGC to many Vehicles on multicast doesn't make sense. One vehicle to multiple QGCs? How would that work?

sanderux:
@DonLakeFlyer The multicast address 224.0.0.1 is the IPv4 reserved address that tell's routers to forward the packet to all connected clients and all clients on the same subnet listen to it. 
http://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml

sanderux:
@dogmaphobic we have telemetry running over vpn over a 3g connection. The reason that we have it multicast is 3-fold;

1. We do indeed allow multiple groundstations and have tested this extensively
2. Having multiple groundstations the multicast method allows a single telemetry stream to pass through the 3g connection saving bandwidth
3. You could reason broadcast would work just as well and it does, but android and linux do not receive broadcast messages until they have sent a packet on that network. They do receive multicast directly.

We have tested 
- single drone, single GCS
- single drone multiple GCS
- multiple drone, single GCS
- multiple drone, multiple GCS

It all works very well as long as you set unique mavlink id's

sanderux:
What would be the benefit of setting it as an option? I would opt for as much plug & play as possible. If users put their drones on the same network they would receive both broadcasts anyway

dogmaphobic:
I get confused enough with one-to-one :P

As long as the user knows what they are doing. I can see all kinds of can-of-worms scenarios with multiple instances of QGC talking to one vehicle. As to make it optional, it just a matter of not knowing how this will affect everybody else. Leave as is for now and let's pay close attention how it behaves.

sanderux:
EU drone regulations allow extended line of sight flights if there is a second operator on a different location. Multiple telemetry connections with multiple emergency stop or RTL buttons active make the drone comply with such regulations.

sanderux:
PS: multicasting is not the default behavior of px4 or mavproxy type relays, the 'option' is already set by enabling this on that side. I don;t see how this differs from the default behavior that responds to broadcasts (also not optional)

dogmaphobic:
Broadcast is passive while multicast requires an active configuration. So far, broadcast is only used for discovery. The vehicle broadcasts its heartbeat and QGC responds. Once the vehicle receives a response, it switches to unicast with that one QGC only.

Yes, multicast would only be possible if you use something that enables it, such as a companion computer or dedicated (WiFi) radio. As said, let's get this in and see how it goes in the long run.

PR #4438, opened 2017-01-13
Video recording

jaxxzer:
This pull request will provide support for recording the gstreamer video stream (not UVC devices). When a stream is present, a small icon is displayed in the lower right hand corner of the FlightView. The icon can be clicked to start recording to a new file, or to stop recording the current file. The icon's shape indicates the current recording status, a square for currently recording and a circle for not recording. The icon is present and functional whether the video is full screen, or in the pip window.

screenshot 2017-01-13 16 01 13
screenshot 2017-01-13 16 01 28
screenshot 2017-01-13 16 01 40

The location that the video is saved can be selected through a file browser in the general settings page. The file name of the videos is QGC-yyyy-mm-dd_hh.mm.ss.mkv. The video format is matroska, chosen because you will still get a playable video file if the application crashes and the video muxer is unable to finalize the file. The video output to both file and the VideoSurface is stable and has been tested extensively (with udp, I haven't tested rtsp). I have saved properly formatted videos that were 10+GB in size and 1.5hours+ in length. The video is also formatted properly if the stream is interrupted or stopped, or if the application is closed normally while recording.

screenshot 2017-01-13 16 01 59

I had to add a work-around for gst_bus_add_watch not working on Windows. Instead of adding a watch to the bus, a timer with an interval of 0 is used to manually pop messages off the bus, and call the handler.


NaterGator:
@jaxxzer how are timestamps working? When I did this in the past I had to rebase timestamps downstream of the tee element on the filesink side based on recording start time. Without 'fixed' timestamps certain media players would wait for whatever the preroll time was before recording started to actually start presenting frames.

jaxxzer:
The pipeline when the application is launched looks like this:

datasource-->rtph264depay-->h264parse-->tee-->queue-->avdec_h264-->_videoSink

We start recording by creating the three elements on the bottom and connecting them to a src pad on the tee:

datasource-->rtph264depay-->h264parse-->tee-->queue-->avdec_h264-->_videoSink
                                                                          |
                                                                          +-->queue-->matroskamux-->filesink

When the icon is clicked to stop recording, these steps take place in order to finalize the video file and to prevent interrupting the video stream displayed on the VideoSurface:

- Create a second temporary pipeline
- Remove the three elements from the first pipeline, and put them in the temporary pipeline
- Inject an EOS event at the beginning of the temporary pipeline
- Wait for the EOS message to be posted to the bus of the temporary pipeline.
- Shutdown the temporary pipeline, and free the temporary pipeline and its elements.

When the application is closed, or the stream is stopped, an EOS event is sent to the main pipeline. Then we wait (blocking) for EOS to be posted to the bus before freeing the pipeline. If we are recording, this process will also finalize the video file.

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -196,30 +252,42 @@ void VideoReceiver::start()
...
-            gst_object_unref(bus);
-            bus = NULL;
-        }
+//        if ((bus = gst_pipeline_get_bus(GST_PIPELINE(_pipeline))) != NULL) {
+//            gst_bus_add_watch(bus, _onBusMessage, this);
+//            gst_object_unref(bus);
+//            bus = NULL;
+//        }
+        // Workaround for above watch on Windows
+        _busCheckTimer.start(0);
```
NaterGator: Yikes, don't do this! It pegs a core. I have some example code that works for windows...

jaxxzer: Thank you. I am eager to see it.

jaxxzer:
@NaterGator I'm not sure how the timestamps are being sorted out, I haven't had any issues like you mentioned. That's interesting though, please let me know if you can try it and see if you have the same issue here. The filesink is created and destroyed every time video recording starts/stops, and it starts recording immediately after connecting to the pipeline, so it shouldn't ever see a timestamp prior to when it starts recording.

NaterGator:
Right, but it joins the pipeline and has its time managed by the pipeline. I guess it depends; in my case I had some more complicated stuff going on (injecting private MPEG-TS data streams, MISB 0601 stream, etc).  If you haven't had a problem maybe gstreamer is managing it better nowadays.

src/VideoStreaming/VideoReceiver.cc
NaterGator: You were actually on the right track with the bit you commented out. You just need to use g_signal_connect to "sync-message" on the bus and gst_bus_enable_sync_message_emission. From there you need to respect that the callback handler will be called from an arbitrary thread and **_only_** use signals (you can connect them to lambdas, slots, whatever you want) to interact with your QObject in the original thread. 

The Qt signal/slot mechanism will automatically ensure that slot is invoked on the correct thread regardless of the thread that libgstreamer calls you back from.

jaxxzer: Great, I'll try it thanks. What about https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstBus.html#gst-bus-set-sync-handler ?

NaterGator: See here: https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstBus.html#gst-bus-enable-sync-message-emission 

You don't actually want to _become_ the bus sync handler, you just want to be signaled about the messages on the bus so you can act on them accordingly (such as EOS) without being responsible for handling them.

NaterGator:
Windows has precompiled Qt headers which hides the fact that QDir is not included now in VideoManager.cc. Add #include <QDir> to VideoManager.cc to continue the build.

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -417,7 +396,7 @@ gboolean VideoReceiver::_onBusMessage(GstBus* bus, GstMessage* msg, gpointer dat
     Q_UNUSED(bus)
     Q_ASSERT(msg != NULL && data != NULL);
     VideoReceiver* pThis = (VideoReceiver*)data;
-    pThis->_onBusMessage(msg);
+    pThis->busMessage(msg);
```
NaterGator: Ideally you'd actually do all the GstMessage processing here (what is currently done in _handleBusMessage) and use signals to perform specific actions on the VideoReceiver object (EOS handling, state change) by passing and necessary data as POD through the signal instead of the GstMessage pointer. 

Since you don't hold your own refcount to the message (you're called synchronously on the owner's thread) you can't ensure that it will live on after you return to the caller in the callback thread. When you emit a signal with the message pointer to a different thread, the pointer itself is queued and it becomes highly likely the underlying GstMessage pointed to is no longer valid by the time the slot gets invoked on the QObject's thread. 

You basically need two signals: msgEOSReceived() and msgStateChanged() that connect to slots that invoke the code in those sections of the switch statement. The GST_MESSAGE_ERROR block doesn't reference any state on your VideoReceiver object so that can go ahead and run in the callback thread context.

jaxxzer:
I should mention, this has been tested on Windows, Mac, and Ubuntu 14.04.

src/VideoStreaming/VideoReceiver.cc
jaxxzer: Thanks for the clarification! I've done just that, but also added a signal for Error because some action does need to be taken on the VideoReceiver object, after all.

src/VideoStreaming/VideoReceiver.h
```diff
@@ -25,41 +26,88 @@
...
 public:
+    Q_PROPERTY(bool recording READ recording NOTIFY recordingChanged)
+
     explicit VideoReceiver(QObject* parent = 0);
     ~VideoReceiver();
 
 #if defined(QGC_GST_STREAMING)
-    void setVideoSink(GstElement* sink);
+    void setVideoSink(GstElement* _sink);
+#endif
```
NaterGator: Move this endif below bool stopping() and I think you're good to go.

src/FlightDisplay/FlightDisplayView.qml
```diff
@@ -260,6 +260,42 @@ QGCView {
...
+                width:                      height * 0.625
+                sourceSize.width:           width
+                source:                     "/qmlimages/CameraIcon.svg"
+                fillMode:                   Image.PreserveAspectFit
+                color:                      "white"
+            }
+
+            MouseArea {
+                anchors.fill:   parent
+                onClicked:      QGroundControl.videoManager.videoReceiver.recording ? QGroundControl.videoManager.videoReceiver.stopRecording() : QGroundControl.videoManager.videoReceiver.startRecording()
```
jaxxzer: When built without gstreamer support, videoManager.videoReceiver is NULL, and we get the error
TypeError: Cannot read property 'recording' of null

What is the best way to check for this?

src/ui/preferences/GeneralSettings.qml
```diff
@@ -41,6 +41,14 @@ QGCView {
 
     QGCPalette { id: qgcPal }
 
+    FileDialog {
```
DonLakeFlyer: The QML file dialog is pretty horrendous. We only use it as a last resort on mobile builds. Give me a couple days and I'll make a QGCFileDialog QML control that uses native dialog on desktop and Qml dialog on mobile. I need it for some work I"m doing anyways.

DonLakeFlyer: @jaxxzer Actually how about this. Leave this as is. Then add an Issue to convert to new QGCFileDialog control that I will create. Then once I'm done with that, I'll convert this stuff over. That way you don't need to wait for me.

DonLakeFlyer:
Also before this gets merged in I'd like @dogmaphobic to take a look. I don't think he's back in town yet, so may still be a little while.

src/ui/preferences/GeneralSettings.qml
DonLakeFlyer: Sorry for all the noise. I just added the Issue myself: #4456

dogmaphobic:
I just got back. I have a bit of a logjam but I will try this as soon as I get a chance.

dogmaphobic:
We also need to think about this a bit more. The advantages of recording the video within QGC is obvious but there are times where this would become redundant. In my original test platform, the recording is handled at the capture side. I have an Odroid handling the capture off a Logitech camera. Video is recorded by the Odroid to its eMMC storage. I never got around to it but the idea was to implement the proper Start/Stop video recording MAVLink messages. 

For now, this PR will work fine as it is the only way to *officially* record the video stream. However, we will be implementing an interface for camera control, which in turn will tell the “camera” to start/stop recording video itself. When that’s implemented, we will need a way to differentiate who is doing the recording.

DonLakeFlyer:
Gus, maybe we need to be careful about the start recording icon on the video screen? Maybe something to try to make it more clear video is being recorded locally, not on the vehicle/camera?

DonLakeFlyer:
@jaxxzer Can you clean up the conflict and I'll merge

PR #4458, opened 2017-01-21
New PID tuning screen

anton-matosov:
- Introduced fact stepper control that is mobile friendly
- Added new screen for PID tuning for the PX4 copters

Other changes:
 - Implement button repeat if its long pressed (must have for mobile stepper)


LorenzMeier:
Can you please share screenshots for UI pull requests? Thanks!

anton-matosov:
image
image
image

anton-matosov:
Roadmap:
 - v2 will have quick filters for simple/advanced settings set, pitch/roll lock
 - v3 simple graph for TPA settings (it's much easier to understand this way)
 - v4 attitude data plotting

dagar:
Nice. I'll fill in FW once this is merged.

src/AutoPilotPlugins/PX4/PX4PIDTuningComponent.cc
```diff
@@ -0,0 +1,113 @@
...
+QStringList PX4PIDTuningComponent::setupCompleteChangedTriggerList(void) const
+{
+    return QStringList();
+}
+
+QUrl PX4PIDTuningComponent::setupSource(void) const
+{
+    QString qmlFile;
+
+    switch (_vehicle->vehicleType()) {
```
DonLakeFlyer: I'd change this to use Vehicle::fixedWing() and Vehicle::multiRotor() instead of a big switch statement. That way all of this sort of switch statement style checking is centralized. Also if new vehicle types show up you change the checks in one place - Vehicle and all other places work correctly.

src/AutoPilotPlugins/PX4/PX4PIDTuningComponent.cc
```diff
@@ -0,0 +1,113 @@
...
+            qmlFile = "qrc:/qml/PX4PIDTuningComponentCopter.qml";
+            break;
+        case MAV_TYPE_VTOL_DUOROTOR:
+        case MAV_TYPE_VTOL_QUADROTOR:
+        case MAV_TYPE_VTOL_TILTROTOR:
+        case MAV_TYPE_VTOL_RESERVED2:
+        case MAV_TYPE_VTOL_RESERVED3:
+        case MAV_TYPE_VTOL_RESERVED4:
+        case MAV_TYPE_VTOL_RESERVED5:
+            qmlFile = "qrc:/qml/PX4PIDTuningComponentOther.qml";
```
DonLakeFlyer: Can you instead change this such that the PID Tuning component is not added to the list if !_vehicle->fixedWing() && !_vehicle->multiRotor().

src/AutoPilotPlugins/PX4/PX4PIDTuningComponentCopter.qml
```diff
@@ -0,0 +1,177 @@
...
+    panelTitle:   "PID Tuning"
+    pageDescription: "PID Tuning is used to setup Proportional Integral Derivative controllers"
+
+    steppersModel: ListModel {
+
+///////////////////////////////////////////////////////////////
+        ListElement {
+            title:          qsTr("Roll P")
+            param:          "MC_ROLL_P"
+            group:          "Roll"
```
DonLakeFlyer: Could you put these group names into a readonly string property so that they aren't repeated over and over again?

src/AutoPilotPlugins/PX4/PX4PIDTuningComponentCopter.qml
```diff
@@ -0,0 +1,177 @@
...
+ ======================================================================*/
+
+import QtQuick              2.5
+import QtQuick.Controls     1.4
+
+import QGroundControl.Controls  1.0
+
+FactStepperGroup {
+    anchors.fill: parent
+    panelTitle:   "PID Tuning"
```
DonLakeFlyer: Any user visible string needs a qsTr("...") around it to mark it for translation.

src/AutoPilotPlugins/PX4/PX4PIDTuningComponentOther.qml
```diff
@@ -0,0 +1,31 @@
...
+
+ QGROUNDCONTROL is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with QGROUNDCONTROL. If not, see <http://www.gnu.org/licenses/>.
+
+ ======================================================================*/
```
DonLakeFlyer: This is the old file header. New one is:
```
/****************************************************************************
 *
 *   (c) 2009-2016 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
 *
 * QGroundControl is licensed according to the terms in the file
 * COPYING.md in the root of the source code directory.
 *
```

src/FactSystem/FactControls/FactStepper.qml
```diff
@@ -0,0 +1,115 @@
+import QtQuick 2.5
```
DonLakeFlyer: Missing file header

src/FactSystem/FactControls/FactStepper.qml
```diff
@@ -0,0 +1,115 @@
...
+import QGroundControl.ScreenTools   1.0
+import QGroundControl.FactSystem 1.0
+import QGroundControl.Palette 1.0
+import QGroundControl.Controls 1.0
+
+Item {
+    property bool enabled: true
+    property alias fact: factInput.fact
+
+    property string valueText:          fact ? fact.valueString : "N/A"
```
DonLakeFlyer: Qml coding style has internal properties use same C++ style ```_``` prefix to identify public versus private variables. So if these are internal use only they should be ```_valueText``` for example/

src/FactSystem/FactControls/FactStepper.qml
```diff
@@ -0,0 +1,115 @@
...
+            height: parent.height
+            text: "-"
+            enabled: fact ? (factValue > fact.min) : false
+
+            onClicked: {
+                incrementWithScale(-1)
+            }
+        }
+
+        QGCTextField {
```
DonLakeFlyer: Why not FactTextField? You can still override as needed but you still get all the base functionality.

AndKe:
nice , please add actual parameter name somewhere,  it's better, but still a bit scary.

The ultimate nightmare, I would not dare to touch in field is this:
(It does not tell me anything about what is changed, and how much )
screenshot from 2017-01-22 19-58-51

src/AutoPilotPlugins/PX4/PX4PIDTuningComponent.cc
anton-matosov: Sounds good. I was following PX4TuningComponent example.

src/AutoPilotPlugins/PX4/PX4PIDTuningComponent.cc
anton-matosov: Currently I have nothing for fixed wing, so I have changed to show current PID only for multirotor and VTOL

src/AutoPilotPlugins/PX4/PX4PIDTuningComponentCopter.qml
anton-matosov: Done

src/AutoPilotPlugins/PX4/PX4PIDTuningComponentCopter.qml
anton-matosov: Done

src/AutoPilotPlugins/PX4/PX4PIDTuningComponentOther.qml
anton-matosov: done

src/FactSystem/FactControls/FactStepper.qml
anton-matosov: Done

src/FactSystem/FactControls/FactStepper.qml
anton-matosov: Ok. I will change that. That's my first QML experience

src/FactSystem/FactControls/FactStepper.qml
anton-matosov: Because of different editing and validation behavior. FactTextField has very "desktop" behavior of showing up popup dialogs. This is unacceptable for mobile UI.
Editing is different in FactStepper as it has +/- buttons that support long press for continuous  increment/decrement. This kind of editing needs delaying of value setting. 
Considering that FactTextField doesn't support customization of any of these features, it is much easier to use QGCTextField directly

anton-matosov: Another option here would be to use FactTextField and override `onEditingFinished` `onHelpClicked` to have at least some code reuse.

anton-matosov:
@DonLakeFlyer unfortunately I can't use SetupPage as a base for the PID tuning page. The issue is in QGCFlickable inside SetupPage. Flickable can't be used as a parent for the ListView, because ListView itself is Flickable and needs a fixed parent size with predefined height, otherwise it collapses to 0 height

src/FactSystem/FactControls/FactStepper.qml
DonLakeFlyer: Since you not only support using +- buttons but you also support typing in a value you need to validate that values that are entered in there. This is why FactTextField shows right side dialogs if you enter in values which fail the validation specified by the parameter meta data. This isn't a mobile or desktop difference, this is just basic field validation which is a required thing to keep parameters values clean.

DonLakeFlyer:
Do you really need to use a ListView? Are you running into a performance problem? You should just be able to create a component which is a ColumnLayout which is then created inside the SetupPage flickable.

src/AutoPilotPlugins/PX4/PX4PIDTuningComponentCopter.qml
DonLakeFlyer: Crap, I've run into that before, forgot about it.

src/AutoPilotPlugins/PX4/PX4PIDTuningComponent.cc
DonLakeFlyer: Those Vehicle::fixedWing ... things are new and the PX4TuningComponent code is older than that. Could you fix up the PX4TuningComponent code as well to use the new way to do it while you are in this space?

DonLakeFlyer:
> It does not tell me anything about what is changed, and how much

@AndKe For the parameters that control those the actual values don't really mean much to a user. Other than make them slightly bigger or smaller to gain some affect. As an example RC Roll/Pitch Feel is a value that goes from 0-100 with 0 being soft and 100 being crisp. The actual values don't mean much unless you try them out by flying them. Not really sure how to make something like that any more clear.

src/AutoPilotPlugins/PX4/PX4PIDTuningComponent.cc
anton-matosov: sure

anton-matosov:
Yes, I really need ListView, as in current version I have sections which are not supported with raw repeater.
In the next version I am planning to use DelegateModel for quick filtering
Good news. I have found a workaround to the problem of listview in flickable. code is checked in.

I will add validation to stepper control (most likely will refactor FactEdit to be reusable for me)

anton-matosov:
image
image

anton-matosov:
Ok, I have tried to use FactTextField and now I remember why I have moved out of it.
The validation error causes full fledged editor to appear on screen:
image

DonLakeFlyer:
> Ok, I have tried to use FactTextField and now I remember why I have moved out of it.
> The validation error causes full fledged editor to appear on screen:

If the user types in a bad value into the edit box that is exactly what you want to happen. Not sure what the problem is? That is how all of QGC works.

src/FactSystem/FactControls/FactStepper.qml
```diff
@@ -0,0 +1,130 @@
...
+    id: stepper
+    property bool enabled: true
+    property alias fact: factInput.fact
+    property alias showHelp: factInput.showHelp
+    property alias showUnits: factInput.showUnits
+
+    property string _valueText:          fact ? fact.valueString : "N/A"
+    property double _factValue:          Number(_valueText)
+
+    property double minimumValue:       !fact || isNaN(fact.min) ? 0 : fact.min
```
DonLakeFlyer: Why the isNaN on these? min/max meta data should always be set to valid values.

src/FactSystem/FactControls/FactStepper.qml
```diff
@@ -0,0 +1,130 @@
...
+        FactTextField {
+            id: factInput
+            width : parent.width / 2
+            height: parent.height
+
+            horizontalAlignment: TextInput.AlignHCenter
+
+            text:       _valueText
+
+            fact: _fact
```
DonLakeFlyer: Don't see where this ```_fact``` value is coming from. Doesn't seem to exist.

src/FactSystem/FactControls/FactStepper.qml
```diff
@@ -0,0 +1,130 @@
...
+    property double stepRatio:          0.1
+
+    implicitHeight: 50
+    implicitWidth: implicitHeight * 4
+
+    signal validationError(string text)
+    signal helpClicked
+
+    Component.onCompleted: {
+        factInput.validationError.connect(validationError)
```
DonLakeFlyer: I don't get what this is doing?

src/FactSystem/FactControls/FactStepper.qml
```diff
@@ -0,0 +1,130 @@
...
+    property bool enabled: true
+    property alias fact: factInput.fact
+    property alias showHelp: factInput.showHelp
+    property alias showUnits: factInput.showUnits
+
+    property string _valueText:          fact ? fact.valueString : "N/A"
+    property double _factValue:          Number(_valueText)
+
+    property double minimumValue:       !fact || isNaN(fact.min) ? 0 : fact.min
+    property double maximumValue:       !fact || isNaN(fact.max) ? 1 : fact.max
```
DonLakeFlyer: Why all the checking for !fact? If there is no fact set then you don't really want to hide that fact since it is badly written code.

src/FactSystem/FactControls/FactStepper.qml
```diff
@@ -0,0 +1,130 @@
...
+            }
+
+            onValidationError: {
+                // do nothing by default
+            }
+
+            onHelpClicked: {
+                // do nothing by default
+            }
+        }
```
DonLakeFlyer: I think this should just be this or something close to it:
```
       FactTextField {
             id: factInput
             width : parent.width / 2
             height: parent.height
             horizontalAlignment: TextInput.AlignHCenter

             function setFactValue(newValue) {
                if (fact.validate(newValue, false /* convertOnly */) == "") {
                    fact.value = newValue
                }
                 factValueChangeDelay.start()
             }
         }
```

DonLakeFlyer:
Also can you rebase against latest master. I want to build this myself to run but can't since it's too far behind master and conflicts.

DonLakeFlyer:
Also some things for future revs:
- I'd like to see this just be another page/tab/something on the existing Tuning page. Not sure how the UI for that sort of thing should look like yet, since it would be the first example. Something like basic/advanced tabs.
- Also I'd like to see this look more like the Mission Planner screen: http://ardupilot.org/copter/docs/tuning.html. Where the groupings roll/pitch/... are in there own group box using the QGC style group box which looks like this:
screen shot 2017-01-31 at 8 33 43 pm. On a laptop this should then fix in a single screen so you don't have to scroll.

src/FactSystem/FactControls/FactStepper.qml
anton-matosov: This is the way to create `alias` for the signal. It connects inner signal with outer one. A kind of signal forwarding

src/FactSystem/FactControls/FactStepper.qml
anton-matosov: Good catch. Even compiler didn't complain about it...

src/FactSystem/FactControls/FactStepper.qml
anton-matosov: I followed analogy with `isNaN(fact.increment)` from the slider panel.
And they are not always set. It might be just bad data, but I don't own these values, so I would prefer to check. My default min/max might be not good tho.

src/FactSystem/FactControls/FactStepper.qml
anton-matosov: This is just a null check to avoid ton of errors in the console

anton-matosov: If fact is not set (yet) fall back to default values

src/FactSystem/FactControls/FactStepper.qml
anton-matosov: The idea of the factValueChangeDelay is to prevent to frequent updates of values, as it results on value set drop outs.
This is critical when WIFI or BT is used with QGC on mobile phone and user holds the +/- button to 'scroll' to needed value.

anton-matosov: I am still refactoring FactTextField to be support inheritance in a good way. My goal is to have only 1 data setting flow independently on the way of input

anton-matosov:
Mission Planner is totally unusable on the mobile, but I do 99% of tuning from small Android tablet.
It is also very user unfriendly and not extensible.
Current version of the PID tuning screen is foundation for future versions.
My goal is to have very easy UI to do tuning. At the very late stages I am going to implement a Tuning Wizard that allows novice user to tune their drones without spending hours searching for info on how to do that

Here is the copy of roadmap one more time:
Roadmap:

- v2 will have quick filters for simple/advanced settings set, pitch/roll lock
- v3 simple graph for TPA settings (it's much easier to understand this way)
- v4 attitude data plotting
- v5 regular and fine tune modes (fine tune reduces step size to 10% of the original one)
- v6 safety checks to prevent flyaway and `quick disarm` button
- v7 (new) PID tuning wizard targeting novice users

anton-matosov:
Latest version with validation handling
image

AndKe:
@anton-matosov I've been using APM Planner 2 for years, entering values in full paramater list, and Ctrl-W to send, - your approach seems very much like the most convenient GUI for this purpose I've ever seen.

DonLakeFlyer:
@anton-matosov Why are you circumventing the standard validation failure handling ui (and help) for parameters? This is a standard part of QGC ui framework which is used everywhere a parameter is edited. I don't want that to be different from one place to another.

DonLakeFlyer:
The group-boxing on your latest visual is not quite correct. There should be a single shaded rectangle for all the Roll values, then another shade rectangle for all the Pitch values and so forth. Not a separate shaded rectangle for each value. I'd suggest splitting up your list model into separate models for Roll/Pitch/... Then I think you can create an array of the ListModels and use a Repeater to draw instead of a ListView. This puts the grouping outside the stepper control. This would better fit changes later on where these group boxes should fill the screen left/right/up/down (like MissionPlanner) instead of just an up/down scrolling list. You want to be able to set all PIDs without the need to scroll if at all possible.

DonLakeFlyer:
Also just FYI: I've been planning on creating a mobile variant of FactTextField which is close to the Tower control which allows for decent increment on mobile. This would then show up everywhere you use a FactTextField in QGC on mobile. Just hadn't gotten around to it yet. It's a bit more complicated to make it work everywhere because you run into ui real estate issues.

anton-matosov:
@DonLakeFlyer 
> Why are you circumventing the standard validation failure handling ui (and help) for parameters?
 1. It is not usable on mobile, at all
 1. It is not usable even on desktop if you just want to increment/decrement parameter
 1. It is not standard, take a look at Tuning UI
 1. I am trying to improve UI and make it user friendly instead of being developer friendly

> The group-boxing on your latest visual is not quite correct. There should be a single shaded rectangle for all the Roll val
 1. Sorry, I am not going to blindly follow what MissionPlanner is doing, even if you are so used to it. It is not user and mobile friendly and/or scalable and not extensible 
 1. My PR is a contribution to the project, my proposal. If you want to do it your own way, please feel free to do it yourself. I am out.

anton-matosov:
This is v2 of UI
image

LorenzMeier:
@anton-matosov The v2 UI looks fantastic:

@DonLakeFlyer What Anton is contributing is really fantastic and really user driven.

DonLakeFlyer:
Sorry Anton I haven't been explaining myself well. I've been living out of the back of a camper van for the last 3 weeks and am still on the road as I type this. So I only have a few minutes here and there to look a things as I get internet through a straw at whatever random campground I'm at. I've been doing a poor job of keeping up and spending the right amount of time on important things. Let me try to explain myself better:

- The +/- increment support is great. I hope you didn't think I thought that was a bad idea. In reality this sort of thing should work on all FactTextFields when you are running on mobile. Since mobile numeric entry is so crappy. Hence my comment on how Tower does mobile well in this regard with there increment control (and QGC currently does not). I'll likely attempt to do that once I get time back home (put your increment stuff everywhere).
- My comments on looking at mission planner was mainly to be able to get more information onto a single screen when using a laptop or tablet. So that you didn't need to scroll to set multiple parameters. On laptops/desktops there is a lot of horizontal real estate you can take advantage of. You can also design the horizontal real estate usage such that you use it when available and don't when not by using either a Flow or laying out the group boxes yourself. As an example the Summary page works this way as do other setup pages so that they reflow based on available screen real estate (using the group box shading visual). I think your new v2 ui which I'm just seeing now is much better (less scrolling) and is fine to go in. 
- My only ui complaint was the new validation logic/ui (and help button) which is only triggered when the user types a bad value into the entry field. One of the reasons the standard mechanism is important is that is provides more than just validation. There are additional features in there such as: Force Save to override a value into a parameter which goes against the meta data min/max, also Restore to Default is in there which is handy to put things back to where they started. Another reason a standard app-wide validation result/help mechanism is important is that on mobile eventually I want the increment capability everywhere. Hence the logic/ui needs to work in a way that requires little ui redesign to make space for it. This is why it currently is a pop up side dialog control. You don't have to design UI specifically for it. If the standard validation side dialog thingy does not work on small phone size screens then that should be fixed globally in one place so that it works well everywhere. Would love to hear your input on possibilities for that. It's the first I've heard from folks that this wasn't workable on small screens.

Sorry I didn't spend the time to explain myself at all. I greatly appreciate your work on this. Should have said that before but I was going to fast with my available time. I am not even remotely an amateur on PID tuning. So experienced help here is hugely appreciated and will be a major addition to QGC. 

Normally if I was at home I would have had the time to devote to this and would have just pushed it through and let it grow in further pulls. Instead of piling on a bunch of crap change this, change that comments. I was trying to get you to deal with things I should have just handled myself later on. I'm usually devoting full time to QGC so what you're experiencing with me is hopefully out of the ordinary. I should be back home and back to full time on QGC in about a week.

anton-matosov:
Hey @DonLakeFlyer, thanks for your reply. Your points are well taken.
I am following agile approach in all my projects. I prefer to deliver to users as soon as possible (as soon as I have something working, because something is always better then nothing) and start collecting feedback on what is working and what can be improved.
Additionally I am always using what I have just build and iterate from there. I have started with standard FactTextEdit and its validation logic. It didn't work well, so I have start improving it. Actually I have started from scratch. The 3 main reasons for that were:
 1. Existing UI is very desktop like, and trying to adopt desktop to mobile doesn't work. That is why "Mobile First" design principle was born. So my goal is to first make UI work for mobile, then we can make use of bigger real estate
 1. Existing UI is widely used and I didn't want to break it right away with something that is not proven to be better. My approach is to try out new flow in some isolated piece of the app. If it proves it worthiness, we can integrate it elsewhere. This helps to move faster and don't waste time on massive refactoring that can give zero or negative improvement
 1. I needed the way to adjust sever parameters at the same time (e.g. Roll/Pitch P and D) and do it as fast as possible. Dialog based editors didn't allow me to do that. In mobile world you should do as much stuff inline as possible, and avoid intrusive experiences.

I am still going to follow my original roadmap (see below) and will provide incremental updates (PRs) aligned to it.

Roadmap:
- v2 will have quick filters for simple/advanced settings set, pitch/roll lock (edit. there will be filters for roll, pitch, yaw, tpa and P, I, D )
- v3 simple graph for TPA settings (it's much easier to understand this way)
- v4 attitude data plotting
- v5 regular and fine tune modes (fine tune reduces step size to 10% of the original one)
- v6 safety checks to prevent flyaway and quick disarm button
- v7 (new) PID tuning wizard targeting novice users

anton-matosov:
Most recent v2 UI
image
image

DonLakeFlyer:
> Existing UI is very desktop like, and trying to adopt desktop to mobile doesn't work. That is why "Mobile First" design principle was born. So my goal is to first make UI work for mobile, then we can make use of bigger real estate

FYI: The current ui design target for QGC is laptop or tablet and the ui is optimized for them (basically larger form factor than small phone). The QGC ui on small phone size screens should currently be usable but may still be painful to use in certain cases. Especially with respect to vehicle setup and mission planning. That should be possible but non-optimal. Flight monitoring on small screens should be good as well since there has been a focus there. A true optimal ui pass for small screen will happen, but probably not in this release since there is so much other work to do.

Can you reopen this pull so it can run against CI systems and check unit tests? Also I still want to run it myself to test for any detrimental affect on other FactTextField usage.

DonLakeFlyer:
Also can you let me know when this pull is in a stable state so I can complete the code review and merge?

anton-matosov:
> FYI: The current ui design target for QGC is laptop or tablet and the ui is optimized for them 
I don't want to be too negative about things, Don, but current QGC is purely desktop app (or laptop, which is the same thing in terms of UI format definitions) which was scaled down to fit into the tablet screen size. 
Making app a Mobile Friendly is not about fitting it into the device's screen size. It is about making it user centric. In most cases it is total rethink of how user interacts with app, what tasks user performs in which circumstances, how navigation between screens happens (preferred way is to use ios/android/windows platform specific format). These items are just tip of an iceberg, more is covered in Apple Human Interface Guidelines https://developer.apple.com/ios/human-interface-guidelines/overview/design-principles/ and Google Material Design https://developer.android.com/design/index.html

anton-matosov:
> Also can you let me know when this pull is in a stable state so I can complete the code review and merge?
I will create another PR once v2 is ready for merge. 
You can check it out from here https://github.com/anton-matosov/qgroundcontrol/tree/PID-Tuning-v2

DonLakeFlyer:
> I don't want to be too negative about things, Don, but current QGC is purely desktop app (or laptop, which is the same thing in terms of UI format definitions) which was scaled down to fit into the tablet screen size.

@anton-matosov There are important reasons for that. I'll write up an explanation separately in the dev guide then link to that.

DonLakeFlyer:
@anton-matosov QGC UI design pattern writeup: https://donlakeflyer.gitbooks.io/qgroundcontrol-developers-guide/content/UIDesignPattern.html

PR #4462, opened 2017-01-23
Add mouse area to the FactSliderPanel in order to avoid unwanted value changes while scrolling

anton-matosov:
Add mouse area to the FactSliderPanel in order to avoid unwanted value changes while scrolling, but still allow scrolling to change values if slider is explicitly focused

### DIFF
--- a/src/QmlControls/FactSliderPanel.qml
+++ b/src/QmlControls/FactSliderPanel.qml
@@ -12,6 +12,7 @@ import QtQuick              2.5
 import QtQuick.Controls     1.4
 
 import QGroundControl.FactSystem    1.0
+import QGroundControl.FactControls  1.0
 import QGroundControl.Palette       1.0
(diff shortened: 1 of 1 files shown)


src/QmlControls/FactSliderPanel.qml
```diff
@@ -110,6 +111,26 @@ Column {
                         onValueChanged: {
                             if (_loadComplete) {
                                 fact.value = value
+                                    }
+                                }
+
+                                updateValueWhileDragging:   true
```
DonLakeFlyer: Not sure this is a good idea. On small screens like a phone, dragging these things around can be a bit hard. Hence you hit a number of bad values while you are trying to get to the final right place. You don't want the intermediate values to make it to the vehicle.

anton-matosov: Ok. Makes sense.

PR #4463, opened 2017-01-23
Joystick calibration rework

jaxxzer:
This is a rework of the joystick calibration process to allow selecting a transmitter mode (TX mode) for the joystick input. The mode options are 1-4 as described [here](http://www.rc-airplane-world.com/rc-transmitter-modes.html). This option is placed under the Advanced Settings in the joystick configuration page.

Instead of mapping functions to raw axis indices, functions are mapped to real-life axes (stickLeftX, stickLeftY, stickRightX, stickRightY). This mapping is changed by selecting the joystick TX mode. The joystick TX mode setting is firmware-specific, so a user can use different TX modes for copter and sub, for example. The default TX mode is 2 for all firmwares except sub, whose default is 3.

The functionAxis saved settings have been replaced with real-life axis mappings, and firmware specific TX mode settings.

The attitude controls in the joystick configuration page will always reflect the current TX mode, and the text has been changed for Sub to show Forward/Lateral instead of Pitch/Roll.

screenshot 2017-01-23 15 41 01
screenshot 2017-01-23 15 41 17

The calibration process has changed by prompting the user to move the left/right sticks on the joystick instead of the throttle/pitch sticks. The real life axes are mapped to raw axis indices reported by the joystick in this step.

screenshot 2017-01-23 15 43 11

SDL2 provides a gamecontroller subclass of joysticks in which the real-life axis mappings are already done by SDL2, and all game controller axis mappings will be consistent. For this reason, game controllers will have a default calibration loaded, and won't require calibration. The option to (re)calibrate a gamecontroller joystick is still available, but the joystick will be enabled automatically without requiring an initial calibration.

### DIFF
--- a/src/FirmwarePlugin/APM/ArduSubFirmwarePlugin.h
+++ b/src/FirmwarePlugin/APM/ArduSubFirmwarePlugin.h
@@ -69,6 +69,8 @@ public:
     // Overrides from FirmwarePlugin
     int manualControlReservedButtonCount(void);
 
+    int defaultJoystickTXMode(void) final { return 3; }
+
     bool supportsThrottleModeCenterZero(void);
 
     bool supportsManualControl(void);
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -106,6 +106,11 @@ int FirmwarePlugin::manualControlReservedButtonCount(void)
     return -1;
 }
 
+int FirmwarePlugin::defaultJoystickTXMode(void)
+{
+    return 2;
+}
+
 bool FirmwarePlugin::supportsThrottleModeCenterZero(void)
 {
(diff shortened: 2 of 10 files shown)


DonLakeFlyer:
Is there a reason this can't work exactly like RC Cal handles transmitter modes? That way it doesn't require any code change more than showing the correctly different stick images. Much simpler also then RC and Joystick continue to mirror each other from a setup/functionality standpoint.

jaxxzer:
That's a good idea. From what I can see now, the downside would be requiring recalibration to change the mode, which is probably not an issue for most users. The default calibration on gamecontrollers would also require a default TX mode, or a TX mode in settings. I'll give it a shot and see if there is anything else that crops up.

DonLakeFlyer:
@jaxxzer Close this one right?

jaxxzer:
Yes.

PR #4465, opened 2017-01-24
Auto connect on start for UDP links

AlexeyBulatov:
This PR
- sets autoconnect flag on start for selected UDP link that helps if you use QGC for outgoing connection and for links with another listening port (not only for 14550). To use this feature set checkbox in the link configuration:
screenshot from 2017-01-24 16-15-33
-  changes return code of virtual method ```isAutoConnectAllowed``` from true to false. This way allowing to autoconnect only for UDP links and prohibiting for all other link types.

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -990,3 +990,14 @@ QList<LinkInterface*> LinkManager::links(void)
 
     return rawLinks;
 }
+
+void LinkManager::startAutoConnectedLinks(void)
+{
+    SharedLinkConfigurationPointer conf;
+
+    for(int i = 0; i < _sharedConfigurations.count(); i++) {
+        conf = _sharedConfigurations[i];
+        if (conf->isAutoConnect())
+            createConnectedLink(conf);
+    }
+}
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -439,6 +439,9 @@ bool QGCApplication::_initForNormalAppBoot(void)
                     "Your saved settings have been reset to defaults.");
     }
 
+    // Connect links with flag AutoconnectLink
+    toolbox()->linkManager()->startAutoConnectedLinks();
+
     if (getQGCMapEngine()->wasCacheReset()) {
         showMessage("The Offline Map Cache database has been upgraded. "
                     "Your old map cache sets have been reset.");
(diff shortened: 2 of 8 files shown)


DonLakeFlyer:
Thanks @AlexeyBulatov. Take me a day or two to look through this again.

DonLakeFlyer:
Sorry again for the delay on this. Still haven't made it back home yet to have time to look at this.

src/comm/LinkManager.cc
```diff
@@ -426,6 +426,9 @@ void LinkManager::loadLinkConfigurationList()
                                 pLink->setAutoConnect(autoConnect);
                                 pLink->loadSettings(settings, root);
                                 addConfiguration(pLink);
+                                if (pLink->isAutoConnect()) {
+                                    createConnectedLink(pLink);
```
DonLakeFlyer: @dogmaphobic I think the positioning of this is ok. My only concern is whether this point in the code is maybe a little too early in boot time to start a link? Any comments?

DonLakeFlyer:
@dogmaphobic Including Gus for opinion as well. The automatially connect check box is disabled on all other links types. Wondering about invisible instead. These disabled checkbox sort of things tend to confuse people as to why they are disabled and think something is wrong.

src/comm/LinkManager.cc
dogmaphobic: The loadLinkConfigurationList() function is called (from QGCApplication.cc) before checking to see if the settings are to be cleared, which would invalidate the settings loaded. The code above seems fine but the function call in QGCApplication.cc has to be moved after testing for `_settingsUpgraded`.

dogmaphobic:
> The automatially connect check box is disabled on all other links types. Wondering about invisible instead.

The original idea was that any comm link could be set to auto connect. That's why those check boxes are there. Since the time I wrote that, just about everything related to links have changed (it pre-dates any sort of automatic connection), which would justify a revisit in that entire comm link settings panel.

src/comm/LinkManager.cc
DonLakeFlyer: @AlexeyBulatov Can you make that change as well?

DonLakeFlyer:
@AlexeyBulatov Can you change the checkbox to be visible based on auto-connectability as opposed to enabled?

After the two changes this should be good to merge.

AlexeyBulatov:
@DonLakeFlyer Fixed and rebased on master.

DonLakeFlyer:
Awesome, thanks @AlexeyBulatov

PR #4466, opened 2017-01-24
MAVLinkLogManager: fix use of wrong variable

bkueng:
previously the start log button was always disabled, even for PX4.

Fixes regression from 9d413108f8116a5ba049f41d94929789f799a509

@dogmaphobic @ChristophTobler


dogmaphobic:
CI seems to be stalled. I will wait a bit longer and restart it if nothing changes.

dogmaphobic:
Mac OS builds are taking forever. One of them failed due to timeout. I restarted it.

dogmaphobic:
Restarting build...

PR #4470, opened 2017-01-25
Rework joystick calibration (2)

jaxxzer:
This is functionally equivalent to https://github.com/mavlink/qgroundcontrol/pull/4463, implemented differently according to Don's suggestion. The mode selection behavior in joystickConfigController now matches that of RC calibration.

screenshot 2017-01-24 23 33 31
screenshot 2017-01-24 23 33 44

The major difference is that selecting a mode just remaps the axes in _rgFunctionAxis, and the output logic and saved settings remain unchanged. The saved settings for the function->axis mappings are always saved as TX Mode 2 (current settings) mappings. A few settings are added for the saved joystick TX mode while running different airframes. The settings are forward/backward compatible tested on 3.1.1.

The calibration prompts haven't been changed in this one, and still ask to move throttle/pitch sticks, according to the selected mode.


src/Joystick/Joystick.cc
```diff
@@ -198,6 +286,42 @@ void Joystick::_saveSettings(void)
...
+        temp[_mapFunctionMode(newMode, function)] = _rgFunctionAxis[_mapFunctionMode(currentMode, function)];
+    }
+
+    for(int function = 0; function < maxFunction; function++) {
+        newMapping[function] = temp[function];
+    }
+
+}
+
+void Joystick::setTXMode(int mode) {
```
jaxxzer: This is to change the mode mapping after calibration has been performed. I disabled the mode mapping selection for joysticks during calibration. I noticed it stays enabled for RC calibration, but that should probably be disabled, too.

src/VehicleSetup/JoystickConfig.qml
```diff
@@ -643,12 +644,57 @@ SetupPage {
...
+
+                    QGCLabel {
+                        text: "TX Mode:"
+                    }
+
+                    QGCRadioButton {
+                        exclusiveGroup: modeGroup
+                        text:           "1"
+                        checked:        controller.transmitterMode == 1
+                        enabled:        !controller.calibrating
```
jaxxzer: Mode mapping selection is disabled here during calibration.

src/Joystick/Joystick.cc
jaxxzer: The TX mode can be changed before or after calibration, but not during. The TX mode is saved in settings, but the function axis settings are not affected. They are adjusted after the settings are loaded to match the TX mode setting, and they are adjusted again before saving. The function axis settings always reflect a mode 2 configuration (as they always have).

jaxxzer: To clarify,
> The function axis settings always reflect a mode 2 configuration (as they always have).
The _saved_ function axis settings always reflect a mode 2 configuration.

DonLakeFlyer: > The TX mode can be changed before or after calibration

Ok, that's kind of what I thought. My concern is that this makes joystick cal different from rc cal which does not support this. It also add an extra level of complexity. Do you think this is important. I'd rather keep it simpler and just like RC cal. But if it's good functionality then ok.

jaxxzer: I think this reflects a difference in joysticks and RC transmitters, RC transmitters may come built for a specific mode that is hard-wired. You are selecting the mode that corresponds to the RC transmitter in your hands. Joysticks aren't made this way, so you are selecting the mode that you want your joystick to emulate.

If we remove this functionality, then we won't be able to auto-calibrate gamecontrollers, we won't be able to have vehicle-specific settings, and we will have to recalibrate the joystick in order to change the mode. I don't know if it will be clear that you need to recalibrate joysticks in order to change the mode either. We could have a message somewhere that says you must recalibrate the joystick to change the mode. Or we could disable the mode selection at all times except, as the first step in calibration.

In my opinion it is 'good functionality', but I understand your concerns. It is not absolutely necessary to leave this in. I did structure the commits so it would be straightforward to take out or revert.

DonLakeFlyer: All good enough explanation for me

DonLakeFlyer:
@jaxxzer Sorry but I'm a little scattered on whether there was anything left to do here. Is this all good to merge now from where you stand? I'm clear from my end.

jaxxzer:
Forcing recalibration as an extra measure. This will handle case in which someone may have previously calibrated joystick and intentionally mapped sticks in a different way than the pictures presented during the calibration process in order to achieve their desired mappings.

Good to go for me.

DonLakeFlyer:
Ok. Thanks @jaxxzer

Issue #4426, opened 2017-01-12
Plan: MAV_CMD_NAV_LAND item should zero out altitude when added


DonLakeFlyer:

DonLakeFlyer:
Fixed

PR #4471, opened 2017-01-26
PX4 FW default landing altitude 0m

dagar:


DonLakeFlyer:
@dagar Does this work correctly? There is some code inside MissionItem i think which defaults altitude  from the C++ side of things. Not sure if the order is correct in that the json will change alt after that default is set.

dagar:
It worked in the one case I tried it, which was adding a series of waypoints and changing the last to a land. Is this the correct way to be setting defaults, or do we need to figure out how it works with the MissionItem code?

DonLakeFlyer:
Ok, the special altitude handling thing only applies to MAV_CMD_NAV_WAYPOINT so this will work fine.

DonLakeFlyer:
@dagar Looking at this some more. Shouldn't this apply to everything: PX4/APM all vehicle types?

DonLakeFlyer:
Related to #4426

dagar:
Yes, I suppose it should.

DonLakeFlyer:
Can you put in a issue for this pointing back to here. Then I'll fix it up.

PR #4472, opened 2017-01-26
reboot vehicle from param menu

dagar:
There's probably a better place for this, but some param changes require a reboot to take effect so this seemed like a good place to hide it.


DonLakeFlyer:
Agree on placement, but better than nothing for now.

PR #4473, opened 2017-01-26
guided mode altitude slider better maximum

dagar:
https://github.com/mavlink/qgroundcontrol/pull/4216


DonLakeFlyer:
Give me a day or so to look at this. Seems simple but I think there were some PX4/ArduPilot differences here with what the slider means.

DonLakeFlyer:
Ok, understand now. This is good.

PR #4476, opened 2017-01-27
Improve release logging, deal with MAVLinkDecoder shutdown

NaterGator:
Just a few simple tweaks to improve things. MAVLinkDecoder was not properly parented (the whole threading the right way can of worms) so this PR removes the parent pointer we aren't using (false sense of proper QObject parenting) and uses a signal to quit the thread and move the object back to the creating thread context. MainWindow waits 1s for the thread to finish and then invokes deleteLater on it so the object is destroyed when the event loop returns.

Also turned on contextual information for release builds in the message logger so we have a trace back in source files when users submit logs.

### DIFF
--- a/src/ui/MainWindow.cc
+++ b/src/ui/MainWindow.cc
@@ -274,6 +274,11 @@ MainWindow::MainWindow()
 
 MainWindow::~MainWindow()
 {
+    // Enforce thread-safe shutdown of the mavlink decoder
+    mavlinkDecoder->finish();
+    mavlinkDecoder->wait(1000);
+    mavlinkDecoder->deleteLater();
+
     // This needs to happen before we get into the QWidget dtor
     // otherwise  the QML engine reads freed data and tries to
     // destroy MainWindow a second time.
@@ -290,8 +295,7 @@ QString MainWindow::_getWindowGeometryKey()
 void MainWindow::_buildCommonWidgets(void)
 {
     // Add generic MAVLink decoder
-    // TODO: This is never deleted
-    mavlinkDecoder = new MAVLinkDecoder(qgcApp()->toolbox()->mavlinkProtocol(), this);
+    mavlinkDecoder = new MAVLinkDecoder(qgcApp()->toolbox()->mavlinkProtocol());
     connect(mavlinkDecoder.data(), &MAVLinkDecoder::valueChanged, this, &MainWindow::valueChanged);
 
     // Log player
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
The changes in moveToThread end up doing this same thing don't they? They also cause only this code to use a different coding pattern for thread creation than all the other places that create thread. I don't think that is a good thing even if it ends up being a little more correct.

DonLakeFlyer:
Or maybe I'm not understanding the movetothread changes?

src/ui/MainWindow.cc
```diff
@@ -274,6 +274,11 @@ MainWindow::MainWindow()
 
 MainWindow::~MainWindow()
 {
+    // Enforce thread-safe shutdown of the mavlink decoder
+    mavlinkDecoder->finish();
+    mavlinkDecoder->wait(1000);
```
DonLakeFlyer: A bit concerned about slowing down shutdown even more. I think you can do a wait using MultiSignalSpy on the thread closing out can't yet. I think there is some sort of signal there you can wait on?

NaterGator:
@DonLakeFlyer with respect to moveToThread:
The intended mechanism to use threads in Qt is via composition, not inheritance. We should instantiate a worker thread and move the QObject onto the worker thread event loop. In this design QThread parentless and cleaned up in the dtor of the new QObject and the created QObject lifetime is managed through signals/slots (connect creating QObject::destroyed to created QObject::deleteLater). In our implementation the thread and object are the same thing which breaks this design (if we stop the thread all event processing stops until the object is moved). When we stop the thread the event loop processing for the object dies so signals and slots stop working for it. The object stays resident in memory and isolated without any event loop processing. This change ensures that the QObject (in our case also the QThread) has a valid event loop to process signals/slots for it after the worker thread is shut down. 

To avoid the wait pattern (which is really a thread join with timeout and should return almost instantly) we can bind the QThreads stopped signal to QObject's deleteLater slot and trust the event loop will perform the delete next time it spins. 

Either way there isn't really much of a delay added assuming the thread has not deadlocked.

DonLakeFlyer:
I understand the correct pattern. Unfortunately no QGC code uses it. For me consistency if it ends up being the same thing ends up being better. But this is fine.

Also sorry, reading the QThread::wait call incorrectly. Thought it would wait always. This is fine.

NaterGator:
@DonLakeFlyer no harm in reality checking the intent/implementation of wait. I absolutely want to be meticulous about responsiveness. 

We can work on the threading patterns in a future release.

DonLakeFlyer:
@NaterGator Can you put in an Issue about fully cleaning all threading pattern code?

Issue #3980, opened 2016-08-26
Stop sending MANUAL_CONTROL messages when joystick is unplugged.


jaxxzer:
When the joystick is enabled, if the joystick is unplugged, qgc continues to send manual control messages with the last input received before the joystick was unplugged.

Joystick input does not pick up again when the joystick is plugged back in. The only way to regain control of the vehicle after the joystick is unplugged is to restart qgc.

DonLakeFlyer:
@rjehangir Something that will get better with SDL v2?

rjehangir:
Yes. Currently SDL1.2 is used on QGC and cannot detect when a device is removed, so there's no good way to do this.

DonLakeFlyer:
@rjehangir Fixable now?

rjehangir:
@DonLakeFlyer Yes, it should be fixable. This will be taken care of when the hotplugging feature is used. I've got branch I'm working on for that.

PR #4477, opened 2017-01-27
Joystick Hotplugging

jaxxzer:
I have tested this extensively under Windows and Ubuntu with multiple joysticks being connected/disconnected. When a Joystick is unplugged, the manual control messages stop, which fixes https://github.com/mavlink/qgroundcontrol/issues/3980.

An **important note** here, when a joystick is removed, the JoystickSDL thread associated with it continues to run, and the pointer to the object is lost. If the joystick is plugged back in, a new thread is started. I am not sure how to delete the thread safely, perhaps someone can lend a hand [here](https://github.com/mavlink/qgroundcontrol/pull/4477/files#diff-772bdcd04c41931b045c56907653a23cR92).

I have also had some infrequent issues with segfaults on SDL_JoystickClose, it seems like the pointer returned by SDL_JoystickOpen is not always valid. This is even the case when calling SDL_JoystickOpen, then immediately calling SDL_JoystickClose. It appears to me to be an issue with SDL. This has only popped up as an issue outside of normal use cases with multiple joysticks plugged in, where joysticks were being opened and closed very many times.


jaxxzer:
That appears to do the trick. Thank you for the hint, I will update this after I have worked it in.

src/Joystick/JoystickManager.cc
```diff
@@ -92,11 +92,10 @@ void JoystickManager::_setActiveJoystickFromSettings(void)
             qCDebug(JoystickManagerLog) << "Releasing joystick:" << i.key();
             if (_activeJoystick && !newMap.contains(_activeJoystick->name())) {
                 qCDebug(JoystickManagerLog) << "\twas active";
-                _activeJoystick->stopPolling();
-                _activeJoystick = NULL;
+                setActiveJoystick(NULL);
             }
-            // FIXME: Don't leave this hanging! We need to free the object.
-            //i.value()->deleteLater();
+            i.value()->wait(1000);
```
NaterGator: I believe you also want `i.value()->stopPolling();` here before the wait...

DonLakeFlyer:
> } else {
>     i.value()->stopPolling();
> }
> i.value()->wait(1000);
> i.value()->deleteLater();

Isn't there some sort of signal you can wait on with QSignalSpy instead of doing a wait?

src/Joystick/JoystickManager.cc
```diff
@@ -146,3 +187,31 @@ void JoystickManager::setActiveJoystickName(const QString& name)
...
     setActiveJoystick(_name2JoystickMap[name]);
 }
+
+void JoystickManager::_updateAvailableJoysticks(void)
+{
+#ifdef __sdljoystick__
+    SDL_Event event;
+    while (SDL_PollEvent(&event)) {
+    	switch(event.type) {
+        case SDL_QUIT:
```
DonLakeFlyer: Coding style issues here. ```case``` should line up with ```switch```

src/Joystick/JoystickManager.cc
```diff
@@ -146,3 +187,31 @@ void JoystickManager::setActiveJoystickName(const QString& name)
...
+            break;
+        case SDL_JOYDEVICEADDED:
+            qCDebug(JoystickManagerLog) << "Joystick added:" << event.jdevice.which;
+            _setActiveJoystickFromSettings();
+            break;
+        case SDL_JOYDEVICEREMOVED:
+            qCDebug(JoystickManagerLog) << "Joystick removed:" << event.jdevice.which;
+            _setActiveJoystickFromSettings();
+            break;
+		default:
```
DonLakeFlyer: ```default``` should line up with ```switch```

DonLakeFlyer:
Generally looks ok. But I don't have a joystick with me on the road to test out. If it's working fine for Nate as well then I'm ok with this.

DonLakeFlyer:
@NaterGator Can you run/try this as well?

NaterGator:
I'll try to give it a go tomorrow with some equipment but I'm actually doing some remote field work this week and not near my development machine at the office. Will report back tomorrow.

DonLakeFlyer:
@NaterGator Have you had time for this yet?

DonLakeFlyer:
@jaxxzer I'm back in town now. Can you fix conflict. Then I'll do a double-check test and merge.

DonLakeFlyer:
LGTM

Issue #4479, opened 2017-01-28
MAVLink 2.0 messages appearing in inspector but not analyze view


jgoppert:
![image](https://cloud.githubusercontent.com/assets/473772/22396428/5d800730-e527-11e6-90b3-6fbcc4536008.png)

This doesn't make mush sense to me. I must be missing something obvious.

mhkabir:
I can confirm this. The messages in the v2 space aren't coming up in plots.

mhkabir:
Fixed here : https://github.com/jgoppert/qgroundcontrol/commit/ff3b9727c9077fb1777d218eff02c4352187f738

PR #4480, opened 2017-01-28
Fix analyze view to work for mavlink 2.0

jgoppert:
This fixes integer overflow issues with mavlink 2 msgid's and get the analyze view working for mavlink 2 messages.

Closes https://github.com/mavlink/qgroundcontrol/issues/4479

image

### DIFF
--- a/src/ui/MAVLinkDecoder.h
+++ b/src/ui/MAVLinkDecoder.h
@@ -2,7 +2,27 @@
 #define MAVLINKDECODER_H
 
 #include <QObject>
+#include <QHash>
+
 #include "MAVLinkProtocol.h"
+
+struct SystemData {
+    /**
+     * @brief Holds mavlink system data
+     */
+    SystemData() :
+        componentID(-1),
+        componentMulti(false),
+        onboardTimeOffset(0),
+        onboardToGCSUnixTimeOffsetAndDelay(0),
+        firstOnboardTime(0) {
+    }
+    int componentID;            ///< Multi component detection
+    bool componentMulti;        ///< Multi components detected
+    quint64 onboardTimeOffset;  ///< Offset of onboard time from Unix epoch (of the receiving GCS)
+    qint64 onboardToGCSUnixTimeOffsetAndDelay; ///< Offset of onboard time and GCS Unix time
+    quint64 firstOnboardTime;   ///< First seen onboard time
+};
 
 class MAVLinkDecoder : public QThread
 {
@@ -26,16 +46,10 @@ protected:
     /** @brief Shift a timestamp in Unix time if necessary */
     quint64 getUnixTimeFromMs(int systemID, quint64 time);
 
-    static const size_t cMessageIds = 256;
-
-    mavlink_message_t receivedMessages[cMessageIds];        ///< Available / known messages
     QMap<uint16_t, bool> messageFilter;                     ///< Message/field names not to emit
     QMap<uint16_t, bool> textMessageFilter;                 ///< Message/field names not to emit in text mode
-    int componentID[cMessageIds];                           ///< Multi component detection
-    bool componentMulti[cMessageIds];                       ///< Multi components detected
-    quint64 onboardTimeOffset[cMessageIds];                 ///< Offset of onboard time from Unix epoch (of the receiving GCS)
-    qint64 onboardToGCSUnixTimeOffsetAndDelay[cMessageIds]; ///< Offset of onboard time and GCS Unix time
-    quint64 firstOnboardTime[cMessageIds];                  ///< First seen onboard time
+    QHash<int, mavlink_message_t> msgDict; ///< dictionary of all mavlink messages
+    QHash<int, SystemData> sysDict; ///< dictionary of all systmes
     QThread* creationThread;                                ///< QThread on which the object is created
 };
 
--- a/src/ui/MAVLinkDecoder.cc
+++ b/src/ui/MAVLinkDecoder.cc
@@ -10,16 +10,6 @@ MAVLinkDecoder::MAVLinkDecoder(MAVLinkProtocol* protocol) :
     // http://blog.qt.digia.com/blog/2010/06/17/youre-doing-it-wrong/
     moveToThread(this);
 
-    memset(receivedMessages, 0, sizeof(mavlink_message_t)*256);
-    for (unsigned int i = 0; i<255;++i)
-    {
-        componentID[i] = -1;
-        componentMulti[i] = false;
-        onboardTimeOffset[i] = 0;
-        onboardToGCSUnixTimeOffsetAndDelay[i] = 0;
-        firstOnboardTime[i] = 0;
-    }
-
     // Fill filter
     // Allow system status
 //    messageFilter.insert(MAVLINK_MSG_ID_HEARTBEAT, false);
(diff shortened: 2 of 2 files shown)


src/ui/MAVLinkDecoder.h
```diff
@@ -25,7 +25,7 @@ public slots:
     /** @brief Shift a timestamp in Unix time if necessary */
     quint64 getUnixTimeFromMs(int systemID, quint64 time);
 
-    static const size_t cMessageIds = 256;
+    static const size_t cMessageIds = 512;
```
mhkabir: Why 512? 
Vendor-specific messages in v2 are allocated higher IDs : https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/uAvionix.xml

We can probably remove this check completely : 
```
if (message.msgid >= cMessageIds) {
       // No support for messag ids above 255
       return;
}
```

jgoppert: well it actually allocates a buffer that big, so we might need to figure out another way to do this then, allocating an array of 10,000+ pointers might be ugly, not out of the realm of possibility. It would probably be better to use a std::map so it can be dynamically sized.

jgoppert: https://github.com/mavlink/qgroundcontrol/blob/e9bee197f6e47a19dfd256a3c347d3dcb43597b9/src/ui/MAVLinkDecoder.h#L30

jgoppert: @dagar are these test failures something I should worry about?

dagar:
Yes those failures look real.

    ********* Start testing of ComplexMissionItemTest *********
    Config: Using QtTest library 5.5.1, Qt 5.5.1 (x86_64-little_endian-lp64 shared (dynamic) release build; by     GCC 4.9.1 20140922 (Red Hat 4.9.1-10))
    PASS   : ComplexMissionItemTest::initTestCase()
    QWARN  : ComplexMissionItemTest::_testDirty() QObject::connect: invalid null parameter
    /home/travis/build.sh: line 57: 11797 Aborted                 (core dumped) ./debug/QGroundControl --unittest

You can run the tests locally with the debug build `QGroundControl --unittest`.

DonLakeFlyer:
From my initial quick look at supporting mavlink 2 messages in analyze it didn't seem easily supported because of the the message id space changes. The real fix is much more complicated than this.

jgoppert:
@DonLakeFlyer To fit the full message space yes, but there are a lot of new messages in this lower area that we can easily support with this patch. To support the full message range we will need to switch from an array to a dictionary.

DonLakeFlyer:
Yeah, I can live with that. Analyze is desktop only so extra memory isn't as much of a problem. Can you put in an issue about fixing up decoder to correctly handle all messages?

DonLakeFlyer:
I'll take a look at the UT failures. Not sure what is going on there, seems strange. May not get done with that till tomorrow.

jgoppert:
Yes, will do that.

jgoppert:
Posted it here: https://github.com/mavlink/qgroundcontrol/issues/4489

DonLakeFlyer:
@jgoppert In looking quickly at the MavlinkDecoder code I think it should actually be pretty easy to fix this for real for all message ids. Just change these:
```
    mavlink_message_t receivedMessages[cMessageIds];        ///< Available / known messages
    int componentID[cMessageIds];                           ///< Multi component detection
    bool componentMulti[cMessageIds];                       ///< Multi components detected
    quint64 onboardTimeOffset[cMessageIds];                 ///< Offset of onboard time from Unix epoch (of the receiving GCS)
    qint64 onboardToGCSUnixTimeOffsetAndDelay[cMessageIds]; ///< Offset of onboard time and GCS Unix time
    quint64 firstOnboardTime[cMessageIds];                  ///< First seen onboard time
```

To be QMap's instead of arrays with the map key being the message id (uint32_t). Then you don't need cMessageIds at all and you can handle an id of any value. Should be an easy change. Could you do that instead?

DonLakeFlyer:
@jgoppert Can you make the change to QMaps? If not, no big deal I'll do it. It's simple but won't get around to it for a week when I finally get back home.

jgoppert:
Yeah, will do, I just got swamped this weekend. Patching now.

jgoppert:
@DonLakeFlyer I tested this against PX4 sitl and was able to receive messages.  I need to rebase the mavlink 2 work I'm doing on this to make sure it works fine for mavlink 2 msg id's, but there is not reason it shouldn't since it is all dictionary based now. Also, while looking at QMap docs I ran into QHash which seemed to be more what we wanted and gave a faster lookup time. Once I test with mavlink 2 messages I think it will be good to merge.

jgoppert:
It looks like we have some unrelated, as far as I can tell, bugs in travis @dagar @DonLakeFlyer :

```bash
The command "if [ "${SPEC}" != "macx-ios-clang" ]; then make -j$JOBS; else xcodebuild -IDEBuildOperationMaxNumberOfConcurrentCompileTasks=$JOBS -configuration Release CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO | xcpretty -c && $(exit ${PIPESTATUS[0]}); fi" exited with 0.
39.93s$ if [ "${TRAVIS_BRANCH}" != "master" ]; then if [[ "${SPEC}" = "linux-g++-64" && "${CONFIG}" = "debug" ]]; then mkdir -p ~/.config/QtProject/ && cp ${TRAVIS_BUILD_DIR}/test/qtlogging.ini ~/.config/QtProject/ && ./debug/QGroundControl --unittest; elif [[ "${SPEC}" = "macx-clang" && "${CONFIG}" = "debug" ]]; then mkdir -p ~/Library/Preferences/QtProject/ && cp ${TRAVIS_BUILD_DIR}/test/qtlogging.ini ~/Library/Preferences/QtProject/ && ./debug/qgroundcontrol.app/Contents/MacOS/QGroundControl --unittest; fi fi
Settings location "/home/travis/.config/QGroundControl.org/QGroundControl_unittest.ini" Is writable?: true
Filter rules "qt.network.ssl.warning=false\n*Log.debug=false\n"
MAVLinkLogManagerLog: MAVLink logs directory: "/home/travis/.local/share/QGroundControl.org/QGroundControl_unittest/MAVLinkLogs"
VideoReceiver::start() failed. Error with data source for gst_element_factory_make()
VideoReceiver::start() failed
```

DonLakeFlyer:
@dogmaphobic Has anything changed with respect to build files and video?

DonLakeFlyer:
Interesting data on QMap versus QHash: https://woboq.com/blog/qmap_qhash_benchmark.html

jgoppert:
@DonLakeFlyer I had a look at the post, thanks for that. After reading that which are you leaning towards?  Typically we send around 20-ish messages which suggests from the post the performance may be similar. I think the worst case is for a small number of elements QHash will take about twice as long, but this lookup time is probably all in the wash anyways. I'm willing to change if you would prefer QMap. I think it is just a string replace QHash for QMap.

DonLakeFlyer:
I think you are right with QHash.

DonLakeFlyer:
I'm trying to repro the UT crash.

jgoppert:
Did you mean to close this?

DonLakeFlyer:
Crap I thought you had put in a new pull for the changes. My bad. Thought this was old hence I wasn't paying attention to it!

jgoppert:
Yeah, the name needs to be updated.

DonLakeFlyer:
Can you rebase to master. It has a needed unit test fix to get them to pass.

jgoppert:
OK, rebased.

PR #4486, opened 2017-01-30
WIP: Support for mavlink dynamic field estimator messages

jgoppert:
We are using this for the development of IEKF but it is made to be general and work with all estimators, such as LPE/ EKF2. @LorenzMeier @priseborough

image

It is now functional, and you can test it with the iekf branch of px4 which has the new messages and is using them. There are also uORB backend equivalents.

https://github.com/PX4/Firmware/pull/6045

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "libs/mavlink/include/mavlink/v1.0"]
 	path = libs/mavlink/include/mavlink/v1.0
-	url = https://github.com/mavlink/c_library_v1.git
+	url = https://github.com/jgoppert/c_library_v1.git
 [submodule "src/GPS/Drivers"]
 	path = src/GPS/Drivers
 	url = https://github.com/PX4/GpsDrivers.git
 [submodule "libs/mavlink/include/mavlink/v2.0"]
 	path = libs/mavlink/include/mavlink/v2.0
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,9 +1,9 @@
 [submodule "libs/mavlink/include/mavlink/v1.0"]
 	path = libs/mavlink/include/mavlink/v1.0
-	url = https://github.com/mavlink/c_library_v1.git
+	url = https://github.com/jgoppert/c_library_v1.git
```
jgoppert: I switched this over temporarily to build the new messages for me on a separate repo.

src/ui/MAVLinkDecoder.cc
```diff
@@ -305,6 +305,239 @@ void MAVLinkDecoder::emitFieldValue(mavlink_message_t* msg, int fieldid, quint64
...
+            {
+                uint8_t id_val = id_array[j];
+                QString id = QString("%1").arg(id_val);
+                switch (id_val) {
+                case MAV_FIELD_UNUSED:
+					// use number
+                    break;
+                case MAV_FIELD_POS_N:
+                    id = "pos_N";
+                    break;
```
jgoppert: These switch cases for turning the fields and sensor enums into strings could move into some mavlink util header if that would be more appropriate, it would make it simpler for users and less bloated here.

PR #4490, opened 2017-01-31
More tweaks

dogmaphobic:
* Remove some unused code in MainToolbarController
 * Allow plugin to define fixed font size, toolbar height, etc.

### DIFF
--- a/src/QmlControls/ScreenTools.qml
+++ b/src/QmlControls/ScreenTools.qml
@@ -47,7 +47,7 @@ Item {
     property real largeFontPointSize:       10
 
     property real availableHeight:          0
-    property real toolbarHeight:            defaultFontPixelHeight * 3
+    property real toolbarHeight:            0
 
     readonly property real smallFontPointRatio:      0.75
     readonly property real mediumFontPointRatio:     1.25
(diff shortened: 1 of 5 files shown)


src/QmlControls/ScreenTools.qml
```diff
@@ -107,7 +108,10 @@ Item {
         property real   fontWidth:    contentWidth
         property real   fontHeight:   contentHeight
         Component.onCompleted: {
-            var baseSize = QGroundControl.baseFontPointSize;
+            var baseSize = QGroundControl.corePlugin.options.defaultFontPointSize
```
DonLakeFlyer: I still think this would be better done through the ability for a plugin to override any setting. That way it requires no code change in the mainline code.

src/api/QGCOptions.h
```diff
@@ -29,6 +29,9 @@ class QGCOptions : public QObject
     Q_PROPERTY(bool     definesVideo                READ definesVideo               CONSTANT)
     Q_PROPERTY(uint16_t videoUDPPort                READ videoUDPPort               CONSTANT)
     Q_PROPERTY(QString  videoRSTPUrl                READ videoRSTPUrl               CONSTANT)
+    Q_PROPERTY(double   toolbarHeightMultiplier     READ toolbarHeightMultiplier    CONSTANT)
+    Q_PROPERTY(double   defaultFontPointSize        READ defaultFontPointSize       CONSTANT)
+    Q_PROPERTY(bool     enablePlanViewSelector      READ enablePlanViewSelector     CONSTANT)
```
DonLakeFlyer: This is ok for now I think. But it would be way better for it to be on FirmwarePlugin and on a per-vehicle basis. This ability is something folks have asked for. For example ArduSub doesn't support Missions yet, but it is a superclass of ArduPilot which does support missions. Would be nice if they could turn off Plan when connected to an ArduSub.

src/QmlControls/ScreenTools.qml
dogmaphobic: It needs to define a baseline default. I have to rethink this whole thing at some point. Once I do that, we can go through these and make them right.

PR #4501, opened 2017-02-01
Add support for the AUAV X2.1 Flight Controller

pkocmoud:
Don - Please review my boardFallback entries.

Note - Line location was chosen to fall in productID numeric order.


src/VehicleSetup/FirmwareUpgradeController.cc
```diff
@@ -407,6 +418,8 @@ QHash<FirmwareUpgradeController::FirmwareIdentifier, QString>* FirmwareUpgradeCo
     case Bootloader::boardIDAeroCore:
         return &_rgAeroCoreFirmware;
     case Bootloader::boardIDMINDPXFMUV2:
```
davids5: Duplicate case it should be ```boardIDAUAVX2_1```

DonLakeFlyer:
All looks good. Could you try connecting and flashing on Windows with a build which has the fallbacks for you boards removed. I think they may not be needed. If both work fine without them you can remove them and this should be good to go.

pkocmoud:
Removed the extra entries, tested good. Can you please merge?

pkocmoud:
@DonLakeFlyer Please merge this.

dagar:
I don't have this hardware, but the changes look correct.

Issue #4500, opened 2017-02-01
Parameter non default color


ChristophTobler:
![qgc_color_non_default](https://cloud.githubusercontent.com/assets/14769636/22508048/8835b308-e888-11e6-8378-dfd77e1f984f.png)
I don't think that black on dark grey is a very good choice ;)

DonLakeFlyer:
@dogmaphobic Most likely related to your recent palette changes?

dogmaphobic:
I only changed one color. The yellow of the "Active Button" is now more subdued. I will check this and see what's going on...

DonLakeFlyer:
Hmm, looking at code this should be warningText color from palette which is unrelated.

PR #4510, opened 2017-02-03
Fix wrong QColor argument encoding.

dogmaphobic:
Fix for #4500 

It was using a bogus color encoding and QColor was taking is as zero (black). I'm not sure what I did as that wasn't the color I was editing when I got this change in.


DonLakeFlyer:
Thanks Gus

PR #4515, opened 2017-02-06
Tidying up Mission Settings and Survey Settings

dogmaphobic:
Mostly making the controls sizes more consistent and legible. This is a first stab at this. There is still way too much info crammed into one dialog. Help text also makes it too busy but not sure what to do about them.

It was this:
old

And for now it looks like this:
new


DonLakeFlyer:
Did you do the same set of changes for SimpleMissionItem as well? If not, It's going to be out of whack now.

dogmaphobic:
> Did you do the same set of changes for SimpleMissionItem as well?

Nope. How do I get there?

DonLakeFlyer:
My bad I just went and looked at it. I thought it had some of those text+line header thingies in it. So I thought it was out of whack with that. But it doesn't so they are fine. 

I had another pull which cleaned up the sizing of the fields as well by using a GridLayout. I'll use that to deal with this one.  So you are good.

PR #4519, opened 2017-02-06
Add PX4 default meta data

LorenzMeier:


DonLakeFlyer:
@LorenzMeier Is this built against latest master. Seems to be hitting a UT crash which I thought was fixed a while back.

DonLakeFlyer:
@LorenzMeier Can you rebase. This is hitting a unit test problem which was fixed in master.

PR #4531, opened 2017-02-09
Dealing with lack of Internet connectivity.

dogmaphobic:
When you have no Internet connection (Tablet out in the field), tile downloading can bog the system down. Specially on Android. That's because the system will try to download all and every tile. As each request normally times out in around a minute, you ended up with hundreds of tile requests in the queue.

This change keeps track of the availability of (real) Internet connection. If there is none, QGC won't bother trying to download any tiles. If you have them cached, they will be displayed. Otherwise it will immediately show a "No Tile" tile.

### DIFF
--- a/src/QtLocationPlugin/QGCMapEngine.cpp
+++ b/src/QtLocationPlugin/QGCMapEngine.cpp
@@ -130,11 +130,13 @@ QGCMapEngine::QGCMapEngine()
     , _maxMemCache(0)
     , _prunning(false)
     , _cacheWasReset(false)
+    , _isInternetActive(false)
 {
     qRegisterMetaType<QGCMapTask::TaskType>();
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
Would be great to have the lack/availability of an internet connection as a method available anywhere. Other places this could be used to give better errors.

dogmaphobic:
Makes sense. I'll move it up the chain but I'm not sure the test I'm using will work everywhere on the planet. I try to see if 8.8.8.8:53 (google DNS) is responding. Isn't Google blocked in some places?

DonLakeFlyer:
Hmm yeah. I didn't look at the code. I figured it was some internal Qt comm check or something. Hit qgroundcontrol.com instead?

PR #4540, opened 2017-02-13
Parameter editor dialog mobile

DonLakeFlyer:
@anton-matosov pointed out that the parameter editor/validation dialog was majorly busted/unusable on small screens. Fixes:

* Expand hit area for Help button on mobile builds to make it easier to touch
* Rearranged items in dialog so that they fit on our target min screen size (iPhone 5s)

iPhone 5s screen shot:
img_1361


DonLakeFlyer:
FYI: There still seem to be some virtual keyboard problems on iOS which I am looking. At various point the virtual keyboard refues to come up.

PR #4545, opened 2017-02-14
Experimenting with Video / Photo commands

dogmaphobic:
Chicken and the egg problem. I need this on this end so I can test the firmware side. This is just the beginning of what will eventually become video / photo control.


DonLakeFlyer:
Where is it? Screenshots?

Issue #3725, opened 2016-07-04
Land abort button (MAV_CMD_DO_GO_AROUND)


dagar:
I'm implementing landing abort for PX4 fixed wing so I'd like to add a nice big abort button only when the vehicle is actually landing. 

How should I do this?
Unless I'm missing something obvious there's currently no way for the GCS to know if the mission is on on a landing waypoint?

DonLakeFlyer:
You'd need to plumb an isLanding into FirmwarePlugin. Then figure out some way to determine that (not sure how you do that). Then tweak the guided buttons to change based on isLanding. Also need top check for support for GO_AROUND in supportedCommands.

DonLakeFlyer:
Vehicle gets current mission item index updates. If the current mission item (you can look it up in MissionManager) is a landing one (not sure what that is) set an isLanding bool in the Vehicle?

dagar:
Isn't it possible we don't have the mission locally?

DonLakeFlyer:
We always have the mission locally that we are flying. That's how we show waypoints onthe Fly view. If we don't or it is out of date, that is a bug. Only way to be out of date (without a bug) would be is the Vehicle modified the mission on it's own. But that would be incorrect firmware I think.

dagar:
Alright that sounds acceptable for now. Maybe with mavlink 2 I could add an isLanding to EXTENDED_SYS_STATE.

Antiheavy:
Interesting topic.  I realize this is old, but I have a suggestion:

What if the [Abort] button is always shown on screen after takeoff, maybe with the other action buttons at the bottom of QGC.  Anytime the [Abort] button is pressed MAV_CMD_DO_GO_AROUND is sent and the aircraft will climb to a safe altitude while roll limited and then loiter above the Land waypoint. 

Would this show up as an "abort" mode type?  Is my understanding correct that any other command will exit the abort mode - e.g. pressing RTL again to go back and try another landing?  Should this be tied to having a valid Land waypoint?

cheers.

dagar:
Most of the the specifics of MAV_CMD_DO_GO_AROUND will be a PX4 issue. 
There's not a lot of detail in the mavlink spec. http://mavlink.org/messages/common (search for MAV_CMD_DO_GO_AROUND) and QGC is only responsible for sending the command (if we can find a way to display it at the right time).

dagar:
Implemented

PR #4546, opened 2017-02-14
Addition of "Abort" button

dkschrad:
Added an "Abort" button to allow the user to abort the current landing attempt by sending MAV_CMD_DO_GO_AROUND.  Currently, Abort button is always displayed during flight.

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewWidgets.qml
+++ b/src/FlightDisplay/FlightDisplayViewWidgets.qml
@@ -191,6 +191,7 @@ Item {
         readonly property int confirmGoTo:          8
         readonly property int confirmRetask:        9
         readonly property int confirmOrbit:         10
+        readonly property int confirmAbort:         11
 
         property int    confirmActionCode
         property real   _showMargin:    _margins
@@ -237,6 +238,9 @@ Item {
                 _activeVehicle.guidedModeOrbit()
                 //-- Center on current flight map position and orbit with a 50m radius (velocity/direction controlled by the RC)
                 //_activeVehicle.guidedModeOrbit(QGroundControl.flightMapPosition, 50.0)
+                break;
+            case confirmAbort:
+                _activeVehicle.abortLanding()
                 break;
             default:
                 console.warn(qsTr("Internal error: unknown confirmActionCode"), confirmActionCode)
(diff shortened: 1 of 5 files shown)


src/FlightDisplay/FlightDisplayViewWidgets.qml
```diff
@@ -356,6 +363,13 @@ Item {
                     onClicked:  _guidedModeBar.confirmAction(_guidedModeBar.confirmOrbit)
                 }
 
+                QGCButton {
+                    pointSize:  _guidedModeBar._fontPointSize
+                    text:       qsTr("Abort")
+                    visible:    _activeVehicle && _activeVehicle.flying
```
DonLakeFlyer: Should be ```visible: _activeVehicle && _activeVehicle.flying && _activeVehicle.fixedWing``` since this is only for fixed wing isn't it?

src/Vehicle/Vehicle.cc
```diff
@@ -1838,6 +1838,16 @@ void Vehicle::pauseVehicle(void)
...
 }
 
+void Vehicle::abortLanding(void)
+{
+    sendMavCommand(defaultComponentId(),
+                   MAV_CMD_DO_GO_AROUND,
+                   true,        // show error if fails
+                   50);         // this is a dummy value that is currently ignored
+
+    _firmwarePlugin->abortLanding(this);
```
DonLakeFlyer: Not sure why any of this needs to be in FirmwarePlugin. This is a generic mavlink command which works both on PX4 and ArduPilot stacks. So there shouldn't be anything that is vehicle specific.

src/Vehicle/Vehicle.h
```diff
@@ -355,6 +355,9 @@ class Vehicle : public FactGroup
     /// Command vehicle to kill all motors no matter what state
     Q_INVOKABLE void emergencyStop(void);
 
+    /// Command vehicle to abort landing
+    Q_INVOKABLE void abortLanding(void);
```
DonLakeFlyer: Can you change this to: ```abortLanding(double climbOutAltitude)``` and pass in the hardcoded value of 50 into it? That makes it more future ready.

src/FlightDisplay/FlightDisplayViewWidgets.qml
dagar: Yes it's only fixed wing.

DonLakeFlyer:
Also can you add an Issue about implementing something for the climb out altitude value. Don't want to forget that is currently hardcoded.

dagar:
Closes https://github.com/mavlink/qgroundcontrol/issues/3725

DonLakeFlyer:
Looks fabulous now, thanks. I'm following the build failures. So far looks like travis problems, not your problems.

DonLakeFlyer:
Finally got this all the way through Travis after fighting with it for days!

PR #4552, opened 2017-02-15
Switch to Qt5.8 [WIP -Do not merge]

DonLakeFlyer:


dogmaphobic:
`QStyleOptionFrameV3`

Yes, I've run into that. I've switched to 5.8 since it came out as a Release Candidate with no issues other than those warnings. Assuming we should be replacing that code at some point any way, I would simply silence the warnings for the qwt module in the mean time.

Issue #4434, opened 2017-01-13
ArduPilot: Complete work on FRAME_CLASS, FRAME_TYPE


DonLakeFlyer:
Need to add back parameter setting

Issue #4374, opened 2016-12-31
ArduPilot: Support new FRAME_CLASS, FRAME_TYPE airframe setup


DonLakeFlyer:
Also need to look at the firmware collapsing in Copter going on with this.

rmackay9:
For AC3.5 we are introducing a FRAME_CLASS parameter and renaming FRAME to FRAME_TYPE so it would be good to update the frame selection screen.

Possible values from FRAME_CLASS are 0:Undefined, 1:Quad, 2:Hexa, 3:Octa, 4:OctaQuad, 5:Y6, 6:Heli, 7:Tri, 8:SingleCopter, 9:CoaxCopter
Possible values for FRAME_TYPE are 0:Plus, 1:X, 2:V, 3:H, 4:V-Tail, 5:A-Tail, 10:Y6B

In case it helps, this is the equivalent screen I've mocked up for MP.  I'm sure QGroundControl will look very different but perhaps it gives a clue as to what it could look like.
![frameclasstype](https://cloud.githubusercontent.com/assets/1498098/21575403/62cb1f70-cf4e-11e6-9ec1-445c83aa4f74.png)

Equivalent to-do for MP: https://github.com/ArduPilot/MissionPlanner/issues/1442

OmahaCounsel:
This issue is NOT solved. There is no way to configure a y6b . The coax is the only choice. Even after loading the developer build.

DonLakeFlyer:
@OmahaCounsel Do you mean this configuration:
<img width="386" alt="screen shot 2017-02-18 at 2 22 10 pm" src="https://cloud.githubusercontent.com/assets/5876851/23097313/b01a9f90-f5e5-11e6-97ae-89a7f0eb0a0b.png">

What OS are you using? Maybe the build drops aren't working correctly.

OmahaCounsel:
I have never see that screen.
My operating system is Windows 10 home. 64 bit running on a Samsung notebook.
I tried the qground control with both with and with out the px4 stack. Would only work with px4 stack but then no option for  a y6 b.

With respect,



Brent W. Nicholls, J.D., LL.M
Kasaby & Nicholls, LLC
300s. 19th Street #300
Omaha, NE 68102
402.502.0600 (p)
402.615.5926.(f)


On Feb 18, 2017, at 4:23 PM, Don Gagne <notifications@github.com<mailto:notifications@github.com>> wrote:


[...]

PR #4557, opened 2017-02-16
ArduPilot: Load params from new Airframe page

DonLakeFlyer:
Fix for #4374, #4434


OmahaCounsel:
There is still. No Option For a y6b

Issue #2910, opened 2016-02-28
can't connect to Pixhawk via Bluetooth on Android


kd0aij:
"service not available"

DonLakeFlyer:
@dogmaphobic Should bluetooth work on Android?

dogmaphobic:
What device? The TBS Crossfire works. At least it was working the last time I used it.

gregswf:
I'm using bluetooth with PX4 stack, it works fine on ASUS zenpad 7 but it does not connect on Galaxy Note4, I get "service not available"

dogmaphobic:
I need to know what "Bluetooth" device is plugged to the Pixhawk.

kd0aij:
Mine shows up as HC-06 and the chip is labeled BT417

dogmaphobic:
This? I need to get one to test. The only Bluetooth device I have is the Crossfire.

![hc-06](https://cloud.githubusercontent.com/assets/749243/15866898/af01ce68-2cae-11e6-98f4-874adf3b1f00.jpg)

dogmaphobic:
I found a datasheet for it. Did you have to set it up using the "AT" commands as described?

https://www.olimex.com/Products/Components/RF/BLUETOOTH-SERIAL-HC-06/resources/hc06.pdf

kd0aij:
It's been 2.5 years since I got mine. Don't remember, but I probably did set the baud rate using an AT command first.

kd0aij:
And that one looks almost identical to mine: http://smile.amazon.com/Kootek%C2%AE-Bluetooth-Serial-Module-Arduino/dp/B00DQ4A7O6?ie=UTF8&psc=1&redirect=true&ref_=oh_aui_search_detailpage

dogmaphobic:
Got it in the mail today.

The good news, I can't get it to connect either. Meaning I could reproduce the issue and I can debug it. The bad news, I won't be able to do much for the next couple of weeks 8P

DonLakeFlyer:
@dogmaphobic Do you think this is important enought to hold 3.0 release for?

dogmaphobic:
Good question. Bluetooth was ever really tested using only the Crossfire. Today was the first time I connected something else. It took me an hour just to get the thing configured and paired with a Nexus 9. Bluetooth is inherently unreliable. 

[...]

PR #4565, opened 2017-02-17
Fix regression bug with Bluetooth.

dogmaphobic:
This is in response to #2910.
It fixes a bug introduced by #4349.
The Bluetooth configuration was not being set and the code would crash when using it.


DonLakeFlyer:
Thanks Gus. In all honesty I bet there are still some problems lurking out ther from that pull. I kind of expected it since it was such a major PITA change.

PR #4566, opened 2017-02-17
Fly View: Use new ToolStrip control

DonLakeFlyer:
* Switch Fly view to ToolStrip
* ToolStrip short screen optimization
* Add Zoom in/out to Plan view ToolStrip


DonLakeFlyer:
This pull got completely out whack. I'll put it back together again once I see if this passes

PR #4568, opened 2017-02-18
Instrument widgets

dogmaphobic:
Move instrument widget off to a loader.
This allows a custom plugin the option to replace the entire widget with its own.
In the process of working with the instrument widget, I've removed a hole pile of unnecessary *properties*, which where passed around and around. Things like *active*, which would indicate if there was an active vehicle. And others such as *heading*, *roll*, *pitch*, etc., which are all properties of Vehicle, which is accessible globally.
While at it, removed some unused cruft from the toolbar controller. I'm looking into eventually removing the toolbar controller and move its functionality to Vehicle where it truly belongs. The toolbar was the very first QML code added to QGC, long before the Vehicle controller came into existence.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -21,6 +21,7 @@ include(QGCCommon.pri)
 
 TARGET   = QGroundControl
 TEMPLATE = app
+QGCROOT  = $$PWD
 
 DebugBuild {
     DESTDIR  = $${OUT_PWD}/debug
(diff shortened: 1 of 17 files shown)


DonLakeFlyer:
> In the process of working with the instrument widget, I've removed a hole pile of unnecessary properties, which where passed around and around. Things like active, which would indicate if there was an active vehicle. And others such as heading, roll, pitch, etc., which are all properties of Vehicle, which is accessible globally.

Awesome, I've been staring at those for a while thinking they were useless.

PR #4571, opened 2017-02-18
Add/use new SettingsManager implementation

DonLakeFlyer:
* Holds all the settings for the app
* Allows core plugin to override default values

Work left to do:
* At this point only the settings which were already SettingsFacts are converted. Need to move all setting to this new model.
* Add string support to FactSystem to support string settings
* Add visible meta data value to FactMetaData such that SettingsFacts can have visiblity meta data. Then allow core plugin to override the visible setting.
* Convert settings ui to use the new visibility meta data


DonLakeFlyer:
@dogmaphobic FYI...

Settings are now accessed from qml like this:
```
import QGroundControl.SettingsManager 1.0

...
    fact: QGroundControl.settingsManager.defaultMissionItemAltitude
...
```

Core plugin can override settings default values by implementing this override:
```
    /// Allows the core plugin to override the default value for the specified setting
    ///     @param name - Setting name
    ///     @param defaultValue - Standard default value for setting
    /// @return New default value for setting, if no override just return passed in defaultValue
    virtual QVariant overrideSettingsDefault(QString name, QVariant defaultValue);
```

Setting names are available from ```SettingsManager``` statics to reference in overrideSettingsDefault custom implementations.

Issue #4572, opened 2017-02-18
Mission Planning: last used waypoint altitude function should ignore special cases such as LAND


Antiheavy:
Special case waypoints such as LAND should ignore the last used waypoint altitude and these waypoints should also be ignored by other waypoints when determining last used altitude.

Currently if waypoints are placed numerically after a LAND or inserted earlier into the mission they default to the land altitude (usually set to 0 altitude, although at time of writing https://github.com/mavlink/qgroundcontrol/issues/4522 is still open) - this is probably bad as you wouldn't want to "fly" into the ground.  LAND waypoints should not follow the last used altitude and other waypoints should not follow Land waypoints' altitude.

This should be extended to ignore landing patterns and other complex waypoint (or mission item) types that have multiple elements at different altitudes.

Also check the behavior of: Takeoff, Loiter, VTOL takeoff and transition, VTOL transition and land.  These currently do not follow the last used altitude as they are hard coded to 50 m.  However, waypoints placed after these follow the altitude of these waypoints (with the exception of Takeoff waypoints) .  Takeoff is the only waypoint type that seems to ignore this logic both ways - it does not follow last used altitude and other waypoints do not follow its altitude.  I'm not sure if this is actually the desired behavior for Takeoff waypoint, but I think it is fine.  In fact, this is how the Land waypoint should work too.  For consistency, if a waypoint ignores the last used altitude then its altitude(s) should also be ignored.

(i'm using PX4 and Windows QGC v3.1.3)

DonLakeFlyer:
I'm also thinking the current logic to determine last used altitude is wrong with respect to the fact that it always pulls the last altitude in the list even if you are inserting and item in the middle of the mission. I think it should pull the altitude to use by only walking back from the insert point.

Antiheavy:
I guess walking back from the insert point would work for waypoints placed at the end of a mission list too.  As long as the walking back logic skips over special case waypoints such as Takeoff, Land, Patterns, and maybe others.

PR #4584, opened 2017-02-21
Plan: Change semantics for pulling altitude from mission items for new items

DonLakeFlyer:
New semantics:
* When a new item is inserted we look backwards from that item for a previous MAV_CMD_NAV_WAYPOINT and use the altitude and frame from that item as the defaults for the new item.
* The semantics above only apply to newly added MAV_CMD_NAV_WAYPOINT items
* Removed support for look back/set on new items for acceptance radius. Since acceptance radius is not normally shown on new items it's dangerous to override the default (which is 0) on something that isn't shown.

This differs from previous support which would find the last mission item which specified a coordinate and was not a land item. The issue with that method were:
* If you inserted a new item in the middle of a mission it would still use the value from the last mission item. This just didn't make any sense.
* I tried to exclude special case items to avoid pulling  previous altitude from them (MAV_CMD_NAV_TAKEOFF for example). This special casing mechanism wasn't feasible movinf forward as new items are added. The special casing will most likely fall out of date from new items being supported which would be a bad thing.

Fix for #4572

### DIFF
--- a/src/MissionEditor/MissionEditor.qml
+++ b/src/MissionEditor/MissionEditor.qml
@@ -851,7 +851,7 @@ QGCView {
                     ]
 
                     onClicked: {
-                        switch (index == 0) {
+                        switch (index) {
                         case 0:
                             _addWaypointOnClick = checked
                             break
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
This pull also includes a fix for adding items in the Plan view.

PR #4587, opened 2017-02-21
fixed uploading of bin files

sverling:
`_binFormat` was never set to true, as the function returned before that assignment, resulting in failing uploads of bin fails


DonLakeFlyer:
Thanks

PR #4590, opened 2017-02-21
Settings fixes

DonLakeFlyer:


DonLakeFlyer:
Cancelled OSX builds instead of waiting for travis to run them on their tin can OSX infrastructure. need to make forward progress so merging.

PR #4598, opened 2017-02-21
Upgrade to Qt5.8, MSVC 2015 on Windows 

NaterGator:


NaterGator:
@DonLakeFlyer as far as I now know this is good to go. Installer works on Windows, assuming you tested OS-X and Linux in your PR.

DonLakeFlyer:
Awesome, thanks Nate. I'll repackage and then send the whole thing through again.

dogmaphobic:
I have not tested QGC on Windows, let alone this new VS2015 build so I don't know if it will have the same issue. I had to port another project a couple of weeks ago to Windows (Qt using VS2015) and I noticed that copying the VS2015 runtime DLLs wasn't enough. It would work on Windows 10 but not on Windows 7. The only way I could get it running was using the "Redistributable" installer. I just added it to my nsis script.

DonLakeFlyer:
@NaterGator Nate?

NaterGator:
Gus, interesting. I've been using the VS2015 builds internally for a few weeks and haven't run into that problem with Windows 7. I believe we are OK but we'll find out from users soon no doubt.

dogmaphobic:
You don't have the issue because you have VS2015 installed (I assume you do). The issue happens on a "clean" system (non developer).

It took me a while to figure out. Windows doesn't tell you it's missing a DLL. It just crashes with a stupid dialog that doesn't explain anything.

This is what I added to the nsis script (there is probably a better way to do it):

```
Section "C++ Redistributables (required)"

    SectionIn RO

    SetOutPath $INSTDIR

    File "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\redist\1033\vcredist_x64.exe"
    ExecWait '"$INSTDIR\vcredist_x64.exe" /passive /norestart'

SectionEnd
```

NaterGator:
That's an option, and certainly we can go that route. These were cleanly imaged Windows 7 laptops though.

dogmaphobic:
Weird, I used the same set of files QGC uses for copying the DLLs. Who knows. Maybe I use something else there that needs something other DLL not included in the copy list used by QGC.

NaterGator:
Could be. I use process explorer from technet to view the DLLs loaded by the running program to verify nothing other than system DLLs are loaded outside the bundled app path.

This is fairly new though, so we'd be wise to keep a sharp eye for reports of weirdness.

PR #4600, opened 2017-02-22
RFC: Support for setups of video streaming for ArduPilot

AlexeyBulatov:
This PR needs changes that are proposed in this pull request mavlink/mavlink#664. Once they are merged, this will get compiled.
We might also add a new capability setting in Mavlink to dynamically create tabs like the ones shown below.
Add new screen of setups:
screenshot from 2017-02-03 18-44-43
And new summary screen:
screenshot from 2017-02-03 18-44-57
To set video streaming use the following mavlink commands:
DO_CONTROL_VIDEO and VIDEO_START_CAPTURE.
To stop:
VIDEO_STOP_CAPTURE.
Quality settings are updated only after start command.

I’m eager to hear back from you!


DonLakeFlyer:
Looks like a great start, thanks. 

Shouldn't this be in Settings/General with the rest of the video settings as opposed to requiring a vehicle connection to set up? Also Gus is adding camera buttons to Fly view. Seems like that should integrate with this.

@dogmaphobic Gus can you take a look?

dogmaphobic:
MAV_CMD_VIDEO_START_CAPTURE and MAV_CMD_VIDEO_STOP_CAPTURE are meant to control video recording. That is, to tell an on board camera (GoPro or whatever) to start/stop recording. There should be a separate set of messages to deal with video streaming.

All this should also be "common", not APM specific. We really should nail down these video/photo messages before it becomes a mess so huge nothing will work together.

Looking at the current MAVLink specs, I see:

Messages for camera settings:

MAV_CMD_SET_CAMERA_SETTINGS_1
MAV_CMD_SET_CAMERA_SETTINGS_2
MAV_CMD_REQUEST_CAMERA_SETTINGS
MAV_CMD_REQUEST_CAMERA_INFORMATION
MAV_CMD_REQUEST_STORAGE_INFORMATION

Taking stills:

MAV_CMD_IMAGE_START_CAPTURE
MAV_CMD_IMAGE_STOP_CAPTURE (in case the above started a time lapse)

Recording videos:

MAV_CMD_VIDEO_START_CAPTURE
MAV_CMD_VIDEO_STOP_CAPTURE
MAV_CMD_REQUEST_CAMERA_CAPTURE_STATUS

What I don't see is how to know what a camera is capable of (list of resolution and frame rates for video, encoding types, file types, list of supported ISO values, etc.) These things are not practical to be sent by MAVLink messages. We would need a way to get these camera capabilities some other way. Not to mention the need for the vehicle to announce the fact it has a camera and supports these messages.

And most importantly about this PR in particular, I don't see anything handling *Video Streaming*.

In my opinion, these messages should be better thought out and established before anything is added to QGC. Otherwise, we will get stuck with a broken system, using wrong messages for things they weren't meant to be used for.

This PR is a great start but it should not go in until these messages are sorted out and this moved to "common". This is not APM specific. It is (or it should be) part of the common set of messages.

dogmaphobic:
> Shouldn't this be in Settings/General with the rest of the video settings as opposed to requiring a vehicle connection to set up?

Normally yes, but if you want to set resolution and whatnot, you would need to know what the camera is capable of, which you don't until you connect to the vehicle and query it. Note that there are no messages for these queries yet neither there are messages to handle video streaming specifically, with the exception of one solitary (MAV_CMD_DO_CONTROL_VIDEO) message that is awfully inadequate.

dogmaphobic:
I added some comments here: https://github.com/mavlink/mavlink/pull/664

Once that's nailed down, we can proceed with this PR is a sane manner.

PR #4608, opened 2017-02-23
Move video settings to new SettingsGroup mechanism

DonLakeFlyer:
* GeneralSettings/Video section now supports visibility on each item


DonLakeFlyer:
Merging this in so I can make forward progress.

PR #4611, opened 2017-02-24
Add joystick indicator to show actual output

nanthony21:
This adds a second indicator dot to the roll, pitch, yaw, and throttle bars to show what is actually being output by the manualControl signal after the exponential and deadband is applied.

I have also changed the message that is displayed when you start joystick calibration so that the user knows that the deadzone is being measured at this time.

While working on this I noticed that the extremes of joystick movement don't actually change the output, I believe this may be a bug.

### DIFF
--- a/src/VehicleSetup/JoystickConfigController.cc
+++ b/src/VehicleSetup/JoystickConfigController.cc
@@ -81,7 +81,7 @@ JoystickConfigController::~JoystickConfigController()
 /// @brief Returns the state machine entry for the specified state.
 const JoystickConfigController::stateMachineEntry* JoystickConfigController::_getStateMachineEntry(int step)
 {
-    static const char* msgBegin =           "Allow all sticks to center as shown in diagram.\n\nClick Next to continue";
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
My concern here is whether this is just going to confuse regular people as to what it means?

@jaxxzer ?

nanthony21:
Maybe it would be better to just have one indicator per bar but have the bars on the left show what is actually being sent in the manual control? The bars on the right already show the raw input from the joystick and I think it would be very useful to have visual feedback on the effects of exponential and deadband.

jaxxzer:
I like the idea, and it appears to work correctly, but it is confusing. I think it would be better to disable the indicators by default, and put the option to enable them under advanced settings, or to display this information separately from the physical joystick positions.

nanthony21:
How about adding a button in advanced options to toggle between displaying input/output?

DonLakeFlyer:
> Maybe it would be better to just have one indicator per bar but have the bars on the left show what is actually being sent in the manual control?

That seems reasonable to me.

nanthony21:
Ok, I'll give that a try. Any thoughts on the issue mentioned above where the axes in the manualControl signal get to 100% when the joystick axis is actually ~75%?

nanthony21:
Please close and  see #4628

PR #4612, opened 2017-02-24
Added altitude to the Vehicle coordinate (QGeoCoordinate)

dogmaphobic:
QGeoCoordinate has 3 components: latitude, longitude and altitude. Up to now, only lat/lon were set, leaving the altitude fixed at 0. I'm now adding the altitude so I can compute an accurate distance between QGC and the vehicle.

In addition, instead of using functions to set lat/lon/alt, I'm setting them all at once and generating a single signal instead of multiple signals for the same variable as it was the case.

### DIFF
--- a/src/Vehicle/Vehicle.h
+++ b/src/Vehicle/Vehicle.h
@@ -645,8 +645,11 @@ public:
     QString vehicleImageCompass () const;
 
 public slots:
-    void setLatitude(double latitude);
-    void setLongitude(double longitude);
+    /// Sets the firmware plugin instance data associated with this Vehicle. This object will be parented to the Vehicle
+    /// and destroyed when the vehicle goes away.
+    void setLatitude    (double latitude);
+    void setLongitude   (double longitude);
+    void setAltitude    (double altitude);
 
 signals:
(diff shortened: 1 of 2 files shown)


dogmaphobic:
@DonLakeFlyer Can you please take a look at this and make sure it doesn't affect something unexpected? I don't think it does as you only use the vehicle's QGeoCoordinate to position it on a 2D map. I need the 3D coordinates to compute accurate distances.

src/Vehicle/Vehicle.h
```diff
@@ -645,8 +645,11 @@ class Vehicle : public FactGroup
     QString vehicleImageCompass () const;
 
 public slots:
-    void setLatitude(double latitude);
-    void setLongitude(double longitude);
+    /// Sets the firmware plugin instance data associated with this Vehicle. This object will be parented to the Vehicle
+    /// and destroyed when the vehicle goes away.
```
DonLakeFlyer: Is this comment in the right place? Don't understand it.

dogmaphobic: That comment was already there. I didn't understand it and I assumed you did understand it :P

PR #4613, opened 2017-02-24
Fix rtspsrc timeout storage type, add a few ms of jitterbuffer

NaterGator:
I have been getting crashes on windows with rtspsrc. Took a bit of hunting with the debugger but ultimately I realized the argument passed for timeout is a 32-bit value.  Since g_object_set uses varargs and the argument size was not properly matched, the null to indicate the last arg was read as part of the timeout and undefined behavior resulted when trying to interpret the remaining varargs. 

Additionally this PR includes 17msec of buffering on rtsp sources to give rtpjitterbuffer a chance to work; the impact should be imperceptible.

### DIFF
--- a/src/VideoStreaming/VideoReceiver.cc
+++ b/src/VideoStreaming/VideoReceiver.cc
@@ -205,7 +205,7 @@ void VideoReceiver::start()
             }
             g_object_set(G_OBJECT(dataSource), "uri", qPrintable(_uri), "caps", caps, NULL);
         } else {
-            g_object_set(G_OBJECT(dataSource), "location", qPrintable(_uri), "latency", 0, "udp-reconnect", 1, "timeout", static_cast<guint64>(5000000), NULL);
+            g_object_set(G_OBJECT(dataSource), "location", qPrintable(_uri), "latency", 17, "udp-reconnect", 1, "timeout", static_cast<guint64>(5000000), NULL);
         }
 
         if ((demux = gst_element_factory_make("rtph264depay", "rtp-h264-depacketizer")) == NULL) {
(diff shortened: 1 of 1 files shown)


src/VideoStreaming/VideoReceiver.cc
```diff
@@ -205,7 +205,7 @@ void VideoReceiver::start()
             }
             g_object_set(G_OBJECT(dataSource), "uri", qPrintable(_uri), "caps", caps, NULL);
         } else {
-            g_object_set(G_OBJECT(dataSource), "location", qPrintable(_uri), "latency", 0, "udp-reconnect", 1, "timeout", 5000000, NULL);
+            g_object_set(G_OBJECT(dataSource), "location", qPrintable(_uri), "latency", 17, "udp-reconnect", 1, "timeout", static_cast<guint64>(5000000), NULL);
```
dogmaphobic: Is 17ms enough? That's half a frame (assuming 30fps). Would 35ms be any more noticeable?

Good grief on the 32 -> 64 bit. I would assume the compiler would have better sense. Good hunt result!

NaterGator: 17ms helps, obviously a whole frame is going to be more accommodating of jittery networks. No, most people won't notice a single frame delay. Regardless on these sorts of links TCP-based streams are pretty much out the window and all we can feasibly solve here is packet jitter and reordering. 

I have no objections to increasing 33-35ms range, though we've apparently gotten away with no jitter buffer for a while which tells me either few people are using this feature or RTSP streaming setups are using pretty robust wireless links.


The gobject type issue _couldn't_ be caught by the compiler since it is a variadic function. I'm not aware of any MSVC probes or guards we could have turned on to catch this.

DonLakeFlyer:
@dogmaphobic I assume you'll merge when you guys are happy with this...

dogmaphobic:
Yeah, I will merge as is and experiment with it.

PR #4614, opened 2017-02-24
Fix shutdown crash

DonLakeFlyer:


src/main.cc
```diff
@@ -253,6 +253,7 @@ int main(int argc, char *argv[])
         exitCode = app->exec();
     }
 
+    app->_shutdown();
```
NaterGator: By this point the event loop is gone too. We should probably call [sendPostedEvents](http://doc.qt.io/Qt-5/qcoreapplication.html#sendPostedEvents) to destroy any stragglers after invoking _shutdown.

DonLakeFlyer: Hmm. I think I need to look at this some more to find a better place.

NaterGator: The only real risk that I can see is objects living on other threads may still be alive and running at this point, emitting signals and interacting with objects that lived on the main thread. It shouldn't be a huge deal since event processing is dead by this point (so signals would never reach their respective slots).

DonLakeFlyer: Nothing in toolbox or main window lives on another thread so that should be ok. I'm going to leave this as is. It still may not be exactly correct. But it's better than before and calling sendPostedEvents may lead to other strange things we've never seen on shutdown.

PR #4616, opened 2017-02-24
Disable scrolling on QGCComboBox

jaxxzer:
Currently, settings are easily changed by accident when scrolling a page, when the mouse passes over a combobox while scrolling. It is also easy to not notice this when it occurs. This patch will disable changing the options on comboboxes via scrolling, by handling the event (do nothing) with a mousearea over the combobox.

### DIFF
--- a/src/QmlControls/QGCComboBox.qml
+++ b/src/QmlControls/QGCComboBox.qml
@@ -39,4 +39,24 @@ ComboBox {
             }
         }
     }
+
+    // Capture Wheel events to disable scrolling options in ComboBox.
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Awesome. This has made me crazy forever!

jaxxzer:
Me too, I don't want to take all the credit:
http://stackoverflow.com/questions/33076960/qt-qml-combobox-override-wheel-events

DonLakeFlyer:
Yeah, I saw that stackoverflow when I just did some poking around to see if there was some better way to deal with this.

PR #4626, opened 2017-02-25
Allow setting the video streaming aspect ratio.

dogmaphobic:
Up to now, the video display for video streaming would stretch to fit the window. This new setting allows you to set the proper aspect ratio. If the window is wider than the video, you get black bars on the sides. Setting the aspect ratio to 0.0 makes it ignore it and it will stretch the video to fit.

Also changed the "NO VIDEO" message to a more optimistic "WAITING FOR VIDEO" :P

### DIFF
--- a/src/Settings/Video.SettingsGroup.json
+++ b/src/Settings/Video.SettingsGroup.json
@@ -27,5 +27,13 @@
     "longDescription":  "Directory to save videos to.",
     "type":             "string",
     "defaultValue":     ""
+},
+{
+    "name":             "VideoAspectRatio",
+    "shortDescription": "Video Aspect Ratio",
+    "longDescription":  "Video Aspect Ratio (width / height). Use 0.0 to ignore it.",
+    "type":             "float",
+    "decimalPlaces":    6,
+    "defaultValue":     1.777777
 }
 ]
--- a/src/ui/preferences/GeneralSettings.qml
+++ b/src/ui/preferences/GeneralSettings.qml
@@ -505,6 +505,20 @@ QGCView {
                         }
                         Row {
                             spacing:    ScreenTools.defaultFontPixelWidth
+                            visible:    QGroundControl.videoManager.isGStreamer && QGroundControl.videoManager.isGStreamer && videoSource.currentIndex < 2
+                            QGCLabel {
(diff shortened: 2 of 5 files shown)


src/Settings/Video.SettingsGroup.json
```diff
@@ -27,5 +27,13 @@
...
     "type":             "string",
     "defaultValue":     ""
+},
+{
+    "name":             "VideoAspectRatio",
+    "shortDescription": "Video Aspect Ratio",
+    "longDescription":  "Video Aspect Ratio (width / height). Use 0.0 to ignore it.",
+    "type":             "float",
+    "decimalPlaces":    6,
+    "defaultValue":     1.777777
```
DonLakeFlyer: Should this be an enum with a combo instead of having to type in direct ratios?

src/ui/preferences/GeneralSettings.qml
```diff
@@ -505,6 +505,20 @@ QGCView {
                         }
                         Row {
                             spacing:    ScreenTools.defaultFontPixelWidth
+                            visible:    QGroundControl.videoManager.isGStreamer && QGroundControl.videoManager.isGStreamer && videoSource.currentIndex < 2
```
DonLakeFlyer: Need to add && QGroundControl.settingsManager.videoSettings.aspectRatio.visible

DonLakeFlyer: That way overrides can hide.

NaterGator:
Any reason to not use [this](https://github.com/mavlink/qgroundcontrol/blob/418631ace7b6948e7d2aeb5cf9b84d76438f4d31/src/VideoStreaming/gstqtvideosink/gstqtvideosinkbase.cpp#L82-L95)?

dogmaphobic:
>Should this be an enum with a combo instead of having to type in direct ratios?

Aspect ratios can be rather arbitrary. In order to have an enum, with a pre-determined list, I would need to know what the camera supports, which we don't.

>Need to add && QGroundControl.settingsManager.videoSettings.aspectRatio.visible

Got it.

>Any reason to not use this?

Not really. The video is eventually rendered into a QML control (OpenGL). That can be stretched regardless of the source. I'm handling the control's width/height myself.

PR #4628, opened 2017-02-26
Changed joystick indicators to show output rather than input

nanthony21:
I also moved the deadband indicators to the right side column and added an explanation of how to calibrate the deadband. Also corrected a misspelled function.

### DIFF
--- a/src/AutoPilotPlugins/Common/RadioComponentController.cc
+++ b/src/AutoPilotPlugins/Common/RadioComponentController.cc
@@ -429,7 +429,7 @@ void RadioComponentController::_inputStickDetect(enum rcCalFunctions function, i
                 _rgChannelInfo[channel].rcMax = value;
             }
             
-            _signalAllAttiudeValueChanges();
+            _signalAllAttitudeValueChanges();
             
             _advanceState();
         }
@@ -582,7 +582,7 @@ void RadioComponentController::_resetInternalCalibrationValues(void)
         _rgFunctionChannelMapping[i] = _chanMax();
     }
     
-    _signalAllAttiudeValueChanges();
+    _signalAllAttitudeValueChanges();
 }
 
 /// @brief Sets internal calibration values from the stored parameters
@@ -654,7 +654,7 @@ void RadioComponentController::_setInternalCalibrationValuesFromParameters(void)
         }
     }
     
-    _signalAllAttiudeValueChanges();
+    _signalAllAttitudeValueChanges();
 }
 
 void RadioComponentController::spektrumBindMode(int mode)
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
Previous discussion in #4611

DonLakeFlyer:
@jaxxzer What do you think?

jaxxzer:
I'll take a look when I get home later tonight. @nanthony21 what is the motivation for moving the dead and indicators?

nanthony21:
@jaxxzer since the left hand indicators are now showing the output (deadband has already been applied) it would be misleading to show the deadband on the same bar.

jaxxzer:
This is ok with me in general, but I would like to see two things:
1. Only display the additional deadband help text during calibration if deadbands are enabled.
2. The axis indicators should still move when joystick input is disabled.

DonLakeFlyer:
@nanthony21 Can you change for #1? Then I think we are good.
@jaxxzer For #2 That is not something that has changed with this pull is it?

nanthony21:
I'll work on 1. 

I think 2 has been around for a while. It would be nice to fix though. I'll take a look.

jaxxzer:
It looks like #2 was not a result of this pr, just something I noticed while testing this.

nanthony21:
This has turned out to be more complicated than I expected. I wanted to use an "if" statement to change msgBegin depending on if deadband was enabled. However, the fact that rgStateMachine is "static const" prevented the message from getting updated once it had already been defined. Trying to make rgStateMachine so it isn't "static const" caused crashes which I think are related to the helper images not being able to load.

https://github.com/mavlink/qgroundcontrol/blob/21923ab9afcf55a9eff0adcdd894ce0617588067/src/VehicleSetup/JoystickConfigController.cc#L84

I'm a bit of a novice with C++ so let me know if I'm missing something obvious.

Another option would be to put the deadband-specific instructions next to the deadband checkbox. Or we could just remove it altogether, people will probably figure it out pretty easily.

DonLakeFlyer:
Modifying that structure isn't possible unless the whole thing is changed from static const and requires other changes as well.

Do you have a screenshot of what this looks like now?

nanthony21:
I just reversed the commit so the instruction messages will be unchanged in this pull.

DonLakeFlyer:
So we good to go with this now?

jaxxzer:
I'll take another look now.

jaxxzer:
LGTM

DonLakeFlyer:
Ok. I'll merge once travis catches up to the other pulls.

DonLakeFlyer:
Trying to kick in Travis run...

PR #4648, opened 2017-02-27
Switch to Qt 5.8

DonLakeFlyer:


DonLakeFlyer:
I'm going to pull the trigger on this which will likely lead to some additional problems but we'll fix them as we find them.

PR #4661, opened 2017-02-28
Plan: MissionItemEditor - Workaround strange lost clicks possible Qt bug

DonLakeFlyer:
Only shows up on Android. Busted in 5.8, works in 5.5.


DonLakeFlyer:
It fixes the instances I could find:
* Left ToolStrip button after you cancel a DropPanel
* Plan view: Mission item list

There could be more. If you find yourself having to click things twice (all the time) it is likely the same problem which needs fixing.

PR #4662, opened 2017-02-28
Increase max flightmap zoom level to 22

jaxxzer:
This approximately doubles the scale of the map at maximum zoom. This helps facilitate flights/missions with particularly small area of operation.

### DIFF
--- a/src/QtLocationPlugin/QGCMapUrlEngine.h
+++ b/src/QtLocationPlugin/QGCMapUrlEngine.h
@@ -23,7 +23,7 @@
 #include <QNetworkReply>
 #include <QMutex>
 
-#define MAX_MAP_ZOOM (20.0)
+#define MAX_MAP_ZOOM (22.0)
 
 class UrlFactory : public QObject {
     Q_OBJECT


dogmaphobic:
Did you try this with all map providers? Chances are you will get 404 on most requests at that level.

DonLakeFlyer:
Which provider goes up to 22?

jaxxzer:
It worked for google maps, but one scroll wheel increment less than 22.

jaxxzer:
I was going to raise the tile thing as a separate issue after I looked into it more. Is it possible to circumvent the tile loading errors at high zoom levels, by just zooming in on what we have, like a picture saved to disk?

DonLakeFlyer:
> Is it possible to circumvent the tile loading errors at high zoom levels, by just zooming in on what we have, like a picture saved to disk?

There is already an issue in for that. It's not an easy change.

DonLakeFlyer:
@dogmaphobic So we can't do this because it will general missing tiles depending on your map provider right?

jaxxzer:
> @dogmaphobic So we can't do this because it will general missing tiles depending on your map provider right?

Correct, also this broke somewhere after the switch from QT 5.5

PR #4675, opened 2017-03-01
Add firmware information to parameter file header

jaxxzer:
This adds some information about the autopilot firmware that would be useful in troubleshooting issues on user forums, among other things.

Here are some examples of what the parameter file header looks like:
APM:
```
# Onboard parameters for Vehicle 1
#
# Stack: ArduPilot
# Vehicle: Fixed Wing
# Version: 3.7.1 
# Git Revision: 22b5c415
#
# Vehicle-Id Component-Id Name Value Type
```

PX4:
```
# Onboard parameters for Vehicle 1
#
# Stack: PX4 Pro
# Vehicle: Multi-Rotor
# Version: 1.6.0 dev
# Git Revision: e892a51afb597632
#
# Vehicle-Id Component-Id Name Value Type
```


DonLakeFlyer:
Handy, thanks.

PR #4678, opened 2017-03-01
Change to use new dynamic map visual system

DonLakeFlyer:


DonLakeFlyer:
This also lays the groundwork for better drag support of mission items on touch systems

PR #4680, opened 2017-03-02
Add joystick indicator icon for Sub

jaxxzer:
This is disabled for other Vehicles. Joystick Indicator is placed with vehicle indicators, as it is required for Sub operation. This also corrects an issue with the RCRSSIIndicator making a blank gap in the indicators. This issue was introduced after the conversion to plugins.

js


src/ui/toolbar/JoystickIndicator.qml
```diff
@@ -0,0 +1,102 @@
...
+import QGroundControl.MultiVehicleManager   1.0
+import QGroundControl.ScreenTools           1.0
+import QGroundControl.Palette               1.0
+
+// Joystick Indicator
+Item {
+    width:          joystickRow.width * 1.1
+    anchors.top:    parent.top
+    anchors.bottom: parent.bottom
+    visible:        activeVehicle ? activeVehicle.sub : false
```
DonLakeFlyer: Instead of doing a visible check here is it possible to put the check in core plugin and just not add the page. This would save to perf/memory so it's not always there.

jaxxzer: I will try this out. This would also resolve the RCRSSIIndicator issue in a better way.

jaxxzer: @dogmaphobic Do you have any advice on how to proceed here?

DonLakeFlyer: Oh, wait. That's not possible. You don't have a vehicle at that point. Sorry about that.

jaxxzer: Could we `connect` `MultiVehicleManager::activeVehicleChanged` ?

DonLakeFlyer: Wait I'm wrong again! This method could be modified to pass in the Vehicle just like the FirmwarePlugin methods work. @dogmaphobic ?

DonLakeFlyer:
Gus has new mechanism ready to go

DonLakeFlyer:
Closing this since you should use the new mechanism

PR #4681, opened 2017-03-02
Fix mode indicator vertical alignment.

dogmaphobic:


DonLakeFlyer:
Don't see how these changes would cause the UT problem.

PR #4695, opened 2017-03-03
Improve deadband

nanthony21:
This pull does two things.

1: it removes a section of code that made it so that the output of an axis (manual control signal) reached 100% when the input from that axis was only at ~75%. This means that the last 25% of axis movement was useless. I'm not sure what the intended purpose of this code was.

2: Previously, if an axis had deadband applied then the maximum output on that axis would be 100%-percentDeadband. To give an example, if an axis had 40% deadband then moving the stick all the way over would only result in 60% at the output. This pull causes the sensitivity of each axis to scale with deadband so that 100% input always results in 100% output.

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -431,18 +431,6 @@ void Joystick::run(void)
                 throttle_accu = std::max(static_cast<float>(-1.f), std::min(throttle_accu, static_cast<float>(1.f)));
                 throttle = throttle_accu;
             }
-
-            float roll_limited = std::max(static_cast<float>(-M_PI_4), std::min(roll, static_cast<float>(M_PI_4)));
-            float pitch_limited = std::max(static_cast<float>(-M_PI_4), std::min(pitch, static_cast<float>(M_PI_4)));
-            float yaw_limited = std::max(static_cast<float>(-M_PI_4), std::min(yaw, static_cast<float>(M_PI_4)));
-            float throttle_limited = std::max(static_cast<float>(-M_PI_4), std::min(throttle, static_cast<float>(M_PI_4)));
-
-            // Map from unit circle to linear range and limit
-            roll =      std::max(-1.0f, std::min(tanf(asinf(roll_limited)), 1.0f));
-            pitch =     std::max(-1.0f, std::min(tanf(asinf(pitch_limited)), 1.0f));
-            yaw =       std::max(-1.0f, std::min(tanf(asinf(yaw_limited)), 1.0f));
-            throttle =  std::max(-1.0f, std::min(tanf(asinf(throttle_limited)), 1.0f));
-            
             if ( _exponential ) {
                 // Exponential (0% to -50% range like most RC radios)
                 // 0 for no exponential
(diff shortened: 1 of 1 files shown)


jaxxzer:
I believe that point 1 was to convert the circular limits of a gamepad joystick to the square limits of a conventional RC Transmitter. I'm not sure, but I think the joysticks handle this internally, and it would be ok to remove. I'll check to make sure though.

DonLakeFlyer:
I believe @LorenzMeier had put in #1.  I don't quite understand what it was for either.

DonLakeFlyer:
> I'll check to make sure though.

@jaxxzer Any update? If removing this code is wrong. It means for example you won't be able to arm with a joystick since it won't go all the way to the right.

jaxxzer:
Haven't had the chance yet, but will keep that point in mind when I am able to investigate.

jaxxzer:
@nanthony21 Please remove change 1. This is necessary to simultaneously reach the endpoint of the two axes on one joystick.

Change 2 looks good to me, but please reformat the if-else block with braces for clarity. Nice catch.

nanthony21:
I’ve fixed the formatting. Before I reverse #1 I would like to make a case for why I think it should stay in. In the following images the black outline represents the physical motion of the stick and the green area represents how the stick is mapped to an axis in the software.

With the joystick code as it currently is in master, a circularly cropped stick will look like this:
 
case3

The upside to this is that the stick can reach every part of the axis mapping. The drawback is that much of the stick’s movement range doesn’t actually do anything. Having this shrunken axis mapping also means that precision control will be more difficult because the vehicle will be more sensitive to movement of the stick.

However, many sticks that a user may use are not circularly cropped. We have no idea what controller layout is being used so we shouldn’t assume that it is a circularly cropped stick. In the case of many controllers the mapping will look like this:
 
case4

This has all the downsides I mentioned earlier without any benefit.

If we keep #1 in then the mappings will look like this:

case1
case2  

In this case a stick which can move through the whole range will be mapped exactly as you would expect. For a circularly cropped stick it is true that you won’t be able to reach the corners of the axis mapping but then entire range of stick movement  will still be useful. If I’m not mistaken this is the whole reason that sticks are sometimes made with circular cropping anyway, they ensure that you can’t achieve a larger magnitude vector in diagonal directions than you would in a pure X or Y direction.

Let me know what you think.

jaxxzer:
@DonLakeFlyer I would actually prefer @nanthony21's method for operating Sub. I don't think this would work on other vehicles for things like you mentioned, arming/disarming etc, we don't do these things in Sub. What do you think about making it a checkbox option under Advanced Settings?

nanthony21:
I don't think it should affect arming, right?

DonLakeFlyer:
I'm not super interested in optioning this thing to death.

jaxxzer:
Then I think we should take it out before merging this PR.

nanthony21:
I've reversed #1

DonLakeFlyer:
@jaxxzer Good to go now right?

jaxxzer:
Yes.

Issue #4635, opened 2017-02-26
FixedWing Landing Pattern: Remaining work


DonLakeFlyer:
Remaining work:
* Persistence
* Mission stats calculations

Finished:
* Click to set current item
* More work on visuals - much better now, feedback still welcome

Questions:
* Exit cross-track MAV_CMD_NAV_LOITER_TO_ALT command is set to 0. Does this mean the vehicle will exit the loiter on a direct line from loiter to land waypoints center?
* Need feedback on default values

Current state:
* Ready for SITL testing to gather feedback before I finish out the details. (Once #4634 goes through)

DonLakeFlyer:
@antiheavy FYI...

NaterGator:
Exiting the loiter on a tangent point is virtually all I've seen in use because it provides for the most predictable transition to landing approaches. Also computing the glide slope from the breakout point to landing coordinate is helpful to determine if you are within the aerodynamic margins of your craft. 

I've succeeded in getting NaNs on the landing point distance and heading; I think it is happening when distance == heading, but it is a bit hard to say. Around those same singular points I've also seen the "L" abbreviation disappear on the two position indicators.

Antiheavy:
copying my comments from https://github.com/mavlink/qgroundcontrol/pull/4599 to here:

I get the following error messages (yellow pop-up bar) when uploading missions with Landing Patterns:

- "Error during Mission communication with Vehicle: Vehicle returned error: Command is not supported (MAV_MISSION_UNSUPPORTED). Vehicle only has a partial list of mission items."
- "IGN MISSION_ITEM: Busy"

The Loiter-to-Alt and Land waypoints get taken by the autopilot as normal waypoints at default altitude. This is confirmed by downloading from the autopilot after attempted upload.

Otherwise looking great! Here is some additional feedback:

Default "Loiter altitude" of 40 m is good. Above most trees and local power lines in the world. Obviously pilot must adjust on their own for hilly terrain, giant trees, large structures, etc.

Default "Landing distance" should be 300 m. 40 m alt and 300 m distance provides a 7-8 degree glide slope which is reasonable for most small fixed wings. A future ability for users to set their own defaults for things like this will be nice.

"Landing heading" should be more like how runways work in aviation. The heading angle should be the direction of flight of the aircraft, not the direction the aircraft is landing from (i.e. 180 degrees opposite from how it is in this PR). (I feel there might be other places in QGC where this is "wrong", wind direction comes to mind)

[...]

PR #4703, opened 2017-03-03
Fixed Wing Landing Pattern

DonLakeFlyer:
* Correct cross track value loiter to be tangent exit for item sent to vehicle
* Visuals shows tangent flight path
* Fix mission item editor command json for xtrack enum

screen shot 2017-03-03 at 2 45 14 pm

### DIFF
--- a/src/MissionEditor/FWLandingPatternMapVisual.qml
+++ b/src/MissionEditor/FWLandingPatternMapVisual.qml
@@ -26,6 +26,8 @@ Item {
     property var _itemVisuals: [ ]
     property var _mouseArea
     property var _dragAreas: [ ]
+    property var _loiterTangentCoordinate
+    property var _flightPath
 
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Related to #4635

DonLakeFlyer:
Please, no more geometry for a while!

NaterGator:
For what it is worth you can also just project the coordinate by radius meters from the origin of the loiter at a bearing of bearing(loiter,land) (+/-) 90° (where the sign is determined by loiter being anti/clockwise. 

Anyways, you did it the hard way and that is fine too ;)

DonLakeFlyer:
I don't think that works in all case, like this one? The tangent isn't always at 90 degrees. Unless I'm seeing something wrong.
screen shot 2017-03-03 at 3 49 48 pm

PR #4711, opened 2017-03-05
Uninitialized variable warning.

dogmaphobic:
Fixing warning.


DonLakeFlyer:
Hmm. WarningsAsError is still turned off?

PR #4715, opened 2017-03-06
Camera list for survey comes from FirmwarePlugin

DonLakeFlyer:


DonLakeFlyer:
```
    /// Returns a list of CameraMetaData objects for available cameras on the vehicle.
    virtual const QVariantList& cameraList(const Vehicle* vehicle);
```

PR #4717, opened 2017-03-06
dynamic param load timeout

bkueng:
This dynamically sets the param load retry timeout based on the measured param receiving rate. The previous fixed timeout did not account for different link speeds, resulting in unnecessary high waiting times.

I tested with different HW (SITL, Pixracer, Pixhawk), different connections (UDP, USB, Pixracer WiFi), and various param drop scenarios (none, single, burst of 50, every second param).
Even with 50% drops, I still got an acceptable param load time.

Results: the timeout is large enough so that it never triggered too early. For a usb connection I got 60ms, for a pixracer via WiFi 200ms timer timeout.

@DonLakeFlyer The default timeout of 30s is incredibly high. I see that it has been recently changed due to some problems with APM (https://github.com/mavlink/qgroundcontrol/pull/4687). Maybe it needs different handling for APM vs PX4?

TODO:
- test using a telemetry link
- test with APM

### DIFF
--- a/src/FactSystem/ParameterManager.cc
+++ b/src/FactSystem/ParameterManager.cc
@@ -63,6 +63,8 @@ ParameterManager::ParameterManager(Vehicle* vehicle)
 {
     _versionParam = vehicle->firmwarePlugin()->getVersionParam();
 
+    _intervalUpdateTimer.invalidate();
+
     if (_vehicle->isOfflineEditingVehicle()) {
         _loadOfflineEditingParams();
         return;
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Please no don't ever merge anything into parameter manager without letting me look at it! 

This code is extremely fragile to changes and needs all sorts of testing across multiple firmwares before it can go through. If you aren't intimately familiar with it you should not be in here.

The 30 second timeout was from testing I was doing and is just a bad push. It just needs to be changed back and this will all go away.

DonLakeFlyer:
I also don't see why all of this complexity is needed. The original timeout was 5 seconds which should be long enough to handle anything. The real change is to speed up retries on initial request misses since at that point we don't have to worry about choking out the link since we are waiting for all params to load before we proceed.

bkueng:
@DonLakeFlyer The original timeout was 3 seconds, and it was causing problems as well. I did not create this PR because of the 30s timeout.
This is the problems with 3s: it could happen that only roughly 50% of the params were loaded (because of drops), then the rest was loaded with 3s timeouts, 10 params at a time. This was painfully slow, on links that could be much faster. It happened for example on the Intel Aero.

I know things are complex & fragile here, but please give it a proper review and/or suggest alternatives.

DonLakeFlyer:
> The real change is to speed up retries on initial request misses since at that point we don't have to worry about choking out the link since we are waiting for all params to load before we proceed.

What I said.

DonLakeFlyer:
Also 10 param batching is fine.

bkueng:
> The real change is to speed up retries on initial request misses since at that point we don't have to worry about choking out the link since we are waiting for all params to load before we proceed.

Exactly. And this PR does that. If we want to restore the timeout after initial param load, we can do that. I also agree that 10 param batch is fine.

DonLakeFlyer:
For initial loading there is no need to wait for a timeout to ask for missing indices. You can just ask for more as soon as you get one down. 

Also if reponses start to go beyond 3 seconds * 5 retries there is simply something wrong with the connection. I don't see why you would adjust timeouts to be that long. You are just hiding a problem either with your communication set up or bad firmware. This will then in turn bite you further down the road when the mission protocol kicks in and then the geo fence protocol and then the rally point protocol. All of these protocols will also crash and burn if the link is so flaky timeouts measured in seconds are going off. They also all have retry counts for no response. And they will fail in the same wa.

By increasing timeouts to unreasonable values you are not doing to user any favors.

bkueng:
> For initial loading there is no need to wait for a timeout to ask for missing indices. You can just ask for more as soon as you get one down.

Then let's do that. I thought if a whole bunch of ~50 params is dropped, re-requesting them immediately would overload the link, which is why I adjust the timeouts with this PR (and I only decrease them, that's the whole point of this PR).

DonLakeFlyer:
You use the same batch recount logic and re-request code to never have more than batchCount on the wire. So the link never gets hammered.

PR #4719, opened 2017-03-06
ParameterManager: make sure there is progress even on connection loss

bkueng:
What could happen before on connection loss:
- a full batch of params reached the retry limit
- goto Out; was executed, but actually no new params got requested
- retry timer did not get restarted

The result was no progress anymore.

There's still the problem that when we lose connection during param load, some params will time out, reach the maximum retry count and thus QGC will show the error that it could not retrieve all params.

What we could do is:
- stop the retry timer on connection loss
- on connection regain:
  - if (!_initialLoadComplete) { add all failedReadParamIndexMap to waitingReadParamIndexMap}
  - start the retry timer

This will not work when a connection stops, all retries fail (and qgc will already show the error), and only then a connection loss is triggered. Increasing the maximum number of retries would help with that.

@DonLakeFlyer @dogmaphobic does that sound reasonable? I don't want to make things more complicated than necessary.

### DIFF
--- a/src/FactSystem/ParameterManager.cc
+++ b/src/FactSystem/ParameterManager.cc
@@ -543,10 +543,6 @@ void ParameterManager::_waitingParamTimeout(void)
         }
 
         foreach(int paramIndex, _waitingReadParamIndexMap[componentId].keys()) {            
-            if (++batchCount > _maxBatchSize) {
-                goto Out;
-            }
-
             _waitingReadParamIndexMap[componentId][paramIndex]++;   // Bump retry count
             if (_disableAllRetries || _waitingReadParamIndexMap[componentId][paramIndex] > _maxInitialLoadRetrySingleParam) {
                 // Give up on this index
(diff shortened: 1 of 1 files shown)


dogmaphobic:
Why do we get this "Connection Lost" only while parameters are being downloaded?

bkueng:
I don't know yet. I checked on the FW side, and the link is not overloaded.

DonLakeFlyer:
I'm not quite sure what this is supposed to fix. But whatever it is this is not correct as far as I can tell. 

I don't quite understand what the problem is that is being fixed? If the communication with the vehicle is lost for so long that the parameter retries end up timing out. Then there isn't much to do other than figure out why the vehicle is not communicating.

bkueng:
Let me try to explain: it can come into a state where param loading just stalls indefinitely.
Assume param loading stops all at once. Then retry logic kicks in and requests a batch of 10 params. Lets say all of these time out as well. The retry is repeated `_maxInitialLoadRetrySingleParam` times, for the same batch. Then things go wrong in the loop:
https://github.com/mavlink/qgroundcontrol/pull/4719/files#diff-2d238aec2be2b9c6088dd17733f68679L545
is iterated, all of the iterations land in `Give up on this index`, and because we increased `batchCount`, we do the `goto Out;` on the last iteration. Also `paramsRequested` is still false at that point, so retry timer is **not** enabled and we did not send any new request, but `_waitingWriteParamNameMap` is actually not empty, thus `_checkInitialLoadComplete();` cannot complete either.
We are then in a 'locked up' state, because `batchCount` increased but in fact we did not call `_readParameterRaw`.

DonLakeFlyer:
Ok, thanks. I'll take a look at that case. I'm in there now and I'll merge this into my changes as needed. Leaving open for now until my stuff goes through.

DonLakeFlyer:
No longer needed with new code.

PR #4726, opened 2017-03-07
Add battery capacity field

nanthony21:
From what I understand from here: https://pixhawk.org/users/battery_config
the battery capacity is important in order for the battery life estimation to function well. Since it is important it should appear in the battery configuration page and not just in the parameters page. This pull request adds a field to the "Power" page that allows the user to change the BAT_CAPACITY parameter

### DIFF
--- a/src/AutoPilotPlugins/PX4/PowerComponent.qml
+++ b/src/AutoPilotPlugins/PX4/PowerComponent.qml
@@ -289,7 +289,7 @@ SetupPage {
                         QGCLabel {
                             text:   (battNumCells.value * battLowVolt.value).toFixed(1) + ' V'
                         }
-/*
+
                         QGCLabel {
                             text: qsTr("Battery Capacity:")
                         }
@@ -300,7 +300,9 @@ SetupPage {
                             fact:       battCap
                             showUnits:  true
                         }
-*/
+
+                        Item { width: 1; height: 1; Layout.columnSpan: 3 }
+
                         QGCLabel {
                             text:               qsTr("Voltage divider")
                         }
--- a/src/AutoPilotPlugins/PX4/PowerComponent.qml
+++ b/src/AutoPilotPlugins/PX4/PowerComponent.qml
@@ -38,6 +38,7 @@ SetupPage {
             property Fact battVoltageDivider:   controller.getParameterFact(-1, "BAT_V_DIV")
             property Fact battAmpsPerVolt:      controller.getParameterFact(-1, "BAT_A_PER_V")
             property Fact uavcanEnable:         controller.getParameterFact(-1, "UAVCAN_ENABLE", false)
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I could have sworn we used to have that in there and then it was taken out for some reason.

@dogmaphobic Do you remember anything about that?

dogmaphobic:
> I could have sworn we used to have that in there and then it was taken out for some reason.

Yes, I had it there when I first put this together but @LorenzMeier told me to take it out: https://github.com/mavlink/qgroundcontrol/pull/1372

nanthony21:
To summarize the previous discussion in #1372: It was decided that the battery capacity parameter wasn't very useful anyway and should be left in the background.

Until recently I didn't know that this parameter existed and I found that the battery estimation functioned terribly due to the voltage drooping during use. Ever since I started using the battery capacity parameter it seems to be working much better. Perhaps something has changed in the firmware since the previous discussion?

DonLakeFlyer:
I'd ask the firmware folks. Or maybe move this issue over to the firmware side so you can get some firmware eyeballs on it.

DonLakeFlyer:
@nanthony21 You are going to get the go ahead from firmware side before we can merge.

dagar:
I don't know the history of the discussion, but when that parameter is supplied it's used for the battery percentage calculation, which generally works better than voltage, but only if the usable capacity is known and you start with the fully charged (and have a calibrated current sensor).

Might be best to leave this one hidden behind an advanced box or in the param tree?

DonLakeFlyer:
That sort of rings a bell. Seems like I remember something about maybe Lorenz being concerned about bad values in here making things kick of bad values. Hmm, not sure what to do. Let me try to ping Lorenz tomorrow on it.

nanthony21:
I checked out the firmware code that handles battery estimation.  It appears that  every line of the code is newer than that previous discussion so maybe some of the problems he was worried about have been fixed. 

One problem I see is that the code doesn't account for the use of a partially charged battery, I'll add this feature when I get a chance.

However, between the current based estimation and the voltage based estimation, the code errs on the safe side by choosing the estimation that indicates the lowest remaining energy. This means that there is no risk of a bad BAT_CAPACITY parameter causing the battery to be overly discharged. 

If you think it'd be better I can move everything into the advanced settings section. Otherwise it's probably best to just close the PR and keep the capacity setting in the parameters page.

DonLakeFlyer:
Let's keep the pull open for now till I can talk with Lorenz.

DonLakeFlyer:
> but only if the usable capacity is known and you start with the fully charged

Talked with Lorenz: Daniel nailed it. It's too dangerous to always assume battery is fully charged. If it's not, the calculations are wrong. So a total power user feature which means it should stay just in parameter editor like all the other high end stuff.

PR #4731, opened 2017-03-08
Clean up the FW landing pattern math, kill warning

NaterGator:


DonLakeFlyer:
I'll have to look at this to see if I can understand this new way. The reason is that I understand how the current code works and does it's calculation. If I don't understand it then I can't maintain it.

NaterGator:
No prob, check out the linked wiki [image](https://en.wikipedia.org/wiki/Thales%27_theorem#/media/File:Thales%27_Theorem_Tangents.svg). I'm doing the same thing you were doing, except I'm using the triangle postulate to directly calculate ∠TOP. I then apply the original loiter->land bearing to rotate ∠TOP correctly.

NaterGator:
Also, forgot to mention the reason I went after this. I'm _fairly_ confident your method assumes an equirectangular map projection (mapping geo to screen coordinates before computing slope) which is probably a safe bet given the current map plugins, but not sure it is safe to assume in the future.

DonLakeFlyer:
I'm going to need to move these calcuations into the controller side. I think that is still doable but not quite sure yet since they are based on qml Map methods. The reason is that we want to have the heading calculated from the tangent line. It is currently calculated from waypoint to waypoint.

DonLakeFlyer:
Ok, I got it. Much simpler. Thanks Nate. Now I need to move this to the C++ side so the values are not just visual.

NaterGator:
@DonLakeFlyer  no problem. These are all QGeoCoordinate methods so they should translate to C++ effortlessly.

DonLakeFlyer:
Replaced by #4770

PR #4732, opened 2017-03-09
Plan:Survey: All survey values are now SettingsFacts

DonLakeFlyer:
This way the various values you set in a survey will persist from run to run.


DonLakeFlyer:
Also allows all survey settings defaults to be override by core plugin.

PR #4735, opened 2017-03-09
Giving up on OS X in Travis

DonLakeFlyer:


DonLakeFlyer:
Can do this since it will stop putting up OSX daily builds.

Issue #4534, opened 2017-02-11
Allow override of enumerated parameter values


jaxxzer:
Can we add an additional field to manually enter parameter values for those parameters that only present a dropdown list of choices? For example, ATC_ACCEL_Y_MAX in Copter. If there is concern about making this feature available to general users, maybe we could have an option for it in the Application Settings?

DonLakeFlyer:
Make sense

jaxxzer:
Done!

PR #4740, opened 2017-03-10
Add manual entry option for enumerated and bitmask parameters

jaxxzer:
Addresses https://github.com/mavlink/qgroundcontrol/issues/4534

### DIFF
--- a/src/QmlControls/ParameterEditorDialog.qml
+++ b/src/QmlControls/ParameterEditorDialog.qml
@@ -83,17 +83,6 @@ QGCViewDialog {
             anchors.left:   parent.left
             anchors.right:  parent.right
 
-            // Checkbox to allow manual entry of enumerated or bitmask parameters
-            QGCCheckBox {
-                id:         manualEntry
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Can you change this as follows:

If you run with PX4 stack you will see an Advanced section to the editor dialog:
screen shot 2017-03-11 at 9 40 55 am

 If you check the box and additional section opens up with an additional PX4 only option:
screen shot 2017-03-11 at 9 41 03 am

Can you make the advanced section be available to all firmwares. Then make it show when it is needed (default to closed). And then move your thing down there. Remove the (advanced user) text.

jaxxzer:
Ok, can do. I am out of town for the next ten days so updates from me may be infrequent until I return

jaxxzer:
screenshot 2017-03-22 16 32 10



screenshot 2017-03-22 16 31 13

DonLakeFlyer:
Sorry @jaxxzer. Didn't realize you had fixed this up. Thanks.

PR #4746, opened 2017-03-11
Guided bar confirm: Better usability with touch

DonLakeFlyer:


DonLakeFlyer:
This increases the height to min touch sizing. It also allows you to start the drag further to the left of the control in case you miss with your finger.

PR #4748, opened 2017-03-11
Move VehicleComponent::prerequisiteSetup method to AutopilotPlugin

DonLakeFlyer:
This makes VehicleComponents reusable from core plugin without requiring override of prerequisiteSetup.

The new rule for core plugins is that if they are re-using setup pages they should store them in the protected variables in the base plugin. Then the AutopilotPlugin:: prerequisiteSetup call will works even is some pages are not loaded. The core plugin AutopilotPlugin implementation can still override the prerequisiteSetup method if needed as well.


DonLakeFlyer:
@dogmaphobic This fixes the problem with the VehicleComponent reaching around to other components.

PR #4766, opened 2017-03-14
Exponential Slider

nanthony21:
This pull request changes the Joystick._exponential property from bool to float. It also replaces the exponential checkbox in the joystickConfig.qml with a slider allowing the user to set the exponential as they please.

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -445,9 +445,7 @@ void Joystick::run(void)
             
             if ( _exponential != 0 ) {
                 // Exponential (0% to -50% range like most RC radios)
-                // 0 for no exponential
-                // -0.5 for strong exponential
-                //float expo = -0.35f;
+                //_exponential is set by a slider in joystickConfig.qml
 
                 // Calculate new RPY with exponential applied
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
I'm leaving @jaxxzer in charge of all the joystick reviews. I think he's been out for a week. Probably back in a couple days.

jaxxzer:
There appears to be an issue with the qml here. @nanthony21 can you verify/fix this?
screenshot 2017-03-22 14 29 21

nanthony21:
This appears to be a problem with QGCSlider.qml itself. Lines 24 and 25 raise: "ReferenceError: TextSingleton is not defined"

 https://github.com/mavlink/qgroundcontrol/blob/1723a2d2611014c3d1f56204d062defafc359c21/src/QmlControls/QGCSlider.qml#L24

It appears to be caused by this commit: 9944728e0b0a92063c9824327c1b83781bba16e6

If you add "import QtQuick.Controls.Private 1.0" back into QGCSlider.qml it works fine

DonLakeFlyer:
That's a screwup from the revert of moving all imports to 5.8 latest. You can add it back.

nanthony21:
Should be good now.

PR #4779, opened 2017-03-15
Allowing to skip going to the Summary page.

dogmaphobic:
If another panel is controlling vehicle connection/disconnection, we need a way to skip the panel from switching away from underneath you.


dogmaphobic:
> That autoswitch to summary is kind of annoying in general. I wonder if we should just pull it out competely?

Not completely. What we need is a way to tell QGC where to go to programmatically depending on the circumstances. Most of the time you just stay put. But if the vehicle has a configuration issue for instance, it makes sense to go to the Summary page.

DonLakeFlyer:
Ok. Seems fine then

DonLakeFlyer:
FYI: I killed the OSX installer build since they are all failing. You can merge without that.

Issue #4765, opened 2017-03-14
Survey - camera orientation flipped for overlap calculation?


Antiheavy:
I think the Landscape vs Portrait camera options might be reversed when calculating the overlap for a Survey.  At least is looks reversed for the side-lap calculation at least, I can't tell about the frontal-lap.

![image](https://cloud.githubusercontent.com/assets/12117027/23883934/dd473cee-0837-11e7-8a75-5674ba5fc0f4.png)

edit:
Additional bug that might be related: 
https://github.com/mavlink/qgroundcontrol/issues/4767

DonLakeFlyer:
Pull doesn't fix this.

PR #4799, opened 2017-03-19
Survey bugs

DonLakeFlyer:
* Fixed some signalling bug with Portrait/Landscape
* Fixed item editor initial load setup bug with respect to fixed camera orientations
* Make sure to use clipToViewPort=false everywhere on Map.fromCoordinate, Map.toCoordinate


DonLakeFlyer:
Fix for #4765

PR #4805, opened 2017-03-20
Stop timeout when closing tile reply.

dogmaphobic:


DonLakeFlyer:
I assume this fixes the crash, awesome. If so can you connect the two together and then mark the issue as closed?

dogmaphobic:
I could not reproduce the crash. The only way I could think it could happen was if a timeout were to occur as you were exiting the program. There is only a minute chance of that ever happening. No response from a tile request for 10 seconds AND exiting? Because of that, I can't say categorically that this fixes anything.

DonLakeFlyer:
Ok. I was able to hit it a number of times. I'll try again with this.

Issue #4787, opened 2017-03-16
Mission item "Takeoff", dangerous bug.


AndKe:
Just discovered that a "takeoff" , sets absolute altitude of 50m 
That's well below ground level.

Additionally: takeoff item should have editable altitude.
When mapping an valley, or any place with a hill nearby, it's preferable to fly from the high elevation, just to ensure good telemetry link over a much bigger area.  In such cases, there is no point in climbing 50meters, (just to descend soon after.)

AndKe:
Extra info; it also affects a good takeoff command loaded from the plane - will be cahnged to ABS when uploading.  (even if the takeoff mission item was not touched)

DonLakeFlyer:
> takeoff item should have editable altitude.

@AndKe Which firmware? PX4 or ArduPilot?

AndKe:
ArduPilot.  - the result should be equally bad on both, unless PX4 treats anything as relative altitude.

DonLakeFlyer:
ArduPilot doesn't support takeoff altitude. Confused?

DonLakeFlyer:
Oh, wait I was looking at ArduPilot code wrong. Sorry.

AndKe:
Just check the fact that mavlink does have altitude type flag also for
takeoff, and - in QGC - creating Takeoff does not let you let you select..
Also:
I did also: (connect AP - Mavproxy - qgroundcontrol)
-Created takeoff in QGC
-uploaded to AP
-read back to mavproxy,
-in mavproxy: corrected to relative altitude.
-uploaded from mavproxy to AP
-read in QGC , (did not touch takeoff item)
-then just write mission from QGC to AP.
-takeoff was again modified to absolute altitude..

- yes, I do some checking before bitching :)

DonLakeFlyer:
Unfortunately not a simple fix because takeoff specifies an altitude but not a coordinate which screws up the concept of whether there is MAV_FRAME or now.

AndKe:
The yaw is also be important , when doing autotakeoff from runway ,
catapult or throwing is not so critical...

DonLakeFlyer:
Related to #4426

DonLakeFlyer:
@AndKe By tomorrow new daily builds show have this fix. Could you test?

AndKe:
yes, I am out flying, will do later today.

AndKe:
I was about to test, but when I compile current head , I get : 
```
 ./QGroundControl 
[...]

PR #4807, opened 2017-03-20
Plan: Waypoint now supports camera section

DonLakeFlyer:
This will roundtrip correclty through vehicle:
screen shot 2017-03-20 at 12 42 07 pm


DonLakeFlyer:
This pull also contains some of the scaffolding to fix #4787

PR #4813, opened 2017-03-22
Allow custom build to override app name information

DonLakeFlyer:


DonLakeFlyer:
This way settings don't conflict

PR #4817, opened 2017-03-22
Map Tile Set Import/Export

dogmaphobic:
This is the fist part of Import/Export of map tiles sets.

This PR implements export and import from disk. You can select what sets you want to export and these will be saved into a file. You can then "import" this file into another instance of QGC.

### DIFF
--- a/src/QtLocationPlugin/QGCMapEngineData.h
+++ b/src/QtLocationPlugin/QGCMapEngineData.h
@@ -389,12 +389,19 @@ public:
         emit exportCompleted();
     }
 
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
When you're down with this can you update this: https://donlakeflyer.gitbooks.io/qgroundcontrol-user-guide/content/DailyBuildChanges/DailyBuildNewFeatures.html

So many new things going in. need to keep track to let people know.

dogmaphobic:
The trick will be remembering this :)

PR #4835, opened 2017-03-23
New toolbar and user model for plan view

DonLakeFlyer:
Now when you go to plan view you see this:
screen shot 2017-03-23 at 4 04 04 pm

@dogmaphobic  Had the fantastic idea to do this. I just implemented it:
* Reuses toolbar real estate to show mission stats
* If the mission is dirty a button will show up on the right saying either "Upload" or "Save" depending on whether you are connected to a vehicle.
* When you click the Fly button to go back to Fly view you will be prompted to Upload/Save if your mission is dirty before you leave.

Hoping this makes the user model of needing to send the vehicel to the vehicle a little more understanable. It's work in progress from a standpoint of user model but everything there is working.


DonLakeFlyer:
The one gotcha with this is that the button in the toolbar currently only works with missions. It doesn't send geofence/rally point. This will all get fixed as well.

PR #4841, opened 2017-03-24
Plan toolbar: Add battery space

DonLakeFlyer:
* Plus minor repositioning for better readability
* Battery is placeholder only at this point. Real values coming soon.


DonLakeFlyer:
screen shot 2017-03-23 at 9 10 30 pm

PR #4856, opened 2017-03-26
New Takeoff, Start Mission, Resume Mission

DonLakeFlyer:
All of these will automatically arm the vehicle and then do the requested action.

Still working out support for these with ArduPilot.


dagar:
I'm not sure if it's relevant, but in PX4 I implemented MAV_CMD_MISSION_START. If a valid mission is available, it switches to auto:mission at the given index and arms.

DonLakeFlyer:
No that's not quite what I need. I need to be able to land in the middle of a mission and have the vehicle remember current mission item even through a power down to change battery. And the power back up, mission still there, correct current item still there. Switch to Mission flight mode, it goes back to the next item in sequence and finishes the mission.

DonLakeFlyer:
When I flew this in SITL it seems like the vehicle reset the current mission item when it landed.

PR #4859, opened 2017-03-26
General Settings tweaks.

dogmaphobic:


DonLakeFlyer:
This whole thing needs to be reworked to not use fixed widths. It currently blows up on small screens. So adding more won't help.

src/ui/preferences/GeneralSettings.qml
```diff
@@ -140,7 +140,7 @@ QGCView {
                         //-----------------------------------------------------------------
                         //-- Base UI Font Point Size
                         Row {
-                            visible: _appFontPointSize.visible
+                            visible: _appFontPointSize ? _appFontPointSize.visible : false
```
DonLakeFlyer: Why is this needed? This should never be missing.

dogmaphobic: It was throwing a Javascript exception because _appFontPointSize wasn't defined.

DonLakeFlyer:
It's also going to conflict like crazy with the stuff I'm doing with save paths already.

src/ui/preferences/GeneralSettings.qml
DonLakeFlyer: That's a bug somewhere then. Settings facts should never be missing.

PR #4863, opened 2017-03-27
New Plan a mission sync/save user model

DonLakeFlyer:
The way you work with missions in QGC has changed massively. There is no longer the concept of "Sync". In fact there is no Sync button at all any more. Let me walk you through it. Let's walk through and example where you are connected to a vehicle which has no current mission on it. It starts out with the mission editor looking like this:
screen shot 2017-03-26 at 8 30 09 pm

You can either load an existing mission or create a new one. When you load an existing mission by clicking one of the mission files, the mission will be loaded and also sent automatically to the vehicle.

You can create a new mission by just started to add way points. New missions can be named as well. In this example I've named my mission. You can see that the list of mission to load is now gone.
screen shot 2017-03-26 at 8 30 33 pm

Here I've added a bunch of waypoints. That's all the same.
screen shot 2017-03-26 at 8 30 46 pm

I now switch back to the Fly view and the mission is automatically sent to the vehicle. Since I named the mission the mission is also save to file automatically.

Here you can see the mission in the Fly View:
screen shot 2017-03-26 at 8 30 55 pm

Then if you go back to Plan view again you will see that there is now a Clear Mission button. Clicking this will clear the mission from Plan and the vehicle.
screen shot 2017-03-26 at 8 31 05 pm

### DIFF
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -646,6 +646,21 @@ void MissionController::loadFromFile(const QString& filename)
     MissionController::_scanForAdditionalSettings(_visualItems, _activeVehicle);
 
     _initAllVisualItems();
+
+    QString filenameOnly = filename;
+    int lastSepIndex = filename.lastIndexOf(QStringLiteral("/"));
+    if (lastSepIndex != -1) {
+        filenameOnly = filename.right(filename.length() - lastSepIndex - 1);
+    }
+    QString extension = AppSettings::missionFileExtension;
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
I'm sure there are still quirks in here and user model holes which I need to work through. But it's pretty close.

LorenzMeier:
Awesome!

DonLakeFlyer:
Complaints go here: https://github.com/mavlink/qgroundcontrol/issues/4868

PR #4873, opened 2017-03-28
Fix video save path in general settings

jaxxzer:
This fixes broken path selection.


DonLakeFlyer:
Related to #4861

DonLakeFlyer:
Thanks

PR #4879, opened 2017-03-28
QGCPalette work

dogmaphobic:
Adding colors for alert messages:

- alertBackground
- alertBorder
- alertText

I've also added some macros for color declaration and definition to make it simpler to create new colors.

A color is now defined (header file) as:
```
    DEFINE_QGC_COLOR(window, setWindow)
```

The definition (cc file):
```
    DECLARE_QGC_COLOR(window, "#ffffff", "#ffffff", "#222222", "#222222")
```

That's all there is to it.


DonLakeFlyer:
Oops I thought i was ok top delete 5.8 from travis. But forgot pulls would still use it till my thing went through. Will close/reopen once my stuff is through to build.

PR #4903, opened 2017-04-02
Plan: Support battery calculations

DonLakeFlyer:
Firmware plugin can now support battery information:
```
    /// Returns the data needed to do battery consumption calculations
    ///     @param[out] mAhBattery Battery milliamp-hours rating (0 for no battery data available)
    ///     @param[out] hoverAmps Current draw in amps during hover
    ///     @param[out] cruiseAmps Current draw in amps during cruise
    virtual void batteryConsumptionData(Vehicle* vehicle, int& mAhBattery, int& hoverAmps, int& cruiseAmps) const;
```

Plan view then uses this information to display number of batteries needed and first battery change waypoint in the Plan toolbar.

### DIFF
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -437,3 +437,11 @@ bool FirmwarePlugin::_armVehicle(Vehicle* vehicle)
     }
     return vehicle->armed();
 }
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
Next step is much better hover and cruise time calculations based on waypoint types.

Issue #4909, opened 2017-04-03
AnalyzeView controller lifetime tied to current page


NaterGator:
The controller backing AnalyzePages is created when the page is loaded and destroyed when it is replaced in the loader. For example, with a log download in progress you cannot switch to GeoTag Images or your log download will terminate.

@DonLakeFlyer if you're okay with it, I'll submit a PR to instantiate the controllers for each AnalyzePage on the AnalyzeView. When loaded the AnalyzePage can access its controller from the view.

DonLakeFlyer:
Yeah these should stay live once you go there.

NaterGator:
Ok. I'm cooking up a mavlink console page as well, which is how I came across this. I'll have a PR in just a bit.

DonLakeFlyer:
Fixed

PR #4912, opened 2017-04-03
Make AnalyzeView controllers persistent

NaterGator:
For #4909

### DIFF
--- a/src/AnalyzeView/AnalyzeView.qml
+++ b/src/AnalyzeView/AnalyzeView.qml
@@ -36,6 +36,14 @@ Rectangle {
     readonly property real  _verticalMargin:        _defaultTextHeight / 2
     readonly property real  _buttonWidth:           _defaultTextWidth * 18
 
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
@dogmaphobic I think this is fine, but can you look?

DonLakeFlyer:
@dogmaphobic Ok, for log download?

dogmaphobic:
> Ok, for log download?

Sorry, I've been swamped. I looked through the code and it all seems fine. Nothing special is done during construction/destruction so I don't see any issues keeping it alive. 

I only went through the code. I did not actually test it.

DonLakeFlyer:
Crap. @NaterGator Can you rebase?

NaterGator:
Sure thing, just a sec.

DonLakeFlyer:
Travis passed. Appveyor seems to have gone bonkers. Should be ok.

PR #4913, opened 2017-04-03
Add mavlink console support

NaterGator:
This PR adds mavlink console support to the analyze page. 

I have tested it over USB with recent PX4 firmware and it seems to do decent job. Very basic ANSI code handling is done so that top(like) applications work as expected. As of right now I don't have command history; you just type your command into the text field at the bottom and press enter to issue it over mavlink. 

Action shot:
image

Would be happy to get feedback and improve this so we can get it merged.

### DIFF
--- a/qgcresources.qrc
+++ b/qgcresources.qrc
@@ -56,6 +56,7 @@
         <file alias="GeoFence.svg">src/AutoPilotPlugins/PX4/Images/GeoFence.svg</file>
         <file alias="GeoFenceLight.svg">src/AutoPilotPlugins/PX4/Images/GeoFenceLight.svg</file>
         <file alias="GeoTagIcon">src/AnalyzeView/GeoTagIcon.svg</file>
+        <file alias="MavlinkConsoleIcon">src/AnalyzeView/MavlinkConsoleIcon.svg</file>
         <file alias="LandMode.svg">src/AutoPilotPlugins/PX4/Images/LandMode.svg</file>
         <file alias="LandModeCopter.svg">src/AutoPilotPlugins/PX4/Images/LandModeCopter.svg</file>
         <file alias="LightsComponentIcon.png">src/AutoPilotPlugins/APM/Images/LightsComponentIcon.png</file>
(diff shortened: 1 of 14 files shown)


DonLakeFlyer:
Is this a mavlink generic thing?

NaterGator:
It is a part of the mavlink common spec: http://mavlink.org/messages/common#SERIAL_CONTROL_DEV_SHELL 

I honestly can't speak as to whether other AP stacks have implemented it. Are you thinking this needs to be built with FirmwarePlugin support?

DonLakeFlyer:
As long as it fails gracefully if not supported this is fine.

NaterGator:
Looks like it is supported in ardupilot for at least a few stacks: https://github.com/ArduPilot/ardupilot/blob/c808ee2f497e7a7a6d570632a3224e655b377323/libraries/GCS_MAVLink/GCS_serial_control.cpp#L62-L65

The current failure mode is you just get no console output if the AP stack doesn't support it. I could show something akin to a spinner to indicate the GCS is expecting(/waiting?) to receive a response from a user command.

DonLakeFlyer:
I don't think it's worth getting fancy if not supported.

jaxxzer:
This is working with APM. Something I've noticed is that I must hit enter repeatedly to get the lines output by nsh to feed out into this console. Is this happening with PX4 as well?

NaterGator:
@jaxxzer no, that is definitely not the case on PX4. I wonder if APM expects an empty ACK (SERIAL_CONTROL_FLAG_RESPOND, or something similar). 

I'll see if I can track that down.

NaterGator:
@DonLakeFlyer if @jaxxzer can confirm the fix for APM and you're happy I'm good with this going into daily builds for wider testing.

DonLakeFlyer:
Ok. I haven't had time to look through it yet. I'll merge after I look.

DonLakeFlyer:
@NaterGator Can you move the stuff from UAS to Vehicle. UAS is deprecated. I'm slowly moving everything out of there and it will eventually go away.

NaterGator:
Will do, I was confused where it should live.

NaterGator:
Ok, rebased on master and moved the mavlink handling onto Vehicle.

PR #4917, opened 2017-04-04
Bulk mission altitude change, plus pile of bug fixes

DonLakeFlyer:
Bulk Altitude change:
* Change the default mission altitude in Mission Settings
* You are prompted whether you want to apply this new altitude to your existing mission
* Yes and all items except LAND items are updated to new altitude

Bug fixes:
* Takeoff item editor for PX4 multi-rotor was showing pitch field
* Slide to Confirm thingy on mobile was too tall visually
* Toolstrip recalc to determine whether to show/hide text labels working incorrectly
* Planned home position waypoint not draggable


DonLakeFlyer:
@AndKe It's not a grid editor, but I think it covers one of the big cases of bulk update.

PR #4938, opened 2017-04-07
New Arm/Disarm toolbar indicator

DonLakeFlyer:
It's not very pretty but functionally it works:
* This shows up in the toolbar if a vehicle has auto-disarm turned off or if the vehicle is a fixed wing. Still discussing whether this should always be there. The reason for fixed wing is that a fixed wing vehicle doesn't spin props like multi-rotor so it isn't very clear whether the vehicle is armed or not.
* If you click on it you can arm/dismarm with a confirmation. Clicking does nothing if you are flying (still thinking about what to do here).

screen shot 2017-04-07 at 12 07 29 pm

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -3,6 +3,7 @@
         <file alias="FactSystemTest.qml">src/FactSystem/FactSystemTest.qml</file>
     </qresource>
     <qresource prefix="/toolbar">
+        <file alias="ArmedIndicator.qml">src/ui/toolbar/ArmedIndicator.qml</file>
         <file alias="BatteryIndicator.qml">src/ui/toolbar/BatteryIndicator.qml</file>
         <file alias="GPSIndicator.qml">src/ui/toolbar/GPSIndicator.qml</file>
         <file alias="MessageIndicator.qml">src/ui/toolbar/MessageIndicator.qml</file>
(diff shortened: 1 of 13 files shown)


DonLakeFlyer:
@antiheavy

DonLakeFlyer:
Maybe click on it while flying is where emergency stop should go?

DonLakeFlyer:
The reasoning for possible showing it all the time is that even with multi-rotor props spinning on armed if the vehicle is some distance away from you it is still hard to tell.

DonLakeFlyer:
@LorenzMeier FYI

Antiheavy:
I've also been thinking about Emergency Stop and how it is different than Disarm.  Is the idea of Emergency Stop is to stop spinning motors the motors?  I'd think Disarm is an effective tool for this when the aircraft is not in a flying state.

However, for Fixed Wing I'd prefer "Disarm" is locked out in flight.  I still want the ability to disable/enable the propulsion motor(s) separately from flight control servos.  This is an effective way to emergency land a fixed wing, i.e. kill the motor but let the servos maintain airspeed and control as the aircraft glides down.  Not sure how this would fit into the UI and how the system differentiates propellers from servos.

DonLakeFlyer:
Emergency stop kills motors. It should still keep servos going though.

Antiheavy:
In that case I think your idea of Emergency Stop there is good.  When not flying show "Disarm", when flying show "Emergency Stop" with "Are you super duper sure?!?!" confirmation dialog.

I've noticed in HIL simulation that re-arming in flight after Emergency Stop causes the home point to be updated and the Altitude-rel to be reset above the ground.  Not super desirable, but by that point you are already in an emergency situation so maybe you work with what you've got.

dogmaphobic:
Maybe it's a bit too subtle but it doesn't take all that space in the tool bar:

armed

DonLakeFlyer:
Yeah, I figure you'd do something with it. Looks great to me.

PR #4939, opened 2017-04-07
Sub updates

jaxxzer:
This fixes up a couple things for Sub UI.


DonLakeFlyer:
ArduPlane has had no testing for guided mode so should not be turned on. Unless you tested the whole thing yourself?

jaxxzer:
No, I wasn't sure about this actually. I saw that Plane does have a guided mode, and the last commit here was just to retain the current behavior for other vehicles.

Issue #4905, opened 2017-04-02
PX4 Pro - Mission Items: Firmware supports Yaw, ui info out of date


DonLakeFlyer:

DonLakeFlyer:
MAV_CMD_NAV_WAYPOINT
MAV_CMD_NAV_LAND
MAV_CMD_NAV_TAKEOFF
MAV_CMD_NAV_VTOL_TAKEOFF
MAV_CMD_NAV_VTOL_LAND

Issue #4902, opened 2017-04-02
Survey: YAW is defaulting to North


tops4u:
When setting MIS_YAWMODE=0 (as defined by Waypoint) and using Survey with a non default Grid Angle the MC will always use YAW=0°. 
My expectation was that if I set a specific Grid Angle that it would use this for YAW. 
Currently I have IMHO no way to fly a Survey with a fixed YAW Angle for the whole Survey.

For all other Settings of MIS_YAWMODE the YAW Angle of the Waypoint should have not influence, so SURVEY should IMHO take YAW as specified from the Survey Grid Angle.

DonLakeFlyer:
Survey does not pay any attention to mission yaw parameters. You are expected to set that up correctly. For Survey you should be using MIS_YAWMODE=1 so that the vehicle always points in the direction of flight. Is there a problem with using MIS_YAW_MODE=1?

tops4u:
Yes. We intend to have YAW in a fixed direction for the whole mission. Currently this is impossible. Would it be possible to habe Survey take over the Grid Angle. In any other Setting of MIS_YAWMODE this would have no Impact.

DonLakeFlyer:
You are going to need to handle this manually. The way to do it is to set a waypoint before the mission with the heading set to what you want. Then set the parameter to not change yaw. The problem is that as far as I currently know. Neither PX4 Pro or ArduPilot firmware support yaw in a waypoint mission command item. 

I don't see adding this as a feature to survey unless it becomes a common request (and the firmwares support it). Survey is already complicated enough with all the settings.

If you find that either of the firmwares supports setting yaw in a waypoint command then let me know and I'll update QGC to allow for that. If not, I'd ask them to start supporting it and once again I'll update to that new support.

tops4u:
@DonLakeFlyer How am I going to handle this manually? Does Survey not _simply_ translate the Survey Area into Waypoints - where "simply" refers to my understanding. 
When I set YAW for PX4 Waypoints this works perfectly well until I reach Survey where the UAV will turn YAW to north and continue the Survey with a North Heading. So I thought that when Survey creates the Waypoints I could simply apply the GRID Angle as YAW Setpoint!? 

Am I misstaken?

DonLakeFlyer:
Hmm, I'll have to do some poking around the firmware code.

DonLakeFlyer:
What I've discovered so far:
* PX4 Pro supports Yaw in waypoint
* ArduPilot does not

[...]

PR #4942, opened 2017-04-08
Add support for NaN in mission command params

DonLakeFlyer:
* Added new setting for MissionCmdParamInfo: nanUnchanged. This indicates that this CMD_LONG parameter supports NaN to signal the value is unchanged.
* Use this new settings on the heading params for the appropriate commands. Also fixed up a bunch of problems with PX4 MavCmdInfo capabilities being out of date.
* Update SimpleItemEditor to support NaN fields.

Here is what it looks like:
screen shot 2017-04-07 at 8 52 05 pm

* Check the box for Heading to specify a heading
* Leave unchecked to set as NaN which means leave heading alone

Fixes for: #4905, #4902

### DIFF
--- a/src/FirmwarePlugin/PX4/MavCmdInfoCommon.json
+++ b/src/FirmwarePlugin/PX4/MavCmdInfoCommon.json
@@ -7,7 +7,7 @@
         {
             "id":           16,
             "comment":      "MAV_CMD_NAV_WAYPOINT",
-            "paramRemove":  "2,3,4"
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
I should also say that Survey now sets heading to NaN in the waypoints it creates. This allows you to create a survey where you set a fixing heading before the survey (assume you also set the correct parameter which turns off heading follows mission direction).

PR #4953, opened 2017-04-09
Exception created to improve TTS saying ADSB

lvale:
To improve the way that TTS engines say ADSB


DonLakeFlyer:
Can't these be "I D" and " A D S B"?

lvale:
I tried " A D S B " and didn't sound right, so had to opt to some "phonetic" approach
Using both AVA, SAMANTHA and ALLISON TTS voices

lvale:
the ID substitution was from @dogmaphobic and it also sounds right

DonLakeFlyer:
Ok, got it.

PR #4960, opened 2017-04-11
Fix ToolStrip sizing/visability

DonLakeFlyer:
@jaxxzer I reversed to initial state of the toolstrip. If there is no vehicle the ToolStrip always shows. It's only after the vehicle shows up that it used guided mode to hide. I don't like that fact that for the majority use case (guided mode supported) it comes and goes which is strange visually. So I've switched the strangeness to the minority use case.e


jaxxzer:
Ok, that's a good idea, works for me.

Issue #4445, opened 2017-01-16
Integrate crash reporting tools


NaterGator:
At least on non-mobile applications we would be well served to handle crashes more gracefully. As mentioned [here](https://github.com/mavlink/qgroundcontrol/pull/2680#issuecomment-173740648) a decent choice would be the Chromium project's [breakpad](https://chromium.googlesource.com/breakpad/breakpad/) library. 

We can start small and just generate crash dumps when an unhandled exception occurs (users would have to manually upload them). Later down the line it would be nice to have a built in console / tool to facilitate users submitting stored crash reports.

NaterGator:
@DonLakeFlyer what do you think about integration with https://drdump.com ? We can upload the release build symbols to their service and collect crash dumps from Windows there.

DonLakeFlyer:
That sound like a great idea. Do we need to start flipping some bits in the installer to get better crash reports?

jaxxzer:
@NaterGator have you done any work on this? I'm interested in helping get this together.

NaterGator:
I looked into this and we would have to package binary blobs that we can't audit from a software company in Russia.  Didn't / doesn't feel right. I'd like to find an open-source and auditable solution to this problem.

DonLakeFlyer:
Breakpad support is working it's way into 3.4 release hopefully

patrickelectric:
Hey guys, any update on it ? 
probably waiting for 5.12 to have the debug symbols ?
Any motive to use breakpad and not other project ?

patrickelectric:
I'm adding a crash report for one application, you can see the output window here:
![screenshot from 2018-11-15 19-27-36](https://user-images.githubusercontent.com/1215497/48583863-669ac380-e90f-11e8-9e2a-0783905b9c0e.png)

After doing the deploy and tests for windows , linux and mac. I'll add it in QGC.

DonLakeFlyer:
Would be fantastic to get this all working cross-platform. Huge win.

jaxxzer:
We are doing some work with [KCrash](https://api.kde.org/frameworks/kcrash/html/index.html
) and [drkonqi](https://github.com/KDE/drkonqi
)

We do intend to attempt some integration with QGC. Will let you all know how that goes.

DonLakeFlyer:
The listed support for KCrash is "FreeBSD, Linux, Windows". Does that include OSX and Android?

patrickelectric:
[...]

PR #4977, opened 2017-04-13
Breakpad support for android

DonLakeFlyer:


DonLakeFlyer:
@dagar The android build now builds with -g so it has symbols in there. Seems like if I put the linux dump_syms tool in the repo I could create the breakpad symbol file in travis. Upload that to S3 as well. And then there must be some tool I can run to strip the symbols from the .so before we push up to S3? That seems possible doesn't it?

dagar:
@DonLakeFlyer 

    strip --strip-all release/libQGroundControl.so

NaterGator:
@DonLakeFlyer @dagar this is relevant to #4445 too. If Android is getting breakpad support we should consider adding it for Windows... at least for CI builds.

dagar:
It sounds like we should be able to use it everywhere?

Do we need something like Mozilla Socorro? https://wiki.mozilla.org/Breakpad

DonLakeFlyer:
Yeah, trying on Android first since we have some unexplained crashes there. Then we can:
* Move to all OS
* Figure out an auto-upload mechanism/server

DonLakeFlyer:
FYI: It's not really automatic to get this to work crossplatform. You need to figure each one out separately. So it's not just a matter of flipping a build switch.

NaterGator:
Understood. If it is any consolation I already have it implemented in Windows.

DonLakeFlyer:
If you guys could figure out how to get this working with upload and a server that would be awesome.

Issue #4989, opened 2017-04-15
Calibration predialog text collapse problem


feemi:
![calibration screen bug](https://cloud.githubusercontent.com/assets/1747208/25062378/dce0664e-21d3-11e7-964e-eea44f86ddfc.png)

There is a text collapse on calibration predialog while opening compass and accelerometer calibration

PR #4990, opened 2017-04-15
Fixed calibration predialog text collapse problem.

feemi:
#4989 issued problem resolved.


DonLakeFlyer:
Thanks @feemi 

Hmm, this is some pretty old Qml code. All spacing should be based off of ratio of ScreenTools.defaultFontPixelHeight or Width depending on direction. Fixed pixels don't work across different device sizes.

Could you instead find a good font ratio to use here?

feemi:
"ScreenTools.defaultFontPixelWidth / 2" looks fine I think.

DonLakeFlyer:
Perfect

feemi:
I just ammended the code again

PR #5032, opened 2017-04-20
Make qgroundcontrol less error-prone for cloning behind corporate firewalls

rhvonlehe:
This commit simply changes one of the submodules from git to https transfers making it less error prone behind stupid corporate firewalls that are too restrictive.


DonLakeFlyer:
Thanks.

PR #5033, opened 2017-04-20
Use a master controller for all Plan elements

DonLakeFlyer:
This is the first pull in a sequence of pulls which eventually will fix the incredibly nasty user model problem which is that if you start planning a mission offline and then you connect a vehicle. The mission you were planning just mysterisouly disappears. Needless to say that can be quite annoying. I had "forgotten" about this. Mainly I think because it was such a royal pain to fix.

* The first major change is to the user model of dealing with missions, geo fences and rally points. Previously these were treated as individual elements. Each needing there own file load/save, upload/download and so forth.
* The new model is that there is just one top level thing call a "Plan". A Plan file (file extension .plan) contains a mission, geo-fence and rally points in a single file. When you upload/download QGC upload/downloads all elements (mission, fence, rally) to the vehicle.

Next steps:
* Fix loading of old file formats: .mission, .waypoints using the new system. I broken that with this changed.
* Finally fix the real problem of disappearing missions.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -472,6 +472,7 @@ HEADERS += \
     src/MissionManager/MissionManager.h \
     src/MissionManager/MissionSettingsItem.h \
     src/MissionManager/PlanElementController.h \
+    src/MissionManager/PlanMasterController.h \
     src/MissionManager/QGCMapPolygon.h \
     src/MissionManager/RallyPoint.h \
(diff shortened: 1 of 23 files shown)


DonLakeFlyer:
I also completely forgot to say that this pull removes support for auto-sync as well. That was an interesting attempt to make things better and learn things from. But in the end it was too confusing and had too many quirky border cases which caused strange results.

PR #5037, opened 2017-04-21
[WIP] try qtquickcompiler

dagar:


dogmaphobic:
Let's see how it goes. When I tried this earlier, it also required `QT += qml-private`. Also, I was under the impression this was only available for the commercial license?

dagar:
It's in the open source version now.

DonLakeFlyer:
Isn't a different license though?

dagar:
Need license clarification.

PR #5041, opened 2017-04-22
Add wind rose in survey pattern editor

patrickelectric:


DonLakeFlyer:
Can you add some details of what this is. Like some words and screenshots?

patrickelectric:
The crosswind is a know best practice to perform the survey operation. To help the user it was added a wind rose to perform the input  of wind direction.
selection_465
selection_466
selection_467

DonLakeFlyer:
Thanks for putting the effort into this. My concern is to add more complexity to an already complex ui. Given that, I'm not sure whether this helps or makes things even more complicated. I assume the problem this is trying to fix is to help the user translate from say a wind direction of NE to something like 22 degrees of grid angle. But that seems pretty simple to do. It would need to valuable enough and solve a good problem to justify yet even more stuff in this ui.

The other things it does is provide a non-numeric way of entering and angle which may be easier than typing. That would be nice and has been in the back of my head for a while. But something built around a slider I think would be better for that since it allows for more values.

@AndKe, @antiheavy: You guys fly fixed wing survey a lot. as this helpful? Or is it already easy enough to translate wind direction into numbers?

AndKe:
@DonLakeFlyer  I do rotate a mission depending on the wind, When arriving at a site, estimating something >7m/s at altitude , I may rotate the mission to avoid direct tailwind that may give high groundspeed..  This usually depends on desired GSD. 
or I may change the direction of the survey to make narrower turns when near mountain side. 

In any case, being at the site and seeing the map, I would not have any use for a tool to help me modify mission, wind directions change and it's not a even a +/-10 degree science.    

The only wind related calculation the software could do, is what I do using custom systems below: (but that is kind of extreme, but nice, for a GCS)
I calculate the minimum altitude over terrain in the mission,(survey areal is not flat), enter expected maximum effective  tailwind. Based on lens FOV, sensor resolution, maximum groundspeed , I then get two things:  required shutter speed (below which motion blur will appear)  and minimum photo interval (I trigger by distance, but high ground speed will reduce time, interval between shots need to be > than the minimum camera is able to store photos at.) 

So while I do not expect the planning tool to take wind into account, I would like to see wind direction and speed during flight as a part of flight monitoring tools. 

 (unless you want to make the survey tool very advanced)

Antiheavy:
Good topic.

Yes, flying crosswind is a best practice for fixed wing survey patterns.  Ideally the pilot would start the survey pattern downwind and snake through the pattern working upwind.  However, in real life there are always additional considerations such as terrain, restrictions on overflying adjacent land, buildings, etc.  If you have a long skinny area you may just want to suck it up and fly with a tailwind to avoid turning all the time.  Wind is just one consideration.

I DO like the idea of a easier way to set the angle of the survey vs typing in a number.  This could be accomplished many ways.  I actually think @patrickelectric wind rose is pretty cool, although the user has to know the direction of the arrow is actually 90 degrees to the resulting flight path and the starting position will be on the downwind side.  This might be a little confusing to some users, although once they figure it out they might appreciate this shortcut button instead of typing in 135 degrees and then realizing that is opposite of what they want and then having to do the math in their head again and type in 315 degrees.

I fully support the battle against complexity, especially as it relates to improving safety and reliability.  I'll admit I started typing this comment with the idea that this idea, while neat, wasn't worth it.  However, the more I think about it the more I really like this wind rose idea.  A lot of our users are doing agricultural mapping in wide open areas and would probably find this shortcut useful.

AndKe:
ok , now it seems we are talking about a rotation tool rather than anything wind-related, the most intuitive thing could be to let user grab something like the red dot with arrows (for illustration purposes), and drag it around, while the grid would always be perpendicular to the direction between map center, and the grab-drag item . 
screenshot from 2017-04-23 19-29-03

that would be totally foolproof, user would not need to know how to add/subtract at all :)

patrickelectric:
@AndKe Good idea, I will try to develop something like it.

DonLakeFlyer:
@patrickelectric What @AndKe describes is what I've had in the back of my head for a long time.  Never had to time to do though. That would be much better. The starting point is SurveyMapVisuals.qml.

Antiheavy:
I like both ideas.  Something along the lines of what AndKe sketched is probably better overall.

DonLakeFlyer:
@patrickelectric I would rather have the direct manipulation of the grid angle on the survey item only instead of the compass rose since it adds no additional complexity to the Survey item editor.

patrickelectric:
image

patrickelectric:
I modified the last version to have more control over the angle, showing the entering point and selecting the wind direction. I think that the survey item have to much complexity in smartphones/tablets, editing the angle in survey editor make more sense to me.

DonLakeFlyer:
Looking good, Some changes:
* The compass rose image doesn't really look enough like a compass rose
* You should use QGCColoredImage to color the control using the palette. I'd try a color of windowShadeDark. And make sure it works in both color themes.
* Only show the control for fixed wing vehicles
* The control doesn't show up in all the cases where grid angle is shown.

AndKe:
looks good

patrickelectric:
@DonLakeFlyer,  everything done.

selection_484
selection_483

DonLakeFlyer:
Awesome. Hope to get to this tonight.

DonLakeFlyer:
The angle tool doesn't show up with Manual Grid:
screen shot 2017-05-08 at 7 33 10 pm

DonLakeFlyer:
Also is it possible to make the spinner control thing go away if it loses focus. Right now you can get into weird states if you click on other things like this:
screen shot 2017-05-08 at 7 40 45 pm

patrickelectric:
@DonLakeFlyer Done.

patrickelectric:
In manual grid the FactTextField is not respecting the Layout.colunSpan and I don't know why :(
image

DonLakeFlyer:
I'll take a look.

DonLakeFlyer:
Yeah, you run into a nasty layout problem there. It will require some larger restructuring which I can deal with myself. Thanks for all the work on this. Nice stuff.

PR #5048, opened 2017-04-23
Plan: More major re-architecture

DonLakeFlyer:
* Plan and Fly view shows now update and not update at the right times among other things

This should be the last of the major changes. Now just testing and bug fixing.


DonLakeFlyer:
This pull also fixes a bug with the camear trigger button in the instrument panel.

PR #5073, opened 2017-05-01
Pile of fixes

DonLakeFlyer:
* Convert Q_ASSERT to qWarning. Other than unit tests Q_ASSERT should never be used.
* Radio Buttons and Check Boxes now have better mobile touchability/visuals
* Vehicle trajectory lines and camera trigger point indicators on Fly view are now cleared at appropriate times.
* Plan view: Mobile file picker allows selecting .plan and .mission files.
* PX4 Safety component page had missing icons
* Vehicle flight mode was announced in cases where the flight mode didn't actually change


NaterGator:
How about qCritical instead of qWarning? Asserts that are failing should be taken pretty seriously.

DonLakeFlyer:
I don't use qCritical anywhere else in QGC. The highest ever used is qWarning. So matching that.

PR #5095, opened 2017-05-05
Add support for FMUv4PRO and FMUv5 boards

LorenzMeier:
This adds support for the full-sized FMUv4PRO design and the FMUv5 STM32F7 based generation.


DonLakeFlyer:
@dagar Do you know what the AppVeyor failure is? Could it be because this branch is not off of a fork?

dagar:
I'm not sure where it's getting that s3 region from, the correct one is in the config. Let me poke around.

LorenzMeier:
Fixed and rebased. When this passes it's good to go.

PR #5099, opened 2017-05-06
Fixes

DonLakeFlyer:
* Show advanced ui confirmation now mentions possible voiding of warranty
* Max alt for missions settings is 400ft, you can Force Save to go higher

### DIFF
--- a/src/Settings/App.SettingsGroup.json
+++ b/src/Settings/App.SettingsGroup.json
@@ -52,6 +52,7 @@
     "type":             "double",
     "defaultValue":     50.0,
     "min":              0.0,
+    "max":              121.92,
     "units":            "meters",
     "decimalPlaces":    2
 },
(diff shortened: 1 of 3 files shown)


kikislater:
Max 400ft is very limited. Why choosing this value ?

DonLakeFlyer:
400ft is max allowed in US. In reality the altitude in unlimited you just need to use Force Save to override the max.

kikislater:
Ok I will force it in that case ... 
But there is not only US users, is it only for US ?

dogmaphobic:
According to this, your limit is around 492' (150m)
http://www.mlvdrone.fr/rules-for-flying-recreational-drones
Each country will have their own rules. The question is how to make it "global"?

DonLakeFlyer:
I needed to pick some max to prevent fly away from bad values. Also keep in mind that this is Mission Settings only. It's mainly a simple stop-gap. You can still adjust individual waypoints to what you want.

kikislater:
I don't fly only in France, so it's not only 150m but you're right, make it global or not that is a question ...

PR #5100, opened 2017-05-06
Flight timer

DonLakeFlyer:
* Flight Time is now available from the Values Widget: Vehicle.flightTime
* Default Values Widget setup includes Flight Time

Existing QGC users who have already modified the Values Widget will need to add Flight Time manually if they want it.

screen shot 2017-05-06 at 2 57 19 pm

### DIFF
--- a/src/FactSystem/Fact.cc
+++ b/src/FactSystem/Fact.cc
@@ -270,6 +270,18 @@ QString Fact::_variantToString(const QVariant& variant, int decimalPlaces) const
         }
     }
         break;
+    case FactMetaData::valueTypeElapsedTimeInSeconds:
+    {
+        double dValue = variant.toDouble();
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
@dogmaphobic I added Flight Time as a standard part of the Vehicle FactGroup.

DonLakeFlyer:
Oops I missed pointing out that this pull also allows a core plugin to override the default setup for the Values Widget.

hamishwillee:
> Oops I missed pointing out that this pull also allows a core plugin to override the default setup for the Values Widget.

Is that something you need to record in the devguide?

DonLakeFlyer:
Yes. The dev guide currently includes no information on core plugin support at all.

PR #5103, opened 2017-05-07
Mission settings: No max altitude

DonLakeFlyer:
Decided to back to the max altitude of 400ft. Instead third parties can set as they like in their versions using settings override.


DonLakeFlyer:
@kikislater FYI

Issue #5078, opened 2017-05-02
Camera gimbal setup changing wrong parameters


jaxxzer:
After the RC/Servo input/output split in APM (and subsequent changes in QGC to support this), the servo limits and reverse options in the camera setup tab are applying to the RCn_ input parameters, instead of the SERVOn_ output parameters.

DonLakeFlyer:
Have you verified this? QGC has parameter remapping. Is that not working?

jaxxzer:
Yes, I've verified this. The RCn_FUNCTION -> SERVOn_FUNCTION parameter remapping is working, but the limits and reverse are still applied to the RCn_ group.

PR #5104, opened 2017-05-08
ArduPilot Camera Setup:  Fix parameter mapping

DonLakeFlyer:
Fix for #5078 

@jaxxzer Can you review this?


jaxxzer:
Yes.

jaxxzer:
Looks good, as far as I can see, this doesn't affect Radio setup. As long as that's the case, I think this is good to go.

PR #5111, opened 2017-05-08
Plan: Add support for SET_CAMERA_MODE command

DonLakeFlyer:
* Add support for new SET_CAMERA_MODE command in Camera Section. This allows you to change photo/video mode.
* You can also add the SET_CAMERA_MODE command manually through change command. This gives you additional access to audio recording options.
* If you add a Survey item to your plan Mission Settings camera section is automatically set to specify Photo Mode.


hamishwillee:
@DonLakeFlyer If you think any PR will result in us needing a new image for the site, please tag me with what you'd like to see.

DonLakeFlyer:
My guess would be that almost all the the images are out of date now. I think the think the thing to do is to use the Stable 3.2 release which is on the way out to be clean starting point. Then after that I can tag visual changes incrementally.

hamishwillee:
Works for me. What I'm trying to do is work out a process that makes it easy for you to share any work out to others (currently me, but ideally the "community") and not get lost.

DonLakeFlyer:
Should we have a github label on pulls which have a visual change?

hamishwillee:
Yes, if that is the easiest way to track changes. What might be even cooler is see if we can automate some tooling to auto-generate screenshots of everything on a regular basis.

Issue #4754, opened 2017-03-12
Mission planning / Survey with non contiguous areas


lvale:
I was doing some planning and the area allowed was very irregular, which shouldn't be a problem.

The planning ignores some parts of the survey area and the end result is a incomplete mission

Easily reproducible anytime (using latest Master) on OSX

<img width="1440" alt="screen shot 2017-03-12 at 17 39 55" src="https://cloud.githubusercontent.com/assets/3130363/23834259/711f5e8a-074b-11e7-87b1-16fd7b487795.png">

<img width="1273" alt="screen shot 2017-03-12 at 17 31 10" src="https://cloud.githubusercontent.com/assets/3130363/23834292/10729ee8-074c-11e7-9e29-1ae296bd9fdf.png">

Planning a mission

https://youtu.be/KPUsQ8s9WAk

DonLakeFlyer:
Fixed

PR #5115, opened 2017-05-09
Survey area is always convex polygon

DonLakeFlyer:
Fix for #4754


DonLakeFlyer:
screen shot 2017-05-08 at 6 48 43 pm

PR #5121, opened 2017-05-09
Add firmware-specific factgroups

jaxxzer:
This will add some valuable information for Sub operation in the vehicle values widget. Because these options are sub-specific, they are allocated at runtime if the vehicle is MAV_TYPE_SUBMARINE.

I think the way the options are created could use a rework in general to be firmware-specific, but maybe that should wait till after 3.2. In the meantime, I can rework the implementation in this PR if it is not appropriate; I am open to alternative approaches.

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -102,7 +102,7 @@
         <file alias="QGroundControl/Controls/qmldir">src/QmlControls/QGroundControl.Controls.qmldir</file>
         <file alias="QGroundControl/Controls/RallyPointEditorHeader.qml">src/PlanView/RallyPointEditorHeader.qml</file>
         <file alias="QGroundControl/Controls/RallyPointItemEditor.qml">src/PlanView/RallyPointItemEditor.qml</file>
-		<file alias="QGroundControl/Controls/RallyPointMapVisuals.qml">src/PlanView/RallyPointMapVisuals.qml</file>
+        <file alias="QGroundControl/Controls/RallyPointMapVisuals.qml">src/PlanView/RallyPointMapVisuals.qml</file>
         <file alias="QGroundControl/Controls/RCChannelMonitor.qml">src/QmlControls/RCChannelMonitor.qml</file>
         <file alias="QGroundControl/Controls/RoundButton.qml">src/QmlControls/RoundButton.qml</file>
         <file alias="QGroundControl/Controls/SectionHeader.qml">src/PlanView/SectionHeader.qml</file>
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
I think the way to do this is to have FirmwarePlugin expose a list of FactGroup objects. Then Vehicle exposes that list. And then values widget has an additional repeater for those extra groups. Then its fully generic. Make sense?

jaxxzer:
Yes, thank you for the input. I'll try that.

jaxxzer:
@DonLakeFlyer I've reworked this, but not exactly according to your suggestions. I think this is a little simpler:

- Firmware plugin [exposes a factgroup](https://github.com/mavlink/qgroundcontrol/pull/5121/files#diff-7431665394c54b0bc65104e9ec82c677R298) (I'm not sure what the motivation for a QVariantList is here).

- Vehicle[ exposes that factgroup](https://github.com/mavlink/qgroundcontrol/pull/5121/files#diff-02e75c7124017cb6a653bdfaea5daa5eR369), if available, by simply adding it to it's own factgroups.

- Views widget functions the same (no added repeater).

- Updating firmware facts is handled by firmware plugin.

The last commit in view widget qml is to handle qml errors in the case that the saved firmware-specific options are not currently available (different firmware currently in use).

I can rework this again, this felt like a better way to me.

DonLakeFlyer:
> I'm not sure what the motivation for a QVariantList is here

The reason for the QVariantList is so that FirmwarePlugin can expose multiple FactGroups instead of just one.


> Vehicle exposes that factgroup, if available, by simply adding it to it's own factgroups.

Yeah, better way.

DonLakeFlyer:
With your way of having vehicle merge FactGroup then you don't need a QVariantList since that is to expose object lists to QML. You just need to return a QList<FactGroup*> from FirmwarePlugin.

jaxxzer:
I've added the QList for firmwareplugin factgroups.

FactGroups can be nested. Would it be better to leave it as a single root factgroup for firmwareplugin, and nest additional factgroups into that root instead of appending all factgroups to a list?

DonLakeFlyer:
Hmm, good point. I think it still may be valuable to be able to add multilple top level groups. Although don't have a big reason for it other than it gives you both options which doesn't seem bad.

jaxxzer:
Ok. This is the only bit that bugs me after the change:
https://github.com/mavlink/qgroundcontrol/pull/5121/commits/34ac53c4915c357555bb16f2efa6b684ce86d18c#diff-02e75c7124017cb6a653bdfaea5daa5eR369

I could add a list of factgroup names to firmware plugin, or we could add an 'name' member to factgroup, so they may identify themselves.

DonLakeFlyer:
Yeah that's nasty. I still think merging them together makes sense. Just pass a QStringList of names through as well.

DonLakeFlyer:
Not sure what is going on with Travis here. Close/Reopen to get new rebase.

jaxxzer:
Thanks, did not know that trick!

DonLakeFlyer:
@jaxxzer Can you fix up the overrides in you pull as well? Not sure why this failure is only showing up now in this pull. Strange.

jaxxzer:
Ok, so I'm not crazy. Yes, I'll fix it up.

DonLakeFlyer:
@jaxxzer This needs the instance data cleaned up then it should be good to go.

jaxxzer:
@DonLakeFlyer Can you please clarify? I don't see where the problem is.

DonLakeFlyer:
Crap my comments have been sitting in a stupid pending review things for days now!

jaxxzer:
No problem @DonLakeFlyer. I've applied the changes, I hope it's in line with what you had in mind.

DonLakeFlyer:
@jaxxzer Can you rebase? There seems to be more stuff in here than there should be.

jaxxzer:
@DonLakeFlyer Done. I'm not sure what you mean more stuff than there should be. I don't think the deltas changed after rebasing.

PR #5130, opened 2017-05-10
Bugfix for vehicle health instrument view

jaxxzer:
The view was switching to the camera controls view.


jaxxzer:
To reproduce, bang your autopilot against a desk.

DonLakeFlyer:
Funny this is that Gus literally asked me yesterday how to simulate a vehicle health failure. I didn't think of your repro steps :)

PR #5133, opened 2017-05-11
ULog Geotagging Support v2

mhkabir:
@LorenzMeier @DonLakeFlyer I removed the major Survey command changes, and added a tiny mod to allow retraction of camera lenses when mission is complete. It will not affect any existing setups, but still allow fully featured support on the autopilot side till we can phase in the new commands.

Open concerns : 
1. With default Survey settings, we should be capturing images even on turnarounds to optimize 3D reconstruction. This is currently not the case, and I don't see a way to change it from the UI.
2. The camera is not enabled till the 1st survey waypoint. For cameras supporting power control, this means ~3s for them to bring the lens out and start shooting. Therefore, 3-4 triggers are missed at the beginning. (e.g on our QX1 + MAP2 reference setup)
3. The PR doesn't build in one CI instance due to init issues for the structs. @LorenzMeier has a solution for it I think.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -450,6 +450,8 @@ DebugBuild { PX4FirmwarePlugin { PX4FirmwarePluginFactory  { APMFirmwarePlugin {
 
 HEADERS += \
     src/AnalyzeView/ExifParser.h \
+    src/AnalyzeView/ULogParser.h \
+    src/AnalyzeView/PX4LogParser.h \
     src/CmdLineOptParser.h \
     src/FirmwarePlugin/PX4/px4_custom_mode.h \
     src/FlightDisplay/VideoManager.h \
@@ -635,6 +637,8 @@ AndroidBuild {
 
 SOURCES += \
     src/AnalyzeView/ExifParser.cc \
+    src/AnalyzeView/ULogParser.cc \
+    src/AnalyzeView/PX4LogParser.cc \
     src/CmdLineOptParser.cc \
     src/FlightDisplay/VideoManager.cc \
     src/FlightMap/Widgets/ValuesWidgetController.cc \
--- a/src/AnalyzeView/ExifParser.cc
+++ b/src/AnalyzeView/ExifParser.cc
@@ -56,7 +56,7 @@ double ExifParser::readTime(QByteArray& buf)
     return tagTime.toMSecsSinceEpoch()/1000.0;
 }
 
-bool ExifParser::write(QByteArray &buf, QGeoCoordinate coordinate)
+bool ExifParser::write(QByteArray& buf, GeoTagWorker::cameraFeedbackPacket& geotag)
 {
     QByteArray app1Header("\xff\xe1", 2);
     uint32_t app1HeaderInd = buf.indexOf(app1Header);
@@ -141,7 +141,7 @@ bool ExifParser::write(QByteArray &buf, QGeoCoordinate coordinate)
     gpsData.readable.fields.gpsLatRef.tagID = 1;
     gpsData.readable.fields.gpsLatRef.type = 2;
     gpsData.readable.fields.gpsLatRef.size = 2;
-    gpsData.readable.fields.gpsLatRef.content = coordinate.latitude() > 0 ? 'N' : 'S';
+    gpsData.readable.fields.gpsLatRef.content = geotag.latitude > 0 ? 'N' : 'S';
 
     gpsData.readable.fields.gpsLat.tagID = 2;
     gpsData.readable.fields.gpsLat.type = 5;
@@ -151,7 +151,7 @@ bool ExifParser::write(QByteArray &buf, QGeoCoordinate coordinate)
     gpsData.readable.fields.gpsLonRef.tagID = 3;
     gpsData.readable.fields.gpsLonRef.type = 2;
     gpsData.readable.fields.gpsLonRef.size = 2;
-    gpsData.readable.fields.gpsLonRef.content = coordinate.longitude() > 0 ? 'E' : 'W';
+    gpsData.readable.fields.gpsLonRef.content = geotag.longitude > 0 ? 'E' : 'W';
 
     gpsData.readable.fields.gpsLon.tagID = 4;
     gpsData.readable.fields.gpsLon.type = 5;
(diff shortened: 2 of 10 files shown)


src/MissionManager/SurveyMissionItem.cc
```diff
@@ -1062,7 +1062,7 @@ bool SurveyMissionItem::_appendMissionItemsWorker(QList<MissionItem*>& items, QO
         MissionItem* item = new MissionItem(seqNum++,
```
mhkabir: @DonLakeFlyer We never seem to hit this point after flying the Survey (even if there is no refly). Some sort of a logical issue somewhere. Can you have a look?

mhkabir: Ah, it's `_imagesEverywhere`. Since it is not set, we never reach execute the inner statement. I'm not sure why we need to check that at the end of the survey anyway. Can we get rid of the check?

Also, related - my open concern No. 1 in the PR's opening post.

src/MissionManager/SurveyMissionItem.cc
```diff
@@ -1062,7 +1062,7 @@ bool SurveyMissionItem::_appendMissionItemsWorker(QList<MissionItem*>& items, QO
...
                                             MAV_CMD_DO_SET_CAM_TRIGG_DIST,
                                             MAV_FRAME_MISSION,
-                                            0.0,                    // trigger distance (off)
+                                            -1.0,                   // trigger distance (off)
```
LorenzMeier: Any concerns with leaving this at 0?

DonLakeFlyer: No idea if -1 works with ArduPilot. I know and have tested 0 to work with ArduPilot.

DonLakeFlyer:
> With default Survey settings, we should be capturing images even on turnarounds to optimize 3D reconstruction.

If you talk with folks doing Surveys they will tell you that is incorrect and they do no want images from the turns.

src/MissionManager/SurveyMissionItem.cc
DonLakeFlyer: Refly does not actually "refly" the survey. It just generates additional paths on the single survey.

DonLakeFlyer:
> I removed the major Survey command changes, and added a tiny mod to allow retraction of camera lenses when mission is complete.

I don't see a code change for this?

DonLakeFlyer:
> The camera is not enabled till the 1st survey waypoint. For cameras supporting power control, this means ~3s for them to bring the lens out and start shooting.

What command does that? It would be needed to be added to Mission Settings if there is a Survey. That said, this whole thing of automatically adding things to Mission Settings to prep the camera for taking phots is starting to get very complicated and very vehicle specific. I'm concerned about adding more there.

src/MissionManager/SurveyMissionItem.cc
mhkabir: @LorenzMeier This is what turns the camera off for PX4. It doesn't break older PX4 either.

mhkabir:
> If you talk with folks doing Surveys they will tell you that is incorrect and they do no want images from the turns.

Can you put me in touch with some of them? :) I'm curious as to why they wouldn't want it for 3D reconstruction. The feature matching would just perform better from the theoretical perspective.

mhkabir:
> I don't see a code change for this?

https://github.com/mavlink/qgroundcontrol/pull/5133/files#diff-147f32bf78728184cfa84bd1d0994dd4R1065

mhkabir:
> What command does that?

As currently implemented, `DO_SET_TRIGG_DISTANCE`.

src/MissionManager/SurveyMissionItem.cc
mhkabir: My point is, can we remove the `_imagesEverywhere` check so that the camera actually turns off at the end of any survey?

src/MissionManager/SurveyMissionItem.cc
DonLakeFlyer: Is there some problem with 0?

DonLakeFlyer:
> I'm curious as to why they wouldn't want it for 3D reconstruction. The feature matching would just perform better from the theoretical perspective.

There are many reasons. Some of them:
* While the vehicle is in the turn it is unstable and the image quality is low, even on a multi-rotor. Hence it makes the mathcing worse.
* Also keep in mind that most surveys are done from fixed wing (with no gimbal), where the images from most of the the turn are useless. 
* Also fixed wing use large turnarounds outside of the survey area so the vehicle is rock stable once it ends the survey area. That's a lot of unimportant images taken.
* For larger surveys storage space can be critical hence limiting the photos to only absolutely what is need is important.

DonLakeFlyer:
> As currently implemented, DO_SET_TRIGG_DISTANCE.

You mean the -1 as opposed to using 0?

mhkabir:
> You mean the -1 as opposed to using 0?

No! When I say "as implemented", I mean as on current QGC master. It is the first DO_SET_TRIGG_DISTANCE call with distance > 0 at the time of entering the Survey. I have not changed this in any manner.

src/MissionManager/SurveyMissionItem.cc
mhkabir: Yes - with 0, for cameras with power control (on PX4) it turns the camera off (lens closes, camera powers down). We don't want that to happen during the turnarounds. We want the camera to be powered, but not taking photos.

DonLakeFlyer:
There is no other way to tell the camera to power up and be ready to take photos?

mhkabir:
> There is no other way to tell the camera to power up and be ready to take photos?

There is.  The `DO_TRIGGER_CONTROL` command (which was implemented in my previous branch) was for this purpose - to change the trigger/camera state without affecting other things. If you send it with param1 = 1.0 (enable trigger and power) and param3 = 1.0 (pause shooting), it will tell the camera to power on and be ready to take shots. Then the first Survey waypoint would un-pause, and off we go.

DonLakeFlyer:
I'm having a really hard time understanding whether there is an existing problem with QGC survey code. Or there is only a problem because the PX4 camera support has changed from how it previously worked and has now broken QGC. Or possibly new features are being added to old commands which now cause new problems in QGC which previously did not exist.

As I have said before QGC is in a lockdown phase for a stable release (as is PX4 stack) as are other companies who use QGC for their vehicles. Survey has been tested on multiple commerical vehicles with cameras using PX4 stack and has been verified to work correctly. Hence any changes in there are going to get massive scrutiny. Yet somehow something is suddenly wrong. But as far as I can tell from hours of actual vehicle testing it all works. I'm not going to redo all of that testing including telling other corporations that they will need to retest Survey as well because Survey generation changed so late in the game.

Point me at a critical bug in the existing Survey code and we can discuss how to fix it. But don't point me at a problem in the existing Survey code which is caused by the fact that the firmware changed out from under QGroundControl. Or some new feature is wanted that didn't exist before. It's just too late for that. I'll be happy to discuss those sorts of changes after QGC Stable 3.2 goes out the door. But not now.

DonLakeFlyer:
A very simple case of this: Looking through the ArduPilot code it specifically checks for 0 on TRIGG_DIST to not trigger the camera any longer. Hence the change to -1, just broke ArduPilot survey support.

DonLakeFlyer:
In realtionship to my explanation above. Is the concept of powered cameras turning on and off based on TRIGG_DIST values something new? Or would that have been a problem with previous versions of firmware as well as it relates to TRIGG_DIST=0 functionality turning the camera off?

DonLakeFlyer:
FYI on CI break: The new C++ initializer syntax ```{ }`` generally is a royal pain to get to work cross platform. The easiest thing is to fallback to old-school and just use memset.

mhkabir:
> A very simple case of this: Looking through the ArduPilot code it specifically checks for 0 on TRIGG_DIST to not trigger the camera any longer. Hence the change to -1, just broke ArduPilot survey support.

Ok, I'm reverting. I don't see a way to fix this for now without making a breaking API change. Let's talk about it after v3.2 goes out of the door.

mhkabir:
@LorenzMeier, the the compromise is that in the end - for cameras with power control we will leave power control unused and they will remain powered on throughout the flight and during landing.

In case we weren't clear yet on the problem I had been trying to solve - if someone was using e.g PX4 v1.5.5 with a QX1 + MAP2, the DO_SET_TRIGG_DIST with distance=0 sent from QGC would cause the camera to power down. No one caught it because the power control implementation had been broken on the PX4 side, which I fixed - exposing this bug.

mhkabir:
One last thing - I understand that Survey is in freeze for now, but is there a way to change the `_imagesEverywhere` flag anywhere from the UI? Several bits of code never get hit because that flag is never set.

DonLakeFlyer:
The _imagesEverywhere flag is in there because I'm not sure if I will expose the option for it yet. Old versions of QGC took images throughout the survey. Then from user feedback I changed to only taking images within the survey area. But I specifically did not add a UI option for images everywhere because the survey UI was busy enough already. if I get user feedback that it is still needed ti wil be ready for use. If not, I will remove at some point later down the road.

Also @mhkabir: In one of your pulls you had changed the delay which happens after the image is taken during hover and capture from 0.5 to 1 second. Is there a camera reason to do that. 0.5 seconds is too slow for some cameras to do be done taking the image for example?

DonLakeFlyer:
Pull still needs the changes for cross-compile. Failing Linux builds.

DonLakeFlyer:
All good now, thanks.

PR #5138, opened 2017-05-12
Set flying flag regardless of armed status

jaxxzer:
This is a bug. Why not be able to change the flying flag to false when we are disarmed?

I think the intention was to not set the flag to true when we are disarmed, but that is [handled by the caller](https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc#L437).


DonLakeFlyer:
Yeah, strange.

PR #5143, opened 2017-05-13
USB modems

AndKe:


DonLakeFlyer:
@AndKe Could you try a couple things with these radios first? I'm wondering why the fallback checks aren't working. Spefically run with LinkManagerLogVerbose and then plug each one in. You should see detailed port information go by for each. Then could you past that here. Thanks.

AndKe:
Well, the fallback isn't working because the FTDI adapter's string is
"Future Technology Devices International, Ltd FT232 USB-Serial (UART) IC"
and the CP210 is:
"Cygnal Integrated Products, Inc. CP210x UART Bridge / myAVR mySmartUSB light"

:)

DonLakeFlyer:
Ah, ok. Thanks for checking. And thanks for the pull.

AndKe:
I am happy to help with trivial things I can handle :)

PR #5144, opened 2017-05-13
Initial localization work.

dogmaphobic:
This is an initial, token start of localizing QGC.

I've added a script that collects all the strings that are set to be localized and creates the localization source file to be sent to crowding for translation. This is just the beginning of the work. There is a great deal of work to be done before all strings used throughout QGC are ready for exporting by that script. 

Localization is a complex task. It's not just a matter of translating every instance of a string, but things need to be in context. In addition, strings that are built from parts or used for comparison, need special handling as well and this is the case in several places throughout the code.

For instance, take this [example](https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/APM/APMPowerComponentController.cc#L106-L109):

```
if (text == "Connect battery now") {
      emit connectBattery();
      return;
}
```

Whatever __*Connect battery now*__ gets translated to, it has to match exactly whatever that source is otherwise the code above won't work. This should not be comparing a string but rather a constant that corresponds to the string.

String constants also don't work directly. There are various [instances](https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/APM/APMCompassCal.cc#L385-L398) of them as well:

```
    static const char* rgOrientationStrs[] = {
        "back",		// tail down
        "front",	// nose down
        "left",
        "right",
        "up",		// upside-down
        "down",		// right-side up
        "error"
    };
```

Here is a starting point for what needs to be done:

http://doc.qt.io/qt-5/i18n-source-translation.html
http://doc.qt.io/qt-5/qtlinguist-index.html

In summary, this gets the ball rolling but there is a tremendous amount of work to be done before QGC can be properly localized. The code was not built with localization in mind and retroactively making it so will not be an easy task.

The localization source file included in this PR (qgc.ts) is ready to start. It can already be sent to crowdin as the `lupdate` tool is smart enough to append new strings added to the code. With that said, it would be best to minimize drastic changes to existing strings already set for translation.

The actual code that loads the *compiled*, translated files will come in a new PR after this one.

### DIFF
--- a/src/QGCQFileDialog.h
+++ b/src/QGCQFileDialog.h
@@ -15,6 +15,7 @@
 #error Should not be included in mobile builds
 #endif
 
+#include <QCoreApplication>
 #include <QFileDialog>
 
 /// @file
@@ -40,7 +41,7 @@
 */
 
 class QGCQFileDialog : public QFileDialog {
-    
+    Q_DECLARE_TR_FUNCTIONS(QGCQFileDialog)
 public:
 
     //! Static helper that will return an existing directory selected by the user.
--- a/localization/README.md
+++ b/localization/README.md
@@ -1,0 +1,25 @@
+To generate (or update) a source translation file, use the "lupdate" command line tool. It can be found in the Qt distribution.
+
+(Easiest to just run the `gen_translation_source.sh` script, or get the commands from within the script and run them manually)
+
+This will parse all the source files and generate a language translation file called qgc.ts. This file should be copied into whatever language it is to be translated to and sent to translation. The translation can either be done directly within the (XML) file, using Qt's "Linguist" tool, or uploaded to crowdin.
+
+For instance, to localize to Germany German as an example you first copy the original source:
+```
+cp qgc.ts qgc_de-DE.ts
+```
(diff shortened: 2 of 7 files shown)


DonLakeFlyer:
All of the above examples are actually the worst examples for fix. The reason is those strings are being compared to strings which come from the vehicle. How you fix that can get crazy. Translate teh vehicle side as well. To what language? What if the vehicle language doesn't match the qgc language. Yuck.

src/QGCQFileDialog.h
```diff
@@ -40,7 +41,7 @@
 */
 
 class QGCQFileDialog : public QFileDialog {
-    
+    Q_DECLARE_TR_FUNCTIONS(QGCQFileDialog)
```
DonLakeFlyer: Ah, had never heard of that before. I wonder where else this is a problem. Those may be harder to find. Does the Qt stuff spit out and error for that somehow?

dogmaphobic:
> Does the Qt stuff spit out and error for that somehow?

Yes, the `lupdate` tool will detect and let you know. That's how I found the problem to begin with.

PR #5146, opened 2017-05-13
Pull firmware custom version from PX4

DonLakeFlyer:


DonLakeFlyer:
@dogmaphobic Do you know what the compiler magic is to get Linux to shut up on the travis failure?

PR #5147, opened 2017-05-13
[WIP] Initial PX4 GeoFence implementation [DO NOT MERGE, for testing only]

DonLakeFlyer:
This pull is meant to be tested with this PX4 Firmware pull: https://github.com/PX4/Firmware/pull/6888.

* It supports a single inclusion GeoFence.
* QGC will request the GeoFence at vehicle connect
* If you create a GeoFence from Plan and Upload, QGC should theoretically be sending the right stuff. Although I can't get far into the process currently.
* Does not currently test capability bit, assumes to be running against firmware that supports new protocol.

@bkueng If you run this again your pull. When you connect to a vehicle, you will see an error pop up which says there was a mission_type mismatch. This is because when QGC asks for the fence from the vehicle. The vehicle responsds with a MISSION_COUNT which has type set to MAV_MISSION_TYPE_MISSION instead of MAV_MISSION_TYPE_FENCE. Similar things happens when you upload a fence, the response from the vehicle doesn't seem to be typed correctly. As far as I can tell QGC is setting the right types from it's side.

You can see in detail what is going on inside QGC if you turn on all the logging assocaited with a Plan: 
*GeoFenceControllerLog
* GeoFenceManagerLog
* MissionControllerLog
* MissionManagerLog
* PlanManagerLog
* PlanMasterControllerLog
* RallyPointControllerLog
* RallyPointManagerLog

FYI: The difference between a Controller and a Manager is that a Manager handles the protocol with the vehicle and knows nothing about the ui. The Controller connects the data from the Manager to the user interface.


DonLakeFlyer:
Wait, I think it's not working because I didn't switch to link the mavlink 2.

DonLakeFlyer:
Yup. That gets me a little further. I can now connect to a vehicle with no polygons on it and it works. Working on upload.

DonLakeFlyer:
I add MockLink support for geofence. With that I can upload/download polygons correctly. There are still signalling problems so the Fly view won't show correctly and the sync progress indicators are out of whack. But it's enough to hammer a bit on the firmware side.

Now when I try to upload a fence I get a MAV_MISSION_ERROR on sending the first vertex.

DonLakeFlyer:
Nope. My screw up. Still working on it.

DonLakeFlyer:
Ok. Load is working now. I can round trip a polygon through MockLink and the Fly view is working as well.

LorenzMeier:
@bkueng PR for you here which should allow to wrap-up the implementation.

DonLakeFlyer:
This still has a ways to go on the QGC side:
* Multiple polygon support
* Exclusion versus/inclusion
* Rally Points

src/MissionManager/PlanManager.cc
```diff
@@ -0,0 +1,945 @@
...
+    
+    if (!_checkForExpectedAck(AckMissionRequest)) {
+        return;
+    }
+    
+    mavlink_msg_mission_request_decode(&message, &missionRequest);
+    qCDebug(PlanManagerLog) << "_handleMissionRequest sequenceNumber" << missionRequest.seq;
+
+    if (missionRequest.mission_type != _planType) {
+        _sendError(MissionTypeMismatch, QString("_handleMissionCount Incorrect mission_type received expected:actual %1:%2").arg(_planType).arg(missionRequest.mission_type));
```
bkueng: This should be: `_handleMissionRequest` Incorrect mission_type...

src/MissionManager/PlanManager.cc
```diff
@@ -0,0 +1,945 @@
...
+    // We can get a MISSION_ACK with an error at any time, so if the Acks don't match it is not
+    // a protocol sequence error. Call _checkForExpectedAck with _retryAck so it will succeed no
+    // matter what.
+    if (!_checkForExpectedAck(_expectedAck)) {
+        return;
+    }
+    
+    mavlink_msg_mission_ack_decode(&message, &missionAck);
+    if (missionAck.mission_type != _planType) {
+        _sendError(MissionTypeMismatch, QString("_handleMissionCount Incorrect mission_type received expected:actual %1:%2").arg(_planType).arg(missionAck.mission_type));
```
bkueng: This should be: `_handleMissionAck` Incorrect mission_type...

DonLakeFlyer:
> does QGC make sure there's only 1 transaction in progress at a time?

Yes it does.

DonLakeFlyer:
* Now includes support for adding/displaying multiple inclusion/exclusion polygons.
* No support for removing polygons after you add them
* After you add a polygon make sure to modify it or move it since new ones will be added at exactly the same size/place
* Polygon editing visuals are on for all polygons which makes things get crazy real fast

Still much work to do on ui editing model, but enough to test more deeply. All tested with MockLink round-tripping.

Like I said it gets cray fast. Two inclusions and one exclusion:
screen shot 2017-05-15 at 3 02 20 pm

bkueng:
Nice! I tested the inclusion & exclusion polygons and an exclusion circle in SITL, which all work.
However changing the `Breach Action` does not do anything. Is it supposed to change the `GF_ACTION` param? Changing this param manually to e.g. RTL works though.

DonLakeFlyer:
The parameter settings side of this is kind of in flux. Not quite sure where to put that. For now, change fence params from parameter editor or Safety page.

What would be really nice is way to not use any parameters for any of the fence/ rally stuff. Any touch of parameters makes it not mavlink generic.

bkueng:
Did you already implement the safe/rally points? I'm wondering if we really need the `MAV_CMD_FENCE_RETURN_POINT`? It could just use the rally points.

DonLakeFlyer:
> Did you already implement the safe/rally points? I'm wondering if we really need the MAV_CMD_FENCE_RETURN_POINT? It could just use the rally points.

Rally is NYI on QGC side. I should get to it in a day or so. I agree that with rally points you don't need breach return. I think the issue may be that capability bits for fence and rally are separate? Can't remember. If so, then that would be a problem. Maybe just once capability bit, both fence and rally?

DonLakeFlyer:
* Working towards better editing ui for fence
* You can drag circular areas to move them
* You can now delete fences after adding
* You can switch a fence from inclusion to exclusion
* Exclusion visuals for polygon

screen shot 2017-05-22 at 6 14 58 pm

bkueng:
> I agree that with rally points you don't need breach return. I think the issue may be that capability bits for fence and rally are separate? Can't remember. If so, then that would be a problem. Maybe just once capability bit, both fence and rally?

They're currently separate capability bits, but I agree it would make sense to merge them. Alternatively Lorenz and I were discussing that we can use the home position as the first rally point (which would happen completely internal, as it does not need to be transmitted). This way, even if you don't have any rally points, there's still a valid RTL location for geofence violations.
An idea is also to assign priorities to each rally point, so that we can realize an alternative landing position, by using a rally point with a higher prio than the home position. Or just go to the closest if all have the same priority.

Then we should also agree on terminology: internally in FW we use safe points, which is more understandable to me than rally points. If rally point is already the generally accepted term, I can go with that as well.

hamishwillee:
> Then we should also agree on terminology: internally in FW we use safe points, which is more understandable to me than rally points. If rally point is already the generally accepted term, I can go with that as well.

Safe points sounds much better to me, but I do know in the wider community including Ardupilot rally point is generally accepted. I'd be trying to work with Ardupilot to get the naming convention changed.

DonLakeFlyer:
> Alternatively Lorenz and I were discussing that we can use the home position as the first rally point (which would happen completely internal, as it does not need to be transmitted). This way, even if you don't have any rally points, there's still a valid RTL location for geofence violations.

But you need an altitude for a rally point for fixed wing. FW traditionally goes to the rally point and loiters waiting for you to take over.

DonLakeFlyer:
Like Hamish says, rally point terminology has been in ArduCopter terminology for a while now.

LorenzMeier:
@DonLakeFlyer To not get ourselves stuck in whatever got arbitrarily chosen at some point we will need to continue to evolve terminology. A significant installed user base (as in DJI's) is a good indicator that we want to keep a term. In this particular case here however, when we cross from DYI space over to consumer space we need to make sure we use the right terminology for consumers. We do significantly detract from the UX if we're not self-conscious on that front.

hamishwillee:
QGC supports px4 and ardupilot. If we called them safe points how would this affect the QGC UI and the documentation. Would they use the same MAVLink messages for setting/querying? If this can all be done fairly transparently to users then I'm with Lorenz - lets call them safe points. If they are part of the common MAVLink messages ardupilot will probably be happy enough to support both.

DonLakeFlyer:
It doesn't really matter what mavlink calls them. That is all internal. QGC can show whatever words you want for them.

On May 23, 2017, at 4:21 PM, Hamish Willee <notifications@github.com<mailto:notifications@github.com>> wrote:


QGC supports px4 and ardupilot. If we called them safe points how would this affect the QGC UI and the documentation. Would they use the same MAVLink messages for setting/querying? If this can all be done fairly transparently to users then I'm with Lorenz - lets call them safe points. If they are part of the common MAVLink messages ardupilot will probably be happy enough to support both.

hamishwillee:
>  QGC can show whatever words you want for them.

Also while [we have docs here](https://docs.qgroundcontrol.com/en/PlanView/PlanRallyPoints.html) that say rally point, it would be trivial to name as Safe/Rally Points and add a line inside explaining that they are referred to differently by ArduPilot.

DonLakeFlyer:
Updating editing interface for fences:
screen shot 2017-05-25 at 6 20 43 pm

bkueng:
@DonLakeFlyer I pushed 2 fixes, have a look at the commits.

bkueng:
@DonLakeFlyer I found a regression here: the currently acctive waypoint is not updated during a mission.

DonLakeFlyer:
@bkueng This is replaced by the other branch now right?

bkueng:
No, except if you have a branch that I'm not aware of. Can you merge master into this to bring it up-to-date?

DonLakeFlyer:
Isn't the GeoFenceAirMap branch in my repro this branch plus to AirMap branch? So that's the branch to work on?

bkueng:
Yes that's true. But for now I'd like to keep things separate, with this workflow: do Geofence fixes/extensions/UI stuff on this branch and keep Airmap rebased on this. Then we can merge the Geofence more quickly, while at the same time continuing independently on Airmap. Does this make sense or is it too complicated?

Another thing: we need a way to distinguish user from airmap polygons when downloading from the vehicle. Should we add an identifier to each polygon? What do you suggest?

DonLakeFlyer:
> Does this make sense or is it too complicated?

Makes sense.

> we need a way to distinguish user from airmap polygons when downloading from the vehicle. Should we add an identifier to each polygon? What do you suggest?

I need to put more thought into the user model for airmap polygons. It throws a bit of a wrench in the geofence only being part of a mission concept.

LorenzMeier:
@bkueng Can you rebase this?

bkueng:
merged master (was much easier). There were some conflicts: `APMRalllyPointManager` got deleted in this branch and modified on master.
Tested, geofence & mission upload still work.

LorenzMeier:
qgc_airmap_polygons_cluttered

mrpollo:
@DonLakeFlyer do you think you can attend the [dev call this coming Wednesday (July 19th)](http://discuss.px4.io/t/weekly-dev-call-of-july-19-2017/3917)?

Issue #5127, opened 2017-05-10
Guided Takeoff ArduPilot: Should use PILOT_TKOFF_ALT for takeoff altitude


DonLakeFlyer:
Related to #5063

DonLakeFlyer:
PILOT_TKOFF_ALT seems to default to 0 unfortunately. So if it =0 or the parameter is missing, then I called back to 2.5 meters default. Otherwise the parameter values is used.

Issue #5062, opened 2017-04-26
Altitude change pop-up very short range and heading random change


lvale:
Latest QGC master, latest ArduPilot Copter master

Clicking on the Action to change altitude the elevator on the right only allows very small increments (10m ?)

<img width="1261" alt="screen shot 2017-04-26 at 03 13 05" src="https://cloud.githubusercontent.com/assets/3130363/25415319/d9cd2a38-2a2e-11e7-8857-33036710a263.png">
<img width="976" alt="screen shot 2017-04-26 at 03 12 24" src="https://cloud.githubusercontent.com/assets/3130363/25415320/d9efd272-2a2e-11e7-9111-e64a09008afa.png">

Also changing altitude is changing also the heading for some undetermined heading

DonLakeFlyer:
Willing to consider new range. Any opinions?

lvale:
I would go for a log slider, meter by meter on the first half, then tens of meters, and maximum range below 400ft (120m), and/or the possibility of inputting a number.

On a usability, and I know it's not easy, we have to move from the extreme left to the bottom middle to the extreme right of the screen for a single process. Couldn't we have these elements to be closer from one another ?

DonLakeFlyer:
Log slider is way too much work for now. Also don't think I'll go back to allowing numeric entry. Just trying to close things out with major usability problems.

pritamghanghas:
heading change is Ardupilot problem. It doesn't honor SET_POSITION_TARGET_GLOBAL_INT yaw parameter. There are several holes in arducopter mavlink implimentation. It is like anytime I use a command something is not implimented especailly when you have anything to do with yaw. Another day I was trying to specify rate of yaw change in a yaw change command, arducopter ignores that as well.

amilcarlucas:
@pritamghanghas I did a couple of PRs for missing YAW implementations in Arducopter. Those PRs did not get accepted in the master branch, but if you apply them localy it might help you.

pritamghanghas:
@amilcarlucas thank you for that. I generally stay away from touching flight code. But it is a pity that those didn't get accepted. QGC can't really use something that is not in mainline. Though I can use them personally.

DonLakeFlyer:
I'm going to stay away from guided heading change for now. Depending on vehicle settings you can sort of achieve it with goto location. May consider at some point.

PR #5150, opened 2017-05-14
Cubic altitude slider

DonLakeFlyer:
Problems:
* The range of the altitude slider for multi-rotor was too small (10m up and down from current altitude)
* Using a large linear range slider on a small screens creates large jumps in altitude for small movement of the slider
* Multi-Rotor can tend to live closer to the ground more safely than a fixed-wing from a standpoint of minimum altitude
* Third parties want to set min/max for altitude slider for their custom builds

Solutions:
* Use a cubic range on the slider. This allows for fine adjustment around the current altitude, but also allows the upper/lower end of the range to hit a full range adjustment. Thanks to @lvale for the suggestion. I tried a number of exponent and cubic seems to provide the best adjustment.
* Have separate min/max range for multi-rotor and fixed wing. (Multi-Rotor: 2m<->400ft, Fixed Wing: 10m<->400 ft)
* Store min/max alt values as settings so third party core plugin can override.

Fixes for #5127, #5062

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -517,6 +517,7 @@ HEADERS += \
     src/Settings/AppSettings.h \
     src/Settings/AutoConnectSettings.h \
     src/Settings/FlightMapSettings.h \
+    src/Settings/GuidedSettings.h \
     src/Settings/RTKSettings.h \
     src/Settings/SettingsGroup.h \
     src/Settings/SettingsManager.h \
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
This pull also has a fix for #5127. Which sets the guided takeoff altitude to PILOT_TAKEOFF_ALT so it is configurable.

PR #5167, opened 2017-05-17
Allow (mobile) plugins to define telemetry logs.

dogmaphobic:


src/QGCApplication.cc
```diff
@@ -503,31 +503,39 @@ void QGCApplication::criticalMessageBoxOnMainThread(const QString& title, const
...
-        QString dtFormat("yyyy-MM-dd hh-mm-ss");
-
-        int tryIndex = 1;
-        QString saveFileName = nameFormat.arg(QDateTime::currentDateTime().toString(dtFormat)).arg("");
-        while (saveDir.exists(saveFileName)) {
-            saveFileName = nameFormat.arg(QDateTime::currentDateTime().toString(dtFormat)).arg(QStringLiteral(".%1").arg(tryIndex++));
-        }
-        QString saveFilePath = saveDir.absoluteFilePath(saveFileName);
+    // Are we supposed to log?
+    if(toolbox()->settingsManager()->appSettings()->telemetrySave()->rawValue().toBool()) {
```
DonLakeFlyer: You shouldn't need this. You should never get here if that isn't set.

DonLakeFlyer: The only case would be if you crash and there is a leftover log file. But if that happens then save was on at the time you crashed. Hence you would want to save that anyway.

DonLakeFlyer: Ah, I hate the stupid review thing...

dogmaphobic:
I don't know what I did but your notes, along with my answers disappeared. I guess I see why you hate these code reviews :)
Either way, yes. I had added that before understanding the *Always Write To Temp* deal. I've removed it.

PR #5169, opened 2017-05-17
Vagrant qt5.7 5064

### LINKED ISSUE #5064: Vagrant using Qt 5.5
The README.md references QT 5.7 is required, but `.vagrantconfig.yml` references 5.5.

Looks like `.vagrantconfig.yml` needs to be updated.

double16:
Fixes #5064


double16:
I get a clean build including the Linux deploy package now.

I found this note at the top of `'.travis.yml`:
```shell
# if you update this file, please consider updating Vagrantfile too
```
😂

DonLakeFlyer:
@dagar Do you know anything about vagrant to review this?

hamishwillee:
There is a new requirement for rsync to be installed. Can this be avoided - rsync on Windows in a pain in the bum:
```
==> default: Syncing folders to the host VM...
"rsync" was not detected as installed in your guest machine. This
is required for rsync synced folders to work. In addition to this,
Vagrant doesn't know how to automatically install rsync for your
machine, so you must do this manually.
```

hamishwillee:
It would also be awesome if we could somehow have a shared yml file for the dependencies with travis.

double16:
There is nothing in the Vagranfile changes that would add a dependency on rsync. I suspect this is due to your Vagrant install. rsync is the fallback when VM provider specifics don't work.

Which version of Vagrant do you have? Are you using VirtualBox, VMware? If VirtualBox you should install the 'vagrant-vbguest' plugin to update the VirtualBox drivers inside the VM so that shared folders work. VMware has a plugin as well, but it must be purchased.

double16:
Which dependencies in travis do you want to automatically pull in? `.travis.yml` is a YAML file and we can parse it in the Vagrantfile. For apt dependencies, it looks like that can easily be determined. Qt might be a bit more difficult, probably need to use some regex on `.travis.yml`.

hamishwillee:
> Which dependencies in travis do you want to automatically pull in? .travis.yml is a YAML file and we can parse it in the Vagrantfile. For apt dependencies, it looks like that can easily be determined. Qt might be a bit more difficult, probably need to use some regex on .travis.yml.

The Qt ones :-) What I am saying is that ideally the build dependencies should be in a common file so that any changes to the travis build are also used in the Vagrant build.

DonLakeFlyer:
Unfortunately that won't work because the Qt that travis uses is a stripped down version of the install to reduce build times. Whereas I think vagrant wants the full Qt install right?

hamishwillee:
> Unfortunately that won't work because the Qt that travis uses is a stripped down version of the install to reduce build times. Whereas I think vagrant wants the full Qt install right?

Well, it needs enough of Qt to run the UI. So probably :-(

hamishwillee:
OK, so I installed the latest vagrant and virtualbox. The problem appears to be that I don't have docker installed (didn't realise it was a dependency). The docker toolbox gives me errors and my computer won't run docker for windows (it is home edition). UPshot - I don't think I can test this further. I had hoped for a build that didn't have any dependencies for the user other than vagrant :-(

double16:
Docker is chosen because of vagrant's logic to determine a default provider. Should prefer VM based boxes over docker now.

double16:
I'm thinking we can make the `.travis.yml` file the common source of dependencies and the Vagrantfile can read it. My latest push shows the apt packages being pulled from `.travis.yml` and also the global environment.

What do you think about setting global environment variables for the QT version and download URLs? Perhaps move all of the (applicable) values from `.vagrantconfig.yml` into `.travis.yml`?

As for the minimal Qt version, please correct me, but it looks like the Linux deployable package is built using that. When installing the package, would Ubuntu then install additional Qt packages from using apt-get?

Vagrantfile
```diff
@@ -6,10 +6,15 @@
...
 
 current_dir    = File.dirname(File.expand_path(__FILE__))
-configfile        = YAML.load_file("#{current_dir}/.vagrantconfig.yml")
-yaml_config = configfile['configs']['dev']
+configfile     = YAML.load_file("#{current_dir}/.vagrantconfig.yml")
+travisfile     = YAML.load_file("#{current_dir}/.travis.yml")
+yaml_config    = configfile['configs']['dev']
 
 Vagrant.configure(2) do |config|
+  # This trick is used to prefer a VM box over docker
```
hamishwillee: This didn't work. I did a pull, and get the same error as before (on both vagrant destroy and up):
```
D:\Github\testdelete\qgroundcontrol>vagrant up
A Docker command executed by Vagrant didn't complete successfully!
The command run along with the output from the command is shown
below.

Command: ["docker", "ps", "-a", "-q", "--no-trunc", {:notify=>[:stdout, :stderr]}]

Stderr: error during connect: Get http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.28/containers/json?all=1: open //./pipe/docker_engine: The system cannot find the file specified. In the default dae
 be run elevated to connect. This error may also indicate that the docker daemon is not running.


Stdout:
```

hamishwillee:
> What do you think about setting global environment variables for the QT version and download URLs? Perhaps move all of the (applicable) values from .vagrantconfig.yml into .travis.yml?

Questions for @dagar :-)

double16:
Try removing your '.vagrant' directory in the source folder. It may be
recording that you've tried docker before. I don't have access to a Windows
machine to try this. Going by the docs and my Mac.

hamishwillee:
@double16 That worked thanks - build went through to completion. 

The docs indicate that you should be able to view and drive the instance through the VM (or the host if it is compatible - on windows obviously not). However this is UI-less VM I think so not particularly useful for me as a "try things out" environment.  Can we make the UI work too?

DonLakeFlyer:
Isn't vagrant meant to be used to set up a development environment or a user such that they can build the product and possible change/debug? If so, the min install of Qt which travis uses is a hacked version of Qt which has all sorts of things ripped out of it. It's just enough to drop a build from travis. For a development environment really want the full Qt don't you?

double16:
@DonLakeFlyer You're correct on the purpose of Vagrant. Who knows how to provide an updated Qt 5.7.1 download? How was the Qt 5.5 download created? If someone drops it in S3 I can change the Vagrantfile to use it.

DonLakeFlyer:
I'll make one for you to use.

DonLakeFlyer:
I have it built but S3 doesn't seem to want to upload anything right now.

DonLakeFlyer:
https://s3-us-west-2.amazonaws.com/qgroundcontrol/dependencies/Qt5.7.1-linux-vagrant.tar.bz2

hamishwillee:
It all builds OK. At the end I'm left with a prompt like this in VirtualBox. How do I log in and run the UI/QGC?

image

double16:
Restart the VM and you'll get a graphical login.

hamishwillee:
It works :-) From my perspective this is good to integrate. Thank you.

Some side questions for docs:
1. I assume that just using "vagrant" as a password  for the VM is correct?
2. Say I wanted to update the daily build to the latest head revision and rebuild within the same VM, what is the right process? E.g.
    - Do a git pull in windows ...
    - Open terminal in /home/vagrant/
    - Call what?

double16:
1. Yes, ‘vagrant’ as the password is a standard for Vagrant.

To rebuild with the latest code:
1. `git pull` on host (Windows, OS X, Linux, …)
2. `vagrant provision` will re-run the provisioning steps and they are idempotent, sort of :), you’ll get updating Ubuntu packages if available, but if not it doesn’t re-install them

OR

3. If you’re using the graphical login, build with Qt Creator. It’s installed and it did open the project for me. It didn’t seem to automatically detect a ‘desktop’ kit, but I’m not that familiar with it. (I stumbled through getting it to run on my Mac)

DonLakeFlyer:
Is this good to merge now?

double16:
Do we want to define the Qt version has a global environment variable in
travis.yml and pull that from the Vagrantfile? Should we add a build for
Vagrant using docker to Travis?

double16:
Maybe we should merge this as-is since it fixes something that is broken and I can submit another PR for the Travis build changes.

hamishwillee:
Merging as-is makes sense to me. The rest sounds like it may need some more thought.

dagar:
Looks good to me. Let's discuss possibly switching over travis-ci in another PR.

peterbarker:
Moving the Vagrant creation script out into a separate script would be nice at this point IMO.

If someone's working in there then they might like to look at that (otherwise I'll look at it next deps-change (unless someone beats me to that one too ;-) )

PR #5178, opened 2017-05-18
 -DoAdd Fall rate percent in Landing Pattern editor [DO NOT MERGE till 3.3]

patrickelectric:

### DIFF
--- a/src/MissionManager/FWLandingPattern.FactMetaData.json
+++ b/src/MissionManager/FWLandingPattern.FactMetaData.json
@@ -1,6 +1,6 @@
 [
 {
-    "name":             "Landing distance",
+    "name":             "Landing dist",
     "shortDescription": "Distance between landing and loiter points.",
     "type":             "double",
     "units":            "m",
@@ -42,5 +42,15 @@
     "units":            "m",
     "decimalPlaces":    1,
     "defaultValue":     0.0
+},
+{
+    "name":             "Fall rate",
+    "shortDescription": "Fall rate between landing and loiter altitude.",
+    "type":             "double",
+    "units":            "%",
+    "min":              0.1,
+    "max":              100,
+    "decimalPlaces":    1,
+    "defaultValue":     12.0
 }
 ]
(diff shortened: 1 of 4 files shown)


patrickelectric:
image

image

DonLakeFlyer:
@patrickelectric Can you make the fall rate just a non-modifiable value which is calculated? It's way too late in the game to be changing the internals for Landing pattern calculation now. We can do that once 3.2 goes out the door. But right now things are locked down too tight to make this sort of change.

Other option is to just hold on to this until 3.2 is done and then put it through.

DonLakeFlyer:
The more I think about it the more I think we just hold this till 3.2 goes out the door.

patrickelectric:
No problem

DonLakeFlyer:
Tested and works like a charm. Thanks @patrickelectric

DonLakeFlyer:
I screwed this up and closed without merging!

Antiheavy:
Is "Fall rate" the best term to use here?  "Glide ratio", "Landing slope", "Descent rate", or other variations might be more commonly used aircraft terminology.

DonLakeFlyer:
Descent rate seems good?

Kjkinney:
I think descent rate is the incorrect thing to use when planning a fixed wings landing path.

- Descent rate suggests that the aircraft is descending at a constant rate from the last waypoint to the land waypoint. This is false as both PX4 and Arduplane will change the descent rate to hit a glideslope.
- Using a percentage to describe a descent rate doesn't make sense.
- Descent rate is already used in unit/s to show how quickly the aircraft is descending at any given time. It is also used universally in the aviation community in unit/s to describe how quickly an aircraft will be ascending or descending



I suggest that this is switched to a glideslope in degrees.

- Glideslope/Glidepath is the universal aviation term when describing an airplane's path to a runway.
- Glideslope is also used in PX4 parameters to set the fixed wings path on landing, meaning that it will reject a mission if the glideslope is set too steep in px4. Arduplane uses glideslopes as well, however, I do not believe they are parameter defined. This would allow them to more accurately set a glideslope for landing rather than having to guess/do extra math.
- Glideslope would be defined in degrees to match aviation terms, as well as to match PX4 and Arduplane parameters and documentation.

Examples:
mls-vertical-coverage

main-qimg-5a23f17811f19b673d8204fab7184dea

glidang

DonLakeFlyer:
@Kjkinney Could you move this to a new Issue and we can discuss there.

PR #5181, opened 2017-05-22
WIP: Video streaming support

AlexeyBulatov:
This PR based on added https://github.com/mavlink/mavlink/pull/677 and new https://github.com/mavlink/mavlink/pull/729.
It adds the following:
1. New class CameraInformationFactGroup and function for handle mavlink message CAMERA_INFORMATION.
2. Function for handle VIDEO_STREAM_CAPABILITIES message and download file by its URL.
3. Common VideoStreamingComponent for `Vehicle Setup` menu.
4. VideoStreamingComponentController and VideoStreamingFactGroup classes for get/set video stream parameters.
5. New summary screen:
screenshot from 2017-05-22 17-37-20
6. New screen of setups:
screenshot from 2017-05-22 17-36-28

Sequence of working:
1. QGC sends MAV_CMD_REQUEST_CAMERA_INFORMATION with camera_id 0 for all cameras.
2. Vehicle sends N (number of cameras) messages CAMERA_INFORMATION.
3. For every camera QGC sends MAV_CMD_REQUEST_VIDEO_STREAM_CAPABILITIES.
4. Vehicle answers url of file using VIDEO_STREAM_CAPABILITIES.
5. After QGC downloads and parses this files.
6. Then QGC sends REQUEST_VIDEO_STREAM_INFORMATION for all cameras.
7. Vehicle sends VIDEO_STREAM_INFORMATION.
8. Then user can change settings with SET_VIDEO_STREAM_SETTINGS, start video streaming with MAV_CMD_VIDEO_START_STREAMING and stop it with MAV_CMD_VIDEO_STOP_STREAMING.

We can reuse class CameraMetaData, but it should be reworked.
Support for choice of protocol (udp, rtsp) haven't been implemented yet. For this will be used URI. For udp connection `udp:[ip]:[port]`, for rtsp `rtsp`. Also available protocols can be stored in json file with another capabilities.

Example of json file with video stream capabilities:
[
{
    "name":"fps",
    "type":"double",
    "units":"Hz",
    "enumStrings":"auto,10 Hz,20 Hz,30 Hz,40 Hz",
    "enumValues":"0,10,20,30,40",
    "defaultValue":"0"
},
{
    "name":"resolution",
    "type":"uint16",
    "units":"pix X pix",
    "enumStrings":"auto,640x480,720x576,1280x720",
    "enumValues":"0,1,2,3",
    "defaultValue":"0"
},
{
    "name":"bitRate",
    "type":"uint32",
    "units":"b/s",
    "enumStrings":"auto,10 Mb,15 Mb,20 Mb,25 Mb",
    "enumValues":"0,10000000,15000000,20000000,25000000",
    "defaultValue":"0"
},
{
    "name":"rotation",
    "type":"uint16",
    "units":"deg",
    "enumStrings":"0 deg,90 deg,180 deg,270 deg",
    "enumValues":"0,90,180,270",
    "defaultValue":"0"
}
]


otaviobp:
Are you planning to have the camera streaming capabilities implemented on the flight stack? Do you have a flight stack with that support implemented so we can test this?

AlexeyBulatov:
For test this I use proxy which can process messages and setup video stream. I don't plan to integrate this to flight stack for now, maybe in future..

DonLakeFlyer:
I'm confused. If no flight stack supports this. Why should it be added to QGC?

dogmaphobic:
It's a chicken and the egg problem. You need this in, in order to build a flight stack to test. With that said, this is all still pretty much in flux. In my opinion, there are way too many MAVLink messages related to camera/video with way too many specialized arguments, some of which the GCS has no way to know how to offer options for their settings. These messages should be reduced to a minimal core set and add a new set of Get/Set messages using arbitrary Key/Value pairs. For that to work, you need a camera definition file that allows the GCS to build a UI with the proper settings and options. If a camera lacks this definition file (or the system doesn't have enough resources to implement it), then only the core set of messages would be used.

There is an ongoing discussion about this: http://discuss.px4.io/t/mavlink-camera-api-discussion/3278

PR #5184, opened 2017-05-23
Update vagrant file to lastest QT version

khancyr:
Just update the vagrantfile with what is done in travis.
Tested on Ubuntu 16.04

### DIFF
--- a/.vagrantconfig.yml
+++ b/.vagrantconfig.yml
@@ -1,13 +1,13 @@
 configs:
     dev:
-        'qt_deps_tarball': 'Qt5.5.1-linux.tar.bz2'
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
The issue with this is that the min version you are installing has missing pieces:
* Debug symbols are stripped out
* Tools like QtCreator are not in there

Various other things which I can't remember. The min version is mainly stripped down to just build from the command line. Not for a user  needing a development environment.

khancyr:
ho ! It was the first time I try to compile QGC and the vagrantfile was broken. 
Should I just change the min tar by the complet one or a more complet fix is wanted ?

khancyr:
closed because I didn't that another PR handle this

Issue #5069, opened 2017-04-27
Support negative thrust in MANUAL_CONTROL


khancyr:
Hello,

I am implementing MANUAL_CONTROL message for APMRover and it needs the support of negative thrust to go backward. But currently Qgroundcontrol only send positive values on throttle. 
Maybe adding a new button to enable the negative throttle should be safe enough to prevent mistakes. On Arducopter, negative thrust are rejected.

khancyr:
done in #5185

PR #5185, opened 2017-05-23
Joystick: add a button to support negative thrust

khancyr:
Hello ! 
This address https://github.com/mavlink/qgroundcontrol/issues/5069. But I never use Qt before, so I am not sure of the button logic.
It has been tested with https://github.com/ArduPilot/ardupilot/pull/6128 on SITL successfully!

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -463,7 +463,7 @@ void Joystick::run(void)
 
             // Adjust throttle to 0:1 range
             if (_throttleMode == ThrottleModeCenterZero && _activeVehicle->supportsThrottleModeCenterZero()) {
-                if (!_negativeThrust) {
+                if (!_activeVehicle->supportsNegativeThrust() || !_negativeThrust) {
                     throttle = std::max(0.0f, throttle);
                 }
             } else {
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -55,6 +55,7 @@ Joystick::Joystick(const QString& name, int axisCount, int buttonCount, int hatC
     , _rgButtonValues(NULL)
     , _lastButtonBits(0)
     , _throttleMode(ThrottleModeCenterZero)
+    , _negativeThrust(false)
     , _exponential(0)
     , _accumulator(false)
     , _deadband(false)
@@ -462,7 +463,9 @@ void Joystick::run(void)
 
             // Adjust throttle to 0:1 range
             if (_throttleMode == ThrottleModeCenterZero && _activeVehicle->supportsThrottleModeCenterZero()) {
-                throttle = std::max(0.0f, throttle);
+                if (!_negativeThrust) {
+                    throttle = std::max(0.0f, throttle);
+                }
             } else {
                 throttle = (throttle + 1.0f) / 2.0f;
             }
@@ -685,6 +688,22 @@ void Joystick::setThrottleMode(int mode)
 
     _saveSettings();
     emit throttleModeChanged(_throttleMode);
+}
+
+bool Joystick::negativeThrust(void)
+{
+    return _negativeThrust;
+}
+
+void Joystick::setNegativeThrust(bool allowNegative)
+{
+    if (_negativeThrust == allowNegative) {
+        return;
+    }
+    _negativeThrust = allowNegative;
+
+    _saveSettings();
+    emit negativeThrustChanged(_negativeThrust);
 }
 
 float Joystick::exponential(void)
--- a/src/VehicleSetup/JoystickConfig.qml
+++ b/src/VehicleSetup/JoystickConfig.qml
@@ -296,7 +296,7 @@ SetupPage {
                         Connections {
                             target: _activeJoystick
 
-                            onManualControl: throttleLoader.item.axisValue = (-2*throttle+1)*32768.0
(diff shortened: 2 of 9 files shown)


khancyr:
rebased

src/Joystick/Joystick.cc
```diff
@@ -687,6 +689,19 @@ void Joystick::setThrottleMode(int mode)
...
 }
 
+bool Joystick::negativeThrust(void)
+{
+    return _negativeThrust;
+}
+
+void Joystick::setNegativeThrust(bool allowNegative)
+{
+    _negativeThrust = allowNegative;
```
DonLakeFlyer: There should be a check here whether allowNegative is actually different than the current setting. And only in that case do you need to do anything here. That prevent unnecessary signaling.

DonLakeFlyer:
@jaxxzer Can you take a quick look at this. I believe this pull is correct. But I haven't been in the joystick code in ages.

khancyr:
Rebased and add a check to prevent unnecessary signaling.

jaxxzer:
Ok, from looking at it it seems ok, but I'm doing some testing.

I wonder if it's better to make the throttle range scaling a property of the firmwareplugin. I think that would be a cleaner all-around approach. @khancyr Is there any reason that an APM Rover user would want to operate without this box checked, or would you recommend all Rover users to check the box?

The only place I see where making it a part of the firmwareplugin wouldn't be a good idea is if apm or px4 support inverted flying with helis. For those who are flying helis inverted though, I bet they don't want to do it with a joystick.

DonLakeFlyer:
> I wonder if it's better to make the throttle range scaling a property of the firmwareplugin.

The reason for it not to be there is for a generic mavlink user to be able to use it. Question is whether that is important enough to add yet another option to the ui. Would ne nice to not have antoher checkbox.

src/VehicleSetup/JoystickConfig.qml
```diff
@@ -453,6 +453,14 @@ SetupPage {
 
                                     onClicked: _activeJoystick.throttleMode = 1
                                 }
+
+                                QGCCheckBox {
```
jaxxzer: Should the visibility of this option depend on https://github.com/mavlink/qgroundcontrol/pull/5185/files#diff-4dd9fb0da73f9007de515d2da5b6dbfdR422?

jaxxzer: It would seem the answer is yes https://github.com/mavlink/qgroundcontrol/pull/5185/files#diff-7179de35ff534916a12fd2e757566fe8R464

jaxxzer:
What about making the visibility of the option depend on firmwareplugin? And generic mavlink and APM Rover will support the option and be the only two FWs to display it.

jaxxzer:
Also, the Joystick indicator slider behavior is funky with this option engaged. The indicator slides off the left hand side of the screen when the values go negative. @khancyr 

Other than that looks to be OK.

DonLakeFlyer:
> What about making the visibility of the option depend on firmwareplugin? And generic mavlink and APM Rover will support the option and be the only two FWs to display it.

Perfect.

khancyr:
Ideally, if we follow the mavlink specification the reverse thrust option should be always on and the vehicule must be in charge of dealing with negative value. It is the case in arducopter. For rover, I don't see why it shouldn't be on too by default. I will look at how to make this on by default. 
Sub could that adavantage of this too and remove the scaling done in ardusub code.

jaxxzer:
>Sub could that adavantage of this too and remove the scaling done in ardusub code.

Agree, but can't deal with this right now. Also need to be careful with changing this. I don't think the mavlink definition is clear enough for both the GCS to know what to send, and the vehicle to know what to do. The GCS and the Vehicle need to figure out how it is going to be done with some rules outside of mavlink spec.

For example, does -1000 always correspond to one stick extreme, and +1000 correspond to the other stick extreme, in which case, does 'negative thrust' mean throttle below 50% in APM copter terms? Or is APM Copter only going to expect positive values and clamp anything negative coming in to 0?

jaxxzer:
@khancyr, can you fix the throttle indicator going off screen, and make the visibility of this option a property of firmwareplugin as discussed above?

khancyr:
@jaxxzer Sorry for the delay, I need some time to understant how QGC works. It should be good now !
I just didn't find how to allow negativethrust on generic vehicle ... so only Rover got the override for now.

khancyr:
@jaxxzer ping !

jaxxzer:
Ok, looks good to me now. Thanks @khancyr and sorry for the delay.

jaxxzer:
> The reason for it not to be there is for a generic mavlink user to be able to use it. Question is whether that is important enough to add yet another option to the ui. Would ne nice to not have antoher checkbox.

Right now this option is only visible to Rover, not generic vehicles.

DonLakeFlyer:
@jaxxzer So this is good to merge now?

jaxxzer:
It's fine with me, the only bit missing is that it won't show up for generic mavlink vehicles. I don't think that's too important.

DonLakeFlyer:
@khancyr Can you rebase against master and fix up the conflicts and then I'll merge this through.

khancyr:
@DonLakeFlyer @jaxxzer Sorry I had very limited access until now... I have rebased.
I could do another PR later to permit negativethrust on generic vehicle

DonLakeFlyer:
Perfect thanks

PR #5202, opened 2017-05-25
trying to fix daily build uploads

DonLakeFlyer:


dagar:
How did this ever work before?

DonLakeFlyer:
Don't know. Also don't know if this is working yet. Still waiting for a master to go through. If it does, i'd bet appveyor changed something inernally.

PR #5218, opened 2017-05-27
Use image for GCS location on map

DonLakeFlyer:
This allows plugin to override with better graphics


DonLakeFlyer:
@dogmaphobic The image is biased to setting height only. That way plugins can create something that is non-square with just an override.

jaxxzer:
How do you display the gcs location on the map? Is it available on non-mobile platforms?

DonLakeFlyer:
It's available anywhere Qt supports it's location service. I know it works on OSX if you have WiFI turned on. Not sure about windows.

PR #5226, opened 2017-05-30
Bugfix for factcomboboxes on Sub safety page

jaxxzer:


jaxxzer:
@DonLakeFlyer Can you squeeze this one in now?

PR #5233, opened 2017-06-03
Automatic protocol version handshake

LorenzMeier:
This extension allows to negotiate with all connected vehicles the highest possible MAVLink version supported by all. It is compatible with legacy vehicle software (interprets NACK and timeouts as v1.0 requests). This allows to have the best possible version enabled without any user configuration.

Before connecting QGC:

```
instance #0:
	mavlink chan: #0
	no telem status.
	flow control:	OFF
	rates:
	tx: 23.215 kB/s
	txerr: 0.000 kB/s
	rx: 0.000 kB/s
	rate mult: 1.000
	accepting commands: YES
	MAVLink version: 1
	transport protocol: UDP (14556)
```

After connecting QGC with this extension:

```
instance #0:
	GCS heartbeat:	988721 us ago
	mavlink chan: #0
	type:		GENERIC LINK OR RADIO
	flow control:	OFF
	rates:
	tx: 30.495 kB/s
	txerr: 0.000 kB/s
	rx: 0.021 kB/s
	rate mult: 1.000
	accepting commands: YES
	MAVLink version: 2
	transport protocol: UDP (14556)
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "libs/mavlink/include/mavlink/v1.0"]
-	path = libs/mavlink/include/mavlink/v1.0
-	url = https://github.com/mavlink/c_library_v1.git
 [submodule "src/GPS/Drivers"]
 	path = src/GPS/Drivers
 	url = https://github.com/PX4/GpsDrivers.git
--- a/src/Vehicle/MultiVehicleManager.cc
+++ b/src/Vehicle/MultiVehicleManager.cc
@@ -131,6 +131,7 @@ void MultiVehicleManager::_vehicleHeartbeatInfo(LinkInterface* link, int vehicle
 
     Vehicle* vehicle = new Vehicle(link, vehicleId, componentId, (MAV_AUTOPILOT)vehicleFirmwareType, (MAV_TYPE)vehicleType, _firmwarePluginManager, _joystickManager);
     connect(vehicle, &Vehicle::allLinksInactive, this, &MultiVehicleManager::_deleteVehiclePhase1);
+    connect(vehicle, &Vehicle::requestProtocolVersion, this, &MultiVehicleManager::_requestProtocolVersion);
     connect(vehicle->parameterManager(), &ParameterManager::parametersReadyChanged, this, &MultiVehicleManager::_vehicleParametersReadyChanged);
 
     _vehicles.append(vehicle);
@@ -159,6 +160,30 @@ void MultiVehicleManager::_vehicleHeartbeatInfo(LinkInterface* link, int vehicle
     }
 #endif
 
+}
+
+/// This slot is connected to the Vehicle::requestProtocolVersion signal such that the vehicle manager
+/// tries to switch MAVLink to v2 if all vehicles support it
+void MultiVehicleManager::_requestProtocolVersion(unsigned version)
+{
+    unsigned maxversion = 0;
+
+    if (_vehicles.count() == 0) {
+        _mavlinkProtocol->setVersion(version);
+        return;
+    }
+
+    for (int i=0; i<_vehicles.count(); i++) {
+
+        Vehicle *v = qobject_cast<Vehicle*>(_vehicles[i]);
+        if (v && v->maxProtoVersion() > maxversion) {
+           maxversion = v->maxProtoVersion();
+        }
+    }
+
+    if (_mavlinkProtocol->getCurrentVersion() != maxversion) {
+        _mavlinkProtocol->setVersion(maxversion);
+    }
 }
 
 /// This slot is connected to the Vehicle::allLinksDestroyed signal such that the Vehicle is deleted
--- a/src/Vehicle/MultiVehicleManager.h
+++ b/src/Vehicle/MultiVehicleManager.h
@@ -95,6 +95,7 @@ private slots:
     void _vehicleParametersReadyChanged(bool parametersReady);
     void _sendGCSHeartbeat(void);
     void _vehicleHeartbeatInfo(LinkInterface* link, int vehicleId, int componentId, int vehicleMavlinkVersion, int vehicleFirmwareType, int vehicleType);
+    void _requestProtocolVersion(unsigned version);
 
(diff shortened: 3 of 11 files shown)


LorenzMeier:
Matching firmware PR: https://github.com/PX4/Firmware/pull/7344

LorenzMeier:
@DonLakeFlyer Firmware PR is through. If you can review / cross-test we can get this in (but its not needed for v3.2).

LorenzMeier:
@DonLakeFlyer Full documentation: https://mavlink.io/en/mavlink-version.html#version-handshaking

DonLakeFlyer:
This isn't quite right. It treats a links as a single group. Either mavlink 1 or 2 based on lowest common denominator vehicle reporting. It should really be on a link by link basis based only on the vehicles on the specific link. That said, I think that gets way more complicated. This should be fine for 3.2. And maybe forever.

I still don't get how if there is a radio that doesn't support mav 2 in the middle how this works. Mavlink defaults to mav 1 for status flags doesn't it? So initial send of MAV_CMD_REQUEST_PROTOCOL_VERSION will be over mavlink 1 protocol. And it will go through. Is the link supposed to start out as mav 2 and then fall back? If so I don't see the code for that.

LorenzMeier:
@DonLakeFlyer The reason I don't want to treat it link-by-link, vehicle-by-vehicle is that managing that configuration space is super error prone (imagine that you can send geofence over one link, but not over another - how do you want to deal with that?). Anyone who knows what he is doing will use MAVLink 2 exclusively. The only reason to have detection and fallback is to ensure that random users don't get bitten when they upgrade to a new QGC version and their old plane is not yet supporting MAVLink 2 throughout.

But really, the main focus here is to enable a safe upgrade path - I'm not really trying to support mixed setups.

In terms of detection: The trick is that the PROTOCOL_VERSION message is sent in MAVLink 2 framing. That ensures that the link supports 2.0. But the message will in the future be also important for cases where we have v2.1 or v2.5.

LorenzMeier:
This is good to go. If the remote vehicle does not support discovery but sends MAVLink 2 it will stick with that as default.

LorenzMeier:
@DonLakeFlyer I haven't tested this yet on hardware, but wanted to see if it fits the flash. This branch teaches the SiK radio to respond properly to PROTOCOL_VERSION, which in turn would allow QGC to detect wether or not the radio can do MAVLink 2. It likely will require adjustments to this pull, but this would be the watertight solution.

LorenzMeier:
I will test this one round more and then get it in. @bkueng This is needed for polygons and stuff as it switches the system automatically.

LorenzMeier:
Re-tested: With just USB connected the link auto-switches to MAVLink 2, with a 3DR Radio it stays at MAVLink 1. And with forced MAVLink 2 it also seems to work:

```
nsh> mavlink status

instance #0:
	GCS heartbeat:	810149 us ago
	mavlink chan: #0
	type:		3DR RADIO
	rssi:		219
	remote rssi:	219
	txbuf:		100
	noise:		56
	remote noise:	45
	rx errors:	0
	fixed:		0
	flow control:	ON
	rates:
	tx: 0.138 kB/s
	txerr: 0.000 kB/s
	rx: 0.059 kB/s
	rate mult: 0.058
	accepting commands: YES
	MAVLink version: 2
	transport protocol: serial (/dev/ttyS1 @57600)

instance #1:
	mavlink chan: #1
	no telem status.
	flow control:	OFF
	rates:
	tx: 0.308 kB/s
	txerr: 0.000 kB/s
	rx: 0.014 kB/s
	rate mult: 0.050
	accepting commands: YES
	MAVLink version: 2
	transport protocol: serial (/dev/ttyS2 @57600)

instance #2:
	GCS heartbeat:	877322 us ago
	mavlink chan: #2
	type:		USB CDC
	flow control:	OFF
	rates:
	tx: 1.518 kB/s
	txerr: 0.000 kB/s
	rx: 0.021 kB/s
	rate mult: 1.000
	accepting commands: YES
	MAVLink version: 2
	transport protocol: serial (/dev/ttyACM0 @2000000)
nsh> 
```

From QGC output: Switching outbound to mavlink 2.0 due to incoming mavlink 2.0 packet

LorenzMeier:
This still needs work - I still need to test with the upgraded radio.

DonLakeFlyer:
Looking through this I can't quite see if it delays the initial plan request until it knows the mavlink version?

LorenzMeier:
@DonLakeFlyer Yes it does, because it queues the command requesting the protocol version and only fully initializes when that is received.

I've also taken care of the SiK / 3DR radio: This PR now contains an updated link to a stable build that is a) MAVLink 2.0 enabled and b) reacts correctly to the handshaking mechanism. Code for that is here: https://github.com/LorenzMeier/SiK/pull/1

If a user has not yet updated his 3DR radio QGC will tell him now. It will still work as a link but a little less well than with the new binary. With this addition this should be watertight and good upgrade path for the user base. It's properly documented, takes care of the corner cases and does not leave existing users behind.

@dogmaphobic You will want to upgrade your ESP8266 bridge too. Here are the handshaking docs: https://mavlink.io/en/mavlink-version.html

From my side this is now all good to go.

DonLakeFlyer:
> Yes it does, because it queues the command requesting the protocol version and only fully initializes when that is received.

I still don't see how this causes _startPlanRequest to not be called until the protocol version is known. The code in Vehicle still triggers _startPlanRequest on capability bits known and parameters ready. I don't see any checked for protocol being known as well.

LorenzMeier:
Let me re-check it. I was relatively certain that the remaining load could not happen until the first commands were handled, but I might have missed a parallel code path. I will update here.

LorenzMeier:
@DonLakeFlyer I made the dependency explicit.

DonLakeFlyer:
@LorenzMeier Yup. Looks good. Any reason to not merge now. Then I can use this in my GeoFence stuff as well.

LorenzMeier:
Ok, merged.

PR #5247, opened 2017-06-05
Work on Video Streaming (mostly related to recording it)

dogmaphobic:
Adding mp4 and mov files types in addition to mkv
Adding mux GStreamer plugin libraries to Android (to enable video recording on Android)
Adding an option to limit the amount to disk used for saving videos
Adding an option to show/hide the REC button (allowing plugins to handle the recording in their own way)

### DIFF
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -359,7 +359,7 @@ QGCView {
             anchors.right:      _flightVideo.right
             height:             ScreenTools.defaultFontPixelHeight * 2
             width:              height
-            visible:            QGroundControl.videoManager.videoRunning && QGroundControl.videoManager.recordingEnabled
+            visible:            QGroundControl.videoManager.videoRunning && QGroundControl.settingsManager.videoSettings.showRecControl.rawValue
             opacity:            0.75
 
             Rectangle {
--- a/src/FlightDisplay/FlightDisplayViewVideo.qml
+++ b/src/FlightDisplay/FlightDisplayViewVideo.qml
@@ -23,9 +23,8 @@ import QGroundControl.Controllers   1.0
 
 Item {
     id: root
-    property double _ar:        QGroundControl.settingsManager.videoSettings.aspectRatio.rawValue
-    property bool _showGrid:    QGroundControl.settingsManager.videoSettings.gridLines.rawValue > 0
-
+    property double _ar:            QGroundControl.settingsManager.videoSettings.aspectRatio.rawValue
+    property bool   _showGrid:      QGroundControl.settingsManager.videoSettings.gridLines.rawValue > 0
     Rectangle {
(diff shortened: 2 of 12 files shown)


dogmaphobic:
@jaxxzer @NaterGator Could you please take a quick peek at this?

Two things of notice:

For `mov` and `mp4`, I had to add `_sink->parse    = gst_element_factory_make("h264parse", NULL);` for them to work. As it didn't affect the original matroskamux, I didn't make it conditional.

VLC will play any of the 3 file types fine but QuickTime won't. Neither will a normal gstreamer pipeline. Something like:

`gst-launch-1.0 filesrc location=2017-06-05_01.01.18.mkv ! matroskademux ! avdec_h264 ! osxvideosink`

Am I missing something?

Thanks!

NaterGator:
h264parse element is a good idea and does not need to be removed. 

For the offending file on OSX can you just tryin gst-launch-1.0 playbin url=file://2017-06-05_01.01.18.mkv?

NaterGator:
We may also want to include MPEG-TS as a container format option; it is specifically called out by MISB [MISP-2017.2](http://www.gwg.nga.mil/misb/docs/misp/MISP-2017.2.pdf) and is a good container format for video streams originating over lossy links.

dogmaphobic:
No dice, but with that said, I've never had success with playbin on macOS. It doesn't work with any file, not just the ones created above.

```
$ gst-launch-1.0 playbin url=file:///Volumes/Data/Documents/QGroundControl/Video/2017-06-05_14.11.54.mkv
Setting pipeline to PAUSED ...
ERROR: Pipeline doesn't want to pause.
Setting pipeline to NULL ...
```

I will take a look at adding MPEG-TS. I need to see what dependencies there are for `mpegtsmux`.

NaterGator:
On your original pipeline you can always set 
`gst-launch-1.0 --gst-debug=3 filesrc location=2017-06-05_01.01.18.mkv ! matroskademux ! avdec_h264 ! osxvideosink`

And try increasing debug level past 3 if you need more verbosity. Are you able to share a sample clip?

jaxxzer:
> We can also insert our own private data stream in MPEG2 TS and, for example, record mavlink traffic passing over links time-synchronized with the video feed which makes for really elegant playback options.

I'm very interested in this, great information, thanks.

jaxxzer:
@dogmaphobic Your pipeline works for me on ubuntu after changing osxvideosink to autovideosink. Is autovideosink available on osx?

dogmaphobic:
>Is autovideosink available on osx?

No, that's why I use osxvideosink. It works for everything else.

Here is a very short video captured by QGC (compressed with zip as Github would not allow uploading otherwise):

[2017-06-05_14.11.54.mkv.zip](https://github.com/mavlink/qgroundcontrol/files/1052975/2017-06-05_14.11.54.mkv.zip)

PR #5253, opened 2017-06-05
Survey: Settable entry location

DonLakeFlyer:
This allows you to specicy the entry location you want for the survey grid. Important for:
* Creating shortest distance survey flight path from takeoff to land
* Moving entry point to work around obstacles in path

Implementing this was a royal PITA. It really needs a comprehensive unit test to make sure it is all working which will come next. Wouldn't be suprised by boundary case problems.


DonLakeFlyer:
@antiheavy Guessing this will be very useful for you.

Antiheavy:
Cool. Looking forward to using it.

PR #5279, opened 2017-06-11
Add 500000 baud to all builds

DonLakeFlyer:


DonLakeFlyer:
@antiheavy I don't know if this will work. But you can try once it is up.

PR #5288, opened 2017-06-13
Solve error in survey input via windrose

patrickelectric:


patrickelectric:
Before:
antes


After:
funfa

Issue #5232, opened 2017-06-02
RTK GPS UI work and enhancements


mhkabir:
@LorenzMeier @dogmaphobic @DonLakeFlyer We need some UI work to get RTK GPS to a stage where it can be used by the average user. This is applicable for both APM and PX4 flight stacks. This would allow us to cleanly support the inexpensive UBlox M8P modules (e.g ProfiCNC Here+ or the Drotek XL RTK). I have only tested with the Drotek module till now, and everything works internally. However, the status of Survey-in, whether RTK GPS is connected at all, etc. is only visible in the console and this is not suitable for an user.

### RTK status indicator
There is need for an icon on the top panel visible when RTK GPS connected - but _irrespective_ of whether a  vehicle is connected.

![image](https://cloud.githubusercontent.com/assets/5557676/26741774/62600f3c-47f8-11e7-9d12-64d72c73f3ff.png)

It would be an icon on the top bar (to the left of "Waiting For Vehicle Connection" maybe?) like the one we have for the vehicle's GPS status, and show the following information when clicked : 
- Satellite count for RTK base
- Accuracy of base position
- Survey-in status for base (survey-in position fixed or not, etc.)
- Time survey-in has been running if not completed.

### Map indication

The base station location derived from the RTK base station should be used and displayed as QGC's location on non-mobile platforms (the "Q" icon on the map).

#### Galileo support
The PX4 GpsDrivers stack needs update to support the RTCM corrections (message XXXX) from satellites in the Galileo constellation. I will send in a PR with these changes soon. On the M8P modules, firmware version 3.01 from UBlox is needed. 

#### MAVLink 2.0 support
We dearly need MAVLink 2.0 to optimize packet size for the RTCM injection messages. Currently we just send the full 182-byte message even though only the top 22 bytes or so are filled for some messages (e.g base position). MAVLink 2 will automatically chop off the empty packets - allowing for a much leaner transmission requirement for the uplink.

#### Verify bandwidth requirements
[...]

PR #5296, opened 2017-06-14
[WIP] First stab at GPS RTK indicator

LorenzMeier:
This is not yet ready for review / testing, but shows the direction I'm trying to take it into. @DonLakeFlyer If you beat me to it it would be fantastic to have an indicator in the toolbar representing the local RTK unit.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -812,6 +812,7 @@ HEADERS+= \
     src/FirmwarePlugin/FirmwarePlugin.h \
     src/FirmwarePlugin/FirmwarePluginManager.h \
     src/Vehicle/MultiVehicleManager.h \
+    src/Vehicle/GPSRTKFactGroup.h \
     src/Vehicle/Vehicle.h \
     src/VehicleSetup/VehicleComponent.h \
 
@@ -836,6 +837,7 @@ SOURCES += \
     src/FirmwarePlugin/FirmwarePlugin.cc \
     src/FirmwarePlugin/FirmwarePluginManager.cc \
     src/Vehicle/MultiVehicleManager.cc \
+    src/Vehicle/GPSRTKFactGroup.cc \
     src/Vehicle/Vehicle.cc \
     src/VehicleSetup/VehicleComponent.cc \
 
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -6,6 +6,7 @@
         <file alias="ArmedIndicator.qml">src/ui/toolbar/ArmedIndicator.qml</file>
         <file alias="BatteryIndicator.qml">src/ui/toolbar/BatteryIndicator.qml</file>
         <file alias="GPSIndicator.qml">src/ui/toolbar/GPSIndicator.qml</file>
+        <file alias="GPSRTKIndicator.qml">src/ui/toolbar/GPSRTKIndicator.qml</file>
         <file alias="MessageIndicator.qml">src/ui/toolbar/MessageIndicator.qml</file>
         <file alias="ModeIndicator.qml">src/ui/toolbar/ModeIndicator.qml</file>
         <file alias="RCRSSIIndicator.qml">src/ui/toolbar/RCRSSIIndicator.qml</file>
@@ -209,6 +210,7 @@
         <file alias="Vehicle/VehicleFact.json">src/Vehicle/VehicleFact.json</file>
         <file alias="Vehicle/BatteryFact.json">src/Vehicle/BatteryFact.json</file>
         <file alias="Vehicle/GPSFact.json">src/Vehicle/GPSFact.json</file>
+        <file alias="Vehicle/GPSRTKFact.json">src/Vehicle/GPSRTKFact.json</file>
         <file alias="Vehicle/WindFact.json">src/Vehicle/WindFact.json</file>
         <file alias="Vehicle/VibrationFact.json">src/Vehicle/VibrationFact.json</file>
         <file alias="Vehicle/TemperatureFact.json">src/Vehicle/TemperatureFact.json</file>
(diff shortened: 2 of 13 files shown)


bkueng:
I brought this into a first working state:
qgc_rtk_survey-in_toolbar

Needs more work on the UI and the status should be shown even if no vehicle is connected. Related: #5232

DonLakeFlyer:
So awesome, thanks @bkueng

hamishwillee:
That's cool. Is it just me who scans "Survey-in Active" as "Survey Inactive"?

DonLakeFlyer:
@bkueng Could you:
* Move the RTK FactGroup to QGroundControlQmlGlobal
* Put RTK FactGroup into it's own source/header files
* #ifdef it to be out for mobile builds both in source and qgroundcontrol.pro

Then I think this is perfect for 3.2. For 3.3 we can make it work when no vehicle is connected.

bkueng:
@DonLakeFlyer I'll do that.

@hamishwillee Yes it takes some time to get used to the term. To me it looks ok, but if you have somehing better, we can use that.

bkueng:
I made the following updates:
- used QGroundControlQmlGlobal & moved the source into separate src/header files.
- added the ifdefs for mobile (I think the new source files don't need to be ifdefed)
- the icon is only visible if there is a GPS connected (and a vehicle)
- the icon is red during survey-in (meaning the user should wait), then turns white (meaning it's ready to be used)
- UI & labeling updates.

Looks like this:
qgc_rtk_gps_status2

Just to test I added the QML stuff to MainToolBar.qml, so that the icon is visible without a vehicle connection:
qgc_rtk_gps_status3
It works, but I think the icon is too large, and I did easily see how to change it. @DonLakeFlyer I think this way would be preferred, if we can make it work.

hamishwillee:
@bkueng 
> Yes it takes some time to get used to the term. To me it looks ok, but if you have somehing better, we can use that.

Hard to say without knowing what this does? I am guessing you have a QGC planning mode called "survey-in" where you define  some boundaries and QGC auto-runs/plans a survey inside that boundary. How does RTK fit into that - just makes sure that you stick very closely to those boundaries?

Anyway, without knowing anything about this you could try "Survey-Within Active"

bkueng:
@hamishwillee survey-in is an RTK-related term, coined by u-blox. It describes the startup procedure to get an accurate position for the base station. It is described here: https://docs.px4.io/en/advanced_features/rtk-gps.html. I'll update that once we get this merged.

@DonLakeFlyer CI failure is a build timeout. I tried restarting several times, but always the same problem.

hamishwillee:
Thanks @bkueng . Doing a search on RTK and survey-in shows up nothing useful, so my guess is that from an end user perspective this is an unhelpful term. 

Perhaps "RTK Calibration Active", "RTK GPS Calibration Active", "RTK GPS Setup Active", "Precision GPS Setup Active". I admit not as pithy, but actually meaningful (and there is a chance that they will find something useful if they search on RTK GPS"

bkueng:
Ok, so the u-blox term hasn't caught on. I agree then it should be something meaningful. How about "Acquiring Precise Position"?

The settings page needs updating as well then:
qgc_rtk_survey-in_settings
We can say: "Required position accuracy" instead of "survey in accuracy".

hamishwillee:
I don't mind "Acquiring Precise Position"? but it sounds like something you do only once, and in fact what I think we're doing is calibrating our equipment or setting it up to get precision position information. 

Perhaps "Initiating Precision Positioning", "Starting Precision Location", "Precision Location Setup", "Precision Location Activation" (I quite like the last one, though of course this would be the text on the prompt, not "Precision Location Activation Active" :-)

For the settings page, sure! Alternatives there, maybe "Positional accuracy"  or "Minimal accuracy" (my preference)

bkueng:
Yes you only do it once, then you can fly as many flights as you want, with different vehicles, etc, as long as you keep the base GPS running and at the same location. For that it's an accurate description of what's happening (the base station is actually determining an accurate position by averaging and converging over time).
Either of your suggestions works as well. I think you should see it in action, then you would see that the "Current Accuracy" converges towards 0 over time.

DonLakeFlyer:
@antiheavy For folks doing surveys with RTK they tend to know what "survey-in" means don't they?

DonLakeFlyer:
@bkueng Just getting back in town. I'll look at this tonight.

hamishwillee:
@bkueng - lets see what @Antiheavy says to @DonLakeFlyer. If this is only ever going to be shown to users with knowledge of RTK then Survey-in is OK. We'd make the link for ordinary users in the docs.

DonLakeFlyer:
@bkueng For showing with no vehicle available it should still be small like a toolbar indicators. It's just that it would require more changes to MainToolbarIndicators to make it work since that only shows when there is a vehicle.

DonLakeFlyer:
@bkueng Looks great. Big thanks. Also for 3.3: https://github.com/mavlink/qgroundcontrol/issues/5319

Antiheavy:
The "Survey-in" term isn't a universally understood thing, but in the context of these specific uBlox GPS receivers I think it is fine to use since that is the term uBlox uses.  This could be revisited in the future if Qgroundcontrol is extended to support more/other RTK systems.

"Base Station" is a fairly universally understood term in the world of RTK GPS referring to the (usually stationary) GPS receiver/equipment on the ground side.  To avoid the "Survey Inactive" wording problem I probably would have said something like "Base station status: Surveying-In" and "Base station status: sending RTCM" or something along those lines.

At any rate, what you have here is better than watching a stream of text print to the console so thanks.

As a future improvement I suggest making it more clear what GPS is for the vehicle and what GPS icon is for the "Base Station".   Or better yet, just combine them into 1 toolbar indicator that is typically vehicle centric, but when clicked on for details provides details of both the vehicle GPS and also the base station GPS.

DonLakeFlyer:
Forgot to enter icon issue for 3.2: https://github.com/mavlink/qgroundcontrol/issues/5324. @dogmaphobic?

DonLakeFlyer:
Also added a new issue for RTK wording which copies Todd's comment above: https://github.com/mavlink/qgroundcontrol/issues/5325

Francklin2:
Hi 
I worked on a RTKlib interface on raspberry( RTKLIB_Touchscreen_GUI ), for the base I use a image showing a groundstation, to be more clear you may use a icon like the second one upper left here for the ground base.
image

Thanks for your nice work on RTK intégration

DonLakeFlyer:
@Francklin2 Thanks. Could you move these images into a separate Issue for the RTK icon. And then I'll look at changing it.

Francklin2:
I put them  in the icon issue #5324 (looks like the correct one)

PR #5303, opened 2017-06-16
Mission Settings: Work on usability

DonLakeFlyer:
There is confusion around what the camera commands do in Mission Settings. The key is that they take affect immediately. So if you turn on the camera you are going to get photos/video from the ground on up.

Default visuals:
screen shot 2017-06-16 at 2 40 09 pm

Camera section manually opened:
screen shot 2017-06-16 at 2 40 17 pm


DonLakeFlyer:
More changes (changed name to Mission Start):
screen shot 2017-06-16 at 2 59 57 pm

PR #5314, opened 2017-06-20
Enable only the gestures we're interested in.

dogmaphobic:
Qt 5.9 has added a few more map gestures, some of which QGC doesn't yet handle (rotation, tilt, etc.) To avoid these, I'm now explicitly defining the desired (enabled) gestures for maps.


dogmaphobic:
Mac Installer build failed twice due to timeout. Jamming this through as everything else passed.

Issue #5163, opened 2017-05-16
Intrument panel overlaps multi-vehicle radios


DonLakeFlyer:
![screen shot 2017-05-16 at 11 55 33 am](https://cloud.githubusercontent.com/assets/5876851/26123115/9f739bf0-3a2e-11e7-90bb-506c8549230c.png)

DonLakeFlyer:
@dogmaphobic Can you take this one. The code here seems different from last I looked. Things used to be top anchored to the bottom of these radios. Now the instrument stuff seemds to be anchored to the bottom. Don't want to screw anything up here with plugin and so forth.

dogmaphobic:
Gazebo is broken. That's the only way I know how to test multiple vehicles (other than actually starting multiple vehicles, which is a bit of a pain :P ) I will take a look when that gets fixed.

DonLakeFlyer:
Start up two MockLink vehicles

hamishwillee:
And just should anyone find this in future, there is information about [Mock Link here](https://dev.qgroundcontrol.com/en/tools/mock_link.html)

PR #5322, opened 2017-06-22
Multi vehicle: Fix positioning of instrument panel

DonLakeFlyer:
Fix for #5163


DonLakeFlyer:
@dogmaphobic Can you double-check this?

Issue #5276, opened 2017-06-10
android/qgroundcontrol auto record tlog in a directory under /sdcard not working


marcmerlin:
I believe qgroundcontrol has a way to manually save tlogs, but it should do so automatically with no user input.
I did a few flights with qgroundcontrol on android yesterday, and none of those flights got recorded on my phone. 
I have "enable automatic logging" in Mavlink 2.0 Logging (PX4 Firmware Only), but it didn't do anything. Lower down in 'saved log files', I have absolutely nothing.

Most of the time, sadly qgroundcontrol stopped running due to the phone going to sleep, or some other reason, so the main point is that it should automatically record all flights without user input until the app dies or gets terminated (which can be done at inconvenient times on android vs a desktop version).
While I do like qgroundcontrol's interface and extra functionality, for now I have to go back to Tower for the reliable auto track recording feature.
But I'm looking forward to going back to qgroundcontrol as soon as the track recording is reliable :)

DonLakeFlyer:
@dogmaphobic I thought you had tuned on support for telemetry logs on Android but now on by default. But it look like the settings doesn't show up.

dogmaphobic:
It "should" be there. I'll take a look when I get back.

dogmaphobic:
I spent some time on this today and I can't quite tell what's going on. The _Start Log_ is properly sent and acknowledged, the log file is created but no log data is ever received. I'm using SITL as I don't have anything functional at the moment. After arming/disarming the "vehicle" several times, I ended up with several log files but obviously all with zero length as no data ever arrived.

DonLakeFlyer:
If the setting is showing up and I can turn it on then I can debug from there.

dogmaphobic:
It's all there. The same thing is happening on desktop. No log data arrives.

dogmaphobic:
Does SITL support this?

DonLakeFlyer:
Sure. Telemetry logs are just mavlink traffic.

dogmaphobic:
No, this is not telemetry log. These are ulog logs (as described above).

dogmaphobic:
>I have "enable automatic logging" in Mavlink 2.0 Logging (PX4 Firmware Only)

DonLakeFlyer:
[...]

PR #5357, opened 2017-06-28
Enable telemetry logging on mobile builds

dogmaphobic:
@DonLakeFlyer 

Here you go. This is for #5276


DonLakeFlyer:
I didn't realize this was all that was needed. I thought stuff was compiled out. Duh!

PR #5358, opened 2017-06-28
Support for custom build installers for Windows/OSX

DonLakeFlyer:
This allows custom builds which have overriden things like application name and so forth have those overrides flow through to the installers.


DonLakeFlyer:
@dogmaphobic @NaterGator Could you guys take a look a this? The changes basically remove a bunch of hardcoded references to things like app executable, org name and so forth. This way if a custom build overrides those things they will flow correctly into the information created by the installers. Current only Windows and OSX supported.

Still work left to do here:
* Windows installer branding

PR #5361, opened 2017-06-29
GeoTagController : Handle logging dropouts correctly

mhkabir:
This fixes the case where the log doesn't contain a particular trigger packet (logging dropout), but the image has been captured correctly. 

Without this fix, a missing log entry would throw off geotagging for all the subsequent frames. Critical fix, so should go into the release.


mhkabir:
@DonLakeFlyer @dagar Does the CI push/store the built app somewhere? Would like to provide some testers with the download link for the built app from this branch.

dagar:
Yes, but only for branches off the main repo.

PR #5373, opened 2017-06-30
ExternalLibs: Fix compilation on windows 64bit

mantelt:
Add switch to QGCExternalLibs.pri to check whether 32bit or 64bit version of SDL should be used on windows


mantelt:
QGC did not compile on my windows 7 with QT 5.7.1 and  MSVC2015_64bit

DonLakeFlyer:
Thanks

PR #5385, opened 2017-07-02
Deal with macdeployqt relocation issues

dogmaphobic:
Yet another attempt at the macOS relocation issue:

For some unknown reason, `macdeployqt` is messing the relocation of some frameworks during the build phase. This appears to only happen on Travis (I have never been able to reproduce this on a local build).

The "fix" is to let it run (it saves a boatload of work in picking the needed various Qt frameworks and plugins) but run a subsequent, wholesale relocation afterwards. This will guarantee that all frameworks are properly relocated.


DonLakeFlyer:
Thanks Gus.

mhkabir:
Thanks for this! I reproduced the issue locally yesterday, but I didn't realise at at the time that it's the same thing we've been facing on Travis.

PR #5407, opened 2017-07-07
Add latitude and longitude in Mission Settings

senosahisnu:
Add latitude and longitude in mission settings, so enable user to input coordinate of planned home position.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ user_config.pri
 *.suo
 *.uhf.txt
 *.opensdf
+*Thumbs.db
 thirdParty/qserialport-build-desktop/
 thirdParty/qserialport/bin/
 thirdParty/qserialport/lib/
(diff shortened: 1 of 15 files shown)


DonLakeFlyer:
This was there before and I took it out since it didn't seem useful. Why you do need such exact positioning of planned home position?

senosahisnu:
Well, in some project, it much easier to type the coordinate rather than zooming to the region and make sure the center of the map is the intended home position. Hopefully this will help others as well.

DonLakeFlyer:
Given the fact that you can both drag the home position to exactly where you want as well as center it to the map. I'd rather keep this simpler from a UI perspective. I haven't gotten much feedback that this as a important use case. I'll collect additional feedback from 3.2 stable and if I get enough complaints I'll put back lat/lon back into field settings. As I said, I had it in there before for a while during 3.2 development and specifically removed it because the UI got too noisy/confusing.

I certainly appreciate the pull request. But I try hard to keep to QGC user interface as minimal as possible and try to not to have to many different ways to do the same thing. For now, I'd rather not add this.

I've added an issue: https://github.com/mavlink/qgroundcontrol/issues/5418 so this doesn't get lost.

Issue #5411, opened 2017-07-07
"Simple" control mode selectable while in PX4


nkhoit:
The "Simple" flight mode from Ardupilot in the mode selection drop-down is selectable while you're flying a PX4 drone.

DonLakeFlyer:
There is a "Simple" flight mode in PX4 as well.

dagar:
Actually there isn't. https://github.com/PX4/Firmware/blob/master/src/modules/commander/px4_custom_mode.h#L45

I've copied this header into QGC to try and keep them straight. https://github.com/mavlink/qgroundcontrol/blob/348949d3755d70682a94b2996eb931196fb4cff5/src/FirmwarePlugin/PX4/px4_custom_mode.h

DonLakeFlyer:
@LorenzMeier What's up with Simple? I put it in ages ago when you told me to. Did this disappear?

LorenzMeier:
It did. I'm wondering if I should quickly add it or not.

LorenzMeier:
Fixed here: https://github.com/mavlink/qgroundcontrol/pull/5422

PR #5414, opened 2017-07-08
PX4 copy current custom modes

dagar:
https://github.com/mavlink/qgroundcontrol/issues/5411

The majority of these modes seem pretty common across many autopilots. Any interest in trying to handle this in the common mavlink spec?


LorenzMeier:
Replacing..

PR #5415, opened 2017-07-08
Fix mode switching

LorenzMeier:
The mode switching for PX4 was generating excessive feedback hiding the user-facing error message and was leaving the vehicle potentially in an unintended armed state.

### DIFF
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -485,7 +485,7 @@ bool FirmwarePlugin::_setFlightModeAndValidate(Vehicle* vehicle, const QString& 
         vehicle->setFlightMode(flightMode);
 
         // Wait for vehicle to return flight mode
-        for (int i=0; i<30; i++) {
+        for (int i=0; i<22; i++) {
             if (vehicle->flightMode() == flightMode) {
                 flightModeChanged = true;
                 break;
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -452,12 +452,15 @@ void PX4FirmwarePlugin::guidedModeChangeAltitude(Vehicle* vehicle, double altitu
 
 void PX4FirmwarePlugin::startMission(Vehicle* vehicle)
 {
-    if (!_armVehicleAndValidate(vehicle)) {
-        qgcApp()->showMessage(tr("Unable to start mission: Vehicle failed to arm."));
-        return;
-    }
 
-    _setFlightModeAndValidate(vehicle, missionFlightMode());
+    if (_setFlightModeAndValidate(vehicle, missionFlightMode())) {
+        if (!_armVehicleAndValidate(vehicle)) {
+            qgcApp()->showMessage(tr("Unable to start mission: Vehicle rejected arming."));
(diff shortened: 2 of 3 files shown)


src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -485,7 +485,7 @@ bool FirmwarePlugin::_setFlightModeAndValidate(Vehicle* vehicle, const QString&
         vehicle->setFlightMode(flightMode);
 
         // Wait for vehicle to return flight mode
-        for (int i=0; i<30; i++) {
+        for (int i=0; i<22; i++) {
```
LorenzMeier: This is 2.2 seconds, which is being tried 3 times, so a total of 6.6 seconds. It was 9 seconds before, which is evil because the operator would never expect to still get an action 10 seconds after he operated the UI.

We might want to still lower it to 1.5.

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
```diff
@@ -452,12 +452,15 @@ void PX4FirmwarePlugin::guidedModeChangeAltitude(Vehicle* vehicle, double altitu
 
 void PX4FirmwarePlugin::startMission(Vehicle* vehicle)
 {
-    if (!_armVehicleAndValidate(vehicle)) {
-        qgcApp()->showMessage(tr("Unable to start mission: Vehicle failed to arm."));
-        return;
-    }
 
-    _setFlightModeAndValidate(vehicle, missionFlightMode());
+    if (_setFlightModeAndValidate(vehicle, missionFlightMode())) {
```
LorenzMeier: If setting the flight mode fails it will now not arm.

src/Vehicle/Vehicle.cc
```diff
@@ -789,7 +789,7 @@ void Vehicle::_handleHilActuatorControls(mavlink_message_t &message)
 
 void Vehicle::_handleCommandAck(mavlink_message_t& message)
 {
-    bool showError = true;
+    bool showError = false;
```
LorenzMeier: If we didn't explicitly register for an error to be shown (below) we should not show spurious messages to the user. Any such "debugging info" should go into the app log.

LorenzMeier:
@DonLakeFlyer Should definitely go into 3.2 to prevent those evil corner-cases.

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
DonLakeFlyer: The problem is that with my testing (mission with waypoint too far) setting the flight mode doesn't fail. See discussion in #5415

src/FirmwarePlugin/FirmwarePlugin.cc
DonLakeFlyer: Yeah, I just picked something arbitrary. Make sense.

src/Vehicle/Vehicle.cc
DonLakeFlyer: Make sense. I could see how this could happen when it retries. If the vehicle actually got both commands it will correctly ack twice.

DonLakeFlyer:
All good stuff

LorenzMeier:
I've fixed it on the Firmware side to be consistent also for mission check fails. Will go into 1.6.4.

PR #5425, opened 2017-07-11
Use either A or mA for the current, but not centiAmps

jliphard:
At present, part of qgroundcontrl uses centiAmps as the units of the current. For first time users, and for most people in general, thinking in centiAmps (100 centiAmps = 1 Amp) is confusing. How about using either Amps or milliAmps? Using Amps might be best? In that case, some variables will need to go from UINT16 to FLOAT. (Presumably centiAmps where chosen to accommodate very large drones, since 200A = 20,000 centiAmps, whereas 200,000 milliAmps would exceed the UINT16).

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -903,6 +903,10 @@ void Vehicle::_handleSysStatus(mavlink_message_t& message)
         _batteryFactGroup.current()->setRawValue(VehicleBatteryFactGroup::_currentUnavailable);
     } else {
         // Current is in Amps, current_battery is 10 * milliamperes (1 = 10 milliampere)
+        // To reduce confusion and misunderstandings, suggest not to work in units of centiAmperes. 
+        // Consider either Amps or milliAmps. Note that at some point, this is converted into UINT16. 
+        // If people prefer A, then suggest FLOAT for _batteryFactGroup.current(); 
+        // If people prefer mA, suggest UINT32 large drones with large current draws > 65,535 mA are happy, too. 
         _batteryFactGroup.current()->setRawValue((float)sysStatus.current_battery / 100.0f);
     }
     if (sysStatus.voltage_battery == UINT16_MAX) {


DonLakeFlyer:
I'm confused. What exactly are you fixing here? Internally "current" is stored as Amps in a float. Where do you see it used as centiamps and/or UINT16?

jliphard:
Hi @DonLakeFlyer - Yes the current is float and has units of Amps _except_ in one place, which is in src/Vehicle/Vehicle.cc, line 905:

// Current is in Amps, current_battery is 10 * milliamperes (1 = 10 milliampere)
_batteryFactGroup.current()->setRawValue((float)sysStatus.current_battery / 100.0f);

I suggest that _batteryFactGroup.current() receives the current in units of Amps or milliAmps, but not centiAmps.

DonLakeFlyer:
The spec for the SYS_STATUS mavlink message is set in stone and can't be changed.

jliphard:
>The spec for the SYS_STATUS mavlink message is set in stone and can't be changed.

Ok - this seems like something that should be improved within the mavlink spec then. I'll open an issue there. Thanks, Don!

DonLakeFlyer:
The spec for existing fields in mavlink messages can't be changed.

jliphard:
The how about adding

<field type="int32_t" name="current_battery_mA" units="mA">Battery current in milliAmperes, -1: autopilot does not measure the current</field>

to the spec? Current code would keep working, and precision mA-resolution current monitoring would become possible?

dagar:
Is the issue the usability of what's displayed or limited resolution? Regardless of the mavlink spec we could still default to showing the current appropriately.

DonLakeFlyer:
> Regardless of the mavlink spec we could still default to displaying the current appropriately.

Which QGC already does.

jliphard:
> Which QGC already does.

Except in one place, which is in QGC->Widgets->Analyze->Airspeed.

PR #5427, opened 2017-07-12
iOS fixes

dogmaphobic:
I made these fixes for the iOS app store

 * New, required NSBluetoothPeripheralUsageDescription
 * Enabled QtQuick compiler
 * Hide _Save Path_ from the General Settings. The path was huge, breaking the whole UI as it was accommodating the size. I think there is no point setting that path on iOS anyway as you can only save it within the app's own data directory.


DonLakeFlyer:
> Hide Save Path from the General Settings.

Hmm, I'll have to look and see what I did. I thought I had already done that. Must have screwed somethiong up

PR #5437, opened 2017-07-15
Survey settings: Default to 70% overlap as sane default

LorenzMeier:
10% is really not workable for any stitching engine.

### DIFF
--- a/src/MissionManager/Survey.SettingsGroup.json
+++ b/src/MissionManager/Survey.SettingsGroup.json
@@ -64,7 +64,7 @@
     "min":              0,
     "max":              85,
     "units":            "%",
-    "defaultValue":     10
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
70%? That seems very high. @antiheavy Any opinions here for frontal and side overlap defaults?

Antiheavy:
Lorenz is correct, 10% is way too low for any realistic application. 

70% is a fine default for most stitching engines, although we like to use 80% long-lap and 60-65% side-lap for most of our stitching flights. 

If you don't plan to stitch but still want total ground coverage you stil want maybe 30% overlap to account for real world operational tolerances.

DonLakeFlyer:
Thanks Todd. I think we are all good here then with 70/70.

Issue #5397, opened 2017-07-04
Never pan map to lat/lon ~0.000 / 0.000


AndKe:
Reproduce:
Start application
Wait till an UAV with mission onboard connnects.
Once mission is downloaded, you will briefly fee it in the map, before map pan to UAV that does not have GPS fix, in the middle of nowhere.   Should operator wish to verify/see mission, then he needs to zoom out and do so manually:

![screenshot from 2017-07-04 09-44-40](https://user-images.githubusercontent.com/1468842/27821172-ab8bccde-60a1-11e7-989c-93d309003b2e.png)

DonLakeFlyer:
Sounds like more cases of ArduPilot making me crazy by sending crap gps values! I thought I had cauterzied all of these.

DonLakeFlyer:
This is a 0/0 lat/lon coming through from first fake home position in ArduPilot.

PR #5444, opened 2017-07-15
Handle bogus lat/lon from ArduPilot

DonLakeFlyer:
Fix for #5397


DonLakeFlyer:
Carry this to master once it passes

PR #5450, opened 2017-07-16
Allow custom overrides of windows installer icons

DonLakeFlyer:


DonLakeFlyer:
This needs to be merged to stable when done

PR #5460, opened 2017-07-17
Wind rose solution

patrickelectric:
This will not allow the scroll of PlanView while using wind rose

### DIFF
--- a/src/PlanView/SurveyItemEditor.qml
+++ b/src/PlanView/SurveyItemEditor.qml
@@ -737,7 +737,7 @@ Rectangle {
                 var angle = Math.round(Math.atan2(point.y, point.x) * 180 / Math.PI)
                 windRoseCanvas.requestPaint()
                 windRosePie.angle = angle
-                gridAngleText.text = - angle
+                gridAngleText.text = angle
                 gridAngleText.editingFinished();
             }
         }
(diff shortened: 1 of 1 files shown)


patrickelectric:
Need more work on it.

patrickelectric:
@DonLakeFlyer Finished.

DonLakeFlyer:
Why is it off for Mobile?

patrickelectric:
I thought that putting the rose only to desktop users make more sense, but I can add this to mobile too.

DonLakeFlyer:
I think it is way more useful on mobile where typing numbers for angle is hard, as opposed to desktop where typing numbers is way easier to do.

patrickelectric:
@DonLakeFlyer I think that everything is correct now.

DonLakeFlyer:
Still not quite right. As you can see from here 0 degrees shows up as 90 degrees:
screen shot 2017-08-08 at 11 46 27 am

Everything else seems to be working fine.

DonLakeFlyer:
Ah, I see.

DonLakeFlyer:
Let me take on more quick code review look and this is ready to go.

DonLakeFlyer:
Ok, all good. Thanks Patrick. Great new feature.

Issue #5436, opened 2017-07-14
Recorded video stream is not playable


dogmaphobic:
There is still some issue with the recorded video stream as a lot of video players can't play them, including Qt itself. QuickTime player won't play but VLC will play them though.

@jaxxzer Any clues?

I used this snipped to try it. It can play any other .mp4 file. It just won't play the ones recorded by QGC.

```
Video {
width: 1280
height: 720
source: "/path/to/recorded/video.mp4"
autoLoad: true
onErrorStringChanged: {
console.log(videoContent.error + ' ' + videoContent.errorString)
}
}
```

jaxxzer:
Is .mp4 the only one you are having trouble with? Do you have a qgc branch or project you can push so I can use that?

NaterGator:
Any samples of videos that fail in most players?

dogmaphobic:
>Is .mp4 the only one you are having trouble with?

No, it's not the container. The other containers have the same issue. Well, .mov and .mp4. VLC is the only thing that will play .mkv and VLC seems to handle the problem video just fine so I can't tell.

Try this: http://www.grubba.com/2017-07-14_17.56.55.mp4 (not the most interesting video but...)

NaterGator:
Looks like the decode timestamps are bogus:

```C:\Users\nweib\Videos>ffmpeg -i 2017-07-14_17.56.55.mp4 -f null -
ffmpeg version N-86789-g4c1aac8 Copyright (c) 2000-2017 the FFmpeg developers
  built with gcc 7.1.0 (GCC)
[...]

PR #5461, opened 2017-07-17
Add TCP-Mpeg2TS support to available video feeds

zthorson:
This adds Mpeg2 Transport stream over TCP support as a possible video feed in the gstreamer pipeline.

This is used in cases where a local ground station receives the video feed from the aircraft, and rebroadcasts it out to multiple ground stations or devices over a wireless link.   Since the network is local and broadcast to more than a single client, occasional packet drops between the ground station and devices occur, which TCP helps to mitigate.

It is not recommended to use TCP/MPEG2_TS as a datalink to directly to the aircraft, since the re transmits that occur over TCP can build up during a loss of link condition causing additional bandwidth overhead and latency issues.

### DIFF
--- a/src/FlightDisplay/VideoManager.cc
+++ b/src/FlightDisplay/VideoManager.cc
@@ -59,6 +59,7 @@ VideoManager::setToolbox(QGCToolbox *toolbox)
    connect(_videoSettings->videoSource(),   &Fact::rawValueChanged, this, &VideoManager::_videoSourceChanged);
    connect(_videoSettings->udpPort(),       &Fact::rawValueChanged, this, &VideoManager::_udpPortChanged);
    connect(_videoSettings->rtspUrl(),       &Fact::rawValueChanged, this, &VideoManager::_rtspUrlChanged);
+   connect(_videoSettings->tcpUrl(),        &Fact::rawValueChanged, this, &VideoManager::_tcpUrlChanged);
 
 #if defined(QGC_GST_STREAMING)
 #ifndef QGC_DISABLE_UVC
@@ -111,6 +112,13 @@ VideoManager::_rtspUrlChanged()
 }
 
 //-----------------------------------------------------------------------------
+void
+VideoManager::_tcpUrlChanged()
+{
+    _restartVideo();
+}
+
+//-----------------------------------------------------------------------------
 bool
 VideoManager::hasVideo()
 {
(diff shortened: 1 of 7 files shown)


DonLakeFlyer:
@dogmaphobic @jaxxzer Video folks, can you take a look for code review?

dogmaphobic:
I'm way too swamped to test this properly right now. @jaxxzer if you could take a look, that would help a lot. You might want to look into #5436 too :P

NaterGator:
Just doing a read through the code looks sane. I'm not entirely sure the pipelineUp change  / assuming the pipeline will unref everything is necessary but we'll just watch for leaks I guess.

The thing I'm less convinced of is that TCP is _ever_ appropriate for real time video streaming applications. The only thing you can accomplish here is accumulated latency in the pipeline with occasional skips to get back to realtime. For real time streaming I have never seen MPEG2-TS work more effectively over TCP than it does over UDP. 

I suppose there is no harm in having the option for TCP and I certainly like MPEG2-TS as a streaming container format, I just feel like we should be steering users towards UDP for a better experience.

zthorson:
@NaterGator 

The piplineUP change was done when I started getting a bunch of warning messages after the pipeline failed to link.  It would get the pipeline setup, then fail to link resulting in a teardown.  During the teardown, the logs would be flooded with warnings not to dereference the components directly, and instead let the pipline do it.  It is probably not strictly necessary, but removed some unhelpful log spam during certain fault conditions.

I agree that for a standard aircraft ---> qgroundcontrol that TCP is not the way to go due to the issues that you mentioned.   This was added mostly for the situation described below, where rebroadcasting data is done over a somewhat lossy link, resulting in higher than normal image tearing at the qgroundcontrol end.    For our particular application (fixed wing aircraft, not manually flown), reducing image tearing was given a higher priority than increased latency.


                        UDP Aircraft Feed                       Local Network (can be lossy)
aircraft ---[UDP|MPEG2TS|H264]---> receiver ----[TCP|MPEG2TS|H264]----> QgroundControl
                                                                          |---[TCP|MPEG2TS|H264]----> Mobile Video Player #1
                                                                          |---[TCP|MPEG2TS|H264]----> Mobile Video Player #2

DonLakeFlyer:
Going to wait a little bit more till Gus has some free time to look through this.

DonLakeFlyer:
@jaxxzer Did you have a chance to look through this?

jaxxzer:
I won't have the time for this for another week or so.

NaterGator:
@zthorson for the network topology you describe you'd be *much* better off using multicast if your networking equipment will cooperate.

zthorson:
@NaterGator

We had experimented with multicast in the past, and found that wifi implementation of multicast was spotty at best across different routers.  In many routers, broadcast of a multicast from a wired link through to the wireless clients was not implemented at all.

The TCP local rebroadcasting also helped to negate some interference issues that can occur in ground stations that are near other sources of 2.4Ghz interference (DJI transmitters, cell towers, etc)

DonLakeFlyer:
@dogmaphobic Gus are you clearer now?

DonLakeFlyer:
I did a code review through. Even though I don't know much about how video works it all seems to make sense and it's been sitting here forever so I'm going to merge this in.

PR #5474, opened 2017-07-19
New Mavlink GeoFence and Rally Point support

DonLakeFlyer:
This is a replacement for #5147. It has everything that had as well as:
* GeoFence/Rally point ui driven by capability bits

This should be fairly close to something that could go into daily builds.


bkueng:
Tested in SITL, seems to work well, except for the following:
- the currently active waypoint is not updated during a mission. I verified that the vehicle is sending the update.
- When starting QGC before SITL, I get a geofence transaction error. Seems to be a mavlink 1 vs 2 issue, when I merge https://github.com/mavlink/qgroundcontrol/pull/5233, it works.
- Are the capability bits respected? When testing agains SITL master, I get geofence upload & download errors. Also when the link is on mavlink 1, geofencing transactions must be disabled.

DonLakeFlyer:
> Are the capability bits respected?

I missed some places.

DonLakeFlyer:
@bkueng Can you give this one a try?

I haven't fixed active waypoint yet.

bkueng:
> @bkueng Can you give this one a try?

Done: upload & download works, when I do a 'Remove All', I still get 'Error during GeoFence communication with Vehicle: Mission remove all, maximum retries exceeded'.

DonLakeFlyer:
@bkueng GeoFence  remove all doesn't work because it returns back a MISSION_COUNT with a mission type instead of a geofence type:
```
VehicleLog: _missionLoadComplete requesting GeoFence
PlanManagerLog: "loadFromVehicle T:GeoFence read sequence"
PlanManagerLog: "_requestList T:GeoFence _planType:_retryCount" 1 0
PlanManagerLog: "_handleMissionCount T:GeoFence Incorrect mission_type received expected:actual" 1 0
PlanManagerLog: "_handleMissionCount T:GeoFence Incorrect mission_type received expected:actual" 1 0
PlanManagerLog: "_handleMissionCount T:GeoFence Incorrect mission_type received expected:actual" 1 0
PlanManagerLog: "Retrying T:GeoFence REQUEST_LIST retry Count" 1
```

DonLakeFlyer:
> the currently active waypoint is not updated during a mission. I verified that the vehicle is sending the update.

I just tested this. When I fly a mission I get this:
screen shot 2017-07-22 at 10 53 42 am
and according to the mavlink inspector I'm not getting any MISSION_CURRENT messages.

bkueng:
Are you sure you tested the latest branch? I pushed a fix for this quite a while ago: https://github.com/PX4/Firmware/pull/7238/commits/0d5f7f8d239ffa90b141bb2a59d105db54d0f706 in https://github.com/PX4/Firmware/pull/7238.

DonLakeFlyer:
Let me try again.

DonLakeFlyer:
With latest firmware I get this:
screen shot 2017-07-24 at 2 17 05 pm
As well as the geofence problem.

bkueng:
Most likely a mavlink 2 problem. You need to set `MAV_PROTO_VER` to 2 until #5233 gets merged. Use `mavlink status` on the console to check which protocol version is used.
Also, as I wrote above, we really should check if we're using mavlink 1 and disable geofencing & rally points in that case. If someone is using a link which does not support mavlink 2 he/she will run into the same issues.

DonLakeFlyer:
What is MAV_PROTO_VER? I've never seen that. I think the problem is when I tested I didn't turn the vehicle to mav 2 always.

DonLakeFlyer:
When I switch the vehicle to mavlink 2 it all works.

LorenzMeier:
We definitely need https://github.com/mavlink/qgroundcontrol/pull/5233 in for this because otherwise too many users will stumble over the settings.

DonLakeFlyer:
It's not quite #5233. It's just the code doesn't check for mavlink 2 running at all on the link.

LorenzMeier:
To summarise it quickly: MAVLink 1 is EOL, it is limiting the ongoing software development and we need to enable systems that support it to automatically switch to v2 and gracefully fall back to v1.

Since all PX4 releases setting the fence capability bit support MAVLink 2 the solution is not to check for protocol versions, but to enable both sides to switch to v2 without the user needing to fiddle with settings.

DonLakeFlyer:
I understand all of that but I still don't see how you don't need to check both the capability bits for fence/radio (which is implemented) as well as whether the link is running mav 2 (which is not implemented yet).

LorenzMeier:
@DonLakeFlyer with #5233 merged, should we rebase and get this in?

DonLakeFlyer:
I need to take a quick look again. This pull disables ArduPilot geofence/rally support since I need to rework it now. Want to look at how hard it's going to be to get that back in.

DonLakeFlyer:
Closing this for final new pull

PR #5480, opened 2017-07-21
FTP Fixes

bkueng:
This adds two fixes for FTP. I tested the full functionality repeatedly in SITL & on a Pixracer via USB and it works well, except when a message is dropped. In that case a transaction is immediately aborted, because of this:
https://github.com/mavlink/qgroundcontrol/blob/9af74e972bec5ec7616248e89c37b247e241ba28/src/uas/FileManager.cc#L326. We need to add retransmission to make this more robust.

Might be worth reverting https://github.com/mavlink/qgroundcontrol/pull/4144.

There are fixes on the vehicle side as well: https://github.com/PX4/Firmware/pull/7665

### DIFF
--- a/src/uas/FileManager.cc
+++ b/src/uas/FileManager.cc
@@ -239,6 +239,7 @@ void FileManager::_writeAckResponse(Request* writeAck)
 {
     if(_writeOffset + _writeSize >= _writeFileSize){
         _closeUploadSession(true /* success */);
+        return;
     }
 
     if (writeAck->hdr.session != _activeSession) {
--- a/src/ui/QGCUASFileView.cc
+++ b/src/ui/QGCUASFileView.cc
@@ -198,7 +198,7 @@ void QGCUASFileView::_commandComplete(void)
         _currentCommand = commandNone;
         _setAllButtonsEnabled(true);
         statusText = "Download complete";
-    } else if (_currentCommand == commandDownload) {
+    } else if (_currentCommand == commandUpload) {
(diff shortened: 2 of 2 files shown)


DonLakeFlyer:
I'm still not sure this is super useful without re-querying for missed blocks? What do you think?

bkueng:
I agree to make it really usable we need the re-query. I was thinking now that we're just past a release, we can enable it, mark as WIP, so that others might pick it up and help with the missing pieces.

Another thing: I was considering using FTP to push elevation data to the vehicle. @LorenzMeier and I think it's probably the easiest way to do, instead of having yet another protocol. What do you think?

LorenzMeier:
@DonLakeFlyer The data links we are interested in (long-range, digital) tend to not support all IP protocols. Even on Linux. So this is pretty appropriate even for next-generation systems. We would need to use this or something like UDT anyway.

DonLakeFlyer:
> Another thing: I was considering using FTP to push elevation data to the vehicle.

ArdiPilot already has an elevation data request system in mavlink. Is there a good reason to do something different? That will just make it harder on the QGC side.

LorenzMeier:
Yes, a very good one: it makes no sense to have one-off protocols for something that is always the same: transferring files.

DonLakeFlyer:
One persons file is another persons data. You could say the same thing for missions or parameters. Terrain data will need more of a protocol around it than just simple file transfer though. You can make life hard for QGC or easy to support in a generic way.

PR #5484, opened 2017-07-21
Fixups for building without gstreamer/videostreaming support

jaxxzer:
This removes the recording button from view when there is no gstreamer available.

### DIFF
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -43,6 +43,7 @@ QGCView {
     property var    _geoFenceController:    _planMasterController.geoFenceController
     property var    _rallyPointController:  _planMasterController.rallyPointController
     property var    _activeVehicle:         QGroundControl.multiVehicleManager.activeVehicle
+    property var    _videoReceiver:         QGroundControl.videoManager.videoReceiver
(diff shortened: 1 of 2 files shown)


src/FlightDisplay/FlightDisplayView.qml
```diff
@@ -370,14 +371,14 @@ QGCView {
...
-            visible:            QGroundControl.videoManager.videoReceiver.videoRunning && QGroundControl.settingsManager.videoSettings.showRecControl.rawValue
+            visible:            _videoReceiver && _videoReceiver.videoRunning && QGroundControl.settingsManager.videoSettings.showRecControl.rawValue
             opacity:            0.75
 
             Rectangle {
                 anchors.top:        parent.top
                 anchors.bottom:     parent.bottom
                 width:              height
-                radius:             QGroundControl.videoManager && QGroundControl.videoManager.videoReceiver && QGroundControl.videoManager.videoReceiver.recording ? 0 : height
+                radius:             QGroundControl.videoManager && _videoReceiver && _videoReceiver.recording ? 0 : height
```
DonLakeFlyer: That's a bit of an odd way to do it instead of using visible? Not quite as clear.

jaxxzer: Setting the radius to 0 makes it a square, setting the radius to height makes it a circle. It should always be visible when the parent is (when we have a video stream we can record).

DonLakeFlyer: Ah, ok.

PR #5486, opened 2017-07-21
Make recording status more obvious

jaxxzer:
We had a disappointing report of someone loosing a lot of footage they thought they were recording because they had accidentally set their hand on the touchpad or something and didn't notice they turned it off. In an effort to prevent these types of happenings, we are trying to make it more obvious when recording is active/inactive. This makes the recording button visibility toggle at 1Hz. I thought it might be distracting to do it this way, but I don't think it is after seeing it.


dogmaphobic:
There could be an option to start recording when armed (and stop when disarmed). That way you can't forget. I've done this so many times it's not even funny. Even on DJI's.

DonLakeFlyer:
@jaxxzer Can you rebase and I'll merge

jaxxzer:
Hi @DonLakeFlyer, I'm working with Daniel on this and will let you know when we are finished with it.

jaxxzer:
This is good to go on our end, rebase complete.

PR #5488, opened 2017-07-22
Re-enable FTP in menu

LorenzMeier:
This re-enables FTP now that Beat has fixed some underlying issues. The main motivation for it is that even high-bandwidth data links tend to only support UDP or even just a single data channel. Having a tunnelled, packetised file transfer is actually appropriate. MAVLink also ensures data integrity (via authentication in MAVLink 2) and therefore offers a viable solution.

The only caveat we need to add (@bkueng) is immediate re-transmission when something drops. It has to work on an unreliable link.

There are a number of use cases where having the ability to get / set files is powerful. This includes:

  * Terrain data
  * Camera meta data (we should reconsider going to JSON instead of XML @dogmaphobic)
  * Vehicle meta data (would avoid the common issue we're having right now with param meta)

It makes no sense to handle these one-off like with the log download tool.

### DIFF
--- a/src/qgcunittest/UnitTestList.cc
+++ b/src/qgcunittest/UnitTestList.cc
@@ -52,6 +52,7 @@ UT_REGISTER_TEST(MissionControllerTest)
 UT_REGISTER_TEST(MissionManagerTest)
 UT_REGISTER_TEST(RadioConfigTest)
 UT_REGISTER_TEST(TCPLinkTest)
+UT_REGISTER_TEST(FileManagerTest)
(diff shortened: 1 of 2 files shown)


dogmaphobic:
>Camera meta data (we should reconsider going to JSON instead of XML @dogmaphobic)

Json is not appropriate for this. It's meant for query/response of blobs of data, not for protocol definition. Camera Definition is more than listing parameters. It defines rules, exceptions, localization, etc.

What could be done, if data size is an issue, is to gzip it for transfer. With that said, at ~30k these files are not exactly large.

Issue #4267, opened 2016-11-30
KML Parser and Exporter


saitej09:
Hi

Are you guys gonna include a KML Parser to load KML objects on to the Map and also for exporting the map missions in to KML?
Thanks

PR #5499, opened 2017-07-25
Add basic KML functionality

patrickelectric:


patrickelectric:
This will solve #4267

patrickelectric:
selection_536
selection_534
selection_535

patrickelectric:
selection_050

DonLakeFlyer:
Some changes:
* Can you move KML code into MissionManager (not quite right, but better than at root)
* Can you follow the pattern for json save where PlanMasterController only creates/saves the outer object and the code for the real object saving is in the appropriate controller. Then move the KML save code to MissionController instead of PlanMasterController. Then you don't need to expose the load and convert items to public from MissionController.
* Can you Save KML... to the last item in the sync popup. That keeps the upload/download, save/load pattern for this first items which I think makes more sense.

Let me me know if any of this doesn't make sense.

DonLakeFlyer:
There are also build breaks in the code

patrickelectric:
[x]-move KML code into MissionManager (not quite right, but better than at root)
[]-follow the pattern for json save where PlanMasterController only creates/saves the outer object and the code for the real object saving is in the appropriate controller. Then move the KML save code to MissionController instead of PlanMasterController. Then you don't need to expose the load and convert items to public from MissionController.
[]-Save KML... to the last item in the sync popup. That keeps the upload/download, save/load pattern for this first items which I think makes more sense.

patrickelectric:
@DonLakeFlyer Now is better ?

DonLakeFlyer:
Yup, that looks like what I was talking about. Let me take another detailed look through and then should be good to merge.

src/MissionManager/MissionController.cc
```diff
@@ -256,6 +257,49 @@ bool MissionController::_convertToMissionItems(QmlObjectListModel* visualMission
...
+    save(missionJson);
+    _loadItemsFromJson(missionJson, visualItems, error);
+    _convertToMissionItems(visualItems, missionItens, this);
+
+    float altitude = missionJson[_jsonPlannedHomePositionKey].toArray()[2].toDouble();
+
+    QString coord;
+    QStringList coords;
+    for(const auto& item : missionItens) {
+        switch(item->command()) {
```
DonLakeFlyer: This code should be change to use the MissionCommonUIInfo data which has meta data for each mission command:
```
    const MissionCommandUIInfo* uiInfo = qgcApp()->toolbox()->missionCommandTree()->getUIInfo(_controllerVehicle, item->command());
    if (uiInfo && uiInfo->specifiesCoordinate() && !uiInfo->standaloneCoordinate())  {
        // Add KML for a mission item which specifies a coordinate that the vehicle flies through
    }
```

DonLakeFlyer:
Also there is still a build break.

patrickelectric:
oh, problems in debug mode, will solve it today.

patrickelectric:
@DonLakeFlyer Now is it ok ?

src/MissionManager/MissionController.cc
DonLakeFlyer: This still needs to be fixed.

patrickelectric: Okey dokey, sorry for missed that.

patrickelectric: @DonLakeFlyer Done.

patrickelectric:
@DonLakeFlyer we can merge that ?

DonLakeFlyer:
I tried to save a KML file, but when I loaded back into Google Earth it did zoom into the right location but no lines showed up. Is there something else I need to do in Google Earth.

DonLakeFlyer:
Nevermind, I figure out how to get lines to show. I think there is an altitude calculation problems somewhere, still looking.

DonLakeFlyer:
[KMLTest.plan.zip](https://github.com/mavlink/qgroundcontrol/files/1285546/KMLTest.plan.zip)

Try loading this plan file without a vehicle connected and save to KML. I think the home position altitude is being calculated incorrectly, or maybe the takeoff item. The first point is way up in the air.

patrickelectric:
@DonLakeFlyer no problem here to open it.
image


Are you doing the same thing that this gif ?
https://media.giphy.com/media/l1J9A3v2CMmmFQPza/giphy.gif

patrickelectric:
Now, it appears that have a problem with Home position, ill take a look.

patrickelectric:
image

patrickelectric:
Can you check now @DonLakeFlyer ?

DonLakeFlyer:
Yup, all good now thanks.

PR #5501, opened 2017-07-26
Handle joysticks with duplicate names

jaxxzer:
Also handle a segfault issue that has been plaguing me for a while by commenting out call to SDL_JoystickClose

### DIFF
--- a/src/Joystick/JoystickSDL.cc
+++ b/src/Joystick/JoystickSDL.cc
@@ -145,7 +145,10 @@ void JoystickSDL::_close(void) {
 
         if (SDL_JoystickInstanceID(sdlJoystick) != -1) {
             qCDebug(JoystickLog) << "\tID:" << SDL_JoystickInstanceID(sdlJoystick);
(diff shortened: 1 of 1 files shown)


hamishwillee:
@jaxxzer Does this have a docs impact? 
Specifically, it sounds like you can have multiple joysticks connected, and that this change allows that to work. However we don't have any information about multiple joysicks [here](https://docs.qgroundcontrol.com/en/SetupView/Joystick.html).

Can you help me understand this?

jaxxzer:
We have allowed multiple joysticks to be connected for quite some time. We are always using only one joystick at any time. The joystick we are using is selected from the dropdown on the joystick page. This change handles the case in which two joysticks with the same name identifier are connected, in order to allow selecting between the two, instead of overwriting the first instance while we enumerate the second and then ending up with only one option in the dropdown.

jaxxzer:
IMO, this does not have a docs impact. @hamishwillee

hamishwillee:
Thanks @jaxxzer

PR #5502, opened 2017-07-26
Add joystick button limit

dheideman:
Limits the maximum number of joystick buttons for which functions can be added on vehicle settings, joystick setup page, to 16.

Prevents issues of unconfigurable buttons outside the range 0~15 from appearing to be configurable, such as: https://cdn-business2.discourse.org/uploads/bluerobotics/original/2X/8/886612ffc44d2d46447e44976e857bec158ec7f3.png

Closes bluerobotics/qgroundcontrol#21

### DIFF
--- a/src/VehicleSetup/JoystickConfig.qml
+++ b/src/VehicleSetup/JoystickConfig.qml
@@ -27,6 +27,8 @@ SetupPage {
     pageName:           qsTr("Joystick")
     pageDescription:    qsTr("Joystick Setup is used to configure a calibrate joysticks.")
 
+    readonly property real _maxButtons: 16
+
     Connections {
         target: joystickManager
         onAvailableJoysticksChanged: {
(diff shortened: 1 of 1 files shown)


hamishwillee:
@dheideman @DonLakeFlyer Is this just sub, and in what version will this appear? Probably a useful note to add to the [joystick docs](https://docs.qgroundcontrol.com/en/SetupView/Joystick.html)

dheideman:
This was written for sub but should also apply to any other systems using joysticks in a similar fashion.

So yes, just sub.

jaxxzer:
@hamishwillee The constraint is the 16 bit field in the [MANUAL_CONTROL](http://mavlink.org/messages/common#MANUAL_CONTROL) message where the button values are sent (to all types of vehicles). So this should affect all vehicles.

hamishwillee:
Thanks guys. So would it be reasonable to add the following comment to the docs (for the current version/now)

"Up to 16 joystick buttons can be assigned to vehicle functions (If more than 16 button checkboxes are displayed only the first 16 will be used)"

jaxxzer:
@hamishwillee That would be correct for 3.2. After this merge, only 16 max will ever appear for assignment.

hamishwillee:
Thanks @jaxxzer . I ended up adding the following note "Select the flight modes/vehicle functions activated by each joystick button. A maximum of 16 joystick *button actions* can be set.". 

This works for all versions, because even if more are displayed at least it is clear why button 17+ does not work.

PR #5504, opened 2017-07-27
Message ID buffer overrun

dogmaphobic:
There is some code in UAS.cc that maps messages IDs to components. It used to use an array of 256 elements, which assumed messages ID were an uint8_t. With MAVLink V2, messages are now 24 bits. New messages arriving with an ID > 255 would cause a buffer overrun.

I replaced the hardcoded array with a map. Given that I'm not sure what this code does, it would not hurt for @DonLakeFlyer to take a look.


NaterGator:
A QHash is probably more appropriate since storage order isn't important, but I think we can deal with the overhead for now ;)

dogmaphobic:
I "assumed" QMap was faster when searching as the bulk of the calls are searching for messages.

In quotes because I did not test for speed. Just assuming it based on logic (which could as well be faulty).

NaterGator:
Not a huge deal; the Qt container class performance is summarized [here](http://doc.qt.io/qt-5/containers.html#algorithmic-complexity).

DonLakeFlyer:
I've been looking at this code since I've never seen it before! It's overkill to track every message that comes through since it is only use for two specific messages. I'll fix it up to make it work simpler.

PR #5508, opened 2017-07-29
Mission/Survey related fixes

DonLakeFlyer:
* Default mission altitude is respected on newly added Surveys altitude is the fixed value
* If the user applies a new altitude to the entire mission any Survey in the mission will be switched to Altitude and have the altitude applied. This is a safety thing, where the user is trying to raise the height of a mission to get above an obstacle. The Survey needs to respect that as well.
* Child indicators for mission items are show in darker color
screen shot 2017-07-29 at 11 27 09 am


DonLakeFlyer:
Forgot one:
* Tweaked the text colors for disable text fields to make them more readable

PR #5513, opened 2017-07-30
Plan: Terrain display in MissionItemStatus

DonLakeFlyer:
This is a WIP so expect some quirks, also the implementation is not super performant from a network traffic standpoint. Uses AirMap terrain apis.

Red terrain indicator shows flight below terrain:
screen shot 2017-07-30 at 12 41 53 pm


kikislater:
Nice ! 
Just curious, why to choose this api ? I didn't find their sources for these elevation data. Do you know if it's srtm 30 or 90 ?

DonLakeFlyer:
We are doing other work with AirMap, this is just one of the things. Not sure of the data set I'll get info for that when I update the release docs.

PR #5518, opened 2017-08-01
MAVLink decoder: Run at lowest priority

LorenzMeier:
There is no reason to run it at any other priority, since its only for UI / inspection.


DonLakeFlyer:
Can the charts keep up at lowest priority?

PR #5520, opened 2017-08-02
Internet Test

dogmaphobic:
* Switch test from `8.8.8.8:53` to `www.github.com:80`
 * Route Internet test function to main app.

```
bool QGCApplication::isInternetAvailable()
```

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -668,3 +668,8 @@ void QGCApplication::qmlAttemptWindowClose(void)
 {
     QMetaObject::invokeMethod(_rootQmlObject(), "attemptWindowClose");
 }
+
+bool QGCApplication::isInternetAvailable()
+{
+    return getQGCMapEngine()->isInternetActive();
+}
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Ooh and global api too. Awesome.

anassinator:
Just curious:

Are there still any governments blocking https://github.com? It's been a reoccurring thing (https://en.wikipedia.org/wiki/Censorship_of_GitHub). Is this a worry?

Also, I wouldn't expect it to matter, and not completely sure how HSTS and HTTPS redirects work, but would port 443 be better in any way instead of port 80?

dogmaphobic:
> Also, I wouldn't expect it to matter, and not completely sure how HSTS and HTTPS redirects work, but would port 443 be better in any way instead of port 80?

Port 80 will always be available. Note that I am not *GET*ing anything. It's a simple TCP connection and no data is transferred. If there are *Great Firewalls*, it would be on port 443.

anassinator:
Ah ok! Thanks for clearing it up 🙂

dogmaphobic:
@DonLakeFlyer This has been merged to master. You can merge the `internetTest` branch to Stable_V3.2 whenever you want (may wait until folks here have a chance to test it).

DonLakeFlyer:
Can folks in this discussion verify that the newest daily build works for you and then I'll merge to Stable.

DonLakeFlyer:
Yup

PR #5525, opened 2017-08-03
Add reload functionality for vehicle setup panels

jaxxzer:
This addresses a long-standing issue described [here](https://github.com/bluerobotics/qgroundcontrol/issues/19).


jaxxzer:
I think this needs a little bit of rework. I'll reopen when ready.

PR #5548, opened 2017-08-11
New Mavlink 2 spec GeoFence and Rally Point support

DonLakeFlyer:
screen shot 2017-08-10 at 4 02 40 pm

Note: ArduPilot does not support the new GeoFence/Rally Point spec. With this pull ArduPilot support for GeoFence/Rally is temporarily disabled until I work it back into this new scheme.


DonLakeFlyer:
Note: PX4 support for this is only in master firmware.

PR #5561, opened 2017-08-12
Offline map: Use FlightMap as base

DonLakeFlyer:
This way:
* GCS location is shown
* It uses centering semantics of FlightMap: centere same as flight map , center to gcs locaiton if available
* Center on Location works
* Note: Vehicle is not shown

Fix for #5554

### DIFF
--- a/src/QtLocationPlugin/QMLControl/OfflineMap.qml
+++ b/src/QtLocationPlugin/QMLControl/OfflineMap.qml
@@ -331,12 +331,14 @@ QGCView {
         id:                 panel
         anchors.fill:       parent
 
-        Map {
-            id:                 _map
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@dogmaphobic Do you see any problems with this? I use FlightMap now as the base for all maps to get consistent behavior.

dogmaphobic:
There should not be any difference. Note that offline maps and FlightMap got out of sync a while back when map types went into settings manager (and the Json metadata). Offline maps uses the map manager and displays a list of a lot more map types that can be used by FlightMap (which is limited to only what is listed in the Json file). At some point we need to reconcile those.

PR #5565, opened 2017-08-13
Split internet test into separate DNS lookup and connection.

dogmaphobic:
To test if you have Internet connection, the code used to test a connection to 8.8.8.8:53 (google DNS). It appears that some routers are now blocking TCP connections to port 53. So instead, we started using a TCP connection to "github.com" (80). The problem with using a host name rather than an IP address is that the code must first do a name look up before it can attempt a connection. The lookup itself may take up to a minute on some devices. If you were to attempt to exit QGC while that thread was attempting a connection, it would lock the thread until the call returned. There is no way to cancel a connection attempt.

This PR splits the DNS lookup from the connection. The lookup, unlike the "look up and connect" call, can be canceled.

Why bother with the connection if the lookup succeeded? Because you may have had a connection and no longer have. The OS would have cached the lookup. The connection proves that you are indeed connected to the Internet.

This PR is a branch of Stable_3.2. It can be merged into it directly (as it is setup to do when done) as well as master.

### DIFF
--- a/src/QtLocationPlugin/QGCTileCacheWorker.cpp
+++ b/src/QtLocationPlugin/QGCTileCacheWorker.cpp
@@ -52,6 +52,7 @@ QGCCacheWorker::QGCCacheWorker()
     , _defaultCount(0)
     , _lastUpdate(0)
     , _updateTimeout(SHORT_TIMEOUT)
+    , _hostLookupID(0)
 {
 
 }
--- a/src/QtLocationPlugin/QGCTileCacheWorker.cpp
+++ b/src/QtLocationPlugin/QGCTileCacheWorker.cpp
@@ -73,6 +73,9 @@ QGCCacheWorker::setDatabaseFile(const QString& path)
 void
 QGCCacheWorker::quit()
 {
+    if(_hostLookupID) {
+        QHostInfo::abortHostLookup(_hostLookupID);
+    }
     _mutex.lock();
     while(_taskQueue.count()) {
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Thanks Gus. I'll merge across to master when done.

dogmaphobic:
Hmm, unit test is failing. Let me take a look at that first...

dogmaphobic:
Unit tests passed just fine (mac OS). Running those builds again to see if something changes (too lazy to setup a Linux build environment).

DonLakeFlyer:
Some kind of shutdown problem since all unit tests are finished and passed.

dogmaphobic:
Right, which is exactly the problem this "fix" is supposed to address. This doesn't happen on macOS though. I will take a look at it tomorrow.

dogmaphobic:
Alright, don't know what else is wrong with this. I run a build on Linux (Ubuntu 16.04) and unit tests finished fine with no errors or crashes:

```
********* Finished testing of QGCMapPolygonTest *********
ALL TESTS PASSED
Done
After app delete
Debugging has finished
```

NaterGator:
That last commit fixed it, the travis mac builder just ran out of time because it is extremely slow.

dogmaphobic:
Ha! Thanks for pointing that out. I didn't even think to look. I just assumed it was still failing. Yes, Mac builds only succeed at 3AM on Sundays :P

PR #5568, opened 2017-08-14
New PageView control in instrument panel

DonLakeFlyer:
Reworked intrument panel page ui:
* New PageView control which supports this page ui. Also handles each page having the ability to require a settings button in a standard way.
* To make it more obvious that there are multiple pages available
* To make it more obvious as to how to switch pages
* Custom builds can now override/add/remove pages
* The special handling of Unhealthy Sensors is no longer there. This is also a trial. I figure you will get notice from the Vehicle with a message. Do you really need to hit people over the head with more than that?
* Always use smaller alternate instrument panel for all builds. This is a trial for now. I'll be doing more work on this, possibly allowing the ability to size it.

screen shot 2017-08-14 at 1 35 36 pm


DonLakeFlyer:
@dogmaphobic I forgot to mention that I changed QGCSettings to QmlPageInfo. Other than the more generic name to make is usable for more than Settings pages it is exactly the same. So when any custom that is using QGCSettings pulls from master it will just need to update the name to QmlPageInfo.

dogmaphobic:
Yes, it will be fun when I get to merge cameraControl back to master :P

DonLakeFlyer:
This pull shouldn't have any specific effect on that I think. You should be able to just take over CameraPageWidget for the new camera control stuff.

PR #5575, opened 2017-08-15
MockLink: Simulate movement of ADSB vehicle

DonLakeFlyer:


hamishwillee:
How will this work? According to [the docs](https://dev.qgroundcontrol.com/en/tools/mock_link.html) Mocklink doesn't support flight. Upshot, what extra will need to be said in the Mocklink doc.

DonLakeFlyer:
The simulated ADSB vehicles just circle around the vehicle. No big deal. I don't think anything specific needs to be documented about it.

hamishwillee:
Thanks. Sounds reasonable.

PR #5578, opened 2017-08-16
Level Horizon Indicator

enginc:
Level horizon green Indicator fix


DonLakeFlyer:
Thanks for the pull but it was not created correctly as you can see from the list of changes. Looks like you pulled master across with your pull to Stable.

DonLakeFlyer:
I'm closing this since it is very out of whack

PR #5579, opened 2017-08-16
Add wind direction pointer

patrickelectric:
Add wind direction pointer in:
 - QGCCompassWidget
 - CompassRing

### DIFF
--- a/qgcresources.qrc
+++ b/qgcresources.qrc
@@ -113,6 +113,7 @@
         <file alias="scale_end.png">src/FlightMap/Images/scale_end.png</file>
         <file alias="scaleLight.png">src/FlightMap/Images/scaleLight.png</file>
         <file alias="scale_endLight.png">src/FlightMap/Images/scale_endLight.png</file>
+        <file alias="windPointerGreen.svg">src/FlightMap/Images/windPointerGreen.svg</file>
         <file alias="adsbVehicle.svg">src/FlightMap/Images/adsbVehicle.svg</file>
         <file alias="vehicleArrowOutline.svg">src/FlightMap/Images/vehicleArrowOutline.svg</file>
         <file alias="vehicleArrowOpaque.svg">src/FlightMap/Images/vehicleArrowOpaque.svg</file>
--- a/qgcresources.qrc
+++ b/qgcresources.qrc
@@ -113,6 +113,7 @@
         <file alias="scale_end.png">src/FlightMap/Images/scale_end.png</file>
         <file alias="scaleLight.png">src/FlightMap/Images/scaleLight.png</file>
         <file alias="scale_endLight.png">src/FlightMap/Images/scale_endLight.png</file>
+        <file alias="windPointer.svg">src/FlightMap/Images/windPointer.svg</file>
         <file alias="windPointerGreen.svg">src/FlightMap/Images/windPointerGreen.svg</file>
         <file alias="adsbVehicle.svg">src/FlightMap/Images/adsbVehicle.svg</file>
         <file alias="vehicleArrowOutline.svg">src/FlightMap/Images/vehicleArrowOutline.svg</file>
--- a/src/FlightMap/Images/windPointer.svg
+++ b/src/FlightMap/Images/windPointer.svg
@@ -1,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Generator: Adobe Illustrator 18.1.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+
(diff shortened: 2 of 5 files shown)


patrickelectric:
selection_060
selection_061

DonLakeFlyer:
This seems like a good idea but I"m concerned around confusion as to what is the vehicle compass direction and which is the wind direction.

patrickelectric:
@DonLakeFlyer in both cases ?

DonLakeFlyer:
Mostly in the combined compass case, but that is the default. @anitheavy Wondering if you seem some good way to do this?

DonLakeFlyer:
@antiheavy Any ideas here?

Antiheavy:
nice topic!  Sorry I missed your first ping 18 days ago (spelling goof).  It is a wonderful and important improvement to add a wind graphic.

Wind direction graphics I've seen in the past usually combine wind speed into the direction arrow somehow.  I think it would be much more useful if the operator can see direction and speed in a single glance.

I don't like the green and red indicator arrow in the combined artificial horizon.  It is too confusing and too small to really convey the meaning.  (I'd like the red compass arrow in the combined artificial horizon to be much more prominent too, but that is a different topic).

If there isn't space to show a good graphic in the artificial horizon or compass area, maybe a dedicated graphic somewhere else is acceptable?

I'm not the worlds most skilled graphic artist, but I will spend a few minutes later today/tomorrow trying to sketch up some ideas.  This is an important feature.

DonLakeFlyer:
Maybe a dedicated graphic for fixed wing?

Antiheavy:
> Maybe a dedicated graphic for fixed wing?

I think it is more than just a Fixed Wing thing.  I know the VTOL firmware guys are working on wind estimation for multi-copters.  Hybrid vehicles like quad-planes like to know about wind too.

DonLakeFlyer:
Yeah, I lump those two together.

Antiheavy:
here are some ideas of things I've seen in the past on other ground stations.  I'm suggesting any of these are better for QGC, just other examples.

image

I like patrickelectric's idea to combine into the widget vs a stand-alone graphic somewhere... still trying to thing of ideas how to simplify/clarify it more...

Antiheavy:
so sticking with patrickelectric's original concept, I like the large compass version, i like that the green triangle/arrow is behind the white compass pips, although i think it should be brighter and more visible.  also I think the wind speed should be overlayed somewhere and made obvious, similar to how the heading value is displayed in the middle of the compass.  here are a couple ideas:
image

Antiheavy:
also, i feel like the roll indicator on the artificial horizon is confusing.  I realize this is off topic (maybe i'll make a separate issue post about this).  anyway, thinking about the wind indicator made me think of it.  see here:
image

Antiheavy:
here are some ideas for the combined artificial horizon and compass with wind speed indicator added:

Clean up the roll indicator, add colored text to make it clear the green arrow is wind and what magnitude:
image

Also, the red compass arrow always bothered me because I think it points the wrong way (should point the direction of the front of the aircraft i think).  Here I changed it to orange, and also made the heading magnitude text orange to make it clear that number goes with the orange arrow (seems we should avoid red text as that indicates a warning/problem):
image

I do not like the way the orange heading text looks crowded inside the horizon, so i moved it and wind speed down to the bottom corners.  the gear goes away because the whole dang thing is a button anyway:
image

of course the next logic step is to add altitude and speed to the other corners:
image

I dunno, this are just ideas.  what do you guys think?

DonLakeFlyer:
Thanks Todd. Helpful. I need to think through this also with respect to the hacking around I've been doing on instrument display in daily builds.

fredrikfalkman:
How about using meteorological wind arrows on the border of the compass? They are the standard way to represent both wind direction and wind speed and would offer the combined info at a glance.

http://tgftp.nws.noaa.gov/fax/PLBZ08.gif

moreba1:
The speed and direction of the wind can be on the left side of the screen  or Maybe it's okay to be near of the position display .

patrickelectric:
Just to update this PR, I'll not continue this :)
If no one will take it, this PR can be closed.

DonLakeFlyer:
Still haven't had time to figure out what to do here. Stacked up behind:
* Getting Stable 3.3 done
* Being out of town forever...

I'll get back to this. Leave it here for now.

tstastny:
Just chiming in -- I would vote for something close to @Antiheavy 's green arrow (color could of course be different). Looks very nice and simple, can easily see if the vehicle is facing into or going with the wind. I would for sure be available to pop out quickly and flight test this whenever there is a PR.

PR #5463, opened 2017-07-18
Initial mav 2 signing support [WIP - DO NOT MERGE]

DonLakeFlyer:
Completely untested. Trying to open up testing of: https://github.com/PX4/Firmware/pull/7530.

You can turn on mav 2 signing and provide your key from the mavlink settings page.

Keep in mind that QGC still does not negotiate mavlink 2 on/off based on capability bits. It just turns on mav 2 if it gets incoming mav 2. So to test with this pull you'll need to turn on mav 2 from the vehicle side first.

@simonegu, @LorenzMeier  I only sort of understand the instructions for implementing mav 2 signing. So first thing would be a code review of this to see if it makes any sense.

PR #5582, opened 2017-08-17
Mavlink2 signing support [WIP - Do not merge]

DonLakeFlyer:
Replacement for #5463


LorenzMeier:
@DonLakeFlyer The setting is only applied after restarting QGC it seems.

DonLakeFlyer:
@dogmaphobic This is somplace else now right?

PR #5586, opened 2017-08-17
Survey fixes

DonLakeFlyer:
* Camera shot calculation had multiple off by ones
* Hover and Capture, images in turnaround options missing from Manual Grid


DonLakeFlyer:
Show be against stable

PR #5588, opened 2017-08-17
Allow disconnect/reconnect of RTK GPS

DonLakeFlyer:
@bkueng Can you code review this?


DonLakeFlyer:
If I remember correctly there is some port close funkiness on windows with respect to just pulling the plug. Hence the code in LinkManager to do it the way it does.

Issue #5577, opened 2017-08-16
RTSP streaming do not work when QGC has no internet access


SaliniVenate:
I'm trying to stream a video from drone to QGC running on ubuntu desktop over RTSP. The ubuntu machine is connected to drone over a wifi hosted by drone. Every time the ubuntu machine is disconnected from internet the QGC is not able to receive the video stream. However when I run vlc or gstreamer pipeline to receive the same video stream , it works. 
For video streaming should QGC be dependent on internet access?

DonLakeFlyer:
You must have something wrong in your setup. There is no depedancy on internet for video that I know of. 

@dogmaphobic Does this make any sense to you?

dogmaphobic:
No, it doesn't. The video connection is peer-to-peer with no dependency on anything else. 

I'm curious though, what is the URL used?

SaliniVenate:
The url used is : rtsp://192.168.8.1:8554/video13. This RTSP server is running in Intel Aero companion board.

1. With Internet : QGC is able to receive video streamed by drone
2. No Internet : QGC does not receive the video stream. Restarting the video (i.e choosing Video Stream Disabled and then choosing RTSP Video Stream in Settings) does not help. However VLC or gst pipeline is able to receive the video stream.
3. Internet disconnected while Video is streaming: QGC continues to receive video without disruption

The video manager and video reciever logs with internet disconnected is attached.
[logs_internet_disconnected.txt](https://github.com/mavlink/qgroundcontrol/files/1230194/logs_internet_disconnected.txt)

PR #5611, opened 2017-08-28
Set NoProxy for the TCP Socket used for Videostreaming

SaliniVenate:
QGC is not able to receive the RTSP video stream from the drone, when the Ubuntu machine which runs QGC is disconnected from the Internet. This issue can be resolved by setting no-proxy for the TCP socket used for video-streaming. Please refer the issue #5577 for details.

### DIFF
--- a/src/VideoStreaming/VideoReceiver.cc
+++ b/src/VideoStreaming/VideoReceiver.cc
@@ -181,6 +181,7 @@ VideoReceiver::_timeout()
     //   found to be working, only then we actually start the stream.
     QUrl url(_uri);
     _socket = new QTcpSocket;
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@dogmaphobic @jaxxzer Is this ok?

dogmaphobic:
Yes, that's fine. It's the same done for UDP links:
https://github.com/mavlink/qgroundcontrol/blob/master/src/comm/UDPLink.cc#L250
Not sure why the lack of Internet would cause a proxy to take over but the change is benign.

jaxxzer:
I can't lend any insight here. Interesting/good to know though.

DonLakeFlyer:
Great, thanks everyone.

Issue #5535, opened 2017-08-07
APMRover : guided not supported


khancyr:
Hello,

APMRover support guided command but QGC said it isn't support

DonLakeFlyer:
Should be better now

Issue #5187, opened 2017-05-23
ArduPlane: Add guided tools support


DonLakeFlyer:

DonLakeFlyer:
Fixed in #5652

PR #5636, opened 2017-09-06
Better Guided Action support for ArduPilot

DonLakeFlyer:
This centralizes common guided action support into APMFirmwarePlugin such that it does not need to be repeated for each vehicle type (Copter/Plane/Rover). The following guided actions are now supported across all vehicles and the left side action tools work for these: 
* RTL
* Pause - Puts the vehicle in the appropriate Pause flight mode
* GoTo Location - Click in map to go to location
* Change Altitude - Available from Action tool

@khancyr This is going to clash with your pull #5597 since it centralizes most guided support. Hence there is no longer need for specific Rover support. Once this goes through you'll need to rebase and fix up. Should be pretty easy. Mostly just pulling stuff out.

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -7,10 +7,6 @@
  *
  ****************************************************************************/
 
-
-/// @file
-///     @author Don Gagne <don@thegagnes.com>
-
 #include "APMFirmwarePlugin.h"
 #include "APMAutoPilotPlugin.h"
 #include "QGCMAVLink.h"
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
Fix for #5535, #5187

DonLakeFlyer:
@tridge I use set_position_target_local_ned to change altitude. This works for Copter, but doesn't seem to work for Plane.

DonLakeFlyer:
@tridge I'd also like to be able to start a mission in some standard way. For Copter I take the vehicle off to 1 meter and then change to Auto. Not sure how to do the same with Plane or Rover. Is there some standard way I can do this, or I do I need seperate implementation for each?

khancyr:
@DonLakeFlyer , ok no problem. 
I think that plane don't support set_position_target for now ... 
About Rover, a mission start is just go to Auto and arm, or arm and go auto , the mission should start.
About Copter, I would say that takeoff to at least 2m is better ,@rmackay9, any advice ?

Issue #4328, opened 2016-12-20
Crash on android when direct usb connection to board


icesword001:
On android, In the beginning，connection QGC is successful, everything is all right; 
However, after waiting for tens of seconds to minutes, the QGC would Crash; 
I found, debugging is a collapse in the "SerialLink::_readBytes()" class, but the reasons is unknown

[Fatal] 16:03:24.206 ASSERT: "size == 0 || offset < 0 || size_t(offset) >= sizeof(QArrayData)" in file E:\Qt\Qt5.5.1_android\5.5\android_armv7\include/QtCore/qarraydata.h, line 54 File:() Line:(0)
@LorenzMeier

DonLakeFlyer:
Are you using an Android daily build? There was a crash after some amount of time was fixed in Daily.

icesword001:
I use the QGroundControl v3.0.2  package（from github）; 
There are the same problems.

DonLakeFlyer:
3.0.2 is a stable buid not a daily build. To get daily build you need to sign up for the android beta: https://donlakeflyer.gitbooks.io/qgroundcontrol-user-guide/content/download_and_install.html

DonLakeFlyer:
@icesword001 I'll re-open if it is still crashing on a Daily Build.

DonLakeFlyer:
Retested on daily build. Working fine

JonReacher:
@DonLakeFlyer I'm still having this issue on a stable build for months, even to the latest 3.2.2. Haven't ever been able to connect for more than a 10 seconds directly to the board.

LorenzMeier:
@dogmaphobic maybe related to what you found?

mhkabir:
Can confirm that this is still an issue.

dogmaphobic:
I found a nasty bug when QGC receives a message with a message ID > 255. It causes buffer overrun and all things go haywire after that. The behavior is random as the affected places change based on the message and platform. I will push the fix shortly.

DonLakeFlyer:
I can reproduce this. But I'm having probelsm with my tablet where I can't get it to go into wireless debugging mode to be able to catch the crash.

birchera:
[...]

PR #5637, opened 2017-09-06
Fix in serial communication on Android

birchera:
Fixes https://github.com/mavlink/qgroundcontrol/issues/4328.

**Proposed changes**:
* Replaces the QGCRingBuffer class with the QRingBuffer class in the serial communication module.
* This solves a bug in memory access when reading (could not figure out actual bug though).

@DonLakeFlyer can you have a look at it?


DonLakeFlyer:
Does this fix the USB crash?

birchera:
Yes, it does for us.

DonLakeFlyer:
Sorry I'm an idiot. Missed it.

DonLakeFlyer:
This is more @dogmaphobic  since he brought across the android serial stuff.

DonLakeFlyer:
Once this is ok'ed this needs to go into Stable_V3.2 as well.

dogmaphobic:
The change seems logical. I don't remember where this came from originally. Serial connection under Android has always been sort of a bastard child any way. I really don't like that Java spaghetti mixed in. At some point that whole thing should be replaced with real code.

DonLakeFlyer:
Ok, great. Big thanks @birchera for this fix. I'm going to use this to get a Stable_V3.2 pull as well before I merge this in.

PR #5652, opened 2017-09-13
Guided takeoff

DonLakeFlyer:
* Rework guided takeoff to be a capability bit
* Restructure ArduPilot guided takeoff support to work for vtol


DonLakeFlyer:
Note: Due to the way ArduPlane identifies VTOL vehicles I've had to enable Takeoff for all fixed wing aircraft. If you try to use it on a non-vtol aircraft it will show an error telling you it is not supported.

PR #5653, opened 2017-09-14
Allow resizing pip window

jaxxzer:
screenshot 2017-09-07 11 15 54
screenshot 2017-09-07 11 18 22
screenshot 2017-09-07 11 19 26


dogmaphobic:
Have you tried this on a mobile device? I wonder if those buttons will even fit.

jaxxzer:
I have not, this might be ok for tablet sized devices, but maybe we should disable it for mobile.

DonLakeFlyer:
YOu can base on screen size on mobile. There are a couple flash short and tiny or something like that.

jaxxzer:
Ok, I can work on the mobile logic if @dogmaphobic @DonLakeFlyer are into the concept.

src/QmlControls/QGCPipable.qml
```diff
@@ -24,6 +24,8 @@ Item {
 
     signal  activated()
     signal  hideIt(bool state)
+    signal  bigger()
```
jaxxzer: maybe shrink() and grow() are better names.

DonLakeFlyer:
@dogmaphobic No resizing for ScreenTools.isShortScreen * ScreenTools.isTinyScreen?

DonLakeFlyer:
Some problems:
1) The buttons are created with fixed pixel width/height. This means you have no idea how large they will really be on various devices. Everything must be font based or ScreenTools.minTouchPixels.
2) Even if you fix 1 the buttons they will need to be large to make them work with touch. Which will cause them to overlap a bunch of video.

Seems like a better way would be to allow the user to drag from the upper right corner to resize. Then you can put a small resize indicator there, but the touch area can be much larger (ScreenTools.minTouchPixels) such that it will work on mouse/touch on all devices.

DonLakeFlyer:
Also I think if the resize goes beyond some percentage of the window size it should just flip views when the user stops dragging.

jaxxzer:
> Seems like a better way would be to allow the user to drag from the upper right corner to resize.

Ok Don, agreed. I can work on this. 

> Also I think if the resize goes beyond some percentage of the window size it should just flip views when the user stops dragging.

I think this might not be a good idea, because if the user wants to make the pip as big as possible, they will expand it, it will flip the active views on them, then they will have to tap it again to flip back. I think it should just grow to a point then stop, and tapping/clicking once on the PiP should be the only thing that flips the views.

DonLakeFlyer:
> I think it should just grow to a point then stop, and tapping/clicking once on the PiP should be the only thing that flips the views.

Sounds fine

jaxxzer:
@DonLakeFlyer check it out:
screenshot 2017-09-19 17 54 21

jaxxzer:
The buttons (except recording icon) are hidden unless mouse is over video window.

NaterGator:
This is really nice @jaxxzer!

DonLakeFlyer:
Looks fantastic.

> The buttons (except recording icon) are hidden unless mouse is over video window.

Can you make it not work that way on Mobile. There is no mouse-over concept with touch screens.

jaxxzer:
> Can you make it not work that way on Mobile. There is no mouse-over concept with touch screens.

Mobile bites me again. I wonder how this will look on mobile. I'm not set up to test on mobile, maybe I should do that...

PS The new .svg is not perfect, but it's as good as I could manage.

DonLakeFlyer:
I can run this on mobile to see how it works. That said if you are going to be getting into global type ui changes like this it would be a good idea to be able to test on mobile as well.

DonLakeFlyer:
I think with the new mechanism you don't need small screen test. I would assume that drag handle simply doesn't show at all if they pip starts out at it's maximum possible size. Or in other words screen is so small, you can resize because you are already at the limit.

DonLakeFlyer:
Tested on both desktop and android tablet. Works fantastic. Merging.

DonLakeFlyer:
> I think with the new mechanism you don't need small screen test. I would assume that drag handle simply doesn't show at all if they pip starts out at it's maximum possible size. Or in other words screen is so small, you can resize because you are already at the limit.

@jaxxzer Where did this end up. Not sure if it works this way or not?

jaxxzer:
I'm not sure what you mean.

The new button will always show up on mobile as it is now.

DonLakeFlyer:
I was wondering if there was a case where the pip standard size could not be made any larger on small screens. Since it might already be at max size. I don' think that the case though. I think you can always resize some amount even on tiny screens? How is default size and max size calculated?

jaxxzer:
It's all percentage based. Default size is 20% (Set in flightView). And min/max size are 10% 75%.

src/FlightDisplay/FlightDisplayView.qml
```diff
@@ -49,7 +49,7 @@ QGCView {
     property bool   _isPipVisible:          QGroundControl.videoManager.hasVideo ? QGroundControl.loadBoolGlobalSetting(_PIPVisibleKey, true) : false
     property real   _savedZoomLevel:        0
     property real   _margins:               ScreenTools.defaultFontPixelWidth / 2
-    property real   _pipSize:               mainWindow.width * 0.2
+    property real   _pipSize:               flightView.width * 0.2
```
jaxxzer: Here

src/QmlControls/QGCPipable.qml
```diff
@@ -22,17 +22,100 @@ Item {
     property bool isHidden:  false
     property bool isDark:    false
 
+    // As a percentage of the window width
+    property real maxSize: 0.75
```
jaxxzer: and here

DonLakeFlyer:
Kind of what I thought. So it will work on any screen size, no matter how small.

jaxxzer:
> Kind of what I thought. So it will work on any screen size, no matter how small.

Correct.

PR #5655, opened 2017-09-14
Allow user-selectable brand image

jaxxzer:
Some of our users use screen grab programs to produce deliverables and would like to have their business identified in the shot. This PR allows for that.

screenshot 2017-09-14 12 04 52

### DIFF
--- a/src/ui/toolbar/MainToolBarIndicators.qml
+++ b/src/ui/toolbar/MainToolBarIndicators.qml
@@ -95,8 +95,12 @@ Item {
 
         property bool   _outdoorPalette:        qgcPal.globalTheme === QGCPalette.Light
         property bool   _corePluginBranding:    QGroundControl.corePlugin.brandImageIndoor.length != 0
-        property string _brandImageIndoor:      _corePluginBranding ? QGroundControl.corePlugin.brandImageIndoor : (_activeVehicle ? _activeVehicle.brandImageIndoor : "")
-        property string _brandImageOutdoor:     _corePluginBranding ? QGroundControl.corePlugin.brandImageOutdoor : (_activeVehicle ? _activeVehicle.brandImageOutdoor : "")
+        property string _userBrandImageIndoor:  QGroundControl.settingsManager.appSettings.userBrandImageIndoor.value
+        property string _userBrandImageOutdoor: QGroundControl.settingsManager.appSettings.userBrandImageOutdoor.value
+        property bool   _userBrandingIndoor:    _userBrandImageIndoor.length != 0
+        property bool   _userBrandingOutdoor:   _userBrandImageOutdoor.length != 0
+        property string _brandImageIndoor:      _userBrandingIndoor ? _userBrandImageIndoor : (_corePluginBranding ? QGroundControl.corePlugin.brandImageIndoor : (_activeVehicle ? _activeVehicle.brandImageIndoor : "") )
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
"Link Brand Images"?

jaxxzer:
@DonLakeFlyer 'Link Brand Images' = Use the same image for both indoor and outdoor theme.

DonLakeFlyer:
Seems unnecessary. Simple better?

jaxxzer:
Ok, maybe remove the 'Link' option, and use the default image if the user has not selected a custom image for the current theme.

Or just one image field that will be used for both themes?

DonLakeFlyer:
Use the opposite themed brand image if they only specify one.

jaxxzer:
Got it.

DonLakeFlyer:
Could you add doc in the Daily Build new features page of the qgc docs for this?

jaxxzer:
Yes.

DonLakeFlyer:
Also the whole section group box and the individual fields need to respect SettingsFact.visible state. This allows custom builds to hide things.

jaxxzer:
Ooo, thanks for the tip :+1:

dheideman:
Brand image settings without "link" fact:

removed linked option

Still need to fix the "visible" state.

dheideman:
I just realized that all the brand image facts are lumped in with the AppSettings group internally but have a separate box on the settings page.  Would it be better to move the image path selection into the Miscellaneous box in General Settings, move the facts into their own group internally, or leave it as-is?

src/ui/toolbar/MainToolBarIndicators.qml
```diff
@@ -100,8 +100,8 @@ Item {
...
         property bool   _userBrandingIndoor:    _userBrandImageIndoor.length != 0
         property bool   _userBrandingOutdoor:   _userBrandImageOutdoor.length != 0
-        property string _brandImageIndoor:      _userBrandingIndoor ? _userBrandImageIndoor : (_corePluginBranding ? QGroundControl.corePlugin.brandImageIndoor : (_activeVehicle ? _activeVehicle.brandImageIndoor : "") )
-        property string _brandImageOutdoor:     _userBrandingOutdoor ? _userBrandImageOutdoor : (_corePluginBranding ? QGroundControl.corePlugin.brandImageOutdoor : (_activeVehicle ? _activeVehicle.brandImageOutdoor : "") )
+        property string _brandImageIndoor:      _userBrandingIndoor ? _userBrandImageIndoor : (_userBrandingOutdoor ? _userBrandImageOutdoor : (_corePluginBranding ? QGroundControl.corePlugin.brandImageIndoor : (_activeVehicle ? _activeVehicle.brandImageIndoor : "") ) )
+        property string _brandImageOutdoor:     _userBrandingOutdoor ? _userBrandImageOutdoor : (_userBrandingIndoor ? _userBrandImageIndoor : (_corePluginBranding ? QGroundControl.corePlugin.brandImageOutdoor : (_activeVehicle ? _activeVehicle.brandImageOutdoor : "") ) )
```
jaxxzer: @dheideman Can you reformat this for readability?

dheideman:
@DonLakeFlyer I went ahead and made a new fact group for Brand Image since it doesn't feel like it would fit in with Miscellaneous.  Let me know if there's a better way to handle this.

DonLakeFlyer:
I think that's fine. That way easier to hide the whole thing from the FactGroup

Issue #5455, opened 2017-07-17
Voice reads "m" as minutes rather than meters


jliphard:
For some error messages, the message contains distances in units of meters. E.g. "Mission not accepted - move landing approach 3m further" (or similar). The voice reads this as "3 minutes" rather than "3 meters".

dagar:
Interesting, I had no idea it was expanding units. I wonder if that's intentional.

These error messages are coming from the firmware. For the proper fix long term this goes back to our favorite topic - http://discuss.px4.io/t/event-interface/2318/11

PR #5656, opened 2017-09-14
Numeric speech: Negative numbers and "m" postfix

DonLakeFlyer:
Correct speech for numbers like: -10.5m


DonLakeFlyer:
Fix for: #5455

PR #5660, opened 2017-09-15
Fix text overflow on sub safety page

jaxxzer:


DonLakeFlyer:
Interesting, I've never noticed Text.Wrap before. I always used Text.WordWrap. But .Wrap is better really.

PR #5663, opened 2017-09-17
PX4 Firmware add DO_SET_ROI

dagar:
Added for testing https://github.com/PX4/Firmware/pull/7939

@ndepal FYI

### DIFF
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -271,6 +271,7 @@ QList<MAV_CMD> PX4FirmwarePlugin::supportedMissionCommands(void)
          << MAV_CMD_DO_SET_SERVO
          << MAV_CMD_DO_CHANGE_SPEED
          << MAV_CMD_DO_LAND_START
+         << MAV_CMD_DO_SET_ROI
          << MAV_CMD_DO_MOUNT_CONFIGURE
          << MAV_CMD_DO_MOUNT_CONTROL
          << MAV_CMD_SET_CAMERA_MODE


ndepal:
What functionality does this add? What is missing from QGC at the moment (stable release) is the ability to set ROI points on the map. This doesn't enable that, does it?

dagar:
This lets you insert an ROI command in the mission sequence, but you have to manually populate the fields. The next step could be a helper to set the position on the map.

DonLakeFlyer:
You just add a waypoint and then change the command to ROI. You have the ability to move it around.

ndepal:
I can already add a ROI command by adding a waypoint, choosing "show all values", changing the command to `MAV_CMD_NAV_ROI`, and setting `param1` to 3.
The GPS coordinates are already correct according to where I set the waypoint, but after changing the mode the waypoint is displayed on the map just next to the preceding waypoint instead of at its location.

It would be really nice if the ROI item showed up on the map according to its GPS coordinates and would allow changing the coordinates the same way we can with waypoints.

dagar:
Yes that would be pretty slick. I think it still needs to show next to the real waypoint so the user knows where it's executed, but then have another piece pointing at the region.

Do you want to create a new QGC issue describing how it should work?

@DonLakeFlyer does this overlap with any gimbal work that's already been done?

DonLakeFlyer:
> I can already add a ROI command by adding a waypoint, choosing "show all values", changing the command to MAV_CMD_NAV_ROI, and setting param1 to 3.

If you are using a new daily build you don't have to use show all values. Just the normal change command from the category ui. ROI will be under the Camera category.

DonLakeFlyer:
I just tested and it is working correctly. I think the editor for it needs some work though. @dagar Can you put and issue in for the editor. Also can specify what exactly is supported. There a number of params in there. If not all supported the command json needs to be updated for px4.

ndepal:
@dagar You're right, there needs to be some visual connection to the waypoint at which the ROI takes effect.

@DonLakeFlyer When you say it works correctly, do you mean that you can drag around the ROI on the map as I described above?

I wanted to try out how it works with the daily build (Linux AppImage) but it's giving mean error about the Qt version. Did the requirements change since the last stable release?
The exact error is: `/tmp/.mount_0VGAPo/QGroundControl: /usr/lib/x86_64-linux-gnu/libQt5Xml.so.5: version 'Qt_5' not found (required by /tmp/.mount_0VGAPo/QGroundControl)`.
`/usr/lib/x86_64-linux-gnu/libQt5Xml.so.5` exists and symlinks to `libQt5Xml.so.5.5.1`.

DonLakeFlyer:
> When you say it works correctly, do you mean that you can drag around the ROI on the map as I described above?

I mean you don't need to use "Show all values" to change the command.

dagar:
For ROI PX4 currently supports next waypoint and coordinate. The others could be implemented if anyone actually wants them.

@ndepal might be better able to explain how the gui should work.

PR #5665, opened 2017-09-18
Make FTP more robust

bkueng:
This adds retransmission & requesting missing parts for burst downloads.

Tested with up to 20% packet drops in both directions.

PR for the Firmware: https://github.com/PX4/Firmware/pull/7979

### DIFF
--- a/src/comm/MockLinkFileServer.cc
+++ b/src/comm/MockLinkFileServer.cc
@@ -36,9 +36,22 @@ MockLinkFileServer::MockLinkFileServer(uint8_t systemIdServer, uint8_t component
     _errMode(errModeNone),
     _systemIdServer(systemIdServer),
     _componentIdServer(componentIdServer),
-    _mockLink(mockLink)
+    _mockLink(mockLink),
+    _lastReplyValid(false),
+    _lastReplySequence(0),
+    _randomDropsEnabled(false)
 {
+    srand(0); // make sure unit tests are deterministic
+}
 
+void MockLinkFileServer::ensureNullTemination(FileManager::Request* request)
+{
+    if (request->hdr.size < sizeof(request->data)) {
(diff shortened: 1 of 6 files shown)


bkueng:
@DonLakeFlyer how do I execute single unit tests?

DonLakeFlyer:
--unitest:foo

Also would be great if there was a MockLink based unit test for this (block retry). You can see other cases in MockLink of how for example Mission Protocol can be tested with it and set to various failure states.

bkueng:
Alright I'll look into that.

DonLakeFlyer:
@bkueng Does this change require the firmware change as well. Or will it work against old firmware versions.

bkueng:
There is already a MockLink based unit test. I extended it with random drops to test the retry behavior. 
And I also found this for the file download tests:
```
// Trying to write test code for read and burst mode download as well as implement support in MockLineFileServer reached a point
// of diminishing returns where the test code and mock server were generating more bugs in themselves than finding problems.
```
https://github.com/mavlink/qgroundcontrol/blob/master/src/qgcunittest/FileManagerTest.cc#L198
So I'm not sure if it's worth the efforts of enabling & fixing them. What do you think?

The Firmware PR (https://github.com/PX4/Firmware/pull/7979) is required to guard against any kind of drops/retry.
However if you run this against an older FW version, it will still work. The only failure case is this: if an ack from the FW gets dropped and QGC times out & resends the request, FW will execute the request twice. For instance if you try to open a file or rename it, it will look like it failed, even though the FW executed the first request successfully. This is only a problem for requests that change state, it's ok for pure data retrieval requests, such as reading a file chunk.

DonLakeFlyer:
Yeah, I sort of remember fighting with that. I guess we are good for now then.

DonLakeFlyer:
Thanks @bkueng

PR #5667, opened 2017-09-18
FlightDisplayViewWidgets: increase the maximum widget size from 200 to 400

bkueng:
On large/high DPI screens, the whole UI scales nicely when increasing the
font size, except for the attitude widget, that was limited to 200 pixels.

Before:
qgc_small_indicator
Now:
qgc_small_indicater_larger
It's better, but IMO it could still be a bit larger.

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewWidgets.qml
+++ b/src/FlightDisplay/FlightDisplayViewWidgets.qml
@@ -42,8 +42,7 @@ Item {
         if(ScreenTools.isMobile) {
             return ScreenTools.isTinyScreen ? mainWindow.width * 0.2 : mainWindow.width * 0.15
         }
-        var w = mainWindow.width * 0.15
-        return Math.min(w, 200)
+        return ScreenTools.defaultFontPixelWidth * 30
     }
 
     function _setInstrumentWidget() {


DonLakeFlyer:
The problem here is that is it using pixels. The width should be a ratio of default font width. Then it will scale with the display.

bkueng:
I changed it to scale with the font size.

This is with the default 10pt font size and a window size of ~1100x750px:
qgc_instrument_size_10pt

This is with12pt font size and a maximized window size at Full HD:
qgc_instrument_size_12pt

This is with15pt font size and a maximized window size at Full HD:
qgc_instrument_size_15pt

Mobile is unchanged.

DonLakeFlyer:
@dogmaphobic Can you look at this? Wondering about mobile. Should it something like the default is ScreenTools.defaultFontPixelWidth * 30, but on mobile it checks to make sure that width isn't larger than the percentages you have?

dogmaphobic:
> Wondering about mobile.

This change only affects desktop builds. Mobile still uses the original sizing code.

DonLakeFlyer:
Right. I'm wondering if it should be changed as I describe.

dogmaphobic:
That would be an issue only if you get a 20" tablet :)

DonLakeFlyer:
Ok, fine like this then.

PR #5671, opened 2017-09-19
Some smaller UX improvements

bkueng:
UX improvements that have bothered me (and others too) for quite some time now:
- activate the keyboard focus to the parameter value when start to edit a parameter
- accept numpad enter key to change a param
- allow to search for multiple independent words in parameter names & descriptions.

There's one more thing I wanted to do: set the focus to the mavlink console text input when opening the console view. But I could not get that to work. @DonLakeFlyer would be great if you could do that.

### DIFF
--- a/src/QmlControls/ParameterEditor.qml
+++ b/src/QmlControls/ParameterEditor.qml
@@ -32,7 +32,7 @@ QGCView {
     property Fact   _editorDialogFact: Fact { }
     property int    _rowHeight:         ScreenTools.defaultFontPixelHeight * 2
     property int    _rowWidth:          10      // Dynamic adjusted at runtime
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
The focus/enter changes are going to require significant testing on mobile platforms. I've tried to fix this multiple times in the past and every time I've had to back it out because it causes strange problems. Specifically the focus gets stuck somewhere and then the rest of the ui becomes unresponsive.

Can you split out the parameter search changes into a seperate pull since that should go through more easily.

bkueng:
I tested on my Android (7.0) phone, and setting the focus works. Will probably need to be tested on different versions and iOS.
I excluded setting the focus on mobile, so that this should be good to merge.

DonLakeFlyer:
Let me do some testing on this.

PR #5678, opened 2017-09-20
Firmware-specific fact metadata

jaxxzer:
Point being to change labels like 'airspeed' and 'altitude' to speed and depth, or whatever's appropriate.


DonLakeFlyer:
I understand the need for this. But I would much rather have some sort of generic mechanism where the plugin can replace meta data. Which at this point is not thought through. I don't want a bunch of ```if vehicletype``` stuff in Vehicle.

jaxxzer:
Ok, back to the drawing board. Thanks @DonLakeFlyer.

DonLakeFlyer:
Can you enter an Issue for this and I'll put some thought into it.

jaxxzer:
https://github.com/mavlink/qgroundcontrol/issues/5685

PR #5679, opened 2017-09-20
LogDownloadController: only skip entry if size==0 && APM firmware

bkueng:
On PX4, under certain circumstances and in rare cases, there can be log
files with 0 size. In that case QGC ignored the entry and tried to
re-request the entry indefinitely.

@MaEtUgR you can approve here :)


DonLakeFlyer:
This would cause Refresh to never finish right? If so, I've hit this as well. Awesome.

bkueng:
> This would cause Refresh to never finish right?

Exactly.

Did you merge this now or accidentally close it?

DonLakeFlyer:
Ahh! Wrong button.

PR #5680, opened 2017-09-20
MissionItemStatus: Make waypointValuesDisplay with variable width

patrickelectric:

### DIFF
--- a/src/PlanView/MissionItemStatus.qml
+++ b/src/PlanView/MissionItemStatus.qml
@@ -19,6 +19,7 @@ import QGroundControl.FactSystem    1.0
 import QGroundControl.FactControls  1.0
 
 Rectangle {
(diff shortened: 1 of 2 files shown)


patrickelectric:
canto
canto2

DonLakeFlyer:
Did you check that when it reaches maximum width the scrolling still works?

DonLakeFlyer:
Can you make it show Home if there is only a home position? I think that may make it less confusing to folks as to what the thing is used for.

patrickelectric:
Here a gif of this PR. About the home item, I'll take a look on it.
itemstatus

patrickelectric:
image

DonLakeFlyer:
Looks perfect @patrickelectric. Thank you so much.

PR #5683, opened 2017-09-21
Update MissionController.cc

yl5006:
save to kml use double number with 10 valid digits。


DonLakeFlyer:
Through the rest of the QGC code lat/lon is persisted to 7 digits and altitude to 2 digits after decimal place. Could you change to that so it is consistent? Thanks.

DonLakeFlyer:
Perfect, thank you so much.

PR #5684, opened 2017-09-21
QGCMapPolygon: Fix bug in center polygon item

patrickelectric:
When creating new points, center need to be updated.

### DIFF
--- a/src/MissionManager/QGCMapPolygon.cc
+++ b/src/MissionManager/QGCMapPolygon.cc
@@ -44,7 +44,7 @@ void QGCMapPolygon::_init(void)
 {
     connect(&_polygonModel, &QmlObjectListModel::dirtyChanged, this, &QGCMapPolygon::_polygonModelDirtyChanged);
     connect(&_polygonModel, &QmlObjectListModel::countChanged, this, &QGCMapPolygon::_polygonModelCountChanged);
-    connect(&_polygonModel, &QmlObjectListModel::countChanged, this, &QGCMapPolygon::_updateCenter);
+    connect(this, &QGCMapPolygon::pathChanged, this, &QGCMapPolygon::_updateCenter);
 }
 
 const QGCMapPolygon& QGCMapPolygon::operator=(const QGCMapPolygon& other)
@@ -284,8 +284,6 @@ void QGCMapPolygon::removeVertex(int vertexIndex)
 
     _polygonPath.removeAt(vertexIndex);
     emit pathChanged();
-
-    _updateCenter();
 }
 
 void QGCMapPolygon::_polygonModelCountChanged(int count)


patrickelectric:
Survey bug:
bug

Fix:
corrigido

DonLakeFlyer:
Do you know why this code:
```
void QGCMapPolygon::_init(void)
{
    connect(&_polygonModel, &QmlObjectListModel::dirtyChanged, this, &QGCMapPolygon::_polygonModelDirtyChanged);
    connect(&_polygonModel, &QmlObjectListModel::countChanged, this, &QGCMapPolygon::_polygonModelCountChanged);
    connect(&_polygonModel, &QmlObjectListModel::countChanged, this, &QGCMapPolygon::_updateCenter);
}
```

Isn't working to update the center? The count should change which should trigger a call to _updateCenter.

patrickelectric:
I'll check it.

patrickelectric:
Done.

> Since `_updateCenter` uses `_polygonPath` to calculate new center point,
> it's necessary to connect `pathChanged` with `_updateCenter`

DonLakeFlyer:
Just to make sure I understand. QmlObjectListModel::countChanged was signalled when the QmlObjectListModel changed, but before the path version of the polygon change. Hence _updateCenter was always working on old data?

DonLakeFlyer:
The QGCMapPolygonTest unit test is failing Can you take a look?

DonLakeFlyer:
Also would be awesome to include a new test which covers this case that was broken.

patrickelectric:
> Just to make sure I understand. QmlObjectListModel::countChanged was signalled when the QmlObjectListModel changed, but before the path version of the polygon change. Hence _updateCenter was always working on old data?
Exactly !

patrickelectric:
I'll check the tests asap.

DonLakeFlyer:
It's failing because the centerChanged signal is being emitted twice when it thinks it should only be emitted one.

patrickelectric:
@DonLakeFlyer This new approach will work in all cases and only call _updateCenter once.

patrickelectric:
Is this the correct approach to solve the problem ?

DonLakeFlyer:
Let me take a look. If it turns out to be a real pain to fix the double signalling we can live with it. The centerChanged signal isn't really cannot to anything which would propogate to other parts of the system causing possible perf problems due to running twice.

patrickelectric:
@DonLakeFlyer Probably will, because is [the same thing that I have done..](https://github.com/mavlink/qgroundcontrol/pull/5684/files)

DonLakeFlyer:
Not quite if you look at it. One change was missing.

DonLakeFlyer:
Replaced by #5715

PR #5687, opened 2017-09-22
PlanToolBar: Change upload button visibility

patrickelectric:
The upload button is necessary only when the mission have more than the home point


DonLakeFlyer:
I'd rather not do this. Things which come and go tend to confuse the user. Since they don't understand why. Also Sync/Upload is still available. Is there a problem with the Upload being there when it doesn't do much?

patrickelectric:
@DonLakeFlyer This is what happen when you try to upload.
image

DonLakeFlyer:
That's a bug in the mission protocol. Can yo uenter an issue for that?

patrickelectric:
Ok, no problem.

PR #5693, opened 2017-09-22
Display warning if the user exceeds video storage limit

dheideman:
Some of our customers have complained about loss of video footage due to the video-storage-limiting auto-video-delete feature.  For longer missions the default 2 GB limit on video storage is insufficient and the lack of a notification when the limit is reached exacerbates the problem.  As recorded footage is often very valuable to our customers, we feel the prospect of auto-deletion of footage by default is incredibly risky.

This PR removes the auto-deletion of footage from the video storage limiter and instead displays and announces a warning when either:
- the prescribed limit is reached during recording, or
- a recording is started after the limit has been reached.

The displayed message also displays the current video storage limit and free space on the disk.
screen shot 2017-09-22 at 11 35 33 am

### DIFF
--- a/src/VideoStreaming/VideoReceiver.cc
+++ b/src/VideoStreaming/VideoReceiver.cc
@@ -24,6 +24,7 @@
 #include <QDir>
 #include <QDateTime>
 #include <QSysInfo>
+#include <QStorageInfo>
 
 QGC_LOGGING_CATEGORY(VideoReceiverLog, "VideoReceiverLog")
 
@@ -78,6 +79,7 @@ VideoReceiver::VideoReceiver(QObject* parent)
     connect(this, &VideoReceiver::msgStateChangedReceived, this, &VideoReceiver::_handleStateChanged);
     connect(&_frameTimer, &QTimer::timeout, this, &VideoReceiver::_updateTimer);
     _frameTimer.start(1000);
+    connect(&_videoStorageTimer, &QTimer::timeout, this, &VideoReceiver::_checkVideoStorage);
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Not sure when this video storage limit came into play. This doesn't seem to be handled in a very user friendly way. This pull just seems like a bandaid to the real problem which is that anything which has a limit should be designed in a more user-friendly way:
* You should be able to turn it on/off. Desktop: default off. Mobile: default on.
* This actual limit size should be configurable with useful default set for mobile.
* When you reach the limit, the oldest file should be removed to make space for new data so you don't lose things.

It would be great if someone created a generic mechanism to support such a thing. Then we can use it for both video and telemetry logs which fill up space.

dheideman:
@DonLakeFlyer To address your points, in order:
- I agree.  Desktop/laptop computers are in less of a need to conserve disk space.  However the defaults are stored in a json fact file, so separate files would probably need to be loaded to implement the default off/on states.
- The limit is configurable.  Default: 2048 MB
- Given the first point, this would be acceptable.  The current master already does this, but in all cases and it cannot be disabled.

If you know of any alternatives for setting defaults based on whether the device is mobile, please let me know.  I'm not too keen on the aforementioned solution.

DonLakeFlyer:
Just requires addition of mobileDefault key support in json reader. Should be a simple change.

DonLakeFlyer:
And then I would use that for both on/off default and size limit default.

dheideman:
Sounds good.  I'll see what I can do.

dheideman:
Let me rework this and I'll submit a new PR.

jaxxzer:
> When you reach the limit, the oldest file should be removed to make space for new data so you don't lose things.

This is the current behavior, but I don't think it's a good idea for QGC to auto delete anything. It should be up to the user to make more space.

This PR changes this behavior to remind the user that their allocation has been exceeded, but we keep recording anyway. The user will get the warning every time they start recording another video until they make more space or change their settings.

Issue #5694, opened 2017-09-22
Plan: Structure Scan discussion


DonLakeFlyer:
This issue is for discussion of Structure Scan support in QGC. The simple concept from a functional standpoint is:
* Draw a polygon
* Specify the altitude for the first layer
* Specify the number of layers up you want to go from the first layer
* Specify the amount of height between each layer
* Specify a camera trigger distance to use while traversing the path
* Generate a mission which follows the path of the polygon exterior for each layer triggering the images along the way

Leaving the user model out of it for now and sticking with functional requirements only. What else is needed? What am I missing?

DonLakeFlyer:
Quick example:
<img width="1106" alt="screen shot 2017-09-22 at 2 21 19 pm" src="https://user-images.githubusercontent.com/5876851/30764776-5acc8ae2-9fa1-11e7-9a2a-35f7105fce07.png">

lvale:
Hi @DonLakeFlyer 

Perhaps a desired picture resolution, and a picture target location

DonLakeFlyer:
> Perhaps a desired picture resolution, and a picture target location

I was thinking about resolution. If I think structure scan is just a vertical survey then I can consider the polygon to show to surface of the area to survey. Then if you specified a resolution the vehicle would back off the appropriate amount from the surface to get that resolution. But I'm not sure if that is how structure scan is used.

The picture target location would be perpendicular to the polygon edge facing in.

lvale:
what if you do a structure scan from the inside of a target?. Think survey the inside of the new apple building :)

birchera:
For concave corners the achievable angle of incident vs proximity to structure itself could be a factor. Not sure how to handle it though.

Also, I would think about having the option for a start and an end point. Otherwise it is limited to objects where you can fly around on all sides. Maybe a line would be better suited than a polygon.

DonLakeFlyer:
> what if you do a structure scan from the inside of a target?

> Also, I would think about having the option for a start and an end point.

Yup and yup, great stuff. Thanks guys.

DonLakeFlyer:
Functional support still missing:
* Circular scans

For discussion:
* Does it make sense to be able to set up structure scan like a vertical survey. With things like image resolution dictating distance from structure and so forth.
* Should there be support for video as well?
* Should there be support for running the gimbal angled down towards the structure?

lvale:
[...]

PR #5708, opened 2017-09-27
Initial Structure Scan support

DonLakeFlyer:
Related to #5694.

* Mission generation/upload does work
* File load/save not working
* Various stats, like distance, camera shots and so forth are still not calculated
* Requires a gimbal which supports yaw. This is due to parameters which force vehicle yaw while flying a mission. Working on a solution for this.

screen shot 2017-09-26 at 7 23 54 pm


DonLakeFlyer:
Meant to be functional but not user friendly. Much, much work let to do.

PR #5715, opened 2017-09-29
Fix polygon center updates

DonLakeFlyer:
Replacement for #5684.

@patrickelectric Can you run this and make sure it handles the cases you tried? I believe it works now with no double signalling. It's just a variant of you pull.

Also there seems to be a possible Qt bug where the center indicator sometimes disappears. Just noticed that. I wonder if that is new with 5.9. Never saw that before. Happens on master.


patrickelectric:
Good, I was in a trip sorry for the delay.
Tested and working with same test as #5684.

DonLakeFlyer:
Thans Patrick

PR #5720, opened 2017-09-30
Custom Settings

dogmaphobic:
Allow custom plugins to define their own list of settings panels


DonLakeFlyer:
Why this as well as QmlComponentInfo?

PR #5729, opened 2017-10-05
Add option to disable video auto-delete

dheideman:
Rework of #5693.

This PR adds the option to enable/disable the video auto-delete feature.  Said feature is enabled by default for mobile builds.  For non-mobile builds, it defaults to disabled.

Max storage limits are also adjusted.  Mobile builds will default to 2 GB (same as current versions), whereas non-mobile builds, when auto-delete is enabled, will default to 10 GB.


DonLakeFlyer:
@dogmaphobic Do you have time to look at this one? I won't have time till I get back home.

PR #5731, opened 2017-10-06
add support for flashing APM to AUAVX2.1

kd0aij:


src/VehicleSetup/FirmwareImage.cc
```diff
@@ -238,7 +238,7 @@ bool FirmwareImage::_px4Load(const QString& imageFilename)
 
     uint32_t firmwareBoardId = (uint32_t)px4Json.value(_jsonBoardIdKey).toInt();
     uint32_t actualBoardId = _boardId == Bootloader::boardIDPX4FMUV3 ? Bootloader::boardIDPX4FMUV2 : _boardId;;
```
DonLakeFlyer: This line here is used to convert an actual board id to a compatible one. Such that the != check works correctly. Then isCompatible kind of does the same thing. Could you toss this line and move the V3->V2 compatibility into your new routine which is better.

QGCSetup.pri
```diff
@@ -122,7 +122,8 @@ LinuxBuild {
         libQt5Svg.so.5 \
         libQt5Test.so.5 \
         libQt5Widgets.so.5 \
-        libQt5XcbQpa.so.5
+        libQt5XcbQpa.so.5 \
+        libQt5TextToSpeech.so.5
```
DonLakeFlyer: Awesome, thanks Mark. I haven't had time to fix this from the road.

src/VehicleSetup/FirmwareImage.cc
kd0aij: oh, I didn't notice what that line was doing.

DonLakeFlyer:
Thanks Mark

PR #5732, opened 2017-10-07
use CMD_MISSION_START to start mission and select waypoint

dagar:
This should be fine for startMission (in the PX4 plugin), but I'm not sure about setCurrentMissionSequence.

### DIFF
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -482,15 +482,7 @@ void PX4FirmwarePlugin::guidedModeChangeAltitude(Vehicle* vehicle, double altitu
 
 void PX4FirmwarePlugin::startMission(Vehicle* vehicle)
 {
-
-    if (_setFlightModeAndValidate(vehicle, missionFlightMode())) {
-        if (!_armVehicleAndValidate(vehicle)) {
-            qgcApp()->showMessage(tr("Unable to start mission: Vehicle rejected arming."));
-            return;
-        }
-    } else {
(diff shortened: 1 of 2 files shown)


Antiheavy:
We are seeing some issues where if a waypoint was manually selected on the flight immediately prior to the current one, Start Mission will jump to that waypoint instead of the Takeoff waypoint.  I wonder if this change would be one way to fix that.

Antiheavy:
> We are seeing some issues where if a waypoint was manually selected on the flight immediately prior to the current one, Start Mission will jump to that waypoint instead of the Takeoff waypoint. I wonder if this change would be one way to fix that.

nevermind, it turns out using this command when trying to select another waypoint other than the takeoff waypoint when on the ground causes the system to try to takeoff to that waypoint, which is problematic! 

However, this might still be valid to be used by the "Start Mission" slider as a method to guarantee that the correct takeoff waypoint is selected when doing a Mission takeoff.

dagar:
We can add an additional check on the firmware side for FW that prevents arming into a mission that's not on a takeoff waypoint when landed.

DonLakeFlyer:
It's gonna take me a while to have time to think about this. Not back home till next Wed.

DonLakeFlyer:
Is there anything with switching to this command which makes it PX4 version specific. Or do you old versions of PX4 support this command as well?

dagar:
Not really, it's been for well over a year.

It's definitely better for mission start, but maybe for active waypoint select we should add something to prevent it from being used when landed.

dagar:
I've split out the "Start Mission" portion here. https://github.com/mavlink/qgroundcontrol/pull/5741

PR #5735, opened 2017-10-10
Update TCPLink.cc

tzekian12:
Fix TCP disconnection problem.


DonLakeFlyer:
I don't have anything to test with TCP so I assume this is good.

PR #5741, opened 2017-10-12
PX4 use CMD_MISSION_START to start mission

dagar:
On PX4 MAV_CMD_MISSION_START is a single command that 
 - sets active waypoint
 - changes to mission mode
 - arms

### DIFF
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -482,15 +482,7 @@ void PX4FirmwarePlugin::guidedModeChangeAltitude(Vehicle* vehicle, double altitu
 
 void PX4FirmwarePlugin::startMission(Vehicle* vehicle)
 {
-
-    if (_setFlightModeAndValidate(vehicle, missionFlightMode())) {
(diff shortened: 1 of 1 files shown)


dagar:
Any update?

DonLakeFlyer:
There is a problem in that FirmwarePlugin::startMission is used by both Start Mission and Continue Mission (used while flying) guided actions. In order to make this work a new continueMission needs to be plumbed through FirmwarePlugin -> Vehicle -> GuidedActionsController.qml. The continueMission implementation would just switch to mission flight mode. If you could plumb that through then I believe this will work.

Antiheavy:
> There is a problem in that FirmwarePlugin::startMission is used by both Start Mission and Continue Mission (used while flying) guided actions. In order to make this work a new continueMission needs to be plumbed through FirmwarePlugin -> Vehicle -> GuidedActionsController.qml. The continueMission implementation would just switch to mission flight mode. If you could plumb that through then I believe this will work.

@DonLakeFlyer we are considering this PR as part of a solution to https://github.com/PX4/Firmware/issues/10657 .  Is your comment above still valid?  Is always forcing the mission to start at the beginning an acceptable feature of the "Start Mission" slider (right now it does not do this)?

PR #5758, opened 2017-10-20
update PX4 airframe and parameter metadata

dagar:


dagar:
I'll look at automating this soon.

PR #5778, opened 2017-10-23
fix LogDownloadController: id is uint16

bkueng:
Fixes log download if there are more than 256 logs


bkueng:
AppVeyor CI failure looks unrelated:
```
Uploading artifact "QGroundControl-installer.exe" (96,815,465 bytes) to S3 bucket "qgroundcontrol" as "builds/fix_log_id_type/QGroundControl-installer.exe"
A WebException with status ReceiveFailure was thrown.
```
@DonLakeFlyer could you have a look?

DonLakeFlyer:
I think that's just because AppVeyor gets confused if the branch is not from a fork.

PR #5827, opened 2017-11-07
[WIP] add VTOL transition to commands

dagar:


DonLakeFlyer:
VTOL transition is on the toolbar now

PR #5828, opened 2017-11-07
Maps: added Eniro as map provider (Norway,Sweden,Finland,Denmark,Poland)

AndKe:


src/Settings/FlightMap.SettingsGroup.json
```diff
@@ -3,8 +3,8 @@
     "name":             "MapProvider",
     "shortDescription": "Currently selected map provider for flight maps",
     "type":             "uint32",
-    "enumStrings":      "Bing,Google,Statkart,Mapbox,Esri",
-    "enumValues":       "0,1,2,3,4",
+    "enumStrings":      "Bing,Google,Statkart,Eniro,Mapbox,Esri",
+    "enumValues":       "0,1,2,3,4,5",
```
DonLakeFlyer: You've changed the values for MapBox and Esri won't that break existing settings (same for the header change)? @dogmaphobic ?

AndKe: I tried to maintain the same order, nor did I see Mapbox & Ersi as an option in GUI, the json file is a mystery to me.

DonLakeFlyer: The json is the meta data for the map provider settings Fact which in turn drives the contents of the ui combo box which shows up in the General page of settings. Json meta data for Fact  ui  is something that is used throughout all the ui. The enum values persist the map provider the user has selected. You can't change existing values since users may have them persisted in their settings. If the numbers change then suddenly the maps change out from under users.

DonLakeFlyer: Mapbox and Esri only show up if you have specified tokens for them

AndKe: ok, so I should move Eniro to the end of the list - correct ?
"enumStrings":      "Bing,Google,Statkart,Mapbox,Esri,Eniro",

AndKe: nope, by changing order in .json , I now see "mapbox" as the fourth option, and no Eniro...

AndKe: changed .json and FlightMapSettings.h enum...

dogmaphobic:
Restarted Mac build as Travis, as usual, botched it.

PR #5831, opened 2017-11-10
Crashlytics (fabric.io) implementation test

Buwaggish:
Contains a (on purpose) crash when you select "Delete All Settings" from the general settings.


dogmaphobic:
This is a test branch and not meant for merging.

Issue #4506, opened 2017-02-02
PX4: Hobbs meter


LorenzMeier:
@dogmaphobic These two parameters implement a 64bit microseconds total flight time count (hobbs meter):

```
Flight time in microseconds = (LND_FLIGHT_T_HI << 32) | LND_FLIGHT_T_LO
```

dogmaphobic:
I assume this is milliseconds instead? The values I'm getting only make sense if it's milliseconds (5+ hours instead of 19 seconds).

LorenzMeier:
@bkueng has the details

dogmaphobic:
Nope, it's microseconds alright. The autopilot I was testing with indeed only had 19 seconds of "armed" time.

dogmaphobic:
I armed it for 10 seconds and the timer went up 10 seconds.

DonLakeFlyer:
QGC cannot currently support volatile parameters. This will causes the parameter cache hit to fail 100% of the time. Supporting volatile parameters is doable but non-trivial. Unless you want the cache to no longer work I'd suggest, either:
* Don't support this through a parameter so cache doesn't break
* Live with broken cache and create/implement new volatile parameter architecture. This won't happen before I head out of town in a week and then I"m gone for two weeks. So QGC users will have to live with full param load for a while.
* Back out firmware change until QGC supports volatile parameters so cache continues to work

DonLakeFlyer:
Also FYI: I do think adding support for volatile parameters to QGC is a good thing. It's just a matter of timing.

dogmaphobic:
Note that the parameters themselves have been there for a very long time. I've just added code to read and process them. The cache would have been broken for a very long time as well. This would explain why I felt the cache system would fail quite frequently. Any time you "arm" the vehicle, the hobbs meter increase and the parameter data changes.

NaterGator:
I noticed this a while back. I'm not sure if we have a convenient way to flag volatile params, maybe it makes sense to use a leading underscore as a volatile qualifier? We can hash everything else and explicitly request all volatile params? We could just assign them a value of 0 or NaN during hash generation so we ensure a change in any volatile parameter name is detected.

DonLakeFlyer:
> I'm not sure if we have a convenient way to flag volatile params, maybe it makes sense to use a leading underscore as a volatile qualifier?

It should be a new field for the parameter metadata on firmware side. Then add volatile to FactMetaData and then update ParameterManager to handle the flags.

[...]

PR #5834, opened 2017-11-10
Adding access to the vehicle hobbs meter

dogmaphobic:
Finally, in response to https://github.com/mavlink/qgroundcontrol/issues/4506

Added access to a string containing the vehicle's hobbs meter.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -2873,6 +2873,25 @@ void Vehicle::sendPlan(QString planFile)
     PlanMasterController::sendPlanToVehicle(this, planFile);
 }
 
+QString Vehicle::hobbsMeter()
+{
+    static const char* HOOBS_HI = "LND_FLIGHT_T_HI";
+    static const char* HOOBS_LO = "LND_FLIGHT_T_LO";
+    //-- TODO: Does this exist on non PX4?
+    if (_parameterManager->parameterExists(FactSystem::defaultComponentId, HOOBS_HI) &&
+        _parameterManager->parameterExists(FactSystem::defaultComponentId, HOOBS_LO)) {
+        Fact* factHi = _parameterManager->getParameter(FactSystem::defaultComponentId, HOOBS_HI);
+        Fact* factLo = _parameterManager->getParameter(FactSystem::defaultComponentId, HOOBS_LO);
+        uint64_t hobbsTimeSeconds = ((uint64_t)factHi->rawValue().toUInt() << 32 | (uint64_t)factLo->rawValue().toUInt()) / 1000000;
+        int hours   = hobbsTimeSeconds / 3600;
+        int minutes = (hobbsTimeSeconds % 3600) / 60;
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Could you instead plumb this through as one of the Vehicle FactGroup things. THen it will be available from the instrument panel.

dogmaphobic:
>Could you instead plumb this through as one of the Vehicle FactGroup things. THen it will be available from the instrument panel.

I will need to figure out how to do that. What can I use as a starting point (something to copy)?

DonLakeFlyer:
If you look in Vehicle.h you'll find:
```
    // FactGroup object model properties

    Q_PROPERTY(Fact* roll               READ roll               CONSTANT)
    Q_PROPERTY(Fact* pitch              READ pitch              CONSTANT)
    Q_PROPERTY(Fact* heading            READ heading            CONSTANT)
    Q_PROPERTY(Fact* groundSpeed        READ groundSpeed        CONSTANT)
    Q_PROPERTY(Fact* airSpeed           READ airSpeed           CONSTANT)
    Q_PROPERTY(Fact* climbRate          READ climbRate          CONSTANT)
    Q_PROPERTY(Fact* altitudeRelative   READ altitudeRelative   CONSTANT)
    Q_PROPERTY(Fact* altitudeAMSL       READ altitudeAMSL       CONSTANT)
    Q_PROPERTY(Fact* flightDistance     READ flightDistance     CONSTANT)
    Q_PROPERTY(Fact* distanceToHome     READ distanceToHome     CONSTANT)
```

Those are the top level facts exposed to the instrument panel through the vehicle FactGroup. Just pull the string on any of those and you should find the places. It's basically a variant of Fact and Fact meta data setup.

dogmaphobic:
I've added but I don't think that system handles a "string" value. It shows up as a single "0" instead of the actual (valid) string. I will take a look at it later...

dogmaphobic:
For what's worth, the value comes into separate 32-bit parameters. They must be computed and transformed from a 64-bit time in microseconds to a string representing hhhh:mm:ss. That's what the current `hobbsMeter` property returns. A QString value.

While at it, as they are parameters, they are only updated when you connect. It doesn't increment until you request the parameters again. I also don't understand how the parameter cache works when this value changes each time you connect.

DonLakeFlyer:
If you set the type to string in the meta data and Fact creation it should work. It could be a bug in Fact::enumOrValueString which that page uses.

> I also don't understand how the parameter cache works when this value changes each time you connect.

I basically means the cache is now broken and will never hit.

DonLakeFlyer:
Supporting volatile parameters and keeping the cache working is going to be a bunch of work.

dogmaphobic:
I've plumbed it through Vehicle FactGroup (in addition to the existing *property*). The issue I was having was one of initialization. I need to signal a value change when all parameters are loaded.

dogmaphobic:
And this is what it looks like for now:

screen shot 2017-11-12 at 3 30 48 pm

DonLakeFlyer:
Nice

PR #5842, opened 2017-11-15
[WIP] Elevation data for offline use

birchera:
* Will extend the ElevationProvider to provide elevation data offline (cached).
* Will allow downloading elevation data as "tiles" for specified regions similar to offline maps.

This is work in progress. Currently this is a first draft of the back end. The actual interaction points are still evolving.

FYI @LorenzMeier @bkueng


DonLakeFlyer:
Something like this should be integrated into Offline Maps somehow. When I'm download map tiles there is an additional checkbox for downloading elevation data as well. A completely seperate system (or may just UI) doesn't seem like a great idea.

birchera:
I looked at the offline map things before and thought it rather not suited to also handle elevation data in a way that we can still access it for planning purposes. But that was no profound investigation.

> When I'm download map tiles there is an additional checkbox for downloading elevation data as well.

Where can you check that? If you can give me a pointer I can check it out.

DonLakeFlyer:
I see two things here:
1) The existing offline maps cache system supports storage and retrieval of "tiles" based on lat/lon position. Maybe it needs some more work to make it more generic, but I wonder why we need two systems to store/retrieve information based on lat/lon positions.
2) Even if 1 is in appropriate I think it makes sense to tie togther the user interface of offline map set download/creation to terrain data. When you create an offline map set you could also check "Terrain "data" which would then also cache the terrain data for the same region as specified by the offline map. Then the same ui provides the way to manage the map/terrain data as a single unit.

DonLakeFlyer:
@dogmaphobic Would also be good to get Gus to chime in here when he gets back from Zurich.

birchera:
Thanks for the insights.

Regarding 1. I planned to store the datasets in some kind of database when selected as offline available. I'll look further into the offline maps cache system to see whether I can tweak it to that end. If that works well we could also replace the hash table with it.

Regarding 2. I did not work on the UI exposure yet. Having it as an option with the offline maps definitely makes sense. Another thing I need to look into is the complete life cylce of the data.

DonLakeFlyer:
Offline Maps is a SQL database which Gus has added geolocation query support to. I think the lifecycle of the data should just match the Offline Maps lifecycle. Only extra thing needed is a max space for the system cache just like offline maps has.

birchera:
I went through the offline map implementation. I can see how this would work out if we cache the json files we download in this data base (triggered by downloading an offline map).

Apart from that, it looks to me like deeper integration could quickly become messy. They are not actual images but text files (json) to be saved. So all the map plugin interface functionality would need to distinguish. Also it seems to me like the handy functionality of figuring out when to load and delete local or online tiles is all embedded within the qt map.

So here is what I propose:
* Online usage remains separate like it is now.
* There is an option when creating a new offline tile set whether to include elevation.
* These elevation tiles are downloaded within the ElevationProvider, but then stored via the QGCTileCacheWorker in the database.
* Each time before downloading an elevation tile in the ElevationProvider, the database is queried to check if we already have that specific tile.

The offline map things are quite a complex piece of code and my insight limited. So if you disagree, please do point me in the right direction.

DonLakeFlyer:
Going to need to wait until Gus gets back so he can comment since that is all his code.

birchera:
Implemented the offline caching with the existing database. @dogmaphobic please check if that is along the lines of what you proposed.

Regarding behavior, there is now the option to download offline elevation data individually or as a checkbox-option together with the maps. They appear as separate datasets in the list.

LorenzMeier:
@dogmaphobic @bkueng It would be good to continue to drive this towards merging. We need to find a modus operandi that does not depend on Don being around to merge. Let me know if you want me to review.

dogmaphobic:
CI has been flaky but it looks like the build is truly broken:
```

**Comment by dogmaphobic** [comment]
CI has been fluffy but it looks like the build is truly broken:

```
/home/travis/build/mavlink/qgroundcontrol/src/Terrain.cc: In member function ‘void ElevationProvider::_fetchedTile()’:
/home/travis/build/mavlink/qgroundcontrol/src/Terrain.cc:214:31: error: no matching function for call to ‘QHash<QString, TerrainTile>::insert(QString&)’
             _tiles.insert(hash), *terrainTile);
                               ^
/home/travis/build/mavlink/qgroundcontrol/src/Terrain.cc:214:31: note: candidate is:
In file included from ../Qt5.9-linux/5.9.1/gcc_64/include/QtCore/qdebug.h:45:0,
                 from ../Qt5.9-linux/5.9.1/gcc_64/include/QtCore/qloggingcategory.h:44,
                 from ../Qt5.9-linux/5.9.1/gcc_64/include/QtCore/QLoggingCategory:1,
                 from /home/travis/build/mavlink/qgroundcontrol/src/QGCLoggingCategory.h:17,
                 from /home/travis/build/mavlink/qgroundcontrol/src/TerrainTile.h:4,
                 from /home/travis/build/mavlink/qgroundcontrol/src/Terrain.h:12,
                 from /home/travis/build/mavlink/qgroundcontrol/src/Terrain.cc:10:
../Qt5.9-linux/5.9.1/gcc_64/include/QtCore/qhash.h:757:52: note: QHash<K, V>::iterator QHash<K, V>::insert(const Key&, const T&) [with Key = QString; T = TerrainTile]
 Q_INLINE_TEMPLATE typename QHash<Key, T>::iterator QHash<Key, T>::insert(const Key &akey,
                                                    ^
../Qt5.9-linux/5.9.1/gcc_64/include/QtCore/qhash.h:757:52: note:   candidate expects 2 arguments, 1 provided
/home/travis/build/mavlink/qgroundcontrol/src/Terrain.cc:214:46: error: expected ‘;’ before ‘)’ token
             _tiles.insert(hash), *terrainTile);
                                              ^
```

birchera:
I received the API specifications [here](https://developers.airmap.com/v2.0/docs/elevation-api) and corrected the query accordingly. This should now be ready.

I noticed that the latency for the API requests is sometimes pretty large, so I was wondering whether we could increase the [timeout for tile download](https://github.com/wingtra/qgroundcontrol/blob/07bb23653d81c2a5836c889abb7fe98e46415483/src/QtLocationPlugin/QGeoMapReplyQGC.cpp#L175-L178). Or do you see any reason not to @dogmaphobic? In case we can also make it a parameter and set it specifically for the elevation tile download.
Alternatively we could try reducing the size of the tiles.

dogmaphobic:
>I noticed that the latency for the API requests is sometimes pretty large

How large? The current 10 seconds is already way longer than necessary to accommodate LTE/4G. If we are to change the timeout, we ought to do it just for this as it's the exception. Regular tile download will issue several requests simultaneously and it relies on a somewhat quick timeout to give up trying.

What is the payload and how large is it? If it's ASCII Json and it's larger than a couple kilobytes, it should probably be gziped to avoid the wasted bandwidth.

birchera:
Yes, it is ASCII Json and the file size is ~33kB. If we decrease the tile size we can of course bring it down. But I think it is the server that is not super responsive.

DonLakeFlyer:
From quick look through it seem like this only store data. I don't see any change in code to be able to query this data like where the mission item status display shows you terrain heights?

birchera:
@DonLakeFlyer [this](https://github.com/mavlink/qgroundcontrol/pull/5842/files#diff-1b9a6c96da2fa1731fb352c2dd5fc0a3R94) line calls the instance of QGeoTiledMapReplyQGC, which first queries the database and if the data is not there, fetches it online. Not much coding in this respect, since we recycle much of the existing structure.

DonLakeFlyer:
Ah, thanks @birchera. I figured it was in there somewhere. I forgot how the ElevationProvider thing worked. I was looking in the wrong place.

DonLakeFlyer:
Some questions on this before I get into a big code review:
- What format are the tiles stored into the database with?
- Can you add a screenshot of what it looks like to turn on offline terrain download.
- How does terrain download affect the offline download progress indications. Tile count? Data size?
- How does terrain download relate to Max Cache values as specified in Offline Maps options? These are important on mobile which has limited resources.

birchera:
> What format are the tiles stored into the database with?

The tiles are stored as the json files that we receive from the download. The map engine has no functionality to convert or interpret this data.

> Can you add a screenshot of what it looks like to turn on offline terrain download.

| Unselect elevation data | Select elevation data | Downloading |
| --- | --- | --- |
| screenshot from 2017-12-15 11 17 22 | screenshot from 2017-12-15 11 17 31 | screenshot from 2017-12-15 11 17 59 |

> How does terrain download affect the offline download progress indications. Tile count? Data size?

As you can see in the third screenshot above, elevation tile sets are shown as separate sets. This allows also import / export, as well as start / stop / resume of download and delete individually from the respective map tile set.

> How does terrain download relate to Max Cache values as specified in Offline Maps options? These are important on mobile which has limited resources.

Elevation tiles add towards the grand total, since they live within the same database. However, their data size is much smaller than that of map tiles. A 0.01°x0.01° elevation tile is only slightly above 5kB.

DonLakeFlyer:
Thanks @birchera. Awesome stuff.

@dogmaphobic Do you think it would be possible to add a hook to be able to compile these tiles down to binary for storage?

dogmaphobic:
> @dogmaphobic Do you think it would be possible to add a hook to be able to compile these tiles down to binary for storage?

If it's a data structure that doesn't change, only the number of elements do, yes. I would need to actually know what it's been stored :P

DonLakeFlyer:
I was thinking the tiles could be store as just double arrays.

DonLakeFlyer:
I'd like to get the tiles stored in binary instead of json if possible before this goes through. That way it isn't a database format change.

dogmaphobic:
What's the status with this? I just built it and it all seem to be working. The only missing thing is saving the "tile" in binary form instead of Json, which shouldn't be too complicated to implement. It looks like you have a fixed header (bounds and stats) and an array of arrays (what are those "carpets"?)

In summary, this can be merged as is (after rebasing) and I can deal with the storage later (bad) or I can take over and add the binary implementation (better). The reason merging this as is is bad is because the "tiles" saved by this implementation won't be compatible with the binary version.

DonLakeFlyer:
Unless the tiles are versioned somehow won't letting this through without binary store mean a database format change when binary is supported requiring everyone to blow away old database. That doesn't sound like a good idea.

dogmaphobic:
There is a concept of ID for the _type_ of "tiles". If you change the ID, you don't need to blow away the database (though the old "tiles" will stay there unused).

DonLakeFlyer:
I'd rather wait and get it right. I think this also makes sense from waiting until after 3.3 goes out for this. It's a big piece of code and big code changes to dump in so close to a Stable going out.

birchera:
I see the point that we want to get this straight from the beginning, so we don't need to wipe the database later.

We are also internally converging towards a release. So I cannot promise that I will be able to do it within the next weeks. The timeline I can give is week 9 / 10.

Alternatively, @dogmaphobic if you want to contribute the binary implementation that is also fine with me. The carpet is the array of ground elevation altitudes on the rectangle specified by the request, inflated to the nearest outer pixels of the underlying elevation tiles.

dogmaphobic:
> Alternatively, @dogmaphobic if you want to contribute the binary implementation that is also fine with me.

Ok. If you don't mind, I will hijack your PR, implement the binary conversion and submit as a new PR. Either way, this won't get into 3.3 in time, though I should have it done shortly.

philipoe:
Hi @dogmaphobic @birchera, whats the status with this? I could definitely help with testing and also a bit of coding where that makes sense. In the end, what we want to do is to actually verify the feasibility of a whole path (not just the waypoints) w.r.t. terrain collisions in QGC. The current online-elevation-data-fetch system is too slow for that I assume. If you guys focus on finishing this PR here, then I can start working on the full-path-terrain-checking.

birchera:
Open things to do:
* Update to diverged master.
* Implement the conversion to and from a compact binary format.

I will be able to tackle this earliest from the 12th march.

DonLakeFlyer:
> then I can start working on the full-path-terrain-checking.

Which means what?

philipoe:
> Which means what?

A simple way to check for possible terrain collisions along the whole path. If i understand https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/VisualMissionItem.cc#L179 correctly, then currently the terrain is only checked at the actual coordinate of a mission item/waypoint. I would then for example assume straight line paths and check along that whole line, or for loitering waypoints (for fixed-wings) check the whole area where the aircraft loiters.

DonLakeFlyer:
@birchera I think it makes sense to wait until I get at least the first full pass at Corridor Scan terrain follow work I'm doing through before going back into this. I keep going into the terrain query code and changing things around as I work on that.

birchera:
That is unfortunate. I updated and sorted things out over the past 2 days. Was just waiting for binary compression to work for pushing.
I also wonder whether your efforts here make sense, given that the only thing we need to download with this patch is carpets. Other than you needing it right now, of course.
To give you a timeline, I have time to push this again on monday, such that we could update this PR latest tuesday.

LorenzMeier:
Hey Don - it would be good if we could find a way to add finer granularity to the merging process. This PR has been pending for a long time and I'm worried that we're discouraging contributions. Anything that I or @dogmaphobic can help with to facilitate that?

DonLakeFlyer:
If you read the thread you will see that the reason it has been pending for so long is because I wanted terrain tiles to be stored in binary format. It doesn't make sense from a perf standpoint to have to reparse big json datasets over and over and over again. If it went through as is it would have required a database reset when it finally switched to binary with another pull. That seemed like a really bad idea, hence best to get it right the first time.

Fast forward to 3 months later: The pull has not been updated with the requested changes. I have now gotten deep into terrain work which means I've had to change the apis related to terrain queries. Which in turn causes this pull to take a hit.

Happy for anyone to explain to me some way to make this process better. I can't wait around to make forward progress when things sit around for that long.

DonLakeFlyer:
> I also wonder whether your efforts here make sense, given that the only thing we need to download with this patch is carpets.

That is correct, but it is still an async protocol (even when eventually connected to the offline database) on the client side which is where the complications come in. Terrain differs from map tiles in that the client code needs to wait for the terrain data to come back before it can proceed with calculations. Whereas map tiles are strictly visual and no client code waits on them to perform further calculations. The query to a server part is easy (whether that be direct to AirMap or the offline server), getting the async part correct is not.

DonLakeFlyer:
@birchera Once this pull: https://github.com/mavlink/qgroundcontrol/pull/6231 goes through the terrain stuff should be stable enough for me to abstract out the query part such that it can either run offline or online. Then you can plug offline support directly into that.

birchera:
> Fast forward to 3 months later: The pull has not been updated with the requested changes.

We only had a ~6 weeks slot to work on this last year and initial reviewing took quite some time as well. Unfortunately I don't fully control my priorities. For the future I think we either need to cope with these inconveniences or sync more tightly in the initial efforts.

In any case the code structure looks now very mature and it should be easy to integrate the offline caching part. Thanks for pushing that so quickly @DonLakeFlyer.

DonLakeFlyer:
> We only had a ~6 weeks slot to work on this last year and initial reviewing took quite some time as well. Unfortunately I don't fully control my priorities.

I have no problem with that whatsoever. Everyone is busy. Just trying to explain to Lorenz what is going on.

birchera:
@DonLakeFlyer I updated this again to master and pushed our solution for the compressed representation of the elevation tiles. Can you let us know what you think about it?

In order to test this, please change the terrain query interface [here](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.h#L127) and [here](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.h#L175). Carpet provider is not implemented, as this is nowhere used right now, and I am also not sure how I would do it, given the tessellation with the tiles.

DonLakeFlyer:
Fantastic. I'll take a look at this today. Let me think about carpet a bit to see what I really need there. I don't think I need the actual points.

DonLakeFlyer:
Current thinking is what I need instead of true carpet is just the min/max height for the carpet as opposed to all the values. Is that doable?

DonLakeFlyer:
Ok. Gave look through the top part of this code. Connected up and both individual coord queries for mission items as well as Path queries for Corridor Scan all worked fine.

@dogmaphobic Have you already looked through the map tile connection part of this before?

If Gus gives the ok, then you can switch over internally to it. Then I'll merge. And we can continue carpet discussion.

birchera:
> Current thinking is what I need instead of true carpet is just the min/max height for the carpet as opposed to all the values. Is that doable?

We also thought about using min/max of an area for a custom feature. However, we found we could more efficiently replace it with an according path call in the end. If you want to implement it for an area, I would propose the following (for minimum as an example):
* Specify a polygon rather than a rectangle aligned with north
* Go through all the terrainTiles that contain part of the polygon
* Iterate through all the points that lie within the polygon and check them for minimum value
* Heuristic to speed up the process: Check whether terrainTile min from the stats are already higher than the minimum value found so far. If so, skip the whole tile.

> then you can switch over internally to it.

I assume that means switching to TerrainOfflineAirMapQuery interface. I'll do that shortly.

DonLakeFlyer:
Yes, I will need a polygon for min/max. The example usage would be for a standard survey (non-terrain) to be able to show whether there is any issue with terrain collision with survey area.

DonLakeFlyer:
@dogmaphobic Review?

DonLakeFlyer:
@dogmaphobic Review?

dogmaphobic:
> @dogmaphobic Review?

I'm not getting these. For all others, I get a specific email with a notice of mention. I bumped into this only because I was specifically looking to see what was going on and noticed your request. I'll take a look.

>Have you already looked through the map tile connection part of this before?

Yes, early on and it all seemed right at the time. Let me build it and test to see it all again.

DonLakeFlyer:
Are notifications turned off on the repo for you?

DonLakeFlyer:
Gus says ready to go. But unit tests are crashing somewhere related to terrain. Looking...

DonLakeFlyer:
Still fighting with a crashing unit test which doesn't repro for me on OSX.

PR #5845, opened 2017-11-17
Fix compilation on Ubuntu 17.04 as described in issue 5843

pritamghanghas:


dogmaphobic:
This is a dupe of #5854, which I have already lined up for a merge.

PR #5848, opened 2017-11-19
MissionController: Add takeoff only with supported vehicles

patrickelectric:

### DIFF
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -340,7 +340,10 @@ int MissionController::insertSimpleMissionItem(QGeoCoordinate coordinate, int i)
     newItem->setCommand(MavlinkQmlSingleton::MAV_CMD_NAV_WAYPOINT);
     _initVisualItem(newItem);
     if (_visualItems->count() == 1) {
-        newItem->setCommand(_controllerVehicle->vtol() ? MavlinkQmlSingleton::MAV_CMD_NAV_VTOL_TAKEOFF : MavlinkQmlSingleton::MAV_CMD_NAV_TAKEOFF);
+        MavlinkQmlSingleton::Qml_MAV_CMD takeoffCmd = _controllerVehicle->vtol() ? MavlinkQmlSingleton::MAV_CMD_NAV_VTOL_TAKEOFF : MavlinkQmlSingleton::MAV_CMD_NAV_TAKEOFF;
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
This code is incorrect.  The Vehicle::takeoffVehicleSupported is for guided commands not for mission commands. What problem are you trying to fix here?

patrickelectric:
Thank you for reply @DonLakeFlyer, this patch remove takeoff in mission commands in vehicles that don't have this feature, like ArduSub.
Can you say the best approach to solve this problem ?

DonLakeFlyer:
ArduSub should implement FirmwarePlugin::supportedMissionCommands. Then something like that:

```
    if (_visualItems->count() == 1) {
        MAV_COMMAND takeoffCmd = _controllerVehicle->vtol() ? MAV_CMD_NAV_VTOL_TAKEOFF : MAV_CMD_NAV_TAKEOFF;
        if (_controllerVehicle->firwarePlugin()->supportedMissionCommands().contains(takeoffCmd)) {
            newItem->setCommand((MavlinkQmlSingleton::Qml_MAV_CMD)takeoffCmd;
    }
```

patrickelectric:
@DonLakeFlyer is this implementation correct ?

src/MissionManager/MissionController.cc
```diff
@@ -1033,8 +1036,9 @@ void MissionController::_recalcWaypointLines(void)
         // If we still haven't found the first coordinate item and we hit a takeoff command, link back to home
         if (firstCoordinateItem &&
                 item->isSimpleItem() &&
-                (qobject_cast<SimpleMissionItem*>(item)->command() == MavlinkQmlSingleton::MAV_CMD_NAV_TAKEOFF ||
-                 qobject_cast<SimpleMissionItem*>(item)->command() == MavlinkQmlSingleton::MAV_CMD_NAV_VTOL_TAKEOFF)) {
+                (!_controllerVehicle->firmwarePlugin()->supportedMissionCommands().contains(MAV_CMD_NAV_TAKEOFF) ||
```
DonLakeFlyer: This part should not change. This is testing whether a takeoff command is already in the mission.

patrickelectric: without this, when planning the mission, the home do not connect with the first item.

DonLakeFlyer: Ok, I get it now. I wasn't understanding the code correctly. This is right.

DonLakeFlyer:
This is all good to go now

PR #5849, opened 2017-11-19
MissionItemStatus: Below altitude color change according to vehicle

patrickelectric:


DonLakeFlyer:
It seem like the real change here would be to not show terrain values for a sub?

DonLakeFlyer:
@patrickelectric @jaxxzer ?

patrickelectric:
The terrain alttitude will be removed when using AUV or ASV.
This PR will be updated.

patrickelectric:
Moving to #5891

PR #5851, opened 2017-11-19
MissionItemEditor: Add label text functionality to work with complex items

patrickelectric:

### DIFF
--- a/src/PlanView/MissionItemEditor.qml
+++ b/src/PlanView/MissionItemEditor.qml
@@ -67,8 +67,13 @@ Rectangle {
         anchors.verticalCenter: commandPicker.verticalCenter
         anchors.leftMargin:     _margin
         anchors.left:           parent.left
-        text:                   missionItem.homePosition ? "H" : missionItem.sequenceNumber
         color:                  _outerTextColor
+        text: {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I don't think this is a good idea. There are times you need to know the sequence number like when using a GoTo command. Not sure how change from number to letter adds value since it already says "Survey" right next to it. Is there a good reason why this would be useful?

patrickelectric:
Just to be consistent with "L", "T" and "H".

DonLakeFlyer:
Only H shows up in item editors. This is because "Planned Home Position" is a wierd special case I think that is ok. But everything else is numeric. There is no L or H:
screen shot 2017-12-11 at 8 55 59 am

patrickelectric:
I was talking about the terrain altitude panel 
screenshot from 2017-12-12 09-25-15

DonLakeFlyer:
Yeah, I guess you could say it's all a bit consistent in it's inconsistency :)

DonLakeFlyer:
The idea for the Terrain view is that all you have is the letter or number there. So it uses the letter. Whereas the item editors also have the full names right next to the number. So it just sticks with the number.

PR #5867, opened 2017-12-04
'manual control lost' when connected joystick with armed UAV

gigorvd:
Hello @dagar 
The test is done on Windows and Android. Version 3.2.5
When a Mavlink UAV is connected, with the activated USB joystick there is a very strong boost of the program. This causes a constant loss of communication with the UAV and the message: manual control lost.

Each time the user moves the map, the central coordinate and zoom level are stored in the configuration file, and this is read every time to change the position of the map on the screen. When you move the map then this reading / writing is performed many times per second and the system is working incorrectly.

If you redefine these functions in such a way that the map coordinate and zoom level are stored simply in the variable, then the speed grows several times and the map becomes comfortable. And in the configuration file, you can save the coordinate when you exit the program.

### DIFF
--- a/src/QmlControls/QGroundControlQmlGlobal.cc
+++ b/src/QmlControls/QGroundControlQmlGlobal.cc
@@ -50,8 +50,6 @@ QGroundControlQmlGlobal::QGroundControlQmlGlobal(QGCApplication* app, QGCToolbox
     settings.beginGroup(_flightMapPositionSettingsGroup);
     _coord.setLatitude(settings.value(_flightMapPositionLatitudeSettingsKey, 0).toDouble());
     _coord.setLongitude(settings.value(_flightMapPositionLongitudeSettingsKey, 0).toDouble());
-
-    settings.beginGroup(_flightMapPositionSettingsGroup);
     _zoom = settings.value(_flightMapZoomSettingsKey, 2).toDouble();
     //if config file is clear
     if(_zoom == 0) _zoom = 17.0;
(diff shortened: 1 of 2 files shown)


dagar:
Thanks @gigorvd 
I've pushed your branch to a branch off the main repo so that the CI system will create an android package for testing. https://github.com/mavlink/qgroundcontrol/tree/pr-android-slowmap_gigorvd

src/QmlControls/QGroundControlQmlGlobal.cc
```diff
@@ -41,11 +44,30 @@ QGroundControlQmlGlobal::QGroundControlQmlGlobal(QGCApplication* app, QGCToolbox
...
     // We clear the parent on this object since we run into shutdown problems caused by hybrid qml app. Instead we let it leak on shutdown.
     setParent(NULL);
+    // Load last coordinates and zoom from config file
+    QSettings settings;
+    
+    settings.beginGroup(_flightMapPositionSettingsGroup);
+    _coord.setLatitude(settings.value(_flightMapPositionLatitudeSettingsKey, 0).toDouble());
+    _coord.setLongitude(settings.value(_flightMapPositionLongitudeSettingsKey, 0).toDouble());
+
+    settings.beginGroup(_flightMapPositionSettingsGroup);
```
dogmaphobic: I get it this came to be because of a copy and paste but you might as well remove the redundant `settings.beginGroup()` for the zoom level.

src/QmlControls/QGroundControlQmlGlobal.cc
```diff
@@ -41,11 +44,30 @@ QGroundControlQmlGlobal::QGroundControlQmlGlobal(QGCApplication* app, QGCToolbox
...
 {
-
+    // Save last coordinates and zoom to config file
+    QSettings settings;
+    
+    settings.beginGroup(_flightMapPositionSettingsGroup);
+    settings.setValue(_flightMapPositionLatitudeSettingsKey, _coord.latitude());
+    settings.setValue(_flightMapPositionLongitudeSettingsKey, _coord.longitude());
+     
+    settings.beginGroup(_flightMapPositionSettingsGroup);
```
dogmaphobic: Same as above

dogmaphobic:
By the way, the CI error is the usual Travis messed up Mac build system.

PR #5886, opened 2017-12-14
Add tuning page for Sub vehicles

jaxxzer:
screenshot 2017-12-14 15 10 46


DonLakeFlyer:
This FactSlider thing is really nice. But I don't think those +/- minus buttons are going to be clickable on mobile. Two options:
1) Make them regular size buttons so they are clickable.
2) Don't use QGCButton and create your own clickable area for them using a control and QGCMouseArea. QGCMouseArea automatically expands the click area on mobile.

jaxxzer:
Nice point; will do.

DonLakeFlyer:
Did you check to see what that looks like on mobile? minTouchPixels can be kinda big. It's mostly used to increases the touch area, but not the actual visuals. Which is what QGCMouseArea does for you automatically.

patrickelectric:
[Here](https://media.giphy.com/media/3o752emN7Wo7KZ0P9S/giphy.gif) you can check the PR running in 1920x1080, android device.

DonLakeFlyer:
Hmm, pretty big. But good enough for now

PR #5894, opened 2017-12-18
Suppress mission ack error until target_system is properly handled

sanderux:
@DonLakeFlyer the groundstation is receiving ACK messages from the onboard computer and does not inspect target_system. Can we suppress this message (only one found) until the system IDs are properly checked?

### DIFF
--- a/src/MissionManager/PlanManager.cc
+++ b/src/MissionManager/PlanManager.cc
@@ -588,7 +588,8 @@ void PlanManager::_handleMissionAck(const mavlink_message_t& message)
     switch (savedExpectedAck) {
     case AckNone:
         // State machine is idle. Vehicle is confused.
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I'd assume it's just as easy to fix this correctly or is there some gotcha there?

sanderux:
We could fix higher level, ignore every packet that doesnt have target_system = gqc or 0
but i dont know what issues we might face doing that

DonLakeFlyer:
Could you change the code to just output a debug message to the PlanManagerLog instead?

sanderux:
Like this?

DonLakeFlyer:
Perfect, this should be good enough.

sanderux:
let is pass through ci, i havent tested it

PR #5895, opened 2017-12-18
Cleanup record button

jaxxzer:


DonLakeFlyer:
What exactly does this do?

jaxxzer:
It's non-functional. Just refactoring. I'll amend it to rid of the whitespace changes in a moment.

jaxxzer:
This was in a branch we used to fix some problems with the blinking animation during state changes, but those same problems were resolved when @dogmaphobic moved the video record button to a widget. Sorry if the branch name is misleading here.

jaxxzer:
Actually, this does change the visual style slightly (it will make the camera picture fade with the red background). I like the way it is currently better. I'm closing this, sorry for the noise.

jaxxzer:
Thanks @dogmaphobic for the visibility->opacity change, I think it is a good improvement.

PR #5898, opened 2017-12-18
Update new lights step parameter setup for ArduSub

jaxxzer:


dogmaphobic:
Are you closing and re-opening to restart a failed Travis build? If that's the case, you can restart just the failed build instead of the whole thing. The Mac build usually fails because of timeouts.

jaxxzer:
Yes that's what I'm doing. I think I need push access to the repo to restart the build through the travis UI. Am I wrong?

dogmaphobic:
> I think I need push access to the repo to restart the build through the travis UI. Am I wrong?

I'm not sure what you need. @DonLakeFlyer? Every push I ever make, unless it's at 3AM on Sunday, fails the Mac build and I always just restart that one build.

DonLakeFlyer:
@jaxxzer If you get a failure on the mac build don't worry about. That only works on 3am Sunday as Gus says. If you don't have admin on travis then you can't re-run individual builds.

PR #5902, opened 2017-12-20
Bugfix joystick input handling with multi-vehicle scenario

jaxxzer:
_loadSettings should not be called in the Vehicle constructor, it may enable/disable the joystick for the active vehicle when a new vehicle connects.

_loadSettings should be called when the active vehicle changes.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -182,7 +182,8 @@ Vehicle::Vehicle(LinkInterface*             link,
 {
     _addLink(link);
 
-    connect(_joystickManager, &JoystickManager::activeJoystickChanged, this, &Vehicle::_activeJoystickChanged);
+    connect(_joystickManager, &JoystickManager::activeJoystickChanged, this, &Vehicle::_loadSettings);
+    connect(qgcApp()->toolbox()->multiVehicleManager(), &MultiVehicleManager::activeVehicleAvailableChanged, this, &Vehicle::_loadSettings);
 
     _mavlink = _toolbox->mavlinkProtocol();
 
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
> _loadSettings should not be called in the Vehicle constructor, it may enable/disable the joystick for the active vehicle when a new vehicle connects.

I don't get this. A Vehicle object instance is associated directly with a single vehicle. Yet this change seems to connect to other vehicles coming and going which seems wrong.

jaxxzer:
_loadSettings may be a little bit misleading, as it is more like _loadJoystickEnabled, the only per-vehicle setting.

The problem presently is that when new vehicle connects, the joystick is disabled or enabled according to the new vehicle settings, however when a new vehicle connects, it is not always the active vehicle. This means that the active vehicle starts receiving joystick updates according to the new vehicle settings.

This change makes the joystick enabled setting get loaded when the active vehicle changes, to match the new active vehicle settings.

In any case, if joystick is enabled, the joystick updates are always sent to only the active vehicle.

DonLakeFlyer:
It seems like the concept of enabling joystick on a vehicle by vehicle basis seems wrong. Would it be better if joystick enabled was global instead?

jaxxzer:
I had the same thought, but I think it needs to remain a per vehicle setting, otherwise there will be conflicts between MANUAL_CONTROL and actual RC input, unless we can count on the autopilot handling that.

DonLakeFlyer:
I think the real solution is to disallow multi-vehicle is you are using a joystick. How does using a joystick with multiple vehicles make any sense at all? There is only one joystick.

jaxxzer:
Yes, the only cases would be if you wanted to monitor other vehicles while controlling only one (we do this to monitor the boat position while controlling the ROV), or control a swarm :).

This PR will work for the first case.

PR #5904, opened 2017-12-20
Formatting fixes

jaxxzer:
Currently indicator widgets look funny on laptops with 4k screens.

## Before:

qgc-old-formatting

## After:

qgc-formatting-fixes-11-10pt

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewWidgets.qml
+++ b/src/FlightDisplay/FlightDisplayViewWidgets.qml
@@ -41,6 +41,8 @@ Item {
     function getPreferredInstrumentWidth() {
         if(ScreenTools.isMobile) {
             return mainWindow.width * 0.25
+        } else if(ScreenTools.isHugeScreen) {
+            return mainWindow.width * 0.11
         }
         return ScreenTools.defaultFontPixelWidth * 30
     }
--- a/src/QmlControls/ScreenTools.qml
+++ b/src/QmlControls/ScreenTools.qml
@@ -60,6 +60,7 @@ Item {
     property bool isDebug:          ScreenToolsController.isDebug
     property bool isTinyScreen:     (Screen.width / Screen.pixelDensity) < 120 // 120mm
     property bool isShortScreen:    ScreenToolsController.isMobile && ((Screen.height / Screen.width) < 0.6) // Nexus 7 for example
+    property bool isHugeScreen:     Screen.width >= 1920*2
 
     readonly property real minTouchMillimeters: 10      ///< Minimum touch size in millimeters
     property real minTouchPixels:               0       ///< Minimum touch size in pixels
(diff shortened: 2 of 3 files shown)


DonLakeFlyer:
Confused on before an after. Code change seems to make widget smaller on large screen?

src/FlightDisplay/FlightDisplayViewWidgets.qml
```diff
@@ -39,8 +39,11 @@ Item {
     QGCPalette    { id: qgcPal }
 
     function getPreferredInstrumentWidth() {
+        ScreenTools.printStats()
         if(ScreenTools.isMobile) {
             return mainWindow.width * 0.25
+        } else if(ScreenTools.isHugeScreen) {
+            return mainWindow.width * 0.11
         }
         return ScreenTools.defaultFontPixelWidth * 30
```
jaxxzer: Before it was always some # of pixels * 30.
After it is 11% of screen width if using 4k screen (which has tiny pixels and apparently the same defaultFontPixelWidth as lower res screens).

DonLakeFlyer: Ah, I see.

src/FlightDisplay/FlightDisplayViewWidgets.qml
```diff
@@ -39,8 +39,11 @@ Item {
     QGCPalette    { id: qgcPal }
 
     function getPreferredInstrumentWidth() {
+        ScreenTools.printStats()
```
DonLakeFlyer: Can you pull this out?

src/QmlControls/ScreenTools.qml
```diff
@@ -90,6 +91,10 @@ Item {
         }
     }
 
+    function printStats() {
```
DonLakeFlyer: Can you change this to printScreenStats?

dogmaphobic:
Has this been tested on a Mac with Retina display? Windows handles high DPI display poorly and this may be a Windows only issue. In addition, there are settings that must be configured for Qt to handle high DPI displays properly. I've been running QGC on a 4K display for over a year and have not seen this issue.

http://doc.qt.io/qt-5/highdpi.html

jaxxzer:
Thanks for pointing that out @dogmaphobic. I will investigate this further; this was only tested on a windows display.

src/FlightDisplay/FlightDisplayViewWidgets.qml
DonLakeFlyer: I was confused that this isn't a mobile case.

DonLakeFlyer:
I think this instrument width sizing needs an end to end rework. It just seems like hack on top of hack. There should be something which is combination of font sizing and percentage of window width that should be workable for everything.

DonLakeFlyer:
Something where font size ratio is a minimum and window width percentage is a maximum.

DonLakeFlyer:
Merging this in and then I'll test on various devices to see what things look like.

jaxxzer:
Looks ok on early 2015 12 inch macbook with 2304x1440 display.

jaxxzer:
Only the iMac desktops have 4k+ widths that would fall into the case where this pr changes behavior. https://en.wikipedia.org/wiki/Retina_Display

PR #5910, opened 2017-12-23
RTCMMavlink: fix sequence numbers

bkueng:
sequence numbers are only increased after a new RTCM message, while only
the fragment id is incremented within a fragmented message.

Only relevant for APM vehicles, but I do not have an APM vehicle to test this.


DonLakeFlyer:
Thanks @bkueng

PR #5915, opened 2017-12-27
Renaming new flight mode to SmartRTL

lvale:
Correction on the definitive naming of the new Flight Mode


DonLakeFlyer:
Thanks @lvale

PR #5920, opened 2017-12-28
Serial connect fixes

LorenzMeier:
These two commits prevent two bad things on app boot:

  - We stop connecting to ports that are part of the system or are already open under a different alias
  - We continue to run the event loop while connecting

Combined this makes the boot / connect experience of QGC much better and much more responsive. At least on my machine with multiple serial devices connected.

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -502,12 +502,6 @@ void LinkManager::_updateAutoConnectLinks(void)
 
         if (portInfo.getBoardInfo(boardType, boardName)) {
 
-            if (portInfo.isSystemPort()) {
-                // Don't connect to system ports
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
Do you know exactly why the code is trying to open an incorrect port? If it is then there may be something wrong with the matching code somewhere. It will only get info on all ports, which should work fine even for already open ports. It should only try to open ports which match vendor ids or the regexp fallbacks. Need to know that first, before knowing that the fix is.

DonLakeFlyer:
Or is it just the double-open problem?

LorenzMeier:
The intent was to a) proven double opens (and I verified that they happen) and b) ensure that system ports are ruled out. However, we can drop the system ports.

LorenzMeier:
Re-tested again and already stripped system and duplicate ports from the list at the list population stage (as you suggested Don).

PR #5922, opened 2017-12-29
Add camera definitions for Parrot Sequioa RGB and Monochrome sensors

samduff:
Pretty self explanatory I hope! I did a double check with Mission Planner for the same survey, getting same results for time between photos, number of photos, lanes, etc..


DonLakeFlyer:
Thanks

PR #5936, opened 2017-12-31
Fix iOS build

DonLakeFlyer:
This gets it closer to working. I think there are still some bundle identifier problems which require manual futzing. But this is a lot closer.


DonLakeFlyer:
@dogmaphobic Can you sanity check?

DonLakeFlyer:
Merging this in since I need it for testing.

PR #5946, opened 2018-01-02
Add NMEA GPS support for GCS position

jaxxzer:
This will allow the use of USB/Serial GPS for computers or mobile devices without built in positioning. The user must specify the serial port and baudrate for reading data.

screenshot 2018-01-02 18 43 47

### DIFF
--- a/src/PositionManager/PositionManager.cpp
+++ b/src/PositionManager/PositionManager.cpp
@@ -22,6 +22,7 @@ QGCPositionManager::QGCPositionManager(QGCApplication* app, QGCToolbox* toolbox)
 QGCPositionManager::~QGCPositionManager()
 {
     delete(_simulatedSource);
+    delete(_nmeaSource);
 }
 
 void QGCPositionManager::setToolbox(QGCToolbox *toolbox)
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
Ah, nice feature thanks. That settings ui needs to be coded to respect the visible bit on the settings.

jaxxzer:
Hi, can you please point out which visible bit you are referring to? I am missing something.

DonLakeFlyer:
A SettingsFact has a visible property associated with it. Custom builds can override Fact meta data as well as the visibility of the setting through QGCCorePlugin:
```
    /// Allows the core plugin to override the setting meta data before the setting fact is created.
    ///     @param metaData - MetaData for setting fact
    /// @return true: Setting should be visible in ui, false: Setting should not be shown in ui
    virtual bool adjustSettingMetaData(FactMetaData& metaData);
```

 If you look at GeneralSettings.qml you'll see references to foo.visible all over the place to respect this propert.

jaxxzer:
Ok cool, it is already hidden when autoconnect settings is hidden. The qml stuff has the same parent as the other autoconnect settings. https://github.com/mavlink/qgroundcontrol/pull/5946/files#diff-fb7fceee8e3e9e26de059b4ec2919762R463

DonLakeFlyer:
That's only the top level bit. You can hide entire groups. Or you can hide individual items. If you go down a bit from that code you can see how each individual FactCheckBox in the repeater can be hidden as well based on visible bit.

jaxxzer:
Ok got it, fixed. Curious why ios is in the checks for autoconnect settings? I matched the other autoconnect settings.

DonLakeFlyer:
> Curious why ios is in the checks for autoconnect settings?

Hmm, that's slighty whacky. It's because ios only support WiFi connection. But it's not quite done correctly I don't think. Android has similar partial support.

DonLakeFlyer:
https://github.com/mavlink/qgroundcontrol/issues/5973

Issue #5967, opened 2018-01-04
Enable Android QGC for installation on Chrome OS


wallbanger123:
Most of the other Mavlink based GCS apps are available - I'd really like to install QGroundControl.

https://developer.android.com/topic/arc/index.html

PR #5969, opened 2018-01-04
AndroidManifest.xml update to enable ChromeOS support

dagar:
I'm not sure if this is the right way to do this, but apparently the only reason QGC isn't already available to Chromebook's (with the Play store) is the location.GPS requirement.

image

https://github.com/mavlink/qgroundcontrol/issues/5967

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -59,6 +59,7 @@
 
     <!-- Support devices without USB host mode since there are other connection types -->
     <uses-feature android:name="android.hardware.usb.host" android:required="false"/>
+    <uses-feature android:name="android.hardware.location.GPS" android:required="false" />
     <!-- The following comment will be replaced upon deployment with default features based on the dependencies of the application.
          Remove the comment if you do not require these default features. -->
     <!-- %%INSERT_FEATURES -->


dagar:
@wallbanger123 do you have a chromebook you could try sideloaded this on?

wallbanger123:
I do.

dagar:
Android package for this branch. https://s3-us-west-2.amazonaws.com/qgroundcontrol/builds/pr-android_chrome/QGroundControl.apk

wallbanger123:
It runs. It did not recognize a USB connection to my Pixhawk. That may be a ChromeOS/Android limitation, I need to do some tinkering. Bluetooth (which I use with a DragonLink) is able to scan and see the DragonLink TX, but it seems to time out when trying to connect. Tower does the same thing.

I'm not able to test a network based connection, but I would imagine that would work without an issue.

I'll have some more time to work with in over the weekend. Thanks for getting this out so fast!

dagar:
Awesome!

dagar:
Thanks Gus. We'll keep an eye on it after it goes out through Play Beta.

wallbanger123:
Bluetooth is working. The trick is to pair the bluetooth connection in ChromeOS first, and then launch QGC.

PR #5984, opened 2018-01-06
WIP: Try turning iOS back on in Travis

DonLakeFlyer:


dogmaphobic:
Without app signing?

DonLakeFlyer:
Just to prevent build breaks

PR #6000, opened 2018-01-11
Offline map fixes

dogmaphobic:
* Fixed an issue when importing a tile set that already exists would cause an error. When an imported set has a same name, it is renamed so the name is unique. The original logic had an error and was allowing the "new" set to be added with the same name, which would cause an error. In addition, if a new tile set brings no new unique tiles, it is now ignored.

* Fixed an issue where the progress was based on the total number of incoming tiles but this total can change if a given tile already exists in the database (it is not added).


dogmaphobic:
Cool, round number PR :)

Issue #6007, opened 2018-01-12
Missing parts in a survey mission using the current daily build (12.01.2018) on Android


ThorstenDP:
Hi all,

I just came across this:

![screenshot_20180112-230010](https://user-images.githubusercontent.com/5606072/34897191-1f83ffd4-f7ed-11e7-975a-ac29ee8d40e5.jpg)

DonLakeFlyer:
Need the .plan file. Zip it and attach here.

ThorstenDP:
Will do tomorrow. Where is it located? The mission folder is empty. I'll check again tomorrow...

DonLakeFlyer:
You have to save it yourself and it should be in the misison folder.

ThorstenDP:
This should be it: [Missions.zip](https://github.com/mavlink/qgroundcontrol/files/1627670/Missions.zip)

PR #6026, opened 2018-01-15
Use a much larger fudge factor to increase initial transect length

DonLakeFlyer:
This prevents intersection misses on more cases. Fix for #6007


Antiheavy:
oohhh neat.  the issue in https://github.com/mavlink/qgroundcontrol/issues/6007 happens to us fairly regularly too - enough to be annoying.

PR #6029, opened 2018-01-16
Stable v3.2

GlidLov:
i'm using a tablet configuration of QGC. 

Wifi is occupied by a RTSP digital video streamer from the UAV. Wifi network is dead-end.
(BT is occupied by telemetry tunelling)

I would need to keep internet connection to DL map, share some data on the cloud and download air traffic information. It's good to keep the connection alive..

But Android stack (i think IOS also) try to connect by wifi network first, and if dead end, it will loose connection despite having 3g cell network avaiabe. Also some samsung phones have a "intelligent network" option which automatically switch to 3g id the wifi is dead end, but it does not work well.

Is there a way that would let QGC choose to D/L data from cell network while keeping a dead end wifi video stream working?


dogmaphobic:
It has to be. The changes have nothing to do with the description. It looks more like an Issue that got inserted as a PR with some weird changes along the way.

PR #6037, opened 2018-01-16
Handle settings when armed

dogmaphobic:
Instead of blocking all settings if the vehicle is armed, check the provided `VehicleComponent::allowSetupWhileArmed()` method. This change not only disables the settings button, but also shows @jaxxzer's overlay only when appropriate.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSafetyComponentSub.qml
+++ b/src/AutoPilotPlugins/APM/APMSafetyComponentSub.qml
@@ -22,7 +22,6 @@ import QGroundControl.ScreenTools   1.0
 SetupPage {
     id:                 safetyPage
     pageComponent:      safetyPageComponent
-    visibleWhileArmed:   true
 
     Component {
         id: safetyPageComponent
--- a/src/AutoPilotPlugins/Common/SetupPage.qml
+++ b/src/AutoPilotPlugins/Common/SetupPage.qml
@@ -21,8 +21,9 @@ import QGroundControl.Controllers   1.0
 
 /// Base view control for all Setup pages
 QGCView {
-    id:         setupView
-    viewPanel:  setupPanel
+    id:             setupView
+    viewPanel:      setupPanel
+    enabled:        !_shouldDisableWhenArmed
 
     property alias  pageComponent:      pageLoader.sourceComponent
     property string pageName:           vehicleComponent ? vehicleComponent.name : ""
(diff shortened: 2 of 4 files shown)


DonLakeFlyer:
The button shouldn't disable. It should stay enabled so the user can see the message which says unavailable when armed. Otherwise they don't know what is going on. This matches how it worked before.

dogmaphobic:
> The button shouldn't disable.

Are you sure about that? I would assume that's expected that some things change state when a vehicle is present and/or armed. That overlay is a bit messy. I didn't do anything because it was only shown in rare cases where you somehow were in it when the vehicle was armed. If the buttons are to remain enabled, I need to come up with something a little more polished.

Note the "Firmware" button disabled. I left it here for contrast. It should be obvious you can't do certain things when the vehicle is armed.

screen shot 2018-01-16 at 7 43 36 pm

dogmaphobic:
Alright, buttons back to enabled. I can make it "pretty" later.

jaxxzer:
Consider you may also want to check on something in one of those pages while the vehicle is armed. Before, the overlay was opaque, I made it transparent for this reason. This way, you can see the settings, but you can't edit them.

dogmaphobic:
I've tweaked the UI. The semi opaque overlay remains the same but the "Disable while armed" text is now next to the title and away from the middle of the screen.

screen shot 2018-01-17 at 12 00 37 pm

DonLakeFlyer:
Old code didn't use opaque. It just had a rectangle with the working. I'd rather not change the semantics from the way it worked before. I don't want to have to explain over and over again why the buttons are disabled. The message saves me customer support.

dogmaphobic:
The message is there in red. Should I change this? To what? I don't know what this used to look like before @jaxxzer's change.

DonLakeFlyer:
I think what you have is fine. What jaxxzer says about sort of seeing values is intersting. 

Previous was a black rectangle that said some centered words in white. Look in old pull if you want to know exact. Yours is better.

jaxxzer:
For convenience: https://github.com/mavlink/qgroundcontrol/commit/f11f006b1b3c41a44ea8777413e6b4dd464b9a37#diff-9592a40212849f1b9dc1e22e173496a7

DonLakeFlyer:
@dogmaphobic Can you rebase this to master to get units test working?

PR #6049, opened 2018-01-18
Hide popup icon until mouseover

jaxxzer:
just like the other icons


DonLakeFlyer:
What about touch?

src/QmlControls/QGCPipable.qml
```diff
@@ -128,7 +128,7 @@ Item {
         fillMode:       Image.PreserveAspectFit
         anchors.left:   parent.left
         anchors.top:    parent.top
-        visible:        !isHidden && !inPopup && !ScreenTools.isMobile && enablePopup
```
jaxxzer: touch is already  handled here. We never show this button on mobile.

PR #6059, opened 2018-01-23
Robustify mission upload/download

julianoes:
This is the result of 2 days debugging mission upload and download over a lossy link.

The way I debugged this was to filter out some of the mavlink mission messages on the firmware side.

For more information, please check the commit messages.

Tested with current PX4/Firmware master but also with https://github.com/PX4/Firmware/pull/8750.

### DIFF
--- a/src/MissionManager/PlanManager.cc
+++ b/src/MissionManager/PlanManager.cc
@@ -36,8 +36,7 @@ PlanManager::PlanManager(Vehicle* vehicle, MAV_MISSION_TYPE planType)
 {
     _ackTimeoutTimer = new QTimer(this);
     _ackTimeoutTimer->setSingleShot(true);
-    _ackTimeoutTimer->setInterval(_ackTimeoutMilliseconds);
-    
+
     connect(_ackTimeoutTimer, &QTimer::timeout, this, &PlanManager::_ackTimeout);
 }
 
@@ -245,6 +244,24 @@ void PlanManager::_ackTimeout(void)
 
 void PlanManager::_startAckTimeout(AckType_t ack)
 {
+    switch (ack) {
+    case AckMissionItem:
+        // We are actively trying to get the mission item, so we don't want to wait as long.
+        _ackTimeoutTimer->setInterval(_retryTimeoutMilliseconds);
+        break;
+    case AckNone:
+        // FALLTHROUGH
+    case AckMissionCount:
+        // FALLTHROUGH
+    case AckMissionRequest:
+        // FALLTHROUGH
+    case AckMissionClearAll:
+        // FALLTHROUGH
+    case AckGuidedItem:
+        _ackTimeoutTimer->setInterval(_ackTimeoutMilliseconds);
+        break;
+    }
+
     _expectedAck = ack;
     _ackTimeoutTimer->start();
 }
--- a/src/MissionManager/PlanManager.cc
+++ b/src/MissionManager/PlanManager.cc
@@ -30,6 +30,7 @@ PlanManager::PlanManager(Vehicle* vehicle, MAV_MISSION_TYPE planType)
     , _transactionInProgress(TransactionNone)
     , _resumeMission(false)
     , _lastMissionRequest(-1)
+    , _missionItemCountToRead(-1)
     , _currentMissionIndex(-1)
     , _lastCurrentIndex(-1)
 {
@@ -317,6 +318,7 @@ void PlanManager::_handleMissionCount(const mavlink_message_t& message)
         for (int i=0; i<missionCount.count; i++) {
             _itemIndicesToRead << i;
         }
+        _missionItemCountToRead = missionCount.count;
         _requestNextMissionItem();
     }
 }
@@ -460,7 +462,7 @@ void PlanManager::_handleMissionItem(const mavlink_message_t& message, bool miss
         return;
     }
 
-    emit progressPct((double)seq / (double)_missionItems.count());
+    emit progressPct((double)seq / (double)_missionItemCountToRead);
     
     _retryCount = 0;
     if (_itemIndicesToRead.count() == 0) {
--- a/src/MissionManager/PlanManager.h
+++ b/src/MissionManager/PlanManager.h
@@ -144,6 +144,7 @@ protected:
     QList<int>          _itemIndicesToWrite;    ///< List of mission items which still need to be written to vehicle
     QList<int>          _itemIndicesToRead;     ///< List of mission items which still need to be requested from vehicle
     int                 _lastMissionRequest;    ///< Index of item last requested by MISSION_REQUEST
+    int                 _missionItemCountToRead;///< Count of all mission items to read
     
     QList<MissionItem*> _missionItems;          ///< Set of mission items on vehicle
     QList<MissionItem*> _writeMissionItems;     ///< Set of mission items currently being written to vehicle
--- a/src/MissionManager/PlanManager.h
+++ b/src/MissionManager/PlanManager.h
@@ -28,11 +28,11 @@ Q_DECLARE_LOGGING_CATEGORY(PlanManagerLog)
 class PlanManager : public QObject
 {
     Q_OBJECT
-    
+
 public:
     PlanManager(Vehicle* vehicle, MAV_MISSION_TYPE planType);
     ~PlanManager();
-    
+
     bool inProgress(void) const;
     const QList<MissionItem*>& missionItems(void) { return _missionItems; }
 
@@ -41,11 +41,11 @@ public:
 
     /// Last current mission item reported while in Mission flight mode
     int lastCurrentIndex(void) const { return _lastCurrentIndex; }
-    
+
     /// Load the mission items from the vehicle
     ///     Signals newMissionItemsAvailable when done
     void loadFromVehicle(void);
-    
+
     /// Writes the specified set of mission items to the vehicle
     /// IMPORTANT NOTE: PlanManager will take control of the MissionItem objects with the missionItems list. It will free them when done.
     ///     @param missionItems Items to send to vehicle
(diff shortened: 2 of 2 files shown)


LorenzMeier:
@julianoes CI failed.

dogmaphobic:
Rebase to master and these build issues should go away. Appveyor switched to Qt 5.9.3 and there were a few unused variables on build with video streaming disabled. I've fixed both issues. Not sure about the Android failure but that looks like a fluke.

src/MissionManager/PlanManager.cc
```diff
@@ -218,9 +218,16 @@ void PlanManager::_ackTimeout(void)
             }
         } else {
             // Vehicle did not request all items from ground station
-            _sendError(AckTimeoutError, tr("Vehicle did not request all items from ground station: %1").arg(_ackTypeToString(_expectedAck)));
-            _expectedAck = AckNone;
-            _finishTransaction(false);
+            if (_retryCount > _maxRetryCount) {
+                // Vehicle did not respond to MISSION_ITEM, try again
```
bkueng: Is the comment `try again` correct?

DonLakeFlyer:
"This adds retrying to send a mission item in case a mission item gets lost on the way to the drone."

I"m confused. This doesn't make sense to me. The mission upload is a pull style protocol controlled by the vehicle side. If the vehicle side doesn't get an item it should specifically ask for it again. The ground side should not just send another one. The vehicle should keep track of which items it has and which it doesn't and know exactly if it missed something. Why would the vehicle side send the final ack for a mission upload complete if it didn't get all the mission items?

This is exactly how QGC handles the mission download sequence which is a pull style protocol controlled from the ground station side. If it doesn't get a mission item it just asks for it again.

DonLakeFlyer:
The various delete calls that were added were not needed:
* QFile objects close themselves when they go out of scope. As do lots of other Qt thingies.
* A QObject which is allocated with another object as parent will automatically be destructed when the parent QObject is destroyed.

Both of these are common patterns which are used all over the place.

DonLakeFlyer:
"Terrain: clean up reply"

This change doesn't really smell right. QGC doesn't not own these reply objects. I noticed some other code similar to this in QGCFileDownload which also looks strange. @dogmaphobic Any comment on this. You may know more here.

DonLakeFlyer:
"PlanManager: keep a flag for int mode"

This changes looks a bit strange as well. What is described "keep the mode based on capability" isn't really what is implemented. The flag changes based on messages it gets as well as capability bit. Did this change fix some sort of bug that was found? Globals make my skin crawl.

DonLakeFlyer:
Just so it doesn't get lost. Here is the other place I found code futzing with the reply object:
https://github.com/mavlink/qgroundcontrol/blob/master/src/QGCFileDownload.cc#L107

DonLakeFlyer:
From reading through the QNetworkReply::finished and QNetworkReply::close docs it looks to me like it will call close itself and free up resources after the finish signal come backs.

Calling deleteLater on the QNetworkReply which you get from QNetworkAccessManager::get may be correct accordind to this note on QNetworkReply docs:

```Note: Do not delete the object in the slot connected to the error() or finished() signal. Use deleteLater().```

If so, there may be other places in code where these things are leaking. It not quite clear as to who is responsible to free this.

**Comment by DonLakeFlyer** [comment]
From what I can tell from internet searching the QNetworkReply should have deleteLater called on it from the finished singal. Given that I think there is a leak here:
https://github.com/mavlink/qgroundcontrol/blob/master/src/QtLocationPlugin/QGCMapTileSet.cpp#L280

**Comment by DonLakeFlyer** [comment]
Searched for other leaks. Found this possible thing:
https://github.com/mavlink/qgroundcontrol/blob/master/src/Camera/QGCCameraControl.cc#L1471

Also the QGCMapTileSet may not really be leaking because the deleteLater is in the error signal.

File: src/MissionManager/PlanManager.cc (hunk 1 above)
Comment (julianoes) [reply src/MissionManager/PlanManager.cc]: Not really, thanks, I fixed it.

**Comment by julianoes** [comment]
@bkueng @DonLakeFlyer @dogmaphobic thanks for your reviews and comments.

I've rebased and removed all commits which just have random cleanup and delete calls since they didn't seem to be correct.

> "This adds retrying to send a mission item in case a mission item gets lost on the way to the drone."
> 
> I"m confused. This doesn't make sense to me. The mission upload is a pull style protocol controlled by the vehicle side. If the vehicle side doesn't get an item it should specifically ask for it again. The ground side should not just send another one. The vehicle should keep track of which items it has and which it doesn't and know exactly if it missed something. Why would the vehicle side send the final ack for a mission upload complete if it didn't get all the mission items?

I have changed the text of the commit message, it was confusing:

```
PlanManager: re-send mission item if it got lost

This allows a mission item to be sent again if it is requested again
even if it has been sent before.
```
Basically, what happened before was that we could not request a mission item again because it was already removed from the `_itemIndicesToWrite` once sent.

> "PlanManager: keep a flag for int mode"
> 
> This changes looks a bit strange as well. What is described "keep the mode based on capability" isn't really what is implemented. The flag changes based on messages it gets as well as capability bit. Did this change fix some sort of bug that was found? Globals make my skin crawl.

This was required because we know also retransmit a mission item and need to remember if it was int or float mode.

> This is exactly how QGC handles the mission download sequence which is a pull style protocol controlled from the ground station side. If it doesn't get a mission item it times out and asks for it again.

I agree for the mission download. However, for the mission upload I'm not sure who is "pulling", so who should retry.

I'm looking at https://mavlink.io/en/protocol/mission.html and it's not entirely clear to me because the timeout seems to always be on the QGC side. However, the retransmit is (or was) implemented on the drone side for the upload, so the drone needs to re-request the mission item.

julianoes:
I also need to look at the different timeouts that happen and do more debugging. I'll update this tomorrow again.

LorenzMeier:
@julianoes Now that I'm revisiting this the PX4 PR might not have been correct. The "pulling" side should retry. On upload the GCS just tells the drone to pull and then it pulls. Can you please consolidate and send a new PR?

The MAVLink docs are actually incorrect on the diagram. The timeout should be on the drone side in this case. I now see how this caused confusion.
https://mavlink.io/en/protocol/mission.html

LorenzMeier:
Issue on Firmware side to remind us to roll back some of the changes: https://github.com/PX4/Firmware/issues/8791

DonLakeFlyer:
> Basically, what happened before was that we could not request a mission item again because it was already removed from the _itemIndicesToWrite once sent.

Ah, ok. That makes sense. Let me look through this part again.

Given your explanations your other stuff looks good then.

DonLakeFlyer:
> Basically, what happened before was that we could not request a mission item again because it was already removed from the _itemIndicesToWrite once sent.

Number of things:
* In looking at the code I don't see how that would be the case. _handleMissionRequest will always respond with the item. The only thing that could happen would be if the GCS side timed out before the vehicle side retried the request. The _itemIndicesToWrite only comes into play to provide a different error message to the user when the GCS side times out. It doesn't affect the protocol sequence at all.
* The new code still handles the protocol wrong as far as I can see. The vehicle should request again  on a timeout if it misses an item. The GCS side should not resend automatically on a timeout. Which goes back to my original comments about the vehicle side being in control of hardening a pull protocol.
* Worried about descreasing the timeouts. The GCS side timeout needs to be larger than the vehicle side time such that the vehicle side has a chance to retry before the GCS side complains. I think keeping the GCS side high is safer to prevent the GCS side timing out before the vehicle side. In reality it's a bit of a whole in the spec. The spec should likely specify the timeouts for the GCS and Vehicle side such that they will work correctly.

So as far as I can tell the QGC code is correct. There are two possible problems here:
* The firmware is not retrying when it doesn't get a MISSION_ITEM in response to a MISSION_REQUEST.
* The timeouts for retries on the firmware side are too long and the QGC side too short such that the QGC side times out and fails the sequence before the firmware side times out and retries.

DonLakeFlyer:
I just built a version of QGC which drops every other response to MISSION_REQUEST by not sending the MISSION_ITEM keeping the rest of the code the same. This simulates packet loss. I also bumped the QGC side timeout to 10 seconds just to make sure we weren't having collision on timeouts. I then ran this against master firmware SITL. The firmware side never re-requested the dropped item. This shows an incorrect implementation of retry hardening in the firmware code. The vehicle side should have timed out on the lack of the MISSION_ITEM response and re-sent the MISSION_REQUEST.

DonLakeFlyer:
Branch is here to use for testing: https://github.com/DonLakeFlyer/qgroundcontrol/tree/MissionWriteSimulatePacketLoss. Then turn on PlanManagerLog and you can see exactly what is going on.

julianoes:
@DonLakeFlyer yes I agree with what you found and was trying similar drops. I think the pulling side needs to have a lower timeout for the retransmission then the "serving" side which eventually times out. 

Right now it looks like QGC does not wait long enough for the case where the Firmware is pulling.

julianoes:
@LorenzMeier yes I'll go through the Firmware changes again and revert what is needed.

julianoes:
@DonLakeFlyer ok I've slimmed this PR down to only two changes. One to fix the progress bar and the other one to use a different timeout when retrying actively to request items. For more info, see the commit messages.

DonLakeFlyer:
@julianoes I think the timeout change will be ok. But given my history of hand to hand combat with the mission and parameter protocols and the differences between PX4, ArduPilot and random mavlink firmware implementations of these tells me that this should not go into Stable 3.3 which is about to release. In 3.4 it should get enough bake time to be sure it has no strange gotchas.

julianoes:
@DonLakeFlyer that's ok. The current implementation works as long as there are no message drops, so nothing is broken for that case.

DonLakeFlyer:
Awesome, thanks Julian.

PR #6061, opened 2018-01-25
Android hotplugging

nanthony21:
QGroundControl on android can currently only detect a new joystick on startup. If a joystick is added or removed after startup then this will not be detected and the joystick interface will simply freeze up.

This pull request uses QGamepadManager to signal when a gamepad/joystick has been removed or added. The active joystick is then set or cleared appropriately.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -225,7 +225,8 @@ QT += \
     svg \
     widgets \
     xml \
-    texttospeech
+    texttospeech\
+    gamepad
 
 # Multimedia only used if QVC is enabled
 !contains (DEFINES, QGC_DISABLE_UVC) {
--- a/src/Joystick/JoystickAndroid.cc
+++ b/src/Joystick/JoystickAndroid.cc
@@ -69,7 +69,7 @@ JoystickAndroid::~JoystickAndroid() {
 QMap<QString, Joystick*> JoystickAndroid::discover(MultiVehicleManager* _multiVehicleManager) {
     bool joystickFound = false;
     static QMap<QString, Joystick*> ret;
-
+    QMap<QString, Joystick*> newRet;
     _initStatic(); //it's enough to run it once, should be in a static constructor
 
     QMutexLocker lock(&m_mutex);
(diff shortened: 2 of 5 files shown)


dogmaphobic:
Rebase to master and these build issue should go away. Appveyor switched to Qt 5.9.3 and there were a few unused variables on build with video streaming disabled. I've fixed both issues.

nanthony21:
Looks like it's ok now.

DonLakeFlyer:
The problem here is that it adds the use of a new Qt module: gamepad. That tends to require all sort of other things to be fixed up for that:
* README for Qt Install
* Various parts of QGCSetup.pri and QGCInstaller.pri which fart around with plugins in the built images
* Linux AppImage support

nanthony21:
Ok. Do you think that it is worth making these changes? 

Right now that are some issues with QGamePad on windows, but once those are fixed it could potentially be used on all platforms which would prevent Qgroundcontrol from needing to maintain multiple joystick implementations.

Even if the qgamepad portion doesn't get added now, I think it would be good to keep the changes to JoystickAndroid.cc. These changes will be needed for handling joystick connects/disconnects on android regardless of how it is implemented.

DonLakeFlyer:
I think wait until QGamePad is ready for prime time. Then cut over to it fully. Then pay the price for all the other crap you need to deal with when you add a new module.

DonLakeFlyer:
@nanthony21 So I think the answer is to close this for now?

nanthony21:
Yes. I think it will probably be a while before QGamePad becomes more stable.

patrickelectric:
@nanthony21 this PR is working ? I would like to continue this if possible.

nanthony21:
I'm not able to test it again at the moment but I would assume that it still works. I would encourage you to try it out.

Unfortunately support for android gamepads won't be able to be merged into master until support for qgamepad is improved. But that doesn't mean you can compile a version of QGC with this PR in it.

patrickelectric:
Hello Nick, 

QGamePad has compatibility with android since Qt 5.10, or at least it is what I can see in this folder here:
https://github.com/qt/qtgamepad/tree/9152a285cdb072bc622e693f1a74f2ee3eda5503/src/plugins/gamepads

If I'm not wrong, you did this PR in qgc while it was in 5.9, right ?
I would like **very much** to use it and remove SDL framework.

nanthony21:
If I remember correctly the issue wasn't with android compatibility. The problem was that qgamepad was buggy in windows and we didn't want to switch over unless it was going to work with all major platforms. However, if those issues are fixed now then I think it would be a great idea to make the switch, you may want to check with @donlakeflyer before investing too much time in it though.

I probably won't be able to work on this anytime in the near future (i'm actually not even using USB game pad control anymore). But if you decide to take this on I'll be happy to  help if I can.

DonLakeFlyer:
I"m fine with switching but it's going to need a ton of testing across all supported OS to verify it works. Which is something I don't have the time for.

On Aug 16, 2018, at 7:01 PM, Nick Anthony <notifications@github.com<mailto:notifications@github.com>> wrote:


If I remember correctly the issue wasn't with android compatibility. The problem was that qgamepad was buggy in windows and we didn't want to switch over unless it was going to work with all major platforms. However, if those issues are fixed now then I think it would be a great idea to make the switch, you may want to check with @DonLakeFlyer<https://github.com/DonLakeFlyer> before investing too much time in it though.

I probably won't be able to work on this anytime in the near future (i'm actually not even using USB game pad control anymore). But if you decide to take this on I'll be happy to help if I can.

PR #6064, opened 2018-01-28
QGC: Solve problem with limitAngleToPMPId when angle is bigger than PI

patrickelectric:

### DIFF
--- a/src/QGC.cc
+++ b/src/QGC.cc
@@ -61,14 +61,14 @@ double limitAngleToPMPId(double angle)
         {
             while (angle < -M_PI)
             {
-                angle += M_PI;
+                angle += 2.0f * M_PI;
             }
         }
         else if (angle > M_PI)
         {
             while (angle > M_PI)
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Should this go into Stable 3.3 as well?

patrickelectric:
I believe so, this function should return values from 0-2pi

DonLakeFlyer:
Ok. Can you submit the change against Stable_V3.3 branch as well then?

DonLakeFlyer:
I'm going to hold this from merging until there is the same pull on the Stable side. That way it won't get lost.

patrickelectric:
@DonLakeFlyer done.

patrickelectric:
@DonLakeFlyer This commit will be cherry-pick to master (or rebased) ?

DonLakeFlyer:
I'll do a merge from Stable to master at some point. So we are all good.

PR #6072, opened 2018-01-31
Add support for HIGH_LATENCY2

DonLakeFlyer:
* Flight modes still not implemented. Waiting on message change.
* Vehicle are created from HIGH_LATENCY2 messages.
* Also includes various cauterization of UI things which should not be available on high latency link:
  * Parameter and mission are not requested.
  * No parameter list diplay.
  * Mission upload/download not allowed

### DIFF
--- a/src/FactSystem/ParameterManager.cc
+++ b/src/FactSystem/ParameterManager.cc
@@ -81,7 +81,17 @@ ParameterManager::ParameterManager(Vehicle* vehicle)
     // Ensure the cache directory exists
     QFileInfo(QSettings().fileName()).dir().mkdir("ParamCache");
 
-    refreshAllParameters();
+    if (_vehicle->highLatencyLink()) {
+        // High latency links don't load parameters
+        _parametersReady = true;
+        _missingParameters = true;
+        _initialLoadComplete = true;
+        _waitingForDefaultComponent = false;
+        emit parametersReadyChanged(_parametersReady);
+        emit missingParametersChanged(_missingParameters);
+    } else {
+        refreshAllParameters();
+    }
 }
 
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
@dogmaphobic @dagar FYI

acfloria:
@DonLakeFlyer thanks for implementing it.

I did some test in SITL with the PX4 implementation and there are two things I noticed:
    - The altitude and relative altitude are always 0 although I verified that the HIGH_LATENCY2 message contains a non-zero altitude.
   - Would it be possible to adopt the display for example for the climb rate while in the high_latency mode? As in the message only the maximum absolute value is sent it leads to can lead to confusion when during a landing a positive value is displayed without more information.

DonLakeFlyer:
> Would it be possible to adopt the display for example for the climb rate while in the high_latency mode?

Not quite sure what you mean. If you want the instrument panel to automatically add climb rate while in high latency mode. That isn't currently possible and would be a lot of work. The user can add it manually if they want it.

> As in the message only the maximum absolute value is sent it leads to can lead to confusion when during a landing a positive value is displayed without more information.

Don't undestand what this mean. Currenty climb_rate shows the value that comes across telemetry.

acfloria:
Some variables do not contain the newest value but the maximum magnitude observed since the last high latency message. For example [eph, epv, and climb rate](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L4335-L4338). So would it be possible to change the description of those variables while in the high latency mode. For example:
climb rate -> max climb rate magnitude

DonLakeFlyer:
Ah, I see. That would not be an easy thing to do in QGC.

PR #6079, opened 2018-02-01
Finer control for plugins

dogmaphobic:
* Split save mission to Json file in two so you have the option to saving it to a file or to a QByteArray Json blob.
* Virtualize all camera functions to allow deeper access.


dogmaphobic:
10 commits? Not sure what went on. The diff looks correct though.

PR #6091, opened 2018-02-01
More work in high latency

DonLakeFlyer:
* Command ack timeout based on normal/high latency link (2 minute timeout for high latency)
* Correctly update ack timeout and high latency bit at correct times
* Fix altitude values from HIGH_LATENCY2. Absolute altitude will display correctly. Relative altitude is NaN'ed to get good display of no value there.

screen shot 2018-02-01 at 3 10 46 pm

Related to #6072


DonLakeFlyer:
Missed:
* Don't sent GCS heartbeat to vehicle through high latency link

PR #6096, opened 2018-02-02
Add highLatency property to LinkConfiguration

acfloria:
- Add a highLatency variable to the LinkConfiguration and set the variable in the LinkInterface accordingly.
- Wenn manually connecting a link there is now a checkbox for UDP, TCP, and serial links to specify if the new link is a high latency link.

### DIFF
--- a/src/comm/LinkConfiguration.cc
+++ b/src/comm/LinkConfiguration.cc
@@ -37,6 +37,7 @@ LinkConfiguration::LinkConfiguration(const QString& name)
     , _name(name)
     , _dynamic(false)
     , _autoConnect(false)
+    , _highLatency(false)
 {
     _name = name;
     if (_name.isEmpty()) {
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
@dogmaphobic You two guys should discuss this. Since it differs from how Gus and I discussed how to plumb the hgh latency bit. With the fact that a real high latency implementation needs a custom link implementation.

DonLakeFlyer:
The concept is whether it is possible to support high latency without needing to implementation a custom link

DonLakeFlyer:
We may need a Skype discussion to talk throught this.

acfloria:
I would vote that it should be possible to use high latency without implementing a custom link if a standard link could be used for the high latency communication. If everyone implements their custom link probably the same work is done multiple time which is a waste of resources.

I agree that a Skype call would be the best way to find a common solution.

acfloria:
UDP:
udp
Serial:
serial
Log Replay:
log_replay

DonLakeFlyer:
Can you move the checkbox to be at the bottom of the settings instead of the top?

acfloria:
Ok, I moved the checkbox to the bottom of the configurations. You can see the updated screenshots below.

Serial new:
serial_new

TCP new:
tcp_new

Issue #6103, opened 2018-02-03
Crash in terrain code (Stable 3.3)


dogmaphobic:
I've been getting crashes randomly and today I was able to tell where (but not why). The crash occurs here:

https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain.cc#L149

Adding it here to keep track.

DonLakeFlyer:
Yeah, I hit that last night as well then forgot about it. I know what it is.

PR #6104, opened 2018-02-03
Terrain crash

DonLakeFlyer:
Fix for #6103


DonLakeFlyer:
Need to move to master as well.

PR #6108, opened 2018-02-04
Clickable deadzone

nanthony21:
This makes it so that joystick deadband for each axis can be adjusted by clicking on the corresponding axis on the right side of the screen and dragging vertically. I have also added a short explanation of the deadband setting process which will appear below the deadband checkbox when it is enabled.

While the current method of setting works pretty well I've found that it often times doesn't end up quite right. This will make it easy to adjust the settings without going through the whole calibration process again.

### DIFF
--- a/src/VehicleSetup/JoystickConfig.qml
+++ b/src/VehicleSetup/JoystickConfig.qml
@@ -537,6 +537,21 @@ SetupPage {
                                     onClicked:  controller.deadbandToggle = checked
                                 }
                             }
+                            Row{
+                                width: parent.width
+                                spacing: ScreenTools.defaultFontPixelWidth
+                                visible: advancedSettings.checked
+                                QGCLabel{
+                                    width:       parent.width * 0.85
+                                    font.pointSize:     ScreenTools.smallFontPointSize
+                                    wrapMode:           Text.WordWrap
+                                    text:   qsTr("Deadband can be set during the first ") +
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Screenshot?

nanthony21:
screen shot 2018-02-04 at 11 35 59 pm
A message appears below the deadband checkbox when it is enabled to provide a brief description of how the deadband can be set.


screen shot 2018-02-05 at 12 01 22 am
Axis 1 is green because it has been clicked on. Dragging up or down will adjust the deadband. When the mouse is released the deadband indicator will turn back to red.

DonLakeFlyer:
"dragging vertically"? That seems a bit strange given the indicators are horizontal? Done so you can see what you are doing?

Also wondering about the ui and tablet usage. Does it work with touch?  Have you tried that?

nanthony21:
I initially had it dragging horizontally but since the axis monitors are on the right side of the screen there is very little room to drag to the right if the application is in full-screen. Small added benefits are that it is easier to see what your are doing and that it is intuitive that up=increase deaband, down=decrease deadband, with left/right it is less intuitive.

I've tested it on a tablet and it worked fine. The axis monitors are small so it can be a little difficult to click on them, but it's not unreasonable. The sensitivity is such that it's very easy to accurately adjust the deadband without overshooting, even on a touchscreen.

DonLakeFlyer:
Ok, thanks Nick.

PR #6110, opened 2018-02-05
UDP Change

dogmaphobic:
The UDP link keeps a list of every host that sends a message and when replying on that link, it sends all messages to every host it received messages from (the client list).

The problem was that up to now, the list was a map (`host["address"] = port`). That meant if a vehicle had a component that sends messages from a different port, it would replace the port on the existing host (in the map). When QGC would sent any message to the vehicle, it would only get sent to the port of whichever component sent a message last.

This fix replaces the map with an actual list of address/port pairs. A vehicle with one or more components sending UDP messages from different ports will all be added to the "client list" and when QGC sends a message, it now gets sent to all components within a same address (not just the last to have sent a message.)

This is separate from the list of "target" addresses. Target addresses are those manually entered when creating an UDP link. QGC will send messages to these regardless of having received a message from them or not.

@DonLakeFlyer This is also bringing in Stable_V3.3. That terrain bug was crashing way too often so I merged those changes during development. Let me know if you want to exclude the merge for whatever reason.

### DIFF
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -493,6 +493,7 @@ void MissionController::removeAll(void)
 {
     if (_visualItems) {
         _deinitAllVisualItems();
+        _visualItems->clearAndDeleteContents();
         _visualItems->deleteLater();
         _settingsItem = NULL;
         _visualItems = new QmlObjectListModel(this);
--- a/src/MissionManager/SurveyMissionItem.h
+++ b/src/MissionManager/SurveyMissionItem.h
@@ -296,7 +296,7 @@ private:
     static const char* _jsonFixedValueIsAltitudeKey;
     static const char* _jsonRefly90DegreesKey;
 
-    static const int _hoverAndCaptureDelaySeconds = 1;
+    static const int _hoverAndCaptureDelaySeconds = 2;
 };
 
 #endif
--- a/src/QGC.cc
+++ b/src/QGC.cc
@@ -61,14 +61,14 @@ double limitAngleToPMPId(double angle)
         {
             while (angle < -M_PI)
             {
-                angle += M_PI;
+                angle += 2.0f * M_PI;
             }
         }
         else if (angle > M_PI)
         {
             while (angle > M_PI)
             {
-                angle -= M_PI;
+                angle -= 2.0f * M_PI;
             }
         }
     }
--- a/src/Terrain.cc
+++ b/src/Terrain.cc
@@ -33,6 +33,8 @@ TerrainBatchManager::TerrainBatchManager(void)
 void TerrainBatchManager::addQuery(ElevationProvider* elevationProvider, const QList<QGeoCoordinate>& coordinates)
 {
     if (coordinates.length() > 0) {
+        qCDebug(ElevationProviderLog) << "addQuery: elevationProvider:coordinates.count" << elevationProvider << coordinates.count();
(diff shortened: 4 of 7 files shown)


NaterGator:
A thought about this: from a routing standpoint it may make sense to have the mavlink decoder fire a callback that notifies the incoming *Link of the SysId/CompId detected so that links can register proper routing tables to that SysId/CompId. 

Maybe too big of a change for this commit, but something we should consider?

dogmaphobic:
> qDeleteAll(_sessionTargets);

I didn't know this existed. Learning something new everyday.

> A thought about this: from a routing standpoint it may make sense to have the mavlink decoder fire a callback that notifies the incoming *Link of the SysId/CompId detected so that links can register proper routing tables to that SysId/CompId.

Yes, this is what I meant by this needing more scrutiny. There are different scenarios and some won't work even with this change. This needs to be discussed at a protocol level, including the autopilot.

I always assumed one link shared by multiple components (one address, one port). You send a message to a given component ID and whoever is receiving (listening on the port) would route it to the proper component, however way it is connected to it. This is normally the case for when you are connected directly to the autopilot and send a command to some component off of it. Another scenario is having a companion computer in between. You "talk" to the companion computer, who in turn routes the messages to the autopilot and/or components connected to it.

I just never expected to see a component bound to its own UDP port and sending messages to the GCS from this extra port. This creates a bit of a mess. Without this change, you would end up talking to one or the other but never both. With this change, QGC will "properly" send the messages to both at the same time. Ideally, it should make a distinction and pair the compID to the port, which is nowhere simple, as this is something unique to UDP and the mavlink decoder is link agnostic.

To make matters worse, I noticed that the autopilot will "route" some messages off to the component but not all. That means the component will receive these messages in duplicate.

The way I see it, if a component has its own "link" (a separate port in this case), nothing sent to the autopilot should be "routed" to this component. Otherwise, why should the component have its own link to begin with?

And lastly, in the case of Gazebo for instance, I have two network interfaces: Ethernet and WiFi. I see things coming from 3 separate addresses (en0, wan0 and localhost). This is random as it depends what address each one got when using `INADDR_ANY`. QGC handles this on the receiving end by translating everything to 127.0.0.1 but I don't think the various modules in Gazebo do this on their own. I end up seeing messages from autopilot to component in triplicate.

DonLakeFlyer:
> You "talk" to the companion computer, who in turn routes the messages to the autopilot and/or components connected to it.

I'm not sure that is how it work. I think you send your companion computer messages with a specific component id. These are sent to the autopilot, who forwards them to the companion. QGC<->autpilot<->companion.

I still don't quite understand how all this work but it just doesn't "smell" right to me. May need a real Skype discussion.

dogmaphobic:
> I'm not sure that is how it work. I think you send your companion computer messages with a specific component id. These are sent to the autopilot, who forwards them to the companion. QGC<->autpilot<->companion.

Both ways work. It depends on where the link is handled. If you have Wifi or something that must be handled outside the autopilot, a companion computer would handle the link to the GCS and route the traffic to the autopilot. That's how the ESP8266 works for example. If the link is direct to the autopilot, then it's the other way around. It shouldn't matter. As long as it's one link (one port) none of this matters. The issue is if the autopilot and the, say camera, have their own separate "links" (in this case, same _link_ different ports). This is also fine as long as commands sent to the camera compID received by the autopilot are not "routed" from the autopilot to the camera. Otherwise the camera will get everything twice.

PR #6117, opened 2018-02-06
Joystick axes can now be monitored even when joystick control is disabled

nanthony21:
The current Joystick implementation uses a tristate enumerator to indicate the calibration state of the joystick. However it seems that the state is never set to CalibrationModeOff except sometimes when the Joystick is first plugged in. This makes the use of this tristate enumerator confusing and results in a situation where disabling the joystick not only stops sending out ManualControl packets but also stops processing input from the joystick all together.

This pull request replaces the tri-state enumerator with two boolean values. One to indicate whether the joystick is currently being calibrated and another to indicate whether the joystick should be transmitting ManualControl packets. This allows the user to monitor the value of each joystick axis in QGroundControl even when joystick control is disabled.


DonLakeFlyer:
Makes sense

Issue #4564, opened 2017-02-17
Disconnecting a link when multiple links are manually connected to one vehicle


mhugentobler:
I connect to my vehicle via a Serial and a UDP link. When both links are connected and I want to disconnect only the Serial link by clicking "disconnect" in the link settings, often, tough not every time, I cannot reconnect this link, because the "connect" button is not clickable. The only thing that helps is to wait for the "communication lost" message and then click "disconnect" in the main tool bar. After that obviously both of my links are disconnected and not only the Serial one. 

In v3.0 this was not an issue. Whenever you would disconnect the Serial link while the UDP link was still connected, you could reconnect at any time, while having persistent connection via UDP.
Since v3.1 this does not work reliably anymore.

DonLakeFlyer:
Can you try a daily build? There have been a number of fixes around this which have been made since the last stable.

mhugentobler:
Just tried it with current master (as of now). The problem is still there: connect serial -> connect udp, then disconnect and reconnect serial a couple times. Usually at the second disconnection of the serial link the problem described above occurs.

DonLakeFlyer:
Sorry I misread your issue. Thought it was multi-vehicle.

DonLakeFlyer:
You should turn off all auto-connect from Settings and try this again using manual links only. Most likely auto-connect getting in the way.

mhugentobler:
I have never used auto-connect before (always all disabled), so I suppose the issue occurs due to something else.

DonLakeFlyer:
I understand what is going on here now. It's a big change to fix so it will go in first for 3.3.

DonLakeFlyer:
@mhugentobler Some of the connect/disconnect semantics have been reworked. Is this still happening with latest daily builds?

mhugentobler:
Since I am no longer working with QGC I cannot comment on this. Maybe @acfloria knows more

acfloria:
@DonLakeFlyer I checked it today with two serial links.

1. I disabled autoconnect.

2. Open a serial connection to ttyUSB0

3. Open a serial connection to ttyACM0. 

4. Try to disconnect ttyUSB0. On the link screen it still shows as connected but after some time I get the `Communication lost` message, shortly followed by a `Communication regain` notification. Though even after this notification I cannot for example upload a mission. I can not edit ttyUSB0 as it still is shown as connected. 

[...]

PR #6121, opened 2018-02-07
Fix for #4564

acfloria:
When two links are connected to one vehicle and the priority link was deleted first, the _priorityLink still contained a shared pointer to the deleted link which caused [#4564](https://github.com/mavlink/qgroundcontrol/issues/4564).

Now if the priority link is deleted and at least one other link is available. The first of the available links is set as the priority link.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -1492,7 +1492,16 @@ void Vehicle::_updatePriorityLink(void)
     }
 #endif
 
-    if (!newPriorityLink && !_priorityLink.data() && _links.count()) {
+    bool found_priority_link = false;
+    if (!newPriorityLink && _priorityLink.data()) {
+        for (int i=0; i<_links.count(); i++) {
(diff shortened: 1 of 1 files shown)


src/Vehicle/Vehicle.cc
```diff
@@ -1492,7 +1492,16 @@ void Vehicle::_updatePriorityLink(void)
     }
 #endif
 
-    if (!newPriorityLink && !_priorityLink.data() && _links.count()) {
+    bool found_priority_link = false;
+    if (!newPriorityLink && _priorityLink.data()) {
+        for (int i=0; i<_links.count(); i++) {
+            if (_priorityLink->getLinkConfiguration() == _links[i]->getLinkConfiguration()) {
```
DonLakeFlyer: Is there a reason this is not ```if (_priorityLink.data() != link)```?

DonLakeFlyer: Reason I ask is that in looking at the code for this method it's really confusing. Working on rewriting with your fix.

acfloria:
Replaced by [#6127](https://github.com/mavlink/qgroundcontrol/pull/6127)

PR #6127, opened 2018-02-08
Fix bugs in updating priority link

DonLakeFlyer:
* Also handles correct prioritization of high latency links

Replacement for #4564.

@acfloria Can you take a look at this. I believe this fixes the leaking shared pointer. I can now disconnect two links on the same vehicle. It also makes the code way more understandable. The previous code was making my head hurt.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -1500,8 +1500,11 @@ void Vehicle::_updatePriorityLink(void)
     // Check for the existing priority link to still be valid
     for (int i=0; i<_links.count(); i++) {
         if (_priorityLink.data() == _links[i]) {
-            // Still valid
-            return;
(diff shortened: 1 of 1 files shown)


acfloria:
Tested it and deleting the priority link works fine.

acfloria:
After thinking about it a bit more:
What is the reason you abort if the priority link changed. For example if we first connected a high latency link and later a serial link then we would switch to the serial link as priority link.

DonLakeFlyer:
> For example if we first connected a high latency link and later a serial link then we would switch to the serial link as priority link.

Yup, good point.

DonLakeFlyer:
> What is the reason you abort if the priority link changed. For example if we first connected a high latency link and later a serial link then we would switch to the serial link as priority link.

Fixed by last commit

acfloria:
Looks good

PR #6130, opened 2018-02-09
Support remainder of HIGH_LATENCY2 fields

DonLakeFlyer:
* Flight mode now works
* Changed to new FLAGS

**Important note for PX4 firmware**
Although PX4 only uses 16 bits for custom_mode flags. They are in the high 16 bits of the 32 bit custom_mode value of HEARTBEAT. This means that you cannot directly copy between HEARTBEAT.custom_mode and HIGH_LATENCY2.custom_mode and expect it to work. In order to do this I'm using the following structure which needs to be matched on the PX4 implementation of HIGH_LATENCY2.custom_mode when sending the message. Or in other words something like HIGH_LATENCY2.custom_mode = cm.custom_mode_hl.

```
union px4_custom_mode {
	struct {
		uint16_t reserved;
		uint8_t main_mode;
		uint8_t sub_mode;
	};
    struct {
        uint16_t reserved_hl;
        uint16_t custom_mode_hl;
    };
    uint32_t data;
	float data_float;
};
```

### DIFF
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -620,4 +620,9 @@ QGCCameraControl* FirmwarePlugin::createCameraControl(const mavlink_camera_infor
     return NULL;
 }
 
+uint32_t FirmwarePlugin::highLatencyCustomModeTo32Bits(uint16_t hlCustomMode)
+{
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
I also forgot to say that if QGC gets a HIGH_LATENCY2 message it assumes the vehicle is armed since it doesn't really know.

PR #6138, opened 2018-02-13
[WIP] Initial gimbal control

dagar:


dagar:
This is more or less working. There's definitely more work to do, but if there's nothing objectionable can we talk about getting it to a safe mergeable state?

src/Vehicle/Vehicle.cc
```diff
@@ -2511,20 +2592,45 @@ void Vehicle::setCurrentMissionSequence(int seq)
 
 void Vehicle::sendMavCommand(int component, MAV_CMD command, bool showError, float param1, float param2, float param3, float param4, float param5, float param6, float param7)
 {
-    MavCommandQueueEntry_t entry;
+    bool updated = false;
+
+    if (_mavCommandQueue.size() > 1) {
```
dagar: @dogmaphobic please review this part carefully. To prevent excessive command queuing any additional duplicate commands are collapsed into what's already queued (but not in flight). You only care about the last command getting out and acknowledged.

dogmaphobic:
@dagar 

May latest changes:

* Adjust frequency based on link type. For normal (high bandwidth) links, use the usual 25Hz message stream. For slow links, drop the frequency down to 5Hz. Disable it altogether for high latency links.
* Change gimbal handling to timer based instead of event driven. Unlike joystick messages, these are MAVLink commands, which will cause the autopilot to send back an ACK. Event driven would saturate the link with way too many messages. The timer is fixed at 4Hz as this is more than enough for controlling the gimbal. And even then, gimbal changes are only sent if they actually occurred (as opposed to Joysticks, which continuously send a stream of messages).

dagar:
This looks good. Let's verify things still work well for @sanderux's gimbal and squash merge this.

sanderux:
Latency over mavlink made it impractical in rate mode.
Tried position mode and that works better with the latency, now sliders would help a lot to get full range and better accuracy.

dogmaphobic:
It's quite stale by now. Let me sync it up with master and see how it goes...

dogmaphobic:
@dagar I've got the branch up-to-date with master, resolved a bunch of conflicts and fixed some stupid error with GStreamer trying to log to the root directory (that thing has to be fixed properly or go away). I've tested with a Typhoon and it all seems to be working. Ready to merge if you are.

P.S. Hadn't I changed some of the virtual joystick graphics? I have a vague recollection of doing something but I can't say for sure.

dagar:
> What if the command parameters contradict each other? A same command but with a parameter that enables/disables something?

This was intentional, although I don't know if it's the best choice in all scenarios. I was imagining things like changing the flight mode then immediately changing to something else.

Do you see a scenario where it's problematic?

dogmaphobic:
> Do you see a scenario where it's problematic?

Nothing off the top of my head. @DonLakeFlyer would probably have a better idea. It would have to be something sent really close to each other, otherwise it would not be in the "queue".

```
           if (entry.command == command) {
                entry.component = component;
                entry.command = command;
```

What if two consecutive (same) commands are sent to two different component IDs?
And just being OCD, `entry.command = command` is not necessary :)

dagar:
Yes different components would be a valid case. 

This really only matters when the link becomes heavily congested or lossy and these commands start piling up.

dagar:
Rebase is nasty. The diff isn't that big, so it's probably easier overall (and safer) to reapply manually.

dogmaphobic:
Yup, hand picking the changes :P

RyanHurst:
Is anyone actively working on this? If I wanted to start working on this, should I start here, or should I start from scratch in a new PR?

jafrado:
I tested this for Gus a few months ago, it works well with the right hardware. IIRC he was going to rebase and merge but there has been so much rewrite probably Gus should comment here.

dogmaphobic:
There is work being done on this. A new PR is coming shortly.

PR #6154, opened 2018-02-22
Mouse Bleed Through

dogmaphobic:
Create a "Dead Mouse Area" control to eat mouse events underneath desired areas. This is to prevent mouse events from leaking through controls underneath the area.
Replace original "NoMouseThroughRectangle" as this is more versatile.
Prevent mouse through events from leaking underneath the Fly View tool strip.


akontsevich:
@dogmaphobic Could You clarify please: why this control is necessary, why these mouse events  leaks are happen? Is it some QTBUG, feature or some problem with outdated QtQuick.Controls 1 used in QGC? Or there is some problem with `FlightMap` where these are propagated for example?

dogmaphobic:
They are used over wide areas upon which controls are placed. See example below. We make the entire area in red as `DeadMouseArea` so it will capture all mouse events not captured by elements higher up in the hierarchy. For example, if you were to touch in between those buttons (yellow arrow), the touch would be sent to the layer below (the map), and some times, you just don't want that to happen. It's not a bug of Qt or anything. It's just a way to prevent mouse events from reaching lower levels within a defined region.

Screen Shot 2019-11-25 at 7 29 08 PM

akontsevich:
> It's not a bug of Qt or anything. 

Thanks much! I see. So standard behavior in whether Qt Widgets or QML is to propagate event down until some element says `event accepted =  true` - correct? And probably components like `Button` does this automatically?

PR #6155, opened 2018-02-22
MainWindowHybrid: Close app if something goes wrong with item

patrickelectric:

### DIFF
--- a/src/ui/MainWindowHybrid.qml
+++ b/src/ui/MainWindowHybrid.qml
@@ -22,7 +22,12 @@ Item {
     }
 
     function attemptWindowClose() {
-        mainWindowInner.item.attemptWindowClose()
+        try {
+            mainWindowInner.item.attemptWindowClose()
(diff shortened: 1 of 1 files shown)


dogmaphobic:
Makes sense

DonLakeFlyer:
It causes the mainwindow to be closed before shutdown dialog prompt is displayed. So if you cancel shutdown you are left in a state where QGC is running but there is no main window.

patrickelectric:
This approach solved some issues that I was having. If something goes wrong with QGC, it's almost impossible to close it, the close signal is handled by it owns components. Theoretically this PR should not give any problems, sorry about that.

DonLakeFlyer:
The thing is that ```mainWindowInner.item.attemptWindowClose()``` does next to nothing. All it does it pop a yes/no dialog and then set a flag. So hard to see how that would ever throw an exception.

PR #6157, opened 2018-02-23
Vehicle handles its own heartbeats (Not to merge yet)

dogmaphobic:
* Let the vehicle instance control when to send heartbeats (GCS to Vehicle)
* Allow firmware plugin to define heartbeat timeout and frequency for high latency links

Until now, MultiVehicleManager handled all heartbeats from QGC out to the connected links. This moves that logic to the individual vehicles themselves. That's because only they know how to properly handle high latency links.

I'm not merging this until Don is back and have a chance to check and see if this doesn't mess up anything else.

### DIFF
--- a/src/Vehicle/Vehicle.h
+++ b/src/Vehicle/Vehicle.h
@@ -752,6 +752,8 @@ public:
     void _setHomePosition(QGeoCoordinate& homeCoord);
     void _setMaxProtoVersion (unsigned version);
 
+    QList<LinkInterface*>       links() { return _links; }
+
     /// Vehicle is about to be deleted
     void prepareDelete();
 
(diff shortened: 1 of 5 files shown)


dogmaphobic:
@DonLakeFlyer I forgot about this. Can you see if this makes sense?

src/Vehicle/Vehicle.h
```diff
@@ -752,6 +752,8 @@ class Vehicle : public FactGroup
     void _setHomePosition(QGeoCoordinate& homeCoord);
     void _setMaxProtoVersion (unsigned version);
 
+    QList<LinkInterface*>       links() { return _links; }
```
DonLakeFlyer: Really needed? I assume private. If public I don't really want vehicle links to bleed out because there is some reference counting stuff that could get screwed up if they hand out to other parts of code.

DonLakeFlyer:
Other than my own thing looks ok.

dogmaphobic:
This is so old that's mummified by now. If this was still an issue, we would have heard about it...

Issue #6165, opened 2018-02-27
apt-get install gstreamer1.0* is not working on 16.04 


barzanisar:
I had the same issue as in https://github.com/mavlink/qgroundcontrol/issues/4303

On a ubuntu 16.04 64bit system the

    sudo apt-get install gstreamer1.0*

fails because there is a i386 dependency somewhere. 

The last few lines of the error were:

**The following packages have unmet dependencies:
 gstreamer1.0-hybris:i386 : Depends: libmedia1:i386 but it is not going to be installed
E: Unable to correct problems, you have held broken packages.**

I tried this solution:

list=$(apt-cache --names-only search ^gstreamer1.0-* | awk '{ print $1 }' | grep -v gstreamer1.0-hybris)
sudo apt-get install $list

yet, I was getting the same error  when I built QGC in Qt creator:
 error: gstreamer-video-1.0 development package not found

so I installed the following package which resolved the issue:

sudo apt-get install libgstreamer-plugins-base1.0-dev

PR #6166, opened 2018-02-27
added additional packages needed to build QGC

barzanisar:
These packages are needed to solve the issue described in https://github.com/mavlink/qgroundcontrol/issues/6165

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ You **need to install Qt as described below** instead of using pre-built package
     * Windows: Make sure to install VS 2015 32 bit package.
 
 ###### Install additional packages:
-* Ubuntu: sudo apt-get install speech-dispatcher libudev-dev libsdl2-dev
+* Ubuntu: sudo apt-get install speech-dispatcher libudev-dev libsdl2-dev libgstreamer1.0-0 gstreamer1.0-plugins-base libgstreamer-plugins-base1.0-dev gstreamer1.0*
 * Fedora: sudo dnf install speech-dispatcher SDL2-devel SDL2 systemd-devel
 * Arch Linux: pacman -Sy speech-dispatcher
 * Windows: [USB Driver](http://www.pixhawk.org/firmware/downloads) to connect to Pixhawk/PX4Flow/3DR Radio


README.md
```diff
@@ -41,7 +41,7 @@ You **need to install Qt as described below** instead of using pre-built package
     * Windows: Make sure to install VS 2015 32 bit package.
 
 ###### Install additional packages:
-* Ubuntu: sudo apt-get install speech-dispatcher libudev-dev libsdl2-dev
+* Ubuntu: sudo apt-get install speech-dispatcher libudev-dev libsdl2-dev libgstreamer1.0-0 gstreamer1.0-plugins-base libgstreamer-plugins-base1.0-dev gstreamer1.0*
```
hamishwillee: I assume similar packages will not be needed for Fedora, archilinux, windows?

barzanisar: I didn't try on those. I tried in ubuntu which gave me some build errors and found out that these packages were also needed. If I later try with windows, I'll update this document accordingly.

hamishwillee: Works for me. Added @DonLakeFlyer as reviewer :-)

dogmaphobic:
Thanks! This was long needed. Regarding Windows, only Linux has this issue as the packages are split into a zillion separate smaller packages with inter/extra dependencies. All other platforms are shipped as a single, monolithic set of libraries.

I would assume other Linux distributions would require equivalent work but the packages should be the same. I would imagine just the package manager would be different.

DonLakeFlyer:
The packages for gstreamer were here: https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoStreaming/README.md. Not on the main README. Not sure having them be in two places (and different) is a good idea?

DonLakeFlyer:
Also the set here https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoStreaming/README.md was working. Is this setup better?

DonLakeFlyer:
These updated instructions fail on a clean up to date 16.04 image. Working on correct set. Then I'll revert this pull and update the streaming side readme. Something changed in latest 16.04 updates to break things from what was previously verified to work during stable 3.3 rollout.

PR #6177, opened 2018-03-06
Joystick: Add circle correction option

patrickelectric:
Some digital joysticks have already this correction
Add habilit to add or remove circle correction

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -25,6 +25,7 @@ const char* Joystick::_throttleModeSettingsKey =        "ThrottleMode";
 const char* Joystick::_exponentialSettingsKey =         "Exponential";
 const char* Joystick::_accumulatorSettingsKey =         "Accumulator";
 const char* Joystick::_deadbandSettingsKey =            "Deadband";
+const char* Joystick::_circleCorrectionSettingsKey =    "Circle_Correction";
 const char* Joystick::_txModeSettingsKey =              NULL;
 const char* Joystick::_fixedWingTXModeSettingsKey =     "TXMode_FixedWing";
 const char* Joystick::_multiRotorTXModeSettingsKey =    "TXMode_MultiRotor";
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
I need way more details for these pulls. I don't have the time to figure out what this does by looking at the code. Screenshots, more explanation...

patrickelectric:
This code allow the possibility to not use diagonal corrections in digital joysticks.
image
This image shows how the analog works, it's necessary to limit the X and Y distance to make the diagonal.
In digital joysticks this correction is not necessary.
In a analog joystick X and Y can have max value of 1, but in the diagonal it'll have 0.70710678118 ou cos(pi/4) for both. Meanwhile the digital joystick will have the maximum value 1 in the X, Y and in the diagonal :video_game:

DonLakeFlyer:
Ok, so the user has to know whether they can turn this on or not based on the joystick they are using right? Looks like you added checkbox which shows up all the time? Can't quite tell without a screenshot. Seems like this would be better under Advanced to not clutter UI?

DonLakeFlyer:
I guess it depends on how command it will be to use this setting?

patrickelectric:
> Ok, so the user has to know whether they can turn this on or not based on the joystick they are using right? Looks like you added checkbox which shows up all the time? Can't quite tell without a screenshot. Seems like this would be better under Advanced to not clutter UI?
Yes, now with the last commit is in under advanced menu.

> I guess it depends on how command it will be to use this setting?
This allow any user to use a digital joystick without losing 30% of precision in each axis.

DonLakeFlyer:
Ok, awesome thanks. Will merge in a bit. Hand holding some other stuff through rifght now.

DonLakeFlyer:
@patrickelectric This pull is incorrect. Previously circle correction was on by default. Now it is off by default. Needs to go back to on by default. Can you fix this?

patrickelectric:
Yep, I'll take a look.

patrickelectric:
https://github.com/mavlink/qgroundcontrol/pull/6284

PR #6184, opened 2018-03-07
Allow setting MAVLINK dialect in user_config.pri

mantelt:
I propose to add this small change to the qgroundcontrol project file in order to allow setting a different MAVLINK dialect in user_config.pri by adding a line ```MAVLINK_CONF = dialect```
Otherwise, it either needs to be set as a qmake argument (might get lost when setting up qt project from scratch), or hard-coded into QGCExternalLibs.pri, which might cause merge conflicts at some point.

### DIFF
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -21,14 +21,32 @@ WindowsBuild {
 # In the mean time, it’s possible to define a completely different dialect by defining the
 # location and name below.
 
+# check for user defined settings in user_config.pri if not already set as qmake argument
 isEmpty(MAVLINKPATH_REL) {
-    MAVLINKPATH_REL = libs/mavlink/include/mavlink/v2.0
+    exists(user_config.pri):infile(user_config.pri, MAVLINKPATH_REL) {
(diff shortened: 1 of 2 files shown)


dogmaphobic:
You might was well add all three variables, allowing full control over the MAVLink dialect. That would allow not only to define a different dialect, but also to define a new path altogether so you can point to some custom built MAVLink.

https://github.com/mavlink/qgroundcontrol/blob/master/QGCExternalLibs.pri#L24-L32

In addition, I would move your additions to `QGCExternalLibs.pri` itself, right before the variables evaluation. This would make it more logical as they are next to each other and easier to understand.

dogmaphobic:
By the way, disregard the Travis build failure below. Travis has a vey high failure rate with macOS all on its own.

mantelt:
Addressed comment by @dogmaphobic :
- [x] Moved import of MAVLINK related config to QGCExternalLibs.pri
- [x] now checking for all three MAVLINK variables

thanks for the input, good ideas!

dogmaphobic:
Thanks!

PR #6186, opened 2018-03-07
Multi camera Support

dogmaphobic:
Added (UI) support for vehicles exposing more than one camera.

@lbegani I've not been able to test with more than one camera as I only have one at a time (per vehicle). Please, let me know how this goes.

The selector is under camera settings:

screen shot 2018-03-07 at 5 08 18 pm

screen shot 2018-03-07 at 5 08 29 pm


src/FlightMap/Widgets/CameraPageWidget.qml
```diff
@@ -202,6 +203,23 @@ Column {
...
+                        QGCLabel {
+                            text:           qsTr("Camera Selector:")
+                            width:          _labelFieldWidth
+                            anchors.verticalCenter: parent.verticalCenter
+                        }
+                        QGCComboBox {
+                            id:             cameraSelector
+                            model:          _isCamera ? _dynamicCameras.cameraLabels : []
+                            width:          _editFieldWidth
+                            onActivated:    _dynamicCameras.cameras = index
```
lbegani: This throws an error - `qrc:/qml/CameraPageWidget.qml:219: TypeError: Cannot assign to read-only property "cameras"` 
Changing it to `_dynamicCameras.currentCamera = index` fixes it

src/FlightMap/Widgets/CameraPageWidget.qml
```diff
@@ -202,6 +203,23 @@ Column {
...
+                            currentIndex:   0
```
lbegani: The selector always shows the first item in the list of the camera
Changing it to `currentIndex:   _dynamicCameras.currentCamera` fixes the issue

src/FlightMap/Widgets/CameraPageWidget.qml
dogmaphobic: Yup, I had the same error someplace else (from copying/pasting from the original code) and missed this one. Things you miss when you can't test :P

src/FlightMap/Widgets/CameraPageWidget.qml
dogmaphobic: Same idea.

Thanks!

lbegani:
Found two issues - 
1. Hitting record button sends image/video capture request to the first component instead of selected camera component
2. Capture request is sent only once. Even though image capture from px4-sitl gazebo camera was success, there was no request sent for second time in spite of hitting the record button

dogmaphobic:
@lbegani I finally got around to take a look at this. I've got the branch up-to-date with master and fixed the issue of sending all commands to the first camera only. Your second issue is probably a lack of a `MAVLINK_MSG_ID_CAMERA_CAPTURE_STATUS` message after an image capture? QGC will consider the camera "busy" until it gets a `MAVLINK_MSG_ID_CAMERA_CAPTURE_STATUS` message indicating the image was indeed captured.

lbegani:
Thanks @dogmaphobic . The second issue is fixed in https://github.com/Dronecode/camera-manager.

PR #6190, opened 2018-03-11
PlanToolBar: Enable focus of upload button when clicked

### LINKED ISSUE #6150: Commit all UI field on mission upload
@DonLakeFlyer You can get into a quite unexpected state when uploading a mission:

  - Add all waypoints
  - Edit one, enter a new altitude but leave the cursor in the field and do NOT press enter
  - Hit upload mission

The mission will upload with the old altitude. Can we make sure to commit all fields on hitting the mission upload button?

We found out the hard way by crashing a drone - it came in too low.

Williangalvani:
This causes the signal editingFinished() to be emitted and the current value of
the textField to be updated and validated before uploading.

fix #6150

Signed-off-by: Willian Galvani <williangalvani@gmail.com>

### DIFF
--- a/src/PlanView/PlanToolBar.qml
+++ b/src/PlanView/PlanToolBar.qml
@@ -301,7 +301,10 @@ Rectangle {
         enabled:                !_controllerSyncInProgress
         visible:                !_controllerOffline && !_controllerSyncInProgress && !uploadCompleteText.visible
         primary:                _controllerDirty
-        onClicked:              planMasterController.upload()
+        onClicked: {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
This is not the correct fix. This fix is to set ```activeFocusOnPress: true``` in QGCButton and QGCCheckBox which in turn fixes the problem everywhere. QGCComboBox may be missing something as well.

DonLakeFlyer:
I'll get around to it some time today.

PR #6192, opened 2018-03-12
Firmware Upgrade Fixes

dogmaphobic:
* Custom, single firmware upgrades were fixed to "stable", even if the user selected "custom".
* The code, when selecting the firmware type was using the combo box model index instead of the actual type defined by the list model. This would only work for when updating the standard flight stack. If you were trying to flash a custom firmware for "SingleFirmwareMode" (custom builds) or for PX4 Flow, the _index_ would point to _Beta_ instead.

@DonLakeFlyer, unrelated to this fix, while looking, I noticed this error:

`qrc:qml/FirmwareUpgrade.qml:127: ReferenceError: hideDialog is not defined`

It comes from here: https://github.com/mavlink/qgroundcontrol/blob/firmwareUpgrade/src/VehicleSetup/FirmwareUpgrade.qml#L127

I don't quite understand how these dialogs work and what this is trying to "hide".

### DIFF
--- a/src/VehicleSetup/FirmwareUpgrade.qml
+++ b/src/VehicleSetup/FirmwareUpgrade.qml
@@ -168,7 +168,7 @@ SetupPage {
                     function accept() {
                         hideDialog()
                         if (_singleFirmwareMode) {
-                            controller.flashSingleFirmwareMode()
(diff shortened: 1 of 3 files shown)


dogmaphobic:
Looks like Appveyor changed Qt to something else...

PR #6205, opened 2018-03-13
Update gstreamer Linux install instructions

DonLakeFlyer:
Changes in latest Ubuntu 16.04 were causing them to fail.

@barzanisar When you created #6166 was there a reason you didn't update the instructions to match what you found to work in #6165?

@dogmaphobic Can you take a look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ You **need to install Qt as described below** instead of using pre-built package
     * Windows: Make sure to install VS 2015 32 bit package.
 
 ###### Install additional packages:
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Bueller, anyone...

barzanisar:
@DonLakeFlyer the packages I added in https://github.com/mavlink/qgroundcontrol/pull/6166 includes the package mentioned in https://github.com/mavlink/qgroundcontrol/issues/6165

Only after installing these packages, I could build QGC from Qt creator.

Also, installing gstreamer1.0* still gives the error:

**The following packages have unmet dependencies:
 gstreamer1.0-hybris:i386 : Depends: libmedia1:i386 but it is not going to be installed
E: Unable to correct problems, you have held broken packages.**

But QGC builds so I just ignored this error.

DonLakeFlyer:
> But QGC builds so I just ignored this error.

Ok. That makes sense now. My version is a combination of the two which fixes the problem and avoids the error.

PR #6206, opened 2018-03-13
Add support for VTOL transition from joystick buttons

DonLakeFlyer:
Also fixed shutdown crash related to joystick


JonReacher:
@DonLakeFlyer This isn't working in 3.3.2. Great feature, btw, very useful.

DonLakeFlyer:
It's only available in daily builds.

PR #6209, opened 2018-03-13
Convex survey support

DonLakeFlyer:
* Fixed dog slow loading of complex KML
* Complicated concave KML loads for generating survey grids was causing bad transect generation
* Change survey grid generation such that the polygon is no longer converted to convex
* New transect intersection algorithm which works with convex polygons

Examples:
screen shot 2018-03-13 at 3 39 14 pm

This .mov shows how it all work while dynamically changing the grid angle of a survey generated from a complex concave KML load:
https://s3-us-west-2.amazonaws.com/qgroundcontrol/DonLakeFlyer/ConvexGrid.mov

### DIFF
--- a/src/MissionManager/QGCMapPolygon.cc
+++ b/src/MissionManager/QGCMapPolygon.cc
@@ -258,6 +258,19 @@ void QGCMapPolygon::appendVertex(const QGeoCoordinate& coordinate)
     emit pathChanged();
 }
 
+void QGCMapPolygon::appendVertices(const QList<QGeoCoordinate>& coordinates)
+{
+    QList<QObject*> objects;
+
+    foreach (const QGeoCoordinate& coordinate, coordinates) {
+        objects.append(new QGCQGeoCoordinate(coordinate, this));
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
@antiheavy I'm going to need you help testing this rather large bug fix! I've hammered on it pretty well, but an extra set of eyes will help. It's actual quite a simple code change, but right in the middle of grid generation so important to not screw up.

DonLakeFlyer:
This came about from having a hard time fixing the code which converted from concave to convex polygons. It was buggy and failed in various cases. While thinking about that, I stumbled across an idea for a new way to interect transects against convex polygons. Which turned out to be simple and it even seems to work!

Antiheavy:
The convex problem is much trickier than it seems at first!  Happy to help test.

Antiheavy:
Ping me once it is in Master, or send me a build.

Antiheavy:
been playing around with this for a few days and it seems to work pretty well.
image

image

image

image

DonLakeFlyer:
Thanks a ton for hammering on this Todd.

Issue #6204, opened 2018-03-13
Mission Terrain Support without firmware terrain following


DonLakeFlyer:
This is to start the discussion on supporting altitude above terrain in missions. This specific discussion assumes that the firmware does **not** have support for MAV_FRAME_GLOBAL_TERRAIN which would require some sort of terrain tile push to the vehicle. The assumption is that this is solely done on the ground station by setting appropriate mission item heights and/or possibly adding new wayoints to handle a terrain problem in between two added waypoints.

DonLakeFlyer:
@birchera @philipoe @AndKe @antiheavy @sanderux FYI. Does this statement above make sense as a problem statement?

philipoe:
Yes. As a bare minimum, the user should be warned to indicate if at any position along a simplified path between the waypoints (e.g. straight line path between two position WPs) the vehicle is close to colliding with terrain. 

Visualizing the terrain in a nicer way such that the user can see exactly where a terrain collision might occur is of course harder than just outputting a warning. Possible thoughts: 1) visualizing the terrain along the path in "google earth style" in the terrain altitude widget (although this is probably too much information in such a small window) or 2) changing the path's color in the map view as a function of distance to terrain (e.g. just marking it in red if a collision could happen)?

AndKe:
One scary scenario is that aircraft may fail to fetch elevation data during mission.  , be it due to GCS link, or internet access.

Planning a mission that does not use MAV_FRAME_GLOBAL_TERRAIN may be useful to closely verify/see the result, and making sure the aircraft will have all necessary data.

On the other hand, if we can be sure the aircraft have the data needed, and/or verify *any* mission's elevation would be an even greater tool (as moving an mission item would instantly account for the elevation if the aircraft were using such navigation mode)
 
AP2 has the "mission elevation" widget, in complex terrain, it's very useful.

and should be visualized right in the:
![screenshot from 2018-03-13 16-35-54](https://user-images.githubusercontent.com/1468842/37352059-b0508532-26dc-11e8-8e7e-da6f44561479.png)

Antiheavy:
I posted a "feature request" issue about this a while back.  I'll add this to the discussion:  https://github.com/mavlink/qgroundcontrol/issues/5524  


> 2) changing the path's color in the map view as a function of distance to terrain (e.g. just marking it in red if a collision could happen)?

I've used systems in the past that worked like this and found it a good method.  Simple.

[...]

PR #6225, opened 2018-03-15
Simple mission terrain support

DonLakeFlyer:
screen shot 2018-03-15 at 1 03 28 pm

The above example shows a plan which was done with all waypoints 20 meters above a varied terrain. You can see this as the consistent delta from the terrain bar for waypoints 2-4 at the bottom of the screen.

* Simple mission items can now specify an altitude to be above terrain
* Altitude mode is specified by a dropdown which has the options: Alt (rel), AMSL, Above Terrain
* If you select Above Terrain the altitude you specify is the height above terrain
* Creating/Updating plans which use Above Terrain altitude mode require an internet connection to communicate with the terrain server
* Once the plan is saved you no longer need an internet connection
* Below the Altitude field you will see the actual AMSL altitude for the item
* Until the terrain query returns Actual AMSL Alt will show --.--
* You will not be able to save/upload plans which use terrain data until all the values are updated based on terrain data from server
* The Altitude Mode, Altitude and Actual AMSL Alt values are persisted to saved Plan files
* If you load a plan back from the vehicle which uses Above Terrain Altitude Mode it will return back as an AMSL Altitude Mode item.
* Terrain is not currently supported by complex items
* This **does not** support the ArduPilot terrain following feature. It is just adjusting the actual heights of the mission items using terrain data.

Related to #6204

### DIFF
--- a/src/JsonHelper.cc
+++ b/src/JsonHelper.cc
@@ -97,7 +97,7 @@ bool JsonHelper::validateKeyTypes(const QJsonObject& jsonObject, const QStringLi
         QString valueKey = keys[i];
         if (jsonObject.contains(valueKey)) {
             const QJsonValue& jsonValue = jsonObject[valueKey];
-            if (types[i] == QJsonValue::Null && jsonValue.type() == QJsonValue::Double) {
+            if (jsonValue.type() == QJsonValue::Null &&  types[i] == QJsonValue::Double) {
                 // Null type signals a NaN on a double value
                 continue;
             }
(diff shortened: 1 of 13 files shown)


DonLakeFlyer:
As always this is a fairly large scale new feature. So extra caution with initial usage would be a good idea :)

terial:
Hey Don, any plans to support Ardupilot terrain following?

Kelly-Foster:
+1.

DonLakeFlyer:
Not at the moment.

PR #6226, opened 2018-03-16
Corridor Scan: Store/Load generated items from Plan file

DonLakeFlyer:
How complex items work now:
* Only the meta data is stored in the plan file. The actual mission items are not.
* This means that when a complex item is loaded from a plan file it must regenerate the actual mission items from the meta data.

This is a problem because:
* If the algorithm for item generation changes from release to release the complex item may generate a mision which is slightly different from when it was created.
* If the complex item contains terrain references then it would need internet access to re-generate the mission items. This would not work if you are trying to load a plan onto a vehicle at the field with no internet access.

The solution:
* Store the generated mission data in the plan file as well as the meta data.
* When you load a plan with a complex item, the complex items use the stored items instead of re-generating.
* If you dirty the complex item in any way, the stored mission items are tossed and it reverts back to generating the items onthe fly.
* The other benefit is that non-QGC ground stations can load plan files with complex items and use the stored items to populate a mission.

Currently this mechanism is only support by Corridor Scan. It will enventually flow to all complex items.


DonLakeFlyer:
Related to #6204

PR #6231, opened 2018-03-17
Corridor Scan initial terrain support

DonLakeFlyer:
screen shot 2018-03-17 at 9 12 17 am

* Terrain follow checkbox is live
* The various terrain follow settings Tolerance, ... are NYI
* Current implemenation only adjusts existing waypoints for terrain heights. It does not add additional waypoints in between yet.

The end result looks like this. A little hard to see, but you can see all points adjusted for 10m over terrain. If you look closely you can also see why just this is inadequate where the one transect goes through terrain:
screen shot 2018-03-17 at 9 12 06 am

Related to #6204

### DIFF
--- a/src/MissionManager/CorridorScanComplexItem.cc
+++ b/src/MissionManager/CorridorScanComplexItem.cc
@@ -380,9 +380,11 @@ void CorridorScanComplexItem::_rebuildTransectsPhase1(void)
                 // Extend the transect ends for turnaround
                 double azimuth = transect[0].azimuthTo(transect[1]);
                 extensionCoord = transect[0].atDistanceAndAzimuth(-_turnAroundDistanceFact.rawValue().toDouble(), azimuth);
+                extensionCoord.setAltitude(qQNaN());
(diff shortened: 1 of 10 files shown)


moreba1:
i can't save mission with terrain following in kml and see in google earth . this error appears : WAITING ON TERRAIN DATA FROM SERVER FOR CORRECT ALTITUDE values.

DonLakeFlyer:
Are you using absolute latest daily build? Older builds sometimes didn't download all terrain data. Also if the terrain data set is large it can take a minute of so to complete downloading.

moreba1:
Thank you. my problem was solved with google satellite map and new version. but in the menu of terrain altitude , one point is displayed for total corridor scan.

DonLakeFlyer:
> but in the menu of terrain altitude , one point is displayed for total corridor scan.

That is the way it currently works. If you want to verify individual items you need to read it back from the vehicle.

moreba1:
yes , this is a good idea .

PR #6233, opened 2018-03-17
Restructure for TerrainQueryInterface abstract base class

DonLakeFlyer:
This restructures the terrain query code such that all queries go through a TerrainQueryInterface abstract base class definition. This way the actual terrain query mechanism can be coded without the upper level client code changes affecting it. I have implemented the AirMap version. Then the offline code just needs to implement another version of this interface and change the code in the three places that instantiates TerrainAirMapQuery to the offline implementation. Should be fairly straightforward.

```
/// Base class for offline/online terrain queries
class TerrainQueryInterface : public QObject
{
    Q_OBJECT

public:
    TerrainQueryInterface(QObject* parent) : QObject(parent) { }

    /// Request terrain heights for specified coodinates.
    /// Signals: coordinateHeights when data is available
    virtual void requestCoordinateHeights(const QList<QGeoCoordinate>& coordinates) = 0;

    /// Requests terrain heights along the path specified by the two coordinates.
    /// Signals: pathHeights
    ///     @param coordinates to query
    virtual void requestPathHeights(const QGeoCoordinate& fromCoord, const QGeoCoordinate& toCoord) = 0;

    /// Request terrain heights for the rectangular area specified.
    /// Signals: carpetHeights when data is available
    ///     @param swCoord South-West bound of rectangular area to query
    ///     @param neCoord North-East bound of rectangular area to query
    ///     @param statsOnly true: Return only stats, no carpet data
    virtual void requestCarpetHeights(const QGeoCoordinate& swCoord, const QGeoCoordinate& neCoord, bool statsOnly) = 0;

signals:
    void coordinateHeights(bool success, QList<double> heights);
    void pathHeights(bool success, double latStep, double lonStep, const QList<double>& heights);
    void carpetHeights(bool success, double minHeight, double maxHeight, const QList<QList<double>>& carpet);
};
```

FYI: @birchera 
Related to #5842

### DIFF
--- a/src/MissionManager/TransectStyleComplexItem.cc
+++ b/src/MissionManager/TransectStyleComplexItem.cc
@@ -374,7 +374,7 @@ void TransectStyleComplexItem::_adjustTransectPointsForTerrain(void)
 
         bool surveyEdgeIndicator = transectPoint.altitude() == _surveyEdgeIndicator;
         if (_followTerrain) {
-            transectPoint.setAltitude(_transectsPathHeightInfo[i].rgHeight[0] + altitude);
+            transectPoint.setAltitude(_transectsPathHeightInfo[i].heights[0] + altitude);
         } else {
             transectPoint.setAltitude(altitude);
         }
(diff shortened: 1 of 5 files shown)


hamishwillee:
@DonLakeFlyer While this is fresh in your mind, perhaps a few words about the abstraction in https://dev.qgroundcontrol.com/en/ ? That way at least other companies can know that creating another interface is possible.

DonLakeFlyer:
This is just a temporary thing to make life easier for Andreas. Not really meant to create a supported pluggable terrain query system.

PR #6235, opened 2018-03-18
Corridor Scan: Tolerance based terrain follow

DonLakeFlyer:
* Tolerence based terrain follow generation is working now. 
* Next up: Terrain waypoints need to be adjust for max climb/descent rates
* Screenshots from extreme example of 10m above terrain with 5m tolerance.

Corridor Scan:
screen shot 2018-03-18 at 2 54 26 pm

Loaded back from vehicle so you can see all the generated points:
screen shot 2018-03-18 at 2 54 54 pm

KML:
screen shot 2018-03-18 at 2 57 57 pm

Related to #6204


sanderux:
Looks great Don! could i suggest having a default max climb/descend rate?

DonLakeFlyer:
> could i suggest having a default max climb/descend rate?

The default is 0, which indicates do not adjust for max climb/descent. If you set them to other values they will be remembered.

PR #6243, opened 2018-03-21
FlightDisplay : Remove sentence about disabling arming check

khancyr:
We don't want average users to disable the arming check. If the problem is trigger, it is generally because there is a problem so the check shouldn't be advice to be disable.


DonLakeFlyer:
Seems reasonable to me

PR #6270, opened 2018-03-27
PX4 SensorsSetup Airspeed check FW_ARSP_MODE no longer bool

dagar:
We changed the metadata of FW_ARSP_MODE so it's no longer a boolean.

https://github.com/PX4/Firmware/issues/9120


DonLakeFlyer:
Fixed here: https://github.com/mavlink/qgroundcontrol/pull/6272

dagar:
Thanks

PR #6272, opened 2018-03-27
Allow type coercion for FW_ARSP_MODE param

DonLakeFlyer:
Previous code was broken due to no type coercion from int to bool. Without it, possible breaks due to meta data changes.

### DIFF
--- a/src/AutoPilotPlugins/PX4/SensorsComponentSummaryFixedWing.qml
+++ b/src/AutoPilotPlugins/PX4/SensorsComponentSummaryFixedWing.qml
@@ -25,7 +25,7 @@ FactPanel {
     property Fact airspeedDisabledFact: controller.getParameterFact(-1, "FW_ARSP_MODE")
     property Fact airspeedBreakerFact:  controller.getParameterFact(-1, "CBRK_AIRSPD_CHK")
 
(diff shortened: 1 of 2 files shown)


src/AutoPilotPlugins/PX4/SensorsComponentSummaryFixedWing.qml
```diff
@@ -25,7 +25,7 @@ FactPanel {
     property Fact airspeedDisabledFact: controller.getParameterFact(-1, "FW_ARSP_MODE")
     property Fact airspeedBreakerFact:  controller.getParameterFact(-1, "CBRK_AIRSPD_CHK")
 
-    property bool _airspeedVisible:     airspeedDisabledFact.value === false && airspeedBreakerFact.value !== 162128
+    property bool _airspeedVisible:     airspeedDisabledFact.value == 0 && airspeedBreakerFact.value !== 162128
```
dagar: airspeedDisabledFact -> airspeedModeFact?

DonLakeFlyer: Better, but not super critical

PR #6285, opened 2018-03-30
rally point upload and download functionality

arwagoner:
Adds the ability to upload rally points to and download from the vehicle.


DonLakeFlyer:
Ah awesome, thanks. Never quite completed. Did you check this against GeoFenceManager? It should just follow the same pattern/code that is in there.

arwagoner:
Yes, I tried to follow as closely to GeoFenceManager as possible.

DonLakeFlyer:
Did you have anything to test this against? MockLink supports Rally Points. But is the PX4 Firmware pull for Rally Points in master now?

arwagoner:
I tested it with this branch of PX4 firmware https://github.com/PX4/Firmware/pull/7743 It isn't in master yet.

DonLakeFlyer:
Great, thanks. Merging

Issue #6264, opened 2018-03-25
QGC 3.3 Joystick: Joystick/ No RC Checks Manual Control lost.


tubeme:
PX4 1.7.3
QGC 3.3
Joystick: Mad Catz F.L.Y.5
Flying Wing. (VTOL)

When Joystick/ No RC Checks is used on arming it reports manual control lost and commands RTL which is rejected with a message Rejected RTL in a loop all the time until i press disarm.

![173 manual control lost on joystick no rc](https://user-images.githubusercontent.com/12617619/37880474-b629b74e-3091-11e8-9b97-43fc054d9ca5.png)

DonLakeFlyer:
I believe this is caused by the fact that when you initiall boot QGC joystick commands don't automatically start sending. You have to go to the Joystick page to get them to start, even if joystick is enabled.

DonLakeFlyer:
@jaxxzer @nanthony21 Joystick folks. Could someone help out and fix this?

If you turn on JoystickValuesLog can you see that they don't start pumping out until you go to the joystick page. I believe this is the case with 3.3 as well so the fix needs to start on that branch.

nanthony21:
@DonLakeFlyer I dug into it a bit today. It looks like this was introduced by my pull request #6117. I'll fix it this weekend. It looks like the "Stable v3.3" branch is unaffected.

DonLakeFlyer:
This should be fixed now

PR #6288, opened 2018-03-31
Fix joystick activation

nanthony21:
This fixes issue #6264 which was caused in PR #6117.

This pull request removes the `outputEnabled` member of the `joystick` class as it serves a reduntant function to the `joystickEnabled` member of the `vehicle` class.

It also changes the `vehicle` class to separate the enabling/disabling of the joystick from the starting/stopping of the thread that polls the joystick. This way we can still monitor the joystick even when it is disabled.

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -445,7 +445,7 @@ void Joystick::run(void)
             }
         }
 
-        if (_outputEnabled && _calibrated) {
+        if (_activeVehicle->joystickEnabled() && _calibrated) {
             int     axis = _rgFunctionAxis[rollFunction];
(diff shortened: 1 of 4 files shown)


nanthony21:
Side note: the Joystick class has a member referencing the activeVehicle and the Vehicle class has a member referencing the activeJoystick.
This makes for a circular reference which can cause a number of issues.
Would it be a good idea to rearrange things so that Joystick doesn't need to know anything about the active vehicle?

DonLakeFlyer:
> the Joystick class has a member referencing the activeVehicle and the Vehicle class has a member referencing the activeJoystick.

Looking at the code I don't see how that can be changed.

Issue #2878, opened 2016-02-24
Add controller tuning widget


LorenzMeier:
@DonLakeFlyer @kd0aij @dagar This would have the parameters on the left, with a drop-down selecting between different controllers (rate, attitude, position, etc) and a realtime graph on the right showing desired vs. actual.

It might also have for the tuning parameters drop-down to map to a RC channel. With the UI setting that in realtime would be feasible for RC tuning, so I would argue that this is not blocked on Firmware support. Probably APM has similar needs and the instrument could be shared and jointly maintained.

dagar:
Here's a rough mockup of what we were thinking.

![qgc_parameter_tuning_and_plotting](https://cloud.githubusercontent.com/assets/84712/13296507/dc8d8ebc-dafb-11e5-87fe-fbe3ba921bec.png)

If someone could help establish the initial framework (controller selection screen, realtime plotting, etc) I could help with the specifics (setup plots and params for each) and testing of course.

@mujahid-abdulrahim-p FYI

DonLakeFlyer:
My PixRacer shows up today. If I end up having to tune it myself you may end up seeing this sooner than later :)

kd0aij:
Does Qt have no "spinner" widgets for incrementing/decrementing values?

DonLakeFlyer:
It does now with the Qt 5.5 we switched to a few months back.

kd0aij:
It would be nice to be able to spec a floating point increment/decrement
for gain fields (in the absence of an RC function to do that). Currently, I
have to type in the entire contents of the field on my tablet, when all I
want to do is add or subtract a little. But I doubt it's practical to press
a tiny up/down arrow symbol with frozen fingers in bright sunlight.

DonLakeFlyer:
Right spin buttons don't work on tablet. Need a more custom spin button style implementaion like tower.

tubeme:
I copy this from another issue #2328 I just stumbled there on the topic:

The current implementation of "Set to RC" is very basic and sometimes tedious to use.

First once I assign knobs to params, and then I reassign the knob to other param and vice versa and I loose track of what the assignment is, I cannot see what is the current assignment. So I have to reset all the assignments in order to be sure I reassign them the correct way.

[...]

PR #6295, opened 2018-04-02
PX4 PID tuning support

DonLakeFlyer:
screen shot 2018-04-01 at 5 35 55 pm

This is a pretty major working in progress from a user model standpoint. Mostly trying thing out at this point. Here are the possibly non-obvious things:
* You get to it from Tuning and then check Advanced
* When the page loads it saves the parameter values for the current tuning set internally. The idea is as you futz with them you have an "oh shit" button to get you back to what are hopefully good values by clicking the Reset button.
* As you adjust and get better values you can save them as new reset values by clicking the Save Values button.
* The +/- buttons increment the values by the amount specified in the Increment/Decrement % combo box. This way you can start by futzing with the 15% changes and then as you fine tune you can drop to 10% or 5% increments.
* You can start/stop and clear the graphs using the Chart buttons.

The above currently is only supported for PX4 Pro firmware and multi-rotors. But that said the whole thing is implemented in a generic fashion such that it just works off of generic lists of parameters. So all I need to do is figure out what all the other param sets are to implemenent other vehicle types and ArduPilot support.

In order to implement all of this I added a bunch of generic support as well:
* SetupPages have a new Advanced checkbox which you can turn on and use
* Added rollRate, pitchRate, yawRow to Vehicle object main FactGroup
* Added new setpoint FactGroup to vehicle with the values: roll, pitch, yaw, rollRate, pitchRate, yawRate
* All new FactGroup values are available for display in the Instrument Panel

Related to #2878

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -224,6 +224,7 @@
         <file alias="Vehicle/ClockFact.json">src/Vehicle/ClockFact.json</file>
         <file alias="Vehicle/GPSFact.json">src/Vehicle/GPSFact.json</file>
         <file alias="Vehicle/GPSRTKFact.json">src/Vehicle/GPSRTKFact.json</file>
+		<file alias="Vehicle/SetpointFact.json">src/Vehicle/SetpointFact.json</file>
         <file alias="Vehicle/SubmarineFact.json">src/Vehicle/SubmarineFact.json</file>
         <file alias="Vehicle/TemperatureFact.json">src/Vehicle/TemperatureFact.json</file>
         <file alias="Vehicle/VehicleFact.json">src/Vehicle/VehicleFact.json</file>
--- a/src/AutoPilotPlugins/Common/SetupPage.qml
+++ b/src/AutoPilotPlugins/Common/SetupPage.qml
@@ -10,6 +10,7 @@
 import QtQuick          2.3
 import QtQuick.Controls 1.2
 import QtQuick.Dialogs  1.2
+import QtQuick.Layouts  1.2
 
 import QGroundControl               1.0
 import QGroundControl.FactSystem    1.0
@@ -30,6 +31,8 @@ QGCView {
     property string pageDescription:    vehicleComponent ? vehicleComponent.description : ""
     property real   availableWidth:     width - pageLoader.x
     property real   availableHeight:    height - pageLoader.y
+    property bool   showAdvanced:       false
+    property alias  advanced:           advancedCheckBox.checked
 
     property var    _activeVehicle:         QGroundControl.multiVehicleManager.activeVehicle
     property bool   _vehicleArmed:          _activeVehicle ? _activeVehicle.armed : false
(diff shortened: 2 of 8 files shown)


bkueng:
Nice!
@MaEtUgR I think you're interested in this

MaEtUgR:
So true, this will make tuning so much more convenient!

I have the suggestion to allow tuning roll and pitch at the same time because in practise that's what you need most of the time even if your vehicle is not perfectly symmetric. It's
1. better to let the mixer take care of asymmetric actuator placement geometry and
2. even if there is asymmetric weight distribution you typically first tune to a certain value and then fine tune asymmetry from there.

I assume the increments are 5% of what was the old value. e.g. if you start out with the default `ROLLRATE_D = 0.003` and press the plus button three times you get `0.003 * (1.05)^3 = 0.003472875`. Are totally uneven numbers no concern?

hamishwillee:
I presume that this will change the way we do tuning, and hence some of the docs like https://docs.px4.io/en/advanced_config/pid_tuning_guide_multicopter.html ?

If so, what should be the trigger for updates (now, next QGC release?) and who should update the process/docs?

DonLakeFlyer:
FYI: Daily builds are still spitting the new tuning page out correctly: https://github.com/mavlink/qgroundcontrol/issues/6296. Until then you have to build yourself to get it.

> I have the suggestion to allow tuning roll and pitch at the same time because in practise that's what you need most of the time even if your vehicle is not perfectly symmetric.

Not sure how you tune two axes at the same time? You move the vehicle in both roll and pitch at the same time. Or are you talking about copying the same values if you tune roll to pitch.

> I assume the increments are 5% of what was the old value. e.g. if you start out with the default ROLLRATE_D = 0.003 and press the plus button three times you get 0.003 * (1.05)^3 = 0.003472875. Are totally uneven numbers no concern?

Correct. Something like that although they are rounded to the decimal places specified in the parameter meta data. Don't see why numbers need to be even.

MaEtUgR:
@DonLakeFlyer 
> Not sure how you tune two axes at the same time? You move the vehicle in both roll and pitch at the same time. Or are you talking about copying the same values if you tune roll to pitch.

Yes, the user should heve the option to lock them to the same value first because roll and pitch are most of the time not completely different. Subsequent fine tuning is then still possible.

> Correct. Something like that although they are rounded to the decimal places specified in the parameter meta data. Don't see why numbers need to be even.

Technically you are completely right. The only reason I'd prefer rounded values is human readability/siplicity to copy the value over. No strong feelings there...

DonLakeFlyer:
> I presume that this will change the way we do tuning, and hence some of the docs like

What I would look to see if instructions for how to tune while flying. The concept of holding a pile of swinging Ginsu knives in my hand just seems crazy as the only way to do it. Not something I ever plan to do. For my usage pattern, I choose and airframe which gives me a flyable vehicle. Then I just want to make it better from there. I'm not starting from scratch.

PR #6305, opened 2018-04-04
Make the Message Indicator respect palette settings

DonLakeFlyer:


DonLakeFlyer:
@dogmaphobic The one change in here is that both error and warning text use the warning palette color. There is text which indicates the difference so I think this is ok. Didn't want to deal with adding yet another palette color. Also I don't think I've ever seen any "warning" text from firmware.

dogmaphobic:
We can tweak these down the road. The entire UI needs some re-planning. Not some fundamental changes but rather, something to make it more consistent throughout.

PR #6306, opened 2018-04-05
PID Tuning: Make QtCharts usage desktop only

DonLakeFlyer:
Qt Charts doesn't work with our mobile code at the moment

### DIFF
--- a/QGCSetup.pri
+++ b/QGCSetup.pri
@@ -102,6 +102,7 @@ LinuxBuild {
 
     # QT_INSTALL_LIBS
     QT_LIB_LIST = \
+        libQt5Charts.so.5 \
         libQt5Core.so.5 \
         libQt5DBus.so.5 \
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
@dogmaphobic The jenkins servers need to have Qt Charts installed for Linux and OSX builds now.

dogmaphobic:
I can handle my macOS Jenkins slave. @dagar the Linux slave will need that too.

dogmaphobic:
By the way, did you look through the https://github.com/mavlink/qgroundcontrol/tree/switchToQQuickView/src branch, which basically had the same objective?

DonLakeFlyer:
> did you look through the...

I didn't. But I basically did the same thing for testing. Then decided to leave it out for now. I remember mobile being crazy flaky till we switched to full Qml. So, paranoid about taking a eve half step back towards that.

PR #6308, opened 2018-04-05
Set video capture status for missions

dogmaphobic:
When setting up a mission using video capture (as opposed to image capture), tell the autopilot to send capture updates (every 5 seconds). This will tell the UI to update the current status as the video "recording" started by an autopilot command instead of an user command.

### DIFF
--- a/src/MissionManager/CameraSectionTest.cc
+++ b/src/MissionManager/CameraSectionTest.cc
@@ -70,7 +70,7 @@ void CameraSectionTest::init(void)
                                                              MAV_CMD_VIDEO_START_CAPTURE,
                                                              MAV_FRAME_MISSION,
                                                              0,                             // Reserved (Set to 0)
-                                                             0,                             // No CAMERA_CAPTURE_STATUS streaming
+                                                             VIDEO_CAPTURE_STATUS_INTERVAL, // CAMERA_CAPTURE_STATUS (default to every 5 seconds)
                                                              NAN, NAN, NAN, NAN, NAN,       // param 3-7 reserved
                                                              true,                          // autocontinue
                                                              false),                        // isCurrentItem
@@ -896,13 +896,6 @@ void CameraSectionTest::_testScanForStartVideoSection(void)
     QCOMPARE(_cameraSection->settingsSpecified(), false);
     visualItems.clear();
 
-    invalidSimpleItem.missionItem() = _validStartVideoItem->missionItem();
-    invalidSimpleItem.missionItem().setParam2(10);    // must be 0
-    visualItems.append(&invalidSimpleItem);
-    QCOMPARE(_cameraSection->scanForSection(&visualItems, scanIndex), false);
-    QCOMPARE(visualItems.count(), 1);
-    QCOMPARE(_cameraSection->settingsSpecified(), false);
-    visualItems.clear();
 }
 
 void CameraSectionTest::_testScanForStopVideoSection(void)
--- a/src/MissionManager/CameraSectionTest.cc
+++ b/src/MissionManager/CameraSectionTest.cc
@@ -931,6 +931,14 @@ void CameraSectionTest::_testScanForStopVideoSection(void)
     QCOMPARE(visualItems.count(), 1);
     QCOMPARE(_cameraSection->settingsSpecified(), false);
     visualItems.clear();
+
+    invalidSimpleItem.missionItem() = _validStartVideoItem->missionItem();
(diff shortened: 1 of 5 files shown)


src/MissionManager/CameraSectionTest.cc
```diff
@@ -896,13 +896,6 @@ void CameraSectionTest::_testScanForStartVideoSection(void)
     QCOMPARE(_cameraSection->settingsSpecified(), false);
     visualItems.clear();
 
-    invalidSimpleItem.missionItem() = _validStartVideoItem->missionItem();
```
DonLakeFlyer: This test is still valid. But should be changed to something like setParam2(VIDEO_CAPTURE_STATUS_INTERVAL + 1) to make sure it fails given new setting.

DonLakeFlyer:
Problems:
* This will break loading old missions which use video capture back from the vehicle. I think that's probably ok.
* If there is new support in VIDEO_CAPTURE the command json needs to be updated such that the editor for it supports it.

That said, wouldn't it be better if VIDEO_CAPTURE just streamed updates by default? That is how camera trigger works. You don't need to tell it to send CAMERA_TRIGGER messages back. It just does it.

dogmaphobic:
> If there is new support in VIDEO_CAPTURE the command json needs to be updated such that the editor for it supports it.

I wasn't even aware that Json file existed. Now that I see it, that's where the default should come from. The value set within the command should also come from what the user enters. I have no idea where that's stored but I can dig through.

With that said, is there a way to flag some parameters as "no control" or something so it is not shown? Like the "reserved" below:

screen shot 2018-04-05 at 2 33 18 pm

> That said, wouldn't it be better if VIDEO_CAPTURE just streamed updates by default?

Unlike image capture status, this is indeed optional and it really only applies to missions. When the user starts recording videos from within the UI, the handling of the status is different and controlled by the UI itself. For missions, the UI has no way to know the camera started recording as the autopilot is what is starting/stopping it. That's why we need to set this up and that's how the API is defined.

DonLakeFlyer:
> That's why we need to set this up and that's how the API is defined.

Ah, sounds fine then.

DonLakeFlyer:
> Now that I see it, that's where the default should come from.

Those defaults only apply to using the editors. When you are futzing with things like CameraSection you have to do things all on your own.

> With that said, is there a way to flag some parameters as "no control" or something so it is not shown? Like the "reserved" below:

Not sure what you mean. My guess is that the json specified for VIDEO_CAPTURE is wrong and just need to be update. You only show values you want.

They way it works is that there is a hierachy of command ui:
* Generic->Firmware->Vehicle Type

From MissionManager/MavCmdInfoCommon.json:
```
        {
            "id":           2500,
            "rawName":      "MAV_CMD_VIDEO_START_CAPTURE",
            "friendlyName": "Start video capture",
            "description":  "Start video capture.",
            "category":     "Camera",
            "param1": {
                "label":            "Reserved",
                "default":          0,
                "decimalPlaces":    0
            }
        },
```

This should match the generic mavlink spec. It's wrong. There is no reason for Reserved to be there.

If PX4 supports all the params specified by mavlink then you are doing if you update this. If PX4 doesn't support some of the params or uses them differently than the spec specifies then you also need to add overrides to the upper levels of the hierarchy.

src/MissionManager/CameraSectionTest.cc
DonLakeFlyer: This test is for reading back a camera section in from the vehicle such that it goes back into the camera section of a waypoint instead of a separate VIDEO_START_CAPTURE mission item. When the user is saying "Start Video" from the camera section combo they have no option to change the interval. It is fixed by the code.

src/MissionManager/MavCmdInfoCommon.json
```diff
@@ -1041,23 +1031,18 @@
...
             "description":  "Start video capture.",
             "category":     "Camera",
-            "param1": {
-                "label":            "Reserved",
-                "default":          0,
-                "decimalPlaces":    0
+            "param2": {
+                "label":            "Status Interval",
+                "default":          0.2,
+                "decimalPlaces":    2
```
DonLakeFlyer: units = seconds?

DonLakeFlyer:
Looks fine except for units of seconds

src/MissionManager/MavCmdInfoCommon.json
dogmaphobic: Hz. Already added it.

dogmaphobic:
Merging it. The Jenkins failure is that its Linux build is missing Qt Charts.

Issue #6169, opened 2018-03-01
Add "radar" view using DISTANCE_SENSOR messages from autopilot


rmackay9:
ArduPilot supports object avoidance using various rangefinders (ZED 3D camera, 360 lidar, single lidar).  As part of this feature, the AP_Proximity library sends DISTANCE_SENSOR messages with the appropriate distance and orientation to the ground station.

This request is to add a "radar" like view so the operator can see the distances to nearby objects.  If the view can handle 8 horizontal directions that would be great.  adding up and down somehow would also be really nice.

DonLakeFlyer:
@dogmaphobic Did you already add something like this for the Typhoon?

dogmaphobic:
Yes, I did something quite simple using the `MAVLINK_MSG_ID_DISTANCE_SENSOR` message.

DonLakeFlyer:
Can that be brought upstream and support all the horizontal values?

dogmaphobic:
Not exactly and as it is, it's not that useful as that was made for a particular type of sensor working in a predictable way. QGC would need something more abstract. There are messages giving sensor values but nothing to enable/disable it for instance. If you have something but it's disabled, you will get the message but nothing ever changes, which is not too useful. You need to know it's active (and a way to make it active/inactive). Martina has also been working on some multidimensional sensors (she has done a boatload of work on that) and there are new messages for those. We need a plan so I know what to implement and how.

For example:
https://github.com/PX4/Firmware/pull/8896

DonLakeFlyer:
> If you have something but it's disabled, you will get the message but nothing ever changes, which is not too useful.

That would just seem to be a broken/studid firmware implemention? No?

DonLakeFlyer:
@rmackay9 Is there a way to simulate this with ArduPilot SITL for testing/implemenation?

DonLakeFlyer:
Hints on testing with SITL:
I need to check but normaly if you activate the sonar on sitl, you should have it on distance sensor msg. With mavproxy : load Tools/autotest/default_params/copter_rangefinder.parm or something like that. Then relaunch Sitl and you should have sonar and the distance sensor message with orientation pitch 270

DonLakeFlyer:
Hmm, I missed the all level and up/down sensors. I'll add to the pull.

DonLakeFlyer:
Ok, so the values are all plumbed through and can be displayed. I'll add a better display in the next release.

[...]

PR #6312, opened 2018-04-05
Add DISTANCE_SENSOR support to Vehicle FactGroup [WIP]

DonLakeFlyer:
Provides values display for level orientation (not all) distance sensors in Values instrument page.

Related to #6169

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -222,6 +222,7 @@
         <file alias="USBBoardInfo.json">src/comm/USBBoardInfo.json</file>
         <file alias="Vehicle/BatteryFact.json">src/Vehicle/BatteryFact.json</file>
         <file alias="Vehicle/ClockFact.json">src/Vehicle/ClockFact.json</file>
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
screen shot 2018-04-05 at 4 08 46 pm

So I need to figure out what to call these rotations. For first pass I want to plumb all level orientations plus up/down. Compass directions make some sense but they would be relative to the vehicle frame. The whole RotationYaw90 thing makes my skin crawl a bit. Want something more user friendly than that.

And yes, a radar like display is coming...

DonLakeFlyer:
Also figuring out how to handle multiple sensors...

DonLakeFlyer:
@maraster FYI

DonLakeFlyer:
Ok, I added all level orientations and up/down:
screen shot 2018-04-05 at 5 28 13 pm

There is support for only a single sensor reporting. It remembers the id of the first DISTANCE_SENSOR message it gets. It then ignores data from any sensor with a different id.

DonLakeFlyer:
Note the naming I used above.

PR #6316, opened 2018-04-06
Support for low battery state (and remaining time)

dogmaphobic:
Adding support for the new, extended low battery state. It provides flags about the current state of the (intelligent) battery as well as remaining time (when it reaches a low battery state).

This is just the plumbing within Vehicle, providing the state and remaining time. This has yet to be added to the UI and presented to the user.

The possible states are:

- Battery is not in low state. Normal operation.
- Battery state is critical, return / abort immediately.
- Battery state is too low for ordinary abort sequence. Perform fastest possible emergency stop to avoid damage.
- Battery failed, damage unavoidable.
- Battery is diagnosed to be broken or an error occurred, usage is discouraged / prohibited.


DonLakeFlyer:
I need to put some work into automating FactGroup setup more. It's always just a ton of copy paste.

PR #6319, opened 2018-04-07
Added clickable link and ScrollView to CustomCommandWidget

Rogni:


DonLakeFlyer:
Clickable link is nice. But if you to make it scroll you should use QGCFlickable and make it only scroll in the vertical direction.

DonLakeFlyer:
This should also set flickableDirection

src/ViewWidgets/CustomCommandWidget.qml
```diff
@@ -87,15 +87,20 @@ QGCView {
...
-                textFormat:         Text.RichText
-                visible:            !loader.visible
+                contentHeight:      textOutput.height
+                flickableDirection: QGCFlickable.VerticalFlick
+                QGCLabel {
+                    id:                 textOutput
+                    width:              container.width
+                    wrapMode:           Text.WordWrap
+                    textFormat:         Text.RichText
+                    visible:            !loader.visible
```
DonLakeFlyer: The visible check should be on the parent flickable

PR #6327, opened 2018-04-08
Allow map tile import/export on mobile

DonLakeFlyer:


DonLakeFlyer:
@dogmaphobic Can you check this?

PR #6331, opened 2018-04-09
Guard against crash in terrain code

DonLakeFlyer:
Still not sure of the exact cause, but this will help debug.

### DIFF
--- a/src/TerrainTile.cc
+++ b/src/TerrainTile.cc
@@ -114,6 +114,10 @@ double TerrainTile::elevation(const QGeoCoordinate& coordinate) const
         // Get the index at resolution of 1 arc second
         int indexLat = _latToDataIndex(coordinate.latitude());
         int indexLon = _lonToDataIndex(coordinate.longitude());
+        if (indexLat == -1 || indexLon == -1) {
+            qCWarning(TerrainTileLog) << "Internal error indexLat:indexLon == -1" << indexLat << indexLon;
+            return -1.0;
+        }
         qCDebug(TerrainTileLog) << "indexLat:indexLon" << indexLat << indexLon << "elevation" << _data[indexLat][indexLon];
         return static_cast<double>(_data[indexLat][indexLon]);
     } else {


birchera:
@DonLakeFlyer how exactly did you experience the crash?

We also found one issue, which we haven't figured out yet. Seems like the allocation of the elevation array fails sometimes.

```
SIGABRT -6 0x0000000000002609
0  libc.so                        0xb4473868 tgkill
1  libc.so                        0xb4470fe3 pthread_kill
2  libc.so                        0xb4446bad raise
3  libc.so                        0xb4442321 (Missing)
4  libc.so                        0xb4440388 abort
5  libgnustl_shared.so            0xa9eb2f5b __gnu_cxx::__verbose_terminate_handler()
6  libgnustl_shared.so            0xa9eb18ed __cxxabiv1::__terminate(void (*)())
7  libgnustl_shared.so            0xa9eb1961 std::terminate()
8  libgnustl_shared.so            0xa9eb1bbd __cxa_throw
9  libgnustl_shared.so            0xa9eb2043 operator new(unsigned int)
10 libgnustl_shared.so            0xa9eb2097 operator new[](unsigned int)
11 libQGroundControl.so           0x9431dd04 TerrainTile::TerrainTile(QByteArray)
12 libQGroundControl.so           0x941a323c TerrainTileManager::_fetchedTile()
```

It is possible that this only happens in our customized code, but I thought it might help to share this.

DonLakeFlyer:
The crash is referencing into _datra. I can't reproduce. So not sure exactly whether the indices are wrong or _data is bad. Will know a little more when @antiheavy retests.

DonLakeFlyer:
Doesn't new throwing mean running out of memory?

Antiheavy:
@DonLakeFlyer I will try to test today's daily build tonight (traveling all day).

Antiheavy:
I was able to get today's Windows daily build (2018-04-10) to crash in the same way.  Using a different Pixracer and window's laptop this time.  Connected to a Pixracer over USB cable , Play around with editing and uploading missions, eventually the application bombs.  Easily repeatable.

This is the version of QGC I had:
image

Link to download the QGCCrashDump log:
https://we.tl/ptI2zKpisW

DonLakeFlyer:
Ah, crap. I was hoping to not crash and get log output. I need to put more effort into crash debugging to help out here.

birchera:
> Doesn't new throwing mean running out of memory?

Yes, I also found why this was the case. Some devices had the old format (json) in the database. This gave a lot of inf tiles, which were stored but since they showed a different id were reloaded.
Since no one else should have used the intermediate version this should not be a problem.

PR #6334, opened 2018-04-10
New FactValueSlider control

DonLakeFlyer:
Currently only in use on the px4 tuning page. The old ui just had standard sliders here. If you actually tried to use them on mobile outside they were pretty much useless. Your finger is right over the slider so you can't really see if you are actually sliding it, especially for small incrememts. You also had no idea how much you were sliding since there were no values. So trying to tune hover throttle say was a royal PITA.

screen shot 2018-04-10 at 10 12 16 am

The new UI shows values. You can click on the values to left/right to adjust. You can flick the control left/right to get to more values.

This control is the starting point for replacing all value input ui in QGC.

Note: This image shows strange values for units of these properties. @dagar We had talked about cleaning up the tuning property meta data. Could you look at these as well for all vehicle types?


DonLakeFlyer:
The % ones are fine.

PR #6338, opened 2018-04-11
travis: Send AppImage and dmg to transfer.sh

patrickelectric:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -200,7 +200,12 @@ before_deploy:
   - cd ${TRAVIS_BUILD_DIR}
   - if [[ "${SPEC}" = "linux-g++-64" && "${CONFIG}" = "installer" ]]; then
         ./deploy/create_linux_appimage.sh ${TRAVIS_BUILD_DIR} ${SHADOW_BUILD_DIR}/release ${SHADOW_BUILD_DIR}/release/package;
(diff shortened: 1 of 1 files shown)


patrickelectric:
This will allow people to test and run PR and branchs that are not deployed.

DonLakeFlyer:
Isn't the real problem here that AppImage creation is part of travis script instead of part of QGC build makefiles? That could be changed. Not sure why things need to be posted to someplace else?

patrickelectric:
It would be nice to have any way to access appimage and dmg deployment files of PR and branchs, to allow anyone to test if their work is working correct.
transfer.sh is pretty easy to use, as you can see, the travis output will print the link for download.
Moving such things outside of travis script will help, but will not use the convenience/facility of travis CI.
E.g: In my case I do not have a windows and mac pc with all necessary tools to perform my work, all development is done in linux and I do my tests with travis/appveyor outputs in the others OSs.

dagar:
Yes, although it's only branches and tags from the main repo.

image

https://s3-us-west-2.amazonaws.com/qgroundcontrol/builds/${BRANCH}/QGroundControl-installer.exe

patrickelectric:
Good to know that ! Ty @DonLakeFlyer and @dagar, if you both think that this PR is not necessary for this development use case, can be closed.

DonLakeFlyer:
Let me think about it a bit. I think there is still a hole to fill with this.

DonLakeFlyer:
Let's just go with that. I think it's good enough. Don't want to add extra time to builds by pushign to another place.

PR #6341, opened 2018-04-11
Link read/write: Better guarding against crashes

DonLakeFlyer:
In looking through Stable 3.3 android crash reports. There are quite a few in readBytes of various link types. These changes guard against possible null pointer references.

### DIFF
--- a/src/comm/UDPLink.cc
+++ b/src/comm/UDPLink.cc
@@ -136,36 +136,35 @@ void UDPLink::removeHost(const QString& host)
 
 void UDPLink::_writeBytes(const QByteArray data)
 {
-    if (!_socket)
-        return;
-
-    QStringList goneHosts;
-    // Send to all connected systems
-    QString host;
-    int port;
-    if(_udpConfig->firstHost(host, port)) {
-        do {
-            QHostAddress currentHost(host);
-            if(_socket->writeDatagram(data, currentHost, (quint16)port) < 0) {
-                // This host is gone. Add to list to be removed
-                // We should keep track of hosts that were manually added (static) and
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
@dogmaphobic Can you double-checlk this?

src/comm/UDPLink.cc
```diff
@@ -136,36 +136,35 @@ void UDPLink::removeHost(const QString& host)
...
+        if(_udpConfig->firstHost(host, port)) {
+            do {
+                QHostAddress currentHost(host);
+                if(_socket->writeDatagram(data, currentHost, (quint16)port) < 0) {
+                    // This host is gone. Add to list to be removed
+                    // We should keep track of hosts that were manually added (static) and
+                    // hosts that were added because we heard from them (dynamic). Only
+                    // dynamic hosts should be removed and even then, after a few tries, not
+                    // the first failure. In the mean time, we don't remove anything.
+                    if(REMOVE_GONE_HOSTS) {
```
dogmaphobic: This is not right. A write error simply means the TCP stack had issues sending the datagram. There is no relationship with a  "gone" host as there is no feed back if the target ever received the datagram.

dogmaphobic:
Other than the weirdness with the "gone" hosts (which was there before your changes), everything else looks fine. In fact, I wonder how it was working before.

DonLakeFlyer:
There has been various people complaing about problems with UDP. Can't quite remember details. Something about QHostAddress with no address in it or something like that.

DonLakeFlyer:
Should whatever the problem is you describe above be fixed in 3.3? Or something for next release?

PR #6348, opened 2018-04-14
Jenkins update to linux and android docker 2018-04-12

dagar:


dagar:
This won't pass until after the docker updates are still building (and the Jenkins builds are restarted).

dagar:
If Jenkins starts archiving the binaries we can kill off travis.

Issue #6352, opened 2018-04-14
Offline Terrain: Crashing on Windows


DonLakeFlyer:
@birchera @dogmaphobic Will need your help on this one. This shows up on Windows because the airmap queries are https and we don't install OpenSSL on windows. That said, it exposes a code path which is crashing. No idea how to fix.

The problem starts here: https://github.com/mavlink/qgroundcontrol/blob/master/src/QtLocationPlugin/QGeoMapReplyQGC.cpp#L156. It spits out: ```Fetch tile error: "Error creating SSL context ()"```.

It will then make it to here: https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L442. Which is incorrect. The tile should have errored out. Instead it builds a tile with 0 bytes of data: ```Received some bytes of terrain data:  0```.

Which then eventually crashes here: https://github.com/mavlink/qgroundcontrol/blob/master/src/TerrainTile.cc#L122.  I'm assuming because _data is some crazy incorrect size or something like that caused by the bad tile build.

I see two bugs here:
* Why didn't the tile error out since it got a network error?
* This code here: https://github.com/mavlink/qgroundcontrol/blob/master/src/TerrainTile.cc#L47 needs more sanity checking on the input data.

DonLakeFlyer:
@antiheavy Here is your crash.

DonLakeFlyer:
FYI: I'll figure out how to deal with the OpenSSL thing. I just needs the bugs fixed.

birchera:
> FYI: I'll figure out how to deal with the OpenSSL thing. I just needs the bugs fixed.

I think the libeay32 is outdated, we had that on one specific computer earlier. I can supply that fix later today.

I'll also propose some data sanity check for the terrain tiles.

Antiheavy:
for what it is worth, i can only get this crash to occur when a fixed wing landing pattern is in the mission plan.  I've tried and tried and cannot reproduce the crash without the fixed wing landing pattern.  With a fixed wing landing pattern the crash sometimes happens quickly, sometimes after many minutes, but it always happens.  Windows 10.

DonLakeFlyer:
Ok I see the pulls for openssl and the tile validation stuff. @dogmaphobic What about the fact that the error doesn't propogate correctly?

DonLakeFlyer:
I don't want to merge those other pulls until the network reply thing is figured out otherwise it will hide the problem.

[...]

PR #6354, opened 2018-04-15
Fix OpenSSL for Windows

birchera:
This should fix the open ssl libraries for certain Windows (Home) systems.
* Update libeay32
* Add libssl32 and sslleay32

@DonLakeFlyer can you please check whether this fixes the problems described in https://github.com/mavlink/qgroundcontrol/issues/6352?


DonLakeFlyer:
Thanks I hadn't look at that and couldn't remember if we installed those. I'll give it a try today.

PR #6355, opened 2018-04-15
Guard terrain tile from reading beyond the bytearray end

birchera:
Fixing the problem when reading in incomplete terrain data.

If the bytearray does not contain all data, the isValid flag never gets set to true, which should prevent any related crashes.


DonLakeFlyer:
Instead of a big pile of if's can't you check for the minimum header size you need?

DonLakeFlyer:
Related to #6352

PR #6358, opened 2018-04-16
Separate terrain from tile maps within cached tile request system.

dogmaphobic:
In response to #6352 

I know nothing about SSL on Windows or the parsing of Elevation data. In looking through the existing code, what I saw were some uses of the system that weren't quite right.

The terrain query code, makes use of `QGeoTiledMapReplyQGC` for obvious reasons. It handles all the caching and Internet requests. On the other hand, `QGeoTiledMapReplyQGC` was meant to be used by the QtLocation plugin and not from an outside "client".

The proper fix, which I started and eventually gave up due to the time it would take to do it properly, would be to create a base, virtual class handling the basics. You request a "tile" and you received it (or an error if that's the case). Internally it looks for the "tile" in the cache and if not found, gets it off the Internet. This is what `QGeoTiledMapReplyQGC` does. But the base, virtual class would limit its function to tile retrieval.

A derived class would provide its services to the QtLocation plugin (along with all the other functions that pertain to map tiles) and the terrain code would instantiate its own, separate implementation.

In absence of that, what I've done was to separate all the code related to terrain data from regular map tiles within `QGeoTiledMapReplyQGC`. There is a new signal, which only the terrain code uses to received the "tile" back (or an error if that's the case). When the request is for terrain data, none of the code within `QGeoTiledMapReplyQGC` call the base class functions (`setFinished()`, `setCached()`, etc.) as these only apply to map tiles.

From the perspective of the terrain code, not much changed. To request a "tile" it still creates an instance of `QGeoTiledMapReplyQGC` as always. The difference is that it connects to just one signal, dedicated to terrain data: 

```
    void terrainDone            (QByteArray responseBytes, QNetworkReply::NetworkError error);
```

This signal is called no matter what and it will either contain the requested data or an error.

With that said, this needs quite a bit of testing, especially under error conditions.

### DIFF
--- a/src/QtLocationPlugin/QGeoMapReplyQGC.cpp
+++ b/src/QtLocationPlugin/QGeoMapReplyQGC.cpp
@@ -123,23 +123,23 @@ QGeoTiledMapReplyQGC::networkReplyFinished()
     }
     QByteArray a = _reply->readAll();
     QString format = getQGCMapEngine()->urlFactory()->getImageFormat((UrlFactory::MapType)tileSpec().mapId(), a);
-
-    // convert "a" to binary in case we have elevation data
+    //-- Test for a specialized, elevation data (not map tile)
     if ((UrlFactory::MapType)tileSpec().mapId() == UrlFactory::MapType::AirmapElevation) {
-
         a = TerrainTile::serialize(a);
-        if (a.isEmpty()) {
-            emit aborted();
-            return;
+        //-- Cache it if valid
+        if(!a.isEmpty()) {
+            getQGCMapEngine()->cacheTile(UrlFactory::MapType::AirmapElevation, tileSpec().x(), tileSpec().y(), tileSpec().zoom(), a, format);
         }
-
+        emit terrainDone(a, QNetworkReply::NoError);
+    } else {
+        //-- This is a map tile. Process and ache it if valid.
+        setMapImageData(a);
+        if(!format.isEmpty()) {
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
Thanks Gus. I'll merge this in first. Which I'll then test separately on Windows before I merge in @birchera pulls.

DonLakeFlyer:
Ok, so I'm not crashing any more. But I'm also not getting the Fetch tile error output which seems strange. I should still be getting that since the tiles are not being downloaded. And I still am setup without new SSL stuff.

DonLakeFlyer:
I didn't turn on TerrainQueryLog I can try that and see what it spits out.

DonLakeFlyer:
Maybe I don't get the SSL thing any more because before I was running old version of SLL which Qt didn't work with. And now I just don't have the DLLs?

DonLakeFlyer:
I was expecting to see this spit out: ``           qWarning() << "Fetch tile error:" << _reply->errorString();
 + ```

dogmaphobic:
Yes, in hindsight, I should have changed those `qCDebug()` to `qWarning()`. There is no more `_reply->errorString()` at the terrain level. It no longer has that kind of deep access. It either gets a "tile" (from anywhere) or it doesn't. It does get a (numeric) `QNetworkReply::NetworkError` error if there was one. In the case above, it's an "unknown" error any way.

PR #6366, opened 2018-04-18
UDP Fixes

dogmaphobic:
* Fixed wrong logic when testing _socket for wrtites. It was never writing (sending anything) out the UDP port. This affected all platforms

* Work around a new Windows issue where the call to list the local network interfaces takes for freaking ever to return, blocking the UDP thread altogether (neither receiving anything nor sending anything). This was Windows only.

During this process, I also found that UDP on Qt 5.9.3 is broken on Windows. If you happen to be using it, it won't talk to anything on UDP. Use 5.9.4 or greater.

### DIFF
--- a/src/comm/UDPLink.cc
+++ b/src/comm/UDPLink.cc
@@ -68,27 +68,6 @@ static QString get_ip_address(const QString& address)
     return QString("");
 }
 
-static bool is_ip_local(const QHostAddress& add)
-{
-    // In simulation and testing setups the vehicle and the GCS can be
-    // running on the same host. This leads to packets arriving through
-    // the local network or the loopback adapter, which makes it look
-    // like the vehicle is connected through two different links,
-    // complicating routing.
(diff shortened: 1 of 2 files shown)


acfloria:
I checked it on Ubuntu 16.04 and the UDP link works again. Thanks for the fix.

Antiheavy:
Seems to work again in Window's too.  thanks!

DonLakeFlyer:
@dogmaphobic Was this caused by a bad merge from Stable (which I recently did to try to fix mobile crashes) or was it just a bug in existing code? Reason I ask is that various folks have been reporting udp problem prior to my stable stuff.

DonLakeFlyer:
Ah, ok. I totally screwed that up with the merge conflict from Stable. I think the windows only problem may have been what other people had been hitting as well. So fingers crossed you changed fixed those other peoples problems as well.

PR #6365, opened 2018-04-17
[WIP] videonode: Do not finish if material is different

patrickelectric:
Fix bluerobotics/qgroundcontrol#164

Signed-off-by: Patrick J.P <patrickelectric@gmail.com>

PR #6368, opened 2018-04-19
Fix video popup on windows

patrickelectric:
This is a weird issue, and very inconsistent, this patch **appears** (we are talking about windows, nothing is for sure) to solve the problem.
Besides me, someone can test and confirm ?
Also, replace #6365.
[AppVeyor download link.](https://ci.appveyor.com/api/buildjobs/rsrnqcvg8pk71q3e/artifacts/QGroundControl-installer.exe)

### DIFF
--- a/src/VideoStreaming/VideoReceiver.cc
+++ b/src/VideoStreaming/VideoReceiver.cc
@@ -218,6 +218,7 @@ VideoReceiver::start()
         return;
     }
 #if defined(QGC_GST_STREAMING)
+    _stop = false;
     qCDebug(VideoReceiverLog) << "start()";
 
     if (_uri.isEmpty()) {
@@ -433,6 +434,7 @@ void
 VideoReceiver::stop()
 {
 #if defined(QGC_GST_STREAMING)
+    _stop = true;
     qCDebug(VideoReceiverLog) << "stop()";
     if(!_streaming) {
         _shutdownPipeline();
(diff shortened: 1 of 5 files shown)


src/VideoStreaming/VideoReceiver.cc
```diff
@@ -872,7 +874,7 @@ VideoReceiver::_updateTimer()
                 stop();
             }
         } else {
-            if(!running() && !_uri.isEmpty() && _videoSettings->streamEnabled()->rawValue().toBool()) {
+            if(!_stop && !running() && !_uri.isEmpty() && _videoSettings->streamEnabled()->rawValue().toBool()) {
```
jaxxzer: We should call this `_stopped`

patrickelectric: Theoretically, this variable comes from the user input and not to handle the pipeline, if stop comes after "stopping the video stream" I would agree.

dogmaphobic:
The code "looks" right. I won't be able to test this (Windows) any time soon though...

patrickelectric:
This PR appears to solve ubuntu problem also, as described [here](https://github.com/bluerobotics/qgroundcontrol/issues/164#issuecomment-383749346).

patrickelectric:
It appears to have a problem related to this modifications, I'll put it as WIP just to make sure.

patrickelectric:
Correction done.

DonLakeFlyer:
@patrickelectric Good to merge now?

patrickelectric:
Yep, everything was tested and working.

PR #6376, opened 2018-04-25
Fixed checked state logic 

dogmaphobic:
Also restore `checkedValue` and `uncheckedValue` properties, which are used by `APMCameraComponent.qml`

### DIFF
--- a/src/FactSystem/FactControls/FactCheckBox.qml
+++ b/src/FactSystem/FactControls/FactCheckBox.qml
@@ -8,7 +8,8 @@ import QGroundControl.Controls 1.0
 
 QGCCheckBox {
     property Fact fact: Fact { }
-
+    property variant checkedValue:   1
+    property variant uncheckedValue: 0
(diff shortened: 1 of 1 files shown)


src/FactSystem/FactControls/FactCheckBox.qml
```diff
@@ -17,5 +18,5 @@ QGCCheckBox {
 
     text: qsTr("Label")
 
-    onClicked: fact.value = checked ? 1 : 0
+    onClicked: fact.value == checked ? checkedValue : uncheckedValue
```
DonLakeFlyer: Shit! I've worked around this all over the place in FactCheckBox usage thinking it was come strange Qml quirk! Not a bug in FactCheckBox.

dogmaphobic:
I bumped into that by chance. I was reinstating `checkedValue` and `uncheckedValue` and happened to notice it :P

Issue #6141, opened 2018-02-14
QGC doesn't play well with other Follow Me sources (DroneCore)


hamishwillee:
DroneCore can act as a source for PX4 follow me mode - supplying position updates ([Guide here](https://docs.dronecore.io/v/develop/en/guide/follow_me.html)). However it can't be used with QGC because as soon as the mode changes to Follow Me QGC starts sending follow me updates - in our  [test](https://github.com/dronecore/docs/pull/94#issuecomment-365507954) QGC starts sending (Lat, Lan) = (0, 0).

I'm not sure of the right solution. Some thoughts:
- QGC should recognise that it did not initiate Follow Me mode and not send updates
- QGC should not send updates if it does not have a valid location source (partial solution)
- Other?

> Ground station operating system.
>  QGroundControl version and build (daily, stable, self-built from source, etc.)

Doesn't matter. (Tested 3.2 on Linux)

>Autopilot board: Pixhawk I, Pixhawk Mini, Pixhawk 2, etc.

Doesn't matter. Tested on simulator.

> Autopilot firmware (e.g. PX4, ArduPilot, custom) and version.

Quad.

> Exact steps to reproduce the problem. Starting from booting QGroundControl to when the problem occurs.

We run the [Follow Me Example](https://docs.dronecore.io/v/develop/en/examples/follow_me.html) with QGC attached on a linux laptop and Sitl. The example is already in the develop source tree - the link is to the docs that are about to be integrated.

hamishwillee:
@julianoes FYI

julianoes:
@hamishwillee I don't know follow me, I'm afraid. I'll let @shakthi-prashanth-m respond.

hamishwillee:
@julianoes OK. It is more a QGCt really a "follow me" question -

shakthi-prashanth-m:
I think QGC can't distinguish whether Mode change was done by user or detected in heartbeat.
From https://github.com/mavlink/qgroundcontrol/blob/4e9b1de50fea7a473facf799e2b5da8f5ad05798/src/FollowMe/FollowMe.h#L25

Follow Me slot `FollowMe::followMeHandleManage` is triggered when Flight mode in QGC tool is **FollowMe**.
https://github.com/mavlink/qgroundcontrol/blob/e37b067c0c77e7993685f761d02c4c232db1b91f/src/Vehicle/Vehicle.cc#L208-L209

Then it starts sending Motion report to vehicle at some interval.

[...]

PR #6380, opened 2018-04-26
Allow manual control of when to send FOLLOW_TARGET messages.

dogmaphobic:
Currently, the GCS location is only sent to the vehicle when the flight mode is "Follow Me". This allows manual control of when (and how frequently) to send these messages. Even if not in follow me mode.


hamishwillee:
May not be applicable, but hope this does not add to this problem with Follow Me: https://github.com/mavlink/qgroundcontrol/issues/6141

DonLakeFlyer:
Perfect

PR #6381, opened 2018-04-26
Download improvements 

patrickelectric:
This PR tries to minimize issues related to SSL.
- Move urls from https to http.
- Ignore SSL errors while downloading data with QGCFileDownload.
- Print human-readable description of SSL errors.

### DIFF
--- a/src/QGCFileDownload.cc
+++ b/src/QGCFileDownload.cc
@@ -76,6 +76,7 @@ bool QGCFileDownload::download(const QString& remoteFile, bool redirect)
     networkRequest.setAttribute(QNetworkRequest::User, localFile);
 
     QNetworkReply* networkReply = get(networkRequest);
+    networkReply->ignoreSslErrors();
     if (!networkReply) {
         qWarning() << "QNetworkAccessManager::get failed";
         return false;
--- a/src/QGCFileDownload.cc
+++ b/src/QGCFileDownload.cc
@@ -87,6 +87,17 @@ bool QGCFileDownload::download(const QString& remoteFile, bool redirect)
     connect(networkReply, static_cast<void (QNetworkReply::*)(QNetworkReply::NetworkError)>(&QNetworkReply::error),
             this, &QGCFileDownload::_downloadError);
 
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Can you please provide way more detail with your pull requests? What exactly are you trying to fix here would be helpful? Are you trying to work around this: https://github.com/mavlink/qgroundcontrol/pull/6354 which has already been fixed?

src/QGCFileDownload.cc
```diff
@@ -76,6 +76,7 @@ bool QGCFileDownload::download(const QString& remoteFile, bool redirect)
     networkRequest.setAttribute(QNetworkRequest::User, localFile);
 
     QNetworkReply* networkReply = get(networkRequest);
+    networkReply->ignoreSslErrors();
```
DonLakeFlyer: I think @dogmaphobic deals with this sort of things in map tiles differently. Would be good to be consistent with that. @dogmaphobic ?

patrickelectric:
Hi @DonLakeFlyer, Sorry for not being so clear. Description updated.

patrickelectric:
Probably:
- QGCFileDownload: Continue download if a SSL error happens  … f6a6c6d
- QGCFileDownload: Print SSL error and SSL Certificate error  … 3fd4d14
Are the best one to continue with this patch, the others just try to minimize any future problem, if there is any.

DonLakeFlyer:
I think @dogmaphobic Has done something different what that to just ignore cert errors for map tiles for examples. Gus?

DonLakeFlyer:
Most of the other SSL stuff is done now. I'm going to pull out the changes to File Download ssl error stuff and create a separate pull for that.

patrickelectric:
Ty Don, let me know if I can help to test it.

jaxxzer:
Is any part of this still relevant? Should we rebase to get the human-readable error messages?

DonLakeFlyer:
I don't think so. But that said SSL doesn't work on latest Ubuntu due to the Qt/SSL strangeness.

PR #6382, opened 2018-04-27
_intersectLinesWithPolygon() bug fixes

guxkung:
- reuse of variable i where it shouldn't
- reduce number of looping in bubble-sorting


DonLakeFlyer:
Makes sense. Will be nice perf improvement for large vertex count polygons which tend to come in from KML files.

PR #6383, opened 2018-04-28
Sensor Cal: Handle case where no mag present [WIP]

nathantsoi:
for https://github.com/PX4/Firmware/pull/9361
and https://github.com/PX4/Firmware/pull/9310

### DIFF
--- a/src/AutoPilotPlugins/PX4/SensorsComponent.cc
+++ b/src/AutoPilotPlugins/PX4/SensorsComponent.cc
@@ -20,11 +20,14 @@ const char* SensorsComponent::_airspeedBreakerParam =   "CBRK_AIRSPD_CHK";
 const char* SensorsComponent::_airspeedDisabledParam =  "FW_ARSP_MODE";
 const char* SensorsComponent::_airspeedCalParam =       "SENS_DPRES_OFF";
 
+const char* SensorsComponent::_magDisabledParam =  "ATT_W_MAG";
+const char* SensorsComponent::_magCalParam =  "CAL_MAG0_ID";
+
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Is the decision on ATT_W_MAG a locked down decision? It doesn't quite match things like how the same concept with airspeed is done.

nathantsoi:
Not at all, I was just getting it working with what we had at the moment. I'm happy to implement whatever is best

DonLakeFlyer:
Ok. This needs to wait then. Also quick look at this pull shows it will crash QGC and/or cause jscript complaints due to the fact ATT_W_MAG will not exist if the user is running QGC against older firmware. For new parameters you always need to guard against them not existing.

nathantsoi:
Ahh, good to know. Could you point me to an existing example of a guard for a missing param in the code that I can learn from?

DonLakeFlyer:
It kind of varies by usage. But for something like this, let's assume you have a new bool parameter which you want to indicate the presence of something. Previously the assumption was that this thing always existed. Pseudocode'ish:

C++:
bool magExists = true; // Always exists on older firmware
if (...->parameterExists("MAG_EXISTS")) {
    fact = controller.getParameter("MAG_EXISTS");
    magExists = fact->rawValue()->toBool();
}

Qml:
property bool _magExists: controller.parameterExists(-1, "MAG_EXISTS") ? controller.getParameterFact(-1, "MAG_EXISTS") : true

Something along those lines.

nathantsoi:
Awesome, thanks!

bkueng:
Thanks @nathantsoi and @DonLakeFlyer , I continued this in https://github.com/mavlink/qgroundcontrol/pull/6510

PR #6388, opened 2018-04-30
UnitsSettings: rework/bugfix metadata translators setup

jaxxzer:
This is some refactoring for readability, as well as a bugfix for temperature units to get the proper translator (checking toLower vs checking both toLower and raw string). There is one outstanding (preexisting) issue:

The units settings work fine for the Temperature category, but not battery.temperature. I do not see why, this should be determined by the json, and they specify the same units:
https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/BatteryFact.json#L35
https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/TemperatureFact.json#L7

### DIFF
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -616,7 +616,7 @@ void FactMetaData::setBuiltInTranslator(void)
         for (size_t i=0; i<sizeof(_rgBuiltInTranslations)/sizeof(_rgBuiltInTranslations[0]); i++) {
             const BuiltInTranslation_s* pBuiltInTranslation = &_rgBuiltInTranslations[i];
 
-            if (pBuiltInTranslation->rawUnits == _rawUnits.toLower()) {
+            if (pBuiltInTranslation->rawUnits.toLower() == _rawUnits.toLower()) {
                 _cookedUnits = pBuiltInTranslation->cookedUnits;
                 setTranslators(pBuiltInTranslation->rawTranslator, pBuiltInTranslation->cookedTranslator);
                 return;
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
The strings in FactMetaData::_rgAppSettingsTranslations are all supposed to be lowercase. The temperature ones are not C/F should b c/f which is why the translation isn't working with the current code. The values in the data structure should be changed not the code.

DonLakeFlyer:
> but not battery.temperature

Bizarre! Not sure what is going on there.

DonLakeFlyer:
> The strings in FactMetaData::_rgAppSettingsTranslations are all supposed to be lowercase.

Oh, wait. Those values also get used as user display which all worked with lowercase before. But doesn't really for C and F which really want to be capitalized on the way out.

src/FactSystem/FactMetaData.cc
```diff
@@ -893,8 +912,15 @@ const FactMetaData::AppSettingsTranslation_s* FactMetaData::_findAppSettingsDist
 {
     for (size_t i=0; i<sizeof(_rgAppSettingsTranslations)/sizeof(_rgAppSettingsTranslations[0]); i++) {
         const AppSettingsTranslation_s* pAppSettingsTranslation = &_rgAppSettingsTranslations[i];
-        if (pAppSettingsTranslation->rawUnits == rawUnits.toLower() &&
-           (pAppSettingsTranslation->unitType == UnitDistance && pAppSettingsTranslation->unitOption == qgcApp()->toolbox()->settingsManager()->unitsSettings()->distanceUnits()->rawValue().toUInt())) {
+
+        if (!(rawUnits == pAppSettingsTranslation->rawUnits) && !(rawUnits.toLower() == pAppSettingsTranslation->rawUnits)) {
```
DonLakeFlyer: Shouldn't this just be one test which does toLower on both strings? More understandable.

src/FactSystem/FactMetaData.cc
```diff
@@ -875,13 +875,32 @@ void FactMetaData::_setAppSettingsTranslators(void)
...
+
+            if (!(_rawUnits == pAppSettingsTranslation->rawUnits) && !(_rawUnits.toLower() == pAppSettingsTranslation->rawUnits)) {
+                continue;
+            }
+
+            uint settingsUnits = 0;
+
+            switch (pAppSettingsTranslation->unitType) {
+            case UnitDistance:
+                settingsUnits = qgcApp()->toolbox()->settingsManager()->unitsSettings()->distanceUnits()->rawValue().toUInt();
```
DonLakeFlyer: While cleaning this up would also be good to put ```qgcApp()->toolbox()->settingsManager()->unitsSettings()``` into a variable and using that instead of repeating all of it the over and over and over again.

DonLakeFlyer:
Missed one?
```
void FactMetaData::setBuiltInTranslator(void)
{
    if (_enumStrings.count()) {
        // No translation if enum
        setTranslators(_defaultTranslator, _defaultTranslator);
        _cookedUnits = _rawUnits;
        return;
    } else {
        for (size_t i=0; i<sizeof(_rgBuiltInTranslations)/sizeof(_rgBuiltInTranslations[0]); i++) {
            const BuiltInTranslation_s* pBuiltInTranslation = &_rgBuiltInTranslations[i];

            if (pBuiltInTranslation->rawUnits == _rawUnits.toLower()) {
                _cookedUnits = pBuiltInTranslation->cookedUnits;
                setTranslators(pBuiltInTranslation->rawTranslator, pBuiltInTranslation->cookedTranslator);
                return;
            }
        }
    }

    // Translator not yet set, try app settings translators
    _setAppSettingsTranslators();
}
```

src/FactSystem/FactMetaData.cc
DonLakeFlyer: Which you can't do because one is a char* not a QString.

DonLakeFlyer: Maybe just move pAppSettingsTranslation->rawUnits into a QString ahead of time so you can toLower it.

DonLakeFlyer:
This is why battery.temperature doesn't work:
```
void FactMetaData::_setAppSettingsTranslators(void)
{
    // We can only translate between real numbers
    if (!_enumStrings.count() && (type() == valueTypeDouble || type() == valueTypeFloat)) {
        for (size_t i=0; i<sizeof(_rgAppSettingsTranslations)/sizeof(_rgAppSettingsTranslations[0]); i++) {
            const AppSettingsTranslation_s* pAppSettingsTranslation = &_rgAppSettingsTranslations[i];
            if (pAppSettingsTranslation->rawUnits == _rawUnits.toLower() && (
                (pAppSettingsTranslation->unitType == UnitTemperature && pAppSettingsTranslation->unitOption == qgcApp()->toolbox()->settingsManager()->unitsSettings()->temperatureUnits()->rawValue().toUInt()) ||
                (pAppSettingsTranslation->unitType == UnitSpeed       && pAppSettingsTranslation->unitOption == qgcApp()->toolbox()->settingsManager()->unitsSettings()->speedUnits()->rawValue().toUInt()) ||
                (pAppSettingsTranslation->unitType == UnitDistance    && pAppSettingsTranslation->unitOption == qgcApp()->toolbox()->settingsManager()->unitsSettings()->distanceUnits()->rawValue().toUInt()) ||
                (pAppSettingsTranslation->unitType == UnitArea        && pAppSettingsTranslation->unitOption == qgcApp()->toolbox()->settingsManager()->unitsSettings()->areaUnits()->rawValue().toUInt())))
            {
                _cookedUnits = pAppSettingsTranslation->cookedUnits;
                setTranslators(pAppSettingsTranslation->rawTranslator, pAppSettingsTranslation->cookedTranslator);
                return;
            }
        }
    }
}
```

Int translation is not supported.

jaxxzer:
> Int translation is not supported.

Seems like a weird restriction. Any idea why?

jaxxzer:
> Shouldn't this just be one test which does toLower on both strings? More understandable.
Agreed.

jaxxzer:
Thanks for the review @DonLakeFlyer. Will apply requested changes.

DonLakeFlyer:
> Seems like a weird restriction. Any idea why?

Due to possible loss of precision from translation. I think that kind of makes sense. Not for this specific use case of battery.temp. But for an int parameter the value could change when the user edited it. Instead of dealing with unexpected consequences of allowin int translation. It may be better to change battery.temp to double with no decimals of 0.

jaxxzer:
Aha. Kinda sorta makes sense now. I understand the consideration of int params changing at least.

jaxxzer:
@DonLakeFlyer changes applied

DonLakeFlyer:
All good now, thanks.

Issue #6387, opened 2018-04-30
*Daily build* - Rel.Abs.AGL. in Waypoint editor lose sync


Antiheavy:
testing Daily build 2018-04-29

three screenshots with the AGL, Abs, Rel radio selectors, notice the altitude preview pane in the lower left.  There is no change between AGL and Abs (should be a very noticable change) and there is a strangely  large change between Abs and Rel. (wrongly).  Seems to depend on which order you click/switch between the different altitude modes.
![image](https://user-images.githubusercontent.com/12117027/39413213-c259464e-4bec-11e8-9a0f-0978ffaffd2a.png)
![image](https://user-images.githubusercontent.com/12117027/39413217-d3da9102-4bec-11e8-90fa-7bbb687fc5cc.png)
![image](https://user-images.githubusercontent.com/12117027/39413221-df83e68e-4bec-11e8-8d1b-d2f370557ec5.png)

DonLakeFlyer:
Ah, I knew I'd miss something. This is all new, like 10 minutes old new! Some signalling is missing somewhere.

DonLakeFlyer:
Fixed

Issue #6342, opened 2018-04-12
MAV_FRAME_GLOBAL_TERRAIN_ALT not showing as Show all values


AndKe:
I tried to use current master profesionally today.
QGC detected and handled RTK great.
I drew a mission , all waypoints were terrain relative, 1.5m above (by LIDAR).
When trying to upload I QGC denied to do so because of missing /fetching terrain data 
(I have a screenshot with the exact wording, not here.) 

I did get map tiles, so even with only GPRS link, we did had internet connection, but QGC would stop the show.  (if I did not had other GCS systems to choose from)

in any case:
fetching terrain data should NOT be enforced that hard, especially when we fly with LIDAR as primary data source.

This is rather serious, it would stop the operation.

AndKe:
Err I would disagree.
Ehen EKF is set to use Lidar as primary altitude source, a mission with *no* elevation data will/does work just fine, in fact, the elevation data is irrelevant/not used as the lidar will define the altitude for those points.

So this mission is a valid one, perfectly working with apmplanner/mavproxy etc, but QGC decide to enforce the requirement of terrain elevation.  - which apparently failed (and does not have a progess bare or other visual indicators.)

the area covered were less than 200x100 meters, so it was not a lot of data, to day the least. :)

In any case: I would think it was nice to warn me about missing elevation data, but denying operation is overkill for any user that know ArduPilot.

DonLakeFlyer:
I think you are missing the point. This feature does not support ArduPilot terrain following. It just is setting absolute heights (MAV_FRAME_GLOBAL) into mission items. It is not setting the MAV_FRAME on mission items to MAV_FRAMEGLOBAL_TERRAIN_ALT.

I think in mission planner there is a thing called "check heights" or something like that which adjust the mission items heights. It is like that.

AndKe:
I do not know about mission planner, can run it in mono as a test, but it's a windows thingy :)

So I used a test mission linke:
```

QGC WPL 110
0	1	0	16	0	0	0	0	69.7561874389648438	18.8628387451171875	241.979995727539063	1
1	0	10	16	0	0	0	0	69.7556915283203125	18.8636550903320313	1.5	1
2	0	10	16	0	0	0	0	69.7549414195638349	18.8628494739532471	1.5	1
3	0	10	16	0	0	0	0	69.7550082456289431	18.8623988628387451	1.5	1
[...]

PR #6389, opened 2018-04-30
Mission Item: Mavlink MAV_FRAME_GLOBAL_TERRAIN_ALT support

DonLakeFlyer:
screen shot 2018-04-29 at 11 35 37 pm

* New **TerrF** radio for altitude mode. This supports MAV_FRAME_GLOBAL_TERRAIN_ALT.
* TerrF radio will only show up for firmwares which support it: currently ArduPilot only
* Also fixed height calculations for alt/terrain display when switching between alt modes

Fixes for #6342 and #6387


DonLakeFlyer:
Note that although QGC supports MAV_FRAME_GLOBAL_TERRAIN_ALT it does not support terrain altitude height requests from the vehicle. So this will only work with ArduPilot if you are using a lidar height sensor. Or however you want terrain height work using a height sensor with ArduPilot.

PR #6404, opened 2018-05-01
Convert Survey complex item to use TransectStyleComplexItem

DonLakeFlyer:
screen shot 2018-05-01 at 2 49 16 pm

screen shot 2018-05-01 at 2 48 32 pm

This is a massive change to the Survey complex item to convert it to the user same new transect generation  and camera calculation code as Corridor Scan (TransectStyleComplexItem, CameraSpec, CameraCalc). The most important thing this does for Survey is add Terrain support.

I can pretty much guarantee that I introduced bugs to Survey whether you are using terrain following or not. Most likely problem areas:
* Terrain generated surveys. Be careulf with using this to start. Verify by saving to KML. Also verify by round trippping through the vehicle and double checking generated waypoints.
* Loading old Survey files. This conversion also massively changes the Survery json file format. The good news is that it matches Corridor Scan and provides common objects for camera and grid information. The bad news is, well it's different from the old format. So the old formats need to be converted on the fly.

So I could certainly used help hammering on this to wring out the bugs.

Features which are not yet re-implemented (coming in the next few days):
* Hover and Capture
* Refly at 90 degress
These don't get used much and add more complexity to an already massive pull. Do I decided for releas without them for the first go round.

Related to #6204.

### DIFF
--- a/src/MissionManager/CameraSpec.cc
+++ b/src/MissionManager/CameraSpec.cc
@@ -121,8 +121,8 @@ bool CameraSpec::load(const QJsonObject& json, QString& errorString)
 
     _sensorWidthFact.setRawValue        (json[_sensorWidthName].toDouble());
     _sensorHeightFact.setRawValue       (json[_sensorHeightName].toDouble());
-    _imageWidthFact.setRawValue         (json[_imageWidthName].toDouble());
-    _imageHeightFact.setRawValue        (json[_imageHeightName].toDouble());
+    _imageWidthFact.setRawValue         (json[_imageWidthName].toInt());
+    _imageHeightFact.setRawValue        (json[_imageHeightName].toInt());
     _focalLengthFact.setRawValue        (json[_focalLengthName].toDouble());
     _landscapeFact.setRawValue          (json[_landscapeName].toBool());
     _fixedOrientationFact.setRawValue   (json[_fixedOrientationName].toBool());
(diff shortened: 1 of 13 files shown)


DonLakeFlyer:
I forgot the include the fact that in reworking the Survey UI I temporarily remove the fixed wing compass rose grid angle changer thingy. I'll be putting that back in soon as well.

dgorissen:
This is great, thanks Don! I had a quick try with todays daily build. Without thinking I created (in retrospect) a far too big survey and I would get stuck on the  "cannot save file, waiting for terrain data to download" message. A smaller survey worked fine (verified in google earth).

Perhaps some kind of progress bar to show terrain loading and/or a warning when creating massive surveys (or ability to cancel)?

As an aside, while testing I found a fairly reproducible crash:
- Start QGC
- I load the following survey: [surveytest.plan.txt](https://github.com/mavlink/qgroundcontrol/files/1974033/surveytest.plan.txt)
- Go to the console view, scroll around a bit
- Click back on the flight-plan icon
- Segfault

image

Im on OSX El Capitan. And gad enabled logging for terrain related things (2 items) and set the debug level to 4 (if that matters).

DonLakeFlyer:
> I would get stuck on the "cannot save file, waiting for terrain data to download" message.

I believe there is a bug in querying for a large number of terrain values. It should work. The first time you may need to way as it caches the data. But then once your area is cached it should be fast. Could you enter a separate issue for that. I keep forgetting to debug it.

Can you also put in two separate issues for the terrain progress feature and the crash. Thanks.

DonLakeFlyer:
And thanks for testing this out.

DonLakeFlyer:
Crashes should be fixed now. Also this: https://github.com/mavlink/qgroundcontrol/pull/6463 will show you terrain download progress.

Issue #5769, opened 2017-10-20
Parameter progress bar visibility in daylight.


AndKe:
The thin green progress bar is very hard to see when in difficult viewing conditions. 
When at >95% , it's even hard to see if it IS there (as in 95% ), or disappeared - as in 100%.
please consider adding a plain , smaller, progress bar with numeric count, something like:
             [|||||||| 120/770        ]

or a significantly thicker green line

Antiheavy:
I agree the current thin bar is too hard to see.  Especially when it gets hung up at ~95% and you're squinting and asking yourself "is that still going? or did it complete?...."

PR #6408, opened 2018-05-02
Large parameter progress on outdoor theme

DonLakeFlyer:
Fix for #5769. This was making me crazy as well so I finally did something about it:
* Only visible if using Outdoor theme
* Only visible while parameters are downloading
* You can click it to hide it if you want to move on to do something else.

screen shot 2018-05-02 at 3 37 27 pm

### DIFF
--- a/src/ui/toolbar/MainToolBar.qml
+++ b/src/ui/toolbar/MainToolBar.qml
@@ -206,12 +206,60 @@ Rectangle {
         }
     }
 
-    // Progress bar
+    // Small parameter download progress bar
     Rectangle {
-        id:             progressBar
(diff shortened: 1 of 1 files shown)


juvinski:
Hi Don,

and about the mission download progress bar ? Based on your change I tested on PlanToolBar and I think is  a great feature too.

DonLakeFlyer:
> and about the mission download progress bar

Not sure what you mean. That already has a big Done thing which is pretty easy to see.

juvinski:
Hi, yes the done message is nice but the download bar is hard to see under sunlight

DonLakeFlyer:
Ok, I get it now

Issue #5730, opened 2017-10-05
**Feature request** improvements to replay function


Antiheavy:
Replay is a VERY useful function to have. The current implementation is too basic. These two improvements would make it much more useable:

* Ability to play forward and play backward.
* Ability to play forward/backward and faster/slower speeds.

Here is a couple concepts of how that might look:
![image](https://user-images.githubusercontent.com/12117027/31240565-bb008fc4-a9c6-11e7-9c0a-ae2cb169419a.png)

somewhat related to:
https://github.com/mavlink/qgroundcontrol/issues/4268
https://github.com/mavlink/qgroundcontrol/issues/3601
https://github.com/mavlink/qgroundcontrol/issues/3375
https://github.com/mavlink/qgroundcontrol/issues/7412

fredrikfalkman:
Also, the ability to enter a specific time would be nice. And of course, live scrubbing!

Antiheavy:
> , live scrubbing!

live scrubbing works somewhat better in recent builds of QGC.

Antiheavy:
It was suggested to provide more context to this request and specific things people are trying to do with replay.  Here are some more thoughts:

1) **Case of a lost vehicle:** For purposes of trying to find the vehicle it is very helpful to know last known: location, altitude, ground course, speeds, and attitude (angles).  A print out of these things could be helpful, but a replay that allows easy forward/reverse and slow/fast playback can really help reconstruct the final moments before the vehicle was lost to determine both what cause the vehicle to be lost and also how large the search area might be.

2) **Analyze widget playback:**  sometimes we'd like to playback certain telemetry items and watch them slower or faster.  Often we might want to replay a large section of log back very quickly to build out those strip charts faster.

Antiheavy:
I think I can close this issues as it is generally address by: https://github.com/mavlink/qgroundcontrol/pull/7686

There are still some headaches, but I've opened them as stand alone issues:
https://github.com/mavlink/qgroundcontrol/issues/7710
https://github.com/mavlink/qgroundcontrol/issues/7709
https://github.com/mavlink/qgroundcontrol/issues/7708

PR #6414, opened 2018-05-03
Slightly better Log Replay

DonLakeFlyer:
* Status Bar shows current time in log
* Parameter Manager no longer complains about timeouts

### DIFF
--- a/src/FactSystem/ParameterManager.h
+++ b/src/FactSystem/ParameterManager.h
@@ -7,9 +7,7 @@
  *
  ****************************************************************************/
 
-
-#ifndef ParameterManager_H
-#define ParameterManager_H
+#pragma once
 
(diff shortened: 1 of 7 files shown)


DonLakeFlyer:
Related to #5730

src/FactSystem/ParameterManager.h
```diff
@@ -7,9 +7,7 @@
  *
  ****************************************************************************/
 
-
-#ifndef ParameterManager_H
-#define ParameterManager_H
+#pragma once
```
patrickelectric: yay!

DonLakeFlyer: Yeah, if you're even in a file working and see the old style switch it over.

PR #6425, opened 2018-05-05
Video: Support for OpenSolo

AndKe:
attempting to get Solo video for video any quad.  - no side effects for non-solo vehicles.

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -366,6 +366,10 @@ bool APMFirmwarePlugin::_handleIncomingStatusText(Vehicle* vehicle, mavlink_mess
                     supportedMinorNumber = 4;
                     break;
                 case MAV_TYPE_QUADROTOR:
+                    // Start TCP video handshake with ARTOO in case it's a Solo
+                    qDebug() << "Trying to request Solo Video";
+                    _soloVideoHandshake(vehicle);
+                    break;
                 case MAV_TYPE_COAXIAL:
                 case MAV_TYPE_HELICOPTER:
                 case MAV_TYPE_SUBMARINE:


DonLakeFlyer:
@dogmaphobic Can you comment on this. Wondering if there would be an implications if this ends up connected to a port which isn't the solo video start thingy. Basically all this does is this:

```
const char* APMFirmwarePlugin::_artooIP =                   "10.1.1.1"; ///< IP address of ARTOO controller
const int   APMFirmwarePlugin::_artooVideoHandshakePort =   5502;       ///< Port for video handshake on ARTOO controller

void APMFirmwarePlugin::_soloVideoHandshake(Vehicle* vehicle)
{
    Q_UNUSED(vehicle);

    QTcpSocket* socket = new QTcpSocket();

    socket->connectToHost(_artooIP, _artooVideoHandshakePort);
    QObject::connect(socket, static_cast<void (QTcpSocket::*)(QAbstractSocket::SocketError)>(&QTcpSocket::error), this, &APMFirmwarePlugin::_artooSocketError);
}
```

Which happens to kick off the video to stream on the Solo. The idea is to do for all ArduPIlot quads. Do you think connecting to same randim IP/Port on a vehicle which is not a Solo would cause a problem. It doesn't send anything out on that port.

@AndKe I'm thinking in order to make this work. The connect and _artooSocketError should be removed. Since it will likely fail on non-solo vehicle. Did you try what happens on a Quad which is not a solo?

DonLakeFlyer:
I also asked the ArduPilot folks on their gitter channel if there was some way they had to identify when this handshake was needed.

AndKe:
@DonLakeFlyer 
The connection is not to Solo, but to the controller(which acts as router) , 10.1.1.1 , (solo is 10.1.1.10)
There are no side effects for a non-solo computer, should it have even have the same IP , then it would require that that said computer had port 5502 (not a common port for services) open for connections - for anything to happened, that would only open a connection - nothing more would happen.   
 (10.0.0.0/8 net is private net range, and non-routable to internet)

If a device on given address is not present, or the port not accepting connections (regardless if it's denying connection, or simply dropping those packets), the connection would simply fail, - which you can test very easily by running my branch , while running a quad in SITL (it will try, but you won't see any effect)   

It's totally harmless.  - and yes, I did test with SITL.

You *could* detect solo, not by mavlink from ArduCopter, but by Artoo connection or maybe something in dronekit that can tell. - that would just be a overly complex solution.

In the meantime, "nc 10.1.1.1 5502"  is the Linux command that does the same job, you can fire it away to any and all of your devices to see nothing happen :)

DonLakeFlyer:
> the connection would simply fail, - which you can test very easily by running my branch

Ok. That will still seem to mean that the connect signal error path should be removed? Otherwise it is going to pop an error on vehicles without the artoo controller thing wont it. Since the connect fails, the signal will trigger and the error will come up.

dogmaphobic:
> Wondering if there would be an implications if this ends up connected to a port which isn't the solo video start thingy.

As stated above, the connection itself is quite harmless. It's a private network and on a non-standard port. The chances of bumping into something you shouldn't is virtually zero.

With that said, is that all it takes to start the video? Just to make a TCP connection and nothing exchanged? That's quite odd and not exactly reliable. 

Also, the code creates a socket and a connection but the socket is never deleted/cleared. It has a, so far useless error handler. It should at minimum delete the socket, which is out of scope at that point. 

Is the video controlled by an "open connection"? How to do you stop the video? Closing the TCP socket? If that's the case, why isn't it implemented?

In summary, it all may "work" but it seems like a hack.

AndKe:
@dogmaphobic 

> Just to make a TCP connection and nothing exchanged?

yes, Controller will start forwarding video to the IP that made connection (QGC) port 5600

> How to do you stop the video?

you don't , dont need too. The controller will transmit to another device if another device creates the port 5502 connection, otherwise, it does not care if the host port closes.
AKA: if I have a 5502 connection, then kill the video player (port 5600 closes), - it will resume just fine if I open the video player.   - If I terminate the 5502 connection, 5600 will stop getting data as well. 

@DonLakeFlyer 
> Otherwise it is going to pop an error on vehicles without the artoo controller thing wont it.

I haven't seen that part of code work,  there is no need for any error handling.

DonLakeFlyer:
Should be all good to go now with my tweaks to @AndKe pull in the replacement pull. Thanks @AndKe for pushing on this.

Issue #5870, opened 2017-12-07
Solo video won't work with updated solos.


AndKe:
Currently, we check for Solo by looking for STATUSTEXT "APM:Copter solo-*"
if it's found, we do the connectToHost(_artooIP, _artooVideoHandshakePort);
which would activate the Solo video..   
More and more Solos are running ArduCopter code, and they er not that easy to identify.
... and it's kind of dirty to expect user to run "nc 10.1.1.1 5502"  to make it work.

I see two good solutions:
1: try the connection blindly, it does not harm as long it does not generate lots of traffic (which it dont)
2: have an checkbox, an option to try this (and/or maybe other ip/port) connection.

in case of solution 1 , I could test and submit a PR with that...

BigOkie:
I believe this is happening to me as well; running a Solo with Arducopter 3.5.4 firmware.  Tower displays video just fine, QGC not so much.

dagar:
@BigOkie can you expand on that? Does QGC video sometimes work, but intermittently?

DonLakeFlyer:
The Solo needs to be told from the ground station to start video. The trigger for QGC to do that is based on text message output which doesn't happen when you are using an ArduPilot build on the solo. Only for original Solo firmware. I just haven't had time to get around fixing this yet.

> in case of solution 1 , I could test and submit a PR with that...

@AndKe If you could make the pull that would be great.

AndKe:
https://github.com/mavlink/qgroundcontrol/pull/6425

I know that the IMX6 from solo have been integrated into some lager RPAS, so if we se absolutely no side effects, this approach may very well be expanded to any vehicle.

- testet, works fine for OpenSolo + ArduCopter  (not only the old Solo-fork)

BigOkie:
@dagar I used to get it to work intermittently but thinking back on that now it was likely due to me still having the Solo firmware loaded, and not Arducopter.  I'll just wait for the correction.  I have Solex also and while it seemed to have given me fits in the past, it seems to be working much better now.  Not sure why.

PR #6430, opened 2018-05-07
ArduPilot: Start possible Solo video

DonLakeFlyer:
Replacement for #6425.

* Fixes a bug in #6425 
* Handles possible errors coming back from connect. When we specifically know what are connected to a Solo show the error if a video start fails. When we are unsure of Solo connection don't show errors from video start, since these will popup when the vehicle is not a Solo.


DonLakeFlyer:
Fixed #5870

PR #6431, opened 2018-05-07
Vehicle. Expose sensor status to Qml

DonLakeFlyer:
Related to #6364. FYI @philipoe 

The various bits from the mavlink SENSOR_STATUS message are now available from the Vehicle object in Qml using .sensorsPresentBits, .sensorsEnabledBits, .sensorsHealthBits. Including an enum for the bit values. So you can write code like this:

```
import QGroundControl.Vehicle 1.0

property bool magEnabled: _activeVehicle.sensorsEnabledBits & Vehicle.SysStatusSensor3dMag
```

Also Phillip: If you want I can make a set of unhealthy bits like there are on the c++ side, which is a combination of enabled & healthy if you want as well.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -1302,6 +1302,7 @@ void Vehicle::_handleSysStatus(mavlink_message_t& message)
     if (newSensorsUnhealthy != _onboardControlSensorsUnhealthy) {
         _onboardControlSensorsUnhealthy = newSensorsUnhealthy;
         emit unhealthySensorsChanged();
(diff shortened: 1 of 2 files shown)


philipoe:
> Also Phillip: If you want I can make a set of unhealthy bits like there are on the c++ side, which is a combination of enabled & healthy if you want as well.

This would be perfect, because then i could completely remove the cumbersome string parsing in https://github.com/mavlink/qgroundcontrol/pull/6364/ .

DonLakeFlyer:
I'll add that to this pull.

philipoe:
Ready to merge? From my side this is OK.

PR #6436, opened 2018-05-09
Move ArduPilot PreArm handling to generic

DonLakeFlyer:
* PreArm duplicated messages are filtered such that they are spoken only if more than 10 seconds from last time spoken.
* PreArm messages do not pop up a QGC message box
* PreArm messages still show up in message dropdown
* PreArm messages show up in the center of the flight map. If no PreArm message shows up again for 10 second it will be removed from the flight map.

@dagar Can you test this with PX4?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,7 +15,7 @@ environment:
 install:
   - git submodule update --init --recursive
   - call "%ProgramFiles(x86)%\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86
-  - set PATH=C:\Qt\Tools\QtCreator\bin;C:\Qt\5.9.4\msvc2015\bin;%PATH%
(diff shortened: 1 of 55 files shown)


DonLakeFlyer:
@dagar Can you test?

dagar:
Great! I'll give it a try.

Could we also do this for anything that contains "PREFLIGHT"? https://github.com/PX4/Firmware/blob/master/src/modules/commander/PreflightCheck.cpp#L131

On the PX4 side I'd still like to change them eventually, but that would get us through the transition.

dagar:
Which MAV_SEVERITY do you think these should be?

dagar:
I'm thinking about treating it differently if the check is done periodically vs as a result of the user requesting to arm.

DonLakeFlyer:
Rebase went bonkers.

PR #6455, opened 2018-05-16
[WIP]: added important tuning parameters/sliders for plane

barzanisar:
These are all the tuning parameters Fixed Wing users find useful for tuning, as discussed with @RomanBapst.

### DIFF
--- a/src/AutoPilotPlugins/PX4/PX4TuningComponentPlane.qml
+++ b/src/AutoPilotPlugins/PX4/PX4TuningComponentPlane.qml
@@ -27,7 +27,7 @@ SetupPage {
             sliderModel: ListModel {
                 ListElement {
                     title:          qsTr("Roll sensitivity")
-                    description:    qsTr("Slide to the left to make roll control faster and more accurate. Slide to the right if roll oscillates or is too twitchy.")
+                    description:    qsTr("FW_R_TC: Slide to the left to make roll control faster and more accurate. Slide to the right if roll oscillates or is too twitchy.")
                     param:          "FW_R_TC"
                     min:            0.2
                     max:            0.8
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I don't necessarily agree with the direction this is heading. Setup pages are for parameters which are commonly changed. This seems like every fixed wing tuning parameter that is possible. Do most normal users really need to tune all of these things! Parameters which are rarely changed are left to be edited in the parameter editor.

Think of it this way:
* The user picks a standard fixed wing airframe which in turns sets various defaults.
* Beyond that what do they really need to tweak to get a flying vehicle.

There are multiple modes of tuning:
* Tuning from a standard airframe, for non-power users. This is what the tuning page is mostly now.
* PID tuning which is power user stuff and require charts and what not. This is what the Advanced thing from daily builds is heading towards.
* Power user tuning of all parameters. This is more left to the parameter editor. The current goal is not to have a page which contains a UI for every parameter on the vehicle.

DonLakeFlyer:
The PID tuning and Power user tuning of all related params could theoretically go together on the UI. That way you have a simple ui for non-power users, and the advanced one for people who know whta they are doing.

DonLakeFlyer:
I think something like this could work:
* Page: Simple tuning params (like throttle hover for multi-rotor)
* Page: PID Tuning
* Page: Full set of tunable params

Opinions?

@RomanBapst Is this list of params all really simple tuning, or could they be split up?

barzanisar:
> Page: Full set of tunable params

Why do we need a page for full set of params when we have them in params? Maybe we can have just the commonly used ones in the "Page: PID Tuning".

DonLakeFlyer:
> Why do we need a page for full set of params when we have them in params?

I'm not sure this is needed. I think you just do to parameter editor for that.

PR #6459, opened 2018-05-17
PreFlight CheckList changes

DonLakeFlyer:
* Less layering. I reorganized the control hierarchy to have everthing in a single control. Also removed uneeded control layering to simplify things. This made it way easier to flow the sizing automatically up/down the hierarchy.
* Allow DropPanel controls to deal with interior component changing size dynamically
* If there is not vehicle the button is not enabled
* Vehicle id only shows when there are multiple vehicles
* Rename some controls to make use clearer
* Fixed some coding style and missing qsTr

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -150,6 +150,7 @@
         <file alias="QGroundControl/FlightDisplay/GuidedActionsController.qml">src/FlightDisplay/GuidedActionsController.qml</file>
         <file alias="QGroundControl/FlightDisplay/GuidedAltitudeSlider.qml">src/FlightDisplay/GuidedAltitudeSlider.qml</file>
         <file alias="QGroundControl/FlightDisplay/MultiVehicleList.qml">src/FlightDisplay/MultiVehicleList.qml</file>
+		<file alias="QGroundControl/FlightDisplay/PreFlightCheckList.qml">src/FlightDisplay/PreFlightCheckList.qml</file>
         <file alias="QGroundControl/FlightDisplay/qmldir">src/FlightDisplay/qmldir</file>
         <file alias="QGroundControl/FlightMap/CameraTriggerIndicator.qml">src/FlightMap/MapItems/CameraTriggerIndicator.qml</file>
         <file alias="QGroundControl/FlightMap/CenterMapDropButton.qml">src/FlightMap/Widgets/CenterMapDropButton.qml</file>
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
@philipoe Can you take a look at this? I'm not sure if this control layering is what it will finally end up being. Supporting custom pre-flight checklists may change that. For for now it makes the code much easier to work with.

Qml Coding Style:
```
            QGCButton {
                id:         resetButton
                width:      1.2 * ScreenTools.defaultFontPixelHeight
                height:     1.2 * ScreenTools.defaultFontPixelHeight
                opacity:    0.2 + (0.8 * (QGroundControl.multiVehicleManager.vehicles.count > 0))
                tooltip:    qsTr("Reset the checklist (e.g. after a vehicle reboot)")

                onClicked: checklist.resetNrClicks()

                Image { source:"/qmlimages/MapSyncBlack.svg" ; anchors.fill: parent }
            }
```

Note the tab allignment of properties. Also always a space around operators (+, *, ...). And also there are places with no space after ```{``` or before ```}```. Also lots of missing using os ```qsTr()```. Could you make those changes as well?

philipoe:
Tested this. Issues found:

- The initial bug that i found here https://github.com/mavlink/qgroundcontrol/pull/6364/commits/13358501e3736921da3d0c7641e506c094a471af , i.e. the fact that once you close the check list, all check list items that depend on the number of times the user clicked them get reset, is now back. This was the very reason why in the commit above i separated the GUI (dropPanel, rectangle, etc objects) from the checklist data (which was in the CheckList.qml file). I think this bug is quite annoying for a user so we need to avoid it.
- The "reset checklist" button is now actually _enabled_ when there is no vehicle. This is not the case for me in the current master (see old PR at https://github.com/mavlink/qgroundcontrol/pull/6364, where you said it is enabled for you, but for me it always worked). So somethign must be wrong.

General comments:
 - I specifically introduced a) the ObjectModel structure and b) the separation of the GUI (in FlightDisplayView.qml) and the data (in CheckList.qml) to give us more flexibility w.r.t. custom checklists, i.e. the idea was that the CheckList.qml would just load the custom checklist items (maybe even from an xml file, which ObjectModel actually supports) and then the GUI in FlightDisplayView.qml just displays that custom data via the ObjectModel. This seems to be the main thing that was reverted in this PR. I have limited experience in QML/QGC, but thought my initial approach makes sense (it also as mentioned above avoids the "check list items reset when the check list is closed" bug).

DonLakeFlyer:
> the fact that closing the check list drop panel resets all check list items that depend on user clicks, is now back.

Hmm, I totally screwed that up. I'll fix that.

> I specifically introduced a) the ObjectModel structure and b) the separation of the GUI (in FlightDisplayView.qml, but this could of course also be outsourced into a PreFlightCheckListGUI.qml file) and the data (in CheckList.qml) to give us more flexibility w.r.t. custom checklists, i.e. the idea was that the CheckList.qml would just load the custom checklist items (maybe even from an xml file, which ObjectModel actually seems to support) and then the GUI in FlightDisplayView.qml just displays that custom data via the ObjectModel.

Yeah I get that, but there was too much control layering making it difficult to work with (pass sizing information up/down hieraerchy) and understand. I"m still thinking about how to customize. I have some ideas but I'll write up separately.

DonLakeFlyer:
> Yes, but do you mean just locally or in this PR?

I meant after this PR goes through.

DonLakeFlyer:
@philipoe This screws up your original stuff too much. I'm closing and will just submit a drop panel resize fix.

philipoe:
Ok, i think i still got what you are aiming for (including the fixes you proposed). I will wait for whatever PR you put up next but would otherwise fix those things in my next PR.

Issue #6440, opened 2018-05-10
Survey: Custom builds can no longer default to specific camera


DonLakeFlyer:
Broken by TransectStyleComplexItem changes

Antiheavy:
https://github.com/mavlink/qgroundcontrol/pull/6464

PR #6464, opened 2018-05-18
Save fact settings for new Transect code

DonLakeFlyer:
Survey, Corridor Scan, Structure Scan now save last settings. This also allows custom builds to override all these values.


DonLakeFlyer:
Fix for #6440

Issue #6438, opened 2018-05-10
Tuning page discussion


barzanisar:
![image](https://user-images.githubusercontent.com/26625030/39863455-fb51b8c0-5446-11e8-8b53-33468762a424.png)

When I compile master or [stable v3.3](https://github.com/mavlink/qgroundcontrol/tree/Stable_V3.3) branch, I get these weird versions of tuning sliders. However, when I open the binary version, the sliders look normal. Can anyone figure out why this is so? Is there any special way of compiling to see the tuning sliders?
@DonLakeFlyer @dogmaphobic

DonLakeFlyer:
If you are getting the new ui for sliders on 3.3 branch then something is wrong with your build. I'd delete build directory and start again. I just tried building stable 3.3 and it was fine using the old slider ui. And complaining about missing roll/pitch parameters.

Not quite sure what you mean by wierd? On master the sliders use a new ui control for changing values. Also master has roll/pitch removed since the parameters controlling those are gone.

barzanisar:
Ok so I checked out a branch from master to solve https://github.com/PX4/Firmware/issues/9379

[These](https://github.com/barzanisar/qgroundcontrol/blob/pr-barza-tuning-sliders/src/AutoPilotPlugins/PX4/PX4TuningComponentCopter.qml#L65-L207) are the changes I made.

This is what the tuning sliders look like:

![image](https://user-images.githubusercontent.com/26625030/39922924-d997a27c-5521-11e8-8b4e-3b699572025c.png)

![image](https://user-images.githubusercontent.com/26625030/39922960-fcc187fe-5521-11e8-995e-c528d422eeb7.png)

![image](https://user-images.githubusercontent.com/26625030/39922984-11501898-5522-11e8-9d53-d9cd32063534.png)

I don't know why some of the sliders show negative values, while some (for e.g. MC_PITCHRATE_FF) show zero. 

Also, is this the new UI for tuning sliders or something is wrong?

DonLakeFlyer:
> Also, is this the new UI for tuning sliders or something is wrong?

This is the new UI for tuning sliders.

> I don't know why some of the sliders show negative values, while some (for e.g. MC_PITCHRATE_FF) show zero.

Most likely missing parameter meta data problems. Can you double check the meta data for those params and make sure it is fully specified. Min/Max/Increment.

barzanisar:
[...]

PR #6465, opened 2018-05-18
Use tabs for Tuning page

DonLakeFlyer:
Related to discussion here: #6438

The idea is to have 3 categories of tuning available for each vehicle type.

Basic:
screen shot 2018-05-18 at 12 42 39 pm
This is the existing Tuning page. It should contains the simple controls to adjust flight characteristics of a standard airframe. Meant for use by non-power users who already have a decent flying vehicle already.

Advanced:
screen shot 2018-05-18 at 12 45 34 pm
This is the place for a designed page of parameter values which relate to power user tweakable tuning params. Used when you are creating a new vehicle from scratch and a standard airframe didn't get you a decent flying vehicle. Or for tweaking a vehicle beyond "decent" flying characteristics.

PID Tuning:
screen shot 2018-05-18 at 12 43 04 pm
This is the existing Advanced PID Tuning.

### DIFF
--- a/src/AutoPilotPlugins/PX4/PX4TuningComponentCopter.qml
+++ b/src/AutoPilotPlugins/PX4/PX4TuningComponentCopter.qml
@@ -28,499 +28,486 @@ SetupPage {
         Column {
             width: availableWidth
 
-            Component.onCompleted: {
-                showAdvanced = !ScreenTools.isMobile
-            }
-
             FactPanelController {
                 id:         controller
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
So for example for a Fixed Wing you have things like maybe min turning radius parameter in the Advanced page. That sort of thing.

DonLakeFlyer:
I'm going to dump this pull since it makes PID Tuning which is power user stuff to up front.

PR #6467, opened 2018-05-19
Special preflight failure handling

DonLakeFlyer:
Replacement for #6436

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -648,7 +648,7 @@ QObject* QGCApplication::_rootQmlObject()
 void QGCApplication::showMessage(const QString& message)
 {
     // PreArm messages are handled by Vehicle and shown in Map
-    if (message.contains("PreArm")) {
+    if (message.startsWith(QStringLiteral("PreArm")) || message.startsWith(QStringLiteral("PREFLIGHT"))) {
         return;
     }
 
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -647,8 +647,8 @@ QObject* QGCApplication::_rootQmlObject()
 
 void QGCApplication::showMessage(const QString& message)
 {
-    // Special case hack for ArduPilot prearm messages. These show up in the center of the map, so no need for popup.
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
As far as severity I think these should be errors. But the problem with that is that it causes QGC to pop error popups for them which is annoying as well. Especially with the special handling. Maybe the error popping code should filter these out as well to not pop and error box. Opinions?

src/QGCApplication.cc
```diff
@@ -647,8 +647,8 @@ QObject* QGCApplication::_rootQmlObject()
 
 void QGCApplication::showMessage(const QString& message)
 {
-    // Special case hack for ArduPilot prearm messages. These show up in the center of the map, so no need for popup.
-    if (message.contains("PreArm:")) {
+    // PreArm messages are handled by Vehicle and shown in Map
+    if (message.contains("PreArm")) {
```
dagar: OR PREFLIGHT FAIL?
At the moment we also have a couple instances of "PREFLIGHT ADVICE".

DonLakeFlyer: Oh, wait I forgot that was there. I'll change that as well. Which in turn means these should be shown with high severity and QGC wille at them.

DonLakeFlyer: And I'll change to PREFLIGHT prefix only.

dagar:
I'm going to test the QGC changes and play with different MAV_SEVERITY levels based on attempted arming or running the check periodically.

dagar:
Questions
 - I'm seeing what text directly on the map. Is it supposed to be in a yellow box?
 - Should we also remove the prefix string (PreArm: or PREFLIGHT FAIL:) before display?
 - On the firmware side is it worth limiting these to a single failure message at a time? Displaying multiple doesn't seem to work very well.

DonLakeFlyer:
> I'm seeing what text directly on the map. Is it supposed to be in a yellow box?

The exact text of the Status Text messaage shows up in a big font in the center of the map.

PR #6479, opened 2018-05-21
Fixed Wing Landing Pattern: Visualize landing area

DonLakeFlyer:
screen shot 2018-05-21 at 12 27 24 pm
Landing area is 100x15 meters


DonLakeFlyer:
@antiheavy FYI

Antiheavy:
I wonder if a dashed line boarded around the landing area would better imply / convey the possibility of landing outside the area?  I’m excited to use this though.

DonLakeFlyer:
Drawing dashed lines on the map will be a royal pain. It's not support natively. So you have to fall back to doing things all by yourself with low level drawing elements. Turns a 10 minute feature into a multi-hour feature.

PR #6494, opened 2018-05-23
Better help for alt mode

DonLakeFlyer:
What each one looks like:
screen shot 2018-05-23 at 3 20 40 pm
screen shot 2018-05-23 at 3 20 47 pm
screen shot 2018-05-23 at 3 20 52 pm
screen shot 2018-05-23 at 3 20 57 pm

### DIFF
--- a/src/MissionManager/SimpleMissionItem.cc
+++ b/src/MissionManager/SimpleMissionItem.cc
@@ -706,13 +706,16 @@ void SimpleMissionItem::_terrainAltChanged(void)
     }
 
     if (qIsNaN(terrainAltitude())) {
+        qDebug() << "1";
         // Set NaNs to signal we are waiting on terrain data
         _missionItem._param7Fact.setRawValue(qQNaN());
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@AndKe FYI...

AndKe:
Great, once you are at it, please mention the difference between source of data, and where it is applied for AGL and terrf (at GCS, incorporated into waypoints, Vs onboard, based on terr messages.)

DonLakeFlyer:
That's going to be end up being way too much text. @hamishwillee Do you have any ideas where. There is small space for text, but not too much.

DonLakeFlyer:
Merging this since it's better. Then we can discuss with Hamish for something even better.

AndKe:
What about something short like:
"Calculated from terrain data" ->  "GCS add Google's altitude data"  (or other vendor, or "internet")
"Using terrain reference frame" -> "Above terrain reference in AP"

In both cases, it's clear where the calculation is done, and it's not very long.

PR #6496, opened 2018-05-24
Pre-Flight Checklist: Coding style cleanup

DonLakeFlyer:
This is 90% coding style. There are two real changes in here:
* Using properties to name states
* A couple places where boolean logic was changed to make it easier to understand

Should be a no-op from a functionality standpoint.

### DIFF
--- a/src/FlightDisplay/CheckList.qml
+++ b/src/FlightDisplay/CheckList.qml
@@ -1,3 +1,12 @@
+/****************************************************************************
+ *
+ *   (c) 2009-2016 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ *
+ * QGroundControl is licensed according to the terms in the file
+ * COPYING.md in the root of the source code directory.
+ *
+ ****************************************************************************/
+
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@philipoe One thing I don't understand is the need for both _state and _nrClicks? It make things kind of confusing. They seem to do the essentially the same thing and seems like _nrClicks could be gotten rid of with code changes.

DonLakeFlyer:
The other thing I want to do is change this names:
* QGCCheckListItem -> PreFlightCheckButton
* CheckList -> PreFlightCheckList

I'l do that in a pull all it's own. Then I think it's to the point where I can start paying around with it for 3rd party customers.

DonLakeFlyer:
Yup. I'll fix. After name changes go through. My next thing is to package things like the battery check list item into a single reusable control. No funtionality change, just packaging. That way a custom build can just override PreFlightCheckList.qml and use the controls simply to create their own custom pre-flight. And then if battery checklist control gets better/new feature they get all that when they pull from master again.

Issue #6451, opened 2018-05-15
Feature Request: Modified Survey Mode


jstrebel:
We regularly fly a number of agricultural crops with a multi-spectral camera. 
(speed approx 13-15m/sec) Spacing approx 20m.
In missions with relatively moderate winds, we encounter the limits of flight plan control on a regular basis, at the turns at the end of the field. In order to reach the trajectory during the return flight with sufficient precision, we must, depending on the wind the mission at the ends by hand. In certain places, and depending on wind force, we can adjust the turnaround distance to solve the problem. However, there are often obstacles that make this impossible. We believe that a modified survey mode would help make it easier and more efficient to use.

To illustrate the flight mission I enclose a picture with the waypoints.
Actual survey mode: UAV flies 1,2,3,4,5 --- to mission end sequentially.
Desired “Modified survey mode”:
UAV flies points: 1,2,3, 180°radius turn to 30, heading toward 31, 180°radius turn to 7,
heading toward 6, 180°radius turn to 35, heading toward 34, 180°radius turn to 10, and so on.

This feature would be extremely helpful for Fixed Wing UAV’s because of their turn radius
is determined by the type of the Airplane.

![buenzen_apr-27](https://user-images.githubusercontent.com/2244311/40053055-20483986-5840-11e8-990f-23208733b79d.jpg)

DonLakeFlyer:
> We believe that a modified survey mode would help make it easier and more efficient to use.

Modified how? The plan is to allow for surveys to fly alternate lines. NYI, but coming.

DonLakeFlyer:
Sorry just seeing your example, looking...

Kjkinney:
This is similar to this issue https://github.com/mavlink/qgroundcontrol/issues/5866 also posted here. @Antiheavy what are your thoughts? I feel like this would be incredibly beneficial for fixed wings.

DonLakeFlyer:
The issue with skipping an arbitrary number of transects (skip determined by turnig radius) is what you do when you get to the end and have to go back and do the ones you skipping. That turnaround gets complicated if you are trying to fly shortest path.

Antiheavy:
Zamboni mode!  +1 for this.  I like the idea of a simple "fly alternating lines" checkbox.  You're right, it usually gets inefficient if you try to skip N lines at a time.

DonLakeFlyer:
So do I because it's easy. But it won't work if your transects are still too close to each other.

sanderux:
[...]

PR #6497, opened 2018-05-24
Survey: Support for flying alternate transects (fixed wing only)

DonLakeFlyer:
Related to #6451. Trivial to implement with new transect code. Two things:
* This doesn't try to do anything fance with the turaround after the first pass.
* The Plan visuals and not super useful will all the overlapping flight paths. I'll get around to doing something there at some point.


DonLakeFlyer:
And I forgot to add this to Corridor Scan as well.

sanderux:
For a corridor scan the turnaround distance can be set much larger

DonLakeFlyer:
Maybe I'll leave corridor scan alone for now then. Less is more.

sanderux:
Hi Don,
I cant find this feature in the current daily build appimage for linux. Is that correct?

sanderux:
confirmed

PR #6502, opened 2018-05-24
Track offline editing settings only for offline editing vehicle

okalachev:
When using QGC in multi-vehicle mode, there is a problem, that vehicles' firmware plugins are changed to base `FirmwarePlugin`. It happens when [this](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/PlanMasterController.cc#L115) code is executed. So, instead of normal modes widget, after switching between vehicles we see something like this:

Haven't figured out, why this bug is not happening in single-vehicle mode, but this change fixes it. Actually, I think, we don't want to subscribe to offline editing settings changes, if we are not in offline editing vehicle.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -413,11 +413,13 @@ void Vehicle::_commonInit(void)
     connect(_rallyPointManager, &RallyPointManager::error,          this, &Vehicle::_rallyPointManagerError);
     connect(_rallyPointManager, &RallyPointManager::loadComplete,   this, &Vehicle::_rallyPointLoadComplete);
 
-    // Offline editing vehicle tracks settings changes for offline editing settings
-    connect(_settingsManager->appSettings()->offlineEditingFirmwareType(),  &Fact::rawValueChanged, this, &Vehicle::_offlineFirmwareTypeSettingChanged);
-    connect(_settingsManager->appSettings()->offlineEditingVehicleType(),   &Fact::rawValueChanged, this, &Vehicle::_offlineVehicleTypeSettingChanged);
-    connect(_settingsManager->appSettings()->offlineEditingCruiseSpeed(),   &Fact::rawValueChanged, this, &Vehicle::_offlineCruiseSpeedSettingChanged);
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
This change is wrong. The fact that the offline editing vehicle type information tracks the last vehicle you connected to is correct and should not change. That is important since offline editing defaults to PX4 Firmware. If you are an ArduPilot user and the first time you use QGC and connect it to ArduPilot. You then don't want to have to tell QGC when planning a mission offline that you are ArduPilot. This makes that all work seamlessly. The bug is somewhere else.

If you can put a set of specific steps to reproduce in an issue I can likely find it quickly and fix.

okalachev:
Actually, we're experimenting with custom firmware plugins (a plugin for a charging station for a drone). I want to see the PX4 drone vehicle and the charging station "vehicle" simultaneously. 

With this patch I do, but without it, when [this](https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L493) code is executied on the first vehicle (e. g. PX4), its plugin is changed to `FirmwarePlugin`. Then, after I switch vehicles, the mode toolbar is broken.

DonLakeFlyer:
Something is wrong with your plugin implementation then. QGC works fine when running against a PX4 vehicle and an ArduPilot vehicle at the same time.

okalachev:
https://github.com/CopterExpress/qgroundcontrol/blob/charging_station/src/FirmwarePlugin/ChargingStation/ChargingStationFirmwarePlugin.h
https://github.com/CopterExpress/qgroundcontrol/blob/charging_station/src/FirmwarePlugin/ChargingStation/ChargingStationFirmwarePlugin.cc
https://github.com/CopterExpress/qgroundcontrol/blob/charging_station/src/FirmwarePlugin/ChargingStation/ChargingStationFirmwarePluginFactory.h
https://github.com/CopterExpress/qgroundcontrol/blob/charging_station/src/FirmwarePlugin/ChargingStation/ChargingStationFirmwarePluginFactory.cc

Here is the code of my plugin. Can you check it please?

PR #6504, opened 2018-05-25
Simplify PreFlightCheckButton implementation

DonLakeFlyer:
* New implementation makes it easier to implement check buttons
* Changed visuals such that they are readable outdoors on a tablet
* Checklist is only available when you have a vehicle connected
* Should be a mostly no-op functionality wise except for manual checks only support single step now.

screen shot 2018-05-25 at 10 07 19 am

Known work left:
* I've tested but I'm sure I still broke something. I'll be continuing on this right away. So I'll find anything as I go along.
* Move the color usage into QGCPalette
* Support double manual steps


DonLakeFlyer:
Also I added the ability to do a sat count check to Sensors button.

PR #6506, opened 2018-05-25
Cleanup layout issues throughout QGC

dogmaphobic:
There were several layout errors throughout the program that went unnoticed until I built it with Qt 5.11. 

In summary: Don't use `anchor` anything within a layout control.

@DonLakeFlyer I didn't test 5.10 but 5.11 is working fine on macOS, iOS, Android and Windows (I have not yet tested on Linux).

### DIFF
--- a/src/AnalyzeView/LogDownloadPage.qml
+++ b/src/AnalyzeView/LogDownloadPage.qml
@@ -51,8 +51,7 @@ AnalyzePage {
 
             TableView {
                 id: tableView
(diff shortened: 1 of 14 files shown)


DonLakeFlyer:
Yeah, looks like good complaints that it makes now.

dogmaphobic:
Tested on Linux as well. As far as I can tell, we're ready to do a wholesale switch to 5.11.

PR #6513, opened 2018-05-26
Transect alt mode fixes

DonLakeFlyer:
* When turning off Terrain Follow the item should change to Relative Altitude
* When going from a manual camera to a custom camera should change to Relative Altitude
* Added support for custom build hiding of Relative Altitude checkbox in Survey and Corridor Scan
* Wrote unit tests to verify all the above so it never breaks again


Antiheavy:
> When turning off Terrain Follow the item should change to Relative Altitude

seems to work.




> When going from a manual camera to a custom camera should change to Relative Altitude

this one is confusing....

DonLakeFlyer:
When you specify a camera the calculations and corresponding mission are always generated using relative alts. So it switches back to relative (and disables the alt mode checkbox). For manual you can do it however you want since you are specifying the altitude yourself.

PR #6518, opened 2018-05-28
Do not send virtual joystick commmands over high latency links

acfloria:
This can block a high latency link. In addition due to the latency it does not make sense to send control commands anyway over that link.

### DIFF
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -484,12 +484,12 @@ QGCView {
             z:                          _panel.z + 5
             width:                      parent.width  - (_flightVideoPipControl.width / 2)
             height:                     Math.min(ScreenTools.availableHeight * 0.25, ScreenTools.defaultFontPixelWidth * 16)
-            visible:                    (_virtualJoystick ? _virtualJoystick.value : false) && !QGroundControl.videoManager.fullScreen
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
I wonder about adding checks for high latency for these things. You have the same problem with a regular joystick. A high latency link is a pretty high end usage pattern that I would think the user kind of needs to know what they are doing to some extent. I"m worried about adding a lot more checks all over the place which could lead to bugginess in mainline usage.

acfloria:
I partially agree, only an experienced user will operate Satcom. Nevertheless I think some checks are good because those actions could really block the link. Those are the mission upload, parameter synch and the virtual joystick. For the first two we have checks, for the last one not, that is why I wanted to fix it.

With regards to the regular joystick we don't use one. But if someone is flying using SatCom and the regular Joystick then transmitting the joystick commands must be stopped or the link will be blocked by the joystick commands.

acfloria:
@DonLakeFlyer any decision/update from your side?

DonLakeFlyer:
I think what would be better here would be to also hide the virtual sticks in this case. That way the user at least wouldn't expect them to do something. They may be confused as to why they aren't visible but I think that's ok for this case.

acfloria:
@DonLakeFlyer I implemented your suggestions, can you have a look at them?

acfloria:
I missed to check if the active vehicle exists in FlightDisplayView.qml, this is now fixed.

PR #6521, opened 2018-05-28
Add settingsGroup qualifer to settings override

DonLakeFlyer:
The api for QGCCorePlugin::adjustSettingsMetaData which incorrect in that it only take the meta data as a parameter. In order to fully qualify which settings this is about you also need the settings group name. Without it there can be namespace collisions on the name which currently exist between Survey and Corridor Scan settings.

Note that this is a breaking change to an API for custom builds. Easy to deal with but all the less breaking.

### DIFF
--- a/src/FactSystem/Fact.cc
+++ b/src/FactSystem/Fact.cc
@@ -62,8 +62,6 @@ Fact::Fact(FactMetaData* metaData, QObject* parent)
     , _deferredValueChangeSignal(false)
     , _valueSliderModel         (NULL)
 {
-    // Allow core plugin a chance to override the default value
-    qgcApp()->toolbox()->corePlugin()->adjustSettingMetaData(*metaData);
     setMetaData(metaData, true /* setDefaultFromMetaData */);
 }
 
--- a/src/FactSystem/SettingsFact.cc
+++ b/src/FactSystem/SettingsFact.cc
@@ -20,19 +20,19 @@ SettingsFact::SettingsFact(QObject* parent)
     QQmlEngine::setObjectOwnership(this, QQmlEngine::CppOwnership);
 }
 
-SettingsFact::SettingsFact(QString settingGroup, FactMetaData* metaData, QObject* parent)
-    : Fact(0, metaData->name(), metaData->type(), parent)
-    , _settingGroup(settingGroup)
-    , _visible(true)
+SettingsFact::SettingsFact(QString settingsGroup, FactMetaData* metaData, QObject* parent)
+    : Fact          (0, metaData->name(), metaData->type(), parent)
+    , _settingsGroup(settingsGroup)
+    , _visible      (true)
 {
     QSettings settings;
 
-    if (!_settingGroup.isEmpty()) {
-        settings.beginGroup(_settingGroup);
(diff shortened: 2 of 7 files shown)


DonLakeFlyer:
@dogmaphobic Can you take a look at this?

src/FactSystem/Fact.cc
```diff
@@ -62,8 +62,6 @@ Fact::Fact(FactMetaData* metaData, QObject* parent)
     , _deferredValueChangeSignal(false)
     , _valueSliderModel         (NULL)
 {
-    // Allow core plugin a chance to override the default value
-    qgcApp()->toolbox()->corePlugin()->adjustSettingMetaData(*metaData);
```
DonLakeFlyer: @dogmaphobic I also noticed this which is somewhat incorrect. In that the adjustSettingsMetaData was for settings not all Facts. At this point there is no settings group qualifier and now way to deal with name collisions. As far as I know nobody uses this ability. Only the settings overrides are used in current custom build customers.

dogmaphobic: I'm now getting to this. Has `adjustSettingMetaData()` been removed altogether? This is used to both hide and set the (default) values of facts in custom builds. For example, if I don't want virtual joysticks, I would:

```
if (metaData.name() == AppSettings::virtualJoystickName) {
    metaData.setRawDefaultValue(false);
    return false;
}
```

That sets the Fact default to `false` and tells QGC to not show the option under settings (retuning false).

src/FactSystem/SettingsFact.cc
```diff
@@ -20,19 +20,19 @@ SettingsFact::SettingsFact(QObject* parent)
...
 
-    if (!_settingGroup.isEmpty()) {
-        settings.beginGroup(_settingGroup);
+    if (!_settingsGroup.isEmpty()) {
+        settings.beginGroup(_settingsGroup);
     }
 
     // Allow core plugin a chance to override the default value
-    _visible = qgcApp()->toolbox()->corePlugin()->adjustSettingMetaData(*metaData);
+    _visible = qgcApp()->toolbox()->corePlugin()->adjustSettingMetaData(settingsGroup, *metaData);
```
dogmaphobic: Ok, so here is where it's actually called. Right?

dogmaphobic: Ok, it's all working as it should. I've updated some random "custom" build and it's all working:

```
if (settingsGroup == AppSettings::settingsGroup && metaData.name() == AppSettings::virtualJoystickName) {
    metaData.setRawDefaultValue(false);
    return false;
}
```

DonLakeFlyer:
The tricky thing is that multiple of the SettingsGroups have SettingsGroup::settingsGroup == "". Which means in the custom build adjustMetaData implementation you can't if/else on the settingsGroup comparision. You have to let it fall through to another settingsGroup == Foo::settingsGroup test since there could be multiple. Ends up strange, but I couldn't figure out any other way to make it work. In reality a SettingsGroup should always have a QSettings::beginGroup hierarchy for it's settings. But unfortunately the SettingsGroup stuff was create after a bunch of settings were already store in the root settings tree.

PR #6522, opened 2018-05-29
Survey: Allow VTOL vehicles to fly alternat transects

DonLakeFlyer:
Also fixed bug in Corridor Scan time between shots calculation.


DonLakeFlyer:
@sanderux FYI

sanderux:
Cool, ready to merge? I will also need to update our client that is using corridor scan, they have just started

PR #6530, opened 2018-05-30
Use a menu to delete a vertex

dogmaphobic:
When editing a survey polygon, clicking on a vertex would delete that vertex. It was too easy to click/touch it by mistake. This changes it to a double click for deleting a vertex.

In addition:

For corridor scans, the same double click applies when you want to delete a vertex. When you single click however, you get the menu asking if you want to delete that particular vertex or load a KML file. Also, the "Delete Vertex" menu item only shows if there are more than 2 vertices.

Also fixed a layout/anchor issue within corridor scan.

### DIFF
--- a/src/FlightMap/MapItems/MissionItemIndicatorDrag.qml
+++ b/src/FlightMap/MapItems/MissionItemIndicatorDrag.qml
@@ -29,7 +29,6 @@ Rectangle {
     property var itemCoordinate ///< Coordinate we are updating during drag
 
     signal clicked
-    signal doubleClicked
     signal dragStart
     signal dragStop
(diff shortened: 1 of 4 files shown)


dogmaphobic:
Switching gears. The distinction of tap/double tap on mobile isn't that great. Half the time it takes it as a single click. Instead, I've added a context menu to all vertices (and the center mark for polygons) where the first option is always an option to delete the selected vertex (not shown if you cannot delete it.)

screen shot 2018-05-30 at 10 38 58 pm

Antiheavy:
@dogmaphobic nice improvement.

PR #6532, opened 2018-05-30
jenkins: agent fix

lamping7:
Currently an overall pipeline agent is assigned to the whole build, essentially blocking the assigned slave from performing any build steps as the individual stages define the agent to use. The blocking prevents it from doing other jobs in the queue until the whole QGC build is done. This fixes that, and adds small formatting fixes.

The images below shows how slaves are blocked when there are other tasks to do in the build queue.
screenshot from 2018-05-29 18-55-33

screenshot from 2018-05-29 18-59-33

FYI @dagar


lamping7:
FYI: not being a previous QGC contributor, these changes aren't being tested per the security settings we have in Jenkins. However, the risk associated with these changes are very minimal.

dagar:
Thanks!

PR #6539, opened 2018-06-01
Support for Sync - Load KML

DonLakeFlyer:
This makes the user model for creating complex mission items from KML files much easier:
screen shot 2018-06-01 at 11 07 24 am

If needed you are prompted for which type of item you want to create from the KML:
screen shot 2018-06-01 at 11 07 45 am

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -515,6 +515,7 @@ HEADERS += \
     src/Joystick/Joystick.h \
     src/Joystick/JoystickManager.h \
     src/JsonHelper.h \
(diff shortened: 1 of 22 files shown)


Antiheavy:
will this append a complex mission item to the end of the existing on-screen mission, or will it blow away an existing on-screen mission?  I'm thinking of the case where there is a Takeoff and Landing Pattern required for fixed wings.

DonLakeFlyer:
Append, just like selecting from the Pattern button.

Antiheavy:
> Append, just like selecting from the Pattern button.

sounds good.

PR #6546, opened 2018-06-03
Remove QtMultimedia import when not needed

DonLakeFlyer:
@dogmaphobic I found these other two references to QtMultimedia which don't seem to be needed.


dogmaphobic:
QtMultimedia is used for UVC video.

PR #6547, opened 2018-06-03
Add missing libraries to Linux build

dogmaphobic:
Also, add test for an existing window before attempting to close it. This is for when the main window isn't created because of some error and when you attempt to close it, the code tries to call a function within it. As it's not there, the application can never exit.


dogmaphobic:
Looks like Jenkins is still using Qt 5.9.2 and 5.9.3.

PR #6555, opened 2018-06-04
Even more PreFlight reorg/support

DonLakeFlyer:
* Sensors check is now just health check including AHRS
* GPS button is separate check which supports turning on/off 3d lock and sat count checks. Allows user override of low sat count. Default preflight is set to require 3d lock (user cannot click past) and shows warnings for sat count < 10 which user can click past.
* Battery button support override of low percentage check. Default preflight has low percent set to 40 and user cannot click past.
* Reworked the way user overrides work on PreFlightCheckButton
* Added slight delay to completed group closing so use can see all things green for a moment prior to next group opening.


dogmaphobic:
We need a bit better coordination regarding this whole PreFlight check list so it doesn't clash with the ongoing work with Airmap. Though they provide different services (Airmap concerns rules and regulations and this new _PreFlight Checks_ concerns with airworthiness), we may not want two of these.

screen shot 2018-06-04 at 4 30 36 pm

DonLakeFlyer:
At some point. But I'm not even sure it's worth that in the first release. I think waiting to see where either of these end up would be a good idea. Otherwise we'll likely just grind to a halt trying to merge two things which will likely change out from under us.

PR #6580, opened 2018-06-08
Fix for vehicle widget bug

### LINKED ISSUE #6579: FactGroupName vehicle translatable text problem
If localization used FactGroupName translatable string cause widget broken because of special characters on translation because of configuration file stores characters badly.
`property string factGroupName: qsTr("Vehicle")`

Other factgroups
`const char* Vehicle::_gpsFactGroupName =            "gps";
const char* Vehicle::_battery1FactGroupName =       "battery";
const char* Vehicle::_battery2FactGroupName =       "battery2";`

feemi:
Fixes #6579


dogmaphobic:
That's correct. That must have been "localized" by mistake.

PR #6586, opened 2018-06-14
Split constructor of HeartbeatTimer up in two functions

acfloria:
Previously the first activeChanged signal got lost because it was not connected yet to the LinkInterface which resulted in a link incorrectly shown as inactive.


DonLakeFlyer:
It seems like ```init``` would be better named ```start``` to make thing more clear?

acfloria:
I named it `init` as this function should be called only once. In addition it is not similar to the `start()` function of a timer and I wanted to make that clear with that name.

But if you want me to rename it I can do it.

DonLakeFlyer:
You can leave it.

PR #6590, opened 2018-06-15
Windows OpenSSL: Use libs which static link to CRT

DonLakeFlyer:
* Previous libs had dynamic link to specific version of CRT which may or may not be on users machine.
* Added better logging to debug SSL problems
* Also added code to ignore SSL errors


DonLakeFlyer:
@birchera I change the online airmap query code to ignore SSL cert errors as well as have better logging of ssl errors. This makes debugging strange SSL errors like the one fixed here much easier. I tried finding a similar place in the offline code to do the same thing but couldn't. Could you do the same SSL ignore and logging in that code?

PR #6595, opened 2018-06-15
qgroundcontrol: Use qtquickcompiler by default

patrickelectric:


DonLakeFlyer:
Ah nice. Didn't realize 5.11 has open source compiler.

DonLakeFlyer:
One thiing I noticed with using this is that QTCreator doesn't check the "Enable Qt Quick Compiler" option in Build Steps. Does that still mean it is really using the compiler?

DonLakeFlyer:
Duh: Just clicked that button and it just added CONFIG+=qtquickcompiler. Guess QtCreator isn't smart enough to pick it up from the build files.

PR #6606, opened 2018-06-19
Turn "Sync" menu into a proper "File" menu

dogmaphobic:
First stab at reworking the _Sync_ menu:

The _Sync_ menu within the Plan View tool strip is basically a regular _File_ menu. As such, let's call it what it is.

* Renamed _Sync_ to _File_
* Moved it to the top of the list as this is the standard for _File_ menus
* Renamed _Remove All_ to _New_ (and moved it to the top of the list as this is the standard)
* Renamed _Load_ to _Open_
* Added a _Save_ (current) and renamed the existing one to _Save As_ (to new file)
* Hide both _Download_ and _Upload_ buttons if you are not connected (offline)
* Disable the _New_ button if the plan is empty
* Disable save options if there is nothing to save

In addition, I've adjusted the width of the tool strip so long words such as _Waypoint_ or _Checklist_ would fit within it.

Blank mission and not connected to anything:
screen shot 2018-06-19 at 11 46 44 am

Selecting _New_ with a plan loaded/created:
screen shot 2018-06-19 at 10 02 23 am

Connected to a vehicle:
screen shot 2018-06-19 at 10 04 10 am

Android for reference:
screen

### DIFF
--- a/src/MissionManager/PlanMasterController.cc
+++ b/src/MissionManager/PlanMasterController.cc
@@ -280,6 +280,7 @@ void PlanMasterController::loadFromFile(const QString& filename)
         return;
     }
 
+    QFileInfo fileInfo(filename);
     QFile file(filename);
 
     if (!file.open(QIODevice::ReadOnly | QIODevice::Text)) {
@@ -288,8 +289,8 @@ void PlanMasterController::loadFromFile(const QString& filename)
         return;
     }
 
-    QString fileExtension(".%1");
-    if (filename.endsWith(fileExtension.arg(AppSettings::planFileExtension))) {
+    bool success = false;
+    if(fileInfo.suffix() == AppSettings::planFileExtension) {
         QJsonDocument   jsonDoc;
         QByteArray      bytes = file.readAll();
 
@@ -319,17 +320,31 @@ void PlanMasterController::loadFromFile(const QString& filename)
                 !_geoFenceController.load(json[_jsonGeoFenceObjectKey].toObject(), errorString) ||
                 !_rallyPointController.load(json[_jsonRallyPointsObjectKey].toObject(), errorString)) {
             qgcApp()->showMessage(errorMessage.arg(errorString));
+        } else {
+            success = true;
         }
-    } else if (filename.endsWith(fileExtension.arg(AppSettings::missionFileExtension))) {
+    } else if (fileInfo.suffix() == AppSettings::missionFileExtension) {
         if (!_missionController.loadJsonFile(file, errorString)) {
             qgcApp()->showMessage(errorMessage.arg(errorString));
+        } else {
+            success = true;
         }
-    } else if (filename.endsWith(fileExtension.arg(AppSettings::waypointsFileExtension)) ||
-               filename.endsWith(fileExtension.arg(QStringLiteral("txt")))) {
+    } else if (fileInfo.suffix() == AppSettings::waypointsFileExtension || fileInfo.suffix() == QStringLiteral("txt")) {
         if (!_missionController.loadTextFile(file, errorString)) {
             qgcApp()->showMessage(errorMessage.arg(errorString));
+        } else {
+            success = true;
         }
+    } else {
+        //-- TODO: What then?
     }
+
+    if(success){
+        _currentPlanFile.sprintf("%s/%s.%s", fileInfo.path().toLocal8Bit().data(), fileInfo.completeBaseName().toLocal8Bit().data(), AppSettings::planFileExtension);
+    } else {
+        _currentPlanFile.clear();
+    }
+    emit currentPlanFileChanged();
 
     if (!offline()) {
         setDirty(true);
@@ -352,6 +367,14 @@ QJsonDocument PlanMasterController::saveToJson()
     return QJsonDocument(planJson);
 }
 
+void
+PlanMasterController::saveToCurrent()
+{
+    if(!_currentPlanFile.isEmpty()) {
+        saveToFile(_currentPlanFile);
+    }
+}
+
 void PlanMasterController::saveToFile(const QString& filename)
 {
     if (filename.isEmpty()) {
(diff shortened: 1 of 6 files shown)


dogmaphobic:
@DonLakeFlyer When I load a plan file, why isn't it editable?

DonLakeFlyer:
The problem is that the Sync menu isn't really just a "File" menu. It is a file menu and a sync menu. It may make sense to separate the two functions into separate buttons? 

That problem shows up the most with getting rid of "Remove All" and replacing it with New. In reality you want both of these. The reason is that there needs to be a simple mechanism to clear a plan fully from the vehicle. This is what "Remove All" does. You can't really replace it with "New".

DonLakeFlyer:
Another thing would be whether Plan and KML open/save should be separate buttons. I still think so. But these could be collapsed so that you picked the file type in the file dialog, but that would be harder to do with thee mobile dialogs.

DonLakeFlyer:
> When I load a plan file, why isn't it editable?

No idea. Is that happening on master?

DonLakeFlyer:
> It may make sense to separate the two functions into separate buttons?

Or maybe break the File menu drop panel into two separate sections. The file part and the sync part.

DonLakeFlyer:
Also just to be clearer on the difference between New and Remove All. Remove All is basically a "New" which begins creation of an new Plan, and also clears the Plan from the Vehicle. Whereas I would expect New to just begin creation of a Plan and not affect the vehicle. That is if it follows normal file menu semantics.

dogmaphobic:
> The reason is that there needs to be a simple mechanism to clear a plan fully from the vehicle. This is what "Remove All" does. You can't really replace it with "New".

I think it makes more sense to have a specific button to reset/clear a mission within the vehicle. It's more obvious than "creating a blank mission" and "uploading it to the vehicle", which is the current model.

>Or maybe break the File menu drop panel into two separate sections. The file part and the sync part.

With an extra button to clear a mission from the vehicle, that may not be necessary. These buttons (Download/Upload/Reset) would only show up if connected. I think they still fit within the context of a file menu.

>Another thing would be whether Plan and KML open/save should be separate buttons. I still think so. 

I thought about that but the complexity of dealing with it for mobile wasn't worth the effort. At least not in this first shot.

DonLakeFlyer:
@antiheavy FYI. Gus is working to make the Plan view more "normal" with respect to file management.

dogmaphobic:
> No idea. Is that happening on master?

Yes, I had never done that so I didn't know if it was by design or not. When I load a plan, there are no editable "dots".

screen shot 2018-06-19 at 1 08 09 pm

(this is my current Airmap branch, which is synced to master but does not include the changes in this PR)

dogmaphobic:
For clearing a mission from the vehicle. Something like this? (I haven't put too much thought on the layout and wording yet)

screen shot 2018-06-19 at 1 27 08 pm

DonLakeFlyer:
> For clearing a mission from the vehicle. Something like this?

Sure.

Also for disable versus invisible. I tend towards disable since it makes the UI stable. It also gives people a hint as to what is possible to do. Or tehy may be confused when the buttons are not there.

dogmaphobic:
> Also for disable versus invisible.

The idea was to maintain the separation of offline versus connected and active. It follows the rest of the UI, which transforms itself when you have a vehicle connected. Buttons that are always present in all states are never hidden, but disabled instead (Save, Save As, New, etc.)

DonLakeFlyer:
> It follows the rest of the UI, which transforms itself when you have a vehicle connected.

Some does, some doesn't. All the DropPanel implementations follow the disable when no vehicle model. Looks at things like Center tool and so forth.

Antiheavy:
> I tend towards disable since it makes the UI stable. It also gives people a hint as to what is possible to do. Or tehy may be confused when the buttons are not there.

I came here to say this.  Greyed out is better than disappearing.

dogmaphobic:
Ok, new iteration:

* Create new, specific button for clearing mission from vehicle
* Adjust sizing of DropPanel

@DonLakeFlyer, the _Clear Vehicle Mission_ button calls `masterController.removeAllFromVehicle()`, which also clears the mission within the plan view. How do I only reset the autopilot but leave the mission in plan view alone? Also, is there a way to tell if there is a mission currently loaded in the autopilot? That way, the button is only enabled if that's the case. Otherwise it should be disabled.

screen shot 2018-06-20 at 9 00 18 am

DonLakeFlyer:
> the Clear Vehicle Mission button calls masterController.removeAllFromVehicle(), which also clears the mission within the plan view. How do I only reset the autopilot but leave the mission in plan view alone?

PlanMasterController::removeAll

> Also, is there a way to tell if there is a mission currently loaded in the autopilot? That way, the button is only enabled if that's the case. Otherwise it should be disabled.

No current way to determine that from the Plan view.

dogmaphobic:
> PlanMasterController::removeAll

That, like the others, clears everything. I couldn't find one that simply cleared the vehicle and left the mission within the Plan View in place.

dogmaphobic:
I phrased the confirmation dialog in such a way that it will tell you both the existing mission and the vehicle will be cleared.

screen shot 2018-06-20 at 3 10 53 pm

DonLakeFlyer:
> That, like the others, clears everything. I couldn't find one that simply cleared the vehicle and left the mission within the Plan View in place.

Could be added easily.

dogmaphobic:
> Could be added easily.

For you! :P

Antiheavy:
> Also, is there a way to tell if there is a mission currently loaded in the autopilot?

@dagar remember our discussions about need for mission sync status communication between firmware and ground station?  another example of why.

DonLakeFlyer:
> remember our discussions about need for mission sync status communication between firmware and ground station? another example of why.

This is a bit different. This is just a missing API in QGC. The Fly View knows if there is a mission on the Vehicle but the Plan View does not.

dogmaphobic:
I'm assuming this is good as it is. The only pending issue was the ability to clear the mission on a vehicle without affecting an existing mission within the Plan View. I could not think of a scenario where that would be important. If something comes up, we can tweak it in further PRs.

PR #6610, opened 2018-06-20
Draggable radius adjustment for Orbit

DonLakeFlyer:
screen shot 2018-06-19 at 6 09 22 pm


Antiheavy:
I realize this is probably being designed for multicopter, but it would be intesting to see how it could have similar use with fixed wing Pause/Hold mode and also fixed wing loiter type mission waypoints.

DonLakeFlyer:
Yeah, doable. Add an Issue

PR #6612, opened 2018-06-20
More random corrections around the code

patrickelectric:
I'm not sure if I _can_ update qwt related code.

### DIFF
--- a/src/AnalyzeView/ULogParser.cc
+++ b/src/AnalyzeView/ULogParser.cc
@@ -57,7 +57,7 @@ QString ULogParser::extractArraySize(QString &typeNameFull, int &arraySize)
         return typeNameFull;
     }
 
-    arraySize = typeNameFull.mid(startPos + 1, endPos - startPos - 1).toInt();
+    arraySize = typeNameFull.midRef(startPos + 1, endPos - startPos - 1).toInt();
     return typeNameFull.mid(0, startPos);
 }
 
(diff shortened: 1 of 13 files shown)


DonLakeFlyer:
FYI: qwt is an external libary which we shouldn't change.

DonLakeFlyer:
Are changes from QLatin1String to QStringLiteral always safe? If so, why?

These sorts of fixups are great, but hard to verify nothing was broken.

patrickelectric:
Hi Don, I'll remove the qwt changes ASAP.
Yes, QLatin1String to QStringLiteral is always safe :)

Here's a summary of usages of QLatin1String that allocate memory:

1. `QString s = "foo"; // Allocates, use QStringLiteral("foo") instead`

2. `QString s = QLatin1String("foo"); // Allocates, use QStringLiteral("foo") instead`

    2.1 `QString s = QLatin1String(""); // No allocation. QString is optimized for this case, so it's safe for empty literals`

3. `QString s = QStringLiteral("foo"); // No allocation`

4. `QString s = QString::fromLatin1("foo"); // Allocates, use QStringLiteral`

5. `QString s = QString::fromUtf8("foo"); // Allocates, use QStringLiteral`

6. `s == "foo" // Allocates, use QLatin1String`

7. `s == QLatin1String("foo) // No allocation`

8. `s == QStringLiteral("foo") // No allocation`

9. `QString {"append", "compare", "endsWith", "startsWith", "indexOf", "insert",`
   `         "lastIndexOf", "prepend", "replace", "contains" } // They all have QLatin1String overloads, so passing a QLatin1String is ok.`

10. `QString::fromLatin1("foo %1").arg(bar) // Allocates twice, replacing with QStringLiteral makes it allocate only once.`

patrickelectric:
@DonLakeFlyer  done.

patrickelectric:
IC build fail with: `java.io.IOException: Unexpected termination of the channel`
lol

PR #6613, opened 2018-06-20
Goto Location uses COMMAND_INT

DonLakeFlyer:
* Fix for #6041
* Fix COMMAND_LONG sending


DonLakeFlyer:
These are no longer shown

Issue #6445, opened 2018-05-13
Planner: mission time/distance ignore hold time, and loops.


AndKe:
If a waypoint has a hold-time specified, this does not affect the flight time.
also:
if waypoints are repeated (jump-to-item x times)  , this does not affect flight time or distance.

DonLakeFlyer:
Nope hold time never worked. There is a ton of missing time calculation if you start to use things beyond the basic. I think it's more likely to have to fail time calc as oppposed to figuring out how to calc for every crazy mission item possibility.

AndKe:
It does not need to be that way :)  QGC could loop thru a missions items that contain navigation & flow and actually sum them up.  - but yes, that's more work.   
Maybe the presented data could at least display warning when encountering a mission item it can't handle, or is known to brake it.

DonLakeFlyer:
Nothing needs to be that way given infinite coding time :)

DonLakeFlyer:
@dagar This is another example where it would be awesome for the firmware code (or something) to be somehow responsible for calcing flight times. Since that code is the only thing which really nows flight characteritics.

DonLakeFlyer:
Similar (sideways sort of) to our landing pattern distance calc thing.

dagar:
Good to know. Should we keep a list somewhere?

DonLakeFlyer:
Hold/Delay work now. Jump doesn't.

AndKe:
great, that's good news for payload that require hovering, it simplifies such missions.

IeiuniumLux:
@DonLakeFlyer, what do you mean with Jump doesn't work?  It's broken in QGC v3.4.2 or in the PX4 firmware v1.8.0?

DonLakeFlyer:
If you have a jump in your mission QGC will not calculate the time/distance correctly.

Issue #5334, opened 2017-06-23
Fix flight time estimation triggering


DonLakeFlyer:
Hold time for waypoint and delay command is supported but does not trigger recalc when mission item values change.

PR #6615, opened 2018-06-20
Plan: Fix additionalTimeDelay calcs

DonLakeFlyer:
Realized there was code for hold time and delay calcs which was just not working.

Related to #6445


DonLakeFlyer:
Fix for #5334

Issue #6266, opened 2018-03-25
Text-to-speech not working in latest stable and nightly in MacOS


anassinator:
This works perfectly fine with v3.2.7 on the same system (macOS 10.12.6). Linux seems to work.

Steps to reproduce:
1. Start QGC and make sure the "Mute all audio output" option is unticked
2. Connect to the telemetry kit
3. Send any `STATUSTEXT` message from the FCU with a high severity (for example by attempting to arm without pressing the enable switch)
4. Receive the `STATUSTEXT` message in a yellow popup, but notice that no voice is spoken.

Board: Pixhawk Mini
Firmware: PX4 v1.7.3

DonLakeFlyer:
@dogmaphobic Can you look?

dogmaphobic:
I could not reproduce this. TTS is working just fine here (master, Qt 5.9.3, macOS 10.13.3)

anassinator:
Are there some logs that I can produce that could help?

DonLakeFlyer:
Downloaded daily builds are not working for me either. @dogmaphobic Can you help with this? Likely builds are being built incorrectly somehow.

dogmaphobic:
@DonLakeFlyer The Qt blob (on S3) must be messed up. If I use it to run qmake, look at the last message:

```
~/build$ ./Qt5.11-clang_64/5.11.0/clang_64/bin/qmake -r ~/github/work/UpstreamQGC/qgroundcontrol.pro CONFIG+=installer -spec macx-clang
Info: creating stash file /Users/gus/build/.qmake.stash
Project MESSAGE: Qt version 5.11.0
Project MESSAGE: Mac build
Project MESSAGE: QGroundControl version 3.3.2 build 742 describe Development master:4dc7b9d64 2018-06-20 10:59:36 -0700
Project MESSAGE: Build Installer
Project MESSAGE: Release flavor
Project MESSAGE: Using Default QtLocation headers
Project MESSAGE: Using MAVLink dialect 'ardupilotmega'.
Project MESSAGE: Including support for Zeroconf (Bonjour)
Project MESSAGE: Including support for video streaming
Project MESSAGE: Preparing GStreamer Framework
Project MESSAGE: This project is using private headers and will therefore be tied to this specific Qt module build version.
Project MESSAGE: Running this project against other versions of the Qt modules may crash at any arbitrary point.
Project MESSAGE: This is not a bug, but a result of using Qt internals. You have been warned!
[...]

PR #6618, opened 2018-06-20
Remove duplicate travis install

DonLakeFlyer:


DonLakeFlyer:
Related to discussion in #6266

Issue #6625, opened 2018-06-21
Fly view:  Map click, and rightclick does pan map (guided defunct)


AndKe:
Guided is impossible to use (current master) - any left/right click pan map in a .. inverse way.
https://youtu.be/RPfuwqN4blk

DonLakeFlyer:
Should be fixed now

PR #6635, opened 2018-06-22
Plan map drag indicator fix

feemi:
Fix for #6625


DonLakeFlyer:
This isn't the correct fix. The correct fix is to find the places which use this control and pass in the mapControl property.

DonLakeFlyer:
Previously the control just reach up and used the assume "map" property from higher up the hierarchy. But that control id is not always there. Hence it needs to be passed in.

DonLakeFlyer:
Thanks a bunch though. This tells me what I screwed up.

DonLakeFlyer:
I'll dig around and find all the places.

feemi:
i was feeling that it was not correct fix but it was working :) thanks for quick reply

DonLakeFlyer:
Awesome, thanks

DonLakeFlyer:
Can you check my pull instead. This one is still wrong due to the name change of the property.

Issue #6628, opened 2018-06-21
Video settings ui screwed up


DonLakeFlyer:
![screen shot 2018-06-21 at 11 32 35 am](https://user-images.githubusercontent.com/5876851/41738431-e22961b0-7546-11e8-96bf-ce07ee42d559.png)

PR #6636, opened 2018-06-22
Settings page which works at any size

DonLakeFlyer:
After a ridiculous amount of time fighting with Qml Layout I finally have a Settings page which works at any windows/font size. It's not quite as pretty as the previous one, but pretty close.

screen shot 2018-06-22 at 9 48 43 am

iPhone 5s:
screen shot 2018-06-22 at 9 48 19 am


DonLakeFlyer:
This fixes #6628 as well.

PR #6637, opened 2018-06-22
Add missing mapControl property setting

DonLakeFlyer:
Fix for #6625

### DIFF
--- a/src/MissionManager/QGCMapPolygonVisuals.qml
+++ b/src/MissionManager/QGCMapPolygonVisuals.qml
@@ -328,6 +328,7 @@ Item {
 
         MissionItemIndicatorDrag {
             id:         dragArea
+            mapControl: _root.mapControl
             z:          _zorderDragHandle
             visible:    !_circle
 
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
@feemi Could you see if this fixes the problem? For some reason it doesn't repro for me.

feemi:
i am on it

feemi:
@DonLakeFlyer it seems the problem fixed for mission points and geofence but it is not working for rally points

app output:

qrc:///qml/QGroundControl/FlightMap/MissionItemIndicatorDrag.qml:55: TypeError: Cannot call method 'toCoordinate' of undefinedqrc:///qml/QGroundControl/FlightMap/MissionItemIndicatorDrag.qml:55: TypeError: Cannot call method 'toCoordinate' of undefined
qrc:///qml/QGroundControl/FlightMap/MissionItemIndicatorDrag.qml:55: TypeError: Cannot call method 'toCoordinate' of undefined

DonLakeFlyer:
I must have screwed one up. I'll fix it.

PR #6658, opened 2018-06-26
Comm lost fixes

acfloria:
- Reset the comm lost timer with every mavlink message instead of the heartbeat

- Fix the link fallback in case of multiple links


DonLakeFlyer:
Change looks good. I'd say merge and then I'll do some testing later today with ArduPilot. Ok with that?

acfloria:
Fine with me. I tested it with PX4 regular telemetry, downloading logs over usb link and px4 multilink (serial + high latency)

PR #6667, opened 2018-06-28
Update to Upstream Master

acfloria:
Tested the following cases:
- SITL
- SenseSoar2 without mppt on the ground
- SatCom simulator

I tested the functionalities I could using the three test cases and I did not find an issue.


acfloria:
Sorry PR to the wrong repo.

PR #6674, opened 2018-07-02
Flickable QmlTest content

Rogni:
make flickable content in QmlTest.qml view


DonLakeFlyer:
Is there some way to redo this change such that it doesn't end up in such a massive diff? All it really needs is a QGCFlickable around the outermost Column control. Given this is debug code I don't want to to break and also don't want to waste time trying to pour through a massive diff.

PR #6680, opened 2018-07-03
AppSettings: Generate copy-paste code

patrickelectric:

### DIFF
--- a/src/Settings/AppSettings.cc
+++ b/src/Settings/AppSettings.cc
@@ -18,14 +18,14 @@
 const char* AppSettings::name =                                         "App";
 const char* AppSettings::settingsGroup =                                ""; // settings are in root group
 
-const char* AppSettings::offlineEditingFirmwareTypeSettingsName =       "OfflineEditingFirmwareType";
-const char* AppSettings::offlineEditingVehicleTypeSettingsName =        "OfflineEditingVehicleType";
-const char* AppSettings::offlineEditingCruiseSpeedSettingsName =        "OfflineEditingCruiseSpeed";
-const char* AppSettings::offlineEditingHoverSpeedSettingsName =         "OfflineEditingHoverSpeed";
-const char* AppSettings::offlineEditingAscentSpeedSettingsName =        "OfflineEditingAscentSpeed";
-const char* AppSettings::offlineEditingDescentSpeedSettingsName =       "OfflineEditingDescentSpeed";
-const char* AppSettings::batteryPercentRemainingAnnounceSettingsName =  "batteryPercentRemainingAnnounce";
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
I don't get what this is? In general you can't change the public names of settings member variables since it will cause all custom builds which reference those to break and need to be updated.

dogmaphobic:
The intent is sound. It replaces that boatload of code with the macros I created. The problem is that the names changed in the process, which will need to be reflected in custom builds that use them. Not sure why the macro definition is shown as an addition. I had already added that somewhere.

dogmaphobic:
Oh, now I see why it's an "addition". I've added to the Airmap branch, which has not been merged yet. @patrickelectric, look here: https://github.com/mavlink/qgroundcontrol/blob/Airmap/src/Settings/SettingsGroup.h#L20-L46
I had already done that and replaced some of those overly verbose implementations in that branch. This PR will make life quite complicated when merging. 

Do not merge this PR until we can better assess the ramifications to the changes already in the Airmap branch (which is essentially up-to-date with master). If you think your macro implementation is better, I would suggest making the changes in the Airmap branch instead.

dogmaphobic:
And here is an example using it:

https://github.com/mavlink/qgroundcontrol/blob/Airmap/src/Airmap/AirMapSettings.h
https://github.com/mavlink/qgroundcontrol/blob/Airmap/src/Airmap/AirMapSettings.cc

I did not change existing settings because of the lower case change it incurs. I need to evaluate what is and what isn't safe to change. The idea was for things moving forward to use the macros.

DonLakeFlyer:
I totally missed the macro part!

@patrickelectric I've mentioned this before. But I need way more detail descriptions (or any at all!) for your pull requests. Otherwise I just waste my time trying to figure out what something is.

DonLakeFlyer:
The value of cutting everything over to some new macro would outweight the PITA of updating all the customs builds I manage.

patrickelectric:
@DonLakeFlyer, sorry ! I really need to be more explicit in mr PRs.
@dogmaphobic I didn't see this branch ! Glad that I was in the correct path though.
The are any roadmap to release a 4.0v of QGC that breaks compatibility to improve this kind of things ? Just wondering.
And, if Gus is working on it, this PR can be closed.

DonLakeFlyer:
I'll let you and Gus work out which version you want to use. All seems like goodness. The really important thing is that the actual settings names from a QSetttings standpoint can't change since users will lose settings. Dealing with code changes in custom builds is ok since there is enough ROI to justify it. The other thing I would say is that if we are going to switch to these macros I'd like to see them used everywhere appropriate so we don't have a mix of new/old.

DonLakeFlyer:
@dogmaphobic Pulled across him similar thing from the branch he had it in.

Issue #6679, opened 2018-07-03
Add NXPHlite V3


davids5:
@DonLakeFlyer 

Would you please add the USB auto connect for PX4 NXPHlite v3.x

CONFIG_CDCACM_VENDORID=0x1FC9
CONFIG_CDCACM_VENDORSTR="NXP"
 CONFIG_CDCACM_PRODUCTSTR="PX4 NXPHlite v3.x"
CONFIG_CDCACM_PRODUCTID=0x001c

The firmware will be under development for a bit but will be in nxphlite-v3_default.px4 when released.

davids5:
I should have also mentioned the board_id is 28 decimal (0x1c). It is the same as the product id.

davids5:
@dogmaphobic would you be able to add this?

dogmaphobic:
> @dogmaphobic would you be able to add this?

Sure! Blindly but you can let me know if it works...

DonLakeFlyer:
What is the board id from the bootloader for this board?

davids5:
board_id is 28 decimal (0x1c). It is the same as the product id

LorenzMeier:
A pull request with those extensions would be most appreciated. It has probably reached the point where board support should be factored out of the code and into an XML file or equivalent. I'm closing this as it is not a real issue and something that could be easily contributed. Thanks for submitting the PR and I'm closing this here.

PR #6688, opened 2018-07-06
Add auto connect support for the PX4 NXPHlite v3.x

dogmaphobic:
From #6679 

@davids5 I've added the info for auto connect. I didn't mess with the firmware upload.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,7 +15,7 @@ environment:
 install:
   - git submodule update --init --recursive
   - call "%ProgramFiles(x86)%\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" x86
-  - set PATH=C:\Qt\Tools\QtCreator\bin;C:\Qt\5.11.0\msvc2015\bin;%PATH%
+  - set PATH=C:\Qt\Tools\QtCreator\bin;C:\Qt\5.11.1\msvc2015\bin;%PATH%
   - mkdir %LOCALAPPDATA%\QtProject && copy test\qtlogging.ini %LOCALAPPDATA%\QtProject\
   - ps: |
       Write-Host "Installing GStreamer..." -ForegroundColor Cyan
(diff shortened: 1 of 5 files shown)


davids5:
@dogmaphobic is there Windows build artifact I can grab and test?

davids5:
@dogmaphobic - I build and test it on a linux VM and it segfaults on insertion.

_prgFirmware is null.

image

I have a fix [here](https://github.com/mavlink/qgroundcontrol/pull/6700)

davids5:
@dogmaphobic @DonLakeFlyer @LorenzMeier  - where is the repo/src for the windows inf file in the driver package for windows located?

davids5:
@LorenzMeier 

As this sits it will allowing users to select a file local FW file to upload on linux and mac.

For Windows we need to update the inf file in px4driver.msi  - Where is the src to that?

For the long term the deployment URLS will need to be added in qgc.

LorenzMeier:
@dogmaphobic Looks like a build failure on AppVeyor.

dogmaphobic:
> @dogmaphobic Looks like a build failure on AppVeyor.

AppVeyor keeps changing the version of Qt and removing the old ones. It's a rat's race to keep track of it. I'll do the switch for Windows only.

dogmaphobic:
> For Windows we need to update the inf file in px4driver.msi - Where is the src to that?

I have no idea. @DonLakeFlyer, do you know?

davids5:
This is still open

@LorenzMeier

As this sits it will allow users to select a file local FW file to upload on linux and mac.

For Windows we need to update the inf file in px4driver.msi - Where is the src to that?

For the long term the deployment URLS will need to be added in qgc.

DonLakeFlyer:
The Windows drivers comes Michael Oborne and Mission Planner. I coped it from there. Not sure where the source is for that.

PR #6706, opened 2018-07-11
Remove boost dependency from airmap public interface

JonasVautherin:
Update QGC to work with the latest master of airmapd (i.e. without depending on boost).

Note that the airmapd binaries for macOS are in the repo. Should I update them, or should they not be versioned in the first place?

@dogmaphobic: FYI

### DIFF
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -153,28 +153,23 @@ contains (DEFINES, DISABLE_AIRMAP) {
     message("Skipping support for AirMap (manual override from user_config.pri)")
 } else {
     AIRMAPD_PATH = $$PWD/libs/airmapd
-    # Temporary until we get rid of boost
-    exists($${AIRMAPD_PATH}/include/boost) {
-        MacBuild|iOSBuild {
-            exists($${AIRMAPD_PATH}/macOS/Qt.5.9) {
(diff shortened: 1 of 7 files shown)


dogmaphobic:
The airmapd binary is linked against Qt. It must match the version used by QGC. The fact that QGC, which now uses Qt 5.11 works with airmapd built against Qt 5.9 is a fluke. The existing binary should be replaced by a new one, built against Qt 5.11 (and properly versioned).

JonasVautherin:
Well, my local built is using Qt 5.11 for both QGC and airmapd. Obviously this PR will not work as the binaries are not the right ones :sweat_smile:.

I guess my question was whether I should add the binaries to the repo or not. I usually don't do this, but if that's normal practice in the qgroundcontrol repo then I can add them. Should I add my binaries for macOS and Linux (built against Qt 5.11.0)? Or only macOS?

thomasgubler:
➤ Jonas Vautherin commented:

Travis fails because it tries to build for armv7 (but Airmap most likely has this issue as well), and Windows I don't know.

dogmaphobic:
The AppVeyor failure is due to them changing the Qt version. This has been adjusted in a different PR. The Travis failure is standard Travis flakiness.

PR #6713, opened 2018-07-13
Add Trimble MB-Two GPS support

bkueng:
This adds base station support for the Trimble MB-Two RTK GPS.
- The device gets automatically detected & connected
- Survey-in works similar as for u-blox, except that no minimum accuracy can be specified
- After that it starts streaming RTCM correction messages

GPS driver update: https://github.com/PX4/GpsDrivers/pull/28


DonLakeFlyer:
@bkueng Can you rebase ( hat will fix appveyor problem) and fix merge conflict? I'd like to get this through today since I"m branching for Stable 3.4 today.

PR #6714, opened 2018-07-13
Update to latest GPS drivers, add Trimble RTK support

DonLakeFlyer:
This is a replacement for #6713. Sorry @bkueng I couldn't wait and wanted this into Stable. So I fixed it up myself. I tested using my ublox RTK and all was still working.


DonLakeFlyer:
@bkueng It's failing cross platform builds. So look slike this won't go into 3.4. I'll let you take a look. You can use either yours or my branch to work from.

PR #6717, opened 2018-07-14
Add Trimble MB-Two GPS support #2

bkueng:
Rebased version of #6713.

Let's see if CI passes...


DonLakeFlyer:
Awesome, thanks @bkueng

PR #6728, opened 2018-07-19
alow position edit in Polyline

yl5006:
name: add  Polyline position edit

add
image
like this 
all  mapPolygon and MapCircle can be add  position edit menu

### DIFF
--- a/src/MissionManager/QGCMapPolylineVisuals.qml
+++ b/src/MissionManager/QGCMapPolylineVisuals.qml
@@ -142,6 +142,12 @@ Item {
             onTriggered:    mapPolyline.removeVertex(menu._removeVertexIndex)
         }
 
+        MenuItem {
+            id:             editVertexItem
+            text:           qsTr("Edit position..." )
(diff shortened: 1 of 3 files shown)


src/MissionManager/QGCMapPolylineVisuals.qml
```diff
@@ -152,6 +158,14 @@ Item {
         }
     }
 
+    Component {
+        id: editPositionDialog
+        EditPositionDialog {
+            coordinate:  mapPolyline.vertexCoordinate(menu._removeVertexIndex)
```
DonLakeFlyer: mapPolyline.path[menu._removeVertexIndex] should give you the same things without needing to change the signature for vertexCoordinate.

DonLakeFlyer:
Thanks, nice new feature.

src/MissionManager/QGCMapPolylineVisuals.qml
yl5006: yes ， I miss it

PR #6730, opened 2018-07-24
Adding a slider to set the mission path angle

dogmaphobic:
This is in addition to the existing text edit control. You can either enter a numeric value directly or use the slider.

screen shot 2018-07-24 at 2 10 51 pm


DonLakeFlyer:
@antiheavy You latest stable doesn't have this. Do you want it in your stable?

dogmaphobic:
Done!

DonLakeFlyer:
Thanks Gus.

PR #6736, opened 2018-07-28
create new vworld map for Korea

stmoon:
Hi, 

I implemented the new map for Korea. It is clearer and has more information for Korea. 
I think it is very useful for Korea drone guys.

Thanks in advance

Best Regards,
SungTae Moon


DonLakeFlyer:
@dogmaphobic Can you look at this?

dogmaphobic:
It all looks correct. This is quite a bit of a narrow feature. Similar to Statkart. I need to rethink the whole mapping thing so QGC will continue to offer these without overcrowding the options for most users.

PR #6751, opened 2018-08-02
alow Polyline position edit

yl5006:
alow position edit in polyline。
image

in EditPositionDialog  
if use  coordinate = mapPolyline.path[menu._removeVertexIndex]
there will be bind loop  property


DonLakeFlyer:
> if use coordinate = mapPolyline.path[menu._removeVertexIndex]

Huh. Strange. But it is Qt after all! So you get used to strange.

Issue #6684, opened 2018-07-05
Need word wrap in QGCLabel on Offline Map Options Dialog


Rogni:
Need word wrap in QGCLabel on Offline Map Options Dialog
![2018-07-05 23-39-47](https://user-images.githubusercontent.com/20958918/42347045-455766bc-80ad-11e8-9f0f-8d07ddf6a182.png)

PR #6754, opened 2018-08-02
Wrap text in Offline Map Options

DonLakeFlyer:
Fix for #6684


DonLakeFlyer:
screen shot 2018-08-02 at 8 00 39 am

Issue #6723, opened 2018-07-16
Terrain bug, seeming Zero terrain height values for a couple WP after downloading off vehicle


Antiheavy:
Happened to notice this after downloading a mission from a vehicle (SITL).  WP 70 and 71 both show what looks like Zero MSL altitude which is incorrect for this location:

![image](https://user-images.githubusercontent.com/12117027/42782166-b66a1cce-890d-11e8-8fe6-45a1d5a1eed1.png)

I tried switching from "Rel" to "AGL" for these two waypoints and they jumped down below the ground when "AGL" was used, confirming QGC had the wrong terrain value there (i forgot to screen shot that).

After adjusting WP 71, then the terrain updates to the correct terrain height value which confirms the wrong value was shown:
![image](https://user-images.githubusercontent.com/12117027/42782250-f75bab9e-890d-11e8-83df-268b7903ce9a.png)

I believe I had a not-so-great internet connection because after uploading and downloading again I had a lot of yellow bars that persisted:
![image](https://user-images.githubusercontent.com/12117027/42782327-31a4b30e-890e-11e8-9ab9-53052fcfbaae.png)
and terrain network errors in the console:
![image](https://user-images.githubusercontent.com/12117027/42782360-49faba98-890e-11e8-943d-17225b27ad6b.png)

I cannot reproduce yet... been trying.

Antiheavy:
here is the mission plan file (attached zip folder).
[couple-WPs-missing-terrain.zip](https://github.com/mavlink/qgroundcontrol/files/2199267/couple-WPs-missing-terrain.zip)

DonLakeFlyer:
Or in other words the hash points to a tile but the validation says the coordinate is not in the tile due to precision error.

DonLakeFlyer:
#6755 Prevents bad surveys from being generated. But does not fix the underlying problem of the actual failure

DonLakeFlyer:
I believe the real fix is to change persistance of tile corners to doubles. But that is a file format change which is a little tricky.

DonLakeFlyer:
Even with this last pull there is still a bug here where this plan (
[WPs-missing-terrain-2018-08-02 copy.plan.txt](https://github.com/mavlink/qgroundcontrol/files/2258277/WPs-missing-terrain-2018-08-02.copy.plan.txt)
) no doesn't cause bad elevations to be loaded. But it still doesn't fully load elevation data.

DonLakeFlyer:
[...]

PR #6761, opened 2018-08-04
More terrain query fixes/logging

DonLakeFlyer:
* Don't allow intermediate coordinate(0,0) state to flow down to terrain query
* Better terrain logging
* FW Landing Pattern - Fix terrain query on land point after file load


DonLakeFlyer:
More hardening for #6723

Issue #6764, opened 2018-08-05
"Set Radius..." on circle survey's is in meters only.  does not change with change of distance units.


Antiheavy:
"Set Radius" on circle survey's is in meters only.  It does not change with change of distance units.  Today's QGC daily build.

To reproduce:

1) change distance units to "Feet".
2) create a Survey Pattern in the Plan screen.
3) change it to a "Circle".
4) select "Set Radius..."
5) notice that the text entry has no units associated with it, but that the radius drawn on-screen is in meters.

DonLakeFlyer:
Oopsie. I'll fix in Stable.

PR #6769, opened 2018-08-06
Polygon Circle: Respect app units

DonLakeFlyer:
Respect app units in setting polygon circle radius


DonLakeFlyer:
Fix for #6764

Issue #6760, opened 2018-08-04
ArduPilot-Flashing Chibios builds


lvale:
At the moment QGC is only aware of px4 type firmware, and ArduPilot is moving away from these. Currently the newer firmware builds (apj) can+t even be selected as a custom firmware.

DonLakeFlyer:
Yup. I'll get this going once I get Stable 3.4 out the door.

DonLakeFlyer:
Also new bootloader and board detection from com port naming.

lvale:
Thanks @DonLakeFlyer 
If you could "squeeze" this to the 3.4 release would be great :)

DonLakeFlyer:
Maybe in a point release after the first one goes out. Not sure. Depends on how big of a change.

PR #6785, opened 2018-08-09
ArduPilot: ChibiOS flash support for Pixhawk family boards

DonLakeFlyer:
Note: Support for non-pixhawk family boards is NYI.

Related to #6760


DonLakeFlyer:
This should finally be all good to go once it passes CI. I"ll get to non-Pixhawk boards soon.

PR #6789, opened 2018-08-13
Plan: Fix defaulting when you "Change Command"

### LINKED ISSUE #6783: QGC Master fails to upload mission to ArduPilot
Hi @DonLakeFlyer 
screen shot 2018-08-09 at 10 01 10

defining just a simple mission, when trying to upload to vehicle, this comes up. 

Apparently part of the mission is uploaded.

DonLakeFlyer:
* Param 1-4 are reset to 0
* Param 5-7 are reset to 0 depending on new command usage

Fixes #6783


DonLakeFlyer:
@lvale Can you play around with this. What it was doing previously was quite incorrect from leaving values around from the previous command.

PR #6798, opened 2018-08-14
If APM firmware type is disabled, skip all the firmware uploading

dogmaphobic:
Also cleaning up annoying c++14 warnings as we go along...

### DIFF
--- a/src/VehicleSetup/FirmwareUpgradeController.cc
+++ b/src/VehicleSetup/FirmwareUpgradeController.cc
@@ -36,20 +36,20 @@ struct FirmwareToUrlElement_t {
 
 uint qHash(const FirmwareUpgradeController::FirmwareIdentifier& firmwareId)
 {
-    return ( firmwareId.autopilotStackType |
+    return static_cast<uint>(( firmwareId.autopilotStackType |
              (firmwareId.firmwareType << 8) |
-             (firmwareId.firmwareVehicleType << 16) );
+             (firmwareId.firmwareVehicleType << 16) ));
 }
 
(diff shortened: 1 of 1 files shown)


dogmaphobic:
This is a small part of a larger objective. The code is so bloated it's almost unusable on an not so old iPad. In this particular case, I was looking at this file to see how to go about extricating all of that to an online source (so it doesn't have to be all hardcoded).

In general, it's unrelated to APM. It just happened to be a simple thing within this one file. The code/memory footprint is getting too big. Everything is built in, even if you won't ever use it. I'm looking at ways to modularize it. We're also relying way too much on Javascript and that's causing CPU pains as well.

DonLakeFlyer:
> We're also relying way too much on Javascript and that's causing CPU pains as well.

Are you seeing that specifically on iPad? In the past I've found that iOs is dog slow writing to the disk. So if there is too much of that it just craps out. Specifically happened when the vehicle side was sending parameter updates which then in turn caused the parameter cache to be re-written over and over and over again. This was with a Solo, so I turned off parameter cache update completely on ArduPilot connects.

dogmaphobic:
It maxed out RAM and it's paging. So, yes, it's a case of disk I/O but as a result of not enough RAM.

DonLakeFlyer:
Memory leak?

dogmaphobic:
Not really. It's immediately, as soon as you boot it.

DonLakeFlyer:
Hmm, haven't looked at memory in ages. Worth a look in next stable.

PR #6812, opened 2018-08-21
Add Support for GPS Inject / Emlid Reach M+

GlidLov:
Currently, MP is the only GS sw which seems to support GPS inject function (supported by arducopter currently) together with EMLID products (Reach M+). It's a very good product which integrates well with UAV. I think it should be supported by the PX4 community, at least from a GS point of view. 
PX4 would need implementation of ERB protocol, which i suggest to do in the future.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,6 @@ Note: This file only contains high level features or important fixes.
 
 ## 3.4
 
-### 3.4.1 - Not yet released
+### 3.4.1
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Is there a reason you are submitting a Pull Request for this?

GlidLov:
DonLakeFlyer are you asking to me? Maybe i haven't used the proper way to ask a feature? Reason is that integrated rtk equipment is becoming popular

GlidLov:
sorry, thank you

PR #6814, opened 2018-08-21
Log Replay: Back up further from end to determine length

DonLakeFlyer:


DonLakeFlyer:
It's not happy the way I brought this over from Stable. I'll just merge stable into master instead.,

PR #6822, opened 2018-08-24
Adding Airmap integration to QGC

dogmaphobic:
This PR is the result of a few months of work integrating [Airmap](https://www.airmap.com) to QGC.


hamishwillee:
@dogmaphobic As posted in https://github.com/mavlink/qgroundcontrol/pull/6828#issuecomment-416078094

> I assume that in future we will want to ease integration of other services into QGC? If so, then can you take some notes as you work on correct project for integrating this sort of thing - we could add into the developer guide.
>
> Does this make sense?

lvale:
@dogmaphobic Currently Airmap gives errors (networking mostly) and that makes uploading missions fail because of the elevation that is sent is NaN

Should a regular issue be opened ?

[E] at /home/travis/build/mavlink/qgroundcontrol/src/Terrain/TerrainQuery.cc:480 - "Elevation tile fetching returned error ( QNetworkReply::NetworkError(UnknownNetworkError) )"

hnimfjdfefamachh
glgfjpbgmmgkboio

dogmaphobic:
> Currently Airmap gives errors (networking mostly) and that makes uploading missions fail because of the elevation that is sent is NaN

This is not related to the code added in this PR. This PR is only about flight planning and monitoring. Elevation data uses a different API and is separate from this.

PR #6828, opened 2018-08-26
Try to build with airmap-qt for iOS

JonasVautherin:
Add support for Airmap for iOS.


QGCSetup.pri
```diff
@@ -58,6 +58,14 @@ MacBuild {
     }
 }
 
+iOSBuild {
+    # AirMap
+    contains (DEFINES, QGC_AIRMAP_ENABLED) {
+        QMAKE_POST_LINK += && rsync -a $$BASEDIR/libs/airmapd/iOS/Qt.5.11.0/* $$DESTDIR/$${TARGET}.app/Frameworks/
+        #QMAKE_POST_LINK += && install_name_tool -change "@rpath/libairmap-qt.0.0.1.dylib" "@executable_path/../Frameworks/libairmap-qt.0.0.1.dylib" $$DESTDIR/$${TARGET}.app/Contents/iOS/$${TARGET}
```
JonasVautherin: I believe that my `rpath` (coming from airmapd) is already correct, so this may be unnecessary here. Needs to be confirmed, though.

hamishwillee:
@JonasVautherin I assume that in future we will want to ease integration of other services into QGC? If so, then can you take some notes as you work on correct project for integrating this sort of thing - we could add into the developer guide.

Does this make sense?

JonasVautherin:
@hamishwillee: Will do!

Issue #6770, opened 2018-08-07
Weird KML


dogmaphobic:
[testpolygon.kml.txt](https://github.com/mavlink/qgroundcontrol/files/2264744/testpolygon.kml.txt)

If you open this KML (renamed to kml.txt so I could attach) and move the indicated vertex, things seem to go haywire.

<img width="1203" alt="screen shot 2018-08-06 at 10 30 18 pm" src="https://user-images.githubusercontent.com/749243/43751043-2f5da7ce-99c9-11e8-93bd-3d053282d2c0.png">

<img width="1203" alt="screen shot 2018-08-06 at 10 30 42 pm" src="https://user-images.githubusercontent.com/749243/43751048-341d3be4-99c9-11e8-8f65-214fb90643a5.png">

DonLakeFlyer:
I'll take a look at some point. But not too concerned with really strange shapes working. More concerned that theoretically this case should work fine.

thomasgubler:
I had a quick look. The _polygonModel object that is used internally for the polygon in the transects generation seems to stay correct. (That can be seen from the screenshot: the white lines follow the correct polygon). I have yet to take a closer look at the _polygonPath object that (I assume) gets picked up in qml for display.

DonLakeFlyer:
Ah, interesting. They could be out of sync. The other thing is possibly a Qt bug where the UI is not updating correctly to _polygonPath changes. That wouldn't suprise me at all either.

thomasgubler:
PR with possible fix here: https://github.com/mavlink/qgroundcontrol/pull/6838

PR #6838, opened 2018-08-29
[QmlObjectListModel] fix list insert

thomasgubler:
Fixes https://github.com/mavlink/qgroundcontrol/issues/6770

I don't think it makes sense to insert the objects at location `i` (`i` stays constant,  `j`  is changed).

Note that  `insert` is called from `append` with `i == _objectList.count()`

### DIFF
--- a/src/QmlControls/QmlObjectListModel.cc
+++ b/src/QmlControls/QmlObjectListModel.cc
@@ -194,7 +194,7 @@ void QmlObjectListModel::insert(int i, QList<QObject*> objects)
         }
         j++;
 
-        _objectList.insert(i, object);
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Yeah, that seems wrong. But it's strange that it seems so wrong that things would screw up all over the place. Is this method not really used much, can't remember.

DonLakeFlyer:
@dogmaphobic Can you do you cherry pick magic and bring this to 3.4?

dogmaphobic:
Sure!

dogmaphobic:
Done! (2e36cbf256f20a39ec9aac8fb1fbaa7c1591cf73)

PR #6845, opened 2018-09-04
Jenkinsfile use QGC Airmap API header

dagar:


dagar:
@dogmaphobic could you give this a try?

image

dogmaphobic:
It's all working. Thanks!

Issue #6846, opened 2018-09-04
Elevation data not available sets NaN into Planned Home Position altitude


lvale:
If there is a problem with networking, or even no network available, defining a mission makes override the terrain height defined by the user and Mission Upload to vehicle sends a NaN which the flight stack does not support.

![hnimfjdfefamachh](https://user-images.githubusercontent.com/3130363/45045890-58f1d380-b06c-11e8-8fc5-5b663039a7c9.png)
![glgfjpbgmmgkboio](https://user-images.githubusercontent.com/3130363/45045891-58f1d380-b06c-11e8-9ba1-c235c57abfa8.png)

ChristopherOlson:
Would like to add that Android seems to work fine. The above screenshots are from my linux machine. The AirMap elevation data is cached fine with Android. The Linux version fails to get it. Also same problem noted on Windows version.

![screenshot_2018-09-04-12-23-30](https://user-images.githubusercontent.com/22397132/45047153-db16d380-b03d-11e8-88d2-be1531fcd2bb.png)

![screenshot_2018-09-04-12-29-56](https://user-images.githubusercontent.com/22397132/45047340-6728fb00-b03e-11e8-9979-f35eec06fd7f.png)

DonLakeFlyer:
What OS @lvale? Some versions of Android have problems with SSL libs. As do some versions of ArchLinux.

DonLakeFlyer:
That said I think there is an error path here where if the home position height query fails the NaN flows through to the top.

DonLakeFlyer:
@ChristopherOlson Are you using the latest version of QGC? Windows should be working fine.

lvale:
@DonLakeFlyer if i disable networking it happens on all OS's, But the biggest issue here is really the NaN coming up, and then that "kills" uploading missions

lvale:
latest build of QGC

ChristopherOlson:
@DonLakeFlyer I just referenced a related issue where it is happening on Windows. I don't actually use Windows, don't have a computer with Windows, have no need for a computer with Windows. On my Android tablet it works fine and fetches and caches the tiles for the AirMap elevation data within seconds. it is failing on two different linux machines even with internet connection.

Using the daily build on one and 3.4.2 on the other. Using 3.4.2 on the Android tablet.

DonLakeFlyer:
@ChristopherOlson What flavor/version of Linux?

ChristopherOlson:
[...]

PR #6850, opened 2018-09-04
Plan: Fix terrain NaN making it all the way to Planned Home altitude.

DonLakeFlyer:
Fix for #6846


DonLakeFlyer:
Well that didn't work.

PR #6852, opened 2018-09-05
Enhancement: added X-UAV Mini Talon airframe in new group Plane V-Tail

fredowski:
The X-UAV Mini Talon is a V-Tail plane. This commit adds the
new group "Plane V-Tail" together with a svg image for the
new group.

The X-UAV Mini Talon in the new group "Plane V-Tail" looks like this:

qgc-mini-talon

I added the new group to separate it from the group "Plane A-Tail" which I found
misleading. I would not search for a V-Tail airframe in the group A-Tail.

I have prepared a PR PX4/Firmware#10414 for the corresponding airframe in the PX4 firmware. We use
the X-UAV Mini Talon in our [research project](https://www.hs-augsburg.de/homes/beckmanf/dokuwiki/doku.php?id=searchwing) and I plan to maintain and adapt the
parameters.


dagar:
Looks good to me.

DonLakeFlyer:
Would it be possible to collapse some the Plane airframe into a single top-level group? As it currently stands there is a "Plane" group which only has two selections in it. There is a Plane A-Tail group which has a single item in it and this adds yet another top level group for V-Tail which also has a single item in it. Couldn't these all be in a single top level grouping?

DonLakeFlyer:
Multi-rotor has similar problems. I don't see how have a huge pile of top level groups you need to scrool through improves usability.

dagar:
I agree with the grouping problem, but wasn't sure where or how to address it. I was hoping for a solution that would allow maintaining the graphic in some capacity.

@hamishwillee FYI

fredowski:
When we started with the X-UAV Mini Talon we could not find it in the list of planes. Then we started with the Albatros in the Plane A-Tail group because we thought this one is at least closer to the V-Tail that we have. I only knew that Albatros was A-Tail because it was in that group. So the aspect that there is some Group like "Flying Wing", "Standard Plane", "V-Tail", "A-Tail" gives you a starting point if you do not find your specific type. When we started with the Bixler3, we just used "Standard Plane" which was fine. Flying a standard plane with a V-Tail configuration would not be a good idea. The idea of putting airframes in one group which are fundamentally different seems a good idea to me. O.k. and there is the nice graphic...

DonLakeFlyer:
We can deal with the grouping later. Doesn't need to help up this pull.

fredowski:
Thanks everybody for reviewing the patch!

hamishwillee:
My 2 bits. A few choices:
1. Lose the images - everything in flat list under "Plane", "Copter" etc.
2. Add a level of indirection. Ie top level is plane, then down to plane options.
3. Add a filter for all, plane, copter, etc. If you do this then I would *seed* it filter with type of current frame if one is stored.

This problem is not going away. FWIW I would do 3. 

@fredowski Also FWIW, thanks for this PR :-)

PR #6858, opened 2018-09-07
Simple item fix

dogmaphobic:
This was pushed against Stable 3.4 but not to master (https://github.com/mavlink/qgroundcontrol/pull/6825)

@DonLakeFlyer: Was that intentional?


DonLakeFlyer:
I was supposed to bring it across but forgot.

PR #6860, opened 2018-09-08
Check for new application Version at boot [WIP DO NOT MERGE]

DonLakeFlyer:
@dogmaphobic Just putting this up so you can look at the diff. Do you see any problems conceptually with this? I know it still needs lot of work, ui, custom build overrides and what not. But is the direction correct. Any holes which will make it simply not work at all.

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -24,6 +24,7 @@
 #include <QStyleFactory>
 #include <QAction>
 #include <QStringListModel>
+#include <QRegularExpression>
 
 #ifdef QGC_ENABLE_BLUETOOTH
 #include <QBluetoothLocalDevice>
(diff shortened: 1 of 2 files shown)


dogmaphobic:
Hmm... Maybe for Linux. For macOS and Windows, I've been using this for ages:

https://sparkle-project.org
https://winsparkle.org

It just works. If finds out if there is a new version, tells you about it and if you accept, it downloads it, installs it and restarts. All automatic.

DonLakeFlyer:
For this first go round I'm planning something dirt simple. Tell them there is a new version and pop the download page in the browser. Otherwise I think I'd just go with the Qt installer thing but that is way more work.

dogmaphobic:
Ha! I didn't realize the Qt installer had its own self update thingy. Either way, I agree that for the short term this is just fine.

DonLakeFlyer:
Just something quick for current Stable. I'll finish it up when I get back home.

PR #6863, opened 2018-09-10
First cut at Chinese localization

dogmaphobic:
The bulk of QGC's strings have been localized to Chinese. There are some new strings that have been added since, which I will be updating shortly. This also makes it obvious the deficiencies of having hard-coded strings defined within Json and XML files, which are not localized (all dropdown boxes in General Settings below as an example). 

screen shot 2018-09-10 at 1 39 00 pm

screen shot 2018-09-10 at 1 40 13 pm


hamishwillee:
@dogmaphobic Even though there are a bunch of untranslated hard-coded strings (which presumably we can fix?) this is a great first step.

dogmaphobic:
> which presumably we can fix?

The ones within QGC yes. These come from Json files defining options and parameters. The trick would be producing the proper localization files off of them so they can be sent to Crowdin. The main issue are strings that come from the firmware. Both dynamically (over the wire) or through the parameter files. None of those have any means of localization.

bys1123:
Thanks, with this version we can also easy to start proofreading now. 20% finished.
Our team can also help on hard-coded strings, we have Qt developers, but need a guideline or something, in case we mass the code.

bys1123:
Do you have 3.4.4 release date? Seems closer. I want to make sure that I can finish proofreading about 50%, need one or two days, makes translations reliable. Do I have that time? @dogmaphobic

hamishwillee:
@bys1123 I wouldn't stress about any particular release - even if we released what we have now it will be  massive improvement.

I wouldn't touch the hard code strings until @dogmaphobic has had a chance to plan a consistent way of handling them. Gus, any ETA on when you might look at that ... days, months?

dogmaphobic:
> ... until @dogmaphobic has had a chance to plan a consistent way of handling them. Gus, any ETA on when you might look at that ... days, months?

The strings that are generated within the QGC repository itself are easier to deal with. I need to wait for Don to coordinate it but it shouldn't be much work on the QGC side. It will be some work in trying to sort out how to get these strings to Crowdin and back.

The strings coming from the firmware are a whole other ball of wax. We would need to get the firmware folks involved as the bulk of the work must be done on that end.

hamishwillee:
@bys1123 Don is away for at least two weeks, and coordinating the other strings will happen some time after that. Upshot, not something to factor in at this point.

Thanks @dogmaphobic . Strings from firmware are going to be a pain, but I suspect that QGC is better off defining what it needs before talking to the Firmware side.

bys1123:
@dogmaphobic Chinese translation proofreading finished, ready for release.

hamishwillee:
@dogmaphobic Can we get that into the daily build please?

dogmaphobic:
I just updated the strings here: https://github.com/mavlink/qgroundcontrol/pull/6901

PR #6865, opened 2018-09-11
Added settings for vehicles' icon and trajectory points colors to clearly distinguish verhicles of different groups.

JloveU:
---
name: Added settings
about: Added settings for vehicles' icon and trajectory points colors to clearly distinguish verhicles of different groups.

---

**Log Files and Screenshots**

default
default
default


DonLakeFlyer:
Interesting, but way too complex. A much better solution would be to automatically cycle through a list of colors with no user selectable setting.

DonLakeFlyer:
Sorry much to complex from an end user standpoint. Happy to discuss alternatives.

PR #6868, opened 2018-09-12
[WIP] RTK: possibility to set the base position

bkueng:
- The first commit adds a retry to open the device (see https://github.com/mavlink/qgroundcontrol/commit/5556cb9f37a33acfe583c9094a63ac7cf219696a)
- requires a GPS submodule udpate (https://github.com/PX4/GpsDrivers/pull/31)

@DonLakeFlyer you can take over from here and rework the second commit as needed. The idea is as following:
- You can get the final accurate base position after a survey-in from [here](https://github.com/PX4/GpsDrivers/pull/31/files#diff-03cf76148bc1d45d2195bbcd8f36f0dbR110). Note that for Trimble, the position during a survey-in is NaN, and will only be set on the last survey-in status update, when successfully completed.
- Later on you can pass the position to `setBasePosition` during startup.

### DIFF
--- a/src/GPS/GPSProvider.cc
+++ b/src/GPS/GPSProvider.cc
@@ -44,8 +44,21 @@ void GPSProvider::run()
     _serial = new QSerialPort();
     _serial->setPortName(_device);
     if (!_serial->open(QIODevice::ReadWrite)) {
-        qWarning() << "GPS: Failed to open Serial Device" << _device;
-        return;
+        int retries = 60;
+        // Give the device some time to come up. In some cases the device is not
+        // immediately accessible right after startup for some reason. This can take 10-20s.
+        while (retries-- > 0 && _serial->error() == QSerialPort::PermissionError) {
+            qCDebug(RTKGPSLog) << "Cannot open device... retrying";
+            msleep(500);
+            if (_serial->open(QIODevice::ReadWrite)) {
+                _serial->clearError();
+                break;
+            }
+        }
+        if (_serial->error() != QSerialPort::NoError) {
+            qWarning() << "GPS: Failed to open Serial Device" << _device << _serial->errorString();
+            return;
+        }
     }
     _serial->setBaudRate(QSerialPort::Baud9600);
     _serial->setDataBits(QSerialPort::Data8);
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Yup, just back in town now. Will do.

src/GPS/GPSProvider.cc
```diff
@@ -72,6 +85,13 @@ void GPSProvider::run()
         }
         gpsDriver->setSurveyInSpecs(_surveyInAccMeters * 10000, _surveryInDurationSecs);
 
+        // fixed base position:
+        double latitude = xyz;
+        double longitude = xyz;
+        float altitude = xyz;
+        float position_accuracy = xyz;
+        gpsDriver->setBasePosition(latitude, longitude, altitude, position_accuracy);
```
DonLakeFlyer: @bkueng What is the position_accuracy field here?

bkueng: You get that from the survey-in status: https://github.com/PX4/GpsDrivers/blob/master/src/gps_helper.h#L112. Not sure how important it is. If unknown, I'd use something like 50cm.

bkueng:
> Should it only be allowed to set base position from the actual survey-in result? Or it should it also be allowed to set that value yourself, assuming you know it somehow?

Allowing a manual entry would make sense I think. For example when switching computers.

src/GPS/GPSProvider.cc
DonLakeFlyer: Wondering if really needed? Poking around in Mission Planner UI which already support this feature shows that it doesn't have this value.

bkueng: You need to provide a value for the u-blox when setting the position. So I would say we should do it properly and use the correct value (since we already have it).

DonLakeFlyer: Got it

DonLakeFlyer: for setBasePosition call is position_accuracy in mm?

bkueng: Yes in mm: https://github.com/PX4/GpsDrivers/blob/master/src/base_station.h#L76

PR #6870, opened 2018-09-15
Move the generated qml cache files to a separate directory

zhxt:


DonLakeFlyer:
Why is this needed?

zhxt:
@DonLakeFlyer 
> Why is this needed?

For short:
to make the build directory looks more clean.

A long version:
When I run a `ls ` cmd in the build directory after successfully build, the output full filled the whole screen, the first thought came to my mind is put the generated files to a separated dir like the other generated files do (moc_* in moc,  ui_* in ui, qrc_* in rcc , etc ...), so this PR was born.

DonLakeFlyer:
I guess that makes sense. @dogmaphobic Is this ok?

zhxt:
@DonLakeFlyer @dogmaphobic I was thinking will it be better to put it **permanently** with:
https://github.com/mavlink/qgroundcontrol/blob/2d37f689938f48609a9815f2613333220a675a99/QGCCommon.pri#L172-L176

or let it be **configurable** with `CONFIG += qtquickcompiler`?

DonLakeFlyer:
I would argue against configurable. Too many options just confuse things.

zhxt:
Totally agree, will update it to put them together.

zhxt:
Removed prefixed "$${OUT_PWD}".
since it will generates  'full path' prefixed makefile target for implicit rules.
```
make: *** No rule to make target 'qmlcache/src_QtLocationPlugin_QMLControl_OfflineMap_qml.cpp', needed by 'obj/src_QtLocationPlugin_QMLControl_OfflineMap_qml.o'.  Stop.
```
```
# diff -uN Makefile Makefile.new
...
-qmlcache/src_QtLocationPlugin_QMLControl_OfflineMap_qml.cpp: /home/zhxt/playground/qgroundcontrol/src/QtLocationPlugin/QMLControl/OfflineMap.qml \
+/home/zhxt/playground/qgroundcontrol/build/qmlcache/src_QtLocationPlugin_QMLControl_OfflineMap_qml.cpp: /home/zhxt/playground/qgroundcontrol/src/QtLocationPlugin/QMLControl/OfflineMap.qml \
...
```
This could be a bug of qmake.

DonLakeFlyer:
@dogmaphobic Is this useful?

DonLakeFlyer:
The QMLCACHE_DIR setting doesn't seem to work in all version of Qt. So for now we can't do this.

PR #6884, opened 2018-09-25
Prevent clicks going through top level items

### LINKED ISSUE #6879: Goto are created below the top bar
Sometimes I when I press an icon in the top bar I miss it slightly and it creates instead a goto on the map beneath the half transparent top bar. The same bar happens beneath the HUD.

DonLakeFlyer:
Fixes #6879


DonLakeFlyer:
@dogmaphobic Do you see any problems with this?

PR #6891, opened 2018-09-26
Stable desktop version check

DonLakeFlyer:
This is a dirt simple implementation but it gets the job done. If version is out of date it shows this:
screen shot 2018-09-26 at 9 10 08 am

For custom builds this feature defaults to off. Custom build can override to turn on. From QGCCorePlugin.h:
```
    /// Returns the url to download the stable version check file. Return QString() to indicate no version check should be performed.
    /// Default QGC mainline implemenentation returns QGC Stable file location. Default QGC custom build code returns QString().
    /// Custom builds can override to turn on and provide their own location.
    /// The contents of this file should be a single line in the form:
    ///     v3.4.4
    /// This indicates the latest stable version number.
    virtual QString stableVersionCheckFileUrl(void) const;

    /// Returns the user visible url to show user where to download new stable builds from.
    /// Custom builds must override to provide their own location.
    virtual QString stableDownloadLocation(void) const { return QString("qgroundcontrol.com"); }
```

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,7 +4,8 @@ Note: This file only contains high level features or important fixes.
 
 ## 3.4
 
-### 3.4.4 - Not yet released
+### 3.4.4
+* Stable desktop versions now inform user at boot if newer version is available.
 * Multi-Vehicle Start Mission and Pause now work correctly. Issue #6864.
 
 ### 3.4.3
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
Last commit modifies this such that custom builds will default to having this feature off.

PR #6898, opened 2018-09-27
Move ArduPilot preflight check handling to generic QGC

DonLakeFlyer:
Replacement for #6436

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSensorsComponentController.cc
+++ b/src/AutoPilotPlugins/APM/APMSensorsComponentController.cc
@@ -102,7 +102,7 @@ void APMSensorsComponentController::_startLogCalibration(void)
 {
     _hideAllCalAreas();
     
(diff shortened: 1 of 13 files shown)


DonLakeFlyer:
@dagar Can you do some testing with this?

bkueng:
I tested this against https://github.com/PX4/Firmware/pull/10586: it works, but can we check the log level? Some messages are only warnings and will still allow arming (and thus should not appear on the screen).

Can you add a semitransparent background box around the text? It's a bit hard to read depending on the location.

DonLakeFlyer:
@bkueng What severity should show on the map?

bkueng:
>= critical

DonLakeFlyer:
@bkueng Can you try it again.

bkueng:
It works, thanks. I noticed that github rendered my reply not as expected. I meant to say `>= critical`.

DonLakeFlyer:
> I meant to say `>= critical`.

I'll change that.

PR #6899, opened 2018-09-28
Update PX4 Firmware metadata Fri Sep 28 14:14:57 UTC 2018

dagar:


dagar:
@dogmaphobic OSX failures look real (unrelated to this PR).

dogmaphobic:
I wonder why it only happens in that system. I'm rebooting and updating Xcode...

dogmaphobic:
It was a (new) warning within the Airmap SDK stuff. I assume the build is set to fail on warning? I wonder how it passes all those zillion warnings with mavlink.h. Either way, I "fixed" both here: https://github.com/mavlink/qgroundcontrol/pull/6903

DonLakeFlyer:
If I remember correctly I thought only the Linux build is set to fail on warning because new clang awhile back turned on all sorts of random crap and it was spewing output.

PR #6901, opened 2018-09-28
Updating Chinese localization

dogmaphobic:
screen shot 2018-09-28 at 11 16 26 am


bys1123:
Looks like this is the previous build translation version, not the latest one, I think maybe you should click "build & download" not "Download latest". I corrected about 2000 strings these days, here only shows 300 lines changed. 
Can you help check again? thanks very much.

bys1123:
Yes, I built it. This is right now, thanks. 👍

PR #6902, opened 2018-09-28
Warn when not using latest stable version of Ardupilot.

Williangalvani:
This warns the user when the running firmware is not the latest stable.
The objective is to help the users keep an up-to-date firmware and reduce the incidence of problems due to incompatibility with older firmware versions.

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -16,6 +16,7 @@
 #include "APMSensorsComponentController.h"
 #include "MissionManager.h"
 #include "ParameterManager.h"
+#include "QGCFileDownload.h"
 
 #include <QTcpSocket>
 
@@ -1003,4 +1004,26 @@ void APMFirmwarePlugin::startMission(Vehicle* vehicle)
     } else {
         vehicle->sendMavCommand(vehicle->defaultComponentId(), MAV_CMD_MISSION_START, true /*show error */);
     }
+}
+
+QString APMFirmwarePlugin::_getLatestVersionFileUrl(Vehicle* vehicle)
+{
+    const static QString baseUrl("http://firmware.ardupilot.org/%1/stable/PX4/git-version.txt");
+    if (vehicle->fixedWing()) {
+        return baseUrl.arg("Plane");
+    }
+    if (vehicle->vtol()) {
+        return baseUrl.arg("Plane");
+    }
+    if (vehicle->rover()) {
+        return baseUrl.arg("Rover");
+    }
+    if (vehicle->sub()) {
+        return baseUrl.arg("Sub");
+    }
+    return baseUrl.arg("Copter");
+}
+
+QString APMFirmwarePlugin::_versionRegex() {
+    return QStringLiteral(" V([0-9,\\.]*)$");
 }
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -13,7 +13,9 @@
 #include "CameraMetaData.h"
 #include "SettingsManager.h"
 #include "AppSettings.h"
+#include "QGCFileDownload.h"
 
+#include <QRegularExpression>
 #include <QDebug>
 
 QGC_LOGGING_CATEGORY(FirmwarePluginLog, "FirmwarePluginLog")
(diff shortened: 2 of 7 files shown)


src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
```diff
@@ -1015,3 +1017,89 @@ void APMFirmwarePlugin::startMission(Vehicle* vehicle)
...
+        this,
+        [vehicle, this](QString remoteFile, QString localFile) {
+            _versionFileDownloadFinished(remoteFile, localFile, vehicle);
+        });
+    downloader->download(versionFile);
+}
+
+ void APMFirmwarePlugin::_versionFileDownloadFinished(QString& remoteFile, QString& localFile, Vehicle* vehicle)
+{
+    qCDebug(APMFirmwarePluginLog) << "Download complete" << remoteFile << localFile;
```
DonLakeFlyer: This should use QObject::sender() here to deleteLater the QGCFileDownload.

DonLakeFlyer:
This is really nice stuff. I wonder if it will annoy people though? I've noticed PX4 folks tend to stay up to date with latest Stable, but ArduPilot folks tend to lag behind commonly. I guess we'll find out :)

DonLakeFlyer:
Alsoo did you think you could do the PX4 side as well? It should be even simple than ArudPilot since it's just a single file.

DonLakeFlyer:
Also would is possible the use the same code in FirmwareUpgradeController which also reads this file to parse the version file? You could make a static method in say the ArduPilot firmware plugin. And then the saqme code could be used in both places.

DonLakeFlyer:
And then the same thing for the PX4 version file parsing.

Williangalvani:
I updated it to support PX4 and did some refactoring.
I had to comment the code checking the kind of version in use because latest PX4 is reporting itself as a "dev" version.

DonLakeFlyer:
> because latest PX4 is reporting itself as a "dev" version.

@dagar Bug in dropped PX4 stable builds? Seems like I remember reading something about that somewhere else.

src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -663,3 +667,73 @@ uint32_t FirmwarePlugin::highLatencyCustomModeTo32Bits(uint16_t hlCustomMode)
...
+        [vehicle, this](QString remoteFile, QString localFile) {
+            _versionFileDownloadFinished(remoteFile, localFile, vehicle);
+            sender()->deleteLater();
+        });
+    downloader->download(versionFile);
+}
+
+void FirmwarePlugin::_versionFileDownloadFinished(QString& remoteFile, QString& localFile, Vehicle* vehicle)
+{
+    Q_UNUSED(remoteFile);
```
jaxxzer: remoteFile is used here.

src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -663,3 +667,73 @@ uint32_t FirmwarePlugin::highLatencyCustomModeTo32Bits(uint16_t hlCustomMode)
...
+        &QGCFileDownload::downloadFinished,
+        this,
+        [vehicle, this](QString remoteFile, QString localFile) {
+            _versionFileDownloadFinished(remoteFile, localFile, vehicle);
+            sender()->deleteLater();
+        });
+    downloader->download(versionFile);
+}
+
+void FirmwarePlugin::_versionFileDownloadFinished(QString& remoteFile, QString& localFile, Vehicle* vehicle)
```
jaxxzer: I think the implementation here needs to be moved back to ardupilotfirmwareplugin, and a new one made for px4. 

For PX4, as far as I can see, there is no guarantee that the first regex match will be the stable version number. It might be a beta release, prerelease, or old release. While none of these things are true at the moment, there aren't any guarantees that it will remain this way.

src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -13,9 +13,13 @@
 #include "CameraMetaData.h"
 #include "SettingsManager.h"
 #include "AppSettings.h"
+#include "QGCFileDownload.h"
 
+#include <QRegularExpression>
 #include <QDebug>
 
+QGC_LOGGING_CATEGORY(FirmwarePluginLog, "FirmwarePluginLog")
```
jaxxzer: I recommend a separate commit

src/FirmwarePlugin/FirmwarePlugin.cc
Williangalvani: That's why the regex is looking for "vX.X.X Stable". All stable versions since 1.5.5 (2017-01-25) in [the file](https://api.github.com/repos/PX4/Firmware/releases) follow this naming.  But I agree this is not as robust.

At [FirmwareUpgradeController.cc](https://github.com/mavlink/qgroundcontrol/blob/master/src/VehicleSetup/FirmwareUpgradeController.cc#L848) the json is properly processed, but then it is also iterated from top to bottom until it finds the first beta (prerelease=true) and first stable release (prerelease=false).

The current way there's less code, but might break if they change the naming standard.

jaxxzer: My concern is also the github api returning the list in chronological order. https://developer.github.com/v3/repos/releases/#list-releases-for-a-repository

It will probably continue to work as-is. We can wait till to fix it if it breaks.

src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -663,3 +667,73 @@ uint32_t FirmwarePlugin::highLatencyCustomModeTo32Bits(uint16_t hlCustomMode)
...
+        return;
+    }
+    int stableMajor = versionNumbers[0].toInt();
+    int stableMinor = versionNumbers[1].toInt();
+    int stablePatch = versionNumbers[2].toInt();
+
+    if ((vehicle->firmwareMajorVersion() != stableMajor)
+        || (vehicle->firmwareMinorVersion() != stableMinor)
+        || (vehicle->firmwarePatchVersion() != stablePatch)
+      //|| (vehicle->firmwareVersionType() != FIRMWARE_VERSION_TYPE_OFFICIAL)  // commented because px4 stable is showing as "dev"
```
DonLakeFlyer: This is a bug in the PX4 release process for the latest build. This should be uncommented.

DonLakeFlyer: It was working fine before.

src/FirmwarePlugin/FirmwarePlugin.cc
DonLakeFlyer: I'm slightly uncomfortable with this not being broken out to PX4 and ArduPilot versions. Mainly because the _versionRegEx mechanism is not super clear at first glance. But that said it works, so fine.

Williangalvani:
I've changed the things @jaxxzer commented, and reworded the message.
This is how it looks now:

screenshot from 2018-10-03 11-14-26

DonLakeFlyer:
I think the check needs to be whether Stable is newer than the build being run as opposed to != to the build being run. Otherwise this is going to get annoying really fast for developers running new firmware bits.

Williangalvani:
Ok, I changed it, now it checks version number first, and the release is only checked if the versions match.
This is to catch the potential case where a user is using say, 1.0-dev, and then 1.0-stable is released, which would mean the user is behind.
What do you think?

DonLakeFlyer:
Looks good. We can adjust as needed if we get any complaints.

Issue #6621, opened 2018-06-21
Display coordinates of RTK base station


mkunzBGC:
I am using QGroundControl 3.3.2 with APM:Copter 3.5.5 and a Here+ RTK base station and rover. While RTK works, there seems to be no way to find out the coordinates of the base station. This would be valuable information e.g. when performing surveys of the same area on different days!

As a first step, having just the coordinates displayed would help me a lot. According to Don Gagne, this should be quite easy to achieve:

> Don Gagne@DonLakeFlyer       Jun 20 19:34

>>    When using QGroundControl for RTK, is there a way to find out the coordinates of the base station?

>Add an Issue and I think this can be easily added to the GPSRTK FactGroup which would make it available to the instrument panel.

On the long run, it would be nice to be able to save and restore base station coordinates, similar to how MissionPlanner handles RTK ([link](http://ardupilot.org/copter/docs/common-here-plus-gps.html#base-module-setting-using-mission-planner)). 

Finally, for people like me who would also like to use RTK with QGC for ground-based surveying, a possibility to do copy and paste with base and rover positions would be wonderful (currently, rover coordinates are available via Mavlink Inspector, but I cannot copy from there).

Thanks for your great work,
Martin

hamishwillee:
@DonLakeFlyer You might want to close this

DonLakeFlyer:
> While RTK works, there seems to be no way to find out the coordinates of the base station. This would be valuable information e.g. when performing surveys of the same area on different days!

This statement is still correct. Those coordinates should be available for display in the instrument panel.

PR #6907, opened 2018-10-01
Add support for specifying fixed RTK base position

DonLakeFlyer:
This allows you to specify a fixed RTK base position and skip survey-in. If you have a valid survey-in "Save Current Base Position" will enable allowing you to copy the values from the survey-in to setttings.

Replacement for #6868. Related to #6621.

screen shot 2018-10-01 at 9 49 36 am

### DIFF
--- a/src/GPS/GPSProvider.cc
+++ b/src/GPS/GPSProvider.cc
@@ -10,6 +10,8 @@
 
 #include "GPSProvider.h"
 #include "QGCLoggingCategory.h"
+#include "QGCApplication.h"
+#include "SettingsManager.h"
 
 #define GPS_RECEIVE_TIMEOUT 1200
 
@@ -17,7 +19,7 @@
 
 #include "Drivers/src/ubx.h"
 #include "Drivers/src/ashtech.h"
-#include "Drivers/src/gps_helper.h"
+#include "Drivers/src/base_station.h"
 #include "definitions.h"
 
 //#define SIMULATE_RTCM_OUTPUT //if defined, generate simulated RTCM messages
@@ -44,8 +46,21 @@ void GPSProvider::run()
     _serial = new QSerialPort();
     _serial->setPortName(_device);
     if (!_serial->open(QIODevice::ReadWrite)) {
-        qWarning() << "GPS: Failed to open Serial Device" << _device;
-        return;
+        int retries = 60;
+        // Give the device some time to come up. In some cases the device is not
+        // immediately accessible right after startup for some reason. This can take 10-20s.
+        while (retries-- > 0 && _serial->error() == QSerialPort::PermissionError) {
+            qCDebug(RTKGPSLog) << "Cannot open device... retrying";
+            msleep(500);
+            if (_serial->open(QIODevice::ReadWrite)) {
+                _serial->clearError();
+                break;
+            }
+        }
+        if (_serial->error() != QSerialPort::NoError) {
+            qWarning() << "GPS: Failed to open Serial Device" << _device << _serial->errorString();
+            return;
+        }
     }
     _serial->setBaudRate(QSerialPort::Baud9600);
     _serial->setDataBits(QSerialPort::Data8);
@@ -54,7 +69,7 @@ void GPSProvider::run()
     _serial->setFlowControl(QSerialPort::NoFlowControl);
 
     unsigned int baudrate;
-    GPSHelper* gpsDriver = nullptr;
+    GPSBaseStationSupport* gpsDriver = nullptr;
 
     while (!_requestStop) {
 
@@ -70,7 +85,11 @@ void GPSProvider::run()
             gpsDriver = new GPSDriverUBX(GPSDriverUBX::Interface::UART, &callbackEntry, this, &_reportGpsPos, _pReportSatInfo);
             baudrate = 0; // auto-configure
         }
-        gpsDriver->setSurveyInSpecs(_surveyInAccMeters * 10000, _surveryInDurationSecs);
+        gpsDriver->setSurveyInSpecs(_surveyInAccMeters * 10000.0f, _surveryInDurationSecs);
+
+        if (_useFixedBaseLoction) {
+            gpsDriver->setBasePosition(_fixedBaseLatitude, _fixedBaseLongitude, _fixedBaseAltitudeMeters * 10000.0f, _fixedBaseAccuracyMeters * 10000.0f);
+        }
 
         if (gpsDriver->configure(baudrate, GPSDriverUBX::OutputMode::RTCM) == 0) {
 
@@ -108,12 +127,27 @@ void GPSProvider::run()
     qCDebug(RTKGPSLog) << "Exiting GPS thread";
 }
 
-GPSProvider::GPSProvider(const QString& device, GPSType type, bool enableSatInfo, double surveyInAccMeters, int surveryInDurationSecs, const std::atomic_bool& requestStop)
-    : _device(device)
-    , _type(type)
-    , _requestStop(requestStop)
-    , _surveyInAccMeters(surveyInAccMeters)
-    , _surveryInDurationSecs(surveryInDurationSecs)
+GPSProvider::GPSProvider(const QString& device,
+                         GPSType    type,
+                         bool       enableSatInfo,
+                         double     surveyInAccMeters,
+                         int        surveryInDurationSecs,
+                         bool       useFixedBaseLocation,
+                         double     fixedBaseLatitude,
+                         double     fixedBaseLongitude,
+                         float      fixedBaseAltitudeMeters,
+                         float      fixedBaseAccuracyMeters,
+                         const std::atomic_bool& requestStop)
+    : _device                   (device)
+    , _type                     (type)
+    , _requestStop              (requestStop)
+    , _surveyInAccMeters        (surveyInAccMeters)
+    , _surveryInDurationSecs    (surveryInDurationSecs)
+    , _useFixedBaseLoction      (useFixedBaseLocation)
+    , _fixedBaseLatitude        (fixedBaseLatitude)
+    , _fixedBaseLongitude       (fixedBaseLongitude)
+    , _fixedBaseAltitudeMeters  (fixedBaseAltitudeMeters)
+    , _fixedBaseAccuracyMeters  (fixedBaseAccuracyMeters)
 {
     qCDebug(RTKGPSLog) << "Survey in accuracy:duration" << surveyInAccMeters << surveryInDurationSecs;
     if (enableSatInfo) _pReportSatInfo = new satellite_info_s();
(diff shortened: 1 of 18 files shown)


DonLakeFlyer:
@bkueng In my testing this seems to be working. But I am not RTK expert so could definitely use some additional testing.

DonLakeFlyer:
@hamishwillee Do you think you could add the doc for this into the Daily Build page?

src/GPS/GPSProvider.cc
```diff
@@ -70,7 +85,11 @@ void GPSProvider::run()
             gpsDriver = new GPSDriverUBX(GPSDriverUBX::Interface::UART, &callbackEntry, this, &_reportGpsPos, _pReportSatInfo);
             baudrate = 0; // auto-configure
         }
-        gpsDriver->setSurveyInSpecs(_surveyInAccMeters * 10000, _surveryInDurationSecs);
+        gpsDriver->setSurveyInSpecs(_surveyInAccMeters * 10000.0f, _surveryInDurationSecs);
+
+        if (_useFixedBaseLoction) {
+            gpsDriver->setBasePosition(_fixedBaseLatitude, _fixedBaseLongitude, _fixedBaseAltitudeMeters * 10000.0f, _fixedBaseAccuracyMeters * 10000.0f);
```
bkueng: Altitude is given in meters, and accuracy in mm (one 0 too much):
`_fixedBaseAltitudeMeters, _fixedBaseAccuracyMeters * 1000.0f`

src/QGCApplication.cc
```diff
@@ -197,7 +199,7 @@ QGCApplication::QGCApplication(int &argc, char* argv[], bool unitTesting)
             QMessageBox msgBox;
             msgBox.setInformativeText(tr("You are running %1 as root. "
                                          "You should not do this since it will cause other issues with %1. "
-                                         "%1 will now exit. "
+                                         "%1 will now exiQGroundControl.gpsRtkt. "
```
bkueng: typo?

src/Settings/RTK.SettingsGroup.json
```diff
@@ -20,5 +20,55 @@
...
+},
+{
+    "name":                 "FixedBasePositionLatitude",
+    "shortDescription":     "Base Position Latitude",
+    "longDescription":      "Defines the latitude of the fixed RTK base position.",
+    "type":                 "double",
+    "defaultValue":         0,
+    "min":                  -90,
+    "max":                  90,
+    "decimalPlaces":        7,
```
bkueng: Can you increase this to 9? That's what's used internally.

src/Vehicle/GPSRTKFact.json
```diff
@@ -2,35 +2,62 @@
...
     "decimalPlaces":    1,
-    "units":            "m"
+    "units":            "m",
+    "default":          null
+},
+{
+    "name":             "currentLatitude",
+    "shortDescription": "Current Survey-In Latitude",
+    "type":             "double",
+    "decimalPlaces":    7,
```
bkueng: Also: 9

src/GPS/GPSProvider.cc
DonLakeFlyer: Oops, thanks.

Yeah I was thinking about disabling the other ones. Let me look at that.

Other question: The field editors show a maximum number of decimal places. Although internally the full value is stored, visibly only 7 digits are shown after the decimal. What do you think the visible precision should me. Is 7 too few?

src/QGCApplication.cc
DonLakeFlyer: Yup.

src/Settings/RTK.SettingsGroup.json
DonLakeFlyer: Ah, thanks. I missed this before I posted my above decimal place question.

src/GPS/GPSProvider.cc
DonLakeFlyer: Disregard decimal place dup question. Already answered,.

DonLakeFlyer:
screen shot 2018-10-02 at 12 58 54 pm

DonLakeFlyer:
@mkunzBGC Do you have time to test once I merge in?

DonLakeFlyer:
@bkueng Is altitude and accuracy ok with 2 digit precision?

mkunzBGC:
@DonLakeFlyer Yes, I can test once a daily build is available.

DonLakeFlyer:
> he RTK icon will only show up when a vehicle is connected. Can you change that to always show up indepenent from a vehicle?

Yup, known problem (there an issue for it somewhere). The way the toolbar works it's a non-trivial change though.

DonLakeFlyer:
@mkunzBGC New daily builds with this should be available in about an hour.

mkunzBGC:
@DonLakeFlyer Today I tested QGroundControl Development HEAD:defdcdc41 2018-10-04 14:37:54 -0700, which includes the option to save and restore the base station position. My setup consisted of a Here+ RTK base and rover module, both running the 1.40 firmware. The rover was connected to a 3DR Pixhawk running ArduCopter 3.5.5. Rover and base antenna were set up under open sky, elevated 2 and ~1.2 m from the ground, respectively. They were statically mounted at a distance of ~5m horizontally and ~1.5m vertically. The base station saw 20 satellites all the time, the rover used 15-19.

During all testing I waited until RTK fixed mode before taking coordinates. I measured the position of the rover three times, restarting the rover and QGroundControl in between. Saving and restoring the base position worked as expected. The positions are within 2.5cm horizontally and 2 cm vertically to each other, which is a pretty good result.

However, the determined base station altitude was strange in that it differed from the rover by more than 40m (268m for the base, 223.4m for the rover). I am at 50.909°N and 11.565°E if that matters.

One suggestion: When saving the base position, several messages pop up: "Change of 'Base Position [Latitude/Longitude/Altitude/Accuracy]' value requires restart of QGroundControl to take effect". I think this is not necessary until the "Use specified base position" radio button is activated.

Thanks for adding this useful feature!

DonLakeFlyer:
@mkunzBGC Thanks fo much for testing.
@bkueng Any ideas on the altitude thing?

mkunzBGC:
@DonLakeFlyer The difference in altitude is likely related to the conversion between ellipsoid to geoid altitude as explained [here](http://www.gpspassion.com/forumsen/topic.asp?TOPIC_ID=10915). The NMEA stream from a different GPS receiver, captured at a location few kilometers away, indicates that the reference ellipsoid is 47.5 m higher than the geoid here. This fits nicely to the difference I saw in the base station altitude.

I imagine that the ublox chipset needs the ellipsoid altitude, but for the user interface it would be desirable to work with geoid altitude. If the base station outputs the altitude correction then QGroundControl could do a conversion that is invisible to the user. If this info is not available I would suggest a hint to the user right next to the field were the base station position is indicated.

bkueng:
@mkunzBGC Sorry I dropped this. You are correct: the vehicle reports the mean sealevel altitude, whereas the RTK base station outputs and uses the ellipsoid altitude.
Either solution is fine with me: add a remark to the UI or do the conversion.

DonLakeFlyer:
So "Base Position Altitude" as shown in image in previous command is WGS84. But what other value is AMSL from where specifically?

mkunzBGC:
"Base Position Altitude" is ellipsoid altitude, i.e. uses the coordinate system where the surface of the WGS84 reference ellipsoid is at altitude 0. The reference ellipsoid is a very simple representation of the earth and therefore deviates quite a bit from what we call mean sea level. GPS receivers include a lookup table that allows them to convert from ellipsoid altitude to altitude above mean sea level.  The NMEA message GGA includes this converted altitude in field 9 (alt) *and* the applied offset in field 11 (sep).
I guess that the lookup tables are different from receiver model to receiver model and could even change with firmware updates. However, I would assume that they are the same for base and rover. As the correction is (afaik) always applied to the rover, I would expect the most consistent results if they are also applied to the base station altitude.
@DonLakeFlyer Is the info from GGA field 11 available at the time when the user would save the base position, i.e. does the RTK base send this info? If not, use ellipsoid altitude and inform the user, e.g. "This is altitude above the WGS84 ellipsoid, not altitude above mean sea level"
If the info is available, the correction could be applied to the altitude. However, the base station internally uses ellipsoid altitude, so the question is how to load a saved position. I see three options:

1. Wait until the base has a good GPS fix so that it tells you the correct separation, allowing the conversion from the saved altitude to ellipsoid altitude.
2. Save the separation together with the base station coordinates. It is not enough to save this "silently" - it should be another field where the user can enter values.
3. Define the the position of the base station by lat, lon and ellipsoid altitude, but inform the user in a read-only way about the altitude above mean sea level.

I would prefer option 3.

bkueng:
Option 3 sounds good to me too. If we start to add offsets we introduce inaccuracies.
A fixed lookup-table would work too, but I'm not sure if there's a well-defined standard for it.

DonLakeFlyer:
> Define the the position of the base station by lat, lon and ellipsoid altitude

I get that I think. It's just label the altitude of "Base Position Altitude" as WGS84. So text changes to something like "Base Position Alt (WGS84)"?

>  inform the user in a read-only way about the altitude above mean sea level.

Inform exactly where?

mkunzBGC:
I made a mockup of an input mask that would imho be (1) intuitive (the user immediately knows which values need to be entered), (2) explicit (the user sees that two different altitude references exist and gets enough info to look up details if desired) and (3) error-proof (potential inconsistencies in altitude conversion do not influence the RTK solution because only the ellipsoid altitude is used):
inputmaskrtkbase3

DonLakeFlyer:
I'd rather not have to get into WGS84 to AMSL conversion unless I really need to. Does that value come the the GPS as well?

DonLakeFlyer:
So Lorenz is trying to figure out whether there is a WGS84 to AMSL conversion code available. At this point it appears to be ublox proprietary. Until then the best I can do is change the label on Base Position Altitude to state that it's WGS84.

mkunzBGC:
@DonLakeFlyer Does the base station send GGA messages to the host computer? If so, you can just use the "sep" value in field 11 for the conversion from ellipsoid to asl altitude. If the base station does not provide this NMEA sentence, is there another way to get the separation used by the ublox chipset? It is an approximation, so calculating the separation in QGroundControl will likely be inconsistent to what the ublox chipset does for the rover altitude.

DonLakeFlyer:
Those would be questions for @bkueng

bkueng:
> @DonLakeFlyer Does the base station send GGA messages to the host computer?

It uses NAV-PVT, available in QGC from here: https://github.com/mavlink/qgroundcontrol/blob/master/src/GPS/GPSProvider.cc#L162. It contains both altitudes. However these altitudes are the current measurement, not based on the survey-in, and thus less accurate.

PR #6912, opened 2018-10-08
Add motor test page for sub

Williangalvani:
This adds a motor test interface for Sub. This interface is important to test individual motors to check if they are connected correctly, diagnose issues and reverse motor direction when necessary.

This is what it looks like:
deepin-screen-recorder_select area_20181008111918

### DIFF
--- a/src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc
+++ b/src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc
@@ -86,7 +86,8 @@ const QVariantList& APMAutoPilotPlugin::vehicleComponents(void)
             _powerComponent->setupTriggerSignals();
             _components.append(QVariant::fromValue((VehicleComponent*)_powerComponent));
 
-            if (_vehicle->multiRotor() || _vehicle->vtol()) {
+            int versionInt = _vehicle->firmwareMajorVersion() * 100 + _vehicle->firmwareMinorVersion() * 10 + _vehicle->firmwarePatchVersion();
+            if (_vehicle->sub() && versionInt >= 353) {
                 _motorComponent = new MotorComponent(_vehicle, this);
                 _motorComponent->setupTriggerSignals();
                 _components.append(QVariant::fromValue((VehicleComponent*)_motorComponent));
--- a/src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc
+++ b/src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc
@@ -40,10 +40,7 @@ APMAutoPilotPlugin::APMAutoPilotPlugin(Vehicle* vehicle, QObject* parent)
     , _subFrameComponent        (NULL)
     , _flightModesComponent     (NULL)
     , _powerComponent           (NULL)
-#if 0
-        // Temporarily removed, waiting for new command implementation
     , _motorComponent           (NULL)
-#endif
     , _radioComponent           (NULL)
     , _safetyComponent          (NULL)
     , _sensorsComponent         (NULL)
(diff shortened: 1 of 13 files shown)


qgroundcontrol.qrc
```diff
@@ -53,13 +53,15 @@
         <file alias="MockLink.qml">src/ui/preferences/MockLink.qml</file>
         <file alias="MockLinkSettings.qml">src/ui/preferences/MockLinkSettings.qml</file>
         <file alias="MotorComponent.qml">src/AutoPilotPlugins/Common/MotorComponent.qml</file>
+        <file alias="APMSubMotorComponent.qml">src/AutoPilotPlugins/APM/APMSubMotorComponent.qml</file>
```
DonLakeFlyer: These should be in the firmware specific files.

src/AutoPilotPlugins/Common/MotorComponent.cc
```diff
@@ -48,7 +48,12 @@ QStringList MotorComponent::setupCompleteChangedTriggerList(void) const
 
 QUrl MotorComponent::setupSource(void) const
 {
-    return QUrl::fromUserInput(QStringLiteral("qrc:/qml/MotorComponent.qml"));
```
DonLakeFlyer: This should be left alone and the ArduPilot code should have it's own implementation.

src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc
```diff
@@ -89,15 +86,12 @@ const QVariantList& APMAutoPilotPlugin::vehicleComponents(void)
             _powerComponent->setupTriggerSignals();
             _components.append(QVariant::fromValue((VehicleComponent*)_powerComponent));
 
-#if 0
-    // Temporarily removed, waiting for new command implementation
-
-            if (_vehicle->multiRotor() || _vehicle->vtol()) {
-                _motorComponent = new MotorComponent(_vehicle, this);
+            int versionInt = _vehicle->firmwareMajorVersion() * 100 + _vehicle->firmwareMinorVersion() * 10 + _vehicle->firmwarePatchVersion();
```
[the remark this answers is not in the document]
Williangalvani: Or hopefully, https://github.com/mavlink/qgroundcontrol/pull/6913 gets merged soon

src/Vehicle/Vehicle.cc
```diff
@@ -3290,13 +3331,10 @@ void Vehicle::setSoloFirmware(bool soloFirmware)
     }
 }
 
-#if 0
-// Temporarily removed, waiting for new command implementation
-void Vehicle::motorTest(int motor, int percent, int timeoutSecs)
+void Vehicle::motorTest(int motor, int percent)
 {
-    doCommandLongUnverified(_defaultComponentId, MAV_CMD_DO_MOTOR_TEST, motor, MOTOR_TEST_THROTTLE_PERCENT, percent, timeoutSecs);
+    sendMavCommand(_defaultComponentId, MAV_CMD_DO_MOTOR_TEST, true, motor, MOTOR_TEST_THROTTLE_PERCENT, percent, 0, 0, MOTOR_TEST_ORDER_BOARD);
```
DonLakeFlyer: So is MOTOR_TEST_ORDER_BOARD the real connection order as opposed to the strange Mission Planner motor order that ArduPilot is using. Did you implement normal motor number ordering support in ArduPilot or only for ArduSub?

DonLakeFlyer: For example for multi-rotor motor numbering?

DonLakeFlyer:
Am I the only person that hates the GitHub Start a Review thing? Gets me every time.

patrickelectric:
> Am I the only person that hates the GitHub Start a Review thing? Gets me every time.

It got me here too..

src/Vehicle/Vehicle.cc
DonLakeFlyer: The reason I ask is because this is in the generic code. For example when PX4 implement motor test (using normal motor numbering) would this be the setting for them as well. Hence it is correct for all firmwares.

Williangalvani: Yeah, this is the real order.
IIRC @jaxxzer got this implemented into ArduSub only.

Williangalvani: I'm open to suggestions here.
We could expose the order to be used
`motorTest(int motor, int percent, MOTOR_TEST_ORDER order)`

DonLakeFlyer: > Yeah, this is the real order.

This is fine then

PR #6913, opened 2018-10-08
Create Helper functions to simplify version checking

Williangalvani:
This is to simplify cases where both major, minor, and patch versions had to be checked independently, which resulted in hard to read code

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSafetyComponentSummarySub.qml
+++ b/src/AutoPilotPlugins/APM/APMSafetyComponentSummarySub.qml
@@ -13,7 +13,7 @@ FactPanel {
     color:          qgcPal.windowShadeDark
 
     property var _activeVehicle:     QGroundControl.multiVehicleManager.activeVehicle
-    property bool _firmware34: _activeVehicle.firmwareMajorVersion == 3 && _activeVehicle.firmwareMinorVersion == 4
+    property bool _firmware34:       _activeVehicle.versionCompare(3, 5, 0) < 0
 
     QGCPalette { id: qgcPal; colorGroupEnabled: enabled }
     FactPanelController { id: controller; factPanel: panel }
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -740,3 +740,35 @@ void FirmwarePlugin::_versionFileDownloadFinished(QString& remoteFile, QString& 
         qgcApp()->showMessage(message.arg(vehicle->firmwareVersionTypeString(), currentVersion, version));
     }
 }
+
+int FirmwarePlugin::versionCompare(Vehicle* vehicle, int major, int minor, int patch)
+{
+    int currMajor = vehicle->firmwareMajorVersion();
+    int currMinor = vehicle->firmwareMinorVersion();
+    int currPatch = vehicle->firmwarePatchVersion();
+
+    if (currMajor == major && currMinor == minor && currPatch == patch) {
+        return 0;
+    }
+
+    if (currMajor > major
+       || (currMajor == major && currMinor > minor)
+       || (currMajor == major && currMinor == minor && currPatch > patch))
+    {
+        return 1;
+    }
+    return -1;
+}
+
+int FirmwarePlugin::versionCompare(Vehicle* vehicle, QString& compare)
+{
+    QStringList versionNumbers = compare.split(".");
+    if(versionNumbers.size() != 3) {
+        qCWarning(FirmwarePluginLog) << "Error parsing version number: wrong format";
+        return -1;
+    }
+    int major = versionNumbers[0].toInt();
+    int minor = versionNumbers[1].toInt();
+    int patch = versionNumbers[2].toInt();
+    return versionCompare(vehicle, major, minor, patch);
+}
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -713,25 +713,12 @@ void FirmwarePlugin::_versionFileDownloadFinished(QString& remoteFile, QString& 
     }
 
     qCDebug(FirmwarePluginLog) << "Latest stable version = "  << version;
-    QStringList versionNumbers = version.split(".");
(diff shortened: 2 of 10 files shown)


src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -740,3 +727,35 @@ void FirmwarePlugin::_versionFileDownloadFinished(QString& remoteFile, QString&
...
+    int currMajor = vehicle->firmwareMajorVersion();
+    int currMinor = vehicle->firmwareMinorVersion();
+    int currPatch = vehicle->firmwarePatchVersion();
+
+    if (currMajor == major && currMinor == minor && currPatch == patch) {
+        return 0;
+    }
+
+    if (currMajor > major
+       || (currMajor == major && currMinor > minor)
```
patrickelectric: =>

src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -740,3 +727,35 @@ void FirmwarePlugin::_versionFileDownloadFinished(QString& remoteFile, QString&
...
+       || (currMajor == major && currMinor == minor && currPatch > patch))
+    {
+        return 1;
+    }
+    return -1;
```
patrickelectric:
```
currMajor >= major
currMinor >= minor
currPatich > patch
```

src/Vehicle/Vehicle.h
```diff
@@ -744,6 +744,11 @@ class Vehicle : public FactGroup
     Q_INVOKABLE void triggerCamera(void);
     Q_INVOKABLE void sendPlan(QString planFile);
 
+    /// Used to check if running current version is equal or higher than the one being compared.
+    //  returns 1 if current > compare, 0 if current == compare, -1 if current < compare
+    Q_INVOKABLE int versionCompare(QString compare);
```
patrickelectric: QString&

src/FirmwarePlugin/FirmwarePlugin.h
```diff
@@ -309,6 +309,11 @@ class FirmwarePlugin : public QObject
     /// Used to check if running firmware is latest stable version.
     virtual void checkIfIsLatestStable(Vehicle* vehicle);
 
+    /// Used to check if running current version is equal or higher than the one being compared.
+    /// returns 1 if current > compare, 0 if current == compare, -1 if current < compare
+    int versionCompare(Vehicle* vehicle, QString compare);
```
patrickelectric: QString&

src/FirmwarePlugin/FirmwarePlugin.cc
patrickelectric: =>

src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -740,3 +727,35 @@ void FirmwarePlugin::_versionFileDownloadFinished(QString& remoteFile, QString&
...
+    if (currMajor == major && currMinor == minor && currPatch == patch) {
+        return 0;
+    }
+
+    if (currMajor > major
+       || (currMajor == major && currMinor > minor)
+       || (currMajor == major && currMinor == minor && currPatch > patch))
+    {
+        return 1;
+    }
```
patrickelectric: Is it more simple to compare:
```
currMajor >= major
currMinor >= minor
currPatch > patch
```

src/FirmwarePlugin/FirmwarePlugin.h
patrickelectric: QString&

src/Vehicle/Vehicle.h
patrickelectric: QString&

DonLakeFlyer:
Can you change this to disregard dev versions? If the user is using a dev version then I think all bets are off. It's hard to say what they may have done or want to do.

src/FirmwarePlugin/FirmwarePlugin.cc
Williangalvani: could you elaborate?
Doesn't look like it would work comparing 4.0.0 with 3.0.0

Williangalvani:
Hey Don,

This is currently used for selecting metadata files or hiding qml sections from firmware that don't support it, I'd prefer to keep the dev versions check apart (like on the "latest release" check).

How do you see dev version checking in this function? should it return that it is always newer?

jaxxzer:
I'd like to point out that this is a non-functional change and we have not been filtering for dev prior. This is only refactoring for readability.

The dev check can still be put in place where needed, and this helper just helps you sort out the x.x.x part.

DonLakeFlyer:
> This is currently used for selecting metadata files or hiding qml sections from firmware that don't support it, I'd prefer to keep the dev versions check apart (like on the "latest release" check).
> 
> How do you see dev version checking in this function? should it return that it is always newer?

Sorry, didn't realize that. I think a dev version should always be newer than whatever is current stable.

DonLakeFlyer:
So as far as I'm concerned this is ok to go. Should I merge?

Williangalvani:
just let me rebase and add the function to the motor test stuff

src/FirmwarePlugin/FirmwarePlugin.cc
patrickelectric: I need to stop reviewing stuff at night.

Williangalvani:
done

src/AutoPilotPlugins/APM/APMSafetyComponentSummarySub.qml
```diff
@@ -13,7 +13,7 @@ FactPanel {
     color:          qgcPal.windowShadeDark
 
     property var _activeVehicle:     QGroundControl.multiVehicleManager.activeVehicle
-    property bool _firmware34: _activeVehicle.firmwareMajorVersion == 3 && _activeVehicle.firmwareMinorVersion == 4
+    property bool _firmware34:       _activeVehicle.versionCompare(3, 5, 0) < 0
```
jaxxzer: aye, this is not equivalent! should be _activeVehicle.versionCompare(3, 5, 0) == 0

@Williangalvani

PR #6923, opened 2018-10-15
Update PX4 Firmware metadata Mon Oct 15 13:47:56 UTC 2018

dagar:
Latest auto generated PX4 airframe metadata. Adds the X-UAV Mini Talon.

### DIFF
--- a/src/AutoPilotPlugins/PX4/AirframeFactMetaData.xml
+++ b/src/AutoPilotPlugins/PX4/AirframeFactMetaData.xml
@@ -579,6 +579,27 @@
       <output name="AUX3">feed-through of RC AUX3 channel</output>
     </airframe>
   </airframe_group>
+  <airframe_group image="AirframeUnknown" name="Plane V-Tail">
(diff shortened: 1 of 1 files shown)


dagar:
@tony3dr fyi

dagar:
@fredowski @hamishwillee - something's not quite right with the metadata. We're missing the Plane V-Tail image.
https://github.com/mavlink/qgroundcontrol/pull/6852

We'll need to get this sorted out across Firmware, user guide, QGC.

dagar:
This is safe to merge, but the core issue still needs to be looked at.

dagar:
> What is the procedure to introduce the image file?

Good question. I don't know, but let's find out.

As for the metadata we should let the bot update it automatically from the source.

fredowski:
Maybe the bot will keep the image info once it is there. I will make a new PR with the image name added.

fredowski:
I opened a new PR #6925 which fixes the autogenerated code. It adds the image and fixes the class info.

PR #6925, opened 2018-10-16
fixed group image for Plane V-Tail group and talon class

fredowski:
The generated patch from the PX4BuildBot in commit dc8549ace8e04e53 from PR #6923 did not include the image for the Plane V-Tail group. The class was also malformed. This patch fixes the problems from the autogenerated code. 

The code was already o.k. from PR #6852 but that code was later removed
from the PX4BuildBot because the mini talon frame was not in the codebase at
that point of time.

So hopefully the buildbot will leave the image in there...

I opened a PR PX4/Firmware#10699 to avoid the malformed class info.

### DIFF
--- a/src/AutoPilotPlugins/PX4/AirframeFactMetaData.xml
+++ b/src/AutoPilotPlugins/PX4/AirframeFactMetaData.xml
@@ -579,12 +579,9 @@
       <output name="AUX3">feed-through of RC AUX3 channel</output>
     </airframe>
   </airframe_group>
-  <airframe_group image="AirframeUnknown" name="Plane V-Tail">
+  <airframe_group image="PlaneVTail" name="Plane V-Tail">
     <airframe id="2200" maintainer="Friedrich Beckmann &lt;friedrich.beckmann@hs-augsburg.de&gt;" name="X-UAV Mini Talon">
-      <class>Plane
-The Mini Talon does not have a wheel and
-no flaps. I leave them here because the mixer
-computes also wheel and flap controls.</class>
+      <class>Plane</class>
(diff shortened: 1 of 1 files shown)


dagar:
I think it needs to be fixed by the generator.

@hamishwillee

hamishwillee:
> I think it needs to be fixed by the generator.

@dagar The generation of correct metadata (ie addition of image PlaneVTail.jpg in XML and markdown) done in https://github.com/PX4/Firmware/pull/10717. I presume supply of correct image is managed by some other path (?) ie AM I expected to manually add the image to Dev and User guide or do these come in from elsewhere (ditto for QGC - I assume it is getting image from another path?)

@fredowski In case it isn't clear, @dagar is saying that this PR should be closed because the data should come in correct from PX4.

fredowski:
Thanks for clarification how the procedure for the a new airframe with image is.

Add airframe in init.d - PR PX4/Firmware#10414
Update botscript to include the image info for the airframe group -  PX4/Firmware#10717
Upload image to devguide - PX4/Devguide#634
Upload image to userguide - PX4/px4_user_guide#369
Upload image to qgc - PR mavlink/qgroundcontrol#6852

Thanks everybody!

hamishwillee:
@dagar Are the last three correct? i.e. can you confirm that there is no automated mechanism to push the images to the devguide, user guide and QGC? This is just FMI

@fredowski Just FYI "Update botscript to include the image info" is only required for a new airframe group - in this case Plane V-Tail. If you were to add another airframe with this same group you wouldn't need to update the script. 

If the image in a group is unknown then the script will display a placeholder image - you can see this if you scroll down from your [airframe to simulation](http://dev.px4.io/en/airframes/airframe_reference.html#plane-v-tail)

Thanks for your patience. It has been a while since I had to add a new group.

PR #6942, opened 2018-10-25
AndroidManifest.xml increase targetSdkVersion 19 -> 26

dagar:
This is a Google Play store requirement as of November 1st, 2018.
https://android-developers.googleblog.com/2017/12/improving-app-security-and-performance.html

### DIFF
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -48,7 +48,7 @@
             <!-- Background running -->
         </activity>
     </application>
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
What about the min sdk version? Is that out of date with reality as well?

dagar:
With Qt I believe our min sdk hardly matters. We could bump it to filter out ancient or questionable devices.

https://developer.android.com/about/dashboards/

DonLakeFlyer:
Should this be merged then?

PR #6948, opened 2018-10-25
Use MAV_COMP_ID_AUTOPILOT1 instead of MAV_COMP_ID_MISSIONPLANNER

### LINKED ISSUE #2907: Wrong target component ID
Target component should be the vehicle's comp_id (same as received on its heartbeat) and not MAV_COMP_ID_MISSIONPLANNER.

Example:
https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/MissionManager.cc#L123

MAV_COMP_ID_MISSIONPLANNER is the component ID for the GCS MissionPlanner.

DonLakeFlyer:
Fixes #2907

### DIFF
--- a/src/MissionManager/PlanManager.cc
+++ b/src/MissionManager/PlanManager.cc
@@ -114,7 +114,7 @@ void PlanManager::_writeMissionCount(void)
                                         _dedicatedLink->mavlinkChannel(),
                                         &message,
                                         _vehicle->id(),
-                                        MAV_COMP_ID_MISSIONPLANNER,
+                                        MAV_COMP_ID_AUTOPILOT1,
                                         _writeMissionItems.count(),
                                         _planType);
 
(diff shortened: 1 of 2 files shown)


hamishwillee:
@edwinhayes You might want to test this change in tomorrow's daily build - see if it fixes #2907

edwinhayes:
@hamishwillee Thanks.  Should be able to confirm on Monday.

edwinhayes:
@hamishwillee Still seems to be broken, but the date in the titlebar is still 2018-09-06.  Are the 'daily build' downloads actually built daily?  Can I check the build status somehow?  I'm on a Mac FYI, so could be specific to that?

hamishwillee:
@edwinhayes I don't know! @DonLakeFlyer may be able to confirm whether there is something odd going on with the builds.

DonLakeFlyer:
@dagar Just saw the OSX daily build drops moved to Jenkins. Looks like they are not updating maybe.

edwinhayes:
Tested and working.  Thanks guys.

hamishwillee:
Thanks for testing @edwinhayes !!

@DonLakeFlyer Thanks very much. Nice to see Feb 2016 bug getting fixed :-)

PR #6950, opened 2018-10-25
MavLinkProtocol: Fix constructor warnings

patrickelectric:
This fix more than 8k warnings messages.
https://api.travis-ci.org/v3/job/446387610/log.txt
screenshot from 2018-10-25 19-06-51

### DIFF
--- a/src/comm/MAVLinkProtocol.cc
+++ b/src/comm/MAVLinkProtocol.cc
@@ -53,6 +53,8 @@ const char* MAVLinkProtocol::_logFileExtension = "mavlink";             ///< Ext
 MAVLinkProtocol::MAVLinkProtocol(QGCApplication* app, QGCToolbox* toolbox)
     : QGCTool(app, toolbox)
     , m_enable_version_check(true)
+    , _message({})
+    , _status({})
     , versionMismatchIgnore(false)
     , systemId(255)
     , _current_version(100)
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
But that removes initialization doesn't it. The current code sets the whole thing to 0 doesn't it?

patrickelectric:
https://github.com/pixhawk/mavconn/blob/master/src/comm/lcm/mavlink_message_t.hpp#L40
yep, no initialization.

patrickelectric:
Oh it's a struct, and g++ things that we want to set the default values with:
{.checksum = 0, .magic = 0, ...}

patrickelectric:
@DonLakeFlyer 	Sorry and fixed !

patrickelectric:
@DonLakeFlyer 	Fixed.

DonLakeFlyer:
I don't know much about this new-fangled C++ stuff. You've verified that zeros out the structure?

dogmaphobic:
Oops, I just "fixed" this during my dealings with Android earlier today. It was spewing tons of warnings with the Android compiler so I reverted to _old school_ initialization.

https://github.com/mavlink/qgroundcontrol/pull/6956

patrickelectric:
@dogmaphobic you can do just {0} in the initialization list, like I'm doing in this PR. memset is not necessary.

dogmaphobic:
Like Don, I'm suspicious of _new-fangled C++ stuff_. I started with straight assembly in the 70s and that's all I used for a decade (yes, I'm that old but I started programing when I was 13.) For years I thought _C_ was for lazy people. Don't get me started with c++ :P 

I'm (half) joking. Given how badly implemented these new things are across all platforms, I just feel safer manually doing what I want done. Especially when I need those structures to be zeroed out.

Issue #6939, opened 2018-10-24
QGC Daily Build (Android) Crashes on start


b14ckyy:
QGC App in Daily Build for Android crashes on start since at least all versions since Oct 18th. 

Crash Reports submitted via Android function. A few minutes ago. 

Android Version: 8.0.0
Phone: Xperia XZ
Firmware Build: 41.3.A.2.184
Security Patch: Sept. 1st 2018
Kernel: 3.18.66

GGC Build: 3.4.4-325.g4885d2f50

Behavior:
- App starts with white screen and Title Bar
- GPS is active and tries to fetch position
- Crash Popup appears and app closes

Steps tried to workaround:
- app data deleted > not solved
- reinstalled > not solved
- Disabled Location Services > not solved
- Flight Mode > not solved

Not yet tested on another phone.

PR #6956, opened 2018-10-26
Android crash fix

dogmaphobic:
This fixes the crash on boot on Android (any mobile actually).

This would only happen the first time you were to run QGC on a device. 

Also getting rid of some build warnings while at it.


DonLakeFlyer:
Fixes #6939

PR #6958, opened 2018-10-27
Move to pragma once

patrickelectric:
Move from ifdef to pragma once


dogmaphobic:
As you are already changing a pile of files, you might as well do a `s/(c) 2009-2016/(c) 2009-2018/g;` (no escapes but you get the idea) through all .cc/.cpp/.h/.qml. I keep meaning to do it and never get around to it.

patrickelectric:
@dogmaphobic 	I did the changes only in the files in this PR, will try to do the same across the next PRs.

patrickelectric:
It appears that the jenkins machine is full
`No space left on device`

DonLakeFlyer:
@dagar Jenkins failure

PR #6961, opened 2018-10-28
Style tweaks

Williangalvani:
This applies a few changes in style in a attempt to standardize things.
Suggestions are welcome!

 ## Make checkboxes darker when disabled:
before:

later:

 ## Make combobox follow color style of textfields
before:

after:

## Make disabled textfields have no border, and black text on clear background in dark theme
before:

after:

The safety pages look like this, then (sorry for the bad image alignment):
oldblack

oldwhite


DonLakeFlyer:
All looks nice. What about disabled combo boxes. Do they have the outer border removed as well like text fields?

Williangalvani:
> All looks nice. What about disabled combo boxes. Do they have the outer border removed as well like text fields?

I didn't think of that, I'll try it and report back here

Williangalvani:
@DonLakeFlyer ,Yes, that is removed too:
(disregard the white line, it's a mistake I made editing the image)
before:
before
after:
after

DonLakeFlyer:
OK, great. Thanks.

Issue #5215, opened 2017-05-27
Survey grid still go out of bounds


lvale:
When planning a survey grid where there are angles on the survey area, QGC does not plan for points within the survey area. Screenshot represents that, where the vehicle would fly above houses, and this is a issue with privacy

latest master:

<img width="1435" alt="screen shot 2017-05-27 at 05 05 09" src="https://cloud.githubusercontent.com/assets/3130363/26517898/cd4797f6-429a-11e7-93f7-7cb22282d7b7.png">

lvale:
Also with different survey angle, note a NaN on the Survey Area

<img width="1435" alt="screen shot 2017-05-27 at 05 22 08" src="https://cloud.githubusercontent.com/assets/3130363/26517963/a5085ed6-429c-11e7-98d2-c5ee9e60fbf0.png">

DonLakeFlyer:
The first image is on purpose. Survey does not yet support concave polygons (or is that convex). Anyway it's the one you drew!

The last one is a bug.

PR #6962, opened 2018-10-29
Feature: Concave Polygon Support by Splitting into Sub-Polygons

### LINKED ISSUE #5830: Survey: Support concave polygons
Currently only convex are supported.

thomasgubler:
This PR adds a new mode for planning a survey which makes the planned flight path stay within concave polygons.
This is achieved by splitting the polygon into concave sub polygons and treating each as an independent area to be surveyed. Because this introduces additional turns, the flight time might be increased. Hence, the original (current) mode that does not support concave polygons is preserved to give user the option to chose the mode they like.

Original mode (stays default after this PR):
screenshot from 2018-10-29 08-30-29

New mode that splits polygons:
screenshot from 2018-10-29 08-31-01

The UI work for toggling between the modes was done by @dogmaphobic 

@dogmaphobic this is a cherrypick of the original branch on top of master

Fixes #5830 #5215

### DIFF
--- a/src/MissionManager/SurveyComplexItem.cc
+++ b/src/MissionManager/SurveyComplexItem.cc
@@ -1115,7 +1115,7 @@ void SurveyComplexItem::_rebuildTransectsPhase1Worker(bool refly)
     }
 
     // Create list of separate polygons
-    QList<QPolygonF> polygons;
+    QList<QPolygonF> polygons{};
     _PolygonDecomposeConvex(polygon, polygons);
 
     // iterate over polygons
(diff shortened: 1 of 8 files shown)


thomasgubler:
@dagar Jenkins fails because of disk space
```
remote file operation failed: /tmp/jenkins/workspace/ontrol_feature_concave_poly-OHOX4QDIGQQT2D5UXRX55AB7A57QEOQTLCBZL7KRVOGAALDJXIPQ at hudson.remoting.Channel@34016e9d:ec2_docker_slave (i-0c530ec0b6c989cc5): java.nio.file.FileSystemException: /tmp/jenkins/workspace/ontrol_feature_concave_poly-OHOX4QDIGQQT2D5UXRX55AB7A57QEOQTLCBZL7KRVOGAALDJXIPQ: No space left on device
```

dagar:
Zombie slave strikes again. Restarting...

DonLakeFlyer:
@thomasgubler Awesome change. Will take me a little to look through.

dogmaphobic:
The idea was to come up with an algorithm for path planning. As @thomasgubler found out, this is a rather hairy proposition. This interim approach makes sure the vehicle does not fly outside the defined polygon boundary but causes multiple turns as it flies each segment separately.

Here are a couple of papers describing solutions: 

https://pdfs.semanticscholar.org/2b0b/20ed4e861fcdad1a27ab9cebf59f25bca629.pdf

https://www.sciencedirect.com/science/article/pii/S0957417416300306

It just needs translating them into code :P

thomasgubler:
> @thomasgubler New features like this also need an update to changelog.md

done

thomasgubler:
rebased on top of master

DonLakeFlyer:
Can you add some comments to the decompose polygon code which explains how it works? Or maybe refer to some article URL which has an algorithem explanation? I'm going to have to maintain this going forward so I need all the help I can get.

Another question is I don't quite understand how essentially the invididual survey sections are connected together. How is this done in such a way that the connectinh flight path never crosses the excluded area. Also is that connecting flight path always generated in such a way that it is the shortest legal flight path?

thomasgubler:
> Another question is I don't quite understand how essentially the invididual survey sections are connected together. How is this done in such a way that the connectinh flight path never crosses the excluded area. Also is that connecting flight path always generated in such a way that it is the shortest legal flight path?

Good point. There is no special code handling that. I initially planned on adding that but then I noticed when testing that with most (meaningful survey shapes) this happens already automatically. That being said, you can get it to connect the polygons outside of the main polygon but in most cases slightly changing the polygon will fix it. Hence I did not spend time on it. So I think it should be useful for a typical use case already like this unless a user tries to really setup some strangely shaped polygons.

thomasgubler:
@DonLakeFlyer I added a link to the algorithm description, cleaned up some debug code  and rebased on master

DonLakeFlyer:
> when testing that with most (meaningful survey shapes) this happens already automatically.

My testing shows the same thing which I think makes this good enough.

DonLakeFlyer:
Some changes:
* The setting for Split Concave polygons should just be a SettingsFact inside a Survey
* It should not be an AppSettings
* The Split Concave SettingsFact should default to true
* The version number of Survey item needs to be bumped up by 1
* The Split Concave setting needs to be saved/loaded to/from the Survey item json
* When loading a old version Survey item, Split Concave should be set to false

thomasgubler:
@dogmaphobic If I understand this correctly the requested changes are all about the setting option that you implemented. I think it would be very inefficient if I try to fix that as I have zero knowledge about all that code. Could you have a look please? (Not urgent compared to all the other ongoing things)

dogmaphobic:
> If I understand this correctly the requested changes are all about the setting option that you implemented.

Sorry, I wasn't paying attention. Yes, it makes a lot more sense for me to look into this. I'll take a look.

dogmaphobic:
> Sorry, I wasn't paying attention. Yes, it makes a lot more sense for me to look into this. I'll take a look.

Done

src/MissionManager/SurveyComplexItem.cc
```diff
@@ -196,6 +201,7 @@ bool SurveyComplexItem::_loadV4(const QJsonObject& complexObject, int sequenceNu
 
     _gridAngleFact.setRawValue              (complexObject[_jsonGridAngleKey].toDouble());
     _flyAlternateTransectsFact.setRawValue  (complexObject[_jsonFlyAlternateTransectsKey].toBool(false));
+    _splitConcavePolygonsFact.setRawValue   (complexObject[_jsonSplitConcavePolygonsKey].toBool(true));
```
DonLakeFlyer: I believe this will assert if the key is missing (which it will be in older files). How to handle this will depend on how you handle the version bump from 4 to 5. Following how I handled this In other places with a similar version bump:
* Save sets version to 5
* Load: Make the key required if version=5
* Load: Do a contains check before accessing on load
* Load: If not there, set to false

src/MissionManager/SurveyComplexItem.cc
```diff
@@ -1087,7 +1093,7 @@ void SurveyComplexItem::_rebuildTransectsPhase1WorkerSinglePolygon(bool refly)
     if (_loadedMissionItemsParent) {
         _loadedMissionItems.clear();
         _loadedMissionItemsParent->deleteLater();
-        _loadedMissionItemsParent = NULL;
+        _loadedMissionItemsParent = nullptr;
```
DonLakeFlyer: I've been using the Q_NULLPTR thing instead of NULL. I think we should come to some concensus on that. I'm hoping Q_NULLPTR would be more future proof than a direct C++ thing?

src/MissionManager/SurveyComplexItem.cc
dogmaphobic: >I hate this f'ing git review thing

Tell me about it. I get completely lost once you get to more than a couple of comments.

>If you look at FixedWingLandingComplexItem you can see an example it a complex item which has gone through many version bumps and how it was handled on load.

Will do

src/MissionManager/SurveyComplexItem.cc
dogmaphobic: I agree. At some point I will look into a global search and replace...

DonLakeFlyer:
Thanks everybody

Gongyihang:
Screenshot from 2019-12-27 10-47-45
Thank you for your previous work
I have noticed that you have optimized the case that the upper and lower boundaries are concave, and the left and right boundaries have not been optimized. I hope you can do the same for the left and right boundaries!
Looking forward to your reply!

DonLakeFlyer:
> I have noticed that you have optimized the case that the upper and lower boundaries are concave, and the left and right boundaries have not been optimized.

That is just a fluke of the current implementation. Concave polygons is still not supported. Unforunately the algorithm in this pull does not work and was disabled.

Issue #6914, opened 2018-10-09
Ambiguous altitude references


hamishwillee:
Can we use specific frame rather than "Abs" in altitude references throughout - ie MSL, WGS84 and AGL?

For example,  consider the waypoint altitude editor:
- "Abs" radio button becomes "WGS84"
  - Remove the comment no comment (minimally remove "Absolute" as this is redundant).
- For the AGL option it says *calculated from terrain data / Abs Alt 530.00 m*. However it is not clear whether the alt is MSL or WGS84. Would be better as (assuming MSL): 
* MSL alt 530.00 m calculated from terrain data*

 ![image](https://user-images.githubusercontent.com/5368500/46641464-7828d700-cbbd-11e8-8b9b-aadcbfa3d1ba.png)

I do appreciate that the absolute reference can be determined from the comments, but users won't necessarily read those carefully enough. Better to remove any ambiguity early.

jacobsenmd:
I strongly support this change. In the aviation world the term "absolute altitude" means AGL altitude. The mavlink ecosystem has used this term incorrectly for years to mean MSL or WGS84 (it is ambiguous which). The incorrect use of terminology will present an increasingly serious problem as UAS are integrated into national airspace systems, so now is the right time for QGC and other components of the ecosystem to adopt standard aviation conventions. Eliminating the word 'absolute' and replacing it with specific frames is an important step.

See: https://www.thebalancecareers.com/types-of-altitudes-in-aviation-282941.

Antiheavy:
+1 for ditching "Abs" and replacing with "MSL".

However, this discussion could quickly expand into the differences between WGS84, "MSL", "amsl", EGM96, EGM2008, and other GNSS receiver specific altitude references.  This is a complicated topic and the ambiguity may still remain.  Importantly, different autopilot firmware and GNSS receivers use different altitude datums and assumptions.  @priseborough any comments?

DonLakeFlyer:
> please don't invent "QGS84"

That's just a typo on WGS84.

AndKe:
I know, I was trying to be funny, and failed.  :)

DonLakeFlyer:
Ah, now I get it. Duh!

hamishwillee:
Hi @Antiheavy , @AndKe 

> +1 for ditching "Abs" and replacing with "MSL".

To be very precise, we want to replace Abs with whatever the global frame actually "is". In the example it is WGS84 (not MSL). 

[...]

PR #6986, opened 2018-11-04
Use correct terminology for Altitude values

DonLakeFlyer:
screen shot 2018-11-03 at 12 50 07 pm
screen shot 2018-11-03 at 12 50 25 pm
screen shot 2018-11-03 at 12 50 31 pm

Related to #6914


hamishwillee:
Thanks @DonLakeFlyer . 

@jacobsenmd FYI, "Abs" is now "AMSL" in QGC UI - so main issue fixed. As discussed this value is not necessarily precisely AMSL (it is whatever UBlox GPS is spitting out), but this is closest match we can get for now.

PR #6990, opened 2018-11-07
RTSP UDP timeout

dogmaphobic:
Moving the timeout to a (protected) variable so derived classes can access it.
Cleaning up some c++ warnings while at it.


dogmaphobic:
Jenkins failed because new files were added to the (qmake) project but not to the (cmake) project it uses. How do we keep these things in sync?

DonLakeFlyer:
Which is why I don't get the cmake thing.

PR #6993, opened 2018-11-07
Parameter Editor: Support editing of non-default component params

### LINKED ISSUE #6964: Parameter Editor: Doesn't show parameters from all components
Only shows default component.

DonLakeFlyer:
Fix #6964

screen shot 2018-11-07 at 12 42 04 pm


DonLakeFlyer:
This got trashed a while back and was never put back.

edwinhayes:
Is this supposed to be complete/working?  For me, the behaviour on daily build is a little weird: clicking on the first non-default component in the menu on the RHS works as expected, but then if you click on _another_ non-default component the parameters don't update.

DonLakeFlyer:
Can you move that into a separate issue?

edwinhayes:
Yep, just wanted to check it wasn't known unfinished first.

PR #6994, opened 2018-11-08
PreFlight: Reset not working correctly

DonLakeFlyer:
It would leave the last group expanded. 

@Williangalvani In this pull I turned off version checks for everythiong other than stable firmware builds. It was popping up all over the place with various SITL testing. Reason being non stable builds tend to have version strangeness.


Williangalvani:
Ok, no problem.

PR #7007, opened 2018-11-14
Survey, CorridorScan: Add camera start to transect entry even when taking images throughout…

DonLakeFlyer:
Problem: 
* You are flying a fixed wing survey
* You are in the middle of flying a large survey area and you fly out of comms range
* This will initiate an RTL (which will stop the camera)
* You quickly reposition your antenna and regain comms
* You put the vehicle back into Mission flight mode to continue the mission

If the survey was generated with "Take images in turnaround" on the rest of your survey is trashed since the camera will not be restart and no images will be taken.

Solution: Add a camera start command to the beginning of each transect no matter what the setting being used. This way at most you lose part of a transect but the entire survey isn't a loss.

While in the code I fixed a number of problem relating to trigger distance = 0 which should not add any camera commands at all.


DonLakeFlyer:
@antiheavy This will need some good testing of various combinations.

PR #7008, opened 2018-11-15
Add tuning page for Sub

Williangalvani:
This adds a tuning page for submarines:

peek 2018-11-15 12-09


DonLakeFlyer:
Seems odd that the FactTextField uses different visuals. Makes it unclear that you can click on it.

Williangalvani:
@DonLakeFlyer I removed the custom styles:
subtuning4
The slider handle is looking huge, but I worry a smaller one might be bad for mobiles.

DonLakeFlyer:
Seems fine for now.

PR #7016, opened 2018-11-17
New Resume Mission sequence

DonLakeFlyer:
The current Resume Mission / Battery Change sequence is confusing a lot of people for various reasons. The new sequence place Resume Mission in the Flight Plan complete dialog instead of using confirm sliders:

screen shot 2018-11-17 at 12 46 32 pm

And then if you do rebuild the mission, afterwards you just get the regular Start Mission slider.

### DIFF
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -119,7 +119,6 @@ QGCView {
 
     Connections {
         target:                     _missionController
-        onResumeMissionReady:       guidedActionsController.confirmAction(guidedActionsController.actionResumeMissionReady)
         onResumeMissionUploadFail:  guidedActionsController.confirmAction(guidedActionsController.actionResumeMissionUploadFail)
     }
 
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@antiheavy @chenny408 FYI

Antiheavy:
What happens if “resume mission from X” button is pressed when the vehicle is already disconnected (e.g. in the middle of a battery swap or lost Comms after landing)?

Is there a way to confirm that the replanned mission did indeed make it up to the vehicle before “start mission” again?

DonLakeFlyer:
> What happens if “resume mission from X” button is pressed when the vehicle is already disconnected (e.g. in the middle of a battery swap or lost Comms after landing)?

Upload will fail. And you go into a Resume Mission Upload failed sequence which allows you to upload again. That said it's kind of crappy experience. I think I can make that better as well.

> Is there a way to confirm that the replanned mission did indeed make it up to the vehicle before “start mission” again?

Basically same answer as above. If Resume Mission upload fails it's a special error case.

Antiheavy:
@m-Skelton

PR #7017, opened 2018-11-17
ArduPilot: Don't starts streams from QGC side. Assume firmware side does the right thing

DonLakeFlyer:
Related to #7104

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -587,18 +587,6 @@ void APMFirmwarePlugin::_adjustCalibrationMessageSeverity(mavlink_message_t* mes
     mavlink_msg_statustext_encode_chan(message->sysid, message->compid, 0, message, &statusText);
 }
 
-void APMFirmwarePlugin::initializeStreamRates(Vehicle* vehicle)
-{
-    vehicle->requestDataStream(MAV_DATA_STREAM_RAW_SENSORS,     2);
-    vehicle->requestDataStream(MAV_DATA_STREAM_EXTENDED_STATUS, 2);
-    vehicle->requestDataStream(MAV_DATA_STREAM_RC_CHANNELS,     2);
-    vehicle->requestDataStream(MAV_DATA_STREAM_POSITION,        3);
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
@jaxxzer This assumes ArduSub doesn't need them either. Ok?

edwinhayes:
Oh, is this why QGC daily build is not longer streaming for me?!  What is the reason for this?

tridge:
@DonLakeFlyer this will break ArduPilot badly. We don't yet support message intervals, although the PR for that is likely to go in soon. We will need the stream rates set for quite a while though, for all the stable releases

jaxxzer:
> this will break ArduPilot badly.

It looks like [Plane](https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/GCS_Mavlink.cpp#L520), Rover, Tracker stream by default. [Copter](https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/GCS_Mavlink.cpp#L347) and Sub do not. They should though.

If the idea is that the user will have to select their own streamrates depending on their link, then they may still do that via the parameters menu. Not necessarily 'broken' as much as forcing this configuration step on the user instead of making assumptions for them about their application.

DonLakeFlyer:
> It looks like [Plane](https://github.com/ArduPilot/ardupilot/blob/master/ArduPlane/GCS_Mavlink.cpp#L520), Rover, Tracker stream by default.

Not quite. For example Plane streams by default but not at a rate appropriate for a gcs link. So gcs still needs to setup for that as well.

DonLakeFlyer:
> They should though.

I thought so to. And in all honesty I still don't quite get why that can't be done. You'll have to talk with @tridge about that part. For now I'm going to quickly add an option which allows the end user to turn off GCS requesting stream starts.

jaxxzer:
> Not quite. For example Plane streams by default but not at a rate appropriate for a gcs link. So gcs still needs to setup for that as well.

True. This was a lapse on my part when I tested this. I was not working with fresh firmware defaults!

DonLakeFlyer:
@noircogi FYI

Noircogi:
Thanks.  I saw the discussion with Tridge.  As long as there is a way to either specify the values in QGC or opt out of the initialization, my problem will be addressed.

PR #7022, opened 2018-11-19
Flight Plan Complete: Better handling of comm last sequence

DonLakeFlyer:
If connection is lost on active vehicle and there is the possibility of a Resume Mission you get this:
screen shot 2018-11-19 at 11 04 20 am

Which achieves two things:
* Even more hinting as to how to do a Resume Mission
* Prevents user from clicking on the Flight Plan complete buttons while there is not vehicle really there

If the connect comes back it will go back to normal Flight Plan Complete stuff. If the connection closes the Flight Plan Complete dialog closes as well.


DonLakeFlyer:
@antiheavy More work to add some sanity to Resume Mission

PR #7026, opened 2018-11-20
Allow Safety page editing while in flight

DonLakeFlyer:


Antiheavy:
tested to work great in SITL

PR #7034, opened 2018-11-23
Flash: Add support for board/firmware detection using ArduPilot port naming

DonLakeFlyer:
If the board shows up using the ArduPilot port naming scheme then that will determine the board type as well as the firmware directly to load from. Otherwise it falls back to the board id detection mechanism.

This pull also changes the order of the ArduPilot firmware list to show ChibiOS build first.


DonLakeFlyer:
Note: QGC still only supports flashing through the standard bootloader. It can't flash naked ChibiOS boards.

PR #7039, opened 2018-11-27
AppVeyor: Store symbol artifacts only for Stable builds

DonLakeFlyer:
@dagar Does this look right? YAML is new to me.


DonLakeFlyer:
Need to keep artifact storage on AppVeyor from blowing up until they can fix their side of things.

PR #7041, opened 2018-11-27
Handle mismatched range for RSSI values.

### LINKED ISSUE #7013: QGC on IOS version 3.2 from App store reports RXRSSI incorrectly.
The reported value appears to be a the raw 8-bit number with a percentage.  Ie:  a true RSSI of 99% shows up as 254%.  Lower values appear as 248% etc.

dogmaphobic:
APM uses 0-255 to indicate RSSI while QGC expects a range of 0-100 (%). QGC also expects 255 to indicate the lack of RSSI altogether but this is not available on APM.

I have no way to test this so it would help if people gave a closer look.

Fixes #7013


DonLakeFlyer:
Thanks Gus

dogmaphobic:
Ok then, it looks like this solved it. Let me know if there are still places where this sort of thing doesn't work.

Issue #7046, opened 2018-11-29
Builds but does not run on Windows 10


kossumov:
I'm on Windows 10 x64
- I've installed: 
  - Qt Creator, 
  - Visual Studio 2015 (with c++ compilers), 
  - Windows SDK
- Cloned the project
- Did everything according to https://dev.qgroundcontrol.com/en/getting_started/

Builds fine
Now, I want to run the project

## Expected Behavior
Runs fine and I see qgroundcontroll UI

## Current Behavior
Crashes  (doesn't run)
Error in C:\Qt\5.11.0\msvc2015\include\QtCore\qobject.h (line 250)

## Steps to Reproduce:
1. Take a clean computer with Windows 10
2. Do everything according to https://dev.qgroundcontrol.com/en/getting_started/
3. Build the project
4. Run the project

## System Information
- Operating System: Windows 10
- QGC Version: 3.4.4
- QGC build: self-built from source

## Detailed Description
Here is a screenshot
![untitled](https://user-images.githubusercontent.com/8853887/49245944-c8d1e900-f424-11e8-95c0-2860a7db79a4.png)

## Log Files and Screenshots
Here is a dump file: 
https://drive.google.com/file/d/1YFJ6XHGyes1ZtMxnDMhIa3xmLB-znMva/view?usp=sharing

## Extra
I have tried to build and run QGC on 3 different machines with Windows 10. All of them have the same issue

DonLakeFlyer:
I think this and #7032 are the same bug.

dogmaphobic:
I've seen this several times on Windows only. It happens immediately on boot when creating a `Fact`. I haven't been able to figure out where it's coming from and it doesn't happen every time.

kossumov:
Tried on 3 different machines. Many times. 
Happens every time

kossumov:
The last commit by @DonLakeFlyer fixed the problem.
Thanks ;)

DonLakeFlyer:
@kossumov Thanks for the detailed report.

Issue #7032, opened 2018-11-22
Android: Daily build crashes at startup


fingadar:
The actual daily build on Android tablet crashes immediately after startup.
Tested with Android 6 and 7, same result.

The 3.4 release version works problem-free on same tablets.

Can I give any more useful information? I'm not an Android expert ;-)

DonLakeFlyer:
Hmm, something started going haywire around Nov 8th when crashes started to spike up. WIth 125 reports of java.lang.UnsatisfiedLinkError crash which means a native call is missing somewhere.

@dogmaphobic Anything ring a bell changing recently with android?

DonLakeFlyer:
And of course it doesn't repro for me.

DonLakeFlyer:
@fingadar Is this with a vehicle ready to connect to? Or just no vehicle there and boot.

fingadar:
That doesn't make any difference. I tested it with or without vehicle (UDP connected), with the same result. I think, qGroundControl crashes long before it tries to connect.

DonLakeFlyer:
@findagar Could you check if QGC has the USB port captured in application settings. If so, remove that settings. Also turn off all autoconnect in settings and try again.

DonLakeFlyer:
@fingadar Also what exact devices are you testing on and what OS version is on each one?

fingadar:
So, I did some short tests:

Medion X10302 with Android 6: Release version (3.4.4?) works, daily build crashes
Medion X10302 with Android 7: Release version works, daily build crashes
Stone aged Nexus7 (1st gen) with Android 4.4.4: both the release version and the daily build are working.

The X10302 is my CGS tablet that worked flawlessly with qGroundGontrol for 1 to 2 years now.

Crash means, the app is closed immediately after starting. It does not load the user interface, but at once the "app closed windows" ist shown. So I cannot do or change anything. 

In the installed release version, no USB port is configured. Deinstalling the Release version an reinstalling the Daily Build makes no difference.

What I noticed, the Daily Build does not ask for any permissions during installation. But even giving all permissions by hand makes no differences, it crashes anyway.

DonLakeFlyer:
@fingadar A new Android daily should have posted a few minutes ago. Can you give it a try?

DonLakeFlyer:
I believe this is fixed now. Hopefully @fingadar can verify. Otherwise Google Play Dashboard should tell us from crash reports.

fingadar:
[...]

PR #7047, opened 2018-11-29
Initialization of Fact prior to Qt init causes crash.

DonLakeFlyer:
I believe this is the fix for #7046 and #7032. The timeframe of that code change looks right from a standpoint of when Android crashes started popping up.


dogmaphobic:
That makes sense. In the debugger, I couldn't figure out the stack trace. `this` was valid but the `Fact` was "empty".

PR #7048, opened 2018-11-30
Revert "ArduPilot: Don't starts streams from QGC side. Assume firmware side does the right thing"

DonLakeFlyer:
Reverts mavlink/qgroundcontrol #7017. Needs a different change.


DonLakeFlyer:
Reverting since this breaks more things than it fixes.

PR #7052, opened 2018-11-30
Set videoSource to UDP if connected to an ArduSub vehicle and video is disabled.

Williangalvani:
Since video is required to control a sub, this sets the video to UDP when a Sub vehicle connects if the video is disabled.


Williangalvani:
@DonLakeFlyer Is there a way to test if the options is "unset"? 
It would be better if we could set to UDP only if this has not been touched by the user.

PR #7060, opened 2018-12-04
Android: Switch to NDK 16b

DonLakeFlyer:


DonLakeFlyer:
Ok, so the summary of what I did was:
* Changed to SDK and Build tools v25
* Changed to NDK 16B
* Removed video support for work around application crash (#7032) and/or build problems. Once we figure out how to fix that, video will go back in. But I figure it's better to keep drop android daiy builds with missing features, than no daily builds at all.

PR #7064, opened 2018-12-05
Orbit supports display/changing rotation direction

DonLakeFlyer:
screen shot 2018-12-05 at 11 29 05 am

* QGCMapCircle:
  * Show an optional rotation direction arrows
  * Clicking rotation arrow will reverse direction
*  Orbit uses new QGCMapCircle to support setting rotation direction

Related to https://github.com/PX4/Firmware/pull/10989#issuecomment-444546964

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -16,6 +16,7 @@ Note: This file only contains high level features or important fixes.
 * Survey Planning: add mode that supports concave polygons
 * Support loading polygons from SHP files
 * Bumped settings version (now 8). This will cause all settings to be reset to defaults.
+* Orbit visuals support changing rotation direction
 
 ## 3.4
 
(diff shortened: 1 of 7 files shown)


hamishwillee:
@DonLakeFlyer Is this just PX4, or will it also display in ArduPilot circle mode?
Can I test this on current master SITL?

DonLakeFlyer:
This is PX4 only since it uses the no DO_ORBIT command (which is mavlink generic). If you link through the firmware pull you can see the branch you can test with.

MaEtUgR:
@DonLakeFlyer Latest QGC daily just after this pr only ever sends counter clockwise commands (negative radius) no matter how I switch around with the new arrows.

DonLakeFlyer:
I hadn't merged the other Orbit pull yet. Should be good on new daily with this in it: https://github.com/mavlink/qgroundcontrol/pull/7066

PR #7066, opened 2018-12-06
Show Orbit telemetry on map

DonLakeFlyer:
screen shot 2018-12-05 at 6 39 29 pm

Example above shows a vehicle which is partially through it's first orbit. Still would like an indicator in the center to say this is an orbit but that will come later. This is at least enough to test telemetry working.

* Orbit circle is drawn on map from orbit telemetry values
* Orbit circle will go away after 3 seconds of no orbit telemetry

Related to https://github.com/PX4/Firmware/pull/10989

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewMap.qml
+++ b/src/FlightDisplay/FlightDisplayViewMap.qml
@@ -354,6 +354,13 @@ FlightMap {
         }
     }
 
+    // Used to show orbit status telemetry from the vehicle
+    QGCMapCircleVisuals {
+        mapControl:     parent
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
@hamishwillee More Orbit stuff

MaEtUgR:
Still only negative radius in commands (like with https://github.com/mavlink/qgroundcontrol/pull/7064#issuecomment-444820355).
I realized I have to also fill the correct radius sign in the telemetry.

DonLakeFlyer:
Rotation finally fixed!

PR #7081, opened 2018-12-10
Some Android joysticks have axis mapped to multiple channels.

dogmaphobic:
When connecting these (such as the Ipega BT gamepads) and trying to calibrate, a single stick would end up moving two channels simultaneously.


DonLakeFlyer:
Would this only show up on android. Wondering about SDL based joystick code.

PR #7082, opened 2018-12-10
Add QtGamepad library support for platforms that currently do not hav…

MatejFranceskin:
Currently gamepads are not supported on all platforms.
This PR adds generic QT gamepad support.

It was tested on Linux and Android. On Linux it worked correctly with previously undetected gamepad (Ipega 9083). On Android some mappings were overlapping so the previous solution was kept.

### DIFF
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -21,6 +21,7 @@ linux {
         message("Linux build")
         CONFIG += LinuxBuild
         DEFINES += __STDC_LIMIT_MACROS
+        DEFINES += QGC_ENABLE_GAMEPAD
         linux-clang {
             message("Linux clang")
             QMAKE_CXXFLAGS += -Qunused-arguments -fcolor-diagnostics
@@ -29,6 +30,7 @@ linux {
         message("Linux R-Pi2 build")
         CONFIG += LinuxBuild
         DEFINES += __STDC_LIMIT_MACROS __rasp_pi2__
+        DEFINES += QGC_ENABLE_GAMEPAD
     } else : android-g++ {
         CONFIG += AndroidBuild MobileBuild
         DEFINES += __android__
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
Wasn't there some problem with GamePad QT support not working cross platform or something like that.

DonLakeFlyer:
@nanthony21 Didn't you try this already?

nanthony21:
Yes, please see PR #6061.

The decision at the time was that even though it worked, it wasn't worth going through the steps to properly add QGamePad to the project unless it was going to be a complete replacement for all platforms.

I think at that time QGamePad had issues on Windows. If those issues are fixed now then I think we should go ahead and add QGamePad for all platforms. This should greatly improve support for android and linux as well as allow us to streamline a lot of the current code which is platform specific into platform-agnostic code.

MatejFranceskin:
At first I wanted to add QGamePad for all platforms, but it had issues on Android. I suggest that it is added only for platforms that currently don't have any joystick support at all.

DonLakeFlyer:
> I suggest that it is added only for platforms that currently don't have any joystick support at all.

All platforms, except for iOS have Joystick support.  Bifurcating even further with the joystick backend code will just make things more complicated than they already are. Hence, waiting for QGamePad to reach reality on all platforms and then switch everything to that.

DonLakeFlyer:
@MatejFranceskin FYI: The problem with having different implementations which are per-OS is that it greatly increases the support burden from a standpoint of understanding the code and fixing bugs in it. I tend to push back on this because when bugs are reported against these different sets of code it tends to end up being me that has to spend the time to understand them all. Hence unless there is big ROI to bifurcate, it's my opinion that it's best to keep things simple.

MatejFranceskin:
I completely agree - the code stays on the branch in case anybody needs it.

jaxxzer:
> Yeah, Qt's gamepad, as is, doesn't quite make sense. It's only handles, as the name implies, "gamepads". Only a limited number of buttons and analog sticks. It would be useless for anything beyond a gamepad.

Is this any different from the SDL joystick classes we are using now?

dogmaphobic:
>Is this any different from the SDL joystick classes we are using now?

When I plug in a Seitek X52 I get well over 40 buttons and dozens of analog channels. Far more  than what Gamepad would support.

jaxxzer:
> When I plug in a Seitek X52 I get well over 40 buttons and dozens of analog channels. Far more than what Gamepad would support.

[That's too bad](https://doc.qt.io/qt-5/qgamepad.html).

PR #7084, opened 2018-12-10
Orbit: Show center point telemetry and flight mode

DonLakeFlyer:
I've changed the px4 flight mode header according to what I think will be done here: https://github.com/PX4/Firmware/pull/10989#issuecomment-445849668. If that ends up being wrong I will adjust before I merge this.

screen shot 2018-12-10 at 11 22 40 am


MaEtUgR:
I changed the header according to your suggestion: https://github.com/PX4/Firmware/compare/fdd5bf8a8115ba8ef89da26d3e6fb04e2d6cb6ed..f2c803687d71bf04cff95d04462ad4b663333269

PR #7087, opened 2018-12-12
[WIP] Add support to receive NMEA streams via udp

Williangalvani:
This enables QGC to receive an NMEA stream via UDP instead of serial.
It also shows the direction of the groundstation if the information is present in the NMEA stream.

Without heading information:
screenshot from 2018-12-10 16-37-25

With heading information:
screenshot from 2018-12-10 17-28-34

It can be tested with [this script](https://gist.github.com/Williangalvani/ff07e3f877f7ce9569557e3888f43cae).

run `./udpgpstestpy --noheading --ip 127.0.0.1 --port 14401` to test without heading information
and `./udpgpstestpy --ip 127.0.0.1 --port 14401` to test with heading information

### DIFF
--- a/src/PositionManager/PositionManager.cpp
+++ b/src/PositionManager/PositionManager.cpp
@@ -14,6 +14,7 @@
 QGCPositionManager::QGCPositionManager(QGCApplication* app, QGCToolbox* toolbox)
     : QGCTool           (app, toolbox)
     , _updateInterval   (0)
+    , _gcsHeading       (NAN)
     , _currentSource    (NULL)
     , _defaultSource    (NULL)
     , _nmeaSource       (NULL)
@@ -61,6 +62,7 @@ void QGCPositionManager::setNmeaSourceDevice(QIODevice* device)
 void QGCPositionManager::_positionUpdated(const QGeoPositionInfo &update)
 {
     QGeoCoordinate newGCSPosition = QGeoCoordinate();
+    qreal newGCSHeading = update.attribute(QGeoPositionInfo::Direction);
 
     if (update.isValid()) {
         // Note that gcsPosition filters out possible crap values
@@ -71,6 +73,10 @@ void QGCPositionManager::_positionUpdated(const QGeoPositionInfo &update)
     if (newGCSPosition != _gcsPosition) {
         _gcsPosition = newGCSPosition;
         emit gcsPositionChanged(_gcsPosition);
+    }
+    if (newGCSHeading != _gcsHeading) {
+        _gcsHeading = newGCSHeading;
+        emit gcsHeadingChanged(_gcsHeading);
     }
 
     emit positionInfoUpdated(update);
--- a/src/PositionManager/PositionManager.cpp
+++ b/src/PositionManager/PositionManager.cpp
@@ -51,8 +51,19 @@ void QGCPositionManager::setToolbox(QGCToolbox *toolbox)
 
 void QGCPositionManager::setNmeaSourceDevice(QIODevice* device)
 {
+    // stop and release _nmeaSource
     if (_nmeaSource) {
+        _nmeaSource->stopUpdates();
+        disconnect(_nmeaSource);
+
+        // if _currentSource is pointing there, point to null
+        if (_currentSource == _nmeaSource){
+            _currentSource = nullptr;
+        }
+
         delete _nmeaSource;
+        _nmeaSource = nullptr;
+
     }
     _nmeaSource = new QNmeaPositionInfoSource(QNmeaPositionInfoSource::RealTimeMode, this);
     _nmeaSource->setDevice(device);
--- a/src/PositionManager/PositionManager.h
+++ b/src/PositionManager/PositionManager.h
@@ -25,7 +25,8 @@ public:
     QGCPositionManager(QGCApplication* app, QGCToolbox* toolbox);
     ~QGCPositionManager();
 
-    Q_PROPERTY(QGeoCoordinate gcsPosition READ gcsPosition NOTIFY gcsPositionChanged)
+    Q_PROPERTY(QGeoCoordinate gcsPosition  READ gcsPosition  NOTIFY gcsPositionChanged)
+    Q_PROPERTY(qreal          gcsHeading   READ gcsHeading   NOTIFY gcsHeadingChanged)
 
     enum QGCPositionSource {
         Simulated,
@@ -35,6 +36,8 @@ public:
     };
 
     QGeoCoordinate gcsPosition(void) { return _gcsPosition; }
+
+    qreal gcsHeading() { return _gcsHeading; }
 
     void setPositionSource(QGCPositionSource source);
 
@@ -50,11 +53,13 @@ private slots:
 
 signals:
     void gcsPositionChanged(QGeoCoordinate gcsPosition);
+    void gcsHeadingChanged(qreal gcsHeading);
     void positionInfoUpdated(QGeoPositionInfo update);
 
 private:
     int             _updateInterval;
     QGeoCoordinate  _gcsPosition;
+    qreal           _gcsHeading;
 
     QGeoPositionInfoSource*     _currentSource;
     QGeoPositionInfoSource*     _defaultSource;
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -479,6 +479,28 @@ void LinkManager::_updateAutoConnectLinks(void)
         createConnectedLink(config);
         emit linkConfigurationsChanged();
     }
+#ifndef __mobile__
+    // check to see if nmea gps is configured for UDP input, if so, set it up to connect
+    if (_autoConnectSettings->autoConnectNmeaPort()->cookedValueString() == "UDP Port")
+    {
+        if (_nmeaSocket.localPort() != _autoConnectSettings->nmeaUdpPort()->cookedValue().toUInt()
+                || _nmeaSocket.state() != UdpIODevice::BoundState) {
+            qCDebug(LinkManagerLog) << "Changing port for UDP NMEA stream";
+            _nmeaSocket.close();
+            _nmeaSocket.bind(QHostAddress::AnyIPv4, _autoConnectSettings->nmeaUdpPort()->cookedValue().toUInt());
+            _toolbox->qgcPositionManager()->setNmeaSourceDevice(&_nmeaSocket);
+        }
+        //close serial port
+        if (_nmeaPort) {
+            _nmeaPort->close();
+            delete _nmeaPort;
+            _nmeaPort = nullptr;
+            _nmeaDeviceName = "";
+        }
+    } else {
+        _nmeaSocket.close();
+    }
+#endif
 
 #ifndef NO_SERIAL_LINK
     QStringList currentPorts;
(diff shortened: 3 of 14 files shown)


DonLakeFlyer:
I think the fact that GCS location with heading looks like like a vehicle is a bad idea. Keep the Q and put the indicator behind it or something like that. Or a hollow indicator plus the Q. Something to make it obvious this is the ground station not a vehicle.

Williangalvani:
You are right, I'll do that.

Williangalvani:
Updated:
screenshot from 2018-12-14 13-59-29

jaxxzer:
@DonLakeFlyer, @patrickelectric and I are giving this a review, we'll take the WIP label off when we are ready.

DonLakeFlyer:
Shouldn't there be some configuration for this? Turn off on, port?

jaxxzer:
Shouldn't there be some configuration for this? Turn off on, port?

We could add it. Right now it's analogous to the 14550 hardcoded for mavlink.

DonLakeFlyer:
> We could add it. Right now it's analogous to the 14550 hardcoded for mavlink.

Right but that kind of a mavlink standard. There isn't any standard for this is there?

src/comm/UdpIODevice.cc
```diff
@@ -0,0 +1,42 @@
...
+ * COPYING.md in the root of the source code directory.
+ *
+ ****************************************************************************/
+
+#include "UdpIODevice.h"
+#include <algorithm>
+
+UdpIODevice::UdpIODevice(QObject *parent) : QUdpSocket(parent)
+{
+    _buffer = QByteArray();
```
patrickelectric: This is not necessary.

src/comm/UdpIODevice.cc
```diff
@@ -0,0 +1,42 @@
...
+}
+
+bool UdpIODevice::canReadLine() const
+{
+    return _buffer.indexOf('\n') > -1 || hasPendingDatagrams();
+}
+
+qint64 UdpIODevice::readLineData(char *data, qint64 maxSize)
```
patrickelectric: Take a look in `QTextStream::readLine`

jaxxzer:
> Right but that kind of a mavlink standard. There isn't any standard for this is there?

I think it the mavlink port became standard by way of hard coding it into the gcs, so we could create the standard for the nmea input in the same way by saying "this is the port to use if you want to do this".

It's also analogous to the video port default being 5600 (which we do allow to change).

@DonLakeFlyer Do you prefer there to be an extra option + ui element or a hardcoded port? We can add an extra ui, but it's one more thing for the user to set up in a case where there is really no need to have more than one option (I might be wrong).

DonLakeFlyer:
You can default to the standard port you want but I think there needs to be the option to change and turn off.

src/comm/UdpIODevice.cc
```diff
@@ -0,0 +1,42 @@
...
+#include "UdpIODevice.h"
+#include <algorithm>
+
+UdpIODevice::UdpIODevice(QObject *parent) : QUdpSocket(parent)
+{
+}
+
+bool UdpIODevice::canReadLine() const
+{
+    return _buffer.indexOf('\n') > -1 || hasPendingDatagrams();
```
jaxxzer: The pending datagrams do not necessarily contain a newline.

You should read any pending datagrams in, then

`return _buffer.indexOf('\n') > -1`

src/comm/LinkManager.cc
```diff
@@ -478,8 +479,28 @@ void LinkManager::_updateAutoConnectLinks(void)
...
+            qCDebug(LinkManagerLog) << "Creating UDP socket for NMEA stream";
+            _nmeaSocket = new UdpIODevice(this);
+        }
+        if (_nmeaSocket->localPort() != _autoConnectSettings->nmeaUdpPort()->cookedValue().toUInt()) {
+            qCDebug(LinkManagerLog) << "Changing port for UDP NMEA stream";
+            _nmeaSocket->close();
+            _nmeaSocket->bind(QHostAddress::AnyIPv4, _autoConnectSettings->nmeaUdpPort()->cookedValue().toUInt());
+            _toolbox->qgcPositionManager()->setNmeaSourceDevice(_nmeaSocket);
+        }
+    }
```
jaxxzer: Match conditional brace style to the rest of the surrounding code.

src/ui/preferences/GeneralSettings.qml
```diff
@@ -498,6 +498,7 @@ QGCView {
 
                                     model:  ListModel {
                                         ListElement { text: "disabled" }
+                                        ListElement { text: "UDP Port"}
```
jaxxzer: There is no option for serial here... but also before the patch?

jaxxzer: Of course it lists the serial devices connected (if any).

Williangalvani: Yeah, it is populated dinamically

src/comm/LinkManager.cc
```diff
@@ -478,8 +479,28 @@ void LinkManager::_updateAutoConnectLinks(void)
...
+
         emit linkConfigurationsChanged();
     }
+#ifndef __android__
+    if (_autoConnectSettings->autoConnectNmeaPort()->cookedValueString() == "UDP Port")
+    {
+        if (_nmeaSocket == nullptr)
+        {
+            qCDebug(LinkManagerLog) << "Creating UDP socket for NMEA stream";
+            _nmeaSocket = new UdpIODevice(this);
```
patrickelectric: it's much easier to use a unique smart pointer than checking for nullptr and using new all the place.

src/comm/UdpIODevice.cc
Williangalvani: I took a look.
While it can work on a ByteArray, It wasn't happy when it got resized.
without resizing the ByteArray, it would just store all the data in memory, instead of freeing what got consumed.

src/comm/UdpIODevice.cc
jaxxzer: Maybe use this class should use the parent readyRead signal to read in the datagrams as soon as they are ready. http://doc.qt.io/qt-5/qiodevice.html#readyRead

src/PositionManager/PositionManager.h
```diff
@@ -36,6 +37,8 @@ class QGCPositionManager : public QGCTool {
 
     QGeoCoordinate gcsPosition(void) { return _gcsPosition; }
 
+    qreal gcsHeading(void) { return _gcsHeading;  }
```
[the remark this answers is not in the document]
patrickelectric: http://www.stroustrup.com/examples_short.pdf
Search for "abomination".

src/PositionManager/PositionManager.cpp
```diff
@@ -70,6 +71,8 @@ void QGCPositionManager::_positionUpdated(const QGeoPositionInfo &update)
     }
     if (newGCSPosition != _gcsPosition) {
         _gcsPosition = newGCSPosition;
+        _gcsHeading = update.attribute(QGeoPositionInfo::Direction);
```
jaxxzer: What if the heading changed from the previous update, but the position did not? This code will not run to update the heading.

src/ui/preferences/GeneralSettings.qml
Williangalvani: done!

src/PositionManager/PositionManager.cpp
Williangalvani: Gps signals don't usually stay really fixed, but who knows what data the users might feed in here?
Fixed it.

src/ui/preferences/GeneralSettings.qml
```diff
@@ -532,6 +535,17 @@ QGCView {
                                         nmeaBaudCombo.currentIndex = index;
                                     }
                                 }
+
+                                QGCLabel {
+                                    text:       qsTr("NMEA stream UDP port")
+                                    visible:    nmeaPortCombo.currentText == "UDP Port"
+                                    }
```
jaxxzer: whitespace

src/PositionManager/PositionManager.cpp
```diff
@@ -52,7 +52,12 @@ void QGCPositionManager::setToolbox(QGCToolbox *toolbox)
 void QGCPositionManager::setNmeaSourceDevice(QIODevice* device)
 {
     if (_nmeaSource) {
+        if (_nmeaSource == _currentSource){
+            _currentSource = nullptr;
```
jaxxzer: make sure you perform the appropriate cleanup [here](https://github.com/mavlink/qgroundcontrol/pull/7087/commits/45321bfb16406cbc0021604f496eecf41d80533b#diff-dc7e6678e48a96d0ba57eab3b2676189R96)

src/comm/LinkManager.cc
jaxxzer: *cough* `else {`

src/ui/preferences/GeneralSettings.qml
```diff
@@ -518,13 +519,18 @@ QGCView {
...
                                         var index = nmeaPortCombo.find(QGroundControl.settingsManager.autoConnectSettings.autoConnectNmeaPort.valueString);
                                         nmeaPortCombo.currentIndex = index;
+                                        if (QGroundControl.linkManager.serialPorts.length === 0) {
+                                            nmeaPortCombo.model.append({text: "Serial <none available>"})
+                                        }
                                     }
                                 }
 
                                 QGCLabel {
+                                    visible:          nmeaPortCombo.currentText !== "UDP Port" && nmeaPortCombo.currentText !== "disabled"
```
jaxxzer: suggest a property to reduce the opportunity for typo induced bugs.

src/comm/UdpIODevice.cc
```diff
@@ -0,0 +1,44 @@
...
+bool UdpIODevice::canReadLine() const
+{
+    return _buffer.indexOf('\n') > -1;
+}
+
+qint64 UdpIODevice::readLineData(char *data, qint64 maxSize)
+{
+    int length = _buffer.indexOf('\n') + 1; // add 1 to include the '\n'
+    if (length == -1) {
+        return 0;
```
jaxxzer: This condition will never be true. http://doc.qt.io/qt-5/qbytearray.html#indexOf

src/comm/UdpIODevice.h
```diff
@@ -0,0 +1,34 @@
...
+ *
+ **/
+
+class UdpIODevice: public QUdpSocket
+{
+    Q_OBJECT
+public:
+    bool     canReadLine() const;
+    qint64   readLineData(char *data, qint64 maxSize);
+    explicit UdpIODevice(QObject *parent = nullptr);
```
jaxxzer: let's declare constructor first.

jaxxzer: maybe remove `explicit`.

src/PositionManager/PositionManager.cpp
```diff
@@ -61,15 +62,18 @@ void QGCPositionManager::setNmeaSourceDevice(QIODevice* device)
...
+    qreal newGCSHeading = update.attribute(QGeoPositionInfo::Direction);
 
     if (update.isValid()) {
         // Note that gcsPosition filters out possible crap values
         if (qAbs(update.coordinate().latitude()) > 0.001 && qAbs(update.coordinate().longitude()) > 0.001) {
             newGCSPosition = update.coordinate();
         }
     }
-    if (newGCSPosition != _gcsPosition) {
+    if (newGCSPosition != _gcsPosition || newGCSHeading != _gcsHeading) {
```
jaxxzer: This should be split into two independent condtitions

src/ui/preferences/GeneralSettings.qml
```diff
@@ -502,9 +502,13 @@ QGCView {
                                 QGCComboBox {
                                     id:                     nmeaPortCombo
                                     Layout.preferredWidth:  _comboFieldWidth
+                                    property string audioDisabled: "Disabled"
```
jaxxzer: I think you are mixed up.
`audio`

PR #7094, opened 2018-12-17
PARAM_VALUE parsing changes, Rover hack fix

DonLakeFlyer:
* ArduPilot: Assume non-autopilot components follow current PARAM_VALUE spec. Related to #7049
* PX4: Hack to allow Rover setup pages to display. Related to https://github.com/PX4/Firmware/issues/10969

PR #7099, opened 2018-12-23
Rover isFlying check fix

DonLakeFlyer:
* Fix the fixf or #7094
* Add --clear-cache command line options. This will erase the parameter value cache for all vehicles as well as the parameter and airframe meta data cached files.


DonLakeFlyer:
@hamishwillee Can you update the docs for the new --clear-cache option?

PR #7114, opened 2019-01-02
Add support to receive NMEA streams via UDP

Williangalvani:
(I'm opening this one with a clean slate after the review in #7087  )

This enables QGC to receive an NMEA stream via UDP instead of serial.
It also shows the direction of the groundstation if the information is present in the NMEA stream.

Without heading information:
screenshot from 2018-12-10 16-37-25

With heading information:
screenshot from 2018-12-10 17-28-34

It can be tested with [this script](https://gist.github.com/Williangalvani/ff07e3f877f7ce9569557e3888f43cae).

run `./udpgpstestpy --noheading --ip 127.0.0.1 --port 14401` to test without heading information
and `./udpgpstestpy --ip 127.0.0.1 --port 14401` to test with heading information


jaxxzer:
ping @DonLakeFlyer let us know if you have any questions!

src/comm/LinkManager.cc
```diff
@@ -479,6 +479,28 @@ void LinkManager::_updateAutoConnectLinks(void)
         createConnectedLink(config);
         emit linkConfigurationsChanged();
     }
+#ifndef __mobile__
+    // check to see if nmea gps is configured for UDP input, if so, set it up to connect
+    if (_autoConnectSettings->autoConnectNmeaPort()->cookedValueString() == "UDP Port")
+    {
```
DonLakeFlyer: Minor nit, but coding style should have this on the previous line.

src/comm/LinkManager.cc
```diff
@@ -479,6 +479,28 @@ void LinkManager::_updateAutoConnectLinks(void)
...
+        if (_nmeaSocket.localPort() != _autoConnectSettings->nmeaUdpPort()->cookedValue().toUInt()
```
DonLakeFlyer: Internal C++ code would normally use rawValue and not cookedValue. cookedValues are normally for display to the user. Is there are reason why cookedValue is used here?

src/comm/LinkManager.cc
Williangalvani: sorry, fixed.

src/comm/LinkManager.cc
Williangalvani: my mistake, just fixed it.

Issue #7115, opened 2019-01-03
Joystick buttons on Arducopter do not work (reserved by firmware)


highfreq:
Hello, i am trying to configure my joystick with QGroundcontrol. The joystick axis seems to work and calibrate fine.

The problem is with buttons, where there should be the button configuration menu it just says “button reserved by firmware” and i can’t do anything with them. They seem to work because when i push them the lights light up.
Please provide an unambiguous set of steps to reproduce the current behavior

1. Try to use Joystick with Arducopter
2.
3.
4.

I tested latest QGC version on pretty much all platforms.

PR #7120, opened 2019-01-04
Remove the concept of reserved buttons for manual control

DonLakeFlyer:
Neither PX4 nor ArduPilot need this. Fix for #7115

@jaxxzer Could you or one of your folks review and test this? I don't want to screw you guys up and you know most about joystick stuff.

Also the joystick code send button bits to the MANUAL_CONTROL message. Seems like it would make sense to never send those button bits since QGC takes all the buttons. Opinions?

### DIFF
--- a/src/FirmwarePlugin/APM/ArduSubFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/ArduSubFirmwarePlugin.cc
@@ -164,11 +164,6 @@ int ArduSubFirmwarePlugin::remapParamNameHigestMinorVersionNumber(int majorVersi
     return majorVersionNumber == 3 ? 6 : Vehicle::versionNotSetValue;
 }
 
-int ArduSubFirmwarePlugin::manualControlReservedButtonCount(void)
-{
-    return 0;
-}
-
 void ArduSubFirmwarePlugin::initializeStreamRates(Vehicle* vehicle) {
     vehicle->requestDataStream(MAV_DATA_STREAM_RAW_SENSORS,     2);
     vehicle->requestDataStream(MAV_DATA_STREAM_EXTENDED_STATUS, 2);
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -228,13 +228,6 @@ bool APMFirmwarePlugin::setFlightMode(const QString& flightMode, uint8_t* base_m
     return found;
 }
 
-int APMFirmwarePlugin::manualControlReservedButtonCount(void)
-{
-    // We don't know whether the firmware is going to used any of these buttons.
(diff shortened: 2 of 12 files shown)


DonLakeFlyer:
> Also the joystick code send button bits to the MANUAL_CONTROL message. Seems like it would make sense to never send those button bits since QGC takes all the buttons. Opinions?

@jaxxzer I pulled out the code that sends button pressed bits to MANUAL_CONTROL since that seemed to make the most sense.

jaxxzer:
I'll test this.

DonLakeFlyer:
I tested with my Logitech+PX4 Pro and it seem ok but could use SITL testing with ArduPilot.

jaxxzer:
ArduSub does use these buttons.

DonLakeFlyer:
> ArduSub does use these buttons.

From MANUAL_CONTROL.buttons on the firmware side?

jaxxzer:
> From MANUAL_CONTROL.buttons on the firmware side?

Indeed. It is something I intend to get away from, but it will take some time.

DonLakeFlyer:
Ah, ok. I'll put that back then.

DonLakeFlyer:
But if ArduSub uses these buttons bits then it would seem that ArduSub needs to reserve from use in Joystick config? Which would mean none of this change is valid.

DonLakeFlyer:
And that the correct change is for standard APMFirmwarePlugin to return 0 for none reserved and ArduSub to return how ever many buttons it wants to reserve.

jaxxzer:
I haven't taken a good look at the patch yet. You are probably right.

src/FirmwarePlugin/APM/ArduSubFirmwarePlugin.cc
```diff
@@ -164,11 +164,6 @@ int ArduSubFirmwarePlugin::remapParamNameHigestMinorVersionNumber(int majorVersi
     return majorVersionNumber == 3 ? 6 : Vehicle::versionNotSetValue;
 }
 
-int ArduSubFirmwarePlugin::manualControlReservedButtonCount(void)
-{
-    return 0;
```
DonLakeFlyer: This is what seems odd in previous code. I would have expected that if ArduSub wanted those button bits that it would have reserved them for use.

DonLakeFlyer: I see it now. Not really because it used a jsScriptButton flag thing to turn off all of the normal button setting ui.

Which means I think:
* The reserved button thing can go away since ArduSub already has the other jsScript flag thing to signal usage.
* The button bits should be passed to MANUAL_CONTROL

Which I think means that this pull is correct to the above spec.

jaxxzer:
Looking now

jaxxzer:
lgtm, thanks!

PR #7121, opened 2019-01-05
Include prebuilt android OpenSLL libs

DonLakeFlyer:
Never versions of Android required apps to include their owns versions of OpenSLL libs.


DonLakeFlyer:
@dogmaphobic Can you look through this as well as trying on your Android stuff? Easiest test is to add some waypoints over the US. If Qt spits out SSL errors then it's not working.

PR #7124, opened 2019-01-05
Mobile visual cleanup

DonLakeFlyer:
Big pile of changes for better visuals/layout on Mobile. 

Largest obvious visual change is a major simplification of the radio button and checkbox code. On mobile the checkboxes are larger for better touch targets. This also increased the sise of the checkmark image inside to indicate checked. Which in turn looked like crap when blown up in size. Instead of fighting with it I simplified the check box visuals in a way which works well at any size:
screen shot 2019-01-05 at 10 39 58 am

The set of changes make mobile look much better and also lets QGC run much better on smaller screens.


DonLakeFlyer:
Heres the Plan view running on a cheap 5" android phone:
screenshot_2019-01-05-13-57-49

Quite usable.

Issue #5383, opened 2017-07-02
Android: Possible conflict with QGC and Tower connecting to Sik radio


rsilk1949:
I am a new user who is looking for a good GCS for Android/Sik radio setup.
First off this configuration works with Tower.

The USB/serial device is not automatically recognised so I tried adding a link. I did not see any choice for USB or OTG, so I asumed Serial. Baud rate was set to 57600. When I hit the Connect button, I get an error "Error on link. Error connecting. Could not create port. No such file or directory"

Also, I could not find a connect button except in the Settings/Comm Links page. I did see a screenshot of the Mac app which did have a connect button on the main page. This should be available on Android UI too IMO.

DonLakeFlyer:
Are you using a daily build? If not, then try that. If that still doesn't work we can discuss further from there.

https://docs.qgroundcontrol.com/en/releases/daily_builds.html

rsilk1949:
Like many potential users, I have no way to install Android apps except via Google  Play. So I will have to wait for the  next release version.

DonLakeFlyer:
The daily build is available via Google Play. Follow the link I posted above.

rsilk1949:
Installed beta version and still get the same result.

kikislater:
Hi, 
I have the same behaviour ! Video to understand below : 
https://youtu.be/RktFWS_NyjE

DonLakeFlyer:
When the popup came on the screen when you plugged in the radio you clicked Cancel. Which in turn did not give acess for QGC to talk with the radio. When that pop up comes up you need to click the QGC icon in the popup.

kikislater:
No, it's wrong. Even with cancel, Tower have access to usb serial whereas qgc not !
I could remake a video. It will not give the other popup to grant authorisations. Difference between Qgroundcontrol and Tower is that Qgroundcontrol is not allowed to modify system parameters.

![qgc_com miui securitycenter](https://user-images.githubusercontent.com/13238064/39414863-72a7dd34-4c4d-11e8-8707-690cef57c2b6.jpg)

![tower_com miui securitycenter](https://user-images.githubusercontent.com/13238064/39414865-772d9f6a-4c4d-11e8-914e-081534f6631d.jpg)

First paramter in Tower shows is 'modify system parameters'. I have other app like drone deploy, dji go4, drone harmony and I have this parameter in app and popup to grant authorisations

DonLakeFlyer:
[...]

PR #7131, opened 2019-01-07
PR Taisync support for Android

MatejFranceskin:
Ported TaiSync Support to Android platform.


android/res/xml/device_filter.xml
```diff
@@ -2,5 +2,6 @@
 <resources>
     <!-- Allow anything connected -->
     <usb-device />
+    <usb-accessory />
```
dogmaphobic: Is there a way to limit this to the Taisync? The Android documentation says you should though I can't quite figure out what information you would provide that would _filter_ the Taisync only:

_In this case, the following resource file should be saved in res/xml/accessory_filter.xml and specifies that any accessory that has the corresponding model, manufacturer, and version should be filtered. The accessory sends these attributes the Android-powered device:_
```
<?xml version="1.0" encoding="utf-8"?>

<resources>
    <usb-accessory model="DemoKit" manufacturer="Google" version="1.0"/>
</resources>
```

MatejFranceskin: Yes indeed - I limited it to:
<usb-accessory model="android.usbaoa" manufacturer="taisync" version="1.0"/>

DonLakeFlyer:
My concern with this is that it doesn't conflict with usb radios/boards being plugged in. I just went througha nightmare change to hopefully fix #5383. My concerns is that this changes regular usb device behavior as well. Is there any overlap there with respect to ```usb-device``` showing up as ```usb-accesory``` as well?

MatejFranceskin:
Found a bug in the implementation - it recognizes USB accessory only when device is in developer or debug mode. Please wait for the fix and more testing.

MatejFranceskin:
> Found a bug in the implementation - it recognizes USB accessory only when device is in developer or debug mode. Please wait for the fix and more testing.

It turns out that this is how it is implemented and cannot be changed:
Currently android usb multimedia transmission requires the support of aoa protocol, there are currently two versions 0x2D00 and 0x2D01. If android terminal aoa version is 0x2D01 must enable usb debuging mode. http://www.android-doc.com/tools/adk/aoa2.html.
Android devices that we tested use 0x2D01 version and require USB debugging to be switched on.

MatejFranceskin:
> My concern with this is that it doesn't conflict with usb radios/boards being plugged in. I just went througha nightmare change to hopefully fix #5383. My concerns is that this changes regular usb device behavior as well. Is there any overlap there with respect to `usb-device` showing up as `usb-accesory` as well?

I don't think that this implementation interferes with the was USB devices are handled. I tested it with direct USB connection to AP and it works like it always did.

PR #7136, opened 2019-01-08
PR android joystick improvement

MatejFranceskin:
name:  Bugfix: Joystick is not visible if connected after start of QGC

On Android the Joysticks are probed only at the start of QGC. If the Joystick is connected after QGC is started it never appears in the system.

1. Drone switched on
2. QGC is started and connected to the drone
3. Bluetooth joystick or gamepad is connected to the system
4. Joystick tab does not appear in configuration

Expected behavior:
Joystick should appear when it is available to the system regardles even after QGC is up and running

Fix:
Android joysticks are periodically checked and probed in the same was as SDL joysticks on other platforms.


dogmaphobic:
This is the case for all platforms, not just Android. You were able to fix this on Android because it has a very separate, specific code base (with clear support for dynamic plugging in and out). The other platforms use SDL and something similar would have to be done there.

dogmaphobic:
> For Android there was a TODO which is now implemented.

Ha! Perfect. I was only vaguely aware of that code, never had spent any actual time looking into any of it. The notion of joystick not showing up if not already connect at (QGC) boot time was based on history. I missed the fact that had been fixed since.

DonLakeFlyer:
This pull seems to have more stuff in it from the commit list? Doesn't seem correct.

MatejFranceskin:
This PR is derived from #7131 - I will close it for now and wait for 7131 first.

PR #7145, opened 2019-01-15
Support for Septentrio base station RTK

MatejFranceskin:
Septentrio base station is recognized by QGC and configured to send RTCM RTK messages to the flight controller.
Tested on the actual hardware by https://github.com/bkueng

### DIFF
--- a/src/GPS/definitions.h
+++ b/src/GPS/definitions.h
@@ -63,10 +63,9 @@ public:
     static void usleep(unsigned long usecs) { QThread::usleep(usecs); }
 };
 
-static inline void usleep(unsigned long usecs) {
+static inline void gps_usleep(unsigned long usecs) {
     Sleeper::usleep(usecs);
 }
-
 
 typedef uint64_t gps_abstime;
 
--- a/src/GPS/definitions.h
+++ b/src/GPS/definitions.h
@@ -50,6 +50,8 @@
 #include "vehicle_gps_position.h"
 #include "satellite_info.h"
 
+#define M_DEG_TO_RAD 		M_PI / 180.0
+#define M_RAD_TO_DEG 		180.0 / M_PI
 #define M_DEG_TO_RAD_F 		0.01745329251994f
 #define M_RAD_TO_DEG_F 		57.2957795130823f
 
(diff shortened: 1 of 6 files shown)


src/GPS/definitions.h
```diff
@@ -50,6 +50,8 @@
 #include "vehicle_gps_position.h"
 #include "satellite_info.h"
 
+#define M_DEG_TO_RAD 		M_PI / 180.0
```
bkueng: Can you add brackets around these?

MatejFranceskin: Yes - done.

DonLakeFlyer:
> in terms of differences to the other RTKs: the Septentrio RTK does not support survey-in, however you can set the base station position.

If that's the case doesn't there need to be more changes than this? For example there is UI in the RTK dropdown which shows status of survey in.

bkueng:
> If that's the case doesn't there need to be more changes than this? For example there is UI in the RTK dropdown which shows status of survey in.

Not for the dropdown, it will directly show the RTK streaming. In the config maybe, since the survey-in params won't have any effect.

MatejFranceskin:
When I started to work with Septentrio I also wondered why don't we just measure positions for some time and average the measurements to get base station position and Septentrion support told me that I could never accurate position this way. You have to place base station for 2-3h at an unknown point and then do post-processing calculations that take in account lots of stuff including for example solar weather to get its accurate position.
I think that it should be "use specified base station postion" by default. Base stations offer different ways of getting its position including for example connection to Precise Point Positioning system, so we could leave the choice of method to the user and proprietary software that is bundled with the base station. RTK is used to achieve cm accuracy and survey-in mode is only slightly better then ordinary GNSS + SBAS L1+L2 capable receiver without base station. I doesn't make sense to use equipment that is so expensive for such results.
Here are also Trimble instructions on how to execute a RTK mission: https://kb.unavco.org/kb/print-617.html

bkueng:
@DonLakeFlyer 
> So RTK settings default to "Perform Survey-In". What happens if I connect one of these with this set?

It will just use the first position that the base gets.

@MatejFranceskin you are correct and while I agree for high-end systems, currently people use QGC with a u-blox M8P as base, where the survey-in is part of the usual work-flow. So I would not change the default, but we should reconsider that when/if high-end systems become more commonly used.

DonLakeFlyer:
This is good to go then

PR #7146, opened 2019-01-17
Camera and Video Streams

dogmaphobic:
Moved the video stream API into the camera API (as opposed to a stand alone API, which you could not query stream options and no easy way to associate a camera with a camera/vehicle).

QGC already supported the full camera API where each vehicle may expose one or more cameras. Now each camera may expose one or more video streams. The discovery and configuration is automatic. No need to go into General Settings and type URI, aspect ratio, etc. (though that is still supported for cameras/vehicles, which do not implement this API).

Documentation is already under way and can be found here: https://mavlink.io/en/services/camera.html

@DonLakeFlyer The Travis unit test is failing in `FactSystemTestGeneric` on Android (core dump). I don't know what could be causing that.

### DIFF
--- a/src/Camera/QGCCameraControl.cc
+++ b/src/Camera/QGCCameraControl.cc
@@ -10,6 +10,7 @@
 #include "SettingsManager.h"
 #include "VideoManager.h"
 #include "QGCMapEngine.h"
+#include "QGCCameraManager.h"
 
 #include <QDir>
 #include <QStandardPaths>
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
From looking at Travis history something seems to have gone haywire about 3 days ago.

DonLakeFlyer:
Last change to .travis.yml was 19 days ago which doesn't match up.

PR #7148, opened 2019-01-17
PR android joystick improvement v2

MatejFranceskin:
Bugfix: Joystick is not visible if connected after start of QGC

On Android the Joysticks are probed only at the start of QGC. If the Joystick is connected after QGC is started it never appears in the system.

    Drone switched on
    QGC is started and connected to the drone
    Bluetooth joystick or gamepad is connected to the system
    Joystick tab does not appear in configuration

Expected behavior:
Joystick should appear when it is available to the system regardles even after QGC is up and running

Fix:
Android joysticks are periodically checked and probed in the same was as SDL joysticks on other platforms.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,9 +2,18 @@
 
 Note: This file only contains high level features or important fixes.
 
-## 3.5
+## 3.6
 
-### 3.5.0 - Daily Build
+### 3.6.0 - Daily Build
+
+* No changes yet
+
+### 3.5.0 - Stable
+* Plan GeoFence: Fix loading of fence from intermediate 3.4 code
+* Structure Scan: Fix loading of structure scan height
+* ArduPilot: Fix location of planned home position when not connected to vehicle. Issue #6840.
+* Fix loading of parameters from multiple components. Would report download complete too early, thus missing all default component params.
+* Fix file delete in mobile file dialogs
 * Add support for specifying fixed RTK based station location in Settings/General.
 * Added Airmap integration to QGC
 * Added ESTIMATOR_STATUS values to new estimatorStatus Vehicle FactGroup. These are now available to display in instrument panel.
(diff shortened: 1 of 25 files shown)


DonLakeFlyer:
Is there a way to do this based on a joystick attach signal somehow? Kind of like
```
            <meta-data android:resource="@xml/device_filter" android:name="android.hardware.usb.action.USB_DEVICE_ATTACHED"/>
```

The reason is that there is already so much stuff going on in timers in QGC. All of that kills battery since it is constantly running. Would be better if it was on a trigger somehow.

MatejFranceskin:
I agree with that. I couldn't find a way at the beginning but will investigate again.

MatejFranceskin:
I implemented listening for bluetooth.ACL_CONNECTED event and then start the timer that checks periodically for 5 seconds as the gamepad is not immediately available after CONNECTED event. This eliminates constant polling.

dogmaphobic:
@MatejFranceskin Can you rebase to master? I will then merge when completed.

MatejFranceskin:
Rebased and re-tested. Thanks!

DonLakeFlyer:
@MatejFranceskin We are getting tons of crashes reported on Google Play with this stack:

java.lang.UnsatisfiedLinkError: 
  at org.mavlink.qgroundcontrol.QGCActivity.nativeUpdateAvailableJoysticks (Native Method)
  at org.mavlink.qgroundcontrol.QGCActivity.access$800 (QGCActivity.java:65)
  at org.mavlink.qgroundcontrol.QGCActivity$3.onReceive (QGCActivity.java:169)
  at android.app.LoadedApk$ReceiverDispatcher$Args.run (LoadedApk.java:1185)
  at android.os.Handler.handleCallback (Handler.java:751)
  at android.os.Handler.dispatchMessage (Handler.java:95)
  at android.os.Looper.loop (Looper.java:154)
  at android.app.ActivityThread.main (ActivityThread.java:6682)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1520)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1410)

Looks to be a really big problem. Not sure if this code is working correctly?

Issue #4293, opened 2016-12-06
Joystick Settings and MANUAL_CONTROL issues


larryyonge:
The MANUAL_CONTROL MAVLink message sends the joystick information from QGC to PX4 when PX4 is compiled and run as SITL. There are several issues related to the QGC joystick settings and the joystick throttle value that appears in the MANUAL_CONTROL message. This issue is base on QGC build cloned on November 1, 2016.

1)        The QGC joystick settings "Full down stick is zero throttle" and "Center stick is zero throttle" appear to behave incorrectly:
a.        Selecting "Center stick is zero throttle" results in 0-1000 throttle values in the MANUAL_CONTROL message where 0 = full back joystick throttle position, 1000 = full forward throttle position, and "linear" in between (see issue 2) below). I believe it should be 0 for joystick position between full back and center, and then ramp linearly from 0 at center joystick throttle to 1000 at full forward joystick throttle position.
b.        Selecting "Full down stick is zero throttle" results in 500-1000 throttle values in the MANUAL_CONTROL message where 500 = full back joystick throttle position, 1000 = full forward throttle position). It should be 0-1000.

2)        The range of throttle values (between 0 and 1000) in the MANUAL_CONTROL message appear to be non-linear. With the joystick calibrated (I tried this several times), when the joystick throttle is mechanically in the center, and the slider in QGC (joystick settings) is also centered, (and with the problem described in 1) above not fixed), the throttle value in the MANUAL_CONTROL message is 577 (should be 500 if linear) for the "Center stick is zero throttle" setting, and 788 (should be 750 if linear) for the "Full down stick is zero throttle".

3)        If the "Center stick is zero throttle" and “Spring loaded throttle smoothing” are both selected, then the throttle value in the MANUAL_CONTROL message appears to be stuck at 1000 (never changes regardless of the joystick throttle position).

4)        Not sure if this is an issue or not: QGC initially populates the MANUAL_CONTROL message with x=1000, y=-1000, z=1000, r=-1000 (not the actual joystick values) and continues to send these values in this message for some time (I only checked up to 30 seconds). I assume some event causes QGC to send the actual joystick values (and it does), but not sure what that event is. I would expect the actual joystick values would be used from the beginning.

DonLakeFlyer:
@rjehangir Is there anything wrong here?

rjehangir:
[...]

PR #7163, opened 2019-01-19
Joystick: Default to ThrottleModeDownZero

DonLakeFlyer:
Related to #4293.

@jaxxzer Does this affect you guys. Down Zero makes way more sense as default for fixed wing and multi-rotor.

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -71,7 +71,7 @@ Joystick::Joystick(const QString& name, int axisCount, int buttonCount, int hatC
     , _rgCalibration(nullptr)
     , _rgButtonValues(nullptr)
     , _lastButtonBits(0)
-    , _throttleMode(ThrottleModeCenterZero)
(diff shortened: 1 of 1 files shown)


jaxxzer:
I'll check

jaxxzer:
Is this safe/desireable? My understanding was that center is the default so that there is no throttle applied on an idle center-sprung throttle stick (which most gamepads have).

DonLakeFlyer:
The majority use case for joystick is multi-rotor. With a multi-rotor center throttle is essentially hover in most flight modes. That's the reason most commerical multi-rotor controllers work the same way. Spring loaded centered throttle stick.

jaxxzer:
OK. This change is fine/ non-functional for ArduSub. I don't know if this will interfere with takeoff or pre-arm checks on other vehicles, but I suspect it might.

DonLakeFlyer:
You have to hold the throttle down to arm like a regular transmitter.

PR #7178, opened 2019-01-21
Prevent testing nullified camera info request

dogmaphobic:


dogmaphobic:
It looks like Jenkins is not updating the GPS submodule?

PR #7184, opened 2019-01-23
Allow custom fact json files for different vehicles

Williangalvani:
This enables vehicle-specific fact jsons to take care of things like displaying "depth" instead of altitude and "dive time" instead of "flight time".

Also added headingToGCS fact.


DonLakeFlyer:
This isn't a good way to do this for two reason:
* It requires duplicating the entire json file which means that it can easily get out of sync with the origin file.
* Custom builds already have the ability to override fact settings metadata through the QGCCorePlugin::adjustSettingMetaData on a case by case basis

The correct way to do this is to allow firmware plugins to adjust metadata on Fact just like custom builds do.

jaxxzer:
@DonLakeFlyer we've changed it up

DonLakeFlyer:
Still not quite what I'm looking for but close enough for now. This is safer than a full change.

PR #7188, opened 2019-01-24
Add UDP audio support

Williangalvani:
This enables QGC to receive and playback audio via an UDP port. Having an audio stream improves the video experience and is an extra feedback channel for the pilot to better understand what is going on with the vehicle and surroundings.

It adds an audio pipeline with adjustable volume, settings for it, and qml interface to adjust volume.

audio1

audio2

It can be tested with this gstreamer pipeline:
`gst-launch-1.0 -v -e audiotestsrc ! audioconvert ! rtpL16pay ! udpsink host=192.168.15.2 port=5601`

### DIFF
--- a/src/VideoStreaming/VideoReceiver.cc
+++ b/src/VideoStreaming/VideoReceiver.cc
@@ -26,6 +26,7 @@
 #include <QDir>
 #include <QDateTime>
 #include <QSysInfo>
+#include <algorithm>
 
 QGC_LOGGING_CATEGORY(VideoReceiverLog, "VideoReceiverLog")
 
@@ -60,9 +61,12 @@ VideoReceiver::VideoReceiver(QObject* parent)
     , _stopping(false)
     , _sink(nullptr)
     , _tee(nullptr)
+    , _volume(0)
     , _pipeline(nullptr)
     , _pipelineStopRec(nullptr)
     , _videoSink(nullptr)
+    , _audioPipeline(nullptr)
+    , _gstVolume(nullptr)
     , _socket(nullptr)
     , _serverPresent(false)
     , _rtspTestInterval_ms(5000)
@@ -84,6 +88,7 @@ VideoReceiver::VideoReceiver(QObject* parent)
     connect(this, &VideoReceiver::msgStateChangedReceived, this, &VideoReceiver::_handleStateChanged);
     connect(&_frameTimer, &QTimer::timeout, this, &VideoReceiver::_updateTimer);
     _frameTimer.start(1000);
+    updateAudioPipeline();
 #endif
 }
 
@@ -91,6 +96,7 @@ VideoReceiver::~VideoReceiver()
 {
 #if defined(QGC_GST_STREAMING)
     stop();
+    _stopAudio();
     if(_socket) {
         delete _socket;
     }
@@ -467,6 +473,73 @@ VideoReceiver::start()
     _starting = false;
 #endif
 }
+
+#if defined(QGC_GST_STREAMING)
+void
+VideoReceiver::_startAudio()
+{
+    // Run audio
+    uint32_t port = _videoSettings->audioUdpPort()->rawValue().toUInt();
+
+    QString pipeline = QStringLiteral("udpsrc port=%1 "
+        "! application/x-rtp, media=audio, clock-rate=44100, encoding-name=L16, encoding-params=1, channels=1, payload=96 "
+        "! rtpL16depay name=rtp0 "
+        "! audioconvert "
+        "! volume volume=1.0 name=volume0 "
+        "! queue "
+        "! autoaudiosink sync=false").arg(port);
+
+    qCDebug(VideoReceiverLog) << "Opening audio pipeline: " << pipeline;
+
+    _audioPipeline = gst_parse_launch(static_cast<char*>(pipeline.arg(port).toLocal8Bit().data()) , nullptr);
+
+    if (_audioPipeline) {
+        gst_element_set_state(_audioPipeline, GST_STATE_PLAYING);
+
+        _gstVolume = gst_bin_get_by_name(GST_BIN(_audioPipeline), "volume0");
+        qCDebug(VideoReceiverLog) << "volume0:" << _gstVolume;
+        setVolume(_videoSettings->audioVolume()->rawValue().toFloat());
+    }
+}
+
+void
+VideoReceiver::updateAudioPipeline()
+{
+    _stopAudio();
+    if (_videoSettings->audioEnabled()->rawValue().toBool()) {
+        _startAudio();
+    }
+}
+
+void
+VideoReceiver::_stopAudio() {
+    qCDebug(VideoReceiverLog) << "Stopping audio pipeline";
+    if(_audioPipeline) {
+        if(_gstVolume) {
+            gst_object_unref(_gstVolume);
+            _gstVolume = nullptr;
+        }
+        gst_element_set_state(_audioPipeline, GST_STATE_NULL);
+        gst_object_unref(_audioPipeline);
+        _audioPipeline = nullptr;
+    }
+}
+
+void
+VideoReceiver::setVolume(float vol)
+{
+    // TODO: replace by std::clamp after moving to cpp17
+    vol = std::max<float>(0.0, std::min<float>(vol, 1.0));
+
+    if(_gstVolume) {
+        _volume = vol;
+        g_object_set(_gstVolume, "volume", _volume, nullptr);
+        qCDebug(VideoReceiverLog) << "Set volume:" << vol;
+    } else {
+        qCDebug(VideoReceiverLog) << "No volume control";
+    }
+}
+#endif
 
 //-----------------------------------------------------------------------------
 void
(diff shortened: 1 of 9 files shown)


src/VideoStreaming/VideoReceiver.cc
```diff
@@ -468,6 +473,65 @@ VideoReceiver::start()
...
 }
 
+#if defined(QGC_GST_STREAMING)
+void
+VideoReceiver::_startAudio()
+{
+    // Run audio
+    GError *error = NULL;
+    uint32_t port = _videoSettings->audioUdpPort()->rawValue().toUInt();
+    QString pipeline = QStringLiteral("udpsrc port=%1 ! application/x-rtp, media=audio, clock-rate=44100, encoding-name=L16, encoding-params=1, channels=1, payload=96 ! rtpL16depay name=rtp0 ! audioconvert ! volume volume=1.0 name=volume0 ! queue ! autoaudiosink sync=false");
```
jaxxzer: Let's break this line up and add some whitespace in the surrounding code

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -468,6 +473,65 @@ VideoReceiver::start()
 #endif
 }
 
+#if defined(QGC_GST_STREAMING)
+void
+VideoReceiver::_startAudio()
+{
+    // Run audio
+    GError *error = NULL;
```
jaxxzer:
```suggestion
```

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -468,6 +473,65 @@ VideoReceiver::start()
...
+    uint32_t port = _videoSettings->audioUdpPort()->rawValue().toUInt();
+    QString pipeline = QStringLiteral("udpsrc port=%1 ! application/x-rtp, media=audio, clock-rate=44100, encoding-name=L16, encoding-params=1, channels=1, payload=96 ! rtpL16depay name=rtp0 ! audioconvert ! volume volume=1.0 name=volume0 ! queue ! autoaudiosink sync=false");
+    qCDebug(VideoReceiverLog) << "Opening audio pipeline: " << pipeline.arg(port);
+    _audioPipeline = gst_parse_launch(static_cast<char*>(pipeline.arg(port).toLocal8Bit().data()) , &error);
```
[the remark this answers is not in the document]
jaxxzer: Use nullptr in place of NULL.

src/VideoStreaming/VideoReceiver.cc
jaxxzer:
```suggestion
    QString pipeline = QStringLiteral("udpsrc port=%1 ! application/x-rtp, media=audio, clock-rate=44100, encoding-name=L16, encoding-params=1, channels=1, payload=96 ! rtpL16depay name=rtp0 ! audioconvert ! volume volume=1.0 name=volume0 ! queue ! autoaudiosink sync=false").arg(port);
```

Remove arg(port) below

src/VideoStreaming/VideoReceiver.cc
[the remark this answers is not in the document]
patrickelectric: It makes more sense to check the error variable nop ?

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -468,6 +473,65 @@ VideoReceiver::start()
...
+        gst_object_unref(_audioPipeline);
+        _audioPipeline = NULL;
+    }
+}
+
+void
+VideoReceiver::setVolume(float vol)
+{
+    if(vol < 0 || vol > 1) {
+        vol = 1;
```
patrickelectric: Maybe
```
vol = std::clamp(vol, 0, 1);
```

dogmaphobic:
This needs a bit more forethought. This PR only handles UDP audio and it assumes that's the only audio source.

* Audio enable/disable should be more generic as well as volume.
* VideoReceiver should export a property telling it supports audio (something like `hasAudio()`) That way you not only check for the enable/disable flag but also to see if there is any audio to enable/disable.
* This PR only handles UDP audio while RTSP is by far more common. If audio is something desirable, it would be nice to have support for RTSP as well.
* The setting for enable/disable should only be shown if the underlying protocol supports it.

jaxxzer:
Are these things we can work up to? Or does it all need to happen at once? Maybe some audio is better than no audio for the moment.


Can you elaborate more on 'should be more generic' and 'if the underlying protocol supports it'? Do we have any protocols other than rtsp and udp?

jaxxzer:
We can certainly help with the rtsp stream, we just need to play with it and get it going over here.

dogmaphobic:
>Are these things we can work up to? Or does it all need to happen at once?

Technically yes. This can be safely merged and we can handle the rest later.

>Can you elaborate more on 'should be more generic' 

Instead of `audioUdpVolume`, it should be just `audioVolume`. Same for `audioUdpEnabled`. Change it to `audioEnabled`. That would make these "generic". You would not be using UDP audio with any other pipeline other than RTP UDP. That way, when you set enable/disable or volume, it's for the _current_ pipeline (if VideoReceiver::hasAudio() is true). For now we have:

- RTSP
- UDP RTP
- TCP
- MPEG TS
- Taisync

RTSP, TCP and Taisync all handle both video and audio using a single connection, though in its current form, Taisync does not support audio. MPEG TS is a weird one off, way too specialized and I doubt it will ever be used on a regular basis. 

Given the above, the only option specific to UDP is when using UDP RTP. That's the only time you should see "UDP Audio Port" as an option.

Also note that there is no provision for audio within the MAVLink video API. For a vehicle that properly implements a camera interface with video streaming, this would only apply to multiplexed connections (RTSP and TCP). VideoReceiver can detect if the stream has audio and set the `hasAudio` to true.

With all that said, I can easily fix all of this after this is merged.

dogmaphobic:
Oh, and we also have UVC video, which is in parallel and independent from GStreamer. That could potentially also have audio.

jaxxzer:
> Instead of audioUdpVolume, it should be just audioVolume

> it's for the current pipeline 

Oh, of course!

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -468,6 +474,73 @@ VideoReceiver::start()
...
+
+    _audioPipeline = gst_parse_launch(static_cast<char*>(pipeline.arg(port).toLocal8Bit().data()) , nullptr);
+
+    if (_audioPipeline) {
+        gst_element_set_state(_audioPipeline, GST_STATE_PLAYING);
+
+        _gstVolume = gst_bin_get_by_name(GST_BIN(_audioPipeline), "volume0");
+        qCDebug(VideoReceiverLog) << "volume0:" << _gstVolume;
+        setVolume(_videoSettings->audioVolume()->rawValue().toFloat());
+    }
```
jaxxzer: We should print an error message if it fails.

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -468,6 +474,73 @@ VideoReceiver::start()
...
+        "! queue "
+        "! autoaudiosink sync=false").arg(port);
+
+    qCDebug(VideoReceiverLog) << "Opening audio pipeline: " << pipeline;
+
+    _audioPipeline = gst_parse_launch(static_cast<char*>(pipeline.arg(port).toLocal8Bit().data()) , nullptr);
+
+    if (_audioPipeline) {
+        gst_element_set_state(_audioPipeline, GST_STATE_PLAYING);
+
```
patrickelectric: Print GError message here. Check others places that GError are being ignored.

src/VideoStreaming/VideoReceiver.cc
jaxxzer: advice still applies

Williangalvani: Do you mean a `qgcApp()->showMessage()` ?

jaxxzer:
```suggestion
    } else {
        qCDebug(VideoReceiverLog) << "Failed to get audio pipleine"
    }
```
or similar

Williangalvani: I don't expect both `_audioPipeline` and `error` to be null at the same time.
isn't [line 497](https://github.com/mavlink/qgroundcontrol/pull/7188/files#diff-ff57b92b141a56c61545a49ceedbe906R497) enough?

Williangalvani: Done.
I expected that if `error` was set then `_audioPipeline` would be `NULL`, but looking again at the docs that is not always the case, my bad.

DonLakeFlyer:
This shouldn't be merged until I snap the next stable branch

Jaeyoung-Lim:
Any updates on this?

espresso0000:
Is there any audio support in any of the QGC?

jaxxzer:
> Any updates on this?

>Is there any audio support in any of the QGC?

There is no audio streaming support in QGC. It will not come until after 3.6 is released at the earliest. This will also have to wait until after the flux in the video streaming code is sorted: https://github.com/mavlink/qgroundcontrol/issues/7842

joecfd:
Is any audio support for rtsp streaming?

PR #7190, opened 2019-01-25
VideoStreaming: Check for gstreamer in pkg-config in mac

patrickelectric:
Use pkg-config to find for gstreamer installation 

### DIFF
--- a/src/VideoStreaming/VideoStreaming.pri
+++ b/src/VideoStreaming/VideoStreaming.pri
@@ -37,6 +37,13 @@ LinuxBuild {
         CONFIG      += VideoEnabled
         INCLUDEPATH += $$GST_ROOT/Headers
         LIBS        += -F/Library/Frameworks -framework GStreamer
+    } else {
+        CONFIG      += link_pkgconfig
+        packagesExist(gstreamer-1.0) {
+            #- gstremaer framework in pkg-config
+            CONFIG    += VideoEnabled
+            PKGCONFIG += gstreamer-1.0 gstreamer-video-1.0
+        }
     }
 } else:iOSBuild {
     #- gstreamer framework installed by the gstreamer iOS SDK installer (default to home directory)


patrickelectric:
@dogmaphobic ^

patrickelectric:
@dogmaphobic 	^

patrickelectric:
@dogmaphobic 	^

dogmaphobic:
I don't ever use Homebrew to install GStreamer. I use several versions of GStreamer and have them all installed individually, only changing a symlink to point to the desired version. I do that with all platforms.

I need to find some time to install it using brew and see how the whole process behaves. With that said, I don't like the idea to have arbitrary versions of dependent libraries. We need a consistent set of dependencies so when an error comes up, finding out that it is because you are using a much more recent (or older) version of some dependent library becomes quite hard to track it.

patrickelectric:
@dogmaphobic 	Tracking old gst versions as like Qt version is just wrong for development proposers. This just makes the development and update processes harder and also increases the necessary time to update this kind of packages.

I have done a number of contributions to push QGC compatibility to the last Qt version (since I just update everything for each release always). The same thing applies with GST.

I'm not removing the usage of old and bug-ish gstreamer versions, this PR only allow others developers to use homebrew packages, that I believe is 90% of the mac developers. 

Can you point **how this package can break your workflow** ?
Can you also point why developers or anyone that want to build  QGC from scratch **can be harmed from this patch** ?
Do you want a warning message for the pkg-config if case to alarm the user ?
Also, this package will use the hardcoded path as default, after that will check for pkg-config and do a check if the package exist. And I'm pretty sure that homebrew does not have old gst packages (I'm thankful for that).

This exactly same comment applies to weird/wrong/hardcoded stuff around the code:
https://github.com/mavlink/qgroundcontrol/pull/7222/files

dogmaphobic:
I missed the whole file and only saw the change. I assumed it was "replacing" the existing logic. No, I see no problem with this change as it is.

>Do you want a warning message for the pkg-config if case to alarm the user ?

A message, showing the non standard version in use is a good idea. When we ship a stable release of QGC, it is based on known versions of dependencies. If you later build your own QGC using a new version of some external library and that causes an issue, there is no easy way to determine the source of the problem. Invariably you will see issues being reported only to find out people are using unsupported/untested versions or various things (Qt, GStreamer, etc.) A _Stable_ release by definition is based on known and tested dependencies. Very specific versions of these dependencies.

>this PR only allow others developers to use homebrew packages, that I believe is 90% of the mac developers.

Only developers who know what they are doing. I would say the opposite is true. The majority of the cases are from people who clone the repo and try to build it. It then doesn't work and they get frustrated because they are using an untested/unsupported version of some dependency.

DonLakeFlyer:
> Only developers who know what they are doing. I would say the opposite is true. The majority of the cases are from people who clone the repo and try to build it. It then doesn't work and they get frustrated because they are using an untested/unsupported version of some dependency.

My feeling is that the QGC build files should be locked to the version dependency that is currently supported/tested in the releases which are automatically dropped. By that I mean a hard stop, build fails if wrong dependency. Anyone who knows what they are really doing can modify the build files to relax this. The reason is that I spend a large amount of time dealing with people screwing up when building the product themselves and really don't know what they are doing. This is the 99% use case as far as I can see. I don't want to make it any worse than it already is.

tcanabrava:
@DonLakeFlyer , @dogmaphobic I agree with you guys that for a *tagged* release the version should be set, but for *master* it's counterproductive. 
on cmake you can specify find_package(blah VERSION EXACT) that will only configure if that *specific* version of the package is found, or find_package(blah) for packages that don't really need to have this hard dep.

This way we have the best of worlds:
1 - We can use brew / pacman / whatever to install the package 
2 - We can use find_package from within cmake to look for the installed binaries
3 - we can die hard if, in a released tag, the package is wrong
3 - we can allow it to pass in a development build as the developer should know what he's doing.

PR #7192, opened 2019-01-28
QGCExternalLibs: Test Qt version before adding AirMap

patrickelectric:
This is necessary to avoid problems with others Qt versions.


dogmaphobic:
For the short term, this is enough. For the long term, once Airmap finishes the SDK, we need to decide if we're going to continue with binaries or having it as a submodule. If we continue with binaries, Qt versioning will need a bit more care.

PR #7203, opened 2019-01-31
Don't automatically enable joystick updates to vehicle

### LINKED ISSUE #7182: Don't automatically enable joystick after calibration.
I think 'Click Next to enable joystick' should be changed to 'Joystick calibration complete, check enable joystick to begin sending joystick updates to the vehicle.'

Calibrating and enabling output should be two independent steps that the user triggers.

jaxxzer:
fix #7182

- The joystick updates to the vehicle will be disabled during calibration.
- The user will not have to click 'next' after the last step of calibration
- When the calibration is finished, the user will have to click 'enable' themselves to start sending data to the vehicle. This means they can confirm the calibration is correct before enabling the updates.

### DIFF
--- a/src/VehicleSetup/JoystickConfig.qml
+++ b/src/VehicleSetup/JoystickConfig.qml
@@ -364,9 +364,12 @@ SetupPage {
 
                             QGCCheckBox {
                                 id:         enabledCheckBox
-                                enabled:    _activeJoystick ? _activeJoystick.calibrated : false
-                                text:       _activeJoystick ? _activeJoystick.calibrated ? qsTr("Enable joystick input") : qsTr("Enable not allowed (Calibrate First)") : ""
-                                onClicked:  _activeVehicle.joystickEnabled = checked
+                                enabled:    _activeJoystick ? (_activeJoystick.calibrated && !controller.calibrating): false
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
How safe is this. I'm wrapping up Stable. Not sure about putting this in there.

DonLakeFlyer:
> * The user will not have to click 'next' after the last step of calibration
> * This means they can confirm the calibration is correct before enabling the updates.

The concept of the last step is that is gives the user the ability to check the calibration and cancel out of a bad calibration. This matches how RC cal works. Not sure how this new way is better? I would guess it's going to generate a lot of "I calibrated my joystick and it sill doesn't work complaints" because people forget to enable.

jaxxzer:
> The concept of the last step is that is gives the user the ability to check the calibration and cancel out of a bad calibration.

Ok, I guess I had not noticed this, and I was achieving the same effect through a different means.

> Not sure how this new way is better?

1. The user will decide when it is ok to send joystick commands to the vehicle. In the current version it is unclear that that will automatically happen immediately after clicking 'next'. I consider this patch safer than the current behavior. When they are finished with the calibration, they may verify everything is ok before enabling the commands to the vehicle, and recalibrate if it is not ok.
2. The purpose and function of the 'enable' checkbox is explained to the user so they will understand how this works better. I would expect that this clarification will reduce the amount of support/questions regarding this portion of setup.

Issue #7204, opened 2019-02-02
mission transfer error when uploading to arducopter 3.6.5 (and 3.6.4)


maciek01:
The following template is for Issue/Bug reporting only. https://docs.qgroundcontrol.com/en/Support/Support.html#github-issues

For questions about how to use or build QGC see: http://qgroundcontrol.com/#resources

If you are using an old stable release please upgrade first and verify your issue before reporting.
----

## Expected Behavior
Please tell us what should happen as detailed as possible.

should be able to upload this mission.

this mission used to upload fine. I recently added gimbal angle controls. after that i am unable to upload.
I dont have the good version of the mission unfortunately.

## Current Behavior
Please describe what happens instead of the expected behavior.

upload fails with "mission transfer error. retry transfer. Error: Vehicle returned error: X/Param5 has invalid value (MAV_MISSION_INVALID_PARAM5_X). Param5: nan."

when u reload mision from vehcle it shows just a WP 3.

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1. upload mission file to QGC
2. select upload button
3. error pops up
4.

## System Information
When posting bug reports, include the following information
- Operating System: Ubuntu 18.4
- QGC Version: HEAD:f585fd801 2019-01-23 23:41:10 +0000
- QGC build: daily
- Flight Controller: Pixhawk 2
- Autopilot (with version): Ardupilot Copter 3.6.5

## Detailed Description
Provide further details about your issue/bug.

## Log Files and Screenshots
- [QGC Console Logs](https://docs.qgroundcontrol.com/en/Support/Support.html#reporting-bugs)
- Autopilot logs when available (post a link).
- Screenshots of QGC to help identify the current issue/bug behavior.

[wilmington3.plan.txt](https://github.com/mavlink/qgroundcontrol/files/2824829/wilmington3.plan.txt)
![screenshot from 2019-02-02 15-08-37](https://user-images.githubusercontent.com/3790977/52169089-d182bc00-2700-11e9-9b20-86b95ea15028.png)

DonLakeFlyer:
Hmm, there is a MAV_CMD_SET_CAMERA_MODE command in there somehow. That is not supported by ArduPilot. Could you have initially created this plan while not connected to a vehicle and had the Vehicle Info set to PX4 Pro at the time and then checked the Mode checkbox in the Camera section.

[...]

PR #7205, opened 2019-02-03
Plan-Waypoint: Show camera mode if plan contains it

DonLakeFlyer:
Normally the Camera Mode setting is not shown for ArduPilot since it is not supported. But if a plan file has a MAV_CMD_SET_CAMERA_MODE even if it is ArduPilot that setting should be shown so the user can see it and change it.


DonLakeFlyer:
Related to #7204

PR #7210, opened 2019-02-04
QGCTextField: Fix crappy font rendering on Windows 

DonLakeFlyer:


DonLakeFlyer:
@AndKe I think this finally fixes the crappy font rendering of text fields on Windows. There should be new Stable 3.5 available by tomorrow with it in it. Could you give it a try?

AndKe:
I just tried current master, on Linux - (as always)  - I think you may have meant "Linux" .
Anyway: all text seems to fit just fine, regardless of mission item type.

PR #7215, opened 2019-02-07
OfflineMaps: Fix tile set counts

DonLakeFlyer:
If you created a new tile set and included elevation data the tile counts stored in the two separate tile sets would be the total tile count for both image and elevation tiles. This would cause the system to think download was incomplete if you left the offline maps page and went back to it.

The fix is to maintain the tile counts separately for image and elevation tile sets and to store the appropriate value for each in the database.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,9 @@ Note: This file only contains high level features or important fixes.
 
 ## 3.5
 
+### 3.5.1 - Not yet released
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
I had to remove the tileXY0 properties since they were no longer calculcated. I"m not sure they made any sense in previous code anyway. Since they would be the union of the image and elevation tiles. For which I'm not sure there is any value. Also the properties weren't used anywhere I could find anyway.

DonLakeFlyer:
I'm going to merge this in so Sentera can provide testing. If it's wrong i'll fix after the fact.

PR #7216, opened 2019-02-08
Change joystick deadband adjustment to dragging horizontally

### LINKED ISSUE #7214: Deadband adjustment by dragging vertically is not working
The instructions say to drag vertically on top of the axis monitor to adjust deadband.

I believe that because now the page is too tall with all the button options, the scrolling has been enabled, so when I drag vertically, instead of adjusting the deadband I scroll the page.
deadband2

Tested using master in Arch Linux

Williangalvani:
This changes the deadband adjustment from dragging vertically to dragging horizontally.
deadbandpr

Fix #7214

### DIFF
--- a/src/VehicleSetup/JoystickConfig.qml
+++ b/src/VehicleSetup/JoystickConfig.qml
@@ -849,20 +849,24 @@ SetupPage {
 
 
                                 MouseArea {
-                                    id:             deadbandMouseArea
-                                    anchors.fill:   parent.item
-                                    enabled:        controller.deadbandToggle
+                                    id:                 deadbandMouseArea
+                                    anchors.fill:       parent.item
+                                    enabled:            controller.deadbandToggle
+                                    preventStealing:    true
 
-                                    property real startY
+                                    property real startX
+                                    property real direction
 
                                     onPressed: {
-                                        startY = mouseY
+                                        startX = mouseX
+                                        direction = startX > width/2 ? 1 : -1
                                         parent.item.deadbandColor = "#3C6315"
                                     }
                                     onPositionChanged: {
-                                        var newValue = parent.item.deadbandValue + (startY - mouseY)*15
+                                        var mouseToDeadband = 32768/(width/2) // Factor to have deadband follow the mouse movement
+                                        var newValue = parent.item.deadbandValue + direction*(mouseX - startX)*mouseToDeadband
                                         if ((newValue > 0) && (newValue <32768)){parent.item.deadbandValue=newValue;}
-                                        startY = mouseY
+                                        startX = mouseX
                                     }
                                     onReleased: {
                                         controller.setDeadbandValue(modelData,parent.item.deadbandValue)


DonLakeFlyer:
You just need to set preventStealing: true on the MouseArea to fix this. That keeps the scroll area below it from stealing the mouse drags. That said, do you know of there was reason for it to drag vertical instead of horizontal? If now you should figure out who did than and ping them here as to why? There may be something important there.

To me the horizontal drag seems to make more sense.

Williangalvani:
Dragging horizontally makes more sense to me too.
@nanthony21 Was there a specific reason for dragging vertically here https://github.com/mavlink/qgroundcontrol/commit/41a994128311ce92d0cdbfe7726493adbe4b3227?

nanthony21:
At the time it seemed like doing it vertically was easier when I tested on a touch screen. Nothing wrong with doing it horizontally though, you just might want to test it out on a tablet first.

DonLakeFlyer:
@Williangalvani Do you have the ability to check android tablets?

Williangalvani:
@DonLakeFlyer I do have an Android tablet here, I'll try to set up the build environment again. I think I had it working on another machine.

Williangalvani:
I had a bad timing trying to build QGC for my X86 android tablet. I'll probably get an ARM one in the next weeks, but if someone could test it before, that would be nice.

DonLakeFlyer:
@Williangalvani Did you ever get a tablet. I can't test this myself since I don't have an android joystick o test with.

Williangalvani:
> @Williangalvani Did you ever get a tablet. I can't test this myself since I don't have an android joystick o test with.

Yes, I got it. I added this to my to-do list for the week.

Williangalvani:
I got my setup working.

something I didn't realize is that my way is a little counter-intuitive that in the right-side of the bar the controls work inverted. 

Other than that it works ok. You can actually move your finger down while dragging so you can see the bar moving when you move the finger sideways.

The vertical control has the same issue of needing to take the finger out of the way, but maybe it is more intuitive in the sense that up is more, down is less (there is no inversion). 

I'm also thinking of adding an option to use the mouse wheel too since that is more intuitive for desktop users.

What do you think is the best way to go?

DonLakeFlyer:
> something I didn't realize is that my way is a little counter-intuitive that in the right-side of the bar the controls work inverted.

I don't understand what you mean by that.

> I'm also thinking of adding an option to use the mouse wheel

Mouse wheel has a tendency to screw things up since it is also used to scroll the entire view up/down when the window is small.

Williangalvani:
> I don't understand what you mean by that.

Left side makes sense, right side not so much. 
joystick
If you agree dragging horizontally is better, I'll fix it. Otherwise I'll go the `preventStealing: true` route.


> Mouse wheel has a tendency to screw things up since it is also used to scroll the entire view up/down when the window is small.

Good point.

DonLakeFlyer:
Got it. Left/Right should be the same. Drag outward from the center point bar gets bigger.

DonLakeFlyer:
I think horizontal is probably ok.

Williangalvani:
Fixed it.
joystick done

I'll test in android to make sure it works ok there too.

Williangalvani:
Works fine in Android too.

PR #7220, opened 2019-02-10
Switch Windows to VC 2017 64 Bit

DonLakeFlyer:
Also fixes for the installer:
* QGC version is shown in the installer
* QGC Icon is shown in control panel add/remove programs
* Better information for add/remove programs entry


src/VideoStreaming/VideoStreaming.pri
```diff
@@ -70,7 +70,6 @@ LinuxBuild {
 
         # Copy GStreamer plugins
         QMAKE_POST_LINK += $$escape_expand(\\n) xcopy \"$$GST_ROOT_WIN\\lib\\gstreamer-1.0\\*.dll\" \"$$DESTDIR_WIN\\gstreamer-plugins\\\" /Y $$escape_expand(\\n)
-        QMAKE_POST_LINK += $$escape_expand(\\n) xcopy \"$$GST_ROOT_WIN\\lib\\gstreamer-1.0\\validate\\*.dll\" \"$$DESTDIR_WIN\\gstreamer-plugins\\validate\\\" /Y $$escape_expand(\\n)
```
DonLakeFlyer: @dogmaphobic These "validate" dlls don't exist in the 64 bit version. Is there a reason why they are needed? Things still compile fine.

DonLakeFlyer: @jaxxzer Not sure if you may know?

jaxxzer: Oof, I guess they are not needed, but I really have no clue. To my knowledge, we aren't using it.

dogmaphobic:
I can't remember. It's been almost 3 years since I added that to Windows. What I do remember was going through the build and adding things as needed. That is, the build would fail, I would see what was missing and adding. It could be that we are now using a much newer version of GStreamer? That was all done for 1.5.2. I think we are now using 1.14.4?

DonLakeFlyer:
This still says 1.5.2 is being used: https://github.com/mavlink/qgroundcontrol/tree/master/src/VideoStreaming?

DonLakeFlyer:
Which is what I compiled with

PR #7221, opened 2019-02-10
Appveyor: Switch to Stable 3.5

DonLakeFlyer:


DonLakeFlyer:
@dagar Maybe this was preventing the new WIndows Stable from being pushed up. Not quite sire, but don't think so...

PR #7224, opened 2019-02-12
Fix cmake building

sithlord48:
---
name: Fix building with Cmake
---

**Steps to fix building**
1. Remove -Werror compiler flag (not in qmake file and cause build to fail)
2. Correctly detect and link SDL2 on Linux

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,6 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 	add_compile_options(
 		-Wall
 		-Wextra
-		-Werror
 
 		-Wno-address-of-packed-member # ignore for mavlink
 	)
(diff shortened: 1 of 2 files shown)


dagar:
Thanks @sithlord48. If it's not obvious cmake isn't fully supported yet, but I think it's definitely worth transitioning. 

Did you see the CI output? http://ci.px4.io:8080/blue/organizations/jenkins/mavlink%2Fqgroundcontrol/detail/PR-7224/1/pipeline

image

sithlord48:
So strange as this works on my distro perfectly with SDL2::SDL2 linked and will not build if i revert to the old linking with SDL_LIBRARIES

dagar:
@sithlord48 what's your environment? Are any of the other SDL2 variables populated? SDL2_INCLUDE_DIR, etc?

sithlord48:
I'm on arch.
I think I may have found the issue. Looks like sdk2 has a bug around 2.0.8 where the returned Cmake dir is empty. I've tested here and the only thing returning a proper path is "SDL2_DIR" . Linking for me is only working with SDL2::SDL2

I've changed the patch to check for an empty SDL2_INCLUDE_DIRS and if found use SDL_DIR and link to SDL2::SDL2 . This is working locally for me . Lets see if we pass the bot test this time.

PR #7237, opened 2019-02-16
Check box check mark using vector image

dogmaphobic:
It now looks like this:

screen shot 2019-02-16 at 5 28 42 pm

screen shot 2019-02-16 at 5 28 49 pm


DonLakeFlyer:
Thanks Gus. Any ideas on PartiallyChecked? The opacity I think sort of worked with the box. But not sure about the check.

DonLakeFlyer:
That said I think this is good enough for now.

PR #7245, opened 2019-02-21
Feature: add option to hide MAVLink Logging from settings

stefandunca:
The option flag is accessible from QGCOptions


DonLakeFlyer:
Interesting I guess but why? Would rather keep things simple.

DonLakeFlyer:
Gus explained. This is fine.

PR #7246, opened 2019-02-22
Added more cameras

dogmaphobic:
* Added a few more Sony and Yuneec cameras 
* Made camera description indicate lens focal length and aperture
* For zoom lenses, the indicated focal length (and the data) is the lens in its widest focal length

### DIFF
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -347,185 +347,322 @@ const QVariantList& FirmwarePlugin::cameraList(const Vehicle* vehicle)
     if (_cameraList.size() == 0) {
         CameraMetaData* metaData;
 
-        metaData = new CameraMetaData(tr("Sony NEX-5R 20mm"),  //http://www.sony.co.uk/electronics/interchangeable-lens-cameras/ilce-qx1-body-kit/specifications
-                                      23.2,                 //http://www.sony.com/electronics/camera-lenses/sel16f28/specifications
-                                      15.4,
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
You can't change the name of these. It will break existing Plan files which reference them by name.

dogmaphobic:
Oops...

Some of the names were just wrong, while others needed further clarification. For instance, when using something with a zoom lens, it needs to be specified that the settings only apply for when the lens is set to its widest (shortest) focal length. Can we add an ID in addition to the name? If this were to ever be localized, that would break this as well. 

I guess it's time to take up on your idea of using SQL for these things...

DonLakeFlyer:
An ID would be better but not trivial to bolt in now. Localization only breaks if you hand Plan files between people on different locales. In that case it will fall back to Custom Camera with the settings for the Camera that it couldn't find. Nothing is really lost just some annoyance. But if the names change then that same fall back will trigger if you load a Plan file from one rev ot QGC to the newer on.

DonLakeFlyer:
@dogmaphobic Can you switch the name back?

dogmaphobic:
Some or all? Some were just wrong: _ Sequioa_? How is the user to know what lens or focal length does the data applies to?

DonLakeFlyer:
All. None of the names can be changed right withouy breaking Plan files. I'll figure out how to fix this up with the big Camera pull I'm working on. But that is a ways out.

dogmaphobic:
https://github.com/mavlink/qgroundcontrol/pull/7256

PR #7250, opened 2019-02-26
Localization work

dogmaphobic:
Fix error in German Translation. The original translation went from `Firmware Files (*.px4 *.bin *.ihx)` to `Firmware-Dateien (*.px4.bin *.ihx)` (messing with the `*.px4` and `*.bin` extensions.)

I've also updated several languages in the process (all in various levels of incompleteness).

Added a missing `.so`

This needs some testing on Linux.


DonLakeFlyer:
These things will need to go into Stable_V3.5.1 as well since they are pretty broken right?

dogmaphobic:
The only real fix was the mess done with `Firmware Files (*.px4 *.bin *.ihx)` to `Firmware-Dateien (*.px4.bin *.ihx)`

I made separate commits for this reason. If anything, I would only cherry pick that one commit. The missing .so stopped the error output to the console but I saw no real difference at runtime. It would need some more testing.

PR #7256, opened 2019-02-28
Revert Camera Names

dogmaphobic:
Revert the camera names to their original names to avoid messing up old plan files, which use the camera name string to match the camera to the plan.


dogmaphobic:
Is Jenkins out of sync?

DonLakeFlyer:
> Is Jenkins out of sync?

Yes. I'm not quite sure what file you change which you add new sources.

PR #7257, opened 2019-02-28
cmake add APMMavlinkStreamRateSettings.cc to Settings

dagar:


DonLakeFlyer:
Ah, their in the individual directories. No wonder I couldn't find them.

dogmaphobic:
Note that the addition of `APMMavlinkStreamRateSettings.cc/h` broke the builds where ArduPilot is disabled. I'm about to push a PR excluding all this code for builds that do not support ArduPilot. This will need to be reflected in cmake as well.

dagar:
Tag me and I'll make the cmake change when I'm back.

dogmaphobic:
Speaking of cmake, how does it handle "custom" builds?

dogmaphobic:
I guess it doesn't. It's even hardcoded to "ardupilotmega". This doesn't seem it will work for custom builds where a lot of what is included/built can be changed based on custom configurations.

dagar:
We should go over the full list for feature parity.

PR #7262, opened 2019-03-04
Add support for AppImage customization

stefandunca:
Add optional support to source binaries names, title and icon from ENV
Also copy all dependencies of ICU libs to cover all configurations
please review: @dogmaphobic @DonLakeFlyer @dagar


dagar:
Looks good to me if the existing builds are still working.

dogmaphobic:
> Looks good to me if the existing builds are still working.

I can't easily test that here but it looks no different than the current Windows installer setup. Merging it now and we deal with it later if somethings go bonkers.

PR #7276, opened 2019-03-07
Introduce custom maps in QGC utilizing Mapbox styles

OlavAa:
---
name: New feature: Custom maps in QGC based on Mapbox' "style"

---

**Describe the feature**
New feature: Enable use of styled (customized) Mapbox maps. 
- Introduce two new attributes (Mapbox account name, Mapbox style ID)
- Capture values in settings for offline maps
- Introduce new API in QGCMapUrlEngine.cpp, to retrieve "[styled maptiles](https://docs.mapbox.com/api/maps/#retrieve-raster-tiles-from-styles)"

**User story**
1. User selects: Settings > Offline Maps > Options
2. User enters: 
    Mapbox Access Token
    Mapbox User Name (new)
    Mapbox Style ID (new)
    Clicks "Save"
3. Still in Settings, user selects: General
4. In Miscellaneous section, user selects:
    Map Provider = Mapbox
    Note: Bug in QGC: User must restart app before Mapbox is selectable. Not introduced by me.
5. User then enjoy a custom map as pr design in his account in "Mapbox Studio"

**Expected behavior**
The map made available is as designed in Mapbox Studio (section: Styles)
Suggested implementation is simplified: If new attributes are filled in, QGC will disregard the map type selection for Map provider Mapbox and download styled tiles instead. 

**Log Files and Screenshots**
*Always* add screenshots and the QGC console log.

**Drone (please complete the following information):**
- Feature of interest for any dronetype
- Photo of the IMU / autopilot setup if possible.

**Additional context**
Feature need triggered during [discussions for Ardurover](https://discuss.ardupilot.org/t/mapping-a-lake-with-ardupilot/32401), where new ability for mapping depth/position from NMEA data captured by a transducer was presented. The resulting bathymetric map was obviously of interest to use as base map in the GCS. The now retired Tower app had abilities to use Mapbox project, [now faulty](https://discuss.ardupilot.org/t/mapping-a-lake-with-ardupilot/32401/62). A proof of concept presented was later presented for [how to realize this in QGC](https://discuss.ardupilot.org/t/mapping-a-lake-with-ardupilot/32401/80). General feedback is that this feature should be made available in GCS in general, Michael Osborne (MP) was said to have shown interest. 
I personally have this feature available as an app on several devices and am happy. It has been a challenge since I'm a newbie. Now I'm finally trying to make a pull request to provide the feature for QGC users in general.
I do however have an opinion that the solution in this pull request is an "initial" (but fully working) solution. [A superior user experience would be possible with a different implementation as elaborated here](https://discuss.ardupilot.org/t/mapping-a-lake-with-ardupilot/32401/96). To create this I would however need to make more challenging changes in how map types are selected. I also expect other developers might have strong opinions about such changes. So I decided to create this request for now.
Please note that I have not been able to figure out how to run [these described tests yet](https://dev.qgroundcontrol.com/en/contribute/unit_tests.html). Any advice for a newbie like me is highly appreciated.


DonLakeFlyer:
@dogmaphobic I don't know much about the mapbox stuff. Not sure if this breaks existing mapbox support or makes it more confusing?

OlavAa:
> @dogmaphobic I don't know much about the mapbox stuff. Not sure if this breaks existing mapbox support or makes it more confusing?

User must enter his account name and style ID in addition to mapbox token to enable functionality and is not likely to do that for no other reason than to use his style. It does not break existing functionality, but it does override the maptype selection. If user wants to switch to another mapbox alternative, one of the new attributes would have to be deleted. That is probably confusing, and not very good usability.
Would it be better to define the style option as a new map provider? And to hide that provider in FlightMapSettings.cc if any of token, account or style ID is left blank?

OlavAa:
Clicked the wrong button, sorry.

DonLakeFlyer:
@OlavAa Can you include a screen shot of the new MapBox settings UI?

OlavAa:
Of course. Here's a screenshot from an android tablet.
IMG_20190323_172224

dogmaphobic:
Hold on to Mapbox stuff. I've been in contact with Mapbox about implementing a new, Mapbox specific API altogether. Also note that these "other" maps are not even visible within the map view. Only for offline maps, which makes them unreachable. Map settings (`FlightMapSettings`) was written after the implementation of the map engine but it has a limited, hard-coded set of options. All the extra options are not exposed and you cannot use them in the map view.

OlavAa:
@dogmaphobic New API for Mapbox sounds promising. But note that I try to utilize a Mapbox user's own designs, not the off the shelf "others" already described in the map engine. A good implementation would in my opinion only require the user to add his Mapbox user name in addition to the token. [API to get the style name and IDs](https://docs.mapbox.com/api/maps/#list-styles) exists, and the response "name" could populate the map type dropdown. The [API to download the desired styled maptiles](https://docs.mapbox.com/api/maps/#retrieve-raster-tiles-from-styles) is the one I've used for this more-of-a-hack (to overcome the FlightMapSettings restrictions) suggested pull request.
Is something like this within reach of how you see Mapbox being utilized?

OlavAa:
Sample screenshot from an Android phone displaying a 3D-effect depth map included into a Mapbox style. One obvious use case for Qgroundcontrol boat user.
Screenshot_20190323-215555

DonLakeFlyer:
@dogmaphobic So what should we so with this? It seems like a nice new capability?

dogmaphobic:
I'm meeting with Mapbox next week to discuss a "proper" implementation in QGC. That would mean the ability to use standard maps for general, open source use cases, and specialized maps (and or use cases) for commercial users using their own accounts. That would mean switching away from our own QtLocation plugin to the Mapbox Qt plugin. If that works out, all the work in this PR will still be applicable but the implementation will be different.

OlavAa:
Thanks, this effort is highly appreciated!

DonLakeFlyer:
Questions:
* Does this switch all map stuff to MapBox or just add the ability to use MapBox as well as Google/Bing?
* Would this new MapBox support allow users to add custom maps somehow, not just commercial people?

OlavAa:
I guess clarifications wrt Mapbox and APIs are due shortly. Imo: Mapbox will stay as an alternative provider. And with "commercial" actor I expect anyone that has created a Mapbox account (needed to customize maps in their Studio tool) also can utilize their maps from QGC. Mapbox has payment plans for their services, but the most simple (entry) plan does include enough offerings for free to handle my needs at least.

OlavAa:
@dogmaphobic Any update on a solution for Mapbox?

seangrant82:
@dogmaphobic any updates on implementing the Mapbox Qt plugin? Would this in theory allow full vector tile support with layering?

jh7rnn:
Any updates on Mapbox user styles integration?

PR #7288, opened 2019-03-14
Mission (programmed flight) Simulation mode

GlidLov:
A useful function to add would be a "programmed flight Simulation mode" which lets user to virtually flight a mission like it's show during actual flight, but while FC or whole drone is on bench, showing images footprints in case of survey auto trigger. Sensefly ground control app has this, you can try to compare.
Also, during the simulate path, real mav (or other output) command should be sent to gimbal and camera so, if connected, payload would act as it would in flight.


DonLakeFlyer:
Was this submitted by mistake?

GlidLov:
why? i thought PRs are for requesting new features, am i wrong?

DonLakeFlyer:
A pull request is for submitting actual code changes. An Issue is used for a feature request.

PR #7294, opened 2019-03-19
Misprints in parameters naming and description

EgoPingvina:
---
name: smal misprints in parameters

---


dogmaphobic:
Thanks for catching them but these parameter metadata files are autogenerated. The fixes must go to the sources:

Here: https://github.com/PX4/Firmware/blob/20f870137b66c61df0ba0dcd6a96a8deab01f92e/src/modules/fw_att_control/fw_att_control_params.c#L362

And here: https://github.com/PX4/Firmware/blob/c303e36ed0834e9d19ccb611979868b4287035e9/src/modules/vtol_att_control/standard_params.c#L114

EgoPingvina:
ok, i`m going to fix it in this sources

PR #7300, opened 2019-03-21
Add a language selector to the General Settings

dogmaphobic:
You can now select the localization of QGC. By default, it will try to use the locale given by the underlying OS.

Screen Shot 2019-03-20 at 10 37 24 PM


hamishwillee:
Nice. Certainly make testing easier.

Issue #7323, opened 2019-03-26
Taking pictures crashes on cameras without video streams


tpwrules:
I am developing a camera but for various reasons it will not offer a stream preview in QGC. Unfortunately, this means that asking a picture to be taken by clicking on the red button crashes QGC. The issue appears to be these lines:
https://github.com/mavlink/qgroundcontrol/blob/6af7e8f2f40f390bb17f3b56e6c910010165787e/src/Camera/QGCCameraControl.cc#L366-L371

If no stream has been set up, something in line 370 ends up null (I would assume the return value of videoReceiver()) , so attempting to call a function on it crashes. Commenting out that line fixes the problem. This issue probably also happens on line 1477. There probably needs to be an option to disable local image capture, and enforce disabling it when the camera does not offer a video stream. This is already checked for in QGCCameraControl::_checkForVideoStreams().

Let me know if you need more analysis and I will be happy to dig a bit deeper.

dogmaphobic:
Yup! That makes sense. No one has ever run into this because video has always been present. I'll look into it.

Issue #7324, opened 2019-03-26
Camera capture status isn't handled correctly for photos


tpwrules:
When a capture starts, the camera manager sets its status to in progress, and resets it to idle when it receives a CAMERA_CAPTURE_STATUS message which says the camera is now idle. But when you take a photo or photo sequence, the manager never starts the timer to send a MAV_CMD_REQUEST_CAMERA_CAPTURE_STATUS.

Here, it resets the request count as if it is about to start the timer:
https://github.com/mavlink/qgroundcontrol/blob/6af7e8f2f40f390bb17f3b56e6c910010165787e/src/Camera/QGCCameraControl.cc#L364-L365

And here, it starts the timer to re-request the status if the start capture command failed:
https://github.com/mavlink/qgroundcontrol/blob/6af7e8f2f40f390bb17f3b56e6c910010165787e/src/Camera/QGCCameraControl.cc#L681-L688

But it never starts the timer when the command succeeds, like it does here for video:
https://github.com/mavlink/qgroundcontrol/blob/6af7e8f2f40f390bb17f3b56e6c910010165787e/src/Camera/QGCCameraControl.cc#L656-L663

I'm not sure what the correct solution is and I'm not sure what the obligation is for the camera to send CAMERA_CAPTURE_STATUS periodically. However, if my camera sends it right after it sends the CAMERA_IMAGE_CAPTURED message even if it never receives a request, everything works correctly.

dogmaphobic:
The camera is supposed to send a `CAMERA_IMAGE_CAPTURED` message for each image it captures. QGC will not request it.

Video is a bit different. When QGC sends a `MAV_CMD_VIDEO_START_CAPTURE` command, it sets param2 to 0, which means it is telling the camera not to send `CAMERA_CAPTURE_STATUS` automatically when recording video. QGC will request it on its own periodically. The reason for this is historic as this point as the camera that caused issues (it wasn't respecting param2) has since been fixed.

tpwrules:
When QGC takes a photo, it sets the `_photoStatus` to `PHOTO_CAPTURE_IN_PROGRESS`:
https://github.com/mavlink/qgroundcontrol/blob/6af7e8f2f40f390bb17f3b56e6c910010165787e/src/Camera/QGCCameraControl.cc#L364

My camera sends `CAMERA_IMAGE_CAPTURED` when it captures an image, and this causes the total image counter to increment and an image marker to show up on the map. 

[...]

PR #7325, opened 2019-03-27
Make sure video streaming is running before accessing it.

dogmaphobic:
Fixes https://github.com/mavlink/qgroundcontrol/issues/7323


dogmaphobic:
Also adding code to request the camera capture status when a photo is taken interactively. This should fix https://github.com/mavlink/qgroundcontrol/issues/7324

PR #7341, opened 2019-04-03
Disable Logging

dogmaphobic:
This adds the option to disable logging. Both locally (QGC) and Vehicle (to its microSD card). You control each individually.

Here you disable logging within QGC:

Screen Shot 2019-04-03 at 2 38 29 PM

And here within the vehicle:

Screen Shot 2019-04-03 at 2 38 20 PM


hamishwillee:
IMO the text for these isn't completely unambiguous/obvious. Up to you, but some thoughts:

1. There are now enough options for Logging to have its own section, in which case you could  pull out of miscellaneous and call the section QGC Vehicle Logging
1. I prefer to spell out what local means - so  maybe "Disable QGC-local logs".
1. "Disable vehicle logging" would probably better as "Disable on-vehicle logging" (vehicle logging could mean either on-vehicle or local).

dogmaphobic:
I knew you would go through this with a fine comb :) I'm about to leave for dinner. In the mean time, the only thing is that local has to be separate from vehicle. The vehicle setting is on a per vehicle basis and it is only available when a vehicle is connected. That's why I put it under the "Safety" panel. I didn't know where else to put it. Back in a bit...

hamishwillee:
:-)

Completely agree that "Safety" is the only reasonable place to put the per-vehicle log settings. 

Given there is only one setting, then the only other logical place right now might be a "Miscellaneous" sidebar section, with this as the first and currently only option. I'm happy _enough_ with "Safety" though for now.

Longer term though, you might consider having a (per-vehicle) Log sidebar. This would include this enable/disable setting and also vehicle-specific options to upload logs, and to push them to Flight review (on PX4 only) - as per the MAVLink 2 logs. Basically everything to do with a particular vehicle and logs in the one place.  
- We might also consider if adding a link to the QGC local logging with a little explanatory text. Not sure if that helps or hinders the "story"

DonLakeFlyer:
"Local logging" is what?

hamishwillee:
> "Local logging" is what?

That's what I said :-). Logging that is stored on QGC. My suggestion was "Disable QGC-local logs".
Other variations: "Disable on-QGC logging", "Disable QGC MAVLink Logs".

It is tempting to say "Disable QGC Logs" - but that implies the logging of QGC.

dogmaphobic:
I'm still not sure how to properly word them. The vehicle is simpler. It just doesn't write anything to its microSD card. On the QGC side, QGC is constantly writing the telemetry to disk, regardless if you ask for it or not. The reason being that if you decide to turn on after the fact, it already has telemetry stored (good in case of a crash or something). When you _disable local logging_, you are telling QGC not to write any telemetry to disk. Ever (well, until you re-enable it.)

hamishwillee:
Re QGC side, perhaps I'm missing something, but does the distinction matter to end users? For me the complicated story is QGC-local logging vs on-vehicle logging. That is why I keep on prepending QGC- when I mention local, because it isn't intuitive that both exist.

hamishwillee:
But if you wanted you could say "Disable storing telemetry logs to disk"

dogmaphobic:
It needs to be clear that no logs are written to disk on one side or the other (or both). That's the part that matters to the user. Ideally, we would have one button (checkbox) that says "no logs at all are written to disk--anywhere" but that is not practical.

>because it isn't intuitive that both exist.

hamishwillee:
I'm not sure you'd have such a box anyway, since you probably still want to configure per-vehicle logging (if you had a fleet)

Everywhere else you call the QGC side log a telemetry log. We could perhaps reverse the order and have a default enabled checkbox: "Save telemetry logs to disk". 
If you had a log section it could be called "Telemetry Logs from Vehicle" and then you could just have "Save logs to disk", "log from arm" etc.

dogmaphobic:
Ok,

I've flipped the logic so the option is now "Enable Logging" and it defaults to true.
Grouped the logging options within General Settings.

Screen Shot 2019-04-04 at 7 37 31 AM

Screen Shot 2019-04-04 at 7 37 21 AM

hamishwillee:
To my mind, much better. Possible tiny improvement, because you have context now from heading:
image

I'm now on holiday. Goodbye :-)

dogmaphobic:
> Possible tiny improvement, because you have context now from heading:

Done!

Enjoy your time off :)

DonLakeFlyer:
> When you _disable local logging_, you are telling QGC not to write any telemetry to disk. Ever (well, until you re-enable it.)

Ah, got it. Hmm, not sure how to word that. "Save logs to disk" isn't quite it either since that is only temp file if the others are turned off. But not sure how to get better than that. Maybe "Disable all telemetry logging"? I would keep the word telemetry in that one.

dogmaphobic:
Changed from just telemetry logs to _any_ data persistence, which includes caching map tiles.

Screen Shot 2019-04-05 at 11 59 17 AM

DonLakeFlyer:
Disable all data persistence? What would you use that for?

PR #6034, opened 2018-01-16
Correct video file timestamping issues

jaxxzer:
Fix #5436.

This needs some testing and review, but appears to do the trick. I have one issue in that dropping frames or blocking on a pad probe in the recording branch, also appears to drop/block in the live stream branch. Though, from everything I have read, they should be independent. I must be missing something. The result is that when recording starts, the live stream stops until the next keyframe arrives, but only the recording branch should be waiting for the keyframe. @NaterGator @dogmaphobic, any clue?

PR #7343, opened 2019-04-04
Fix: When recording starts the video stops until a key frame is received

MatejFranceskin:
When streaming video is displayed and the recording is started the video screen goes black for some amount of time, depending on when the next key frame is received.
This behaviour was implemented so that the recorded video would start with a key frame, but the side effect is to loose video for few seconds when you start recording which is unacceptable.
After this change the recorded video might start with intra frames which as in some players displayed as "garbled" image until the first key frame is received.

### DIFF
--- a/src/VideoStreaming/VideoReceiver.cc
+++ b/src/VideoStreaming/VideoReceiver.cc
@@ -130,14 +130,12 @@ VideoReceiver::grabImage(QString imageFile)
 static void
 newPadCB(GstElement* element, GstPad* pad, gpointer data)
 {
-    gchar* name;
-    name = gst_pad_get_name(pad);
-    //g_print("A new pad %s was created\n", name);
+    gchar* name = gst_object_get_name(reinterpret_cast<GstObject*>(pad));
     GstCaps* p_caps = gst_pad_get_pad_template_caps (pad);
     gchar* description = gst_caps_to_string(p_caps);
-    qCDebug(VideoReceiverLog) << p_caps << ", " << description;
+    qCDebug(VideoReceiverLog) << "A new pad was created - name: " << name << ", caps: " << description;
     g_free(description);
-    GstElement* sink = GST_ELEMENT(data);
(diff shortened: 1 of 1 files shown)


jaxxzer:
Original patch fyi: https://github.com/mavlink/qgroundcontrol/pull/6034

The frames are dropped on the `queue` inside the `_sink` box here: https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoStreaming/VideoReceiver.cc#L655

As far as I understand the dropping frames for the recording branch should have no effect on the `_videosink`, as it's happening downstream.

I have not observed the behavior mentioned.

Some players will refuse to open the file at all if we revert this.

MatejFranceskin:
I understand why it was implemented. I tried to play recorded file with VLC and ffplay without problems. From user perspective it is much worse that you get black video until key frame. Usually you start recording when there is something interesting and here you get blacked out. It happens on all platforms.
Is it possible that the behaviour changed with new version of gstreamer? What was GST version when you tested this?

jaxxzer:
> From user perspective it is much worse that you get black video until key frame. Usually you start recording when there is something interesting and here you get blacked out.

I'll point out that before or after this patch, you aren't going to get anything (but garbage) on the recording until the first keyframe. It's black video until keyframe or corrupted video until keyframe.

In doing some tests today, @Williangalvani has been able to reproduce this issue. It appears related to the keyframe interval which varies between cameras.

Williangalvani:
Yeah, I can reproduce it with a fake video stream only. Doesn't happen with our cameras.
 > From user perspective it is much worse that you get black video until key frame.

It is bad. But I think we should try to fix it properly instead of generating bad videos. I'll take a look into it too.

MatejFranceskin:
I have solved both problems in a new PR: https://github.com/mavlink/qgroundcontrol/pull/7361
@Williangalvani and @jaxxzer - could you please check how that works for your setups.

PR #7346, opened 2019-04-05
Hack around ArduPilot problems with unsupported NaN in Loiter commands

DonLakeFlyer:
Hack fix for #6783


DonLakeFlyer:
I changed the code such that the base level defaults for the LOITER commands default to 0 instead of NaN. And then only PX4 stack changes them to nanUnchanged and defaults to NaN. This is not theoretically correct to mavlink spec, and will cause mavlink generic vehicles to not be able to support headings in LOITER commands but it's the only workaround possible.

PR #7348, opened 2019-04-07
StructureScanFix: Starting altitude

FlavioTonelli:
Bug fix:
When scanning a structure from the top the starting altitude shall be equal to the structure height.


DonLakeFlyer:
Yup, thanks. Missed that change when the concept of Structure Height changed a while back.

DonLakeFlyer:
@FlavioTonelli Can you submit this same fix to the Stable_V3.5.1 branch as well?

PR #7351, opened 2019-04-14
MAV_CMD_NAV_WAYPOINT message shall express the desired yaw angle in d…

FlavioTonelli:
The desired yaw angle at a waypoint shall be expressed in degree.
Here is the code inside the PX4 firmware that handles the MAV_CMD_NAV_WAYPOINT message:

> switch (mavlink_mission_item->command) {
 case MAV_CMD_NAV_WAYPOINT:
 mission_item->nav_cmd = NAV_CMD_WAYPOINT;
 mission_item->time_inside = mavlink_mission_item->param1;
 mission_item->acceptance_radius = mavlink_mission_item->param2;
 mission_item->yaw = wrap_2pi(math::radians(mavlink_mission_item->param4));
 break;

> 
The yaw field is already converted on the firmware side from degree to radians.

### DIFF
--- a/src/MissionManager/MavCmdInfoCommon.json
+++ b/src/MissionManager/MavCmdInfoCommon.json
@@ -52,7 +52,7 @@
             },
             "param4": {
                 "label":            "Heading",
-                "units":            "radians",
+                "units":            "deg",
                 "nanUnchanged":     true,
                 "default":          null,
                 "decimalPlaces":    2


DonLakeFlyer:
This needs to be discussed with the MAVLink working group with respect to the spec and what both PX4 and ArduPilot support. There are many other commands which currently have heading specified in radians (just search for radians in MissionManager/MavCmdInfoCommand.json). In fact all headings in commands are currently espressed in radians as far as QGC is concerned. Changing just one of them doesn't make sense.

dagar:
PX4 is currently interpreting this field as degrees. It's kind of crazy that it's missing units in the spec.

FlavioTonelli:
I get your point of view.
I suggest anyway to merge the modification just for helping basic-user in every day flight.
In the actual stable version it is tricky to specify a desired yaw for the drone. The interface ask for a desired degree heading value but, to let it work properly, the user has to insert the value in degree "converted to degree" (value*180)/pi
In order to fly 116 degree you have then to insert 6649.

DonLakeFlyer:
> I suggest anyway to merge the modification just for helping basic-user in every day flight.

I can't do that beause this change applies to all firmwares. It will affect ArduPilot as well and I'm not sure if that is radians or degrees. 

To change for PX4 only change the same file in the PX4 firmware only: https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/PX4/MavCmdInfoCommon.json. And change it for all the heading values in other commands as well.

I'll talk with ArduPilot to see how they work.

DonLakeFlyer:
> To change for PX4 only change the same file in the PX4 firmware only:

I'd hold off till I hear back from ArduPilot.

DonLakeFlyer:
The full list is: 
* MAV_CMD_NAV_WAYPOINT
* MAV_CMD_NAV_LOITER_UNLIM
* MAV_CMD_NAV_LOITER_UNLIM
* MAV_CMD_NAV_LOITER_TIME
* MAV_CMD_NAV_LAND
* MAV_CMD_NAV_TAKEOFF

@dagar Is it ok to change all those to degrees?

DonLakeFlyer:
ArduPilot doesn't implement these yet, so degress should theoretically be ok for everyone.

auturgy:
@wickedshell

dagar:
> The full list is:
> 
> * MAV_CMD_NAV_WAYPOINT
> * MAV_CMD_NAV_LOITER_UNLIM
> * MAV_CMD_NAV_LOITER_UNLIM
> * MAV_CMD_NAV_LOITER_TIME
> * MAV_CMD_NAV_LAND
> * MAV_CMD_NAV_TAKEOFF
> 
> @dagar Is it ok to change all those to degrees?

I confirmed these are all used as angles in degrees. The range and wrapping might be another thing worth clarifying in the spec.

DonLakeFlyer:
@FlavioTonelli I'll clean this up. I'm going to leave this pull open until I put in a replacement which has changes for all the commands. Then I'll close this.

PR #7356, opened 2019-04-16
Better handling of NaN in LOITER_* commands

DonLakeFlyer:
Previous change to handle #6783 was not completely correct. This fixes that:
* PX4: Fixes handling back to spec
* ArduPilot
  * Maintains multi-rotor heading defaulting to 0 instead of NaN
  * Fixed wing now shows correct x-track options
  * Fixed Wing Landing Pattern item does not show camera options and defaults them to off.

### DIFF
--- a/src/FirmwarePlugin/APM/MavCmdInfoCommon.json
+++ b/src/FirmwarePlugin/APM/MavCmdInfoCommon.json
@@ -10,6 +10,36 @@
             "paramRemove":  "2"
         },
         {
+            "id":                   17,
+            "comment":              "MAV_CMD_NAV_LOITER_UNLIM",
+            "param4": {
+                "label":            "Heading",
+                "units":            "radians",
+                "default":          0,
+                "decimalPlaces":    2
+            }
+        },
+        {
+            "id":                   18,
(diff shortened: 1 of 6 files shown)


hamishwillee:
@DonLakeFlyer Are you saying that ArduPilot uses 0 instead of NaN for loiter commands - e.g. [MAV_CMD_NAV_LOITER_UNLIM](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_UNLIM)?
- Would I be correct in saying that use of NaN is a recent change/refinement of the requirement? 
- Are they aware of the change/working on it?

Ie basically I want everyone to follow the spec, but would like to confirm the problem before chasing them.

DonLakeFlyer:
> Are you saying that ArduPilot uses 0 instead of NaN for loiter commands

Correct.

> Would I be correct in saying that use of NaN is a recent change/refinement of the requirement?

I think it more of an omission in the spec as opposed to a change. Without the NaN the definition of the command doesn't make any sense and isn't really usable.

> * Are they aware of the change/working on it?

I tried having a discussion with them and gave up.

PR #7361, opened 2019-04-17
Configure video stream

MatejFranceskin:
Add parameters that allow setting video frame rate, bit rate and encoding from QGC.
To properly detect video stream encoding the gstreamer pipeline has to use decodebin plugin.
Besides standard h264 streams it was tested with HEVC (h265) encoded video streams.


Williangalvani:
This generated a bad video for me:
Screenshot from 2019-04-17 14-13-07

The ball should be filled, and the time interval of the video file is also bad. It is currently  playing 25 seconds of 9 seconds.

I looked at the issue of the video going out when clicking record and came up with my own solution here: https://github.com/mavlink/qgroundcontrol/pull/7362

Williangalvani:
I just tested here. I'm still losing video after clicking record. Also I could not play the recorded file, it opens but it it is like it has only one frame and is finished instantly.
I tested with this pipeline: `gst-launch-1.0 videotestsrc pattern=ball ! video/x-raw,width=640,height=480,framerate=30/1 !  videoconvert ! x264enc ! video/x-h264, profile=baseline ! rtph264pay ! udpsink host=127.0.0.1 port=5600`

dogmaphobic:
@MatejFranceskin Is this safe/ready to merge?

MatejFranceskin:
Not yet - I have to fix case Willian Galvani found where there is a problem.

PR #7362, opened 2019-04-17
Delay adding filesink to recording branch until after an I-frame arrives

Williangalvani:
This solves the issue where the video is lost for a few seconds after clicking record as the pipelines waits for the filesink to preroll but it does not receive a valid buffer because the probe is dropping anything that is not an I-frame, stalling the whole pipeline. 
This issue is mostly noticeable in videos with larger intervals between I-frames.

I tested with this pipeline:
```gst-launch-1.0 videotestsrc pattern=ball ! video/x-raw,width=1920,height=1080,framerate=30/1 !  videoconvert ! x264enc bitrate=5000 ! video/x-h264, profile=baseline ! rtph264pay ! udpsink host=127.0.0.1 port=5600```

Before:
before pr
After:
after pr

### DIFF
--- a/src/VideoStreaming/VideoReceiver.cc
+++ b/src/VideoStreaming/VideoReceiver.cc
@@ -712,17 +712,18 @@ VideoReceiver::startRecording(const QString &videoFile)
     gst_object_ref(_sink->mux);
     gst_object_ref(_sink->filesink);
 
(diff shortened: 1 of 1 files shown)


MatejFranceskin:
Well done! I tested it and it works perfectly.

LorenzMeier:
Thanks!

Issue #7355, opened 2019-04-16
Structure Scan Total Distance, Time are zero


Stifael:
In the mission plan-view, `Total Mission Distance` and `Time` are not updated and remain 0. 

![qgc_distance_0](https://user-images.githubusercontent.com/13449161/56195017-632d7480-6034-11e9-8cf9-dfbc36047613.png)

PR #7364, opened 2019-04-18
the battery and distance of structure scan always show 0

lichenyuneec:
Add distance calculation of structure scan.

### DIFF
--- a/src/MissionManager/StructureScanComplexItem.cc
+++ b/src/MissionManager/StructureScanComplexItem.cc
@@ -84,6 +84,7 @@ StructureScanComplexItem::StructureScanComplexItem(Vehicle* vehicle, bool flyVie
     connect(&_cameraCalc, &CameraCalc::isManualCameraChanged, this, &StructureScanComplexItem::_updateGimbalPitch);
 
     connect(&_layersFact,                           &Fact::valueChanged,            this, &StructureScanComplexItem::_recalcScanDistance);
+    connect(&_flightPolygon,                        &QGCMapPolygon::pathChanged,    this, &StructureScanComplexItem::_recalcScanDistance);
 
     _recalcLayerInfo();
 
@@ -514,8 +515,6 @@ void StructureScanComplexItem::_rebuildFlightPolygon(void)
         _entryVertex = savedEntryVertex;
     }
 
-    _recalcScanDistance();
-
     emit coordinateChanged(coordinate());
     emit exitCoordinateChanged(exitCoordinate());
 }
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
I try to get some time to take a look at this tonight. Thanks for the fix.

Fix for #7355

src/MissionManager/StructureScanComplexItem.cc
```diff
@@ -83,6 +83,8 @@ StructureScanComplexItem::StructureScanComplexItem(Vehicle* vehicle, bool flyVie
 
     connect(&_cameraCalc, &CameraCalc::isManualCameraChanged, this, &StructureScanComplexItem::_updateGimbalPitch);
 
+    connect(&_layersFact,                           &Fact::valueChanged,            this, &StructureScanComplexItem::_recalcScanDistance);
```
DonLakeFlyer: The signalling should also be connected to:
```
    connect(&_flightPolygon,  &QGCMapPolygon::pathChanged,    this, &StructureScanComplexItem::_recalcCameraShots);
```

src/MissionManager/StructureScanComplexItem.cc
```diff
@@ -519,6 +513,9 @@ void StructureScanComplexItem::_rebuildFlightPolygon(void)
     } else {
         _entryVertex = savedEntryVertex;
     }
+
+    _recalcScanDistance();
```
DonLakeFlyer: This will calc recalc the scan distance at times when it is not needed since the number of camera shots can change without the scan distance being affected. The signal connected to the flight path polygon is the better way to do it. Since the two cases which affect scan distance are polygon change or layer count change.

src/MissionManager/StructureScanComplexItem.cc
```diff
@@ -598,3 +595,26 @@ void StructureScanComplexItem::_signalTopBottomAltChanged(void)
...
+
+    double surfaceHeight = qMax(_structureHeightFact.rawValue().toDouble() - _scanBottomAltFact.rawValue().toDouble(), 0.0);
+    scanDistance += surfaceHeight;
+
+    if (!qFuzzyCompare(_scanDistance, scanDistance)) {
+        _scanDistance = scanDistance;
+        emit complexDistanceChanged();
+    }
+
+    qCDebug(MissionControllerLog) << "StructureScanComplexItem--_recalcScanDistance layers: "
```
DonLakeFlyer: This should be ```StructureScanComplexItemLog```

lichenyuneec:
@DonLakeFlyer I've changed it as you suggested. Thank you for your advice.

DonLakeFlyer:
Great thanks for the fix

PR #7374, opened 2019-04-23
Fixed FlightMap stealing mouse from ToolStrip and DropPanel in PlanView

stefandunca:
Map control gestures work in PlanView over the ToolStrip and DropPanel and they break other MouseArea functionality for controls overlaid on top of map. Moving the ToolStrip outside FlightMap control fixes this issue.

### DIFF
--- a/src/PlanView/PlanView.qml
+++ b/src/PlanView/PlanView.qml
@@ -574,87 +574,90 @@ QGCView {
                     border.width:   object.lineWidth
                 }
             }
+        }
 
-            ToolStrip {
-                id:                 toolStrip
-                anchors.leftMargin: ScreenTools.defaultFontPixelWidth
-                anchors.left:       parent.left
-                anchors.topMargin:  _toolButtonTopMargin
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
The way to fix this is to use DeadMouseArea. If you look at ToolStrip it always has one behind everything. Are you sure that the Toolstrip itself lets the mouse events through. The DropPanel needs the same change made to it with a DeadMouseArea behind everything.

I'm concerned that changing that parent of ToolStrip will cause problems.

stefandunca:
I don't have an explanation but I've tried all the possible combinations of `DeadMouseArea` on top of buttons inside the `ToolStrip`, different `z` order of controls and `MouseArea` settings ... . The only working solution was this one. I will cancel this change, however, it's included in `qgc4` branch. I also noticed the same effect in other views on top of the `Map`. I will investigate this another time but it seems that `MapGestureArea` inside `Map` control might have a higher priority of capturing the mouse from other `MouseArea` overplayed on top.

Issue #4210, opened 2016-11-14
Make android virtual joystick have option for snap back to center


jgoppert:
https://github.com/mavlink/qgroundcontrol/blob/cbee0fb9aab043924b4c5df7bbc3d62685522ff1/src/FlightDisplay/VirtualJoystick.qml#L41

Similar to the gamepad, it would be nice to make the touch joystick interface also have a mode where it snaps back to zero for altitude control mode.

Or, intelligently switch to snap back to zero mode when in altittude control mode and make the interface disappear in auto mode.

DonLakeFlyer:
The changes for auto-center throttle don't really fix this request.

HTRamsey:
Closing in favor of #11590

Issue #5897, opened 2017-12-18
Make virtual joystick usable


DonLakeFlyer:
After flying for a week in Zimbabwe with virtual joystick I would deem it a piece of crap. Needs work:
* ~~Throttle needs to auto-center when flying multi-rotor~~
* ~~If throttle auto-centers needs to deal with fact that you need full throttle down for arming~~
* Your thumbs cover up where the stick really is. Need additional indicators to the side of the sticks to show position.
* On tablet the joystick pads cover up various parts of the ui. Need some way to make this work better.

DonLakeFlyer:
> * Throttle needs to auto-center when flying multi-rotor
> * If throttle auto-centers needs to deal with fact that you need full throttle down for arming

These two lead to a safety problem where it's too easy to leave the throttle in a bad position on multi-rotor.

DonLakeFlyer:
> * Throttle needs to auto-center when flying multi-rotor

Done

> * If throttle auto-centers needs to deal with fact that you need full throttle down for arming

Not going to fight with this. Use the ui to arm no the stick in this case.

PR #7375, opened 2019-04-25
Improve virtual joystick functionality 

patrickelectric:
- Helps with: #5897
- Fix: #4210

**Describe the bug**
Helps and fix listed issues.
The patch allows the user to configure the virtual joystick throttle to centralize like a normal gamepad joystick.

**Expected behavior**
When selecting the correct settings, the virtual joystick throttle will centralize after user input.

**Log Files and Screenshots**
image

deepin-screen-recorder_Select area_20190425003818

This helps vehicles that allow reverse joystick inputs like:
- Rover
- Sub
- Boats


patrickelectric:
Maybe @peterbarker @jaxxzer inputs could validate that.

DonLakeFlyer:
@hamishwillee Any ideas on better wording for "Centralize Joystick"?

patrickelectric:
I was thinking between "centralize" and "spring-loaded", but the second one is not so clear for newer users.

DonLakeFlyer:
Maybe Auto-Center or something along those lines.

hamishwillee:
@DonLakeFlyer Auto-Centre or Auto-ReCentre are the best I can come up with too.

patrickelectric:
@DonLakeFlyer @hamishwillee fixed !
image

hamishwillee:
Just generally, good to be consistent re capitalisation across the app - ie here we have _Joystick_ capitalised, but every other second word is not.

DonLakeFlyer:
As I run through things I've been moving to capatilize everything.

patrickelectric:
Well, the text was already there, should I fix it or do you want to create a second PR to fix all strings in the settings menu ?

DonLakeFlyer:
So I'm remembering why it didn't start out this way. With throttle starting centered you can't arm the vehicle.

DonLakeFlyer:
@patrickelectric I would worry about stuff you didn't change.

patrickelectric:
It was tested with the ROV and it's working.
Some vehicles indeed does not arm with centralized joysticks, but others do :)
Maybe we should configure the joystick automatically for each vehicle in the future.

DonLakeFlyer:
The majority of vehicles out there don't arm with center throttle (Multi-Rotor, Plane). Which means lots of "Hey, how come I can't arm?" support issues.

patrickelectric:
Hmm, maybe should we enable this configuration only with rover and submarines ?

DonLakeFlyer:
Make sense.

patrickelectric:
@DonLakeFlyer auto-center configuration will only be available for sub and rover now

taileron:
my proposal would be to add curves e.g. to the virtual thottle stick with the possibility of a huge dead band in the middle and enough expo function. Tried this with a historic app called FlyPad with enough expo and throttle dead band it was easy possible to fly a copter if necessary also in acro mode.

DonLakeFlyer:
@patrickelectric So I've been looking at this change in relationship to #8774. It breaks the auto-center of the right stick. That's fine I fixed that. But I don't understand how the throttle center setting works. It seems like it is meant to support reverse on the virtual joystick. But as far as I can tell it doesn't. The reason being the range of values being sent is still from [0, 1] [stick down, stick up]. Whereas a reverse value needs to be negative. I've verified this using a daily build and PX4 SITL Rover. You can't make it go in reverse.

DonLakeFlyer:
Another question would be why the setting is needed at all instead of making it autocenter all the time if the current vehicle is a rover/sub?

patrickelectric:
Hi @DonLakeFlyer ArduSub does the map from [1100, 1900] to [-1, +1] for the throttle and map back to the ESC to be [1100 (full reverse), 1900 (full forward)].
And about to center the joystick by default for sub, I totally agree.

DonLakeFlyer:
So that is wrong according to mavlink spec. Reverse is negative values for MANUAL_CONTROL. And my recent changes will break the ability of reverse in ArduSub since QGC now does reverse according to spec. I've verified in SITL PX4 and ArduPilot Rovers that this works correctly for them.

DonLakeFlyer:
The way it works is if the vehicle is either a rover or a sub the virtual joystick send full range positive/negative values.

Williangalvani:
Yep, it looks like ArduSub is out of spec there. I'm thinking of patching ArduSub to take the correct range, but this will likely make older versions unusable with QGC master.
So we'll also need to set the Sub minimum supported version to 4.1 (next release)

What do you guys think? (also @jaxxzer)

DonLakeFlyer:
FYI: The virtual joystick change is in Stable as well since it was a safety issue.

Williangalvani:
Reverse is negative values for MANUAL_CONTROL. And my recent changes will break 
     the ability of reverse in ArduSub since QGC now does reverse according to spec. 

@DonLakeFlyer are these changes already in master? Do they also affect USB Joysticks?

DonLakeFlyer:
> are these changes already in master?

Yes and stable.

> Do they also affect USB Joysticks?

Virtual stick only

PR #7376, opened 2019-04-26
Pattern Preset support

DonLakeFlyer:
This allows you to save settings for a Pattern item as a named reusable preset. So for example if you fly a lot of say NDVI surveys at the same altitude and overlap settings you can save that as a preset. Then you create a new survey for a new area, you just select that Preset and you are done setting up the Survey.

Althought the internal archiecture for this is set up to work with all pattern types it is only supported by Survey now. Once we work through the usability of this on Survey then I'll implement in other Patterns.

Screen Shot 2019-04-26 at 10 31 18 AM

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,7 @@ Note: This file only contains high level features or important fixes.
 
 ### 3.6.0 - Daily Build
 
+* Plan/Pattern: Support named presets to simplify commonly used settings setup. Currently only supported by Survey.
 * ArduCopter: Handle 3.7 parameter name change from CH#_OPT to RC#_OPTION.
 * Improved support for flashing/connecting to ChibiOS bootloaders boards.
 * Making the camera API available to all firmwares, not just PX4.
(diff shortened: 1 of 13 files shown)


DonLakeFlyer:
The docs for this don't seem to be building correctly right now. But they are here: https://docs.qgroundcontrol.com/en/PlanView/PatternPresets.md

DonLakeFlyer:
Also this was a pretty major change to support this. So be careful when using a preset to make sure it really changed the settings to what they should be by switching back to Custom to check the actual settings.

DonLakeFlyer:
For now you can read the .md file directly: https://docs.qgroundcontrol.com/en/PlanView/PatternPresets.md which is somewhat helpful.

DonLakeFlyer:
This greatly simplifies the UI if you are using a preset:
Screen Shot 2019-04-26 at 11 20 58 AM

Further support will be added such that custom build can define their own presets as built in from the start.

One question is whether there is some useful presets which could be included with regular QGC as well.

Antiheavy:
After playing around with it, the functionality is there and seems to work quite well.  I'm trying to decide whether the camera is a subset of the preset or if the preset is a subset of the camera...  arguments both ways....

DonLakeFlyer:
Yeah, the camera thing is a little wierd. I see the need, but I'm not sure it's quite correct as it currently stands.

Antiheavy:
Also, this ability for the user to save custom profiles for things could be useful for Preflight Checklist.

PR #7383, opened 2019-04-27
Add support for geofence return point. 

DonLakeFlyer:
* You can drag the "B" indicator around to change location.
* Also fixed a bunch of bugs in geofence/rally support.

Screen Shot 2019-04-27 at 12 30 45 PM

Screen Shot 2019-04-27 at 12 31 00 PM


DonLakeFlyer:
@peterbarker Fence return point support.

PR #7389, opened 2019-04-28
Pr tweaks toolstrip qmltest

stefandunca:
Improve QmlTest and ToolStrip

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -26,6 +26,11 @@
 #include <QStringListModel>
 #include <QRegularExpression>
 #include <QFontDatabase>
+#ifdef Q_OS_LINUX
+#ifndef __mobile__
+    #include <QMessageBox>
+#endif
+#endif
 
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
Can you include before/after screenshots for things like this so we can see what the ui change looks like?

dogmaphobic:
> Can you include before/after screenshots for things like this so we can see what the ui change looks like?

It doesn't invalidate the request, but note that this PR is meant for the new QtQuick branch, not master. Once this branch is ready, there will be a great deal of explaining, screen shots, etc.

stefandunca:
#### ToolStrip:
Before: before_7389
After: after_7389

#### QmlTest:
Before: before_palette
After: after_palette

dogmaphobic:
CI errors are to be expected. Not only Jenkins is not setup properly, unit test is disabled for the time being. Feel free to merge (to qgc4)

DonLakeFlyer:
Ah, I didn't catch this was in qgc4. Sorry

PR #7407, opened 2019-05-02
Pid Tuning: Stream rate adjustment, Automatic flight mode switching

DonLakeFlyer:
* QGC will increase stream rates on ATTITUDE and ATTITUDE_TARGET messages for usable graphs. If you are running with low bandwidth telemetry you may need to adjust the vehicle settings further to make bandwidth room for these faster rates.
* Add "Automatic Flight Mode Switching". If you try to tune a vehicle in Stabilize mode while trying to capture the charts with Start/Stop on the ui it can become quite a handful for one person. The concept here is you turn this mode on. Then for example in PX4 you start in Positionor or Hold mode so the vehicle is stable. When you click Start, the vehicle switches to Stabilized, you then twitch to a side to generate data in Stablized mode and when you click Stop the vehicle goes to Hold flight mode. That way it's isn't running off on you while you are futzing with the PID Tuning ui.

Screen Shot 2019-05-02 at 4 46 48 PM

### DIFF
--- a/src/AutoPilotPlugins/PX4/PX4TuningComponentCopter.qml
+++ b/src/AutoPilotPlugins/PX4/PX4TuningComponentCopter.qml
@@ -61,26 +61,6 @@ SetupPage {
                         max:            15
                         step:           1
                     }
-                    /*
-  These seem to have disappeared from PX4 firmware!
-                    ListElement {
-                        title:          qsTr("Roll sensitivity")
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
I was able to get reasonable graphs on PX4 by switching the mav mode to Minimal and mav rate to 2kb/sec over the radio. Still a WIP as to whether this will work well for tuning over something like a SiK radio. I'll do some flight testing tomorrow with it.

DonLakeFlyer:
I also changed the sliders on the basic tuning pages back to regulars sliders.

PR #7424, opened 2019-05-07
Update throttle % fact from VFR_HUD message

DonLakeFlyer:
Fix for #7421


Antiheavy:
tested using a replay telemetry file.  works!
image

PR #7429, opened 2019-05-09
Warn user when multiple devices are connected while doing the firmware upgrade

patrickelectric:
**Describe the bug**
The user does not have any feedback around how many devices are connected and what kind of device should be removed.
QGC also detects normal FTDIs and Arduinos as SiK radios.

**Expected behavior**
This PR add a human friendly message warning the user that multiple devices are connected and need to be removed before performing the firmware upgrade.

image


DonLakeFlyer:
I like the less techy "Detected devices:" text much better.

DonLakeFlyer:
I wait, the text above sort of explains it. That's ok. But there looks to be a typo in "detected !" with a space between "detected" and the "!". Is that right or just a quirk of the display?

patrickelectric:
Yes, I just removed the extra space before the '!'.

DonLakeFlyer:
Thanks @patrickelectric

PR #7432, opened 2019-05-11
Firmware Upgrade: Use ArduPilot manifest to get firmware listings

DonLakeFlyer:
The manifest of available firmwares is downloaded from the ArduPilot web site. This then gives you the list of available firmwares which can be flashed to the board.

The UI is reworked such that you can pick ChibiOS/NuttX and then vehicle type followed by the actual firmware.

If the board leads to a single firmware (like when the ArduPilot bootloader indicates specific board type) you get that specific firmware to flash:
Screen Shot 2019-05-11 at 1 45 17 PM

If it is a generic board like an old Pixhawk then you have to pick the firmware you want:
Screen Shot 2019-05-11 at 1 45 42 PM

I would consider this a bit of a WIP since I'm not still too happy with all the choices the user has to make in some cases (for example the old Pixhawk 2 case aboce). But I need to put some more though into that first to figure out what to do about it.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -599,6 +599,7 @@ HEADERS += \
     src/Settings/AppSettings.h \
     src/Settings/AutoConnectSettings.h \
     src/Settings/BrandImageSettings.h \
+    src/Settings/FirmwareUpgradeSettings.h \
     src/Settings/FlightMapSettings.h \
     src/Settings/FlyViewSettings.h \
     src/Settings/OfflineMapsSettings.h \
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
@tridge FYI

bys1123:
Love this! Maybe this feature only for advanced config is better.

DonLakeFlyer:
> Maybe this feature only for advanced config will be better?

Given the way ArduPilot firmwares work (variants for each board type) that is not possible. If using the old bootloader the user needs to choose the right one or possibly get some sort of loss of functionality. You'll have to ask the ArduPilot folks exactly what you lose if you pick the wrong one.

DonLakeFlyer:
Once question would be whether ChibiOS decision should move to Advanced only and just default to ChibiOS. Up to the ArduPilot folks to decide.

PR #7435, opened 2019-05-12
Restart video stream after change

MatejFranceskin:
- Add camera parameters for video stream bitrate, frames per second and encoding.
- When these parameters where changed from QGC, the video stream was not restarted properly and sometimes it was started twice
- If H265 encoding is used change pipeline so that H265 decode is used.


dogmaphobic:
@MatejFranceskin, if you rebase this against master I can now merge it.

MatejFranceskin:
Rebased.

PR #7443, opened 2019-05-15
Move to QtQuick

dogmaphobic:
This is primarily a move from QtWigets to QtQuick. Until now, QGC was still running its legacy QtWidget core with a QtQuick container for the UI. This moves everything away from QtWidgets altogether and QGC is now fully QtQuick. From a user perspective, nothing is radically different. This is mostly an internal architectural change, which will allow for better UI experience as we make use of newer and updated controls. It also standardize the code base for all platforms. QGC had many features only available on desktop builds. The code will now be mostly the same across all platforms.

This move will go in steps, where this is the first and largest one. Several legacy widgets were QtWidgets only and those will have to be “ported” over to QtQuick. Principally, the _Analyze_ widget (the real time charting) has been temporarily dropped. That code will need to be ported over to QtQuick. Like the Analyze widget, the same will need to happen for _Custom Command_, _Onboard Files_ and _HIL Config_ (is that even used?)

Old QtWidgets
Screen Shot 2019-05-15 at 12 27 22 PM

Most new controls are now based on QtQuick 2.11 and QtQuick.Controls 2.4. The bulk of QGC is still using 2.3 and 1.2 respectively and will be updated over time.

Besides the move to QtQuick, a few other things:

- QGC now requires Qt 5.11 or greater. The idea is to standardize on Qt 5.12 (LTS). Just waiting for a solution for Windows as Qt dropped support for 32-bit.
- New, QtQuick MAVLink Inspector. The basics are already there but it still needs the ability to filter compID.
- Fixed application storage location on iOS. It was trying to save things where it could not.
- Basic support for secondary, thermal imaging with video streaming. If a camera provides both visual spectrum and thermal imaging, you have the option of displaying both at the same time.
- Better handling of fonts for Korean and Chinese locales. QGC now has builtin fonts for Korean (where some unusable font was being used). I still need to know if Chinese will need its own font as well.

New MAVLink Inspector
Screen Shot 2019-05-15 at 12 35 47 PM

As this is, at least internally, a rather large change, chances are things may have gotten broken here and there. This will need to go through quite a bit of testing to identify new issues so we can fix them as they are found.


bys1123:
Where is the analyze widget?

dogmaphobic:
>Where is the analyze widget?

_Several legacy widgets were QtWidgets only and those will have to be “ported” over to QtQuick. Principally, the Analyze widget (the real time charting) has been **temporarily dropped**._

Issue #7444, opened 2019-05-15
Android: Large number of crashes from QGCActivity.openAccesory


DonLakeFlyer:
This code is crashing for lots of people on Stable release. Hundreds plus crashes reported on Google Play:

Today, 11:38 AM on app version 352000
HTC One M9 (htc_himaulatt), Android 7.0
Report 1

java.lang.RuntimeException: 
  at android.app.LoadedApk$ReceiverDispatcher$Args.run (LoadedApk.java:1179)
  at android.os.Handler.handleCallback (Handler.java:751)
  at android.os.Handler.dispatchMessage (Handler.java:95)
  at android.os.Looper.loop (Looper.java:173)
  at android.app.ActivityThread.main (ActivityThread.java:6523)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:938)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:828)

Caused by: java.lang.NullPointerException: 
  at org.mavlink.qgroundcontrol.QGCActivity.openAccessory (QGCActivity.java:674)
  at org.mavlink.qgroundcontrol.QGCActivity.access$200 (QGCActivity.java:65)
  at org.mavlink.qgroundcontrol.QGCActivity$2.onReceive (QGCActivity.java:110)
  at android.app.LoadedApk$ReceiverDispatcher$Args.run (LoadedApk.java:1159)

DonLakeFlyer:
@MatejFranceskin This one if crashing all over the place as well.

DonLakeFlyer:
Not yet in stable branch

DonLakeFlyer:
@MatejFranceskin Can you move this to Stable_V3.5.1?

PR #7448, opened 2019-05-16
Fix for java open accessory crash

MatejFranceskin:
Bug reported in #7444

Caused by: java.lang.NullPointerException:
at org.mavlink.qgroundcontrol.QGCActivity.openAccessory (QGCActivity.java:674)

I could not reproduce it in my environment but few commits agot line 674 was 
Log.i(TAG, "openAccessory: " + usbAccessory.getSerial());
which means that usbAccessory was null when openAccessory was called.

This PR makes sure that usbAccessory is checked for null before being used.


DonLakeFlyer:
@MatejFranceskin This fix also needs to be in Stable_V3.5.1 branch.

Issue #7445, opened 2019-05-15
Android: Large number of crashes from QGCActivity.nativeUpdateAvailableJoysticks


DonLakeFlyer:
This is being reporting as 100s of crashes on the Google Play store stable build:

Today, 9:58 AM on app version 352000
Samsung Galaxy Note5 (noblelte), Android 7.0
Report 1

java.lang.UnsatisfiedLinkError: 
  at org.mavlink.qgroundcontrol.QGCActivity.nativeUpdateAvailableJoysticks (Native Method)
  at org.mavlink.qgroundcontrol.QGCActivity.access$800 (QGCActivity.java:65)
  at org.mavlink.qgroundcontrol.QGCActivity$3.onReceive (QGCActivity.java:169)
  at android.app.LoadedApk$ReceiverDispatcher$Args.run (LoadedApk.java:1185)
  at android.os.Handler.handleCallback (Handler.java:751)
  at android.os.Handler.dispatchMessage (Handler.java:95)
  at android.os.Looper.loop (Looper.java:154)
  at android.app.ActivityThread.main (ActivityThread.java:6682)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1520)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1410)

DonLakeFlyer:
@MatejFranceskin It's critical that you look at this one. It's blowing up all over the place on people using Stable build.

MatejFranceskin:
I am looking at it but the problem is that I cannot reproduce it on my devices. It is ok with or without joystick.
Do you have any more information? It could be dependent on Android version.

MatejFranceskin:
When did this problems start? Immediately after it was included in the build or later?
Did we change target SDK version recently?

DonLakeFlyer:
Looks like all over the place:
![Screen Shot 2019-05-16 at 9 41 11 AM](https://user-images.githubusercontent.com/5876851/57871925-621c6e00-77bf-11e9-8b55-6ebdfa720029.png)

Do you have access to Google Play Dev Store? Probably should. Daniel should be able to get it for you. Then you can see crash reports directly.

sjd69:
I've replicated this issue using a Note 8 (Qualcomm) as well as a S9+ (Qualcomm).

[...]

PR #7449, opened 2019-05-16
Set Android Joystick native methods from JNI_OnLoad

MatejFranceskin:
Bug report: #7445

The problem could not be reproduced on my devices but I found that the joystick native methods were not registered at the same point of time as the serial port native methods so it is very likely that they were not registered yet when this crash happens.
Solution registers joystick native methods from JNI_OnLoad - in the same way as serial port JNI methods.


DonLakeFlyer:
Should be able to track this on Daily Build reporting to see crashes go away

DonLakeFlyer:
@MatejFranceskin This fix needs to brought to Stable_V3.5.1 branch so stable build gets it.

PR #7452, opened 2019-05-16
Localization Work

dogmaphobic:
- Automated the ingestion of localization from Crowdin
- Automated the generation of language resources into the application
- Added all languages that come from Crowdin, even if empty.
- Allow dynamic language changes


bys1123:
Tried this version, I can't switch to Chinese version.

PR #7455, opened 2019-05-17
main: Returning critical messsage and an error code when QGC is alredy running

patrickelectric:

---
name: QGC fails silently when it fails to initialize 
about: There is no user feedback when it fails to start

---

**Describe the bug**
This patch provides two feedback if QGC fail to initialize, one as a system error number with the return and a critical message with a human string for the user.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -19,6 +19,7 @@
 #include <QApplication>
 #include <QIcon>
 #include <QSslSocket>
+#include <QMessageBox>
 #include <QProcessEnvironment>
 #include <QHostAddress>
 #include <QUdpSocket>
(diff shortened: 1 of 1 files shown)


dogmaphobic:
That message is only shown on the console. It will not normally be seen on any OS unless you launch QGC from the command line (nowhere for mobile). If you are happy with that, I'm fine with the PR.

patrickelectric:
@dogmaphobic I changed the PR to create a popup message.
image

dogmaphobic:
I think that makes more sense if you truly want the user to know why "nothing" happened. I don't think we need to worry about mobile as you can't launch a second instance of anythings so I would exclude from mobile builds. With that said, I see two issues:

1) The string should probably be made translatable:
```
QMessageBox::critical(nullptr, QStringLiteral("Error"),
	QString(tr("Another instance of QGroundControl is already running. Please close the other instance first and try again."))
);
```
2) Does this work on a non QtWidget environment? Just making sure as I assume you run it fine already.

dogmaphobic:
```
        // QApplication is necessary to use QMessageBox
	        QApplication errorApp(argc, argv);
```
Ha! Disregard issue 2 above :)

patrickelectric:
@dogmaphobic thank you! I did lost the tr on my previous version. 
`QStringLiteral`  or `QString` is not necessary to use with QObject::tr, since it results a QString already.

dogmaphobic:
> I did lost the tr on my previous version.
> `QStringLiteral` or `QString` is not necessary to use with `QObject::tr`, since it results a `QString` already.

Yup, I'm not at home and I'm using this tiny laptop so disregard the lack of attention.

patrickelectric:
@dogmaphobic 	any update ?

dogmaphobic:
Sorry, it fell through the cracks...

PR #7456, opened 2019-05-17
FirmwareUpgrade: Fix firmware type layout with long strings

patrickelectric:

**Describe the bug**
Actual layout breaks if string of both types are bigger than the actual parent width.

Old:

New:


patrickelectric:
Oh, fixed in master.

PR #7458, opened 2019-05-18
cmake fix build and delete obsolete files

dagar:


dogmaphobic:
I'm offline until Monday and then it's crazy week in Zürich. I'll see what I can do...

dagar:
Building now, but somethings not quite right.

Screenshot from 2019-05-18 15-08-12

dagar:
Cmake was missing `qgcimages.qrc`. Seems to be working now.

Screenshot from 2019-05-18 15-11-23

dagar:
@dogmaphobic I think the only thing I need from you is to comment on the file deletions.

dogmaphobic:
`cc1plus: out of memory allocating 268435456 bytes after a total of 238735360 bytes`

Any ideas what that could be?

stefandunca:
> `cc1plus: out of memory allocating 268435456 bytes after a total of 238735360 bytes`
> 
> Any ideas what that could be?

It sounds like docker container run out of memory.

dagar:
Rebased.

dogmaphobic:
If the issue is within the Docker container, we'll have to wait for @DonLakeFlyer to get back as I have no idea how that is setup.

The Windows issue (AppVeyor) seems related to trying to shove a tag (string) into some numerical field (or something like it).
```
QGroundControl_resource.rc(6) : error RC2104 : undefined keyword or key name: LastQtWidgetBuild
356
```

patrickelectric:
I just tested the PR here and it's working fine locally. Thanks @dagar

dagar:
The Linux release failure on Jenkins is also present in master (and unrelated to cmake). I'll merge this and we can look at fixing it separately.

PR #7463, opened 2019-05-20
Load Korean fonts when needed

dogmaphobic:
We have specific fonts for Korean, which were being loaded when the user selected "Korean" from the list of languages. However, when QGC was set to "System" locale, no test was done to see if System == Korean.

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -391,12 +391,6 @@ void QGCApplication::setLanguage()
             break;
         case 12:
             locale = QLocale(QLocale::Korean);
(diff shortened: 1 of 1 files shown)


kdnguru:
Dear Gus, Thank you for effort. In case of Samsung Android phone, after select "Korean" language, Samsung phone dos not show specific fonts as "NanumGothic-Regular".

kdnguru:
Screenshot_20190521-163918

kdnguru:
Thank you for fixing Korean font name

PR #7467, opened 2019-05-21
MissionManager: correct frame on Rally mission items

peterbarker:
Frame is incorrect; altitudes used are relative.


DonLakeFlyer:
KIcking this through the build system again. I think I may have fixed this in the pull for mavlink negotation but don't remember. So can't hurt.

PR #7469, opened 2019-05-22
Handle auto video stream

dogmaphobic:
If a camera supports the full camera API, including the video streaming API, both camera and video streaming and handled through one same interface. In this case, the "Video" widget is disabled as it becomes redundant.


dogmaphobic:
Travis build for Android is broken for whatever reason. That's a Travis issue, not QGC.

PR #7476, opened 2019-05-27
Change default autopilot to MAV_AUTOPILOT_GENERIC when requesting parameters meta

dvornikov-aa:
ParameterManager uses MAV_TYPE_QUADROTOR as a system type to request parameters meta. Custom FirmwarePlugins may not support quadrotors so it's better to use MAV_TYPE_GENERIC for this purpose.

### DIFF
--- a/src/FactSystem/ParameterManager.cc
+++ b/src/FactSystem/ParameterManager.cc
@@ -1179,7 +1179,7 @@ void ParameterManager::_initialRequestTimeout(void)
 QString ParameterManager::parameterMetaDataFile(Vehicle* vehicle, MAV_AUTOPILOT firmwareType, int wantedMajorVersion, int& majorVersion, int& minorVersion)
 {
     bool            cacheHit = false;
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Hmm, the current code in QGC is simply wrong since it makes a hard-coded assumption of the vehicle type. But your change doesn't quite do the right thing either since it basically does the same thing. I believe correct fix is:

```
    FirmwarePlugin* plugin = vehicle->firmwarePlugin();
```

DonLakeFlyer:
In looking at my suggestion that isn't really correct either since there are other places in ParameterManager which need to get the firmware plugin when it doesn't have a vehicle available. For current usage this still all works fine. 

It would only fail if a custom build for an ArduPilot vehicle removed support for quadrotor in it's FirmwarePluginFactory. Which is a possibility but hasn't been hit yet by anyone. It will require some FirmwarePlugin architecture changes to fix this.

I realize you are trying to create a whole new FirmwarePlugin implementation but QGC isn't really set up for this right now. Hence you are going to run into things like this likely one after the other. It just isn't set up to do that right now.

I'll try to figure out some way to fix this, since an ArduPilot custom build could hit it. But you still may hit a boat anchor somewhere along the way which will not be able to be worked around without major change.

PR #7477, opened 2019-05-28
Add csv logging of telemetry

### LINKED ISSUE #7322: Export layman's log (.csv output!)
In a world of mavlink, missions, binary logs, xmls, and proxies, working out how to produce a log with the information one is after can be a daunting and time consuming task, even for advanced users. At the time of writing, Blue Robotics' best recommendation for producing a .csv from a log file requires a python installation and command line tools. This is far from easy as it can be.

This is a proposal for a simple-as-possible solution that will serve a majority of users and use cases.

Most users will already have some idea of how they would like to handle the data they are interested in. They usually want to import data into a spreadsheet program or other analysis tool like matlab. These tools support importing .csv files. 

The proposed solution is for QGC to export a .csv file with some ~20-50  or so curated fields/columns such as gps coordinates, altitude, attitude, temperature, etc. To keep things simple, the keys chosen should be human readable and self explanatory (including units!). Whatever values/keys we decide to log, there won't be any user configuration; all columns/keys that we choose will always be logged. Spreadsheet programs will be able to filter/render the data however is required in post. The decision to forgo user configuration means that all available data will be in the original csv (no forgetting to turn on logging) while making the implementation very simple (no ui, no configuration), and maintaining minimal overhead.

One suggestion for deciding the columns/keys to be logged is to just use everything that appears as an option for display in the vehicle values widget, shown below. This is the same suggestion as put forward in the [video overlay](https://github.com/mavlink/qgroundcontrol/issues/7286) proposal.

image

Williangalvani:
This makes each vehicle log it's facts to a `.csv` file named as `<timestamp> vehicle<vehicle_ID>.csv`
This is a sample file:  [2019-05-21 18-22-17_vehicle2.txt](https://github.com/mavlink/qgroundcontrol/files/3228873/2019-05-21.18-22-17_vehicle2.txt)

This requires **no interaction of the user whatsoever**, the .csv will be created automatically just like the .tlogs.

This allows users to easily open the data with any spreadsheet software or Matlab, without going through the hassle of using mavlogdump or Mission Planner to deal with a tlog, for example.

close #7322

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -111,6 +111,7 @@ Vehicle::Vehicle(LinkInterface*             link,
     , _soloFirmware(false)
     , _toolbox(qgcApp()->toolbox())
     , _settingsManager(_toolbox->settingsManager())
+    , _csvLogTimer(this)
     , _joystickMode(JoystickModeRC)
     , _joystickEnabled(false)
     , _uas(nullptr)
@@ -302,6 +303,7 @@ Vehicle::Vehicle(LinkInterface*             link,
     connect(&_adsbTimer, &QTimer::timeout, this, &Vehicle::_adsbTimerTimeout);
     _adsbTimer.setSingleShot(false);
     _adsbTimer.start(1000);
+    _initializeCsv();
 }
 
 // Disconnected Vehicle for offline editing
@@ -323,6 +325,7 @@ Vehicle::Vehicle(MAV_AUTOPILOT              firmwareType,
     , _soloFirmware(false)
     , _toolbox(qgcApp()->toolbox())
     , _settingsManager(_toolbox->settingsManager())
+    , _csvLogTimer(this)
     , _joystickMode(JoystickModeRC)
     , _joystickEnabled(false)
     , _uas(nullptr)
@@ -4005,6 +4008,53 @@ void Vehicle::_pidTuningAdjustRates(bool setRatesForTuning)
     _setpointFactGroup.setLiveUpdates(setRatesForTuning);
 }
 
+void Vehicle::_initializeCsv()
+{
+    QString now = QDateTime::currentDateTime().toString("yyyy-MM-dd hh-mm-ss");
+    QString fileName = QString("%1 vehicle%2.csv").arg(now).arg(_id);
+    QDir saveDir(_toolbox->settingsManager()->appSettings()->telemetrySavePath());
+    _csvLogFile.setFileName(saveDir.absoluteFilePath(fileName));
+
+    if (!_csvLogFile.open(QIODevice::Append)) {
+        qCWarning(VehicleLog) << "unable to open file for csv logging, Stopping csv logging!";
+        return;
+    }
+
+    QTextStream stream(&_csvLogFile);
+    QStringList allFactNames;
+    allFactNames << factNames();
+    for (const QString& groupName: factGroupNames()) {
+        for(const QString& factName: getFactGroup(groupName)->factNames()){
+            allFactNames << QString("%1.%2").arg(groupName, factName);
+        }
+    }
+    qCDebug(VehicleLog) << "Facts logged to csv:" << allFactNames;
+    stream << "Timestamp," << allFactNames.join(",") << "\n";
+
+    connect(&_csvLogTimer, &QTimer::timeout, this, &Vehicle::_writeCsvLine);
+    _csvLogTimer.start(1000);
+}
+
+void Vehicle::_writeCsvLine()
+{
+    QStringList allFactValues;
+    QTextStream stream(&_csvLogFile);
+
+    // Write timestamp to csv file
+    allFactValues << QDateTime::currentDateTime().toString();
+    // Write Vehicle's own facts
+    for (const QString& factName : factNames()) {
+        allFactValues << getFact(factName)->cookedValueString();
+    }
+    // write facts from Vehicle's FactGroups
+    for (const QString& groupName: factGroupNames()) {
+        for (const QString& factName : getFactGroup(groupName)->factNames()) {
+            allFactValues << getFactGroup(groupName)->getFact(factName)->cookedValueString();
+        }
+    }
+
+    stream << allFactValues.join(",") << "\n";
+}
 
 #if !defined(NO_ARDUPILOT_DIALECT)
 void Vehicle::flashBootloader(void)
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
> This requires **no interaction of the user whatsoever**

I don't see that as a positive thing. For a number of reasons:
* Hitting the disk kills battery. Since there is no way to turn this off, the user has no mechanism to save battery life.
* Mobile devices can have very limited storage. If the user has no plans of ever using these it increases the rate at which they need to manage that limited storage.

Because of that:
* I think there needs to be an option for turning this on/off just like there is for telemetry logs.
* I think it should also follow the telemetry log setting for only Saving if armed. Lot's of useless data can be generated by just connecting to the vehicle. I think following that setting would be a good idea.

I'm also concerned about limited usability. At a rate of once per second you can lose a lot of interaction depending on what value you are looking at. For example it you are trying to graph something to do with say multi-rotor controls that is way to slow to be of value. It may fit your specific usage pattern nicely, but I'm not convinced it fits a generic usage pattern that well. Given that I think it should default to being off.

If you really/really want a way to convert telemetry logs to .csv format then why not add a feature to Log Replay which can do that? Load the telemetry log from there, and then provide the user with the option to convert the whole thing to csv. That would give you full data, at full update rate.

jaxxzer:
Thank you for the comments @DonLakeFlyer, we can update this according to your recommendation. 

I will make some explanations on the motive for this implementation:
- the *general* use case we are targeting here is 'beginners'; we want something simple because nothing simple exists.
- there are already tools out there that can manipulate binary logs and produce tailored csvs. These are all hard to use for a beginner. We could roll our own with better features, but then exposing mavlink fields for selection becomes hard to implement and hard to use, just like the tools that already exist.
- this is not for **deep** flight analysis or debugging of things like rc values (there are already tools for that), but rather only the values of obvious interest to most users, like coordinates and altitude. The majority of these values change at low frequency. The value labels will also be easy to identify and require no acquaintance with or understanding of mavlink, for example: "altitude" instead of [GLOBAL_POSITION_INT.alt](https://mavlink.io/en/messages/common.html#GLOBAL_POSITION_INT). 

The use case we are imagining here is some user or class of students gets their first drone, takes a flight with qgc, and can get some data into matlab or excel with minimum effort. Today, it is very difficult to navigate the mavlink, flight stack, and log tool documentation to get from flight to workable data.

jaxxzer:
I'll see if mission planner can lend any inspiration in this area. I'm also interested if there are any log analysis tools out there that I'm unaware of. Right now our best recommendation is mavlogdump.py, which is a command line tool.

Our desire is essentially ease of use.

DonLakeFlyer:
I'm fine with the output using the FactGroups to do it for simplicity. My main concern is how this is integrated into QGC. Given the fact the conversion is already written I could see a new tab under Analyze which would input a log file and then convert to csv. That wouldn't get in peoples way for the most folks who won't use it. It also doesn't automatically write things to disk.

src/ui/preferences/GeneralSettings.qml
```diff
@@ -420,6 +420,14 @@ Rectangle {
                                 enabled:    promptSaveLog.checked && !disableDataPersistence.checked
                                 property Fact _telemetrySaveNotArmed: QGroundControl.settingsManager.appSettings.telemetrySaveNotArmed
                             }
+                            FactCheckBox {
+                                id:         promptSaveCsv
+                                text:       qsTr("Save CSV log of telemetry data")
+                                fact:       _saveCsvTelemetry
+                                visible:    _saveCsvTelemetry && _saveCsvTelemetry.visible
```
DonLakeFlyer: No need for ```_saveCsvTelemetry &&```. This will never be null.

Williangalvani: Without it, I get `qrc:/qml/GeneralSettings.qml:427: TypeError: Cannot read property 'visible' of null` when closing QGC. I think that is why the other options have it too.

dogmaphobic: That's correct. I have code identical to that everywhere as those things are indeed NULL during shutdown of the UI.

DonLakeFlyer: Wow, that's annoying. Hadn't seen those in there. They are still missing in quite a few places. There should be a different way to shutdown to make this not happen. I think we can tear down the Windows first before we close the toolbax and this will go away. Anyway, this is fine then. I'll look at changing shutdown as well.

DonLakeFlyer: This null crap happens because the QQmlApplicationEngine is not explictly destroyed in the shutdown sequence. So it ends up being deleted last when all the other toolbox stuff is already gone. I fixed it. I'll remove all the null check cruft in that pull.

So @Williangalvani you should pull out the null checks in your changes.

Williangalvani: Done!

src/Vehicle/Vehicle.cc
```diff
@@ -4005,6 +4011,63 @@ void Vehicle::_pidTuningAdjustRates(bool setRatesForTuning)
...
+
+    if (!_csvLogFile.open(QIODevice::Append)) {
+        qCWarning(VehicleLog) << "unable to open file for csv logging, Stopping csv logging!";
+        return;
+    }
+
+    QTextStream stream(&_csvLogFile);
+    QStringList allFactNames;
+    allFactNames << factNames();
+    for (auto groupName: factGroupNames()) {
```
DonLakeFlyer: I want to keep the codebase away from using ```auto``` in almost all cases.

src/Vehicle/Vehicle.cc
```diff
@@ -4005,6 +4011,63 @@ void Vehicle::_pidTuningAdjustRates(bool setRatesForTuning)
...
+        return;
+    }
+
+    QStringList allFactValues;
+    QTextStream stream(&_csvLogFile);
+
+    // Write timestamp to csv file
+    allFactValues << QDateTime::currentDateTime().toString();
+    // Write Vehicle's own facts
+    for (auto fact : factNames()) {
```
DonLakeFlyer: This is kind of the classic example of why I think ```auto``` is bad. In most cases ```fact``` would represent a Fact object. But in this case, it does not. It is a string. Which you only realize if look further in the code.

DonLakeFlyer: I know I'm a pain. But I have strong opinions on coding style and readability.

DonLakeFlyer:
Fantastic thanks @Williangalvani

src/Vehicle/Vehicle.cc
jaxxzer: good call.

PR #7479, opened 2019-05-29
Quick workaround to fix qwdiget leftovers

stefandunca:
After switching to QML only UI, the `QMessageBox` instances trigger the following error: "QWidget: Cannot create a QWidget without QApplication."
This PR converts QMessageBox popups into regular warnings.


dogmaphobic:
Replaced by #7532

PR #7481, opened 2019-05-30
PX4ParameterMetaData: Remove default version

dvornikov-aa:
Current implementation has an issue: any XML file is interpreted as a valid metadata (even if it's not). Modern px4 firmware has a correct version in the metadata. Why the default version has been added in the first place?


DonLakeFlyer:
What exact problem/bug are you fixing here? I don't understand.

DonLakeFlyer:
After looking through this code path some more to see how the callers use this method I get it now. I think your change is correct (in combination with you other pull). But I need to stare at it a bit more to make sure.

I don't think parameter meta data was always versioned correctly if you go back in time. But as it stands now I think this is always the case as you say.

PR #7482, opened 2019-05-30
Implement overlay on recorded videos by using subtitle files

### LINKED ISSUE #7286: Add telemetry overlay to saved video
This is one of the most highly requested features of sub users. The idea is to have telemetry such as depth, location, time etc overlaid on the video file in order to produce and maintain better records and deliverables. The common standard in the industry is simple text overlay like this:
image

There are subtitle formats ([ssa](https://en.wikipedia.org/wiki/SubStation_Alpha)) that allow positioning and formatting of the text. Media players such as VLC have support for rendering these subtitle files. QGC could produce these subtitle files with the telemetry data synchronized to the video. In order to keep it simple, we recommend limiting the available telemetry overlay to a set of <=20 or so common values like gps coordinates, altitude, temperature etc. One thought was to just use the available 'vehicle values widget' (pictured below) and add a column like 'export' to choose to add a value to the subtitle file. We would like to force the text placement also, for example a title and date at the top right, and all telemetry values in the bottom left (no user configuration).

image

We have considered several other approaches, but we favored the subtitle file for these reasons:
- The original video remains available without any 'burned in' overlay
- Media players already support rendering the subtitle files and burning them into the video file in post, we don't have to implement any of the rendering side ourselves
- As minimal as possible user configuration and intervention - easy for them!
- Implementation requires simply producing the subtitle file which is in ascii format - easy for us!

If you guys approve of the approach, @patrickelectric @Williangalvani and I can do the work to make it happen.

Williangalvani:
This patch allows QGC to record a data overlay while recording the video. This is implemented by producing an independent subtitle file for use with each video that is recorded to file.  The subtitle file is a text file in the [SubStation Alpha](https://en.wikipedia.org/wiki/SubStation_Alpha) format. This is supported by your usual media players, where you can control if the overlay is displayed on playback, or optionally combine the video and subtitle file and export a single video file with the subtitles 'burned in'.

This patch will close #7286. No additional UI elements or user configuration/options are introduced. The values that will be recorded are the same as those that are selected for display in the vehicle values widget (telemetry display). Telemetry values are sampled and recorded at a fixed 1Hz. The file is produced automatically.

The result looks like this:

Screenshot from 2019-05-30 10-12-21

The data rate is about 1kb/s. In a test with FHD video, 3 hours of footage took 8.5GB of disk, with a 7.4MB subtitle file.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -1189,12 +1189,14 @@ HEADERS += \
     src/VideoStreaming/VideoStreaming.h \
     src/VideoStreaming/VideoSurface.h \
     src/VideoStreaming/VideoSurface_p.h \
+    src/VideoStreaming/SubtitleWriter.h \
 
 SOURCES += \
     src/VideoStreaming/VideoItem.cc \
     src/VideoStreaming/VideoReceiver.cc \
     src/VideoStreaming/VideoStreaming.cc \
     src/VideoStreaming/VideoSurface.cc \
+    src/VideoStreaming/SubtitleWriter.cc \
 
 contains (CONFIG, DISABLE_VIDEOSTREAMING) {
     message("Skipping support for video streaming (manual override from command line)")
--- a/src/FlightDisplay/VideoManager.cc
+++ b/src/FlightDisplay/VideoManager.cc
@@ -81,6 +81,7 @@ VideoManager::setToolbox(QGCToolbox *toolbox)
     _updateSettings();
     if(isGStreamer()) {
         startVideo();
+        _subtitleWriter.setVideoReceiver(_videoReceiver);
     } else {
         stopVideo();
     }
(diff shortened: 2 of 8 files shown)


dogmaphobic:
This is cool. We can tweak it as needed later.

DonLakeFlyer:
Are there any docs written for this?

Williangalvani:
> Are there any docs written for this?

Working on it. I'll open a PR to put it in the daily features soon.

CharlesY987:
How does it work under Windows ? I'm using QGroundControl to manipulate a BlueROV2. Could you help me ?

Williangalvani:
> How does it work under Windows ? I'm using QGroundControl to manipulate a BlueROV2. Could you help me ?

Take a look [at the docs here](https://docs.qgroundcontrol.com/en/FlyView/VideoOverlay.html).
This only works for playback, and is not present on any stable releases yet. If you want to test with a [daily build](https://docs.qgroundcontrol.com/en/releases/daily_builds.html), I recommend using VLC for playback.

LPNautic:
Hi
When can we have this function in Ardusub Stable ??
Thanks

petegreg:
This is a great feature! It would also be very helpful to have the flight mode included in the overlay.

PR #7483, opened 2019-05-30
Add What3Words API implementation

szulak:
This PR adds a simple implementation of What3Words (https://what3words.com) platform which could be useful in the future.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -362,6 +362,7 @@ INCLUDEPATH += \
     src/Terrain \
     src/VehicleSetup \
     src/ViewWidgets \
+    src/What3Words \
     src/Audio \
     src/comm \
     src/input \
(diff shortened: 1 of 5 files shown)


jaxxzer:
What3Words is a system for specifying geographical coordinates with 3-word combinations instead of numbers/coordinates. The system provides a lot of resolution and makes it easy to remember/communicate locations compared to conventional addresses/methods. It's a cool idea.

@szulak It looks like you have added the API, but haven't put it to use. Do you have plans for subsequent work here? Does this API require an internet connection to work?

szulak:
@jaxxzer Yes, I do. I would like to add an option to display What3Words address near the latitude and longitude pair in some widgets.

It would be configurable via settings panel as it requires the internet connection in order to work.

DonLakeFlyer:
> The system provides a lot of resolution and makes it easy to remember/communicate locations compared to conventional addresses/methods.

Correct. Useful for transcribing lat/lon positions without mistakes and or communicating lat/lon positions without mistakes over a radio. But I don't get the key value for QGC? We are already running into issues with App size on mobile platforms so we have to be careful about adding things. I'm not sure the additional ui/code for this has enough value.

PR #7498, opened 2019-06-05
Parameter meta data cleanup

DonLakeFlyer:
Fix problems with parameter meta data methods not conforming to spec as well as using incorrect vehicle type.

Replacement for #7481, #7480, #7476


DonLakeFlyer:
@dvornikov-aa This should accomplish what you were changing in your pulls.

dvornikov-aa:
@DonLakeFlyer I've tested this PR. Everything is working.

DonLakeFlyer:
@dvornikov-aa Thanks for verifying

PR #7512, opened 2019-06-10
User hardware video decoding on Android

MatejFranceskin:
GStreamer is setup to always use software video decoding which is consuming more CPU power and battery.
This change introduces hardware video decoding on Android using Android Media gstreamer plugin which is always optimized for particular platform.
If hardware decoding fails for some reason the system falls back to software decode.


xgdwhy:
hi, there is a problem that using android media codec use more time and higher latency than gst 1.5.2 using software video decoder

xgdwhy:
and when play video , the frame often jump to the frame have played

Issue #7517, opened 2019-06-12
Strange audio language on Ubuntu 18.04 + workaround 


ArkadiuszNiemiec:
## Expected Behavior
I expect the QTextToSpeech to use system locale to generate audio or TTS.

## Current Behavior
I get English TTS in a strange accent which is almost not understandable.

## Steps to Reproduce:
1. Fresh Ubuntu 18.04 configured with US locales (installed in Poland and connected to local wifi)
2. Install QGroundControl (daily or stable)
3. Connect to PX4
4. Listen

## System Information
- Ubuntu 18.04
- QGC build: Daily, stable or even builded from source

## Detailed Description
I have checked the `speech-dispatcher` and it works with `espeak`. I can confirm that it works perfectly by typic `spd-say "Hold flight mode"` All builds of QGroundControl have strange English accent and I cannot find a source of it.

In the meantime I have found a workaround: I can get clean English by adding this line:
```
_tts->setLocale(QLocale())
```
to https://github.com/mavlink/qgroundcontrol/blob/eee967c549d4958907f97c716f98b1c44c9b963f/src/Audio/AudioOutput.cc#L19-L24

Even stranger: I have checked the `_tts` in the debugger and it shows `en_US` before `_tts->setLocale(QLocale())` but still it fixes it.

I can also confirm that I have `en_US` locale by watching QGroundControl logs:
```
System reported locale: QLocale(English, Lating, UnitedStates) "en_US"
```

ArkadiuszNiemiec:
This bug report confirms my theory: https://bugreports.qt.io/browse/QTBUG-73552

dogmaphobic:
This is a dupe of some other issue but your bug research made a lot simpler to fix it. On it...

dogmaphobic:
Can you please try this and see if this makes any sense? As I explained in the PR, we may need to do the opposite (force it to use English always, regardless of your locale): https://github.com/mavlink/qgroundcontrol/pull/7518

dogmaphobic:
See https://github.com/mavlink/qgroundcontrol/pull/7518#issuecomment-501668327

ArkadiuszNiemiec:
Fixed by #7518

Niel66:
It was not fixed in the .appimage file
Here is the accent in which it spoke
[...]

PR #7518, opened 2019-06-13
Handle text to speech locale

dogmaphobic:
In response to #7517 

This sets the Text to Speech engine to the current locale. I'm not sure how this will work as 99% of the spoken sentences are in English and are not localized. @ArkadiuszNiemiec could please test this and see how it behaves? This "fix" might be the opposite of what is required. I'm thinking it makes more sense to set the TTS engine to always use English regardless of the current locale as that's the only source of text it will ever see. If that's the case, let me know and I will switch to "always use English" instead.

I can't really test as my system is set to its default (en_US). If I switch the language in QGC, the spoken text is obviously still English but with a weird accent.

### DIFF
--- a/src/Audio/AudioOutput.cc
+++ b/src/Audio/AudioOutput.cc
@@ -20,7 +20,14 @@ AudioOutput::AudioOutput(QGCApplication* app, QGCToolbox* toolbox)
     : QGCTool(app, toolbox)
     , _tts(new QTextToSpeech(this))
 {
+    _tts->setLocale(QLocale::system());
     connect(_tts, &QTextToSpeech::stateChanged, this, &AudioOutput::_stateChanged);
+    connect(qgcApp(), &QGCApplication::languageChanged, this, &AudioOutput::_languageChanged);
+}
+
+void AudioOutput::_languageChanged(const QLocale locale)
+{
+    _tts->setLocale(locale);
 }
 
 bool AudioOutput::say(const QString& inText)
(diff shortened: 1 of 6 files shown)


ArkadiuszNiemiec:
You are right, it should be always set to English if it does not have strings in other languages. I am not able to test it with other locales as my system is also set to `en_US` but I can say that this PR fixes my problem. Explicitly setting the locale of `_tts` object fixes the bug in QT https://bugreports.qt.io/browse/QTBUG-73552

ArkadiuszNiemiec:
I have build this PR and it works perfectly when the language is set to English or System. Otherwise it TTS speaks in strange english (as it should), so I will keep this build but would love a commit with the language of TTS always set to English.

ArkadiuszNiemiec:
I am not a contributor and cannot commit in this PR :disappointed: 
This code is tested and working:
```
AudioOutput::AudioOutput(QGCApplication* app, QGCToolbox* toolbox)
    : QGCTool(app, toolbox)
    , _tts(new QTextToSpeech(this))
{
    _tts->setLocale(QLocale("en_US"));
    connect(_tts, &QTextToSpeech::stateChanged, this, &AudioOutput::_stateChanged);
}
```

dogmaphobic:
I've inverted the logic and I am now forcing the TTS engine to use English. But only for Linux. When I do that on macOS, the output sounds like an asthmatic Stephen Hawking instead of the natural speech you normally get. If this issue shows up on other platforms, we will need to find a way to work around this.

ArkadiuszNiemiec:
@dogmaphobic looks great! Running perfectly on my system.

ArkadiuszNiemiec:
@dogmaphobic I am not able to build a release version on my Ubuntu. Do you know when will this be available in the daily build?

dogmaphobic:
In a few minutes. The daily runs right after the PR is pushed. The build itself is quick but the queue within Travis not so much.

ArkadiuszNiemiec:
Oh, that's great! Thank you for your work!

PR #7521, opened 2019-06-13
Initial support for PX4 Rover

dogmaphobic:
This is just the beginning. Currently, a lot of the variables that define a vehicle type are specified on a per firmware plugin type. That works for APM given that it has different firmwares for each vehicle type but not so much for PX4 where one single firmware supports everything.

This change assigns a vehicle type during the creation of an instance of `FirmwarePlugin`. It further implements `supportsNegativeThrust()` based on the vehicle type (for PX4 only).

Now we just need to do the other 99.9% of missing rover support :P

@DonLakeFlyer, @dagar The whole Joystick/Radio setup is hardcoded for full, 3 axis attitude in addition to throttle. We need to sort out how to handle just "steering" and "throttle" (and a button for brake, etc.)

### DIFF
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -50,11 +50,6 @@ FirmwarePluginFactoryRegister* FirmwarePluginFactoryRegister::instance(void)
     return _instance;
 }
 
-FirmwarePlugin::FirmwarePlugin(MAV_TYPE vehicleType)
-{
-    _vehicleType = vehicleType;
-}
-
 AutoPilotPlugin* FirmwarePlugin::autopilotPlugin(Vehicle* vehicle)
 {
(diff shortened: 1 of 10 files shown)


src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -50,6 +50,11 @@ FirmwarePluginFactoryRegister* FirmwarePluginFactoryRegister::instance(void)
     return _instance;
 }
 
+FirmwarePlugin::FirmwarePlugin(MAV_TYPE vehicleType)
```
DonLakeFlyer: You can't do this. There is a single FirmwarePlugin created for all PX4 vehicles. The same one just keeps getting handed out.

dogmaphobic: Ha! So all those calls (such as `supportsNegativeThrust()`) must pass a `Vehicle` instance then as there is no other way to figure it out.

PR #7528, opened 2019-06-14
Moved the preflight checklist to a proper, modal dialog.

dogmaphobic:
I don't quite grok the over-complexity/hardcoded dependencies of that Preflight Checklist. I will deal with that when I find the time. In the mean time, I've made it into a proper, modal dialog using current QtQuick controls.

Screen Shot 2019-06-14 at 3 24 18 PM

### DIFF
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -627,7 +627,6 @@ Item {
                 {
                     name:               "Checklist",
                     iconSource:         "/qmlimages/check.svg",
-                    dropPanelComponent: checklistDropPanel
                 },
                 {
                     name:       _guidedController.takeoffTitle,
@@ -660,6 +659,8 @@ Item {
                 guidedActionsController.closeAll()
                 if(index === 0) {
                     mainWindow.showPlanView()
+                } else if(index === 1) {
+                    checklistDropPanel.open()
                 } else {
                     var action = model[index].action
                     if (action === -1) {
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
I'm not sure it makes a lot of sense for regular QGC. The checklist tends to be vehicle and type specific. Which is the reason it's off by default. The main usage at this point is in custom builds for commercial vehicles.

dogmaphobic:
All I did was to make it non modal. It was already there.

DonLakeFlyer:
You click away to close it?

dogmaphobic:
Either use the Escape key or click anywhere outside.

DonLakeFlyer:
Does the height calc work the same way? I can't quite tell. I remember fighting with that for Sentera.

dogmaphobic:
> Does the height calc work the same way? I can't quite tell. I remember fighting with that for Sentera.

That whole thing needs a lot of work. The implementation is way too convoluted. The main Popup size/position is fine. If the dialog within can't handle dynamic sizing, that's another story. It seemed to all work though.

If this is to be used only by custom builds, it needs to be exposed as such. Everything in there is hard coded. This will need to be moved to a QGCCorePlugin method, which, if implemented, provides the UI. If this is to be absent from QGC proper, the method returns nothing.

The reason for this PR was a complaint that, while the Preflight Checklist dialog was up, you could go into General Settings, disable it and it would still be there. You could also go from flight to plan view and it would stay put, in the way of other UI elements. Making it modal was the logical "fix". By using a proper `Popup`, it also grays out the rest of the UI.

DonLakeFlyer:
> The reason for this PR was a complaint that, while the Preflight Checklist dialog was up, you could go into General Settings, disable it and it would still be there.

Isn't that a DropPanel bug? Close DropPanels if you change views?

dogmaphobic:
>Isn't that a DropPanel bug? Close DropPanels if you change views?

Yes, we need to do away with `DropPanel` and covert them all to Popup as well. This was the first one.

Issue #7427, opened 2019-05-09
Rally point upload/download sometimes "not supported"


rmackay9:
ArduPilot is adding support for the MISSION_ITEM protocol (PR is here:https://github.com/ArduPilot/ardupilot/pull/10409) but during testing with QGC we are finding that, for whatever reason, QGC is often (but not always) showing "This vehicle does not support Rally Points". 

Here it is not working:
![QGC](https://user-images.githubusercontent.com/1498098/57427356-824baa80-725e-11e9-877a-ee0895d24d93.png)

Here it is working moments later after a restart:
![QGC_working](https://user-images.githubusercontent.com/1498098/57427523-3816f900-725f-11e9-9bed-cbe92c953930.png)

rmackay9:
I wonder if perhaps QGC is not requesting the AUTOPILOT_VERSION using MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES because I'm seeing this odd behaviour:

1. Start board and QGC and connect - it will not be possible to upload rally point
2. Shutdown QGC, connect with MP
3. Disconnect MP, re-connect with QGC - it will be possible to upload rally points

peterbarker:
The first two lines of output here are my debug:

```
##### Got autopilotversion!
setcapa(48079)
VehicleLog: "Vehicle does support MAVLink 2 but the link does not allow for it."
VehicleLog: "Vehicle supports Mavlink 2.0"
VehicleLog: "Vehicle supports MISSION_ITEM_INT"
VehicleLog: "Vehicle supports MISSION_COMMAND_INT"
VehicleLog: "Vehicle does not support GeoFence"
VehicleLog: "Vehicle supports RallyPoints"
VehicleLog: Delaying _startPlanRequest due to parameters not ready
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
VehicleLog: _sendMessageMultipleNext: 66
[...]

PR #7533, opened 2019-06-15
Mavlink protocol version negotiation fixes

DonLakeFlyer:
Related to #7427

Fixed protocol negotation when the vehicle does not support MAV_CMD_REQUEST_PROTOCOL_VERSION.

Here is how the negotation now works:
* Start with the assumption that the capability bits are known
* Send MAV_CMD_REQUEST_PROTOCOL_VERSION to the vehicle
* Start a timeout on the COMMAND_ACK for MAV_CMD_REQUEST_PROTOCOL_VERSION
  * If GCS receives COMMAND_ACK.result != MAV_RESULT_ACCEPTED or times out waiting for the COMMAND ACK then set the link to Mavlink 2 if the capability bits say it is supported
  * If GCS receives COMMAND_ACK.result == MAV_RESULT_ACCEPTED start a timeout waiting to receive a PROTOCOL_VERSION messsage
    * If PROTOCOL_VERSION is never received then set the link to Mavlink 1
    * If PROTOCOL_VERSION message is received set the link accordingly to 1 or 2

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -141,6 +141,7 @@ Vehicle::Vehicle(LinkInterface*             link,
     , _telemetryTXBuffer(0)
     , _telemetryLNoise(0)
     , _telemetryRNoise(0)
+    , _mavlinkProtocolRequestComplete(false)
     , _maxProtoVersion(0)
     , _vehicleCapabilitiesKnown(false)
     , _capabilityBits(0)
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@peterbarker I believe this should work for ArduPilot. Can you verify that against what you know about how it works.
@hamishwillee This differs from existing docs.

DonLakeFlyer:
I still need to do some more testing with ArduPilot in various cases but I think this is good.

peterbarker:
On Sat, 15 Jun 2019, Don Gagne wrote:

> I still need to do some more testing with ArduPilot in various cases but I think this is good.

Just tested in the same way I recall testing before - and it all seems to 
work, which is fantastic!

But please, do test it yourself :-)

Thanks,

Peter

DonLakeFlyer:
Ah, good news. Thanks for testing @peterbarker. This is in the latest daily builds now. So you can use those.

hwang12345:
> Related to #7427
> 
> Fixed protocol negotation when the vehicle does not support MAV_CMD_REQUEST_PROTOCOL_VERSION.
> 
> Here is how the negotation now works:
> 
> * Start with the assumption that the capability bits are known
> * Send MAV_CMD_REQUEST_PROTOCOL_VERSION to the vehicle
> * Start a timeout on the COMMAND_ACK for MAV_CMD_REQUEST_PROTOCOL_VERSION
>   
>   * If GCS receives COMMAND_ACK.result != MAV_RESULT_ACCEPTED or times out waiting for the COMMAND ACK then set the link to Mavlink 2 if the capability bits say it is supported
>   * If GCS receives COMMAND_ACK.result == MAV_RESULT_ACCEPTED start a timeout waiting to receive a PROTOCOL_VERSION messsage
>     
>     * If PROTOCOL_VERSION is never received then set the link to Mavlink 1
>     * If PROTOCOL_VERSION message is received set the link accordingly to 1 or 2

I found QGroundControl can not receive protocol_version message even vehicle has send protocol version message.i can not found the reason.

PR #7539, opened 2019-06-20
Expose BATT_AMP_OFFSET and add Blue Robotics power sense module

Williangalvani:
This exposes the BATT_AMP_OFFSET parameter for ArduPilot vehicles and adds an entry for the [Blue Robotics Power Sense Module R2](https://www.bluerobotics.com/store/comm-control-power/elec-packages/psm-asm-r2-rp/)

Screenshot from 2019-06-20 16-30-02

Screenshot from 2019-06-20 17-32-42

### DIFF
--- a/src/AutoPilotPlugins/APM/APMPowerComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMPowerComponent.qml
@@ -337,6 +337,15 @@ SetupPage {
                 }
 
                 ListElement {
+                    text:       qsTr("Blue Robotics Power Sense Module R2")
+                    voltPin:    2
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
How far back does BATT_AMP_OFFSET param go as far as being available in firmwares? If it's missing QGC will puke on it.

Williangalvani:
The parameters is old: https://github.com/ArduPilot/ardupilot/commit/8f108e2eb3b931b9fd988a1bdba6d0a14ec70d25
I also tested the stable releases for the other vehicles with no issues.

PR #7555, opened 2019-06-28
Fix dangling pointer access

stefandunca:
Because of `_cameras` QML bindings check for `nullptr` won't work in the binding pipeline. The dangling pointer access will cause a runtime fault

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -551,8 +551,11 @@ Vehicle::~Vehicle()
 void Vehicle::prepareDelete()
 {
     if(_cameras) {
-        delete _cameras;
+        // because of _cameras QML bindings check for nullptr won't work in the binding pipeline
+        // the dangling pointer access will cause a runtime fault
+        auto tmpCameras = _cameras;
         _cameras = nullptr;
+        delete tmpCameras;
         emit dynamicCamerasChanged();
         qApp->processEvents();
     }


dogmaphobic:
Nasty business. I wonder how many of these are loose around the code. @DonLakeFlyer, I haven't tested on Windows but this might be the answer to that Windows crash.

src/Vehicle/Vehicle.cc
```diff
@@ -551,8 +551,11 @@ Vehicle::~Vehicle()
 void Vehicle::prepareDelete()
 {
     if(_cameras) {
-        delete _cameras;
```
DonLakeFlyer: FYI: For Qt you almost never want to use c++ delete. It is way safer to alway suse ->deleteLater due to various QObject references.

DonLakeFlyer:
> I haven't tested on Windows but this might be the answer to that Windows crash.

The windows crash is on boot. This code would fix when a Vehicle goes away which wouldn't be the case I think.

src/Vehicle/Vehicle.cc
stefandunca: I was thinking on this as well but It looked safer to keep the current tested behavior. Otherwise there is need for investigation: will the `control loop` run so that the object is destroyed? is there any side effect for the object living after the `Vehicle` is deleted?

DonLakeFlyer: I don't quite get what this prepareDelete thing is supposed to do and why it is needed? If the list of cameras is associated with the vehicle then why isn't the QObject for it parented to Vehicle? Then it will be destroyed when the Vehicle goes away and you don't need any of this prepareDelete thing.

PR #7556, opened 2019-06-30
CMake updates: Enable VideoStreaming while using CMake as buildsystem

tcanabrava:
VideoStreaming is not enabled while building using CMake, we missed the gstreamer libraries and a few other things. I also took the time to make cmake a bit more easy to handle.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,40 +127,8 @@ if(NOT QT_MKSPEC MATCHES "winrt")
 	)
 endif()
 
-if(${CMAKE_BUILD_TYPE} MATCHES "Debug")
-	include(CTest)
-	enable_testing()
-	add_definitions(-DUNITTEST_BUILD)
-endif()
-
-if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-	# clang and AppleClang
-	add_compile_options(
-		-Wall
-		-Wextra
-
-		-Wno-address-of-packed-member # ignore for mavlink
-	)
-
-elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-	# GCC
-	if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 4.9)
-		add_compile_options(-fdiagnostics-color=always)
-	endif()
-
-	add_compile_options(
-		-Wall
-		-Wextra
-	)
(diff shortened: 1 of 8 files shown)


dagar:
Awesome! Thanks @tcanabrava 

Can you address to CI failures on Jenkins? I'm happy to merge this once it's passing. http://ci.px4.io:8080/blue/organizations/jenkins/mavlink%2Fqgroundcontrol/detail/PR-7556/1/pipeline


So far cmake is my little side project (that's probably an annoyance for @DonLakeFlyer and @dogmaphobic at the moment). I'm hoping it won't be too long until we can switch over properly. At the moment I'd say the main blocker is making the Qt Android and iOS workflow as seamless with cmake as it currently is with qmake.

tcanabrava:
> Awesome! Thanks @tcanabrava
> 
> Can you address to CI failures on Jenkins? I'm happy to merge this once it's passing. http://ci.px4.io:8080/blue/organizations/jenkins/mavlink%2Fqgroundcontrol/detail/PR-7556/1/pipeline
> 
> So far cmake is my little side project (that's probably an annoyance for @DonLakeFlyer and @dogmaphobic at the moment). I'm hoping it won't be too long until we can switch over properly. At the moment I'd say the main blocker is making the Qt Android and iOS workflow as seamless with cmake as it currently is with qmake.

I'v also tried to make them switch - tougth it was three or four years ago. it's good to see that somebody restarted what I tried in the past. I have tons of problems with qmake that are not even funny to start with, and I belive having a strong cmake buildsystem will help a lot.

tcanabrava:
it's annoying to test when your machine builds and travis doesn't.

dagar:
Thank you!

DonLakeFlyer:
> At the moment I'd say the main blocker is making the Qt Android and iOS workflow as seamless with cmake as it currently is with qmake.

It also needs full support for the custom build override capability which is the in qmake files. I don't think that's in there yet either. @dagar You can look at Sentera build as a good example for this.

dogmaphobic:
Or just rename `custom-example` to `custom`.

https://github.com/mavlink/qgroundcontrol/tree/master/custom-example

dagar:
Should we get an issue going to track what it would actually take to switch over 100%?

PR #7564, opened 2019-07-04
Airframe Setup (custom types)

dogmaphobic:
Have a separate panel for Airframe Setup if the firmware is using a custom type:

Screen Shot 2019-07-03 at 5 00 53 PM


hamishwillee:
Cool. Technically "that" is better English:
image

dogmaphobic:
> Technically "that" is better English:

Done.

PR #7566, opened 2019-07-06
Use std::min and std::max instead of rolling our own

tcanabrava:


src/FactSystem/FactMetaData.cc
```diff
@@ -1284,28 +1284,12 @@ QMap<QString, FactMetaData*> FactMetaData::createMapFromJsonArray(const QJsonArr
 
 QVariant FactMetaData::cookedMax(void) const
 {
-    // We have to be careful with cooked min/max. Running the raw values through the translator could flip min and max.
```
DonLakeFlyer: Better to keep the comments here to explain why this is important. Also most QGC code uses qMin/qMax which knows how to coerce between the various Qt types. And deals with any cross-platform wierdness.

PR #7569, opened 2019-07-09
Remove mentions to MSVC < 2015

tcanabrava:
We don't support any other msvc anyway


tcanabrava:
linux release breaking with a windows only change?

dogmaphobic:
> linux release breaking with a windows only change?

Build system issue, not code issue:

`out of memory allocating 134217728 bytes after a total of 188006400 bytes`

dagar:
That's been broken for weeks. http://ci.px4.io:8080/blue/organizations/jenkins/mavlink%2Fqgroundcontrol/activity?branch=master

PR #7570, opened 2019-07-09
Cleanup append for initialization

tcanabrava:
<< for initialization of lists and vectors are bad because they can realloc, passing the whole list using bracket initializations are much faster as there's only one allocation.


dagar:
It also fixes the linux release issue on Jenkins (docker).

PR #7574, opened 2019-07-11
Old instrument panel layout as alternative

tilaktilak:
We propose to re-add a layout for the instrument panel that is similar to the old one as an alternative setting.

- Added a setting in "General Settings/Fly View" to enable or disable the use of the vertical layout (Disabled by default).

- Renamed `QGCInstrumentWidgetAlternate.qml` to `QGCInstrumentWidget.qml` as it has became the main one, and put the vertical one in `QGCInstrumentWidgetAlternate.qml`.   
Looks like the `QGCInstrumentWidget.qml` file was not maintained anymore (See 5df0103)

Here how it looks : 
Kept the changes from the current version and stacked up Attitude and Compass widget differently.
image

Here the setting added : "Use Vertical Instrument Panel"
image

Tests made :
- Resize window 
- Change font size
- Insure setting is disabled by default (So the current Instrument panel is used)
- Insure setting is persistent while restarting the app

@edwinhayes FYI


dogmaphobic:
This seems reasonable to me but let's wait for @DonLakeFlyer to get back as he has a better handle on that widget and its use.

tilaktilak:
@DonLakeFlyer Did you have any chance to review this request ? I'll be glad to have your opinion on this.
Thanks in advance.

DonLakeFlyer:
I'll get to it at some point once I finish other fires. Just back from out of town today.

DonLakeFlyer:
Seems to be ok. But the problem with the old instrument panel is that it's simply too tall. Leaving no space for Values on things like tablets or phones. So the vertical display only works on laptops. Also the need for such large instrument display just didn't seem to add a lot of value. Curious as to why you think such large display is useful? For most people flying missions things should just work without having to stare at intrument displays the whole time.

The other reason I push back is that major changes in UI which are behind the non-default settings have a tendency to break since so few people use them. Hence I try to keep them to as few as possible.

edwinhayes:
@DonLakeFlyer Sure, it does use much more real estate; we wouldn't suggest changing the default back to this.  But not everyone is using a tablet, and/or only flying waypoint missions; we're also interested in the case of operating from a desktop with a large monitor, in which case the little instruments are rather tiny.  Our teleoperation experiences have taught us that having a usable set of instruments is still important.

Equally, we understand the desire to keep things in the mainline simple; we'll be making lots of other changes which we can plugin in our own build, but since this is a significant layout option for the default instrument cluster, seemed like this would be prudent to include.

DonLakeFlyer:
@tilaktilak Have you tested this out/code it to make sure the values widget only grows in height to available amount of real esate based on window height.  Without that the lower instrument can go outside the bottom of the window and not be seen if there are too many values displayed.

tilaktilak:
I have tested that reducing the window to the minimum size will have a similar behavior as the old layout : Crop the value page and show a scrollbar inside.    

I have : `_availableValueHeight: maxHeight - (attitude.height + compass.height)`   

And the max height of the PageView is set as this :
```
        PageView {
            id:                 _valuesWidget
            anchors.margins:    1
            anchors.left:       parent.left
            anchors.right:      parent.right
            maxHeight:          _availableValueHeight
        }
```    

Here the minimum size : 
image
   

Here another screenshot to show the scrollbar :
image

And in fullsize : 
image

src/Settings/FlyView.SettingsGroup.json
```diff
@@ -18,5 +18,11 @@
     "shortDescription": "Show/Hide Log Replay status bar",
     "type":             "bool",
     "defaultValue":     false
+},
+{
+    "name":             "alternateInstrumentPanel",
+    "shortDescription": "Use Alternate Instrument Panel instead of the default one",
```
DonLakeFlyer: This should match the wording show to the user to some extent. So should use the word "Vertical".

DonLakeFlyer:
Other than the one wording changes this is good to go then

tilaktilak:
I changed the wording then :-)

PR #7575, opened 2019-07-11
Dual battery display

stefandunca:
Support for the new changes in `PX4` firmware: https://github.com/PX4/Firmware/pull/12034

**Changes:**
- Provide the battery info from `SYS_STATUS` and `HIGH_LATENCY2` message as `Vehicle.batterySummary` and keep the old us `Vehicle.battery` and `Vehicle.battery2` for first and second battery (optional).
- Take individual battery voltage from battery cell voltages
- Optional more battery details in the detail box
- Bigger fonts for better visibility on small tablets

Note:
 - Keeps backward compatibility

pr-battery
pr-battery_details

### DIFF
--- a/src/AutoPilotPlugins/PX4/PowerComponent.qml
+++ b/src/AutoPilotPlugins/PX4/PowerComponent.qml
@@ -130,7 +130,7 @@ SetupPage {
                                     QGCTextField { id: measuredVoltage }
 
                                     QGCLabel { text: qsTr("Vehicle voltage:") }
-                                    QGCLabel { text: controller.vehicle.battery.voltage.valueString }
+                                    QGCLabel { text: controller.vehicle.batterySummary.voltage.valueString }
 
                                     QGCLabel { text: qsTr("Voltage divider:") }
                                     FactLabel { fact: battVoltageDivider }
@@ -144,7 +144,7 @@ SetupPage {
                                         if (measuredVoltageValue === 0 || isNaN(measuredVoltageValue)) {
                                             return
                                         }
-                                        var newVoltageDivider = (measuredVoltageValue * battVoltageDivider.value) / controller.vehicle.battery.voltage.value
+                                        var newVoltageDivider = (measuredVoltageValue * battVoltageDivider.value) / controller.vehicle.batterySummary.voltage.value
                                         if (newVoltageDivider > 0) {
                                             battVoltageDivider.value = newVoltageDivider
                                         }
@@ -188,7 +188,7 @@ SetupPage {
                                     QGCTextField { id: measuredCurrent }
 
                                     QGCLabel { text: qsTr("Vehicle current:") }
-                                    QGCLabel { text: controller.vehicle.battery.current.valueString }
+                                    QGCLabel { text: controller.vehicle.batterySummary.current.valueString }
 
                                     QGCLabel { text: qsTr("Amps per volt:") }
                                     FactLabel { fact: battAmpsPerVolt }
@@ -202,7 +202,7 @@ SetupPage {
                                         if (measuredCurrentValue === 0) {
                                             return
                                         }
-                                        var newAmpsPerVolt = (measuredCurrentValue * battAmpsPerVolt.value) / controller.vehicle.battery.current.value
+                                        var newAmpsPerVolt = (measuredCurrentValue * battAmpsPerVolt.value) / controller.vehicle.batterySummary.current.value
                                         if (newAmpsPerVolt != 0) {
                                             battAmpsPerVolt.value = newAmpsPerVolt
                                         }
--- a/src/AutoPilotPlugins/PX4/PowerComponent.qml
+++ b/src/AutoPilotPlugins/PX4/PowerComponent.qml
@@ -130,7 +130,7 @@ SetupPage {
                                     QGCTextField { id: measuredVoltage }
 
                                     QGCLabel { text: qsTr("Vehicle voltage:") }
-                                    QGCLabel { text: controller.vehicle.batterySummary.voltage.valueString }
+                                    QGCLabel { text: controller.vehicle.battery1.voltage.valueString }
 
                                     QGCLabel { text: qsTr("Voltage divider:") }
                                     FactLabel { fact: battVoltageDivider }
@@ -144,7 +144,7 @@ SetupPage {
                                         if (measuredVoltageValue === 0 || isNaN(measuredVoltageValue)) {
                                             return
                                         }
-                                        var newVoltageDivider = (measuredVoltageValue * battVoltageDivider.value) / controller.vehicle.batterySummary.voltage.value
+                                        var newVoltageDivider = (measuredVoltageValue * battVoltageDivider.value) / controller.vehicle.battery1.voltage.value
                                         if (newVoltageDivider > 0) {
                                             battVoltageDivider.value = newVoltageDivider
                                         }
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
My concern here is how this may break existing users configuration of the instrument panel. Not quite sure if it does or not.

DonLakeFlyer:
Also this seems to apply semantics to the SYS_STATUS mavlink message which are not really spec'ed. Specifically that they provide a summary or I guess a combination of all the batteries on the vehicle. If that's the case then that need to go through mavlink change appproval. I'm not sure how that would affect ArduPilot reporting for example.

src/AutoPilotPlugins/PX4/PowerComponent.qml
```diff
@@ -130,7 +130,7 @@ SetupPage {
                                     QGCTextField { id: measuredVoltage }
 
                                     QGCLabel { text: qsTr("Vehicle voltage:") }
-                                    QGCLabel { text: controller.vehicle.battery.voltage.valueString }
```
DonLakeFlyer: I don't think this is correct. Current PowerComponent setup ui is only valid for a single battery. How does firmware handle multiple batteries? I would assume each battery has it's own power controller which would mean they could differ?

DonLakeFlyer: This is how ArduPilot power setup works. There are two separate full sets of parameters for each battery.

DonLakeFlyer:
> * Bigger fonts for better visibility on small tablets

Not quite sure how this differs from the other dropdowns font usage. But they should be consistent. It will look strange if the font size differs for no reason between different dropdown text.

DonLakeFlyer:
Is live current draw really something that needs to be in the dropdown. Seems like overkill to have a "More Details" for a single field. I don't see how a normal user would need to see that value ever.

DonLakeFlyer:
> My concern here is how this may break existing users configuration of the instrument panel. Not quite sure if it does or not.

Staring at this some more I think this is ok. 

I wonder though if we should transition from "battery" as the FactGroup name for battery 1 to "battery1" in order to be more clear when picking instrument panel values? This would require back compat support in ValuesWidgetController.

DonLakeFlyer:
> Seems like overkill to have a "More Details" for a single field.

I see that from the code there are "more" more details than are shown in the example. So I guess this is ok. One comment would be to add the "more details" values only to the bottom of the list as opposed to having them show up as new values in the middle of the lists. I think this will make it easier for the user to see what got added.

DonLakeFlyer:
Another possible problem is that this dropdown is now possibly taller than it ever has been. Does it handle max-height sizing and scrolling when running on say something like an iPhone 5S sized screen?

edwinhayes:
Instantaneous current consumption is an important metric on some aircraft in some phases of flight; the current lines take up about as much space as the checkbox, so I'd suggest just having them in by default, and getting rid of the checkbox.

DonLakeFlyer:
> so I'd suggest just having them in by default, and getting rid of the checkbox.

There are more things hidden behind this checkbox than current.

I would suggest making the "More Details" a FactSettings in AppSettings. That would it will persist for people who alway wants that visible.

edwinhayes:
Oh right, I see.  Disregard.  I didn't actually look at the code, just responded to:

> Is live current draw really something that needs to be in the dropdown. Seems like overkill to have a "More Details" for a single field. I don't see how a normal user would need to see that value ever.

DonLakeFlyer:
To summarize my ramblings:
* If a larger font is being used for Battery Status for readability then the other pulldowns then the other pulldowns should be updated as well.
* According to here: https://github.com/PX4/Firmware/pull/12034 SYS_STATUS battery values report the lowest of the two batteries, hence calling the values from here batterySummary only sort of makes sense. But that said I'm not sure what would be better.
* According to here: https://github.com/PX4/Firmware/pull/12034#event-2492405421 parameters for the second battery will be added. Hence PowerComponent page should not be changed to use summary values. It should continue to use battery 1 values since it currently only supports the first battery.
* The "More Details" checkbox should be on an AppSettings such that it is remembered.
* The fields that show up for "More Details" should be added to bottom of each battery list to make it clearer as to what was added.
* The Battery dropdown contents should be in a QGCFLickable and be sized according to available height.
* The "battery" FactGroup should be renamed to "battery1" to match "battery2" naming. And then back compat adjustment should be coded into ValuesWidgetController.cpp.

stefandunca:
> If a larger font is being used for Battery Status for readability then the other pulldowns then the other pulldowns should be updated as well.

That is supposed to be an improvement after reports of low visibility in the field, by regular users, on small android tablets. I don't have enough filed testing experience to take this decision therefore I will revert to regular.

> According to here: PX4/Firmware#12034 (comment) parameters for the second battery will be added. Hence PowerComponent page should not be changed to use summary values. It should continue to use battery 1 values since it currently only supports the first battery.

I've tried to be on the safe side and keep the old behavior. The individual battery voltage message wasn't used therefore I don't have a good understanding of the implications now that I use it. Will it be available on all the firmware versions QGC is connected to?. Ideally this should be based on the `fw` version.

> The "More Details" checkbox should be on an AppSettings such that it is remembered.

I was thinking of it but I was planning it as an improvement after getting some feedback from the users.

> The fields that show up for "More Details" should be added to bottom of each battery list to make it clearer as to what was added.

It makes sense. I tied to keep them closer based on the importance.

> The Battery dropdown contents should be in a QGCFLickable and be sized according to available height.

I haven't thought that will go out of the screen anytime soon.

> The "battery" FactGroup should be renamed to "battery1" to match "battery2" naming. And then back compat adjustment should be coded into ValuesWidgetController.cpp.

DonLakeFlyer:
FYI: SYS_STATUS mavlink discussion is happening here: https://github.com/mavlink/mavlink/pull/1182. It's seems to be tending towards SYS_STATUS not really being a "summary" but we'll see where it ends up.

> The individual battery voltage message wasn't used therefore I don't have a good understanding of the implications now that I use it.

Ah, I missed that part. This is ok for now then until whatever SYS_STATUS ends up doing and/or the additional parameters for battery 2 come online. At some point I think we are going to have to change to pull all Battery 1 fact values from BATTERY_STATUS instead of SYS_STATUS. I'm going to create an Issue for that so it doesn't get lost.

DonLakeFlyer:
Just and FYI for now until it gets locked down but I think SYS_STATUS battery values are heading towards being battery 0 only not a summary: https://github.com/mavlink/mavlink/pull/1182

DonLakeFlyer:
Also dual battery is going to show dual indicators in the toolbar as well.

hamishwillee:
> Also dual battery is going to show dual indicators in the toolbar as well. But I'd wait until the mavlink discussion closes out to make any further changes.

FWIW Dual icons works great for 2 batteries, but you might want to code such that beyond some number of batteries multiple icons are hidden behind a parent icon (there is no limit to the possible number in the spec).

stefandunca:
This was just a quick fix/workaround for the PR that was merged. I will wait for the discussion to settle, understand the specs and changes then do a proper implementation.

DonLakeFlyer:
Whatever happened with all of this?

LorenzMeier:
@DonLakeFlyer Looking at the long back and forth I think what would have been needed would have been a short call to settle the architecture so the contribution can be aligned and brought in. It sounds like that is still missing.

DonLakeFlyer:
I believe this happened in the mavlink calls but no conclusion was come to. Isn’t that correct hamish? Battery_status is still kind on whatever the firmware wants it to be. I’ll take a look at this pull again when I get home.

DonLakeFlyer:
Also I think this is just toolbar changes. Power setup need fixing as well.

stefandunca:
Part of the changes was done in the following PR: https://github.com/mavlink/qgroundcontrol/pull/6474

PR #7576, opened 2019-07-11
WIP: Throttle gimbal joystick

sanderux:
Fix to keep joysticks centered and throttle output

### DIFF
--- a/src/FlightDisplay/VirtualJoystick.qml
+++ b/src/FlightDisplay/VirtualJoystick.qml
@@ -17,10 +17,14 @@ import QGroundControl.Palette       1.0
 import QGroundControl.ScreenTools   1.0
 import QGroundControl.Vehicle       1.0
 
+
+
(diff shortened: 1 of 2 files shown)


src/FlightDisplay/VirtualJoystick.qml
```diff
@@ -69,12 +87,21 @@ Item {
...
+                });
+
             }
         }
         onXAxisChanged: {
             if(_gimbalMode) {
-                setGimbal()
+                    _seq = _seq + 1
+                    var myseq = _seq
+                    delay(100, function() {
```
jafrado: shouldn't the delay be configurable?

sanderux:
This is a temporary fix for a 1 year old PR. It is against another branch for a very specific task. It will not go in the standard system. Having said that, its comparable to the interval in the timers that are also fixed

PR #7583, opened 2019-07-16
Reset Parameters Warning Message

dogmaphobic:
_Reset all to firmware default's_ functionality in QGC resets all the parameters to factory defaults, which includes UAVCAN nodes. In the case of UAVCAN ESCs this means that all the motor parameters are wiped out and the ESCs are no more operational.

I'm adding a warning message alerting the user about this.

Screen Shot 2019-07-16 at 8 29 57 AM


DonLakeFlyer:
So this is sent like this:
```
    _vehicle->sendMavCommand(MAV_COMP_ID_ALL,
                             MAV_CMD_PREFLIGHT_STORAGE,
                             true,  // showError
                             2,     // Reset params to default
                             -1);   // Don't do anything with mission storage
```

Which means all components parameters will reset. Should the wording be changing to something to say that instead and then maybe use UAVCAN as an example to call it out?

dogmaphobic:
> Which means all components parameters will reset. Should the wording be changing to something to say that instead and then maybe use UAVCAN as an example to call it out?

Makes sense.

dogmaphobic:
I've updated the message:
Screen Shot 2019-07-18 at 4 59 35 PM

PR #7595, opened 2019-07-18
Joystick make over

dogmaphobic:
I've done some work with the Joystick code. Primarily adding new actions available for button assignment and cleaning up the UI a bit. This is a first pass. I still have some more work to do, especially related to gimbals. I've added support for handling gimbals using joysticks analog channels but there aren't any MAVLink messages that can handle them yet.

The code wasn't handling buttons properly. Instead of a single action, you would get a stream of actions just like an RC channel works. That doesn't work for "actions" that expect a simple "button down" state. The code now handles just a single "button down" and executes the action assigned to it only once. The original MANUAL_CONTROL message still contains the proper button bitmap as it used to before. I have in my list to add options for button repeats (and frequency) on a per button basis. For now, you get one call per button press.

I've added support to up 64 buttons. The original code had a limit of 16 due to the way the MANUAL_CONTROL message works. Given that the majority of the functionality is handled within QGC, there is no reason to limit to 16. The button bitmap sent through MANUAL_CONTROL is still cropped to 16 bits (buttons).

All internal actions are now handled through signals instead of direct calls. The joystick code runs on its own thread and none of those calls were thread safe.

Added a few more actions that can be assigned to joystick buttons. More to come...

I split the Joystick setup page into 4 sub-panels.

General:
Screen Shot 2019-07-18 at 3 14 19 PM

Button Assignment:
Screen Shot 2019-07-18 at 3 09 24 PM

Calibration:
Screen Shot 2019-07-18 at 3 14 24 PM


hamishwillee:
- Existing problem, but to me Joystick is relatively rare setup - having it as #3 item in the sidebar seems overly high priority. 
- Is the way to enable this unchanged? https://docs.qgroundcontrol.com/en/SetupView/Joystick.html#enabling-px4-joystick-support
- Is RC mode the same as mode 1 etc on an RC controller?
- I haven't otherwise sanity checked this in any way - I'll have a play when you say you're done.

hamishwillee:
What is this 
image

hamishwillee:
Is the calibration guided? ie when you press start it tells you what to do?

dogmaphobic:
>Existing problem, but to me Joystick is relatively rare setup - having it as #3 item in the sidebar seems overly high priority.

All of those settings are "rare" under normal circumstances. In normal, custom builds of QGC, most of these settings are hidden under an "Advanced" mode. For the regular QGC user, invariably a developer, these settings are actually quite often used. There is a certain sequence of events and the list of options somewhat follow that order. @DonLakeFlyer would know better about that.

Also note that I did not change anything fundamentally. The dialog is essentially in the same place showing the same things. I just split them apart into somewhat logical blocks. The bulk of the changes in this PR is the code behind it all. This is what this panel used to look like:

Screen Shot 2019-07-18 at 10 04 12 PM

>Is the way to enable this unchanged?

Yes, it's unchanged. I just moved the calibration to its own sub panel.

>Is RC mode the same as mode 1 etc on an RC controller?

Yes, it determines the functions of the 4 "sticks"

>What is this? I'm guessing it is where you select the buttons, which you then assign in the button page.

It becomes obvious once you run it. It's simply a monitor giving you feedback of the buttons and sticks as you activate them.

DonLakeFlyer:
> * Existing problem, but to me Joystick is relatively rare setup - having it as #3 item in the sidebar seems overly high priority.

This is true, but it only shows up if you actually have a joystick plugged in. Hence if you have a joystick plugged in you are more than likely going to use it and it will need cal.

src/Vehicle/Vehicle.cc
```diff
@@ -4002,6 +4010,53 @@ void Vehicle::gimbalControlValue(double pitch, double yaw)
         MAV_MOUNT_MODE_MAVLINK_TARGETING);   // MAVLink Roll,Pitch,Yaw
 }
 
+void Vehicle::gimbalPitchStep(int direction)
```
FlavioTonelli: I got confused by the naming here.  It is not properly the direction. I can pass to it an integer such as 5 and as a result I get a 5 degree step.
Wouldn't be better to declare it as a bool? something like rotateClockwise?
An other option can be to extend the args like (int stepSize,bool clockwise)
This consideration apply to various function in this class.

src/Vehicle/Vehicle.cc
```diff
@@ -4002,6 +4010,53 @@ void Vehicle::gimbalControlValue(double pitch, double yaw)
...
+{
+    if(_haveGimbalData) {
+        //qDebug() << "Pitch:" << _curGimbalPitch << direction << (_curGimbalPitch + direction);
+        double p = static_cast<double>(_curGimbalPitch + direction);
+        gimbalControlValue(p, static_cast<double>(_curGinmbalYaw));
```
FlavioTonelli: curGinmbalYaw shall be renamed to curGimbalYaw

src/Vehicle/Vehicle.cc
dogmaphobic: This is work in progress. We need new MAVLink messages to handle this. 
Here is the work on this: https://github.com/mavlink/mavlink/pull/1174

With that said, the idea when using "steps" is indicating direction (negative/positive) but also rate (1 to 100). If you want to slowly move one step in the positive direction you would send a "1". If you want to move as fast as it gets to the negative direction, you would use -100.

You have 3 scenarios:
* Single button press: Move one "step" command in one direction or another
* Button press and hold: Stream "step" messages
* Analog dial/knob: Stream current dial position messages

"Steps" are for now arbitrary. Yes, the logical mapping for a "step" for gimbal is "angle" but is it 1°? 0.1°? 10°? Hence the "step".

src/Vehicle/Vehicle.cc
dogmaphobic: Yup! :P

dogmaphobic:
I'm also adding a "repeat" option to buttons (along with a repeat rate). These are for actions that allow it. When sending a high rate of repeats, it makes no sense to send MAVLink commands. Hence the need for messages to handle these.

Screen Shot 2019-07-25 at 1 48 06 PM

DonLakeFlyer:
Wouldn't it be better to have start/stop type commands with a rate? Button down - start gimbal down at certain rate. Then button up - stop gimbal down. For example?

src/Vehicle/Vehicle.cc
FlavioTonelli: >With that said, the idea when using "steps" is indicating direction (negative/positive) but also rate (1 to 100). If you want to slowly move one step in the positive direction you would send a "1". If you want to move as fast as it gets to the negative direction, you would use -100.
Yes the logic make sense. On this point I was just concerning about the naming. "Direction" seems to me a variable that can only assume +1 or -1 as values.

dogmaphobic:
There is one already for zooming. The action is "Continuous Zoom". I'm not aware of a gimbal that supports this but I will add the same set of commands/UI for it as well. Then we need to handle the rate. Some setting somewhere. Simple when you have one device you know all the options. Not so easy when working with arbitrary things that may or may not do what you want.

DonLakeFlyer:
Hate it when reality gets in the way!

src/VehicleSetup/JoystickConfigController.cc
```diff
@@ -731,31 +720,50 @@ bool JoystickConfigController::throttleAxisReversed(void)
...
-    
+    emit gimbalPitchAxisMappedChanged(gimbalPitchAxisMapped());
+    emit gimbalYawAxisMappedChanged(gimbalYawAxisMapped());
+
     emit rollAxisReversedChanged(rollAxisReversed());
     emit pitchAxisReversedChanged(pitchAxisReversed());
     emit yawAxisReversedChanged(yawAxisReversed());
     emit throttleAxisReversedChanged(throttleAxisReversed());
+    emit gimbalPitchAxisReversedChanged(pitchAxisReversed());
+    emit gimbalYawAxisReversedChanged(yawAxisReversed());
```
FlavioTonelli: It seems wrong to me. It should be gimbalYawAxisReversed().

denna10:
I installed QGC on my PC yesterday (v3.5.4), but the joystick setup interface still looks the same way as @dogmaphobic demonstrates above under "This is what this panel used to look like:". Is there something I'm doing wrong? I was interested in setting the joystick for control of the gimbal.

dogmaphobic:
> Is there something I'm doing wrong? I was interested in setting the joystick for control of the gimbal.

These changes are in master. 3.5.4 is a stable build. You need to run a daily for these new things (knowing that it is development build): 

https://docs.qgroundcontrol.com/en/releases/daily_builds.html

denna10:
> > Is there something I'm doing wrong? I was interested in setting the joystick for control of the gimbal.
> 
> These changes are in master. 3.5.4 is a stable build. You need to run a daily for these new things (knowing that it is development build):
> 
> https://docs.qgroundcontrol.com/en/releases/daily_builds.html

Thanks for the quick, clear and helpful reply! If it is not to much trouble, may I ask, if I would like to suggest support for two joysticks (one for pilot and the other for camera operator) as input, should I open a new issue or are there other ways to suggest new features?

dogmaphobic:
> may I ask, if I would like to suggest support for two joysticks (one for pilot and the other for camera operator) as input, should I open a new issue or are there other ways to suggest new features?

I've been wanting this myself. Not only for two operators as you describe but also to be able to have additional buttons and/or analog channels/sliders coming from a second (or more) "joystick". It would be a major undertaking however, given how the existing architecture is laid out.

denna10:
> > may I ask, if I would like to suggest support for two joysticks (one for pilot and the other for camera operator) as input, should I open a new issue or are there other ways to suggest new features?
> 
> I've been wanting this myself. Not only for two operators as you describe but also to be able to have additional buttons and/or analog channels/sliders coming from a second (or more) "joystick". It would be a major undertaking however, given how the existing architecture is laid out.

I see. Yes, flying with a nice joystick/yoke, throttle and rudder, for example, could be quite the experience. Thanks again!

PR #7609, opened 2019-07-23
CentOS Build Instructions

dogmaphobic:


hamishwillee:
@dogmaphobic It is excellent that you have created this doc, but is the root of the qgc tree the right place?

I would have though alongside the other docs on this topic: https://dev.qgroundcontrol.com/en/getting_started/#build-qgroundcontrol

dogmaphobic:
I always expect to find build instructions in the root directory of a project. That's how it's always been done. Yes, we could certainly have an expanded version within the proper docs but they would be in addition to these.

hamishwillee:
Entirely up to you and Don. I've added link to these instructions in the devguide: https://github.com/mavlink/qgc-dev-guide/pull/68

DonLakeFlyer:
We don't have the the normal build instructions in the root of the repo any more. So I wouldn't put these there either.

hamishwillee:
@dogmaphobic @dogmaphobic Do you want me to just pull these into the dev guide properly (and delete from root). Either way, agree together and let me know.

dogmaphobic:
If it makes it consistent, sure. It's just "normal" to have build instructions within the root directory of a project.

hamishwillee:
> If it makes it consistent, sure. It's just "normal" to have build instructions within the root directory of a project.

It does make it consistent. "Normal" is relative - my experience is that it is normal to have build instructions in or linked from the readme.md, and everything else is up for grabs. Given how many files you have in the root of qgc repo, good to keep things uncluttered.

DonLakeFlyer:
I'd pull into dev guide

PR #7610, opened 2019-07-23
Fix AirMap

jzaturensky:
Fixes issue with AirMap plugin flight planning features not showing up


dogmaphobic:
That looks right. The only issue is the lack of `#if defined(QGC_AIRMAP_ENABLED)` around the new code to prevent it from being compiled when Airmap support is not enabled.

jzaturensky:
@dogmaphobic Any other changes?

dogmaphobic:
>@dogmaphobic Any other changes?

Sorry, I've been buried this past few days...

PR #7611, opened 2019-07-23
Fix logs Download on Android

stefandunca:
Downloading the firmware logs doesn't work on Android. This is a quick fix.
`QGCFileDialog.qml`: dumps the downloaded files in the  `Logs` sub-folder of application's default data path. Optionally the current selection can go in a custom subfolder in the same `Logs` folder for organizational purposes (like labeling).
`LogDownloadController.cc`: create the new subfolder if folder path if missing
`LogDownloadPage.qml`: fix theme colors making text visible in certain theme configurations

### DIFF
--- a/src/AnalyzeView/LogDownloadController.cc
+++ b/src/AnalyzeView/LogDownloadController.cc
@@ -604,6 +604,7 @@ LogDownloadController::_prepareLogDownload()
     } else {
         _downloadData->filename += ".bin";
     }
+
     _downloadData->file.setFileName(_downloadPath + _downloadData->filename);
     //-- Append a number to the end if the filename already exists
     if (_downloadData->file.exists()){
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
I'm kind of concerned about allow folder select on mobile devices. The file systems across android/ios and so forth are funky. Why now just make Analyze download automatically to the default save path on mobile? That would match the other file save usage for things like missions and so forth.

stefandunca:
All folder creation are done in the default application's logs save path as expected. If user doesn't type anything for the folder name it will just download all selected files in the `Logs`. I didn't make any improvements I've just fixed the workflow based on the current implementation that is used in other parts as well.

DonLakeFlyer:
> All folder creation is done in the default application's logs save path as expected. If user doesn't type anything for the folder name it will just download all selected files in the `Logs`.

It still makes this implementation inconsistent with all other places where you save files (Plan Files, Parameters, ...) on Mobile. I want to keep file handling consistent across all of Mobile. Which is to always just go to the default location without any ability to change that. 

I also don't see much value in allowed new folder creation within the limited directory tree. Hence I don't see changing Mobile to follow this model everywhere either.

dogmaphobic:
@stefandunca What is the conclusion here? This is out of sync with the custom build. If this cannot go to main QGC, we need to sort out a way to allow a plugin to change the behavior on its own. It can't stay out of sync (files outside custom cannot be different).

dogmaphobic:
@stefandunca Is this still relevant?

stefandunca:
Was a quick hack at pilot's request. I drop it until somebody misses it.

Issue #7586, opened 2019-07-17
Feature Proposal - Home Heading / COG Angle / Next WP Heading in Compass


tilaktilak:
We worked on some indicators in the compass widget to show :
- Home Heading
- COG Angle (ie. The GPS Heading from trajectory)
- Heading of the next Waypoint

We also added the ability to lock the North up (The red vehicle arrow locked up).

From our first implementation it looks like this :
![Untitled2](https://user-images.githubusercontent.com/6662416/61343206-165e7180-a8a1-11e9-89d3-44e7cf47f6fa.png)

Settings Page :
![image](https://user-images.githubusercontent.com/6662416/61345499-87a22280-a8a9-11e9-95ef-549425452821.png)

Here the current implementation : https://github.com/mavlink/qgroundcontrol/compare/master...Aeronavics:compass_home_n_COG

Wanted to have some feedback if it worth integrating those features in the mainline.

DonLakeFlyer:
All seems valuable to me. I would change the settings text to be just "Show" from "Show/Hide".

@antiheavy Opinions?

edwinhayes:
If people think this is too busy, we could condense the three show/hide checkboxes into one?  If someone wants COG but not home, they can rebuild themselves? (But personally, I don't mind long lists of configuration options).

Also, suggest changing "Lock North up on Compass" to "Lock Compass 'North-Up'".  Oh, hang on, the usual default behaviour is for North-Up, so the checkbox should be "Lock Compass 'Nose-Up'" so that  having all the boxes unchecked delivers the existing 'normal' QGC experience?

edwinhayes:
Oh, also remove the word "angle".  And change "Show Home heading to Compass" to "Show Heading to Home on Compass" so that it's consistent with "Show Heading to Next Waypoint on Compass".

Antiheavy:
I like this.  +1 vote from me.

Antiheavy:
> All seems valuable to me. I would change the settings text to be just "Show" from "Show/Hide".

correct.  checkbox options like that should just be "Show"

Antiheavy:
>  we could condense the three show/hide checkboxes into one?

seems okay.  maybe "Show extra heading indicators on compass"

How does a new user know what indicator means what thing?

edwinhayes:
> How does a new user know what indicator means what thing?

 @tilaktilak can update the documentation page online to explain?

tilaktilak:
[...]

PR #7613, opened 2019-07-24
Feature Proposal - Home Heading / COG Angle / Next WP Heading in Compass

tilaktilak:
**As discussed in #7586, here the correspoding P.R.**

We worked on some indicators in the compass widget to show :

- Home Heading
- COG Angle (ie. The GPS Heading from trajectory)
- Heading of the next Waypoint

We also added the ability to lock the North up (The red vehicle arrow locked up).

From our first implementation it looks like this :
image

Settings Page :
image

Here the current implementation : master...Aeronavics:compass_home_n_COG

Wanted to have some feedback if it worth integrating those features in the mainline.


DonLakeFlyer:
* The screen shot for settings page is old right? There is just a single setting.
* I'd like to make heading to home always shown,  not controlled by setting

You should test multi-vehicle mode where the compass indicator is shown as a tiny little indicator. It may be required to turn all of these off in that mode no matter what setting.

tilaktilak:
- That's true, let me update the screenshots, in fact there is also the "Lock North up" setting :
image

- I Agree for the Home heading could be relevant to have it at any time

- Is it possible to test it on simulation ? I'll probably need to test it to insure my changes works.

DonLakeFlyer:
> * Is it possible to test it on simulation ? I'll probably need to test it to insure my changes works.

Yes. You can start up multiple Mock Links.

DonLakeFlyer:
Did something go bonkers here?

edwinhayes:
Ah, yes... it does seem to have.  This is gonna have been my fault. :-(

DonLakeFlyer:
I think the original pull was merged wasn't it?

tilaktilak:
I think it's fine now, we've made some changes in our repos yesterday. Sorry about that :-)

BTW, I have made some tests on the "Multi-Vehicle" yesterday, it appear that the Single/Multi vehicle view selector was overlayed by the Instrument panel, not clear that it's a side effect of my changes on the instrument panel, my fix :    
Add a top anchor for FlightDisplayViewWidgets if the SingleMultiSelector is visible : https://github.com/mavlink/qgroundcontrol/pull/7613/commits/aa6fb469ec1e2d57469e42ef66c9f2a7aa88f2d2#diff-202f6dde27553787cd006134f715b64fR522 

I will disable the indicators in the multi-vehicle panel then

tilaktilak:
So now it's hidden in the Multi-Vehicle panel, I also fixed some anchors on Multi-vehicle view.
@DonLakeFlyer can you have a look ?

PR #7631, opened 2019-07-28
Improve CentOS instructions

MaEtUgR:
Thanks to @dogmaphobic for providing these useful steps to get QGC working under CentOS. I changed/added all the main points I found out while setting it all up in a VM, and a toughpad in a UAV Components Micronav device. This is supposed to save other people a lot of time since CentOS behaves quite different compared to Ubuntu.

### DIFF
--- a/CentOS.md
+++ b/CentOS.md
@@ -1,40 +1,99 @@
-## Building QGC on a stock CentOS:
+# Running QGC on CentOS 7
 
+# OS Installation
+- Fetch the latest CentOS 7 iso from here: http://isoredirect.centos.org/centos/7/isos/x86_64/CentOS-7-x86_64-DVD-1810.iso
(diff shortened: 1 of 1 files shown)


CentOS.md
```diff
@@ -61,19 +120,13 @@ For a debug/test build:
 ```
 For a release build:
 ```
-../Qt/5.12.4/gcc_64/bin/qmake ../qgroundcontrol/qgroundcontrol.pro spec linux-g++ CONFIG+=qtquickcompiler
+../Qt/5.12.4/gcc_64/bin/qmake ../qgroundcontrol/qgroundcontrol.pro -spec linux-g++ CONFIG+=qtquickcompiler
```
MaEtUgR: @dogmaphobic I was assuming there needs to be a dash here since the debug command contains it and I was successfully using the command iwth the dash for both builds.

dogmaphobic: Yes, the missing dash is a typo.

PR #7646, opened 2019-07-30
ArduCopter: Add Follow, Flow Hold, ZigZag flight modes

DonLakeFlyer:
Also bumped to max limit for Go To Location to 10k.


Antiheavy:
10 kilometers!  Hmmm....  wonder if we will want that user or plugin setable....

DonLakeFlyer:
I'll change GoTo Location max to be settable

PR #7649, opened 2019-07-31
Std ComboBox

dogmaphobic:
This replaces the old QGCComboBox with a simpler, current QtQuick 2.0 version. I'm sure there was a reason at some point for that code to be so convoluted but it turns out it breaks apart when you add it to a Popup window. It causes the focus to get stuck and you can't touch anything afterwards. Instead of fighting it, I just replaced with a current ComboBox, which seems to work just fine.


DonLakeFlyer:
> I'm sure there was a reason at some point for that code to be so convoluted

Because the built in one would put the dropdown in the wrong place on Android.

PR #7654, opened 2019-08-01
Update QGC menus to match the application's theme

stefandunca:
Drop the native looking menus and customize the menus  based on theme including scaling factor.
Change the base QGCMenu and QGCMenuItem components


stefandunca:
@yasendimov if you can take over and check why the menu is not adjusting based on size. Thanks

yasendimov:
image

PR #7666, opened 2019-08-05
Disable preflight checklist if vehicle is armed.

dogmaphobic:
Title says it all.


hamishwillee:
@dogmaphobic Does this also hide the checklist if it currently displayed when the vehicle is armed?

dogmaphobic:
In theory, you can't arm the vehicle while the checklist is up. At least not from the UI.

PR #7667, opened 2019-08-06
Fix MSVC 2015 "fatal error C1060: compiler is out of heap space"

samsalway:
I followed the Getting Started page, I've got Qt 5.11.3 with MSVC 2015 32 bit on Windows 8.1 Pro, and I got this error.  The solution was simple.


dogmaphobic:
Perfect. Thanks!

PR #7674, opened 2019-08-11
Global: change null or 0 to nullptr

muramura:
I think nullptr is better than specifying NULL pointer as NULL or 0 for C ++ language.

I saw a warning displayed in the QT tool where NULL or 0 was set in the NULL pointer setting.
I think it's easier to spot real problems if you deal with tool warnings.


DonLakeFlyer:
@dogmaphobic Can you think of a case where a NULL to nullptr change will break something?

dogmaphobic:
No, it's all good.

tcanabrava:
thanks, this fixes so many warnings in my system. can this get in?

DonLakeFlyer:
I've been trying to get the android build to run.

DonLakeFlyer:
@dagar Do you have any ideas as to what is going on in travis android builds. It can't seem to see github.com

dagar:
image

I'm not sure what's going on, but I'm tempted to properly move everything (except windows) to Jenkins where we can drop a lot of this cruft we've strung together for build environment setup, packaging, and deployment.

DonLakeFlyer:
I'm fine with moving to cmake and Jenkins as long as everything in current qmake/travis/appveyor side is supported. A good test case for that is Sentera custom build. Or I guess the custom build thats in the repo as well. There is all sorts of override crap in qmake .pro for custom builds.

DonLakeFlyer:
My main concern with Jenkins is the fact that the build server is not in the cloud right? It's just a machine Gus has. Which means, if Gus does on vacation and it craps out we are toast.

dagar:
> I'm fine with moving to cmake and Jenkins as long as everything in current qmake/travis/appveyor side is supported. A good test case for that is Sentera custom build. Or I guess the custom build thats in the repo as well. There is all sorts of override crap in qmake .pro for custom builds.

They aren't necessarily related, I just happened to only add the cmake builds to Jenkins because we have orders of magnitude more capacity there. It will build both until everything is 100% in the cmake world (which probably means waiting for Qt to improve cmake integration for Android and iOS).

image

dagar:
> My main concern with Jenkins is the fact that the build server is not in the cloud right? It's just a machine Gus has. Which means, if Gus does on vacation and it craps out we are toast.

All the linux resources are in the cloud (dynamically provisioned ec2 slaves). For MacOS we have a hosted mac mini (macstadium.com) in addition to the stupid fast machine Gus donates.

DonLakeFlyer:
Ah, I see. Again I'm fine with moving to whatever as long as it works fully.

DonLakeFlyer:
I fixed the conflict. If it all passes again. I'll merge it without the Android passing.

DonLakeFlyer:
@muramura Thanks for the pull

muramura:
@DonLakeFlyer san.. 
Merged, Thank's.

PR #7678, opened 2019-08-13
custom-example: Add GPS info to the Vehicle Indicator

stefandunca:
Add one row at the top of the Custom Vehicle Indicator control to show: lat, lon and hdop

Screenshot from 2019-08-13 11-12-10

### DIFF
--- a/custom-example/res/CustomFlyView.qml
+++ b/custom-example/res/CustomFlyView.qml
@@ -266,68 +266,59 @@ Item {
             rowSpacing:             ScreenTools.defaultFontPixelHeight * 0.5
             columns:                7
             anchors.centerIn:       parent
-        //-- Chronometer
-            QGCColoredImage {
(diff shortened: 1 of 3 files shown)


custom-example/res/CustomFlyView.qml
```diff
@@ -266,68 +266,59 @@ Item {
...
-                sourceSize.height:      height
                 Layout.alignment:       Qt.AlignVCenter | Qt.AlignHCenter
                 color:                  qgcPal.text
-        }
-        QGCLabel {
-            text: {
-                    if(activeVehicle)
-                        return secondsToHHMMSS(activeVehicle.getFact("flightTime").value)
-                return "00:00:00"
+                text:                   "lat:"
```
dogmaphobic: I would capitalize it:

_Lat:_
_Lon:_

PR #7680, opened 2019-08-14
Object avoidance

dogmaphobic:
Initial support for object avoidance.
The base message processing is within the vehicle class. The (custom) UI for now only shows dots representing obstacles. A new UI is forthcoming.

Screen Shot 2019-08-14 at 5 20 16 PM

Screen Shot 2019-08-14 at 5 26 39 PM


dogmaphobic:
Added code to properly handle units within the distance slider.

Screen Shot 2019-08-14 at 6 16 17 PM

hamishwillee:
What is "minimum distance"? - the distance below which you will no longer get notification?

How is the max range of slider populated?

dogmaphobic:
This is just the bare minimum for developers to start testing. It is not meant for normal consumption. The UI is not even present on normal builds.

hamishwillee:
@dogmaphobic Sure. Note that I may well not look at this again. If this goes into production these things will still not make sense, but "not my fault guv"

dogmaphobic:
What I meant is that the UI for it, along with the copy, has yet to be designed/done. @yasendimov is working on the UI/UX.

dogmaphobic:
For what's worth, this is how it is documented:

[_Minimum distance the vehicle should keep to all obstacles_](https://docs.px4.io/v1.9.0/en/advanced_config/parameter_reference.html)

_Comment: Only used in Position mode. Collision avoidace is disabled by setting this parameter to a negative value_

The range comes from the parameter metadata.

dogmaphobic:
It shows a range from 1 to 15 (meters). Negative values to disable it. 0 is not used apparently.

hamishwillee:
Thanks. That was my guess (I wrote the collision docs). FWIW I don't know why minimum distance feels a bit off to me - it is accurate. 

I just wish I could come up with something better than "Closest approach", "Buffer zone size". Ignore me.

Issue #2829, opened 2016-02-17
Critical errors not displayed aggressively enough for the user, leads to wrong focus on QGC errors


LorenzMeier:
<img width="1392" alt="screen shot 2016-02-17 at 16 13 58" src="https://cloud.githubusercontent.com/assets/1208119/13113933/87710176-d591-11e5-8fe8-226b2b17b01b.png">

LorenzMeier:
And actually this a usability issue: The real error is shown in a critical message from the system. The critical messages should probably be shown all in the lower message pane:

<img width="1392" alt="screen shot 2016-02-17 at 16 15 10" src="https://cloud.githubusercontent.com/assets/1208119/13113968/b6c69008-d591-11e5-9515-6a9a9a7a833c.png">

DonLakeFlyer:
@dogmaphobic Shouldn'y all critical errors be showing up down there? I have noticed sometimes they don't pop. Not sure why.

dogmaphobic:
I can't find anywhere in the code base where these *critical* messages are ever sent to that message box. There is one handler within Vehicle.cc that receives it but all it does is to send its own signal, which no one listens to:

https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L928

The only messages that are shown in that lower message area are those that are explicitly called. That is, only fully formatted messages are ever sent there. For example, things like these:

https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L466-L478

In other words, critical messages coming from vehicles are not currently been sent there at all. I will add code to the signal above (Vehicle.cc#L928) and have the critical messages sent there.

dogmaphobic:
Disregard the above. MainWindowInner has a handler for `formatedMessage` (Vehicle.cc#L928 above). Looking...

dogmaphobic:
Done in #3241. It was a side effect of some changes Don had done, I assume to prevent the message handler from sending messages before the main window was ready to receive them.

DonLakeFlyer:
> I assume to prevent the message handler from sending messages before the main window was ready to receive them.

I vaguely remember something like that

PR #7682, opened 2019-08-14
Turn on popup errors for mobile builds

DonLakeFlyer:
Mobile builds were not showing pop up errors.

Caused by sequence of changes sort of doc'ed here: https://github.com/mavlink/qgroundcontrol/issues/2829. 

I believe there was an initial problem with error messages popping up before the main window was created either causing crashes or a strange race condition on main window creation. There was a fix for that which went in which screwed thing up. Then a fix for that screw up went in which only worked on desktop builds.

Mobile builds don't really have the same race condition crash problems. But even so I changed the code to match enabling popup errors to only display after the main qml window was created.


DonLakeFlyer:
@dogmaphobic Can you just sanity check my thinking/code here?

PR #7693, opened 2019-08-16
CMake: Set warning level and disable -Waddress-of-packet-member

tcanabrava:
Mavlink fails terribly on the check for -Waddress-of-packet-member
but this is not a problem of QGC, if I don't remove this warning
it's impossible to look at the terminal for QGC warnings.
This needs to be fixed however, but this is not the place.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,12 @@ set(CMAKE_AUTOUIC ON)
 set(CMAKE_AUTORCC ON)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -10,6 +10,12 @@ set(CMAKE_AUTOUIC ON)
 set(CMAKE_AUTORCC ON)
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
```
DonLakeFlyer: There is no "or" logical capability in cmake?

dagar:
Rebase to get it through CI?

CMakeLists.txt
dagar: add_compile_options?

tcanabrava: @DonLakeFlyer there's an or operator but for some options clang and gcc differ, so I'd prefer to keep them separated even if they add the same options for now. we don't know about the future. :)

@dagar I'll update this using add_compile_options

PR #7694, opened 2019-08-16
Add qml files into custom_targets

tcanabrava:
When we use CMake as a project manager - to be less restrictive about what kind of IDE the developer wants to use, we should add the Qml files into custom_targets so the IDE can display the Qml files in hierarchy. Tested in QtCreator


dagar:
Rebase to get it through CI?

PR #7701, opened 2019-08-18
Change goto message to set_position for ArduCopter

tilaktilak:
**Goto Location does not work anymore on ArduCopter**

The correct message as sent by MissionPlanner is SET_POSITION_GLOBAL

**To Reproduce**
1. Drone flying a mission
2. Click on map to send a Goto Location
3. Error + Drone does not fly GotoLocation

**Explanation**
The current message sent in this case is a MissionItem to set a new waypoint, Arducopter now need a SET_POSITION_TARGET_GLOBAL_INT.
This branch implement a ArduCopter specific version of guidedModeGotoLocation to send the correct message.

Tested and works with the last version of Ardupilot simulator.

### DIFF
--- a/src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.cc
@@ -136,3 +136,44 @@ bool ArduCopterFirmwarePlugin::vehicleYawsToNextWaypointInMission(const Vehicle*
     return true;
 }
 
+void ArduCopterFirmwarePlugin::guidedModeGotoLocation(Vehicle* vehicle, const QGeoCoordinate& gotoCoord)
+{
+    if (qIsNaN(vehicle->altitudeRelative()->rawValue().toDouble())) {
+        qgcApp()->showMessage(QStringLiteral("Unable to go to location, vehicle position not known."));
+        return;
+    }
+
+    setGuidedMode(vehicle, true);
+
+    if (vehicle->capabilityBits() & MAV_PROTOCOL_CAPABILITY_COMMAND_INT) {
+
+        mavlink_message_t msg;
+        mavlink_set_position_target_global_int_t cmd;
+
+        memset(&cmd, 0, sizeof(cmd));
+
+        cmd.target_system    = static_cast<uint8_t>(vehicle->id());
+        cmd.target_component = static_cast<uint8_t>(vehicle->defaultComponentId());
+        cmd.coordinate_frame = MAV_FRAME_GLOBAL_RELATIVE_ALT;
+        cmd.type_mask = 0x0DF8; // Position Only
+        cmd.lat_int = static_cast<int>(gotoCoord.latitude()*1e7);
+        cmd.lon_int = static_cast<int>(gotoCoord.longitude()*1e7);
+        cmd.alt = vehicle->altitudeRelative()->rawValue().toFloat();
+
+        MAVLinkProtocol* mavlink = qgcApp()->toolbox()->mavlinkProtocol();
+
+        mavlink_msg_set_position_target_global_int_encode_chan(
+                static_cast<uint8_t>(mavlink->getSystemId()),
+                static_cast<uint8_t>(mavlink->getComponentId()),
+                vehicle->priorityLink()->mavlinkChannel(),
+                &msg,
+                &cmd);
+
+        vehicle->sendMessageOnLink(vehicle->priorityLink(), msg);
+    }
+    else{
+        qgcApp()->showMessage(QStringLiteral("Unable to go to location, vehicle cannot receive INT command."));
+        return;
+    }
+
+}
--- a/src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.h
+++ b/src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.h
@@ -59,6 +59,7 @@ public:
 
     // Overrides from FirmwarePlugin
     void    guidedModeLand                      (Vehicle* vehicle) final;
+    void    guidedModeGotoLocation          (Vehicle* vehicle, const QGeoCoordinate& gotoCoord) override;
     const FirmwarePlugin::remapParamNameMajorVersionMap_t& paramNameRemapMajorVersionMap(void) const final { return _remapParamName; }
     int     remapParamNameHigestMinorVersionNumber(int majorVersionNumber) const final;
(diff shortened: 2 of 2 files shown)


edwinhayes:
@tilaktilak - Per our discussion, probably want to account for resending packets which don't make it through.

src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.cc
```diff
@@ -136,3 +136,44 @@ bool ArduCopterFirmwarePlugin::vehicleYawsToNextWaypointInMission(const Vehicle*
...
+void ArduCopterFirmwarePlugin::guidedModeGotoLocation(Vehicle* vehicle, const QGeoCoordinate& gotoCoord)
+{
+    if (qIsNaN(vehicle->altitudeRelative()->rawValue().toDouble())) {
+        qgcApp()->showMessage(QStringLiteral("Unable to go to location, vehicle position not known."));
+        return;
+    }
+
+    setGuidedMode(vehicle, true);
+
+    if (vehicle->capabilityBits() & MAV_PROTOCOL_CAPABILITY_COMMAND_INT) {
```
DonLakeFlyer: This only applies to sending COMMAND_INT (https://mavlink.io/en/messages/common.html#COMMAND_INT) mavlink messages. It doesn't really apply to other messages are far as I know. At least not accordint to the mavlink spec.

src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.h
```diff
@@ -59,6 +59,7 @@ class ArduCopterFirmwarePlugin : public APMFirmwarePlugin
 
     // Overrides from FirmwarePlugin
     void    guidedModeLand                      (Vehicle* vehicle) final;
+    void    guidedModeGotoLocation          (Vehicle* vehicle, const QGeoCoordinate& gotoCoord) override;
```
DonLakeFlyer: In reality it's a bug that all of these aren't marked override like this new one. '''final''' prevents custom builds from overriding the QGC provided implementations which should be allowed.

src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.cc
DonLakeFlyer: @hamishwillee Not sure if this is another point of confusion/discussion in the spec or not?

src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.cc
```diff
@@ -136,3 +136,44 @@ bool ArduCopterFirmwarePlugin::vehicleYawsToNextWaypointInMission(const Vehicle*
...
+
+        mavlink_message_t msg;
+        mavlink_set_position_target_global_int_t cmd;
```
edwinhayes: Per discussion about MAVLink confusion; maybe rename this, since it isn't a Command that you're sending?

src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.cc
```diff
@@ -136,3 +136,44 @@ bool ArduCopterFirmwarePlugin::vehicleYawsToNextWaypointInMission(const Vehicle*
...
+
+        memset(&cmd, 0, sizeof(cmd));
+
+        cmd.target_system    = static_cast<uint8_t>(vehicle->id());
+        cmd.target_component = static_cast<uint8_t>(vehicle->defaultComponentId());
+        cmd.coordinate_frame = MAV_FRAME_GLOBAL_RELATIVE_ALT;
+        cmd.type_mask = 0x0DF8; // Position Only
+        cmd.lat_int = static_cast<int>(gotoCoord.latitude()*1e7);
+        cmd.lon_int = static_cast<int>(gotoCoord.longitude()*1e7);
+        cmd.alt = vehicle->altitudeRelative()->rawValue().toFloat();
```
edwinhayes: This isn't really related to this implementation in particular: can't QGeoCoordinate store either a 2D or 3D position?  If this method is called with a 3D position, should this really be the behaviour?  It seems pretty non-intuitive.

DonLakeFlyer: The coord passed in does not have an altitude in it. Just lat/lon is provided.

edwinhayes: Is that just a function of the fact that right now, the only way of setting a guided mode target is by clicking on the map though?  What if I wanted to implement some other UI element which provided a 3D coordinate?

DonLakeFlyer: > What if I wanted to implement some other UI element which provided a 3D coordinate?

It's a function of the current ui interface (which doesn't allow altitude change on goto) as well as what is expected from the FirmwarePlugin method. They were implemented at the same time. Although totally possible to add from ui and interface side and folks have been asking for altitude change in combination with goto location anyway.

edwinhayes: So, it might be ok then to change the implementations such that if the coordinate supplied is 3D, it sets the altitude as requested, but if the coordinate is just 2D, it uses the current altitude (as is the present behaviour)?

DonLakeFlyer: Sure but that will require the corresponding ui for it as well. I would put that in a separate pull.

DonLakeFlyer: @tilaktilak To implement that you can look at how the Guided Pause command always shows the altitude slider and pulls values from it.

tilaktilak:
By doing some tests today I realize that Goto location is working as is for an ArduPilot simulator tagged 3.6.9 or 3.6.10 but not for master. So it might be a regression on ArduPilot side, let me check because if it's the case this P.R. may be closed.

tilaktilak:
So I tested several branches of ArduPilot :

- master : Not working : 
image

- Branch Copter-3.6 : Working
- Tag 3.6.10 and 3.6.9 : Working

DonLakeFlyer:
Does regular mission upload work?

tilaktilak:
Yep, I think it only occurs in Guided Mode. I'll open an issue in ArduPilot to see what's going on.

PR #7703, opened 2019-08-19
Remove dynamic cast

tcanabrava:
Remove dynaic casts thru the software for objects that inherti from QObject.
qobject_cast does not use RTTI and thus is much faster. Also, use `auto` keyword to not repeat class names when the class name is explicit.

### DIFF
--- a/custom-example/src/CustomPlugin.cc
+++ b/custom-example/src/CustomPlugin.cc
@@ -47,7 +47,7 @@ customQuickInterfaceSingletonFactory(QQmlEngine*, QJSEngine*)
 {
     qCDebug(CustomLog) << "Creating CustomQuickInterface instance";
     CustomQuickInterface* pIFace = new CustomQuickInterface();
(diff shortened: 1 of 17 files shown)


DonLakeFlyer:
I'm continue to be way less than excited about using auto anywhere. To me it just obfuscates things. Your specific usage case here is certainly less than evil to me. But it makes me look in what I would think to be a non-standard place to figure out the type. I think it makes code less readable "at a glance" which is not necessarily a good thing.

tcanabrava:
Don, I believe it’s more readable this way - but I would never use auto
when we don’t have a type formally written nearby.
Sometimes the type is too big specially when we use together with dynamic
cast or with iterators.
So I would like to keep *this* kind of auto usage that I added here, not
the others (that I don’t use anyway)
Feels good enough?



Em seg, 19 de ago de 2019 às 22:47, Don Gagne <notifications@github.com>
escreveu:

DonLakeFlyer:
I'm fine with this specific use case. But I don't want to see it bleed out into other things. It's hard to police "only use auto when it doesn't really suck". Especially when it pretty much sucks 99% of the time.

PR #7707, opened 2019-08-20
De escalate method

tcanabrava:
Trying to improve the code quality of some methods.

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -393,25 +393,15 @@ void LinkManager::loadLinkConfigurationList()
         // Find out how many configurations we have
         int count = settings.value(LinkConfiguration::settingsRoot() + "/count").toInt();
         for(int i = 0; i < count; i++) {
-            QString root = LinkConfiguration::settingsRoot() + QStringLiteral("/Link%1").arg(i); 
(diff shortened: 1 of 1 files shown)


tcanabrava:
Because it’s not used by anyone besides the private part of the class. This
makes recompilation faster as if I change the signature of this function I
don’t touch the header and thus I don’t recompile everything that touches
it.

I know that this is not how qgc handles private functions and I put on
purpose to open for questioning: should we move private implementations of
methods that *dont* rely on members of the class outside of the header?

I

Em seg, 26 de ago de 2019 às 20:16, Don Gagne <notifications@github.com>
escreveu:

tcanabrava:
That was a quick answer, if you have time please take a look at this cppcon
talk

https://youtu.be/WLDT1lDOsb4

If you want I rework this patch to have the method as a private
implementation. But really consider having non dependent functions as
separated entities. Link time is already too big for qgc.


Em seg, 26 de ago de 2019 às 22:40, Don Gagne <notifications@github.com>
escreveu:

tcanabrava:
@DonLakeFlyer fair enough. There's also a second part of this that I personally like, that you don't have if the code is in a class: The compiler will warn if you are not using a function in an unamed namespace = easier to remove dead code. I'v found out a few files that where compiled but never used in QGC by pure luck.

Still, I'll rework this to be part of the class - or at least a lambda.

PR #7714, opened 2019-08-21
Add oucoming messages in tlog

tilaktilak:
It appear really useful to have both messages sent and received in the telemetry log from QGC :
- To have the complete story of which order trig which response from the drone
- To analyze all messages sent/received
- To retrieve all orders sent while replaying a mission

Here the changes I propose : 

- Add a signal emitted by writeBytes of each LinkInterface (TCP/UDP for the moment)
- Connect this signal to MAVLinkProtocol::logSentBytes in LinkManager
- MAVLinkProtocol::logSentBytes will log messages sent from
  QGC

### DIFF
--- a/src/comm/MAVLinkProtocol.cc
+++ b/src/comm/MAVLinkProtocol.cc
@@ -160,6 +160,52 @@ void MAVLinkProtocol::resetMetadataForLink(LinkInterface *link)
 }
 
 /**
+ * This method parses all outcoming bytes and log a MAVLink packet.
+ * @param link The interface to read from
+ * @see LinkInterface
+ **/
+
+void MAVLinkProtocol::logSentBytes(LinkInterface* link, QByteArray b){
+
+    uint8_t mavlinkChannel = link->mavlinkChannel();
+    static mavlink_message_t _sent_message;
+
+    for (int position = 0; position < b.size(); position++) {
+
+        if(mavlink_parse_char(mavlinkChannel, static_cast<uint8_t>(b[position]), &_sent_message, &_status)){
+
+            if (!_logSuspendError && !_logSuspendReplay && _tempLogFile.isOpen()) {
+                uint8_t buf[MAVLINK_MAX_PACKET_LEN+sizeof(quint64)];
+
+                // Write the uint64 time in microseconds in big endian format before the message.
+                // This timestamp is saved in UTC time. We are only saving in ms precision because
+                // getting more than this isn't possible with Qt without a ton of extra code.
+                quint64 time = static_cast<quint64>(QDateTime::currentMSecsSinceEpoch() * 1000);
+                qToBigEndian(time, buf);
+
+                // Then write the message to the buffer
+                int len = mavlink_msg_to_send_buffer(buf + sizeof(quint64), &_sent_message);
+
+                // Determine how many bytes were written by adding the timestamp size to the message size
+                len += sizeof(quint64);
+
+                // Now write this timestamp/message pair to the log.
+                QByteArray b(reinterpret_cast<const char*>(buf), len);
+
+                if(_tempLogFile.write(b) != len)
+                {
+                    // If there's an error logging data, raise an alert and stop logging.
+                    emit protocolStatusMessage(tr("MAVLink Protocol"), tr("MAVLink Logging failed. Could not write to file %1, logging disabled.").arg(_tempLogFile.fileName()));
+                    _stopLogging();
+                    _logSuspendError = true;
+                }
+            }
+        }
+    }
+
+}
+
+/**
  * This method parses all incoming bytes and constructs a MAVLink packet.
  * It can handle multiple links in parallel, as each link has it's own buffer/
  * parsing state machine.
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
Sorry out of town. I think this is ok. But I need to think on it a little to see whether it will affect playback. Have you tried playing back these log files to see what happens:
* Mavlink inspector
* Charting
* ...

tilaktilak:
I made a simple mission to test. The yaw angle is changed with the joystick and this is Ardupilot simulator.
Everything looks consistent, but I am not really familiar with Replay so I could miss some bugs. Here my telemetry log ( extension renamed from tlog to log).

[2019-08-26 09-12-06.log](https://github.com/mavlink/qgroundcontrol/files/3538642/2019-08-26.09-12-06.log)

tilaktilak:
I correct the mention : @Antiheavy if you can have a look :-)

src/comm/MAVLinkProtocol.cc
```diff
@@ -159,6 +159,52 @@ void MAVLinkProtocol::resetMetadataForLink(LinkInterface *link)
...
+ **/
+
+void MAVLinkProtocol::logSentBytes(LinkInterface* link, QByteArray b){
+
+    uint8_t mavlinkChannel = link->mavlinkChannel();
+    static mavlink_message_t _sent_message;
+
+    for (int position = 0; position < b.size(); position++) {
+
+        if(mavlink_parse_char(mavlinkChannel, static_cast<uint8_t>(b[position]), &_sent_message, &_status)){
```
DonLakeFlyer: You can't use the same mavlink channel here. It will screw up the link statistics stored in the mavlink status struct associated with the channel. In order to do this it will require a separately reserved mavlink channel for the decode. But that isn't really possible since it will cut in half the number of vehicles you can connect to. There is a way to change mavlink to allocate status structs instead of the using the fixed array. But QGC does not use that and it's a bit of work to switch to that, but doable.

DonLakeFlyer: Sorry I didn't realize this would be a problem initially.

tilaktilak: Actually I don't need that test on my function as it's called by a signal emitted exactly when the message is sent. The buffer only contains the current message and starts by this message. Let me change this !

tilaktilak:
I made some changes on the logSendBytes function to simplify it and avoid calling mavlink_parse_char.
@DonLakeFlyer

Antiheavy:
I can’t speak to the technical implementation, but the feature of logging both in and out data in the TLOG seems very useful, especially for a log replay. I think?

DonLakeFlyer:
I built this pull on OSX. Booted it up with no vehicle and I just get "Mavlink logging failed..." over and over again.

tilaktilak:
Ok let me have a look !

tilaktilak:
I might have cut too short [here](https://github.com/mavlink/qgroundcontrol/pull/7714/commits/babe2b7c422b5b0a94b94dc6f8c56acd0b3e0a28), I missed the test to insure log is not suspended or tempFile is open.    
I think it's fixed, @DonLakeFlyer can you run a test again ?

src/comm/TCPLink.cc
```diff
@@ -81,6 +81,7 @@ void TCPLink::_writeBytes(const QByteArray data)
 
     if (_socket) {
         _socket->write(data);
+        emit bytesSent(this, data);
```
DonLakeFlyer: I only see TCP and UDP adding the bytesSent signal. What about SerialLink and BLuetoohLink?

DonLakeFlyer:
I've tested this and it seems to work fine. The main problem right now in Master is that the MAVLink Inspector is missing functionality that it previously had. The current implementation only support a single vehicle which means that the GCS side messages are never displayed in it. So although this does seem to work correctly by testing it using Log Replay in Stable and Mavlink Inspector there. On master there isn't much you can do with these newer telemetry logs.

I'm a bit concerned about merging this in before the mavlink inspector is fixed.

tilaktilak:
I am not sure we want MavLink inpector to shows QGC outcoming commands as it does a pretty good job to shows telemetry messages coming from the aircraft but It might be confusing adding all upcoming messages their.

DonLakeFlyer:
> Are you refering to this issue to add a filter by SystemId/CompId

No, it's covered here: #7770. The outgoing messages won't show up at all in the mavlink inspector. Without that there is no way to use/look at the outgoing data from the log.

> Depending on which tool you use to analyze the logs, it's still really useful to log both ends

What other tool is there to analyze telemetry logs?

tilaktilak:
> What other tool is there to analyze telemetry logs?

I don't know if there is a normal workflow in QGC, but the simplest one is mavlogdump to analyze the text file directly, pymavlink itself it pretty handy as well. That's for what I use but there is a lot of tools out there to analyze mavlink logs.

DonLakeFlyer:
Ok, I'm fine with merging this then when travis passes

DonLakeFlyer:
Thanks @tilaktilak

PR #7716, opened 2019-08-21
MissionItem memset to 8 instead of 0

tilaktilak:
This memset seems to create an issue with GotoLocation on Ardupilot : https://github.com/ArduPilot/ardupilot/issues/12085

@DonLakeFlyer

### DIFF
--- a/src/MissionManager/MissionManager.cc
+++ b/src/MissionManager/MissionManager.cc
@@ -46,7 +46,7 @@ void MissionManager::writeArduPilotGuidedMissionItem(const QGeoCoordinate& gotoC
     mavlink_message_t       messageOut;
     mavlink_mission_item_t  missionItem;
 
-    memset(&missionItem, 8, sizeof(missionItem));
+    memset(&missionItem, 0, sizeof(missionItem));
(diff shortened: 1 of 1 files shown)


src/MissionManager/MissionManager.cc
```diff
@@ -46,7 +46,7 @@ void MissionManager::writeArduPilotGuidedMissionItem(const QGeoCoordinate& gotoC
     mavlink_message_t       messageOut;
     mavlink_mission_item_t  missionItem;
```
dagar: You could also drop the memset entirely. 
```suggestion
    mavlink_mission_item_t  missionItem{};
```

DonLakeFlyer: I think that will work. Not quite sure though. We've had problems with some of the newer c++ stuff working on all platforms.

DonLakeFlyer:
@tilaktilak Can you create a pull for Stable_V3.5.1 as well?

DonLakeFlyer:
Also since this is a fairly important bug fix it should be noted in changelog.md

PR #7717, opened 2019-08-22
Make advanced mode persistent

stefandunca:
Keep the state of advanced mode. Default is enabled for QGC and disabled for custom-example

### DIFF
--- a/custom-example/src/CustomPlugin.cc
+++ b/custom-example/src/CustomPlugin.cc
@@ -86,7 +86,8 @@ CustomPlugin::CustomPlugin(QGCApplication *app, QGCToolbox* toolbox)
     : QGCCorePlugin(app, toolbox)
 {
     _pOptions = new CustomOptions(this, this);
-    _showAdvancedUI = false;
+    if(!_settings.contains(_kShowAdvancedUI))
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
I don’t this will work for most commercial customers. They want to keep people out of advanced mode as mush as possible and it is only used for customer support reasons.

stefandunca:
> I don’t this will work for most commercial customers. They want to keep people out of advanced mode as mush as possible and it is only used for customer support reasons.

They still want to use it for development as well, and triggering this easter-egg is annoying.I was thinking of a  --dev parameter

dogmaphobic:
This only affects custom builds. Main QGC doesn't pay attention to "advanced mode". If you move everything to the custom part (and away from `QGCCorePlugin`), it doesn't affect any other existing custom build.

If you want to make this "official", use a proper setting. Move it to `AppSettings`. That way a custom plugin will have control over how this is used.

DonLakeFlyer:
> This only affects custom builds.

Ah. I would say it's wrong for the custom build example as well. That's not going to be what most custom build people want.

Just add an --advanced-mode command line option if you want it start to in advanced all the time.

PR #7718, opened 2019-08-23
Initial support for link pairing

dogmaphobic:
Pairing requires a pairing daemon running on the vehicle. Documentation and MAVLink specifications are coming shortly.

First, pairing must be enabled:
Screenshot from 2019-08-22 22-55-17

Instead of "Waiting for Vehicle Connection", if pairing is enabled the pairing icon is shown. Once clicked/touched you will be presented with a dialog with the pairing options and instructions (which are not quite there yet).
Screenshot from 2019-08-22 22-56-44

Once a pairing option is selected, the instructions for pairing a vehicle/link with that option is presented (again, not quite there yet)
Screenshot from 2019-08-22 22-57-23

Pairing
Screenshot from 2019-08-22 22-58-07

Once you have one or more paired links, when you select the "Pairing" icon on the toolbar, you get a list of paired vehicles to chose from.
Screenshot from 2019-08-22 22-59-05


DonLakeFlyer:
What's with all the OpenSSL stuff in here?

QGCSetup.pri
```diff
@@ -87,7 +87,7 @@ WindowsBuild {
     }
 
     DEPLOY_TARGET = $$shell_quote($$shell_path($$DESTDIR_WIN\\$${TARGET}.exe))
-    QMAKE_POST_LINK += $$escape_expand(\\n) $$QT_BIN_DIR\\windeployqt --no-compiler-runtime --qmldir=$${BASEDIR_WIN}\\src $${DEPLOY_TARGET}
```
DonLakeFlyer: What does this change do?

dogmaphobic: Something to do with what to copy and not to copy (system runtime libraries) @MatejFranceskin Do you know why this was necessary?

DonLakeFlyer: If it has to to with the C Runtime. Was it verified that new versions of installed QGC on a virgin windows machine still work?

Issue #7698, opened 2019-08-17
Weird line shows up in Action selection window, then fades


Antiheavy:
In recent daily builds, a weird line shows up in the "Select Action" window, then it fades away after a few seconds.   Seems weird.

![image](https://user-images.githubusercontent.com/12117027/63205757-acd3ad80-c06e-11e9-95f1-48e1d4622037.png)

PR #7724, opened 2019-08-26
Guided action list: Fix list scrolling

DonLakeFlyer:
Fix for #7698


hamishwillee:
Since you were just in the code, how hard would it be to skip the action list altogether if there is only one action?

PR #7753, opened 2019-09-03
Flight Modes

dogmaphobic:
* Add access to all (unfiltered) flight modes.
* Have joystick use the unfiltered list of flight modes for button assignment.

Screen Shot 2019-09-03 at 11 28 21 AM

### DIFF
--- a/src/FirmwarePlugin/FirmwarePlugin.h
+++ b/src/FirmwarePlugin/FirmwarePlugin.h
@@ -86,6 +86,12 @@ public:
         return QStringList();
     }
 
+    /// Returns the list of all flight modes.
+    virtual QStringList allFlightModes(Vehicle* vehicle) {
+        Q_UNUSED(vehicle);
+        return QStringList();
+    }
+
     /// Returns the name for this flight mode. Flight mode names must be human readable as well as audio speakable.
     ///     @param base_mode Base mode from mavlink HEARTBEAT message
     ///     @param custom_mode Custom mode from mavlink HEARTBEAT message
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -168,6 +168,21 @@ QStringList PX4FirmwarePlugin::flightModes(Vehicle* vehicle)
     return flightModes;
 }
 
+QStringList PX4FirmwarePlugin::allFlightModes(Vehicle* vehicle)
+{
+    QStringList flightModes;
+    foreach (const FlightModeInfo_t& info, _flightModeInfoList) {
+        bool fw = (vehicle->fixedWing() && info.fixedWing);
+        bool mc = (vehicle->multiRotor() && info.multiRotor);
+        // show all modes for generic, vtol, etc
+        bool other = !vehicle->fixedWing() && !vehicle->multiRotor();
+        if (fw || mc || other) {
+            flightModes += *info.name;
+        }
+    }
+    return flightModes;
+}
+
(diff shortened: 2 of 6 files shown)


DonLakeFlyer:
This is going to break existing custom implemenations which override FirmwarePlugin::flightModes to provide their own filtered list. It would be better to change the signature of the existing FirmwarePlugin::flightModes to add a flag to ignore vehicle type. This way custom builds will fail and need fix up. Otherwise they'll just change functionality silently.

dogmaphobic:
I've added a new function to get all modes. The original one is still there unchanged. This was a direct result of the plugin implementing `FirmwarePlugin::flightModes()` but wanting all modes available for button assignment (not just the filtered list).

https://github.com/mavlink/qgroundcontrol/blob/qgc-centos/src/FirmwarePlugin/FirmwarePlugin.h#L82-L93

DonLakeFlyer:
> but wanting all modes available for button assignment (not just the filtered list).

That will break custom builds where commercial vehicle operators don't want their users to see full flight mode list even in joystick setup. It was specifically changed a while back to do that on customer request.

dogmaphobic:
Got it. I assumed the filtered list was for the Flight Mode indicator/selector UI (toolbar) only. Now we need to sort out how to enable all modes for joystick button assignment. That is, for when you want the UI selectable modes filtered but all modes available for joystick buttons.

dogmaphobic:
For example, it doesn't make a whole lot of sense to have "Land" or "Take Off" available as a flight _mode_ within the selector (hence the filtered list) but you might want to assign a button to it.

DonLakeFlyer:
> I assumed the filtered list was for the Flight Mode indicator/selector UI (toolbar) only.

It was originally was. Then I got complaints about it not affecting joystick.

> That is, for when you want the UI selectable modes filtered but all modes available for joystick buttons.

That would seem like a custom build tweak by overriding flightModes. To me the default behavior of filter everywhere is better for upstream.

dogmaphobic:
>That would seem like a custom build tweak by overriding flightModes. To me the default behavior of filter everywhere is better for upstream.

The plugin would need to override joystick configuration not flight modes. You can't tell what you are filtering from `FirmwarePlugin::flightModes()`

DonLakeFlyer:
> For example, it doesn't make a whole lot of sense to have "Land" or "Take Off" available as a flight _mode_ within the selector (hence the filtered list) but you might want to assign a button to it.

Hmm, a bit hard to say from an upstream standpoint as to how people are doing things. Not sure if things covered by guided actions should be removed from flight mode dropdown. Maybe...

DonLakeFlyer:
> The plugin would need to override joystick configuration not flight modes. You can't tell what you are filtering from `FirmwarePlugin::flightModes()`

Hmm, yeah. I think flightModes usage is kind of getting hacked around for too many things. Maybe it's time to split into toolbarFlightModes and joystickFlightModes. Would need to review usage of flightModes to see if it is used anywhere else.

dogmaphobic:
Instead of `FirmwarePlugin::allFlightModes()` rename it to `FirmwarePlugin::joystickFlightModes()`. Then it is clear what the list is for and the plugin can override it or just map it to its normal, filtered list.

dogmaphobic:
:) We both came up with the same idea at the same time

DonLakeFlyer:
I think trying Land/Takeoff out of the toolbar is an interesting thing as well. Less is more.

dogmaphobic:
Are you saying we should filter out Land/Takeoff by default from the toolbar selector? I agree...

DonLakeFlyer:
> Are you saying we should filter out Land/Takeoff by default from the toolbar selector?

yup

PR #7771, opened 2019-09-05
Plan/Fly: Remove child item display

DonLakeFlyer:
Items which have no lat/lon no longer display in views.


dogmaphobic:
@DonLakeFlyer 
Why was this removed?

PR #7772, opened 2019-09-05
WIP: tap/click on video stream preview to switch main/preview stream sources

stefandunca:
- [ ] The video doesn't frame properly in its view
- [ ] Handle main is map
- [x] Handle fullscreen

### Main stream in full view
pip_switch-main_is_main
### After tap on preview secondary stream takes full view and main is shown in preview
pip_switch-thermal_is_main
### Tap again on preview will switch to main stream in full view

### DIFF
--- a/custom-example/res/CustomCameraControl.qml
+++ b/custom-example/res/CustomCameraControl.qml
@@ -149,7 +149,7 @@ Item {
                     width:              buttonSize
                     height:             buttonSize
                     iconSource:         "/custom/img/thermal-standard.svg"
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Need some visuals to explain exactly what this change does...

DonLakeFlyer:
@stefandunca Thanks. So there is not real extra ui here. Just using existing stuff in better way.

dogmaphobic:
@stefandunca Is this still relevant?

Issue #7775, opened 2019-09-05
Layer shown while editing Mission/Fence/Rally


tilaktilak:
This issue is a follow-up to the discussion from #7765

When we edit the Fence or Rally point the Mission Layer is hidden. It's useful to show all layers not depending on what we are editing (Fence, Rally or Mission). Other layers that are not editable at the moment could be put in gray.

Need to be done in accordance with @DonLakeFlyer work on complex item visuals.

DonLakeFlyer:
So the problem here is working through the clutter of showing everything on the screen. Does it prevent you from getting to the fence or rally point item on the screen that you need to interact with. 

I think fence is likely easier since a fence would normally be outside the area of a mission. But even with that, getting to the polygon tool in the center may not be easy because it is covered up by other stuff.

Rally points could be much worse with a complex mission like a survey in combination with a survey for example.

To make it all work it may require dynamic manipulation of the z-order of elements such that the active editing layer elements are moved to the top of the z-order.

DonLakeFlyer:
There is also the issue of the "current Item" for mission items. I've changed the complex item code such that it show more minimal ui when the item is not current. That minimal ui should also be shown if the editing layer is not Mission. But requires extra work.

tilaktilak:
> So the problem here is working through the clutter of showing everything on the screen. Does it prevent you from getting to the fence or rally point item on the screen that you need to interact with.
> 
> I think fence is likely easier since a fence would normally be outside the area of a mission. But even with that, getting to the polygon tool in the center may not be easy because it is covered up by other stuff.
> 
> Rally points could be much worse with a complex mission like a survey in combination with a survey for example.
> 
> To make it all work it may require dynamic manipulation of the z-order of elements such that the active editing layer elements are moved to the top of the z-order.

Yeah, I don't have something really specific in mind at this stage but the goal is to keep kind of a shadow of the other layers on the map but they are not editable.

tilaktilak:
> There is also the issue of the "current Item" for mission items. I've changed the complex item code such that it show more minimal ui when the item is not current. That minimal ui should also be shown if the editing layer is not Mission. But requires extra work.

I think it's something we have to keep in mind doing that but it's related to https://github.com/mavlink/qgroundcontrol/pull/7779 only right ?

DonLakeFlyer:
> I think it's something we have to keep in mind doing that but it's related to #7779 only right ?

[...]

PR #7779, opened 2019-09-05
Fly: Hide waypoint indicators unless they are really needed

DonLakeFlyer:
Here is what we have now:
Screen Shot 2019-09-05 at 4 16 40 PM

That's a massive amount of crap in front of the users faces most of which has no value.

So instead we hide the waypoint indicators except for the one the vehicle is travelling towards:
Screen Shot 2019-09-05 at 4 14 09 PM
Screen Shot 2019-09-05 at 4 14 24 PM

The waypoints indicators show up if you Pause the vehicle. This allows you to then do a GoTo Waypoint by clicking on one:
Screen Shot 2019-09-05 at 4 14 40 PM

Next step: Direction Arrows like I have working in Plan view.

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewMap.qml
+++ b/src/FlightDisplay/FlightDisplayViewMap.qml
@@ -233,7 +233,7 @@ FlightMap {
             map:                flightMap
             largeMapView:       mainIsMap
             masterController:   masterController
-            isActiveVehicle:    _vehicle.active
+            vehicle:            _vehicle
 
             property var _vehicle: object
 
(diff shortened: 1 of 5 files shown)


tilaktilak:
Is it working when the flight plan is retrieved from the drone ?
Is the drone always on top of the WP layer ? 
As mentionned in https://github.com/mavlink/qgroundcontrol/issues/7775 what do we show in PlanView ?

DonLakeFlyer:
> Is the drone always on top of the WP layer ?

I think there are still z-order problems there.

> Is it working when the flight plan is retrieved from the drone ?

Not sure what you mean. Is "what" working? These visuals work this way in Fly view no matter where the items came from.

> As mentionned in #7775 what do we show in PlanView ?

Plan view always shows waypoint indicators since you need to be able to select them to edit and/or drag them around.

DonLakeFlyer:
The changes in this pull affect Fly view only.

sanderux:
@DonLakeFlyer Not showing the sequence numbers is very much affecting our workflow when using external analyzers. We indicate problems based on these numbers.

Could i request to enable them with a setting?

DonLakeFlyer:
Make sense, can you add a separate issue

PR #7785, opened 2019-09-07
Enable ArduPilot motor testing

DonLakeFlyer:
I implemented this something like two years ago. But I've been hoping that ArduPilot would support motor testing by motor output number as opposed to the letter thing which just seems really odd to me. Anyway, that still hasn't happened and it's a key setup element which is missing. So I just switched the existing code over to the motor letter model.

Along with it I fixed up the motor test page to be safer. Which hopefully at some stage can be used to support PX4 motor testing as well.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,7 @@ Note: This file only contains high level features or important fixes.
 
 ### 3.6.0 - Daily Build
 
+* ArduPilot: Add Motor Test vehicle setup page
 * Compass Instrument: Add indicators for Home, COG and Next Waypoint headings.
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
@jaxxzer I changed the Vehicle::motorTest method which I think you use to include a timeout:
```
    /// Test motor
    ///     @param motor Motor number, 1-based
    ///     @param percent 0-no power, 100-full power
    ///     @param timeoutSec Disabled motor after this amount of time
    Q_INVOKABLE void motorTest(int motor, int percent, int timeoutSecs);
```

Which means you may need to update your motor test page.

PR #7804, opened 2019-09-09
Add Takeoff SimpleItem if first item is complex

### LINKED ISSUE #7799: ArduPilot: Takeoff must be first mission command
Following #7773 and #6776 

There is no TakeOff added if we create a mission with a complex pattern.
Need to insure a takeoff is added at the beginning of every ArduPilot mission.

tilaktilak:
Fixes #7799

As it's done for SimpleItem insertion, add a SimpleItem with TakeOff command before the ComplexItem


fredowski:
Hi, this results in a build failure because the variable i is not declared.

dogmaphobic:
> Hi, this results in a build failure because the variable i is not declared.

Yup, I just found it myself as well.

Issue #7774, opened 2019-09-05
Terrain Following Fix and Improvements


tilaktilak:
This issue is a follow-up to the discussion from #7765 

 ## Terrain Altitude of Planned Home is not determined properly
At first it's not determined, but works if we move it. Here the error shown :   
```
TerrainQueryLog: Error in fetching elevation tile. Empty response.
```
![image](https://user-images.githubusercontent.com/6662416/64291679-e25c1e80-cfbc-11e9-8d8c-fe8d1086d6df.png)

TODO :

- Currently QGC does not support ArduPilot vehicle terrain requests where terrain maps are uploaded to the vehicle such that things like an RTL and follow terrain

- Changing to a new terrain server and supporting vehicle terrain requests. (Probably a big item)

DonLakeFlyer:
So the first phase for this is to switch to a new terrain server. I want to do that like ArduPilot does which is using height files (.hgt, https://dds.cr.usgs.gov/srtm/version2_1/Documentation/SRTM_Topo.pdf). Auterion has agreed to host the server. ArduPilot folks are ok with QGC hitting there servers until Auterion can stand up the data.

I had already done some false starts at coding this. The trickiest part is going to be the fact that normally height files are zipped. Which means QGC needs cross-platform unzip capability. That likely means including the zlib source in QGC and building it ourselves. Would could create a different format for zipping which would work with Qt quncompress but having unzip as a general capability in QGC is useful for more than just terrain. If we had unzip we could read .kmz files instead of just .kml files.

Once we get the file downloaded (using QGCDownload), and unzipped. It's just a matter of translating into the binary format for our offilne database. TerrainQuery.* and TerrainTile.* are a good place to start looking at that.

tilaktilak:
That's great for #7790 ! 
I can have a look at zlib and how to compile it then.

tilaktilak:
I notice that zlib looks present already in our repo : https://github.com/mavlink/qgroundcontrol/tree/master/libs/zlib/windows

edwinhayes:
The implementation should to be able to support DEM tiles from a GIS tile server in GeoTIFF format (for example).  What is the reason for using the SRTM height format in the first place?

DonLakeFlyer:
> What is the reason for using the SRTM height format in the first place?

Because it's a compact format which compresses well. Which in turn means download times are fast.

DonLakeFlyer:
> I notice that zlib looks present already in our repo, at least for windows :

[...]

PR #7818, opened 2019-09-18
Terrain Provider step 1 : Generic MapProviders

tilaktilak:
Issue related : https://github.com/mavlink/qgroundcontrol/issues/7774
I started to work on the integration of a new providers for terrain.
I also want to have the ability add/remove custom providers for both map and terrain.

So this is to make room for a big coming job :
- Make a generic interface for a MapProvider
- Subscribe new map provider by adding a new `MapProvider` to a hash table
- Remove duplicate/hard coded list of providers from backend / ui sides. The ui will retrieve all necessary info from the hash table, simple way to insure consistency between front and back.

Now we can :
- Implement several Terrain providers with several protocols and use still use the cache offline feature.
- Improve OfflineMaps menu to cache terrain tiles at different zoom, from several providers.
- Create a UI menu to add custom map provider (WMS compatible for instance), this will instantiate and subscribe a MapProvider.

@edwinhayes


DonLakeFlyer:
"Improve OfflineMaps menu to cache terrain tiles at different zoom, from several providers."

I don't quite get how this would be used. To me you always want the best terrain information available. Also say you are planning a mission which follow terrain, there is no way in that process to say I want to use terrain at a specific soom level.

DonLakeFlyer:
@dogmaphobic Gus is going ot have to look at this.

tilaktilak:
> "Improve OfflineMaps menu to cache terrain tiles at different zoom, from several providers."
> 
> I don't quite get how this would be used. To me you always want the best terrain information
available. 
 
 I agree with the fact that we want best terrain info most of the time, but depending on the area covered and the precision of the provider (can be up to 5m, or even more if a custom DEM is used), it make sense to choose a relevant zoom level / definition to the mission.

> Also say you are planning a mission which follow terrain, there is no way in that process to say I want to use terrain at a specific soom level.

I am referring to the "Offline Maps" menu as the way to cache elevation tiles.

edwinhayes:
> I don't quite get how this would be used. To me you always want the best terrain information
available.

- Imagine the case of planning a mission that covers 500km: you don't necessarily want to sample at 30m intervals; you might want to instead perform some kind of regionation so that you're only 'following' the highest point in each 1km square.

- Sometimes, you might want to load in a custom DEM; for instance, flying around a building site; but the DEM might be at a really high resolution - 1-2cm for instance.

- This allows rendering elevation information on the map.

edwinhayes:
> Also say you are planning a mission which follow terrain, there is no way in that process to say I want to use terrain at a specific soom level.

This is definitely a legit consideration; we will need to give some consideration to what we might need to do about this, but that'll be a separate thread.

DonLakeFlyer:
I'm not saying that all terrain tiles stored need to be at the same resolution. When I'm saying is that I don't understand why you would store multiple tiles for the same area at differing resolution with some being lower.

> * Imagine the case of planning a mission that covers 500km: you don't necessarily want to sample at 30m intervals; you might want to instead perform some kind of regionation so that you're only 'following' the highest point in each 1km square.

That sort of thing isn't going to be supported by the feature which sends tiles to the vehicle. At least I don't think that's the case. The ability to build such a mission is more of a new feature of planning. You don't need multiple levels of tile information to do that. You just need the abilityu to query a maximum height for a certain area.

DonLakeFlyer:
> but depending on the area covered and the precision of the provider (can be up to 5m, or even more if a custom DEM is used), it make sense to choose a relevant zoom level / definition for the mission.

That would be a feature of the mission generation not necessarily a feature of storing zoom levels on elevation tiles. Given the fact that I don't see the cases where you want to use lower levels of terrain resolution as being something used very often. I don't seen storing a lot of additional information to support a rarely used thing to be a good idea.

Also I you implement such a thing based on elevation tile zoom it means it will only work if you have the lower resolution data. That doesn't seem like a good thing either. You want it to work with whatever data you have and it is possible to support that. If you support building lower resolution terrain missions from higher resolution data why would you need to do it any other way then that?

DonLakeFlyer:
Wait, I'm thinking about this wrong. I've been thinking only of why you want lower resolution data.  If you want to store a smaller area of higher resolution data while you still have existing say normal resolution coverage in that area you need to do that using a higher zoom level which may cover a smaller tile size geographically than the lower tile. Ok, I get that.

tilaktilak:
The thing is we also want to take advantage of what is present in the code at the moment in term of tiling/caching. The way it was started with the support of airmap elevation in Offline was great ( Using the cache from the map engine through OfflineMaps menu, so tiles are stores in the sql database).
My approach here is simply to create a generic interface for map providers from several sources and either for a base map or an elevation map. So it become fairly easy to change from a terrain provider to another, same for the basemap.
It's also a refactor to avoid hard-coded list of map providers in QGCMapEngine.cc QGCMapEngineManager.cpp, QGCMapUrlEngine.h and all the `if "Airmap Elevation"` in several places as well.
I also think that it's a good method to smooth the transition from Airmap to the next solution that will be decided. It's an open door to custom DEM for mapping purpose, caching options , add custom basemap and all of this at run time !

DonLakeFlyer:
@tilaktilak Yeah all makes great sense. Gus, who wrote the original map code is on vacation now. When he gets back I'll ping him to look at this. I know very little about this code unfortunately.

tilaktilak:
Following discussion here : #7864

@cosmarc 
So I rework a little bit the TerrainQuery code + create a generic class for MapProvider ( Then a big hash table contains all providers ). 
This is beneficial for both map/elevation. We can create a child of MapProvider that will fetch data from a WMS server, or even a SRTM mirror, same for 
I ended up adding gdal to QGC to open geotiff and apply transformation, so with gdal onboard theorically we can open what we want (See here : https://gdal.org/api/raster_c_api.html#_CPPv48GDALOpenPKc10GDALAccess) ! 
On my current version QGC search in a folder for .tiff files to try to retrieve elevation data from it, if it's not available fallback to Airmap elevation.
My current working branch is really messy (include some UI changes on OfflineMap that are just random experiments to connect things together) but it's here : https://github.com/Aeronavics/qgroundcontrol/tree/MapProviders_step2
I Also had in mind adding the ability to manage several elevation provider ( and class them by priority ) and same thing with the map provider (Adding the ability to overlay maps and put an orthophoto on the top of a basemap as it appear useful for our customers, that is basically the purpose of my messy UI changes in MapProviders_step2 branch).

There is a first PR related to the creation of a generic class for MapProviders here : https://github.com/mavlink/qgroundcontrol/pull/7818

My work/challenges at the moment : 
- Link gdal statically to optimize QGC binary size.
- I manage to compile my code on Windows, Linux and Android with dynamic libs but have no access to an OSX platform.
- I need a nice UI menu to connect this backend ( How to add  elevation providers map layers from a given list, add custom element in this list (with a WMS url or a local folder of tile), handle layers ordering, opacity for basemap. Integrate the ability to overlay maps in FlightMap item ( I though about a repeater of FlightMap items somewhere between the current map and the items on the map ).

dogmaphobic:
@tilaktilak Can you update the PR so I can merge it?

tilaktilak:
Done ! :-)

dogmaphobic:
There are some build errors:
```
compiling src/QtLocationPlugin/ElevationMapProvider.cpp
1695/home/travis/build/mavlink/qgroundcontrol/src/QtLocationPlugin/MapProvider.cpp:66:59: error: use of undeclared identifier 'pow'
1696    return static_cast<int>(floor((lon + 180.0) / 360.0 * pow(2.0, z)));
```

tilaktilak:
Let me have a look

PR #7819, opened 2019-09-19
Pairing manager - Microhard pairing procedure improvements

MatejFranceskin:
This PR addresses issues that were found during testing of the pairing procedure:
- Do not hardcode 192.168.168 as IP prefix - extract it from air unit IP.
- Stop video during pairing and start it only when pairing manager successfully connected.
- Automatically connect to last paired device.
- Maintain a list of connected devices in pairing manager
- Implemented exchange of public keys, signing and encrypting of connect message and verification of origin of connect response.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,12 +28,12 @@ matrix:
         - name: "Android 32 bit"
           dist: trusty
           language: android
-          env: SPEC=android-clang CONFIG=installer BITNESS=32-bit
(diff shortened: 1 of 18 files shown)


edwinhayes:
@tilaktilak - We might want to help test this out.

DonLakeFlyer:
@MatejFranceskin This should be closed?

PR #7821, opened 2019-09-20
Add EKF vision velocity fusion to EKF_AID_MASK

kamilritz:
This adds the possibility to set the fusion of external vision velocity with the EKF_AID_MASK.
The external vision velocity will be introduced with: https://github.com/PX4/ecl/pull/642


DonLakeFlyer:
PX4 parameter meta data is pulled over regularly by the firmware folks. Why an individual pull just for this?

Issue #7743, opened 2019-08-31
Segmentation fault on start with latest Daily (Ubuntu 18.04)


MaEtUgR:
## Expected Behavior
QGC starts normally like the older daily versions I was running before.

## Current Behavior
The latest daily I downloaded today (31.08.2019) segfaults immediately on start before the window comes up:
![segfault](https://user-images.githubusercontent.com/4668506/64065507-a9aff400-cc0e-11e9-93a3-3627fc0bc4ac.PNG)

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1. `wget https://s3-us-west-2.amazonaws.com/qgroundcontrol/builds/master/QGroundControl.AppImage`
1. `chmod +x QGroundControl.AppImage`
1. `./QGroundControl.AppImage`
1. See segfault

## System Information
- Operating System: Ubuntu 18.04 in VMWare on Windows 10 host
- QGC Version: Latest Daily 31.08.2019 16:50 CET, I have no idea how to extract the version from the binary if it doesn't start.

## Detailed Description
I did not attach any autopilot via USB or UDP. Older versions are still working fine e.g. the last I used on that computer was the daily build based on aa2c794c0 (quite old).

dogmaphobic:
The error is shown in the console as soon as you launch it:
```
$ ./QGroundControl.AppImage 
/tmp/.mount_u6DmxD/QGroundControl: error while loading shared libraries: libcrypto.so.1.0.0: cannot open shared object file: No such file or directory
```

That is a missing OpenSSL dependency. @MatejFranceskin

MaEtUgR:
@dogmaphobic Thanks for the quick check. Why do I not have the same error as you mention in your comment? I have OpenSSL installed. Here's the text out of the screenshot above:
```
maetugr@ubuntuvm:~$ ./QGroundControl.AppImage
Settings location "/home/maetugr/.config/QGroundControl.org/QGroundControl.ini" Is writable?: true
Filter rules "*Log.debug=false\n"
GStreamer debug output: "/home/maetugr/Documents/QGroundControl/Logs/gst/gstreamer-log.txt"
Error loading text-to-speech plug-in "speechd"
serialnmea: No known GPS device found. Specify the COM port via QT_NMEA_SERIAL_PORT.
Failed to set Geoclue positioning requirements. Geoclue error: org.qtproject.QtDBus.Error.InvalidObjectPath
QGCPositionManager error 2
[...]

PR #7823, opened 2019-09-22
Disable pairing

DonLakeFlyer:
There is a long list of issues against this with nobody looking at them. I'm tired of fighting with this on my own builds as well. Until this can be made to work on all OS and verified to be working correctly and run correctly on virgin installs of all OS it needs to stay off.

Related to #7743 and all the other Pairing issues

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -313,6 +313,9 @@ DebugBuild {
 
 include(src/QtLocationPlugin/QGCLocationPlugin.pri)
 
+# Until pairing can be made to work cleanly on all OS it is turned off
+DEFINES+=QGC_DISABLE_PAIRING
+
 # Pairing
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I would also question why this is in master at all since as far as I can tell it is not supported on all OS.

DonLakeFlyer:
@stefandunca This should really move to your custom build until it is fully ready for prime time.

stefandunca:
@DonLakeFlyer we've moved the work on PairingManager on our own fork. I'm working to fix the OpenSSL conflict then merge the fixes back to master
> I would also question why this is in master at all since as far as I can tell it is not supported on all OS.
For now it works on Android and Linux but we target all the OSes in the end.
> This should really move to your custom build until it is fully ready for prime time.
The UI work is done in custom.

Thanks

PR #7824, opened 2019-09-22
Setting for microhard channel/frequency

MatejFranceskin:
Add a setting in Microhard settings for the channel used during communication.
During pairing phase a different pairing channel is used which is by default set to 78. The channel that is set in Microhard settings panel is then transfered to pairing-manager on companion computer side and in order to establish a connection both sides have to switch to this user selected channel.
MicrohardChannel

### DIFF
--- a/src/Microhard/MicrohardManager.cc
+++ b/src/Microhard/MicrohardManager.cc
@@ -18,6 +18,11 @@
 #define SHORT_TIMEOUT 2500
 #define LONG_TIMEOUT  5000
 
+// Microhard pMDDL 2350 constants
+#define MICROHARD_CHANNEL_START   1     // First MH channel
+#define MICROHARD_CHANNEL_END     81    // Last MH channel
+#define MICROHARD_FREQUENCY_START 2310  // First MH frequency in MHz
+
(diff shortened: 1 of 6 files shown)


src/Microhard/MicrohardManager.cc
```diff
@@ -25,23 +25,36 @@ static const char *kNET_MASK            = "NetMask";
...
 static const char *kCFG_PASSWORD        = "ConfigPassword";
 static const char *kENC_KEY             = "EncryptionKey";
+static const char *kPAIR_CH             = "PairingChannel";
+static const char *kCONN_CH             = "ConnectingChannel";
 
 //-----------------------------------------------------------------------------
 MicrohardManager::MicrohardManager(QGCApplication* app, QGCToolbox* toolbox)
     : QGCTool(app, toolbox)
 {
+    for (int i = 1; i < 82; i++) {
```
stefandunca: How about having the hard-coded values (nr. of channels and frequency offset ) as constants and a comment on where are they sourced from?

MatejFranceskin: Yes - I defined and commented the constants.

PR #7827, opened 2019-09-23
New flight trajectory display algorithm

DonLakeFlyer:
Problems with previous algorithm:
* Performance was crap. Too many points were added. Mobile builds had to limit path length due to perf problems.
* Run SITL it higher speed or replay log at higher speed an the flight path was really bad.

New algorithm is distance and azimuth change based with tolerances set to reduce point count yet still show high quality fligh path. This new code reduces point count for lines on the map by 80%+. But if you compare old to new the flight path shown is actually closer to the true flight path.


Antiheavy:
Cool!

PR #7839, opened 2019-09-29
Plan: Presets - tweak settings from tab

DonLakeFlyer:
* This allows you to complete a Survey setup all from the Presets tab
* Added a setting which specifies whether the Presets or Grid tab is shown as initial selection. Default is show Grid. Custom builds can override to show Presets for situations where most customers will use presets.

Screen Shot 2019-09-28 at 6 31 34 PM

### DIFF
--- a/src/PlanView/SurveyItemEditor.qml
+++ b/src/PlanView/SurveyItemEditor.qml
@@ -61,7 +61,7 @@ Rectangle {
             anchors.left:   parent.left
             anchors.right:  parent.right
 
-            Component.onCompleted: currentIndex = 0
+            Component.onCompleted: currentIndex = QGroundControl.settingsManager.planViewSettings.displayPresetsTabFirst.rawValue ? 2 : 0
 
             QGCTabButton { text: qsTr("Grid") }
             QGCTabButton { text: qsTr("Camera") }
@@ -255,7 +255,11 @@ Rectangle {
                 text:   qsTr("Statistics")
             }
 
-            TransectStyleComplexItemStats { }
+            TransectStyleComplexItemStats {
+                anchors.left:   parent.left
+                anchors.right:  parent.right
+                visible:        statsHeader.checked
+            }
         } // Grid Column
 
         Column {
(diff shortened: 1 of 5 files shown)


Antiheavy:
So the type of camera isn't assumed as part of each preset, yes?  Meaning, you can apply a preset, then chose different cameras and the survey will adjust accordingly?  I'm trying to decide if this is a good approach or not (vs having the camera type assigned as part of the preset).  I'm on the fence, arguments both ways.

DonLakeFlyer:
> Is the type of camera isn't assumed as part of each preset?

The camera is part of the Preset, but after you select the preset you can go to the Camera tab and change it. Applying a preset in a sense just sets a bunch of values. After you do that you can tweak whatever you want.

DonLakeFlyer:
> Applying a preset in a sense just sets a bunch of values.

This is a fundamental difference in how presets work now. 

Previously:
The preset was a setting on the survey and saved with the survey. It would set all the values to the preset values and then hide all the UI for modifying the survey other than a few key values: Angle, Entry Point. 

Initially this caused a problem in that the camera was also locked down. So I changed it such that you could decide whether the camera was included in the preset or not at save time. That added extra complexity to the whole thing from a user model standpoint. But it would also prevent you from doing things like using a Preset in an area say where you also needed to adjust for terrain. Since terrain was Off in the preset there was no way to change it since the UI for it was hidden. This sort of thing may be ok for specific use cases but in a general sense for a diverse QGC customer base it's just too limiting.

New approach:
In order to deal with the problems with the initial implication I changed the preset model such that the preset is _not_ saved with the Survey. Using a preset just sets the values of the survey to what is stored in the preset. The values stored in a preset are everything available from the Survey editor ui. As before it does not information associated with the polygon shape. After you apply a preset you can change anything you want or not. This is also the reasoning behind changing from selecting the preset from a combo box as a single action. That made more sense with the previous implementation. With the current implementation the concept of selecting a preset and then clicking "Apply" matches more closely the way a preset actually works.

PR #7846, opened 2019-10-01
Fix uninitialized variable that might cause an undefined behavior

stefandunca:
CentOS startup stack dump:
```
#0  __base_ctor (this=<optimized out>, acolor=...) at /home/user/Qt/5.12.4/gcc_64/include/QtGui/qcolor.h:302
#1  thumbJoystick (this=<optimized out>) at moc/../../qgroundcontrol/src/QGCMapPalette.h:57
#2  QGCMapPalette::qt_static_metacall (_o=<optimized out>, _c=<optimized out>, _id=<optimized out>, _a=<optimized out>, _a=<optimized out>, _id=<optimized out>, _c=<optimized out>, _o=<optimized out>) at moc/moc_QGCMapPalette.cpp:122
#3  0x000000000044df3b in QGCMapPalette::qt_metacall (this=0x35dd8a0, _c=QMetaObject::ReadProperty, _id=1, _a=0x7fffbeff2650) at moc/moc_QGCMapPalette.cpp:178
```

### DIFF
--- a/src/QGCMapPalette.h
+++ b/src/QGCMapPalette.h
@@ -64,7 +64,7 @@ signals:
     void lightColorsChanged(bool lightColors);
     
 private:
-    bool _lightColors;
+    bool _lightColors = false;
 
     static const int _cColorGroups = 2;
 


src/QGCMapPalette.h
```diff
@@ -64,7 +64,7 @@ class QGCMapPalette : public QObject
     void lightColorsChanged(bool lightColors);
     
 private:
-    bool _lightColors;
+    bool _lightColors = false;
```
DonLakeFlyer: Current QGC codebase doesn't use this style of member settings anywhere currently. It all uses constructor code style of setting these up. I'd rather not mix and match that usage pattern since it means you may have to look in two places to find what things are set to.

DonLakeFlyer: Looks like there are a number of cases of this style already in the codebase. So fine for now.

dogmaphobic: It seems more "right" to define the default right next to the definition instead of the constructor. It's easier to _not_ forget to initialize it and you don't have to pay attention to the initialization order, which the compiler complains if you do it in the wrong order within the constructor.

DonLakeFlyer: I don't disagree with that. But the problem is that you can't do everything there. Hence you are left with some things in the constructor and some things in the header. Leaving you with two places to look for things.

PR #7886, opened 2019-10-08
Wizard Mode for Create Plan

DonLakeFlyer:
The new Create Plan UI now puts the mission items into a "Wizard Mode" which will step you through what you need to do next. All of the automatic Wizard stuff only works from Create Plan. If you create a Plan by hand using the tools, things work like they used to without these "Done..." buttons.

Here is an example sequence for Create Survey on a Fixed Wing:

After you click Survey you end up here, where you futz with the Takeoff item. Once you are happy with it, you click "Done Adjusting" which will take you to the Survey item.
Screen Shot 2019-10-08 at 10 40 11 AM

This is where you end up after you clicked "Done Adjusting" on the Takeoff item. The image shows what it looks like after I added a polygon. Once you have your survey to where you want it you click "Done With Polygon". Which will take you to the Fixed Wing Landing Pattern item.
Screen Shot 2019-10-08 at 10 40 38 AM

The starting points for FW Landing is to click to set landing point just like before. 
Screen Shot 2019-10-08 at 10 40 50 AM

But once you do that you end up in this new place, which allows you to adjust the loiter point. This adjust works differently in this "WIzard Mode" then if you were just creating a Plan by hand. When you drag it just changes the angle of the approach. This way the defaults for the FW Landing Pattern are still respected as far as Glide Slope Angle and/or Landing Distance.
Screen Shot 2019-10-08 at 10 41 04 AM

### DIFF
--- a/src/MissionManager/CorridorScanPlanCreator.cc
+++ b/src/MissionManager/CorridorScanPlanCreator.cc
@@ -10,6 +10,7 @@
 #include "CorridorScanPlanCreator.h"
 #include "PlanMasterController.h"
 #include "MissionSettingsItem.h"
+#include "FixedWingLandingComplexItem.h"
 
 CorridorScanPlanCreator::CorridorScanPlanCreator(PlanMasterController* planMasterController, QObject* parent)
     : PlanCreator(planMasterController, MissionController::patternCorridorScanName, QStringLiteral("/qmlimages/PlanCreator/CorridorScanPlanCreator.png"), parent)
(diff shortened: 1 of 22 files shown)


Antiheavy:
Neat!  Some comments:

- There should be a "Set to vehicle location" button for both Takeoff and Landing Pattern.  

- Default Takeoff WP distance should be set out in front of the vehicle based on current heading and the specified climb rate and cruise speed of the vehicle.  Rotating the Takeoff WP from home should have this distance locked similar to how you propose with Landing Pattern.

- I wonder if the Waypoint and Pattern buttons on the left toolbar should only be available when planning a Custom Mission?  
image

- Most users don't know that the "hamburger" is an options menu.  Almost nobody I talk with realizes they can "insert" waypoints and patterns via that button.  Maybe this isn't a problem needing a solution right now, but just a comment.

- When using one of these Wizards, can we force the Takeoff to always be at the beginning and force only 1 landing pattern to always be at the end?  I see a lot of new users accidentally placing multiple landing patterns or accidentally placing waypoints and surveys AFTER the landing.  this sort of ties to my previous comments.

DonLakeFlyer:
Travis keep crapping out on 64 bit build. I'll try to make that work in a separate pull. It's not related to any changes here.

Issue #7870, opened 2019-10-05
Survey Pattern: qground crashed when setting 0 value to spacing


SajyBug:
when i create new survey pattern and set 0 value to spacing the app get to slow and not working well 
![Screenshot from 2019-10-05 15-28-30](https://user-images.githubusercontent.com/21136372/66254732-78839000-e787-11e9-8487-77e5792bbaa1.png)

hamishwillee:
@SajjadRahnama You do realise that a zero spacing means an infinitely dense grid right? Effectively you have created billions of waypoints to represent this line.

@DonLakeFlyer I guess this means enforcing minimum non-zero grid spacing?

PR #7894, opened 2019-10-09
Plan Patterns: Minimum transect spacing of 0.5 meters

DonLakeFlyer:
If you let the transect spacing get too small you end up with way too many transects and QGC will blow up from perf problems. In order to prevent this the minimum transect spacing is set internally to 0.5 meters. Anything lower than that will automatically just generate a single transect.

Fix for #7870


hamishwillee:
Awesome. Will the UI also restrict entry to this range?

DonLakeFlyer:
> Will the UI also restrict entry to this range?

There isn't a simple way to do that. The transect spacing in some cases is a calculated value based on altitude, camera specs and so forth. So you just end up with really narrow spacing based on a pile of other settings.

Issue #7875, opened 2019-10-05
White screen on flight screen in Android 8.0 phone


Hwurzburg:
beginning with the first daily builds in Oct or so (but I only restarted using daily builds on 9/30), on start up the flight screen shows the map fine with QGC gps location....after connecting and loading params, switching to fly screen is white in all the map areas....planner screen is still fine

sometimes, if I switch from connections screen quickly after params start downloading, the fly screen will be okay with map tiles shown

also, I have noticed that mode changes fail often under these conditions....

as of now, the daily builds on my Motorola RAZR Android 8.0 phone are not useable...

going back to the stable version, there is no issue with white screen

DonLakeFlyer:
@Hwurzburg Can you look at Settings/Console to see if there is any interesting log output in there when this happens.

Hwurzburg:
okay...white screen does not happen when using OTG USB to connect to vehicle....only when using telem (actually I use a 3DR Sik radio to BT repeater and then connect QGC via the BT)....
console shows errors like :"Elevation tile fetching returned error(QnetworkReply:operation cancelled error) and TLS initiailization failed

if this is of any help

PR #7896, opened 2019-10-09
Toolbar changes plus other stuff

DonLakeFlyer:
* Change order of toolbar indicators to keep visual ones together follow by combo box ones
* Fix spacing on toolbar to get rid of strange gap
* Fix Fly View white screen problem #7875
* Change Multi-Vehicle selector to a combo box

Screen Shot 2019-10-09 at 1 13 52 PM

### DIFF
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -323,16 +323,16 @@ const QVariantList &FirmwarePlugin::toolBarIndicators(const Vehicle* vehicle)
     //-- Default list of indicators for all vehicles.
     if(_toolBarIndicatorList.size() == 0) {
         _toolBarIndicatorList = QVariantList({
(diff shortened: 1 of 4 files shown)


hamishwillee:
FWIW I do not love the new layout. I find it easier to quickly locate the status information (most relevant) when it is against a corner than in the center. The vehicle, armed state, mode are much less interesting, and were a nice visual separator from the menu.

DonLakeFlyer:
> when it is against a corner than in the center

Not quite sure what you mean. Most stuff is in the same position it was before. Except for things most people don't see like multi-vehicle and rtk.

That said I'm not super happy with the combo box visuals. At some point I plan to come back with that. Maybe remove border and change background color to match such that only the dropdown indicators are there. But for now it solves the problem.

hamishwillee:
Ignore me, thinking of a customised UI with different constraints.

PR #7905, opened 2019-10-11
Display the Tools button also while param filtering is active

kamilritz:
A few parameter changes require a vehicle reboot. After changing such a parameter the search filter is in most cases still active. So you have to clear your search before you can access the Tools button to restart the vehicle. I do not see a reason why we should not be able to access the Tools button while having an active search filter.

This is not tested.

### DIFF
--- a/src/QmlControls/ParameterEditor.qml
+++ b/src/QmlControls/ParameterEditor.qml
@@ -97,7 +97,6 @@ Item {
         anchors.bottom: header.bottom
         anchors.right:  parent.right
         text:           qsTr("Tools")
-        visible:        !_searchFilter
         onClicked:      toolsMenu.popup()
     }
 


dogmaphobic:
I don't see any immediate reason why the _Tools_ button is hidden while search is active. But I assume @DonLakeFlyer had a reason for doing so. I would wait for his input.

Screen Shot 2019-10-14 at 10 05 14 AM

DonLakeFlyer:
Likely because there is a refresh parameters item in there which would have to do some special stuff if you are in the middle of a search. Maybe reset to not searching after list is refreshed (easie), or re-apply the search after you refresh (harder). So as it stands now just making it visible won't work correctly in all cases. There may be other things in that menu which could cause problems in the middle of a search context.

kamilritz:
@DonLakeFlyer Thank you very much for your explanation. Would it be possible for you to came up with a working solution? I have little experience with QT GUIs.
This would be much appreciated!

PR #7910, opened 2019-10-11
setting minimum roll_limit to 20 deg

moreba1:
Setting minimum fw_roll_lim to 20 deg #7910


DonLakeFlyer:
You cannot change the parameter meta data like this. It comes from firmware side. If you think it should be changed you should be talking with firmware folks.

PR #7919, opened 2019-10-16
Fix initial Map Provider setting

tilaktilak:
Fix : 
The default value for MapType / MapProvider was not defined


DonLakeFlyer:
Thanks

PR #7921, opened 2019-10-16
Photo gallery

caleridas:
Implement facilities to:
- take photos on the drone
- download them into QGC
- view and manage them in a simple way in a photo gallery part

The implementation provided here is also an exercise to improve the ways in which features are kept modular, with as little as possible interference with other code. Some necessary prerequesite baseline hooks are implemented in first few commits of this series, the rest is then fully encapsulated.

This is also an experiment to see how we can bring unit tests into QGroundControl. This is known to be problematic with qmake (only option is to have subdir-project files -- as done here, although not wired to main build).

### DIFF
--- a/src/PhotoGallery/AbstractPhotoTrigger.h
+++ b/src/PhotoGallery/AbstractPhotoTrigger.h
@@ -1,0 +1,60 @@
+/****************************************************************************
+ *
+ *   (c) 2019 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ *
+ * QGroundControl is licensed according to the terms in the file
+ * COPYING.md in the root of the source code directory.
+ *
+ ****************************************************************************/
+
+#pragma once
+
+#include <QObject>
+
+/// A "photo" operation in progress
+///
+/// Represents a photo being taken. This object exists while this is in
+/// progress and can be used by callers to track operation.
+class AbstractPhotoTriggerOperation : public QObject {
+    Q_OBJECT
+public:
+    ~AbstractPhotoTriggerOperation() override;
+
+    virtual bool finished() const = 0;
+    /// Only meaningful in finished.
+    virtual bool success() const = 0;
+    /// Id of the photo, if operation succeeded.
+    virtual QString id() const = 0;
+
+signals:
+    /// Signal emitted when the operation has finished.
+    ///
+    /// This object is alive until after emitting this signal has finished, it
+    /// must not be referenced after delivery of this signal has completed.
+    void finish();
+};
+
+/// Abstract interface to trigger taking a photo.
+class AbstractPhotoTrigger : public QObject {
+    Q_OBJECT
+public:
+    ~AbstractPhotoTrigger() override;
+
+    explicit AbstractPhotoTrigger(QObject * parent = nullptr);
+
+    /// Take a photo.
+    ///
+    /// \returns Operation in progress or nullptr.
+    ///
+    /// Starts taking a photo. If no photo can presently be taken at all
+    /// (e.g. no camera) then this may return nullptr to indicate failure.
+    /// Otherwise, returns a continuation object to represent the operation in
+    /// progress.
+    ///
+    /// The continuation object is alive for as long as the operation is in
+    /// progress, up and until its "finish" signal has been emitted.
+    ///
+    /// This method should also ensure that the operation cannot delay
+    /// "indefinitely" (but e.g. instead cancel after some timeout).
+    virtual AbstractPhotoTriggerOperation * takePhoto() = 0;
+};
--- a/src/PhotoGallery/AsyncDownloadPhotoTrigger.cc
+++ b/src/PhotoGallery/AsyncDownloadPhotoTrigger.cc
@@ -1,0 +1,155 @@
+/****************************************************************************
+ *
+ *   (c) 2019 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ *
+ * QGroundControl is licensed according to the terms in the file
+ * COPYING.md in the root of the source code directory.
+ *
+ ****************************************************************************/
+
+#include "AsyncDownloadPhotoTrigger.h"
+
+#include <QNetworkReply>
+
+#include "PhotoFileStore.h"
+
+AsyncDownloadPhotoTriggerOperation::~AsyncDownloadPhotoTriggerOperation()
+{
+}
+
+bool AsyncDownloadPhotoTriggerOperation::finished() const
+{
+    return _finished;
+}
+bool AsyncDownloadPhotoTriggerOperation::success() const
+{
+    return _success;
+}
+
+QString AsyncDownloadPhotoTriggerOperation::id() const
+{
+    return _id;
+}
+
+void AsyncDownloadPhotoTriggerOperation::finishSuccess(QString id)
+{
+    _finished = true;
+    _success = true;
+    _id = std::move(id);
+    emit finish();
+    delete this;
+}
+
+void AsyncDownloadPhotoTriggerOperation::finishFailure()
+{
+    _finished = true;
+    _success = false;
+    emit finish();
+    delete this;
+}
+
+AsyncDownloadPhotoTrigger::~AsyncDownloadPhotoTrigger()
+{
+    failPhotoOperation();
+}
+
+AsyncDownloadPhotoTrigger::AsyncDownloadPhotoTrigger(
+    std::function<bool()> photo_trigger_fn,
+    const Config & config,
+    PhotoFileStore * store ,
+    QObject * parent)
+    : AbstractPhotoTrigger(parent),
+      _photo_trigger_fn(std::move(photo_trigger_fn)),
+      _config(config)
+{
+    _photo_timeout.setSingleShot(true);
+    connect(&_photo_timeout, SIGNAL(timeout()), this, SLOT(failPhotoOperation()));
+    connect(&_download_manager, SIGNAL(finished(QNetworkReply*)), this, SLOT(downloadFinished(QNetworkReply*)));
+    if (store) {
+        setStore(store);
+    }
+}
+
+AsyncDownloadPhotoTriggerOperation * AsyncDownloadPhotoTrigger::takePhoto()
+{
+    if (_store && !_active_op) {
+        if (_photo_trigger_fn()) {
+            _active_op = new AsyncDownloadPhotoTriggerOperation();
+            _photo_timeout.start(_config.photo_timeout / std::chrono::milliseconds(1));
+            return _active_op;
+        } else {
+            return nullptr;
+        }
+    } else {
+        return nullptr;
+    }
+}
+
+PhotoFileStore * AsyncDownloadPhotoTrigger::store() const
+{
+    return _store;
+}
+
+void AsyncDownloadPhotoTrigger::setStore(PhotoFileStore * store)
+{
+    failPhotoOperation();
+    _store = store;
+}
+
+
+void AsyncDownloadPhotoTrigger::completePhotoWithURI(const QString & uri)
+{
+    // We can be called in two different scenarios:
+    // - Photo was triggered through "takePhoto" API function: In this case
+    //   there must be an active operation, but no download yet (because
+    //   we would otherwise already have initiated download). Initiate download
+    //   and track it.
+    // - "Unsolicited": Some photo was taken, but we did not ask for it. In
+    //   the latter case, also initiate download, but do not bother tracking.
+    if (_active_op && !_active_reply) {
+        _active_reply = _download_manager.get(QNetworkRequest(QUrl(uri)));
+        // NB: QNetworkReply has progress signals, could pass that on.
+    } else {
+        _download_manager.get(QNetworkRequest(QUrl(uri)));
+    }
+}
+
+void AsyncDownloadPhotoTrigger::completePhotoFailed()
+{
+    // See comments to completePhotoWithURI regarding this notification.
+    if (_active_op && !_active_reply) {
+        failPhotoOperation();
+    }
+}
+
+void AsyncDownloadPhotoTrigger::failPhotoOperation()
+{
+    if (_active_op) {
+        _active_op->finishFailure();
+        _active_op = nullptr;
+    }
+    if (_active_reply) {
+        _active_reply->abort();
+        _active_reply = nullptr;
+    }
+}
+
+void AsyncDownloadPhotoTrigger::downloadFinished(QNetworkReply * reply)
+{
+    if (reply == _active_reply) {
+        if (reply->error()) {
+            _active_op->finishFailure();
+        } else {
+            QByteArray data = reply->readAll();
+            QString id = _store->add(reply->url().fileName(), std::move(data));
+            _active_op->finishSuccess(id);
+        }
+        _active_op = nullptr;
+        _active_reply = nullptr;
+    } else {
+        if (!reply->error()) {
+            QByteArray data = reply->readAll();
+            QString id = _store->add(reply->url().fileName(), std::move(data));
+        }
+    }
+}
--- a/src/ui/toolbar/MainToolBar.qml
+++ b/src/ui/toolbar/MainToolBar.qml
@@ -162,6 +162,17 @@ Item {
                     }
                 }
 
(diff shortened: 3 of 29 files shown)


DonLakeFlyer:
To me a photo gallery is best left to the OS picture file system capabilities than bolting in to QGC. Most modern OS already have good picture viewing capabilty. Don't see why this would need to be part of QGC?

caleridas:
For desktop use that is true, filesystem browser will actually work better.

The intended use is in the field (possibly tablets / tablet pcs) where it not convenient to constantly switch back and forth between QGC and a filesystem view. The workflow this is targetted at is roughly that the drone is taking pictures, and operator wants to immediately see images while drone is still in the air, get a quick glance whether they are "good enough" or other shots need to be taken, tag them, discard useless ones. (NB that the UI as written here does not yet fully satisfy all of that).

The other part of the changeset is that this will actually make QGC download the pictures from the drone -- without this, the pictures are not even there for filesystem browser to deal with them.

src/QGCApplication.cc
```diff
@@ -558,7 +558,10 @@ bool QGCApplication::_initForNormalAppBoot()
     // Exit main application when last window is closed
     connect(this, &QGCApplication::lastWindowClosed, this, QGCApplication::quit);
 
-    _qmlAppEngine = toolbox()->corePlugin()->createRootWindow(this);
+
+    _qmlAppEngine = new QQmlApplicationEngine(this);
+    _qmlAppEngine->rootContext()->setContextProperty("multiVehicleManager", toolbox()->multiVehicleManager());
```
DonLakeFlyer: Why are you pushing multiVehicleManager into the Qml context. It is already available from QGroundControl.multiVehicleManager.

src/PhotoGallery/AbstractPhotoTrigger.h
```diff
@@ -0,0 +1,66 @@
...
+ */
+class AbstractPhotoTriggerOperation : public QObject {
+    Q_OBJECT
+public:
+    ~AbstractPhotoTriggerOperation() override;
+
+    /**
+     * \brief Whether this operation has finished.
+     */
+    virtual bool finished() const = 0;
```
DonLakeFlyer: The Coding Style for comments is to use ```///``` style comments not ```/* .. */```. Also don't add comments for things which are obvious by there method naming. Like here. Commenting that this represents "finished" doesn't add any value and just takes up room on the header file.

https://github.com/mavlink/qgroundcontrol/blob/master/CodingStyle.h
https://github.com/mavlink/qgroundcontrol/blob/master/CodingStyle.cc

DonLakeFlyer:
> The intended use is in the field (possibly tablets / tablet pcs) where it not convenient to constantly switch back and forth between QGC and a filesystem view.

In thinking about mobile use you can't switch away from QGC to another app on a mobile device anyway. So given that this makes sense. I get it now.

src/PhotoGallery/AsyncDownloadPhotoTrigger.cc
```diff
@@ -0,0 +1,146 @@
+#include "AsyncDownloadPhotoTrigger.h"
```
DonLakeFlyer: There should be the copyright header in all these new files.

src/QGCApplication.cc
caleridas: Removed, was not aware that this is exposed.

src/PhotoGallery/AbstractPhotoTrigger.h
caleridas: Reformatted for coding style.

I am not clear about the intended difference between /// and // -- I take it one is exported docstrings, the other isn't? In that case the way this is set now is as intended.

I did *not* follow the "_" for protected/private methods. This makes a mess as methods a) change visibility during development and b) via "using" in subclasses. The redudancy means too many pointless changes to me.

src/ui/toolbar/MainToolBar.qml
```diff
@@ -121,6 +121,18 @@ Item {
                 }
             }
 
+            QGCToolBarButton {
+                id:                 photoGalleryButton
+                anchors.top:        parent.top
+                anchors.bottom:     parent.bottom
+                icon.source:        "/qmlimages/PhotoGallery.svg"
+                visible:            QGroundControl.corePlugin.showAdvancedUI
```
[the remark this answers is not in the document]
caleridas: I am not clear about this or the intended meaning of "advanced" -- can you clarify? Basically this follows the pattern around it.

src/PhotoGallery/AsyncDownloadPhotoTrigger.cc
caleridas: Added to all files.

caleridas:
Updated as per review comments. Also, UI interactions are complete now (can select/delete photos, and can take photos directly from within gallery). Yes, UI can be made nicer, but want to get basic workflow integrated first.

src/ui/toolbar/MainToolBar.qml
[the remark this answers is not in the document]
dogmaphobic: QGC has a concept of "Advanced Mode". When advance mode is enabled (not the default) some features that are otherwise hidden become visible. This is used by custom builds of QGC, where certain things should be hidden from the "casual" user (things like dangerous parameters and such). In this case, this new Image Gallery feature does not need to be hidden, therefore, you can remove the `visible` attribute altogether.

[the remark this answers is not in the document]
caleridas: removed visibility spec

caleridas:
Screenshot showing two images selected, to be deleted using "trash can icon" in lower left. Lower right has icon to take new picture.

Screenshot_20191022_105226

DonLakeFlyer:
Can you just use a regular checkbox control for selection? No need to make up new ui things for that.

tilaktilak:
> The intended use is in the field (possibly tablets / tablet pcs) where it not convenient to constantly switch back and forth between QGC and a filesystem view.

> In thinking about mobile use you can't switch away from QGC to another app on a mobile device anyway. So given that this makes sense. I get it now.

Does it make sense having this feature for all the platform then or mobile only 
I am concern about how this "wall" view will render on a mobile device (10" screen or smaller), does it make sense to show only 1 picture at the time ?

I think we should re-use what's already in the code for the payload management and preview : 


    take photos on the drone
    download them into QGC
    view and manage them in a simple way in a photo gallery part

tilaktilak:
>   The intended use is in the field (possibly tablets / tablet pcs) where it not convenient to constantly switch back and forth between QGC and a filesystem view.

  >  In thinking about mobile use you can't switch away from QGC to another app on a mobile device anyway. So given that this makes sense. I get it now.

Does it make sense having this feature for all the platform then or mobile only
I am concern about how this "wall" view will render on a mobile device (10" screen or smaller), does it make sense to show only 1 picture at the time ?

Also, I definitively think we should re-use what's already in the code for the payload management and preview :

image
It's beneficial to have a picture preview during a mission to check if the camera settings are good ( Iso / Shutter) and we still want to show the FlyView as the drone is flying.
This window can be extended to have a deeper look at each individual images and trash them if needed.

From a UI-architecture point of view it make more sense to have this feature closer to the payload management rather than having it at the same level as General Settings/ Vehicle Setup/ Plan View/ Fly View. 

Can you be more clear on what you're intend to do here : 

- take photos on the drone : 
    Is that related to a flight plan generation with MissionItems ?
- download them into QGC 
   How does this works and what can of hardware setup do we need to benefit from it ? 

Also you said that this work is an exercise trying to be more modular, which is really great btw !. Can you give us a quick explanation of the new generic interfaces/classes/modules you're implementing.

caleridas:
> Can you just use a regular checkbox control for selection? No need to make up new ui things for that.

That would have been nice indeed :/ I tried a more "regular" approach using e.g. QScrolledArea etc. and could have put QCheckBox in -- but these are widgets, and QGC is QGuiApplication, meaning no QWidget. My attempt to simply change it to QApplication resulted in the GUI layout being messed up completely, so I gave up that attempt.

caleridas:
> > The intended use is in the field (possibly tablets / tablet pcs) where it not convenient to constantly switch back and forth between QGC and a filesystem view.
> 
> > In thinking about mobile use you can't switch away from QGC to another app on a mobile device anyway. So given that this makes sense. I get it now.
> 
> Does it make sense having this feature for all the platform then or mobile only
> I am concern about how this "wall" view will render on a mobile device (10" screen or smaller), does it make sense to show only 1 picture at the time ?

That renders okay, it actually works down to photo screen size. You can zoom between 'single picture' and 2, 4 and 6 column modes. (Actually, the operation is closely inspired by Android's photo gallery).

> Also, I definitively think we should re-use what's already in the code for the payload management and preview :
> 
> image
> It's beneficial to have a picture preview during a mission to check if the camera settings are good ( Iso / Shutter) and we still want to show the FlyView as the drone is flying.
> This window can be extended to have a deeper look at each individual images and trash them if needed.

I would consider this an orthogonal feature (also useful certainly) -- you will still need something that allows you to manage the entirety of images taken. Working with pictures one by one the user will almost certainly lose track of what was kept/discarded.

> 
> From a UI-architecture point of view it make more sense to have this feature closer to the payload management rather than having it at the same level as General Settings/ Vehicle Setup/ Plan View/ Fly View.
> 
> Can you be more clear on what you're intend to do here :
> 
> * take photos on the drone :
>   Is that related to a flight plan generation with MissionItems ?

Actually taking the pictures is supported already (manually, or maybe by missions), I just additionally wired this up into the gallery.

> * download them into QGC
>   How does this works and what can of hardware setup do we need to benefit from it ?

Camera manager informs us by mavlink message when a picture has been taken. In this message it specifies a URL through which the photo is accessible. QGC retrieves it from there. Not certain of specific hardware setup, some kind of camera on the drone would be nice obviously :)

> Also you said that this work is an exercise trying to be more modular, which is really great btw !. Can you give us a quick explanation of the new generic interfaces/classes/modules you're implementing.

Hm no this is not about implementing generic classes, but avoiding to interfere with the rest of QGC as much as possible:
- there is only one commit that touches core QGC, to handle the camera mavlink messages that were forgotten and exposes them as Qt signals
- in the actual implementation, there is only one class (in one file) that makes the "connection" between QGC and the new view, the rest is completely independent. Combined with QML this forms the wiring between the two (via signals and controller objects being hooked into).
- The "mental model" that I had while writing this is that it should be possible to make such a change entirely a compile time feature -- no ifdefs scattered through code, just include or don't include the relevant files in compile, and the feature is complete. This also means that there is no "static registration" anywhere in the main code etc.

tilaktilak:
Thanks for your explanation.
I am not quite sure I understand the use case though, as the "photo gallery" is more likely a tool used after the flight as a post-flight step. It required time and concentration to go through all images, tag them, trash them, ... In practice, on field during a mission the best you can do is take a picture, look at it, trash it if you want, continue your flight, take a picture, ...
For that purpose you need to preview a single image at the time the same way QGC preview video payloads, with probably a single shot button.

> I would consider this an orthogonal feature (also useful certainly) -- you will still need something that allows you to manage the entirety of images taken. Working with pictures one by one the user will almost certainly lose track of what was kept/discarded.

Ok, then this is definitively a post-flight operation. In that case you may want to download a bunch of pictures at the time from the camera ? 
    
> That would have been nice indeed :/ I tried a more "regular" approach using e.g. QScrolledArea etc. and could have put QCheckBox in -- but these are widgets, and QGC is QGuiApplication, meaning no QWidget. My attempt to simply change it to QApplication resulted in the GUI layout being messed up completely, so I gave up that attempt.

Not sure what is the problem here but you should at least reuse the sprites already present in the repo, like : 
https://github.com/Aeronavics/qgroundcontrol/blob/master/src/QmlControls/checkbox-check.svg
https://github.com/Aeronavics/qgroundcontrol/blob/master/resources/camera.svg
https://github.com/Aeronavics/qgroundcontrol/blob/master/resources/TrashDelete.svg

Also, does this gallery preview shots taken from a mission item ( on a survey ) ?
How everything behave with a lot of pictures at high resolution ?
When are the images removed from the tablet memory ? At the end of the mission ?

tilaktilak:
> Sure I can hook these up. However, they are monochrome without outline and by themselves will simply "disappear" on mathing (e.g. white) background which is not ideal form UI perspective. I can take these, produce outlined versions and use them, would you favor that?

> The "hand-drawn" things currently in code are of course only meant as stand-in, I am open to any suggestion how to do it.

Well you can't sanely write a function that will draw a trash icon and a camera . https://github.com/mavlink/qgroundcontrol/blob/24330f494bd947512919fb236b58093772a2aefb/src/PhotoGallery/PhotoGalleryView.cc#L35

The concern of having an icon "disappear" as already been addressed in QGC, we create a proper button with a background color that respect the theme, here an example : 

image

You can have a look at those files :
https://github.com/mavlink/qgroundcontrol/blob/master/src/QmlControls/QGCToolBarButton.qml
https://github.com/mavlink/qgroundcontrol/blob/master/src/QmlControls/QGCColoredImage.qml

src/PhotoGallery/AbstractPhotoTrigger.h
DonLakeFlyer: > I am not clear about the intended difference between /// and // -- I take it one is exported docstrings, the other isn't? In that case the way this is set now is as intended.

QGC coding style is to use ```//``` commenting versus ```/* ... */```. For doxygen the coding style is to use ```///```.

> I did _not_ follow the "_" for protected/private methods. This makes a mess as methods a) change visibility during development and b) via "using" in subclasses. The redudancy means too many pointless changes to me.

It is important to know whether a method is public or private from a readability standpoint to understand the scoping of the method without needing to look at the header. Same thing for "_foo" class member variable naming. Also I would consider moving a method from private to public requiring more work a good thing. Since it make you look at the code in a deeper way to make sure it's ok to move this to public.

caleridas: > > I am not clear about the intended difference between /// and // -- I take it one is exported docstrings, the other isn't? In that case the way this is set now is as intended.
> 
> QGC coding style is to use `//` commenting versus `/* ... */`. For doxygen the coding style is to use `///`.
> 
> > I did _not_ follow the "_" for protected/private methods. This makes a mess as methods a) change visibility during development and b) via "using" in subclasses. The redudancy means too many pointless changes to me.
> 
> It is important to know whether a method is public or private from a readability standpoint to understand the scoping of the method without needing to look at the header. Same thing for "_foo" class member variable naming. Also I would consider moving a method from private to public requiring more work a good thing. Since it make you look at the code in a deeper way to make sure it's ok to move this to public.

The class member variable naming is sensible, the visibility is not.

FTR, none of the 'major' coding styles is doing that (boost / standard C++, Google, Qt, ...) and some of them have published extensive rationale explaining the reasoning for choices. I am not clear what makes QGC "exceptional" or where some of the C++ experts behind those are missing crucial insight from QGC development. One particular thing for certain is that the convention is logically at odds with the language as any private/protected inheritance and/or using directives changing visibility render the name invalid. 

For the class 'user' it does not matter as the visibility is enforced by compiler without naming convention. For class 'implementor' it still does not matter as there is no difference in calling either, and the implementor needs to know about the detailed semantics of each called function anyways which cannot be conveyed in a single character.

There is a gradual line of moving between "member functions", "static member functions" and "free functions" where any such naming conventions is contributing useless red tape. (Free functions cannot be legally named with an underscore). There is also a gradual process in defining the member exposure of implementation(!) classes when working on unit test coverage, again adding nothing but red tape.

That said, I am going to change implementation according to style, but it is not going to change my opinion that it is entirel pointless.

DonLakeFlyer:
So I've been reading through this discussion. I get the concept of being able to "preview" and image take from the drone. But this implementation goes 10 miles beyond that. To me their are issues with:
* How you can download multiple full resolution images over even a fast link without it taking forever.
* Creating a huge piece of user interface to manage photos which doesn't really fall into something QGC should be doing.

Maybe the specific really high end hardware you are using and usage scenario make this a good feature. But that makes this a custom build feature for your specific build. I don't see it as a mainline QGC feature.

src/PhotoGallery/AbstractPhotoTrigger.h
DonLakeFlyer: I'll take all of that into account in considering a change to the coding style.

PR #7924, opened 2019-10-16
Fly: Disable Pause button if fixed wing on approach

DonLakeFlyer:
This prevents the user from possibly causing the plane to loiter much too low to the ground.


Antiheavy:
This is good. Thanks. I’ll add that the proper user action if on landing approach is to Abort and not Pause. This helps enforce that.

PR #7933, opened 2019-10-18
Toolbar: Indicator spacing

DonLakeFlyer:
Fix problem with extra empty space on iOs toolbar. Toolbar indicators now require a showIndicator property to determine whether the parent Loader item should be visible or not. Ths indicators themselves no longer set the visible property nor do they set width to 0 with invisible.


DonLakeFlyer:
@dogmaphobic Don't know if any of your custom builds have custom tooolbar indicators. If they do they will need to switch to this new mechanism to visibility.

PR #7936, opened 2019-10-19
Switch to Android API level 28 to meet Google Play requirements

DonLakeFlyer:


tilaktilak:
Shouldn't we change the Android API from qgroundcontrol containers as well ? https://github.com/mavlink/containers/blob/master/Dockerfile_qgc-build-android

DonLakeFlyer:
I not quite sure what that is. Maybe Jenkins? If it is, I continue to say that I just don't have time to keep two completely separate build systems up to date. It's hard enough to keep one working correctly. Travis and appveyor drop the actual releases so that's all I really care about now.

tilaktilak:
Alright, if there is no added value for Jenkins pipeline you're right we should simplify the CI.
 
It's the repo that holds dockerfiles with a "ready to build" environment. Those images are used for Jenkins at the moment but it can also be considered as a dev tool by itself. 

With a 1line command like : `docker pull mavlink/qgc-build-android`  you have a complete Qt/Android cross compile environment that can be used to develop on QGC, to build dependency libs for QGC on Android, generate Apk, run unit-tests, ...

Same deal with linux dev env : `docker pull mavlink/qgc-build-linux` 

IMHO those are valuable tools :-)

Anyway, it's a probably the subject of another discussion.

PR #7941, opened 2019-10-21
PX4: enable motor test UI

bkueng:
@DonLakeFlyer it only took the firmware several years to get it done :D
But here it is: https://github.com/PX4/Firmware/pull/13240

I changed the logic a bit. Previously, when moving a slider all the way down or when disabling the safety switch, the motors would still turn until the timeout triggered (of 1s). This does not inspire confidence, so I updated it to as long as the switch is enabled, the command is sent.

Remaining issues:
- QGC shows the wrong image: instead of x it showed me a + configuration on a quad.
- it works well on USB and wifi. But via telemetry (I used a holybro link), commands get buffered on the QGC side, and arrive with an increasing delay (when you wait a bit after enabling the safety slide, I experienced something like 20s).

### DIFF
--- a/src/AutoPilotPlugins/Common/MotorComponent.qml
+++ b/src/AutoPilotPlugins/Common/MotorComponent.qml
@@ -20,9 +20,10 @@ SetupPage {
     id:             motorPage
     pageComponent:  pageComponent
 
-    readonly property int _barHeight:       10
-    readonly property int _barWidth:        5
-    readonly property int _sliderHeight:    10
+    readonly property int _barHeight:           10
+    readonly property int _barWidth:            5
+    readonly property int _sliderHeight:        10
+    readonly property int _motorTimeoutSecs:    3
 
     FactPanelController {
         id:             controller
@@ -32,7 +33,13 @@ SetupPage {
         id: pageComponent
 
         Column {
-            spacing: 10
+            spacing: ScreenTools.defaultFontPixelHeight
+
+            QGCLabel {
+                text:       qsTr("Warning: Unable to determine motor count")
+                color:      qgcPal.warningText
+                visible:    controller.vehicle.motorCount == -1
+            }
 
             Row {
                 id:         motorSliders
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@bkueng I'm heading out of town for a number of weeks and won't have time to look at this now. But could you look at the APMMotorComponent.qml implementation. The UI there is new an much better/safer than the original one I implement ages ago. My plan was to make that as the starting point for a generic one or possibly just use it as the basis for the new PX4 one. I think it would be better to move this over to use for PX4 as well and tweak as needed.

DonLakeFlyer:
@bkueng Also big thanks for implementing this.

DonLakeFlyer:
ALso the APM version doesn't show airframe images. Mainly because it doing that correctly 100% of the time is incredibly difficult. Especially since config files can really do any sort of custom motor ordering they want. May add back at some point, but for now the testing part by itself is more imporant.

bkueng:
I tested the APM version, with the following remarks:
- it is considerably more intuitive if the motors update while dragging.
- if there is a timeout for safety, that is ok. But I can see cases where you want to just keep the motors spinning, and having that option would be useful.

> Mainly because it doing that correctly 100% of the time is incredibly difficult. Especially since config files can really do any sort of custom motor ordering they want. May add back at some point, but for now the testing part by itself is more imporant.

That is true. It would be useful for users to see it though (at least for the very common setups, like a quad).

DonLakeFlyer:
> * it is considerably more intuitive if the motors update while dragging.

I tried that and it ends up being difficult to not create a backlog of motor commands to the vehicle such that when you let go, they are still getting pumped to the vehicle. Which in turn causes a command to stop motors to be delayed. I tried various ways to make this work but it all got too complicated. Either causing delayed reponse to stopping motors or live reponse to dragging being to slow as to be not useful. So I opted for safety of always being able to quickly respond to motor stop.

DonLakeFlyer:
> * But I can see cases where you want to just keep the motors spinning, and having that option would be useful.

You are correct that I opted way towards the safety side. My take on this is if we get complaints we can consider adding. But it's scary to think you can't get the motors to stop. Maybe QGC crashes/hang at exactly the wrong time and so forth. Also full access to motor testing is still available from mavlink console for hardcore folks.

DonLakeFlyer:
I believe @dagar was the one who also originally complaining about the PX4 version being to lax on safety.

bkueng:
Ok I updated it to use the same logic as APM (except for `motorIndexToLetter`), which works as well.
I think we can still get the dynamic updating to work, for example by dropping old commands in the queue, as new ones come in.

PR #7944, opened 2019-10-23
rename collision prevention parameters to match Firmware

mrivi:
Sync collision prevention parameters with Firmware https://github.com/PX4/Firmware/pull/13135
FYI @baumanta


DonLakeFlyer:
Does this break backwards compatibility. Or was this old parameter name never released in a stable?

dogmaphobic:
> Does this break backwards compatibility. Or was this old parameter name never released in a stable?

It was certainly never used. I just added it to QGC for the first time not long ago.

mrivi:
@DonLakeFlyer @dogmaphobic can we get this in? This was never in a stable release

PR #7945, opened 2019-10-24
Cleanup map providers

cosmarc:
Cleaning up all map providers added by @tilaktilak . There were no changes in terms of behaviour, just performance and code style improvements. Main points:
- Always use `QStringLiteral` to generate the data for the string at compile time. Since there's a lot of string processing involved in building up the URLs 
- Use `override` in all derived classes to avoid re-declaring methods.
- Try to match as much as possible the code style
- Try to avoid creation of temporary/useless `QString` objects
- Remove useless includes
- Remove some patch file that got in by mistake (couldn't find any use of it)


src/QtLocationPlugin/GoogleMapProvider.cpp
```diff
@@ -105,37 +108,28 @@ void GoogleMapProvider::_tryCorrectGoogleVersions(
...
-                                 QNetworkAccessManager* networkManager) {
+QString GoogleStreetMapProvider::_getURL(const int x, const int y, const int zoom, QNetworkAccessManager* networkManager) {
     // http://mt1.google.com/vt/lyrs=m
-    QString server  = "mt";
-    QString request = "vt";
-    QString sec1    = ""; // after &x=...
-    QString sec2    = ""; // after &zoom=...
+    QString server  = QStringLiteral("mt");
+    QString request = QStringLiteral("vt");
+    QString sec1; // after &x=...
```
tilaktilak: Why removing the initialization ? Could be QStringLiteral("") ?
What is the motivation of passing everything in QStringLiteral ?

src/QtLocationPlugin/EsriMapProvider.cpp
```diff
@@ -1,61 +1,51 @@
...
-QString
-EsriWorldStreetMapProvider::_getURL(int x, int y, int zoom,
-                                    QNetworkAccessManager* networkManager) {
-    Q_UNUSED(networkManager);
-    return QString("http://services.arcgisonline.com/ArcGIS/rest/services/"
-                   "World_Street_Map/MapServer/tile/%1/%2/%3")
-        .arg(zoom)
-        .arg(y)
-        .arg(x);
+static const QString WorldStreetMapUrl = QStringLiteral("http://services.arcgisonline.com/ArcGIS/rest/services/World_Street_Map/MapServer/tile/%1/%2/%3");
```
tilaktilak: Why are you moving all URLs out of class scope ?

src/QtLocationPlugin/GoogleMapProvider.cpp
cosmarc: - There is no need for initialization of statically allocated `QString`  objects because the default constructor will be called anyways giving you [an empty and null QString](https://doc.qt.io/qt-5/qstring.html#QString). To have the initialization would cost us some extra processing in the QString internals which is not needed since in `GoogleMapProvider::_getSecGoogleWords` they get wiped again. 
- When you pass a `char[]` to methods that take `QString` as argument a temporary `QString` is created and that requires some memory allocations and conversions in the implementation of `QString` that are quite expensive. You can see this if you do some performance profiling using QtCreator, especially on the maps/terrain data part where we build all those urls for each tile.

src/QtLocationPlugin/EsriMapProvider.cpp
cosmarc: Just to avoid allocating 2 objects at every request instead of just one (because arg always returns a *copy* of the initial string). You can as well add them in the class as static members (in the header).

cosmarc:
@DonLakeFlyer @dogmaphobic any idea why CI takes so long?

dogmaphobic:
> @DonLakeFlyer @dogmaphobic any idea why CI takes so long?

Travis open source is quite slow. Priority is given to paid customers. The build machines are overtaxed and we only get a single build at a time (or something low like that). Might as well use Commodore 64s to build :P

cosmarc:
@dogmaphobic Would be nice if you have some time to shortly review the PR as well and get it in if everything is ok. @tilaktilak please have a look again. Thanks a lot!

cosmarc:
@dogmaphobic Is it still possible that after 5 days the CI didn't got through or it's a glitch and I should retrigger it somehow?

dogmaphobic:
No, it’s all good. The Jenkins build is not properly setup and you can ignore it for now. I have not however, had the time to go over this PR in detail.

PR #7947, opened 2019-10-24
Fix test for rally and geofence support.

dogmaphobic:
Both Geofence and Rally tests for support check not only the flight controller capability field but also the (MAVLink) protocol version. However, there was nothing keeping track of version changes. If you were to start QGC with no connections, QGC creates a dummy vehicle in order to handle mission planning. If later you were to connect to a real vehicle, the handing over between dummy and real vehicle was not updating the protocol version. You ended up with protocol version "0" and the test for Geofence and Rally would return false.

@DonLakeFlyer Do we really need to check protocol version if the capability flag is set? Also, that signal name (`Vehicle::requestProtocolVersion()`) is not exactly in line with the use in this context. I didn't spend a great deal of time trying to make sense of it though.


DonLakeFlyer:
> Do we really need to check protocol version if the capability flag is set?

Yes. Just because the vehicle supports fence/rally and/or mavlink 2 that doen't mean the link itself supports mavlink 2. There is also funkiness with how PX4/ArduPilot differ on reporting capability bits with respect to mavlink 2 being turned on for the vehicle which also causes this to be needed.

PR #7963, opened 2019-10-30
CameraManager: bug, cleanups in CameraPageWidget.qml

olliw42:
CameraPageWidget.qml is a bit of a wild coding. This PR attempts to streamline it a bit.

1) bug: QGCCameraControl.CAMERA_MODE_UNDEFINED is not existing, hence changed to QGCCameraControl.CAM_MODE_UNDEFINED.

(I must admit I'm "surprised" that such an error doesn't trigger any compiler warning/error)

2) QGCCameraControl.CAMERA_MODE_VIDEO,  QGCCameraControl.CAMERA_MODE_PHOTO used instead of plain 1 and 0

3) once _camera has been checked there is no need to do _dynamicCameras.cameras.get(_curCameraIndex).cameraMode, replaced by _camera.cameraMode

4) _camera && _camera && _camera.hasModes changed to _camera && _camera.hasModes

5) _cameraPhotoIdle renamed to _photoIdle:  Unlike the other _cameraXXXX variables which relate to the _camera.cameraMode variable, _cameraPhotoIdle relates to a _camera.xxxStatus variable, similar to e.g. _cameraRecording. Hence the naming is streamlined to match the other _camera.xxxStatus.

6) _canShoot appeared only twice and in both instances in combination with !_cameraModeUndefined, which was thus moved up, which makes the logic also cleaner and more transparent

7) variable definitions somewhat reordered to keep logically similar things together

these are all supposed to be NFC. I have tested and things work for me.

### DIFF
--- a/src/FlightMap/Widgets/CameraPageWidget.qml
+++ b/src/FlightMap/Widgets/CameraPageWidget.qml
@@ -32,22 +32,22 @@ Column {
 
     property var    _dynamicCameras:        activeVehicle ? activeVehicle.dynamicCameras : null
     property bool   _isCamera:              _dynamicCameras ? _dynamicCameras.cameras.count > 0 : false
+    property int    _curCameraIndex:        _dynamicCameras ? _dynamicCameras.currentCamera : 0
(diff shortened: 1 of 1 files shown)


dogmaphobic:
I'm not sure anybody has ever really used this CameraPageWidget. All development using the camera API has been done in custom builds (see https://github.com/mavlink/qgroundcontrol/tree/master/custom-example)

Thanks for cleaning it up :)

PR #7964, opened 2019-10-30
CameraManager: Add support for BATTERY_STATUS

olliw42:
some cameras are powered from a battery which may dry out during flight. It thus would be most convenient to get the battery charging state reported to and displayed by the CameraManager.

This PR is adding such support.

It currently listens to incoming BATTERY_STATUS messages, and displays the battery_remaining value as an integer percentage in the widget.

There could be future improvements. For instance, it would be nice to also have a graphical representation of the battery percentage, but I'm not yet familiar enough with QGC/Qt to know how to do that. Also, currently it expects the camera to voluntarily send out the BATTERY_STATUS messages (e.g. every 5 secs, or so). I think this makes sense and is fine, but one could consider requesting a stream (I think it doesn't make sense to periodically request it single shot). One also might issue a warning dialog then the battery level falls below a certain value. But I think it's a good start as it is.

Personally I consider this a great additional piece of information!

The PR also increases the font size of the text, I really found it extremely small. This certainly might be personal preference, so if you don't like it, only grab the first commit.

I've tested this and it works great for me.


dogmaphobic:
This is interesting. @hamishwillee, until now, no camera has ever reported its battery status (all cameras I am aware of run off the system battery). What @olliw42 did here seems quite clever. It might be a good idea to document this so this becomes the "standard" for camera battery status.

olliw42:
well, I made an effort to get something into the standard, https://github.com/mavlink/mavlink/issues/1202, but concluded there won't be a consensus easily, so did this out of desperation. At the end of the day I'm just interested in the feature, if it's standard or not. Glad to hear though you seem to like it.

hamishwillee:
So this gets battery status message associated with the camera component id and displays that in camera control dialog for the current camera? And presumably it works for any camera component that publishes a status, and isn't rendered if there is no battery information?

@dogmaphobic If you're proposing adding this to the MAVLink camera API docs then sure. Would be better if you took first crack at it though. OK?

olliw42:
qgc-battery

src/FlightMap/Widgets/CameraPageWidget.qml
```diff
@@ -67,15 +68,21 @@ Column {
...
+        font.pointSize: ScreenTools.defaultFontPointSize
         anchors.horizontalCenter: parent.horizontalCenter
         visible: _camera && _storageReady
     }
+    QGCLabel {
+        text: _camera ? qsTr("Battery: ") + _camera.batteryRemainingStr : ""
+        font.pointSize: ScreenTools.defaultFontPointSize
+        anchors.horizontalCenter: parent.horizontalCenter
+        visible: _camera && _batteryReady
+    }
```
dogmaphobic: `_batteryReady` already tests for `_camera`
```
_batteryReady:  _camera && _camera.batteryRemaining >= 0
```
Which makes this a bit redundant:
```
visible: _camera && _batteryReady
```
I would normally merge and deal with it later but I'm not planning on working on this for a while.

olliw42: nice 
the same holds for _storageReady, used just few lines up the same way. I grabbed the occasion to also correct this

PR #7965, opened 2019-10-30
CameraManager: bug in CameraPageWidget.gml, stream variable not defined

olliw42:
a variable _streamingEnabled is used, but isn't declared anywhere.

is clearly a bug. It also produced a error message in the output, and there also was a spurios gray window showing up briefly when one displayed the video/photo settings, which empirically doesn't show up now anymore.

I couldn't figure out for sure how this value should be set, so I simply declared it and set it to false. But left a comment that this needs to be determined, and added what is done in the VideoPageWidget.qml, in case this might give a hint.


src/FlightMap/Widgets/CameraPageWidget.qml
```diff
@@ -42,6 +42,7 @@ Column {
     property real   _labelFieldWidth:       ScreenTools.defaultFontPixelWidth * 30
     property real   _editFieldWidth:        ScreenTools.defaultFontPixelWidth * 30
     property bool   _communicationLost:     activeVehicle ? activeVehicle.connectionLost : false
+    property bool   _streamingEnabled:      false //TODO: determine what it should be, VideoPageWidget.qml does QGroundControl.settingsManager.videoSettings.streamConfigured
```
dogmaphobic: Looks like it came from here:
https://github.com/mavlink/qgroundcontrol/blob/master/src/FlightMap/Widgets/VideoPageWidget.qml#L37

This video section under settings was copied and pasted from `VideoPageWidget.qml`

https://github.com/mavlink/qgroundcontrol/blob/master/src/FlightMap/Widgets/CameraPageWidget.qml#L392-L434

olliw42: not sure I understand your message
you suggest I should use the same line as in your first link? 
I'm happy to copy it :)

dogmaphobic: No (or maybe). I was just pointing out where it came from. I don't remember the order but I think the video widget came first (along with those video specific settings). When the other one was created, some of the video settings were brought along but it missed the `_streamingEnabled` definition.

olliw42:
rebased on master, and variable deleted (I somehow did forget doing this in this PR)

olliw42:
any reason this simple PR is stuck?
:)

PR #7968, opened 2019-11-01
Parameter write progress bar

dakejahl:
Added in logic to display the progress bar when writing parameters from QGC to the vehicle via MAVLink.

This is especially useful when writing a large parameter file to the vehicle for the first time or when transferring over an unreliable data link.

Tested successfully on Ubuntu 18.04.

I did notice every time I did a `Load from file...`it would give me this error, however this happened before my changes, so I imagine it is probably related to the build being a debug build, and I am running using *qgroundcontrol-start.sh*

The presence of the error did not effect anything, all parameter writes were successful.

Selection_035

### DIFF
--- a/src/FactSystem/ParameterManager.cc
+++ b/src/FactSystem/ParameterManager.cc
@@ -274,6 +274,17 @@ void ParameterManager::_parameterUpdate(int vehicleId, int componentId, QString 
         _setLoadProgress((double)(_totalParamCount - readWaitingParamCount) / (double)_totalParamCount);
     }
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@dakejahl I don't think this will work quite correctly in the case where you have both reads and writes pending which is possible. I think the read and write progress bar update code needs to be merged together into a single thing.

DonLakeFlyer:
I've been putting some more thought into this and supporting this correctly is actually quite complex. I'm working on it now.

PR #7969, opened 2019-11-01
PageView: white wheel, smaller

olliw42:
The PageView currently shows a black gear symbol in the header combo box, which is very had to see given the dark gray background of the combo box.

This PR changes it to a white gear symbol, and also reduces the size of it for better style.

### DIFF
--- a/src/QmlControls/PageView.qml
+++ b/src/QmlControls/PageView.qml
@@ -33,10 +33,10 @@ Rectangle {
             anchors.leftMargin:     _margins
             anchors.left:           parent.left
             anchors.verticalCenter: parent.verticalCenter
-            source:                 "/res/gear-black.svg"
+            source:                 "/res/gear-white.svg"
             mipmap:                 true
             width:                  parent.height -(_margins * 2)
-            sourceSize.width:       width
+            sourceSize.width:       width * 0.75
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Changing it to black isn't correct either. This should be a QGCColoredImage using the palette color of windowText. That way it works with palette changes.

@dogmaphobic We need an .svg gear we can color. Not sure what that means from an svg standpoint. I don't think we can color the black or white one.

dogmaphobic:
>We need an .svg gear we can color. Not sure what that means from an svg standpoint. I don't think we can color the black or white one.

You can color either one using `QGCColoredImage`. I suggested that change because both svg files are there. Yes, the item can be replaced with a `QGCColoredImage` instead.

DonLakeFlyer:
> You can color either one using `QGCColoredImage`

Hmm, I tried that and it didn't work for some reason. I"ll try again.

DonLakeFlyer:
Figured it out. QGCColoredImage has some differences with respect to needing to set both width/height

olliw42:
thx!

DonLakeFlyer:
This has been annoying me for ages, but just never got around to i

Issue #7869, opened 2019-10-05
Kind reminder: CameraManager VideoStop still broken


olliw42:
the issue with the camera manager's video start/stop described in https://github.com/mavlink/qgroundcontrol/issues/7605 still exists in todays daily build

it unfortunately makes QGC of quite little use for folks wanting to use the camera manager

for your convenience I copy over the description and close the older issue:

When in video mode, pressing the red round button sends a startVideo() and the button goes square as usual, but the button gets GRAY and doesn't remain red. From then on it's not possible to get QGC emit a stopVideo() by clicking it. In photo mode it works.

olliw42:
@dogmaphobic @MatejFranceskin @hamishwillee 

I've added names which I believe I've seen before here in this context, sorry if I pinged you incorrectly

I'm confused.

The above bug is now existing for 3 month. It is a tiny bug, but it has the drastic consequence that the camera manager on the whole is essentially useless. 

I had now two explicit cases where folks wanted to use the camera manager but went away when I told them that they have to use a daily build of 3 months ago. This furthers my own feeling that it is very sad that QGC cannot be used for that purposes.

if possible, I really very much would appreciate feedback - even just one sentence - as regards to what's up here. 

Is it only and only me seeing this bug? Is the camera manager not used by anyone except of me? Is there no interest anymore in the camera manager from you side? ...? ...? ...?

thx in advance.

dogmaphobic:
Sorry but I've been out for most of the past several weeks and buried in other stuff in between. All of the above has been working fine in all devices/setups I have access to, which would indicate some edge case or configuration issue. With that said, I will see if I find a slot of time this week to go over your reports and try to see what could be going on. I have a suspicion the camera is not sending back camera status messages to QGC.

olliw42:
first, many thx for your feedback, really appreciated.

I'm using this on a Win7 computer, and I simply run the daily-installer, clicking yes everywhere, i.e also install the drivers etc.. 
It is not immediately obvious to me what I could do wrong.
Also, it was working fine before the QtQuick conversion (not sure I use the right words, but few months back you changed something like from QtWidgets to QtQuick or so, before that it was all working fine, after that it stopped working fine).

The camera is sending back camera status messages. I can see that also in QGC when I enable the respective log. And as said, when I use the "old" daily build of 5 months ago it's all working fine. So, I would rule out that possibility, and effectively would rule out any such extrinsic issues, but believe it's inside QGC.

[...]

PR #7970, opened 2019-11-01
CameraManager: Remove two bugs, resolve issue #7689

olliw42:
this resolves two bugs:

1) resolves issue #7869: Video stop broken

The MouseArea of the shutter button is enabled/disabled depending on the boolean _canShoot, which was determined as

`property bool   _canShoot:              !_videoRecording && _cameraPhotoIdle && ((_storageReady && _camera.storageFree > 0) || _storageIgnored)
`

Therefore, when a video was recording, and _videoRecording = true, the MouseArea is disabled, and cannot be clicked anymore, which is however what one has to do to stop the video. 

The same issue also exists for taking photos, where it however only bites for timelapses, not for single shot.

The PR corrects this by removing the dependence on _videoRecording and _cameraPhotoIdle.

Note: The bug is IMHO very clearly due to a wrong logic. It was previously argued that things are working fine for many other systems and that issue #7869 is an edge case or issue with my specific configuration. This seems to be not the case. Also, very strangely I admit, things seemed to have worked several months back, despite the same logical flaw being present. I really don't understand this, and only can speculate. But as said, I think the above logic is just wrong. And correcting it as described makes things nicely working for me.

2) Photo Idle

photo idle was determined as 

` property bool   _photoIdle:             _camera && _camera.photoStatus === QGCCameraControl.PHOTO_CAPTURE_IDLE`

However, the photo status can also be >= QGCCameraControl.PHOTO_CAPTURE_LAST, it especially is set to PHOTO_CAPTURE_STATUS_UNDEFINED at startup which obviously also indicates an idle photo. The PR corrects the condition to take this into account. This avoids strange conditions during the initialization state of the CameraManager.

COMMENT: This PR has also applied the PRs  #7963 and #7965.


dogmaphobic:
This will take a while to digest. I will take a look tomorrow. I have not kept track of this camera widget. Instead, I've been using the custom build version:
https://github.com/mavlink/qgroundcontrol/blob/master/custom-example/res/CustomCameraControl.qml

olliw42:
> Instead, I've been using the custom build version:
https://github.com/mavlink/qgroundcontrol/blob/master/custom-example/res/CustomCameraControl.qml

ahhh ... so it would have been more convenient if I would have based on this too?
If it's so to say the better "default", wouldn't it be more convenient to kind of have that example merged/moved over to the default place?
Your's look more advanced on a quick glance, and it's maybe not convenient to have two things to be worked on as default

olliw42:
PR #7971 added to here, as suggested over there

dogmaphobic:
> ahhh ... so it would have been more convenient if I would have based on this too?
> If it's so to say the better "default", wouldn't it be more convenient to kind of have that example merged/moved over to the default place?
> Your's look more advanced on a quick glance, and it's maybe not convenient to have two things to be worked on as default

No, they are both equality important. I just have neglected the camera widget in the "normal" QGC given that no one seemed to be using. The custom build produces a different UI/UX for QGC. It's more camera centric and meant to be user focused. The "normal" QGC is primarily a development tool with a plethora of options that don't exactly concern to a user interested in simply flying (and taking pictures/recording videos).

olliw42:
many thx for the clarifications

>  The "normal" QGC is primarily a development tool with a plethora of options that don't exactly concern to a user interested in simply flying (and taking pictures/recording videos).

well, but the "normal" QGC is what is mostly used by "average-Joe" ! At least anyone I've heard off using QGC doesn't use a custom made one. And also "average-Joe" wants to benefit from cool stuff ... :)

anyway, I guess, I may conclude that it's not totally useless to go on with spending effort on the normal camera support :) :)
(this PR is then actually kind of a must, since the can't-stop-Video issue really made things useless)

BTW: How can I compile the said custom example? I couldn't figure that out. I would be curious to see this other UI.

dogmaphobic:
> well, but the "normal" QGC is what is mostly used by "average-Joe" ! At least anyone I've heard off using QGC doesn't use a custom made one. And also "average-Joe" wants to benefit from cool stuff ... :)

It's not yet complete. Eventually I want to have both versions built and available.

>BTW: How can I compile the said custom example? I couldn't figure that out. I would be curious to see this other UI.

Just rename "custom-example" to "custom", delete your existing build directory and build it all from scratch. The build automatically picks up the "custom" directory and integrates it into the build. I normally just keep a symlink to it (`ln -s custom-example custom`).

olliw42:
since this has become kind of the "bug" thread, I've added removal of another (tiny) bug 
:)

(_isCamera and _camera are not true for the same conditions, hence when _camera is used one should check for _camera, not _isCamera)

olliw42:
> Just rename "custom-example" to "custom", delete your existing build directory and build it all from scratch. 

just as feedback: I did as you described and this nicely made the compile run thorugh. Thx for the explanation. However, the exe doesn't work. When done from within QtCreator it complains that the exe can't be accessed because of privileges or because it is not existing. When started from the explorer QGC pops up with a somewhat screwed layout, but disappers after some seconds with an error message. So doesn't appear to work on Win7. I thought I just let you know. :)

dogmaphobic:
You need to restart QtCreator when switching from "normal" to "custom". As to starting QGC from Explorer on Windows, this only works for release builds.

olliw42:
> You need to restart QtCreator when switching from "normal" to "custom". As to starting QGC from Explorer on Windows, this only works for release builds.

ah, ok ... I'll try that :). Thx for explaining. You may realize I'm not very good with this build system stuff in general :) :)

Issue #7471, opened 2019-05-23
Camera Manager: Collection of behaviors to imporve


olliw42:
I've made much progress with implementing a CameraManager support into the STorM32 gimbal controller to the point that it is practically usable, and in the practical tests I start to observe various behaviors of the camera manager GUI implementation which - IMHO - could be improved for better user friendliness. Instead of raising an issue for each, I thought I rather collect them here (pl tell if you prefer single issues for each).

<strike>**1) Stream Selector field shown even if no STREAM capability**
See #7450.</strike>
EDIT: resolved

**2) In photo mode single shot or interval mode should be indicated.**

It's easy to forget what the last setting for the photo mode (single or interval) was, and not remembering it can lead to unexpected behavior. I thus suggest to have a graphical element on the GUI's "frontpage" showing this. E..g. this often seen "multiple-clocks" graphics.

I would suggest to extend this to a "submode" concept. Cameras like the GoPro e.g. support several submodes for each video and photo, single or interval would then be seen as submodes. For  the same reasons it would be highly useful if the camera's submode would be easily visible to the user. This may cause an issue for submodes for which the GUI/GCS doesn't have a graphics to represent, but this should be solvable, and be it that in this case the GUI doesn't show a graphics. 

<strike>**3) Don't allow changes of camera mode (photo vs video) while a video is recording or a a photo interval is running.**
Currently that's possible, and results in strange states for the red button.</strike>
EDIT: done

<strike>**4) In photo mode provide a graphical feedback when the next photo can be taken.**
In photo mode with single shots the response is very slow, i.e. it takes few seconds before the next photo can be taken. If a hit on the red button was successful currently can only be inferred indirectly through an increasing number of photos. I suggest to provide some graphical mechanism to show to the user when it is ready for the next shot. E.g., the red button could get square, and round again when ready for the next shot.</strike>
EDIT: resolved

**5) Clarify the behavior and relationship as regards CAM_MODE and CAMERA_CAP_FLAGS.**

[...]

PR #7971, opened 2019-11-01
CameraManager: disallow mode change during shooting

olliw42:
Currently it was possible to change the camera mode while the video is recording or a photo timelpase is running.

This PR prevents this by disabling the respective MouseArea when a video or photo is running.

Resolves point (3) in issue #7471.
In combination with PR #7689 it also resolves point (4) in issue #7471.


dogmaphobic:
You might as well consolidate this into https://github.com/mavlink/qgroundcontrol/pull/7970

olliw42:
if you prefer me doing this
I've split because the other PR is about bugs, and this here about a mere issue
but I'm happy to do :)

olliw42:
added to PR #7970 
thus I close

PR #7977, opened 2019-11-05
ParameterEditorController: MAVLink COMP ID names for components

olliw42:
This shows the MAVLink name for a component instead of "Component XXX" in the parameter editor window. See the screenshot below.

I'm not sure this is the sort of thing you guys like, but I like it (obviously)(LOL). 

The mavlink hash could probably go to a more central location. Pl advise.

cheers, Olli

qgc-compidname

### DIFF
--- a/src/QmlControls/ParameterEditorController.cc
+++ b/src/QmlControls/ParameterEditorController.cc
@@ -19,6 +19,45 @@
 
 #include <QStandardPaths>
 
+
+const QHash<int, QString> _mavlinkCompIdHash {
+    { MAV_COMP_ID_CAMERA,   "Camera1" },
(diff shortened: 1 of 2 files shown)


olliw42:
depending on which approach you prefer more, this could be replaced by PR #7985.

DonLakeFlyer:
Can you change the wording to be something like this: "Comp: Camera6 (105)". This makes it a bit more clear to the user that these are coming from a separate component.

olliw42:
sure.
I will incorporate this once it is decided if either this PR or PR  #7985 is preferred
:)

olliw42:
closed in favor of PR #7985

Issue #7962, opened 2019-10-30
Advanced Tuning Setup Crashes QGC


dogmaphobic:
Adding here to keep track.
Hitting the _Advanced_ checkbox causes QGC to segfault. It's somewhere deep within the QtChart module.

<img width="1026" alt="Screen Shot 2019-10-30 at 12 22 12 PM" src="https://user-images.githubusercontent.com/749243/67877718-7f948880-fb10-11e9-974c-3576fc21458a.png">

<img width="653" alt="Screen Shot 2019-10-30 at 12 23 50 PM" src="https://user-images.githubusercontent.com/749243/67877732-858a6980-fb10-11e9-81f1-38b6e3853ce8.png">

stefandunca:
Fix here: https://github.com/mavlink/qgroundcontrol/pull/7978

DonLakeFlyer:
Reopening since I backed out the QGuiApplication->QApplication change since it caused other bad problems.

DonLakeFlyer:
Also advanced tuning is removed from the build until a workaround is figure out

DonLakeFlyer:
@dogmaphobic Is tried doing what the Qml Charts docs say to switch QGuiApplication->QApplication in a small test app and it worked. Didn't you say this didn't work for you?

dogmaphobic:
> Didn't you say this didn't work for you?

If solves the QtChart problem but causes MessageBox to close the entire application on exit.

DonLakeFlyer:
But I though you made a small test app and were able to reproduce the MessageBox problem. I can't reproduce the MessageBox problem in a small test app. Hence it look like it has someting specific to do with QGC. Which I think it's means there wmay be some way to make QApplicaiton work.

dogmaphobic:
Weird... I just loaded the small test project and indeed, it all works. I could swear it was closing the app when I clicked the OK button within the MessageBox. This needs some digging...

DonLakeFlyer:
@dogmaphobic Fixed now?

PR #7978, opened 2019-11-05
Fix enabling Advanced Tuning crash

stefandunca:
Fix for https://github.com/mavlink/qgroundcontrol/issues/7962

It seems that QtCharts require `QApplication` instance instead of current used `QGuiApplication` even that we use only QML controls without initializing QWidgets. See `Note: ` for the hint: https://doc.qt.io/qt-5.12/qtcharts-index.html

Then crash stack in QT looks unrelated to the issue


DonLakeFlyer:
Why not QGuiApplciation for everything then?

PR #7982, opened 2019-11-07
Simplify mission item list

tilaktilak:
Some few changes to improve and simplify the Mission Item List :

* Add a trash icon to remove easily an item
* Insert on the current element of the list and not at the end ( "insert Waypoint" / "insert Pattern" from the Hamburger are issued by the left-hand toolbar )
* Hamburger removed as every features are now redundant ( Change Command issued by clicking on the Item name, "Show all values" checkbox added at the end of the Item, Edit position accessible through Show all values, Item number in Item name )

Here how it looks : 

Peek 2019-11-07 13-57


DonLakeFlyer:
I like exposing a simpler way to delete but it causes too many problems:
* Many people don't find the ability to click on the command name to change commands. This is why it is in the Hamburger which for a number of users is the only place they know it exists.
* "Show all Values" which is a power user from hell feature is now exposed to everyone. For 99.9% of the user base we don't really want people seeing or using this. It also make the editors more complex than they already are which is a problem.
* I specifically removed the item numbers from the editors to remove generally useless clutter. I don't think adding them back is a positive change.
* Edit Position is now a missing feature

Things I like:
* I like the insert from current selection. It think that makes a lot of sense.

It may instead be possible to add a delete icon to the left instead. But there is already an indicator there for item completion. I'm also not really that sure how often delete is actually used such that promoting it to top level ui is iumportant.

Antiheavy:
I'm a big fan of anything that can simplify the mission item list.

I like the idea of a trash/delete icon at the top level.  I think people make enough mistakes when mission planning that this is useful and more satisfying to the user.  In fact, in our experience people use the hamburger menu primarily to find the delete button, secondarily to find insert, and rarely for all the other advanced features.  Even the knowledge that delete is accessible right there makes mission planning less stressful to some users.

>     * Many people don't find the ability to click on the command name to change commands. This is why it is in the Hamburger which for a number of users is the only place they know it exists.

Many of the users I've worked with don't find the hamburger "discoverable".  Meaning they have no clue that those 3 lines means a menu they can click on.  I also get it, that many experienced users are now very used to that menu and may find a change difficult.  
Maybe an edit icon added to the end of the command name button to imply clicking here allows you to edit?  Maybe there is a way to combine everything that is currently in the hamburger menu into the change command menu as one giant edit menu?
image



>     * "Show all Values" which is a power user from hell feature is now exposed to everyone. For 99.9% of the user base we don't really want people seeing or using this. It also make the editors more complex than they already are which is a problem.

yes, this very important power user feature should be hidden away.

>     * I specifically removed the item numbers from the editors to remove generally useless clutter. I don't think adding them back is a positive change.

Yes, the mission item numbers start getting confusing as soon as you start adding complex items or have hidden (non position) items.  Better that they are not shown on the top level in my opinion.  Or maybe there is a fancy way to show them during certain situations where they are useful and hide them most other times?

tilaktilak:
Thanks for your feedback.

The idea of my work here is to propose something with less duplication and less hidden tools.

- I agree with the fact that "show all values" needs to be hidden somehow, that makes sense. 

    Can't it be an easter egg kind of thing if it's an advanced user feature? 

- Agree with the number of MissionItem especially if that was removed for a reason before. 

- Also, I do think that "Delete" should be exposed more clearly as it's the second tool you use after insert. 

- > Edit Position is now a missing feature

    You can already change the position manually on the "Show All Value" page


- >     * Many people don't find the ability to click on the command name to change commands. This is why it is in the Hamburger which for a number of users is the only place they know it exists.

> 
> Many of the users I've worked with don't find the hamburger "discoverable". Meaning they have no clue that those 3 lines means a menu they can click on. I also get it, that many experienced users are now very used to that menu and may find a change difficult.

Things have to be discoverable anyway. What about using the same upgrade that was made on Flight modes? 
image

hamishwillee:
FWIW I like delete very visible. Even more, I'd like to be able to undo 2 or 3 changes because I'm often accidentally clicking on the map.

DonLakeFlyer:
> I like the idea of a trash/delete icon at the top level.

It makes sense to bring Delete to the front somehow if it truly gets used a decent amount of the time. If not, then I think it's a mistake to bring it to the front. I just adds additional clutter for an infrequently used command. It seems like @Antiheavy end users experience leans towards delete does get a decent amount of usage.

> Yes, the mission item numbers start getting confusing as soon as you start adding complex items or have hidden (non position) items. Better that they are not shown on the top level in my opinion. Or maybe there is a fancy way to show them during certain situations where they are useful and hide them most other times?

> I like numbers to, because that provides easy way to locate same item on the map. I guess I could click the item on the map to open the editor, but it isn't quite as natural to me.

@hamishwillee I think you haven't been running daily builds for a while because there are no longer numbers anywhere in the ui (except for the hamburger).

> What about using the same upgrade that was made on Flight modes?

Yeah, i've considered that. Haven't gone there yet because it's not actually a combobox it leads to a dialog. But maybe that's not the end of the world.

> * You can already change the position manually on the "Show All Value" page

Changing position from Show All Values is not equivalent to what you can do from Edit Position (Lat/Lon, UTM, Set to Vehicle position, ...)

> Maybe there is a way to combine everything that is currently in the hamburger menu into the change command menu as one giant edit menu?

Ick. That's just a bit too strange.

My ultimate conclusion is two things:
* The hamburger can't go away since there is too much stuff in there with no other place to go.
* There should be room from a Delete icon for selected items. Some of the recent changes to the item ui makes a clearer place to put it to the left if the command name.
* The command name as a Button which allows you to change command is not clear to most people that it can be clicked.

Let me see what I can some up with.

DonLakeFlyer:
This has all be override bvyt other changes.

PR #7984, opened 2019-11-07
Filter mavlink messages by component ID

dogmaphobic:
You can now filter the messages in the MAVLink inspector by Component ID:

Screen Shot 2019-11-07 at 3 15 18 PM

The component ID is now shown in the list in order to identify a message when multiple component ID are being displayed at once:

Screen Shot 2019-11-07 at 3 15 29 PM

Note that the filter is only available if more than one Component ID is actually found:

Screen Shot 2019-11-07 at 3 18 17 PM


olliw42:
great!

wouldn't it be good to not only see the number in the combo box but also/instead the MAVLink default ID name/MAV_TYPE, e.g. instead of only 154 to display 154-GIMBAL1, and instead of 1 to display 1-QUADROTOR (if it is a quadrotor)?

PR #7985, opened 2019-11-08
ParameterManager, ParameterEditorController: Nice category & groups for components

olliw42:
Description:

this PR provides a better presentation  of the parameters of components in the `Parameter` window. It replaces the title for a component by the MAVLink name of the component if such a name is available, e.g. `Gimbal1` instead of `Component 154`.  This PR can also group the parameters according to their first characters, as it's common MAVLink practice, as compared to the previous grouping into one section `All`. It makes the code, I believe, generally more extensible.

This PR from the user perspective accomplishes essentially the same thing as the recent PR #7977, but does it in a different way (vide infra).

_Comment:_ This PR thus could (should?) replace the recent PR #7977. I'm not sure however about your polices and desired practices and hence keep it "alive". Pl advise.

What's the point:

Currently, the handling of the parameters is quite asymmetric for the vehicle and the components. The vehicle's category&groups are build and handled in the `ParameterManager` class, while the component's category&groups are build and handled in the `ParameterEditorController` class. The result is a not so nice code since what is essentially the same thing is done very differently in the code. The recent PR #7977 stays within that architecture. 

This new PR moves the handling for components also into the `ParameterManager` class, and uses the same structures and tables as are used for the vehicle. The result is what I consider a cleaner since more symmetric and more canonical architecture, which also I believe reflects itself in a more natural code.

Function-wise it at first may seem to not achieve much more than the old PR, however the new architecture is much better suited for extensions. E.g., it just needs just a simple code change in one place to modify the grouping for the components. 

Looking forward to hear your opinions. Cheers, Olli

PS: The PR also renames `name` in few places to `paramName`, for more consistent code.


DonLakeFlyer:
@olliw42 Do you have some before/after screenshots?

olliw42:
sure, see the picture below

if one changes these lines to https://github.com/mavlink/qgroundcontrol/pull/7985/files#diff-2d238aec2be2b9c6088dd17733f68679R582-R587
to the single line `componentCategoryMap[category]["All"] += paramName;` then it will look EXACTLY like the screenshot shown in PR #7977. I think the easiness with which such changes can be made, which would be cumbersome in the current architecture, reflects and demonstrates the better architecture of this here.

qgc-compid2

src/FactSystem/ParameterManager.cc
```diff
@@ -527,20 +563,66 @@ QStringList ParameterManager::parameterNames(int componentId)
...
+    // Must be able to handle being called multiple times
+    componentCategoryMap.clear();
+
+    // Fill parameters into the group determined by param name
+    for (const QString &paramName: _mapParameterName2Variant[componentId].keys()) {
+        int i = paramName.indexOf("_");
+        if (i > 0) {
+            componentCategoryMap[category][paramName.left(i)] += paramName;
+        } else {
+            componentCategoryMap[category]["Misc"] += paramName;
```
DonLakeFlyer: Missing a tr()

src/FactSystem/ParameterManager.cc
```diff
@@ -527,20 +563,66 @@ QStringList ParameterManager::parameterNames(int componentId)
...
+    }
+}
 
-    for (const QString &name: _mapParameterName2Variant[_vehicle->defaultComponentId()].keys()) {
-        Fact* fact = _mapParameterName2Variant[_vehicle->defaultComponentId()][name].value<Fact*>();
-        _defaultComponentCategoryMap[fact->category()][fact->group()] += name;
+QString ParameterManager::getComponentCategory(int componentId)
+{
+    if (_mavlinkCompIdHash.contains(componentId)) {
+        return QString("Component %1  (%2)").arg(_mavlinkCompIdHash.value(componentId)).arg(componentId);
```
DonLakeFlyer: Should be ```tr``` not ```QString```

src/FactSystem/ParameterManager.cc
```diff
@@ -1439,7 +1521,7 @@ void ParameterManager::_loadOfflineEditingParams(void)
     }
 
     _addMetaDataToDefaultComponent();
-    _setupDefaultComponentCategoryMap();
+    _setupDefaultComponentCategoryMap(); //TODO: check if this is really only for the default, or needs to be done for all components !!
```
DonLakeFlyer: This is correct. Offline editing params will never have params from anything other than the autopilot component. Comment can be removed

DonLakeFlyer:
@olliw42 All good stuff. This is the way to do it. Fix up the comments and I'll merge. Thanks

olliw42:
thx for reviewing and commenting.
all comments adopted, code compiled and tested.

olliw42:
you are happy with the location of the _mavlinkCompIdHash hash in ParameterManager.cc?

having spend more time on the QGC code base it seems to me it might be better located higher up the food chain, e.g. in Vehicle.cc. Indeed UAS.cc has a very similar need, see
https://github.com/mavlink/qgroundcontrol/blob/master/src/uas/UAS.cc#L143-L172. 
Both UAS.cc and ParameterManager could access it then via their _vehicle variable. Would avoid duplication and bring some unification, and it might be useful elsewhere too.

In case you would agree, should I add it to this PR or would you want me to do it in a separate PR?

DonLakeFlyer:
> you are happy with the location of the _mavlinkCompIdHash hash in ParameterManager.cc?

Good enough for now. In reality it should be at global scope somewhere. That referenced UAS code is some old cruft that is no longer used I believe. I have been slowly getting rid of UAS, rewriting what is used and moving to Vehicle. It's been a multi-year journey that still isn't finished :)

olliw42:
> Good enough for now. 

ok. Great.

> That referenced UAS code is some old cruft that is no longer used I believe.

I've noted that it's being worked towards getting rid of it. However, it is actually still in use, the mavlink messages are passing by it

DonLakeFlyer:
> passing by it

Passing by it sure. But I don't think the information is used anywhere any more. If it ever was.

PR #7989, opened 2019-11-08
cmake: add new sources and delete unused SurveyMissionItem

dagar:


dagar:
Cmake is working again. I'll take another pass and fix the android container separately. https://github.com/mavlink/containers/pull/10

image

PR #7998, opened 2019-11-08
We can only filter by component ID

dogmaphobic:
Reverting the changes done in https://github.com/mavlink/qgroundcontrol/commit/598d1464314c5ad4574c087d0b4e588bf99fde2f

When handling MAVLink messages within the camera manager, we must filter them based solely on the component ID.

### DIFF
--- a/src/Camera/QGCCameraManager.cc
+++ b/src/Camera/QGCCameraManager.cc
@@ -68,7 +68,8 @@ QGCCameraManager::_vehicleReady(bool ready)
 void
 QGCCameraManager::_mavlinkMessageReceived(const mavlink_message_t& message)
 {
-    if(message.sysid == _vehicle->id()) {
+    //-- Only pay attention to camera components, as identified by their compId
+    if(message.sysid == _vehicle->id() && (message.compid >= MAV_COMP_ID_CAMERA && message.compid <= MAV_COMP_ID_CAMERA6)) {
         switch (message.msgid) {
             case MAVLINK_MSG_ID_CAMERA_CAPTURE_STATUS:
                 _handleCaptureStatus(message);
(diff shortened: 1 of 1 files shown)


olliw42:
why?

olliw42:
I think I understand where your confusion lies: 

the filtering in _handleHeartbeat() is perfectly fine, however what one has to do in addition is a filtering for the non-HEARTBEAT messages in _mavlinkMessageReceived(), and here one should filter for the componentID's marked in _handleHeartbeat() as "seen". (i.e. one needs to keep a list of "seen" compIDs).

your attempt in  _mavlinkMessageReceived() is wrong, IMHO

If you like I can do this and PR.

dogmaphobic:
>EDIT: I believe the _cameraInfoRequest QMap provides this list of "seen" already, so it just needs a simple change in _mavlinkMessageReceived(). This appears to work great for me.

Yes, but I didn't have the time to implement and test it. Leaving as it was was causing it to error out when you started having messages (such as battery status) coming from the autopilot. This PR was a simple and safe fix until I had the time to think about it further.

olliw42:
I see. You surely understand that this is a quite different line of reasoning than saying "we must filter them based solely on the component ID"
:)

(and I guess I would have thought that there would be left some trace somewhere, as issue as comment, reminding that this needs to be not forgotten)

but you are right that I haven't done my testing of that previous PR sufficiently well ... I just now note I haven't had my autopilot to emit a battery message, so I didn't notice that issue. That's certainly what I'm to blame for. Sorry for it.

I've PR-ed the suggested change, along with docs of my testing of it.

DonLakeFlyer:
Can this get resolved somehow? Quickly? The constant console output spew from the camera stuff makes debugging anything really hard.

DonLakeFlyer:
This pull https://github.com/mavlink/qgroundcontrol/pull/8026 comments out the qWarning which spits out over and over and over and over and over...

olliw42:
I'd like to stress that the annoying stream of messages is not just noise which one wants to silence but indeed reflects a serious flaw in the code ... so it's good that it is there!

DonLakeFlyer:
> I'd like to stress that the annoying stream of messages is not just noise which one wants to silence but indeed reflects a serious flaw in the code ... so it's good that it is there, and one shouldn't just silence it. :)

Then fix the damn bug like I asked!

olliw42:
ähm, sir, I suggested a fix along with some serious testing in PR #8000. Not sure what else I could do.

:)

DonLakeFlyer:
Of course I understand the warning was important. And of course I noted that I removed it here so that it could be put back when the problem was fixed. And of course I expected it to be put back when the problem was fixed.

> Not sure what else I could do.

Uncomment the warning when the problem is fixed instead of posting a stream of comments about it.

olliw42:
???

I am totally lost. 

Dogmaphobic noted an issue and suggested a solution, which I considered incorrect and thus very shortly after within less of a day provided an alternative which I think is corrected and PRed it. 

It's you guys decision and power to merge it or not, not mine. But not to merge it but blame me for ongoing qWarnings doesn't look appropriate. 

IMHO.

olliw42:
so pl let me understand this: 8 days ago a very annoying bug was found. A solution is available since 7 days. 3 days ago you were pressing for urgent solution. But the available solution is still not merged. 

I obviously miss an important point here, I would appreciate if it could be clarified :)

I obviously write since I too find the bug very annoying ... :) :)

DonLakeFlyer:
Gus is out of town. I asked him to take a look.

olliw42:
this is disturbing. I genuinely believe in arguments, but arguments are totally denied or ignored here. With the result that code is merged which is known to be flawed. Don't tell me that this isn't the opposite of competent programming. Very disturbing.
:-1:

dogmaphobic:
> 8 days ago a very annoying bug was found. A solution is available since 7 days

This indicated to me it was done and ready.

> With the result that code is merged which is known to be flawed.

If it wasn't ready, why wasn't it flagged as such (or fixed)?

PR #7999, opened 2019-11-09
CameraManager: Replace sCompId string by compId int

olliw42:
for some (strange) reason the _cameraInfoRequest QMap uses a QString as index for the int-valued mavlink message componentID. This PR replaces it by a int.


olliw42:
should be a safe change of code, shouldn't it?

PR #8001, opened 2019-11-09
MAVLinkInspector: Button fields too short

olliw42:
The fields in the mavlink inspector were too short for displaying also well when the message frequencies are >= 10 Hz.

This PR widens them by 3 default pixel widths, which - on my screen - is sufficient to also display frequencies above 100 Hz well. 
(I don't really understand the math, since the minimum width for the id and name is 3+28 = 31, and 100.0Hz would be another 7, so that a minimum width of the button of 40 sounds OK, but the math is obviously not as simple, and above my paygrade. Empirically 43 appears to be good).

cheers, Olli

Before:
note the missalignement on the right edge
qgc-mavinspectorwidth-001

After:
(taken during connection where e.g. the autopilots PARAM_VALUE comes whith high frequency)
qgc-mavinspectorwidth-002

### DIFF
--- a/src/AnalyzeView/MAVLinkInspectorPage.qml
+++ b/src/AnalyzeView/MAVLinkInspectorPage.qml
@@ -94,7 +94,7 @@ Item {
                         selectionValid  = true
                         curMessageIndex = index
                     }
-                    Layout.minimumWidth: ScreenTools.defaultFontPixelWidth * 40
+                    Layout.minimumWidth: ScreenTools.defaultFontPixelWidth * 43
                 }
             }
         }


DonLakeFlyer:
That sort of width calculation always seems to blow up and be incorrect without requiring a big amount of wasted padding. Which in turn causes things to work like crap on small screens.

Why not dynamically calc the width of the buttons? The comp id and Hz take some fixed amount of width space. Then the button has an onCompleted which updates a max message name width based on the size of the message name label. Then the width of the flickable is the width of the max calced width. Which then also sets to width of the buttons. Things like the Settings and Setup button flickables do basically the same thing to know how wide they should be.

olliw42:
ah, like it's done in the ParameterEditorController qml I was touching before ... I'll try that. 
That's an interesting lesson, thx.

olliw42:
I've tried doing this, and failed to get anything which I would like or find acceptable. 
I'm not sure it's as easy as suggested since the Hz fields are not populated on component complet, so one essentially ends up with just the minimumWidth of the MAVLinkButton of 3+28=31. So I tried to fool around with adjusting dynamically via onMessageHzChanged but didn't got anything I liked.
So, sorry, I don't have the skills to do what you suggested. The above is the best I can offer. If it's not sufficient so it be.
:)

olliw42:
resolved, so I close
THX

Issue #7957, opened 2019-10-28
Plan Wizard:  make unplanned wizard items more obvious and Warn/prevent Upload


Antiheavy:
The little red question mark isn't a very obvious indicator that the wizard item is unplanned and needs finishing.

![image](https://user-images.githubusercontent.com/12117027/67720589-5524bd00-f9a2-11e9-8f5f-f48a89152297.png)

Maybe use much larger question marks?  or maybe highlight the entire box in red as well?  Or something else that makes it obvious and intuitive.
![image](https://user-images.githubusercontent.com/12117027/67720670-96b56800-f9a2-11e9-923d-f1aa90217407.png)

Also, should Upload be warned against or blocked if there are unfinished wizard items?
![image](https://user-images.githubusercontent.com/12117027/67720834-0fb4bf80-f9a3-11e9-96b8-1003509fe94b.png)

DonLakeFlyer:
> Also, should Upload be warned against or blocked if there are unfinished wizard items?

Upload should error out in this case. If not, it's a bug. The button should be available though. Same for Save.

PR #8002, opened 2019-11-09
Plan: Incomplete visuals, Polygon/line toolbar

DonLakeFlyer:
* Polygon/line toolbar changes
  * Centered
  * Only as big as it needs to be
  * Same background coloring as left toolstrip
* Incomplete data indication changes
  * Entire item is borded in red if incomplete
  * Text to explain why incomplete

Shows new toolbar and incomplete data visuals:
Screen Shot 2019-11-09 at 4 35 30 AM

Incomplete due to terrain data:
Screen Shot 2019-11-09 at 4 34 24 AM

Related to #7957

### DIFF
--- a/src/MissionManager/QGCMapPolygonVisuals.qml
+++ b/src/MissionManager/QGCMapPolygonVisuals.qml
@@ -516,10 +516,11 @@ Item {
         id: toolbarComponent
 
         PlanEditToolbar {
-            x:          mapControl.centerViewport.left
-            y:          mapControl.centerViewport.top
-            width:      mapControl.centerViewport.width
-            z:          QGroundControl.zOrderMapItems + 2
+            anchors.horizontalCenter:       mapControl.left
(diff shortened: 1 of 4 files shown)


Antiheavy:
FWIW, I find the term "polygon" redundant in the "Polygon Tools" title and "Trace Polygon".  Suggest just "Trace".  Also, just "Trace" is used in the Corridor Tools so might as well be consistent.

DonLakeFlyer:
> I find the term "polygon" redundant in the "Polygon Tools" title and "Trace Polygon".

Agree on "Trace", disagree on "Polygon Tools". If all you see is Basic/Circle/Trace/Load KML then it's not entirely clear at first glance what these things are for.

Antiheavy:
I meant to say “polygon” was shown twice, hence redundant (literally). Yes, the menu probably needs a title like “polygon tools” or something along those lines. Maybe “survey tools” to be consistent with “corridor tools”. Or maybe corridor tools should be “line tools” instead.

Issue #8003, opened 2019-11-09
mainWindow.showMessageDialog click Ok, entire app close


DonLakeFlyer:
Repro steps:
* Go to Plan view
* Click Survey in the Create Plan thingy
* File/Save As
* Dialog pops up about incomplete data
* Click Ok
* App closes

dogmaphobic:
I don't have the answer yet but I've located the source. It started with this fix: https://github.com/mavlink/qgroundcontrol/pull/7978

`ChartView` in PIDTuning.qml was crashing QGC. @stefandunca found this:
https://bugreports.qt.io/browse/QTBUG-49895

That fixed the `ChartView` crash but now it seems `MessageDialog` is causing issues if `QApplication` is used instead of `QGuiApplication`.

dogmaphobic:
It must be some weird combination of Qt modules or something. Putting together a bare minimum QML app showing both a `ChartView` and a `MesageDialog` works just fine (only if using `QApplication`):

<img width="612" alt="Screen Shot 2019-11-09 at 11 45 25 AM" src="https://user-images.githubusercontent.com/749243/68532010-a5771580-02e6-11ea-9999-d2e1891abb2e.png">

If you do the same in QGC, the entire applications closes when you dismiss any `MesageDialog`. Just go to General Settings, check _Clear all settings on next start_ and hit "No". It closes all of QGC when using `QApplication` (it works fine when using `QGuiApplication` but then `ChartView` crashes.)

The bare minimum app:

```javascript
import QtQuick          2.12
import QtQuick.Window   2.12
import QtQuick.Controls 2.4
import QtQuick.Dialogs  1.3

import QtQuick          2.0
import QtCharts         2.0

Window {
    visible:            true
    width:              1280
    height:             800
    title:              qsTr("Hello World")

    Rectangle {
        anchors.fill:   parent
        color:          "black"
    }

    MessageDialog {
        id:                 simpleMessageDialog
        title:              "Test Dialog"
        text:               "Test Text"
        standardButtons:    StandardButton.Ok
        modality:           Qt.ApplicationModal
        visible:            false
    }

    ChartView {
        anchors.centerIn:   parent
        width:              400
        height:             300
        theme:              ChartView.ChartThemeBrownSand
        antialiasing:       true
        PieSeries {
            id: pieSeries
            PieSlice { label: "eaten"; value: 94.9 }
            PieSlice { label: "not yet eaten"; value: 5.1 }
        }
    }

[...]

Issue #7992, opened 2019-11-08
Shutdown while parameter writes are not complete


DonLakeFlyer:
Should warn user.

PR #8005, opened 2019-11-10
Param progress bar and other stuff

DonLakeFlyer:
* Initial parameter download progress works like it always did
* You will now get a progress bar for parameter writes. It shows up as soon as a parameter is updated and will go away when it gets a response back from the vehicle. If multiple writes are batched up like with a file load, then you get a progress bar whose length matches the batch count. And it will go away once all writes are acked.
* You will now get a progress bar for parameter reads just like writes. This shows up in some of the calibration ui where sets of params are redownloaded after a calibtation.
* If there are pending parameter writes and you try to close QGC you will get a warning with a yes/no close option.
* Fixed the parameter load errors display when loading from a file.

### DIFF
--- a/src/FactSystem/ParameterManager.cc
+++ b/src/FactSystem/ParameterManager.cc
@@ -94,6 +94,63 @@ ParameterManager::~ParameterManager()
     delete _parameterMetaData;
 }
 
+void ParameterManager::_updateProgressBar(void)
+{
+    int waitingReadParamIndexCount = 0;
+    int waitingReadParamNameCount = 0;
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
Replacement for #7968. Fix for #7992.

DonLakeFlyer:
@dakejahl FYI

DonLakeFlyer:
@dogmaphobic @stefandunca I backed out #7978 in this pull since the cure was worse (#8003) than bug. I'll take a look at tuning.

PR #8008, opened 2019-11-10
MGRS coordinates

MatejFranceskin:
This PR adds possibility to display vehicle and waypoints positions in MGRS coordinates.
Code for converting between Geodetic, UTM and MGRS coordinates was taken from GeodeticLib [https://geographiclib.sourceforge.io/html/](url) which is licensed under MIT/X11 license.
Previous UTM conversions were replaced with conversions from GeodeticLib library.

MGRS description: [https://en.wikipedia.org/wiki/Military_Grid_Reference_System](url)

Vehicle MGRS position:
MGRSPosition

Editing waypoints:
MGRSEditPosition


hamishwillee:
FWIW Longer term should the position options be configurable by app settings? Ie most users will only use one of these, so configuring for that group might make the behaviour even more obvious

PR #8009, opened 2019-11-11
Plan mission item editor: Delete icon and Command Picker ui

DonLakeFlyer:
* Add Delete icon
* Rework command picker visuals to be more understandable as something to click on. Only visuals are changed. Clicking still goes to command picker dialog.
* No changes to hamburger

Screen Shot 2019-11-10 at 6 29 54 PM

Related to #7982

### DIFF
--- a/src/PlanView/MissionItemEditor.qml
+++ b/src/PlanView/MissionItemEditor.qml
@@ -3,6 +3,7 @@ import QtQuick.Controls             2.4
 import QtQuick.Controls.Styles      1.4
 import QtQuick.Dialogs              1.2
 import QtQml                        2.2
+import QtQuick.Layouts              1.11
 
 import QGroundControl               1.0
 import QGroundControl.ScreenTools   1.0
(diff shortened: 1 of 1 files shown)


tilaktilak:
Great! 
Can the trash be at the left of the hamburger? so we put the clickable icon altogether.
Also, I changed the trash icon a little bit to make it looks like a rubbish bin as it's going to be shown on every MissionItem now: 8f4f2780305086c293d99de8e5bc050ab6446028
Did you integrate the "insert current" thing to it?

DonLakeFlyer:
> Can the trash be at the left of the hamburger? so we put the clickable icon altogether.

Not without taking away space for the command name which is already tight.

If you look at this screen shot you can see how there are three columns in which things all line up in horizontally. These column line up whether the item is selected or not. Left: Not Ready indicator/Trash icon, Center: Command, Right: Hamburger. Long command names will expand right up to the hamburger. Also note how the command names all line up whether item is selected or not. So there is already available space on the left due to the possible display of the Not Ready indicator. Placing the trash to the right would force taking space away from the command name, and/or not lining things up horizontally. Since everything in this first row of information is clickable. I don't see how grouping trash next to hamburger makes a big difference in usability.

Screen Shot 2019-11-10 at 7 14 54 PM

> Did you integrate the "insert current" thing to it?

No.

> Also, I changed the trash icon a little bit to make it looks like a rubbish bin

I just used the one already in there. It's fine if you want to change to a better looking one. Can you take the trash and insert stuff from your pull into a new pull?

DonLakeFlyer:
The above image also shows that I need to work on lining up the trash icons with the Not Ready indicators horizontally.

tilaktilak:
I prepare a PR with the "Insert Current" and the Trash Icon.
I think we should remove "Delete" and "Change Command" from the hamburger menu as it's now clearly exposed by the MissionItem list.

Antiheavy:
IMHO the trash icon on the left just looks wrong. I wonder if the hamburger would work better on the left?

DonLakeFlyer:
> I wonder if the hamburger would work better on the left?

Given the fact that the hamburger gets minimal usage I would contend that putting on the left is even "wronger"

PR #8010, opened 2019-11-11
Mavlink Inspector: Dynamic sizing of message buttons

DonLakeFlyer:
* Dynamically size message buttons based on max message name width
* Comp id and Hz columns are fixed width, message name is variable
* Also reduced left/right insets on Button to use less space.

Replacement for #8001

Screen Shot 2019-11-10 at 7 49 59 PM

### DIFF
--- a/src/AnalyzeView/MAVLinkInspectorPage.qml
+++ b/src/AnalyzeView/MAVLinkInspectorPage.qml
@@ -22,11 +22,12 @@ Item {
     anchors.fill:           parent
     anchors.margins:        ScreenTools.defaultFontPixelWidth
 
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@olliw42 Can you test this after it goes through? I think you have a better multi-component setup than I do to test with.

olliw42:
looks great
here a screen shot in the initial phase where you see IDs in various ranges and also high rates ... i.e. both "ends" of the dynamics are captured here
qgc-mavinsp-001

DonLakeFlyer:
@olliw42 Thanks for testing

Issue #8004, opened 2019-11-10
Plan Wizard: Custom plan should not have wizard items, let user plan from scratch


Antiheavy:
This could be just my opinion, but it seems like if someone selects "custom" they should be allowed to plan an entire mission from scratch, just like they always have been able to with QGC.  i.e. "custom" should launch them into the full blown traditional plan screen with no wizard helpers on the right side.

![image](https://user-images.githubusercontent.com/12117027/68537100-55b73f00-0324-11ea-96e7-cc29781c34ac.png)

Who knows if a takeoff an a couple normal waypoints are even what are intended in their custom mission.... could be something totally different, actually probably is something different in most cases.

DonLakeFlyer:
Yeah, I agree. I was just trying some things out there.

PR #8013, opened 2019-11-11
Plan: Create Plan

DonLakeFlyer:
* Create Plan tools are no longer a separate overlay. They are part of the standard File drop panel
* Create Plan - Blank just clears any existing Plan from the view and hides the File panel. This differs from previous New button in that it **does not** clear the plan from the vehicle.
* File drop panel will show on first display if there are no items in the plan
* Reworked layout of previously existing File drop panel items
* SectionHeader control no longer automatically sets width. This allows you to set width using either anchors or Layout.fillWidth.

Screen Shot 2019-11-11 at 6 04 09 AM


DonLakeFlyer:
Fixes #8004

Antiheavy:
Ah, neat.  I'm looking forward to playing around with this.  Some initial reactions:

1) Would it be possible to have this menu automatically open if there is no existing mission, similar to how the previous Create Plan was displayed automatically.  Otherwise I fear the "How do I get started?" problem will still be there because the getting started actions will be buried under the File menu.

2) One thing I (and people I previewed to) really liked about the previous version of Create Plan was that it was big and in your face.  It was the most prominent thing on the screen therefore was it obvious what the user should do to move forward.  I worry that putting these visuals seemingly underneath a menu will make it less clear you should select one of those options to move forward.

DonLakeFlyer:
> 1. Would it be possible to have this menu automatically open if there is no existing mission, similar to how the previous Create Plan was displayed automatically.

It does if there is no mission already on the vehicle, or if offline.

PR #8018, opened 2019-11-11
Simplify mission item list 2

tilaktilak:
Related to https://github.com/mavlink/qgroundcontrol/pull/8009 and https://github.com/mavlink/qgroundcontrol/pull/7982

- Insert on the current element of the list and not at the end ( "insert Waypoint" / "insert Pattern" from the Hamburger are issued by the left-hand toolbar )
- Change Trash Icon
- Change Hamburger Menu ( "Delete", "Change Command" and "Insert*" handled by MissionItem list )


DonLakeFlyer:
I get that the some of the hamburger menu items are redundant now. But I'm still not quite sure if the right way to go is to pull them out. Some users are used to going through that path to get things done. I'd like to think some more on what is left with trash and edit poisition and the hamburger. For now can you comment these out with a note instead? I don't want to lose the code in case they come back. After it sits for a while it's likely I'll just pull it out. But for now...

DonLakeFlyer:
Thanks

PR #8019, opened 2019-11-12
Avoid QGC to shutdown connecting to a flying Ardupilot

tilaktilak:
I have noticed that master crash when I connect to a flying Ardupilot simulator. 
After investigation, it's an 'index out of range' on TrajectoryPoints because we try to access `_points[points.count()-1]` and `point.count()=0`. 
Here a proposal for a fix, this needs to be checked though! :-)


DonLakeFlyer:
> After investigation, it's an 'index out of range' on TrajectoryPoints because we try to access `_points[points.count()-1]` and `point.count()=0`.

That is supposed to be guarded against by _lastPoint.isValid. But since clear doesn't reset that correctly you can end up in a state where _lastPoint is valid and the points array is empty.

PR #8022, opened 2019-11-12
Fix EasterEgg overlayed by Flickable item

tilaktilak:
Easter Egg was not working anymore since the last changes.


DonLakeFlyer:
The problem is that the whole clicking in blank space thing doesn't really work well. There might not be any blank space left. The easter egg really needs to move to be inside one of the toolbar buttons. COuld you move it to be associated with the Fly button instead?

tilaktilak:
Yep I'll have a look

PR #8025, opened 2019-11-13
APMHeliComponent - update for Copter 4.0

ChristopherOlson:
@DonLakeFlyer this updates the Heli Setup for Copter 4.0

### DIFF
--- a/src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc
+++ b/src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc
@@ -111,7 +111,7 @@ const QVariantList& APMAutoPilotPlugin::vehicleComponents(void)
                 _components.append(QVariant::fromValue((VehicleComponent*)_followComponent));
             }
 
-            if (_vehicle->vehicleType() == MAV_TYPE_HELICOPTER) {
+            if (_vehicle->vehicleType() == MAV_TYPE_HELICOPTER && (_vehicle->versionCompare(4, 0, 0) >= 0)) {
                 _heliComponent = new APMHeliComponent(_vehicle, this);
                 _heliComponent->setupTriggerSignals();
                 _components.append(QVariant::fromValue((VehicleComponent*)_heliComponent));
--- a/src/AutoPilotPlugins/APM/APMHeliComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMHeliComponent.qml
@@ -37,35 +37,43 @@ SetupPage {
             property real _margins:     ScreenTools.defaultFontPixelHeight
             property bool _showIcon:    !ScreenTools.isTinyScreen
 
-            property Fact _hSvMan:      controller.getParameterFact(-1, "H_SV_MAN")
-            property Fact _hSv1Pos:     controller.getParameterFact(-1, "H_SV1_POS")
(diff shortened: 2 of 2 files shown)


DonLakeFlyer:
Can you add screenshot(s)

ChristopherOlson:
Sure, Don.
Full screen layout
imenjhhjoilejfhe

cejgfgmijkfpcfep

Medium screen layout
iknkjhkkeadibfpb

lepfmpodohiaccem

Small screen like smart phone goes to single-column.

Thanks!

ChristopherOlson:
@DonLakeFlyer I also suggested to Bill to shorten some of the names of the @DisplayName in the metadata. Some of those are pretty long. I can get nicer grouping on the sizes of the boxes if some of those were shorter. But this gets everything working for the 4.0 beta testers using the Daily Builds.

DonLakeFlyer:
So this is for Copter 4.0. Are any of the parameters new and in Copter 4.0 only? If so this is going to pop missing parameter warnings if run against version < 4.0.

ChristopherOlson:
Don, yes 4.0 is TOTALLY different than 3.6. Even the names of the swashplate parameters were changed, some like LAND_COL_MIN were totally eliminated. Unfortunately, there is no possibility of backwards compatibility with 3.6 with this. The Parameters section will handle it, but the Heli Setup, where it calls for specific parameters is not compatible. Maybe I could make a note on the top of the page that this is only for helicopters running Copter 4.0 and newer? For older versions of Copter use Stable 3.5.1?

ChristopherOlson:
@DonLakeFlyer the reason the swashplate params were changed is because there is a complete new swashplate library for heli in 4.0. Even the old servo position params are no longer used.

Realistically, many, if not most, users on Copter with heli are flying my ArduHeli builds instead of stock Copter, since the ArduHeli builds have all the latest stuff that's coming out in 4.0. ArduHeli 3.6.11 IS 100% compatible with this new setup page. Copter 3.6.11 is not compatible.

DonLakeFlyer:
> Realistically, many, if not most, users on Copter with heli are flying my ArduHeli builds instead of stock Copter, since the ArduHeli builds have all the latest stuff that's coming out in 4.0.

Then this should be for ArduHeli 4.0 only then. You need to tweak the APMAutoPilotPlugin.cc such that it only gets added for 4.0 firmware.

ChristopherOlson:
OK, not sure how to do that. But I'll take a look at it this evening and see if I can figure it out.

DonLakeFlyer:
There are other examples in there that do a version compare before adding setup pages.

ChristopherOlson:
Thanks Don! Will take a look this evening when I have more time to examine that.

ChristopherOlson:
@DonLakeFlyer I included a commit that (from what I tested) blocks the Heli Setup from appearing if the version is not 4.0 or greater. I hope I did this right, but it seems to work fine here in my testing.

DonLakeFlyer:
Perfect thanks for working to keep this up to date.

ChristopherOlson:
@DonLakeFlyer I did some more testing with three different controllers and versions. The Version lockout for prior to 4.0 definitely works as expected.

Thanks for making this heli setup for us in the first place! I like the design - very easy to maintain.

PR #8044, opened 2019-11-15
Parameters: Add category "All"

olliw42:
This PR adds a category "All" to the parameters page.

Reason/Motivation:

I read the thread "Most wanted QGC features"  https://discuss.px4.io/t/most-wanted-qgc-features/7031 and found these comments

user Vavooon: https://discuss.px4.io/t/most-wanted-qgc-features/7031/11
_"Separating parameters by categories is also a questionable feature: Every time I need to change a param I can switch between categories in order to find it. Search feature is cool but I do not always remember param name."_

user Antiheavy: https://discuss.px4.io/t/most-wanted-qgc-features/7031/15
_"With regards to hiding the parameters, QGC already supports custom plugins that allow an “advanced mode” and a “standard” mode. The standard mode has whatever custom items turned off/hidden you program it to be. Advanced mode is basically the full QGC. Maybe there could be a community consensus about what would be hidden in a upstream standard build and a mechanism for how to enable/disable would need to be determined."_

DonLakeFlyer: https://discuss.px4.io/t/most-wanted-qgc-features/7031/26
_"That said it would be easy to add an All Parameters view."_

which immediately triggered me because this in fact was and is the No.1 beef I have with QGC from day one, namely that it's ultra inconvenient to adjust parameters because it's essentially impossible to get an overview and find which are there unless you know all of them by heart (and I don't know ca 800 params by heart...)(and I don't want to know ca 800 params by heart...). For instance, for ArduPilot the AHRS parameters are spread over ALL THREE categories, Standard, Advanced and Other !!! I have no idea how one should be able to get along with this. As pointed out, the search is not a substitute since it requires also good knowing.

I thus wished from day one that there would be a category showing ALL parameters in a MAVLink sorted manner, like MissionPlanner's "FullParameter Tree". And I just learned now it's not only me.

So I'm excited about this PR LOL :) :)

Screenshots:
qgc-catall-001
qgc-catall-002


DonLakeFlyer:
I'm not convinced the categorization really adds any value at all. Wonder about just removing them completely?

olliw42:
I don't think you ask the right person to answer that question.

If you ask me however for my opinion, here we go, I do have one:

I don't think the categorization is a bad or useless thing per see. I am the kind of person which wouldn't use it, and I'm presumably not the only one, but I don't find it hard to imagine that there are plenty of those which would in fact prefer a categorization. That is, it's probably really a matter of preference, and thus highly subject per see. QGC would probably benefit from satisfying them all.

The issue with the categories is IMHO not the concept per see, but the actual categorization of the parameters into categories, and the practical difficulties with doing this. 

Having seen some of your comments in that context I got the feeling that there is a certain level of frustration with that (don't get me wrong, you contain that well in your comments but every once in a while it shimmers through),i.e. that you have to do that for e.g. for ArduPilot. I find that very understandable given the amount of time and effort which has to be spend on this, and this annew with every new version. This practical difficulty I think however should not be an argument to abandon the categorization concept.

For the majority of users of QGC the higher-level dedicated setup interfaces, such as the gimbal page and alike, are probably most desired. Which implies that the actual shape of the categories is of comparatively less relevance. This however too should IMHO not be an argument to abandon the categorization concept.

So, the concept adopted by QGC of "Standard" and "Advanced" you probably want to continue.

What I however would do is to get rid of the "Other" category. I mean, either a parameter is standard or advanced, but what argument should make it "other"? Never ever touch it? But when why displaying it at all? So, the advanced user knows? Or is there something like a super-advanced user?

I would assume that the "Other" category is a result from the practical difficulties mentioned before. What I however would rather do is to sort any parameter which is not standard into advanced, and sort any parameter which cannot be put into group be put into a "Misc" group.

One also could argue that advanced should mean to expose all parameters. That is, that the Advanced category actually should be what "All" is. A user may actually rather expect this to be the case. 

So, you see, I would come up with the categories "Standard", "Advanced", "All" or only "Standard", "All", depending on the mentioned choice.

Finally a comment as regards the practical difficulty of categorization: It is not the only, but it is one of the main reasons for why I was suggesting the COMPONENT_INFORMATION and the associated xml definiton file concept, and am trying to push it forward, not because it would be of any personal benefit to me but because I strongly believe that it would be of much benefit to the community. I would wish the community would more clearly see the benefits (IMHO the community as an entity doesn't appear to be always good at seeing benefits ;)). I in fact laid out my thoughts in that much detail in the above in order to raise the insight that that mechanism would be a canonical solution (also) to the categorization.

:)

Last but not least: If you want me to implement the change to Standard-Advanced-All or Standard-All, I would happily do it.

DonLakeFlyer:
FYI: Categories are not really a QGC thing. They are defined by the parameter metadata.

DonLakeFlyer:
My take on this is that if you are in the Parameter Editor you have already fallen off the power user cliff. You need to set something which is not covered by vehicle setup ui. You are not going there to just peruse the list of parameters (for which categories might be helpful). 

You are going there either to:
* Set a specific parameter you already know about
* Maybe search for a parameter you forgot the name of

In either case categorization does't help you at all with the main usage of the parameter editor. Or in the specific parameter case it gets in the way.

I think the change here should be to:
* Get rid of categorization
* Use a single section header for the autopilot component which has all parameters in it.
* Use a more friendly naming for it than the generic component section naming scheme. Likely leave off the ID and just put the name "Autopilot".

DonLakeFlyer:
> I would assume that the "Other" category is a result from the practical difficulties mentioned before.

Just FYI: The Other category are parameters for which there is not metadata specified so QGC doesn't know where to put them. In reality for ArduPilot it's possible to fix this up to be better such that they group with their other ```FOO_``` grouped friends. And I've alerady fixed that with a pull coming. But for PX4 that is not possible since PX4 uses human readable grouping for params:
Screen Shot 2019-11-16 at 8 50 59 AM
The human readable grouping I think is actually useful and should become part of the COMPONENT_INFORMATION metadata spec.

olliw42:
> FYI: Categories are not really a QGC thing. They are defined by the parameter metadata.

I think I understood that. :) For the argument it doesn't matter however who is the who. It needs to be defined.

> My take on this is that if you are in the Parameter Editor you have already fallen off the power user cliff.

I think I understood that this is what you were implying with your first comment. However, I still think one probably doesn't want to go that drastic. I mean, the Standard category really doesn't hurt, and many users may have gotten used to it, and just because one doesn't hear complains it doesn't mean that no users like it. So, if it's my take, I still would go with Standard-Advanced-All or Standard-All.

> Just FYI: The Other category are parameters for which there is not metadata specified so QGC doesn't know where to put them. In reality for ArduPilot it's possible to fix this up to be better such that they group with their other FOO_ grouped friends. 

I think we are saying the same with different words.

olliw42:
> The human readable grouping I think is actually useful and should become part of the COMPONENT_INFORMATION metadata spec.

I put my answer in a extra post, since I think that's now important: 

Although the xml file is obviously linked to the COMPONENT_INFORMATION message spec, the definition file entries themselves are actually kind of their own spec in the sense that they are largely unrelated to the COMPONENT_INFORMATION message, and microservice more generally. That's analogous to  how it is for the camera micro service. For instance, it will be quite simple to add definitions to the xml file, any time later, without affecting the COMPONENT_INFORMATION spec.

There had been ZERO discussion on the xml file entries. I mocked up my own version here ... obviously because nothing has happened for 1/2 year in that regards ... I think this kind of gave me the right to do so ;) 

I'd like to put that also in the context of your comment in the other thread, _"Can you hold off pushing on that. I want to have time to review it again and possibly propose additional changes before it goes in."_ I suspect that you were rather meaning the xml file definition and not actually the COMPONENT_INFORMATION message itself. If that is the case, you might consider releasing your hold on merging the COMPONENT_INFORMATION message, Both hamishwille and julianoes considered it good to go. :)

src/FactSystem/ParameterManager.cc
```diff
@@ -661,6 +661,8 @@ void ParameterManager::_setupDefaultComponentCategoryMap(void)
     for (const QString &paramName: _mapParameterName2Variant[_vehicle->defaultComponentId()].keys()) {
         Fact* fact = _mapParameterName2Variant[_vehicle->defaultComponentId()][paramName].value<Fact*>();
         defaultComponentCategoryMap[fact->category()][fact->group()] += paramName;
+        int i = paramName.indexOf("_");
+        defaultComponentCategoryMap[tr("All")][(i > 0) ? paramName.left(i) : tr("Misc")] += paramName;
```
DonLakeFlyer: Are you explicitly not using group names here on purpose? I'm sort of 50/50 on whether that is good or bad. For PX4 the group names are way more human readable than just a name prefix.

olliw42: yes

although I can't really speak for PX4

I totally agree that the mavlink-ish naming should go and be replaced by human readable text. This should IMHO however so for both the groups and the parameter names. The Camera manager is a good example of there it should end up. 

However, that's not at all the current status (the COMPONENT_INFORMATION mechanism aims at getting there). The current state of affairs is in fact quite inconsistent, in that the parameters are still named by their mavlink names. This is so in QGC, also for PX4 (it is the first entry in the list and not the 3rd entry !!!). But this is so also in the docs !!! I.e., when you search for a parameter in the docs you have to search for the mavlink names, since these are given there (true for both PX4 and ArduPilot). Therefore, currently I think mavlink-ish sectioning is much more helpful. 

Once things are in a state there the mavlink names don't appear to the user, because only human readable names are used, both in QGC and the docs, I would agree with a human readable grouping also for a "All" category. However, although I certainly wish this state would be reached soon, experience shows that the timelines are long. Hence that choice.

olliw42: so, I myself am happy with this, I'm actually 50/50 with whether one should also for a "All" category take out eventual last digits from the group name, like you do in this other PR of yours. Split minded here. :)

DonLakeFlyer: > whether one should also for a "All" category take out eventual last digits from the group name

Not quite sure what you mean there. 

But in thinking about this a bit I think All should use group names. Which should end up showing the same visual on ArduPilot but show humn readable group names for PX4.

olliw42: group.remove(QRegExp("[0-9]*$")); 

GUI design is subjective :)

DonLakeFlyer: Ok, I get it now. I'd just leave it as is. My concern is that it will work and be good in most cases but may not be correct in all. The cases where it works it where the number indicates a series of parameters. But that may not always be the case for why a number is used. Imagine something like FOOV2_BAR.   Removing in that case is incorrect.

DonLakeFlyer: I think it's just this last change and this can go in?

olliw42: thx for asking, much appreciated. This actually fell a bit off my mind, also since the simplification of removing "Other" you already did. I will rework it the next days.

olliw42: ok, looked at it. I think the last change you mentioned is missing is to show human readable group names for PX4. I unfortunately couldn't figure that out beyond doubt, and currently don't have a setup to test, so I'm sorry to say that I can't accomplish this task.

I hope to find time over Christmas break to get me into PX4, but for the moment that's the situation. Not sure if it's better to keep this PR open for a later starting point or to close it. Pl decide.

DonLakeFlyer: > I unfortunately couldn't figure that out beyond doubt, and currently don't have a setup to test

You can use MockLinks to test things without requiring a real connection. App Settings/MockLink

olliw42: ah, very interesting comment. I'll see what I can make of it :) THX

olliw42: first, the MockLink is kinda cool, isn't it :D

I've played with this now, but there isn't no satisfying solution possible IMHO

I've changed my addition to `defaultComponentCategoryMap[tr("All")][fact->group()] += paramName;` which I think is what is supposed to be done.

the problem is the group 'Misc'. For PX4 there is a category 'Other' with group 'Misc' (to my surprise, I thought we killed the 'Other', so it must be a 'hardwired' thing), and for ArduPilot all unassigned parameters are put into a group 'Misc', which is found in the category 'Advanced'.

Therefore, with the above algorithm, there is also a group 'Misc' in the 'All' category, which is however not what one wants to have, since e.g. a parameter which starts with 'COM_' should be put into the group 'Commander' where all the other 'COM_' parameters are located. That is, the main purpose of having an 'All' category in the first place is defeated.

Now one could think that one could take that parameter from 'Misc' and try to find the proper group. In the example of the 'COM_' one could search for the group there the other 'COM_' params are located. However, this obviously only works unambiguously if either (i) any group has just parameters with the same header and/or (ii) if parameters with the same header are not distributed in several groups. However none of these conditions are fullfilled for PX4. There are several examples, the groups 'SensorCalibration' and 'Sensors' are most clear: In both one finds params with header 'CAL_' and 'SENS_', so it's impossible to sort out where a 'SENS_' or 'CAL_' parameter should go to. Ergo, no solution possible. Q.E.D.

A solution would only be possible if unassigned parameters could not ever occur, which obviously isn't fulfilled.

I have no idea how to proceed from here. :)

Issue #7956, opened 2019-10-28
Plan Wizard: suggested text for Takeoff dialog


Antiheavy:
After showing the [WIP] Takeoff dialog to a number of people here is some suggested improvements to the text:

Suggestions:
"Move the P near the planned launch location and move the T to the desired climbout location."
"Move P (Planned Home) near the launch location and move T (Takeoff) to the climbout location."
also
"Ensure clear of obstacles and into the wind."  (only add the "into the wind" part if you are a fixed wing)

Current:

![image](https://user-images.githubusercontent.com/12117027/67720197-4f7aa780-f9a1-11e9-99ca-b6edc7bc05dc.png)

tilaktilak:
We can probably have a "click-to-add' instead of making it appears on the center of the map and explain to the user that he has to move it ? That will make user experience more sensible.    

Also I think the Takeoff / Planned Home behavior is different depending on MultiCopter, Plane, ArduPilot, PX4, ... I put that here.

DonLakeFlyer:
> We can probably have a "click-to-add'

Done.

> here is some suggested improvements to the text:

I lost track of this one when I was making changes in #8049. I'll fix this up.

DonLakeFlyer:
Wording changes are done

PR #8049, opened 2019-11-17
PLan: Takeoff item user model changes

DonLakeFlyer:
Given how different firmwares and vehicles deal with takeoff items which do or don't include a location. Trying to explain how this all works now is next to impossible since it's so complicated. The simplest way to explain how it works now is:
* The user is asked to click in the map to set initial location when appropriate. Sometimes that sets launch location, sometimes takeoff location, sometime both!
* The editor drops in/out of wizard mode appropriately
* A fixed wing takeoff which requires a coordinate sets the takeoff item coordinate to be at a 30% climb out from the launch location

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,7 +6,7 @@ Note: This file only contains high level features or important fixes.
 
 ### 3.6.0 - Daily Build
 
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
Also given how complicated this all is it would be a miracle if I didn't screw something up

DonLakeFlyer:
Related to #7956

PR #8053, opened 2019-11-18
cmake: add ADSB

dagar:


dagar:
The build failure is Linux Release with the existing qmake build system.

image

The build slaves probably need a slight bump. Alternatively perhaps the resource data could be split up a bit.

PR #8055, opened 2019-11-18
Set Rally Point default value to 0.0 and get it from Fact's default

stefandunca:
The Rally Point default value is taken from user settings `Default Mission Altitude`.
This PR does the following:
 - Take the default value from the `Fact`'s default value if NOT fixed wing. Otherwise use the same behaviour `Default Mission Altitude`
 - Rename the misleading `Altitude` to `Relative Altitude` (relative to home altitude)
 - Set the default value to `0.0` (use home altitude by default)

### DIFF
--- a/src/MissionManager/RallyPoint.FactMetaData.json
+++ b/src/MissionManager/RallyPoint.FactMetaData.json
@@ -12,10 +12,11 @@
     "decimalPlaces":    7
 },
 {
-    "name":             "Altitude",
+    "name":             "Relative Altitude",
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
I think this makes sense for everythig but fixed wing where a rally point is just a loiter point.

stefandunca:
> I think this makes sense for everythig but fixed wing where a rally point is just a loiter point.

Good catch. I've updated the PR to keep the old behaviour if fixed wing. Will this work?

DonLakeFlyer:
Great all good now

PR #8058, opened 2019-11-19
Firmware update: add Durandal-v1

bkueng:


DonLakeFlyer:
Does this also require an update to USBBoardInfo.json for autoconnect? Or does this new board match the existing Holybro entry:
```
        { "vendorID": 12642, "productID": 0,        "boardClass": "Pixhawk",    "name": "Holybro" },
```

bkueng:
Yes, it already matches, and auto-connects.

PR #8061, opened 2019-11-20
Removed compile warnings on Linux

MatejFranceskin:
Building QGC on Ubuntu 19.10 with gcc (Ubuntu 9.2.1-9ubuntu2) 9.2.1 20191008 produces lots of warnings. Some of these warnings were already disabled in builds for other platforms so I added them also to Linux build.


DonLakeFlyer:
@stefandunca There a number of these which come from mavlink headers. Those should be fixed in mavlink to get mavlink headers to compile clean. Can you create an issue against mavlink so this can be fixed? I'm fine with turning those off until mavlink gets fixed.

```no-deprecated-copy``` What is causing this one?

DonLakeFlyer:
FYI: I'll see if pulling a new eigen will get rid of all the crap from there.

MatejFranceskin:
There are a lot of deprecated-copy warnings coming from QT headers:

/home/matej/Qt/5.12.6/gcc_64/include/QtCore/qvariant.h: In constructor ‘QVariant::QVariant(QVariant&&)’:
/home/matej/Qt/5.12.6/gcc_64/include/QtCore/qvariant.h:273:25: warning: implicitly-declared ‘QVariant::Private& QVariant::Private::operator=(const QVariant::Private&)’ is deprecated [-Wdeprecated-copy]
  273 |     { other.d = Private(); }
      |                         ^
/home/matej/Qt/5.12.6/gcc_64/include/QtCore/qvariant.h:399:16: note: because ‘QVariant::Private’ has user-provided ‘QVariant::Private::Private(const QVariant::Private&)’
  399 |         inline Private(const Private &other) Q_DECL_NOTHROW
      |                ^~~~~~~

PR #8065, opened 2019-11-22
Microservice versioning Proof of Concept

ItsTimmy:
Related:

 - [Microservice versioning document](https://docs.google.com/document/d/1zK_JcJbhtbRA3W0MgnXSNWMUxOcBu6vXocrjZVMXki4/edit#)
 - https://github.com/PX4/Firmware/pull/13565
 - https://github.com/PX4/Devguide/pull/945

This PR uses [this common.xml](https://github.com/ItsTimmy/mavlink/blob/pr-microservice-versioning/message_definitions/v1.0/common.xml), and [this C library](https://github.com/ItsTimmy/c_library_v2) generated from it.

This is still a work in progress, because the exact API for microservice versioning has not been finalized. In addition, there is still work to be done on this code:

- [x] Clean up old code that is currently commented out
- [x] Determine how the mission microservice should act with systems that do not perform the microservice version handshake
- [x] Update comments in the Mission microservice to reflect the changes
- [ ] Update MAVLink module to include new auto-generated constants for service IDs and new messages
- [ ] Thorough testing

This is currently a minimum functioning prototype. Work is now being done in PX4 and MAVSDK to reach the same state, while the specifications are finalized and MAVLink is updated.

More context: https://docs.google.com/document/d/1zK_JcJbhtbRA3W0MgnXSNWMUxOcBu6vXocrjZVMXki4/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,7 @@ Qt*-linux*.tar.*
 .vs/
 
 libs/airmapd/include/boost
+
+# Jetbrains IDEs
+.idea/
+cmake-build-debug/
--- a/src/MissionManager/PlanManager.cc
+++ b/src/MissionManager/PlanManager.cc
@@ -198,12 +198,19 @@ void PlanManager::_ackTimeout(void)
         break;
     case AckServiceVersion:
         if(_retryCount > _maxRetryCount){
-            _sendError(VehicleError, tr("Request mission service version failed, maximum retries exceeded."));
-            _finishTransaction(false);
+            //_sendError(VehicleError, tr("Request mission service version failed, maximum retries exceeded."));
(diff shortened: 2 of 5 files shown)


hamishwillee:
FWIW

> Determine how the mission microservice should act with systems that do not perform the microservice version handshake

Exactly as it does before this change - ad hoc fallback.

>  Update MAVLink module to include new auto-generated constants for service IDs and new messages

Hmmm. The new messages need to be updated, and they can go in as WIP anytime. We hadn't actually discussed/considered generation of the services and version numbers via the toolchain. It makes sense though.

ItsTimmy:
Yes, I think it is important to generate constants like `MAVLINK_SERVICE_ID_MISSION` and so on. I don't know how the service numbers should be generated as constants. Perhaps `MAVLINK_SERVICE_MISSION_MAX_VERSION` or something? But that is so far less important to me.

PR #8071, opened 2019-11-24
Qmlsink simple verion

tcanabrava:
Cleaned up implementation of the Qml Videosink.

This implementation handles only UDP connections, as I tried to make the code as simple as possible.
it does not handles record / subtitles: dom asked me to only do video streaming. I have removed tons of legacy code but I havent removed all of it, there's a lot of code that I haven't touched as I don't really understand what it's doing.

@dogmaphobic said that he does not like the idea of using gst_parse_string as it would need all of the gst instalation: this is not true, we can pack only what we use. 

This implementation removes around 7K line of code from the original one, and plays video.


tcanabrava:
This is a huge update in this patch series trying to address most of the issues:
- Patches in small increments, every single commit is compillable.
- Easy to follow the developer's state of mind.
- Handles UDP H264 and H265 streams (and I believe the pipeline creation via gst_parse_launch is the way to go for this), also handles 'no' video stream with a tv pipelne in case of no video configured (I believe that it's better UX not to remove elements completely from the screen - I personally recompiled QGC a few times trying to get the video output window, only to discover that I needed to enable video in the settings)

- Enables Play / Pause directly in the PIP controller ( to showcase how the controllers should be written).

There's also another patch for qmlglsink inclusion that keeps the software as is, probably it should be added before my patch, think of this as a try to do an archectural change, and not just to replace the sink.

src/Vehicle/Vehicle.cc
```diff
@@ -1676,11 +1676,9 @@ void Vehicle::_updateArmed(bool armed)
             _trajectoryPoints->stop();
             _flightTimerStop();
             // Also handle Video Streaming
-            if(qgcApp()->toolbox()->videoManager()->videoReceiver()) {
-                if(_settingsManager->videoSettings()->disableWhenDisarmed()->rawValue().toBool()) {
-                    _settingsManager->videoSettings()->streamEnabled()->setRawValue(false);
-                    qgcApp()->toolbox()->videoManager()->videoReceiver()->stop();
-                }
+            auto *videoSettings = _settingsManager->videoSettings();
+            if(videoSettings->disableWhenDisarmed()->rawValue().toBool()) {
```
patrickelectric: I don't know if videoSettings can be null but:
`if(videoSettings && videoSettings->disableWhenDisarmed()->rawValue().toBool())`

tcanabrava:
Video settings cannot be null, it’s created before the object is
constructed and set beforehand.

It’s hard to see the object creation hierarchy’s because of the QCGToolbox
madness.

src/Camera/QGCCameraControl.h
```diff
@@ -322,6 +322,7 @@ class QGCCameraControl : public FactGroup
     void    thermalModeChanged              ();
     void    thermalOpacityChanged           ();
     void    storageStatusChanged            ();
```
jaxxzer: try to match the existing whitespace style

src/VideoStreaming/VideoStreaming.cc
```diff
@@ -143,6 +143,11 @@ void initializeVideoStreaming(int &argc, char* argv[], char* logpath, char* debu
             qCritical() << "gst_init_check() failed: " << error->message;
             g_error_free(error);
         }
+        // GStreamer needs the sink to be created before any Qml elements
+        // so that the Qml elements are registered in the system. so we create
+        // and free it.
+        GstElement *sink = gst_element_factory_make ("qmlglsink", nullptr);
```
jaxxzer: This is funny (to me), do you have any reference url?

src/VideoStreaming/VideoSurface.cc
```diff
@@ -0,0 +1,96 @@
...
+    Q_UNUSED(data)
+    if (!_shouldStartVideo) {
+        return node;
+    }
+    auto *pipeline = _videoReceiver->pipeline();
+    // Do not try to play an already playing pipeline, pause it first.
+    GstState state;
+    gst_element_get_state(pipeline, &state, nullptr, 5);
+    if ( state == GST_STATE_PLAYING ) {
+         gst_element_set_state(pipeline, GST_STATE_PAUSED);
```
jaxxzer: why is this necessary? it seems like we can do nothing if it is already playing

src/VideoStreaming/VideoStreaming.cc
```diff
@@ -28,6 +28,8 @@
 #endif
 
 #include "VideoStreaming.h"
+#include "VideoSurface.h"
+#include "VideoManager.h"
```
jaxxzer: `VideoReceiver.h`?

src/Vehicle/Vehicle.cc
tcanabrava: videoSettings cant be null, it's created before this object and set thru the strange QGCToolBox lifecycle

src/VideoStreaming/VideoSurface.cc
tcanabrava: This is just the first implementation - a followup patch in the sieries implements pause / play. I did this just to make sure things where running and that I was not double-playing it.

src/VideoStreaming/VideoStreaming.cc
tcanabrava: VideoManager is the current class that initializes the VideoReceiver. one of the things I plan to do is to merge two classes. let's see.

src/QmlControls/QGCPipable.qml
```diff
@@ -247,4 +248,32 @@ Item {
             }
         }
     }
+
+    Rectangle {
+        id:                     buttonScreenShoot
```
jaxxzer: `buttonScreenShot`

tcanabrava:
typos on the title, when I submited an update it created a new PR. closing this one.

src/VideoStreaming/VideoStreaming.cc
jaxxzer: @tcanabrava I'm still interested in seeing a reference.

jaxxzer:
> typos on the title, when I submited an update it created a new PR. closing this one.

.. and the new pr is here: https://github.com/mavlink/qgroundcontrol/pull/8261

PR #8089, opened 2019-12-04
Add support for ModalAI FC v1 FW Update

modaltb:
Added support for the ModalAI FC v1 (this will also will cover VOXL-Flight board).

Tested with boot loader update: https://github.com/PX4/Bootloader/commit/d83169d50c333ab33d8206a8a29f6efc9605c39b


modaltb:
Thanks!

Issue #8087, opened 2019-12-04
Mission commands with terrain altitudes are being uploaded as absolute altitudes when uploaded to AP


rmackay9:
We've noticed recently that when a user creates a mission command with altitude frame is set to "Terrain" that it's actually being sent as an absolute altitude.

ArduPilot supports mission commands with the altitude specified as an altitude above terrain so we would much prefer if the alt-frame was sent as MAV_FRAME_GLOBAL_TERRAIN_ALT

Here is the corresponding issue on the ArduPilot issues list: https://github.com/ArduPilot/ardupilot/issues/12917

rmackay9:
@DonLakeFlyer, any thoughts on this one?  We don't think we can fix it on the AP side because as far as we can tell QGC is not sending the terrain alt.

DonLakeFlyer:
QGC does not support MAV_FRAME_GLOBAL_TERRAIN_ALT frame. The "Above Terrain" support is a calculated absolute height based on terrain data and the requested amount above terrain. The docs specify that ArduPilot terrain following is not supported. But it's in release notes as opposed to main docs. I think MP has a sort of similar thing with "Verify Heights" or something like that.

DonLakeFlyer:
Reopening since I thought I had specifically put in some warning text when using ArduPilot and Above Terrain to say that this does not support ArduPilot terrain following.

DonLakeFlyer:
Hmm, seems to have gone missing in action. Need to look into adding that again.

peterbarker:
On Tue, 10 Dec 2019, Don Gagne wrote:

> QGC does not support MAV_FRAME_GLOBAL_TERRAIN_ALT frame. The "Above Terrain" support is a calculated absolute height based on terrain data and the
> requested amount above terrain. The docs specify that ArduPilot terrain following is not supported. But it's in release notes as opposed to main docs. I
> think MP has a sort of similar thing with "Verify Heights" or something like that.

Ah, thanks!

So what it comes to is that QGC calculates a terrain offset and sends that 
to the autopilot as an AMSL height.

That's... interesting.  It immediately means you can't round-trip that 
from the autopilot, of course - but, then, your enhanced GCS-side items 
make that difficult anyway.

I imagine, then, that QGC also can't supply terrain tiles to the autopilot 
when requested?

Do you have a fundamental problem with the concept of a "terrain-relative 
altitude", or is just something which has never really popped up to the 
top of a long list?

[...]

PR #8099, opened 2019-12-10
Warn about Above Terrain usage

DonLakeFlyer:
Reduce confusion with vehicle based terrain following mavlink support:
Screen Shot 2019-12-10 at 3 10 38 PM

Related to #8087.

### DIFF
--- a/src/FactSystem/FactControls/AltitudeFactTextField.qml
+++ b/src/FactSystem/FactControls/AltitudeFactTextField.qml
@@ -1,6 +1,19 @@
-import QtQuick                      2.3
+/****************************************************************************
+ *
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
@hamishwillee Can you look at the wording?

hamishwillee:
I think it is pretty unambigous (i.e. OK). Whether it can be improved depends on whether the use of "set" in the first sentence is significant? Here are some alternative proposals

> 'Above Terrain' sends absolute altitudes to the vehicle (not height above terrain). 

or 

> 'Above Terrain' sends items with absolute altitudes to the vehicle (not height above terrain). Altitudes are calculated from terrain height at the location and the requested altitude above terrain.

Issue #8080, opened 2019-12-02
ROIs are not clearly associated with waypoint and visa versa


hamishwillee:
QGC Daily build 20191202 on Ubuntu 18.04 running in VM.

I believe that when an ROI is defined it becomes active from the preceding waypoint, until the next ROI. There are two problems with the current implementation:
1. It is not at all obvious how you decide "I no longer want to use any ROI" from a particular waypoint (?)
2. It is not at all obvious which part of the mission ROIs belong to.

Specifically for 2, to find out where an ROI starts working you need to click the ROI, then go to mission list, then find and select the previous waypoint, To find out where it ends you would need to travel down the mission item list until you find another ROI definition. 

That seems complicated. Minimally it would be great if selecting an ROI highlights its associated waypoint in some way, and visa versa on the map. Even better would be if you could highlight an ROI and see all segments it applies to.

PR #8101, opened 2019-12-11
Plan: Cancel ROI, ROI visual indication

DonLakeFlyer:
* ROI button will change to Cancel ROI and back at appropriate times
* When an ROI item is selected the flight path lines which are affected by the ROI change color
Screen Shot 2019-12-11 at 7 51 17 AM

Fix for #8080


hamishwillee:
That is really so much better. Thanks.

PR #8108, opened 2019-12-11
Only accept ATTITUDE messages from the vehicle

olliw42:
In a system there might be more than one component emitting a MAVLink ATTITUDE message, for instance a flight controller and a gimbal.

Currently the code is not discriminating against the different sources, with the result that the instrument widget gets confused, i.e. will appear to wildly flip around, as it shows the attitudes of the various sources as they arrive.

This PR adds a check for the sysid and compid of the message and rejects it if they don't match the vehicles IDs.

Works great for my setup.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -1112,6 +1112,11 @@ void Vehicle::_handleAttitudeWorker(double rollRadians, double pitchRadians, dou
 
 void Vehicle::_handleAttitude(mavlink_message_t& message)
 {
+    // only accept the attitude message from the vehicle's flight controller
+    if (message.sysid != _id || message.compid != _compID) {
+        return;
+    }
+
     if (_receivingAttitudeQuaternion) {
         return;
     }
@@ -1124,6 +1129,11 @@ void Vehicle::_handleAttitude(mavlink_message_t& message)
 
 void Vehicle::_handleAttitudeQuaternion(mavlink_message_t& message)
 {
+    // only accept the attitude message from the vehicle's flight controller
+    if (message.sysid != _id || message.compid != _compID) {
+        return;
+    }
+
     _receivingAttitudeQuaternion = true;
 
     mavlink_attitude_quaternion_t attitudeQuaternion;


DonLakeFlyer:
I think the problem is way larger than this. 90% of the _handleFoo implementations are only meant to be handled if they are from the autopilot component. 

Also I hadn't really seen that _compID thing before. Not sure that should be used as the test. It won't always be the autopilot component. It could be a PX4 Flow component id for example. In which case the handlers should be called.

Instead of adding ifs to all the _handleFoo implementations I think the right thing to do is to change Vehicle::_mavlinkMessageReceived to call out into Vehicle::_mavlinkMessageReceivedFromAutopilot and Vehicle::_mavlinkMessageReceivedFromComponent. Then the _handleFoo things go into the switch statement in the appropriate FromAutopilot/Component method.

olliw42:
I totally agree that the problem very likely is bigger and this phenomenon isn't only possible for the attitude message. I don't think there would be an easy solution however, since there are so many messages and in principle each of them would have to be checked individually and carefully. So I would think that going case by case might be the more doable path, and then, once for sufficiently many messages it has been figured out, to comense the code pieces.

I understand the concern with _compID. I did test this for various setup scenarios and for all it worked as intended, and also the code logic pointed me to this However, this certainly doesn't proof anything like completness. 

I think this actually also points to a bigger problem, namely that there is no central place there the info on the seen components is stored. One can find such pieces at several places in the code, but none is systematic or keeps all relevant info (sysid, compid, mavtype, active/live, etc). In the COMPONENT_INFORMATION PR I made an attempt of something which could be evolved into such a central repository. But currently it's a bit missing.

If you are not comfortable with this PR pl feel encouraged to put it on hold, or to close it and move it to an issue so that the issue is not lost.

:)

DonLakeFlyer:
@hamishwillee To me this is another mavlink spec problem. The simplest example is ATTITUDE message. If multiple components are sending that message which one is the one for the vehicle itself.

olliw42:
that's IMHO easy to answer, i.e., has been already resolved: The one with the sys/compId which the vehicle has revealed via its heartbeat, and the vehicle reveals itself via the mav type.

that is, what one would need to do is to analysze the heartbeats and build up a table with the seen sysid/compids and the mavtype, and store that in a central place accessible by "anyone". 

with such a table all such questions can be always answered

:)

hamishwillee:
Pretty much what @olliw42 said. The autopilot component will have the autopilot component ID and will also have a [MAV_TYPE](https://mavlink.io/en/messages/common.html) that is a vehicle type rather than a "component type" -e.g. camera. 

> I think the problem is way larger than this. 90% of the _handleFoo implementations are only meant to be handled if they are from the autopilot component.

I think the MAVLink ecosystem evolved from a very system/autopilot centric point of view. Personally I'm still working on getting my head around the rules for components.

DonLakeFlyer:
> The autopilot component will have the autopilot component ID and will also have a [MAV_TYPE](https://mavlink.io/en/messages/common.html) that is a vehicle type rather than a "component type" -e.g. camera.

Which is how QGC currently does it. The problem with that mechanism is that when a new MAV_TYPEs are added GCS code need to change to take that into account.

What @olliw42 said was that you can determine the main autopilot component ID as well as whether a HEARTBEAT is from the main autopilot component based on MAV_AUTOPILOT != MAV_AUTOPILOT_INVALID. Which I have changed in the pull that's up there. That is a way better way to do it since it is stable with respect to new things getting added after a gcs release is done. From the docs if you poke around enough this seems to be a correct assumption.

DonLakeFlyer:
> that's IMHO easy to answer, i.e., has been already resolved: The one with the sys/compId which the vehicle has revealed via its heartbeat, and the vehicle reveals itself via the mav type.

According to what doc in the mavlink spec? And is that rule true for all messages? What about for things which aren't quite easily recognizable as a directly "vehicle" related like camera messages. Where do those come from? If they come from multiple places how to do you know one is the one you really want. Or another example CAMERA_IMAGE_CAPTURED currently comes through the main autopilot component but CAMERA_INFORMATION comes through a seperate camera component. If you happen to get a CAMERA_INFORMATION from the main autopilot does that override the values from the camera component. Who knows? A table of component ids helps you in know way to understand that ahead of time when you are writing code.

olliw42:
> The autopilot component will have the autopilot component ID and will also have a [MAV_TYPE](https://mavlink.io/en/messages/common.html) that is a vehicle type

@hamishwillee : it would be useful to be super precise in this matter now.

I believe some few months ago there was a longer discussion about it and I believe that it was finally consent that the comp ID is **suggested** to be the standard ID but that it **does NOT HAS** to be the standard ID !! That is, a component may have the id but also may not have it. I was phrasing it in the words that the types must be determined by mav_type and not component id. 

So, I believe the consent was that the sentence must be

_The autopilot component may have the autopilot component ID but will have a [MAV_TYPE] that is a vehicle type_

We should be precise on this

olliw42:
> > that's IMHO easy to answer, i.e., has been already resolved: The one with the sys/compId which the vehicle has revealed via its heartbeat, and the vehicle reveals itself via the mav type.
> 
> According to what doc in the mavlink spec? And is that rule true for all messages? What about for things which aren't quite easily recognizable as a directly "vehicle" related like camera messages. Where do those come from? If they come from multiple places how to do you know one is the one you really want. Or another example CAMERA_IMAGE_CAPTURED currently comes through the main autopilot component but CAMERA_INFORMATION comes through a seperate camera component. If you happen to get a CAMERA_INFORMATION from the main autopilot does that override the values from the camera component. Who knows? A table of component ids helps you in know way to understand that ahead of time when you are writing code.

> According to what doc in the mavlink spec? 

that's a very valid question. As you noted elsewhere the doc isn't very easy on this. I refer to discussions in mavlink git issues some months ago /which there spread out themselves unfortunately), so - with some searching though - I could bring by quotes. The result of that discussion unfortunately hasn't made it yet into the specs. There is an open issue listing this, https://github.com/mavlink/mavlink/issues/1219, last bullet. 

> And is that rule true for all messages? 

IMHO: sure


> What about for things which aren't quite easily recognizable as a directly "vehicle" related like camera messages. Where do those come from? If they come from multiple places how to do you know one is the one you really want. Or another example CAMERA_IMAGE_CAPTURED currently comes through the main autopilot component but CAMERA_INFORMATION comes through a seperate camera component. If you happen to get a CAMERA_INFORMATION from the main autopilot does that override the values from the camera component. Who knows? A table of component ids helps you in know way to understand that ahead of time when you are writing code.

I must say, I really don't understand the issue here. IMHO it's really supersimple: 

The existing components in a MAVLink system must emit a heartbeat, so for any source sysid/compid it is precisely known if such a component exists and of what type it is. This gives a complete picture of the physical and/or logical components in the system. 

It is complete in the sense that it is ALL the information which is available within how the MAVLink protocol is constructed.

Of course, and I think this is what you somewhat confuse with, there can be information one would like to know which are however NOT retrievable through the fabrics of the MAVLink protocol but would require specific messages. And here I agree, MAVLink is in some evolving state (which is GOOD !!!) and therefor some inconsistency or legacies exist in the design of the message set as it momentarily exist (that's unavoidable if one wants evolve without breaking, right).

The camera handling is indeed a very good example to think through and sort out things, and also for 
In your specific example ...


... ohhh, I have to catch my train ... I will continue with outlining my answer to your example

:)

hamishwillee:
Re https://github.com/mavlink/qgroundcontrol/pull/8108#issuecomment-565321120

> I believe some few months ago there was a longer discussion about it and I believe that it was finally consent that the comp ID is suggested to be the standard ID but that it does NOT HAS to be the standard ID !

There was a long discussion. I don't believe there was "agreement" on that. Certainly I would see that as potentially breaking.

hamishwillee:
> What @olliw42 said was that you can determine the main autopilot component ID as well as whether a HEARTBEAT is from the main autopilot component based on MAV_AUTOPILOT != MAV_AUTOPILOT_INVALID. Which I have changed in the pull that's up there. That is a way better way to do it since it is stable with respect to new things getting added after a gcs release is done. From the docs if you poke around enough this seems to be a correct assumption.

Yes, that works as a mechanism for identifying whether a heartbeat is from an autopilot component vs not an autopilot. 
You might also use the the autopilot component id (as per previous comment making this "optional" was discussed but IMO not broadly "agreed" (essentially I thought this might break "stuff", and would not make that decision myself).

I'll leave you an Ollie to fight out the discussion on maintaining a component id table. FWIW I do tend to agree with him that being able to differentiate the source of a message is important for handling.

olliw42:
> > I believe some few months ago there was a longer discussion about it and I believe that it was finally consent that the comp ID is suggested to be the standard ID but that it does NOT HAS to be the standard ID !
> 
> There was a long discussion. I don't believe there was "agreement" on that. Certainly I would see that as potentially breaking.

hhhhhmmmmmm ... I think we really should go back and look this up ... I'm kind of sure there was a PR or sentence or such with the suggested change ... I'm a bit puzzled now since it feels a bit like a step back

that's IMHO now really a VERY VERY crucial point to settle once and for all. The implication of a determined component ID is that one then indeed would NOT have to first inspect the heartbeat, since the id then says it all ... so one is talking here about very basic basics of the fabric of the MAVLink protocol ... IMHO

there is no need to fight over a table or not ... it's a only a code thing ... currently similar tasks are located in several places, which makes me think to centralize it, but one can keep it also that way

LorenzMeier:
@DonLakeFlyer The change looks correct to me.

LorenzMeier (approved):

mrpollo:
I'm going to rebase locally and re-submit!

Issue #5846, opened 2017-11-17
New Guided Action - Set ROI


fredrikfalkman:
This is a feature request for an added funktionality for the Guided Go-to by clicking in the map.

Clicking in the map might give multiple choices aside from the current Go-to:

- Go-to location
- Go-to location and set as ROI
- Set as ROI

DonLakeFlyer:
This will be implemented in 3.6

dogmaphobic:
Done in https://github.com/mavlink/qgroundcontrol/pull/8109

PR #8109, opened 2019-12-11
ROI for manual flight

dogmaphobic:
Provided the autopilot supports it, you can now set a Region Of Interest during manual flight. The camera will keep pointing to it until you cancel it or set another ROI point.

Screen Shot 2019-12-11 at 3 04 09 PM


hamishwillee:
- How do you cancel it? 
- How do you set it?
- How does a user know when this is active?
- How do they know when it is not active?
- FMI how does QGC know autopilot supports it?

If you look at this image you see an ROI being ignored by the vehicle. I know that this is because the vehicle is not in position mode (which I believe is the only mode this works in) for PX4. But there is no way for a user to guess at this.

So FWIW there needs to be visual queues for a user to explain when/how this is working. Minimally perhaps a rename both of the icon and the menu option for setting it to *Position mode ROI*. Grey/translucent icon when not active.

dogmaphobic:
There is more to come. The vehicle did not ignore the ROI. The gimbal was constantly aiming at the ROI point. I will fill in the blanks once I'm done with the PR.

hamishwillee:
OK, thanks. I'll keep quiet until you're done.

dogmaphobic:
If the vehicle supports ROI (for now, it's hardwired for PX4 multirotor only), you will see a new indicator within the toolbar:
Screen Shot 2019-12-11 at 7 16 26 PM

You set an ROI by clicking on a point on the map:
Screen Shot 2019-12-11 at 7 24 49 PM

When an ROI point is set, the indicator lights up:
Screen Shot 2019-12-11 at 7 16 42 PM

Clicking on the indicator gives you the option to cancel ROI:
Screen Shot 2019-12-11 at 7 17 30 PM

DonLakeFlyer:
Merge?

DonLakeFlyer:
Fixes #5846

Issue #7585, opened 2019-07-17
Vehicle setup qml timers remain active after navigating to fly view


jaxxzer:
As noted in #7275 :
> There is another bug involved, and that is that the code/timers in the vehicle setup qml pages keeps running after the fly view is selected. So if motor test page is open in the background (while fly view is active), then it is sending motor test commands after you arm the vehicle. 

This is confirmed to occur on master for the motor setup page, I haven't investigated yet to see if any other setup pages are affected.

DonLakeFlyer:
The pages from each view (Settings, Setup, Plan, Fly) continue to run even if they are not being shown.

DonLakeFlyer:
The pull reference is the first step to providing a mechanism to fix this. When done it will provide a system for a page to prevent a view switch. Hence if you are in the middle of a calibration or an active motor test you won't be able to switch views. Preventing arming will be a little trickier. But view switch is a worse problem I think.

DonLakeFlyer:
> I haven't investigated yet to see if any other setup pages are affected.

They are. Things which are say calibration relation with a start/stop to calibration will be affected.

DonLakeFlyer:
@jaxxzer Once #8113 goes through you'll have the mechanism to fix your motor test page.

jaxxzer:
ok! :clap:

DonLakeFlyer:
@jaxxzer This should get fixed before I cut a stable which is soon

DonLakeFlyer:
Time is running out to get this fix into stable.

Williangalvani:
That was fixed in my latest PR, you can close it!

PR #8112, opened 2019-12-12
View switch prevention support

DonLakeFlyer:
Previously the right side dialogs were parented to the child window the came from. Now they are parented to the main window. This causes a new problem when the right dialog pops up from a validation error. The same basic problem has always existed though when changing pages within a view like Settings pages. The nice thing is that the new way right dialogs are rooted to mainWindow makes it way easier to fix.  For example this sequence:
* App Settings/General
* Type -10 into Default Mission Altitude. Don't move focus away
* Click on any other App Settings button (or switch views from the toolbar)
* The view will switch and you will be left with a either a screwed up validation dialog or a validation dialog showing over the wrong view

In order to fix this there is a new mainWindow.preventViewSwitch bool which dictates whether a view switch should be allowed. This is automatically kept up to date by the right hand dialog code. But anything which wants to do a view switch needs to respect it. 

Currently the App Settings tabs and the Toolbar buttons respect it. And I just realized I need to hook it up to the Setup pages as well. I'll do that in a follow on commit.

### DIFF
--- a/src/QmlControls/QGCToolBarButton.qml
+++ b/src/QmlControls/QGCToolBarButton.qml
@@ -14,20 +14,25 @@ import QGroundControl.Controls      1.0
 import QGroundControl.Palette       1.0
 import QGroundControl.ScreenTools   1.0
 
+// Important Note: Toolbar buttons must manage their checked state manually in order to support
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
This is related to #7585.

PR #8113, opened 2019-12-12
Prevent view switching apis support a stack

DonLakeFlyer:
The prevent view switching methods are all on mainWindow:
* pushPreventViewSwitch() - Call to prevent a view switch, should have a matched popPreventViewSwitch
* popPreventViewSwitch() - Matching call to pop the prevent view switch stack
* preventViewSwitch() - Return true if view switching should be prevented. This is already used in all the right places to prevent view switching. Implementors of things like individual setup pages shouldn't need to call this.

I've changed the current usage over to these new methods. But I have not used them inside any of the setup pages yet. Example usage would be:
* Radio Component
* When the user starts a calibration pushPreventViewSwitch() should be called
* When the user finishes/cancels the calibration popPreventViewSwitch() should be called

Support for fixing #7585

### DIFF
--- a/src/VehicleSetup/SetupView.qml
+++ b/src/VehicleSetup/SetupView.qml
@@ -41,7 +41,7 @@ Rectangle {
 
     function showSummaryPanel()
     {
-        if (mainWindow.preventViewSwitch) {
+        if (mainWindow.preventViewSwitch()) {
             return
         }
         if (_fullParameterVehicleAvailable) {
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
FYI: In case anyone is wondering why this is a stack. It's because a setup page may prevent a view switch and then while prevented pop a right dialog. Those two actions need to stack prevents.

PR #8115, opened 2019-12-12
Component specific handling of mavlink messages

DonLakeFlyer:
Replacement for #8108

This splits Vehicle based mavlink message handling into two categories:
```
    void _mavlinkMessageFromAutopilotOnly(LinkInterface* link, mavlink_message_t message);
    void _mavlinkMessageFromAnyComponent(LinkInterface* link, mavlink_message_t message);
```

The import thing to look at is what message is in each category. This current breakout is a safe guess at what I think makes sense. It leans towards putting things in AnyComponent for safety. The goal is to keep the values which should only be vehicle associated in the AutopilotOnly handling. 

But in reality this is a significant mavlink spec problem in that how do you know which of these messages should come from where? And if the same messages comes from multiple components which one of them applies to the Vehicle itself.

@olliw42, @dogmaphobic Can you guys both take a look a this?

@hamishwillee FYI

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -178,7 +178,7 @@ Vehicle::Vehicle(LinkInterface*             link,
     , _messagesSent(0)
     , _messagesLost(0)
     , _messageSeq(0)
-    , _compID(0)
+    , _mainCompID(0)
     , _heardFrom(false)
     , _firmwareMajorVersion(versionNotSetValue)
     , _firmwareMinorVersion(versionNotSetValue)
@@ -378,7 +378,7 @@ Vehicle::Vehicle(MAV_AUTOPILOT              firmwareType,
     , _messagesSent(0)
     , _messagesLost(0)
     , _messageSeq(0)
-    , _compID(0)
+    , _mainCompID(0)
     , _heardFrom(false)
     , _firmwareMajorVersion(versionNotSetValue)
     , _firmwareMinorVersion(versionNotSetValue)
@@ -818,12 +818,14 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
     _messagesReceived++;
     emit messagesReceivedChanged();
     if(!_heardFrom) {
-        if (message.msgid == MAVLINK_MSG_ID_HEARTBEAT) {
-            // This first hearbeat which comes through the vehicle is always the main autopilot component.
-            _heardFrom = true;
-            _mainCompID = message.compid;
-            _messageSeq = message.seq + 1;
+        // This message to a vehicle is alway a hearbeat from the main autopilot component.
+        if (message.msgid != MAVLINK_MSG_ID_HEARTBEAT) {
+            qWarning() << "Internal error: First message to vehicle is not HEARTBEAT" << message.msgid;
+            return;
         }
+        _heardFrom = true;
+        _mainCompID = message.compid;
+        _messageSeq = message.seq + 1;
     } else {
         if(_mainCompID == message.compid) {
             uint16_t seq_received = static_cast<uint16_t>(message.seq);
@@ -2130,7 +2132,7 @@ int Vehicle::motorCount(void)
             SUB_FRAME_CUSTOM
         };
 
-        uint8_t frameType = parameterManager()->getParameter(_compID, "FRAME_CONFIG")->rawValue().toInt();
+        uint8_t frameType = parameterManager()->getParameter(_mainCompID, "FRAME_CONFIG")->rawValue().toInt();
 
         switch (frameType) {  // ardupilot/libraries/AP_Motors/AP_Motors6DOF.h sub_frame_t
 
@@ -3160,7 +3162,7 @@ void Vehicle::setCurrentMissionSequence(int seq)
         priorityLink()->mavlinkChannel(),
         &msg,
         static_cast<uint8_t>(id()),
-        _compID,
+        _mainCompID,
         static_cast<uint8_t>(seq));
     sendMessageOnLink(priorityLink(), msg);
 }
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -818,13 +818,14 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
     _messagesReceived++;
     emit messagesReceivedChanged();
     if(!_heardFrom) {
-        if(message.msgid == MAVLINK_MSG_ID_HEARTBEAT) {
+        if (message.msgid == MAVLINK_MSG_ID_HEARTBEAT) {
+            // This first hearbeat which comes through the vehicle is always the main autopilot component.
             _heardFrom = true;
-            _compID = message.compid;
+            _mainCompID = message.compid;
             _messageSeq = message.seq + 1;
         }
     } else {
-        if(_compID == message.compid) {
+        if(_mainCompID == message.compid) {
             uint16_t seq_received = static_cast<uint16_t>(message.seq);
             uint16_t packet_lost_count = 0;
             //-- Account for overflow during packet loss
@@ -850,7 +851,7 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
         return;
     }
 
-    if (message.compid == MAV_COMP_ID_AUTOPILOT1) {
+    if (message.compid == _mainCompID) {
         _mavlinkMessageFromAutopilotOnly(link, message);
     } else {
         _mavlinkMessageFromAnyComponent(link, message);
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -635,6 +635,165 @@ void Vehicle::resetCounters()
     _heardFrom          = false;
 }
 
+void Vehicle::_mavlinkMessageFromAutopilotOnly(LinkInterface* link, mavlink_message_t message)
+{
+    switch (message.msgid) {
+    case MAVLINK_MSG_ID_HOME_POSITION:
+        _handleHomePosition(message);
+        break;
+    case MAVLINK_MSG_ID_HEARTBEAT:
+        _handleHeartbeat(message);
+        break;
+    case MAVLINK_MSG_ID_RC_CHANNELS:
+        _handleRCChannels(message);
+        break;
+    case MAVLINK_MSG_ID_RC_CHANNELS_RAW:
+        _handleRCChannelsRaw(message);
+        break;
+    case MAVLINK_MSG_ID_BATTERY_STATUS:
+        _handleBatteryStatus(message);
+        break;
+    case MAVLINK_MSG_ID_SYS_STATUS:
+        _handleSysStatus(message);
+        break;
+    case MAVLINK_MSG_ID_RAW_IMU:
+        emit mavlinkRawImu(message);
+        break;
+    case MAVLINK_MSG_ID_SCALED_IMU:
+        emit mavlinkScaledImu1(message);
+        break;
+    case MAVLINK_MSG_ID_SCALED_IMU2:
+        emit mavlinkScaledImu2(message);
+        break;
+    case MAVLINK_MSG_ID_SCALED_IMU3:
+        emit mavlinkScaledImu3(message);
+        break;
+    case MAVLINK_MSG_ID_VIBRATION:
+        _handleVibration(message);
+        break;
+    case MAVLINK_MSG_ID_EXTENDED_SYS_STATE:
+        _handleExtendedSysState(message);
+        break;
+    case MAVLINK_MSG_ID_AUTOPILOT_VERSION:
+        _handleAutopilotVersion(link, message);
+        break;
+    case MAVLINK_MSG_ID_PROTOCOL_VERSION:
+        _handleProtocolVersion(link, message);
+        break;
+    case MAVLINK_MSG_ID_HIL_ACTUATOR_CONTROLS:
+        _handleHilActuatorControls(message);
+        break;
+    case MAVLINK_MSG_ID_GPS_RAW_INT:
+        _handleGpsRawInt(message);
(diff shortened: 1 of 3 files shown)


olliw42:
as implied by other comments by me, I think the key thing to do is to keep a table of the sys/compId & mavtypes as seen by the heartbeats, and to make it accessible more globally 

(the vehicle class, as it does the central mavlink message handling, could be a good location for that structure)

how to split off the handling of the messages is then - IMHO - really just a matter of convenience and taste but not of principle

there in fact can be cases there a distinction into vehicle/any-on-else isn't the best solutions, as there can be messages which can be relevant to any type of component, such as e.g. the COMPONENT_INFORMATION message, which potentially is emitted by any type of component, also the flight controller.

src/Vehicle/Vehicle.cc
```diff
@@ -840,6 +792,69 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
...
+    if (!_firmwarePlugin->adjustIncomingMavlinkMessage(this, &message)) {
+        return;
+    }
+
+    // Give the Core Plugin access to all mavlink traffic
+    if (!_toolbox->corePlugin()->mavlinkMessage(this, link, message)) {
+        return;
+    }
+
+    if (message.compid == MAV_COMP_ID_AUTOPILOT1) {
```
olliw42: this line of code is IMHO - and I believe I correctly understand the MAVlink spec in that matter - eaxctly what one should NOT do, ie., to identify the components type by the standard IDs. 

This type of approach is found repeatedly in the QGC code, but isn't correct (it may work in most practical cases, but isn't guaranteed to work since it's not aligned with the spec)

DonLakeFlyer:
> I think the key thing to do is to keep a table of the sys/compId & mavtypes as seen by the heartbeats, and to make it accessible more globally

I don't see how that helps. For example if a gcs gets a SCALED_IMU message from MAV_COMP_ID_AUTOPILOT1, MAV_COMP_ID_IMU and MAC_COMP_ID_GIMBAL which one is the right one to use. Heartbeats and MAV_TYPE don't help with that.

src/Vehicle/Vehicle.cc
DonLakeFlyer: > I believe I correctly understand the MAVlink spec in that matter

Where does it say anything at all about this?

olliw42:
to maybe be a bit clearer, my suggestion would be to do something along this, just in e.g. the vehicles main mavlink receive 
... and with the obvious modifications to make it appropriate for the main:
* e.g. broadcast sysid's should be considered, it's here taken out since comp_info must not be broadcast
* e.g. _handleHeartbeatMessage() would when not trigger anything, but just populate the structure, and populate not only the compId but also the type, (and collect the other info available from heartbeat such as armed etc. as it is already done)

https://github.com/mavlink/qgroundcontrol/blob/6f1b186d1a36fc1347519e71960b75e8dc0a427a/src/Vehicle/ComponentManager.cc#L441-L476

DonLakeFlyer:
That only helps somewhat for components which sends a heartbeat, which is not required as far as I know. Eventually at some point you have to case out on something. Pick your poison: COMP_ID, MAV_TYPE, ... Don't make sense to me that one if better than another.

olliw42:
> 
> 
> > I think the key thing to do is to keep a table of the sys/compId & mavtypes as seen by the heartbeats, and to make it accessible more globally
> 
> I don't see how that helps. For example if a gcs gets a SCALED_IMU message from MAV_COMP_ID_AUTOPILOT1, MAV_COMP_ID_IMU and MAC_COMP_ID_GIMBAL which one is the right one to use. Heartbeats and MAV_TYPE don't help with that.

they do, and only they do:

these three components have to, by spec, send out heartbeats with their sysId and compId, and their type populated properly. In your example e.g.

autopilot:
sysid = 1
compID = AUTOPLIOT1 = 1, but note that by spec it can be whatever you want
mav_autopilot = MAV_AUTOPILOT_PX4 (or any other except MAV_AUTOPILOT_INVALID)
mav_type =chose as appropriate, but not any of the compoent types

the entry mav_autopilot != MAV_AUTOPILOT_INVALID identifies it as the autopilot

gimbal:
sysid = 1
compId = GIMBAL1, but note that by spec it can be whatever you want
mav_autopilot = MAV_AUTOPILOT_INVALID
mav_type = MAV_TYPE_GIMBAL

the entries  mav_autopilot == MAV_AUTOPILOT_INVALID && mav_type = MAV_TYPE_GIMBAL identifies it as a gimbal

IMU:
sysid = 1
compId = MAV_COMP_ID_IMU, but note that by spec it can be whatever you want
mav_autopilot = MAV_AUTOPILOT_INVALID
mav_type = not type yet in the spec, pl raise a PR to add MAV_TYPE_IMU

the entries  mav_autopilot == MAV_AUTOPILOT_INVALID && mav_type = MAV_TYPE_IMU identifies it as a imu

In your list you have now this info (since sysid can be assumed to be identical for a vehicle we don't really have to store it)

list[0] = (1, 1, autopilot)
list[1] = (1, 154, gimbal)
list[2] = (1, 200, imu)

So, when an attitude message with sysid = 1 and compId = 154 arrives you can unambiguously determine from that list that its for the gimbal. Similar for the autopilot and imu.

This is btw how MAVLink worked since ever. The later addition of the standard component IDs and the latest clarification that they are recommended but not mandatory doesn't change the basic mechanism that the origin of a message is retrievable from the IDs and the knowledge gained from the heartbeats.

Hope that helps clarifying this.

olliw42:
> 
> 
> That only helps somewhat for components which sends a heartbeat, which is not required as far as I know. Eventually at some point you have to case out on something. Pick your poison: COMP_ID, MAV_TYPE, ... Don't make sense to me that one if better than another.

I was too slow

a MAVLink compliant component must emit a heartbeat

https://mavlink.io/en/services/heartbeat.html: 

 _"The heartbeat protocol is used to advertise the existence of a system on the MAVLink network, along with its system and component id, vehicle type, flight stack, component type, and flight mode."_

_"Components must regularly broadcast their HEARTBEAT"_

DonLakeFlyer:
> Components must regularly broadcast their HEARTBEAT

If that is reality, then I get it. I'll check some things when I get home.

olliw42:
I rarely say I'm sure, but I'm sure it's reality. Ask @hamishwillee

DonLakeFlyer:
@hamishwillee You might want to look at the mavlink docs with respect to a components required to send a heartbeat. I'm still not convinced there are cases where firmware forwards message from some other thing on the vehicle using a different component id and you don't get heartbeats on the comp id.

DonLakeFlyer:
OK, the last commit changes MultiVehicleManager to match spec with respect to detecting a vehicle on a link. And then Vehicle is changed to us the _mainCompId member to case out when to go where.

This clears things up slightly but still not fully as to where to expect messages to come from. For example, currently in QGC the camera manager has it's own mavlink message handler which handles messages from specifically camera components by comp id. Where in the mavlink species it say that for example the CAMERA_INFORMATION message only comes from camera components.

src/Vehicle/Vehicle.h
```diff
@@ -1482,7 +1484,7 @@ private slots:
     uint                _messagesSent;
     uint                _messagesLost;
     uint8_t             _messageSeq;
-    uint8_t             _compID;
+    uint8_t             _mainCompID;
```
olliw42: the naming is maybe a bit unfortunate, for the vehicle's sysid we have _id, and for it's compid we have _mainCompID ... 
maybe streamlining to e.g.

_mainSysID, _mainCompID
_sysID, _compID
_vehicleSysID, vehicleCompID 

or alike could improve code readability

src/Vehicle/Vehicle.cc
```diff
@@ -840,6 +792,70 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
...
+    //-- Check link status
+    _messagesReceived++;
+    emit messagesReceivedChanged();
+    if(!_heardFrom) {
+        if (message.msgid == MAVLINK_MSG_ID_HEARTBEAT) {
+            // This first hearbeat which comes through the vehicle is always the main autopilot component.
+            _heardFrom = true;
+            _mainCompID = message.compid;
+            _messageSeq = message.seq + 1;
+        }
```
olliw42: why relying on this assumption, and not doing the proper mav_autopilot != invalid check?

why not grabbing the opportunity and building up a compid/type table here for use by other code pieces?

DonLakeFlyer: > why relying on this assumption, and not doing the proper mav_autopilot != invalid check?

Because _mainCompID must be set on the first message since it is used elsewhere. If something changes and the assumption breaks all hell will break lose in other parts of the code as well. So it's way safer to check the assumption. Also this last commit was bad.

DonLakeFlyer: > why not grabbing the opportunity and building up a compid/type table here for use by other code pieces?

What other code pieces? Nothing else needs it at the moment as far as I can tell.

olliw42: I don't understand this. And I believe it's incorrect.

There is simply no guarantee that the first heartbeat must be the one from the autopilot. Hence in order to determine the autopilots heartbeat you must properly check by

if (message.msgid == MAVLINK_MSG_ID_HEARTBEAT && message.mav_autopilot !=  MAV_AUTOPILOT_INVALID ) 

that's the only safe since conform test. You are of course allowed to add whatever additional test you like on top of this, but adding these additional tests but not the relevant type test doesn't make it way safer.

Also, you actually don't test that the first heartbeat is from the autopilot! In fact, now you not only assume that the first heartbeat is from the autopilot, but even that the very first message from the autopilot is a heartbeat.

If you are worried of bad things happen before _compId is set, you can add an exist before you go to the separate handlers. This would then be safe and would work.

IMHO

:)

olliw42: if one is worried that some flight stacks don't set their heartbeat correct, one could improve the check to 

if (message.msgid == MAVLINK_MSG_ID_HEARTBEAT && (message.mav_autopilot != MAV_AUTOPILOT_INVALID || message.compid == MAV_COMP_ID_AUTOPILOT1) )

:)

hamishwillee:
> You might want to look at the mavlink docs with respect to a components required to send a heartbeat. I'm still not convinced there are cases where firmware forwards message from some other thing on the vehicle using a different component id and you don't get heartbeats on the comp id.

That is what the spec says, and what I believe was reviewed by Lorenz. However whether he checked this specific point carefully enough I do not know. Also whether everything is compliant is another unknown. 

I have not questioned this, and it makes sense to me because of the routing rules. To route a non-broadcast message to another component on a different channel, the routing component must already have received a heartbeat from that component. 

I'm not certain exactly what you are saying above. I *think* you are saying that an autopilot might act as a proxy for its connected components; resending our relevant messages with its own component id and swallowing any heartbeat (if it was sent at all). That I guess is possible, but how would you know?

src/Vehicle/Vehicle.cc
DonLakeFlyer: > There is simply no guarantee that the first heartbeat must be the one from the autopilot.

From a mavlink spec standpoint you are correct. From a QGC standpoint of how MultiVehicleManager (which creates Vehicles) and Vehicle code interact you are incorrect.

DonLakeFlyer:
> I'm not certain exactly what you are saying above.

What I'm saying is that when I look at the mavlink docs, the concept of components requiring a heartbeat and the component ids are spread across doc which don't really give an overview of a component itself. There is doc on "Heartbeat Protocol" which "mentions" components. There is also doc on MAV_COMPONENT:
"_Component ids (values) for the different types and instances of onboard hardware/software that might make up a MAVLink system (autopilot, cameras, servos, GPS systems, avoidance systems etc.). Components must use the appropriate ID in their source address when sending messages. Components can also use IDs to determine if they are the intended recipient of an incoming message. The MAV_COMP_ID_ALL value is used to indicate messages that must be processed by all components. When creating new entries, components that can have multiple instances (e.g. cameras, servos etc.) should be allocated sequential values. An appropriate number of values should be left free after these components to allow the number of instances to be expanded._"
Which mentions nothing about heartbeats.

Then if you look at the HEARTBEAT message spec there are some rules in there which state different settings in there if you are component.

I think to make this completely clear there should be a specific section dedicated to the concept of a component. Which specifies the rules all in one place.

DonLakeFlyer:
FYI: I'm going to wait until I get home to test this some more before I merge. I want to make sure with testing that the mav_autopilot != MAV_AUTOPILOT_INVALID thing is reality as opposed just mavlink spec. If not I'll go back to what I was doing before for vehicle detection in MultiVehicleManager.

olliw42:
> To route a non-broadcast message to another component on a different channel, the routing component must already have received a heartbeat from that component.

I think not. At least ArduPilot doesn't do it that way. ArduPilot opens a route as soon as it sees ANY message with a new source id !! It doesn't wait for heartbeats to open. 

What this effectively means is that messages of a something which emits mavlink messages but not the heartbeat would be routed (I refrained from calling this something a component since in my view a component must emit heartbeats).

src/Vehicle/Vehicle.cc
olliw42: I agree and confirm that I don't know much about the interaction between MultiVehicleManager and Vehicle code. :)
I wonder however what sort of interaction this could be so that it so vulnerable to what is happening in the vehicle's mavlink receive handling. My feeling is there is then something to improve in architecture for some future version.

olliw42:
> FYI: I'm going to wait until I get home to test this some more before I merge. I want to make sure with testing that the mav_autopilot != MAV_AUTOPILOT_INVALID thing is reality as opposed just mavlink spec. If not I'll go back to what I was doing before for vehicle detection in MultiVehicleManager.

my feeling is that it maybe would be better to leave it as is and subject it to a consistent & thorough analysis some when later and not to rush into some changes now.

From my side, the attitude thing, which triggered this here, isn't really very important in the sense of being super urgent.

PR #8121, opened 2019-12-13
Remove Fly/Plan from toolstrips

DonLakeFlyer:
These were added for a specific custom build. I'm wrapping up stable and they need to go now since they still have not been implemented correctly. I've commented out what I could so they can be put back correctly.


DonLakeFlyer:
@dogmaphobic FYI

Issue #8086, opened 2019-12-04
QGC using MISSION_ITEM rather than MISSION_ITEM_INT when uploading to AP


rmackay9:
I've noticed that when QGC is uploading missions to AP (i.e. Copter-4.0.0) it's using the less accurate MISSION_ITEM message rather than MISSION_ITEM_INT.  The equvalent report for ArduPilot is over here: https://github.com/ArduPilot/ardupilot/issues/12961

DonLakeFlyer:
QGC requires a response to the AUTOPILOT_VERSION message with the item_int capability bit set for it to use MISSION_ITEM_INT. I don't have an ArduPilot setup on my laptop right now (and don't have enough internet to set it up) so I can't test ArduPilot. But I did test PX4 and it worked as expected. There "shouldn't" be a difference in this negotiation between PX4 and ArduPilot.

If you turn on logging for VehicleLog it will show you the capabilities which are being set at first connect. You should also be able to see the AUTOPILOT_VERSION message which QGC saw in the Mavlink Inspector.

peterbarker:
@DonLakeFlyer I've just had another play with this.  I can confirm that it is using `MISSION_REQUEST_INT` to *download* items correctly.  However, at the same time, it is *uploading* items using `MISSION_ITEM`.  And the reason is... it's kinda working according to spec!

Both MissionPlanner and MAVProxy work out-of-spec; they convert requests for `MISSION_ITEM` into `MISSION_ITEM_INT` - and ArduPilot only ever sends requests for `MISSION_ITEM` (https://github.com/ardupilot/ardupilot/blob/master/libraries/GCS_MAVLink/MissionItemProtocol.cpp#L318).  MissionPlanner checks the vehicle's compatibility bits before doing that.

For PX4 Firmware (If my reading is correct), if you don't download the items first (giving the GCS an opportunity to indicate it can do _INT), it will MISSION_ITEM.  I haven't delved, but it looks like you may get interesting effects with the `_using_int` state if there are multiple GCSs floating around.

Really, there should probably have been `MISSION_COUNT_INT` and `MISSION_REQUEST_LIST_INT` to disambiguate all of this.

DonLakeFlyer:
@peterbarker ArduPilot doesn't send the mission item int capability bit? That is what triggers upload int usage.

DonLakeFlyer:
[...]

PR #8122, opened 2019-12-13
Hold: Only use MISSION_ITEM_INT

DonLakeFlyer:
This pull request relates to a **breaking** mavlink mission protocol spec change I want to make which would need to be coordinate/testing across al major GCS makers. The change is basically to always use MISSION_ITEM_INT no matter what. I have tested this pull against PX4 and it works fine. Given discussion in #8086 I assume it will work fine for PX4 as well. But I won't be able to test that till next week.

Here it is:
* Mark MISSION_ITEM, MISSION_REQUEST and MAV_PROTOCOL_CAPABILITY_MISSION_INT as deprecated
* Change spec for MISSION_REQUEST to always respond with MISSION_ITEM_INT item **NOT** MISSION_ITEM. This is a breaking change.

Related to #8086

### DIFF
--- a/src/MissionManager/PlanManager.cc
+++ b/src/MissionManager/PlanManager.cc
@@ -348,25 +348,14 @@ void PlanManager::_requestNextMissionItem(void)
     qCDebug(PlanManagerLog) << QStringLiteral("_requestNextMissionItem %1 sequenceNumber:retry").arg(_planTypeString()) << _itemIndicesToRead[0] << _retryCount;
 
     mavlink_message_t message;
-    if (_vehicle->capabilityBits() & MAV_PROTOCOL_CAPABILITY_MISSION_INT) {
-        mavlink_msg_mission_request_int_pack_chan(qgcApp()->toolbox()->mavlinkProtocol()->getSystemId(),
-                                                  qgcApp()->toolbox()->mavlinkProtocol()->getComponentId(),
-                                                  _dedicatedLink->mavlinkChannel(),
-                                                  &message,
-                                                  _vehicle->id(),
-                                                  MAV_COMP_ID_AUTOPILOT1,
-                                                  _itemIndicesToRead[0],
-                _planType);
-    } else {
-        mavlink_msg_mission_request_pack_chan(qgcApp()->toolbox()->mavlinkProtocol()->getSystemId(),
+    mavlink_msg_mission_request_int_pack_chan(qgcApp()->toolbox()->mavlinkProtocol()->getSystemId(),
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@hamishwillee, @peterbarker, @rmackay9, @wickedshell FYI

If someone wants to use this through mavlink before I can do it. That's fine. I won't be able to get to it till mid-next week. If at all possible I'd like to release the QGC Stable which is going out with this change.

I don't think there is any way to do with with a breaking spec change. But that said I think the simplification is worth it.

DonLakeFlyer:
MY other proposal would be to allow MISSION_ITEM_INT.x/y to be something like max int to represent a NaN like optional params work for MISSION_ITEM.

hamishwillee:
> If someone wants to push this through mavlink before I can do it. That's fine. I won't be able to get to it till mid-next week. If at all possible I'd like to release the QGC Stable which is going out with this change.
> I don't think there is any way to do with with a breaking spec change. But that said I think the simplification is worth it.

The "right" MAVLink way to do this would be to deprecate `MISSION_REQUEST`, `MISSION_ITEM` and leave `MAV_PROTOCOL_CAPABILITY_MISSION_INT` alone. Then add a new capability or version MAV_PROTOCOL_CAPABILITY_MISSION_VER2 so there is an easy way to detect the break behaviour (things that no longer support the `MISSION_REQUEST`).

Whether or not we make this break, does it not make sense to make the change easy to detect?
Further, if we do this, we should also try pull in this change to the protocol: https://github.com/mavlink/mavlink/issues/1171#issuecomment-538161203

NOTE: I haven't been tracking the discussion properly - why can't stacks stop using `MISSION_REQUEST` and use `MISSION_REQUEST_INT` as per the spec?

@auturgy FWIW this is the kind of thing formalised microservices versions helps. This is now messy.

> MY other proposal would be to allow MISSION_ITEM_INT.x/y to be something like max int to represent a NaN like optional params work for MISSION_ITEM.

I don't understand what you are suggesting here. Something like `MISSION_ITEM_INT.x -- PARAM5 / local: x position in meters * 1e4, global: latitude in degrees * 10^7. UINT32_MAX indicates not supported`. What's the benefit?

auturgy:
@hamishwillee sigh. I’m past arguing about that. Simply put, juice not worth the squeeze. Let’s not spread that debate to here though.

DonLakeFlyer:
I'm going to merge this in even though it isn't quite to spec. It works fine with both PX4 and ArduPillot.

DonLakeFlyer:
Waiting till after stable and I put up the mavlink discussion before this gone in.

PR #8123, opened 2019-12-13
Update GUI Tests

Adakar:
- Fix issue with reading and setting values of the drone parameters like
speed, altitude. After this change the framework should be able to to
get and set values as long as the value is separated from the unit
abbreviation with a white space char.

- Update real names of the GUI elements from the MAVLink Inspector view

- Rename suite_analyze/tst_mavlink_inspector to
tst_mavlink_menu_entries_width and change it so it checks if the size of
the displayed messages is correct but doesn't check which messages are
displayed. This should be covered by another test case.

- Add "sim" and "no-sim" tags to test cases based on the fact if a
test requires a running simulator

- New test suite for testing Map Zoom Level

- Some minor cleanups


dogmaphobic:
Perfect. Thanks @Adakar!

PR #8126, opened 2019-12-14
Must manually check the button when clicked.

dogmaphobic:
Otherwise, "Log Download" remains the only selected button:

Screen Shot 2019-12-14 at 12 25 11 PM


DonLakeFlyer:
I broke this with all the don't allow view switch stuff right?

dogmaphobic:
> I broke this with all the don't allow view switch stuff right?

I don't think it's related. I had noticed similar behavior elsewhere even before you made those changes.

dogmaphobic:
My guess is the use of `button` from QtQuick 1.x (and the ExclusiveGroup stuff). The only reason I have not switched them yet is because the `style` stuff is all gone and customization is now done in a different way. Not that complicated but who has the time? :P

PR #8152, opened 2019-12-22
Remove OpenSSL Android libs

LorenzMeier:
These should be pulled in either locally through a download or a GIT submodule
as maintaining a security-related library in-tree is not acceptable from a
cybersecurity perspective. The risk for it to be out-of-date and a security
risk is too high.


DonLakeFlyer:
Using a submodule changes nothing with respect to being out of date. Any mechanism requires some manual process to update. Doesn't really matter what that is.

PR #8178, opened 2019-12-25
Deploy tools: Make paths consistent and add QGStreamer

LorenzMeier:
This makes it easier to deploy on remote targets and also includes iOS for GStreamer.

@DonLakeFlyer I found the previous naming convention unnecessarily complex (two ways of specifying Qt versions, you never know which one was meant).

This will one-time break the Travis scripts as you need to add the last dot release, but that shouldn't be an issue once you update.

### DIFF
--- a/deploy/MakeQtTravisTarball.sh
+++ b/deploy/MakeQtTravisTarball.sh
@@ -11,16 +11,27 @@ if [ ! -d ${QT_DIRECTORY} ]; then
 	exit 1
 fi
 
-QT_FULL_VERSION=5.11.0
-QT_BASE_VERSION=5.11
+QT_FULL_VERSION=5.12.6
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
The naming convention is on purpose. The directory names do not include the point version so that it requires less updates to travis scripts as you move from 5.12.*++ to another. That said the travis scripts could be changed to make this easier using variables. But for now I'd rather not change these since I have them all working. I'm tired of futzing with getting stable builds together and I still have lots of other stuff that needs to be done. I don't want to have to go back and do it all over again. The tarbarll shell scripts don't really work. I took a pile of notes when I created the latest set so I could update the scripts. But I won't get to that till later.

DonLakeFlyer:
Closing since it's reference by the Issue which is open

PR #8191, opened 2019-12-30
Changing file encoding(UTF8 to UTF8 with BOM) to fix compiling error

wuuhii:
Changing file encoding(UTF8 to UTF8 with BOM) to fix compiling error


DonLakeFlyer:
@wuuhii Don't quite get why only these two files are screwed up? Any ideas?

wuuhii:
> 
> 
> @wuuhii Don't quite get why only these two files are screwed up? Any ideas?

May be these files contain non-ascii characters

DonLakeFlyer:
> May be these files contain non-ascii characters

Hmm, they shouldn't. @dogmaphobic  any ideas here? Wondering if these files need to be cleaned out somehow from crap chars?

dogmaphobic:
> @dogmaphobic any ideas here? Wondering if these files need to be cleaned out somehow from crap chars?

I don't see anything weird. No non ASCII characters are there:

Screen Shot 2019-12-31 at 2 58 09 PM

wuuhii:
图片
Maybe I make a mistake. It is just a warning. But I don't know how to cancel the compiler option.

dogmaphobic:
Do you have some editor that shows the hex dump of the file? Maybe something gets changed when you clone it? Trying looking at the place where the error occurs and see if there are indeed Unicode characters in there.

wuuhii:
I'll try tomorrow

PR #8193, opened 2019-12-30
Don't show GCS heading

DonLakeFlyer:
99% of the time the heading value is random since it comes from the location information and/or or device GPS. And that heading value is only valid if it is moving fairly rapidly in a direction.


DonLakeFlyer:
I'm going to do this in a different way which continues to allow custom build position manager overrides to specify the heading. Under the assumption they are sending quality data.

PR #8199, opened 2020-01-02
MAVLink Graphing

dogmaphobic:
Adding graphing to the MAVLink Inspector.

The UI currently allows you to create two charts and each chart is limited to 6 series. This is arbitrary. The hardware you are using will ultimately determine the real limits.

* Charting is disabled for Android 32-bit (MAVLink message disassembly is not working on 32-bit ARM)
* QGCApplication now derives from `QApplication` for all platforms. I'm not sure why it was limited to desktop. QtCharts require an `QApplication` derived application (it doesn't support `QGuiApplication`).
* Made the MAVLink Inspector page fully resizable.

Minor:
* Made the checkbox control "gray" when disabled.
* Made the model data for the Analyze Page come from the CorePlugin API instead of having it hardcoded within the QML page.

Android (64-Bit)
screen

Windows
Capture

Ubuntu 18.04
Screenshot from 2020-01-02 13-21-32

macOS showing with two plots:
Screen Shot 2020-01-02 at 3 14 28 PM

The MAVLink Inspector page, like all Analyze pages is detachable (on non mobile platforms):
Screen Shot 2020-01-02 at 3 17 08 PM


DonLakeFlyer:
Cool stuff!

DonLakeFlyer:
> * MAVLink message disassembly is not working on 32-bit ARM

This isn't fixable?

LorenzMeier:
Overall this is really awesome! I do have a couple of minor things that I will collect in an issue, but nothing major. CPU load: No plots 30% (good, means there is no overhead there if not needed), one plot: About 50%, two plots: about 70%. Given how resource intense high-speed plotting is, this seems acceptable.

LorenzMeier:
I've captured the TODO here: https://github.com/mavlink/qgroundcontrol/issues/8200

jafrado:
Wow! It's stellar; thank you guys! On the performance issue, it pays to render Qt Graphics on a well endowed GL machine (sans the typical generic integrated graphics). For example, on Windows with an RTX1080, if you change the GPU affinity, those loading numbers come down an additional 10-20% for 4 graphs so it works well on a debug station and tracks HW capabilities. I have one feature request - use QSurface3D as in https://doc.qt.io/qt-5/qtdatavisualization-qmloscilloscope-example.html so one can view state trending over time.

bkueng:
Very cool, thanks Gus!
I noticed some issues:
- a vehicle disconnect and reconnect does not clean some state, the entries appear twice:
image
- switching back to another tab (e.g. MAVLink Console) and back does not remember the state and plots.

dogmaphobic:
> * a vehicle disconnect and reconnect does not clean some state, the entries appear twice:

This is a weird one. If a vehicle disconnects, the entire state (vehicle instance along with all messages) should have been deleted. I need to take a look.

>switching back to another tab (e.g. MAVLink Console) and back does not remember the state and plots.

I need to think how to go about this one. This is a function of preventing QGC from consuming too much memory (an issue on mobile devices). When you switch away to something else, the view is _replaced_, shutting down the old one. On non mobile builds, you can "detach" those windows. In that case, they remain active within their own window while you load different analyze views.

Saving state is in my todo list. The idea you can name and save/load specific sets of datapoint, range, scale etc.

>EDIT: also the list of mavlink messages could be sorted.

It is sorted by message ID. I can change it to sort alphabetically.

bkueng:
>  It is sorted by message ID. I can change it to sort alphabetically.

Yes alphabetically would make it easier to find a message.

PR #8202, opened 2020-01-03
Continue work on MAVLink Graphing

dogmaphobic:
* Enable charting on Android 32-bit
* Skip message parsing if the message is not being consumed.
* Work around message disassembly issue on ARM 32-bit.
* Fix memory leaks.

Two charts, six series running on a 6-year-old, dual core Nexus 9 (Android 32-bit):
screen


LorenzMeier:
Awesome!

Issue #3946, opened 2016-08-15
Support RTK gps connection on Android


DonLakeFlyer:

geofrancis:
i would very much like this feature, it would reduce the ground station to a phone using remote RTCM

dlwalter:
@DonLakeFlyer Is there any feasible path forward for this feature or is it an underlying hardware limitation?  We would be very interested in supporting this however we can.

DonLakeFlyer:
It's not a hardware limitation. It's a problem since QGC android serial port support is completely different than everything else. Hence it doesn't just come for free like most other cross platform stuff. So it's some special case work for android builds.

DonLakeFlyer:
I tried some quick changes to make this work. With that I'm not sure if there is a hardware or android os limitation here. I can open the port for the RTK but I can't read from it. Could be some sort of driver problem which doesn't exist an android devices. Not sure.

dlwalter:
Do you have a branch you started any work on?

DonLakeFlyer:
No I tossed it. But it's pretty easy to get started. Look at LinkManager.cc and see where the rtk gps is commented out with and ifdef __mobile__. Change that to an ifdef __ios___ and then keep pull the string of compiling telling you things are missing. Basically RTK should only be out for ios buid. When you're all done you'll find that in GPSProvider which providers callbacks for serial port read/writes that for some reason although the ports opens you can't read anything from it.

DonLakeFlyer:
Theoretically if you can get the serial port reading/writing it should just work.

tilaktilak:
On that topic, I started to have a look at how to connect via Wifi to a base station. I think it's probably more practicable for a tablet, solve the serial issue and should represent a similar amount of work. What do you think ?

DonLakeFlyer:
> connect via Wifi to a base station.

Are you talking about NTRIP? That would be a great feature as well. Lots of folks have asked.

tilaktilak:
I was thinking of opening a TCP or UDP socket that streams RTCM and as QSocket and QSerial are parent classes, use the same code to parse RTCM. 
That should be as complex as trying to fix serial on Android and be beneficial for all platforms :-)

DonLakeFlyer:
Was is the thing that is going to stream RTCM? Do those exist?

tilaktilak:
Any wifi base station should do that. 
[...]

PR #8210, opened 2020-01-10
[WIP- NO MERGE] Android RTK Sandbox

dlwalter:
This is just a test branch to sandbox development for RTK on Android

### DIFF
--- a/src/GPS/GPSProvider.cc
+++ b/src/GPS/GPSProvider.cc
@@ -208,7 +208,7 @@ int GPSProvider::callback(GPSCallbackType type, void *data1, int data2)
         }
         case GPSCallbackType::writeDeviceData:
             if (_serial->write((char*) data1, data2) >= 0) {
-                if (_serial->waitForBytesWritten(-1))
+                //if (_serial->waitForBytesWritten(-1))
                     return data2;
             }
             return -1;
--- a/src/GPS/GPSProvider.cc
+++ b/src/GPS/GPSProvider.cc
@@ -98,7 +98,6 @@ void GPSProvider::run()
 
         qCDebug(RTKGPSLog) << "Configuring RTK Base Station";
         if (gpsDriver->configure(baudrate, GPSDriverUBX::OutputMode::RTCM) == 0) {
-            qCDebug(RTKGPSLog) << "Configuration Set";
             /* reset report */
             memset(&_reportGpsPos, 0, sizeof(_reportGpsPos));
 
@@ -114,13 +113,11 @@ void GPSProvider::run()
 
                     if (helperRet & 1) {
                         publishGPSPosition();
-                        qCDebug(RTKGPSLog) << "publishGPSPosition";
                         numTries = 0;
                     }
 
                     if (_pReportSatInfo && (helperRet & 2)) {
                         publishGPSSatellite();
-                        qCDebug(RTKGPSLog) << "publishGPSSatellite()";
                         numTries = 0;
                     }
                 } else {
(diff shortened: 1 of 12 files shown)


dlwalter:
https://github.com/mavlink/qgroundcontrol/issues/3946

src/GPS/GPSProvider.cc
```diff
@@ -202,7 +203,7 @@ int GPSProvider::callback(GPSCallbackType type, void *data1, int data2)
         }
         case GPSCallbackType::writeDeviceData:
             if (_serial->write((char*) data1, data2) >= 0) {
-                if (_serial->waitForBytesWritten(-1))
+                //if (_serial->waitForBytesWritten(-1))
```
dlwalter: @bkueng It looks like this may be the problem on android.  I can get RTK in a working state if I comment out this check.

src/comm/LinkManager.cc
```diff
@@ -620,7 +622,6 @@ void LinkManager::_updateAutoConnectLinks(void)
             }
         }
     }
-
 #ifndef __android__
```
dlwalter: @DonLakeFlyer Is there a way around this "android only supports a single serial connection"?  Having a SiK Radio plugged in via FTDI and the RTK GPS results in QGC only detecting only one or the other.

src/comm/LinkManager.cc
```diff
@@ -673,8 +674,19 @@ void LinkManager::_updateAutoConnectLinks(void)
         _autoConnectRTKPort.clear();
     }
 #endif
+#else
```
dlwalter: @DonLakeFlyer The "Check for RTK GPS connection gone" on line 669 results in the RTK GPS stopping immediately after configuration.  Is there a good way to handle the disconnection that you guys can recommend?

src/GPS/GPSProvider.cc
bkueng: Hmm, I don't remember why I added this. It might not be required at all.

src/comm/LinkManager.cc
DonLakeFlyer: Possibly but I'm not quite sure what it would take to make that work.

src/GPS/GPSProvider.cc
DonLakeFlyer: Is waitForBytesWritten simply not support by the android serial port code? Might not be.

dakejahl:
@dlwalter what's the status with this PR? Did you have any luck? I'd like to get this support into QGC, how can I help?

dlwalter:
It's been a while since I've looked at it so I'll have to refresh.  The gist of the problem is that I couldn't get android and QGC to connect to 2 serial devices at the same time.  So I could get the RTK GPS connected and running but couldn't also have a 3DR radio connected for telemetry to the aircraft.

src/comm/LinkManager.cc
dakejahl: How does QGC differentiate between connected serial devices? It looks like it does it by the vendorID and productID, in which case most FTDIs are the same ("vendorID": 1027, "productID": 24577)

See this https://github.com/mavlink/qgroundcontrol/blob/master/src/comm/USBBoardInfo.json#L36

dakejahl:
> So I could get the RTK GPS connected and running but couldn't also have a 3DR radio connected for telemetry to the aircraft.

Is this because the RTK connected via FTDI is getting discovered incorrectly? Any idea how the desktop case handles this properly?

dlwalter:
I believe RTK devices like ublox enumerate as a CDC device, not FTDI.

dakejahl:
Ohh gotcha, the gps unit is connected via USB. Okay well I'm going to take a look at this soon, I'll report back if/when I make any progress.

dlwalter:
Yeah, I also tried changing the vendor/product ids in QGC to use a SLAB UART device to connect via serial but couldn't get both that and the radio to work simultaneously.

PR #8218, opened 2020-01-14
Joystick: Use find_package for mac and linux

patrickelectric:
Fix cmake build in macOS

### DIFF
--- a/src/Joystick/CMakeLists.txt
+++ b/src/Joystick/CMakeLists.txt
@@ -24,24 +24,7 @@ target_link_libraries(Joystick
 
 target_include_directories(Joystick PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
-if(LINUX)
-        find_package(SDL2 REQUIRED)
-        if (IS_DIRECTORY ${SDL2_INCLUDE_DIRS})
-            include_directories(${SDL2_INCLUDE_DIRS})
-            string(STRIP "${SDL2_LIBRARIES}" SDL2_LIBRARIES)
-            target_link_libraries(Joystick PRIVATE ${SDL2_LIBRARIES})
-        else()
-            include_directories(${SDL2_DIR})
-            target_link_libraries(Joystick PRIVATE SDL2::SDL2)
-        endif()
-
-elseif(APPLE)
-        include_directories(${CMAKE_SOURCE_DIR}/libs/lib/Frameworks/SDL2.framework/Headers)
-        target_link_libraries(Joystick PRIVATE -F${CMAKE_SOURCE_DIR}/libs/lib/Frameworks "-framework SDL2")
-        set_target_properties(Joystick PROPERTIES
-                        MACOSX_BUNDLE YES
-        )
-elseif(WIN32)
+if(WIN32)
 	include_directories(libs/lib/sdl2/msvc/include)
 	# Support both 32 and 64 bit builds
 	if (${CMAKE_SIZEOF_VOID_P} MATCHES 8)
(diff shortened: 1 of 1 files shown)


patrickelectric:
@dagar more love for cmake

DonLakeFlyer:
cmake isn't really being used anywhere. And I'm also not super interested in keeping it up to date. It's hard enough for me to support a single build system. It's impossible to support two. Unless cmake becomes the standard for the CI system for all OS (which it can't as far as I understand) then I don't seen any reason for it to exist.

tcanabrava:
Of course it can, cmake is the industry default for ages and qmake is being
deprecated, qt itself is moving to cmake.

DonLakeFlyer:
Then prove it and put together a pull which switches all CI environments over to cmake and supports all of the custom build features in the current system and also convinces Gus that he can use it to create iOS builds which can be pushed to the iOS store. Until then for me cmake is just an annoyance which I don't have time to deal with.

> qt itself is moving to cmake.

That doesn't really mean anything. Qt doesn't release to all the places we do.

patrickelectric:
> Then prove it and put together a pull which switches all CI environments over to cmake and supports all of the custom build features in the current system and also convinces Gus that he can use it to create iOS builds which can be pushed to the iOS store. Until then for me cmake is just an annoyance which I don't have time to deal with.

If all PRs that improve the cmake build system get merged this will be possible.. 
There is two options, **remove cmake** entirely from QGC or **merge the PRs that improve and fix cmake**.

dagar:
For cmake we need to either figure out the workflow for Android and iOS build/deploy/upload that's equivalent to QtCreator with qmake or wait until Qt switches over entirely. Once Android and iOS are in place I volunteer to fill in all the other gaps.

After that literally everything is better in the cmake world. Finding dependencies, creating packages, handling platform differences, and even build times.

@DonLakeFlyer  I thought gradually getting cmake support in place piece by piece would be the lazy way to switch over some day, but if it's a problem then please ignore it and defer to me.

patrickelectric:
<3 @dagar

DonLakeFlyer:
> I thought gradually getting cmake support in place piece by piece would be the lazy way to switch over some day, but if it's a problem then please ignore it and defer to me.

@dagar The problem is that gradually has equaled more than a year++. That's a bit more than gradual! I don't care much what system is used I just care that I'm not left cleaning up a mess of missing stuff. To me it's time to as they say "Shit or get off the pot!". The requirements are:
* Build all OS versions in CI, travis and appveyor
* Can build both Auterion and Sentera custom build as test cases for custom support
* Can build iOS and android builds which can be pushed to store
* Android build should be push to store from CI
* iOS build would still be some manual magic that Gus does. If cmake could automate that as well well even better. But were aren't there yet so that isn't a requirement.

DonLakeFlyer:
Also any full switchover needs to wait until 4.0 goes out. But certainly switching right away after that would make sense.

DonLakeFlyer:
The other thing is that for android using Qt Creator and the simulator is never used. But for iOS I also the Xcode build (Spit out by qmake) plus run QGC in the simulator all the time. So that support can't disappear either.

dagar:
> > I thought gradually getting cmake support in place piece by piece would be the lazy way to switch over some day, but if it's a problem then please ignore it and defer to me.
> 
> @dagar The problem is that gradually has equaled more than a year++. That's a bit more than gradual! I don't care much what system is used I just care that I'm not left cleaning up a mess of missing stuff. To me it's time to as they say "Shit or get off the pot!". The requirements are:
> 
> * Build all OS versions in CI, travis and appveyor
> * Can build both Auterion and Sentera custom build as test cases for custom support
> * Can build iOS and android builds which can be pushed to store
> * Android build should be push to store from CI
> * iOS build would still be some manual magic that Gus does. If cmake could automate that as well well even better. But were aren't there yet so that isn't a requirement.

Actually I'm fairly confident most if not all of those particular areas will be saner. One of the reasons I started going down the cmake path is that handling all of these things with qmake and our own crappy scripts on the side was such a dumpster fire.

My only hesitation in pushing for a full switchover sooner was the out of box QtCreator device debug/deploy experience for Android and iOS. It's been quite a while since I've last tried (probably the "more than a year++" you mentioned) so maybe time to take another look.

@tcanabrava any comment or suggestions?

DonLakeFlyer:
> was the out of box QtCreator device debug/deploy experience for Android and iOS

Yup. I missed the debug/deploy as important for Android as well.

Issue #7382, opened 2019-04-27
QGroundControl 3.5.1 won't start on Ubuntu 16.04


sfalexrog:
QGroundControl v3.5.1 for Linux (packaged as an AppImage) won't start on Ubuntu 16.04.6 LTS due to missing `libicui18n.so.56`. Current ICU package version for Ubuntu 16.04.6 is 55.

## Expected Behavior
QGroundControl should launch.

## Current Behavior
QGroundControl fails to launch with the following message:

```bash
$ ./QGroundControl.AppImage 
/tmp/.mount_4m096M/QGroundControl: error while loading shared libraries: libicui18n.so.56: cannot open shared object file: No such file or directory
```

## Steps to Reproduce:

1. Update packages on Ubuntu 16.04 to their latest versions
2. Download the QGroundControl v3.5.1 AppImage for Linux and set run permissions (`chmod +x ./QGroundControl.AppImage`)
3. Open terminal and navigate to QGroundControl.AppImage
4. Try to run the image (`./QGroundControl.AppImage`)

## System Information

- Operating System: Ubuntu 16.04.6 LTS
- QGC Version: 3.5.1
- QGC build: stable

jnunezortuno:
Also here with:

Operating System: Ubuntu 18.10 (cosmic)
QGC Version: 3.5.1

Have no problems with previous 3.5.0 version

sfalexrog:
Looks like the bundled Qt libraries link explicitly to `libicui18n.so.56`, `libicuuc.so.56` and `libicudata.so.56`. The AppImage used to contain these libraries with their version suffixes, but now it contains these libraries without the suffixes.

I've tried unpacking the AppImage and creating symlinks for the missing files (`libicudata.so.56 -> libicudata.so`, `libicui18n.so.56 -> libicui18n.so`, `libicuuc.so.56 -> libicuuc.so`), and this allowed QGroundControl to launch.

I believe this bug might have something to do with this commit: https://github.com/mavlink/qgroundcontrol/pull/7222/commits/10bf1605b5f6d52b5859bb4f160daf044a46e921 - as far as I can tell, the intention was to use system libraries if possible, but it probably should have contained extra logic for launching QGroundControl from AppImage. Maybe we could create symlinks for the missing libraries during AppImage creation?

IeiuniumLux:
I'm also using Ubuntu 16.04.6 LTS and QGC 3.5.1 just failed with the same error when I tried to launch it.

[...]

PR #8219, opened 2020-01-14
Remove copying libicu in buildtime

tcanabrava:
This seems quite wrong, on build time we should not copy
libraries, a target for `make package` shouldbe used instead

on cmake this is easier to handle as we have a INSTALL command
that runs only on make install

The problem with the current approach is that the comand fails
after the build is successfully finished, making any IDE to
think that the software did not compiled correctly

### DIFF
--- a/QGCSetup.pri
+++ b/QGCSetup.pri
@@ -120,13 +120,21 @@ LinuxBuild {
         libicui18n.so* \
         libQt5TextToSpeech.so.5
 
-    !contains(DEFINES, __rasp_pi2__) {
-        # Some Qt distributions link with *.so.56
-        QT_LIB_LIST += \
-            libicudata.so.56 \
-            libicui18n.so.56 \
-            libicuuc.so.56
-    }
+# TODO: Discover what to do here, checking for a specific
+# version of libicu is broken and gives compilation errors
+# when the app actually successfully build.
+#
+# for now, I'm just commenting those lines as it does not appear
+# to matter.
+# cp: cannot stat '/usr/lib/libicudata.so.56': No such file or directory
+#
+#    !contains(DEFINES, __rasp_pi2__) {
+#        # Some Qt distributions link with *.so.56
+#        QT_LIB_LIST += \
+#            libicudata.so.56 \
+#            libicui18n.so.56 \
+#            libicuuc.so.56
+#    }
 
     for(QT_LIB, QT_LIB_LIST) {
         QMAKE_POST_LINK += && $$QMAKE_COPY --dereference $$[QT_INSTALL_LIBS]/$$QT_LIB $$DESTDIR/Qt/libs/


DonLakeFlyer:
> This seems quite wrong

It might seem quite wrong, but it is working for some reason? It also is not failing in CI or in any of the Ubuntu builds I do.

tcanabrava:
The Linux world is not just Ubuntu; I’ve pasted the error message below and
it’s easy to see the reason for the error. You copy a specific version of
the library without checking if it exist.

DonLakeFlyer:
> The Linux world is not just Ubuntu

Ok. But you just commented out build rules which must have been there for a reason! Did you go back and look to see why that was added and by whom and what problem it fixed? And did you verify that your change doesn't break the fix that this was targeted to?

tcanabrava:
This is not a build rule, it’s a package rule of sorts. If this is inside a
“make package” target I wouldn’t try to fix as I would never use it. For a
“make” target (that should just build the binaries) this is indeed wrong.

I can add a target “make package” of sorts in qmake and move this copying
to it if you feel that’s more appropriate.

DonLakeFlyer:
> (that should just build the binaries)

That is not what the build does. It also sets things up for creating a release every time you build. Such that you can create a tarball for the run scripts or however the app image creation stuff works. You remove this and you break release creation. It's been like this since the dawn of time.

tcanabrava:
don, why do you feel so strongly against creating a “make deploy” or “make
release”?

In an old review of mine you stated that you didn’t like a change I
proposed as it would be against common code and for you having a common
base code it’s better as it’s easier for all developers to know what to
expect.

The common thing for any project with auto tools, cmake, qmake, is to use
the install target “make install”. Yes this would break the release
scripts, but it would break the release scripts only once.

You don’t work 100% of the time in Linux, I do. And I have this error 100%
of the time that I issue a make; let me be clear: I will not break
anything, it’s already broken.

jaxxzer:
I have had this error for quite some time on linux:

**cp: cannot stat '/usr/lib/libicudata.so.56': No such file or directory**

```
$ make
linking release/QGroundControl
echo Copying files && cp -f -R /home/jacob/git/qgroundcontrol/resources/flightgear /home/jacob/git/qgroundcontrol/build/release && mkdir -p /home/jacob/git/qgroundcontrol/build/release/Qt/libs && mkdir -p /home/jacob/git/qgroundcontrol/build/release/Qt/plugins && cp -f --dereference /usr/lib/libQt5Charts.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Core.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5DBus.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Gui.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Location.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Multimedia.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5MultimediaQuick.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Network.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5OpenGL.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Positioning.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5PositioningQuick.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5PrintSupport.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Qml.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Quick.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5QuickControls2.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5QuickTemplates2.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5QuickWidgets.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5SerialPort.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Sql.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Svg.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Test.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Widgets.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5XcbQpa.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5Xml.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libicui18n.so* /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libQt5TextToSpeech.so.5 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libicudata.so.56 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libicui18n.so.56 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference /usr/lib/libicuuc.so.56 /home/jacob/git/qgroundcontrol/build/release/Qt/libs/ && cp -f --dereference --recursive /usr/lib/qt/plugins/bearer /home/jacob/git/qgroundcontrol/build/release/Qt/plugins/ && cp -f --dereference --recursive /usr/lib/qt/plugins/geoservices /home/jacob/git/qgroundcontrol/build/release/Qt/plugins/ && cp -f --dereference --recursive /usr/lib/qt/plugins/iconengines /home/jacob/git/qgroundcontrol/build/release/Qt/plugins/ && cp -f --dereference --recursive /usr/lib/qt/plugins/imageformats /home/jacob/git/qgroundcontrol/build/release/Qt/plugins/ && cp -f --dereference --recursive /usr/lib/qt/plugins/platforminputcontexts /home/jacob/git/qgroundcontrol/build/release/Qt/plugins/ && cp -f --dereference --recursive /usr/lib/qt/plugins/platforms /home/jacob/git/qgroundcontrol/build/release/Qt/plugins/ && cp -f --dereference --recursive /usr/lib/qt/plugins/position /home/jacob/git/qgroundcontrol/build/release/Qt/plugins/ && cp -f --dereference --recursive /usr/lib/qt/plugins/sqldrivers /home/jacob/git/qgroundcontrol/build/release/Qt/plugins/ && cp -f --dereference --recursive /usr/lib/qt/plugins/texttospeech /home/jacob/git/qgroundcontrol/build/release/Qt/plugins/ && cp -f --dereference --recursive /usr/lib/qt/plugins/xcbglintegrations /home/jacob/git/qgroundcontrol/build/release/Qt/plugins/ && cp -f --dereference --recursive /usr/lib/qt/qml /home/jacob/git/qgroundcontrol/build/release/Qt/ && cp -f /home/jacob/git/qgroundcontrol/deploy/qgroundcontrol-start.sh /home/jacob/git/qgroundcontrol/build/release && cp -f /home/jacob/git/qgroundcontrol/deploy/qgroundcontrol.desktop /home/jacob/git/qgroundcontrol/build/release && cp -f /home/jacob/git/qgroundcontrol/resources/icons/qgroundcontrol.png /home/jacob/git/qgroundcontrol/build/release
Copying files
cp: cannot stat '/usr/lib/libicudata.so.56': No such file or directory
make: *** [Makefile:2475: release/QGroundControl] Error 1
```

DonLakeFlyer:
> don, why do you feel so strongly against creating a “make deploy” or “make release”?

I only feel strongly against breaking something which is working and I have to spend the time myself to figure out how to fix it. As I said above someone added this for a reason to fix something. Without any proof or reasearch to prove me wrong I assume removing it will break something in fully supported Ubuntu builds.

Also keep in mind that as far as I'm concerned there is no such thing as "make anything". There is only Qt Creator and CONFIG settings which affects the make. As far as why the build does what seems like setup/installer oriented things even when you haven't specified CONFIG+=installer. I don't know, it's the way it's always been as far as I can remember.

If someone wants to change the build system to clean all of that up I'm fine with it. But the first step is to understand why this specific copy is there and verify how to change it without breaking something. And then if the build system is going to change the phases it uses to skip the QGCSetup.pri steps and only do those when creating an installer then that also needs to be fully fully investigated across all OS as to whether it will cause problems. 

"_But the first step is to understand why this specific copy is there and verify how to change it without breaking something._" Also this first step needs to happen no matter what since ArchLinux or some other version of Linux which seems to get decent usage is broken. It would be nice for someone else to help with this, but it not I'll get around to it at some point. And by "breaking something" i mean the dropped Ubunutu builds which I consider to be the main supported version of QGC Linux.

DonLakeFlyer:
@tcanabrava The change is from the discussion here: https://github.com/mavlink/qgroundcontrol/issues/7382 which seems to have broken builds on Ubuntu 16.04. I'm not sure if we just drop support for 16.04 and go back to what it was before as long as the verifies as working on 18.04 or if there is some other way to keep all the linux versions as working happily.

tcanabrava:
16.04 is an LTS release that is supported till 2021, I'm ok with adding code on QGC to support it. please give me a few days to install 16.04 and create a patch that will fix things for it and for a newer linux.

DonLakeFlyer:
Thanks @tcanabrava. Also after this new stable goes out I'd be happy to talk about how to reorganize the build/installer phases. There is a lot of ancient history there I've never gone into much.

DonLakeFlyer:
@tcanabrava I'm going to cut Stable in one or two days. So if you have a fix for this it needs to go in now.

tcanabrava:
Give me till tomorrow and I’ll deliver.

DonLakeFlyer:
Awesome, thanks so much @tcanabrava

tcanabrava:
#8379

PR #8223, opened 2020-01-16
cmake: Add COMPILE_QML and DEBUG_QML options

patrickelectric:
This helps to improve the QML performance with COMPILE_QML and debug QML problems with DEBUG_QML

@dagar more cmake updates

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.10)
 
 project(QGroundControl LANGUAGES C CXX)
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
> DEBUG_QML

This should not be on by default. It causes random crashes in various OS when debugging. It's only usable for really specific stuff. I always have to turn it of in Qt Creator qmake builds.

dagar:
Awesome!

dagar:
FYI both options are off by default.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.12)
```
dagar: Any objection to leaving it at 3.10 for now? That's the cmake version that ships with the current Ubuntu LTS (18.04), and where a lot of developers/users live.

patrickelectric: Moved to 3.10

PR #8224, opened 2020-01-16
cmake: Add and use git module 

patrickelectric:
Move from hardcoded git binary to git module


DonLakeFlyer:
@dagar You should just merge these in when you are happy with them

PR #8233, opened 2020-01-17
Add a timer in between low battery warning messages.

### LINKED ISSUE #8232: Low battery warning announced too frequently (audio voice)
**Problem Description**
Every time the battery percentage changes and the battery percentage is below the battery low voltage threshold, QGC will announce a low battery warning as well as the current percentage.
On most of the existing battery setups the battery percentage can change quite quickly and this will cause QGC to spit out low battery messages very frequently.
Not only can this be very annoying but it also does not help if the last sentence is not finished before the new one starts, essentially leading to the user not even hearing what the current battery percentage is anymore.

**Suggestions**
I see a couple of ways how to improve this, possible even combinations of those

- require minimum time between announcing warnings, e.g. 20 or 30 seconds
- only report at steps of e.g. 5%
- do not announce if battery percentage goes up

dogmaphobic:
QGC will now wait 30 seconds before issuing a new (spoken) low battery warning. When the low battery reaches 50% of the threshold, the delays go down to every 15 seconds.

The original code had the announcement go out for every percentage change, which could turn into a stream of spoken messages on top of each other.

Closes #8232

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -300,6 +300,7 @@ Vehicle::Vehicle(LinkInterface*             link,
     // Start csv logger
     connect(&_csvLogTimer, &QTimer::timeout, this, &Vehicle::_writeCsvLine);
     _csvLogTimer.start(1000);
+    _lastBatteryAnnouncement.start();
 }
 
 // Disconnected Vehicle for offline editing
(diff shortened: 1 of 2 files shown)


dogmaphobic:
@DonLakeFlyer, something is messed up with the Android build:
```
Set executable permissions for: /home/travis/.gradle/wrapper/dists/gradle-2.2.1-all/6dibv5rcnnqlfbq9klf8imrndn/gradle-2.2.1/bin/gradle
```
A new gradle version causing issues? I have no idea how to fix that.

DonLakeFlyer:
That was kind of on purpose in case your battery goes to shit fast. But maybe your toast then anyway so it doesn’t matter.

PR #8234, opened 2020-01-17
Let the sender control the joystick message rate.

dogmaphobic:
The joystick messages (set_attitude_target, set_position_target_local_ned, etc.) were being throttled by the `UAS::setExternalControlSetpoint()` function when the sending functions are also controlling the frequency rate. That was causing issues where the message rate was all over the place. These message should go out on a constant rate, regardless if things changed or not.

### DIFF
--- a/src/uas/UAS.cc
+++ b/src/uas/UAS.cc
@@ -59,10 +59,6 @@ UAS::UAS(MAVLinkProtocol* protocol, Vehicle* vehicle, FirmwarePluginManager * fi
     controlPitchManual(true),
     controlYawManual(true),
     controlThrustManual(true),
-    manualRollAngle(0),
-    manualPitchAngle(0),
-    manualYawAngle(0),
-    manualThrust(0),
 
 #ifndef __mobile__
     fileManager(this, vehicle),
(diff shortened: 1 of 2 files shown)


MaEtUgR:
I got a binary from @MatejFranceskin and tested. It gives me still kind of jittery between 11 and 50Hz joystick messages but at least the average is consistently just below 25Hz and if I lower the update rate in the advanced tab to single digits I get very consistently just a tick ~0.5% below the expected rate.

I blamed where that logic was originally added: https://github.com/mavlink/qgroundcontrol/commit/bbeb00e305b2e9f797ed5571a96a9ebe504efb84

MaEtUgR:
Thanks a lot @dogmaphobic !!

dogmaphobic:
> still kind of jittery between 11 and 50Hz joystick messages

That should not happen. The thread runs at a consistent interval and messages are sent within the given rate. Are you measuring this on the sending side or on the receiving side? When I measure the sending side, I see a constant stream. If you are measuring on the receiving end, could it be lost messages?

MaEtUgR:
> Are you measuring this on the sending side or on the receiving side?

Receiving side, PX4 SITL binary running on the same machine printing out all the timings as the messages come in. I hope there are no messages lost in this setup but in any case, I also got consistent results with lower rates...

PR #8243, opened 2020-01-20
Vehicle: Change date format in csv output

Williangalvani:
This sets the datetime format to a format easier for the spreadsheet software to interpret.
Before this patch, it prints "tue sep 3 11:32:36 2019"


jaxxzer:
LGTM

PR #8251, opened 2020-01-22
VideoPageWidget: Enable camera control if using gstreamer

patrickelectric:
Since gstreamer is enable, we can record and do mostly everything


patrickelectric:
ping @dogmaphobic

dogmaphobic:
Sorry, it's been a bit crazy these past few days. The change makes sense within the context of VideoPageWidget (the _Video_ part).

PR #8261, opened 2020-01-28
Qmlsink simple version

tcanabrava:


dogmaphobic:
It builds but fails at runtime (macOS). What GStreamer version contains `org.freedesktop.gstreamer.GLVideoItem`?

```
(QGroundControl:2786): GStreamer-CRITICAL **: gst_object_unref: assertion 'object != NULL' failed
MAVLinkLogManagerLog: MAVLink logs directory: "/Users/gus/Documents/QGroundControl/Logs"
System reported locale: QLocale(English, Latin, United States) "en_US"
Loading localization for "en_US"
Map Cache in: "/Users/gus/Library/Caches/QGCMapCache300" / "qgcMapCache.db"
qrc:/qml/MainRootWindow.qml:329 Type FlightDisplayView unavailable
qrc:/qml/QGroundControl/FlightDisplay/FlightDisplayView.qml:343 Type FlightDisplayViewVideo unavailable
qrc:/qml/QGroundControl/FlightDisplay/FlightDisplayViewVideo.qml:24 module "org.freedesktop.gstreamer.GLVideoItem" is not installed
```

dogmaphobic:
Android is not finding GStreamer for some reason:
```
Project MESSAGE: Skipping support for video streaming (GStreamer libraries not installed)
```
The build is failing as it is not properly excluding GStreamer if not enabled:
```
compiling /home/travis/build/mavlink/qgroundcontrol/src/VideoStreaming/SubtitleWriter.cc
1664/home/travis/build/mavlink/qgroundcontrol/src/VideoStreaming/VideoReceiver.cc:725:5: error: unknown type name 'GstSample'
1665    GstSample *videobuffer = nullptr;
1666    ^
```

tcanabrava:
> It builds but fails at runtime (macOS). What GStreamer version contains `org.freedesktop.gstreamer.GLVideoItem`?
> 
> ```
> qrc:/qml/QGroundControl/FlightDisplay/FlightDisplayViewVideo.qml:24 module "org.freedesktop.gstreamer.GLVideoItem" is not installed
> ```
If the internet is correct, it's on gstreamer-good-plugins, and the first version of it was in gstreamer 1.6

https://gstreamer.freedesktop.org/documentation/qmlgl/qmlglsink.html
https://gstreamer.freedesktop.org/releases/1.6/

dogmaphobic:
This time I built using 1.16.2:
```
$ gst-launch-1.0 --version
gst-launch-1.0 version 1.16.2
GStreamer 1.16.2
```
Same runtime error:
```
(QGroundControl:6409): GStreamer-CRITICAL **: 12:05:21.457: gst_object_unref: assertion 'object != NULL' failed
qrc:/qml/MainRootWindow.qml:329 Type FlightDisplayView unavailable
qrc:/qml/QGroundControl/FlightDisplay/FlightDisplayView.qml:343 Type FlightDisplayViewVideo unavailable
qrc:/qml/QGroundControl/FlightDisplay/FlightDisplayViewVideo.qml:24 module "org.freedesktop.gstreamer.GLVideoItem" is not installed
```

I don't think GLVideoItem is available on macOS. That reminded me of an old discussion we had about this:
```
$ gst-inspect-1.0 qmlglsink
No such element or plugin 'qmlglsink'
```
https://gitter.im/mavlink/qgroundcontrol?at=5d6524df68406739f9904d6a

tcanabrava:
gus, this is unrelated to GStreamer, it's a GStreamer plugin on the -good package.

run: 

> ➜  ~ gst-inspect-1.0 | grep qml
it should tell you:

```
qmlgl:  qmlglsink: Qt Video Sink
qmlgl:  qmlglsrc: Qt Video Source

```

if you don't have this answer from the gst-inspect, then you need to install gst-plugins-good (probably brew has it)

dogmaphobic:
>then you need to install gst-plugins-good (probably brew has it)

That's not how this works on macOS. You download a single package, which has everything already in it. You don't install it piecemeal. We do not use homebrew for GStreamer. That would require users to install and maintain GStreamer on their own. Only Linux works like that.

jaxxzer:
Here is a link back to the relevant conversation on gitter with @patrickelectric https://gitter.im/mavlink/qgroundcontrol?at=5d67154a07d1ff39f89f2742

tcanabrava:
> > then you need to install gst-plugins-good (probably brew has it)
> 
> That's not how this works on macOS. You download a single package, which has everything already in it. You don't install it piecemeal. We do not use homebrew for GStreamer. That would require users to install and maintain GStreamer on their own. Only Linux works like that.

Gus, I don't know what to say... besides all of the developers that I know in mac os use brew to install **developer tools**, and if you are **building** a software from sources you are either a developer or someone that knows how to handle the command line. 

I can assure you that qmlglsink works if installed by brew using the gst-plugins-good meta package, I don't understand the issue with it (as most developer tools for mac are installed with brew).

The other part of your comment is **users** should not care about brew, this part I agree. and they don't need, following the gstreamer documentation those should be inside the app bundle.

https://gstreamer.freedesktop.org/documentation/deploying/mac-osx.html?gi-language=c
https://stackoverflow.com/questions/32807973/installing-gstreamer-1-0-on-mac-os-x-mavericks
https://stackoverflow.com/questions/30702842/cant-find-get-launch-1-0-from-gstreamer/30873313#30873313

dogmaphobic:
If you add a dependency to QGC based on software installed by Homebrew, invariably you are forcing users to also install Homebrew (and the dependent packages) in order for the runtime to work. Yes, it is possible to circumvent that by writing code that during the build process collects these dependencies, adjust the search path, bundle them into the distributable package and document the process so others trying to build QGC will know what they need to do.

The documentation must also explain what specific version of GStreamer should be installed. The reason we use pre-packaged versions of every dependency is because there is no way to support arbitrary versions of things. Things should "just work" and the only way to make sure that happen is to make sure all the parts used are known to work (Linux is the exception based on the assumption you know what you are doing if you are using it.)

If that's the case, this will need to be added and tested along with the PR.

I have no particular issues with the source of the dependency (within reason) provided that the dependency only exists during the building process. We cannot require users to have to install them.

tcanabrava:
I don’t have a problem adding the documentation for that based on brew,
however I don’t have a Mac - so I’m using the build failures on Jenkins to
fix whatever it tells me that it’s broken.

And I’ll need a Mac to create the scripts to port.

About the versions, this is also another thing that cmake could help with,
as you can do “find_package(Something VERSION 1.0 EXACT)” and it fails
during configure time.

I’ll take the time to update the documentation to build based on this patch
(also, note that same issues will happen with the other qmlsink patch)

PR #8262, opened 2020-01-28
Bring back non-coordinate display and waypoint sequence number

stefandunca:
The `Jump to` mission waypoint doesn't make sense without knowing the item numbers, therefore, bring back waypoint numbers and display of non-position waypoints like "Jump to". Hiding the non-position waypoints hides important mission actions.

Removed by the following PRs:
 - Remove non-coordinate items: #7771
 - Hide sequence numbers: #7830

This PR:
Screenshot from 2020-01-28 21-09-03

Current master:
Screenshot from 2020-01-28 21-10-08

### DIFF
--- a/src/FlightMap/MapItems/MissionItemIndicator.qml
+++ b/src/FlightMap/MapItems/MissionItemIndicator.qml
@@ -32,6 +32,7 @@ MapQuickItem {
             id:                 _label
             checked:            _isCurrentItem
             label:              missionItem ? missionItem.abbreviation : ""
+            index:              missionItem ? missionItem.sequenceNumber : 0
             gimbalYaw:          missionItem.missionGimbalYaw
             vehicleYaw:         missionItem.missionVehicleYaw
             showGimbalYaw:      !isNaN(missionItem.missionGimbalYaw)
--- a/src/PlanView/SimpleItemMapVisual.qml
+++ b/src/PlanView/SimpleItemMapVisual.qml
@@ -110,6 +110,22 @@ Item {
             missionItem:    _missionItem
             sequenceNumber: _missionItem.sequenceNumber
             onClicked:      _root.clicked(_missionItem.sequenceNumber)
+            // These are the non-coordinate child mission items attached to this item
+            Row {
+                anchors.top:    parent.top
+                anchors.left:   parent.right
+                Repeater {
+                    model: _missionItem.childItems
+                    delegate: MissionItemIndexLabel {
+                        z:                      2
+                        label:                  object.abbreviation.length === 0 ? object.sequenceNumber : object.abbreviation.charAt(0)
+                        checked:                object.isCurrentItem
+                        child:                  true
+                        specifiesCoordinate:    false
+                        onClicked:              _root.clicked(object.sequenceNumber)
+                    }
+                }
+            }
         }
     }
 }
--- a/src/QmlControls/MissionItemIndexLabel.qml
+++ b/src/QmlControls/MissionItemIndexLabel.qml
@@ -25,7 +25,7 @@ Canvas {
     property real   gimbalYaw
     property real   vehicleYaw
     property bool   showGimbalYaw:          false
-    property bool   showSequenceNumbers:    false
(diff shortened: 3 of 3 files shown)


hamishwillee:
I know @Antiheavy will be happy with this. Might be cool to make this a setting, or numbers become visible when you start using things that need them - e.g. jumps.

DonLakeFlyer:
I pretty much 100% disagree with this. For the incredibly limited usage of Jump you can see what the waypoints numbers are from the hamburger menu. It's not worth adding useless noise to the ui just for Jump.

@antiheavy What are your complaints about this. News to me.

DonLakeFlyer:
There are many other possibilities to make Jump easier to use. A simple one is to update the help text for it to show the user how to get the item number they need.

DonLakeFlyer:
Also if you click the T to get the bottom mission item status display you'll see item numbers are still in there. And you can click on the items in there to select as well. So many different ways to workaround.

DonLakeFlyer:
Also my change here is meant to be significant. I want to hear from users why they think they can't live without numbering. Take that feedback and make the needed adjustments. My expectation is to break some things to make forward progress on simplification.

DonLakeFlyer:
Also creating a pull which backs out functionality which was changed for a reason is not the way to do things. Either start a discussion in an Issue with problems or a thread on QGC slack.

Antiheavy:
I don’t recall advocating for bringing the numbers back (although my memory could be wrong). Either way, I’m a fan of not showing mission item numbers by default and having other ways for the advanced user to view numbers.  There are much better ways to present the mission to the typical user than a list of item numbers.  If an advanced user wants to do advanced things like GoTo it almost makes more sense to have a detailed list in a table view that could be accessed through an advanced view button.

DonLakeFlyer:
Also just FYI: My plan for post Stable is to propose a mavlink spec for mission upload errors and feasilbility check failures which brings them into the 21st century. Similar to how you don't need to know code line numbers to see an error reported by an IDE. So you won't need to know misison item numbers for that either. And then the errors will just tie into the new "Red question mark" problem indicator thingy which has been introduced.

Antiheavy:
Actually,  thinking about it more, there are certainly use cases where waypoint numbers shown on the map can be useful. E.g when communicating with visual observers or coordinating with others involved with an operation.

 Maybe it should be an app setting option or a toggle option somewhere in the plan/fly screens?

DonLakeFlyer:
> Actually, thinking about it more, there are certainly use cases where waypoint numbers shown on the map can be useful. E.g when communicating with visual observers or coordinating with others involved with an operation.

Can you start an Issue and include this as a possible problem with no numbers. Then I can collect it all up over some time and figure out how to handle.

DonLakeFlyer:
Also the real fix for this is to stop using MAV_CMD_DO_JUMP and support MAV_CMD_DO_JUMP_TAG. That doesn't have all the usability problems associated with DO_JUMP:
* Doesn't require knowing about sequence numbers
* Doesn't have the problem of sequence number being unstable causing jumps to be incorrect. If you add/remove a mission item prior to the jump your screwed. Even and indirect change like survey transect spacing changing prior to a jump will screw things up.

DonLakeFlyer:
> Also the real fix for this is to stop using MAV_CMD_DO_JUMP and support MAV_CMD_DO_JUMP_TAG. That doesn't have all the usability problems associated with DO_JUMP:

This is still my first recommendation. If that won't happen and you really want to make DO_JUMP work for real then I think the following things need to happen:
* Setup do jump item edit like takeoff special case where the use first clicks another item visuallyt to set the jump location
* Change recalc code to handle additions/deletions and automatically adjust do jump locations as needed

All of that is a bunch of work which to me is wasted given the much better usage of DO_JUMP_TAG.

stefandunca:
@DonLakeFlyer this is my first encounter with the mission planning and I go by a lot of assumptions, I only saw broken features. This was an internal request and after some digging, I found those changes that carried no information at all and looked as removing features. Thank you for clarifying all this.

PR #8268, opened 2020-01-31
Register to QML in the constructor directly

tcanabrava:
We where registering in the SetToolbox method, but if that's called
twice, then we'd have a double registering. This simplifies the code
a bit.


tcanabrava:
> Calling `setToolbox()` twice would be an error. With that said, I don't particularly have any issues with moving it to the constructor.

it would. but there's currenly no way for us to be sure it's only happening once, perhaps a Q_ASSERT(!m_toolbox) to crash if we double assert it?

(I would try to get rid of this toolbox class in the future, it feels that it was a class for the widget version)

dogmaphobic:
@tcanabrava Please, update this and I will merge it.

PR #8270, opened 2020-01-31
The first bunch of Qt code improvements

alexeylysenko:
Fixed few performance issues in Qt code. For example, got rid of unnecessary memory allocations.

### DIFF
--- a/src/Camera/QGCCameraControl.cc
+++ b/src/Camera/QGCCameraControl.cc
@@ -835,7 +835,7 @@ QGCCameraControl::_loadCameraDefinitionFile(QByteArray& bytes)
         qCDebug(CameraControlLog) << "Saving camera definition file" << _cacheFile;
         QFile file(_cacheFile);
         if (!file.open(QIODevice::WriteOnly)) {
-            qWarning() << QString("Could not save cache file %1. Error: %2").arg(_cacheFile).arg(file.errorString());
+            qWarning() << QString("Could not save cache file %1. Error: %2").arg(_cacheFile, file.errorString());
         } else {
             file.write(originalData);
         }
@@ -1132,7 +1132,7 @@ QGCCameraControl::_handleLocalization(QByteArray& bytes)
         QDomNode locale = locales.item(i);
         QString name;
         read_attribute(locale, kName, name);
-        if(name.toLower().startsWith(localeName)) {
+        if(name.startsWith(localeName, Qt::CaseInsensitive)) {
             return _replaceLocaleStrings(locale, bytes);
         }
     }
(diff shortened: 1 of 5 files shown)


src/Camera/QGCCameraControl.cc
```diff
@@ -1171,12 +1171,14 @@ void
...
-            _paramIO[paramName]->setParamRequest();
+    auto iterator = _paramIO.constBegin();
+    while (iterator != _paramIO.constEnd()) {
+        if (iterator.value()) {
+            iterator.value()->setParamRequest();
         } else {
-            qCritical() << "QGCParamIO is NULL" << paramName;
+            qCritical() << "QGCParamIO is NULL" << iterator.key();
         }
+        ++iterator;
```
tcanabrava: Nak. even tougth iterator is faster than creating a array of keys, this version is worse to read and easier to create mistakes. sometimes it's more iportant readability than speed, and looking at what the code is trying to do, the old, and the new versions are wrong.

for (const auto param : qAsConst(_paramIO) {
 if (param) {
   param->setParamRequest();
 } 
}

if we are only using the name of the param to print on the qCritcal(), the param should have a ->name() method.

src/Camera/QGCCameraControl.cc
```diff
@@ -1386,26 +1388,34 @@ QGCCameraControl::_updateRanges(Fact* pFact)
...
+        const auto &value = resetIterator.value();
+        key->setEnumInfo(_originalOptNames[value], _originalOptValues[value]);
+        if(!updates.contains(key->name())) {
+            _paramIO[key->name()]->optNames = _originalOptNames[value];
+            _paramIO[key->name()]->optVariants = _originalOptValues[value];
+            emit key->enumsChanged();
+            qCDebug(CameraControlVerboseLog) << "Restore full set of options for:" << key->name() << _originalOptNames[key->name()];
+            updates << key->name();
         }
+        ++resetIterator;
```
tcanabrava: way way worse to read.  (and the old code was also bad to read).
Please don't just try to make things faster, but they should be faster and better in readability.
Perhaps this code should change to std::map / std::unordered_map as the iteration thru range-based-for has key and value?
to we *really* need key and value?
you see that you are double-searching values based on keys just like the old code?

src/FactSystem/ParameterManager.cc
```diff
@@ -356,7 +356,7 @@ void ParameterManager::_parameterUpdate(int vehicleId, int componentId, QString
             qCDebug(ParameterManagerVerbose1Log) << _logVehiclePrefix(-1) << "Not restarting _waitingParamTimeoutTimer (all requests satisfied)";
         }
     }
-\
+
```
tcanabrava: this part of the file I like. but you are missing qAsConst in all maps as range-based for will create a copy of the containers.

PR #8272, opened 2020-02-02
Code documentation and updates

dogmaphobic:
We had some quite primitive support for Doxygen code documentation. I've updated it along with adding a few high level diagrams. 

This PR changes quite a few files but none of the changes are code changes:

* Update copyright date throughout
* Updated code (Doxygen) documentation in several, main classes
* Added QML to Doxygen (it documents functions and properties)
* Remove a couple of no longer used files

Screen Shot 2020-02-02 at 4 47 00 PM


hamishwillee:
Sounds good. Do you plan to autogenerate and publish the output somewhere - or is this just a "you will now be able to generate this yourself" fix?

dogmaphobic:
> or is this just a "you will now be able to generate this yourself" fix?

More likely. Its audience is too narrow for broad distribution and if that's the case, the code would need to be really fully documented. As is, I would say less than 30% of it is properly documented. Maybe after this is better cleaned up.

hamishwillee:
Cool. Perhaps worth adding a note that this is supported and how to do a build in https://dev.qgroundcontrol.com/en/ though. I'll leave to you :-)

Issue #8249, opened 2020-01-22
Proposal: keyboard shortcut for extracting still frames from video stream


jaxxzer:
This functionality has been around in QGC, a bit behind the scenes. I propose using the spacebar to trigger a video frame saved to file as a picture. The keyboard key may be reassigned as an extra or subsequent feature.

PR #8274, opened 2020-02-03
Add screenshoot capabilities to the Pipeline

tcanabrava:
Plug a screenshoot button on the PIP that takes a screenshoot
of the currently running pipeline to showcase how the interaction
with the new Sink should be made

### DIFF
--- a/src/QmlControls/QGCPipable.qml
+++ b/src/QmlControls/QGCPipable.qml
@@ -198,7 +198,7 @@ Item {
         radius:                 ScreenTools.defaultFontPixelHeight / 3
         visible:                true
         color:                  isDark ? Qt.rgba(0,0,0,0.75) : Qt.rgba(0,0,0,0.5)
-
+        focus: visible
         Image {
             width:              parent.width  * 0.75
             height:             parent.height * 0.75
@@ -213,6 +213,11 @@ Item {
             anchors.fill: parent
             onClicked: {
                 print("Clicked!")
+                takeSnapshot();
+            }
+        }
+        Keys.onPressed: {
+            if (event.key === Qt.Key_Space) {
                 takeSnapshot();
             }
         }
--- a/src/QmlControls/QGCPipable.qml
+++ b/src/QmlControls/QGCPipable.qml
@@ -212,7 +212,6 @@ Item {
         MouseArea {
             anchors.fill: parent
             onClicked: {
-                print("Clicked!")
                 takeSnapshot();
             }
         }
--- a/src/QmlControls/QGCPipable.qml
+++ b/src/QmlControls/QGCPipable.qml
@@ -189,10 +189,9 @@ Item {
             }
         }
     }
-
     Rectangle {
         id:                     buttonScreenShoot
-        anchors.left:           buttonStop.right
+        anchors.left:           openPIP.right
         anchors.bottom:         parent.bottom
         height:                 ScreenTools.defaultFontPixelHeight * 2
         width:                  ScreenTools.defaultFontPixelHeight * 2
@@ -213,6 +212,7 @@ Item {
         MouseArea {
             anchors.fill: parent
             onClicked: {
+                print("Clicked!")
                 takeSnapshot();
             }
         }
--- a/src/QmlControls/QGCPipable.qml
+++ b/src/QmlControls/QGCPipable.qml
@@ -33,6 +33,7 @@ Item {
     signal  hideIt(bool state)
     signal  newWidth(real newWidth)
     signal  popup()
+    signal takeSnapshot()
 
     MouseArea {
         id: pipMouseArea
@@ -185,6 +186,34 @@ Item {
             anchors.fill: parent
             onClicked: {
                 pip.hideIt(false)
+            }
+        }
+    }
+
+    Rectangle {
+        id:                     buttonScreenShoot
+        anchors.left:           buttonStop.right
+        anchors.bottom:         parent.bottom
+        height:                 ScreenTools.defaultFontPixelHeight * 2
+        width:                  ScreenTools.defaultFontPixelHeight * 2
+        radius:                 ScreenTools.defaultFontPixelHeight / 3
+        visible:                true
+        color:                  isDark ? Qt.rgba(0,0,0,0.75) : Qt.rgba(0,0,0,0.5)
+
+        Image {
+            width:              parent.width  * 0.75
+            height:             parent.height * 0.75
+            sourceSize.height:  height
+            source:             "/res/buttonRight.svg"
+            mipmap:             true
+            fillMode:           Image.PreserveAspectFit
+            anchors.verticalCenter:     parent.verticalCenter
+            anchors.horizontalCenter:   parent.horizontalCenter
+        }
+        MouseArea {
+            anchors.fill: parent
+            onClicked: {
+                takeSnapshot();
             }
         }
     }
--- a/src/QmlControls/QGCPipable.qml
+++ b/src/QmlControls/QGCPipable.qml
@@ -191,19 +191,19 @@ Item {
     }
     Rectangle {
         id:                     buttonScreenShoot
-        anchors.left:           openPIP.right
-        anchors.bottom:         parent.bottom
+        anchors.bottom: parent.bottom
+        anchors.horizontalCenter: parent.horizontalCenter
         height:                 ScreenTools.defaultFontPixelHeight * 2
         width:                  ScreenTools.defaultFontPixelHeight * 2
         radius:                 ScreenTools.defaultFontPixelHeight / 3
-        visible:                true
+        visible:                !isHidden && (ScreenTools.isMobile || pipMouseArea.containsMouse)
         color:                  isDark ? Qt.rgba(0,0,0,0.75) : Qt.rgba(0,0,0,0.5)
         focus: visible
         Image {
             width:              parent.width  * 0.75
             height:             parent.height * 0.75
             sourceSize.height:  height
-            source:             "/res/buttonRight.svg"
+            source:             "/qmlimages/camera_photo.svg"
             mipmap:             true
             fillMode:           Image.PreserveAspectFit
             anchors.verticalCenter:     parent.verticalCenter
--- a/src/Camera/QGCCameraControl.cc
+++ b/src/Camera/QGCCameraControl.cc
@@ -388,12 +388,11 @@ QGCCameraControl::takePhoto()
             _setPhotoStatus(PHOTO_CAPTURE_IN_PROGRESS);
             _captureInfoRetries = 0;
             //-- Capture local image as well
-            if(qgcApp()->toolbox()->videoManager()) {
-                QString photoPath = qgcApp()->toolbox()->settingsManager()->appSettings()->savePath()->rawValue().toString() + QStringLiteral("/Photo");
-                QDir().mkpath(photoPath);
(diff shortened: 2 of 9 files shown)


tcanabrava:
sigh - I need to block this from building on non gstreamer.

src/QmlControls/QGCPipable.qml
```diff
@@ -188,4 +189,37 @@ Item {
...
     }
+    Rectangle {
+        id:                     buttonScreenShoot
+        anchors.left:           openPIP.right
+        anchors.bottom:         parent.bottom
+        height:                 ScreenTools.defaultFontPixelHeight * 2
+        width:                  ScreenTools.defaultFontPixelHeight * 2
+        radius:                 ScreenTools.defaultFontPixelHeight / 3
+        visible:                true
+        color:                  isDark ? Qt.rgba(0,0,0,0.75) : Qt.rgba(0,0,0,0.5)
```
patrickelectric:
```
color: "black"
opacity: isDark ? 0.75 : 0.5
```

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -966,6 +968,61 @@ VideoReceiver::_keyframeWatch(GstPad* pad, GstPadProbeInfo* info, gpointer user_
...
+void VideoReceiver::takeSnapshot() {
+#if defined(QGC_GST_STREAMING)
+    qDebug() << "Taking Snapshot";
+    if (!_videoSink) {
+        return;
+    }
+
+    qDebug() << "Trying to take a snapshot";
+    GstSample *videobuffer = nullptr;
+    GstCaps *caps;
```
patrickelectric: Missing nullptr ?

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -966,6 +968,61 @@ VideoReceiver::_keyframeWatch(GstPad* pad, GstPadProbeInfo* info, gpointer user_
...
+    gint width, height;
+    GstMapInfo map;
+
+    g_object_get(G_OBJECT(_videoSink), "last-sample", &videobuffer, NULL);
+    if (!videobuffer) {
+        qDebug() << "No video buffer, exiting";
```
patrickelectric: This file contains a debug category, maybe you should use it.

src/QmlControls/QGCPipable.qml
```diff
@@ -188,4 +189,37 @@ Item {
...
+        focus: visible
```
jaxxzer: this seems like it could interfere with other items that might want focus, is it right?

src/QmlControls/QGCPipable.qml
```diff
@@ -188,4 +189,37 @@ Item {
...
+            source:             "/res/buttonRight.svg"
+            mipmap:             true
+            fillMode:           Image.PreserveAspectFit
+            anchors.verticalCenter:     parent.verticalCenter
+            anchors.horizontalCenter:   parent.horizontalCenter
+        }
+        MouseArea {
+            anchors.fill: parent
+            onClicked: {
+                print("Clicked!")
```
jaxxzer: stale debug

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -966,6 +968,61 @@ VideoReceiver::_keyframeWatch(GstPad* pad, GstPadProbeInfo* info, gpointer user_
...
+        qDebug() << "Cant get width or height from caps";
+        return;
+    }
+
+    GstBuffer *snapbuffer = gst_sample_get_buffer(videobuffer);
+    gst_buffer_map (snapbuffer, &map, GST_MAP_READ);
+
+    uchar* bufferData = reinterpret_cast<uchar*>(map.data);
+    QImage::Format imageFormat = QImage::Format_RGB32;
+    // QImage::InvertMode invertMode = QImage::InvertRgb;
```
jaxxzer: stale? leave an explanation if commented code should stay around

jaxxzer:
This was made to close #8249 

Some feedback about the functionality I experienced with https://github.com/mavlink/qgroundcontrol/pull/8274/commits/c314d95ba0045aa6d45e9724b152b7e571d1ce0f:
- the spacebar makes a png from the video stream - great!
- the video stream pauses for a brief moment ~300ms when a picture is saved, this is ok to me for now, but something we should try to improve
- we should save these to a dedicated 'pictures' directory instead of putting them in the 'video' directory
- I got a segmentation fault on taking my 6th picture. This 6th picture is a fraction of the memory size of the other pictures:
```
Taking Snapshot
Trying to take a snapshot
Image saved in "/home/jacob/Documents/QGroundControl/Video/2020-02-05_10.32.01.png"
Taking Snapshot
Trying to take a snapshot
Image saved in "/home/jacob/Documents/QGroundControl/Video/2020-02-05_10.32.08.png"
Taking Snapshot
Trying to take a snapshot
Image saved in "/home/jacob/Documents/QGroundControl/Video/2020-02-05_10.32.11.png"
Taking Snapshot
Trying to take a snapshot
Image saved in "/home/jacob/Documents/QGroundControl/Video/2020-02-05_10.32.18.png"
Taking Snapshot
Trying to take a snapshot
Image saved in "/home/jacob/Documents/QGroundControl/Video/2020-02-05_10.32.22.png"
Taking Snapshot
Trying to take a snapshot
Segmentation fault (core dumped)
```
image

- the pictures that did save correctly are funny:


- the qml is a bit funny:

1. it does not look anchored right (but some normal menu items are bad too)
1. it will go over the map
image

patrickelectric:
> * it does not look anchored right (but some normal menu items are bad too)
> * it will go over the map

From your window manager, it appears that you are using plasma.
There is a bug in plasma style manager, start QGC with universal style.
```
QT_QUICK_CONTROLS_STYLE=universal qgroundcontrol
```

src/VideoStreaming/VideoReceiver.cc
andrewvoznytsa: It will not work like that in case of qmlglsink packaged into GstBin (see https://github.com/mavlink/qgroundcontrol/blob/master/src/FlightDisplay/VideoManager.cc#L287-L379). To make that working you need to add some property to GstBin (let's call is also "last-sample") and forward call to qmlglsink inside that bin.

andrewvoznytsa: > This file contains a debug category, maybe you should use it.

I'm planing to cleanup logging in whole VideoReceiver - as for me @tcanabrava can leave that as is, I'll take care of that.

src/VideoStreaming/VideoReceiver.cc
andrewvoznytsa: IMO this code should check buffer format and if it is not RGB32 (RGBA as qmlglsink advertises) then simply fail - potential crash place (low probability, but...).

src/VideoStreaming/VideoReceiver.cc
andrewvoznytsa: And one more thing - you need to analyze GstVideoInfo (you can get it from sink caps and sample's pool caps) to get raw pointer to the video buffer. Sometimes pixel 0, 0 is not aligned with begging of memory block (sample buffer as try to do) and depending on hardware and GL implementation it is not necessary that image row length will be equal to sizeof(pixel) * width - sometimes it is longer because of buffer aligning reasons. You need to use thing called pitch/stride to find beginning of next pixel row.

IMO the best possible implementation in your case (where performance does not really matter but maintenance and stability is the key) - it is to find some gstreamer tool that converts GstSample into some image evaluating GstVideoInfo, as you need. It is not big fun to deal with implementation-depended buffer aligning. I suspect that in your case it will be 99.9% aligned as you expect but can't guarantee it for all cases.

jaxxzer:
> There is a bug in plasma style manager, start QGC with universal style.

That improves the situation:
image

tcanabrava:
That button is the photo button, I used a icon that was already in the
system.

jaxxzer:
> That button is the photo button, I used a icon that was already in the
system.

My comment (or red arrow) is referring to the positioning/anchoring of the icon.

src/QmlControls/QGCPipable.qml
tcanabrava: done.

src/VideoStreaming/VideoReceiver.cc
tcanabrava: done.

src/QmlControls/QGCPipable.qml
tcanabrava: yes, it's right. we need focus *or* a customized focus handler, to trigger actions with the keyboard.

src/QmlControls/QGCPipable.qml
tcanabrava: done.

src/VideoStreaming/VideoReceiver.cc
tcanabrava: not stale, the picture is being saved with the wrong colors and I'm experimenting.

tcanabrava:
> > That button is the photo button, I used a icon that was already in the
> > system.
> 
> That is understood, my comment (or red arrow) is referring to the positioning/anchoring of the icon.

ah, ok. The position is weird because it's anchoring to the right of another icon that is currently hidden because the mouse is not hovering the video element. I'll change the position to center.

src/VideoStreaming/VideoReceiver.cc
andrewvoznytsa: There are should be some kind 'inverted' RGB32 image format - most likely problem is related to color component order - what you have might be ABGR. Try to look for some BGR32 or InvertedRGB32 - something like.

src/QmlControls/QGCPipable.qml
tcanabrava: not really done, I'll update this in a bit.

jaxxzer:
@tcanabrava most everything from my last review is better. No more crashes, the photos appear correct, the anchoring is better. There are only two things now:

- We should make a 'photos' directory for QGC photos instead of putting them in 'video' directory
- The space bar only works when the mouse hovers over this area: 
image

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -1009,52 +1009,52 @@ VideoReceiver::_keyframeWatch(GstPad* pad, GstPadProbeInfo* info, gpointer user_
...
-    qDebug() << "Trying to take a snapshot";
+     qCDebug(VideoReceiverLog) << "Trying to take a snapshot";
     GstSample *videobuffer = nullptr;
-    GstCaps *caps;
+    GstCaps *caps = nullptr;
     gint width, height;
     GstMapInfo map;
 
     // The video sink now is a GstBin that *contains* a sink.
     GstElement *realSink = gst_bin_get_by_name(GST_BIN(_videoSink), "qmlsink");
```
andrewvoznytsa: What you did will work. But it expects that _videosink is GstBin (not necessary, if we support customization - it could be just GstElement) and it expects that GstBin contains element called "qmlsink" (it was called "qmlglsink" in my branch, IIRC) which is also not necessary to be true.

Proper way would be to install "last-sample" property at GstBin level (probably around this line https://github.com/mavlink/qgroundcontrol/blob/master/src/FlightDisplay/VideoManager.cc#L351) and then simply query last-sample as you did in your previous version.

DonLakeFlyer:
Closing as abandoned

PR #8277, opened 2020-02-04
Checklist enforcing

dogmaphobic:
I've added an option to enforce the use (and passing) the preflight checklist. The option is under general settings, though it is only visible if this is implemented by a custom build (regular QGC does not have preflight checklist enabled).

If enabled, you won't be able to arm, take off, start a mission etc. until all checklist tests are passed.


Antiheavy:
Good feature!

PR #8278, opened 2020-02-05
Remove unnecessary pointer check before delete 

### LINKED ISSUE #8247: Remove unnecessary null pointer checks
[An extra null pointer check is not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like the following.
- [AirspaceManager](https://github.com/mavlink/qgroundcontrol/blob/1bd2ba525547c9c1ae7b68368a016fcc9bd61fb5/src/AirspaceManagement/AirspaceManager.cc#L49 "Destructor for AirspaceManager")
- [GPSManager::disconnectGPS](https://github.com/mavlink/qgroundcontrol/blob/e7e3818dd178bd5b3652eb3a304c1011387dc189/src/GPS/GPSManager.cc#L84 "GPSManager::disconnectGPS function")
- [QGCCorePlugin_p](https://github.com/mavlink/qgroundcontrol/blob/48f4ae193ae8728de35390cb71aa3068910e5a57/src/api/QGCCorePlugin.cc#L37 "Destructor for QGCCorePlugin_p")

patrickelectric:
Fix #8247 
> The C++ language guarantees that delete p will do nothing if p is null. Since you might get the test backwards, and since most testing methodologies force you to explicitly test every branch point, you should not put in the redundant if test.

### DIFF
--- a/src/AirspaceManagement/AirspaceManager.cc
+++ b/src/AirspaceManagement/AirspaceManager.cc
@@ -48,21 +48,16 @@ AirspaceManager::AirspaceManager(QGCApplication* app, QGCToolbox* toolbox)
 //-----------------------------------------------------------------------------
 AirspaceManager::~AirspaceManager()
 {
-    if(_advisories) {
-        delete _advisories;
-    }
-    if(_weatherProvider) {
-        delete _weatherProvider;
-    }
-    if(_ruleSetsProvider) {
-        delete _ruleSetsProvider;
-    }
-    if(_airspaces) {
-        delete _airspaces;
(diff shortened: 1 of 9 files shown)


elfring:
[Further update candidates were left over](https://github.com/mavlink/qgroundcontrol/blob/2e033034dc3944d329bb9fc3a52018d34d527d8e/src/GPS/GPSManager.cc#L84 "Update candidate: GPSManager::disconnectGPS()").

PR #8286, opened 2020-02-07
Bringing preflight checklist to mainline QGC

dogmaphobic:
It's back and it's no longer generic. There is a model for each vehicle type:

* Fixed Wing
* Multirotor
* Rover
* Submarine
* VTOL

If these checks fail, it will load a generic preflight checklist.

Note that I've only trimmed the "Generic" check list when creating the individual models. Someone with actual experience using those should come and properly fill in the required/desired check list items.

Screen Shot 2020-02-07 at 1 48 28 PM

A custom build can override the (QGCCorePlugin) method collecting the URI to the checklist (the main loader) and completely customize it if they want.

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -19,6 +19,14 @@
         <file alias="TelemetryRSSIIndicator.qml">src/ui/toolbar/TelemetryRSSIIndicator.qml</file>
         <file alias="VTOLModeIndicator.qml">src/ui/toolbar/VTOLModeIndicator.qml</file>
     </qresource>
+    <qresource prefix="/checklists">
+        <file alias="DefaultChecklist.qml">src/FlightDisplay/DefaultChecklist.qml</file>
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
> Someone with actual experience using those should come and properly fill in the required/desired check list items.

That's kind of the main problem. I'm not sure there is something that can be made up which makes sense to everyones usage pattern.

dogmaphobic:
The basics are the same and it's stuff that you should pay attention each time. This is true for "real" aviation as well.

DonLakeFlyer:
If it's good enough should it default to on? Maybe we try that after this stable goes out. But default the thing you added to prevent arming to off.

Antiheavy:
I’m willing to help with text for a “generic” fixed wing checklist. Let me know if/when this would be helpful.

dogmaphobic:
Anytime. The code is merged in. You just need to add/remove whatever you think is pertinent here:

https://github.com/mavlink/qgroundcontrol/blob/master/src/FlightDisplay/FixedWingChecklist.qml

PR #8297, opened 2020-02-09
use mavlink fork of gst-plugins-good

dagar:
I'm having issues cloning https://gitlab.freedesktop.org/andrew.voznytsa/gst-plugins-good. There seems to be a github mirror (https://github.com/GStreamer/gst-plugins-good), but it's missing the commit qgc is currently pointing at (9d782fad9dc). https://github.com/GStreamer/gst-plugins-good/compare/9d782fad9dc0384ba86ecae64511c193f6149f93...c00796eaa5fb5fde271fd1081f5fab57b970e428

When it's possible to clone https://gitlab.freedesktop.org/andrew.voznytsa/gst-plugins-good again let's get it pushed to a branch on a fork we can control (https://github.com/mavlink/gst-plugins-good).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,4 +9,4 @@
 	url = https://github.com/Auterion/android_openssl
 [submodule "libs/gst-plugins-good"]
 	path = libs/gst-plugins-good
-	url = https://gitlab.freedesktop.org/andrew.voznytsa/gst-plugins-good.git
+	url = https://github.com/mavlink/gst-plugins-good.git


dagar:
FYI @andrewvoznytsa

DonLakeFlyer:
Why isn't this repo under mavlink directly? Shouldn't need a mirror, it should be just there.

dagar:
It's now under mavlink with the default branch containing the commit used here. https://github.com/mavlink/gst-plugins-good

DonLakeFlyer:
Ok, not sure if that means the issue I created is fixed and can close or not? Can you do what's appropriate with that new issue?

dagar:
Fixes #8301

PR #8300, opened 2020-02-09
cmake add qmlglsink support (gst-plugins_good)

dagar:
FYI @andrewvoznytsa @tcanabrava @patrickelectric


dagar:
There's probably a better way to do some of this, but so far it works for me on Linux. Let's get it in and improve it incrementally.

dagar:
There's probably a better way to do some of this, but so far it works for me on Linux. Let's get it in and improve it incrementally.

Issue #8203, opened 2020-01-04
SiK radios and Mavlink2 not working really


olliw42:
I do have a copter which uses SiK 433MHz radios, and I can't get it working with QGC. QGC shows a warning that it uses Mavlink1 and also its console logging say that it is switching to Mavlink1. In addition the connection is very fragile and connection gets lost often and repeatedly, and further strange things happen. I don't have any issue when using MissionPlanner with the same setup, it just runs "for ever", and one can prove that Mavlink2 messages get through the SiK radios by the fact that MP receives messages with ID larger than 256, such as e.g MOUNT_ORIENTATION.

I use ArduCopter 4.0, have a STorM32 gimbal which is emitting Mavlink2 messages connected to it, use a QGC daily build of some days ago, and have just updated the SiK radios to the latest firmware with MissionPlanner.

My impression is that the issue is that the logic of QGC for switching to Mavlink1 or Mavlink2 isn't doing the right thing and makes things hang up. In MissionPlanner I can see a vehicle 51 with component ID 68 emitting a RADIO_STATUS, which is obviously the SiK radio, and it's maybe this RADIO_STATUS which confuses QGC. MissionPlanner's logic seems to make all parties happy and use Mavlink2.

Many thx!
Olli

![gcs-sik-01](https://user-images.githubusercontent.com/6089567/71768573-67c5f000-2f17-11ea-9617-321a721059de.jpg)
![qgc-sik-02](https://user-images.githubusercontent.com/6089567/71768575-6a284a00-2f17-11ea-8f09-2ab8e4f034b6.jpg)
![qgc-sik-03](https://user-images.githubusercontent.com/6089567/71768682-a90acf80-2f18-11ea-920e-4fc67c61209d.jpg)

auturgy:
The SiK firmwares from QGC and Missionplanner are different: Missionplanner uses the Ardupilotmega “RADIO” message (id 166), QGC uses the Common “RADIO_STATUS” message (id 109). It would surprise me (as I’d have thought it would have come up before), but as you’re using ArduPilot perhaps try the Missionplanner SiK firmware, and see if the problem persists?

olliw42:
oh ... thanks a lot for that info, it certainly is total news to me, THX

[...]

PR #8302, opened 2020-02-09
Disable SiK Radio mavlink 1 downgrade code

DonLakeFlyer:
Fix for #8203

MAVLinkProtocol had code which would downgrade the link to mavlink 1 based on the firmware of the MAVLINK_MSG_ID_RADIO_STATUS packet. This doesn't work because ArduPilot SiK Radio firmware always sends MAVLINK_MSG_ID_RADIO_STATUS as a mavlink 1 packet, even if the vehicle is sending mavlink 2.

### DIFF
--- a/src/comm/MAVLinkProtocol.cc
+++ b/src/comm/MAVLinkProtocol.cc
@@ -319,6 +319,11 @@ void MAVLinkProtocol::receiveBytes(LinkInterface* link, QByteArray b)
                 emit vehicleHeartbeatInfo(link, _message.sysid, _message.compid, highLatency2.autopilot, highLatency2.type);
             }
 
+#if 0
+            // Given the current state of SiK Radio firmwares there is no way to make the code below work.
+            // The ArduPilot implementation of SiK Radio firmware always sends MAVLINK_MSG_ID_RADIO_STATUS as a mavlink 1
+            // packet even if the vehicle is sending Mavlink 2.
+
             // Detect if we are talking to an old radio not supporting v2
             mavlink_status_t* mavlinkStatus = mavlink_get_channel_status(mavlinkChannel);
             if (_message.msgid == MAVLINK_MSG_ID_RADIO_STATUS && _radio_version_mismatch_count != -1) {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@LorenzMeier FYI since you added this code. This is the reason for my proposal to bring telemetry radio firmwares under mavlink central control with a single implementation overseen by all parties.

DonLakeFlyer:
@olliw42 I thinking some more about this I'm wondering whether this should go back in. The reason being that QGC tells you to update your firmware. If you use QGC to do that, this problem gets fixed. It also does catch the case where you are running pre-mavlink2 support sik radio firmware and warns appropriately. The thing which is a bit whacky is when it warns to upgrade firmware and you are running the mavlink 2 supporting ArduPilot variant. I this case theoretically you don't need to update firmware, but if you do the problem does fix itself. Also the PX4 version of SiK firmware does a better job I think of the new mavlink v2 protocol negotation whereas the ArduPilot firmware I don't think supports that. So it's not necessarily a bad thing to switch to the PX4 version.

Also as a side I'm trying to get these two firmwares to converge into one on the mavlink forum.

olliw42:
I'm sorry that I am a bit silent on this, QGC however fell a bit down on my list of things.

I can make a comment to this statement from top of my head however: As I pointed out in #8203 updating the firmware DOES NOT HELP THE ISSUE !!!! At least not with updating with the ArduPilot's version.

>The thing which is a bit whacky is when it warns to upgrade firmware and you are running the mavlink 2 supporting ArduPilot variant. I this case theoretically you don't need to update firmware, but if you do the problem does fix itself.

This statement is INCORRECT !!! I did upgrade and it did not help. And it can't as I demonstrated by the code.

 I don't know anything about PX4 version but form your text it reads to me one might suffer all sorts of incompatibility issues. Also, it's no option for me anyhow since, for some reason I never could figure out, I never managed to flash anything with QGC, only MP works for me.

So, the point is, you are making IMHO some inappropriate assumption here. Instead of forcing anyone else who have latest firmware installed into pain, I rather would argue that folks still running old Mavlink1 SiKs should upgrade ...

:):):)

Issue #5950, opened 2018-01-03
Log-Download Timed Out?


Marcusqjumpfxu:
Hi there,

I use QGroundControl quite often and I've discovered an issue with the download of Log-Files.

First the download is way to much than the size of the actual Log. Second the download timed out.

I've been connected via USB to my Pixhawk. OS: Ubuntu. QGroundControl: Nightly 2017-12-25.
![1](https://user-images.githubusercontent.com/35077586/34536130-85d4e1ca-f0c4-11e7-9d85-59f1ed04858b.png)
![3](https://user-images.githubusercontent.com/35077586/34536136-8a84ca0a-f0c4-11e7-89a1-79e4d3c4e586.png)
![2](https://user-images.githubusercontent.com/35077586/34536178-a83e7244-f0c4-11e7-9905-4b852961561d.png)

I've used an USB extension cable, but even without the extension cable it time out.
I use the newest stable firmware 3.5.4 ArduCopter.

The terminal gives me the following output if I try to download a log file:

```
LogDownloadLog: Request log data (id: 41 offset: 551880 size: 1080 )
LogDownloadLog: Request log data (id: 41 offset: 551880 size: 1080 )
LogDownloadLog: Request log data (id: 41 offset: 551880 size: 1080 )
LogDownloadLog: Request log data (id: 41 offset: 551880 size: 1080 )
LogDownloadLog: Request log data (id: 41 offset: 552960 size: 46080 )
LogDownloadLog: Request log data (id: 41 offset: 552960 size: 46080 )
LogDownloadLog: Request log data (id: 41 offset: 552960 size: 46080 )
LogDownloadLog: Request log data (id: 41 offset: 552960 size: 46080 )
Too many errors retreiving log data. Giving up.
```
In APM Planner 2.0.24 it works flawless.

I've posted this also on the [ardupilot discussion page](https://discuss.ardupilot.org/t/log-download-timed-out/24326).

DonLakeFlyer:
@dogmaphobic Do the size requests from the log make sense. Can't remember how that works. It suddenly bumps to a large size.

DonLakeFlyer:
I tried downloading some logs through SITL and it of course worked every time.

NaterGator:
@DonLakeFlyer if the segment the long downloader was working on had gaps (likely over any kind of less-than-perfect link) the smaller size can make sense (it would be the size of the missing contiguous gap).

DonLakeFlyer:
[...]

PR #8312, opened 2020-02-11
Infinite retries to log download

DonLakeFlyer:
Change log download to infinite retries on data download. If things go haywire and data stops coming across the data transfer rate will fall to 0 and the user can cancel. This should make it work even on really lossy connections.

Fix for #5950

### DIFF
--- a/src/AnalyzeView/LogDownloadController.cc
+++ b/src/AnalyzeView/LogDownloadController.cc
@@ -293,6 +293,24 @@ LogDownloadController::_findMissingEntries()
     }
 }
 
+void LogDownloadController::_updateDataRate(void)
+{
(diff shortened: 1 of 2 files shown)


fredowski:
Hi Don, thanks for the update. I tested this with the ESP8266 wifi module running firmware version 1.2.2 and Arduplane 4.0.3. The behavior for the log download is 

  * The download starts and reports 20kB/s rate
  * After approx. 60kB the download is stuck. QGC reports permantly reduced rates. No data is received.
  * After 8s to 20s the download continues for another maybe 40kB. 
  * After two minutes I have received around in total 300kB. 

I have not tested again the USB download. Do you have an idea where the log download protocol is described? Is this maybe a packet queuing problem where the esp stores many packets but cannot forward them and finally dumps the packets?

PR #8323, opened 2020-02-13
Don't report missing parameter FRAME_CLASS

### LINKED ISSUE #8319: Missing FRAME_CLASS parameter with ArduSub in frame setup page
Using QGC master and ArduSub 4.0.

Screenshot from 2020-02-12 16-02-30

Williangalvani:
ArduSub doesn't have this parameter, this gets rid of the "missing param" warning.

Fix #8319

### DIFF
--- a/src/AutoPilotPlugins/APM/APMAirframeComponentController.cc
+++ b/src/AutoPilotPlugins/APM/APMAirframeComponentController.cc
@@ -139,7 +139,7 @@ static QString s_findImageResourceRover(int frameClass, int frameType)
 }
 
 APMAirframeComponentController::APMAirframeComponentController(void)
-    : _frameClassFact   (getParameterFact(FactSystem::defaultComponentId, QStringLiteral("FRAME_CLASS")))
+    : _frameClassFact   (getParameterFact(FactSystem::defaultComponentId, QStringLiteral("FRAME_CLASS"), false /* reportMissing */))
     , _frameTypeFact    (getParameterFact(FactSystem::defaultComponentId, QStringLiteral("FRAME_TYPE"), false /* reportMissing */))
     , _frameClassModel  (new QmlObjectListModel(this))
 {


src/AutoPilotPlugins/APM/APMAirframeComponentController.cc
```diff
@@ -139,7 +139,7 @@ static QString s_findImageResourceRover(int frameClass, int frameType)
 }
 
 APMAirframeComponentController::APMAirframeComponentController(void)
-    : _frameClassFact   (getParameterFact(FactSystem::defaultComponentId, QStringLiteral("FRAME_CLASS")))
+    : _frameClassFact   (getParameterFact(FactSystem::defaultComponentId, QStringLiteral("FRAME_CLASS"), false /* reportMissing */))
```
DonLakeFlyer: This should be ArduSub only not for all firmwares. That said why is ArduSub page using APMAirframeComponenController? Seems like it should have it's own custom component here.

jaxxzer: > This should be ArduSub only not for all firmwares.

Why? This controller downloads default parameters for a given build/model from the ardupilot repository. That process is the same for each ardupilot vehicle.

> That said why is ArduSub page using APMAirframeComponenController? Seems like it should have it's own custom component here.

I will give you a brief history. please pay attention because it is a little confusing.
in mid-2016, ArduCopter had only the FRAME_TYPE parameter (no FRAME_CLASS) to choose +, x, octorotor, etc, ArduRover had no parameters relating to frame config, and ArduSub had the FRAME_CONFIG parameter that serves the same function as ArduCopter's FRAME_TYPE parameter. I added ArduSub support to the APMAirframeComponentController at that point in time, everything was the same or analagous between ardusub and arducopter.

Between now and then... ArduRover has added the FRAME_CLASS parameter to choose between boat, car, etc. Rover support was added to APMAirframeComponentController because this new parameter worked like the ones from sub and copter. Copter has added the FRAME_CLASS paremter **in addition to** the FRAME_TYPE parameter, so copter's frame setup has become multi-tiered and in the end different from rover and copter.

I think it's not necessary to create a new controller for just sub (note that there are already several checks in the source here to distinguish between rover and copter firmware instead of distinct classes for each). Let us know how you would like to proceed.

DonLakeFlyer: > This controller downloads default parameters for a given build/model from the ardupilot repository. That process is the same for each ardupilot vehicle.

That UI was disconnected ages ago from copter since as far as I could tell those files were not maintained in a decent state. Are you still using that?

DonLakeFlyer: If you are still using the param download code, then it make sense to use this controller. In which case then ok with the change.

If you are not, then the controller doesn't nothing for you as far as I can tell and you should just use the default FactPanelController.

Williangalvani: > If you are still using the param download code, then it make sense to use this controller. In which case then ok with the change.

That is right, the only reason we need it is for [downloading the default parameters](https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/APM/APMSubFrameComponent.qml#L217).

Can we merge it, then?

Issue #8296, opened 2020-02-09
Create Plan: Survey mission missing key steps


Antiheavy:
The new Create Plan function goes a long way to help people figure out mission planning on their own.  However, the Survey (and others) plan wizards miss some key steps that requires the user to go back and adjust the Survey (and others) every time.  

I think a lot of this could be avoided by letting the user choose a Preset (if they have any) and rotate the transects right in the wizard.

Here is a mock-up of what I am thinking:
![image](https://user-images.githubusercontent.com/12117027/74095174-9d2fa180-4ab2-11ea-8a31-c55310d33020.png)

This becomes more than just creating the Polygon so maybe the button becomes [Done Survey], not sure...

DonLakeFlyer:
I don't see that as being "Pre-Wizard" complete. To me it's just a change to show the Presets tab first if there are any presets. The presets tab already has transect rotation in it. Why isn't that good enough?

Antiheavy:
It's worth a discussion.  

When I'm using a Survey Wizard, I need to do that following every single time:
1) define the polygon
2) choose one of my preset survey types (I feel like there should always be a "custom" option in the presets drop-down that expands the pane to show all the adjustable items, but that is probably a separate discussion),
3) adjust the direction (based on geometry and wind). 

The above 3 items make me have to click on Survey after I've completed the Landing Pattern to go back and change everything around.  It's additionally confusing because I'm presented with all the details of the Landing Pattern once it is completed, and those are items I almost never change.  It's a disjointed thought process and not intuitive for users.

I don't need the other items in the Presets pane at this stage.  When using a Wizard I don't need buttons for [Apply Preset], [Delete Preset], or [Save Settings as New Preset].  I also don't necessarily need the statistics right away.

DonLakeFlyer:
Ok, I understand now. I wasn't thinking about the wizard moving you to the next item. And then having to come back.

PR #8326, opened 2020-02-13
Survey: Preset selection in wizard mode

DonLakeFlyer:
If the user has saved Presets then there is a second step to wizard mode.

First step continues to be polygon setup:
Screen Shot 2020-02-13 at 9 30 00 AM

But if you have any presets defined then when you click Done With Polygon you go here:
Screen Shot 2020-02-13 at 9 30 15 AM
(yellow button is a fluke of where I had the mouse when I captured).

Then when you click Done Adjusting you have completed the wizard.

Fix for #8296 (Survey only at this point).

### DIFF
--- a/src/PlanView/SurveyItemEditor.qml
+++ b/src/PlanView/SurveyItemEditor.qml
@@ -28,6 +28,9 @@ Rectangle {
     property real   _fieldWidth:                ScreenTools.defaultFontPixelWidth * 10.5
     property var    _vehicle:                   QGroundControl.multiVehicleManager.activeVehicle ? QGroundControl.multiVehicleManager.activeVehicle : QGroundControl.multiVehicleManager.offlineEditingVehicle
     property real   _cameraMinTriggerInterval:  missionItem.cameraCalc.minTriggerInterval.rawValue
+    property bool   _polygonDone:               false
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@antiheavy Can you test/verify for usability when this comes through. Then after things are good, I'll move to the other Patterns as well.

Antiheavy:
> @Antiheavy Can you test/verify for usability when this comes through. Then after things are good, I'll move to the other Patterns as well.

yes, excited to test!  I feel like we've discussed this before, but is the [Apply Preset] button really necessary, or can choosing a drop-down option just make that the selection?

DonLakeFlyer:
> I feel like we've discussed this before, but is the [Apply Preset] button really necessary, or can choosing a drop-down option just make that the selection?

Doing it that way is more complex to implement. I think also kinds of screws up the Apply/Delete button usage. I can look at some point but I"m trying to do simple things without affecting stable at this point. This new code just reuses existing stuff so it's pretty safe.

Antiheavy:
I did a quick sanity test and this is working as advertised.  thanks!  I'll come back to my thoughts about the drop down menu and "apply" buttons in a separate issue for future reference.

DonLakeFlyer:
Thanks for checking Todd.

PR #8328, opened 2020-02-13
Support colored text for checklist section headers

dakejahl:
Added properties to enable changing the color of the checklist section headers based on the pass/fail status of the checklist items.

If any item in the group fails, the group header will display the `failedTextColor`. Once all items pass, the color will be the `passedTextColor`. Otherwise, the color will remain `pendingTextColor`

### DIFF
--- a/src/QmlControls/PreFlightCheckButton.qml
+++ b/src/QmlControls/PreFlightCheckButton.qml
@@ -32,6 +32,7 @@ QGCButton {
     property bool   telemetryFailure:               false   ///< true: telemetry check failing, false: telemetry check passing
     property bool   allowTelemetryFailureOverride:  false   ///< true: user can click past telemetry failure
     property bool   passed:                         _manualState === _statePassed && _telemetryState === _statePassed
+    property bool   failed:                         _manualState === _stateFailed || _telemetryState === _stateFailed
 
     property int _manualState:          manualText === "" ? _statePassed : _statePending
     property int _telemetryState:       _statePassed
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Can you add some screenshots to show what this does?

dakejahl:
checklist_colored_headers

DonLakeFlyer:
Interesting but to me that looks really nasty from a visual standpoint. Don't see how it adds much value. Other problem is too much text coloring beyond the normal palette makes things even harder to read when the sun is on your screen.

If you really need control over this then you should add new colors to the palette for this. Regular QGC would just use the standard text color for all of them and you can change them to whatever your want in your custom build.

dakejahl:
Are you suggesting I add something like 
```
DECLARE_QGC_COLOR(checklistFailedText,                 "#9d9d9d", "#000000", "#707070", "#ffffff")
...
```
Which default to just black text for all 3 colors and then override these in the custom plugin like 
```
    QColor myCustomColor(58, 170, 53);

    if (colorName == QStringLiteral("checklistFailedText")) {
        colorInfo[QGCPalette::Light][QGCPalette::ColorGroupEnabled] =
                colorInfo[QGCPalette::Light][QGCPalette::ColorGroupDisabled] = myCustomColor;
        colorInfo[QGCPalette::Dark][QGCPalette::ColorGroupEnabled] =
                colorInfo[QGCPalette::Light][QGCPalette::ColorGroupDisabled] =myCustomColor;
    }
```

DonLakeFlyer:
Correct. As long as all the new palette entries start with the same values as the existing "text" entry.

PR #8329, opened 2020-02-13
Enforce checklist order property

dakejahl:
Added a property to the PreFlightCheckModel to enforce checklist order.

This can be set to `false` by a custom build that wishes to see all checklist subsections and complete them in any order.

### DIFF
--- a/src/QmlControls/PreFlightCheckModel.qml
+++ b/src/QmlControls/PreFlightCheckModel.qml
@@ -12,12 +12,17 @@ import QtQml.Models 2.1
 
 ObjectModel {
     id: _root
+    property bool enforceOrder: true
 
     function reset() {
         for (var i=0; i<_root.count; i++) {
             var group = _root.get(i)
             group.reset()
-            group.enabled = i === 0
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I don't quite get what this does?

dakejahl:
It allows the checklist to be completed in any order. Currently you have to complete each group sequentially.

DonLakeFlyer:
Ah, ok. So you need to add an enforceOrder option to QGCOptions which normal builds sets the true to retain current behavior. Then custom build overrides that to false and you get what you want.

dakejahl:
These changes just require the custom build to explicitly set `enforceOrder` to false in their implementation of the `PreFlightCheckModel`, which is done within their `custom/CustomBuildPreFlightCheckModel`

DonLakeFlyer:
> These changes just require the custom build to explicitly set `enforceOrder` to false in their implementation of the `PreFlightCheckModel`, which is done within their `custom/CustomBuildPreFlightCheckModel`

Ah, ok great. Fine then. Simpler.

PR #8330, opened 2020-02-13
Fix checklist (passed) header

dakejahl:
Fixes issue where the checklist is marked as **(passed)** when only a single group has passed.

With this PR the checklist will only be marked as **(passed)** once every group in the checklist has passed.

### DIFF
--- a/src/FlightDisplay/PreFlightCheckList.qml
+++ b/src/FlightDisplay/PreFlightCheckList.qml
@@ -42,14 +42,18 @@ Rectangle {
                 group = checkListRepeater.itemAt(index + 1)
                 group.enabled = true
                 group._checked = true
-            } else if ((index + 1) == checkListRepeater.count) {
-                // If the last group is passed, we mark the entire checklist as passed.
-                _passed = true
-                checklistDropPanel.close()
             }
-        } else {
-            _passed = false
         }
+
+        // Walk the list and check if any group is failing
+        var allPassed = true
+        for (var i=0; i < checkListRepeater.count; i++) {
+            if (!checkListRepeater.itemAt(i).passed) {
+                allPassed = false
+                break
+            }
+        }
+        _passed = allPassed;
     }
 
     //-- Pick a checklist model that matches the current airframe type (if any)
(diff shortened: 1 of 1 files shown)


src/FlightDisplay/PreFlightCheckList.qml
```diff
@@ -42,9 +42,14 @@ Rectangle {
                 group = checkListRepeater.itemAt(index + 1)
                 group.enabled = true
                 group._checked = true
+            } else if ((index + 1) == checkListRepeater.count) {
```
DonLakeFlyer: This won't work unless the group order is enforced. And you added the feature to allow order to not be enforced. Is that right?

DonLakeFlyer: The fix at quick glance is to walk the list using checkListRepeater.itemAt(...).passed for testing

src/FlightDisplay/PreFlightCheckList.qml
```diff
@@ -42,9 +42,14 @@ Rectangle {
...
+                // If the last group is passed, we mark the entire checklist as passed.
+                _passed = true
+                checklistDropPanel.close()
```
DonLakeFlyer: I'm not so sure about this. There were original problems for example with collapsing groups when they passed. It tends to happen when the last item to pass it not under user control. Things just magically collapse/go away and the user has no idea what just happen. With the time on group passing you can see the last item go green visually and then a short time after it closes. That way the user knows what happened. By closing the drop down right away and last pass you run into the same exact problem. I think it might be ok to auto-close, but if so it needs to work on a delayed timer basis just like the groups. That way you can see it mark as passed, register that in your head and then you see it close.

src/FlightDisplay/PreFlightCheckList.qml
dakejahl: Yup you're right, I've fixed the logic.

src/FlightDisplay/PreFlightCheckList.qml
dakejahl: Gotcha. I think now that you can just click away from the check list to minimize it the auto closing is not as important. I removed it.

DonLakeFlyer:
Yup, should be all good now

PR #8337, opened 2020-02-15
ROI enable/disable setting for Plan View

dakejahl:
Added setting to control visibility of ROI in the Plan View. This is to support custom builds in which supported vehicles do not use a gimbal and thus cannot use this feature.


DonLakeFlyer:
This should got into PlanViewSettings.h/cc and PlanView.SettingsGroup.json

Is there also the same problem with click to ROI support in the Fly View?

src/PlanView/PlanView.qml
```diff
@@ -53,6 +53,7 @@ Item {
     property int    _toolStripBottom:                   toolStrip.height + toolStrip.y
     property var    _appSettings:                       QGroundControl.settingsManager.appSettings
     property var    _planViewSettings:                  QGroundControl.settingsManager.planViewSettings
+    property bool   _enableROI:                         _appSettings ? _appSettings.enableROI : false
```
DonLakeFlyer: You shouldn't need the null check here. The settings objects area always available

Antiheavy:
> Is there also the same problem with click to ROI support in the Fly View?

I don't think so.  At least I don't see an ROI button with our custom plugin:
image

DonLakeFlyer:
There is an ROI state button in the toolbar and you click in the map and then pick ROI from the menu.

dakejahl:
Looks like the vehicle plugin assumes ROI is only supported for multirotors. Should we use that same assumption for Plan View instead of this new Setting?
https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc#L226-L235
https://github.com/mavlink/qgroundcontrol/blob/master/src/FlightDisplay/GuidedActionsController.qml#L110

Antiheavy:
> Looks like the vehicle plugin assumes ROI is only supported for multirotors.

I'm not familiar enough with how ROI works, but I'd think in the generic case, you'd want it to work with fixed wing and any vehicle that had a Gimbal or the ability to point the camera at a location and still effectively navigate at the same time.  Our fixed wing just used NADIR pointing mapping cameras so I don't have a dog in the game, but I'm just thinking about the generic case.

DonLakeFlyer:
@dogmaphobic Hmm, that doesn't look correct. For generic QGC I don't think you can't base ROI support on vehicle type. It should be available everywhere since you have no idea if there is gimbal support. The custom builds can override capabilities bits to specify ROI support. And then that capability bit should drive bothe the ROI in Fly and the ROI in Plan.

dakejahl:
Maybe we add support for this in MAVLink?

https://github.com/mavlink/mavlink/issues/1304

hamishwillee:
> Maybe we add support for this in MAVLink?

I think that is a good idea. Might be covered by the new gimbal API. Can you only do ROI with a gimbal (my guess is that you can do ROI on an MC without one)?

> Is there also the same problem with click to ROI support in the Fly View?

Generally yes. This shouldn't be available if the system doesn't support ROI. 

But then IMO it should be completely reworked anyway - the bit of the UI it is currently should only be used for status indication. If it were me, the button would move into the toolbar and arguably only be available in position mode, since that is the only mode in which an ROI can be set. The ROI status icon might still exist, but only to show that ROI is actually active.

DonLakeFlyer:
Wait I forgot about this:
```
    /// Used to determine whether a vehicle has a gimbal.
    ///     @param[out] rollSupported Gimbal supports roll
    ///     @param[out] pitchSupported Gimbal supports pitch
    ///     @param[out] yawSupported Gimbal supports yaw
    /// @return true: vehicle has gimbal, false: gimbal support unknown
    virtual bool hasGimbal(Vehicle* vehicle, bool& rollSupported, bool& pitchSupported, bool& yawSupported);
```

This gives custom builds the ability to tell QGC whether the vehicle has a gimbal or not. That is what should drive ROI availability. But it's not fully plumbed through to Fly/Plan. Let me fix all of this up using that.

dogmaphobic:
> That is what should drive ROI availability.

I would wait on that. Gimbal support in MAVLink is about to have a complete overhaul.

DonLakeFlyer:
When I went to work on #8383 I discovered I couldn't use this hasGimbal for ROI. So my statement is wrong.

dakejahl:
https://github.com/mavlink/qgroundcontrol/pull/8383

Issue #8334, opened 2020-02-14
MAVCMD retransmit an infinite number of time


FlavioTonelli:
In Vehicle::_handleCommandAck() the function _sendNextQueuedMavCommand() is called every time an ack is received.

In the case,GCS continuously receive a stream of ack that are not related to the MAVCMD QGC is trying to transmit then the retransmission logic break. GCS start retransmit the current CMD an infinity number of time, blocking all the other MAVCMD in queue.

DonLakeFlyer:
@FlavioTonelli If you have a repro, can you verify this fix?

DonLakeFlyer:
Reopen due to discussion on pull.

FlavioTonelli:
It seems it is working fine now. Adding or not the target id field is not preventing this issue to be closed. I still suggest to implement that to be more compliant to the MAVLink specification.

DonLakeFlyer:
> I still suggest to implement that to be more compliant to the MAVLink specification.

That's still marked as WIP which generally means don't use.

PR #8338, opened 2020-02-15
Fix placement of Vehicle::_sendNextQueuedMavCommand

DonLakeFlyer:
Fix for #8334

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -3469,6 +3469,7 @@ void Vehicle::_handleCommandAck(mavlink_message_t& message)
         _mavCommandAckTimer.stop();
         showError = _mavCommandQueue[0].showError;
         _mavCommandQueue.removeFirst();
+        _sendNextQueuedMavCommand();
     }
 
     emit mavCommandResult(_id, message.compid, ack.command, ack.result, false /* noResponsefromVehicle */);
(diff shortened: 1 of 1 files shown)


FlavioTonelli:
To let it be more robust I suggest also to check if the ack.target_system is QGC.

DonLakeFlyer:
> To let it be more robust I suggest also to check if the ack.target_system is QGC.

Hmm, missed this before I merged. That said I don't quite see how hose additional params will work. COMMAND_ACK.target_system is a newly added V2 value. Which means it defaults to 0 on old system. Given that how do you determine that value is not set and the target system id is actually 0. Is 0 not allowed for ids?

FlavioTonelli:
As far as I know 0 is equivalent to set the target system to broadcast. So it shall not be used as id of a specific MAVLink system.
What I'm afraid of is the case when:
I have 3 MAVLink system of which 1 is QGC( I have reproduced it with a drone and a companion computer). If I send command from the drone to the companion computer and I ack from the companion computer to the drone, QGC shall not handle that ack if the ack it's not target to him. That's the reason why target_system helps differenziate.
Cleaner solution seems to me to just handle the ack that has an ID equal to zero(MAVLink V1 case and/or broadcast) and do not handle the ack that specify a target not equal to the one of QGC).

PR #8340, opened 2020-02-16
Start Mission slider appears when checklist is complete

dakejahl:
When not using the preflight checklist, if a vehicle has a mission the **Start Mission** slider will pop up as soon as the vehicle connects.  This PR now provides the same behavior when using the checklist, as soon as the checklist is complete the **Start Mission** slider will appear

### DIFF
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -70,6 +70,12 @@ Item {
     readonly property string    _mainIsMapKey:          "MainFlyWindowIsMap"
     readonly property string    _PIPVisibleKey:         "IsPIPVisible"
 
+    on_CanArmChanged: {
+        if (_guidedController.showStartMission && _canArm) {
+            _guidedController.confirmAction(_guidedController.actionStartMission)
+        }
+    }
+
     function setStates() {
         QGroundControl.saveBoolGlobalSetting(_mainIsMapKey, mainIsMap)
         if(mainIsMap) {
--- a/src/FlightDisplay/FlightDisplayView.qml
+++ b/src/FlightDisplay/FlightDisplayView.qml
@@ -76,6 +76,12 @@ Item {
         }
     }
 
+    onVisibleChanged: {
+        if (visible && activeVehicle && !_canArm) {
+            checklistDropPanel.open()
+        }
+    }
+
     function setStates() {
         QGroundControl.saveBoolGlobalSetting(_mainIsMapKey, mainIsMap)
         if(mainIsMap) {


DonLakeFlyer:
Didn't @dogmaphobic just already add that somewhere? Or did that just prevent arming?

dakejahl:
Yes but the check for `_canArm` is inside of the `onShowStartMissionChanged` handler
```
            onShowStartMissionChanged: {
                if (showStartMission && _canArm) {
                    confirmAction(actionStartMission)
                }
            }
```

DonLakeFlyer:
> An alternate solution (which was my first solution) was to add the `_canArm` logic to the `GuidedActionsController.qml` however this just duplicates logic and is a bit messier overall.

That is how it should be done. Otherwise you will miss signalling of _canArm changes to pop show mission and right time and have to do random workaround stuff like this pull. Why does it duplicate logic?

dakejahl:
> Why does it duplicate logic?

It requires adding this to the `GuidedActionsController.qml`
```
    property bool   _useChecklist:                  QGroundControl.settingsManager.appSettings.useChecklist.rawValue && QGroundControl.corePlugin.options.preFlightChecklistUrl.toString().length
    property bool   _enforceChecklist:              _useChecklist && QGroundControl.settingsManager.appSettings.enforceChecklist.rawValue
    property bool   _canArm:                        activeVehicle ? (_useChecklist ? (_enforceChecklist ? activeVehicle.checkListState === Vehicle.CheckListPassed : true) : true) : false
```

And it requires removing `_canArm` from 
```
            onShowStartMissionChanged: {
                if (showStartMission && _canArm) {
                    confirmAction(actionStartMission)
                }
            }
```

Becayse there is a data race where the `GuidedActionsController` has emitted the `showStartMissionChanged` signal which trigger the handler in `FlightDisplayView`, however the `_canArm` variable in `FlightDisplayView` has not actually updated `_canArm` to the latest value. (I found this when testing).

So I went with the impl in this PR since it was less convoluted.

DonLakeFlyer:
The data race problem happens with all of these values. That's why in GuidedActionConfirm the actual display of a confirm slider is on a delay to allow for the rest of the values to catch up. That way an initial change true, which immediately after becomes false prevents the slider from popping up.

dakejahl:
FYI I edited my above comment

Also, I added another commit that opens the checklist when switching to the Flight View if enforcing the usage of the checlist. Let me know your thoughts on moving the logic to `GuidedActionsController` and I can rework this PR if needed.

DonLakeFlyer:
> it may be worth moving the `_canArm` logic out of Flight View entirely and moved directly into the logic in `GuidedActionsController` for

Yes that is how it should be. It should just be another bool used in the right places for the command enabler properties. Also the new bools should be added to the _outputState implementation so you can debug things from a stable build by turning on guidedActionsControllerLog

dakejahl:
I've done some work that is more comprehensive and actually fixes the core issues.

Work is continued [here](https://github.com/mavlink/qgroundcontrol/pull/8358)

PR #8341, opened 2020-02-16
Fix PreFlightBatteryCheck issue with NaN

dakejahl:
When no battery is connected PX4 reports NaN as the battery percent remaining. This causes the preflight battery check to pass when it shouldn't. This fixes it.

### DIFF
--- a/src/FlightDisplay/PreFlightBatteryCheck.qml
+++ b/src/FlightDisplay/PreFlightBatteryCheck.qml
@@ -25,6 +25,7 @@ PreFlightCheckButton {
 
     property int    failurePercent:                 40
     property bool   allowFailurePercentOverride:    false
-    property var    _batPercentRemaining:           activeVehicle ? activeVehicle.battery.percentRemaining.value : 0
+    property var    batteryValue:                   activeVehicle ? activeVehicle.battery.percentRemaining.value : 0
+    property var    _batPercentRemaining:           batteryValue ? batteryValue : 0
     property bool   _batLow:                        _batPercentRemaining < failurePercent
 }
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Oops, yeah I changed that recently which then screwed this up. @dakejahl Can you check whether the battery toolbar indicator has problems now as well?

dakejahl:
I think it's working as expected
battery_status

DonLakeFlyer:
Looking at the code the problem is bigger than this. In reality if percentRemaining is NaN then it means that the vehicle is not reporting that value. Which is really a different error than your battery is dead. Wondering how you are hitting this?

dakejahl:
>Wondering how you are hitting this?

Flight controller plugged in via USB.

DonLakeFlyer:
Ah, ok. The force to 0 seems ok then.

src/FlightDisplay/PreFlightBatteryCheck.qml
```diff
@@ -25,6 +25,7 @@ PreFlightCheckButton {
 
     property int    failurePercent:                 40
     property bool   allowFailurePercentOverride:    false
-    property var    _batPercentRemaining:           activeVehicle ? activeVehicle.battery.percentRemaining.value : 0
+    property var    batteryValue:                   activeVehicle ? activeVehicle.battery.percentRemaining.value : 0
```
DonLakeFlyer: This should be _batteryValue

src/FlightDisplay/PreFlightBatteryCheck.qml
```diff
@@ -25,6 +25,7 @@ PreFlightCheckButton {
...
+    property var    _batPercentRemaining:           batteryValue ? batteryValue : 0
```
DonLakeFlyer: Making this an actual NaN check would be clearer

dakejahl: How do I do that? Using `_batteryValue.NaN` or `_batteryValue === NaN` doesn't seem to work.

DonLakeFlyer: isNaN(_batteryValue) Qml is jscript.

Issue #8208, opened 2020-01-08
Strange altitude sent in follow mode


twistedwings:
I switched to follow mode and walked along a flat area.
This is the altitude ArduCopter got:

![image](https://user-images.githubusercontent.com/562809/72014950-23ad5500-3261-11ea-8c43-d3956f2fff3d.png)

see also [https://discuss.ardupilot.org/t/strange-altitude-in-follow-mode/50781/4](url)

ArduCopter 4.0.0
QGC Android daily 2020-01-03

DonLakeFlyer:
I'm going to need to add some additional logging to the builds to figure out what is going on here. Hope to get that some in a day or so.

Thanks for the bug reports.

Antiheavy:
I assume QGC is using the location services from the device (some android device in this case)?  Depending on the quality of your device's GPS signal, it could be possible that your device's altitude estimate is really that terrible...  or maybe its a bug...

DonLakeFlyer:
> I assume QGC is using the location services from the device (some android device in this case)? Depending on the quality of your device's GPS signal, it could be possible that your device's altitude estimate is really that terrible... or maybe its a bug...

That is correct as is the comment on crap GPS Alt. I hope to have time soon to add logging of location info in QGC so that it can be verified to be that or some other problem.

DonLakeFlyer:
@twistedwings New daily builds should be available in an hour. Can you use the new version. Turn on FollowMeLog console logging. Capture the output while moving around and attach here? This should tell me what is really being reported as altitude from the device.

twistedwings:
may take a while as it's raining all day long.

twistedwings:
Where do I enable FollowMeLog? I cannot find that option.

![image](https://user-images.githubusercontent.com/562809/72628506-bed3b800-394e-11ea-9520-d77706c70b61.png)

dogmaphobic:
Under "Console" in _General Settings_:

<img width="1026" alt="Screen Shot 2020-01-17 at 1 24 22 PM" src="https://user-images.githubusercontent.com/749243/72632555-e580f700-392c-11ea-9d3b-49486360eb11.png">

twistedwings:
I don't have that, do I need a newer version?

![image](https://user-images.githubusercontent.com/562809/72634500-ff393300-395a-11ea-9f6f-db6fcb8bb830.png)

dogmaphobic:
It's in the current master (daily) build.

[...]

Issue #7811, opened 2019-09-13
Support of ArduPilot Copter/Rover Follow mode


rmackay9:
ArduPilot Copter and Rover have a [Follow flight mode](http://ardupilot.org/rover/docs/follow-mode.html) which is implemented on the flight controller.    There are pros and cons of implementing the logic in the vehicles vs the ground station but in any case, this enhancement request is to make use of the flight controller based Follow mode for regular "FollowMe" where the vehicle follows the person holding the ground station computer.

To make this work, the ground station must publish it's position at 1hz (or faster if possible) to the vehicle using the GLOBAL_POSITION_INT message.  Beyond this it's possible that the GCS user could modify the position of the vehicle etc with one of the existing mavlink parameters but I haven't looked into this much.

FYI we have a somewhat related [issue on the AP side here](https://github.com/ArduPilot/ardupilot/issues/8517) to support QGC's GCS based follow-mode (i.e. the logic is done on the ground station computer) but I think we should support both methods.

Some videos of the mode in action:

- [Copter following a plane](https://www.youtube.com/watch?v=0s1ZBSB0tvM)
- [Rovers following each other](https://www.youtube.com/watch?v=_g9SkK0IhRk)
- [Boats following each other](https://www.youtube.com/watch?v=ZSzI7XddT4A)

DonLakeFlyer:
Should it publish GLOBAL_POSITION_INT all the time to any ArduPilot vehicle? Or just when the flight mode is follow me?

rmackay9:
@DonLakeFlyer, it only needs to send the position when it's in Follow mode.  The vehicle will stop and hold position if it's not getting position updates.

DonLakeFlyer:
@rmackay9 I assume GLOBAL_POSITION_INT.relative_alt should always be 0?

DonLakeFlyer:
@rmackay9 Also ok to set GLOBAL_POSITION_INT.hdg = UINT16_MAX or do you need that?

rmackay9:
@DonLakeFlyer, for ground vehicles it won't matter but for flying vehicles it can.  On the AP side we have a parameter (FOLL_ALT_TYPE) which allows the user to decide if it will use the relative or absolute altitude from the lead vehicle/GCS.  By default is uses the relative alt.

I guess it depends upon whether we want the feature to be useful on ski hills or not.

DonLakeFlyer:
[...]

PR #8350, opened 2020-02-17
ArduPilot: Change follow to work above home altitude only

DonLakeFlyer:
Initial starting point is this:
Screen Shot 2020-02-17 at 12 06 33 PM

When you click Enable for the first time you get this:
Screen Shot 2020-02-17 at 12 06 40 PM
This is due to the fact that the default value set by the firmware are not usable by QGC setup. You can also end up here if you futz parameters in the parameter editor and then to go this page. Clicking 'Reset To Supported Settings' will set things up to a decent set of workable defaults.

Which then gets you here:
Screen Shot 2020-02-17 at 12 17 02 PM
I'm hoping this is all self explaining.

Then to active Follow Me you change to Follow Me flight mode.

Some important things to note:
* The height you are specifying is a relative height above launch/home position.
* If you are using a mode which requires a heading from the gcs then you may need to walk in a specific direction for a short distance such that the gcs gps has a valid heading to send.

Related to #8711

@hamishwillee Hoping this is enough details for you to help with docs.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMFollowComponent.FactMetaData.json
+++ b/src/AutoPilotPlugins/APM/APMFollowComponent.FactMetaData.json
@@ -20,7 +20,7 @@
 },
 {
     "name":             "height",
-    "shortDescription": "Vertical distance from ground station to vehicle",
+    "shortDescription": "Vertical distance from Launch (home) position to vehicle",
     "type":             "double",
     "min":              0,
     "decimalPlaces":    1,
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
Also fix for #8208 since it doesn't work that way any more.

DonLakeFlyer:
@twistedwings Do you time to give it a try. Sorry it's been such an adventure with this new feature!

jstrebel:
Don, is this great feature also usuable for PX4?
Would be great. Jakob

DonLakeFlyer:
Follow has been available for PX4 for a while. Search for PX4 Follow Me. I doesn't have a UI though. Set up through parameter editor.

jchengethub:
@DonLakeFlyer : I am trying to make a RC car with Pixhawk, Rover to follow a person who walking with QGC installed in Android phone. The purpose of this job is to help people who are inconvenient to carry things. I read the related issues (#7811, 7841, 8350) and find out the "Follow Me" mode had been removed from stable (#8413) for some reasons. Is it possible to download this version of your work and then try to modify by Qt for the function of RC car follows person safely. Many thanks

Issue #8356, opened 2020-02-19
Popup checklist on Flight View if using enforceChecklist


dakejahl:
I think if the **enforceChecklist** setting is enabled and the checklist is not complete, the checklist should popup as soon as you navigate to the Flight View _from a different view_. This is helpful because it makes it very clear to the user that they need to complete the checklist before they're allowed to arm.

The issue with this is the the `visible` property of the Flight View (as well as all the other views). If you are already in Flight View and click the Flight View button again, it will mark the view as `visible = false` and then `visible = true` 
https://github.com/mavlink/qgroundcontrol/blob/master/src/ui/MainRootWindow.qml#L102-L118

This prevents using the `onVisibleChanged` signal to popup the checklist, because if you're trying to go into advanced mode in a custom build you need to be able to click the Flight View multiple times _without_ the checklist popping up. 

Effectively the current behavior of clicking the view button of a view you are already in is to "refresh" the view. Is this useful? I would expect clicking the button for a view you are already in to do nothing. 

**edit:** this is pretty easy to handle properly, I opened the issue because I was unsure if there is a good reason to set all views to false before setting the desired view to true.

DonLakeFlyer:
It's just a coding simplification. The concept of turn everything off, and then turn the one thing you want back on is a pretty common pattern to use since it leads to way less and simpler code. Generally it doesn't cause problems because of the delay in UI rendering. You'll never see the intermediate state. But if you are connected directly to that property to do something then you see it.

PR #8358, opened 2020-02-19
Fixed interactions between PreFlightCheckList and the FlightView

dakejahl:
This PR accomplishes 3 things:

1. Checklist logic for `_canArm` when using **enforceChecklist** has moved to the GuidedActionsController. This allows the Flight View to properly catch the signals that depend on the preflight checklist (canArm, startMission, etc)

2. Updating of the `Vehicle.checkListState` in the PreFlightCheckList from the `onAllChecksPassedChanged` signal, as opposed to the `onVisibleChanged` signal. 

3. Added a timer with a 1 second delay to forcing opening the checklist when navigating to the Flight View if using **enforceChecklist**. This timer also triggers closing the checklist once complete.


dakejahl:
We might want to have a quick discussion on the issue raised in https://github.com/mavlink/qgroundcontrol/issues/8356

PR #8361, opened 2020-02-19
Fix possibility of AUTOPILOT_VERSION message crashing QGC

stefandunca:
`FirmwarePlugin::checkIfIsLatestStable` is executed on every `AUTOPILOT_VERSION` message received from the vehicle which starts a file download to check the latest stable version. There were cases reported by autopilot sending too much of these and breaking QGC. Also on a failed attempt, we leave the object hanging.
This PR limits the firmware version file fetch to one instance only and logs the error reports as debug.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc
+++ b/src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc
@@ -79,8 +79,9 @@ const QVariantList& APMAutoPilotPlugin::vehicleComponents(void)
                 _components.append(QVariant::fromValue((VehicleComponent*)_radioComponent));
             }
 
+            FwVersion vehicleVersion(_vehicle->firmwareMajorVersion(), _vehicle->firmwareMinorVersion(), _vehicle->firmwarePatchVersion());
             // No flight modes component for Sub versions 3.5 and up
-            if (!_vehicle->sub() || (_vehicle->versionCompare(3, 5, 0) < 0)) {
+            if (!_vehicle->sub() || (vehicleVersion.compare(FwVersion(3, 5, 0)) < 0)) {
                 _flightModesComponent = new APMFlightModesComponent(_vehicle, this);
                 _flightModesComponent->setupTriggerSignals();
                 _components.append(QVariant::fromValue((VehicleComponent*)_flightModesComponent));
@@ -94,7 +95,7 @@ const QVariantList& APMAutoPilotPlugin::vehicleComponents(void)
             _powerComponent->setupTriggerSignals();
             _components.append(QVariant::fromValue((VehicleComponent*)_powerComponent));
 
-            if (!_vehicle->sub() || (_vehicle->sub() && _vehicle->versionCompare(3, 5, 3) >= 0)) {
+            if (!_vehicle->sub() || (_vehicle->sub() && vehicleVersion.compare(FwVersion(3, 5, 3)) >= 0)) {
                 _motorComponent = new APMMotorComponent(_vehicle, this);
                 _motorComponent->setupTriggerSignals();
                 _components.append(QVariant::fromValue((VehicleComponent*)_motorComponent));
@@ -111,7 +112,7 @@ const QVariantList& APMAutoPilotPlugin::vehicleComponents(void)
                 _components.append(QVariant::fromValue((VehicleComponent*)_followComponent));
             }
 
-            if (_vehicle->vehicleType() == MAV_TYPE_HELICOPTER && (_vehicle->versionCompare(4, 0, 0) >= 0)) {
+            if (_vehicle->vehicleType() == MAV_TYPE_HELICOPTER && (vehicleVersion.compare(FwVersion(4, 0, 0)) >= 0)) {
                 _heliComponent = new APMHeliComponent(_vehicle, this);
                 _heliComponent->setupTriggerSignals();
                 _components.append(QVariant::fromValue((VehicleComponent*)_heliComponent));
@@ -130,7 +131,7 @@ const QVariantList& APMAutoPilotPlugin::vehicleComponents(void)
                 _lightsComponent->setupTriggerSignals();
                 _components.append(QVariant::fromValue((VehicleComponent*)_lightsComponent));
 
-                if(_vehicle->versionCompare(3, 5, 0) >= 0) {
+                if(vehicleVersion.compare(FwVersion(3, 5, 0)) >= 0) {
                     _subFrameComponent = new APMSubFrameComponent(_vehicle, this);
                     _subFrameComponent->setupTriggerSignals();
                     _components.append(QVariant::fromValue((VehicleComponent*)_subFrameComponent));
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -827,6 +827,7 @@ void APMFirmwarePlugin::_artooSocketError(QAbstractSocket::SocketError socketErr
 
 QString APMFirmwarePlugin::internalParameterMetaDataFile(Vehicle* vehicle)
 {
+    FwVersion vehicleVersion(vehicle->firmwareMajorVersion(), vehicle->firmwareMinorVersion(), vehicle->firmwarePatchVersion());
     switch (vehicle->vehicleType()) {
     case MAV_TYPE_QUADROTOR:
     case MAV_TYPE_HEXAROTOR:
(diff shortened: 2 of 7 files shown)


DonLakeFlyer:
FirmwarePlugin instances can't maintain any state associated with a vehicle since it is a singleton. Also there may be multiple checkIfIsLatestStable downloads in play at the same time if you have multiple vehicles connected. So it is required to support more than one at a time. Also just for FYI the AutoPilotPlugin is different in that there is an instance per Vehicle. So that sort of thing is ok there.

The way to deal with the FirmwarePlugin singleton thing is to maintain all state in Vehicle. There should be a bool in Vehicle which specifies whether the firmware version check was already done oce. And the bool should be checked here: https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L1347 so that the FirmwarePlugin call is only made once per vehicle.

For error cleanup, the code is missing a similar connect to QGCFileDownload::error which should do a sender->deleteLater() as well.

stefandunca:
> FirmwarePlugin instances can't maintain any state associated with a vehicle since it is a singleton.

It just keeps the QGCFileDownload instance for decoupling vehicle events (`AUTOPILOT_VERSION`) from firmware version file downloading. The vehicle instance is temporarily kept in the message, which might be a problem if it is deleted while waiting for the version download. This might be easily fixed by keeping and passing the Vehicle instance as a shared pointer. However, it's a huge change outside the scope of this PR, hence the TODO.

> Also there may be multiple checkIfIsLatestStable downloads in play at the same time if you have multiple vehicles connected. So it is required to support more than one at a time. Also just for FYI the AutoPilotPlugin is different in that there is an instance per Vehicle. So that sort of thing is ok there.

I will fix this use-case by implementing the following two-step procedure:
 - downloading the version file if missing from cache or too old
 - use the cached file version for checking the firmware update availability

> The way to deal with the FirmwarePlugin singleton thing is to maintain all state in Vehicle. There should be a bool in Vehicle which specifies whether the firmware version check was already done oce. And the bool should be checked here: https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L1347 so that the FirmwarePlugin call is only made once per vehicle.

I don't see this as a good reason to pollute the already huge Vehicle class with this responsibility. The `AUTOPILOT_VERSION` message should be sent only once when requested by Vehicle using `MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES`. The quick fix I was trying to do is for stability/security, for when the system misbehaves and sends too many AUTOPILOT_VERSION which opens too many download sessions and QGC crashes.

> For error cleanup, the code is missing a similar connect to QGCFileDownload::error which should do a sender->deleteLater() as well.

The second connect I added does exactly this.

DonLakeFlyer:
Ok understood, so my comments above are pretty much wrong. There isn't any vehicle oriented state here since there is only a single "latest version of firmware" for all vehicles no matter what. So the concept of multiple downloads in play is incorrect. I wasn't thinking about this the right way.

> * use the cached file version for checking the firmware update availability

Certainly a nice optimization. But is it really needed right now? It will also delay notifying the user of new releases depending on how long you accept the cache. Sometimes that can be important if the new release has a critical fix in it. QGC itself always also checks the internet for new version every time you boot. I think that's all fine for now with respect to trying to wrap up a Stable.

> The vehicle instance is temporarily kept in the message, which might be a problem if it is deleted while waiting for the version download. This might be easily fixed by keeping and passing the Vehicle instance as a shared pointer.

Shared pointers can be a real pain. The way this is handled in many other places is to connect a signal the firmware version back to the vehicle and then let the vehicle do the version compare and show the ui. That way if the Vehicle goes away the signal gets disconnected automatically. In reality this matches much better from an architecture standpoint since it's odd to have FirmwarePlugin pop ui.

> I don't see this as a good reason to pollute the already huge Vehicle class with this responsibility. The `AUTOPILOT_VERSION` message should be sent only once when requested by Vehicle using `MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES`.

How are you going to prevent the ui which says there is an new version from popping up multiple times then if this happens? Won't crash any more but still pretty annoying. A bool in Vehicle is still the simplest way to fix this with less code.

> The second connect I added does exactly this.

Yup. Missed that.

So in general I continue to think that preventing a Vehicle from requesting the version check more than once is the way to go with a bool. Simpler, less code and correct place as far as I'm concerned. Also much safer change right before a Stable goes out.  The rest of the problems (caching, vehicle crashing case, even the memory loss on error) can go into an Issue and be dealt with in the next point release. I'd be happy to take care of those after Stable.

DonLakeFlyer:
> So in general I continue to think that preventing a Vehicle from requesting the version check more than once is the way to go with a bool. Simpler, less code and correct place as far as I'm concerned. Also much safer change right before a Stable goes out.

I'd be happy to make this change if you want and I'll add an Issue to clean up the other stuff. I know it's a pain to fight with QGC architecture at times.

stefandunca:
> Certainly a nice optimization. But is it really needed right now? It will also delay notifying the user of new releases depending on how long you accept the cache.

I will do more testing tomorrow and it will enter our manual testing cycle later next week. You can postpone it for this release or get something quickly and safe in.

> Shared pointers can be a real pain.

I can't agree on this and the gain that managed languages took over C++ is a testimony and part of it due to object reference management. I also worked on much bigger projects where all the high-level module management was done using managed pointers (shared/weak) and it worked exceptionally.

> The way this is handled in many other places is to connect a signal the firmware version back to the vehicle and then let the vehicle do the version compare and show the UI. That way if the Vehicle goes away the signal gets disconnected automatically. In reality, this matches much better from an architecture standpoint since ...

For certain using the observer pattern to do memory and dependency management is much error-prone, hard to understand and has a bigger performance penalty. Also debugging dangling/misusage of shared instances is much easier than tracking signals and slots. Not to mention that is the right way to go when you want to manage an object that has no dependency on its parent's lifespan.

> it's odd to have FirmwarePlugin pop UI.
I agree that the separation between business logic and UI is the weakest link in the entire architecture and should be the major focus going forward.

> How are you going to prevent the ui which says there is a new version from popping up multiple times then if this happens? Won't crash any more but still pretty annoying. A bool in Vehicle is still the simplest way to fix this with less code.
It might be simple but it ads another dependency and responsibility to the Vehicle class which is already a GOD object hard to track and grasp.

> I'd be happy to make this change if you want and I'll add an Issue to clean up the other stuff. I know it's a pain to fight with QGC architecture at times.
I already did the quickest thing I could do to check the major issues and there is still more to be done to fix it properly.

DonLakeFlyer:
> Shared pointers can be a real pain.

I'm not commenting on the general use of shared pointers. I'm talking about retrofitting shared pointer usage specfically for a single object like Vehicle into a QGC with generally does not use shared pointers. I did the same thing with LinkManager links to solve a problem that couldn't be solved any other way and yes it was better technically but was a royal pain to work it through the codebase. 100% shared pointers top to bottom is nice, partial usage is not.

DonLakeFlyer:
There is an amount of good stuff in here with respect to caching fw versions. But what I call "rearranging the deck chairs" with respect to version compare doesn't really accomplish anything other than possibly introduce bugs. In fact it makes the code larger and more complex as far as I'm concerned. In the process you've broken ArduSub. The version compare apis were on vehicle as invokable for a reason.

stefandunca:
> But what I call "rearranging the deck chairs" with respect to version compare doesn't really accomplish anything other than possibly introduce bugs.

I was trying to make it more understandable based on my struggle to debug the issue and hoping that others won't go through the same process. I call this going slowly towards cleaner code that is testable, doesn't take too much time and the devs are not afraid to fix it fearing that will break things. 

> In the process you've broken ArduSub. The version compare apis were on vehicle as invokable for a reason.

I think this supports some of my points. What I thought I was doing was only directly called the compare function instead of going through the vehicle redirection thus breaking the direct undeeded dependency. If there is something more then it's hidden and tribal knowledge. Still, there is value for you to share the reason.

I was looking through creating a set of unit tests for FirmwarePlugin to complete the change. This is not straight forward with the current approach. It will require mocking a vehicle instance.

stefandunca:
Replaced by: https://github.com/mavlink/qgroundcontrol/pull/8377

DonLakeFlyer:
I don't understand what is simpler than the Vehicle maintains it's state. It understands what version of firmware it is running and you can ask it to compare it's known version to a specific version number. That's pretty simple stuff. Your definition of cleaner is radically different than mine.

There is nothing more simple than a Q_INVOKABLE. Which means the call is being made through Qml. And a simple search will find you those references. That is not tribal knowledge that is basic Qt/Qml C++ interaction.

The pattern is that almost all traffic goes directly through Vehicle which in turns hides the fact that it talks to FirmwarePlugin behind the scenes. This provides a simple way to access Vehicle related functionality through both C++ and Qml. And it removes Firmware plugin code/signature changes from the code which references Vehicle such that if FirmwarePlugin changes the code above which references the Vehicle does not. Only the interior usage of FirmwarePlugin inside Vehicle does.

DonLakeFlyer:
> The pattern is that almost all traffic goes directly through Vehicle which in turns hides the fact that it talks to FirmwarePlugin behind the scenes.

The reason is that there is an "almost" here is because that pattern is mostly followed for code exposed to Qml.

DonLakeFlyer:
> I call this going slowly towards cleaner code that is testable, doesn't take too much time and the devs are not afraid to fix it fearing that will break things.

The really important thing to understand is that QGC doesn't really fit a model where it's ok to no be overly worried about breaking things. I've had this same discussion with many folks, so this isn't just you. 

QGC has very few devs working on it, maintaining it and testing it. It also has next to no unit or functional test coverage. You put those two things together and a number of things happen:
* There is a very limited amount of devs available to do code reviews (mainly just me or Gus)
* It is easy to break things without finding the problem till months later
* The people who find the problem are end users. They have two responses to finding the problem. The majority response will be to consider QGC crap and move on. Possibly dropping out of the PX4 ecosystem altogether. And the minority will file an Issue so we can know about the problem and fix it.
* The original developer who changed the code is long gone, not interested, working on something else now or doesn't have the time to come back to fix the bug they introduced.

Which in turn leads to me having to do two things:
* Spend a huge amount of time doing code reviews trying to find bugs in "cleaned up" code
* Spend a huge amount of time fixing other peoples bugs

So any sort of rewriting of working code needs to be driven by a large problem of understandability, usability or bugginess. The new video work is a great example of an area which met the bar for rework. Otherwise it is best to leave it alone. Without that forward progress on QGC will screech to a halt and quality will drop. This isn't just an opinion. This is my day to day reality of over 6+ years of rewriting QGC completely from scratch to what it is now. I know it's a PITA to contributors. But it is meant to create less of a PITA for me who deals with all the day to day support of QGC. Maybe that will change with Auterion contributing additional QGC devs/testing/customer support to the mix. But it hasn't happened yet.

There is also some amount of this talked about this here in the dev guide: https://dev.qgroundcontrol.com/en/ui_design/multi_device_pattern.html.

stefandunca:
Thanks for sharing. This clarifies a lot of things I see around QGC codebase.

PR #8364, opened 2020-02-19
QGCMavlink: Add GCC version check to use address-of-packed-member

patrickelectric:


patrickelectric:
@DonLakeFlyer I'm not sure which warning from mavlink you are getting with GCC 7.4.0 when compiling QGC, so I just put -Wall. PS: -Wall does not help with GCC 9+

patrickelectric:
It appears to be working under travis, let me know if it's not working for you @DonLakeFlyer.

PR #8379, opened 2020-02-21
Try to fix ubuntu 16.04 library copying

tcanabrava:
Only copy specific ubuntu libraries if in ubuntu 16.04.

### DIFF
--- a/QGCSetup.pri
+++ b/QGCSetup.pri
@@ -121,14 +121,14 @@ LinuxBuild {
         libicui18n.so* \
         libQt5TextToSpeech.so.5
 
-    DISTRO_NAME = $$system(cat /etc/os-release | grep "NAME=")
-    contains(DISTRO_NAME, "Ubuntu") {
-        VERSION = = $$system(cat /etc/os-release | grep "VERSION=")
-        contains(VERSION, "16.04") {
-            QT_LIB_LIST += \
-            libicudata.so.56 \
-            libicui18n.so.56 \
-            libicuuc.so.56
+        exists($$[QT_INSTALL_LIBS]/libicudata.so.56) {
+            QT_LIB_LIST += libicudata.so.56
+        }
+        exists($$[QT_INSTALL_LIBS]/libicui18n.so.56) {
+            QT_LIB_LIST += libicui18n.so.56
+        }
+        exists($$[QT_INSTALL_LIBS]/libicuuc.so.56) {
+            QT_LIB_LIST += libicuuc.so.56
         }
     }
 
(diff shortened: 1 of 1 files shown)


tcanabrava:
@DonLakeFlyer  - This *should* fix it, I'm waiting for the buildsystem to see if it goes thru.
I'll close the other PR.

tcanabrava:
@DonLakeFlyer  you know more qmake than me, I have no idea why this code is failing, can you spare two minutes? Seems sane: try to copy the files only if they exist in the filesystem. for some reason it's failing on debug builds.

QGCSetup.pri
```diff
@@ -121,12 +121,15 @@ LinuxBuild {
         libicui18n.so* \
         libQt5TextToSpeech.so.5
 
-    !contains(DEFINES, __rasp_pi2__) {
-        # Some Qt distributions link with *.so.56
-        QT_LIB_LIST += \
-            libicudata.so.56 \
-            libicui18n.so.56 \
-            libicuuc.so.56
+        exists($$[QT_INSTALL_LIBS]/libicudata.so.56) {
```
DonLakeFlyer: I think this should be either just $$QT_INSTALL_LIBS, or $${QT_INSTALL_LIBS}. Something seems to be confusing qmake breaking some variable expansion. This is just a guess.

DonLakeFlyer:
No idea what is going on here...

booo:
@tcanabrava Are you soure the failing build is related to your changes? Can you maybe rebase and try again? I couldn't find log output related to your changes.

And I think we should just use libicudata.so as it is provided by xenial, bionic and my archlinux system.

tcanabrava:
> @tcanabrava Are you soure the failing build is related to your changes? Can you maybe rebase and try again? I couldn't find log output related to your changes.
> 
> And I think we should just use libicudata.so as it is provided by xenial, bionic and my archlinux system.

yes, my changes ignore the libicudata, so it breaks on CI because the CI uses a specific version, I'm on archlinux too - so my fixes here fixes thigns for arch, breaks for CI.

jlucke86:
Why is the library copy hard coded to libicudata.so.56? I've managed to get builds working on yocky poky thud, zeus and warrior and rhel8.1 they all provide different versions of that library and they all work without issue.

booo:
@jlucke86 Could you check my change: https://github.com/mavlink/qgroundcontrol/pull/8599

I think we should just link against libicudata.so.

DonLakeFlyer:
Fails build

PR #8383, opened 2020-02-21
Drive Plan view ROI support from firmware capability bits

DonLakeFlyer:
This makes Plan view using the same capability bit as Fly view for ROI support. I've left the bit set on only for multi-rotor which isn't technically correct. But closer to vehicle majority usage.

Replacement for #8337.

@dakejahl So to turn off ROI support fully in Sentera custom you remove the ROI capability in the Sentera firmware plugin override.


DonLakeFlyer:
@dogmaphobic I assume this isn't a problem for Auterion

DonLakeFlyer:
We are going to have to sort out getting a truer (not hardcoded) understanding of whether a vehicle has a gimbal on it for not or has ROI support somehow. But that's a problem for later.

PR #8386, opened 2020-02-21
Fix view visible property toggle on redundant click

dakejahl:
Fixes an issue where if clicking the view button in the main toolbar for a view you're already in, the view will toggle the `visible` property causing the `onVisibleChanged` handler to run. This is an issue for code that relies on that signal, i.e https://github.com/mavlink/qgroundcontrol/pull/8358/files

### DIFF
--- a/src/ui/MainRootWindow.qml
+++ b/src/ui/MainRootWindow.qml
@@ -99,42 +99,49 @@ ApplicationWindow {
         return _rgPreventViewSwitch[_rgPreventViewSwitch.length - 1]
     }
 
-    function viewSwitch(isPlanView) {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
All that duplication is just nasty. 

What I would do is expose a function called something like showPreflightChecklistIfNeeded on the fly view. And then do this:
```
    function showFlyView() {
        viewSwitch(false)
        flightView.visible = true
        flightView.showPreflightChecklistIfNeeded()
    }
```

dakejahl:
https://github.com/mavlink/qgroundcontrol/pull/8389

PR #8391, opened 2020-02-22
Vagrant: move to bionic64

peterbarker:
xenial fails with this error:

    default: Project ERROR: gstreamer-gl-1.0 development package not found

bionic64 just works.  So move to it :-)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -15,7 +15,7 @@ Vagrant.configure(2) do |config|
   config.vm.provider "virtualbox"
   config.vm.provider "vmware_fusion"
 
-  config.vm.box = "ubuntu/xenial64"
+  config.vm.box = "ubuntu/bionic64"
   config.vm.provider :docker do |docker, override|
     override.vm.box = "tknerr/baseimage-ubuntu-16.04"
   end


DonLakeFlyer:
Xenial is our out of the box supported platform though. So it's best to match what CI is using. @andrewvoznytsa Any ideas here?

andrewvoznytsa:
Xenial does not work for us - we need bionic (and it is just because
gstreamer-gl is not supported by xenial, see
https://packages.ubuntu.com/search?keywords=libgstreamer-gl1.0-0
). Gstreamer GL is critical for qmlglsink support. CI uses bionic, BTW

And keep in mind that we have high chances to move to Ubuntu 20.04 in 2
months - gstreamer 1.14 (latest for bionic) has some bug (see
https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/issues/700)
 which was fixed in gstreamer 1.16 (will be available in 20.04).

It is possible to resolve xenial support issue building missing packages on
our side. But we need very serious reason for that - it will require some
noticeable efforts.

DonLakeFlyer:
Crap, my bad. I'm not up on Linux much. CI is bionic not xenial. So this is vagrant thing is fine.

@andrewvoznytsa When you say Xenial won't work. Do you mean you can't build on Xenial or you can't run on Xenial?

andrewvoznytsa:
I never really tested on xenial - I noticed that one of required build
dependencies is missing there and after short consulting with other guys it
was decided to move to bionic.

I think with some effort QGC (video stuff) can build and run on xenial.

andrewvoznytsa:
From our perspective xenial support will probably mean shipping (and
maintaining) gstreamer development and run time files for QGC developers.

DonLakeFlyer:
Ok, that's a new thing I didn't know about. Is it possible to get the build to automatically disable video when building on Xenial?

andrewvoznytsa:
Sure, disabling video support if some components are missing or outdated
would be helpful - we can move to new LTS in very smooth manner.

Issue #8398, opened 2020-02-23
Settings: Better system to hide groups when all settings are invisible


DonLakeFlyer:
This is a real pain right now. Also there are many groups in there which still don't have hide support coded on

Antiheavy:
It would also be nice to be able to control which application settings are revealed in Advanced Mode.

Issue #8396, opened 2020-02-23
Survey Presets do not work with "Manual (no camera specs)"


Antiheavy:
Survey Presets do not work with "Manual (no camera specs)" selected.  You can save the name, but when you select the Preset nothing changes with the mission.  It seems like you should be able to set up a Preset that used manual distance and spacing.  We would like to for our use case anyway.
![image](https://user-images.githubusercontent.com/12117027/75101755-5b6d2380-55a7-11ea-9689-b15f701a0124.png)
![image](https://user-images.githubusercontent.com/12117027/75101759-6cb63000-55a7-11ea-9321-8506e5275ec9.png)

Antiheavy:
Looks like https://github.com/mavlink/qgroundcontrol/pull/8400#event-3066636366 fixed this.  thanks!

PR #8400, opened 2020-02-23
Fix CameraCalc loading

DonLakeFlyer:
* Call was made to isManualCamera before the camera name was loaded. Causing it to base decisions on previous camera name. Which would in turn cause Manual camera Preset load to fail.
* Added Internal Error message display for preset load failures.
* Fix for #8398


Antiheavy:
Maybe you mean fix for: https://github.com/mavlink/qgroundcontrol/issues/8396

Issue #8347, opened 2020-02-16
Request: eliminate need for [Apply Preset] button in Surveys


Antiheavy:
The [Apply Preset] button is awkward and inconsistent with other QGC drop down selectors. 
![image](https://user-images.githubusercontent.com/12117027/74611573-b8a63800-50c2-11ea-887f-a023a68eb71e.png)

Most other drop down selectors in QGC operate by changing immediately to the item selected.  This is the most intuitive method and most drop down selectors in other applications generally work this way.

I think most people expect drop down selectors to do 2 things:
a) work as selector buttons
b) indicate current state

The current Presets drop down doesn't do either of the above.  The fact that it may not show the correct state is the most concerning to us (our users will be upset when the vehicle flies the wrong altitude and overlap, even though the Preset words were shown right in their face on screen).

Suggestions:
![image](https://user-images.githubusercontent.com/12117027/74611800-06bc3b00-50c5-11ea-9898-b6e4376b6943.png)

1) Preset drop down could initially show "_select a preset_" in italics or somehow to indicate nothing has been selected. (there are probably other methods that can work here too)
2) allow the drop-down menu to apply the settings when picked.  This way the preset drop down also shows current state.
3) if someone tabs away and then back, it should revert back to "_select a preset_" since you don't know what the user has changed in the mean-time.  Then if a drop down is selected it can be trusted to be showing the correct state again.
4) remove the [Apply Preset] button
5) suggestion: a simpler [New Preset] button should be enough to get the point across, especially since the saving/naming dialog provides extra context:
![image](https://user-images.githubusercontent.com/12117027/74611812-189dde00-50c5-11ea-8cdf-b927cd8d464b.png)

This should apply to the Survey Wizard Preset selector the same way.

Antiheavy:
Examples of other drop-down selectors in QGC that work the way I'm requesting here:

Camera selector:
![image](https://user-images.githubusercontent.com/12117027/74894388-73009e00-5354-11ea-82d3-49ba57654f21.png)

Camera settings in Mission Start pane:
![image](https://user-images.githubusercontent.com/12117027/74894490-bce98400-5354-11ea-97ee-837c867c95f2.png)

![image](https://user-images.githubusercontent.com/12117027/74894477-b5c27600-5354-11ea-9f64-cb3054e98795.png)

Flight Mode selector:
[...]

PR #8403, opened 2020-02-25
Improve survey presets drop down

dakejahl:
This PR implements the suggested improvements to the Survey Presets UI as discussed in 
https://github.com/mavlink/qgroundcontrol/issues/8347

survey_presets

And the wizard as well
survey_presets_wizard

### DIFF
--- a/src/MissionManager/ComplexMissionItem.cc
+++ b/src/MissionManager/ComplexMissionItem.cc
@@ -37,7 +37,11 @@ QStringList ComplexMissionItem::presetNames(void)
 
     settings.beginGroup(presetsSettingsGroup());
     settings.beginGroup(_presetSettingsKey);
-    return settings.childKeys();
+
+    names << "Select a preset";
+    names += settings.childKeys();
+
+    return names;
 }
 
 void ComplexMissionItem::loadPreset(const QString& name)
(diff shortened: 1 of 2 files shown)


src/MissionManager/ComplexMissionItem.cc
```diff
@@ -37,7 +37,11 @@ QStringList ComplexMissionItem::presetNames(void)
 
     settings.beginGroup(presetsSettingsGroup());
     settings.beginGroup(_presetSettingsKey);
-    return settings.childKeys();
+
+    names << "Select a preset";
```
Antiheavy: I wonder if it would add some clarity if this text was _italic_?

dakejahl: Yeah I wasn't quite sure how to do it, I found one way but it was a little bit hacky feeling.

Antiheavy:
This is working awesome.  It is more intuitive in my opinion.

@dakejahl  There is still a problem where, staying inside the Survey mission item the whole time, you can select a Preset, switch to the Grid tab, change some settings, and then go back to the Presets tab and it still shows the Preset you selected (which is now no longer true because you might have changed some things).  The drop down should revert back to "Select a preset" any time you switch into that tab (because you don't know what the user might have changed while away).

The button sizing could use some tweaking.  There is some awkward spacing as it stand now.  I suggest side-by-side buttons for New Preset and Delete Preset:
image

DonLakeFlyer:
> New Preset

That is way bad as well. The current button text helps explain to the user what a Preset is. Whereas "New" has completely different connotations as to what it may be doing. 

I don't see any of this as being better than what is there now. My reasons in the discussion issue. This lowers the usability of the product to remove one click.

PR #8409, opened 2020-02-25
Preset work: Allow custom build to specify which presets are built in

DonLakeFlyer:
* Built in presets cannot be deleted. Not pretty but simple enough for this late in the game with Stable.
* Deleting a preset pops an Are you sure? dialog


DonLakeFlyer:
@dakejahl FYI

PR #8411, opened 2020-02-25
Move Video Settings to the Video Element

tcanabrava:
This is for QGC 4.1+, not to be merged now.

with the idea of the multiple video streams we need to start moving some code around to be able to do the jump later on. As any video will have a different setting, we can't rely on the settings panel anymore for it, it should be tied to the video.
The QGCPipable now has a button for configuring the video:
image

When the button is clicked, the setting appears on the middle of the screen:
image

This will not enable multiple settings per video, but will facilitate when we get there.


src/VideoStreaming/VideoReceiver.h
```diff
@@ -152,4 +152,3 @@ protected slots:
     bool            _showFullScreen;
     VideoSettings*  _videoSettings;
```
andrewvoznytsa: According to our discussion _videoSettings should not be member of VideoReceiver - this is the only one path to make VideoReceiver a bit lighter and standalone as first step to make simple test application for CI and video development.

dogmaphobic:
Note that this is only valid for "dumb" cameras. If using a camera that properly implements the camera/video API, most of these settings aren't applicable.

tcanabrava:
The ones that are not applicable will behave the same way as the old
settings one.

tcanabrava:
Just some updates on the *why* I started this:
qgc3

This is the current configuration part for the Video / Recording.
Labels will not appear completely
Impossible to read and understand what's going on (now imagine this with multiple videos...)

My fix for that is to move some of the controls to the PIP - the ones that should not be behind a configuration - wall, like 'start / stop' recording, 'start / stop' video:

qgc1

And for the folder / filename of the video being recorded, this goes to the Configuration pane, easily accessed via the gear icon on the video itself.
qtc2


Some cameras can be correctly configured - but that's just related to streaming, not related to video recording.

tcanabrava:
(I used wrong icons in the bar below because I did not had correct images for them, it's `start/pause`, `show grid`, `record/stop recording`, `open configuration`.

src/VideoStreaming/VideoReceiver.h
tcanabrava: resolved by  #8455

tcanabrava:
@dogmaphobic - updated & rebased against master.

DonLakeFlyer:
Closing as abandoned

tcanabrava:
The code was approved by people almost a year ago, this is not abandoned.
It’s quite frustrating to have lots of code ready for merge and mass closed
as abandoned because the maintainers didn’t bother to look.

PR #8420, opened 2020-02-25
QML: Make an exact comparison

muramura:
I think it's better to do a type-matched comparison.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMAirframeComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMAirframeComponent.qml
@@ -109,7 +109,7 @@ SetupPage {
                             anchors.fill: parent
                             onClicked: {
                                 if (!airframeCheckBox.checked || !combo.valid) {
-                                    if (_frameTypeAvailable && object.defaultFrameType != -1) {
+                                    if (_frameTypeAvailable && object.defaultFrameType !== -1) {
                                         _frameType.rawValue = object.defaultFrameType
                                     }
                                     airframeCheckBox.checked = true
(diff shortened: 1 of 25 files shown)


DonLakeFlyer:
> I think it's better to do a type-matched comparison.

Why? As far as I'm concerned this is a dangerous change which could lead to subtle bugs. The code wants to check that two values are equal not that there types may actually differ. In a fixed type language this sort of thing makes sense. But in a variable types language like Qml script it doesn't as much. Making this change would require checking that every single fact type (which are variably typed) matches the type of the variable in the script you are checking against. For example there can be cases where a parameter Fact is a double. But there are fixed values which are ints which may be tested against. As far as I understand strict comparison will now fail that check. Finding those cases throughout the could be a nightmare to figure out when strict comparison is right and when it is wrong.

dogmaphobic:
> As far as I understand strict comparison will now fail that check.

This is important to confirm (I suspect it would fail as well). There are a ton of cases where the value within the _Fact_ is a float but used everywhere in QML as simple integers.

tcanabrava:
> > As far as I understand strict comparison will now fail that check.
> 
> This is important to confirm (I suspect it would fail as well). There are a ton of cases where the value within the _Fact_ is a float but used everywhere in QML as simple integers.

I actually got this bug yesterday when I rewrote a bit of code that returned fact.rawvalue = 0 but the type in the fact is a bool.

DonLakeFlyer:
> I actually got this bug yesterday when I rewrote a bit of code that returned fact.rawvalue = 0 but the type in the fact is a bool.

Yeah, that's a similar problem and the reason why this is dangerous. It's too easy to screw this up in script code and the fact that it converts in most cases can save your ass. In reality this script side no equals what happens on the C++ side. Where the Fact variant will automatically by converted to the type you are comparing against. I'm going to close unless someone has a better reason to do this.

muramura:
@DonLakeFlyer san. Comments thanks.

Issue #8415, opened 2020-02-25
Loading KML and SHP files moves map to 0,0 Africa


Antiheavy:
In recent Daily builds whenever we load a KML or SHP file for a Survey the map view flies to lat,lon 0,0 off the coast of Africa.  The KML and SHP "seems" to load correctly at the proper location.  This happens on both KML and SHP files which are known to work fine with older versions of QGC. 

![image](https://user-images.githubusercontent.com/12117027/75285216-d6b71b00-57db-11ea-86aa-791484c1b874.png)

DonLakeFlyer:
Can you link to one that breaks. Then ones I"m testing are working.

Antiheavy:
[Joe 160.zip](https://github.com/mavlink/qgroundcontrol/files/4252942/Joe.160.zip)

[Farm Demo.zip](https://github.com/mavlink/qgroundcontrol/files/4252945/Farm.Demo.zip)

Attached "Joe160" SHP and "Farm Demo" KML.  Both of these work normally with older QGC version.  

To reproduce (I've done this on 3 different Windows machines now):
1) Plan View
2) File -> Survey
3) Click [Done Adjusting] for Takeoff
3) Click ]Load KML/SHP...]
4) Browse to file, click Open
5) Observe map pans to 0,0 off coast of Africa
6) Observe the Survey is correctly created according to the KML/SHP

PR #8424, opened 2020-02-26
Map Fit: Skip invalid coords

DonLakeFlyer:
* When an item is in an intermediate state (wizard not complete) the coordinate for it may be NaN. Which was confusing the map fit functions causing the map to zoom fully out and move to 0,0 to fit it.
* Fix for #8415


Antiheavy:
yep.  working in Stable_v4.0 build.  Not working in today's Daily build, but maybe that is expected...

DonLakeFlyer:
Daily build is going to fall behind a little while I fight with Stable.

PR #8445, opened 2020-02-29
Compiler switch for Survey Preset JSON generation

dakejahl:
This allows developers to optionally save a Survey Preset as a JSON file when selecting "Save preset" in the Survey Plan. 

This JSON files end up in the "Missions" directory, where you can just copy/paste them into your source tree and include them in your custom build.

```
<qresource prefix="/foo">
        <file alias="bar.json">patch/bar.json</file>
</qresource>
```

The from C++ file load code is `":/foo/bar.json"`

### DIFF
--- a/src/MissionManager/ComplexMissionItem.h
+++ b/src/MissionManager/ComplexMissionItem.h
@@ -86,8 +86,6 @@ protected:
     void        _savePresetJson (const QString& name, QJsonObject& presetObject);
     QJsonObject _loadPresetJson (const QString& name);
 
-    void        _saveSettingsValueAsJson(const QSettings& settings, const QString& name);
-
     bool _isIncomplete = true;
 
     QMap<QString, FactMetaData*> _metaDataMap;
--- a/src/MissionManager/ComplexMissionItem.h
+++ b/src/MissionManager/ComplexMissionItem.h
@@ -15,6 +15,10 @@
 
 #include <QSettings>
 
+#include <QGCToolbox.h>
+#include <SettingsManager.h>
+
+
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Why go back and forth through the QSettings object from binary and so forth. Much simpler would be to just breakout QJsonDocument(presetObject) into a variable right in _savePresetJson. And then stuff everything in the #if 0 in _savePresetJson instead of breaking out into a separate method. The separate method seems to indicated that it's useful for something other than a one-off build which it is not. Don't want anyone using that for something else.

DonLakeFlyer:
I know I'm a pain. I'm super focused on readability of the code base for anywone walking up to it.

DonLakeFlyer:
@Thanks @dakejahl

dakejahl:
It make sense, I really appreciate good feedback!

src/MissionManager/ComplexMissionItem.h
```diff
@@ -15,6 +15,10 @@
 
 #include <QSettings>
 
+#include <QGCToolbox.h>
```
DonLakeFlyer: @dakejahl  I noticed this after I merged. QGC includes should be "foo.h", only Qt includes are "<foo>". Also you should keep the Qt and QGC include blocks together. So in this case the include you added should have gone right underneath the other existing QGC include at the top of the file without an extra newlines. Make sense?

I've fixed it up in my terrain pull which merged conflicted in these file which is how I notied.

dakejahl: Yup makes sense, I was not aware of the convention,

PR #8446, opened 2020-02-29
Preflight checklist when no vehicle is connected

dakejahl:
Rather than using the DefaultChecklist when no vehicle is connected, we can first check to see if we have an idea about which vehicle type is expected by first checking **QGroundControl.settingsManager.appSettings.offlineEditingVehicleType**

### DIFF
--- a/src/FlightDisplay/PreFlightCheckList.qml
+++ b/src/FlightDisplay/PreFlightCheckList.qml
@@ -23,9 +23,7 @@ Rectangle {
     color:      qgcPal.windowShade
     radius:     3
 
-    property real   _verticalMargin:                ScreenTools.defaultFontPixelHeight / 2
-    property var    _offlineEditingVehicleType:     QGroundControl.settingsManager.appSettings.offlineEditingVehicleType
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
This ends up being a little wierd from a code duplication standpoint. You should be able to do something like this:
```
var vehicle = activeVehicle
if (!vehicle) {
   vehicle = QGroundControl.multiVehicleManager.offlineEditingVehicle
}
```

And then fall through to the existing code tweaked to use vehicle instead of activeVehicle. Does that make sense?

DonLakeFlyer:
Perfect thanks

dakejahl:
Yes that's a much simpler way to do it, I didn't realize the multi vehicle manager exposed that

dakejahl:
Should I open a PR into master for this as well or does v4.0 get merged
back in to master?

DonLakeFlyer:
I'll merge Stable_V4.0 across to master today sometime. There a pile of stuff in there that needs to come across.

PR #8449, opened 2020-02-29
Mavlink terrain protocol support

DonLakeFlyer:
* Currently only works with ArduPilot. PX4 firmware side support should be coming soon as well.
* Not yet supported by Patterns. That will come soon in another pull.
* I've done a bunch of testing myself flying over terrain, but best be careful when using this since it's a big pile of new code.

If you are using a firmware which supports it. You will now see the additional "Terrain Frame" option for altitudes.
Screen Shot 2020-02-29 at 3 45 48 PM

Fly view has a new indicator which shows the progress of uploading terrain heights to the vehicle.
Screen Shot 2020-02-29 at 10 23 54 AM

When all the terrain info the vehicle asked for has been uploaded you'll see it say Done. Then after 30 seconds it will disappear to get out of the way. if the Vehicle asks for more terrain data it will come back.
Screen Shot 2020-02-29 at 10 35 02 AM

So create Plans using Terrain Frame, wait for the data to upload to the vehicle and you can fly the mission which follows terrain and do thing like Terrain based RTL and so forth.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,9 +2,13 @@
 
 Note: This file only contains high level features or important fixes.
 
+## 4.1 - Daily Build
+
+* Support mavlink terrain protocol which queries gcs for terrain height information. Allows planning missions with TERRAIN\_FRAME.
+
 ## 4.0
 
-### 4.0.0 - Daily Build
+### 4.0.0 - Stable
 
 * Added ROI option during manual flight.
 * Windows: Move builds to 64 bit, Qt 5.12.5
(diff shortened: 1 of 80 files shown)


DonLakeFlyer:
Screen Shot 2020-02-29 at 3 58 50 PM

Note that it's a WIP to figure out what to do with the two terrain options. The old one which stuffs AMSL value direction into the waypoint altitude. And the new one where terrain data is uploaded to the vehicle and the vehicle handles it. Leaning towards it being an either/or option. If the vehicle supports mavlink terrain protocol then only show Terrain Frame (likely named something else) and hide the other one. And the reverse if the vehicle does not support the terrain protocol.

Knowing whether there is vehicle support is driven by the MAV_PROTOCOL_CAPABILITY_TERRAIN capability bit.

PR #8456, opened 2020-03-02
checks for gst.h file

BishalSingha:
The **build** fails because the package "gstream" was not found. This should be the required package and if not found then **cmake** should raise an error. This PR fixes that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,9 +85,10 @@ endif()
 #
 find_package(PkgConfig)
 pkg_check_modules(GST
+	REQUIRED
     gstreamer-1.0>=1.14
     gstreamer-video-1.0>=1.14
-    gstreamer-gl-1.0>=1.14
+    gstreamer-gl-1.0>=1.14	
 )
 
 if (GST_FOUND)


CMakeLists.txt
```diff
@@ -85,9 +85,10 @@ endif()
 #
 find_package(PkgConfig)
 pkg_check_modules(GST
+	REQUIRED
```
andrewvoznytsa: Actually gstreamer-1.0 is not required package - it is optional. If gstreamer and related packages are not found then QGC should build without video streaming support (and it is how it builds using qmake, see .pro/.pri files). If you want to get it merged you need to keep it working properly - make gstreamer optional and compile with video disabled if related packages were not found.

CMakeLists.txt
```diff
@@ -85,9 +85,10 @@ endif()
...
     gstreamer-1.0>=1.14
     gstreamer-video-1.0>=1.14
-    gstreamer-gl-1.0>=1.14
+    gstreamer-gl-1.0>=1.14
```
andrewvoznytsa: Thanks for that finding - we missed that. But could you please remove extra spaces at the end of line?

jaxxzer:
It looks like this should be closed until the build with video streaming is handled automatically according to whether or not the package is found.

PR #8472, opened 2020-03-04
Added missing files into Vehicle's cmake file

alexeylysenko:


alexeylysenko:
It's been fixed by @tcanabrava in #8534

PR #8473, opened 2020-03-04
Added parent to the new FactMetaData objects to be owned by parent

alexeylysenko:
Otherwise it leads to memory leakage, no one is responsible to destroy FactMetaData object.


alexeylysenko:
@dagar @dogmaphobic could you please review this PR? Does it make sense?

src/FirmwarePlugin/PX4/PX4ParameterMetaData.cc
```diff
@@ -214,7 +214,7 @@ void PX4ParameterMetaData::loadParameterFactMetaDataFile(const QString& metaData
                 
                 // Now that we know type we can create meta data object and add it to the system
                 
-                metaData = new FactMetaData(foundType);
+                metaData = new FactMetaData(foundType, this);
```
DonLakeFlyer: This is correct. Can you also remove the Q_CHECK_PTR under this since it does nothing. Can could you also check the ArduPilot variant of this same routine to see if it has the same problem.

alexeylysenko: Yup, it has a memory leak as well
`rawMetaData = new APMFactMetaDataRaw();` 
but the situation is a bit different.
APMFactMetaDataRaw is not QObject-derived class, so we can't use parent to manage object's lifetime.

As a suggestion we may use smart pointer instead of raw one. For example,
`typedef QMap<QString, std::shared_ptr<APMFactMetaDataRaw>> ParameterNametoFactMetaDataMap;`
What's your opinion?

IMO it'd be an overkill to just derive from QObject to solve the memory leakage.

DonLakeFlyer: > IMO it'd be an overkill to just derive from QObject to solve the memory leakage.

Sort of. But since QObject parentage mechanism is used throughout the QGC codebase, whereas the shared pointer pattern is used nowhere. Then QObject parentage is a better choice for codebase consistency.

DonLakeFlyer: Also the parent for APMFactMetaDataRow which is APMParameterMetaData is a QObject.

alexeylysenko: Implemented as you proposed (APMFactMetaDataRow is QObject now)

DonLakeFlyer:
Perfect, thanks so much

PR #8483, opened 2020-03-05
fixed bug: the highlight mode is not the actually mode

qter188:
fixed bug: the highlight mode is not the actually mode


cuhome:
image
Wrong pattern mapping

DonLakeFlyer:
Thanks for the fix

Issue #8478, opened 2020-03-04
Waypoint numbers optional


sanderux:
Although the view is much cleaner, external validators and phone support on mission plans makes the lack of waypoint numbers very impractical.

I propose an on/off switch, setting or showing the numbers in the waypoint list.

Saijin-Naib:
Agree with this. I like being able to validate the order without having to open the plan in a text editor.

DonLakeFlyer:
> validate the order

@Saijin-Naib Can you explain some more what you mean by that. Why do you need to validate the order?

Saijin-Naib:
OCD-ness, not trusting the generated plans, ensuring that the transects flow the way I expect them to (especially important on cross-grid), etc. etc. etc.

I'd really love a means to inspect each individual waypoint/command generated for a Survey as well to validate it's properties...

DonLakeFlyer:
Got it. Thanks for the info.

DonLakeFlyer:
> I'd really love a means to inspect each individual waypoint/command generated for a Survey as well to validate it's properties...

FYI: You can get that if you download the Plan back from the vehicle after you upload. Then it will degrade to individual commands.

Saijin-Naib:
> FYI: You can get that if you download the Plan back from the vehicle after you upload. Then it will degrade to individual commands.

Thanks for the tip/work-around. That's very valuable. An in-app way to do the same would likely be easier, but this suits for now.

DonLakeFlyer:
Also for folks trying to deal with no waypoints numbers. You can see waypoints from the Terrain display (Click T at bottom of Plan). Those item are also clickable to set current item.

Antiheavy:
We had some of our users start trialing the newest QGC Stable this week and the first peice of feedback was "where did the waypoint numbers go? can I enable them somehow? we use those to communicate to eachother about where the aircraft is."

I was planning on seeing if people would even notice the numbers where gone... turns out they noticed right away.  ha!

I think a Fly View Settings option would be pretty useful for this.

DonLakeFlyer:
> we use those to communicate to each other about where the aircraft is.

That's a major problem which I'm getting feedback on from multiple places. 

At this point I even wonder whether a show.hide thing is feasible with default to off. If majority of people end up turning that on, then it's a failure.

sanderux:
@DonLakeFlyer 
[...]

PR #8494, opened 2020-03-05
Plan: Default for terrain display is visible

DonLakeFlyer:
* Hoping this will cover some of these case for waypoint numbering
* The setting is remembered, so you can hide and it will stay hidden

Related to #8478


Antiheavy:
this is good.  we were going to ask for this.  it is a good safety feature.  sounds like others prefer to have it shown too.

PR #8497, opened 2020-03-05
Much better VTOL support [Do Not Merge]

DonLakeFlyer:
VTOL support was in a state of very unhelpful mess.

Screen Shot 2020-03-05 at 12 55 42 PM

Screenshot shows multiple changes, some specific to VTOL, some generic:
* VTOL Takeoff now works like Fixed Wing takeoff in that it positions the Takeoff item away from the Launch location. And it asks the user to futz both, warning them about wind and so forth.
* The command description is now shown all the time even when in incomplete/wizard mode. This helps explain a bit about the move this thing to here, move that thing to there.
* A number of the command description help texts were updated to better reflect what they do and or make more sense with respect to the QGC planning ui terminology.
* A number of commands were using the word "Heading" for labels when they should have been "Yaw".
* The Tooltrip Return/Land button now says "Land" for a VTOL and when you click it a VTOL_LAND command is added.
* Fix some bugs where the family of land commands did not have their altitude correctly defaulted to 0.
* Land commands have an altitude value associated with them. Differing from almost all other commands this altitude is the altitude of the ground. Which can be quite confusion. So in those cases be a little more wordy on what the altitude is actually for.

Screen Shot 2020-03-05 at 1 03 17 PM

### DIFF
--- a/src/FirmwarePlugin/PX4/MavCmdInfoMultiRotor.json
+++ b/src/FirmwarePlugin/PX4/MavCmdInfoMultiRotor.json
@@ -26,7 +26,7 @@
             "id":           22,
             "comment":      "MAV_CMD_NAV_TAKEOFF",
             "paramRemove":  "1",
-            "description":  "Hover straight up to specified altitude. Then travel to specified position.",
+            "description":  "Hover straight up to specified altitude. Then travel to specified position."
         }
     ]
 }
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
This has a "Do Not Merge" on it because I head out of town in two weeks and will be gone for a little over a month. The changes here are not major but they are also not minor to the point of not being worried about dumping into stable and then disappearing.

Instead of waiting I put the pull up now to give VTOL follks a heads up on changes.

DonLakeFlyer:
Also I moved VTOL_TAKEOFF and VTOL_LAND into the "Basic" category for the command picker.

DonLakeFlyer:
In looking back at this although I'd really like to get this into a Stable 4.0.* somewhere it's just way too much to got into a point release. I'm going to move it to master and put it through.

PR #8509, opened 2020-03-06
Clean up code used only on custom builds.

dogmaphobic:


DonLakeFlyer:
Should be in Stable_V4.0 as well I think. Not sure if it really causes a problem or just annoying?

dogmaphobic:
Just annoying. The page is not really useful until someone edits it to contain a proper VTOL checklist.

PR #8511, opened 2020-03-06
Mobile file dialog: Handle multiple extensions correctly

DonLakeFlyer:
Loading a shape file would fail since it would add the .kml file extension to the file


Antiheavy:
tested to be working correctly in our build.  thanks.

PR #8517, opened 2020-03-07
fixed sensors check when no vehicle connected

dakejahl:
It's obvious


DonLakeFlyer:
This needs to go into Stable

DonLakeFlyer:
No need to close, just another pull for Stable as well.

PR #8519, opened 2020-03-08
Audio: Speak last buffered message

muramura:
I hear a buffered message for a long time.
I think the last message is meaningful.
I decided to speak the last buffered message.


DonLakeFlyer:
I don't see how throwing out messages is a good idea? There could be something important you miss. Who's to say the last message is more important than any of the others?

DonLakeFlyer:
@muramura Do you agree that what I said is correct? Or is there something more to it?

DonLakeFlyer:
Closing unless someone can explain how this isn't the wrong thing to do

PR #8523, opened 2020-03-08
Plan: Much better error reporting for mission transfer errors

DonLakeFlyer:
* Now includes item sequence number as well as command name
* Better structuring of the error string information for readability
* Create unit test to test the error string output

Screen Shot 2020-03-08 at 11 26 34 AM

Above is an example of the current sate of mission transfer error reporting. It is pretty much useless. No information on which item this applies to. No clue as to what the command is that is causing the problem. And that's just a start.


DonLakeFlyer:
This is just a start to work from. End goal is to show error right in the corresponding item.

PR #8536, opened 2020-03-10
Plan: CONDITION_GATE command support

DonLakeFlyer:
Problem:
If you start/stop camera triggering in each transect at the survey edges you do not get precise coverage of the survey area. This is due to waypoint acceptance triggering prior to the vehicle actually getting to the wp position. So for survey edge entry you tend to get images prior to the survey edge. And for survey edge exit you will miss images which go all the way out to the edge.

The new CONDITION_GATE command works just like a wp from a standpoint of the vehicle flying to it. But it delays the mission state machine until the vehicle flies through an imaginary line which is orthogonal to it and the previous waypoint. If you put this in front of a camera trigger then it means the triggering will occur at the gate position instead of before it. Hence the triggering will start/stop right at the survey edges.

The goal is to no longer need to capture images throughout the survey including the turnarounds to get correct coverage of the survey area.

Note: This requires firmware support for MAV_CMD_CONDITION_GATE which is currently a WIP.

Changes:
* Added MAV_CMD_CONDITION_GATE the the ui info command tree
* Updated MAV_CMD_DO_SET_CAM_TRIGG_DIST to include ui for param2:shutter and param3:trigger
* Added setting to turn on usage on CONDITION_GATE in Settings/General/Plan View section (defaults to off). This is a temporary mechanism into a fancier vehicle connection based firmware version check can be implemented.
* Survey item will use a CONDITION_GATE instead of a NAV_WAYPOINT prior to a camera trigger start/stop.
* The trigger stop at survey edge exit now sets MAV_CMD_DO_SET_CAM_TRIGG_DIST.param2=1 to trigger an immediate image as well as stopping distance triggering. This way it gets correct image coverge of the exit edge of the survey area.
* Complete rewrite of Survey command sequence generation both to support CONDITION_GATE and also to make it more understandale to folks who haven't been staring at it for months.
* New comprehensive unit test for survey command sequence generation.

Remaining work:
* CONDITION_GATE support for Corridor Scan. Including similar rewrite for readability.
* The CONDITION_GATE item editor should slurp up camera and flight speed commands like NAV_WAYPOINT does. Since it is essentially a variation on a waypoint.
* Since camera trigger off not sets MAV_CMD_DO_SET_CAM_TRIGG_DIST.param2=1 to trigger immediately even when stopping distance triggering. Those items are not sucked into the camera section of the previous wp since they don't match the criteria. They will show up as individual commands.

Replacement for #8497


LorenzMeier:
PX4 Implementation: https://github.com/PX4/Firmware/pull/11878

Issue #8489, opened 2020-03-05
4.0 video streaming high latency


jaxxzer:
The introduction of `rtpjitterbuffer` in the pipeline has significantly increased the latency on the video streaming.

In controlled tests (ethernet connection, udp, h264 1080p 30fps):

QGC 3.5: 202ms
QGC 4.0: 389ms

jaxxzer:
https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoStreaming/VideoReceiver.cc#L234
https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoStreaming/VideoReceiver.cc#L383

PR #8551, opened 2020-03-11
Add Low latency mode for video

Williangalvani:
Screenshot from 2020-03-11 21-01-19

This setting disables the rtpjitterbuffer and sets the videosink to non-sync (assynchronous? mode. This reduces the latency here from ~380ms to ~200ms.

@jaxxzer can you please test?

### DIFF
--- a/src/VideoStreaming/VideoReceiver.cc
+++ b/src/VideoStreaming/VideoReceiver.cc
@@ -395,7 +395,7 @@ VideoReceiver::_makeSource(const QString& uri)
         gst_element_foreach_src_pad(source, _padProbe, &probeRes);
 
         if (probeRes & 1) {
-            if (probeRes & 2) {
+            if (probeRes & 2 && !_videoSettings->lowLatencyMode()->rawValue().toBool()) {
                 if ((buffer = gst_element_factory_make("rtpjitterbuffer", nullptr)) == nullptr) {
                     qCCritical(VideoReceiverLog) << "gst_element_factory_make('rtpjitterbuffer') failed";
                     break;
@@ -414,7 +414,11 @@ VideoReceiver::_makeSource(const QString& uri)
                 }
             }
         } else {
-            g_signal_connect(source, "pad-added", G_CALLBACK(_linkPadWithOptionalBuffer), parser);
+            if (_videoSettings->lowLatencyMode()->rawValue().toBool()) {
+                g_signal_connect(source, "pad-added", G_CALLBACK(newPadCB), parser);
+            } else {
+                g_signal_connect(source, "pad-added", G_CALLBACK(_linkPadWithOptionalBuffer), parser);
+            }
         }
 
         g_signal_connect(parser, "pad-added", G_CALLBACK(_wrapWithGhostPad), nullptr);
@@ -568,6 +572,9 @@ VideoReceiver::start()
         qCWarning(VideoReceiverLog) << "Failed because video sink is not set";
         return;
     }
+
+    g_object_set(_videoSink, "sync", !_videoSettings->lowLatencyMode()->rawValue().toBool(), NULL);
+
     if(_running) {
         qCDebug(VideoReceiverLog) << "Already running!";
         return;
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
This needs to be a setting? It kinds of reads like "check this box to make your video not suck"!

Williangalvani:
> This needs to be a setting? It kinds of reads like "check this box to make your video not suck"!

@andrewvoznytsa pointed out that some encoders can send frames out of orders, which the sync takes care of.
The rptjitterbuffer is useful for people with 4G links which seem to enjoy sending packets out of order.

DonLakeFlyer:
There is no way to autodetect that somehow? if not, then we need some good way to explain this completely techy thing to a user

@Williangalvani Can you screenshot the new setting?

@hamishwillee Getting hamish involved...

Williangalvani:
> There is no way to autodetect that somehow?

We'll have to ask @andrewvoznytsa. Maybe based on the amount of bad frames decoded? I'd still prefer some artifacts over a high latency video.

Making it low-latency-only or making it the default both sound ok to me, I was just being cautious as it could harm other users I'm not aware of.

jaxxzer:
This will fix #8489 . The setting is working over here, no app restart is necessary.

In controlled tests (ethernet connection, udp, h264 1080p 30fps):

QGC 3.5: 202ms
QGC 4.0: 389ms

After this patch:
box unchecked: 393ms
box checked: 205ms

What this does is enable/disable an internal buffer that will wait for missing/out-of-order packets for 200ms. Then it reorders the packets for a nice display. This is very good for wireless connections, but it adds up to 200ms latency to the video stream, which is quite noticeable to a pilot. If the internal frame buffer is disabled, the video stream may have more tearing/breaking and lower framerate (especially for a poor wireless connection), but the 200ms latency overhead of the internal frame buffer is eliminated.

I suggest calling the setting 'disable (200ms) frame buffer' or something along those lines.
 @Williangalvani and I can help @hamishwillee with the docs.

DonLakeFlyer:
Ok, so I'm maybe fingers crossed at a point with Stable that this next point release may last more than a couple days. So I'd like to wait on it until we get some good wording here. Then I'll merge and get a point release out.

src/VideoStreaming/VideoReceiver.cc
```diff
@@ -395,7 +395,7 @@ VideoReceiver::_makeSource(const QString& uri)
         gst_element_foreach_src_pad(source, _padProbe, &probeRes);
 
         if (probeRes & 1) {
-            if (probeRes & 2) {
+            if (probeRes & 2 && !_videoSettings->lowLatencyMode()->rawValue().toBool()) {
```
andrewvoznytsa: There is also another place (https://github.com/mavlink/qgroundcontrol/blob/Stable_V4.0/src/VideoStreaming/VideoReceiver.cc#L238) where rtpjitterbuffer could be inserted. I think lowLatencyMode() option should be consistent across all cases - please add it.

andrewvoznytsa:
> There is no way to autodetect that somehow? if not, then we need some good way to explain this completely techy thing to a user

There are some ways to autodetect if this rtp jitter buffer is needed (in fact this is something that good jitter buffers do - they are able to adjust buffer length depending on QoS). But they are very risky to be added in the last moment and not really trivial to do. Our silver bullet for 4.0 is to explain that `if you see some video lag AND video quality is good then you can try to experiment with low latency check box - enabling low latency (disabling buffering) switches QGC into 3.x buffering mode. In this mode you can expect a bit choppy video and/or frame losing in some cases` - that's what user needs to know (feel free to re-phrase into more readable text).

Williangalvani:
Updated, thank you @andrewvoznytsa for the help.

patrickelectric:
Tipo: It's possible for QGC to inform the user about the video latency:
https://gstreamer.freedesktop.org/documentation/additional/design/tracing.html?gi-language=c

DonLakeFlyer:
Is this ready to merge?

Williangalvani:
I am running some tests here on a real vehicle.

DonLakeFlyer:
Thanks folks. I'll spit out a point release tomorrow.

PR #8554, opened 2020-03-13
Set correct parameter type for signal

alexeylysenko:
terrainCollision property is a boolean value, but signal's parameter is double - it has to be bool too.

### DIFF
--- a/src/MissionManager/VisualMissionItem.h
+++ b/src/MissionManager/VisualMissionItem.h
@@ -194,7 +194,7 @@ signals:
     void altDifferenceChanged           (double altDifference);
     void altPercentChanged              (double altPercent);
     void terrainPercentChanged          (double terrainPercent);
-    void terrainCollisionChanged        (double terrainCollision);
+    void terrainCollisionChanged        (bool terrainCollision);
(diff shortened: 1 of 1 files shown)


src/MissionManager/VisualMissionItem.h
```diff
@@ -194,7 +194,7 @@ class VisualMissionItem : public QObject
     void altDifferenceChanged           (double altDifference);
     void altPercentChanged              (double altPercent);
     void terrainPercentChanged          (double terrainPercent);
-    void terrainCollisionChanged        (double terrainCollision);
+    void terrainCollisionChanged        (bool terrainCollision);
```
DonLakeFlyer: How come this wasn't causing signal connect compiler failures. Is nobody using this no the C++ side? Wondering if there is some connect code somewhere.

alexeylysenko: I tried to find the connect() with the signal and had no luck. It's the reason why it compiles successfully.

DonLakeFlyer: OK, thanks. That means it's only used on the Qml side.

PR #8555, opened 2020-03-13
Replaced not equal operator for double values to qFuzzyCompare

alexeylysenko:
Using != operator is not a way to go for comparing double values. Moreover, qFuzzyCompare is already widely used in the file, so the fix would be good for keeping consistency too.


alexeylysenko:
@DonLakeFlyer Does it make sense?

DonLakeFlyer:
Yes and no. The reason being is that gimbal yaw is a user entered value. A fuzzy compare tends to important when real numbers are being put through some sort of calculation. And then the outputs of those calculations are compared. Also the gimbal yaw values the user can enter are limited to no decimal places. So theoretically this makes some amount of sense, but in reality it doesn't matter. The largest agrument is around consistency, which I do agree with.

PR #8556, opened 2020-03-13
Used isValid() to check whether geo coordinates have been properly set

alexeylysenko:
IMO isValid() is more accurate way to validate the geo coordinates than comparing longitude and latitude with 0. In addition, expression like double == int may not work as expected.

### DIFF
--- a/src/MissionManager/VisualMissionItem.cc
+++ b/src/MissionManager/VisualMissionItem.cc
@@ -162,7 +162,7 @@ void VisualMissionItem::setMissionVehicleYaw(double vehicleYaw)
 
 void VisualMissionItem::_updateTerrainAltitude(void)
 {
-    if (coordinate().latitude() == 0 && coordinate().longitude() == 0) {
+    if (!coordinate().isValid()) {
         // This is an intermediate state we don't react to
         return;
     }
--- a/src/MissionManager/VisualMissionItem.cc
+++ b/src/MissionManager/VisualMissionItem.cc
@@ -166,7 +166,7 @@ void VisualMissionItem::_updateTerrainAltitude(void)
         // This is an intermediate state we don't react to
         return;
     }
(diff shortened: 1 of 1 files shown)


src/MissionManager/VisualMissionItem.cc
```diff
@@ -162,11 +162,11 @@ void VisualMissionItem::setMissionVehicleYaw(double vehicleYaw)
 
 void VisualMissionItem::_updateTerrainAltitude(void)
 {
-    if (coordinate().latitude() == 0 && coordinate().longitude() == 0) {
```
DonLakeFlyer: This is incorrect. lat=0, lon=0 is a valid coordinate.

DonLakeFlyer: The other thing to know is that at times for various commands ArduPilot (but not PX4) can use lat=0, lon=0 to indicate a lack of position in a command. So you may see other cases of this.

alexeylysenko: True, but do we really want to check whether long and lat is zero (we want to skip processing "null island"?) Or it's just a wrong way to catch invalid coordinate.

DonLakeFlyer: In this specific case there was some transient state which was causing it. I don't quite remember the details. Without going back through commit history to try to find those details it's best to leave it alone and assume the semantics are correct. As opposed to changing it and possibly breaking something in a subtle way.

alexeylysenko: I see, is it worth to add a comment about that to avoid confusion?

DonLakeFlyer: Other than what is already there not sure what else is known to be added

alexeylysenko:
Ok, thanks for explanation

PR #8562, opened 2020-03-15
WIP interpret HTTP for RTSP

sanderux:
WIP DO NOT MERGE


sanderux:
Yes i realize that, this was just a pr to generate an apk to test.
current master has broken RTSP support

As for the redirect (should we ever need it) i disagree as the http url must be polled in the timeout loop as the RTSP URL might have changed

PR #8564, opened 2020-03-15
Excluded ANDROID OS from LINUX variable in cmake

alexeylysenko:
In cmake UNIX flag is set also for ANDROID OS. Since we have constructions like
```
if (LINUX)
elseif (ANDROID)
endif()
```
Android OS has to be excluded from LINUX flag, otherwise code under elseif will be unreachable.


alexeylysenko:
@dagar @tcanabrava can you confirm this?

patrickelectric:
Can we merge this ?

PR #8570, opened 2020-03-16
Sub: Add fact for roll/pitch toggle, remove spaces from fact names

### LINKED ISSUE #8360: Sub: Add Roll-Pitch toggle visual or sonorous feedback
> Would it be possible to create an additional speech announcement for when Roll-Pitch toggle is engaged/disengaged, along with a APMSubInfo style display similar to input hold.

From: https://discuss.bluerobotics.com/t/roll-pitch-toggle-speech-activation-indicator/6894

### LINKED ISSUE #8705: QGC crashes when trying to click APMSubInfo in Values widget
Open ardusub SITL and try to change de displayed values.
segfault

It seems to be trying to access the fact group "aPMSubInfo" instead of "APMSubInfo"

Williangalvani:
This will allow sub users to see if the ROV is in the Roll/pitch toggle
mode using the APMSubInfo section in the info widget

needs to wait for https://github.com/ArduPilot/ardupilot/pull/13630

Fix #8360 
Fix #8705

### DIFF
--- a/src/FirmwarePlugin/APM/ArduSubFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/ArduSubFirmwarePlugin.cc
@@ -135,7 +135,7 @@ ArduSubFirmwarePlugin::ArduSubFirmwarePlugin(void):
         _remapParamNameIntialized = true;
     }
 
-    _nameToFactGroupMap.insert("APMSubInfo", &_infoFactGroup);
+    _nameToFactGroupMap.insert("apmSubInfo", &_infoFactGroup);
 
     _factRenameMap[QStringLiteral("altitudeRelative")] = QStringLiteral("Depth");
     _factRenameMap[QStringLiteral("flightTime")] = QStringLiteral("Dive Time");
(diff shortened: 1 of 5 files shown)


jaxxzer:
@DonLakeFlyer can we get a merge?

src/FirmwarePlugin/APM/ArduSubFirmwarePlugin.cc
```diff
@@ -287,6 +289,7 @@ const char* APMSubmarineFactGroup::_lightsLevel1FactName        = "lights 1";
 const char* APMSubmarineFactGroup::_lightsLevel2FactName        = "lights 2";
 const char* APMSubmarineFactGroup::_pilotGainFactName           = "pilot gain";
```
DonLakeFlyer, line 290: So this is the first I'm noticing these. Although I think it will currently work I'd recommend against fact names with spaces in them. You could get hosed further down the road by some other change which can't handle spaces. Or if you fact name gets used in an unexpected ways like for example persistance or things like that. There aren't any places in the code where fact names have spaces.

Williangalvani: I just pushed another two commits taking spaces out of fact names

DonLakeFlyer:
Some of these are not camelCase "rollpitch...", "rangefinder..." maybe others. Not technically required. But common style is better.

Williangalvani:
@DonLakeFlyer update. "Rangefinder" apparently is just one word.
I renamed APMSubInfo to apmSubInfo too.

hamishwillee:
> "Rangefinder" apparently is just one word.

If you insist on two words, you can always use distance sensor :-)

Williangalvani:
Rebased to fix CI, can you take a look at this, @DonLakeFlyer ?

Issue #7791, opened 2019-09-08
Feature Request: add "current mission item" (or current mission index) to Value Widget list


Antiheavy:
Would be useful to see the active mission item/index in the values widget for certain situations.  Especially with the newer simplified Fly screen visuals, sometimes it is not obvious what the active mission item is.  Having this option in the Values Widget gives the advanced user an option to always see what is active, even if off screen or hidden or whatever.

![image](https://user-images.githubusercontent.com/12117027/64490519-41bb6800-d223-11e9-8ef0-6f4f0d89f929.png)

DonLakeFlyer:
I would think that with the new visuals, the current mission item is way more obvious than it used to be. The only case would be if it is off screen. In general I'm moving away from showing mission item numbers anywhere. I just don't see who they add any value.

Other than offscreen, I don't quite get why current mission item is unclear?

Antiheavy:
In addition to off-screen, there are times when the current mission item is not a waypoint and is sucked up/hidden.  In particular before Takeoff.  When troubleshooting an issue it can be valuable to know if the mission is correctly indexed and starting at 1.  this request isn't for an always shown value, but one that would be turned on for troubleshooting or in special situations.

DonLakeFlyer:
> there are times when the current mission item is not a waypoint and is sucked up/hidden.

Hmm, I was think that all of these cases would only last for like a few milliseconds after which the next waypoint would go current. That may not be the case. Need to think about the other item types. May need to do something special here.

> When troubleshooting an issue it can be valuable to know if the mission is correctly indexed and starting at 1. this request isn't for an always shown value, but one that would be turned on for troubleshooting or in special situations.

That's all good.

Antiheavy:
Had a user request for this feature today.  Related to no longer seeing mission item numbers on the map.  

loosely related to: https://github.com/mavlink/qgroundcontrol/issues/8478

Antiheavy:
upstream PR merged.  thanks!  https://github.com/mavlink/qgroundcontrol/pull/8577#event-3175996117

PR #8577, opened 2020-03-20
WP index to the values widget

dakejahl:
Adds current WP index as a Vehicle Fact for use to display in the values widget.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -77,6 +77,7 @@ const char* Vehicle::_altitudeAMSLFactName =        "altitudeAMSL";
 const char* Vehicle::_flightDistanceFactName =      "flightDistance";
 const char* Vehicle::_flightTimeFactName =          "flightTime";
 const char* Vehicle::_distanceToHomeFactName =      "distanceToHome";
+const char* Vehicle::_missionItemIndexFactName =    "missionItemIndex";
 const char* Vehicle::_headingToNextWPFactName =     "headingToNextWP";
 const char* Vehicle::_headingToHomeFactName =       "headingToHome";
 const char* Vehicle::_distanceToGCSFactName =       "distanceToGCS";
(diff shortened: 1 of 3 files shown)


Antiheavy:
Addresses https://github.com/mavlink/qgroundcontrol/issues/7791

DonLakeFlyer:
The naming for all of this should be "Mission Item Index" and variations of that. Indices are associated with all mission items not just waypoints.

src/Vehicle/Vehicle.cc
```diff
@@ -3984,6 +3989,13 @@ void Vehicle::_updateHeadingToNextWP()
     }
 }
 
+void Vehicle::_updateWpIndexNumber()
+{
+    const int _currentIndex = _missionManager->currentIndex();
```
DonLakeFlyer: This shouldn't use "_" for the variable name.

src/Vehicle/Vehicle.cc
```diff
@@ -3984,6 +3989,13 @@ void Vehicle::_updateHeadingToNextWP()
...
+
+    _wpIndexNumberFact.setRawValue(_currentIndex + 1);
```
DonLakeFlyer: The "+1" must be based on !firmwarePlugin()->sendHomePositionToVehicle(). If this is false then you +1 the index. This accounts for a different in how PX4 and ArduPilot handle mission items. ArduPilot always has a fake home position item in index 0, PX4 does not. QGC always has a home position item in index 0. So item indices coming directly from mavlink must be massaged appropriately.

dakejahl:
Thanks for the review, I've updated the PR with the requested changes.

src/Vehicle/VehicleFact.json
```diff
@@ -104,6 +104,12 @@
     "decimalPlaces":    1,
     "units":            "m"
 },
+{
+    "name":             "missionItemIndex",
+    "shortDescription": "Mission Item Index",
+    "type":             "uint16",
+    "units":            ""
```
DonLakeFlyer: This should just be left out

DonLakeFlyer:
Thanks Jake

src/Vehicle/VehicleFact.json
```diff
@@ -104,6 +104,11 @@
     "decimalPlaces":    1,
     "units":            "m"
 },
+{
+    "name":             "missionItemIndex",
+    "shortDescription": "Mission Item Index",
+    "type":             "uint16",
```
DonLakeFlyer: Unit tests are failing because there is an extra "," here

PR #8579, opened 2020-03-20
FollowMe: Fix "Always" settings for streaming GCS position

samuelctabor:
Currently it looks like GCS motion reports will only be sent to vehicles that are in the appropriate "Follow" mode even when the "Stream GCS Position" setting is "Always". I've confirmed that GCS motion reports (FOLLOW_TARGET mavlink messages) aren't received using 4.0.4, whereas they were in version 3.5.6, when in a non-follow mode.

This PR should fix the issue. I haven't been able to fully test it as I'm not set up to build on Android and don't have a GPS I can use on my development machine.


DonLakeFlyer:
Yup, thanks for the fix

PR #8582, opened 2020-03-21
Emit sensorHilChanged signal only if it was really changed

alexeylysenko:
Originally sensorHilChanged() was emitted every time enableSensorHIL() was called.
However, we need to emit signal only when the value was changed.


DonLakeFlyer:
This is all fine, but this is all dead code that should have been removed. HIL is no longer supported.

PR #8583, opened 2020-03-21
Removed unreachable code

alexeylysenko:
The 'rejected' variable is always false, so code under if (rejected) will be never executed.


booo:
As far as I can see `rejected` is now a unused variable.

DonLakeFlyer:
Yup looks like years ago there was an attempt to reject bad calibration but it didn't work.

PR #8590, opened 2020-03-23
Adjust minimum application window size

### LINKED ISSUE #8486: Minimum window size is too large.
## Expected Behavior
The QGC window can be reduced up to very minimal sizes (even if not all elements of the GUI are visible)

## Current Behavior
The window can only be reduced up to a certain amount which is much larget than it should be as there is still plenty of space.

## Steps to Reproduce:
1. Open latest QGC.
2. Try to reduce window size by dragging a corner.
3. It only works up to a certain point.

## Log Files and Screenshots
Screenshot from 2020-03-05 14-57-48

From the screenshot above you can see the minimal size that I can currently set the QGC window to. There is plenty of space it takes way more than half the screen in both directions. Unless I'm missing a setting this is a bug that should be fixed. Not sure if other users are experiencing this as well.

dogmaphobic:
In response to https://github.com/mavlink/qgroundcontrol/issues/8486

I'm using an arbitrary minimum size where the UI _mostly_ fits. Needs testing on other platforms. Especially when connected to low DPI monitors.

Screen Shot 2020-03-23 at 5 23 39 PM

Closes #8486

### DIFF
--- a/src/ui/MainRootWindow.qml
+++ b/src/ui/MainRootWindow.qml
@@ -25,8 +25,8 @@ import QGroundControl.FlightMap     1.0
 /// All properties defined here are visible to all QML pages.
 ApplicationWindow {
     id:             mainWindow
-    minimumWidth:   ScreenTools.isMobile ? Screen.width  : Math.min(215 * Screen.pixelDensity, Screen.width)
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I'm not super worried about desktop folks making the windows so small it doesn't work. If they do it to themselves, then it's there own problem.

DonLakeFlyer:
Can you bring to Stable as well

Issue #8592, opened 2020-03-24
App crash when using Plan Wisard: Plan -> Create Plan -> Survey


appetito:
## Expected Behavior
Starts creating survey plan with wisard
## Current Behavior
Segmentation fault

## Steps to Reproduce:
1. Select Plan view
2. Select Survey option in Create Plan list

## System Information
When posting bug reports, include the following information
- Operating System: MacOS X
- QGC Version: 4.0.0, 4.0.4
- QGC build: stable
- Flight Controller: not connected, radionink mini pix
- Autopilot (with version): not connected, arduplane v 4.0.1

## Detailed Description
Provide further details about your issue/bug.

## Log Files and Screenshots
macos conole log:
MissionControllerLog: _addMissionSettings
MissionControllerLog: _recalcWaypointLines homePositionValid false
MissionControllerLog: _recalcMissionFlightStatus
MissionControllerLog: _recalcWaypointLines homePositionValid false
MissionControllerLog: _recalcMissionFlightStatus
MissionControllerLog: _recalcWaypointLines homePositionValid true
MissionControllerLog: _recalcMissionFlightStatus
MissionControllerLog: _recalcWaypointLines homePositionValid true
MissionControllerLog: _recalcMissionFlightStatus
MissionControllerLog: _recalcWaypointLines homePositionValid true
Segmentation fault: 11

DonLakeFlyer:
Caused by changes made to support #8023

DonLakeFlyer:
Thanks for much for the bug report. This has been fixed and I just marked v4.0.5 to go out. Android v4.0.5 will start rolling out to store in a few minutes.

PR #8594, opened 2020-03-24
Plan: Remove display of incomplete item line connections

### LINKED ISSUE #8023: Incomplete Pattern items cause disjointed display
If the "Survey" Wizard item is selected and then a Waypoint is "inserted" seemingly at the end of the mission, but in fact it is being added before the Planned Home.

image

DonLakeFlyer:
* This essentially backs out the change to fix #8023. The correct fix for this is way more complex than I originally thought. And the existing code was buggy and causing crashes. Not possible to deal with in current stable. Will have to wait for point release.
* Fixed item sequence numbering problem for incomplete items. This could cause more than one item to have the same sequence number. Which would in turn cause things like multiple items being marked as current at the same time.


DonLakeFlyer:
Fix for #8592

PR #8601, opened 2020-03-25
add emergency stop as joystick button action

booo:
This commit adds the emergency stop as button action to the joystick.
Enables a user to map emergency stop to a button of the joystick.

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -64,6 +64,7 @@ const char* Joystick::_buttonActionGimbalUp =           QT_TR_NOOP("Gimbal Up");
 const char* Joystick::_buttonActionGimbalLeft =         QT_TR_NOOP("Gimbal Left");
 const char* Joystick::_buttonActionGimbalRight =        QT_TR_NOOP("Gimbal Right");
 const char* Joystick::_buttonActionGimbalCenter =       QT_TR_NOOP("Gimbal Center");
+const char* Joystick::_buttonActionEmergencyStop =      QT_TR_NOOP("Emergency Stop");
 
 const char* Joystick::_rgFunctionSettingsKey[Joystick::maxFunction] = {
     "RollAxis",
@@ -664,6 +665,7 @@ void Joystick::startPolling(Vehicle* vehicle)
             disconnect(this, &Joystick::gimbalYawStep,      _activeVehicle, &Vehicle::gimbalYawStep);
             disconnect(this, &Joystick::centerGimbal,       _activeVehicle, &Vehicle::centerGimbal);
             disconnect(this, &Joystick::gimbalControlValue, _activeVehicle, &Vehicle::gimbalControlValue);
+            disconnect(this, &Joystick::emergencyStop,      _activeVehicle, &Vehicle::emergencyStop);
         }
         // Always set up the new vehicle
         _activeVehicle = vehicle;
@@ -687,6 +689,7 @@ void Joystick::startPolling(Vehicle* vehicle)
             connect(this, &Joystick::gimbalYawStep,      _activeVehicle, &Vehicle::gimbalYawStep);
             connect(this, &Joystick::centerGimbal,       _activeVehicle, &Vehicle::centerGimbal);
             connect(this, &Joystick::gimbalControlValue, _activeVehicle, &Vehicle::gimbalControlValue);
+            connect(this, &Joystick::emergencyStop,      _activeVehicle, &Vehicle::emergencyStop);
             // FIXME: ****
             //connect(this, &Joystick::buttonActionTriggered, uas, &UAS::triggerAction);
         }
@@ -1008,6 +1011,8 @@ void Joystick::_executeButtonAction(const QString& action, bool buttonDown)
             _localYaw   = 0.0;
             emit gimbalControlValue(0.0, 0.0);
         }
+    } else if(action == _buttonActionEmergencyStop) {
+      if(buttonDown) emit emergencyStop();
     } else {
         qCDebug(JoystickLog) << "_buttonAction unknown action:" << action;
     }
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
This will not work from a safety standpoint. It would be too easy to accidentally hit the emergency stop button and have your vehicle fall out of the sky. This is why from the UI the guided actions are sliders. They require specific action beyond a click to make them work for safety reason. It's also the reason why this wasn't added ages ago. It requires some sort of new mechanism, possibly hold down for some amount of time to make it work safely.

booo:
I think hitting an emergency stop button should be as easy as possible.

I'm not 100% familiar with the flight modes/states and the transitions but I think a emergency stop button would be nice e.g. when the vehicle is on the ground, armed and someone approaches the vehicle. Or if the vehicle hovers over ground and someone approaches and disarm is impossible.

How is this handled with the standard radio? On the radio it's a switch you have to switch?

How could we implement a hold down for some time?

DonLakeFlyer:
I think possibly adding the option for buttons to require clicking twice within a short timeframe would be the way to go. Then you could use that option for emergency stop. Which would in turn prevent inadvertent presses from crashing the vehicle.

DonLakeFlyer:
I've posted a link to this pull on QGC Slack channel to gather some feedback on other peoples opinions.

Kaigott:
Having flown a multi-thousand dollar octocopter with QGC's joysticks, I definitely think it would be too dangerous to have an emergency stop button that could accidentally be hit. 

I like all of the other ideas up there, and if I were to design it, I would probably want this:

- The emergency stop button is easily accessible and pressable during armed flight modes, but only when the vehicle is **on the ground.** From a safety standpoint, if someone were to walk towards the drone when the propellers were spinning (inadvertently, of course), it would be nice to kill the motors quickly. 

- During the flight, the emergency stop button is still easily accessible, but it becomes much more difficult to accidentally press it. As an idea, I like the sliders that QGC currently uses, but it would be nice if this slider can appear where the button was pressed, rather than at the bottom of the screen. In an emergency, it is much easier to activate a function (like the kill switch) if you are able to find the slider sooner rather than later. 
tl;dr basically have the kill switch button near the joystick, and when you press and hold the button, a slider appears directly above the button where you can slide it to kill the vehicle, without having to find the slider at the bottom of the screen.

Overall, I think this addition is a good idea because safety is definitely one of the top priorities. Thanks for the great idea!

TSC21:
> Having flown a multi-thousand dollar octocopter with QGC's joysticks, I definitely think it would be too dangerous to have an emergency stop button that could accidentally be hit.

I do think this feature should be set on a button that has a cover on it. Like a missile-launch button.

dagar:
Can you add a small time threshold? I'm imagining a momentary button you're required to hold down for even 0.5 s (rather than instantaneous). I'd like to see the same mechanism for arm/disarm now that I think of it.

DonLakeFlyer:
> * During the flight, the emergency stop button is still easily accessible, but it becomes much more difficult to accidentally press it.

This is just talking about joystick buttons. Emergency Stop is already available from the QGC ui. If you click the Armed toolbar indicator in a state where you can't just disarm it shows the Emergency Stop slider.

DonLakeFlyer:
> Can you add a small time threshold? I'm imagining a momentary button you're required to hold down for even 0.5 s (rather than instantaneous). I'd like to see the same mechanism for arm/disarm now that I think of it.

Doable. The two possible mechanisms are a hold down to activate or a double-click type of thing.

DonLakeFlyer:
This whole thing kind of make you wonder whether default for all buttons should be some sort of hold time. Kind of like the solo controller used to work.

TSC21:
> > Can you add a small time threshold? I'm imagining a momentary button you're required to hold down for even 0.5 s (rather than instantaneous). I'd like to see the same mechanism for arm/disarm now that I think of it.
> 
> Doable. The two possible mechanisms are a hold down to activate or a double-click type of thing.

Again, if a RC have a button with a cover, why not offer that option as well?

dagar:
> > Can you add a small time threshold? I'm imagining a momentary button you're required to hold down for even 0.5 s (rather than instantaneous). I'd like to see the same mechanism for arm/disarm now that I think of it.
> 
> Doable. The two possible mechanisms are a hold down to activate or a double-click type of thing.

Hold down to activate with some visual on screen progress might be pretty nice. Thinking of something that might potential avoid a repeatedly tapping during a panic. You could also have an advanced option to bypass this for people with the right hardware (a trigger cover).

DonLakeFlyer:
> Hold down to activate with some visual on screen progress might be pretty nice. Thinking of something that might potential avoid a repeatedly tapping during a panic. You could also have an advanced option to bypass this for people with the right hardware (a trigger cover).

Like it

booo:
First of all the emergency stop action should be assignable to the joystick in some way. As far as I can see this is consensus.

Second there are multiple ways to implement a stop switch or in general trigger actions with a button. For each action and the corresponding button, it should be possible to e.g. configure a hold down time, double click, a repetition count or something a like. The user should be able to configure the behaviour of the button in the button assignment view.

The emergency stop is not copter specific. A rover can have a emergency too. 

I think especially the emergency stop button should be configured in a way that it is easy to trigger. All delays in the mechanism defeat the idea of a emergency stop button and potentially cause harm.

Thanks a lot for all the feedback!

dagar:
> All delays in the mechanism defeat the idea of a emergency stop button and potentially cause harm.

I disagree, but that's why we should make it configurable. I would argue that if the situation warrants an instantaneous terminate where even a fraction of a second is too long to hold then you probably should be using something more robust (and less complex) than a joystick + QGC + datalink to initiate termination. 

I believe we should try an optimize for a good default experience in common situations (eg plug a gamepad in), but keep advanced options available for users that need/want it. Accidentally tapping a gamepad momentary shouldn't be enough to arm, disarm, or terminate.

booo:
We maybe should consider button combinations and special stick positions as configuration options too?

Is anybody aware of an abstraction layer for SDL that e.g. supports button combinations. It doesn't feel right to implement all this again.

dagar:
> We maybe should consider button combinations and special stick positions as configuration options too?

Yes might be another interesting option if the configuration page can be made somewhat intuitive.

DonLakeFlyer:
I'll merge this in after my pull goes through. And then you'll be able to put something like emergency stop on a two buttons press to be safe. (If you want)

booo:
Are there any actions we should make available too? If you let me know which ones are actually missing I can create another pull request.

Thanks a lot for your work!

PR #8608, opened 2020-03-26
Add UVify Core in the firmware downloader.

limhyon:
- UVify Core board support is implemented.


DonLakeFlyer:
Also I just realized after I merge that you should use better commit messages which describe what you changed.

PR #8614, opened 2020-03-27
Removed unnecessary memory allocation

alexeylysenko:
From Qt docs: https://doc.qt.io/qt-5/qtglobal.html#qEnvironmentVariableIsEmpty


alexeylysenko:
@andrewvoznytsa Do you agree?

PR #8621, opened 2020-03-29
Fix "Pause" behaviour for ArduPilot

samuelctabor:
Pausing the mission via the "Pause" button on the Fly screen sets the vehicle to the relevant flight mode (LOITER, HOLD or BRAKE depending on type) but then the mode is immediately set to GUIDED so that the altitude can be set via the slider. This causes the "Resume Mission" and "Change Alt" actions to be unavailable.

This change causes the mode to change to GUIDED if the altitude change requested is more than 1cm, and also recognises the mission as paused when in GUIDED mode so that the "Resume Mission" and "Change Alt" actions are shown.

### DIFF
--- a/src/FlightDisplay/GuidedActionsController.qml
+++ b/src/FlightDisplay/GuidedActionsController.qml
@@ -217,7 +217,7 @@ Item {
     property var _actionData
 
     on_FlightModeChanged: {
-        _vehiclePaused =        activeVehicle ? _flightMode === activeVehicle.pauseFlightMode : false
+        _vehiclePaused =        activeVehicle ? _flightMode === activeVehicle.pauseFlightMode || activeVehicle.guidedMode: false
         _vehicleInRTLMode =     activeVehicle ? _flightMode === activeVehicle.rtlFlightMode || _flightMode === activeVehicle.smartRTLFlightMode : false
         _vehicleInLandMode =    activeVehicle ? _flightMode === activeVehicle.landFlightMode : false
         _vehicleInMissionMode = activeVehicle ? _flightMode === activeVehicle.missionFlightMode : false // Must be last to get correct signalling for showStartMission popups
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -912,7 +912,10 @@ void APMFirmwarePlugin::guidedModeChangeAltitude(Vehicle* vehicle, double altitu
         return;
     }
 
-    setGuidedMode(vehicle, true);
+    // Only set guided mode if altitude change has been requested.
+    if (abs(altitudeChange)>0.01) {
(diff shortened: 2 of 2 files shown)


DonLakeFlyer:
Hmm, what happens if the vehicle is travelling quickly and you do a pause and and altitude change? Does the vehicle end up overshooting the pause location greatly because it switched to guided too fast to change alt?

samuelctabor:
Hey Don

I'm more familiar with ArduPlane, and in that case LOITER and GUIDED modes are very similar except that GUIDED will accept target position changes (like the altitude change from the slider in QGC). In either mode the target location is set when the mode is entered so the aircraft will circle back when it overshoots.

For ArduCopter it appears BRAKE mode will try to stop ASAP and not go back to the location the mode was entered - so yes it will end up beyond the pause location if travelling at speed.

If this is undesirable maybe the pause flight mode should be changed from BRAKE to GUIDED in ArduCopterFirmware.cpp?

The altitude change from the slider is sent as an offset from the current target location stored on the autopilot. For this to have any effect the mode needs to be GUIDED.

This change just skips setting mode to GUIDED in the case that the altitude change is negligible, so the vehicle will stay in LOITER (in the case of Plane) at the current target location and altitude.

If the slider is changed, it will still set mode to GUIDED, the same as current behaviour, so that the change in target altitude is applied.

In either mode the target location is saved by the autopilot when the mode is entered, so there is no effect on the horizontal position of the target.

If the vehicle is moving rapidly it will go back to the

DonLakeFlyer:
> If this is undesirable maybe the pause flight mode should be changed from BRAKE to GUIDED in ArduCopterFirmware.cpp?

No, your change is great and make sense. I'm just trying to understand other unintended possible consequences of this Pause->Guided transition.

samuelctabor:
Sure, understood!

src/FlightDisplay/GuidedActionsController.qml
```diff
@@ -217,7 +217,7 @@ Item {
     property var _actionData
 
     on_FlightModeChanged: {
-        _vehiclePaused =        activeVehicle ? _flightMode === activeVehicle.pauseFlightMode : false
+        _vehiclePaused =        activeVehicle ? _flightMode === activeVehicle.pauseFlightMode || activeVehicle.guidedMode: false
```
DonLakeFlyer: I'm going to have to play around with this change a bit. In hindsight I don't quite remember why it also checked guided mode. There may be something important there.

DonLakeFlyer: Sorry I had this backward. You added the guided mode check. Why? This causes pause to not show up in some cases where it should.

samuelctabor:
I added it because ArduPlane and ArduCopter will go to GUIDED mode if pause.mission is selected and the altitude slider is changed - so it seems like this should also be recognised as being in the paused condition.

DonLakeFlyer:
Closing since this is covered by another pull

Issue #8623, opened 2020-03-30
[Feature Request] Dragging landing location on map translates both landing location and landing loiter equally.


kosakwayne:
When speaking with users, they often find it complicated to fine tune the landing location after it has been initially placed on the map. Currently, when moving the landing location on the map the landing loiter stays in the same location and the glide slope angle changes. This current behavior when paired with locking the landing loiter map movement to only be rotational forces the user to change the landing location and adjusting the glide slope to translate the loiter location. 

What if we set it up so translating the landing location also translates the landing loiter. This gives the user the option of rotating landing direction by rotating the loiter, or translating the landing and loiter by simply translating the landing location. This also emphasizes locking the glide slope angle unless intentionally changing it by changing the specific value. 

![image](https://user-images.githubusercontent.com/33092810/77928114-a268cb80-726d-11ea-9967-38b09b14ec7c.png)

DonLakeFlyer:
Make sense

PR #8624, opened 2020-03-30
Plan FW Landing: Drag of land/loiter maintains land distance/heading as appropriate

DonLakeFlyer:
* Dragging Landing position will maintain distance and heading
* Dragging Loiter position will only adjust angle
* Distance/Glide Slope is now only adjust from the item editor values

Fix for #8623


Antiheavy:
nice!   FYI @kosakwayne

Antiheavy:
Just testing this now.... wow...  this is actually life changing.

DonLakeFlyer:
> wow... this is actually life changing.

You must be under a stay at home order and are really bored!

PR #8625, opened 2020-03-31
Show mission items in Geofence edit mode

stefandunca:
Show mission items in geofence and rally edit mode.

Add interactive flag to the mission items that don't have it and disable interactions in Geofence and Rally edit mode.
Propagate transparency to the mission items in order to enable it in non-mission edit mode.

Mission edit mode
mission_always

Geofence edit mode
fence_always

Rally Points edit mode
rally_always

Mission from the screenshots
[geofence_test.zip](https://github.com/mavlink/qgroundcontrol/files/4527610/geofence_test.zip)

### DIFF
--- a/src/PlanView/FWLandingPatternMapVisual.qml
+++ b/src/PlanView/FWLandingPatternMapVisual.qml
@@ -24,6 +24,7 @@ Item {
     id: _root
 
     property var map        ///< Map control to place item in
+    property bool interactive: true
 
     signal clicked(int sequenceNumber)
 
@@ -178,6 +179,7 @@ Item {
         MouseArea {
             anchors.fill:   map
             z:              QGroundControl.zOrderMapItems + 1   // Over item indicators
+            visible:        _root.interactive
 
             readonly property int   _decimalPlaces:             8
 
@@ -199,6 +201,7 @@ Item {
             mapControl:     _root.map
             itemIndicator:  _loiterPointObject
             itemCoordinate: _missionItem.loiterCoordinate
+            visible:        _root.interactive
 
             property bool _preventReentrancy: false
 
@@ -224,6 +227,7 @@ Item {
             mapControl:     _root.map
             itemIndicator:  _landingPointObject
             itemCoordinate: _missionItem.landingCoordinate
+            visible:        _root.interactive
 
             onItemCoordinateChanged: _missionItem.moveLandingPosition(itemCoordinate)
         }
--- a/src/FlightMap/MapItems/MissionItemIndicator.qml
+++ b/src/FlightMap/MapItems/MissionItemIndicator.qml
@@ -38,6 +38,7 @@ MapQuickItem {
             showGimbalYaw:      !isNaN(missionItem.missionGimbalYaw)
             highlightSelected:  true
             onClicked:          _item.clicked()
+            opacity:            _item.opacity
 
             property bool _isCurrentItem:   missionItem ? missionItem.isCurrentItem || missionItem.hasCurrentChildItem : false
         }
--- a/src/FlightMap/MapItems/MissionItemIndicatorDrag.qml
+++ b/src/FlightMap/MapItems/MissionItemIndicatorDrag.qml
@@ -71,6 +71,7 @@ Rectangle {
         drag.maximumX:      itemDragger.parent.width - parent.width
         drag.maximumY:      itemDragger.parent.height - parent.height
         preventStealing:    true
+        enabled:            itemDragger.visible
 
         onClicked: {
             focus = true
--- a/src/MissionManager/QGCMapPolygonVisuals.qml
+++ b/src/MissionManager/QGCMapPolygonVisuals.qml
@@ -303,7 +303,7 @@ Item {
 
             sourceItem: SplitIndicator {
                 z:          _zorderSplitHandle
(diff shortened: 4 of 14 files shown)


hamishwillee:
FWIW I can see that you made mission items semi-transparent in geofence editing but not visa versa. In mission you can also see the "B" and it is same colour etc as a mission item so a little confusing.

For me it is a little confusing to sort wood from trees.

Ideas (which might be used in parallel)
1. Geofence uses a different accent colour altogether - e.g. purple or red. After all, it is a barrier. Ditto for the B.
1. There is a toggle to turn the ineditable items on/off in a particular view, or to control their opacity. 
1. Just a little more transparent :-)

src/PlanView/FWLandingPatternMapVisual.qml
```diff
@@ -178,6 +179,7 @@ Item {
         MouseArea {
             anchors.fill:   map
             z:              QGroundControl.zOrderMapItems + 1   // Over item indicators
+            enabled:        _root.interactive
```
DonLakeFlyer, line 182: I'd change this to visible which would seem safer. I think they end up being the same thing. But Qml is quirky as hell so who knows.

DonLakeFlyer:
Thanks for the pull Stefan. 

Comments:
* To me this whole thing should just switch to showing everything all of the time.
* With the opacity change showing all non current selection items as lower opacity non-interactive like you currently have. Mission is selected both fence and rally are shown but with lower opacity...
* The flight path lines should have opacity changed on them as well when not current
* In running your code I think non interactive opacity may need to go down to 0.5
* Testing this code shows the polyline tools for corridor scan are broken. Haven't looked into why. Add a corridor scan from a create plan template. When you click corridor scan item polyline tools don't show up any more.
* I think zorder for items will need to change dynamically when you switch between Mission/Fence/Rally. With the currrent selection bringing zorder to the front. Otherwise things like rally points could be under a survey polygon or a waypoint. Same thing could theoretically happen with fence polygon editing UI.

DanielePettenuzzo:
@stefandunca I tested this on linux. Everything works as expected except for when selecting the "Rally" tab. In this case I can still see my geofence but my mission disappears. Is this expected or is it an issue? 

Moreover I agree with @hamishwillee  that we we should discuss about colors because when the view is crowded due to many mission elements, it is difficult to understand which point is which.

DonLakeFlyer:
> Moreover I agree with @hamishwillee that we we should discuss about colors because when the view is crowded due to many mission elements, it is difficult to understand which point is which.

Starting point for this should be to lower the opacity as I said above. If that doesn't work then color plus opacity could be the next try. But I'd rather do opacity only first.

hamishwillee:
@stefandunca Which new images?

stefandunca:
> @stefandunca Which new images?

Strange, it didn't save. I just fixed it. How about now?

DonLakeFlyer:
@stefandunca What's the white triange in the images? Also can you attach the Plan you are using for testing?

DonLakeFlyer:
I did some simple testing with this. I created a simple mission and then a simple fence around it:
Screen Shot 2020-04-18 at 9 04 49 AM

Then when I went back to Mission from Fence the editing ui for the fence was still on, which is incorrect:
Screen Shot 2020-04-18 at 9 05 03 AM

I then went to Rally and the mission disappeared:
Screen Shot 2020-04-18 at 9 05 10 AM

hamishwillee:
This is a heap better and _probably_ usable/sufficient. 

FWIW though, if it were me I would **definitely** change geofence to another colour. Both here and in the editing screen. Probably also change the rally point colour too. Just my 2 bits.

stefandunca:
> Then when I went back to Mission from Fence the editing ui for the fence was still on, which is incorrect:

I keep showing it but with half transparency.

> I then went to Rally and the mission disappeared:

It is not clear to me if this is a requirement. My understanding is that the rally points are not related to mission, they are more to the landscape, independent from a specific mission. So you can have the same rally point with different missions. Is this right or should I add the mission also in Rally Point view?

> ... if it were me I would definitely change geofence to another colour. Both here and in the editing screen. Probably also change the rally point colour too ...

Any ideas about what color should they take? That can also go into another PR focused only on color alternatives.

stefandunca:
> Then when I went back to Mission from Fence the editing ui for the fence was still on, which is incorrect:

I have just fixed in the last amend

DonLakeFlyer:
> Any ideas about what color should they take? That can also go into another PR focused only on color alternatives.

I would deal with this separately. I can look at it after all this goes through.

> My understanding is that the rally points are not related to mission, they are more to the landscape, independent from a specific mission, so you can have the same rally point with different missions.

Sort of. Theoretially what you say could be correct. But rally points are associated with a "Plan" which is a combination of mission, geofence and rally points. There is no such thing as a reusable set of rally points as a user model concept. Hence they should display just like fence all the time.

stefandunca:
> ... There is no such thing as a reusable set of rally points as a user model concept. Hence they should display just like fence all the time.
Update PR to accommodate this. Added new screenshots and mission file.

DonLakeFlyer:
OK, thanks. Can you squash this down to a commit and I'll merge it in. Thanks

stefandunca:
> OK, thanks. Can you squash this down to a commit and I'll merge it in. Thanks

Done. Thanks

DonLakeFlyer:
Awesome, thanks @stefandunca

PR #8626, opened 2020-03-31
Plan: New wizard done model

DonLakeFlyer:
How things used to work:
* When you created a plan from a template say like survey...
* You would get a red Takeoff item (and others) which indicated it was incomplete. It told you what to futz and then to click "Done Adjusting" when done. At which point the editor switches back to normal editing mode.
* Same thing for items which required you to set up a Polygon/Polyline. 
* When you clicked "Done Adjusting" it would automatically move you to the next incomplete item as well. In this case that would be the Survey item

Problems:
* When you clicked to position the takeoff item. You then clicked "Done Adjusting" which in fact wasn't correct. You may need to still adjust the altitude for example. So you aren't really "Done Adjusting".
* Also the fact that when you clicked "Done Adjusting" it automatically moved you to the next incomplete item could cause you to miss the fact that there were other important things you really needed to take care of before you were "Done".
* In the case where the incomplete item was due to no Polygon/Polyline yet set for the item. When you added a Polygon say by tracing and clicked Done Tracing. You still needed to the click Done Adjusting which is a pain.
* Red highlights for incomplete were not quite strong enough visually.

To deal with the above I changed the user model for how these incomplete items work:
* Changed visuals to make incomplete items more prominent.
* Changed the "Done Adjusting" wording to just "Done"
* For items which just need a valid Polygon/Polyline to be completed there is no "Done" button. Finishing the add of the Polygon/Polyline will mark the item as no longer incomplete.
* When an item transitions away from incomplete it no longer automatically selects the next incomplete item.

Here you can see the new visuals for items which have a Done button:
Screen Shot 2020-03-31 at 10 57 30 AM

Here you can see the new visuals for items which only have Polygon/Polyline requirement:
Screen Shot 2020-03-31 at 10 57 52 AM

### DIFF
--- a/src/MissionManager/CorridorScanComplexItem.cc
+++ b/src/MissionManager/CorridorScanComplexItem.cc
@@ -41,18 +41,19 @@ CorridorScanComplexItem::CorridorScanComplexItem(PlanMasterController* masterCon
         _cameraCalc.distanceToSurface()->setRawValue(qgcApp()->toolbox()->settingsManager()->appSettings()->defaultMissionItemAltitude()->rawValue());
     }
 
-    connect(&_corridorWidthFact,    &Fact::valueChanged,                                this, &CorridorScanComplexItem::_setDirty);
-    connect(&_corridorPolyline,     &QGCMapPolyline::pathChanged,                       this, &CorridorScanComplexItem::_setDirty);
+    connect(&_corridorWidthFact,    &Fact::valueChanged,                            this, &CorridorScanComplexItem::_setDirty);
+    connect(&_corridorPolyline,     &QGCMapPolyline::pathChanged,                   this, &CorridorScanComplexItem::_setDirty);
 
-    connect(&_cameraCalc,           &CameraCalc::distanceToSurfaceRelativeChanged, this, &CorridorScanComplexItem::coordinateHasRelativeAltitudeChanged);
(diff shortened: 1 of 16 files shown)


DonLakeFlyer:
@antiheavy Lots of stuff changing here to improve usability problems which your folks (and others) have been complaining about.

Antiheavy:
Do you envision the Survey Wizard item will still have the simplified view to select a preset and angle as soon as a polygon is onscreen?  Or maybe it defaults to preset view first in the full blown Survey Pane?

DonLakeFlyer:
> Do you envision the Survey Wizard item will still have the simplified view to select a preset and angle as soon as a polygon is onscreen?

No. It was specifically removed on purpose. I don't want to create a parallel editing universe with this model. Because...

> Or maybe it defaults to the preset view first in the full blown Survey Pane?

Yes, it does. It always did. If you have presets the preset pane is shown first. So once you add a polygon you automatically go to the Presets pane by default in the Sentera case since you have Preset installed from scratch. And now that I think of that I didn't test that was still working. I'll double check.

Antiheavy:
> And now that I think of that I didn't test that was still working. I'll double check.

nope.  defaults to the main survey settings even with presets saved.

DonLakeFlyer:
> defaults to the main survey settings even with presets saved.

Ok, that's a bug. I'll fix it.

PR #8633, opened 2020-04-03
Reset the checklist when the activeVehicle changes

dakejahl:


DonLakeFlyer:
@dakejahl I brought this across to master and now I get this:
```
qrc:/qml/PreFlightCheckList.qml:37: TypeError: Cannot call method 'reset' of undefined
```

Which I think happens if you are not using checklists. Not sure if the same thing happens on Stable but if it does then it need fixing there as well.

Also now that I look a this, there is a magic reference to activeVehicle here:
```
    property var  vehicleCopy:       activeVehicle
```

Which just happens to be further up the hierarchy. This code should have it's own activeVehicle property. Not depend on something up the hierarchy would could be removed from that code and then it breaks this code.

dakejahl:
I see what you're saying. I think I remember being a little confused when I did this. How would I get an instance of `activeVehicle` into the checklist? Because the checklist is loaded with a loader item... I'm not entirely sure how to set properties of the qml object when it's instantiated like that (for the purposes of passing in the activeVehicle)

DonLakeFlyer:
You don't pass it in. You just make your own:
```property var _activeVehicle: QGroundControl.multiVehicleManager.activeVehicle```

PR #8637, opened 2020-04-05
Mobile: Turn off persistence of AppSettings::savePath

DonLakeFlyer:
On iOS builds the mobile save path includes an application UID in it. If QGC persists this and then you install a newer version that UID can change. Which in turn will cause QGC to not be able to load data from the correct location.

### DIFF
--- a/src/FactSystem/SettingsFact.cc
+++ b/src/FactSystem/SettingsFact.cc
@@ -77,3 +77,8 @@ void SettingsFact::_rawValueChanged(QVariant value)
 
     settings.setValue(_name, value);
 }
+
+void SettingsFact::setNoUpdate(void)
+{
+    disconnect(this, &Fact::rawValueChanged, this, &SettingsFact::_rawValueChanged);
+}
--- a/src/FactSystem/SettingsFact.h
+++ b/src/FactSystem/SettingsFact.h
@@ -26,6 +26,10 @@ public:
     Q_PROPERTY(bool visible MEMBER _visible CONSTANT)
 
     // Must be called before any references to fact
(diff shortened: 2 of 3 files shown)


DonLakeFlyer:
@antiheavy @dakejahl To test this on your custom iOS builds:
* Using the same exact repro steps which shows the problem
* When you have the first install make sure there are some files like a Plan file you can load
* And the second install/run which was failing check:
  * You now can pick files from the standard location
  * You can still see the Plan file you left from the first version

Antiheavy:
thanks @DonLakeFlyer , we'll give it a try as soon as we can!

bryanrahn:
@DonLakeFlyer I just tested the following:

1. Fresh app install, creating and saving a new flight plan
2. Update the app
3. Open previously saved flight plan

The correct file panel on the right was displayed. This is fixed! 👍🏽

Antiheavy:
> Update:
> 
> I think I found another issue. Digging in more.

I think the issue we found was that this isn't backwards compatible with previous versions.  Maybe that is just the way it is?

DonLakeFlyer:
> I think the issue we found was that this isn't backwards compatible with previous versions.

I don't know what that means? How can you have a previous version of your custom iOS build when you've never released one before.

Antiheavy:
> How can you have a previous version of your custom iOS build when you've never released one before.

we mean from internal test builds.

DonLakeFlyer:
I need to update this to include not reading from settings either. Since that could have been screwed up from previous builds.

DonLakeFlyer:
> I need to update this to include not reading from settings either. Since that could have been screwed up from previous builds.

I'm going to redo this pull to do this correctly.

PR #8643, opened 2020-04-08
Vehicle: ArduPilot does not support MAV_CMD_REQUEST_PROTOCOL_VERSION

muramura:
I am using QGC on ArduPilot.
I will be issued by the MAVLink message "MAV_CMD_REQUEST_PROTOCOL_VERSION" from QGC.
ArduPilot returns "unsupported".
ArduPilot has not supported this message in over three years.
Therefore, in the case of ArduPilot, this message should not be sent.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -283,10 +283,13 @@ Vehicle::Vehicle(LinkInterface*             link,
                        MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES,
                        false,                                   // No error shown if fails
                        1);                                      // Request firmware version
-        sendMavCommand(MAV_COMP_ID_ALL,
-                       MAV_CMD_REQUEST_PROTOCOL_VERSION,
-                       false,                                   // No error shown if fails
-                       1);                                      // Request protocol version
+
+        if (!apmFirmware()) {
+            sendMavCommand(MAV_COMP_ID_ALL,
+                           MAV_CMD_REQUEST_PROTOCOL_VERSION,
+                           false,                                   // No error shown if fails
+                           1);                                      // Request protocol version
+        }
     }
 
     _firmwarePlugin->initializeVehicle(this);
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
This pull is incorrect. QGC understands that MAV_CMD_REQUEST_PROTOCOL_VERSION is not supported by all firmwares and is supposed to handle it correctly.

Please detail the steps to reproduce this in a separate Issue. Be very explicit with the exact version of both QGC and ArduPilot you are using and against what vehicle type. Also make sure to provide the exact order of how you do things which then results in this error message being displayed with respect to whether the vehicle was connected prior to starting QGC that sort of thing. There is likely some specific sequence of steps which has a bug.

muramura:
@DonLakeFlyer san. Comments Thanks.

I implemented the following procedure.
1. Start ArduCopter V4.0.4-dev with SITL. 
2. Activate QGC on QT.

The following is the result of the above procedure.
APM: ArduCopter V4.0.4-dev (37d89d21)
APM: 7c414dbec5d648c689eb0165c4e5e004
APM: Frame: QUAD
Got MAVLink msg: COMMAND_ACK {command : 520, result : 0}
Got MAVLink msg: COMMAND_ACK {command : 519, result : 3}
Got MAVLink msg: COMMAND_ACK {command : 511, result : 0}

I got the following results with this PR.
APM: ArduCopter V4.0.4-dev (37d89d21)
APM: 7c414dbec5d648c689eb0165c4e5e004
APM: Frame: QUAD
Got MAVLink msg: COMMAND_ACK {command : 520, result : 0}
Got MAVLink msg: COMMAND_ACK {command : 511, result : 0}

DonLakeFlyer:
Sure but did QGC pop an error ui saying "Unsupported Command"? QGC code understands it can get an unsupported ack from MAV_CMD_REQUEST_PROTOCOL_VERSION and handles it correctly. And in fact it is required that it knows if MAV_CMD_REQUEST_PROTOCOL_VERSION is unsupported in order to negotiate mavlink v1/v2 protocol change correctly.

DonLakeFlyer:
What is the exact bug you saw?

muramura:
@DonLakeFlyer san.
QGC can tell you what type of firmware you have in your heartbeat.
QGC is sending unsupported messages without seeing this information.
QGC should not send unsupported messages by looking at this information.

Some other communication protocols implement the procedure of not issuing commands that are not supported by the protocol version, but issuing other commands.

DonLakeFlyer:
QGC implements the mavlink v1/v2 protocol negotiation per mavlink spec. Dealing with MAV_CMD_REQUEST_PROTOCOL_VERSION being unsupported is part of that spec. QGC works with any vehicle which supports mavlink (not just PX4 or ArduPilot) without knowing ahead of time what it does/does not support. Not sending MAV_CMD_REQUEST_PROTOCOL_VERSION in certain known cases breaks the protocol negotiation code since that is not how the spec works. And would in turn requiring writing a different v1/v2 negotiation protocol which is ArduPilot specific which doesn't make any sense.

muramura:
@DonLakeFlyer san. 
I understand. Thank you.

Issue #8639, opened 2020-04-06
Land start marker displayed on previous waypoint compared to what is executed


DanielePettenuzzo:
## Expected Behavior
If I set the land start marker to a specific waypoint in the plan UI I expect the vehicle to go to this waypoint to start a landing procedure (eg. if I trigger and RTL during the mission).

## Current Behavior
The vehicle starts the landing maneuver on the next waypoint compared to what was displayed in the mission plan UI. 

## Steps to Reproduce:
1. Plan a VTOL mission that has a land start marker
2. Trigger an RTL before the mission in over
3. The vehicle will not go to the waypoint that had the land start associated to it but to the next one instead

## System Information
When posting bug reports, include the following information
- Operating System: [e.g. Arch Linux]
- QGC Version: master from April 6, 2020 (git hash: 793d8d9632694d64153c8)
- QGC build: self-built from source
- Flight Controller: standard vtol in gazebo
- Autopilot (with version): PX4 master from April 6, 2020 (git hash: 92de0048410cf07c2c8)

## Log Files and Screenshots
Mission planning:
![qgc-landstart-plan-1](https://user-images.githubusercontent.com/35664507/78565299-ff402500-781d-11ea-8eea-8514d7d91f3b.png)
Mission execution:
![qgc-landstart-issue-1](https://user-images.githubusercontent.com/35664507/78565313-05360600-781e-11ea-933f-9bdbad1e0170.png)

DanielePettenuzzo:
FYI @RomanBapst @sfuhrer

DonLakeFlyer:
Hmm, that's a tricky one. For mission commands which do not include a position the visual representation is to highlight the previous item which has a location. For these types of commands this indicates that those commands will be executed "after" that waypoint. You can have commands after a DO_LAND_START such as bringing the gimbal back to neutral and stopping camera commands. All of which would happen prior to the new WP command. If you highlighted instead the "next" waypoint that would be even less correct since it would seem to show that the commands are the DO_LAND_START are not executed.

So yes it's odd for this specific case, but it's more correct from a consistent standpoint of how QGC deals with this.

DonLakeFlyer:
> For mission commands which do not include a position the visual representation is to highlight the previous item which has a location.

Would love to here opinions on whether in general that makes sense.

> So yes it's odd for this specific case, but it's more correct from a consistent standpoint of how QGC deals with this.

[...]

PR #8644, opened 2020-04-08
Don't add RTL for VTOLs and FixedWings

stefandunca:
This change removes the automatically added RTL waypoint for VTOL and FixedWing when autogenerating plans from templates.

"We want the operator to coordinate his landing. An RTL is not really predictable. That's why we want operators to work with land start markers, loiter to altitude, and land as a planned mission landing" @Yannick


DonLakeFlyer:
I'm going to be adding a VTOL landing pattern likely sometime today which will change all of this. Discussion is happening here: https://github.com/mavlink/qgroundcontrol/issues/8639#issuecomment-610803999

So this change will be superceded by all of that.

DonLakeFlyer:
Also should have said: This part of the change itself is correct. I'll just have all of it lumped together into a single pull with insertLandItem using a new VTOL landing pattern.

DonLakeFlyer:
> This part of the change itself is correct.

Sorry, not quite. Thought maybe I could merge this to use as base. But with new VTOL landing pattern this code you modified won't change.

DonLakeFlyer:
This no longer applies

PR #8645, opened 2020-04-08
Plan: VTOL landing pattern

DonLakeFlyer:
Added new VTOL Landing Pattern complex item:
* Works very similarly to FW Landing Pattern
* Sequence is: DO_LAND_START, camera futzing (optional), MAV_CMD_NAV_LOITER_TO_ALT, VTOL_LAND
* Modified VTOL default takeoff straight up distance to takeoff point to be 30 meters. Would like opinions on the the correct default for this value.
* VTOL loiter exit to land point also defaults to 30 meters. Would like opinions on the the correct default for this value.

There is an issue somewhere I can't find asking to set FW Landing Pattern heading to match the heading set on takeoff climbout. Under the assumption climbout was set to go into the wind. I would think the same thing applies to VTOL Landing Pattern loiter to land path. Once I figure out how to do that this will happen for VTOL as well.

Related to #8639 

Screen Shot 2020-04-08 at 1 02 28 PM

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,8 @@ Note: This file only contains high level features or important fixes.
 ## 4.1 - Daily build
 
 * Support mavlink terrain protocol which queries gcs for terrain height information. Allows planning missions with TERRAIN\_FRAME.
+* Fly: New instrument values display/editing support
+* Plan: Added new VTOL Landing Pattern support
 
 ## 4.0
 
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -618,6 +618,7 @@ HEADERS += \
     src/MissionManager/TakeoffMissionItem.h \
     src/MissionManager/TransectStyleComplexItem.h \
     src/MissionManager/VisualMissionItem.h \
+    src/MissionManager/VTOLLandingComplexItem.h \
     src/PositionManager/PositionManager.h \
     src/PositionManager/SimulatedPosition.h \
     src/Geo/QGCGeo.h \
@@ -823,6 +824,7 @@ SOURCES += \
     src/MissionManager/TakeoffMissionItem.cc \
     src/MissionManager/TransectStyleComplexItem.cc \
     src/MissionManager/VisualMissionItem.cc \
+    src/MissionManager/VTOLLandingComplexItem.cc \
     src/PositionManager/PositionManager.cpp \
     src/PositionManager/SimulatedPosition.cc \
     src/Geo/QGCGeo.cc \
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -242,6 +242,8 @@
         <file alias="VibrationPageWidget.qml">src/FlightMap/Widgets/VibrationPageWidget.qml</file>
         <file alias="VideoPageWidget.qml">src/FlightMap/Widgets/VideoPageWidget.qml</file>
         <file alias="VirtualJoystick.qml">src/FlightDisplay/VirtualJoystick.qml</file>
+        <file alias="QGroundControl/Controls/VTOLLandingPatternMapVisual.qml">src/PlanView/VTOLLandingPatternMapVisual.qml</file>
+        <file alias="VTOLLandingPatternEditor.qml">src/PlanView/VTOLLandingPatternEditor.qml</file>
     </qresource>
     <qresource prefix="/json">
         <file alias="ADSBVehicleManager.SettingsGroup.json">src/Settings/ADSBVehicleManager.SettingsGroup.json</file>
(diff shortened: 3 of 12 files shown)


DonLakeFlyer:
Once this goes in I could use testing on this from VTOL people.

sfuhrer:
@DonLakeFlyer I tested it in SITL a bit. Really cool feature, this improves VTOL usability a lot!

- Overall I find the usage intuitive and easy (hope the same is the case for less experienced users). 
- I especially like the "Set to vehicle heading" button, as if the user know what he's doing, he places the vehicle with the nose into the wind, and thus after pressing this button the backtransition and landing part is also with the nose into the wind.
- The 2 text boxes "Heading" and "Landing Dist" are currently inside the Landing point settings, but are rather settings for the loiter position, no? I actually don't think these need to be exposed like that, the as dragging the loiter to the correct position and/or using the "Set to vehicle heading" is the correct workflow, no?
image
- I think the default you chose make sense

DonLakeFlyer:
@sfuhrer Thanks for testing.

> * The 2 text boxes "Heading" and "Landing Dist" are currently inside the Landing point settings, but are rather settings for the loiter position, no?

It depends on your point of view. I think either is correct. The Heading dispalyed is from Loiter to Land. And Landing Distance is is in relationship to both points.

> * I actually don't think these need to be exposed like that

Some people like dragging some dont. For example changing just the heading by typing allows you to keep the same exact landing distance you had before. Whereas dragging affects both heading and distance which may or may not be what you want.

Antiheavy:
> Some people like dragging some dont. For example changing just the heading by typing allows you to keep the same exact landing distance you had before. Whereas dragging affects both heading and distance which may or may not be what you want.

exactly.   Although there was a recent change to FW Landing Pattern to only let the dragging effect the heading, not distance.  This is important to FW as there is a minimum landing distance based on glideslope and loiter altitude (and a few other parameters).  It sounds like VTOL also has it's own set of constraints for backtransition.  You should think about if you want VTOL Landing Pattern to have a locked distance for dragging or if you are okay with users screwing it up constantly ;-)   Of course another approach could be for QGC to have a minimum drag distance that was somehow set in the app settings.

DonLakeFlyer:
> Although there was a recent change to FW Landing Pattern to only let the dragging effect the heading, not distance.

Yeah I wasn't sure if that was needed or not. I think it depends on this:

> It sounds like VTOL also has it's own set of constraints for backtransition.

If there is a standard distance from loiter to land distance which will work in most cases and it's important that distance doesn't get smaller than that default then locking the distance during drag would be a good thing. I don't know enough about VTOL to say yes or no to that.

sfuhrer:
@Antiheavy good point about the min distance from loiter to land. For VTOL this is not as crucial as for FW, but still important. And I guess in a product you would even want to fix the distance and only add the possibility to change the heading.
To quickly show how PX4 handles the backtransition: It calculates the distance from home to make start the backtransition based on the parameter VT_B_DEC_MSS (deceleration during backtransition in m/s/s) and the current groundspeed. The distance is then v^2/(2*VT_B_DEC_MSS).
The exact groundspeed we can't know at the planning stage, but with can estimating it by taking the parameter FW_AIRSPD_TRIM for it, times a factor of 1.2 I would say. We then should prob adjust this factor after doing some flight testing.
What do you guys think, should we only put the initial distance from loiter to land to the thus calculated distance, or enforce it, or only enforce the minimum distance to and allow larger ones? I would go for the latter I think.

Antiheavy:
> only enforce the minimum distance to and allow larger ones?

This seems best to me as well.  I know there has been resistance to depending on QGC to read vehicle params to aid in mission planning - mostly due to the offline planning scenario.  Maybe one ideas is adding Landing Pattern defaults to the "Plan View" section under General Settings, something like this:
image

DonLakeFlyer:
> I know there has been resistance to depending on QGC to read vehicle params to aid in mission planning - mostly due to the offline planning scenario.

Exactly. Solvling the offline problem correctly is way beyond a small change. Hence they need to be fixed values, but can be settings as Todd points out. If someone can tell me what they think the default land distance for VTOL is I can put it in. 

I would make it work this way:
* Minimum distance is enforced. Users always have the "Force Save" override to go below.
* Loiter to Land distance is locked when dragging  from the ui. If the user needs to modify distance they can do it from the entry field.

sfuhrer:
@DonLakeFlyer your suggestions make total sense. I would put 50m as min distance and 100m as default.

hamishwillee:
@DonLakeFlyer Can I please confirm how this works? I'm fairly sure, but this is a bit of a discussion.
1. The purpose is to drop the height of a VTOL flying in fixed wing mode to an acceptable level before transitioning to multicopter mode for landing? 
   - This being faster?, more battery efficient? safer? than descending from from height in MC mode?
1. The land consists of a (FW) loiter where vehicle descends to landing altitude, then a vtol land waypoint that gets placed by default where the vehicle takes off
1. The settings look like below

   image

Most of this is documented in QGC, but for camera, at what point do the images/video stop being captured - from land start, during loiter, during VTOL land?

DonLakeFlyer:
> 1. The purpose is to drop the height of a VTOL flying in fixed wing mode to an acceptable level before transitioning to multicopter mode for landing?

Correct. MC is expensive battery wise. So it is better to drop altitude in FW mode. It also is important with respect to creating a landing sequence which begins with DO_LAND_START. This allows you to abort a landing a do it over again if needed.

> 2\. At what point do the images/video stop being captured - from land start, during loiter, during VTOL land?

The sequence is: DO_LAND_START, stop photo/video, LOITER_TO_ALT, VTOL_LAND

> 2\. What does set to vehicle location actually do?

Set the location of the landing point to be the current vehicle location.

> 2\. What does set to vehicle heading actually do.

Set the heading from loiter to land to the current vehicle heading.

Set to vehicle location/heading are handy to use to match your landing pattern to physical field locations/dimension/wind.

Also note: All of these answers apply to FW Landing Pattern item as well.

PR #8649, opened 2020-04-09
Always generate AppSettings::savePath at runtime for mobile builds

DonLakeFlyer:
Replacement for #8637


DonLakeFlyer:
@antiheavy Can you guys this is out using the same process as the other pull?

Need to wait to hear back from @dogmaphobic to make sure this doesn't cause any other custom build problems.

Antiheavy:
> @Antiheavy Can you guys this is out using the same process as the other pull?
> 
> Need to wait to hear back from @dogmaphobic to make sure this doesn't cause any other custom build problems.

Yep! But might not be until next week sometime when we get back to syncing with upstream and testing another development build.

PR #8655, opened 2020-04-12
Value range visual change support

DonLakeFlyer:
I'm hoping this is fairly self explanatory. As always major WIP but it should all be working. Certainly more ui tweaking needed to make it easier to use, but functionality behind the scenes is all hooked up.

Change the icon or text color based on value:
Screen Shot 2020-04-12 at 4 22 54 PM

Change the icon based on value:
Screen Shot 2020-04-12 at 4 23 19 PM

Changed the opacity of the icon based on value:
Screen Shot 2020-04-12 at 4 22 32 PM

The direction I'm heading is to also re-do the toolbar to make it more of a useful piece of real estate. Currently it takes up a lot of space for not a lot of value. The direction is to make it configurable just like the instrument value page (using this same code) such that you can specify what values you want to see there all the time. And then the instrument value page becomes secondary. For example you don't want to miss your vehicle plummeting to the ground because you are farting around with the camera settings page in the instrument panel which in turn hides the instrument values. So critical instrument values in the toolbar, secondary in the instrument panel. End users have full control over tweaking. Custom builds have full control over creating there own default setups.


DonLakeFlyer:
Also once I'm closer to being done I want to gather some folks together to help create the new default setup for both toolbar and instrument values for each vehicle type.

Issue #8585, opened 2020-03-23
"Set RC to Param..." button no action 4.0.1


TheSoundMan000:
Windows 10
4.0.1 Downloaded direct from website
Pixhawk 4/Pixhawk 2, both latest stable PX4

Drone configured and maidened with obvious PID tuning necessary. PARAM1 and PARAM2 assigned to channel 11 and 12 respectively (analog knobs). Select a desired paramater eg MC_ROLLRATE_P and advanced then "Set RC to Param..." and nothing happens.

Completely unable to assign RC control to tune the PIDs mid flight.

Expecting something like the below but not making it to the dialog window.
![](https://docs.px4.io/v1.9.0/images/qgc/setup/parameters_radio_channel_mapping.jpg)

DonLakeFlyer:
@dogmaphobic This is sitll missing from Stable

TheSoundMan000:
Is it available in the latest daily? I might give that a burn if so. Thanks for your help guys.

DonLakeFlyer:
It's not available anywhere

TheSoundMan000:
Ah right... so what version is depicted in the above screen shot? That's straight from the PX4 1.9.0 manual. I don't mind if its beta or anything like that just chasing an implementation of some form. Is this a new feature or an obsolete one or whats the situation?

DonLakeFlyer:
The is QGC v3.5.6, it is missing from newer versions. You can get older version from the GitHub releases page.

PR #8657, opened 2020-04-14
Set RC To Param support

DonLakeFlyer:
* Fixes: #8585, #8382
* Also includes some changes to showPopupDialog:
  * No need to pass in title and buttons. You now set them in the component itself.
  * Ability to pass additional dialog properties from the showPopupDialog call. This allow you to provide external context to the dialog if needed.


DonLakeFlyer:
Screen Shot 2020-04-14 at 12 07 16 PM

PR #8662, opened 2020-04-15
Add support for multi-press joystick buttons

DonLakeFlyer:
Related to discussion in #8601

Screen Shot 2020-04-15 at 2 18 04 PM


DonLakeFlyer:
Note that this changes the existing behavior of assigning the same action to multiple buttons. That was allowed before as well. But a single button press on any of the buttons assigned the same action would work. The loss of that capability doesn't seem as important as the new capability added.

PR #8664, opened 2020-04-16
Vehicle: fix removal of hashtag in statustext

julianoes:
This fixes a bug where a URL containing a hashtag (#) was displayed without the hashtag.

This is untested, feed forward programming :smile:.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -1000,8 +1000,17 @@ void Vehicle::_chunkedStatusTextCompleted(uint8_t compId)
 
     // If the message is NOTIFY or higher severity, or starts with a '#',
     // then read it aloud.
-    if (messageText.startsWith("#") || severity <= MAV_SEVERITY_NOTICE) {
+    bool readAloud = false;
+
+    if (messageText.startsWith("#")) {
         messageText.remove("#");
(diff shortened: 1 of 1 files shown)


src/Vehicle/Vehicle.cc
```diff
@@ -1000,8 +1000,17 @@ void Vehicle::_chunkedStatusTextCompleted(uint8_t compId)
 
     // If the message is NOTIFY or higher severity, or starts with a '#',
     // then read it aloud.
-    if (messageText.startsWith("#") || severity <= MAV_SEVERITY_NOTICE) {
+    bool readAloud = false;
+
+    if (messageText.startsWith("#")) {
         messageText.remove("#");
```
DonLakeFlyer, line 1006: This isn't correct either. QString::remove removes all occurances of the specified character. This should only remove the first char from the string.

julianoes: True, I'll fix it.

julianoes: Done in 68b649e9a.

julianoes:
@DonLakeFlyer I think this is ready for merging.

DonLakeFlyer:
Thanks @julianoes

PR #8666, opened 2020-04-16
Use different UI for App/Vehicle messaging

DonLakeFlyer:
* Vehicle messages use the same modeless popup as before
* Application messages now use mainRootWindow.simpleMessageDialog
* From C++ code:
  * Application messages now use qgcApp()->showAppMessage(...)
  * Vehicle messages use qgcApp()->showVehicleMessage(...)

This way the user can tell the difference between a message which comes from the vehicle and an error message which is coming from QGC itself. It also sets the stage for being smarter about how to display vehicle messages since we now know the difference between the two.


DonLakeFlyer:
Also forgot to say that mainRootWindow.simpleMessageDialog now supports being called multiple time and all the dialogs will show correctly.

Issue #4485, opened 2017-01-29
Same version firmware update should overwrite existing meta


jgoppert:
When developing firmware we are changing params a lot. It seems we have to delete ~.config/QGroundcontrol* each time so it will upload the new data. If the version numbers match it seems it won't update. The behavior  I think should be that if the uploaded version number is >= current it should update.

jgoppert:
I think the related logic is here: https://github.com/mavlink/qgroundcontrol/blob/master/src/FactSystem/ParameterManager.cc#L1095

We maybe need to invalidate the cache if a firmware file is uploaded with a the same version as a file in cache.

DonLakeFlyer:
On the surface this seems like an ok change. Need to think a little more to make sure,

DonLakeFlyer:
Supported now. But will be replaced by COMPONENT_INFORMATION support

PR #8670, opened 2020-04-18
Cache new metadata if versions are equal

DonLakeFlyer:
Fixes for https://github.com/PX4/Firmware/issues/11072, #4485


NuclearManD:
This does not appear to be in the latest release - when do you think this will reach a release of QGroundControl?

DonLakeFlyer:
> latest release

This is in daily builds, it is not in stable. What is your definition of "latest release"?

NuclearManD:
I suppose if it's available here:
https://docs.qgroundcontrol.com/en/getting_started/download_and_install.html

Yea, sorry, that was sort of vague.  I suppose I'm really asking "When will this PR make it into the version of QGC we download from your docs or website"

DonLakeFlyer:
Both Stable and Daily builds are available for download from the website.

PR #8672, opened 2020-04-20
Fix QGCPipable "Transparency" for mouseEvents

tcanabrava:
QGCPipable used to eat all the mouse events so it was impossible
to add anything in the map or the video (like a row of commands)
This patch adds a new button that changes how the event handling works
buy using a hoverHandler that reacts only on hover, and a new
button that triggers the swap between map and video.

The icon of the Button is wrong - but I'm not an artist :)

This is needed to add more complex controls to the Video Item underneath

### DIFF
--- a/src/QmlControls/QGCPipable.qml
+++ b/src/QmlControls/QGCPipable.qml
@@ -8,8 +8,8 @@
  ****************************************************************************/
 
 
-import QtQuick                      2.3
-import QtQuick.Controls             1.2
+import QtQuick                      2.12
+import QtQuick.Controls             2.12
 
 import QGroundControl               1.0
 import QGroundControl.ScreenTools   1.0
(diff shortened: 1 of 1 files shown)


tcanabrava:
@dogmaphobic @DonLakeFlyer please review, using MouseArea to get containsMouse breaks logic for any clickable area underneath not in the same qml equivalent of DOM - but in the same Z.

MatejFranceskin:
I have tested it and it works, but I see one problem in the GUI logic.
Now instead of click on the PiP video to enlarge it you have to click the button, but the button becomes visible only when you hover the mouse on it, so on touch screen you will never see it and will have to "know" that you have to click on the upper part of the picture to enlarge it.

DonLakeFlyer:
> but the button becomes visible only when you hover the mouse on it

As a rule you can't hide/show controls based on hover in QGC.

DonLakeFlyer:
From a usability standpoint I don't see the benefit of adding controls to a pip'ed view being important enough to lose the simplicity of clicking anywhere in the PIP to switch. I would content that adding more complexity to something while PIP'ed is not a good thing. If you need full control, then bring it to full window. That said there is no context with examples as to why this new feature is interesting or how it would be used. So maybe it's useful, but without context I don't see the need to add complexity.

PR #8673, opened 2020-04-20
remove stray debug messages

remster:
I can only imagine this was a stray debug message
foobar

and is so, this is a perfect material for my warmup PR.


DonLakeFlyer:
Oops screwed up pull, sorry.

PR #8677, opened 2020-04-21
QGCApplication: Use forward declaration in QQmlApplicationEngine

patrickelectric:

### DIFF
--- a/src/QGCApplication.h
+++ b/src/QGCApplication.h
@@ -11,7 +11,6 @@
 
 #include <QApplication>
 #include <QTimer>
-#include <QQmlApplicationEngine>
 #include <QElapsedTimer>
 
 #include "LinkConfiguration.h"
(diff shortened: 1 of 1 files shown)


patrickelectric:
ping @DonLakeFlyer

src/QGCApplication.h
```diff
@@ -30,6 +29,8 @@
 #include "OpalLink.h"
 #endif
 
+class QQmlApplicationEngine;
```
DonLakeFlyer, line 32: Can you move this down with the other forward references and remove the obvious comment about circular header include.

patrickelectric: This is not fixing any circular header includes, as the comment in the top of the others forward references. But Ok.

patrickelectric:
Done.

PR #8678, opened 2020-04-21
Fix Qt environment style in linux systems

patrickelectric:
QGC appears broken when compiled in linux environment that does not provide QT_QUICK_CONTROLS_STYLE
Without this patch, that is what I have here:
image

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -250,6 +250,7 @@ QT += \
     positioning \
     qml \
     quick \
+    quickcontrols2 \
     quickwidgets \
     sql \
     svg \
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -28,6 +28,7 @@
 #include <QFontDatabase>
 #include <QQuickWindow>
 #include <QQuickImageProvider>
+#include <QQuickStyle>
 
 #ifdef QGC_ENABLE_BLUETOOTH
 #include <QBluetoothLocalDevice>
@@ -217,6 +218,11 @@ QGCApplication::QGCApplication(int &argc, char* argv[], bool unitTesting)
                 }
             }
             permFile.close();
+        }
+
+        // Set default QtQuick style if not configured
+        if (QString(getenv("QT_QUICK_CONTROLS_STYLE")).isEmpty()) {
+            QQuickStyle::setStyle("Universal");
         }
     }
 #endif


DonLakeFlyer:
I've built and run master like a 100 times a day on Linux in the last week without problems. Are you sure you have everything set up correctly? What changed which suddenly caused you build to not work.

patrickelectric:
@DonLakeFlyer Some linux DE does not set / use this variable, I believe that anyone that is using the last Plasma version has this problem. @tcanabrava can confirm this.
My setup:
```
Operating System: Arch Linux 
KDE Plasma Version: 5.18.4
KDE Frameworks Version: 5.69.0
Qt Version: 5.14.2
Kernel Version: 5.6.5-arch3-1
OS Type: 64-bit
Processors: 16 × AMD Ryzen 7 2700 Eight-Core Processor
Memory: 15.7 GiB of RAM

```

tcanabrava:
Patrick is correct, the reasoning is not really.
this is a bug loading qc1 and qc2 in a system that uses a different theme as "Universal" for the whole DE, QGroundControl tries to load it's on theme and it gets wrong information from the pre-loaded theme. 
Forcing the theme to be Universal for QGC only fixes it.

@DonLakeFlyer if you want to verify, use a up-to-date distro running Plasma.

DonLakeFlyer:
I"m concerned about unintended consequences of this change. For example setting the theme to Universal does what? I tried hacking QGC to spit out what theme it is currently running and it output and empty string for QQuickStyle::name(). Is "Universal" different then what say QGC is using on a vanilla Ubuntu LTS version?

Also the fact these global environment variables: https://doc.qt.io/qt-5/qtquickcontrols2-environment.html (which I'm seeing for the first time) affect all Qt applications seems crazy as a default thing. You can even turn off hover events globally. Seems like insanity to me.

@dogmaphobic Can you look at this. You maybe have more experience with control2 since you switched things over.

tcanabrava:
That's not really crazy as default: That's the only way to have a unified look & feel. What I feel crazy as default is that projects such ours theme every single bit of it so it does not resemble any common look & feel - it brings us bugs (such as this one or the combobox bug we had for years).

Currently the bug affects all users of plasma (and that's the majority of linux users)

I would really like to see QGroundControl to integrate in my desktop theme to be honest, inherit the system pallete and work with it, but I know this is far fetched. :)

Pehaps there's a way to say "I don't want to load the system theme".

DonLakeFlyer:
> it brings us bugs

I'll trade two bugs in the many year history of QGC, for the hundreds of bugs that would be caused by having the ui be different on every single OS.

patrickelectric:
Ok, so what is the status ? Should we merge it or fix it in a different way ?

DonLakeFlyer:
Waiting to hear with @dogmaphobic has to say.

patrickelectric:
Ping @dogmaphobic 
@DonLakeFlyer is he still working for qgroundcontrol ?

DonLakeFlyer:
Gus is around he's just been crazy busy for weeks now. He'll get to this but it will take time.

dogmaphobic:
I won't have the time to test any of this. Just make sure these changes are wrapped around `#if defined (__linux__)` as that seems to be the only platform with these issues. The changes are really, really small so if something blows up, it won't be hard to come back and revisit it.

patrickelectric:
@dogmaphobic and @DonLakeFlyer the changes are already inside of:
```
#ifdef Q_OS_LINUX
#ifndef __mobile__
```

PR #8679, opened 2020-04-21
Improve category log menu

patrickelectric:
- Add Set all button
- Add search category

Video: https://streamable.com/0k4gf9

### DIFF
--- a/src/QmlControls/AppMessages.qml
+++ b/src/QmlControls/AppMessages.qml
@@ -11,6 +11,7 @@ import QtQuick                  2.3
 import QtQuick.Controls         1.2
 import QtQuick.Controls.Styles  1.4
 import QtQuick.Dialogs          1.2
+import QtQuick.Layouts          1.12
 
 import QGroundControl               1.0
 import QGroundControl.Palette       1.0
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Love the search. But don't see whan anyone would ever use Set All?

patrickelectric:
Well, we never know, at least is an option.

DonLakeFlyer:
I'd rather not have it there. The reason is that I spend a lot of time asking people to turn on various logging options to debug things remotely. I've had many cases where folks then do this from the command line and don't do what I ask and just turn on all logging. Which then makes it next to impossible for me to find the information I'm looking for. Making this too easy from the UI will likely make that problem for me even worse. If you really, really need to do that you can still do it from the command line.

patrickelectric:
@DonLakeFlyer "Set all" removed

patrickelectric:
Ping

DonLakeFlyer:
If you put all of this inside the flickable it will work much better on small screens like phones. Those just don't have much vertical real estate to devote something which sticks on the top.

patrickelectric:
@DonLakeFlyer done

PR #8680, opened 2020-04-22
Fix magic reference to vehicle in PreFlightChecklist

dakejahl:
See https://github.com/mavlink/qgroundcontrol/pull/8633#event-3236058051 for reference.


DonLakeFlyer:
Can you bring this to master as well?

Issue #8273, opened 2020-02-03
Grid Survey `distanceToSurface` used as MAV_FRAME_GLOBAL_RELATIVE_ALT


remster:
## Expected Behavior
With the takeoff on a good hill and a grid survey **below** that hill; like so:
![hill](https://user-images.githubusercontent.com/8327961/73634593-cfc83d00-4661-11ea-8bf4-da07c4fa9fc3.png)
- QGC should recalculate the transects based on distance to surface at the survey's centroid;
- it should be possible to set the **rel** altitude of that survey (as it is possible for a waypoint) to a negative value - and fly below the takeoff point (that's high on a hill);

## Current Behavior
QGC [recalculates the survey transects based `CameraCalc::distanceToSurface`](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/CameraCalc.cc#L163), which sounds correct because the closer the vehicle is to the surface the smaller bit of ground its field-of-view sees and the denser should the transects be. However, QGC [defaults `CameraCalc::distanceToSurface` to the **mission item altitude**](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/SurveyComplexItem.cc#L83) which (with default alt mode) is **not** distance to surface where the survey is but distance to surface at takeoff (which may be hundreds of metres below or above the survey). The consequence of this is that, as one moves the survey between hills and valleys, the transects do not change despite the distance to surface does change. 

Importantly the `.plan` must separate `MAV_CMD_NAV_WAYPOINT` altitude from distanceToSurface as (for MAV_FRAME_GLOBAL_RELATIVE_ALT) they are two different values:
```
"TransectStyleComplexItem": {
    "CameraCalc": {
        "AdjustedFootprintFrontal": 25,
        "AdjustedFootprintSide": 25,
        "CameraName": "Manual (no camera specs)",
        "DistanceToSurface": 50, <<-- distance to surface at survey centroid
        "DistanceToSurfaceRelative": true,
        "version": 1
    },
    "CameraShots": 98,
    "CameraTriggerInTurnAround": true,
    "FollowTerrain": false,
    "HoverAndCapture": false,
    "Items": [
        {
            "autoContinue": true,
            "command": 16,
            "doJumpId": 3,
            "frame": 3,
            "params": [
                0,
                0,
                0,
                null,
                35.78288117563359,
                -121.33179433564969,
                50 <<-- distance to surface at mission takeoff
            ],
[...]

Issue #8560, opened 2020-03-15
[Feature Request] option to prevent mission upload if there is a known terrain intersection


Antiheavy:
This may be a can-of-worms topic, but I think it might be worth considering a Plan View Settings option to prevent/allow mission upload if there is a known terrain intersection with a waypoint.  I'd imagine a similar pop-up warning to the one that prevents upload if there are incomplete plan wizard items.

For example:

![image](https://user-images.githubusercontent.com/12117027/76693341-5133c780-6631-11ea-9c15-e792aef1b041.png)

There is a somewhat related larger discussion about more robust mission validators here: https://github.com/mavlink/qgroundcontrol/issues/8510

Antiheavy:
Related to this request, it would be nice for a Plan View option for "minimum terrain clearance warning" that would warn user about uploading a mission that had waypoints planned too close to the ground (trees!):
![image](https://user-images.githubusercontent.com/12117027/79808991-021a3a00-8335-11ea-86bd-36603cc36ef7.png)

PR #8682, opened 2020-04-22
Calculate transect altitudes from distanceToSurface and terrain

remster:
Addresses: https://github.com/mavlink/qgroundcontrol/issues/8273

This is the most minimalist manner of addressing the problem I could think of. Here I chose to calculate altitudes for transect waypoints from `CameraCalc::distanceToSurface` **and from terrain query**.
This way of fixing it leaves the two related problems intact:
- There is no visual discrepancy between the "Altitude" UI element for a plain waypoint and for a Survey - but their interpretation is very different. The former is treated as altitude of the waypoint (relative or otherwise), the latter is treated as `distanceToSurface` below the survey polygon (more akin to AGL), which cannot be used directly as transect waypoint alt (what this PR is about)
- There is logically and structurally no such thing as [`CameraCalc::distanceToSurfaceRelative`](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/CameraCalc.h#L36). There isn't logically because for `CameraCalc` this distance is always absolute (grid survey: vehicle towards ground directly below; structure survey: vehicle towards structure sideways). There isn't structurally because `CameraCalc` never cares what what the value of `distanceToSurfaceRelative` is, which is a hint this member doesn't belong to its dwelling. It is a consequence of trying to treat `CameraCalc::distanceToSurface` both to correctly calculate transects (AGL) and as transect waypoint (relative or otherwise) altitude. 

I do have a draft fix that addresses those two problems and instead flips it and chooses to calculate `distanceToSurface` **from** waypoint altitude, but this draft is way more invasive than this PR.

Because of [this](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L44) (which i think is dangerous to keep) this PR and all function related to terrain (QGC key datapoint) are very difficult to test. How about:
```
    if (qgcApp()->runningUnitTests()) {
         QList<double>() result;
         for (auto const& point: coordinates) {
             result.push_back(
                std::round((point.longitude - std::round(longitude))*100)
             );
         }
        return;
    }
```
To enable standard ersatz/deterministic terrain values? I could test the affected function against it. 

Here's how this PR treats a 
### Higher Ground
new-hill

`MissionController` sees this:
```
0 absoluteAltitude 153
1 absoluteAltitude 203
2 absoluteAltitude 530.147
```
And the plan (relative alt/above takeoff) renders transects at 50m from the avg elevation seen under the polygon survey and at correct alt above takeoff.
```
            {
                "AMSLAltAboveTerrain": null,
                "Altitude": 50,
                "AltitudeMode": 1,
                "autoContinue": true,
                "command": 22,
                "doJumpId": 2,
                "frame": 3,
                "params": [
                    15,
                    0,
                    0,
                    null,
                    50.91853487,
                    16.75337956,
                    50
                ],
                "type": "SimpleItem"
            },
            {
                "TransectStyleComplexItem": {
                    "CameraCalc": {
                        "AdjustedFootprintFrontal": 22.03125,
                        "AdjustedFootprintSide": 14.687499999999998,
                        "CameraName": "Manual (no camera specs)",
                        "DistanceToSurface": 50,
                        "DistanceToSurfaceRelative": true,
                        "version": 1
                    },
                    "CameraShots": 28100,
                    "CameraTriggerInTurnAround": true,
                    "FollowTerrain": false,
                    "HoverAndCapture": false,
                    "Items": [
                        {
                            "autoContinue": true,
                            "command": 16,
                            "doJumpId": 3,
                            "frame": 3,
                            "params": [
                                0,
                                0,
                                0,
                                null,
                                50.878818400458165,
                                16.685102742687075,
                                377.147152194211
                            ],
                            "type": "SimpleItem"
                        },
```
### Lower Ground
valley

```
            {
                "AMSLAltAboveTerrain": null,
                "Altitude": 50,
                "AltitudeMode": 1,
                "autoContinue": true,
                "command": 22,
                "doJumpId": 2,
                "frame": 3,
                "params": [
                    15,
                    0,
                    0,
                    null,
                    50.86198616,
                    16.7080792,
                    50
                ],
                "type": "SimpleItem"
            },
            {
                "TransectStyleComplexItem": {
                    "CameraCalc": {
                        "AdjustedFootprintFrontal": 22.03125,
                        "AdjustedFootprintSide": 14.687499999999998,
                        "CameraName": "Manual (no camera specs)",
                        "DistanceToSurface": 50,
                        "DistanceToSurfaceRelative": true,
                        "version": 1
                    },
                    "CameraShots": 28100,
                    "CameraTriggerInTurnAround": true,
                    "FollowTerrain": false,
                    "HoverAndCapture": false,
                    "Items": [
                        {
                            "autoContinue": true,
                            "command": 16,
                            "doJumpId": 5,
                            "frame": 3,
                            "params": [
                                0,
                                0,
                                0,
                                null,
                                50.948412671101465,
                                16.746262568059883,
                                -471.0328311011905
                            ],
                            "type": "SimpleItem"
                        },
```

### Higher Ground on Master
master-hill

```
0 absoluteAltitude 154
1 absoluteAltitude 204
2 absoluteAltitude 204
```
rams the survey into the Sobotka hill
```
            {
                "AMSLAltAboveTerrain": null,
                "Altitude": 50,
                "AltitudeMode": 1,
                "autoContinue": true,
                "command": 22,
                "doJumpId": 2,
                "frame": 3,
                "params": [
                    15,
                    0,
                    0,
                    null,
                    50.92656236,
                    16.75926895,
                    50
                ],
                "type": "SimpleItem"
            },
            {
                "TransectStyleComplexItem": {
                    "CameraCalc": {
                        "AdjustedFootprintFrontal": 22.03125,
                        "AdjustedFootprintSide": 14.687499999999998,
                        "CameraName": "Manual (no camera specs)",
                        "DistanceToSurface": 50,
                        "DistanceToSurfaceRelative": true,
                        "version": 1
                    },
                    "CameraShots": 28100,
                    "CameraTriggerInTurnAround": true,
                    "FollowTerrain": false,
                    "HoverAndCapture": false,
                    "Items": [
                        {
                            "autoContinue": true,
                            "command": 16,
                            "doJumpId": 3,
                            "frame": 3,
                            "params": [
                                0,
                                0,
                                0,
                                null,
                                50.88015638818496,
                                16.68840304403717,
                                50 <- BOOM!
                            ],
                            "type": "SimpleItem"
                        },
```

### DIFF
--- a/src/MissionManager/TransectStyleComplexItem.cc
+++ b/src/MissionManager/TransectStyleComplexItem.cc
@@ -9,7 +9,7 @@
 
 #include "TransectStyleComplexItem.h"
 #include "JsonHelper.h"
-#include "MissionController.h"
+#include "PlanMasterController.h"
 #include "QGCGeo.h"
 #include "QGCQGeoCoordinate.h"
 #include "SettingsManager.h"
@@ -71,6 +71,7 @@ TransectStyleComplexItem::TransectStyleComplexItem(PlanMasterController* masterC
     connect(&_cameraTriggerInTurnAroundFact,            &Fact::valueChanged,            this, &TransectStyleComplexItem::_rebuildTransects);
     connect(_cameraCalc.adjustedFootprintSide(),        &Fact::valueChanged,            this, &TransectStyleComplexItem::_rebuildTransects);
     connect(_cameraCalc.adjustedFootprintFrontal(),     &Fact::valueChanged,            this, &TransectStyleComplexItem::_rebuildTransects);
+    connect(masterController->missionController(),      &MissionController::plannedHomePositionChanged, this, &TransectStyleComplexItem::_rebuildTransects);
 
     connect(&_turnAroundDistanceFact,                   &Fact::valueChanged,            this, &TransectStyleComplexItem::complexDistanceChanged);
     connect(&_hoverAndCaptureFact,                      &Fact::valueChanged,            this, &TransectStyleComplexItem::complexDistanceChanged);
@@ -355,24 +356,7 @@ void TransectStyleComplexItem::_rebuildTransects(void)
     }
 
     _rebuildTransectsPhase1();
-
-    if (_followTerrain) {
-        // Query the terrain data. Once available terrain heights will be calculated
-        _queryTransectsPathHeightInfo();
-    } else {
-        // Not following terrain, just add requested altitude to coords
-        double requestedAltitude = _cameraCalc.distanceToSurface()->rawValue().toDouble();
-
-        for (int i=0; i<_transects.count(); i++) {
-            QList<CoordInfo_t>& transect = _transects[i];
-
-            for (int j=0; j<transect.count(); j++) {
-                QGeoCoordinate& coord = transect[j].coord;
-
-                coord.setAltitude(requestedAltitude);
-            }
-        }
-    }
+    _queryTransectsPathHeightInfo();
 
     // Calc bounding cube
     double north = 0.0;
@@ -538,6 +522,36 @@ void TransectStyleComplexItem::_adjustTransectsForTerrain(void)
         }
 
         emit lastSequenceNumberChanged(lastSequenceNumber());
+    } else {
+        double requestedAltitude = std::numeric_limits<double>::min();
+        for (const auto& transect : _transectsPathHeightInfo) {
+            for (const auto& path : transect) {
+                for (const auto& height : path.heights) {
+                    requestedAltitude = std::max(height, requestedAltitude);
+                }
+            }
+        }
+        requestedAltitude += _cameraCalc.distanceToSurface()->rawValue().toDouble();
+        if (coordinateHasRelativeAltitude()) {
+            requestedAltitude -= masterController()->missionController()->plannedHomePosition().altitude();
+        }
+        for (int i=0; i<_transects.count(); i++) {
+            QList<CoordInfo_t>& transect = _transects[i];
+            for (int j=0; j<transect.count(); j++) {
+                QGeoCoordinate& coord = transect[j].coord;
+                coord.setAltitude(requestedAltitude);
+            }
+        }
+    }
+    if (_transects.count()) {
+        if (_transects.first().count()) {
+            _coordinate.setAltitude(_transects.first().first().coord.altitude());
+            emit coordinateChanged(coordinate());
+        }
+        if (_transects.last().count()) {
+            _coordinate.setAltitude(_transects.last().last().coord.altitude());
+            emit exitCoordinateChanged(exitCoordinate());
+        }
     }
 }
 
--- a/src/MissionManager/TransectStyleComplexItem.cc
+++ b/src/MissionManager/TransectStyleComplexItem.cc
@@ -71,7 +71,6 @@ TransectStyleComplexItem::TransectStyleComplexItem(PlanMasterController* masterC
     connect(&_cameraTriggerInTurnAroundFact,            &Fact::valueChanged,            this, &TransectStyleComplexItem::_rebuildTransects);
     connect(_cameraCalc.adjustedFootprintSide(),        &Fact::valueChanged,            this, &TransectStyleComplexItem::_rebuildTransects);
     connect(_cameraCalc.adjustedFootprintFrontal(),     &Fact::valueChanged,            this, &TransectStyleComplexItem::_rebuildTransects);
-    connect(masterController->missionController(),      &MissionController::plannedHomePositionChanged, this, &TransectStyleComplexItem::_rebuildTransects);
 
     connect(&_turnAroundDistanceFact,                   &Fact::valueChanged,            this, &TransectStyleComplexItem::complexDistanceChanged);
     connect(&_hoverAndCaptureFact,                      &Fact::valueChanged,            this, &TransectStyleComplexItem::complexDistanceChanged);
@@ -109,6 +108,7 @@ TransectStyleComplexItem::TransectStyleComplexItem(PlanMasterController* masterC
     connect(this,                                       &TransectStyleComplexItem::wizardModeChanged,           this, &TransectStyleComplexItem::readyForSaveStateChanged);
 
     connect(&_surveyAreaPolygon,                        &QGCMapPolygon::isValidChanged, this, &TransectStyleComplexItem::readyForSaveStateChanged);
+    connect(masterController->missionController(),      &MissionController::plannedHomePositionChanged, this, &TransectStyleComplexItem::_adjustTransectsForTerrain);
 
     setDirty(false);
 }
@@ -523,23 +523,38 @@ void TransectStyleComplexItem::_adjustTransectsForTerrain(void)
 
         emit lastSequenceNumberChanged(lastSequenceNumber());
     } else {
-        double requestedAltitude = std::numeric_limits<double>::min();
+        //We're not following terrain - this means that all the transects are on the same altitude
+        //(relative or absolute). Nominally, this altitude should position the vehicle at
+        //cameraCalc.distanceToSurface() from the ground, but the ground is uneven and we have
+        //to pick a representative spot/reference. We could pick:
+        //(1) The higest hill under the survey polygon for safety;
+        //(2) Average/median height under the polygon to minimise the extent where we
+        //    inevitably disobey distanceToSurface.
+        //(3) Choose the survey's centroid as reference, though it's unclear why that would
+        //    be better than any other option.
+        //We pick (2).
+        size_t noOfSamples = 0;
+        double avgAltitude = 0;
         for (const auto& transect : _transectsPathHeightInfo) {
             for (const auto& path : transect) {
                 for (const auto& height : path.heights) {
-                    requestedAltitude = std::max(height, requestedAltitude);
+                    avgAltitude += height;
+                    noOfSamples++;
                 }
             }
         }
-        requestedAltitude += _cameraCalc.distanceToSurface()->rawValue().toDouble();
-        if (coordinateHasRelativeAltitude()) {
-            requestedAltitude -= masterController()->missionController()->plannedHomePosition().altitude();
-        }
-        for (int i=0; i<_transects.count(); i++) {
-            QList<CoordInfo_t>& transect = _transects[i];
-            for (int j=0; j<transect.count(); j++) {
-                QGeoCoordinate& coord = transect[j].coord;
-                coord.setAltitude(requestedAltitude);
+        if (noOfSamples) {
+            avgAltitude /= noOfSamples;
+            avgAltitude += _cameraCalc.distanceToSurface()->rawValue().toDouble();
+            if (coordinateHasRelativeAltitude()) {
+                avgAltitude -= masterController()->missionController()->plannedHomePosition().altitude();
+            }
+            for (int i=0; i<_transects.count(); i++) {
+                QList<CoordInfo_t>& transect = _transects[i];
+                for (int j=0; j<transect.count(); j++) {
+                    QGeoCoordinate& coord = transect[j].coord;
+                    coord.setAltitude(avgAltitude);
+                }
             }
         }
     }


DonLakeFlyer:
Thanks @remster. I'll take a look at this. Take me a little bit. I'm in the middle of a bunch of stuff.

DonLakeFlyer:
"Average/median height under the polygon to minimise the extent where we inevitably disobey distanceToSurface."

Conceptually I think this is a good approach. It's more a matter of somehow making that known to the user which requires some more tweaking I think. I need to look back the the Survey UI again some more.

DonLakeFlyer:
I also think the planned home position altitude should not be used at all in the case of a camera based survey. The altitude in there may not be correct since the actual position for home is only truly set at flight time. It would be much better for camera based survey to always be written out in AMSL. Which is what you've been saying all along. But as you say likely requires a larger change. In reality I'd much rather make this correct and make sense than limit the change. Let me look through the QGC code again and see what it would take.

DonLakeFlyer:
I think what I'll try to do is clean up the whole distanceToSurfaceRelative thing in a separate pull. Without changing the fact that the surface itself is assumed to be at the AMSL height of planned home. Then once all that change is done you can add your average height for the polygon thing and use that as the surface AMSL height instead. And then after that I think it will still require some UI tweaking in yet another pull to make it more object to the user what is really happening. Right now the help text says things about distance to surface which explains nothing. 

Does that make sense?

remster:
Thanks you for giving it attention this quickly @DonLakeFlyer !
**Practical comments first:**
> I think what I'll try to do is clean up the whole distanceToSurfaceRelative thing in a separate pull.

Very glad to see we're on the same page. I went this way. [Landed with code](https://github.com/mavlink/qgroundcontrol/pull/8686/files) where `TrasectStyleComplexItem` grew the same [altitude properties](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/SimpleMissionItem.h#L36) as SimpleMissionItem, which made me think they should really move to `VisualMissionItem`. Interestingly, w/o throwing any functionality i removed more code than added. Please take a look - maybe you will find it useful.

Practically; I have reasonable time constraint on this - about a month. If you think you could land that pull beforehand all is good. But if you think you can't - would you consider merging this PR as a stopgap?

> Does that make sense?

Yes it does.

**Now for loose comments:**
> I also think the planned home position altitude should not be used at all in the case of a camera based survey.

You have a point here but it begs a question:. If "actual position for home is only truly set at flight time" why do we give it the user to manipulate at all? I can think of those two where it's valuable **to the user**:
- where the terrain model is unknown/unreachable or un-trusted;
- where the user wishes for the drone to climb this high from the assumed takeoff (because they know the building is this high) and either the user would feel responsible for taking off from that takeoff exactly or let QGC know to translate to AMSL internally so they don't have to feel responsible.

> It would be much better for camera based survey to always be written out in AMSL. Which is what you've been saying all along I believe.

If i recall correctly on this front I said:
- it beats me why would a single `.plan` have individual waypoints expressly using varying/interchanging altitude modes. Why wouldn't altitude mode be a pan-mission setting? 
- i did not succeed creating a `.plan` that uses anything else but relative alt - but i didn't try very hard.

So as far as I am concerned: it would be much better if the entire plan was written in AMSL (default), bearing in mind circumstances (no terrain model available) when it cannot be. Only the user can decide to put QGC into that "dont't trust terrain" mode and the UI should reflect it.

thank you once again.

DonLakeFlyer:
> [Landed with code](https://github.com/mavlink/qgroundcontrol/pull/8686/files) where `TrasectStyleComplexItem` grew the same [altitude properties](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/SimpleMissionItem.h#L36) as `SimpleMissionItem`

Been there done that. Then decided to throw it all out.

DonLakeFlyer:
> * it beats me why would a single `.plan` have individual waypoints expressly using varying/interchanging altitude modes. Why wouldn't altitude mode be a pan-mission setting?

Purely historical reasons and needing to support all of what mavlink can support. From a usability standpoint it's insanity.

DonLakeFlyer:
I you want to play around with WIP:
[DistanceToSurface.patch.txt](https://github.com/mavlink/qgroundcontrol/files/4519361/DistanceToSurface.patch.txt)

DonLakeFlyer:
I should have it finished up by tomorrow.

DonLakeFlyer:
FYI @antiheavy

So the problem with this change is that it requires terrain information to calc the average surface height. Which in turn means you need an internet connection if you don't have the terrain tiles for the area you are working in already cached. It's certainly way better to do this. But if you don't have the tiles you're screwed.

It seems like it's going to need some sort of workaround mechanism if you are stuck without internet and still really need to generate a survey on the fly. I'm thinking a hamburger workaround which allows you to manually specify the AMSL alt of the surface you plan to survey.

I really don't want transect planning to get back into the whole Altitude Mode thing to solve this problem.

Antiheavy:
> > * it beats me why would a single `.plan` have individual waypoints expressly using varying/interchanging altitude modes. Why wouldn't altitude mode be a pan-mission setting?
> 
> Purely historical reasons and needing to support all of what mavlink can support. From a usability standpoint it's insanity.

yeah, would be better/safer IMO to force missions to follow only one altitude system throughout the mission.

DonLakeFlyer:
Another question is what to rename the "Altitude" field to. It isn't really altitude. It's the distance above the surface your are surveying. This is true whether you are using follow terrain or not now. Structure Scan uses "Scan Distance".

src/MissionManager/TransectStyleComplexItem.cc
```diff
@@ -538,6 +522,51 @@ void TransectStyleComplexItem::_adjustTransectsForTerrain(void)
...
 
         emit lastSequenceNumberChanged(lastSequenceNumber());
+    } else {
+        //We're not following terrain - this means that all the transects are on the same altitude
+        //(relative or absolute). Nominally, this altitude should position the vehicle at
+        //cameraCalc.distanceToSurface() from the ground, but the ground is uneven and we have
+        //to pick a representative spot/reference. We could pick:
+        //(1) The higest hill under the survey polygon for safety;
+        //(2) Average/median height under the polygon to minimise the extent where we
+        //    inevitably disobey distanceToSurface.
```
Antiheavy, line 532: I'm envisioning trying to explain this logic to the average user (at least our users) and I think their heads would explode, (except that maybe they wouldn't because I don't think I can even figure out how to explain it in a way they'd even start to understand).  Doesn't this whole approach makes the transect widths completely unpredictable?  I feel like even an advanced user might want some control over what the altitude reference datum would be.  Maybe I'm missing something.

DonLakeFlyer: I don't think you'd ever need to explain it to anyone. It will just do something reasonable even when you are not flying over a billiard table.The point is that if you are trying to hit a specific GSD for your images. You need to know the height of the surface you are surveying and then you need to fly a distance above that surface to get the specified GSD. If you are not flying over a billiard table using a relative altitude above the home position will not work since you distance above the surface will be incorrect. If you are flying over a billiard table then it works excatly like it did before.

DonLakeFlyer: > If you are flying over a billiard table then it works exactly like it did before.

That is as long as you have decent quality terrain data.

DonLakeFlyer: The problem with the current code comes about when the altitude of the home position differs from the surface altitude of the place you are surveying.

DonLakeFlyer: With the old code if there was a big enough difference between the alt of the home position and the general alt of the survey area there was no way to set up a survey with GSD or overlap settings that would work at all.

Antiheavy: > Another question is what to rename the Survey/Corridor Scan "Altitude" field to. It isn't really altitude.

Maybe "Height" ?

Antiheavy:
> Another question is what to rename the Survey/Corridor Scan "Altitude" field to. It isn't really altitude.

Maybe "Height"?

src/MissionManager/TransectStyleComplexItem.cc
Antiheavy: Yep, that all makes sense.  My worry here is that even small tweaks to the polygon might cause the "average/median" calculation to change unpredictably which might be confusing.  Maybe I'm over-thinking it a bit.

Antiheavy:
This may be a separate issue needing a separate PR, but my biggest concern is accurate warning and prevention of a plan with terrain intersections within a survey area (including turnarounds).  The current altitude strip indicator doesn't calculate this correctly.  We've seen users fly into the sides of hills due to this.
image


Maybe this PR is putting in place some of the pieces to be used for more correct terrain warnings?  Thinking towards this issue: https://github.com/mavlink/qgroundcontrol/issues/8560

Antiheavy:
> > Another question is what to rename the Survey/Corridor Scan "Altitude" field to. It isn't really altitude.
> 
> Maybe "Height"?

@DonLakeFlyer thinking about this more, maybe you'd want to show both somehow.  It is important that "altitude" be shown and it should be the value that the drone uses to fly (rel, AMSL, AGL, whatever).  There could be 2nd value in displaying a height above ground ("AGL" maybe) as a reference.  Or in case of a Survey or Corridor the average height (with this PR).  Lots of other drone systems I've used in the past have shown multiple altitude references, usually the primary was more prominent.  It's a struggle to make it simple though.

remster:
> This may be a separate issue needing a separate PR, but my biggest concern is accurate warning and prevention of a plan with terrain intersections within a survey area (including turnarounds). The current altitude strip indicator doesn't calculate this correctly. We've seen users fly into the sides of hills due to this.
> image
> 
> Maybe this PR is putting in place some of the pieces to be used for more correct terrain warnings? Thinking towards this issue: #8560
> 
> More on topic to this PR, I fear the "average" calculation method can still cause users to fly into terrain with a false sense of security.

I did stop to think about this. `TransectStyleComplexItem` is a `VisualMissionItem` and so it has `terrainAltitude`. This altitude is just set to whatever `coordinate` returns (which is the survey entry). `TransectStyleComplexItem` should adjust its `terrainAltitude` to the highest point under the survey polygon.

remster:
@DonLakeFlyer @Antiheavy what do you think of this: 
> Because of this (which i think is dangerous to keep) this PR and all function related to terrain (QGC key datapoint) are very difficult to test. How about:

    if (qgcApp()->runningUnitTests()) {
         QList<double>() result;
         for (auto const& point: coordinates) {
             result.push_back(
                std::round((point.longitude - std::round(longitude))*100)
             );
         }
        return;
    }

DonLakeFlyer:
@remster So Todd and I have been having a discussion on this. Still agree that current code is wrong. One question is for you why using Follow Terrain doesn't work for you scenarios? If you set the tolerances high on a follow terrain survey you'll basically end up with the same thing.

DonLakeFlyer:
I made a note of the unit test thing and I'll take a look at it.

DonLakeFlyer:
The more I play around with this the more I think there is no need a new average height of the survey area option. 

I think th should be two options (which essentially exist today already):
1) Assume Level Ground yes - This is basically how the current code works. There still needs to be UI changes to fix various things relating to distance to surface and manual grids. But internally it's functionally the same.
2) Assume Level Ground no - This is terrain follow as currently exists. It fully supports the survey area being above or below the home position. If the survey area is generally flat but just at a different altitude then you get the same thing as you would get by using an average terrain height for the surface height. If you end up getting more intertitial points than you want added you can up the tolerence. But if you're getting intertitials at a decent tolerance it means average surface height wouldn't have worked anyway.

I did find a bug in the terrain follow tolerance code. This would cause interstiatial points to be left in even they didn't exceed tolerance. With that fix terrain follow surveys over relatively flat areas look the same as what and average surface height survey would generate.

remster:
@DonLakeFlyer 
So i would need to try (which i shortly will) whether "following terrain" will work for me - i need to learn how the levers work). I will get back on that before Monday. It sounds like if the levers have enough headroom I can indeed land with what i need. Though I have to reiterate, i never used AMSL.

But I am having cognitive dissonance between:
> Still agree that current code is wrong.

and
> Assume Level Ground:Yes - This is the default to mimic existing behavior. This is basically how the current code works. 

It would make sense to me if you suggested that all surveys should follow terrain - just should be more or less relaxed about this. Essentially if one unlicks "follow terrain" the UI simply relaxes the parameters to the maximum degree. And you can remove the `!_followTerrain` path entirely from the code. 

I understand this path only works if - by chance - most of terrain under survey is at the same elevation as home. There must be something you're not telling me. What does "Assume Level Ground:Yes" mean?

DonLakeFlyer:
> What does "Assume Level Ground:Yes" mean?

It means that your entire mission takes place over mostly the same level of terrain altitude. From takeoff to survey and back.

> It would make sense to me if you suggested that all surveys should follow terrain

I don't necessarily disagree with that. The biggest problem is that is not what current users expect. Most other systems don't work this way.

DonLakeFlyer:
The easiest way to explain it is that if the area you are working with is level ground then using relative altitudes throughout the survey pattern will yield you the highest quality results. Once you throw into the mix using any sort of terrain data for calculations then you are at the whim of the quality of that data. If that is your only choice then it's the best you can do. But if you don't need to do that you are better off not to.

Antiheavy:
I still think there should be 3 high level altitude options a user should select for planning a whole mission (not per item):

- **Relative** : height above home by default, what most people are used to for better or worse, doesn't require terrain (although it is nice for warnings/safety)
- **AGL** : required terrain, internally MSL calculated by terrain + AGL, I think(?) this option would solve @remster 's goal, special exception for landing patterns (we'd probably switch to this by default in our system if it was implemented right)
- **MSL** : for the rare users of high flyers that play with the big boys

I'm not trying to derail the discussion on this specific PR, but I think giving up on allowing each item to use it's own altitude reference solves half the problem here.

DonLakeFlyer:
> but I think giving up on allowing each item to use it's own altitude reference solves half the problem here.

Only if you say that you cannot create a camera based survey in relative mode. Since there is simply no way to support that other than over a billiard table surface.

remster:
@DonLakeFlyer I am sorry, when you wrote "Assume Level Ground" I somehow read it as "Assume Ground Level" - the pitfalls of throwing syntax to aid language's expression ;)

> Once you throw into the mix using any sort of terrain data for calculations then you are at the whim of the quality of that data

Yes, i did remark earlier "where the terrain model is unknown/unreachable, un-trusted" so I have to agree with you here. But:
- calling "Assume Level Ground" as "Follow Terrain" is at least ambiguous to the user. I now understand that: to not follow terrain means (for locally good reasons) to distrust terrain model and not use it **at all**.
- QGC still has to accurately warn the user operating the "Assume Level Ground" mode when the ground doesn't appear quite as level through the lens of the available terrain model. This means using "max height below the survey polygon" in the terrain profile in the bottom-left and not just terrain at survey entry.
- QGC still has to get rid of 'distanceToSurfaceRelative` from the code one day - Switching "follow terrain" **on** forces AMSL (not strictly necessary, but at least removes noise). Switching "follow terrain" **off** allows the user to operate in relative or AMSL references, the latter of which is what they're expressly avoiding. So when they switch "follow terrain" they should be implicitly and exclusively in relative. This brushes onto what @Antiheavy said on pan mission altitude model. If the user can trust terrain in one bit of the mission, why would they stop trusting it another? I think then AMSL, AGL and Relative should be a mission setting that influence [defaults](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/SimpleMissionItem.cc#L110) of all mission items within it.

DonLakeFlyer:
My current take on this is:
* Leave things as is with respect to how it currently works if you are not using terrain. As far as I understand this is exactly how other planners work as well like Mission Mapper, DroneDeploy. Although most don't support a follow terrain option like QGC odes.
* The docs are being updated to explain how non-terrain surveys work: https://github.com/mavlink/qgc-user-guide/pull/263. Explaining that terrain follow should be used if your area is non-level.
* I have documented in the code the strangeness around distanceToSurfaceRelative being just an expediency in keep the code simple. As well as how CameraCalc is still used to persist manual grids and what fields are used. But in reality it has no relationship to camera calculations since it is only used for manual grids.

> * Switching "follow terrain" **on** forces AMSL (not strictly necessary, but at least removes noise).

Yes it is strictly necessary. You can't follow terrain using relative altitudes based on a planned home position altitude. The position you specified when you planned the mission not connected to the vehicle could be different from when you fly the mission which in turns means the altitude for it can be different as well. Which then trashes all your altitudes. A boundary case sure, but real.

> * Switching "follow terrain" **off** allows the user to operate in relative or AMSL references

Not really. Operating in relative or absolute is only allowed for manual grids. Non-terrain camera based surveys are always relative and cannot be changed.

> * So when they switch "follow terrain" **off** they should be implicitly and exclusively in relative.

If you are using a manual grid, when you switch off follow terrain you will be initially set to relative altitude. But after that you can change it to absolute. When you are flying a manual grid you are truly specifying an altitude. You are not specifying a distance to surface. Hence just like all other waypoint altitude setting you can do what you want with it.

DonLakeFlyer:
> Shall i close this PR now? Has been serving as a good discussion board..

I think almost. I want to make sure I'm still not missing something important.

Antiheavy:
>     * If the user can trust terrain in one bit of the mission, why would they stop trusting it in another?

exactly

remster:
> You can't follow terrain using relative altitudes based on a planned home position altitude. The position you specified when you planned the mission not connected to the vehicle could be different from when you fly the mission

Doesn't this argument hold for every mission that uses a relative altitude? I.e.: you better take off from where you said you would or hope the ground is level. But then - if the ground is level you also don't have to worry about following terrain in either altitude mode. There is still some conflation in the air - but I will move on.

I will now test "follow terrain". Having chatted with my colleague, I fear our reference drone has problems flying AMSL so i may find myself in a pickle. Hopefully I will know more early next week. Thank you. I will close this PR on a 48 hour work-day inactivity.

DonLakeFlyer:
> Doesn't this argument hold for every mission that uses relative altitude?

Yes it does. You better know what you are doing if you are using relative altitudes and flying over uneven ground.

DonLakeFlyer:
> I fear our reference drone has problems flying AMSL so i may find myself in a pickle.

If that is the case then nothing which uses any terrain data will work. Including your original average height proposal. Your only choice would be to use a manual grid and set the altitude manually. You can use the camera calc at first to get calculations for a level ground situation. Then if you switch to manual grid the spacing and trigger distance will keep the previous values. Then you adjust the altitude of the grid manually based on a relative altitude that is appropriate to get the distance to surface you want.

remster:
@DonLakeFlyer 
Yes. When we give this to our drone:
```
{
    "fileType": "Plan",
    "geoFence": {
        "circles": [
        ],
        "polygons": [
        ],
        "version": 2
    },
    "groundStation": "QGroundControl",
    "mission": {
        "cruiseSpeed": 15,
        "firmwareType": 12,
        "hoverSpeed": 5,
        "items": [
            {
                "AMSLAltAboveTerrain": null,
                "Altitude": 373.38,
                "AltitudeMode": 2,
                "autoContinue": true,
                "command": 22,
                "doJumpId": 1,
                "frame": 0,
                "params": [
                    15,
                    0,
                    0,
                    null,
                    4x.312000684350494,
                    -79.74213822415534,
                    373.38
                ],
                "type": "SimpleItem"
            },
            {
                "AMSLAltAboveTerrain": null,
                "Altitude": 373.38,
                "AltitudeMode": 2,
                "autoContinue": true,
                "command": 16,
                "doJumpId": 2,
                "frame": 0,
                "params": [
                    0,
                    0,
                    0,
                    null,
                    4x.31177524,
                    -79.74126501,
                    373.38
                ],
                "type": "SimpleItem"
            }
        ],
        "plannedHomePosition": [
            4x.312000684350494,
            -79.74213822415534,
            366
        ],
        "vehicleType": 2,
        "version": 2
    },
    "rallyPoints": {
        "points": [
        ],
        "version": 2
    },
    "version": 1
}
```
It just shoots out to the sky and climbs 366 above takeoff and not, as we hoped, 366m AMSL (7m above takeoff). 

I need to engage with the vendor.

With what MAVLINK command does QGC communicate the altitude reference and where (in the code) does do it?
I understand the .plan carries it in:
"AltitudeMode": 2,
and 
"frame": 0,

I can't find the bit in the [spec](https://mavlink.io/en/messages/common.html#MAV_FRAME) that I our drone disobeys.

DonLakeFlyer:
TransectStyleComplexItem::appendMissionItems is what builds the actual mission commands. They are just MAV_CMD_NAV_WAYPOINT commands with the MAV_FRAME set appropriately. First step is to send the survey to the vehicle and then download it back and check that the altitudes make sense in there. If they do then there is something wrong with your vehicle. You can also output to KML to visual a 3d terrain follow survey.

remster:
Before i do the survey, i need to iron the SimpleMissionItem and have the drone just fly to that waypoint using AMSL.

DonLakeFlyer:
> 1. I understand the .plan carries it in:
>    ["AltitudeMode": 2](https://github.com/mavlink/qgroundcontrol/blob/master/src/QmlControls/QGroundControlQmlGlobal.h#L59),
>    and
>    ["frame": 0](https://mavlink.io/en/messages/common.html#MAV_FRAME_GLOBAL),

That all looks correct. For AMSL

remster:
> > 1. I understand the .plan carries it in:
> >    ["AltitudeMode": 2](https://github.com/mavlink/qgroundcontrol/blob/master/src/QmlControls/QGroundControlQmlGlobal.h#L59),
> >    and
> >    ["frame": 0](https://mavlink.io/en/messages/common.html#MAV_FRAME_GLOBAL),
> 
> That all looks correct. For AMSL

That;s what i figure, but this is `.plan` and not MAVLINK. How does the drone learns to interpret `MAV_CMD_NAV_WAYPOINT` elevation as AMSL (and not rel)? The [message spec](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_WAYPOINT) doesn't leave room for alt reference.

DonLakeFlyer:
Mission items are COMMAND_LONG/INT

PR #8686, opened 2020-04-22
WIP/Experiment only - calculate distanceToSurface

remster:
this is the first pass draft attempt to addressing https://github.com/mavlink/qgroundcontrol/issues/8273
Sharing just to show @DonLakeFlyer  how far i got on this path.


DonLakeFlyer:
Thanks. I have the first pass mostly working already. I'll likely get it in tonight/tomorrow. I'm taking a simpler approach where the altitude is always just simple distance to surface. I decided that support altitude modes for manual grids just isn't worth the additional complexity. Everything is distance to surface and that surface AMSL height is then your code for polygon average. So the new code will have a surfaceAMSL value which you can then plug your calced value into.

PR #8689, opened 2020-04-24
Plan - Terrain Follow: Fix tolerance adjustment step

DonLakeFlyer:
* Tolerance adjustment step was leaving unecessary waypoints within transects
* Plus a bunch of smaller visual tweaks


arnaudTilak:
Why did the follow terrain disappeared ?

Issue #2368, opened 2015-12-07
Map zoom level should go higher


DonLakeFlyer:
If you use Mission Planner or Tower you'll see that they allow you to zoom higher past the level where there are tiles available. They just blow up the last tile available to a larger size. This makes sense since the last zoom level available if too low for accurate positioning. Not sure exactly how high to go. Lorenz may have an idea of that in relationship to GPS accuracy.

LorenzMeier:
The three related issues are closed, closing as done.

PR #8690, opened 2020-04-24
QtLocationPlugin: Maximize the map to 21.

muramura:
I use Google's satellite maps.
I tend to fly my multicopters in areas where the area is small.
Google supports up to ZOOM size 21.
I would change the maximum from 20 to 21.

Screenshot from 2020-04-24 19-02-16

### DIFF
--- a/src/QtLocationPlugin/QGCMapUrlEngine.h
+++ b/src/QtLocationPlugin/QGCMapUrlEngine.h
@@ -24,7 +24,7 @@
 #include "MapboxMapProvider.h"
 #include "ElevationMapProvider.h"
 
-#define MAX_MAP_ZOOM (20.0)
+#define MAX_MAP_ZOOM (21.0)
 
 class UrlFactory : public QObject {
     Q_OBJECT


DonLakeFlyer:
If I remember correctly the issue here is that not all map providers support up to 21. Hence you can't just change the max. @dogmaphobic Knows more

patrickelectric:
From: https://github.com/mavlink/qgroundcontrol/issues/2368

patrickelectric:
I have went to places where "20" is not available in QGC, so I don't see why not to support the maximum zoom available in Google.

DonLakeFlyer:
> so I don't see why not to support the maximum zoom available in Google.

Because someone needs to implement it correctly and only @dogmaphobic knows how to do that. This has come up before and just changing this constant doesn't work as far as I remember.

DonLakeFlyer:
Also #2368 is a completely different feature request.

patrickelectric:
@dongfang this is at least a helper for the same thing, we could have real zoom over stretch for some areas.

muramura:
@DonLakeFlyer san, @patrickelectric san. Comments Thanks.

I want to incorporate VWorld's zoom processing into other map provider processing.
https://github.com/mavlink/qgroundcontrol/blob/master/src/QtLocationPlugin/GenericMapProvider.cpp#L66.
I learned that the ZOOM specification of the map is influenced by "cameraCaps.setMaximumZoomLevel".
I've found that each map provider has its own characteristics and scales differently.
I don't think all map-providers support 20.0 for "MAX_MAP_ZOOM".

DonLakeFlyer:
> Comments Thanks.

This requires @dogmaphobic who is crazuy busy right now so expect it to take a while. I don't know anything about the map internals.

jaxxzer:
I tried this in the past: https://github.com/mavlink/qgroundcontrol/pull/4662

DonLakeFlyer:
The two options here are:
1) Allow all maps to go to the highest zoom level possible for any map. Even if the tiles will show empty for some map types.
2) Manually generate the tiles for the missing zoom levels by just blowing them and stuffing them in. This would allow you go a few steps beyond even what the max level across all systems is. Which would still be useful to position things.

I've always wanted to support 2. But unfortunately I don't know how to do either one since I know next to nothing about this code.

PR #8692, opened 2020-04-27
Introduce weight unit and separate distance unit from altitude

remster:
Sooner or later drones will start carrying payload and this PR introduces units of weight.
It also separates distance units from altitude units as this is often a practice (e.g.: US military always measures altitude in feet, but often measure distance in "km" - we've all seen in Hollywood movies when they say "clicks" - it's true).

### DIFF
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -32,6 +32,11 @@ const qreal FactMetaData::UnitConsts_s::milesToMeters =         1609.344;
 const qreal FactMetaData::UnitConsts_s::feetToMeters =          0.3048;
 const qreal FactMetaData::UnitConsts_s::inchesToCentimeters =   2.54;
 
+//Weight
+const qreal FactMetaData::UnitConsts_s::ouncesToGrams =         28.3495;
+const qreal FactMetaData::UnitConsts_s::poundsToGrams =         453.592;
+
+
 static const char* kDefaultCategory = QT_TRANSLATE_NOOP("FactMetaData", "Other");
 static const char* kDefaultGroup    = QT_TRANSLATE_NOOP("FactMetaData", "Misc");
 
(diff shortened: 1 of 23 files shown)


src/FactSystem/FactMetaData.cc
```diff
@@ -51,24 +56,31 @@ const FactMetaData::AppSettingsTranslation_s FactMetaData::_rgAppSettingsTransla
...
+    { "alt m",  "m",        FactMetaData::UnitAltitude,    UnitsSettings::AltitudeUnitsMeters,         FactMetaData::_defaultTranslator,                   FactMetaData::_defaultTranslator },
     { "cm/px",  "cm/px",    FactMetaData::UnitDistance,    UnitsSettings::DistanceUnitsMeters,         FactMetaData::_defaultTranslator,                   FactMetaData::_defaultTranslator },
     { "m/s",    "m/s",      FactMetaData::UnitSpeed,       UnitsSettings::SpeedUnitsMetersPerSecond,   FactMetaData::_defaultTranslator,                   FactMetaData::_defaultTranslator },
     { "C",      "C",        FactMetaData::UnitTemperature, UnitsSettings::TemperatureUnitsCelsius,     FactMetaData::_defaultTranslator,                   FactMetaData::_defaultTranslator },
-    { "m^2",    "m^2",      FactMetaData::UnitArea,        UnitsSettings::AreaUnitsSquareMeters,       FactMetaData::_defaultTranslator,                   FactMetaData::_defaultTranslator },
+    { "m^2",    "m\u00B2",  FactMetaData::UnitArea,        UnitsSettings::AreaUnitsSquareMeters,       FactMetaData::_defaultTranslator,                   FactMetaData::_defaultTranslator },
```
remster, line 64: while at it change to superscript - in case this isn't a favour, i will remove.

src/QmlControls/QGroundControlQmlGlobal.h
```diff
@@ -145,6 +145,22 @@ class QGroundControlQmlGlobal : public QGCTool
 
     QString appSettingsDistanceUnitsString(void) const { return FactMetaData::appSettingsDistanceUnitsString(); }
 
+    /// Converts from meters to the user specified distance unit
+    Q_INVOKABLE QVariant metersToAppSettingsAltitudeUnits(const QVariant& meters) const { return FactMetaData::metersToAppSettingsAltitudeUnits(meters); }
```
remster, line 149: This is becoming ugly and turning `QGroundControlQmlGlobal.h` into a blob. This is needed though when QML calculates altitudes (as opposed to showing values stored as `Facts`) - I think it has every privilege to want to do so and it is correct to supply QML with functions like these. But maybe it's time to wrap them into a `UnitCoversion` class that `QGroundControlQmlGlobal` exposes.

src/FactSystem/FactMetaData.cc
DonLakeFlyer: @dogmaphobic Do you know anything about Qt Fonts. Will this cause problems?
@remster do you have screenshots of what this looks like?

DonLakeFlyer:
So I would change UnitDistance to UnitHorizontalDistance and UnitAltitude to UnitVerticalDistance. I think that is clearer to what they are for.

src/FactSystem/FactMetaData.cc
DonLakeFlyer: Is it possible somehow to have the table only be horizontal distance units. And then the vertical distance variants for all of these are generated automagically by prepending "vertical " to the front of the horizontal distance ones? There needs to be vertical variants for all the horizontal ones not just meters.

remster: > @remster do you have screenshots of what this looks like?
IS:
superscript
WAS:
... (coming up)

DonLakeFlyer: Ah, super nice. Definitely want that if there any strange issues.

remster: So i understand what you want here:
> Is it possible somehow to have the table only be horizontal distance units. And then the vertical distance variants for all of these are generated automagically by prepending "vertical " to the front of the horizontal distance ones?

but do not here: 
> There needs to be vertical variants for all the horizontal ones not just meters.

Do you mean, on top of "m" also: "meter", "meters" - I don't understand why these raw extra units are there at all (neglecting the British spelling i might add) as "raw units" aren't (I may be wrong) presented to the user?
Or do you somehow also mean speed and area? Please just be more specific, I'll comply.

src/QmlControls/QGroundControlQmlGlobal.h
DonLakeFlyer: I don't disagree. Will need to pull the string on usage of it through all the Qml.

src/FactSystem/FactMetaData.cc
remster: > Ah, super nice. Definitely want that if there aren't any strange issues.

That I don't know how to go around verifying. Any hints how these cooked units are localised?

> There is on language which uses a different font, maybe Korean. Wonder if that will works as well. These are localized yet, but at some point we'll get around to supporting that.

I can do you Polish :)

src/MissionManager/KMLPlanDomDocument.cc
```diff
@@ -84,8 +84,8 @@ void KMLPlanDomDocument::_addFlightPath(Vehicle* vehicle, QList<MissionItem*> rg
                 QString htmlString;
                 htmlString += QStringLiteral("Index: %1\n").arg(item->sequenceNumber());
                 htmlString += uiInfo->friendlyName() + "\n";
-                htmlString += QStringLiteral("Alt AMSL: %1 %2\n").arg(QString::number(FactMetaData::metersToAppSettingsDistanceUnits(coord.altitude()).toDouble(), 'f', 2)).arg(FactMetaData::appSettingsDistanceUnitsString());
-                htmlString += QStringLiteral("Alt Rel: %1 %2\n").arg(QString::number(FactMetaData::metersToAppSettingsDistanceUnits(coord.altitude() - homeCoord.altitude()).toDouble(), 'f', 2)).arg(FactMetaData::appSettingsDistanceUnitsString());
+                htmlString += QStringLiteral("Alt AMSL: %1 %2\n").arg(QString::number(FactMetaData::metersToAppSettingsHorizontalDistanceUnits(coord.altitude()).toDouble(), 'f', 2)).arg(FactMetaData::appSettingsHorizontalDistanceUnitsString());
```
remster, line 87: Please note, I have so far introduced the facility to use a "vertical metre", but i am not yet surfacing it in the UI. I may extend this PR to do so, but this may be quite involved hunting all those `Facts` that use "m" for altitude.

remster:
> So I would change UnitDistance to UnitHorizontalDistance and UnitAltitude to UnitVerticalDistance. I think that is clearer to what they are for. And use the "Vertical" terminology in the method names.

Done: https://github.com/mavlink/qgroundcontrol/pull/8692/commits/1e98a7abc8da7f14fb348880ad43b4435739413a

src/QmlControls/QGroundControlQmlGlobal.h
remster: Since you asked me to rename existing methods (e.g.: ["distance" to "horizontalDistance"](https://github.com/mavlink/qgroundcontrol/pull/8692/commits/1e98a7abc8da7f14fb348880ad43b4435739413a#diff-b4d15c63e4db01b93ae2cb70e47602d1R64)) I had to update Qml so i did take the liberty of wrapping all those unit methods in `QGroundControlQmlGlobal` into [QmlUnitsConversion](https://github.com/mavlink/qgroundcontrol/pull/8692/commits/1e98a7abc8da7f14fb348880ad43b4435739413a#diff-d0673baf68efbb0ab1f10f243c8dce5aR1) and diligently chased all qml references.

src/FactSystem/FactMetaData.cc
DonLakeFlyer: > Any hints how these cooked units are localised?

They aren't currently. If you look in QGCApplication.cc you'll see that Korean has a different font for some reason. Could you hack your build and verify that font works with these as well?

DonLakeFlyer:
@hamishwillee This addition of the concept of "vertical" units should affect the mavlink spec as well. Fields which are vertical distances should use these new unit names so that conversion can work correctly.

src/FactSystem/FactMetaData.cc
remster: There you go @DonLakeFlyer 
korean

DonLakeFlyer: Perfect. Seems fine then.

remster:
@DonLakeFlyer just a reminder on this that we have this issue unresolved:
> There needs to be vertical variants for all the horizontal ones not just meters.

DonLakeFlyer:
@remster Can you enter an issue for carrying across vertical support for all the horizontal units.

PR #8702, opened 2020-04-29
Allow to move/reorder objects within QmlObjectListModel

akontsevich:
Often necessary to move/reorder objects within the list for example in ListView dragging, waypoints reordering, etc etc. And `move()` method should be added and organized properly to allow above and work correctly.

### DIFF
--- a/src/QmlControls/QmlObjectListModel.cc
+++ b/src/QmlControls/QmlObjectListModel.cc
@@ -125,6 +125,22 @@ bool QmlObjectListModel::removeRows(int position, int rows, const QModelIndex& p
     return true;
 }
 
+void QmlObjectListModel::move(int from, int to)
+{
+    if(0 <= from && from < count() && 0 <= to && to < count() && from != to) {
+        // Workaround to allow move item to the bottom. Done according to
+        // beginMoveRows() documentation and implementation specificity:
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Is there way you get your editor to not futz with whitespace unless you really changed something? Lots of unneeded whitespace changes here.

DonLakeFlyer:
I'm guessing there is something in Qt Creator which puts that sort of spacing in. But don't want to fight back and forth in commits removing then and then other commits they show back up.

akontsevich:
> I'm guessing there is something in Qt Creator which puts that sort of spacing in. But don't want to fight back and forth in commits removing then and then other commits they show back up.

There is an option to clear unnecessary white-spaces in Qt Creator:

image

I think it is a good practice to have it turned on for all. But ok, I removed that changes from the commit: https://github.com/mavlink/qgroundcontrol/pull/8702/commits/ea2130691ac7ea2a6d8c97f5c3be91c8f7a7fefd

DonLakeFlyer:
> I think it is a good practice to have it turned on for all.

It's already next to impossible to get people to read the docs on how to build fully before they report something wrong. Adding yet another step they need to do just makes things harder.

DonLakeFlyer:
Thanks for the pull

akontsevich:
Thank much for acceptance and the merge!!!

Issue #8700, opened 2020-04-29
Update custom-example


akontsevich:
Need to update `custom-example` to recent API changes as seems it is outdated.

## Expected Behavior
Build and run `/custom` successfully.

## Current Behavior
Build error:

`.../qgroundcontrol/custom/src/AutoPilotPlugin/CustomAutoPilotPlugin.cc:47:32: error: error: ‘class Vehicle’ has no member named ‘hilMode’
`
## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1. rename `custom-example` directory to `custom` before running `qmake`
2.  delete the build directory before running `qmake`
3.  build in Qt Creator

## System Information
When posting bug reports, include the following information
- Operating System: openSUSE Tumbleweed
- QGC Version: 9.3.1
- QGC build: 20200406 [revision 6db837a5288ee3ca5ec504fbd5a765817e556ac2] (SUSE Linux)

akontsevich:
Thanks @DonLakeFlyer for the update. Builds and works fine for me now. 

However it is not clear now how to create custom MainWindow now: https://github.com/mavlink/qgroundcontrol/pull/8709/files#r418476063 - any clarification? Thanks!

akontsevich:
> Thanks @DonLakeFlyer for the update. Builds and works fine for me now.
> 
> However it is not clear now how to create custom MainWindow now: https://github.com/mavlink/qgroundcontrol/pull/8709/files#r418476063 - any clarification? Thanks!

Got it: `createRootWindow()` still virtual.

PR #8709, opened 2020-04-30
Major update to simplify and document the custom build example

DonLakeFlyer:


DonLakeFlyer:
Related to #8700

custom-example/src/CustomPlugin.cc
```diff
@@ -130,118 +107,66 @@ QVariantList&
...
-QQmlApplicationEngine*
-CustomPlugin::createRootWindow(QObject *parent)
-{
-    QQmlApplicationEngine* pEngine = new QQmlApplicationEngine(parent);
-    pEngine->addImportPath("qrc:/qml");
-    pEngine->addImportPath("qrc:/Custom/Widgets");
-    pEngine->addImportPath("qrc:/Custom/Camera");
-    pEngine->rootContext()->setContextProperty("joystickManager",   qgcApp()->toolbox()->joystickManager());
-    pEngine->rootContext()->setContextProperty("debugMessageModel", AppMessages::getModel());
-    pEngine->load(QUrl(QStringLiteral("qrc:/qml/MainRootWindow.qml")));
```
akontsevich, line 215: How to create custom MainWindow now? Any example? Thanks!

src/api/QGCCorePlugin.cc
```diff
@@ -435,14 +435,18 @@ void QGCCorePlugin::instrumentValueAreaCreateDefaultSettings(const QString& defa
...
+    QQmlApplicationEngine* qmlEngine = new QQmlApplicationEngine(parent);
+    qmlEngine->addImportPath("qrc:/qml");
+    qmlEngine->rootContext()->setContextProperty("joystickManager", qgcApp()->toolbox()->joystickManager());
+    qmlEngine->rootContext()->setContextProperty("debugMessageModel", AppMessages::getModel());
+    return qmlEngine;
+}
+
+void QGCCorePlugin::createRootWindow(QQmlApplicationEngine* qmlEngine)
+{
+    qmlEngine->load(QUrl(QStringLiteral("qrc:/qml/MainRootWindow.qml")));
```
akontsevich, line 449: Hmm, what if I want to change this URL like it was possible in previous example? Could You fix this please to allow custom MainWindow as well? I was good in previous version to specify our own MainWindow. Or it is still possible by overriding `createRootWindow()`? 

Ok, seems it is still possible as it is `virtual`, thanks!

custom-example/src/CustomPlugin.cc
akontsevich: Go it! NVm

custom-example/README.md
```diff
@@ -2,27 +2,26 @@
...
 
-```
-DEFINES += DISABLE_BUILTIN_ANDROID
-```
+* Assumes an "Off The Shelf" purchased commercial vehicle. This means most vehicle setup is hidden from the user since they should mostly never need to adjust those things. They would be set up correctly by the vehicle producing company prior to sale.
+* The above assumption cause the QGC UI to adjust and not show various things. Providing an even simpler experience to the user.
+* The full experience continues to be available in "Advanced Mode".
+* Brands the build with various custom images and custom color palette which matches corporate branding of the theoretical commercial company this build is for.
+* Customizes portions of the interface such as you can see in the above screenshot which shows a custom instrument widget replacing the standard QGC ui.
+* It also overrides various QGC Application settings to hide some settings the users shouldn't modify as well as adjusting defaults for others.
```
akontsevich, line 24: Any example of this? Thanks!

DonLakeFlyer:
```
    /// Allows the core plugin to override the visibility for a settings group
    ///     @param name - SettingsGroup name
    /// @return true: Show settings ui, false: Hide settings ui
    virtual bool overrideSettingsGroupVisibility(QString name);

    /// Allows the core plugin to override the setting meta data before the setting fact is created.
    ///     @param settingsGroup - QSettings group which contains this item
    ///     @param metaData - MetaData for setting fact
    /// @return true: Setting should be visible in ui, false: Setting should not be shown in ui
    virtual bool adjustSettingMetaData(const QString& settingsGroup, FactMetaData& metaData);
```

There is usage in the custom example.

akontsevich: I see, thanks!

PR #8712, opened 2020-05-02
QtLocationPlugin: Add Japan Map

muramura:
I am a QGC customer in Japan.
I would like to use the maps provided by the Geographical Survey Institute of Japan.
I would add five different types of maps.

standard map
Screenshot from 2020-05-02 10-46-51

photo map
Screenshot from 2020-05-02 11-40-47

elevation map by color
Screenshot from 2020-05-02 12-25-15

anaglyph
Screenshot from 2020-05-02 11-55-59

inclination diagram
Screenshot from 2020-05-02 12-24-34

### DIFF
--- a/src/QtLocationPlugin/GenericMapProvider.cpp
+++ b/src/QtLocationPlugin/GenericMapProvider.cpp
@@ -8,6 +8,41 @@
  ****************************************************************************/
 #include "GenericMapProvider.h"
 
+static const QString JapanStdMapUrl = QStringLiteral("https://cyberjapandata.gsi.go.jp/xyz/std/%1/%2/%3.png");
+
+QString JapanStdMapProvider::_getURL(const int x, const int y, const int zoom, QNetworkAccessManager* networkManager) {
+    Q_UNUSED(networkManager)
+    return JapanStdMapUrl.arg(zoom).arg(x).arg(y);
+}
+
+static const QString JapanSeamlessMapUrl = QStringLiteral("https://cyberjapandata.gsi.go.jp/xyz/seamlessphoto/%1/%2/%3.jpg");
+
+QString JapanSeamlessMapProvider::_getURL(const int x, const int y, const int zoom, QNetworkAccessManager* networkManager) {
+    Q_UNUSED(networkManager)
+    return JapanSeamlessMapUrl.arg(zoom).arg(x).arg(y);
+}
+
+static const QString JapanAnaglyphMapUrl = QStringLiteral("https://cyberjapandata.gsi.go.jp/xyz/anaglyphmap_color/%1/%2/%3.png");
+
+QString JapanAnaglyphMapProvider::_getURL(const int x, const int y, const int zoom, QNetworkAccessManager* networkManager) {
+    Q_UNUSED(networkManager)
+    return JapanAnaglyphMapUrl.arg(zoom).arg(x).arg(y);
+}
+
+static const QString JapanSlopeMapUrl = QStringLiteral("https://cyberjapandata.gsi.go.jp/xyz/slopemap/%1/%2/%3.png");
+
+QString JapanSlopeMapProvider::_getURL(const int x, const int y, const int zoom, QNetworkAccessManager* networkManager) {
+    Q_UNUSED(networkManager)
+    return JapanSlopeMapUrl.arg(zoom).arg(x).arg(y);
+}
+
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Expect this review to take a while since Gus is very busy right now

src/QtLocationPlugin/QGCMapUrlEngine.cpp
```diff
@@ -75,6 +75,12 @@ UrlFactory::UrlFactory() : _timeout(5 * 1000) {
     _providersTable["VWorld Satellite Map"] = new VWorldSatMapProvider(this);
 
     _providersTable["Airmap Elevation"] = new AirmapElevationProvider(this);
+
+    _providersTable["Japan Std"] = new JapanStdMapProvider(this);
```
DonLakeFlyer, line 79: Is there some more specific naming for these maps than "Japan foo"? That is way too generic.

src/QtLocationPlugin/GenericMapProvider.cpp
```diff
@@ -8,6 +8,76 @@
  ****************************************************************************/
 #include "GenericMapProvider.h"
 
+static const QString JapanStdMapUrl = QStringLiteral("https://cyberjapandata.gsi.go.jp/xyz/std/%1/%2/%3.png");
+
+QString JapanStdMapProvider::_getURL(const int x, const int y, const int zoom, QNetworkAccessManager* networkManager) {
+    Q_UNUSED(networkManager)
+
+    int compulsionZoom = zoom;
+    if (compulsionZoom < 2) {
```
DonLakeFlyer, line 17: The _getURL implementations should not filter the zoom level. If data is not available at that zoom level the requests should fail. The new master code in combination Qt Map implementation automatically handles missing tiles in way such that it will zoom the best tile it has to to higher zoom level for example.

I also believe that if you change the zoom level from what is requested then it will cache incorrect tile information in the database. You will end up with repeated tiles in the map.

src/QtLocationPlugin/QGCMapUrlEngine.cpp
muramura: I change std to contour.

src/QtLocationPlugin/GenericMapProvider.cpp
muramura: I would remove the ZOOM filter process.

src/QtLocationPlugin/QGCMapUrlEngine.cpp
DonLakeFlyer: I"m not talking about "Std" versus contour. I'm talking about only identifying this class of maps with the prefix "Japan". That provides no detail as to what service is providing these maps. How is a user supposed to understand where these maps come from.

DonLakeFlyer: Which leads me to also think there should be some other piece of more detailed information for each map provider that is user visible. So that these more specific custom area maps can tell the user some more information than just a name. But for now all we have is a name, so that is the only place we have to provide at least some level of detail.

muramura: @DonLakeFlyer san. I get the maps from the Geospatial Information Authority of Japan
I think the official name is as follows.
I think this is a long string of characters.
What should I do?

Name:.
Geospatial Information Authority of Japan

DonLakeFlyer: It looks like from here: https://www.gsi.go.jp/ENGLISH/ that they use the initials GSI to refer to themselves. So how about "Japan (GSI)" as a prefix.

muramura: @DonLakeFlyer san. Thank you. I agree. I'll change.

muramura: I want to register more than one GSI map.
When I changed it to "Japan(GSI)", the map type was not displayed.
I think "Japan-GSI" is fine.

Screenshot from 2020-09-03 06-57-53

DonLakeFlyer: Yes, that's fine

DonLakeFlyer:
Thanks so much for the pull

muramura:
@DonLakeFlyer san. Thank you, Marge. It's nice to have the GSI map of Japan available in Japan.

PR #8721, opened 2020-05-04
Don't allow user to enter a bad formed rtsp url

stefandunca:
Add a regex validator to the RTSP URL textfield to match the allowed URL.
Gstreamer doesn't handle well the bad formated rstp urls. A space in front will break finding the stream
as it happened with some users.

### DIFF
--- a/src/FactSystem/FactControls/FactTextField.qml
+++ b/src/FactSystem/FactControls/FactTextField.qml
@@ -45,6 +45,7 @@ QGCTextField {
             validate:       true
             validateValue:  _validateString
             fact:           _textField.fact
+            validator:      _textField.validator
         }
     }
 
@@ -52,6 +53,7 @@ QGCTextField {
         id: helpDialogComponent
         ParameterEditorDialog {
             fact: _textField.fact
+            validator:      _textField.validator
         }
     }
 }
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
We tried using these validation things in the Fact system ages ago. They don't work at all from a usability standpoint. In some cases they simply don't let you type things which is a mystery to the user as to why. Looks like in the case of regexp you can type whatever you want but then if it doesn't match pattern it just gets tossed. Type something wrong in the field, move away to another control it stays showing the bad value leaving the user thinking everything is ok. Move to a different page, come back and suddenly the changes you made are gone like magic. No notice to the user they did something wrong.

The real way to implement this is to add support for script based value validation to the FactMetaData system. That way it works just like everything else and provide consistent feedback on bad values as well.

It could be one offed with custom handling of onEditingFinished in GeneralSetting. But it would be way nicer to do it for real since that would be used in other places as well. There are other values which could use the same thing.

Anyway, as this pull stands it doesn't meet the usability requirements for QGC.

DonLakeFlyer:
Give me a couple days and I'll add support for script based validation.

stefandunca:
I tested it and it works fine. It won't let you type anything that doesn't match an URL. If work to do it even better I will close this PR. Just, please, post the PR in here or mention this one.

DonLakeFlyer:
> It won't let you type anything that doesn't match an URL.

Which is why these validator things are a usability nightmare.

Issue #8578, opened 2020-03-20
gimbal yaw indicator erroneously propagates to followup waypoints


remster:
## Expected Behavior
When the gimbal checkbox isn't checked on a waypoint - the [visual](https://github.com/mavlink/qgroundcontrol/blob/master/src/QmlControls/MissionItemIndexLabel.qml#L60) shouldn't show.

## Current Behavior
it shows
![yaw-bug](https://user-images.githubusercontent.com/8327961/77190386-70bb6c00-6ad9-11ea-9453-9e884323577d.png)

## Steps to Reproduce:
1. create a plan with a number of waypoints
2. check camera->gimbal on the Nth
3. all waypoints that follow Nth show the same gimbal heading/visual

## System Information
When posting bug reports, include the following information
- Operating System: Ubuntu 18.04
- QGC Version: a22c17b455a5cca6018d0f5abed4b3c2537b7a18
- QGC build: off master 
- Flight Controller: N/A
- Autopilot (with version): N/A

## Detailed Description
I believe this may be the fix:
```
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -1329,15 +1329,8 @@ void MissionController::_recalcMissionFlightStatus()
             _missionFlightStatus.vehicleSpeed = newSpeed;
         }
 
-        // Look for gimbal change
-        double gimbalYaw = item->specifiedGimbalYaw();
-        if (!qIsNaN(gimbalYaw)) {
-            _missionFlightStatus.gimbalYaw = gimbalYaw;
-        }
-        double gimbalPitch = item->specifiedGimbalPitch();
-        if (!qIsNaN(gimbalPitch)) {
-            _missionFlightStatus.gimbalPitch = gimbalPitch;
-        }
+        _missionFlightStatus.gimbalYaw = item->specifiedGimbalYaw();
+        _missionFlightStatus.gimbalPitch = item->specifiedGimbalPitch();

```

DonLakeFlyer:
If you adjust the gimbal, that adjustment continues on until you make another adjustment. Leaving the gimbal settings unchecked does not bring the gimbal back to neutral they stay where they were before. Hence the visual continue to show.

remster:
Thanks. Is there a purpose (feature/use-case) to this behaviour?
[...]

PR #8729, opened 2020-05-05
Introduce gimbal yaw visiblity setting

remster:
Addresses: https://github.com/mavlink/qgroundcontrol/issues/8578
And this is what it achieves:

gimbal_yaw

I.e.: the specific gimbal visual can now be visible only when the user actually assigns value to its gimbal yaw.


DonLakeFlyer:
I think a bool with the name `showGimbaYawlOnlyWhenSet` would be more understable.

remster:
I thought maybe one day you would find a reason to `showGimbalWhen=Never`.
Your comment addressed: https://github.com/mavlink/qgroundcontrol/pull/8729/commits/472b2acf78f8d992416c5ee8ca8ac8853f36743e

Issue #7662, opened 2019-08-05
VTOL Mission Time estimation is not cruise speed based


abdulmajid92:
Dear,

I found an issue in Plan view. When I want to make a mission, the mission flight time estimation based on vehicle  hover speed, which should be based on vehicle cruise speed.

![Capture](https://user-images.githubusercontent.com/14164260/62434691-8e62dd80-b763-11e9-850d-4a4a38db71ae.JPG)

moreba1:
YES , it is wrong.

moreba1:
@DonLakeFlyer 
can it added to realease v4.5 milestone ?

PR #8735, opened 2020-05-07
Default state of VTOL in mission

moreba1:
The default state of most VTOLs in mission (for example survey ) is cruise. so cruise speed must be used for flight time. 

https://github.com/mavlink/qgroundcontrol/issues/7662

### DIFF
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -1544,6 +1544,8 @@ void MissionController::_recalcMissionFlightStatus()
         }
 
         // Update VTOL state
+        //VTOL default state is criuse 
+        vtolInHover = false;
         if (simpleItem && _controllerVehicle->vtol()) {
             switch (simpleItem->command()) {
             case MAV_CMD_NAV_TAKEOFF:


DonLakeFlyer:
This is not a correct fix. QGC tracks the vtol state hover or cruise based on the mission items themselves. The initial state needs to be based on whether the mission starts from the ground or not (which is a setting). If it starts from the ground it starts in hover. If it start in the air, then the best you can do is assume hover. I'll take a look at fixing it.

moreba1:
Most  VTOLs start mission in VTOL takeoff. missions after the transition to fixed-wing and before the transition to multirotor must be known as cruise state.

DonLakeFlyer:
> If it start in the air, then the best you can do is assume hover.

Typo: "assume cruise".

DonLakeFlyer:
The new daily build should be available in an hour. Can you test to verify my fix works in the following cases:
* Mission starts on ground with a MAV_CMD_NAV_VTOL_TAKEOFF. Hover is initial state.
* Mission starts on ground with a MAV_CMD_NAV_TAKEOFF. Cruised is initial state. It will be a bit tricky to create this mission. You'll need to add a waypoint and change it.
* Mission starts in the air which means it assumes cruise as the starting point.
* VTOL transition commands within the mission correctly adjust

moreba1:
I tested daily build. time of mission is wrong and it is hover speed based .you can test it, changing cruise speed don't change mission time.
image

DonLakeFlyer:
Can  you attach the plan file and tell me what you are expecting for time?

moreba1:
I expected for this plan 17:40 min time with 20 m/s cruise speed.
The total distance of the mission is 20867 m.
[vtolplan.zip](https://github.com/mavlink/qgroundcontrol/files/4602846/vtolplan.zip)

DonLakeFlyer:
Are you sure you're running latest daily. I get this:
Screen Shot 2020-05-14 at 7 13 27 PM

moreba1:
> 
> 
> Are you sure you're running latest daily. I get this:
> Screen Shot 2020-05-14 at 7 13 27 PM

how much is your hover speed?

DonLakeFlyer:
That comes from the Plan file:
```
    "mission": {
        "cruiseSpeed": 20,
        "firmwareType": 12,
        "hoverSpeed": 20,
```

moreba1:
please change hoverspeed to 5 m/s .

PR #8742, opened 2020-05-11
Added configurable value grid to toolbar

DonLakeFlyer:
Screen Shot 2020-05-11 at 1 59 25 PM

Screen Shot 2020-05-11 at 2 03 07 PM


Antiheavy:
Neat!

DonLakeFlyer:
@antiheavy I'm also working my way to vehicle type specific initial settings for this. Would be helpful to hear from you on what you think should be up top and what should be on the side. The goal is critical information up top which is alway visible (even from another view). And secondary information to the side. Possibly to the point of not only having a values page, but say when flying a mission it switches to a mission page which provides you with mission details, time left, distance left maybe stuff like that.

Antiheavy:
I think altitude is applicable for all aircraft. 

After that it opinions can vary a lot I think.  
Airplanes often care a lot about airspeed and groundspeed, but maybe those could live on the right side by default.

Issue #8484, opened 2020-03-05
4.0 android/mobile qml issues


jaxxzer:
In each of these screenshots, some item or text is too wide and is being cropped

![image](https://user-images.githubusercontent.com/8315108/75944215-f7224d80-5e4b-11ea-80ac-3022a5f156d2.png)

![image](https://user-images.githubusercontent.com/8315108/75944230-ff7a8880-5e4b-11ea-9b00-1551d1488812.png)

![image](https://user-images.githubusercontent.com/8315108/75944239-030e0f80-5e4c-11ea-87c9-20001c2a3036.png)

![image](https://user-images.githubusercontent.com/8315108/75944252-073a2d00-5e4c-11ea-9091-08a3c2177441.png)

tcanabrava:
some of those are also happening on desktop since 3.x

DonLakeFlyer:
What size device. Target mobile is 10" tablet. Although things will sort of work below that we generally don't try to fix layout issues on smaller devices.

jaxxzer:
> What size device. Target mobile is 10" tablet. Although things will sort of work below that we generally don't try to fix layout issues on smaller devices.

Oh, it's an 8" device. I did not know about the 10" thing.

DonLakeFlyer:
There is some small amount of write up on it here: https://dev.qgroundcontrol.com/en/ui_design/multi_device_pattern.html

jaxxzer:
noting here for @tcanabrava to check the font size options in the general settings menu, that can throw some of the layouts off too.

jaxxzer:
@Williangalvani merged pull request has fixed the problem on linux. The problem persists on windows.

PR #8747, opened 2020-05-13
[WIP] QGCApplication: set style for all OS

jaxxzer:
Following up with https://github.com/mavlink/qgroundcontrol/pull/8736 in regard to https://github.com/mavlink/qgroundcontrol/issues/8484

This ui fix was previously applied to linux only, but windows was still demonstrating the same problem

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -223,14 +223,14 @@ QGCApplication::QGCApplication(int &argc, char* argv[], bool unitTesting)
             }
             permFile.close();
         }
-
-        // Set default QtQuick style if not configured
-        if (QString(getenv("QT_QUICK_CONTROLS_STYLE")).isEmpty()) {
-            QQuickStyle::setStyle("Universal");
-        }
     }
 #endif
 #endif
+
+    // Set default QtQuick style if not configured
+    if (QString(getenv("QT_QUICK_CONTROLS_STYLE")).isEmpty()) {
+        QQuickStyle::setStyle("Universal");
+    }
 
     // Setup for network proxy support
     QNetworkProxyFactory::setUseSystemConfiguration(true);


DonLakeFlyer:
I"m confused. This was stated to me as a Linux only problem?

jaxxzer:
There was a bit of a mixup. The original ticket applied to android specifically.
We should have opened a new issue for the https://github.com/mavlink/qgroundcontrol/pull/8736 fix, as it focused on a similar ui issue that arose on desktop after the 4.0 stable release.

This is what users are currently seeing in windows on 4.0.6:
image

DonLakeFlyer:
> This is what users are currently seeing in windows on 4.0.6:

What I don't understand is why this changes fix it? Things should be positioned and sized based on default font point sizes. I don't see how that is affected by this change?

DonLakeFlyer:
So I see two things here:
* The sizing problems which I will take a look at but just look like implementation bugs with respect to ratio based sizing/positioning. Changing a control style may move things around and make these bugs appear to be gone but in reality they are not. Even after whatever is screwed up with sizing/positioning is fixed things can still end up not fitting. Anything that has the possibility of happening should be in a flickable so you can still get to everything. Users can bump the font size up which can easily cause things to get bigger than the available space. And the user should still be able to interact with all the parts of the ui.
* The control style thing. According to docs there is a "Default Style" which is what you get if you don't specify a specific style. QGC doesn't want to have it's ui changed based on some environment variable. Hence I would say that the correct change is to always set to default style to bast over any random environment variable setting. And that Universal Style is incorrect since that is some fancy Microsoft guidelines style.

DonLakeFlyer:
The problem with the Video Page is that it uses a TextField instead of a QGCTextField:
```
        TextField {
            id: videoFileName
            visible: QGroundControl.videoManager.isGStreamer
            width: 100
        }
```

 This TextField then comes from the control "style" and hence is affected by whatever style is selected. When you switch to a "Universal Style" that style just happens to have a smaller control for TextFields which all of a sudden makes things fit. Also page isn't in a flickable which means it can grown larger than it's bound without the suer being able to get to what they need to go to. 

I'll fix all of this up in Stable and bring across.

DonLakeFlyer:
Also the file name label is randomly in a different size font than everything else!

jaxxzer:
@DonLakeFlyer thanks for all of that. This was a bit of a shot in the dark on my part.

jaxxzer:
For what it's worth, the build artifact for windows is working better after this fix.

DonLakeFlyer:
It smelled funny right from the start. I should have went with my gut initially.

DonLakeFlyer:
> the build artifact for windows is working better after this fix.

Just dumb luck on the TextField implementation in Universal Style being smaller and letting things fit.

DonLakeFlyer:
This is all fixed up now

PR #8749, opened 2020-05-13
Video widget

DonLakeFlyer:
* Fix video page widget control usage and sizing
* Always apply "Default" style to Qml Controls. In general this shouldn't be needed since we have our own custom control set. But there are a few like sliders which use the vanilla control. By forcing to Default style it prevents the QT_QUICK_CONTROLS_STYLE from randomly changing the QGC ui.

Screen Shot 2020-05-13 at 12 52 09 PM

Replacement for #8747 and friends.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,7 +4,9 @@ Note: This file only contains high level features or important fixes.
 
 ## 4.0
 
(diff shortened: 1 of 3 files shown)


jaxxzer:
It works thanks!

Issue #8746, opened 2020-05-13
Joystick button flight modes don't load on startup


almatthew:
## Expected Behavior
Joystick flight mode button actions should persist.

## Current Behavior
All flight behavior _except_ flight mode selections persist.

## Steps to Reproduce:
1. Start QGroundControl, connect to an APM autopilot
2. Connect a USB gamepad or joystick
3. Make at least one joystick button action a flight mode change (e.g. Loiter or Guided)
4. Close QGroundControl
5. Open QGroundControl, connect to an APM autopilot
6. Connect a USB gamepad or joystick (if not already connected)
7. In the Joystick settings section, find that the flight mode change actions have been removed

## System Information
- Operating System: OS X & Windows
- QGC Version: Stable_V4.0 (and associated tags)
- QGC build: stable, self-built
- Flight Controller: APM-supported APs
- Autopilot (with version): APM 4.0.3 (though I would imagine PX4 would fail as well)

## Detailed Description
This appears to be due to the way the flight mode button actions are populated at start up in the Joystick class.  Specifically, the flight mode button actions are not populated on startup, even though the flight mode button actions have been saved properly in the ini file.  This appears to be due to the lack of an `activeVehicle`: https://github.com/mavlink/qgroundcontrol/blob/Stable_V4.0/src/Joystick/Joystick.cc#L1070.

## Log Files and Screenshots
Setting the button actions:
<img width="1283" alt="Setting button actions" src="https://user-images.githubusercontent.com/5853291/81828455-a6d00780-94ee-11ea-8e41-ecc8e9d216ff.png">

After restarting QGC:
<img width="1284" alt="Subsequent start of QGC" src="https://user-images.githubusercontent.com/5853291/81828475-aa638e80-94ee-11ea-8469-9b9b6d39ac61.png">

DonLakeFlyer:
Thanks @almatthew.

> Specifically, the flight mode button actions are not populated on startup, even though the flight mode button actions have been saved properly in the ini file.

Ah. You checked the .ini file yourself already and the values are correct in there? That was what I was going to ask you to do first.

DonLakeFlyer:
So it's just flight mode change buttons which don't stick then right?

almatthew:
Correct, the `.ini` includes the flight modes I selected.  And yes, it appears that its only the flight modes that don't persist, all other actions seem to work fine.

almatthew:
[...]

PR #8750, opened 2020-05-13
Fix disappearing flight mode button assignments

DonLakeFlyer:
The load settings for joystick buttons was validating the action against the currently known set of available actions. Tossing the actions which were not in the list. At the time load settings is called there isn't a vehicle yet hence the flight modes aren't in the list. Hence all flight mode assignments were being dropped.

I think it's better to just leave these in there in any case. It is possible to set up joystick for Vehicle A which has a different set of flight modes available than Vehicle B. It's better to have the flight mode change just fail with an error in that case.

The new way it works from a ui perspective is that if a button is truly assigned to a flight mode which currently does not exist then it shows "No Action" which isn't quite correct. But good enough I think then trying to get fancy.

Fix for #8746

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -279,7 +279,7 @@ void Joystick::_loadSettings()
 
     for (int button = 0; button < _totalButtonCount; button++) {
         QString a = settings.value(QString(_buttonActionNameKey).arg(button), QString()).toString();
-        if(!a.isEmpty() && _findAssignableButtonAction(a) >= 0 && a != _buttonActionNone) {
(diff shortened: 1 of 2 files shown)


almatthew:
@DonLakeFlyer this fix works well on my bench, thanks!

DonLakeFlyer:
Thanks for testing @almatthew. Once Gus has time to double-check I'll merge and bring to master as well. May be a slight delay since Gus is crazy busy right now.

DonLakeFlyer:
Thanks @dogmaphobic. I think this mechanism is a bit safer since it also handles vehicle type changes where the flight modes differ between vehicle types.

PR #8752, opened 2020-05-13
FactMetaData: Add custom validator support

DonLakeFlyer:
FactMetaData:
```
    // Custom function to validate a cooked value.
    //  @return Error string for failed validation explanation to user. Empty string indicates no error.
    typedef QString (*CustomCookedValidator)(const QVariant& cookedValue);
```

```
    /// Sets a custom cooked validator function for this metadata. The custom validator will be called
    /// prior to the standard validator when convertAndValidateCooked is called.
    void setCustomCookedValidator(CustomCookedValidator customValidator) { _customCookedValidator = customValidator; }
```

This then ties into the normal Fact validation system. Such that the users get an explanation and an error popup dialog when validation fails.

Related to #8721

### DIFF
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -424,6 +424,13 @@ bool FactMetaData::convertAndValidateCooked(const QVariant& cookedValue, bool co
 
     errorString.clear();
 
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@stefandunca FYI

PR #8755, opened 2020-05-15
First run prompts

DonLakeFlyer:
Modified the support for first run prompts and added a new prompt for offline vehicle settings. Custom builds which were using the previous mechanism will need to adjust their code. But it should be trivial changes. I'll be creating docs for how it all works later today. 

Screen Shot 2020-05-15 at 9 44 00 AM


DonLakeFlyer:
@stefandunca FYI

DonLakeFlyer:
@dakejahl This new support can also be used for implementing startup tips in a standard way. I'll explain in the docs.

DonLakeFlyer:
https://github.com/mavlink/qgc-dev-guide/blob/master/en/custom_build/FirstRunPrompts.md

PR #8756, opened 2020-05-16
fix too many fail message box when failed to start mission

lizn520:
it's too many error message boxs when failed to start mission，like the below picture:
image

it's caused by the 3 time arming tries so I add a function to set vehicle armed  with no arm fail message, after fixed like:

image

### DIFF
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -718,7 +718,7 @@ bool FirmwarePlugin::_armVehicleAndValidate(Vehicle* vehicle)
 
     // We try arming 3 times
     for (int retries=0; retries<3; retries++) {
-        vehicle->setArmed(true);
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Hmm, yeah that's nasty. 

But I would instead do the following:
* Change Vehicle::setArmed to `void setArmed   (bool armed, bool showError = true);` instead of making a new method. You'll have to tweak some code which uses setArmed in a slot in other places.
* Use the showError=false variant in FirmwarePlugin::_armVehicleAndVehicle
* Change FirmwarePlugin::_armVehicleAndVehicle to only call setArm once. Given the fact that arming is a command long it already has all the retry logic it needs.

DonLakeFlyer:
This pull https://github.com/mavlink/qgroundcontrol/pull/9198 basically does what your pull did and a but more to fix the problem. Thanks for pushing on this. It should be better now.

PR #8757, opened 2020-05-18
Fix incorrect units of gradient on planner screen.

samuelctabor:
The gradient was displayed with "%" as the unit, whereas it is actually degrees.


samuelctabor:
Thanks Don, changed to "deg"

DonLakeFlyer:
This was superceded by another pull

PR #8758, opened 2020-05-18
Fix more and more deprecated issues

patrickelectric:


patrickelectric:
@DonLakeFlyer this fix a bunch of warning problems related to the deprecated functions.
We are also unable to build QGC anymore, since warnings appears to be errors now, so anyone that is using a modern Qt version is unable to build QGC right now.

jaxxzer:
`asprintf` was introduced in Qt 5.5 https://doc.qt.io/qt-5/qstring.html#asprintf. There is a warning in the doc:

> Warning: We do not recommend using QString::asprintf() in new Qt code. Instead, consider using QTextStream or arg(), both of which support Unicode strings seamlessly and are type-safe.

patrickelectric:
Both QTextStream and arg does not support cformat style.

jaxxzer:
refactor is still possible, just ditch the format string and use precision parameter of `arg` to format the floats. NBD to me though, the small patch is nice.

DonLakeFlyer:
> so anyone that is using a modern Qt version is unable to build QGC right now.

If you are doing that you are living in the territory of "unsupported". That said there should be an easier way to turn off warnings as errors. I'm in the middle of reworking all that. I'll fix it up. If you don't see it soon, ping me. I may forget. I've been out of town so lots of catchup to do.

PR #8684, opened 2020-04-22
Pr small video tweaks

tcanabrava:
Some changes on top of @andrewvoznytsa `pr-small-video-tweaks` as he's no longer active.

PR #8759, opened 2020-05-18
Pr small video tweaks

jaxxzer:
This replaces https://github.com/mavlink/qgroundcontrol/pull/8684, it's been rebased on master. I will pick up where @tcanabrava left off.


src/VideoManager/VideoManager.cc
```diff
@@ -282,11 +282,6 @@ VideoManager::startRecording(const QString& videoFile)
         return;
     }
 #if defined(QGC_GST_STREAMING)
-    if (!_videoReceiver[0]) {
```
jaxxzer, line 285: @MatejFranceskin can you explain the reasoning for applying this patch?

PR #8762, opened 2020-05-19
Add debian documentation, fix for hardcoded dot path

AJRepo:
Updated the documentation to indicate correct packages for "apt" to import to build documentation. 

Fixed hardcoded path for dot command which caused doxygen command to fail on stock debian OS.


AJRepo:
Something odd in the pull request. I'm deleting this and will re-do as a branch instead.

PR #8764, opened 2020-05-20
Debian doc updates

AJRepo:
Updated the documentation to indicate correct packages for "apt" to import to build documentation.

Fixed dot errors (hardcoded path for dot)  in script building doxygen docs on Ubuntu 

Fixes TravisCI errors since Trusty has been replaced with xenial.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,13 +29,13 @@ jobs:
           sudo: required
         - stage: "Build"
           name: "Android 32 bit"
-          dist: trusty
+          dist: xenial
           language: android
           env: SPEC=android-clang CONFIG=installer BITNESS=32 GSTREAMER_NAME=armv7
           sudo: false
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
The other Linux builds have moved to bionic. So I think android should likely move there as well. Not sure if there is a problem with that or not.

AJRepo:
I first tried Bionic but the TravisCI build failed. I didn't include that in this pull request. 

I tried with Xenial and the TravisCI build succeeded. So this is a minimum change to get builds to continue to work on the TravisCI platform. 

The Trusty builds will continue to fail because the apt repositories that the build scripts use are no longer provided by Ubuntu.

DonLakeFlyer:
Ok, great. Thanks for the explanation and the fixes.

Issue #8765, opened 2020-05-20
Compiler warnings turned to errors


akontsevich:
Errors converted back to warnings: https://github.com/mavlink/qgroundcontrol/commit/c24029938e88d7a45a04f4e4e64bf588f595afed, however still can't build QGC:

> `...qgroundcontrol/src/Vehicle/Vehicle.cc:1189:48: error: error: taking address of packed member of ‘__mavlink_attitude_target_t’ may result in an unaligned pointer value [-Werror=address-of-packed-member]`

> `...qgroundcontrol/src/Vehicle/Vehicle.cc:1926:9: error: error: taking address of packed member of ‘__mavlink_rc_channels_t’ may result in an unaligned pointer value [-Werror=address-of-packed-member]`

Any ideas?

PR #8769, opened 2020-05-22
-Waddress-of-packed-member build error fix

akontsevich:
Resolves https://github.com/mavlink/qgroundcontrol/issues/8765:
- [x] -Waddress-of-packed-member build error fix
- [x] Also fixed `/custom` build

### DIFF
--- a/custom-example/custom.pri
+++ b/custom-example/custom.pri
@@ -35,14 +35,14 @@ CONFIG  += QGC_DISABLE_PX4_PLUGIN_FACTORY
 DEFINES += CUSTOMHEADER=\"\\\"CustomPlugin.h\\\"\"
 DEFINES += CUSTOMCLASS=CustomPlugin
 
-TARGET   = MyGroundStation
+TARGET   = CustomQGroundControl
 DEFINES += QGC_APPLICATION_NAME='"\\\"Custom QGroundControl\\\""'
 
 DEFINES += QGC_ORG_NAME=\"\\\"qgroundcontrol.org\\\"\"
 DEFINES += QGC_ORG_DOMAIN=\"\\\"org.qgroundcontrol\\\"\"
 
 QGC_APP_NAME        = "Custom QGroundControl"
-QGC_BINARY_NAME     = "MyGroundStation"
+QGC_BINARY_NAME     = "CustomQGroundControl"
 QGC_ORG_NAME        = "Custom"
 QGC_ORG_DOMAIN      = "org.custom"
 QGC_APP_DESCRIPTION = "Custom QGroundControl"
--- a/custom-example/custom.pri
+++ b/custom-example/custom.pri
@@ -42,7 +42,7 @@ DEFINES += QGC_ORG_NAME=\"\\\"qgroundcontrol.org\\\"\"
 DEFINES += QGC_ORG_DOMAIN=\"\\\"org.qgroundcontrol\\\"\"
 
 QGC_APP_NAME        = "Custom QGroundControl"
-QGC_BINARY_NAME     = "CustomQGroundControl"
+QGC_BINARY_NAME     = "MyGroundStation"
 QGC_ORG_NAME        = "Custom"
 QGC_ORG_DOMAIN      = "org.custom"
 QGC_APP_DESCRIPTION = "Custom QGroundControl"
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -50,7 +50,6 @@ linux {
         DEFINES += QGC_GST_TAISYNC_ENABLED
         DEFINES += QGC_GST_MICROHARD_ENABLED 
         QMAKE_CXXFLAGS_WARN_ON += -Werror \
-            -Wno-address-of-packed-member \     # Mavlink headers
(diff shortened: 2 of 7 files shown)


custom-example/custom.pri
```diff
@@ -42,7 +42,7 @@ DEFINES += QGC_ORG_NAME=\"\\\"qgroundcontrol.org\\\"\"
 DEFINES += QGC_ORG_DOMAIN=\"\\\"org.qgroundcontrol\\\"\"
 
 QGC_APP_NAME        = "Custom QGroundControl"
-QGC_BINARY_NAME     = "CustomQGroundControl"
```
DonLakeFlyer, line 45: Why the name change here?

DonLakeFlyer:
Is there no way to update QGCCommon.pri to turn off the warnings? This is where all the other stuff is handled. Doing in directly in source code doesn't work well because it can mover around and show up in new files. It also pollutes the source with crap.

akontsevich:
> Is there no way to update QGCCommon.pri to turn off the warnings? This is where all the other warnings stuff is handled. Doing in directly in source code doesn't work well because it can mover around and show up in new files. It also pollutes the source with crap.

@DonLakeFlyer, of course it is possible, however 2 things there:
- in recent commits you have turned warnings in to errors, so I assumed that was done intentionally for some reason and did not want to alter `QGCCommon.pri` because of that. And it is turned on explicitly for Android here: https://github.com/mavlink/qgroundcontrol/blob/master/QGCCommon.pri#L53
- To solve the issue I just took QGC code as an example from here: https://github.com/mavlink/qgroundcontrol/blob/master/src/comm/QGCMAVLink.h#L26-L32 assuming this is the right way for you to solve this.

So because of the above I went by the way of changing sources. Any other thoughts, ideas?

custom-example/custom.pri
akontsevich: To align with the `TARGET` here: https://github.com/mavlink/qgroundcontrol/blob/master/custom-example/custom.pri#L38. Otherwise I got build errors.

DonLakeFlyer: Ah, ok. Then TARGET should change to match QGC_BINARY_NAME not the other way around.

DonLakeFlyer:
Hmm. In looking at this more: https://github.com/mavlink/qgroundcontrol/blob/master/src/comm/QGCMAVLink.h#L26-L32 I like it better since it localizes to that specific case as opposed to the entire build. Let me look at the android build for a minute to see what makes more sense.

DonLakeFlyer:
Hmm. Now that I look more at this: https://github.com/mavlink/qgroundcontrol/blob/master/src/comm/QGCMAVLink.h#L26-L32 I like it better since it is localized to a specific header instead of turning it off across the entire build which is what QGCCommon.pri does.

Instead of disabling at a complete header scope could you instead turn off the global disable for android and put the warning disable directly in the C++ code location(s) instead which causes the problem. This limits the scope to only where it is exactly needed. Thus leaving the error to pop in places which it make be a good indicator of a true problem.

custom-example/custom.pri
akontsevich: Ok: https://github.com/mavlink/qgroundcontrol/pull/8769/commits/f27dbb68faff09360ee90ee9228dc62af5d0b13e

akontsevich:
> Instead of disabling at a complete header scope could you instead turn off the global disable for android and put the warning disable directly in the C++ code location(s) instead which causes the problem. This limits the scope to only where it is exactly needed. Thus leaving the error to pop in places which it make be a good indicator of a true problem.

Ok removed warning disabling on:
- mac build: https://github.com/mavlink/qgroundcontrol/blob/master/QGCCommon.pri#L121
- on android: https://github.com/mavlink/qgroundcontrol/blob/master/QGCCommon.pri#L53

Commit: https://github.com/mavlink/qgroundcontrol/pull/8769/commits/efcd990cd41b49a8dda4bb8fa1dd1b648210c4d9 - builds fine on Linux and for android, then should be fine for mac as well as clang used on both.

DonLakeFlyer:
Right but I'd like to move the warning disabling out of the Vehicle/UAS.h headers since that is a pretty high level global spot as well which could hide other true problems. And instead wrap the actual C++ code itself which is causing the error with push and pop of the warning off/on. That way it limits the scope to the smallest possible place. Leaving other possible real problems which may show up in the future causing build failures.

akontsevich:
> Right but I'd like to move the warning disabling out of the Vehicle/UAS.h headers since that is a pretty high level global spot as well which could hide other true problems. And instead wrap the actual C++ code itself which is causing the error with push and pop of the warning off/on. That way it limits the scope to the smallest possible place. Leaving other possible real problems which may show up in the future causing build failures.

Ah, I see, ok: https://github.com/mavlink/qgroundcontrol/pull/8769/commits/ce674f46cd38b070aae2e2b89036c6e40fcd49ec

DonLakeFlyer:
Great thanks for the help.

akontsevich:
> Great thanks for the help.

Thanks You!

Issue #6866, opened 2018-09-11
Saving DNS names for TCP and UDP comm links


exp131:
Hello,
I have multiple vehicles connected to QGC via TCP comm link, and there is a little issue with it. When I create TCP link and enter a DNS name QGC finds and save an IP address for this link. But if IP address is changes QGC doesn't track this change, so I need to edit all connections with DNS name to update an IP address. I suggest to save DNS name if entered instead of IP address and discover IP address when "Connect" button pressed.
Regards,
Vitaly.

dwiel:
I can't even enter a dns name once. I enter the name and click OK, but the hostname reverts to the previous ip address. It is possible this is because the dns name isn't found, but there is no indication that that is what is happening.

PR #8784, opened 2020-05-26
Feature request: Save hostnames instead of host addresses for TCP/UDP links

sfalexrog:
**Background**: This mainly concerns drones that are used for indoor flights. We are using QGroundControl to connect to vehicles that are in a local wireless network. The vehicles themselves are equipped with an onboard computer running ROS, with `mavros` serving as, among other things, a MAVLink proxy.

**The problem**: The drones may connect to different networks and may be given different IP addresses. Name resolution inside the network is handled by `avahi`, dropping the requirement to configure the network's DHCP/DNS servers. Thus, a drone with a single hostname may have different IP addresses in different networks, during different runs, etc.

This is not a big problem for ssh, ROS, and other tools, since the drone may be accessed by its hostname, but QGroundControl seems to only store IP addresses and resolve them at link configuration time.

**Proposed solution**: I suggest storing unresolved hostnames for TCP and UDP connections and resolve them at a later time (ideally when the connection is established). This seems to be easier for TCP connections (where address is only used when connection is established), UDP probably requires more thought.

Right now having an unresolvable hostname in a UDP link results in several seconds of blank window upon starting QGroundControl. Changing networks while QGroundControl is running is not handled as well.

Addresses #6866

### DIFF
--- a/src/comm/TCPLink.cc
+++ b/src/comm/TCPLink.cc
@@ -151,6 +151,11 @@ bool TCPLink::_hardwareConnect()
     _socket = new QTcpSocket();
 
     QSignalSpy errorSpy(_socket, static_cast<void (QTcpSocket::*)(QAbstractSocket::SocketError)>(&QTcpSocket::error));
+    auto address = _tcpConfig->address();
+    if (address.isNull())
+    {
+        emit communicationError(tr("Link Error"), tr("Error on link %1. Unable to resolve hostname %2").arg(getName(), _tcpConfig->host()));
+    }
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
A UDPLink is already threaded. You should be able to move the address resolution from the constructor to inside the threaded part of UDPLink startup such that it doesn't hang the app. Isn't that possible?

DonLakeFlyer:
@sfalexrog Updates?

DonLakeFlyer:
Closing as abandoned

Issue #8774, opened 2020-05-25
QGC Vistual Stick take control Ardupilot after takeoff 10s


bys1123:
QGC Vistual Stick take control Ardupilot after takeoff 10s makes drone full speed flight backwards and hit the pilot.

This maybe mostly the pilot's mistake but I think if ArduPilot wouldn't disarm or QGC wouldn't take control especially after 10 seconds that will give more safety and prevent this unsafety happen.

Firmware: Ardupilot
OS: MacOS
Log: [2020-05-24 16-08-59.zip](https://github.com/mavlink/qgroundcontrol/files/4678010/2020-05-24.16-08-59.zip)

Vistual Stick looks like this on MacOS, but the pilot forget it:
![image](https://user-images.githubusercontent.com/4748411/82827931-99e5e780-9ee2-11ea-90b1-5e80935d2788.png)

bys1123:
tlog playback: https://youtu.be/bHAEzjJyW0k

DonLakeFlyer:
Hmm, haven't look at virtual joystick in ages. I believe the right stick it supposed to auto-center though. So when you let you it should go back to neutral. I'll take a look at what is going on there. I think the auto-center is broken.

DonLakeFlyer:
No, the right stick doesn't auto-center. Only the left stick yaw auto-centers. Trying to remember why. I think it relates to fixed wing versus multi-rotor usage. I think it makes sense to have a setting to auto-center the right stick in both axes. That setting should default to on. Does that make sense?

bys1123:
Thanks for reply. The most important part that I'm wondering is why it take control the drone after takeoff few seconds. This could make lots of troubles, 
And the doc said it only can control PX4.

DonLakeFlyer:
> The most important part that I'm wondering is why it take control the drone from RC after takeoff few seconds. This could make lots of troubles,
> And the doc said it only can control PX4.

ArduPilot added support for the mavlink MANUAL_CONTROL message a while back which means that joystick both real and virtual are now supported. The docs are out of date. As to why there is a delay, I don't know. Could be some ArduPilot quirk/bug. I can do some SITL testing to verify.

DonLakeFlyer:
> I believe the right stick it supposed to auto-center though.

The auto-center support was broken in the 4.0 code. It's supposed to do this all the time.

wolfling-052:
[...]

PR #8785, opened 2020-05-26
Rework reverse/center throttle support. Fix right stick centering bug.

DonLakeFlyer:
* Fix for #8774. Related to #7375
* Fixes right stick auto-center
* Adds reverse support for rover/sub which sends correct negative thrust values for reverse
* Fixes throttle auto-center with respect to reverse as well

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,6 +7,7 @@ Note: This file only contains high level features or important fixes.
 ## 4.0.7 - Not yet released
 
 * Fix video page sizing
+* Virtual Joystick: Fix right stick centering. Fix/add support for rover/sub reverse throttle support.
 
 ### 4.0.6 - Stable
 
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
Need to hear back on questions in #7375 since I don't understand how that would have ever worked. Need to make sure I didn't break something.

DonLakeFlyer:
@jaxxzer FYI. There seems to be some strangeness with respect to virtual joystick and reverse support. This pull corrects that. But if for some reason ArduSub doesn't quite match spec which is the reason #7375 was working then I just broke it. Since throttle now work per mavlink spec for reverse and validated using PX4 SITL Rover testing. I'll try to do some ArduPilot Rover testing as well to check that there isn't some difference there.

DonLakeFlyer:
Would be good to get an ArduSub test against these changes to make sure I didn't break anything.

DonLakeFlyer:
I tested this on ArduPilot Copter and Rover and it is working correctly. I'm going to merge now since this needs to go out ASAP since the current state is dangerous in ArduPilot when you mix RC and joystick. Not sure about PX4 yet.

PR #8790, opened 2020-05-28
Deploy custom start script for custom build

akontsevich:
2 fixes proposed for custom build:

- [x] Deploy custom start script for custom build
- [x] Add custom `QT_LIB_LIST` in custom build (prevent its overwriting in `QGCSetup.pri`)

### DIFF
--- a/QGCSetup.pri
+++ b/QGCSetup.pri
@@ -90,7 +90,7 @@ LinuxBuild {
     QMAKE_POST_LINK += && mkdir -p $$DESTDIR/Qt/libs && mkdir -p $$DESTDIR/Qt/plugins
 
     # QT_INSTALL_LIBS
-    QT_LIB_LIST = \
+    QT_LIB_LIST += \
         libQt5Charts.so.5 \
         libQt5Core.so.5 \
         libQt5DBus.so.5 \
@@ -160,7 +160,11 @@ LinuxBuild {
     }
 
     # QGroundControl start script
-    QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/deploy/qgroundcontrol-start.sh $$DESTDIR
-    QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/deploy/qgroundcontrol.desktop $$DESTDIR
-    QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/resources/icons/qgroundcontrol.png $$DESTDIR
+    contains (CONFIG, QGC_DISABLE_CUSTOM_BUILD) | !exists($$PWD/custom/custom.pri) {
+        QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/deploy/qgroundcontrol-start.sh $$DESTDIR
+        QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/deploy/qgroundcontrol.desktop $$DESTDIR
+        QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/resources/icons/qgroundcontrol.png $$DESTDIR
+    } else {
+        include($$PWD/custom/custom_deploy.pri)
+    }
 }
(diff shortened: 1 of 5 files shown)


QGCSetup.pri
```diff
@@ -90,7 +90,7 @@ LinuxBuild {
     QMAKE_POST_LINK += && mkdir -p $$DESTDIR/Qt/libs && mkdir -p $$DESTDIR/Qt/plugins
 
     # QT_INSTALL_LIBS
-    QT_LIB_LIST = \
+    QT_LIB_LIST += \
```
akontsevich, line 93: This prevents `QT_LIB_LIST` overwriting from `custom.pri`

QGCSetup.pri
```diff
@@ -160,7 +160,11 @@ LinuxBuild {
...
     # QGroundControl start script
-    QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/deploy/qgroundcontrol-start.sh $$DESTDIR
-    QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/deploy/qgroundcontrol.desktop $$DESTDIR
-    QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/resources/icons/qgroundcontrol.png $$DESTDIR
+    contains (CONFIG, QGC_DISABLE_CUSTOM_BUILD) | !exists($$PWD/custom/custom.pri) {
+        QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/deploy/qgroundcontrol-start.sh $$DESTDIR
+        QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/deploy/qgroundcontrol.desktop $$DESTDIR
+        QMAKE_POST_LINK += && $$QMAKE_COPY $$BASEDIR/resources/icons/qgroundcontrol.png $$DESTDIR
+    } else {
+        include($$PWD/custom/custom_deploy.pri)
```
akontsevich, line 168: Deploy custom start scripts for custom build

custom-example/custom.pri
```diff
@@ -55,6 +55,14 @@ RESOURCES += \
 QML_IMPORT_PATH += \
    $$PWD/res
 
+LinuxBuild {
+
+    # Our QT_INSTALL_LIBS
+    QT_LIB_LIST += \
+        libQt5VirtualKeyboard.so.5 \
```
akontsevich, line 62: Allow to add custom Qt libs

DonLakeFlyer:
What you want to achieve makes sense. But it would be much nicer if there was a way to instead get the startup script to be generated on the fly based on the existing variables or something like that. That way the guts of it could be shared from regular to custom. This change creates a lot of duplication in custom. Which means changes to the regular side would need to be copied to the custom side as well. Does that make sense. I think it would be mostly doable.

akontsevich:
> What you want to achieve makes sense. But it would be much nicer if there was a way to instead get the startup script to be generated on the fly based on the existing variables or something like that. That way the guts of it could be shared from regular to custom. This change creates a lot of duplication in custom. Which means changes to the regular side would need to be copied to the custom side as well. Does that make sense. I think it would be mostly doable.

Custom duplicates only 3 things:
- desktop file
- icon which is different
- start script which could be any custom content as well so we can't pre-generate it for all and this makes things less flexible for custom build users. While with `$$PWD/custom/custom_deploy.pri` we can deploy whatever we need without modifying QGC code.

If You still need some generation, could we make this in separate task, please? We need this change urgent.

DonLakeFlyer:
Ok, I'll deal with making this better later.

akontsevich:
> Ok, I'll deal with making this better later.

Thanks much, @DonLakeFlyer!!!

dnkagamarra:
Hi there!
@DonLakeFlyer are there any updates on this? I'd like to use a custom icon for MacOS & Windows (and eventually Linux) but I can't find anything on the docs about how to do it.
Thanks :D

DonLakeFlyer:
Windows already has custom icon support. If you look at the build files you should be able to find it. Other OS still do not have support for this out of the box

PR #8793, opened 2020-05-29
Cherry pick virtual joystick fix from Stable

DonLakeFlyer:
Fix for #8785 into master


MaEtUgR:
@DonLakeFlyer Thank you for fixing the virtual joystick. Awesome!

PR #8800, opened 2020-05-30
Toolbar tweaks

DonLakeFlyer:
* Fly View shows full tool bar
* Plan view shows only plan view toolbar
* Other views show view switching, values and comm loss
* View switching mechanism tweaks
* Log Replay toolbar has a Close button so you can close it without needing to go back to Settings


Antiheavy:
> Log Replay toolbar has a Close button so you can close it without needing to go back to Settings

hooray!

PR #8811, opened 2020-06-05
Help with Qt 5.15 integration

patrickelectric:


patrickelectric:
Ping @DonLakeFlyer

src/Camera/QGCCameraControl.cc
```diff
@@ -1274,17 +1274,26 @@ QGCCameraControl::_processConditionTest(const QString conditionTest)
     qCDebug(CameraControlVerboseLog) << "_processConditionTest(" << conditionTest << ")";
     int op = TEST_NONE;
     QStringList test;
+
+    auto split = [&conditionTest](const QString& sep ) {
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
```
DonLakeFlyer, line 1279: All of this ifdef version checking is just incredibly nasty. I would much rather just move Daily to 5.15 LTS first. And then clean up after. Have you been running 5.15 a bunch. Have you seen a reason to not just move to it?

patrickelectric: > All of this ifdef version checking is just incredibly nasty. I would much rather just move Daily to 5.15 LTS first. And then clean up after. Have you been running 5.15 a bunch. Have you seen a reason to not just move to it?

I'm using without problems right now for my work, I spend my friday night (corona virus feelings) polishing the PR and fixing a couple of issues. 

About the 5.15, we should take some time to fix all warning messages (a bunch in QML) for deprecated features, I have not done in this PR since my objetive was to compile and make it work, there is a couple of runtime warnings about deprecated features that we are using, this is really important since Qt 5.15 helps to catch up with Qt 6. Doing this now will help QGC to move to Qt 6 in the future without spending much time.

This also help us to create future proof code, creating code for future Qt versions.

Oh, this PR removes almost all warnings also, there is only a single warning message in my build, but it's related to the extern C code in the gstreamer pipeline configuration, I could check it later after this PR is merged, and finally, have a build without warnings in QGC.

DonLakeFlyer: > About the 5.15, we should take some time to fix all warning messages (a bunch in QML) for deprecated features

I agree with that. I just don't see any reason to have an ifdef'ed codebase which supports something less then 5.15. If there don't seem to be major problems with 5.15 then I can start doing the work to move the CI systems over to that.

patrickelectric: I did the ifdef to allow CI to pass, we could merge the patch in the way that it's now, and remove this and all ifdefs that exist in QGC related to Qt version, I saw a couple while working on it. 

Let me know what you prefer.

patrickelectric: And to help with that: https://github.com/mavlink/qgroundcontrol/pull/8815

PR #8814, opened 2020-06-08
Fix up windows warnings as errors

DonLakeFlyer:
Also includes initial work for `WarningsAsErrorsOff` support


deiuch:
I think, I have an idea on how to fix it appropriately for Windows. Warning that is threated as an error states that used character code in source file is not supported by the source character set (for example, mine is 1251). Adding flag `/utf-8` solves everything for me. But is this correct? Does project work only with UTF-8? I am new here :D

PR #8813, opened 2020-06-08
QGCMapPolygon select vertexes functionality

akontsevich:

PR #8820, opened 2020-06-09
appSettingsSpeedUnitsString() method added

akontsevich:
`appSettingsSpeedUnitsString()` method added to `QmlUnitsConversion` class.

### DIFF
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -1080,6 +1080,26 @@ const FactMetaData::AppSettingsTranslation_s* FactMetaData::_findAppSettingsArea
     return nullptr;
 }
 
+const FactMetaData::AppSettingsTranslation_s* FactMetaData::_findAppSettingsSpeedUnitsTranslation(const QString& rawUnits)
+{
+    for (size_t i=0; i<sizeof(_rgAppSettingsTranslations)/sizeof(_rgAppSettingsTranslations[0]); i++) {
+        const AppSettingsTranslation_s* pAppSettingsTranslation = &_rgAppSettingsTranslations[i];
+
+        if (rawUnits.toLower() != pAppSettingsTranslation->rawUnits.toLower()) {
+            continue;
+        }
+
+        uint settingsUnits = qgcApp()->toolbox()->settingsManager()->unitsSettings()->speedUnits()->rawValue().toUInt();
+
+        if (pAppSettingsTranslation->unitType == UnitSpeed
+                && pAppSettingsTranslation->unitOption == settingsUnits) {
+            return pAppSettingsTranslation;
+        }
+    }
+
+    return nullptr;
+}
+
 QVariant FactMetaData::metersToAppSettingsHorizontalDistanceUnits(const QVariant& meters)
 {
     const AppSettingsTranslation_s* pAppSettingsTranslation = _findAppSettingsHorizontalDistanceUnitsTranslation("m");
@@ -1200,7 +1220,7 @@ QVariant FactMetaData::appSettingsWeightUnitsToGrams(const QVariant& weight) {
 
 QString FactMetaData::appSettingsSpeedUnitsString()
 {
-    const AppSettingsTranslation_s* pAppSettingsTranslation = _findAppSettingsHorizontalDistanceUnitsTranslation("m/s");
+    const AppSettingsTranslation_s* pAppSettingsTranslation = _findAppSettingsSpeedUnitsTranslation("m/s");
     if (pAppSettingsTranslation) {
         return pAppSettingsTranslation->cookedUnits;
     } else {
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -1198,6 +1198,15 @@ QVariant FactMetaData::appSettingsWeightUnitsToGrams(const QVariant& weight) {
     }
 }
 
+QString FactMetaData::appSettingsSpeedUnitsString()
+{
+    const AppSettingsTranslation_s* pAppSettingsTranslation = _findAppSettingsHorizontalDistanceUnitsTranslation("m/s");
+    if (pAppSettingsTranslation) {
+        return pAppSettingsTranslation->cookedUnits;
+    } else {
+        return QStringLiteral("m/s");
+    }
+}
 
 double FactMetaData::cookedIncrement(void) const
 {
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -1033,9 +1033,6 @@ const FactMetaData::AppSettingsTranslation_s* FactMetaData::_findAppSettingsUnit
         case UnitWeight:
             unitOption = unitsSettings->weightUnits()->rawValue().toUInt();
             break;
-        default:
-            assert(false);
-            break;
         }
 
         if (pAppSettingsTranslation->unitType == type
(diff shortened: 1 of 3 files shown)


akontsevich:
@DonLakeFlyer what is wrong with Ci again? :slightly_smiling_face:

dogmaphobic:
> @DonLakeFlyer what is wrong with Ci again? 🙂

Unittest crashed:
```
./debug/qgroundcontrol-start.sh --unittest
Settings location "/home/travis/.config/QGroundControl.org/QGroundControl_unittest.ini" Is writable?: true
Filter rules "*Log.debug=false\n"
System reported locale: QLocale(English, Latin, United States) ; Name "en_US" ; Preffered (used in maps):  "en-US"
Request for unknown firmware plugin factory -1
Aborted (core dumped)
```

akontsevich:
> Unittest crashed:

@dogmaphobic I saw above - this tells me nothing as ALL TESTS PASSED on my side. Also my change should not affect any tests.

akontsevich:
I see same multiple bugs here: https://travis-ci.org/github/mavlink/qgroundcontrol/pull_requests and here: https://travis-ci.org/github/mavlink/qgroundcontrol/jobs/697396623 - seems something with the CI.

dogmaphobic:
> seems something with the CI.

Yup, I restarted the build and it still the same.

The last error before the crash came from here:
https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/FirmwarePluginManager.cc#L52
Or here:
https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/FirmwarePluginManager.cc#L66

akontsevich:
> > seems something with the CI.
> 
> Yup, I restarted the build and it's still the same.
> 
> The last error before the crash came from here:
> https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/FirmwarePluginManager.cc#L52
> Or here:
> https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/FirmwarePluginManager.cc#L66

@dogmaphobic that is not mine, came from previous PRs, strange CI did not crash there or CI was ignored on merge. Like @DonLakeFlyer [said in previous PR](https://github.com/mavlink/qgroundcontrol/pull/8813#issuecomment-641399487) unit tests run with QT_FATAL_WARNINGS on. So these warnings should be wrapped into 

```
    if (!qgcApp()->runningUnitTests()) {
            qWarning() << ...
    }
```

BTW, all unittests runs and QGC runs are littered a lot now by these warnings, which is very inconvenient. Could we turn them off or fix their cause? Thanks!

akontsevich:
@dogmaphobic You see the same in all others pull requests: https://travis-ci.org/github/mavlink/qgroundcontrol/pull_requests

akontsevich:
FYI, rebased to recent master CI fixes.

akontsevich:
ALL TESTS PASSED now!!! :slightly_smiling_face:

DonLakeFlyer:
Could you help out a little more here? The fact that the majority of the code in FactMetaData::_findAppSettings*Translation implementations is common means that it could be refactored into a single worker function which takes as a parameter the units type you are looking for. And then the specific methods just call the worker appropriately. Thus reducing the possibility of bugs caused by needing to change the common code and forgetting to do it in multiples places.

akontsevich:
> Could you help out a little more here? The fact that the majority of the code in FactMetaData::_findAppSettings*Translation implementations is common means that it could be refactored into a single worker function which takes as a parameter the units type you are looking for. And then the specific methods just call the worker appropriately. Thus reducing the possibility of bugs caused by needing to change the common code and forgetting to do it in multiples places.

I agree, ok: https://github.com/mavlink/qgroundcontrol/pull/8820/commits/d3faaf5ee981572899b3786b276269261f095ffc - tested.

src/FactSystem/FactMetaData.cc
```diff
@@ -1012,66 +1012,32 @@ const FactMetaData::AppSettingsTranslation_s* FactMetaData::_findAppSettingsHori
...
-{
-    for (size_t i=0; i<sizeof(_rgAppSettingsTranslations)/sizeof(_rgAppSettingsTranslations[0]); i++) {
-        const AppSettingsTranslation_s* pAppSettingsTranslation = &_rgAppSettingsTranslations[i];
-
-        if (rawUnits.toLower() != pAppSettingsTranslation->rawUnits.toLower()) {
-            continue;
+        uint settingsUnits = 0;
+        switch (type) {
+        case UnitHorizontalDistance:
+            settingsUnits = qgcApp()->toolbox()->settingsManager()->unitsSettings()->horizontalDistanceUnits()->rawValue().toUInt();
```
DonLakeFlyer, line 1018: Ok, just one last thing please. Can you move `qgcApp()->toolbox()->settingsManager()->unitsSettings()` into a variable as well to deal with the same repetition leads to bugs problem.

akontsevich: Absolutely agree, looks much better: https://github.com/mavlink/qgroundcontrol/pull/8820/commits/78cdbbbd0e2d3ef718a23b1ca5d0ae61481eda8d

src/FactSystem/FactMetaData.cc
```diff
@@ -1012,67 +1012,34 @@ const FactMetaData::AppSettingsTranslation_s* FactMetaData::_findAppSettingsHori
...
+            unitOption = unitsSettings->speedUnits()->rawValue().toUInt();
+            break;
+        case UnitTemperature:
+            unitOption = unitsSettings->temperatureUnits()->rawValue().toUInt();
+            break;
+        case UnitWeight:
+            unitOption = unitsSettings->weightUnits()->rawValue().toUInt();
+            break;
+        default:
+            assert(false);
```
DonLakeFlyer, line 1037: Oops sorry, just saw this now. You can't used asserts in QGC code. That causes the app to crash, which is never a good idea for a GCS. For switch statements QGC compiles will fail build if not all cases are covered and there is no default. So there is no need for a default here. If a new unit gets added later the build will fail until this switch is updated.

Also just FYI: The way to handle switch statements for which a default case is needed since not all the cases over covered. And for which that default indicates an error, the general pattern is to use qWarning with text spitting out "Internal Error: ..." to indicate a coding error.

akontsevich: We used assert not to indicate all cases are covered or not but to check if no other value will be passed which is an error. Additional program logic check. Above `_setAppSettingsTranslators()` method contains default for some reason which is not correct. But ok, changed to your coding standard: https://github.com/mavlink/qgroundcontrol/pull/8820/commits/8a33a574431d0249f7ab3236a035d00aceafa97f

DonLakeFlyer: "We used assert not to indicate all cases are covered or not but to check if no other value will be passed which is an error. "

The variable is typed as UnitsType which means it will be a compiler error to get anything else. I guess you could cast a crap value to a UnitsType which a default will catch. But to me, it's better to have the compiler complain when a new type is added to let you know the switch must be updated than handle that case.

akontsevich: while enums widely used in QML we won't get compiler errors in many cases ;)

DonLakeFlyer: The way to catch something like that is to validate the value at the point of entry with separate code. That way you get both validation and compiler errors with new enums added. As it currently stands QGC does not generally validate enums which come in from the Qml side. It unfortunately expects Qml code to be correct which isn't the best thing.

PR #8831, opened 2020-06-12
Softlock workaround fix

joshanne:
This allows any dialog component to nest itself up to twice.

The simplest example is.
* Drop a way point
* On the right, click the hamburger
* Click Edit Position
* Click the "**?**" button on one of the facts (eg. Lat / Lon)

Prior to #8826, this would have resulted in a soft lock.

The current implementation on master results in the "ParamaterEditorDialog" not being shown.

This proposed implementation allows to display a dialog within a dialog (which appears to be the maximum nesting I have found)

### DIFF
--- a/src/PlanView/MissionItemEditor.qml
+++ b/src/PlanView/MissionItemEditor.qml
@@ -218,7 +218,10 @@ Rectangle {
                 QGCMenuItem {
                     text:           qsTr("Edit position...")
                     visible:        missionItem.specifiesCoordinate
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Ah, we had thought there weren't any cases left where a dialog popped another dialog. Oops. The way to fix this is to instantiate Drawers on the fly as needed. That way you can stack as many as you want. I would rather do that than just move the limit from 1 to 2.

joshanne:
Indeed, that's the appropriate fix. I'll update this shortly.

DonLakeFlyer:
I'm working on refactoring this to fix the problem

PR #8835, opened 2020-06-13
LinkInterface/Vehicle uses mutexes to synchronize writing bytes

DonLakeFlyer:
Previously writing bytes out to a link went through a sequence of multiple queued messages in Vehicle and LinkInterface to prevent thread safety issues. The problem is that doing it that way is dog slow. So slow that in cases where a joystick was being used on a device with a very slow CPU the MANUAL_CONTROL messages didn't make it through to the vehicle at a high enough frequency. It also chew more CPU than needed.

What happens now is that both Vehicle and LinkInterface use mutexes for thread safety in the following calls which are threadsafe:
* LinkInterface::writeBytesThreadSafe
* Vehicle::sendMessageOnLinkThreadSafe
* Vehicle::sendJoystickDataThreadSafe - Joystick calls this directly instead of posting queued signal

The end result is Mavlink message sending QGC->Vehicle which is measured in nanoseconds instead of 100s of milliseconds. And also when using a joystick a significant reduction in CPU usage.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -59,7 +59,6 @@ QGC_LOGGING_CATEGORY(VehicleLog, "VehicleLog")
 const QString guided_mode_not_supported_by_vehicle = QObject::tr("Guided mode not supported by Vehicle.");
 
 const char* Vehicle::_settingsGroup =               "Vehicle%1";        // %1 replaced with mavlink system id
-const char* Vehicle::_joystickModeSettingsKey =     "JoystickMode";
 const char* Vehicle::_joystickEnabledSettingsKey =  "JoystickEnabled";
 
 const char* Vehicle::_rollFactName =                "roll";
@@ -116,7 +115,6 @@ Vehicle::Vehicle(LinkInterface*             link,
     , _soloFirmware(false)
     , _toolbox(qgcApp()->toolbox())
     , _settingsManager(_toolbox->settingsManager())
-    , _joystickMode(JoystickModeRC)
     , _joystickEnabled(false)
     , _uas(nullptr)
     , _mav(nullptr)
@@ -235,7 +233,6 @@ Vehicle::Vehicle(LinkInterface*             link,
 
     _addLink(link);
 
-    connect(this, &Vehicle::_sendMessageOnLinkOnThread, this, &Vehicle::_sendMessageOnLink, Qt::QueuedConnection);
     connect(this, &Vehicle::flightModeChanged,          this, &Vehicle::_handleFlightModeChanged);
     connect(this, &Vehicle::armedChanged,               this, &Vehicle::_announceArmedChanged);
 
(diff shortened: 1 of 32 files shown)


src/Vehicle/Vehicle.cc
```diff
@@ -2104,40 +2100,26 @@ void Vehicle::_linkInactiveOrDeleted(LinkInterface* link)
...
 {
-    if (!link || !_links.contains(link) || !link->isConnected()) {
-        return false;
-    }
-
-    emit _sendMessageOnLinkOnThread(link, message);
-
-    return true;
-}
+    QMutexLocker lock(&_sendMessageOnLinkMutex);
```
dogmaphobic, line 2106: Is this lock needed given that `link->writeBytesThreadSafe()` has its own lock? What is it protecting?

DonLakeFlyer: Makes sure this stuff is thread safe as well:
```
    if (!link->isConnected()) {
        return false;
    }

    // Give the plugin a chance to adjust
    _firmwarePlugin->adjustOutgoingMavlinkMessage(this, link, &message);
```

DonLakeFlyer: Since the joystick send will come through this path from another thread.

DonLakeFlyer:
Also  there is gimbal control signal which when that gets fully implemented should switch to using this path as well.

src/Vehicle/Vehicle.cc
DonLakeFlyer: This `link->isConnected()` probably isn't too problematic. But this `_firmwarePlugin->adjustOutgoingMavlinkMessage` is since it maintains a mavlink channel status if I remember correctly.

DonLakeFlyer:
I should have also mentioned that this pull removes all the advanced joystick modes. Only joystick usage for RC is now supported.

DonLakeFlyer:
Crap, closed by accident.

src/Vehicle/Vehicle.cc
dogmaphobic: Yup, It doesn't look like it is needed. The `adjustOutgoingMavlinkMessage()` function is 99.9% void. The only time something actually is done is for [this](https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc#L511-L518):
```
    case MAVLINK_MSG_ID_PARAM_SET:
        _handleOutgoingParamSet(vehicle, outgoingLink, message);
```
Might as well put a mutex in there rather than annoying the other 99.9% of the time this never used.

DonLakeFlyer: The problem is that it's a firmwareplugin override so you have no idea what can happen in there in a custom build. Without a change to the FirmwarePlugin method to mark as threadsafe required and likely a name change to prevent old unthreadsafe code from leaking through it needed. I can certainly do that. Seems like a better way to go intstead of the upper layers doing things which won't be needed most of the time.

DonLakeFlyer:
@dogmaphobic All good now?

dogmaphobic:
> @dogmaphobic All good now?

Yup! :)

lukegluke:
@DonLakeFlyer, @dogmaphobic this one leads to https://github.com/mavlink/qgroundcontrol/issues/9751

PR #8838, opened 2020-06-14
typo in roiModeSupported  comment 

roiegeron-spearuav:
Fixed small typo in comment

### DIFF
--- a/src/FlightDisplay/FlyViewMap.qml
+++ b/src/FlightDisplay/FlyViewMap.qml
@@ -533,13 +533,13 @@ FlightMap {
         }
 
         onClicked: {
-            if (guidedActionsController.guidedUIVisible || (!guidedActionsController.showGotoLocation && !guidedActionsController.showOrbit)) {
+            if (guidedActionsController.guidedUIVisible || (!guidedActionsController.showGotoLocation && !guidedActionsController.showOrbit && !guidedActionsController.showROI)) {
                 return
             }
             orbitMapCircle.hide()
             gotoLocationItem.hide()
             var clickCoord = _root.toCoordinate(Qt.point(mouse.x, mouse.y), false /* clipToViewPort */)
-            if (guidedActionsController.showGotoLocation && guidedActionsController.showOrbit) {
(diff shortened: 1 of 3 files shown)


src/FlightDisplay/FlyViewMap.qml
```diff
@@ -533,13 +533,13 @@ FlightMap {
...
         onClicked: {
-            if (guidedActionsController.guidedUIVisible || (!guidedActionsController.showGotoLocation && !guidedActionsController.showOrbit)) {
+            if (guidedActionsController.guidedUIVisible || (!guidedActionsController.showGotoLocation && !guidedActionsController.showOrbit && !guidedActionsController.showROI)) {
                 return
             }
             orbitMapCircle.hide()
             gotoLocationItem.hide()
             var clickCoord = _root.toCoordinate(Qt.point(mouse.x, mouse.y), false /* clipToViewPort */)
-            if (guidedActionsController.showGotoLocation && guidedActionsController.showOrbit) {
+            if (guidedActionsController.showGotoLocation && (guidedActionsController.showOrbit || guidedActionsController.showROI)) {
```
DonLakeFlyer, line 542: This isn't correct. It won't do the right thing if goto is not shown but orbit and roi are. This block should be change like this:

```
if (guidedActionsController.showGotoLocation && !guidedActionsController.showOrbit && !guidedActionsController.showROI) {
    ...
}  else if (guidedActionsController.showOrbit && ! guidedActionsController.showGotoLocation && !guidedActionsController.showROI) {
    ....
} else if (...) {
    ....
} else {
    Handle the case where a menu needs to be shown to pick from
}

roiegeron-spearuav: of goto is enabler for ROI and Orbit then the if case is very simple
I made the changes and committed

DonLakeFlyer: This is still incorrect since it doesn't follow what I showed above.

DonLakeFlyer:
In looking through this I decided to simplify the menu handling to always show to menu even when only a single item. I also found other bugs in _hideROI and _hideOrbit and fixed them and replaced this pull with #8893. Thanks for finding the problem.

Issue #8839, opened 2020-06-14
New Feature: Altitude mode which applies to entire plan


DonLakeFlyer:
This is what the waypoint editor for a Plan looks like:
![Screen Shot 2020-06-14 at 1 56 44 PM](https://user-images.githubusercontent.com/5876851/84603907-ff285b00-ae46-11ea-9ae0-1a4389a8f2a3.png)

Note the big giant altitude setting control which allows you to set the altitude as well as set the altitude frame that you want to use. This same big UI shows up in an mission item which includes an altitude. The reason being, mavlink supports setting the altitude frame on each individual mission item. Hence QGC supports it as well.

The problem is that what mavlink support and what end users use in reality can be pretty far apart. Way more than the vast majorty of user create plans for which all altitudes are relatives. For these folks this big, pick your alt frame from hell ui is confusing and not needed.

In order to simplify things QGC is moving to a model where you can specific an altitude mode for an entire mission. This is done through Mission Start and defaults to relative:
![Screen Shot 2020-06-14 at 2 01 17 PM](https://user-images.githubusercontent.com/5876851/84603958-8c6baf80-ae47-11ea-84c5-cbdd6420cf0c.png)

Now if you add anything which has an altitude, like a waypoint you get this. Much simple and easier to understand for most users:
![Screen Shot 2020-06-14 at 2 01 58 PM](https://user-images.githubusercontent.com/5876851/84603970-a86f5100-ae47-11ea-871f-54aaf69e6baf.png)

The options for the plan wide altitude mode are the same as what you saw in the big altitude ui:
![Screen Shot 2020-06-14 at 2 04 41 PM](https://user-images.githubusercontent.com/5876851/84604018-03a14380-ae48-11ea-8d1d-9b3282f4e728.png)

If you pick a global alt mode other than relative you still get the little text hits about what the alt fields are doing:
![Screen Shot 2020-06-14 at 2 06 47 PM](https://user-images.githubusercontent.com/5876851/84604083-5a0e8200-ae48-11ea-8158-ff09e97b5144.png)

If you pick "Mixed Modes" you are back to the big ass alt mode ui on each item:
![Screen Shot 2020-06-14 at 2 07 55 PM](https://user-images.githubusercontent.com/5876851/84604106-875b3000-ae48-11ea-8304-11dffb0c1c28.png)

So to wrap up for most users they likely won't even notice this change since they were always using relative anyway. The UI just got simpler for them.

[...]

PR #8840, opened 2020-06-14
Initial support for global plan-wide altitude mode setting

DonLakeFlyer:
Details are here: #8839


DonLakeFlyer:
@antiheavy I believe this is something we had discussed somewhere along the way.

PR #8841, opened 2020-06-15
Fix vertical units oversight

remster:
Very simple and embarrassing oversight i have left behind [here](https://github.com/mavlink/qgroundcontrol/pull/8692).


remster:
now redundant with https://github.com/mavlink/qgroundcontrol/pull/8820/commits/053140562fb2713eacacffa8105891a062a6bb6d

PR #8843, opened 2020-06-15
Transect style re factor

DonLakeFlyer:
* Refactored the tab bar and terrain follow from Survey into reusable TransectStyleComplexItem common controls. This is prep for terrain follow support in Corridor Scan which is also a TransectStyleComplexItem.
* Move Terrain Follow settings into its own tab
* Change tabs from text to icons to fit more tabs. Grid - Camera - Terrain - Presets
* Related to #8339

Screen Shot 2020-06-15 at 10 30 27 AM

Screen Shot 2020-06-15 at 10 30 37 AM


Antiheavy:
Looks nice.

PR #8846, opened 2020-06-17
Add a follow terrain test

remster:
This PR introduces one, basic, not awfully valuable, but first nonetheless test that puts `TransectStyleComplexItem::_followTerrain` under stress. 

It does so by introducing two fake regions around [Point Nemo](https://oceanservice.noaa.gov/facts/nemo.html), where they have little chance of colliding with real terrain - should that ever needed to be put under test and checked in as a terrain tile:
-  [flat region at 10m above ASL](https://github.com/mavlink/qgroundcontrol/pull/8846/files#diff-3c691d138cca295bea62f8d66ae860abR317);
- [slope region](https://github.com/mavlink/qgroundcontrol/pull/8846/files#diff-3c691d138cca295bea62f8d66ae860abR329) rising (east to west) from -100 to 1000m above ASL.

Each region is a bounding box 0.1 x 0.1 degrees across (about 11km x 11km if it were on the equator). I argue that the uniformity of these regions, simplicity and extensibility mean that a lot of tests can be written against this. I admit an individual test case should be able to command the roughness of the terrain it requires, but most test cases don't and if they do - i think this can be extended to cover custom terrains either by adding more regions (sine wave region, valley, mountain) and if that's not enough, tests can check in specific terrain tiles. The latter may be an argument to move this lower down the stack to Qt Location Plugin. I thought that wasn't necessary at this stage and it's best to consider plugin a plugin - but I am open to moving it. 

This PR is related to [this](https://github.com/mavlink/qgroundcontrol/pull/8682) rejection and the necessity I face to migrate to AMSL - which I have to test.

### DIFF
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -331,6 +331,7 @@ VisualMissionItem* MissionController::_insertSimpleMissionItemWorker(QGeoCoordin
             }
         }
     }
+    newItem->setMissionFlightStatus(_missionFlightStatus);
     if (visualItemIndex == -1) {
         _visualItems->append(newItem);
     } else {
@@ -371,6 +372,7 @@ VisualMissionItem* MissionController::insertTakeoffItem(QGeoCoordinate /*coordin
             newItem->setAltitudeMode(static_cast<QGroundControlQmlGlobal::AltitudeMode>(prevAltitudeMode));
         }
     }
+    newItem->setMissionFlightStatus(_missionFlightStatus);
     if (visualItemIndex == -1) {
         _visualItems->append(newItem);
     } else {
@@ -1884,9 +1886,8 @@ void MissionController::_initVisualItem(VisualMissionItem* visualItem)
     connect(visualItem, &VisualMissionItem::specifiedVehicleYawChanged,                 this, &MissionController::_recalcMissionFlightStatusSignal, Qt::QueuedConnection);
     connect(visualItem, &VisualMissionItem::terrainAltitudeChanged,                     this, &MissionController::_recalcMissionFlightStatusSignal, Qt::QueuedConnection);
     connect(visualItem, &VisualMissionItem::additionalTimeDelayChanged,                 this, &MissionController::_recalcMissionFlightStatusSignal, Qt::QueuedConnection);
-    connect(visualItem, &VisualMissionItem::lastSequenceNumberChanged,                  this, &MissionController::_recalcSequence);
 
-    visualItem->setMissionFlightStatus(_missionFlightStatus);
+    connect(visualItem, &VisualMissionItem::lastSequenceNumberChanged,                  this, &MissionController::_recalcSequence);
 
     if (visualItem->isSimpleItem()) {
         // We need to track commandChanged on simple item since recalc has special handling for takeoff command
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -331,7 +331,6 @@ VisualMissionItem* MissionController::_insertSimpleMissionItemWorker(QGeoCoordin
             }
         }
     }
-    newItem->setMissionFlightStatus(_missionFlightStatus);
     if (visualItemIndex == -1) {
         _visualItems->append(newItem);
     } else {
@@ -372,7 +371,6 @@ VisualMissionItem* MissionController::insertTakeoffItem(QGeoCoordinate /*coordin
             newItem->setAltitudeMode(static_cast<QGroundControlQmlGlobal::AltitudeMode>(prevAltitudeMode));
         }
     }
-    newItem->setMissionFlightStatus(_missionFlightStatus);
     if (visualItemIndex == -1) {
         _visualItems->append(newItem);
     } else {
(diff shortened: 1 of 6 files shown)


src/MissionManager/MissionController.cc
```diff
@@ -1886,9 +1884,10 @@ void MissionController::_initVisualItem(VisualMissionItem* visualItem)
     connect(visualItem, &VisualMissionItem::specifiedVehicleYawChanged,                 this, &MissionController::_recalcMissionFlightStatusSignal, Qt::QueuedConnection);
     connect(visualItem, &VisualMissionItem::terrainAltitudeChanged,                     this, &MissionController::_recalcMissionFlightStatusSignal, Qt::QueuedConnection);
     connect(visualItem, &VisualMissionItem::additionalTimeDelayChanged,                 this, &MissionController::_recalcMissionFlightStatusSignal, Qt::QueuedConnection);
-
     connect(visualItem, &VisualMissionItem::lastSequenceNumberChanged,                  this, &MissionController::_recalcSequence);
 
+    visualItem->setMissionFlightStatus(_missionFlightStatus);
```
remster, line 1889: I have moved this here as it wasn't invoked for the complex items and rendered `TransectStyleComplexItem::_missionFlightStatus` uninitialised and in turn [this code](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/TransectStyleComplexItem.cc#L712) didn't do anything useful.

src/MissionManager/TransectStyleComplexItemTest.cc
```diff
@@ -226,21 +209,57 @@ void TransectStyleComplexItemTest::_testAltMode(void)
...
+TestTransectStyleItem::TestTransectStyleItem(PlanMasterController* masterController, QObject* parent)
     : TransectStyleComplexItem      (masterController, false /* flyView */, QStringLiteral("UnitTestTransect"), parent)
     , rebuildTransectsPhase1Called  (false)
     , recalcComplexDistanceCalled   (false)
     , recalcCameraShotsCalled       (false)
 {
-
+    // We use a 100m by 100m square test polygon
+    const double edgeDistance = 100;
+    surveyAreaPolygon()->appendVertex(UnitTestTerrainQuery::linearSlopeRegion.center());
```
remster, line 239: this polygon lies in the slope region and thus in can count on emulated terrain values in the test cycle.

src/MissionManager/MissionController.cc
DonLakeFlyer: So the fact that setMissionFlightStatus is called in the insert workers likely incorrect. I'd have to go back to see why they were added. The problem is that there are values in _missionFlightStatus which are dynamic and are specific to the point in the recalc sequence. For example vehicle speeds can be adjusted by mission items. So they are tracked as the recalc is done and the current status is stuffed into the items.

Also the problem you are seeing with TransectStyleComplexItem actually indicates a different bug. Which is that if the vehicle speed given to TransectStyleComplesItem changes then terrain follow must be fully rebuilt. Since the vehicle speed affects the generation of interstitial waypoints. Which if you look here:
```
void TransectStyleComplexItem::setMissionFlightStatus(MissionController::MissionFlightStatus_t& missionFlightStatus)
{
    ComplexMissionItem::setMissionFlightStatus(missionFlightStatus);
    if (!qFuzzyCompare(_cruiseSpeed, missionFlightStatus.vehicleSpeed)) {
        _cruiseSpeed = missionFlightStatus.vehicleSpeed;
        emit timeBetweenShotsChanged();
    }
}
```

You'll see that isn't happening.

I'll take a look at figuring what is they correct way to deal with is. Until then this pull will have to wait those fixes. And you should take out you changed for setMissionFlightStatus so my changes when they are ready will merge in.

remster: done: https://github.com/mavlink/qgroundcontrol/pull/8846/commits/f38ab2610945bd8dd9a8cc7920369973337f66f8

src/MissionManager/TransectStyleComplexItem.cc
```diff
@@ -51,7 +51,7 @@ TransectStyleComplexItem::TransectStyleComplexItem(PlanMasterController* masterC
     , _terrainAdjustMaxClimbRateFact    (settingsGroup, _metaDataMap[terrainAdjustMaxClimbRateName])
     , _terrainAdjustMaxDescentRateFact  (settingsGroup, _metaDataMap[terrainAdjustMaxDescentRateName])
 {
-    _terrainQueryTimer.setInterval(_terrainQueryTimeoutMsecs);
+    _terrainQueryTimer.setInterval(qgcApp()->runningUnitTests() ? 0 : _terrainQueryTimeoutMsecs);
```
DonLakeFlyer, line 54: So this timer is used to batch up repeated calls to query terrain as well as de-dup calls. This can happen from an end user futzing with things in the ui as well as the internal code itself not being super careful about asking for the same thing more than once. The first case can't happen in a unit test, but the second case of internal code asking for the same thing multiple times can. So in order to replicate what actually happens in the real world the unit test timeout should be something like say 10 msecs. That way the test past ends up going all the way out to the main loop and then back when the timer fires. Which in turn replicates what happen in the normal non unit test code path.

remster: good point, done: https://github.com/mavlink/qgroundcontrol/pull/8846/commits/50180bf88d353a04bd70df71f92d8c21c005dd0e

PR #8848, opened 2020-06-20
Cube orange yellow

DonLakeFlyer:
The Orange and Yellow Cubes expose a composite usb device over a single usb physical connections. The first of the ports is the mavlink connection. The second connection by default is SLCAN and QGC should not auto-connect to it.


DonLakeFlyer:
Related to #8789

PR #8849, opened 2020-06-20
Remove unused Shapes module causing AppImage to fail

DonLakeFlyer:


Noircogi:
This commit seems to break the IOS build.
The label Stable_V4.0 builds fine.
Syncing to this PR causes build failures.
I found it trying to build an IOS version with compass priority support. Nothing after Stable_V4.0 builds.

DonLakeFlyer:
I just built an XCode project using this:
* `~/Qt/5.12.6/ios/bin/qmake -r ~/repos/qgroundcontrol/qgroundcontrol.pro`
* Built it in XCode
* Ran it in the simulator
* Connected to a MockLink
* Went to Console

All worked fine. Are you using 5.12.6?

Noircogi:
Thank you Don.  Yes, that method works fine.
I was able to build top of tree and verify working compass priority on IOS.
In the past, I've used QTCreator to generate the .xcodeproj file.  It works prior to this PR.

PR #8858, opened 2020-06-23
Fix Windows build with wrong character set

deiuch:
See reverted #8814.


DonLakeFlyer:
I don't understand what problem this fixes?

deiuch:
> I don't understand what problem this fixes?



> I don't understand what problem this fixes?

The project did not build on MSVC due to "warnings as errors", and this flag fixes the warning that blocked the building process. I was able to fix the build with this flag.

DonLakeFlyer:
Are you running latest master? If so, what language is your machine set to?

DonLakeFlyer:
@deiuch I believe the bad characters in source files have been removed such that this is no longer needed. Can you retry building latest master.

PR #8867, opened 2020-06-30
Remove windows only zlib

DonLakeFlyer:


hamishwillee:
@julianoes This isn't what MAVSDK was expecting to exist for component information file decompression right?

julianoes:
@hamishwillee yes it is but I don't know what this change exactly does.

DonLakeFlyer:
What does MAVSDK have to do with QGC code?

dogmaphobic:
I'm assuming the discussion is around mavftp compression. Don removed this binary, which is only for Windows, to replace it with a submodule to provide zlib to all platforms.

DonLakeFlyer:
Correct. Which also I believe turned out to be wrong to remove since zlib is supposedly available by default on all platforms except Windows. Hence you only need a windows lib. I'm working on verifying that right now to see if it will work.

julianoes:
Sorry, it's not about MAVSDK. It's just the MAVLink spec.

PR #8873, opened 2020-07-02
Change gradient indicator calculation from degrees to percent

mattstraehl:
Came across this as we copied some of the indicator logic to a new AMC widget. If my understanding is correct and "gradient" is basically the same as "slope" and we want to display it in percent, I think the calculation should be `100 * altDifference / distance`.

Alternatively the `_gradientText` unit [on line 55](https://github.com/mavlink/qgroundcontrol/blob/master/src/PlanView/PlanToolBarIndicators.qml#L55) could be changed to degrees.

### DIFF
--- a/src/PlanView/PlanToolBarIndicators.qml
+++ b/src/PlanView/PlanToolBarIndicators.qml
@@ -38,7 +38,7 @@ Item {
     property real   _rowSpacing:                ScreenTools.isMobile ? 1 : 0
     property real   _distance:                  _statusValid && _currentMissionItem ? _currentMissionItem.distance : NaN
     property real   _altDifference:             _statusValid && _currentMissionItem ? _currentMissionItem.altDifference : NaN
-    property real   _gradient:                  _statusValid && _currentMissionItem && _currentMissionItem.distance > 0 ? (Math.atan(_currentMissionItem.altDifference / _currentMissionItem.distance) * (180.0/Math.PI)) : NaN
+    property real   _gradient:                  _statusValid && _currentMissionItem && _currentMissionItem.distance > 0 ? 100 * _currentMissionItem.altDifference / _currentMissionItem.distance : NaN
     property real   _azimuth:                   _statusValid && _currentMissionItem ? _currentMissionItem.azimuth : NaN
     property real   _heading:                   _statusValid && _currentMissionItem ? _currentMissionItem.missionVehicleYaw : NaN
     property real   _missionDistance:           _missionValid ? missionDistance : NaN


DonLakeFlyer:
Yeah, this has gone back/forth without ever being correct multiple times! I wonder what is more understandable percent of degrees though.

@antiheavy, @AndKe I think you guys had some original input here

mattstraehl:
Good question what is more understandable. For what it's worth, the traffic signs that indicate slope use percent :)

DonLakeFlyer:
Probably @sanderux as well. Since their mission validator does stuff with max gradients.

AndKe:
Incline in percent is common on steeper European/Scandinavian roads. 
In manned aviation, normal descent/climb rate is still most often ft/min,  given that we always use SI units on UAVs the normal is m/min 
The climb rate limiting us is at Vy (velocity for best climb rate)  when planning, we know that if you need to clear a 1000m mountain in 10min then a climb rate of 100m/min is not necessary sufficient ..in case of tailwind it may be much too little (or plenty with good headwind)

But still, we calculate based on known rate of climb/descent given as m/min - not percentage nor degrees.
Usually one would circle to a target altitude well before taking chances.

During approach/landing, same is used (m/min)  - the maximum sink rate is limited by Vne, but we would think of glide ratio in case of loss of propulsion..   and yet again, the practical outcome in terms of ground distance travelled is very affected by wind.

To properly estimate m/min between waypoints limit we would need to take into account cruise speed ,max speed(on case of descend), and some TECS parameters - then the user could compare that with POH and a glance know the likelihood of performing that maneuver.

Degrees and percent is basically the same thing, with the same lack of "complete picture"  I do agree that percent may be a little more easy to understand/visualize. 
The ultimate would be to calculate based on approximated actual performance.  If not connected, then only cruise speed.

The regulations for UAV's are getting more and more strict, there is no way around trying to adopt much of the established aviation terminology - it is a good thing to try to stick to the format of ft/minute or meter/minute for this.

Antiheavy:
I don't have a strong opinion about how gradient is handled or calculated between waypoints with the following exception:
- the Fixed Wing Landing Pattern edit dialog should still display "glide slope" to the user measured in degrees.  In aviatin, "degrees" is the most common unit of measure for landing glide slopes.  It is also the unit of measure used by both PX4 and Arduplane firmware parameters that control fixed wing landing glide slopes.
image

sanderux:
Although i dont have a strong opinion on which method is used, i care about consistency. Operation manuals should specify some sort of maximum (even if a rough estimate derived from m/s). Constantly changing this will make it impossible to explain to users or properly set limits.

I propose showing both percent and estimated climb rate (for consistency i would use m/s, not m/min).

DonLakeFlyer:
So previous Stable builds showed gradient as "%" even though the value it was showing was actually degrees! And then in master things got out of whack. Since stable was degrees I"m going to put the value back to degrees and show the units correctly as degrees.

PR #8874, opened 2020-07-02
Add high latency handle

patrickelectric:
Add HIGH_LATENCY message handle


patrickelectric:
@DonLakeFlyer

DonLakeFlyer:
What's the difference between these two messages? Is there something lost from using the old one?

patrickelectric:
Hi Don, sorry I was OOO from the last couple of days.
Yes, we lose a bit of information like failure2Sensor, vdop, hdop and others non so important info, this message is a bit simpler but has the basic variables to allow mission status in QGC.

dagar:
The original `HIGH_LATENCY` should have been marked WIP, but we weren't doing that at the time.

Issue #8853, opened 2020-06-21
RC channels 15 and 16 not changing


lucasdemarchi:
On the Radio Setup screen there's no feedback for channels 15 and 16. I noticed this while implementing those channels on https://github.com/lucasdemarchi/dema-rc/

Initially I thought it was a bug on my RC sw stack, then on ArduPilot end. Finally got to check with mavproxy and I can see the values changing there. So I suspect there's something in QGC not updating the values.

I used QGC nightly to test.

DonLakeFlyer:
As far as I know ArduPilot only used to support up to 14 channels. Do you know when 15/16 support was added? And is the new support only for 16 instead of 18?

lucasdemarchi:
Number of channels depend on the board. The Linux boards always supported max 16 channels. The stm32 boards currently support 18 channels and I don't see any board overriding that. On a quick look this has always been the case since the move to Chibios for its HAL around January 2018

lucasdemarchi:
Also, SITL supports max 16 channels

DonLakeFlyer:
K, thanks. I'll have to look at why the code seems to think it needs to understand a max channels limit. The radio cal code is ancient history so it could just be stupid.

AndKe:
Current AP/AC show parameters for RC1-16 , log to dataflash RCIN,C1-14 , while some RCProtocol libraries like SBUS,  ST24, Spectrum,  contain 1-18 , but apparently go nowhere.. so yes,  there is some mess, but you can safely stop at 16.
Also, to make sure, for compatibility issues, you may check for the presence of a parameter name like RC16 before displaying ch16 as an option.

DonLakeFlyer:
So it's `RC_CHANNELS` and `RC_CHANNELS_RAW` are fighting with each other. I think way (way) back when it was something like ArduPilot only supported one of them. I think I can get rid of support for RC_CHANNELS_RAW completely now.

DonLakeFlyer:
Can you try a stable build from tomorrow to verify this is working now?

DonLakeFlyer:
Should be another daily build in about an hour

AndKe:
Current HEAD build works as expected.
Thank you, - you can close this case, and maybe look at the Disco ADB connection :)

lucasdemarchi:
Thanks!

PR #8889, opened 2020-07-07
Remove RC_CHANNELS_RAW support

DonLakeFlyer:
* Remove handling of `RC_CHANNELS_RAW` relying o n`RC_CHANNELS` only. Handling both was causing incorrect channel counts to be sent to Radio Cal. This fixes *#8853 
* Fixed flash find board loop

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -732,9 +732,6 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
     case MAVLINK_MSG_ID_RC_CHANNELS:
         _handleRCChannels(message);
         break;
-    case MAVLINK_MSG_ID_RC_CHANNELS_RAW:
-        _handleRCChannelsRaw(message);
-        break;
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
The RC_CHANNELS change should go to Stable as well

AndKe:
When compiled, RC15 and RC16 (in radio settings) is just in the middle. (1500us)  - while the RC_CHANNELS toggle between 1000 and 2000. 
This was verified by connecting AP2 and checking "mavlink inspector" widget.
Visualization of the channels 1..14 seems normal.

DonLakeFlyer:
Can you capture a telemetry log while you are wiggling 15 and 16? You'll have to turn on take log while not armed in settings.

> when did compilation output directory change from build/release to build/staging?

Recently to work around a Qt bug on OSX.

AndKe:
@DonLakeFlyer  
[RC15_16_wiggle_wiggle_yeah!.zip](https://github.com/mavlink/qgroundcontrol/files/4891729/RC15_16_wiggle_wiggle_yeah.zip)

PR #8893, opened 2020-07-08
Fix ROI, Orbit, Goto handling

DonLakeFlyer:
Related to #8838


DonLakeFlyer:
This needs to go to stable as well

PR #8902, opened 2020-07-09
Support map zoom to level 23

DonLakeFlyer:
* Discovered that Qt will zoom beyond the level for which it has tiles
* I also discovered that when Bing doesn't have a tile available instead of and error out it sends an image of an empty tile graphic
* It requires a bit of insanity with respect to detecting Bing sending empty tiles. But it seems to work. I detect the empty tile graphic and generic an error internal to QGC instead.
* In order to make this work you need to get the Bing "no tile" tiles out of the cache. Which means you need to delete your map cache file completely.
* In another pull I'm going to try to find the bogus tiles in the cache and delete them automatically.
* Related to #8690


src/QtLocationPlugin/QGeoMapReplyQGC.cpp
```diff
@@ -135,11 +142,19 @@ QGeoTiledMapReplyQGC::networkReplyFinished()
         }
         emit terrainDone(a, QNetworkReply::NoError);
     } else {
-        //-- This is a map tile. Process and cache it if valid.
-        setMapImageData(a);
-        if(!format.isEmpty()) {
-            setMapImageFormat(format);
-            getQGCMapEngine()->cacheTile(getQGCMapEngine()->urlFactory()->getTypeFromId(tileSpec().mapId()), tileSpec().x(), tileSpec().y(), tileSpec().zoom(), a, format);
+        if (a == _bingNoTileImage) {
```
dogmaphobic, line 145: This is an expensive comparison. Maps used to have an ID. Not sure how the new code handles it. The id is in `tileSpec().mapId()`. If simple, I would make sure it's Bing before doing a byte array comparison.

DonLakeFlyer: Cap, I meant to do that then forgot.

PR #8930, opened 2020-07-21
Fix joystick reconnect bug - srr-qgc branch

cwteal:
When a joystick would be disconnected then reconnected none of the
settings would be set correctly.  Noticing this did not occur on the
upstream I tracked down the introduction of this bug to the commit
8cd3c085f81ccc20289ee0e46427bdc309acf1f5.  Upon further inspection I
found that we should be calling _loadSettings and the other
initialization methods in the constructor to account for the case when a
joystick is unplugged and reconnected without the connection to the
activeVechicle changing.

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -265,6 +265,7 @@ Joystick::Joystick(const QString& name, int axisCount, int buttonCount, int hatC
         _buttonActionArray.append(nullptr);
     }
 
+    _activeVehicleChanged(_multiVehicleManager->activeVehicle());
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Closing, abandoned

cwteal:
@DonLakeFlyer - I'm not sure I understand the process.  The fix is valid.  Maybe I was supposed to do something different with the PR?  I didn't "abandon" this PR but it seems that the maintainers weren't interested?

DonLakeFlyer:
> but it seems that the maintainers weren't interested?

YOu'd have to ask the Auterion folks who manage this branch. I know nothing about it.

cwteal:
@DonLakeFlyer - thanks for your response.  I guess they aren't interested.

Best regards!

PR #8935, opened 2020-07-26
Cleanup duplicated magic numbers in joystick code

DonLakeFlyer:


src/Joystick/Joystick.cc
```diff
@@ -470,7 +477,7 @@ void Joystick::run()
         _update();
         _handleButtons();
         _handleAxis();
-        QGC::SLEEP::msleep(20);
+        QGC::SLEEP::msleep(qMin(static_cast<int>(1000.0f / _maxAxisFrequencyHz), static_cast<int>(1000.0f / _maxButtonFrequencyHz)) / 2);
```
DonLakeFlyer, line 480: This duplicates the previous value of 20 at current defaults. But adjusts if you futz with the max Hz values.

Issue #8934, opened 2020-07-24
Crash bug in UDPLink::_writeBytes()


dogmaphobic:
I've been chasing this all day and no definitive answer. It takes a while to get it to happen on my regular computer but it is quite easy to reproduce on Android. But as it depends on sending lots of data (Joystick), not easy to test unless you have an Android device with a joystick.

Just launch QGC, connect to anything (even Gazebo) and make sure you have a joystick and it is enabled. Once connected, you don't need to do anything. Just wait. On Android, it will crash within a minute. On desktops, it may take a while. It's a game of chance. No hard rule on when it happens (if you really feel like it, switching the Joystick code to send messages at 200Hz speeds up the crash interval).

The crash is here:
https://github.com/mavlink/qgroundcontrol/blob/master/src/comm/UDPLink.cc#L189

`target` is bogus. Not NULL but bogus.

The call comes from here:
https://github.com/mavlink/qgroundcontrol/blob/master/src/comm/UDPLink.cc#L182

When you hit the crash, the contents of `_sessionTargets` is valid. Note that `_sessionTargets` is only written to at the very beginning of the connection (literally when QGC receives the first packet from a vehicle). I don't fully grasp the nuances of the voodoo behind these c++11 constructs. The doc says the above gets a copy of the pointer. It looks like instead of:
```
for(UDPCLient* target: _sessionTargets) {
```
It should be:
```
for(auto& target: _sessionTargets) {
```
What I can tell is if I switch to the explicit:
```
for(int i = 0; i < _sessionTargets.count(); i++) {
    UDPCLient* target = _sessionTargets[i];
```
the crash goes away.

Before pushing a PR, I would like to hear more about those who actually eat c++11 for breakfast. These sort of things seem to be all over the place in the code base.

DonLakeFlyer:
Seems to me like your change is just rearranging the deck chairs on a possible stack corruption or something. Code one way and the variable is on the stack, code another and the variable is in a register maybe. Or the addition of `int i` moved the stack around as well. Linux debug should tell you if stack is walked over. OSX/Xcode you can build/run with the memory check thingy which was how I figured out this one: https://github.com/mavlink/qgroundcontrol/pull/8885/files

DonLakeFlyer:
Crash is caused by the hidden iterators which operate behind the scenes when using C++ range_for. Both read and write bytes can come in on different thread and try to use iterators on the same object at the same time and then boom.

[...]

PR #8936, opened 2020-07-26
Udp link mutex

DonLakeFlyer:
* Add mutex on _sessionTargets access
* Remove all C++ range_for usage
* General coding style cleanup
* Fix for #8934

### DIFF
--- a/src/comm/UDPLink.cc
+++ b/src/comm/UDPLink.cc
@@ -7,14 +7,6 @@
  *
  ****************************************************************************/
 
-
-/**
- * @file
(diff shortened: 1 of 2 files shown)


Antiheavy:
Is this fixing a very specific case or is it fixing a bug that could impact any UDP connection? Curious if we should update our application for this since we use UDP.

DonLakeFlyer:
In general anything that meets the bar for Stable should be picked up by custom builds which are based on that Stable. This crash can happen to anyone. It is more likely to happen with joystick users since it generates more gcs->vehicle traffic so more change of read/write overlap.

PR #8943, opened 2020-07-28
Prefer ALTITUDE message over GPS messages

DonLakeFlyer:


DonLakeFlyer:
@dagar FYI: In response the GPS messages with no GPS discussion

Issue #8923, opened 2020-07-17
Plan - VTOL: Mission command editors don't work correctly


DonLakeFlyer:
Currently the command editors for VTOL come from the generic command tree so they end up with all sorts of crap in there that is not needed. They should instead be contextual to the current VTOL state (MR/FW) at that point in the mission. So if the VTOL is in MR mode then it should show the MR editors. If it is in FW mode it should show FW editors.

DonLakeFlyer:
I also discovered while working on this that Mission time calculation was screwed up with respect to speed/vtol state changes.

DonLakeFlyer:
FYI: I'm using this issue to track all the big ticket work items to fix screwed up support in VTOL. With #8954 I'm almost done. The last big thing is to show VTOL commands as being available on when the vehicle is a VTOL. Current code always shows all VTOL commands as being available from the command picker no matter what type of vehicle.

DonLakeFlyer:
All the really big stuff is done now.

PR #8947, opened 2020-07-29
Plan: Fix recalc bugs associated with vtol state, speed, vehicle/gimbal yaw

DonLakeFlyer:
* Found/fixed a pile of bugs related to MissionController::_recalcMissionFlightStatus
* Gimbal/Vehicle yaw calc/display was buggy
* Current plan flight speed was being applied one mission item too early. Throwing off things like time calcs
* Current VTOL state was being applied on mission item too early
* Related to #8923


DonLakeFlyer:
Probably should have mentioned I created unit test to verify these fixes and keep them from coming back,

PR #8954, opened 2020-07-30
Plan: Show correct item editor based on current VTOL state in mission

DonLakeFlyer:
* Related to #8923

Example Plan with transitions:
* Plan starts with VTOL takeoff so QGC knows current state after that is FW
* Here you can see the Loiter to Alt editor which is showing fixed wing values
Screen Shot 2020-07-30 at 12 58 31 PM
* After that there is a transition to MR
Screen Shot 2020-07-30 at 12 58 39 PM
* And here you can see Loiter to Alt after that showing MR values
Screen Shot 2020-07-30 at 12 58 46 PM


dagar:
FYI @sfuhrer @RomanBapst

PR #8955, opened 2020-07-31
Fix zlib setup on Windows

lukegluke:


DonLakeFlyer:
Carp, I had seen that reported before but that it was correct. THanks

PR #8967, opened 2020-08-06
View change done using Drawer

DonLakeFlyer:
* Clicking the current tool button in the toolbar opens a drawer on the left edge to change view
* Easter egg for advanced mode is now click on version text
* Easter egg for show touch area is now shift-click on version text

Screen Shot 2020-08-06 at 2 09 07 PM


Antiheavy:
Comment 1:


> Easter egg for advanced mode is now click on version text

this seems too easy to accidentally press, especially on a mobile device.  maybe it could be a triple tap/click or something?

DonLakeFlyer:
> this seems too easy to accidentally press

Makes sense, I'll switch to double click.

Antiheavy:
Comment 2: (this is a big one)


> Clicking the current tool button in the toolbar opens a drawer on the left edge to change view

This UI change seems like an opportunity.  QGC has always had a problem with the top toolbar buttons giving equal prominence to Fly, Plan, Analyze, Vehicle Setup, Application Settings.  Fly and Plan should be given different prominence IMHO.

For most QGC users, Fly and Plan are used every flight.  They are mission action and execution oriented.  Some time ago there was a version of QGC Master that had the Fly and Plan view toggle buttons at the top of the Left Toolbar.  The top of the Left Toolbar is a great place for the Fly/Plan view toggle as it keeps the user focused on the map and on the Left Toolbar which is the primary way to interact with the vehicle.  All the stuff a user does every time is kept in once place and closer together in context.  A single click toggle button without having to move the mouse/finger would be a step forward in ease of use.  The button to change what the Left Toolbar looks like and does is right there at the top of the toolbar; it just makes sense.  This small change is a big step toward making things more intuitive.  It's also a big step toward unifying the Fly and Plan screens (at least mentally).

The new drawer now requires double the number of move-click actions to switch views.  This is a step backwards in ease of use for the very common task of switching between Fly and Plan views.

Right now, almost all other Top Toolbar icons work as toggles which display and hide their popup item when pressed.  This is easy and intuitive.  Maybe the Application Settings, Vehicle Settings, and Analyze tool could also be toggle buttons?  Or, if the goal is to consolidate all the menus under one button could we let the drawer have it's own multi-level navigation inside of it and let the whole thing toggle on/off with the single Top Toolbar button?

DonLakeFlyer:
All good comments. Let me think about it some.

PR #8970, opened 2020-08-07
Allow custom AndroidManifest.xml and icons

remster:
The main problem this PR solves is for custom QGC to have its own Android identity and thus allows QGC and custom QGC run on the same Android device side by side.


custom-example/android/AndroidManifest.xml
```diff
@@ -0,0 +1,84 @@
+<?xml version="1.0"?>
+<manifest package="org.custom.qgroundcontrol" xmlns:android="http://schemas.android.com/apk/res/android" android:versionName="-- %%INSERT_VERSION_NAME%% --" android:versionCode="-- %%INSERT_VERSION_CODE%% --" android:installLocation="auto">
```
remster, line 2: Please note `INSERT_VERSION_CODE` and `INSERT_VERSION_NAME` that [allegedly](http://www.cutehacks.com/blog/2015/5/28/one-version-number-to-rule-them-all) appeared in Qt5.2  (I tested them to be there on Qt.5.12.5, while QGC is now on Qt.5.12.6) 

They work against:
```
+ANDROID_VERSION_NAME="1.2.3"
+ANDROID_VERSION_CODE=123
```
Which I am not setting here, but could write some script that automatically counts the commits for the `..code` and replicates `$${CUSTOM_QGC_VERSION}` for the `..name`.

custom-example/android/AndroidManifest.xml
```diff
@@ -0,0 +1,84 @@
+<?xml version="1.0"?>
```
remster, line 1: I did try generating the `package` from `QGC_ORG_NAME`, but it aint all that easy and it need to agree with the folder structure where the `QGCActivity.java` lives. Since custom QGCs may have different permissions, all things considered, I think to replicate the entire AndroidManifest.xml is more honest.

android.pri
```diff
@@ -1,9 +1,26 @@
...
 QT += androidextras
+
 ANDROID_PACKAGE_SOURCE_DIR = $$PWD/android
+exists($$PWD/custom/android) {
+    message("Merging $$PWD/custom/android/ -> $$PWD/android/")
+    ANDROID_PACKAGE_SOURCE_DIR = $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR
+
+    system($$QMAKE_MKDIR $$ANDROID_PACKAGE_SOURCE_DIR)
+    system($$QMAKE_COPY_DIR $$PWD/android/* $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR)
+    system($$QMAKE_COPY_DIR $$PWD/custom/android/* $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR)
```
remster, line 12: `androiddeployqt` [only allows](https://doc.qt.io/qt-5/deployment-android.html#androiddeployqt) a single input folder so i had no better ideas but to merge these two folders in the build tree and give that to `androiddeployqt`

android.pri
```diff
@@ -1,9 +1,26 @@
...
+}
+
+exists($$PWD/custom/android/AndroidManifest.xml) {
+    OTHER_FILES += \
+    $$PWD/custom/android/AndroidManifest.xml
```
remster, line 17: This allows editing whichever AndroidManifest.xml in Qt Creator 
custom_manifest

DonLakeFlyer:
I've been staring at this for a couple days now. It just all seems so nasty to have full duplication of the manifest file. That will get out of sync quickly. To me it would be better to have something kind of like an updateqrc.py script which would take master as input and then futz with it somehow.

akontsevich:
> I've been staring at this for a couple days now. It just all seems so nasty to have full duplication of the manifest file. That will get out of sync quickly. To me it would be better to have something kind of like an updateqrc.py script which would take master as input and then futz with it somehow.

I complete disagree: there could be many custom situations in the manifest file:

- additional permissions
- package name, labels & icons
- components
- hardware
- etc

Which could differ, so you can't generate all the cases with script.

DonLakeFlyer:
There are two problems to solve here:

1) Make simple customizations of app name, org, icon and so forth work for the android custom build. I believe this is the way more than majority use case. Duplicating the manifest file in such a way that it can get out of sync with master is totally not the way to solve this problem. I had thought that Qt had solved this problem with replacement macros. If 5.12 doesn't have everything needed we will be moving to 5.15 sometime soon. I would assume they have solved the problem by 5.15?

2) Make advanced customization allowable including major changes to the manifest. This will be used rarely. If this is what you want then in reality you are better off modifying the manifest file directly in mainline code in your fork. That way at least merges from upstream still make it into your custom version of the manifest. If there is a completely separate manifest file for custom then you will constantly have to manually keep the part you want in sync by hand. And it require all this file copying craziness in the build system. That just doesn't make sense to me.

akontsevich:
> I've been staring at this for a couple days now. It just all seems so nasty to have full duplication of the manifest file. That will get out of sync quickly. To me it would be better to have something kind of like an updateqrc.py script which would take master as input and then futz with it somehow.

I complete disagree: there could be many custom situations in the manifest file:

- additional permissions
- package name, labels & icons
- components
- hardware
- etc

Which could differ, so you can't generate all the cases with script.

> There are two problems to solve here:
> 
>     1. Make simple customizations of app name, org, icon and so forth work for the android custom build. I believe this is the way more than majority use case. Duplicating the manifest file in such a way that it can get out of sync with master is totally not the way to solve this problem. I had thought that Qt had solved this problem with replacement macros. If 5.12 doesn't have everything needed we will be moving to 5.15 sometime soon. I would assume they have solved the problem by 5.15?

Yep, 5.15 has something: https://doc.qt.io/qt-5/deployment-android.html#androidmanifest-xml Not sure however if it can cover all sections like: add/remove permissions records for ex.

akontsevich:
>  1. Make simple customizations of app name, org, icon and so forth work for the android custom build. I believe this is the way more than majority use case. Duplicating the manifest file in such a way that it can get out of sync with master is totally not the way to solve this problem. I had thought that Qt had solved this problem with replacement macros. If 5.12 doesn't have everything needed we will be moving to 5.15 sometime soon. I would assume they have solved the problem by 5.15?

Yep, 5.15 has something: https://doc.qt.io/qt-5/deployment-android.html#androidmanifest-xml Not sure however if it can cover all sections like: add/remove permissions records for ex.

akontsevich:
BTW, @DonLakeFlyer, custom java files solution is ok?

DonLakeFlyer:
So refreshing myself on how android build works:
* From looking at it I don't think there is a way to get a different application icon other than copying from custom into the android dir tree. Does that seem correct?
* If you look at .travis.yml and how it uses tools/update_android_version.sh you'll see how the build uses this to update the version and package name automatically
* So some combination of bringing update_android_version.sh into the main build and using androiddeployqt substituions should get us most of the way there for case 1 I describe above.
* And then you still have the directory tree copy semantics, but in general it is only needed for the app icon. This way you don't have the issue of having to duplicate files which can get out of sync with mainstream changes for the majority use case. The default custom build example should only use this level of icon setting ability as the example.
* Then since the directory tree copy stuff id there you can use it for more advanced things like full replacement of manifest or activity java.
* Given this I would make the directory tree copy no optional but part of the standard custom build. And update the custom example to use it with a custom app icon for android.

I'm not sure in detail if all of this will work. But I hope it's close enough from a direction standpoint.

remster:
@DonLakeFlyer 
> So some combination of bringing update_android_version.sh into the main build and using androiddeployqt substituions should get us most of the way there for case 1 I describe above.

I can land with a single AndroidManifest where we just replace the package name in the build tree, but also allow the custom build to supply an entire manifest.

> And then you still have the directory tree copy semantics, but in general it is **only needed for the app icon**. 
The problem I see is in the identity of the `org.mavlink.qgroundcontrol.QGCActivity` - nested [here](https://github.com/mavlink/qgroundcontrol/blob/master/android/src/org/mavlink/qgroundcontrol/QGCActivity.java): `qgroundconrol/android/src/org/mavlink/qgroundcontrol` "should always match your project's base package name where you keep your activities and other app code." ([see](https://developer.android.com/guide/topics/manifest/manifest-intro))

If a custom build comes with a different package name, it must come with a `CustomQGCActvity.java` nested in the directory stucture that reflects this package as shown [here](https://github.com/mavlink/qgroundcontrol/pull/8970/files#diff-689e02569220e8bf2db43d3169defd24R1).

DonLakeFlyer:
Hmm, didn't know that. Currently QGC drops a daily and stable version of QGC to the store. It only massages the manifest with the two different package names. It doesn't touch the java as far as I know. Yet it still works.

Is there no way with the java code to do macro replacement type stuff like c++. So the package name could come from somewhere else?

android.pri
```diff
@@ -1,9 +1,33 @@
...
 ANDROID_PACKAGE_SOURCE_DIR = $$PWD/android
+android_source_dir_target.target = android_source_dir
+PRE_TARGETDEPS += $$android_source_dir_target.target
+QMAKE_EXTRA_TARGETS += android_source_dir_target
+
+exists($$PWD/custom/android) {
+    message("Merging $$PWD/custom/android/ -> $$PWD/android/")
+    ANDROID_PACKAGE_SOURCE_DIR = $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR
+
+    android_source_dir_target.commands = $$QMAKE_MKDIR $$ANDROID_PACKAGE_SOURCE_DIR && \
```
remster, line 14: This allows, but doesn't force, the custom build to supply its own AndroidManifest.xml and java files.

android.pri
```diff
@@ -1,9 +1,33 @@
...
+
+    ANDROID_PACKAGE_SOURCE_DIR = $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR
+    android_source_dir_target.target = android_source_dir
+    PRE_TARGETDEPS += $$android_source_dir_target.target
+    QMAKE_EXTRA_TARGETS += android_source_dir_target
+
+    android_source_dir_target.commands = $$QMAKE_MKDIR $$ANDROID_PACKAGE_SOURCE_DIR && \
+            $$QMAKE_COPY_DIR $$PWD/android/* $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR && \
+            $$QMAKE_COPY_DIR $$PWD/custom/android/* $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR && \
+            $$QMAKE_STREAM_EDITOR -i \"s/package="org.mavlink.qgroundcontrol"/package=\"$$QGC_ORG_DOMAIN.qgroundcontrol\"/\" $$ANDROID_PACKAGE_SOURCE_DIR/AndroidManifest.xml
```
remster, line 18: This hi-jacks `QGC_ORG_DOMAIN` for the purpose of identifying the custom QGC app for android. Do you want me to define `QGC_ANDROID_PACKAGE` instead to be explicit?

remster:
@DonLakeFlyer Your comments addressed - and I am keeping things as minimal as i can. 
- allows, but doesn't force, the custom build to supply AndroidManifest.xml - if custom doesn't supply, the package name will be replaced to `$$QGC_ORG_DOMAIN.qgroundcontrol`. If the custom build does supply it, it will be as is.
- allows, but doesn't force, the custom build to supply its java classes. You were right - the activity nesting does not need to match the package. If:
 - the custom build doesn't supply its java classes and thus doesn't supply custom activity 
 - both custom build and vanilla QGC are installed on the device.
 - someone launches an intent towards `org.mavlink.qgroundcontrol.QGCActivity`, my android expert colleague tells me, Android will ask what specific app one wishes to launch.

android.pri
DonLakeFlyer: Yeah I think a `QGC_ANDROID_PACKAGE` would be better.

DonLakeFlyer:
@remster Also thanks for dealing with all the change requests here. I think we'll end up with a nice system which is simple to use in most case but can still extend to complex usage.

android.pri
remster: Done: https://github.com/mavlink/qgroundcontrol/pull/8970/commits/9adc28bd0f454502fa27f8af7fc441bbee318a81

android.pri
```diff
@@ -1,9 +1,33 @@
...
+
+    ANDROID_PACKAGE_SOURCE_DIR = $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR
+    android_source_dir_target.target = android_source_dir
+    PRE_TARGETDEPS += $$android_source_dir_target.target
+    QMAKE_EXTRA_TARGETS += android_source_dir_target
+
+    android_source_dir_target.commands = $$QMAKE_MKDIR $$ANDROID_PACKAGE_SOURCE_DIR && \
+            $$QMAKE_COPY_DIR $$PWD/android/* $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR && \
+            $$QMAKE_COPY_DIR $$PWD/custom/android/* $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR && \
+            $$QMAKE_STREAM_EDITOR -i \"s/package=\\\"org.mavlink.qgroundcontrol\\\"/package=\\\"$$QGC_ANDROID_PACKAGE\\\"/\" $$ANDROID_PACKAGE_SOURCE_DIR/AndroidManifest.xml
```
remster, line 18: This renders:
```
... sed -i "s/package=\"org.mavlink.qgroundcontrol\"/package=\"com.airmap.airboss\"/" /home/remek/Projects/airmap/airmap/build-qgroundcontrol-Android_for_arm64_v8a_Clang_Qt_5_12_8_for_Android_ARM64_v8a-Release/ANDROID_PACKAGE_SOURCE_DIR/AndroidManifest.xml
```
In turn the build folder:
```
[remek][~/.../airmap/airmap/qgroundcontrol][android-manifest]$ cat ../build-qgroundcontrol-Android_for_arm64_v8a_Clang_Qt_5_12_8_for_Android_ARM64_v8a-Release/ANDROID_PACKAGE_SOURCE_DIR/AndroidManifest.xml
<?xml version="1.0"?>
<manifest package="org.custom.qgroundcontrol"  ...
```
and similarly: 
```
cat ../build-qgroundcontrol-Android_for_arm64_v8a_Clang_Qt_5_12_8_for_Android_ARM64_v8a-Release/android-build/AndroidManifest.xml
<?xml version="1.0"?>
<manifest package=""org.custom.qgroundcontrol"
```

remster:
@DonLakeFlyer your comment addressed - at it seems i have caught a bug in the earlier commits (new escape sequences) - i have tested as much as a i could up to running the app on my phone

DonLakeFlyer:
So I tried testing this and found this:
* CustomPlugin.g includes a direct include GstVideoReceiver.h. This is not needed and makes it not compile if you don't have gstreamer installed. I know, not you, but could you fix in this pull?
* The pull is not up to date with master so it won't build. It needs a 'python updateqrc.py' done on it.
* I don't see changes to do this: "So some combination of bringing update_android_version.sh into the main build and using androiddeployqt substituions should get us most of the way there for case 1 I describe above." The version number should be set in the build system (if possible) instead of .travis.yml script.
* Also somewhere along the way the 32/64 apks are renamed to use real app names. Don't quite remember where that happens. But the end result of the build should be an appropriately named apk.
* After I fixed those two things I built and deployed:
  * In my build directory things were built into a directory literally named "ANDROID_PACKAGE_SOURCE_DIR" instead of the normal "android-build" directory. The android.pri file is missing some '$$'. 
  * The app was only sort of correctly showing the custom build name after installing. There are two versions of the name one with and without spaces. "Custom QGroundControl" and "CustomQGroundControl". The one with the space in it should be used in the places which are shown to the user like the app name in the app list on the tablet. Right now the app shows as "CustomQGroundControl"
  * The app did not have the custom build icon like it should. It just showed the regular QGC icon in the app list.

remster:
> So I tried testing this and found this:
> 
> * CustomPlugin.g includes a direct include GstVideoReceiver.h. This is not needed and makes it not compile if you don't have gstreamer installed. I know, not you, but could you fix in this pull?

done : https://github.com/mavlink/qgroundcontrol/pull/8970/commits/e969a644c064cfe4ec3ff32de474bb52067d59d6

> * The pull is not up to date with master so it won't build. It needs a 'python updateqrc.py' done on it.

done 

> * I don't see changes to do this: "So some combination of bringing update_android_version.sh into the main build and using androiddeployqt substituions should get us most of the way there for case 1 I describe above." The version number should be set in the build system (if possible) instead of .travis.yml script.

working on it
> * Also somewhere along the way the 32/64 apks are renamed to use real app names. Don't quite remember where that happens. But the end result of the build should be an appropriately named apk.

working on it
> * After I fixed those two things I built and deployed:
>   
>   * In my build directory things were built into a directory literally named "ANDROID_PACKAGE_SOURCE_DIR" instead of the normal "android-build" directory. The android.pri file is missing some '$$'.

It is deliberate to create a intermediate `./build/ANDROID_PACKAGE_SOURCE_DIR` that is not ./build/android-build`. The latter is created by `androiddeployqt` from the former. On master androiddeployqt` works on the source tree `./android` on this PR is works on the build tree `./build/ANDROID_PACKAGE_SOURCE_DIR` containing merged ./android` and `./custom/android`

>   * The app was only sort of correctly showing the custom build name after installing. There are two versions of the name one with and without spaces. "Custom QGroundControl" and "CustomQGroundControl". The one with the space in it should be used in the places which are shown to the user like the app name in the app list on the tablet. Right now the app shows as "CustomQGroundControl"

working on it
>   * The app did not have the custom build icon like it should. It just showed the regular QGC icon in the app list.

What is checked into this branch are QGC icons, so it will be indistinguishable. I will tamper with this icon to make it obvious.

DonLakeFlyer:
@remster Why don't you wrap up the stuff you are confident with and remove the stuff you are not. And then I can merge that. Then for the stuff you pulled out that are questionable submit them in a separate pull after this and I can look at working on them further.

remster:
> @remster Why don't you wrap up the stuff you are confident with and remove the stuff you are not. And then I can merge that. Then for the stuff you pulled out that are questionable submit them in a separate pull after this and I can look at working on them further.

@DonLakeFlyer Done!

PR #8972, opened 2020-08-07
Fly: Move flight mode, vtol mode, armed into main status

DonLakeFlyer:
VTOL example:
Screen Shot 2020-08-07 at 10 45 38 AM

Arm/Disarm/Emergency stop are in main status dropdown:
Screen Shot 2020-08-07 at 10 51 08 AM

### DIFF
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -320,9 +320,6 @@ const QVariantList& FirmwarePlugin::modeIndicators(const Vehicle*)
     if(_modeIndicatorList.size() == 0) {
         _modeIndicatorList = QVariantList({
             QVariant::fromValue(QUrl::fromUserInput("qrc:/toolbar/ROIIndicator.qml")),
-            QVariant::fromValue(QUrl::fromUserInput("qrc:/toolbar/ArmedIndicator.qml")),
-            QVariant::fromValue(QUrl::fromUserInput("qrc:/toolbar/ModeIndicator.qml")),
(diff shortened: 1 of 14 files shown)


Antiheavy:
I see there are multiple PR's related to these UI improvements, but I'll leave some comments here for discussion:

- Overall this is awesome.

- hiding the manual Arm under the vehicle status seems like a good place.  Will Emergency Stop be hidden under there too (i'm sure there will be debate about if Emergency Stop should be that hidden, but I think it's okay)?

- The Sensor Status drop down is great.  Is it worth considering expanding this to include other things such as Mission status/Feasibility and potentiality other system checks?  Could this drop down accommodate custom plugin items?

- At the moment the top toolbar is too cluttered and the item spacing is inconsistent and overall confusing.  Things just need to be spaced out more.  In my mind there are currently 3 logical functions in the top toolbar which might help to separate by white space:  
-- Application and Vehicle Settings menus
-- Vehicle status and mode
-- Vehicle telemetry

DonLakeFlyer:
> * Will Emergency Stop be hidden under there too (i'm sure there will be debate about if Emergency Stop should be that hidden, but I think it's okay)?

Disarm is in there if the vehicle is armed and it will emergency stop if the vehicle is flying which is how it worked before.

> * Is it worth considering expanding this to include other things such as Mission status/Feasibility and potentiality other system checks?

Good ideas, but no way in mavlink to get that information.

> * Could this drop down accommodate custom plugin items?

What would be better is to expand mavlink to handle additional status and then route through that.

> * At the moment the top toolbar is too cluttered and the item spacing is inconsistent and overall confusing.

Yup, still all in flux as I put more work into it. Most likely going to move instrument values out of there (I don't think the two places for values is working) and put them in a new values display which is more like the one in the custom build example.

PR #8976, opened 2020-08-09
Fly: Telemetry value bar

DonLakeFlyer:
The goal of the work is to allow you to be able to see your telemetry value and work with the camera at the same time.

* Add new customizable telemetry value bar at center bottom
* Each vehicle class has it's own defaults for the bar and settings space
* Guide action confirm rework
* Remove values from toolbar
* Remove values from instrument panel
* Update custom example to latest mainline code

Remaining work:
* Move Vibration page to Analyze
* Rework camera/video stream right panel ui

Multi-Rotor example
Screen Shot 2020-08-09 at 10 13 06 AM

Fixed Wing default values:
[ Altitude, Climb Rate, Air Speed, Flight Time]
[ Distance From Home, GroundSpeed, Throttle Pct, Total Flight Distance ]
Note that icons/labels are WIP...
Screen Shot 2020-08-09 at 10 11 36 AM


DonLakeFlyer:
Same as before custom builds can specify the defaults for their vehicles. Also more importantly they can set up fancier ranges for coloring and things like that. So for example you could color airspeed value red/yellow based on value ranges which would let you know you are getting into trouble with respect to flying too slowly.

Antiheavy:
Neat. Will the action sliders sit above this new display?

DonLakeFlyer:
Screen Shot 2020-08-09 at 10 13 48 AM

Antiheavy:
The dynamic resizing of the Telemetry Value bar is a little weird, it twitches from wider to narrower depending on how many digits are in each telemetry field:
image

DonLakeFlyer:
> it twitches from wider to narrower depending on how many digits are in each telemetry field:

Yup. I haven't figured out how to fix that yet.

PR #8983, opened 2020-08-13
Add Vehicle FactGroup for ESC_STATUS values

PonomarevDA:
I added `esc_status` to Vehicle FactGroup and tested it using own uavcan esc and PX4 devices with implemented mavlink `esc_status` message from this PR: https://github.com/PX4/Firmware/pull/14408.
So, it's working, but you need to use new version of [c mavlink library](https://github.com/mavlink/c_library_v2) (current mavlink version from QGC doesn't have `esc_status`, it appeared in https://github.com/mavlink/mavlink/pull/1322), for example I'm using [cdcea1c6 version](https://github.com/mavlink/c_library_v2/tree/cdcea1c68565e435d18f88c12a4d5db20b070df1).
When qgroundcontrol update [c mavlink library](https://github.com/mavlink/c_library_v2) version, it will be able to use this PR.

In screen below you can see implemented feature: it shows rpm from my uavcan ESC (it doesn't have current and voltage sensors, so they are displayed as zero).

image


DonLakeFlyer:
The standard for shortDesc is sentence like capitalization. Can you capitalize the first words. Also to update to lastest mavlink just go into the mavlink submodule dir (`libs/mavlink/include/mavlink/v2.0`) in your repo and do a `git pull origin master` and include that in your pull.

src/Vehicle/EscStatusFactGroup.json
```diff
@@ -0,0 +1,98 @@
+{
+    "version":  1,
+    "fileType": "FactMetaData",
+    "QGC.MetaData.Facts":
+[
+{
+    "name":             "index",
+    "shortDesc":        "Index of the first ESC in this message",
+    "type":             "uint8_t",
+    "decimalPlaces":    2,
```
DonLakeFlyer, line 10: Remove decimalPlaces for ints

DonLakeFlyer:
Also I assume these values have units like V Voltage, maybe current is maH? For ones that make sense you should specifiy units. Look at FactMetaData.cc for units which support conversion and use those if possible. Or look at other json files for examples of things like voltage and current.

DonLakeFlyer:
This is failing all builds

src/Vehicle/EscStatusFactGroup.json
PonomarevDA: removed

PonomarevDA:
I fixed all your notes.
It didn't build because of old mavlink version. Now it's almost ok, but problem with local was founded. I should specify path to json file somewhere. I'll do it.

PonomarevDA:
@DonLakeFlyer, I understood that I should use Qrowdin to update the translation files.
I tried to use [qgc-lupdate.sh script](https://github.com/mavlink/qgroundcontrol/blob/master/translations/qgc-lupdate.sh). At first time it failed because one of files have non UTF-8 encoding (json works only with UTF-8). [This commit](https://github.com/mavlink/qgroundcontrol/commit/7f593de24c7a53619193d6c74322dc905c03463f#diff-0b18929be7a2738625690364674c3752) changed it for src/comm/USBBoardInfo.json from UTF-8 to UTF-8 with BOM. After returning the encoding script works well and I see that it uses my new src/Vehicle/EscStatusFactGroup.json file.
So, should I push fix in src/comm/USBBoardInfo.json and/or modified translation/*.ts files in this PR?
I think anyway I should wait for translation updating.

DonLakeFlyer:
What OS and OS Language do you have your machine set to?

PonomarevDA:
> What OS and OS Language do you have your machine set to?

Ubuntu 18.04, English (US)

DonLakeFlyer:
Ok. Don't worry about translations or the USBBoardInfo thing. I'll deal with that. Also I just realize the I screwed you up with the latest changes I made to Vehicle FactGroups. I moved all Vehicle FactGroups to there own source file and they now handle the mavlink message parsing in there own code instead of Vehicle. Sorry about that. You'll have to update your pull to the new model. Take a look here at this for an example of how to set it up. It should be a fairly simple change.

* https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/VehicleEstimatorStatusFactGroup.h
* https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/VehicleEstimatorStatusFactGroup.cc

PonomarevDA:
I made this changes. It builds and seems to be working fine (except translation problem).
I should also notice about mavlink version. I choose [2fa36812 version](https://github.com/mavlink/c_library_v2/tree/2fa368124a7803a3f968b386406804d217f010a5) because in [the next commit (d87878bc)](https://github.com/mavlink/c_library_v2/commit/d87878bce48d49fc3ab7305b7ffc907590f3468e#diff-2f2230fb909cac5a125331d9e3ee7ce0R7191) new argument appear in `mavlink_msg_battery_status_pack()` methods, but current version of QGC call these methods without new argument, so now it can't be build on mavlink d87878bc or higher. In other words, I choose the maximum working version.
I need to do a test with my ESC device to make sure that the PR is working fine. So, I will write you when I do it.

DonLakeFlyer:
This is failing CI because the json file is not added to resources

PonomarevDA:
So, I fixed it. Also I tested it with my ESC device yesterday. It was good.

DonLakeFlyer:
All good, thanks

PR #8985, opened 2020-08-14
item color added for listview, combobox

bijungsm:
comboboxes listviews have items. Items are displayed in button color. However, the item is not a button. Not beautiful So I added the item color. It is necessary to separate the color of the item to make it beautiful. The colors and qml files will need to be changed later.


DonLakeFlyer:
I would rather see a full pull which has palette changes as well as control code changes so I can see what this will look like. Just adding stuff to the palette which is not yet used is not super helpful.

DonLakeFlyer:
Closing this until there is something useful here

PR #8988, opened 2020-08-14
Mavlink Inspector: Support displaying messages from multiple systems

DonLakeFlyer:
* Here is an example of a log replay. You can see the system combo shows up with two entries. 255 is QGC outgoing telemetry, and 1 is the incoming vehicle telemetry.
Screen Shot 2020-08-14 at 3 10 57 PM
* Also fixed a problem where values for a message were not updated unless the message was selected in the ui. This made analyzing logs not super useful. Since you couldn't see the last value for the message. And in fact if you selected a message, new values came in for that message, selected some other message, new values came in on the previous message and then you went back to select the previous message it would show you stale data with no indicator of that.

### DIFF
--- a/src/AnalyzeView/MAVLinkInspectorController.cc
+++ b/src/AnalyzeView/MAVLinkInspectorController.cc
@@ -179,6 +179,7 @@ QGCMAVLinkMessage::QGCMAVLinkMessage(QObject *parent, mavlink_message_t* message
         QGCMAVLinkMessageField* f = new QGCMAVLinkMessageField(this, msgInfo->fields[i].name, type);
         _fields.append(f);
     }
+    update(message);
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


src/AnalyzeView/MAVLinkInspectorController.cc
```diff
@@ -222,10 +223,7 @@ void
 QGCMAVLinkMessage::update(mavlink_message_t* message)
 {
     _count++;
-    //-- If we are not consuming this message, no need to parse it
-    if(!_selected && !_fieldSelected) {
-        return;
-    }
```
DonLakeFlyer, line 228: @dogmaphobic I assume you put this in for performance. But it makes the inspector either not shows the user what they want to see, or in other cases it shows wrong data. Was there another reason other than perf for this?

dogmaphobic: Yes, it's hundreds of thousands signal/slot calls each.

dogmaphobic: (per second)

DonLakeFlyer: Fixed https://github.com/mavlink/qgroundcontrol/pull/8991

PR #8989, opened 2020-08-15
Set default landing heading to takeoff heading

mattstraehl:
For **VTOL** and **FW** set default landing heading to takeoff heading when `TakeoffMissionItem` is added in `PlanView`.

VTOL:

Screenshot from 2020-08-14 20-50-07

FW:

Screenshot from 2020-08-14 20-48-03

@DonLakeFlyer, please take a look and let me know if anything is missing.

### DIFF
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -166,6 +166,7 @@ void MissionController::_newMissionItemsAvailableFromVehicle(bool removeAllReque
         _visualItems->deleteLater();
         _visualItems  = nullptr;
         _settingsItem = nullptr;
+        _takeoffMissionItem = nullptr;
         _updateContainsItems(); // This will clear containsItems which will be set again below. This will re-pop Start Mission confirmation.
 
         QmlObjectListModel* newControllerMissionItems = new QmlObjectListModel(this);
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
Thanks so much @mattstraehl

src/MissionManager/MissionController.cc
```diff
@@ -360,6 +360,7 @@ VisualMissionItem* MissionController::insertTakeoffItem(QGeoCoordinate /*coordin
 {
     int sequenceNumber = _nextSequenceNumber();
     TakeoffMissionItem * newItem = new TakeoffMissionItem(_controllerVehicle->vtol() ? MAV_CMD_NAV_VTOL_TAKEOFF : MAV_CMD_NAV_TAKEOFF, _masterController, _flyView, _settingsItem, this);
+    _takeoffMissionItem = newItem;
```
DonLakeFlyer, line 363: So it gets set here. But I don't see any code which clears it. The cases where this item goes away needs to be handled and null out the ptr. Also a takeoff item can come in through a plan from the vehicle or one that is loaded from a file. Those cases need to set this as well.

mattstraehl: @DonLakeFlyer, thanks so much for pointing this out, totally missed it. I just updated the PR, please let me know in case anything else is missing.

DonLakeFlyer:
Tested and all look goods. Thanks for bringing this down

PR #8990, opened 2020-08-15
localization: Organize the same line elements.

muramura:
I organized the same line elements.

### DIFF
--- a/localization/qgc_ko.ts
+++ b/localization/qgc_ko.ts
@@ -5,24 +5,20 @@
     <name>APMAirframeComponent</name>
     <message>
       <location filename="../src/AutoPilotPlugins/APM/APMAirframeComponent.qml" line="74"/>
-      <location filename="../src/AutoPilotPlugins/APM/APMAirframeComponent.qml" line="74"/>
       <source>Airframe is currently not set.</source>
       <translation>기체유형이 설정되지 않음</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Translation file are output by crowdin. So as far as I understand modifying the manually won't have an affect. Since the next push down from Crowdin to QGC will overwrite it (and possible cause merge conflicts).

hamishwillee:
Correct. 
The extra line is in the **source** that crowdin is getting - so this is a .ts file generation issue. @dogmaphobic do you have any power over what is generated?

Gus, crowdin shows duplicate files - are they both valid? If not, which should we delete? 
image

DonLakeFlyer:
> The extra line is in the **source** that crowdin is getting - so this is a .ts file generation issue.

I don't get where this "extra line" is? If I look at the diff it looks like the changes are maybe only in spaces or tabs or something? There is no change in the non-whitespace stuff.

Screen Shot 2020-08-24 at 8 41 46 AM

hamishwillee:
@DonLakeFlyer The diff shows the file after the duplicate line has been removed. So the source has line with tag `location` twice.

image

DonLakeFlyer:
Duh!

DonLakeFlyer:
Which all still all comes back to not understanding why this pull is needed? @muramura? Are these extra lines causing some sort of problem?

DonLakeFlyer:
Closing unless I get more info

muramura:
It works fine even if there are multiple same lines.

PR #9001, opened 2020-08-18
Ensure showGimbalOnlyWhenSet as custom builds may have different defaults

remster:
This test has been failing on our custom QGC as we have a different default for `showGimbalOnlyWhenSet`


DonLakeFlyer:
This is fine. But what would be better would be to include a test to make sure this works correctly in both cases. Then it won't get broken.

PR #9018, opened 2020-08-22
Add support for custom Mapbox style

shatyuka:
Similar to this #7276 PR.

I need to add some special marks on the satellite map, like takeoff and land zone. But currently QGC does not support modify the offline map atlas, so I turn to Mapbox custom style service. However QGC only support presets provided by Mapbox.

I add a map type named "Custom", and user can provide their Mapbox user name and style ID in "Offline Maps" options below "Mapbox Access Token".

Also because I add a new map type, it won't break or interfere with original functions.

Mapbox style can be designed [here](https://www.mapbox.com/studio/)

Screenshots:
Screenshot_20200822_195806
Screenshot_20200822_195932


DonLakeFlyer:
Do you know if the way mapbox accesses tiles from a url is the same as the goole url format? That is somewhat of a standard.

shatyuka:
Not exactly same, but it doesn't matter. We use different provider classes to deal with it. See [here](https://github.com/mavlink/qgroundcontrol/tree/master/src/QtLocationPlugin)
Also, the original Mapbox support is functional. I just implement the custom style api for mapbox.

DonLakeFlyer:
So I'm trying to figure out how to test this. Starting with something simple. Just getting regular mapbox support to work. I created a mapbox account. Got the default public token put that into maps settings. Picked a mapbox map and it doesn't work. Is there something else I need to do?

DonLakeFlyer:
Do you really need MapBox user name? Can't you create your own private token which exposes the styles? That at least seems possible from what I can see from looking at mapbox

DonLakeFlyer:
From some more digging into how mapbox works this looks good, thanks.

PR #9019, opened 2020-08-22
Full support for BATTERY_STATUS mavlink message

DonLakeFlyer:
Previously:
* QGC used a mix of `SYS_STATUS` and a few of the `BATTERY_STATUS` fields for single battery status to the user
* Toolbar showed battery status for a single battery and only `SYS_STATUS` level of battery information influence the indicator
* Only the first two batteries where exposed as a Vehicle Battery Fact group
* Reporting of spoken battery levels warnings and coloring of toolbar battery information was based on levels hard-coded into the QGC code

Now:
* QGC only pulls battery information from `BATTERY_STATUS`, it **DOES NOT** use any battery information from `SYS_STATUS`
* An essentially infinite number of batteries are available through the Vehicle Battery Fact groups. And all of the `BATTERY_STATUS` information for all batteries is available for display through the Telemetry Values Bar.
* The toolbar will now shows icons for as many batteries as are reported by the Vehicle
* Each battery shows an Icon as well as a piece of text. The text shown varies based on what is available in this order:
   * `BATTERY_STATUS.battery_remaining`
   * `BATTERY_STATUS.voltages` totalled up value
   * `BATTERY_STATUS.charge_state'
   * Icon only
* The battery icon and text will change to a palette color based on `BATTERY_STATUS.charge_state' value:
  * text : `MAV_BATTERY_CHARGE_STATE_UNDEFINED` 
  * text : `MAV_BATTERY_CHARGE_STATE_OK`
  * colorOrange : `MAV_BATTERY_CHARGE_STATE_LOW`
  * colorRed:  `MAV_BATTERY_CHARGE_STATE_CRITICAL`, `MAV_BATTERY_CHARGE_STATE_EMERGENCY`, `MAV_BATTERY_CHARGE_STATE_FAILED`, `MAV_BATTERY_CHARGE_STATE_UNHEALTHY`
  * text: Anything else
* QGC only supports spoken battery level warnings if `BATTERY_STATUS.charge_state' is supported.
  * It will report a transition to a higher level of warning a single time. With the current warning level starting out as `MAV_BATTERY_CHARGE_STATE_OK`.
  * If the charge state then changes to say `MAV_BATTERY_CHARGE_STATE_LOW` it will announce "battery low".
  * If there is more than one battery on the vehicle it will include the battery id in the announcement
  * If the last warning charge state was say `MAV_BATTERY_CHARGE_STATE_CRITICAL` and then it lowers to `MAV_BATTERY_CHARGE_STATE_LOW` and later one goes back up to `MAV_BATTERY_CHARGE_STATE_CRITICAL` that will not generate any announcements since it had already announced critical
  * If you land disarm and rearm the warnings already spoken will be reset so you get them again

Here you can see two batteries on the toolbar. The one on the left is in the `MAV_BATTERY_CHARGE_STATE_LOW` state:
Screen Shot 2020-08-22 at 11 34 02 AM

The Battery Indicator popup now provides full information coming from `BATTERY_STATUS`:
Screen Shot 2020-08-22 at 11 34 08 AM

This pull also contains some architectural changes as new support:
* There is a mechanism in QGCMAVLink.h to expose MAVLink enums to qml
  * All magic numbers in Qml related to mavlink enums should transition to this mechanism
  * These are expose through the `MAVLink` singleton and the `QGroundControl.MAVLink` module
  * So this sort of thing works:
```
import QGroundControl.MAVLINK 1.0
property bool chargeStateValueAvailable: battery.chargeState !== MAVLInk.MAV_BATTERY_CHARGE_STATE_UNDEFINED
```
* VehicleBatteryFactGroup has been pulled out of Vehicle and is in it's own source files
  * How it is bolted into Vehicle creates the new pattern for getting all fact group code out of vehicle
  * The Fact Group code itself does the job of watching the mavlink traffic and updating the values it cares about
  * No more massive switch statements and tons of handleThis, handleThat methods for deal with all the messages and getting the data into the Fact Groups
  * So far on VehicleBatteryFactGroup works like this. I'll be cleaning up all the others which should modularize Vehicle and reduce it's size greatly.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -705,6 +705,7 @@ HEADERS += \
     src/Vehicle/TrajectoryPoints.h \
     src/Vehicle/Vehicle.h \
     src/Vehicle/VehicleObjectAvoidance.h \
+    src/Vehicle/VehicleBatteryFactGroup.h \
     src/VehicleSetup/JoystickConfigController.h \
     src/comm/LinkConfiguration.h \
     src/comm/LinkInterface.h \
@@ -924,6 +925,7 @@ SOURCES += \
     src/Vehicle/TrajectoryPoints.cc \
     src/Vehicle/Vehicle.cc \
     src/Vehicle/VehicleObjectAvoidance.cc \
+    src/Vehicle/VehicleBatteryFactGroup.cc \
     src/VehicleSetup/JoystickConfigController.cc \
     src/comm/LinkConfiguration.cc \
     src/comm/LinkInterface.cc \
--- a/src/FactSystem/FactGroup.cc
+++ b/src/FactSystem/FactGroup.cc
@@ -18,8 +18,6 @@
 #include <QFile>
 #include <QQmlEngine>
 
-QGC_LOGGING_CATEGORY(FactGroupLog, "FactGroupLog")
-
 FactGroup::FactGroup(int updateRateMsecs, const QString& metaDataFile, QObject* parent, bool ignoreCamelCase)
     : QObject(parent)
     , _updateRateMSecs(updateRateMsecs)
@@ -135,6 +133,8 @@ void FactGroup::_addFact(Fact* fact, const QString& name)
     }
     _nameToFactMap[name] = fact;
     _factNames.append(name);
+
+    emit factNamesChanged();
 }
 
 void FactGroup::_addFactGroup(FactGroup* factGroup, const QString& name)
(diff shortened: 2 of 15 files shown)


hamishwillee:
@DonLakeFlyer Love this. What will happen on the app bar if there are 10 batteries (re scaling etc).

DonLakeFlyer:
MockLink can fake two batteries which I used for testing. If you have 10 batteries you get 10 batteries and the toolbar scrolls. At some point I will think about doing things like not showing batteries marked a payload in the toolbar. And at some point somebody needs to figure out an overall single level battery status indication mechanism. But I don't want to go down that rathole again right now. So for now you get all the batteries!

hamishwillee:
Sensible. Remember of course we also have SMART_BATTERY messages :-). See comment here https://github.com/mavlink/mavlink/pull/1456

PR #9044, opened 2020-08-26
COMPONENT_INFORMATION:COMP_METADATA_TYPE_PARAMETER Changed parameter name regex support

DonLakeFlyer:
* Original support didn't work
* You now specify a group regex pattern in the name `.../(regex)/...` to match on param names
* You can refer to the captured text from the regex in `shortDesc` and `longDesc` using the `/1` tag which will be replaced
* Example:
```
    {
      "name": "CAL_ACC/(\\d+)/_EN",
      "type": "Int32",
      "group": "Sensor Calibration",
      "category": "System",
      "shortDesc": "Accelerometer /1 enabled",
      "default": 1,
      "decimalPlaces": 3,
      "minValue": -2.14748e+09,
      "maxValue": 2.14748e+09
    },
```

### DIFF
--- a/src/Vehicle/CompInfoParam.cc
+++ b/src/Vehicle/CompInfoParam.cc
@@ -76,30 +76,58 @@ void CompInfoParam::setJson(const QString& metadataJsonFileName, const QString& 
         }
 
         FactMetaData* newMetaData = FactMetaData::createFromJsonObject(parameterValue.toObject(), emptyDefineMap, this);
-        _nameToMetaDataMap[newMetaData->name()] = newMetaData;
+
+        QRegularExpression      regexNameIncludesRegex("/\\(.+\\)/");
+        QRegularExpressionMatch match = regexNameIncludesRegex.match(newMetaData->name());
+        if (match.hasMatch()) {
+            QString regexParamName = newMetaData->name();
+            regexParamName.replace(QRegularExpression("/(\\(.+\\))/"), "\\1");
+            newMetaData->setName(regexParamName);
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Note that this solves previous problems as well as being more powerful in its ability to collapse dups beyond a simple index.

hamishwillee:
Good to have match on the number in the body. FYI, this is the yaml definition in PX4. We could move transparently from one format to the other. Why match on `/1` rather than `{n}`?
```yaml
    - group: Battery Calibration
      definitions:
        ...
        BAT${i}_V_LOAD_DROP:
            description:
                short: Voltage drop per cell on full throttle
                long: |
                    This implicitly defines the internal resistance
                    to maximum current ratio for battery 1 and assumes linearity.
                    A good value to use is the difference between the
                    5C and 20-25C load. Not used if BAT${i}_R_INTERNAL is
                    set.

            type: float
            unit: V
            min: 0.07
            max: 0.5
            decimal: 2
            increment: 0.01
            reboot_required: true
            num_instances: *max_num_config_instances
            instance_start: 1
            default: [0.3, 0.3]
```

DonLakeFlyer:
At one point in my changes there was the possibility of multiple capture groups. Hence `/2'... might be possible. But then I backed away from that and with that the `/1` got left in. I'm likely take another pass at his before I update the spec.

hamishwillee:
OK, let me know how you go. Also if someone in PX4 committed to make corresponding changes. I can look at it, but I won't if someone else in the team is already doing so.

DonLakeFlyer:
> Also if someone in PX4 committed to make corresponding changes.

Yes. Lorenz, Beat and I will be discussion soon.

PR #9047, opened 2020-08-31
FirmwarePlugin: APM: Update link for stable version

patrickelectric:


DonLakeFlyer:
Does ArduPilot release all versions for all boards at the same time?

patrickelectric:
@DonLakeFlyer I think so, but I'm just updating the URL address, they have a new name for the old PX4.

DonLakeFlyer:
I'm wondering if this really works since it targets a single board to get versions from.

PR #9061, opened 2020-09-08
Customization Improvements

ArnaudThiercelin:
# Summary
This allows for QGroundControl to be customized in a structure were it is included as a submodule:

**MyQGC:**
/custom
/qgroundcontrol (submodule)

This is achieved by adding another directory search for the custom folder.
This adds clean structure from the repo side of things of which code is custom and which code is to go upstream.
It also allows for the use of other dependencies without tainting the qgroundcontrol repo.


qgroundcontrol.pro
```diff
@@ -86,17 +86,34 @@ exists(user_config.pri):infile(user_config.pri, CONFIG) {
...
+        message("* Found custom build *")
         CONFIG  += CustomBuild
         DEFINES += QGC_CUSTOM_BUILD
         # custom.pri must define:
         # CUSTOMCLASS  = YourIQGCCorePluginDerivation
         # CUSTOMHEADER = \"\\\"YourIQGCCorePluginDerivation.h\\\"\"
-        include($$PWD/custom/custom.pri)
+        include($$CUSTOM_PRI)
+    } else {
+        CUSTOM_PRI = $$PWD/../custom/custom.pri
```
DonLakeFlyer, line 104: I would change the out of tree dir name to be something less generic like 'qgc-custombuild' to be safer against collisions.

DonLakeFlyer:
Also I think the python updateqgrc.py will need to be updated to make it work if custom code is outside the tree.

ArnaudThiercelin:
@DonLakeFlyer I'm not really sure it's going to be such a problem to have ../custom because worst case, your build system won't work.
If you had another custom folder unrelated, you won't have the expected build files and it won't do anything but not build.
I did the changes nonetheless as it's also harmless.

For the updateqgrc.py, I've looked at it and I don't find any issues. The custom code isn't meant to be in both places or outside of it. It should really just be in either folder.
As part of it, I'm also working on changing the custom file to use CUSTOM_DIR that would be define by the QGC toolchain rather than fixed paths. This will come in another PR.

DonLakeFlyer:
> For the updateqgrc.py, I've looked at it and I don't find any issues.

The python script needs to read files from the regular QGC source tree. It does that by reaching up one directory level from it's current location which is going to be the wrong place if custom is outside the tree:
https://github.com/mavlink/qgroundcontrol/blob/master/custom-example/updateqrc.py#L31

ArnaudThiercelin:
@DonLakeFlyer fixed.

qgroundcontrol.pro
```diff
@@ -86,17 +86,34 @@ exists(user_config.pri):infile(user_config.pri, CONFIG) {
 # is present. It's useful to run "regular" builds to make sure you didn't
 # break anything.
 
+CUSTOM_DIR = $$PWD/qgc-custom
```
DonLakeFlyer, line 89: This is not needed and should be removed

qgroundcontrol.pro
```diff
@@ -86,17 +86,34 @@ exists(user_config.pri):infile(user_config.pri, CONFIG) {
...
+
 contains (CONFIG, QGC_DISABLE_CUSTOM_BUILD) {
     message("Disable custom build override")
 } else {
-    exists($$PWD/custom/custom.pri) {
-        message("Found custom build")
+    CUSTOM_PRI = $$PWD/qgc-custom/custom.pri
```
DonLakeFlyer, line 94: This breaks/changes existing behavior for all current users of custom builds. In place custom dir should stay named 'custom' to not break existing users.

custom-example/updateqrc.py
```diff
@@ -29,10 +29,16 @@ def process(src, exclusion, dst):
     file3.close()
 
 def main():
-    if(os.path.isfile(qgc_exc)):
-        process(os.path.join("../",qgc_rc), qgc_exc, qgc_rc)
-    if(os.path.isfile(res_exc)):
-        process(os.path.join("../",res_rc), res_exc, res_rc)
+    if (os.path.isfile(qgc_exc)):
+        qrc_path = os.path.join("../", qgc_rc)
+        if (!os.path.exists(qrc_path)):
```
DonLakeFlyer, line 34: python not operator is 'not', '!' doesn't work. At least it doesn't on the python installed on ubuntu 18.04

DonLakeFlyer:
Both inside tree custom build and outside trees are broken.

Custom build inside the build tree:
* Pull your changed branch
* Renamed `custom-example` to custom
* Delete qgroundcontrol.qrc and qgcresources.qrc files
* Rebuild resource files using `python updateqrc.py' command line from the `custom` dir
  * Fails on syntax error for '!' usage
  * Change to use correct 'not' logical operation, try again
* Build
  * Doesn't build because new build files expect the dir to be named 'qgc-custom' which breaks back compat
  * Change make file back to 'custom' for in tree builds, try again
  * All sorts of resource errors during build. Looks like the updateqrc.py didn't work correctly. Didn't look into why.

Custom dir outside the build tree:
* Pull your changed branch
* Move `custom_example` to `../qgc-custom`
* Delete qgroundcontrol.qrc and qgcresources.qrc files
* Rebuild resource files using updated `python updateqrc.py' command line from the `qgc-custom` dir
  * Error `IOError: [Errno 2] No such file or directory: './qgroundcontrol/qgroundcontrol.qrc'`. Needs to be `..` to go up and then back down to main qgc source.

PR #9063, opened 2020-09-09
fixed typo in VirtualJoystick.qml

roiegeron-spearuav:
activeVehicle - >_activeVehicle
caused the virtual joystick to malfunction


DonLakeFlyer:
Thanks for the fix

PR #9076, opened 2020-09-17
Add Nora support

czhj:
Add Nora support

### DIFF
--- a/src/comm/USBBoardInfo.json
+++ b/src/comm/USBBoardInfo.json
@@ -21,7 +21,7 @@
         { "vendorID": 1155, "productID": 41775,     "boardClass": "Pixhawk",    "name": "PX4 FMU ModalAI FCv1" },
         { "vendorID":12642, "productID": 75,        "boardClass": "Pixhawk",    "name": "PX4 DurandalV1" },
         { "vendorID": 4104, "productID": 1,         "boardClass": "Pixhawk",    "name": "PX4 FMU UVify Core" },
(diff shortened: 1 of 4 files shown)


thopiekar:
Could you add some more info about what you did and what it is needed for? It would ease the review and make people understand the purpose of your PR faster.
Thank you for your contribution! :)

src/comm/USBBoardInfo.json
```diff
@@ -21,7 +21,7 @@
         { "vendorID": 1155, "productID": 41775,     "boardClass": "Pixhawk",    "name": "PX4 FMU ModalAI FCv1" },
         { "vendorID":12642, "productID": 75,        "boardClass": "Pixhawk",    "name": "PX4 DurandalV1" },
         { "vendorID": 4104, "productID": 1,         "boardClass": "Pixhawk",    "name": "PX4 FMU UVify Core" },
-        { "vendorID": 12643, "productID": 76,       "boardClass": "Pixhawk",    "name": "PX4 CUAV X7(PRO)" },
+        { "vendorID": 12643, "productID": 76,       "boardClass": "Pixhawk",    "name": "CUAV Flight Controller" },
```
DonLakeFlyer, line 24: @czhj Just double-checking: The VID/PIDs for all your boards are the same?

cuhome: yes, all boards use the same usb id, because it is very expensive.

PR #9088, opened 2020-09-25
QGCPostLinkCommon: Remove duplicate echo

thopiekar:
We have `QMAKE_POST_LINK += echo "Post Link Common"` in line 12 already.


DonLakeFlyer:
Yup

PR #9101, opened 2020-10-06
VehicleLinkManager: Major refactor on how QGC manages comm links

DonLakeFlyer:
Most people know that QGC support the display of multiple vehicles. But QGC also supports more complex setups of comm link and vehicles. For example:
* Single vehicle which has multiple comm links associated with it
* A single comm link which exposes multiple vehicles through it
* Concept of a "High Latency" link which limits both uplink and download traffic to/from the vehicle
* The ability to switch between links if a vehicle has more than one. QGC itself always talks to the vehicle through what is called the "primary" link. QGC pulls download telemetry for all links available.
* QGC automatically switches between multiple links to a single vehicle based on priority

All of this leads to a high level of complexity in QGC code with respect to link<->vehicle management. That code has evolved over time to supports all these features. And as happens that lead to it being a mess of spaghetti and complexity. It also leads to various crashes in QGC when links would go away at just the wrong time. In order to fix this, link management was completely re-written into a separate VehicleLinkManager class which hangs off the Vehicle.

This pull is the first of multiple to fix link management problems and includes these changes:
* New VehicleLinkManager class
* LinkInterface and LinkConfiguration objects are always handed out of reference as either strong or weak referenced pointers. This prevents crashes caused by different disconnect/shutdown sequencing.
* Better/correct reporting on communication lost on a single link in a situation where a vehicle has multiple links.
* Comprehensive unit tests to test all the crazy combinations
* New MockLink support:
  * Create multiple links for a single vehicle
  * Better support for creating simulated high latency links
  * Ability to simulate things like communication loss on a MockLink
* Change both the internal terminology and the terminology exposed to the user for multiples links. The "Primary" link is the link which QGC uses to talk QGC->Vehicle. "Secondary" links are only used to get telemetry from Vehicle->QGC.
* New MultiSignalSpyV2 implementation which doesn't require to manually setup the signals. It does it automatically through introspection on the QMetaObject info. This makes usage in units test much easier to use.

Work remaining:
* Much more work on QGC to correctly use the primary link for what I call mini-protocols. For example the request to download all params takes place on the primary link. If the primary link changes in the middle of the parameter download for whatever reason the download should fail. Right now depending on how unlucky you are it may fail or it may crash.
* Better MockLink support and unit tests for Single Link -> Multiple Vehicle case.


DonLakeFlyer:
@hamishwillee Here is how High Latency links work in QGC. 

Whenever a high latency is made primary:
  * First link to the vehicle is a high latency link.
  * User selected the high latency link to be primary.
  * QGC switched to this link automatically as primary due to other links disconnecting or having comm lost.

QGC will send the `MAV_CMD_CONTROL_HIGH_LATENCY` command to start transmission on the link.

If the opposite occurs in that a high latency is changed away from primary. QGC will send the command to stop transmission on the link.

This is done to limit the transmission across the high latency link to only when necessary. The reason being it is likely a high cost ($$$ per byte) link that you only want to use when you absolutely need it.

PR #9103, opened 2020-10-08
[Hold for 4.2] add pseudo oblique function to CMD 206 CAM_TRIGG_DIST

igorsgcampos:
**Problem**
A solution commonly used to increase the coverage and efficiency in mapping surveys is to have multiple cameras at oblique angles to capture a wider areas per flight line, this also has potential benefits of capturing more data from facades and tall objects. However, adding cameras increase cost, weight and complexity of the equipment, on the other hand, mapping cameras are generally capable of shooting multiple frames per second and are usually triggered about once every two seconds or less frequently in mapping missions. We could better use the hardware capabilities if we moved the camera mount between shots while decreasing the shooting distance to keep the overlap settings.

**Solution**
I decided to implement a solution that uses the already existing MAV_CMD_DO_SET_CAM_TRIGG_DIST, plus some unused parameters, 4~6, to issue MAV_CMD_DO_MOUNT_CONTROL between camera shots to move the camera from left to right in stages, right after it passes half the shooting distance, allowing time for mount actuation and camera exposure. To keep the frontlap the solution expects the GCS to adjust the trigger distance accordingly. The parameters usage are as follows:

4: number of oblique positions (this will generally be 2 or 3, but could be more, depending on desired frontlap, flying speed and hardware capabilities (how fast are the camera and mount))
5: the maximum angle to roll the camera left and right from the center, the solution calculates the interval angle between positions from this information and the previous parameter
6: the pitch angle for mounts that support pitching

**Implementation**
I proposed the change to mavlink standard here: https://github.com/mavlink/mavlink/pull/1508
I implemented the proposed solution here: https://github.com/PX4/Firmware/pull/15882

**Alternative**
Other way would be to add a new command, but since it has many aspects that are essentially the same as the CAM_TRIGG_DIST, I thought it would be better to add a few parameters to the existing command, as it would also allow it to work under Mavlink 1 with the limited number of 255 commands.

**Test data / coverage**
I tested it with the gazebo_typhoon_h480 model and checked the outcome from the video stream in QGC.

**Additional context**
Later I plan to write a paper comparing the gains in area coverage and any accuracy change from using this solution with relation to the traditional survey mode. I am feeling confident about a positive outcome.


DonLakeFlyer:
> Later I plan to write a paper comparing the gains in area coverage and any accuracy change from using this solution with relation to the traditional survey mode. I am feeling confident about a positive outcome.

To me that needs to be the first step to prove this will actually work. With specific details on how things like front and sidelap need to be calculated based on settings.

@antiheavy FYI

Antiheavy:
> @Antiheavy FYI

seems ok.  or at least, this shouldn't negatively impact the way we use the camera trigger commands.

igorsgcampos:
build will break until mavlink submodule is updated with the changes from https://github.com/mavlink/mavlink/pull/1508

DonLakeFlyer:
I don't know if you've rebased to master recently. But of not my recent changes to support terrain adjustment in turnaround segments has likely caused conflicts with this code. Mainly because the building of mission items had to be reworked. It should be fairly straightforward to understand how the new code works. There wasn't any other way to fix this problem without the change. Sorry about that.

igorsgcampos:
> I don't know if you've rebased to master recently. But of not my recent changes to support terrain adjustment in turnaround segments has likely caused conflicts with this code. Mainly because the building of mission items had to be reworked. It should be fairly straightforward to understand how the new code works. There wasn't any other way to fix this problem without the change. Sorry about that.

No problem Don, already rebased/fixed my commits. Thanks for the heads-up!

igorsgcampos:
@DonLakeFlyer one more question, when I build this with the target to Android apk, the file ends up being ~42MB, whereas the latest daily is ~53MB. Plus, my build does not get the solo video link, despite the fact that it is already rebased onto master, whereas the latest daily does get it successfully.

Having the camera feed would help seeing the camera behavior, but I can either have the feed or the oblique survey command. Could you point me to instructions on how to get this properly working on my android build?

DonLakeFlyer:
Sounds like you may not have android gstreamer installer: https://github.com/mavlink/qgroundcontrol/tree/master/src/VideoReceiver

igorsgcampos:
I followed the instrucitons, but I must have some conflicting software (maybe ROS, it likes to give headaches):

The following packages have unmet dependencies:
 gstreamer1.0-python3-dbg-plugin-loader : Conflicts: gstreamer1.0-python3-plugin-loader but 1.14.5-0ubuntu1~18.04.1 is to be installed
 gstreamer1.0-python3-plugin-loader : Conflicts: gstreamer1.0-python3-dbg-plugin-loader but 1.14.5-0ubuntu1~18.04.1 is to be installed
E: Unable to correct problems, you have held broken packages.

DonLakeFlyer:
Qmake output should tell you whether it is building with or without gstreamer support.

DonLakeFlyer:
Qmake output should tell you whether it is building with or without gstreamer support. Any linux complaints should be irrelevant I believe since the package is for android not linux.

igorsgcampos:
Actually, I noticed it was a conflict between 2 packages on the list and kept the dbg one, it installed successfully, but the apk build still has no video, is there any config to tell the compiler to build with video?

igorsgcampos:
nevermind, I noticed additional instructions for android at the bottom, I will follow those instead of the linux ones.

igorsgcampos:
Cleaning and rebuilding seemed to solve it! Thanks for the directions!

igorsgcampos:
Damn, I'm having a hard time fixing these tests.

DonLakeFlyer:
FYI: I haven't look at what is failing. But there is currently some sort of race condition in master VehicleLinkManagerTest which can cause it to fail intermittently. Haven't figured it out yet.

igorsgcampos:
It is likely due to this:

> I also appended to the surveys DO_MOUNT_CONTROL commands to point the gimbal down and forward before and after the survey, as this had to be manually done. If the user forgot about this, it would end up with a survey facing forward.

The mission is having an unexpected number of waypoints or having the wrong waypoint types.

DonLakeFlyer:
If you add a survey to a Plan the gimbal will already be pointed down since the Mission Start will be modified to do this when the survey gets added. Why isn't that good enough?

igorsgcampos:
It wasn't behaving like this when I tested a couple months ago with PX4 in SITL. It kept the camera pointing at 0,0,0 orientation, which faces forward normally. The one at the end is needed to ensure that the camera is at 0 degrees roll at the end with this new command.  I noticed the need for it when I landed the Solo and the gimbal touched the ground because it was at 20 degrees roll left.

src/MissionManager/TransectStyleComplexItem.cc
```diff
@@ -1128,6 +1147,9 @@ void TransectStyleComplexItem::_buildAndAppendMissionItems(QList<MissionItem*>&
             } else {
                 _appendWaypoint(items, missionItemParent, seqNum, mavFrame, 0 /* holdTime */, coordInfo.coord);
             }
+            if (lastSurveyExit) {
```
igorsgcampos, line 1150: @DonLakeFlyer This variable is not working as intended. It is not entering this condition if there is a turnaround. It wasn't adding the camera mount command at the end of the survey mission.

igorsgcampos:
@DonLakeFlyer I have been busy writing a paper about this new feature, I hope I will be able to take a better look at the tests soon, but I have experienced an issue where I tried using an existing boolean variable that is supposed to be true when it is the last transect of the survey. Can you help me determining the correction for it?

src/MissionManager/TransectStyleComplexItem.cc
DonLakeFlyer: Are you sure you are fully up to date with latest master. There were a pile of bugs there I fixed.

igorsgcampos: It was already up to date, found the issue, I was adding only after lastSurveyExit, not after lastExitTurnaround. This fixed it:
```
            if (lastExitTurnaround) {
                _appendCameraMountPitchItem(items, missionItemParent, seqNum, 0);
            }
```

igorsgcampos:
Dear @DonLakeFlyer,

I have finally submitted the paper I was writing on this new feature yesterday. The results are very promising, in field experiments an 80% gain in efficiency was observed.

I believe I am still going to add a toggle to disable the footprint adjustments for surveys focused on 3D model generation. In these usecases the transect separation would be the same, but less flights would be necessary to cover an area with nadir and oblique images at the 4 directions and/or oblique angles, as these images would be acquired in a double-grid mission. I believe a toggle like:

Survey Type: ○ 2D ◉ 3D

DonLakeFlyer:
Thanks so much for your patience @igorsgcampos. I think I may have finally finished stable today. So hope to get back to this soon.

igorsgcampos:
@DonLakeFlyer I added the radio button to select the survey type and I think I fixed the tests to account for the additional DO_MOUNT_CONTROL at the beginning and end of survey missions, but the test failed at `StructureScanComplexItemTest::_testSaveLoad()`, but I didn't touch any structure scan code:

```
FAIL!  : StructureScanComplexItemTest::_testSaveLoad() 'newItem->load(items[0].toObject(), 10, errorString)' returned FALSE. ()
   Loc: [/home/travis/build/mavlink/qgroundcontrol/src/MissionManager/StructureScanComplexItemTest.cc(118)]
```

I cleaned up the code a bit and pushed a new commit, maybe it was just a random failure. The full log is available at:
https://travis-ci.org/github/mavlink/qgroundcontrol/jobs/756473153

igorsgcampos:
@DonLakeFlyer it appears it wasn't a random failure. But this one I am clueless about the root cause. I would appreciate your input whenever you find some time to have a look at it.

DonLakeFlyer:
The structure can unit test is failing in the new code you added to CameraCalc because of the `_jsonCamposSurvey2DKey` being listed as a bool but being defined as a double in the json. Not sure why those keys exist at all for a structure scan. I would not expect them to exist in that case since your new code should not come into play there. Just debug through the calls and you can see it.

igorsgcampos:
> The structure can unit test is failing in the new code you added to CameraCalc because of the `_jsonCamposSurvey2DKey` being listed as a bool but being defined as a double in the json. Not sure why those keys exist at all for a structure scan. I would not expect them to exist in that case since your new code should not come into play there. Just debug through the calls and you can see it.

@DonLakeFlyer Thanks so much, I fixed it now, it was my mistake as I copied and pasted the code. I wouldn't have figured out this quick without your help. I believe it will now pass all tests.

mrpollo:
@igorsgcampos nice to see you are still going strong on your journey, thank you for the amazing contributions!

@DonLakeFlyer have all your comments been addressed, is there anything missing before merging?

igorsgcampos:
@mrpollo Thanks for all the support! It will be nice to see this running in the future releases and getting the feedback from the users.

@DonLakeFlyer I forgot to mention this in the previous comment, but to implement it I just followed what was already in place, using the same logic that the other variables in the survey settings grid had. Sorry about the unit testing, I am not familiar with Qt creator (in fact I only use it to compile/run, but prefer to use the VS code otherwise) and didn't know how to run them, that is why I fixed the others where it was easier to identify the root cause from the logs, but got stuck in this one.

DonLakeFlyer:
@igorsgcampos I'm finally getting to this. Sorry for the massive delay. You had talked about writing a paper. Do you have anything I could read like that?

DonLakeFlyer:
> Sorry about the unit testing, I didn't know how to run them as I am not familiar with Qt creator

Unit tests don't have anything to do with Qt Creator. They are run from the command line option '--unittest' to run all or '--unittest:UnitTestName' to run a specific test.

DonLakeFlyer:
I created a simple one of these using master PX4, kept the defaults for the values and testing using SITL typhoon and the gimbal never did anything. It stayed pointing forward the whole time.

igorsgcampos:
> I created a simple one of these using master PX4, kept the defaults for the values and testing using SITL typhoon and the gimbal never did anything. It stayed pointing forward the whole time.

I did have to make a few adjustments to the SITL (I remember the tilt angle was not enough in the Typhoon settings, but this was later corrected), I will check if they are all synced with my fork of the simulator. A video of it working is shown at: https://www.youtube.com/watch?v=p3yD-LFPS9E

igorsgcampos:
> @igorsgcampos I'm finally getting to this. Sorry for the massive delay. You had talked about writing a paper. Do you have anything I could read like that?

The paper is finished, but still under peer review, I can send you, but kindly request you to keep it undisclosed to anyone else.

Since I couldn't find your e-mail address I sent it to your LinkedIn profile.

DonLakeFlyer:
> but kindly request you to keep it undisclosed to anyone else.

No problem. I just want to use it to understand how it works more deeply.

DonLakeFlyer:
Aha! Now I see why I was confused. In looking at that paper what you describe as "Roll" is actually manipulating the yaw axis of the gimbal not the roll axis. Is that correct?

DonLakeFlyer:
Or maybe I'm misunderstanding how gimbal angles work. So for example if I set Pitch to 0 degrees. And then set a roll value the gimbal would actually change Yaw in the horizontal plane, not roll.

DonLakeFlyer:
Sorry I'm confused. It is really roll.

igorsgcampos:
Yes, if the pitch is 0 I think there will be no effect on roll, I haven't tested this scenario as it doesn't make much sense for survey missions.

igorsgcampos:
Hey @DonLakeFlyer,

What are your thought on this? I would appreciate your feedback after you tried it.

Best regards,

Igor

DonLakeFlyer:
> I created a simple one of these using master PX4, kept the defaults for the values and testing using SITL typhoon and the gimbal never did anything. It stayed pointing forward the whole time.

I never progressed past this. So I can't test it. Can you verify it working against SITL master?

igorsgcampos:
> > I created a simple one of these using master PX4, kept the defaults for the values and testing using SITL typhoon and the gimbal never did anything. It stayed pointing forward the whole time.
> 
> I never progressed past this. So I can't test it. Can you verify it working against SITL master?

There is something fishy going on with the simulator, but the messages are being sent and received by CAMPOS/vmount logic respectively, which I checked through the some logging messages:

```
INFO  [navigator] Takeoff to 50.0 meters above home.
INFO  [commander] Takeoff detected
[Wrn] [Publisher.cc:185] Queue limit reached for topic /gazebo/default/user_camera/pose, deleting message. This warning is printed only once.
INFO  [vmount] Configured primary gimbal control sysid/compid from 0/0 to 1/1
INFO  [vmount] received VEHICLE_MOUNT_MODE_MAVLINK_TARGETING: roll/pitch/yaw: 0.00/-90.00/0.00
INFO  [tone_alarm] notify negative
INFO  [camera_trigger] received OBLIQUE_SURVEY
INFO  [camera_trigger] adjust roll: sys/comp/angle: 1/1/0.00
INFO  [vmount] received VEHICLE_MOUNT_MODE_MAVLINK_TARGETING: roll/pitch/yaw: 0.00/-90.00/0.00
    1 image files updated
INFO  [camera_trigger] adjust roll: sys/comp/angle: 1/1/20.00
INFO  [vmount] received VEHICLE_MOUNT_MODE_MAVLINK_TARGETING: roll/pitch/yaw: 20.00/-90.00/0.00
    1 image files updated
INFO  [camera_trigger] adjust roll: sys/comp/angle: 1/1/-20.00
INFO  [vmount] received VEHICLE_MOUNT_MODE_MAVLINK_TARGETING: roll/pitch/yaw: -20.00/-90.00/0.00
    1 image files updated
```

It is weird as I attempted the simulation several times while adding these messages, but in a few instances, the vehicle tilted the camera as it was supposed to do in the gazebo simulator.

The messages were added in camera_trigger.cpp:807 and input_mavlink.cpp:636:

`PX4_INFO("adjust roll: sys/comp/angle: %i/%i/%.2f", _target_system, _target_component, (double) vcmd.param2);`


`PX4_INFO("received VEHICLE_MOUNT_MODE_MAVLINK_TARGETING: roll/pitch/yaw: %.2f/%.2f/%.2f",(double) vehicle_command.param2, (double) vehicle_command.param1, (double) vehicle_command.param3);`

igorsgcampos:
Perhaps @JaeyoungLim could also take a look on this and help us out.

igorsgcampos:
Perhaps @Jaeyoung-Lim could also take a look on this and help us out.

DonLakeFlyer:
@igorsgcampos Lost track of this. Any updates on getting SITL working so I can test this?

igorsgcampos:
@DonLakeFlyer Sorry for taking a while to reply, I was pretty busy this last week. BTW, the paper I mentioned is now finally published, make sure you check the acknowledgements section at: https://doi.org/10.5194/isprs-archives-XLIII-B1-2021-139-2021

I will attempt to print the roll angle of the camera in the log instead of relying on the gazebo visualization for it. I am pretty sure it is unrelated to my changes.

igorsgcampos:
@DonLakeFlyer I will join tomorrow's dev-call, can you mark this for discussion please?

julianoes:
Looks like the feature went into MAVLink and PX4 but not here yet. @igorsgcampos any chance you want to resurrect the PR?

DonLakeFlyer:
I'm going to close this since it's been sitting here for so long. Can be re-opened if new work is done against it.

PR #9110, opened 2020-10-12
Remove useless codes 

2997ms:
Fix the useless code in RadioComponentController.cc

### DIFF
--- a/src/AutoPilotPlugins/Common/RadioComponentController.cc
+++ b/src/AutoPilotPlugins/Common/RadioComponentController.cc
@@ -866,6 +866,7 @@ void RadioComponentController::_loadSettings(void)
     settings.beginGroup(_settingsGroup);
     _transmitterMode = settings.value(_settingsKeyTransmitterMode, 2).toInt();
     settings.endGroup();
+
     if (!(_transmitterMode == 1 || _transmitterMode == 2)) {
         _transmitterMode = 2;
     }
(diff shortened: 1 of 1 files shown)


2997ms:
@DonLakeFlyer

src/AutoPilotPlugins/Common/RadioComponentController.cc
```diff
@@ -867,9 +867,7 @@ void RadioComponentController::_loadSettings(void)
     _transmitterMode = settings.value(_settingsKeyTransmitterMode, 2).toInt();
     settings.endGroup();
 
-    if (_transmitterMode != 1 || _transmitterMode != 2) {
-        _transmitterMode = 2;
-    }
+    _transmitterMode = 2;
```
DonLakeFlyer, line 870: Not sure why you think this is not needed. The current code is correct and required to validate the two valid settings from _transmitterMode.

2997ms: `if (_transmitterMode != 1 || _transmitterMode != 2)`
This is always true.

DonLakeFlyer: Ah, right it's screwed up. Should be `if (!(_transmitterMode == 1 || _transmitterMode == 2))`

2997ms: @DonLakeFlyer Addressed

2997ms: @DonLakeFlyer

DonLakeFlyer:
Great thanks

PR #9112, opened 2020-10-14
Param load from file: Support unknown params, diff ui

DonLakeFlyer:
* You no longer need to load a parameter file multiple times to get params updated which were not currently being reporting from the vehicle
* You are presented with a diff ui which shows you the differences between the file and the vehicle
* You can selectively send specific param updates from the file

Screen Shot 2020-10-14 at 11 48 08 AM


dagar:
Awesome!

Issue #8654, opened 2020-04-12
[Feature Request] warn user when trying to save a Preset name with dis-allowed characters


Antiheavy:
**Problem:**
If you try to save a preset with a dis-allowed character (e.g. "map 25/25") the preset fails to save and there is no feedback to the user.
We've heard about this causing confusion with a bunch of users already (seems everyone wants to use a special character, probably a "/" in many cases).

**Request:**
Warn the user when saving Preset names with special characters is not allowed and let them type in a different name.

![image](https://user-images.githubusercontent.com/12117027/79082131-12089d00-7ce9-11ea-97e1-2ac549800e65.png)

PR #9113, opened 2020-10-15
[FIX for issue 8654] Add form validation for survey preset names

hankminden:
When trying to add a preset name with a disallowed character we would fail silently which is a confusing experience for the user. Instead it should tell the user they cannot use that character and allow them to choose a different preset name. 

Original issue: 
https://github.com/mavlink/qgroundcontrol/issues/8654

### DIFF
--- a/src/PlanView/SurveyItemEditor.qml
+++ b/src/PlanView/SurveyItemEditor.qml
@@ -373,8 +373,12 @@ Rectangle {
 
             QGCViewDialog {
                 function accept() {
-                    if (presetNameField.text != "") {
-                        _missionItem.savePreset(presetNameField.text)
+                    if (presetNameField.text.includes("/")){  // form validation because savePreset will fail silently if presetNameField includes "/"
+                        presetNameError.text = "/ is not allowed in preset names"
+                    } else if(presetNameField.text === ""){
+                        presetNameError.text = "Preset name must not be empty"
+                    } else {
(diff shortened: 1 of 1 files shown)


src/PlanView/SurveyItemEditor.qml
```diff
@@ -373,8 +373,12 @@ Rectangle {
 
             QGCViewDialog {
                 function accept() {
-                    if (presetNameField.text != "") {
-                        _missionItem.savePreset(presetNameField.text)
+                    if (presetNameField.text.includes("/")){  // form validation because savePreset will fail silently if presetNameField includes "/"
+                        presetNameError.text = "/ is not allowed in preset names"
```
DonLakeFlyer, line 377: These strings should use qsTr("...") for translating.

src/PlanView/SurveyItemEditor.qml
```diff
@@ -398,6 +402,10 @@ Rectangle {
                         id:                 presetNameField
                         Layout.fillWidth:   true
                     }
+                    QGCLabel {
+                        id:                 presetNameError
+                        color:              "red"
```
DonLakeFlyer, line 407: The color here should use QGCPalette `warningText` color instead of a hardcoded color.

DonLakeFlyer:
@hankfinney Updates?

PR #9115, opened 2020-10-16
PX4 Setup - Power: Fix broken voltage/current divider calculators

DonLakeFlyer:
Broken by previous multi-battery pull

### DIFF
--- a/src/AutoPilotPlugins/PX4/PowerComponent.qml
+++ b/src/AutoPilotPlugins/PX4/PowerComponent.qml
@@ -418,7 +418,8 @@ SetupPage {
                     title:   qsTr("Calculate Voltage Divider")
                     buttons: StandardButton.Close
 
-                    property var _controller: controller
+                    property var        _controller:        controller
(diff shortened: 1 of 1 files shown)


bys1123:
Fixed https://github.com/PX4/Firmware/issues/15986

bys1123:
Do you have any idea that why old 3.5.6 version can load these new BATn_* parameters? Thanks!

DonLakeFlyer:
> Do you have any idea that why old 3.5.6 version can load these new BATn_* parameters?

It can't. The firmware syncs BAT1_* params to BAT_ behind the scenes.

bys1123:
Oh, can't, right, sorry, typo.
So, new firmware can't compatible on battery settings/messages with old QGC? It didn't syncs BAT_* params to BAT1_* backwards.

DonLakeFlyer:
> It didn't syncs BAT_* params to BAT1_* backwards.

I believe there is still a bug on that in firmware which was fixed here: https://github.com/PX4/Firmware/pull/15985.

Stable should be able to sort of work with single battery setup if the firmware parameter sync stuff works correctly. It is missing setup for BAT1_SOURCE though so in reality not so sure if that causes a problem.

PR #9117, opened 2020-10-17
Allow use of AutoConnect and HighLatency options for most link types.

### LINKED ISSUE #9064: TCP connection auto-connect
When using a TCP connection that you've already configured it would be nice if it could attempt to auto-connect (configurable).

brad112358:
I've often wished to have my Bluetooth telemetry link connected automatically at startup, but the option to do so was disabled for some reason.  When I checked, I found a similar feature request for TCP links.  Looking at the code, it was not clear to me why only specific link types allowed the high latency option to be set.  Latency is not unique only to specific link types as there is no way of knowing what specific air transport might be at the other end of a connection.   The ability to automatically connect on startup is useful for most link types.

This trivial change modifies the LinkConfiguration class to allow the user to choose these options.   Remove some now unneeded overrides and add overrides to disable these options for Mock links, the only link type where it may not make sense to allow them.

fixes #9064


DonLakeFlyer:
> Looking at the code, it was not clear to me why only specific link types allowed the high latency option to be set.

I think that make sense. But instead of just changing some bools around such that they are always true. Can you instead just remove to code totally which deals with optional high latency?

> I've often wished to have my Bluetooth telemetry link connected automatically at startup, but the option to do so was disabled for some reason.

I vaguely remember some history there as some reason why. It's probably going to take some poking around in really old pulls to figure out if any information was left around as to why.

DonLakeFlyer:
Here it is: https://github.com/mavlink/qgroundcontrol/pull/4264 with the comment:

"Autoconnect should only be allowed for UDP links since other link types will need more sophisticated code to check whether the port exists before it tries to connect. Otherwise it will through continuous errors when the device is not connected."

Which I think is still correct but will need to think about it a bit.

brad112358:
> I think that make sense. But instead of just changing some high latency bools around such that they are always true. Can you instead just remove to code totally which deals with optional high latency?

I assume you mean remove the code which disables the checkbox for high latency, not all the high latency logic.  But Mock links still have the high latency checkbox disabled in my pull request.  Should high latency be allowed there for testing that logic?  I suppose it could be useful.  In this case, yes, we can remove highLatencyAllowed and isHighLatencyAllowed.  Do you want to allow the high latency flag for Mock links and, thus, all links?

brad112358:
> Autoconnect should only be allowed for UDP links since other link types will need more sophisticated code to check whether the port exists before it tries to connect. Otherwise it will through continuous errors when the device is not connected.

I don't think this is currently a concern.  Setting the AutoConnect flag does not currently cause links to be automatically reconnected when disconnected or when the link fails in some way.  IsAutoConnect() is only called at startup.  Though, it would be nice if a reconnect were attempted for Bluetooth connections when they disconnect, that is not currently the case, and this PR does not attempt to fix that.

DonLakeFlyer:
> I assume you mean remove the code which disables the checkbox for high latency, not all the high latency logic.

Correct.

> But Mock links still have the high latency checkbox disabled in my pull request.

MockLinks can simulate high latency links so it should be on for them. I'm pretty sure it turned in on in a recent pull. Your master may be a little behind.

> I don't think this is currently a concern. Setting the AutoConnect flag does not currently cause links to be automatically reconnected when disconnected or when the link fails in some way.

Great thanks for the explanation. I think it's fine then on all links. For now I'd leave the ability to turn off support for it just in case we discover something down the road which causes problems on some link type. If it all works fine for a while then at some point I'll remove the optional support code to clean up.

brad112358:
OK.  So, it sounds like you are suggesting allowing high latency setting for Mock Links.  That makes sense.  And, leave the logic in place which can grey out the option for autoconnect.  Is it OK to leave the autoconnect option greyed out for Mock Links?  What about the logic which can grey out the high latency option?  Leave it but with no link type using it?  Any other change I should make to the PR?

DonLakeFlyer:
You know what might as well clean everything up now. If there are problems I can put stuff back. So remove all the "optional" nature of checking high latency and autoconnect. All links support them and they are always shown.

DonLakeFlyer:
Yup, perfect. Thanks.

PR #9118, opened 2020-10-17
Clang 12 compile error fixes

LorenzMeier:
These are necessary to get master to compile with Clang 12. At first sight it might look like a naive fix - what was initially a reference is now a copy. However, if you trace the code path in the inner loop, you will find that in all of these cases either the inner loop already created a local copy, and/or discarded the const modifier (which requires again taking a copy).

So based on my quick assessment the new compiler version keeps us more honest about the invariants.

### DIFF
--- a/src/AutoPilotPlugins/AutoPilotPlugin.cc
+++ b/src/AutoPilotPlugins/AutoPilotPlugin.cc
@@ -61,7 +61,7 @@ void AutoPilotPlugin::parametersReadyPreChecks(void)
     _recalcSetupComplete();
 
     // Connect signals in order to keep setupComplete up to date
-    for(const QVariant componentVariant: vehicleComponents()) {
+    for(QVariant componentVariant: vehicleComponents()) {
         VehicleComponent* component = qobject_cast<VehicleComponent*>(qvariant_cast<QObject *>(componentVariant));
         if (component) {
             connect(component, &VehicleComponent::setupCompleteChanged, this, &AutoPilotPlugin::_recalcSetupComplete);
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
Seems like a really odd change. Is there no way to get these range_for things to not copy? How did you complile with clang12?

DonLakeFlyer:
> Is there no way to get these range_for things to not copy?

Or does it just have to do with how QVariant and QJsonValue works?

dagar:
> > Is there no way to get these range_for things to not copy?
> 
> Or does it just have to do with how QVariant and QJsonValue works? What's the error you get?

Screen Shot 2020-10-18 at 2 25 23 PM

dagar:
This one's better.
Screen Shot 2020-10-18 at 2 29 02 PM

DonLakeFlyer:
OK, understood. 

I've been fixing the first example of not creating copies as I see them. I think errors as warnings is not on for all OS. I'll have to look at that. Should be build failure and they spit out on current Ubuntu 18.* compiler.

And then I assume the warning about reference not doing anything must be new since I've never seen that one before.

src/AutoPilotPlugins/AutoPilotPlugin.cc
```diff
@@ -61,7 +61,7 @@ void AutoPilotPlugin::parametersReadyPreChecks(void)
     _recalcSetupComplete();
 
     // Connect signals in order to keep setupComplete up to date
-    for(const QVariant componentVariant: vehicleComponents()) {
+    for(QVariant componentVariant: vehicleComponents()) {
```
DonLakeFlyer, line 64: Seems strange though that QVariantList doesn't support reference. I guess it is what it is.

PR #9133, opened 2020-10-20
[NEW] - Configure branding logo margins

TheLegendaryJedi:
Some logos that have an image together with text become not readable or not recognizable at all with the default margins. So I just added in settings the possibility to change that logo margins for a custom ones.

By default I left the ones already configured.

Thanks

### DIFF
--- a/src/ui/preferences/GeneralSettings.qml
+++ b/src/ui/preferences/GeneralSettings.qml
@@ -33,7 +33,9 @@ Rectangle {
     property Fact _savePath:                            QGroundControl.settingsManager.appSettings.savePath
     property Fact _appFontPointSize:                    QGroundControl.settingsManager.appSettings.appFontPointSize
     property Fact _userBrandImageIndoor:                QGroundControl.settingsManager.brandImageSettings.userBrandImageIndoor
+    property Fact _userBrandImageIndoorMargin:          QGroundControl.settingsManager.brandImageSettings.userBrandImageIndoorMargin
     property Fact _userBrandImageOutdoor:               QGroundControl.settingsManager.brandImageSettings.userBrandImageOutdoor
+    property Fact _userBrandImageOutdoorMargin:         QGroundControl.settingsManager.brandImageSettings.userBrandImageOutdoorMargin
     property Fact _virtualJoystick:                     QGroundControl.settingsManager.appSettings.virtualJoystick
     property Fact _virtualJoystickAutoCenterThrottle:   QGroundControl.settingsManager.appSettings.virtualJoystickAutoCenterThrottle
 
@@ -61,6 +63,7 @@ Rectangle {
 
     property string gpsDisabled: "Disabled"
     property string gpsUdpPort:  "UDP Port"
+
 
     readonly property real _internalWidthRatio: 0.8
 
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
Why is this both in App settings and BrandImage settings?

TheLegendaryJedi:
Hi, I just checked how the image was implemented and did the same for the margins. Should I remove it from one of them?

DonLakeFlyer:
Can you include screenshots of the new settings ui? Also did you test with and without margins that the brand image continue to show/hide appropriately based on window size?

TheLegendaryJedi:
Yes I can, and I tested it.

Screenshot from 2020-10-27 17-22-31
Screenshot from 2020-10-27 17-22-28
Screenshot from 2020-10-27 17-22-19
Screenshot from 2020-10-27 17-22-15
Screenshot from 2020-10-27 17-21-55
Screenshot from 2020-10-27 17-21-40
Screenshot from 2020-10-27 17-21-37

src/ui/preferences/GeneralSettings.qml
```diff
@@ -1078,13 +1099,34 @@ Rectangle {
                                     onAcceptedForLoad:  _userBrandImageOutdoor.rawValue = "file:///" + file
                                 }
                             }
+
+
+                            QGCLabel {
+                                text:           qsTr("Outdoor Image Margin")
+                                visible:        userBrandImageOutdoorMargin.visible
+                            }
+                            SpinBox {
```
DonLakeFlyer, line 1108: We don't have a QGC version of a SpinBox. You can't use the Qt controls directly since they don't work with the QGC pallete selection/theming. This will need to just a a QGCTextField.

TheLegendaryJedi: Can we make a SpinBox for QGC?

DonLakeFlyer: > Can we make a SpinBox for QGC?

It depends on who "we" is :) I don't have time to do it myself right now. But I assume it should be doable. If you do, start with controls 2 version.

TheLegendaryJedi: Ok, I'll try to do it. Any documentation to helping me in this journey :rofl: ?

DonLakeFlyer: Not really. Other than looking at how the other QGCFoo controls do it. They need to respect the palette colors and font sizing. I would change this to a QGCTextField to complete this pull. And then try doing the spin box in a separate pull. It certainly would be nice to have a spin box as a usable QGC control.

TheLegendaryJedi: Ok thanks :)

TheLegendaryJedi:
Hi @DonLakeFlyer, just added the required QGCSpinBox

DonLakeFlyer:
Could you move the QGC SpinBox to it's own separate pull and make this pull with just an existing FactTextField. And then we can iterate on the spinbox separately.

For the separate spinbox pull the place to add the control to the QmlTest.qml page which you can see here. There is where new control begin there life and then the page is used to verify they are working correctly.
Screen Shot 2020-11-05 at 11 03 54 AM

PR #9136, opened 2020-10-21
Fixed 5.15 build on Android (removed deprecated API)

akontsevich:
Fixed 5.15 build on Android - removed deprecated API: `QLatin1Literal() -> QLatin1String()`.

**P.S. BTW**, Why Android build on 5.15 have multiple link errors? Any ideas how to fix? Update AndroidManifest.xml, @patrickelectric : https://wiki.qt.io/Qt_for_Android_known_issues ?

### DIFF
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -57,7 +57,6 @@ linux {
             -Wno-parentheses-equality           # android gstreamer header files
         QMAKE_CFLAGS_WARN_ON += \
             -Wno-unused-command-line-argument   # from somewhere in Qt generated build files
-        QMAKE_LINK += -nostdlib++ # Hack fix?: https://forum.qt.io/topic/103713/error-cannot-find-lc-qt-5-12-android
         target.path = $$DESTDIR
         equals(ANDROID_TARGET_ARCH, armeabi-v7a)  {
             DEFINES += __androidArm32__
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -231,13 +231,6 @@ CONFIG(debug, debug|release) {
 
 SOURCE_DIR = $$IN_PWD
 
-!iOSBuild {
-    OBJECTS_DIR  = $${OUT_PWD}/obj
-    MOC_DIR      = $${OUT_PWD}/moc
-    UI_DIR       = $${OUT_PWD}/ui
-    RCC_DIR      = $${OUT_PWD}/rcc
-}
-
 LANGUAGE = C++
 
 LOCATION_PLUGIN_DESTDIR = $${OUT_PWD}/src/QtLocationPlugin
--- a/android/AndroidManifest.xml
+++ b/android/AndroidManifest.xml
@@ -7,11 +7,10 @@
                 <category android:name="android.intent.category.LAUNCHER"/>
                 <action android:name="android.hardware.usb.action.USB_DEVICE_ATTACHED"/>
                 <action android:name="android.hardware.usb.action.USB_DEVICE_DETACHED"/>
-                <action android:name="android.bluetooth.device.action.ACL_CONNECTED" />
-                <action android:name="android.bluetooth.device.action.ACL_DISCONNECTED" />
+                <action android:name="android.bluetooth.device.action.ACL_CONNECTED"/>
+                <action android:name="android.bluetooth.device.action.ACL_DISCONNECTED"/>
                 <action android:name="android.hardware.usb.action.USB_ACCESSORY_ATTACHED"/>
             </intent-filter>
-
             <meta-data android:resource="@xml/device_filter" android:name="android.hardware.usb.action.USB_DEVICE_ATTACHED"/>
             <meta-data android:resource="@xml/device_filter" android:name="android.hardware.usb.action.USB_DEVICE_DETACHED"/>
             <meta-data android:resource="@xml/device_filter" android:name="android.hardware.usb.action.USB_ACCESSORY_ATTACHED"/>
@@ -36,13 +35,11 @@
             <meta-data android:value="@string/ministro_needed_msg" android:name="android.app.ministro_needed_msg"/>
             <meta-data android:value="@string/fatal_error_msg" android:name="android.app.fatal_error_msg"/>
             <!--  Messages maps -->
-
             <!-- Splash screen -->
             <!--
             <meta-data android:name="android.app.splash_screen_drawable" android:resource="@drawable/logo"/>
             -->
             <!-- Splash screen -->
-
             <!-- Background running -->
             <!-- Warning: changing this value to true may cause unexpected crashes if the
                           application still try to draw after
@@ -52,7 +49,7 @@
             <!-- Background running -->
         </activity>
     </application>
-    <uses-sdk android:minSdkVersion="16" android:targetSdkVersion="28"/>
+    <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="28"/>
 
     <!-- Needed to keep working while 'asleep' -->
     <uses-permission android:name="android.permission.WAKE_LOCK"/>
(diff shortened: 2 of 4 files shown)


akontsevich:
> **P.S. BTW**, Why Android build on 5.15 have multiple link errors? Any ideas how to fix? Update AndroidManifest.xml, @patrickelectric : https://wiki.qt.io/Qt_for_Android_known_issues ?

No, fixing `AndroidManifest.xml` does not help. Will recheck building with 5.12.9.

akontsevich:
Full list of link errors: [link_errors.txt](https://github.com/mavlink/qgroundcontrol/files/5426303/link_errors.txt)

patrickelectric:
It appears that you are missing some android dependencies in general, but since I'm not an android developer myself I can't be of further help.

akontsevich:
https://github.com/mavlink/qgroundcontrol/pull/9136#issuecomment-714802202

> It appears that you are missing some android dependencies in general, but since I'm not an android developer myself I can't be of further help.

The problem in `stdlib` here: https://github.com/mavlink/qgroundcontrol/blob/master/QGCCommon.pri#L60. For 5.15 need to remove it and patch `AndroidManifest.xml`.

akontsevich:
> [#9136 (comment)](https://github.com/mavlink/qgroundcontrol/pull/9136#issuecomment-714802202)
> 
> > It appears that you are missing some android dependencies in general, but since I'm not an android developer myself I can't be of further help.
> 
> The problem in `stdlib` here: https://github.com/mavlink/qgroundcontrol/blob/master/QGCCommon.pri#L60. For 5.15 need to remove it...

@DonLakeFlyer I think this line could be removed for **5.12** as well as it was **ndk20** problem while currently Qt Creator installs and uses **ndk21** by default. I checked: builds fine with Qt 5.12 and ndk21, so removing `-nostdlib++` hack fix: https://github.com/mavlink/qgroundcontrol/pull/9136/commits/dc898635816757787001a9c61abc52dc868d0caa

akontsevich:
Another problem here building APK:
```
qmlimportscanner: No such file or directory: ".../Code/build-AirBoss-Android_Qt_5_15_1_Clang_Multi_Abi-Debug/rcc/arm64-v8a/qmake_qmake_qm_files.qrc"
Invalid json output from qmlimportscanner.
19:38:59: The process "~/Qt/5.15.1/android/bin/androiddeployqt" exited with code 4.
Error while building/deploying project qgroundcontrol (kit: Android Qt 5.15.1 Clang Multi-Abi)
When executing step "Build Android APK"
```
Seems need to regenerate templates: https://forum.qt.io/topic/110662/unable-to-build-project-on-android-with-qt-5-14-qmlimportscanner-no-such-file-or-directory/5 however not sure new templates will work with 5.12

akontsevich:
> Seems need to regenerate templates: https://forum.qt.io/topic/110662/unable-to-build-project-on-android-with-qt-5-14-qmlimportscanner-no-such-file-or-directory/5 however not sure new templates will work with 5.12

No luck for now, same errors.

akontsevich:
Fixed APK build error under 5.15 removing unnecessary assignments:  https://github.com/mavlink/qgroundcontrol/pull/9136/commits/54d2028e47a6b4a29f450dde42b56dd576f771f0 - builds fine with 5.12 and 5.15 now. For 5.15 probably need to alter `AndroidManifest.xml` as mentioned here: https://wiki.qt.io/Qt_for_Android_known_issues

akontsevich:
>  For 5.15 probably need to alter `AndroidManifest.xml` as mentioned here: https://wiki.qt.io/Qt_for_Android_known_issues

Yep, just checked it locally: need to update all the templates to deploy and run APK correctly. However they do not work with 5.12. Thoughts?

akontsevich:
> Yep, just checked it locally: need to update all the templates to deploy and run APK correctly. However they do not work with 5.12. Thoughts?

I can commit updated Android templates to current PR if You ok to start building daily Android builds with Qt 5.15? What do You think, @DonLakeFlyer, feasible? 

### Templates:

Screenshot_20201026_161810

DonLakeFlyer:
I"m not ready to move to 5.15 yet. If there is something needed which you know how to fix but for that which will break compat with 5.12 then you can put it in a separate pull. And when I'm ready to move I'll use it. I have a bunch of more important stuff stacked up in front of move to 5.15.1.

akontsevich:
I see, ok, I'll put that stuff into a separate pull. In current nothing breaks 5.12 compatibility so You may merge this.

DonLakeFlyer:
Awesome, thanks. I should have time to look through this today and merge.

QGCCommon.pri
```diff
@@ -57,7 +57,6 @@ linux {
             -Wno-parentheses-equality           # android gstreamer header files
         QMAKE_CFLAGS_WARN_ON += \
             -Wno-unused-command-line-argument   # from somewhere in Qt generated build files
-        QMAKE_LINK += -nostdlib++ # Hack fix?: https://forum.qt.io/topic/103713/error-cannot-find-lc-qt-5-12-android
```
DonLakeFlyer, line 60: Why remove this now?

QGCCommon.pri
```diff
@@ -232,13 +231,6 @@ CONFIG(debug, debug|release) {
 
 SOURCE_DIR = $$IN_PWD
 
-!iOSBuild {
```
DonLakeFlyer, line 235: Also why is this removed?

src/AnalyzeView/ULogParser.cc
```diff
@@ -14,25 +14,25 @@ ULogParser::~ULogParser()
 
 int ULogParser::sizeOfType(QString& typeName)
 {
-    if (typeName == QLatin1Literal("int8_t") || typeName == QLatin1Literal("uint8_t")) {
```
DonLakeFlyer, line 17: In most other parts of the code I've given up on stuff like this and just do:
```
if (typeName == "int8_t") || ty
```

It causes extra run time code to be run, but it has no real impact on performance and the readability of the code without it is way better. So I lean to readability as opposed to trying to reduce a performance hit you can't measure above microseconds.

QGCCommon.pri
akontsevich: Mentioned here: https://github.com/mavlink/qgroundcontrol/pull/9136#issuecomment-715374359 - not necessary for ndk21, if not removed - link errors under 5.15, either works fine with 5.12

QGCCommon.pri
akontsevich: See this comment: https://github.com/mavlink/qgroundcontrol/pull/9136#issuecomment-715662938 - these assignments are not necessary with either QT version (`qmake` does all automatically) and break multi-ABI build in Qt 5.15, see: [QTBUG-81477](https://bugreports.qt.io/browse/QTBUG-81477). I've spent 1 week searching and fixing all of these. :wink:

src/AnalyzeView/ULogParser.cc
akontsevich: I just replaced deprecated API to new one. I did not diagnose it is necessary or not. So remove `QLatin1Literal` completelly? Only in this file or in `APMAirframeComponentController.cc` as well?

DonLakeFlyer: It's all fine. Just letting you know for other changes you make. No need to go back and change these.

akontsevich: ok, thanks, do we merging this? :slightly_smiling_face:

QGCCommon.pri
DonLakeFlyer: If you don't have this what is the directory structure?

DonLakeFlyer: You also talk about multi-abi build. Are you talking about creating a package bundle which includes both 32 and 64 bit support? QGC doesn't support package bundles.  I'm not sure switching to it would be a positive thing. CI won't work with it either.

android/AndroidManifest.xml
```diff
@@ -52,7 +49,7 @@
             <!-- Background running -->
         </activity>
     </application>
-    <uses-sdk android:minSdkVersion="16" android:targetSdkVersion="28"/>
```
DonLakeFlyer, line 55: I think this is ok. Is 21 the first one that support 64 bit?

akontsevich: Seems it first which looses gcc support and starts clang buildings. Min version to build 5.15.

QGCCommon.pri
akontsevich: > If you don't have this what is the directory structure you end up with and does that directory structure screw up the final installer step and/or the CI scripts which push things to the store. That code reaches into various places in the build treee.

As I said `qmake` crates exact same directory structure automatically with either 5.12 or 5.15 so works fine.

> You also talk about multi-abi build. Are you talking about creating a package bundle which includes both 32 and 64 bit support? 

It is possible but not necessary. Currently it works as is with 5.12. With 5.15 I'm adding `ANDROID_ABIS="arm64-v8a"` as an argument to `qmake` and it builds only 64 bit abi bundle as before.

>QGC doesn't support package bundles. CI won't work with it either.

You do not need to change anything in CI or support package bundles. Just add `ANDROID_ABIS="arm64-v8a"` and Qt does all automatically - apk produced still.

DonLakeFlyer:
Ok, I'll merge and the follow through CI to the store to make sure there aren't problems there,

akontsevich:
> Ok, I'll merge and the follow through CI to the store to make sure there aren't problems there,

Thanks much!

akontsevich:
> I"m not ready to move to 5.15 yet. If there is something needed which you know how to fix but for that which will break compat with 5.12 then you can put it in a separate pull. And when I'm ready to move I'll use it. I have a bunch of more important stuff stacked up in front of move to 5.15.1. It takes a bunch of prep work to move to get CI to work.

Created PR as requested: https://github.com/mavlink/qgroundcontrol/pull/9154

PR #9143, opened 2020-10-28
replace non-ascii characters in source code

kemen209:
These non-ascii characters in source code will generate warning C4819 on Windows 10 with codepage set to  936, which in turn prevent from building the entire project.

NOTE: these non-ascii characters are found by using RegEx search with [^\x00-\x7F]


src/FactSystem/FactMetaData.cc
```diff
@@ -61,17 +61,17 @@ const FactMetaData::AppSettingsTranslation_s FactMetaData::_rgAppSettingsTransla
     { "cm/px",  "cm/px",    FactMetaData::UnitHorizontalDistance,    UnitsSettings::HorizontalDistanceUnitsMeters, FactMetaData::_defaultTranslator,                   FactMetaData::_defaultTranslator },
     { "m/s",    "m/s",      FactMetaData::UnitSpeed,                 UnitsSettings::SpeedUnitsMetersPerSecond,     FactMetaData::_defaultTranslator,                   FactMetaData::_defaultTranslator },
     { "C",      "C",        FactMetaData::UnitTemperature,           UnitsSettings::TemperatureUnitsCelsius,       FactMetaData::_defaultTranslator,                   FactMetaData::_defaultTranslator },
-    { "m^2",    "m\u00B2",  FactMetaData::UnitArea,                  UnitsSettings::AreaUnitsSquareMeters,         FactMetaData::_defaultTranslator,                   FactMetaData::_defaultTranslator },
```
DonLakeFlyer, line 64: Ah, interesting. I was wondering if this superscript format would work everywhere. Guess not.

DonLakeFlyer:
Thanks for the pull

PR #9144, opened 2020-10-28
fix issue of 'Error writing to QHostAddress("")'

kemen209:
Platform: Windows 10 x64 
qgroundcontrol version: master

Step to reproduce:
(1) Connect QGC to px4, wait until QGC show "Connected" on the UI
(2) Then, shutdown px4. And QGC start to show warning message "Error writing to QHostAddress("") 32762"

Error message in QGC console:
Screen Shot 2020-10-23 at 10 20 17
"


kemen209:
Debugging show that the readDatagram in UDPLink's readBytes return -1(no idea why even it passed condition checking of hasPendingDtagrams()), which lead to empty host address.


Some debugging print sentences added in the following:
Screen Shot 2020-10-27 at 09 38 01

Screen Shot 2020-10-26 at 18 05 30

src/comm/UDPLink.cc
```diff
@@ -182,7 +182,10 @@ void UDPLink::readBytes()
         QHostAddress sender;
         quint16 senderPort;
         //-- Note: This call is broken in Qt 5.9.3 on Windows. It always returns a blank sender and 0 for the port.
-        _socket->readDatagram(datagram.data(), datagram.size(), &sender, &senderPort);
```
DonLakeFlyer, line 185: Very strange. Can you include a comment on the `if -1` explaining why it is needed. I also assume the comment about 5.9.3 is no longer needed. Can you remove that?

kemen209: I print out the error message of the socket: "Connection reset by peer".

Screen Shot 2020-11-02 at 16 46 46

And the message "Connection reset by peer" will keep pop up until manually "Disconected" the link on QGC side.

Figure out the error code is WSAECONNRESET on Windows, as stated in [windows-sockets-error-code](https://docs.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2):

> An existing connection was forcibly closed by the remote host. This normally results if the peer application on the remote host is suddenly stopped, the host is rebooted, the host or remote network interface is disabled, or the remote host uses a hard close.

And in our scenario, the peer application which is PX4 is closed. I think it may not close the link properly(NOT verified). Since we can't control the behavior of other application, condition checking on our-side return value(-1) is required.

DonLakeFlyer: Ok, so comment should say something along those lines. "If the other end is reset then it will still report data available, but will fail on the readDatagram call". Something like that.

kemen209: adding comments as request.

DonLakeFlyer:
@kemen209 Thanks so much for figuring this all out.

PR #9145, opened 2020-10-28
fix warning C5208 on windows(stop the build process)

kemen209:
Building QGC with visual studio 2019 ver 16.6(with x86-windows-msvc2017-64bit)(not vs2017) on Windows 10, it stop with warning C5208.

As state in Microsoft doc(https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5208?view=vs-2019), the  fix is simple: add a name to struct typedef.

### DIFF
--- a/src/Vehicle/Vehicle.h
+++ b/src/Vehicle/Vehicle.h
@@ -1133,7 +1133,7 @@ private:
     void _waitForMavlinkMessageMessageReceived(const mavlink_message_t& message);
 
     // requestMessage handling
-    typedef struct {
+    typedef struct RequestMessageInfo {
         bool                        commandAckReceived; // We keep track of the ack/message being received since the order in which this will come in is random
         bool                        messageReceived;    // We only delete the allocated RequestMessageInfo_t when both happen (or the message wait times out)
         int                         msgId;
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
The concept that somehow a struct name is required for a typedef seems odd to me. I can't seem to understand what that does is saying with respect to why it is needed. I wondering whether a better change is to add an ignore for the warning. Not sure if that is dangerous or not. Otherwise it seems like this is just going to keep popping up since QGC is not currently running against VS 2019 as it's supported compiler.

kemen209:
As stated on Microsoft page, the Unnamed classes within a typedef declaration can't have any members other than:
non-static data members,
member classes,
member enumerations,
and default member initializers.

and the restriction is meant to ensure the simplicity of structs that have typedef names for linkage purposes. They must be simple enough that no linkage calculations are necessary before the compiler gets to the typedef name for linkage.

kemen209:
I'm had little knowledge about the build process, so i don't know what is the proper way to ignore the warning. I will leave that to someone else.

DonLakeFlyer:
Ok, fine. We'll just go with adding the names then.

PR #9150, opened 2020-10-29
Fix MISSION_ITEM_INT bugs

DonLakeFlyer:
Fix problem with incorrect futzing of x/y params.

### DIFF
--- a/src/MissionManager/PlanManager.cc
+++ b/src/MissionManager/PlanManager.cc
@@ -394,8 +394,8 @@ void PlanManager::_handleMissionItem(const mavlink_message_t& message, bool miss
         param2 =        missionItem.param2;
         param3 =        missionItem.param3;
         param4 =        missionItem.param4;
-        param5 =        (double)missionItem.x / qPow(10.0, 7.0);
-        param6 =        (double)missionItem.y / qPow(10.0, 7.0);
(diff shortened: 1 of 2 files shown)


src/MissionManager/PlanManager.cc
```diff
@@ -394,8 +394,8 @@ void PlanManager::_handleMissionItem(const mavlink_message_t& message, bool miss
         param2 =        missionItem.param2;
         param3 =        missionItem.param3;
         param4 =        missionItem.param4;
-        param5 =        (double)missionItem.x / qPow(10.0, 7.0);
-        param6 =        (double)missionItem.y / qPow(10.0, 7.0);
```
FlavioTonelli, line 398: @DonLakeFlyer looks like you used two time missionItem.x

DonLakeFlyer: Ah crap, thanks

DonLakeFlyer: I need to switch the unit tests to MISSION_ITEM_INT so they will catch this

DonLakeFlyer: @FlavioTonelli Thanks so much for catching this

FlavioTonelli: Just lucky that I scrolled into this pull request ;D

PR #9159, opened 2020-11-04
dist: Moving Windows files into separate directory

thopiekar:
Just to get the OS-specific files out of the root directory.

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com> (github: thopiekar)


DonLakeFlyer:
Seems ok. Will only screw custom build people that aren't overriding the icons correctly. Thanks

PR #9161, opened 2020-11-04
Linux: Move to Qt 5.15.1

DonLakeFlyer:
* First step of moving all OS to Qt 5.15.1 which will be the supported Qt for upcoming Stable 4.1
* Getting it in early enough to get some bake time on it while working towards Stable


DonLakeFlyer:
I'm giving up on this for 4.1 stable

PR #9163, opened 2020-11-05
QGCMAVLink: add missing class for Sub

Williangalvani:
This fixes features that check for a sub not working, such as the "lights" and "frame" tabs in vehicle setup


src/comm/QGCMAVLink.cc
```diff
@@ -113,6 +113,8 @@ QGCMAVLink::VehicleClass_t QGCMAVLink::vehicleClass(MAV_TYPE mavType)
     case MAV_TYPE_QUADROTOR:
     case MAV_TYPE_COAXIAL:
     case MAV_TYPE_HELICOPTER:
+    case MAV_TYPE_SUBMARINE:
```
DonLakeFlyer, line 116: This is right in the middle of the multirotor block. It's going to return submarine for the multi-rotor type above it.

Williangalvani: Oops!
I moved it under boat now.

Issue #7416, opened 2019-05-04
(Android)Initial Time Value Issue in Flight Plan Tab


AGurel:
For android devices, in Flight Plan tab, initial time value is 01:00:00. However in other devices no problem.

I solve it as below;

In src/PlanView/PlanToolBar.qml file, line 78, i added the bold section. In debug mode, i saw that _missionTime's value is NaN, but after click Flight Plan tab the _missionTime's value is 0.

function getMissionTime() {
 if(isNaN(_missionTime) **|| _missionTime == 0**) {
 return "00:00:00"
 }
 var t = new Date(0, 0, 0, 0, 0, Number(_missionTime))
 return Qt.formatTime(t, 'hh:mm:ss')
 }

![original](https://user-images.githubusercontent.com/7662882/57179885-8016d400-6e8b-11e9-9ce9-b3f2b27f1e84.jpeg)
![debuged](https://user-images.githubusercontent.com/7662882/57179886-8016d400-6e8b-11e9-8edc-3722a46e707f.jpeg)

DonLakeFlyer:
Sorry I lost track of this bug. I installed both stable and daily and they both worked fine. From the screenshots It looks like you are running an old build.

DonLakeFlyer:
I'll reopen if you can make this happen on a new daily.

DonLakeFlyer:
@AGurel Can you take a screenshot with the Vehicle Info section of Mission Start expanded?

AGurel:
> @AGurel Can you take a screenshot with the Vehicle Info section of Mission Start expanded?

ok. you can see it below.

![Screenshot_20210107-110838](https://user-images.githubusercontent.com/7662882/103868014-576fc480-50d9-11eb-932f-e785df36e190.jpg)

PR #9166, opened 2020-11-07
Update PlanToolBarIndicators.qml

AGurel:
For android devices, in Flight Plan tab, initial time value is 01:00:00. However in other devices no problem. In debug mode, i saw that _missionTime's value is NaN, but after click Flight Plan tab the _missionTime's value is 0.

I reported this as an issue before. https://github.com/mavlink/qgroundcontrol/issues/7416#issue-440324533


DonLakeFlyer:
Bug doesn't reproduce

DonLakeFlyer:
Closing until I get possibly more details in #7416. I'll reopen if valid.

PR #9175, opened 2020-11-13
ParameterEditor: mention that it resets sensors calibration and frame setup

Williangalvani:
We had users pressing that and hoping the vehicle would still be operational. This makes it a little clearer that it won't.


DonLakeFlyer:
This is pretty specific wording to only a few cases of what gets reset. Something along the lines of "all vehicle settings, setup and calibrations" would be way better.

Williangalvani:
Yes, that is way better. Updated.

DonLakeFlyer:
Perfect, thanks

PR #9178, opened 2020-11-18
Fix map scale behaviour

patrickelectric:
- Fix scale showing in video when map is in windowState
- Show scale when map is in windowState

old version:
Screen Capture_QGroundControl_20201118152907

new version:
Screen Capture_QGroundControl_20201118152721

### DIFF
--- a/src/FlightDisplay/FlyViewMap.qml
+++ b/src/FlightDisplay/FlyViewMap.qml
@@ -49,6 +49,7 @@ FlightMap {
     property var    _rallyPointController:      planMasterController.rallyPointController
     property var    _activeVehicleCoordinate:   _activeVehicle ? _activeVehicle.coordinate : QtPositioning.coordinate()
     property real   _toolButtonTopMargin:       parent.height - mainWindow.height + (ScreenTools.defaultFontPixelHeight / 2)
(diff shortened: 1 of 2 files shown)


patrickelectric:
Ping @DonLakeFlyer

src/FlightDisplay/FlyViewMap.qml
```diff
@@ -572,4 +573,16 @@ FlightMap {
         }
     }
 
+    MapScale {
```
DonLakeFlyer, line 576: Adding another MapScale control at this level of the ui hierarchy screws up the layering of ui elements for custom build overrides. Why can't this be fixed just in FlyViewWidgetLayer?

patrickelectric: Because the map is a component itself when you popup in a new window. 
Are you suggesting to not have the scale after you popup ?

DonLakeFlyer: Hmm, ok right. I didn't realize the widgets are not shown at all in this mode. In fact I had no idea you could even do this! Although duplicating the MapScale makes my skin crawl a bit it's the simplest way to fix this for now.

PR #9180, opened 2020-11-19
improved README on UDP Pipeline of VideoReceiver s.t. it is closer to…

lukelu520:
… the implemented pipeline of the source code

After reviewing the source code of [GstVideoReceiver.cc](https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoReceiver/GstVideoReceiver.cc), we found that below UDP pipeline illustrates the implementation better.

`gst-launch-1.0 udpsrc port=5600 caps='application/x-rtp, media=(string)video, clock-rate=(int)90000, encoding-name=(string)H264' ! rtpjitterbuffer ! parsebin ! decodebin ! autovideosink fps-update-interval=1000 sync=false
`

So we propose to add the above pipeline in the README, allowing more people (especially new developers who are not familiar with gstreamer) get the big picture of QGC's video stream.

Best,
-- Luke


DonLakeFlyer:
@jaxxzer @dakejahl @andrewvoznytsa Can any video folks comment on this? It's all greek to me.

andrewvoznytsa:
Technically that change is correct. From the other side it does not have big sense because if we are about to receive UDP@H.264 stream the we explicitly inform udpsrc that encoding-name=(string)H264 (which means H.264 video) and therefore we don't need flexibility/automation in choosing parser (parsebin) and decoder (decodebin) - we already know that for H.264 stream we need h264parse and avdec_h264 (and it is what parsebin and decodebin will use after some probing). In this simple case parsebin and decodebin are overkill and are just hiding processing mechanism. I prefer to keep it clear for everyone and show them which exact elements are needed to handle this video stream. Having exactly specified pipeline (no heuristics) also is helpful with debugging. But it is up to maintainers to decide what they'd see in the docs.

dakejahl:
you could also mention removing rtpjitterbuffer will reduce video latency, since that is what low latency mode is doing

DonLakeFlyer:
I have no context to determine the correctness of this. So hoping ArduSub/Sentera folks who are know heavy video users can determine what is correct here.

lukelu0520:
@dakejahl , that's a very good comment, I've appended in this PR.
@andrewvoznytsa @DonLakeFlyer , thanks for your comments. The purpose of the PR is providing the second way to receive udp video stream, which is more general and tells more about QGC's VideoReceiver. One of functions of README is to let people understand the big picture.  

In the README, we've listed 2+ ways on the sending side, why not having two ways on the receiving side ? (one is in specific, another one is in general ).

DonLakeFlyer:
@dakejahl Jake, you're deep into this. Let me know when you are happy with where this stands and I'll merge.

dakejahl:
Yes this is representative of the pipeline. This is the kind of example you'd find on stack overflow when looking for gstreamer pipeline examples so I think it is really helpful to have in the README

DonLakeFlyer:
It seems like some more explanation other than "or this one" would be helpful. I don't quite understand why there needs to be two examples. If it makes sense to have two then can there be more details on the differences between the two?

mrpollo:
IMHO none of the examples tell you anything if you aren't already familiar with gstreamer. As a developer, I appreciate the diversity of examples I can try. Still, ultimately it would be better if someone would take the time to explain in detail what a gstreamer pipeline is, what the components are, the requirements, how to debug a pipeline, and how to plan for building yours.

Perhaps we should have a quick graph explaining what the pipeline is doing similar to this:

sample pipeline

Or maybe something like this

another sample pipeline

I would also clearly separate the example pipelines from the installation instructions by making title headers for each.

Overall, I think this section of the docs needs much improvement that would fall out of the scope of the authors' original intention.

My proposal to the author is to keep the script as-is, and include a quick paragraph explaining what the pipeline is doing, and how its different from the rest of the examples, keeping in mind non-gstreamer capable developers will be reading the documentation, and if you have time to throw in a diagram, even better, but I'm not going to push my luck!

DonLakeFlyer:
That level of detail isn't really the job of QGC docs. The is the job of gstreamer docs. That's kind of like requiring QGC docs to include how to use Qt or how to write C++ code.

On Mar 18, 2021, at 2:25 PM, Ramon Roche ***@***.******@***.***>> wrote:



IMHO none of the examples tell you anything if you aren't already familiar with gstreamer. As a developer, I appreciate the diversity of examples I can try. Still, ultimately it would be better if someone would take the time to explain in detail what a gstreamer pipeline is, what the components are, the requirements, how to debug a pipeline, and how to plan for building yours.

Perhaps we should have a quick graph explaining what the pipeline is doing similar to this:

[sample pipeline]<https://camo.githubusercontent.com/3d179729bf5688d42c354221bf39b5156cbddbf8f9d7f6204b7ad62fd2e6ee3d/68747470733a2f2f75706c6f61642e77696b696d656469612e6f72672f77696b6970656469612f636f6d6d6f6e732f612f61352f4753747265616d65725f6578616d706c655f706970656c696e652e737667>

Or maybe something like this

[another sample pipeline]<https://camo.githubusercontent.com/543163a2dc685462e361d92af9f1814989dabba464f604017742d18e56f94405/68747470733a2f2f6773747265616d65722e667265656465736b746f702e6f72672f646f63756d656e746174696f6e2f6170706c69636174696f6e2d646576656c6f706d656e742f696e74726f64756374696f6e2f696d616765732f73696d706c652d706c617965722e706e67>

I would also clearly separate the example pipelines from the installation instructions by making title headers for each.

Overall, I think this section of the docs needs much improvement that would fall out of the scope of the authors' original intention.

My proposal to the author is to keep the script as-is, and include a quick paragraph explaining what the pipeline is doing, and how its different from the rest of the examples, keeping in mind non-gstreamer capable developers will be reading the documentation, and if you have time to throw in a diagram, even better, but I'm not going to push my luck!

cc @hamishwillee<https://github.com/hamishwillee> for awareness

mrpollo:
It's not at all comparable, and I would argue that having a few sections on how to best leverage Qt would even help this project gain more adopters and unblock newcomers, same with gstreamer and any other dependencies. 

I'm not arguing about having a "how to get started with gstreamer" guide, just a basic breakdown of what QGC is expecting from your pipeline, broken down into pieces that are easily digestible and give you enough to continue on your own if you understand gstreamer or not, by allowing you to know what to search for, and where the pieces fit.

Ultimately you are the maintainer, you are free to decide what you want on your docs. I'm not going to push more on this matter.

DonLakeFlyer:
It's more about what QGC supports or doesn't as far as video types/sources as opposed to understanding gstreamer pipelines. You don't need to understand anything about gstreamer pipelines necessarily to pump video into QGC. It's just a video source which can come from anywhere/anything. It has nothing really to do with gstreamer on the video provider side. That most likely is just some piece of hardware which is spitting out video. There may be some gstreamer pipeline in the middle of it, there may not be. It's more of a video spec problem. Some things work some don't. The pipeline examples for creating video samples is in essence just a way to encompass that video spec in a testable way without real hardware.

DonLakeFlyer:
> and I would argue that having a few sections on how to best leverage Qt would even help this project gain more adopters and unblock newcomers

The existing dev guide talks about things like the architecture and things like Qt QML which are the main developer tool. Sure it's sparse and could always be 100 times larger with more information but it is what it is.

DonLakeFlyer:
True video devs are welcome to correct me since I am way far away from knowing all about video. I just know enough to be dangerous.

DonLakeFlyer:
Better than nothing I guess.

PR #9199, opened 2020-11-22
Take MulticastLock on Android to ensure broadcast packets can be received

### LINKED ISSUE #8365: UDP broadcast not working on Android
When running ardupilot with the ":bcast" flag, it sends the mavlink data to the broadcast address and when the first "client" responds it locks down to the address that responded. This works fine when running QGC on my Linux machine, but it's not working on Android.

I think this is related to this QT issue https://bugreports.qt.io/browse/QTBUG-73138 and the workaround is to acquire  WifiManager.MulticastLock (it's a global lock, so any app acquiring this will make other apps to work).

From @DonLakeFlyer:

    Hmm, maybe. I know it used to work. But may have broken with change to 5.11. Enter an issue and I'll try to reproduce.

brad112358:
Workaround for QTBUG-73138
fixes #8365

### DIFF
--- a/android/src/org/mavlink/qgroundcontrol/QGCActivity.java
+++ b/android/src/org/mavlink/qgroundcontrol/QGCActivity.java
@@ -77,7 +77,7 @@ public class QGCActivity extends QtActivity
     private static PendingIntent                        _usbPermissionIntent = null;
     private TaiSync                                     taiSync = null;
     private Timer                                       probeAccessoriesTimer = null;
-    private static WifiManager.MulticastLock            _cLock;
+    private static WifiManager.MulticastLock            _wifiMulticastLock;
     
     public static Context m_context;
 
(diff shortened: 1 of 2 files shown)


android/src/org/mavlink/qgroundcontrol/QGCActivity.java
```diff
@@ -76,7 +77,8 @@
     private static PendingIntent                        _usbPermissionIntent = null;
     private TaiSync                                     taiSync = null;
     private Timer                                       probeAccessoriesTimer = null;
-
+    private static WifiManager.MulticastLock            _cLock;
```
DonLakeFlyer, line 80: Could you change this to something a bit more descriptive like '_wifiMulticastLog' or something. If you just see '_cLock' further down the code its not clear at a glance what it is.

DonLakeFlyer:
Perfect thanks so much

PR #9201, opened 2020-11-22
Virtual joystick defaults to auto-center

DonLakeFlyer:
* This is a safer option for multi-rotor/rover/sub. It's isn't necessarily the best option for fixed wing, but I'd question the sanity of someone flying a fixed wing with their thumbs anyway. They can always turn it off.
* There is a problem with respect to arming the vehicle with auto-center which are firmware specific:
  * PX4: Allows you arm with the throttle centered in modes like Position. It does not allow it in Stabilize. To me this makes a lot of sense since most joysticks virtual or not auto-center both axes.
  * ArduPilot: Require throttle to be low no matter what if you try to arm. So for vehicles which don't support reverse throttle auto-center doesn't really work. The reason being you can't hold the virtual joystick throttle down and click in the toolbar to arm. I'm going to post an issue against ArduPilot asking for ability to arm in modes where a center throttle is a perfectly valid state to fly and start a takeoff with.
* Related to #5897

### DIFF
--- a/src/QmlControls/JoystickThumbPad.qml
+++ b/src/QmlControls/JoystickThumbPad.qml
@@ -20,6 +20,9 @@ Item {
     property bool   _processTouchPoints:    false
     property color  _fgColor:               QGroundControl.globalPalette.text
     property color  _bgColor:               QGroundControl.globalPalette.window
+    property real   _hatWidth:              ScreenTools.defaultFontPixelHeight
+    property real   _hatWidthHalf:          _hatWidth / 2
+
     property real   stickPositionX:         _centerXY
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@jaxxzer ArduSub folks can you test this with your stuff to make sure I didn't screw you up.

patrickelectric:
ping @Williangalvani

Williangalvani:
Well, it turned out that it was not working before this pr anyway...
I opened https://github.com/mavlink/qgroundcontrol/pull/9204 to fix it.

I tested #9204 over this pr and that worked well.

DonLakeFlyer:
Thanks folks

PR #9204, opened 2020-11-23
VirtualJoystick: remove sub from yAxisPositiveRangeOnly condition

Williangalvani:
Sub currently handles throttle in the 0-1000 range
https://github.com/ArduPilot/ardupilot/issues/8818


DonLakeFlyer:
Do regular joysticks need to same change?

Williangalvani:
> Do regular joysticks need to same change?

Regular joysticks are working fine on master and stable

Issue #9202, opened 2020-11-23
QGC window opening at saved screen coordinates even if they become invalid


vincentdavoust:
## Expected Behavior
QGC window always opens on visible screen even when displays are changed / removed

## Current Behavior
QGC is launched properly but is not displayed. (clicking on the icon loses focus on other windows but does not display, using windows-tab shows window miniature but cannot switch it to visible on screen)

## Steps to Reproduce:
1. Have dual display
2. Open QGC window and position it on screen 2 (the display on the right, could be screen 1 if main screen is switched)
3. Close QGC
4. Disconnect screen 2
5. Relaunch QGC

## System Information
When posting bug reports, include the following information
- Operating System: Windows 10
- QGC Version: 4.0.11
- QGC build: stable

## Detailed Description
I believe the issue comes from using the saved window coordinates in `%AppData%/QGroundControl/QGroundControl.ini [MainWindowState]` without checking that they are within 
`0 <= x < QApplication::desktop()->screenGeometry().width() - width`
and
`0 <= y < QApplication::desktop()->screenGeometry().height() - height`
and also probably constraining width and height to the full display size as a maximum.

vincentdavoust:
I am planning to make a fix/PR for this shortly, except if this is already something that is being handled ?
Also any input on where to handle this the cleanest would be appreciated.

DonLakeFlyer:
This ends up being a Qt thing. I think folks have tried fixing it in the past and failed. I believe because there was some Qt support missing to make it work. That was a while back though. It's possible Qt 5.12 has what is needed to make this work.

vincentdavoust:
I think I've fixed it with an `if` in MainWindowSavedState.qml, making th PR now.

PR #9212, opened 2020-11-25
#9202 Bugfix : Only restoring saved window state if parameters make sense on current screen configuration

vincentdavoust:
`src/QmlControls/MainWindowSavedState.qml` used to restore window state if it finds saved parameters in `%AppData%/QGroundControl.org/QGroundControl.ini`
This leads to bugs if, for example, the saved window position is on a screen which is no longer connected.

This PR adds a simple `if` clause which lets the main window open with Qt default geometry in case the saved geometry is found to be invalid.

### DIFF
--- a/src/QmlControls/MainWindowSavedState.qml
+++ b/src/QmlControls/MainWindowSavedState.qml
@@ -32,12 +32,24 @@ Item {
     }
 
     Component.onCompleted: {
-        if (_enabled && s.width && s.height) {
-            window.x = s.x;
-            window.y = s.y;
-            window.width = s.width;
-            window.height = s.height;
-            window.visibility = s.visibility;
+        var isValidScreenGeometry =
+                (
+                    (s.x      >= 0                                         ) &&
+                    (s.y      >= 0                                         ) &&
+                    (s.width  >  0                                         ) &&
+                    (s.height >  0                                         ) &&
+                    (s.x      <=  Screen.desktopAvailableWidth  - s.width  ) &&
+                    (s.y      <=  Screen.desktopAvailableHeight - s.height ) &&
+                    (s.width  <=  Screen.desktopAvailableWidth             ) &&
+                    (s.height <=  Screen.desktopAvailableHeight            )
+                )
+
+        if (_enabled && isValidScreenGeometry) {
+            window.x            = s.x
+            window.y            = s.y
+            window.width        = s.width
+            window.height       = s.height
+            window.visibility   = s.visibility
         }
     }
 


src/QmlControls/MainWindowSavedState.qml
```diff
@@ -32,12 +33,24 @@ Item {
     }
 
     Component.onCompleted: {
-        if (_enabled && s.width && s.height) {
-            window.x = s.x;
-            window.y = s.y;
-            window.width = s.width;
-            window.height = s.height;
-            window.visibility = s.visibility;
+        _isValidScreenGeometry = (
```
DonLakeFlyer, line 36: Can you change this to:
```
        _isValidScreenGeometry = (
                                     (s.x      >= 0                                         ) &&
                                     (s.y      >= 0                                         ) &&
                                     (s.width  >  0                                         ) &&
                                     (s.height >  0                                         ) &&
                                     (s.x      <=  Screen.desktopAvailableWidth  - s.width  ) &&
                                     (s.y      <=  Screen.desktopAvailableHeight - s.height ) &&
                                     (s.width  <=  Screen.desktopAvailableWidth             ) &&
                                     (s.height <=  Screen.desktopAvailableHeight            ) &&
                                  )
```

To remove redundancy and for better grouping for readability.

src/QmlControls/MainWindowSavedState.qml
```diff
@@ -19,6 +19,7 @@ Item {
     property Window window
 
     property bool _enabled: !ScreenTools.isMobile && QGroundControl.corePlugin.options.enableSaveMainWindowPosition
+    property bool _isValidScreenGeometry : true
```
DonLakeFlyer, line 22: Why does this need to be a property? Should just be a `var` in `onCompleted`.

DonLakeFlyer:
@vincentdavoust So still two things:
* The _isValidScreenGeometry should be just isValidScreenGeometry. The '_' coding style is used to indicate a private property in Qml and/or private/protected class method/variable in C++.
* Once all done can you reorganize it all into a single commit

vincentdavoust:
@DonLakeFlyer What do you think of making `isValidScreenGeometry` a function ? Would that be considered even beter style for qml, or not necessarily ?
```
        function isValidScreenGeometry() {
            return (
                    (s.x      >= 0                                         ) &&
                    (s.y      >= 0                                         ) &&
                    (s.width  >  0                                         ) &&
                    (s.height >  0                                         ) &&
                    (s.x      <=  Screen.desktopAvailableWidth  - s.width  ) &&
                    (s.y      <=  Screen.desktopAvailableHeight - s.height ) &&
                    (s.width  <=  Screen.desktopAvailableWidth             ) &&
                    (s.height <=  Screen.desktopAvailableHeight            )
                   )
        }
        if (_enabled && isValidScreenGeometry()) {
            window.x            = s.x
            window.y            = s.y
            window.width        = s.width
            window.height       = s.height
            window.visibility   = s.visibility
        }
```

DonLakeFlyer:
Since the valid screen geometry check is only ever used in one place and the Component.onCompleted implementation is pretty short, leaving it inline is a better choice from a readability standpoint.

DonLakeFlyer:
@vincentdavoust Perfect, thanks

DonLakeFlyer:
@vincentdavoust Didn't realize this was on Stable. I'm not pushing out new Stables right now since 4.1 is working its way out the door. Can you bring to master as well?

PR #9099, opened 2020-10-05
Upgrade gstreamer to 1.18.1

jaxxzer:

PR #9213, opened 2020-11-25
Enable and configure hardware decode priority 

### LINKED ISSUE #8631: UDP H.264 was way better on 3.5.6, with quality and latency has degraded significantly.
QGC 3.5.6 vs. QGC 4.x
----

System:

MacBook Pro 2018 OSX 10.15.3
and
Acer Nitro 5, with Nvidia 1650GPU, Ubuntu 18.04

both got same issue...

## Expected Behavior
UDP H.264 without gray images or any error, very little latency.

## Current Behavior
UDP H.264 very long latency, often error in image
on the old 3.5.6. perfect....

## Steps to Reproduce:
1.
2.
3.
4.

## System Information
Play and run...  gStreamer created video stream with speed-preset=superfast 

## Detailed Description
gst-launch-1.0 videotestsrc ! video/x-raw,framerate=20/1 ! videoscale ! videoconvert ! x264enc tune=zerolatency bitrate=3000 speed-preset=superfast ! rtph264pay ! udpsink host=10.0.0.100 port=5600

patrickelectric:
- nvidia's hardwre code - tested by me
- vaapi hardware decode AMD - tested by @Williangalvani 

The older video implementation was using a specific pipeline, with the new video backend, now QGC is hostage of `decodebin` to take the decisions about the pipeline construction. 
Moving to `decodebin3` and enabling the hardware decode creates a much better scenario with a modern `decodebin` alternative that can handle a better construction and decision about the pipeline and elements that is used.  

Fix #8631
Needs #9099 to work under windows

### DIFF
--- a/src/FactSystem/FactMetaData.h
+++ b/src/FactSystem/FactMetaData.h
@@ -148,6 +148,9 @@ public:
     /// Used to add new values to the enum lists after the meta data has been loaded
     void addEnumInfo(const QString& name, const QVariant& value);
 
+    /// Used to remove values from the enum lists after the meta data has been loaded
+    void removeEnumInfo(const QVariant& value);
+
     void setDecimalPlaces           (int decimalPlaces)                 { _decimalPlaces = decimalPlaces; }
     void setRawDefaultValue         (const QVariant& rawDefaultValue);
     void setBitmaskInfo             (const QStringList& strings, const QVariantList& values);
--- a/src/Settings/Video.SettingsGroup.json
+++ b/src/Settings/Video.SettingsGroup.json
@@ -133,8 +133,10 @@
     "name":             "forceVideoDecoder",
     "shortDesc":        "Force specific category of video decode",
     "longDesc":         "Force the change of prioritization between video decode methods, allowing the user to force some video hardware decode plugins if necessary.",
-    "type":             "string",
-    "default":          "",
+    "type":             "uint32",
+    "enumStrings":      "Default,Force software decoder,Force NVIDIA decoder,Force VA-API decoder,Force DirectX3D 11 decoder",
+    "enumValues":       "0,1,2,3,4",
+    "default":           0,
     "qgcRebootRequired": true
 }
 ]
--- a/src/Settings/Video.SettingsGroup.json
+++ b/src/Settings/Video.SettingsGroup.json
@@ -128,6 +128,14 @@
     "longDesc":  "If this option is enabled, the rtpjitterbuffer is removed and the video sink is set to assynchronous mode, reducing the latency by about 200 ms.",
     "type":             "bool",
     "default":     false
+},
+{
+    "name":             "forceVideoDecoder",
+    "shortDesc":        "Force specific category of video decode",
+    "longDesc":         "Force the change of prioritization between video decode methods, allowing the user to force some video hardware decode plugins if necessary.",
+    "type":             "string",
+    "default":          "",
+    "qgcRebootRequired": true
 }
 ]
 }
--- a/src/Settings/VideoSettings.cc
+++ b/src/Settings/VideoSettings.cc
@@ -29,6 +29,12 @@ const char* VideoSettings::videoSourceMPEGTS            = "MPEG-TS (h.264) Video
 const char* VideoSettings::videoSource3DRSolo           = "3DR Solo (requires restart)";
 const char* VideoSettings::videoSourceParrotDiscovery   = "Parrot Discovery";
 
+const char* VideoSettings::forceVideoDecoderDefault = "Default";
+const char* VideoSettings::forceVideoDecoderSoftware = "Force software decoder";
+const char* VideoSettings::forceVideoDecoderNVIDIA = "Force NVIDIA decoder";
+const char* VideoSettings::forceVideoDecoderVAAPI = "Force VA-API decoder";
+const char* VideoSettings::forceVideoDecoderD3D11 = "Force DirectX3D 11 decoder";
+
 DECLARE_SETTINGGROUP(Video, "Video")
 {
     qmlRegisterUncreatableType<VideoSettings>("QGroundControl.SettingsManager", 1, 0, "VideoSettings", "Reference only");
@@ -63,6 +69,24 @@ DECLARE_SETTINGGROUP(Video, "Video")
         videoSourceVarList.append(QVariant::fromValue(videoSource));
     }
     _nameToMetaDataMap[videoSourceName]->setEnumInfo(videoSourceList, videoSourceVarList);
+
+    QStringList forceVideoDecoderList{
+        forceVideoDecoderDefault,
+        forceVideoDecoderSoftware,
+        forceVideoDecoderNVIDIA,
+#ifdef Q_OS_LINUX
+        forceVideoDecoderVAAPI,
+#endif
+#ifdef Q_OS_WIN
+        forceVideoDecoderD3D11,
+#endif
+    };
+    QVariantList forceVideoDecoderVarList;
+    for (const QString& forceVideoDecode: forceVideoDecoderList) {
+        forceVideoDecoderVarList.append(QVariant::fromValue(forceVideoDecode));
+    }
+    _nameToMetaDataMap[forceVideoDecoderName]->setEnumInfo(forceVideoDecoderList, forceVideoDecoderVarList);
+
     // Set default value for videoSource
     _setDefaults();
 }
@@ -74,6 +98,8 @@ void VideoSettings::_setDefaults()
     } else {
         _nameToMetaDataMap[videoSourceName]->setRawDefaultValue(videoDisabled);
     }
+
+    _nameToMetaDataMap[forceVideoDecoderName]->setRawDefaultValue(forceVideoDecoderDefault);
 }
 
 DECLARE_SETTINGSFACT(VideoSettings, aspectRatio)
@@ -103,6 +129,28 @@ DECLARE_SETTINGSFACT_NO_FUNC(VideoSettings, videoSource)
         connect(_videoSourceFact, &Fact::valueChanged, this, &VideoSettings::_configChanged);
     }
     return _videoSourceFact;
+}
+
+DECLARE_SETTINGSFACT_NO_FUNC(VideoSettings, forceVideoDecoder)
+{
+    if (!_forceVideoDecoderFact) {
+        _forceVideoDecoderFact = _createSettingsFact(forceVideoDecoderName);
+
+        _forceVideoDecoderFact->setVisible(
+#ifdef Q_OS_LINUX
+            true
+#else
+#ifdef Q_OS_WIN
+            true
+#else
+            false
+#endif
+#endif
+        );
+
+        connect(_forceVideoDecoderFact, &Fact::valueChanged, this, &VideoSettings::_configChanged);
+    }
+    return _forceVideoDecoderFact;
 }
 
 DECLARE_SETTINGSFACT_NO_FUNC(VideoSettings, udpPort)
--- a/src/Settings/VideoSettings.cc
+++ b/src/Settings/VideoSettings.cc
@@ -29,12 +29,6 @@ const char* VideoSettings::videoSourceMPEGTS            = "MPEG-TS (h.264) Video
 const char* VideoSettings::videoSource3DRSolo           = "3DR Solo (requires restart)";
 const char* VideoSettings::videoSourceParrotDiscovery   = "Parrot Discovery";
 
-const char* VideoSettings::forceVideoDecoderDefault = "Default";
-const char* VideoSettings::forceVideoDecoderSoftware = "Force software decoder";
-const char* VideoSettings::forceVideoDecoderNVIDIA = "Force NVIDIA decoder";
-const char* VideoSettings::forceVideoDecoderVAAPI = "Force VA-API decoder";
-const char* VideoSettings::forceVideoDecoderD3D11 = "Force DirectX3D 11 decoder";
-
 DECLARE_SETTINGGROUP(Video, "Video")
 {
     qmlRegisterUncreatableType<VideoSettings>("QGroundControl.SettingsManager", 1, 0, "VideoSettings", "Reference only");
@@ -70,22 +64,18 @@ DECLARE_SETTINGGROUP(Video, "Video")
     }
     _nameToMetaDataMap[videoSourceName]->setEnumInfo(videoSourceList, videoSourceVarList);
 
-    QStringList forceVideoDecoderList{
-        forceVideoDecoderDefault,
-        forceVideoDecoderSoftware,
-        forceVideoDecoderNVIDIA,
+    const QVariantList removeForceVideoDecodeList{
 #ifdef Q_OS_LINUX
-        forceVideoDecoderVAAPI,
+        VideoDecoderOptions::ForceVideoDecoderDirectX3D,
 #endif
 #ifdef Q_OS_WIN
-        forceVideoDecoderD3D11,
+        VideoDecoderOptions::ForceVideoDecoderVAAPI,
 #endif
     };
-    QVariantList forceVideoDecoderVarList;
-    for (const QString& forceVideoDecode: forceVideoDecoderList) {
-        forceVideoDecoderVarList.append(QVariant::fromValue(forceVideoDecode));
+
+    for(const auto& value : removeForceVideoDecodeList) {
+        _nameToMetaDataMap[forceVideoDecoderName]->removeEnumInfo(value);
     }
-    _nameToMetaDataMap[forceVideoDecoderName]->setEnumInfo(forceVideoDecoderList, forceVideoDecoderVarList);
 
     // Set default value for videoSource
     _setDefaults();
@@ -98,8 +88,6 @@ void VideoSettings::_setDefaults()
     } else {
         _nameToMetaDataMap[videoSourceName]->setRawDefaultValue(videoDisabled);
     }
-
-    _nameToMetaDataMap[forceVideoDecoderName]->setRawDefaultValue(forceVideoDecoderDefault);
 }
 
 DECLARE_SETTINGSFACT(VideoSettings, aspectRatio)
--- a/src/Settings/VideoSettings.h
+++ b/src/Settings/VideoSettings.h
@@ -35,6 +35,7 @@ public:
     DEFINE_SETTINGFACT(streamEnabled)
     DEFINE_SETTINGFACT(disableWhenDisarmed)
     DEFINE_SETTINGFACT(lowLatencyMode)
+    DEFINE_SETTINGFACT(forceVideoDecoder)
 
     Q_PROPERTY(bool     streamConfigured        READ streamConfigured       NOTIFY streamConfiguredChanged)
     Q_PROPERTY(QString  rtspVideoSource         READ rtspVideoSource        CONSTANT)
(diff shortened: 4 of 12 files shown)


patrickelectric:
The PR also follows @andrewvoznytsa comment about forcing hardware priority as an optional feature.

src/VideoReceiver/GstVideoReceiver.h
```diff
@@ -103,7 +103,7 @@ protected slots:
 
 protected:
     virtual GstElement* _makeSource(const QString& uri);
-    virtual GstElement* _makeDecoder(GstCaps* caps, GstElement* videoSink);
+    virtual GstElement* _makeDecoder();
```
andrewvoznytsa, line 106: You changed class API in a way that breaks derived classes. The fact that you don't see them does not mean that they do no exist - be careful with `virtual`.

Perhaps it would be useful to look into a year old discussion regarding video refactoring - it explains a lot.

patrickelectric: Ok, I see what are you talking about, I'll change it back.

src/VideoReceiver/GStreamer.cc
```diff
@@ -133,18 +133,22 @@ blacklist()
     changeRank("avdec_h264", GST_RANK_MARGINAL + 1);
```
andrewvoznytsa, line 133: That's not correct and breaks normal user experience.

src/VideoReceiver/GstVideoReceiver.h
andrewvoznytsa: > Moving to decodebin3 and enabling the hardware decode creates a much better scenario with a modern decodebin alternative that can handle a better construction and decision about the pipeline and elements that is used.

@patrickelectric can you please show better construction and decision about the pipeline and elements that are used?

Also, are you sure that decodebin3 works on Android and iOS (in the way how do you use it)? I recall that these platforms were not so straightforward.

patrickelectric: decodebin failed to use nvidia hardware decoders and create the correct filtercaps for glupload
I also got similar problems with some vaapi decoders. 

I have not tested with android yet.

DonLakeFlyer:
Having to have settings in video to support specific video hardware seems crazy. Why doesn't gstreamer work correctly in the first place? Seems like a pretty fundamental flaw/bug in gstreamer that it doesn't use hardware decoding when it is actually available.

Poking around here: https://gstreamer.freedesktop.org/documentation/tutorials/playback/hardware-accelerated-video-decoding.html?gi-language=c says this: "The GStreamer developers often rank hardware decoders lower than the software ones when they are defective. This should act as a warning."

src/VideoReceiver/GStreamer.cc
```diff
@@ -108,17 +108,37 @@ static void qgcputenv(const QString& key, const QString& root, const QString& pa
...
+    // Enable VAAPI drivers
+    for(auto name : {"vaapimpeg2dec", "vaapimpeg4dec", "vaapih263dec", "vaapih264dec", "vaapivc1dec", "vaapidecodebin"}) {
+        changeRank(name, GST_RANK_PRIMARY);
+    }
 
-    if ((plugin = gst_registry_lookup_feature(reg, "bcmdec")) != nullptr) {
-        qCCritical(GStreamerLog) << "Disable bcmdec";
-        gst_plugin_feature_set_rank(plugin, GST_RANK_NONE);
+    // Enable NVIDIA's proprietary APIs for hardware video acceleration
+    for(auto name : {"nvh265dec", "nvh265sldec", "nvh264dec", "nvh264sldec"}) {
```
andrewvoznytsa: Just came over https://gstreamer.freedesktop.org/documentation/nvcodec/nvh264dec.html?gi-language=c and found its rank is 'primary'. So why we need to assign GST_RANK_PRIMARY to nvh264dec again? Aren't other elements also 'primary' already? Is not, then, is not this extra option in the QGC settings useless?

src/VideoReceiver/GstVideoReceiver.h
patrickelectric: Android works without problem.

src/VideoReceiver/GStreamer.cc
andrewvoznytsa: @patrickelectric looked on rank of a few HW decoders and seems that a lot of them use GST_RANK_PRIMARY in contrast others which are GST_RANK_PRIMARY + 1. The best solution would be to iterate over hw decoders and give them all rank GST_RANK_PRIMARY + 1 or work in reverse way - call this option 'lower software decoding priority' and set avdec_h264 rank to GST_RANK_PRIMARY - 1 or similar. But only if this options is enabled.

patrickelectric: @andrewvoznytsa I'm doing some tests based on this idea, but also with a second approach where I could filter all hardware decodes and their rank and let the user choose how to configure it.

We got a weird problem where gstreamer was trying to use vaapi, that mostly has higher ranks compared to nvdpau, with the proprietary nvidia driver, resulting in a non working pipeline.

I'm going to do more tests and share here.

andrewvoznytsa: @patrickelectric perhaps it would be even better to let advanced users to choose which decider to blacklist and which to make higher priority - just to list and one 'advanced' option. And leave default gstreamer ranks for most of users. It is not trivial task to find config which works good for everyone.

patrickelectric: @andrewvoznytsa that's exactly what I'm planning to do.

patrickelectric:
I have discussed a bit more, and creating a UI to handle all hardware decoders is tricky and may be not necessary, for a first version, we should be fine creating a group of valid hardware decoders for the user and allowing his to change between them.

src/Settings/VideoSettings.cc
```diff
@@ -29,6 +29,10 @@ const char* VideoSettings::videoSourceMPEGTS            = "MPEG-TS (h.264) Video
 const char* VideoSettings::videoSource3DRSolo           = "3DR Solo (requires restart)";
 const char* VideoSettings::videoSourceParrotDiscovery   = "Parrot Discovery";
 
+const char* VideoSettings::forceVideoDecoderDefault = "Default";
```
jaxxzer, line 32: `Software Decoder (Default)`

src/Settings/VideoSettings.h
```diff
@@ -44,6 +45,10 @@ class VideoSettings : public SettingsGroup
     Q_PROPERTY(QString  mpegtsVideoSource       READ mpegtsVideoSource      CONSTANT)
     Q_PROPERTY(QString  disabledVideoSource     READ disabledVideoSource    CONSTANT)
 
+    Q_PROPERTY(QString  defaultVideoDecoder     READ defaultVideoDecoder    CONSTANT)
+    Q_PROPERTY(QString  vaapiVideoDecoder       READ mpegtsVideoSource      CONSTANT)
```
jaxxzer, line 49: copy paste errors on READ

src/VideoReceiver/GStreamer.cc
```diff
@@ -105,20 +105,50 @@ static void qgcputenv(const QString& key, const QString& root, const QString& pa
...
+    // Set rank for specific features
+    changeRank("bcmdec", GST_RANK_NONE);
 
-    if ((plugin = gst_registry_lookup_feature(reg, "bcmdec")) != nullptr) {
-        qCCritical(GStreamerLog) << "Disable bcmdec";
-        gst_plugin_feature_set_rank(plugin, GST_RANK_NONE);
+    if (forceVAAPI || forceNVIDIA) {
+        changeRank("avdec_h264", GST_RANK_MARGINAL + 1);
+    } else {
+        changeRank("avdec_h264", GST_RANK_PRIMARY);
```
jaxxzer, line 137: return?

src/Settings/VideoSettings.cc
patrickelectric: The "default" is not necessary software decoder, we just don't change the gstreamer rankings.

DonLakeFlyer:
> I have discussed a bit more, and creating a UI to handle all hardware decoders is tricky and may be not necessary, for a first version, we should be fine creating a group of valid hardware decoders for the user and allowing his to change between them.

Then the question for me would be whether hardware decode should be on by default? If we can test cross platform that they seem to work fine then I would say yes.

patrickelectric:
> Then the question for me would be whether hardware decode should be on by default? If we can test cross platform that they seem to work fine then I would say yes.

The idea of this patch is to be a first step toward to this direction, by default we are not going to change the plugin/rank behavior, but at the same time we are going to empower the user allowing him to change with hardware decode he wants to use.

Since the hardware x operating system x driver x gstreamer version creates a complex range of options that makes it hard to hardcode the hardware decoder, for a first attempt, we should be fine to create an user selected option.

For windows we are almost sure that we can hardcode the hardware video decoder, but for linux it makes a much more complex and tricky environment, that is probably impossible for us to hardcode a specific plugin.

DonLakeFlyer:
There shouldn't be an isLinux thing to control the video decoder setting visibility. That should be controlled by the _videoSettings.forceVideoDecoder.visible property itself (which is also not handled correctly in the qml changes in this pull). In the C++ side of VideoSettings it should force the setting to enabled/disabled visible/invisible on the appropriate platforms. The Qml should know nothing about that.

patrickelectric:
@DonLakeFlyer Done.

jaxxzer:
It's working well on Windows here.

src/VideoReceiver/GStreamer.cc
jaxxzer: 'Default' option should not change any ranks.

src/Settings/VideoSettings.cc
jaxxzer: Ok, that's good. I think we should have an additional option to force software decoder.

src/VideoReceiver/GStreamer.cc
jaxxzer: In reference also to @andrewvoznytsa comment [here](https://github.com/mavlink/qgroundcontrol/pull/9213/files#discussion_r533461326)

I think we should
- remove this line, do nothing to lower the rank of any decoder
- if user selects 'Default' - we do nothing to change the ranks
- if user selects 'Force *' - we set the rank of the forced decoder PRIMARY + 1.

src/VideoReceiver/GStreamer.cc
```diff
@@ -105,20 +105,58 @@ static void qgcputenv(const QString& key, const QString& root, const QString& pa
...
     }
 
-    GstPluginFeature* plugin;
+    // Decrease software rank to not interfere with next plugins
+    changeRank("avdec_h264", GST_RANK_MARGINAL + 1);
+
+    // Enable VAAPI drivers
+    if (forceVAAPI) {
+        for(auto name : {"vaapimpeg2dec", "vaapimpeg4dec", "vaapih263dec", "vaapih264dec", "vaapivc1dec", "vaapidecodebin"}) {
+            changeRank(name, GST_RANK_PRIMARY);
```
jaxxzer, line 144: If what if there is another h264 decoder that has primary rank? We aren't forcing here in that case. See my other suggestion in the lines above.

src/VideoReceiver/GStreamer.cc
patrickelectric: > 'Default' option should not change any ranks.

Default, in the actual patch. Does not change any rank, the early return will be used.

src/VideoReceiver/GStreamer.cc
patrickelectric: That's all vaapi decoders.

src/VideoReceiver/GStreamer.cc
patrickelectric: Done.

src/Settings/VideoSettings.cc
patrickelectric: done

src/VideoReceiver/GStreamer.cc
jaxxzer: I think this has been addressed by forcing the decoder to the user selection, without changing the default behavior.

jaxxzer:
> Having to have settings in video to support specific video hardware seems crazy. Why doesn't gstreamer work correctly in the first place? Seems like a pretty fundamental flaw/bug in gstreamer that it doesn't use hardware decoding when it is actually available.
> 
> Poking around here: https://gstreamer.freedesktop.org/documentation/tutorials/playback/hardware-accelerated-video-decoding.html?gi-language=c says this: "The GStreamer developers often rank hardware decoders lower than the software ones when they are defective. This should act as a warning."

Gstreamer will use hardware decoding when available - to a point. It depends on the graphics hardware, the graphics drivers installed on the OS, and the drivers that are supported by gstreamer, and the relative stability of the graphics backends in gstreamer. The Direct3D 11 decoder is supported on windows in gstreamer 1.18, but it is considered less stable than the software decoder by gstreamer, so gstreamer defaults to the software decoder. I've had a dozen people test the D3D decoder in QGC on Windows and with improved results, so we want people to be able to use it, but not force them to because of gstreamer's 'stability warning'.

I think we can eventually get to a point where the best situation is configured automagically, but we need some more feedback and data on which decoder work best in each situation.

DonLakeFlyer:
> I think we can eventually get to a point where the best situation is configured automagically, but we need some more feedback and data on which decoder work best in each situation.

Ok, understood.

jaxxzer (approved):

DonLakeFlyer (changes requested):

src/Settings/Video.SettingsGroup.json
```diff
@@ -128,6 +128,14 @@
     "longDesc":  "If this option is enabled, the rtpjitterbuffer is removed and the video sink is set to assynchronous mode, reducing the latency by about 200 ms.",
     "type":             "bool",
     "default":     false
+},
+{
+    "name":             "forceVideoDecoder",
+    "shortDesc":        "Force specific category of video decode",
+    "longDesc":         "Force the change of prioritization between video decode methods, allowing the user to force some video hardware decode plugins if necessary.",
+    "type":             "string",
```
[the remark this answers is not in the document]
patrickelectric: I was unable to find something that matches your description in QGC and in the *.json files of the settings, there is no "unsigned" values that I could find, only "win", "float", "int32", "SimpleItem".
Can you point a settings that matches what are you talking about  ?

src/QGCApplication.cc
```diff
@@ -356,9 +356,18 @@ QGCApplication::QGCApplication(int &argc, char* argv[], bool unitTesting)
     // We need to set language as early as possible prior to loading on JSON files.
     setLanguage();
 
+    // toolbox should be created before gstreamer initialization
     _toolbox = new QGCToolbox(this);
     _toolbox->setChildToolboxes();
 
+#if defined(QGC_GST_STREAMING)
+    bool forceSoftware = toolbox()->settingsManager()->videoSettings()->forceVideoDecoder()->rawValue() == VideoSettings::forceVideoDecoderSoftware;
```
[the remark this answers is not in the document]
patrickelectric: This is now possible since blacklist was separated from init call, done.

src/Settings/Video.SettingsGroup.json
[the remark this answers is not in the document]
DonLakeFlyer: * Supported types are here: https://github.com/mavlink/qgroundcontrol/blob/master/src/FactSystem/FactMetaData.cc#L96
* Example usage: https://github.com/mavlink/qgroundcontrol/blob/master/src/Settings/Video.SettingsGroup.json#L63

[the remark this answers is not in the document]
patrickelectric: Thanks @DonLakeFlyer, but how is it possible to control the visibility of each option based in the operating system ?
From what I can see, the fact system does not have a function to remove items only to add, I also was unable to find any configuration regard the operating system.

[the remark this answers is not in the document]
DonLakeFlyer: I was only commenting to the using of a string versus an enum for a settings. The way you were setting the visible property as needed in the settings C++ code was correct. The FactSystem json parser does not have the ability to set visible on an OS by OS base directly in the json. The only support sort of like that is `default` versus `mobileDefault` (https://github.com/mavlink/qgroundcontrol/blob/master/src/FactSystem/FactMetaData.cc#L136).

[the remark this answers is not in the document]
patrickelectric: Hi @DonLakeFlyer, thanks, this should be done in the last version, I'm defining the default configuration in the settings file and the others in the VideoSettings class, where the enumInfo is added by the operating system, this will save the configuration as numbers as requested and also control each possible value based in the operating system.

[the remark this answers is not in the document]
DonLakeFlyer: Sorry I wasn't reading your question correctly. I thought you were still talking about `visible`. The way to handle an enum list which is different based on OS is to include the full list in the json. And then on the C++ side in the settings code you ask for the enum info, remove the ones which are inappropriate for the OS and then stuff it back in. This way the translation mechanism uses the standard json translation mechanism. It also removes confusion around some part of the enum in one place and some part in another place. I believe there are examples of this pattern in other settings someplace.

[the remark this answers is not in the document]
patrickelectric: Hi @DonLakeFlyer, no problem, since I was unable to find how to remove such items, I have created a function in the fact system to do such thing, I hope that is ok.

src/FactSystem/FactMetaData.h
```diff
@@ -148,6 +148,9 @@ class FactMetaData : public QObject
     /// Used to add new values to the enum lists after the meta data has been loaded
     void addEnumInfo(const QString& name, const QVariant& value);
 
+    /// Used to remove values from the enum lists after the meta data has been loaded
+    void rmEnumInfo(const QString& name);
```
DonLakeFlyer, line 152: Can you name is `removeEnumInfo`? This needs to work from the value not the string. The strings are not unique since they can be translated. Only the values are unique.

patrickelectric: Thanks for pointing this out, the last version contains the requested changes.

DonLakeFlyer:
There should be a public enum in VideoSettings which mirrors the enum values in the json. There shouldn't be magic numbers in the C++ side of things. Looks at UnitsSettings for examples of the same pattern.

patrickelectric:
Using enum over raw values.

src/VideoManager/VideoManager.cc
```diff
@@ -104,6 +104,11 @@ VideoManager::setToolbox(QGCToolbox *toolbox)
    connect(pVehicleMgr, &MultiVehicleManager::activeVehicleChanged, this, &VideoManager::_setActiveVehicle);
 
 #if defined(QGC_GST_STREAMING)
+    bool forceSoftware = _videoSettings->forceVideoDecoder()->rawValue() == 1;
```
DonLakeFlyer, line 107: This should be using the enum values.

src/Settings/VideoSettings.h
```diff
@@ -35,6 +35,16 @@ class VideoSettings : public SettingsGroup
     DEFINE_SETTINGFACT(streamEnabled)
     DEFINE_SETTINGFACT(disableWhenDisarmed)
     DEFINE_SETTINGFACT(lowLatencyMode)
+    DEFINE_SETTINGFACT(forceVideoDecoder)
+
+    enum VideoDecoderOptions {
+        Default = 0,
```
DonLakeFlyer, line 41: Can you prefix all of these with `ForceVideoDecoder`. Referencing like `VideoSettings::Default` is just too generic to mean much, whereas `VideoSettings::ForceVideoDecoderDefault` is understandable at a glance.

src/VideoManager/VideoManager.cc
patrickelectric: Sorry for that, fixed.

src/Settings/VideoSettings.h
patrickelectric: Using ForceVideoDecoder prefix now.

DonLakeFlyer (approved):

DonLakeFlyer:
Hmm, there's a race condition in VehicleLinkManagerTest causing it to fail intermittently. Unrelated to this pull.

PR #9217, opened 2020-11-27
Events interface (first iteration)

bkueng:
It will replace the `STATUSTEXT` mavlink messages with event ID's + metadata.
Full details: https://docs.google.com/document/d/18qdDgfML97lItom09MJhngYnFzAm1zFdmlCKG7TaBpg/edit

PR for PX4: https://github.com/PX4/PX4-Autopilot/pull/16293

@DonLakeFlyer @dogmaphobic can you go through it to check if you miss something?
In particular:
- event descriptions can contain URL's which should eventually be clickable and open the link.
- event descriptions can contain params (they can be tagged as a URL). Clicking on those should open the param edit dialog and I believe this is possible with custom URL handlers.
- for health reporting you get this information: https://github.com/mavlink/qgroundcontrol/commit/0aa362be6ca1024b08c59abb861f0d1d6c055f93#diff-2a76975fc3bdcd8b6958c58a3db89aa591bc14ce457863bfb0ce7b5b8e9c6827R88. It can be used to drive part of the UI: e.g. disabling a 'start mission' or 'takeoff' button, show the GPS icon in red, etc. Let me know what you'd like to have here, I'm sure you have more ideas.

Other:
- There's a new submodule: https://github.com/bkueng/libevents (that implements the mavlink protocol and event parsing)
- The 2. commit contains a test for using sensor calibration based on the events interface for reliable transfer.

Still missing:
- use COMPONENT_INFO to load metadata
- UI work

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "libs/xz-embedded"]
 	path = libs/xz-embedded
 	url = https://github.com/Auterion/xz-embedded.git
+[submodule "libs/libevents/libevents"]
+	path = libs/libevents/libevents
+	url = https://github.com/mavlink/libevents.git
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -102,6 +102,19 @@ INCLUDEPATH += libs/eigen
 DEFINES += NOMINMAX
 
 #
+# [REQUIRED] Events submodule
+HEADERS+= \
+	libs/libevents/libevents/libs/cpp/protocol/receive.h \
+	libs/libevents/libevents/libs/cpp/parse/parser.h \
+	libs/libevents/libevents/libs/cpp/generated/events_generated.h \
+	libs/libevents/libevents_definitions.h
+SOURCES += \
+	libs/libevents/libevents/libs/cpp/protocol/receive.cpp \
+	libs/libevents/libevents/libs/cpp/parse/parser.cpp \
+	libs/libevents/definitions.cpp
+INCLUDEPATH += libs/libevents
+
+#
 # [REQUIRED] shapelib library
 INCLUDEPATH += libs/shapelib
 SOURCES += \
--- a/libs/libevents/definitions.cpp
+++ b/libs/libevents/definitions.cpp
@@ -1,0 +1,27 @@
+/****************************************************************************
+ *
+ * (c) 2020 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ *
+ * QGroundControl is licensed according to the terms in the file
+ * COPYING.md in the root of the source code directory.
+ *
+ ****************************************************************************/
+
+#include "libevents_definitions.h"
+
+#include <QGCLoggingCategory.h>
+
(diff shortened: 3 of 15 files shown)


DonLakeFlyer:
I took a first quick look through this. I will need to spend more time with it. My first comments:
* For the translation parameter substitution stuff `{1>` there usually also needs to be a mechanism to deal with plurals. As an example looks at handling plurals in [Qt](https://doc.qt.io/qt-5/i18n-source-translation.html#:~:text=Qt%20indexes%20each%20translatable%20string%20by%20the%20translation,the%20Q_OBJECT%20macro%20in%20each%20new%20class%20definition.)
* Sending links to parameters is problematic. It talks about popping the parameter editor dialog. But for many users they have no idea what the actual parameter they are setting is named. They come at it directly from the setup page ui, not the parameter editor list. In fact many third party custom builds hide the full parameter editor list from their users. Going right into the edito dialog may be pretty confusing for these folks.
* The spec seems to talk about mavlink dialects for things. For example it has what looks to be flight modes in mavlink headers for `FLIGHT_MODE_CATEGORY`. Same thing for stuff like `ARMING_CHECKS_SUMMARY`. To me mavlink dialect isn't the way to handle this. These sorts of firmware specific things should come through in COMPONENT json metadata files.

I need to make a pass through this from a GCS perspective looking at what things QGC does on it's own that should really be controlled by the firmware side using this mechanism.

bkueng:
Thanks for the initial feedback.

> For the translation parameter substitution stuff {1> there usually also needs to be a mechanism to deal with plurals. As an example looks at handling plurals in Qt

I'll have a look.

> Sending links to parameters is problematic. It talks about popping the parameter editor dialog. But for many users they have no idea what the actual parameter they are setting is named. They come at it directly from the setup page ui, not the parameter editor list. In fact many third party custom builds hide the full parameter editor list from their users. Going right into the edito dialog may be pretty confusing for these folks.

I see that and this is mostly intended for developers/dyi and would then only be shown if that profile is selected. So not relevant for end products.
At the same time you'll see the parameter description when opening the dialog, so you don't need to know the param name.

> The spec seems to talk about mavlink dialects for things. For example it has what looks to be flight modes in mavlink headers for FLIGHT_MODE_CATEGORY. Same thing for stuff like ARMING_CHECKS_SUMMARY. To me mavlink dialect isn't the way to handle this. These sorts of firmware specific things should come through in COMPONENT json metadata files.

Not quite clear what you mean, there's nothing dialect specific. If the GCS needs to know the flight modes to the category assignment, then that should be in the json metadata file.

bkueng:
@DonLakeFlyer this is now ready:
- preflight checks not added yet (still expecting you to have a look)
- there's compatibility support: px4 sends STATUSTEXT in addition to each event, and appends a tab to the message, which is then ignored in QGC. Existing QGC just ignores the tab (including speech output)
- events are displayed in the same ui as existing, this will need some rework:
qgc_events_interface

DonLakeFlyer:
@bkueng Looked through this. All seems reasonable to me. Ok to merge?

bkueng:
Yes, I don't have anything more to add at the moment.

Issue #9219, opened 2020-11-29
flighMapPosition not restored on Android


remster:
## Expected Behavior
When the app is restarted it should restart with the mapview matching that when it previously exited. Especially when Android kills QGC as it's going to the background. 

## Current Behavior
The code that tries to achieves this lives in the [`QGroundControlQmlGlobal` d'tor](https://github.com/mavlink/qgroundcontrol/blob/master/src/QmlControls/QGroundControlQmlGlobal.cc#L40) and is, as such, sensitive to whether the app goes away gracefully. In my experiments it rarely does on Android.

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1. Start QGC on Android
2. Pan somewhere
3. Kill QGC using the phone's recent apps menu
4. Start QGC again and observe that the map is looking at 0,0 and at the coarsest zoom.
![Screenshot_20201129-101701](https://user-images.githubusercontent.com/8327961/100537851-3390a780-322c-11eb-87c7-657f4faca8f3.jpg)

## System Information
When posting bug reports, include the following information
- Operating System: Android 10
- QGC Version: master at `1e6e1c6e43678dea1519650cfe835eb804cbdae6`
- QGC build: N/A
- Flight Controller: N/A
- Autopilot (with version): N/A

PR #9220, opened 2020-11-29
Save coordinates on every, settled viewport change

remster:
Addresses: https://github.com/mavlink/qgroundcontrol/issues/9219


DonLakeFlyer:
Yup destructor is wrong. Why complicate this with a timer though? Just add the save to the set* methods.

remster:
> Yup destructor is wrong. Why complicate this with a timer though? Just add the save to the set* methods.

I did initially - it's called repetitively as the pan progresses - i thought to spare the disk these useless write requests and wait for the user to stop dragging the map. Sure - the OS will probably handle this way better. Up to you - just say the word.

DonLakeFlyer:
Hmm, yeah. Qt itself has its own internal buffering prior to writing to disk though and then you have to OS itself. So I still think simpler is better for now.

remster:
Done: https://github.com/mavlink/qgroundcontrol/pull/9220/commits/ce08f690aa5a9f9adad0852e4350c68bbf22b41b

DonLakeFlyer:
Great. Thanks for the fix @remster

aib:
`QSettings` calls `sync` in its destructor, absolutely destroying performance where it is honored. See #10070.

Issue #9215, opened 2020-11-26
Cube Orange PX4 switch fails with "Unable to find specified firmware for board type"


dell640:
[As per discussion at px4.io](https://discuss.px4.io/t/cube-orange-px4-stack-issue/19754/7) with @DonLakeFlyer I'm rising this as issue:

I would like to upload PX4 stack to my Cube Orange, unfortunately without success.

When I upgrade my Cube to PX4 via MissionPlanner, upload is successful, however Cube hangs in bootloader and stack is never loaded and running.

When I try to upgrade to PX4 from QGC:
![image](https://user-images.githubusercontent.com/11704829/100325072-676b9300-2fc8-11eb-937b-d4430d9816df.png)
I’m getting this error:
![image](https://user-images.githubusercontent.com/11704829/100325096-70f4fb00-2fc8-11eb-8b45-704969d851a4.png)

I have two Cube Orange units here, both behaves the same.

Testing computer running W10.

Tried multiple QGC versions from 3.5.6 up to the latest and also daily build. All have the same “error”.

Also I tried to uninstall USB drivers via Device Manager, that does not help. Then reinstalled UAV drivers, but problem persist.
Tried another computer (also W10) with same results.

DonLakeFlyer:
Turn on console logging for FirmwareUpgradeLog as described here and then attach the log so I can look at it.

DonLakeFlyer:
Also wait a minute these screenshots are not from a Daily build. They are from Stable. You need to use a daily build to get Cube Orange fixes.

DonLakeFlyer:
> Turn on console logging for FirmwareUpgradeLog as described here and then attach the log so I can look at it.

Never mind I can repro it myself.

DonLakeFlyer:
This is because of the ArduPilot bootloader on the board returning a different board id. Trying to figure out how to make it work...

dell640:
> 
> 
> This is because of the ArduPilot bootloader on the board returning a different board id. Trying to figure out how to make it work...

Thanks for checking that. As stated at first post, when proper PX4 firmware (v1.11.2 - cubepilot_cubeorange_default.px4) is uploaded via MissionPlanner (where is goes smoothly) Cube hangs in bootloader and PX4 stack is not running properly. So, maybe it is deeper incompatibility issue.
Looking forward to your discoveries.

DonLakeFlyer:
[...]

PR #9221, opened 2020-11-29
VehicleSetup: firmware upgrade use static map of px4 boards names

dagar:


dagar:
Tested on a few random boards like the CUAV v5+, CubeOrange, etc.

DonLakeFlyer:
Yeah, much simpler thanks. But can you leave the board ids in the header and use them in `px4_board_name_map `? Having raw numbers there just seems really odd. Also having them in a separate header provides a consistent place for them which is firmware independent. I often look up a board id when I'm debugging things.

DonLakeFlyer:
Fix for #9215

DonLakeFlyer:
@dagar Could you get this working? I need it for Stable.

dagar:
Needs fmu-v2/v3 testing.

DonLakeFlyer:
> Needs fmu-v2/v3 testing.

I'll do that today.

PR #9225, opened 2020-11-30
COMPONENT_INFORMATION change ftp spec

DonLakeFlyer:
* From `mavlinkftp://` to `mftp://` to save characters in message packet
* Add support for specifying component id in uri. Format `mftp://[;comp=<id>]<path>`. This allows the component which delivers the metadata files to differ from the component you are requesting it from. So for example you could store all the metadata on a companion computer.


DonLakeFlyer:
I will be pushing a corresponding mavlink docs spec change for this as well

PR #9227, opened 2020-12-01
Ardupilot: request telemetry streams upon reconnection

### LINKED ISSUE #9209: Ardupilot: Vehicle Fact data goes stale after reconnect
When using Ardupilot and QGC, after a disconnect/reconnect event (changing batteries on the vehicle for example) the Vehicle Fact data never updates to new values, it just shows the old values.

QGC: Stable_V4.0
APM: v4.0.5

dakejahl:
Fixes #9209 

@DonLakeFlyer let me know if this looks good to you, I tested on my end and it does the trick

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -673,7 +673,9 @@ void APMFirmwarePlugin::initializeStreamRates(Vehicle* vehicle)
 
 void APMFirmwarePlugin::initializeVehicle(Vehicle* vehicle)
 {
-    vehicle->setFirmwarePluginInstanceData(new APMFirmwarePluginInstanceData);
+    if (!vehicle->firmwarePluginInstanceData()) {
+        vehicle->setFirmwarePluginInstanceData(new APMFirmwarePluginInstanceData);
+    }
 
     if (vehicle->isOfflineEditingVehicle()) {
         switch (vehicle->vehicleType()) {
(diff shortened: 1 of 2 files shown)


dakejahl:
ugh I actually realized last night that this line can cause a memory leak in this PR
https://github.com/mavlink/qgroundcontrol/blob/Stable_V4.0/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc#L676

DonLakeFlyer:
Number of things:
* Re-using `FirmwarePlugin::initializeVehicle` breaks the contract for that method which is described as "Called when Vehicle is first created to perform any firmware specific setup.". Instead I would make a `FirmwarePlugin::reInitializeVehicleAfterCommLoss` method with the appropriate contract. The default implementation is do nothing and then APMFirmwarePlugin overrides it. 
* On the APMFirmwaerPlugin override call `initializeVehicle`. And then add comments on both `initializeVehicleAfterCommLoss ` and 'initializeVehicle` methods as to exact usage. Such that further down the road it will be clear if the two start to need to do diverge and do different things.
* The call to initializeVehicleAfterCommLoss needs to be called on any links which regain comms not just priority links.
* The fix on master will be different and go into VehicleLinkManager.

DonLakeFlyer:
@dakejahl Updates?

dakejahl:
Haven't done it yet, I have it on my TODO list though along with the CSV log memory limit. Been busy with other things + transitioning into holiday mode :santa:

DonLakeFlyer:
I"m going to close this instead of just letting it sit here. You can reopen and or resubmit when you have it done

PR #9228, opened 2020-12-01
Fix race condition for sending GCS heartbeat

### LINKED ISSUE #9210: _sendGCSHeartbeat() race condition crash
I believe there is a race condition between 
https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/MultiVehicleManager.cc#L49
and
https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/MultiVehicleManager.cc#L133
Which causes random crashes.

I will look into it more in a few days.

dakejahl:
Should fix #9210, it's hard to reproduce so I will let you know if our crashes go away after testing this more.

### DIFF
--- a/src/Vehicle/MultiVehicleManager.cc
+++ b/src/Vehicle/MultiVehicleManager.cc
@@ -134,7 +134,13 @@ void MultiVehicleManager::_vehicleHeartbeatInfo(LinkInterface* link, int vehicle
     _vehicles.append(vehicle);
 
     // Send QGC heartbeat ASAP, this allows PX4 to start accepting commands
-    _sendGCSHeartbeat();
+    if (_gcsHeartbeatEnabled) {
+        _gcsHeartbeatTimer.stop();
+        _sendGCSHeartbeat();
+        _gcsHeartbeatTimer.start();
+    } else {
+        _sendGCSHeartbeat();
+    }
 
     qgcApp()->toolbox()->settingsManager()->appSettings()->defaultFirmwareType()->setRawValue(vehicleFirmwareType);
 


dakejahl:
CI is failing, looks like a Qt bug
```
C:\projects\qgroundcontrol\src\Vehicle\MultiVehicleManager.cc(137): error C2039: 'active': is not a member of 'QTimer'
```
For whatever reason Qt on Android doesn't know about `QTimer::active()`

Changing to
```
if (_gcsHeartbeatEnabled) {
...
}
```

DonLakeFlyer:
Can you tell me about the details on why this crashes?

dakejahl:
Honestly I am not sure, but my guess is that the function is non-reentrant for some reason. When a vehicle heartbeat is seen for the first time, the function `_sendGCSHeartbeat()` is sent. This same function is also sent on an interval from a QTimer. Even if this isn't the cause of the bug (it may not be, it only happens on Android and I don't have the Android device right now, but my guy is testing it today), I still think this fix is valid since we probably don't want to be creating situations where we have reentrant functions. The other thought is that the link is trying to get used before it's fully constructed... I think having the QTimer get started in the CTOR is probably not the best way to do it either, so we could think about changing where that timer is started, giving the app more time to get everything in place.

DonLakeFlyer:
Timer firing happens on the same thread that it was created on. Also timer firing only happens from the event queue so there is no way that you could end up in _sendGCSHearbeat from `MultiVehicleManager::_vehicleHeartbeatInfo` and the timer at the same time. I don't think this code is going to make any difference. There is something else wrong likely having to do with iterating the link list. But I don't see anything there either since it holds strong references which should prevent the links from being ripped out from under it.

DonLakeFlyer:
I think it might be a multi-threaded link shutdown sequencing problem here:
```        if (link->isConnected() && !link->linkConfiguration()->isHighLatency()) {```

The link is still held alive by the shared pointer but the configuration is gone, hence `link->linkConfiguration()` is null. Digging further to verify that can happen...

**Comment by dakejahl** [comment]
That's really good to know, the way Qt processes events has always been something I'm a bit confused about, like what events happen where and when. So what about moving the start of the QTimer for sending GCS heartbeat to somewhere other than the CTOR? For more context, I _**only**_ ever see this crash when the app launches.

**Comment by DonLakeFlyer** [comment]
> So what about moving the start of the QTimer for sending GCS heartbeat to somewhere other than the CTOR? For more context, I _**only**_ ever see this crash when the app launches.

That wouldn't make any difference. Given the way timers work. These sorts of crashes are almost always related to link references. I just made a huge pass fixing it all up. But there is something I missed.

Is the crash over radio telemetry or USB?

**Comment by DonLakeFlyer** [comment]
One thing to try would be to force this connect to be queued:
```    connect(_mavlinkProtocol, &MAVLinkProtocol::vehicleHeartbeatInfo, this, &MultiVehicleManager::_vehicleHeartbeatInfo);
```

That should happen automatically since _mavlinkProtocol and MultiVehicleManager are on different threads. Signals which cross threads are always supposed to be queued to get the slot to run in the appropriate thread. But maybe something is going wrong in Qt setup.

DonLakeFlyer:
Try putting through a pull with this changed:
```        if (link->isConnected() && !link->linkConfiguration()->isHighLatency()) {```

Add a null check on link->linkConfiguration() prior to reference. Then I'll merge that through and we should be able to see from Google Play whether it's fixed or not. Since this crash shows up in reporting there.

**Comment by DonLakeFlyer** [comment]
Wait, I think you might have something on the CTOR but it relates to how QGCToolbox works not general Qt. This:
```
    if (_gcsHeartbeatEnabled) {
        _gcsHeartbeatTimer.start();
    }
```

Should be in setToolbox so it only startup once all the tools have finished completion. Can you move that and add the null check on linkConfiguration and we'll see what happens.

**Comment by dakejahl** [comment]
oh yeah I see now in `MultiVehicleManager::setToolbox()`
```
_mavlinkProtocol =           _toolbox->mavlinkProtocol();
```
I wonder if the timer fires before `multiVehicleManager->setToolbox()` is called and it therefore crashes on a nullptr for 
```
mavlink_msg_heartbeat_pack_chan(_mavlinkProtocol->getSystemId(),
                                            _mavlinkProtocol->getComponentId(),
                                            link->mavlinkChannel(),
                                            &message,
                                            MAV_TYPE_GCS,            // MAV_TYPE
                                            MAV_AUTOPILOT_INVALID,   // MAV_AUTOPILOT
                                            MAV_MODE_MANUAL_ARMED,   // MAV_MODE
                                            0,                       // custom mode
                                            MAV_STATE_ACTIVE);       // MAV_STATE
```

PR #9233, opened 2020-12-01
Fix sendGCSHeartbeat() crash

dakejahl:
Attempted fixes for #9210 
See #9228 for more context.


DonLakeFlyer:
@dakejahl Remind me in say 5 days to look at Google Play to see if the crashes are gone.

PR #9235, opened 2020-12-04
Airship airspeed sensor update

antonerasm:
This pull request makes it possible to view and calibrate airspeed sensors for airships.


DonLakeFlyer:
@antonerasm Thanks for the pull

PR #9237, opened 2020-12-05
COMPONENT_INFORMATION use LZMA compression as standard

DonLakeFlyer:
Change the compress standard from gzip to lzma which gives much higher compression rations.


DonLakeFlyer:
Doesn't work on Android

Issue #9236, opened 2020-12-05
Current Daily: terrain altitude bar seems wrong


Antiheavy:
Current daily build seems to have something going on with the terrain display.  I'm testing indoors with a vehicle connected and no GPS fix (or home location yet).  QGC seems to understand the local position ground altitude (as shown in the Mission Start), however, it is showing the mission plan as being completed below the terrain.

Also: does the RED flight path lines indicate a problem, in this case below terrain?

Screenshot:
![image](https://user-images.githubusercontent.com/12117027/101247845-45b58e80-36e1-11eb-8033-1023d98172a0.png)

another example
![image](https://user-images.githubusercontent.com/12117027/101247913-c5435d80-36e1-11eb-8619-443c8a093035.png)

DonLakeFlyer:
Yeah, something screwed up there. I thought I already had a 4.1 issue on it somewhere but can't find now. Can you attach the plan you are testing with.

The red line indicates below terrain.

Antiheavy:
>  Can you attach the plan you are testing with.

couple examples attached in this zip folder.
[below-ground-missions.zip](https://github.com/mavlink/qgroundcontrol/files/5647686/below-ground-missions.zip)

moreba1:
1- when we click on "SET TO MAP CENTER" button home altitude set to 0.
2- when we change location of Launch with mouse (home) , altitude of other relative waypoint does not changed .

DonLakeFlyer:
> 1- when i click on "SET TO MAP CENTER" button home altitude set to 0. also in end of plan file altitude of "plannedHomePosition": is 0.
> 2- when i change location of Launch with mouse (home) , altitude of other relative waypoint does not changed .

Can you put these in a separate issue?

moreba1:
> 
> 
> > 1- when i click on "SET TO MAP CENTER" button home altitude set to 0. also in end of plan file altitude of "plannedHomePosition": is 0.
> > 2- when i change location of Launch with mouse (home) , altitude of other relative waypoint does not changed .
> 
> Can you put these in a separate issue?

these are reason of terrain altitude bar issue.

DonLakeFlyer:
> these are reason of terrain altitude bar issue.

So far, not according to the plan files attached to this issue. They have good launch altitudes in them. I think this a separate issue, possibly related somehow but maybe not.

DonLakeFlyer:
[...]

Issue #9207, opened 2020-11-23
Height bar issue in daily build version


moreba1:
currently follow terrain removed from survey and AGL alt setting for the waypoint. also, the height bar widget shows the altitude of waypoints lower than the terrain ground altitude.

![image](https://user-images.githubusercontent.com/14200369/99999931-1e53ed00-2dd6-11eb-93bd-bf0a79d07d9f.png)

PR #9241, opened 2020-12-07
Fix fuzzyCompare boolean comparison

DonLakeFlyer:
Related to #9236, #9207


DonLakeFlyer:
New daily should be out in couple hours. @moreba1, @antiheavy Can you guy double check with some testing?

PR #9242, opened 2020-12-07
COMPONENT_INFORMATION use LZMA compression as standard

bkueng:
@DonLakeFlyer this should work better. It's using xz-embedded from https://tukaani.org/xz/embedded.html directly.

Tested on Linux so far.


mrpollo:
@DonLakeFlyer could you please help us with a review?

DonLakeFlyer:
I haven't had time to look at this. I'm working on fixes for next stable. I'm keeping @bkueng up to date with my status.

DonLakeFlyer:
@bkueng Is there a reason for this to be an Auterion submodule as opposed to including directly in the QGC repro like most other libs?

bkueng:
I would keep it as a submodule (for the git history and easy updating), but where it's stored really doesn't matter. The reason why it's a mirror (and not https://git.tukaani.org/?p=xz-embedded.git) is that CI had issues with cloning (`
gnutls_handshake() failed: Handshake failed`, see https://github.com/mavlink/qgroundcontrol/pull/9242/commits/733367885e90e16b8e283144455b075d5a11477d)

DonLakeFlyer:
> easy updating

Ah, I didn't realize it was a direct mirror. Sounds fine then. I'll merged this in once Stable goes out.

DonLakeFlyer:
@bkueng I think the file extension should be lzma. I believe the .xz file extension implies the possibility of a multiple file archive compression which is incorrect for this usage in mavlink. It can only be as single compressed file.

bkueng:
`xz` is more generic and supports both, while `lzma` is the legacy format (and command), and for the correct output for mavlink you have to use the `xz` command (`xz --format=xz` to be explicit). So using `.lzma` would be incorrect.

DonLakeFlyer:
Ah, sorry. I was thinking the .xz format was a multi-stream archive format which it is not. It is a single stream format.

bkueng:
So is this good to merge?

DonLakeFlyer:
I have't run it yet. Just had a quick look through the code. Still spending all my time on stable. I just marked 4.1.1 which is hopefully the stable I will announce and I'll be done.

DonLakeFlyer:
@bkueng Ok, finally had time to take a full look. Can you remove the MockLink gzip files (src/comm/MockLink.Version.MetaData.json.gz) from the repo and the resource file?

bkueng:
Done and rebased

mrpollo:
@DonLakeFlyer This PR is ready to go whenever you have a chance to merge!

Issue #9234, opened 2020-12-02
APM Sub: enabling motor test induces hundreds of mavlink command error popups


jaxxzer:
After the 'enable motor test' switch is clicked on the motor test page, hundreds of popups appear with a message like
"unable to send mavlink command: previous command not acked"

The messages come so rapidly my desktop froze and I was unable to make a screenshot.

jaxxzer:
This should be fixed for 4.1

DonLakeFlyer:
Broken by the recent changes to not queue mavlink commands. The previous command still hasn't finished before the next one is sent.

DonLakeFlyer:
@Williangalvani This is fixed right?

Williangalvani:
Yes!

PR #9243, opened 2020-12-08
Sub motor test fixes

Williangalvani:
- Stops the timer when it shouldn't be sending motor tests messages. This fixes some "motor test command failed" after user clicked the motor test toggle to stop the test/disarm, but the vehicle state in qgc hasn't received a disarmed heartbet yet

 - Disables showError for sub, fixing #9234

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -3150,9 +3150,9 @@ void Vehicle::setSoloFirmware(bool soloFirmware)
     }
 }
 
-void Vehicle::motorTest(int motor, int percent, int timeoutSecs)
+void Vehicle::motorTest(int motor, int percent, int timeoutSecs, bool showError)
 {
-    sendMavCommand(_defaultComponentId, MAV_CMD_DO_MOTOR_TEST, true, motor, MOTOR_TEST_THROTTLE_PERCENT, percent, timeoutSecs, 0, MOTOR_TEST_ORDER_BOARD);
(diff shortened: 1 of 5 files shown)


jaxxzer (approved):
LGTM

src/Vehicle/Vehicle.cc
```diff
@@ -3152,7 +3152,10 @@ void Vehicle::setSoloFirmware(bool soloFirmware)
 
 void Vehicle::motorTest(int motor, int percent, int timeoutSecs)
 {
-    sendMavCommand(_defaultComponentId, MAV_CMD_DO_MOTOR_TEST, true, motor, MOTOR_TEST_THROTTLE_PERCENT, percent, timeoutSecs, 0, MOTOR_TEST_ORDER_BOARD);
+    // Sub requires periodic motortest messages, and we send them at a high rate to improve responsiveness.
+    // On higher latency connection, `showError=true` caused lots of "Unable to send command: Waiting on previous response to same command." errors.
+    bool showError = !sub();
```
DonLakeFlyer, line 3157: This isn't correct. This is an ArduSub problem not a "sub" problem isn't it?

Williangalvani: Right. What do you think of just exposing showError on the signature `void Vehicle::motorTest(int motor, int percent, int timeoutSecs, bool showError)` and letting qml decide it?

DonLakeFlyer: That sounds fine.

Williangalvani: ok, updated!

Issue #7635, opened 2019-07-28
Linux AppImage fix link libts


tilaktilak:
In [create_linux_appimage.sh](https://github.com/mavlink/qgroundcontrol/blob/1f283ad0dda59f06d3df9cdebbf02a7ed42453e1/deploy/create_linux_appimage.sh#L61) this link is dead.
We could replace this line by : ` wget -c --quiet "http://old-releases.ubuntu.com/ubuntu/pool/universe/t/tslib/libts-0.0-0_1.0-11_amd64.deb"

DonLakeFlyer:
Is this something really new? I just rebuilt stable 3.5 and the app image worked for me when I ran it.

DonLakeFlyer:
Also if this is currently failing if you could help to fix the script such that it fails the build if the wgets fail that would be awesome.

tilaktilak:
working on it at the moment,
I probably go for something like that : 
```
# exit when any command fails
set -e

wget -c --no-verbose http://ftp.us.debian.org/debian/pool/main/u/udev/udev_175-7.2_amd64.deb
wget -c --no-verbose http://ftp.us.debian.org/debian/pool/main/s/speech-dispatcher/speech-dispatcher_0.8.8-1_amd64.deb
wget -c --no-verbose http://ftp.us.debian.org/debian/pool/main/libs/libsdl2/libsdl2-2.0-0_2.0.2%2bdfsg1-6_amd64.deb
```

I added set -e to exit in case of an error
I changes --quiet to --no-verbose so it does not print everything but the errors   

And that make me realized that a lot of the wget result in a 404 error, that's mainly because those packages are old (from debian wheezy). 
It probably worth to have a look of the all script to update the dependencies.    

Maybe @dagar can help me on this ? As I've seen he wrote this script at first.

DonLakeFlyer:
Sorry, I don't know much about this at all to help. I'm not much of a Linux person.

DonLakeFlyer:
> I just rebuilt stable 3.5 and the app image worked for me when I ran it.

So it's unclear to me if the AppImage which has been released for Stable 3.5 is broken or not?

tilaktilak:
It's also working for me. But for instance libts is meant to be used for touchscreen i guess, so not sure if my ubuntu  is a completely good test config.

DonLakeFlyer:
[...]

PR #9253, opened 2020-12-15
create_linux_appimage.sh: stop trying to wget from not-found URL(s)

pestophagous:
I discovered the issues with this script while using it to build an AppImage containing a modified application binary of [CustomQGC](https://github.com/mavlink/qgroundcontrol/blob/master/custom-example/custom.pri). I am working for one of several teams that got involved with [US Defense drone-related initiatives](https://www.defense.gov/Newsroom/Releases/Release/Article/2318799/defense-innovation-unit-announces-suas-product-availability-to-provide-secure-c/) recently.

For now, this issue is only pertinent to those who **deploy** QGC. That probably isn't even a majority of the developers. So this isn't affecting a lot of people.

However, I do think you should add `set -Eeuxo` to your bash script A.S.A.P.  The next time a depended-upon URL goes offline, it may not be as innocuous for your AppImages. Having an "early warning" (a script failure in CI) will help avoid the pain of publicly releasing an AppImage that could immediately be found to be unlaunchable by scores of users.

Now that I ran into the issue, I have subsequently discovered that it was previously reported as issue #7635 

I don't want to "spam" by posting my assessment in 2 places. Therefore, I will kindly request that anyone reading this take a look at what I said on the issue: https://github.com/mavlink/qgroundcontrol/issues/7635#issuecomment-744906359

### DIFF
--- a/deploy/create_linux_appimage.sh
+++ b/deploy/create_linux_appimage.sh
@@ -1,6 +1,6 @@
-#!/bin/bash -x
+#!/bin/bash
 
-#set +e
+set -Eeuxo pipefail # https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/
 
 if [[ $# -eq 0 ]]; then
   echo 'create_linux_appimage.sh QGC_SRC_DIR QGC_RELEASE_DIR'
@@ -41,8 +41,6 @@ APPDIR=${TMPDIR}/$APP".AppDir"
 mkdir -p ${APPDIR}
 
 cd ${TMPDIR}
-wget -c --quiet http://ftp.us.debian.org/debian/pool/main/u/udev/udev_175-7.2_amd64.deb
-wget -c --quiet http://ftp.us.debian.org/debian/pool/main/s/speech-dispatcher/speech-dispatcher_0.8.8-1_amd64.deb
 wget -c --quiet http://ftp.us.debian.org/debian/pool/main/libs/libsdl2/libsdl2-2.0-0_2.0.2%2bdfsg1-6_amd64.deb
 
 cd ${APPDIR}
@@ -56,12 +54,6 @@ dpkg -x libdirectfb-1.2-9_1.2.10.0-5.1_amd64.deb libdirectfb
 cp -L libdirectfb/usr/lib/x86_64-linux-gnu/libdirectfb-1.2.so.9 ${APPDIR}/usr/lib/x86_64-linux-gnu/
 cp -L libdirectfb/usr/lib/x86_64-linux-gnu/libfusion-1.2.so.9 ${APPDIR}/usr/lib/x86_64-linux-gnu/
 cp -L libdirectfb/usr/lib/x86_64-linux-gnu/libdirect-1.2.so.9 ${APPDIR}/usr/lib/x86_64-linux-gnu/
-
-# copy libts-0.0-0
-wget -c --quiet http://ftp.us.debian.org/debian/pool/main/t/tslib/libts-0.0-0_1.0-11_amd64.deb
-mkdir libts
-dpkg -x libts-0.0-0_1.0-11_amd64.deb libts
-cp -L libts/usr/lib/x86_64-linux-gnu/libts-0.0.so.0 ${APPDIR}/usr/lib/x86_64-linux-gnu/
 
 # copy QGroundControl release into appimage
 rsync -av --exclude=*.cpp --exclude=*.h --exclude=*.o --exclude="CMake*" --exclude="*.cmake" ${QGC_RELEASE_DIR}/* ${APPDIR}/


deploy/create_linux_appimage.sh
```diff
@@ -58,7 +56,7 @@ cp -L libdirectfb/usr/lib/x86_64-linux-gnu/libfusion-1.2.so.9 ${APPDIR}/usr/lib/
 cp -L libdirectfb/usr/lib/x86_64-linux-gnu/libdirect-1.2.so.9 ${APPDIR}/usr/lib/x86_64-linux-gnu/
 
 # copy libts-0.0-0
-wget -c --quiet http://ftp.us.debian.org/debian/pool/main/t/tslib/libts-0.0-0_1.0-11_amd64.deb
+wget -c --quiet http://old-releases.ubuntu.com/ubuntu/pool/universe/t/tslib/libts-0.0-0_1.0-11_amd64.deb
```
DonLakeFlyer, line 59: I wonder if this should be removed. From the Issue comments if I remember it has something to do with touchscreens. Not sure why it would be QGCs job to install that. If a Linux user was using a touchscreen wouldn't they need to install this themselves to get it to work generically?

pestophagous: The following is an excerpt of `apt-cache show libts-dev`:

Homepage: http://tslib.org
Description-en: touch screen library, development files
 Tslib is an abstraction layer for touchscreen panel events, as well as a
 filter stack for the manipulation of those events.
 
 So, yes, it is specific to touch screens.
 
 Regarding the idea of just removing that line from `create_linux_appimage.sh`, I would tend to agree.
 
 Since no QGroundcontrol AppImage has actually contained libts for many releases, that seems like enough evidence that it wasn't really needed.  I would tend to say "remove it and wait til anyone complains."
 
 Corollary: if someone does complain and some variation of this `wget libts` gets re-added to the script, then at least you can then capture better documentation about the rationale.
 
 It seems that libts was added back in 2016, in case this helps jog anyone's memory: https://github.com/mavlink/qgroundcontrol/commit/78d203bedf758462

DonLakeFlyer: Can you remove then. 

Also if I remember, udev is maybe related to flashing. Can you verify that flashing still works on a machine without it? Sorry but Linux is a foreign animal to me so I'm not super helpful on understanding the exact details of why these things may be needed. My concern would be that most Linux users may also have installed the PX4 firmware build pre-reqs which I think installs udev (which is how my VM is set up). So for them it may work. But for someone with a clean Linux it may not.

pestophagous: I will remove the libts line right away.

Regarding the rest:

quoting:

```
libudev.h provides APIs to introspect and enumerate devices on the local system.
```

from here: https://www.freedesktop.org/software/systemd/man/libudev.html

Labeling *my* weaknesses: Windows is a (mostly) foreign animal to *me*, so I am 100% empathetic to your situation. I'm happy to work in a bi-partisan way with you "across the OS aisle" ;) 

I think the odds of libudev not being on a modern machine used for "Desktop" usage is very low.

I just checked the reverse dependencies (things that depend on libudev) using `apt-cache rdepends libudev1`, and I see that all these would bring udev with them (or else not work):

```
...
  network-manager
  libudev-dev
  libnm0
  libnm-glib4
  libgnome-desktop-3-17
  libgnome-bluetooth13
  libd3dadapter9-mesa-dev
  libd3dadapter9-mesa
  initramfs-tools-bin
  xserver-xorg-core
  pulseaudio
  bluez
  xserver-xorg-input-xwiimote
  xserver-xorg-input-evdev
  vlc-plugin-base
  ubuntu-core-libs
  scanbd
  ring-daemon
  libgnome-desktop-3-17
  libgnome-bluetooth13
  libdevmapper1.02.1
  libclutter-1.0-0
...
```

... among many others.

I think there need **not** be concern for libudev being missing. However, the other feature of using an `AppImage` (as you know) is to include a particular **version** of the library.  So I suppose that could still be an issue.

After I opened this pull request, I subsequently read through a small handful of bug reports on QGroundcontrol related to gstreamer, and to the discussion/debate surrounding inclusion of gstreamer in the AppImage.

I would recommend that any concerns about libudev be bundled into the general debate on the AppImage issues. At some point it seems clear that someone will need to "own" and sort through all of that.

For this (small) PR, I'm merely suggesting to remove lines from a script where all the involved lines have failed for some time.  

Whether (and how) to bundle these failing-for-some-time libs in the future seems arguably out of scope. (I could argue it both ways, though.)

DonLakeFlyer: I just checked a virgin Ubuntu 18.04 and it has udev installed in it. So doesn't seem to be needed with that either. So seems perfectly fine to not include.

PR #9266, opened 2020-12-21
UX improvements around MAVLink Console

bkueng:
image

Changes:
- revolutionary feature: user can select & copy output (to be fair there is a caveat, see below)
- auto-scrolling: update is paused when scrolling up, any key input will scroll to the bottom
- feel is more like a shell w/o extra text field.
  (on mobile there's still a separate text field as input is handled differently)
- handle multi-line commands from clipboard
- coloring of ERROR's and WARNing's
- set text focus to shell input on init
- some fixes around cursor + newline handling (see commits)

### Performance
The use of a `TextArea` is quite slow when the output grows. This is mitigated by:
- limit update rate to 30hz
- limit history to 500 lines

With this I'm seeing a CPU load increase for QGC from idle ~22% to ~28% with continous updates (running `top`), vs. 22% to 24% on master.
The only alternative to getting this faster I see is to do a custom QML element in C++.

Tested on Linux and Android. Would be good to give a quick test on MacOS and Windows as well (I don't have the setup).

### DIFF
--- a/src/AnalyzeView/MavlinkConsoleController.cc
+++ b/src/AnalyzeView/MavlinkConsoleController.cc
@@ -256,6 +256,15 @@ MavlinkConsoleController::writeLine(int line, const QByteArray &text)
     auto rc = rowCount();
     if (line >= rc) {
         insertRows(rc, 1 + line - rc);
+        if (rowCount() > _max_num_lines) {
+            int count = rowCount() - _max_num_lines;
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
I tested on Window where it worked fine. Can't get my Mac build to work anymore. I'll try it once I figure it out. For now I'm guessing this is ok.

PR #9267, opened 2020-12-21
Fix Segfault on vehicle reboot

### LINKED ISSUE #9260: Segfault on vehicle reboot on master
Happened to me once during a vehicle reboot connected via USB. I didn't try much to reproduce.
```
Temp log "/tmp/FlightData893795.mavlink"
LinkManager::sharedLinkInterfaceForLink returning nullptr
./staging/qgroundcontrol-start.sh: line 10: 50355 Segmentation fault      (core dumped) "${HERE}/QGroundControl" "$@"
```

bkueng:
When rebooting the vehicle from param editor, the vehicle would call `VehicleLinkManager::closeVehicle()` from `Vehicle::_mavlinkMessageReceived`. The link was then removed, and deleted as the ref count went to 0. If `MAVLinkProtocol::receiveBytes` was then parsing further messages, the same link was accessed again, leading to invalid memory access.
This PR fixes it by checking the validity of the link object after message handling. It could be solved differently, e.g. by closing the connection outside of the `MAVLinkProtocol::receiveBytes` call.

See individual commits for details.

Fixes #9260

Debugging output:
```
VehicleLinkManagerLog: _removeLink: "5b59370"
LinkManagerLog: LinkManager::_linkDisconnected "PX4 FMU V5 on ttyACM0 (AutoConnect)" 3

Thread 1 "QGroundControl" received signal SIGSEGV, Segmentation fault.
```
With the stack trace:
```
#0  0x000000000044d05c in QString::QString(QString const&) (this=0x7fffffffbd58, other=...)
    at /opt/install/qt/5.12.5/gcc_64/include/QtCore/qstring.h:957
#1  0x000000000054d08f in LinkConfiguration::name() const (this=0x0)
    at /home/beat/px4/src/qgroundcontrol/src/comm/LinkConfiguration.h:39
#2  0x00000000007358fb in VehicleLinkManager::_addLink(LinkInterface*)
    (this=0x4f899e0, link=0x5b59370)
    at /home/beat/px4/src/qgroundcontrol/src/Vehicle/VehicleLinkManager.cc:174
#3  0x00000000007345b8 in VehicleLinkManager::mavlinkMessageReceived(LinkInterface*, __mavlink_message) (this=0x4f899e0, link=0x5b59370, message=...)
    at /home/beat/px4/src/qgroundcontrol/src/Vehicle/VehicleLinkManager.cc:36
#4  0x000000000070dd96 in Vehicle::_mavlinkMessageReceived(LinkInterface*, __mavlink_message)
    (this=0x55367a0, link=0x5b59370, message=...)
    at /home/beat/px4/src/qgroundcontrol/src/Vehicle/Vehicle.cc:562
#5  0x0000000000727e28 in QtPrivate::FunctorCall<QtPrivate::IndexesList<0, 1>, QtPrivate::List<LinkInterface*, __mavlink_message>, void, void (Vehicle::*)(LinkInterface*, __mavlink_message)>::call(void (Vehicle::*)(LinkInterface*, __mavlink_message), Vehicle*, void**) (f=
    (void (Vehicle::*)(Vehicle * const, LinkInterface *, __mavlink_message)) 0x70dbd0 <Vehicle::_mavlinkMessageReceived(LinkInterface*, __mavlink_message)>, o=0x55367a0, arg=0x7fffffffc340)
    at /opt/install/qt/5.12.5/gcc_64/include/QtCore/qobjectdefs_impl.h:152
#6  0x0000000000726865 in QtPrivate::FunctionPointer<void (Vehicle::*)(LinkInterface*, __mavlink_message)>::call<QtPrivate::List<LinkInterface*, __mavlink_message>, void>(void (Vehicle::*)(LinkInterface*, __mavlink_message), Vehicle*, void**) (f=
    (void (Vehicle::*)(Vehicle * const, LinkInterface *, __mavlink_message)) 0x70dbd0 <Vehicle::_mavlinkMessageReceived(LinkInterface*, __mavlink_message)>, o=0x55367a0, arg=0x7fffffffc340)
    at /opt/install/qt/5.12.5/gcc_64/include/QtCore/qobjectdefs_impl.h:185
#7  0x0000000000724a20 in QtPrivate::QSlotObject<void (Vehicle::*)(LinkInterface*, __mavlink_message), QtPrivate::List<LinkInterface*, __mavlink_message>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*)
    (which=1, this_=0x65a2bf0, r=0x55367a0, a=0x7fffffffc340, ret=0x0)
    at /opt/install/qt/5.12.5/gcc_64/include/QtCore/qobjectdefs_impl.h:414
#8  0x00007ffff2873df6 in QMetaObject::activate(QObject*, int, int, void**) ()
    at /opt/install/qt/5.12.5/gcc_64/lib/libQt5Core.so.5
#9  0x000000000095eb85 in MAVLinkProtocol::messageReceived(LinkInterface*, __mavlink_message)
    (this=0x7ffff7fa9010, _t1=0x5b59370, _t2=...) at moc_MAVLinkProtocol.cpp:346
#10 0x00000000007583a4 in MAVLinkProtocol::receiveBytes(LinkInterface*, QByteArray)
```

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -567,6 +567,7 @@ void LinkManager::_updateAutoConnectLinks(void)
                         pSerialConfig->setBaud      (boardType == QGCSerialPortInfo::BoardTypeSiKRadio ? 57600 : 115200);
                         pSerialConfig->setDynamic   (true);
                         pSerialConfig->setPortName  (portInfo.systemLocation());
+                        pSerialConfig->setAutoConnect(true);
 
                         SharedLinkConfigurationPtr  sharedConfig(pSerialConfig);
                         createConnectedLink(sharedConfig, boardType == QGCSerialPortInfo::BoardTypePX4Flow);
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -103,40 +103,38 @@ void LinkManager::createConnectedLink(LinkConfiguration* config)
 
 bool LinkManager::createConnectedLink(SharedLinkConfigurationPtr& config, bool isPX4Flow)
 {
-    LinkInterface* link = nullptr;
+    SharedLinkInterfacePtr link = nullptr;
 
     switch(config->type()) {
 #ifndef NO_SERIAL_LINK
     case LinkConfiguration::TypeSerial:
-        link = new SerialLink(config, isPX4Flow);
+        link = std::make_shared<SerialLink>(config, isPX4Flow);
         break;
 #else
     Q_UNUSED(isPX4Flow)
 #endif
     case LinkConfiguration::TypeUdp:
-        link = new UDPLink(config);
+        link = std::make_shared<UDPLink>(config);
         break;
     case LinkConfiguration::TypeTcp:
-        link = new TCPLink(config);
+        link = std::make_shared<TCPLink>(config);
         break;
 #ifdef QGC_ENABLE_BLUETOOTH
     case LinkConfiguration::TypeBluetooth:
-        link = new BluetoothLink(config);
+        link = std::make_shared<BluetoothLink>(config);
         break;
 #endif
     case LinkConfiguration::TypeLogReplay:
-        link = new LogReplayLink(config);
+        link = std::make_shared<LogReplayLink>(config);
         break;
 #ifdef QT_DEBUG
     case LinkConfiguration::TypeMock:
-        link = new MockLink(config);
+        link = std::make_shared<MockLink>(config);
         break;
 #endif
     case LinkConfiguration::TypeLast:
         break;
     }
-
-    QScopedPointer<LinkInterface> scopedLink(link);
 
     if (link) {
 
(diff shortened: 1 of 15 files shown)


src/comm/LinkManager.cc
```diff
@@ -221,7 +220,7 @@ SharedLinkInterfacePtr LinkManager::sharedLinkInterfacePointerForLink(LinkInterf
         }
     }
 
-    qWarning() << "LinkManager::sharedLinkInterfaceForLink returning nullptr";
```
DonLakeFlyer, line 224: This should really stay as a qWarning since it indicates an internal error. And qWarning will also fail units tests if it hits. Super useful to have it pop all the time not behind a category flag just like it popped for this crash.

src/comm/MAVLinkProtocol.cc
```diff
@@ -197,7 +197,8 @@ void MAVLinkProtocol::receiveBytes(LinkInterface* link, QByteArray b)
     // Since receiveBytes signals cross threads we can end up with signals in the queue
     // that come through after the link is disconnected. For these we just drop the data
     // since the link is closed.
-    if (!_linkMgr->containsLink(link)) {
```
DonLakeFlyer, line 200: If you keep both the containsLinks and the weak pointer check then I believe the null ptr return from LinkManager can stay a qWarning.

src/comm/LinkManager.cc
```diff
@@ -103,41 +103,39 @@ void LinkManager::createConnectedLink(LinkConfiguration* config)
...
 bool LinkManager::createConnectedLink(SharedLinkConfigurationPtr& config, bool isPX4Flow)
 {
-    LinkInterface* link = nullptr;
+    SharedLinkInterfacePtr link = nullptr;
 
     switch(config->type()) {
 #ifndef NO_SERIAL_LINK
     case LinkConfiguration::TypeSerial:
-        link = new SerialLink(config, isPX4Flow);
+        link = std::make_shared<SerialLink>(config, isPX4Flow);
```
DonLakeFlyer, line 111: I assume this causes the constructors needing to be public. Kind of a bummer since nobody else should ever create these. But not the end of the world.  With the link constructors going public then I believe the friend class to LinkManager in all the Links can go away as well. Can you remove those. I'm 90% sure that's the only reason they are friended togehter.

src/comm/MAVLinkProtocol.cc
```diff
@@ -362,6 +363,10 @@ void MAVLinkProtocol::receiveBytes(LinkInterface* link, QByteArray b)
             // Reset message parsing
             memset(&_status,  0, sizeof(_status));
             memset(&_message, 0, sizeof(_message));
+
+            if (linkPtr.expired()) {
```
DonLakeFlyer, line 367: Can this move up to be right after the emit so that nothing else can sneak in between? Also with a comment as to why it's needed.

src/comm/MAVLinkProtocol.cc
bkueng: I'm not a fan 2 calls or iterating the list twice (even though it's a small one), so I added a flag.

src/comm/MAVLinkProtocol.cc
bkueng: yup, done.

src/comm/LinkManager.cc
bkueng: Yes that's the downside. You could work around it using factory methods, but that didn't really look appealing to me.
Indeed they can be removed.

DonLakeFlyer:
Thanks @bkueng

PR #9269, opened 2020-12-23
main: Fix mac build when running Qt 5.15

patrickelectric:
QProces::execute(QString) was deprecated, QProcess::execute(QString, QStringList) should be used


jaxxzer:
LGTM

PR #9274, opened 2020-12-26
Add indexes to prevent Query Blocking with full search for Big offline map.

evnngsky:
If TileDB is bigger than 2GB, 
--- sq = QString("SELECT COUNT(size), SUM(size) FROM Tiles WHERE tileID IN (SELECT A.tileID FROM SetTiles A join SetTiles B on A.tileID = B.tileID WHERE B.setID  ---
This query holds so long time QGCTileCache Worker.
I solved my problems with adding an index like that.


LorenzMeier:
Makes sense and thanks!

LorenzMeier (approved):

DonLakeFlyer:
Same thing here. This was waiting for post stable so it could get some bake time on it before release.

DonLakeFlyer:
This pull is also incorrect in that there is no error checking on the index creation.

PR #9283, opened 2020-12-29
PX4 Config: Horizon level fix

LorenzMeier:
Leveling the horizon is not a requirement and can cause more harm than good. We are defaulting now to a green indicator to suggest to average users that they do not need to do anything.

There are likely more fixes coming, so I've kept this as draft.


DonLakeFlyer:
That would seem to say that the indicator for it should just be removed completely if it will never go red. Correct?

LorenzMeier:
I opted for consistency with the remaining ones, but yes, that is a valid conclusion / position as well.

DonLakeFlyer:
Replaced

PR #9289, opened 2020-12-31
Fix up GST Debug Level UI

DonLakeFlyer:
* Use names instead of numbers
* Set the qgc reboot required flag
* Note: I only added 0-7 level. Previous code had level 8 in there but I can't find anywhere that level 8 actually exists and what it is.

Screen Shot 2020-12-30 at 4 13 22 PM


DonLakeFlyer:
I added another change to this to prevent search controls from being clipped:
Screen Shot 2020-12-30 at 4 25 51 PM

PR #9291, opened 2020-12-31
Move to Qt 5.12.10, change to 7zip for Qt min zip format

DonLakeFlyer:
* Working towards moving Stable to use latest Qt 5.12 point release which is 5.12.10 instead of 5.12.6 which is an odd Qt release and hard to get to
* Format for zipped Qt min packages on S3 is moving to 7zip. This is because 7zip is what is needed for Windows. Moving everything to 7zip standardizes things
* Create new shell scripts which will build these min zip packages from a standard Qt install


DonLakeFlyer:
The main reason to move from 5.12.6 to 5.12.10 is that I'm going to change the build to FAIL if the EXACT Qt version isn't being used. I'm tired of dealing with answering questions from people using the wrong Qt over and over (and over) again. I'll add a CONFIG flag backdoor you can use to disable if you really mean to do this.

PR #9305, opened 2021-01-05
Fix channel indexing bug in Flight Modes page for Ardupilot

dakejahl:
This fixes an off by 1 error in the Flight Modes page in Ardupilot. It's obvious once you see it, but took me forever to figure out. The resultant effect was every channel was off by 1, so when channel 8 was active channel 7 would light up yellow.

image

### DIFF
--- a/src/AutoPilotPlugins/APM/APMFlightModesComponentController.cc
+++ b/src/AutoPilotPlugins/APM/APMFlightModesComponentController.cc
@@ -107,7 +107,7 @@ void APMFlightModesComponentController::_rcChannelsChanged(int channelCount, int
 
     for (int i=0; i<_cChannelOptions; i++) {
         _rgChannelOptionEnabled[i] = QVariant(false);
-        channelValue = pwmValues[i+6];
+        channelValue = pwmValues[i+5];
         if (channelValue > 1800) {
             _rgChannelOptionEnabled[i] = QVariant(true);
         }


DonLakeFlyer:
This isn't correct. I believe the bug is here:
```
    property bool   _ch7OptAvailable:           controller.parameterExists(-1, "CH7_OPT")
    property int    _rcOptionStart:             _ch7OptAvailable ? 7 : 6
    property int    _rcOptionStop:              _ch7OptAvailable ? 12 : 16
```

Should be:
```
    property bool   _ch7OptAvailable:           controller.parameterExists(-1, "r.RC7_OPTION")
```

dakejahl:
I thought the same thing initially. but what that check is actually for is for Arducopter 3.6 and earlier, only channel options 7 thru 12 were available, see https://ardupilot.org/copter/docs/channel-7-and-8-options.html

This fix is indeed correct, digest what this is doing
```
    for (int i=0; i<_cChannelOptions; i++) {
        _rgChannelOptionEnabled[i] = QVariant(false);
        channelValue = pwmValues[i+5];
        if (channelValue > 1800) {
            _rgChannelOptionEnabled[i] = QVariant(true);
        }
    }
```

The array `_rgChannelOptionEnabled` is simply an array of booleans which indicates if a channel option is enabled (for arducopter versions greater than 3.6, these are options 6 thru 16). Channel 6 corresponds in index 5 in `pwmValues`.

dakejahl:
Perhaps we remove support for Ardupilot version 3.6 and below? Last v3 release was Dec 2019

DonLakeFlyer:
I was wrong. I believe the bug fix is actually here which is incorrectly indexing into the enabled array:
```
                                    color:              controller.channelOptionEnabled[modelData + (_ch7OptAvailable ? 1 : 0)] ? "yellow" : qgcPal.text
```

DonLakeFlyer:
That said, the fact that the min version for ArduCopter in QGC is 3.6 likely should be updated which would get rid of all this insanity. But it's too late to do that I think for stable now. And should be cleaned up once this stable goes out and the min supported version in QGC bumped up to something less ancient to simplify.

dakejahl:
No that's not right either.

This line right here is looking at the channel PWM value, and if it's "selected" (greater than 1800) then the index in `_rgChannelOptionEnabled` is set to true. Index zero of this array is either channel 6 (APM v4 and higher) or channel 7 (APM v3.6 and below), hence why it should be `+5` for v4 and higher (starting at channel 6) and `+6` for v3.6 and below (starting at channel 7)

```
    for (int i=0; i<_cChannelOptions; i++) {
        _rgChannelOptionEnabled[i] = QVariant(false);
        channelValue = pwmValues[i+6];
        if (channelValue > 1800) {
            _rgChannelOptionEnabled[i] = QVariant(true);
        }
    }
```

DonLakeFlyer:
I missed something. The C code is wrong as well. Here it the full fix:
```
diff --git a/src/AutoPilotPlugins/APM/APMFlightModesComponent.qml b/src/AutoPilotPlugins/APM/APMFlightModesComponent.qml
index d4fef1ef2..11b3f4069 100644
--- a/src/AutoPilotPlugins/APM/APMFlightModesComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMFlightModesComponent.qml
@@ -211,7 +211,7 @@ SetupPage {
                                 QGCLabel {
                                     anchors.baseline:   optCombo.baseline
                                     text:               qsTr("Channel option %1 :").arg(index)
-                                    color:              controller.channelOptionEnabled[modelData] ? "yellow" : qgcPal.text
+                                    color:              controller.channelOptionEnabled[modelData + (_ch7OptAvailable ? 1 : 0)] ? "yellow" : qgcPal.text
                                 }
 
                                 FactComboBox {
diff --git a/src/AutoPilotPlugins/APM/APMFlightModesComponentController.cc b/src/AutoPilotPlugins/APM/APMFlightModesComponentController.cc
index 425bd3ed0..4e251a5b2 100644
--- a/src/AutoPilotPlugins/APM/APMFlightModesComponentController.cc
+++ b/src/AutoPilotPlugins/APM/APMFlightModesComponentController.cc
@@ -107,7 +107,7 @@ void APMFlightModesComponentController::_rcChannelsChanged(iparapaparapparapparpppappappppparallels@parallels:~/repos/qgroundcontrol$ git diff
diff --git a/src/AutoPilotPlugins/APM/APMFlightModesComponent.qml b/src/AutoPilotPlugins/APM/APMFlightModesComponent.qml
index d4fef1ef2..11b3f4069 100644
--- a/src/AutoPilotPlugins/APM/APMFlightModesComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMFlightModesComponent.qml
@@ -211,7 +211,7 @@ SetupPage {
                                 QGCLabel {
                                     anchors.baseline:   optCombo.baseline
                                     text:               qsTr("Channel option %1 :").arg(index)
-                                    color:              controller.channelOptionEnabled[modelData] ? "yellow" : qgcPal.text
+                                    color:              controller.channelOptionEnabled[modelData + (_ch7OptAvailable ? 1 : 0)] ? "yellow" : qgcPal.text
                                 }
 
                                 FactComboBox {
diff --git a/src/AutoPilotPlugins/APM/APMFlightModesComponentController.cc b/src/AutoPilotPlugins/APM/APMFlightModesComponentController.cc
index 425bd3ed0..4e251a5b2 100644
--- a/src/AutoPilotPlugins/APM/APMFlightModesComponentController.cc
+++ b/src/AutoPilotPlugins/APM/APMFlightModesComponentController.cc
@@ -107,7 +107,7 @@ void APMFlightModesComponentController::_rcChannelsChanged(int channelCount, int
 
     for (int i=0; i<_cChannelOptions; i++) {
         _rgChannelOptionEnabled[i] = QVariant(false);
-        channelValue = pwmValues[i+6];
+        channelValue = pwmValues[i+5];
         if (channelValue > 1800) {
             _rgChannelOptionEnabled[i] = QVariant(true);
         }
```

The whole thing is freakin horrendous!

DonLakeFlyer:
I wanted to get this in before I marked stable so I create a replacement. Also found another bug with channel 16. Tested on SITL to be working correctly.

PR #9307, opened 2021-01-06
Create field for manual Magnetic Declination setup for Submarines

### LINKED ISSUE #7310: Feature: Add declination setup
Moved from: bluerobotics/qgroundcontrol#143
Original author: @jaxxzer

### Issue description
The autopilot has a declination lookup, but it won't use it without gps lock. We should have manual declination entry on sensors/compass setup.

Williangalvani:
This introduces a Magnetic Declination field for Subs:
declination

Two possible discussion points:

 - ~Should we show it as degrees instead of radians ? Is something already there in the fact system that could help with the conversion?~
 - Having the checkbox inverted (user clicks "Manual declination" instead of "Automatic Declination") makes more sense from a user perspective. This could be done by adding an "inverted" property to FactCheckBox, but [this binding](https://github.com/mavlink/qgroundcontrol/blob/master/src/FactSystem/FactControls/FactCheckBox.qml#L18) would double in size.
 
Fix #7310

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSensorsComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMSensorsComponent.qml
@@ -446,6 +446,37 @@ SetupPage {
                                 model:      _orientationsDialogShowCompass ? 3 : 0
                                 delegate:   singleCompassSettingsComponent
                             }
+
+                            QGCLabel {
+                                id:         magneticDeclinationLabel
+                                width:      parent.width
+                                visible:    globals.activeVehicle.sub && _orientationsDialogShowCompass
+                                text:       qsTr("Magnetic Declination")
+                            }
+
+                            Column {
+                                visible:            magneticDeclinationLabel.visible
+                                anchors.margins:    ScreenTools.defaultFontPixelWidth
+                                anchors.left:       parent.left
+                                anchors.right:      parent.right
+                                spacing:            ScreenTools.defaultFontPixelHeight
+
+                                QGCCheckBox {
+                                    id:                           manualMagneticDeclinationCheckBox
+                                    text:                         qsTr("Manual Magnetic Declination")
+                                    property Fact autoDecFact:    controller.getParameterFact(-1, "COMPASS_AUTODEC")
+                                    property int manual:          0
+                                    property int automatic:       1
+
+                                    checked:    autoDecFact.rawValue === manual
+                                    onClicked:  autoDecFact.value = (checked ? manual : automatic)
+                                }
+
+                                FactTextField {
+                                    fact:       sensorParams.declinationFact
+                                    enabled:    manualMagneticDeclinationCheckBox.checked
+                                }
+                            }
                         } // Column
                     } // QGCFlickable
                 } // QGCViewDialog
--- a/src/FirmwarePlugin/APM/APMSensorParams.qml
+++ b/src/FirmwarePlugin/APM/APMSensorParams.qml
@@ -92,4 +92,6 @@ Item {
     property bool compass2Calibrated:               compass2Available ? compass2OfsXFact.value != 0.0  && compass2OfsYFact.value != 0.0  &&compass2OfsZFact.value != 0.0 : false
     property bool compass3Calibrated:               compass3Available ? compass3OfsXFact.value != 0.0  && compass3OfsYFact.value != 0.0  &&compass3OfsZFact.value != 0.0 : false
     property var  rgCompassCalibrated:              [ compass1Calibrated, compass2Calibrated, compass3Calibrated ]
+
+    property Fact declinationFact:                  factPanelController.getParameterFact(-1, "COMPASS_DEC")
 }
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -57,6 +57,7 @@ const char* FactMetaData::_enumBitmaskArrayDescriptionJsonKey   = "description";
 const FactMetaData::BuiltInTranslation_s FactMetaData::_rgBuiltInTranslations[] = {
     { "centi-degrees",  "deg",  FactMetaData::_centiDegreesToDegrees,                   FactMetaData::_degreesToCentiDegrees },
     { "radians",        "deg",  FactMetaData::_radiansToDegrees,                        FactMetaData::_degreesToRadians },
+    { "rad",            "deg",  FactMetaData::_radiansToDegrees,                        FactMetaData::_degreesToRadians },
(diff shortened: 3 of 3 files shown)


DonLakeFlyer:
Fact system already has radians to degrees conversion:
```
// Built in translations for all Facts
const FactMetaData::BuiltInTranslation_s FactMetaData::_rgBuiltInTranslations[] = {
    { "centi-degrees",  "deg",  FactMetaData::_centiDegreesToDegrees,                   FactMetaData::_degreesToCentiDegrees },
    { "radians",        "deg",  FactMetaData::_radiansToDegrees,                        FactMetaData::_degreesToRadians },
    { "gimbal-degrees", "deg",  FactMetaData::_mavlinkGimbalDegreesToUserGimbalDegrees, FactMetaData::_userGimbalDegreesToMavlinkGimbalDegrees },
    { "norm",           "%",    FactMetaData::_normToPercent,                           FactMetaData::_percentToNorm },
};
```

You can make the checked/unchecked values for a FactCheckBox be anything you want by using the `FactCheckBox.checkedValue` and `FactCheckBox.uncheckedValue` properties. There is no need for a special "invert" mechanism.

src/AutoPilotPlugins/APM/APMSensorsComponent.qml
```diff
@@ -446,6 +446,34 @@ SetupPage {
                                 model:      _orientationsDialogShowCompass ? 3 : 0
                                 delegate:   singleCompassSettingsComponent
                             }
+
+                            QGCLabel {
+                                width:      parent.width
+                                visible:    globals.activeVehicle.sub && _orientationsDialogShowCompass
```
DonLakeFlyer, line 452: The best way to handle visibility for these sort of things is to have the logic only in one place. Either with a single _showFoo property or with say an id on the label `magneticDeclinationLabel' and then the `Column` uses ``magneticDeclinationLabel.visible` to determine visibility. This prevent two separate sets of visibility logic from getting out of sync and causing bugs.

src/FirmwarePlugin/APM/APMSensorParams.qml
```diff
@@ -92,4 +92,7 @@ Item {
     property bool compass2Calibrated:               compass2Available ? compass2OfsXFact.value != 0.0  && compass2OfsYFact.value != 0.0  &&compass2OfsZFact.value != 0.0 : false
     property bool compass3Calibrated:               compass3Available ? compass3OfsXFact.value != 0.0  && compass3OfsYFact.value != 0.0  &&compass3OfsZFact.value != 0.0 : false
     property var  rgCompassCalibrated:              [ compass1Calibrated, compass2Calibrated, compass3Calibrated ]
+
+    property Fact autoDeclinationFact:              factPanelController.getParameterFact(-1, "COMPASS_AUTODEC")
```
DonLakeFlyer, line 96: Are these parameters available in all versions of ArduPilot firmware going all the way back to the min supported ArduPilot version for each vehicle type. If not, referencing them like this will cause the parameters missing error to pop depending on the firmware you are using.

src/AutoPilotPlugins/APM/APMSensorsComponent.qml
```diff
@@ -446,6 +446,34 @@ SetupPage {
...
+                                anchors.margins:    ScreenTools.defaultFontPixelWidth
+                                anchors.left:       parent.left
+                                anchors.right:      parent.right
+                                spacing:            ScreenTools.defaultFontPixelHeight
+
+                                FactCheckBox {
+                                    id:         manualMagneticDeclination
+                                    text:       qsTr("Automatic Declination")
+                                    fact:       sensorParams.autoDeclinationFact
+                                    visible:    true
```
DonLakeFlyer, line 469: There is no need for a visible property here.

src/AutoPilotPlugins/APM/APMSensorsComponent.qml
```diff
@@ -446,6 +446,34 @@ SetupPage {
...
+                                }
+
+                                FactTextField {
+                                    fact:       sensorParams.declinationFact
+                                    visible:    !sensorParams.autoDeclinationFact.value
```
DonLakeFlyer, line 474: The pattern for things like this more follows disabling if the controlling checkbox is in a state. Making things come/go visually can be more confusing to folks.

src/AutoPilotPlugins/APM/APMSensorsComponent.qml
Williangalvani: ok, done!

src/AutoPilotPlugins/APM/APMSensorsComponent.qml
Williangalvani: ok, fixed

Williangalvani:
I had to add "rad" to the translations table as Ardupilot is using that instead of "radians" (see https://github.com/ArduPilot/ardupilot/commit/4c45b1c7ca893b54f2ac3204a7adddeb0238af15).

The issue with 
```
uncheckedValue: 1
checkedValue: 0
```
Is that the current binding is checking for literals `0` and `false`. This meant that when I checked the box, it made the fact 0, and then the binding unticked the box.

I had to change FactCheckBox.qml so it compares with `uncheckedValue` instead. I also changed the compare to `==` so we don't need to check if it is a boolean or not.

src/FirmwarePlugin/APM/APMSensorParams.qml
Williangalvani: This parameter is over 9 years old. It works fine on latest releases, but I couldn't find the "min supported ardupilot version for each vehicle type" information anywhere...

I'll test the oldest versions we have metadata for, to be sure

DonLakeFlyer: > but I couldn't find the "min supported ardupilot version for each vehicle type" information anywhere...

Forgot that I removed that code this release.

> This parameter is over 9 years old

This is fine then.

DonLakeFlyer:
> Is that the current binding is checking for literals `0` and `false`. This meant that when I checked the box, it made the fact 0, and then the binding unticked the box.

Oh crap, this is a bool. You are correct. Instead of futzing with FactCheckBox which is dangerous at this point instead just make your own control work like you want using QGCCheckBox instead.

DonLakeFlyer:
Also lastly can you squash it down to a single commit with a good message

DonLakeFlyer:
> Also lastly can you squash it down to a single commit with a good message

Actually the only problem is the two commits to APMSensorsComponent. The separation of the rest makes sense.

Williangalvani:
I had extra commits because I has considering dropping the changes to FactCheckBox and inverting the control.

I took out the change to FactCheckBox.qml and used a QGCCheckBox instead. What do you think about it?

DonLakeFlyer:
Perfect, thanks

jaxxzer:
@patrickelectric can we backport this to 4.1?

PR #9311, opened 2021-01-07
Fix APM PowerComponent to use the battery FactGroups

spark404:
The power screen in the APM vehicle setup didn't properly show the current voltage and current.

I'm not much of a Qt developer, but looking at other pieces of the code i found something that seems to fix the problem.

Please advice if this is the correct way to fix this. Happy to adjust the PR if needed.

Environment:
* Mac OS X Big Sur
* Drone running ArduCopter 4.0.5
* Single battery installed in Drone


DonLakeFlyer:
Perfect. Thank you so much for the bug fix pull.

zzaurak:
just tried this PR and it does not work with me, throws an error:
```
qrc:/qml/APMPowerComponent.qml:221: TypeError: Cannot read property 'current' of null
qrc:/qml/APMPowerComponent.qml:220: TypeError: Cannot read property 'voltage' of null
```
Env:
* Mac OS 10.15
* ArduPlane 4.0.6
* Single battery installed in Drone

DonLakeFlyer:
That is because you don't have two batteries. It won't prevent anything from working correctly. The errors are from the second battery. I'll create an issue to clean this up after stable goes out. This code needs to be rewritten anyway since ArduPilot supports more than two batteries.

zzaurak:
Okay thanks, for the clarification!

DonLakeFlyer:
@zzaurak And thanks for testing/reporting bugs. Great to have folks helping out.

PR #9313, opened 2021-01-08
update Gstreamer version in readme to 1.18.1

kemen209:
Fix doc issue of commit 6afd4178d46a181bc3bcfdf9197179b478c193fc  and 465836c4bcd5e61be1b9c9aa4220b53a6552adba

NOTE: Since there are no update to gstreamer in IOS, i'm keeping the doc of IOS section remain still.

### DIFF
--- a/src/VideoReceiver/README.md
+++ b/src/VideoReceiver/README.md
@@ -2,7 +2,7 @@
 
 ## Video Streaming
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
iOS will move to 1.18 as well with the stable that is currently being worked on. So that should be updated as well.

LorenzMeier (approved):
Thanks!

PR #9315, opened 2021-01-08
Fix video recording time when camera provides recording_time_ms

spark404:
QTime::msec only returns the msec component. Replaced with msecsTo to get the actual elapsed time in msec.

Environment:
* Mac OS X Big Sur
* Drone running ArduCopter 4.0.5
* Single battery installed in Drone
* Custom camera adapter in Basic Mavlink mode

### DIFF
--- a/src/Camera/QGCCameraControl.cc
+++ b/src/Camera/QGCCameraControl.cc
@@ -779,7 +779,7 @@ QGCCameraControl::_setVideoStatus(VideoStatus status)
 void
 QGCCameraControl::_recTimerHandler()
 {
-    _recordTime = static_cast<uint32_t>(_recTime.msec());
+    _recordTime = static_cast<uint32_t>(_recTime.msecsTo(QTime::currentTime()));
     emit recordTimeChanged();
 }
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Is this the fix for #9316?

spark404:
> Is this the fix for #9316?

I think so, i noticed the same behaviour.

DonLakeFlyer:
It seems correct, but I've asked @dogmaphobic to verify.

zzaurak:
i just tested this PR, and can confirm it fixes #9316

DonLakeFlyer:
@spark404 There were a number of these QTimer elapsed time changes done in preparation for moving to Qt 5.15. Would it be possible for you to do a global search to see if there are other cases in the codebase of this same problem?

spark404:
> @spark404 There were a number of these QTimer elapsed time changes done in preparation for moving to Qt 5.15. Would it be possible for you to do a global search to see if there are other cases in the codebase of this same problem?

Sure thing 👍

spark404:
I did a compile using 5.15.2 with `DEFINES += QT_DISABLE_DEPRECATED_BEFORE=0x051500`. There are a few issues:

QGCTileCacheWorker.cpp:multiple_issues
LogDownloadController.cc:178
QGCApplication.cc:433 
SerialLink.cc:180
SubtitleWriter.cc:153 

But none of them are related to the deprecation of start and elapsed in the QTime. Do I think we are good. The few issues that are time related have something to do with to and from Time_t being deprecated.

DonLakeFlyer:
Ok great, thanks for the fix

PR #9317, opened 2021-01-08
Make GStreamer debugging simpler/clearer from an end user support perspective

DonLakeFlyer:
* Created new VideoAllLog category which can be used to turn on all the individual video logging categories instead of having to do them one by one
* Separated logging from QGC GStreamer and GStreamer api debug level logging into their own categories. This way you can more easily tell what comes from where

So now customer support is something along these lines:
* Turn on VideoAllLog
* Set GStreamer debug output level to `Debug`

Trying to make QGC GStreamer support more easily remotely debuggable

### DIFF
--- a/src/QGCLoggingCategory.cc
+++ b/src/QGCLoggingCategory.cc
@@ -15,6 +15,8 @@
 
 #include <QSettings>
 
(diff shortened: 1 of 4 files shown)


jaxxzer:
nice thanks

PR #9321, opened 2021-01-10
fix warning C4334(lead to error 2220) on debug build with vs comipiler

kemen209:
Wonder if 'unsigned long long' is proper here, since quint64 may be defined to other type on other platform.

Maybe should use static_cast to convert the shift result ?


kemen209:
Fixed in commit: 946486c8e3f7cc4179f906b2c17ff7a5a5406026 ,so close this.

Issue #9186, opened 2020-11-20
Fix custom example code to take into account new telemetry values display


DonLakeFlyer:
Current custom example has it's own custom values display which doesn't really make sense any more.

PR #9323, opened 2021-01-10
Custom example update

### LINKED ISSUE #9314: updateqrc.py code python version
fallowing line throws error in python 3, it is an easy fix, and error mesage itself show how to fix it and there are online converters avaliable.
Should we change it, or should we mention it anywhere ?
Error :
```
 File "updateqrc.py", line 28
    print 'Excluded:', line
          ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print('Excluded:', line)?
```
line:
https://github.com/mavlink/qgroundcontrol/blob/44be3032b1f49ab0ea2dc540b522746eed7bf1b2/custom-example/updateqrc.py#L28
fix : 
`  print('Excluded:', line)`

DonLakeFlyer:
Fix #9314, #9186

### DIFF
--- a/custom-example/custom.qrc
+++ b/custom-example/custom.qrc
@@ -1,7 +1,4 @@
 <RCC>
-    <qresource prefix="/custom">
-        <file alias="CustomFlyViewOverlay.qml">res/CustomFlyViewOverlay.qml</file>
-    </qresource>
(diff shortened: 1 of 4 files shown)


LorenzMeier (approved):
Makes sense, nice individual commits!

PR #9329, opened 2021-01-13
linux_appimage: canonicalize release_dir argument

potaito:
When running the script and providing a local path as an argument for the release directory, the `rsync` command would fail, because `rsync` changes the directory and the relative path no longer applies. This only happens when `rsync` is ran inside the script. If I run the exact same command in the command line, it works. I'm not a bash ninja, so I don't know why. But canonicalizing the path fixes the problem :smile: 

- fixes errors when providing path arguments as local paths
- fixes error when output directory does not already exist

### DIFF
--- a/deploy/create_linux_appimage.sh
+++ b/deploy/create_linux_appimage.sh
@@ -88,5 +88,6 @@ chmod a+x ./appimagetool-x86_64.AppImage
 
 ./appimagetool-x86_64.AppImage ./$APP.AppDir/ ${TMPDIR}/$APP".AppImage"
 
+mkdir -p ${OUTPUT_DIR}
 cp ${TMPDIR}/$APP".AppImage" ${OUTPUT_DIR}/$APP".AppImage"
 
--- a/deploy/create_linux_appimage.sh
+++ b/deploy/create_linux_appimage.sh
@@ -8,7 +8,7 @@ if [[ $# -eq 0 ]]; then
   exit 1
 fi
 
-QGC_SRC=$1
+QGC_SRC=$(readlink -f $1)
 
 QGC_CUSTOM_APP_NAME="${QGC_CUSTOM_APP_NAME:-QGroundControl}"
 QGC_CUSTOM_GENERIC_NAME="${QGC_CUSTOM_GENERIC_NAME:-Ground Control Station}"
@@ -29,6 +29,7 @@ if [ ! -f ${QGC_RELEASE_DIR}/${QGC_CUSTOM_BINARY_NAME} ]; then
 fi
 
 OUTPUT_DIR=${3-`pwd`}
+OUTPUT_DIR=$(readlink -f $OUTPUT_DIR)
 echo "Output directory:" ${OUTPUT_DIR}
 
 # Generate AppImage using the binaries currently provided by the project.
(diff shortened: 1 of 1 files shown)


potaito:
Building appimage still not fully succeeding on my machine. Will probably canonicalize everything :D

potaito:
This should do it

potaito:
I have not yet been able to launch the AppImage. Not sure if this is because of the changes I proposed. But when launching, I get 
```
./AppRun: 10: /tmp/tmp.sWa8xaAgVv/CustomQGC.AppDir/QGroundControl: not found
```
where the tpm.xxxx folder is the one uesd on my machine for creating the AppImage. Looking into the issue.

DonLakeFlyer:
Can you provide an example command line which fails prior to this change. It was all working fine for me.

potaito:
> Can you provide an example command line which fails prior to this change. It was all working fine for me.

Local file path arguments like
```
create_linux_appimage.sh . ../build-xzy/release ./appimage
```

potaito:
I just with latest qgroundcontrol master. Tried to replicate the full travis workflow for linux builds, and the appimage creation fails when local paths are used.

```
# Fake travis variables
export TRAVIS_BUILD_DIR=$PWD
export SHADOW_BUILD_DIR=build   # <- in travis this is actually /tmp/something

# run qmake
echo "\n----- BUILD STEP ----\n"
mkdir $SHADOW_BUILD_DIR 
cd $SHADOW_BUILD_DIR
export STABLE_OR_DAILY=StableBuild;
export CONFIG=installer;
export CODESIGN=nocodesign
export SPEC=linux-g++-64 CONFIG=installer

# compile
qmake -r ${TRAVIS_BUILD_DIR}/qgroundcontrol.pro CONFIG+=${CONFIG} CONFIG+=${STABLE_OR_DAILY} CONFIG+=${CODESIGN} -spec ${SPEC};
make -j`nproc --all`

# Create AppImage
echo "\n----- Create AppImage ----\n"
./deploy/create_linux_appimage.sh ${TRAVIS_BUILD_DIR} ${SHADOW_BUILD_DIR}/staging ${SHADOW_BUILD_DIR}/package;
```

fails with

```
Output directory: build/package
sending incremental file list
rsync: change_dir "/tmp/tmp.94hNDFSV4i//build/staging" failed: No such file or directory (2)
```

But when canonicalizing the build folder like it is in travis (`/tmp/something` instead of `./build`), it works.

potaito:
The resulting AppImage works fine with the modified script :)

deploy/create_linux_appimage.sh
```diff
@@ -87,5 +88,6 @@ chmod a+x ./appimagetool-x86_64.AppImage
 
 ./appimagetool-x86_64.AppImage ./$APP.AppDir/ ${TMPDIR}/$APP".AppImage"
 
+mkdir -p ${OUTPUT_DIR}
```
DonLakeFlyer, line 91: Doesn't `mkdir` error out of the directory is already there? That will fail CI.

potaito: Nah. `-p` is the secret weapon for that :)
```
 -p, --parents     no error if existing, make parent directories as needed
```

DonLakeFlyer: Ah, sorry missed that

potaito:
Thanks :grin: :+1:

PR #9330, opened 2021-01-13
Refactor sub default param loading, allow loading params for sub 4.0

Williangalvani:


jaxxzer (approved):

jaxxzer:
@DonLakeFlyer Could you cherry-pick this commit to the 4.1 branch? This would be a nice fix for us to have on stable.

DonLakeFlyer:
Just create a branch against Stable_V4.1 and then cherry pick this to that branch. And then submit a pull request from that and set the branch for it to be Stable_V4.1. That way it all runs through the normal CI/review process.

PR #9339, opened 2021-01-15
Refactor CameraControl

spark404:
Refactor CameraControl
- Separate Definition File Parsing to dedicated class
- Separate HTTP Downloading to dedicated class
- Move Video Stream Info class to own files
- Move Camera Options to own files

I was trying to understand how the Camera Component worked and refactored some of it in the meantime. If this is useful I'm happy to take feedback and spend some more time cleaning this up. If it isn't, no worries. The exercise was worth it :-) The idea was that this refactoring improves separation of concerns and makes testing of internal bits easier. Though no tests have been added yet by me.

The functional testing I did on this bit is with a ArduPilot and a camera component. Scenarios tested are with a definition file and without a definition file.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -108,6 +108,10 @@ WindowsBuild {
     QMAKE_TARGET_PRODUCT        = "$${QGC_APP_NAME}"
 }
 
+MacBuild {
+    QMAKE_MACOSX_DEPLOYMENT_TARGET = 10.12
+}
+
 #-------------------------------------------------------------------------------------
 # iOS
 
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
In general I like this. But as it currently stands there are no unit tests to back up this code and the original contributor isn't around any more. Which leaves it to me to review all of this blindly and make sure it doesn't introduce any bugs. Unfortunately right now I just don't have time for that. So given this code is functionally working fine now I would rather leave it as is. Sorry about that.

spark404:
Hey Don, that's a fair point. Don't worry about it. I'll see if I can do anything about unit testing this, without those its hard to review. And as you say it's currently working fine.

DonLakeFlyer:
Closing for now

PR #9359, opened 2021-01-20
MissionManager: get currentMissionIndex from High_lat messages

Davidsastresas:
High latency messages provide wp_num field, which is meant to be used instead of MISSION_CURRENT when we are under high latency link. 

This PR provides support in mission manager so it takes the currentMissionIndex from high latency as well, so the highlighting of the active Waypoint over the map works under High latency as well.

### DIFF
--- a/src/MissionManager/MissionManager.cc
+++ b/src/MissionManager/MissionManager.cc
@@ -221,6 +221,14 @@ void MissionManager::generateResumeMission(int resumeIndex)
 void MissionManager::_mavlinkMessageReceived(const mavlink_message_t& message)
 {
     switch (message.msgid) {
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@Davidsastresas The duplication of code here is really bad. Can you instead create a worker function like `_updateMissionCurrent` and have the variants call that with the new mission index instead?

Davidsastresas:
You are completely right, I am sorry. I wasn't sure If I should refactor or leave it that way to modify less of the code present, but It is much nicer refactored indeed. I changed it as you suggested, but called the new method _updateMissionIndex, so it is more differentiable than _handleMissionCurrent. Let me know If I need to modify anything else. Thanks!

PR #9361, opened 2021-01-21
main.cc: Different runguard names for custom builds

Davidsastresas:
This way we can run at the same time regular QGC and custom QGC and runguard will not block the execution of each other.

This was talked in slack a few months back, and it was suggested to use the executable name for the runguard. This way the runguards of daily and stable will block each other, but custom builds would be able to run independently.

I wasn't able to get the executable name, but with the approach of this commit we have the same effect, only Custom will be compatible with non custom, but stable and daily will still have same runguard and thus won't be able to run at the same time.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -223,7 +223,14 @@ bool checkAndroidWritePermission() {
 int main(int argc, char *argv[])
 {
 #ifndef __mobile__
-    RunGuard guard("QGroundControlRunGuardKey");
+    // We make the runguard key different for custom and non custom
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
The application name is available in the code from the `QGC_APPLICATION_NAME` define. You should use that and add "RunGuardKey" to the end to build the string.

DonLakeFlyer:
Also need a rebase to get through CI

Davidsastresas:
Done and rebased! Let me know if you need me to modify anything. Thanks!

PR #9362, opened 2021-01-21
PlanToolBarIndicators: Fix mission time representation

Davidsastresas:
For missions of less than 1 hour this indicator was showing 1 hour plus the minutes of the mission. Also for missions longer than 24 hours we had overflow here. 

This commit fixes those issues, and now missions longer than 24 hours are indicated dd hh mm ss.

### DIFF
--- a/src/PlanView/PlanToolBarIndicators.qml
+++ b/src/PlanView/PlanToolBarIndicators.qml
@@ -18,7 +18,7 @@ Item {
 
     property var    missionItems:               _controllerValid ? _planMasterController.missionController.visualItems : undefined
     property real   missionDistance:            _controllerValid ? _planMasterController.missionController.missionDistance : NaN
-    property real   missionTime:                _controllerValid ? _planMasterController.missionController.missionTime : NaN
+    property real   missionTime:                _controllerValid ? _planMasterController.missionController.missionTime : 0
     property real   missionMaxTelemetry:        _controllerValid ? _planMasterController.missionController.missionMaxTelemetry : NaN
     property bool   missionDirty:               _controllerValid ? _planMasterController.missionController.dirty : false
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I believe this is still screwed up with respect to missionTime and _missionTime using different values to indicate not time. One uses NaN and the other 0. It should be a simple fix to take your pull and fix it up. I'll do that, and the you can look. I'm trying to get another stable cleaned up today and marked so I'll put it there first.

DonLakeFlyer:
Also in missionController the missionTime value uses 0 to indicate no time as well.

Davidsastresas:
I tested this and it seemed to work fine with both 0 time missions and missions with time. Let me know if you want me to do some changes here, or what the conflicts are and I will fix it. I know it is not ideal but that is the best approach that came to my mind using qml dates. Thanks!

DonLakeFlyer:
Look at #9371 to see. There is a missing change to set _missionTime to 0 not NaN and the check should be a zero check. Problem was mixed usage of NaN and 0 to represent no time. Should have been 0 everywhere.

And then there is all your other fixes for specifying year to fix less than an hour thing and the more than one day stuff all of which is fine.

DonLakeFlyer:
The bug is a bit elusive since the it OS dependent relating to jscript implementation of the Date object.

Davidsastresas:
I see it now. Understood. Maybe we can close this and move the conversation to  #9371

DonLakeFlyer:
> Maybe we can close this and move the conversation to #9371

That is for stable only. Can you fix your's up to match the additional changes in my pull? Then master will get the mission time = 0 fixes and you more than a day stuff as well.

Davidsastresas:
> > Maybe we can close this and move the conversation to #9371
> 
> That is for stable only. Can you fix your's up to match the additional changes in my pull? Then master will get the mission time = 0 fixes and you more than a day stuff as well.

Ah sorry, I didn't realize that was for stable only. I included your changes from #9371. Let me know if you need me to modify anything else. Thanks!

DonLakeFlyer:
@Davidsastresas Perfect, thanks

PR #9366, opened 2021-01-22
Fixed an issue where MSG_ID_PING are incorrectly handled

jlucke86:
Fixed an issue where MSG_ID_PING of type response get parsed assuming ping type request.

The expected ping sequence is shown https://mavlink.io/en/services/ping.html

Current implementation wrongly respond to MSG_ID_PING which contain a ping response, with another ping response.


LorenzMeier:
Thanks for fixing this! Looking into the CI failure.

LorenzMeier:
Will hold this for any 4.x fixes in the next couple of days.

DonLakeFlyer:
CI failure is google play still broken on master. I'll cherry pick to Stable and the CI there which works will double verify the fix by passing.

PR #9370, opened 2021-01-22
Fix AirMap telemetry support in QGroundControl

jennerl:


src/MissionManager/MissionController.cc
```diff
@@ -2509,14 +2509,18 @@ void MissionController::_updateTimeout()
                 case MAV_CMD_NAV_WAYPOINT:
                 case MAV_CMD_NAV_LAND:
                     if(pSimpleItem->coordinate().isValid()) {
+                        double alt = 0.0;
+                        if (!pSimpleItem->altitude()->rawValue().isNull()) {
+                            alt = pSimpleItem->altitude()->rawValue().toDouble();   // for some reason, the item's coordinate().altitude() is out of sync with the altitude()
+                        }
                         if((MAV_CMD)pSimpleItem->command() == MAV_CMD_NAV_TAKEOFF) {
                             takeoffCoordinate = pSimpleItem->coordinate();
+                            minAlt = maxAlt = alt;
```
jennerl, line 2518: On takeoff, initialize min and max altitude to the takeoff altitude.

src/MissionManager/MissionController.cc
```diff
@@ -2509,14 +2509,18 @@ void MissionController::_updateTimeout()
                 case MAV_CMD_NAV_WAYPOINT:
                 case MAV_CMD_NAV_LAND:
                     if(pSimpleItem->coordinate().isValid()) {
+                        double alt = 0.0;
```
jennerl, line 2512: get the current item's altitude, if available. Somehow, `pSimpleItem->coordinate()->altitude()` is incorrect but the altitude in `pSimpleItem->altitude()` is correct.

PR #9371, opened 2021-01-25
Fix mission time display

DonLakeFlyer:
This is a simpler version of #9362 for Stable


Davidsastresas:
Just tested this and it solves the issue of indicating one hour when less than 1 hour missions, but the more than one day overflow is still present. Thanks!

DonLakeFlyer:
> but the more than one day overflow is still present

Yup, that is what I was shooting for. I want to multi-day stuff to get more bake time in master.

PR #9377, opened 2021-01-27
joystick: fix crash on app exit

bkueng:
This fixes a pure virtual method call exception during exit when a joystick is connected.

The problem was that ~Joystick() waited for the thread to exit, while the
thread called the virtual method _close(). But at this point the destructor
of the derived class (JoystickSDL in my case) was already called.

To resolve the problem a separate stop() method is added that is called
before deleting the object.


DonLakeFlyer:
@bkueng Does this only show up in the debugger? Or can a real user hit this? Wondering whether this needs to go to Stable.

DonLakeFlyer:
I've cancelled the build for this. I'm trying to speed up getting a tagged stable build through. Travis is dog slow. I'll restart once tag build is done and out.

DonLakeFlyer:
@bkueng Can you rebase to get CI fixes

Issue #9263, opened 2020-12-18
Crash referencing stale _activeVehicle in Joystick code


DonLakeFlyer:
Crash is here: https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L3804 caused by referencing a stale _activeVehicle in the Joystick code.

It hit this using PX4 SITL:
* Start sitl with an iris
* Start QGC with joystick enabled
* Stop SITL
* QGC shows comm lost
* Click Disconnect
* Boom

It's some sort of race condition. I can't get it to reproduce now. But from looking at the joystick code _activeVehicle is set in Joystick::startPolling. But when a vehicle is being destructed I don't see how that _activeVehicle gets cleared. I'm not familiar with this code at all but something seems wrong here with respect to there not being a stopPolling in the Vehicle code.

DonLakeFlyer:
@jaxxzer Can someone from your team look at this. Likely faster since I'm not familiar with the joystick code any more.

jaxxzer:
@DonLakeFlyer  we are trying to replicate it here.

DonLakeFlyer:
I can't make it happen again myself. Will likely need to be tracked down through code inspection. When looking through code myself something seemed fishy as I described above. But I don't know that code enough to say exactly what. Smells wrong through.

PR #9378, opened 2021-01-28
fix joystick: avoid accessing invalid pointer after vehicle is destroyed

bkueng:
Joystick::_activeVehicle was still pointing to a vehicle after it got destroyed (e.g. on disconnect) and the thread continued to access it.

Ideally Joystick::_activeVehicle would be changed to a weak pointer.

@DonLakeFlyer your assessment in https://github.com/mavlink/qgroundcontrol/issues/9263 is correct, there was no protection. I did not see a crash, but I saw the pointer being dereferenced after free.

This should fix https://github.com/mavlink/qgroundcontrol/issues/9263

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -459,6 +459,10 @@ Vehicle::~Vehicle()
     delete _mav;
     _mav = nullptr;
 
+    if (_joystickManager) {
+        _startJoystick(false);
+    }
+
 #if defined(QGC_AIRMAP_ENABLED)
     if (_airspaceVehicleManager) {
         delete _airspaceVehicleManager;
@@ -1876,6 +1880,7 @@ void Vehicle::_startJoystick(bool start)
             joystick->startPolling(this);
         } else {
             joystick->stopPolling();
+            joystick->wait(500);
         }
     }
 }


LorenzMeier (approved):

src/Vehicle/Vehicle.cc
```diff
@@ -459,6 +459,10 @@ Vehicle::~Vehicle()
     delete _mav;
     _mav = nullptr;
 
+    if (_joystickManager) {
```
DonLakeFlyer, line 462: If a Vehicle is being destructed after the joystick manager is gone then that indicates some sort of worse internal coding error. All Vehicles should be gone by the time toolbox tools are destroyed. If this fixed the crash then it's just hiding a larger shutdown ordering problem.

In looking at Vehicle constructor I don't understand with the joystick manager is passed in to it. That should just be pulled from the toolbox.

DonLakeFlyer: Wait, am I reading this wrong? Is this needed for the offline vehicle which has a null pointer here? Something here still seems out of whack.

bkueng: > Is this needed for the offline vehicle which has a null pointer here?

Correct

DonLakeFlyer: Got it, thanks.

DonLakeFlyer:
@bkueng Can you cherry pick this to a pull for Stable_V4.1?

PR #9256, opened 2020-12-16
PX4 MC PID tuning page

bkueng:
This brings the PID tuning page into a minimally working state.

qgc_pid_tuning_wifi

And over a telemetry link:
qgc_pid_tuning_telem
Unfortunately the update rate is not sufficient (around 10Hz) for it to be useful.

I removed the attitude, since that is generally not required to be tuned for MC.

### TODO
Before merging:
-  [x] make the advanced the default, remove hover & min throttle sliders
-  [x] make sure things are still corrrect for other vehicle types
-  [x] fix overlapping text for slider
-  [ ] descriptions can probably be improved

For later:
- disable graph over telemetry link (e.g. if update rate is < 20Hz)
- keep data when switching between axis (graph + clipboard)
- add ability to lock roll+pitch sliders together
- start graph on arming if not already running
- mouse zooming might be useful
- potentially show `MC_ROLLRATE_MAX` in the graph

Apart from tuning, I think an additional behavior tab would be useful, with sliders for max velocity and responsiveness (jerk limits, max acceleration, which ideally is a combined slider).

@DonLakeFlyer feel free to push the branch for further changes/improvements.

PR #9379, opened 2021-01-28
PX4 PID tuning page continued

bkueng:
This continues the work from #9256:
- makes the plots + mavlink streams more generic
- adds mouse wheel zooming
- adds all MC controllers (attitude, velocity + position)
- adds MC_AIRMODE + THR_MDL_FAC params
  - with links to docs (opening the external link did not work for me though, due to a Qt vs. system lib mismatch)
- adds MPC_POS_MODE for velocity + position controllers (this could possibly be set automatically, but we need to avoid user confusion and ensure it's always restored)
- adds vtol
- adds fw (not finished yet)

This is the MC rate controller page for a VTOL:
qgc_pid_tuning_vtol_rate

Basic FW TECS plot:
qgc_pid_tuning_plane_tecs

~Known issue: there is no scrollbar shown when the page overflows the vertical window size. @DonLakeFlyer I'd appreciate if you can have a look.~ fixed

Thanks to @dagar @MaEtUgR @RomanBapst for your inputs

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -687,6 +687,7 @@ HEADERS += \
     src/Vehicle/GPSRTKFactGroup.h \
     src/Vehicle/InitialConnectStateMachine.h \
     src/Vehicle/MAVLinkLogManager.h \
+    src/Vehicle/MAVLinkStreamConfig.h \
     src/Vehicle/MultiVehicleManager.h \
     src/Vehicle/StateMachine.h \
     src/Vehicle/SysStatusSensorInfo.h \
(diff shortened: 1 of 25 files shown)


dagar:
> adds MC_AIRMODE + THR_MDL_FAC params
> with links to docs (opening the external link did not work for me though, due to a Qt vs. system lib mismatch)

Are either of these going to be actively adjusted during tuning? I think it would be better to keep the tuning tab focused on the absolute minimum you're adjusting based on the particular plots. Should we separately start capturing things in a static vehicle specific setup tab?

bkueng:
THR_MDL_FAC could be separate, but I didn't see enough of those that justifies another tab. And the rate controller is the right place, as you start to see oscillations at either higher or lower throttle if it's wrong.
I don't want to overload the plots either though.

farhangnaderi:
@bkueng Shall I test this with the latest Master? Cause behaviors are different between Stable and Master QGC.

bkueng:
This is still an open PR, so you'll need to build it yourself.

farhangnaderi:
My bad, mistaken with this,

https://github.com/mavlink/qgroundcontrol/pull/9256

mrpollo:
@bkueng hey can you please let us know when this is ready for merge? we are eager to get this in :)

bkueng:
Yes so am I. It's waiting for @RomanBapst to do the fixed-wing part. If that takes much longer I can leave that out for now.

bkueng:
I disabled fixed-wing for now

PR #9386, opened 2021-02-01
Arch setup script

MaEtUgR:
Finally I get into QGC development let's see where that leads.
FYI @julianoes 

- I added the setup script for the installations I had to do and am planning to extend the documentation with the workflow I'm using.
- The icu 56 specific version staging seems to be required for Ubuntu version 16.04 but on distributions that don't stock this version anymore it leads to an error. So I added a condition that just takes it if it exists and otherwise the existing version. This should ensure that as long as our deployment CI has the library it deploys and for any other person who builds it this should not lead to an error.

### DIFF
--- a/tools/setup/arch.sh
+++ b/tools/setup/arch.sh
@@ -1,0 +1,13 @@
+sudo pacman -Sy --noconfirm --needed \
+    speech-dispatcher \
+    qt5-base \
+    qt5-charts \
+    qt5-graphicaleffects \
+    qt5-location \
+    qt5-multimedia \
+    qt5-quickcontrols \
+    qt5-quickcontrols2 \
+    qt5-serialport \
+    qt5-speech \
+    qt5-tools \
+    ;
--- a/QGCPostLinkCommon.pri
+++ b/QGCPostLinkCommon.pri
@@ -104,12 +104,17 @@ LinuxBuild {
         libicui18n.so* \
         libQt5TextToSpeech.so.5
 
-    !contains(DEFINES, __rasp_pi2__) {
+    exists($$[QT_INSTALL_LIBS]/libicudata.so.56) {
         # Some Qt distributions link with *.so.56
         QT_LIB_LIST += \
             libicudata.so.56 \
             libicui18n.so.56 \
             libicuuc.so.56
+    } else {
+        QT_LIB_LIST += \
+            libicudata.so \
+            libicui18n.so \
+            libicuuc.so
     }
 
     for(QT_LIB, QT_LIB_LIST) {


mrpollo:
Nice to see you here @MaEtUgR 🎉 

@DonLakeFlyer What's missing in this PR to get it merged?

MaEtUgR:
Anything missing? I'm successfully using this setup on two machines now and documented the use in https://github.com/mavlink/qgc-dev-guide/pull/116

julianoes (approved):
Looks sane.

tools/setup/arch.sh
```diff
@@ -0,0 +1,13 @@
+sudo pacman -Sy --noconfirm --needed \
+    speech-dispatcher \
+    qt5-base \
```
DonLakeFlyer, line 3: Pulling Qt libs like this is incorrect since it just pulls the Qt version which is associated with the OS. QGC builds use a specific version of Qt which must be installed as described in the dev guide instructions. Building QGC from random versions of Qt is not unsupported.

Also including this script file duplicates the instructions which are in the deb guide. Thus creating two places which can get out of sync. I don't see the value of including a script file which will ultimately install just a single package.

MaEtUgR: @DonLakeFlyer Thanks for having a look!

> it just pulls the Qt version which is associated with the OS

That is correct. The reason for the pull request that I'm successfully using this workflow without any issue. It's much more convenient and faster to set up and requires orders of magnitude less memory. The goal is to document this experience to people using Arch who I'd assume know about their OS's packet versioning. I'll also continue using this workflow and would follow up on the script if anything happens to break.

Is that acceptable or do I have to remove the script?

MaEtUgR: Maybe for additional context, I do this for PX4 since 2019:
https://github.com/PX4/PX4-Autopilot/blob/9d11d94e7ef51139b1a3a8aaa380a256f8e3162f/Tools/setup/arch.sh

DonLakeFlyer: > The reason for the pull request that I'm successfully using this workflow without any issue.

Just because it compiles doesn't mean there are not subtle bugs introduced by using a version of Qt which nobody else is using. QGC documentation states that it requires a specific version of Qt for a reason. That reason is years of experience pointing to that requirement being necessary though many fixes of subtle bugs introduced by new/different versions of Qt. If you go against that you are creating a version of QGC that has been verified as working correctly by nobody.

MaEtUgR: I'm not trying to step on anyone's foot here. To come back to my question do I hear it's not acceptable and I should revert the toolchain documenting changes?

DonLakeFlyer: Yes. Because the only thing valid in the script is the one install of speech-dispatcher which is already doc'ed (or it was) in the readme. If it was more complex than that I script might make sense. But not given current state of affairs.

DonLakeFlyer: > I'm not trying to step on anyone's foot here.

Understood. I get a bit frustrated because I've have this same conversation of Qt version so many times with different people. Maybe I need to update the README with more details as to no in general you really really shouldn't do this and why.

hamishwillee: @DonLakeFlyer I've added https://github.com/mavlink/qgc-dev-guide/pull/125 to at least add a reason in the build instructions. I think adding in the readme as well might make sense.

PR #9390, opened 2021-02-02
Add millisec to video file name for better sync with flight logs

tzuriel-spearuav:


mrpollo:
Thanks for the clean PR @tzuriel-spearuav I'll bring this up with the dev team

mrpollo:
@tzuriel-spearuav the PR looks clean, what is the reason behind this change request?

DonLakeFlyer:
This doesn't make sense to me since it adds a ton of noise to filenames.

LorenzMeier (changes requested):
Adding milliseconds here seems to trying to solve a different problem and is adding complexity.

PR #9391, opened 2021-02-02
FlightMap: Add widget with gimbal orientation feedback

spark404:
I'm working on adding control for a gimbal mount in qgroundcontrol. This PR adds a simple widget that displays the current feedback provided by ArduPilot (MOUNT_STATUS) message.

Functionality:
 * Visual feedback on gimbal angles from ArduPilot in a widget on the FlyView screen
 * Option to change the gimbal mode from a settings popup from the widget
 * Option to click on a map location to set as GPS tracking target for the gimbal
 
image

image

image

### DIFF
--- a/qgcimages.qrc
+++ b/qgcimages.qrc
@@ -195,5 +195,6 @@
         <file alias="Yield.svg">src/ui/toolbar/Images/Yield.svg</file>
         <file alias="ZoomMinus.svg">src/FlightMap/Images/ZoomMinus.svg</file>
         <file alias="ZoomPlus.svg">src/FlightMap/Images/ZoomPlus.svg</file>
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
@spark404 Sorry for losing track of this for forever! The issue is that MOUNT_STATUS is an Ardupilot dialect specific mavlink message. It is not part of the common standard dialect. Current QGC is only implementing to common mavlink dialect. I believe there is newish (couple years old now I think) camera and gimbal messages which provide gimbal status in the common dialect. Ardupilot would need to support the standard mavlink spec and then gimbal status could be displayed from that which would make it generic support.

DonLakeFlyer:
Also not sure why there is a need for a new "Gimbal Focus" thing when ROI targeting is already supported?

PR #9393, opened 2021-02-03
Vehicle: hide 'No GPS Lock' message if GPS is not present

### LINKED ISSUE #9380: "No Gps Lock for Vehicle" shows up on top of video even for GPS-less vehicles.
## Expected Behavior
Unobstructed video.

## Current Behavior
"No GPS Lock for Vehicle" shows at the center of the video

## Steps to Reproduce:
 - Connect a Pixhawk with ArduSub 4.0.2 flashed
 - Put the video in full screen in the fly view.

## System Information
- Operating System: [e.g. iOS 12.1, Windows 8, macOS Mojave 10.14.1, Ubuntu 14]
- QGC Version:  4.1 and master
- QGC build: self-built from source
- Autopilot (with version): ArduSub 4.0.2

We could check if the vehicle has a GPS connected, or check if a GPS is crucial for this type of vehicle.
It would also be fine if this was displayed only over the map (as the map is not that useful without a GPS signal), but not over the video.

Williangalvani:
Fix #9380

With this, "No GPS Lock for Vehicle" will only show if the vehicle reports a GPS connected.

PX4 seems to always report a GPS. if there is none connected at boot, it's reported with an "error" status.
ArduPilot only reports a GPS after it is connected.

### DIFF
--- a/src/Vehicle/Vehicle.h
+++ b/src/Vehicle/Vehicle.h
@@ -237,6 +237,7 @@ public:
     Q_PROPERTY(bool                 readyToFly                  READ readyToFly                                                     NOTIFY readyToFlyChanged)
     Q_PROPERTY(QObject*             sysStatusSensorInfo         READ sysStatusSensorInfo                                            CONSTANT)
     Q_PROPERTY(bool                 allSensorsHealthy           READ allSensorsHealthy                                              NOTIFY allSensorsHealthyChanged)    //< true: all sensors in SYS_STATUS reported as healthy
+    Q_PROPERTY(bool                 requiresGpsFix              READ requiresGpsFix                                                 NOTIFY requiresGpsFixChanged)
 
     // The following properties relate to Orbit status
     Q_PROPERTY(bool             orbitActive     READ orbitActive        NOTIFY orbitActiveChanged)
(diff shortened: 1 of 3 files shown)


src/Vehicle/Vehicle.h
```diff
@@ -554,6 +555,7 @@ class Vehicle : public FactGroup
     bool            readyToFly                  () { return _readyToFly; }
     bool            allSensorsHealthy           () { return _allSensorsHealthy; }
     QObject*        sysStatusSensorInfo         () { return &_sysStatusSensorInfo; }
+    bool            requiresGpsFix              () { return static_cast<bool>(_onboardControlSensorsPresent & SysStatusSensorGPS); }
```
patrickelectric, line 558: this function should be const.

patrickelectric, line 558: this function should be const

Williangalvani: right! fixed it.

mrpollo:
thanks @Williangalvani

Williangalvani:
Do you have a gps connected?

atyshka:
No I don’t, the point is it’s indoors and I want to use vision_position_estimate messages instead of gps

Williangalvani:
Then I guess the issue is that PX4 is sending the gps flag of the  `the onboard_control_sensors_present` flags of the `SYS_STATUS` message.

PR #9395, opened 2021-02-04
Video: added preset for Yuneec Mantis G

AndKe:


mrpollo:
@AndKe thanks for the PR, can you share more details about this? how did you test this, and where did you find the static IP for the **rtsp** video?

AndKe:
Tested it with an actual Mantis G, the static IP of the vehicle is easy to spot when connected to it's WLAN.

LorenzMeier:
Is the Mantis G still commercially available?

AndKe:
@LorenzMeier Yes, - we bought two recently. and got a bugfix for firmware in northern regions as well (the official was toilet bowling BADLY, then behaved strangely even in the "MANUAL" mode)
Is there a reason to believe they will stop selling soon ?

src/VideoManager/VideoManager.cc
```diff
@@ -692,6 +693,8 @@ VideoManager::_updateSettings(unsigned id)
         settingsChanged |= _updateVideoUri(0, QStringLiteral("udp://0.0.0.0:5600"));
     else if (source == VideoSettings::videoSourceParrotDiscovery)
         settingsChanged |= _updateVideoUri(0, QStringLiteral("udp://0.0.0.0:8888"));
+    else if (source == VideoSettings::videoSourceYuneecMantisG)
+        settingsChanged |= _updateVideoUri(0, QStringLiteral("rtsp://192.168.42.1:554/live"));
```
patrickelectric, line 697: is that always the IP address ?

AndKe: @patrickelectric yes, the IP addresses are static, same every time.

LorenzMeier (approved):
This is fine to go in, but we will clean up the whole implementation over the next couple of months and so all implementations not following the camera protocol will be removed. The camera protocol is here: http://mavlink.io/en/services/camera.html

AndKe:
> This is fine to go in, but we will clean up the whole implementation over the next couple of months and so all implementations not following the camera protocol will be removed. The camera protocol is here: http://mavlink.io/en/services/camera.html

That will destroy more than do any good, there are plenty of use-cases and custom machines that do not integrate the video info/control with MAVLink

PR #9400, opened 2021-02-04
cmake updates and fixes

dagar:
- use any Qt 5.12.x or 5.15.x version found
    - prefers 5.12 over 5.15
    - overridden by setting QT_VERSION
 - add Compression (and XZ_EMBEDDED) to build


dagar:
Tested on Ubuntu 20.04 + Qt 5.15

PR #9401, opened 2021-02-04
Mac os build action

DiegoJRAleixandre:
**Problem**
We were having issues building and deploying for macOS via Travis.

**Solution**
Implement  build and deploy via Github Actions.

An action is created which triggers every time a tag is pushed. It will then proceed to build and deploy the artifacts generated by the build (in this case just the .dmg) as assets of the release of the same name as the tag pushed.

**Additional context**
Probably all deployments could be implemented in this same action, so every release could automatically create the assets for windows, android and macOS

### DIFF
--- a/.github/workflows/buildQGCandDeploy.yml
+++ b/.github/workflows/buildQGCandDeploy.yml
@@ -1,0 +1,76 @@
+name: Build QGC and upload to release 
+
+on: 
+    push:
+      #Every time a tag with name v* (e.g: vX.Y.Z) is pushed,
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

dagar (approved):

DonLakeFlyer:
This partially duplicates the travis build CI but not all of it. Is there some plan to make this take over all of CI and turn off travis? It also seems to refer to some private action from someplace else. Also what does merging this actually do?

PR #9412, opened 2021-02-05
cmake: Qt5QGCConfiguration: Check for FOUND_QT_VERSIONS before continue

patrickelectric:
It'll use the system installation if there is no Qt installation under home

### DIFF
--- a/cmake/Qt5QGCConfiguration.cmake
+++ b/cmake/Qt5QGCConfiguration.cmake
@@ -9,6 +9,9 @@ if(NOT QT_VERSION)
 		$ENV{HOME}/Qt/5.12.*
 		$ENV{HOME}/Qt/5.15.*
 	)
+	if(NOT FOUND_QT_VERSIONS)
+		return()
+	endif()
 	list(SORT FOUND_QT_VERSIONS) # prefer 5.12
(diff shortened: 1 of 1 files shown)


dagar:
> It'll use the system installation if there is no Qt installation under home

You don't think it's ever worth trying that?

patrickelectric:
Are you talking about the system Qt version or the home folder ?
To me there is no reason to have the home folder detection, because it should use the user PKG_* env variables to choose, but I do know that there is a bunch of developers that don't know how to deal with this kind of thing and the home detection may be necessary to decrease the support level.
TLDR: System configuration/installation > home folder

patrickelectric:
ping @dagar

dagar:
I agree the HOME directory thing is hacky, but works now with the defaults of the online installer.

I interpreted your change as it would ONLY work with the home directory Qt install (early return), but if that's actually the opposite then it sounds good to me.

dagar (approved):

patrickelectric:
Hi @dagar, glad that we agree, the early return happens when the HOME detection fails, otherwise it fails completely since it only uses the HOME installation right now. With the return, if the HOME detection fails, it'll continue with the normal PKG_CONFIG method using the system installation

PR #9413, opened 2021-02-05
Set initial values for sliders in PhotoVideoControl from underlying fact

oysteinskotheim:
Before this fix, sliders with camera settings in PhotoVideoControl would all be set to their minimum values (see below) everytime the popup dialog is displayed. After the fix, these sliders will have their initial value set based on the underlying fact

Screenshot_20210205-161943

### DIFF
--- a/src/FlightMap/Widgets/PhotoVideoControl.qml
+++ b/src/FlightMap/Widgets/PhotoVideoControl.qml
@@ -466,6 +466,7 @@ Rectangle {
                                 minimumValue:               parent._fact.min
                                 stepSize:                   parent._fact.increment
                                 visible:                    parent._isSlider
+                                value:                      parent._fact.value
                                 updateValueWhileDragging:   false
                                 onValueChanged:             parent._fact.value = value
                                 Component.onCompleted:      value = parent._fact.value


patrickelectric (approved):

krisklau:
What is the holdup for getting this merged?

src/FlightMap/Widgets/PhotoVideoControl.qml
```diff
@@ -466,6 +466,7 @@ Rectangle {
                                 minimumValue:               parent._fact.min
                                 stepSize:                   parent._fact.increment
                                 visible:                    parent._isSlider
+                                value:                      parent._fact.value
```
DonLakeFlyer, line 469: I don't understand why this is needed given the `Component.onCompleted` call below?

oysteinskotheim: I had to add this to get the sliders to behave as expected. Otherwise they are reset to their minimum value every time the camera settings dialog is displayed.

DonLakeFlyer: There was a Qt bug with sliders with respect to setting the value property like this. You'll see in other usage of this control it follows the pattern of setting the value in onCompleted to work around it (look at SurveItemEditor/qml for an example). I don't have a camera that can show these sliders and SITL doesn't expose them either. So I have nothing to test this with. It should work fine with the onCompleted since I was able to verify that it works correctly in the other usages of it. I don't want to change to the value property setting to expose what I remember as a subtle bug in the control. Can you debug this such to determine why the onCompleted isn't working correctly? Are you using a different version of Qt than what is specific (5.12.6)?

DonLakeFlyer:
Replaced by #9732

PR #9420, opened 2021-02-07
Adding support for map from custom URL

TobinHall:
I wanted to add support for the [basemap](https://basemaps.linz.govt.nz/) service provided by Land Information New Zealand.
They provide both WMTS and XYZ tiling service.

It was simple to add LINZ as a GenericMapProvider, but I was using a free rate-limited 90 day API key embedded in the URL.
I will contact LINZ for an API key for QGroundControl.

In the mean time, I thought it would add a lot of flexibility to allow users to specify their own URLs.
Several GenericMapProviders use the same pattern of substitution on a const QString:
```
static const QString JapanStdMapUrl = QStringLiteral("https://cyberjapandata.gsi.go.jp/xyz/std/%1/%2/%3.png");

QString JapanStdMapProvider::_getURL(const int x, const int y, const int zoom, QNetworkAccessManager* networkManager) {
    Q_UNUSED(networkManager)
    return JapanStdMapUrl.arg(zoom).arg(x).arg(y);
}
```
A simple way to achieve custom URLs for conforming services is just to add an app setting for the URL:

```
QString CustomURLMapProvider::_getURL(const int x, const int y, const int zoom, QNetworkAccessManager* networkManager) {
    Q_UNUSED(networkManager)
    QString url = qgcApp()->toolbox()->settingsManager()->appSettings()->customURL()->rawValue().toString();
    return url.arg(zoom).arg(x).arg(y);
}
```
Users just need to format their URL with the QString arg substitutions %1 %2 %3 for zoom x y respectively. Eg:
`https://basemaps.linz.govt.nz/v1/tiles/aerial/EPSG:3857/%1/%2/%3.png?api=c01exz21103e9cwryvnervpx2qa
`

This syntax means that no additional decoding is necessary, but is slightly less intuitive than substitutions that are more descriptive. 
Provided that this syntax is well documented, I don't think it will be a problem.


src/QtLocationPlugin/QMLControl/OfflineMap.qml
```diff
@@ -368,6 +369,23 @@ Item {
                         visible:        _esriFact ? _esriFact.visible : false
                         font.pointSize: _adjustableFontPointSize
                     }
+
+                    Item { width: 1; height: 1; visible: _customURLFact ? _customURLFact.visible : false }
+                    QGCLabel { text: qsTr("Custom Map URL"); visible: _customURLFact ? _customURLFact.visible : false }
```
patrickelectric, line 374: `visible: _customURLFact && _customURLFact.visible`

src/QtLocationPlugin/QMLControl/OfflineMap.qml
```diff
@@ -368,6 +369,23 @@ Item {
...
+                    FactTextField {
+                        fact:               _customURLFact
+                        visible:            _customURLFact ? _customURLFact.visible : false
```
patrickelectric, line 377: You can have a property for visibility to avoid the same code in multiple places

src/QtLocationPlugin/GenericMapProvider.h
```diff
@@ -60,6 +60,17 @@ class JapanReliefMapProvider : public MapProvider {
     QString _getURL(const int x, const int y, const int zoom, QNetworkAccessManager* networkManager) override;
 };
 
+class CustomURLMapProvider : public MapProvider {
+    Q_OBJECT
+  public:
+    CustomURLMapProvider(QObject* parent = nullptr)
+        : MapProvider(QStringLiteral(""), QStringLiteral(""),
```
patrickelectric, line 67: Use default constructor `QStringLiteral()` to avoid the creation of an empty string

DonLakeFlyer:
@TobinHall YEs, this is great stuff. Sorry for the delay in looking at it.  I had seen the LINZ pull and had in the back of my head this exact change instead. But similar to the problem you ran into with a key, shouldn't this also add support for an api key along with it. Is that not standardized enough to make that be part of this as well? Won't a lot of services require a key?

src/QtLocationPlugin/GenericMapProvider.h
DonLakeFlyer: @patrickelectric I've moved away from trying to use QStringLiteral everywhere and just use naked strings most places now. The reason being is that I think the impact to readability of the code is higher than the benefit of the unmeasurable perf hit of the on the fly generation/allocation.

src/QtLocationPlugin/QMLControl/OfflineMap.qml
TobinHall: Sure. I did wonder about this pattern. This implementation was consistent with the rest of the file.

src/QtLocationPlugin/QMLControl/OfflineMap.qml
TobinHall: Do you have an example of this somewhere in code?
I am interested in using it, I just have not had time to familiarise myself with QML, so I have been borrowing from existing code.

TobinHall:
> @TobinHall YEs, this is great stuff. Sorry for the delay in looking at it. I had seen the LINZ pull and had in the back of my head this exact change instead. But similar to the problem you ran into with a key, shouldn't this also add support for an api key along with it. Is that not standardized enough to make that be part of this as well? Won't a lot of services require a key?

Hi Don. LINZ did provide me with a permanent API key for this project. That is why I also implemented that PR.
API key wise, the LINZ key is part of the URL, so I didn't split the key into extra fields or anything.
I think the key being part of the URL might be enough for a lot of services.
I'm not sure how else the API key would be implemented, web dev is not my area of expertise.

src/QtLocationPlugin/QMLControl/OfflineMap.qml
DonLakeFlyer: The if check for fact being null shouldn't be needed here as far as I can tell. This code is pretty old, not sure why is has all the null checks? Most of this stuff will never be null.

DonLakeFlyer:
Ah, sorry missed that. The key just in the url makes sense. Could you change the format of the url:

`https://basemaps.linz.govt.nz/v1/tiles/aerial/EPSG:3857/%1/%2/%3.png?api=c01exz21103e9cwryvnervpx2qa`
to be this instead:
`https://basemaps.linz.govt.nz/v1/tiles/aerial/EPSG:3857/{zoom}/{x}/{y}.png?api=c01exz21103e9cwryvnervpx2qa`

And then replace on the names instead? That way the format is more easily self documenting. I think it makes sense to include an example in the `longDesc` as well.

TobinHall:
> Ah, sorry missed that. The key just in the url makes sense. Could you change the format of the url:
> 
> `https://basemaps.linz.govt.nz/v1/tiles/aerial/EPSG:3857/%1/%2/%3.png?api=c01exz21103e9cwryvnervpx2qa`
> to be this instead:
> `https://basemaps.linz.govt.nz/v1/tiles/aerial/EPSG:3857/{zoom}/{x}/{y}.png?api=c01exz21103e9cwryvnervpx2qa`
> 
> And then replace on the names instead? That way the format is more easily self documenting. I think it makes sense to include an example in the `longDesc` as well.

Sounds good.
I have added friendlier substitutions of {x} {y} and {z} or {zoom}.

Regarding other quirks in the implementation, this is consistent with the other elements in each respective file.

If there are changes to be made, I think it makes sense to apply them everywhere for consistency. 

Does it make sense to merge this PR as is and create another to fix all of the problem patterns? That way, this implementation is not exceptional.

DonLakeFlyer:
> Does it make sense to merge this PR as is and create another to fix all of the problem patterns? That way, this implementation is not exceptional.

Yup. Makes sense.

@TobinHall One last question. The example has your api key in it. Is that ok?

TobinHall:
> @TobinHall One last question. The example has your api key in it. Is that ok?

That key is the API key that LINZ generated specifically for this project. It is the key that is used in the LINZ pull request. 
I think it is still worth merging that pull request too, I know several people personally who use QGroundControl in New Zealand.

DonLakeFlyer:
@TobinHall Thanks so much for the pull

PR #9432, opened 2021-02-12
missing libicudata.so.56 problem

AuroraRAS:
fix dependencies for Post Link Common. We cannot be sure that the libicu
version in all environments is 56.

Signed-off-by: AuroraRAS <chplee@gmail.com>

### DIFF
--- a/QGCPostLinkCommon.pri
+++ b/QGCPostLinkCommon.pri
@@ -105,11 +105,11 @@ LinuxBuild {
         libQt5TextToSpeech.so.5
 
     !contains(DEFINES, __rasp_pi2__) {
-        # Some Qt distributions link with *.so.56
+        # Some Qt distributions link with *.so
         QT_LIB_LIST += \
-            libicudata.so.56 \
-            libicui18n.so.56 \
-            libicuuc.so.56
+            libicudata.so \
+            libicui18n.so \
+            libicuuc.so
     }
 
     for(QT_LIB, QT_LIB_LIST) {


LorenzMeier (approved):

LorenzMeier:
Thanks!

QGCPostLinkCommon.pri
```diff
@@ -105,11 +105,11 @@ LinuxBuild {
         libQt5TextToSpeech.so.5
 
     !contains(DEFINES, __rasp_pi2__) {
-        # Some Qt distributions link with *.so.56
+        # Some Qt distributions link with *.so
         QT_LIB_LIST += \
-            libicudata.so.56 \
-            libicui18n.so.56 \
-            libicuuc.so.56
```
patrickelectric, line 112: Just a comment, usually @DonLakeFlyer is not a huge fan of a non version specific dependencies, so that may be a problem.
Check the others PRs for here: https://github.com/mavlink/qgroundcontrol/pulls?q=is%3Apr+libicudata

AuroraRAS:
seems some problem there, I will close this pull and try to find other solution.

QGCPostLinkCommon.pri
AuroraRAS: You reminded me. I just comment that lines out, and all build process is working fine. 
Seems my Qt distributions don't need link with that .so files, but I don’t know how to determine which distributions need them.

DonLakeFlyer: > is not a huge fan of a non version specific dependencies, so that may be a problem.

This was a Qt thing where it required a specific version to work with. With latest 5.12 version that requirement may have changed. Don't know.

Issue #9433, opened 2021-02-12
UI Issues in QGC v4.1 and recent. 


sunny52266:
1. Certain UI features not visible until the application is scaled to around 70 percent on android phone. My current phone is oneplus 6. 

2. The translucent nature of the main dock causes readability issues in sunlight specially when the satellite imagery is turned on and bare ground is in the background. For example: ![Screenshot_20210213-001142.jpg](https://user-images.githubusercontent.com/68956206/107812627-c7850080-6d91-11eb-905c-7208cca83169.jpg)

Davidsastresas:
fixed in  #9436

DonLakeFlyer:
I think this screenshot is showing more bugs than just the visibility problem. You can see how the coloring of the text versus the labels differs. I need to take a look at that to see if there is an additional problem.

DonLakeFlyer:
Indoor theme looks like crap as well over certain maps. I think in both cases the values panel needs to be opaque.

Davidsastresas:
That makes sense. What do you think is the cause of the other bug? it seems like the bold feature of the digits is not what it should right?

DonLakeFlyer:
> 1. Certain UI features not visible until the application is scaled to around 70 percent on android phone. My current phone is oneplus 6.

@sunny52266 What else is missing?

sunny52266:
1. Some times the back button does not exit the application. 

2. When use vertical instrument panel is checked, the instrument panel disappears. 

3. There is no option for trigger camera now as there was in previous versions. 

4. Second GPS info does not appear in the gps parameters selection in the parameters dock type thingy on the main display that needed to be opaque. 

These are the ones from the top of my head.

DonLakeFlyer:
> 2\. When use vertical instrument panel is checked, the instrument panel disappears.

Fixed. Will show up when a new stable is marked.

> 3\. There is no option for trigger camera now as there was in previous versions.

This should be fixed as well. Will show up when new stable is marked.

> 4\. Second GPS info does not appear in the gps parameters selection in the parameters dock type thingy on the main display that needed to be opaque.

This was not supported in previous versions. Support was just added and is supported in Daily builds.

sunny52266:
@DonLakeFlyer Must be good then. Shall I close this issue now?

PR #9436, opened 2021-02-13
TelemetryValuesBar: custom transparency by settings

Davidsastresas:
Under some basemaps the default alpha value for this panel was too transparent for the user to see clearly the values. This commit makes this alpha channel variable by a setting under appsettings.

This addresses the issue #9433

### DIFF
--- a/src/FlightDisplay/TelemetryValuesBar.qml
+++ b/src/FlightDisplay/TelemetryValuesBar.qml
@@ -7,22 +7,23 @@
  *
  ****************************************************************************/
 
-import QtQuick                      2.12
-import QtQuick.Layouts              1.12
+import QtQuick                        2.12
+import QtQuick.Layouts                1.12
 
-import QGroundControl               1.0
-import QGroundControl.ScreenTools   1.0
-import QGroundControl.Vehicle       1.0
-import QGroundControl.Controls      1.0
-import QGroundControl.Palette       1.0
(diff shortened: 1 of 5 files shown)


LorenzMeier (approved):

Davidsastresas:
Thank you! Maybe we can close the issue related: https://github.com/mavlink/qgroundcontrol/issues/9433 

I can't close it myself. Thanks!

DonLakeFlyer:
I don't think this is the correct way to handle this. Just like the toolstrip and HUD indicators are not opaque for readability I think given the problems with opacity this control should just be made opaque as well. I also don't think there should be a setting for turn it on/off since it doesn't add much value and just adds more stuff to the already long list of settings.

There is likely a more elegant way to handle this from a visual perspective but brute force opaque for now I think is the best choice.

@Davidsastresas Opinions?

DonLakeFlyer:
Also remember that there is the option to set opacity on individual elements based on value ranges. I don't think that will work well if the background is semi-transparent as well.

Davidsastresas:
I agree 100% with your opinion about this. I just wasn't sure if changing it to opaque would be approved, but yes I think in the short therm it makes sense to make it opaque and we can figure out what to do with it more elegantly in the future.

DonLakeFlyer:
I can revert this PR and then once that goes through can you make a new PR which changes to opaque.

Issue #9435, opened 2021-02-13
Option to display values for GPS2 instrument panel (the selectable one)


pompecukor:
On the user configurable display values, would be nice to be able to also show GPS2 (count and HDOP).
Right now only possible to show for GPS1. one, but that is already show on the horizontal bar at the top,

For us this is mainly useful for when we are testing the aircraft. On QGC there is just no was to make GPS2 information to show.
BTW same goes for secondary Airspeed sensor. 

Thank you.

Davidsastresas:
Partially fixed in #9437. For the airspeed matter I think we need development from mavlink and autopilot firmware as well, as I couldn't see where is currently sent the second airspeed sensor info.

Davidsastresas:
Okey, apparently Airspeed 2 is sent right now from Ardupilot by a "named float" mavlink message. This is a workaround until we have support for it from mavlink, so we should really open an issue in mavlink, and when we have proper support in mavlink for this add it to QGC.

eduardkieser:
Hi @Davidsastresas, I have a related question. I have implemented a sensor driver in lua for an unsupported temperature sensor. I would like to view the values of that sensor in the "Instrument Panel". The message is being sent as a `named_float`, and on mission planner I can configure the quick view to display the value. Is there any way that I can view named floats in the QGC gui?

Davidsastresas:
I don't think we have support for that right now, but it can be added. Can you open a separate issue for that matter? thanks!

eduardkieser:
sure thing. [10564](https://github.com/mavlink/qgroundcontrol/issues/10564).

PR #9437, opened 2021-02-13
Support for mavlink GPS2_RAW message

Davidsastresas:
This PR adds support for GPS2_RAW messages, implemented within the factgroup system so the new field GPS2 can be selected in the main custom telemetry widget.

I've inherited vehicleGPSfactgroup totally and reimplemented the handle message function. This has been tested in v4.1.1, master isn't compiling to me right now. For the PR I rebased to master.

It has been tested with Ardupilot SITL and it works fine.

This PR addresses part of #9435, the airspeed part probably requires development of autopilot firmware as well, I can't see where mavlink is sending the second airspeed values.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -90,6 +90,7 @@ const char* Vehicle::_hobbsFactName =               "hobbs";
 const char* Vehicle::_throttlePctFactName =         "throttlePct";
 
 const char* Vehicle::_gpsFactGroupName =                "gps";
+const char* Vehicle::_gps2FactGroupName =               "gps2";
 const char* Vehicle::_windFactGroupName =               "wind";
 const char* Vehicle::_vibrationFactGroupName =          "vibration";
 const char* Vehicle::_temperatureFactGroupName =        "temperature";
@@ -143,6 +144,7 @@ Vehicle::Vehicle(LinkInterface*             link,
     , _hobbsFact                    (0, _hobbsFactName,             FactMetaData::valueTypeString)
     , _throttlePctFact              (0, _throttlePctFactName,       FactMetaData::valueTypeUint16)
     , _gpsFactGroup                 (this)
+    , _gps2FactGroup                (this)
     , _windFactGroup                (this)
     , _vibrationFactGroup           (this)
     , _temperatureFactGroup         (this)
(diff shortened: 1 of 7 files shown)


LorenzMeier (approved):

src/Vehicle/Vehicle.cc
```diff
@@ -1025,6 +1032,27 @@ void Vehicle::_handleGpsRawInt(mavlink_message_t& message)
     }
 }
 
+void Vehicle::_handleGps2Raw(mavlink_message_t& message)
```
DonLakeFlyer, line 1035: @LorenzMeier This doesn't seem correct to me. Isn't the GPS position fused into a best position from multiple GPS in the main GPS mavlink telemetry and then the coordinate should only be update only from that telemetry? This would seem to cause the vehicle coordinate to ping back and forth between the telemetry from multiple GPS.

Davidsastresas: Your concern makes sense. I just assumed both GPS would be reporting the same coordinate, but you are right. Although I am not sure what would be the best way to handle it.

I could be wrong but from my understanding, the "best position fused" you are talking about must be the global_position_int message, which we are taking into consideration as well for updating the _coordinate from vehicle. So maybe we should leave gps_raw_int and gps2_raw only for the GPS fact groups, and leave the _coordinate update to global_position_int? 

I think this would make sense for Ardupilot firmware, but I am not sure how PX4, or how the "standard" mavlink rules apply to this.

DonLakeFlyer: The main fused lat/lon values come from global_position_int. QGC remembers if it has gotten a global_position_int message. Only in the case where QGC is not getting a global_position_int message will gps_raw_int values be used to update lat/lon for the vehicle as a fallback. I think that is all that is really needed and in reality it's unclear whether the gps_raw_int fallback should be in there at all or whether QGC should only ever rely on global_position_int. But the code has been that way for ages now, so who knows. All of that said it still points to not updating lat/lon from gps2_raw and only updating fact group value from that. So can you remove that code?

@LorenzMeier Any comments on the whole gps_raw_int fallback for lat/lon? Seems possibly not needed.

Davidsastresas: Done! https://github.com/mavlink/qgroundcontrol/pull/9453

PR #9447, opened 2021-02-17
MissiongManager: ComplexMissionItem: Save preset as CborValue variant

### LINKED ISSUE #9423: Stable_V4.1: "Apply Preset" fails with error, missing keys
When trying to "Apply Preset", I get an error about missing keys in the JSON object. I looked at the .json file for the preset, and it indeed has these keys it says are missing. This happens even after creating a fresh new preset from a Basic Survey as seen in the image below. I just created a preset called "test", saved it as a new preset, and then tried to load it. 

image

patrickelectric:
Fix #9423

### DIFF
--- a/src/MissionManager/ComplexMissionItem.cc
+++ b/src/MissionManager/ComplexMissionItem.cc
@@ -79,7 +79,11 @@ void ComplexMissionItem::_savePresetJson(const QString& name, QJsonObject& prese
     QSettings settings;
     settings.beginGroup(presetsSettingsGroup());
     settings.beginGroup(_presetSettingsKey);
-    settings.setValue(name, QCborMap::fromJsonObject(presetObject).toCborValue().toByteArray());
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
+    settings.setValue(name, QJsonDocument(presetObject).toBinaryData());
+#else
+    settings.setValue(name, QCborMap::fromJsonObject(presetObject).toCborValue().toVariant());
+#endif
 
     // Use this to save a survey preset as a JSON file to be included in the build
     // as a built-in survey preset that cannot be deleted.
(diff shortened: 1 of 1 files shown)


dakejahl:
Actually, let me elaborate.

This now _does_ work, as long as the preset being loaded is created using the same QCbor fucntion.
```
settings.setValue(name, QCborMap::fromJsonObject(presetObject).toCborValue().toVariant());
```

This means that presets created in QGC 4.0 will not work with 4.1, which is a problem for users. Is there a good reason why we moved from QJsonDocument to QCborValue? I see the docs for 5.15 (we are still on 5.12) mark it as deprecated, but it does say it is kept around specifically to maintain backwards compatibility
https://doc.qt.io/qt-5/qjsondocument-obsolete.html

patrickelectric:
> Just tested, still the same error

You need to save first and after that apply a pattern, otherwise you are going to load a broken pattern, since both load and save functions were broken. If the problem still exist, please provide the steps to replicate it.

dakejahl:
See above comment

patrickelectric:
Hi @dakejahl, the change was done to get QGC ready for Qt 6.0 and with less pain to make the portability later, allowing the maintainability and sustainability of the software development without an aggregate pain in the future.

A check for older Qt builds version was added.

dakejahl:
Great, this works! Thanks!

DonLakeFlyer:
Is this only a problem in master no stable?

dakejahl:
It's a problem in Stable_V4.1

DonLakeFlyer (approved):

DonLakeFlyer:
@patrickelectric Can you cherry pick this into a pull for Stable_V4.1?

patrickelectric:
> @patrickelectric Can you cherry pick this into a pull for Stable_V4.1?

Yes sir!

Issue #9448, opened 2021-02-17
Video (h264): some machines show frames in the wrong order in low-latency mode


Williangalvani:
## Expected Behavior
Smooth video

## Current Behavior

Frames are rendered out of orders, causing things to move back and forward.

## Steps to Reproduce:

Please provide an unambiguous set of steps to reproduce the current behavior

1. Setup a H264 video
2. Enable low-latency mode

## System Information
  - Operating System: Arch Linux (we had reports mostly on windows though)
  - QGC Version: master, > 4.0

https://user-images.githubusercontent.com/4013804/108220905-fdb9db80-7115-11eb-9d79-97fe776f6046.mp4

PR #9449, opened 2021-02-17
VideoReceiver: Remove low-latency mode and set rtpjitterbuffer to a shorter buffer

Williangalvani:
Potential fix for #9448 

This replaces the default 200ms in rtpjitterbuffer for 70ms.

We could allow the user to fine tune this value. but I don't think we should allow that much customization

Needs testing!

### DIFF
--- a/src/Settings/Video.SettingsGroup.json
+++ b/src/Settings/Video.SettingsGroup.json
@@ -123,13 +123,6 @@
     "default":     false
 },
 {
-    "name":             "lowLatencyMode",
-    "shortDesc": "Tweaks video for lower latency",
-    "longDesc":  "If this option is enabled, the rtpjitterbuffer is removed and the video sink is set to assynchronous mode, reducing the latency by about 200 ms.",
-    "type":             "bool",
-    "default":     false
-},
-{
     "name":             "forceVideoDecoder",
     "shortDesc":        "Force specific category of video decode",
     "longDesc":         "Force the change of prioritization between video decode methods, allowing the user to force some video hardware decode plugins if necessary.",
--- a/src/Settings/VideoSettings.cc
+++ b/src/Settings/VideoSettings.cc
@@ -106,7 +106,6 @@ DECLARE_SETTINGSFACT(VideoSettings, enableStorageLimit)
 DECLARE_SETTINGSFACT(VideoSettings, rtspTimeout)
 DECLARE_SETTINGSFACT(VideoSettings, streamEnabled)
 DECLARE_SETTINGSFACT(VideoSettings, disableWhenDisarmed)
-DECLARE_SETTINGSFACT(VideoSettings, lowLatencyMode)
 
 DECLARE_SETTINGSFACT_NO_FUNC(VideoSettings, videoSource)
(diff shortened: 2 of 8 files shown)


LorenzMeier:
I'm not clear on the testing level - did you test this and it fixed your issue and you are looking for more community cross testing?

mrpollo:
For Testing: Needs an h264 source through UDP

LorenzMeier (approved):
On a code level this looks good, but we need to test against test video streams and against real HW targets.

DanielePettenuzzo:
I tested this pr on my video setup and I don't see any regressions although I cannot reproduce the issue on v4.1.0. 
My test setup is the following:
- QGC v4.1.0 running on Arch linux
- QGC low latency mode enabled in the video settings
- UVC camera connected to my laptop via usb and encoded to H264 over RTP with the following gst-launch pipeline:
```
gst-launch-1.0 -v v4l2src device=/dev/video2 ! videoconvert ! x264enc tune=zerolatency ! rtph264pay ! udpsink host=127.0.0.1 port=5600
```
These are some screen recordings of QGC while I was running some tests on v4.1.0 (before this pr):
- https://youtu.be/LZc9M12MU-I
- https://youtu.be/kjndTM0Ht44
- https://youtu.be/h8rFBlsuZw4

I haven't tested on Windows yet because I need to prepare the same setup there. 

FYI @andrewvoznytsa

andrewvoznytsa:
IMO it makes more sense to have some UI control to tune latency (in range off .. some big value like 1000ms) where 'off' would mean no rtpjitterbuffer in the pipeline as it was before for those who wants 3.5.6 pipeline. I'm not sure that it will work properly on all systems - there are high chances that on some Androids and iOSes 70ms will be not enough - it is better to give users control over that that to find blue pill for everyone.

Williangalvani:
> I cannot reproduce the issue on v4.1.0.
> https://youtu.be/h8rFBlsuZw4

I could see the issue at 4 seconds at the laptop screen. Is this video not from v4.1.0?

> IMO it makes more sense to have some UI control to tune latency (in range off .. some big value like 1000ms) where 'off' would mean no rtpjitterbuffer in the pipeline as it was before for those who wants 3.5.6 pipeline.

Yes, I think you are right, and I'm currently leaning towards this option. It would turn the "low-latency mode" checkbox into a slider/field for the buffer size.

Anyway, I'm not particularly in a rush with this PR. I'll work on a test setup to make sure this doesn't introduce any regressions latency-wise on other setups.

DonLakeFlyer:
This has been sitting at Draft for months now. Can I close?

Williangalvani:
> This has been sitting at Draft for months now. Can I close?

I intend to continue this in the coming weeks. My vehicle was taken apart as we are upgrading a lot of things so I can't test properly right now. As soon as it is back in one piece I should pick this up again.
That said, I could open it again later if the staleness bothers you.

holden-zenith:
should this be closed?

Williangalvani:
sure, I'll pick this back up next time it bothers me.

PR #9452, opened 2021-02-18
Make telemetryvaluesbar opaque

Davidsastresas:
Following the discussion in https://github.com/mavlink/qgroundcontrol/issues/9433 this PR makes the panel opaque. It already includes the revert PR : https://github.com/mavlink/qgroundcontrol/pull/9450

Let me know if something additional is needed. Thanks!

PR #9472, opened 2021-02-22
TelemetryValuesBar: change panel from transparent to opaque

Davidsastresas:
Following the request from #9452, this is a cherry-pick of that fix over Stable v4.1.


DonLakeFlyer:
Thanks @Davidsastresas

Issue #9425, opened 2021-02-09
Landing approach shows red in Fly view, but orange in Plan view


Antiheavy:
Is the red intentional to indicate caution?  If so, it seem like we should use something different that the same red color used to indicate terrain intersection in the Plan screen.  Maybe orange triangle with red boarder or something?   I also suspect this might just be a bug...

Fly view:
![image](https://user-images.githubusercontent.com/12117027/107393049-f2383600-6abf-11eb-9b1a-b96e47cd950b.png)

Plan view:
![image](https://user-images.githubusercontent.com/12117027/107393091-fc5a3480-6abf-11eb-9cbf-957a223cbad2.png)

Mission file:
[red-fly-view-landing.zip](https://github.com/mavlink/qgroundcontrol/files/5952642/red-fly-view-landing.zip)

DonLakeFlyer:
Most likely just a bug

Antiheavy:
This bug is still there in the latest stable.

DonLakeFlyer:
I'll associate that with #9495. Once that gets taken care of we'll see if it still happens.

Antiheavy:
this is still happening with the latest test build on stable with attempted fixes for #9495 
![image](https://user-images.githubusercontent.com/12117027/111873362-faf11700-895d-11eb-9afc-ab6279c5be7d.png)

GIF:
![Landing approach still red](https://user-images.githubusercontent.com/12117027/111873396-25db6b00-895e-11eb-82fa-3c4f6ddf46f9.gif)

Mission file:
[landing-approach-red.zip](https://github.com/mavlink/qgroundcontrol/files/6175773/landing-approach-red.zip)

DonLakeFlyer:
Should be fixed now

Antiheavy:
seems fixed based on testing latest daily!  thanks.

DonLakeFlyer:
@Antiheavy Thanks so much for helping with testing as always.

blackbareng:
We are having similar issues as mentioned above when planning a mission for a small fixed wing. 

We’ve planned this exact mission several times before, but in the last few days, the landing pattern is yellow when planning, but turns red when uploaded. The glide slope is 6 degrees. No different than it’s ever been. 

Is this the same bug discussed above?

DonLakeFlyer:
Save the plan that isn't working and attach it here

blackbareng:
Hello Don,

The plan is attached below, as well as two images from our controller.

[...]

PR #9476, opened 2021-02-22
Ignore terrain collision on beginning/end of takeoff/land segments

### LINKED ISSUE #9466: Terrain collision detection too sensitive near the Landing Area and Launch (fixed wing patterns)
The terrain intersection calculation is too sensitive around the Land point (and Launch point too actually) when using typical fixed wing landing and takeoff patterns.  The landing pattern and takeoff line often show as red when they shouldn't.  

Terrain models such as the one QGC is using have all sorts of weird effects at polygon (or post) edges and they use who-knows-what styles of smoothing functions - they just aren't designed for ultra close-up inspection.  The accuracy and fidelity of the terrain model is fine for mission planning at altitude, but isn't appropriate to be trying to determine a terrain intersection such as a small bump that may or may-not actually be on the ground right near where the pilot is standing anyway.

**Suggested improvement:**  ignore terrain intersections within X meters horizontal distance of the Landing and Launch locations.  Might need to play around to determine X, but in past experience I've seen 3-10 meters used.

terrain intersect calc too sensitive

DonLakeFlyer:
* The start/end of takeoff/land segments are ignored with respect to terrain collision. The distance ignored is 30 meters from beginning/end.
* Fix #9466, #9425

### DIFF
--- a/src/MissionManager/ComplexMissionItem.cc
+++ b/src/MissionManager/ComplexMissionItem.cc
@@ -124,9 +124,9 @@ void ComplexMissionItem::addKMLVisuals(KMLPlanDomDocument& /* domDocument */)
     // Default implementation has no visuals
 }
 
-void ComplexMissionItem::_appendFlightPathSegment(const QGeoCoordinate& coord1, double coord1AMSLAlt, const QGeoCoordinate& coord2, double coord2AMSLAlt)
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
@anitheavy Can you give this a try? Interim stable builds can be installed from here: https://qgroundcontrol.s3-us-west-2.amazonaws.com/builds/Stable_V4.1/QGroundControl-installer.exe

PR #9481, opened 2021-02-24
Support for CustomSettings 

dakejahl:
This PR adds the ability for custom plugin users to define their own SettingsGroup, which they can access via the SettingsManager. I've added it to the custom-example for reference on usage. 

I'll open a PR into master after this has passed review.

### DIFF
--- a/custom-example/custom.pri
+++ b/custom-example/custom.pri
@@ -35,6 +35,9 @@ CONFIG  += QGC_DISABLE_PX4_PLUGIN_FACTORY
 DEFINES += CUSTOMHEADER=\"\\\"CustomPlugin.h\\\"\"
 DEFINES += CUSTOMCLASS=CustomPlugin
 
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
Does this really need to bolt onto SettingsManager? Why not just add to the custom plugin? Then this allows for adding as many groups as you want. I've had in the back of my head removing SettingsManager from the qml class hierarchy since it does nothing and just adds noise to coding.

dakejahl:
Oh yeah I guess the usage would be exactly the same except instead of
```
qgcApp()->toolbox()->settingsManager->() ...
```
you'd do
```
qgcApp()->toolbox()->corePlugin->()->customSettingsManager() ...
```
Is that what you are suggesting?

DonLakeFlyer:
`qgcApp()->toolbox()->corePlugin->()->customSettings()`

There is no need to expose a "Settings Manager" concept since it does nothing useful. Just bolt the SettingsGroups right to the core plugin and skip the middle man. This require no custom build script changes to make happen. It's just regular custom build stuff.

PR #9492, opened 2021-02-26
Updated AirMap support

jennerl:
Resolves issues with AirMap's support in QGC to allow users to file flight plans, send telemetry and utilize AirMap's other services.

Also resolves an issue with altitude handling in creating the flight's bounding cube.

Supporting header files and libs removed from this repo in favor of being downloaded and installed at build time.


remster:
CI fails with:
```
linking staging/QGroundControl
/tmp/shadow_build_dir/libs/airmap-platform-sdk/linux/Qt.5.12/libairmap-qt.so: undefined reference to `std::__cxx11::basic_ostringstream<char, std::char_traits<char>, std::allocator<char> >::basic_ostringstream()@GLIBCXX_3.4.26'
collect2: error: ld returned 1 exit status
```
Quick inspection suggest it may be related to https://github.com/airmap/platform-sdk/pull/155/files (but shouldn't as `StreamLogger` is not instantiated and is header-only) or https://www.neuron.yale.edu/phpBB/viewtopic.php?t=3880 - is there a way of emulating what travis does locally?

LorenzMeier:
What is the state here?

PR #9494, opened 2021-03-01
Add a hill region to testable TerrainQueries.

remster:
`TerrainQueries` typically don't work in the test cycle, but a while back i have contributed [special areas around Point Nemo](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.h#L301) where testing terrain-following `TransectStyleComplexItems` works.

Yet, we were missing a hill to verify transects are breaking on its top as expected and for LoS analysis. This only adds another useful region next to ones that are already there.


src/Terrain/TerrainQuery.cc
```diff
@@ -901,6 +910,19 @@ QList<double> UnitTestTerrainQuery::_requestCoordinateHeights(const QList<QGeoCo
             long dx = regionSizeDeg/one_second_deg;
             double fraction = 1.0 * x / dx;
             result.append(std::round(UnitTestTerrainQuery::LinearSlopeRegion::minAMSLElevation + (fraction * UnitTestTerrainQuery::LinearSlopeRegion::totalElevationChange)));
+        } else if (hillRegion.contains(coordinate)) {
+            double arc_second_meters = (6371000. / 3600.) * (M_PI / 180);
```
patrickelectric, line 914: What is `6371000` ? Please create an intermediary variable to describe this magic number.

Issue #9501, opened 2021-03-02
Huge performance loss when a lot of "picture taken" icons are displayed


Davidsastresas:
When running a mission, if mavlink messages announcing pictures taken from a survey arrive we draw a camera icon over the map. For a few icons QGC works great, but on large surveys ( lots of icons displayed ) QGC starts to behave badly, super laggy. It is 100% connected to the number of camera icons in the map. It seems that has a huge impact in screen rendering or QML engine has a hard time to display them all. 

I took a quick look at it and we are not doing anything different from any other map icons, so maybe it is just the mapquickitem being not super optimized for large amounts of icons? I suspect this comes from qt/qml performance of this qml type directly. Does anybody know anything about this? I would love to help, I thought on limiting the number of picture taken icons, similar to how we limit the lenght of the vehicle trajectory, but maybe there is actually a solution preserving the icons?

Thanks!

Antiheavy:
We've noticed this in the past and had to stop using the photo icons altogether.   It's a real bummer because most users expect some sort of photo location feedback when flying mapping missions.  Of course, there can be 1000's of photos taken on some mapping missions so whatever the solution ends up being, it needs to be performant/efficient.

Davidsastresas:
Thank you, lets see if any of the devs more familiarized with the inners of qml, etc sees the issue and points me to the right direction for working over it. Thanks!

DonLakeFlyer:
I know about it but haven't looked at it to figure it out. 

@Davidsastresas I wonder if it is related to using too many display of the camera icon which I think is an svg file. Could you try changing the camera indicator to a regular qml circular something using a Rectangle and see if the perf is still bad?

Davidsastresas:
@DonLakeFlyer  Yes, it is a .svg file. Sure, I will test it this week. Besides testing simple qml shapes, do you think it is worth testing other icon format, or any other way of preserving the icon? Thanks!

DonLakeFlyer:
> do you think it is worth testing other icon format, or any other way of preserving the icon?

Not sure yet. Theoretically the best possible perf should be from using OpenGL drawing and then you instance the image with a shader such that it is not dup'ed for each icon. But that a bit tricky if you don't know OpenGL.

Davidsastresas:
I tested with regular Qml types and there is no performance issue. I made a camera like icon based on Qml rectangles, the PR is here: #9513

[...]

PR #9513, opened 2021-03-04
CameraTriggerIndicator: use a qml based icon instead:

Davidsastresas:
There were performance issues when too many icons of this kind were displayed on the map, probably because of the old image based nature of the icon. This commit changes it to a qml rectangles based icon. 

As discussed in issue #9501, the optimal way of addressing this would be by openGL shaders. This commit paints a camera icon based on qml rectangles instead, I am not familiarized with the openGL shaders approach right now, and I thought this could be a nice provisional fix.

The performance of QGC is not affected apparently anymore with this new icon. Some screenshots of the change:

Old icon:
old

New icon:
new

Let me know your thoughts, I will be happy to change the PR if needed. Thanks!

### DIFF
--- a/src/FlightMap/MapItems/CameraTriggerIndicator.qml
+++ b/src/FlightMap/MapItems/CameraTriggerIndicator.qml
@@ -29,13 +29,39 @@ MapQuickItem {
 
         readonly property real _radius: ScreenTools.defaultFontPixelHeight * 0.6
 
-        QGCColoredImage {
-            anchors.margins:    3
-            anchors.fill:       parent
-            color:              "white"
-            mipmap:             true
-            fillMode:           Image.PreserveAspectFit
-            source:             "/qmlimages/camera.svg"
+        // Bigger rectangle of camera icon
+        Rectangle {
+            id:                       cameraIconFrameRectangle
+            anchors.horizontalCenter: parent.horizontalCenter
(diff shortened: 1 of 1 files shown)


src/FlightMap/MapItems/CameraTriggerIndicator.qml
```diff
@@ -29,13 +29,52 @@ MapQuickItem {
...
+        // Old icon, disabled due to performance issues when too much icons in the map 
+        
+        // QGCColoredImage {
+        //     anchors.margins:    3
+        //     anchors.fill:       parent
+        //     color:              "white"
+        //     mipmap:             true
+        //     fillMode:           Image.PreserveAspectFit
+        //     source:             "/qmlimages/camera.svg"
+        // }
```
patrickelectric, line 41: Why is this code here ? It should be just removed.

Davidsastresas: As this was a provisional patch until we sort out  how to do it with openGL shaders I thought it would be nice to leave that commented out. It is removed now, thanks for commenting.

Davidsastresas:
I saw the new "move to stable" label. Let me know if you need me to rebase this or whatever. Thanks!

DonLakeFlyer:
@Davidsastresas You can cherry pick to a pull to the Stable branch?

PR #9518, opened 2021-03-07
Remove airmap libs

jennerl:
AirMap libs and headers removed from qgroundcontrol repo in favor of downloading the same at build time if AirMap support is enabled.


LorenzMeier (approved):

LorenzMeier:
Could you look into the travis build failure?

PR #9521, opened 2021-03-11
Fix return _userBrandImageOutdoor

WTPENGUIN:
I found that it returns _userBrandingOutdoor instead of an image string.

### DIFF
--- a/src/ui/toolbar/MainToolBar.qml
+++ b/src/ui/toolbar/MainToolBar.qml
@@ -146,7 +146,7 @@ Rectangle {
 
         function brandImageOutdoor() {
             if (_userBrandingOutdoor) {
-                return _userBrandingOutdoor
+                return _userBrandImageOutdoor
             } else {
                 if (_userBrandingIndoor) {
                     return _userBrandingIndoor


patrickelectric (approved):

src/ui/toolbar/MainToolBar.qml
```diff
@@ -146,7 +146,7 @@ Rectangle {
 
         function brandImageOutdoor() {
             if (_userBrandingOutdoor) {
-                return _userBrandingOutdoor
+                return _userBrandImageOutdoor
             } else {
                 if (_userBrandingIndoor) {
                     return _userBrandingIndoor
```
DonLakeFlyer, line 152: This is wrong as well isn't it?

The problem here is the readability of `_userBrandingOutdoor` not being very useful as to explaining what it really means. I think the code would be more readable if these two properties where removed and the length checked right in the ifs. There is some dupplication here, but I think the better readability overrules. Could you change the code to do that instead? Both the 'brandImageIndoor' and 'brandImageOutdoor' functions are screwed up and need fixing.

WTPENGUIN: Okay. I'll fix the code soon. I close the request.

PR #9522, opened 2021-03-11
Use default QStringLiteral constructor over building a new empty QString

patrickelectric:


patrickelectric:
It appears that msvc hates it, going to close it for now.

PR #9525, opened 2021-03-11
CameraTriggerIndicator: use a qml based icon instead:

Davidsastresas:
There was performance issues when too many icons of this kind were displayed on the map, probably because of the old image based nature of the icon. This commit changes it to a qml rectangles based icon.

This is a cherry-pick from #9513 to Stable branch, after @DonLakeFlyer  request. Thanks!


DonLakeFlyer:
Thanks @Davidsastresas

Issue #9539, opened 2021-03-18
missing RadioComponentController screen in custom build


mnumanuyar:
## Expected Behavior
radio component parameter screen should shown in a custom build that does not support px4 

## Current Behavior
"qrc:/qml/RadioComponent.qml:47:13: RadioComponentController is not a type" is shown in console

## Steps to Reproduce: 
1. start a custom buld that has "CONFIG  += QGC_DISABLE_PX4_PLUGIN QGC_DISABLE_PX4_PLUGIN_FACTORY" in custom.pri
2. connect stil 
3.goto -> vehicle setup -> radio
4. observe the empty screen and "qrc:/qml/RadioComponent.qml:47:13: RadioComponentController is not a type" in console

## System Information
- Operating System: kubuntu 18.04
- QGC Version: 4.1
- QGC build: daily.  
- Flight Controller: ardupilot sitl

## Detailed Description
The problem is RadioComponentController qml type is registered in px4 firmware plugin which is disabled "CONFIG  += QGC_DISABLE_PX4_PLUGIN QGC_DISABLE_PX4_PLUGIN_FACTORY". Since it is supported in ardupilot, it should be defined in QGCApplication.cc

PR #9540, opened 2021-03-18
RadioComponentController fix for Custom builds

mnumanuyar:
Signed-off-by: numan.uyar <numan.uyar@titra.com.tr>

related to issue: #9539

### DIFF
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -15,6 +15,7 @@
 #include "AppSettings.h"
 #include "QGCFileDownload.h"
 #include "QGCCameraManager.h"
+#include "RadioComponentController.h"
 
 #include <QRegularExpression>
 #include <QDebug>
@@ -46,6 +47,12 @@ FirmwarePluginFactoryRegister* FirmwarePluginFactoryRegister::instance(void)
     }
 
     return _instance;
+}
+
+
+FirmwarePlugin::FirmwarePlugin(void)
+{
+    qmlRegisterType<RadioComponentController>       ("QGroundControl.Controllers",                       1, 0, "RadioComponentController");
 }
 
 AutoPilotPlugin* FirmwarePlugin::autopilotPlugin(Vehicle* vehicle)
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
I assume you are doing this because you have an ArduPilot custom build?

Since the RadioComponentController implementation is common to both PX4 and ArduPilot the correct change is to move this to the FirmwarePlugin base class. Moving it to QGCApplication causes it to be compiled in even if you are not using either PX4 or ArduPilot firmwares which is incorrect.

mnumanuyar:
You are absolutely correct! 
I think FirmwarePlugin does not have a constructor, should I create one and then us it in PX4FirmwarePlugin and APMFirmwarePlugin, or is there a better way?

DonLakeFlyer:
Yes, create a constructor for FirmwarePlugin. I don't think you need to mod px4 or apm firmware plugin code to make it work,.

mnumanuyar:
It should be ok now

src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -48,6 +49,12 @@ FirmwarePluginFactoryRegister* FirmwarePluginFactoryRegister::instance(void)
     return _instance;
 }
 
+
+FirmwarePlugin::FirmwarePlugin(void)
+{
+    qmlRegisterType<RadioComponentController>       ("QGroundControl.Controllers",                       1, 0, "RadioComponentController");
```
DonLakeFlyer, line 55: Can you pull the tabs out of here?

src/FirmwarePlugin/FirmwarePlugin.h
```diff
@@ -41,6 +41,8 @@ class FirmwarePlugin : public QObject
     Q_OBJECT
 
 public:
+    FirmwarePlugin ();
```
DonLakeFlyer, line 44: Also the extra space here to match coding style

DonLakeFlyer (changes requested):
Should be all god after the minor spacing fixes

DonLakeFlyer:
@mnumanuyar Can you update?

DonLakeFlyer:
@mnumanuyar Can you update?

mnumanuyar:
sorry, I posted somewhere else but forget to put it also here. Unfortunately I am a little busy with work related stuff. I will probably have more time in 2-3 weeks.

mnumanuyar:
Also thank you for merging and care

PR #9553, opened 2021-03-23
Vehicle.cc: Support Ardupilot in hobbsMeter()

Davidsastresas:
hobbs meter was working only for PX4. In ardupilot the parameter for this matter is STAT_FLTTIME. This PR adds support for this.

I separated the debug log messages for each of the cases, let me know if any change is needed here. Thanks!

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -566,6 +566,24 @@ QObject* APMFirmwarePlugin::_loadParameterMetaData(const QString& metaDataFile)
     return metaData;
 }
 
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
So I've never looking at the initial implementation of this. It is incorrect with respect to allowing firmware specific code to bleed into mainline source instead of being abstracted through a FirmwarePlugin method. Could you fix this up. Should be fairly straight forward to abstract it for any firmware type.

Davidsastresas:
Totally agree. Sure, I will take a look at it over this week. Thanks!

Davidsastresas:
I think it should be ok now! let me know if any modification is needed. Thanks!

LorenzMeier (approved):
Thanks!

PR #9554, opened 2021-03-23
COMPONENT_INFORMATION extension and rework

bkueng:
From https://github.com/mavlink/mavlink/pull/1610, will need to be updated when https://github.com/mavlink/mavlink/pull/1610 is merged.

---
@DonLakeFlyer could it be that something is still running when stopping the mock link?
I'm doing:
- start PX4 mock link
- stop mock link
- start PX4 SITL
- output (note the `Received old packet seqNum expected`, which I don't get w/o starting the mock link):
```

_parametersReady true

VehicleLinkManagerLog: _removeLink: "7286270"
~MockLink MockLink(0x7286270)
Adding target QHostAddress("127.0.0.1") 18570
Temp log "/tmp/FlightData279216.mavlink"
VehicleLinkManagerLog: _addLink: "UDP Link (AutoConnect)" "39b78d0"
ComponentInformationManagerLog: Requesting component information "COMP_METADATA_TYPE_GENERAL"
ComponentInformationManagerLog: Downloading metadata json "mftp://etc/extras/component_general.json.xz"
FTPManagerLog: download fromURI: "mftp://etc/extras/component_general.json.xz" to: "/tmp"
FTPManagerLog: _downloadState.fullPathOnVehicle:_downloadState.fileName "/etc/extras/component_general.json.xz" "component_general.json.xz"
FTPManagerLog: _sendRequestExpectAck opcode: "Open File RO" seqNumber: 1
FTPManagerLog: _mavlinkMessageReceived: hdr.opcode:hdr.req_opcode:seqNumber "Ack" "Open File RO" 2
FTPManagerLog: _openFileROAckOrNak: Ack  - sessionId:openFileLength 0 240
FTPManagerLog: _burstReadFileWorker: starting burst at offset:firstRequest:retryCount 0 true 0
FTPManagerLog: _sendRequestExpectAck opcode: "Burst Read File" seqNumber: 3
FTPManagerLog: _mavlinkMessageReceived: Received old packet seqNum expected:actual 4 2 hdr.opcode:hdr.req_opcode "Ack" "Open File RO"
FTPManagerLog: _mavlinkMessageReceived: hdr.opcode:hdr.req_opcode:seqNumber "Ack" "Burst Read File" 4
FTPManagerLog: "_burstReadFileAckOrNak: Ack offset(0) size(239) burstComplete(0)"
FTPManagerLog: _mavlinkMessageReceived: Received old packet seqNum expected:actual 5 4 hdr.opcode:hdr.req_opcode "Ack" "Burst Read File"
FTPManagerLog: _mavlinkMessageReceived: hdr.opcode:hdr.req_opcode:seqNumber "Ack" "Burst Read File" 5
FTPManagerLog: "_burstReadFileAckOrNak: Ack offset(239) size(1) burstComplete(0)"
FTPManagerLog: _mavlinkMessageReceived: Received old packet seqNum expected:actual 6 5 hdr.opcode:hdr.req_opcode "Ack" "Burst Read File"
FTPManagerLog: _mavlinkMessageReceived: hdr.opcode:hdr.req_opcode:seqNumber "Nak" "Burst Read File" 6
FTPManagerLog: _burstReadFileAckOrNak EOF
FTPManagerLog: _sendRequestExpectAck opcode: "Reset Sessions" seqNumber: 7
FTPManagerLog: _mavlinkMessageReceived: Received old packet seqNum expected:actual 8 6 hdr.opcode:hdr.req_opcode "Nak" "Burst Read File"
FTPManagerLog: _mavlinkMessageReceived: hdr.opcode:hdr.req_opcode:seqNumber "Ack" "Reset Sessions" 8
```


src/FactSystem/FactMetaData.cc
```diff
@@ -1436,6 +1436,8 @@ FactMetaData* FactMetaData::createFromJsonObject(const QJsonObject& json, QMap<Q
 
     if (json.contains(_categoryJsonKey)) {
         metaData->setCategory(json[_categoryJsonKey].toString());
+    } else {
+        metaData->setCategory("Standard");
```
bkueng, line 1440: @DonLakeFlyer I added this to preserve the current PX4 behavior. Are there cases (e.g. other components) where this is not desired?

DonLakeFlyer:
I would guess a bug in `FTPManager::_expectedIncomingSeqNumber` not being set/reset properly from a previous protocol sequence.

src/FactSystem/FactMetaData.cc
DonLakeFlyer: The "Standard" category concept is a PX4 firmware specific thing. The QGC default is actually "Other" which is setup up in the constructor. So this will change behavior I think moving ArduPilot params to a new category in some cases. Why is this needed?

DonLakeFlyer: It's all fairly ancient history so I may have it somewhat wrong but it's something along those lines.

bkueng: Ok I make sure PX4 sets this then

DonLakeFlyer:
> I would guess a bug in `FTPManager::_expectedIncomingSeqNumber` not being set/reset properly from a previous protocol sequence.

Wait no. That's not quite right. The FTPManager is on the Vehicle so you should get a new one when you create the new SITL Vehicle. So not sure.

src/FactSystem/FactMetaData.cc
DonLakeFlyer: Make sense. I believe I had to add a workaround in the xml parser for the metadata for the missing categories ages ago.

bkueng:
>  Wait no. That's not quite right. The FTPManager is on the Vehicle so you should get a new one when you create the new SITL Vehicle. So not sure.

Mystery solved: it was PX4 broadcast that was enabled for a short time on master, which led to duplicate packets sent out until it got connected. So not specifically related to the mock link.

src/Vehicle/CompInfo.h
```diff
@@ -27,17 +28,43 @@ class CompInfo : public QObject
...
+
+    const COMP_METADATA_TYPE  type;
+    Vehicle* const      vehicle                = nullptr;
+    const uint8_t       compId                 = MAV_COMP_ID_ALL;
+
+private:
+    friend class CompInfoGeneral;
+
+    struct Uris {
+        bool                crcMetaDataValid            = false;
```
DonLakeFlyer, line 53: I don't get these `Valid` flags? Why isn't a CRC always required?

src/Vehicle/CompInfoGeneral.cc
```diff
@@ -0,0 +1,93 @@
...
+        return;
+    }
+
+    QJsonArray rgSupportedTypes = jsonObj[_jsonMetadataTypesKey].toArray();
+    for (QJsonValue typeValue : rgSupportedTypes) {
+        int type = typeValue["type"].toInt(-1);
+        if (type == -1)
+            continue;
+        Uris uris;
+        uris.uriMetaData = typeValue["uri"].toString();
```
DonLakeFlyer, line 78: I think it would be safer to validate the full type object values as well. I would assume `MetaData` is required and then the rest are optional. But if a uri is specified for any the crc is required.

src/Vehicle/CompInfoGeneral.cc
```diff
@@ -0,0 +1,93 @@
...
+        uris.crcMetaData = typeValue["fileCrc"].toVariant().toLongLong(); // Note: can't use toInt()
```
DonLakeFlyer, line 79: Not sure why `toInt` doesn't work. Could you add more detail to the comment as to why.

src/Vehicle/CompInfoGeneral.h
```diff
@@ -15,27 +15,30 @@
...
+
     // Overrides from CompInfo
     void setJson(const QString& metadataJsonFileName, const QString& translationJsonFileName) override;
 
 private:
-    QList<COMP_METADATA_TYPE>   _supportedTypes;
+    QMap<COMP_METADATA_TYPE, Uris>   _supportedTypes;
 
-    static const char*          _jsonSupportedCompMetadataTypesKey;
+    static const char*          _jsonMetadataTypesKey;
```
DonLakeFlyer, line 43: The pattern for json is to have the keys all in statics instead of using strings directly. This removes bugs from any possible need to use the keys in more than one place and having them be different or if the key name changes then a single code change affects all usage.

src/comm/MockLink.cc
```diff
@@ -1576,14 +1576,8 @@ bool MockLink::_handleRequestMessage(const mavlink_command_long_t& request, bool
...
         if (_firmwareType == MAV_AUTOPILOT_PX4) {
-            switch (static_cast<int>(request.param2)) {
-            case COMP_METADATA_TYPE_VERSION:
-                _sendVersionMetaData();
-                return true;
-            case COMP_METADATA_TYPE_PARAMETER:
-                _sendParameterMetaData();
-                return true;
-            }
+            _sendVersionMetaData();
```
DonLakeFlyer, line 1579: Shouldn't this be changed to `_sendGeneralMetaData`?

src/comm/MockLink.General.MetaData.json
```diff
@@ -4,5 +4,7 @@
     "modelName":                    "QGC MockLink Model",
     "firmwareVersion":              "1.0.0",
     "hardwareVersion":              "1.0.0",
-    "supportedCompMetadataTypes":   [ 0, 1 ]
-} 
\ No newline at end of file
+	"metadataTypes": [
+		{"type": 1, "uri": "mftp://[;comp=1]parameter.json.xz", "fileCrc": 1}
```
DonLakeFlyer, line 8: `"fileCrc": 1` - I assume this is because crc checking isn't implemented yet right? I could seem to find it.

src/comm/MockLink.cc
```diff
@@ -1619,45 +1613,21 @@ void MockLink::_sendVersionMetaData(void)
...
     mavlink_message_t   responseMsg;
 #if 1
-    char                metaDataURI[MAVLINK_MSG_COMPONENT_INFORMATION_FIELD_METADATA_URI_LEN]       = "mftp://[;comp=1]version.json";
+    char                metaDataURI[MAVLINK_MSG_COMPONENT_INFORMATION_FIELD_GENERAL_METADATA_URI_LEN]       = "mftp://[;comp=1]general.json";
 #else
-    char                metaDataURI[MAVLINK_MSG_COMPONENT_INFORMATION_FIELD_METADATA_URI_LEN]       = "https://bit.ly/31nm0fs";
+    char                metaDataURI[MAVLINK_MSG_COMPONENT_INFORMATION_FIELD_GENERAL_METADATA_URI_LEN]       = "https://bit.ly/31nm0fs";
 #endif
-    char                translationURI[MAVLINK_MSG_COMPONENT_INFORMATION_FIELD_TRANSLATION_URI_LEN] = "";
+    char                peripheralsMetaDataURI[MAVLINK_MSG_COMPONENT_INFORMATION_FIELD_PERIPHERALS_METADATA_URI_LEN]       = "https://bit.ly/31nm0fs";
```
DonLakeFlyer, line 1620: This is a bitly? Shouldn't it be included directly in the source for mocklink?

DonLakeFlyer (changes requested):

src/Vehicle/CompInfo.h
bkueng: For translation (and potentially fallback) there's no CRC, as the autopilot does not see that file. So I added the flag to all, but yes, for the main metadata it needs to be valid and this one could be removed.

src/Vehicle/CompInfoGeneral.cc
bkueng: Ok, I added a check for empty uri and crc set

src/Vehicle/CompInfoGeneral.cc
bkueng: Because it's a signed int and it returns 0 for >2^31 inputs. Qt 6 will come with a 64bit version.

src/Vehicle/CompInfoGeneral.h
bkueng: I generally agree, but this is so localized with a single usage that I don't consider it worth doing.

src/comm/MockLink.cc
bkueng: Indeed, updated

src/comm/MockLink.cc
bkueng: Ah I didn't intend to set this one at all, just a copy-paste left-over.

src/Vehicle/CompInfoGeneral.h
DonLakeFlyer: That is always how things start. But is almost never how they end up.  Someone else comes in add new stuff later on, dups the string and boom, you're in trouble.

PR #9550, opened 2021-03-20
Plan: Fix Loiter directions

DonLakeFlyer:
* One form of recalc code wasn't taking into account loiter direction setting
* Fix #9514

PR #9555, opened 2021-03-23
Plan - FW Landing: Fix Loiter directions

DonLakeFlyer:
Cherry pick #9550 to master


DonLakeFlyer:
@antiheavy There should be a new daily for this in an hour or so

PR #9567, opened 2021-03-28
libs/shapelib: Skip installing executables

thopiekar:
.. which don't exist. The whole CMake file is a mess actually.
It would be much more elegant to add an option to turn it off.
Best of all in the upstream project and then adding it as a submodule here.

Anyway, this will workaround an error when running "make install".

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com>


thopiekar:
Merged via https://github.com/mavlink/qgroundcontrol/pull/9569

PR #9568, opened 2021-03-28
qmake: Adding QGC_ENABLE_UNIX_INSTALLATION

thopiekar:
So we can install via "make install" like normal people and don't need to rely on own post-build-script copying stuff from right to left.

I was working on this while trying to make qmake working with flatpak's builder.
However, there was an issue - I think the make build template is not bug-free - so I had to switch to CMake.
Eventually, this might be useful for someone of you. Therefore sharing it :)

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com>

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -1447,5 +1447,27 @@ contains (CONFIG, QGC_DISABLE_INSTALLER_SETUP) {
     include(QGCPostLinkInstaller.pri)
 }
 
+#
+# Unix installation
+#
+
+contains (CONFIG, QGC_ENABLE_UNIX_INSTALLATION) {
+    message("binary: $${OUT_PWD}/$${DESTDIR}/QGroundControl")
+
+    bin_qgroundcontrol.path = $${PREFIX}/bin/
+    bin_qgroundcontrol.files = $${OUT_PWD}/$${DESTDIR}/QGroundControl
+
+    share_qgroundcontrol.path = $${PREFIX}/share/qgroundcontrol/
+    share_qgroundcontrol.files = $${IN_PWD}/resources/
+
+    share_pixmaps.path = $${PREFIX}/share/pixmaps/
+    share_pixmaps.files = $${IN_PWD}/resources/icons/qgroundcontrol.png
+
+    share_applications.path = $${PREFIX}/share/applications/
+    share_applications.files = $${IN_PWD}/deploy/qgroundcontrol.desktop
+
+    INSTALLS += bin_qgroundcontrol share_qgroundcontrol share_pixmaps share_applications
+}
+
 DISTFILES += \
     src/QmlControls/QGroundControl/Specific/qmldir
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -1447,27 +1447,22 @@ contains (CONFIG, QGC_DISABLE_INSTALLER_SETUP) {
     include(QGCPostLinkInstaller.pri)
 }
 
-#
-# Unix installation
-#
-
-contains (CONFIG, QGC_ENABLE_UNIX_INSTALLATION) {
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

DonLakeFlyer:
This seems to add yet another installation type which could break further down the road. At this point there is only AppImage which is supported. Not sure why this is needed as an additional thing?

thopiekar:
As mentioned above, it is a result of a work in process. I was working on flatpak software distribution and within flatpak manifests, which are needed to build flatpaks, you can choose between project templates. This can be for example: cmake, qmake, autotools or whatever you want to do manually via "simple".
I knew before starting with my work, that qmake is mainly used and tried to build something on top of it. During this I got complaints from flatpak-builder that it tried to run "make install" but there was "nothing to do".
Therefore, to avoid any interferences with the current software distributions, I decided to add a define, which turns the needed install steps selectively on.

Generally, I tend not to throw things away, if I'm not certain that someone else might still need it. Therefore offering it here to get merged or not. Depends on you whether you see an purpose or not :wink: 
For myself, it won't be needed when looking at the current flatpak progress, because I gave up with qmake and continued with cmake. If I heard correctly on Slack, qmake is going to die with Qt6 and will be replaced, therefore I decided to continue with CMake since it is the future in longer term. :slightly_smiling_face:

DonLakeFlyer:
> If I heard correctly on Slack, qmake is going to die with Qt6 and will be replaced

I've been hearing that for years now. I'll believe it when I see it. Also who knows when we'll actually get to Qt 6. First sepect before that is 5.15. As it currently stands cmake cannot be used as the one build system to spit out all OS versions of QGC. Only qmake is capable of that. And CI only uses qmake. So anything which is cmake only will not see much of the light of day for now. The folks using cmake are a subset of QGC users who mostly only use Linux. I for example never use cmake build scipts since they don't replicate all the things I need.

thopiekar:
Haha, ok, didn't know that this exists for years. For me, it is quite new that qmake is going to die and actually never heard an official statement by the Qt company. 

What are all these things? Is it possible to create a list of things that are missing in CMake?
I don't have any overview of what needs to be added into CMake and generally done before we could move away from qmake.

But back to the topic also:  How do you think about this PR? Shall I retry without the QGC_ENABLE_UNIX_INSTALLATION switch as done in the CMake variant of this PR and see what the CI/CD says?
Or is there no interest in having this in qmake merged generally? :confused:

DonLakeFlyer:
> What are all these things? Is it possible to create a list of things that are missing in CMake?

Simpler to give examples:
* Build, deploy to device and debug an Android build using Qt Creator
* Generate an XCode build project so you can build the iOS version in XCode

DonLakeFlyer:
A traditional "make install" thing is only useful for a single person no a single machine. Not super useful for folks like third party commercial users of custom build who want to distribute thing to the customers. So doesn't seem super important. Seems like building able to create something like an AppImage directly from the build with CONFIG+=installer which you can then just use or distribute would be more useful.

thopiekar:
Ok. I made a quick scan of the repo and it doesn't seem that "make install" is being used anywhere.
I'm going to make a test and remove the QGC_ENABLE_UNIX_INSTALLATION config switch as I did with @patrickelectric for CMake. Since it doesn't seem that the install target is not being used anywhere, it shouldn't hurt anyone, I guess.

thopiekar:
Looks fine :sweat_smile:

DonLakeFlyer:
@thopiekar Can you rebase this into a single commit with a better commit message?

thopiekar:
Hmm, if I put any more time into this, I would better close this PR and #9585 by doing the following:
* Merging all commits here into a single commit as you wish
* Put the changes there into one commit regarding airmap
* Rename the whole PR into something generic: "qmake: improvements around Linux"

@DonLakeFlyer What do you think? :)

DonLakeFlyer:
Makes sense

PR #9179, opened 2020-11-18
[Hold for 4.2] deploy: windows: Moving files into common distribution directory

thopiekar:
The general goal is to tidy up the project folder and give better focus on other contents of the project.
As a first iteration, I would like to tidy up files, which are Windows related. Technically seen I'm pretty sure that we won't notice any differences, since some libraries and images have been moved and some config files cleaned up.

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com>

PR #9569, opened 2021-03-28
CMake: Bundle of fixes

thopiekar:
Basically needed for my flatpak build to work.
What you see is what you get.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required(VERSION 3.10)
 
 project(QGroundControl LANGUAGES C CXX)
 
+include(GNUInstallDirs)
+
 set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug;Release;RelWithDebInfo;MinSizeRel;Coverage")
 
 set(CMAKE_CXX_STANDARD 14)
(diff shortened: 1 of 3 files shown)


thopiekar:
Hope we can speed PRs like #9179 somehow.
Would be glad to do some cleanup - I'm not happy about adding another desktop file here, but I don't see how to generate files in qmake, which is the major build tool here afaik.
Hope it dies soon, since there are much more elegant ways to handle files in CMake :smile:

libs/shapelib/CMakeLists.txt
```diff
@@ -189,7 +189,7 @@ foreach(executable ${executables})
     get_target_property(${executable}_LOC ${executable} LOCATION)
     file(APPEND ${CMAKE_CURRENT_BINARY_DIR}/script.sed "s?\\./${executable}?${${executable}_LOC}?\n")
   endif(BUILD_TEST)
-  install(TARGETS ${executable} DESTINATION ${CMAKE_INSTALL_BINDIR})
+  # install(TARGETS ${executable} DESTINATION ${CMAKE_INSTALL_BINDIR})
```
patrickelectric, line 192: You just remove the line, there is no reason to comment it, that's why we use version control.

CMakeLists.txt
```diff
@@ -186,6 +186,29 @@ endif()
 
 target_link_libraries(QGroundControl PRIVATE qgc)
 
+option(ENABLE_INSTALL_TARGET "Enable real installation of QGC when running install" OFF)
```
patrickelectric, line 189: Why is this an option ? If someone is using `make install` in QGC it should install QGC.
This should be the default behavior.

patrickelectric (changes requested):
Great patch, just need some minor corrections to get it merged.

libs/shapelib/CMakeLists.txt
thopiekar: Well, I was confused why the shapelib sources have been added as a copy instead of a submodule. Since I'm planning to put some effort into cleaning up this, too, I actually didn't care whether it looks messy ot not.
But yeah, far enough. I'm going to clean it up.

CMakeLists.txt
thopiekar: Since I'm not sure whether other projects or any of our distributions (Windows installer, Linux AppImage, etc.) will hit any issues.
So I added the option as I need it explicitly for my Flatpak builds.
But yes, I can remove it and see what the CI/CD says.
I'm just not able test the macOS dmg..

Or wait.. is qmake not being used everyhere at the moment?

thopiekar:
Thanks! And thank *you* for the quick review! 😉

CMakeLists.txt
thopiekar: While crossing fingers - fixed via: https://github.com/mavlink/qgroundcontrol/pull/9569/commits/2a6af31adfadbe4095ad10e275b892d74dbb344d

libs/shapelib/CMakeLists.txt
thopiekar: https://github.com/mavlink/qgroundcontrol/pull/9569/commits/c82b10a477b282795120da0ef89032fd7edb0ad9

thopiekar:
All done, waiting for the results of CI/CD. :)

dagar (approved):

patrickelectric (approved):
I have manually removed the merge commit (check [`git rebase`](https://git-scm.com/docs/git-rebase)) and [squashed](https://blog.chilij.com/posts/git-fixup-autostash-autosquash/) unnecessary commits.

thopiekar:
Thanks @patrickelectric

PR #9572, opened 2021-04-01
Fix vehicle map item opacity

Hunter522:
Noticed at some point the active vehicle map icon kept being transparent instead of fully opaque. After investigating found that it relied on an old property `Vehicle.active` which got removed in 66753e2111b1b7139664f8e354f2919c5b2e37f2

This PR fixes that and brings back the opaque icon.

# Screenshots

Current:

Screenshot from 2021-03-31 22-37-30

Fix:

Screenshot from 2021-03-31 22-38-29

Fix (with mutliple vehicles):

Screenshot from 2021-03-31 22-51-48

### DIFF
--- a/src/FlightMap/MapItems/VehicleMapItem.qml
+++ b/src/FlightMap/MapItems/VehicleMapItem.qml
@@ -31,6 +31,7 @@ MapQuickItem {
     anchorPoint.y:  vehicleItem.height / 2
     visible:        coordinate.isValid
 
+    property var    _activeVehicle: QGroundControl.multiVehicleManager.activeVehicle
     property bool   _adsbVehicle:   vehicle ? false : true
     property real   _uavSize:       ScreenTools.defaultFontPixelHeight * 5
     property real   _adsbSize:      ScreenTools.defaultFontPixelHeight * 2.5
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

src/FlightMap/MapItems/VehicleMapItem.qml
```diff
@@ -41,7 +42,7 @@ MapQuickItem {
         id:         vehicleItem
         width:      vehicleIcon.width
         height:     vehicleIcon.height
-        opacity:    vehicle ? (vehicle.active ? 1.0 : 0.5) : 1.0
+        opacity:    (vehicle && _activeVehicle) ? ((vehicle === _activeVehicle) ? 1.0 : 0.5) : 1.0
```
DonLakeFlyer, line 45: Yes, good bug find. I think this would make the code a little simpler/clearer as to what needs to happen with respect to ADSB vehicle as well as multiple vehicles with only one being active:

`opacity:    _adsbVehicle ||  vehicle === _activeVehicle ? 1.0 : 0.5`

Hunter522: That's much simpler and I didn't even think about ADSB vehicles. Done in latest commit.

DonLakeFlyer:
Thanks for the fix @Hunter522

DonLakeFlyer:
@Hunter522 Can you cherry pick to a pull for the Stable_V4.1 branch?

Hunter522:
@DonLakeFlyer, sure see PR https://github.com/mavlink/qgroundcontrol/pull/9589

PR #9575, opened 2021-04-04
ScreenTools.qml: decouple font from default width/height

Davidsastresas:
I realized when changing the fonts in QGC all the proportions in the GUI go nuts. This is because screentools takes the base measurement from the font we are using.

This PR decouples the basic units of GUI from font used, so a change in font doesn't mess up with GUI proportions. I left hardcoded "opensans" which is the default until now.

I've tested adding new fonts and now the dimensions of the GUI are unaffected. 

I am not sure if this is the best way to approach this, let me know your thoughts please. Thanks!

### DIFF
--- a/src/QmlControls/ScreenTools.qml
+++ b/src/QmlControls/ScreenTools.qml
@@ -162,7 +162,7 @@ Item {
     Text {
         id:     _textMeasure
         text:   "X"
-        font.family:    normalFontFamily
+        font.family:    "opensans"
         property real   fontWidth:    contentWidth
         property real   fontHeight:   contentHeight
         Component.onCompleted: {


Antiheavy:
Any chance these two issues are related/solved by your approach?

DonLakeFlyer:
> Any chance these two issues are related/solved by your approach?

Doubt it. Those are likely caused by an iOS version which is newer than the version of Qt we are using. That will likely get fixed when we move to Qt 5.15.

Multinnov:
I just checked those issues and I don't think this PR makes a difference. This PR only makes possible to change the font used by QGC and keep all the gui proportions the same. 

Before this PR, GUI proportions were tied to font width/height, so some fonts messed up a lot gui proportions.

Davidsastresas:
I just checked those issues and I don't think this PR makes a difference. This PR only makes possible to change the font used by QGC and keep all the gui proportions the same. 

Before this PR, GUI proportions were tied to font width/height, so some fonts messed up a lot gui proportions.

DonLakeFlyer:
> This is because screentools takes the base measurement from the font we are using.

Which is exactly how it is supposed to work. All direct positioning in QGC is based on the font size being used. This allows the UI to be created in a resolution independent way and also adjust if the user adjusts the font sizes in settings.

If you can provide repro steps for exactly what you are seeing when something goes wrong I can help you out.

Davidsastresas:
I just changed the font by another one, super similar to opensans, and I noticed the left toolbar, the one where you command take off, actions etc, was super, super wide.

It is actually solved to me already, I just thought It could be interesting to made a PR, but if that is the way it should work no problem at all, I can close the PR. thanks!

DonLakeFlyer:
There is possibly still a bug here for folks that want to change fonts. But more likely a layout bug in the qml for the toolbar. Can you provide a simple diff in an issue which I can then use to reproduce? Also include a before/after screenshot would be helpful.

Davidsastresas:
I haven't thought on that possibility. Sure!

Davidsastresas:
I just tested again and for some reason I am not seeing the actions toolbar super screwed in dimensions as before. It is possible that I messed up something back when I tested the new fonts with scaling or whatever, but I am not able to reproduce it now.

Closing this PR, sorry for the trouble!

PR #9582, opened 2021-04-06
Applying some general freedesktop conventions

thopiekar:
As needed for great flatpak integration.


deploy/org.mavlink.qgroundcontrol.metainfo.xml
```diff
@@ -0,0 +1,39 @@
...
+    </p>
+  </description>
+  
+  <launchable type="desktop-id">org.mavlink.qgroundcontrol.desktop</launchable>
+  <screenshots>
+    <screenshot type="default">
+      <image>https://docs.qgroundcontrol.com/assets/quickstart/ConnectedVehicle.jpg</image>
+    </screenshot>
+  </screenshots>
+..
```
patrickelectric, line 30: This two dots should be here ?

thopiekar: Heck?! Let me correct this.

thopiekar: Done. Like your reviews @patrickelectric - you have eagle eyes :stuck_out_tongue:

patrickelectric (approved):

deploy/org.mavlink.qgroundcontrol.metainfo.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.mavlink.qgroundcontrol</id>
+  
+  <name>QGroundControl</name>
+  <summary>UAS ground control station</summary>
+  
+  <metadata_license>MIT</metadata_license>
```
thopiekar, line 8: Maybe this needs discussion still. Within the generator website, I couldn't choose either GPL and/or Apache. :confused:

deploy/org.mavlink.qgroundcontrol.metainfo.xml
```diff
@@ -0,0 +1,39 @@
...
+  <project_license>GPL-3.0-only</project_license>
```
thopiekar, line 9: And here only one of the licenses. Don't know whether mixes are possible within these metadata files.

thopiekar (commented):
Got some license questions open still.

deploy/org.mavlink.qgroundcontrol.metainfo.xml
thopiekar: Found a solution to declare GPL3 and Apache2.

deploy/org.mavlink.qgroundcontrol.metainfo.xml
thopiekar: Found a solution to declare GPL3 and Apache2.

LorenzMeier (approved):

mrpollo:
@thopiekar can you please help us with the instructions on how to build and deploy a flatpak bundle? we need this to be documented on the QGC docs

patrickelectric:
@thopiekar, can also be helpful to have a cmake command to generate as output the flatpak itself, this will make the documentation simple and straightforward for everybody.

thopiekar:
@mrpollo Will do so. But need #9585 first. After that I'll create a PR with the flatpak manifest. However, without this PR building the flatpak will only be possible without airmap since I can't download anything during build-time.

@patrickelectric That shouldn't be difficult. Nevertheless, I'll keep with the plans discussed with @DonLakeFlyer and keep focusing on qmake.
Something I mentioned on Slack already: Does any of you have an overview about the current status of the CMake files? Do we build with the Airmap SDK there, for example? Any other differences. I think we should collect these things somewhere. :)

PR #9591, opened 2021-04-15
Fix internet access test

### LINKED ISSUE #9484: Joystick: MANUAL_CONTROL messages are not regular
## Expected Behavior
MANUAL_CONTROL messages should be sent with a regular interval when there is a joystick connected (or virtual joystick is enabled.

## Current Behavior
Every 30 seconds, something disrupts the messages, causing large latency spikes

## Steps to Reproduce:
1. Open SITL (any?)
2. Open QGC
3. Close QGC
4. Check latest telemetry log

## System Information

- QGC build: QGC 88b2b8f5
- Ardusub SITL master

## Detailed Description
This causes ArduSub to complain about "Lost Manual Control". Which is correct. While this is usually pretty quick, I just looked a a users tlog where this freeze peaked at 25 seconds! Every time the blue line rises above the orange one, ArduSub emits a "Manual Control Lost" warning.

image
User log:
[2021-02-17 17-04-38.zip](https://github.com/mavlink/qgroundcontrol/files/6036717/2021-02-17.17-04-38.zip)

## Investigation

I did some digging already, and this seems to be caused by [QGCMapEngine::testInternet()](https://github.com/mavlink/qgroundcontrol/blob/master/src/QtLocationPlugin/QGCMapEngine.cpp#L390) (the only relevant timer fired every 30s). 

This is a regular log of SITL:
image

This is SITL with `checkInternet=false` in `[General]` (is there no UI for it?)
 
image

So it seems that on some setups, the internet check can take long enough to disrupt regular operation of QGC...

patrickelectric:
Fix #9484


Williangalvani:
This seems to improve things a lot. The peaks of 300~400ms are down to 80ms
image

mrpollo:
here's the failed test on travis @patrickelectric 

```
FAIL!  : VehicleLinkManagerTest::_multiLinkSingleVehicleTest() Compared pointers are not the same
2207   Loc: [/home/travis/build/mavlink/qgroundcontrol/src/Vehicle/VehicleLinkManagerTest.cc(167)]
2208QDEBUG : VehicleLinkManagerTest::_multiLinkSingleVehicleTest() ~MockLink MockLink(0x55729f62ddc0)
2209QDEBUG : VehicleLinkManagerTest::_multiLinkSingleVehicleTest() ~MockLink MockLink(0x55729f53e6e0)
2210QDEBUG : VehicleLinkManagerTest::_connectionRemovedTest() MockLink MockLink(0x55729f53e6e0)
2211QDEBUG : VehicleLinkManagerTest::_connectionRemovedTest() say " manual flight mode"
2212QDEBUG : VehicleLinkManagerTest::_connectionRemovedTest() _parametersReady true
2213QDEBUG : VehicleLinkManagerTest::_connectionRemovedTest() say "communication lost"
2214QDEBUG : VehicleLinkManagerTest::_connectionRemovedTest() ~MockLink MockLink(0x55729f53e6e0)
```

mrpollo:
@patrickelectric looking at the `VehicleLinkManagerTest` tests the only `QCOMPARE` that is actually comparing objects rather than booleans, strings, or numbers is [on line 167, comparing the `mockLink1` to a `QSignalSpy` which is calling `Vehicle::initialConnectComplete`](https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/VehicleLinkManagerTest.cc#L167)

[I tracked down the chain of events until I stumbled upon this ](https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/InitialConnectStateMachine.cc#L111)
```
vehicle->_waitForMavlinkMessage(_waitForProtocolVersionResultHandler, connectMachine, MAVLINK_MSG_ID_PROTOCOL_VERSION, 1000);
```
which is setting the timeout for `void Vehicle::_waitForMavlinkMessageMessageReceived(const mavlink_message_t& message)` to 1000ms which might be leading to the timeout in TravisCI

https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L2976

Just a theory, since I have observed other timeouts on `VehicleLinkManagerTest` tests related to timeouts that mostly hang on waits

patrickelectric:
@mrpollo I have just rebased over master, should be fine now.

mrpollo:
Awesome thanks @patrickelectric, glad the flaky tests are gone, did you get a chance to look at PR https://github.com/mavlink/qgroundcontrol/pull/9595 ?

It's something we should probably discuss at the next developer call.

booo:
Unfortunately this brakes the build on my system:

```/home/borgers/projects/qgroundcontrol/src/QtLocationPlugin/QGCTileCacheWorker.cpp: In member function ‘void QGCCacheWorker::_lookupReady(QHostInfo)’:
/home/borgers/projects/qgroundcontrol/src/QtLocationPlugin/QGCTileCacheWorker.cpp:1200:79: error: use of ‘auto’ in lambda parameter declaration only available with ‘-std=c++14’ or ‘-std=gnu++14’
 1200 |         connect(socket, &QAbstractSocket::errorOccurred, this, [this, socket](auto error) {
      |
 ```

I guess we should use explicit types here as suggested in this thread:

https://stackoverflow.com/questions/60983917/error-use-of-auto-in-lambda-parameter-declaration-only-available-with-std-c

Issue #9346, opened 2021-01-18
Intermittent failures with VehicleLinkManagerTest


DonLakeFlyer:
```
********* Start testing of VehicleLinkManagerTest *********
1931Config: Using QtTest library 5.12.6, Qt 5.12.6 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 5.3.1 20160406 (Red Hat 5.3.1-6))
1932PASS   : VehicleLinkManagerTest::initTestCase()
1933QDEBUG : VehicleLinkManagerTest::_simpleLinkTest() MockLink MockLink(0x564ad3127870)
1934QDEBUG : VehicleLinkManagerTest::_simpleLinkTest() say " manual flight mode"
1935QDEBUG : VehicleLinkManagerTest::_simpleLinkTest() _parametersReady true
1936QDEBUG : VehicleLinkManagerTest::_simpleLinkTest() ~MockLink MockLink(0x564ad3127870)
1937PASS   : VehicleLinkManagerTest::_simpleLinkTest()
1938QDEBUG : VehicleLinkManagerTest::_simpleCommLossTest() MockLink MockLink(0x564ad329e9b0)
1939QDEBUG : VehicleLinkManagerTest::_simpleCommLossTest() say " manual flight mode"
1940QDEBUG : VehicleLinkManagerTest::_simpleCommLossTest() _parametersReady true
1941QDEBUG : VehicleLinkManagerTest::_simpleCommLossTest() say "communication lost"
1942QDEBUG : VehicleLinkManagerTest::_simpleCommLossTest() say "communication regained"
1943QDEBUG : VehicleLinkManagerTest::_simpleCommLossTest() QGCApplication::showAppMessage unittest title:message "QGroundControl_unittest" "Communication regained"
1944QDEBUG : VehicleLinkManagerTest::_simpleCommLossTest() say "communication lost"
1945QDEBUG : VehicleLinkManagerTest::_simpleCommLossTest() ~MockLink MockLink(0x564ad329e9b0)
1946PASS   : VehicleLinkManagerTest::_simpleCommLossTest()
1947QDEBUG : VehicleLinkManagerTest::_multiLinkSingleVehicleTest() MockLink MockLink(0x564ad329e9b0)
1948QDEBUG : VehicleLinkManagerTest::_multiLinkSingleVehicleTest() MockLink MockLink(0x564ad34d0fa0)
[...]

PR #9595, opened 2021-04-16
Stability Fixes - Vehicle Link Manager Test

keith-bennett-airmap:
This PR aims to fix [#9346](https://github.com/mavlink/qgroundcontrol/issues/9346). For AirMap, the JIRA ticket [BOSS-616].

Several issues identified and fixed.

* Many racey uses of [`if (weak_ptr::expired()) {... weak_ptr::lock() ... }`](https://en.cppreference.com/w/cpp/memory/weak_ptr/expired)
  racey weak_ptr expired lock

* `MockLink` had its own `_mavlinkChannel` shadowing the inherited `LinkInterface::_mavlinkChannel`. So `MockLink` always used channel 0 for lots of work which caused `VehicleLinkManagerTest::_multiLinkSingleVehicleTest` to fail very frequently when its two links would use the same mavlink buffer to parse a reply to the vehicle. Its failure would poison subsequent tests. 
  MockLink _handleIncomingMavlinkBytes racey
  That diagram is huge and hard to follow because it reflects how deep the two call stacks are with signals. This table might be easier to understand (top is most-recent call):
  | Thread 1 | Thread 2 |
  | :- | -: |
  | [mavlink_parse_char](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MockLink.cc#L550) (racey->) | (<-racey) (same) |
  | [MockLink::_handleIncomingMavlinkBytes](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MockLink.cc#L543) | (same) |
  | [MockLink::_writeBytesQueued](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MockLink.cc#L514) ([signal handler](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MockLink.cc#L94)) | (same) |
  | [MockLink::writeBytesQueuedSignal](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MockLink.cc#L501) (signal) | (same) |
  | [MockLink::_writeBytes](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/LinkInterface.cc#L52) | (same) |
  | [LinkInterface::writeBytesThreadSafe](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/Vehicle.cc#L1670) | [LinkInterface::writeBytesThreadSafe](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/MultiVehicleManager.cc#L388) |
  | [Vehicle::sendMessageOnLinkThreadSafe](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/Vehicle.cc#L2838) | [MultiVehicleManager::_sendGCSHeartbeat](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/MultiVehicleManager.cc#L132) |
  | [Vehicle::_sendMavCommandFromList](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/Vehicle.cc#L2753) | ([signal handler](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/MultiVehicleManager.cc#L60)) [MultiVehicleManager::_vehicleHeartbeatInfo](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/MultiVehicleManager.cc#L70) |
  | [Vehicle::_sendMavCommandWorker](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/Vehicle.cc#L2637) | (signal) [MAVLinkProtocol::vehicleHeartbeatInfo](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MAVLinkProtocol.cc#L315) |
  | [Vehicle::sendMavCommandWithHandler](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/InitialConnectStateMachine.cc#L112) | ([signal handler](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/LinkManager.cc#L153)) [MAVLinkProtocol::receiveBytes](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MAVLinkProtocol.cc#L195) |
  | [InitialConnectStateMachine::_stateRequestCapabilities](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/Vehicle.cc#L239) ([initial state function](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/Vehicle/InitialConnectStateMachine.cc#L22)) | (signal) [MockLink::bytesReceived](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MockLink.cc#L493) |
  | | [MockLink::respondWithMavlinkMessage](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MockLink.cc#L320) |
  | | [MockLink::_sendHeartBeat](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MockLink.cc#L208) |
  | | (thread start & timer) [MockLink::_run10HzTasks](https://github.com/mavlink/qgroundcontrol/blob/2e093b834983f2956fec507938f142fd61915f92/src/comm/MockLink.cc#L145-L171) |

* `MAVLinkProtocol::receiveBytes` didn't lock the link's shared_ptr which meant undefined behavior was possible (dereferencing a raw pointer) if the connection closed before `MAVLinkProtocol` wasn't done. For example, when the mavlink channel 0 buffer was corrupted, it can stall the initial connection state machine. Then `_multiLinkSingleVehicleTest` gives up waiting for the initial connection complete signal then test teardown occurs. But if there's a heartbeat signal pending to be processed in the second thread then the link may be destroyed by the time the MAVLinkProtocol wants to access it. This undefined behavior often manifested in subsequent test failures.

* `_multiLinkSingleVehicleTest` also didn't understand that two links starting at the same time could have _either_ one of them set as the primary link (basically depends on which one starts sending heartbeats first).

The solutions aimed to fix the above issues are as follows:
* added the ability for a `LinkInterface` to desire an additional auxiliary mavlink channel and currently the `MockLink` makes use of the auxiliary channel for its reverse-direction message processing.
* added a mutex around the `MockLink`'s use of the auxiliary channel and ensured that the `MockLink` no longer shadows the `LinkInterface::_mavlinkChannel`.
* fixed a bunch of racey uses of `weak_ptr`
* ensured that `MAVLinkProtocol` holds a lock on the link's `shared_ptr` while it's processing data with that link. It continues to check, at the end of the message processing loop, to see if the link might have been orphaned suggesting a disconnect.
* `_multiLinkSingleVehicleTest` checks link1 and link2 to see who won the race for the primary link

I ran this test in a loop for about 12 hours last night and it was successful for every iteration.
`while xvfb-run --auto-servernum ./qgroundcontrol-start.sh --unittest:VehicleLinkManagerTest --logging:MAVLinkProtocolLog,VehicleLog,MockLinkLog,InitialConnectStateMachineLog,LinkInterfaceLog,"$(date -Is)" > log.txt; do sleep 1; done`

### DIFF
--- a/src/comm/LinkInterface.cc
+++ b/src/comm/LinkInterface.cc
@@ -13,39 +13,8 @@
 
 QGC_LOGGING_CATEGORY(LinkInterfaceLog, "LinkInterfaceLog")
 
-bool LinkInterface::Channel::reserve(LinkManager& mgr)
-{
-    if (_set) {
-        qCWarning(LinkInterfaceLog) << "Channel set multiple times";
-    }
-    uint8_t channel = mgr.reserveMavlinkChannel();
-    if (0 == channel) {
-        qCWarning(LinkInterfaceLog) << "Channel reserve failed";
-        return false;
-    }
-    qCDebug(LinkInterfaceLog) << "Channel::reserve" << channel;
-    _set = true;
-    _id = channel;
-    return true;
-}
-
-void LinkInterface::Channel::free(LinkManager& mgr)
-{
-    qCDebug(LinkInterfaceLog) << "Channel::free" << _id;
-    if (0 != _id) {
-        mgr.freeMavlinkChannel(_id);
-    }
-    _set = false;
-    _id = 0;
-}
-
-uint8_t LinkInterface::Channel::id(void) const
-{
-    if (!_set) {
-        qCWarning(LinkInterfaceLog) << "Call to LinkInterface::mavlinkChannel with _mavlinkChannel.set() == false";
-    }
-    return _id;
-}
+// The LinkManager is only forward declared in the header, so the static_assert is here instead.
+static_assert(LinkManager::invalidMavlinkChannel() == std::numeric_limits<uint8_t>::max(), "update LinkInterface::_mavlinkChannel");
 
 LinkInterface::LinkInterface(SharedLinkConfigurationPtr& config, bool isPX4Flow)
     : QThread   (0)
@@ -67,17 +36,46 @@ LinkInterface::~LinkInterface()
 
 uint8_t LinkInterface::mavlinkChannel(void) const
 {
-    return _mavlinkChannel.id();
+    if (!mavlinkChannelIsSet()) {
+        qCWarning(LinkInterfaceLog) << "Call to MAVLinkChannel::id with isSet() == false";
+    }
+    return _mavlinkChannel;
 }
 
-bool LinkInterface::_reserveMavlinkChannel(LinkManager& mgr)
+bool LinkInterface::mavlinkChannelIsSet(void) const
 {
-    return _mavlinkChannel.reserve(mgr);
+    return (LinkManager::invalidMavlinkChannel() != _mavlinkChannel);
 }
 
-void LinkInterface::_freeMavlinkChannel(LinkManager& mgr)
+bool LinkInterface::_allocateMavlinkChannel()
 {
-    return _mavlinkChannel.free(mgr);
+    // should only be called by the LinkManager during setup
+    Q_ASSERT(!mavlinkChannelIsSet());
+    if (mavlinkChannelIsSet()) {
+        qCWarning(LinkInterfaceLog) << "_allocateMavlinkChannel already have " << _mavlinkChannel;
+        return true;
+    }
+
+    auto mgr = qgcApp()->toolbox()->linkManager();
+    _mavlinkChannel = mgr->allocateMavlinkChannel();
+    if (!mavlinkChannelIsSet()) {
+        qCWarning(LinkInterfaceLog) << "_allocateMavlinkChannel failed";
+        return false;
+    }
+    qCDebug(LinkInterfaceLog) << "_allocateMavlinkChannel" << _mavlinkChannel;
+    return true;
+}
+
+void LinkInterface::_freeMavlinkChannel()
+{
+    qCDebug(LinkInterfaceLog) << "_freeMavlinkChannel" << _mavlinkChannel;
+    if (LinkManager::invalidMavlinkChannel() == _mavlinkChannel) {
+        return;
+    }
+
+    auto mgr = qgcApp()->toolbox()->linkManager();
+    mgr->freeMavlinkChannel(_mavlinkChannel);
+    _mavlinkChannel = LinkManager::invalidMavlinkChannel();
 }
 
 void LinkInterface::writeBytesThreadSafe(const char *bytes, int length)
--- a/src/comm/LinkInterface.cc
+++ b/src/comm/LinkInterface.cc
@@ -51,7 +51,7 @@ void LinkInterface::_setMavlinkChannel(uint8_t channel)
     if (_mavlinkChannelSet) {
         qWarning() << "Mavlink channel set multiple times";
     }
-    qCDebug(LinkInterfaceLog) << QThread::currentThread() << this << "setMavlinkChannel" << channel;
+    qCDebug(LinkInterfaceLog) << "setMavlinkChannel" << channel;
     _mavlinkChannelSet = true;
     _mavlinkChannel = channel;
 }
@@ -61,7 +61,7 @@ void LinkInterface::_setMavlinkAuxChannel(uint8_t channel)
     if (_mavlinkAuxChannelSet) {
         qWarning() << "Mavlink channel set multiple times";
     }
-    qCDebug(LinkInterfaceLog) << QThread::currentThread() << this << "setMavlinkChannel" << channel;
+    qCDebug(LinkInterfaceLog) << "setMavlinkChannel" << channel;
     _mavlinkAuxChannelSet = true;
     _mavlinkAuxChannel = channel;
 }
--- a/src/comm/LinkInterface.cc
+++ b/src/comm/LinkInterface.cc
@@ -37,7 +37,7 @@ LinkInterface::~LinkInterface()
 uint8_t LinkInterface::mavlinkChannel(void) const
 {
     if (!mavlinkChannelIsSet()) {
-        qCWarning(LinkInterfaceLog) << "Call to MAVLinkChannel::id with isSet() == false";
+        qCWarning(LinkInterfaceLog) << "mavlinkChannel isSet() == false";
     }
     return _mavlinkChannel;
 }
--- a/src/comm/LinkInterface.cc
+++ b/src/comm/LinkInterface.cc
@@ -38,6 +38,14 @@ uint8_t LinkInterface::mavlinkChannel(void) const
     return _mavlinkChannel;
 }
 
+uint8_t LinkInterface::mavlinkAuxChannel(void) const
+{
+    if (!_mavlinkAuxChannelSet) {
+        qWarning(LinkInterfaceLog) << "Call to LinkInterface::mavlinkAuxChannel with _mavlinkAuxChannelSet == false";
+    }
+    return _mavlinkAuxChannel;
+}
+
 void LinkInterface::_setMavlinkChannel(uint8_t channel)
 {
     if (_mavlinkChannelSet) {
@@ -46,6 +54,16 @@ void LinkInterface::_setMavlinkChannel(uint8_t channel)
     qCDebug(LinkInterfaceLog) << QThread::currentThread() << this << "setMavlinkChannel" << channel;
     _mavlinkChannelSet = true;
     _mavlinkChannel = channel;
+}
+
+void LinkInterface::_setMavlinkAuxChannel(uint8_t channel)
+{
+    if (_mavlinkAuxChannelSet) {
+        qWarning() << "Mavlink channel set multiple times";
+    }
+    qCDebug(LinkInterfaceLog) << QThread::currentThread() << this << "setMavlinkChannel" << channel;
+    _mavlinkAuxChannelSet = true;
+    _mavlinkAuxChannel = channel;
 }
 
 void LinkInterface::writeBytesThreadSafe(const char *bytes, int length)
--- a/src/comm/LinkInterface.cc
+++ b/src/comm/LinkInterface.cc
@@ -10,6 +10,8 @@
 #include "LinkInterface.h"
 #include "QGCApplication.h"
 
+QGC_LOGGING_CATEGORY(LinkInterfaceLog, "LinkInterfaceLog")
+
 LinkInterface::LinkInterface(SharedLinkConfigurationPtr& config, bool isPX4Flow)
     : QThread   (0)
     , _config   (config)
@@ -23,7 +25,7 @@ LinkInterface::LinkInterface(SharedLinkConfigurationPtr& config, bool isPX4Flow)
 LinkInterface::~LinkInterface()
 {
     if (_vehicleReferenceCount != 0) {
-        qWarning() << "~LinkInterface still have vehicle references:" << _vehicleReferenceCount;
+        qWarning(LinkInterfaceLog) << "~LinkInterface still have vehicle references:" << _vehicleReferenceCount;
     }
     _config.reset();
 }
@@ -31,7 +33,7 @@ LinkInterface::~LinkInterface()
 uint8_t LinkInterface::mavlinkChannel(void) const
 {
     if (!_mavlinkChannelSet) {
-        qWarning() << "Call to LinkInterface::mavlinkChannel with _mavlinkChannelSet == false";
+        qWarning(LinkInterfaceLog) << "Call to LinkInterface::mavlinkChannel with _mavlinkChannelSet == false";
     }
     return _mavlinkChannel;
 }
@@ -41,6 +43,7 @@ void LinkInterface::_setMavlinkChannel(uint8_t channel)
     if (_mavlinkChannelSet) {
         qWarning() << "Mavlink channel set multiple times";
     }
+    qCDebug(LinkInterfaceLog) << QThread::currentThread() << this << "setMavlinkChannel" << channel;
     _mavlinkChannelSet = true;
     _mavlinkChannel = channel;
 }
@@ -66,7 +69,7 @@ void LinkInterface::removeVehicleReference(void)
             disconnect();
         }
     } else {
-        qWarning() << "LinkInterface::removeVehicleReference called with no vehicle references";
+        qWarning(LinkInterfaceLog) << "LinkInterface::removeVehicleReference called with no vehicle references";
     }
 }
 
--- a/src/comm/LinkInterface.cc
+++ b/src/comm/LinkInterface.cc
@@ -99,7 +99,7 @@ void LinkInterface::removeVehicleReference(void)
             disconnect();
         }
     } else {
-        qCWarning(LinkInterfaceLog) << "LinkInterface::removeVehicleReference called with no vehicle references";
+        qCWarning(LinkInterfaceLog) << "removeVehicleReference called with no vehicle references";
     }
 }
 
--- a/src/comm/LinkInterface.cc
+++ b/src/comm/LinkInterface.cc
@@ -8,9 +8,44 @@
  ****************************************************************************/
 
 #include "LinkInterface.h"
+#include "LinkManager.h"
 #include "QGCApplication.h"
 
 QGC_LOGGING_CATEGORY(LinkInterfaceLog, "LinkInterfaceLog")
+
+bool LinkInterface::Channel::reserve(LinkManager& mgr)
+{
+    if (_set) {
+        qCWarning(LinkInterfaceLog) << "Channel set multiple times";
+    }
+    uint8_t channel = mgr.reserveMavlinkChannel();
+    if (0 == channel) {
+        qCWarning(LinkInterfaceLog) << "Channel reserve failed";
+        return false;
+    }
+    qCDebug(LinkInterfaceLog) << "Channel::reserve" << channel;
+    _set = true;
+    _id = channel;
+    return true;
+}
+
+void LinkInterface::Channel::free(LinkManager& mgr)
+{
+    qCDebug(LinkInterfaceLog) << "Channel::free" << _id;
+    if (0 != _id) {
+        mgr.freeMavlinkChannel(_id);
+    }
+    _set = false;
+    _id = 0;
+}
+
+uint8_t LinkInterface::Channel::id(void) const
+{
+    if (!_set) {
+        qCWarning(LinkInterfaceLog) << "Call to LinkInterface::mavlinkChannel with _mavlinkChannel.set() == false";
+    }
+    return _id;
+}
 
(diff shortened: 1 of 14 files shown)


DonLakeFlyer:
@keith-bennett-airmap Awesome stuff. Let me stare at this a bit today and I'll merge.

src/comm/LinkInterface.cc
```diff
@@ -23,28 +25,47 @@ LinkInterface::LinkInterface(SharedLinkConfigurationPtr& config, bool isPX4Flow)
 LinkInterface::~LinkInterface()
 {
     if (_vehicleReferenceCount != 0) {
-        qWarning() << "~LinkInterface still have vehicle references:" << _vehicleReferenceCount;
+        qWarning(LinkInterfaceLog) << "~LinkInterface still have vehicle references:" << _vehicleReferenceCount;
```
DonLakeFlyer, line 28: Doesn't this need to be `qCWarning`? Or does this work as well?

DonLakeFlyer:
Does the usage of a second mavlink channel by MockLink need to bleed into the mainline code? Seems like it would be simpler for MockLink to just have access to reserve and free the extra channel as needed. And then LinkManager doesn't need to be overly complicated with this aux stuff. I think that's possible but not sure.

src/comm/LinkInterface.cc
inetknght: There [is a difference between `qC`-prefix (`qCDebug`, `qCWarning`, etc) and `q`-prefix (`qDebug`, `qWarning`, etc) logging functions](https://doc.qt.io/qt-5/qloggingcategory.html#checking-category-configuration). Both "work". The former is somewhat more efficient than the latter. I left the call as-is other than to ensure that it gets put into the correct logging channel. I don't anticipate this code to be performance sensitive but if it were performance sensitive then it should move to `qCWarning`.

keith-bennett-airmap: There [is a difference between `qC`-prefix (`qCDebug`, `qCWarning`, etc) and `q`-prefix (`qDebug`, `qWarning`, etc) logging functions](https://doc.qt.io/qt-5/qloggingcategory.html#checking-category-configuration). Both "work". The former is somewhat more efficient than the latter. I left the call as-is other than to ensure that it gets put into the correct logging channel. I don't anticipate this code to be performance sensitive but if it were performance sensitive then it should move to `qCWarning`.

DonLakeFlyer: All other places in the code which use categorized debug out use the `qC` variants. Mainly because I thought it was the only way to do it. That said I think it would be best to continue to follow that pattern for consistency.

DonLakeFlyer:
I think there is some ArduPilot code which does mavlink channel 0 futzing as well. Let me look.

DonLakeFlyer:
And then once it's all switched to reserving their own channels on the fly. The concept of a reserved channel 0 can go away.

keith-bennett-airmap:
> Does the usage of a second mavlink channel by MockLink need to bleed into the mainline code?

I'm not sure I understand the question. Are you asking whether the auxiliary mavlink channel should be available to UDP, TCP, serial, etc links?

They need a separate buffer for sending and receiving for sure. Any link which expects to use the mavlink buffer for a partial message (so definitely TCP and serial) will need an auxiliary channel. Consider a TCP socket who's received a partial message (because the TCP frame was split by a router or something) or a serial link who's receiving a message byte-by-byte. If the buffer is in the middle of processing that data and then QGC wants to send a message, such as a heartbeat or something, then the buffer will be clobbered. This can manifest in high packet loss but might also stall the link (eg, if it thinks it needs to receive an abnormally large message) or cause invalid message processing (I could probably craft a malicious CRC packet if I tried).

> Seems like it would be simpler for MockLink to just have access to reserve and free the extra channel as needed. And then LinkManager doesn't need to be overly complicated with this aux stuff. I think that's possible but not sure.

I think it would be best to move away from the preallocated global mavlink buffers and move toward having the links provide their own buffers. I did a little bit of spelunking around the mavlink library and it looked like there was partial support for that but the candy wrappers QGC uses just expect a channel ID to point to the preallocated global buffers.

DonLakeFlyer:
> * Many racey uses of [`if (weak_ptr::expired()) {... weak_ptr::lock() ... }`](https://en.cppreference.com/w/cpp/memory/weak_ptr/expired)

Yup, that's majorly screwed up.

DonLakeFlyer:
> I'm not sure I understand the question

What I'm saying is that in you instead:
* Make LinkManager::reserveMavlinkChannel/freeMavlinkChannel public methods
* Change MockLink/APMFirmwarePlugin to just call these to get the extra channel they need to work with
* No additional changes needed to LinkManager

To me this is much simpler and doesn't complicate LinkManager with this aux channel stuff.

keith-bennett-airmap:
I see. Yes that can work too.

mrpollo:
@patrickelectric you might have seen this happen to some of your PRs as well!

DonLakeFlyer:
It is also likely worthwhile to have a debug check in APMFirmarePlugin to make sure it usage of the one mavlink channel is always on the main thread. That's is how it works now, but who knows in the future.

DonLakeFlyer:
Also once this is all worked out it will need to cherry pick to Stable as well.

keith-bennett-airmap:
@DonLakeFlyer 

I don't see automated tests for `APMFirmwarePlugin`. How do I exercise the changes to the `APMFirmwarePlugin`?

I added to `APMFirmwarePlugin` the reserve/free of a mavlink channel via the `LinkManager` anywhere channel 0 was being used. However I see what looks like possibly other problems with APMFirmwarePlugin. I'm not familiar with the class so perhaps you can help me to understand:

* `APMFirmwarePlugin::_handleOutgoingParamSetThreadSafe` uses a mutex when accessing the channel. That suggests that all of the  other uses of the channel should be mutexed as well. This is also the only function that uses the mutex. Without a way to exercise the firmware plugin code it's difficult for me to tell whether this is correct but it stands out like a sore thumb to my experience in threading and network comms.

* I extracted the channel int and bool into its own small class, `LinkInterface::Channel`. I think the bool determining whether a channel has been set can simply be replaced with a check for a non-zero channel id. What are your thoughts about that?

* Is there only ever one `LinkManager`? I think that should be correct anyway. Perhaps the `LinkInterface::Channel` can hide its dependency on it through `qgcApp()->toolbox()->linkManager()` in the `.cc` file and removing the pointer. That would make the added code to `APMFirmwarePlugin` cleaner for sure

DonLakeFlyer:
> How do I exercise the changes to the `APMFirmwarePlugin`?

For the ArduPilot parameter futzing stuff MockLink simulates that for ArduPilot vehicles. So you can use that to verify it's isn't completely hosed. Trying to simulate/test the thread safety part would be really tricky to try to hit a collision would be really trick and I don't see a good way to do that right now.

I still don't see the need for the LinkInterface code to have to know anything about the concept of an Aux channel. It just bleeds complexity into LinkInterface that it doesn't need to know about. All to support a single use case for MockLink. All that is needed is for the ability for outside code to reserve/free channels as needed to use them how they see fit.

Same goes for the `LinkInterface::Channel` class. I think it adds complexity and not much value. The current way `LinkManager::createConnectedLink` reserves the mavlink channel is fine and easily understood.

LinkManager is a QGCTool. QGCTools are singletons. QGCTools are always assumed to be available. So you never need to check whether your call to the toolbox returns null for a tool. If it did all hell would break lose immediately since it breaks a fundamental concept that tools are always available. You also should never pass a tool as a parameter to a method. If any code needs a tool it just asks for it from the toolbox as needed. 

What I think should happen:
* LinkManager should expose public reserve/free channel methods. Max uint8 can be used as the signal for not set or failure of reserve call. LinkManager can expose a constant for this value.
* `LinkManager::reservedMavlinkChannel` should be removed and channel 0 should be able to be returned as usable by the reverse mthod
* Mocklink: The additional reserved mavlink channel is used for outgoing mavlink traffic Vehicle->GCS direction. MockLink should just call LinkManager reserve/free channel in constructor/destructor to get the mavlink channel it needs.
* APMFirmwarePlugin: There is a single instance of APMFirmwarePlugin per vehicle class that has been connected to. So for example if you are connected to multiple copters there is a single APMFirmwarePlugin instance. But if you are connected a copter and a plane there will be two instances. The need for a mavlink channel here (the channel 0 usage) is just to adjust incoming mavlink messages on the fly on the way in/out. It does not maintain any persistent status associated with the channel across usage. It's more of a hack usage than anything else for re-encoding, but given how mavlink works I don't see any other way to do it. Given all of that there really only needs to be a single channel reserved to work with. So this could be a static in the class. And the constructor can reserve it if it already hasn't done so. Since there is only one of these usage to it needs to be protected with a mutex. 

> * APMFirmwarePlugin::_handleOutgoingParamSetThreadSafe

You are correct that this smells bad. This is a bit of a different case than all of the above stuff since it is working with the outgoing links mavlink channel. I did a little looking and it seems correct. But certainly could use more eyeballs.

DonLakeFlyer:
FYI: ArduPilot SITL is also a good way to sanity check. But requires a lot of setup if you've never done before. If not, I have a setup for it and can do it.

keith-bennett-airmap:
> > How do I exercise the changes to the `APMFirmwarePlugin`?
> 
> For the ArduPilot parameter futzing stuff MockLink simulates that for ArduPilot vehicles. So you can use that to verify it's isn't completely hosed.

I run the unit tests before pushing so that should suffice then.

> Trying to simulate/test the thread safety part would be really tricky to try to hit a collision would be really trick and I don't see a good way to do that right now.

There are [tools](https://stackoverflow.com/a/39041217/1111557) to detect this kind of problem. I'm using Thread Sanitizer and Address Sanitizer. TSan is particularly noisy in this codebase but it helped a lot in tracking down the flaky race condition.

I *really* want to add them to CI but Qt's custom mutex implementation works against us here. [Google (author of TSan)](https://github.com/google/sanitizers/issues/460) and [Qt](https://bugreports.qt.io/browse/QTBUG-79770) seem to disagree about who owns the problem. Alas, that's a rabbit hole.

> I still don't see the need for the LinkInterface code to have to know anything about the concept of an Aux channel. It just bleeds complexity into LinkInterface that it doesn't need to know about. All to support a single use case for MockLink. All that is needed is for the ability for outside code to reserve/free channels as needed to use them how they see fit.

I removed the Aux channel from the `LinkInterface` in previous commits.

> Same goes for the `LinkInterface::Channel` class. I think it adds complexity and not much value. The current way `LinkManager::createConnectedLink` reserves the mavlink channel is fine and easily understood.

I disagree, but not strongly. So I'll remove the `LinkInterface::Channel`.

Reserving and freeing a channel smells a lot like what RAII solves and that's a core feature of C++.

> LinkManager is a QGCTool. QGCTools are singletons. QGCTools are always assumed to be available. So you never need to check whether your call to the toolbox returns null for a tool. If it did all hell would break lose immediately since it breaks a fundamental concept that tools are always available. You also should never pass a tool as a parameter to a method. If any code needs a tool it just asks for it from the toolbox as needed.

Good to know. Thank you!

> What I think should happen:
> 
>     * LinkManager should expose public reserve/free channel methods. Max uint8 can be used as the signal for not set or failure of reserve call. LinkManager can expose a constant for this value.

Previous commits already made the reserve/free channel methods public.

I'm not sure about using max uint8 if this needs to be merged into a stable release though: that smells like an interface contract change. Do any downstream forks use the reserve/free mechanism? They would undoubtedly check for a 0 since that's how it's done now. Changing it to max uint8 instead would be a silent change that won't cause compiler errors but _will_ change the interface contract.

I think that's also somewhat of an argument for wrapping this into its own class.

>     * `LinkManager::reservedMavlinkChannel` should be removed and channel 0 should be able to be returned as usable by the reverse mthod

I agree channel 0 should be able to be allocated. I still worry about anywhere that has 0 as a hardcoded magic number but I think this is progress.

It's still a silent interface contract change and I don't think that should get cherry picked into stable; now not only will 0 _not_ be a reserved-but-unallocated channel, it will be a possible answer to reserve request. Any existing code that checks for 0 will erroneusly think that the `LinkManager` ran out of channels and there won't be a compiler error to point out that the contract has changed.

I think a good way to ensure that any dependent code gets updated would be to change the function name: `allocateMavlinkChannel` instead of `reserveMavlinkChannel`. I think that will also help to disambiguate `reserve()` in the action sense and `reserved` in the _do not use_ sense.

>     * Mocklink: The additional reserved mavlink channel is used for outgoing mavlink traffic Vehicle->GCS direction. MockLink should just call LinkManager reserve/free channel in constructor/destructor to get the mavlink channel it needs.

Yes.

>     * APMFirmwarePlugin: There is a single instance of APMFirmwarePlugin per vehicle class that has been connected to. So for example if you are connected to multiple copters there is a single APMFirmwarePlugin instance. But if you are connected a copter and a plane there will be two instances. The need for a mavlink channel here (the channel 0 usage) is just to adjust incoming mavlink messages on the fly on the way in/out. It does not maintain any persistent status associated with the channel across usage. It's more of a hack usage than anything else for re-encoding, but given how mavlink works I don't see any other way to do it. Given all of that there really only needs to be a single channel reserved to work with. So this could be a static in the class. And the constructor can reserve it if it already hasn't done so. Since there is only one of these usage to it needs to be protected with a mutex.

That's info I didn't have. Thank you and agree.

> 
> 
> > * APMFirmwarePlugin::_handleOutgoingParamSetThreadSafe
> 
> You are correct that this smells bad. This is a bit of a different case than all of the above stuff since it is working with the outgoing links mavlink channel. I did a little looking and it seems correct. But certainly could use more eyeballs.

:+1:
Honestly I need to learn a lot more about the mavlink protocol and, in particular, the library. My background has a lot of experience in network protocols and C++ and it smells like there's room to improve. But, one thing at a time.

> FYI: ArduPilot SITL is also a good way to sanity check. But requires a lot of setup if you've never done before. If not, I have a setup for it and can do it.

Cool. I watched [the video](https://www.youtube.com/watch?v=pJGFkZmGV6o) from the [setup page](https://ardupilot.org/dev/docs/building-setup-linux.html#building-setup-linux). It seems fairly straightforward use of a VM. I'll look at getting that set up today.

DonLakeFlyer:
> I'm not sure about using max uint8 if this needs to be merged into a stable release though: that smells like an interface contract change. Do any downstream forks use the reserve/free mechanism? They would undoubtedly check for a 0 since that's how it's done now. Changing it to max uint8 instead would be a silent change that won't cause compiler errors but _will_ change the interface contract.

The method name for reserve has changed though right? So anyone calling it from custom build will fail. Still given that I think your suggestion to move away from reserved->allocate makes the contract change even more clear. Also with detailed comments on the method about 0 no longer being a special case. The incoming message futzing stuff was added to handle strangeness in ArduPilot with respect to not quite following mavlink spec. It was added using a generic mechanism but I'd be surprised if there is anyone else out there using to massage mavlink traffic so something else on the way in. Also in the current code the concept of channel 0 is not a flag to signal not set. It was simply a hack to keep a channel aside for out of band use.

There are also only two mavlink methods which can get screwed up with the channel 0 thing: *_pack_chan and *_encode_chan. I reviewed usage of all of those after your code changes and found no remaining usage of channel 0 hack. I'd like to move this across to stable, so sorry to ask for one more change but I think your "allocate" name change is warranted to signal even more specifically the whole channel 0 change thing. And then I think it's ok to move to Stable.

> Reserving and freeing a channel smells a lot like what RAII solves and that's a core feature of C++.

I get that. But my opinion is that when the usage pattern is simple in that there is a single point of initialization and a single point of cleanup and the init/cleanup is also dead simple you are better off not hiding it behind a layer in order to improve the readability of what is really happening. Maybe I'm just too much of and old geezer to follow these patterns everywhere and I need to change my opinions. My experience move me towards leaning very hard on readability as being overly important.

DonLakeFlyer:
@keith-bennett-airmap Also thanks so much for all of your work here. This is all really fantastic stuff.

keith-bennett-airmap:
> I think your "allocate" name change is warranted to signal even more specifically the whole channel 0 change thing.

That made it into this PR via this commit [c826618e64bac61133f32bb813d092116dd11e21](https://github.com/mavlink/qgroundcontrol/pull/9595/commits/c826618e64bac61133f32bb813d092116dd11e21#diff-bf8226457392bb38828f1531a71ab07369c8974d64c8b64ca100ccf48c04ac41L114-L117); search for `LinkManager.h` in the Github diff browser window will bring you to it

> Also thanks so much for all of your work here. This is all really fantastic stuff.

I'm glad to be able to help!

keith-bennett-airmap:
> But my opinion is that when the usage pattern is simple in that there is a single point of initialization and a single point of cleanup and the init/cleanup is also dead simple you are better off not hiding it behind a layer in order to improve the readability of what is really happening.

In C that makes sense. In C++ there's sometimes a `throw` that can make cleanup a bit of a mess.  In this PR the changes are around calls to C library so there isn't risk to of a throw. So it's not a problem to me.  In my experience I've found that creating a RAII class, even for small types, helps with that since now the compiler is in charge of ensuring correct cleanup no matter how the code mutates over the years.

I think the concept is strongly embraced by the C++ community; for example we now have smart pointers and smart locks. It also helps to reduce code duplication: cleanup code is now only in the object destructor instead of duplicated in 30 places. And the objects can help a lot when interfacing with C libraries that have their own allocation/cleanup routines.

My rule of thumb basically comes to: write it once is great. Writing it twice is allowable copy-pasting. Writing it a third time is copy-paste for the sake of time or a class for future-proofing. Writing it a fourth time necessitates its own class. In this case, I'd written the channel changes in LinkInterface and MockLink. When I learned the channel changes were needed in the APMFirmwarePlugin too, that's the third time to me.

DonLakeFlyer:
> That made it into this PR via this commit

Great, this is all done then. Can you bring to Stable as well.

PR #9606, opened 2021-04-21
Fix state machine not being advanced in all cases

DonLakeFlyer:
The initial connect state machine could get stuck prior to downloading parameters.


DonLakeFlyer:
The Vehicle::waitForMavlinkMessage implementation has timeout support. If either there is no ack or if you never get the message the system falls back to mavlink 1. The code is written such that it's ok to not determine protocol version. It's doesn't retry in this case. Are you saying there is a bug in that flow. Or are saying the way it is currently coded is incorrect and protocol version determination should be retried.

DonLakeFlyer:
Ok, got it. There is a bug in Vehicle::waitForMessage then. It doesn't use requestMessage because that support is newer and the old protocol version stuff used request_protocol_version which I moved over to this new code. I'll fix waitForMessage and then probably move to requestMessage instead since it's a simpler flow.

Issue #9563, opened 2021-03-26
Make it easier/clearer to exit video fullscreen mode


jaxxzer:
When streaming video, if you double click the video surface, the video will grow to fullscreen mode. User reports it was unclear how to exit this mode (you have to double click the surface again).

User suggested something people would naturally try, like the escape key.

DonLakeFlyer:
I think something like a simple message that says double-click to exit full screen or something which shows up if you have the mouse over the video might work nicely.

Ljung323:
@jaxxzer @DonLakeFlyer 
I've just tried to implement this feature in #9612.
Could you have a check whether it's intended behavior?

holden-zenith:
This should be closed

PR #9612, opened 2021-04-22
Show message to exit video fullscreen mode

Ljung323:
Show how to exit video fullscreen mode when mouse hovers on the video view (https://github.com/mavlink/qgroundcontrol/issues/9563#issuecomment-808545622)

https://user-images.githubusercontent.com/82586164/115717387-1912c280-a3b5-11eb-97cc-c534cf12b950.mov

### DIFF
--- a/src/FlightDisplay/FlyViewVideo.qml
+++ b/src/FlightDisplay/FlyViewVideo.qml
@@ -67,18 +67,16 @@ Item {
     QGCLabel {
         text: qsTr("Double-click to exit full screen")
         font.pointSize: ScreenTools.largeFontPointSize
-        visible: QGroundControl.videoManager.fullScreen && flyViewVideoMouseArea.isHovered
+        visible: QGroundControl.videoManager.fullScreen && flyViewVideoMouseArea.containsMouse
         anchors.centerIn: parent
     }
 
     MouseArea {
         id: flyViewVideoMouseArea
-        property bool isHovered: false
         anchors.fill:       parent
         enabled:            pipState.state === pipState.fullState
         hoverEnabled: true
         onDoubleClicked:    QGroundControl.videoManager.fullScreen = !QGroundControl.videoManager.fullScreen
-        onHoveredChanged: isHovered = !isHovered
     }
 
     ProximityRadarVideoView{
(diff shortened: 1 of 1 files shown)


src/FlightDisplay/FlyViewVideo.qml
```diff
@@ -63,10 +64,21 @@ Item {
...
+    QGCLabel {
+        text: qsTr("Double-click to exit full screen")
+        font.pointSize: ScreenTools.largeFontPointSize
+        visible: QGroundControl.videoManager.fullScreen && flyViewVideoMouseArea.isHovered
+        anchors.centerIn: parent
+    }
+
     MouseArea {
+        id: flyViewVideoMouseArea
+        property bool isHovered: false
```
DonLakeFlyer, line 76: Why is this property needed? Why can't `flyViewVideoMouseArea.hovered` be used directly?

Ljung323: @DonLakeFlyer 
I tried to use the pre-defined `hovered`, but it is not defined for MouseArea. 
([Button](https://doc.qt.io/qt-5/qml-qtquick-controls2-control.html#hovered-prop) has `hovered` property, but it seems [MouseArea](https://doc.qt.io/qt-5/qml-qtquick-mousearea-members.html) does not have it)

Ljung323: Also, renaming `isHovered` to `hovered` does not work. (other names than `hovered` works)
I don't figure out the reason, but when `property bool hovered` is defined, `onHoveredChanged` is not triggered.

DonLakeFlyer: > I tried to use the pre-defined `hovered`, but it is not defined for MouseArea.

That's certainly odd. Looks like it is 'containsMouse` from MouseArea which tells you what you need. You should be able to use that as the reference for showing/hiding the help test.

Ljung323: @DonLakeFlyer 
Thank you for the feedback, `containsMouse` worked perfectly. a9c0d52
On the first trial for `containsMouse`, I did not enable `hoverEnabled` which triggers `containsMouse`.

https://user-images.githubusercontent.com/82586164/119072360-4d19eb80-ba26-11eb-92a3-bba7173afee2.mov

DonLakeFlyer: Great. Much simpler now. Thanks.

PR #9614, opened 2021-04-23
Replace deprecated functions

booo:
Replaces deprecated functions with the not deprecated counterparts.

Without these fixes I can't build QGC on archlinux with `qmake ... && make`.

### DIFF
--- a/src/Airmap/QJsonWebToken/src/qjsonwebtoken.cpp
+++ b/src/Airmap/QJsonWebToken/src/qjsonwebtoken.cpp
@@ -186,9 +186,9 @@ bool QJsonWebToken::setAlgorithmStr(QString strAlgorithm)
 	// set algorithm
 	m_strAlgorithm = strAlgorithm;
 	// modify header
-	m_jdocHeader = QJsonDocument::fromJson(QObject::trUtf8("{\"typ\": \"JWT\", \"alg\" : \"").toUtf8()
+	m_jdocHeader = QJsonDocument::fromJson(QObject::tr("{\"typ\": \"JWT\", \"alg\" : \"").toUtf8()
 		                                 + m_strAlgorithm.toUtf8()
-		                                 + QObject::trUtf8("\"}").toUtf8());
+		                                 + QObject::tr("\"}").toUtf8());
 
 	return true;
 }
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
These are useful to move forward to a newer Qt version which is something we are working towards. 

But be careful it you aren't using the prescribed version of Qt with QGC. You are running a version of QGC which has no eyeballs against finding bugs. It's rare that moving to a new version of Qt has not caused any subtle bugs in QGC.

booo:
Ok. Thanks for the feedback.

I checked my commits again and think the QElapsedTimer class should be used instead of the QTime class to check for the elapsed time. I can rework the commits if there is an interest in merging them later on.

patrickelectric:
I would recommend to move to `QElapsedTimer`.

DonLakeFlyer:
> I can rework the commits if there is an interest in merging them later on.

Definitely interested in the updates since we will be moving to new Qt soon.

src/Airmap/QJsonWebToken/src/qjsonwebtoken.cpp
```diff
@@ -186,9 +186,9 @@ bool QJsonWebToken::setAlgorithmStr(QString strAlgorithm)
 	// set algorithm
 	m_strAlgorithm = strAlgorithm;
 	// modify header
-	m_jdocHeader = QJsonDocument::fromJson(QObject::trUtf8("{\"typ\": \"JWT\", \"alg\" : \"").toUtf8()
+	m_jdocHeader = QJsonDocument::fromJson(QObject::tr("{\"typ\": \"JWT\", \"alg\" : \"").toUtf8()
```
DonLakeFlyer, line 189: Is it ok to switch away from trUtf8 because there is no non-ascii in the strings?

booo: How would this look like?

I think using utf8 strings is fine. Calling tr is totally unnecessary in this context I guess.

DonLakeFlyer: > Calling tr is totally unnecessary in this context I guess.

I believe that is correct. Can you just remove the 'tr' then?

src/Airmap/QJsonWebToken/src/qjsonwebtoken.cpp
```diff
@@ -186,9 +186,9 @@ bool QJsonWebToken::setAlgorithmStr(QString strAlgorithm)
...
 		                                 + m_strAlgorithm.toUtf8()
-		                                 + QObject::trUtf8("\"}").toUtf8());
+		                                 + QObject::tr("\"}").toUtf8());
```
patrickelectric, line 191: Why this is being translated ?

booo: I raised the same question in my commit message ;) I think it's totally unnecessary.

I'm a bit tired fighting with qgc and all the changes that brake my local build when I do a rebase every other day. Feel free to adopt the pull request.

DonLakeFlyer:
> I would recommend to move to `QElapsedTimer`.

Closing since this is the way to go

PR #9630, opened 2021-05-04
PX4: allow flight mode setup without sensors calibrated

julianoes:
Before this change it is not allowed to set up the flight modes unless all sensor calibrations are completed.

I don't think that makes sense because it is possible to calibrate RC without having the sensors calibrated, and it's the logical next step right after that.

It is also a problem if HITL (hardware-in-the-loop) is used where the sensors are not marked as calibrated but you still need to be able to configure the flight modes. This is what prompted me to suggest this change.

### DIFF
--- a/src/AutoPilotPlugins/PX4/PX4AutoPilotPlugin.cc
+++ b/src/AutoPilotPlugins/PX4/PX4AutoPilotPlugin.cc
@@ -156,8 +156,6 @@ QString PX4AutoPilotPlugin::prerequisiteSetup(VehicleComponent* component) const
                 return _airframeComponent->name();
             } else if (_radioComponent && !_radioComponent->setupComplete()) {
                 return _radioComponent->name();
-            } else if (_sensorsComponent && !_sensorsComponent->setupComplete()) {
-                return _sensorsComponent->name();
             }
         }
     } else if (qobject_cast<const PX4RadioComponent*>(component)) {


LorenzMeier (approved):

DonLakeFlyer:
This is because HIL support in QGC was removed about a year ago. I though HIL support was dead? Previous to this change HIL mode was checked for flight mode setup.

julianoes:
HIL as such still works, just not involving QGC.

DonLakeFlyer:
Ok, so I think the solution is to bring back the "HIL Mode" checking changes and leave out all the x-Plane insanity stuff. If you look at git blame for this line you can see the commit where it was removed. If someone could help bring back only the "HIL Mode" relevant parts then this would all work better.

LorenzMeier:
I think this change is harmless overall, because being more flexible on the setup UX doesn't hurt normal users either. I found it to be too rigid for real systems as well.

DonLakeFlyer:
> I think this change is harmless overall

I haven't had to time to go back and figure out if that is true or not. That check was there for some reason otherwise it wouldn't have been added. I just need to do some digging as to whether that reason is crap or not which is entirely possible.

DonLakeFlyer:
I believe there were multiple places in QGC code which checked for HIL mode. This change likely only fixes one of them. Looking at the original commit removing it will show the others.

julianoes:
But this doesn't check for HITL mode. It checks if sensors are ok. But to select a flight mode there is no good reason that the sensors need to be ok. Even the RC calibration is allowed.

DonLakeFlyer:
Look at this pull: https://github.com/mavlink/qgroundcontrol/pull/8617

DonLakeFlyer:
If HIL mode is still a thing with respect to using QGC then the "HIL Mode" changes should be brought back and the x-plane stuff continues to be left out.

DonLakeFlyer:
> But to select a flight mode there is no good reason that the sensors need to be ok.

I looked through both the simple and advanced flight mode setup code and don't see a connection to sensor cal requirement. So theoretically on visual code inspection this is ok. But theoretically doesn't always pan out.

@julianoes Have you verified that both simple and advanced flight mode setup don't puke on a virgin board setup (for some obscure reason) without doing sensor cal first?

PR #9633, opened 2021-05-07
github: workflows: Do not release mac binaries

patrickelectric:
Avoid any kind of release for now, only upload artifacts for action test


DonLakeFlyer:
#9631 covers the cases from your two workflows now. So this is no longer needed.

PR #9635, opened 2021-05-09
Font - fix return of resource name by font family name

tzuriel-spearuav:


dogmaphobic:
I only vaguely remember when the issue with fonts running on Korean systems came up. Whoever brought the issue up at the time also told us what font (names) to use. You would need a Korean installation to know what the current, proper names are. I would leave it as is (the same for both) until someone either from SK or who knows what to do chimes in.

dogmaphobic:
@kdnguru would know:

https://github.com/mavlink/qgroundcontrol/pull/7463#issuecomment-494265830

DonLakeFlyer:
@kdnguru Can you look at this pull with respect to getting the Koren bold font working correctly with this change?

patrickelectric (approved):
Looks sane

DonLakeFlyer:
@tzuriel-spearuav Thanks for the fixes on this.

PR #9414, opened 2021-02-06
Update mavlink v2.0 submodule Sat Feb  6 06:55:09 UTC 2021

dagar:

PR #9638, opened 2021-05-11
Correct fix for PROTOCOL_VERSION message lost

DonLakeFlyer:
This is the correct set of fixes for what was attempted/discussed in #9414. 

@bkueng FYI. Your initial comment was correct. I simulated lost message to test. Thanks for looking.


DonLakeFlyer:
This still isn't fully correct. There are more bugs here with respect to the Vehicle::requestMessage implementation.

PR #9640, opened 2021-05-12
fix cmake issues on windows

kistlin:
ignore typical IDE folders (CLion, VSCode)
exclude none windows compiler flags, make zlib include/linker files available
- cl : Command line error D8021 : invalid numeric argument '/Wno-missing-braces'
- ..\src\VehicleSetup\FirmwareUpgradeController.cc(29): fatal error C1083: Cannot open include file: 'zlib.h': No such file or directory

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+.idea/
+.vscode/
+cmake-build-*/
+out/
 libs/lib/Frameworks/GStreamer.framework
 *_qmlcache.qrc
 *.swp
--- a/libs/CMakeLists.txt
+++ b/libs/CMakeLists.txt
@@ -1,5 +1,8 @@
 add_subdirectory(qtandroidserialport)
 add_subdirectory(shapelib)
+if (WIN32)
+    add_subdirectory(zlib)
+endif (WIN32)
 if (GST_FOUND)
     add_subdirectory(qmlglsink)
 endif()
--- a/libs/shapelib/CMakeLists.txt
+++ b/libs/shapelib/CMakeLists.txt
@@ -33,11 +33,13 @@ option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
 
 cmake_policy(SET CMP0026 OLD) # Disallow use of the LOCATION property for build targets
 
-add_compile_options(
-	-Wno-missing-braces
-	-Wno-sign-compare
-	-Wno-return-type
-)
+if(NOT MSVC)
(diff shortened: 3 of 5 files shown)


patrickelectric:
Sadly, the cmake github action was removed, is there a reason for this @DonLakeFlyer ?
It was first added to avoid breaks on the cmake build system and also to help us review cmake pull requests.

thopiekar:
@patrickelectric Whut?! Good that I didn't spent too much time on this :sweat_smile:

DonLakeFlyer:
> It was first added to avoid breaks on the cmake build system and also to help us review cmake pull requests.

Ah, sorry. Didn't realize that was for cmake verification. It thought it was for travis replacement which I've started working on. Let me finish the S3 part this weekend and then you can add a cmake job back in to the workflow for verification.

DonLakeFlyer:
In reality I'm not super interested in having cmake in the CI system. To me the concept of having to update/verify two different build systems to submit a change is insanity. We can discuss at next weeks dev meeting.

kistlin:
Managing two build systems makes no sense I agree. What will be the procedure now?
Because I have some more changes in CMake files for Windows. They fix gstreamer related issues and contain some cleanups.

Should I open the pull-requests and we just leave them? So people find it if one wants to use CMake on Windows?
Or no more pull-requests for now?

And is a move to CMake somewhere on the Roadmap?
Maybe together with a move to Qt6 (if planned)? As Qt6 itself uses CMake to build and the documentation got updated with CMake commands.

For example QFile documentation (https://doc.qt.io/qt-6/qfile.html)
```
find_package(Qt6 COMPONENTS Core REQUIRED)
target_link_libraries(mytarget PRIVATE Qt6::Core)
```

Even though qmake is obviously still supported.

DonLakeFlyer:
> Should I open the pull-requests and we just leave them? So people find it if one wants to use CMake on Windows?
> Or no more pull-requests for now?

The procedure up to now has been that folks like @dagar or @patrickelectric have been reviewing the cmake pulls and merging them in.

> Maybe together with a move to Qt6 (if planned)? As Qt6 itself uses CMake to build and the documentation got updated with CMake commands.

That is most likely the timeframe where we move fully since I assume that will mean the cmake will have full end to end support in Qt. The plan for next Qt version to move to has been 5.15 not 6. But at this point I'm not up to speed on whether a move directly to 6 is feasible.

PR #9641, opened 2021-05-12
Stable v4.1 - create an "autoconfig" function for RTSP video streaming (IP setting)

GlidLov:
It would be a good idea if there's an auto detect mechanism for the RTSP string (at least for IP address).
One option would be:
- take the ip from comm link (detect which one is the active connection - Q it's already doing it)
- then take that IP and put it into RTSP string. 

This would work when using a secondary tablet connected with herelink, so if it changes IP or you change from a wifi network to hotspot mode, you would not need to change your config.

Another method would be to make a list of RTSP urls, like it's in comm links' IP host list, so Q would scan them when opening. User could fill that list with some "preset" urls, so no need to change everytime you open app

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -43,7 +43,8 @@ install:
       Write-Host "Installed" -ForegroundColor Green
 
 build_script:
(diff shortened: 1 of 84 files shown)


DonLakeFlyer:
Not sure what this is. Looks like a bad pull given the commits.

There already is a system for autodetect of RTSP. It is specified in the new mavlink camera protocol.

DonLakeFlyer:
I'm assuming this is a mistaken pull. Closing

PR #9644, opened 2021-05-14
Hardening of inital connect sequence to boundary conditions

DonLakeFlyer:
The initial connect sequence had bugs in it with respect to lost responses to things like `AUTOPILOT_VERSION` or 'PROTOCOL_VERSION'. This could cause the connect sequence to hang and never complete. Made a pile of changes to fix this up as well as create MockLink support and units tests to validates the failure conditions. 

* MockLink - Support REQUEST_MESSAGE:AUTOPILOT_VERSION, REQUEST_MESSAGE:PROTOCOL_VERSION
* Fix bugs in Vehicle::requestMessage
* Switch to all INT mission items
* Create unit tests to exercise Initial Connect state mache state failure conditions
* Change initial connect sequence state machine to use REQUEST_MESSAGE for `AUTOPILOT_VERSION` or 'PROTOCOL_VERSION' states

### DIFF
--- a/src/Vehicle/InitialConnectStateMachine.cc
+++ b/src/Vehicle/InitialConnectStateMachine.cc
@@ -19,7 +19,7 @@
 QGC_LOGGING_CATEGORY(InitialConnectStateMachineLog, "InitialConnectStateMachineLog")
 
 const StateMachine::StateFn InitialConnectStateMachine::_rgStates[] = {
-    InitialConnectStateMachine::_stateRequestCapabilities,
+    InitialConnectStateMachine::_stateRequestAutopilotVersion,
     InitialConnectStateMachine::_stateRequestProtocolVersion,
     InitialConnectStateMachine::_stateRequestCompInfo,
     InitialConnectStateMachine::_stateRequestParameters,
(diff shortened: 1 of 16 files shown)


src/Vehicle/InitialConnectStateMachine.cc
```diff
@@ -198,6 +169,35 @@ void InitialConnectStateMachine::_waitForAutopilotVersionResultHandler(void* res
...
+    if (failureCode != Vehicle::RequestMessageNoFailure) {
+        qCDebug(InitialConnectStateMachineLog) << "REQUEST_MESSAGE:AUTOPILOT_VERSION failed. Setting no capabilities";
+        uint64_t assumedCapabilities = 0;
+        if (vehicle->_mavlinkProtocolRequestMaxProtoVersion >= 200) {
+            // Link already running mavlink 2
+            assumedCapabilities |= MAV_PROTOCOL_CAPABILITY_MAVLINK2;
+        }
+        if (vehicle->px4Firmware() || vehicle->apmFirmware()) {
+            // We make some assumptions for known firmware
+            assumedCapabilities |= MAV_PROTOCOL_CAPABILITY_MISSION_INT | MAV_PROTOCOL_CAPABILITY_COMMAND_INT | MAV_PROTOCOL_CAPABILITY_MISSION_FENCE | MAV_PROTOCOL_CAPABILITY_MISSION_RALLY;
```
DonLakeFlyer, line 196: @bkueng This specific change I wonder about. It make some assumptions about known firmwares if the AUTOPILOT_VERSION message is lost. The case where this could theoretically go wrong is when say the vehicle is running modified PX4 firmware which still reports as PX4 but the customization has specifically turned on some of these features. In that case this specific path would turn them back on. What do you think?

DonLakeFlyer:
This fixes the problems found by @bkueng talked about here: https://github.com/mavlink/qgroundcontrol/pull/9606. It ended up being somewhat involved to fix well. Hence the unit testing from hell to verify.

src/Vehicle/InitialConnectStateMachine.cc
bkueng: I think this is ok, as I think it's very rare that someone would turn one of those off. And with the retry logic it should be fairly robust, possibly the most likely failure case is a router in between that consistently drops the message.

bkueng (approved):
I didn't check in detail, but looks reasonable. Thanks @DonLakeFlyer

ManuelMeraz:
This pull request introduced a bug in the initialization sequence. After the the AUTOPILOT_VERSION is received by QGroundControl, it fails to go beyond the PROTOCOL_VERSON step in the state machine because of the following line:

https://github.com/DonLakeFlyer/qgroundcontrol/blob/846e7fa04a8f810726de27d670c3416f99f3b6b4/src/Vehicle/InitialConnectStateMachine.cc#L222
```
            qCDebug(InitialConnectStateMachineLog) << "Sending REQUEST_MESSAGE:PROTOCOL_VERSION";
            vehicle->requestMessage(_protocolVersionRequestMessageHandler,
                                    connectMachine,
                                    MAV_COMP_ID_AUTOPILOT1,
                                    MAVLINK_MSG_ID_AUTOPILOT_VERSION);
```
As you can see the vehicle is supposed to expect a PROTOCOL_VERSON_MESSAGE but instead `MAVLINK_MSG_ID_AUTOPILOT_VERSION` is being requested. `MAVLINK_MSG_ID_AUTOPILOT_VERSION`  needs to be changed to `MAVLINK_MSG_ID_PROTOCOL_VERSION` in line 222. I will open up a pull request at some point, but wanted to point out this bug that currently exists in master.

PR #9681, opened 2021-05-26
Fix build problems with docker

patrickelectric:
Fixes two problems:
  - Fix and use default user to avoid the creation/modification of files as root
  - Remove git commands since it breaks git internals when running inside docker


mrpollo:
Thanks @patrickelectric 🎉

Issue #9678, opened 2021-05-25
Compile error: libevents/libs/cpp/parse/parser.h: No such file or directory


heeplr:
## Expected Behavior
Successful build.

## Current Behavior
```
In file included from /var/tmp/portage/app-misc/qgroundcontrol-9999/work/qgroundcontrol-9999/src/Vehicle/EventHandler.h:18,
                 from /var/tmp/portage/app-misc/qgroundcontrol-9999/work/qgroundcontrol-9999/src/Vehicle/Vehicle.cc:52:
/var/tmp/portage/app-misc/qgroundcontrol-9999/work/qgroundcontrol-9999/src/Vehicle/HealthAndArmingChecks.h:16:10: fatal error: libevents/libs/cpp/parse/parser.h: No such file or directory
```

## System Information
- Operating System: Gentoo Linux amd64
- QGC Version: current master
- QGC build: -

## Log Files and Screenshots
s. attached [build.log](https://github.com/mavlink/qgroundcontrol/files/6539405/build.log)

WTPENGUIN:
Try this command in the source code folder.

git submodule update --init --recursive

heeplr:
I double checked the source tree and the file is there. It's just not found.

The compile command doesn't seem to contain the needed include path. The only include paths inside "libs" are:
* /libs/eigen 
* /libs/mavlink/include/mavlink/v2.0
* /libs/mavlink/include/mavlink/v2.0/ardupilotmega
* /libs/mavlink/include/mavlink/v2.0/common
* /libs/shapelib

The problem is probably caused by missing cmake integration (?) and triggered when building outside of the sourcetree. It should be reproducable with:
```
$ git clone --recursive -j8 https://github.com/mavlink/qgroundcontrol.git
$ cd qggroundcontrol
$ git submodule update --recursive
$ mkdir build
$ cd build
$ cmake ..
$ make
```

DonLakeFlyer:
Change the Events submodule INCLUDEPATH in QGCExternalLibs.pri to this:
```
INCLUDEPATH += \
        libs/libevents \
        libs/libevents/libs/cpp/parse
```

We're still not exactly sure why that is needed since it is a relative include. But it fixes it. I'll make a pull for this today.

heeplr:
Does that fix it for you? Neither this nor the actual path ("libs/libevents/libevents/libs/cpp/parse") does for me.

[...]

PR #9682, opened 2021-05-26
Rework altitude mode internals/ui

DonLakeFlyer:
* I'm working my towards support Terrain Frame (MAV_FRAME_GLOBAL_TERRAIN_ALT) in Surveys
* First step was taking another cleanup pass through the "Altitude Mode" concept
* Changed naming of "Above Terrain" to "Calculated Above Terrain" to be more specified
* Change Alt Mode selection from Combo dropdown to full dialog to provide better information on each possibility
* If you always use Relatative Alts and don't know anything about what I'm talking about then the UI for you will look exactly like it did before

Mission Start
* New Alt Mode picker dialog
* You also get this dialog when picking Alt Mode from a simple item if you are using Mixed Modes
Screen Shot 2021-05-26 at 2 05 06 PM

Simple Item Editor
* New ui for altitude value display. Showing non-relative example. Relative Alts ui is still as before.
Screen Shot 2021-05-26 at 2 03 22 PM

### DIFF
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -112,8 +112,9 @@ SOURCES += \
 	libs/libevents/libevents/libs/cpp/protocol/receive.cpp \
 	libs/libevents/libevents/libs/cpp/parse/parser.cpp \
 	libs/libevents/definitions.cpp
-INCLUDEPATH += libs/libevents
-
+INCLUDEPATH += \
+        libs/libevents \
+        libs/libevents/libs/cpp/parse
 #
 # [REQUIRED] shapelib library
 INCLUDEPATH += libs/shapelib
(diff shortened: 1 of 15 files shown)


DonLakeFlyer:
Forgot to say that this also includes the fix for #9678

Antiheavy:
Excited to see this.  We’ve been using the older “follow terrain” in QGC 4.0 a lot recently. We have found some pitfalls. Would like to learn more about the latest developments. I’ll try poking around the code…

DonLakeFlyer:
The support I'm adding will be ArduPilot only. As far as I know PX4 doesn't yet support mavlink terrain frame which requires lidar for height or the firmware needs to support the mavlink terrain tile protocol which loads height tiles to the vehicle. There are currently no planned changes to which is basically the existing Calculated Above Terrain support in Survey which generates all the interstitial waypoints manually. If you have problem with the existing functionality write up an issue which details and I'll see what is fixable.

PR #9684, opened 2021-05-28
RC to Param: Fix swapped values

### LINKED ISSUE #9637: Set RC to Param - mismatch values
I am testing "Set RC to Param" function in my custom autopilot firmware. It seems like values "Scale" and "Center value" are switched.

In version 3.5.6 it looks ok, but since version 4.x.x it doesn't work properly.

DonLakeFlyer:
Fix #9637


patrickelectric (approved):
Looks correct based on the mavlink API.

PR #9691, opened 2021-06-03
PX4FirmwarePlugin: ignore minimum takeoff altitude

julianoes:
The old behaviour adjusts the takeoff altitude set by the user with a slider to be higher than the minimum takeoff altitude as definied by the param MIS_TAKEOFF_ALT. This can be very confusing (and dangerous) as the slider to enter the altitude does not reflect that. It always goes from 0m to 100m.
For example: the param is set to 10m. Now the user can command a takeoff and set the slider to 1.5m. The drone will then takeoff and climb to 10m completely without warning.

I don't think it makes much sense to magically change the user input without feedback. Therefore, I removed the adjustment.
With this change the drone will just go to the altitude commanded by the slider.

### DIFF
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -389,17 +389,6 @@ void PX4FirmwarePlugin::_mavCommandResult(int vehicleId, int component, int comm
     }
 }
 
-double PX4FirmwarePlugin::minimumTakeoffAltitude(Vehicle* vehicle)
-{
-    QString takeoffAltParam("MIS_TAKEOFF_ALT");
-
-    if (vehicle->parameterManager()->parameterExists(FactSystem::defaultComponentId, takeoffAltParam)) {
-        return vehicle->parameterManager()->getParameter(FactSystem::defaultComponentId, takeoffAltParam)->rawValue().toDouble();
-    } else {
-        return FirmwarePlugin::minimumTakeoffAltitude(vehicle);
-    }
-}
-
 void PX4FirmwarePlugin::guidedModeTakeoff(Vehicle* vehicle, double takeoffAltRel)
 {
     double vehicleAltitudeAMSL = vehicle->altitudeAMSL()->rawValue().toDouble();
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
I assume the MIS_TAKEOFF_ALT parameter exists for a reason? Some sort of danger taking off lower than that? Shouldn't the value of that param control the lower bound of the takeoff slider instead?

julianoes:
@DonLakeFlyer from what I understand, the param sets the safe altitude for when you don't specify any altitude in the takeoff command. However, as soon as you command a takeoff in QGC and set the altitude with the slider, you probably know better what you want anyway.

One argument for a minimum altitude was to hover at 2.5m, so higher than people and thus avoid crashing into heads. That works if you're outdoor but leads to a crash/safety issue when you try it indoor.
The other argument was for fixedwing where you don't want to - by accident - take off to 2.5m and then loiter at that altitude because you will likely crash into terrain. However, this takeoff button does not seem activated for the fixedwing case anyway.

DonLakeFlyer:
Got it. Change seems reasonable then.

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
```diff
@@ -408,10 +397,7 @@ void PX4FirmwarePlugin::guidedModeTakeoff(Vehicle* vehicle, double takeoffAltRel
         return;
     }
 
-    double takeoffAltRelFromVehicle = minimumTakeoffAltitude(vehicle);
-    double takeoffAltAMSL = qMax(takeoffAltRel, takeoffAltRelFromVehicle) + vehicleAltitudeAMSL;
-
-    qDebug() << takeoffAltRel << takeoffAltRelFromVehicle << takeoffAltAMSL << vehicleAltitudeAMSL;
+    double takeoffAltAMSL = takeoffAltRel + vehicleAltitudeAMSL;
```
patrickelectric, line 400: missing a const here, but should not hold this PR to get merged.

patrickelectric (approved):
The code looks ok, but I'm not a PX4 expert.

Issue #9546, opened 2021-03-19
Does not create Altitude type "Terrain Frame" in Surveys Plans


framosnt:
Does not create the Altitude type "Terrain Structure" in Plans

Situation:
Use the QGROUND for drones with Radars or Lidar for RangeFinder for the drone to maintain the altitude of the distance sensor (follow terrain), so that it makes a Survey flight plan with the Terrain Frame option, which is determined in the START MISSION, it is also accepted in the creation of individual Waypoints, but, it is not created when used in the Survey option with "Follow Terrain", that for my flight tests, it would be enough to create the waypoints with the Option "MAV_FRAME_GLOBAL_TERRAIN_ALT", that's it!
----

## Expected Behavior
Please tell us what should happen as detailed as possible.

## Current Behavior
Please describe what happens instead of the expected behavior.

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1.
2.
3.
4.

## System Information
When posting bug reports, include the following information
- Operating System: [e.g. iOS 12.1, Windows 8, macOS Mojave 10.14.1, Ubuntu 14]
- QGC Version: [e.g. 3.4.4]
- QGC build: [e.g., daily, stable, self-built from source, etc...]
- Flight Controller: [e.g., Pixhawk I, Pixhawk Mini, Pixhawk 2, etc.]
- Autopilot (with version): [e.g., PX4 1.8.1, Ardupilot Copter 3.6.1]

## Detailed Description
Provide further details about your issue/bug.

## Log Files and Screenshots
- [QGC Console Logs](https://docs.qgroundcontrol.com/en/Support/Support.html#reporting-bugs)
- Autopilot logs when available (post a link).
- Screenshots of QGC to help identify the current issue/bug behavior.

DonLakeFlyer:
@LorenzMeier What's the status of `MAV_FRAME_GLOBAL_TERRAIN_ALT` in PX4? Two questions:
* Does it work with Lidar vehicle?
* Does it support the mavlink terrain tile protocol?

DonLakeFlyer:
[...]

PR #9705, opened 2021-06-11
Support terrain frame in Survey/Corridor Scan as well as simple items

DonLakeFlyer:
This is a pretty massive set of changes to make all of this work:
* Survey/Corridor Scan now support terrain frame natively for usage with mavlink terrain tile protocol or lidar height sensing
* Terrain Profile display shows correct flight profiles when using terrain frame in either simple waypoint missions or patterns
* Fix for #9546

There is still some things not working which I will fix in subsequent pulls:
* When loading a pattern which using terrain frame from a file the terrain profile display does not update. Upload to vehicle will be fine. The plan file itself and data within is correct.
* min/max AMSL alt statistics when loading terrain frame back from a file will be incorrect
* Likely other bugs due to large scale change. Take care in testing this out with real vehicles. Best to round trip through the vehicle to double check the pattern generation.

Simple example using Terrain Frame on waypoints:
Screen Shot 2021-06-11 at 11 10 14 AM

Survey example using Terrain Frame:
Screen Shot 2021-06-11 at 11 10 56 AM

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -180,6 +180,7 @@
         <file alias="QGroundControl/Controls/TakeoffItemMapVisual.qml">src/PlanView/TakeoffItemMapVisual.qml</file>
         <file alias="QGroundControl/Controls/ToolStrip.qml">src/QmlControls/ToolStrip.qml</file>
        	<file alias="QGroundControl/Controls/ToolStripHoverButton.qml">src/QmlControls/ToolStripHoverButton.qml</file>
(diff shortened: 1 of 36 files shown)


DonLakeFlyer:
To be clear this adds support for MAV_FRAME = MAV_FRAME_GLOBAL_TERRAIN_ALT. In order to use that your vehicle must support either the mavlink terrain tile protocol (ArduPilot yes, PX4 no) or using a lidar for height info (ArduPilot yes, PX4 don't know/maybe?).

PR #9708, opened 2021-06-12
Fix PlanMapItems typo error

WTPENGUIN:
When I connected multiple vehicles, I found that the vehicle mission displayed on the screen was not updated properly. I don't know if it's related to this, but I found out that the bug was fixed after applying it. Please review my PR.


DonLakeFlyer:
@WTPENGUIN Thanks for the fix

PR #9709, opened 2021-06-14
VWorld request URL and request method have been changed

WTPENGUIN:
Hello, VWorld request URL and request method have been changed, so l open a PR. I'd appreciate it if you could review my PR(As I am a beginner developer of QGC, I would appreciate it if you could point out if there is anything that does not comply with the rules or made a mistake).


WTPENGUIN:
VWorld Token

I added the VWorld API key input box and tested the map to come out properly by entering the API key that I have (VWorld is a Korean map provider).

maptest

DonLakeFlyer:
@WTPENGUIN Thanks for the pull

PR #9718, opened 2021-06-22
`UnitTest::cleanup()` shall drain any remaining signals or events.

keith-bennett-airmap:
For AirMap, this is PR related to BOSS-741.

Some signals, events, or objects are outliving the unit test which emits it. Any signals or events which do (including deleting objects marked with `deleteLater()`) may cause segmentation faults or use-after-free errors.

## Expected Behavior
Each unit test shall be isolated and should not pollute later unit tests with incorrect, incomplete, invalid, data nor with signals or events pointed to objects that have already been destroyed. Every unit test should have consistent 

This isn't directly a problem with unit tests (per se) but rather with the design of some of the objects being tested.

## Current Behavior
Some unit tests leak objects with (wild) pointers pointing to stack- or heap-based objects whose lifetimes exist exclusively within the unit test. The leaked objects make attempts to access the stack- or heap after the objects have expired and cause problems including spurious failures and segmentation faults during later unit tests.

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1. Checkout or merge this branch
2. Build and run unit tests in Debug mode, or with Address Sanitizer.
- a. In Debug mode, you may observe "random" crashes or some tests may unexpectedly fail.
- b. In Address Sanitizer mode, Address Sanitizer will almost undoubtedly discover at least one heap-use-after-free error, and possibly some additional errors.

## System Information
irrelevant

## Detailed Description
Provide further details about your issue/bug.
Unit tests are correct to exercise code. Some of the exercises are incomplete in that the unit test does not ensure that the code is well-isolated. A test which isn't well isolated may cause problems that make you think are due to other, unrelated, tests.

This PR _does not actually fix anything_. Instead, it exposes the problems so that they can be identified (eg, via Address Sanitizer (see [previous](https://github.com/mavlink/qgroundcontrol/pull/9595#issuecomment-824974766) [mentions](https://github.com/mavlink/qgroundcontrol/issues/9626)). Because it doesn't relate to any one single fix I have therefore opted to keep the change in this PR separate from any distinct fixes.

## Log Files and Screenshots
- Note that there will be a follow-up PR which addresses the screenshot and log below
- Screenshot from 2021-06-22 10-57-56
```PASS   : SpeedSectionTest::_testDirty()
[...]

PR #9719, opened 2021-06-22
Fix polluted CI caused by leaking certain specific `VisualMissionItem`s

keith-bennett-airmap:
Related to #9718; this PR addresses AirMap's BOSS-741.

The problem is that some of the `VisualMissionItem`s lifetimes are exceeding the `MissionController` object that they're attached to and leaking beyond the scope of the unit test in which they were created.

By leaking beyond the scope of the unit test, the `TerrainQuery` classes will return an answer after the unit test has finished... and after another (or several) other unit tests have started. By exceeding the lifetime of its associated MissionController, its internal `_missionController` pointer is now wild. Accessing that pointer causes `heap-use-after-free` or segmentation fault.

The problem lies in the design of the TerrainQuery (it's far too easy to misuse), the design of `VisualMissionItem` (if its lifetime needs to be tied to the `MissionController` then the `MissionController` should create it), and `QmlObjectListModel` (for many reasons, not least of which is that it doesn't actually _own_ the objects in it, and you have to _actively_ tell it to delete its pointers... and doing so will only call `deleteLater()` on the objects instead of deleting them immediately). All of these combine to make the lifetimes of each of these objects far too complicated and easy to get wrong.

This table might be easier to understand (top is most-recent call):

| Thread 1 (during `UnitTest::cleanup()` |
| - |
| [:fire:](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/MissionController.cc#L2262) Address Sanitizer detects heap-use-after-free of MissionController object |
| [MissionController::plannedHomePosition()`](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/SimpleMissionItem.cc#L1079) |
| [SimpleMissionItem::amslEntryAlt()](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/VisualMissionItem.cc#L240) |
| [VisualMissionItem::_amslEntryAltChanged()](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/SimpleMissionItem.cc#L164) (slot) |
| [VisualMissionItem::terrainAltitudeChanged()](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/VisualMissionItem.cc#L210) (signal) |
| [VisualMissionItem::_terrainDataReceived()](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/VisualMissionItem.cc#L200) (slot) |
| [TerrainAtCoordinateQuery::terrainDataReceived()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L728) (signal) |
| [TerrainAtCoordinateQuery::_signalTerrainData()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L635) |
| [TerrainAtCoordinateQuery::_batchFailed()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L686) |
| [TerrainAtCoordinateBatchManager::_coordinateHeights()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L577) (slot) |
| [TerrainQueryInterface::coordinateHeightsReceived()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L848) (signal) |
| [UnitTestTerrainQuery::requestCoordinateHeights()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L286) |
| [TerrainOfflineAirMapQuery::requestCoordinateHeights()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L625) |
| [TerrainAtCoordinateBatchManager::_sendNextBatch](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L576) (slot) |
| [QTimer::timeout()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L576) (signal) |
| QApplication::processEvents() |
| UnitTest::cleanup() |

So: that call stack which demonstrates that the timer's callback has outlived the unit test. But where does the timer's callback get started? See below.

To be clear, we're looking at [`TerrainAtCoordinateBatchManager::_batchTimer`](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.h#L198).

Ultimately, the problem is caused by a violation of [this comment](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.h#L202-L209). The comment is in a random place and there is no strict guarantee that the object will live its necessary lifetime. So this is ultimately a very big design problem that needs to be resolved.

| Thread 1 (_possibly_ during unit test run, or also possibly during `UnitTest::cleanup()`) |
| - |
| [_batchTimer.start()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L587) |
| [_TerrainAtCoordinateBatchManager->addQuery()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Terrain/TerrainQuery.cc#L718) |
| [_currentTerrainAtCoordinateQuery->requestData()](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/VisualMissionItem.cc#L203) |
| [VisualMissionItem::_reallyUpdateTerrainAltitude(https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/VisualMissionItem.cc#L52)]() (slot) |
| [QTimer::timeout()](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/VisualMissionItem.cc#L52) (signal) |

Why yes, that _is_ another timer. But, how does that one get started? Consider the following call stack:

| Thread 1 (during unit test run) |
| - |
| [_updateTerrainTimer.start()](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/VisualMissionItem.cc#L185) |
| [VisualMissionItem::_updateTerrainAltitude()](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/VisualMissionItem.cc#L173) (slot) |
| [VisualMissionItem::coordinateChanged()](https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/VisualMissionItem.h#L54) (signal) |

Okay, so `VisualMissionItem` (such as a SimpleMissionItem) gets called from _just about anywhere_ -- any time the visual mission item's coordinate is written to. Any unit test which creates a VisualMissionItem and assigns it a coordinate "could" trigger this problem if it doesn't take care to ensure that the timers are stopped or expired! Some surveys could create hundreds or thousands of timers. Indeed, some surveys might create and destroy many VisualMissionItems.

This PR does _not_ aim to fix the poor design of the `TerrainQuery`. It is aimed to triage the heap-use-after-free and help fix some of the polluted unit tests.

You should take note that the `VisualMissionItem` seems to also outlive the unit test. That's because of yet another very big design flaw: the `QmlObjectListModel`, and _nearly every single use of `QObject::deleteLater`_.

Address Sanitizer is pretty great. One of the pretty cool things it does, in addition to telling us exactly _what_ was used after it was freed, is to also tell us where that object was first allocated and finally freed.

The initial log output, [including changes from the branch which drains the application event queue between each test](https://github.com/airmap/qgroundcontrol/tree/keith/BOSS-741/fix-tests), is [here](https://github.com/mavlink/qgroundcontrol/files/6697575/A.txt). It simply demonstrates that there is (at least one) heap-use-after-free in the base branch.

[This log output](https://github.com/mavlink/qgroundcontrol/files/6697617/F.txt) roughly corresponds to commit `8ecee67f43c68641a8a7a34adadb49b1b7a16c32` wherein I'm logging the lifetime of each `SimpleMissionItem` to eliminate each leak.

### DIFF
--- a/src/MissionManager/VisualMissionItem.cc
+++ b/src/MissionManager/VisualMissionItem.cc
@@ -19,6 +19,8 @@
 #include "PlanMasterController.h"
 #include "QGC.h"
 
+QGC_LOGGING_CATEGORY(VisualMissionItemLog, "VisualMissionItemLog")
+
 const char* VisualMissionItem::jsonTypeKey =                "type";
 const char* VisualMissionItem::jsonTypeSimpleItemValue =    "SimpleItem";
 const char* VisualMissionItem::jsonTypeComplexItemValue =   "ComplexItem";
@@ -31,6 +33,7 @@ VisualMissionItem::VisualMissionItem(PlanMasterController* masterController, boo
     , _controllerVehicle(masterController->controllerVehicle())
 {
     _commonInit();
+    qCDebug(VisualMissionItemLog) << this << "ctor";
 }
 
 VisualMissionItem::VisualMissionItem(const VisualMissionItem& other, bool flyView, QObject* parent)
@@ -40,6 +43,7 @@ VisualMissionItem::VisualMissionItem(const VisualMissionItem& other, bool flyVie
     *this = other;
 
     _commonInit();
+    qCDebug(VisualMissionItemLog) << this << "ctor";
 }
 
 void VisualMissionItem::_commonInit(void)
@@ -75,7 +79,13 @@ const VisualMissionItem& VisualMissionItem::operator=(const VisualMissionItem& o
 }
 
 VisualMissionItem::~VisualMissionItem()
-{    
+{
+    if (_currentTerrainAtCoordinateQuery) {
+        disconnect(_currentTerrainAtCoordinateQuery, &TerrainAtCoordinateQuery::terrainDataReceived,
+                this, &VisualMissionItem::_terrainDataReceived);
+        _currentTerrainAtCoordinateQuery = nullptr;
+    }
+    qCDebug(VisualMissionItemLog) << this << "dtor";
 }
 
 void VisualMissionItem::setIsCurrentItem(bool isCurrentItem)
@@ -206,6 +216,7 @@ void VisualMissionItem::_reallyUpdateTerrainAltitude(void)
 
 void VisualMissionItem::_terrainDataReceived(bool success, QList<double> heights)
 {
+    qCDebug(VisualMissionItemLog) << this << "_terrainDataReceived";
     _terrainAltitude = success ? heights[0] : qQNaN();
     emit terrainAltitudeChanged(_terrainAltitude);
     _currentTerrainAtCoordinateQuery = nullptr;
--- a/src/MissionManager/VisualMissionItem.cc
+++ b/src/MissionManager/VisualMissionItem.cc
@@ -19,8 +19,6 @@
 #include "PlanMasterController.h"
 #include "QGC.h"
 
-QGC_LOGGING_CATEGORY(VisualMissionItemLog, "VisualMissionItemLog")
-
 const char* VisualMissionItem::jsonTypeKey =                "type";
 const char* VisualMissionItem::jsonTypeSimpleItemValue =    "SimpleItem";
 const char* VisualMissionItem::jsonTypeComplexItemValue =   "ComplexItem";
@@ -33,7 +31,6 @@ VisualMissionItem::VisualMissionItem(PlanMasterController* masterController, boo
     , _controllerVehicle(masterController->controllerVehicle())
 {
     _commonInit();
-    qCDebug(VisualMissionItemLog) << this << "ctor";
 }
 
 VisualMissionItem::VisualMissionItem(const VisualMissionItem& other, bool flyView, QObject* parent)
@@ -43,7 +40,6 @@ VisualMissionItem::VisualMissionItem(const VisualMissionItem& other, bool flyVie
     *this = other;
 
     _commonInit();
-    qCDebug(VisualMissionItemLog) << this << "ctor";
 }
 
 void VisualMissionItem::_commonInit(void)
@@ -85,7 +81,6 @@ VisualMissionItem::~VisualMissionItem()
                 this, &VisualMissionItem::_terrainDataReceived);
         _currentTerrainAtCoordinateQuery = nullptr;
     }
-    qCDebug(VisualMissionItemLog) << this << "dtor";
 }
 
 void VisualMissionItem::setIsCurrentItem(bool isCurrentItem)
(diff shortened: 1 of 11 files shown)


src/MissionManager/VisualMissionItem.cc
```diff
@@ -75,7 +75,12 @@ const VisualMissionItem& VisualMissionItem::operator=(const VisualMissionItem& o
 }
 
 VisualMissionItem::~VisualMissionItem()
-{    
+{
+    if (_currentTerrainAtCoordinateQuery) {
+        disconnect(_currentTerrainAtCoordinateQuery, &TerrainAtCoordinateQuery::terrainDataReceived,
```
DonLakeFlyer, line 80: I though Qt did this automatically when a QObject was deleted? Any signals in/out will be disconnected.

DonLakeFlyer:
So in regular QGC code VIsualMissionItems superclasses are parented to MissionController. So they end up being destructed at the same time thus prevented dangling reference problems. But the unit tests parent the items to themselves which breaks the behavior and lifetime handling.

I think the correct thing to do is to change VisualMissionItem to not allow parent to be specified. And instead force the contract by automatically always parent to the master controller in the base class. This way the lifetime of both are tied together.

Also this:
```
void VisualMissionItemTest::cleanup(void)
{
    UnitTest::cleanup();
    _masterController->deleteLater();
}
```
should not be a deleteLater. I believe "most" usage of deleteLater is incorrect throughout the codebase. But the problem is that I don't think it's 100% true.

keith-bennett-airmap:
If I understand you correctly @DonLakeFlyer: you're saying that it should be invalid for a `VisualMissionItem` or any of its derived types to exist before, after, or effectively without an associated `MissionController`?

DonLakeFlyer:
> If I understand you correctly @DonLakeFlyer: you're saying that it should be invalid for a `VisualMissionItem` or any of its derived types to exist before, after, or effectively without an associated `MissionController`?

That is correct

DonLakeFlyer:
Also to be clear PlanMasterControl and friends (MissionController, RallyPointControl, ...) are essentially globals. Except that there are two of them: one for the Fly View and one for the Plan View. Which is why they need to be passed into all the VIsualMissionItems since both views have their own independent set of items.

keith-bennett-airmap:
Superseded by https://github.com/mavlink/qgroundcontrol/pull/9727.

PR #9722, opened 2021-06-24
When splitting a polygon, update the selected vertex index if necessary.

keith-bennett-airmap:
When splitting a polygon, the selected vertex needs to be updated. Otherwise the selected vertex won't always be the same one that the user selected prior to the split.


DonLakeFlyer:
Unit testing to validate. Awesome.

PR #9727, opened 2021-06-25
Fix MissionController/VisualMissionItem lifecycle problems

DonLakeFlyer:
* MissionController and VisualMissionItem are tied together from a lifecycle standpoint. They must come and go at the same time.
* This change enforces that contract whereas before it was only a coding standard that was followed incorrectly in various places like unit tests.
* Replacement for #9719

### DIFF
--- a/src/MissionManager/CameraCalcTest.cc
+++ b/src/MissionManager/CameraCalcTest.cc
@@ -32,8 +32,13 @@ void CameraCalcTest::init(void)
 
 void CameraCalcTest::cleanup(void)
 {
+    delete _masterController;
     delete _cameraCalc;
     delete _multiSpy;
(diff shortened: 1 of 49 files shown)


DonLakeFlyer:
@keith-bennett-airmap Can you run unit tests through with your memory stuff? This should fix the problems.

keith-bennett-airmap (approved):
It looks good to me. It seems to solve the immediate problem regarding leaking objects that cause CI pollution. It runs clean with Address Sanitizer with `ASAN_OPTIONS=detect_leaks=0`.

DonLakeFlyer:
Grea. Thanks for testing @keith-bennett-airmap

PR #9731, opened 2021-06-30
Fix warnings related to not used variables 

patrickelectric:


DonLakeFlyer:
@patrickelectric FYI: I've stopped using Q_UNUSED as a coding style moving forward and switched to commenting out the arguments instead:

```
void foo(int param, int /*unused*/)
```

The reason being that when the parameter eventually gets used the Q_UNUSED tended to get left hanging around. Whereas commenting out forces the change when you use it.

PR #9732, opened 2021-06-30
PhotoVideoControl.qml: fix Slider setting some facts to minimum:

Davidsastresas:
for some reason the QGCSlider that is assigned to the activesettings of _mavlinkcamera was setting the facts to its minimum value when that menu was instantiated. 

I realized this testing the camera protocol with a Workswell WirisPro  camera. This only seemed to affect some facts instantiated by a facttextfield. The fact that was being reset had the following .xml description:

parameter name="PIP_OPACITY" type="uint8" default="50" min="10" max="100"

but for instance there was another pretty similar that behaved correctly:

parameter name="PALETTE" type="uint8" default="0"

I have not dug deeper than this into all the fact generation from the .xml, maybe the bug is coming from there, but this commit fixes the behaviour.

### DIFF
--- a/src/FlightMap/Widgets/PhotoVideoControl.qml
+++ b/src/FlightMap/Widgets/PhotoVideoControl.qml
@@ -467,8 +467,19 @@ Rectangle {
                                 stepSize:                   parent._fact.increment
                                 visible:                    parent._isSlider
                                 updateValueWhileDragging:   false
-                                onValueChanged:             parent._fact.value = value
-                                Component.onCompleted:      value = parent._fact.value
+                                property bool initialized:  false
+
+                                onValueChanged: {
+                                    if (!initialized) {
+                                        return
+                                    }
+                                    parent._fact.value = value
+                                }
+
+                                Component.onCompleted: {
+                                    value = parent._fact.value
+                                    initialized = true
+                                }
                             }
                             QGCSwitch {
                                 checked:        parent._fact ? parent._fact.value : false


DonLakeFlyer:
Is it possible for you to debug this to understand why the Component.onCompleted is not working? I was trying to debug this myself but SITL doesn't support a camera with any settings and I don't have anything else to try with.

DonLakeFlyer:
Something like:
```
Component.onCompleted: { console.log("got here", parent._fact.value);  value = parent._fact.value }
```

To see what's going wrong.

Davidsastresas:
Okey, I debugged further and it seems the problem is in:

onValueChanged:             parent._fact.value = value

I debugged as you said in Component.onCompleted but it is like the onValueChanged happened before the debug messagnes in Component.onCompleted.

I commented out the onValueChanged and it doesn't initialize that to the minimum value anymore.

What I don't understand is why that is only happening to that particular parameter I told you in the first comment. The rest of parameters are doing just good, I don't understand the relationship. Do you have anything else in mind I could test here?

DonLakeFlyer:
Ok, I get it now thanks. And remember the slider strangeness.

The onValueChanged is happening because the fact has a min set to it which is >0. The default for Slider.value is 0. And then the min is changed to something above 0 which in turn causes Slider.value to change to that minimum which in turn kicks off a valueChanged signal. All of this happen before onCompleted.

If you look at other examples of slider with min != 0 they have a loadComplete bool which is initially set to false. In onValueChanged if !loadComplete then the fact is not updated. And then in the onComplete which sets the initial value the loadComplete is set to true after the initial value is set into the slider.

And then there should not be an value: binding.

Make sense?

DonLakeFlyer:
All of which is insanity. Not sure if there is a better way to make it all work with less crap around it.

Davidsastresas:
Okey, I understand now, it makes sense!

So the issue really is that onValueChanged signal shouldn't be connected until component.onCompleted. Just out of curiosity, where did you find this? is there any place I can study further qml base elements?

I updated it as you said and did push force.

Thanks!

DonLakeFlyer:
> Just out of curiosity, where did you find this?

Just trial an error.

Issue #9642, opened 2021-05-13
Battery monitor doesn't get updated after changing the battery and restarting the flight controller


buckker:
Hi

I'm using QGC V4.1.2 on a Android device (CUAV H16) with a Nora flight controller (ardupilot, copter 4.0.7). I noticed that the battery monitor is stuck after changing the battery for a new flight. The only way to get correct values is to close the app and restart it. After that QGC shows the correct battery information. 

I normally use Mission Planner. If I restart the flight controller without closing Mission Planner, the battery information will be showed correcty. 

Many thanks in advance

Michael

DonLakeFlyer:
This is a dup of #9209

DonLakeFlyer:
There should be a new daily in a few hours. Can you test it out?

PR #9737, opened 2021-07-01
Restart ArduPilot streams as needed

DonLakeFlyer:
Fix for #9209, #9642


DonLakeFlyer:
@dakejahl Jake, do you think you could test this as well?

PR #9750, opened 2021-07-07
Fix deprecated Qt 5.15 usage

DonLakeFlyer:


src/Airmap/AirMapWeatherInfoManager.cc
```diff
@@ -10,9 +10,8 @@
 #include "AirMapWeatherInfoManager.h"
 #include "AirMapManager.h"
 
-
-#define WEATHER_UPDATE_DISTANCE 50000                   //-- 50km threshold for weather updates
-#define WEATHER_UPDATE_TIME     30 * 60 * 60 * 1000     //-- 30 minutes threshold for weather updates
+#define WEATHER_UPDATE_DISTANCE     50000           //-- 50km threshold for weather updates
+#define WEATHER_UPDATE_TIME_MSECS   30 * 60 * 1000  //-- 30 minutes threshold for weather updates
```
DonLakeFlyer, line 14: As far as I can tell this value was incorrect. It is supposed to be milliseconds.

src/Airmap/QJsonWebToken/src/qjsonwebtoken.cpp
```diff
@@ -186,9 +186,9 @@ bool QJsonWebToken::setAlgorithmStr(QString strAlgorithm)
 	// set algorithm
 	m_strAlgorithm = strAlgorithm;
 	// modify header
-	m_jdocHeader = QJsonDocument::fromJson(QObject::trUtf8("{\"typ\": \"JWT\", \"alg\" : \"").toUtf8()
+    m_jdocHeader = QJsonDocument::fromJson(QStringLiteral("{\"typ\": \"JWT\", \"alg\" : \"").toUtf8()
```
DonLakeFlyer, line 189: This doesn't seem to need to use `tr`? Correct?

jennerl: what's the motivation to change it?

jennerl: I've never touched this code...AirMap didn't author it

DonLakeFlyer: Oh, that's right. This is Gus' code. The calls in this pull are deprecated in Qt 5.15 and won't compile without warnings. These changes get rid of the warnings. This all should be fine then.

PR #9752, opened 2021-07-08
Android: Serial port fix possible crash

lukegluke:
Delete the local JNI reference after use to fix possible crashes after some time periodically scanning ports on several android devices:
JNI ERROR (app bug): local reference table overflow (max=512)

https://stackoverflow.com/questions/26685491/jni-error-app-bug-local-reference-table-overflow-max-512


DonLakeFlyer:
Seems reasonable

PR #9755, opened 2021-07-08
fixes #9754

### LINKED ISSUE #9754: CMake build breaks with gstreamer found because of Microhard/Taisync
## Expected Behavior
Build files can be built.

## Current Behavior
Link phase fails with
```
LINK : fatal error LNK1104: cannot open file 'Microhard.lib'
```

## Steps to Reproduce:
1. Generate build files with gstreamer enabled
2. Leave Microhard and Taisync disabled

## System Information
- Windows 10
- current master (54b15e48c6f569ea955b03d020c2e72180b53322)
- CMake build

## Detailed Description
The problem comes with changes made in PR #9733 and #9736.
An option was added to enable/disable Microhard and Taisync when gstreamer is found.
If gstreamer is found and Microhard or Taisync is off, they are still added in `src/CMakeLists.txt` line 195.

```
if(GST_FOUND)
	target_link_libraries(qgc PUBLIC Microhard)
	target_link_libraries(qgc PUBLIC Taisync)
endif()
```
When build files are generated, the linker now tries to find those libraries. But they are excluded in other places, because the options for Microhard and Taisync are disabled.

kistlin:
- if gstreamer is not found but Microhard or Taisync is enabled
  issue an error message and cancel CMake generation
- exclude the target_link_libraries call if Microhard or Taisync
  is not enabled
- if Microhard or Taisync is enabled add compile definitions to
  enable #defines in the source code, else there are build errors

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,10 @@ if (GST_FOUND)
 
     option(QGC_GST_MICROHARD_ENABLED "Enable microhard" OFF)
     option(QGC_GST_TAISYNC_ENABLED "Enable taisyng" OFF)
+else()
+	if (QGC_GST_MICROHARD_ENABLED OR QGC_GST_TAISYNC_ENABLED)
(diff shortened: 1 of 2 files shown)


patrickelectric (approved):

patrickelectric:
Thanks @kistlin

PR #9757, opened 2021-07-09
Limit disconnection scope when switching active vehicle - don't disconnect PlanMapItems

remster:
This affects multiple vehicle scenarios.
First up, https://github.com/mavlink/qgroundcontrol/pull/9708 was not a typo. `masterController` (renamed to `planMasterController`) referred to the vehicle-scoped `PlanMasterController` instantiated [here](https://github.com/mavlink/qgroundcontrol/blob/master/src/FlightDisplay/FlyViewMap.qml#L295). After https://github.com/mavlink/qgroundcontrol/pull/9708 nothing refers to said `PlanMasterController` so either we have dead code or https://github.com/mavlink/qgroundcontrol/pull/9708 didn't quite get the designer's intent. This happens when we revert https://github.com/mavlink/qgroundcontrol/pull/9708:

change-active mp4

I.e.: 
- the [`PlanMapItems`](https://github.com/mavlink/qgroundcontrol/blob/master/src/FlightMap/MapItems/PlanMapItems.qml) stops receiving updates from the vehicle it is connected to.  
- but we can see both plans! and not just the active one like on master. 

We need to show both plans at once and that also seems what @DonLakeFlyer [once wanted](https://github.com/mavlink/qgroundcontrol/commit/6f6439cb5ec518b97da3a07ab23bc990ce182e6d).

the [`masterController` instances](https://github.com/mavlink/qgroundcontrol/blame/master/src/FlightDisplay/FlyViewMap.qml#L295) lose updates from the vehicle because they are disconnected from it [here](https://github.com/mavlink/qgroundcontrol/pull/9757/files#diff-47961bc8678bd6e019f8a66ebcc0b52dab2779118165ed1401c6e595d0630535L130). This seems a bit inconsiderate from the `PlanMasterController` - if not rude. A specific instance of it disconnects everyone else from `_managerVehicle->missionManager()`. It assumes all interests other components have at knowing what's going on with a specific vehicle are not important. It smells like [too much wild carding](https://github.com/mavlink/qgroundcontrol/pull/9757/files#diff-47961bc8678bd6e019f8a66ebcc0b52dab2779118165ed1401c6e595d0630535R129) indeed - but this time **I** am guessing what @DonLakeFlyer meant with this change: https://github.com/mavlink/qgroundcontrol/commit/916d6ebd4b583d4d7609960f75d84b749b58e32e#diff-47961bc8678bd6e019f8a66ebcc0b52dab2779118165ed1401c6e595d0630535L117

### DIFF
--- a/src/FlightMap/MapItems/PlanMapItems.qml
+++ b/src/FlightMap/MapItems/PlanMapItems.qml
@@ -27,9 +27,9 @@ Item {
 
     property var    _map:                       map
     property var    _vehicle:                   vehicle
-    property var    _missionController:         masterController.missionController
-    property var    _geoFenceController:        masterController.geoFenceController
-    property var    _rallyPointController:      masterController.rallyPointController
+    property var    _missionController:         planMasterController.missionController
+    property var    _geoFenceController:        planMasterController.geoFenceController
+    property var    _rallyPointController:      planMasterController.rallyPointController
     property var    _guidedController:          globals.guidedControllerFlyView
     property var    _missionLineViewComponent
 
--- a/src/FlightMap/MapItems/PlanMapItems.qml
+++ b/src/FlightMap/MapItems/PlanMapItems.qml
@@ -27,9 +27,9 @@ Item {
 
     property var    _map:                       map
     property var    _vehicle:                   vehicle
-    property var    _missionController:         planMasterController.missionController
-    property var    _geoFenceController:        planMasterController.geoFenceController
-    property var    _rallyPointController:      planMasterController.rallyPointController
+    property var    _missionController:         masterController.missionController
+    property var    _geoFenceController:        masterController.geoFenceController
+    property var    _rallyPointController:      masterController.rallyPointController
     property var    _guidedController:          globals.guidedControllerFlyView
     property var    _missionLineViewComponent
 
--- a/src/MissionManager/PlanMasterController.cc
+++ b/src/MissionManager/PlanMasterController.cc
@@ -127,9 +127,9 @@ void PlanMasterController::_activeVehicleChanged(Vehicle* activeVehicle)
 
     if (_managerVehicle) {
         // Disconnect old vehicle. Be careful of wildcarding disconnect too much since _managerVehicle may equal _controllerVehicle
-        disconnect(_managerVehicle->missionManager(),       nullptr, nullptr, nullptr);
-        disconnect(_managerVehicle->geoFenceManager(),      nullptr, nullptr, nullptr);
-        disconnect(_managerVehicle->rallyPointManager(),    nullptr, nullptr, nullptr);
+        disconnect(_managerVehicle->missionManager(),       nullptr, this, nullptr);
+        disconnect(_managerVehicle->geoFenceManager(),      nullptr, this, nullptr);
+        disconnect(_managerVehicle->rallyPointManager(),    nullptr, this, nullptr);
     }
 
     bool newOffline = false;
--- a/src/FlightDisplay/FlyViewMap.qml
+++ b/src/FlightDisplay/FlyViewMap.qml
@@ -286,7 +286,7 @@ FlightMap {
         PlanMapItems {
             map:                    _root
             largeMapView:           !pipMode
-            planMasterController:   _planMasterController
+            planMasterController:   masterController
             vehicle:                _vehicle
 
             property var _vehicle: object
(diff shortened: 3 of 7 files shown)


WTPENGUIN:
Since I am a beginner developer, I think it is correct that I did not understand the designer's intention. Thanks for correcting the part I was wrong.

keith-bennett-airmap:
@remster I don't recall any unit tests failing before this commit and I don't see any updated or added unit tests. Please make sure there's a test which would cover the problem solved.

remster:
> @remster I don't recall any unit tests failing before this commit and I don't see any updated or added unit tests. Please make sure there's a test which would cover the problem solved.

Happy to once @DonLakeFlyer is happy with the principle. I am worried there was a reason for the broad wildcard

src/FlightMap/MapItems/PlanMapItems.qml
```diff
@@ -27,9 +27,9 @@ Item {
 
     property var    _map:                       map
     property var    _vehicle:                   vehicle
-    property var    _missionController:         planMasterController.missionController
-    property var    _geoFenceController:        planMasterController.geoFenceController
-    property var    _rallyPointController:      planMasterController.rallyPointController
+    property var    _missionController:         masterController.missionController
+    property var    _geoFenceController:        masterController.geoFenceController
+    property var    _rallyPointController:      masterController.rallyPointController
```
DonLakeFlyer, line 32: I need to stare at this a little more to make sure. But I believe this change is incorrect. This should continue to the planMasterController which is passed in as a property. The bug is in FlyViewMap.qml. The planMasterController property on PlanMapItems in there should be bound to masterController which is the controller associated with each vehicle.

src/MissionManager/PlanMasterController.cc
```diff
@@ -127,9 +127,9 @@ void PlanMasterController::_activeVehicleChanged(Vehicle* activeVehicle)
 
     if (_managerVehicle) {
         // Disconnect old vehicle. Be careful of wildcarding disconnect too much since _managerVehicle may equal _controllerVehicle
-        disconnect(_managerVehicle->missionManager(),       nullptr, nullptr, nullptr);
-        disconnect(_managerVehicle->geoFenceManager(),      nullptr, nullptr, nullptr);
-        disconnect(_managerVehicle->rallyPointManager(),    nullptr, nullptr, nullptr);
+        disconnect(_managerVehicle->missionManager(),       nullptr, this, nullptr);
+        disconnect(_managerVehicle->geoFenceManager(),      nullptr, this, nullptr);
+        disconnect(_managerVehicle->rallyPointManager(),    nullptr, this, nullptr);
```
DonLakeFlyer, line 132: Makes sense

DonLakeFlyer:
> I don't recall any unit tests failing before this commit and I don't see any updated or added unit tests. Please make sure there's a test which would cover the problem solved.

This is in the Qml UI code which has no unit tests associated with it. It's pretty tricky to write testing code for the UI. I used to have a little bit but it was all to flaky so I turned them all off.

remster:
> @remster I don't recall any unit tests failing before this commit and I don't see any updated or added unit tests. Please make sure there's a test which would cover the problem solved.

there: https://github.com/mavlink/qgroundcontrol/pull/9757/commits/707f482b25235da0b5d44215383cf8e7d3a5ecbf

src/FlightMap/MapItems/PlanMapItems.qml
remster: Done: https://github.com/mavlink/qgroundcontrol/pull/9757/commits/1bba400819eb7ac92d51044f3efb168b4eacfb38

keith-bennett-airmap (approved):
lgtm

DonLakeFlyer:
Are the new unit test methods really needed as opposed to using MultiSignalSpyV2?

remster:
> Are the new unit test methods really needed as opposed to using MultiSignalSpyV2?

@DonLakeFlyer there: https://github.com/mavlink/qgroundcontrol/pull/9757/commits/edd5d4194e181b78341ec016742ac00d0beb31e9 - sorry didn't know of `MultiSignalSpyV2`. I did know of `MultiSignalSpy`, but I didn't have a multi-signal problem and its signature (masks) confused me so i looked elsewhere.

DonLakeFlyer:
Yeah the V1 version was a pain to use, hence the V2 changes.

PR #9758, opened 2021-07-09
Enable translation of videoSource strings

positron96:
Enables translation of video source setting by utilizing Fact mechanism of having a list of raw values for internal usage and a list of translated (cooked) values for display purposes. During the creation of VideoSettings group, two lists are fed to videoSource metadata. Then during operation raw values are used in code. Translation should be loaded from qt translation system.


DonLakeFlyer:
Fantastic, thanks.

PR #9762, opened 2021-07-12
joystick range multiplier

Anonym007Pro:
I have the problem that my controller only covers a range from 1300 to 1700.
the servos then do not move far enough.
I therefore simply have completely parallel to the _exponential
the _joystick_multi built in.
That would help me incredibly if it was simply included in the next release and I can't imagine that I'm the only one who has the problem.

**possible mistakes:**
I don't know if the numbers can get too big.

and I know that I have installed everything correctly and have not forgotten anything.

**Thank you so much!
great software!**


Anonym007Pro:
oops.... 
I will delete that.

PR #9764, opened 2021-07-13
BoardInfo: Add FMUv5X

mrpollo:
@LorenzMeier @dagar, what would be the proper `vendorID` for FMUv5X?


src/comm/USBBoardInfo.json
```diff
@@ -54,6 +55,8 @@
         { "regExp": "^PX4 BL FMU v6U.x$",   "boardClass": "Pixhawk" },
         { "regExp": "^PX4 FMU v6X.x$",      "boardClass": "Pixhawk" },
         { "regExp": "^PX4 BL FMU v6X.x$",   "boardClass": "Pixhawk" },
+        { "regExp": "^PX4 FMU v5X.x$",      "boardClass": "Pixhawk" },
+        { "regExp": "^PX4 BL FMU v5X.x$",   "boardClass": "Pixhawk" },
         { "regExp": "^PX4 FMU v5.x$",       "boardClass": "Pixhawk" },
```
mrpollo, line 60: @dagar not sure how is the regex working on the other end, but would the dot "." on this line need to be escaped to avoid the regex from catching an X or any other char other than the dot?

LorenzMeier:
Vendor: Pixhawk?

LorenzMeier (approved):

davids5:
@LorenzMeier USB Vendor ID is set to 0x3185

src/comm/USBBoardInfo.json
```diff
@@ -13,6 +13,7 @@
         { "vendorID": 9900, "productID": 33,        "boardClass": "Pixhawk",    "name": "AUAV X2.1 FMU V2" },
         { "vendorID": 9900, "productID": 48,        "boardClass": "Pixhawk",    "name": "MindPX FMU V2" },
         { "vendorID": 9900, "productID": 50,        "boardClass": "Pixhawk",    "name": "PX4 FMU V5" },
+        { "vendorID": 9900, "productID": 51,        "boardClass": "Pixhawk",    "name": "PX4 FMU V5X" },
```
davids5, line 16: @mrpollo @LorenzMeier The bootloader that has been out for years use 12677

mrpollo: Thanks, I sent the following https://github.com/mavlink/qgroundcontrol/pull/9816

PR #9769, opened 2021-07-15
first add FMT firmwarePlugin,copy from PX4

wenzhicode:


wenzhicode:
VERY Sorry, I submitted it in the wrong place.

PR #9773, opened 2021-07-19
Enable MavLink Inspector on IOS builds per default

### LINKED ISSUE #9772: Enabeling the MAVLink Inspector on IOS per default
Hi everyone,
In the current IOS version of QGroundControl the Analyze Section is not visible and the MavLink Inspector is not available. I have seen somewhere in the release Notes for 4.X that the graph libs have been updated to now also work on IOS and Android (On android everything works fine)

After spending some time on setting up QT for IOS yesterday I did run a build from latest master and saw that it still does not include MAVLink Inspector, so I enabled it and found it to be working fine on my iPhone and iPad(2017) (althought the iPhone XS screen size is a bit too small to see everything)

So there are 2 things: First I would vote for enabeling MavLink Inspector per default on IOS. It works fine and is also available on Android, also it is a must have feature for tuning custom vehicles (I am playing a lot with ArdupilotRover)

Second the IOS app would need to be updated to the latest version on the AppStore... is there any specific reason this was not done ? Maybe I can help ?

Greetings, and thanks for all the great work on this project
Lukas

s00500:
Closes #9772


DonLakeFlyer:
So the original need for this was because the inspector used to be implemented using Qt Widgets which were not available of tablet builds. Since it has been converted to Qml it can run anywhere. Because of that the whole QGC_ENABLE_MAVLINK_INSPECTOR thing isn't really needed. 

If possible could you instead reverse the state of QGC_ENABLE_MAVLINK_INSPECTOR to QGC_DISABLE_MAVLINK_INSPECTOR,  with the default for it not being defined. This means by default all builds get the inspector and the build files can be cleaned up to remove all the += QGC_ENABLE_MAVLINK_INSPECTOR crap.

s00500:
> So the original need for this was because the inspector used to be implemented using Qt Widgets which were not available of tablet builds. Since it has been converted to Qml it can run anywhere. Because of that the whole QGC_ENABLE_MAVLINK_INSPECTOR thing isn't really needed.
> 
> If possible could you instead reverse the state of QGC_ENABLE_MAVLINK_INSPECTOR to QGC_DISABLE_MAVLINK_INSPECTOR, with the default for it not being defined. This means by default all builds get the inspector and the build files can be cleaned up to remove all the += QGC_ENABLE_MAVLINK_INSPECTOR crap.

Sounds like a good plan, I implemented the requested changes, please check :-)

DonLakeFlyer:
Awesome, thanks so much

PR #9778, opened 2021-07-22
Improve TelemetryValues widget

patrickelectric:
- Improves editing experience and user feedback experience from value settings
- Add option to change layout to right corner (older style) or to the bottom (new one)

More patches will come after this one to rework the add/remove logic for the panel values. 

https://user-images.githubusercontent.com/1215497/127000674-0937e7e5-62ab-4321-8e2a-5fd62fe30957.mp4


DonLakeFlyer:
This is missing the custom build edge layout negotiation mechanism through QGCToolInsets. That said in looking at the current code I think it's wrong already.

DonLakeFlyer:
> This is missing the custom build edge layout negotiation mechanism through QGCToolInsets. That said in looking at the current code I think it's wrong already.

If you want you can just blow that part off and I can deal with it. If so, can you enter an issue on it an assign to me and I'll take care of it.

src/FlightDisplay/FlyViewWidgetLayer.qml
```diff
@@ -151,7 +151,38 @@ Item {
...
+                }
+            },
+
+            State {
+                name: "right"
+                when: !telemetryPanel.bottomMode
+
+                AnchorChanges {
+                    target: telemetryPanel
+                    anchors.top: photoVideoControl.bottom
```
DonLakeFlyer, line 180: The photoVideoControl is not always visible. It can be turned off by a setting.

patrickelectric: Done, check the latest version of the video.

jaxxzer:
We got several feedback about it was not clear that you could or how you would reconfigure the telemetry values, and also that the placement was less aesthetic or convenient than earlier versions. I think adding the 'anchor position' button and the 'edit/lock' button are good improvements to address that feedback.

One suggestion for futher improvement I would make is that when 'edit mode' is enabled, that you draw a border around each cell on mouseover, to give the user an indication that they can click on an individual cell to configure it.

jaxxzer (approved):
I can't speak to the implementation or patch details, but I approve the functional change.

DonLakeFlyer:
> We got several feedback about it was not clear that you could or how you would reconfigure the telemetry values, and also that the placement was less aesthetic or convenient than earlier versions.

Yeah. I think this needs some re-thinking. I was moving more towards how the custom build example does it's UX which seemed to work better. But the current state doesn't work that well.

PR #9782, opened 2021-07-23
Improve parameter bitmask value on parameter widget

patrickelectric:
Show a list of selected bit and the string value for each one over the "Unknown: X" string

https://user-images.githubusercontent.com/1215497/126830106-ae5157a0-c66d-454e-917c-e0886db4ea0f.mp4

### DIFF
--- a/src/FactSystem/Fact.cc
+++ b/src/FactSystem/Fact.cc
@@ -303,6 +303,39 @@ QVariantList Fact::bitmaskValues(void) const
     }
 }
 
+/**
+ * @brief Provide a list of selected strings based on the fact value with the bitmaskString/bitmaskValues map
+ *
+ * @return QStringList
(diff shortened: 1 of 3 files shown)


patrickelectric:
ping @DonLakeFlyer

jaxxzer:
I think it's better and more user friendly.

jaxxzer (approved):

DonLakeFlyer:
Seems to me that this new text should just be returned directly in `valueString` for this case instead of adding a new property.

DonLakeFlyer:
> Seems to me that this new text should just be returned directly in `valueString` for this case instead of adding a new property.

@patrickelectric Can you update?

DonLakeFlyer:
Good enough for now

PR #9789, opened 2021-07-27
WIP: APMSubFrameComponent: Improve Sub frame selection

patrickelectric:
- Use frame option as parameter source
    - Remove the duplicated step when the usuer need to select frame and load parater and again the frame


Williangalvani:
I have mixed feelings about this...
can you add a note in the text (qml) of how to restore the default parameters for the current frame? Can you click the frame if it is already selected?

I'm also wondering if we should ask for confirmation before loading the parameters, or at least show a notce instead of doing it silently.

PR #9793, opened 2021-07-28
JoystickConfigButton: Disable joystick buttons that can't be used with vehicle

patrickelectric:
old:
image

new:

image


Williangalvani:
What if instead of hiding the buttons entirely, we replace the comboboxes with "no firmware support for button <number>" or something similar?
on some builds, one of my regular buttons gets mapped to 16. I see it activating and I know I can't use it. but if nothing showed up I would just be confused.

patrickelectric:
@Williangalvani done as requested

Williangalvani (approved):

patrickelectric:
Rebased.

PR #9794, opened 2021-07-29
WIP: Remove mavlink fork of gst-plugins-good

patrickelectric:
Uses original repository of gst-plugins-good.
Tested under android and linux.

### DIFF
--- a/.github/workflows/linux_debug.yml
+++ b/.github/workflows/linux_debug.yml
@@ -14,7 +14,7 @@ defaults:
 
 env:
   SOURCE_DIR:   ${{ github.workspace }}
-  QT_VERSION:   5.12.6
+  QT_VERSION:   5.15.2
 
 jobs:
   build:
--- a/.github/workflows/linux_release.yml
+++ b/.github/workflows/linux_release.yml
@@ -14,7 +14,7 @@ defaults:
 
 env:
   SOURCE_DIR:   ${{ github.workspace }}
-  QT_VERSION:   5.12.6
+  QT_VERSION:   5.15.2
   ARTIFACT:     QGroundControl.AppImage
 
 jobs:
--- a/.github/workflows/macos_release.yml
+++ b/.github/workflows/macos_release.yml
@@ -14,7 +14,7 @@ defaults:
 
 env:
   SOURCE_DIR:   ${{ github.workspace }}
-  QT_VERSION:   5.12.6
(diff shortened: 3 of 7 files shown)


DonLakeFlyer:
If I remember correctly the reason for this fork is that it contains windows fixes which are not in the normal repo.

andrewvoznytsa:
Confirming - if you need to get it working properly in case of multiple
video streams on Windows then you have to use this fork.

DonLakeFlyer:
> if you need to get it working properly in case of multiple video streams

And multiple video stream is part of the camera mavlink spec isn't it?

andrewvoznytsa:
> > if you need to get it working properly in case of multiple video streams
> 
> And multiple video stream is part of the camera mavlink spec isn't it?

Unfortunately - I don't know - mavlink is still black box for me.

DonLakeFlyer:
> Confirming - if you need to get it working properly in case of multiple video streams on Windows then you have to use this fork.

Which is either something already supported or if not, requested as a feature. So it seems bad to lose these fixes. @patrickelectric  What's the underlying reason to remove the way it is now?

andrewvoznytsa:
BTW, there were some activities on qmlglsink/gstreamer side regarding Windows support recently. I think they have the fix in the master. At some point it will be possible to drop this qmlglsink maintenance.

holden-zenith:
@patrickelectric The fix for the windows bug @DonLakeFlyer mentioned is included in [1.19.2](https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/-/merge_requests/1032) and can be found in this gst-plugins-good [here](https://github.com/GStreamer/gst-plugins-good)

holden-zenith:
@patrickelectric The fix for the windows bug @DonLakeFlyer mentioned is included in [1.19.2](https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/-/merge_requests/1032) and can be found in this gst-plugins-good [here](https://github.com/GStreamer/gst-plugins-good)

DonLakeFlyer:
@zdanek Maybe something else you can look into with the video stuff?

zdanek:
I will take a look

HTRamsey:
As of 1.19.x gst-plugins-good is no longer separate from the base repo, and since only a single plugin is used from this I think it is better to use the approach used in #11258

PR #9798, opened 2021-07-30
QGCComboBox: Check for model before iterating over it

patrickelectric:

### DIFF
--- a/src/QmlControls/QGCComboBox.qml
+++ b/src/QmlControls/QGCComboBox.qml
@@ -53,7 +53,7 @@ T.ComboBox {
     }
 
     function _adjustSizeToContents() {
-        if (_onCompleted && sizeToContents) {
+        if (_onCompleted && sizeToContents && model) {
             _largestTextWidth = 0
             for (var i = 0; i < model.length; i++){
                 textMetrics.text = model[i]


src/QmlControls/QGCComboBox.qml
```diff
@@ -53,7 +53,7 @@ T.ComboBox {
     }
 
     function _adjustSizeToContents() {
-        if (_onCompleted && sizeToContents) {
+        if (_onCompleted && sizeToContents && model) {
```
DonLakeFlyer, line 56: I would rather have this fail in this case. Why would not having a model at all for a combo box be a reasonable thing?

patrickelectric: So, what should be the correct approach for the following line ?
`fact: controller.parameterExists(-1, "PARAMETER") ? controller.getParameterFact(-1, "PARAMETER") : null;`

The parameter may not exist and we need to populate fact with an empty or invalid state, since it appears not possible to do so with fact using `[]` and `null/undefined`, what should be the correct approach ?
To me, it makes sense to check for valid models inside the combobox.

DonLakeFlyer: > The parameter may not exist and we need to populate fact with an empty or invalid state

Yup, according to the FactComboBox implementation a null Fact is a perfectly fine thing. As is no model on a combo. So concept of catching failures doesn't make sense.

PR #9801, opened 2021-07-30
Load AirMap dynamically to avoid loading unavailable components 

patrickelectric:
Avoid loading AirMap if the support does not exist

### DIFF
--- a/src/FlightDisplay/FlyViewInstrumentPanel.qml
+++ b/src/FlightDisplay/FlyViewInstrumentPanel.qml
@@ -23,11 +23,21 @@ Column {
 
     property real availableHeight
 
-    AirspaceControl {
-        id:                 airspaceControl
-        width:              parent.width
-        planView:           false
-        visible:            QGroundControl.airmapSupported ? QGroundControl.settingsManager.airMapSettings.enableAirMap.rawValue : false
+    Component {
+        id: airspaceControlComponent
+        AirspaceControl {
+            id:                 airspaceControl
+            width:              parent.width
+            planView:           false
+            visible:            QGroundControl.airmapSupported && QGroundControl.settingsManager.airMapSettings.enableAirMap.rawValue
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


patrickelectric:
It would be nice to have someone of AirMap to test it

patrickelectric:
@keith-bennett-airmap ping

keith-bennett-airmap:
Hi @patrickelectric ,

I recall from the conversation that you'd said an error message keeps popping up in builds without AirMap configured. When I build QGC with [`DISABLE_AIRMAP`](https://github.com/mavlink/qgroundcontrol/blob/0ed94504b265fb589d7c2bf974c7417bcdb94d54/QGCExternalLibs.pri#L231) defined during the configure step, I do not see any problem messages showing up during flight. This is with [a recent `master`](https://github.com/mavlink/qgroundcontrol/commit/0ed94504b265fb589d7c2bf974c7417bcdb94d54). I also went so far as to remove [the definition of `QGC_AIRMAP_ENABLED`](https://github.com/mavlink/qgroundcontrol/blob/0ed94504b265fb589d7c2bf974c7417bcdb94d54/QGCExternalLibs.pri#L269) and still did not encounter any error messages or issues that I could easily identify.

https://user-images.githubusercontent.com/72767521/130007596-50b9d79e-1d89-41eb-a48e-f9165737e479.mp4

Can you clearly describe the steps to reproduce the issue you were having? I'd like to better understand the problem that this PR tries to solve.

Thanks,
Keith

DonLakeFlyer:
> Can you clearly describe the steps to reproduce the issue you were having? I'd like to better understand the problem that this PR tries to solve.

@patrickelectric ?

patrickelectric:
The idea here is to disable the components in runtime since airmap will not exist when building QGC without airmap support.

Davidsastresas:
As the removal of Airmap in the codebase is in master already I think this can be closed.

PR #9815, opened 2021-08-08
Added SmartAP board USB identifier

aviaks:
Added SmartAP board USB identifier to support the hardware detection properly.


aviaks:
Could you review and merge please? Many thanks in advance!

PR #9835, opened 2021-08-21
Find vehicles on the network with zero-conf (mdns)

patrickelectric:
Find vehicles that uses mavlink type:
https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=mavlink

Supports both tcp and udp connections. 
- [x] Tested on linux.
- [x] Tested on mac. 
- [x] Tested on windows.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "libs/eigen"]
 	path = libs/eigen
 	url = https://gitlab.com/libeigen/eigen.git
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
@patrickelectric This doesn't apply to android?

patrickelectric:
@DonLakeFlyer I have not tested under android but it should

PR #9842, opened 2021-08-27
APM: Mission Flight Speed

dakejahl:
APM firmware does indeed support the MAV_CMD_DO_CHANGE_SPEED command for changing the speed for a mission. I tested this in sim against Copter-4.0.7 and QGC master as of today.

### DIFF
--- a/src/PlanView/MissionSettingsEditor.qml
+++ b/src/PlanView/MissionSettingsEditor.qml
@@ -36,7 +36,7 @@ Rectangle {
     property bool   _waypointsOnlyMode:             QGroundControl.corePlugin.options.missionWaypointsOnly
     property bool   _showCameraSection:             (_waypointsOnlyMode || QGroundControl.corePlugin.showAdvancedUI) && !_controllerVehicle.apmFirmware
     property bool   _simpleMissionStart:            QGroundControl.corePlugin.options.showSimpleMissionStart
-    property bool   _showFlightSpeed:               !_controllerVehicle.vtol && !_simpleMissionStart && !_controllerVehicle.apmFirmware
+    property bool   _showFlightSpeed:               !_controllerVehicle.vtol && !_simpleMissionStart
     property bool   _allowFWVehicleTypeSelection:   _noMissionItemsAdded && !globals.activeVehicle
 
     readonly property string _firmwareLabel:    qsTr("Firmware")


julled:
@booo interesting for us?

booo:
Yes, of course this is of interest for the SearchWing project too.

DonLakeFlyer:
> APM firmware does indeed support the MAV_CMD_DO_CHANGE_SPEED command for changing the speed for a mission.

Yeah, that was known but there was some problem which I seem to remember related to inconsistent use of the values in that message. I'll have to look and see if I can remember. I specifically didn't support it because of some problem I ran into, not because I didn't think it existed.

DonLakeFlyer:
The problem related to DO_CHANGE_SPEED.param1 support for 0=airspeed and 1=groundspeed. Ages ago when I originally implement I vaguely remember ArduPilot not supporting param1 and forcing to one or the other (don't remember which). QGC requires param1 support for it to work. Does ArduPilot support this now?

dakejahl:
>Does ArduPilot support this now?

Yes it supports all 4 options.

**Speed type (0=Airspeed, 1=Ground Speed, 2=Climb Speed, 3=Descent Speed)**

holden-zenith:
It forces to [groundspeed](https://github.com/ArduPilot/ardupilot/blob/b2f00d469e0549742a20030f97b85bd9af03c2cd/ArduCopter/mode_auto.cpp#L1522). Also in [doc](https://ardupilot.org/copter/docs/common-mavlink-mission-command-messages-mav_cmd.html#mav-cmd-do-change-speed)

DonLakeFlyer:
Ok, so according to @holden-zenith (and the code) this won't work since only groundspeed is supported. I'm closing for now until there is other proof that it does work.

julled:
Wouldnt groundspeed also be useful for various of applications already? 
At least for me it would help alot and could be pointed out to the user via GUI or documentation.

DonLakeFlyer:
> Wouldnt groundspeed also be useful for various of applications already?

It would. But the way QGC works is when you change flight speed it changes airspeed for fixed wings and groundspeed for multi-rotors. There is no groundspeed only thing in current QGC code. Also groundspeed for fixed wing is dangerous I assume.

PR #9851, opened 2021-09-01
fix implicitly defined properties in Connections

booo:
Implicitly defined properties are deprecated in newer Qt versions.
This changes the syntax of the property functions to the new style.

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewVideo.qml
+++ b/src/FlightDisplay/FlightDisplayViewVideo.qml
@@ -79,7 +79,7 @@ Item {
 
                 Connections {
                     target: QGroundControl.videoManager
-                    onImageFileChanged: {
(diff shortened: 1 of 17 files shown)


booo:
I'm not sure if this changes is ok for qt version 5.12. See https://stackoverflow.com/questions/62297192/qml-connections-implicitly-defined-onfoo-properties-in-connections-are-deprecat

patrickelectric:
That's a great PR to start Qt5.15 migration

DonLakeFlyer:
Awesome, thanks

booo:
I think there are still a few missing. Maybe I can fix them in another run later on.

DonLakeFlyer:
Yeah, there is still a pretty big pile left. For now I've turned off the warning.

Issue #9847, opened 2021-08-31
Cannot run v4.1.4 AppImage: GLIBCXX_3.4.26 and GLIBC_2.29 not found


mortenfyhn:
I cannot run the uploaded v4.1.4 AppImage.

I'm on Ubuntu 18.04.

* Download v4.1.4 AppImage from the github release page [here](https://github.com/mavlink/qgroundcontrol/releases/tag/v4.1.4). (The one from the user guide [here](https://docs.qgroundcontrol.com/master/en/getting_started/download_and_install.html#ubuntu) is currently identical (md5sum), so this applies to both.)
* `chmod +x ./QGroundControl.AppImage`
* `./QGroundControl.AppImage`

gives

```
/tmp/.mount_QGroundHDUWS/QGroundControl: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by /tmp/.mount_QGroundHDUWS/QGroundControl)
/tmp/.mount_QGroundHDUWS/QGroundControl: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /tmp/.mount_QGroundHDUWS/QGroundControl)
```

The current daily build (appears to be 048081c) runs fine. Maybe v4.1.4 is build with a different compiler / otherwise different version of something, which makes it incompatible with Ubuntu 18.04.

I also recall v4.1.3 running fine, but I cannot find a downloadable AppImage for that version anymore. Is it possible to build one and attach it to the github release for v4.1.3?

I observe this issue on multiple different computers running Ubuntu 18.04.

Appears similar to [this](https://docs.qgroundcontrol.com/master/en/Support/troubleshooting_qgc.html#glibc_2_27) problem for Ubuntu 16.04 users.

Cristian-wp:
Same problem...do you find any solution?

mehmetyldz87:
Hi 
We have same problems. 

https://discuss.px4.io/t/latest-qgc-release/23972/4

patrickelectric:
Can you try https://github.com/mavlink/qgroundcontrol/suites/3677341316/artifacts/89532134 ?
It's the artifact from #9853

mortenfyhn:
> Can you try https://github.com/mavlink/qgroundcontrol/suites/3677341316/artifacts/89532134 ?
> It's the artifact from #9853

Runs properly on my machine!

mehmetyldz87:
[...]

PR #9853, opened 2021-09-03
Update Action to target ubuntu 18.04

mrpollo:
we previously made the mistake of only targeting ubuntu 20.04 leaving a lot of users behind, this PR hopefully fixes that.

### DIFF
--- a/.github/workflows/linux_debug.yml
+++ b/.github/workflows/linux_debug.yml
@@ -18,7 +18,7 @@ env:
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


patrickelectric:
It appears to be fine and the artifact works here, do you have someone using ubuntu 18.04 to test ?
Maybe someone from #9847

patrickelectric (approved):

patrickelectric:
Appears to fix the problem.

PR #9855, opened 2021-09-03
CI: re-enable S3 artifact upload

mrpollo:
Checking if the updated S3 key works

### DIFF
--- a/.github/workflows/linux_debug.yml
+++ b/.github/workflows/linux_debug.yml
@@ -81,7 +81,7 @@ jobs:
         run:  ccache -s
 
       - name: Setup for unit tests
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
> Checking if the updated S3 key works

The way to do this is to just use the aws cli locally with the new key and see if you can upload something from a command line. If that works, then this should work as well. But only after things are merged? Have you verified you can upload things using aws cli with the new key?

mrpollo:
Yeah I tested the keys locally, this was just to check if it would work on CI, but it didn't, I'm not really sure what's going on with S3/AWS

DonLakeFlyer:
@mrpollo Can you email me the key information and also the command line you used to test? I'll take a look as well to see if I can figure anything out.

PR #9872, opened 2021-09-16
Add "Precision Land" and "Land" to the flight mode dropdown for PX4

RafikHachana:
This change makes both the "Precision Landing" and "Land" modes settable for PX4 setups from within the QGC UI. This might be useful and convenient for users who switch to these modes frequently.


DonLakeFlyer:
So this turns on Land as a settable flight mode for Fixed Wings. Is that something that is even possible? What happens when you pick that?

RafikHachana:
> So this turns on Land as a settable flight mode for Fixed Wings. Is that something that is even possible? What happens when you pick that?

Yes, it is possible. The vehicle would automatically switch to Multi-Rotor and land. 
Besides, the Land mode was already set to "FW compatible" before I made it settable through the dropdown menu, which means that it was already supported for FW.

DonLakeFlyer:
> The vehicle would automatically switch to Multi-Rotor and land.

I'm talking about Fixed Wing not VTOL. Previously with FW, Land was just a reported mode which it would change to during a landing sequence. You couldn't actually command a FW to Land flight mode you could only tell it to RTL. What happens when you tell a FW to switch to Land flight mode?

RafikHachana:
> > The vehicle would automatically switch to Multi-Rotor and land.
> 
> I'm talking about Fixed Wing not VTOL. Previously with FW, Land was just a reported mode which it would change to during a landing sequence. You couldn't actually command a FW to Land flight mode you could only tell it to RTL. What happens when you tell a FW to switch to Land flight mode?

Sorry I didn't understand at first. Setting Land manually doesn't work with FW (I assumed it does since the corresponding boolean was set to true). I have removed "Land" from the dropdown menu. Now the diff only makes PrecLand settable for VTOL.

PR #9876, opened 2021-09-17
main: allow quitting using Ctrl+C on Linux

julianoes:
This catches the signals SIGINT and SIGTERM to shut down QGC.

I find this quite convenient as a user and developer. Any objections?

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -211,6 +211,18 @@ bool checkAndroidWritePermission() {
 }
 #endif
 
+// To shut down QGC on Ctrl+C on Linux
+#ifdef Q_OS_LINUX
+#include <csignal>
+
+void sigHandler(int s)
+{
+    std::signal(s, SIG_DFL);
+    QApplication::instance()->quit();
+}
+
+#endif /* Q_OS_LINUX */
+
 //-----------------------------------------------------------------------------
 /**
  * @brief Starts the application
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
> I find this quite convenient as a user and developer. Any objections?

Is that a normal thing for Linux apps? Seems a bit odd to me. Isn't Ctrl+C the shortcut for copy? If so, doesn't this prevent copy from working?

julianoes:
I would argue it's normal, yes. It's Ctrl+C in the console (copy is Ctrl+Shift+C there) but it is not caught when the UI/app is focused.

booo:
For the console it's quite normal but as far as I know for other apps it's not. I guess there are conventions we should follow here. Please look them up.

julianoes:
@booo let me find a few examples:
CLion -> quits
vlc -> quits
gazebo -> quits
libreoffice -> quits
VSCode -> forks away somehow and give the terminal back
chromiu -> quits

I yet have to find one that does not quit with Ctrl+C :smile:.

julianoes:
Is that convincing enough? Or do I need to gather more evidence? :smile: Or what are your concerns?

patrickelectric (approved):
As a Linus user for more than 10 years, this is indeed a fix and makes qgc more friendly for Linux.

booo:
What are we talking about exactly? Are we talking about a general key combination that works always when the focus is on the graphical user interface or are we talking about a combination that only works when we start QGC from the commandline and focus is on the commandline?

patrickelectric:
@booo if you are a user on mac or linux and start QGC binary when building from source from the command line, it's not possible to kill QGC without  the user interface or clicking in X, this adds the **common** linux behavior to kill the process with ctrl+c.
If you are not a linux/mac user/developer this behavior will not make sense for you, but for the user base that I'm saying it does do.

julianoes:
> that only works when we start QGC from the commandline and focus is on the commandline?

Only that.

booo:
Ok. I misunderstood the proposed change. I'm fine with the change as it is.

If you don't mind add this tiny bit of extra information to the commit body. Thanks for your work!

julianoes:
Ok, done, rebased and force-pushed.

PR #9887, opened 2021-09-24
Don't resolve tcp host when adding a config

valbok:
Currently when adding a hostname in string form for tcp connection,
resolving to IP is done immediately, and in case of an error
the host is not stored and used default one instead.

Also the ip might be not known when adding the config, but hostname is known,
which also might be not resolvable at time when the config is being created.

Since hostName in QAbstractSocket::connectToHost
may be an IP address in string form (e.g., "43.195.83.32"),
or it may be a host name (e.g., "example.com").
QAbstractSocket will do a lookup only if required.

So suggesting to totally remove resolving of ip from TCPLink.
Resolving will be performed in QTcpSocket::connectToHost() when actual connection is requested.

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -479,8 +479,7 @@ void LinkManager::_addZeroConfAutoConnectLink(void)
             }
 
             auto link = new TCPConfiguration(tcpName);
-            QHostAddress address(hostname);
-            link->setAddress(address);
+            link->setHost(hostname);
             link->setPort(service.port());
             link->setAutoConnect(true);
             link->setDynamic(true);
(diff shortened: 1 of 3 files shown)


valbok:
1. Setting name `host` has not been changed - https://github.com/mavlink/qgroundcontrol/pull/9887/files#diff-497d2e78aec64b524ab94467b71141764c941fe72a0d735cf297530f171a3a6fR202
2. Setting `host` previously _always_ contained resolved IP address (`address().toString()`), since it has been resolved while adding a link.
3. Now (after this PR) this setting contains either unresolved hostname or IP address provided by a user via GUI - QTcpSocket::connectToHost() does not care.

So if the config contains previously stored IP address, it will be loaded and passed to QTcpSocket::connectToHost() as previously.

valbok:
@DonLakeFlyer Any objections about this?
it now stores hostname to config and resolves it while connecting, instead of resolving the host while adding connection link.

holden-zenith:
There are a couple issues with the comm links and this is a good step forward

mrpollo:
thanks @valbok

Issue #9629, opened 2021-05-04
libairmap-qt.so.1.1.0 not found


julianoes:
## Expected Behavior
I would like to just be able to build QGC and run it, so I can do development.
I use these steps:
```
mkdir build && cd build
~/Qt/5.12.9/gcc_64/bin/qmake -r ../qgroundcontrol.pro CONFIG+=force_debug_in
make -j8
```

## Current Behavior
It builds but doesn't start:

```
./staging/QGroundControl: error while loading shared libraries: libairmap-qt.so.1.1.0: cannot open shared object file: No such file or directory
``` 

Even though the file is there:
```
./staging/Qt/libs/libairmap-qt.so.1.1.0
./libs/airmap-platform-sdk/linux/Qt.5.12/libairmap-qt.so.1.1.0
```

## Steps to Reproduce:
See above.

## System Information
When posting bug reports, include the following information
- Operating System: Arch Linux
- QGC Version: master (2b2d9d826cfd4353c64b98312836b763cadbfa66)

booo:
Did you try something like this: `LD_LIBRARY_PATH=./staging/Qt/libs ./staging/QGroundControl`?

julianoes:
@booo thanks. Yes I use that until it's fixed.

jennerl:
@julianoes `LD_LIBRARY_PATH` is set to include libs in `./staging/Qt/libs` via https://github.com/mavlink/qgroundcontrol/blob/master/deploy/qgroundcontrol-start.sh#L3

Are you using that script to start QGC?

julianoes:
Do I need to use that script? What's wrong with just starting the executable directly? Sorry, I'm playing a bit ignorant here on purpose trying to understand why we need such things.

booo:
For an explanation of the LD_LIBRARY_PATH please follow this link: https://stackoverflow.com/a/7148086

We build libraries and link against them. If the shared libraries are not in your env they can't be found when you start the script. Therefore use the LD_LIBRARY_PATH to add them to your env or use the start.sh script.

I think you can close this ticket. It's not a bug.

PR #9893, opened 2021-09-29
deploy: fix UI not starting

julianoes:
This fixes the problem I had on Arch Linux when building and starting with:

```
mkdir build && cd build
~/Qt/5.15.2/gcc_64/bin/qmake -r ../qgroundcontrol.pro
make -j8
build/staging/qgroundcontrol-start.sh
```

Where I got the error:
```
qrc:/qml/MainRootWindow.qml:25:1: Type ApplicationWindow unavailable
file:///home/julianoes/Qt/5.15.2/gcc_64/qml/QtQuick/Controls.2/qmldir: plugin cannot be loaded for module "QtQuick.Controls": Cannot install type 'VerticalHeaderView' into protected module 'QtQuick.Controls' version '2'
<Unknown File>: Cannot install type 'HorizontalHeaderView' into protected module 'QtQuick.Controls' version '2'
<Unknown File>: Cannot install element 'SplitHandle' into protected module 'QtQuick.Controls' version '2'
<Unknown File>: Cannot install type 'SplitView' into protected module 'QtQuick.Controls' version '2'
...
```

From what I can tell, the only thing that needs to happen in qgroundcontrol-start.sh is that libairmap-cpp.so.2.0.0 is found.
Maybe I'm missing something but this patch works for me.

### DIFF
--- a/deploy/qgroundcontrol-start.sh
+++ b/deploy/qgroundcontrol-start.sh
@@ -1,8 +1,6 @@
 #!/bin/sh
 HERE="$(dirname "$(readlink -f "${0}")")"
-export LD_LIBRARY_PATH="${HERE}/usr/lib/x86_64-linux-gnu":"${HERE}/Qt/libs":$LD_LIBRARY_PATH
-export QML2_IMPORT_PATH="${HERE}/Qt/qml"
-export QT_PLUGIN_PATH="${HERE}/Qt/plugins"
+export LD_LIBRARY_PATH="${HERE}/Qt/libs":$LD_LIBRARY_PATH
 
 # hack until icon issue with AppImage is resolved
 mkdir -p ~/.icons && cp ${HERE}/qgroundcontrol.png ~/.icons


patrickelectric:
Inside the build folder you can just call qgroudcontrol binary directly, there is no need to call start.sh

julianoes:
@patrickelectric that's what I would assume but it doesn't work:

```
julianoes@fractal ‹ master ● › : ~/src/qgroundcontrol
[0] % (cd build && make -j8)                                         
...

julianoes@fractal ‹ master ● › : ~/src/qgroundcontrol
[0] % cd build 

julianoes@fractal ‹ master ● › : ~/src/qgroundcontrol/build
[0] % staging/QGroundControl 
staging/QGroundControl: error while loading shared libraries: libairmap-cpp.so.2.0.0: cannot open shared object file: No such file or directory

julianoes@fractal ‹ master ● › : ~/src/qgroundcontrol/build
[127] % cd staging 

julianoes@fractal ‹ master ● › : ~/src/qgroundcontrol/build/staging
[0] % ./QGroundControl 
./QGroundControl: error while loading shared libraries: libairmap-cpp.so.2.0.0: cannot open shared object file: No such file or directory
```

I've reported this in #9629 and was told that I should be using `LD_LIBRARY_PATH` or the script.

julianoes:
Is something wrong with my changes? From what I can see these paths are no longer valid.

Williangalvani:
I have the exact same issue.

julianoes:
@dakejahl so your change doesn't seem to help my problem.

What I find confusing is that you would have to set the library path to the Qt libs installed system-wide. Wouldn't they be known/found automatically? In my view, only the local / "special" libs need to be manually added to the `LD_LIBRARY_PATH`.

I'm going to fire up my Ubuntu 18.04 VirtualBox and check this.

julianoes:
@dakejahl I tested this in Ubuntu 18.04 and it starts fine with my changes. Note that I'm using the build and start scripts as listed above.

How are you starting it when it does not work? Maybe this has something to do with the AppImage?

dakejahl:
>How are you starting it when it does not work?

Using the script.

keith-bennett-airmap:
Building with `qmake` as your instructions, I see:
```bash
$ mkdir -p build/qmake && cd build/qmake
$ ~/storage/Qt/5.15.2/gcc_64/bin/qmake -r ../../qgroundcontrol.pro
$ make -j$(nproc)
<snip success for brevity>
$ ldd ./staging/QGroundControl | grep -i "not found"
        libairmap-cpp.so.2.0.0 => not found
```

Building with Qt Creator, I see:
```bash
$ ldd ./build/build-qgroundcontrol-Desktop_Qt_5_15_2_GCC_64bit-Debug/staging/QGroundControl  | grep -i "not found"`
        libairmap-cpp.so.2.0.0 => not found
```

There are a bunch of other libs that are in the same `${HERE}/Qt/libs`, such as `libQt5*.so.5`. But when I search for one of them with `ldd`, I see it's not pointing to that one. Instead it's pointing to the original in my Qt installation directory. It points to the correct one if `LD_LIBRARY_PATH` is used.
```bash
$ ls -lah ./staging/Qt/libs | grep libQt5Gui
-rwxr-xr-x 1 kbennett kbennett 8.4M Oct 15 13:16 libQt5Gui.so.5
$ ldd ./staging/QGroundControl | grep -i "Qt5Gui"
        libQt5Gui.so.5 => /home/kbennett/storage/Qt/5.15.2/gcc_64/lib/libQt5Gui.so.5 (0x00007f02c2cf1000)
$ LD_LIBRARY_PATH=./staging/Qt/libs ldd ./staging/QGroundControl | grep -i "Qt5Gui"
        libQt5Gui.so.5 => ./staging/Qt/libs/libQt5Gui.so.5 (0x00007f36b559c000)
```

Using `LD_LIBRARY_PATH` is poor behavior and hiding it behind a startup script doesn't change that. [The correct fix would be to tell the **runtime linker** where to find the libraries](
https://stackoverflow.com/a/8482308). There's an additional compiler flag for that. Here's some information (using GCC as the guinea pig, I'm most familiar with it:
* `-L` is usually used to point to a library that gets linked **at compile time**, eg by the computer compiling QGC (eg, a cross compiler).
* `-rpath` is usually used to point to a library that gets linked **at runtime**, eg by the computer running QGC.

It looks like [clang should be compatible (and thus, macOS and Android)](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-rpath).

I'm not 100% sure which qmake variable to use, there are a few candidates.

So the scope of the problem isn't just airmap. Airmap's lib is just the most annoying since it's typically downloaded to the build directory and not usually installed to the system, so a default system-provided lib isn't found, so it becomes the most visible. But if QGC is built with a different version of Qt than the system-provided one then there may be other runtime problems too.

I'm not 100% sure which qmake variable to use. There are a few candidates [[0](https://forum.qt.io/topic/26321/does-qmake-support-a-qmake_-variable-for-setting-rpath-link/5)]. I'd venture a guess that the correct one would be [`QMAKE_RPATHDIR`](https://doc.qt.io/qt-5/qmake-variable-reference.html#qmake-rpathdir):
> Specifies a list of library paths that are added to the executable at link time so that the paths will be preferentially searched at runtime.
> 
> When relative paths are specified, qmake will mangle them into a form understood by the dynamic linker to be relative to the location of the referring executable or library. This is supported only by some platforms (currently Linux and Darwin-based ones) and is detectable by checking whether QMAKE_REL_RPATH_BASE is set.


I can't promise I'll get a fix in today but I'll try to get a PR to fix within a week unless someone else can fix it. I don't currently have a Mac to test with. I moved recently and my Mac is still packed away. But I can test Linux and Android if someone makes a PR.

keith-bennett-airmap:
This ended up being a bit of a rabbit hole. Using `QMAKE_RPATHDIR` solved the immediate problem with libairmap and the Qt libraries, but it exposed another problem: that the Qt plugins don't get `rpath` set and that it seems that they don't respect `qt.conf`. So the only solution for them is to continue to use `LD_LIBRARY_PATH`, or... use a tool called `patchelf`.

Consider:

```
$ rm -Rf build/ && mkdir -p build/qmake && cd build/qmake
$ qmake -r ../../qgroundcontrol.pro 2>&1 >/dev/null # silence qmake for brevity
$ make -j$(nproc) 2>&1 >/dev/null # silence make for brevity
$ cd staging
$ cat <<EOF > qt.conf
[Paths]
Prefix=./Qt
Libraries=libs
EOF
$ find ./Qt -iname '*.so' -o -iname '*.so.5' | while read -r so; do patchelf --set-rpath '$ORIGIN/'"$(realpath ./Qt/libs --relative-to $(dirname "${so}"))" "${so}"; done
$ ./QGroundControl
```

So, all of the libraries under `./Qt/plugins` and `./Qt/qml` (`find ./Qt -iname '*.so' -o -iname '*.so.5'`) need to have `patchelf --set-rpath` to update the RPATH to point to the `./Qt/libs` dir. Since each library has a different path, `realpath ./Qt/libs --relative-to` is used to generate the relative path from that library to the Qt libraries dir. After doing that, I can successfully run QGroundControl without the start script:

Screenshot from 2021-10-18 16-46-48

That is with some updates to the `QGCPostLinkCommon.pri` as found [here](https://github.com/mavlink/qgroundcontrol/compare/master...airmap:keith/qmake_rpathdir).

There's clearly some more work to do.

* I didn't have to install `patchelf` but it's documented that it's not a common executable. I don't know when/where it was installed. If this route is desired then it would need to be documented and/or added to some dev environment setup script
* The startup script would still be needed for an AppImage icon?
* I don't know about non-Linux platforms...

julianoes:
Not required anymore since #9947.

PR #9896, opened 2021-10-02
FlightDisplay: put zoom buttons on the left of the scale

jolivain:
The map scale width changes when the zoom up/down buttons are clicked.
When buttons are on the right, and when the user quickly clicks a zoom
up/down button several times, the first click change the scale, which
changes the buttons position. Subsequent clicks are most of the time not
where the user wanted to put them.

Putting buttons on the left of the scale will keep them at a fixed
position, fixing this unfortunate behavior.

Signed-off-by: Julien Olivain <ju.o@free.fr>

### DIFF
--- a/src/FlightDisplay/FlyViewMap.qml
+++ b/src/FlightDisplay/FlyViewMap.qml
@@ -586,7 +586,7 @@ FlightMap {
         anchors.left:       parent.left
         anchors.top:        parent.top
         mapControl:         _root
-        buttonsOnLeft:      false
+        buttonsOnLeft:      true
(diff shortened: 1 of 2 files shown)


bpmooch:
@DonLakeFlyer learning the ropes here, is this a good example of ticket that should be closed?

DonLakeFlyer:
It's a good problem to fix. I swear there was a reason they were in that order. But I can't remember so might as well change. And maybe we discover later on why it was that way.

DonLakeFlyer:
And for example the Windows build should be re-run to see what is going on there.

PR #9901, opened 2021-10-05
Simple factory calibration button implemented in QGC.

batinkov:
1. In-field calibration should be made visible by default on the calibration window if calibration parameters are already loaded.
2. Otherwise, factory calibration option should be provided.
3. A smaller button for factory calibration reset should be provided on the calibration window to allow the user to reset to factory calibration parameters. (since this is HW specific, a check for the loaded FW should be done, in order to tell the user if this action is possible or not - only FMU-V5X compatible atm)

Changes in the UI:
FactoryReset


patrickelectric:
Great idea!

batinkov:
@bkueng could you review the changes above?

PR #9905, opened 2021-10-07
Fix internet access test

### LINKED ISSUE #9484: Joystick: MANUAL_CONTROL messages are not regular

patrickelectric:
Move #9591 to stable.
Fix #9484 on stable.

### DIFF
--- a/src/QtLocationPlugin/QGCTileCacheWorker.cpp
+++ b/src/QtLocationPlugin/QGCTileCacheWorker.cpp
@@ -1184,18 +1184,28 @@ QGCCacheWorker::_lookupReady(QHostInfo info)
 #else
     _hostLookupID = 0;
     if(info.error() == QHostInfo::NoError && info.addresses().size()) {
-        QTcpSocket socket;
+        auto socket = new QTcpSocket();
         QNetworkProxy tempProxy;
         tempProxy.setType(QNetworkProxy::DefaultProxy);
-        socket.setProxy(tempProxy);
-        socket.connectToHost(info.addresses().first(), 80);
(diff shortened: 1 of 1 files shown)


holden-zenith:
This looks like it was implemented

DonLakeFlyer:
Closing since this is stale

Issue #9917, opened 2021-10-13
Gstreamer Android build error


akontsevich:
## Expected Behavior
Build for Android with GStreamer

## Current Behavior
Build errors happen:

```
compiling .../qgroundcontrol/libs/qmlglsink/gst-plugins-good/ext/qt/qtitem.cc
.../qgroundcontrol/libs/qmlglsink/gst-plugins-good/ext/qt/gstqtglutility.cc:65:7: error: static_assert failed due to requirement 'sizeof *(&_debug) == sizeof(void *)' "Expression evaluates to false"
  if (g_once_init_enter (&_debug)) {
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~
.../qgroundcontrol/gstreamer-1.0-android-universal-1.18.1/x86/include/glib-2.0/glib/gthread.h:249:5: note: expanded from macro 'g_once_init_enter'
    G_STATIC_ASSERT (sizeof *(location) == sizeof (gpointer));       \
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
```

## Steps to Reproduce:
1.  Copy files like mentioned here: [master/src/VideoReceiver/README.md#android](https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoReceiver/README.md#android)
2. Build QGC for Android
3. Have above errors

## System Information
- Operating System: Ubuntu 20.04
- QGC Version: latest commit from github
- QGC build: daily

### Build log 
[buildlog.txt](https://github.com/mavlink/qgroundcontrol/files/7340792/buildlog.txt)

akontsevich:
What could be the cause of the errors, @DonLakeFlyer? GStreamer build for Qt 5.15 Android Multi-Abi changes was not updated?

PR #9920, opened 2021-10-14
Fixed GStreamer support in Android build for Qt 5.15

akontsevich:
Resolves https://github.com/mavlink/qgroundcontrol/issues/9917 issue specifically for new Qt 5.15 Multi-Abi build. GStreamer version updated to [`gstreamer-1.0-android-universal-1.18.5`](https://gstreamer.freedesktop.org/data/pkg/android/1.18.5/gstreamer-1.0-android-universal-1.18.5.tar.xz) to fix [the crash](https://github.com/mavlink/qgroundcontrol/issues/9917#issuecomment-942703344).

See more info about gstreamer latest stable version: https://gstreamer.freedesktop.org/download/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,9 +107,9 @@ install:
 
   # android dependencies: qt, gstreamer, android-ndk
   - if [ "${SPEC}" = "android-clang" ]; then
-        wget --quiet https://qgroundcontrol.s3-us-west-2.amazonaws.com/dependencies/gstreamer-1.0-android-universal-1.18.1.tar.xz &&
-        mkdir gstreamer-1.0-android-universal-1.18.1 &&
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
> Clean build works for Android arm64-v8a -- including the cameras. But Android armeabi-v7a still crashes on startup on my tablet, Galaxy S7+ with Android 11

Is the crash related to gstreamer or is it just a regular old switched to 5.15 crash? Also make sure you are up to date with very latest master since my android 5.15 fixes just went in recently.

FYI: travis.yml is no longer being used since it no longer runs, but I have a pull coming which has android CI starting to work in GitHub Actions CI. So I'll carry this over to there if this is a correct change.

DonLakeFlyer:
FYI: I'm incorporating these changes into here: https://github.com/mavlink/qgroundcontrol/pull/9916

PR #9927, opened 2021-10-16
Added BLE controller support

tadeuszk733:
Added initial support for the Gamesir T1D bluetooth low energy game controller


DonLakeFlyer:
Uh, 43 commits? Also seems to include random changes to unrelated things. Closing in hopes of another cleaner pull which makes more sense.

PR #9931, opened 2021-10-18
Patched QMake for Android, executed when custom sub-project exists.

QDroneDev:


QDroneDev:
I ran into a series of QMake failures when I tried building for Android on Windows, while using a custom sub project.  I resolved them, and then confirmed the changes worked for the same target platform built via macOS.

PR #9932, opened 2021-10-19
QMake patch for Android build in context of using custom subproject.

QDroneDev:


highfreq:
Is there a particulr reason why this fix is blocked after all checks passed?

QDroneDev:
Should I request a merge onto a different (dev?) branch perhaps?

If it were faster and easier to contribute here, I'd do so a whole lot more!  I work on a close source version of QGC daily, but am discouraged to share code as a result of the red tape (such this patch never being approved).  I assume it's simply very hard for the maintainers to keep up with all the merge requests?

DonLakeFlyer:
Sorry for taking so long. I was hoping someone else would take over majority of maintenance. That didn't happen. So for now I'm back to doing it myself. Hopefully you'll see better response now.

With regards to this pull, newer versions of OSX have similar but different problems with sed. I"m considering moving this replace scraping stuff to some other mechanism that will work everywhere. Not sure what that is yet. If I don't figure that out I'll use your mods. Stay tuned.

DonLakeFlyer:
This should just wait until we convert the custom build to cmake instead of fighting with make.

PR #9947, opened 2021-10-24
set RPATH for QGroundControl and libraries on Linux

keith-bennett-airmap:
This tries to address the issues brought up in https://github.com/mavlink/qgroundcontrol/issues/9629 and https://github.com/mavlink/qgroundcontrol/pull/9893.

- Remove the necessity of `LD_LIBRARY_PATH` in the startup script
- Remove the necessity of `QML2_IMPORT_PATH` in the startup script
- Remove the necessity of `QT_PLUGIN_PATH` in the startup script
- Fix libraries not knowing where eachother at at and causing linker failures, mixed versions, or bugs during runtime.
- Install `qt.conf` file to help Qt with finding its own stuff

Okay so each of those are quite related to each other but I'll try to explain each one. This all starts with `rpath`, or `runtime path`. That describes to a library where it can find other libraries. A system-installed library will look under (for example) `/usr/lib`. When you build a program and need a version different from the system, and that library has dependencies, you are supposed to make sure they all understand how to find each other -- and that's the purpose of `rpath`.

When you build an executable (the program or a library) you tell the linker what rpath to use for the library. On gcc or clang you use `-rpath`. That then bakes the value into the final library. If you copy that library around then it's expected that you also copied its dependencies around and maintained the same directory structure. Adding the `rpath` to QGroundControl is easy: that's what the `QMAKE_RPATHDIR` variable does. However, it _only_ updates the `rpath` for QGroundControl and specifically _does not_ update the `rpath` for dependent libraries. So removing the `LD_LIBRARY_PATH` would cause QGroundControl _directly_ to find all of its libraries but the dependencies were still not finding what they needed.

QGC would copy the Qt libraries to `Qt/libs`. First, that's different from Qt default which is `Qt/lib` [0]. One of several ways to tell Qt to use something other than its default is to use a `qt.conf` [0] file. Other documented [1] [2] [3] ways are to use the environment variables such as `QML2_IMPORT_PATH`, `QT_PLUGIN_PATH`, et al. I argue `qt.conf` is the most correct because it goes far to removing the need for environment variables and a wrapper script so that's where `qt.conf` comes in.

After adding `qt.conf`, the Qt C++ libraries under `Qt/libs` were able to resolve all of their dependencies during runtime start. However, some of the Qt imports failed. It turns out that the Qt QML libraries under `Qt/qml` don't respect `qt.conf` and/or link back to the Qt main libraries under `Qt/libs`. So their rpaths needed to be updated to fix that issue. That necessated a script to fix-up _all_ libraries under the `Qt/` directory.

And even after that there were _still_ some problems. I'd noted, while inspecting the libraries that were imported, that `libQt5WaylandClient` and `libQt5WaylandCompositor` were being imported but weren't being copied into that directory. So _most_ of the Qt libs were Qt 5.15 ... but some were Qt 5.12. And that caused further issues.

And finally, it turns out that some libraries have multiple dependencies in different paths. So it turns out that `rpath` in the executable file can contain multiple paths split by a colon (same as `PATH`). And in realizing that, I also realized that some libraries might not *have* an rpath. So the script to fix up rpath had to be adjusted to ensure that it prepended if the rpath already existed or set if not, and in doing so had to also look at what is the rpath for that lib.

And now I can run `./QGroundControl` and not have to use the startup script. I hope this fixes the issues that @julianoes had with using the script in the appimage.

Screenshot from 2021-10-23 22-37-17

[0]: `qt.conf`: [Overriding Paths](https://doc.qt.io/qt-5/qt-conf.html)
[1]: `QML_IMPORT_PATH`: [Importing QML Modules](https://doc.qt.io/qt-5/qtqml-syntax-imports.html#qml-import-path)
[2]: `QML2_IMPORT_PATH`: [QML Import Path](https://doc.qt.io/qt-5/qtqml-syntax-imports.html#qml-import-path)
[3]: `QT_PLUGIN_PATH` [The Plugin Directory](https://doc.qt.io/qt-5/deployment-plugins.html#the-plugin-directory)


julianoes:
I did some testing:

#### Ubuntu 20.04

master:
```
build/staging/QGroundControl: error while loading shared libraries: libairmap-cpp.so.2.0.0: cannot open shared object file: No such file or directory
```

This PR:
```
/home/julianoes/src/upstream/qgroundcontrol/deploy/linux-fixup-rpaths.bash: line 85: patchelf: command not found
```
And then, after installing `patchelf`:
```
(cd build && make -j8) && build/staging/QGroundControl
make: Nothing to be done for 'first'.
qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

[1]    761373 abort (core dumped)  build/staging/QGroundControl
```

With gdb:
```
Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7dd0859 in __GI_abort () at abort.c:79
#2  0x00005555556246c8 in msgHandler(QtMsgType, QMessageLogContext const&, QString const&)
    (type=<optimized out>, context=..., msg=...)
    at /home/julianoes/src/upstream/qgroundcontrol/src/QmlControls/AppMessages.cc:41
#3  0x00007ffff21ac1f0 in qt_message(QtMsgType, QMessageLogContext const&, char const*, __va_list_tag*)
    () at /home/julianoes/src/upstream/qgroundcontrol/build/staging/Qt/libs/libQt5Core.so.5
#4  0x00007ffff2194b44 in QMessageLogger::fatal(char const*, ...) const ()
    at /home/julianoes/src/upstream/qgroundcontrol/build/staging/Qt/libs/libQt5Core.so.5
#5  0x00007ffff42381b0 in QGuiApplicationPrivate::createPlatformIntegration() ()
    at /home/julianoes/src/upstream/qgroundcontrol/build/staging/Qt/libs/libQt5Gui.so.5
#6  0x00007ffff42384bd in QGuiApplicationPrivate::createEventDispatcher() ()
    at /home/julianoes/src/upstream/qgroundcontrol/build/staging/Qt/libs/libQt5Gui.so.5
#7  0x00007ffff23a84b3 in QCoreApplicationPrivate::init() ()
    at /home/julianoes/src/upstream/qgroundcontrol/build/staging/Qt/libs/libQt5Core.so.5
#8  0x00007ffff423aedc in QGuiApplicationPrivate::init() ()
    at /home/julianoes/src/upstream/qgroundcontrol/build/staging/Qt/libs/libQt5Gui.so.5
#9  0x00007ffff5086b59 in QApplicationPrivate::init() ()
    at /home/julianoes/src/upstream/qgroundcontrol/build/staging/Qt/libs/libQt5Widgets.so.5
#10 0x0000555555646346 in QGCApplication::QGCApplication(int&, char**, bool)
    (unitTesting=false, argv=0x7fffffffdc38, argc=@0x7fffffffd94c: 1, this=0x5555575dbbb0)
    at /home/julianoes/src/upstream/qgroundcontrol/src/QGCApplication.cc:186
#11 main(int, char**) (argc=<optimized out>, argv=<optimized out>)
    at /home/julianoes/src/upstream/qgroundcontrol/src/main.cc:369
```

Then, after wiping the build folder and starting fresh it ran fine.

julianoes:
Tested on Arch Linux as well and it works :tada:!

DonLakeFlyer:
Is this good to merge now?

keith-bennett-airmap:
Yes I think it is. @julianoes ?

julianoes:
Yes, all good. Do we have to document the `patchelf` build dependency somewhere?

DonLakeFlyer:
> Do we have to document the `patchelf` build dependency somewhere?

Yes, the dev-guide build instructions will need to be updated.

PR #9949, opened 2021-10-26
FactMetadata: fix bitset metadata parsing from json

### LINKED ISSUE #9854: Bitmasks are broken on Windows and Linux
When setting a parameter that is a bitmask, the decimal value is wrong. The first problem is when the parameter is default, ie EKF2_AID_MASK or SDLOG_PROFILE is 1, the selected checkboxes are wrong.

The other problem is that the checkboxes are not ORing a binary value, but ORing decimal values.

image
image
image
image
image
image

bkueng:
- use int's instead of double's
- use bitshift to match the logic in the xml

@DonLakeFlyer I did not see any internal uses of the bitmask in any json config file, so this should not affect anything else, but maybe you're aware of something?

Fixes #9854

### DIFF
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -1297,6 +1297,7 @@ FactMetaData* FactMetaData::createFromJsonObject(const QJsonObject& json, QMap<Q
 
     QStringList     rgDescriptions;
     QList<double>   rgDoubleValues;
+    QList<int>      rgIntValues;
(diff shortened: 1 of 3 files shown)


AlexKlimaj:
Is this in the daily? If so, it isn't fixed.

DonLakeFlyer:
Yes, but keep in mind daily builds are not being pushed up right now. You have to build yourself or get it from a GitHub artifact.

PR #9954, opened 2021-10-27
Debounce app reboot messages

### LINKED ISSUE #9344: Initial setup: restart popup for every unit option
## Expected Behavior
When setting up QGC for the first time it asks if the user wants to use imperial or metric units. After changing to metric (from the deafult imperial) there's a popup saying the change only has an effect after restarting QGC.

## Current Behavior
Instead of one popup there's one for each possible field probably because they correctly get changed all at the same time after the first one was switched.

## Steps to Reproduce:
1. Load the latest QGC daily.
2. Start for the first time.
3. Select metric units in the first option of the first dialogue.
4. See a lot of popups on top of each other.

## System Information
- Operating System: Manjaro 20.2.1
- QGC Version: daily:
image

DonLakeFlyer:
Fix #9344

### DIFF
--- a/src/FactSystem/Fact.cc
+++ b/src/FactSystem/Fact.cc
@@ -742,9 +742,9 @@ void Fact::_checkForRebootMessaging(void)
     if(qgcApp()) {
         if (!qgcApp()->runningUnitTests()) {
             if (vehicleRebootRequired()) {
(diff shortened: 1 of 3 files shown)


MaEtUgR:
Awesome, thanks!

PR #9964, opened 2021-10-29
docker container for building Android packages

asmirnou:
QGroundControl from the Google Play Store (even Daily Test Build) was not updated for a long time. The daily build APK for Android devices available for [download](https://docs.qgroundcontrol.com/master/en/releases/daily_builds.html#daily-builds) is not the recent version either. Here is a Docker container, which one can use to build the current version.

Most of the code was taken from [this](https://github.com/mavlink/containers) repository and adapted for the current version of Qt. The container can be used for both architectures: `armeabi-v7a` and `arm64-v8a`.

## Usage

Before using the Container, you have to build the image. You can accomplish this using docker, running the following script from the root of the QGC source code directory.

```
docker build --file ./deploy/docker/Dockerfile-build-android -t qgc-android-docker .
```

A debug keystore which is used to sign an Android app during development needs a specific alias and password combination.
To create a debug keystore, use:
```bash
keytool -genkey -v -keystore android_release.keystore -storepass android \
     -alias QGCAndroidKeyStore \
     -keypass android \
     -keyalg RSA -keysize 2048 -validity 10000
```
Replace the original "android/android_release.keystore" file in QGC source code directory, then use the Container to build QGC:

```bash
mkdir build
docker run --rm -v ${PWD}:/project/source -v ${PWD}/build:/project/build \
     -e ANDROID_KEYSTORE_PASSWORD=android -e ANDROID_TARGET_ARCH=arm64-v8a \
     qgc-android-docker
```

To build APK for another architecture, remove `-e ANDROID_TARGET_ARCH=arm64-v8a` option.

## Notes

The base image can be specified as follows: `--build-arg BASE_IMAGE=ubuntu:18.04`. This is useful for Linux-based container if one needs to build QGC for OS, where newer version of `GLIBC_2.29' is not available.


DonLakeFlyer:
> Do you have time to help us create a Github Action that uses this image to build android on pull request?

The goal would be to modify the existing android action to use docker instead

asmirnou:
If it gets outdated, then along with the Linux container, which is already in repository. The argument doesn't look like a blocker to me.

Unfortunately, I have other commitments at present and do not have much time to help.

Issue #9022, opened 2020-08-24
Loss of MAVLink transmission when enabling joystick


MaEtUgR:
## Expected Behavior
Gamepad style joystick when calibrated and activated provides manual control stick input to the autopilot and doesn't interfere with other functionality.

## Current Behavior
When I have a Joystick (Microsoft XBox 360 controller I'm always using) connected, calibrated (tried recalibration) and activated (doesn't happen when the gamepad is connected but the checkbox is off since QGC start) that MAVLink transmission stops working. As soon as I'm once in this state no command gets sent anymore (I tried mode switch, parameter request, reboot, arm), MAVLink shell doesn't react anymore and when I start QGC in this state the "Vehicle 1 did not react to parameter request" message gets shown. Only solved by unplugging or disabling the joystick device and restarting QGC.

- The onscreen virtual joystick does not suffer from this

## Steps to Reproduce:
1. Plug USB gamepad that was not configured in QGC before
2. Start QGC
3. Plug pixhawk autopilot to computer
4. See autopilot connected and everything (MAVLink commands) working
5. Calibrate and enable joystick
6. Get stuck in a state where no commands are sent to the autopilot anymore, telemetry still shows

## System Information
- Operating System: Verified the same behavior on Windows 10 and Manjaro Linux VM
- QGC Version:
![image](https://user-images.githubusercontent.com/4668506/91042241-b8f33300-e611-11ea-87e2-aa35a1383684.png)

- QGC build: Downloaded from https://docs.qgroundcontrol.com/master/en/releases/daily_builds.html
- Flight Controller: Pixhawk 4 mini
- Autopilot (with version): PX4 latest master (https://github.com/PX4/Firmware/commit/979243f38f9c4fad14ec47b622d7b95f92dafe21)

## Misc Context
I was testing calibration and noticed I have manual control lost at the end of every calibration: https://github.com/PX4/Firmware/pull/15235#issuecomment-678353384 and wanted to check if with a real instead of virtual joystick I'll have the same problem.

MaEtUgR:
I checked again on latest QGC daily:
![image](https://user-images.githubusercontent.com/4668506/95754838-e5d3d600-0ca3-11eb-9782-5daea6cb656f.png)
And it still has the issue.

DonLakeFlyer:
Does that happen on SITL as well?

BrummbQ:
I can confirm this issue. How can I test it on SITL?

[...]

PR #9976, opened 2021-11-02
Force write bytes to happen on correct thread

DonLakeFlyer:
* Fix for #9022
* The #8835 change was incorrect in that it caused the call to QIODevice::write to possibly happen on a different thread than the link itself was on. This worked for some link types and not for others.
* In order to fix this I moved back to signaling the write bytes to the link. This will cause them to be queued and show up on the correct thread if the threads differ.


src/PositionManager/PositionManager.cpp
```diff
@@ -39,7 +45,34 @@ void QGCPositionManager::setToolbox(QGCToolbox *toolbox)
 
    if (!_defaultSource) {
        //-- Otherwise, create a default one
+#if 0
```
keith-bennett-airmap, line 48: `#if 0` is an ugly wart

src/comm/LinkInterface.cc
```diff
@@ -80,10 +82,7 @@ void LinkInterface::_freeMavlinkChannel()
 
 void LinkInterface::writeBytesThreadSafe(const char *bytes, int length)
 {
-    QByteArray byteArray(bytes, length);
-    _writeBytesMutex.lock();
-    _writeBytes(byteArray);
-    _writeBytesMutex.unlock();
+    emit _invokeWriteBytes(QByteArray(bytes, length));
```
keith-bennett-airmap, line 85: From [documentation](https://doc.qt.io/qt-5/qbytearray.html#QByteArray-1):
> QByteArray makes a deep copy of the string data.

:+1:

keith-bennett-airmap (commented):
It looks like the changes to to `src/comm/LinkInterface.h/cc` are the only changes needed. The other changes look like they're unrelated

DonLakeFlyer:
This has two commits in it. The second one is to make QT_FATAL_WARNING usable again. They were required for me to be able to debug this problem. They are not required to fix the joystick problem. They are just required to make this type of debugging possible again.

DonLakeFlyer:
I'll squash the android change together.

DonLakeFlyer:
The #if 0 is to leave the old code in, both to explain the Qt strangeness and also under the hope that at some point Qt fixes this and we can just move back to that simpler direct call.

PR #9977, opened 2021-11-02
MavlinkConsoleController: add target sysid/compid

julianoes:
The SERIAL_CONSOLE message now includes a target_system and target_component which prevents serial messages being sent to the wrong vehicle in case there are multiple vehicles connected.

More background:
https://github.com/mavlink/mavlink/pull/1725


julianoes:
I tested this against PX4 and it still works.

PR #9984, opened 2021-11-03
Joystick: Introduce custom MAV commands

valbok:
In case when you would like to bind some MAV commands
to some joystick's buttons, you can now create a json file
which will be parsed at runtime.

It is usefull when you need to send a custom command from the joystick.
But this command is not supported by the firmware, or not relevant to be included to QGC.

Example:
Create a file JoystickMavCommands.json and place it with a binary:
"commands": [
        {
            "id":                   31010,
            "name":                 "Enable Autonomy",
            "param1":               1.0
        }
]

After executing the QGC, you will be able to bind "Enable Autonomy" to a button,
by clicking to the button, MAV_CMD_USER_1 (31010) with param1 1.0 will be sent.

If there is no json file, nothing will happen.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -583,6 +583,7 @@ HEADERS += \
     src/FollowMe/FollowMe.h \
     src/Joystick/Joystick.h \
     src/Joystick/JoystickManager.h \
+    src/Joystick/JoystickMavCommand.h \
     src/JsonHelper.h \
     src/KMLDomDocument.h \
     src/KMLHelper.h \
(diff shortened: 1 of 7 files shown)


DonLakeFlyer:
I'm not very up to date on how these button actions things works. Can you include a screen shot of the joystick config UI where these new actions show up?

valbok:
Screenshot_2021-11-23_11-49-48
Screenshot_2021-11-23_11-49-02

valbok:
If stored settings have unknown/not found action, it falls back to No Action.

DonLakeFlyer:
Thanks, look great.

PR #9994, opened 2021-11-06
Support for various languages

muramura:
Change untranslatable text.


booo:
Can you please use a more descriptive commit message and add one or two sentences to the body of the commmit.

Otherwise it looks good to me.

Issue #9995, opened 2021-11-07
The UnitsSettings choices do not provide translation resources.


bols-blue:
Tell us a bit about the feature:
- What problem does it solve?
It will be automatically detected by linguist and added to the translation resource.
sorce code comment "units settings can't be loaded from json since it creates an infinite loop of meta data loading."

- What does it do?
translation

- What flight stacks must it work with? (All, PX4, ArduPilot)
All

- Relevant vehicle types? (All, multirotor, fixed-wing, VTOL, submarine, etc.)
All

Any additional context you can provide will make the feature easier to evaluate (e.g. mockups, detailed specification, etc.)

![Screenshot from 2021-11-06 09-59-39](https://user-images.githubusercontent.com/184669/140640467-9304a343-692e-4244-aba7-674056f6686b.png)

Fixed translation resources and patching locally
![Screenshot from 2021-11-07 18-44-34](https://user-images.githubusercontent.com/184669/140640436-c299cc7e-fcbf-45f5-96f5-fd9e2e46d5eb.png)

Issue #9962, opened 2021-10-28
Instructions: Completing move and translations to master


DonLakeFlyer:
Over a year ago the localization process for QGC was updated as follows:
* Branch support for stable/daily releases
* Support for translating json files
* Automatic updates of translated strings submitted as pull request to QGC from Crowdin

![Screen Shot 2021-10-28 at 11 44 48 AM](https://user-images.githubusercontent.com/5876851/139316633-a1723ab4-91d3-41c5-8bfd-766e2ebd7aec.png)

The new structure can be seen above:
* The master/translations hierarchy is where new translations need to happen
* The Stable 4.0 is the previous translation set for which there should no longer be work done against

This new structure changed the process in a few significant ways:
* The automatic updates don't happen until the translations are Approved on Crowdin
* The change in structure broke the connection of the previous set of strings in Stable 4.0 to the new master/translations location. The screenshot above for Turkish shows the problem. Stable shows almost complete, but the new location shows very little translation.

How to fix this
Since Crowdin supports Translation Memory it's a fairly mechanical process to pick the translations which previously existed. In the following screenshot you can see an exact match which you would select, Save and Approve.

![Screen Shot 2021-10-28 at 11 54 44 AM](https://user-images.githubusercontent.com/5876851/139318052-987bee30-faf3-407c-9f9d-4f93c1fbdcb2.png)

I know a PITA. But hopefully just a one time thing. Thanks everyone for helping out.

FYI: Since Chinese and Korean were previously fully translated I was able to mechanically use Crowdin to filter previous submitted translations which were a year old assuming they were correct and approved them. So Chinese/Korean should be done I hope.

booo:
The idea is to create proper documentation about the process?

DonLakeFlyer:
Essentially yes.

booo:
I started to transfer some of the German Translations but it's really a shitty job.

Do you think it's possible to get some funding for these tasks and out source these to proper translators? I see two benefits: We don't take away valuable developer time from the project and someone with proper language skills creates a good translation.

DonLakeFlyer:
> Do you think it's possible to get some funding for these tasks and out source these to proper translators? I see two benefits: We don't take away valuable developer time from the project and someone with proper language skills creates a good translation.

That would be nice. But so far I haven't been able to get much help with anything at all.

PR #9996, opened 2021-11-07
Change raw string to tr

bols-blue:
please show #9995

### DIFF
--- a/src/Settings/UnitsSettings.cc
+++ b/src/Settings/UnitsSettings.cc
@@ -23,7 +23,7 @@ DECLARE_SETTINGSFACT_NO_FUNC(UnitsSettings, horizontalDistanceUnits)
         // Distance/Area/Speed units settings can't be loaded from json since it creates an infinite loop of meta data loading.
         QStringList     enumStrings;
         QVariantList    enumValues;
-        enumStrings << "Feet" << "Meters";
+        enumStrings << QObject::tr("Feet") << QObject::tr("Meters");
         enumValues << QVariant::fromValue(static_cast<uint32_t>(HorizontalDistanceUnitsFeet))
                    << QVariant::fromValue(static_cast<uint32_t>(HorizontalDistanceUnitsMeters));
         FactMetaData* metaData = new FactMetaData(FactMetaData::valueTypeUint32, this);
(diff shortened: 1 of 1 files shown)


src/Settings/UnitsSettings.cc
```diff
@@ -23,7 +23,7 @@ DECLARE_SETTINGSFACT_NO_FUNC(UnitsSettings, horizontalDistanceUnits)
         // Distance/Area/Speed units settings can't be loaded from json since it creates an infinite loop of meta data loading.
         QStringList     enumStrings;
         QVariantList    enumValues;
-        enumStrings << "Feet" << "Meters";
+        enumStrings << QObject::tr("Feet") << QObject::tr("Meters");
```
muramura, line 26: In this file, "tr(" is used.
Is there any reason to clarify the "QObject::tr(" class in this process?
I thought it would be better to combine it with other processes.
Also, there are only 5 CPP language files that use this notation.

booo:
Please try to describe why the change from raw string to QObject::tr is necessary and add it to your commit message. Thanks.

bols-blue:
@booo Thank you for comment.
I tried to rewrite it, but how about it?

bols-blue:
I have a question.
Is it better to include translation resources in a pull request?

booo:
Thanks.

If you want to add a translation or improve one please use https://crowdin.com/project/qgroundcontrol

We are still working on the documentation: https://github.com/mavlink/qgroundcontrol/issues/9962

DonLakeFlyer:
Why can't these all be `tr` instead of `QObject::tr`?

bols-blue:
I used QObject :: tr because it was displayed ”tr not defined.”
when I check it now, it seems that Unitsetting :: tr is used as tr, so I fixed it.

muramura:
@bols-blue san
I am no longer detectable by QT's LUPDATE with the current support.
Is there a statement in the LUPDATE of your QT?

bols-blue:
@muramura san
I know that. Perhaps the linguist is not evaluating the macro correctly.
I was looking for a setting. Priority is given to undoing and closing.

muramura:
This statement "tr("Weight units")" is assigned to the UnitsSettings class.
I thought that the others could be assigned to the UnitsSettings class as well.
I used this specification to display in the LUPDATE.
I also set the translation settings to display the unit settings in Japanese.

DonLakeFlyer:
> I used QObject::tr because it was displayed ”tr not defined.”
> when I check it now, it seems that Unitsetting::tr is used as tr, so I fixed it.

I'm a bit confused as to where this currently stands. I tried doing a compile just using tr("...") and it compiled fine for me. Are you saying that isn't the problem. The problem is that lupdate doesn't find the strings and output the correctly updated .ts files? If so does this latest change bot compile and work with lupdate?

mrpollo:
is this PR ready to come in? I can't tell by the discussion, and the changeset looks good to me

muramura:
I am saying that this PR change was not discovered by the ludate command.
I am hoping the change will be found in the lupdate command.

bols-blue:
"tr" cannot be lupdated.
So "QObject::tr" is used in this PR.
I would like you to change "QObject::tr" to "tr" in another issue and merge it.

PR #10022, opened 2021-11-16
Vehicle: fix potential out-of-bound access when receiving SERIAL_CONTROL msg

bkueng:


booo:
Could you add a explanation to the commit body? There is plenty of text in the corresponding ticket.

bkueng:
> Could you add a explanation to the commit body? There is plenty of text in the corresponding ticket.

I did not because imo the code change makes this obvious already.

Issue #9946, opened 2021-10-22
Remaining Github Actions based CI work


DonLakeFlyer:
First lets start with the current state of affairs:
* There is no longer fully automated support for uploading any builds whether stable or daily to S3 and/or Google Play. This all fell apart when Travis crapped out.
* Daily builds which are referenced by the docs are out of date my multiple months
* The Google Play daily build in the store hasn't been updated in close to a year
* GitHub Actions have been created for Linux/Mac/Windows which verify build and test on all pull requests
* Android GitHub Actions only compiles to APK as a first step towards switching to Android App Bundles
* Android build has been updated for Qt 5.15 as well as new scheme for daily build version number generation as step towards end goal of app bundle

Remaining work:
* [x] Although the GitHub Actions are setup to upload to S3 the action for this is currently disabled. The reason is that it doesn't work. Someone with both S3 and GitHub knowledge/access will need to figure out how to make this work.
* [ ] The Android build needs to be finished up with respect to generating an Android App Bundle with both 32 and 64 bit versions in it. It close, but not fully there. It fails on obscure Gradle errors.
* [ ] Run GitHub Actions when a new release is created and upload installers to S3

booo:
Sorry, I'm not super familiar with QGC and iOS. Do we need a workflow for the iOS build too?

DonLakeFlyer:
> Sorry, I'm not super familiar with QGC and iOS. Do we need a workflow for the iOS build too?

Yes we do. We've in fact never had iOS in CI.

bols-blue:
How about changing the daily build documentation?
Users can download it from github without having to store their daily builds in S3.

MaEtUgR:
I just stumbled on this trying to download the latest versions from https://docs.qgroundcontrol.com/master/en/releases/daily_builds.html
For Windows I get a version from 25.09.2021 and for Linux there's just a 403 forbidden.

> Users can download it from github without having to store their daily builds in S3.

As far as I know there are no daily builds stored on GitHub so if a user is not building locally (like e.g. me) he can't get the latest version. Please tell me if I'm wrong.

mrpollo:
@MaEtUgR you can download the artifacts out of any build, that's how I'm getting my daily builds right now, go to this link https://github.com/mavlink/qgroundcontrol/actions and look for the "Windows Release" tab, and click on the latest successful build you will be able to download the artifacts

DonLakeFlyer:
[...]

PR #10025, opened 2021-11-16
CI: Upload stable release artifacts

mrpollo:
fixes one of the todos on #9946 for uploading stable releases to S3 with GitHub Actions

### DIFF
--- a/.github/workflows/linux_release.yml
+++ b/.github/workflows/linux_release.yml
@@ -1,6 +1,8 @@
 name: Linux Release
 
 on:
+  release:
+    types: [released]
   push:
     branches:
     - 'master'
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
That's not how releases are done in QGC though. They are based on tags which I think are different from a github action standpoint. I'll try to find some time to document the releases process int he dev guide such that someone could automate it.

mrpollo:
Tags are fine, we can manually convert a tag to a release, which should work with this change, or I can change this PR to watch for the created tags event, and regex for semver.

mrpollo:
@DonLakeFlyer, I just realized what you are saying is that we don't make "release builds" like this, and we need a separate release build workflow.

I will wait for your documented instructions and work on a separate CI workflow just for stable releases

DonLakeFlyer:
Basically the stable builds need to kick off based on a "v*" matching tag. And then they need to push to the "latest" directory on S3 which you can see the url structure from the doc links.

DonLakeFlyer:
Also the CONFIG+=DailyBuild needs to be remove when it's a stable build.

PR #10031, opened 2021-11-21
[Hold for Flatpak] QMake: Don't override AIRMAP_PLATFORM_SDK_PATH

thopiekar:
In case you want to use a different SDK installation from a custom path, you can set AIRMAP_PLATFORM_SDK_PATH to whatever you want.
However, there is no check whether it is already defined and therefore it will get overwritten in any case.

Let's avoid it!

### DIFF
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -285,7 +285,9 @@ contains (DEFINES, DISABLE_AIRMAP) {
 } else:exists(user_config.pri):infile(user_config.pri, DEFINES, DISABLE_AIRMAP) {
     message("Skipping support for AirMap (manual override from user_config.pri)")
 } else {
-    AIRMAP_PLATFORM_SDK_PATH    = $${OUT_PWD}/libs/airmap-platform-sdk
+    !defined(AIRMAP_PLATFORM_SDK_PATH, var) {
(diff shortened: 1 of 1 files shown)


thopiekar:
I'm pretty sure that I made a PR with this in the past, but looks like I missed to cherry-pick it into my last bundle.
Still need to practically test it since I'm not a QMake expert.

thopiekar:
Works as expected. Ready to be merged! :smile:

DonLakeFlyer:
I'm not quite sure this make sense from a general concept. This allows you to override one specific build setting. Whereas there are hundreds of other ones which you can't override. Like say where QGC expects to find gstreamer files. Not sure why this one is somehow more special than any of the other settings in the build files? Normally if you need some special thing like this you just modify the build files for your temporary use case. Why is this so important there need to be permanent support for this?

thopiekar:
Pew, I remember that we had this conversation before and @jennerl was fine with it.
We can put this on hold as it is part of the Flatpak proof-of-concept still and not really needed on your production side to be honest.

Actually, there are bigger problems to get the Flatpak accepted, but I wanted to share this before it gets lost at least.

DonLakeFlyer:
Airmap is gone now

PR #10051, opened 2021-12-01
Joystick: fix visualization

### LINKED ISSUE #10043: Joystick calibration does not seem to stick
## Expected Behavior
I want to use my Logitech Extreme 3D Pro joystick. I expect that I can calibrate it and then use it.

## Current Behavior
I do the calibration and it works correctly, however, afterwards it doesn't look like I can actually use it.

Selection_053

Here I press Next, and now it should all be saved.

So back at this screen, I would assume the axes show up, however, only the buttons work, the axes don't move.

Selection_054

When I try to calibrate PX4 to use RC, it tells me that it needs at least 4 channels, so it supposedly gets zero.

At the end of the calibration, I see the log:
```
JoystickLog: Invalid axis index -1
JoystickLog: Invalid axis index -1
```

## Steps to Reproduce:
As above.

## System Information
- QGC Version: master (7d960a878d3b2ee5731c9ea337f9838f87d9ac8b)
- Flight Controller: Pixhawk 4
- Autopilot (with version): PX4 master

julianoes:
It turns out this signal was not called anywhere.

Fixes #10043.

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -676,6 +676,8 @@ void Joystick::_handleAxis()
                     buttonPressedBits |= buttonBit;
                 }
             }
+            emit axisValues(roll, pitch, yaw, throttle);
+
             uint16_t shortButtons = static_cast<uint16_t>(buttonPressedBits & 0xFFFF);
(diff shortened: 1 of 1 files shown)


julianoes:
Not sure what is wrong with the Android build.

PR #10067, opened 2021-12-08
Add support for Bluetooth low energy joysticks

funsafe-math:
Currently only one the Gamesir T1d is supported, 
The changes add a menu option for joysticks
image
Where  you can discover and connect to your joystick
image


DonLakeFlyer:
If this is to be added it should integrate into the existing joystick ui not add a whole new joystick settings ui to QGC.

Issue #10069, opened 2021-12-10
ArduPilot: Tuning page broken by recent changes


DonLakeFlyer:
Looks like the `params` property is no longer available which causes the tuning page for ArduCopter (maybe others as well) to fail.

DonLakeFlyer:
This can be tested with ArduPilot MockLinks. Although if you are going to be modifying mainline generic QGC code would be good to understand how to test with ArduPilot SITL as well. It's pretty easy to set up.

PR #10072, opened 2021-12-13
ArduPilot: fix Tuning page

bkueng:
Tested against APM SITL:
image

Fixes https://github.com/mavlink/qgroundcontrol/issues/10069

And for PX4: save & restore autotuning checkbox when switching between tabs


DonLakeFlyer:
Thanks @bkueng

PR #10085, opened 2021-12-17
Add option to update home position

julled:
We have the usecase of flying a fixed wing from a (moving) ship. Thus we have the problem of having a moving homepoint over time.

In APM its possible to update the homeposition via MAV_CMD_DO_SET_HOME. After the command is sent, the drone uses this  new HOMEPOINT. 

This PR uses a internal or external GPS source (from the positionmanager) and sends the new home point as MAV_CMD_DO_SET_HOME message if the QGCPositionManager sends a gcsPositionChanged signal. 

I tested these changes using the SimulatedPosition.cc from the QGC positionmanager and the APM simulator with the latest master (where this commit is included https://github.com/ArduPilot/ardupilot/commit/1d5ea351c41c0af2118e57a80e5727a1ed77b4aa - it fixes homepoint setting while already in RTL) and it works like a charm.  See video https://cloud.hs-augsburg.de/s/kfJfjFjLPAakoG2

I added a setting for this feature in the options.


julled:
This feature is somehow similar to the "movinghome" plugin of mavproxy.

https://github.com/ArduPilot/MAVProxy/blob/master/MAVProxy/modules/mavproxy_movinghome.py

src/Settings/FlyView.SettingsGroup.json
```diff
@@ -66,6 +66,12 @@
     "units":            "m",
     "default":     1000,
     "min":              1
+},
+{
+    "name":             "updateHomePosition",
+    "shortDesc": "Update GCS' homeposition coordinates in case of change on Autopilot",
```
booo, line 72: Should be "home position" not homeposition.

The description is a bit confusing. Maybe it should be "Send updated GCS' home position to autopilot in case of change of home position."

src/Vehicle/Vehicle.cc
```diff
@@ -3660,6 +3661,23 @@ void Vehicle::_updateDistanceToGCS()
     }
 }
 
+void Vehicle::_updateHomepoint()
+{
+    if(_toolbox->settingsManager()->flyViewSettings()->updateHomePosition()->rawValue().toBool()
```
booo, line 3666: Maybe saving these values in variables such as updateHomePosition and supportsSetHome helps with readability.

booo (approved):

booo:
This is a feature we missed while testing our UAS for SAR at sea. Very important for our work at sea. Would appreciate a merge a lot.

src/Vehicle/Vehicle.cc
julled: see commit https://github.com/mavlink/qgroundcontrol/pull/10085/commits/a743b4ca277e862b89333743b50e40e8647a9444

src/Settings/FlyView.SettingsGroup.json
julled: see commit https://github.com/mavlink/qgroundcontrol/pull/10085/commits/776ada0bef6f91345b4e5d7368137c9e94be5a54

mrpollo:
@bkueng do you think this will affect PX4 users?

kpfmoi23ipm:
> @bkueng do you think this will affect PX4 users?

according to https://docs.px4.io/master/en/flight_modes/mission.html the command is supported by px4.

bkueng:
Yes it's supported and should be added here as well: https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc#L268

booo:
Is there something left to do we can do to get this pull request merged?

bkueng:
Yes please add the command to https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc#L268 as well.

julled:
I could add, but got no chance to test it for PX4 aswell. Can one of you do this ?

RomanBapst:
@julled You can ping me if it's ready for testing

julled:
@RomanBapst i have added the command and it should be now possible to test this for PX4.

RomanBapst:
@julled Actually, can you tell me how to simulate different gcs positions? I never did that.

julled:
Easiest hardware soultion would be to use a Tablet with GPS and QGC and walk around. Or you could feed GPS positions via a USB GPS to a laptop based QGC.

If you actually want to simulate GPS changes within QGC:
* Activate Simulation via https://github.com/mavlink/qgroundcontrol/blob/master/src/PositionManager/PositionManager.cpp#L82
* Implement the wanted simulated movement in https://github.com/mavlink/qgroundcontrol/blob/master/src/PositionManager/SimulatedPosition.cc#L61 (e.g. a circle)

RomanBapst:
@julled Tested with PX4 and it works just fine, thanks! The home position moves with the GCS position on the map and internally the home position is updated.
image

Might be something unrelated to this PR but I only seem to be able to access the general settings once. If I return to the flyview and try a second time clicking on the blue QGC icon then nothing happens. Maybe it's unrelated and we have to debug it separately.

RomanBapst:
@julled Never mind about the app settings not working, this was already fixed before.

RomanBapst (approved):

PR #10094, opened 2021-12-25
QGCCameraControl: allow to read camera information xml via ftp

dayjaby:
This PR heavily reuses the MAVLink FTP capabilities already present in QGroundControl and allows to specify camera definition URLs in `CAMERA_INFORMATION` with the mftp prefix, like `mftp://[;comp=100]infos/camera_info.xml`.

Tested it against the MAVSDK camera manager example PR: https://github.com/mavlink/MAVSDK/pull/1655

TODO:

- re-use component ID from CAMERA_INFORMATION as default for the ftp communication, as this would allow URLs like `mftp://infos/camera_info.xml`
- test / handle communication aborts properly

### DIFF
--- a/src/Camera/QGCCameraControl.cc
+++ b/src/Camera/QGCCameraControl.cc
@@ -2052,6 +2052,7 @@ void QGCCameraControl::_ftpDownloadComplete(const QString& fileName, const QStri
         return;
     }
 
+    _cached = true;
     QByteArray bytes = xmlFile.readAll();
     emit dataReady(bytes);
 }
--- a/src/Camera/QGCCameraControl.cc
+++ b/src/Camera/QGCCameraControl.cc
@@ -11,6 +11,7 @@
 #include "VideoManager.h"
 #include "QGCMapEngine.h"
 #include "QGCCameraManager.h"
+#include "FTPManager.h"
 
 #include <QDir>
 #include <QStandardPaths>
@@ -1950,6 +1951,21 @@ QGCCameraControl::_handleDefinitionFile(const QString &url)
 {
     //-- First check and see if we have it cached
     QFile xmlFile(_cacheFile);
+
+    QString ftpPrefix(QStringLiteral("%1://").arg(FTPManager::mavlinkFTPScheme));
+    if (url.startsWith(ftpPrefix, Qt::CaseInsensitive)) {
+        int ver = static_cast<int>(_info.cam_definition_version);
+        QString fileName = QString::asprintf("%s_%s_%03d.xml",
+            _vendor.toStdString().c_str(),
+            _modelName.toStdString().c_str(),
+            ver);
+        connect(_vehicle->ftpManager(), &FTPManager::downloadComplete, this, &QGCCameraControl::_ftpDownloadComplete);
+        _vehicle->ftpManager()->download(url,
+            qgcApp()->toolbox()->settingsManager()->appSettings()->parameterSavePath().toStdString().c_str(),
+            fileName);
+        return;
+    }
+
     if (!xmlFile.exists()) {
         qCDebug(CameraControlLog) << "No camera definition file cached";
         _httpRequest(url);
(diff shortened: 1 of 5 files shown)


src/Camera/QGCCameraControl.cc
```diff
@@ -2020,6 +2036,26 @@ QGCCameraControl::_downloadFinished()
     //reply->deleteLater();
 }
 
+void QGCCameraControl::_ftpDownloadComplete(const QString& fileName, const QString& errorMsg)
+{
+    qCDebug(CameraControlLog) << "QGCCameraControl::_ftpDownloadComplete fileName:errorMsg" << fileName << errorMsg;
```
patrickelectric, line 2041: You don't need to set class and function name, that can be configured  via environment variable:
https://doc.qt.io/qt-5/debug.html

src/Vehicle/FTPManager.cc
```diff
@@ -72,7 +72,11 @@ bool FTPManager::download(const QString& fromURI, const QString& toDir)
     }
     lastDirSlashIndex++; // move past slash
 
-    _downloadState.fileName = _downloadState.fullPathOnVehicle.right(_downloadState.fullPathOnVehicle.size() - lastDirSlashIndex);
+    if (fileName.size() == 0) {
```
patrickelectric, line 75: `fileName.isEmpty()`

src/Vehicle/FTPManager.cc
```diff
@@ -641,6 +645,8 @@ bool FTPManager::_parseURI(const QString& uri, QString& parsedURI, uint8_t& comp
         if (!ok) {
             qCWarning(FTPManagerLog) << "Incorrect format for component id" << uri;
             return false;
+        } else {
```
patrickelectric, line 648: Dont need else after return.

src/Vehicle/FTPManager.cc
```diff
@@ -641,6 +645,8 @@ bool FTPManager::_parseURI(const QString& uri, QString& parsedURI, uint8_t& comp
...
+            qCDebug(FTPManagerLog) << "Found compId:" << (int)compId;
```
patrickelectric, line 649: There is no need to cast it to int.

patrickelectric (changes requested):

dayjaby:
@patrickelectric Thanks for your review. I applied your suggested changes.

patrickelectric (approved):
The code appears to be correct

mrpollo:
thanks for the contribution @dayjaby

src/Camera/QGCCameraControl.cc
```diff
@@ -1950,6 +1951,21 @@ QGCCameraControl::_handleDefinitionFile(const QString &url)
 {
     //-- First check and see if we have it cached
     QFile xmlFile(_cacheFile);
+
+    QString ftpPrefix(QStringLiteral("%1://").arg(FTPManager::mavlinkFTPScheme));
+    if (url.startsWith(ftpPrefix, Qt::CaseInsensitive)) {
+        int ver = static_cast<int>(_info.cam_definition_version);
+        QString fileName = QString::asprintf("%s_%s_%03d.xml",
```
mrpollo, line 1958: hey @dayjaby we have a question, we noticed you are saving the file here and then reading the byte array before submitting to `dataReady` we are concerned we are saving the file twice to disk, once when ftp manager downloads, and twice when the caching detection happens, is this why you are using the same filename?

mrpollo (changes requested):

src/Camera/QGCCameraControl.cc
dayjaby: If I remember correctly the file got only transmitted once. But let me test that again.

Why should the cache detection save the file? The FTP manager saves it once, and from that moment onwards QGC should reuse the cached file.

mrpollo: in http "downloads" we are saving the camera definition directly as a byte array, whereas on ftp you are saving it directly as a file then converting it to byte array, im wondering if you can either save as byte array directly, or double check the cache logic that happens after you signal data ready won't create a duplicate file locally

dayjaby: You are right. Didn't expect "loadCameraDefinitionFile" to **save** a file.

With the newest commit, after successful ftp download, it just pretend the xml file is cached to prevent a second file save.

PR #10096, opened 2021-12-28
New QGCPopupDialog rework

DonLakeFlyer:
* This makes it simpler to use and more closely match the existing Qml PopupDialog
* Getting rid of QGCViewDialog right hand dialogs I've converted a few in this pull. I'll convert the rest in a subsequent pull.
* I added QGCPopupDialog a while back as a means to move away from the right hand dialogs to a more standard popup dialog. Just getting around to moving things over now.


DonLakeFlyer:
I'm going to merge this so I can proceed with the next step

PR #10097, opened 2021-12-28
Fix crash in QGC and report configuration errors to the configuring pilot

keith-bennett-airmap:
Fix crash in QGC and report configuration errors to the configuring pilot

For BOSS-899, https://github.com/mavlink/qgroundcontrol/issues/10035

- Fix crash in QGC:
    Application key wasn't set and so the AirMap shared state object
    hadn't been created.

- Report configuration errors to the configuring pilot:
    When attempting to load the flight list, the application key and
    pilot's username/password must be set.


bkueng (approved):
Makes sense

Issue #10070, opened 2021-12-11
Map view very slow


aib:
The main map view is very slow (<1 - 10 FPS) to respond to scrolling and zooming. I have a decent computer running the latest Ubuntu.

The rest of the application (i.e. Application Setup) seems fine but I can't be sure as there is UI lag of around a few hundreds of milliseconds.

## Expected Behavior
On my friend's Mac it runs fine. I would guess at least 30-60 FPS.

## Current Behavior
As stated above.

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1. Download v4.1.4 AppImage
2. `rm -rf ~/QGroundControl* ~/.config/QGroundControl*`
3. Run the app
4. Click "OK" twice
5. Zoom and/or scroll

## System Information
When posting bug reports, include the following information
- Operating System: Ubuntu 21.10
- QGC Version: 4.1.4 (AppImage prints "v4.1.3")
- QGC build: GitHub releases
- Flight Controller: none
- Autopilot (with version): none

## Detailed Description
Also tried the latest `master` (Docker build), same behavior.

## Log Files and Screenshots
- [QGC Console Logs](https://docs.qgroundcontrol.com/en/Support/Support.html#reporting-bugs)
[console.log](https://github.com/mavlink/qgroundcontrol/files/7696829/console.log)
- Autopilot logs when available (post a link).
N/A
- Screenshots of QGC to help identify the current issue/bug behavior.
N/A

I could actually use some pointers for debugging this behavior in the code. Could someone point me to the most likely culprit, or at least the map widget and/or a render loop, if applicable?

adk4davsys:
Hello, I am facing the same issue with the latest appimage, have you found any solution to this issue? My map view is extremely slow to input and virtual joystick keeps giving me manual control lost error as well.

aib:
Not yet. Haven't made much progress, either. Currently lost in the Qt code.

adk4davsys:
Thanks for your reply. I also noticed when I am doing HITL with Gazebo, even opening QGC is slow. I am just going to revert back to previous version where I did not have this stuttering issue. Hope you'll figure out.

aib:
You are right, if I go back far enough the issue disappears. 4.0.11 is good whereas 4.1.4 is bad. That should give me the two commits to start dissecting. Thanks!

aib:
[...]

PR #10101, opened 2022-01-04
Save map position using a timer to prevent FPS decrease (#10070)

aib:
Based on commit d0008a5f307086227197bff1ba72e6ff1090eeab. See #9220.

Added if clauses so prolonged repeated changes don't prevent saves. Fixes the issue for me.


akontsevich (approved):
Should work fine.

holden-zenith (approved):
Tested with debug output and works great. No need to write any faster

mrpollo:
@DonLakeFlyer I'm not too keen on this fix, if the current functionality has some issues with performance on slower systems, would be best if we removed the change detection save feature entirely and replace it with a manual save after done?

mrpollo:
thanks @DonLakeFlyer 💯 

thanks for the fix @aib

PR #10103, opened 2022-01-05
More work on QGCPopupDialog, and QGCSimpleMessageDialog usage

DonLakeFlyer:
* This is the second phase of the work happening in this area
* Even more work done to make QGCPopupDialog instead of "sort of" work like a standard Qml Dialog it works exactly like it now
* Converted all usage of QGCViewMessage to either `mainWindow.showMessage` or direct `QGCSimpleMessageDialog` usage
* Converted a few more QGCViewDialog usages to the new QGCPopupDialog
* Added new support to `mainWindow.showMessage` to make it way easier to write code which needs to do something based on a simple yes/no style questions. It now has two additional optional parameters to specify buttons and accept function all within the showMessage function call. Example:
```
mainWindow.showMessage(
    "title", 
    "text", 
    StandardButton.Yes | StandardButton.No,
    function() { doAcceptStuff here } )
```

Remaining work:
* Convert remaining usage of QGCViewDialog to QGCPopupDialog

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -164,7 +164,6 @@
         <file alias="QGroundControl/Controls/QGCToolInsets.qml">src/QmlControls/QGCToolInsets.qml</file>
         <file alias="QGroundControl/Controls/QGCViewDialog.qml">src/QmlControls/QGCViewDialog.qml</file>
         <file alias="QGroundControl/Controls/QGCViewDialogContainer.qml">src/QmlControls/QGCViewDialogContainer.qml</file>
-        <file alias="QGroundControl/Controls/QGCViewMessage.qml">src/QmlControls/QGCViewMessage.qml</file>
         <file alias="QGroundControl/Controls/qmldir">src/QmlControls/QGroundControl/Controls/qmldir</file>
         <file alias="QGroundControl/Controls/RallyPointEditorHeader.qml">src/PlanView/RallyPointEditorHeader.qml</file>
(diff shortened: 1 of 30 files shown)


Williangalvani:
Nice cleanup.
Plus this fixes #10113

patrickelectric (approved):

Williangalvani:
> Nice cleanup. Plus this fixes #10113

upon double-checking it, I noticed I had this unstaged:

```
diff --git a/src/QmlControls/QGCPopupDialog.qml b/src/QmlControls/QGCPopupDialog.qml
index e61f67967..1a3de0691 100644
--- a/src/QmlControls/QGCPopupDialog.qml
+++ b/src/QmlControls/QGCPopupDialog.qml
@@ -56,7 +56,7 @@ Popup {
     property alias  acceptButtonEnabled:    acceptButton.enabled
     property alias  rejectButtonEnabled:    rejectButton.enabled
     property var    dialogProperties
-    property bool   destroyOnClose:         true
+    property bool   destroyOnClose:         false
     property bool   preventClose:           false
 
     signal accepted
```

@DonLakeFlyer Can we make destroyOnClose false by default?

DonLakeFlyer:
> Can we make destroyOnClose false by default?

The standard usage of QGCPopupDialog is to be used in a Component. This limits memory usage to only when the dialog is used. Hence the default of true. I tried finding a way to determine whether it's being used from a Component or a directly in the hierarchy so I could warn of incorrect usage and couldn't find a way.

PR #10106, opened 2022-01-07
Allow vehicles that have built-in joystick support to also use QGC joystick functions.

Williangalvani:
This exposes some of QGC's built-in functions to these users, specifically for control of video streams.
The options are filtered so that there are no repeated functions (there are both firmware and QGC implementation for many of the functions).

This is the final result:
joystick

Right now this only affects Sub.

### DIFF
--- a/src/VehicleSetup/JoystickConfigButtons.qml
+++ b/src/VehicleSetup/JoystickConfigButtons.qml
@@ -163,21 +163,53 @@ ColumnLayout {
                     }
                 }
 
-                FactComboBox {
-                    id:             mainJSButtonActionCombo
-                    width:          ScreenTools.defaultFontPixelWidth * 15
-                    fact:           hasFirmwareSupport ? controller.getParameterFact(-1, parameterName) : null;
-                    visible:        hasFirmwareSupport
-                    indexModel:     false
+                QGCComboBox {
+                    id:                         buttonActionCombo
+                    width:                      ScreenTools.defaultFontPixelWidth * 26
+                    property Fact fact:         controller.parameterExists(-1, parameterName) ? controller.getParameterFact(-1, parameterName) : null
+                    property Fact fact_shift:   controller.parameterExists(-1, parameterShiftName) ? controller.getParameterFact(-1, parameterShiftName) : null
+                    property var factOptions:   fact ? fact.enumStrings : [];
+                    model:                      [..._activeJoystick.assignableActionTitles, ...factOptions]
+                    property var isFwAction:    currentIndex >= _activeJoystick.assignableActionTitles.length
                     sizeToContents: true
-                }
 
+                    function _findCurrentButtonAction() {
+                        // Find the index in the dropdown of the current action, checks FW and QGC actions
+                        if(_activeJoystick) {
+                            if (fact && fact.value > 0) {
+                                // This is a firmware function
+                                currentIndex = _activeJoystick.assignableActionTitles.length + fact.value
+                                // For sanity reasons, make sure qgc is set to "no action" if the firmware is set to do something
+                                _activeJoystick.setButtonAction(modelData, "No Action")
+                            } else {
+                                // If there is not firmware function, check QGC ones
+                                currentIndex = find(_activeJoystick.buttonActions[modelData])
+                            }
+                        }
+                    }
+
+                    Component.onCompleted:  _findCurrentButtonAction()
+                    onModelChanged:         _findCurrentButtonAction()
+                    onActivated:            function (optionIndex) {
+                        var func = textAt(optionIndex)
+                        if (factOptions.indexOf(func) > -1) {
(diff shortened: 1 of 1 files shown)


src/VehicleSetup/JoystickConfigButtons.qml
```diff
@@ -163,24 +165,85 @@ ColumnLayout {
...
-                    fact:           hasFirmwareSupport ? controller.getParameterFact(-1, parameterName) : null;
-                    visible:        hasFirmwareSupport
-                    indexModel:     false
+                QGCComboBox {
+                    id:                         buttonActionCombo
+                    width:                      ScreenTools.defaultFontPixelWidth * 26
+                    property Fact fact:         controller.parameterExists(-1, `BTN${index}_FUNCTION`) ? controller.getParameterFact(-1, `BTN${index}_FUNCTION`) : null
+                    property Fact fact_shift:   controller.parameterExists(-1, `BTN${index}_SFUNCTION`) ? controller.getParameterFact(-1, `BTN${index}_SFUNCTION`) : null
+                    property var factOptions:   fact ? fact.enumStrings : [];
+                    property var qgcActions:    _activeJoystick.assignableActionTitles.filter(function(s){return s.includes("Camera") || s.includes("Stream") || s.includes("Zoom") || s.includes("Gimbal")  || s.includes("No Action")})
```
patrickelectric, line 174: You could break this function in a new line to make it easier to read and maintain :laughing: 
```
property var qgcActions: _activeJoystick.assignableActionTitles.filter(function(string){
    return s.includes("Camera") \
        || s.includes("Stream") \
        || s.includes("Stream") \
        || s.includes("Zoom") \
        || s.includes("Gimbal") \
        || s.includes("No Action") \
    })
```

or

```
property var qgcActions: _activeJoystick.assignableActionTitles.filter(function(string){
    return [
        s.includes("Camera")
        , s.includes("Stream")
        , s.includes("Stream")
        , s.includes("Zoom")
        , s.includes("Gimbal")
        , s.includes("No Action")
    ].some(Boolean)})
```

src/VehicleSetup/JoystickConfigButtons.qml
```diff
@@ -163,24 +165,85 @@ ColumnLayout {
...
+                        // Find the index in the dropdown of the current action, checks FW and QGC actions
+                        if(_activeJoystick) {
+                            if (fact && fact.value > 0) {
+                                // This is a firmware function
+                                currentIndex = qgcActions.length + fact.value
+                                // For sanity reasons, make sure qgc is set to "no action" if the firmware is set to do something
+                                _activeJoystick.setButtonAction(modelData, "No Action")
+                            } else {
+                                // If there is not firmware function, check QGC ones
+                                var i = find(_activeJoystick.buttonActions[modelData])
```
patrickelectric, line 189: this intermediary variable is a bit unnecessary, and currentIndex already provides the information that find will return an index like variable.

src/VehicleSetup/JoystickConfigButtons.qml
```diff
@@ -163,24 +165,85 @@ ColumnLayout {
...
 
                 FactComboBox {
-                    id:             shiftJSButtonActionCombo
-                    width:          ScreenTools.defaultFontPixelWidth * 15
-                    fact:           hasFirmwareSupport ? controller.getParameterFact(-1, parameterShiftName) : null;
-                    visible:        hasFirmwareSupport
-                    indexModel:     false
+                    id:         shiftJSButtonActionCombo
+                    width:      ScreenTools.defaultFontPixelWidth * 26
+                    fact:       controller.parameterExists(-1, `BTN${index}_SFUNCTION`) ? controller.getParameterFact(-1, `BTN${index}_SFUNCTION`) : null;
```
patrickelectric, line 235: Should we make a helper function to avoid maintaining this string around the code ?
The same thing for `BTN${INDEX}_FUNCTION`

patrickelectric (changes requested):

patrickelectric (approved):

junwoo091400:
Could you clarify what a FW action, and QGC action is? I have been browsing through the code and found it quite un-intuitive :thinking:

Is FW action, somehow related to component metadata or sth?

Williangalvani:
Hi,

That is an ArduSub peculiarity, I guess.
Ardusub has all these functions implemented in the firmware:
https://github.com/ArduPilot/ardupilot/blob/Sub-4.1/libraries/AP_JSButton/AP_JSButton.h#L11-L113
This means we just tell ArduSub "hey, button x is pressed" and ArduSub does whatever it wants with it. I called this a FW action in this code.

QGC Actions are handled by QGC. QGC can do things such as switching video streams and cameras, as well as sending MAVLink commands to the vehicle.

There is actually an overlap in these, but they are both useful for Sub, as QGC not able to control some Sub functionality with MAVLink, and ArduSub is unable to control video streams, for example.

I hope this was clear enough.

PR #10110, opened 2022-01-10
APM: show sensor names, improve compass setup

Williangalvani:
This shows what sensors are detected and at which buses:
image

It also allows changing compass priority without doing a re-calibration:

image


src/AutoPilotPlugins/APM/APMSensorsComponentSummary.qml
```diff
@@ -26,25 +27,88 @@ Item {
...
+                }
+
+                APMSensorIdDecoder {
+                    horizontalAlignment:    Text.AlignRight
+                    Layout.alignment:       Qt.AlignRight
+
+                    fact: sensorParams.rgCompassPrio[index]
+                }
+
+                function getPriority (index) {
```
patrickelectric, line 73: Why there is a duplicated function here ?

src/AutoPilotPlugins/APM/APMSensorsComponentSummary.qml
```diff
@@ -26,25 +27,88 @@ Item {
...
         }
 
         VehicleSummaryRow {
-            labelText: qsTr("Accelerometer(s)")
+            labelText: qsTr("Accelerometer(s):")
             valueText: controller.accelSetupNeeded ? qsTr("Setup required") : qsTr("Ready")
         }
+
+        Repeater {
+            model: 3
```
patrickelectric, line 94: Can't we use sensorParams.rgInsId.length ? This will make future proof

src/AutoPilotPlugins/APM/APMSensorsComponentSummary.qml
```diff
@@ -26,25 +27,88 @@ Item {
...
+            }
+        }
+
+        VehicleSummaryRow {
+            labelText: qsTr("Barometer(s):")
+            valueText: ""
+        }
+
+        Repeater {
+            model: 3
```
patrickelectric, line 107: Same thing with length here

src/FirmwarePlugin/APM/APMSensorIdDecoder.qml
```diff
@@ -0,0 +1,152 @@
...
+        2: 'SPI',
+        3: 'UAVCAN',
+        4: 'SITL',
+        5: 'MSP',
+        6: 'EAHRS'
+    }
+
+    property var compassTypes: {
+        0x01: 'HMC5883_OLD',
+        0x07: 'HMC5883',
```
patrickelectric, line 36: Should this be in numeric order to avoid confusion ?

patrickelectric (changes requested):

src/FirmwarePlugin/APM/APMSensorIdDecoder.qml
Williangalvani: this is pretty much a direct copy/paste from https://github.com/ArduPilot/MAVProxy/blob/9bf8b00fbc355650b060e546110877c7898baa81/MAVProxy/modules/lib/mp_util.py#L384

I think it is easier to just keep it the same.

src/AutoPilotPlugins/APM/APMSensorsComponentSummary.qml
Williangalvani: bad rebase :sweat_smile:

patrickelectric: This is still here

src/FirmwarePlugin/APM/APMSensorIdDecoder.qml
```diff
@@ -0,0 +1,152 @@
...
+        0x05: 'MSP',
+        0x06: 'SDP3X',
+        0x07: 'UAVCAN',
+        0x08: 'ANALOG',
+        0x09: 'NMEA',
+        0x0A: 'ASP5033'
+    }
+
+    function decode (device) {
+        var devid = parseInt(device.valueString)
```
patrickelectric, line 118: It would be nice to add a reference of where this decode logic was extracted.

patrickelectric (changes requested):

src/AutoPilotPlugins/APM/APMSensorsComponentSummary.qml
Williangalvani: what about now?

patrickelectric (approved):

PR #10115, opened 2022-01-12
Implement fast compass calibration for Ardupilot Vehicles

Williangalvani:
fastcompass

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSensorsComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMSensorsComponent.qml
@@ -22,6 +22,7 @@ import QGroundControl.Controls      1.0
 import QGroundControl.ScreenTools   1.0
 import QGroundControl.Controllers   1.0
 import QGroundControl.ArduPilot     1.0
+import QGroundControl.QGCPositionManager    1.0
 
 SetupPage {
     id:             sensorsPage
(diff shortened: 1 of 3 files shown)


src/AutoPilotPlugins/APM/APMSensorsComponent.qml
```diff
@@ -499,6 +522,76 @@ SetupPage {
...
+                                    width:      parent.width
+                                    visible:    useMapPositionCheckbox.checked
+                                    wrapMode:   Text.WordWrap
+                                    text:       qsTr(`Lat: ${_mapPosition.latitude.toFixed(4)} Lon: ${_mapPosition.longitude.toFixed(4)}`)
+                                }
+
+                                FactTextField {
+                                    id:         northCalLat
+                                    visible:    !useGcsPositionCheckbox.checked && !useMapPositionCheckbox.checked && northCalibrationCheckBox.checked
+                                    text:       "0.00"
```
patrickelectric, line 583: Why _gcsPosition is not being used for latitude ?

src/AutoPilotPlugins/APM/APMSensorsComponent.qml
```diff
@@ -411,11 +412,33 @@ SetupPage {
...
+                               var lon = parseFloat(northCalLon.text)
+                               if (useMapPositionCheckbox.checked) {
+                                   lat = _mapPosition.latitude
+                                   lon = _mapPosition.longitude
+                               }
+                               if (useGcsPositionCheckbox.checked) {
+                                   lat = _gcsPosition.latitude
+                                   lon = _gcsPosition.longitude
+                               }
+                               controller.calibrateCompassNorth(lat, lon, compassMask())
```
patrickelectric, line 440: You should check if lat and lon are `NaN`, since `parseFloat` (and maybe mapPosition or gcsPotion) could return invalid values.

patrickelectric (changes requested):

src/AutoPilotPlugins/APM/APMSensorsComponent.qml
Williangalvani: changed both to zero. mid-writing this I decided to use a label when showing the gcs position, instead of the textfield.

patrickelectric (approved):

PR #10119, opened 2022-01-14
Actuators: minor updates

bkueng:
- disable motor assignment button if geometry does not have motors
- increase motor spin timeout to 3s when assigning motors 
  So that ESC's can initialize. Note that some are faster than others and 1s is enough in some cases.

I can bring this to stable as well.


patrickelectric (approved):
The code looks correct

bkueng:
Just fixed the variable naming convention

PR #10123, opened 2022-01-18
Support for GuidedAction Vtol Takeoff

RomanBapst:
Currently QGC only exposes a simple takeoff (multi rotor takeoff or whatever you want to call it) in the guided actions menu.
For hybrid VTOL however, it would be convenient for the user to have an option to execute a takeoff and directly transition into forward flight after reaching the desired takeoff height. The vehicle could then establish a predefined loiter which allows it to remain in sight of the user and does not require any immediate action from the user.

The approach described above is implemented in this pull request and currently supports PX4 Firmware only. If there is interest from the Ardupilot community, I'd be happy to help adding support for their flight stack as well.

Below you can find a link to a video which demonstrates this functionality:
https://www.youtube.com/watch?v=AwHqYNDKTEI

**Implementation Details**

The implementation relies on [MAV_CMD_NAV_VTOL_TAKEOFF](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_VTOL_TAKEOFF) in order to send the 3D location of the loiter circle to the flight controller.
This mavlink command is actually not really well documented, I'm hoping we can improve that as part of this PR.
The altitude specified in the command (param 7) holds the relative transition height above ground.
The height to which the vehicle climbs after the front transition is currently set by the flight controller but we could discuss to add it to the command as well (e.g. param 1 is still available). This would give the user the option to specify both transition and loiter altitude from the ground station. I'm looking forward to hear some opinions regarding this.

The position of the loiter circle is currently constrained to what I deemed reasonable values. The expectation is generally that the user roughly aligns the nose of the vehicle with the wind. The loiter circle center is then constrained to an angle of 90 degrees centered about the nose of the vehicle. Furthermore, the maximum and minimum distance of the loiter center to the current position of the vehicle is constrained taking into account the loiter radius of the circle.

As mentioned above, I'm currently only adding support for the PX4 flight stack and will post the link to the respective PR here shortly:

**Future work**
 Once VTOL takeoff as a guided action is supported, the next step would be supporting deterministic VTOL land approaches from the fly view, similar to what is shown in this video:
 https://auterion.com/effortless-and-safer-vtol-deployment/


RomanBapst:
@patrickelectric Would you mind reviewing this?

DonLakeFlyer:
This has been sitting for ages with nobody looking at it. Not sure if this canoe brought up to date?

DonLakeFlyer:
@RomanBapst Do you think you could help resurrect this?

julianoes:
@RomanBapst last call...

DonLakeFlyer:
Closing. Can re-open if someone picks up work on it.

PR #9701, opened 2021-06-11
(cherry-pick to stable) Vehicle: hide 'No GPS Lock' message if GPS is not present

Williangalvani:
Just getting #9393 to stable.

We are still recommending 4.0 to our users due to this and a couple of other minor issues.

PR #10127, opened 2022-01-20
Implement 'Set EKF Origin' in fly view

Williangalvani:
This is useful for ardupilot for navigation using odometry and other gps-less modes.
set_ekf_origin

Is this relevant for PX4? should it be hidden behind a `supportsSettingEkfOrigin()` in FirmwarePlugin?

This needs testing. My sensors are dead...


patrickelectric (approved):

bkueng:
PX4 implements this as well. However from a UI perspective I'm not sure where it makes sense to expose this to all users.

dagar:
FYI @mcsauder

dagar:
How about slightly more generic naming as the message (SET_GPS_GLOBAL_ORIGIN) isn't EKF specific and it isn't necessarily an EKF on the other side?

http://mavlink.io/en/messages/common.html#SET_GPS_GLOBAL_ORIGIN

mrpollo:
@bkueng @dagar do you have time to help me push this through the finish line? its a good feature that could benefit users

mcsauder (approved):
Looks great.  Sorry it took so long.  Thanks for your work!

mcsauder:
Fix up the conflicts and maybe rebase.  After that it should be good to go!

-Mark

mrpollo:
Thanks @mcsauder I went ahead and fixed the conflicts so we can get this in ASAP. I would like to make a new point release soon.

bkueng (commented):
- At the minimum please change ekf to estimator
- Can you only expose this if no gps is used? It's in SYS_STATUS.

patrickelectric:
ping @Williangalvani can you update the code based on the review and rebase over master ?

DonLakeFlyer:
I can clean this up. But is there some way to understand when this command is necessary? As far as I can tell from the current state of the code showSetEkfOrigin is always true. So every QGC user is going to see this new menu item. Even though a massive percentage of users will have no idea what it's used for.

Williangalvani:
With Ardupilot, this is useful when you are flying a vehicle with no absolute positioning. In our case, we have ROVs with very accurate DVL (think velocity sensors) which can be used for navigation. But first the user has to manually tell the vehicle where it is for it to have a starting point.

DonLakeFlyer:
Yeah I get that part. But I"m assuming this is only needed in certain situations. For example: Should this option only be shown if QGC isn't receiving any GPS telemetry? OR something else like that?

Williangalvani:
Right, from Ardupilot's perspective. we should be able to show this only if we have GLOBAL_POSITION_INT showing lat/long == (0,0)
There might be a better option, let me ask the other devs

DonLakeFlyer:
> * Can you only expose this if no gps is used? It's in SYS_STATUS.

Oh, wait I just saw this. Does this work for PX4 and ArduPilot?

Williangalvani:
Yes that makes sense! it is actually how I hid the "No GPS Fix" message for sub in here:
https://github.com/mavlink/qgroundcontrol/pull/9701/files#diff-86909c8722a0a27e4936cfbe08a7e0b463c6f8100575adf84375e312659e9508R558

src/FlightDisplay/GuidedActionsController.qml
```diff
@@ -127,6 +130,7 @@ Item {
     property bool showROI:              _guidedActionsEnabled && _vehicleFlying && __roiSupported && !_missionActive
     property bool showLandAbort:        _guidedActionsEnabled && _vehicleFlying && _fixedWingOnApproach
     property bool showGotoLocation:     _guidedActionsEnabled && _vehicleFlying
+    property bool showSetEkfOrigin:     true
```
DonLakeFlyer, line 133: So then this should change to: 
```
property bool showSetEkfOrigin:     _activeVehicle.sensorsPresentBits & 32 // MAV_SYS_STATUS_SENSOR_GPS
```

DonLakeFlyer:
@Williangalvani ?

PR #10137, opened 2022-01-24
Small ui tweaks

gillamkid:
This PR fixes a couple things that don't look right.

1. PhotoVideoControl.qml
Outdoor mode before change
beforeLight
Outdoor mode after the change
afterLight

2. QGCComboBox.qml.qml
Alignment before change
beforeMavDark
Alignment after change
afterMavDark

3. MavlinkSettings.qml
Indoor mode before change
beforeMavLight
Indoor mode after the change
afterMavLight


mrpollo:
Great, thanks for the small but really helpful PR @gillamkid 👍

PR #10147, opened 2022-02-07
Linux: incremental build speedup

bkueng:
Time used for incremental build when changing a single .cc file:
- master: 26s
- timestamp check for library update: 9s
- timestamp check + gold linker: 5s

So overal build time reduces from 26s to 5s.

@keith-bennett-airmap

### DIFF
--- a/deploy/linux-fixup-rpaths.bash
+++ b/deploy/linux-fixup-rpaths.bash
@@ -47,40 +47,47 @@ find "${SEARCHDIR}" \
     -executable \
     2>/dev/null |
 while IFS='' read -r library; do
-    # Get the library's current RPATH (RUNPATH)
-    # Example output of `readelf -d ./build/build-qgroundcontrol-Desktop_Qt_5_15_2_GCC_64bit-Debug/staging/QGroundControl`:
-    #  0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/Qt/libs:/home/kbennett/storage/Qt/5.15.2/gcc_64/lib]
-    #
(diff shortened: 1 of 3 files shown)


deploy/linux-fixup-rpaths.bash
```diff
@@ -47,40 +47,47 @@ find "${SEARCHDIR}" \
...
-        readelf -d "${library}" |
-        grep -P '^ 0x[0-9a-f]+ +\(RUNPATH\) ' |
-        sed -r 's/^ 0x[0-9a-f]+ +\(RUNPATH\) +Library runpath: \[(.*)\]$/\1/g'
-    )"
-    set -o pipefail
 
-    # Get the directory containing the library
-    library_dir="$(dirname "${library}")"
+    # readelf is expensive, so keep track of updates with a timestamp file
+    if [ ! -e "$library.stamp" -o "$library" -nt "$library.stamp" ]; then
```
keith-bennett-airmap, line 52: FYI: [shellcheck](https://shellcheck.net/) presents a warning (not error):

```
Line 52:
    if [ ! -e "$library.stamp" -o "$library" -nt "$library.stamp" ]; then
                               ^-- [SC2166](https://github.com/koalaman/shellcheck/wiki/SC2166) (warning): Prefer [ p ] || [ q ] as [ p -o q ] is not well defined.
```

Other than that, it looks good to me.

bkueng: Thanks for checking. Let's avoid it then: https://github.com/mavlink/qgroundcontrol/pull/10154

Issue #10016, opened 2021-11-12
Ubuntu 18.04 Stable AppImage for v4.1.4 is Outdated and Unusable 


3dpinta:
Issue #9847 brought up a the problem of Ubuntu 18.04 users not being able to run the latest version (v4.1.4) of QGC. Issue #9853 address #9847 and appears to fix the problem but as the later comments in Issue #9847 show, the latest github releases and the AppImage linked to in the downloads sections still point to the version that will only work on Ubuntu 20.04. If the correct version can be linked, that would be great.

Information requested:
- Ubuntu 18.04 LTS
- v4.1.4 stable (from the downloads page). Have to roll back to 4.0.11 to get a functioning GCS. There are no AppImage files for 4.1.0-4.1.3 stable available to pick up more recent improvements/fixes.
- Cube Orange, Cube Black
- PX4 latest and ArduCopter 4.0.3
- Downloaded latest stable image, followed instructions at https://docs.qgroundcontrol.com/master/en/getting_started/download_and_install.html, and when starting via command line and from icon, I continue to get the error from Issue #9847

Jaeyoung-Lim:
I can confirm that QGroundcontrol is broken on ubuntu 18.04
```
/tmp/.mount_QGrouneKMv3t/QGroundControl: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /tmp/.mount_QGrouneKMv3t/QGroundControl)
```

mrpollo:
@3dpinta @Jaeyoung-Lim we will be releasing a new stable release this week that will be compatible with Ubuntu 18.04 again. We are sorry for the inconvenience.

3dpinta:
@mrpollo Can the correct one at least be linked? If someone needs to use an older version for compatibility purposes, it would be helpful to get to it rather than have to go rebuild it themselves.

patrickelectric:
[check 4.1.5 release ](https://github.com/mavlink/qgroundcontrol/releases/tag/v4.1.5)

ES-Alexander:
> [check 4.1.5 release ](https://github.com/mavlink/qgroundcontrol/releases/tag/v4.1.5)

This is [confirmed to be working](https://discuss.bluerobotics.com/t/video-streaming-is-not-working-on-linux-pc/11523/12). This issue can be closed.

ashishmagar600:
[...]

PR #10153, opened 2022-02-09
use upstream gst-plugins-good

### LINKED ISSUE #10140: Use upstream gst-plugins-good
- What problem does it solve?
  Using the upstream https://github.com/GStreamer/gst-plugins-good allows us to track changes much easily. In particular https://github.com/mavlink/qgroundcontrol/issues/9669 and the associated PR https://github.com/mavlink/gst-plugins-good/pull/1 would not have been needed had we used the upstream repo instead of our own fork.
- What does it do?
  Replace our fork of https://github.com/mavlink/gst-plugins-good with upstream https://github.com/GStreamer/gst-plugins-good
  All
  All
  N/A
  N/A

Any additional context you can provide will make the feature easier to evaluate (e.g. mockups, detailed specification, etc.)

Following conversation from above mentioned issue and PR.

Looks like the only difference is this one commit: https://github.com/mavlink/gst-plugins-good/commit/9d782fad9dc0384ba86ecae64511c193f6149f93

Understanding what it does and if the capability has already been incorporated into upstream (if not then sending the appropriate PR upstream) will be needed to resolve this issue.

acxz:
issue for multiple qmlglsink instances for windows has been fixed in 1.20.0 release

closes #10140

### DIFF
--- a/.github/workflows/linux_debug.yml
+++ b/.github/workflows/linux_debug.yml
@@ -19,7 +19,7 @@ env:
 
 jobs:
   build:
-    runs-on:  ubuntu-22.04
+    runs-on:  ubuntu-20.04
 
     steps:
       - name: Checkout repo
--- a/.github/workflows/linux_debug.yml
+++ b/.github/workflows/linux_debug.yml
@@ -19,7 +19,7 @@ env:
 
 jobs:
   build:
-    runs-on:  ubuntu-18.04
+    runs-on:  ubuntu-22.04
 
     steps:
       - name: Checkout repo
--- a/.github/workflows/android_32_release.yml
+++ b/.github/workflows/android_32_release.yml
@@ -79,9 +79,9 @@ jobs:
       - name:               Install gstreamer
         working-directory:  ${{ github.workspace }}
         run: |
-            wget --quiet https://gstreamer.freedesktop.org/data/pkg/android/1.18.5/gstreamer-1.0-android-universal-1.18.5.tar.xz
-            mkdir gstreamer-1.0-android-universal-1.18.5
-            tar xf gstreamer-1.0-android-universal-1.18.5.tar.xz -C gstreamer-1.0-android-universal-1.18.5
+            wget --quiet https://gstreamer.freedesktop.org/data/pkg/android/1.20.0/gstreamer-1.0-android-universal-1.20.0.tar.xz
+            mkdir gstreamer-1.0-android-universal-1.20.0
+            tar xf gstreamer-1.0-android-universal-1.20.0.tar.xz -C gstreamer-1.0-android-universal-1.20.0
 
       # This will set GIT_BRANCH_NAME environment variable
       - name: Git branch name
--- a/.github/workflows/android_64_release.yml
+++ b/.github/workflows/android_64_release.yml
@@ -79,9 +79,9 @@ jobs:
       - name:               Install gstreamer
         working-directory:  ${{ github.workspace }}
         run: |
-            wget --quiet https://gstreamer.freedesktop.org/data/pkg/android/1.18.5/gstreamer-1.0-android-universal-1.18.5.tar.xz
-            mkdir gstreamer-1.0-android-universal-1.18.5
-            tar xf gstreamer-1.0-android-universal-1.18.5.tar.xz -C gstreamer-1.0-android-universal-1.18.5
+            wget --quiet https://gstreamer.freedesktop.org/data/pkg/android/1.20.0/gstreamer-1.0-android-universal-1.20.0.tar.xz
+            mkdir gstreamer-1.0-android-universal-1.20.0
+            tar xf gstreamer-1.0-android-universal-1.20.0.tar.xz -C gstreamer-1.0-android-universal-1.20.0
 
       # This will set GIT_BRANCH_NAME environment variable
       - name: Git branch name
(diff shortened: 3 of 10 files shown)


acxz:
Looks like the CI is adding multiple entries to `.gitmodules`?
https://github.com/mavlink/qgroundcontrol/runs/5120305065?check_suite_focus=true#step:3:130

Is there a way to run
```
git submodule sync
git submodule init
git submodule update
```
?

acxz:
Getting the following error if I do not include `remote` under the `submodule` field:
```
Errors during submodule fetch:
	libs/qmlglsink/gst-plugins-good
error: could not fetch origin
```

However, it seems that adding `remote` will update the submodules to newer commits. The build also fails later on, which could be due to the submodules being updated to the newer commits.

I tried using the `--prune` and the `--prune-flags` options to `git fetch` but still no luck.
Note: this is seems to be a CI issue and I have no issues running `git fetch` locally.

@DonLakeFlyer can you take a look at this CI issue? For now in my PR I have removed the Tag Information block as I don't know if it is required for the rest of the CI build.

acxz:
Can build locally, but looks like some dependency versions on the CI might be outdated resulting in the error we see. Still need to look deeper into this.

acxz:
Looks like we need a gstreamer archive for macos. And a newer ubuntu(22.04) with gstreamer=1.20.0

acxz:
The current macos issue I am seeing in the ci is the following:
```
linking QGroundControl.app/Contents/MacOS/QGroundControl
ld: file not found: @rpath/lib/libgsttag-1.0.0.dylib for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [QGroundControl.app/Contents/MacOS/QGroundControl] Error 1
```

Tracked upstream here: https://gitlab.freedesktop.org/gstreamer/cerbero/-/issues/338

As for the linux builds, I'm guessing with the newer ubuntu release, a lot of people of using it so there is prob a good bit of backlog.

.github/workflows/linux_debug.yml
```diff
@@ -19,7 +19,7 @@ env:
 
 jobs:
   build:
-    runs-on:  ubuntu-18.04
+    runs-on:  ubuntu-20.04
```
DonLakeFlyer, line 22: So the fact that the Linux versions are built on Ubuntu 18 is important since it allows pre-built versions of QGC to run on Ubuntu 18 and 20. Moving to 20 will lose that capability. Maybe we just toss support for 18 with this newest version of QGC. I think 18 is getting pretty old now right?

DonLakeFlyer: So I assume Ubuntu 20 comes with gstreamer 1.20 right?

DonLakeFlyer: Wait just say the note about Ubuntu 22. How can that be feasible then? We are going to need Ubuntu 20 support I would say since that is still pretty recent.

acxz: Yeah I agree with you, I think bumping up to Ubuntu 20.04 is good, but we shouldn't move our ci stack to 22 for a while.

I'll make a separate PR bumping our ubuntu from 18 to 20 and then I'll update this PR with a manual install of gstreamer in the ci if that works?

DonLakeFlyer: That would also mean that the AppImage needs to include the updated version of gstreamer as well so that it just works out of the box without additional installs. Doesn't it?

acxz: Also I have a quick question: if the CI is building the AppImage then does it matter which Ubuntu version we base off of? Does the AppImage contain everything that is needed to run qgroundcontrol?

DonLakeFlyer: "if the CI is building the AppImage then does it matter which Ubuntu version we base off of?" You would think so. But something was linking to a specific OS lib which didn't exist or maybe different version on 20. I think there is a closed issue on it somewhere. I think @mrpollo  discovered the problem. You can probably go through commit blame on the CI file to find the change which was recent.

acxz: Weird I guess TIL.

DonLakeFlyer: @mrpollo Do you remember if there an issue on the Ubuntu 18 problem?

mrpollo: Yeah it was a `glibc` version incompatibility see #10016, the only viable alternative at the time was to downgrade to ubuntu-18, especially since we didn't really have a version to upgrade to ubuntu-20 in the first place

do we have a reason to upgrade now?

acxz: Not really the version that this PR needs is Ubuntu 22 or building gstreamer from source. I've been trying to get gstreamer building from source for now as Ubuntu 22 is too far of a platform to target. Updating to Ubuntu 20 will not help this PR.

DonLakeFlyer: Given that we should keep CI at 18 then to keep back compat.

smagellan:
Maybe I misunderstand something, then please correct me if I wrong
1. QGC builds-in one of gst-good plugins (static linkage/linking).
2. QGC uses GStreamer shared libraries anyway. (dynamic linking)
3. Does it make sense to keep step #1 given upstream GStreamer plugins are fine now?

PR #10157, opened 2022-02-10
limit telemetry bar size by screen size

holden-zenith:
This adds a limit to the number of columns and rows in the telemetry bar, otherwise it gets so big that the buttons go off screen. Also remove Note as the bug has been patched in 5.15.1

### DIFF
--- a/src/QmlControls/HorizontalFactValueGrid.qml
+++ b/src/QmlControls/HorizontalFactValueGrid.qml
@@ -20,10 +20,6 @@ import QGroundControl.Palette       1.0
 import QGroundControl.FlightMap     1.0
 import QGroundControl               1.0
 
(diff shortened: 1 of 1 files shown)


src/QmlControls/HorizontalFactValueGrid.qml
```diff
@@ -126,7 +121,7 @@ T.HorizontalFactValueGrid {
                     Layout.preferredHeight: ScreenTools.minTouchPixels
                     Layout.preferredWidth:  parent.width
                     text:                   qsTr("+")
-                    onClicked:              appendColumn()
```
patrickelectric, line 129: This check should be on the enabled of the button, or in the parent that shows such buttons. Otherwise the user will click the button and nothing will happen.

patrickelectric (changes requested):

mrpollo:
Thanks @holden-zenith

Issue #10122, opened 2022-01-17
 QList::insert(): index out of range warning when uploading a mission contains geofence


hitagisol:
- Ground station operating system: Ubuntu 20.04
- QGroundControl version and build: self build from v4.2 (debug build)
- Autopilot board: Pixhawk 4 (Also reproduceble with SITL)
- Autopilot firmware (e.g. PX4, ArduPilot, custom) and version: PX4 v1.12.3
- Exact steps to reproduce the problem. Starting from booting QGroundControl to when the problem occurs:
  - Boot QGC, link with the Pixhawk or start a SITL such as jmavsim.
  - Plan a mission with geofence
  - upload the mission
  - some Qt warning will shown as: QList::insert(): index out of range.
 
After some investication I found the warning comes from [https://github.com/mavlink/qgroundcontrol/blob/5496f81cdb227a815d17c8daf99614eb37c97f9d/src/MissionManager/GeoFenceManager.cc#L172](url). 
However seems the real problem comes from here: [https://github.com/mavlink/qgroundcontrol/blob/5496f81cdb227a815d17c8daf99614eb37c97f9d/src/QmlControls/QmlObjectListModel.cc#L207](url).

```
 for (QObject* object: objects) {
        QQmlEngine::setObjectOwnership(object, QQmlEngine::CppOwnership);

        // Look for a dirtyChanged signal on the object
        if (object->metaObject()->indexOfSignal(QMetaObject::normalizedSignature("dirtyChanged(bool)")) != -1) {
            if (!_skipDirtyFirstItem || j != 0) {
                QObject::connect(object, SIGNAL(dirtyChanged(bool)), this, SLOT(_childDirtyChanged(bool)));
            }
        }
        j++;

        _objectList.insert(j, object);
    } 
```

The insert is called after the increment of j, which is not the correct usage of QList. I tried with switch the order of j++ and insert then the warning disappears. Or maybe just using QList::append which shall be equally as QList::insert(size(), value) frem the reference of Qt [https://doc.qt.io/qt-5/qlist.html#append](url).

PR #10159, opened 2022-02-10
Fix PlanView warnings

holden-zenith:
This fixes a warning about property width of null not existing when all mission items are removed and fixes the QList insert method that produced an invalid index warning from being one index too high each insertion (#10122).

### DIFF
--- a/src/PlanView/PlanView.qml
+++ b/src/PlanView/PlanView.qml
@@ -795,7 +795,7 @@ Item {
                         map:            editorMap
                         masterController:  _planMasterController
                         missionItem:    object
-                        width:          parent.width
+                        width:          missionItemEditorListView.width
                         readOnly:       false
                         onClicked:      _missionController.setCurrentPlanViewSeqNum(object.sequenceNumber, false)
                         onRemove: {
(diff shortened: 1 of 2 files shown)


src/PlanView/PlanView.qml
```diff
@@ -795,7 +795,7 @@ Item {
                         map:            editorMap
                         masterController:  _planMasterController
                         missionItem:    object
-                        width:          parent.width
+                        width:          parent ? parent.width : 0
```
mrpollo, line 798: @holden-zenith if the parent is null then this component shouldn't exist either, rather than setting the width to 0 can you remove it?

src/QmlControls/QmlObjectListModel.cc
```diff
@@ -220,9 +220,9 @@ void QmlObjectListModel::insert(int i, QList<QObject*> objects)
                 QObject::connect(object, SIGNAL(dirtyChanged(bool)), this, SLOT(_childDirtyChanged(bool)));
             }
         }
-        j++;
 
         _objectList.insert(j, object);
+        j++;
```
mrpollo, line 225: This looks good

mrpollo (changes requested):

PR #10169, opened 2022-02-15
GuidedAltitudeSlider: Refactor to become a generic value slider

RomanBapst:
This is preparation work for enabling speed changes from the fly view.

1) For fixed wings / VTOL the user will be able to change the airspeed setpoint via the speed slider (constrained between min and max airspeed)
2) For multirotors the user will be able to set the cruise speed.

This PR refactors the guidedAltitudeSlider such that it can be used as a generic slider which needs to be configured depending on the quantity at hand (e.g. takeoff height, height change, groundspeed, airspeed, ...)


RomanBapst:
Replaced by https://github.com/mavlink/qgroundcontrol/pull/10174

PR #10170, opened 2022-02-15
Fix #10143: ADSB SBS-1 messages parsing and displaying

### LINKED ISSUE #10143: ADS-B data feed not reliably displayed.
[ADSB-2022-02-02.txt](https://github.com/mavlink/qgroundcontrol/files/7989320/ADSB-2022-02-02.txt)
## Expected Behavior
Planes are displayed on the map when they are sent by ADS-B

## Current Behavior
Planes don't reliably show up. Sometimes they do, then disappear, sometimes they don't even show. I have a web app showing nearby planes, netcat showing my dumb1080 feed from my raspberry pi and Mission Planner reading the same feed shows planes that don't show up on th QGC map. Also the QGC logs viewer shows that QGC is receiving the ADSB feed (see attached screenshot), but the planes shown in the log are not showing on the map.

This is on a Mac if it matters (MacPro 2021 version M1Pro running 12.1

## Steps to Reproduce:
1. Configure QGC to use my local (raspberry pi) ADS-B server at port 30003
2. Go "back" to the main map. 
3. Planes don't display most of the time.
4. Go to Application Settings ->Console turn on "Trace" trace log shows the ADSB data is being received. I've clicked the "Save App Log" button and uploaded a copy of the file that was saved.

## System Information
- OSX 12.1
- 4.2.0
- Stable

## Detailed Description
Screen Shot 2022-02-02 at 10 17 57 AM
Screen Shot 2022-02-02 at 10 16 25 AM

## Log Files and Screenshots
-Screen shots attached
-App log attached.

zdanek:
Hi.
I'm very thrilled that I can push my first PR to QGroundControl. I've spend whole weekend playing with the code. Now I've tidy it up and here it is :)

# Fixes and improvements

## Choice of MSG types
I have analysed lots of ADSB SBS-1 messages from various sources and despite that protocol description states that Callsign should be present only in MSG type 1, it's also often present in MSG types 5 and 6. So I've included those types as well to increase chances to obtain Callsign ASAP.

## Decreasing timeout/cleanup time
I've decreased cleanup (expiry) time do 60s. Previous was 120s. The reasoning behind that is that maybe except for helicopters all aircrafts with ADSB are in constant move with speed expected to be more than 100km/h (small low flying planes). At this speed after 2 mins aircraft should to be further than 4km away (if it's a jet it will be over 30km away!). It makes no sense to keep it frozen on the map. A heli is an exception but it's more likely that we have a plane above our heads and 2min timeout is simply too much IMO. To be discussed.

## Fix in altitude calculation
First of all there was an error in the formula. To convert altitude in `ft` to `m` it should be multiplied by `0.3048`. If you're into dividing then `ft` should be divided by `3.28` **not** by `3.048`. 

Also @timtuxworth is using SBS-1 data source that sends altitude in HAE (GPS altitude) instead of Barometric alt. Those are the values with H at the end.  I've spent about 3 hours to understand what this might be as SBS-1 is documented good on only one web page if found:
[http://woodair.net/sbs/article/barebones42_socket_data.htm](http://woodair.net/sbs/article/barebones42_socket_data.htm)
with no info about H altitude. It mentioned HAE but as an internal feature of SBS BaseStation there. So I deduced it might be HAE as an altitude with H.

The main problem with **NOT** displaying most of @timtuxworth data was that the `ASDBVehicleManager` code exits if it fails converting altitude string (here an int with H character) and thus not showing planes. This was the main reason that Tim opened a bug ticket :)

## Fix of reading data from Socket
I've discovered a major problem with reading of incoming data! As I was experimenting with ADSB messages I discovered that after about 30 minutes all messages were delayed about 15-20 seconds. A message was sent by an ADSB server but it was parsed and used by QGC with a such delay.   The reason was that at every TCP `readyRead` event only one line was read. And if messages flooded QGC a queue of "lines" was rising. I've fixed that and now messages are parsed immediately.

## Other fixes
I've fixed width of ADSB server address in the Settings as it was to narrow to hold the address.
And I've changed a bit storing of `ADSBVehicles` so a Vehicle is also created when a callsign comes in first without knowing a position. This makes it more usable because callsign rarely come, and it's good to store it as soon as it comes. The Vehicle is not shown if it misses Lat and Lon, so it was required to fix a bit UI but without storing callsign it would not show either before Lat, Lon was received.

While looking for display delay (see tcp problem above) I've applied some minor speedups that I think are worth keeping. For example skipping splitting/parsing messages of non-supported types. Why bother CPU with data that will be discarded :)

**I can explain or defend every line in this PR so please ask :)**

# How to test

Connect to own ASDB SBS-1 server OR use my ADSB Fake replay server :)
I've created for this occasion an ADSB SBS-1 replay server that emits messages from csv file with a respect for message record times.

[adsb tools](https://github.com/zdanek/adsb-tools)

Using flags you can enable displaying most important messages and looping. More features to come. It's very useful. Stars would be appreciated :star2:

### DIFF
--- a/src/ADSB/ADSBVehicleManager.cc
+++ b/src/ADSB/ADSBVehicleManager.cc
@@ -23,7 +23,6 @@ ADSBVehicleManager::ADSBVehicleManager(QGCApplication* app, QGCToolbox* toolbox)
 void ADSBVehicleManager::setToolbox(QGCToolbox* toolbox)
 {
     QGCTool::setToolbox(toolbox);
-
     connect(&_adsbVehicleCleanupTimer, &QTimer::timeout, this, &ADSBVehicleManager::_cleanupStaleVehicles);
     _adsbVehicleCleanupTimer.setSingleShot(false);
     _adsbVehicleCleanupTimer.start(1000);
@@ -42,7 +41,7 @@ void ADSBVehicleManager::_cleanupStaleVehicles()
     for (int i=_adsbVehicles.count()-1; i>=0; i--) {
         ADSBVehicle* adsbVehicle = _adsbVehicles.value<ADSBVehicle*>(i);
         if (adsbVehicle->expired()) {
-            qCDebug(ADSBVehicleManagerLog) << "Expired" << QStringLiteral("%1").arg(adsbVehicle->icaoAddress(), 0, 16);
+            qCDebug(ADSBVehicleManagerLog) << "Expired " << QStringLiteral("%1").arg(adsbVehicle->icaoAddress(), 0, 16);
             _adsbVehicles.removeAt(i);
             _adsbICAOMap.remove(adsbVehicle->icaoAddress());
             adsbVehicle->deleteLater();
@@ -50,7 +49,7 @@ void ADSBVehicleManager::_cleanupStaleVehicles()
     }
 }
 
-void ADSBVehicleManager::adsbVehicleUpdate(const ADSBVehicle::VehicleInfo_t vehicleInfo)
+void ADSBVehicleManager::adsbVehicleUpdate(const ADSBVehicle::ADSBVehicleInfo_t vehicleInfo)
 {
     uint32_t icaoAddress = vehicleInfo.icaoAddress;
 
@@ -61,6 +60,7 @@ void ADSBVehicleManager::adsbVehicleUpdate(const ADSBVehicle::VehicleInfo_t vehi
             ADSBVehicle* adsbVehicle = new ADSBVehicle(vehicleInfo, this);
             _adsbICAOMap[icaoAddress] = adsbVehicle;
             _adsbVehicles.append(adsbVehicle);
+            qCDebug(ADSBVehicleManagerLog) << "Added " << QStringLiteral("%1").arg(adsbVehicle->icaoAddress(), 0, 16);
         }
     }
 }
@@ -101,9 +101,7 @@ void ADSBTCPLink::run(void)
 void ADSBTCPLink::_hardwareConnect()
 {
     _socket = new QTcpSocket();
-
     QObject::connect(_socket, &QTcpSocket::readyRead, this, &ADSBTCPLink::_readBytes);
-
     _socket->connectToHost(_hostAddress, static_cast<quint16>(_port));
 
     // Give the socket a second to connect to the other side otherwise error out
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
Wow, sounds like a fantastic set of fixes. Thanks so much. I'll take a look at this some time today.

timtuxworth:
This is huge, thanks for the work zdanek. I now get vehicles showing up pretty much whenever my RadarBox shows traffic. The altitude matches what shows on the RadarBox.com website.
I wonder if you could show a different icon for helicopters? For using this for safety when flying a model, it's very helpful to know if the approaching vehicle is a helicopter or plane.
Airspeed isn't being displayed. On MP there is a mouseover with a field for Airspeed (it doesn't work), but the RadarBox website shows airspeed and this would also be helpful if possible.
The Callsigns don't always match the "official" airline flight number which is what shows on RadarBox. I'm not sure if it's that important, but an example I had a few minutes ago showed WEN3114 on QGC, but the actual flight number is WR3114 (from airline WEN).

timtuxworth:
More of a nice to have - could I put in the server name rather than IP address? If I could put in "raspberrypi.local" as the host address then things will keep working when I take my Radarbox out to the field and connect to a different wifi network (tether to my phone). I can change it no big deal, so like I say - nice to have.

zdanek:
@timtuxworth 
> I wonder if you could show a different icon for helicopters?
Based on ASDB only cannot be done because ADSB data doesn't contain aircraft type. 
It can be checked online  I suppose but it would require some kind of free service to check it. ICAO holds all the data but currently I did not find any open database. Need help.
AFAIK ICAO 24bit hex adddress can tell us directly the country of registration of an aircraft.

> Airspeed isn't being displayed. 
This is good idea. Even better - I saw on Flightradar24 that position is sometimes interpolated based on track and speed. If signal is lost aircraft is till moving and then correct position is fixed when real data comes in.
This is great idea for a Feature Request. You might create it and I could implement it in the future, aside from this PR which is a fix for current issues. Let's separate them.

> The Callsigns don't always match the "official" airline flight number
I cannot fix it. Callsign is as is in ADSB data.

> The timeout on displaying a vehicle might be a bit short
Losing callsign is because all vehicle info was removed. Then aircraft reappeared but from QGC's perspective it's a new one. Then you had to wait until Callsign MSG appeared (it comes very rarely).
We can increase again timeout but maybe simulating aircraft move (see my idea above) might be some kind of compromise. It would prolongate aircraft appearance on map. Still if no more ADSB data would not come it must be removed after some time.

>  could I put in the server name rather than IP address?
Good idea. Please create a feature request.

zdanek:
Ok, so I contacted people from open-sky.org and they are open to use their database 
https://opensky-network.org/data/datasets#d5
it's possible to map aircraft ICAO address to the type. This could be great feature :)

timtuxworth:
I put in a feature request for configurable timeout. https://github.com/mavlink/qgroundcontrol/issues/10171
I've put in a feature request to use server name not IP address https://github.com/mavlink/qgroundcontrol/issues/10172
I've put in a feature request to display different icons for different types of vehicles https://github.com/mavlink/qgroundcontrol/issues/10173

timtuxworth:
On the plane vs heli thing. My local club is 5k from a heliport. We often have helicopters hovering around, way more than planes.

DonLakeFlyer:
@zdanek Could you attach an example csv file here I can use to test QGC with your adsb tool?

timtuxworth:
Here's a log @DonLakeFlyer  https://www.dropbox.com/s/gov4fjc7htzhrlc/radarbox2022-02-16.csv?dl=0

zdanek:
> @zdanek Could you attach an example csv file here I can use to test QGC with your adsb tool?

And here's mine. Look at Warsaw, Poland :)
https://drive.google.com/file/d/1dd-A36_Ww6Gc8HRPRD8FCQ5Fi-iHpkH2/view

Mine data is a bit different than Tim's as I have different receiver. That's good as data might vary.

timtuxworth:
Tylko Lotem!

zdanek:
@timtuxworth 🤩🤩🤩

src/ADSB/ADSBVehicleManager.cc
```diff
@@ -121,72 +119,115 @@ void ADSBTCPLink::_hardwareConnect()
...
+void ADSBTCPLink::_parseAndEmitLocation(ADSBVehicle::ADSBVehicleInfo_t &adsbInfo, QStringList values)
+{
+    bool altOk, latOk, lonOk;
+    int modeCAltitude;
+
+    QString altitudeStr = values[11];
+    // Altitude is either Barometric - based on pressure, in ft
+    // or HAE - as reported by GPS - based on WGS84 Ellipsoid, in ft
+    // If altitude ends with H, we have HAE
+    if (altitudeStr.endsWith('H')) {
```
DonLakeFlyer, line 190: Extra return duplicates the error checking further down the code. Also better to not duplicate string conversion code. I would change as follows:
```
    QString altitudeStr = values[11];
    // Altitude is either Barometric - based on pressure, in ft
    // or HAE - as reported by GPS - based on WGS84 Ellipsoid, in ft
    // If altitude ends with H, we have HAE
    // There's a slight difference between Barometric alt and HAE, but it would require
    // knowledge about Geoid shape in particular Lat, Lon. It's not worth complicating the code
    if (altitudeStr.endsWith('H')) {
        altitudeStr.chop(1);
    }
    modeCAltitude = altitudeStr.toInt(&altOk);
```

DonLakeFlyer:
"First of all there was an error in the formula. To convert altitude in ft to m it should be multiplied by 0.3048. If you're into dividing then ft should be divided by 3.28 not by 3.048."

Interesting, I think Mission Planner has the same bug.

zdanek:
> Interesting, I think Mission Planner has the same bug.

@DonLakeFlyer ups!

https://en.wikipedia.org/wiki/Foot_(unit)

DonLakeFlyer (commented):
The rest all looks great. Good set of fixes.

DonLakeFlyer:
@meee1 Michael I think MP has the same conversion problem with feet to meters. Not sure if there is a problem for you with altitude values with an "H" on the end.

src/ADSB/ADSBVehicleManager.cc
zdanek: I wanted to bail out as soon as possible. But, sure, it reduces a few CPU ticks but shadows code readability. Clean code FTW.

DonLakeFlyer:
On the timeout change: I get what you are saying but I'm concerned with being too short in combination with making assumption on vehicle speeds. I would leave this alone for now.

The better change I think would be to keep the say 1 minute timeout which would cause the vehicle the go slight transparent on the display or something like that. That way you know contact has been lost. And then removed after a longer time. Do you know how often location update come through? Is it really only when the vehicle moves some amount of distance from the previous update? Or is is on sort sort of timed repetition no matter the movement? Anyway, fancier stuff like this doesn't need to be in this pull. And can be left for someone interested to make it better later.

zdanek:
> Do you know how often location update come through?

It's send quite often. Any timeouts are because of bad reception by the adsb hardware.
The idea with fading is cool. Maybe with conjunction on estimating new location based on speed and heading. Or not. To be treated as a new feature.

I propose to revert to 2 mins as we all used to. Then let's create a new feature req with an idea of fading and optionally movement estimation like Flightradar24 does.

WDYT?

DonLakeFlyer:
> I propose to revert to 2 mins as we all used to.

Yes, I think that is safer. Any new stuff can go into another pull when someone has time to implement.

zdanek:
Okay.
Timeout reverted to 120s.

timtuxworth:
> > Interesting, I think Mission Planner has the same bug.
> 
> @DonLakeFlyer ups!
> 
> https://en.wikipedia.org/wiki/Foot_(unit)

Mission Planner can't even handle the "H" altitudes - all I get is Alt: 0

DonLakeFlyer:
@RomanBapst This is all reviewed now and ready to go in

meee1:
> @meee1 Michael I think MP has the same conversion problem with feet to meters. Not sure if there is a problem for you with altitude values with an "H" on the end.

I did have the H issue, but ft to m is a non-issue

RomanBapst:
@DonLakeFlyer Not sure if you meant to tag me but I don't have any merge rights for QGC.

zdanek:
@DonLakeFlyer @RomanBapst ouch. I'm chilling my Champagne to celebrate my first PR into QGC so I just can't wait ;)

RomanBapst:
@bkueng Maybe you can merge this?

DonLakeFlyer:
I meant @mrpollo

zdanek:
@DonLakeFlyer I think all those guys can merge
https://github.com/orgs/mavlink/people

mrpollo:
Thanks for the great contribution @zdanek 🎉 

And also thanks to everyone who chimed in to help on the thread 💯

zdanek:
Thank for merging. It was a pleasure.

PR #10174, opened 2022-02-17
Support speed changes from Flyview

RomanBapst:
This introduces the ability to change multirotor cruise speed and fixed wing equivalent airspeed as part of the guided actions interface.

**What it does for hovering vehicles:**
The slider allows you to set the maximum cruising speed constrained to [0.1, MPC_XY_VEL_MAX]. If you e.g. do a GOTO you can specify the speed at which the vehicles moves to the target location.
You can also pause a mission and change the cruise speed. This will override the current cruise speed of the mission until the drone reaches a new waypoint which specifies a new cruise speed.
As mentioned, currently you do have to pause the mission to update the cruise speed, I image that this can be made a bit more convenient in the future.

**What it does for vehicles in forward flight (fixed wing)**
The slider allows to the set equivalent airspeed setpoints constrained to [FW_AIRSPEED_MIN, FW_AIRSPEED_MAX]

This video shows how this looks like when using PX4 as flight controller. Note that this requires changes on the PX4 side which are present in this PR: https://github.com/PX4/PX4-Autopilot/pull/18834 If you want to test this before it's merged, you need to test against that PR. The video also shows that the fixed wing airspeed parameters are not well setup as the vehicle starts to stall during the loiter at low airspeed.
https://www.youtube.com/watch?v=crJ9kn4S4N4

It would be good if folks from ardupilot could let me know if it makes sense to support this for them as well. I'm happy to help.


sfuhrer (commented):
Really cool! Tested it in simulation and couldn't get it to break. 
Just wondering if using Groundspeed and Airspeed consistently would be better (make it clearer what to expect from the command, especially when flying VTOLs), or if for Multicopter applications Cruise Speed is clearer, like it is currently proposed here.

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
```diff
@@ -410,6 +410,39 @@ void PX4FirmwarePlugin::guidedModeTakeoff(Vehicle* vehicle, double takeoffAltRel
         static_cast<float>(takeoffAltAMSL));    // AMSL altitude
 }
 
+double PX4FirmwarePlugin::maximumHorizontalSpeedMultirotor(Vehicle* vehicle)
+{
+    QString speedParam("MPC_XY_VEL_MAX");
+
+    if (vehicle->parameterManager()->parameterExists(FactSystem::defaultComponentId, speedParam)) {
+        return vehicle->parameterManager()->getParameter(FactSystem::defaultComponentId, speedParam)->rawValue().toDouble();
+    } else {
```
patrickelectric, line 419: else is unnecessary after return.

julled:
I really like this feature! Especially as this generalizes the code towards a generic "GuidedValueSliders".
According to Ardupilot it very possible to change the Air or Groundspeed via MAVLINK:
https://ardupilot.org/plane/docs/common-mavlink-mission-command-messages-mav_cmd.html#mav-cmd-do-change-speed

We would very love to get those changes upstream for our ardupilot based plane usecase.
@booo

src/FlightDisplay/GuidedActionsController.qml
```diff
@@ -452,6 +486,13 @@ Item {
         case actionActionList:
             actionList.show()
             return
+        case actionChangeSpeed:
+            confirmDialog.title = "Change Speed"
+            confirmDialog.hideTrigger = true
+            confirmDialog.title = changeSpeedTitle
```
patrickelectric, line 492: Title is changing twice.

patrickelectric (changes requested):

mrpollo:
@peterbarker hey do you think this is helpful for APM users as well?

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
```diff
@@ -526,6 +559,34 @@ void PX4FirmwarePlugin::guidedModeChangeAltitude(Vehicle* vehicle, double altitu
     }
 }
 
+void PX4FirmwarePlugin::guidedModeChangeGroundSpeed(Vehicle* vehicle, double groundspeed)
+{
+
+    vehicle->sendMavCommand(
```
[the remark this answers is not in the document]
julled: Nevertheless it could be also implemented like you did as a specialization of the FirmwarePlugin class. I dont know hat is prefered by the maintainers.

peterbarker:
> @peterbarker hey do you think this is helpful for APM users as well?

If I'm understanding this PR correctly, then certainly.

One caveat:
   ArduPlane honours the airspeed/groundspeed parameter (param1) for gorundspeed/airspeed.
   ArduCopter honours the airspeed/groundspeed parameter for groundspeed only (at this time), but also supports ascent/descent rates.

This is kind of timely; I'm currently playing around with implementing a `MAV_CMD_DO_CHANGE_RADIUS` - useful for loitering planes.  Having a slider for that would be kind of cool ;-)

src/FirmwarePlugin/FirmwarePlugin.h
```diff
@@ -150,6 +150,15 @@ class FirmwarePlugin : public QObject
...
     virtual double minimumTakeoffAltitude(Vehicle* /*vehicle*/) { return 10; }
 
+    /// @return The maximum horizontal groundspeed for a multirotor.
+    virtual double maximumHorizontalSpeedMultirotor(Vehicle* /*vehicle*/) { return 10; }
+
+    /// @return The maximum equivalent airspeed setpoint.
+    virtual double maximumEquivalentAirspeed(Vehicle* /*vehicle*/) { return 25; }
+
+    /// @return The minimum equivalent airspeed setpoint
+    virtual double minimumEquivalentAirspeed(Vehicle* /*vehicle*/) { return 12; }
```
peterbarker, line 160: Should these be pure-virtual?  Given that this is a fall-out-of-sky value if you get it wrong, defaults don't seem the right way to do things...

RomanBapst:
@peterbarker 

> One caveat:
ArduPlane honours the airspeed/groundspeed parameter (param1) for gorundspeed/airspeed.
ArduCopter honours the airspeed/groundspeed parameter for groundspeed only (at this time), but also supports ascent/descent rates.

Those are pretty much the things I'd need to know in order to go ahead and also make this available to you guys.
I have not idea how this works. For now speed changes will not be available but once it's clear how this should be handled then supporting APM will not be I problem (I hope).

> This is kind of timely; I'm currently playing around with implementing a MAV_CMD_DO_CHANGE_RADIUS - useful for loitering planes. Having a slider for that would be kind of cool ;-)

In PX4 you can use orbit commands and QGC allows you to drag the radius on the map. If you want to change the radius then you can set another orbit and place it on top of the old one -> it's probably not the most user friendly thing but works fine.
But yes, give that this PR refactors the sliders to be generic you could do all sorts of things with it.

image

src/FirmwarePlugin/FirmwarePlugin.h
RomanBapst: @peterbarker If I make them pure then I will need an implementation for Ardupilot, which I think is anyway the goal.
Could you please tell me how I get these values from Ardupilot. I assume it's parameters.

I'm actually wondering if it would make more sense to have the base function return negative values which indicates that the values are not available / not valid.
Because even in the case where you get these vales from parameters, if the param fetch fails then I'd like to return invalid. In that case QGC could take action and disable the speed change button.

peterbarker: > @peterbarker If I make them pure then I will need an implementation for Ardupilot, which I think is anyway the goal. 

Yep, understood.  This shouldn't be a blocker for your PR here, 'though, so maybe mark being `pure-virtual` down as "aspirational" :-)

> Could you please tell me how I get these values from Ardupilot. I assume it's parameters.

Sadly it's a bit more complicated than that.

Plane has `ARSPD_FBW_MIN` and `ARSPD_FBW_MAX`.  There's no maximum groundspeed, but there is a minimum groundspeed in `MIN_GNDSPD_CM`

Copter doesn't have a maximum speed while doing waypoint navigation; it's inferred from the (parameterised) maximum circular angle of the vehicle.  Given it's less important than the fixed-wing numbers, probably easiest to just choose a couple of constants for ArduPilot for now.  10m/s is way too slow, 'though :-)  30m/s would probably make for a reasonable upper bound guestimate.  Not suitable for some, but higher and your UI is going to suffer (almost need expo on the slider in this case!).

> I'm actually wondering if it would make more sense to have the base function return negative values which indicates that the values are not available / not valid. Because even in the case where you get these vales from parameters, if the param fetch fails then I'd like to return invalid. In that case QGC could take action and disable the speed change button.

Seems reasonable.  That also means that if the firmware behaviour changes out from underneath QGC it's going to do something sensible.  (My preference is for using NaNs in such cases; it makes it easier to catch when they get used in calculations when they shouldn't be - assuming QGC is set up for that).

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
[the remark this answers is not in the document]
RomanBapst: @julled That's a good point, thanks for pointing it out. I think if both flight stacks interpret the mavlink message the same way (which ideally they should) then your proposal makes sense and introduces less code changes.
If however, the mavlink messages need to be filled differently due to different interpretations then I think the plugin way is nicer.
It's not net clear how APM does it so I'm going to keep it like this for now but I'm open to change it if we decide to move forward with APM as well.

RomanBapst:
@patrickelectric @julled @peterbarker I've pushed some changes based on the review.

@peterbarker The base class of the firmware plugin now return NAN by default. I also conditioned the speed change action based on the availability of speed limits. For PX4 it checks if the params are available, for APM it will be something else.

I tested both PX4 and APM in SITL and I'm happy with the functionality.

patrickelectric (approved):

mrpollo:
🎉  thanks @RomanBapst

julled:
thanks @mrpollo  also from my side!

julled:
Where in this PR is the ardupilot specific mavlink message sending implemented, as only for px4 there are specific changes to the firmwareplugin?

PR #10186, opened 2022-02-26
use strcpy instead of strncpy

### LINKED ISSUE #10141: Memory overflow warning:  warning: ‘__builtin_strncpy’ writing 50 bytes into a region of size 24
This warning appears on every build on Linux. 
```
compiling moc_qtitem.cpp
linking staging/QGroundControl
In function ‘strncpy’,
    inlined from ‘_handleStatusText’ at /home/andre/prog/qgroundcontrol/src/Vehicle/Vehicle.cc:933:12:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:95:34: warning: ‘__builtin_strncpy’ writing 50 bytes into a region of size 24 overflows the destination [-Wstringop-overflow=]
   95 |   return __builtin___strncpy_chk (__dest, __src, __len,
      |                                  ^
/usr/include/x86_64-linux-gnu/bits/string_fortified.h: In member function ‘_handleStatusText’:
/home/andre/Qt/5.15.2/gcc_64/include/QtCore/qarraydata.h:129:8: note: destination object ‘<anonymous>’ of size 24
  129 | struct QTypedArrayData
      |        ^

```

acxz:
resolves stringop-overflow warning as at compile time b (QByteArray) is always 24, whereas the source is 50. 

~This means strncpy is not really useful in terms of checking length, (infact it is hurtful as it omits the null terminator~

fixes #10141

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -922,7 +922,6 @@ void Vehicle::_chunkedStatusTextCompleted(uint8_t compId)
 
 void Vehicle::_handleStatusText(mavlink_message_t& message)
 {
-    QByteArray  b;
     QString     messageText;
 
     mavlink_statustext_t statustext;
@@ -930,8 +929,7 @@ void Vehicle::_handleStatusText(mavlink_message_t& message)
 
     uint8_t compId = message.compid;
 
-    b.resize(MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN+1);
-    strncpy(b.data(), statustext.text, MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN);
+    QByteArray  b(statustext.text, MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN+1);
     b[b.length()-1] = '\0';
     messageText = QString(b);
     bool includesNullTerminator = messageText.length() < MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN;


src/Vehicle/Vehicle.cc
```diff
@@ -930,7 +930,7 @@ void Vehicle::_handleStatusText(mavlink_message_t& message)
     uint8_t compId = message.compid;
 
     b.resize(MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN+1);
-    strncpy(b.data(), statustext.text, MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN);
+    strcpy(b.data(), statustext.text);
```
bkueng, line 933: Rather use `memcpy`, as `statustext.text` might not be null-terminated.
The compiler warning is a false positive in this case, and `strncpy` is correct.

acxz: > Rather use memcpy, as statustext.text might not be null-terminated.

True, I'll change to use memcpy.

> strncpy is correct

I still don't understand why this is correct. strncpy does compile time length checks right? That is being violated and during runtime we are resizing. Which I believe renders using `strncpy` mute. Not to mention it will remove the null-terminator even if it exists.

acxz: Just compiled with memcpy and it also gives us the same warning. Maybe we should try to change how `b` is initialized and ensure that it is 50 bytes so we don't have to resize?

bkueng: > I still don't understand why this is correct. strncpy does compile time length checks right? 

It's a runtime-check and gcc has some extra checks at compile-time. Hence the compile-check fails but it's correct at runtime. The null termination is enforced on the next line.

> Just compiled with memcpy and it also gives us the same warning.

I thought it might.

> Maybe we should try to change how b is initialized and ensure that it is 50 bytes so we don't have to resize?

Yes you could directly use this one: https://doc.qt.io/qt-5/qbytearray.html#QByteArray-1

acxz: @bkueng I have updated the commit with your suggestion.

acxz:
@AndKe if you could test this PR that would be wonderful

acxz:
I can't seem to reproduce this with the latest gcc version: 1.12.1
so I'll go ahead and close this PR

AndKe:
sorry for the late reply - yes, this issue got resolved a while back.

PR #10187, opened 2022-02-26
allow follow to work with ardupilot copter and rover

tridge:
ardupilot follow mode has been stable for a long time, so enable
also added other missing flight modes


mrpollo:
Thanks @tridge

PR #10192, opened 2022-02-28
[Draft] Make the Log Viewer sortable by date

junwoo091400:
## Description
image
Currently, QGC's Log Viewer page doesn't have sort by date option. While doing a SITL tests, I discovered that the logs created don't have the id in order of the time. Which is annoying since it's hard to spot and download the 'latest' log.

I had this idea for a week, and after around 30 minutes of scanning, I couldn't figure out a concrete way to implement this (since QTableView doesn't support sorting via clicking the header by default). So any help / comment would be appreciated!

### DIFF
--- a/src/AnalyzeView/LogDownloadController.cc
+++ b/src/AnalyzeView/LogDownloadController.cc
@@ -765,6 +765,22 @@ QGCLogModel::clear(void)
 }
 
 //-----------------------------------------------------------------------------
+void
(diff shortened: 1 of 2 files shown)


zdanek:
Maybe a quick-win solution would be sorting all incoming logs by date ascending. Right after receiving, prior to displaying.
Clicking to sort would be a challenging task (or not, as I'm not QT expert) but sorting received data should be simple.

Issue #10178, opened 2022-02-21
QGC Crash When Saving Flight Mode with PX4 Master 1.13.0 Alpha


vincentpoont2:
## Expected Behavior
Flight mode should save, QGC should not crash.

## Current Behavior
QGroundControl Crash when saving flight mode settings with PX4 Master_1.13.0 Alpha

Tested with 
Rev1 Durandal (Support added here https://github.com/PX4/PX4-Autopilot/pull/18628) 
Pixhawk 4

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1.) Using Pixhawk 4 or Rev 1 Durandal Flight controller
2.) QGC (Tried Daily & 4.1.4)
3.) PX4 Master 1.13.0 Alpha
4.) Save flight mode settings 

Durandal with PX4 Master 1.13.0 Alpha
https://user-images.githubusercontent.com/46874772/154998452-af48217e-e0a6-4cf7-9d06-4f806723aa80.mp4
or https://youtu.be/xC99lJGRhtc

https://user-images.githubusercontent.com/46874772/154999205-9a6255f6-353a-44b4-be60-e3838788600b.mp4
or https://youtu.be/MyjieHO4evM

Durandal with PX4 Stable 1.12.3

No issue
**Video**

https://user-images.githubusercontent.com/46874772/155104822-8132b893-1256-4713-a914-d6c8952a73b4.mp4

## System Information
When posting bug reports, include the following information
- Operating System: Window 10
- QGC Version:  QGC Daily & 4.1.4
- QGC build: daily, stable 
- Flight Controller: Durandal Rev1, Pixhawk 4
- Autopilot (with version): PX4 Master 1.13.0 Alpha

vincentpoont2:
Any idea? @mrpollo

mrpollo:
Has this always been an issue on Durandal? if not in which PX4 version did this start?

vincentpoont2:
Reproduced on both Pixhawk 4 & Durandal with Master FW, does not happen with Stable PX4 1.12.3.

davids5:
I can verify this on PX4 master. QGC crashes on both linux and windows.
Linux:
[1561688.874184] QGroundControl[176238]: segfault at 28 ip 00007f1c0c009b10 sp 00007ffdb483f658 error 4 in libQt5Core.so.5[7f1c0bd27000+5e4000]
[...]

PR #10193, opened 2022-03-02
PX4: remove AdvancedFlightModes UI

bkueng:
Got deprecated in https://github.com/PX4/PX4-Autopilot/pull/16444, and led to a segfault when trying to be activated due to the removal of the param RC_ASSIST_TH.

Fixes https://github.com/mavlink/qgroundcontrol/issues/10178

@mrpollo


dagar (approved):

dagar:
Thanks @bkueng. I've been trying to keep track of QGC parameter dependencies in PX4.

https://github.com/PX4/PX4-Autopilot/blob/52221b0bb79b241dd4db3e2cba85d84bb521af70/src/modules/rc_update/rc_update.cpp#L157-L201

https://github.com/PX4/PX4-Autopilot/blob/52221b0bb79b241dd4db3e2cba85d84bb521af70/src/modules/mavlink/mavlink_parameters.cpp#L282-L306

We should do another pass of QGC itself to identify any remaining hard coded parameter access we haven't accounted for.

mrpollo:
Thank you @bkueng

PR #10196, opened 2022-03-03
update ubuntu to 20.04

acxz:
Addresses: https://github.com/mavlink/qgroundcontrol/pull/10153#discussion_r818868188

### DIFF
--- a/.github/workflows/linux_debug.yml
+++ b/.github/workflows/linux_debug.yml
@@ -19,7 +19,7 @@ env:
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


patrickelectric:
Be aware, that doing so will break people that use ubuntu 18.04+ and probably someone from the AppImage community will say something about it.

acxz:
I think the gst update can still be pushed off for a bit, tbh. Don't think we should break existing builds until 18 is EOL.

When creating this PR I hadn't realized AppImages from 20 don't work on 18.

PR #10211, opened 2022-03-14
Update SDL2 library

holden-zenith:
Allows use of new controllers like Sony Dualsense


holden-zenith:
@mrpollo we need the compiled dll

DonLakeFlyer:
Android and OSX also have SDL2 libraries in the repo. This should updated as well to match the windows version.

holden-zenith:
Android does not

mrpollo:
thanks @holden-zenith

PR #10219, opened 2022-03-22
Unify android deployment yamls

Williangalvani:
depends on #10208


mrpollo:
Great work @Williangalvani thanks for streamlining the workflow!

PR #10228, opened 2022-03-28
Fix video record with mavlink camera

patrickelectric:
…nk camera

Mavlink camera can or not record the video, if not, record it locally


ES-Alexander:
Works well (fixes the problem), although once the MAVLink stream loads the video stream controls come up with a timestamp that's just fixed at 00:00:00, which is a bit unintuitive form an interface perspective. That's not an issue that's specific to this PR though, so probably worth merging this and just raising that as a separate issue.

ES-Alexander:
@Williangalvani mentioned this is already fixed (differently) in `master`, and should likely be cherry-picked from https://github.com/mavlink/qgroundcontrol/commit/732c9953b2f5b7da1b81d79e76e38284762aad93

PR #10231, opened 2022-03-29
Fix seriallink error signal connection

holden-zenith:
QSerialPort::error signal is deprecated, use old signal slot syntax because there is both a signal and function named QSerialPort::error

### DIFF
--- a/src/comm/SerialLink.cc
+++ b/src/comm/SerialLink.cc
@@ -177,7 +177,11 @@ bool SerialLink::_hardwareConnect(QSerialPort::SerialPortError& error, QString& 
 
     _port = new QSerialPort(_serialConfig->portName(), this);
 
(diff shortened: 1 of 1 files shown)


src/comm/SerialLink.cc
```diff
@@ -177,7 +177,11 @@ bool SerialLink::_hardwareConnect(QSerialPort::SerialPortError& error, QString&
 
     _port = new QSerialPort(_serialConfig->portName(), this);
 
-    QObject::connect(_port, static_cast<void (QSerialPort::*)(QSerialPort::SerialPortError)>(&QSerialPort::error), this, &SerialLink::linkError);
+#ifdef __android__
```
patrickelectric, line 180: This test should use the Qt version and not the build target

patrickelectric (changes requested):

PR #10235, opened 2022-03-29
sport mode is deprecated

holden-zenith:
> [o) Sport mode deprecated](https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/ReleaseNotes.txt)


src/FirmwarePlugin/APM/ArduCopterFirmwarePlugin.cc
```diff
@@ -64,7 +63,6 @@ ArduCopterFirmwarePlugin::ArduCopterFirmwarePlugin(void)
         APMCopterMode(APMCopterMode::CIRCLE,        true),
         APMCopterMode(APMCopterMode::LAND,          true),
         APMCopterMode(APMCopterMode::DRIFT,         true),
-        APMCopterMode(APMCopterMode::SPORT,         true),
```
patrickelectric, line 67: I believe that you should check the FW version and see if this mode is still valid.

PR #10239, opened 2022-04-01
Add Reboot button to the Main Toolbar for easy access

junwoo091400:
## Problem Statement
Previously, to reboot a vehicle user had to:
1. Click Q icon
2. Click 'Vehicle Setup'
3. Click 'Parameters' all the way down in the left side of the panel
4. Click 'Tools' tab on the right-up side of the screen
5. Click 'Reboot Vehicle' button

This PR exposes that button to the Main Screen so that user has to:
1. Click 'Reboot vehicle' Button.

## Benefits
It is really useful to be able to reboot the vehicle efficiently and quickly. So this PR will help a lot of people in my opinion!

RebootButton_Screenshot from 2022-04-01 18-19-22

### DIFF
--- a/src/ui/toolbar/MainToolBar.qml
+++ b/src/ui/toolbar/MainToolBar.qml
@@ -107,9 +107,26 @@ Rectangle {
         }
     }
 
+    //-- Reboot Button
+    RowLayout {
+        anchors.right:          brandingLogo.left
+        anchors.top:            parent.top
+        anchors.bottom:         parent.bottom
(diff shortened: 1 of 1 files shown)


WTPENGUIN:
It's definitely a convenient feature. But an easy approach can lead to frequent mistakes. It seems like you need to think more about safety measures such as disabling the button in the arming state.

junwoo091400:
Replaced by #10244

PR #10244, opened 2022-04-05
Add Reboot vehicle button in Main Window tool bar & Update Firmware update tab

### LINKED ISSUE #10242: Feature-Proposal] Remove the need to unplug cable from Pixhawk when uploading firmware
## Problem
image
**Currently, the QGC forces the user to unplug the Pixhawk from the QGC Device and re-plug to upload the software** (it catches at the bootloader stage, then enters firmware update mode). This is quite annoying at times since the cable is not always so accessible & it just adds one more step to flashing a new firmware.

## Proposal
We can 'reboot' the vehicle and then go into the firmware update mode, as far as I understand. That wouldn't require user to mess with the cables!

junwoo091400:
## Need
When flashing a firmware, it was tedious to remove the cable every time (since QGC didn't allow rebooting the vehicle in the Firmware tab). The ability to simply reboot & go into bootloader mode was needed.

## Suggested change
image
- Added 'Reboot Vehicle' button that shows up in Main Toolbar (in all Vehicle setup / Analyze Tools / Application Settings)
- Changed messages in Firmware tab to guide user to use this button to upload new firmware (without the need to unplug & plug the USB cable)

This Resolves #10242


ryanjAA:
Great! Will test it out. Thanks! This will be much better.

patrickelectric:
@ryanjAA did you test it ?

patrickelectric:
@junwoo091400 can you break it in multiple commits ? I see that you have added a reboot button but there are some comments, code organization and modifications that does not reflect what te single commit message says.

junwoo091400:
@patrickelectric Commit separating done. Thanks for the feedback :wink:

junwoo091400:
Why was this closed?

junwoo091400:
This replaces #10239

PR #10250, opened 2022-04-08
RTK GPS feature enabled for Android builds.

mnaumov:
Hello!
I have got RTK working on Android devices. Radio-link and RTK device connected to phone via tiny USB switch. It works correctly and QGC sends RTK messages to the vehicle.
The only problem I could not solve:
Case 1 (fail):
Steps to reproduce: Both radio-link and RTK devices are connected. Disconnect RTK device.
Result (problem): Radio-link disconnects. Re-plugging radio-link does not help, it is not visible in the list of connected devices anymore. Only re-plugging USB-switch allows to connect USB devices again.

Case 2 (pass):
Steps to reproduce: Both radio-link and RTK devices are connected. Re-plug radio-link.
Result: Radio-link connected successfully.

Tested on Samsung A32 only.

Conclusion: I suspect CdcAcmSerialDriver closes USB connection in a wrong way (may be it closes not dedicated USB RTK-device, but USB-switch connection). I'm a newbie at Android programming, hope community can help.

Changes performed:
1) Multiple USB devices connected info. _QGCActivity::availableDevicesInfo()_
2) Every _UsbIoManager_ instance for every particular connected device now runs in a separate thread. _UsbIoManager::start()_ added.
3) _QGCActivity::write()_ moved to async mode.
4) _GPSProvider::callback::writeDeviceData_ . Disable _serial->waitForBytesWritten_ because it is not supported currently for Android.

Please review my changes and send me your questions, objections, comments.

Thank you,
Michael Naumov.

### DIFF
--- a/android/src/org/mavlink/qgroundcontrol/QGCActivity.java
+++ b/android/src/org/mavlink/qgroundcontrol/QGCActivity.java
@@ -81,7 +81,6 @@ public class QGCActivity extends QtActivity
     
     public static Context m_context;
 
-    private final static ExecutorService m_Executor = Executors.newSingleThreadExecutor();
 
     private final static UsbIoManager.Listener m_Listener =
             new UsbIoManager.Listener()
@@ -357,7 +356,8 @@ public class QGCActivity extends QtActivity
             String          deviceInfo;
             UsbSerialDriver driver = _drivers.get(i);
 
-            if (driver.permissionStatus() != UsbSerialDriver.permissionStatusSuccess) {
+            if ((driver.permissionStatus() != UsbSerialDriver.permissionStatusSuccess)
+             && (driver.permissionStatus() != UsbSerialDriver.permissionStatusOpen)) {
                 continue;
             }
 
@@ -423,7 +423,7 @@ public class QGCActivity extends QtActivity
 
             UsbIoManager ioManager = new UsbIoManager(driver, m_Listener, userData);
             m_ioManager.put(deviceId, ioManager);
-            m_Executor.submit(ioManager);
+            ioManager.start();
 
             qgcLogDebug("Port open successful");
         } catch(IOException exA) {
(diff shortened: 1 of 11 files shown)


zdanek:
@mnaumov I was doing the same as I belive that RTK GPS can work on Android. I can join you in the efforts.

zdanek:
@mnaumov see my (unfinished) work at [repo ](https://github.com/zdanek/qgroundcontrol/tree/rtk_on_mobile)
I did all changes enabling android only. You've enabled all mobile that means IOS too. In my opinion this is wrong as IOS propably has different serial port architecture / behaviour. 
Also I don't have IOS so I cannot be sure if it works. It's safe to enable RTK on Adroid only.

mnaumov:
@zdanek [zdanek](https://github.com/zdanek)
Hello Zranek! Thank you for reply. I agree my changes should be enabled for Android only.
Did you have a chance to test my changes related to multiple USB devices on Android?

mrpollo:
is this PR ready to come in @mnaumov ?

mnaumov:
@mrpollo Hello Roman! I have tested my changes at real survey missions twice. I'm sure it would be useful.

zdanek:
@mnaumov did you check PC build again with RTK, so we are sure it did not break? Does it still work on PC?
Can you please confirm on which OS. I have Ubuntu. Whick OS do you have?

DonLakeFlyer:
> Case 1 (fail):
> Steps to reproduce: Both radio-link and RTK devices are connected. Disconnect RTK device.
> Result (problem): Radio-link disconnects. Re-plugging radio-link does not help, it is not visible in the list of connected devices anymore. Only re-plugging USB-switch allows to connect USB devices again.

The android code for USB does not expect to support more than one USB device. This expectation is carried through in multiple parts of the code. It "might" be able to be changed to support multiple USB devices. But given how difficult it's been to get the existing code to work correctly with a single USB telemetry device I'm not super interested in possible breaking that for RTK support. At least not without some massive amount of testing and prove that the problem shown in Case 1 has been fixed.

Given all that. Closing until a much more details fix could come through.

Issue #10278, opened 2022-05-04
TradHeli Setup Page for Copter 4.2


bnsgeyer:
ArduCopter 4.2 changes some Tradheli specific parameters and adds a few new ones.  Hoping someone can help me get these changes into the next release of QgroundControl in order to support our Users with the release of ArduCopter 4.2

This will affect Single Heli, Dual Heli and Quad Heli frame types.

Here is a picture that shows the changes to be made.
![image](https://user-images.githubusercontent.com/9907906/166618660-efd73e0f-43af-48c1-ac41-8706aec609f4.png)

Thanks in advance
Bill Geyer 
Tradheli Maintainer

PR #10285, opened 2022-05-11
Tradheli- update heli page for ArduCopter 4.2

bnsgeyer:
This PR addresses issue #10278.  I have never modified this code so I would appreciate guidance on what else had to be changed to support this PR. When will the 4.2 parameters be added to ensure my reference to the assigned variable names are correct.


bnsgeyer:
@DonLakeFlyer or @patrickelectric I would appreciate any assistance you could provide in finishing this PR so it would be ready for the release of arducopter 4.2.  Thanks in Advance!
Bill

mrpollo:
@bnsgeyer thanks for the PR

bnsgeyer:
@mrpollo thanks for merging this. Did you do a functional check on it. I don’t have a way to check it to make sure the new Params worked correctly. I was really just guessing as to what needed changed to make this PR.

PR #10288, opened 2022-05-12
Convert shapelib to submodule

holden-zenith:


patrickelectric:
Thanks @holden-zenith!

PR #10297, opened 2022-05-19
Update Pixhawk FMUv6C vendor id

mrpollo:
We use the generic vendor id available on the bootloader, which is set to Auterion. We plan to change that to either Dronecode or Pixhawk, we realize it's not ideal, but it will have to do for now until we get a USB Vendor ID approved

cc @vincentpoont2 @davids5


mrpollo:
Please test @vincentpoont2

vincentpoont2:
Tested. Please merge.

mrpollo:
Merged, failing tests had unrelated issue documented here #10311

PR #10300, opened 2022-05-23
solve the crash caused by map

jingwenyi:
when I tested qgc in windwos and android , I found a crash. The repeated step is that my copter link qgc by  tcp, then I enter the offlinemap setting , I zoom in and out frequently, qgc quickly crash. I use windbg check the bug, I find the error is "Qt5Network!QNetworkAccessManager", then I find that the code of "_reply->abort()" leaded the crash. I  think we can delete the row.


HTRamsey:
Actually the whole timer/timeout portion needs to be replaced with QNetworkReply TimeoutError in networkReplyError. Having the separate timer is causing duplicate work and possibly leading to some issues.

PR #10327, opened 2022-06-23
Adsb: remove pointer to deleted object

potaito:
There is an illegal pointer access in QGC related to ADSB.

From the QGC code:
- `ADSBVehicleManager` contains a list of all `ADSBVehicle` objects that were allocated
- `ADSBVehicleManager` furthermore holds a key:value map with pointers to the `ADSBVehicle` objects

Once a second `ADSBVehicleManager` checks whether any stored vehicle information is expired. This happens after no information was received for 120 seconds. The `ADSBVehicle` object is then marked for deletion with Qt's `deleteLater()` function. 

However, the map containing the pointer to this object was not groomed. If the airplane now re-enters the area, QGC will access the pointer of an object that was possibly deleted:

This can lead to a segfault

### DIFF
--- a/src/ADSB/ADSBVehicleManager.cc
+++ b/src/ADSB/ADSBVehicleManager.cc
@@ -43,6 +43,10 @@ void ADSBVehicleManager::_cleanupStaleVehicles()
         if (adsbVehicle->expired()) {
             qCDebug(ADSBVehicleManagerLog) << "Expired " << QStringLiteral("%1").arg(adsbVehicle->icaoAddress(), 0, 16);
             _adsbVehicles.removeAt(i);
+
+            uint32_t key = adsbVehicle->icaoAddress();
+            _adsbICAOMap.remove(key);
+
             _adsbICAOMap.remove(adsbVehicle->icaoAddress());
             adsbVehicle->deleteLater();
         }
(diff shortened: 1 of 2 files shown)


src/ADSB/ADSBVehicleManager.cc
```diff
@@ -43,6 +43,10 @@ void ADSBVehicleManager::_cleanupStaleVehicles()
         if (adsbVehicle->expired()) {
             qCDebug(ADSBVehicleManagerLog) << "Expired " << QStringLiteral("%1").arg(adsbVehicle->icaoAddress(), 0, 16);
             _adsbVehicles.removeAt(i);
+
+            uint32_t key = adsbVehicle->icaoAddress();
```
zdanek, line 47: I don't get it.
The same code is in single line 50. :thinking:

potaito: Omg. I swear when I looked I didn't see that. The fix was added 2 years ago: https://github.com/mavlink/qgroundcontrol/commit/30c227eb1a88a6df3ccd328b443d712515c84303

Sorry. This can be closed then :roll_eyes:

potaito:
Already fixed in https://github.com/mavlink/qgroundcontrol/pull/8881
sorry

Issue #10175, opened 2022-02-17
Joystick setAxisFrequency and setButtonFrequency validation seems to be wrong


vesseshhebbar:
## Expected Behavior
When we set Joystick frequency in the UI, it should accept values between Min and Max values

## Current Behavior
When we set Joystick frequency, the UI automatically changes the frequency to the Max frequency.

## Detailed Description
The logic for validating the value in 
Joystick::setAxisFrequency(float val)  and  Joystick::setButtonFrequency(float val)
seems to be wrong.

I was able to fix this bug by swapping qMin and qMax, and by disabling the validators in JoystickConfigAdvanced.qml:

#### Erroneous Code:
void Joystick::setAxisFrequency(float val)
    val = qMin(_minAxisFrequencyHz, val);
    val = qMax(_maxAxisFrequencyHz, val);

#### Correct Code:
    val = qMax(_minAxisFrequencyHz, val);
    val = qMin(_maxAxisFrequencyHz, val);

I also had to remove the validator lines in JoystickConfigAdvanced.qml for both axis and button:

// validator:          DoubleValidator { bottom: _activeJoystick.minAxisFrequencyHz; top: _activeJoystick.maxAxisFrequencyHz; }

PS: Since I'm new to pull requests and submitting patches, someone kindly fix the issue or please handhold me in this process

PR #10345, opened 2022-07-14
Fix joystick set*Frequency boundary checks. #10175

svpcom:
Joystick::setAxisFrequency and Joystick::setButtonFrequency has wrong boundary checks. Instead of boundary check value will set to maximum allowed value.
This was referenced in issues #10175


bkueng (approved):
Looks correct

Issue #10188, opened 2022-02-26
warning: cast between incompatible function types


acxz:
```
/home/acxz/vcs/git/github/acxz/qgroundcontrol/src/VideoReceiver/gstqgcvideosinkbin.c: In function ‘_vsb_get_type’:
/home/acxz/vcs/git/github/acxz/qgroundcontrol/src/VideoReceiver/gstqgcvideosinkbin.c:314:13: warning: cast between incompatible function types from ‘void (*)(GstQgcVideoSinkBinClass *)’ {aka ‘void (*)(struct _GstQgcVideoSinkBinClass *)’} to ‘void (*)(void *, void *)’ [-Wcast-function-type]
  314 |             (GClassInitFunc)_vsb_class_init,
      |             ^
/home/acxz/vcs/git/github/acxz/qgroundcontrol/src/VideoReceiver/gstqgcvideosinkbin.c:319:13: warning: cast between incompatible function types from ‘void (*)(GstQgcVideoSinkBin *)’ {aka ‘void (*)(struct _GstQgcVideoSinkBin *)’} to ‘void (*)(GTypeInstance *, void *)’ {aka ‘void (*)(struct _GTypeInstance *, void *)’} [-Wcast-function-type]
  319 |             (GInstanceInitFunc)_vsb_init,
      |             ^
```

gcc: 11.2.0

acxz:
Closed by #10359

PR #10359, opened 2022-08-06
perform explicit void * cast

acxz:
Solves #10188 by explicitly casting to `void *`.

### DIFF
--- a/src/VideoReceiver/gstqgcvideosinkbin.c
+++ b/src/VideoReceiver/gstqgcvideosinkbin.c
@@ -65,7 +65,7 @@ enum {
 
 static GstBinClass *parent_class;
 
-static void _vsb_init(GstQgcVideoSinkBin *vsb);
+static void _vsb_init(GTypeInstance *instanceData, void *vsbVoid);
 static void _vsb_dispose(GObject *object);
 static void _vsb_get_property(GObject *object, guint prop_id, GValue *value, GParamSpec *pspec);
 static void _vsb_set_property(GObject *object, guint prop_id, const GValue *value, GParamSpec *pspec);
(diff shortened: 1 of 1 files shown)


src/VideoReceiver/gstqgcvideosinkbin.c
```diff
@@ -311,12 +311,12 @@ _vsb_get_type(void)
             sizeof(GstQgcVideoSinkBinClass),
             NULL,
             NULL,
-            (GClassInitFunc)_vsb_class_init,
+            (GClassInitFunc)(void *)_vsb_class_init,
```
bkueng, line 314: I would rather change the method API to `void *, void *` so we don't have to cast here, and do the necessary cast in the method.

acxz: Can you be a bit specific as to what method you are referring to?

Are you referring to `g_type_register_static`? If so I don't think we can since that is part of a dependency and not code that we have written ourselves.

bkueng: The `_vsb_class_init` method. Both arguments are expected to be `void*` pointers.

acxz: @bkueng for `_vsb_class_init` I believe it only has one argument right?

I agree though, I have changed the method API to ingest a void * rather than doing a preliminary cast before calling the method.

bkueng: Yes you need to add a 2. argument, `void* classData`. Then use `Q_UNUSED(classData);` in the method to suppress the unused warning.

acxz: Ah til, `Q_UNUSED`. Thanks for the pointers, @bkueng.

I have added your suggestions and this PR should be good to go now.

src/VideoReceiver/gstqgcvideosinkbin.c
```diff
@@ -316,7 +317,7 @@ _vsb_get_type(void)
             NULL,
             sizeof(GstQgcVideoSinkBin),
             0,
-            (GInstanceInitFunc)_vsb_init,
+            (GInstanceInitFunc)(void *)_vsb_init,
```
bkueng, line 320: Can you do the same here as well and remove the void* cast?

acxz: Sure thing! Done.

acxz:
Thanks for walking me through this PR, learned a couple things along the way!

PR #10361, opened 2022-08-09
Rework fly view map guided mode menu

s-lisovenko:
This PR is designed to make the guided mode menu a little more beautiful and unified in style with QGC

Old style menu (Go to location and Orbit at location):
Screenshot from 2022-08-09 16-47-51

New style menu:
Screenshot from 2022-08-09 16-42-53


bkueng (approved):
Nice

PR #10365, opened 2022-08-11
Rework flight mode menu for toolbar

s-lisovenko:
Add a flight mode menu indicator that matches the QGC style

Old menu style:
Screenshot from 2022-08-11 16-15-58

New menu style:
Screenshot from 2022-08-11 16-05-09


bkueng:
Did you test this on mobile as well? It might use up a bit much space.

s-lisovenko:
> Did you test this on mobile as well? It might use up a bit much space.
No, unfortunately I don't have such an opportunity. But if the screen is small, then this menu will simply scroll.
But, of course, a test on a real device would be a very big plus.

Below is a maximally reduced screen for GCC on Linux:
Screenshot from 2022-08-12 14-12-45

bkueng (approved):
Alright, should be fine then.

PR #10373, opened 2022-08-19
main: Allow user to open QGC

patrickelectric:
Some users report that our check does work

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -248,11 +248,14 @@ int main(int argc, char *argv[])
     RunGuard guard(runguardString);
     if (!guard.tryToRun()) {
         // QApplication is necessary to use QMessageBox
-        QApplication errorApp(argc, argv);
-        QMessageBox::critical(nullptr, QObject::tr("Error"),
-            QObject::tr("A second instance of %1 is already running. Please close the other instance and try again.").arg(QGC_APPLICATION_NAME)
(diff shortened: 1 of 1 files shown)


bkueng:
Why would you want to allow this? There's multiple files like caches that will be accessed by both instances.

patrickelectric:
@bkueng check: https://px4.slack.com/archives/C629RE9H7/p1661268112226079

bkueng (changes requested):
Ok this isn't the right way to resolve this then. We need to find the root cause.

patrickelectric:
@bkueng @mrpollo, so, there is the tradeoff of not allowing users to open QGC under windows 11 or merging the temporary fix until we find the root of the problem.

bkueng:
You can also check if it's running on windows 11. But that's still a pretty bad user experience.

junwoo091400:
> @bkueng check: https://px4.slack.com/archives/C629RE9H7/p1661268112226079

Screenshot of the discussion archived here, as it will get removed soon

image

booo:
What do we do about this change?

I second that we should find the root cause rather than fixing the symptoms.

booo:
I will close this one for now because of lack of activity and because two people think it's not the right way to resolve the problem.

Please reopen if you think this should be merged e.g. temporary as fix.

PR #10386, opened 2022-08-31
Vehicle: update HealthAndArmingCheck

bkueng:
From https://github.com/mavlink/libevents/pull/3

### DIFF
--- a/src/Vehicle/EventHandler.h
+++ b/src/Vehicle/EventHandler.h
@@ -38,6 +38,7 @@ public:
     void setMetadata(const QString& metadataJsonFileName);
 
     const events::HealthAndArmingChecks::Results& healthAndArmingCheckResults() const { return _healthAndArmingChecks.results(); }
+    bool healthAndArmingCheckResultsValid() const { return _healthAndArmingChecksValid; }
 
     int getModeGroup(int32_t customMode);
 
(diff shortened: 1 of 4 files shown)


src/Vehicle/EventHandler.h
```diff
@@ -56,6 +57,7 @@ class EventHandler : public QObject
     QTimer _timer;
     events::parser::Parser _parser;
     events::HealthAndArmingChecks _healthAndArmingChecks;
+    bool _healthAndArmingChecksValid{false};
```
patrickelectric, line 60: The variable value should be set on constructor initializer list, as the current class style.

patrickelectric (changes requested):
The update on libevents could be done in a separated commit (and in preference in a separated PR), an libevents update is necessary for this changes ?

bkueng:
> The update on libevents could be done in a separated commit (and in preference in a separated PR), an libevents update is necessary for this changes ?

Yes it's an API change. I separated the commits

src/Vehicle/EventHandler.h
bkueng: It is consistent, only members initialized from constructor arguments use the initializer list.

bkueng:
Merging as it's needed for PX4. @patrickelectric let me know if you want to discuss something further.

PR #10389, opened 2022-09-02
src: Improve debug messages on the camera stream logic

patrickelectric:


booo:
Looks good to me.

Maybe the prefix of the commit message should be "QGCCameraControl:" and not "src".

PR #10380, opened 2022-08-28
fix Android CI for 32bit and 64bit builds (#10360)

fredowski:
* install and use NDK r21e
  * update QT install action from v2 to v3

Starting with NDK Version r24 llvm will not accept one-dash
long options anymore. This is bug QTBUG-104580

https://bugreports.qt.io/browse/QTBUG-104580

The r23c NDK version did not work due to issues with
stripping symbols. So I install and use r21e. QT 5.15.2 uses
gradle version 5.6.4 with the default android gradle plugin 3.6.0.
With gradle 6.9.2 and android gradle plugin 4.2.2 the 32bits
build worked. NDK r22b is not compatible with gstreamer 1.18 and
results in a failing 64bit build. Therefore r21e is required for
gstreamer support.

Closes: https://github.com/mavlink/qgroundcontrol/issues/10360

Issue #10378, opened 2022-08-23
After Commit: 3f479fc Video is not working


AndreySkyFoxSidorov:
After Commit: 3f479fc Video is not working. Platform Windows/Android

Log:
(QGroundControl.exe:20868): GLib-GObject-WARNING **: 15:45:44.989: invalid uninstantiatable type '<unknown>' in cast to 'GstBin'
(QGroundControl.exe:20868): GStreamer-CRITICAL **: 15:45:44.994: gst_bin_add_many: assertion 'GST_IS_BIN (bin)' failed
(QGroundControl.exe:20868): GLib-GObject-WARNING **: 15:45:44.998: invalid uninstantiatable type '<unknown>' in cast to 'GstElement'
(QGroundControl.exe:20868): GStreamer-CRITICAL **: 15:45:44.999: gst_element_add_pad: assertion 'GST_IS_ELEMENT (element)' failed
(QGroundControl.exe:20868): GLib-GObject-CRITICAL **: 15:45:44.999: g_object_set: assertion 'G_IS_OBJECT (object)' failed
VideoReceiverLog:
(QGroundControl.exe:20868): GLib-GObject-WARNING **: 15:45:45.001: invalid uninstantiatable type 'void' in cast to 'GstBin'
 Unable to find sin
(QGroundControl.exe:20868): GStreamer-CRITICAL **: 15:45:45.002: gst_bin_add_many: assertion 'GST_IS_BIN (bin)' failed
k pad of video sink "rtsp://192.168.0.40:8554/main.265"
(QGroundControl.exe:20868): GLib-GObject-WARNING **: 15:45:45.007: invalid uninstantiatable type 'void' in cast to 'GstElement'
(QGroundControl.exe:20868): GStreamer-CRITICAL **: 15:45:45.012: gst_element_add_pad: assertion 'GST_IS_ELEMENT (element)' failed
(QGroundControl.exe:20868): GLib-GObject-CRITICAL **: 15:45:45.013: g_object_set: assertion 'G_IS_OBJECT (object)' failed

fredowski:
Does that mean that Video is not displayed in the QGC app? I try to fix the Android CI in PR https://github.com/mavlink/qgroundcontrol/pull/10380#issuecomment-1234784608 but the Android QGC app does not display the video stream.

fredowski:
This is fixed with 9b63448fc0631.

PR #10392, opened 2022-09-04
Fix GInstanceInitFunc for register GtypeIfo

s-lisovenko:
I found out that video streaming stopped working after this commit 3f479fc8410d3422dc4e27208b17eebcdb0bd617.
The resulting errors are below:

(QGroundControl:238965): GLib-GObject-WARNING **: 15:02:27.061: invalid uninstantiatable type '<unknown>' in cast to 'GstBin'

(QGroundControl:238965): GStreamer-CRITICAL **: 15:02:27.061: gst_bin_add_many: assertion 'GST_IS_BIN (bin)' failed

(QGroundControl:238965): GLib-GObject-WARNING **: 15:02:27.064: invalid uninstantiatable type '<unknown>' in cast to 'GstElement'

(QGroundControl:238965): GStreamer-CRITICAL **: 15:02:27.064: gst_element_add_pad: assertion 'GST_IS_ELEMENT (element)' failed

(QGroundControl:238965): GLib-GObject-CRITICAL **: 15:02:27.064: g_object_set: assertion 'G_IS_OBJECT (object)' failed
VideoReceiverLog: Unable to find sink pad of video sink "udp://0.0.0.0:4545"

(QGroundControl:238965): GLib-GObject-WARNING **: 15:02:27.065: invalid uninstantiatable type 'void' in cast to 'GstBin'

(QGroundControl:238965): GStreamer-CRITICAL **: 15:02:27.065: gst_bin_add_many: assertion 'GST_IS_BIN (bin)' failed

(QGroundControl:238965): GLib-GObject-WARNING **: 15:02:27.068: invalid uninstantiatable type 'void' in cast to 'GstElement'

(QGroundControl:238965): GStreamer-CRITICAL **: 15:02:27.068: gst_element_add_pad: assertion 'GST_IS_ELEMENT (element)' failed

(QGroundControl:238965): GLib-GObject-CRITICAL **: 15:02:27.068: g_object_set: assertion 'G_IS_OBJECT (object)' failed

Apparently, during refactoring, variables were mixed up in places. 
This PR fixes the problem.
@bkueng Please review this changes since the previous changes were merged into the master branch by you.

### DIFF
--- a/src/VideoReceiver/gstqgcvideosinkbin.c
+++ b/src/VideoReceiver/gstqgcvideosinkbin.c
@@ -114,10 +114,10 @@ _vsb_sink_pad_query(GstPad* pad, GstObject* parent, GstQuery* query)
 static void
 _vsb_init(GTypeInstance *instanceData, void *vsbVoid)
 {
-    Q_UNUSED(instanceData);
+    Q_UNUSED(vsbVoid);
 
     GstQgcVideoSinkBin *vsb;
-    vsb = (GstQgcVideoSinkBin *)vsbVoid;
+    vsb = (GstQgcVideoSinkBin *)instanceData;
 
     gboolean initialized        = FALSE;
     GstElement* glcolorconvert  = NULL;


fredowski:
This would solve https://github.com/mavlink/qgroundcontrol/issues/10378. I had the same problem for the PR https://github.com/mavlink/qgroundcontrol/pull/10380. I simply reverted the original commit. I was not too sure if the Warning fix is o.k. at all because the number of parameters changed. 

I think the plugin uses the [GstPluginInitFunc](https://gstreamer.freedesktop.org/documentation/gstreamer/gstplugin.html?gi-language=c#GstPluginInitFunc) which has just one parameter and not two.

s-lisovenko:
I can't say for sure about this, because I'm not familiar enough with the gstreamer,  but in this case a function from glib-2.0 is used and it has 2 parameters (GTypeInstance   *instance, gpointer  g_class). Perhaps somewhere in the depths, 
[GstPluginInitFunc](https://gstreamer.freedesktop.org/documentation/gstreamer/gstplugin.html?gi-language=c#GstPluginInitFunc)  is called, which expects one parameter (Gst Plugin * plugin).

bkueng (approved):
Sorry I missed this, yes this looks correct.

According to https://www.freedesktop.org/software/gstreamer-sdk/data/docs/2012.5/gobject/gobject-Type-Information.html#GInstanceInitFunc it has 2 arguments.

s-lisovenko:
> Sorry I missed this, yes this looks correct.
> 
> According to https://www.freedesktop.org/software/gstreamer-sdk/data/docs/2012.5/gobject/gobject-Type-Information.html#GInstanceInitFunc it has 2 arguments.

If everything is ok what do you think about renaming vsbVoid to g_class? Because vsbVoid may associate with GstQgcVideoSinkBin (vsb = (GstQgcVideoSinkBin *))?

fredowski:
Is it possible that this are two different things? You are referring to the gobject init funtion but here we have the plugin init function which is a different thing?

When you read here the [description of the GST_PLUGIN](https://gstreamer.freedesktop.org/documentation/gstreamer/gstplugin.html?gi-language=c#gstplugin-page), then they refer to the [GstPluginInitFunc](https://gstreamer.freedesktop.org/documentation/gstreamer/gstplugin.html?gi-language=c#GstPluginInitFunc) which has one parameter.

I think we have here the GstPluginInitfunc and not the gobject funtion that you mention? You can also compare to the aasink plugin which is part of the gst-plugins-good collection:

https://github.com/GStreamer/gst-plugins-good/blob/master/ext/aalib/gstaasink.c#L276

fredowski:
I think I am wrong. The plugin definition is here:

https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoReceiver/gstqgc.c#L37

but the _vsb_init function is used here:

https://github.com/mavlink/qgroundcontrol/blob/master/src/VideoReceiver/gstqgcvideosinkbin.c#L311

where the plugin type function is. And they indeed refer to the "GInstanceInitFunc". Sorry for the confusion.

fredowski (approved):
I think this is o.k. - I will try that in my PR https://github.com/mavlink/qgroundcontrol/pull/10380 to fix the CI pipeline for Android and MacOS.

patrickelectric:
ping @acxz

acxz (approved):
oof sorry about my commit, guys to be honest I was also getting lost in the all the variables and the number of variables that need to be exposed, etc.

Looks like the first argument is the proper argument that needs to be casted and used. Thanks for catching my mistake @s-lisovenko !

Issue #9440, opened 2021-02-15
RTSP video from specific video cam not displayed since 4.1.0


tzuriel-spearuav:
I'm working with Nextvision video camera, and since v4.1.0 video is not displayed.
The log has the following error :
```
[D] at C:\projects\qgroundcontrol\src\QGCApplication.cc:287 - "Settings location "C:/Users/Tzuriel/AppData/Roaming/QGroundControl.org/QGroundControl.ini" Is writable?: true"
[D] at C:\projects\qgroundcontrol\src\QGCLoggingCategory.cc:118 - "Filter rules "*Log.debug=false\nGStreamerAPILog.debug=true\n""
[E] at ../gst/gstelementfactory.c:467 - "(NULL) no such element factory "qmlglsink"!"
[D] at C:\projects\qgroundcontrol\src\QGCApplication.cc:390 - "System reported locale: QLocale(English, Default, Israel) ; Name "en_IL" ; Preffered (used in maps):  "en-US""
[E] at C:\projects\qgroundcontrol\src\QGCApplication.cc:482 - "Error loading source localization for "en_IL""
[E] at C:\projects\qgroundcontrol\src\QGCApplication.cc:487 - "Error loading json localization for "en_IL""
[E] at :0 - "serialnmea: No known GPS device found. Specify the COM port via QT_NMEA_SERIAL_PORT."
[D] at C:\projects\qgroundcontrol\src\Vehicle\MAVLinkLogManager.cc:360 - "MAVLink logs directory: "C:/Users/Tzuriel/OneDrive - SpearUAV/Documents/QGroundControl/Logs""
[D] at C:\projects\qgroundcontrol\src\QtLocationPlugin\QGCMapEngine.cpp:164 - "Map Cache in: "C:/Users/Tzuriel/AppData/Local/cache/QGCMapCache300" / "qgcMapCache.db""
[D] at qrc:/qml/MainRootWindow.qml:51 - "QGCCorePlugin(0x238f83d1180) []"
[E] at qrc:/qml/MainRootWindow.qml:412 - "qrc:/qml/MainRootWindow.qml:412:13: QML ColumnLayout: Possible anchor loop detected on vertical anchor."
[E] at ../gst/videoparsers/gsth264parse.c:1492 - "<h264parse0> broken/invalid nal Type: 1 Slice, Size: 8820 will be dropped"
[...]

PR #10396, opened 2022-09-06
MacOS: update GStreamer 1.18.1 to 1.18.6 and fix CI (#10393 / #10408) 

fredowski:
The CI builds on master for MacOS do not work. This PR updates the CI for MacOS and 

  * MacOS CI: Use official gstreamer package by picking patches from @patrickelectric from stable
  * MacOS CI: Update gstreamer from 1.18.1 to 1.18.6 for MacOS

With GStreamer 1.18.6 the application does not crash anymore from the videofeed

`gst-launch-1.0 videotestsrc ! x264enc ! rtph264pay ! udpsink host=127.0.0.1 port=5600`

Closes: #10393 
Closes: #10408
Closes: #10435

PR #10409, opened 2022-09-11
Video: increase timeout for RTSP video streaming (#9440)

fredowski:
In #9440 the video streaming failed for some cameras and the bunny example RTSP streaming movie. This worked in 4.0.6 and it did not work anymore in 4.1.0. The same unstable behaviour is observed in #10396.

The gstreamer rtsp source will switch to tcp when no data is received via udp after 5 seconds. The default timeout for the gstreamer video receiver is 2 seconds resulting in a permanent teardown and restart until eventually some data is received. As far as I understood the code 4.0.6 did not have a timeout which is in 4.1.0. I change the timeout for RTSP to 8 seconds and leave it at 2 seconds for the other streams. With 8 seconds timeout I can receive the bunny example without problems.

Closes: #9440

I tested this on MacOS with the changes from #10396 included and on Android.

### DIFF
--- a/src/Settings/Video.SettingsGroup.json
+++ b/src/Settings/Video.SettingsGroup.json
@@ -106,7 +106,7 @@
     "type":             "uint32",
     "min":              1,
     "units":            "s",
-    "default":     2
+    "default":     8
 },
 {
     "name":             "streamEnabled",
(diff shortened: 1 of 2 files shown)


fredowski:
This can be tested with the "Bunny" Video stream. 

* QGC: Application Settings -> General -> Video Source: "RTSP Video Stream"
* RTSP URL: rtsp://wowzaec2demo.streamlock.net/vod/mp4:BigBuckBunny_115k.mp4

mrpollo:
LGTM

fredowski:
Thanks for reviewing and merging!

Issue #9411, opened 2021-02-05
disconnect and connect buttons does not updated correctly


mnumanuyar:
## Expected Behavior
There is two issues, but I think they are related.
1. After succesfull connection when disconnect button clicked, if disconnect is sucsesfull, disconnect button should be disabled and connect button should be enabled
2. if connection failed, connect and disconnect button states should not changed
## Current Behavior
1. after succesfull connection when clicked disconnect, Disconnect button remains enabled. If users clicks to a different connection or goes to a different screen, and then comes back, button status is fixed
2. if connection refused, connect button becomes disabled, and disconnect button enabled. Clicking disconnect does nothing.

## Steps to Reproduce: 
1. open gcs,
2. go to appsettings -> CommLinks, chose a connection settings that would not refuse connection.
3. click disconnect 
4. Observe that how much you wait disconnect is not became disabled and connect does not become enabled.

1. open gcs
2. go to appsettings -> CommLinks, chose a connection settings that would refuse connection
3. observe that altough warning pop up appears, button states are changed (connect button becames disabled, disconnect button becomes enabled)
4.click disconnect
5. observe that nothing happens

## System Information
- Operating System: kubuntu 18.04 
- QGC Version: 4.1
- QGC build: daily. sha: 9589cc5f002d55099954428603413a6a00df20fa
- Flight Controller: ardupilot sitl
- Autopilot (with version): ardupilot sha: b113fe15ab30946b1b5105d1f8bd6df9d35696e6 

## Detailed Description
For first problem: A simple work around is simply choosing another connection or going to anoher screen and coming back.

For the second problem: As far as I am aware no workaround exist and only solution is quiting and re opening qgcs.

I am connecting to an sitl vehicle on TCP 127.0.0.1:5760 did not try other connection types.

## Log Files and Screenshots
first problem:
[DisconnectDoesNotUpdate.txt](https://github.com/mavlink/qgroundcontrol/files/5932835/DisconnectDoesNotUpdate.txt)
2nd problem:
[ConnectButtonChangesOnFail.txt](https://github.com/mavlink/qgroundcontrol/files/5932838/ConnectButtonChangesOnFail.txt)

kschoelpple:
Hi,
also noticed same and in fact also applies to Edit button which does appear to not respond when clicked after a connection as been initiated. As the Disconnect button also doesn't respond as expected a change to connection can not be made. Only solution is to re-start Qgroundcontrol.
System in use:
[...]

PR #10412, opened 2022-09-14
fix: comm link bugs

dakejahl:
- Fix UI update when a link is disconnected
- Clean up properly if a link fails to connect

We should probably get rid of all of the std::shared_ptr and use QSharedPointer instead. ~A PR for another day.~ I pushed a commit that replace them, let me know if I should revert it. Seems like using Qt smart pointers is probably better for a Qt project. I can go through and change it everywhere in the code base if we want to do that.

### DIFF
--- a/src/comm/LinkManager.cc
+++ b/src/comm/LinkManager.cc
@@ -160,6 +160,9 @@ bool LinkManager::createConnectedLink(SharedLinkConfigurationPtr& config, bool i
         _mavlinkProtocol->setVersion(_mavlinkProtocol->getCurrentVersion());
 
         if (!link->_connect()) {
+            link->_freeMavlinkChannel();
+            _rgLinks.removeAt(_rgLinks.indexOf(link));
+            config->setLink(nullptr);
             return false;
         }
 
--- a/src/ui/preferences/LinkSettings.qml
+++ b/src/ui/preferences/LinkSettings.qml
@@ -130,7 +130,10 @@ Rectangle {
         QGCButton {
             text:       qsTr("Disconnect")
             enabled:    _currentSelection && _currentSelection.link
-            onClicked:  _currentSelection.link.disconnect()
+            onClicked:  {
+                _currentSelection.link.disconnect()
+                _currentSelection.linkChanged()
+            }
         }
         QGCButton {
             text:       qsTr("MockLink Options")


booo:
Does this relate to https://github.com/mavlink/qgroundcontrol/issues/9411 ?

dakejahl:
yes this is the fix for #9411

src/comm/MAVLinkProtocol.cc
```diff
@@ -364,7 +364,7 @@ void MAVLinkProtocol::receiveBytes(LinkInterface* link, QByteArray b)
 
             // Anyone handling the message could close the connection, which deletes the link,
             // so we check if it's expired
-            if (1 == linkPtr.use_count()) {
+            if (!linkPtr.isNull()) {
```
bkueng, line 367: This does not look correct, it should use the ref count.

dakejahl: There is no function to get the reference count for a QSharedPointer
https://doc.qt.io/qt-5/qsharedpointer.html
https://stackoverflow.com/questions/41934716/reference-count-number-for-qsharedpointer

bkueng: Yes I saw, so we can't use it.

dakejahl: Are you suggesting we can't use QSharedPointer because it doesn't have a reference count function? This doesn't seem like a very good reason, especially since the places where it's used don't really add any value.

dakejahl:
reverted the last commit

src/comm/LinkManager.cc
```diff
@@ -160,6 +160,10 @@ bool LinkManager::createConnectedLink(SharedLinkConfigurationPtr& config, bool i
         _mavlinkProtocol->setVersion(_mavlinkProtocol->getCurrentVersion());
 
         if (!link->_connect()) {
+            link->_freeMavlinkChannel();
+            _rgLinks.removeAt(_rgLinks.indexOf(link));
+            config->linkChanged();
+            config->setLink(nullptr); // weak_ptr is cleaned up
```
bkueng, line 166: `setLink` will already emit `linkChanged`

src/comm/MAVLinkProtocol.cc
bkueng: I'm not sure what you mean by this. The logic here actually prevents subtle crashes, so it's very relevant and changing the semantics breaks it.

dakejahl: oh I overlooked the details. Hmm is there actually an instance in the code base where a slot handling  `messageReceived` actually deletes the link? Doesn't seem quite right. Anyway this PR should be good to go now.

bkueng: Yes I think when rebooting the vehicle if I remember correctly. There's a lot that can happen within receiving messages, and thus it's quite complex.

bkueng (commented):
Can you squash the last commit? Otherwise good to merge, thanks for the fix!

PR #10416, opened 2022-09-16
View: Show correct progressbar also for Mission, Fence and Rally

fredowski:
The progressbar is updated with the actual progress for Mission-, Fence- and Rallypointdownload.

Without the patch the progressbar is only updated for each particular step when it is at 100% and the initial connect statemachine advances. The progress information is already available but it was not used. For slow connections the mission download with many mission items can take quite a while and the progressbar does not indicate the progress without this patch.

### DIFF
--- a/src/Vehicle/InitialConnectStateMachine.cc
+++ b/src/Vehicle/InitialConnectStateMachine.cc
@@ -318,6 +318,8 @@ void InitialConnectStateMachine::_stateRequestMission(StateMachine* stateMachine
         } else {
             qCDebug(InitialConnectStateMachineLog) << "_stateRequestMission";
             vehicle->_missionManager->loadFromVehicle();
+            connect(vehicle->_missionManager, &MissionManager::progressPct, connectMachine,
+                    &InitialConnectStateMachine::gotProgressUpdate);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


mrpollo:
@bkueng can you please help me give this a test? I already went through the codebase with @fredowski on a call and it looks good to me

fredowski:
This video: https://vimeo.com/756880072 shows the non progressing progress bar in release version 4.2.3 and the fixed version from this pull request. With the fix the progress bar also progresses during mission, fence and rally during initial connect.

src/Vehicle/InitialConnectStateMachine.cc
```diff
@@ -353,6 +360,10 @@ void InitialConnectStateMachine::_stateRequestRallyPoints(StateMachine* stateMac
     Vehicle*                    vehicle         = connectMachine->_vehicle;
     SharedLinkInterfacePtr      sharedLink      = vehicle->vehicleLinkManager()->primaryLink().lock();
 
+    if (vehicle->_geoFenceManager->supported() && vehicle->_geoFenceManager != nullptr)
```
bkueng, line 363: You could just unconditionally disconnect (_geoFenceManager is always initialized).
```suggestion
    if (vehicle->_geoFenceManager != nullptr && vehicle->_geoFenceManager->supported())
```

fredowski: Right. I removed the paranoia checks and just disconnect unconditionally. As you pointed out my paranoia check was not even correct... Thanks for reviewing the code!

fredowski:
Thanks for reviewing and merging the code!

PR #10419, opened 2022-09-18
NEW: MavLink FTP binary parameter download (APM)

fredowski:
Ardupilot vehicles support the download of parameters via MavLink FTP which is much faster and possibly more robust for low bandwidth radio links with a higher packet loss rate like RFD868 compared to the existing conventional mavlink parameter download.

The proposed implementation is based on the existing ftp download system which exists in qgc. I added the parameter file parser which then adds the parameters to the fact system. If the ftp parameter download fails, the conventional download is used. For existing vehicles which do not support ftp at all, this results in two additional initial request timeouts. If the vehicle supports ftp but not the parameterfile like px4 vehicles, the overhead are two messages. 

The Ardupilot parameterfile read response as today responds with a fixed 1 MByte file size as the parameterfile is build on the fly and it is not possible to predict the filesize before the transfer starts, because the filesize depends on the used block size. This mismatch between actual transmitted filesize and indicated filesize will result in failure in the current qgc ftp implementation. Therefore I added an additional option to ignore the filesize which is indicated in the mavlink response. As the progress bar is based on the indicated filesize I added PR https://github.com/ArduPilot/ardupilot/pull/21744 to Ardupilot for a better filesize estimation. 

I tested the implementation with the current Arduplane SITL simulation on a Mac with a simulated 64kBit/s link with 30ms delay and a packet loss rate of 5%. I used the packet filtering and dummynet system on MacOS. 

```
sudo pfctl -E
sudo pfctl -f ./pf.conf
=== pf.conf ===
dummynet in proto udp from any to any port 14550 pipe 1
dummynet out proto udp from any port 14550 to any pipe 2
===
sudo dnctl pipe 1 config bw 64Kbit/s delay 30 plr 0.05 
sudo dnctl pipe 2 config bw 64Kbit/s delay 30 plr 0.05
```

The conventional parameter download takes 130 seconds while the ftp parameter download takes 5 seconds. You can see the comparison in this video https://vimeo.com/751000899.

@tridge, @DonLakeFlyer: I added the option to disable the filesize check in the QGC ftp client. At first I wanted to fix the filesize response in Ardupilot but after thinking about it I agree that it is pretty complicated given the special padding design to avoid parameter value corruption during the on-the-fly generation for volatile parameters. The implementation now is not restricted to Ardupilot. If PX4 have any plans to implement this it would work right away.


fredowski:
Here is a video showing the QGroundControl ftp parameter download with a RFD868+ modem and a PixRacer running Arduplane 4.2.1 firmware: https://vimeo.com/752651470

fredowski:
In this video: https://vimeo.com/756925930 QGC with this ftp patch is shown with a simulated 256kBits/s link with Arduplane SITL. You can compare the parameter download speed with the video from here: https://vimeo.com/756880072 showing release 4.2.3 with a simulated 256kBits/s in the context of the progress bar PR.

bkueng (commented):
Nice work, generally looks good.

fredowski:
@bkueng : Thanks for looking at the code! I followed your proposals:

  * changed the variable names according to your idea
  * request @PARM/param.pck instead of @PARM/param.pck/withdefaults=1 as I do not use the default information anyway in QGC and the default parameter style has been introduced later.
  * Check the component id. I check for the component id now such that a parameter request for other component ids than MAV_COMP_ID_AUTOPILOT1 are handled via the conventional download mechanism. The mavftp will request parameters for MAV_COMP_ID_AUTOPILOT1 but the vehicle manager starts with a broadcast request for all components. So if there is a vehicle which has another component, then that component will not see a parameter request broadcast. If I do a parameter request broadcast, then the mavftp is obsolete because the vehicle will send all parameters again.

I guess this behaviour will result in not discovering for example the ESP8266 component here:

https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/APM/APMAutoPilotPlugin.cc#L150

So I think this requires to add some method to discover those components which are now discovered via the detection of parameters which have been obtained by a parameter request broadcast.

bkueng (approved):
> So I think this requires to add some method to discover those components which are now discovered via the detection of parameters which have been obtained by a parameter request broadcast.

I don't see an easy way around this either. I would have expected component discovery happening based on heartbeats. This would be a bigger change though.

fredowski:
@bkueng: Thank you for reviewing the code and merging. Yes I noticed that the (new?) method for service discovery is based on emitting heartbeats, e.g. for gimbals.

PR #10444, opened 2022-09-30
Do not send PROTOCOL_VERSION request when using Ardupilot firmware

WTPENGUIN:
The Ardupilot firmware doesn't seem to handle PROTOCOL_VERSION requests. 

It seems to be using the capabilities field in the AUTOPILOT_VERSION message.

So, When connecting the Ardupilot firmware vehicle after #10436, QGC failed to upload rally points and geofences.

Please review this PR.
(I'm not good with English, so I used a translator.)


WTPENGUIN:
Oops, I accidentally opened a PR as master rather than a branch. We will reopen after closing.

PR #10445, opened 2022-09-30
Do not send PROTOCOL_VERSION request when using Ardupilot firmware

WTPENGUIN:
The Ardupilot firmware doesn't seem to handle PROTOCOL_VERSION requests.

It seems to be using the capabilities field in the AUTOPILOT_VERSION message.

So, When connecting the Ardupilot firmware vehicle after #10436, QGC failed to upload rally points and geofences.

Please review this PR.
(I'm not good with English, so I used a translator.)

### DIFF
--- a/src/Vehicle/InitialConnectStateMachine.cc
+++ b/src/Vehicle/InitialConnectStateMachine.cc
@@ -216,6 +216,9 @@ void InitialConnectStateMachine::_stateRequestProtocolVersion(StateMachine* stat
         if (sharedLink->linkConfiguration()->isHighLatency() || sharedLink->isPX4Flow() || sharedLink->isLogReplay()) {
             qCDebug(InitialConnectStateMachineLog) << "Skipping REQUEST_MESSAGE:PROTOCOL_VERSION request due to link type";
             connectMachine->advance();
+        } else if (vehicle->apmFirmware()) {
+            qCDebug(InitialConnectStateMachineLog) << "Skipping REQUEST_MESSAGE:PROTOCOL_VERSION request due to Ardupilot firmware";
+            connectMachine->advance();
         } else {
             qCDebug(InitialConnectStateMachineLog) << "Sending REQUEST_MESSAGE:PROTOCOL_VERSION";
             vehicle->requestMessage(_protocolVersionRequestMessageHandler,


patrickelectric (approved):
Since PROTOCOL_VERSION is in WIP, that makes sense.
ref: https://mavlink.io/en/messages/common.html#PROTOCOL_VERSION

fredowski:
I can confirm PR #10436 breaks the geofence and rally download for Arduplane in ardupilot master. The log shows that mission download works but geofence and rally are skipped because qgc thinks they are not supported. I tested this with arduplane in SITL.

```
PlanManagerLog: "_requestNextMissionItem T:Mission sequenceNumber:retry" 61 0
PlanManagerLog: "_handleMissionItem T:Mission seq:command:current:ardupilotHomePositionUpdate" 61 16 false false
PlanManagerLog: "_requestNextMissionItem T:Mission sequenceNumber:retry" 62 0
PlanManagerLog: "_handleMissionItem T:Mission seq:command:current:ardupilotHomePositionUpdate" 62 16 false false
PlanManagerLog: "_requestNextMissionItem T:Mission sequenceNumber:retry" 63 0
PlanManagerLog: "_handleMissionItem T:Mission seq:command:current:ardupilotHomePositionUpdate" 63 206 false false
PlanManagerLog: "_requestNextMissionItem T:Mission sequenceNumber:retry" 64 0
PlanManagerLog: "_handleMissionItem T:Mission seq:command:current:ardupilotHomePositionUpdate" 64 16 false false
PlanManagerLog: _readTransactionComplete read sequence complete
PlanManagerLog: _setTransactionInProgress "T:Mission" 0
InitialConnectStateMachineLog: _stateRequestGeoFence: skipped due to no support
InitialConnectStateMachineLog: _stateRequestRallyPoints: skipping due to no support
InitialConnectStateMachineLog: Signalling initialConnectComplete
setCurrentPlanViewSeqNum
setCurrentPlanViewSeqNum
```

fredowski:
I tested the MacOS Version from this PR versus Arduplane in SITL and it fixes the problem. Geofence and Rally points are downloaded again. @WTPENGUIN : Nice work!

fredowski (changes requested):
I think that it should be tested as "else if". If the "sharedLink" in line 216 advances the statemachine, then it should not be advanced again in line 223.

src/Vehicle/InitialConnectStateMachine.cc
```diff
@@ -216,6 +216,11 @@ void InitialConnectStateMachine::_stateRequestProtocolVersion(StateMachine* stat
         if (sharedLink->linkConfiguration()->isHighLatency() || sharedLink->isPX4Flow() || sharedLink->isLogReplay()) {
             qCDebug(InitialConnectStateMachineLog) << "Skipping REQUEST_MESSAGE:PROTOCOL_VERSION request due to link type";
             connectMachine->advance();
+        }
```
fredowski, line 219:
```suggestion
        } else if (vehicle->apmFirmware()) {
              qCDebug(InitialConnectStateMachineLog) << "Skipping REQUEST_MESSAGE:PROTOCOL_VERSION request due to Ardupilot firmware";
```

fredowski:
The split in three commits would not make sense for a merge. Can you squash the three commits to one and add a little explanation in the commit comment?

WTPENGUIN:
@fredowski 
I have grouped 3 commits via force-push.
(I haven't used force-push much, so I made a mistake. sorry.)

fredowski:
You mentioned that you are not too good in english... Maybe:

```
InitialconnectStatemachine: Do not request PROTOCOL_VERSION for APM vehicles

Ardupilot does not support the MAVLINK_MSG_ID_PROTOCOL_VERSION. 
When this is attempted QGC assumes that Ardupilot does not support
the geofence and rally download. So this is skipped to make fence
and rally download work.
```

fredowski:
Yes, if you modify the commits, then you need to force-push. That is not a problem because this only affects your repository and your branch. You can squash the commits with `git rebase -i master` when you are in your fixInitialConnect branch. See: https://git-scm.com/docs/git-rebase . Then you need need to force push again.

WTPENGUIN:
@fredowski 

I also changed the commit message. Thanks for the advice.

src/Vehicle/InitialConnectStateMachine.cc
```diff
@@ -216,6 +216,10 @@ void InitialConnectStateMachine::_stateRequestProtocolVersion(StateMachine* stat
...
```
fredowski, line 219:
```suggestion
        } else if (vehicle->apmFirmware()) {
             qCDebug(InitialConnectStateMachineLog) << "Skipping REQUEST_MESSAGE:PROTOCOL_VERSION request due to Ardupilot firmware";
             connectMachine->advance();
         } else {
```

fredowski: Can you just append the "else if" to line 219?

fredowski (changes requested):

src/Vehicle/InitialConnectStateMachine.cc
WTPENGUIN: Done.

fredowski (approved):
Very nice! Thanks for your support!

PR #10464, opened 2022-10-25
Fix mac custom builds

DonLakeFlyer:


DonLakeFlyer:
I was seeing the new compiler error in my custom builds as well. That is was just on my side. But turns out it was upstream as well. Looks like new compiler in github actions mac environment popping new errors. They come from eigen which is out of our control so the warning has to be turned off.

Davidsastresas:
Thank you! I think this is important and all checks are passed, shall we merge?

DonLakeFlyer:
I don't know what the current process is to merge things in. I still have merge rights. If folks are ok with me merging my own pulls in after they get positive review I can do that.

Davidsastresas:
To be honest I was granted very recently merge rights, so maybe I am not the most indicated person to have an opinion about this, but considering QGC nowadays exists basically because of you I don't think there are much more qualified persons than you for merging/reviewing your own commits, unless you specifically ask for revisions from somebody else.

From my part I trust 100% on your criteria for doing whatever with QGC, but maybe leaving a day or two for the rest of maintainers to take a look and comment in case they feel like, and if no updates go ahead with the merging, I would think it is a safe approach that in my opinion will hardly be wrong for anybody.

Besides the above, I don't think we currently have like an established workflow for merging commits. Maybe we should? I don't know, but from my part it is more than fine that you merge whatever you consider. Thank you Don!

patrickelectric (approved):

patrickelectric:
@DonLakeFlyer During the QGC meeting we are doing reviews and merging PRs, but things like this one can be approved and merged if someone gives the green check :+1:

DonLakeFlyer:
Great thanks folks. I am super busy working on custom build wildlife conservation stuff. But I have a number of custom build things I want to clean up and document better.

PR #10467, opened 2022-10-28
Moved Guided Confirm slider to top of window

DonLakeFlyer:
Here is what the current Guided Confirmation slider looks like using Outdoor colors:
Screen Shot 2022-10-28 at 10 07 14 AM

There are a number of problems with this:
* Since it is part of the Telemetry Bar when you move the mouse over it to slide the slider the telemetry bar edit tools show up
* When it shows up it causes the telemetry bar to move position
* On a small screen like a 10" tablet, stacking telemetry and slider begin to invade the center of the map which is tried to keep clear as much as possible
* The coloring of the slider in the outdoor colors makes it very hard to see in the sun outdoors
* The text explaining what the slider is about is pretty small once again making it hard to read outdoors which is kind of important for these sliders since they are critical functionality

Here is the new version:
Screen Shot 2022-10-28 at 10 04 35 AM

This is better because:
* It keeps the center of the map clearer
* The telemetry bar no longer moves around on you
* The coloring and text size make it is easier to read outdoors

### DIFF
--- a/src/FlightDisplay/FlyViewWidgetLayer.qml
+++ b/src/FlightDisplay/FlyViewWidgetLayer.qml
@@ -106,6 +106,16 @@ Item {
         visible:            !multiVehiclePanelSelector.showSingleVehiclePanel
     }
 
+
+    GuidedActionConfirm {
+        anchors.margins:            _toolsMargin
+        anchors.top:                parent.top
+        anchors.horizontalCenter:   parent.horizontalCenter
+        z:                          QGroundControl.zOrderTopMost
+        guidedController:           _guidedController
+        guidedValueSlider:          _guidedValueSlider
+    }
+
     FlyViewInstrumentPanel {
         id:                         instrumentPanel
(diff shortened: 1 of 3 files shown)


Davidsastresas:
I totally agree with your arguments and I think we indeed improve the user experience with this changes. Thanks!

DonLakeFlyer:
Crap, another problem I thought I was only seeing in my custom builds. The eigen lib has deprecated usage. I'll try updating the Eigen lib to latest to see if it goes away. Otherwise the warnings needs to be silenced for now.

DonLakeFlyer:
@patrickelectric You added eigen right? Do you have any idea why my builds is failing and every once else's pulls are not?

patrickelectric (approved):
@DonLakeFlyer I just restart the CI, from what I could see in your last commit it was a problem on the Qt server, it should be fine now.

DonLakeFlyer:
@patrickelectric Ah, strange. Thanks.

Issue #10465, opened 2022-10-26
QGC does not support all sensor orientations


Jaeyoung-Lim:
## Expected Behavior
Board orientation can be configured for all values in https://mavlink.io/en/messages/common.html#MAV_SENSOR_ORIENTATION are available.

## Current Behavior
The entries in https://mavlink.io/en/messages/common.html#MAV_SENSOR_ORIENTATION  have 35 entries, but the board orientations supported by QGC only has 30 entires.
![image](https://user-images.githubusercontent.com/5248102/198005799-0b3df3b5-1e93-4509-827d-5d536d4ae7f5.png)

This leaves out the orientation `[MAV_SENSOR_ROTATION_ROLL_90_YAW_270](https://mavlink.io/en/messages/common.html#MAV_SENSOR_ROTATION_ROLL_90_YAW_270)`, which I needed for setting up a tailsitter

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1. Connect a autopilot to QGC and start a calibration routine

## System Information
- Operating System: [Ubuntu 18.04]
- QGC Version: [daily]
- QGC build: [daily]
- Flight Controller: [Pixhawk4]
- Autopilot (with version): PX4

## Log Files and Screenshots
- [QGC Console Logs](https://docs.qgroundcontrol.com/en/Support/Support.html#reporting-bugs)
- Autopilot logs when available (post a link).
- Screenshots of QGC to help identify the current issue/bug behavior.

Jaeyoung-Lim:
I noticed that for PX4, the enums are even misaligned, and therefore incorrect

booo:
Thanks for reporting the bug. The enumeration of rotations can be found here:

https://github.com/mavlink/qgroundcontrol/blob/master/src/AutoPilotPlugins/PX4/SensorsSetup.qml#L59

Maybe you can update the enumeration and provide us with a patch?

Jaeyoung-Lim:
@booo I think the problem is more that the enumeration is done separately in QGC, rather than reading it out from mavlink definitions. However, I am not sure how to implement this properly with qmls

junwoo091400:
I can check on this later, as I am nowadays learning a lot about qt 🤔

It would be a matter of
1. C++: Fetching the orientation enums from the mavlink
2. C++: handling the name strings (not sure if we can get them directly)
3. Qml: having a checkbook be able to read the Qlist property or sth similar from the C++ side, and display it

[...]

PR #10471, opened 2022-11-01
PX4: Add missing sensor orientations to PX4 sensor setup UI

booo:
The enumeration of sensor orientation was not up to date. Reported in:

https://github.com/mavlink/qgroundcontrol/issues/10465

I updated the enum based on the mavlink enum from here:

https://mavlink.io/en/messages/common.html#MAV_SENSOR_ORIENTATION


src/AutoPilotPlugins/PX4/SensorsSetup.qml
```diff
@@ -83,10 +83,22 @@ Item {
...
         "ROTATION_ROLL_180_PITCH_270",
-        "ROTATION_PITCH_90_YAW_180",
-        "ROTATION_ROLL_90_PITCH_90"
+        "ROTATION_ROLL_270_PITCH_270",
+        "ROTATION_ROLL_90_PITCH_180_YAW_90",
+        "ROTATION_ROLL_90_YAW_270",
+        "ROTATION_ROLL_90_PITCH_68_YAW_293",
+        "ROTATION_PITCH_315",
+        "ROTATION_ROLL_90_PITCH_315",
+        "ROTATION_CUSTOM
```
Jaeyoung-Lim, line 101:
```suggestion
        "ROTATION_CUSTOM"
```

Jaeyoung-Lim: Probably missing a quote?

booo: Yes. Pushed a fix. Thanks for spotting that!

patrickelectric (approved):

junwoo091400:
Can we merge this? @patrickelectric

booo:
> Can we merge this? @patrickelectric 

Did somebody test if that actually fixes the user problem? I was waiting for feedback from the user.

junwoo091400:
@Jaeyoung-Lim could you try it out on your vehicle? You can get the Linux binary from the Github Actions here: https://github.com/mavlink/qgroundcontrol/actions/runs/3369382358

junwoo091400:
> > Can we merge this? @patrickelectric
> 
> Did somebody test if that actually fixes the user problem? I was waiting for feedback from the user.

I think it could be nice to add a test like this: https://github.com/PX4/PX4-Autopilot/blob/1c5750b292496024669c746c23359e445dceea75/src/modules/mavlink/mavlink_messages.cpp#L147-L231

So that we can detect enum mis-match, would you be able to implement that as well?

patrickelectric:
Codewise it makes sense, but I would prefer someone with PX4 expertise to merge the code.

src/AutoPilotPlugins/PX4/SensorsSetup.qml
```diff
@@ -83,10 +83,22 @@ Item {
...
         "ROTATION_ROLL_180_PITCH_270",
-        "ROTATION_PITCH_90_YAW_180",
-        "ROTATION_ROLL_90_PITCH_90"
+        "ROTATION_ROLL_270_PITCH_270",
+        "ROTATION_ROLL_90_PITCH_180_YAW_90",
+        "ROTATION_ROLL_90_YAW_270",
+        "ROTATION_ROLL_90_PITCH_68_YAW_293",
+        "ROTATION_PITCH_315",
+        "ROTATION_ROLL_90_PITCH_315",
+        "ROTATION_CUSTOM"
```
bkueng, line 101: I expect this will need special handling and show the extra fields to configure the angles. @junwoo091400

booo: I can remove the ROTATION_CUSTOM option and we leave this as an exercise for later.

bkueng: Yes I would remove it for now.

booo: I pushed a new version with the ROTATION_CUSTOM option removed and rebased onto master.

mrpollo:
thanks @booo

PR #10472, opened 2022-11-01
Frames Component Metadata Support & New Frame selection UI

junwoo091400:
## Description
This is the QGC implementation of https://github.com/mavlink/mavlink/pull/1908

It will eventually look something like this in QGC:
Demo

### What will be implemented
1. Support for Frames component metadata
2. New metadata based Frames visualization page (Fallback to old Frames page, if metadata isn't supported by the target)

## TODOs
- [x] Initial crude porting of the files from the [standalone Qt application](https://github.com/junwoo091400/PX4-Airframe-Metadata-support/commit/75d54c40dd234959ab284604fc6c3942c909ef09)
- [x] Refactor to expose the new Frames page in QML in the Airframes tab, and rename the tab to `Frames` :stuck_out_tongue: 
- [x] Implement (temporary) way to visualize the `frames.example.json` from the MAVLink PR, to test functionality
- [ ] Iterate on the UI design, to make it more intuitive (maybe Animation)
- [ ] Implement Component Metadata fetching part from the Autopilot
- [ ] Implement Component Metadata generation in PX4 (To be updated)
- [ ] Test the whole system in action with a PX4 target containing the new Frames metadata!


src/AutoPilotPlugins/PX4/Frames.schema.json
```diff
@@ -0,0 +1,120 @@
...
+                },
+                "url": {
+                    "type": "string",
+                    "description": "Link to the relevant material online about the frame"
+                },
+                "frame_parameter_values": {
+                    "type": "array",
+                    "description": "Values to set for parameters defined in `frame_parameters` settings to make frame selection come into effect",
+                    "minItems": 1
+                },
```
junwoo091400, line 65: @bkueng One thing that I wanted to have was setting the `maxItems` of this part to the length of the `frame_parameters` array below.

However, it seems like such dynamic (using length of an array element as `maxItems` setting) is not supported (not mentioned): https://json-schema.org/understanding-json-schema/reference/array.html#items

bkueng: Yes you can't do everything with a schema

junwoo091400:
Fixed a typo where the `FrameComponent.h` was being included (yet again!) in the Plugins API section at the top of `qgroundcontrol.pro` file. Probably doesn't affect build result though.

junwoo091400:
image

QGC Display is finally working with the https://github.com/mavlink/qgroundcontrol/pull/10472/commits/b14231626dc83edfc0a025cc8100e9833bf1c708

The biggest challenge (which was laughable) was this: https://github.com/mavlink/qgroundcontrol/pull/10472/commits/4eb3b02ab40dbb572bb26aad654d7d85ac59110b

Who would have thought that having a `Frame.qml` would conflict with Qt's own `Frame` data type (didn't know it existed)!!

junwoo091400:
> Not sure why [1fdb68b](https://github.com/mavlink/qgroundcontrol/commit/1fdb68b6ae5be7d1802ebef460300f84529ddbb0) broke the Go to parent frame functionality (at least displaying part), but here's the status!

Fixed with https://github.com/mavlink/qgroundcontrol/pull/10472/commits/5579a9a66887e337700f8459525fb5b101fbf5e8, error was you can't use `anchor` layout keyword inside `Flow` element, so the background rectangle was doing a weird thing :laughing: 

Also, the double heap free bug was fixed with https://github.com/mavlink/qgroundcontrol/pull/10472/commits/1137e034934b4438f741cb0b9421fa9d68730b4b

junwoo091400:
Finally got the pop-up to actually work :vomiting_face:, needed to import this QtQuick.Dialogs library, otherwise the buttons wouldn't show up!

Quite annoying how the QTCreator doesn't give a clue on what's actually wrong. Likely due to the custom components like `QGCPopupDialog` is not being registered in available QMLs though.

junwoo091400:
With the setting parameters part implemented, here's the current UI status:
image

Setting final frame:
image

Will improve the design / pop-up aesthetics :thinking:

DonLakeFlyer:
Is this dead? Work stopped on it a while back.

junwoo091400:
> Is this dead? Work stopped on it a while back.

After the Mavlink PR got rejected, I haven't continued on the feature development more 😞

https://github.com/mavlink/mavlink/pull/1908#issuecomment-1302658456

DonLakeFlyer:
Ok, closing then

PR #10476, opened 2022-11-03
PlanToolBarIndicators: Use more descriptive label for distance to WP

booo:
The current implementation displays two different distances by the same name. One of the distances is actually the distance to the privous waypoint. For the user it is "hard" to tell which one is which. I try to resolve this with a more descriptive label for the distance to the privous waypoint.

An alternative is maybe to rename the other distance to total distance or total mission distance.

### DIFF
--- a/src/PlanView/PlanToolBarIndicators.qml
+++ b/src/PlanView/PlanToolBarIndicators.qml
@@ -158,7 +158,7 @@ Item {
 
             Item { width: 1; height: 1 }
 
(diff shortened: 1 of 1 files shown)


zdanek (approved):
Seems good to me 🙂

Davidsastresas:
Seems ok for me to, but I think there is a typo, It should be "Dist prev WP:" from previous, instead of "Dist priv WP:", right?

PR #10477, opened 2022-11-03
QGCMapPolygonVisuals: Visualize edge lengths of polygon

booo:
This adds information about the edge lengths of the polygon to the map. With the current tools it is quite hard to plan e.g. a square of 1 km x 1 km on the open sea. One can orientate on the map scale but that yields imprecise results.

In trace mode it is not perfect until you have more than two points.

### DIFF
--- a/src/MissionManager/QGCMapPolygonVisuals.qml
+++ b/src/MissionManager/QGCMapPolygonVisuals.qml
@@ -62,7 +62,7 @@ Item {
 
     function addEditingVisuals() {
         if (_objMgrEditingVisuals.empty) {
-            _objMgrEditingVisuals.createObjects([ dragHandlesComponent, splitHandlesComponent, centerDragHandleComponent ], mapControl, false /* addToMap */)
+            _objMgrEditingVisuals.createObjects([ dragHandlesComponent, splitHandlesComponent, centerDragHandleComponent, edgeLengthHandlesComponent ], mapControl, false /* addToMap */)
         }
     }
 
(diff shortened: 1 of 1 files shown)


julled:
A screenshot would help to estimate the impact of the changes.

Currently its not possible to disable the visualization. I think these kind of helping UI elements should not be enabled by default, rather by a user setting.

src/MissionManager/QGCMapPolygonVisuals.qml
```diff
@@ -306,6 +306,25 @@ Item {
...
+            id:             mapQuickItem
+            anchorPoint.x:  sourceItem.width / 2
+            anchorPoint.y:  sourceItem.height / 2
+            visible:        !_circleMode
+
+            property int vertexIndex
+            property real distance
+
+            sourceItem: Text {
+              text: distance.toFixed(2) + " m"
```
julled, line 322: In my opinion cm resolution / precision of 2 is a bit overkill. 
Precision of 1 should be prefered.
Its better to not overfreight the UI with numbers.

booo:
This is what it currently looks like:

screenshot_20221111_140653

DonLakeFlyer:
I get the problem you are trying to solve. But this adds clutter for the 99% of people who don't need this. If you really need detailed survey area creation when not create it in google earth and then load either KML or Shape file with the perfectly created survey region?

DonLakeFlyer:
I"m going to try to modify this a bit so it's now shown all the time. I think it would be worthwhile if it only showed up while you are dragging a vertex.

booo:
> I"m going to try to modify this a bit so it's now shown all the time. I think it would be worthwhile if it only showed up while you are dragging a vertex.

Yes, that sounds like a good plan.

DonLakeFlyer:
Thanks for the pull @booo. This is a good new feature. See my replacement pull for the small updates.

Issue #4503, opened 2017-02-01
Allow copy/paste of console output


jaxxzer:
It would be easier to share debugging output if we could copy and paste the output in the console window rather than taking a screenshot or saving to file.

PR #10478, opened 2022-11-06
MessageIndicator: Enable copy/paste of messages

booo:
Requested by user here:

https://github.com/mavlink/qgroundcontrol/issues/4503

I'm not sure about the `SelectionColor` and `SelectedTextColor` values.

### DIFF
--- a/src/ui/toolbar/MessageIndicator.qml
+++ b/src/ui/toolbar/MessageIndicator.qml
@@ -147,10 +147,13 @@ Item {
                 pixelAligned:       true
 
                 TextEdit {
-                    id:             messageText
-                    readOnly:       true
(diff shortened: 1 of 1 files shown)


julled:
Did you test this if onefinger and twofinger scroll guestures still work on android?

booo:
No. If someone could test this that would be great!

booo:
This is what the current change looks like:

screenshot_20221111_143820

booo:
@DonLakeFlyer can you take a look at this. I think it is super handy for sharing actual error messages. Maybe it should be integrated in your recent changes to the GUI element? Thanks a lot.

DonLakeFlyer:
Looks good to me

Issue #10383, opened 2022-08-29
Allow Parameter Search using REGEX expression (e.g. EKF2_*_CTRL)


junwoo091400:
## Need
A lot of PX4's parameters have a repetitive form that has either a prefix or a suffix. Currently, the Parameter search works like this:
**Given the search phrase "X", search all the parameters with the phrase "X" in it's name OR description.**

However it has two problem / limitations:
1. User may not want to include the search to parameters descriptions
2. To search for a postfix (e.g. `_CTRL`), user need to search using `_CTRL`, which is counter intuitive compared to `*_CTRL`, a regex form.

### REGEX
Currently, to search for the EKF2_*_CTRL parameters, user has to search for '_CTRL', as the * marker as in REGEX expression is not supported.
![image](https://user-images.githubusercontent.com/23277211/187210514-6273ec99-396d-4a00-8b49-e85d1f604287.png)

## Proposal
1. QGC should have an option to include/exclude parameter's 'description' in the Search page
2. The REGEX (or at least * marker) search should be supported

booo:
Can you check if my [proposal](https://github.com/mavlink/qgroundcontrol/pull/10479) comes close to what you are looking for?

PR #10479, opened 2022-11-06
ParameterEditorController: Add regular expressions to parameter search

booo:
This adds regular expression support to the parameter search function. I added the isValid condition to prevent the contains function to spam the log with messages about invalid regular expressions.

User requested this feature here:

https://github.com/mavlink/qgroundcontrol/issues/10383


junwoo091400:
Pretty useful!

image
image

junwoo091400:
However, the CIs seems to be failing. Could you check that?

booo:
> However, the CIs seems to be failing. Could you check that?

I don't think that has anything to do with the changes. The checkout of eigen lib and something else prior to the actual build process fails.

patrickelectric (approved):

PR #10487, opened 2022-11-15
QGCCameraControl: support ftp download of .xz .lzma compressed cam files

olliw42:
title says it

via mftp, compressed camera definition files with extensions .xml.lzma or .xml.xz can be downloaded, are decompressed, and the .xml stored and used. Works also with the caching.

tested using a STorM32 controller as mavlink camera.

:)

EDIT: here a log screenshot showing the thing in action

qgc-mftp-xz

### DIFF
--- a/src/Camera/QGCCameraControl.cc
+++ b/src/Camera/QGCCameraControl.cc
@@ -12,6 +12,7 @@
 #include "QGCMapEngine.h"
 #include "QGCCameraManager.h"
 #include "FTPManager.h"
+#include "QGCLZMA.h"
 
 #include <QDir>
 #include <QStandardPaths>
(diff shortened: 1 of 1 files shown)


bkueng (approved):

olliw42:
REALLY?
merged without any comments, suggestions, etc?
I am shocked :) positively shocked ... :D

you have noticed the change in behavior concerning ftp and caching? previously it would have always done the ftp download, even if a cached file would have been available. Now it uses the cached file. This made more sense to me, but is a change in behavior. 

OT, but let me grab the opportunity, adding comp_metadata support for components would be a much larger effort with deeper code changes and design decissions. If I were to attempt this how should I approach this? Should I open an issue were I first would outline how it could look like and one can discuss how it should look like, or should I jump forward with a smaller PR preparing first the neccessary grounds? 

many thx, sir.

bkueng:
Haha, yes it made sense so I merged it. Yes I saw the change in behavior.

> OT, but let me grab the opportunity, adding comp_metadata support for components would be a much larger effort with deeper code changes and design decissions. If I were to attempt this how should I approach this? Should I open an issue were I first would outline how it could look like and one can discuss how it should look like, or should I jump forward with a smaller PR preparing first the neccessary grounds?

I think an issue makes sense, as it is a bigger change.

patrickelectric:
This breaks cmake build system, check https://github.com/mavlink/qgroundcontrol/pull/10489 for the fix

PR #10488, opened 2022-11-15
MAVFtp: Allow proper compId as default

olliw42:
this PR allows to specify a default compId in MAVFtp downloads, which can be different from MAV_COMP_ID_AUTOPILOT1. 

As before, if the uri includes a [;comp=xxx], the compId used in the ftp is ovberrited to be that specified in the uri. However, if the uri does not contain a compId specifier, it is in general not approriate to assume MAV_COMP_ID_AUTOPILOT1. For instance, if it is a camera.

The PR corrects the behavior for the camera's. It leaves the behaviour unchanged for the ftp downloads in the ParameterManager and componentInformationManager.

The PR has a tiny second change, namely lines https://github.com/mavlink/qgroundcontrol/pull/10488/files#diff-3c77d06c121d54cdd5191ae158a04af0095c441da288069469a948a48abdfaaeR174-R175.  The additional check for sysid should make sense.

Tested with a STorM32 controller which provides a mavlink gimbal and a mavlink camera, which is connected to a ArduPilot flight controller.

The attached screenshots, which show the log in sequence, show that the camera ftp is corretly adressed at CompId 105, even though the uri doesn't have a comp specifier, and the that autopilot param ftp is correctly adressed at CompId 1.

qgc-ftpcompid-1

qgc-ftpcompid-2

### DIFF
--- a/src/Vehicle/FTPManager.h
+++ b/src/Vehicle/FTPManager.h
@@ -43,7 +43,7 @@ public:
     ///                     a dynamic file creation on the vehicle.
     /// @return true: download has started, false: error, no download
     /// Signals downloadComplete, commandError, commandProgress
-    bool download(const QString& fromURI, const QString& toDir, const QString& fileName="", bool checksize = true);
+    bool download(uint8_t fromCompId, const QString& fromURI, const QString& toDir, const QString& fileName="", bool checksize = true);
 
     /// Cancel the current operation
     /// This will emit downloadComplete() when done, and if there's currently a download in progress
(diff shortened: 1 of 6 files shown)


src/Vehicle/FTPManager.h
```diff
@@ -43,7 +43,7 @@ class FTPManager : public QObject
     ///                     a dynamic file creation on the vehicle.
     /// @return true: download has started, false: error, no download
     /// Signals downloadComplete, commandError, commandProgress
-    bool download(const QString& fromURI, const QString& toDir, const QString& fileName="", bool checksize = true);
+    bool download(uint8_t fromCompId, const QString& fromURI, const QString& toDir, const QString& fileName="", bool checksize = true);
```
bkueng, line 46: Can you add the param to the API docs as well?

bkueng (commented):
Looks good. CI fails due to a type mismatch - it needs an explicit cast.

src/Vehicle/FTPManager.h
olliw42: ahh ... didn't even thought about that LOL
yes, sure

olliw42:
> CI fails due to a type mismatch - it needs an explicit cast.

yo, notived, but I am not sure there the cast is exactly needed (I mean compId = MAV_COMP_ID_AUTOPILOT1 was apparently ok)

should it be

`(fromCompId == (uint8_t)MAV_COMP_ID_ALL) `

?

bkueng:
Probably `(fromCompId == MAV_COMP_ID_ALL) ? (uint8_t)MAV_COMP_ID_AUTOPILOT1 : fromCompId` will fix it.

bkueng (approved):

PR #10489, opened 2022-11-16
CMakeLists: Add missing compression on QGC link

patrickelectric:
Fix cmake build


bkueng (approved):
Thanks.
I generally don't think it makes sense to maintain 2 different build systems.

PR #10490, opened 2022-11-16
Allow selection of first item in ardupilot firmware list

DonLakeFlyer:
Ardupilot firmware flash was broken. You couldn't pick the first item in the firmware list. Especially problematic if there was only one firmware option!


patrickelectric (approved):

DonLakeFlyer:
Thanks @patrickelectric

PR #10502, opened 2022-11-28
Fix issue for Ardupilot Vtols and start mission action:

Davidsastresas:
The current code would set the vehicle to guided, then arm, and if fixed wing, then switch to auto.

This small time it remained in guided, until QGC received feedback of vehicle armed and changed to auto, was enough to provoke dangerous situations, because when arming in guided mode, if the vehicle is a tilt rotor vtol, it would arm the rotors in forward flight position, and some airframes are not designed for this situation, leading to a dangerous situation.

This commit changes the behavior of start mission, so if it is an Ardupilot fixed wing or VTOL it will set auto mode, and then arm, which is the intended way of dealing with this in Ardupilot.

For the rest of vehicles the behavior keeps being the same.

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -890,9 +890,18 @@ void APMFirmwarePlugin::startMission(Vehicle* vehicle)
 
     if (!vehicle->armed()) {
         // First switch to flight mode we can arm from
-        if (!_setFlightModeAndValidate(vehicle, "Guided")) {
-            qgcApp()->showAppMessage(tr("Unable to start mission: Vehicle failed to change to Guided mode."));
-            return;
+        // In Ardupilot for vtols and airplanes we need to set the mode to auto and then arm, otherwise if arming in guided
+        // If the vehicle has tilt rotors, it will arm them in forward flight position, being dangerous.
+        if (vehicle->fixedWing()) {
+            if (!_setFlightModeAndValidate(vehicle, "Auto")) {
+                qgcApp()->showAppMessage(tr("Unable to start mission: Vehicle failed to change to Auto mode."));
+                return;
+            }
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
```diff
@@ -890,9 +890,18 @@ void APMFirmwarePlugin::startMission(Vehicle* vehicle)
...
     if (!vehicle->armed()) {
         // First switch to flight mode we can arm from
-        if (!_setFlightModeAndValidate(vehicle, "Guided")) {
-            qgcApp()->showAppMessage(tr("Unable to start mission: Vehicle failed to change to Guided mode."));
-            return;
+        // In Ardupilot for vtols and airplanes we need to set the mode to auto and then arm, otherwise if arming in guided
+        // If the vehicle has tilt rotors, it will arm them in forward flight position, being dangerous.
+        if (vehicle->fixedWing()) {
+            if (!_setFlightModeAndValidate(vehicle, "Auto")) {
+                qgcApp()->showAppMessage(tr("Unable to start mission: Vehicle failed to change to Auto mode."));
```
booo, line 897: Do we need a change to the translation files  for this?

booo:
I assume you tested this somehow?

Davidsastresas:
Indeed, it was tested in Ardupilot SITL. It only changes the logic for fixed wing and VTOL, this is the way a start mission command should work for those vehicles. In SITL it works as it should.

src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
Davidsastresas: Good point. it is reusing a message already present, but we probably need to update translation files. Let me a couple days to fix it. Thanks for spotting it!

PR #10509, opened 2022-12-07
PlanManager: drop spurious items for unexpected mission types

maaaat:
With the short timeout on requesting mission items, some items are requested multiple times. That's usually no problem as spurious items are ignored. However if we moved on to requesting items for a different mission type, say fences, then a spurious duplicate would confuse the manager who would just accept it, and later complain that the command is invalid for this mission type.

The root cause is that the mission type of incoming mission items is not checked, which is an easy fix.

### DIFF
--- a/src/MissionManager/PlanManager.cc
+++ b/src/MissionManager/PlanManager.cc
@@ -389,9 +389,9 @@ void PlanManager::_handleMissionItem(const mavlink_message_t& message)
     mavlink_mission_item_int_t missionItem;
     mavlink_msg_mission_item_int_decode(&message, &missionItem);
 
-    command =       (MAV_CMD)missionItem.command,
-            frame =         (MAV_FRAME)missionItem.frame,
-            param1 =        missionItem.param1;
+    command =       (MAV_CMD)missionItem.command;
+    frame =         (MAV_FRAME)missionItem.frame;
+    param1 =        missionItem.param1;
     param2 =        missionItem.param2;
     param3 =        missionItem.param3;
     param4 =        missionItem.param4;
--- a/src/MissionManager/PlanManager.cc
+++ b/src/MissionManager/PlanManager.cc
@@ -373,24 +373,26 @@ void PlanManager::_requestNextMissionItem(void)
 
 void PlanManager::_handleMissionItem(const mavlink_message_t& message)
 {
-    MAV_CMD     command;
-    MAV_FRAME   frame;
-    double      param1;
-    double      param2;
-    double      param3;
-    double      param4;
-    double      param5;
-    double      param6;
-    double      param7;
-    bool        autoContinue;
-    bool        isCurrentItem;
-    int         seq;
+    MAV_CMD          command;
+    MAV_FRAME        frame;
+    MAV_MISSION_TYPE missionType;
+    double           param1;
+    double           param2;
+    double           param3;
+    double           param4;
+    double           param5;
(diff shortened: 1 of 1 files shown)


booo:
Looks good to me.

Do you think it makes sense to increase the timeout? How many duplicate items do you see?

maaaat:
In my case I see quite a lot of duplicates, by just eyeballing the logs I'd guess about 33%. I did raise the timeout from 250ms to 300ms and all the duplicates disappeared. I did measure the average time between request and response and if my memory isn't failing it was around 270ms or something like that. I'd be in favour of raising the timeout to 300ms.

booo:
> In my case I see quite a lot of duplicates, by just eyeballing the logs I'd guess about 33%. I did raise the timeout from 250ms to 300ms and all the duplicates disappeared. I did measure the average time between request and response and if my memory isn't failing it was around 270ms or something like that. I'd be in favour of raising the timeout to 300ms.

With what setup did you see these results?

maaaat:
Oh it's a custom setup, a prototype with custom avionics, custom ground segment, and some 868MHz radios. I don't know if the PX4 or APM have similar behaviours, I don't have the hardware with me to test that. Would be cool to be able to change the timeout value in custom builds from one of the plugin interfaces, but that would be for a different PR.

patrickelectric (approved):

Davidsastresas (approved):

av-maurits:
I believe this PR results in a bug with PX4 v1.13.0 (haven't tested for higher versions). When sending a MISSION_REQUEST_INT for the mission type MAV_MISSION_TYPE_FENCE, the returned MISSION_ITEM_INT has a mission type that is always labeled as MAV_MISSION_TYPE_MISSION, irrespective of the requested mission type. This PR ensures that any existing geofence will therefore not be processed anymore. 

To reproduce, use QGC v4.2.7, and PX4 v1.13.0 (I run the latter using a Gazebo Docker container). Create a geofence using QGC. Close QGC and reopen. It will display 'GeoFence transfer failed. Error: Mission read failed, maximum retries exceeded.' and 'Operation timeout, aborting transfer'. 

We originally encountered this issue in the field as an old geofence still present in PX4 could not be displayed or removed in QGC. This can unexpectedly trigger failsafes.

benjinne:
> Update: this [issue](https://github.com/PX4/PX4-Autopilot/issues/22069) got fixed in PX4 and will probably be included in 1.14.1.
> 
> I believe this PR results in a bug with PX4 up to v1.14.0. When sending a MISSION_REQUEST_INT for the mission type MAV_MISSION_TYPE_FENCE, the returned MISSION_ITEM_INT has a mission type that is always labeled as MAV_MISSION_TYPE_MISSION, irrespective of the requested mission type. This PR ensures that any existing geofence will therefore not be processed anymore.
> 
> To reproduce, use QGC v4.2.7, and PX4 v1.13.0 (I run the latter using a Gazebo Docker container). Create a geofence using QGC. Close QGC and reopen. It will display 'GeoFence transfer failed. Error: Mission read failed, maximum retries exceeded.' and 'Operation timeout, aborting transfer'.
> 
> We originally encountered this issue in the field as an old geofence still present in PX4 could not be displayed or removed in QGC. This can unexpectedly trigger failsafes.

I'm observing the same behavior and opened an issue for it:
https://github.com/mavlink/qgroundcontrol/issues/10856

PR #10514, opened 2022-12-10
README: Remove obsolete badges

booo:
We use github actions for CI now. The travis and appveyor badges are obsolete.

Do people use gitter?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,6 @@
 
 [![Releases](https://img.shields.io/github/release/mavlink/QGroundControl.svg)](https://github.com/mavlink/QGroundControl/releases)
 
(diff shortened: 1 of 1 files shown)


bkueng (approved):
Gitter can probably go as well.

zdanek:
And why don't we use those badges any more? Is that because building was moved to Github Actions?

booo:
Yes, for the build process we use github actions now. I mentioned that in the commit body.

booo:
> Gitter can probably go as well.

I removed the gitter badge as well. Maybe we should remove/archive the gitter channel(s) and remove all reference from the documentation too?

patrickelectric (approved):

PR #10515, opened 2022-12-11
remove old qtlocation files

holden-zenith:
Removes QtLocation files for an older version of Qt.


booo:
`git grep QtLocationPlugin` find quite a few references to the path you removed. Can you check on that?

And maybe add one or two lines of explanation to your commit message even tough it's quite obvious what we do.

Nice work!

holden-zenith:
All I see are references to the parent directory QtLocationPlugin, instead of the subdirectory qtlocation that I deleted.

booo:
> All I see are references to the parent directory QtLocationPlugin, instead of the subdirectory qtlocation that I deleted.

You are right. Sorry, I got confused by all the directories and subdirectories :)

junwoo091400:
LGTM :thinking:

patrickelectric (approved):

PR #10516, opened 2022-12-11
CI: Remove obsolete Jenkinsfile

booo:
We use github actions for CI now so Jenkins is obsolete. It is part of the git history in case we need a backup.


junwoo091400 (approved):
Nice!

patrickelectric (approved):

PR #10517, opened 2022-12-11
CI: Remove travis CI related configuration

booo:
We use github actions now and I remove the travis configuration file and change the Vagrantfile that uses the travis file for configuration. I remove build_ios script as well as it is nowhere in use and seems to be part of the travis CI setup.


booo:
The change to the Vagrantfile is untested. I hope the syntax is right.

booo:
Is somebody using the Vagrantfile? Otherwise I suggest we remove this as well.

Davidsastresas:
Just a note, regarding the build_ios, it might be worth leaving it just in case. If I am not mistaken, latest QGC is not being uploaded to the appstore neither, so maybe that script comes handy if at some point we tackle that task, what do you think?

booo:
> Just a note, regarding the build_ios, it might be worth leaving it just in case. If I am not mistaken, latest QGC is not being uploaded to the appstore neither, so maybe that script comes handy if at some point we tackle that task, what do you think?

I would rather document the release build process somewhere and/or add this to a github action.

mrpollo:
@booo did you add a pr documenting this yet? should we go ahead and merge?

booo:
> @booo did you add a pr documenting this yet? should we go ahead and merge?

I did not document the release process or added additional information. But anyway I think this is fine to merge.

booo:
I merge this now. If we need the code/script it's archived in git for later generations.

PR #10522, opened 2022-12-19
Metadata translation support

bkueng:
This implements https://github.com/mavlink/mavlink/pull/1934

qgc_metadata_translation

### DIFF
--- a/UnitTest.qrc
+++ b/UnitTest.qrc
@@ -15,6 +15,8 @@
 	<file alias="PolygonBadXml.kml">src/MissionManager/UnitTest/PolygonBadXml.kml</file>
 	<file alias="PolygonBadCoordinatesNode.kml">src/MissionManager/UnitTest/PolygonBadCoordinatesNode.kml</file>
 	<file alias="MockLinkOptionsDlg.qml">src/comm/MockLinkOptionsDlg.qml</file>
+	<file alias="TranslationTest.json">src/qgcunittest/TranslationTest.json</file>
+	<file alias="TranslationTest_de_DE.ts">src/qgcunittest/TranslationTest_de_DE.ts</file>
     </qresource>
     <qresource prefix="/qml">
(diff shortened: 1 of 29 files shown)


booo:
We need some documentation on this feature.

What happens if there is no internet? Can we download information in advance? How? Where does the system download stuff too.

bkueng:
Yes this will be done in mavlink.

There's a 3 day cache implemented, and once expired it will check for updates and if there's no internet, it will still use the cache. So you need internet at least once.

booo:
I think we should add some minimal documentation to QGC documentation too. Describe how QGC behaves and how the caching works.

So for warming the cache I have to connect the FC, let QGC connect to the internet and download the translation for the specific FC? If I need multiple languages I have to change the language and let QGC download again?

A small tool to pre-warm the cache for a matrix of FC and languages would be nice.

bkueng:
> So for warming the cache I have to connect the FC, let QGC connect to the internet and download the translation for the specific FC? If I need multiple languages I have to change the language and let QGC download again?

Yes and yes. We can certainly extend that if you're interested in doing so. But first we'll need some translations.

mrpollo:
hey @bkueng what's the state of this PR? looks useful

bkueng:
Hamish is going through the server/CI setup. It's somewhat independent, but I'd like to get the whole pipeline in place before merging this.

booo:
I would like to see some documentation in the QGC documentation too before we merge. I think it is a useful feature but if we don't describe how it works and how to pre-warm the cache it will only be accessible to the experts.

bkueng:
@booo I added a short section: https://github.com/mavlink/qgc-user-guide/pull/394

PX4 has been updated, so this is getting ready @mrpollo.

mavlink docs: https://github.com/mavlink/mavlink-devguide/pull/498

mrpollo (approved):
super clean, looks good, thanks @bkueng

PR #10525, opened 2022-12-23
FactSystem: ParameterManager: Remove unused variables

patrickelectric:


Davidsastresas (approved):

Davidsastresas:
I've seen _jsonParameterskey is unused as well, shall we remove it as well?

we have another _jsonParameterskey under compInfoParam, and that is used indeed, but not the one in parameterManager.

Thanks!

patrickelectric:
@Davidsastresas done.

PR #10540, opened 2023-01-13
FlightMap: PhotoVideoControl: Allow to take photos and record from any camera

patrickelectric:
QGC is capable to record and capture frames from any camera.
The PR unblocks the controls and break in two difference shot buttons.


booo:
We should focus on a new release. I don't think we should backport/implement more and more stuff into old versions of QGC. And if we backport/implement stuff we should use the right versions. 4.3 would be more appropriate for new features.

If you consider this a bug fix it's fine with me.

booo:
@patrickelectric What do we do about this one?

DonLakeFlyer:
Don't understand if this is a port from master to stable? If so, new stable just went out so this isn't needed.

Or if this is a change which as made only to a stable. If so, this needs to get into master.

Re-open if this is a real change.

PR #10543, opened 2023-01-16
Joystick: Enable button up signal for hat buttons

kjersbry:
Releasing a hat button did not lead to _executeButtonAction being called with BUTTON_UP. This adds the hat buttons to lastBbuttonValues, such that the button up transitions for hat buttons are recorded.


mrpollo:
thanks @kjersbry 🎉

PR #10546, opened 2023-01-17
Partially update mavlink submodule to 97c50e128768c9b95c5542b8fc751198b1af96da

Davidsastresas:
I am working on Remote ID integration for QGC, and the message OPEN_DRONE_ID_ARM_STATUS was missing.

I checked that it was already merged in master mavlink, also in c_library_v2 master, although this commit doesn't update mavlink submodule to master, just to the commit when this message was added.

I tried real quick to update it to master, but there were some more conflicts, and I prefer to spend the time right now on finishing the Remote ID implementation, as no other part of the code ( as far as I know ) needs newer mavlink. We can work later on on updating mavlink to the latest version.

I never updated mavlink submodule upstream before. I placed mavlink submodule HEAD in the newer commit, built the project and checked that no conflicts arose, and commited the change. Please let me know if it should be done differently and I will modify the PR. Thanks!

@bkueng @dagar I saw you updated mavlink in QGC the latest times. I would appreciate very much if you could confirm the above is the correct approach. Thanks!


booo:
I would prefer a stable release of the mavlink repo instead. Is there a minor release that works for you?

Davidsastresas:
Sure, that makes sense. I will investigate this week which one is the closest stable and I will fix this PR. Thanks!

mrpollo:
Closed in favor of https://github.com/mavlink/qgroundcontrol/pull/10600

PR #10549, opened 2023-01-26
Update QGCMapUrlEngine.cpp

Tevhit:
Labels - Terrain in _providersTable


booo:
Please provide some context why this change is needed and add it to the commit message.

PR #10556, opened 2023-02-02
Display Rangefinder Distance

AndKe:
this adds the option to display the telemetry data from RANGEFINDER.distance 
tested, works fine.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -80,6 +80,7 @@ const char* Vehicle::_yawRateFactName =             "yawRate";
 const char* Vehicle::_airSpeedFactName =            "airSpeed";
 const char* Vehicle::_airSpeedSetpointFactName =    "airSpeedSetpoint";
 const char* Vehicle::_xTrackErrorFactName =         "xTrackError";
+const char* Vehicle::_rangeFinderDistFactName =     "rangeFinderDist";
 const char* Vehicle::_groundSpeedFactName =         "groundSpeed";
 const char* Vehicle::_climbRateFactName =           "climbRate";
 const char* Vehicle::_altitudeRelativeFactName =    "altitudeRelative";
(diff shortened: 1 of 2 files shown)


Davidsastresas:
Looks good to me. Just out of curiosity, which firmware are you using? Usually for these matter the distance sensor message is used instead.

AndKe:
Ardupilot - both Copter and Plane emit RANGEFINDER.distance

AndKe:
Thank you. If you have a clue where the issue with this one can be, please tell.  https://github.com/mavlink/qgroundcontrol/issues/10557
If it's trivial, I could maybe fix it myself ... as I would like this to work.

DonLakeFlyer:
So this breaks the usage of common as opposed to ardupilot mavlink dialect support. Which will in turn break custom builds which are PX4 only. The reason being that this mavlink message is only in the ardupilot dialect and not in common. All of this code should be #ifdef'ed in/out based on the NO_ARDUPILOT_DIALECT DEFINE. Look in Vehicle.cc for other examples. This will need to be fixed before a new Stable goes out so we don't screw commerical folks with custom builds.

AndKe:
@DonLakeFlyer  I'll correct that as you requested - I guess you are saying it should be similar to this one:
```
#if !defined(NO_ARDUPILOT_DIALECT)
    void _handleCameraFeedback          (const mavlink_message_t& message);
#endif
```
I am surprised to see compiler stuff like this - does this mean QGC has some PX4-only builds? - what is the purpose/practical  of this?

DonLakeFlyer:
> does this mean QGC has some PX4-only builds?

Not in genral public usage. But custom builds can be created which are PX4 or ArduPilot only. So if you are a commercial vehicle business and have a custom version of QGC. You can build it to only the firmware you use on the vehicle. Cuts down on app size quite a bit and also simplifies some of the ui since it's single firmware.

Davidsastresas:
> > does this mean QGC has some PX4-only builds?
> 
> Not in genral public usage. But custom builds can be created which are PX4 or ArduPilot only. So if you are a commercial vehicle business and have a custom version of QGC. You can build it to only the firmware you use on the vehicle. Cuts down on app size quite a bit and also simplifies some of the ui since it's single firmware.

I wonder if we could instead move that to firmware plugin, instead of relying on a #define? It will not be straightforward as right now a lot of stuff in the frontend is using this from vehicle itself, but maybe in the long term it is worth spending time now on really making QGC firmware agnostic for main architecture? Any thoughts @DonLakeFlyer ?

DonLakeFlyer:
In general I had stopped supporting any ArduPilot specific messages in QGC code. It just got all out of hand. Instead I always asked them to move them to common dialect. The whole concept of dialect specific generic messages like this makes no sense to me.

AndKe:
Before this PR, the widget did not support any rangefinder info, so I don't see this as a conflict. 
The same function that is populating rangefinder data, could very well read both RANGEFINDER and DISTANCE_SENSOR packets totally unaware of which is in use as no system is emitting both.
Penultimately, it would be nice to be able to select *any* mavlink data for display, not only the preset sources.

PR #10558, opened 2023-02-03
Fix Kakute H7 v2 & Kakute H7 Mini Board ID error on QGC

vincentpoont2:
Fix Board ID Mistake according to https://github.com/PX4/PX4-Bootloader/pull/224/files


vincentpoont2:
@mrpollo Can we merge this?

PR #10565, opened 2023-02-10
FirmwareUpgrade: add CubeOrange+

julianoes:
This is according to https://github.com/PX4/PX4-Bootloader/blob/main/board_types.txt

FYI @bugobliterator


julianoes:
Self merging, given no one seems to have an opinion.

PR #10572, opened 2023-02-14
Topomap of Svalbard

AndKe:
Added another map by Statkart - with topographic maps of Svalbard.

### DIFF
--- a/src/QtLocationPlugin/GenericMapProvider.cpp
+++ b/src/QtLocationPlugin/GenericMapProvider.cpp
@@ -65,6 +65,14 @@ QString StatkartMapProvider::_getURL(const int x, const int y, const int zoom, Q
     return StatkartMapUrl.arg(zoom).arg(x).arg(y);
 }
 
+static const QString StatkartBaseMapUrl = QStringLiteral("http://opencache.statkart.no/gatekeeper/gk/gk.open_gmaps?layers=norgeskart_bakgrunn&zoom=%1&x=%2&y=%3");
+
+QString StatkartBaseMapProvider::_getURL(const int x, const int y, const int zoom, QNetworkAccessManager* networkManager) {
+    Q_UNUSED(networkManager)
+    return StatkartBaseMapUrl.arg(zoom).arg(x).arg(y);
+}
+
+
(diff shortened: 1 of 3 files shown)


booo:
Could you please add a few lines of documentation to your commit message or code, e.g. what kind of map it is, who operates the map and is it free for public use.

Out of curiosity what do you do with UAVs in Svalbard? :cold_face: 

Thanks for your contribution.

AndKe:
@booo 
This map provides topographic information for Svalbard.
We have been flying fixed wing to map how ice moves.
the past years, I've flown Multicopter (X8) , MTOW 50kg , AUW25kg with wideband radar to map snow layers, glacier cracks, and now (in a week) polar bear dens.
So it would be very helpful if my PR's got merged :) 
My other PR is about rangefinder-monitoring - which is useful where terrain height is .. unpredictable due to snow.
Ardupilot has been to the north pole with us as well, so Svalbard is not that far north :)

booo:
Could you add the bits of information about the map to your commit message so they don't get lost? I would much appreciate that.

Unfortunately I only have permission to close tickets :/

I would like to contact you about your activities if that is fine with you. Our association has a request for similar activities in Alaska.

booo:
Apart from my remark the change looks good to me!

AndKe:
@booo  I think I fixed the comment.
Are you on the ArduPilot Discord server? my nickname is  "AndKe" - or - feel free to email me if you have/see my email already.

booo:
I don't understand why we hard code so much configuration information here. Maybe it is nice to change that in the future and add a configuration somewhere editable by the user.

AndKe:
@booo  Indeed, especially when MAVProxy has this super-clean way of adding maps - one line, done.  In comparsion: QGC takes 20lines across three files to achive the same.

Penultimately, we would love to see an option to specify different ESPG (projection) too.

booo:
> @booo Indeed, especially when MAVProxy has this super-clean way of adding maps - one line, done. In comparison: QGC takes 20 lines across three files to achieve the same.
> 
> Penultimately, we would love to see an option to specify different ESPG (projection) too.

Could you add a feature request to the bug tracker for that? Would be great. PR is welcome as well :)

AndKe:
The feature request I can do - clean up and add alternative projections is way beyond my skill level.

PR #10575, opened 2023-02-15
Vehicle: use double precision for COMMAND_INT

julianoes:
The whole point of COMMAND_INTs is that they have better resolution for coordinates leveraging int32 for degrees*10^7. This resolution is lost if we use float while handling the data instead of double.


booo:
Sounds reasonable and looks good.

To ease the understanding of the change you could have moved parameter access change to another commit.

booo:
Thanks for your contribution.

julianoes:
Thanks @booo! :heart:

PR #10583, opened 2023-02-18
Fix Analyze Page popup ownership

DonLakeFlyer:
Do this:
* Analyze View
* Mavlink Inspector
* Pop it out to a window
* Back from Analyze View
* Open Vehicle Setup
* You'll see that that mavlink inspector stops showing values

This was caused by the fact that the popout window was created in the context of the Analyze VIew. When that view went away all it's child objects are killed as well. Changed to create analyze popouts in the context of the main ui window.

### DIFF
--- a/src/AnalyzeView/AnalyzeView.qml
+++ b/src/AnalyzeView/AnalyzeView.qml
@@ -127,43 +127,8 @@ Rectangle {
         property string title
 
         Connections {
-            target: panelLoader.item
-            onPopout: {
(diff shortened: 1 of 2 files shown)


Davidsastresas:
Nice catch!

DonLakeFlyer:
Sorry for going around the process but I'm going to merge this in. It is hugely broken and I'm hitting myself in trying to move forward with new stuff in QGC. So...

Issue #10364, opened 2022-08-11
Missing DLL files on Windows.


ES-Alexander:
## Expected Behavior
QGC should be runnable once installed.

## Current Behavior
On a fresh Windows OS, QGC fails to run due to missing `.dll` files (specifically `VCRUNTIME140_1.dll` and `MSVCP140_1.dll`).

## Steps to Reproduce:
1. Fresh install Windows
2. Install QGC
3. Fail to open QGC due to missing DLL warnings

## System Information
- Operating System: Windows 10, 11
- QGC Version: 4.2.3, and presumably many earlier versions
- QGC build: stable
- Flight Controller: N/A
- Autopilot (with version): N/A

## Detailed Description
Issue raised in [this forum post](https://discuss.bluerobotics.com/t/qgroundcontrol-missing-dll-files/12559).

An existing workaround is to install the [Microsoft Visual C++ Redistributables](https://docs.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170), which some other applications just provide as an optional step in their installation processes, but as per [this resolution](https://github.com/bluerobotics/ping-viewer/issues/1018#issuecomment-988084847) it's also possible to bundle the relevant `.dll` files with the application instead of relying on the existence of a shared set used throughout the computer.

If continuing to use the shared files is preferred, it should at least be documented in [the installation instructions](https://docs.qgroundcontrol.com/master/en/getting_started/download_and_install.html) as a troubleshooting note or similar.

PR #10588, opened 2023-02-20
Github actions: Ship missing Windows DLLs

joaoantoniocardoso:
Closes https://github.com/mavlink/qgroundcontrol/issues/10364


DonLakeFlyer:
I believe there are redist vc runtime installers that you are supposed to use (https://learn.microsoft.com/en-us/cpp/windows/deploying-visual-cpp-application-by-using-the-vcpp-redistributable-package?view=msvc-170). I'm not sure just copying this specific set of DLLs will work in all case. Seems like the redist installers should be made part of the NSIS installation process.

joaoantoniocardoso:
> I believe there are redist vc runtime installers that you are supposed to use (https://learn.microsoft.com/en-us/cpp/windows/deploying-visual-cpp-application-by-using-the-vcpp-redistributable-package?view=msvc-170). I'm not sure just copying this specific set of DLLs will work in all case. Seems like the redist installers should be made part of the NSIS installation process.

Usually (but not always), the build process uses the redis installed at system32, then the easiest way to ship the right version would be by directly copying it, with no need to use the Microsoft installer.

The exceptions are probably when we force by passing a different location, or the system32 path doesn't have, then it will search in other places...

I was hoping that the CI would build it for me :eyes: as I don't have a windows virtual machine with a proper development environment configured to test this build on it... Any idea why it didn't run?

DonLakeFlyer:
Ok, sounds good then. MY guess is your windows yaml isn't parsing due to syntax error. I think there is a github workflow syntax checker somewhere on the web.

DonLakeFlyer:
So I forgot how all this stuff works in the Windows build. If you look at QGCPostLinkCommon.pri you'll see where this should go. Not in the github workflow.

joaoantoniocardoso:
Easier than before, thanks :)

I'll make it ready for review after testing, but feel free to merge if you test it before me

mrpollo:
Hey @joaoantoniocardoso @DonLakeFlyer have you had a chance to test this yet?

mrpollo:
No problem @DonLakeFlyer I only posted because I saw your comment from earlier

joaoantoniocardoso:
@mrpollo @DonLakeFlyer Hi, I've been struggling to get the time to work on this, but I suppose it's not working - the CI is not building for windows. 

It seems that `libs/Microsoft/windows/` doesn't have the two additional DLLs (`vcruntime140_1.dll` and `msvcp140_1.dll`) yet, so I've just added a commit with the files (now from version 17, because the `vcruntime140_1.dll` is only available after 17). Here are the hash and versions from `peres`:

image

joaoantoniocardoso:
I was testing the original windows binary (downloaded from [here](https://docs.qgroundcontrol.com/master/en/getting_started/download_and_install.html)) and noticed that we are also shipping a redis installer:

image

After the installation, the executable is at our installed folder, and by running it we can see that the version is the latest 2019 (14.29.30139.0):

image

Even after I'd reran the installer, I searched in the system for `vcruntime140_1.dll` and I couldn't find anything new. I've also manually downloaded the latest redis installer (2022) from Microsoft, but it didn't helped.

---

With the image built from the CI of this PR, it ships the required `vcruntime140_1.dll`, and checking it with [Dependencies](https://github.com/lucasg/Dependencies), it seems to have all that it needs to run:

image

---

Despite that, on my clean Windows 10 x64 in a Virtualbox, I'm not able to run the QGC by any of its shortcuts, neither in the original installation nor after this PR, which might be a bug or some kind of incompatibility with the VM itself, so it would be good if someone else test the application in a non-virtual machine.

With all that, I think this PR is ready for reviews, thanks

joaoantoniocardoso:
> So if the redist installer doesn't install this and adding it manually doesn't fix the problem then why add this manually to QGC?

Sorry, I think I wrote it in a confusing way, but _shipping the missing DLLs do fix [this DLL missing issue](https://github.com/mavlink/qgroundcontrol/issues/10364)._

Despite that, I can't fully test the resulting binary because QGC is not opening on my virtual machine, the QGC process is running, but the GUI never shows up. If QGC is supposed to work in virtual machines, we might have just discovered a new issue, which is (probably) unrelated to this PR.

DonLakeFlyer:
Ah, I understand now. QGC should work fine in a VM. I've been using Windows in a VM for almost my entire development time with QGC. That said, the current version might be broken in a VM. If so it usually has something to do with the video driver. It does seem from other posts I've seen that the Windows version has gone a bit downhill and isn't working on various versions of windows.

@booo @Davidsastresas Something that is going to need looking into...

booo:
Supporting windows not really my top priority ;)

But I can test on an old machine in the next weeks maybe. I can't do more than startup of the QGC binary and a minimal sanity check of the functionality.

Davidsastresas:
I don't have time to test this on a fresh windows install ( I think that was the origin of all this ), but I just tested the artifact generated on the latest commit here, and it works perfectly fine on my windows 10 partition. I connected it to SITL and did some basic sanity check, and everything was working as expected, and I didn't see any error anywhere.

Is there anything else we should double check before merging? I am not particularly familiarized with windows neither. Thanks @joaoantoniocardoso  for working into this!

DonLakeFlyer:
I would merge and see what happens with Stable RC folks.

Davidsastresas:
@DonLakeFlyer I agree. We have 2 reports of this working nicely on windows, so I think what you suggested is the best approach to find out, and it should be safe enough. So, I merged it! Thanks!

PR #10594, opened 2023-02-22
Missionitem: Set condition_yaw default to absolute heading

AndKe:
I find myself changing this every time. In almost all scenarios I've been flying, I prefer to change heading to absolute direction, rather than relative (because relative will accumulate any heading error, especially if the command is executed while turning.)
The absolute heading is the most used/typical way of using this mission item.

### DIFF
--- a/src/MissionManager/MavCmdInfoCommon.json
+++ b/src/MissionManager/MavCmdInfoCommon.json
@@ -457,7 +457,7 @@
                 "label":            "Offset",
                 "enumStrings":      "Relative,Absolute",
                 "enumValues":       "1,0",
-                "default":          1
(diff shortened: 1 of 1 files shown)


booo:
Commit message should be "Missionitem: Set condition_yaw default to absolute heading".

This changes behavior of QGC. Cloud you please explain how this affects users that are used to the old defaults and maybe add this to the message body too.

AndKe:
@booo I've made the changes you requested :)

DonLakeFlyer:
Seems reasonable to me.

booo:
Is this something we should document when we do a new release? In the end it is a change to default behavior.

Issue #10593, opened 2023-02-21
Offline maps: cosmetic issues.


AndKe:
Tiles count does not fit:

![Screenshot from 2023-02-20 23-21-53](https://user-images.githubusercontent.com/1468842/220285309-46b8a3fc-39fb-4eaa-b223-eb02b9ccddae.png)

Apart from "default map" - only the last download gets a green dot - all other are red.  - the "green dot" - whitch I guess means "completed/fetched" is ...not stored.
![Screenshot from 2023-02-21 07-39-10](https://user-images.githubusercontent.com/1468842/220285362-55bb1871-c114-48db-a20d-72b3dd43e155.png)

There is tons of screen area, no need to use such a narrow column.
![Screenshot from 2023-02-21 07-50-45](https://user-images.githubusercontent.com/1468842/220285613-aeab6010-c472-47bb-90c0-a64a9ff23bc9.png)

Davidsastresas:
This was addressed here: #10598 

It should fix the text overlapping. For the buttons going green or not, they will go green when the whole tileset is downloaded. From your screenshot it seems the tilesets labeled in red have not been completely downloaded, thus they are red.

On my screenshots on the PR you can see it worked nicely for me. Maybe you got bad connection or something and they were not able to complete.

Please report back and close the issue if the PR fixes it for you. Thanks for reporting!

AndKe:
@Davidsastresas please let me show you:

[Screencast from 2023-02-22 23-08-20.webm](https://user-images.githubusercontent.com/1468842/220771163-bb2b0781-71c3-47d3-9972-be14a2f73a43.webm)

In other words: after download - they are green and fine, successfully completed. Then I restart QGC, and all are red - but eager to become green is "resumed" with 0 tiles.  Hence my idea about the "dot-status" not being stored.

Davidsastresas:
Sorry, I didn't understood correctly. Just tested, it is working fine for me. It is funny because a couple of months back a partner of mine reported weird stuff happening on QGC and ubuntu 22. We took a look real quick, it wasn't making sense the time we were putting into it so he just rolled back to 20.04 in order to keep working. I just tested on my machine, ubuntu 20.04 and it is working as expected:

https://drive.google.com/file/d/1iwMxc4eBWqv_Oi_mb3pRkKxDXps-4soL/view?usp=sharing

So maybe there is something in ubuntu 22 that is conflicting with QGC some way.

AndKe:
[...]

PR #10598, opened 2023-02-22
OfflineMapButton: fix text overlapping on long tileset descriptions

Davidsastresas:
This addresses #10593

text wrap was added, and also the possibility to occupy 2 lines in case the tileset title is too long.

This is how it looked before the PR:

Screenshot from 2023-02-22 22-57-29

and this is how it looks after the PR:

Screenshot from 2023-02-22 22-57-56


booo:
Makes sense. Looks good. Thanks for your contribution.

PR #10600, opened 2023-02-24
Include support for Remote ID

Davidsastresas:
This PR adds support for remote ID on QGroundControl.

In order to enable, under general settings -> miscellaneous settings there is a new checkbox "enable remote id". 

photo_6037317610116332027_y

After enabling it, a new menu will appear in the left toolbar. ( blue arrow in above picture ).

When a remote ID enabled vehicle is connected, we will see a top toolbar indicator as well:

Screenshot from 2023-02-24 04-53-13

On this indicator we can see which elements of remote ID are not ready. Clicking on them, on this indicator, will open up the Remote ID settings menu ( the same is accessible through the left toolbar under QGC settings):

Screenshot from 2023-02-24 04-44-18

On this menu, besides scrolling, if we click on the indicators on top it will automatically scroll to the submenu we clicked. Here we can configure all the remote ID configuration relative to the GCS. Please note that here the available settings change according to region. There are some differences about what is mandatory and what is not between FAA and EU, and it is taken into account on this implementation.

2 additional commits were added after some feedback from Bluemark innovations, they were kept separated in case in the future we need to go over them again. These commits are:

- Allow basic ID sending to be optional: As the recommended approach is to set this up on the RID device itself. If RID doesn't have a valid basic ID it will complain anyway and we will see it in the menu explained above.
- Allow to clear emergency state: After declaring an emergency, long press emergency button during 3 seconds reverts back this state, it is indicated in the button label.

It has been tested and confirmed to be working nicely on Windows, Linux and Android. On windows and Linux it was tested using a USB GPS, in android the built in GPS of the device.

At the architecture level, we created a remoteidmanager class, and it is included in vehicle. Vehicle will forward mavlink messages to this remoteidmanager class, and the UI will take the remoteidmanager of the active vehicle, so it should work with multivehicle as well.

There is also an additional commit at the beginning updating mavlink submodule. There was a PR already opened about this #10546 and @booo suggested to stick to a minor stable release of mavlink instead. I wasn't able to identify it on the mavlink repo, as tags stopped being updated long time ago. If somebody can point me to where I can find the versioning on mavlink I will fix it.

This PR was a joined effort between Lincesystems, Harris Aerial and Vision Aerial. Special thanks as well to my partner @DiegoJRAleixandre who started this implementation, and to Bluemark innovations for the support during this development.


booo:
> There is also an additional commit at the beginning updating mavlink submodule. There was a PR already opened about this #10546 and @booo suggested to stick to a minor stable release of mavlink instead. I wasn't able to identify it on the mavlink repo, as tags stopped being updated long time ago. If somebody can point me to where I can find the versioning on mavlink I will fix it.

Could you ask in the mavlink source repo if they plan to introduce releases again? Seems to me quite strange that they have no release cycle at all for such an important part of the ecosystem.

Davidsastresas:
> > There is also an additional commit at the beginning updating mavlink submodule. There was a PR already opened about this #10546 and @booo suggested to stick to a minor stable release of mavlink instead. I wasn't able to identify it on the mavlink repo, as tags stopped being updated long time ago. If somebody can point me to where I can find the versioning on mavlink I will fix it.
> 
> Could you ask in the mavlink source repo if they plan to introduce releases again? Seems to me quite strange that they have no release cycle at all for such an important part of the ecosystem.

I agree, there must be something, but I wasn't able to figure it out. I opened an issue in mavlink repo, I will ask on discord as well. Thanks for checking this out!

Davidsastresas:
@booo I asked this on mavlink repo issues, and also in discord threads. It is confirmed mavlink doesn't have versioning. So I guess any commit we pick from its repo is "stable". Auturgy said this is kind of nice because it forces mavlink to be backwards compatible. It seems both PX4 and Ardupilot just pick the latest mavlink commit they are happy with, and go with that.

The optimal would be for sure to include latest mavlink. I tried to do that but some messages changed their fields, so some extra adaptation was needed all over QGC to make this change. I will go through that but I think we have some more important matters to pay attention to first.

So, I wanted to ask your opinion, should we leave this mavlink update for the moment and go forward merging this? Several people has already tested this one, it seems to be working all right. Ardupilot has this part already implemented on the codebase, and it seems PX4 is moving towards adding support for this as well.

Let me know your thoughts. Thanks!

Davidsastresas:
Just saw @mrpollo requested review from @bkueng. Let me know if you need anything on this front, I am available anytime on discord. If there is no objection, I would like to push this within a reasonable timeframe. Is it fine If I mark 10 days from now to merge this if there are no mayor drawback about it until then? Thanks.

optical-o:
@Davidsastresas Great work on the on the initial support for the QGroundControl i have tested the feature with Ardupilot and Dronetag DRI. Hope you find my testing helpful.

Only note that i have found is that the ground station is sending the mavlink messages as broadcasts, which i'm not sure is intended. Because the drone will forward these messages to the ODID transmitter. However, the intended way of receiving the messages on the ODID transmitter is from the drone it self. We have solved this by implementing filter inside the Dronetag DRI to receive the messages only from the flight controller. However, i think the messages should be addressed to the flight-controller only to prevent un-intentional forwarding. Maybe add it to RemoteID settings as checkbox whether for enabling of broadcasting.

image
 
image

qgcimages.qrc
```diff
@@ -149,6 +149,15 @@
         <file alias="PX4/BrandImage">src/FirmwarePlugin/PX4/PX4BrandImage.png</file>
         <file alias="Quad.svg">src/ui/toolbar/Images/Quad.svg</file>
         <file alias="RadioComponentIcon.png">src/AutoPilotPlugins/Common/Images/RadioComponentIcon.png</file>
+        <file alias="RID_EMERGENCY_BACKGROUND_HIGHLIGHT_SVG.svg">src/ui/toolbar/Images/RID_EMERGENCY_BACKGROUND_HIGHLIGHT_SVG.svg</file>
```
bkueng, line 152: Can you stick to the existing convention?
```suggestion
        <file alias="RidEmergencyBackgroundHighlight.svg">src/ui/toolbar/Images/RidEmergencyBackgroundHighlight.svg</file>
```

src/Settings/RemoteID.SettingsGroup.json
```diff
@@ -0,0 +1,169 @@
...
+{
+    "name":         "basicIDType",
+    "shortDesc":    "Basic ID Type",
+    "type":         "uint8",
+    "enumStrings":  "None, SerialNumber(ANSI/CTA-2063), CAA, UTM(RFC4122), Specific",
+    "enumValues":   "0,1,2,3,4",
+    "default":      2
+},
+{
+    "name":         "basicIDUaType",
```
bkueng, line 85: Isn't this coming from the vehicle?

src/Vehicle/RemoteIDManager.cc
```diff
@@ -0,0 +1,424 @@
...
+
+#include <QDebug>
+
+QGC_LOGGING_CATEGORY(RemoteIDManagerLog, "RemoteIDManagerLog")
+
+#define AREA_COUNT 1
+#define AREA_RADIUS 0
+#define SENDING_RATE_MSEC 1000
+#define ALLOWED_GPS_DELAY 5000
+#define RID_TIMEOUT 2000 // Messages should be arriving at 1 Hz, so we set a 2 second timeout
```
bkueng, line 25: Might be better to set it to 2.5, otherwise a single drop might or might not trigger a timeout
```suggestion
#define RID_TIMEOUT 2500 // Messages should be arriving at 1 Hz, so we set a 2 second timeout
```

src/Vehicle/RemoteIDManager.cc
```diff
@@ -0,0 +1,424 @@
...
+
+}
+
+void RemoteIDManager::_sendSelfIDMsg()
+{
+    WeakLinkInterfacePtr weakLink = _vehicle->vehicleLinkManager()->primaryLink();
+
+    if (!weakLink.expired()) {
+        mavlink_message_t       msg;
+        SharedLinkInterfacePtr  sharedLink = weakLink.lock();
```
bkueng, line 187: Please use this pattern to avoid race conditions:
```suggestion
    SharedLinkInterfacePtr sharedLink = weakLink.lock();
    if (sharedLink) {
        mavlink_message_t       msg;
```

src/Vehicle/RemoteIDManager.cc
```diff
@@ -0,0 +1,424 @@
...
+            }
+        } else {
+            _gcsGPSGood = false;
+            emit gcsGPSGoodChanged();
+            qCDebug(RemoteIDManagerLog) << "GCS GPS data is not valid.";
+        }
+        
+    }
+
+    if (!weakLink.expired()) {
```
bkueng, line 323: Same as above

src/Vehicle/RemoteIDManager.cc
```diff
@@ -0,0 +1,424 @@
...
+        mavlink_msg_open_drone_id_basic_id_pack_chan(_mavlink->getSystemId(),
+                                                    _mavlink->getComponentId(),
+                                                    sharedLink->mavlinkChannel(),
+                                                    &msg,
+                                                    _targetSystem,
+                                                    _targetComponent,
+                                                    _id_or_mac_unknown,
+                                                    _settings->basicIDType()->rawValue().toUInt(),
+                                                    _settings->basicIDUaType()->rawValue().toUInt(),
+                                                    reinterpret_cast<const unsigned char*>(ba.constData())),
```
bkueng, line 378: Mavlink does not handle this as string, so you need to make sure the buffer is large enough.
You can use `ba.resize(MAVLINK_MSG_OPEN_DRONE_ID_BASIC_ID_FIELD_UAS_ID_LEN);`

src/Vehicle/RemoteIDManager.cc
```diff
@@ -0,0 +1,424 @@
...
+        checkOperatorID();              // Check if OperatorID is good in case we want to send it from start because of the settings
+        emit commsGoodChanged();
+        qCDebug(RemoteIDManagerLog) << "Receiving ODID_ARM_STATUS from RID device";
+    }
+
+    // Restart the timeout
+    _odidTimeoutTimer.start();
+
+    // CompId and sysId are correct, we can proceed
+    mavlink_open_drone_id_arm_status_t armStatus;
```
bkueng, line 122: PX4 does not send this yet. Not blocking for this PR though.
I'm also not familiar what is required in detail for compliance.

Davidsastresas:
I am having some very busy days, but I am aware of this one. I will go over it during this weekend probably. Thank you very much for the report and the review, I will address all of it later this week. Thanks!

qgcimages.qrc
Davidsastresas: You are totally right, I am sorry. It should be fixed now

src/Settings/RemoteID.SettingsGroup.json
Davidsastresas: Not totally. This can be set on the parameters of remote id device itself, same as basic id message. If it is set in there, it will work perfectly fine, and in fact that seems to be the preferred way to do so.

But, it can also be set from GCS, and it works as follows:

- If none is set on device, and we set it on GCS, it will appear as basic id 1.
- If one is set on rid device, but we also set it in GCS, it will appear as basic id 2 as long as the type is not serial number.
- If one is set on rid device, but a new one is set in GCS with serial number type, it should override the one set in RID device for basic id 1.

src/Vehicle/RemoteIDManager.cc
Davidsastresas: I agree, this should be fixed now!

src/Vehicle/RemoteIDManager.cc
Davidsastresas: Thanks, it should be fixed now.

src/Vehicle/RemoteIDManager.cc
Davidsastresas: You are right, sorry I missed this one. I fixed it and left a comment about how this behaves ( adding padding bytes if array is smaller than size set, or discarding extra bytes if array is bigger than size set )

src/Vehicle/RemoteIDManager.cc
Davidsastresas: I am not an expert on this Remote ID thing, but according to what I learned during this implementation, talking with companies using it, RID device manufacturers, and Ardupilot team, it seems this message is what PX4 should send so the GCS understands the status of the vehicle regarding remote ID.

I am familiarized with Ardupilot but not PX4, so I can not help actively a lot on that from. But, when PX4 adds support for this I will be happy to help on the QGC side if needed. Also if anybody on PX4 team wants to ping me to clarify quickly any doubt they might have I am available anytime, probably the fastest way to reach me is by discord, I am in the dronecode servers, my nick is davidsastresas.

Thanks!

Davidsastresas:
@optical-o you are right, they are sent as broadcast right now. Please allow me a few days to discuss this with the rest of the players on this RID thing, because you are probably right and this is better being sent to the vehicle id and compid. If I confirm this to be fine, I will just modify it to be this way, so no need to use an option for this one. Thank you very much for reporting that!

@bkueng thank you very much for the review. I addressed all the points you mentioned, in additional commits for you to check it easier. But please, let me know before approving it, I will rebase this to master and squash the extra commits so the tree remains clean. Also I want to wait for the matter about broadcasting or not mentioned above.

I wanted to ask you about the link interface pattern you suggested. I checked and there is still a lot of parts of the code where it is done like the first approach I used here. I wasn't aware of that potential race condition. Should we update all the parts of QGC to use this new pattern? If so please let me know and I will work on it. Thanks!

src/Vehicle/RemoteIDManager.cc
bkueng: Ok thanks for clarifying. It should be easy to add this one.

bkueng:
This looks good now, nice work @Davidsastresas!

> I wanted to ask you about the link interface pattern you suggested. I checked and there is still a lot of parts of the code where it is done like the first approach I used here. I wasn't aware of that potential race condition. Should we update all the parts of QGC to use this new pattern? If so please let me know and I will work on it. Thanks!

Yes it should be changed everywhere.

mrpollo:
Hey @DonLakeFlyer can you give it one last sanity check please?

Davidsastresas:
Thanks @bkueng. I talked with @tridge and it seems the preferred way of sending messages is targeting the autopilot. I added this one ( thanks for reporting this @optical-o ) and tested here, and everything seems to be working just the same.

I squashed all the commits resulting from @bkueng review, so this is ready to merge whenever you see fit, maybe after that last check from @DonLakeFlyer .

Thanks!

optical-o:
@Davidsastresas  Today i have tested the latest commits on Ardupilot 4.3.5, which is currently the newest version. And found  out that the System messages is not being sent by the drone. It might not be an issue of this implementation, however i wanted to mention it here so you know.

The UI indicated that everything is ok. Except ARM status indication, which signaled that the Module does not have System message. How does the QGroundCenter verifies that everything is set correctly and was successfully transmitted to the drone ? Or does the UI just signal that local settings are fine ?

DonLakeFlyer:
On the mavlink version thing: I always just pulled HEAD when I updated the submdule.

As for the rest, all looks fantastic.

BluemarkInnovations:
I found a minor bug. 

Issue: if the GCS streams location data (system message) to the ArduRemoteID device and then you remove the GPS source, ArduRemoteID device won't detect it is gone.

ArduRemoteID checks the timestamp in the System message if it is non-zero and keeps changing. So, can you set this timestamp in the system message to zero once there is no GPS data anymore received by the GCS? To signal to ArduRemoteID that the system data is invalid.

Davidsastresas:
> @Davidsastresas Today i have tested the latest commits on Ardupilot 4.3.5, which is currently the newest version. And found out that the System messages is not being sent by the drone. It might not be an issue of this implementation, however i wanted to mention it here so you know.
> 
> The UI indicated that everything is ok. Except ARM status indication, which signaled that the Module does not have System message. How does the QGroundCenter verifies that everything is set correctly and was successfully transmitted to the drone ? Or does the UI just signal that local settings are fine ?

@optical-o  Thanks for reporting back. It goes as follows:

- GPS indicator: If set to FIXED, it will go green if valid lat/long is introduced. If it is set to LIVE, it will go green if valid GPS is received, and latest GCS GPS info is received less than 5 seconds ago. It will go red otherwise.
- Basic ID: will go green if the armstatus message received is "GOOD_TO_ARM". It will go red if armstatus message reports an error string "missing basic_id message". This is not ideal, but it is the only way given the circumstances of detecting precisely when it is just Basic ID failure. A bitmask or something in armstatus message will be more appropriate, but these are the cards we can play with.
- Arm status: will go green if armstatus message is good. If armstatus message contains errors it will go red.
- Comms indicator: will go green as soon as we receive an armstatus message. Will go red as soon as the timeout ( 2.5 seconds ) expires with no message received.
- Operator ID: will only check if the data set in QGC is valid, not empty and valid ID type.

I will get over the next comments in a moment.

Davidsastresas:
> 

Thanks for the review @DonLakeFlyer ! About mavlink, I tried to just pull mavlink HEAD, but some messages implemented in QGC changed, and it needed further rework. It should be quick to do, but I preferred to spend the time on releasing this. I have it noted down, and I will work on updating mavlink to current HEAD as soon as possible. Thanks!

Davidsastresas:
> I found a minor bug.
> 
> Issue: if the GCS streams location data (system message) to the ArduRemoteID device and then you remove the GPS source, ArduRemoteID device won't detect it is gone.
> 
> ArduRemoteID checks the timestamp in the System message if it is non-zero and keeps changing. So, can you set this timestamp in the system message to zero once there is no GPS data anymore received by the GCS? To signal to ArduRemoteID that the system data is invalid.

I spend some hours tracing this one. Indeed, arduremoteid doesn't get if we disconnect GPS. The android opendroneid app keeps reporting the latest valid system position, and the latest valid timestamp. I tested the timestamp to 0 you suggested, but it still does that. I think the issue is somewhere else, in Ardupilot or ArduremoteID.

Here in QGC we are sending the lat/lon as 0 when not valid, as expected by specs. I think we should look into this issue, but I don't think it is on the QGC side.

I suggest we merge this, and if this is addressed in the future in ArduRemoteID or Ardupilot and something changes, we can always come back to it.

Thanks!

DonLakeFlyer:
Looks fine to me from a quick look. Keep in min that I don't know the details of how this all works though. So just a sanity check from me.

Davidsastresas:
Thank you for spotting it @DonLakeFlyer , I missed it after the latest rebase, sorry. Should be fine with #10642

Amrope:
> @Davidsastresas在对 QGroundControl 的初步支持方面做得很好，我已经使用 Ardupilot 和 Dronetag DRI 测试了该功能。希望我的测试对您有所帮助。
> 
> 唯一要注意的是，我发现地面站正在以广播形式发送mavlink消息，我不确定这是不确定的。因为无人机将这些消息转发给ODID发送器。然而，在ODID发送器上接收消息的预期方式是从无人机本身接收消息。我们通过在 Dronetag DRI 内部实施过滤器来仅接收来自飞行控制器的消息解决了这个问题。但是，我认为这些消息应该发送给飞行控制器，以防止无意转发。也许将其添加到RemoteID设置中作为是否启用广播的前台。
> 
> 图像
> 
> 图像

I have Pixhawk6X flight board and DroneTagDRI ，build the Bootlarder is done.  ```
DID_ENABLE 1 @READONLY
DID_OPTIONS 1 @READONLY
DID_MAVPORT 2 @READONLY
DID_CANDRIVER 0 @READONLY
AHRS_EKF_TYPE 3 @READONLY
GPS_TYPE 1
GPS_TYPE2 0
``` but the Dronescanner is not working.
![8f690b66b7cc90c9a5d715737be46d8](https://github.com/mavlink/qgroundcontrol/assets/48429678/91a8bea9-dd97-4713-8adb-b8a07071d195)

PR #10603, opened 2023-02-25
Better user model / ui for critical vehicle messages

DonLakeFlyer:
Problems with current implementation:
* Users don't understand that these messages come directly from the vehicle
* The popup is massive and the X to close is even more massive! Example:
Screen Shot 2023-02-25 at 15 04 04
* Some time you may get multiples of these in a row. Close one, and you just get another. Depending on what the vehicle is sending.
* Current ui can show multiples at one time. Example:
Screen Shot 2023-02-25 at 15 04 15
* There is also a case where you can get a little mystery down array in the popup which means you should scroll because there are more for you to see.
* And then to make things even worse. If you dismiss one of these popups clearing that fact that you saw all critical messages. The Message Indicator in the toolbar would continue to display the Red yield sign indicator meaning there is something bad there you should look at.

The new thing which I think is a bit better and slightly less annoying:
* The popup only ever shows a single critical message and is smaller:
Screen Shot 2023-02-25 at 14 49 23
* If while this popup is being displayed you get additional critical messages it tells you that:
Screen Shot 2023-02-25 at 14 48 00
* And then if that is the case of additional messages you should look at, when you click to close the crticial popup the Message Indicator dropdown will drop showing you all the gory details of all the messages that just came in:
Screen Shot 2023-02-25 at 15 32 50

Also includes in this pull are fixes to MockLink status text sending support which is support to allow you to test this sort of stuff with a MockLink but was broken.

### DIFF
--- a/src/ui/MainRootWindow.qml
+++ b/src/ui/MainRootWindow.qml
@@ -503,15 +503,15 @@ ApplicationWindow {
     //-------------------------------------------------------------------------
     //-- Critical Vehicle Message Popup
 
-    property var    _vehicleMessageQueue:      []
-    property string _vehicleMessage:     ""
-
     function showCriticalVehicleMessage(message) {
         indicatorPopup.close()
         if (criticalVehicleMessagePopup.visible || QGroundControl.videoManager.fullScreen) {
-            _vehicleMessageQueue.push(message)
+            // We received additional wanring message while an older warning message was still displayed.
+            // When the user close the older one drop the message indicator tool so they can see the rest of them.
+            criticalVehicleMessagePopup.dropMessageIndicatorOnClose = true
         } else {
-            _vehicleMessage = message
+            criticalVehicleMessagePopup.criticalVehicleMessage      = message
+            criticalVehicleMessagePopup.dropMessageIndicatorOnClose = false
             criticalVehicleMessagePopup.open()
         }
     }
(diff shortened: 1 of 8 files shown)


mrpollo (changes requested):
Hey @DonLakeFlyer, while I agree this is an improvement over what we already have, would it make sense to standardize the dialog messages once and for all? I made a few mock-ups, I can share the styles if you are interested in following this pattern

QGC UI - attention
QGC UI - error
QGC UI - info
QGC UI - success

DonLakeFlyer:
These two things are different on purpose. There is a ton of confusion around where alerts come from. Making them different makes it a bit more understandable which are driven by the vehicle itself. And which are local from QGC.

Davidsastresas:
Sorry @DonLakeFlyer, I missed your mention here. I wasn't aware of this PR. It looks great to me, and I think we should merge this, because it indeed improves a lot the current behaviour.

What @mrpollo said about unifying the UI for these dialogs makes sense as well, I like it. But I do agree we can work on that after this PR, as what this addresses is in fact needed for the later work if we want to work on the unifying dialogs idea.

So, I suggest to merge this, and keep working on a separate thread about the dialogs. This week I will update #10617 with this matter, and also I will add some of the ideas we have been discussing lately.

booo:
As an incremental change I'm fine with merging this change. But I like the mock-ups much more. I suggest that we create a feature request with the mockups and hopefully someone finds the time to implement them. Shouldn't be to hard to swap the current stuff for the better versions from @mrpollo.

@mrpollo can you move your mockups into a feature request so they don't get lost. After that I'm happy to merge this one.

src/ui/MainRootWindow.qml
```diff
@@ -521,100 +521,85 @@ ApplicationWindow {
...
-                anchors.fill:   parent
-                onClicked: {
-                    criticalVehicleMessageFlick.flick(0,-500)
+        }
+
+        MouseArea {
+            anchors.fill: parent
+            onClicked: {
+                criticalVehicleMessagePopup.close()
+                console.log("Clicked", criticalVehicleMessagePopup.dropMessageIndicatorOnClose);
```
booo, line 598: Do we need the `console.log`? Where does the message pop up?

booo:
I will merge this next time I come around ;)

src/ui/MainRootWindow.qml
DonLakeFlyer: Nope. Left over debugging. I'll fix it.

s-lisovenko:
@DonLakeFlyer What do you think about moving this critical message popup into a separate component widget and removing this bulk of code from MainRoot?

booo:
I merged this for now. I suggest we create two tickets for the suggests made in the review. Are the reviewers up for that?

PR #10555, opened 2023-02-02
Mission item: CONDITION_YAW - added shortest direction

AndKe:
This adds the best option for Yaw changes in Copter  - the shortest way.
Tested, works fine.

PR #10605, opened 2023-02-27
Condition yaw options default

AndKe:
@booo - I made this a separate commit, as you requested in https://github.com/mavlink/qgroundcontrol/pull/10555
:)

### DIFF
--- a/src/MissionManager/MavCmdInfoCommon.json
+++ b/src/MissionManager/MavCmdInfoCommon.json
@@ -449,8 +449,8 @@
             },
             "param3": {
                 "label":            "Direction",
-                "enumStrings":      "Clockwise,Counter-Clockwise",
-                "enumValues":       "1,-1",
+                "enumStrings":      "Clockwise,Shortest,Counter-Clockwise",
+                "enumValues":       "1,0,-1",
                 "default":          1
             },
             "param4": {
--- a/src/MissionManager/MavCmdInfoCommon.json
+++ b/src/MissionManager/MavCmdInfoCommon.json
@@ -451,7 +451,7 @@
                 "label":            "Direction",
                 "enumStrings":      "Clockwise,Shortest,Counter-Clockwise",
                 "enumValues":       "1,0,-1",
-                "default":          1
+                "default":          0
             },
             "param4": {
                 "label":            "Offset",


patrickelectric:
@AndKe that is not on the mavlink spec from what I can see: https://mavlink.io/en/messages/common.html#MAV_CMD_CONDITION_YAW

AndKe:
@patrickelectric yes, you mean that -1 and 1 are described, and "0" is just implicit as default and means the shortest way (also implemented on AC)
also: you see the range be "-1...1"   - so there is a zero there, even if not explicitly mentioned.
all default yaw behavior is "shortest way" - so maybe it is not strange that "0" is to be understood as "no specific direction selected"

patrickelectric (changes requested):
The document currently says:

3: Direction | direction: -1: counter clockwise, 1: clockwise | min: -1 max:1 increment:2
-- | -- | --

So, if the increment is 2, there is no 0. As the direction says.


ArduPilot code for copter also points that 0 is not handled as shortest but as clockwise, we should follow the protocol spec over having assumptions. 

https://github.com/ardupilot/ardupilot/blob/master/ArduCopter/autoyaw.cpp#L116

AndKe:
@patrickelectric that's scary/horrible. 
Planning long missions with radar pointing to the side I've added the same CONDITION_YAW to 120° after each waypoint.
when I selected 1 (clockwise) , and the command was executed while the copter was at 121° ... - it did a 360° turn. 
(this was tested with current master in SITL)
 Then I tried "0" for direction - which worked fine - for many flights with AC4.1.3 (in real life)... was that luck or what?

We should absolutely have a "shortest way" turn.
If it is as you say then you tell me that is 120° is set, with right turn, (because one must choose one) - then if the actual heading is 120.01° the moment the next command is issued - one will expect a 359.99° turn to the right?

patrickelectric:
Yes, I agree, but you should send a PR to the mavlink standard and explain that.

AndKe:
@patrickelectric mavlink standard? where's that? - and how come using "0" actually worked perfectly fine for me?

patrickelectric:
> @patrickelectric mavlink standard? where's that?
The repository is here: https://github.com/mavlink/mavlink, you'll need to create a PR that changes: https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L1365

If that gets merged on the specification, we can merge your PR.

> and how come using "0" actually worked perfectly fine for me?

Not sure, but that's not a defined behavior, anything can happen.

AndKe:
@patrickelectric 
> If that gets merged on the specification, we can merge your PR.

What got merged is a rewrite of my PR, please see:
https://github.com/mavlink/mavlink/commit/02f1575d73b3b7eb09b1cc7ca3a03844aec05858

DonLakeFlyer:
So where is this thing at? Should this be merged?

AndKe:
Yes, this ("shortest") is a now formally a valid choice that should be supported :) 
please see https://mavlink.io/en/messages/common.html#MAV_CMD_CONDITION_YAW

DonLakeFlyer:
Thanks @AndKe

PR #10607, opened 2023-02-28
Modifications for PX4 PR-20495

farhangnaderi:
This PR presents the ability to use new feature in [20495](https://github.com/PX4/PX4-Autopilot/pull/20495) on QGC mission planning view. 

Here is how the change will be seen:

Screenshot from 2023-02-28 18-53-17

Tested this on QGC with the following log file:

https://review.px4.io/plot_app?log=aeea67e7-224f-4f6e-84cb-e753e4b3fc31

The .plan file is also [here](https://drive.google.com/file/d/1W43BJm7aXOsSOAIEzbX8_6nZCoyP1Hkz/view?usp=sharing)

A review will be appreciated so that it could be merged. 

cc: @afwilkin

### DIFF
--- a/src/MissionManager/MavCmdInfoCommon.json
+++ b/src/MissionManager/MavCmdInfoCommon.json
@@ -509,7 +509,7 @@
             "category":     "Flight control",
             "param1": {
                 "label":            "Type",
-                "enumStrings":      "Airspeed,Ground Speed,Ascend Speed, Descend Speed",
+                "enumStrings":      "Airspeed,Ground Speed,Ascend Speed,Descend Speed",
                 "enumValues":       "0,1,2,3",
                 "default":          0
             },
--- a/src/MissionManager/MavCmdInfoCommon.json
+++ b/src/MissionManager/MavCmdInfoCommon.json
@@ -509,8 +509,8 @@
             "category":     "Flight control",
             "param1": {
                 "label":            "Type",
-                "enumStrings":      "Airspeed,Ground Speed",
-                "enumValues":       "0,1",
+                "enumStrings":      "Airspeed,Ground Speed,Ascend Speed, Descend Speed",
+                "enumValues":       "0,1,2,3",
                 "default":          0
             },
             "param2": {
--- a/src/MissionManager/SpeedSection.cc
+++ b/src/MissionManager/SpeedSection.cc
@@ -92,8 +92,8 @@ void SpeedSection::appendSectionItems(QList<MissionItem*>& items, QObject* missi
         MissionItem* item = new MissionItem(seqNum++,
                                             MAV_CMD_DO_CHANGE_SPEED,
                                             MAV_FRAME_MISSION,
-                                            _masterController->controllerVehicle()->multiRotor() ? (_isascend ? 2 /* ascend speed */ : (_isdescend ? 3 /* descent speed */ : 1 /* ground speed */)) : 0 /* airspeed */,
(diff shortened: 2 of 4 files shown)


src/MissionManager/SpeedSection.cc
```diff
@@ -92,7 +92,7 @@ void SpeedSection::appendSectionItems(QList<MissionItem*>& items, QObject* missi
         MissionItem* item = new MissionItem(seqNum++,
                                             MAV_CMD_DO_CHANGE_SPEED,
                                             MAV_FRAME_MISSION,
-                                            _masterController->controllerVehicle()->multiRotor() ? 1 /* groundspeed */ : 0 /* airspeed */,    // Change airspeed or groundspeed
```
DonLakeFlyer, line 95: The SpeedSection code is only used by a Waypoint ui item which also changes flight speed. (And in turn outputs both a waypoint item and a change speed item from the one plan element) Waypoint items don't support the new ascend/descend speed changes nor do I think they should. So this changes is not needed.

src/Vehicle/VehicleFact.json
```diff
@@ -59,6 +59,20 @@
     "decimalPlaces":    1,
     "units":            "m/s"
 },
+{
```
DonLakeFlyer, line 62: VehicleFact.json is used by telemetry values which are exposed though the Vehicle object. This pull is for mission item changes which has no affect on Vehicle telemetry. So this change is incorrect and not needed.

src/MissionManager/SpeedSection.h
```diff
@@ -55,6 +55,8 @@ private slots:
     bool    _available;
     bool    _dirty;
     bool    _specifyFlightSpeed;
```
DonLakeFlyer, line 57: Same thing here. Incorrect and not needed.

src/MissionManager/MavCmdInfoCommon.json
```diff
@@ -509,8 +509,8 @@
             "category":     "Flight control",
             "param1": {
                 "label":            "Type",
-                "enumStrings":      "Airspeed,Ground Speed",
-                "enumValues":       "0,1",
+                "enumStrings":      "Airspeed,Ground Speed,Ascend Speed, Descend Speed",
```
DonLakeFlyer, line 512: There should not be spaces between the individual enum string values. Not the extra space after the comma. It might work fine, but would have to look at the code to determine. Better to remove.

DonLakeFlyer (changes requested):

src/MissionManager/SpeedSection.cc
```diff
@@ -93,7 +93,7 @@ void SpeedSection::appendSectionItems(QList<MissionItem*>& items, QObject* missi
                                             MAV_CMD_DO_CHANGE_SPEED,
                                             MAV_FRAME_MISSION,
                                             _masterController->controllerVehicle()->multiRotor() ? 1 /* groundspeed */ : 0 /* airspeed */,    // Change airspeed or groundspeed
-                                            _flightSpeedFact.rawValue().toDouble(),
```
DonLakeFlyer, line 96: There shouldn't be any changes at all in this file.

farhangnaderi: my bad! corrected!

patrickelectric (approved):

patrickelectric (changes requested):
Please squash it :)

PR #10615, opened 2023-03-09
add content_rating to metainfo and fix metadata_license

christianrauch:
This adds the tag `content_rating` to the `metainfo.xml` using the interactive form at https://odrs.gnome.org/oars.

The `metadata_license` states `GPL-3.0 AND Apache-2.0`. I assume this only applies to the project and not the metadata, i.e. this `metainfo.xml` itself. Usually, these metadata files use one of the "Creative Commons" licences. I propose to use the "CC0-1.0".

### DIFF
--- a/deploy/org.mavlink.qgroundcontrol.metainfo.xml
+++ b/deploy/org.mavlink.qgroundcontrol.metainfo.xml
@@ -1,34 +1,34 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop-application">
   <id>org.mavlink.qgroundcontrol</id>
-  
+
   <name>QGroundControl</name>
   <summary>UAS ground control station</summary>
   <developer_name>Dronecode Project, Inc.</developer_name>
-  
+
   <metadata_license>GPL-3.0 AND Apache-2.0</metadata_license>
   <project_license>GPL-3.0 AND Apache-2.0</project_license>
-  
+
   <recommends>
     <control>pointing</control>
     <control>keyboard</control>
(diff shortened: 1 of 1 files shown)


booo:
Cloud you please describe which metadata we are talking about? And maybe add a link to the metainfo.xml documentation?

I guess it is fine to merge but I will wait for the feedback first.

christianrauch:
> Cloud you please describe which metadata we are talking about?

Well, I mean the only file that is changed by this PR (`deploy/org.mavlink.qgroundcontrol.metainfo.xml`) of course.

> And maybe add a link to the metainfo.xml documentation?

The file is standardised: https://freedesktop.org/software/appstream/docs/chap-Quickstart.html

booo:
Thanks a lot for the clarification.

I personally think the change is fine. I'm new to licenses changes and a bit unsure how to deal with this in a proper way even if it only concerns one stupid metadata file. Do we have to ask the right holders?

christianrauch:
> Do we have to ask the right holders?

Someone has to agree to change the licence of this single xml metadata file. Not sure how this is handled internally, but I assume that whoever has the right to merge PRs can decide this. Looking at the [history of that file](https://github.com/mavlink/qgroundcontrol/commits/master/deploy/org.mavlink.qgroundcontrol.metainfo.xml) I can see that only @thopiekar committed to that file, starting in 2021.

booo:
Thanks again for the explanation!

I will merge this now. The potential right holder had two weeks to respond to the issue. I think that's enough.

Issue #10617, opened 2023-03-13
The next version of the QGC UI/UX


DonLakeFlyer:
So I'm planning on spending my time moving the QGC UI/UX forward to improve usability. First I'll detail out my opinions on high-level problems with the current QGC:

* In some sense QGC's strengths are also it's current weaknesses:
    * Includes support for as much mavlink spec as possible
    * Includes support for both PX4 Pro and Ardupilot firmware
    * Includes support for multiple specific vehicle types
    * Includes support for multi-os and multi-device (laptop, tablet, mouse, touch)
complexity

* The mavlink spec is missing a ton of stuff which relates to allowing a groundstation to understand a vehicles capabilities and/or hardware elements.
* The target user is somewhere between:
    *  Vehicle developer: Someone who is creating a new vehicle from scratch and needs all the initial setup, configuration and debugging options for testing out something new.
    *  Vehicle consumer: Think of as someone who has a DJI Mavic and uses DJI Go.
*  The target UI is somewhere between focusing on a laptop and a tablet.
*  The target UI is focused on screens which are a bit too large. For example tagetting a 10" screen leaves out Herelink sized displays.

All of the above bleeds through in various places in the UI adding complexity and looking a bit schizophrenic. Some examples:

* Since in general QGC doesn't know enough up front whether the vehicle has a camera or a gimbal:

    * When you are creating a mission you are presented with all sorts of camera and gimbal goop even though your vehicle may have none of those.
    * The Fly View shows a video widget even though you may have no video.
    * The Fly View starts on the map instead of video even though your vechicle DOES have video.
    * The camera widget on the fly view shows you video options, when your vechicle only has picture capture capability.

* Since QGC doesn't know if your vehicle has the sensors needed to perform precision landing it shows you the Precision Land flight mode in the dropdown even though it's not possible. Further cluttering the already massive list of flight modes.

* Initial vehicle setup/configuration is mixed in with every day use settings
    * If you go into Vehicle Setup the majority of stuff in there if for configuring a brand new vehicle from scratch. 
    * For a vehicle consumer this is something you would never need. The vehicle is already set up for you.
    * Even for a DIY vehicle developer it is something you do once and then you go fly. You might re-calibrate the compass every once in a while. But the other stuff you don't go back to.
    * And then in the middle of a huge list of one time vehicle developer setup stuff is a nugget of every daty use settings: The Safety page, which contains things like your RTL height.

[...]

PR #10618, opened 2023-03-13
The next version of the QGC UI/UX

DonLakeFlyer:
DO NOT MERGE

This is a pull which goes along with #10617. It's here to generates artifacts folks can you to try out the WIP changes.


booo:
Maybe we could enable the CI on all branches and remove these "please don't merge" PRs? They feel a little bit like additional noise the other developers have to deal with in the limited time they have.

DonLakeFlyer:
I'd assume by now anyone you wanted to try it out has. So this can be closed.

PR #10634, opened 2023-03-23
4.2: Update sdl libraries to work with latest xbox bluetooth controller

patrickelectric:


patrickelectric:
> I don't think we should push new features to the old release. I would rather push for a new major release and add such changes there.
> 
> Can we add the libraries as submodules?

It's not a new feature, is more a bug fix, to support the latest joysticks. 

> Can we add the libraries as submodules?

I was unable to find one, if one exist, it should be applied over master first, since this a major change that may affect the stability of the current stable release.

booo:
> > I don't think we should push new features to the old release. I would rather push for a new major release and add such changes there.
> > Can we add the libraries as submodules?
> 
> It's not a new feature, is more a bug fix, to support the latest joysticks.

I still think new device support is an added feature and not a bug fix. But if you like feel free to merge this. I have no strong opinion about it.

booo:
Is this change in master present?

patrickelectric:
Yes, it's a cherry-pick from master.
This is not a feature, is a bug fix for 4.2, but it appears to not take effect. I should get a new xbox controller to validate that next month.

patrickelectric:
Validated by multiple users.

PR #10635, opened 2023-03-23
fix: check VIDEO_STREAM_STATUS messages whether stream is running

RedFox20:
I discovered this bug when trying to add integration support for a multi-camera mavlink device with multiple streams.

The current code assumed that each Camera would always be streaming, but this is not always the case, and not guaranteed by MAVLink camera protocol. Especially with 3+ cameras, due to data rate limitations only 1 out 3 might be active.

QGC would be stuck assuming that the streams are running, but the camera system could be expecting for a VIDEO_START_STREAMING command for those particular streams.

The VIDEO_STREAM_STATUS `flags` field describes whether the Stream is actually running, and updating the `_currentStream` value according to those flags re-enables video stream switching.

I also slightly improved the camera debug logging to make it easier to debug multi-camera system issues.

### DIFF
--- a/src/Camera/QGCCameraControl.cc
+++ b/src/Camera/QGCCameraControl.cc
@@ -1624,8 +1624,9 @@ QGCCameraControl::handleVideoStatus(const mavlink_video_stream_status_t* vs)
 void
 QGCCameraControl::setCurrentStream(int stream)
 {
-    if (stream != _currentStream && 0 <= stream && stream < _streamLabels.count()) {
-        if(QGCVideoStreamInfo* pInfo = currentStreamInstance()) {
+    if (stream != _currentStream && (stream == -1 || (0 <= stream && stream < _streamLabels.count()))) {
+        // stop the old stream if possible
+        if (QGCVideoStreamInfo* pInfo = currentStreamInstance()) {
             qCDebug(CameraControlLog) << "Stopping stream: " << modelName() << " uri=" << pInfo->uri();
             //-- Stop current stream
             _vehicle->sendMavCommand(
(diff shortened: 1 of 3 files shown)


src/Camera/QGCCameraControl.cc
```diff
@@ -1600,45 +1601,53 @@ void
...
-                    _compID,                                // Target component
-                    MAV_CMD_VIDEO_START_STREAMING,          // Command id
-                    false,                                  // ShowError
-                    pInfo->streamID());                     // Stream ID
-                //-- Update stream status
-                _requestStreamStatus(static_cast<uint8_t>(pInfo->streamID()));
-            }
-            emit currentStreamChanged();
-            emit _vehicle->cameraManager()->streamChanged();
+    if (stream != _currentStream && 0 <= stream && stream < _streamLabels.count()) {
```
patrickelectric: you can just revert the logic, avoid the indentation level completely.

src/Camera/QGCCameraControl.cc
```diff
@@ -1600,45 +1601,53 @@ void
...
+        if (pInfo->isRunning()) { // this is the current active stream
+            _currentStream = _streams.indexOf(pInfo);
+        } else { // we don't know which stream is active, check all of them
+            _currentStream = -1;
+            for (int i = 0; i < _streams.count(); i++) {
+                if (auto pStream = qobject_cast<QGCVideoStreamInfo*>(_streams[i]); pStream->isRunning()) {
+                    _currentStream = i;
+                    break;
+                }
+            }
```
patrickelectric, line 1616: you can use `std::find_if` and `std::distance`

patrickelectric (changes requested):

booo:
@RedFox20 can you update the PR with the suggested changes?

booo:
@RedFox20 please reopen if you find the time to integrate the suggested changes. We would love to integrate your updated PR.

PR #10636, opened 2023-03-24
Keyboard slider switch

DonLakeFlyer:
Screen Shot 2023-03-24 at 11 30 00

SliderSwitch is used by the Guided Action confirmation ui. With touch it's easy to slide the slider with your finger. But if you are running QGC on a laptop with a trackpad, sliding that thing over with your fingers is a royal PITA.

Updates the SliderSwitch control:
* Support holding down the spacebar for confirmation. 
* The length of time you have to hold the spacebar is fairly long at 1.5 seconds. The reason for this is you want to give the user enough time to bail out if this isn't really what they wanted. Otherwise the slider swishes across the screen too fast to give you time to release to need to release the space bar.
* I also updated and centered the slider prompt text


DonLakeFlyer:
Ahhh! I hate git.

PR #10643, opened 2023-03-27
Enable ROIModeCapability for ArduPilot vehicles

sypaq-nexton:
Addresses #10640.


booo:
Please add a comment to the commit message about the fact that some builds don't support ROI features. And/or link the issue with the discussion.

sypaq-nexton:
> Please add a comment to the commit message about the fact that some builds don't support ROI features. And/or link the issue with the discussion.

Done.

booo:
Thanks a lot. 

Do you know if the `MAV_RESULT_UNSUPPORTED` pops up somewhere so the user know about the missing support?

sypaq-nexton:
> Do you know if the `MAV_RESULT_UNSUPPORTED` pops up somewhere so the user know about the missing support?

Yeah, there's a popup that appears on the Fly View.

PR #10645, opened 2023-03-28
Ardupilotmega flag added for Mavlink Dialect

harshmittal2210:
While building for Mavlink common dialect `common`, `mavlink_rangefinder_t` and all supporting functions are undefined as it is defined for only `ardupilotmega` dialect.

Have added flags to check the same. This I faced when I tried building for the custom QGC build with default dialect as `common`

### DIFF
--- a/QGCExternalLibs.pri
+++ b/QGCExternalLibs.pri
@@ -58,6 +58,11 @@ isEmpty(MAVLINK_CONF) {
     }
 }
 
+equals (MAVLINK_CONF, ardupilotmega) {
+    MAVLINK_CONF_ARDUPILOTMEGA = 1
+    DEFINES += MAVLINK_CONF_ARDUPILOTMEGA=$$MAVLINK_CONF_ARDUPILOTMEGA
+}
+
 # If defined, all APM specific MAVLink messages are disabled
 contains (CONFIG, QGC_DISABLE_APM_MAVLINK) {
     message("Disable APM MAVLink support")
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
You should define you custom build as being PX4 only and this define should be there.

harshmittal2210:
@DonLakeFlyer I could not find any `#ifdef` specific for this in `vehicle.cc`, I am not sure which one are you talking about.


> You should define your custom build as being PX4 only and this define should be there.

Here are you referring that the flags that I have mentioned for the mavlink dialect should actually be defined as part of PX4 or Arducopter builds only? 

Also not just for the custom builds but for normal builds also if the dialect is changed then it was showing a compilation error.

DonLakeFlyer:
> I could not find any `#ifdef` specific for this in `vehicle.cc`

https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L791

booo:
As far as I understand an already existing variable should be used in the `#ifdef`. @harshmittal2210 can you make that change and push it to the PR here?

Does it make sense to add a custom build to the CI system? Someone wants to contribute that?

harshmittal2210:
Hi @booo 
Sure I'll make that change and push it. 

I am not sure about the CI system as I have not used it. But yes if someone can guide me I'll try to add that as well.

harshmittal2210:
Hi @DonLakeFlyer @booo 

I have made the changes, is this fine or should I make any more changes?

DonLakeFlyer:
Looks good now

booo:
Thanks for your contribution!

PR #10646, opened 2023-03-28
RemoteIDManager: Fix GPS bad not being detected:

Davidsastresas:
we were using the geopositioninfo.isValid() in order to discard GPS and send lat/long/alt 0, but this function returns true even if latest GPS refresh was long time ago, so now instead we are using _gcsGPSGood, which takes into account the 5 seconds timeout as well.

This fixes the issue where unplugging GPS would still show "gps good" because QGC backend would take just the latest good GPS position, so as long as we had good GPS position at some point this was always true. Now after the 5 seconds timeout, if we don't have further "fresher" GPS updates, this flag will evaluate to false, and thus it will work correctly.


Davidsastresas:
Thanks! I am merging it!

Issue #10544, opened 2023-01-16
Multi-Vehicle: Joystick input gets stuck on a single vehicle


goasChris:
## Expected Behavior
When having multiple devices/vehicles connected to QGC I would expect that you can easily and in a stable manner switch between them when sending joystick/controller inputs.

## Current Behavior
When having two vehicles connected to QGC, it will get stuck sending joystick inputs to a single vehicle after switching from one vehicle to another a few times (1-3). You will need to restart QGC, or recalibrate the controller in order for be able to switch again, but it will get stuck after switching.

## Steps to Reproduce:
1. Have two vehicles set up. We used the docker version of ardusub companion. Set them up to point at QGC IP with port udp/14660. We used vehicle ID 10 and 13 for ours.
2. In QGC, set up a "comm link" using UDP/14660, and your local ipv4 adresse.
3. Confirm in the log that both have connected, and that you get the multi vehicle UI up, so that you can switch between them. Confirm you can arm/disarm both, and that both are delivering mavdata.
4. Have a USB controller set up, calibrated and configured.
5. Move the stick forward, so that you'll send thruster input to the currently selected vehicle (currently 10). Physically confirm the thrusters are engaging. Or you can send a simple arm/disarm command via the controller. 
6. Let go of the stick/button, wait a few sec, and then switch to the other vehicle (now 13) in the drop down menu at the top. Repeat step 5.
7. Repeat step 6 (switch vehicle, up to 2-3 times), and observe how the thruster will not engage on the selected vehicle, nor will arm/disarm via controller input. Instead it is sending the packets to the wrong vehicle for thruster, arm/dismarm etc.
8. Try switching back and forth between the vehicle, and observe it being completely stuck on a single vehicle.

## System Information
- Operating System: Win 10 22H2
- QGC Version: 4.2.3
- QGC build: Stable, from Github
- Flight Controller: Blue Robotics Navigator on pi4
- Autopilot (with version): unsure, rover type

## Detailed Description
When controlling multiple robots (id Vehicle 10 and 13). We get info from both.
Thru the QGC UI we can arm and disarm both. Switching is fine, we get up to date information.

The issue is with USB controller input getting sent to the incorrect mavlink/robot at times.
Usually fine initially, but after switching between boths vehicles 1-3 times, it will get stuck sending USB controller/thrusts to a single vehicle.

[...]

PR #10648, opened 2023-03-29
Joystick input should switch to the active vehicle when the active vehicle changes

vorobotics:
This addresses issue #10544. 

When there are multiple vehicles, and you select a vehicle, the joystick should switch over to the newly active vehicle, and _only_ the active vehicle (so that joystick inputs are not sent to a vehicle you do not intend to control). 

In this PR, we achieve this by handling the `MultiVehicleManager::activeVehicleAvailableChanged` and `MultiVehicleManager::activeVehicleChanged` signals and invoking `setJoystickEnabled()` appropriately.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -178,7 +178,8 @@ Vehicle::Vehicle(LinkInterface*             link,
     _linkManager = _toolbox->linkManager();
 
     connect(_joystickManager, &JoystickManager::activeJoystickChanged, this, &Vehicle::_loadSettings);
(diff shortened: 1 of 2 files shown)


booo:
Thanks for your contribution. Very good catch!

Could your add your description of the problem to the commit message? A safety related commit should not be a oneliner.

Maybe it is worth backporting this to a bug fix release of 4.2.X?

vorobotics:
> Could your add your description of the problem to the commit message? A safety related commit should not be a oneliner.

I have updated the commit message, please check if it is suitable.

PR #10649, opened 2023-03-30
Fix cmake build

s-lisovenko:
-Increase C++ standard to C++17 as for qmake build
-Add missing files for remote id
-Add missing files for custom action

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ include(GNUInstallDirs)
 
 set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug;Release;RelWithDebInfo;MinSizeRel;Coverage")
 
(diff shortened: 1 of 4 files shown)


sypaq-nexton:
Thanks. I didn't realise I broke the CMake build when I added the Custom Actions...

booo:
Could we split this up in multiple commmits?

What are the implications of the C++ standard change?

s-lisovenko:
> Could we split this up in multiple commmits?
> 
> What are the implications of the C++ standard change?

I split the commit into two separate commits.

About С++17 еhere are 2 reasons for increasing the standard
1. qmake build uses c++17
2. Compilation errors
For example for QGCCachedFileDownload.cc file deduction type error
Screenshot from 2023-04-06 07-37-06

booo:
Thanks a lot for your contribution!

PR #10653, opened 2023-04-06
add release tag to metainfo file

christianrauch:
This adds a `release` tag to the `.metainfo.xml`.

According to the [documentation](https://freedesktop.org/software/appstream/docs/chap-Quickstart.html#qsr-app-contents-suggestions), this is only suggested. But it is required for Flathub and would allow using the upstream metainfo file without modifications.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,8 +224,13 @@ install(
     DESTINATION ${CMAKE_INSTALL_DATADIR}/icons/hicolor/128x128/apps/
     RENAME org.mavlink.qgroundcontrol.png
 )
+configure_file(
+    ${CMAKE_SOURCE_DIR}/deploy/org.mavlink.qgroundcontrol.metainfo.xml.in
+    ${CMAKE_BINARY_DIR}/metainfo/org.mavlink.qgroundcontrol.metainfo.xml
+    @ONLY
+)
 install(
-    FILES ${CMAKE_SOURCE_DIR}/deploy/org.mavlink.qgroundcontrol.metainfo.xml
+    FILES ${CMAKE_BINARY_DIR}/metainfo/org.mavlink.qgroundcontrol.metainfo.xml
     DESTINATION ${CMAKE_INSTALL_DATADIR}/metainfo/
 )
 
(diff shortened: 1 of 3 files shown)


deploy/org.mavlink.qgroundcontrol.metainfo.xml
```diff
@@ -24,6 +24,10 @@
 
   <content_rating type="oars-1.1" />
 
+  <releases>
+    <release date="2023-03-21" version="4.2.6"/>
```
patrickelectric, line 28: this tag and date should be generated by the build system over using it manually

patrickelectric (changes requested):

deploy/org.mavlink.qgroundcontrol.metainfo.xml
christianrauch: Is there a reliable way to extract the version? I don't see it in the `CMakeLists.txt`. `git` is not an option as someone may build the source from an archive.

patrickelectric: > Is there a reliable way to extract the version? I don't see it in the `CMakeLists.txt`. `git` is not an option as someone may build the source from an archive.

It's not possible to build from archive since we use submodules already.

christianrauch: Can't you distribute the main source and the source of the dependencies as an archive? I.e. checkout the repo with submodules, delete the `.git` folder and then zip all the files.

How would I get the version in this case? How is it set for the compiled executable?

patrickelectric: GitHub archives in releases does not contain submodules.
Deleting `.git` is not an use case, the file will be updated by the build system on CI with release tags.

christianrauch: Can you give me a pointer where the release version is extracted during CI?

patrickelectric:
https://github.com/mavlink/qgroundcontrol/blob/master/CMakeLists.txt#L48

https://github.com/mavlink/qgroundcontrol/blob/master/QGCCommon.pri#L165

christianrauch:
The automatic release tag creation passes the Flathub CI: https://buildbot.flathub.org/#/builders/33/builds/4165. Thanks for the feedback.

booo:
Is this ready to go?

christianrauch:
> Is this ready to go?

From my side yes.

PR #10660, opened 2023-04-14
libs: OpenSSL update to 1.1.1l_1.0.2u

Davidsastresas:
OpenSSL for android wasn't updated for 3 years. Fetching elevation data in Android wasn't working, it reported a TLS error. Updating to latest openSSL solves the problem.

By the way, openssl for android submodule is pointing to a branch in Auterion repo. That Auterion repo doesn't have any modification at all in this front. Would it make sense to make it point to the original openssl repo Auterion repo is pointing to?

Thank you very much to Arace UAS for reporting this issue!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/mavlink/c_library_v2.git
 [submodule "libs/OpenSSL/android_openssl"]
 	path = libs/OpenSSL/android_openssl
-	url = https://github.com/Auterion/android_openssl
+	url = https://github.com/KDAB/android_openssl
 [submodule "libs/qmlglsink/gst-plugins-good"]
 	path = libs/qmlglsink/gst-plugins-good
 	url = https://github.com/mavlink/gst-plugins-good.git


booo:
> OpenSSL for android wasn't updated for 3 years. Fetching elevation data in Android wasn't working, it reported a TLS error. Updating to latest openSSL solves the problem.

Please add such information to the commit message so we know the reason for the change. Do you have a description of the problem?

Pointing to an official release of the library is a good idea.

Davidsastresas:
> > OpenSSL for android wasn't updated for 3 years. Fetching elevation data in Android wasn't working, it reported a TLS error. Updating to latest openSSL solves the problem.
> 
> Please add such information to the commit message so we know the reason for the change. Do you have a description of the problem?
> 
> Pointing to an official release of the library is a good idea.

Thanks for the review. I extended the commit message, and added a new commit changing the remote directly to github.com/KDAB/android_openssl instead of the Auterion fork. Let me know if it looks good to merge for you. Thanks!

Davidsastresas:
I know. I did a quick search and didn't found nothing really solid around it. This was the repo Auterion ssl was pointing to. I've seen openssl itself has a github repo, but it is not android specific nor Qt. We could delay this PR and try to find out more. Although the PR is actually fixing a big issue in Android for QGC. Not sure how to proceed. What do you think?

booo:
It seems like the best solution for now given our limited time.

Maybe you can create a small ticket that reminds to fix this?

Building from source is not an option? Sorry, I'm totally unfamiliar with the situation and don't have time to dig in.

Davidsastresas:
I agree. Regarding fixing this, I am not sure I would really spend more time in the short term on it. I mean, that is the ssl library we have been using for years, and they are still updating it. I checked quickly and they even updated recently a new branch 3.something to support newer Qt 6.5. So unless pointing to that repo is really a problem I wouldn't really spend more time on it. Do you agree?

booo:
I totally agree.

PR #10661, opened 2023-04-14
Fixed ModalAi Flight Core V2 Firmware Update Not Working On QGC

jnomikos:
image

Updating the firmware on the FlightCore V2 would not work on QGC due to the board not being included on USBBoardInfo.json. After I put the board on USBBoardInfo.json, I could update the firmware like normal. Without its inclusion, commands to the board would time out during update.


Davidsastresas:
Why is it marked as draft? It looks good to me, unless you have any reason to not merge I think it is good to go!

jnomikos:
> Why is it marked as draft? It looks good to me, unless you have any reason to not merge I think it is good to go!

Wanted to make absolute sure that this fixed the issue through more testing, but our fightcore V2 was at the field. Did more testing and it looks good to me.

PR #10667, opened 2023-04-24
[WIP] Improve gimbal support

julianoes:
This work mostly done by @Davidsastresas aims to bring gimbal functionality as much as possible based on the gimbal v2 protocol.

- [x] Support acquire/release gimbal control
- [x] Support multigimbal
- [x] Joystick buttons control: supporting up/down/left/right and center
- [x] Gimbal control acquisition for AP
- [x] ROI to home in AP
- [x] Retract in AP
- [x] Gimbal control acquisition for PX4
- [ ] ROI to home in PX4
- [ ] Retract in PX4
- [ ] ...

FYI @olliw42, @rmackay9

### DIFF
--- a/src/FlightDisplay/FlyViewVideoToolStrip.qml
+++ b/src/FlightDisplay/FlyViewVideoToolStrip.qml
@@ -194,7 +194,7 @@ Item {
                         iconSource:         "/HA_Icons/CAMERA_90.png"
                         onTriggered: {
                             if (_activeVehicle) {
-                                _activeVehicle.gimbalController.sendGimbalManagerPitchYaw(-90, 0) // point gimbal down
+                                _activeVehicle.gimbalController.sendPitchBodyYaw(-90, 0) // point gimbal down
                             }
                         }
                     },
--- a/src/FlightDisplay/FlyViewVideoWidgetLayer.qml
+++ b/src/FlightDisplay/FlyViewVideoWidgetLayer.qml
@@ -16,12 +16,12 @@ Item {
     property var screenXrateInitCoocked
     property var screenYrateInitCoocked
 
-    property var  _gimbalController:         _activeVehicle ? _activeVehicle.gimbalController : undefined  
+    property var  _gimbalController:         _activeVehicle ? _activeVehicle.gimbalController : undefined
     property var  _activeGimbal:             _gimbalController ? _gimbalController.activeGimbal : undefined
     property bool _gimbalAvailable:          _activeGimbal != undefined
     property var  _gimbalControllerSettings: QGroundControl.settingsManager.gimbalControllerSettings
     property var  _activeVehicle:            QGroundControl.multiVehicleManager.activeVehicle
-    
+
     property bool shouldProcessClicks:   QGroundControl.videoManager.fullScreen || flyViewVideoWidgetLayer._gimbalControllerSettings.EnableOnScreenControl.value ||
                                          _activeVehicle
 
@@ -70,7 +70,7 @@ Item {
         }
         sendRateTimer.stop()
         screenXrateInitCoocked = null
-        screenYrateInitCoocked = null 
+        screenYrateInitCoocked = null
         // _activeVehicle.gimbalController.gimbalOnScreenControl(0, 0, false, true, true)
     }
 
@@ -94,7 +94,7 @@ Item {
     // Pitch/Roll indicators
     Rectangle {
         id: pitchRollRectangle
-        width: gimbalPitchLabel.width + gimbalPanLabel.width + ScreenTools.defaultFontPixelWidth * 5
+        width: gimbalPitchLabel.width + gimbalPanLabel.width + ScreenTools.defaultFontPixelWidth * 20
         height: gimbalPitchLabel.height + ScreenTools.defaultFontPixelWidth * 2
         anchors.top: parent.top
         anchors.horizontalCenter: parent.horizontalCenter
@@ -103,7 +103,7 @@ Item {
 
         QGCLabel {
             id: gimbalPitchLabel
-            text: rootItem._gimbalAvailable ? "Tilt: " + rootItem._gimbalController.activeGimbal.curPitch.toFixed(2) : ""
+            text: rootItem._gimbalAvailable ? "Tilt: " + rootItem._gimbalController.activeGimbal.absolutePitch.toFixed(1) : ""
             visible: rootItem._gimbalAvailable
             anchors.top: parent.top
             anchors.left: parent.horizontalCenter
@@ -112,7 +112,7 @@ Item {
 
         QGCLabel {
             id: gimbalPanLabel
-            text: rootItem._gimbalAvailable ? "Pan: " + rootItem._gimbalController.activeGimbal.curYaw.toFixed(2) : ""
+            text: rootItem._gimbalAvailable ? "Pan: " + rootItem._gimbalController.activeGimbal.bodyYaw.toFixed(1) + ", from North: " + rootItem._gimbalController.activeGimbal.absoluteYaw.toFixed(1): ""
             visible: rootItem._gimbalAvailable
             anchors.top: parent.top
             anchors.right: parent.horizontalCenter
--- a/src/FlightDisplay/FlyViewWidgetLayer.qml
+++ b/src/FlightDisplay/FlyViewWidgetLayer.qml
@@ -288,7 +288,7 @@ Item {
     GripperMenu {
         id: gripperOptions
     }
-    
+
     // need to manage full screen here
     FlyViewVideoToolStrip {
         id:                     videoToolStrip
@@ -327,20 +327,20 @@ Item {
         // Pitch/Roll indicators
     Rectangle {
         id: pitchRollRectangle
-        width: gimbalPitchLabel.width + gimbalPanLabel.width + ScreenTools.defaultFontPixelWidth * 5
+        width: gimbalPitchLabel.width + gimbalPanLabel.width + ScreenTools.defaultFontPixelWidth * 20
         height: gimbalPitchLabel.height + ScreenTools.defaultFontPixelWidth * 2
         anchors.top: parent.top
         anchors.horizontalCenter: parent.horizontalCenter
         color: qgcPal.window
         opacity: 0.8
 
-        property var _gimbalController: _activeVehicle ? _activeVehicle.gimbalController : undefined  
+        property var _gimbalController: _activeVehicle ? _activeVehicle.gimbalController : undefined
         property var _activeGimbal:     _gimbalController ? _gimbalController.activeGimbal : undefined
         property bool _gimbalAvailable: _activeGimbal != undefined
 
         QGCLabel {
             id: gimbalPitchLabel
-            text: pitchRollRectangle._gimbalAvailable ? "Tilt: " + pitchRollRectangle._activeGimbal.curPitch.toFixed(2) : ""
+            text: pitchRollRectangle._gimbalAvailable ? "Tilt: " + pitchRollRectangle._activeGimbal.absolutePitch.toFixed(1) : ""
             visible: pitchRollRectangle._gimbalAvailable
             anchors.top: parent.top
             anchors.left: parent.horizontalCenter
@@ -349,7 +349,7 @@ Item {
 
         QGCLabel {
             id: gimbalPanLabel
-            text: pitchRollRectangle._gimbalAvailable ? "Pan: " + pitchRollRectangle._activeGimbal.curYaw.toFixed(2) : ""
+            text: pitchRollRectangle._gimbalAvailable ? "Pan: " + pitchRollRectangle._activeGimbal.bodyYaw.toFixed(1) + ", from North: " + pitchRollRectangle._activeGimbal.absoluteYaw.toFixed(1) : ""
             visible: pitchRollRectangle._gimbalAvailable
             anchors.top: parent.top
             anchors.right: parent.horizontalCenter
--- a/src/FlightMap/MapItems/VehicleMapItem.qml
+++ b/src/FlightMap/MapItems/VehicleMapItem.qml
@@ -49,7 +49,7 @@ MapQuickItem {
             anchors.centerIn:             vehicleItem
             width:                        vehicleItem.width * 2
             height:                       vehicleItem.height * 2
-            property var gimbalYaw:       vehicle && vehicle.gimbalController.activeGimbal ? vehicle.gimbalController.activeGimbal.curYaw : NaN
+            property var gimbalYaw:       vehicle && vehicle.gimbalController.activeGimbal ? vehicle.gimbalController.activeGimbal.bodyYaw : NaN
             rotation:                     gimbalYaw + 180
             onGimbalYawChanged:           canvas.requestPaint()
             visible:                      vehicle && !isNaN(gimbalYaw)
--- a/src/Gimbal/GimbalController.cc
+++ b/src/Gimbal/GimbalController.cc
@@ -12,7 +12,7 @@ QGC_LOGGING_CATEGORY(GimbalLog,         "GimbalLog")
 Gimbal::Gimbal()
     : QObject(nullptr)
 {
-    
+
 }
 
 Gimbal::Gimbal(const Gimbal& other)
@@ -23,24 +23,25 @@ Gimbal::Gimbal(const Gimbal& other)
 
 const Gimbal& Gimbal::operator=(const Gimbal& other)
 {
-    requestInformationRetries = other.requestInformationRetries;                    
-    requestStatusRetries      = other.requestStatusRetries;                
-    requestAttitudeRetries    = other.requestAttitudeRetries;                
-    receivedInformation       = other.receivedInformation;                
-    receivedStatus            = other.receivedStatus;        
-    receivedAttitude          = other.receivedAttitude;            
-    isComplete                = other.isComplete;    
-    retracted                 = other.retracted;    
-    neutral                   = other.neutral;    
-    yawLock                   = other.yawLock;    
-    _haveControl              = other._haveControl;        
-    _othersHaveControl        = other._othersHaveControl;            
-    _curRoll                  = other._curRoll;    
-    _curPitch                 = other._curPitch;    
-    _curYaw                   = other._curYaw;  
+    requestInformationRetries = other.requestInformationRetries;
+    requestStatusRetries      = other.requestStatusRetries;
+    requestAttitudeRetries    = other.requestAttitudeRetries;
+    receivedInformation       = other.receivedInformation;
+    receivedStatus            = other.receivedStatus;
+    receivedAttitude          = other.receivedAttitude;
+    isComplete                = other.isComplete;
+    retracted                 = other.retracted;
+    neutral                   = other.neutral;
+    _yawLock                   = other._yawLock;
     _haveControl              = other._haveControl;
-    _othersHaveControl        = other._othersHaveControl;  
-    _deviceId                 = other._deviceId;    
+    _othersHaveControl        = other._othersHaveControl;
+    _absoluteRoll          = other._absoluteRoll;
+    _absolutePitch         = other._absolutePitch;
+    _bodyYaw               = other._bodyYaw;
+    _absoluteYaw           = other._absoluteYaw;
+    _haveControl              = other._haveControl;
+    _othersHaveControl        = other._othersHaveControl;
+    _deviceId                 = other._deviceId;
 
     return *this;
 }
@@ -54,7 +55,7 @@ GimbalController::GimbalController(MAVLinkProtocol* mavlink, Vehicle* vehicle)
     connect(_vehicle, &Vehicle::mavlinkMessageReceived, this, &GimbalController::_mavlinkMessageReceived);
 }
 
-void 
+void
 GimbalController::setActiveGimbal(Gimbal* gimbal)
 {
     if (!gimbal) {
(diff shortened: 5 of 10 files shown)


rmackay9:
@julianoes,

Really great to see this. For the gimbal control aquisition issue for AP, it's the [GIMBAL_MANAGER_INFORMATION](https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_INFORMATION) support that is needed?

kadabr:
Would it be reasonable to have a separate class like GimbalManager for related functionality to not overload the Vehicle class?

Davidsastresas:
> Would it be reasonable to have a separate class like GimbalManager for related functionality to not overload the Vehicle class?

That is one of the first tasks I would like to perform before undrafting this. Thanks!

julianoes:
> Would it be reasonable to have a separate class like GimbalManager for related functionality to not overload the Vehicle class?

Yes, especially because that will later allow us to have multiple gimbals, something that the protocol should support.

julianoes:
@rmackay9 for control it's actually [https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_STATUS](GIMBAL_MANAGER_STATUS). I want to request that topic at a low rate in a next commit.

julianoes:
@olliw42 thanks for the review. There might be a bit of a misunderstanding. This PR is intended to be, eventually, when it's done to be as much as possible based on the gimbal v2 protocol. In the current state, it is definitely not yet, it's a draft PR for @Davidsastresas, me, and whoever else has input to collaborate.

I would have thought that becomes clear when going through the code. It's clearly in a raw state with lots of TODOs.

I've changed the description to reflect that more closely. Also the title says WIP.

And just last night I added a separate class where we can shift the functionality over, in order to implement things like component IDs and multiple gimbals properly.

So things will happen. Just be patient.

> This statement is unfortunately not true.

> ergo: not yet standard conform

Just for next time: I find that language hard to cope with. It comes across condescending (or in German rechthaberisch). 

---

Ok with that aside, let me respond to some of the (valuable) comments:

> If I'm not mistaken, the angles extracted from GIMBAL_DEVICE_ATTITUDE_STATUS will produce nonsense when the gimbal is tilted down

You're right. Remind me what we should use instead? Or what about you use the GitHub review feature which allows you to do a **Suggestion** where you can put your code in directly.

---
>> GIMBAL_MANAGER_STATUS. I want to request that topic at a low rate

> While a technically possible approach, I think it is the least desirable. I think, the manager should sent it out voluntarily. The reasons are quite similar to why a component is supposed to send out the heartbeat voluntarily, and not only then the heartbeat is requested. Moreover, the manager can do smarter, e.g. send it out more frequently when some status has changed. In short, it makes things just so much easier. Protocol wise, code wise, implementation wise, traffic wise, feature wise.

This is interesting. I had forgotten about it. The reason I suggest to request it from QGC is that is - I believe - generally how it's done. So by default an autopilot sends very little or no data but topics are requested as required. For instance if a ground station does not care about gimbals it shouldn't be flooded with gimbal messages. It depends on bandwidth and use case.

However, as you write, it would be good/best to send it out whenever there is a change and at a low rate in-between. So therefore, the optimum would be, if we could request the message at a low rate, and also get updates whenever it changes. Unfortunately, that's not exactly an option of [MAV_CMD_SET_MESSAGE_INTERVAL](https://mavlink.io/en/messages/common.html#MAV_CMD_SET_MESSAGE_INTERVAL). Looking at it, what is an option is "request it at default rate" and I guess that would make most sense. The gimbal manager can then interpret that as 0.5 Hz + updates.

julianoes:
@olliw42 regarding the build problems, I would create a separate issue. I don't use Windows or QtCreator, so I probably can't help you.

olliw42:
frankly, I think you want my language to come across like this. No chance to do it right, I give up. Cheers,

julianoes:
@olliw42 I'm sorry I was assuming you had seen my other post where I wrote that I actually agree with you on the fact that iterative doesn't always work, for instance when it prevents us from being compatible in the long term. I had only realized that when I was actually working on the PR. But I can't find my post anymore so the dog must have eaten it (or I didn't hit send).
That's why I was surprised why you still really needed to prove that you're right even though I had already agreed with you on that.

Sorry about that! Anyway, you're right, but let's turn that into action and get this PR in a good shape that works for STorM32, ArduPilot, and PX4.

rmackay9:
Hi there, AP now supports the GIMBAL_MANAGER_STATUS and MAV_CMD_DO_GIMBAL_MANAGER_CONFIGURE messages ([PR is here](https://github.com/ArduPilot/ardupilot/pull/23737)) so I think the top checkbox in this PR's description can be checked now.  

There's actually little more work we need to do to better enforce only one controller of the gimbal but that shouldn't hold up this PR.

The next thing we're planning is better discovery of the camera and FOV feedback.

Davidsastresas:
Thanks @rmackay9 . I think next week I will be able to flash AP master on my testing vehicle and advance on the QGC front on top of the amazing work Julian is doing. I don't think we are too far of having this ready. Thanks!

julianoes:
@Davidsastresas I just rebased this on master and continued a little bit:
- I removed some of the Joystick stuff, hoping to add it back in but better and using rate control.
- I added another button to the gimbal UI where we can select the gimbals (if there are multiple connected). I'm not convinced this is the right way and it's not hooked up yet :(.

Davidsastresas:
Thanks! I think next week I will be able to give it another push.

Regarding joystick, funny that you mentioned it, I worked a bit over it a few weeks ago, and I thought on bringing it back. Indeed when this gimbal manager integration is finished it would be great to link it to joystick for gimbal pan/tilt.

Thanks!

Davidsastresas:
I spent some time today testing this with Ardupilot master. The handshake, creation of gimbal instance, etc work perfectly on Ardupilot for single gimbal, but for multiple gimbals it doesn't work, because the messages GIMBAL_DEVICE_ATTITUDE_STATUS would come from the same compid. We have a device id in all the control commands, but not on this feedback message. Julian made a PR that fixes this.

https://github.com/mavlink/mavlink/pull/2009

So after that PR makes it upstream and we adapt the current QGC implementation to account for it, I think everything should be ready from the Autopilot firmware side, and only polishing details on QGC would be needed.

Thanks!

rmackay9:
Hi @Davidsastresas,

Maybe I'm misunderstanding something but AP doesn't send the gimbal-manager-status by default which means the default rate is zero so I think QGC will always need to say what rate it wants (when talking to AP).  I guess we could change it so that AP does send it by default but at this point there are very few consumers of the message so this seems like a waste.  Better for QGC to just request it I think.

0.2 hz seems slow to me but I don't mind really. if you guys think that's fine then I'm fine with it.

Davidsastresas:
@rmackay9 sorry, maybe I wasn't clear. What you said is exactly what I meant. AP doesn't have a default rate for that message, which makes sense as functionality around it is not implemented, so I think it is totally fine that AP doesn't have a default rate for it for the moment.

So right now QGC will ask 3 times to the autopilot to stream that message at the default Autopilot rate. If the autopilot doesn't have a default rate for that message ( default rate 0 ), then QGC will try to request that message at 0.2 Hz.

I do agree 0.2 Hz is very slow, but that message is only carrying information about the ownership of control for that gimbal. So I really think those 0.2 Hz, plus AP sending that message whenever control ownership changes, could be enough. I didn't want to use a higher rate precisely because what you said, there are very few consumers of this message, I don't think it makes sense at this point to use a much higher rate, would be a waste of bandwidth.

rmackay9:
@Davidsastresas, ok, sounds good!

Davidsastresas:
**I rebased to master and push -f, in case anyone is tracking this work**. It was needed because lots of commit got into master since we started this, better to keep it updated.

I added a few commits:
- I brought the latest changes from mavlink around gimbal manager v2, as it is not WIP anymore
- I increased the request status message retries to 6, for some reason for Ardupilot it wasn't enough sometimes at 4, specially when QGC is open, and Ardupilot boots up right when connecting with QGC. As this only happens on the handshake I think it isn't a big deal to increase by 2 the request attempts.
- Latest commit is a lot of rework over GimbalController class. The initial implementation would discover gimbals based on heartbeat, and the internal sorting of gimbals was based on the compid the message they were coming from. This would not allow proper functioning of multigimbal systems where all the gimbals come from the same compid. So I implemented it following:

https://mavlink.io/en/services/gimbal_v2.html

Particularly the part "discovery of gimbal manager" where it says something like "Discovery should be based on GCS sending a MAV_CMD_REQUEST_MESSAGE for GIMBAL_MANAGER_INFORMATION and we should instantiate gimbals when we receive response to this message."

Doing it this way allows it to work for both, gimbal managers on different compids and gimbal managers within same compid.

I tested this on Ardupilot SITL simulating several gimbals and it seems to work fine. Although more testing is needed, maybe we need to polish some details about it once the work keeps advancing.

**NEXT STEPS BEFORE BEING READY TO PUSH**
- I would like to do a rework on the frontend, to allow a nice way to select the active gimbal. I am now moving the gimbals to a qmlobjectlistmodel list, so it is nicer for this purpose.
- Add actions to GimbalController: right now they are over vehicle, and some of them are not using gimbal manager v2, we need to update all actions ( neutral, retract, etc ) to gimbal manager v2. Some of them would be good being buttons-indicators, like yaw lock/unlock, getting the flags from gimbal device I think that can be done now that all the parties have it implemented, it is much more compact and intuitive this way, rather than buttons that send actions with no feedback.

But I think we are close to having it ready. Thanks everybody involved, I am very happy this is moving forward!

Davidsastresas:
In case anybody is following, I rebased to master and pushed -f. I will spend some time on this in the following weeks so it was needed.

Davidsastresas:
I pushed a bunch of commits:

- A debug panel to show gimbal pitch and yaw, just provisional to make debug of functionality easier
- Add representation of gimbal yaw to vehicle overlay in map, it started as a debug functionality but It is actually cool:
Screenshot from 2023-12-07 16-09-30
It will probably change in the future, but it is handy to leave it there now to have visual feedback of camera yaw for debugging.
- Most of gimbal functions that used to be in vehicle moved to gimbalController. This has been moved as it was in vehicle, but those functions need rework, to take into account gimbal control adquisition, etc. Also some of them are APM specific, like setting rc targetting mode.
- Adapt front and backend to new location of gimbal functions.
- Fix gimbal yaw interpretation from gimbal_device_attitude_status. The approach used using Eigen would only interpret yaw as 0-180, and when negative values ( left semi circle considering nose of vehicle ) it would repeat again positive 0-180. It uses now mavlink helper quaternion_to_euler.
- Make gimbal a qobject, and expose qproperties so qml can pick mount telemetry from there. Right now this is not used, right now front end is using the remains of gimbal present in vehicle, but will be moved to this soon. Also this change to Q_Object is needed in order to support properly multi gimbal control.

As it is now, everything works with Ardupilot. Work pending to be done for a minimum working version of this:

- Clean all the functions in gimbalController for gimbal actions. Remove apm specific code, and make sure gimbal adquisition etc is implemented as it should.
- Finish porting all gimbal parts from vehicle to gimbalController
- Finish gimbalController functionality to support multigimbal control, right now it is controlling first gimbal only. This will need to adapt the q_invokables, and also the gimbals vector, convert that to qmlobjectlistmodel so it plays nicer with qml.

Besides the above, what I would like to do before considering this finished:
- Rework UI. With the new implementation completed, hopefully we can make it much more compact ( merge yaw follow/yaw lock button for example as we could have status feedback ). Also I would like to add some indicators of gimbal pitch and yaw. There is a debug panel already showing this, but must be implemented nicely.
- Fix AP camera setup submenu. It needs to get updated as it is referencing a lot of outdated params right now.
- Click on screen functionality: needs to be cleaned up. Settings should be elsewhere ( right now they are in vehicle camera component menu ), probably accesible through the flyview UI, for quick setup of FOV, etc.

Davidsastresas:
I pushed a bunch of new commits. The news are:

- Minor UI fixes: disable map tools when in video view, automatically disable "click on map to get coordinates" when that menu is hidden, etc
- Provisional button to manually release control. It will probably evolve in a release/adquire control button, depending on the current status. Although automatic gimbal adquisition was implemented, more below.
-  The concept of "active gimbal" was implemented, so now all the commands refer to this gimbal, and it was properly linked when sending commands, filling the proper deviceId field. From here, just changing the active gimbal would allow us to operate multiple gimbals, it is what I am working over right now.
- Adquire/release control is working. I tested with multiple instances of QGC with different sysid. When attempting to send a gimbal command, if we are not in control a popup will appear asking us if we want to acquire control. This only happens when others are in control. If nobody is in control, QGC will attempt to get control, but will send the command anyway.

Davidsastresas:
I pushed another bunch of commits. Everything is in a pretty good state right now, I think we could start thinking on trying to get more people involved in testing, in case you are interested @rmackay9. I still want to finish some details before considering to merge this, also in case @julianoes gets to work on this before I come back to it:

- Multigimbal telemetry: right  now we are only showing the active gimbal
- Limit how often the "acquire control" popup can appear. This will popup whenever we attempt to do a command without being in control. It could potentially appear a lot of times in a row, so I want to limit this to only show maximum one every few seconds. video showing this here: https://drive.google.com/file/d/1aDYX5VKTJNY2Vb6BUi-b6FU7uFevOOUL/view?usp=sharing
- Double check destructors: Would be good to link with vehicle destructor to delete gimbal objects.
- Redo UI: The one present right now really needs to be redone. We can merge a lot of buttons into a single button changing functionality depending on state, etc also it could be prettier, some panels are not aligned, etc

The new changes are:

- Acquire/release control button is now automatic: Will show/do release or acquire control depending on if we have or not control.
- Multi gimbal working: There is a select button in the gimbal toolbar, after clicking we will be able to select the current active gimbal. All the commands and telemetry will correspond to this new active gimbal
- Joystick button control working: We can configure now joystick buttons for gimbal up/down/left/right and center.
- Vehicle: I think everything related to gimbal is already moved to gimbalController

Maarrk:
Hi, we're working on our gimbal, and it would be great to get it ready for the v2 protocol from the start.

I see that this has already had a lot of work done, can I ask @julianoes and @Davidsastresas whether you still expect big changes, or just some GUI improvements here and there? I'd love to start using these features ASAP, and just wanted to know your advice on this.

Thanks a lot for the feature, and sorry if I'm bothering you with this

Davidsastresas:
Hello @Maarrk,

I would say the backend is pretty close to be completed, although some small details could be polished still. I would recommend to double check with the gimbal manager V2 documentation from mavlink in case you have doubts with any of it.

But as it is right now, I would say it is very close to what it will be on the first iteration, so for the most of it you can take it as reference for your system.

The UI related to this will change a lot probably, the one present right now was the first proof of concept and will be redone before this PR is completed.

If you want to discuss something in particular feel free to ping me on discord, I am davidsastresas.

Happy to hear a manufacturer is joining this! I wish you best luck on your projects. Best regards!

DonLakeFlyer:
Why is FakeToolStripHoverButton needed?

DonLakeFlyer:
Also need some screen shots of how this all works/looks ...

DonLakeFlyer:
I tried rebasing this myself so could take a look but it turned into a nightmare. It's been sitting too long. Would be good to at least get it rebased.

julianoes:
I had done a rebase recently and not pushed, I think. Let me check.

DonLakeFlyer:
So in looking at this my thinking is that the Gimbal stuff should be a new Toolbar indicator as opposed to controls which are always showing on the Fly View. Then with the new Indicator stuff you can use expanded page for settings and so forth.

rmackay9:
@DonLakeFlyer, thanks very much for the reviews!

Davidsastresas:
I will finish this PR, and rebase it to master, and to 4.3 in the following days. Thank you very much @DonLakeFlyer for taking a look at it. Answering your points:

## FakeToolStripHoverButton: 
This is a work in progress, it is meant to be used in a panel that could combine those buttons with labels, etc in multiple columns and rows, but preserving the aesthetics and dimensions of toolbar buttons. I was playing around with the idea of such a panel, and I found the extra complexity of ToolStripHoverButton, depending on a toolStripAction, harder to play with on this panel concept. Right now it is only used for gimbal selection. Most likely I can get this to work with new properties over ToolStripHoverButton, but as I still don't know what we will finally need I found faster to iterate this way, in a separate UI element.

## Screenshots: 
Yes, definitely. One of the pending items is to redo the UI, so when undrafting this PR my idea was to present a short video and/or screenshots to make evaluation of the PR easier. As the UI was still evolving I was holding on this.

## ROI:
 Video click stuff: I am not sure I follow you on this one. I think maybe you mixed 2 functionalities, the ROI panel, which has functionality that effectively overlaps with standard QGC ROI ( more on that below ), and the click over video to point gimbal.

The click on video feature is meant to be used like this ( min 0:43 - 1:10):

https://www.youtube.com/watch?v=ROyj6F6XieE&t=19s

So if this is what you meant, I don't understand how do you suggest we use DO_SET_ROI_LOCATION instead. That would mean calculating the angle our click represented, interpolate that with terrain altitude to obtain a ROI, and send that instead. Although overkill just for this "click and point" feature, it would be pretty cool. but it would have some issues like, what happens if we click in something in the sky, in a way that we don't have terrain to interpolate with, where we would set the ROI in this case?

Now, If you meant the ROI panel, I agree, that overlaps functionality with the existing ROI functionality. I was playing with the idea of having a panel to introduce manually the coordinates for ROI, as well as the Altitude for ROI, and I also left the shortcut to "get coordinates from map" in case that was useful. It is not polished and it needs some fixes, but I think the concept is cool, a panel to work quickly with ROI, rather than relying on clicking on map and confiming slider every time. Clicking around over the map is a good way of looking around with a gimbal, and I was playing with this. If you don't think this is useful or we should do it differently I am of course open to suggestions.

Please clarify what you meant on this point, I don't think I got it.

## Toolbar indicator versus current panel in fly view approach: 
The idea was that this new toolbar appears in fly view only when a gimbal is detected, so It won't be there all the time. And even when it is there, there is a button to hide it, so It barely occupies the space of a regular toolbar button when collapsed.

At first I thought about adding all of this in a top toolbar indicator, but after playing with the concept, I really don't think it makes sense for gimbal control. **I 100% see your point, the top toolbar is a great place for this on paper, but I think it is quite different from what we usually have in the top toolbar**.

In a vehicle with a gimbal attached, unless it is a survey vehicle and the gimbal stays on the same place most of the time, the user is going to interact a lot with it. And even though the top toolbar is "quick" to click, it is not near as handy as the fly view space is. The fly view toolbar has permanently some actions like Land or RTL that are only triggered once or twice per mission, and yet they have their own super handy space in the left margin. Same with PlanView button, lots of users don't even use plan view at all. These spaces in left and right margins are super handy on small mobile devices because they are within range of the thumbs. You can click them without moving your hand from "stick position". And that is why I tried to place this handy in the flyview, because the user is really going to interact with them, and the handier they are, the nicer the user experience is going to be with gimbals.

Again, This is a draft PR, and I am more than open to discussion, If you consider we should place it following the top toolbar scheme, maybe we can do it like that on default, with a very compact set of controls on the dropdown panel, but then in the expanded settings we could allow the user to place this over fly view instead. 

Would you be okay with this? I really think these are controls the user will interact with a lot, so I think it is worth at least leaving it optional. 

Thank you very much for taking the time to review all of this!

DonLakeFlyer:
> Now, If you meant the ROI panel, I agree,

Yes, this is what I was talking about. As far as editing the position of a point on the map there is already scaffolding for this. For example, modifying the position of a survey area polygon vertex. Click the vertex, then select edit (and I just found a new Qt6 bug that needs fixing!):
Screenshot 2024-03-05 at 8 58 39 AM
This is used all over Plan.

Screenshot 2024-03-05 at 8 52 30 AM

That's a generic position editing dialog which is available for use anywhere. I had already been in the process of cleaning up ROI last week. I got rid of the ROI toolbar thing which didn't make sense. You now just click the ROI to remove it. I was planning on adding drag support which is also a standard ability of map indicators (as is click signaling support). Adding position editing I think makes sense as well. I can do that. You are also correct that the whole slider to confirm an ROI is total overkill. I can remove that.

> Toolbar indicator versus current panel in fly view approach

I don't disagree with the toolbar not being quite a fit for this. That said my main complaint is adding yet another new UI thingy to control the gimbal. As well as chewing more real estate in the fly view. There are already plenty of usable controls/places to support this. Less is more. Also more custom controls that are only used a little create maintenance overhead (as you can see from all the stuff Qt6 broke). There has to be high bang for the buck. The regular Toolstrip seems perfectly fine for it. I was able to hack a mockup in QGC using your branch in a couple minutes:
Screenshot 2024-03-05 at 8 49 21 AM
This uses all the currently available UI parts of QGC to make it work. And it doesn't chew permanent valuable real estate from the fly view. You can put icons in regular QGCButton (although I discovered yet another Qt6 break that icons positioning is broken).

DonLakeFlyer:
Also if the ToolStrip action stuff is too much of a pain to work with we can change it to simplify. No need to make a whole new thing. I already need to go in there to add support for custom builds to add actions to the Fly Toolstrip since I need that myself. When I do that I'll look at revising how that whole thing works.

Davidsastresas:
@DonLakeFlyer I get your point, thank you.

On the next following days I will attempt to rework all of it to better stick to your guidelines. I totally understand your point of maintenance overhead, etc So I think it makes sense to do this first iteration as little invasive as possible.

However, if after this first iteration we find it isn't optimal for gimbal/camera control, I would like to review this again, even if that means major rework of fly view UI. I think cameras/gimbals are a very important part on a good portion of the userbase of QGC, so if after this first minimal/non-invasive iteration we find there is really a lot of room for improvement in user experience, I will pursue that when the time comes, I hope that is fine for you.

Thanks!

DonLakeFlyer:
> However, if after this first iteration we find it isn't optimal for gimbal/camera control, I would like to review this again

Absolutely. That's true for anything.

Davidsastresas:
Since this was included here https://github.com/mavlink/qgroundcontrol/pull/11276 and that other pull is basically where we are iterating the fixes for this gimbal support, I think we could close this one.

PR #10677, opened 2023-05-04
Fix parsering of hygrometer message

roman-dvorak:
Hi, 

[Hygrometer](https://mavlink.io/en/messages/common.html#HYGROMETER_SENSOR) mavlink message uses (is defined) centi-Celsius-degrees units (C*100). But in QGC, the message is currently parasersed as if they were the basic grades of Celsius. 

This PR converts the value from the message to basic unit DegresCelsius. 

image

image

image


booo:
Could you use a more descriptive commit message? Your explanation in the PR is very nice but commit message/description lacks behind.

patrickelectric (approved):

Issue #10557, opened 2023-02-02
Survey produces no mission items if Mission is in "Terrain Frame"


AndKe:
## Expected Behavior
Survey grids should work equally _ regardless of altitude frame.

## Current Behavior
Survey grid is not generated (as output during upload) if terrain frame is selected  

## Steps to Reproduce:
please see video: https://drive.google.com/file/d/1cO4LV8ajPKtE2si55mNuFMI5NBMEyVZg/view?usp=sharing

## System Information
- Operating System: Ubuntu22.10
- QGC Version: 
- QGC build: current HEAD
- Flight Controller: SITL
- Autopilot (with version): Copter, current release.

## Detailed Description
- video says all.

AndKe:
update: This is not really related to mission start, but rather the altitude frame of the survey:
![Screenshot from 2023-02-02 15-12-49](https://user-images.githubusercontent.com/1468842/216348663-e7623fb4-b7c6-492d-bfcf-eabe8fe5b62b.png)

AndKe:
@Davidsastresas how is your schedule?  - back in February I think this was #6  :)

Davidsastresas:
Sorry for taking this long @AndKe. I was still busy with other ongoing projects but I spared some time to fix this one:

10678

It turns out to be a bug where QGC was expecting to have terrain data on the surveys if these had terrain frame. The behaviour you showed in your video was only present when this terrain info was not present, with terrain info the survey was generated correctly, so I pulled that string and it turns out that was the reason.

Thank you very much for reporting this, and sorry for taking this long to address it. Let me know how it goes, and if it fixes the problem for you please close the issue and comment on the PR. Thanks!

PR #10678, opened 2023-05-07
TransectStyleComplexItem.cc: Fix bug generating empty surveys when no terrain data

Davidsastresas:
Selecting terrain data on surverys was generating an empty survey if the polygon was on an area where QGC doesn't have terrain height info about.

This is because QGC was waiting on terrain info to build the path on such surveys, which is not needed at all, because terrain frame just specifies an altitude above terrain, and relies on the vehicle for being able to calculate that height, by terrain database or by altitude sensors, having terrain height information is not needed at all to generate this kind of surveys.

I think this is heritage of a few time ago where QGC didn't support terrain frame for missions, and it was supporting only what we call now AltitudeModeCalcAboveTerrain, so this comprobation was needed.

I double checked what happens on this situation selecting AltitudeModeCalcAboveTerrain instead, and it works correctly, instead of uploading an empty survey it throws an error: 

"Plan is waiting on terrain data from server for correct altitude values."

So I think everything should be working correctly now. In any case, I would appreciate if @DonLakeFlyer and @booo can double check this makes sense.

Thank you very much @AndKe for reporting this issue #10557. Please let me know if this fixes the issue for you.

### DIFF
--- a/src/MissionManager/TransectStyleComplexItem.cc
+++ b/src/MissionManager/TransectStyleComplexItem.cc
@@ -422,12 +422,14 @@ void TransectStyleComplexItem::_rebuildTransects(void)
         return;
     case QGroundControlQmlGlobal::AltitudeModeRelative:
     case QGroundControlQmlGlobal::AltitudeModeAbsolute:
-        // Not following terrain so we can build the flight path now
+    case QGroundControlQmlGlobal::AltitudeModeTerrainFrame:
+        // Terrain height not needed to calculate path, as TerrainFrame specifies a fixed altitude over terrain, doesn't need to know the actual terrain height
+        // so vehicle is responsible for having or not this altitude calculation, so we can build the flight path right away.
         _buildFlightPathCoordInfoFromTransects();
         break;
     case QGroundControlQmlGlobal::AltitudeModeCalcAboveTerrain:
-    case QGroundControlQmlGlobal::AltitudeModeTerrainFrame:
-        // Query the terrain data. Once available flight path will be calculated
+        // Query the terrain data. Once available flight path will be calculated, as on this mode QGC actually calculates the individual altitude for each waypoint
+        // having into account terrain data.
         _queryTransectsPathHeightInfo();
         break;
     }


AndKe:
@Davidsastresas - well done ! It works as expected - thank you.

AndKe:
@Davidsastresas Thank you - this was a major issue for radar use with QGC.

DonLakeFlyer:
Sorry folks, been out of town for a while. I think the reason for terrain frame also waiting to query terrain data was so that the terrain graph would show correctly. Although the flight plan itself doesn't really need terrain data since the vehicle is in control with terrain frame the Plan UI needs it to display the terrain graphs doesn't it?

I haven't been in this code for a while, so I don't quite remember how all this works. With this change does the terrain graph work?

AndKe:
@DonLakeFlyer  his change does not affect how the elevation profile is displayed (it's all at zero altitude when I checked) for two reasons:
The current DEM source does not include my latitude (69.6°)  or where the relevant missions I had were >78°.

The elevation graph is kind of broken anyway because it's missing color coding of relative/ABS/terr/ etc. 
I made a mockup of a color-coded suggestion but failed to find it now.
Flying a lot, I can tell that it's a higher value to clearly distinguish between altitude-types, than the altitudes - or to put it other way : takeoff to rel50m then flying to an abs150m waypoint can mean flying into the ground.   (as the takeoff site may very well be at much higher altitude). Hence: before the graph is of much use (for those that have the DEM) - the altitude types are super-essencial.

As for DEM for the northern hemisphere around Norway - there are open/free DEM sources of very high quality for our latitude ..except there is no way for QGC to use them.

DonLakeFlyer:
I get it doesn't work for the area you are flying but it should work if there is terrain available.

I don't get your complaint about the current display not being valuable. There are basically to lines:
* The flight path altitude (whatever frame you are using)
* The terrain altitude

If those two lines cross the segment turns red to indicate you will hit terrain and the Plan won't be uploaded.

Maybe as you say a picture which show the problems would help me understand.

AndKe:
@DonLakeFlyer 
Please see attached a mixed mode mission: (had to compress it to make github happy)

[Bosleyfjellet.zip](https://github.com/mavlink/qgroundcontrol/files/11525517/Bosleyfjellet.zip)

it looks like:
Screenshot from 2023-05-21 22-28-56

having different color dots/lines for different terrain frames would be helpful.

AndKe:
Yes, my bitching about terrain is kind of unfair, because I always thought of it as a local issue, but given that https://terrain.ardupilot.org provides coverage up to 84°N  - I start to suspect a bug more than a real limitation of QGC.

I would absolutely love to see the terrain-graph working as intended. It would make it easier to validate a mission near mountainous terrain. (basically everywhere in norther Norway)

DonLakeFlyer:
Terrain data comes from Airmap. QGC can only work with what they have. At one point Auterion has talked about taking over the terrain server support and providing new data. Didn’t get anywhere though.

As far as showing terrain frame modes in the graph that would be nice additional information. But that isn't what the initial concept was for. The goal was to show whether you were going to fly into the ground or not. Which can be done without that level of display.

As you can see from @AndKe's example the terrain data is yellow and 0 which means unavailable. My worry is this pull causes that to happen also in areas which have terrain data available to it when you are using terrain frame.

AndKe:
Would't it make sense to get the data from https://terrain.ardupilot.org ? (or the same source)
Airmap's world view of 50°N to 50°S may be suffient for flat-earthers, but otherwise .... it's simply inadequate.
(comment based on https://developers.airmap.com/docs/elevation-api)

DonLakeFlyer:
"Would't it make sense..." Only if it didn't require rewriting the terrain server query code, which it does. So unless someone wants to write the code ...

DonLakeFlyer:
I don't remember exactly because I didn't write this code Gus Grubba did. But I think Airmap funded the development of it.

AndKe:
Who's involved in ArduPilot's terrain service?  Maybe that person can tell if that server can serve the same kind of elevation data.
Also: maybe mapproxy together with terrain.ardupilot.com could translate the format. 
Maybe this could be discussed at next Ardupilot devcall?

DonLakeFlyer:
Don’t know if anyone is actively worked on it in ages. ArduPilot format is a set of files you download by lat/lon with a url format. They got the data from a flight sim app ages ago. AirMap is a rest (I think) json api, which return in a json format. Totally different. Certainly doable to switch with a medium amount of effort. Not easy, not hard.

From: André Kjellstrup ***@***.***>
Date: Tuesday, May 23, 2023 at 13:58
To: mavlink/qgroundcontrol ***@***.***>
Cc: Donald Gagne ***@***.***>, Mention ***@***.***>
Subject: Re: [mavlink/qgroundcontrol] TransectStyleComplexItem.cc: Fix bug generating empty surveys when no terrain data (PR #10678)

Who's involved in ArduPilot's terrain service? Maybe that person can tell if that server can serve the same kind of elevation data.
Also: maybe mapproxy together with terrain.ardupilot.com could translate the format.
Maybe this could be discussed at next Ardupilot devcall?

Davidsastresas:
> Sorry folks, been out of town for a while. I think the reason for terrain frame also waiting to query terrain data was so that the terrain graph would show correctly. Although the flight plan itself doesn't really need terrain data since the vehicle is in control with terrain frame the Plan UI needs it to display the terrain graphs doesn't it?
> 
> I haven't been in this code for a while, so I don't quite remember how all this works. With this change does the terrain graph work?

I just tested this and it works as it should. If no terrain data available the graph is not drawn, and as soon as terrain data keeps coming the graph is shown perfectly, so I think we could forget about this.

Regarding the discussion about other elevation map providers, it is interesting and I think it has value. Maybe we should move the conversation elsewhere, to an issue?

I am thinking maybe there are other providers that could serve the terrain data similarly to how we are getting it now from airmap. Maybe some of those servers have nicer resolution, or resolution at all at extreme latitudes. It is worth checking, and if we find a similar service provider the implementation could be much more straighforward than adapting the whole tile server to Ardupilot data, which as @DonLakeFlyer said, was taken from a flight simulator a long time ago, I am sure we can get some providers with updated and nice data, although maybe they require subscription and api key, but it could be justified for some users. 

Anyway, lets open that conversation in an issue to see if more people is interested on it.

AndKe:
@Davidsastresas  : https://github.com/mavlink/qgroundcontrol/issues/10702   :)

PR #10680, opened 2023-05-08
VideoReceiver: call stop and cleanup teeProbe

julianoes:
This fixes regular segfaults on exit for me.


booo:
Looks good. Thanks for your contribution. Next time please add some details about the segfault if you can.

julianoes:
Thanks @booo and you're right, that would have been helpful. I'm pretty sure the normal gdb backtrace wasn't conclusive. I then used address sanitizer which pointed me to the right direction.

PR #10686, opened 2023-05-15
Auto load : Implementing Auto load feature.

SebastianViollaz:
With this feature now we can save a .plan file in an specific folder and QCG will  automatically load to every dron when it connects. Also you can enable and disble this feature in the AppConfig Autoload checkbox if u like.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,6 @@
     "docs:preview": "vitepress preview docs"
   },
   "dependencies": {
-    "vitepress": "^1.0.0-rc.31"
+    "vitepress": "^1.0.1"
   }
 }


SebastianViollaz:
@Davidsastresas You are right. A confirmattion box could be a nice solution to the problem. Im going to implement that check box and make a new commit.
Also it coul'd be a little annoying to the user (at least for me), because i will have to always confirm. But i could add a "Don't ask me again" option.

SebastianViollaz:
Maybe can save the previus plan in a folder, so the user can access it anyways. But i think the confirmation is the best soluttion

Davidsastresas:
I really think for safety reasons we should make a popup appear whenever we have a plan file on that special folder. I think that is the only way of making this a good compromise between handy and safe. I don't think it is too inconvenient to just confirm a dialog on each QGC boot up, it is straightforward and we save a lot of trouble of possible misunderstandings of users, don't you think?

Thanks!

Davidsastresas:
Thanks, I saw you just fixed that. I am not familiarized with PX4, I am only familiarized with Ardupilot, but I have some concerns.

The way you check this, it seems _tryCacheHashLoad is meant for parameters, not related to plan, am I wrong? Also, what is the relationship with _HASH_CHECK px4 parameter here?

I am not experienced in PX4, but that looks wrong to me.

SebastianViollaz:
Yes it's related with the parameters and no the plan itself. Im looking where to put that. Like when QGC detect a new vehicle connection. I will make a new push correcting that.

SebastianViollaz:
@Davidsastresas  Done not sure why didn't pass the git checks. Any ideas?

SebastianViollaz:
Do u want to undo the commits and commit again only with my commits or something like that or just leave like this?

src/FactSystem/ParameterManager.cc
```diff
@@ -19,6 +19,7 @@
 #include "CompInfoParam.h"
 #include "FTPManager.h"
 
+
```
ahmedak1, line 22: Not necessary?

SebastianViollaz: No. In one of my previus commits (09b6219f41ed370848da3dede5a758b86049dca7) in the line 260 i write the logic to show the autoload checkbox so i need to #include SettingManager.h to acces to the settings. Then i change of place were to write it because here it wasn't the place.
the logic was remeved too but it doest show in git idk why. Check the line 260 to verify

src/QGCApplication.cc
```diff
@@ -746,6 +746,17 @@ void QGCApplication::showCriticalVehicleMessage(const QString& message)
         qWarning() << "Internal error";
     }
 }
+void QGCApplication::autoLoadCheck(){
+
+
+    QObject* rootQmslObject = _rootQmlObject();
+    if (rootQmslObject) {
+        //_settingsManager -> appSettings() -> missionSavePath() + "/AutoLoad.plan"
+        QVariant path = qgcApp()->toolbox()->settingsManager()->appSettings() -> missionSavePath() + "/Autoload.plan";
```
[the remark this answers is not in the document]
SebastianViollaz: Here i just execute askforAutoloadMission that is defide in MainRootWindow.qml. There i send the plan to the active vehicle. I call this in MultipleVEhicleManager.cc, i understand that when i new veihicle connects it became the active one rigth? so the plan will be sent to the last vehicule added.

src/FactSystem/ParameterManager.cc
ahmedak1: Yes. I just meant since the change got removed you can delete this empty line (line 22). There is no need to change the file anymore.

SebastianViollaz: ok

DonLakeFlyer:
I had implemented this same feature a while back for a companies custom build. I never brought it across to mainline because I remember running into also sorts of gotchas in making it work correctly and not screwing things up.

I think an example was lost contact which a vehicle which is already flying a mission. If the link disconnects and then you reconnect while mid-air you can't push the Plan to the vehicle since it is already flying a plan. If I remember correctly there were a pile of other cases like this that I can't remember. So be careful and think it through...

SebastianViollaz:
@DonLakeFlyer maybe we can load only if the state of the vehicle is landed and not armed yet?. 
Thanks for the feedback

DonLakeFlyer:
Yeah, I don't think the problems are insurmountable. You just have to be careful of all the possible cases.

ahmedak1:
I don't really have authority to merge by the way so there's no point in me reviewing it. Maybe you can remove me as a reviewer and wait for someone else to take a look at it?

julianoes:
Should we still get this in? It would require rebasing on top of `master`.

SebastianViollaz:
Im intereted in keep get this in. I´m here if you need to make any change, or you need ,me to do something

julianoes:
Please merge or rebase with mavlink/upstream master and fix the conflicts.

SebastianViollaz:
I will open a new PR due to isssues due to original repo

PR #10689, opened 2023-05-21
Add support for easier interaction between Ardupilot users and Ardupilot remote support server

Davidsastresas:
The functionality was already there with mavlink forwarding, but it wasn't totally convenient because:

- It doesn't disconnect automatically when QGC restarts. This is indeed very cool for standard mavlink forward, because the user probably wants to forward again on next boot, but for this purpose is not desirable because users can forget to disable forwarding and they would keep forwarding to the server without knowing it.
- If the users wanted to forward traffic to another device during the remote support session, it would not be possible.

This PR addresses the issues above.

On the front end, a new configuration component was added to Ardupilot settings menu, and also an indicator in the top toolbar ( only to APMfirmwareplugin, overriding toolbar indicators ) was added so the user is extra aware the traffic is being forwarded.

On the backend, we pretty much replicated what is being done with the mavlink forwarding feature we already had, only instead of at bootup it connects after clicking the button, and the flags in mavlinkprotocol are different as well. Some refactor was done in linkManager as well to not replicate code with this feature. A new setting was added to appSettings to hold the name of ardupilot support servers, so after reboots QGC saves the latest server name and thus it is handier to replicate the remote support session.

Here is a short video explaining the UI: 

https://drive.google.com/file/d/1BnfAmvep__wpZrNzE9JQZl9SiscaPKia/view?usp=sharing

Thanks to @tridge for pointing out the need of this feature in QGC!


tridge:
@Davidsastresas brilliant, thank you! This will help a lot

src/comm/LinkManager.cc
```diff
@@ -185,6 +186,19 @@ SharedLinkInterfacePtr LinkManager::mavlinkForwardingLink()
     return nullptr;
 }
 
+SharedLinkInterfacePtr LinkManager::mavlinkForwardingSupportLink()
+{
+    for (int i = 0; i < _rgLinks.count(); i++) {
```
patrickelectric, line 191: Please use range based loop over manual index.
https://en.cppreference.com/w/cpp/language/range-for

patrickelectric (approved):

src/comm/LinkManager.cc
Davidsastresas: Thanks for the suggestion, I am used to this manual index approach. Just out of curiosity, is there any benefit on the range based for versus manual index? Or is it just that c++ is evolving into that?

I reused that scheme used all over linkManager class. Maybe we should fix it in all the places in a separate PR?

src/Settings/App.SettingsGroup.json
```diff
@@ -325,6 +325,13 @@
     "longDesc":  "Host name to forward mavlink to. i.e: localhost:14445",
     "type":             "string",
     "default":     "localhost:14445"
+},
+{
+    "name":      "forwardMavlinkAPMSupportHostName",
+    "shortDesc": "Ardupilot Support Host name",
+    "longDesc":  "Ardupilot Support server to forward mavlink to. i.e: support.ardupilot:xxxx",
+    "type":      "string",
+    "default":   "support.ardupilot:xxxx"
```
hendjoshsr71, line 334: Small typo
```suggestion
    "longDesc":  "Ardupilot Support server to forward mavlink to. i.e: support.ardupilot.org:xxxx",
    "type":      "string",
    "default":   "support.ardupilot.org:xxxx"
```

hendjoshsr71 (changes requested):
Very nice!

One small typo.

Note during my testing when connecting  to the forwarded traffic via Mission Planner. MP is unable to download the parameters.  When using MP's original MAVLink  forwarding tool the cause of this was not checking the  "Write Access" box. 

This box seems to be reading all the incoming packets on the forwarded stream and then copying them to the original mavlink stream. 

See here: https://github.com/ArduPilot/MissionPlanner/blob/master/ExtLibs/ArduPilot/Mavlink/MAVLinkInterface.cs#LL5390C32-L5390C32

src/Settings/App.SettingsGroup.json
Davidsastresas: Fixed, thanks for spotting that!

Davidsastresas:
Thanks for the review @patrickelectric. I wasn't used to those range based loops, but I investigated a bit and indeed they are nicer. Thank you for the suggestion.

I modified it in the new function added on this PR, and it the standard mavlinkforwardlink() as well. There are more places in that file where it could be changed, and it would be probably safe to do it directly, but I would love to do that change and test it properly before committing, on a separate PR, if that is fine? I left that on a separate commit.

@hendjoshsr71 thanks for spotting that typo, it is fixed as well, that one is squashed on the first commit. About what you say on Mission Planner, maybe that is something that should be indicated in the docs right? But as I understand it, it should not affect this PR, is that correct?

Should we merge this?

Thanks!

Davidsastresas:
Thank you @patrickelectric, of course, I just rebased to master, and tested again the whole procedure and it works alright. I will merge it when the tests finish. Thanks!

PR #10694, opened 2023-05-24
Add GripperAction feature

DieBorr:
In this way if user has gripper_enable option enabled the UI will show gripper action in the UI as a common action like change altitude and if it's disarmed as a specific action with it's own symbol.


DieBorr:
Here is a example of how it works
[video_2023-05-24_22-00-32.zip](https://github.com/mavlink/qgroundcontrol/files/11558570/video_2023-05-24_22-00-32.zip)
:

DieBorr:
Here is a video of how it works with a servo Gripper.
https://github.com/mavlink/qgroundcontrol/assets/47715559/68b31880-829c-4741-8f60-dfc93dddd752

PR #10697, opened 2023-05-25
Add Gripper feature

DieBorr:
Now the user can use a gripper with the GCS UI, when the motors are not armed as a ToolStripAction and when armed as a common action. I have tested this feature with a servo Gripper as shown in the next video: 

https://github.com/mavlink/qgroundcontrol/assets/47715559/29924e27-d682-4048-b18e-a1de9dcae22d


booo:
Where do the GUI icons come from? Did you design them?

Looks quite ok to me. I don't have time for a proper review now. If someone else from the community wants to review please do so and I hit the merge button if you have consensus on a GO.

hamishwillee:
@junwoo091400 Perhaps you might like to look at this given your gripper experience.

DieBorr:
> Where do the GUI icons come from? Did you design them?
> 
> Looks quite ok to me. I don't have time for a proper review now. If someone else from the community wants to review please do so and I hit the merge button if you have consensus on a GO.

Thanks! Yes, the GUI icons are self-designed but not from me, they are from @alexdelatorre. Is nice to hear that the UI looks nice for you.

booo:
I find the naming a bit strange. In the GUI it is "grab", "release". In the code it is "drop", "release", "grab". Maybe we can decided on a naming schema and stick to it?

Given the increasing military use of UAVs drop also does not sound right. In Germany it is forbidden to drop stuff from UAVs too.

If you don't want to rename I'm also fine with a merge.

DieBorr:
> I find the naming a bit strange. In the GUI it is "grab", "release". In the code it is "drop", "release", "grab". Maybe we can decided on a naming schema and stick to it?
> 
> Given the increasing military use of UAVs drop also does not sound right. In Germany it is forbidden to drop stuff from UAVs too.
> 
> If you don't want to rename I'm also fine with a merge.

I'm completely agree with the naming problem, I have worked on it, now I think is more understandable, I have used "Release" instead of drop, also I have renamed some archives.
Referring to the gripper uses I think that each pilot should know the legislation of the place where the UAV is going to fly, regarding the military use of UAV's this feature has cooked on thinking on the hundreds and hundreds of useful possibilities that this feature can do so thinking only in the bad side and deprive the community of its use is not fair. Thank you so much for reviewing the code, this has been one of my first contribution's on QgroundControl project and I hope to continue helping the community.

booo:
Thanks for your swift changes and contribution. Looking forward to many more. Welcome to the community!

PR #10701, opened 2023-05-28
Master pr/support do set home

Davidsastresas:
This PR adds support for setting home clicking on map. It is shown as another action in the submenu when we click over the map with an active vehicle online. It needs slider confirmation as usual, with the following message:

set home guided indicator

As per mavlink specs, this command must use AMSL altitude, so we need to first query terrain altitude, and when received send the command.
    
Several checks have been implemented, one of them is a sanity check where the terrain altitude must be within a range (-500 - 10000 ) that should cover all places on Earth, in order to discard possible corrupted data, and in case terrain altitude is not available at the selected location a popup will appear indicating there are no terrain data for the specified location:

set home failed

I also added vehicle logging that reports some messages related to this when it fails, and also writes a debug message each time home location has changed, and I also extended the description of the "update home position" setting, as it could potentially conflict with this set home feature and I felt the existing tittle was not descriptive enough to make users aware of this.

I tested in Ardupilot SITL extensively and it seems to work all right and report the error popup correctly when terrain data is not available.

Testing in PX4 would be good before considering this good to go, so in case something needs to be changed for PX4 we can move all of this to FirmwarePlugin.

There is a matter that maybe could make sense to add to this implementation, which would be a timeout that aborts this operation if terrain query takes too long to get the data, and warn the user with a popup when this happens. Maybe this could make sense here, although according to my tests the location is always set within a few seconds maximum, or the popup error is reported. I would appreciate more opinions on this front @DonLakeFlyer @booo @patrickelectric as this is the first time I get in contact with the terrain query system.

Thanks to Arace ( https://araceuas.com ) for spotting the need for this one in QGC.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -62,6 +62,8 @@ QGC_LOGGING_CATEGORY(VehicleLog, "VehicleLog")
 #define UPDATE_TIMER 50
 #define DEFAULT_LAT  38.965767f
 #define DEFAULT_LON -120.083923f
+#define SET_HOME_TERRAIN_ALT_MAX 10000
+#define SET_HOME_TERRAIN_ALT_MIN -500
 
 const QString guided_mode_not_supported_by_vehicle = QObject::tr("Guided mode not supported by Vehicle.");
 
@@ -1482,6 +1484,7 @@ void Vehicle::_setHomePosition(QGeoCoordinate& homeCoord)
 {
     if (homeCoord != _homePosition) {
         _homePosition = homeCoord;
+        qCDebug(VehicleLog) << "new home location set at coordinate: " << homeCoord;
         emit homePositionChanged(_homePosition);
     }
 }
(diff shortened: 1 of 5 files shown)


booo:
Sorry, I don't have time for a proper review now but want to let you know that we had problems with different ArduPilot versions in the past and how they accept some home position updates and others not. This depends on the mavlink command you use to send. I think if you use the long_command method the change is the highest to get the home position update accepted. Maybe you can double check on that?

I also wonder if it should be possible to move the home point dot in fly mode. Would be another option to move the home point maybe.

Davidsastresas:
Thanks for the review @booo.

About Ardupilot digesting this or not, I tested in Arducopter master, but maybe it would be good if @tridge could verify this is fine for all AP vehicles, and fix it otherwise. The way it is done on this PR is sending a command_long with params1-4 at 0, and 5-7 indicating lat/lon and terrain altitude.

About moving home position over the map, I thought about it, and It would be very cool, but it would need some additional development on top of this. Also, I am not 100% sure if it would be wise to do so. 

Let me explain, with the current approach, the user needs to consciously select "set home" command. If only dragging home position it would be less obvious for the user what is happening. A double confirmation would be good on this front of course, but I also think it is faster with the current approach, because if home map indicator is too far away, users will need to zoom out, search for home element, drag it around, which could need map panning, etc.

So in short, I think click and dragging home point would be very cool, and I would like to implement it at some point, but I think it will not be really practical,  so from my point of view I would not say it is necessary for this PR. Do you agree?

thanks for the review!

timtuxworth:
Tested on the bench. It was simple, intuitive (click the map) and easy to set the home location on a QioTek Zealot H743 with no connected GPS running ArduPlane 4.3.7.

I have to install this device in a real plane, which might take a few days, I should be able to flight test it soon.

timtuxworth:
Does it do a live query for terrain data over the internet? What if there is no internet at the flying site?

Davidsastresas:
@timtuxworth Thanks for the test. Please report back when you test it on flight.

It does need terrain data, because the autopilots expect the altitude of the new home in above sea level altitude, so we need terrain data.

If there is internet, and the point we requested is not cached already, It will do a live query indeed.

If there is no internet in the flying site it will fail as per the second screenshot of my first message on this thread. If you plan on flying with no internet on a site it is always a good idea to download maps previously. When you download offline maps for an area, elevation maps are downloaded as well.

timtuxworth:
Is this a QGC function? I would usually have pre-downloaded the Terrain data onto the AP.

booo:
@Davidsastresas do you want to wait for feedback from a PX4 expert or should we merge.

To me it looks good and you can merge if you like to. Thanks for your work.

Davidsastresas:
Thanks @booo . I also think it is probably good, but I would like to wait for confirmation from somebody familiar with PX4, as this feature as it is now, over Vehicle and not over FirmwarePlugin, will be available inmediately for PX4, and if for some reason PX4 is not okay with this it could be dangerous, so please lets wait just a bit to have that confirmation. @mrpollo can you confirm, or can you point me to somebody that could take a look at it? Thanks!

mrpollo:
Hey @bkueng can you please review

src/Vehicle/Vehicle.cc
```diff
@@ -4074,6 +4077,60 @@ void Vehicle::flashBootloader()
...
+        // Coord and terrain alt sanity check
+        if (_doSetHomeCoordinate.isValid() && terrainAltitude <= SET_HOME_TERRAIN_ALT_MAX && terrainAltitude >= SET_HOME_TERRAIN_ALT_MIN) {
+            sendMavCommand(
+                        defaultComponentId(),
+                        MAV_CMD_DO_SET_HOME,
+                        true, // show error if fails
+                        0,
+                        0,
+                        0,
+                        0,
```
bkueng, line 4116: param4 should be NaN for default heading (mavlink spec: `NaN to use default heading`).

Davidsastresas: You are right, sorry I missed it, it is fixed now.

Davidsastresas:
Thanks for the review @bkueng, I fixed it and squashed it. I rebased it to master. Considering it is good to go for PX4 as well I will merge when the tests finish. Thanks!

Issue #10654, opened 2023-04-07
Remote ID showing failed GCS GPS for Live GNSS


dirksavage88:
Using the bluemark dronebeacon db201 connected over mavlink serial to the FC.

I am using a usb connected GPS module to the ground station and I am showing the live coordinates update in QGC, along with altitude. The GCS GPS still shows failed. Sometimes altitude will show negative, but this only happens rarely. 

Not sure if the GPS coordinates are updating fast enough, or the failed arm status has something to do with it. Heading shows N/A, horizontal accuracy around ~13m

- Operating System: Ubuntu 20.04
- QGC Version: main
- QGC build: linux, downloaded the appimage
- Flight Controller: holybro 6c
- Autopilot (with version): Px4 main
- Dronebeacon DB201
![Screenshot from 2023-04-06 21-13-46](https://user-images.githubusercontent.com/35986980/230521438-45567ffd-f65e-4d42-b23e-a01563ec3994.png)

BluemarkInnovations:
It could also be that the QGC RemoteID function needs to be adapted for PX4. This is what @Davidsastresas said back in the days "In case you are familiarized with QGC, we will implement it as a plugin in the vehicle class, and from there each of the firmware plugins ( px4 and Ardupilot ), might have some slight differences in how they send/receive messages, and this is the only possible extra development that would be needed to add PX4 support, so probably really straightforward"

Davidsastresas:
It could be the GPS not refreshing fast enough. Are you testing outdoors? if testing indoors this could be the issue. If the GPS data is older than 2 seconds it will be discarded ( as per remote ID specs ) and so the GPS indicator will be red. Also for FAA I think altitude is mandatory, so if altitude is not all right it will fail as well.

@BluemarkInnovations thanks for pinging me!

dirksavage88:
@BluemarkInnovations I have confirmed I have the correct serial port and there is nothing else on tty/ACM*. It also looks like the data is updating so it is not stale data from a previous connection. 

I still think the module is expecting something that we don't have explicitly stated in the mavlink stream: https://github.com/dirksavage88/Firmware/blob/3c0b8d348765a7470fa0ba422d16c07f3adfa22e/src/modules/mavlink/streams/OPEN_DRONE_ID_SYSTEM.hpp#L92

BluemarkInnovations:
[...]

PR #10704, opened 2023-05-29
Remote id fronted: hide all indicators if RID comms is not healthy:

Davidsastresas:
All this remote id feature depends on a device being detected so it makes sense to not show the rest of the flags until a healthy connection has been detected. Now it looks like this until we have healthy connection with RID device ( arm status message received ):

rid settigns
rid indicator

It addresses the issues mentioned here #10654.

### DIFF
--- a/src/ui/preferences/RemoteIDSettings.qml
+++ b/src/ui/preferences/RemoteIDSettings.qml
@@ -42,6 +42,9 @@ Rectangle {
     // Flag to get active vehicle and active RID
     property var  _activeRID:           _activeVehicle && _activeVehicle.remoteIDManager ? _activeVehicle.remoteIDManager : null
 
+    // Healthy connection with RID device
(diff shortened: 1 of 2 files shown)


booo:
Does it make sens to communicate by text what is happening? I get that red is "danger" or "something is broken" but I don't get that a device is missing from that.

Davidsastresas:
@booo It  makes a lot of sense, I hadn't thought about it. I modified it so now when there is no connection, the label shows "NOT CONNECTED" instead of "RID COMMS" just with red background. It looks as follows:

rid3
rid2

I squashed and extended the commit comment. Let me know if this looks good now. Thanks for the review!

booo:
Looks good now.

BluemarkInnovations:
I tested it with both ArduPilot and PX4. Looks indeed good.

PR #10706, opened 2023-06-02
install-qt-linux.sh: Remove use of soon to be deprecated aqt install

kribe48:
The command "aqt install" is deprecated and marked for removal. It has been replaced with "aqt install-qt".

### DIFF
--- a/deploy/docker/install-qt-linux.sh
+++ b/deploy/docker/install-qt-linux.sh
@@ -13,6 +13,6 @@ set -e
 apt-get update
 apt-get install python3 python3-pip -y
 pip3 install aqtinstall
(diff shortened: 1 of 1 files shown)


booo:
Did you test the docker build? Do we use the command in other places?

Cloud you maybe contribute a CI action that triggers a docker build too? :)

kribe48:
@booo I tested the docker build procedure according to https://dev.qgroundcontrol.com/master/en/getting_started/container.html and it seems to be working as it did before. I could give CI action a try! Never added one before. Do you want an action for each platform, or is linux sufficient?

booo:
Thanks for your feedback!

It would be awesome if you could contribute a CI action in another PR! I think an action on a platform you know would be fine for now. Until now we don't test the docker build at all as far as I know.

PR #10711, opened 2023-06-08
MainRootWindow: fix bug toolbar dissapearing:

Davidsastresas:
In the edge scenario where we set video view in full screen and for some reason QGC brings us to setup view ( if the vehicle has calibration/setup pending ) the toolbar would dissapear, thus not being able to get out of that state.


booo:
Thanks for your contribution.

PR #10712, opened 2023-06-13
Fix zh_CN translate error

cuav-zhang3:


booo:
Could you shortly describe what problem you are fixing? It is really hard for me to understand the change because I don't speak Chinese.

cuav-zhang3:
The Chinese display is abnormal. Because arrays are separated by ",". But it is translated as "，".
cd9da82b58f0310e323993dda5fde8d1

DonLakeFlyer:
This needs to come down through crowdin otherwise the next time someone changes the translations this will just get trashed again.

PR #10713, opened 2023-06-15
First version of a CI action for building QGC using a container on linux

kribe48:
This is under development, will assign a reviewer when I am done!

### DIFF
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -32,3 +32,4 @@ jobs:
         run: |
               mkdir build 
               docker run --rm -v ${PWD}:/project/source -v ${PWD}/build:/project/build qgc-linux-docker
+        run-as: non-root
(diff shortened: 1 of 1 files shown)


kribe48:
@booo This is as far as I get. The docker image is successfully built, but when I try to build QGC using the image I get permission issues. I believe that it is due to that the action is executed as root, which means that the instantiated docker container is unable to write the build files since it does not have root access. Do you know if there is any solution to this?

booo:
> @booo This is as far as I get. The docker image is successfully built, but when I try to build QGC using the image I get permission issues. I believe that it is due to that the action is executed as root, which means that the instantiated docker container is unable to write the build files since it does not have root access. Do you know if there is any solution to this?

Sorry, I'm also not that familiar with github actions and docker :/

Did you check the github instructions for actions and docker: 

https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions

Maybe we should not use the WORKDIR in the docker file?

DonLakeFlyer:
Closing this until it works...

PR #10715, opened 2023-06-16
[DRAFT] Include support for Ardupilot terrain servers as elevation data source

Davidsastresas:
I thought it would be good to start making public the current work done over a separate branch, so we can start testing builds when it is functional.

**current state of it:**
De coupling of Airmap in elevation architecture is mostly done, and database management should be fine for this new provider ( not tested yet ).

**latest commit is a work in progress, it doesn't build with it**

If the above indeed is good, we would need to sort out:
- getting url properly and be sure the file gets to be handled nicely by cache engine
- adapt whatever is neccesary in TerrainQuery and TerrainTile when it works with other sources than Airmap. Make sure here it is really provider agnostic, and ideally move the relevant methods to ElevationMapProvider.cpp/h 

Thanks @mwrightE38 for joining!

### DIFF
--- a/src/FlightDisplay/FlyViewMap.qml
+++ b/src/FlightDisplay/FlyViewMap.qml
@@ -588,11 +588,23 @@ FlightMap {
                         globals.guidedControllerFlyView.confirmAction(globals.guidedControllerFlyView.actionROI, clickMenu.coord, roiLocationItem)
                     }
                 }
+
+                QGCButton {
+                    Layout.fillWidth: true
+                    text: "Set home here"
+                    visible: globals.guidedControllerFlyView.showSetHome
+                    onClicked: {
+                        if (clickMenu.opened) {
+                            clickMenu.close()
+                        }
+                        globals.guidedControllerFlyView.confirmAction(globals.guidedControllerFlyView.actionSetHome, clickMenu.coord)
+                    }
+                }
             }
         }
 
         onClicked: {
-            if (!globals.guidedControllerFlyView.guidedUIVisible && (globals.guidedControllerFlyView.showGotoLocation || globals.guidedControllerFlyView.showOrbit || globals.guidedControllerFlyView.showROI)) {
+            if (!globals.guidedControllerFlyView.guidedUIVisible && (globals.guidedControllerFlyView.showGotoLocation || globals.guidedControllerFlyView.showOrbit || globals.guidedControllerFlyView.showROI || globals.guidedControllerFlyView.showSetHome)) {
                 orbitMapCircle.hide()
                 gotoLocationItem.hide()
                 var clickCoord = _root.toCoordinate(Qt.point(mouse.x, mouse.y), false /* clipToViewPort */)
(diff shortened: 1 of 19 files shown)


Davidsastresas:
I just force pushed after some tree cleanup and some fixes, thank you very much  @mwrightE38 for the help. In case anybody else is joining, I think the most urgent stuff to follow after the above is:

- AP terrain servers provide the data zipped, so we need to unzip it. We have some libraries to do so in QGC, but we don't have any that could be used in android right now. It seems QByteArray could decompress a .zip file, in fact there is an example of this in FirmwareImage.cc. But in order to use this QByteArray approach, we need to append in the first 4 bytes of the zip file the size of the uncompressed file, and I am not sure yet how feasible is this.

After that is figured out, I think it should go here: QGCMapTileset.cpp, _networkReplyFinished() so as long as we fetch it online we unzip it and we store it unzipped, so we no longer need to worry about it.

- After the above is sorted out we need to get into TerrainQuery and work on the Ardupilot terrain servers format. Mission planner can be checked for inspiration:

https://github.com/ArduPilot/MissionPlanner/blob/master/ExtLibs/Utilities/srtm.cs

Thanks!

Davidsastresas:
Small update, I pushed some commits. It was implemented a mechanism to check if map providers need to uncompress tiles, and I am in the step of testing uncompressing.

It seems the cache system is getting the tiles and storing them correctly, and the calls to AP elevation server seems to be done correctly.

The QByteArray method doesn't work for uncompressing, and even if it does it is not good to use it, it could break functionality at any moment, we need to use a proper zip library there. Zlib is already implemented in QGC, but not for all platforms. We need to investigate that...

Davidsastresas:
Thanks to @DonLakeFlyer for the tips about uncompressing in QGC, I realized we have zlib available for all platforms in QGC. I gave it a try but unfortunately zlib expects the files in zlib format, which apparently is different from .gzip or .zip. I wasn't aware of this, I am not very familiar with compression in general.

Next time I spare time for this I will investigate more possibilities for this decompresion...

cessna1525:
any updates on this?

DonLakeFlyer:
@Davidsastresas This can be closed now right?

Davidsastresas:
@DonLakeFlyer yes I think so. At some point I would like to work on this again, but I don't know when I will get back to it so we can close it for now. Thanks.

HTRamsey:
@Davidsastresas is there still value in this? I can take a look at it since I have been working on the location code.

DonLakeFlyer:
I don't see it. But maybe someone has a better reason as to why multiple terrain server support would be important.

Davidsastresas:
I really think it is! We are now using Auterion servers, and I don't think we have a lot of guarantee that they will commit to maintain them forever. So for me it is definitely a go.

I wanted to jump back to this at some point, so it would be amazing to have your help here. Thank you!

DonLakeFlyer:
> and I don't think we have a lot of guarantee that they will commit to maintain them forever.

At which point it would make certainly sense to do this. Time would be better spent helping to finish up the ancient pull which allows the import of custom geotiff elevation data. Lot's of folks ask for that. That would be a great new addition to QGC that adds a lot of value.

Davidsastresas:
I agree. However most of the work done on this PR was towards supporting several providers, it wasn't hardcoding it to AP terrain servers.

So I would say priority one would be to de-couple the current Auterion servers implementation to allow several providers, and then work on those geotiffs, and maybe then add AP terrain servers.

DonLakeFlyer:
Sorry, I wasn't paying attention and thought this was just an Issue discussion as opposed to an existing Pull. If it's mostly done then it can't hurt.

PR #10716, opened 2023-06-19
Add airspeed guided action support for Ardupilot fixed wing, fix speed change slider

Davidsastresas:
This PR adds support for Arduplane to change airspeed. The slider will show ONLY if we are in fixed wing firmware. Under actions tab, a new option available "set airspeed" will be available, and after selecting, we will have a slider to select the desired airspeed. The slider will be automatically scaled using ARSPD_FBWA_MAX and MIN parameters. This is sending a MAV_CMD_DO_CHANGE_SPEED, with:

- param 1 = 0 ( airspeed )
- param 2 = airspeed setpoint
- param 3 = -1 (no throttle change)
- param 4 = 0 (absolute speed, not incremental)
- rest of parameters to 0.

In a different commit I also modified the guided slider so it shows "m/s" instead of "m" when we are dealing with speeds. This is hardcoded and it is not getting system unit settings, this is done on purpose as we don't have implemented unit conversion here, it is always meters/s. 

I also modified the conditions in guidedactionscontroller to show this slider. It used to show the airspeed slider only if Vtol in forward flight mode, and the rest of the time will show groundspeed.

Now it will show airspeed if vtol in forward flight or fixed wing ( before this the fixed wing case was not covered ) and it will show groundspeed only if not fixedwing and we have multirrotor speed limits on firmware.

The latest changes are fine with how we are dealing with this in AP, but I am not sure if this is appropiate for PX4. My feeling is that it is, but I would appreciate another look here from anybody familiar with PX4. @mrpollo  can you get somebody to double check this? @RomanBapst was the one adding initial support for this in PX4, maybe he could take a look as well. Thanks!

### DIFF
--- a/src/FlightDisplay/GuidedActionsController.qml
+++ b/src/FlightDisplay/GuidedActionsController.qml
@@ -190,22 +190,24 @@ Item {
     function setupSlider(actionCode) {
         // generic defaults
         guidedValueSlider.configureAsLinearSlider()
+        guidedValueSlider.setIsSpeedSlider(false)
 
         if (actionCode === actionTakeoff) {
                 guidedValueSlider.setMinVal(_activeVehicle.minimumTakeoffAltitude())
                 guidedValueSlider.setValue(_activeVehicle ? _activeVehicle.minimumTakeoffAltitude() : 0)
                 guidedValueSlider.setDisplayText("Height")
         } else if (actionCode === actionChangeSpeed) {
-            if (_activeVehicle.vtolInFwdFlight) {
+            guidedValueSlider.setIsSpeedSlider(true)
+            if (_fixedWing) {
                 guidedValueSlider.setDisplayText("Set Airspeed")
-                guidedValueSlider.setMinVal(_activeVehicle.minimumEquivalentAirspeed())
-                guidedValueSlider.setMaxVal(_activeVehicle.maximumEquivalentAirspeed())
-                guidedValueSlider.setValue(_activeVehicle.airSpeedSetpoint.rawValue)
-            } else {
+                guidedValueSlider.setMinVal(QGroundControl.unitsConversion.metersSecondToAppSettingsSpeedUnits(_activeVehicle.minimumEquivalentAirspeed()).toFixed(1))
+                guidedValueSlider.setMaxVal(QGroundControl.unitsConversion.metersSecondToAppSettingsSpeedUnits(_activeVehicle.maximumEquivalentAirspeed()).toFixed(1))
+                guidedValueSlider.setValue(_activeVehicle.airSpeed.value)
+            } else if (!_fixedWing && _activeVehicle.haveMRSpeedLimits) {
                 guidedValueSlider.setDisplayText("Set Speed")
-                guidedValueSlider.setMinVal(0.1)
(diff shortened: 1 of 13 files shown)


mrpollo:
Thanks for the great contribution, as always @Davidsastresas

Hey @tstastny, @bkueng any feedback on the above ☝️ ?

src/FlightDisplay/GuidedActionsController.qml
```diff
@@ -180,19 +180,22 @@ Item {
...
         guidedValueSlider.configureAsLinearSlider()
+        guidedValueSlider.setIsSpeedSlider(false)
 
         if (actionCode === actionTakeoff) {
                 guidedValueSlider.setMinVal(_activeVehicle.minimumTakeoffAltitude())
                 guidedValueSlider.setValue(_activeVehicle ? _activeVehicle.minimumTakeoffAltitude() : 0)
                 guidedValueSlider.setDisplayText("Height")
         } else if (actionCode === actionChangeSpeed) {
-            if (_activeVehicle.vtolInFwdFlight) {
+            if (_activeVehicle.vtolInFwdFlight || _activeVehicle.fixedWing) {
```
bkueng, line 190: You could use `_fixedWing` here as well

bkueng (approved):
Makes sense to me

RomanBapst:
@Davidsastresas 

`About the issue for the initial value reported by Arace, it is because on the previous implementation it was using airspeed setpoint instead of airspeed. This setpoint was varying a lot during flight, it was extremely unstable changing +- 10 m/s in a matter of seconds, so the initial value for the slider wasn't making sense. I think it makes sense to use airspeed here, and not airspeed setpoint, as the user probably wants to set airspeed based on the current airspeed reading.`

This doesn't make sense to me. When I did the initial implementation if deliberately used the airspeed setpoints as it's not changing erratically as measured airspeed. Given your description above I assume that something is wrong in the calculation of the airspeed setpoint. Can you please check that? I would still prefer the slider to init with the current airspeed setpoint TECS is trying to achieve.

Davidsastresas:
@RomanBapst thanks for the answer.

That makes sense, thanks for clarifying. Maybe that setpoint calculation is what is wrong here. I will take a look at it at some point.

About PX4, I have a setup in my environment, and I test it usually, but I am not as familiarized with PX4 as I am with Ardupilot. In ardupilot I can trace real quick in the code how a particular mavlink message affects all the vehicles, or how a particular telemetry field is generated. In PX4 I do not, so even if I can test it and confirm it works, I still prefer confirmation from anybody really familiar with it, to be 100% we don't break anything. Thanks for the suggestion!

DonLakeFlyer:
Close and reopen to get CI to run again

DonLakeFlyer:
@Davidsastresas is this ready to merge?

Davidsastresas:
@DonLakeFlyer I rebased to current master, tested in Ardupilot, and force pushed. My only concern was about PX4 being fine with it, but as @bkueng confirmed back at the beginning it should be fine.

About setpoint or not discussed with @RomanBapst , I don't have a lot of time to fix the setpoint calculation and test all of this again. As it is now it has been running in production in a custom build for some months now, Flying Ardupilot VTOLs.

So if we are okay with using the actual airspeed and not the airspeed setpoint this should be ready to merge now.

PR #10719, opened 2023-06-23
Vagrantfile: move to use Ubuntu Jammy for Vagrant builds

peterbarker:


Vagrantfile
```diff
@@ -11,7 +11,6 @@ env_global = [
   'JOBS=4',
   'SHADOW_BUILD_DIR=/tmp/shadow_build_dir',
   'CODESIGN=nocodesign',
-  'secure: RGovyUnMw3fp/bHZi058JvANT1rYmNqrsuSYew0cIgirO6YbMHr/rsjwCm1FTYpBl8s1zgr+u2b8ftYnfnCz2YT+Aip4NWrVYpVU0FEmfytGILrnUS0pjlt8m7fU9AKR1ElOSll7yw7e1kftynN39Q321etvwbLZcXon6zz0suE='
```
booo, line 14: Should this be part of the commit? If so what implications has it. I never checked the Vagrantfiles

peterbarker: This seems to be a syntax error to be - not the : vs = - doesn't build with it there, does everything it needs to without AFAICS.

peterbarker:
On Sat, 24 Jun 2023, Philipp Borgers wrote:

> @booo commented on this pull request.
> 
> _______________________________________________________________________________________________________________________________________________________________________________________________________________
> 
> In Vagrantfile:
> 
> > @@ -11,7 +11,6 @@ env_global = [
>    'JOBS=4',
>    'SHADOW_BUILD_DIR=/tmp/shadow_build_dir',
>    'CODESIGN=nocodesign',
> -  'secure: RGovyUnMw3fp/bHZi058JvANT1rYmNqrsuSYew0cIgirO6YbMHr/rsjwCm1FTYpBl8s1zgr+u2b8ftYnfnCz2YT+Aip4NWrVYpVU0FEmfytGILrnUS0pjlt8m7fU9AKR1ElOSll7yw7e1kftynN39Q321etvwbLZcXon6zz0suE='
> 
> Should this be part of the commit? If so what implications has it. I never checked the Vagrantfiles

Well... the PR definitely doesn't work without it.

The line was added to the Vagrantfile with a commit message which doesn't 
explain *why* it was added to the Vagrantfile ("Remove travis-related CI 
configuration") where Vagrant and Travis are unrelated.

peterbarker:
I've now modified the Vagrantfile to increase the maximum number of files open at one time.  The QGC link step seems to have stepped over the 1024-open-files boundary which is the default in Ubuntu Jammy.

This Vagrantfile doesn't work in master right now either due to the syntax error - which I can only assume was merged untested.

booo:
Merged. Thanks for your contribution.

As I mentioned in another ticket about docker it would be nice to have CI steps for these builds too. Otherwise we don't spot problems in the build process based e.g. on Vagrant. Is this something you could contribute?

peterbarker:
On Mon, 10 Jul 2023, Philipp Borgers wrote:

> As I mentioned in another ticket about docker it would be nice to have CI steps for these builds too. Otherwise we don't spot problems in the build process based e.g. on Vagrant. Is this something you could
> contribute?

I don't think it will work end-to-end; bringing up VMs within CI is 
probably not going to end well.

Could potentially add a CI actuion to at least check the syntax is correct 
in the configuration file, however.

PR #10726, opened 2023-07-06
Remove AirMap support

DonLakeFlyer:
Airmap platform SDK is no longer supported by Airmap. Airmap recently removed a library repro which QGC depends on. This in turns causes all QGC builds to fail. So the removal of all Airmap code needs to go in asap so builds will work again.


booo:
Thanks for your work. 

I always wondered why we maintain software that is connected to an external entity at all.

PR #10732, opened 2023-07-13
Stable 4.2: Fix problem with xbox series x joystick

patrickelectric:


patrickelectric:
Validated by users testing under windows with Xbox series x and one controller though BT and cable.

PR #10738, opened 2023-07-19
Fix incorrect group name

DonLakeFlyer:


DonLakeFlyer:
Nope. This can't be corrected without losing end users current settings.

HTRamsey:
@DonLakeFlyer I think we should consider doing this for v5. A bit unfortunate to leave a bug forever, and a major new release is the best time to do something that affects end users

DonLakeFlyer:
I'll fix this up

PR #10740, opened 2023-07-20
feat: replace Airmap terrain data with Copernicus

leonardosimovic:
### The Problem
The Airmap terrain servers (and possibly other Airmap services) are shutting down. The docs  [https://developers.airmap.com/docs/elevation-api](https://developers.airmap.com/docs/elevation-api) are already down while the servers seem still to be running for the time being.

### The Solution
This PR replaces the `Airmap` query with an alternative provider, hosting the [Copernicus GLO-30](https://spacedata.copernicus.eu/collections/copernicus-digital-elevation-model) which was compiled and provided by `© Airbus Defence and Space GmbH` dataset. The Copernicus dataset has generally better global coverage and accuracy than the previous SRTM dataset. 

### Followup or Incremental Work

- [x] Remove `Airmap` references in the code.
- [x] Give credit to `© Airbus Defence and Space GmbH` also in the UI.

### Elevation Provider Credit

Added a small note in the bottom center of terrain plot:

image

#### Latitude zone I: 0-50 degrees north and south

| Airmap |
:-------------------------:
latitude-zone-1_1-airmap

| Copernicus |
:-------------------------:
latitude-zone-1_1-copernicus

#### Latitude zone II: 50-70 degrees north and south:
| Airmap |
:-------------------------:
latitude-zone-2-airmap

| Copernicus |
:-------------------------:
latitude-zone-2-copernicus

#### Latitude zone III:  70-75 degrees north and south:
| Airmap |
:-------------------------:
latitude-zone-3-airmap Note: the SRTM dataset only covers land surface between 60° north and 56° south latitude

| Copernicus |
:-------------------------:
latitude-zone-3-copernicus

#### Latitude zone IV: 75-80 degrees north and south:
| Airmap |
:-------------------------:
latitude-zone-4-airmap Note: the SRTM dataset only covers land surface between 60° north and 56° south latitude

| Copernicus |
:-------------------------:
latitude-zone-4-copernicus

#### Latitude zone V: 80-90 degrees north and south:
| Airmap |
:-------------------------:
latitude-zone-5-airmap
 Note: the SRTM dataset only covers land surface between 60° north and 56° south latitude

| Copernicus |
:-------------------------:
latitude-zone-5-copernicus

### Raw Images and Mission Plans
[qgc-airmap-replacement.zip](https://github.com/mavlink/qgroundcontrol/files/12117119/qgc-airmap-replacement.zip)

### DIFF
--- a/src/QtLocationPlugin/QGCMapEngine.cpp
+++ b/src/QtLocationPlugin/QGCMapEngine.cpp
@@ -200,7 +200,7 @@ QGCMapEngine::addTask(QGCMapTask* task)
 
 //-----------------------------------------------------------------------------
 void
-QGCMapEngine::cacheTile(QString type, int x, int y, int z, const QByteArray& image, const QString &format, qulonglong set)
+QGCMapEngine::cacheTile(const QString& type, int x, int y, int z, const QByteArray& image, const QString &format, qulonglong set)
 {
     QString hash = getTileHash(type, x, y, z);
     cacheTile(type, hash, image, format, set);
@@ -208,7 +208,7 @@ QGCMapEngine::cacheTile(QString type, int x, int y, int z, const QByteArray& ima
 
 //-----------------------------------------------------------------------------
 void
-QGCMapEngine::cacheTile(QString type, const QString& hash, const QByteArray& image, const QString& format, qulonglong set)
+QGCMapEngine::cacheTile(const QString& type, const QString& hash, const QByteArray& image, const QString& format, qulonglong set)
 {
     AppSettings* appSettings = qgcApp()->toolbox()->settingsManager()->appSettings();
     //-- If we are allowed to persist data, save tile to cache
@@ -220,7 +220,7 @@ QGCMapEngine::cacheTile(QString type, const QString& hash, const QByteArray& ima
 
 //-----------------------------------------------------------------------------
 QString
-QGCMapEngine::getTileHash(QString type, int x, int y, int z)
+QGCMapEngine::getTileHash(const QString& type, int x, int y, int z)
 {
     return QString::asprintf("%010d%08d%08d%03d", getQGCMapEngine()->urlFactory()->getIdFromType(type), x, y, z);
 }
@@ -235,7 +235,7 @@ QGCMapEngine::hashToType(const QString& hash)
 
 //-----------------------------------------------------------------------------
 	QGCFetchTileTask*
-QGCMapEngine::createFetchTileTask(QString type, int x, int y, int z)
+QGCMapEngine::createFetchTileTask(const QString& type, int x, int y, int z)
 {
 	QString hash = getTileHash(type, x, y, z);
 	QGCFetchTileTask* task = new QGCFetchTileTask(hash);
@@ -244,7 +244,7 @@ QGCMapEngine::createFetchTileTask(QString type, int x, int y, int z)
 
 //-----------------------------------------------------------------------------
 	QGCTileSet
-QGCMapEngine::getTileCount(int zoom, double topleftLon, double topleftLat, double bottomRightLon, double bottomRightLat, QString mapType)
+QGCMapEngine::getTileCount(int zoom, double topleftLon, double topleftLat, double bottomRightLon, double bottomRightLat, const QString& mapType)
 {
 	if(zoom <  1) zoom = 1;
 	if(zoom > MAX_MAP_ZOOM) zoom = MAX_MAP_ZOOM;
@@ -368,7 +368,7 @@ QGCMapEngine::_pruned()
 
 //-----------------------------------------------------------------------------
 int
-QGCMapEngine::concurrentDownloads(QString type)
+QGCMapEngine::concurrentDownloads(const QString& type)
 {
     Q_UNUSED(type);
     // TODO : We may want different values depending on
--- a/src/TerrainTile.cc
+++ b/src/TerrainTile.cc
@@ -31,140 +31,120 @@ const char*  TerrainTile::_jsonMinElevationKey  = "min";
 const char*  TerrainTile::_jsonAvgElevationKey  = "avg";
 const char*  TerrainTile::_jsonCarpetKey        = "carpet";
 
-TerrainTile::TerrainTile()
-    : _minElevation(-1.0)
-    , _maxElevation(-1.0)
-    , _avgElevation(-1.0)
-    , _data(nullptr)
-    , _gridSizeLat(-1)
-    , _gridSizeLon(-1)
-    , _isValid(false)
+TerrainTile::TerrainTile(const QByteArray& byteArray)
 {
+    // Copy tile info
+    _tileInfo = *reinterpret_cast<const TileInfo_t*>(byteArray.constData());
 
-}
-
-TerrainTile::~TerrainTile()
-{
-    if (_data) {
-        for (int i = 0; i < _gridSizeLat; i++) {
-            delete[] _data[i];
-        }
-        delete[] _data;
-        _data = nullptr;
+    // Check feasibility
+    if ((_tileInfo.neLon - _tileInfo.swLon) < 0.0 || (_tileInfo.neLat - _tileInfo.swLat) < 0.0) {
+        qCWarning(TerrainTileLog) << this << "Tile extent is infeasible";
+        _isValid = false;
+        return;
     }
-}
 
-TerrainTile::TerrainTile(QByteArray byteArray)
-    : _minElevation(-1.0)
-    , _maxElevation(-1.0)
-    , _avgElevation(-1.0)
-    , _data(nullptr)
-    , _gridSizeLat(-1)
-    , _gridSizeLon(-1)
-    , _isValid(false)
-{
+    _cellSizeLat = (_tileInfo.neLat - _tileInfo.swLat) / _tileInfo.gridSizeLat;
+    _cellSizeLon = (_tileInfo.neLon - _tileInfo.swLon) / _tileInfo.gridSizeLon;
+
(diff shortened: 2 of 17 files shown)


src/QtLocationPlugin/QGCMapEngine.cpp
```diff
@@ -35,7 +35,7 @@ Q_DECLARE_METATYPE(QList<QGCTile*>)
 static const char* kDbFileName = "qgcMapCache.db";
 static QLocale kLocale;
 
-#define CACHE_PATH_VERSION  "300"
```
leonardosimovic, line 38: Previously cache terrain tiles won't be compatible with new ones

src/TerrainTile.cc
```diff
@@ -31,140 +31,128 @@ const char*  TerrainTile::_jsonMinElevationKey  = "min";
...
-        double latFraction          = (clampedLat - latIndexLatitude) / tileValueSpacingDegrees;
-
-        // Calc the elevation as the average across the four known points
-        double known00      = _data[latIndex][lonIndex];
-        double known01      = _data[latIndex][lonIndex+1];
-        double known10      = _data[latIndex+1][lonIndex];
-        double known11      = _data[latIndex+1][lonIndex+1];
-        double lonValue1    = known00 + ((known01 - known00) * lonFraction);
-        double lonValue2    = known10 + ((known11 - known10) * lonFraction);
-        double latValue     = lonValue1 + ((lonValue2 - lonValue1) * latFraction);
```
leonardosimovic, line 141: Dropped the bi-linear interpolation since there is no real benefit in doing so. Simple interpolation methods are not suited for terrain data. More advanced interpolation algorithms based on statistical methods, e.g. Kriging, would be ideal. Nearest neighbor and linear interpolation yield about the same error on average.

Davidsastresas:
Thanks @leonardosimovic 

I took a quick look, and regarding the implementation I would only change all the references to Airmap to whatever we want to call this provider.

Regarding the bi-linear approximation or not, what is the gridSize of your tiles right now? The graphs on your comparisons look great, so probably we are good, but I think it would be good to specify somewhere the gridSize of your tiles so we can understand the error we are assuming not doing interpolation, or doing it.

I tested real quick your branch and it works great for me, but I am getting some errors when I create a polygon and then I move it. When creating a survey polygon everything is smooth, I only get the errors when click and drag vertices:

TerrainQueryLog: TerrainTileManager::getAltitudesForCoordinates Internal Error: missing elevation in tile cache
TerrainQueryLog: addPathQuery: signalling failure due to internal error
TerrainTileLog: 0x5580d4118738 Internal error: coordinate outside tile bounds: QGeoCoordinate(40.479999573, -3.3279448085)

I don't think I will be able to spend much more time this week on this matter, so if you could investigate the above it would be great.

Thank you very much!

leonardosimovic:
Hey @Davidsastresas  

Thank you for the feedback!

> Not 100% sure right now, but doing this also we can probably avoid bumping cache version right

I would also prefer not upgrading the cache. However, it is necessary because the new data has a different format and therefore is not compatible. Previously, with Airmap, it was assumed that each carpet query would return a fixed array size `n_lat x n_lon`, regardless of the geolocation, and this has worked well. Our service accounts for converging longitudinal lines as we go to higher latitudes, hence the array size will vary in longitudinal direction.

> Regarding the bi-linear approximation or not, what is the gridSize of your tiles right now?

It has the same resolution as `Airmap` i.e. 30 meters, hence I did not change the static variables or the line sampling method.

>  I tested real quick your branch and it works great for me, but I am getting some errors when I create a polygon and then I move it. When creating a survey polygon everything is smooth, I only get the errors when click and drag vertices:
 
I will have a look!

Davidsastresas:
Great to hear you can also reproduce those warnings!

Regarding Cache, I am not 100% sure of this, but isn't the cache system based on the hash for each tile, and isn't this hash dependant on the name? So, changing the name of the provider, shouldn't the tiles have a different hash and thus not having need to bump cache version? I am not 100% sure though.

Thanks!

leonardosimovic:
Hey @Davidsastresas 

I looked into the warnings some more and traced it back to rounding errors in the `bounds` fields for the request:
 [https://terrain-ce.suite.auterion.com/api/v1/carpet?points=40.42,-3.23,40.43,-3.22](https://terrain-ce.suite.auterion.com/api/v1/carpet?points=40.42,-3.23,40.43,-3.22).
We will soon deploy an updated service, I will include a fix for this issue.

> Regarding Cache, I am not 100% sure of this, but isn't the cache system based on the hash for each tile, and isn't this hash dependant on the name? So, changing the name of the provider, shouldn't the tiles have a different hash and thus not having need to bump cache version? I am not 100% sure though.

This sounds right, but I'm not 100% sure either. When I tested it the cache needed to be bumped otherwise still the perviously cached tiles were loaded 🤔

leonardosimovic:
@Davidsastresas 

I pushed a couple of commits, addressing the Airmap hard-coding + some unrelated performance improvements. I tried to keep the changes to a minimum in order to reduce conflicts with [your PR](https://github.com/mavlink/qgroundcontrol/pull/10715/files#diff-22fec0cfecc787341aa00ccca1a75111edc65b6737b46db472c0fcd754a81ce3). Let me know what you think!

Also, I will let you know as soon as we have deployed our updated service.

DonLakeFlyer:
Thanks so, so much for this work. I took a quick look through and it looks fine. But at this point @Davidsastresas knows way more about this than I do since he was recently in this code.

My one complaint is in the license string being part of the terrain status display. That takes up really (really) critical vertical real estate on small screens, like small tablets or say a Herelink. Can that go someplace else? Couple options:

Where the QGC version number is:
Screen Shot 2023-08-05 at 1 06 08 PM

The help screen in Application settings:
Screen Shot 2023-08-05 at 1 06 38 PM

leonardosimovic:
Hey @DonLakeFlyer! No problem, we're happy to help :)  

> My one complaint is in the license string being part of the terrain status display. That takes up really (really) critical vertical real estate on small screens, like small tablets or say a Herelink. Can that go someplace else?

That's a valid point! My only concern is with the visibility of the copyright notice if we place it in the menus. Let me check with the team and I'll come back to you.

DonLakeFlyer:
Is there some legal thing with respect to where it needs to be?

DonLakeFlyer:
The other options is show it in terrain status for say a few seconds and then hide it.

Davidsastresas:
Hi!, I think I will be able to check this tomorrow. Thanks for the work!

About the label, I agree with Don, it is really not good there, in devices like Herelink screen real state is precious. The idea about showing it for a few seconds I think is a good compromise. Something like 5 seconds will really be noticed. On the draft PR we were making for supporting several elevation providers we will have a dedicated option in general settings for it. It could be another nice place, in a way that the copyright only shows if we select this provider.

Also, It would be good to somehow link that license to the elevationmap provider, instead of being in QGCUrlmapengine as a global variable. Just thinking on the near future when we support more elevation providers. We can manage that later as well, I don't think that should hold back the PR.

Thanks!

leonardosimovic:
Hey everyone, sorry for the delaye!

> The other options is show it in terrain status for say a few seconds and then hide it.

I'm not sure about hiding the label after x seconds from a UX point of view. I don't think the label is super noticeable as it is now, but it's disappearing will cause the plot to resize / elements to shift, this will be noticed for sure by which time the label is already gone. 

Other suggestion would be to move outside and on top of the plot. This won't increase terrain plot background size while the label remains visible.

> On the draft PR we were making for supporting several elevation providers we will have a dedicated option in general settings for it. It could be another nice place, in a way that the copyright only shows if we select this provider.

This would be a very nice solution!

> I just spared some time to check this out. It looks good, but that error when creating a polygon and moving it around is still there

Yeah we haven't yet deployed the fix to the terrain-service. I will let you know as soon as it's in.

> Regarding the 3rd commit, when you declare that kOnlineElevationProviderKey, I think we could do something there that leaves it nicer for future support for more elevation map providers. It could be added as a setting, so the name of the map provider is available globally by QGroundControl settings.

That's definitely the way to go and you already implemented in your draft I believe. Hence I went with the bare minimum so it's a) less cumbersome to rebase for you and b) we can safe time by avoid duplicate work.

DonLakeFlyer:
> Other suggestion would be to move outside and on top of the plot.

I like that much better. If it's just text you can still see the map behind it and make sure it in the zorder below the plan items on the map.

leonardosimovic:
Hey everyone, I cleaned up the commit history and moved the label ontop of the terrain plot:

image

It's not the prettiest solution but I hope it's acceptable for the time being, until there's a dedicated terrain settings section which @Davidsastresas proposed.

DonLakeFlyer:
> It's not the prettiest solution but I hope it's acceptable for the time being

I think it's good enough for now. @Davidsastresas Want to give one final look see an then you can merge.

Davidsastresas:
@DonLakeFlyer as said on my previous comment, if the rest of you are fine with this I am fine myself too, so I am happy with this getting merged as well!

leonardosimovic:
Awesome, thank you all for the reviews!

@Davidsastresas we have redeployed our terrain service, so you shouldn't see the error any more after clearing the map cache 😄

Ryanf55:
> This PR replaces the `Airmap` query with an alternative provider, hosting the [Copernicus GLO-30](https://spacedata.copernicus.eu/collections/copernicus-digital-elevation-model) which was compiled and provided by `© Airbus Defence and Space GmbH` dataset. The Copernicus dataset has generally better global coverage and accuracy than the previous SRTM dataset.


The link to the dataset is dead. For anyone curious, a new link is here: https://portal.opentopography.org/raster?opentopoID=OTSDEM.032021.4326.3

I've received reports that this dataset performs poorly in mountainous areas in Canada.

PR #10742, opened 2023-07-20
Airlink

Uzopb:
Hi. I am adding the ability to connect ground station with Qgroundcontrol software through [AirLink lte module](https://ardupilot.org/plane/docs/common-airlink-telemetry.html). Adding changes to master will help speed up the work. But I also had difficulties after switching to the master branch in the libs/mavlink submodule, so I had to make comments on those parts of the code that broke the build.


Davidsastresas:
Hello @Uzopb,

it is great work, but I think for this to get into master it would need to go in a separate settings menu, and separate class, not over linkManager. I think those are generic menus-classes, and something like this belongs to a separate module.

Please take a look at how this is managed for taisync or microhard. I think something like that is much more appropiate if you mean to push this to master.

Regarding the second commit, It looks like you don't have the submodules up to date. usually a git submodule update --init --recursive would do it. If you were previously working in a branch with a different submodule remote, you can run a git submodule sync command before the submodule update, that should fix it.

Thanks!

Uzopb:
Hi @Davidsastresas 
Thanks for the review, for working through airlink modules, it's really better to follow the example with taisync.

About the second commit and the mavlink submodule.
To get the repository, I used the following commands from the qgroundcontrol developer's guide:
- git clone --recursive -j8 https://github.com/mavlink/qgroundcontrol.git
- git submodule update --recursive

I also tried to do this:
- git clone https://github.com/mavlink/qgroundcontrol.git
- git submodule update --init --recursive

But as a result, there are no airlink files in the mavlink submodule.
To get them, I did a **git pull** (or **checkout master**).
After that I have the airlink files in the mavlink submodule, but
the build qgroundcontrol breaks. 
Could you check it out?

Davidsastresas:
Okey, I understand now sorry. You needed mavlink messages that are not defined in master QGC, correct?

If so, QGC uses the repo https://github.com/mavlink/c_library_v2 which is just the mavlink headers defined in Ardupilotmega.xml compiled an included as submodule. Your new messges must be on that c_library_v2 so we can pick them in QGC. When a commit is done that affects any of the submodules, it is good to include in the commit what tag or hash you are checking out in the submodules, and why ( The why is good to put it leaving a white line, and then a explanation, so the long description of the commit only shows when expanding the commit tittle ).

So somethink like: 

"libs/mavlink: checking out xxxxxxxxx to include xxxx messages:

Integrating airlink we didn't have this and that messages etc etc"

About the build breaking, maybe with that mavlink update something needs to get adjusted in the code. Sometimes some new fields are added to some mavlink messages and it needs to be fixed all around in QGC code. I can go through it, but I am not sure I will be able to spend time on this on the following weeks, so if you can go over it it would be great.

Thanks!

PR #10744, opened 2023-07-24
Correct typos in korean translation

NIA1995:
Correct typos in korean translation.

fixed QString::arg: Argument missing: "항목 1" , 0 error.


DonLakeFlyer:
Was this just changed here or was it changed in Crowdin? It needs to be done through Crowdin otherwise this will just be overwritten the next time someone updates translations.

DonLakeFlyer:
Closing since I'm unsure if this was done correctly

PR #10748, opened 2023-07-29
Allow app data to save to sd card on android

DonLakeFlyer:
* Add setting to General/Misc in the UI to save all application data to the SD card for android builds.
* If option is enabled and no SD card available or write protected user is notified about the problem and that internal storage instead will be used.

This was added to move forward the Herelink version. The current Herelink variant saves video to the SD card. Nice feature, but in reality it makes sense to allow for all data to be on the SD Card which is much easier to work with than the internal storage. Adding this upstream change then allows me to carry it over to the new Herelink version.


mrpollo (approved):
lgtm!

PR #10752, opened 2023-08-02
Fix a few issues with NMEA GPS for GCS position

JMare:
While working on getting Remote ID working with Master QGC I encountered a few small problems in using an NMEA GPS.
I am using a Toughbook FZ-G2 running windows 10, which has an internal GPS module which presents as a NMEA stream on a COM port (COM5 in my case).

# First problem - Serial port list mismatch:
While I was able to select COM5 in the drop down, and I got a lat lon to show up in the Remote ID Page, it would never show altitude. In digging into the issue, I discovered what was happening.
The dropdown menu in the UI was pulling from LinkManager::serialPorts(), which was pulling from QSerialPortInfo::availablePorts(). However, in the LinkManager::_updateAutoConnectLinks(), where the NMEA port is passed to the position manager, it would iterate through a smaller list provided by QGCSerialPortInfo::availablePorts(), which is a override which provides a filtered version of the availablePorts list.
**This drills down to two problems:**
a) the list of serial ports in the UI is not the same as the list which are actually possible to use
b) the override for availablePorts is filtering out my internal NMEA GPS port.

**My Solutions**
for a) This is a simple fix to make the data passed to the UI call the same override, so that the two lists match.
for b) I'm not sure the best solution, I added a check that doesn't skip any links which have the word NMEA in the description.

# Second Problem - latched values on changing position source:
Once I fixed the above so I could select and use the NMEA GPS, I found the following issue:
If I booted up QGC when the GPS didn't have a fix, it would show a fixed GPS position, even though the NMEA source had no fix.
**The cause:**
When QGC Starts, it first selects the default/internal position source. On my device, that provides a position update with the last position the GPS had. Then after a few seconds, the _updateAutoConnectLinks() passes the real NMEA port to PositionManager, which switches to the NMEA GPS. However, since the NMEA GPS has an invalid fix, the values in the UI are latched at the values provided by the internal GPS source. This is really confusing, and makes it look like the GPS has a fix when it doesnt.
**The solution:**
I fixed this by resetting all the stored position values when setPositionSource is called.

### DIFF
--- a/src/PositionManager/PositionManager.cpp
+++ b/src/PositionManager/PositionManager.cpp
@@ -153,6 +153,17 @@ void QGCPositionManager::setPositionSource(QGCPositionManager::QGCPositionSource
     if (_currentSource != nullptr) {
         _currentSource->stopUpdates();
         disconnect(_currentSource);
+
+        // Reset all values so we dont get stuck on old values
+        _geoPositionInfo = QGeoPositionInfo();
+        _gcsPosition = QGeoCoordinate();
+        _gcsHeading =       qQNaN();
+        _gcsPositionHorizontalAccuracy = std::numeric_limits<qreal>::infinity();
+
+        emit gcsPositionChanged(_gcsPosition);
+        emit gcsHeadingChanged(_gcsHeading);
+        emit positionInfoUpdated(_geoPositionInfo);
+        emit gcsPositionHorizontalAccuracyChanged();
     }
 
     if (qgcApp()->runningUnitTests()) {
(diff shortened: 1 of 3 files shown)


bkueng (approved):
Makes sense.

> for b) I'm not sure the best solution, I added a check that doesn't skip any links which have the word NMEA in the description.

Would be nice to identify mavlink composite devices a bit better in the first place, but problably not much we can do now.

JMare:
> Makes sense.
> 
> > for b) I'm not sure the best solution, I added a check that doesn't skip any links which have the word NMEA in the description.
> 
> Would be nice to identify mavlink composite devices a bit better in the first place, but problably not much we can do now.

Agreed. I know on a pixhawk 2 the MAVLink port has "mavlink" in the description. If that were true across other devices that would be pretty easy, but I'm guessing its not that simple. Otherwise a list of VID/PID devices but that would be painful.
Obviously the current solution is mostly working and hopefully this change will fix for a few people.

Thanks for review and merge

PR #10759, opened 2023-08-04
[Bugfix] Fix Flight Mode Switch settings box not showing

junwoo091400:
## About
This was introduced in
https://github.com/mavlink/qgroundcontrol/commit/d5dc4f1af17151d7e7f88372847bc12ac499ccb9, and the refactor didn't fix the case for RC_MAP_FLAPS case correctly

Fixes https://github.com/PX4/PX4-Autopilot/issues/21927

## Note
I haven't tested it, but with the built artifact it could be great if you could give it a test @benjinne

### DIFF
--- a/src/AutoPilotPlugins/PX4/PX4SimpleFlightModes.qml
+++ b/src/AutoPilotPlugins/PX4/PX4SimpleFlightModes.qml
@@ -34,8 +34,8 @@ Item {
             _switchTHList.push("RC_TRANS_TH")
         }
         if (controller.vehicle.fixedWing) {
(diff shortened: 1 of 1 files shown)


junwoo091400:
Test update
Before (from mentioned github issue):
image

The switch page is now normal:
image

It would be awesome to transition to Betaflight-style configuration with setting thresholds, etc correctly in the future, as described here: https://github.com/PX4/PX4-Autopilot/pull/19638#issuecomment-1230457262

benjinne:
I can confirm this fixes my issue

image

DonLakeFlyer:
Yup, this is a correct fix. Thanks.

PR #10762, opened 2023-08-04
Android fixes

DonLakeFlyer:
There were all discovered through my Herelink work but are basic android build problems. The commit text describes the changes in detail.

### DIFF
--- a/src/AnalyzeView/LogDownloadPage.qml
+++ b/src/AnalyzeView/LogDownloadPage.qml
@@ -84,7 +84,7 @@ AnalyzePage {
                                     if(d.getUTCFullYear() < 2010)
                                         return qsTr("Date Unknown")
                                     else
-                                        return d.toLocaleString()
+                                        return d.toLocaleString(undefined, "short")
                                 }
                             }
                             return ""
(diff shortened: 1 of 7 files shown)


Davidsastresas:
Thank you very much Don!

I checked, everything looks good to me. Just one small note:

- What you did to long press the version number to show touch Areas in MainRootWindow.qml, I know it is extremely unlikely that a user could do that accidentally, but maybe we can let user know when this happens/when it is happening? Nothing fancy, something like a popup or something indicating this action was executed, so the user understands and can undo that if desired.

This feature of showing the touch areas in read is pretty cool by the way, very handy!

The rest of the fixes are nice I think, and they will improve it in small screens while not bothering at all in the rest of platforms, so i am good with all of it.

Thank you!

DonLakeFlyer:
> * What you did to long press the version number to show touch Areas in MainRootWindow.qml, I know it is extremely unlikely that a user could do that accidentally, but maybe we can let user know when this happens/when it is happening? Nothing fancy, something like a popup or something indicating this action was executed, so the user understands and can undo that if desired.

Sure I'll add that here. And then I'll merge it.

Davidsastresas (approved):

PR #10763, opened 2023-08-05
Android support for serial/bluetooth disabling

DonLakeFlyer:
QGC has build settings to disable serial port and bluetooth support. The android build was not correctly supporting these settings which respect to updating the AndroidManifest.xml correctly. This is now fixed.

This change was driven by fixing up upstream qgc for Herelink support in ways which require less upstream changes to get a Herelink build to work. In this specific case Herelink does not support usb (serial) or bluetooth.

### DIFF
--- a/QGCCommon.pri
+++ b/QGCCommon.pri
@@ -141,6 +141,10 @@ linux|macx|ios {
     }
 }
 
+NoSerialBuild {
+    message("Serial port support disabled")
+}
+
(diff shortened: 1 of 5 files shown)


Davidsastresas:
I just checked this. It is really cool what you did to modify the androidManifest searching for those comments. I didn't know that could be done on the *.pri files. I love it! Looks great to me, Thanks!

Davidsastresas (approved):

PR #10766, opened 2023-08-08
Geofence updates

DonLakeFlyer:
These changes had been made on the Herelink repo only when in fact they are general QGC features:
* Better ArduCopter fence support
* Support mavlink FENCE_STATUS messaging. Nothing fancy here now. At some point an App Setting to turn on/off as well as the timeout time would be good.

Screen Shot 2023-08-08 at 11 09 48 AM


DonLakeFlyer:
@Davidsastresas More of an FYI if you don't have time to look through.

PR #10771, opened 2023-08-11
Fly View: Fix normal and adsb vehicle sizing

DonLakeFlyer:
The code in VehicleMapItem.qml and FlyViewMap.qml was fighting with each other for vehicle sizing.


ryanjAA:
Works in flight view, disregards when mission-planning so there's an issue with something.

DonLakeFlyer:
Fly View and Plan View are two different things. This change only fixes a problem with Fly View. If you want different sizes in plan view you need to change plan view.

PR #10784, opened 2023-08-24
Handle openssl version mismatch

bkueng:
Some systems (like Ubuntu 22.04) only ship with OpenSSL 3.x, however Qt 5.15.2 links against OpenSSL 1.x.
This results in unresolved symbols for EVP_PKEY_base_id and SSL_get_peer_certificate.
To still get a connection we have to ignore certificate verification (connection is still encrypted but open to MITM attacks)
See https://bugreports.qt.io/browse/QTBUG-115146

It's not a great solution, but the best one I see given the situation.

Thanks to @julianoes for doing all the investigation & bug reporting.

### DIFF
--- a/src/QGCFileDownload.cc
+++ b/src/QGCFileDownload.cc
@@ -60,6 +60,8 @@ bool QGCFileDownload::download(const QString& remoteFile, const QVector<QPair<QN
         return false;
     }
 
+    setIgnoreSSLErrorsIfNeeded(*networkReply);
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
Thanks @bkueng

julianoes (commented):
Amazing, looks good. I'll test this over the coming days but I'm sure you have already done so.

PR #10786, opened 2023-08-24
Vehicle: add support for altitude above terrain as fact for telemetry panel

Davidsastresas:
Add support to have altitude above terrain as displayable telemetry. It queries terrain data when vehicle coordinate is changed to compute the altitude, but it has an elapsed timer to limit the time between those queries to 500ms minimum, because the terrain query system wasn't able to service them at higher rate, and 500ms is fine anyway for this information. This elapsed timer is not super elegant there, if anybody has a nicer idea for it I will be happy to work on it.

Last commit is unrelated, but I realized that function wasn't in the private slot declaration, and it belongs there. Same for first commit.

Thanks to https://araceuas.com for sponsoring this.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -84,6 +84,7 @@ const char* Vehicle::_groundSpeedFactName =         "groundSpeed";
 const char* Vehicle::_climbRateFactName =           "climbRate";
 const char* Vehicle::_altitudeRelativeFactName =    "altitudeRelative";
 const char* Vehicle::_altitudeAMSLFactName =        "altitudeAMSL";
+const char* Vehicle::_altitudeAboveTerrFactName =   "altitudeAboveTerr";
 const char* Vehicle::_altitudeTuningFactName =      "altitudeTuning";
 const char* Vehicle::_altitudeTuningSetpointFactName = "altitudeTuningSetpoint";
 const char* Vehicle::_flightDistanceFactName =      "flightDistance";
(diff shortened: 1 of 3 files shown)


s-lisovenko:
Very useful feature. Thanks!

s-lisovenko:
About the 500 ms timer, do you think it might be better to use a boolean variable that can be set to false when sending a request and to true in the handler slot. In this case, we can not send new requests until the current one is processed?

s-lisovenko:
Also regarding naming new facts, maybe altitudeAGL would be preferable? It is more consistent with avionics.

Davidsastresas:
Thanks for taking a look at it!

About the boolean, it would not be as simple because if for some reason the request is not serviced we would have a deadlock. A "similar"idea is already implemented in the terrainquery implementation itself, so It would have been totally feasible and safe to not implement timer at all, terrainquery mechanism takes care of it. I just tested a bit and I realized the terrain query system wasn't able to provide updates at a higher rate than that, so we just save the minimal extra cpu cycles it would take.

About the fact name, I totally agree, will fix it this week.

Thanks!

DonLakeFlyer:
Two things:
* How do I know if the altitude above terrain fact has a valid value in it? Seems like this should be initialized to nan or something so you know whether an initial terrain query hasn't returned yet.
* I think it might make sense to add an additional check on distance from previous coordinate query to limit even further the calls to the terrain server. I get paranoid about chewing laptop battery to do work which will just return the same value over and over again.

Davidsastresas:
@DonLakeFlyer I agree with your suggestions. I rebased to master and addressed your comments.

I needed to check for altitude change as well, because checking only distance would not query new terrain altitude in the situation when a copter gains altitude in the same coordinate.

I left the commits separated for you to check the changes, but If you confirm it to be fine I will squash them before merging.

thanks!

DonLakeFlyer:
Looks good. Squash it and you can merge it.

Davidsastresas:
Thanks, done!

PR #10788, opened 2023-08-25
Mavlink standard modes protocol

bkueng:
This brings the MAVLink standard modes protocol from https://github.com/mavlink/mavlink/pull/1750.
It allows the GCS to dynamically request the list of modes.
As the messages are in `development.xml`, they are disabled when doing a release build of QGC.

Implementation on the PX4 side: https://github.com/PX4/PX4-Autopilot/pull/20707

@hamishwillee fyi

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -724,6 +724,7 @@ HEADERS += \
     src/Vehicle/MultiVehicleManager.h \
     src/Vehicle/RemoteIDManager.h \
     src/Vehicle/StateMachine.h \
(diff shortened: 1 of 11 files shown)


patrikpordi:
Hey @bkueng,

As far as I understand external modes should show up on the QGC main, however it is still only works on the daily builds.
Could you have a look on @junwoo091400 's issue?

feoz0:
Is the issue related to external modes not showing in QGC main when building from source solved? I am still facing this problem

patrikpordi:
I do not think so!

PR #10790, opened 2023-08-26
CameraIO : Change param type assign method and fixed some compile warning

tbago:
the param type is only used in _sendParameter method, so don't need store the value in init, So I change the param type and param_value assign together.

Fixed some compile warning in CameraIO using Qt 5.15.


DonLakeFlyer:
Two things:
* I don't understand why this changed is needed?
* Also don't understand why fixes for Qt 5.12 are interesting. We don't support 5.12.

tbago:
1. the [_mavParamType](https://github.com/mavlink/qgroundcontrol/blob/master/src/Camera/QGCCameraIO.cc#L49C13-L49C26) is only used in the [QGCCameraParamIO::_sendParameter()](https://github.com/mavlink/qgroundcontrol/blob/master/src/Camera/QGCCameraIO.cc#L131C19-L131C33), so may use local variable more clear, and this two switch code [L1](https://github.com/mavlink/qgroundcontrol/blob/master/src/Camera/QGCCameraIO.cc#L43) [L2](https://github.com/mavlink/qgroundcontrol/blob/master/src/Camera/QGCCameraIO.cc#L143) like repeat code；

 2. just clerical error；

DonLakeFlyer:
If you can resolve the conflicts I'll merge this.

PR #10796, opened 2023-08-30
Restore tool inset functionality

JMare:
When I tried to use insets to position a widget on a custom fly view layer, I discovered that the inset system in master QGC is pretty much non functioning. This PR is my attempt to restore what I think the intended functionality of the system is, and it is a work in progress. There are some elements here which are open to interpretation, so I want some input from others on this.

**The inset system is currently not working because:**
- In the widget layer, very few widgets actually have insets passed out through totalToolInsets
- When the insets are passed through the default FlyViewCustomLayer, all the insets get wiped anyway before they are passed to the map layer
- The FlyViewMap only looks at the center insets to determine whether the vehicle has moved under the UI, this can cause it to not recenter the view if the vehicle has moved under a corner element

**The inset system, as I understand it, is intended to:**

1. Allow widgets in different layers to negotiate space on the flyview. For example, placing a widget on the custom layer that sits below the instrument panel.
2. Allow the map to identify when the vehicle has moved behind a widget, so it can recenter the map

You can verify that current master is not working by flying the aircraft under the instrument panel, you will see that the map view does not recenter (unless you fly off the edge of the screen).

**So on to my solution:**
- I have written a qml component called FlyViewInsetViewer which can be passed a QGCToolInset object and will display it on the screen, this will help in developing and understanding the insets. This is commented out in the flyview, can be enabled while developing when needed, won’t be included as a runtime option.
- I have gone through the widget layer and tried to fill in the correct widgets in each component
- I have corrected the FlyViewCustom layer so it correctly passes through the tool insets unchanged
- I have changed the FlyViewMap to check for the vehicle falling under a corner element

**Still to do:**
- Update the custom-example to correctly set insets
- provide an example in the custom-example of how to use the widget layer insets to position a custom widget
- much more thorough testing
- feedback from others
- consideration of edge cases
- Code tidying, git history cleaning (eg I know I have an unrelated commit in the history, its just for my ease of development on mac, I’ll clean it up before marking this as ready)

**But before I do all the above which is alot of work, I want to ask for feedback**
- Am I on the right track with this?
- Have I misunderstood the inset system?
- Would any or all of these changes be welcome?

Below is a screenshot of the inset viewer showing the total tool insets of a standard build
Untitled(4)

### DIFF
--- a/custom-example/qgcresources.qrc
+++ b/custom-example/qgcresources.qrc
@@ -26,11 +26,15 @@
 		<file alias="JoystickBezel.png">../resources/JoystickBezel.png</file>
 		<file alias="JoystickBezelLight.png">../resources/JoystickBezelLight.png</file>
 		<file alias="land.svg">../resources/land.svg</file>
+		<file alias="layout-bottom.svg">../resources/layout-bottom.svg</file>
+		<file alias="layout-right.svg">../resources/layout-right.svg</file>
 		<file alias="LockClosed.svg">../resources/LockClosed.svg</file>
 		<file alias="LockOpen.svg">../resources/LockOpen.svg</file>
 		<file alias="notile.png">../resources/notile.png</file>
+		<file alias="NoVideoBackground.jpg">../resources/NoVideoBackground.jpg</file>
 		<file alias="Pause.svg">../resources/Pause.svg</file>
 		<file alias="pause-mission.svg">../resources/pause-mission.svg</file>
+		<file alias="pencil.svg">../resources/pencil.svg</file>
 		<file alias="Play">../resources/Play.svg</file>
 		<file alias="PowerButton">../resources/PowerButton.svg</file>
 		<file alias="QGCLogoBlack">../resources/QGCLogoBlack.svg</file>
@@ -51,6 +55,9 @@
 		<file alias="XDelete.svg">../resources/XDelete.svg</file>
 		<file alias="XDeleteBlack.svg">../resources/XDeleteBlack.svg</file>
 		<file alias="waypoint.svg">../resources/waypoint.svg</file>
+		<file alias="Gripper.svg">../resources/Gripper.svg</file>
+		<file alias="GripperRelease.svg">../resources/GripperRelease.svg</file>
+		<file alias="GripperGrab.svg">../resources/GripperGrab.svg</file>
 		<file>../resources/icons/qgroundcontrol.ico</file>
 	</qresource>
 	<qresource prefix="/res/firmware">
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
A further cleanup to make this more understandable would be from this:
```
 leftEdgeTopInset:       toolStrip.leftInset
```
To this:
```
 leftEdgeTopInset:       toolStrip. leftEdgeTopInset
```

This way the properties in the controls which are copied into the QGCToolInsets are more understandable as to what they are and where they go.

JMare:
> A further cleanup to make this more understandable would be from this:
> 
> ```
>  leftEdgeTopInset:       toolStrip.leftInset
> ```
> 
> To this:
> 
> ```
>  leftEdgeTopInset:       toolStrip. leftEdgeTopInset
> ```
> 
> This way the properties in the controls which are copied into the QGCToolInsets are more understandable as to what they are and where they go.

Good point, that will make it much more maintainable.
I have implemented your suggestion, and I will be working through the rest of my todo list throughout the rest of this week.
I may post up screenshots of edge cases for discussion as I go

JMare:
- used updateqrc.py to update the resource files in the custom example
- fixed the custom-example to properly inherit tool insets from the widget layer
- made the custom-example use insets to dodge the virtual joysticks if present
- added an rectangle as an example of how to use parent tool insets to position a custom element on the screen while dodging the toolstrip and the virtual joysticks (this is invisible by default) 

This work is nearly finished. It isn't perfect and doesn't handle every edge case perfectly but its a big improvement over the current state in master. I would welcome any feedback on this work.
In order to test this, go in FlyView.qml and uncomment the FlyViewInsetViewer element, which will allow you to check where the insets are sitting as I have shown in my screenshots.

image
 
 I still need to do a bit more testing, and clean up the git history before I will mark this as a ready to to merge PR.

JMare:
One annoying edge case is what happens if the telemetry values panel is moved to the RHS of the screen. This would logically set the rightEdgeCenterInset, but a shown in the screenshot below, that blocks out a large section of the screen where the vehicle would otherwise be able to be positioned.  The vehicle will also recenter into the center of the yellow rectangle. Screenshot below is a slightly extreme example to demonstrate.
image

JMare:
I have cleaned up the git history on this PR and am marking this as ready for review.
I believe this PR brings the function of the tool inset system back to its original intention.
Would welcome any and all feedback. Happy to strip back this PR back if any elements of this aren't needed. Eg, the inset viewer could be omitted as could the example ui element on the custom layer.

DonLakeFlyer:
> One annoying edge case is what happens if the telemetry values panel is moved to the RHS of the screen.

Yeah, but that is kind of an incorrect use of the right edge of the screen. The right edge should really be used for fixed size, smaller width things which don't block too much.

Sorry, for taking so long to get back to this.

DonLakeFlyer:
@JMare Thanks for putting all the work into this

JMare:
Thanks @DonLakeFlyer!

😄

PR #10810, opened 2023-09-13
fix FirmwarePlugin: increase arming timeout to 1.5s

bkueng:
As the arming flag is set from the heartbeat received at 1Hz, we need to wait a bit longer than 1s.

The 1s timeout sometimes led to this error message when trying to start a mission: "Unable to start mission: Vehicle rejected arming."

Fixes https://github.com/PX4/PX4-Autopilot/issues/22042


DonLakeFlyer:
Thanks @bkueng

PR #10813, opened 2023-09-20
WIP: joystick: Set default throttleMode based on vehicle thrust configuration

patrickelectric:


DonLakeFlyer:
More stable only fixes which don't seem to make sense to me?

Issue #10811, opened 2023-09-13
Richtext message display broken


bkueng:
It looks like https://github.com/mavlink/qgroundcontrol/pull/10603 broke rich text rendering:

![image](https://github.com/mavlink/qgroundcontrol/assets/281593/b14c1dd7-8f03-43ce-995d-7a98552fe35a)

![image](https://github.com/mavlink/qgroundcontrol/assets/281593/ca82e9ff-dcb5-4e00-86b3-756d9a86d0ab)

@DonLakeFlyer could you have a look?

DonLakeFlyer:
Broken by this as well: https://github.com/mavlink/qgroundcontrol/pull/10533

DonLakeFlyer:
To me using formatting in status text seems like a not so great idea. Not sure how you keep both text like this "foo < bar" which could be in any status text message and this "<b>foo</b>" working correctly.

bkueng:
The above comes through the events interface: https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L1623.
And the escaping is missing there as well. Initially I just did it that way to fit into the existing UI elements but ideally we would do a bit more.

DonLakeFlyer:
OK, so can you fix this then? The vehicle message stuff was never meant to understand how to do rich text from the vehicle itself. It only used rich text for internal QGC formatting usage.

You'll need to escape the regular text coming through the event stuff. And then I would hack UASMessageHandler for now to look for a "<br/>" in the message and skip the escaping there if so. And then add rich text format back to the critical message label in MainRootWindow. I started to hack it myself, but I'm not quite so sure about that event code to get it right. I also couldn't get the failsafe to generate an event to test it with.

PR #10818, opened 2023-09-26
Fix Richtext message display & html escaping for events

bkueng:
And allows to use URL's & parameters in messages:
image

Fixes https://github.com/mavlink/qgroundcontrol/issues/10811

### DIFF
--- a/.github/workflows/android_release.yml
+++ b/.github/workflows/android_release.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Get all tags for correct version determination
         working-directory:  ${{ github.workspace }}
         run: |
-          git fetch --all --tags -f
+          git fetch --all --tags -f --depth 1
 
       - name: Install Qt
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
@bkueng Thanks Beat. Just so I understand this correctly you separated the complex status text into two parts:
* The regular part which comes from a normal STATUS_TEXT thing. This part is htmlEscaped.
* The "description" which is the extra fancy possibly rich text part. This part is not htmlEscaped,

bkueng:
Yes it's separated and all cases are escaped, including the description. I had to move the escaping outside, as the events might contain urls. This is done here: https://github.com/mavlink/qgroundcontrol/pull/10818/files#diff-0beffce9af19c5b54b97706a2354f113c6dd6c71a9bde40eec507c85a751eb86R49, which only escapes regular text.

PR #10822, opened 2023-09-28
Fix typo

Davidsastresas:


Davidsastresas:
@DonLakeFlyer I am bringing this fix from your herelink branch.

Davidsastresas:
Sorry, I saw you disabled it, and the reason why, on a later commit. Apologies for not checking it first.

DonLakeFlyer:
Oh yeah. Turns out these settings were in the system prior to the new SettingsGroup code going in with the old name. So I had to keep the previous group naming otherwise it would hose folks.

PR #10823, opened 2023-09-29
Master pr/fix custom android build

Davidsastresas:
android.pri was broken when a custom build has a custom android folder. PRE_TARGETDEPS was updated before the custom build evaluation, so if custom build found, the PRE_TARGETDEPS would still be assigned the previous ( QGC standard ) AndroidManifest.xml.

Also, it was needed to handle the case when we have a custom android folder, but no custom AndroidManifest.xml ( in case we are only changing android icons for instance.

@DonLakeFlyer does it look good to you? Thanks!

### DIFF
--- a/android.pri
+++ b/android.pri
@@ -20,7 +20,10 @@ android_source_dir_target.depends = FORCE
 exists($$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR) {
     message("Merging$$ $$ANDROID_PACKAGE_QGC_SOURCE_DIR and $$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR to $$ANDROID_PACKAGE_SOURCE_DIR")
 
-    android_source_dir_target.target = $$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR/AndroidManifest.xml
+    exists($$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR/AndroidManifest.xml) {
+        android_source_dir_target.target = $$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR/AndroidManifest.xml
+    }
+    
     android_source_dir_target.commands = $$android_source_dir_target.commands && \
             $$QMAKE_COPY_DIR $$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR/* $$ANDROID_PACKAGE_SOURCE_DIR && \
             $$QMAKE_STREAM_EDITOR -i \"s/package=\\\"org.mavlink.qgroundcontrol\\\"/package=\\\"$$QGC_ANDROID_PACKAGE\\\"/\" $$ANDROID_PACKAGE_SOURCE_DIR/AndroidManifest.xml
--- a/android.pri
+++ b/android.pri
@@ -12,7 +12,6 @@ android_source_dir_target.target = $$ANDROID_PACKAGE_QGC_SOURCE_DIR/AndroidManif
 android_source_dir_target.commands = \
     $$QMAKE_MKDIR $$ANDROID_PACKAGE_SOURCE_DIR && \
     $$QMAKE_COPY_DIR $$ANDROID_PACKAGE_QGC_SOURCE_DIR/* $$ANDROID_PACKAGE_SOURCE_DIR
-PRE_TARGETDEPS += $$android_source_dir_target.target
 QMAKE_EXTRA_TARGETS += android_source_dir_target
 android_source_dir_target.depends = FORCE
 
@@ -26,6 +25,8 @@ exists($$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR) {
             $$QMAKE_COPY_DIR $$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR/* $$ANDROID_PACKAGE_SOURCE_DIR && \
             $$QMAKE_STREAM_EDITOR -i \"s/package=\\\"org.mavlink.qgroundcontrol\\\"/package=\\\"$$QGC_ANDROID_PACKAGE\\\"/\" $$ANDROID_PACKAGE_SOURCE_DIR/AndroidManifest.xml
 }
+
+PRE_TARGETDEPS += $$android_source_dir_target.target
 
 # Insert package name into manifest file
 


DonLakeFlyer:
So I'm confused as to what this is supposed to fix. That way it works now (or at least it's supposed to):
* First the entired standard package source dir is copied to a build location. This should include all the original needed android package files.
* Then the entire custom package source dir (if it exists) is copied over that. This means custom builds can override any part the way, all or only specific stuff.
* Either way you always end up with everything you need in the build dir for package stuff

Given that what is really wrong? I think it might be that android_source_dir_target.target should point to $$ANDROID_PACKAGE_SOURCE_DIR. Which is the ultimate end destination of everything?

Davidsastresas:
- In the case of the second commit, line 23, we are assigning to that variable custom/android/androidmanifest.xml, and the only condition to enter there is that custom/android exists. I think this is wrong because we could have overriden just the icons but not the androidmanifest. In fact this is exactly why I spotted it, trying to do a custom build and only populating custom/android/res, but not using a custom AndroidManifest.xml.

- In the case of the first commit, "PRE_TARGETDEPS += $$android_source_dir_target.target" is done prior to to the possible assignment of  android_source_dir_target.target again for our custom AndroidManifest.xml. So in other words, regardless of custom AndroidManifest.xml, we were always loading to PRE_TARGETDEPS the standard QGC AndroidManifest.xml

I hope that makes sense.

DonLakeFlyer:
Ok, so let's go through the existing code. I'm not getting the problem:
```
ANDROID_PACKAGE_SOURCE_DIR          = $$OUT_PWD/ANDROID_PACKAGE_SOURCE_DIR  # Tells Qt location of package files for build
ANDROID_PACKAGE_QGC_SOURCE_DIR      = $$PWD/android                         # Original location of QGC package files
ANDROID_PACKAGE_CUSTOM_SOURCE_DIR   = $$PWD/custom/android                  # Original location for custom build override package files

# We always move the package files to the ANDROID_PACKAGE_SOURCE_DIR build dir so we can modify the manifest as needed

android_source_dir_target.target = $$ANDROID_PACKAGE_QGC_SOURCE_DIR/AndroidManifest.xml
android_source_dir_target.commands = \
    $$QMAKE_MKDIR $$ANDROID_PACKAGE_SOURCE_DIR && \
    $$QMAKE_COPY_DIR $$ANDROID_PACKAGE_QGC_SOURCE_DIR/* $$ANDROID_PACKAGE_SOURCE_DIR
PRE_TARGETDEPS += $$android_source_dir_target.target
QMAKE_EXTRA_TARGETS += android_source_dir_target
android_source_dir_target.depends = FORCE

# Custom builds can override android package file

exists($$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR) {
    message("Merging$$ $$ANDROID_PACKAGE_QGC_SOURCE_DIR and $$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR to $$ANDROID_PACKAGE_SOURCE_DIR")

    android_source_dir_target.target = $$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR/AndroidManifest.xml
    android_source_dir_target.commands = $$android_source_dir_target.commands && \
            $$QMAKE_COPY_DIR $$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR/* $$ANDROID_PACKAGE_SOURCE_DIR && \
            $$QMAKE_STREAM_EDITOR -i \"s/package=\\\"org.mavlink.qgroundcontrol\\\"/package=\\\"$$QGC_ANDROID_PACKAGE\\\"/\" $$ANDROID_PACKAGE_SOURCE_DIR/AndroidManifest.xml
}
```

* ANDROID_PACKAGE_SOURCE_DIR is a qmake variable which tells it where the android package stuff is: manifest, icons, whatever. That is set to the "<build directory>/ANDROID_PACKAGE_SOURCE_DIR" directory. This directory is actually named "ANDROID_PACKAGE_SOURCE_DIR", this not is a variable substitution.
* The regular non-custom QGC build has this content in the android directory of the repo. The ANDROID_PACKAGE_QGC_SOURCE_DIR variable points to it.
* Custom builds can override any one of the file in this android package stuff. The location for any of these overrides (if any) is custom/android. The variable ANDROID_PACKAGE_CUSTOM_SOURCE_DIR.
* The way it is supposed to accomplish this is to:
  * Create the ANDROID_PACKAGE_SOURCE_DIR which is a build system artifact not part of the repo
  * Copy the regular QGC package stuff to that directory
  * Copy the custom overridden package stuff to that same directory. Overwriting one or more of the regular QGC package files.
  * The expectation is this sequence of steps happens every single time you build. It is not targeted to some specific file change target to be done optionally.

My understanding of how it should have worked:
* The combination of `android_source_dir_target.target = $$ANDROID_PACKAGE_QGC_SOURCE_DIR/AndroidManifest.xml` and `android_source_dir_target.depends = FORCE` should mean that the things in `android_source_dir_target` should run every time you build. They should not be dependent on modifying or targeting a specific manifest file or anythign else.
* The first block of `android_source_dir_target` stuff handles the copying of the regular qgc package stuff to the build directory.
* The second block of `android_source_dir_target` stuff handles adds (if needed) the copying of the regular custom overrides package stuff to the build directory.

So questions are:
* Given the explanation of what it is attempting to accomplish. Is the goal incorrect somewhow in that it won't work?
* Given the actual code, why doesn't it do what the goal is?

Davidsastresas:
@DonLakeFlyer first of all thank you very much for taking the time on explaining your point that accurately. 

My build stopped working right after after creating a custom/android/res, and populating it with custom app icons. After the changes above, particularly the one about PRE_TARGETDEPS, I managed to get it to build again, and take the custom icons for android as well.

Answering your latest questions:

**- Given the explanation of what it is attempting to accomplish. Is the goal incorrect somewhow in that it won't work?** Your explanation makes total sense, and I actually spent a ton of time trying to figure out why it wasn't working for me. My understanding of what was happening, and my understanding of why that commit solves it, is the following:

- Before the commit: we do the following:

`PRE_TARGETDEPS += $$android_source_dir_target.target
`
before entering the custom condition. If the custom condition (custom/android) is positive, we would change now our android_source_dir_target.target to the custom AndroidManifest.xml, while it has the previous value as the standard AndroidManifest.xml. 

This is one detail I am not 100% sure of, and could be the key why we are not understanding each other. is the above fine then? when we update the value of android_source_dir_target.target in the custom condition, isn't PRE_TARGETDEPS still holding the old value of android_source_dir_target.target?

Besides that "issue", we have that the entering condition for the "custom loop", is just that custom/android exists. But we don't check if custom/android/AndroidManifest.xml exists. And a couple of lines below we do:

`android_source_dir_target.target = $$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR/AndroidManifest.xml
`
But at this point we have no warranty at all that that custom/android/(ANDROID_PACKAGE_CUSTOM_SOURCE_DIR)AndroidManifest.xml exists. We would be here if we create custom/android/res in order to override the app icons, but we don't have a custom AndroidManifest.xml. In this scenario, wouldn't android_source_dir_target.target get an undefined value?

**- Given the actual code, why doesn't it do what the goal is?** My understanding of it is the previous answer. I am not specially familiarized with qmake/cmake, and I actually spend some considerable time when I spotted this to try to understand what could be happening, but as you are so sure about it, it really makes me doubt and wonder if there is something else going on here in my build setup.

Thanks!

DonLakeFlyer:
> Besides that "issue", we have that the entering condition for the "custom loop", is just that custom/android exists. But we don't check if custom/android/AndroidManifest.xml exists. And a couple of lines below we do:
> 
> `android_source_dir_target.target = $$ANDROID_PACKAGE_CUSTOM_SOURCE_DIR/AndroidManifest.xml `
> 
> But at this point we have no warranty at all that that custom/android/ (ANDROID_PACKAGE_CUSTOM_SOURCE_DIR) AndroidManifest.xml exists. We would be here if we create custom/android/res in order to override the app icons, but we don't have a custom AndroidManifest.xml. In this scenario, wouldn't android_source_dir_target.target get an undefined value, as the path doesn't exist?

What you say is correct. But it shouldn't matter if AndroidManifest.xml is in the custom directories. It should only matter that one version of it ends up in the build directories. Which is guaranteed to be true because it copies regular QGC files first which has everything. And then it overwrites those with whatever is in custom.

As I said in a previous comment I think the problem is that android_source_dir_target.target should be pointing to the build directory where everything is copied to.

> My build stopped working right after after creating a custom/android/res, and populating it with custom app icons.

Let me try to reproduce the problem you are seeing which should help me fix it.

All of this android qmake stuff is kind of magic and a PITA.

Davidsastresas:
Closed in favor of https://github.com/mavlink/qgroundcontrol/pull/10826

PR #10826, opened 2023-10-05
Fix android custom build package file overrides

DonLakeFlyer:
Replacement for #10823

@Davidsastresas Can you give this a try? I believe I figured out what the problem was.


Davidsastresas:
Sorry for taking this long to test @DonLakeFlyer.

I just tested this and it works for me. I understand now why you didn't like the other approach.

Looks great to me, I think we should merge this right away. Thanks for checking it!

DonLakeFlyer:
CI is failing. I'll have time to take a look at it when I get back home in a week.

Davidsastresas:
@DonLakeFlyer I pulled that string of CI failing.

It seems after your commit here, all the stuff we put in QMAKE_POST_LINK is repeated several times in a loop after a build. I tried to track it around, and it is not an issue of any of the .pro and .pri files, all the stuff in those is only executed once, but for some reason this commit provokes this on Android. 

I ran out of time to spend on this, and I didn't come to understand the reason behind this, I just wanted to report back in case you look back into it.

The changes I proposed however ( closed in favor of this PR ) do not provoke this behaviour and seem to work fine, just in case for you to know if you want to compare with it when debugging.

Thanks!

DonLakeFlyer:
> It seems after your commit here, all the stuff we put in QMAKE_POST_LINK is repeated several times in a loop after a build.

Yeah, I've seen that before I believe and fixed it in my custom repo. Let me figure out what I did there and bring it across

DonLakeFlyer:
@Davidsastresas This latest version fixes the looping and is sort of a combination of my changes and your changes from the previous pull. It triggers the file copies and manifest updates whenever the AndroidManifest.xml changes. The one thing it doesn't do is trigger the file copies of something else in the android directory changing: icons or whatever. So if you do taht you'll need clean build. Since CI is clean build, those builds will always work correctly. I can't quite figure out how to make that work. That said, that wasn't working before this pull either. I may go back into this later on to figure out how to fix that. But for now I'm taking a break from this android crap since I've been fighting with it forever and it makes me crazy.

DonLakeFlyer:
I'm going to redo this pull and remove the custom resource updates from it

Davidsastresas:
Thanks Don. I just checked and it is working fine for me as well. As you said you were going to redo the pull and remove the custom resources from it, I did it myself, preserving your authorship, I hope that is fine: https://github.com/mavlink/qgroundcontrol/pull/10826

Thanks!

PR #10829, opened 2023-10-13
Optimize map provider iteration to prevent app crashes. Fixed application crash due to segmentation fault.

s-lisovenko:
Optimize map provider iteration to prevent app crashes. Fixed application crash due to segmentation fault.

Changed from QHashIterator to using a constant iterator for iterating through the map providers in _providersTable. This change prevents application crashes during a map download session caused by the potential modification of an implicitly shared hash.

### DIFF
--- a/src/QtLocationPlugin/QGCMapUrlEngine.cpp
+++ b/src/QtLocationPlugin/QGCMapUrlEngine.cpp
@@ -156,13 +156,9 @@ quint32 UrlFactory::averageSizeForType(const QString& type) {
 }
 
 QString UrlFactory::getTypeFromId(int id) {
-
-    QHashIterator<QString, MapProvider*> i(_providersTable);
-
-    while (i.hasNext()) {
-        i.next();
-        if ((int)(qHash(i.key())>>1) == id) {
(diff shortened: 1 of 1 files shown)


Davidsastresas:
This is great, thank you very much.

How did you find the bug? I mean, is there any way we can "force" that modification of the implicitly shared hash, in order to reproduce the crash?

Thanks!

s-lisovenko:
> This is great, thank you very much.
> 
> How did you find the bug? I mean, is there any way we can "force" that modification of the implicitly shared hash, in order to reproduce the crash?
> 
> Thanks!

I tried to download the cards I needed and the application began to randomly fall into segfault:)
I found that this object is being used from different threads. buy QHash guarantees safe access for reading from different threads, but apparently QHashIterator does not provide such guarantees because the atomic access error occurred precisely when trying to iterate and we had a resource race.


As for how to reproduce it, I think that you need to call this method with QHashIterator from another thread. In my case, the application always crashed after 10-20 seconds, so I didn’t really try to reproduce the error.

patrickelectric (approved):

PR #10835, opened 2023-10-18
Add support for EFI, EKF and Generator telemetry

Davidsastresas:
EFI and Generator are already in upstream mavlink common.xml, but I just realized EKF one is ardupilotmega.xml only.

Should we implement something through firmwareplugin to workaround it? Or we should just drop it and not include it?

For the rest of it these changes have been working for some time in a custom build.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -745,6 +745,8 @@ HEADERS += \
     src/Vehicle/VehicleVibrationFactGroup.h \
     src/Vehicle/VehicleWindFactGroup.h \
     src/Vehicle/VehicleHygrometerFactGroup.h \
+    src/Vehicle/VehicleGeneratorFactGroup.h \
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
> EFI and Generator are already in upstream mavlink common.xml, but I just realized EKF one is ardupilotmega.xml only.

I made the call years ago that I was going to stop supporting mavlink stuff which was in the ardupilot dialect only. It just makes a big unmanageable mess. It's up to ardupilot to get things into common. So I would drop it.

Davidsastresas:
@DonLakeFlyer I understand, it is fair enough.

I rebased to master and I removed EKF, leaving only the EFI and generator, which is in mavlink common.xml.

Thanks!

DonLakeFlyer:
Is this good to merge now. It is still marked as Draft.

Davidsastresas:
Undrafted and merged, thanks!

PR #10836, opened 2023-10-20
Use all mavlink includes instead of ardupilotmega for cmake build

s-lisovenko:
Switched mavlink include option from 'ardupilotmega' to 'all' in the CMakeLists file. This modification will allow the use of all the functionalities offered by the mavlink library. the range of functionalities that can be utilized from the mavlink library.
@bkueng Please check out this PR because it fixes cmake compilation for StandardModes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,7 +164,7 @@ include_directories(
 	libs/libevents
 
 	libs/mavlink/include/mavlink/v2.0
-	libs/mavlink/include/mavlink/v2.0/ardupilotmega
+	libs/mavlink/include/mavlink/v2.0/all
 	libs/mavlink/include/mavlink/v2.0/common
 
 	libs/shapelib


auturgy:
all.xml includes development.xml (which is why it has the common modes messages/enums). You shouldn't be building releases with it.

s-lisovenko:
> all.xml includes development.xml (which is why it has the standard modes messages/enums). You shouldn't be building releases with it.

Thank you for reply.
How to properly connect development to the cmake build? Because mavlink from Ardupilotmega is used by default and the application does not compile under cmake build

auturgy:
There's a number of open issues regarding the build failures: I'm not a qgc dev but I think it would be best for that to be resolved. If the standard modes are now being used in a stable flight stack release then we should look to move them into common. 
In the interim building locally with your cmake change is fine.

s-lisovenko:
> There's a number of open issues regarding the build failures: I'm not a qgc dev but I think it would be best for that to be resolved. If the standard modes are now being used in a stable flight stack release then we should look to move them into common.
> In the interim building locally with your cmake change is fine.

Yes, i noticed that qmake release build used development includes also.

auturgy:
Building releases with development.xml basically makes development.xml redundant. As I said above, I'm not a qgc dev, so in this context it's not up to me, but from a mavlink perspective we created development.xml because the flight stacks weren't respecting "WIP" messages. Sounds like that's being repeated 🤷‍♂️

bkueng (approved):
@auturgy the approach I used to disable development messages in release builds is to exclude them via ifdef: https://github.com/mavlink/qgroundcontrol/pull/10788/commits/489a03bbc1bb02fce06d2e5f9920bf33d5b80e3a. We'll need to replicated that in cmake, at the latest when switching away from qmake.
There's other approaches, like using a different dialect for release builds (which also requires ifdefs), or making it a runtime option.
Either one works for me. I find the runtime option interesting, as it allows users who are ok with potential risk of breakage to still use the functionality.

DonLakeFlyer:
Is this ready to merge now?

DonLakeFlyer:
Thanks @bkueng

PR #10839, opened 2023-10-24
Remove old "offboard compass calibration" code

peterbarker:
We're just about to merge a PR which removes handling of `SET_SENSOOR_OFFSETS` mavlink commands from ArduPilot: https://github.com/ArduPilot/ardupilot/pull/25305

Don added code in 2016 to use ArduPilot's onboard compass calibration...

I've tested QGC after this and it works just fine to calibrate compasses on a CubeOrange


patrickelectric (approved):

DonLakeFlyer:
@peterbarker I don't quite remember how QGC decided which cal variant to use. I don't see a place where the UI is cauterized as well to to show this. Or is that not how this works. Just want to make sure there isn't a handling ui element or button somewhere.

peterbarker:
> @peterbarker I don't quite remember how QGC decided which cal variant to use. I don't see a place where the UI is cauterized as well to to show this. Or is that not how this works. Just want to make sure there isn't a handling ui element or button somewhere.

I believe the ability to enter the offboard compass stuff was removed in this PR.  Perhaps by accident :-)

https://github.com/mavlink/qgroundcontrol/pull/9010/files#diff-82898962f216a84a178ed2fc044fc052effbe97fbf5e7014b4658a391ef3645bL526

src/AutoPilotPlugins/APM/APMSensorsComponentController.cc
```diff
@@ -274,9 +271,6 @@ void APMSensorsComponentController::_mavCommandResult(int vehicleId, int compone
                                      0,             // no delayed start
                                      0);            // no auto-reboot
 
-        } else {
-            // Onboard mag cal is not supported
-            _compassCal.startCalibration();
```
DonLakeFlyer, line 279: @peterbarker This is what triggered the offboard cal. If the onboard was not accepted. So then this change doesn't have any dangling stuff left.

PR #10841, opened 2023-10-26
APMFirmwarePlugin: reposition using MAV_CMD_DO_REPOSITION 

peterbarker:
I've tested this in ArduPilot SITL.

The interface used for this testing was click-on-map and select the "Move to" option.

I tested Rover, Copter and Plane which actually implement DO_REPOSITION, and on Sub which doesn't.  All act as expected.

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.h
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.h
@@ -137,4 +137,7 @@ public:
 
     QTime lastBatteryStatusTime;
     QTime lastHomePositionTime;
+
+    bool  MAV_CMD_DO_REPOSITION_supported = false;
+    bool  MAV_CMD_DO_REPOSITION_unsupported = false;
 };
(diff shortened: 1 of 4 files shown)


patrickelectric (approved):

src/FirmwarePlugin/APM/APMFirmwarePlugin.h
```diff
@@ -137,4 +137,7 @@ class APMFirmwarePluginInstanceData : public QObject
 
     QTime lastBatteryStatusTime;
     QTime lastHomePositionTime;
+
+    bool  MAV_CMD_DO_REPOSITION_supported;
```
DonLakeFlyer, line 141: @peterbarker Are these getting zero initialized someplace else already?

peterbarker: My bad.

We zero all allocations as a matter of course in ArduPilot, so it's going to be a blind spot for me :-)

I've zeroed these here in the header now.

DonLakeFlyer:
Thanks @peterbarker

PR #10842, opened 2023-10-26
QGCTextField: fix forcing the use of virtual numeric keyboard

HmZyy:
**Description:**
This pull request addresses an issue with the input method hints that was causing problems with Android devices, specifically related to the use of the virtual numeric keyboard. The problem arose from an inverted condition, which led to incorrect behavior.

**Changes Made**
Reversed the condition to correctly force the use of the virtual numeric keyboard only when numericValuesOnly is true and the device is not an IOS device.

**Problem**
This condition was causing problems on Android devices by not allowing users to search for parameters or create com link names using alphabetic characters.

C8AD48CF-62D4-4C4A-B44F-98CE179F3121

7F5A305A-31FB-4299-A1DF-3175BE9278BF

### DIFF
--- a/src/QmlControls/QGCTextField.qml
+++ b/src/QmlControls/QGCTextField.qml
@@ -13,8 +13,8 @@ TextField {
     activeFocusOnPress: true
     antialiasing:       true
     inputMethodHints:   numericValuesOnly && !ScreenTools.isiOS ?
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Sorry, for screwing that up

PR #10848, opened 2023-10-31
Remove android sdks from github runner which are newer than 33

DonLakeFlyer:
This forces gradle to use android-33 sdk version. Without this CI builds will fail since gradle blows up with newer SDKs.


DonLakeFlyer:
@kibidev Any ideas here? I'm just poking around in the dark trying to this. Was hoping the problem was similar to your past changes.

DonLakeFlyer:
I believe I have this figure out now

PR #10850, opened 2023-11-03
Vehicle: keep commands in list if in progress

julianoes:
We should not remove commands from the command list when we receive an IN_PROGRESS ack because we should eventually receive a final ack.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -2738,8 +2738,8 @@ double Vehicle::minimumEquivalentAirspeed()
     return _firmwarePlugin->minimumEquivalentAirspeed(this);
 }
 
-bool Vehicle::hasGripper()  const 
-{ 
+bool Vehicle::hasGripper()  const
+{
     return _firmwarePlugin->hasGripper(this);
 }
 
@@ -3850,7 +3850,7 @@ void Vehicle::_handleRawImuTemp(mavlink_message_t& message)
 {
     // This is used by compass calibration
     emit mavlinkRawImu(message);
-    
+
     mavlink_raw_imu_t imuRaw;
     mavlink_msg_raw_imu_decode(&message, &imuRaw);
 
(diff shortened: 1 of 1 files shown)


patrickelectric (approved):

peterbarker:
Ah, glad you're doing this; I was looking at changing the ArduPilot-specific calibration stuff to use this mechanism....

src/Vehicle/Vehicle.cc
```diff
@@ -3254,7 +3254,10 @@ void Vehicle::_handleCommandAck(mavlink_message_t& message)
     int entryIndex = _findMavCommandListEntryIndex(message.compid, static_cast<MAV_CMD>(ack.command));
     bool commandInList = false;
     if (entryIndex != -1) {
-        MavCommandListEntry_t commandEntry = _mavCommandList.takeAt(entryIndex);
+        // We remove the command from the list unless it's still in progress.
```
bkueng, line 3257: While in progress, are retries handled correctly?

bkueng (commented):
The change is fine with me, but [autotune](https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Autotune.cpp) does not follow that semantics and needs to be checked.

DonLakeFlyer:
Beat is correct. Supporting this is more involved than just the list management. I'll take a look.

DonLakeFlyer:
There are also big problems with how MAV_RESULT_IN_PROGRESS and progress values were plumbed through Vehicle::sendMavCommandWithHandler. Given the fact that any command can start to send IN_PROGRESS before sending an ack, all callbacks from Vehicle::sendMavCommandWithHandler as currently implemented need to handle IN_PROGRESS specifically otherwise they will break if the command ever sends that.

DonLakeFlyer:
I should have a pull which cleans this all up in a day or so...

julianoes:
Closing this. Thanks @DonLakeFlyer for #10852.

PR #10852, opened 2023-11-06
Fix up support for MAV_RESULT_IN_PROGRESS

DonLakeFlyer:
Replacement for #10850.

The current problems with Vehicle::sendMavCommandWithHandler in progress support:
* When a MAV_RESULT_IN_PROGRESS ack is received the Vehicle code does the following incorrect things:
  * The command entry is removed from the entry list. Since this is only a progress indication to command is still in play and we are still waiting for the real ack so it should be left in the list.
  * The timeout for the command ack which is used is left alone. This means that it will fire too early with respect to not getting a completion ack. When an in progress ack is received the timeout should be restarted to give more time for the completion ack to be seen.
  * If the command supports retries and we get an in progress we should not retry the command. The reason being we did hear back from the vehicle with respect to the fact that it got the command. In this case the try count should be reset back to 1 such that no retry happens.
* The other problem relates to the fact that the progress indication comes through the same callback as completion acks. This makes writing the ack callback handlers fragile and complex. Specifically in the case where the command did not initially support MAV_RESULT_IN_PROGRESS. You write some simple checks for command success/fail (like result != MAV_RESULT_ACCEPTED). But then the command suddenly returns MAV_RESULT_IN_PROGRESS in a newer version of firmware which causes your code to think the command failed when it reality it didn't. There currently are multiple examples of this in the existing callback handler implementations.

The changes made to fix all this:
* The Vehicle code was changed to fix all the bugs with handling MAV_RESULT_IN_PROGRESS described above. The changes are all in Vehicle.cc/h.
* I split the handling of in progress acks and completions acks into two separate callback handlers. This way you can implement a simple completion callback handler which won't break if you suddenly start getting progress notications. And you can just ignore the progress information if you like. Or if you want you can implement a separate progress only callback. There are a decent number of changes throughout the code because of the new signature to the Vehicle:: sendMavCommandWithHandler method. But they are mostly mechanical changes in nature as opposed to logic changes.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -2960,8 +2960,7 @@ void Vehicle::sendMavCommand(int compId, MAV_CMD command, bool showError, float 
 {
     _sendMavCommandWorker(false,            // commandInt
                           showError,
-                          nullptr,          // resultHandler
-                          nullptr,          // resultHandlerData
+                          nullptr,          // no handlers
                           compId,
                           command,
                           MAV_FRAME_GLOBAL,
@@ -2982,12 +2981,11 @@ void Vehicle::sendCommand(int compId, int command, bool showError, double param1
                 static_cast<float>(param7));
 }
 
-void Vehicle::sendMavCommandWithHandler(MavCmdResultHandler resultHandler, void *resultHandlerData, int compId, MAV_CMD command, float param1, float param2, float param3, float param4, float param5, float param6, float param7)
+void Vehicle::sendMavCommandWithHandler(const MavCmdAckHandlerInfo_t* ackHandlerInfo, int compId, MAV_CMD command, float param1, float param2, float param3, float param4, float param5, float param6, float param7)
 {
     _sendMavCommandWorker(false,                // commandInt
                           false,                // showError
-                          resultHandler,
-                          resultHandlerData,
+                          ackHandlerInfo,
                           compId,
                           command,
                           MAV_FRAME_GLOBAL,
@@ -2998,20 +2996,18 @@ void Vehicle::sendMavCommandInt(int compId, MAV_CMD command, MAV_FRAME frame, bo
 {
     _sendMavCommandWorker(true,         // commandInt
                           showError,
-                          nullptr,      // resultHandler
-                          nullptr,      // resultHandlerData
+                          nullptr,      // no handlers
                           compId,
                           command,
                           frame,
                           param1, param2, param3, param4, param5, param6, param7);
 }
 
-void Vehicle::sendMavCommandIntWithHandler(MavCmdResultHandler resultHandler, void *resultHandlerData, int compId, MAV_CMD command, MAV_FRAME frame, float param1, float param2, float param3, float param4, double param5, double param6, float param7)
+void Vehicle::sendMavCommandIntWithHandler(const MavCmdAckHandlerInfo_t* ackHandlerInfo, int compId, MAV_CMD command, MAV_FRAME frame, float param1, float param2, float param3, float param4, double param5, double param6, float param7)
 {
     _sendMavCommandWorker(true,                   // commandInt
                           false,                  // showError
-                          resultHandler,
-                          resultHandlerData,
+                          ackHandlerInfo,
                           compId,
                           command,
                           frame,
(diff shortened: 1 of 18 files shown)


src/Vehicle/Vehicle.cc
```diff
@@ -3264,12 +3275,21 @@ void Vehicle::_handleCommandAck(mavlink_message_t& message)
...
     if (entryIndex != -1) {
-        MavCommandListEntry_t commandEntry = _mavCommandList.takeAt(entryIndex);
-        if (commandEntry.command == ack.command) {
-            if (commandEntry.resultHandler) {
-                (*commandEntry.resultHandler)(commandEntry.resultHandlerData, message.compid, static_cast<MAV_RESULT>(ack.result), ack.progress, MavCmdResultCommandResultOnly);
+        if (ack.result == MAV_RESULT_IN_PROGRESS) {
+            MavCommandListEntry_t commandEntry = _mavCommandList.at(entryIndex);  // Command has not completed yet, don't remove
+
+            commandEntry.maxTries = 1;              // Vehicle responsed to command so don't retry
+            commandEntry.elapsedTimer.restart();    // We've heard from vehicle, restart elapsed timer for no ack received timeout
```
julianoes, line 3283: Nice, restarting the timer is important!

src/comm/MockLink.cc
```diff
@@ -1052,6 +1055,53 @@ void MockLink::_handleFTP(const mavlink_message_t& msg)
...
+    case MockLink::MAV_CMD_MOCKLINK_RESULT_IN_PROGRESS_ACCEPTED:
+        // Test command which sends in progress messages and then acceptance ack
+        commandResult = MAV_RESULT_ACCEPTED;
+        break;
+    case MockLink::MAV_CMD_MOCKLINK_RESULT_IN_PROGRESS_FAILED:
+        // Test command which sends in progress messages and then failure ack
+        commandResult = MAV_RESULT_FAILED;
+        break;
+    case MockLink::MAV_CMD_MOCKLINK_RESULT_IN_PROGRESS_NO_ACK:
+        // Test command which sends in progress messages and then never sends final result ack
```
julianoes, line 1072: I assume this is to test legacy behavior? It should not be encouraged going forward.

julianoes (approved):
Thanks @DonLakeFlyer, much better than my hotfix!

src/comm/MockLink.cc
DonLakeFlyer: It's to simulate a noisy comm link where the final ack is lost on the wire. QGC handles that case an reports an error back to the handler.

bkueng (commented):
It looks correct.
But I think it breaks the autotune logic: QGC get's MAV_RESULT_IN_PROGRESS, but then no ACK, resulting in a command timeout and `autotune->handleAckError`. Of if `_pollTimer` triggers before the timeout, the command is sent again resulting in the duplicate command logic.

src/Vehicle/Vehicle.h
```diff
@@ -763,21 +763,38 @@ class Vehicle : public FactGroup
...
+    ///     @param ack                  Received COMMAND_ACK
+    ///     @param failureCode          Failure reason. If not MavCmdResultCommandResultOnly only ack.result == MAV_RESULT_FAILED is valid.
+    typedef void (*MavCmdResultHandler)(void* resultHandlerData, int compId, const mavlink_command_ack_t& ack, MavCmdResultFailureCode_t failureCode);
+
+    // Callback info for sendMavCommandWithHandler
+    typedef struct MavCmdAckHandlerInfo_s {
+        MavCmdResultHandler     resultHandler;          ///> nullptr for no handler
+        void*                   resultHandlerData; 
+        MavCmdProgressHandler   progressHandler;
+        void*                   progressHandlerData;    ///> nullptr for no handler
```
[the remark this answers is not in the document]
DonLakeFlyer: Ah, yeah. I'm old so the new C++ stuff doesn't jump to mind immediately :)

DonLakeFlyer:
> But I think it breaks the autotune logic: QGC get's MAV_RESULT_IN_PROGRESS, but then no ACK, resulting in a command timeout and `autotune->handleAckError`. Of if `_pollTimer` triggers before the timeout, the command is sent again resulting in the duplicate command logic.

That goes against mavlink spec doesn't it? Should you always get a final real ack one way of the other (success/fail) to know the command is done?

As far as command retry that is only used for vehicle startup stuff. It's never used for user initiated commands:
```
bool Vehicle::_sendMavCommandShouldRetry(MAV_CMD command)
{
    switch (command) {
#ifdef QT_DEBUG
    // These MockLink command should be retried so we can create unit tests to test retry code
    case MockLink::MAV_CMD_MOCKLINK_ALWAYS_RESULT_ACCEPTED:
    case MockLink::MAV_CMD_MOCKLINK_ALWAYS_RESULT_FAILED:
    case MockLink::MAV_CMD_MOCKLINK_SECOND_ATTEMPT_RESULT_ACCEPTED:
    case MockLink::MAV_CMD_MOCKLINK_SECOND_ATTEMPT_RESULT_FAILED:
    case MockLink::MAV_CMD_MOCKLINK_NO_RESPONSE:
        return true;
#endif

        // In general we should not retry any commands. This is for safety reasons. For example you don't want an ARM command
        // to timeout with no response over a noisy link twice and then suddenly have the third try work 6 seconds later. At that
        // point the user could have walked up to the vehicle to see what is going wrong.
        //
        // We do retry commands which are part of the initial vehicle connect sequence. This makes this process work better over noisy
        // links where commands could be lost. Also these commands tend to just be requesting status so if they end up being delayed
        // there are no safety concerns that could occur.
    case MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES:
    case MAV_CMD_REQUEST_PROTOCOL_VERSION:
    case MAV_CMD_REQUEST_MESSAGE:
    case MAV_CMD_PREFLIGHT_STORAGE:
    case MAV_CMD_RUN_PREARM_CHECKS:
        return true;

    default:
        return false;
    }
}
```

julianoes:
> That goes against mavlink spec doesn't it? Should you always get a final real ack one way of the other (success/fail) to know the command is done?

Correct.

julianoes:
I've tested this with my bench setup where I first discovered the problem and it works!

bkueng:
> That goes against mavlink spec doesn't it? Should you always get a final real ack one way of the other (success/fail) to know the command is done?

Yes. I'm not sure if it was implemented before the MAVLink spec got clarified. We'll need to fix PX4 and think about now to transition. It might be fine to require updating both PX4 and QGC in this case, just the error behavior won't be obvious.

DonLakeFlyer:
Ok, I can hack it to make autotune not screw. Easiest thing for now. Firmware should be fixed though.

DonLakeFlyer:
@bkueng I"m confused as to how the current code (not this pull) works with respect to autotune. Seems to me that it would get the first IN_PROGRESS ack which would in turn remove the command from the list. This means that the autotune code would stop received IN_PROGRESS update through it's handler. Since the handler would have been removed.

Not sure how I can hack this to make it work? The handler needs to stay connected up so in progress still makes it to the autotune code. But when do I remove the auto-tune command from the list?

bkueng:
> Seems to me that it would get the first IN_PROGRESS ack which would in turn remove the command from the list. This means that the autotune code would stop received IN_PROGRESS update through it's handler. Since the handler would have been removed.

Yes exactly.

> The handler needs to stay connected up so in progress still makes it to the autotune code. But when do I remove the auto-tune command from the list?

Not that nice, but maybe add a flag to `MavCommandListEntry_t` to remove it immediately when receiving IN_PROGRESS?

DonLakeFlyer:
> Not that nice, but maybe add a flag to `MavCommandListEntry_t` to remove it immediately when receiving IN_PROGRESS?

I was thinking something like that wouldn't work since Autoune would stop getting updates. But if you are ok with that I can make that change. I'll do something and you can take a look @bkueng

DonLakeFlyer:
@bkueng Can you check this list commit?

bkueng:
Looks good, and I quickly tested it, it works (it can be tested in SITL).

DonLakeFlyer:
I didn't change to using std::function from how the existing code does it since I don't quite have time to deal with that for now. At some point I'll try to come back and update things.

src/Vehicle/Vehicle.cc
hshakula: Can we also agree that restarting the actual timer is more important? Here a copy of the original elapsedTimer is restarted. Because [this](https://github.com/mavlink/qgroundcontrol/blob/39fcc8299ed9824bc972802ec7e678540d374b3d/src/Vehicle/Vehicle.cc#L3280) is a copy.

PR #10855, opened 2023-11-09
Add range checking to Mission Item params

nexton-winjeel:
A number of `MAV_CMD`s have minimum and maximum values for the params. For example, [MAV_CMD_NAV_LOITER_TURNS](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TURNS) has a minimum of 0 for `param1` ("Turns").

This PR adds `"min"` and `"max"` fields to the definitions in `MavCmdInfoCommon.json`, and uses them when setting up the `Fact`s to display when planning a Mission.

### DIFF
--- a/src/MissionManager/MissionCommandUIInfo.h
+++ b/src/MissionManager/MissionCommandUIInfo.h
@@ -55,6 +55,8 @@ public:
     Q_PROPERTY(int          param           READ param          CONSTANT)
     Q_PROPERTY(QString      units           READ units          CONSTANT)
     Q_PROPERTY(bool         nanUnchanged    READ nanUnchanged   CONSTANT)
+    Q_PROPERTY(double       min             READ min            CONSTANT)
+    Q_PROPERTY(double       max             READ max            CONSTANT)
 
     int             decimalPlaces   (void) const { return _decimalPlaces; }
     double          defaultValue    (void) const { return _defaultValue; }
@@ -64,6 +66,8 @@ public:
     int             param           (void) const { return _param; }
     QString         units           (void) const { return _units; }
     bool            nanUnchanged    (void) const { return _nanUnchanged; }
+    double          min             (void) const { return _min; }
+    double          max             (void) const { return _max; }
 
 private:
     int             _decimalPlaces;
@@ -74,6 +78,8 @@ private:
     int             _param;
     QString         _units;
     bool            _nanUnchanged;
+    double          _min;
+    double          _max;
 
     friend class MissionCommandTree;
     friend class MissionCommandUIInfo;
@@ -185,6 +191,8 @@ private:
     static const char* _idJsonKey;
     static const char* _labelJsonKey;
     static const char* _mavCmdInfoJsonKey;
+    static const char* _maxJsonKey;
+    static const char* _minJsonKey;
     static const char* _param1JsonKey;
     static const char* _param2JsonKey;
     static const char* _param3JsonKey;
--- a/src/FactSystem/FactMetaData.cc
+++ b/src/FactSystem/FactMetaData.cc
@@ -153,9 +153,7 @@ FactMetaData::FactMetaData(QObject* parent)
     , _rawDefaultValue      (0)
     , _defaultValueAvailable(false)
     , _rawMax               (_maxForType())
-    , _maxIsDefaultForType  (true)
     , _rawMin               (_minForType())
-    , _minIsDefaultForType  (true)
     , _rawTranslator        (_defaultTranslator)
     , _cookedTranslator     (_defaultTranslator)
     , _vehicleRebootRequired(false)
@@ -177,9 +175,7 @@ FactMetaData::FactMetaData(ValueType_t type, QObject* parent)
     , _rawDefaultValue      (0)
     , _defaultValueAvailable(false)
     , _rawMax               (_maxForType())
-    , _maxIsDefaultForType  (true)
     , _rawMin               (_minForType())
-    , _minIsDefaultForType  (true)
     , _rawTranslator        (_defaultTranslator)
     , _cookedTranslator     (_defaultTranslator)
     , _vehicleRebootRequired(false)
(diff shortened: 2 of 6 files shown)


DonLakeFlyer:
Why the NaN changes to FactMetaData. Is there a usage case for this? In general setting a NaN requires special UI or special code since the user can't enter a NaN into a field through the ui.

nexton-winjeel:
@DonLakeFlyer:
> Why the NaN changes to FactMetaData. Is there a usage case for this? In general setting a NaN requires special UI or special code since the user can't enter a NaN into a field through the ui.

* It's only on the `min` and `max` values of `FactMetaData`.
* It simplifies the handling of invalid values in the JSON -- we don't have to convert `NaN` values when we read them, we can handle them in `FactMetaData`.
* The `NaN` values don't propagate through: passing in a `NaN` will set the `min`/`max` to their default value.
* The specific change is needed so that:
    * the `NaN` isn't accepted due to `isInRawMinLimit()`/`isInRawMaxLimit()` returning `true` for `NaN`; and
    * it doesn't trigger the "Attempt to set min below allowable value ..." warning.

DonLakeFlyer:
Here is how I understand things as they currently stand:
* Only "default" in the json is allowed to be NaN ("null" in the json). Nulls in another other field will generate a type error and the metadata will be defaulted. Any cases other than "default" for NaN values are internal errors in the json. You don't want to hide this when it happens. You make to report it so it can be fixed.
* Facts which support NaN require special UI. Usually a checkbox plus a FactTextField.
* The end user has no capability to enter a NaN using a FactTextField
* Given the above, the only way a NaN could end up making it's way to the setRaw[Min|Max] methods would be from bad C++ code. In that case the current code outputs and error "Attempt to set default value which is outside min/max range" which is correct since this is wrong. Also a case where you want to report he problem so it can be fixed.

So the question is: How are NaN making their way to the setRaw[Min|Max] methods through a correctly written code path?

nexton-winjeel:
> Only "default" in the json is allowed to be NaN ("null" in the json). 

The way the validation works, any field of type `double` (such as these new `min` and `max` fields) is allowed to be set to `null` in the JSON.

> So the question is: How are NaN making their way to the setRaw[Min|Max] methods through a correctly written code path?

OK, now I'm following. 
* I'm defaulting to `NaN` for the `min` and `max` fields if the JSON value couldn't be read.
* I'm used to using `NaN` to signal invalid/unset.

I'm doing this because:
* In the `FactMetaData` `setRaw[Min|Max]` functions, I need a way to re-set the `_[min|max]IsDefaultForType` field, otherwise the min and max values are displayed in the UI for every mission item param.
    * I've chosen to use `NaN` for this, but I could compare against ±`std::numeric_limits<double>::max()`.
* The JSON values are read into a `MissionCmdParamInfo`. 
    * I need a way to signal that the `min` and `max` fields are unset.
    * I've chosen to use a `NaN` for this rather than have an extra field.
* The `MissionCmdParamInfo` is converted to a `FactMetaData` in multiple locations.
    * I've chosen to lower the validation into the `setRaw[Min|Max]` functions, but could validate at each call site instead.

Having said all that, I'll get rid of the `NaN`s by doing the following:
* Set the `min` and `max` fields of `MissionCmdParamInfo` to ±`std::numeric_limits<double>::max()`; and
* Change the validation in the `FactMetaData` `setRaw[Min|Max]` functions to look for the above.

DonLakeFlyer:
> * otherwise the min and max values are displayed in the UI for every mission item param.

Ah, now I understand the problem you are trying to solve. The basic problem is caused by the fact the the min/max stuff and the isDefault ForType implementations keep two separate pieces of information which needs to be kept in sync. This is a crappy implementation. 

The simple way to fix this without needing additional magic inputs to setRawMin/Max is to fix the underlying issue described above:
* FactMetaData changes:
    * Change ```[min|max]IsDefaultForType``` method to determine the result dynamically instead of keeping a separate bool to the side. Just return the comparison of the current value to ```_[min|max|ForType```.
    * Move the code from ```_[min|max]ForType``` into new static public methods: ```[min|max]ForType(ValueType_t type)```. Change existing private methods ```_[min|max]ForType``` to call this. This way you can call these new methods from the ```MissionCommandUIInfo``` code.
    * Leave current ```setRaw[Max|Min]``` as is
 * MissionCommandUIInfo changes:
    * In the constructor initialize your new min/max values to ```FactMetaData::[min|max]ForType(...)```
    * When you read the min/max values from the ui json no need to do anything special in the case where they don't exists in the json since they are already correctly defaulted.

By doing the above the ```isDefaultForType``` will track automatically as you move the min/max values in and out in SimpleMissionItem.

Does that all make sense? It should be pretty straightforward to do it this way.

DonLakeFlyer:
FYI: I'd like to try to get this into the next Stable if possible. That is going to happen in two days.

nexton-winjeel:
@DonLakeFlyer: I've made the suggested changes.

src/MissionManager/MissionCommandUIInfo.h
```diff
@@ -55,6 +55,8 @@ class MissionCmdParamInfo : public QObject {
     Q_PROPERTY(int          param           READ param          CONSTANT)
     Q_PROPERTY(QString      units           READ units          CONSTANT)
     Q_PROPERTY(bool         nanUnchanged    READ nanUnchanged   CONSTANT)
+    Q_PROPERTY(bool         min             READ min            CONSTANT)
```
DonLakeFlyer, line 58: These should be double shouldn't they?

DonLakeFlyer:
Fantastic, thanks!

src/MissionManager/MissionCommandUIInfo.h
nexton-winjeel: Yes, they should. Fixed.

DonLakeFlyer:
@nexton-winjeel Thanks again for this change. This is a nice addition.

PR #10862, opened 2023-11-18
PhotoVideoControl.qml: Fixed TypeError when pressing recording button with no active stream

jnomikos:
A minor bug exists where the record button is enabled even before there is an active stream streaming. 

Hitting the record button before there is an active stream will give this error:
`qrc:/qml/QGroundControl/FlightMap/PhotoVideoControl.qml:104: TypeError: Cannot read property 'capturesPhotos' of null`

My change makes the record button disabled until it is possible to record.

### DIFF
--- a/src/FlightMap/Widgets/PhotoVideoControl.qml
+++ b/src/FlightMap/Widgets/PhotoVideoControl.qml
@@ -101,7 +101,19 @@ Rectangle {
 
     function toggleShooting() {
         console.log("toggleShooting", _anyVideoStreamAvailable)
-        if (_mavlinkCamera && (_mavlinkCamera.capturesVideo || _mavlinkCamera.capturesPhotos) ) {
+
+        // This whole mavlinkCameraCaptureVideoOrPhotos stuff is to work around some strange qml boolean testing 
+        // behavior which wasn't working correctly. This should work:
+        //    if (_mavlinkCamera && (_mavlinkCamera.capturesVideo || _mavlinkCamera.capturesPhotos) ) {
+        // but it doesn't for some strange reason. Hence all the stuff below...
+        var mavlinkCameraCaptureVideoOrPhotos = false
+        if (_mavlinkCamera) {
+            if (_mavlinkCamera.capturesVideo || _mavlinkCamera.capturesPhotos) {
+                mavlinkCameraCaptureVideoOrPhotos = true
+            }
+        }
+        
+        if (mavlinkCameraCaptureVideoOrPhotos) {
             if(_mavlinkCameraInVideoMode) {
                 _mavlinkCamera.toggleVideo()
             } else {


src/FlightMap/Widgets/PhotoVideoControl.qml
```diff
@@ -34,7 +34,7 @@ Rectangle {
 
     // The following properties relate to a simple camera
     property var    _flyViewSettings:                           QGroundControl.settingsManager.flyViewSettings
-    property bool   _simpleCameraAvailable:                     !_mavlinkCamera && _activeVehicle && _flyViewSettings.showSimpleCameraControl.rawValue
+    property bool   _simpleCameraAvailable:                     !_mavlinkCamera && _activeVehicle && _flyViewSettings.showSimpleCameraControl.rawValue && _videoStreamIsStreaming
```
DonLakeFlyer, line 37: This is not a correct fix. This will likely break taking photos when you are in simple camera mode. This change would only allow you to take a photo when video is streaming which doesn't make sense. A simple cameras can do both photos and video.

DonLakeFlyer:
I don't quite understand how this code:
```
if (_mavlinkCamera && (_mavlinkCamera.capturesVideo || _mavlinkCamera.capturesPhotos) ) {
```
can generate that qml error. That would indicate that ```_mavlinkCamera`` is null which should cause the remainder of the if to not be evaluated.

If it's some strange qml wierdness you can split the compound if into two separate if to force the second part to not be evaluated when ```_mavlinkCamera``` is null.

src/FlightMap/Widgets/PhotoVideoControl.qml
jnomikos: Ah my mistake. Did not realize QGroundControl supported taking photos with a simple camera when it is not actively streaming.

jnomikos:
> I don't quite understand how this code:
> 
> ```
> if (_mavlinkCamera && (_mavlinkCamera.capturesVideo || _mavlinkCamera.capturesPhotos) ) {
> ```
> 
> can generate that qml error. That would indicate that ```_mavlinkCamera`` is null which should cause the remainder of the if to not be evaluated.
> 
> If it's some strange qml wierdness you can split the compound if into two separate if to force the second part to not be evaluated when `_mavlinkCamera` is null.

Was definitely some QML weirdness. Problem went away after I did what you suggested.

src/FlightMap/Widgets/PhotoVideoControl.qml
jnomikos: Undone

src/FlightMap/Widgets/PhotoVideoControl.qml
```diff
@@ -101,14 +101,16 @@ Rectangle {
...
                 } else {
-                    _mavlinkCamera.takePhoto()
+                    if(_mavlinkCameraInPhotoMode && !_mavlinkCameraPhotoCaptureIsIdle && _mavlinkCameraElapsedMode) {
+                        _mavlinkCamera.stopTakePhoto()
+                    } else {
+                        _mavlinkCamera.takePhoto()
+                    }
                 }
             }
         } else if (_onlySimpleCameraAvailable || (_simpleCameraAvailable && _anyVideoStreamAvailable && _videoStreamInPhotoMode && !videoGrabRadio.checked)) {
```
DonLakeFlyer, line 116: Hmmm. The above changes aren't going to fall through to the other elses the same way they used to. To avoid this I would do the following with a comment to explain with it's so crazy:
```
// This whole mavlinkCameraCaptureVideoOrPhotos stuff is to work around some strange qml boolean testing // behavior which wasn't working correctly. This should work:
//    if (_mavlinkCamera && (_mavlinkCamera.capturesVideo || _mavlinkCamera.capturesPhotos) ) {
// but it doesn't for some strange reason. Hence all the stuff below...

var mavlinkCameraCaptureVideoOrPhotos = false
if (_mavlinkCamera) {
    if (_mavlinkCamera.capturesVideo || _mavlinkCamera.capturesPhotos) {
        mavlinkCameraCaptureVideoOrPhotos = true
    }
}
if (mavlinkCameraCaptureVideoOrPhotos) {
} else if (...)
```

DonLakeFlyer:
FYI: I'd like to try to get this into the next Stable if possible. That is going to happen in two days.

src/FlightMap/Widgets/PhotoVideoControl.qml
jnomikos: I like this idea. Pushed up changes.

Still super weird that qml is behaving this way. I wonder if this is a bug in 5.15.2 or just something that is happening in this particular area.

jnomikos:
> FYI: I'd like to try to get this into the next Stable if possible. That is going to happen in two days.

Done

DonLakeFlyer:
Great, thanks for the fix

Issue #10861, opened 2023-11-18
Windows Builds Missing Video Streaming


AlexKlimaj:
The latest windows daily builds are missing video streaming.

DonLakeFlyer:
@AlexKlimaj I'll close this once you do the final verification.

AlexKlimaj:
Daily build is working now

PR #10864, opened 2023-11-19
Windows: Look for gstreamer on D drive as well as C

DonLakeFlyer:
* Fix for #10861
* Something changed in the github actions windows runner which places everything on the D drive instead of C


AlexKlimaj:
Daily build is working now

Issue #10591, opened 2023-02-21
Offline map export does not add proper filename extension


AndKe:
## Expected Behavior
exported maps should be saved as filename.qgctiledb

## Current Behavior
saved as filename

## Steps to Reproduce:
1.export map(s)  - specify filename, but dont type the extension

## System Information
- Operating System: ubuntu 22.10
- QGC Version: current master

jnomikos:
I was able to recreate this on Ubuntu 22.04

DonLakeFlyer:
@AndKe Thanks for the bug report

PR #10867, opened 2023-11-23
Fixed exporting tile sets not saving the files with the .qgctiledb extension

jnomikos:
Problem mentioned here: https://github.com/mavlink/qgroundcontrol/issues/10591

Exporting tile sets does not save them with any file extension if none are explicitly defined when writing the file name. This patch fixes that problem.

### DIFF
--- a/src/QtLocationPlugin/QMLControl/QGCMapEngineManager.h
+++ b/src/QtLocationPlugin/QMLControl/QGCMapEngineManager.h
@@ -37,26 +37,27 @@ public:
     };
     Q_ENUM(ImportAction)
 
-    Q_PROPERTY(quint64              tileCount       READ    tileCount       NOTIFY tileCountChanged)
-    Q_PROPERTY(QString              tileCountStr    READ    tileCountStr    NOTIFY tileCountChanged)
-    Q_PROPERTY(quint64              tileSize        READ    tileSize        NOTIFY tileSizeChanged)
-    Q_PROPERTY(QString              tileSizeStr     READ    tileSizeStr     NOTIFY tileSizeChanged)
-    Q_PROPERTY(QmlObjectListModel*  tileSets        READ    tileSets        NOTIFY tileSetsChanged)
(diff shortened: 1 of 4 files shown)


src/QtLocationPlugin/QMLControl/QGCMapEngineManager.h
```diff
@@ -37,26 +37,27 @@ class QGCMapEngineManager : public QGCTool
     };
     Q_ENUM(ImportAction)
 
-    Q_PROPERTY(quint64              tileCount       READ    tileCount       NOTIFY tileCountChanged)
```
zdanek, line 40: please don't change indentation. Although it might not be one's preference (neither mine for example), let's not change it without a real reason. It's harder to read changes/merge the code.

zdanek (changes requested):
looks good for me, but please revert back unnecessary indentation changes.

src/QtLocationPlugin/QMLControl/QGCMapEngineManager.h
jnomikos: Sounds good to me, but just be aware that my change there will not align with the other Q_PROPERTYs.

If we want to keep everything aligned, I could throw my indentation changes in a separate commit.

DonLakeFlyer:
I need to look at this. The file dialog should automatically add the extension if not specified. I think the problem is actually how the file dialog is being  used. That said, the changes to AppSettings to add the file extension definition is good/needed. But the change to add ```QString QGCMapEngineManager::tilesetFileExtension(void) const``` is not needed since this is exposed by AppSettings. No need to put it in two places. I'l likely take this a make a new pull once I figure out what is going on.

DonLakeFlyer:
Thanks for the pull. I reworked slightly to fix the GCFileDialog usage.

jnomikos:
> Thanks for the pull. I reworked slightly to fix the GCFileDialog usage.

Thank you so much!

PR #10872, opened 2023-11-25
Add battery voltage in Toolbar's Instrument Panel

zdanek:
For most of pilots it's more beneficial to know current voltage. They know their batteries / hardware the best so they know how to react on battery voltage more than on percentage only.

Those are  examples, with two batteries, second one has voltage unavailable, but adds a description instead.
image

image

image

image

### DIFF
--- a/src/Vehicle/CMakeLists.txt
+++ b/src/Vehicle/CMakeLists.txt
@@ -98,8 +98,12 @@ add_library(Vehicle
 	VehicleVibrationFactGroup.h
 	VehicleWindFactGroup.cc
 	VehicleWindFactGroup.h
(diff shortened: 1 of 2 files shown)


zdanek:
Btw, looking at those examples, bear in mind that they were based on SITL, that simulates percentage drop, but not voltage. So don't look for an error in displaying save voltage value at every screen even though percentage drops :)

zdanek:
If you need bigger screen slice for better understanding
image

DonLakeFlyer:
Could you horizontal center the multiple text parts in the column? I think it would look nicer that way.

DonLakeFlyer:
I integrated this into #10909

PR #10875, opened 2023-11-27
several fixes

Davidsastresas:
This PR addresses random fixes not significant enough as for having its own PR.

Some comments:

- The commits over ToolStripAction and ToolStripHoverButton don't have any application on current codebase. I ported that from a custom build, and it allows to set buttons in toolstrip with bi-color icons, useful to only highlight one of the colors. The properties about non exclusive allows toolstrips where one of the tools might be permanently enabled, and the property toolstripIndex is a helper to know the "id" of that action in the toolstrip.This is being used on the ongoing work of improving gimbal development: https://github.com/mavlink/qgroundcontrol/pull/10667

- The commits about get offline camera definition file is for custom builds that could embed a camera definition file within QGC resources, so QGC doesn't need internet access for first handshake with a camera. It is not used right now on main QGC codebase either.

- Sendmavdelayed is a helper function to send a delayed mav command through vehicle. Also not used in main QGC codebase, but it is a handy function to have in custom builds.


src/AutoPilotPlugins/APM/APMCameraComponent.cc
```diff
@@ -62,5 +62,5 @@ QUrl APMCameraComponent::setupSource(void) const
 
 QUrl APMCameraComponent::summaryQmlSource(void) const
 {
-    return QUrl::fromUserInput(QStringLiteral("qrc:/qml/APMCameraComponentSummary.qml"));
```
DonLakeFlyer, line 65: How come this is removed?

DonLakeFlyer:
Can you look at the conflicts?

src/AutoPilotPlugins/APM/APMCameraComponent.cc
Davidsastresas: A lot of parameters are outdated there, so it isn't functional right now for the most of it for Ardupilot, so I thought it would be better disabled and force the users to do it by parameters. As part of https://github.com/mavlink/qgroundcontrol/pull/10667 we will go over that menu as well, and leave it updated for the current Ardupilot setup. Is that fine?

Davidsastresas:
@DonLakeFlyer I fixed the conflicts of the latest commit. It turns out after what you did there it wasn't necessary anymore so I dropped it, thanks!

I also rebased to current master. If you took a look at the rest of the commits and they are good, I will merge when CI finishes if that is fine! Thanks!

src/AutoPilotPlugins/APM/APMCameraComponent.cc
DonLakeFlyer: Sure

qgroundcontrol.pro
```diff
@@ -70,9 +70,21 @@ QGC_APP_COPYRIGHT   = "Copyright (C) 2019 QGroundControl Development Team. All r
...
+    } else {
+        QGC_INSTALLER_HEADER_BITMAP = "$$SOURCE_DIR\\deploy\\windows\\installheader.bmp"
+    }
+
+    exists($$PWD/custom/deploy/windows/WindowsQGC.ico) {
+        message("Using custom WindowsQGC.ico")
+        QGC_INSTALLER_ICON = "$$SOURCE_DIR\\custom\\deploy\\windows\\WindowsQGC.ico"
+    } else {
+        QGC_INSTALLER_ICON = "$$SOURCE_DIR\\deploy\\windows\\WindowsQGC.ico"
+    }
```
DonLakeFlyer, line 87: Unless I'm missing something this is not needed. The way you override these sorts of things is by just setting them in custom.pri which is included after this. That will then in turn overwrite the values with the new ones. 

Otherwise you need these exists things all over the place on every single possible config setting.

src/QmlControls/QGCButton.qml
```diff
@@ -7,7 +7,7 @@ import QGroundControl.ScreenTools
 
 Button {
     id:             control
-    hoverEnabled:   true
+    hoverEnabled:   !ScreenTools.isMobile
```
DonLakeFlyer, line 10: Is there some reason for this? Mobile with a mouse?

src/ui/MainRootWindow.qml
```diff
@@ -497,7 +497,6 @@ ApplicationWindow {
                 width:              (backTextLabel.x + backTextLabel.width) - backIcon.x
                 onClicked: {
                     toolDrawer.visible      = false
-                    toolDrawer.toolSource   = ""
```
DonLakeFlyer, line 500: Why is this removed? This means the drawer won't be freed from memory.

Davidsastresas: You are totally right, sorry. I redid this using a timer.

src/QmlControls/QGCButton.qml
Davidsastresas: No no, this disables hover on mobile, so it doesn't have nothing to do with mouse on mobile

The issue is that for some reason in mobile the buttons keep "hovered" after you click them, so it creates a lot of inconsistencies around the UI. I am not sure of the "deep" reason, but in any case the concept of "hover" for touch based controls doesn't make sense, so I thought it would be good to disable hover on mobile. 

After this change, the buttons no longer keep "hovered" after clicking them in Android.

qgroundcontrol.pro
Davidsastresas: I haven't tested, but based on what you said this makes sense. For some reason I thought custom.pri was "evaluated" before this. I am dropping this one. Thank you!

Davidsastresas:
Thanks for the review @DonLakeFlyer.

About the MainRootWindow commit, The problem is, when you open this drawer, for instance going to vehicle setup, and then you click the button that goes back to fly view, until the drawer hides, as we unload it, we see a "blank" drawer, which is very noticeable, specially when the dark theme is used. You were right, the previous approach doesn't unload it, sorry I missed it, so I redid that using a timer that unloads it after 500ms ( I wasn't able to locate the actual precise time the drawer takes to appear/dissapear ).

About the commit disabling hover on mobile, The issue is that for some reason in mobile the buttons keep being "hovered" after you click them, so it creates a lot of color inconsistencies around the UI. I am not sure of the "deep" reason, but in any case the concept of "hover" for touch based controls doesn't make sense, so I thought it would be good to disable hover on mobile. This effectively fixes the issue, and we no longer have "ghost hoverings" in mobile.

About the custom icon in qgroundcontrol.pro, I haven't tested, but based on what you said this makes sense. For some reason I thought custom.pri was "evaluated" before this. I am dropping these two. Thank you!

qgroundcontrol.pro
Davidsastresas: I did a quick test, overriding that from custom.pri, and for some reason is not working. I am trying to debug it now, but I thought I could post here the update.

Davidsastresas:
I tested after dropping the commits in qgroundcontrol.pro, and working exclusively within custom.pri, as it is now, I am not able to override the installheader.bmp, nor the app icon. The installer Icon works though! 

The suggested changes in qgroundcontrol.pro work. But I understand it is not ideal and it would be better to understand why it is not working. So I wanted to report this even though I dropped them.

DonLakeFlyer:
> The problem is, when you open this drawer, for instance going to vehicle setup, and then you click the button that goes back to fly view, until the drawer hides, as we unload it, we see a "blank" drawer, which is very noticeable, specially when the dark theme is used.

Looking at that Drawer code it's little odd that it uses visibility to show instead of open/close call. That said I think visibility changed ends up being the same thing though. Given that it means that you should get a closed signal. So unload the Loader source onClosed instead. Then it should happen after the drawer hides I think. There are couple places the source is cleared (there is a view close function someplace, maybe more) in MainRootWindow make sure you remove them all when you switch to onClosed. Also if it works add a note as to why this is in onClosed for future reference.

src/QmlControls/QGCButton.qml
DonLakeFlyer: Hmm, I had noticed that hover thing as well but hadn't gotten around to figure it out. So this is fine then. I think there are other place where hover is on in other controls. Can you search for that and make sure there aren't other places that should change as well.

DonLakeFlyer:
> as it is now, I am not able to override the installheader.bmp, nor the app icon.

That doesn't make sense since they are all in the same place. Can you test by adding message output in qgroundcontrol.pro after custom.pri is include to see if they end up being changed. You should be able to debug why this is not working.

Davidsastresas:
> > The problem is, when you open this drawer, for instance going to vehicle setup, and then you click the button that goes back to fly view, until the drawer hides, as we unload it, we see a "blank" drawer, which is very noticeable, specially when the dark theme is used.
> 
> Looking at that Drawer code it's little odd that it uses visibility to show instead of open/close call. That said I think visibility changed ends up being the same thing though. Given that it means that you should get a closed signal. So unload the Loader source onClosed instead. Then it should happen after the drawer hides I think. There are couple places the source is cleared (there is a view close function someplace, maybe more) in MainRootWindow make sure you remove them all when you switch to onClosed. Also if it works add a note as to why this is in onClosed for future reference.

I hadn't thought of that! It is the elegant solution I was looking for, thank you. It works. The other places you thought this could be happening were in viewSwitch() function, I checked and it seems it is unused, so I did another commit removing it, and I did push force so it should be on order now.

>> That doesn't make sense since they are all in the same place. Can you test by adding message output in qgroundcontrol.pro after custom.pri is include to see if they end up being changed. You should be able to debug why this is not working.

I know, it doesn't make sense, but I just tested and it does not work. This reminds me of that situation for overriding some Android stuff we dealt with in the past, you ended up fixing it but if I recall correctly it wasn't obvious.

I mean to keep debugging it at some point, but I should setup my windows partition first to build QGC there, as right now I am relying on CI and it is just too slow to work on it efficiently. In any case, it should not be a priority right now, I dropped those commits, we can go back to this in the future.

Let me know if that is fine. Thanks!

Davidsastresas:
@DonLakeFlyer , is it fine for me to merge this now? Thanks!

DonLakeFlyer:
Yup

PR #10879, opened 2023-11-29
[ObstacleDistanceOverlay] Fix 90 degree offset 

dakejahl:
Fixes issue where obstacle distances are displayed with a +90 degree offset. This is due to an error in the coordinate system math along with an error in calculating the degrees for the current distance index. The yaw angle theta is measured CW from North and therefore `r * cos(theta)` is negative Y axis and `r * sin(theta)` is positive X.

**Qt docs**
> The x values increase to the right and the y values increase downwards.
image

There's more still that needs to be done. The ObstacleDistanceOverlayVideo.qml probably needs this fix. There's also the VehicleObjectAvoidance class which seems to be entirely unused other than the `distances` array. This feature is currently broken in PX4 and I don't think anyone has used it in years. I want to resurrect the feature and make it all work again.

### DIFF
--- a/src/FlightDisplay/ObstacleDistanceOverlay.qml
+++ b/src/FlightDisplay/ObstacleDistanceOverlay.qml
@@ -36,7 +36,10 @@ Canvas {
 
     // Converts degrees to index from ranges
     function rangeIdx(deg, increment, offset, len, heading) {
-        const i = (360 - heading + deg - offset) / increment
+        var degrees = deg + offset - heading
+        if (degrees > 360) degrees = degrees - 360
+        if (degrees < 0) degrees = degrees + 360;
+        const i = degrees / increment
         return (len + Math.ceil(i)) % len
     }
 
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
@dakejahl Any updates on this. It's still marked Draft.

dakejahl:
No not yet, I haven't come back to working on this. The first step is to get the feature working again in PX4.

DonLakeFlyer:
@dakejahl Updates?

dakejahl:
@Claudio-Chies has done a ton of work lately with avoidance in PX4 so he might know better than I

Claudio-Chies:
@dakejahl wow awesome! This was on my to-do, didn't know you already solved it! Thanks a lot!
image
looks great, only the video overlay is still offset, as you mentioned above.

Claudio-Chies:
image
Added a Fix for the Video Overlay, basically the problem is that the degrees are in FRD frame starting from Front rotating around Down, but plotting is done in XY Starting at the X line, clockwise, 
so I'm subtracting Pi/2 to fix the issue.

DonLakeFlyer (approved):

PR #10882, opened 2023-11-30
User Guide and DevGuide source migration to Stable v4_3

hamishwillee:
This migrates the source for both QGC user and QGCdev guide into the docs repo.

The sources are migrating to Vitepress, which is a pretty good replacement for gitbook.

There is some work to be done:
1. Google analytics - I need a GA4 key. Need to get off @mrpollo 
1. ~Compress and lower case images~
1. ~Currently working through this updating notes, tips, warnings to the new format.~
1. ~Prettier~
1. ~Lower case file names~
1. ~Set up deployment~
1. Merge this PR

After:
1. Copy commit to master
1. Then do crowdin integration.
1. SOMEONE ELSE should turn off all the extra tests if there are docs-only changes.

@DonLakeFlyer I'm now away until Weds, when I will continue this. Suggest if you want to make changes to docs you make to this branch, but I'd pause it if I was you. 

Also take no new changes from Crowdin or to the other repos. We should lock them for the next week.


DonLakeFlyer:
> Also take no new changes from Crowdin

You mean on the Stable branch or all branches?

hamishwillee:
Any of the **docs** branches - which only have master for translation. QGC itself should continue to take translations as per usual. Hope that answers. I just don't want to have to keep tracking new translations into this text for a week or so.

hamishwillee:
@DonLakeFlyer This "should" be ready to go in. It touches only 2 little parts of the existing setup:
1. Workflows for releases are not run if changes are ONLY in docs or workflow actions. 
2.  The gitignore  doesn't pick up the build files etc from docs toolchain.

There are some "flaws". The translations don't exists all for all files, so you might get a 404 in some cases. But the English version should be solid. I will fix the translated version in master.

Otherwise it is just a big cleanup of docs. I'm going to push when tests pass because I'm fairly confident it won't break anything.

hamishwillee:
@DonLakeFlyer Seems to have worked quite well. YOu can check out the version here: https://docs.qgroundcontrol.com/Stable_V4.3/en/

PR #10876, opened 2023-11-27
Move to Qt 6.6.1

DonLakeFlyer:
I'm putting this up here for folks that want to follow along. I still have a long list of ToDo's to fix up and debugging with respect to Qml changes. But at this point it does boot and connect to vehicles. You'll likely stumble onto parts of the UI which are screwed up or don't work at all. There is a crash somewhere in the bowels of the mapping engine which happens at boot. Happens about 30% of the time, just try again (and again) and it will boot.

As it stands now you need to qmake with `CONFIG+=DISABLE_VIDEOSTREAMING CONFIG+=QT6_DISABLE_DNSENGINE`. The `qmdnsengine` submodule needs cleaning up as does `qmlglsink`. Unfortunately I've never been in any of that code so hopefully other folks can help out to fix these up.

PR #10886, opened 2023-12-02
Move to Qt 6.6.1

DonLakeFlyer:
This is a replacement for #10876. Git wasn't able to make sense out of cleaning up so many commits though interactive rebase. So I had to instead create a diff and recreate the branch with new simple commits.


dayjaby:
@DonLakeFlyer Can you also please update QT_VERSION in deploy/docker?

PR #10909, opened 2023-12-07
Major rework on toolbar indicator dropdowns

DonLakeFlyer:
This is the first step of major ui rework described here: https://github.com/mavlink/qgroundcontrol/issues/10617

The goal of these change are to prevent the user needing to wade through the myriad of vehicle setup and app settings pages once they have reached the stage where they are simply flying an already configured vehicle. Say a commercially purchased vehicle or a DIY vehicle which has already completed all the airframe, radio/sensor cal, flight mode setup and so forth steps. And now you are just flying to vehicle to do work.

The initial dropdown for each toolbar button continues to have the pervious information in the dropdown. But you can now expand them which will show vehicle settings which you may use during every day flight. Some of these settings come from vehicle parameters and some may come from QGC App settings. Within this expanded set of information the FirmwarePlugin has control of part of the page. Such that those settings are firmware specific as well as providing the ability for custom builds to override and customize.

As it currently stands I have implemented the first cut at the PX4 version of these settings. Consider these choices to be a WIP. I'd like to gather input as to what folks think should be made available and not. The ArduPilot ones are NYI and will not show parameter based settings in the expanded drop-down yet. Looking for folks to help fill out the ArduPilot side of things.

All toolbar dropdowns are anchored to the left of the screen and the Q dropdown works just like the other tool dropdowns.
Screenshot 2023-12-07 at 11 42 15 AM

The main status indicator in regular mode:
Screenshot 2023-12-07 at 11 42 24 AM

Expanded mode. Here you can see that the first set of settings (failsafe, data link) are parameter based settings and it also provides access to the full vehicle setup pages.
Screenshot 2023-12-07 at 11 42 35 AM

Here's the rest of the regular and expanded drop downs:
Screenshot 2023-12-07 at 11 42 52 AM
Screenshot 2023-12-07 at 11 43 01 AM
 
The flight mode expanded section also provides new support for hiding flight modes from usage. There is a default set for each firmware type which automatically stars out as hidden:
Screenshot 2023-12-07 at 11 43 10 AM
Screenshot 2023-12-07 at 11 43 19 AM
Screenshot 2023-12-07 at 11 43 26 AM
Screenshot 2023-12-07 at 11 43 33 AM
Screenshot 2023-12-07 at 11 43 42 AM

### DIFF
--- a/.github/workflows/linux_debug.yml
+++ b/.github/workflows/linux_debug.yml
@@ -44,6 +44,9 @@ jobs:
           modules:      qtcharts qtlocation qtpositioning qtspeech qt5compat qtmultimedia qtserialport qtimageformats qtshadertools
           setup-python: true
 
+      - name: Install github environment dependencies for unit test running
+        run:  sudo apt-get install libxcb-xinerama0
+
       - name: Install QGC source dependencies
         run:  sudo apt-get install -y libsdl2-dev
 
--- a/custom-example/qgroundcontrol.qrc
+++ b/custom-example/qgroundcontrol.qrc
@@ -4,7 +4,6 @@
 	</qresource>
 	<qresource prefix="/toolbar">
 		<file alias="ArmedIndicator.qml">../src/ui/toolbar/ArmedIndicator.qml</file>
-		<file alias="BatteryIndicator.qml">../src/ui/toolbar/BatteryIndicator.qml</file>
 		<file alias="GPSIndicator.qml">../src/ui/toolbar/GPSIndicator.qml</file>
 		<file alias="GPSRTKIndicator.qml">../src/ui/toolbar/GPSRTKIndicator.qml</file>
 		<file alias="JoystickIndicator.qml">../src/ui/toolbar/JoystickIndicator.qml</file>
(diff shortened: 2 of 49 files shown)


zdanek:
Can I go first with my voltage added to battery indicator, please?
So I will not step at your work, and vice versa. I've added also batteries on Multivehilce indicators on the right. Maybe that would be better PR with both changes. On top of that you could rework UI as you wish. 
It would be pity to loose my work on this.

Davidsastresas:
You finished this! :) It looks great, thank you!

DonLakeFlyer:
This last commit is a modified replacement for #10872. The default is to show percentage like it always have. But you now have the option to show Percentage, Voltage or both. I didn't want to just show both only all the time because it makes the text really small and I think many PX4 folks determine with percentage only.

Screenshot 2023-12-14 at 10 40 53 AM

Screenshot 2023-12-14 at 10 41 06 AM

DonLakeFlyer:
After this passes CI I'm going to merge

s-lisovenko:
@DonLakeFlyer i found some places with old style (Qt5) qml imports. Could you check?

PR #10912, opened 2023-12-07
Fix QGCSwitch  and ClickableColor after migration to Qt6

s-lisovenko:
After migration to Qt6 these controls were broken

### DIFF
--- a/src/QmlControls/QGCSwitch.qml
+++ b/src/QmlControls/QGCSwitch.qml
@@ -9,20 +9,42 @@
 
 import QtQuick
 import QtQuick.Controls
-import QtQuick.Controls
 
 import QGroundControl.Palette
+import QGroundControl.Controls
 import QGroundControl.ScreenTools
 
 Switch {
-    id: _root
-    QGCPalette { id:qgcPal; colorGroupEnabled: true }
+    id: control
+
+    readonly property int _radius: 3
+
+    QGCPalette {
+        id:                 qgcPal
+        colorGroupEnabled:  true
+    }
+
+    contentItem: QGCLabel {
+        text:               control.text
+        verticalAlignment:  Text.AlignVCenter
+        rightPadding:       control.indicator.width + control.spacing
+    }
+
     indicator: Rectangle {
-        implicitWidth:  ScreenTools.defaultFontPixelWidth * 6
-        implicitHeight: ScreenTools.defaultFontPixelHeight
-        color:          (control.checked && control.enabled) ? qgcPal.colorGreen : qgcPal.colorGrey
-        radius:         3
-        border.color:   qgcPal.button
-        border.width:   1
+        implicitWidth: knob.width * 2
+        implicitHeight: knob.height
+        x: control.width - width - control.rightPadding
+        y: parent.height / 2 - height / 2
+        radius: knob.radius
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Doesn't Switch need to be changed to backgroundItem and contentItem: https://doc.qt.io/Qt-6/qtquickcontrols-customize.html#customizing-switch

s-lisovenko:
> Doesn't Switch need to be changed to backgroundItem and contentItem: https://doc.qt.io/Qt-6/qtquickcontrols-customize.html#customizing-switch

Do you mean using SwitchDelegate instead switch? This is what i see right now with switch.
https://github.com/mavlink/qgroundcontrol/assets/2522054/7464fc98-5bb9-4df3-8602-cce2a0df2574

s-lisovenko:
Delete duplicate imports for QGCSwitch

DonLakeFlyer:
> Do you mean using SwitchDelegate instead switch?

No. In your change it look like you have the the first Rectangle which is actually the background for the control. This should be in the backgroundItem property. And then the indicator is a child of that first Rectangle. This part should be the contentItem. See the link I included on how to style the Switch control in Qt6.

s-lisovenko:
Add background item to the switch control

s-lisovenko:
> > Do you mean using SwitchDelegate instead switch?
> 
> No. In your change it look like you have the the first Rectangle which is actually the background for the control. This should be in the backgroundItem property. And then the indicator is a child of that first Rectangle. This part should be the contentItem. See the link I included on how to style the Switch control in Qt6.

Added background item. I don't think the content item is needed here. So we have separate background and indicator. Please check again

DonLakeFlyer:
> I don't think the content item is needed here.

To be completely correct it does. The previous implementation didn't have it which was wrong. Without the the Text will not the themed correctly to the QGC palette.

s-lisovenko:
> To be completely correct it does. The previous implementation didn't have it which was wrong. Without it the Text will not the themed correctly to the QGC palette.

Sorry I think I misunderstood you. Do yo want to display text information next to a switch? now it's just a stylized switch as in  before QGC implementation? Perhaps you can fix this according to your or the core team's vision? But I think that having the option of a simple switch without text should be left as well.

DonLakeFlyer:
> now it's just a stylized switch as in before QGC implementation

The previous version supported text as well (incorrectly though), since it used the regular Qml Switch as the base control. If you didn't want text you just didn't specify it.

s-lisovenko:
Screenshot from 2024-01-27 21-42-45

Added the ability to display text to QGCSwitch
@DonLakeFlyer please take a look.

src/QmlControls/QGCSwitch.qml
```diff
@@ -9,20 +9,43 @@
...
+    id: control
+
+    readonly property int _radius: 3
+
+    QGCPalette {
+        id:                 qgcPal
+        colorGroupEnabled:  true
+    }
+
+    contentItem: Text {
```
DonLakeFlyer, line 26: This should just be QGCLabel. That way it will automatically follow any styling changes.

s-lisovenko: Yes, changed

DonLakeFlyer:
Thanks for the fix

Issue #10887, opened 2023-12-02
QGC Daily: Moved to Qt version 6.6.3 - Status


DonLakeFlyer:
How you can help:
* Try out a daily build and report issues you find making sure to label the Issue with the "Qt6" label.
* Recommendation is to try your scenarios in SITL first before using QGC Daily on a real vehicle flight
* Join the discussion on the QGC discord channel for questions. Don't clutter GitHub Issues with questions.
* If you can help with QGC development ask on QGC discord channel what remaining work you could help out with. Simplest thing is to find a FIXME-QT6 marker (see below) and fix it. Or find a known issue and fix it.

Current status:
* Windows, Linux, Mac and Android builds are available from the normal Daily Build links on qgrouncontrol.com
* You are likely to run into quirkiness/bugs in the UI. It is going to take some time to find all the missing changes to the Qml UI code in QGC. 
* The required changes to Qml although mostly mechanical in nature are not 100% discoverable through just code inspection. It's going to take running all the parts of the ui and verifying it works correctly to find the remaining needed changes.
* The C++ code should be very close to 100% updated. Those issues show up at compile time as opposed to run time.
   * The largest changes were to the Map plugin which download tiles. So there may still be problems to wring out there. 
* The majority of the top level UI with respect to Vehicle Setup and flight of PX4 and ArduPilot vehicles has been tested to verify the basics are working correctly.
* Plan View has had less focus and is likely to be more flaky than the rest of QGC.

Known Issues:
* Linux 
  * You must manually install `sudo apt install libxcb-cursor0` yourself for the Appimage to boot. The AppImage generation script needs to be updated to fix this.
* Android
  * Joystick, support is not working
* Video is currently disabled in all OS types. Video support is still a WIP.
* Unsure where custom-example stands. Likely needs testing/updates.
* The code is also peppered with FIXME-QT6 comments which marks parts which still need work but at this point don't prevent QGC working at  high level.

Building from master:
* Docs are not yet updated. That will come in the future.
* Qt version you need is 6.6.1
* CI builds are still using qmake for build. So that is known to be in a working state. Transition to make for CI will start soonish.
* Here is a screen shot of the modules needed from Mac OSX. You should be able to figure what you need in your OS from here:
<img width="437" alt="Screenshot 2023-12-01 at 6 45 33 PM" src="https://github.com/mavlink/qgroundcontrol/assets/5876851/1a614e54-9484-4de8-9cc4-79bd5be5007c">

[...]

PR #10915, opened 2023-12-08
Migrate GstGLVideoItem to Qt6

zdanek:
In particular
    * uses new Qt6 qmlsink gstreamer plugin code (switch to new submodule repo)
    * removes unnecessary dependency of QGC to VideoReceiverApp
    * removes dependency of VideoReceiver to Settings submodule (uses only enum from Settings)
    * removes registering of GstGLVideoItem in GStreamer.cc (it's registered in own code)
    * minor cleanup and fixes in VideoReceiverApp
    * fixes dragging map when resizing/clicking on Pip Window

Part of #10887


julianoes:
Thanks for this PR! It works nicely for me, I'm on Linux Mint on top of Ubuntu 22.04.

julianoes:
@zdanek I've rebased this PR in https://github.com/mavlink/qgroundcontrol/pull/10949 and fixed the Android build.

I couldn't push to this PR as you probably don't have the checkbox checked to allow others to push to your fork/branch. If you want, you can just cherry-pick my commits into this PR, or we can merge mine instead. Up to you :man_shrugging:.

julianoes:
@zdanek :wave:

zdanek:
> @zdanek :wave: 
@julianoes sorry, I will do this today afternoon. I was a bit overwhelmed by everyday job, so I had no energy to sit on this in the evenings 🙂

zdanek:
@julianoes Thanks for pushing Android further. All builds pass :)

.gitmodules
```diff
@@ -9,7 +9,7 @@
 	url = https://github.com/KDAB/android_openssl
 [submodule "libs/qmlglsink/gst-plugins-good"]
 	path = libs/qmlglsink/gst-plugins-good
-	url = https://github.com/mavlink/gst-plugins-good.git
```
julianoes, line 12: Do we need this, or can we update the mavlink one with the changes required?

julianoes (changes requested):
Good to go except the submodule change.

.gitmodules
zdanek: It would be great to merge my changes to `mavlink/gst-plugins-good`
I changed to mine, to have a quick path fix my fixes. But, bear in mind that not so long ago it was patricelectric's repo here :) 
So, shall I create a PR to  `mavlink/gst-plugins-good`? Will you or sbd else approve it quite fast?
Shall we wait with current PR until new PR into  `mavlink/gst-plugins-good` goes through?

Maybe we could push with my submodule, until I update  `mavlink/gst-plugins-good`?
This would be something that opens possibility for others to test video stream for bugs. Also non-linux platforms are still do be fixed. I can take Windows, and then eventually Android. I have no Apple hardware (both macbook nor iphone)

julianoes: I should be able to approve and merge it, yes.

zdanek: @julianoes please create a new branch in `mavlink/gst-plugins-good`, it might be for example `qgroundcontrol_qt6` or anythyng else, like `qt6_qmlglsink` 
I have to create PR to existing branch. 
`qgroundcontrol` cannot be upgraded to qt6, as it's still  used by qt5, QGC 4.x version. We cannot break it.

julianoes: Although, as long as the submodule points to the previous commit it should be fine.

julianoes: Ah, unless you want to branch from upstream.

Anyway, I created a branch for you:
https://github.com/mavlink/gst-plugins-good/tree/qgroundcontrol-qt6

And I'll give you permission to the repo, so you can directly push to that one.

DonLakeFlyer: > `` cannot be upgraded to qt6, as it's still used by qt5, QGC 4.x version. We cannot break it.

You shouldn't branch gst-plugins-good. The submodules in the Stable branches of QGC will stay pointed to the place before any mods to plugins-good were made for Qt6. Hence they will build file even though plugins-good has moved on.

You should instead tag plugins-good with a tag like "pre-qt6". Then in the unlikely event that a Stable build needs a fix in plugins-good, that is when plugins-good can be branched from that specific  tag. And then the Stable builds would reference that branch which has the fixes.

zdanek: @DonLakeFlyer  you're right, this is fine, but I'm not comfortable with dangling references. I'd like to have branches, as we have branches with current and stable release.

julianoes (approved):

DonLakeFlyer:
@zdanek @julianoes Thank you guys so, so much for this!

DonLakeFlyer:
Should I merge this now?

PR #10920, opened 2023-12-12
Add FMU V6XRT Board Info

vincentpoont2:
Vendor ID & Product ID is from [fmu v6xrt nsh file](https://github.com/PX4/PX4-Autopilot/tree/main/boards/px4/fmu-v6xrt/nuttx-config/nsh)


vincentpoont2:
@mrpollo @dk7xe I think this should allow user to flash master FW online via QGC directly. 

VendorID & ProductID is from [here](https://github.com/PX4/PX4-Autopilot/blob/ece60b6165753c26f8cbc24cd6b2b2918ef8d8fe/boards/px4/fmu-v6xrt/nuttx-config/nsh/defconfig#L43)

PR #10846, opened 2023-10-30
Hardware: Add support for Pixhawk FMUv6XRT

mrpollo:
This adds support for the new Pixhawk FMUv6XRT Open Standard based on NXP Semiconductor MCU, please note this is using the Dronecode vendorID

cc @davids5 @dk7xe

PR #10921, opened 2023-12-12
Correct FMUv6x-RT ProductID

vincentpoont2:
This looks like a mistake, change FMUv6x-RT ProductID to 29 to match PX4 code. 
The product ID of the firmware used in this product shown in the PX4 Code is 0x001d, which is 29. 

[PX4-Autopilot/boards/px4/fmu-v6xrt/nuttx-config/nsh/defconfig ](https://github.com/PX4/PX4-Autopilot/blob/ece60b6165753c26f8cbc24cd6b2b2918ef8d8fe/boards/px4/fmu-v6xrt/nuttx-config/nsh/defconfig#L43)
[PX4-Autopilot/boards/px4/fmu-v6xrt/nuttx-config/bootloader/defconfig](https://github.com/PX4/PX4-Autopilot/blob/ece60b6165753c26f8cbc24cd6b2b2918ef8d8fe/boards/px4/fmu-v6xrt/nuttx-config/bootloader/defconfig#L44)

It looks like user is unable to flash FW via QGC online due to this mismatch. 

image

### DIFF
--- a/src/VehicleSetup/FirmwareUpgradeController.cc
+++ b/src/VehicleSetup/FirmwareUpgradeController.cc
@@ -62,7 +62,7 @@ static QMap<int, QString> px4_board_name_map {
     {54, "px4_fmu-v6u_default"},
     {56, "px4_fmu-v6c_default"},
     {57, "ark_fmu-v6x_default"},
-    {58, "px4_fmu-v6xrt_default"},
+    {29, "px4_fmu-v6xrt_default"},
     {55, "sky-drones_smartap-airlink_default"},
     {88, "airmind_mindpx-v2_default"},
     {12, "bitcraze_crazyflie_default"},


vincentpoont2:
This fixes the mismatch here https://github.com/mavlink/qgroundcontrol/pull/10846/files, can you double check? @mrpollo

vincentpoont2:
@davids5 Can you take a look?

src/VehicleSetup/FirmwareUpgradeController.cc
```diff
@@ -62,7 +62,7 @@ static QMap<int, QString> px4_board_name_map {
     {54, "px4_fmu-v6u_default"},
     {56, "px4_fmu-v6c_default"},
     {57, "ark_fmu-v6x_default"},
-    {58, "px4_fmu-v6xrt_default"},
+    {29, "px4_fmu-v6xrt_default"},
```
davids5, line 65: @vincentpoont2 I believe this is supposed to be the  **board_id** 
```suggestion
    {35, "px4_fmu-v6xrt_default"},
```

see https://github.com/PX4/PX4-Bootloader/commit/3b381f660490aabfe9885323ce2fb4b0f7f622e6

the [USB VID](CONFIG_CDCACM_VENDORID) and [USB PID](https://github.com/PX4/PX4-Autopilot/blob/main/boards/px4/fmu-v6xrt/nuttx-config/bootloader/defconfig#L44C1-L44C31) are different from the the [board_id](https://github.com/PX4/PX4-Autopilot/blob/main/boards/px4/fmu-v6xrt/firmware.prototype#L2C17-L2C19)

davids5 (changes requested):
Good catch It is wrong. But the value here is the board_id not the the USB product id. (the confusion comes from the fact that in _some_  cases the board id matches the PID

src/VehicleSetup/FirmwareUpgradeController.cc
davids5: @vincentpoont2 - I added a comment to https://github.com/PX4/PX4-Bootloader/blob/main/board_types.txt#L64

vincentpoont2:
> Good catch It is wrong. But the value here is the board_id not the the USB product id. (the confusion comes from the fact that in _some_ cases the board id matches the PID

Good to know!  Is it good to merge now? @davids5

davids5 (approved):

vincentpoont2:
@mrpollo please merge.

mrpollo:
Thanks @vincentpoont2

Issue #10943, opened 2023-12-29
New feature: 3D map viewer integrated into QGC


omid-esrafilian:
I  have recently added a 3D viewer into QGC to visualize the 3D map of the environment, the 3D model of the vehicle (multirotor), and the mission 3D trajectory (including the waypoints). The 3D map for any region of interest can be downloaded from the OpenStreetMap website (.osm file) and imported to this 3D viewer. You can see a demonstration of this new feature from the link below (YouTube video).
I have also attached a screenshot of this viewer integrated into QGC.

Link to the video:
https://youtu.be/aJI5349crG8?si=DUIGgef6J9bnKjxN

The 3D viewer works for any flight control that is compatible with QGC and it supports multiple vehicles as well.
Currently, only a 3D model for multirotors is added to this 3D viewer. However, this can be easily extended for other types of vehicles.
The Qt version that I used for this project is 5.15.2. But it can also be migrated to Qt 6.

Please let me know if I can create a new pull request to merge the code. The current branch that I am working on is Stable_V4.3.

Thanks,

![Screenshot from 2023-12-29 23-16-51](https://github.com/mavlink/qgroundcontrol/assets/115142498/c19ce39e-4e75-4e23-a219-5ceb8d57406c)

DonLakeFlyer:
Thanks for explaining what you are working on. 

My first question is: What is the use case for this? It certainly cool, but what user problem does it solve?

The reason I ask:
* Any new top level feature in QGC can clutter the UI making it harder to use. So if some new thing doesn't have good reasons to be added then it best to leave things simpler.
* Also any large pack of new code which doesn't get much usage tends to get orphaned. In that it's left to the maintainers to keep it up to date and working when the original developer is gone.

omid-esrafilian:
Dear @DonLakeFlyer ,

Thank you for your remarks. 

I can imagine/envision several use cases for this new feature:

- While the QGC is more suited for outdoor settings where GPS is available, there are many scenarios where vehicles need to operate in indoor environments. In these cases, this 3D viewer can potentially enable users to visualize the vehicle location and the 3D map of the environment (e.g. the point cloud) constructed by the robot/vehicle in real-time.

- In the case of simulation, for instance, in SITL or HITL modes, the simulation results, such as the status and movements of the vehicle, can be shown in the QGC (instead of using Gazebo or other similar applications).

[...]

PR #10948, opened 2024-01-06
New feature: 3D viewer integrated into QGC

omid-esrafilian:
Following the issue (#10943), this is a pull request to introduce a 3D viewer new feature into QGC. This 3D viewer can visualize the 3D map of the environment, the 3D model of the vehicle (multirotor), and the mission 3D trajectory (including the waypoints). The 3D map for any region of interest can be downloaded from the OpenStreetMap website (.osm file) and imported to this 3D viewer. 

I have already built and run the code successfully on both Windows and Linux (Ubuntu). Please also note that **Qt Quick 3D** (QT += quick3d) is needed to compile the code.

**Instruction to load a 3D map (.osm file):**
After running the code, click on the icon "3D View" from the side menu on the left. This will direct you to the 3D viewer window. After that, click on the "Setting" Icon (again from the left menu) and select a ".osm" file that has been already downloaded from the OpenStreetMap website. I have also included a sample .osm file in the directory "src/Viewer3D/SampleOsmMap" as an example. Having uploaded the ".osm" file, you should be able to see the 3D map of the environment. You can navigate through the map using the mouse (Right Click: for rotation, Left Clink: for moving, Wheel: for zooming).

On the "Setting menu", you can also set the bias for the altitude/height of the vehicle and its trajectory.

You can see a demonstration of this new feature from the link below as well:

https://youtu.be/aJI5349crG8?si=DUIGgef6J9bnKjxN


ryanjAA:
can this be used for terrain data for 3d terrain? or expanded to that? Pretty cool btw.

DonLakeFlyer:
QGC has terrain data in it already. TerrainOfflineAirMapQuery::requestCarpetHeights should allow you to get terrain tile info.

omid-esrafilian:
Thank you @DonLakeFlyer for the information. I will look into it.

DonLakeFlyer:
Is there a reason why you are creating your own controls? QGC has a full suite of standard controls which follow the ui style for QGC. You should be using those.

omid-esrafilian:
@DonLakeFlyer 
There was no particular reason. I can use the QGC controls instead.

omid-esrafilian:
@DonLakeFlyer 
I have just replaced my controls with QGC standard controls.

DonLakeFlyer:
Can you update the build docs to include the new 3d module: https://docs.qgroundcontrol.com/master/en/qgc-dev-guide/getting_started/index.html. The docs are in the qgc repo. Changed the "en" version.

qgroundcontrol.pro
```diff
@@ -1403,3 +1403,39 @@ LinuxBuild {
 
     INSTALLS += target share_qgroundcontrol share_icons share_metainfo share_applications
 }
+
+
+#-------------------------------------------------------------------------------------
+#For 3D Viewer
+
+QT += \
+    quick3d \
```
DonLakeFlyer, line 1412: Instead of adding this here, you should add it to main place where all modules are included above this.

qgroundcontrol.pro
```diff
@@ -1403,3 +1403,39 @@ LinuxBuild {
...
+    xml
+
+CONFIG += qmltypes
```
DonLakeFlyer, line 1415: I'm not quite sure what this is doing fully. I think it is adding a new qml module. If so, it's not compiling on my Mac. Can you change this to setup up the module like the other QGC qml modules.

qgroundcontrol.pro
```diff
@@ -1403,3 +1403,39 @@ LinuxBuild {
...
+QML_IMPORT_NAME = Viewer3DQmlType
+QML_IMPORT_MAJOR_VERSION = 1
+
+
+SOURCES += \
```
DonLakeFlyer, line 1420: Same thing with this stuff. Since this will just become part of all regular QGC builds. This should just be where all the other source is added. We only tend to split things out like this when that feature is optional.

src/FlightDisplay/FlyView.qml
```diff
@@ -27,13 +27,23 @@ import QGroundControl.Palette
 import QGroundControl.ScreenTools
 import QGroundControl.Vehicle
 
+
+// 3D Viewer modules
+import Viewer3DQmlType             1.0
```
DonLakeFlyer, line 32: The new Qt6 standard is to no include module version numbers.

src/QGCApplication.cc
```diff
@@ -553,6 +553,11 @@ bool QGCApplication::_initForNormalAppBoot()
     QSettings settings;
 
     _qmlAppEngine = toolbox()->corePlugin()->createQmlApplicationEngine(this);
+
+    // 3D Viewer initialization
+    _qgcViewer3D = new QGCViewer3D(this);
+    _qgcViewer3D->initQml(_qmlAppEngine, this);
```
DonLakeFlyer, line 559: So this is probably related to my comment above about the module thing. Can you explain to me what this QGCViewer3D class does. Probably need a high level overview of the architecture, class hierarchy and how it functions to make sure this gets integrated correctly.

DonLakeFlyer (changes requested):

qgroundcontrol.pro
omid-esrafilian: If you mean "CONFIG += qmltypes", yes, indeed, it is the new way (as opposed to qmlRegisterType) of defining and registering QML Types from C++ (based on this reference: https://doc.qt.io/qt-6/qtqml-cppintegration-definetypes.html).
I can change them to the conventional way by using "qmlRegisterType".

But please also note that to be able to compile the code, **"Qt Quick 3D"** needs to be installed using the Qt Maintenance Tool as I highlighted in the figure below:

image

qgroundcontrol.pro
omid-esrafilian: Sure. I will move sources and the modules to the right places.

src/QGCApplication.cc
omid-esrafilian: This class instantiates and initializes all the classes required for the 3D viewer. I created this class just to put all the requirements of the 3D viewer in one place, making it easier to debug or edit without affecting other parts of QGC. However, this class can be removed and its content can be merged with the "QGCApplication" class.

In general, for the moment, the  "QGCViewer3D" class contains three essential classes: 
- OsmParser: to load the OSM map file.
- QmlBackend: for linking the QML and C++.
- MetaDataStreamer: to save and load the user's custom settings for the 3D viewer, such as the path of the OSM map file and the height bias set by the user.

omid-esrafilian:
@DonLakeFlyer 
And yes, sure. I will update the build docs to include the module required for the 3D viewer.

src/FlightDisplay/FlyView.qml
omid-esrafilian: Right, I will remove the versions.

omid-esrafilian:
@DonLakeFlyer 
I have modified the .pro file as you suggested and registered the qml types using qmlRegisterType in QGCApplication.cc. 

So, regarding the "QGCViewer3D" class, should I remove it and merge it with the QGCApplication class?

src/QGCApplication.cc
DonLakeFlyer: So basically this is to make the c++ classes available to Qml code correct. If so you should change this to work like the other C++ classes exposed to Qml already in QGC. Look at QGCApplication::_initCommon. You want to register you classes using qmlRegisterType with your own module name. Then you use Import in Qml and create them as needed in your Qml code. Does that make sense?

omid-esrafilian: In my latest commit today, I have already added some lines of code in the QGCApplication::_initCommon function to register my C++ types using qmlRegisterType and then I imported them in qml flies as you mentioned.

But the QGCViewer3D class does a different job. It makes the instantiated C++ variable visible to qml, and for this, it requires the variable "_qmlAppEngine" which is initialized in QGCApplication::_initForNormalAppBoot(). So, the suitable place to instantiate and initialize QGCViewer3D is inside QGCApplication::_initForNormalAppBoot().

omid-esrafilian: Sorry for the multiple replies.
Please let me add more details about QGCViewer3D class. 

The line of code "_qgcViewer3D = new QGCViewer3D(this);" can be placed anywhere in QGCApplication.cc. However, the function "_qgcViewer3D->initQml(_qmlAppEngine, this);" should be inside "QGCApplication::_initForNormalAppBoot()", after initializing "_qmlAppEngine" and before calling "toolbox()->corePlugin()->createRootWindow(_qmlAppEngine);", since the variables required for the 3D viewer should be initialized before creating the RootWindow. 
Please let me know if that is ok.

DonLakeFlyer: You shouldn't be injecting this into all Qml at a global level. That isn't the way QGC does things. If you look at FactPanelController on both the C++ and Qml side of thing you should see how you make a C++ class/object available to and/or instantiated from the Qml side of the code. 

Basically the way this should work:
* You have worker C++ classes you need to make the Qml work
* You register those class(es) as creatable from Qml using qmlRegisterType in QGCApplication
* Then in your Qml you create them in there

This way:
* They are not global things that magically show up in all Qml
* They only are live when they are needed

omid-esrafilian: Thanks for the explanation. Yes, that makes sense. 
So, in my new commit as you suggested, I have registered all my types/classes with qmlRegisterType and there is no global declaration anymore. I created a new Class Viewer3DFacts to handle accessing C++ variables and classes from the QML side. Please let me know if this looks good to you.

DonLakeFlyer:
Some global things:
* Can you match the coding style: https://github.com/mavlink/qgroundcontrol/blob/master/CodingStyle.cc. Specifically with respect to if/else bracing positions and if's always having braces even single line.
* There are a number of places where the spacing of code which you hav not modified has been altered. This generates unnecessary diffs. Can you limit the changes to place you modified code.
* Once everything is reviewed at the end squash this down to a single commit

.github/workflows/android_release.yml
```diff
@@ -74,7 +74,7 @@ jobs:
           arch:         ${{ matrix.arch }}
           extra:        --autodesktop
           dir:          ${{ runner.temp }}
-          modules:      qtcharts qtlocation qtpositioning qtspeech qt5compat qtmultimedia qtserialport qtimageformats qtshadertools qtconnectivity
+          modules:      qtcharts qtlocation qtpositioning qtspeech qt5compat qtmultimedia qtserialport qtimageformats qtshadertools qtimageformats qtshadertools qtconnectivity qtquick3d
```
DonLakeFlyer, line 77: Duplicated qtimageformats qtshadertools

src/FlightDisplay/FlyViewToolStripActionList.qml
```diff
@@ -21,7 +21,54 @@ ToolStripActionList {
         ToolStripAction {
             text:           qsTr("Plan")
             iconSource:     "/qmlimages/Plan.svg"
-            onTriggered:    mainWindow.showPlanView()
+            onTriggered:
+            {
```
DonLakeFlyer, line 25: An example of brace in wrong place. Start brace should not be on separate line

src/FlightDisplay/FlyViewToolStripActionList.qml
```diff
@@ -21,7 +21,54 @@ ToolStripActionList {
...
+                map_icon.showFlyMap()
+            }
+        },
+        ToolStripAction {
+            id: map_icon
+            text:           qsTr("3D View")
+            iconSource:     "/qmlimages/Viewer3D/City3DMapIcon.svg"
+            onTriggered:{
+                if(viewer3DWindow.z === 0)
+                {
```
DonLakeFlyer, line 36: Same bracing problems here

src/QGCApplication.cc
```diff
@@ -134,16 +144,16 @@
 class FinishVideoInitialization : public QRunnable
 {
 public:
-  FinishVideoInitialization(VideoManager* manager)
-      : _manager(manager)
-  {}
+    FinishVideoInitialization(VideoManager* manager)
```
DonLakeFlyer, line 147: Example of spacing changes in code you didn't modify

src/Viewer3D/CityMapGeometry.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+    m_model_name = m_name;
+    //    setName(m_model_name);
+
+    emit modelNameChanged();
+}
+
+void CityMapGeometry::setCityMap(QString map_path)
+{
+    if(m_city_map_path.compare(map_path) == 0)
```
DonLakeFlyer, line 27: Even single line ifs should have bracing. Single line ifs have a tendency to turn into multi-line down the road. Which can then lead to bugs when people don't noticed they need to add bracing.

src/Viewer3D/Viewer3DFacts.h
```diff
@@ -0,0 +1,29 @@
+#ifndef VIEWER3DFACTS_H
+#define VIEWER3DFACTS_H
```
DonLakeFlyer, line 2: This class should used the terminology "Fact". QGC already has the concept of a "Fact"  and a "FactSystem" which is something else. Don't won't to cause confusion.

DonLakeFlyer (changes requested):

DonLakeFlyer:
> So, regarding the "QGCViewer3D" class, should I remove it and merge it with the QGCApplication class?

I don't understand why this is a global singleton which is always available? This should only be instantiated when the user uses the 3D View so it should be instantiated when the user creates that view.

omid-esrafilian:
@DonLakeFlyer 
Sure, I will carefully review the code to make it QGC coding style compliant and will reduce the unnecessary changes.

Regarding the QGCViewer3D class, I can instantiate it when the user opens/creates the 3D view, but I need to somehow be notified when the 3D view is being opened, which in my opinion again requires a C++ class similar to QGCViewer3D as a global singleton. Unless if there is already a place in the QGC code where I can be notified on the C++ side when the 3D view is being opened (i.e. the 3D view button is clicked). I appreciate it if you could let me know if there is already a way in the QGC code to do it.

omid-esrafilian:
@DonLakeFlyer 
I have reviewed the code to apply the changes you recommended. The 3D Viewer is now initiated when the user opens the 3D view. Please let me know if this looks good to you.

DonLakeFlyer:
Ok, I just ran it for the first time. Some changes from my testing:
* The Setting button should only show when the 3D View is being displayed
* The Settings should use a regular QGCPopupDialog with Open/Cancel buttons. Search the codebase for example usage. There are tons.
* The settings themselves in the Settings dialog should be their own SettingsGroup. This was they persist across runs and use the standard QGS settings persistence model.  I probably wouldn't save the OSM filename in settings, but the rest should persist. And then the fields for editing should use the Fact Controls. Look at FlyViewSettings.*, FlyView.SettingsGroup.json for simple usage example.

Lastly can you update the QGC user docs to explain how to use this feature. As well as explain what the fields are in the settings. You can either do this in this pull (the docs are in the main repo) or in a separate pull.

omid-esrafilian:
@DonLakeFlyer 

In the new commit, I have applied the new changes you suggested.

Among those, I created a new SettingGroup called "Viewer3D.SeetingsGroup.json" and "Viewer3DSeetings.*" for the 3D View. I have retained the OSM filename in the setting for the moment. Please let me know if you have any suggestions or want to put the OSM filename elsewhere. 

Moreover, I have also updated the QGC user docs to explain how to use the 3D View and the Settings.

Please let me know if I need to make further changes.

DonLakeFlyer:
Lastly can you Squash into a single commit then I will merge the pull request.

omid-esrafilian:
@DonLakeFlyer 

At some point, I synced my fork with the original QGC repository, therefore I have other contributors' commits in my Git logs too. Now, I am worried that if I want to squash all my commits, my changes cannot be distinguished from others (or am I mistaken?). Do you have any suggestions?

For your reference, below I included the link to the list of all the commits in my repository:
https://github.com/omid-esrafilian/qgroundcontrol-viewer3D/commits/master/

P.S. I have already squashed all my commits since the last synchronization with the original QGC repository.

DonLakeFlyer:
If you rebase your branch to upstream master then this should fix itself and you should only see your commits in the pull.

omid-esrafilian:
@DonLakeFlyer

Thanks for your reply.
I could solve the problem and squash all the commits into one.
Please let me know if that looks good to you.

DonLakeFlyer:
@omid-esrafilian Thanks for all the work on this. I need to put some thought into how this show be exposed to the user. The Toolstrip buttons are very coveted real estate for this. I'd like to find some different way. But I need to put some thought into it. Hopefully once I have an idea we can discuss further and you could help with the changes.

omid-esrafilian:
@DonLakeFlyer 
My pleasure. And thank you also for all the feedback.
Yes, sure. Please let me know when you would like to make any changes. I would be happy to help.

SebastianViollaz:
Hi @omid-esrafilian i like the new feature.
Some questions about it:
I have this .osm file dowloaded:

image

This is the 3d view showed in QGC:
image

I think that maybe is due to a parameters configuration read error or something.
The 3d model that is showed is only building that is only showed:
image


Why the others buildings are not showed?

omid-esrafilian:
@SebastianViollaz 
Hi, Thanks for your interest.
Can you please share your .osm file with me? Then I can take a look and try it as well.

SebastianViollaz:
[GYT-Estadio.zip](https://github.com/mavlink/qgroundcontrol/files/14450331/GYT-Estadio.zip)
This is the file.
This file have a litlle larger area than the one that I upload, but i it has the same problem, only a few buildings are showed in the 3D view.

omid-esrafilian:
@SebastianViollaz 
Here is what's happening. In the code, I only show the buildings that their heights are specified (either in terms of the levels or in meters). In your .osm file, only a few buildings are like this. However, if I enable the viewer to show all types of buildings, you'll be able to see them all. There are several types of buildings defined on the OpenStreetMap website. I've just included some of them in the code, you can see the results below. Almost all the buildings are displayed now.
image

So, I will include this option in the code in the next pull.

SebastianViollaz:
Thanks that would be great. Thanks for the solution.
Also I've noticed that when drawing the flight path using ROI's Items.
Here is a picture:
image

Maybe when drawing a ROI would be nice to  add an icon or something instead of drawing a line to the coordinates of the roi.
Here is the same mission in the 2D View:
image
Some numbers are skipped in the view because there are some items that aren't visible, This is the real order:
2- ROI - Change speed- 5 -Cancel ROI - 7

omid-esrafilian:
@SebastianViollaz 
Thanks for the feedback.
Yes, true. That shouldn't be like that. I will correct it too.

omid-esrafilian:
@SebastianViollaz 
I have just created a new pull request (#11173) regarding the issues you pointed out. I believe that should solve the problems.
Please let me know if that works.

PR #10970, opened 2024-01-16
RemoteID: validate EU operator ID

julianoes:
This implements the validation of the operator ID as required for the EU remote ID spec.

The way it works is that:
1. The user enters the operator ID including the checksum and secret.
2. The ID is validated using the secret against the checksum.
3. The secret is removed and the public ID is saved, as well as a flag signalling that the ID has been checked.

This work is sponsored by [bluemark.io.](https://bluemark.io/), https://github.com/BluemarkInnovations.

operator-id-2-2024-01-25_13 31 43-ezgif com-video-to-gif-converter

### DIFF
--- a/src/ui/preferences/RemoteIDSettings.qml
+++ b/src/ui/preferences/RemoteIDSettings.qml
@@ -243,7 +243,7 @@ Rectangle {
                         id:                     operatorIDFlag
                         Layout.preferredHeight: flagsHeight
                         Layout.preferredWidth:  flagsWidth
-                        color:                  qgcPal.colorRed //_activeRID ? (_activeVehicle.remoteIDManager.operatorIDGood ? qgcPal.colorGreen : qgcPal.colorRed) : (_offlineVehicle.remoteIDManager.operatorIDGood ? qgcPal.colorGreen : qgcPal.colorRed)
+                        color:                  _activeRID ? (_activeVehicle.remoteIDManager.operatorIDGood ? qgcPal.colorGreen : qgcPal.colorRed) : qgcPal.colorGrey
                         radius:                 radiusFlags
                         visible:                commsGood && _activeRID ? (QGroundControl.settingsManager.remoteIDSettings.sendOperatorID.value || _regionOperation == RemoteIDSettings.RegionOperation.EU) : false
 
@@ -771,12 +771,8 @@ Rectangle {
                     Layout.fillWidth:       true
 
                     border.width:   _borderWidth
-                    border.color:   (_regionOperation == RemoteIDSettings.RegionOperation.EU ||
-                                     QGroundControl.settingsManager.remoteIDSettings.sendOperatorID.value) ?
-                                      (_activeRID ?
-                                        (_activeVehicle.remoteIDManager.operatorIDGood ? qgcPal.colorGreen : qgcPal.colorRed) :
-                                         (_offlineVehicle.remoteIDManager.operatorIDGood ? qgcPal.colorGreen : qgcPal.colorRed)) :
-                                       color
+                    border.color:   (_regionOperation == RemoteIDSettings.RegionOperation.EU || QGroundControl.settingsManager.remoteIDSettings.sendOperatorID.value) ?
+                                    (_activeRID && !_activeVehicle.remoteIDManager.operatorIDGood ? qgcPal.colorRed : color) : color
 
                     GridLayout {
                         id:                         operatorIDGrid
@@ -819,31 +815,35 @@ Rectangle {
                             visible:            QGroundControl.settingsManager.remoteIDSettings.operatorID.visible
                             Layout.fillWidth:   true
                             maximumLength:      20 // Maximum defined by Mavlink definition of OPEN_DRONE_ID_OPERATOR_ID message
-                            property bool editing: false
 	                    onTextChanged: {
-                                // Usually changes are only saved when editing is finished (enter, out of focus)
-                                // but we want to track changes as letters are entered. Therefore, we set the new
-                                // text directly.
-                                // We also need to keep track of when this is being changed from within C++, so when the secret part
-                                // is removed. That's what the editing flag is for.
-                                if (editing) {
-                                    return
-                                }
                                 if (_activeVehicle) {
-                                    _activeVehicle.remoteIDManager.setOperatorID(text)
+                                    _activeVehicle.remoteIDManager.checkOperatorID(text)
                                 } else {
-                                    _offlineVehicle.remoteIDManager.setOperatorID(text)
+                                    _offlineVehicle.remoteIDManager.checkOperatorID(text)
                                 }
                             }
                             onEditingFinished: {
-                                editing = true
                                 if (_activeVehicle) {
-                                    _activeVehicle.remoteIDManager.checkOperatorID(true)
+                                    _activeVehicle.remoteIDManager.setOperatorID()
                                 } else {
-                                    _offlineVehicle.remoteIDManager.checkOperatorID(true)
+                                    _offlineVehicle.remoteIDManager.setOperatorID()
                                 }
-                                editing = false
                             }
+                        }
+
+                        // Spacer
+                        QGCLabel {
+                            text:               ""
+                            visible:            _regionOperation == RemoteIDSettings.RegionOperation.EU
+                            Layout.alignment:   Qt.AlignHCenter
+                            Layout.fillWidth:   true
+                        }
+
+                        QGCLabel {
+                            text:               QGroundControl.settingsManager.remoteIDSettings.operatorID.shortDescription + qsTr(QGroundControl.settingsManager.remoteIDSettings.operatorIDValid.rawValue == true ? " valid" : " invalid")
+                            visible:            _regionOperation == RemoteIDSettings.RegionOperation.EU
+                            Layout.alignment:   Qt.AlignHCenter
+                            Layout.fillWidth:   true
                         }
 
                         QGCLabel {
@@ -857,7 +857,7 @@ Rectangle {
                             onClicked: {
                                 if (checked) {
                                     if (_activeVehicle) {
-                                        _activeVehicle.remoteIDManager.checkOperatorID(false)
+                                        _activeVehicle.remoteIDManager.setOperatorID()
                                     }
                                 }
                             }
--- a/src/ui/preferences/RemoteIDSettings.qml
+++ b/src/ui/preferences/RemoteIDSettings.qml
@@ -47,6 +47,7 @@ Rectangle {
 
     // General properties
     property var  _activeVehicle:       QGroundControl.multiVehicleManager.activeVehicle
+    property var  _offlineVehicle:      QGroundControl.multiVehicleManager.offlineEditingVehicle
     property int  _regionOperation:     QGroundControl.settingsManager.remoteIDSettings.region.value
     property int  _locationType:        QGroundControl.settingsManager.remoteIDSettings.locationType.value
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
> * Validating the input on each keypress rather than when hitting enter, going out of focus.

Why do you need to do this?

julianoes:
> Why do you need to do this?

That's the UX I would expect coming from web forms that mostly work that way.

DonLakeFlyer:
What UX? How does it work from a user perspective?

julianoes:
When you type something in a web form, the indicator whether it's a good password, address, whatever, is done as you enter the characters, if it's well done.

You can see what I mean in the screenshot where it turns green once the entered operator ID is valid.

I'll comment inline to point out what I mean.

src/ui/preferences/RemoteIDSettings.qml
```diff
@@ -814,12 +819,31 @@ Rectangle {
...
+                                // is removed. That's what the editing flag is for.
+                                if (editing) {
+                                    return
+                                }
+                                if (_activeVehicle) {
+                                    _activeVehicle.remoteIDManager.setOperatorID(text)
+                                } else {
+                                    _offlineVehicle.remoteIDManager.setOperatorID(text)
+                                }
+                            }
```
julianoes, line 837: This is required to validate the input as you enter it.

DonLakeFlyer:
> where it turns green once the entered operator ID is valid.

Ah, I missed that somehow! Interesting, but not the way validation works anywhere else in QGC. Not sure it's worth creating a new visual paradigm for field validation. 

Current field validation uses a tooltip to show validation error. You also need to set globals.validationError to true/false if the validation failed/succeeded. Otherwise the user can switch away from that page while there is crap data in the field.

julianoes:
> Otherwise the user can switch away from that page while there is crap data in the field.

Ok, but in this case it won't send the operator ID if it has not been marked as good.

DonLakeFlyer:
> Ok, but in this case it won't send the operator ID if it has not been marked as good.

That's kind of a crappy user experience. It's going to save a bad value to settings isn't it? Even if validated later, it's odd to save something that won't work.

BluemarkInnovations:
> > Ok, but in this case it won't send the operator ID if it has not been marked as good.
> 
> That's kind of a crappy user experience. It's going to save a bad value to settings isn't it? Even if validated later, it's odd to save something that won't work.

I think what @julianoes did not mention is that this PR is about regulation. In the EU you cannot get a UAV certified if you do not have this operatorID check implemented. Also the regulation says that you are not allowed to save the private part of the operator ID. So on one hand you have the crappy user experience, but on the other hand you have the RemoteID regulation that prescribes certain behavior.

DonLakeFlyer:
Ok, so someone needs to explain to me or point me to something which I can read which explain fully how this id is supposed to work. Then I can help with creating a good experience around it.

BluemarkInnovations:
> Ok, so someone needs to explain to me or point me to something which I can read which explain fully how this id is supposed to work. Then I can help with creating a good experience around it.

Sure

See attached picture

operatorID

So a user needs to enter the full operatorID (OPRN). Once it has been validated, only the public part should be broadcast/included in the Remote ID signal. As the next time QGC is started, it should of course re-use the OPRN that has been saved previously (public part only). 

Today I have time to review the PR. A small bug: as long as the user enters the full OPRN and does not leave the input field, also the private part is broadcast. That should not happen, as it is the private part that should not be shared with anyone.

Let me know if you have questions.

BluemarkInnovations:
Relevant standard: https://asd-stan.org/downloads/pren-4709-002-corr/

But I think above OPRN is enough for this PR. The screenshot is from that document.

julianoes:
Ok, I think there are two things to decide on here:

1. Validation per character, or per edit (enter, out of focus): I think validation per character is nicer, but it is not consistent with the rest of QGC, and it requires more complex code to make it work. I'm ok changing this back to validation per edit if you feel strongly about it @DonLakeFlyer.
2. Saving/sending incomplete/wrong operator ID. I will check that we will not send out an invalid/incomplete ID. The fact that we save something wrong/incomplete is not a problem in my eyes as it's shown as red, as is any other field that is wrong.

DonLakeFlyer:
I don't have a problem with per-char validation. I'm just complaining about how validation failure is shown/handled. The red border thing isn't something used for field validation. There are two ways validation failure is currently shown in QGC:
* Tooltip - this is used by value validation like value must be greater than 10
* Warning text next to/below field. This is used in a couple places. When there is a problem the warning text goes visible (using the warning palette color)

Of those for this specific case, the warning text is probably best since I assume the warning text might be pretty involved with respect to explaining what is wrong.

Saving an invalid value is odd, but probably not the end of the world.

julianoes:
Oh, I didn't realize that. I'll change it to text next to the field then.

julianoes:
@DonLakeFlyer just for my undestanding. What does the red or green frame/box around any of the settings mean then?

Davidsastresas:
@julianoes if I recall correctly that colored square is to highlight when that particular field is okay or not, so it is more visual the part that is preventing remote id to be 100% alright.

julianoes:
@Davidsastresas right, thanks! However, I think it only shows green or red if a vehicle is actually connected and data is actually sent out.

julianoes:
@DonLakeFlyer I've added a commit that:
- Simplifies the logic,
- Checks the ID on every input,
- But only saves it on enter or out-of-focus,
- And moves the indication to a text below rather than the border color.

I'll change the PR description screenshot to reflect the changes.

DonLakeFlyer:
> just for my undestanding. What does the red or green frame/box around any of the settings mean then?

I didn't even know such a thing existed. Nothing else in QGC does that except for remote id I guess. Not sure that is a good thing I'd have to look at it.

hamishwillee:
@julianoes Does this requirement/behaviour need to be somehow captured in MAVLink docs? https://mavlink.io/en/services/opendroneid.html

julianoes:
@DonLakeFlyer do you need anything more from me here to get this in?

@hamishwillee I think @BluemarkInnovations could assist with the docs.

BluemarkInnovations:
@hamishwillee The idea of this PR/implementation is that the user interface checks the operator ID. And only the public part is included in the MAVLink message that is sent to the Remote ID module. 

"Provides the operator ID." We could change this text in:

"Provides the operator ID. For some regions like the EU, the user interface needs to check if the operator ID is valid. In the EU, the Operator ID is called OPRN (UAS Operator Registration Number). It contains a public part and a private part. The user interface needs to check if the OPRN is valid. If so, only the public part is sent via MAVLink to the Remote ID module. The private part is considered secret information that only the operator possesses and shall not be stored in an UAS."

DonLakeFlyer:
> do you need anything more from me here to get this in?

Nope, you can merge it when you are ready.

hamishwillee:
> the Remote ID module. The private part is considered secret information that only the operator possesses and shall not be stored in an UAS."

Hi @BluemarkInnovations Could you please create a PR to update the docs? I could do it, but if there are questions then it is better if they go to you. If not, let me know.

BluemarkInnovations:
> Hi @BluemarkInnovations Could you please create a PR to update the docs? I could do it, but if there are questions then it is better if they go to you. If not, let me know.

Yes I do, but not sure where to create a PR. You mean a PR for https://github.com/mavlink/mavlink and common.xml? I mean I don't know where the repository of https://mavlink.io/en/services/opendroneid.html is located.

mahyarmotie:
> This implements the validation of the operator ID as required for the EU remote ID spec.
> 
> The way it works is that:
> 
>     1. The user enters the operator ID including the checksum and secret.
> 
>     2. The ID is validated using the secret against the checksum.
> 
>     3. The secret is removed and the public ID is saved, as well as a flag signalling that the ID has been checked.
> 
> 
> This work is sponsored by [bluemark.io.](https://bluemark.io/), https://github.com/BluemarkInnovations.
> 
> operator-id-2-2024-01-25_13 31 43-ezgif com-video-to-gif-converter [ operator-id-2-2024-01-25_13 31 43-ezgif com-video-to-gif-converter ](https://private-user-images.githubusercontent.com/1419688/299501855-5f9ba7a5-0094-45fc-9cbf-c094d03a5800.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzY4NDAxODcsIm5iZiI6MTczNjgzOTg4NywicGF0aCI6Ii8xNDE5Njg4LzI5OTUwMTg1NS01ZjliYTdhNS0wMDk0LTQ1ZmMtOWNiZi1jMDk0ZDAzYTU4MDAuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDExNCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTAxMTRUMDczMTI3WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MDI2MGU1YzcxMDkwZTJkYmE5YTQ5MGIyNGE2NDQ5YTc1MzM5ZDkwNDI2MDE1Y2ZjY2U5MWM0OWY4NWFhNzIxOCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.En0a27r3mX7GMAFGbW_bhw9ySTpsv21FV06PVBaKwVw) [ ](https://private-user-images.githubusercontent.com/1419688/299501855-5f9ba7a5-0094-45fc-9cbf-c094d03a5800.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzY4NDAxODcsIm5iZiI6MTczNjgzOTg4NywicGF0aCI6Ii8xNDE5Njg4LzI5OTUwMTg1NS01ZjliYTdhNS0wMDk0LTQ1ZmMtOWNiZi1jMDk0ZDAzYTU4MDAuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDExNCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTAxMTRUMDczMTI3WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MDI2MGU1YzcxMDkwZTJkYmE5YTQ5MGIyNGE2NDQ5YTc1MzM5ZDkwNDI2MDE1Y2ZjY2U5MWM0OWY4NWFhNzIxOCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.En0a27r3mX7GMAFGbW_bhw9ySTpsv21FV06PVBaKwVw)

thank you for this explanation, it's very helpful, but i have a problem, with one of our ID we type at the end "-123" and it works but it doesn't work with our another ID , how can we calculate the checksum or 3digit secret?

PR #10972, opened 2024-01-16
PID Tuning: Don't hide charts on mobile

DonLakeFlyer:


DonLakeFlyer:
Just needs restructuring for small screens. I'll deal with that.

PR #10999, opened 2024-01-30
Make RC RSSI Indicator show via checkbox

ryanjAA:
Adds a checkbox to app settings for turning the RC rssi indicator on and off.

### DIFF
--- a/src/Settings/App.SettingsGroup.json
+++ b/src/Settings/App.SettingsGroup.json
@@ -340,6 +340,13 @@
     "longDesc":  "Ardupilot Support server to forward mavlink to. i.e: support.ardupilot.org:xxxx",
     "type":      "string",
     "default":   "support.ardupilot.org:xxxx"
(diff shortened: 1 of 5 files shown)


ryanjAA:
@DonLakeFlyer As discussed in Discord, maybe you hold out on this if you are adding more info in the RSSI indicator.

ryanjAA:
Screen Shot 2024-01-29 at 9 54 49 PM

DonLakeFlyer:
Two problems:
* This is way behind what settings looks like now, hence the conflicts
* Also toolbar indicators now do more than just show status now. There are settings and stuff in the expanded pages. This means disabling them outright may hide more features than just unwanted status.

Because of that I'm closing this. Please create a new Issue and assign it to me with the details on why the RC status indicator is not useful or works incorrectly. Then I can take that into account in my future indicator work.

PR #11009, opened 2024-01-31
fix cmake build: add UTM library

DonLakeFlyer:
Fixes the error:
qrc:/qml/MainRootWindow.qml:23:1: module "QGroundControl.UTMSP" is not installed

I had to add utmsp_dummy.qrc directly to QGC_RESOURCES, adding it to the UTMSP library wasn't enough.


DonLakeFlyer:
My revert on a previous pull seems to have dropped this. So putting it back.

Issue #10985, opened 2024-01-21
Global search for Qt6 onFoo handler value injection


DonLakeFlyer:
ComboBox.onActivated handlers must be updated for Qt6: ```onActivated: (index) =>```
QGCFileDialog.onAcceptedForLoad/Save

DonLakeFlyer:
Removed onClick checking. That is completed.

PR #11014, opened 2024-02-01
Remove redundant code after moving to Qt6 and fix mouse action handling

s-lisovenko:


DonLakeFlyer:
Hmm, those duplicate QtQuick.Controls are odd. I have a feeling they are a global search/replace gone bad. Let me take look back to see what I screwed up in a commit somewhere just to make sure.

DonLakeFlyer:
Yup. it's here:
```
import QtQuick.Controls     1.2
import QtQuick.Controls.Styles  1.4
```
Changed to:
```
import QtQuick.Controls
import QtQuick.Controls
```

QtQuick.Controls.Styles is a Qt5 thing. Qt6 does styling differently so this duplication doesn't indicate some other possible problem that make be lurking.

DonLakeFlyer:
I also update https://github.com/mavlink/qgroundcontrol/issues/10985 to make anything switch through a global check of all onClicked handlers.

DonLakeFlyer:
> make another swing through a global check of all onClicked handlers.

Or did you already do that when you were creating this pull?

s-lisovenko:
> Or did you already do that when you were creating this pull?

I checked everything where needed to use a mouse variable inside. If some handler does not use the mouse then i did not change the function signature

ysfkg:
I don't know if it's right to write this as a comment here. But the new program compiled with qt6 is complete garbage. Due to my busy schedule, I was last interested in qgroundcontrol 2 months ago. At that time it was compiled with qt5 and its performance was very good. But now it has been compiled with qt6 and some changes have been made to the code. Even though I run the application on my i7 computer (I have an nvidia GPU), it causes me to wait for seconds on the task drawing screen. There is no optimization. The application lags, especially during drawing tasks. Whoever did this really messed up the app.

DonLakeFlyer:
> I checked everything where needed to use a mouse variable inside. If some handler does not use the mouse then i did not change the function signature

Ok, great thanks. I can take that off the list then.

DonLakeFlyer:
> I don't know if it's right to write this as a comment here

Folks here contributes their time to QGC for free. Which in turn means you get to use QGC for free as well. Open source works by everyone contributing their time. Whether it be on the coding side, or on the testing and feature improvement side. If you want to help improve QGC from the standpoint of a user, the way to do that is not disparage the developers who have spent hours/months/years of their free time contributing to the project. The way is to contribute yourself on a constructive way by creating an Issue with specific details of the problem:
* What OS are you using?
* What version of the build are you using? What version of the build are you comparing to?
* What problems are you seeing?

And then the folks who volunteer there time on the coding side of thing can fix the problems.

PR #11019, opened 2024-02-04
Fix indicatorDrawer closing + fix onNewFormattedMessage for message inficator

s-lisovenko:


DonLakeFlyer:
So what's the current problem with indicator closing?

s-lisovenko:
> So what's the current problem with indicator closing?

Does not close because the indicator name is incorrect

DonLakeFlyer:
Wow, I screwed that up somewhere along the way didn't I! Thanks for the fixes.

PR #11032, opened 2024-02-08
GPS/Drivers: Update submodule, fix RTK for F9P

julianoes:
This updates the GPS driver submodule to latest main. This includes a fix for RTK for the uBlox F9P where a config failed and prevent proper init of the GPS module.

~Depends on https://github.com/PX4/PX4-GPSDrivers/pull/149~


julianoes:
Pointed submodule to main, force-pushed, and tested again.

bkueng (approved):

PR #11034, opened 2024-02-08
Convert SDL Joystick DB to submodule

HTRamsey:
This defines joystick mappings and the current ones are ancient


DonLakeFlyer:
Are there forward compatible issue here with respect to this being connected to submodule HEAD? Whereas as the code itself is using some specifically version of SDL lib.

HTRamsey:
Not really, unless the format of joystick mappings change randomly one day but I highly doubt SDL would do that. This just makes sure all new joysticks will eventually get their mappings without anybody here doing anything.

DonLakeFlyer:
Ok, thanks. I guess this is easier than copying files. You'll still have to move the submodule pointer when you want new stuff.

PR #11035, opened 2024-02-08
Fix setting default style

HTRamsey:
Fixes `ERROR: QQuickStyle::setStyle() must be called before loading QML that imports Qt Quick Controls 2.`

QQuickStyle calls QQuickStyleSpec which is a global static, so it can be called before the QApp is made. Usually this error will only happen `_exitWithError` but at least it won't output an irrelevant error anymore.


DonLakeFlyer:
Thanks for the fix

PR #11044, opened 2024-02-11
Remove deprecated UAS class

HTRamsey:
The `UAS` class was only still used under two circumstances:

- Sending signals to `LogDownloadController` for message IDs `MAVLINK_MSG_ID_LOG_ENTRY` & `MAVLINK_MSG_ID_LOG_DATA`.
- Sending command `MAV_CMD_START_RX_PAIR` from `RadioComponentController`.

So just move the signals handling the messages and the function `pairRX` into the vehicle class and delete the `UAS` class. Deletes all references to `UAS` and `UASInterface` but leaves `UASMessageHandler` alone for now. I will address that class if this PR gets accepted.


DonLakeFlyer:
Thanks so much for this. The migration from UAS to Vehicle has been a WIP in progress for ages now. Finally gone!

PR #11045, opened 2024-02-11
Update OpenSSL to v3

HTRamsey:
QT6.5+ uses OpenSSL3. This is particularly an issue with downloading offline maps using https. Also, openssl is not included when using cmake but you can just use the following for android (Windows will use native):
```
if (ANDROID)
    FetchContent_Declare(
      android_openssl
      DOWNLOAD_EXTRACT_TIMESTAMP true
      URL      https://github.com/KDAB/android_openssl/archive/refs/heads/master.zip
    )
    FetchContent_MakeAvailable(android_openssl)
    include(${android_openssl_SOURCE_DIR}/android_openssl.cmake)
    add_android_openssl_libraries(${PROJECT_NAME})
endif()
```


DonLakeFlyer:
Ah, I've been noticing those ssl v3 errors on android logs. Haven't had time to look. I assume this is what was causing that?

DonLakeFlyer:
> Also, openssl is not included when using cmake but you can just use the following for android (Windows will use native):

Not quite sure what you mean by this? Shouldn't cmake build replicate what happens in the android.pri which is in the submodule?

HTRamsey:
I just meant that the CMake build hasn't been updated to include openssl yet. There are a handful of differences between the qmake and cmake builds that prevent me from building with cmake right now so I couldn't verify it quickly, but it works with qmake.

Also, I'm pretty sure you can remove the files in `libs/OpenSSL/windows` as I believe Qt will just use other native windows methods for ssl, but it doesn't hurt to leave them either as long as they are updated to 3.0.

DonLakeFlyer:
Ok, thanks. Could you detail what've you've found with cmake and other builds problems into an Issue and assign to me. Don't want to lose track of missing things, possible changed. I've been focused on desktop cmake builds, haven't tried an android build yet.

HTRamsey:
I don't think I can assign you, but I made it

DonLakeFlyer:
@HTRamsey Could Ubuntu be screwed up with OpenSSL as well? There is someone in discord complaining about Ubuntu 22 and OpenSSL not matching certs.

PR #11047, opened 2024-02-11
Linux: do not make usb permissions fatal

HTRamsey:
QGC doesn't have to be used with USB, so there is no reason insufficient serial permissions should be fatal. Also, only show the error if not building with `NO_SERIAL_LINK`.

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -208,24 +208,6 @@ QGCApplication::QGCApplication(int &argc, char* argv[], bool unitTesting)
                     "sudo apt-get remove modemmanager</pre>").arg(qgcApp()->applicationName())));
             return;
         }
-        // Determine if we have the correct permissions to access USB serial devices
-        QFile permFile("/etc/group");
-        if(permFile.open(QIODevice::ReadOnly)) {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
This looks ok. The reason for the hard stop was that users kept screwing this up and reporting bugs that they couldn't connect. I guess instead popping it up every single time is sort of just as good. 

The best fix for this would be two things:
* If any of the USB based autoconnect settings are enabled then check this on boot.
* If autoconnect is off, then do the check when the user tries to instantiate a manual serial link

If you have time, would be nice to make this check less annoying for people who will never use USB, but still have a regular build which include serial port support.

If not, this is still better than before.

HTRamsey:
Yup, I will take a look at it later either today or the next few days.

DonLakeFlyer:
Thanks so much. You want me to merge this in. Or wait for an updated one?

HTRamsey:
Let's just hold off

HTRamsey:
@DonLakeFlyer actually can we just go ahead and merge this and I will get back to the rest later? I have some other things I want to address first for a while.

PR #11048, opened 2024-02-11
3D View changes and fixes

omid-esrafilian:
The following changes have been made regarding the 3D viewer:

- Touchscreen is supported now
- Uploading an OSM file is not mandatory anymore to use the 3D View
- docs is updated with more details on how to use the 3D View and also the build instructions
- Viewer3DManager is also initialized when the 3D view is opened 
- Using Flickable in the 3D view Setting menu
- The camera is initialized facing downward when the 3D view is opened for the first time.


DonLakeFlyer:
> * Using Flickable in the 3D view Setting menu

This is shown in a QGCPopupDialog. QGCPopupDialog automatically sizes itself to fit what you put in it never growing larger than what will fit within the screen. It already has a Flickable in it so if your content needs to scroll to fit the screen it will happen automatically. No need to do it yourself. Was that not working for you?

DonLakeFlyer:
In reality the Viewer3DSettingsMenu should just use QGCPopupDialog as its base class. And name it not with the word Menu. Then you just put that into a Component when you need to use it.

DonLakeFlyer:
And then for dialogs the best structure tends to be:
QGCPopupDialog {
    ColumnLayout {
         ....
    }
}

This way the layout automatically sizes to as much as put into it. Which will in turn correctly size the dialog. Tons of example usage in the codebase.

src/FlightDisplay/FlyViewToolStripActionList.qml
```diff
@@ -66,7 +66,7 @@ ToolStripActionList {
             enabled: false
             visible: enabled
             onTriggered:{
-                viewer3DWindow.settingMenuOpen = !viewer3DWindow.settingMenuOpen
+                viewer3DWindow.settingsDialogOpen = !viewer3DWindow.settingsDialogOpen
```
DonLakeFlyer, line 69: Triggering things other than visibility to happen on boolean state changes is pretty odd. This sort of thing is much better handled by a function in Viewer3DWindow like showSettingsDialog() which then creates the settings dialog component. Much clearer as to what it does.

DonLakeFlyer:
I'd like to also clean up the fact that the 3D View button is always visible even though the majority of users will not use this. I think the way to do this is as follows:
* Make the 3d view settings available as a separate group in Fly View settings
* Add ability to clear the OSM file if not already possible
* Only show the 3D View button in fly view if there is an OSM file specified
* They I wonder what you think about still leaving the Settings button in the Fly View as well when the 3d map is displayed? Is that needed/useful? Or just make them go back to the regular settings place.

omid-esrafilian:
@DonLakeFlyer 
Thanks for the feedback. Sure, I will replace all the boolean assignments with functions to make them more readable.

I will also move the 3D View settings to the Fly View settings in a separate group with the changes you suggested.
Regarding your question, for the moment most of the settings are quite general, so we can remove the setting icon in this case and move them back to the regular settings place. I'd like also to add more settings such as the camera movement/rotation/zoom speed.

about how to enable the 3D view, what about adding a Toggle Switch in the Fly View settings to enable the 3D view? This way, it will not be mandatory to upload an OSM file to use the 3D view. Just in case, if someone only wants to show the 3D vehicle and the waypoints. So, what do you think?

DonLakeFlyer:
> Just in case, if someone only wants to show the 3D vehicle and the waypoints.

Ah, yeah. That makes sense from a good use case.

omid-esrafilian:
@DonLakeFlyer 
I have moved the 3D View settings as a separate group to the Fly View settings as you suggested. There is now an enabled switch to activate the 3D view as well as a clear button for the OSM map file. 
Please let me know if that looks good to you, then I will update the docs accordingly.

DonLakeFlyer:
Looks fantastic, thanks

omid-esrafilian:
@DonLakeFlyer 
Thanks. So, I will update the docs for the 3D View in a separate pull request.

PR #11053, opened 2024-02-12
RemoteIDManager.cc: Fix eventual crash and refactor _getSelfIDDescription

oorahmi:
We experienced an intermittent app crash when flying with Remote ID enabled aircraft.

I noticed there are some buffers in the Open Drone ID messages that aren't being resized to the spec.
* `id_or_mac` needs to always be sized at 20 instead of {NULL}
* OPEN_DRONE_ID_SELF_ID  `description` must always be size 23
* OPEN_DRONE_ID_OPERATOR_ID `operator_id` must always be 20

Original author handled a similar buffer here:
https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/RemoteIDManager.cc#L378
but missed a few more buffers.

After implementing these changes, we no longer experienced an intermittent crash in our app.

Feel free to let me know any edits, but I wanted to notify the community in case this can prevent some future app crashes during flight.

### DIFF
--- a/src/Vehicle/RemoteIDManager.cc
+++ b/src/Vehicle/RemoteIDManager.cc
@@ -24,7 +24,7 @@ QGC_LOGGING_CATEGORY(RemoteIDManagerLog, "RemoteIDManagerLog")
 #define ALLOWED_GPS_DELAY 5000
 #define RID_TIMEOUT 2500 // Messages should be arriving at 1 Hz, so we set a 2 second timeout
 
-const uint8_t* RemoteIDManager::_id_or_mac_unknown = {NULL};
+const uint8_t* RemoteIDManager::_id_or_mac_unknown = new uint8_t[MAVLINK_MSG_OPEN_DRONE_ID_OPERATOR_ID_FIELD_ID_OR_MAC_LEN]();
 
 RemoteIDManager::RemoteIDManager(Vehicle* vehicle)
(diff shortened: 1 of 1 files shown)


Davidsastresas:
Thank you very much @oorahmi. All of it looks good to me. Could you please do a rebase so it ends up being just one commit? Also please, adding to the tittle the code refactor you did in _getSelfIDDescription? Something like:

"RemoteIDManager.cc: Fix eventual crash and refactor:

There was an intermittent app crash when flying with Remote ID enabled aircraft.
some buffers in the Open Drone ID messages aren't being resized to the spec.

- id_or_mac needs to always be sized at 20 instead of {NULL}
- OPEN_DRONE_ID_SELF_ID description must always be size 23
- OPEN_DRONE_ID_OPERATOR_ID operator_id must always be 20

After these changes it no longer crashes"

This way only the first line will appear in compact git log, but we will have the long description, in case somebody in the future is studying that file and sees the change, it is handy to have.

Thank you very much.

Davidsastresas (approved):

oorahmi:
@Davidsastresas Hey David, I made some tweaks to this PR as you mentioned. Feel free to let me know if I should do anything else. Thanks!

Davidsastresas:
It looks good to me. Thank you very much!

PR #11054, opened 2024-02-13
Update FirmwareUpgradeController.cc to add mRo Control Zero Classic board

TOTON95:
Control Zero Classic Board ID and target added


mrpollo:
Did you test this on hardware? You can download a version of QGC built by CI here. I'm not sure if this will work w/o the USB VID/PID, are those already in?

TOTON95:
@mrpollo Thank you for your quick reply, I have confirmed that the USB VID/PID are already listed into `src/comm/USBBoardInfo.json`, I am going to check for the CI Build to confirm that the FW is located successfully.

TOTON95:
@mrpollo I can confirm that it is providing the PX4 build (stable/master) for mRo Control Classic in QGC (Windows).

PR #11058, opened 2024-02-14
Add missing Quick3D lib to linux install

DonLakeFlyer:


TOTON95:
@DonLakeFlyer I can confirm that is working on Ubuntu 20.04 and 22.04, thank you!

DonLakeFlyer:
Great, thanks for the help.

PR #11069, opened 2024-02-16
Cache CI Qt Downloads

HTRamsey:
This adds the automatic caching option to the install-qt-action. Also it removes the extra linux install from the android workflow, which is handled automatically by setting `extra:        --autodesktop` in the android Qt download.

One extra note: CCache isn't working in the workflows and hasn't been for some time, I might try and fix it but it's probably better to just wait until we switch to CMake, where ninja will make ccache less relevant.

### DIFF
--- a/.github/workflows/android_release.yml
+++ b/.github/workflows/android_release.yml
@@ -53,17 +53,6 @@ jobs:
         run: |
           git fetch --all --tags -f --depth 1
 
-      - name: Install Qt6 for Linux (required for Android)
-        uses: jurplel/install-qt-action@v3
-        with:
-          version:      ${{ env.QT_VERSION }}
-          aqtversion:   ==3.1.*
-          host:         linux
-          target:       desktop
-          dir:          ${{ runner.temp }}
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
Ooh, nice. When I saw the title, first thing I thought was maybe you could fix cache caching as well :)  But waiting is fine as well. I haven't gotten back to switching CI to make. If you want to give that a try that would be awesome. It's fine to switch over one OS at a time. I have updates to get OSX to make generating a DMG so I can get that up soon,

.github/workflows/android_release.yml
```diff
@@ -53,17 +53,6 @@ jobs:
         run: |
           git fetch --all --tags -f --depth 1
 
-      - name: Install Qt6 for Linux (required for Android)
```
DonLakeFlyer, line 56: Are you sure removing this works? I couldn't get things to run without it since there were pieces missing. I think the APK file somewhere.

HTRamsey: The autodesktop option was added after the additional linux qt install, supposedly according to `https://github.com/jurplel/install-qt-action/blob/master/README_upgrade_guide.md` you only need the autodesktop option. And actually I did get ccache working on another branch that I can make a PR for. Instead of 0% hits it is like 90%

PR #11071, opened 2024-02-16
Update GStreamer versions to 1.24.0

HTRamsey:
This sets windows and android to use the same gstreamer as MacOS was using. This doesn't fix the fact that linux is still using 16.x and that CMake is set to require 1.20.
I don't see a reason why windows/android shouldn't just use the default 1.22 gstreamer w/ qt6 plugin, and use the custom gst plugins just for linux, alternatively just use QMultiMedia for Linux as their ffmpeg implementation is finally getting decent. Or can we build GStreamer from source for linux?

### DIFF
--- a/.github/workflows/android_debug.yml
+++ b/.github/workflows/android_debug.yml
@@ -99,9 +99,9 @@ jobs:
       - name:               Install gstreamer
         working-directory:  ${{ github.workspace }}
         run: |
-            wget --quiet https://gstreamer.freedesktop.org/data/pkg/android/1.18.5/gstreamer-1.0-android-universal-1.18.5.tar.xz
-            mkdir gstreamer-1.0-android-universal-1.18.5
-            tar xf gstreamer-1.0-android-universal-1.18.5.tar.xz -C gstreamer-1.0-android-universal-1.18.5
+            wget --quiet https://gstreamer.freedesktop.org/data/pkg/android/1.24.0/gstreamer-1.0-android-universal-1.24.0.tar.xz
+            mkdir gstreamer-1.0-android-universal
+            tar xf gstreamer-1.0-android-universal-1.24.0.tar.xz -C gstreamer-1.0-android-universal
 
       - name: Update android manifest
         if: github.ref_name != 'Stable'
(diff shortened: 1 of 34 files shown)


DonLakeFlyer:
@zdanek Can you take look and respond to questions?

zdanek:
Sure.

HTRamsey:
Which part are you answering "sure" to lol

zdanek:
> Which part are you answering "sure" to lol

yeah, seems like insane answer :D  I'm sorry.

> I don't see a reason why windows/android shouldn't just use the default 1.22 gstreamer w/ qt6 plugin, and use the custom gst plugins just for linux, 
 
I'm with that. Leaving old gst version for linux is for compatibility with a Ubuntu 20.04 and higher. Ubuntu 20.04 has 1.16 AFAIK. I have Ubuntu 22.04 with 1.20 by default.

> alternatively just use QMultiMedia for Linux as their ffmpeg implementation is finally getting decent. Or can we build GStreamer from source for linux? 

I'd skip that if we don't need to. Maintenance of own gst build is another burden. We don't use any sophisticated pipelines, so let's stick with defaults.

zdanek:
@HTRamsey did you check video on both platforms or just set sane versions in build scripts?

HTRamsey:
I've checked a bunch of different versions with Android, but I'd have to set up an environment to check with Windows. I also checked 1.20 with linux but that doesn't help since I believe 1.21 is when Qt6 compatibility was added. If your custom gst-good-plugins are backported and working then I think the best thing to do would be to update Windows & Android (I know nothing about MacOS) since I imagine the user base for those are larger than for linux and at least they would get the best possible video experience.

I know the UVC video uses QtMultiMedia, so I have no idea if there is a conflict if someone tries to use a GStreamer backend for QTMultiMedia along with the GStreamer Qt6 plugin.

DonLakeFlyer:
Why can't we move to the latest GStreamer, whatever it is, and then on Linux we include the libs in the AppImage?

DonLakeFlyer:
Moving to at least 1.20 also gets us OSX ARM support

HTRamsey:
I can update this based on our discussion after the CMake stuff. It gets annoying having to update two build systems.

HTRamsey:
Is anybody able to test this on Windows/MacOS?

PR #11073, opened 2024-02-16
Create build upload action

HTRamsey:
Makes a common action for the release workflows

### DIFF
--- a/.github/actions/upload/action.yml
+++ b/.github/actions/upload/action.yml
@@ -1,0 +1,38 @@
+name: 'Upload Release'
+description: 'Uploads a built release file'
+inputs:
+  artifact_name:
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
Ooh, nice. Looks like it's conflicting with some of your other changes with I just merged. Clean up and I'll merge.

DonLakeFlyer:
Yes, squash merging is available. But I generally don't like it because it creates crap commit messages. Single commits aren't required for a pull. Commits should be broken up into functional aspects as make sense. And a single commit should be revertible. I just don't like commits which include basically WIP tracking. Those are the ones that don't squash that well and generate all sorts of noise messaging in the commit message.

Issue #10968, opened 2024-01-15
Stable_V4.3 CI/CD Android is failing


jnomikos:
The following template is for Issue/Bug reporting only. https://docs.qgroundcontrol.com/en/Support/Support.html#github-issues

For questions about how to use or build QGC see: http://qgroundcontrol.com/#resources

If you are using an old stable release please upgrade first and verify your issue before reporting.
----

## Expected Behavior
This CI/CD should work, but it is failing

## Detailed Description
Looks like something might have updated and caused the build to now fail. It is saying compileSdkVersion needs to be specified. I can try specifying this to see if it fixes the error

## Log Files and Screenshots

```
FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* Where:
Build file '/home/runner/work/_temp/shadow_build_dir/android-build/build.gradle' line: 37

* What went wrong:
A problem occurred evaluating root project 'android-build'.
> For input string: "ext10"

* Try:
Run with --stacktrace option to get the stack trace. Run with --debug option to get more log output. Run with --scan to get full insights.
==============================================================================

2: Task failed with an exception.
-----------
* What went wrong:
A problem occurred configuring root project 'android-build'.
> compileSdkVersion is not specified. Please add it to build.gradle

* Try:
Run with --stacktrace option to get the stack trace. Run with --debug option to get more log output. Run with --scan to get full insights.
==============================================================================

* Get more help at https://help.gradle.org/

BUILD FAILED in 17s
Building the android package failed!
make: *** [Makefile:2837: libQGroundControl_arm64-v8a.so] Error 14
Error: Process completed with exit code 2.
##[debug]Finishing: Build
```

PR #11074, opened 2024-02-16
Fix Stable_V4.3 android build workflow

jnomikos:
https://github.com/mavlink/qgroundcontrol/issues/10968

android-34-ext10 was being installed which was causing the build step on Stable-V4.3 to fail. 

I noticed similar issues were experienced in the past here: https://github.com/mavlink/qgroundcontrol/pull/10848

With this change I managed to get past the build step successfully. However, every branch our fork fails at the "Upload build to S3 Bucket" step, so I cannot confirm with full certainty that it will succeed.

Here's that workflow run if you are curious:
https://github.com/AscendEngineering/qgroundcontrol/actions/runs/7936789857/job/21672792328


angtsusiong:
Here is my fork with removing upload to the S3 step, the line of uninstalling android-34-ext10 works well.

image
https://github.com/angtsusiong/qgroundcontrol/actions/runs/7855162091/workflow
https://github.com/angtsusiong/qgroundcontrol/actions/workflows/android_release.yml?query=branch%3AStable_V4.3

jnomikos:
Confirmed it works on my end as well.

DonLakeFlyer:
Yup, this is good

PR #11076, opened 2024-02-17
Correct Precompiled Headers Usage

HTRamsey:


DonLakeFlyer:
cmake builds are broken for me at least on OSX. Couple things:
* Could you add additional GitHub actions which at least just build using cmake for maybe just one of the desktop version. We can use qmake builds still for now to push to CI. But one cmake action in there would keep the cmake build much more clean at least.
* Are you on the PX4 discord server? Would be good to be on there for dev discussions.

DonLakeFlyer:
```
Exception has occurred: FATAL_ERROR
CMake Error at src/CMakeLists.txt:30 (target_precompile_headers):
  Cannot specify precompile headers for target "qgc" which is not built by
  this project.
```

If I comment that out the next problem moves to:
```
Exception has occurred: FATAL_ERROR
CMake Error at /Users/don/Qt/6.6.1/macos/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:305 (add_custom_target):
  add_custom_target cannot create target "qgctest" because another target
  with the same name already exists.  The existing target is a static library
  created in source directory "/Users/don/repos/qgroundcontrol/test".  See
  documentation for policy CMP0002 for more details.
Call Stack (most recent call first):
  /Users/don/Qt/6.6.1/macos/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:313 (qt6_add_binary_resources)
  test/CMakeLists.txt:75 (qt_add_binary_resources)
```

HTRamsey:
Yeah actually I did have a PR that finished the cmake build, but there was starting to be too many conflicts with my other stuff. Lemme see what I can do with the cmake stuff real quick now that the other stuff is merged. The first issue here was having the precompiled headers part before the lib was created. And yeah I am but I've never really used it, I'll look into it.

PR #11078, opened 2024-02-17
Separate Tests from Source

HTRamsey:
Testing is in a bit of a rough shape, moving it out of the source code and creating a mirrored test folder will help make it easier to work with.


DonLakeFlyer:
Seems reasonable

PR #11082, opened 2024-02-19
Android: Update Serial Implementation

HTRamsey:
@DonLakeFlyer I am just making this public in case you have comments. This will probably take a bit of time to finish. I can break it down into multiple PRs when I am done.


justas-:
Following this. Willing to test on our custom Android-based RC controllers.

DonLakeFlyer:
Is this just a goal of better repo organization as well? Or are there functionals changes here as well?

HTRamsey:
The biggest change will be much better USB support. The cleanup is a consequence of me replacing the dated custom QSerialPort implementation and custom implementation of the [usb library](https://github.com/mik3y/usb-serial-for-android) with a simplified interface. Although I did move a couple files around just to better structure the android packaging.

DonLakeFlyer:
Oh, ok. That's all great stuff.

HTRamsey:
It will be a little easier if I can get us switched to CMake entirely first, so that I don't have to update both build systems. I will return to this later.

HTRamsey:
> Following this. Willing to test on our custom Android-based RC controllers.

#11536 may interest you

PR #11086, opened 2024-02-20
Fix incorrect precompiled header including for cmake

s-lisovenko:
Should be included for known target

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,13 +286,6 @@ else()
     )
 endif()
 
-if(BUILD_TESTING)
-    add_subdirectory(test)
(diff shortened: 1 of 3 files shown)


HTRamsey:
The executable really should be made first before including these libraries, and the precompiled headers should be done on the executable instead of the library, which is why I messed this up. Thanks for the fix.

HTRamsey:
Could you also remove the `qt_add_binary_resources(qgctest UnitTest.qrc)` in test/CMakeLists.txt and put `UnitTest.qrc` right under `UnitTestList.cc` in the `qt_add_library`?

DonLakeFlyer:
I manually verified that this fixes the cmake build.

PR #11091, opened 2024-02-21
Rework instrument and video cluster

DonLakeFlyer:
Screenshot 2024-02-21 at 12 10 04 PM

* Telemetry values and instruments are merged together aligned to bottom right edge. Custom builds folks may remember this look being in the old custom build example.
* Photo/Video control is aligned to top/right edge
* Goal is to keep center of screen as clear as possible
* Also created new layout mechanism for custom builds to more easily add to ui layer. They can resource override FlyViewTopRightColumnLayout.qml and/or FlyViewBottomRightRowLayout.qml to to add their own controls into the flow. The custom widget layer continues to work as before as does all the QGCToolInsets stuff. But these new mechanisms allow for less upstream change and simpler mods to get what you want.

More to come in cleaning up the Photo/Video control to not be such a large monstrosity.

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -220,6 +220,7 @@
         <file alias="QGroundControl/FactControls/qmldir">src/QmlControls/QGroundControl/FactControls/qmldir</file>
         <file alias="QGroundControl/FlightDisplay/FlightDisplayViewVideo.qml">src/FlightDisplay/FlightDisplayViewVideo.qml</file>
         <file alias="QGroundControl/FlightDisplay/FlyView.qml">src/FlightDisplay/FlyView.qml</file>
+        <file alias="QGroundControl/FlightDisplay/FlyViewBottomRightRowLayout.qml">src/FlightDisplay/FlyViewBottomRightRowLayout.qml</file>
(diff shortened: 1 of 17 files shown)


s-lisovenko:
@DonLakeFlyer I think this is not a very good solution. Placing the camera controls in the top of screen will cause discomfort to users who use a tablet larger than 10" (for example assembled GCS with integrated joystick).

DonLakeFlyer:
> Placing the camera controls in the top of screen will cause discomfort to users who use a tablet larger than 10"

Why? Because of where their hands are and the sticks?

DonLakeFlyer:
Merging in to continue work. I'll move camera controls again if it turns out to be crap.

s-lisovenko:
> > Placing the camera controls in the top of screen will cause discomfort to users who use a tablet larger than 10"
> 
> Why? Because of where their hands are and the sticks?

Screenshot from 2024-02-21 22-28-51
Screenshot from 2024-02-21 22-28-41

For example this station from aeronav. The optimal area for your thumb to work is from the bottom right corner to the center of the screen

DonLakeFlyer:
Yeah, I see that now. Let me work on it.

s-lisovenko:
> Yeah, I see that now. Let me work on it.

Thank you!

PR #11108, opened 2024-02-23
PendingIntent needs IMMUTABLE flag from Android12

tajisoft:
App crashes on startup on Android 12.

https://developer.android.com/privacy-and-security/risks/pending-intent

Exception
Screenshot 0006-02-23 at 23 53 31


DonLakeFlyer:
@HTRamsey I think this makes sense. What do you think?

HTRamsey:
Yup, I had to do the same in my draft PR. I never tested it but I assume this should be fine.

PR #11115, opened 2024-02-24
Update for Qt 6.6.1, resolves #11104

### LINKED ISSUE #11104: Build QGroundControl in container.
I'm following the Developer's Guide of QGroundControl.

For the part of build using Containers.
I was able to make the docker image by the docker file in /deploy/docker/Dockerfile-build-linux.

but when I tried to build the android apk file with it. the error shows like:
image

I know QGC has been updated to QT 6.6, but the docker file seems like still install the QT 5.15.2.
Is that a bug? or building with docker has been deprecated?
how should I fix that?
Thank you.

clydemcqueen:
I'm not a Qt expert, but this change worked for me.

Another option is to add `ARG QT_MODULES="all"`, installs more, but future-proof.


DonLakeFlyer:
Thanks

Issue #11098, opened 2024-02-22
Fix cmake correctly such that MockLink and Unit Test are available in all Debug builds


DonLakeFlyer:
In this pull #11097 I hacked back in the ability to use MockLink in debug builds so I could make forward progress. This puts back QGC to where it was before where a Debug build always included MockLink and unit tests. There is probably some better cmake standard way to do this which I don't know about.

HTRamsey:
Yup, I'll check it out this weekend

PR #11124, opened 2024-02-26
Make Build Testing an option for debug builds

HTRamsey:
#11098
You had moved UnitTest.qrc but I don't believe that was necessary. This just makes build testing a cmake option that defaults to true when building in debug

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,9 @@ else ()
     add_definitions(-DDAILY_BUILD)
 endif ()
 
+include(CMakeDependentOption)
+cmake_dependent_option(BUILD_TESTING "Enable testing" ON "CMAKE_BUILD_TYPE STREQUAL Debug" OFF)
+
 #######################################################
 #               OS Configuration
 #######################################################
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
Thanks

DonLakeFlyer:
Don't quite understand why. The cmake config says "Building tests" so it's heading down the right path.

DonLakeFlyer:
Adding back the append to QGC_RESOURCES fixes it

HTRamsey:
Oh probably where it says `target_link_libraries(qgc PRIVATE qgctest)` it needs to be `target_link_libraries(qgc PUBLIC qgctest)`. Can you try that real quick? Github is being crazy slow for me right now

DonLakeFlyer:
Ok, I just tried that:
```[build] ninja: error: dependency cycle: cmake_object_order_depends_target_GeoTest -> cmake_object_order_depends_target_MissionManagerTest -> cmake_object_order_depends_target_qgcunittest -> cmake_object_order_depends_target_QmlControlsTest -> qml/qmlcontrolstest/QmlControlsTest.qmltypes -> test/FactSystem/meta_types/qt6factsystemtest_debug_metatypes.json -> test/FactSystem/meta_types/FactSystemTest_json_file_list.txt -> test/Geo/libGeoTest.a -> test/Geo/CMakeFiles/GeoTest.dir/GeoTest_autogen/mocs_compilation.cpp.o -> cmake_object_order_depends_target_GeoTest```

HTRamsey:
alright well I guess we'll just leave things the way they are then lol

PR #11125, opened 2024-02-26
docs for 3D View is updated

omid-esrafilian:
@DonLakeFlyer 
I have updated the docs for how to use the 3D View in this pull request. I have also made some minor changes in the code.


DonLakeFlyer:
@omid-esrafilian I don't remember seeing how much qDebug output you have in your code. If there is a bunch would be good to put the behind a logging category so it can be turned on if needed as opposed to always being output.

omid-esrafilian:
@DonLakeFlyer 
I have just a few qDebug. But yes, sure, I can do that. Is there any standard form of doing it in QGC code? Or if you have any example that I can take a look at, by any chance? Otherwise, I can just define a macro for instance.

PR #11128, opened 2024-02-26
Fix cmake build after migration to qt_add_library

s-lisovenko:
By default, qt_add_library creates a library type based on how Qt was built - in our scenario, it's set to SHARED.


DonLakeFlyer:
I verified that this work by building it myself. No need to wait for it to finish CI since it has no effect on CI yet.

PR #11133, opened 2024-02-26
Fixes for joystick event handling in QML

s-lisovenko:
This update changes the method signatures for "onAxisValueChanged", "onRawButtonPressedChanged", and "onAxisValues" events in QML according to Qt6 requirements. Each event now directly accepts the necessary arguments.


DonLakeFlyer:
Thanks. These things are like to plague! I don't quite understand the benefit of this Qt change. What a PITA.

PR #11135, opened 2024-02-26
Update CMake Deployment

HTRamsey:
This will need a follow up after some more testing, but at least for now the linux_debug workflow has been updated to use CMake and runs the unit tests correctly. Getting this PR merged first will let us verify future CMake changes easier. The other workflows can be converted to CMake after we verify all of the deployment properties for each platform are set correctly.

Also added qtconnectivity which was required for bluetooth.
Ultimately I would like VideoReceiverApp to share the same root CMakeLists.txt and just be a separate executable.

### DIFF
--- a/.github/workflows/linux_debug.yml
+++ b/.github/workflows/linux_debug.yml
@@ -50,24 +50,21 @@ jobs:
           host:         linux
           target:       desktop
           dir:          ${{ runner.temp }}
-          modules:      qtcharts qtlocation qtpositioning qtspeech qt5compat qtmultimedia qtserialport qtimageformats qtshadertools qtquick3d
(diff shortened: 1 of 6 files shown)


HTRamsey:
@DonLakeFlyer I will try to correctly set up everything for MacOS deployment in the follow up to this but I will need your help verifying it.

DonLakeFlyer:
Fantastic, thanks so much

DonLakeFlyer:
The ultimate goal for the Linux Debug and Test workflow would be to build the AppImage (right from cmake, not separate script) and the run the units using the generated AppImage. That way we sanity check the AppImage creation as well.

DonLakeFlyer:
Let me know when you need me to look at Mac. Right now I'm setting up a new Linux box so I can get my head around Mavlink Cameras and Video. 

I've been putting off having to know anything about video for years now. It's such a rats nest though I really need to get my head around it.

PR #11141, opened 2024-02-27
QGCTabButton text variant fix

ben-draper-ascent:


src/QmlControls/QGCTabButton.qml
```diff
@@ -32,7 +32,7 @@ TabButton {
     property bool   _showHighlight:     enabled && (pressed | checked)
     property int    _horizontalPadding: ScreenTools.defaultFontPixelWidth
     property int    _verticalPadding:   Math.round(ScreenTools.defaultFontPixelHeight * heightFactor)
-    property bool   _showIcon:          control.icon.source !== ""
+    property bool   _showIcon:          control.icon.source != ""
```
DonLakeFlyer, line 35: Ah, ok. I see the problem now. icon.source starts out as "undefined" I assume so the !== fails.

DonLakeFlyer:
Thanks

PR #11142, opened 2024-02-27
Add macos debug workflow

HTRamsey:
I have literally no idea how to use a mac but at least this should build with cmake.


DonLakeFlyer:
Awesome, hopefully I can take this the rest of the way to dropping a DMG from it.

Issue #11020, opened 2024-02-06
Video streaming seems not work in windows qt6 qgc


hyeonseung89:
## Expected Behavior
Want to RTSP Video source play and work at qt6 QGC
test source is rtsp://********:8554/*******

## Current Behavior
Recently saw that gstreamer fixes for qt6 qgc were merged.  But it doesn't seem to play in Windows apps. I haven't tried any other OS.

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior

1. Download generated latest app by CI action
2. In video settings, set source to rtsp 
3. input address

## System Information
When posting bug reports, include the following information
- Operating System: Windows 11 Home 22H2
- Graphic Card: rtx3060ti
- QGC Version: master branch
- QGC build: daily
- gstreamer version : 1.18.1
- Qt 6.6.1 msvc 2019 64bit

## Detailed Description
When I compiled it myself up to qt 5.15, it worked normally.

## Log Files and Screenshots
- Screenshots of QGC console.
![image](https://github.com/mavlink/qgroundcontrol/assets/159096725/b8500d37-28df-448c-bfe4-602512b1f427)

DonLakeFlyer:
@zdanek Do you have the ability to test under Windows?

zdanek:
Yes I do.

DonLakeFlyer:
Thank so much for looking at this.

AlexKlimaj:
```[!] at D:\a\qgroundcontrol\qgroundcontrol\src\VideoReceiver\GstVideoReceiver.cc:1370 - "GStreamer error: Could not initialize window system"```

DonLakeFlyer:
I vaguely seem to remember fixes in the qmlsink thingy for Windows. Maybe we lost those somewhere along the way?

NiiAAip:
I obtained some important information through https://www.qt.io/blog/graphics-in-qt-6.0-qrhi-qt-quick-qt-quick-3d. Qt Quick aims to accelerate rendering using hardware. By default, it will be built on the low-level graphical API that is most suitable for the target platform. For example, on Windows, it will default to Direct3D, while on macOS, it will default to Metal. 

So here, this line of code cannot run properly.
https://github.com/mavlink/gst-plugins-good/blob/5ff0a6af0e1fd29bafc9f6644e728d24349d0ce7/ext/qt6/qt6glitem.cc#L527

[...]

PR #11145, opened 2024-02-27
Mac video cmake plus video streaming change

DonLakeFlyer:
* cmake wasn't building video code
* Fixed OSX app bundle permissions for camera/location
* Set Qt to use OpenGL for rendering backend to fix video streaming. Related to #11020


HTRamsey:
I've got a MacOS cmake build workflow added that will hopefully help with stuff like this. I think there's an issue with SDL for MacOS too.

PR #11147, opened 2024-02-28
Create QtInstallerFramework Installer

HTRamsey:
It'll look much better when it's actually configured fully, and running on windows or something. But this is just a quick demo of what it looks like by default. I don't have a windows example to show, but it will create the shortcut and provide an uninstaller/updater and everything.
image
image

Added dependencies will go here
image
image
image

Then there is the maintenance tool in the installation folder, which works exactly like the Qt Maintenance Tool.
image


DonLakeFlyer:
Is it able to launch other installers? Right now QGC installs drivers as a separate msi installer package. That said, I'm not even sure if those drivers are really needed. Could be just ancient history on old windows versions.

HTRamsey:
Yup, we can include the driver.msi as an option in the "Select Components" screen then I believe it just triggers that installation normally. I have to look through the manual a bit more to figure out how to do everything though. I'm pretty sure you just add logic to the installscript.js for each package for anything custom like that.

There's also the controller.qs that can be used for stuff related to the installer functionality overall.

I don't believe any of the stuff that's in those files currently does anything, it was just sample code from the guide I had copied as a reference for a starting place. So all of this is running correctly without any custom logic so far which is nice.

HTRamsey:
@DonLakeFlyer It won't be used yet, but maybe we should merge this in case you or anybody else wants to work on the design. I'll come back to adding driver installs and such after some of the more urgent stuff gets done

DonLakeFlyer:
Sure

PR #11148, opened 2024-02-28
Update CI Workflows

HTRamsey:
Uses the new ubuntu setup script to install all dependencies
Builds android using the latest ubuntu
Moves saving PDB to windows debug
Corrects some variable usage
Fixes some cmake usage
Updates the visual studio shell action to an actively maintained one

### DIFF
--- a/.github/workflows/android_release.yml
+++ b/.github/workflows/android_release.yml
@@ -28,7 +28,7 @@ env:
 
 jobs:
   build:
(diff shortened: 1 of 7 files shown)


DonLakeFlyer:
FYI: For OSX I'm going to do similar to ubuntu.sh and switch to brew install for Gstreamer instead of the frameworks. That is what works with cmake for video on OSX.

PR #11153, opened 2024-02-28
Create LICENSE

HTRamsey:


s-lisovenko:
https://docs.qgroundcontrol.com/master/en/qgc-dev-guide/contribute/licences.html

HTRamsey:
Doesn't it have to be GPL because of the Qt libraries it uses? Like i know QtCharts is GPL. I'm pretty sure GPL requires that all derivative software be GPL as well. The only alternative would be removing GPL based libraries (which I think QtCharts is the only one) and only using LGPL then you could use the more permissive LGPL which is probably what any company using QGC would want to do.

DonLakeFlyer:
I don' know the specifics of all of it. All I know was that the licensing goal for QGC and PX4 is to not require contributions to come back. And everyone who submitted code when the switch was made approved to that licensing.

DonLakeFlyer:
I think it's fine to just leave it in the docs instead of trying to get the intricacies in a single license file.

HTRamsey:
Yeah that's fine. I was just trying to get it to show up in the Readme bar on the homepage.

DonLakeFlyer:
Maybe I should updates the docs to say something about Qt licensing as well.

HTRamsey:
When I was first looking in to this issue a couple years ago, I went by the most restrictive license which was basically determined by which Qt modules were used, and then I removed everything that used GPL. But I'm not a lawyer and was very confused by all this initially.

PR #11154, opened 2024-02-28
CI: Add android debug workflow

HTRamsey:
This doesn't build successfully because of a missing QtQuick3DTools lib, but I only see that available in the desktop Qt build? I'm not sure if this is a Qt bug or there's something I'm misunderstanding. Having set QT_HOST_PATH correctly, it shouldn't be related to the fact that it's cross-compiling for mobile. Although it builds for qmake so that shows I'm probably doing something wrong tho.
`Qt6Quick3D could not be found because dependency Qt6Quick3DTools could not be found.`

Here are the available Android libs:
image
And here's the same for Linux:
image

There is no Qt6Quick3DTools for Android so I don't really understand.

Also, I know there's a lot of workflows and they take forever now. Hopefully we can narrow that down soon but this at least ensures CMake will  work for everything.

### DIFF
--- a/.github/workflows/android_debug.yml
+++ b/.github/workflows/android_debug.yml
@@ -1,0 +1,129 @@
+# The 32 and 64 bit version of these actions should be kept in sync
+name: Android 32/64-bit Debug
+
+on:
+  push:
+    branches:
+      - 'master'
+      - 'Stable*'
+    tags:
+      - 'v*'
+    paths-ignore:
+      - 'docs/**'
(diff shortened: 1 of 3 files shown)


HTRamsey:
So apparently the --autodesktop option does not install all of the same modules as android, so we actually do have to install them separately until aqt fixes that

DonLakeFlyer:
So I can merge this now?

HTRamsey:
yup

PR #11163, opened 2024-02-29
Move files into their appropriate modules

HTRamsey:
I had made a PR like this previously but there was an issue with it. Should be good now. The purpose of this is to make modular isolated libraries to eventually reduce the need for everything to be linked to each other. So when you make changes, it will only rebuild that library.  Ideally all external dependencies will be PRIVATE linked into the module that uses them, and this puts shapelib in MissionManager. This eventually needs to be done with QML to make clean QML modules too since Qt6 has an easy way to include Qml files with qt_add_qml_module, rather than the qmldir nonsense.
This will also be good for unit tests, since you will be able to test that module by itself.


DonLakeFlyer:
Some of these make sense since they are truly specific to the containing class. But some of them are really generic utility functions. Like say QGCGeoCoordinate. That's pretty generic capability which could end up being used in multiple places over time. Same thing with say the KMLDom/Shape stuff. At some point we might want to load shapes into the fly view. It's been a requested feature for a while now. In that case KML or shape would be used there.

HTRamsey:
At the moment since everything is publicly linked to the base qgc project this doesn't actually change anything, but I can revert some of the changes and just do like uasmessagehandler if you want. I'm just not a huge fan of the large collection of miscellaneous files in the root src directory.

DonLakeFlyer:
> I'm just not a huge fan of the large collection of miscellaneous files in the root src directory.

Yeah, I don't disagree with that. Why don't you make Helpers directory or something like that the generic stuff.

HTRamsey:
This one should be good now

DonLakeFlyer:
Nice, I like it

PR #11167, opened 2024-03-01
Fix Linux TLS Error

HTRamsey:


DonLakeFlyer:
What does `egldeviceintegrations` do? Also is there a need for any other library copying in PostLinkCommon given there are new modules?

QGCPostLinkCommon.pri
```diff
@@ -161,7 +161,10 @@ LinuxBuild {
         position \
         sqldrivers \
         texttospeech \
-        multimedia
+        multimedia \
+        egldeviceintegrations \
+        networkinformation \
+        tls
```
julianoes, line 167: This line fixes the TLS error for me on Linux, nice!

HTRamsey:
@DonLakeFlyer https://doc.qt.io/qt-6/embedded-linux.html, probably not necessary but these three plugins were missing from the qmake build compared to the cmake build. The TLS plugin is the only one actually needed for this issue, so I can change this to just add that one if you want. There are a bunch of other libraries that the qmake build is missing compared to the cmake build but no other ones are needed to make this work.

DonLakeFlyer:
I wonder if the whole embedded Linux thing brings along with it other problems as well. That isn't something we have specifically stated as supported. Only desktop Linux and android.

HTRamsey:
Just to keep it simple and stick to the purpose of this PR I'll just add TLS

DonLakeFlyer:
@HTRamsey Thanks so much for figure this out. It has been plaguing folks for a while.

PR #11173, opened 2024-03-01
3D View fixes: showing all types of buildings as well as the ROI

omid-esrafilian:
The following changes have been made in this pull request to improve the 3D View:

- All types of buildings are now fetched from the OSM file and added to the 3D View.
- The ROI is shown with a separate marker (Letter "R") in the 3D View
- The takeoff position is also marked with the letter "T" in the 3D View

Here is a picture demonstrating the new changes:

Screenshot from 2024-03-01 20-54-23

### DIFF
--- a/src/FlightDisplay/FlyViewToolStripActionList.qml
+++ b/src/FlightDisplay/FlyViewToolStripActionList.qml
@@ -16,7 +16,6 @@ ToolStripActionList {
     id: _root
 
     signal displayPreFlightChecklist
-    property bool   _viewer3DEnabled:        QGroundControl.settingsManager.viewer3DSettings.enabled.rawValue
 
     model: [
         ToolStripAction {
@@ -28,7 +27,8 @@ ToolStripActionList {
             }
         },
         ToolStripAction {
-            property bool _is3DViewOpen: viewer3DWindow.isOpen
+            property bool _is3DViewOpen:            viewer3DWindow.isOpen
+            property bool   _viewer3DEnabled:       QGroundControl.settingsManager.viewer3DSettings.enabled.rawValue
 
             id: view3DIcon
             visible: _viewer3DEnabled
@@ -47,7 +47,6 @@ ToolStripActionList {
                     view3DIcon.iconSource =     "/qmlimages/PaperPlane.svg"
                     text=           qsTr("Fly")
                 }else{
-                    viewer3DWindow.close()
                     iconSource =     "/qmlimages/Viewer3D/City3DMapIcon.svg"
                     text =           qsTr("3D View")
                 }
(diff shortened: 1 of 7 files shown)


SebastianViollaz:
Which parameters are  the necessary to draw the object in the 3D view?

Im mapping a zone by my self and the builds are not shoed in the 3D View.
image
This are my parameters in the building (Are the same in other buildings that are showed properly in other file).
I also tried to copy and paste others buildings that are  showed properly in the same and in other files.

I share the file if its necesary:
 [LaVerbena.zip](https://github.com/mavlink/qgroundcontrol/files/14482670/LaVerbena.zip)

SebastianViollaz:
As a feature: Could be something interesting to add an alert when trying to take off while the flight path goes through a building. Maybe could be a box that  alert you that your UAV may chrash with something

SebastianViollaz:
Also if the file selector window  opens in user folder or in an specific one. The  user experience will be better.

When u wan to change or select  the file the folder opens in the SO folder
image

I think it will be better to open at least in the user folder.


Let me know about the changes.
@omid-esrafilian :3

SebastianViollaz:
> he .osm file should contain the following information

NIce, im learning to edit the osm files. Maybe the negative id could be used to draw holes?
Image a pool. Maybe could be used to draw and indoor pool for example??

SebastianViollaz:
Thanks for the response, i will lt you know any other questions/features that come into my mind. 
Tanks for everything :3

omid-esrafilian:
@SebastianViollaz 
No problem.

SebastianViollaz:
Hello again @omid-esrafilian. I have tested the new feature and i would like to let you know, in case you are interested in updating.
Here are some recommendations and ideas that i think that would be nice to keep into mind:
* The take of icon isn't properly showed (at least for me):
image

* Would be nice to add the RTL icon. At the moment the RTL isn't showed and the flight path isn't showed:
image
As you can see in the last point doesn't draw a line to the Take off item like in the 2D view.

* **3D  editor**:   Adding a 3d editor would be very useful  if you have a properly designed 3d MAP. If you are in an indoor environment using a 3D edit would be easier, and will help avoiding collisions during the flight planification. Maybe adding  some arrows to move the point like 3D design apps:
image


* **Textures**: In the osm file you can specify the material of the objects. If instead of drawing a grey box, you can set a default texture for each material the 3D view will  be exponentially more attractive for the user.



I hope this changes to be integrated and let me know if you need some help, i will create issues in this topics to see if anyone else is interested in adding this features.

Thanks :3.

omid-esrafilian:
@SebastianViollaz 
Thanks for the feedback. I have replied to each issue individually. 
However, I am planning to create a new pull request very soon introducing a new interesting feature, along with the fixes for some of the problems you mentioned.

PR #11174, opened 2024-03-02
CMake: Update compile definitions

HTRamsey:
I went through all of the qmake files to see what defines were missing from the cmake side. There were a handful of things that needed to be added as cmake options (like CONFIG_UTM_ADAPTER), and some needed to be added as defines (like __ mobile __). In doing so, I found __ android __ and __ ios __ were missing but it seemed redundant to define those when Q_OS_ANDROID and Q_OS_IOS already exist so I replaced those in the source. 

This also started to fix some of the issues with Android for cmake. There are some issues with almost all of the external libraries for Android but this fixes a couple of them and does a bit of minor rework to better organize how external libraries are included. A separate PR fixing the remaining Android specific issues will follow.

I also found QGC_DISABLE_DNSENGINE and QGC_ZEROCONF_ENABLED had no bearing on if qmdnsengine was included in LinkManager and _addZeroConfAutoConnectLink was called, the check in _addZeroConfAutoConnectLink was only based on __ios__ but qmdnsengine shouldn't even be included if it's not used.

### DIFF
--- a/libs/libevents/CMakeLists.txt
+++ b/libs/libevents/CMakeLists.txt
@@ -1,17 +1,64 @@
-find_package(Qt6 COMPONENTS Core REQUIRED)
+find_package(Qt6 REQUIRED COMPONENTS Core)
 
-add_library(libevents_generated)
-target_include_directories(libevents_generated INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/libevents/libs/cpp/generated)
+qt_add_library(libevents_generated STATIC
+    libevents/libs/cpp/generated/events_generated.h
+)
 
-add_library(libevents_parser
-        definitions.cpp
-        libevents/libs/cpp/parse/parser.cpp
-        libevents/libs/cpp/protocol/receive.cpp
-        )
-target_include_directories(libevents_parser PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(libevents_parser PUBLIC Qt6::Core)
-target_link_libraries(libevents_parser PUBLIC comm)
+target_include_directories(libevents_generated
+    PUBLIC
+        ${CMAKE_CURRENT_SOURCE_DIR}
+        libevents/libs/cpp/generated
+)
 
-add_library(libevents_health_and_arming_checks
-    libevents/libs/cpp/parse/health_and_arming_checks.cpp)
-target_link_libraries(libevents_health_and_arming_checks PRIVATE comm)
+
+qt_add_library(libevents_parser STATIC
+    definitions.cpp
+    libevents_definitions.h
+    libevents/libs/cpp/parse/parser.cpp
+    libevents/libs/cpp/parse/parser.h
+    libevents/libs/cpp/protocol/receive.cpp
+    libevents/libs/cpp/protocol/receive.h
+)
+
+add_subdirectory(libevents/libs/cpp/parse/nlohmann_json)
+
+target_include_directories(libevents_parser
+    PUBLIC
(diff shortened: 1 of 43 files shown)


DonLakeFlyer:
Can you give me an overview of the changes here?

libs/libevents/CMakeLists.txt
```diff
@@ -1,17 +1,67 @@
...
+    PUBLIC
+        libevents_parser
+)
+
+
+qt_add_library(libevents STATIC)
+
+target_include_directories(libevents_health_and_arming_checks
+    PUBLIC
+        {CMAKE_CURRENT_SOURCE_DIR}
```
s-lisovenko, line 60: ${CMAKE_CURRENT_SOURCE_DIR}

HTRamsey:
@DonLakeFlyer yeah sorry I meant to come back and add to it after verifying CI passed but I fell asleep

DonLakeFlyer:
QGC_DISABLE_DNSENGINE was something I added during the Qt6 transition so I could make forward progress. I should have removed it. QGC_ZEROCONF_ENABLED has been around for a while but for all these "enable/disable" things they are likely broken since they are almost never used.

DonLakeFlyer:
Took a quick Look through. All seems to make sense.

HTRamsey:
Should I remove QGC_DISABLE_DNSENGINE and the zeroconf one now then?

s-lisovenko:
@HTRamsey could you also fix HINTS variable?

image

HTRamsey:
@s-lisovenko good catch, I don't know why I miss those. Unfortunately the editor I use, sublimetext, doesn't support cmake syntax

s-lisovenko:
Btw, some parts of QGC do not work correctly with cmake build (for example actuators) I found hat we migrated CI to use cmake build and app may not work correctly.

HTRamsey:
Okay, we'll have to do some significant testing after the cmake implementation is done before switch the release builds to cmake. After this PR and #11154, I will run through and check all the final CMake stuff and see if there's anything wrong. I know a couple of Android properties need to be set at least, and gstreamer for android.

s-lisovenko:
> Okay, we'll have to do some significant testing after the cmake implementation is done before switching the release builds to cmake. After this PR and #11154, I will run through and check all the final CMake stuff and see if there's anything wrong. I know a couple of Android properties need to be set at least, and gstreamer for android needs to be fixed.

Thank you.

s-lisovenko:
> @s-lisovenko Actually it's probably related to how the qml modules are added since that changed in Qt6 and I haven't done much with those in cmake

Yes, most likely it is.

DonLakeFlyer:
> Should I remove QGC_DISABLE_DNSENGINE and the zeroconf one now then?

QGC_DISABLE_DNSENGINE yes, since that was supposed to be temporary. Zeroconf I would leave. It's been in there since forever. Might be some custom build folks out there using it.

HTRamsey:
@DonLakeFlyer This should be ready too now

DonLakeFlyer:
As the changes to cmake for Qt6 get more and more involved we really need to focus on switching over the cmake in CI. Otherwise keeping both build systems up to date is getting harder and harder as what you need to do in each diverge further and further away from each other.

HTRamsey:
@DonLakeFlyer I think most of the major things should be done now. Probably the main limiting issue would be getting android linked with gstreamer in cmake, and I have a draft PR up for that. The qml module stuff may not really be required, as if it were then nothing would work. So maybe I finish that gstreamer stuff then we just test and see if there are any other problems?

PR #11182, opened 2024-03-03
CMake: Create linux appimage

HTRamsey:
I'm not sure linuxdeployqt is needed anymore because of the built in qt_generate_deploy_qml_app_script, but let's try it out anyway. I can only test via CI because the AppImageToolkit/linuxdeployqt guys are hostile about supporting newer os. I believe [this](https://github.com/linuxdeploy/linuxdeploy-plugin-qt) might be the better option

Edit: I have several different methods commented out to try in the cmake file.

### DIFF
--- a/.github/workflows/linux_debug.yml
+++ b/.github/workflows/linux_debug.yml
@@ -84,4 +84,4 @@ jobs:
 
       - name: Run unit tests
         working-directory:  ${{ runner.temp }}/shadow_build_dir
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
So is this draft then. You're just testing, don't merge?

HTRamsey:
Good question, I cleaned it up a little just now but left some of the commented out stuff in the .cmake just in case. It seems to work as-is right now.

DonLakeFlyer:
OK, I'll merge then

PR #11195, opened 2024-03-05
Qml: Register Elements via Qt6 method

HTRamsey:
This registers qml elements in the Qt6 way, I have left the QML_SINGLETON's commented out for now to avoid create duplicate singletons until we are ready to remove the `qmlRegisterSingletonType` code.

### DIFF
--- a/src/AnalyzeView/GeoTagController.h
+++ b/src/AnalyzeView/GeoTagController.h
@@ -77,6 +77,7 @@ private:
 class GeoTagController : public QObject
 {
     Q_OBJECT
+    QML_ELEMENT
 public:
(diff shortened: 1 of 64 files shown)


DonLakeFlyer:
But does this mean that the registration in QGCApplication goes away? If so, how does the new way connect an element to a module?

HTRamsey:
Yup they all go away, when you do qt_add_qml_module in cmake you specify all of the qml files in the module, its resource path, imports, etc. Then it autogenerates all of the qmldirs pointing to each of your qml files. I think you can just point it to the .qrc file as well.

PR #11198, opened 2024-03-05
Do Not Merge: Example for herelink camera switching

DonLakeFlyer:
Screenshot 2024-03-05 at 12 31 37 PM

@julianoes This will need some adjustment for the Herelink custom build overrides. But it should point you in the right direction.


HTRamsey:
Looks like this was just an example? Reopen if you still need it

Issue #11205, opened 2024-03-06
Custom Joystick actions on Android using JoystickMavCommands.json 


ivanzivotic:
## Expected Behavior
For joystick control, there is an option for defining custom Mavlink messages that can later be assigned for the joystick buttons, enabling the user to send custom messages to the vehicle on button click. As I saw in the code that these message definitions are read from the file named JoystickMavCommands.json (line 120 of current master branch in Joystick.cc file), I was able to define custom messages on Windows and Ubuntu OS by putting the file to the working directory from which QGC is being started. However, this is not possible to do on Android as the process of starting an application (QGC in this case) is not done that way. Therefore, I am not sure where should the JSON file be located and if the QGC application even has the permissions for reading it.

## Current Behavior
Currently, there is no way to add this file on Android.

## System Information
When posting bug reports, include the following information
- Operating System: Android 14 (Samsung OneUI 6.0)
- QGC Version: 4.3.0
- QGC build: stable
- Autopilot (with version): Ardupilot Copter 4.3.5

Maybe the fix for this issue would be to set a static path from which the application will try to read the file and instruct users to always use that exact path for configuring these custom messages.

DonLakeFlyer:
This file should be coming from somewhere in the regular Documents/QGroundControl locations not the app root.

DonLakeFlyer:
Also double-check where CustomActions.json is loaded from.

DonLakeFlyer:
The "Enable Custom Actions" settings seems to have been lost in the conversion to new UI style. Need to add it back. There is a CustomActions directory where these should be loaded from.

maxfinbow:
This would be useful. I have a couple of android terminals and would greatly benefit from the custom joystick actions. The current instruction in the GUI is to place the file in /storge/emulated/0/Android/data/org.mavlink.qgroundcontrol/files/QGroundcontrol/MavlinkActions

I can't figure out how to access that path, and from the comments above I assume the intention is to rather have it in a standard documents folder. I see this thread is quite old - doesn't look like this has been done as of v5.0.8. I did create a 'QGroundControl/MavlinkActions/' folder in documents, and put my json file in there but it didn't seem to work

DonLakeFlyer:
The settings is missing and needs to be added back before this will work

DonLakeFlyer:
> The settings is missing and needs to be added back before this will work

No, that is wrong. The setting to select joystick actions is under Fly View Settings, which is strange but it it what it is for now. 

[...]

PR #11207, opened 2024-03-06
Merge fly view and joystick custom actions in a single backend

DonLakeFlyer:
Currently the Fly View and Joysticks have support for defining and executing arbitrary mavlink commands. Right now those are two different systems which work differently and use different file formats/location. This pull combines the two into a single backend which works the same way for both.

Changes:
* Custom action json files are always loaded from the `CustomActions` directory of the normal QGC save path.
* The Fly View loads the file named `FlyViewCustomActions.json`
* The Joysticks load the file named `JoystickCustomActions.json`
* None of the above things are configurable
* The existence of the files are optional

Change in function from before:
* There is no longer a setting to turn off Fly View custom actions
* There is no longer a setting to specify the Fly View custom actions file
* The thinking is that was unnecessary additional complexity which doesn't add much benefit

ToDo:
* Warn the user who was doing it the old way that it has changed and they will need to update their files/locations.

Related to #11205


DonLakeFlyer:
@nexton-winjeel FYI: As to whether to change to allow filename setting is a problem?

nexton-winjeel:
> FYI: As to whether to change to not allow filename setting is a problem?

@DonLakeFlyer: It won't break anything, but it will be an annoyance. The reason we have the ability to select different files from the UI was so that we use the one GCS setup for multiple vehicle configs, and quickly swap between custom actions if we change the active vehicle.

DonLakeFlyer:
Hmm, that's a reasonable use case. I'll put back the ability to switch files.

DonLakeFlyer:
Updates:
* You can pick the Fly View and Joystick action file from Fly View Settings
* User is warned about needing to redo their custom actions settings

Screenshot 2024-03-07 at 10 34 32 AM

DonLakeFlyer:
@nexton-winjeel Can you help with testing this?

nexton-winjeel:
> @nexton-winjeel Can you help with testing this?

@DonLakeFlyer: I pulled the Linux AppImage from https://github.com/mavlink/qgroundcontrol/actions/runs/8197777789/artifacts/1308087842. Confirmed:
* can load and use custom JSON for Fly View;
* can load and use custom JSON for Joystick;
* can change custom JSON at runtime for Fly View; and
* can change custom JSON at runtime for Joystick.

One thing I did note: The `description` field is now required, whereas previously it was optional. The `description` field isn't used by the Joystick side, so may force users to enter an unused field in their custom files.

DonLakeFlyer:
> The `description` field isn't used by the Joystick side, so may force users to enter an unused field in their custom files.

Yeah, that's slightly odd. But I wanted the files to be shareable between joystick and fly. Seems useful to just use a single file for both.

ivanzivotic:
@DonLakeFlyer thanks for this update. Just a note, I think there is a mistake in the docs/en/SUMMARY.md path, causing the docs website to break when trying to open the custom actions tab. Namely, there is an extra '(' at the link start and missing 's' in the filename before the .md extension

DonLakeFlyer:
@ivanzivotic Thanks for letting me know: #11246

Issue #11215, opened 2024-03-08
Joystick Gimbal Up and Gimbal Down button actions should use degrees (not centi-degrees)


rmackay9:
As first reported [here on the ArduPilot forums](https://discuss.ardupilot.org/t/mount-invalid-angle-targets-problem/113885), QGC's Joystick Button actions for "Gimbal Up" and "Gimbal Down" seems to be sending MAV_CMD_DO_MOUNT_CONTROL but with the angles specified in centi-degrees instead of degrees.

![ec50a0ad3800d625643507c98af82ab422e6febc](https://github.com/mavlink/qgroundcontrol/assets/1498098/c3ea6b27-f238-4cb0-a88e-552afdb7fd55)

Thanks to @nexton-winjeel for providing the exact problem which is [here](https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L4264).  We should at least remove this little "apmFirmware()" section but another solution could be to migrate to the MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW command.

By the way, I think this "bug" in QGC was originally added because of a bug in ArduPilot.  [That bug was fixed in Aug 2022](https://github.com/ArduPilot/ardupilot/pull/21530) and went out with Copter-4.3.  Copter-4.4 is the stable version now and we're getting close to releasing 4.5 so the fix has been out there for a long time now.

rmackay9:
Thanks very much!!

PR #11216, opened 2024-03-08
Vehicle: Remove centi-degrees conversion in MOUNT_CONTROL

nexton-winjeel:
Fixes: #11215

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -4262,12 +4262,6 @@ void Vehicle::_altitudeAboveTerrainReceived(bool success, QList<double> heights)
 
 void Vehicle::gimbalControlValue(double pitch, double yaw)
 {
-    if (apmFirmware()) {
-        // ArduPilot firmware treats this values as centi-degrees
-        pitch *= 100;
-        yaw *= 100;
-    }
-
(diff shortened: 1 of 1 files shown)


rmackay9:
Thanks very much for this!

The incorrect use of centi-degrees was fixed in ArduPilot more than 18 months ago.  Thanks!

nexton-winjeel:
@rmackay9: Can you confirm what version of ArduPilot this was fixed in?

@DonLakeFlyer: Are you happy with this fix as-is, or do you want to incorporate a version check so that it can use the legacy behaviour for old versions of ArduPilot?

DonLakeFlyer:
Thanks for the change. It's going to need a version check for back-compat.

rmackay9:
Hi @nexton-winjeel,

The ArduPilot fix was released with 4.3.0 which went out on 31-Oct-2022.  Since then we've also release 4.4 and we're getting close to releasing 4.5.  I still occasionally hear of users back on 4.2 and earlier but not often.  Most users are on 4.4 while companies tend to lag a bit but even most of them are on 4.3 or 4.4.

DonLakeFlyer:
That's pretty far back then. I think it's fine without a version check then.

rmackay9:
Thanks so much!

PR #11220, opened 2024-03-09
CMake: Fix Android Executable

HTRamsey:
At first I was noticing an issue with where the Viewer3D qmldir was. I found out it was related to QT_QML_ROOT_PATH, so I changed that value. Then I had to fix the android app not finding the main function, which meant moving main.cc to the root executable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,7 +183,7 @@ endif()
 #               QGroundControl Target
 #######################################################
 
-qt_add_executable(${PROJECT_NAME} ${QGC_RESOURCES})
+qt_add_executable(${PROJECT_NAME} src/main.cc ${QGC_RESOURCES})
 
 set_target_properties(${PROJECT_NAME}
     PROPERTIES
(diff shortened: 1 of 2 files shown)


omid-esrafilian:
@HTRamsey 
May I ask what was the issue with Viewer3D qmldir? Then I can fix it.

HTRamsey:
@omid-esrafilian I think we're fine now. I think it was expecting the first qmldir in the root Viewer3D folder. It won't matter with just CMake anyways when the qmldirs are autogenerated

omid-esrafilian:
@HTRamsey 
Thanks for the explanation.
I also realized that Qt Creator could not recognize the Viewer3D modules, but there was no compilation error or any problems during the code runtime.
So, I did a little digging and it appears that all the qmldirs should be placed in the src/QmlControls directory, as it is the only path imported by QML_IMPORT_PATH. I moved them there and Qt Creator can find them now. However, still the QGroundControl.Viewer3D modules that I defined using "qmlRegisterType" reminded the same and cannot be recognized by Qt Creator. Despite this, there are no compilation errors and the code ran without any issues.

PR #11221, opened 2024-03-09
Fix comm link switching

### LINKED ISSUE #10815: Can’t change between multiple links
When I have a UDP and serial link to one vehicle, I’m unable to change between them successfully. 

If I select the other link (the one I’m not on) it doesn’t change. I have to go to comm links and hit disconnect which isn’t the intention. 

Go back in version, I see it working around 4.0.10

DonLakeFlyer:
* Fix #10815
* Added autoconnect settings to Comm Links page
* Modified the LinkManager autoconnect so that it doesn't spew continuous autoconnect wait logging


ryanjAA:
Amazing! Will test ASAP. :) 👍 

Thanks!

PR #11222, opened 2024-03-09
Remote id indicator: Convert to new UI style, plus other changes

DonLakeFlyer:
* Converted Remote ID Indicator to new style
* Removed the remote id enable setting:
  * Remote ID settings page is always shown
  * If comms are good in the remote id manager open drone id message are sent to the vehicle
* Added RemoteIDManager.available property. This indicates whether the vehicle support Open Drone ID of not
* Remote ID toolbar indicator will not be shown if the vehicle does not support Mavlink Open Drone ID

ToDo:
* Rework remote id settings to the new ui style
* Allow remote id settings to be configured in expanded page of the toolbar indicator

@Davidsastresas FYI: Other than the enable thing, the functionality is exactly the same. Just visual changes.

Screenshot 2024-03-09 at 12 22 47 PM

### DIFF
--- a/qgroundcontrol.qrc
+++ b/qgroundcontrol.qrc
@@ -107,6 +107,7 @@
         <file alias="QGroundControl/Controls/LabelledButton.qml">src/QmlControls/LabelledButton.qml</file>
         <file alias="QGroundControl/Controls/LabelledComboBox.qml">src/QmlControls/LabelledComboBox.qml</file>
         <file alias="QGroundControl/Controls/LabelledLabel.qml">src/QmlControls/LabelledLabel.qml</file>
+        <file alias="QGroundControl/Controls/RemoteIDIndicatorPage.qml">src/ui/toolbar/RemoteIDIndicatorPage.qml</file>
         <file alias="QGroundControl/Controls/SettingsGroupLayout.qml">src/QmlControls/SettingsGroupLayout.qml</file>
         <file alias="QGroundControl/Controls/InstrumentValueLabel.qml">src/QmlControls/InstrumentValueLabel.qml</file>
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
@Davidsastresas Let me know if I'm missing something about the enable setting and I need to put it back.

Davidsastresas:
Hello @DonLakeFlyer. I took a quick look and all the changes seem fine to me. 

I see you also did some refactor in the qml indicator, etc besides bringing that to the new UI style, very nice. Also, I like how you added that "available" Q_Property to show the indicator or not.

Regarding this available property, versus that setting you removed to show or not Remote ID menus, I think as it is now it should be equivalent. As we have Remote Id menu always visible now it is fine because we can access the menu to fill the data even if no vehicle is still connected, so from a user experience point of view I think it is consistent and makes sense.

The only drawback of removing that "enable setting" is that we can no longer avoid GCS to send remote id messages on purpose.  In other words, as long as a valid remote id sends a heartbeat and QGC receives it, as it is now QGC will engage on Remote id communication and will start sending messages right away.

And this is not wrong, it is technically correct if we wanted to be as compliant as possible with this remote id thing, but I would personally leave somehow the possibility of a user manually enforcing not to send remote id messages from GCS, even if a valid remote ID device is detected. I think some users would appreciate if this option is available for them. What do you think?

DonLakeFlyer:
> The only drawback of removing that "enable setting" is that we can no longer avoid GCS to send remote id messages on purpose. In other words, as long as a valid remote id sends a heartbeat and QGC receives it, as it is now QGC will engage on Remote id communication and will start sending messages right away.

Ah, interesting. That makes sense. I'll add something back to support that.

PR #11227, opened 2024-03-10
GPS: Move RTK handling from App to GPSManager

HTRamsey:
Seemed kinda out of place in QGCApplication so I put it in GPSManager. I left all the signals the way they were in case we make a GPSRTK class or something instead. Also, move RTCM out of extra folder because it seemed unnecessary to have an extra folder for two small files.


src/GPS/GPSManager.cc
```diff
@@ -24,6 +24,49 @@ GPSManager::GPSManager(QGCApplication* app, QGCToolbox* toolbox)
 GPSManager::~GPSManager()
 {
     disconnectGPS();
+    if(_gpsRtkFactGroup)
+    {
+        delete _gpsRtkFactGroup;
+    }
```
julianoes, line 30: `delete` on `nullptr` is already a no-op:

```suggestion
    delete _gpsRtkFactGroup;
```

julianoes (approved):
Small nitpick, otherwise good to go.

bkueng (approved):
I missed that one, looks good

PR #11229, opened 2024-03-10
[Draft] On Emergency Button hit, send MAV_CMD_ODID_SET_EMERGENCY

jnomikos:
On Emergency button hit, sends the new MAV_CMD_ODID_SET_EMERGENCY command which was pushed to upstream MAVLink here: https://github.com/mavlink/mavlink/pull/2086

Discussions about it are here: https://github.com/mavlink/mavlink/wiki/20240228-Dev-Meeting 
and here: https://github.com/mavlink/mavlink/pull/2027#top
-----------

Sends the new MAVLink command MAV_CMD_ODID_SET_EMERGENCY when the emergency button is activated. 

**NOTE:** This is currently a draft because the MAVLink library on QGC needs to be updated to allow for this message to be sent. 

Test Steps
-----------
1. Ensure successfully builds and runs
2. Flash PX4 firmware which has support of this message (I will also create a draft PR for that, but MAVLink also needs to be updated on PX4)
3. Plug in remote id module with flight controller.
4. Ensure remote ID module is detected on QGC
5. Install a drone scanning app to view the status of the drone
6. Hold the emergency button and see if the status goes to emergency

Checklist:
----------

- [ ] [Review Contribution Guidelines](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CONTRIBUTING.md).
- [ ] [Review Code of Conduct](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CODE_OF_CONDUCT.md).

Related Issue
-----------

### DIFF
--- a/src/Vehicle/RemoteIDManager.cc
+++ b/src/Vehicle/RemoteIDManager.cc
@@ -485,6 +485,20 @@ void RemoteIDManager::setEmergency(bool declare)
 {
     _emergencyDeclared = declare;
     emit emergencyDeclaredChanged();
+
+    // We send a MAV Command to the vehicle to start or stop the emergency
+    _vehicle->sendMavCommand(
+        _vehicle->compId(),        // Target component
+        static_cast<MAV_CMD>(12900),  // MAV_CMD_ODID_SET_EMERGENCY - Called from number because from development.xml
+        true,                     // ShowError
+        _emergencyDeclared ? 1 : 0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0);
+
     // Wether we are starting an emergency or cancelling it, we need to enforce sending
     // this message. Otherwise, if non optimal connection quality, vehicle RID device
     // could remain in the wrong state. It is clarified to the user in remoteidsettings.qml


DonLakeFlyer:
How come this is still in development.xml and hasn't moved to common if it has been approved on the MAVLink side of things.

hamishwillee:
MAVLink only move things across when there is a prototype/proof of implementation and acceptance into flight stacks. Otherwise historically we've ended up with lots of crud in common that never got implemented.

DonLakeFlyer:
Ok, got it

DonLakeFlyer:
So I lost track of this. I see this command seems to still only be in development many (many) months later. The way QGC deals with this is that things in development.xml only go into debug builds. Does it still make sense to put this through?

jnomikos:
> So I lost track of this. I see this command seems to still only be in development many (many) months later. The way QGC deals with this is that things in development.xml only go into debug builds. Does it still make sense to put this through?

I think we should keep this open unless PX4 decides to merge my PR for emergency button [here](https://github.com/PX4/PX4-Autopilot/pull/21959)

DonLakeFlyer:
Seems like this has gone majorly stale. Can I close?

jnomikos:
Yeah I would close this. If anything comes of it in the future, I can reopen with updated code

> Seems like this has gone majorly stale. Can I close?

PR #11231, opened 2024-03-10
CMake: Improve Libraries

HTRamsey:
This makes some improvements to the cmake libraries and how everything is linked. I had tried to add the QML modules but was getting circular linking issues because of the way everything is currently so this is a step in that direction.

- All header files are added so they show up in QtCreator
- Qml file entries are sorted to help identify missing ones
- Linking is reduced in some libraries to limit circular dependencies
- find_package is added to each library, this helps make the libraries independent for things like unit tests

Edit: This should probably be rebased after #11237


DonLakeFlyer:
One thing I wonder about is that in old qmake you only needed to add the qml files in one place the resource file. Now with cmake it seems like you need to add them in two places: both the resource files and the cmake build files. It also seems like you can forget to add qml files to the cmake files and only have them in the resource file and the build will still work. But does that means that they won't be compiled from qml to code?

HTRamsey:
@DonLakeFlyer There won't be any resource files either. The 'new' way takes care of auto-generating all of the boilerplate stuff you had to do manually previously between the qrc files and registering files. Here's a random example from one of my projects, and it's all you need to do to get everything added and registered and such.
```
file(GLOB_RECURSE QML_SOURCES RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/*.qml)
file(GLOB QML_RESOURCES RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/res/*/*.*)
qt_add_qml_module(Comms
    URI comms
    VERSION 1.0
    QML_FILES ${QML_SOURCES}
    RESOURCES "${QML_RESOURCES}"
    OUTPUT_TARGETS Comms_targets
    IMPORT_PATH ${QT_QML_OUTPUT_DIRECTORY}
    IMPORTS
        QtCore
        QtQuick
        QtQuick.Controls
        QtQuick.Layouts
        QtQml.Models
        video
        utils
)
```

HTRamsey:
It then generates a qmltypes files, a qmldir file that looks some thing like 
```
module comms
linktarget Commsplugin
optional plugin Commsplugin
classname commsPlugin
typeinfo Comms.qmltypes
import QtCore
import QtQuick
import QtQuick.Controls
import QtQuick.Layouts
import QtQml.Models
import video
import utils
prefer :/qt/qml/comms/
ActionsConfirm 1.0 ActionsConfirm.qml
ActionsController 1.0 ActionsController.qml
ActionsSlider 1.0 ActionsSlider.qml
FollowConfig 1.0 FollowConfig.qml
```

and a qrc file that links to the qmldir that looks like
```
<RCC>
  <qresource prefix="/">
    <file alias="/qt/qml/comms">/home/user/Desktop/dev/Projects/project/build/Desktop_Qt_6_6_2_GCC_64bit-Release/qml/comms</file>

  </qresource>
</RCC>
```
All the listed resources are included automatically as well.

HTRamsey:
Also I believe all of this really helps with linting and compiling the qml for better performance.

DonLakeFlyer:
Ok, great. Should the qml files also move to directories which match there Qml import paths? I think that makes Qt Creator happier as well to find them.

HTRamsey:
No you don't have to do that either. It's all handled during the cmake configuration. So your qml files can be as scattered as you want, but it makes it a little more frustrating to work with if they aren't all together.

HTRamsey:
You can use this
```
set_source_files_properties(ui/res/qtquickcontrols2.conf PROPERTIES QT_RESOURCE_ALIAS qtquickcontrols2.conf)
```
to do the same thing as the file alias in qrc files. And when doing a Qml Singleton you just do this:
```
set_source_files_properties(
    ${CMAKE_CURRENT_SOURCE_DIR}/ui/qml/CameraSettings.qml
    PROPERTIES
        QT_QML_SINGLETON_TYPE TRUE
)
```

PR #11232, opened 2024-03-11
Shapelib: Fetch through Cmake

HTRamsey:
This collects shapelib through fetchcontent in cmake, so that the submodule can be removed when qmake goes. They have done some cleanup to their cmake implementation since 1.5.0 but most of it seems to be coming after 1.6.0.

### DIFF
--- a/.github/workflows/android_debug.yml
+++ b/.github/workflows/android_debug.yml
@@ -136,7 +136,6 @@ jobs:
               -DCMAKE_BUILD_TYPE=Debug \
               -DANDROID_ABI=${{ matrix.eabi }} \
               -DANDROID_PLATFORM=android-23 \
-              -DBUILD_TESTING:BOOL=OFF \
               -DQT_HOST_PATH:PATH=${QT_ROOT_DIR}/../gcc_64 \
               -DQT_DEBUG_FIND_PACKAGE=ON
             cmake --build ${{ runner.temp }}/shadow_build_dir/ --target all
--- a/libs/CMakeLists.txt
+++ b/libs/CMakeLists.txt
@@ -6,7 +6,6 @@ add_subdirectory(zlib)
 add_subdirectory(qmlglsink)
 
 set(BUILD_SHARED_LIBS OFF CACHE BOOL "Force qmdnsengine & shapelib to build as static" FORCE)
-add_subdirectory(shapelib)
 add_subdirectory(qmdnsengine)
 
 qt_add_library(xz STATIC
--- a/src/Utilities/CMakeLists.txt
+++ b/src/Utilities/CMakeLists.txt
@@ -29,6 +29,18 @@ if(MOBILE)
 	)
 endif()
 
+set(BUILD_SHAPELIB_CONTRIB OFF CACHE INTERNAL "")
+set(BUILD_SHARED_LIBS OFF CACHE INTERNAL "")
+set(BUILD_TESTING OFF CACHE INTERNAL "")
+
+include(FetchContent)
+FetchContent_Declare(shapelib
+	GIT_REPOSITORY https://github.com/OSGeo/shapelib.git
+	GIT_TAG v1.6.0
+    GIT_SHALLOW TRUE
+)
+FetchContent_MakeAvailable(shapelib)
+
 target_link_libraries(Utilities
 	PRIVATE
 		Qt6::Qml


DonLakeFlyer:
What happens if you clean your build and then do another make build without an internet connection? Will the build fail because the clean tossed the previously fetched shapelib? I don't want to tie builds to needing an internet connection.

HTRamsey:
When you fetch the content you can specify where the dependencies are downloaded and built, either individually or with global defaults. So then you can separate all submodule files from the source without needing something like a libs folder, and keep all the submodule build files from cluttering up the build directory. Also the downloaded dependencies aren't deleted when you run the build clean command anyways. The behavior of all that stuff is also easily configurable either in cmake itself or in something like QtCreator as well.

HTRamsey:
Since we aren't removing the qmake yet, we could try this and see what you think before we get to where we would remove the submodules and qmake anyways.

DonLakeFlyer:
It seems interesting, but there still is something that rubs me the wrong way about this. 

For example I've used this pattern a number of times in my other projects:
* Need to fix/changed something in a submodule where it needs to be coordinate between main and submodule repos
* Make the changes directly in the submodule which is part of the parent repo
* Make the changes in the main repo
* Test, fart around till I'm happy with changes on both sides
* Create commit to update submodule right from the submodule in main repo
* Wait for submodule to get merged in and then update main repo side of things

I also tend to not trust "clean" and to clean I usually just delete the entire build directory. Which would in turn cause me to need to download again.

HTRamsey:
You could set them to download outside the build directory as well but point taken. What about doing something like what I just updated it to? It would be nice to offer both options because most people have moved away from git submodules as much as possible.

DonLakeFlyer:
> I've used this pattern a number of times in my other projects

Seems to be answered here: https://cmake.org/pipermail/cmake/2019-June/069709.html

DonLakeFlyer:
My main concern is that I do tons of work in the middle of nowhere with no internet. So I don't want to screw myself over by mistake. I'm old, maybe I just need to learn :) Willing to give this a try.

HTRamsey:
Whatever you prefer. I think having it be like this:
```
if(QGC_FETCH_DEPENDS)
	include(FetchContent)
	FetchContent_Declare(shapelib
		GIT_REPOSITORY https://github.com/OSGeo/shapelib.git
		GIT_TAG v1.6.0
	        GIT_SHALLOW TRUE
	)
	FetchContent_MakeAvailable(shapelib)
else()
	add_subdirectory(${CMAKE_SOURCE_DIR}/libs/shapelib shapelib.build)
endif()
```
and defaulting to using git submodules could work fine. Maybe use FETCHCONTENT_FULLY_DISCONNECTED instead of QGC_FETCH_DEPENDS or something I don't know. We can adjust later if needed.

julianoes:
In MAVSDK we do it somewhat similar. We have the notion of a `SUPERBUILD`, with `SUPERBUILD` set it will download everything using `ExternalProject_add` and then build it, locally install it (using `CMAKE_INSTALL_PREFIX`) and pass the install prefix to the main build using `CMAKE_PREFIX_PATH`.

See:
https://github.com/mavlink/MAVSDK/blob/main/third_party/CMakeLists.txt

And tinyxml2 as an example:
https://github.com/mavlink/MAVSDK/blob/main/third_party/tinyxml2/CMakeLists.txt#L29

This allows you to build it without SUPERBUILD when offline, and use system-wide installed dependencies, or supply the dependencies manually using your own prefix paths.

SUPERBUILD works ok when offline, as long as you don't touch any CMakeLists.txt file. Once you do, it will trigger the configure step which will try to download the third party repos again.

It's not optimal either but an alternative way.

@JonasVautherin did a nice write writeup here: https://www.acarg.ch/posts/cmake-deps/

DonLakeFlyer:
I don't think we need to make this optional. I'd much rather just switch to the new things if that is the better way to do things now. Would it be possible to fetch the dependencies into something like a fetch-dependencies folder instead of directly into build? Then you can do a source clean by delete the build dir, and also force a refetch of dependencies by deleting the fetch-dependencies. Or would that sort of things be cmake strangeness?

JonasVautherin:
I personally [don't do superbuilds anymore](https://www.acarg.ch/posts/cmake-superbuild/), but [this](https://www.acarg.ch/posts/cmake-deps/) is still relevant. The idea is:

* Treat your project like if all the dependencies were installed on the system (i.e. by having CMake find them with `find_package`).
* Either install the dependencies on your system (e.g. using the distro package manager), or build and install them somewhere (with `CMAKE_INSTALL_PREFIX`).

This way, anyone is free to choose how to deal with the dependencies. For instance you can build and install the dependencies in `/home/user/my-deps` and have your project somewhere else (say `/home/user/project`). By setting `CMAKE_PREFIX_PATH=/home/user/my-deps`, you tell CMake to look for them there first.

It means that you can clean your project completely independently from the dependencies: your project really does not know whether those are system dependencies or local ones that were "injected" with `CMAKE_PREFIX_PATH`, so you are free to do whatever you want.

PR #11233, opened 2024-03-11
Android: Update Source Package

HTRamsey:
This does a couple things (Tested on Android 14 OnePlus, it is the non-functionality changes from #11082 which I am changing to just focus on the Android Serial Implementation):

- ~~Creates a multi-abi apk for armv7 and arm64-v8a instead of building individually~~
- Updates Gradle and AGP versions
- Switches CI to use Java 17 which is required to use a AGP version that supports sdk 34+.
- Sets target sdk to 34.
- ~~Fixes ANDROID_VERSION_NAME in qmake build~~
- Fixes package name, this should be set in gradle using namespace instead of AndroidManifest.xml package for now on (There was a deprecation warning).
- ~~Moves android initialization code into its own file in AndroidInit.cpp, cleaning up the main file of platform dependent code a bit.~~
- Removes two obsolete android tools, as these values can be set by user variables in the build instead of manually changing afterwards. These variables can be set either through build system variables or variables set in gradle.properties or gradle.settings, so future custom builds should make use of that functionality instead.
- Includes an android USB fix that is required from android 14 onwards
- Adds back the gradle files, which allow various extra customization settings rather than using default Qt ones.

Note: We might be able to leave out the gradlew stuff if we want

Edit: Removed deployment changes for a separate PR, this now just focuses on updating the android package content


android/AndroidManifest.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="org.qtproject.example"
+    package="org.mavlink.qgroundcontrol"
```
julianoes, line 3: :smile:

julianoes (approved):
Very nice!

I assume the version str still works, even after you removed that update_android_version.sh script?

HTRamsey:
I still need to add the logic to set the CMake variable for version string based on daily/release but it will do the exact same thing the tools did.

PR #11235, opened 2024-03-11
Zlib: Fetch through Cmake

HTRamsey:
Fetch zlib through cmake and build. This ensures all platforms are on the same version. When qmake goes, the windows zlib can be removed from the libs folder.


HTRamsey:
That linux release failure is unrelated to this

PR #11237, opened 2024-03-12
Src: Remove Unused Includes

HTRamsey:
This will help limit what modules need to be linked together in cmake. There is no functionality change, and I didn't really do much forward declarations either.

Seems to cut like 5minutes off of the CI build time too.


julianoes (approved):
Wow! Amazing, did you do this automated or manually?

HTRamsey:
About half manual, half automated. Probably a couple hours of manual work unfortunately but seemed worth it

PR #11240, opened 2024-03-12
SDL2: Fetch Through CMake

HTRamsey:
Fetches SDL2 through cmake and just builds the necessary components for joystick. This will make all platforms run the same version. When qmake goes, the libs/sdl2/msvc and Frameworks/SDL2.framework stuff can be deleted.

This means ios and android could use sdl2 as well


cmake/BuildSdl2.cmake
```diff
@@ -0,0 +1,50 @@
+# find_package(SDL2 QUIET)
+if(SDL2_FOUND)
```
DonLakeFlyer, line 2: This seems odd. the find_package is commented out. Won't this always the false?

cmake/BuildSdl2.cmake
```diff
@@ -0,0 +1,50 @@
...
+    message(STATUS "Using SDL2 via find_package")
+endif()
+
+# 2. Try using a vendored SDL library
+if(NOT SDL2_FOUND AND EXISTS "${CMAKE_CURRENT_LIST_DIR}/SDL/CMakeLists.txt")
```
DonLakeFlyer, line 7: Why is this needed?

HTRamsey: It's not, this was while I was waiting to figure out how we were going to handle submodules. Same with the other issue.

PR #11250, opened 2024-03-13
Pinch Zoom Fix

ben-draper-ascent:
The MouseArea fix was dominating the PinchHandler. Figured we should stick to the QML intended implementation of the Map and just disable the views if the tool drawer is open

### DIFF
--- a/src/FlightMap/FlightMap.qml
+++ b/src/FlightMap/FlightMap.qml
@@ -147,42 +147,18 @@ Map {
         property:           "zoomLevel"
     }
 
-    MouseArea {
-        anchors.fill:       parent
-        acceptedButtons:    Qt.LeftButton
+    DragHandler {
+        target: null
 
-        property real startMouseX: 0
-        property real startMouseY: 0
-        property real lastMouseX: 0
-        property real lastMouseY: 0
-
-        onClicked: (mouse) => {
-            var deltaX = Math.abs(startMouseX - lastMouseX)
-            var deltaY = Math.abs(startMouseY - lastMouseY)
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
I had tried that and it didn't work. Even with the fly view invisible, the DragHandler was bleeding through and causing the map to move around. The only way I could get it to work in all cases was to not use it.

ben-draper-ascent:
> I had tried that and it didn't work. Even with the fly view invisible, the DragHandler was bleeding through and causing the map to move around. The only way I could get it to work in all cases was to not use it.

Yes visibility is not enough which is why I changed the Enable property. I tested on my android and did a control test where it did not work and then loaded the new apk to prove it resolved.

DonLakeFlyer:
Ah, I missed that. That's the magic thing! What a pain.

DonLakeFlyer:
What about the Plan View drags bleeding down to the Fly View?

DonLakeFlyer:
Also can you add a comment to the places you enable based on visibility explaining why? That way, years down the road someone doesn't remove them thinking they aren'y really needed.

HTRamsey:
@DonLakeFlyer I could have sworn I remember you adding a StackView to the MainRootWindow to swap between the views, which would have prevented this too I would think

ben-draper-ascent:
> What about the Plan View drags bleeding down to the Fly View?

I tested this as well and its why I added the enabled property in the PlanView item as well. I'll add those comments now

s-lisovenko:
https://github.com/mavlink/qgroundcontrol/pull/11257

PR #11258, opened 2024-03-15
Qmlglsink: Gstreamer Build Fixes

HTRamsey:
This gets gstreamer to link correctly via cmake, and removes the unnecessary gst-plugins-good submodule since we only need the few files of the one plugin. Video is working fine on a Oneplus 9 pro streaming a 8mp 30fps rtsp camera.

### DIFF
--- a/.github/workflows/macos_debug.yml
+++ b/.github/workflows/macos_debug.yml
@@ -62,17 +62,12 @@ jobs:
           cache: true
 
       - name: Install Dependencies
-        run: |
-          brew install ninja
-          brew install SDL2
+        run: brew install ninja SDL2
 
-      - name: Install Gstreamer
-        run: |
-            wget https://gstreamer.freedesktop.org/data/pkg/osx/1.18.6/gstreamer-1.0-devel-1.18.6-x86_64.pkg
-            wget https://gstreamer.freedesktop.org/data/pkg/osx/1.18.6/gstreamer-1.0-1.18.6-x86_64.pkg
-            for package in *.pkg ;
-              do sudo installer -verbose -pkg "$package" -target /
-            done
+      - name: Setup GStreamer
+        uses: blinemedical/setup-gstreamer@v1
+        with:
+          version: 1.18.6
 
       - name: Create build directory
         run:  mkdir ${{ runner.temp }}/shadow_build_dir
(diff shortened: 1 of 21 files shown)


HTRamsey:
@zdanek this look good to you?

julianoes:
Hm, so you're just copying in some gstreamer files? I'm not sure I like that in terms of maintainability. Or did I misunderstand this?

HTRamsey:
> Hm, so you're just copying in some gstreamer files? I'm not sure I like that in terms of maintainability. Or did I misunderstand this?

Most of this is fixing the cmake part that links to gstreamer. The part where I copied gstreamer files in is because QGC has used custom modified gstreamer files for a long time now and the only reason this is still needed is because Linux distros like Ubuntu 22.04 still use an older gstreamer that doesn't support Qt6. The submodule we were using [here](https://github.com/mavlink/gst-plugins-good/tree/5ff0a6af0e1fd29bafc9f6644e728d24349d0ce7) was forked from [here](https://github.com/GStreamer/gst-plugins-good) which is now defunct. Only a single plugin (qt) was ever used from the entire gst-plugins-good, so I removed 99% of the gst-plugins-good that was unused in the submodule, and copied over the only code that's actually important. Either way we are still using custom gstreamer files rather than pulling directly from gstreamer.

julianoes:
> because Linux distros like Ubuntu 22.04 still use an older gstreamer that doesn't support Qt6

I see, what a pain.

Fine with me then.

DonLakeFlyer:
Regarding streamer on Linux. Should we move away from using the OS provided version of streamer on linux and instead use the same version used everywhere else and include it in the appimage?

HTRamsey:
That's a decent option, but I have no idea how significantly it will affect the build time or appimage size

HTRamsey:
@DonLakeFlyer Can you test video on mac?

DonLakeFlyer:
Yes, I can do basic video testing on Mac using PX4 SITL.

HTRamsey:
[of course](https://bugreports.qt.io/browse/QTBUG-114635)

DonLakeFlyer:
I don't really know what Wayland is. But could we drop support for it for now until they fix this somehow?

PR #11259, opened 2024-03-16
Remove Qt5 Compatibility Module

### LINKED ISSUE #11081: Remove usage of Qt5Compat
I think this just comes down to:

- Replacing ColorOverlay & Image in APMSafetyComponent.qml with a QGCColoredImage
- Redesigning QGCColoredImage.qml to use [multieffect](https://doc.qt.io/qt-6/qml-qtquick-effects-multieffect.html) instead of ColorOverlay, can use [colorization & colorizationColor variables](https://doc.qt.io/qt-6/qml-qtquick-effects-multieffect.html#colorization-prop)
- Removing OpacityMask in QGCAttitudeWidget.qml, can use [maskSource](https://doc.qt.io/qt-6/qml-qtquick-effects-multieffect.html#maskSource-prop)
- Removing DropShadow in VehicleMapItem.qml, can use [shadowEnabled](https://doc.qt.io/qt-6/qml-qtquick-effects-multieffect.html#shadowEnabled-prop)
- Removing the module imports from qml, linking in qmake/cmake, and downloads in ci

HTRamsey:
Resolves #11081. Removes all usage of the Qt5 Compatibility Module which I believe was made due to licensing issues with libraries Qt was using for the GraphicalEffects.
This also fixed a couple of issues:

- APMParameterMetaData::getParameterMetaDataVersionInfo was not getting the version numbers correctly
- APMFirmwarePlugin::_handleIncomingStatusText was never setting _coaxialMotors

All Qml visuals are 100% identical except for the ADSB shadow because I am terrible at graphical design. Here is a before and after:
DropShadow_Before
DropShadow_After

@DonLakeFlyer Hopefully you can just tune a couple of the settings [here](https://github.com/HTRamsey/qgroundcontrol/blob/1dcc5f25c2581e83d0f1882e70f3beb76ae35824/src/FlightMap/MapItems/VehicleMapItem.qml#L47) to make it look better. I try to avoid touching Qml files since you're much better at it but I needed to in this circumstance.

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -29,11 +29,10 @@
 #include "QGCLoggingCategory.h"
 
 #include <QTcpSocket>
-#include <QtCore5Compat/QRegExp>
+#include <QtCore/QRegularExpression>
+#include <QtCore/QRegularExpressionMatch>
 
 QGC_LOGGING_CATEGORY(APMFirmwarePluginLog, "APMFirmwarePluginLog")
-
-static const QRegularExpression APM_FRAME_REXP("^Frame: ");
 
 const char* APMFirmwarePlugin::_artooIP =                   "10.1.1.1"; ///< IP address of ARTOO controller
 const int   APMFirmwarePlugin::_artooVideoHandshakePort =   5502;       ///< Port for video handshake on ARTOO controller
@@ -261,21 +260,18 @@ bool APMFirmwarePlugin::_handleIncomingStatusText(Vehicle* /*vehicle*/, mavlink_
     // APM user facing calibration messages come through as high severity, we need to parse them out
     // and lower the severity on them so that they don't pop in the users face.
 
-    QString messageText = _getMessageText(message);
+    const QString messageText = _getMessageText(message);
     if (messageText.contains("Place vehicle") || messageText.contains("Calibration successful")) {
         _adjustCalibrationMessageSeverity(message);
         return true;
     }
 
-    if (messageText.contains(APM_FRAME_REXP)) {
-        // We need to parse the Frame: message in order to determine whether the motors are coaxial or not
-        QRegExp frameTypeRegex("^Frame: (\\S*)");
-        if (frameTypeRegex.indexIn(messageText) != -1) {
-            QString frameType = frameTypeRegex.cap(1);
-            if (!frameType.isEmpty() && (frameType == QStringLiteral("Y6") || frameType == QStringLiteral("OCTA_QUAD") || frameType == QStringLiteral("COAX"))) {
-                _coaxialMotors = true;
-            }
-        }
+    static const QRegularExpression APM_FRAME_REXP("^Frame: (\\S*)");
+    const QRegularExpressionMatch match = APM_FRAME_REXP.match(messageText);
+    if (match.hasMatch()) {
+        static const QSet<QString> coaxialFrames = {"Y6", "OCTA_QUAD", "COAX", "QUAD/PLUS"};
+        const QString frameType = match.captured(1);
+        _coaxialMotors = coaxialFrames.contains(frameType);
     }
 
     return true;
@@ -367,7 +363,7 @@ QString APMFirmwarePlugin::_getMessageText(mavlink_message_t* message) const
 
     // Ensure NUL-termination
     b[b.length()-1] = '\0';
-    return QString(b);
+    return QString::fromLocal8Bit(b, std::strlen(b.constData()));
 }
 
 void APMFirmwarePlugin::_setInfoSeverity(mavlink_message_t* message) const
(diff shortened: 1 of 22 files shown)


src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
```diff
@@ -360,14 +357,10 @@ void APMFirmwarePlugin::adjustOutgoingMavlinkMessageThreadSafe(Vehicle* vehicle,
...
-
-    b.resize(MAVLINK_MSG_STATUSTEXT_FIELD_TEXT_LEN+1);
-    mavlink_msg_statustext_get_text(message, b.data());
-
-    // Ensure NUL-termination
-    b[b.length()-1] = '\0';
-    return QString(b);
+    mavlink_statustext_t data;
+    mavlink_msg_statustext_decode(message, &data);
+    QString text = QString::fromStdString(data.text);
```
julianoes, line 362: I disagree here. The previous implementation was there for a reason. `data.text` is not necessarily 0 terminated by spec, so this can read garbage memory.

What about using [QString::fromLocal8Bit()](https://doc.qt.io/qt-6/qstring.html#fromLocal8Bit) which allows to pass the max length as well?

julianoes (changes requested):

src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
HTRamsey: I'm sure I had a reason, but I don't remember why I even touched this part in this PR. I'll have to go back and check.

HTRamsey: So maybe this and the other fix should be separate PRs, but the reason I did this was because when changing the QRegExp in _handleIncomingStatusText, I noticed it was impossible for _coaxialMotors to be set as true because of the trailing 0's, as shown in this image. Maybe it's better to just check if the QStringLiterals are substrings of frameType.
image

julianoes: I'm confused. A statustext is a string closed with `\0` or not if the full length is used. If you construct QString from it, it will read past the array into garbage and we need to protect against that. I'm not sure how that matters with a regexp? Also, this looks like statustext is potentially somewhat abused for this use case :thinking:?

HTRamsey: So it's not really related to QRegExp, which is why I probably should have made a separate issue/PR, but I tried to fix it when I was changing this code anyways. The problem the way things are is that you cannot do an equality comparison of `frameType` (which is always 50 bytes) directly to "Y6", "OCTA_QUAD", etc. That equality will never be true. So something needs to be changed with the `frameType` comparison, either by dropping the trailing zeros or using substring or something like that.

julianoes: Right, what about if you convert both to QString and then compare? That would drop trailing 0s, right?

HTRamsey: They already are both QStrings. The thing is in `_getMessageText`, the `\0` termination is always added at the very end. So you will always have a bunch of trailing zeros until the 50th byte.

julianoes: So, I guess we should copy into length 51 char array, then do a `strlen` and then use [QString::fromLocal8Bit()](https://doc.qt.io/qt-6/qstring.html#fromLocal8Bit) with the length we have?

HTRamsey: That seemed to work.

src/QtLocationPlugin/GoogleMapProvider.cpp
```diff
@@ -78,19 +79,27 @@ void GoogleMapProvider::_googleVersionCompleted() {
     }
 #endif
 
-    QRegExp reg(QStringLiteral("\"*https?://mt\\D?\\d..*/vt\\?lyrs=m@(\\d*)"), Qt::CaseInsensitive);
-    if (reg.indexIn(html) != -1) {
-        _versionGoogleMap = QString(QStringLiteral("m@%1")).arg(reg.capturedTexts().value(1, QString()));
+    static const QRegularExpression regMap( "\"*https?://mt\\D?\\d..*/vt\\?lyrs=m@(\\d*)", QRegularExpression::CaseInsensitiveOption );
+    const QRegularExpressionMatch matchMap = regMap.match( html );
+    if( matchMap.hasMatch() )
+    {
```
DonLakeFlyer, line 85: FYI: The QGC coding style is to not use an extra line for all of these these. They should be on the line above.

HTRamsey: Sorry I switch between working on QGC and FreeRTOS a lot, and that was FreeRTOS coding style

PR #11263, opened 2024-03-18
Set Message Rates From Mavlink Inspector

HTRamsey:
Uses the currently selected message in the MavlinkInspector of the Analyze View to adjust individual message rates (rather than stream rates for groups of messages like is currently used for ArduPilot).
@DonLakeFlyer Design Advice? Maybe a TextField or SpinBox instead? Also, if you're cool with it I would like to move the QGCMAVLinkMessage and QGCMAVLinkSystem out of MavlinkInspectorController and into comm to be used by MAVLinkProtocol.
MavInspectorMsgRate
MavInspectRates

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -7,6 +7,7 @@ Note: This file only contains high level features or important fixes.
 * New combined compass and attitude instrument
 * You can select between multiple instruments by clicking on the instrument conntrol on a desktop build or press and hold on mobile builds
 * Support for Fly View and Joystick custom mavlink actions has changed. Both the name and formation of the command file is different now. Go to QGC docs to understand how it works now.
+* Support for setting individual Mavlink message rates in the Mavlink Inspector.
 
 ## 4.1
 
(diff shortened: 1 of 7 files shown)


julianoes (approved):
Oh, this is very cool!

DonLakeFlyer:
I think this is fine as is.

> Also, if you're cool with it I would like to move the QGCMAVLinkMessage and QGCMAVLinkSystem out of MavlinkInspectorController and into comm to be used by MAVLinkProtocol.

Something which I have always wanted to do but have never gotten around to which relates to this is to:
* connect the mavlink inspector message keeping stuff to Vehicle
* Expose it through Qml properties on Vehicle, like the FactGroup stuff
* Allow Qml access to the full mavlink message stream like the FactGroup stuff does now

This allows you to create all sorts of telemetry based UI much more easily. And then you don't need to keep adding FactGroups for this or that.

DonLakeFlyer:
@HTRamsey Can you unmark as Draft and I'll merge

DonLakeFlyer:
Could you also update the changelog for this new feature?

PR #11265, opened 2024-03-20
Fix bug with junk characters in 16-character parameter name

ddatsko:
Fix bug with junk characters in 16-character parameter name

Description
-----------
I noticed that when parameters are 16-characters long, I can often see some junk characters in the end of their names when listing all the parameters. Looks like the problem was in the fact that we create a 16-characters buffer and copy the parameter name there. If the name is shorter than 16 characters, everything is fine, as the name is a null-terminated string. However, when the parameter name is exactly 16 characters long, there is no proper null-termination (which is allowed by [MavLink](https://mavlink.io/en/services/parameter.html#parameter-names)), and some junk characters get into the name. Fortunately, this seems to not break anything else, but potentially could lead to the application crash. An example of such behaviour with ArduPilot FCU is depicted below 
Selection_028

This PR fixed the problem by increasing the buffer to 17 characters instead of 16. This ensures that when the parameter name is 16-characters long, the string stays properly null-terminated

Checklist:
----------

### DIFF
--- a/src/FactSystem/ParameterManager.cc
+++ b/src/FactSystem/ParameterManager.cc
@@ -173,7 +173,7 @@ void ParameterManager::mavlinkMessageReceived(mavlink_message_t message)
         mavlink_msg_param_value_decode(&message, &param_value);
 
         // This will null terminate the name string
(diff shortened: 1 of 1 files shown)


julianoes (approved):
That looks correct, thanks!

PR #11266, opened 2024-03-21
JoystickAndroid: Implement acquiring hat button values similar to existing implementation in JoystickSDL

ivanzivotic:
Implement _getHat function in the JoystickAndroid implementation of base Joystick class

Description
-----------
This pull request implements the _getHat(int hat, int i) function of JoystickAndroid class (inherited from Joystick class). Although this function is not called directly from anywhere in the code, my main motivation for this was to implement the same behavior on all supported platforms. It was driven by an existing implementation in the JoystickSDL class, where the hat variable represents a choice of which hat device is being used and i variable signifies one of four buttons available on hat.  
Since there is support for only one hat button on Android operating system (explained [here](https://developer.android.com/develop/ui/views/touch-and-input/game-controllers/controller-input#button)), only value 0 of hat parameter is accepted in the function. The second parameter is interpreted by following the implementation order of the buttons in the JoystickSDL class (up, down, left, right). All other values are considered invalid.  
Since the hat values are represented by Android OS as axes, we have to read the values of X and Y Hat axis and check if they are -1 or 1, as described on [this link](https://developer.android.com/reference/android/view/MotionEvent#AXIS_HAT_X). To achieve that, AXIS_HAT_X and AXIS_HAT_Y values are added as static integers to JoystickAndroid class and are read as part of static init function of the same class (similar to ACTION_DOWN and ACTION_UP android constants, except that they are part of the KeyEvent and not the MotionEvent android class).

Test Steps
-----------
Try calling the _getHat(int, int) function from a custom app on Android devices. Before this change, the behavior was not the same for devices running SDL library comparing to Android.

Checklist:
----------

Related Issue
-----------
None.


julianoes:
@Davidsastresas you might be best at reviewing this.

HTRamsey:
Working through Joystick code now and came across this

PR #11282, opened 2024-04-02
CI build fails:: using qmake for android_debug workflow

omid-esrafilian:
@HTRamsey 
[I apologize in advance if the pull request looks more like an issue report]
The CI build fails for android_debug workflow, an example can be found in #11214.
As a potential solution, I used "qmake" instead of "cmake" for android_debug workflow.
Please let me know if that makes sense.


HTRamsey:
Taken care of in  #11274

omid-esrafilian:
@HTRamsey 
Thanks much.

PR #11306, opened 2024-04-06
GPS: update submodule to main

julianoes:
This is no change, however, we should point to main, not a feature branch.


julianoes:
Looks like CI failures are unrelated. I'm merging this early as the no longer existing submodule branch is going to cause problems in other PRs otherwise.

PR #11308, opened 2024-04-07
Fix colorized icon with Qt5Compat.GraphicalEffects instead MultiEffet from Qt6

s-lisovenko:
At the moment, MultiEffect has limitations in its operation. It does not work with black images. Therefore, the most logical solution would be to use the features from Qt5

https://forum.qt.io/topic/144070/qt6-color-svg-using-multieffect/8

Multieffect:
Screenshot from 2024-04-07 14-36-26

ColorOverlay:
Screenshot from 2024-04-07 14-35-41

### DIFF
--- a/src/QmlControls/QGCColoredImage.qml
+++ b/src/QmlControls/QGCColoredImage.qml
@@ -1,6 +1,6 @@
 import QtQuick
 import QtQuick.Controls
-import QtQuick.Effects
+import Qt5Compat.GraphicalEffects
 
 import QGroundControl.Palette
 
@@ -35,10 +35,9 @@ Item {
         sourceSize.height:  height
     }
 
-    MultiEffect {
-        source: image
-        anchors.fill: image
-        colorizationColor: parent.color
-        colorization: 1.0
+    ColorOverlay {
+        anchors.fill:       image
+        source:             image
+        color:              parent.color
     }
 }


DonLakeFlyer:
@HTRamsey Did you have some other way of fixing this other than reverting back?

s-lisovenko:
> @HTRamsey Did you have some other way of fixing this other than reverting back?

No, at the moment this is the most reliable way. Moreover, it does not depend on the Qt5Compat library. I see no reason not to continue to use it.

DonLakeFlyer:
Ah, ok. I thought this was going back to the compat thing. If it works, it works! Thanks.

s-lisovenko:
@DonLakeFlyer Sorry, seems like it depends on Qt5Compat. it should be presented in OS. So we need to consider to include it in cmake and qmake

DonLakeFlyer:
Ok, I'm going to revert since it will fail the AppImage I think with a missing lib.

s-lisovenko:
> Ok, I'm going to revert since it will fail the AppImage I think with a missing lib.

Yes, please. For now there no alternative in QT6. Maybe in future releases.

s-lisovenko:
@DonLakeFlyer could we reopen this PR and add missing Qt5Compat dependency. I think we should continue using it as non alternative in qt6.

DonLakeFlyer:
Doesn't seem like there is a way to reopen a merged PR.

s-lisovenko:
Ok. I will create another with missing library

PR #11311, opened 2024-04-07
Exclude shapelib and zlib from cmake install process

s-lisovenko:
Modify CMakeLists for shapelib and zlib fetch process to exclude them from install procedure

### DIFF
--- a/src/Compression/CMakeLists.txt
+++ b/src/Compression/CMakeLists.txt
@@ -9,6 +9,7 @@ qt_add_library(compression STATIC
 
 set(ZLIB_BUILD_EXAMPLES OFF CACHE INTERNAL "")
 set(BUILD_SHARED_LIBS OFF CACHE INTERNAL "")
+set(SKIP_INSTALL_ALL ON CACHE INTERNAL "")
 
 include(FetchContent)
 FetchContent_Declare(zlib
--- a/src/Utilities/CMakeLists.txt
+++ b/src/Utilities/CMakeLists.txt
@@ -37,9 +37,13 @@ include(FetchContent)
 FetchContent_Declare(shapelib
 	GIT_REPOSITORY https://github.com/OSGeo/shapelib.git
 	GIT_TAG v1.6.0
-    GIT_SHALLOW TRUE
+	GIT_SHALLOW TRUE
 )
-FetchContent_MakeAvailable(shapelib)
+FetchContent_GetProperties(shapelib)
+if(NOT shapelib_POPULATED)
+	FetchContent_Populate(shapelib)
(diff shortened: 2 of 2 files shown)


HTRamsey:
I assume this removes all the extra shapelib executables? Also I believe zlib has the skip_install options we can use instead

s-lisovenko:
> skip_install

Zlib was fixed.
About shapelib, yes all targets will be remove from install procedure. Do we need something from shapelib inside install folder?

HTRamsey:
Not that I know of. I believe shapelib 1.7.0 will have better cmake options to control install but this should be fine for now

HTRamsey (approved):

s-lisovenko:
Better way is to use cmake 3.28 with ECLIDE_FROM_ALL parameters for FetchContent_Declare function. But I don’t know if this acceptable now.

s-lisovenko:
But maybe more better way i think is to use ExternalProject_add and build and install then use find_package. Like Mavsdk team do it.

HTRamsey:
We can always come back and do it more like that later. At least at the moment this will work and it will guarantee everyone is using the same versions dependencies

DonLakeFlyer:
Is this fix related to me hitting this:
```
don@Donalds-Mac-Pro build % cmake --install .                                                     
-- Install configuration: "Release"
-- Installing: /Users/don/repos/qgroundcontrol/build/staging/lib/libz.1.3.1.dylib
-- Installing: /Users/don/repos/qgroundcontrol/build/staging/lib/libz.1.dylib
-- Installing: /Users/don/repos/qgroundcontrol/build/staging/lib/libz.dylib
CMake Error at _deps/zlib-build/cmake_install.cmake:87 (file):
  file INSTALL cannot find
  "/Users/don/repos/qgroundcontrol/build/_deps/zlib-build/libz.a": No such
  file or directory.
Call Stack (most recent call first):
  src/Compression/cmake_install.cmake:42 (include)
  src/cmake_install.cmake:82 (include)
  cmake_install.cmake:47 (include)
```

HTRamsey:
I believe this PR removes the static lib so it should help with that, although I don't understand why you would get that in the first place

HTRamsey:
By the way, if you are using QtCreator you can go to Run -> Add Deploy Step -> CMake Install to automate that step

s-lisovenko:
> By the way, if you are using QtCreator you can go to Run -> Add Deploy Step -> CMake Install to automate that step

Yes, this PR should fix this error

s-lisovenko:
> I believe this PR removes the static lib so it should help with that, although I don't understand why you would get that in the first place

No, This PR jyst remove it from install process. The problem is that by default zlib use dynamic library
Screenshot from 2024-04-09 21-55-09

PR #11316, opened 2024-04-09
Convert Linux Release to CMake

HTRamsey:


HTRamsey:
We are getting
```
qt.core.qobject.connect: QObject::connect(QObject, QThreadPool): invalid nullptr parameter
qt.core.qobject.connect: QObject::connect(QObject, Unknown): invalid nullptr parameter
qt.core.qobject.connect: QObject::connect(QObject, Unknown): invalid nullptr parameter
```
at the `delete app;` line which is causing the failure since QT_FATAL_WARNINGS is enabled

PR #11319, opened 2024-04-09
Convert Android Release to CMake

HTRamsey:


HTRamsey:
@DonLakeFlyer I believe this is ready but it doesn't have access to the signing secrets so the signing part fails. Maybe we should have dummy secrets for PRs or something?

DonLakeFlyer:
Can we just not sign on pulls, only sign on master merges?

HTRamsey:
This is ready

DonLakeFlyer (approved):

PR #11273, opened 2024-03-26
Quick controls2 basic missing libraries added  

TOTON95:
# Quick controls2 basic missing libraries added  
Description
-----------

Some missing libraries for AppImage were added to `QGCPostLinkCommon.pri` to load the Quick Controls2 Basic resources

Test Steps
-----------

1. Download and enable QGC AppImage as usual into a Linux Distro
2. Run the AppImage
3. Expect to run without issues coming from missing libraries 

Checklist:
----------

- [x] [Review Contribution Guidelines](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CONTRIBUTING.md).
- [x] [Review Code of Conduct](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have tested my changes.

Related Issue
-----------

More details about the problem can be found  in #11272 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #11321, opened 2024-04-09
Revert "Quick controls2 basic missing libraries added  "

HTRamsey:
Reverts mavlink/qgroundcontrol#11273
Just curious if it passes without this, ultimately won't matter switching to cmake

### DIFF
--- a/QGCPostLinkCommon.pri
+++ b/QGCPostLinkCommon.pri
@@ -110,8 +110,6 @@ LinuxBuild {
         libQt6DBus.so.6 \
         libQt6OpenGLWidgets.so.6 \
         libQt6QuickControls2.so.6 \
-        libQt6QuickControls2Basic.so.6 \
-        libQt6QuickControls2BasicStyleImpl.so.6 \
         libQt6SerialPort.so.6 \
         libQt6Gui.so.6 \ 
         libQt6PositioningQuick.so.6 \


TOTON95:
@HTRamsey, weirdly, the outcome is not consistent. Hopefully, CMake will avoid this behavior in the future.

TOTON95:
@HTRamsey I just tested the output of this build:

```bash
Filter rules "*Log.debug=false\nGStreamerAPILog.debug=true\nqt.qml.connections=false"
System reported locale: QLocale(English, Latin, United States) ; Name "en_US" ; Preffered (used in maps):  "en-US"
qt.multimedia.ffmpeg.libsymbolsresolver: Couldn't load OpenSsl library
qt.multimedia.ffmpeg.libsymbolsresolver: Couldn't load VAAPI library
VideoReceiverLog: Stop called on empty URI
VideoReceiverLog: Stop called on empty URI
MAVLinkLogManagerLog: MAVLink logs directory: "/home/totonzx/Documents/QGroundControl Daily/Logs"
Map Cache in: "/home/totonzx/.cache/QGCMapCache300" / "qgcMapCache.db"
qrc:/qml/MainRootWindow.qml:11:1: module "QtQuick.Controls" version 6.6 cannot be imported because:
Cannot load library /tmp/.mount_QGrounrB2xM5/Qt/qml/QtQuick/Controls/Basic/libqtquickcontrols2basicstyleplugin.so: (libQt6QuickControls2Basic.so.6: cannot open shared object file: No such file or directory)

```

HTRamsey:
Well that makes no sense. We'll revisit this after cmake and see if it is still a problem

TOTON95:
@HTRamsey Sure thing, I will be around if you need help to test it.

PR #11315, opened 2024-04-09
Convert MacOS Release to CMake

HTRamsey:
I think it creates a .app file but we need to convert it to a .dmg or something? I'm not familiar with this, but I can look through the QGCPostLinkInstaller.pri and figure something out

PR #11323, opened 2024-04-09
OSX CI using cmake, DMG creation support

DonLakeFlyer:
This is a replacement for #11315.

Theoretically cmake CPack can be used to create the DMG. But I couldn't get that to work so instead of did it the manual hardcoded way. I'll come back to this at some point to try to switch to that. For now, this works and lets us move forward.


.github/workflows/macos.yml
```diff
@@ -0,0 +1,91 @@
...
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-tags: true
+
+      - name: ccache
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          create-symlink: true
+          key: ${{ runner.os }}-Debug
```
HTRamsey, line 41: I forgot to set this part

key: ${{ runner.os }}-${{ matrix.BuildType }}
restore-keys: ${{ runner.os }}-${{ matrix.BuildType }}

.github/workflows/macos.yml
```diff
@@ -0,0 +1,91 @@
...
+
+      - name: Install Qt
+        uses: jurplel/install-qt-action@v3
+        with:
+          version:      6.6.3
+          aqtversion:   ==3.1.*
+          host:         mac
+          target:       desktop
+          dir:          ${{ runner.temp }}
+          modules:      qtcharts qtlocation qtpositioning qtspeech qtmultimedia qtserialport qtimageformats qtshadertools qtconnectivity qtquick3d
```
HTRamsey, line 54: qt5compat

CMakeLists.txt
```diff
@@ -295,3 +295,11 @@ if(LINUX)
     )
     install(SCRIPT "${CMAKE_SOURCE_DIR}/cmake/CreateAppImage.cmake")
 endif()
+if (MACOS)
```
HTRamsey, line 298: Not important but you could make a CreateDMG.cmake and do 
`install(SCRIPT "${CMAKE_SOURCE_DIR}/cmake/CreateDMG.cmake")`
like I did for the linux AppImage

DonLakeFlyer: Ah, right,. I'll clean that up later.

.github/workflows/macos.yml
DonLakeFlyer: Yep I added that

HTRamsey:
I'm guessing this is replaced by #11369.

PR #11331, opened 2024-04-12
Remove qmake build system

HTRamsey:
Removes files that were only used by qmake, submodules that are now fetched in cmake, and other libs that aren't needed anymore. Looks scary big but it's like 99% SDL.


HTRamsey:
-

PR #11332, opened 2024-04-12
Mavlink: Update submodule

HTRamsey:
Needed for  #11229
A few mission related messages have changed, and I am not yet sure how to accommodate them. I've defaulted them to 0 for now which is what PX4 has while Ardupilot hasn't implemented the update yet. MAVSDK also initializes them to 0.


julianoes (commented):
Yes, 0 is backwards compatible by definition for extensions.

PR #11338, opened 2024-04-13
Remove Vagrant

HTRamsey:
It is not used anymore


peterbarker:
Wait, what?!

I .... still do use this.  A lot.

Who claimed this was no longer used, and on what basis?

HTRamsey:
Discussed on Dronecode Discord. If somebody wants to use it then it will need to be updated for Qt6, so as to not confuse people by having a tool in the source that's not usable.

peterbarker:
> Discussed on Dronecode Discord. If somebody wants to use it then it will need to be updated for Qt6, so as to not confuse people by having a tool in the source that's not usable.

So instead you remove all vestiges of the tool so that nobody knows it's there to be fixed?

Adding two line to the script simply stating, "needs fixing for qt6" and having the script exit would have felt a little less hostile.

A brief look at the commit history would have shown the Vagrantfile setup is patched - usually when it causes me problems.

DonLakeFlyer:
Ah, right. Forgot you used this.  So we have both a docker and vagrant setup. Is there some way to have both of these use the same script or something for setup? It's a pain to have to keep all these little bits up to date when things change.

HTRamsey:
I believe Multipass can read in a docker blueprint.

DonLakeFlyer:
Also is there some explanation as to why we would need both a vagrant and a docker setup?

peterbarker:
On Mon, 15 Apr 2024, Don Gagne wrote:

> Also is there some explanation as to why we would need both a vagrant and a docker setup?

Vagrant actually was able to use Docker to create the machine images. 
That was the case since 2017.  It was also able to use vmware fusion...

I've never used either of those virtualisation solutions; I've always used 
it with virtualbox.

Unfortunately, this adds an extra layer which Docker people aren't 
expecting; the dockerfiles are expected to configure the VM, which is why 
there's duplication within the QGC source tree.

We could probably share some code between the initialisation steps (we 
most certainly do in the ArduPilot repository; we re-use the same scripts 
users run by hand to get a native development environment!).  This would 
have some advantages as I don't see an equivalent to this

     # increase the allowed number of open files (the link step takes a
     # lot of open filehandles!):
echo '*               soft    nofile          2048' 
>/etc/security/limits.d/fileno.conf

in the Dockerfile setup files.  That might bite someone at some point!


However, I'm not sure it's reasonable to expect the people managing the 
Dockerfile configuration to worry about breaking people trying to use 
VirtualBox-based (or VMware-Fusion-based?!) virtual machines.  Using those 
VMs isn't necessarily the same workflow as for Docker; someone added GUI 
pacakges to the VirtualBox VMs because they wanted to run QGC on the VM 
rather than on their native desktop, for example.

DonLakeFlyer:
So I'll revert this so we aren't hosing people. @peterbarker Can you do a couple things:
* After I revert, update the vagrant thing so it works with the latest requirements?
* Update the docs to talk about it's existence
* Create an Issue for further discussion on whether we can shared some setup stuff across docker/vagrant to reduce the task of updating multiple things.

peterbarker:
On Tue, 23 Apr 2024, Don Gagne wrote:

> So I'll revert this so we aren't hosing people. @peterbarker Can you do a couple things:

Thanks Don.

I haven't just been bitching about the problem; I have been poking at 
getting things working with the new build system (cribbing a lot of stuff 
from here and there).

https://github.com/mavlink/qgroundcontrol/pull/11379

PR #11339, opened 2024-04-14
App: Use QCommandLineParser

HTRamsey:
Removes the need for a custom command line parser by using the builtin Qt implementation.


HTRamsey:
@DonLakeFlyer Are [these](https://docs.qgroundcontrol.com/master/en/qgc-dev-guide/command_line_options.html) all still correct? They differ somewhat from what is in the code.

PR #11340, opened 2024-04-14
FTPManager: Added list directory support

DonLakeFlyer:
Also added list directory unit tests to verify.

Not currently used by regular QGC. But useful functionality for custom builds accessing files on companion computers.


DonLakeFlyer:
@HTRamsey Could use a quick sanity check on this

HTRamsey:
I think it looks good, I guess you stepped through the unittest with a debugger to verify? Also just a small comment, I like to use Q_FUNC_INFO inside each qDebug instead of typing out the function names.

DonLakeFlyer:
> I like to use Q_FUNC_INFO inside each qDebug instead of typing out the function names.

Ooh, nice. Didn't know about that.

DonLakeFlyer:
So I tried Q_FUNC_INFO. It's super noisy since it's a full function signature instead of just a function name. Doing it manually with a shorter function name only is much more readable although more of a PITA:
```
QDEBUG : FTPManagerTest::_testLostPackets() FTPManagerLog: void FTPManager::_fillMissingBlocksAckOrNak(const MavlinkFTP::Request *) Ack offset:size 0 239
QDEBUG : FTPManagerTest::_testLostPackets() FTPManagerLog: void FTPManager::_fillMissingBlocksWorker(bool) offset:cBytesToRead 956 239
QDEBUG : FTPManagerTest::_testLostPackets() FTPManagerLog: void FTPManager::_sendRequestExpectAck(MavlinkFTP::Request *) opcode: "Read File" seqNumber: 27
QDEBUG : FTPManagerTest::_testLostPackets() FTPManagerLog: void FTPManager::_mavlinkMessageReceived(const mavlink_message_t &) hdr.opcode:hdr.req_opcode:seqNumber "Ack" "Read 
```

DonLakeFlyer:
Maybe at some point I'll try to make a cross platform one which is just function name. That would be useful.

HTRamsey:
That's also the purpose of [qInstallMessageHandler](https://doc.qt.io/qt-6/qtlogging.html#qSetMessagePattern), where you can just format with the function name.

PR #11341, opened 2024-04-15
CMake: Update Deployment

HTRamsey:
This has some CMake and Android CI fixes including adjustments to multiabi builds, macos deployment settings, and compilation definitions. Adds back Android caching for now. Somehow revealed a small bug in QGCTileCacheWorker.cpp


HTRamsey:
@DonLakeFlyer Let me know if any of the extra macos properties made any differences

DonLakeFlyer:
> Let me know if any of the extra macos properties made any differences

Not sure what you want me to look at for this?

.github/workflows/android.yml
```diff
@@ -128,37 +143,24 @@ jobs:
         env:
           GST_VERSION: 1.18.5
         run: |
-            wget --quiet https://gstreamer.freedesktop.org/data/pkg/android/${GST_VERSION}/gstreamer-1.0-android-universal-${GST_VERSION}.tar.xz
+            wget https://gstreamer.freedesktop.org/data/pkg/android/${GST_VERSION}/gstreamer-1.0-android-universal-${GST_VERSION}.tar.xz
```
DonLakeFlyer, line 146: How come removal of quiet. Without this it makes lookout the the raw logs really tough with a massive amount of wget logs.

CMakeLists.txt
```diff
@@ -198,40 +211,40 @@ endif()
...
+elseif(MACOS)
     set_target_properties(${PROJECT_NAME}
         PROPERTIES
             MACOSX_BUNDLE TRUE
             MACOSX_BUNDLE_INFO_PLIST ${CMAKE_SOURCE_DIR}/deploy/mac/MacOSXBundleInfo.plist.in
+            MACOSX_BUNDLE_BUNDLE_NAME ${PROJECT_NAME}
+            MACOSX_BUNDLE_BUNDLE_VERSION ${PROJECT_VERSION}
+            MACOSX_BUNDLE_COPYRIGHT "Copyright (c) 2018 QGroundControl. All rights reserved."
+            MACOSX_BUNDLE_GUI_IDENTIFIER "io.mavlink.qgroundcontrol"
+            MACOSX_BUNDLE_ICON_FILE "${CMAKE_SOURCE_DIR}/resources/icons/macx.icns"
```
DonLakeFlyer, line 240: Are these the additions you are talking?

src/comm/CMakeLists.txt
```diff
@@ -77,6 +77,14 @@ endif()
 
 target_compile_definitions(comm PUBLIC QGC_ENABLE_BLUETOOTH)
 
+if(NOT EXISTS ${CMAKE_SOURCE_DIR}/libs/mavlink/include/mavlink/v2.0/ardupilotmega)
```
DonLakeFlyer, line 80: What is this for? In general QGC doesn't use ardupilotmega dialect. It uses "all" dialect.

CMakeLists.txt
HTRamsey: Yeah, they were taken from [here](https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#properties-on-targets).

.github/workflows/android.yml
HTRamsey: I had an issue at one point where it wasn't downloading correctly and I couldn't tell but I can put it back.

src/comm/CMakeLists.txt
HTRamsey: I don't know to be honest. It was in the qmake stuff

DonLakeFlyer: I think you can get rid of it

HTRamsey: Should I remove its usage in the code too?

.github/workflows/android.yml
DonLakeFlyer: Would be better without it. It ends up dumping page after page of downloading progress.

src/comm/CMakeLists.txt
DonLakeFlyer: Can you point me at that?

HTRamsey: [NO_ARDUPILOT_DIALECT.txt](https://github.com/mavlink/qgroundcontrol/files/14983023/NO_ARDUPILOT_DIALECT.txt)

DonLakeFlyer: If you are talking about thee NO_ARDUPILOT_DIALECT defines. Those should be there. They allow you to create a custom build which only has a single firmware type. Thus reducing the size of your custom build. So those references in the source to ifdef stuff out should stay.

DonLakeFlyer: There is a PX4 one as well to make an ArduPilot only build. It also reduces some of the complexity in the UI since it's simpler to support a single firmware type.

DonLakeFlyer: So two things:
* As I said NO_ARDUPILOT_DIALECT is used to compile out all ArduPilot specific code. It shouldn't really have "DIALECT" in the name, which confuses things with mavlink dialects. It really NO_ARDUPILOT_SUPPORT.
* MAV_AUTOPILOT_ARDUPILOTMEGA is a firmware type from the mavlink spec: MAV_AUTOPILOT_ARDUPILOTMEGA. Is to determine what firmware type the vehicle is running

DonLakeFlyer: In the past QGC used to use the ArduPilot mavlink dialect to build against instead of common. That's because at that time that dialect was a superset of what Ardupilot and PX4 used. So it made thing in QGC be able to talk mavlink with both firmwares. Also ages ago I decided to stop supporting Ardupilot specific mavlink stuff and just focus on generic across the board support because it was too much of a pain to implement things multiple ways. From the mavlink spec side this has been cleaned up such that the "all" dialect is now the one that all firmwares support. Hence the whole "DIALECT" aspect of "NO_ARDUPILOT_DIALECT" no longer makes any sense. Whereas in the past it made more sense.

HTRamsey: Wait then why should we get rid of the NO_ARDUPILOT_DIALECT compile definition? Or are you referring the check if it exists? Like it should be a cmake option instead of checking if the file exists?

DonLakeFlyer: You should get rid of this check:
```
if(NOT EXISTS ${CMAKE_SOURCE_DIR}/libs/mavlink/include/mavlink/v2.0/ardupilotmega)
   target_compile_definitions(${PROJECT_NAME} PUBLIC NO_ARDUPILOT_DIALECT) # TODO: Make this QGC_NO_ARDUPILOT_DIALECT
endif()
```

You shouldn't get rid of NO_ARDUPILOT_DIALECT.

DonLakeFlyer (approved):

PR #11342, opened 2024-04-16
AudioOutput: Improvements

HTRamsey:
Modernizes AudioOutput with new functionality from [Qt6.6](https://doc-snapshots.qt.io/qt6-6.6/whatsnew66.html#qt-texttospeech-module).

- The queue is now handled internally in the QTextToSpeech object.
- Now available to use from QML.
- No longer dependent on QGCApplication making it an isolated utility available everywhere with a singleton using static methods.
- Inherits from QTextToSpeech making all [functions](https://doc.qt.io/qt-6/qtexttospeech.html) available.
- Makes use of a QHash for faster translations
- Uses InvokeMethod for thread safety

### DIFF
--- a/src/Audio/AudioOutput.cc
+++ b/src/Audio/AudioOutput.cc
@@ -7,191 +7,186 @@
  *
  ****************************************************************************/
 
-#include <QDebug>
-#include <QRegularExpression>
-
 #include "AudioOutput.h"
-#include "QGCApplication.h"
-#include "SettingsManager.h"
 
-AudioOutput::AudioOutput(QGCApplication* app, QGCToolbox* toolbox)
-    : QGCTool   (app, toolbox)
-    , _tts      (nullptr)
+#include <QtCore/QRegularExpression>
+
+#define MAX_TEXT_QUEUE_SIZE 20U
+
+Q_LOGGING_CATEGORY( AudioOutputLog, "qgc.audio.audiooutput" );
+// qt.speech.tts.flite
+// qt.speech.tts.android
+
+const QHash<QString, QString> AudioOutput::s_textHash = {
+    { "ERR",            "error" },
+    { "POSCTL",         "Position Control" },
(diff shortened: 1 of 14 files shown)


src/Audio/AudioOutput.cc
```diff
@@ -7,191 +7,186 @@
...
+    { "RTL",            "return To launch" },
+    { "ACCEL",          "accelerometer" },
+    { "RC_MAP_MODE_SW", "RC mode switch" },
+    { "REJ",            "rejected" },
+    { "WP",             "waypoint" },
+    { "CMD",            "command" },
+    { "COMPID",         "component eye dee" },
+    { "params",         "parameters" },
+    { "id",             "eye dee" },
+    { "ADSB",           "Hey Dee Ess Bee" },
```
DonLakeFlyer, line 34: This one seems a little odd: "Hey" for "A". Isn't that wrong? Might as well fix that.

DonLakeFlyer (changes requested):

src/Audio/AudioOutput.cc
HTRamsey: I checked with espeak on linux and I think you can just do "A.D.S.B.", "I.D.", "E.K.F.", etc. I don't know if that's standard though

DonLakeFlyer: MockLink has a "send status text" option which you could hack into and change the one that is spoken to test it out. I can test it on Mac.

https://github.com/mavlink/qgroundcontrol/blob/master/src/comm/MockLink.cc#L1377

DonLakeFlyer: I verified that "A.D.S.B" is spoken correctly on the Mac. I had to change the MAV_SEVERITY_NOTICE text in MockLink to test. The "#" status text isn't causing it to talk for some reason. Not sure if there is a bug there, I'll look.

PR #11348, opened 2024-04-17
CI: Fix Docs Deploy

HTRamsey:
I guess the output from the build step wasn't working like I had expected


HTRamsey:
It still has the node version change, and uploads the build result artifact, but otherwise yes.

hamishwillee:
> Note, we should also update the other workflows to make sure that they ignore changes in the /docs path.

NOte that again ^^^ - every time I make a docs change it forces a rebuild on everything. That doesn't need to happen.

HTRamsey:
They should all have
```
paths-ignore:
      - 'docs/**'
```

PR #11358, opened 2024-04-22
BatteryIndicator remove duplicate qml Component and add flag for battery percentage availability

ivanzivotic:
BatteryIndicator remove duplicate qml Component and add flag for battery percentage availability

Description
-----------
This is a small change that adds the check for the availability of the percentRemaining data of the battery, since it can be unavailable in some autopilot configurations. Since all other values that can be unavailable have the check of the availability on the BatteryIndicator.qml page, same logic is applied for the percentRemaining Fact. In addition, this pull request removes the duplicate batteryValuesAvailableComponent element of type Component, containing all boolean values that define which data is to be displayed. 

Test Steps
-----------
Connect autopilot sending the -1 value of the field _battery_remaining_ in Mavlink BATTERY_STATUS message and check if data is displayed in the battery popup.

Checklist:
----------

### DIFF
--- a/src/ui/toolbar/BatteryIndicator.qml
+++ b/src/ui/toolbar/BatteryIndicator.qml
@@ -159,12 +159,13 @@ Item {
         id: batteryValuesAvailableComponent
 
         QtObject {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Thanks for removing the duplication. I must have screwed that up when I reworked this for the new indicator support

PR #11363, opened 2024-04-23
Revert "Remove Vagrant"

DonLakeFlyer:
Reverts mavlink/qgroundcontrol#11338. Putting this back since folks are using it.


HTRamsey:
Looks like this can be replaced by #11379

PR #11364, opened 2024-04-23
WIP - OSX cmake dmg

DonLakeFlyer:
So this is what I have been testing with.


libs/qmlglsink/CMakeLists.txt
```diff
@@ -87,11 +87,17 @@ if(QGC_ENABLE_VIDEOSTREAMING)
...
 
-	find_package(PkgConfig)
-	if(NOT EXISTS ${GSTREAMER_ROOT})
-		pkg_check_modules(GSTREAMER gstreamer-1.0>=${GST_TARGET_VERSION})
-		if(GSTREAMER_FOUND)
-			set(GSTREAMER_ROOT ${GSTREAMER_PREFIX})
+	if(MACOS)
+		find_library(GSTREAMER_FRAMEWORK
+			NAMES GStreamer
+			HINTS  ${GSTREAMER_ROOT})
```
DonLakeFlyer, line 93: Because we are using a Framework for gstreamer you use find_library to find it. cmake supports finding frameworks in find_library.

libs/qmlglsink/CMakeLists.txt
```diff
@@ -122,43 +128,48 @@ if(QGC_ENABLE_VIDEOSTREAMING)
...
 		elseif(MACOS)
-			set(ENV{PKG_CONFIG_PATH} "${GSTREAMER_ROOT}/Versions/Current/lib/pkgconfig:$ENV{PKG_CONFIG_PATH}")
-			pkg_check_modules(GST IMPORTED_TARGET ${GST_TARGET_MODULES})
-		else()
-			set(ENV{PKG_CONFIG_PATH} "${GSTREAMER_ROOT}/lib/pkgconfig:$ENV{PKG_CONFIG_PATH}")
-			pkg_check_modules(GST IMPORTED_TARGET ${GST_TARGET_MODULES})
+			if (GSTREAMER_FRAMEWORK-NOTFOUND)
+				message(WARNING "Gstreamer Framework Not Found")			
+			else()
+			    set(GST_FOUND TRUE)
```
DonLakeFlyer, line 137: Since the find_library uses a different mechanism to find the frame the GST_FOUND variable is not set automatically. So I set it manually here. I must be doing this wrong somehow.

DonLakeFlyer: Is this specific CMakelists.txt it does see that GST_FOUND was set to TRUE further down the file.

src/VideoReceiver/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ if(GST_FOUND)
             Settings
     )
 
+    message(FATAL_ERROR "GST_FOUND: ${GST_FOUND}")
```
DonLakeFlyer, line 38: But once it gets here it doesn't see that GST_FOUND is set to TRUE.

src/VideoManager/VideoManager.cc
```diff
@@ -32,6 +32,7 @@
 #include "GStreamer.h"
 #include "VideoSettings.h"
 #else
+#error Boom
```
DonLakeFlyer, line 35: And then I use this to verify that the DEFINE is set correctly which it is not.

DonLakeFlyer:
@HTRamsey I added comments to the changes to show what I'm running into. Probably something simple related to my cmake stupidity.

Issue #10329, opened 2022-06-27
Unknown command problem


ldm0330:
Hello, I would like to ask if I updated the latest version of QGC. When I click to create a waypoint during use, it will appear unknown: 84, unknown: 16 and display all parameters. How to solve this problem?

1chick:
# same issue.
- "add waypoint" button can not be clicked in planview.
![image](https://user-images.githubusercontent.com/20314338/179645819-7e75debf-1000-4221-86e1-bda46596b8ed.png)
- unknow waypoint ,command combox is blank 
![image](https://user-images.githubusercontent.com/20314338/179646678-7880159a-97d4-4cd4-bd20-980e1307aa3a.png)

# Enviroment:
1.QGC Release v4.2.3
2.win10/android os
3.ardupilot,Rover v4.2.2 firmware

LnAYuTaK:
你好，我在version 4.2.0中也遇到这个问题，但是我尝试了上面的解决方法并没有解决这个问题，请问下还有其他需要修改的地方么

Hello, I also encountered this problem in version 4.2.0, but I tried the above solution and failed to solve the problem. Is there anything else that needs to be modified

booo:
Could you check if the proposed change fixes your problem? You can find binaries for testing in the pull request and workflow runs: https://github.com/mavlink/qgroundcontrol/pull/10491

booo:
Fix merged into master.

PR #11367, opened 2024-04-24
translation: Fix translation of Item #%1

ROSStargh:
Same bug as in
#10491 

This bug would prevent creating waypoints in plan view.
#10329


DonLakeFlyer:
@ROSStargh These sort of changes need to be made in CrowdIn and then they come back down to QGC automatically from there: https://docs.qgroundcontrol.com/master/en/qgc-dev-guide/#translations

PR #11376, opened 2024-04-25
CI: Add MacOS gstreamer pkgconfig path

HTRamsey:
Using pkgconfig is the easiest way to find gstreamer and dependencies in cmake.

### DIFF
--- a/.github/workflows/macos_debug.yml
+++ b/.github/workflows/macos_debug.yml
@@ -73,6 +73,7 @@ jobs:
             for package in *.pkg ;
               do sudo installer -verbose -pkg "$package" -target /
             done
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
This is all going away. Can you wait on any further macOS stuff until I finish my work. I'm getting conflicts which are slowing me down.

HTRamsey:
Yeah sorry for some reason CI switched to MacOS Arm in the middle of no where and I was trying to figure out what the problem was.

DonLakeFlyer:
> Yeah sorry for some reason CI switched to MacOS Arm in the middle of no where

Yeah, what a PITA!

PR #11380, opened 2024-04-26
libevents: update submodule

bkueng:
Simplifies the integration a bit. We can also use `ExternalProject_Add` instead of the submodule when removing qmake completely.


HTRamsey:
Can you rebase? Qt tools source link was down

bkueng:
I rebased, it's passing now.

PR #11393, opened 2024-04-28
Mavlink: Separate Implementations into new Library

HTRamsey:
Copies much of the mavlink implementation from Analyze View and puts them into their own files in a new library within Comm.
GeoTags should be separated from Analyze View and put into Camera or Utilities or something at some point so that Mavlink lib doesn't rely on AnalyzeView.

There are some things currently in the mavlink files related to Analyze View that don't make sense to be there, but I just wanted this to be a copy/paste thing and not affect any functionality for now.


DonLakeFlyer:
If we're moving things around. Is the whole additional level of "comm" needed? Doesn't seem to add much value. Mavlink could just be right under src. Easier to find. Don't remember what else is in "comm".

DonLakeFlyer:
I like the direction we are heading here. One thing mavlink related if you are interested: If the mavlink inspector backed code gets factored out into a more generic mavlink telemetry object model then theoretically that could be hooked up to Vehicle and allow the Telemetry Values bar access to display any values from any mavlink telemetry. Instead of having to create new FactGroups for everything you can display in the Telemetry UI.

HTRamsey:
This is probably about ready.

PR #11394, opened 2024-04-28
APM: Update Parameter Repository

HTRamsey:
Brings Ardupilot params up to date again. Still need to make this autorun.
Also auto searches for most recent parameters available if vehicle version is higher than the meta data files available.
Previously it would just use the oldest version if the requested one was unavailable.
If the vehicle version is invalid, then search for the oldest available file.

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -686,51 +686,66 @@ void APMFirmwarePlugin::_artooSocketError(QAbstractSocket::SocketError socketErr
     qgcApp()->showAppMessage(tr("Error during Solo video link setup: %1").arg(socketError));
 }
 
-QString APMFirmwarePlugin::_internalParameterMetaDataFile(const Vehicle* vehicle) const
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
I don't know if you have any pull with the ArduPilot folks. But this whole having to update parameter metadata and build into QGC is a solved problem in mavlink. The COMPONENT_INFORMATION support allows parameter metadata to come right from the vehicle. This way you don't need to update QGC every time the metadata changes. Sure would be nice if the Ardupilot folks supported that. Then we could get rid of this whole thing.

DonLakeFlyer (approved):

HTRamsey:
I'd have to double check but I think it does already use that and this is in-case that fails maybe?

DonLakeFlyer:
> but I think it does already use that and this is in-case that fails maybe?

Hmm, interesting. Not sure that's super useful then. Lot's of work to keep up to date for something that isn't supposed to fail. That just hides some other possibly real problem.

mrcdnov:
@DonLakeFlyer do I understand correct, that QGC pull metadata, in my case I'm interested in custom parameters and its description, via mavlink from vehicle ?

PR #11425, opened 2024-04-30
Rework PX4 airspeed cal due to parameter name/functionality changes

### LINKED ISSUE #11252: The latest QGroundControl and the latest PX4 firmware parameters do not match.
## Current Behavior
1:FW_ARSP_MODE, 1:CBRK_AIRSPD_CHK
image

## System Information
- Operating System: ubuntu and windows
- QGC Version: v4.3.0
- Flight Controller:V2.4.8
- Autopilot (with version): PX4 [Stable Release v1.14.0]

DonLakeFlyer:
Fix #11252

### DIFF
--- a/src/AutoPilotPlugins/APM/APMRadioComponent.cc
+++ b/src/AutoPilotPlugins/APM/APMRadioComponent.cc
@@ -124,7 +124,7 @@ void APMRadioComponent::_connectSetupTriggers(void)
 
 void APMRadioComponent::_triggerChanged(void)
 {
-    emit setupCompleteChanged(setupComplete());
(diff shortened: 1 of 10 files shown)


sfuhrer (commented):
I would simplify it and only use SYS_HAS_NUM_ASPD, and enforce the calibration whenever that one is >0. 
FW_USE_AIRSPD doesn't need to be checked.

DonLakeFlyer:
> I would simplify it and only use SYS_HAS_NUM_ASPD, and enforce the calibration whenever that one is >0.
> FW_USE_AIRSPD doesn't need to be checked.

Yup, makes sense

PR #11439, opened 2024-05-01
Issues: Add Question Template

HTRamsey:
Try to limit random unrelated questions


HTRamsey:
My thinking is people are going to do it anyways, at least you can filter it out using this haha

DonLakeFlyer:
Then I was add some text saying where to ask questions, which is not here! With links tot he right places.  And then 90% of folks will ignore it but at least we cut down 10%

PR #11447, opened 2024-05-02
CI: Free up more space in linux builds

HTRamsey:
The OpenSSL submodule is taking up over 50% of allowed space, so just delete it after cloning repo in linux.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -50,6 +50,13 @@ jobs:
           submodules: recursive
           fetch-tags: true
 
+      # Temporary solution
+      - name: Remove unused dependencies
+        run: |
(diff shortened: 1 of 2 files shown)


TSC21:
Screenshot from 2024-05-07 22-08-42

@HTRamsey oddly enough, I am running out of space in the Debug build for Android. Would you say there are other things we can do to further reduce the space being used? This is currently always happening for the Android build for some reason.

HTRamsey:
The android build is a multiabi build of armv7 and arm64v8a, so maybe just use the particular one you need. Or since CI builds with cmake just tell it to not recursively checkout submodules, since they are fetched via cmake.

PR #11449, opened 2024-05-02
QML: Move ui files into ui folder

HTRamsey:
It's generally cleaner to keep QML related stuff together.


DonLakeFlyer:
I disagree. Most qml is backed by C++ controller code. Having to go two different places to get things which are functionally connected together doesn't seem like better to me. When looking at the file hierarchy in an editor this:

Screenshot 2024-05-02 at 8 20 42 AM

Shows me the structure of the code much better than having to go all over the place to see the connections.

HTRamsey:
Yup that's why I didn't move the cpp related stuff like Analyze View. This is strictly Flight Display/Flight Map related Qml files

DonLakeFlyer:
I still don't see a lot of value in an extra level of hierarchy. I would go the opposite way and get rid of the ui directory. Moving up toolbar to Toolbar under src and moving the preferences files to Settings. Then whether things are only Qml or Qml + C++ they are done the same way with Functional or UI Area naming as the top most dir under src.

HTRamsey:
Up to you, I can close this and let you decide later if you want to do anything with it. I think traditionally Qml files were kept together because of how hard it was to get the qmldir stuff set up correctly, but that shouldn't be as much of a problem in Qt6.

PR #11470, opened 2024-05-04
Reorganize some utility files

HTRamsey:
Moves multiple things out of MissionManager and into their respective appropriate modules.


DonLakeFlyer (commented):
The RallyPoint stuff isn't really associated with a Vehicle. It is solely a UI mechanism used by Plan View editing. It's not a generic thing. It's only valid in relationship to ReallyPointController which is in MissionManager.

DonLakeFlyer (commented):
Similar problems with CameraMetaData, CameraSpes: These are not really wholly generic things. It only relates to camera information as used by mission planning. I think moving them to a generic location make it less clear what these are used for.

HTRamsey:
Hmm, I think at least moving the KMLDocument stuff makes sense, since in the past you had mentioned being able to use shape stuff elsewhere. With the rally point, I suppose that's true. 

The CameraSpec stuff looks like it could be merged with CameraMetaData because they hold the same data, just CameraSpec is also meant for qmlobjectlistmodels.

RallyPoint, I do see that that they are only used in Mission based Mavlink commands.

DonLakeFlyer:
> I do think think the other shapes should be available in FlyView, because for example it would be nice to display your GeoFence in FlyView (Which is one of the changes I needed in my custom QGC design).

Yeah the other ones are generic and make sense to be in a generic location.

DonLakeFlyer:
> The CameraSpec stuff looks like it could be merged with CameraMetaData

Yeah, I vaguely remember doing that ages ago and thinking I should really clean that up somehow.

PR #11473, opened 2024-05-05
Utilities: Create DeviceInfo

HTRamsey:
This was to try and reduce dependencies on the qgcapplication. The map module shouldn't be responsible for handling internet, so I made a new file called DeviceInfo that can hold functions related to capabilities of the current device. Also made the jsonhandler slightly more independent.


src/Comms/LinkManager.h
```diff
@@ -51,7 +51,6 @@ class LinkManager : public QGCTool
     LinkManager(QGCApplication* app, QGCToolbox* toolbox);
     ~LinkManager();
 
-    Q_PROPERTY(bool                 isBluetoothAvailable            READ isBluetoothAvailable            CONSTANT)
```
DonLakeFlyer, line 54: The remove of this is the only thing I wonder about. There isn't any Qml currently in regular QGC that I can find that references this. But I wonder if custom build folks have been relying on it? Maybe the network and bluetooth available information should be piped out QGroundControl global thing show Qml still has access?

HTRamsey: Yup, that would work. I thought it would have been used too but was surprised not to find it anywhere. We could just give Qml access to DeviceInfo or something so that all of the functions added there are available.

HTRamsey: we can tackle that later, I'll just leave it as it is for now.

PR #11477, opened 2024-05-05
Add QmlObjectListItem

HTRamsey:
Provides the required functions to be used in a QmlObjectListModel.
This cuts a bit down on repeated code and makes it more clear where these classes are used.
@DonLakeFlyer This is what I was talking about in my reply to the other PR.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -513,6 +513,7 @@ HEADERS += \
     src/QmlControls/QGCQGeoCoordinate.h \
     src/Utilities/QGCTemporaryFile.h \
     src/QGCToolbox.h \
+    src/QmlControls/QmlObjectListItem.h \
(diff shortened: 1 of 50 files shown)


DonLakeFlyer:
Yeah, makes sense

HTRamsey:
I think this is worthwhile, but it needs a thorough double checking. Also, I don't know if QmlObjectListModel should be changed to specifically hold QmlObjectListItems or not.

DonLakeFlyer:
> Also, I don't know if QmlObjectListModel should be changed to specifically hold QmlObjectListItems or not.

50/50. A decent amount of usage of QmlObjectListModel doesn't use dirty bits. Maybe the QmlObjectListItem thing should just be something more generic to indicate a generic QObject with a dirty bit. Not sure switching the ones that don't use a dirty bit adds value.

PR #11487, opened 2024-05-06
CI: Action to install Qt

HTRamsey:
Currently only meant for linux. This way we have more control over setup and don't have to rely on the action being updated.

### DIFF
--- a/.github/actions/dependencies/action.yml
+++ b/.github/actions/dependencies/action.yml
@@ -1,0 +1,70 @@
+name: Dependencies
+description: Install Dependencies
+inputs:
+  target:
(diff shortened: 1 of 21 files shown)


DonLakeFlyer:
Would be nice to carry this through to the other OS as much as possible. Would remove a ton of duplication.

HTRamsey:
I'll look into it soon. Do we need any setup scripts for windows or mac to put in the tools? then I can just call those

DonLakeFlyer:
> Do we need any setup scripts for windows or mac to put in the tools?

Mac, no. Don't remember windows, but don't think so. Other than the NSIS install.

PR #11490, opened 2024-05-06
Clean up some old method Qml Registrations

HTRamsey:
These will be removed anyways, but clean up a little now.


HTRamsey:
@DonLakeFlyer this fixes all of these, so maybe there will be better QML linting in QtCreator.
image

PR #11515, opened 2024-05-10
Hardware: Add support for JFB110

jfbblue0922:
This adds support for the new JFB110 board of Japan Aviation Electronics Industry Ltd.

sorry, I closed PR by mistake, so I'll post it again.

### DIFF
--- a/src/Comms/USBBoardInfo.json
+++ b/src/Comms/USBBoardInfo.json
@@ -40,6 +40,8 @@
 
         { "vendorID": 11694, "productID": 0,        "boardClass": "Pixhawk",    "name": "CubePilot" },
 
+        { "vendorID": 2702,  "productID": 110,      "boardClass": "Pixhawk",    "name": "JFB JFB110" },
+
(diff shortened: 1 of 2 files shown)


TSC21:
Is this board already supported and included in the PX4 repos?

jfbblue0922:
@TSC21 

JFB110 is not supported yet by PX4 repository, but we plan to make a PR soon.
we made this PR first because we are using QGC in our current development.

rmackay9:
This board is supported by [AP and our wiki is here](https://ardupilot.org/copter/docs/common-jae-jfb110.html).

I think this PR adds support for QGC to be able to upload AP firmware to the board perhaps?  In which case I think it should be OK to merge it even if PX4 support is not complete.

I've used the board successfully with AP, I haven't tried uploading using QGC with this PR applied (but I don't think that should necessarily hold up merging this change).

rmackay9:
Thanks @HTRamsey!

jfbblue0922:
@rmackay9
@HTRamsey

Thank you everyone for your support.

PR #11520, opened 2024-05-10
Vehicle: Move Generic Code to MAVLink Module

HTRamsey:
This is generic mavlink code and doesn't have to belong to a vehicle type.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSensorsComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMSensorsComponent.qml
@@ -22,6 +22,7 @@ import QGroundControl.ScreenTools
 import QGroundControl.Controllers
 import QGroundControl.ArduPilot
 import QGroundControl.QGCPositionManager
+import MAVLink
 
 SetupPage {
     id:             sensorsPage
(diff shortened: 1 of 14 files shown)


TSC21:
I am not opposed to the fact that this is being developed at a fast pace, but since this is an open-source project, even if it is the maintainer merging it, shouldn't there be at least a person reviewing and approving the PRs? @HTRamsey @DonLakeFlyer @mrpollo @patrickelectric @peterbarker

TSC21:
> If there are people willing to keep up with it then absolutely yes. I think @DonLakeFlyer ran into the problem of nobody filling in for when he took a hiatus however. I mostly came to help w/ the Qt6/cmake conversion since I had done it before, although I've been trying to help in other areas too that are mostly trivial non-functional cleanups. But I'm cool with calling it quits here and branching off onto my own build again too.

@HTRamsey everyone appreciates your work here and we are glad you can help. I am jus suggesting there probably should be a process in place so at least one approval is required over the PRs so changes can be merged. Maybe that can be discussed in a dev call? Not wanting to make the process slower, bur rather minimally supervised, since this is an open-source project that thousands of people will use and rely on. Thanks

PR #11525, opened 2024-05-13
QtAndroidSerialPort: Update QSerialPortInfo

HTRamsey:
Rebases the custom QtSerialPort QSerialPortInfo class for android onto the latest version from Qt. This is the first step in updating the android serial driver. Ultimately the plan is to remove this whole custom module.


mrpollo:
Awesome work, were you able to test locally with a Pixhawk or similar?

HTRamsey:
Yeah I generally use a samsung tab active4 pro and cube orange+ when I test. Although I have to change one thing real quick. By the way I risked the cube and after #11413 I was able to successfully update the firmware on the cube via the tablet, which was previously impossible.

mrpollo:
Wow that's huge! I don't have an android tablet available so I can't test. I'll get this in shortly.

HTRamsey:
I should have marked it as draft rather than just saying "Although I have to change one thing real quick" haha. I found a bug in Qt that I need to add one change for. We can revert this or I can just make another PR.

PR #11530, opened 2024-05-14
PositionManager: Update Sources

HTRamsey:
Preps PositionManager to use more positioning sources. I have another PR coming up that preps the GPS module to use external connected gps's (other than base stations for RTK, which is currently the only GPS type supported) for positioning sources. This adds the ability to use QtSensors compass for the heading, but doesn't enable it yet (although in testing it works and shows up in Qml). Also does general code quality improvements.

### DIFF
--- a/src/PositionManager/CMakeLists.txt
+++ b/src/PositionManager/CMakeLists.txt
@@ -1,4 +1,4 @@
-find_package(Qt6 REQUIRED COMPONENTS Core Positioning)
+find_package(Qt6 REQUIRED COMPONENTS Core Qml Positioning)
 
 qt_add_library(PositionManager STATIC
     PositionManager.cpp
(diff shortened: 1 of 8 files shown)


DonLakeFlyer (approved):
Have you checked the GPS heading accuracy on a tablet? Last time I tried that from Qt it say heading was accurate but it still was reporting crap.

HTRamsey:
> Have you checked the GPS heading accuracy on a tablet? Last time I tried that from Qt it say heading was accurate but it still was reporting crap.

It was within maybe a range of +/- 20 degrees on a samsung active which isn't great. But the compass device provides a calibration level which can somehow be translated to relative accuracy maybe (wish it was just +/- degrees) and also allows for easy adding of additional backends. So with that serial driver update it would be pretty easy to add any external gps/compass/temperature sensor or whatever to use as data sources in custom builds

DonLakeFlyer:
I also vaguely remember some ArduPilot follow me support not being fully implemented since it required a good heading from the ground. If we can get good heading now with this change maybe that can be fixed as well.

Issue #8304, opened 2020-02-10
Slower telemetry links, support for SET_MESSAGE_INTERVAL command


cooldemo:
Ardupilot since V4 has implemented different approach to selecting telemetry stream rates that allows for greater granularity. This allows to lower the amount of telemetry that is generated and needs to be transferred to the GCS.
It would be nice if this was also implemented in QGC along with the graphical configuration tool to select interval for every telemetry message.
The change involves transmitting command:
MAV_CMD_SET_MESSAGE_INTERVAL
instead of
REQUEST_DATA_STREAM

DonLakeFlyer:
QGC supports adjust the stream rates to what you want:
![Screen Shot 2020-02-10 at 10 07 13 AM](https://user-images.githubusercontent.com/5876851/74176747-2a771100-4bed-11ea-9638-7c9d81a26e43.png)

Adjusting on a message by message basis is sort of interesting but not something I see happening anytime soon.

cooldemo:
I haven't seen this settings on Android and this only uses REQUEST_DATA_STREAM command. You have only 7 settings to adjust and telemetry rate was quite high up to 30kbps with all settings on 1Hz.
MAV_CMD_SET_MESSAGE_INTERVAL allows much finer granularity and intervals between messages up to 60 seconds.

DonLakeFlyer:
> I haven't seen this settings on Android and this only uses REQUEST_DATA_STREAM command.

They are available in latest daily builds.

cooldemo:
Is there any plan to remove deprecated REQUEST_DATA_STREAM interface and switch to MAV_CMD_SET_MESSAGE_INTERVAL ?
This would enable lowering telemetry rates to minimum, possibly supporting slower interfaces over longer range.

DonLakeFlyer:
> Is there any plan to remove deprecated REQUEST_DATA_STREAM interface and switch to MAV_CMD_SET_MESSAGE_INTERVAL ?

No.

> This would enable lowering telemetry rates to minimum, possibly supporting slower interfaces over longer range.

They way people are doing that is to turn off GCS control of streaming and control from vehicle side through parameters.

cooldemo:
> > Is there any plan to remove deprecated REQUEST_DATA_STREAM interface and switch to MAV_CMD_SET_MESSAGE_INTERVAL ?
> 
> No.
> 
> > This would enable lowering telemetry rates to minimum, possibly supporting slower interfaces over longer range.
> 
[...]

PR #11531, opened 2024-05-14
Do not REQUEST_DATA_STREAMs by default, since it's deprecated

TishSerg:
Proof: https://mavlink.io/en/messages/common.html#REQUEST_DATA_STREAM

Description
-----------
MavLink command REQUEST_DATA_STREAM is DEPRECATED: Replaced by MAV_CMD_SET_MESSAGE_INTERVAL. 
So, QGroundControl shouldn't use it by default.

Test Steps
-----------

1. Install QGroundControl on clean machine where it wasn't installed before or wipe QGroundControl saved settings.
2. Go to Application Settings, Telemetry, Stream Rates.
3. See default setting of "Controlled By vehicle" checkbox slider (when it's ON it means REQUEST_DATA_STREAM is disabled).

Checklist:
----------

Related Issue
-----------
#8304 

### DIFF
--- a/src/Settings/App.SettingsGroup.json
+++ b/src/Settings/App.SettingsGroup.json
@@ -253,7 +253,7 @@
     "name":                 "apmStartMavlinkStreams",
     "shortDesc":     "Request start of MAVLink telemetry streams (ArduPilot only)",
     "type":                 "bool",
-    "default":         true,
+    "default":         false,
     "qgcRebootRequired":    true
 },
 {


HTRamsey (approved):
Related to #11263

HTRamsey:
I'd have to double check because I always use a message-intervals-chan1.txt file, but I thought they were generally sent by default.

TishSerg:
Sorry, I don't understand what you're talking about.

> message-intervals-chan1.txt

Also, I can't find that file among QGC sources.

HTRamsey:
It's somewhere in the ardupilot documentation, it's the way to set message rates on bootup

Davidsastresas:
I think this is good. Indeed by default this is usually not needed. So I think it is good that it is disabled by default, but the user can enable if needed for some reason.

This being enabled by default without users expecting it is problematic so I agree with the change.

PR #11534, opened 2024-05-14
Convert static const to constexpr

HTRamsey:
This is a cleanup/optimization mostly of jsonkeys using constexpr


mrpollo:
Hey @HTRamsey I'm a bit scared out of this seemingly safe optimization, I'm wondering if there's anything our test coverage might miss here.

Is this ready for final review?

DonLakeFlyer (commented):
I thought for just variables not functions that `constexpr` implied `const` so you don't need to double up on `constexpr const`

HTRamsey:
I believe that's true for things like ints, but I think for something like a char array constexpr implies `char * const var` rather than `const char * var`

PR #11535, opened 2024-05-14
CI: Create Build install-qt-action action

HTRamsey:
Action to build the typescript action we use to install qt. We need to build this because the latest release lacks fixes from master. Probably will be replaced by a manual qt installation eventually.


.github/workflows/android.yml
```diff
@@ -81,24 +81,8 @@ jobs:
...
-          cache-dependency-path: install-qt-action/action/
-
-      - name: Build jurplel/install-qt-action
-        run: |
-          cd install-qt-action/action/
-          npm ci || npm install
-          npm run build
-        shell: bash
+      - name: Build install-qt-action
+        uses: ./.github/actions/qt
```
mrpollo, line 85: Should this step be used across workflows? or exclusively on Android builds?

HTRamsey: It can be used in all of them, but it's only required in Android until that repository uploads a new built release. Their master currently contains some updates that the Android build process needs, but their previous release still works for all the other builds.

HTRamsey:
They just released an update so we may not need to do this anymore: https://github.com/jurplel/install-qt-action/releases/tag/v4.0.0

DonLakeFlyer (approved):

PR #11545, opened 2024-05-16
Video: Add QtMultimedia Receiver

HTRamsey:
Creates an alternate video receiver backend that uses QtMultimedia. The UVC Receiver is meant to replace FlightDisplayViewUVC.qml and provide a single widget with a changeable video sink. The VideoManager requires a bit of reworking for multiple video sources before actually enabling this so at the moment this is dead/isolated code.

Edit: Since in the past gstreamer has had so many issues for people, this could be a nice backup solution that provides each platform with two additional video backends (ffmpeg + a platform specific one).


src/VideoManager/VideoReceiver/QtMultimedia/QtMultimediaReceiver.cc
```diff
@@ -0,0 +1,315 @@
...
+
+    qCDebug(QtMultimediaReceiverLog) << Q_FUNC_INFO << this;
+}
+
+QtMultimediaReceiver::~QtMultimediaReceiver()
+{
+    qCDebug(QtMultimediaReceiverLog) << Q_FUNC_INFO << this;
+}
+
+// -1 - disable buffer and video sync
```
DonLakeFlyer, line 124: This should use the doxygen `///` and `@param` format.

DonLakeFlyer (approved):
Jusrts need to fix up commenting

PR #11551, opened 2024-05-17
AnalyzeView: Add ULogParser UnitTest

HTRamsey:
Just need someone to provide sample logs


mrpollo:
Can you grab something out of here? https://logs.px4.io/browse

HTRamsey:
Is there an easy way to tell which log will provide camera feedback info? The only public function in ULogParser checks for camera_capture packets

DonLakeFlyer (approved):

PR #11555, opened 2024-05-19
Joystick: Fix SDL Closing

HTRamsey:
- Enable proper closing of SDL Joysticks (disabled due to crashes during closing).
- Fix double updating of SDL joysticks each loop of thread.
- Disable SDL Joystick Events if subsystem failed to initialize (probably redundant but whatever).

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -113,6 +113,8 @@ Joystick::Joystick(const QString& name, int axisCount, int buttonCount, int hatC
     , _multiVehicleManager  (multiVehicleManager)
     , _customActionManager  (qgcApp()->toolbox()->settingsManager()->customMavlinkActionsSettings()->joystickActionsFile())
 {
+    // qCDebug(JoystickLog) << Q_FUNC_INFO << this;
+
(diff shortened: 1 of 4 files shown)


src/Joystick/Joystick.cc
```diff
@@ -113,6 +113,8 @@ Joystick::Joystick(const QString& name, int axisCount, int buttonCount, int hatC
     , _multiVehicleManager  (multiVehicleManager)
     , _customActionManager  (qgcApp()->toolbox()->settingsManager()->customMavlinkActionsSettings()->joystickActionsFile())
 {
+    qCDebug(JoystickLog) << Q_FUNC_INFO << this;
```
DonLakeFlyer, line 116: Should these remain in the production code? Seems more like useful for detailed debugging of a specific scenario as opposed to generic logging. Maybe keep in commented out for usage as needed?

HTRamsey: Sure, I usually just use it to verify correct usage of QObject parenting cleanup

DonLakeFlyer (approved):

PR #11558, opened 2024-05-20
CI: Update Install Qt Action

### LINKED ISSUE #11285: CI: Revert Temporary Android Build Workaround
The changes from #11274 can be reverted once the upstream action is fixed.

HTRamsey:
We don't need to build it manually anymore now that they have updated the release version.
Closes #11285

### DIFF
--- a/.github/workflows/android-macos.yml
+++ b/.github/workflows/android-macos.yml
@@ -38,40 +38,23 @@ jobs:
           distribution: temurin
           java-version: 17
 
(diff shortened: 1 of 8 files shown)


HTRamsey:
Nah, I had #11535 and another branch that used aqtinstall but they aren't merged. Previously we downloaded the master branch of the repo for the install-qt-action and built it manually, but this will just use the release version.

PR #11571, opened 2024-05-30
Camera: switch to new REQUEST_MESSAGE commands

julianoes:
Description
-----------
The used commands like REQUEST_CAMERA_SETTINGS and REQUEST_CAMERA_INFORMATION are deprecated. Therefore, we should gradually try to move to the new REQUEST_MESSAGE commands.

This commit does so by sending REQUEST_MESSAGE by default but falling back to the previous commands on the second try and every other retry after that.

Test Steps
-----------
Need to test it manually with one of my [camera-manager example](https://github.com/julianoes/siyi-a8-mini-camera-manager).

Checklist:
----------

### DIFF
--- a/src/Camera/VehicleCameraControl.cc
+++ b/src/Camera/VehicleCameraControl.cc
@@ -698,7 +698,7 @@ VehicleCameraControl::_mavCommandResult(int vehicleId, int component, int comman
                 break;
                 case MAV_CMD_IMAGE_START_CAPTURE:
                 case MAV_CMD_IMAGE_STOP_CAPTURE:
-                    if(++_captureInfoRetries < 3) {
+                    if(++_captureInfoRetries < 5) {
                         _captureStatusTimer.start(1000);
                     } else {
                         qCDebug(CameraControlLog) << "Giving up start/stop image capture";
@@ -706,14 +706,14 @@ VehicleCameraControl::_mavCommandResult(int vehicleId, int component, int comman
                     }
                     break;
                 case MAV_CMD_REQUEST_CAMERA_CAPTURE_STATUS:
-                    if(++_captureInfoRetries < 3) {
+                    if(++_captureInfoRetries < 5) {
                         _captureStatusTimer.start(500);
                     } else {
                         qCDebug(CameraControlLog) << "Giving up requesting capture status";
                     }
                     break;
                 case MAV_CMD_REQUEST_STORAGE_INFORMATION:
-                    if(++_storageInfoRetries < 3) {
+                    if(++_storageInfoRetries < 5) {
                         QTimer::singleShot(500, this, &VehicleCameraControl::_requestStorageInfo);
                     } else {
                         qCDebug(CameraControlLog) << "Giving up requesting storage status";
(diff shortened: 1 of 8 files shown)


src/Camera/VehicleCameraControl.cc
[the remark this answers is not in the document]
julianoes: I can't have a close look now but I Randy tested it and it worked apparently. If your testing is different, then we have to address it.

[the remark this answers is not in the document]
HTRamsey: Oh nvm if it doesn't support the REQUEST_MESSAGE msgid then it will return MAV_RESULT_FAILED. At some point we should handle MAV_RESULT_UNSUPPORTED somewhere but doesn't matter for this PR.

HTRamsey (approved):

DonLakeFlyer:
Vehicle has built in request message with retries support:
```
    /// Requests the vehicle to send the specified message. Will retry a number of times.
    ///     @param resultHandler Callback for result
    ///     @param resultHandlerData Opaque data passed back to resultHandler
    void requestMessage(RequestMessageResultHandler resultHandler, void* resultHandlerData, int compId, int messageId, float param1 = 0.0f, float param2 = 0.0f, float param3 = 0.0f, float param4 = 0.0f, float param5 = 0.0f);
```

You should use that instead of building your own.

test/Vehicle/SendMavCommandWithSignallingTest.cc
```diff
@@ -38,6 +38,18 @@ void SendMavCommandWithSignallingTest::_testCaseWorker(TestCase_t& testCase)
 
     QCOMPARE(spyResult.wait(10000), true);
     QList<QVariant> arguments = spyResult.takeFirst();
+
+    // Gimbal controler requests MAVLINK_MSG_ID_GIMBAL_MANAGER_INFORMATION on vehicle connection,
+    // and that messes with this test, as it receives response to that command instead. So if we 
+    // are taking the response to that MAV_CMD_REQUEST_MESSAGE, we discard it and take the next
+    // Also, the camera manager requests MAVLINK_MSG_ID_CAMERA_INFORMATION on vehicle connection,
+    // so we need to ignore that as well.
+    while (arguments.at(2).toInt() == MAV_CMD_REQUEST_MESSAGE || arguments.at(2).toInt() == MAV_CMD_REQUEST_CAMERA_INFORMATION) {
```
DonLakeFlyer, line 47: This is kinda nasty and fragile. There is already the concept of creating a connect to a mock link vehicle which does not do the initial vehicle connect sequence. This is specifically for unit test problems like this. Would it be possible to hook in the gimbal request stuff into the vehicle initial connect sequence state machine so that a it only happens if that was run?

julianoes: I can try that but I will have to do a bit more reading to understand these tests. It's all a mystery to me.

DonLakeFlyer:
Let me take a look at it. I’ll see what I can do

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

test/Vehicle/SendMavCommandWithSignallingTest.cc
DonLakeFlyer: Don't worry about it. I'll clean this up after it's merged in. Will be easier for me to deal with it than have you figure out something new to you.

DonLakeFlyer: Same thing the the request message stuff. I can switch that over as well.

DonLakeFlyer:
@julianoes So in looking through these changes the main functional change is the checking of compids is a little different on the camera manager. The rest is to move to request message. Instead of merging this in and then me futzing with it. Can we instead just leave this pull here for now. And I will create a replacement pull for it. That will be easier.

DonLakeFlyer:
Ok, so I'm confused now. I just went to create a pull for this and it looks like this change is already in master:

https://github.com/mavlink/qgroundcontrol/blob/master/src/Camera/QGCCameraManager.cc#L406

DonLakeFlyer:
Didn't look at the date on this. This is super old. The changes in the pull are already in master. I'm working on cleaning up master per my above comments. @julianoes I think you can close this?

julianoes:
Oh, I'm sorry for the confusion. I didn't look at the date either.

PR #11538, opened 2024-05-15
Ardupilot: Add Baro Alt Offset Compensation

HTRamsey:
Uses available pressure & temperature sensors to adjust altitude offset on drones for long flights using the params BARO_GND_TEMP & BARO_ALT_OFFSET. No current method of enabling, but the logic is there.

PR #11572, opened 2024-05-30
Utilities: fixup warnings and linking

julianoes:
This fixes:
- tabs instead of spaces
- reordering of variables
- linking in qmake

Fixup after #11538.

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -258,7 +258,8 @@ QT += \
     texttospeech \
     core-private \
     core5compat \
-    quick3d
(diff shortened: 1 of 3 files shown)


HTRamsey:
@julianoes thank you for the qmake fix again. I wonder if we should have a qmake CI build up temporarily until custom builds are completed for cmake. It's frustrating having to keep up with both.

HTRamsey (approved):

julianoes:
Depends on the timeline of cmake, up to you.

PR #11573, opened 2024-05-30
CI: Create Temporary QMake Build

HTRamsey:
Create a qmake ci build just to verify builds pass until qmake is dropped entirely. Don't build Android Debug as a multi-ABI (which is pointless if you aren't uploading the APK anyways) to compensate for added build time a bit.


julianoes (approved):
Great, thanks!

julianoes:
Looks like it built nicely!

PR #11574, opened 2024-05-30
Camera: allow camera managed by an autopilot

julianoes:
Description
-----------
By querying autopilot for the CAMERA_INFORMATION message, we allow an autopilot to be a proxy for a MAVLink or non-MAVLink camera.

The idea is similar to a gimbal manager implemented by an autopilot.

Test Steps
-----------
Test it with ArduPilot managing a camera. FYI @rmackay9.

Checklist:
----------

Related Issue
-----------

https://github.com/mavlink/mavlink/pull/2121

### DIFF
--- a/src/Camera/QGCCameraManager.cc
+++ b/src/Camera/QGCCameraManager.cc
@@ -78,7 +78,8 @@ void QGCCameraManager::_vehicleReady(bool ready)
 void QGCCameraManager::_mavlinkMessageReceived(const mavlink_message_t& message)
 {
     //-- Only pay attention to camera components, as identified by their compId
-    if(message.sysid == _vehicle->id() && (message.compid >= MAV_COMP_ID_CAMERA && message.compid <= MAV_COMP_ID_CAMERA6)) {
+    if(message.sysid == _vehicle->id() && (message.compid == MAV_COMP_ID_AUTOPILOT1 ||
+        (message.compid >= MAV_COMP_ID_CAMERA && message.compid <= MAV_COMP_ID_CAMERA6))) {
         switch (message.msgid) {
             case MAVLINK_MSG_ID_CAMERA_CAPTURE_STATUS:
                 _handleCaptureStatus(message);


src/Camera/QGCCameraManager.cc
```diff
@@ -78,7 +78,8 @@ void QGCCameraManager::_vehicleReady(bool ready)
 void QGCCameraManager::_mavlinkMessageReceived(const mavlink_message_t& message)
 {
     //-- Only pay attention to camera components, as identified by their compId
-    if(message.sysid == _vehicle->id() && (message.compid >= MAV_COMP_ID_CAMERA && message.compid <= MAV_COMP_ID_CAMERA6)) {
+    if(message.sysid == _vehicle->id() && (message.compid == MAV_COMP_ID_AUTOPILOT1 ||
+        (message.compid >= MAV_COMP_ID_CAMERA && message.compid <= MAV_COMP_ID_CAMERA6))) {
```
hamishwillee, line 82: Technically the component ID for a mavlink camera can be ANYTHING except 0 (broadcast) and 1-6 (autopilot connected). We shouldn't be restricting against the MAV_COMP_ID_CAMERA - MAV_COMP_ID_CAMERA6 but instead against a MAV_TYPE of https://mavlink.io/en/messages/minimal.html#MAV_TYPE_CAMERA

I appreciate that is harder since you'd have to maintain the type vs component mappings somewhere, but if we were never going to do that, why did everyone (other than me) agree using the MAV_TYPE made sense?

If we are using this id (1-6) for anything, such as sending commands to a specific camera, then we presumably also need to make sure that any messages to that camera are either broadcast or addressed to the autopilot, and that the id of the message (if present) is also set.

Just a brain dump, you surely understand all  this better than me.

julianoes: Yes, correct, we could not filter it at all. It would mean traffic is increased to other components that unlikely have cameras attached.

MAV_TYPE would be nice but we only have one heartbeat sent out by the autopilot. Otherwise we'd have to send out duplicate heartbeats which would be an alternative to this implementation but it's unlikely to gain traction.

Those messages would just be addressed to the autopilot. The autopilot will then do whatever proxying is required, I think.

HTRamsey:
Looks like this was merged already

PR #11575, opened 2024-05-30
QtLocationPlugin: Remove Unused Geocoding

HTRamsey:
Remove Unused Geocoding manager, as I don't believe it has ever been used. Removed old Qt license as it has changed.


HTRamsey:
@bkueng This one is pretty straightforward, these files were written like 9 years ago and were never used.

src/QtLocationPlugin/QGeoServiceProviderPluginQGC.cpp
```diff
@@ -1,96 +1,58 @@
...
 #endif
-//-----------------------------------------------------------------------------
-QGeoCodingManagerEngine*
-QGeoServiceProviderFactoryQGC::createGeocodingManagerEngine(
+
+QGeoCodingManagerEngine* QGeoServiceProviderFactoryQGC::createGeocodingManagerEngine(
     const QVariantMap &parameters, QGeoServiceProvider::Error *error, QString *errorString) const
 {
-    return new QGeoCodingManagerEngineQGC(parameters, error, errorString);
+    *error = QGeoServiceProvider::NotSupportedError;
```
bkueng, line 23: According to the [docs](https://doc.qt.io/qt-6/qgeoserviceproviderfactory.html#createGeocodingManagerEngine), `error` might be nullptr

bkueng (commented):
Nice, removing code

src/QtLocationPlugin/QGeoServiceProviderPluginQGC.cpp
HTRamsey: Huh, wonder if they never updated it from Qt5. I checked the QtLocation code and it's impossible to be nullptr but I can add checks anyways

bkueng (approved):

HTRamsey:
sorry just converted my tabs to spaces

Issue #11428, opened 2024-05-01
Restructure Vehicle to Not be a Fact Group


HTRamsey:
Kinda seems odd the vehicle class itself inherits a fact group. Move those facts into their own fact groups.

DonLakeFlyer:
I think that would add an additional level of hierarchy to the object model when referenced from script. vehicle.foo would become vehicle.bar.foo. Which will in turn break folks who have customized the telemetry display. If I remember correctly that stores property names with the full object hierarchy in settings. I'm not 100% sure of that since I haven't been in there in a while.

DonLakeFlyer:
That's right I forgot all those properties were defined in there anyway.

PR #11577, opened 2024-05-30
Vehicle: Create Fact Group

HTRamsey:
Aggregates remaining Vehicle Facts into their own fact group and makes Vehicle inherit from that factgroup rather than a generic fact group. This should result in the same exact Vehicle class, just with a pseudo separate fact group. Helps clean up Vehicle class a bit. Smaller & easier approach than #11428.

A couple of message handling functions are left in Vehicle for now because they use vehicle signals & private variables, and will require a bit more refactoring.

### DIFF
--- a/src/Vehicle/FactGroups/VehicleFactGroup.cc
+++ b/src/Vehicle/FactGroups/VehicleFactGroup.cc
@@ -1,0 +1,241 @@
+/****************************************************************************
+ *
+ * (c) 2009-2020 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ *
+ * QGroundControl is licensed according to the terms in the file
+ * COPYING.md in the root of the source code directory.
+ *
+ ****************************************************************************/
+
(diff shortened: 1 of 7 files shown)


src/Vehicle/FactGroups/VehicleFactGroup.cc
```diff
@@ -0,0 +1,248 @@
...
+    _hobbsFact.setRawValue(QVariant(QString("0000:00:00")));
+}
+
+void VehicleFactGroup::handleMessage(Vehicle* vehicle, mavlink_message_t& message)
+{
+    switch (message.msgid) {
+        case MAVLINK_MSG_ID_ATTITUDE:
+            _handleAttitude(vehicle, message);
+            break;
+
```
DonLakeFlyer, line 88: Standard coding style doesn't have blank lines here. The blank lines just add too much extra air in the source for readability. Also we follow Qt code style which lines up ```case``` at same indent as ```switch```. Although that last thing is not super important.

PR #11578, opened 2024-05-30
Vehicle: Add Change Heading

HTRamsey:
Adds ability to change heading towards location on map, currently only for ardupilot (sub/copter)
image

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -83,8 +83,11 @@ bool APMFirmwarePlugin::isCapable(const Vehicle* vehicle, FirmwareCapabilities c
     if (vehicle->multiRotor()) {
         available |= TakeoffVehicleCapability;
         available |= ROIModeCapability;
+        available |= ChangeHeadingCapability;
     } else if (vehicle->vtol()) {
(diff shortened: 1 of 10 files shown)


HTRamsey:
@DonLakeFlyer What's the best way to get this to only work for copter & sub on ardupilot?

DonLakeFlyer:
If you look at firmware plugin you'll see capability bits. Add a new one for change heading. ArduPilot sets it, PX4 does not. The add a new virtual for the change heading implementation. Plumb that through to Vehicle. Plumb through an changedHeading supported through vehicle as well. Look at how ROI mode does this for an example. Replicate that.

DonLakeFlyer:
Can you make this work for both ArduPilot and PX4? Mavlink command for both are here:
https://github.com/DonLakeFlyer/UAV-RT-TagTracker/blob/master/custom/src/CustomPlugin.cc#L657

Plumb difference through FirmwarePlugin and then you should need check for supported?

HTRamsey:
I just realized it isn't added to PX4FirmwarePlugin::isCapable

DonLakeFlyer:
> Also not sure if all vehicle's on px4 should be capable or not.

I would do the same copter/rover only thing. I've only used this myself on PX4 for multi rotor.

DonLakeFlyer (approved):
I tested this on PX4 SITL and it works

PR #11583, opened 2024-05-31
QtLocationPlugin: Make MapProvider Functions Const

HTRamsey:
This consolidates the MapProviders to have much less duplicated code and makes them entirely const.

This required removing the google version check and converting the Google Satellite Maps from the Google Maps to Google Quick Maps Services, which don't require version numbers as they default to the latest available. The only reason we might want to log the version number is to keep cached tiles and downloaded tiles on the same version, but currently all downloaded google tiles are stuck forever on an ancient version. The google version check can be re-added at a later dated if desired (maybe in QGeoTileFetcherQGC), but it serves little purpose currently.

The purpose in doing this is such that the entire UrlEngine can then become a static lookup class, without having to pull definitions from the mapengine instance. This should speed up live map functionality a bit and is safer memory wise.

Future option is to add a json file to be read from custom builds to easily add custom maps. Also it would be good to combine these definitions with the QGeoTiledMap class so each map type can easily provide copyrights, QGeoCameraCapabilities, etc.

I've tested on all the available maps that worked for me initially without API keys

### DIFF
--- a/src/QtLocationPlugin/MapProvider.cpp
+++ b/src/QtLocationPlugin/MapProvider.cpp
@@ -8,94 +8,103 @@
  ****************************************************************************/
 
 #include "MapProvider.h"
+#include <QGCLoggingCategory.h>
 
-#include <QtNetwork/QNetworkAccessManager>
+#include <QtCore/QLocale>
 #include <QtNetwork/QNetworkRequest>
 
-MapProvider::MapProvider(
-    const QString &referrer, 
-    const QString &imageFormat,
-    const quint32 averageSize, 
-    const QGeoMapType::MapStyle mapStyle, 
-    QObject* parent)
-    : QObject       (parent)
-    , _referrer     (referrer)
-    , _imageFormat  (imageFormat)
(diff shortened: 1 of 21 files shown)


src/QtLocationPlugin/MapProvider.cpp
```diff
@@ -8,94 +8,103 @@
...
+    if (image.startsWith(pngSignature)) {
+        return QStringLiteral("png");
+    }
+
+    static constexpr QByteArrayView jpegSignature("\xFF\xD8\xFF");
+    if (image.startsWith(jpegSignature)) {
+        return QStringLiteral("jpg");
+    }
+
+    static const QByteArrayView gifSignature("\x47\x49\x46\x38");
```
bkueng, line 61:
```suggestion
    static constexpr QByteArrayView gifSignature("\x47\x49\x46\x38");
```

bkueng (approved):
Nice, I've only seen a small inconsistency. Good to merge otherwise.

PR #11595, opened 2024-06-03
CMake: Fix Daily Build Version

### LINKED ISSUE #4882: Why are daily builds tagged as 3.1.2?
The latest [QGC release is 3.1.3](https://github.com/mavlink/qgroundcontrol/releases/tag/v3.1.3).  However, the [daily builds](https://donlakeflyer.gitbooks.io/qgroundcontrol-user-guide/content/download_and_install.html) are labeled 3.1.2. Why is that?

HTRamsey:
Master daily builds would show up as the previous tag.
Closes #4882

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,6 +302,12 @@ elseif(LINUX)
 
 endif()
 
+if(QGC_STABLE_BUILD)
+    set(APP_VERSION_STR ${REL_VERSION})
+else()
+    set(APP_VERSION_STR ${GIT_BRANCH}-${GIT_HASH})
+endif()
+
 add_compile_definitions(
     QGC_APPLICATION_NAME="QGroundControl"
     QGC_ORG_NAME="QGroundControl.org"
--- a/cmake/Git.cmake
+++ b/cmake/Git.cmake
@@ -22,12 +22,20 @@ endif()
 include(CMakePrintHelpers)
 
 execute_process(
-    COMMAND ${GIT_EXECUTABLE} describe --always --tags
+    COMMAND ${GIT_EXECUTABLE} rev-parse --abbrev-ref @
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
-    OUTPUT_VARIABLE APP_VERSION_STR
+    OUTPUT_VARIABLE GIT_BRANCH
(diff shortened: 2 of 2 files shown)


DonLakeFlyer:
This isn't pulling a version tag either

DonLakeFlyer:
``-- REL_VERSION="41d9164d668e795a9ae0605e422c6081d6be9f6b"''

This should be a version tag.

HTRamsey:
Is that what you get when you build it? I get something like `v4.3.0-605-gd49da8ea1`. Then this PR was just to adjust it such that the UI isn't showing v4.3.0 for daily builds after the v4.3.0 release. But thinking about it now, using CMAKE_PROJECT_VERSION might be better such that it shows v5.0.

DonLakeFlyer:
Right that is what you should get. I have this fixed in my cmake custom build pull. Look there for explanation. So this pull is not needed.

HTRamsey:
@DonLakeFlyer I don't think this is actually fixed yet. I still get this in master: 
image

DonLakeFlyer:
That is correct behavior. That ui shows the full git version which includes the commit. The version number from an OS standpoint should only be 4.3.0.

DonLakeFlyer:
The 4.3.0 tag is the last one that is in master. The daily just uses that. I was thinking it might be interesting to do some +1 math on the min version so that daily would be one rev higher. It didn't want to move to a fixed version number (5.0) in daily so that you don't have to keep updating it each time you make a new release.

HTRamsey:
Personally I think this PR is a better method and reduces confusion a bit, but it's up to you. My thoughts were that daily builds should just show branch-hash and stable builds would just show something like v4.4.0. I think the average user doesn't need to know the whole tag considering most of them are just regular users rather than programmers. Maybe the full tag could be somewhere a little more hidden like in the Help tab along with its associated github link or something

HTRamsey:
Makes sense, I do think it is slightly confusing when it appears that master is behind the latest stable but it's not a huge deal.

DonLakeFlyer:
That happened because the 4.4 release was created in a strange way. It was create from Stable 4.3 instead of master. Hence the missing 4.4 tag in master. That isn't the normal way things are done.

PR #11596, opened 2024-06-03
MAVLink: Add Signing

### LINKED ISSUE #8958: Add Support for MAVLink2 Message Signing
- What problem does it solve?
Integrity checks and authentication of GCS and Vehicle prevent hijacking of a vehicle via MAVLink telemetry radio.

- What does it do?
Prevent rogue ground stations from forcing a vehicle to land or upload malicious missions etc.

All

All

MAVLink2 contains this feature (https://mavlink.io/en/guide/message_signing.html) and Mission Planner seems to support it.

MAVLink2

HTRamsey:
Todo: Probably should disable the enable switch until a valid key is entered? 
Closes #8958

image

image

### DIFF
--- a/qgroundcontrol.pro
+++ b/qgroundcontrol.pro
@@ -619,6 +619,7 @@ HEADERS += \
     src/MAVLink/QGCMAVLink.h \
     src/MAVLink/MAVLinkLib.h \
     src/MAVLink/MAVLinkFTP.h \
+    src/MAVLink/MAVLinkSigning.h \
     src/MAVLink/StatusTextHandler.h \
     src/Comms/TCPLink.h \
     src/Comms/UDPLink.h \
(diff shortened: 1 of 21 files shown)


HTRamsey:
@DonLakeFlyer Should I go ahead and add a way to control this in the UI or is that something you want to do later?

DonLakeFlyer:
> Should I go ahead and add a way to control this in the UI

Sure, go ahead. Just make it functional. And then if needed I can clean up visuals later so you don't need to deal with that stuff.

HTRamsey:
@DonLakeFlyer Should this be enabled globally or per comm link?

DonLakeFlyer:
So forwarding is disabled when you are using signing. Seems like there should also be some indicator in the forward setting to indicate that it won't work.

HTRamsey:
I believe it's just the setup_signing message that isn't supposed to be forwarded, as per the documentation. I think the key could already be preconfigured on the forwarded device though, so signing should still work. I'll have to double check

DonLakeFlyer (approved):

PR #11597, opened 2024-06-03
Replace RunGuard with SingleApplication

HTRamsey:
RunGuard is no longer safe to use as of Qt6.6. See [this](https://lists.qt-project.org/pipermail/development/2023-November/044680.html). 
There are several different existing alternative solutions of SingleApplication out there with one being [this](https://github.com/itay-grudev/SingleApplication) and another being [this](https://github.com/KDAB/KDSingleApplication). The first one (which is used in this PR) will handle the issue noted in the Qt link automatically and provides other potentially useful capabilities for the future.


DonLakeFlyer (approved):

HTRamsey:
For me it works fine until I add the QT_DEBUG_PLUGINS=1 part. If there's not a simple solution then perhaps we can just update RunGuard in a way that doesn't use QSharedMemory and QSystemSemaphore

HTRamsey:
I just checked and it happens during closing of the app due to the AppMessages Q_GLOBAL_STATIC still writing debugging messages (like unloading plugins since you have QT_DEBUG_PLUGINS enabled) after the application instance is deleted.

HTRamsey:
@DonLakeFlyer Can you check if you get the same issue? Perhaps we can disable logging output when the application is destroyed or maybe just update RunGuard in a way that doesn't use the deprecated variable types listed in that Qt message

HTRamsey:
I noticed it also doesn't open if the app doesn't close cleanly. Probably a better solution is needed

Issue #11598, opened 2024-06-03
Link Interface Threading is Incorrect


HTRamsey:
The QThread subclassing in the LinkInterfaces is incorrect and is probably causing various difficult to trace issues. ADSBTCPLink as well. None of these should call moveToThread(this). A worker class or QtConcurrent::run can be used for the MockLink, while the rest are fine with signals/slots as they use asynchronous functions (except for one in TCPLink).

Also Avoid shared_ptr with QObjects.

PR #11599, opened 2024-06-04
ADSB: Convert Threading to Signals/Slots

HTRamsey:
Related to #11598. Converts the QThread subclassing to use timed signal/slot processing.

### DIFF
--- a/src/ADSB/ADSBTCPLink.cc
+++ b/src/ADSB/ADSBTCPLink.cc
@@ -8,165 +8,217 @@
  ****************************************************************************/
 
 #include "ADSBTCPLink.h"
-#include "QGCLoggingCategory.h"
+#include <QGCLoggingCategory.h>
 
 #include <QtNetwork/QTcpSocket>
-#include <QtPositioning/QGeoCoordinate>
+#include <QtCore/QTimer>
 
 QGC_LOGGING_CATEGORY(ADSBTCPLinkLog, "qgc.adsb.adsbtcplink")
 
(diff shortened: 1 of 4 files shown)


src/ADSB/ADSBTCPLink.cc
```diff
@@ -8,165 +8,217 @@
...
+        return;
+    }
 
-    QString altitudeStr = values[11];
     // Altitude is either Barometric - based on pressure, in ft
     // or HAE - as reported by GPS - based on WGS84 Ellipsoid, in ft
     // If altitude ends with H, we have HAE
     // There's a slight difference between Barometric alt and HAE, but it would require
     // knowledge about Geoid shape in particular Lat, Lon. It's not worth complicating the code
+    QString altitudeStr = values.at(11);
```
DonLakeFlyer, line 178: Curious as to what the difference is here?

HTRamsey: When debugging it uses a Q_ASSERT to verify a valid index, otherwise it is exactly the same. Just an easy way to catch errors

DonLakeFlyer: So does `[]` operator.

DonLakeFlyer:
FYI: I verified that QList operator [] uses an assert on index reference as well.

HTRamsey:
Oh okay I guess it is the same for Qt classes but not std classes I think. The other reason I used it was to guarantee a const reference which avoids a deep copy of the list. Can also be nice to do something like myList->at(x) rather than *myList[x].

DonLakeFlyer (approved):

PR #11600, opened 2024-06-04
CMake: Add LGPL Option

### LINKED ISSUE #8252: Allow building QGroundControl with only LGPL Qt modules
According to [Qt Licensing](https://doc-snapshots.qt.io/qt5-5.12/licensing.html) and in particular [this list](https://doc-snapshots.qt.io/qt5-5.12/qtmodules.html#gpl-licensed-addons), most of Qt is (also) released under the LGPL license except for a few modules that are GPL / commercial only.

QGroundControl uses mostly LGPL modules. At the moment only MAVLink inspector uses the charts module which is GPL / commercial.

What I am asking for is a CONFIG option or something similar to disable building code that uses non-LGPL modules. This will require changes to qgroundcontrol.pro. Something like this works for the current code base (define QGC_ENABLE_QT_LGPL in user_config.pri or on the command line):

    #-------------------------------------------------------------------------------------
    # MAVLink Inspector
    contains (DEFINES, QGC_ENABLE_MAVLINK_INSPECTOR && !QGC_ENABLE_QT_LGPL) {

Another option would be something like this (easier to understand but more verbose):

    # Enable usage of LGPL-only Qt modules
    contains(CONFIG, QGC_ENABLE_QT_LGPL) {
        # Disable MAVLink inspector as it uses the charts module
        contains (DEFINES, QGC_ENABLE_MAVLINK_INSPECTOR) {
            DEFINES -= QGC_ENABLE_MAVLINK_INSPECTOR
        }
    }

You might remember I started a [discussion](https://discuss.px4.io/t/licensing-question-qgc-qt-open-source/13889) on this.

Thanks!

HTRamsey:
Closes #8252


DonLakeFlyer (commented):
If this removes the usage of Charts then the PX4 tuning page also uses charts. Can't remember if it PX4 only. Might be ArduPilot tuning as well.

DonLakeFlyer:
Also this is interesting, but really wonder whether the additional complexity is needed?

HTRamsey:
I'll have to double check the tuning pages. I don't think it'll really be all that extra complex but we can see what it looks like after we're done with all the other custom cmake stuff. In my opinion this is pretty useful as I have worked with a few companies that won't use anything but LGPL

s-lisovenko:
Agree with @HTRamsey. I also know several companies that are just thinking about abandoning QGC in favor of using only lgpl solutions.

DonLakeFlyer:
> I also know several companies that are just thinking about abandoning QGC in favor of using only lgpl solutions.

If any of those companies actually ever actually gave back to the ecosystem then I'd certainly be more interested :)

s-lisovenko:
> > I also know several companies that are just thinking about abandoning QGC in favor of using only lgpl solutions.
> 
> If any of those companies actually ever actually gave back to the ecosystem then I'd certainly be more interested :)

So far these are startups and one even makes a contribution to qgc in the person of me. But no one can move on with gpl. And the prices for commercial use of qt are simply inadequate, especially since it is also subject to the qt gender license for devices, not to mention the fact that qt prohibits mixing a commercial license with an open sources.

DonLakeFlyer:
Yeah, the Qt pricing is really kinda crap.

PR #11601, opened 2024-06-05
QtLocationPlugin: Make UrlEngine Static

HTRamsey:
- Adds the provider name and map id to the MapProvider class and removes their QObject inheritance.
- Completely converts the UrlEngine to a static class.
- Re-adds the esri maps to the list as previously they never got added after entering the token despite the comment.

### DIFF
--- a/src/QtLocationPlugin/ElevationMapProvider.h
+++ b/src/QtLocationPlugin/ElevationMapProvider.h
@@ -2,16 +2,19 @@
 
 #include "MapProvider.h"
 
-static constexpr const quint32 AVERAGE_AIRMAP_ELEV_SIZE = 2786;
+static constexpr const quint32 AVERAGE_COPERNICUS_ELEV_SIZE = 2786;
 
 class ElevationProvider : public MapProvider
 {
-    Q_OBJECT
-
 protected:
-    ElevationProvider(const QString& imageFormat, quint32 averageSize,
-                      QGeoMapType::MapStyle mapType, QObject* parent = nullptr)
-        : MapProvider(QStringLiteral("https://terrain-ce.suite.auterion.com/"), imageFormat, averageSize, mapType, parent) {}
+    ElevationProvider(const QString &mapName, const QString &referrer, const QString &imageFormat, quint32 averageSize,
+                      QGeoMapType::MapStyle mapType)
+        : MapProvider(
+            mapName,
+            referrer,
+            imageFormat,
+            averageSize,
+            mapType) {}
 
 public:
     bool isElevationProvider() const final { return true; }
@@ -19,12 +22,14 @@ public:
 
 class CopernicusElevationProvider : public ElevationProvider
 {
-    Q_OBJECT
-
 public:
-    CopernicusElevationProvider(QObject* parent = nullptr)
-        : ElevationProvider(QStringLiteral("bin"), AVERAGE_AIRMAP_ELEV_SIZE,
-                            QGeoMapType::StreetMap, parent) {}
+    CopernicusElevationProvider()
+        : ElevationProvider(
+            QStringLiteral("Copernicus Elevation"),
+            QStringLiteral("https://terrain-ce.suite.auterion.com/"),
+            QStringLiteral("bin"),
+            AVERAGE_COPERNICUS_ELEV_SIZE,
+            QGeoMapType::StreetMap) {}
 
     int long2tileX(double lon, int z) const final;
     int lat2tileY(double lat, int z) const final;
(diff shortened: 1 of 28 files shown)


src/QtLocationPlugin/ElevationMapProvider.h
```diff
@@ -6,25 +6,29 @@ static constexpr const quint32 AVERAGE_AIRMAP_ELEV_SIZE = 2786;
...
-
 protected:
-    ElevationProvider(const QString& imageFormat, quint32 averageSize,
-                      QGeoMapType::MapStyle mapType, QObject* parent = nullptr)
-        : MapProvider(QStringLiteral("https://terrain-ce.suite.auterion.com/"), imageFormat, averageSize, mapType, parent) {}
+    ElevationProvider(const QString &mapName, const QString &imageFormat, quint32 averageSize,
+                      QGeoMapType::MapStyle mapType)
+        : MapProvider(
+            mapName,
+            QStringLiteral("https://terrain-ce.suite.auterion.com/"),
```
bkueng, line 14: This was already the case, but since you're changing it, I would pass the url from `CopernicusElevationProvider`

src/QtLocationPlugin/MapProvider.h
```diff
@@ -40,16 +40,19 @@ class QNetworkRequest;
 
 class MapProvider : public QObject
```
bkueng, line 41: Did you have a reason to keep this?
```suggestion
class MapProvider
```

src/QtLocationPlugin/MapProvider.h
```diff
@@ -40,16 +40,19 @@ class QNetworkRequest;
...
+                QGeoMapType::MapStyle mapStyle = QGeoMapType::CustomMap);
+    virtual ~MapProvider();
 
     QString getImageFormat(QByteArrayView image) const;
 
+    // TODO: Download Random Tile And Use That Size Instead?
     quint32 getAverageSize() const { return _averageSize; }
+
     QGeoMapType::MapStyle getMapStyle() const { return _mapStyle; }
+    QString getMapName() const { return _mapName; }
```
bkueng, line 54:
```suggestion
    const QString& getMapName() const { return _mapName; }
```

src/QtLocationPlugin/QGCMapUrlEngine.h
```diff
@@ -17,54 +17,47 @@
...
 
 #define MAX_MAP_ZOOM (23.0)
 
-class QNetworkAccessManager;
 class MapProvider;
 
-class UrlFactory : public QObject {
-    Q_OBJECT
-
+class UrlFactory : public QObject
```
bkueng, line 29: Also here
```suggestion
class UrlFactory
```

src/QtLocationPlugin/MapProvider.cpp
```diff
@@ -15,18 +15,29 @@
 
 QGC_LOGGING_CATEGORY(MapProviderLog, "qgc.qtlocationplugin.mapprovider")
 
+int MapProvider::_mapIdIndex = 1;
```
bkueng, line 18: Might be worth addiing a comment why we start at 1 (if that's important)

src/QtLocationPlugin/QGCMapUrlEngine.h
```diff
@@ -17,54 +17,47 @@
...
-    int             hashFromProviderType            (const QString& type);
-    QString         providerTypeFromHash            (int tileHash);
-
-    QGCTileSet getTileCount(int zoom, double topleftLon, double topleftLat,
+    static QGCTileSet getTileCount(int zoom, double topleftLon, double topleftLat,
                             double bottomRightLon, double bottomRightLat,
-                            const QString& mapType);
+                            QStringView mapType);
+
+    static QList<std::shared_ptr<const MapProvider>> getProviders() { return _providers; }
```
bkueng, line 49:
```suggestion
    static const QList<std::shared_ptr<const MapProvider>>& getProviders() { return _providers; }
```

src/QtLocationPlugin/MapProvider.h
HTRamsey: Isn't this no longer necessary as of C++17?

src/QtLocationPlugin/MapProvider.h
HTRamsey: Only reason being I forgot to remove it

src/QtLocationPlugin/MapProvider.h
bkueng: My understanding is that copy elision would not be applied here as it's not returning a temporary.
And inlining would optimize some cases, but not generally all. E.g. `QString mapName = x.getMapName();`. You could argue that you should use `const QString& mapName = x.getMapName();` in that case too, where I think the compiler would remove all copies, but I'm not sure if it's guaranteed. Maybe you're more familiar with the details here?

HTRamsey:
@bkueng I ended up having to make a few small adjustments to figure out why the test cleanup wasn't working well, which was mostly just setting some functions in QGCMapEngine to static and correcting the parenting of the QGCCacheWorker so that the thread quits correctly.

bkueng (approved):

DonLakeFlyer (approved):

PR #11606, opened 2024-06-06
APM: Add Rover Modes

### LINKED ISSUE #10670: Yet no "Dock" mode available in flight mode selection menu
mode "8" = DOCK mode is still missing, which is already implemented in ardupilot

HTRamsey:
Closes #10670. I really feel like these can be auto configured from the INITIAL_MODE parameter or something but I guess they don't change often enough for it to matter


DonLakeFlyer (approved):

DonLakeFlyer:
I think there is new mavlink spec for getting flight modes from the vehicle. PX4 implements I think which gets rid of all these constant updates.

Issue #11436, opened 2024-05-01
Custom Build Implementation for CMake


HTRamsey:
Custom Builds compatibility needs to be updated for the new CMake build system.

DonLakeFlyer:
Things that are a pain with respect to current way that custom builds work:
* The fact that the custom code and the upstream qgc code are in the same repo causes all sorts of nightmares with respect to bringing upstream qgc up to date in your custom build repro. You spend tons of time resolving merge conflicts and what not on the upstream pull.
* The process of forking the regular qgc repo and then renaming custom_example and go from there means that your starting point for custom files are forever linked back to the original upstream ones. Which in turn means as your custom build gets more complex, you run into tons of merge conflicts when updating to a new upstream qgc.

I think that a mechanism where the custom build source is disconnected from the upstream qgc source would be way better. The process then being:
* Create a brand new empty repo for you custom build
* Copy whatever example custom build stuff you want as a starting point into there
* Somehow reference upstream qgc in your build more like a submodule type thing

This way:
*  upstream qgc is disconnected from a git pull standpoint. If you have a clean custom build implementation which has no upstream qgc changes. Then you just update to newest upstream qgc. No merge conflict hell.
* Your custom code is not connected to the initial example code. So again no merge conflict hell as the example custom build goes through modifications.

mrpollo:
Thanks, @DonLakeFlyer. I have heard from devs maintaining custom forks that things could be better, which is also one of the reasons the project sees few contributions. When someone writes fixes/features downstream, pushing them upstream is hard when the delta is too high, especially when downstream stays the same for a while.

Testing coverage for this effort would likely be required. We need to offer a way for folks to test upstream changes against their custom builds without completely derailing their development efforts.

One way to make this a successful migration would be to have the QGC App itself be a third-party app using the same system we are putting in place. What I mean by QGC App is the app you download from the app stores.

DonLakeFlyer:
> When someone writes fixes/features downstream, pushing them upstream is hard when the delta is too high, especially when downstream stays the same for a while.

[...]

PR #11608, opened 2024-06-09
Custom Build Implementation for CMake (#11436)

garrik:
I wrote the cmake implementation for the custom build as close as possibile to the previous one.
I will be glad to contribute with fix or improvents if necessary.

Related Issue
-----------
#11436

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -181,7 +181,10 @@ if(QGC_BUILD_TESTING)
     add_compile_definitions(UNITTEST_BUILD) # TODO: QGC_UNITTEST_BUILD
 endif()
 
-# option(QGC_CUSTOM_BUILD "Enable Custom Build" OFF)
+option(QGC_CUSTOM_BUILD "Enable Custom Build" OFF)
+if(QGC_CUSTOM_BUILD)
+    add_subdirectory(custom)
+endif()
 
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
Unless I'm reading something wrong I don't see how this works from a standpoint of automatically setting QGC_CUSTOM_BUILD if the custom directory exists like the old system did. What are the expectations as to how you kick off a custom build?

DonLakeFlyer:
I'm going to play around with this a bit to see what matches previous setup a bit more closely. @garrik Thanks so much for putting something together on this.

DonLakeFlyer:
@garrik Can you look at #11609?

garrik:
The custom build is enabled by moving/copying the custom-example folder to custom folder and setting QGC_CUSTOM_BUILD to ON. This is based on what I get from the documentation and from comments on the issue but I may have missed something since I have no previous experience in building QGC with qmake and `qgroundcontrol.pro` file.

garrik:
> @garrik Can you look at #11609?

Looked at it. It streamlines the process of kicking off the custom build and makes code more straightforward. Thank you.

DonLakeFlyer:
@HTRamsey You know more about cmake than I do. I don't see a way to do this without this two step process. Maybe there is some other magic way to make it work?

Issue #11303, opened 2024-04-06
Null Property Errors in Comm Links when Canceling Adding a New Link 


HTRamsey:
![image](https://github.com/mavlink/qgroundcontrol/assets/68555040/e8320b63-e26f-45a1-a75f-25d63929ee7e)

PR #11609, opened 2024-06-11
Custom build cmake support

DonLakeFlyer:
This is derived from #11608. It differs from that version in two ways:
* It automatically detects custom builds by the existence of the "custom" directory
* It provides a two step override mechanism for things like application name such that things like a custom application name can flow through the rest of cmake
  * It does this by including a CustomBuild.cmake file at top of the main cmake file. This file can then override various variables which are needed further down the main cmake file. For example the modified application name is needed by the OSX bundle cmake setup further down the main file.
  * And the further down the main cmake file it adds the custom directory to cmake

It's not finished yet, but I wanted to get comments on this whole two step process thing. Maybe there is a better way to do it.

This pull fixes https://github.com/mavlink/qgroundcontrol/issues/4882 as well


garrik:
As a side note, all variables in `custom/CMakeLists.txt` that are set in PARENT_SCOPE could be moved into the `custom/CustomBuild.cmake` and set normally. The same applies to the `set_property(DIRECTORY ${CMAKE_SOURCE_DIR} APPEND PROPERTY COMPILE_DEFINITIONS ...)` could be moved and reverted to `add_compile_definitions()`.
This would split apart even more the settings for the custom build, but I think it is not a big deal.
I have not a strong and clear opinion on having one or two location to look for custom build editing.

garrik:
Done this way, the `custom\CustomBuild.cmake` file is the place to look at changes related to the custom build that impact the application (QGroundControl scope).
The file `custom\CMakeLists.txt` is the place to look at changes related to the custom build implementation (custom scope).

Davidsastresas:
I am not as familiarized with cmake as I was with qmake, so I can not comment a lot. But those comments about propagating custom app name, etc seem like an improvement versus the old qmake based one, so it sounds great.

A few weeks from now I might bring a custom build to master to experience all of this first hand and hopefully provide more feedback. Thank you!

AndrewPix:
I tried to make a custom build using both this PR and @garrik 's, and I can not connect serial device:
```
qrc:/qml/LinkSettings.qml:215: TypeError: Cannot read property 'settingsURL' of null
qrc:/qml/SerialSettings.qml:169: TypeError: Cannot read property 'stopBits' of null
qrc:/qml/SerialSettings.qml:161: TypeError: Cannot read property 'dataBits' of null
qrc:/qml/SerialSettings.qml:115: TypeError: Cannot read property 'flowControl' of null
```

Any advices how to fix that? This only happens on the custom build.

HTRamsey:
@AndrewPix that should be unrelated. Dunno why it happens only with custom build, see #11303.

HTRamsey:
@DonLakeFlyer I think the overall approach is good. There are some things that I would change such as adjusting the Git part to use our Git.cmake file instead. I also had imagined the custom builds to be triggered by a cmake option, like @garrik did, (which would then check existance of custom folder) but that's a very minor implementation detail. 

Also, I think the whole python qrc file thing will eventually be cleanly replaced by qt_add_resources & qt_add_qml_module but for now getting the bare minimum working as similar to the way it is via qmake is great.

AndrewPix:
> @AndrewPix that should be unrelated. Dunno why it happens only with custom build, see #11303.

I can not add a new serial link in custom build which is work fine in regular build.

DonLakeFlyer:
@HTRamsey Any ideas as to why the git describe for version info isn't working? It doesn't seem like the clone has tags yet fetch-tags = true?

DonLakeFlyer:
So I've made changes and tested OSX custom builds and they are working locally for me. Now need to move to other OS.

HTRamsey:
@DonLakeFlyer see #11595 regarding the Git stuff.

DonLakeFlyer:
It seems like something changed in the repo. I reverted back to the way Stable does it with checkout and tag fetch. But the git describe --always -- tags returns a hash instead of a version tag. Sigh... No idea what going on there. 

If I clone a new qgc at this pull, git describe works and gives me a version tag. But if I do the tag fetch then it breaks and gives me a has.

DonLakeFlyer:
```
      - name: Checkout repo
        uses: actions/checkout@v4
        with:
          submodules: recursive
          fetch-tags: true
          fetch-depth: 0
```

We used to be able to do a shallow fetch and the tag thing worked. Not anymore for some reason. Need ` fetch-depth: 0` to make it work now.

DonLakeFlyer:
This pull fixes #4882 as well

HTRamsey:
> This pull fixes #4882 as well

just fyi if you didn't know you can put that comment in the original PR description and it will auto close the issue when this PR is merged. A little summary of which issues are addressed shows up like in this image
image

DonLakeFlyer:
@HTRamsey What do you think about merging this in now. Only OSX is working, but it's gonna be a while until I can get to the other OSes. I'm running into problems since my dev platform is ARM and that just complicates everything: Linux, Windows, ... Then maybe other folks could help out with other OS following the pattern I took.

HTRamsey:
Sure, I can probably tackle the rest based on your work here. Considering this is the main blocker for moving forwards, I'd say let's move on with it.

PR #11613, opened 2024-06-12
Version release pages and point to Github releases page first

hamishwillee:
There doesn't seem to be any maintenance going on of the release notes pages. That's a problem because we're up to v4.3, and you'd thing that v4.0.0 was the last one.

What this does is aim to give people a clear view of where they can go for release information first (the github release page), without making it impossible to revert to supplying more detailed release notes in future.

In summary 
- At the top level it just points to the releases page on github
- it merges the minor release pages into major release pages for v3 and v4
- At the top of v4 it notes that there is more release info in the github release page. 

When we release v5 we will need to decide if we're going to provide proper release notes or not.

@DonLakeFlyer @julianoes What do you think? Julian, I didn't want to just move unilaterally to "only use the github release note".


julianoes:
Given we didn't update it since v4.0.0 I don't think we should try to keep it in sync, unless someone is keen on maintaining that.

DonLakeFlyer:
> unless someone is keen on maintaining that.

Which is the main problem. I gave up trying to spend my time keeping track of changes for releases notes. Without that, the whole thing isn't very useful.

hamishwillee:
@DonLakeFlyer @julianoes I can't argue that Don's time can be better spent. That said, I personally would like to know when getting the next stable release facts like "it now supports the MAVLink gimbal and camera protocol".

Probably what we need to do is agree what we expect in tag releases like https://github.com/mavlink/qgroundcontrol/releases/tag/v4.4.0 so that you could construct a high level feature-based release note if you wanted.

I'm going to merge this because it is better than what is there now.

@DonLakeFlyer Is there a plan for a new release against PX4 v1.15 which I "understand" is planned soon?

PR #11639, opened 2024-06-27
CMake: Build Fixes

HTRamsey:
Moving some fixes from custom build PR to this.

- Fix spacing formatting in all cmake files
- Fix disabling plugins
- Don't link to UI qml module yet
- Fix including sensors module
- Fix including GNUInstallDirs
- Minor improvement to root CMakeLists.txt formatting 
- Fix Android Builds


DonLakeFlyer (approved):
One thing which would be nice is to keep formatting changes in a single commit. That makes doing a review easier. Since I can just blow past the formatting stuff. If it's all mixed together then I need to poke through everything for real changes.

PR #11640, opened 2024-06-28
New Guided value slider

### LINKED ISSUE #10743: Guided Altitude Slider Linear Slider with Snapping
This is a continuation of a short discussion on Discord, and is my concept for an improvement to the guided altitude slider.
This is something I have modified in a custom build already, but I am making this issue to discuss whether it would benifit others and would be appropriate for integrating into master.

## Background
The current exponential altitude slider is designed to "allows for fine adjustment around the current altitude, but also allows the upper/lower end of the range to hit a full range adjustment" see #5150 for background.

## The issues with the current design
My suggestions are based on my experience of deploying a fleet of multirotors flown exclusively with guided mode through qgroundcontrol (On a 10" tablet). The drones are flown between 0 and 120m. We have found the following issues with the current behavior:
- The exponential slider makes it difficult to precisely make large altitude changes, as the resolution decreases away from center. While it is very easy to "go up 1.5m" it is hard to precisely "go to 80m" which at least in our case is a more common scenario.
- The non linearity makes for a less predictable experience and requires the user to concentrate quite hard to dial in specific altitudes.
- The change altitude slider behaves differently from the takeoff slider

## Solution Mockup
In my custom build, I have written the following: The slider snaps to the tickmarks and the red section shows the minimum guided height, below which you cant drag the slider. I find it works quite well. This is similar to what was suggested in #5795.
image

## When would this not work?
While I appreciate that there may be scenarios in which the exponential slider is a better option for this, I believe they are the minority of use cases. I believe the solution I am proposing would not work when:
- High altitude precision is required - eg the ability to go to 43m exactly, rather than being content with coarser altitude steps
- A very large altitude range is in use, beyond the normal 0-120m
- The vehicle needs to fly below the home point

## My suggestion for integration
- Under the guided command settings section, a drop down to select linear vs exponential slider (default whatever is deemed appropriate for most users)
- a tickbox to enable snapping on the linear slider if desired (default enabled)
- a setting for slider snap resolution, ie a setting of 10 will snap to the nearest of 10 values. We can setup the defaults for this setting to be low for mobile builds, and high for desktop builds. eg a value of 16 is what is used in my example above and is practical on desktop/laptop, but on a phone a value of 8 eg 50ft resolution would probably be more usable. (we could also limit the maximum allowed value depending on platform)

DonLakeFlyer:
Fixes #10743

See the issue for details

### DIFF
--- a/src/Comms/MockLink/MockLink.cc
+++ b/src/Comms/MockLink/MockLink.cc
@@ -32,11 +32,11 @@ QGC_LOGGING_CATEGORY(MockLinkVerboseLog, "MockLinkVerboseLog")
 #if 1
 double      MockLink::_defaultVehicleLatitude =     47.397;
 double      MockLink::_defaultVehicleLongitude =    8.5455;
-double      MockLink::_defaultVehicleAltitude =     488.056;
+double      MockLink::_defaultVehicleHomeAltitude = 488.056;
 #else
 double      MockLink::_defaultVehicleLatitude =     47.6333022928789;
 double      MockLink::_defaultVehicleLongitude =    -122.08833157994995;
-double      MockLink::_defaultVehicleAltitude =     19.0;
+double      MockLink::_defaultVehicleHomeAltitude = 19.0;
 #endif
 int         MockLink::_nextVehicleSystemId =        128;
 const char* MockLink::_failParam =                  "COM_FLTMODE6";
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
@AndKe Can you test with Ardupilot SITL? @Davidsastresas Can you test your speed change stuff?

DonLakeFlyer:
Appveyor? Something change I don't know about?

HTRamsey:
@DonLakeFlyer it just means it was out of date with master. I turned on the button to be able to rebase via a button in the PR  webpage (was getting annoying having to do it locally and push every time) which causes it to not build unless it is in sync with master. Actually kinda nice in my opinion because it ensures a PR will commit correctly before doing it, but we can disable it if you want

HTRamsey:
I am not sure why it is not starting the builds now

PR #11641, opened 2024-06-28
Guided value slider

### LINKED ISSUE #10743: Guided Altitude Slider Linear Slider with Snapping

DonLakeFlyer:
Fixes #10743

See Issue for details

### DIFF
--- a/src/Comms/MockLink/MockLink.cc
+++ b/src/Comms/MockLink/MockLink.cc
@@ -1337,7 +1337,6 @@ void MockLink::_sendGlobalPositionInt(void)
 
 void MockLink::_sendExtendedSysState(void)
 {
-    static uint64_t timeTick = 0;
     mavlink_message_t msg;
 
     mavlink_msg_extended_sys_state_pack_chan(_vehicleSystemId,
--- a/src/Comms/MockLink/MockLink.cc
+++ b/src/Comms/MockLink/MockLink.cc
@@ -32,11 +32,11 @@ QGC_LOGGING_CATEGORY(MockLinkVerboseLog, "MockLinkVerboseLog")
 #if 1
 double      MockLink::_defaultVehicleLatitude =     47.397;
 double      MockLink::_defaultVehicleLongitude =    8.5455;
-double      MockLink::_defaultVehicleAltitude =     488.056;
+double      MockLink::_defaultVehicleHomeAltitude = 488.056;
 #else
 double      MockLink::_defaultVehicleLatitude =     47.6333022928789;
 double      MockLink::_defaultVehicleLongitude =    -122.08833157994995;
-double      MockLink::_defaultVehicleAltitude =     19.0;
+double      MockLink::_defaultVehicleHomeAltitude = 19.0;
 #endif
 int         MockLink::_nextVehicleSystemId =        128;
 
@@ -55,7 +55,7 @@ MockLink::MockLink(SharedLinkConfigurationPtr& config)
     , _mavState                             (MAV_STATE_STANDBY)
     , _firmwareType                         (MAV_AUTOPILOT_PX4)
     , _vehicleType                          (MAV_TYPE_QUADROTOR)
-    , _vehicleAltitude                      (_defaultVehicleAltitude)
+    , _vehicleAltitudeAMSL                      (_defaultVehicleHomeAltitude)
     , _sendStatusText                       (false)
     , _apmSendHomePositionOnEmptyList       (false)
     , _failureMode                          (MockConfiguration::FailNone)
@@ -251,6 +251,8 @@ void MockLink::_run10HzTasks(void)
             _sendGPSPositionDelayCount--;
         } else {
             _sendGpsRawInt();
+            _sendGlobalPositionInt();
+            _sendExtendedSysState();
         }
     }
 }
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
I'm lost. Creating a new pull didn't make it go away as well.

HTRamsey:
can you rebase locally and see what happens?

DonLakeFlyer:
I already re-pulled the version you rebase tossing my old one

DonLakeFlyer:
Let me try making a whole new branch

HTRamsey:
I would drop my rebase and do a merge just to see if one of the older commits didn't rebase well

DonLakeFlyer:
Stupid https://github.com/appveyor/ci/issues/3585

HTRamsey:
So I have no idea why the android builds fails to find gstreamer sometimes, but rerunning it works I think, I will look at it tonight

HTRamsey:
That's one of the things I will add to #11639

arnaudjuin:
> That's one of the things I will add to #11639 Edit: Nvm a quick google search gave me the answer. I can fix later today

Any success or hint ?

HTRamsey:
> > That's one of the things I will add to #11639 Edit: Nvm a quick google search gave me the answer. I can fix later today
> 
> Any success or hint ? 
> 

Sorry I've been working 15hrs the past few days and ran out of time. I will be able to do it tomorrow at the latest

DonLakeFlyer:
@jmare Can you retest?

* Tick marks which are above/below allowed values are grey
* Indicator is always centered
* Fixed problem with setting below min alt

JMare:
Nice!
Away from my computer for a few days, will test on Monday

Davidsastresas:
I added a few commits addressing some issues testing with Ardupilot.

The only one I think should be discussed is the last commit, about changing altitudeAMSL to altitudeRelative. Before this one, the slider would represent the altitude AMSL, even though the label says "alt (rel)" and it always represented relative altitude. @DonLakeFlyer was this a mistake or this brokes PX4 functionality? From what I saw in the code the issue explained above should affect PX4 as well, but I could have missed something.

Other than that I think it should be fine now.

The new slider looks amazing by the way, great job, thank you!

DieBorr:
Hi, I have been taking a look to the slider, I really like it, seems really nice with the actual UI. However there is something that at least for me does not make sense and is that the horizontal flickable is visible when there is a width resize, I don't know if it has been made on purpose, I have a commit ready to push if you confirm that if it wasn't intentional. Correct me if I wrong but I can't figure out a case where the horizontal flickable needs to be visible in the slider.

image

DonLakeFlyer:
> horizontal flickable is visible when there is a width resize

Ah, right that is screwed up. It isn't tracking contentWidth correctly on the flickable

DieBorr:
https://github.com/mavlink/qgroundcontrol/pull/11663 solves overlap and makes some margin between text and slider border.

DonLakeFlyer:
I'm gonna merge this in and then fix the horizontal thing in another pull...

JMare:
@DonLakeFlyer, Tested, and fixed the issue I saw previously - looks good.

The only thought I wanted to raise - which isn't specifically related to this change is:
When doing a change alt, QGroundControl sends the new altitude as an offset from the current altitude.
This offset is calculate by taking the new desired altitude and subtracting the current altitude.

One consequence of that is that if the vehicle is climbing or descending while you do the altitude change, there is an inaccuracy.
For instance, in Ardupilot SITL, if you do a changealt to 25.0m while you are climbing, you may find the vehicle actually settle at 25.5m. I can imagine this would be worse with high climb rates or low telemetry rates.

Its a minor thing but do you know if there is any reason the normal relative altitude couldn't be sent rather than an offset?

DonLakeFlyer:
> Its a minor thing but do you know if there is any reason the normal relative altitude couldn't be sent rather than an offset?

If I remember correctly it's because PX4 firmware side does change alt as a delta offset. Let me look at the code again. Maybe better to send absolute all the way till it gets to mavlink and then adjust there. That way at least ArduPilot would get a good alt.

DonLakeFlyer:
Yeah, I looked at the code. I can reverse this so that an absolutely altitude flows through until the last minute instead of an offset. That way at least ArduPilot can get a solid alt change in the case described above.

JMare:
Nice yes I think that would be a good change, happy to help with testing. (or coding, but it sounds like you have it covered)

DonLakeFlyer:
I had this backwards. It's ArduPilot that needs the offset change not PX4. Let me see if there is some different command I could send to get an absolute altitude change.

PR #11654, opened 2024-07-04
Android: Create Serial Driver Interface

HTRamsey:
Create interface for accessing android java serial functions. The last step after this is to rebase QSerialPort class to the latest QtSerialPort version


DonLakeFlyer:
So what the overall summary to the changes here? Use latest USB support on the Java side of things? Then create a custom QSerialPort which calls into that? How does it differ from the old code?

HTRamsey:
So I had already updated the Java serial code previously, this just provides a direct cpp interface to the java code in AndroidSerial.cpp and fills in the remaining missing Java functions for QSerialPort cpp functions that weren't implemented (flow control, control lines, break, pinoutsignals, etc). The custom Android QSerialPort already existed, so this just cleans it up to use AndroidSerial.cpp and implements those remaining missing functions.

DonLakeFlyer (approved):

PR #11655, opened 2024-07-05
Comms: UDPLink Threading Changes

HTRamsey:
Related to #11598
Also improves zeroconf implementation debugging

Edit: Just realized an AirLink update will have to immediately follow this

### DIFF
--- a/src/Comms/CMakeLists.txt
+++ b/src/Comms/CMakeLists.txt
@@ -109,7 +109,7 @@ if(QGC_ZEROCONF_ENABLED)
     )
     FetchContent_MakeAvailable(qmdnsengine)
 
-    target_link_libraries(Comms PRIVATE qmdnsengine)
+    target_link_libraries(Comms PUBLIC qmdnsengine)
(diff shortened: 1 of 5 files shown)


dakejahl:
I'm getting spammed with this warning, seems to be introduced in this PR. I am testing an indoor quad on wifi, the link is probably pretty lossy. 
image

HTRamsey:
Gotcha, should probably just be a console warning instead

HTRamsey:
@dakejahl do you think #12267 will resolve it?

dakejahl:
@HTRamsey thanks for the quick fix! Yes I think that will do it, I had just commented those out for the time being and don't see any more pop ups

dakejahl:
Haha spoke too soon... got another one "unable to send a datagram"

HTRamsey:
That one is probably coming through "_onSocketErrorOccurred" I guess. Previously pretty much all warnings were just ignored so trying to find the right balance of what to just print to console vs show as popup

PR #11663, opened 2024-07-07
Remove horizontal flickable + UI resize

DieBorr:
Changed margins + horizontalFlickable deleted.


DonLakeFlyer:
This isn't correct. It fixes the symptom as opposed to the problem. The problem is that the content container and the flickable item are not correctly being kept in sync. If they were they would always have equal width which means a horizontal slider will never show up. I'll figure out the bug here.

PR #11668, opened 2024-07-08
Tranlsations: Bring up to date with Crowdin

DonLakeFlyer:
I did a full build on Crowdin and brought all files down to here. These are the only differences. So hopefully after this this will flow through smoothly.


HTRamsey:
Is crowdin able to commit directly so that we don't have to keep track of them? We could also add an action to call qgc-lupdate.sh periodically

DonLakeFlyer:
It only submits pulls automatically when things change. I just forced it manually to make sure everything was in sync.

Automatic lupdate would be nice. Be not sure how you push up to Crowdin automatically. Also not sure what sort of churn this might cause on the Crowdin side. Might be ok.

PR #11671, opened 2024-07-10
Guided altitude changes use relative alt instead of offset

DonLakeFlyer:
Changed guided altitude change to use relative alt instead of offset from current vehicle altitude. This gives precise positioning even when vehicle is climbing/descending.

@JMare Can you test this with ArduPilot? I looked at ArduPilot source and this "should" work.

Related to discussion  in #11641

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -838,20 +838,10 @@ void APMFirmwarePlugin::guidedModeRTL(Vehicle* vehicle, bool smartRTL)
     _setFlightModeAndValidate(vehicle, smartRTL ? smartRTLFlightMode() : rtlFlightMode());
 }
 
-void APMFirmwarePlugin::guidedModeChangeAltitude(Vehicle* vehicle, double altitudeChange, bool pauseVehicle)
+void APMFirmwarePlugin::guidedModeChangeAltitudeAMSL(Vehicle* vehicle, double altitudeAMSL, bool pauseVehicle)
 {
-    if (qIsNaN(vehicle->altitudeRelative()->rawValue().toDouble())) {
-        qgcApp()->showAppMessage(tr("Unable to change altitude, vehicle altitude not known."));
-        return;
-    }
-
     if (pauseVehicle && !_setFlightModeAndValidate(vehicle, pauseFlightMode())) {
         qgcApp()->showAppMessage(tr("Unable to pause vehicle."));
-        return;
-    }
-
-    if (abs(altitudeChange) < 0.01) {
-        // This prevents unecessary changes to Guided mode when the users selects pause and doesn't really touch the altitude slider
         return;
     }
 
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
Once this is verified to work with ArduPilot I'm going to add the fix for the horizontal slider showing up to this as well.

DonLakeFlyer:
@JMare I mean ArduPilot SITL

JMare:
@DonLakeFlyer I tested this with Ardupilot SITL and reviewed the code

As far as altitude control it works great. Did a bunch of takeoffs landings and change altitudes and it behaves exactly as expected. There is now no error if you do a change alt while the vehicle is climbing/descending.

However, now that the SET_POSITION_TARGET is using cmd.coordinate_frame = MAV_FRAME_LOCAL_NED, there is a problem where any change alt returns the vehicle horizontally above the home point. This is because the type mask is set to control x, y, and z, and the x and y are set to zero.

I tried fixing this by changing the type mask to ignore x and y, which I think should be 0xFFFB, rather than 0xFFF8 as it is now.
However when I did this the autopilot doesn't seem to accept the command at all.
I had a squizz at the source and it seems like ardupilot takes an all or nothing approach to the position fields in this message.
See
https://github.com/ArduPilot/ardupilot/blob/630d3fa8a7f3624469f3025b321e2724328f3b17/ArduCopter/GCS_Mavlink.cpp#L1297C51-L1297C87
and
https://github.com/ArduPilot/ardupilot/blob/630d3fa8a7f3624469f3025b321e2724328f3b17/ArduCopter/GCS_Mavlink.cpp#L1177

Sorry, I guess this may have been why it was done this way originally - might not be possible with ardupilot, but i'll have more of a look in the source and see if theres another message or way that it might be doable, because it would be nice if it could be done this way.

JMare:
@DonLakeFlyer I had a look and think about this and I can think of a couple of ways this could work well for ardupilot given the constraints.

Problem:
When sending SET_POSITION_TARGET_LOCAL_NED, x,y,z must all be populated.

Solution 1:
When sending a guided goto command, we store in QGC the x,y,z of the last command to the autopilot.
This means that on the next command, if we are already in guided, we can use the last values of the fields we don't want to change.
For instance, in a change alt command, we send the autopilot the new altitude for z, and the x,y of the last command we sent it.

Solution 2:
When we do a change alt command, if we are already in guided, we request the current POSITION_TARGET_LOCAL_NED from the autopilot using MAV_CMD_REQUEST_MESSAGE. We use this info to populate x,y of the message. The same approach could be used to populate the z field during a goto command. This solution would be more robust incase another GCS sends a guided command

Both of these solutions would have a side affect - which I think is good thing, that if the vehicle was in motion towards a new guided command, changing x,y wouldn't stop its movement in z, and changing z wouldn't stop its movement in x,y. 
Currently if the vehicle is moving in x,y, sending a change altitude command will make it stop dead in x,y. The same applies if you do a change alt followed by a go-to, the aircraft will stop dead in z.
Given that the UI separates X,Y and Z, it would be more intuitive if they operated independantly.

I think this would be an improvement, but I can see that it might be too complicated and there would be edge cases around takeoff, mode switches etc.
What do you think? Id be happy to work on this.

DonLakeFlyer:
I'm really concerned that the asynchronous nature of asking for current local target and then updating might be error problem and have strange edge cases. I think leaving it as is is best for now.

JMare:
Yep makes sense, thanks

PR #11678, opened 2024-07-13
Windows: Fix Install

HTRamsey:
This should hopefully fix running windows and the shortcuts not working.
Edit: I just realized this will mess up the qmake install, so maybe this has to wait until we get rid of qmake?

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -142,8 +142,8 @@ int main(int argc, char *argv[])
     // Allow for command line override of renderer
     for (int i = 0; i < argc; i++) {
         const QString arg(argv[i]);
-        if (arg == QStringLiteral("-angle")) {
-            QCoreApplication::setAttribute(Qt::AA_UseOpenGLES);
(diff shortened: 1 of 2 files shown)


src/main.cc
```diff
@@ -142,8 +142,8 @@ int main(int argc, char *argv[])
     // Allow for command line override of renderer
     for (int i = 0; i < argc; i++) {
         const QString arg(argv[i]);
-        if (arg == QStringLiteral("-angle")) {
-            QCoreApplication::setAttribute(Qt::AA_UseOpenGLES);
+        if (arg == QStringLiteral("-desktop")) {
```
DonLakeFlyer, line 145: Isn't this the default. Same as specifying nothing on the command line?

HTRamsey: It didn't work until I added that so I don't necessarily think so? I was getting a black screen without that.

DonLakeFlyer: ok, fine

PR #11681, opened 2024-07-16
New feature: Option to reverse the virtual joystick layout for left-handed users

demirciAhmet:
An option added to reverse the virtual joystick layout. You can enable the option from:  *Application Settings/Fly View/Left-Handed Mode*

Checklist:
----------


src/UI/preferences/FlyViewSettings.qml
```diff
@@ -179,6 +180,14 @@ SettingsPage {
             enabled:            _virtualJoystick.rawValue
             fact:               _virtualJoystickAutoCenterThrottle
         }
+
+        FactCheckBoxSlider {
+            Layout.fillWidth:   true
+            text:               qsTr("Left-Handed Mode")
```
DonLakeFlyer, line 186: I'm not sure people will understand what this means. I think something like "Left-Handed Mode (swap sticks)" would be better.

demirciAhmet: I agree with that.

Issue #11673, opened 2024-07-11
[v4.4.0] Remote ID GPS GCS Location type fixed not working


julianoes:
## Behavior

In QGC v4.4.0 I'm trying to select Location Type "Fixed" for the remote ID GPS GCS setting. However, the setting does activate. In other words. I click on "Fixed" but it's not accepted.

This seems fixed in master, so I wonder what commit needs backporting to v4.4.1 to fix this.

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1. Set up Remote ID
2. GPC GCS -> Location Type -> Try to select "Fixed"

## System Information

- QGC v4.4.0
- Linux Ubuntu 22.04

## Log Files and Screenshots

![image](https://github.com/mavlink/qgroundcontrol/assets/1419688/bc53ebbc-f727-4482-b2e5-602175aa02f7)

PR #11682, opened 2024-07-17
RemoteID: add note that Fixed only works in EU

julianoes:
It is very confusing when the "Fixed" option just seemingly doesn't work when you select it. Therefore, we should at least add this note.

Closes https://github.com/mavlink/qgroundcontrol/issues/11673.


Davidsastresas:
I agree, this note should have been there from the beginning. Thanks Julian!

PR #11688, opened 2024-07-19
QtLocationPlugin: Update QGeoMapReplyQGC

HTRamsey:
Updates the QGeoMapReplyQGC implementation to be more agnostic to the being used for altitude and tiles, and create the network requests in the tilefetcher instead of the map providers

### DIFF
--- a/src/QtLocationPlugin/MapProvider.cpp
+++ b/src/QtLocationPlugin/MapProvider.cpp
@@ -11,7 +11,6 @@
 #include <QGCLoggingCategory.h>
 
 #include <QtCore/QLocale>
-#include <QtNetwork/QNetworkRequest>
 
 QGC_LOGGING_CATEGORY(MapProviderLog, "qgc.qtlocationplugin.mapprovider")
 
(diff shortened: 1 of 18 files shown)


src/QtLocationPlugin/MapProvider.cpp
```diff
@@ -40,18 +39,9 @@ MapProvider::~MapProvider()
     // qCDebug(MapProviderLog) << Q_FUNC_INFO << this << _mapId;
 }
 
-QNetworkRequest MapProvider::getTileURL(int x, int y, int zoom) const
+QString MapProvider::getTileURL(int x, int y, int zoom) const
```
bkueng, line 42: Shouldn't this return `QUrl`?

src/QtLocationPlugin/MapProvider.h
```diff
@@ -53,8 +55,8 @@ class MapProvider
     QGeoMapType::MapStyle getMapStyle() const { return _mapStyle; }
     const QString& getMapName() const { return _mapName; }
     int getMapId() const { return _mapId; }
-
-    virtual QNetworkRequest getTileURL(int x, int y, int zoom) const;
+    QString getReferrer() const { return _referrer; }
```
bkueng, line 58:
```suggestion
    const QString& getReferrer() const { return _referrer; }
```

src/QtLocationPlugin/QGeoTileFetcherQGC.cpp
```diff
@@ -88,3 +97,31 @@ void QGeoTileFetcherQGC::handleReply(QGeoTiledMapReply *reply, const QGeoTileSpe
...
+    const QByteArray token = mapProvider->getToken();
+    if (!token.isEmpty()) {
+        request.setRawHeader(QByteArrayLiteral("User-Token"), token);
+    }
+    // request.setOriginatingObject(this);
+    request.setAttribute(QNetworkRequest::CacheLoadControlAttribute, QNetworkRequest::PreferCache);
+    request.setAttribute(QNetworkRequest::BackgroundRequestAttribute, true);
+    request.setAttribute(QNetworkRequest::CacheSaveControlAttribute, true);
+    request.setAttribute(QNetworkRequest::DoNotBufferUploadDataAttribute, false);
+    // request.setAttribute(QNetworkRequest::AutoDeleteReplyOnFinishAttribute, true);
```
bkueng, line 122: Did you plan to uncomment these or why are they commented?

HTRamsey: That's an option I was considering but needed to test more after remaining changes and didn't want to forget. It will either be deleted or uncommented by the end, but I can also just remove it now.

bkueng (approved):

Issue #11544, opened 2024-05-16
Clear button offline 


arnaudjuin:
It would be nice to have an option to clear the planning view when QGC is not connected to the vehicle.

smadi0x86:
I agree that having an option to clear the planning view when QGroundControl (QGC) is not connected to the vehicle would be a useful enhancement.

To address this issue, I propose adding a Clear button that is always enabled, regardless of the vehicle connection status. This button would allow users to clear all mission items from the planning view, even when QGC is offline.

**Here's a high-level plan for implementing this feature:**

- Add a Clear button to the toolstrip in the planning view.
- Implement a clearPlanningView function that removes all mission items and resets the view.
- Ensure the Clear button triggers this function.

**Files to edit for this features:**

- https://github.com/mavlink/qgroundcontrol/blob/master/src/PlanView/PlanView.qml

I would love to hear your feedback on this approach or any suggestions you might have before I start working on a pull request.

Thanks!

arnaudjuin:
That's the way I see it. I can also help you with the PR.

smadi0x86:
Hello, I hope you're doing well, Sorry for the delay in working on this, I was on vacation. I've put together a PR (https://github.com/mavlink/qgroundcontrol/pull/11691)  to address the issue and detailed what I've done in it.

Turns out my earlier insights were incorrect after diving deep into the code, the logic to enable the "Clear" button when offline was already in place. I tried tracing and changing the logic to just give a warning when vehicle is offline and execute the `clear` button.

DonLakeFlyer:
You can just use File/Blank to start over again. Is that too confusion? Adding a toolstrip button to clear isn't a great use of important real estate. Options like "Clear" are already under the File button.

smadi0x86:
I totally agree, in fact I was trying to copy the same functionality/logic of creating a new blank plan for clearing via the clear button. I have no problem with both, if the feature requester have any strong takes on why we need to that, we can possibly continue if that's possible.

DonLakeFlyer:
> if the feature requester have any strong takes on why we need to that, we can possibly continue if that's possible.

Or maybe just re-oeganizse/re-name things to make them more clear. We'll see where the. confusion is.

smadi0x86:
Yes that's the way I see it, do you have anything in mind?

DonLakeFlyer:
[...]

PR #11691, opened 2024-07-19
Fix Button Enabling and Data Clearing Behavior for Offline Vehicle State

smadi0x86:
# Fix Button Enabling and Data Clearing Behavior for Offline Vehicle State

Description
-----------
**This update addresses two key issues related to the data clearing functionality within the application:**

**1. Button Enabling:** The "Clear" button is now enabled even when the vehicle is offline. Previously, the button was only enabled when the vehicle was online and not syncing, which prevented users from clearing data when the vehicle was offline.

**2. Data Clearing Behavior:** Adjustments were made to ensure that the removeAllFromVehicle() functions in various controllers (Mission, GeoFence, RallyPoint) are executed even if the vehicle is offline. This change allows the application to clear all relevant data regardless of the vehicle’s online status. Warnings are logged when these actions are attempted while offline, but the operations will still proceed.

**Note:** This PR is still not ready, this was created based on the issue author request for potential collaboration.

After these changes, clearing isn't working when vehicle is offline and I am getting this error even after clicking on `yes`:
```
qml: User cancelled: Did not clear mission items from the vehicle 
```
Test Steps
-----------

**1. Verify Button Functionality:**

- Launch the application and ensure the vehicle is offline.
- Check if the "Clear" button is enabled in the user interface.
- Click the "Clear" button and confirm that the appropriate data clearing actions are attempted.

**2. Validate Data Clearing:**

- When the vehicle is offline, trigger data clearing operations for missions, geofences, and rally points.
- Verify that the operations proceed and that warnings are logged appropriately, indicating the offline status but still performing the clear actions.

**3. Check Online Behavior:**

- Switch the vehicle to an online state.
- Ensure the "Clear" button is still enabled and functions correctly.
- Confirm that data clearing operations proceed as expected when the vehicle is online.

Checklist:
----------

Related Issue
-----------
https://github.com/mavlink/qgroundcontrol/issues/11544

### DIFF
--- a/src/MissionManager/GeoFenceController.cc
+++ b/src/MissionManager/GeoFenceController.cc
@@ -240,12 +240,16 @@ void GeoFenceController::removeAll(void)
 
 void GeoFenceController::removeAllFromVehicle(void)
 {
+    qCInfo(GeoFenceControllerLog) << "GeoFenceController::removeAllFromVehicle called";
     if (_masterController->offline()) {
         qCWarning(GeoFenceControllerLog) << "GeoFenceController::removeAllFromVehicle called while offline";
-    } else if (syncInProgress()) {
-        qCWarning(GeoFenceControllerLog) << "GeoFenceController::removeAllFromVehicle called while syncInProgress";
-    } else {
+    }
+
+    try {
         _geoFenceManager->removeAll();
+        qCInfo(GeoFenceControllerLog) << "All geofences removed from the vehicle";
+    } catch (const std::exception& e) {
+        qCCritical(GeoFenceControllerLog) << "Failed to remove geofences:" << e.what();
     }
 }
 
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -2246,13 +2246,16 @@ bool MissionController::containsItems(void) const
 
 void MissionController::removeAllFromVehicle(void)
 {
+    qCInfo(MissionControllerLog) << "MissionController::removeAllFromVehicle called";
     if (_masterController->offline()) {
-        qCWarning(MissionControllerLog) << "MissionControllerLog::removeAllFromVehicle called while offline";
-    } else if (syncInProgress()) {
-        qCWarning(MissionControllerLog) << "MissionControllerLog::removeAllFromVehicle called while syncInProgress";
-    } else {
-        _itemsRequested = true;
+        qCWarning(MissionControllerLog) << "MissionController::removeAllFromVehicle called while offline";
+    }
+
+    try {
         _missionManager->removeAll();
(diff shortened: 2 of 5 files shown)


DonLakeFlyer:
There are a number of problems here, some I discussed in #11544:

As the UI currently stands the "Clear" button is associated with a vehicle:
Screenshot 2024-08-01 at 9 36 44 AM
So it doesn't make sense for things under "Vehicle" to work when there is no Vehicle. You have also modified methods called removeAllFromVehicle to not do what they say they do. They are no longer associated with a vehicle.

I think it's a bad idea to allow removeAllFromVehicle while a sync is on progress. There is a sync state machine running which you then intersperse another mission command on top off while the sync continues to run.  No idea what sort of strange problems this could lead to. Also don't understand why this is something interesting to support. In order to support well there would need to be the ability to cancel a sync cleanly and then do the remove all.

As I said in the Issue the ability to start over from scratch when not connected to a vehicle is already supported by the  Create a Blank Plan button. I believe that is the exact functionality you want, it's just not called Clear. I realize this may not be as obvious as a "Clear" button but I'm also not quite sure how confusing it is for most people. In order to resolve this it would require rework of the overall UI, not just enabling things that weren't enabled before. At this point in looking at it I'm not super convinced this is a big problem that needs solving. Let me ask the community what they think.

smadi0x86:
> There are a number of problems here, some I discussed in #11544:
> 
> As the UI currently stands the "Clear" button is associated with a vehicle: Screenshot 2024-08-01 at 9 36 44 AM So it doesn't make sense for things under "Vehicle" to work when there is no Vehicle. You have also modified methods called removeAllFromVehicle to not do what they say they do. They are no longer associated with a vehicle.
> 
> I think it's a bad idea to allow removeAllFromVehicle while a sync is on progress. There is a sync state machine running which you then intersperse another mission command on top off while the sync continues to run. No idea what sort of strange problems this could lead to. Also don't understand why this is something interesting to support. In order to support well there would need to be the ability to cancel a sync cleanly and then do the remove all.
> 
> As I said in the Issue the ability to start over from scratch when not connected to a vehicle is already supported by the Create a Blank Plan button. I believe that is the exact functionality you want, it's just not called Clear. I realize this may not be as obvious as a "Clear" button but I'm also not quite sure how confusing it is for most people. In order to resolve this it would require rework of the overall UI, not just enabling things that weren't enabled before. At this point in looking at it I'm not super convinced this is a big problem that needs solving. Let me ask the community what they think.

Yeah as mentioned in the issue this requires a community decision, from my side I changed this PR to a draft until further notice. And yes at this point its a re-design to the UI not only to fix the clear view.

DonLakeFlyer:
I'm closing this since it is incorrect. I've submitted another pull that should make things a little better.

PR #11698, opened 2024-07-23
Fix build without gstreamer or without Qt6 gstreamer plugin

SleepingSoul:
VideoReceiver: provide GLVideoItemStub.h in public library interface if gstreamer is missing or Qt6 plugin wasn't found

Description
-----------
Currently, build is failing due to GLVideoItemStub.h being included in VideoManager.cc:
```
#ifdef QGC_GST_STREAMING
#include "GStreamer.h"
#include "VideoDecoderOptions.h"
#else
#include "GLVideoItemStub.h"
#endif
```
This header file is expected to be provided by VideoReceiver::GStreamer library. However, this library is not built at all due to early return statement in its CMakeLists.txt:
```
if(NOT GST_QT6_PLUGIN_FOUND)
    return()
endif()
```
I have fixed it by still building this library (with a single GLVideoItemStub.h) file.

Test Steps
-----------
I have tested it by successfully building and launching QGroundControl without gstreamer in Windows, in Release configuration.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/src/VideoManager/VideoReceiver/GStreamer/CMakeLists.txt
+++ b/src/VideoManager/VideoReceiver/GStreamer/CMakeLists.txt
@@ -1,12 +1,26 @@
 add_subdirectory(${CMAKE_SOURCE_DIR}/libs/qmlglsink qmlglsink.build)
 
+find_package(Qt6 REQUIRED COMPONENTS Core Quick)
+
 if(NOT GST_QT6_PLUGIN_FOUND)
+    qt_add_library(GStreamerReceiver STATIC
+        GLVideoItemStub.h
+    )
+
+    target_link_libraries(GStreamerReceiver
+        PRIVATE
+            Qt6::Quick
+            Utilities
+        PUBLIC
+            Qt6::Core
+    )
+
+    target_include_directories(GStreamerReceiver PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
+
     return()
 endif()
 
 message(STATUS "Building GStreamer VideoReceiver")
-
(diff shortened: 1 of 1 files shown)


SleepingSoul:
Linux build failure looks false positive to me

HTRamsey:
Yup, Vulkan just updated their latest version which will cause that download to fail temporarily. I will re-run again in like an hour just to double check the build but this looks good to me. I guess without video streaming you shouldn't even need FlightDisplayViewVideo and thus wouldn't need GLVideoItemStub, you could just use a loader with the enabled property, but this works for now.

SleepingSoul:
> Yup, Vulkan just updated their latest version which will cause that download to fail temporarily. I will re-run again in like an hour just to double check the build but this looks good to me.
> 
> I guess without video streaming you shouldn't even need FlightDisplayViewVideo and thus wouldn't need GLVideoItemStub, you could just use a loader with the enabled property, but this works for now. I suppose that will be different when QtMultimedia receiver is integrated into Qml though, so this is probably fine anyways.

Yes, I think in perfect scenario we don't need GLVideoItemStub at all because we can load QML file that uses:
`org.freedesktop.gstreamer.Qt6GLVideoItem`
in runtime at the right moment, or do not load at all if gstreamer Qt6 plugin is missing. This is what I am doing in my own project that uses qt6 plugin. But yeah, it would be more of a refactoring and much more radical change. I hope this simple fix will be OK for now.

HTRamsey:
Looks like vulkan may have removed ubuntu 20.04 for ubuntu 24.04, so I may have to remove that from CI. https://packages.lunarg.com/vulkan/1.3.290/dists/

PR #11704, opened 2024-07-25
Docs: Update Release Version to v4.4

HTRamsey:
Website currently shows 4.3 instead of 4.4, not sure if this is the correct way to update this or not.


DonLakeFlyer (approved):
I think that's right

PR #11714, opened 2024-07-27
Android: Serial Fixes

HTRamsey:
- Remove custom SerialInputOutputManager and use default from library 
- Improve QSerialPortInfo data (was missing a couple of values)
- Create QGCSerialListener class
- Fix some miscellaneous issues and remove some unused functions


android/libs/qtandroidserialport/qserialport_android.cpp
```diff
@@ -38,12 +38,12 @@ bool QSerialPortPrivate::open(QIODevice::OpenMode mode)
         return false;
     }
 
-    /*if (mode & QIODevice::ReadOnly) {
```
DonLakeFlyer, line 41: What's with this uncommented and then commented out stuff?

android/AndroidManifest.xml
```diff
@@ -56,7 +58,6 @@
             <intent-filter>
                 <action android:name="android.hardware.usb.action.USB_DEVICE_ATTACHED" />
                 <action android:name="android.hardware.usb.action.USB_DEVICE_DETACHED" />
-                <action android:name="android.hardware.usb.action.USB_ACCESSORY_ATTACHED" />
```
DonLakeFlyer, line 59: What's the distinction between USB_DEVICE and USB_ACCESORY?

HTRamsey: Accessory was used for Taisync before that was removed. It's just which end acts as the USB Host. https://developer.android.com/develop/connectivity/usb

android/libs/qtandroidserialport/qserialport_android.cpp
HTRamsey: I tried to model this file after Qt's Unix class, but some commented out code doesn't work as I expected yet or I'm not sure it's necessary. I'll either fix it or remove it in a follow up after more testing.

PR #11723, opened 2024-07-30
Comms: Update LinkInterface & LinkConfiguration

HTRamsey:
Initial prep for work involving #11598 (such as #11655) by cleaning up LinkInterface & LinkConfiguration classes
Fixes documentation format


DonLakeFlyer (approved):
It's a bit tricky to review with all the code movement, but seems ok. One thing is the usage of m_config which isn't to coding style. Protected/Private is just _config.

HTRamsey:
Okay, I can change that. I was following Qt's style of member variables being prepended with 'm_' and private functions just starting with '_'

Issue #11713, opened 2024-07-27
MAVLink Signing Improvements


HTRamsey:
Follow up to #11596.
The initial MAVLink Signing implementation supports basic setting of a user provided secret key and sending the setup_signing message to enable/disable signing. Necessary improvements include:

- Cleaning up the QML implementation of setting key & enabling/disabling signing
- Allow unsigned messages options in QML?
- Verify valid secret key input before allowing signing enabled
- Ability to receive setup_signing message from another component?
- Persistent storage of timestamp
- Remove signature and setup_signing message from logs
- Only send setup_signing if connection is wired?

DonLakeFlyer:
Isn't the current implementation also missing incoming packet verification against the signing key?

DonLakeFlyer:
> Isn't the current implementation also missing incoming packet verification against the signing key? I can't seem to find that code.

Ok, I "think" it's included auto-magically as part of the internal C mavlink lib implementation.

DonLakeFlyer:
So I'm still confused by how this is connected into QGC. Problems I see:
* Mavlink signing settings are application not vehicle settings. Hence they can be set when there is no vehicle connected. The current implementation only calls Vehicle::_sendSetupSigning when the signing key value changes. If I setup the key first with no Vehicle and then connect the Vehicle nothing happens. Seems like there needs to be a specific, send the key to the vehicle button of something in order to get the key over. I don't think trying to make this auto-magic at all is a good thing. You also don't want the key going over an un-secure link by accident just be cause you changed a value. To me this should be ab explicit action taken by the user.
* It seems like setup for a channel to support signing (MavlinkSigning::initSigning() is only setup if Vehicle::_sendSetupSigning is called. Doesn't MavlinkSigning::initSigning need to be called at link connection time for all links if signing is enabled?

DonLakeFlyer:
Also not quite sure what should happen on a vehicle with multiple links.

HTRamsey:
This might be a bit more of a pain but I'm wondering about having the signing settings in each link settings when you are adding a link. Then check for changes when the link settings are changed to queue sending the setup_signing message until the link is connected or something. Maybe it only sends setup_signing at initial connection not sure. Then you can get what type of connection the link is and adjust some settings based on if it is a secure connection (serial/ethernet based) or wifi/bluetooth based.

DonLakeFlyer:
I think it's fine to have a single global signing key for all links. Here is what should happen:
[...]

PR #11732, opened 2024-07-31
More work on Mavlink Signing

DonLakeFlyer:
Related to #11713 

Changed:
* No enable/disable setting for signing. The existence of a non-empty key enables signing
* Remove support for a custom mavlink build with MAVLINK_NO_SIGN_PACKET turned on. This just adds complexity. If someone wants to build something like this they can modify QGC themselves.
* Signing is initialized on a channel as soon as a link is created. This means any link created gets signing.
* When the signing key is changed in settings all links are updated to new key
* Sending the key to the vehicle is a user initiated action from Telemetry settings. This only sends setup signing. It no longer also does an init since that would have already happened on link create or key change.
* Signing is not turned on for USB and Log Replay connections

Things I still want to do:
* Give the user some notification if signing initialization fails. You'll see a FIXME in the code for this for now
* Give the user some notification on signing failures like signature mismatches. Right now things just fall on the floor on signature mismatch and Vehicles don't get created since you never see a heartbeat.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -48,7 +48,7 @@
 #include "VehicleObjectAvoidance.h"
 #include "VideoManager.h"
 #include "VideoSettings.h"
-#include <DeviceInfo.h>
+#include "QGCDeviceInfo.h"
 #include <StatusTextHandler.h>
 #include <MAVLinkSigning.h>
 
@@ -164,7 +164,7 @@ Vehicle::Vehicle(LinkInterface*             link,
 
     // Send MAV_CMD ack timer
     _mavCommandResponseCheckTimer.setSingleShot(false);
-    _mavCommandResponseCheckTimer.setInterval(_mavCommandResponseCheckTimeoutMSecs);
+    _mavCommandResponseCheckTimer.setInterval(qgcApp()->runningUnitTests() ? 100 : _mavCommandResponseCheckTimeoutMSecs);
     _mavCommandResponseCheckTimer.start();
     connect(&_mavCommandResponseCheckTimer, &QTimer::timeout, this, &Vehicle::_sendMavCommandResponseTimeoutCheck);
 
@@ -325,8 +325,6 @@ void Vehicle::_commonInit()
     connect(_toolbox->corePlugin(), &QGCCorePlugin::showAdvancedUIChanged, this, &Vehicle::flightModesChanged);
 
     connect(_imageProtocolManager, &ImageProtocolManager::imageReady, this, &Vehicle::_imageProtocolImageReady);
-
-    (void) connect(_settingsManager->appSettings()->mavlink2Signing(), &Fact::rawValueChanged, this, &Vehicle::_sendSetupSigning);
 
     _createStatusTextHandler();
 
@@ -494,12 +492,6 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
             if(packet_lost_count)
                 emit messagesLostChanged();
         }
-    }
-    // TODO: Set Signing AppSetting Based on First Message?
-    const bool mavlinkSigning = message.incompat_flags & MAVLINK_IFLAG_SIGNED;
-    if (mavlinkSigning != _mavlinkSigning) {
-        _mavlinkSigning = mavlinkSigning;
-        emit mavlinkSigningChanged();
     }
 
     // Give the plugin a change to adjust the message contents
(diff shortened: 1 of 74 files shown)


DonLakeFlyer:
I also simplified the determination of whether a link is signed or not with just nullptr on the signing struct instead of looking at the outgoing flag.

HTRamsey:
Yup, I'll take a look at it.

DonLakeFlyer:
So I looked at ArduPilot source, Mission Planner and pymavlink to see if the approach I took above makes sense or not. Let me know if you think it does. It's not super clear from the docs how this should all work.

DonLakeFlyer:
Also just realizing this pul is screwed up. I messed up taking out my testing hacks.

HTRamsey:
Can you remove the isNetworkWired part of the _defaultAcceptUnsignedCallback function? Maybe we can have a LinkInterface function that checks if the link is secure (serial/ethernet) by using the channel from the signing struct that the callback gets. So basically loop through the LinkManagers links, and compare the link's mavlinkChannel to the link_id of the signing struct, and then call isLinkSecure() or something on the matching link.

src/Vehicle/Vehicle.cc
```diff
@@ -495,12 +493,6 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
                 emit messagesLostChanged();
         }
     }
-    // TODO: Set Signing AppSetting Based on First Message?
-    const bool mavlinkSigning = message.incompat_flags & MAVLINK_IFLAG_SIGNED;
-    if (mavlinkSigning != _mavlinkSigning) {
```
HTRamsey, line 500: I would also re-add this, as this is the only way to know if it is actually working or not.

DonLakeFlyer: Yeah, that will be part of the remaining work on showing signing status somehow.

DonLakeFlyer:
> Can you remove the isNetworkWired part of the _defaultAcceptUnsignedCallback function?

Ah, didn't realize that was available. Let me figure out how to make that work like I made LinkManager::isUSBDirect work.

HTRamsey:
One other minor thing I noticed is that when you click immediately on the send signing message button after editing the signing key, the notice about making sure to press that button pops up for like a fraction of a second. I guess it would be better if it didn't pop up at all if you go from editing the key straight to pressing the send button. I had to go in to the code to figure out what it was saying

DonLakeFlyer:
> I assume we can be certain the key will always update before sending the message right?

Hmm, not sure the order there. Let me check to make sure. It happens on focus loss. I'll look at the Click the send button thing as well.

DonLakeFlyer:
```
bool _defaultAcceptUnsignedCallback(const mavlink_status_t *status, uint32_t message_id)
{
    Q_UNUSED(status);

    static const QSet<uint32_t> unsigned_messages({MAVLINK_MSG_ID_RADIO_STATUS});

    return (unsigned_messages.contains(message_id) || QGCDeviceInfo::isNetworkWired());
}
```

Just to double-check. This code as it currently stands is wrong isn't it? This just says that the computer itself is using a wired connection. Whereas it doesn't necessarily mean that QGC is connected to the board over a wired connection.

HTRamsey:
Correct, so it should really only be checked for links like UDP & TCP. You could be using bluetooth and connected via ethernet which would make that part wrong.

DonLakeFlyer:
Sorry, taking me a bit...

Decided to add configurable secure connection support to MockLink for debugging and better end to end signing unit tests.

DonLakeFlyer:
Almost done. Turned out to be pretty involved to support well and have good end to end unit tests

HTRamsey:
Gotcha, sorry I didn't intend for you to have to do that much for this

DonLakeFlyer:
No problem. Nothing to do with your stuff. Just involved to support signing in MockLink. But that in turn makes testing this stuff easier when you don't have a real vehicle.

DonLakeFlyer:
First pass at this is finally done. Founds lots of problems along the way in old code relating to mavlink v2 negotation. Checking CI and then after that I'm going to make a new pull which is broken up into commits which make sense together.

DonLakeFlyer:
@HTRamsey Any idea what is going on with the Linux cmake build? This builds fine for me on OSX. Doesn't seem like a failure that should be OS dependent. Seems more like something missing in cmake build files.

HTRamsey:
It's a problem with one of the unit tests having a segfault, not the build itself

DonLakeFlyer:
I was looking at the unit tests. I see that now. I'll look. Odd that release doesn't build but debug does build on Linux.

HTRamsey:
Oh you're referring to the qmake build not working, I didn't think that really mattered at this point

DonLakeFlyer:
Duh, qmake not cmake! Someday I'll learn to read.

DonLakeFlyer:
Problem with unit tests is that there was a flaky boundary on how long it would take a MockLink vehicle to be fully brought up. With the new updates to MockLink for signing we're touching the boundary where wait times in unit tests are not long enough. Fix is to reduce timeouts on the various retries which occur during the vehicle boot sequence when running unit tests. There is a bunch of this already in there, but more is needed specifically in ComponentInformationManager it looks like. No reason to wait so long for things to happen when running against a MockLink since you have a direct connection to it.

DonLakeFlyer:
Closing this since I'm splitting the changes into multiple pulls

PR #11735, opened 2024-07-31
Remove Deprecated QMake Build System

HTRamsey:
Remove everything related to qmake or no longer needed. Left the git submodules however in the event we end up using them


DonLakeFlyer (approved):

DonLakeFlyer:
Yay!

DonLakeFlyer:
Moving from qmake to cmake has been in the works for years now.  This is both from a QGC standpoint and also from the standpoint of Qt itself. During that transition time we have maintained both build systems. With the move of daily to Qt6 we have changed fully to cmake only.

simiken1234:
So what is now the recommended way to create an appimage of a customized version please?

HTRamsey:
> So what is now the recommended way to create an appimage of a customized version please?

That's in the docs

PR #11740, opened 2024-08-01
Comms: Prepare TCP for Threading Changes

HTRamsey:
Related to https://github.com/mavlink/qgroundcontrol/issues/11598
Since every link type will have to be updated at the same time when the inheritance is moved from QThread to QObject, this is a PR to prepare everything else to simplify that change.


DonLakeFlyer (approved):
All seems reasonable to me

PR #11755, opened 2024-08-06
MissionController.cc: fix bug current mission item not clearing:

Davidsastresas:
In plan view, when selecting a mission item which is a child item ( on the right mission item view flickable for example ), the child item gets set as current item, and thus the parent gets highlighted in the map, as expected.

However, from there, when selecting a different current mission item, the previous highlighted ( parent of the previous selected mission item ) does not get cleared. If repeating this process, we end up with a lot of falsely highlighted items on the map.

This happens because we were forgetting to clear the hasCurrentChildItem property. This commit fixes this.


DonLakeFlyer:
thanks @Davidsastresas

PR #11756, opened 2024-08-06
CMake: Remove Duplicate Install of Resources

HTRamsey:
These are already installed automatically when they are linked to the project sources


HTRamsey:
@DonLakeFlyer This will help with the build size some

PR #11757, opened 2024-08-07
CI: Various Fixes & Cleanup

HTRamsey:
- Merge Windows Arm & Windows x86
- Update GStreamer minor revision
- Standardize formatting/conditionals/order of workflows
- Add geographiclib to installs
- Update MacOS version
- Create Install Android Qt action to reduce duplicated code & prepares for github runners removing r25b NDK


DonLakeFlyer (approved):
So we're clear now on the macOS GitHub deprecation right?

HTRamsey:
Yup, and they are about to remove the cached NDK we use for android so this fixed that too. The next thing to be aware of in the future will probably be when they remove Ubuntu 20.04 which I though was supposed to be this year but not sure. But we can always migrate to doing builds in Docker if we have to.

PR #11762, opened 2024-08-09
SurveyComplexItem: remove px4 hack overriding default turnaround distance

Davidsastresas:
I discovered this trying to override it from a custom plugin. Is this still neccesary? @julianoes could you confirm something like this is needed for PX4, in order to address it differently? It avoids setting a different default from a custom plugin. Thanks!

### DIFF
--- a/src/MissionManager/SurveyComplexItem.cc
+++ b/src/MissionManager/SurveyComplexItem.cc
@@ -38,13 +38,6 @@ SurveyComplexItem::SurveyComplexItem(PlanMasterController* masterController, boo
 {
     _editorQml = "qrc:/qml/SurveyItemEditor.qml";
 
-    // If the user hasn't changed turnaround from the default (which is a fixed wing default) and we are multi-rotor set the multi-rotor default.
-    // NULL check since object creation during unit testing passes NULL for vehicle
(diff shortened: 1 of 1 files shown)


julianoes (commented):
I'm afraid I have no idea. I'd have to play with it in SITL to figure out.

Davidsastresas:
@julianoes, do you know anybody that could confirm this? Should we hold on for the moment on this one then?

julianoes:
"Remove it and see who complains" is my best answer, sadly.

Davidsastresas:
That works for me, thanks Julian, I will give it a couple days more and I will merge.

DonLakeFlyer:
Can you walk that change back through the pulls to see if the comments in the pull have additional information?

DonLakeFlyer:
> "Remove it and see who complains" is my best answer, sadly.

That seems like a bad idea without knowing more details on the original reason to put it in. Comments says it's for PX4, code could be changed to be PX4 only.

DonLakeFlyer:
Ok, I looked through the code. The Transect code already had individual settings for FW and MR turnaround distance which can be custom build overriden using the normal mechanism. The default for MR is 10 meters which matches this code. So that means it should be ok to remove this code. Since by default this code is just setting the turnaround distance to 10 when it is already 10! Old crap from before Transect had MR/FW differentiation capability would be my guess.

Davidsastresas:
@DonLakeFlyer thanks for verifying this!

PR #11774, opened 2024-08-15
Revert "Do not REQUEST_DATA_STREAMs by default, since it's deprecated"

HTRamsey:
Reverts mavlink/qgroundcontrol#11531
Need a better solution due to Copter not initiating streams automatically


TishSerg:
Yeah, that's what I was saying https://github.com/mavlink/qgroundcontrol/pull/11531#issuecomment-2127096773

PR #11778, opened 2024-08-18
QtLocationPlugin: Update QGCMapEngineManager

HTRamsey:
Remove a couple unused signals
Use Application Instance instead of Toolbox for QGCMapEngineManager instance to further separate location plugin from core code & reduce linking/dependencies
Perform code quality maintenance

### DIFF
--- a/src/QmlControls/QGroundControlQmlGlobal.cc
+++ b/src/QmlControls/QGroundControlQmlGlobal.cc
@@ -15,6 +15,7 @@
 #include "FirmwarePluginManager.h"
 #include "AppSettings.h"
 #include "PositionManager.h"
+#include "QGCMapEngineManager.h"
 #ifndef NO_SERIAL_LINK
 #include "GPSManager.h"
 #endif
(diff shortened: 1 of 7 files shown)


src/QtLocationPlugin/QMLControl/QGCMapEngineManager.cc
```diff
@@ -176,196 +155,168 @@ QGCMapEngineManager::startDownload(const QString& name, const QString& mapType)
...
+
+    const QStorageInfo info(path);
+
+    const quint32 total = static_cast<quint32>(info.bytesTotal() / 1024);
+    _diskSpace = total;
+
+    const quint32 free = static_cast<quint32>(info.bytesFree() / 1024);
+    setFreeDiskSpace(free);
+
+    qCDebug(QGCMapEngineManagerLog) << info.rootPath() << "has" << free << "Mbytes available.";
```
bkueng, line 313: The way it's calculated, it's in KB. There was also a comment below saying it's in MB, so maybe the calculation needs to be fixed.
```suggestion
    qCDebug(QGCMapEngineManagerLog) << info.rootPath() << "has" << free << "Kbytes available.";
```

src/QmlControls/QGroundControlQmlGlobal.cc
```diff
@@ -73,7 +74,7 @@ void QGroundControlQmlGlobal::setToolbox(QGCToolbox* toolbox)
 
     _linkManager            = toolbox->linkManager();
     _multiVehicleManager    = toolbox->multiVehicleManager();
-    _mapEngineManager       = toolbox->mapEngineManager();
+    _mapEngineManager       = QGCMapEngineManager::instance();
```
bkueng, line 77: This could probably be init in the constructor now.

src/QtLocationPlugin/QMLControl/QGCMapEngineManager.cc
HTRamsey: DiskSpace/FreeDiskSpace are currently unused, but they could be useful for Don to add in the UI so I just made a note about it

PR #11779, opened 2024-08-18
Settings: Remove unused settings

HTRamsey:
showLargeCompass isn't used
userBrandImageIndoor & userBrandImageOutdoor was duplicated from BrandImageSettings
defaultFirmwareType was duplicated from FirmwareUpgradeSettings
usePairing isn't used


src/Settings/App.SettingsGroup.json
```diff
@@ -179,20 +172,6 @@
     "default":              false,
     "qgcRebootRequired":    true
 },
-{
-    "name":             "userBrandImageIndoor",
```
DonLakeFlyer, line 183: Both of these are still used in the code. I think the UI for setting them was lost somewhere along the way with the settings ui rework. Create #11780

src/Settings/App.SettingsGroup.json
```diff
@@ -235,12 +214,6 @@
     "type":             "string",
     "default":     ""
 },
-{
-    "name":             "defaultFirmwareType",
```
DonLakeFlyer, line 239: Yeah, this is screwed up. There are three settings that are basically doing the similar things: 
* AppSettings.offlineEditingFirmwareClass - This is used by PlanView as the firmware type for plan creation when you are not connected to a vehicle. It is initially set by the user in the first run dialogs. You can also change from the Plan View Mission Start item when offline.
* AppSettings.defaultFirmwareType - This is no longer used. I think it may have been leftover from the change to the offline* settings.
* FirmwareUpgradeSettings.defaultFirmwareType - This remembers the last selection the user used for firmware flashing.

The idea is to remember whether the user is a PX4 or an ArduPilot user. I think the solution is to make all of these just be/reference AppSettings.offlineEditingFirmwareClass.

DonLakeFlyer (changes requested):

src/Settings/App.SettingsGroup.json
HTRamsey: They are used, but not this version of them. The ones in the BrandImageSettings Group are used instead of these

DonLakeFlyer: Ah, right

HTRamsey: Actually seems kinda unnecessary to have a new settingsgroup for just these two so maybe they belong in the appsettings. But I suppose it really doesn't matter that much. I think a few of the settings could use a bit of reorganization

DonLakeFlyer: Not quite sure why/when those moved/duplicated. Was that me? Problem with settings moves is not trouncing existing users stuff.

HTRamsey: Fair, so is this okay then?

DonLakeFlyer: Sure, if it's not used. Look at blame and can't quite figure out why the AppSettings variants were added.

HTRamsey: https://github.com/mavlink/qgroundcontrol/commit/23561c4a5ced26271879bf3e6cf131c6bf241d56

PR #11782, opened 2024-08-19
ADSB: Add Unit Test

HTRamsey:
Creates a unit test for ADSB
Remove need for restarting app when changing ADSB enabled setting
Better handling of TCPLink state
Change from Tool to Application Static to reduce linking and make better isolated/testable module
General code quality improvements

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -3277,36 +3277,37 @@ bool Vehicle::autoDisarm()
     return false;
 }
 
-void Vehicle::_handleADSBVehicle(const mavlink_message_t& message)
+void Vehicle::_handleADSBVehicle(const mavlink_message_t &message)
 {
     mavlink_adsb_vehicle_t adsbVehicleMsg;
-    static const int maxTimeSinceLastSeen = 15;
-
     mavlink_msg_adsb_vehicle_decode(&message, &adsbVehicleMsg);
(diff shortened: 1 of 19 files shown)


src/Vehicle/Vehicle.cc
```diff
@@ -3277,36 +3277,37 @@ bool Vehicle::autoDisarm()
...
 
         if (adsbVehicleMsg.flags & ADSB_FLAGS_VALID_HEADING) {
-            vehicleInfo.heading = (double)adsbVehicleMsg.heading / 100.0;
-            vehicleInfo.availableFlags |= ADSBVehicle::HeadingAvailable;
+            vehicleInfo.heading = adsbVehicleMsg.heading / 1e2;
+            vehicleInfo.availableFlags |= ADSB::HeadingAvailable;
         }
 
-        _toolbox->adsbVehicleManager()->adsbVehicleUpdate(vehicleInfo);
+        (void) QMetaObject::invokeMethod(ADSBVehicleManager::instance(), "adsbVehicleUpdate", Qt::AutoConnection, vehicleInfo);
```
DonLakeFlyer, line 3310: I don't quite get why this would be an invokeMethod. What is the reasoning behind when to use invokeMethod can when to call the method directly?

HTRamsey: I believe it's mostly just standard Qt practice to call slots with that now, so that connection type can be handled automatically even if you know it will be a direct call.

DonLakeFlyer: Ok, still seems a bit odd. But I guess future-proof for possible changes on called object down the road.

DonLakeFlyer (approved):

PR #11785, opened 2024-08-20
CMake: Fetch PX4-GPSDrivers

HTRamsey:
Fetch PX4-GPSDrivers through CMake. Had to add an empty folder to get the include path working


HTRamsey:
@julianoes is there a better way to handle the ../../definitions.h include issue

julianoes:
Without looking in detail, I would say the GPS repo is poorly architected and could use a bit of refactoring.

julianoes:
What about leaving a valid definitions.h in source in the GPS repo, and just allow PX4 (or whoever) to override it.

julianoes (approved):

PR #11786, opened 2024-08-20
CMake: Custom Build Fixes

HTRamsey:
Minor Custom Build Fixes


HTRamsey:
On Ubuntu 22.04 & Ubuntu 20.04 yes. It works on 24.04. Probably the backported gstreamer qt plugin is wrong

PR #11787, opened 2024-08-20
AnalzeView: Use easyexif library

HTRamsey:
Use easyexif for parsing in ExifParser. This library only handles parsing, so another alternative (https://github.com/Exiv2/exiv2) will have to be considered for writing exifs with a library

### DIFF
--- a/src/AnalyzeView/ExifParser.cc
+++ b/src/AnalyzeView/ExifParser.cc
@@ -1,8 +1,11 @@
 #include "ExifParser.h"
 #include "QGCLoggingCategory.h"
 
-#include <QtCore/QtEndian>
+#include <QtCore/QByteArray>
 #include <QtCore/QDateTime>
+#include <QtCore/QtEndian>
+
+#include <exif.h>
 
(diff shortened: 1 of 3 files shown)


DonLakeFlyer (approved):

src/AnalyzeView/ExifParser.cc
```diff
@@ -110,7 +101,7 @@ double readTime(const QByteArray &buf)
     return (tagTime.toMSecsSinceEpoch() / 1000.0);
 }
 
-bool write(QByteArray& buf, const GeoTagWorker::cameraFeedbackPacket& geotag)
+bool write(QByteArray &buf, const GeoTagWorker::cameraFeedbackPacket &geotag)
```
DonLakeFlyer, line 104: FYI: Most of the code uses the `foo& bar` style instead of `foo &bar`. Just pointing out, it's really fine either way. I've always thought the modifier `&/*` should go closer to the type since it's more related to that. I know weird. Just a quirk I have.

PR #11796, opened 2024-08-21
UI fixes

gillamkid:
Description
-----------
I added a few UI fixes/enhancements

1. Added hover highlighting
Buttons already got fully highlighted when checked or pressed. Now they get partially highlighted when hovered

Samples of before (no visual change when mouse hovers the buttons shown)
before1
before2
before3

samples of after - mouse hover highlights slightly (dark mode)
after10
after20
after30

samples of after - mouse hover highlights slightly (light mode)
after-light-10
after-light-20
after-light-30

2. Make some fixes to the "Palette Test" page
the light/dark toggle button was broken and some of the QGCComponent examples lookes messed up

before
before-QmlType

after
after-qmlType

3. Fix QGC icon
The QGC icon shown in the top left corner of the app did not match the typical QGC icon seen everywhere else (icon of QGC shortcut made after an install, icon branding on QGC website, etc). I replaced the instances of the extra-bloated Q with the more common QGC icon

before
a-before

after
a-after

4. fix padding on launch card

before
z-before

after
z-afer


DonLakeFlyer:
Thanks for all the nice cleanup. I'll take a look.

DonLakeFlyer:
If you look at QGCButton you'll see that `hoverEnabled` is disabled on mobile builds. This works around a problem where the hover would get stuck on for android. Since there really isn't "hover" on touch systems this fixes the problem without really changing any ui behavior.

So for any control which uses a raw `Button` as base and has had hover support added to it this same `hoverEnabled` things needs to be there.

These changes affect all button controls except QGCButton. Seems like that should be changed as well to match don't you think?

gillamkid:
> If you look at QGCButton you'll see that `hoverEnabled` is disabled on mobile builds. This works around a problem where the hover would get stuck on for android. Since there really isn't "hover" on touch systems this fixes the problem without really changing any ui behavior.
> 
> So for any control which uses a raw `Button` as base and has had hover support added to it this same `hoverEnabled` things needs to be there.
> 
> These changes affect all button controls except QGCButton. Seems like that should be changed as well to match don't you think?

@DonLakeFlyer, I made the changes you suggested to set `hoverEnabled` and make `QGCButton` have the same hover behavior. Should be good to go now!

gillamkid:
@DonLakeFlyer, this is ready for a second round of review! I made the changes you suggested but haven't added anything else. After I added my fixes I re-did the commits so the fixes could be squashed into the commit they originated from.

Let me know what your preferred fix strategy is. If you'd prefer me to not squash and keep the fix commits separate for easy second round review I could do that next time

mrpollo:
As discussed on the PX4 dev call @gillamkid please split this into atomic changes so we can get reviewed, thank you for your contribution!

DonLakeFlyer:
@gillamkid Closed because you are going to make new pull?

PR #11797, opened 2024-08-21
Terrain: Cleanup Terrain Queries

HTRamsey:
Clean up and better structure Terrain files in prep for #11795
Fixes some QObject parenting
Finishes setting up unit tests

### DIFF
--- a/src/QtLocationPlugin/ElevationMapProvider.cpp
+++ b/src/QtLocationPlugin/ElevationMapProvider.cpp
@@ -13,28 +13,28 @@
  ****************************************************************************/
 
 #include "ElevationMapProvider.h"
-#include "TerrainTile.h"
+#include "TerrainTileCopernicus.h"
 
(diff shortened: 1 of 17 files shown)


DonLakeFlyer:
Are you running some sort of lint thing on stuff before you checkin? Removing tabs and moving around the foo* bar to foo *bar generates all sorts of unnecessary diffs in the commit.

DonLakeFlyer:
> I don't but it would be really nice to get everything standardized because with so many different contributors things have gotten to be a mess so I wanted to do it right and clean when I work on things.

Ok, all good then. I'll change to be closer to what most people like.

DonLakeFlyer (approved):
Are you going to pull out the old AirMap stuff in separate pull?

HTRamsey:
> Are you going to pull out the old AirMap stuff in separate pull?

That or convert it to Copernicus if I can. Not sure why some stuff was converted to Copernicus but not everything.

PR #11809, opened 2024-08-26
Fix broken support links

gillamkid:
[](url)

Description
-----------
I noticed the "Discussion/Support" link in the root README.md had an outdated link. I fixed all occurances of the outdated link. 

https://github.com/user-attachments/assets/e2cf9ffd-069b-4af1-ae9f-3d30020b74b2

There was also a broken link should have been a link to a different page. In this video I show a fix for the broken link, and then the actual link that it is changed to in this PR (the "Console Logging" page instead of the "Reporting Bugs" page)

https://github.com/user-attachments/assets/3bc708b0-d752-4b33-8287-4e44721f8bd0

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -7,7 +7,7 @@ assignees: ''
 
 ---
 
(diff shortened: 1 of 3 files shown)


mrpollo:
Thank you

PR #11813, opened 2024-08-27
AnalyzeView: Fix GeoTagWorker Threading

### LINKED ISSUE #8488: Improved Geo-Tagging
Current Geo Tagging does not process misfires.
Geotagging should be timestamp based (deltas, not straight)

This implementation would work much better: https://github.com/PX4/Firmware/blob/master/Tools/geotag_images_ulog.py

### LINKED ISSUE #11858: GeoTagPage QGCFileDialog Issue
I believe the word 'folder' got mistakenly switched to file, that or the signal file was supposed to be used

qrc:/qml/GeoTagPage.qml:104: Error: Cannot assign [undefined] to QString
Screenshot from 2024-09-08 15-06-58
Screenshot from 2024-09-08 15-07-22

HTRamsey:
Remove QThread inheritance and fix QObject parenting
Fixes #11858
Closes #8488 by using timestamps
Also fixes a couple potential logic issues in corner cases
Pending a fix in ExifParser from #11906 first
I still need one really good log source for testing


src/AnalyzeView/GeoTagController.cc
```diff
@@ -8,120 +8,172 @@
...
 {
-    connect(&_worker, &GeoTagWorker::progressChanged,   this, &GeoTagController::_workerProgressChanged);
-    connect(&_worker, &GeoTagWorker::error,             this, &GeoTagController::_workerError);
-    connect(&_worker, &GeoTagWorker::started,           this, &GeoTagController::inProgressChanged);
-    connect(&_worker, &GeoTagWorker::finished,          this, &GeoTagController::inProgressChanged);
+    // qCDebug(GeoTagControllerLog) << Q_FUNC_INFO << this;
+
+    _worker->moveToThread(_workerThread);
+
+    (void) connect(_worker, &GeoTagWorker::progressChanged, this, &GeoTagController::_workerProgressChanged);
```
DonLakeFlyer, line 30: Is there something new going on with connect which required the (void) now?

DonLakeFlyer (approved):

src/AnalyzeView/GeoTagController.cc
HTRamsey: It's just a pattern carried over from my embedded work. I like to denote that I am intentionally ignoring return values rather than being unaware if it has one or not, helps with collaborating so others know that too

HTRamsey:
@bkueng do you have any sample logs/pics I can use for the unit test?

bkueng:
Not at hand. But if you have a PX4 SITL setup, there's a gazebo classic plugin here: https://github.com/PX4/PX4-SITL_gazebo-classic/blob/main/src/gazebo_camera_manager_plugin.cpp. It's used by the [cam_plane](https://github.com/PX4/PX4-SITL_gazebo-classic/tree/c0e0751341d46108377bbae2ae1bb6da8a5d4106/models/plane_cam) model. I never used it myself though.

PR #11824, opened 2024-08-29
VideoManager: Get the QtMultimedia Alternative Backend Working

HTRamsey:
Will be used as a backup for issues with GStreamer


src/API/QGCCorePlugin.h
```diff
@@ -110,8 +110,6 @@ class QGCCorePlugin : public QGCTool
     /// Allows the plugin to override the creation of the root (native) window.
     virtual void createRootWindow(QQmlApplicationEngine* qmlEngine);
 
-    /// Allows the plugin to override the creation of VideoManager.
```
DonLakeFlyer, line 113: Isn't this going to hose custom builds folks who already use this?

HTRamsey: I debated on leaving it, but it's not used anywhere so it does absolutely nothing right now

DonLakeFlyer: > but it's not used anywhere so it does absolutely nothing right now

Ok. Was it always like that or was it changed somewhere along the way. Guess we can remove and see if someone yells. I think some amount of this stuff was added by Auterion ages ago for their custom QGC. But they are so far forked from upstream they can never come back, so no longer important.

DonLakeFlyer (approved):

PR #11835, opened 2024-09-04
Fix light/dark toggle button on "Palette Test" page

gillamkid:
before
before-light-mode

after
after-light-dark


HTRamsey:
Somewhat unrelated to this PR but just a general comment, I feel like QGCPalette should be a singleton or something instead of having to have QGCPalette { id: qgcPal } everywhere

mrpollo:
Do we want to fix this now? Or can we merge this @HTRamsey?

PR #11836, opened 2024-09-05
Gillamkid/hover highlight

gillamkid:
## I did 2 things in this PR

1. Slightly highlight buttons on hover
* Before buttons would only change color when pressed or checked, but had no visual response when hovered. Now buttons will will have a faint highlight when hovered and a full highlight when pressed or checked

2. Make button icon colors same as text color
    
    Some of the icons where button color (gray), but i changed that so the icon color would be the same as the text color in the button. This is more similar to Google/Apple button design. It is also better for when the color Palette is overwritten with a different color scheme where button where buttonText and buttonHighlight might not look good when mixed together.
    
## before/after QGCButton

https://github.com/user-attachments/assets/adb727fd-c97c-40c4-8d11-d2b9e669c71a

https://github.com/user-attachments/assets/effd78d4-6973-4d3f-a1ad-ffe9541eeef3

## before/after app settings menu

https://github.com/user-attachments/assets/d8d5c23d-05c9-4be7-bdc7-9f95276c804c

https://github.com/user-attachments/assets/c180b997-cde6-4bd7-8f5a-88e5024c2267

## before/after menu buttons

https://github.com/user-attachments/assets/7b54948b-bd19-458c-8022-c46e29e3d555

https://github.com/user-attachments/assets/5e42e054-6f8a-4882-8218-b330e80b6b97

## icon color change justification
make it more similar to google/apple icon buttons

### google icon buttons
Screenshot from 2024-09-05 09-50-25

### apple icon buttons
Screenshot from 2024-09-05 09-49-05

### DIFF
--- a/src/QmlControls/QGCButton.qml
+++ b/src/QmlControls/QGCButton.qml
@@ -30,7 +30,7 @@ Button {
     property alias wrapMode:            text.wrapMode
     property alias horizontalAlignment: text.horizontalAlignment
 
-    property bool   _showHighlight:     enabled && (pressed | hovered | checked)
+    property bool   _showHighlight:     enabled && (pressed | checked)
 
     property int _horizontalPadding:    ScreenTools.defaultFontPixelWidth
     property int _verticalPadding:      Math.round(ScreenTools.defaultFontPixelHeight * heightFactor)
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
Perfect thanks

gillamkid:
@funyarinpa01, good catch, fixed in this new PR
https://github.com/mavlink/qgroundcontrol/pull/11874/commits/f029418cefda1ef97f35e481830a7ba5b2051a30

PR #11845, opened 2024-09-06
Mavlink: Improve Image Transmission Protocol Handling

HTRamsey:
- Fix parenting so ImageProtocolManager is correctly destroyed
- Add starting/ending of image stream, which could be added as buttons to OpticalFlowSensor.qml
- Remove discontinued PX4Flow as it was largely an unused feature and connecting Optical Flow sensors to a GCS via serial has little real-world use. No other Optical Flow sensors have serial links
- Fix working with other sensors other than PX4Flow
- Fix working with multiple vehicles

### DIFF
--- a/custom-example/qgroundcontrol.qrc
+++ b/custom-example/qgroundcontrol.qrc
@@ -76,7 +76,7 @@
 		<file alias="PlanViewSettings.qml">../src/UI/preferences/PlanViewSettings.qml</file>
 		<file alias="PlanViewToolBar.qml">../src/UI/toolbar/PlanViewToolBar.qml</file>
 		<file alias="PreFlightCheckList.qml">../src/FlightDisplay/PreFlightCheckList.qml</file>
-		<file alias="PX4FlowSensor.qml">../src/VehicleSetup/PX4FlowSensor.qml</file>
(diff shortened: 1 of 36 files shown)


HTRamsey:
@DonLakeFlyer is it worth it to expand the flow sensor capabilities in this? I don't actually know how popular that functionality is. I noticed it didn't work with multiple vehicles so I fixed that, and also there's no way of detecting remote mavlink devices (other than px4flow plugged in locally through serial). There's also a bunch of parameters than are available: https://docs.px4.io/v1.11/en/sensor/px4flow.html

DonLakeFlyer:
How long ago did the PX4 Flow disappear?

HTRamsey:
> How long ago did the PX4 Flow disappear?

I believe it was discontinued quite some time ago and compatibility was removed from PX4 entirely like 3 years ago. I also saw there hasn't really been a single mention of the PX4Flow stuff in QGC anywhere online so I didn't think it was worth the extra complexity it added to the link and firmware stuff.

Also forgot to mention this fixes the issue that the image handling stuff only worked with serial connected PX4Flow devices, and not with other OpticalFlow devices connected to autopilots. Also now works with multiple vehicle connections.

DonLakeFlyer:
Ok, all good then

DonLakeFlyer (approved):

PR #11846, opened 2024-09-06
FollowMe: Re-enable for Ardupilot

HTRamsey:
Reenable FollowMe Component in Ardupilot for debug builds
Added Unit Test


DonLakeFlyer:
Is this just for testing then for debug only?

HTRamsey:
> Is this just for testing then for debug only?

I am going to be doing a lot of live tests with large drones so I can make sure everything is reliable/safe before enabling it for everybody else

DonLakeFlyer:
I think I had mentioned this before. I vaguely remember: ArduPilot needs to good heading to do follow me. And I couldn't get a reasonable heading from a tablet.

HTRamsey:
Yup I have a plan for that. I want to expand #11722 to link into the positionmanager. With the updated android serial driver you could plug in an external gps and use that for the following.

PR #11856, opened 2024-09-08
CMake: Reduce MacOS package size

HTRamsey:
GStreamer had a lot of static libraries it didn't need. The next step would be to only copy the required libraries.


DonLakeFlyer:
Yeah, we had the sort of thing before

DonLakeFlyer (approved):

PR #11880, opened 2024-09-14
RTK/GPS toolbar information now shown in single indicator

### LINKED ISSUE #11859: Both RTK and GPS Status Indicators are exactly the same
I think maybe there should be some kind of difference between the two, or maybe you transform the GPS indicator image into the RTK indicator image when the RTK is connected

Screenshot from 2024-09-08 15-29-58

Screenshot from 2024-09-08 15-30-12

DonLakeFlyer:
Fix #11859

No vehicle connected:
Screenshot 2024-09-14 at 11 54 12 AM

Vehicle connected:
Screenshot 2024-09-14 at 11 55 28 AM


HTRamsey:
@DonLakeFlyer did it not show up twice for you? 
image

DonLakeFlyer:
I think I screwed up taking out some hacks I had in there for testing. I'll fix it.

PR #11881, opened 2024-09-15
Enhance battery indicator visuals with new SVGs and color coding

Paschalis:
- Added new SVG images for different battery levels: Critical, Low, Orange, Yellow, YellowGreen, and Green.
- Updated QGCPalette to include color definitions for these new states.
- Modified BatteryIndicator.qml to incorporate the new visual elements.
- Updated the resource file (qgcimages.qrc) to include the new SVG images.

The following GIFs demonstrate the changes:
Before:  
before
After:  
after

Or you can click on the videos below to see the changes:
Before: [before.webm](https://github.com/user-attachments/assets/051669b9-f435-419b-aec4-3cb13a8d9f98)
After: [after.webm](https://github.com/user-attachments/assets/8543f40f-cb87-4a02-91ba-6bb525d693de)


zdanek (approved):
I like it very much. Gives a nice fresh touch and it's more informative.
One thing, are shades of each colors same / according to existing palette?
At recording they look a bit like from shifted shades palette, not similar to what we have now.

Paschalis:
Thank you for the feedback! To maintain consistency with the existing palette, I used the following predefined colors from QGCPalette:

- Green: `qgcPal.colorGreen`
- Yellow-Green: `qgcPal.colorYellowGreen` (`#9acd32`) – **added to the palette**
- Yellow: `qgcPal.colorYellow` (`#ffff00`) – **added to the palette**
- Orange: `qgcPal.colorOrange`
- Red: `qgcPal.colorRed`

I’ve now updated the code to use `qgcPal.colorYellowGreen` instead of the hex value directly. The SVG images might appear darker due to their design, but they follow these updated palette colors. Let me know if any further adjustments are needed!

Light:
[light.webm](https://github.com/user-attachments/assets/99cc7399-b5bb-449c-84f2-fffde59991b8)

Dark:
[dark.webm](https://github.com/user-attachments/assets/ad4f53bd-0c2f-4413-aec0-c20ebcd76db6)

TSC21:
@gillamkid FYI

zdanek:
@Paschalis dark mode is great. But in light mode, those yellow, green and orange are hard to read.
Don't take it personally, let's look for best solution.

What about black border? Or something else? 

We are very close 😊 Can you experiment with something that will allow icons to be visible on white background, in full sun? Remember that often people use QGC in the field 😅

Paschalis:
Hi @zdanek ,

Thanks for the feedback!

Using darker colors will improve readability and contrast better than adding a border. Since the other indicators don’t have borders, adding one only to the battery indicator could disrupt visual consistency. Darker colors will maintain design uniformity and enhance visibility in bright conditions.

[Light.webm](https://github.com/user-attachments/assets/3aec1ebb-18ee-4c62-8d9a-ea6d2c145e89)

DonLakeFlyer:
In general coloring text like this is a bad idea. I just makes it too hard to read in direct sunlight on things like a Herelink or tablet controller. The icon coloring is great. But I'd leave the text color alone. That is optimized for sunlight readability.

DonLakeFlyer:
My other concern is the 80/60/40/20 coloring breaks. I seem to remember QGC having a talking battery warning ages ago with breaks like this and it caused concern. Can't remember if that's still in the code base. I would run that by folks on Discord. I'm guessing the first request will be to be able to make those configurable since different batteries discharge at differing rates. Also I'm guessing there will be questions about how thee values relate to the ArduPilot multi-level warning parameters thingies for battery.

zdanek:
Regarding thresholds, I also immediately thought about users' preferences since, people have their own needs and batteries can be non linear in discharge, so one might want to have own levels to be aware of.

Paschalis:
[Light.webm](https://github.com/user-attachments/assets/0af8c554-1ab4-49e9-b580-36b91f6485fa)

[Dark.webm](https://github.com/user-attachments/assets/0bb9ffe0-23c1-45ed-b140-3e92d1c0dc0b)

I'll proceed with implementing the changes to make the battery level thresholds configurable.

DonLakeFlyer:
> why might the upper percentage thresholds be a concern in terms of visibility or readability?

They are only a concern with visibility/readability on the text. The text should not be colored, the icons are fine.

> which are intended to give users a more nuanced view of battery health, particularly in non-critical ranges.

As it currently stands QGC shows:
* Text color when the battery is reported as MAV_BATTERY_CHARGE_STATE_OK
* Orange for when the battery is reported as MAV_BATTERY_CHARGE_STATE_LOW
* Red for when the battery is reported as MAVLink.MAV_BATTERY_CHARGE_STATE_CRITICAL and worse
* The battery state at which these transitions occur is controlled by the firmware through various parameters.
* I think ArduPilot even allows this setup to be different on each individual battery
* The parameters used differ between firmwares
* Using PX4 as an example the default setting for low battery state is at 15% and critical is at 7%. So the battery color will turn orange at 15% and red at 7%. These settings are global to all batteries

With this new pull the battery display for orange color is hardcoded to the 21%-40% remaining battery range for all vehicles and all batteries on the vehicle. And then red coloring is hardcoded to 20% and below. There is no distinction between Low and Critical states. The hardcoded 20% red trigger sort of relates to mavlink low state. But as you can see in the P4 example, by default it does not match PX4's concept of low battery. It also does not track the vehicles concept of low battery which is configurable.

Although the concept of more nuanced coloring of battery state is good thing the fact that it does not respect what the vehicle considers to be low and critical states is a pretty bad thing.

In order to make this work correctly there needs to be a combination of the two approaches. The telemetry controls the state from MAV_BATTERY_CHARGE_STATE_LOW and worse and the orange to red band of coloring. And then you have a nuanced coloring when battery state is reported as MAV_BATTERY_CHARGE_STATE_OK. The problem is that means you need to know what the percentage is when the transition is made from MAV_BATTERY_CHARGE_STATE_OK to MAV_BATTERY_CHARGE_STATE_LOW.

This would require firmware specific methods in FirmwarePlugin which could report back the percentage associated with the MAV_BATTERY_CHARGE_STATE_LOW transition such that the code can be written correctly. The problem with implementing something like that is that I don't think given the current state of affairs with mavlink and firmware implementations that is possible in all cases. In looking at the parameters in PX4 and ArduPilot it looks like it would be possible for PX4 since the parameters for controlling this are represented as percentages. But for ArduPilot the parameters which control low battery are represented in either milli-amps or voltages, not percentages. Maybe someone familiar with ArduPilot can chime in with some other way to do it?

So it's doable for PX4 but not for ArduPilot. Which then means it would need to be controlled by a supported bit in FirmwarePlugin. We do that for other QGC features which can only be supported by one firmware. So it's not out of the ordinary.

TLDR summary - Can be done correctly for PX4, but not currently (as far as I can see) for ArduPilot.

Davidsastresas:
I second what Don said on the last sentence.

For Ardupilot we really would want to get the low and critical values from parameters, and maybe have one additional medium point which could be configurable. But leaving it hard coded or configurable in so many levels is going to be confusing for Ardupilot users, so I would leave this particular to PX4 for the moment, if PX4 friends are fine with it.

I think it would go as creating some new q_properties and getters in firmware plugin and overriding in PX4 firmware plugin, or alternatively creating a whole toolbar indicator particular to PX4 and another for Ardupilot, and move battery indicator from tool indicator to vehicle indicators group ( so we can specify the qml from each firmware plugin ).

Thanks!

DonLakeFlyer:
> For Ardupilot we really would want to get the low and critical values from parameters,

@Davidsastresas You know ArduPilot better than I do. Do you see a way for QGC to know the battery percentage values for low given the set of parameters available?

Paschalis:
### This update enhances the battery indicator with the following changes:

1. **Battery State Coloring**:
   - Orange for `MAV_BATTERY_CHARGE_STATE_LOW`.
   - Red for `MAV_BATTERY_CHARGE_STATE_CRITICAL`.
   - Coloring respects the vehicle's telemetry thresholds for PX4.
2. **Function Update**: 
   - The function `getBatteryColor()` was updated to apply different color logic depending on whether the **dark** or **light** theme is used. 
   - In the dark theme, the function uses thresholds to set color gradients for better visibility, applying `qgcPal.colorGreen`, `qgcPal.colorYellowGreen`, and `qgcPal.colorYellow`.
   - In the light theme, a default text color is returned for non-critical states.
3. **New Icons**: Added separate battery configuration icons for light and dark themes (`BatteryConf.svg` and `BatteryConfLight.svg`).
4. **Configurable Thresholds**: Users can now configure battery thresholds.

The following images demonstrate the changes and how users can configure the thresholds, along with GIFs showcasing the battery indicator in both light and dark themes:

[Dark.webm](https://github.com/user-attachments/assets/15f36c77-3b3d-48c5-887a-63e5de661c80)

Screenshot from 2024-09-20 22-02-16

[Light.webm](https://github.com/user-attachments/assets/5c37cece-bf5a-4da8-aee3-7189c0d5445a)

Screenshot from 2024-09-20 22-02-30

DonLakeFlyer:
Ok, this is getting much, much better. Thanks. Comments:
* I still don't think the text for the battery display should be colored. As I said above that just kills readability of the actual values out in the field.
* You've got visual indications for 7% and 15% as hardcoded values. In reality you can't show those numbers you can only refer to "low" and "critical" which is I think what you are trying to indicator here. At this point in the code you don't know what the percentage is for low and critical.
* I don't quite understand the need for light and dark theme differences in coloring. In reality "light" is meant for outdoor, direct sunlight use. On things like tablets or hirelings. And "dark" tends to work better out of direct sunlight on things like laptops sitting on a table say. At least that's the concept between the two. My take is that if you can find something that is visible in direct sunlight then that is fine for all usages.

Paschalis:
Current Default Master Version: Image and Text Coloring for Both Light and Dark Themes
```
function getBatteryColor() {
                switch (battery.chargeState.rawValue) {
                case MAVLink.MAV_BATTERY_CHARGE_STATE_OK:
                    return qgcPal.text
                case MAVLink.MAV_BATTERY_CHARGE_STATE_LOW:
                    return qgcPal.colorOrange
                case MAVLink.MAV_BATTERY_CHARGE_STATE_CRITICAL:
                case MAVLink.MAV_BATTERY_CHARGE_STATE_EMERGENCY:
                case MAVLink.MAV_BATTERY_CHARGE_STATE_FAILED:
                case MAVLink.MAV_BATTERY_CHARGE_STATE_UNHEALTHY:
                    return qgcPal.colorRed
                default:
                    return qgcPal.text
                }
            }
```
Our new Version retains the above approach !
```
function getBatteryColor() {
    switch (battery.chargeState.rawValue) {
        case MAVLink.MAV_BATTERY_CHARGE_STATE_OK:
            if (qgcPal.globalTheme === QGCPalette.Dark) {
                // Apply the battery color logic only for the dark theme
                if (!isNaN(battery.percentRemaining.rawValue)) {
                    if (battery.percentRemaining.rawValue > threshold1) {
                        return qgcPal.colorGreen 
                    } else if (battery.percentRemaining.rawValue > threshold2) {
                        return qgcPal.colorYellowGreen 
                    } else {
                        return qgcPal.colorYellow 
                    }
                } else {
                    return qgcPal.text
                }
            } else {
                // For light theme, return a default color
                return qgcPal.text
            }
        case MAVLink.MAV_BATTERY_CHARGE_STATE_LOW:
            return qgcPal.colorOrange
        case MAVLink.MAV_BATTERY_CHARGE_STATE_CRITICAL:
        case MAVLink.MAV_BATTERY_CHARGE_STATE_EMERGENCY:
        case MAVLink.MAV_BATTERY_CHARGE_STATE_FAILED:
        case MAVLink.MAV_BATTERY_CHARGE_STATE_UNHEALTHY:
            return qgcPal.colorRed
        default:
            return qgcPal.text
    }
}
```

> Ok, this is getting much, much better. Thanks. 

Appreciate your feedback and support !

>      *I still don't think the text for the battery display should be colored. As I said above that just kills readability of the actual values out in the field.

The current behavior in the master branch actually changes both the image and text color based on the battery state. My update retains this approach !

>     * You've got visual indications for 7% and 15% as hardcoded values. In reality you can't show those numbers you can only refer to "low" and "critical" which is I think what you are trying to indicate here. At this point in the code you don't know what the percentage is for low and critical.

The values aren't hardcoded; I'm following the default approach by using MAVLink.MAV_BATTERY_CHARGE_STATE_LOW for orange and MAVLink.MAV_BATTERY_CHARGE_STATE_CRITICAL (and others) for red. The color changes are tied to these MAVLink states rather than specific percentages. However, I can update the display to show 
- 'low'   
- 'critical'

 instead of percentage values for clarity !

>     * I don't quite understand the need for light and dark theme differences in coloring. In reality "light" is meant for outdoor, direct sunlight use. On things like tablets or Herelink type controllers. And "dark" tends to work better out of direct sunlight on things like laptops sitting on a table. At least that's the concept between the two. My take is that if you can find something that is visible in direct sunlight then that is fine for all usages.

I completely understand and agree with the concept you described for light and dark themes, which is why I've kept the light theme the same as in the current master version—using Black, Orange, and Red for both the image and text, as these colors work well in outdoor, direct sunlight conditions. Since no one has raised any issues with the current coloring in the light theme, I've decided to keep it as is for now.

DonLakeFlyer:
So I built and ran this to see what it looks likes. Comments:
* Battery coloring from dark theme should be used in all cases
* Text shouldn't be colored
* The setup for defining the percentages isn't super clear. For examples the icons shown in there don't match what is actually shown in the toolbar. I think something along these lines would be better:
  * Screenshot 2024-09-21 at 10 55 31 AM
  * Basic idea is to show the actual different toolbar icon states with what they mean: 100% - Config 1 - Config 2 - Low - Critical
  * I just drew this in Gimp to give an idea. Not fully fleshed out, but hopefully enough to give a directional idea.
  * I think it ends up being more understandable as to what is going on

DonLakeFlyer:
Also can you add a note in ChangeLog.md about this new feature?

Paschalis:
Screenshot from 2024-09-22 01-08-29
Screenshot from 2024-09-22 01-08-14

DonLakeFlyer:
Love it. This is fantastic stuff. I'll do a full code review soon.

Davidsastresas:
> > For Ardupilot we really would want to get the low and critical values from parameters,
> 
> @Davidsastresas You know ArduPilot better than I do. Do you see a way for QGC to know the battery percentage value for low state given the set of parameters available?

Sure! In order to know the low and critical values for voltage we have:
[BATT_LOW_VOLT](https://ardupilot.org/copter/docs/parameters.html#batt-low-volt)
[BATT_CRT_VOLT](https://ardupilot.org/copter/docs/parameters.html#batt-crt-volt)

and in order to know the low and critical values for current we have:
[BATT_LOW_MAH](https://ardupilot.org/copter/docs/parameters.html#batt-low-mah)
[BATT_CRT_MAH](https://ardupilot.org/copter/docs/parameters.html#batt-crt-mah)

and to understand the total capacity of the battery we have:
[BATT_CAPACITY]https://ardupilot.org/copter/docs/parameters.html#batt-capacity-battery-capacity

Please note we have a BATTX set of parameters for each battery monitor configured, so all the above parameters will exist for each instance of battery monitor configured.

Considering the above, the logic would go as taking BATT_CAPACITY as 100% battery, and then understand the percentages for Low and Critical from BATT_LOW_MAH and BATT_CRT_MAH.

I don't mind to work on this part for AP, but it would be good to leave it prepared so the widget takes this info from FirmwarePlugin.

@Paschalis that work looks very nice so far, thanks! However, it would be good if you could do a rebase -i, and arrange the commits so we don't have those "merge into master" commits. Whenever you want to rebase, if you do a git rebase upstream/master instead of git merge, your commits will organize on top of upstream/master, instead of having that odd "merge into master" commit. It makes much easier to track commits in a PR. Thanks!

src/UI/toolbar/BatteryIndicator.qml
```diff
@@ -124,7 +152,7 @@ Item {
                 anchors.bottom:     parent.bottom
                 width:              height
                 sourceSize.width:   width
-                source:             "/qmlimages/Battery.svg"
+                source:             getBatterySvgSource()
                 fillMode:           Image.PreserveAspectFit
                 color:              getBatteryColor()
```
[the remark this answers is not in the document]
DonLakeFlyer: Did you verify this?

src/UI/toolbar/BatteryIndicator.qml
```diff
@@ -278,6 +322,117 @@ Item {
...
+                        height: ScreenTools.defaultFontPixelHeight * 5
+                        width: ScreenTools.defaultFontPixelWidth * 7
+                        fillMode: Image.PreserveAspectFit
+                        color: qgcPal.colorYellowGreen
+                    }
+
+                    FactTextField {
+                        id: threshold1Field
+                        fact: batterySettings.threshold1
+                        validator: IntValidator { bottom: 16 } // Value is at least 16
```
DonLakeFlyer, line 361: This should be done by specifying min in the FactMetaData. Then you get automatic validation.

src/UI/toolbar/BatteryIndicator.qml
```diff
@@ -278,6 +322,117 @@ Item {
...
+                        implicitWidth: ScreenTools.defaultFontPixelWidth * 6  // Reduced width
+                        Layout.leftMargin: -10
+                        //Layout.preferredWidth: ScreenTools.defaultFontPixelWidth * 2  // Reduced width using Layout.preferredWidth
+                        height: ScreenTools.defaultFontPixelHeight * 1.5
+                        onEditingFinished: {
```
DonLakeFlyer, line 366: Then you do final level of validation on a rawValueChanged signal to adjust if needed.

src/UI/toolbar/BatteryIndicator.qml
```diff
@@ -278,6 +322,117 @@ Item {
...
+                        height: ScreenTools.defaultFontPixelHeight * 5
+                        width: ScreenTools.defaultFontPixelWidth * 7
+                        fillMode: Image.PreserveAspectFit
+                        color: qgcPal.colorGreen
+                    }
+
+                    QGCLabel {
+                        Layout.preferredWidth: ScreenTools.defaultFontPixelWidth * 3.5 
+                        text: qsTr("100%")
+                        Layout.leftMargin: -13
```
DonLakeFlyer, line 347: You should never use fix pixels as margins. Everything needs to be based on the current font. Otherwise the ui won't scale to font size changes.

The way to lay this out is as follows:
* Each battery icon plus either fixed percentage or text field should be in its own RowLayout. With the spacing on the layout set to something like font width times 1 or 2 depending on how it looks.
* Then there is a top level RowLayout which has all these RowLayouts as children. And then there is a font based spacing on that to space out the battery+pct visuals.

src/UI/toolbar/BatteryIndicator.qml
```diff
@@ -278,6 +322,117 @@ Item {
                         }
                     }
                 }
+                
+            }
+
+            SettingsGroupLayout {
+                heading:        qsTr("Battery State Display")
+                Layout.fillWidth: true
+                spacing: ScreenTools.defaultFontPixelWidth * 0.5
```
DonLakeFlyer, line 331: The visibility of this group should be based on `batterySettings.visible`. This allows custom builds to hide this section if they want to.

Also each individual settings has a `visible` property.  If not visible then the editing field would normally be completely hidden. But in this caseI would show just a label with the percentage instead of a configurable text field.

Both of these combined give a good level of custom build override capability.

DonLakeFlyer (changes requested):

DonLakeFlyer:
Ok, so then with what @Davidsastresas explains we could plumb FirmwarePlugin to tell us what the battery percentage is for both the Low and Critical states. This would allow us to do the following things:
* Show percentage values in the Low and Critical battery visual display which I think would be helpful
* Add better validation on the thresholds so that they aren't set to things which don't make sense. Specifically threshold2 should not be below the Low percentage.

I think this would be nice addition to this work, but I would deal with it in a separate pull after this goes through.

DonLakeFlyer:
Also after writing up my last comment which include talking about threshold percentage validation it go me thinking about the fact that threshold validation is being done in ui code. In reality this is incorrect. Since these settings can be changed from the C++ side of things as well. These sort of things happen with custom builds.

The better way to do it would be to use the SettingsGroup and FactMetaData systems to do the validation fully on the C++ side. @Paschalis This isn't that hard, but requires knowledge of both those systems. If you want to do it then I can explain how. If not (since you've already put a ton of effort into this) I can do this myself after this gets merged in with the way it is now.

src/UI/toolbar/BatteryIndicator.qml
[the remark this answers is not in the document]
Paschalis: Yes, I verified that the coloring works properly, even though the SVG icons are already colored. The icons display correctly in both Light and Dark themes. I also tested the color application from the palette, and the correct colors show up as expected in both themes without any issues.

https://github.com/user-attachments/assets/21076fcc-a4d6-4210-a480-1fdf844abe6d

https://github.com/user-attachments/assets/1514ad9c-c4be-41b8-9208-9fe356c9a004

src/UI/toolbar/BatteryIndicator.qml
Paschalis: I appreciate your guidance, and I've implemented the layout adjustments as you suggested. Each battery icon, along with its corresponding percentage or text field, is now placed in its own `RowLayout`. I've also set the spacing based on the current font width, ensuring that the UI scales properly with font size changes. The top-level `RowLayout` contains all these individual layouts with appropriate font-based spacing.

src/UI/toolbar/BatteryIndicator.qml
Paschalis: I’ve implemented the validation fully on the C++ side in `BatteryIndicatorSettings.cc`. Now, in the QML, we simply use:

```qml
onEditingFinished: {
    batterySettings.setThreshold2(parseInt(text));
}
```

src/UI/toolbar/BatteryIndicator.qml
Paschalis: I've implemented the final level of validation in the `BatteryIndicatorSettings.cc`. For `threshold1`, I ensure that the value is always at least 16 and no more than 99. If the provided value is below 16, it gets adjusted to 17. If it exceeds 99, it's capped at 99. Additionally, if the new value is less than or equal to `threshold2`, it automatically gets set to one more than `threshold2` to maintain the required order. For `threshold2`, I enforce a minimum value of 16. If the user attempts to set it to 15 or less, it gets adjusted up to 16. Moreover, if the new value is greater than or equal to `threshold1`, it is set to one less than `threshold1` to ensure it remains less.

src/UI/toolbar/BatteryIndicator.qml
Paschalis: I’ve implemented the visibility controls as you suggested, using `visible: batterySettings.visible` for the entire group and `bool thresholdVisible: batterySettings.thresholdEditable` for individual settings.

In the `BatteryIndicatorSettings` class, I’ve added QSettings to manage visibility and editability, defaulting to `true` if no value is set:

```cpp
bool BatteryIndicatorSettings::visible() const {
    QSettings settings;
    return settings.value("ShowBatterySettings", true).toBool();
}

bool BatteryIndicatorSettings::thresholdEditable() const {
    QSettings settings;
    return settings.value("ThresholdEditable", true).toBool();
}
```
However, I’m unsure how to properly configure these two settings to effectively allow custom builds to hide this section if they choose to. Could you provide how to set this up correctly? Your expertise would be greatly appreciated!

Paschalis: In the current setup, if the `thresholdVisible` property is **false**, the `FactTextField` for `threshold1` and `threshold2` will be hidden, and a `QGCLabel` will display the current percentages. This ensures that users can still see the battery thresholds without the editing fields when they're not meant to be configurable.

Screenshot from 2024-09-23 21-15-09

If the `thresholdVisible` property is **true**, the `FactTextField` for both `threshold1` and `threshold2` will be visible, allowing users to edit the thresholds directly. 

Screenshot from 2024-09-23 21-23-33

DonLakeFlyer: > However, I’m unsure how to properly configure these two settings to effectively allow custom builds to hide this section

there is no extra work needed. SettingsGroup and SettingsFact base classes already support this automatically.

src/Settings/BatteryIndicatorSettings.h
```diff
@@ -19,4 +19,21 @@ class BatteryIndicatorSettings : public SettingsGroup
     DEFINE_SETTING_NAME_GROUP()
 
     DEFINE_SETTINGFACT(display)
-};
+    DEFINE_SETTINGFACT(threshold1)
+    DEFINE_SETTINGFACT(threshold2)
+
+    Q_PROPERTY(bool visible READ visible NOTIFY visibleChanged)
```
DonLakeFlyer, line 25: So all f this is incorrect:
* There is no need for adding a visible property. It already exists on both the group and on each setting. You can access them in Qml using `QGroundControl.settingsManager.batteryIndicatorSettings.visible` for visibility which should control whether to show the battery indicator panel. And then `QGroundControl.settingsManager.batteryIndicatorSettings.threshold1.visible` for the thresholds.
* To implement validation on the C++ side you need to implement you own settings fact creation using `DECLARE_SETTINGSFACT_NO_FUNC`. Look at AppSettings.cc for examples. Then in that code you connect to the rawValueChanged signal. And in you signal handler that is where you do the validation.

Paschalis: I streamlined the visibility management by utilizing existing properties in QGroundControl.settingsManager.batteryIndicatorSettings and removed redundant properties. Additionally, I implemented validation for threshold settings using DECLARE_SETTINGSFACT_NO_FUNC 

Screenshot from 2024-09-28 21-16-04

and confirmed the correct storage of new threshold values after verifying defaults from the JSON file.

Screenshot from 2024-09-28 22-54-15

Thanks!

DonLakeFlyer:
Sorry for falling off the face of the earth on this. I was out of town. This all looks great. I reviewed the code and it looks correct. I'm going to double-check it after it's merged in just to be sure in a lot more detail.

PR #11890, opened 2024-09-17
CMake: MacOS Deployment Fixes

### LINKED ISSUE #11860: CI: Fix occasional failure to build DMG.
image

Edit: Try Bundle & DragNDrop CPack?

HTRamsey:
Builds universal build for arm64 & x64
Fixes DMG by using APFS format which is supported MacOS 11+
fixes #11860


DonLakeFlyer:
I wonder about Universal builds. It creates a monster download/install size. Not sure if that a good thing?

HTRamsey:
I think it's 277MB vs 253MB right now right? I don't think that's too bad, especially since we are still including a bunch of gstreamer stuff that isn't necessary. But also we can leave the option and just set it OFF by default

DonLakeFlyer:
Previous stable is 133Mb though. So there is already something out of whack even before this. Maybe just gstreamer. Would be good to know what the real difference is.

HTRamsey:
Is this alright then or leave universal build disabled? Either way the other change fixing the dmg is important

DonLakeFlyer:
> yeah I just checked and the difference is almost entirely gstreamer.

all good then to leave this as universal binary

DonLakeFlyer (approved):

PR #11895, opened 2024-09-18
"Value Display" update

gillamkid:
# Description
This PR has 2 commits
1. Fix button sizes in TelemetryValuesBar
    * In QGC 4.4.2 these buttons look fine, but in the master branch they were super squashed looking
2. Organize "Value Display" so it is more scannable
    * Now it is organized with keys on the left and values/inputs on the right, making the form more intuitive and scannable for first time users.
    * The design of the form is now more similar to other parts of QGC, namely the views seen in "Application Settings"

## Sponsor

## Before

https://github.com/user-attachments/assets/92304f65-fdee-4206-aab8-cc6e3ed7a0cc

## After

https://github.com/user-attachments/assets/52c2dd33-761a-4f0a-8083-f2ba0afa5228

### DIFF
--- a/src/QmlControls/QGCRadioButton.qml
+++ b/src/QmlControls/QGCRadioButton.qml
@@ -42,7 +42,7 @@ RadioButton {
         color:              control.textColor
         opacity:            enabled ? 1.0 : 0.3
         verticalAlignment:  Text.AlignVCenter
-        leftPadding:        control.indicator.width + (_noText ? 0 : ScreenTools.defaultFontPixelWidth * 0.25)
+        leftPadding:        control.indicator.width + (_noText ? 0 : control.leftPadding)
     }
 
 }
--- a/src/QmlControls/SettingsGroupLayout.qml
+++ b/src/QmlControls/SettingsGroupLayout.qml
@@ -18,6 +18,7 @@ ColumnLayout {
     property string heading
     property string headingDescription
     property bool   showDividers:       true
+    property bool   showBorder:         true
 
     property real _margins: ScreenTools.defaultFontPixelHeight / 2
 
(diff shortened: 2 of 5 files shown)


mrpollo:
Changes look good to me, its a really nice improvement.

@DonLakeFlyer Can you do one last final check here? just want to make sure nothing is lost in the transition

DonLakeFlyer:
@gillamkid The mp4 seem to be gone now? Maybe just images showing before after changes?

DonLakeFlyer:
I had taken an initial look at this a bit ago. My concern was with the sizing of the buttons for add/remove columns/row. If they are too small and right next to each other they can be tough to touch with a fat finger on a tablet. I'll need to build it myself and look.

DonLakeFlyer:
> The mp4 seem to be gone now? Maybe just images showing before after changes?

Don't worry about this. I'm just building it myself to look at it.

DonLakeFlyer:
Hmm, this is different than I last saw it. I thought there were little button with images for add/remove row/columns. Anyway:

Here is current master. You can see how the buttons are really small. This is because it uses `minTouchPixels / 2` which in reality is incorrect usaget. Since it is smaller than what is touchable on a tablet. It also expose a bug in QGCButton which is not positioning the text correctly.
Screenshot 2024-09-18 at 11 16 34 AM

Here is what the editable panel looks like with this pull. My opinion is that those buttons end up being way too big and look really strange in this context:
Screenshot 2024-09-18 at 11 11 58 AM

Here is the same thing but using full minTouchPixels size which is the calculated minimum size of something which is touchable on a tablet. It also shows the text positioning bug in QGCButton which needs fixing. I think this looks way better and is still functional on a tablet from a sizing perspective.
Screenshot 2024-09-18 at 11 10 56 AM

src/QmlControls/QGCRadioButton.qml
```diff
@@ -42,7 +42,7 @@ RadioButton {
         color:              control.textColor
         opacity:            enabled ? 1.0 : 0.3
         verticalAlignment:  Text.AlignVCenter
-        leftPadding:        control.indicator.width + (_noText ? 0 : ScreenTools.defaultFontPixelWidth * 0.25)
+        leftPadding:        control.indicator.width + (_noText ? 0 : control.leftPadding)
```
DonLakeFlyer, line 45: Why this change? This is incorrect in that it doesn't adjust sizing based on the configurable font size.

DonLakeFlyer:
> * The design of the form is now more similar to other parts of QGC, namely the views seen in "Application Settings"

Yes, much better to head this direction.

Detailed comments...

Current master which you can see is kind of a mess of complexity and alignment problems:
Screenshot 2024-09-18 at 11 28 18 AM

Here is what this pull looks like. It's better from an alignment perspective. But for me all those divider lines make it look really noisy. Also the Range section needs conversion.
Screenshot 2024-09-18 at 11 33 12 AM

I think you need to take the group layout grouping further than a single group. This is how the new settings stuff works and it ends up looking much better. I also don't think there needs to be toggle visibility. And it groups things more logically. Here is a quick hack on top of your pull to show it heading more in the direction of multiple groupings. Certainly needs more work but I hope you get the idea:
Screenshot 2024-09-18 at 11 46 57 AM

src/QmlControls/SettingsGroupLayout.qml
```diff
@@ -18,6 +18,7 @@ ColumnLayout {
     property string heading
     property string headingDescription
     property bool   showDividers:       true
+    property bool   showBorder:         true
```
DonLakeFlyer, line 21: I'd leave this as is for now. Don't want to create a new UI paradigm of these group layout without borders until it is really needed. That is already mostly already covered by the SectionHeader control.

DonLakeFlyer (changes requested):

DonLakeFlyer:
If you want I can merge this in and then fix up the rest of the stuff. All simple, take a few minutes for me.

gillamkid:
> If you want I can merge this in and then fix up the rest of the stuff. All simple, take a few minutes for me. Thanks so much for cleaning this messy stuff up.

Yeah that'd be great! Not sure when I'd get to it

PR #11902, opened 2024-09-19
Utilities: Update Compression

HTRamsey:
Adds QGCZip for zipping & unzipping using built-in Qt methods
Adds unit tests
Allow finding zlib before building
code cleanup


DonLakeFlyer:
Is this gonna be used someplace though? Hate to carry around code which is never referenced.

HTRamsey:
yup, it's needed for #11795, the ardupilot terrain downloads are .zip's

DonLakeFlyer (approved):

PR #11906, opened 2024-09-19
AnalyzeView: Add Exiv2 for writing exif data

HTRamsey:
Fixes Exif Writing by adding Exiv2 dependency. This also allows much more capabilities in regards to geotagging too. Not sure whether to remove easyexif or not, at least until this has been more thoroughly tested.


DonLakeFlyer:
So readtime2 might be switched over to if testing proves it out?

HTRamsey:
> So readtime2 might be switched over to if testing proves it out?

Yeah, I can remove it now but I also intend to come back to exif stuff soon and expand a bit on it, so I can do it then too. Easyexif is just much easier to work with and is a very small lib but it only reads data.

DonLakeFlyer (approved):

DonLakeFlyer:
> but I also intend to come back to exif stuff soon and expand a bit on it

all good then

PR #11908, opened 2024-09-19
Font problems

### LINKED ISSUE #11876: OSX: qt.qpa.fonts: Populating font family aliases took 149 ms. Replace uses of missing font family "Open Sans Semibold" with one that exists to avoid this cost.
You get this on every startup

DonLakeFlyer:
* Fix #11876
* Fix incorrect usage of bold font
* Fix incorrect usage of fixed width font

There still is a problem here. But I can't find it and for now at least things that are supposed to be bold are actually bold and things that are supposed to be fixed are really fixed.
```qt.qpa.fonts: Populating font family aliases took 161 ms. Replace uses of missing font family "13" with one that exists to avoid this cost. ```


HTRamsey:
Question about the warnings one, isn't the default for all categories to silence Debug and info but output warnings and up? So it shouldn't really matter if it has a category

DonLakeFlyer:
I had forgotten about that. I'll put these back at some point then.

PR #11909, opened 2024-09-19
QtLocationPlugin: Minor Miscellaneous Improvements

HTRamsey:
- Move remaining file caching related stuff to QGeoFileTileCacheQGC to prep converting QGCMapEngine to a better threadWorker handler for  QGCCacheWorker
- Reduce total count of QNetworkAccessManager's, which each create a new thread per instance
- Translate output errors
- Network caching

### DIFF
--- a/src/QtLocationPlugin/QGeoFileTileCacheQGC.h
+++ b/src/QtLocationPlugin/QGeoFileTileCacheQGC.h
@@ -27,20 +27,31 @@ public:
     static quint32 getMaxDiskCacheSetting();
     static void cacheTile(const QString &type, int x, int y, int z, const QByteArray &image, const QString &format, qulonglong set = UINT64_MAX);
     static void cacheTile(const QString &type, const QString &hash, const QByteArray &image, const QString &format, qulonglong set = UINT64_MAX);
-    static QGCFetchTileTask* createFetchTileTask (const QString &type, int x, int y, int z);
-    static QString getCacheFilename() { return QStringLiteral("qgcMapCache.db"); }
+    static QGCFetchTileTask *createFetchTileTask(const QString &type, int x, int y, int z);
+    static QString getDatabaseFilePath() { return _databaseFilePath; }
+    static QString getCachePath() { return _cachePath; }
 
 private:
     // QString tileSpecToFilename(const QGeoTileSpec &spec, const QString &format, const QString &directory) const final;
     // QGeoTileSpec filenameToTileSpec(const QString &filename) const final;
 
+    static void _initCache();
+    static bool _wipeDirectory(const QString &dirPath);
+    static void _wipeOldCaches();
+
     static QString _getCachePath(const QVariantMap &parameters);
     static uint32_t _getMemLimit(const QVariantMap &Parameters);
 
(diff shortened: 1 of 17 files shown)


src/QtLocationPlugin/QGeoFileTileCacheQGC.h
```diff
@@ -18,20 +18,31 @@ class QGeoFileTileCacheQGC : public QGeoFileTileCache
     static quint32 getMaxDiskCacheSetting();
     static void cacheTile(const QString &type, int x, int y, int z, const QByteArray &image, const QString &format, qulonglong set = UINT64_MAX);
     static void cacheTile(const QString &type, const QString &hash, const QByteArray &image, const QString &format, qulonglong set = UINT64_MAX);
-    static QGCFetchTileTask* createFetchTileTask (const QString &type, int x, int y, int z);
-    static QString getCacheFilename() { return QStringLiteral("qgcMapCache.db"); }
+    static QGCFetchTileTask *createFetchTileTask(const QString &type, int x, int y, int z);
+    static QString getDatabaseFilePath() { return _databaseFilePath; }
+    static QString getCachePath() { return _cachePath; }
```
bkueng, line 23: Is this guaranteed to be initialized before it's first used?

src/QtLocationPlugin/QMLControl/QGCMapEngineManager.cc
```diff
@@ -46,7 +47,7 @@ QGCMapEngineManager::QGCMapEngineManager(QObject *parent)
 
     (void) connect(getQGCMapEngine(), &QGCMapEngine::updateTotals, this, &QGCMapEngineManager::_updateTotals);
 
-   _updateDiskFreeSpace();
+    // _updateDiskFreeSpace();
```
bkueng, line 50: Why comment this?

src/QtLocationPlugin/QGeoFileTileCacheQGC.h
HTRamsey: Actually it won't even get used if I remove _updateDiskFreeSpace as mentioned in the other comment

bkueng: Ok, I'd add a comment or an assertion to make sure a future refactoring does not break this.

HTRamsey: Got it, made it so that it had to be instantiated first to provide a parameter to MapEngine & added a comment. But also MapEngine will go away soon anyways.

HTRamsey: Also just FYI the TileCacheWorker is the last major thing to do. Currently it has to service up to 3-4 maps simultaneously while doing offline map downloads, sql queries, amongst other stuff which causes significant issues if you have multiple downloads going at once. So I will turn the map engine into a thread pool worker of some kind

bkueng: Sounds good

bkueng (approved):

PR #11912, opened 2024-09-20
CMake: Move Translations Output To Own Folder

HTRamsey:
Cleans up the build dir a bit. Also includes minor fix for windows cmake


gillamkid:
this seems to have broken translation imports. I made a fix here: https://github.com/mavlink/qgroundcontrol/pull/11934

PR #11916, opened 2024-09-20
Tools: Create File Header Updater Tool

HTRamsey:
Creates a tool to update file headers

Edit: Whoops I just realized some files are a slightly different format messing this up, fixing...


DonLakeFlyer (approved):
Nice thanks

PR #11927, opened 2024-09-22
Use icon for terrain button instead of "T"

gillamkid:
# Description

Makes so the button used to expand/collapse the terrain panel has an icon instead of the letter "T"

## Sponsor

# Before
before-closed
before-expanded

# After
after-closed
after-expanded


DonLakeFlyer:
Looks good, thanks

PR #11928, opened 2024-09-22
Fix terrain chart colors

gillamkid:
# Description

When the theme was set to outdoor mode, the terrain chart's lines/labels weren't ledgible colors. This fixes that

## Sponsor

# Before
before-terrain-light

# After
after-terrain-light

### Before
before-terrain-dark
### After
after-terrain-dark


DonLakeFlyer:
Yup, thanks

PR #11929, opened 2024-09-22
CI: Create Continuous Prerelease

HTRamsey:
This should upload commits to master to a continuous release making it easy for people to download daily builds

### DIFF
--- a/.github/actions/upload/action.yml
+++ b/.github/actions/upload/action.yml
@@ -43,3 +43,11 @@ runs:
       working-directory: ${{ runner.temp }}/shadow_build_dir/${{ inputs.source }}
       run: aws s3 cp ${{ inputs.artifact_name }} s3://qgroundcontrol/latest/${{ inputs.artifact_name }} --acl public-read
       shell: bash
+
+    - name: Create Continuous Release
+      if: ${{ github.event_name != 'pull_request' && github.ref_name == 'master' && !github.event.pull_request.head.repo.fork }}
+      uses: softprops/action-gh-release@v2
+      with:
+        prerelease: true
+        files: |
+          ${{ runner.temp }}/shadow_build_dir/${{ inputs.source }}/${{ inputs.artifact_name }}


HTRamsey:
@mrpollo Thoughts on this?

mrpollo:
This should work, but we must test it as soon as possible. Unfortunately, we can't test until we merge.

By the way, I'm also using the same action package for PX4 here: https://github.com/PX4/PX4-Autopilot/pull/23703.

.github/actions/upload/action.yml
```diff
@@ -43,3 +43,11 @@ runs:
       working-directory: ${{ runner.temp }}/shadow_build_dir/${{ inputs.source }}
       run: aws s3 cp ${{ inputs.artifact_name }} s3://qgroundcontrol/latest/${{ inputs.artifact_name }} --acl public-read
       shell: bash
+
+    - name: Create Continuous Release
+      if: ${{ github.event_name != 'pull_request' && github.ref_name == 'master' && !github.event.pull_request.head.repo.fork }}
```
mrpollo, line 48: Did you test `github.ref_name = 'master'`?  sometimes `ref_name` isn't what we want

mrpollo: I would skip the fork conditional, I don't its needed

mrpollo: Are you creating a release every time you push to master? Is this the desired behavior?

HTRamsey: Rather than every commit, should it be every night/week?

DonLakeFlyer: Daily builds upload to S3 on every merge so seems like it should match that.

DonLakeFlyer (approved):

.github/actions/upload/action.yml
DonLakeFlyer: Also seems like this is way ore important on automating Stables. This just kinda duplicates the S3 thing which is referenced by the docs. I guess some people go here to get releases though.

PR #11960, opened 2024-09-28
CMake: Find all packages in root CMakeLists.txt

HTRamsey:
Apparently there is an occasional issue for some catching all dependencies if you don't find them all in the root cmakelists. Also, disable QtMultimedia receiver backend by default until it is stable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,7 +177,10 @@ find_package(Qt6
         Xml
     OPTIONAL_COMPONENTS
         LinguistTools
+        MultimediaQuickPrivate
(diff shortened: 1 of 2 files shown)


msoos (approved):
Yay, I did a full-on delete `build`, rebuild via `cmake ..` and `make -j18` and this worked!! Nice. Can we please merge this? It's definitely needed for newest Arch, these dependencies have always been there, just hidden, and now cmake complains and refuses to build without explicitly naming them.

msoos:
Pretty please let's merge this!

PR #11969, opened 2024-10-04
Fixing channel issue with ELRS Mavlink-rc: number of channels is not set by default

msoos:
## Description 

If you are using ELRS MAVLINK setup as per: https://www.expresslrs.org/software/mavlink/ then the number of channels will never be set. Hence it will remain 0 -- even though there is clearly control data going through. This is because ELRS is not sending a full `RC_CHANNELS` packet, only channel update packets via `RC_CHANNELS_RAW`, to update the position of the values. Therefore, the number of channels is not set.

This ELRS system forwards BOTH the RC and the Mavlink on the same ELRS link. Using one TX and one RX, you get to have both Mavlink and RC. The Mavlink is forwarded via the backpack chip in the TX to the wifi. You can then connect to this wifi and QGC picks it up automatically, via zero-conf (I think! Very cool) and "just works". Which is great. But the RC doesn't SEEM to work, because the number of channels is never set.

TL;DR: `RC_CHANNELS` is never sent, only `RC_CHANNELS_RAW`, so `chancount` is never set.

## Test Steps

* Using RX ELRS 3.5.0 (release version, not experimental)
* Using TX ELRS 3.5.0 (release version, not experimental)
* Using TX ELRS Backpack 1.5.0 (release version, not experimental)
* Using SpeedyBee F405 Wing (full, not mini) board, Serial 1 connected to ELRS receiver, set up as per: https://www.expresslrs.org/software/mavlink/

We get connectivity of Mavlink forwarded via ELRS + backpack in the TX via WiFi, however, the radio channels cannot be calibrated, and it cannot arm, because the number of channels is seen as 0. With this fix, calibration works and the craft can be armed.

## How to Reproduce
Connect as per above. Observe that calibration is not possible, the QGC says there is no controller connected.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -1452,6 +1452,16 @@ void Vehicle::_handleRCChannels(mavlink_message_t& message)
     };
     int pwmValues[QGCMAVLink::maxRcChannels];
 
+    // Below is a hack that's needed by ELRS
+    // ELRS is not sending a full RC_CHANNELS packet, only channel update
+    // packets via RC_CHANNELS_RAW, to update the position of the values.
+    // Therefore, the number of channels is not set.
+    if (channels.chancount == 0) {
+        for(const auto& channelValue : _rgChannelvalues) {
+            if (*channelValue != UINT16_MAX) channels.chancount++;
+        }
+    }
+
     for (int i=0; i<QGCMAVLink::maxRcChannels; i++) {
         uint16_t channelValue = *_rgChannelvalues[i];
 
--- a/src/AutoPilotPlugins/Common/RadioComponentController.cc
+++ b/src/AutoPilotPlugins/Common/RadioComponentController.cc
@@ -180,6 +180,16 @@ void RadioComponentController::_setupCurrentState(void)
 /// Connected to Vehicle::rcChannelsChanged signal
 void RadioComponentController::_rcChannelsChanged(int channelCount, int pwmValues[QGCMAVLink::maxRcChannels])
 {
+    // Below is a hack that's needed by ELRS
+    // ELRS is not sending a full RC_CHANNELS packet, only channel update
+    // packets via RC_CHANNELS_RAW, to update the position of the values.
+    // Therefore, the number of channels is not set.
+    if (channelCount == 0) {
+        for (int channel=0; channel<16; channel++) {
+            if (pwmValues[channel] != INT16_MAX) channelCount++;
+        }
+    }
+
     for (int channel=0; channel<channelCount; channel++) {
         int channelValue = pwmValues[channel];
 


DonLakeFlyer:
This seems like a bad hack which goes against mavlink spec. Why can't whatever this system is be fixed to respect the mavlink spec?

https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/PX4/PX4ParameterMetaData.cc#L78. 

RC_CHANNEL.channel_count - Total number of RC channels being received. This can be larger than 18, indicating that more channels are available but not given in this message. This value should be 0 when no RC channels are available.

msoos:
Thanks for the response @DonLakeFlyer ! It turns out, they send `RC_CHANNELS_OVERRIDE` messages only (not _RAW). Does that change your mind regarding what should be done here?

DonLakeFlyer:
Hmm, that certainly is a mess. 

I could live with this:
* If RC_CHANNEL.channel_count == 0 then look at the individual chan*_raw values. Count up the number of ones which are not UINT16_MAX. That should be the channel count.
* Comment all this as being a hack to support ELRS (refer to issue below in the comment)
* Create an Issue about all of this with a note to remove it once ELRS is fixed to be correct to mavlink spec

That would work?

DonLakeFlyer:
Any updates on this?

msoos:
@DonLakeFlyer Yes, let me do this. Sorry for the delay.

msoos:
Hi @DonLakeFlyer ,

Can you please check the PR now? It now has your requirements implemented. I checked, it works well. Please merge if you think it's good :)

Thanks,

Mate

src/Vehicle/Vehicle.cc
```diff
@@ -1452,6 +1452,17 @@ void Vehicle::_handleRCChannels(mavlink_message_t& message)
     };
     int pwmValues[QGCMAVLink::maxRcChannels];
 
+    // Below is a hack that's needed by ELRS
+    // ELRS is not sending a full RC_CHANNELS packet, only channel update
+    // packets via RC_CHANNELS_RAW, to update the position of the values.
+    // Therefore, the number of channels is not set.
+    if (channels.chancount == 0) {
+        uint8_t num_non_uint16_max = 0;
```
DonLakeFlyer, line 1460: Can you remove this. It isn't used.

DonLakeFlyer (changes requested):

src/Vehicle/Vehicle.cc
msoos: OOps, yes, fixed! Sorry.

DonLakeFlyer:
All looks good now. Can you squash down to a single commit. Then once the rebuild finishes I'll merge.

msoos:
Nice, thank you so much! I just squashed :) Let me know how it goes :) Thanks so much,

Mate

DonLakeFlyer:
@msoos Thanks for fixing this

msoos:
Thanks so much for merging!

Issue #11693, opened 2024-07-20
Multi Vehicle Panel Overhaul


demirciAhmet:
Hello, I was working on a QGC custom build for a while (mostly on the UI/UX part), and I would now like to contribute to the development of QGC itself. The current multi-vehicle panel in QGroundControl provides basic functionality for monitoring multiple vehicles. However, I think it lacks certain features and usability improvements that could significantly improve the user experience and operational efficiency. Despite the challenges in making comprehensive changes due to the coupling of widgets and the complexity of QGC, I believe it is reasonable to improve the multi-vehicle panel.
### Issues with the Current Multi Vehicle Panel

- The current panel's design does not align well with QGC's overall Fly View design.
- The widgets on the panel are prone to breaking.
- The usability of the panel is not comprehensive. 
- The multi-vehicle panel selector only appears if more than one vehicle is connected, as mentioned in issue [#8076](https://github.com/mavlink/qgroundcontrol/issues/8076). I think the user should be able to see QGC's capability to control multiple vehicles without needing to connect multiple vehicles.
- Actions initiated through the multi-vehicle panel execute without confirmation, which can lead to accidental commands.

### Suggestions

**UI/UX Improvements**:

- Implementing a new solid top-right panel design that is flexible and easier to enhance.
- Implementing a design that aligns better with the overall Fly View.
- Simplifying instrument widgets for multi-vehicle panel usage.
- Simplified versions of the top bar indicators can be used for multi-vehicle panel usage.

**Functionality Enhancements**:

- Allowing users to change the active vehicle directly from the multi-vehicle panel.
- Enabling customization of displayed information for each vehicle, such as speed and altitude.
- Organizing vehicles based on user-defined criteria.
- Introducing sorting and filtering options.
- Adding additional notification indicators for critical events like low battery or signal loss.

**Usability Enhancements**:

- Ensuring that actions initiated through the panel require confirmation to prevent accidental commands.

![panel1](https://github.com/user-attachments/assets/c615f990-d04c-467d-ad7f-fc625105087e)
(The mockup design based on Stable_V4.3, and only covers the *solid top-right panel* suggestion above)

What do you think about this concept design? I am willing to implement this for QGC, and would appreciate your thoughs to ensure the successful implementation of these enhancements. Thank you for considering this feature request.

lebronkey:
This is a multi machine control station, and I think the multi machine list should be optimized

DonLakeFlyer:
[...]

PR #11970, opened 2024-10-06
Multi Vehicle Panel Overhaul

demirciAhmet:
Description
-----------

Related to #11693 

Here are the changes by the panel overhaul:

- [x] New solid top-right panel implemented, changed overall design of the multi vehicle list. 
- [x] To be able to add *guided action confirmation* to the buttons of the multi-vehicle panel, a new logic *"Select vehicles - Perform action"* implemented (now you first select vehicles from the panel by tapping on them, then perform the action to the selected vehicles, otherwise you had to confirm one by one for the each action, which would be very annoying).
- [x] Guided action confirmation for the mv actions
- [x] New swipe pages embedded in the new top-right panel. This way, other components such as the *PhotoVideoControl*, or any other widgets that may be implemented later, can be added to the top-right panel. 

image

https://github.com/user-attachments/assets/1e98bc0d-9e05-45f5-97c9-5f849c8ddbd6

Checklist:
----------


DonLakeFlyer:
Is there a fixed amount of space you devote to vehicle indicators which scrolling? is this why the list is clipped at the bottom?

demirciAhmet:
> Is there a fixed amount of space you devote to vehicle indicators which scrolling? is this why the list is clipped at the bottom?

Yes, it has a fixed height (currently half of the panel height), you scroll through the panel if there are many vehicles. This way there is no need to worry about adding new widgets to the panel in the future.

I thought about separating the lower part of the panel for buttons, and other widgets, which will be stacked page by page.

DonLakeFlyer (changes requested):
So far this is looking really good. Thanks. Biggest problems are real estate problems.

demirciAhmet:
> When there are no vehicles connected there is an artifact left on the screen in the upper right: Screenshot 2025-01-13 at 8 34 54 AM
> 
> When I am connected to multiple vehicles I see that same artifact and there is nothing that indicates there is a multi-vehicle panel available. If I click on the oval thing the panel finally opens up.

Actually, the idea on that point was that the panel itself, which you can open, close, and use even when connected to a single vehicle, should indicate the availability of a multi-vehicle panel. However, as you mentioned earlier, most users operate QGC with a single vehicle, so this might not be the best design approach.

Perhaps the most suitable option would be to convert the panel into something hidden by default but accessible by swiping from the right side of the screen, similar to sidebar menus on Android. When multiple vehicles are connected, the panel will automatically show itself while still allowing users to open and close it via swiping. What do you think?

DonLakeFlyer:
Here is what I would do:
* The visibility of the multi-vehicle panel is controlled through a setting which is part of FlyViewSettings. It defaults to visible.
* Add that as a setting in the Settings / Fly View / General ui
* If the setting in set to visible still only show the panel when multiple vehicles are available
* Rework the active vehicle selector in the toolbar. Make it look more like flight mode selection dropdown instead of a combo. In the expanded page for the dropdown, put the multi-vehicle panel on/off setting in there as well

How about that?

demirciAhmet:
> Here is what I would do:
> 
>     * The visibility of the multi-vehicle panel is controlled through a setting which is part of FlyViewSettings. It defaults to visible.
> 
>     * Add that as a setting in the Settings / Fly View / General ui
> 
>     * If the setting in set to visible still only show the panel when multiple vehicles are available
> 
>     * Rework the active vehicle selector in the toolbar. Make it look more like flight mode selection dropdown instead of a combo. In the expanded page for the dropdown, put the multi-vehicle panel on/off setting in there as well
> 
> 
> How about that?

So, you're suggesting we stick with something similar to the previous approach regarding the on/off point. But it might result in a loss of flexibility and make it harder to use?
 
Isn't this "while still allowing users to open and close it via swiping" approach more user friendly?

DonLakeFlyer:
My thinking was along the lines of you either use the panel or you don't use the panel. If you don't use the panel then you don't want stuff no matter how small cluttering things up. I was not thinking that you might hide/show the panel multiple times during the same flight. In that case, yes some quicker way to access it would be better. But I'm not sure if that an important use case to consider?

I don't want to create new paradigms for the UI unless they are really necessary. And then if something new is added it is added in a consistent way. For example if you can swipe this multi-vehicle thing off the screen. Then why can't you swipe the video thing off the screen when you don't need it and are just flying a single vehicle. If it's not a critical need then I'd rather have consistency in how things are done.

demirciAhmet:
> My thinking was along the lines of you either use the panel or you don't use the panel. If you don't use the panel then you don't want stuff no matter how small cluttering things up. I was not thinking that you might hide/show the panel multiple times during the same flight. In that case, yes some quicker way to access it would be better. But I'm not sure if that an important use case to consider?
> 
> I don't want to create new paradigms for the UI unless they are really necessary. And then if something new is added it is added in a consistent way. For example if you can swipe this multi-vehicle thing off the screen. Then why can't you swipe the video thing off the screen when you don't need it and are just flying a single vehicle. If it's not a critical need then I'd rather have consistency in how things are done.

Oh, I see what you mean now—that makes sense. I’ll probably complete the improvements tomorrow. Should I push the changes as a new commit for easier review, or rebase again and force push?

DonLakeFlyer:
The setting for the panel in the toolbar indicator should use the same wording as in Fly View Settings: "Enable Multi-Vehicle Panel". Also it should be in the expanded portion of the indicator which matches the usage of settings in indicators in all the other indicators.

DonLakeFlyer:
Took a look. Perfect, thanks. This is great new work. I'll merge once it passes CI.

demirciAhmet:
> Took a look. Perfect, thanks. This is great new work. I'll merge once it passes CI.

Glad to hear! Thank you too.

DonLakeFlyer (approved):

PR #11976, opened 2024-10-07
Camera: Move CameraMetaData list from FirmwarePlugin to CameraManager

HTRamsey:
Resolves an old TODO by moving camera metadata list to the rest of the camera stuff
Convert to Q_GADGET since a Q_OBJECT isn't needed


src/FirmwarePlugin/FirmwarePlugin.h
```diff
@@ -307,10 +307,6 @@ class FirmwarePlugin : public QObject
     /// @return A list of QUrl with the indicators
     virtual const QVariantList& modeIndicators(const Vehicle* vehicle);
 
-    /// Returns a list of CameraMetaData objects for available cameras on the vehicle.
-    /// TODO: This should go into QGCCameraManager
-    virtual const QVariantList& cameraList(const Vehicle* vehicle);
```
DonLakeFlyer, line 312: So how does a custom build override the list of available cameras now?

HTRamsey: True, I will re-enable that ability. Personally I think it would be better to read some kinda resource file for this kinda stuff, like a json or whatever. But lemme think about it some first. I do think it's at least nice to have all the camera code together as long as we keep the same functionality.

DonLakeFlyer: "Personally I think it would be better to read some kinda resource file for this kinda stuff, like a json or whatever."

I agree. In general for custom build tweaking I've been moving away from C++ code when possible to resource based stuff which can be overridden. Json for this certainly makes sense.

HTRamsey: So now they can override the json file in the resource system

DonLakeFlyer:
Can you make the camera json file follow the specs for the internal QGC json files (version, type and so forth) and use the helper routines to validate/load? This provides versioning and translation support.

HTRamsey:
Done

src/Camera/QGCCameraManager.cc
```diff
@@ -555,3 +568,103 @@ QGCCameraManager::_stepStream(int direction)
...
+
+    return _cameraList;
+}
+
+QList<CameraMetaData*> QGCCameraManager::_parseCameraMetaData(const QString &jsonFilePath)
+{
+    QList<CameraMetaData*> cameraList;
+
+    QFile file(jsonFilePath);
+    if (!file.open(QIODevice::ReadOnly | QIODevice::Text)) {
```
DonLakeFlyer, line 591: You need to use JsonHelper::openInternalQGCJsonFile to get the translation support. This will do this for you:
* Open file
* Validation
* Transation

HTRamsey: I'll get it right eventually

DonLakeFlyer (approved):

PR #11989, opened 2024-10-09
CI: Create Android Play Store Action

HTRamsey:
Closes #11989
Can't finish or test until dronecode android play store account is created


DonLakeFlyer:
So I have access to the original playstore account again. If you need me to do something I should be able to do it.

HTRamsey:
I'm not entirely sure, but seems this action could be useful: https://github.com/r0adkll/upload-google-play
We could merge this and get you to mess with trying to link it to the account somehow

PR #11990, opened 2024-10-09
Fixed 'Sensors Setup' page not showing GUI

stepan14511:
Fixed Vehicle Setup -> Sensors page for PX4 not showing GUI, made to help user with calibration.

Description
-----------
Used to be:
image
Now it works correctly:
image

Test Steps
-----------
Go to Vehicle Setup -> Sensors. Start, for example, accelerometer calibration and you will see no changes in the right part of the screen. Now it works the same way, as it does in the stable verison (correctly).

Related Issue
-----------
There might be the same issue with others autopilots. I don't not have ability to check.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSensorsComponentController.cc
+++ b/src/AutoPilotPlugins/APM/APMSensorsComponentController.cc
@@ -390,14 +390,12 @@ void APMSensorsComponentController::calibrateGyro(void)
     _vehicle->startCalibration(_calTypeInProgress);
 }
 
-void APMSensorsComponentController::_handleUASTextMessage(int uasId, int compId, int severity, QString text)
+void APMSensorsComponentController::_handleUASTextMessage(int compId, int severity, QString text, QString description)
 {
     Q_UNUSED(compId);
     Q_UNUSED(severity);
+    Q_UNUSED(description);
     
-    if (uasId != _vehicle->id()) {
-        return;
-    }
 
     QString originalMessageText = text;
     text = text.toLower();
--- a/src/AutoPilotPlugins/APM/APMSensorsComponentController.h
+++ b/src/AutoPilotPlugins/APM/APMSensorsComponentController.h
@@ -123,7 +123,7 @@ signals:
     void setAllCalButtonsEnabled                (bool enabled);
 
 private slots:
-    void _handleUASTextMessage  (int uasId, int compId, int severity, QString text);
+    void _handleUASTextMessage  (int compId, int severity, QString text, QString description);
     void _mavlinkMessageReceived(LinkInterface* link, mavlink_message_t message);
     void _mavCommandResult      (int vehicleId, int component, int command, int result, bool noReponseFromVehicle);
 
--- a/src/AutoPilotPlugins/APM/APMSubMotorComponentController.cc
+++ b/src/AutoPilotPlugins/APM/APMSubMotorComponentController.cc
@@ -17,11 +17,11 @@ APMSubMotorComponentController::APMSubMotorComponentController(void)
     connect(_vehicle, &Vehicle::textMessageReceived, this, &APMSubMotorComponentController::handleNewMessages);
 }
 
-void APMSubMotorComponentController::handleNewMessages(int uasid, int componentid, int severity, QString text)
+void APMSubMotorComponentController::handleNewMessages(int componentid, int severity, QString text, QString description)
 {
-    Q_UNUSED(uasid);
     Q_UNUSED(componentid);
     Q_UNUSED(severity);
+    Q_UNUSED(description);
     if (_vehicle->flightMode() == "Motor Detection"
         && (text.toLower().contains("thruster") || text.toLower().contains("motor"))) {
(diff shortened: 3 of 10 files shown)


DonLakeFlyer:
I see what was changed but I don't understand why it was needed. The signals should have worked there way through the new mechanism:
* https://github.com/mavlink/qgroundcontrol/blob/master/src/MAVLink/StatusTextHandler.cc#L338
* https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L4004
* And then the changes to the text parsing seem like a no-op to me.

@stepan14511 Could you help with debugging this to determine why the current code doesn't work?

stepan14511:
@DonLakeFlyer 

- > The signals should have worked there way through the new mechanism

In the mechanism all the signals work absolutely fine. `StatusTextHandler::textMessageReceived` signals are coming to the `Vehicle` object. But `SensorsComponentController` is not connecting to `StatusTextHandler::textMessageReceived`, but to `Vehicle::textMessageReceived`, which is never emited in a `Vehicle` object.

The other solution could be to subscribe straightly to the `StatusTextHandler::textMessageReceived`, but `StatusTextHandler *m_statusTextHandler` is private in the Vehicle class. And it's getter function is removed for some reason:
https://github.com/mavlink/qgroundcontrol/blob/2d4bb34805a338cd1259664018d281850136be99/src/Vehicle/Vehicle.h#L1339

- > And then the changes to the text parsing seem like a no-op to me.

https://github.com/mavlink/qgroundcontrol/blob/2d4bb34805a338cd1259664018d281850136be99/src/AutoPilotPlugins/PX4/SensorsComponentController.cc#L233
The expression in the if statement is always false, since the text variable is following (for example):
`[cal] progress &lt;20&gt;`

Because of this the progress bar is not working.

HTRamsey:
Good catch, I probably missed the PX4 issue when making StatusTextHandler because previously I only tested on ArduPilot but at least it helped catch the SensorsComponentController issue. Connecting to the Vehicle signal is the correct way as I had intended for it to be used.

DonLakeFlyer:
Ok, so isn't one of the fixes to add:
```
    (void) connect(m_statusTextHandler, &StatusTextHandler::textMessageReceived, this, &Vehicle::textMessageReceived);
```
This will get the signal up through Vehicle.

As far as the "<" versus "&lt" stuff. It looks like this changed on the PX4 firmware side of things at some point. Since in the past the "<" certainly worked. My concern is with back compat with older version of firmware. @HTRamsey Is there some Qt magic to get these strings to replace the "&" stuff which real chars? If so, then the cal stuff could jam the strings through that first and then use the non "&" chars for comparison. This would then in turn work on all firmware versions.

stepan14511:
@DonLakeFlyer 
This idea sound better. But it requires a bit more changes. It can be seen here: https://github.com/stepan14511/qgroundcontrol/commit/affe648c3962abe72983bccd1e71eced0b0bd515
If we will agree on this approach, I will merge it to my master to appear in this pull request.

> Is there some Qt magic to get these strings to replace the "&" stuff which real chars?

I couldn't find anything apart from the straightforward replace function.

DonLakeFlyer:
> If we will agree on this approach

I believe this is the way to go.

For converting the &gt things:
https://stackoverflow.com/questions/7696159/how-can-i-convert-entity-characterescape-character-to-html-in-qt

stepan14511:
@HTRamsey
Is there any particular reason why in the line below you are using `toHtmlEscaped()`?
https://github.com/mavlink/qgroundcontrol/blob/3bceda063c591d12663dac7b7a49512eb0614356/src/MAVLink/StatusTextHandler.cc#L338

Removing it solves the issue.

HTRamsey:
Not that I recall, I believe that's what was there already before I moved it into StatusTextHandler

stepan14511:
This is exactly the reason, why old if is not working:
image

Then what would be better: to convert it back in the autopilots or just remove it?

DonLakeFlyer:
Hmm, interesting. I believe somewhere along the the way PX4 starting sending status text messages which were HTML. So the message display starting using a rich text field to display status text.

Could someone find the "blame" on the original toHtmlEscaped? I only vaguely remember those changes. Need to understand why that was put in there. I think maybe @bkueng was involved?

DonLakeFlyer:
I just went backwards through blame. Looks like it was @bkueng. Hopefully he can help explain why this is necessary.

DonLakeFlyer:
So I looked through the code and I believe it's correct that the toHtmlEscaped (somewhere) because both the critical vehicle message popup and and vehicle message dropdown use a richtext control If you don't escape then some things could be assumed to be HTML when they are not. That said I think the place where this escaping should happen should not be at a global level. It should only be when the status text is about to be displayed in one of those control. For generic signaling out from Vehicle of status text. I think it should not be escaped at that point.

I'm pawing through code right now to figure out exactly where that should be.

HTRamsey:
@stepan14511 Is this still necessary? I don't know if this was replaced by the other PR or if this is different

DonLakeFlyer:
Sorry I should have closed this. It was replaced by my pull

Issue #11994, opened 2024-10-10
Validation errors in toolbar indicator pages don't prevent page from closing


DonLakeFlyer:
Need to figure out how to make closing respect `globals.validationError`

DonLakeFlyer:
Is use mechanism that QGCDialog uses to prevent this.

PR #12007, opened 2024-10-13
Rework/Fix validation support

DonLakeFlyer:
* Add validation support to QGCTextField
  * Call QGCTextField.showValidationError/clearValidationError to set/clear 
  * Validation error shown with red border and tooltip
* Rework FactTextField to use new QGCTextField support
* Validation errors tracked by count not bool: globals.validationErrorCount
* New mainWindow.allowViewSwitch function to check whether view switch should be allowed
* Fixed numerous problems with QGCPopupDialog with validation not being triggered correctly. As well as dialogs allowed to close with validation errors.

Related to #11994

### DIFF
--- a/src/FactSystem/FactControls/FactTextField.qml
+++ b/src/FactSystem/FactControls/FactTextField.qml
@@ -8,8 +8,7 @@ import QGroundControl.Controls
 import QGroundControl.ScreenTools
 
 QGCTextField {
-    id: control
-
+    id:                 control
     text:               fact ? fact.valueString : ""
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
@HTRamsey Don't know if you want to review these detailed qml things? Would be good for sanity check but also don't want to waste your time. Previously most of the internal ui stuff had no reviewers.

DonLakeFlyer:
Merging so I can move on to the remaining fixes for #11994

PR #12011, opened 2024-10-14
Status text html escaping in wrong place

DonLakeFlyer:
This is a replacement for #11990. I was crazy hard to get my head around all the back and forth of status text between vehicle and StatusTextHandler. So I created the most minimal set of changes I could come up with to fix the problem. I tested it and it works as far as I can tell.

The basic idea is that status text which flows through the system should not be html escaped until the very last place when it needs to be output to a rich text control. This is critical vehicle error popups and the message indicator dropdown. Status text which flows from Vehicle should raw for people connecting to that signaling.

@stepan14511 Can you test as well.

### DIFF
--- a/src/MAVLink/StatusTextHandler.cc
+++ b/src/MAVLink/StatusTextHandler.cc
@@ -137,9 +137,10 @@ void StatusTextHandler::resetErrorLevelMessages()
     }
 }
 
-void StatusTextHandler::handleTextMessage(MAV_COMPONENT compId, MAV_SEVERITY severity, const QString &text, const QString &description)
+void StatusTextHandler::handleHTMLEscapedTextMessage(MAV_COMPONENT compId, MAV_SEVERITY severity, const QString &text, const QString &description)
 {
     QString htmlText(text);
+
     (void) htmlText.replace("\n", "<br/>");
 
     // TODO: handle text + description separately in the UI
@@ -335,7 +336,7 @@ void StatusTextHandler::_chunkedStatusTextCompleted(MAV_COMPONENT compId)
 
     (void) m_chunkedStatusTextInfoMap.remove(compId);
 
-    emit textMessageReceived(compId, severity, messageText.toHtmlEscaped(), "");
+    emit textMessageReceived(compId, severity, messageText, "");
 }
 
 void StatusTextHandler::_handleTextMessage(uint32_t newCount, MessageType messageType)
--- a/src/MAVLink/StatusTextHandler.h
+++ b/src/MAVLink/StatusTextHandler.h
@@ -57,7 +57,7 @@ public:
     ~StatusTextHandler();
 
     void mavlinkMessageReceived(const mavlink_message_t &message);
-    void handleTextMessage(MAV_COMPONENT componentid, MAV_SEVERITY severity, const QString &text, const QString &description);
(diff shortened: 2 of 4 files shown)


src/MAVLink/StatusTextHandler.cc
```diff
@@ -139,7 +139,7 @@ void StatusTextHandler::resetErrorLevelMessages()
 
 void StatusTextHandler::handleTextMessage(MAV_COMPONENT compId, MAV_SEVERITY severity, const QString &text, const QString &description)
 {
-    QString htmlText(text);
+    QString htmlText(text.toHtmlEscaped());
```
bkueng, line 142: The problem with escaping here is that it will be escaped twice in case of the events code path from https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L1230

That's why I originally did it outside, but did not consider that calibration was affected.
The simplest would be to add a flag as argument to indicate if it's escaped already. But I do agree, would be good to simplify this and avoid the back and forth.

stepan14511:
For me it works the same good, as in the https://github.com/mavlink/qgroundcontrol/pull/11990. Tested all the ways I know and use the QGC, nothing broke.

But, for the history, I will remind you that we decided to fix the signals the other way.

src/MAVLink/StatusTextHandler.cc
DonLakeFlyer: My understanding from code inspection are these path(s):
* StatusTextHandler::mavlinkMessageReceived is the only thing which sucks up status text messages from mavlink
  * It eventually emits StatusTextHandler::textMessageReceived
* Vehicle catches StatusTextHandler::textMessageReceived and calls Vehicle::_textMessageReceived which...
  * Signals Vehicle::textMessageReceived
  * Calls StatusTextHandler::handleTextMessage which...
    * emits newErrorMessage which is escaped text
    * Stuffs escaped text into the message list
* Events come in through Vehicle::_handleEvent - not sure how that gets called. Which then...
  * Also calls StatusTextHandler::handleTextMessage

I don't see how you end up going through StatusTextHandler::handleTextMessage twice. I also don't see how to test this since I don't know how to generate an event.

bkueng: Escaping for the events is done here: https://github.com/mavlink/qgroundcontrol/blob/be1e2d46c95812d3ad700bdc4807eca95829be6b/src/MAVLink/LibEvents/EventHandler.cc#L48.
This is because they can contain URL's which should not be escaped.

If you connect to PX4 SITL and do a takeoff you should get some events.

DonLakeFlyer: Ah, ok. I get it now.

DonLakeFlyer: @bkueng I think I have this correct now. Can you take a look again?

* Status text coming in from events is known to already be escaped
* Regular textMessageReceived signal is not escaped
* StatusTextHandler::handleTextMessage -> StatusTextHandler::handleHTMLEscapedText name change which assumes escaped text on the way in
* Raw status text from vehicle is escaped before call to StatusTextHandler::handleHTMLEscapedText

I haven't had time to test yet. Take me a bit to get a SITL setup to work again on OSX arm.

PR #12013, opened 2024-10-15
PX4: Airframe: fix application crash when selecting Airframe type

dakejahl:
The application will crash when you select an Airframe. This is because the overrideCursor is nullptr
https://doc.qt.io/qt-6/qguiapplication.html#overrideCursor


HTRamsey:
Can you do the same in for APMAirframeComponentController?

PR #12019, opened 2024-10-16
UTMSPManager: Convert to Singleton

HTRamsey:
Convert from tools to singletons


DonLakeFlyer:
@HTRamsey I'd ask the UTMSP folks who created this to review. I'm not familiar with this code.

HTRamsey:
It's mostly copy and paste but that's fair. Should I split out the airlink stuff to a different PR?

HTRamsey:
@royzah since you seem to use this code, could you check/test this? Trying to remove toolbox and use singletons instead

PR #12026, opened 2024-10-18
Fix issue with cached tile set download cancelation

SleepingSoul:
Fix issue with cached tile set download cancelation

Description
-----------
Cancel download did not cancel download. It just set "downloading" flag to "false". This fix resolves this issue.

Test Steps
-----------
I have tested it locally, on release build, by starting offline map download and stopping it

Checklist:
----------

### DIFF
--- a/src/QtLocationPlugin/QGCCachedTileSet.cpp
+++ b/src/QtLocationPlugin/QGCCachedTileSet.cpp
@@ -61,6 +61,11 @@ QString QGCCachedTileSet::downloadStatus() const
 
 void QGCCachedTileSet::createDownloadTask()
 {
+    if (_cancelPending) {
+        setDownloading(false);
+        return;
+    }
+
     if (!_downloading) {
         setErrorCount(0);
         setDownloading(true);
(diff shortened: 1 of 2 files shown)


src/QtLocationPlugin/QGCCachedTileSet.cpp
```diff
@@ -82,11 +87,17 @@ void QGCCachedTileSet::createDownloadTask()
 
 void QGCCachedTileSet::resumeDownloadTask()
 {
```
HTRamsey, line 89: Should we only resume if it is already canceled? If you click Ok or Cancel the UI stops updating the download and this value can get out of sync with whether the button shows to resume/cancel

SleepingSoul: Okay since its merged now I guess it is not necessary anymore

HTRamsey: I tried it out and it didn't help the problem I was noticing, so I figured just merge this and I'll look more into it later. I think it's partly on the QML side of things

SleepingSoul: Okay got it, thanks for explanation

PR #12032, opened 2024-10-24
Use correct aspectRation for gst video

### LINKED ISSUE #11896: aspectRatioChanged signal does not get emitted when the aspect ratio changes
## Expected Behavior
When the stream info changes (on receipt of `VIDEO_STREAM_STATUS` msg) the `_ar` property in `FlightDisplayViewVideo.qml` should be updated.

## Current Behavior
On app start the `_ar` property is set to `1.77 (16:9)`, but because `VideoManager::aspectRatioChanged` is never triggered, `_ar` is never updated and so aspect ration calculations in `FlightDisplayViewVideo.qml` are wrong. I believe this only affects camera tracking functionality.

## Steps to Reproduce:
This is not particularly straight forward issue to see without debugging. But if you have a Mavlink camera that can do tracking, you can do the following
1. Under general settings, set the `Aspect Ratio` to a unrealistic large wide video, eg `5`
2. Restart QGC
3. Initialise the tracking outside of the image on the left (ie on the black bars)
4. Note the tracking box is then displayed well within the image (as the aspect ratio calcs are wrong)

## System Information
- Operating System: Windows 11
- QGC Version: 4.4.1
- QGC build: Built from source, with custom firmware plugin
- Flight Controller: PX4
- Autopilot (with version): PX4 1.15.0 RC2

gillamkid:
Fixes #11896 

# Description
The wrong ratio was being used for GST video (if the video stream aspect ratio did not match the operator provided aspect ratio). This made so thinngs like grid lines weren't properly placed and also made track point setting inaccurate.

## Sponsor

# Before
Note how the grid lines are not accurately placed and the video does not fill the width of the screen even though it is in "Fit Width" mode.
bad-ratio
I generated the sample image using the following command
```
gst-launch-1.0 videotestsrc ! openh264enc ! rtph264pay config-interval=10 pt=96 ! udpsink host=127.0.0.1 port=5600
```
My QGC settings were "Fit Width" in video settings and I made video streaming enabled for h264 port 5600

# After
Note how the grid lines are accurately placed and the video does fill the width of the screen as expected when in "Fit Width" mode.
good-ratio

### DIFF
--- a/src/FlightDisplay/FlightDisplayViewVideo.qml
+++ b/src/FlightDisplay/FlightDisplayViewVideo.qml
@@ -26,7 +26,9 @@ Item {
 
     property bool useSmallFont: true
 
-    property double _ar:                QGroundControl.videoManager.aspectRatio
+    property double _ar:                QGroundControl.videoManager.gstreamerEnabled
+                                            ? QGroundControl.videoManager.videoSize.width / QGroundControl.videoManager.videoSize.height
+                                            : QGroundControl.videoManager.aspectRatio
     property bool   _showGrid:          QGroundControl.settingsManager.videoSettings.gridLines.rawValue
     property var    _dynamicCameras:    globals.activeVehicle ? globals.activeVehicle.cameraManager : null
     property bool   _connected:         globals.activeVehicle ? !globals.activeVehicle.communicationLost : false


HTRamsey:
Nice, I don't know if you're interested in helping look into it, but that entire file actually needs to be reworked, and really most of the video QML stuff. The loader actually doesn't work as expected (videoDisabled isn't even used), and the thermal video should be in a loader. Also, ideally this would be combined with the UVC file somehow and just have the video source or video background changed

gillamkid:
I can't commit to reworking the entire file. I did have the same idea about combining with the UVC file, so it's nice to know that is a welcomed change if I ever decide to take a stab at it.

For now I mostly just make fixes that cause pains when flying the vehicle my company is developing. The messed up GST aspect ratio was giving us grief when setting tracks.

PR #12033, opened 2024-10-25
Fix disappearing loiter circles

gillamkid:
# Description
When video streaming was enabled, toggling between mini video and full screen video mode would make the circles representing loiter circles on the flight path disappear. This PR fixes that issue.

# Steps to Reproduce
1. open a PX4 plane sim. ex
```
cd PX4-Autopilot
make px4_sitl gazebo-classic_plane_lidar
```
2. Open QGC daily
3. create a mission with a loiter waypoint.
make-mission

4. upload the mission to the vehicle and go to Fly view
before-toggle

5. switch to big video mode
big-video

6. switch back to Big map mode
after-toggle
Notice how the loiter circle has collapsed. This is the bug that is fixed. With the fix in this PR the loiter circle properly reappears when toggling the video from mini to full screen.

## Sponsor

### DIFF
--- a/src/PlanView/SimpleItemMapVisual.qml
+++ b/src/PlanView/SimpleItemMapVisual.qml
@@ -97,14 +97,11 @@ Item {
         target: _missionItem.isSimpleItem ? _missionItem : null
 
         onLoiterRadiusChanged: {
-            _loiterVisual.blockSignals = true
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Great, thanks so much

PR #12042, opened 2024-10-29
CMake: Fix MacOS GStreamer RPath

HTRamsey:
Taken from previous stable
https://github.com/mavlink/qgroundcontrol/blob/990ed12f5be9464d737ba0da703a126f63b0eff0/QGCPostLinkInstaller.pri#L22

### DIFF
--- a/cmake/CreateMacDMG.cmake
+++ b/cmake/CreateMacDMG.cmake
@@ -1,6 +1,6 @@
 # include(BundleUtilities)
 
-# message(STATUS "Copy GStreamer framework into bundle")
+message(STATUS "Copy GStreamer framework into bundle")
 file(COPY /Library/Frameworks/GStreamer.framework DESTINATION staging/QGroundControl.app/Contents/Frameworks)
 file(REMOVE_RECURSE staging/QGroundControl.app/Contents/Frameworks/GStreamer.framework/Versions/1.0/bin)
 file(REMOVE_RECURSE staging/QGroundControl.app/Contents/Frameworks/GStreamer.framework/Versions/1.0/etc)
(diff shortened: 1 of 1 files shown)


HTRamsey:
@DonLakeFlyer If this doesn't work can you see if you can try something here? It seems the issue was not fixing the rpath for the extra included gstreamer libs like gstallocators

DonLakeFlyer:
Hmm, not booting for me. I got this once:
```
2024-10-29 12:52:41.159 QGroundControl[10297:14468116] WARNING: AVCaptureDeviceTypeExternal is deprecated for Continuity Cameras. Please use AVCaptureDeviceTypeContinuityCamera and add NSCameraUseContinuityCameraDeviceType to your Info.plist.
arch: /Applications/QGroundControl.app/Contents/MacOS/../Frameworks/GStreamer.framework/Versions/1.0/libexec/gstreamer-1.0/gst-plugin-scanner isn't executable

(.:10297): GStreamer-WARNING **: 12:52:41.244: External plugin loader failed. This most likely means that the plugin loader helper binary was not found or could not be run. You might need to set the GST_PLUGIN_SCANNER environment variable if your setup is unusual. This should normally not be required though.
```

But then it went away which is strange. I'll try to figure this one out.

HTRamsey:
I didn't add this step from the previous stable `ln -sf $${TARGET}.app/Contents/Frameworks $${TARGET}.app/Contents/Frameworks/GStreamer.framework/Versions/1.0/libexec/Frameworks`

HTRamsey:
Oh actually you may have to check permissions on gst-plugin-scanner. I had to mark it as executable for linux

HTRamsey:
It should be okay to still have a system version installed once we get everything correct

DonLakeFlyer:
Right, but you have to keep in mind while testing to remove it to prove it works without it. I've been trying some simple things based on what stable does in it's cmake to fix this but can't find anything that works. It you look at Stable 4.4 cmake support you'll see a much more complex pile of gstreamer install stuff which I assume works (prepare_gstreamer_framework.sh,  osxrelocator.py, QGCDeploy.cmake). I'll give it a try. But it's gonna take me a while to get to it. Somewhat involved.

HTRamsey:
Hopefully we can get it to at least run without crashing relatively easily. I can make a PR to use the older v4.4 method the best I can and just get you to test it and fix any last parts if you want

DonLakeFlyer (approved):

PR #12049, opened 2024-10-30
Gimbal/Joystick: implement rate commands

### LINKED ISSUE #11956: Gimbal manual control doesn't work
## Steps to Reproduce:

Follow the [official instruction](https://docs.px4.io/main/en/advanced/gimbal_control.html)
1. Set `MNT_MODE_OUT=AUX`
2. Set `MNT_MODE_IN=Auto`
3. Configure Gimbal Servo using `Actuators` tab:
```
PWM:       200Hz
AUX 5:     "Gimbal Pitch"
Disarmed:  1000
Minimum:   1100
Maximum:   1900
```
4. Verify that gimbal servo is working using `Actuator Testing`
5. Set `MNT_MAN_PITCH=AUX5`
6. In `Joystick -> Button Assigment` tab set `Gimball Up` and `Gimball Down` to some buttons on connected USB joystick 

Gimbal work automatically:
- It change position after arming
- It changing position relatively to frame position if `MNT_DO_STAB` is enabled. 
So I can assume that gimbal is actually working, but I can't manipulate it manually for some reason

## System Information
- Operating System: Ubuntu 24.10
- QGC Version: 4.4.0 (AppImage)
- QGC build: stable
- Flight Controller: Pixhawk 6c mini
- Flight Controller FW: PX4 Autopilot 1.14.3

julianoes:
This pull request changes the behaviour of the joystick Gimbal Up/Down, Left/Right commands. Instead of sending an angle and increasing that angle with a step (using repeat) we now send the angular rate.

The way it works is that we:
1. Send the angular rate command when either button is pressed.
2. Keep sending that command at 2 Hz to allow the autopilot to stop again with a timeout.
3. Send a last command with rates set to zero when the button is released again.

We should probably allow a user to adjust the angular rate commanded which is currently hard-coded. I'm not sure where the right place for this would be.

Fixes #11956.


HTRamsey:
@julianoes could you set the parent to the GimbalController to be the vehicle and then remove the deleteGimbalController() from the vehicle class? If not I can make a separate PR to do so later

julianoes:
Feel free to add commits @HTRamsey. I'm not too sure what you mean.

HTRamsey:
Okay will do. Just FYI when you do something like in this picture, it sets the vehicle as the parent of the GimbalController. When QObjects are deleted, they also delete all of their children automatically
image

julianoes:
I think that's actually on purpose:
https://github.com/mavlink/qgroundcontrol/blob/617b651ae50047b3d89a585946893d78fa109936/test/Vehicle/RequestMessageTest.cc#L43-L44

HTRamsey:
Oh I guess that function can stay then, although the parenting stuff should still be done

julianoes:
Done @HTRamsey.

HTRamsey:
Nice thanks. `QmlObjectListModel _gimbals;` isn't created dynamically with a parent but since GimbalController calls `_gimbals.clearAndDeleteContents()` it should be fine

Davidsastresas:
I will check this on the next few days, sorry, but I read that matter about making it automatically get destroyed with vehicle destructor, and that was actually a workaround to fix tests, at least in 4.4 for the way tests were done this hack was needed, I don't remember exactly the details but I wanted to mention in case you guys find the same issue. Thanks!

Davidsastresas:
Good catch the gimbal parent front. It wasn't being deleted until now right? so we had memory leaks if vehicle disconnected, not huge but we had, right?

I checked your solution  for rates, very elegant as usual! I like how you managed to control the timer to avoid timeouts. I tested in SITL Ardupilot and it works great. Indeed this is much much nicer than having steps, this feels much more natural, yet you still have somehow precise control to just move a few degrees.

I added the rate used by buttons to the Settings panel in cfdadfddf4a0a8b8234dbb5d998cf46bcda14fc8:
Screenshot from 2024-11-04 17-55-51

On the other commit I fixed the buttons that got broken by the QGCButton restyling, text out of sight

Having confirmed this works correctly, we should update click and drag to use rates too. I was tempted to do it now, but I remember we had some issue in android, apparently android is not processing press and release as desktop, so that was the fundamental issue with click and slide for android, so I think we can address that maybe other time in a different PR.

Thanks Julian!

Davidsastresas (approved):

julianoes:
Amazing, thanks @Davidsastresas! And great that you could test it with ArduPilot and it works!

PR #12063, opened 2024-11-03
Android: Fix Serial Issues

### LINKED ISSUE #12031: Android App freezes after random period when connecting over serial
## Expected Behavior
Please tell us what should happen as detailed as possible.
After launching QGC and connecting to a cube running arducopter (tried with both black and orange), the system works normally.  

## Current Behavior
Please describe what happens instead of the expected behavior.
After a random amount of time, sometimes 1minute sometimes 2hours, the app freezes completly and one CPU core gets stuck at 100%.
This only happens on android when using a serial connection, not UDP.

## Steps to Reproduce:
1. Connect an android device over serial to a cube running arducopter.
2. Wait
3. At some point the app freezes
4. After restarting the app it works again for some time then freezes again

## System Information
- Operating System: Android 9 & 12 tested
- QGC Version: latest master, commit: d915ef0d1e83a172524137e7ab431a63ab3cfc14
- QGC build: daily
- Flight Controller: pixhawk cube black and orange
- Autopilot (with version): Ardupilot Copter 4.5.6

HTRamsey:
Closes #12031
Hopefully fixes freezing after several hours by increasing buffer and adding memory checks
Moves serial code to another java class for better organization
Better debugging output and result verification
Moving serial link to a separate thread should resolve any remaining issues


src/Joystick/JoystickAndroid.cc
```diff
@@ -285,7 +285,7 @@ bool JoystickAndroid::init() {
     return true;
 }
 
-static const char kJniClassName[] {"org/mavlink/qgroundcontrol/QGCActivity"};
+static const char kJniClassName[] {"org/mavlink/qgroundcontrol/QGCUsbSerialManager"};
```
DonLakeFlyer, line 288: Is this correct. Seems strange?

HTRamsey: Yeah, it's because all the serial methods were moved to a separate class "QGCUsbSerialManager" to organize the code a bit.

PR #12068, opened 2024-11-05
Add crsf-bind button

benjinne:
Adds crsf-bind button.
Based on: 
- https://github.com/PX4/PX4-Autopilot/pull/23294
and
- https://github.com/mavlink/mavlink/pull/2162

image

Test Steps
-----------
The button successfully sets a CRSF (crossfire, ELRS, etc) receiver into bind mode

### DIFF
--- a/src/AutoPilotPlugins/Common/RadioComponent.qml
+++ b/src/AutoPilotPlugins/Common/RadioComponent.qml
@@ -389,6 +389,14 @@ SetupPage {
                     }
 
                     QGCButton {
+                        text:       qsTr("CRSF Bind")
+                        onClicked:  mainWindow.showMessageDialog(qsTr("CRSF Bind"),
+                                                                 qsTr("Click Ok to place your CRSF receiver in the bind mode."),
+                                                                 Dialog.Ok | Dialog.Cancel,
+                                                                 function() { controller.crsfBindMode() })
+                    }
+
+                    QGCButton {
                         text:       qsTr("Copy Trims")
                         onClicked:  mainWindow.showMessageDialog(qsTr("Copy Trims"),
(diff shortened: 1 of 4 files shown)


HTRamsey:
Looks fine to me, do you know if there is an enum available for the rxType?

DonLakeFlyer:
@benjinne What happens on the firmware side if you try a CSRF bind on an older firmware that doesn't support this?

DonLakeFlyer:
Also now that the RC_TYPE_foo enums have been added to mavlink can the code be updated to use both of those?

benjinne:
> @benjinne What happens on the firmware side if you try a CSRF bind on an older firmware that doesn't support this?

I believe nothing happens. The version of ELRS supported is 3.4.0+ 
https://www.expresslrs.org/quick-start/binding/?h=bind#updated-binding-procedure-since-expresslrs-340

DonLakeFlyer:
I looked at the ArduPilot code and it's going to get confused by this. It will cause Spektrum bind to start. Are there plans to support this in ArduPilot as well? If not, then I would say this show for PX4 only.

benjinne:
> Also now that the RC_TYPE_foo enums have been added to mavlink can the code be updated to use both of those?

> I looked at the ArduPilot code and it's going to get confused by this. It will cause Spektrum bind to start. Are there plans to support this in ArduPilot as well? If not, then I would say this show for PX4 only.

I'll take a look. I think I can use the enum and we can update ArduPilot

DonLakeFlyer:
Ok, so assuming ArduPilot gets updated as well. Then I'll merge once the magic numbers are fixed.

DonLakeFlyer:
Fantastic, thanks

PR #12081, opened 2024-11-09
Remove qt5compat Module

HTRamsey:
Attempt 2 at this. Since the only reason we needed this lib was for the ColorOverlay, just implement that one element and remove the rest of the module. This prepares for whenever this module is removed entirely (probably closer to Qt7) and also reduces the amount that is downloaded for the build.

Tested and works just like normal


HTRamsey:
@DonLakeFlyer Maybe it's better to just hold on to this until something happens to the qt5compat module rather than integrate this right now, not sure.

DonLakeFlyer:
Yeah I agree, it's kind of a bunch of extra cruft to add.

DonLakeFlyer:
@HTRamsey Close this for now?

PR #12091, opened 2024-11-12
Add Close Application Button in Select Tool Menu for Touch-Only Devices

NiklasVoigt:
This commit introduces a "Close Application" button to the Select Tool menu in QGroundControl. The addition addresses the need for an accessible close option on touch-only devices that lack a physical keyboard (e.g., Steam Deck). By integrating this button directly into the interface, users can easily exit the application without requiring keyboard input, improving usability on mobile and handheld platforms.

Description
-----------

Test Steps
-----------

Checklist:
----------

- [ ] [Review Contribution Guidelines](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CONTRIBUTING.md).
- [ ] [Review Code of Conduct](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CODE_OF_CONDUCT.md).

Related Issue
-----------

### DIFF
--- a/src/UI/MainRootWindow.qml
+++ b/src/UI/MainRootWindow.qml
@@ -334,6 +334,18 @@ ApplicationWindow {
                             }
                         }
 
+                        SubMenuButton {
+                            id:                 closeButton
+                            height:             toolSelectDialog._toolButtonHeight
+                            Layout.fillWidth:   true
+                            text:               qsTr("Close QGroundControl")
+                            onClicked: {
+                                if (mainWindow.allowViewSwitch()) {
+                                    mainWindow.finishCloseProcess()
+                                }
+                            }
+                        }
+
                         ColumnLayout {
                             width:                  innerLayout.width
                             spacing:                0


AlexisTM:
This is enabling using the Steam Deck as a controller without wondering how to exit the application. While there is a way on the Steam Deck, it isn't the case if all handheld.

HTRamsey:
@NiklasVoigt What do you think about using `QGroundControl.appName` instead of "QGroundControl" so that the name changes with custom builds?

NiklasVoigt:
@HTRamsey i've changed the name to the variable, works nice.

DonLakeFlyer:
@AlexisTM It's a bit odd to see this show up on a regular desktop build. But I think the Steam Deck just runs Linux right not android? So we can't just show this on mobile builds. I wonder if making it show up only if the main window is full screen would work?

AlexisTM:
It just runs on Linux indeed. That would definitely work to prevent it to look strange for other users.

Exit QGC challenge, the new game by the creators of "Exit Vi"!

PR #12139, opened 2024-11-26
Fix crash on startup for older Android devices

rubenp02:
# Fix crash on startup for older Android devices

Description
-----------
The overload to the registerReceiver method with a flags parameter was introduced in API level 26 (Android 8.0), which leads to an instant crash on startup on devices running older versions of Android. Adjusted the implementation to use the previous signature of registerReceiver on devices with API levels below 26.

Checklist:
----------


HTRamsey:
@rubenp02 just so you know, when we switch to Qt6.8, the minimum supported android will be 9.0. https://doc.qt.io/qt-6.8/whatsnew68.html

PR #12145, opened 2024-11-27
Update to Qt 6.8

### LINKED ISSUE #12279: QGC crashing on exit
I get this:
`ASSERT failure in PlainType *QtGlobalStatic::ApplicationHolder<(anonymous namespace)::Q_QAS__qgcCorePluginInstance>::pointer() [QAS = (anonymous namespace)::Q_QAS__qgcCorePluginInstance]: "The application static was used without a QCoreApplication instance", file /Users/don/Qt/6.7.3/macos/lib/QtCore.framework/Headers/qapplicationstatic.h, line 55`

From here:
```
VideoManager::~VideoManager()
{
    for (VideoReceiverData &videoReceiver : _videoReceiverData) {
        QGCCorePlugin::instance()->releaseVideoSink(videoReceiver.sink);
        delete videoReceiver.receiver;
        videoReceiver.receiver = nullptr;
    }

    // qCDebug(VideoManagerLog) << Q_FUNC_INFO << this;
}
```

Stack:
Screenshot 2025-01-04 at 9 18 41 AM

HTRamsey:
This is an LTS and will be the version used for v5
Note: Qt6.7 is unusable for Android

closes #12279


HTRamsey:
I had considered supporting both 6.6.3 & 6.8.X (6.7.X is missing files for Android so we will not use that). There are a few new things in Qt 6.8 I want to use but it shouldn't be too hard to support both for a bit. Waiting for more info/decisions from Don too.

PR #12148, opened 2024-11-27
Custom Build: Android overrides for min/target sdk

DonLakeFlyer:
@HTRamsey Can you double-check my cmake stuff? I think it's right, but i'm no cmake expert.


HTRamsey:
Looks like the build is failing. Maybe it's not supposed to be a string? Not sure

DonLakeFlyer:
Hmm, strange. It seems to work for me locally. Is there some other way to do this so it's not a string?

DonLakeFlyer:
Option setting was backwards, duh!

DonLakeFlyer:
Ah! I was wondering where this went. I just fixed it as well at the same time as you.

HTRamsey:
Oh whoops my bad, I was in the middle of testing the new android stuff from the other PR somebody did so I went ahead and handled this too

DonLakeFlyer:
No problem at all. I was just confused when I couldn't find the open pull any more.

HTRamsey:
Ideally since Qt6.8 is LTS we would make that the base for QGC v5. @DonLakeFlyer changed the minimum supported to 25 for his herelink and it works fine. I think based on Qt Documentation it is supposed to be 26 right now but 25 seems to work as well so we just left it for custom builds to choose.

PR #12164, opened 2024-12-02
Ignore key released if autorepeat in slider switch

unjambonakap:
AutoRepeat sends key released events that reset the slider


DonLakeFlyer:
What OS is this. The change seems ok. But it doesn't make sense to me that the space bar reports that it was released when it really wasn't.

unjambonakap:
Archlinux
Thanks for having a look, so much system disparities :)

DonLakeFlyer:
Ok, odd. Thanks for the fix.

PR #12165, opened 2024-12-02
GuidedValueSlider Tweaks

DieBorr:
This commit addresses two main issues:

1. The indicator made with the canvas was asymmetric; the top line was thinner than the bottom one.

2. On large screens, there was an overlap with negative numbers where the minus sign was overlapping with the major ticks.

Here are two screenshots, one showing the problem and the other one with the fix.

Screenshot from 2024-12-03 00-31-03
Screenshot from 2024-12-03 00-29-38

### DIFF
--- a/src/FlightDisplay/GuidedValueSlider.qml
+++ b/src/FlightDisplay/GuidedValueSlider.qml
@@ -124,8 +124,9 @@ Item {
             Layout.fillHeight:  true
             contentWidth:       sliderContainer.width
             contentHeight:      sliderContainer.height
-            flickDeceleration:  0.5
+            flickDeceleration:  2000
             flickableDirection: Flickable.VerticalFlick
+            Layout.rightMargin: -2                          // Vertical Flickable width
 
             Item {
                 id:     sliderContainer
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Thanks for the fixes. There is a lot to like here given the fact that the control has been somewhat screwed up. I like the fact that your changes fix the problem. The things I don't like are:
* It increases the width of the control taking more real estate
* It adds a bunch of additional complexity in the way it solves the problem. Unfortunately added on top of a control which is already crazy to go back into and understand how it works.
* The changes in the flick deceleration slows down the flick making it hard to make large jumps. This was specifically made to flick fast to solve the problem of making large changes like were possible with the old exponential slider. That way it easy to flick, and then you just click/touch again where you want to stop. Or a single flick all the way to max alt.

GIven all of that what I've done is taken your changes and re-arranged them into another pull. In this pull I've also put a bunch of effort into making the code for the control more readable. While I was doing that I ended finding a number of other problems with the control which I also fixed. I'll detail all of that in the pull I'll put up tomorrow which is a variant of this one. Thanks again for putting the effort into make this complicated thing work better.

DieBorr:
Hi, I'm glad you find this useful. I've been away from the project for a few weeks, and what I found is that it has some great ideas that just need a little refinement.

> It increases the width of the control taking more real estate

Regarding this, while I was working on it, I couldn't think of another way to solve the overlap without making the numbers smaller. However, if you have any ideas on how to resolve this, I'd be happy to give it a try.

>The changes in the flick deceleration slows down the flick making it hard to make large jumps. This was specifically made to flick fast to solve the problem of making large changes 

I see, I didn't know that was the intended behavior.

> GIven all of that what I've done is taken your changes and re-arranged them into another pull. In this pull I've also put a bunch of effort into making the code for the control more readable. 

Perfect! Thanks for your time. My idea was just to show the new 'look' of the slider with these fixes, and once they were approved, just clean and organize better the code. I don't have much on my plate this week, so I'll review your commit and help if possible. Don't worry about the code organization; it's just a matter of time to have a solid structure around the object itself.

Issue #12018, opened 2024-10-16
Finish Remote Id refactor


DonLakeFlyer:
* Refactor toolbar indicator into App Settings and flight based indicator settings stuff: Discussion here on what goes where - https://github.com/mavlink/qgroundcontrol/pull/11993
* Make sure to add in GNSS Live serial port setup support

BluemarkInnovations:
See also this comment https://github.com/mavlink/qgroundcontrol/pull/11993#issuecomment-2437712499

DonLakeFlyer:
Let's do the simple UI movement first in one pull and then we can discuss/fix the things @BluemarkInnovations brought up.

PR #12168, opened 2024-12-04
Remove RemoteID Configurations from Toolbar

bytesByHarsh:
Description
-----------
Remove a few config settings related to remote ID as they are now shifted to Application Settings.

Test Steps
-----------
1) Start Mocklink
2) Toolbar -> Remote ID -> Extended Menu

image

Related Issue
-----------
https://github.com/mavlink/qgroundcontrol/issues/12018

### DIFF
--- a/src/UI/preferences/RemoteIDSettings.qml
+++ b/src/UI/preferences/RemoteIDSettings.qml
@@ -275,7 +275,9 @@ SettingsPage {
             }
             SettingsGroupLayout {
                 outerBorderColor: _activeRID ? (_remoteIDManager.armStatusGood ? defaultBorderColor : qgcPal.colorRed) : defaultBorderColor
+                visible:            armStatusLabel.labelText !== ""
                 LabelledLabel {
+                    id :                armStatusLabel
                     label:              qsTr("Arm Status Error")
                     labelText:          _remoteIDManager?_remoteIDManager.armStatusError:"Vehicle Not Connected"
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Couple things...

There is a bug which is causing the empty group layout box at the top. This is the group box for Arm Status Error. The 'visible: labelText !== ""' in there should be on the SettingsGroupLayout control not the Labelled Label. This way the whole group goes away when not needed. This bug might also be in the full settings page.

The expanded page should include a "Configure" button which takes you to the full setup page. You can see it here inthe Battery page:
Screenshot 2024-12-09 at 9 44 48 AM

bytesByHarsh:
> The expanded page should include a "Configure" button which takes you to the full setup page. You can see it here in the Battery page:

Already on clicking `Arm Status`, `GCS GPS` etc it is taking to the configuration page. Then should we add it again?


> There is a bug which is causing the empty group layout box at the top. This is the group box for Arm Status Error. The 'visible: labelText !== ""' in there should be on the SettingsGroupLayout control not the Labelled Label. This way the whole group goes away when not needed. This bug might also be in the full settings page.

image

image

If the above changes seem fine then I'll push the updated code.

bytesByHarsh:
@DonLakeFlyer let me know if this way is fine:

image

DonLakeFlyer:
Yup, perfect

PR #12169, opened 2024-12-04
Add two phase init to QGCCorePlugin

DonLakeFlyer:


HTRamsey:
Some issue about the PX4 libevents repo not being accessible. GitHub is lagging really bad right now so maybe related?

PR #12170, opened 2024-12-04
Vehicle: ignore severity of PX4 preflight messages

MaEtUgR:
### Problem
We had someone change this condition out of confusion downstream and I figured out there is no message with emergency severity containing the string "preflight" in the current PX4 version. I can still check if and where they might have existed on older versions but I'd suggest to remove this since for the events it doesn't make sense to sort the text by severity.

### Solution
Ignore the severity. I think this was added originally such that emergency messages get output as yellow box and spoken even though they had the word "preflight" in it which usually resulted in a transparent box over the map.

### Test Steps
I have not tested the change but went through the existing cases and it shouldn't make a difference with the current version of PX4.

### Checklist:
- [ ] Check what messages ever fulfilled this condition.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -3997,7 +3997,7 @@ void Vehicle::_textMessageReceived(MAV_COMPONENT componentid, MAV_SEVERITY sever
 
     bool skipSpoken = false;
     const bool ardupilotPrearm = text.startsWith(QStringLiteral("PreArm"));
-    const bool px4Prearm = text.startsWith(QStringLiteral("preflight"), Qt::CaseInsensitive) && (severity >= MAV_SEVERITY::MAV_SEVERITY_CRITICAL);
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
That severity check certainly seems odd. Could you also clean this up such that the two bools first check the firmware type on the vehicle as well. They are firmware specific tests so they should only run on vehicles with that firmware.

So for example:
`const bool ardupilotPrearm = apmFirmware() && text.startsWith(QStringLiteral("PreArm"));`

DonLakeFlyer:
@MaEtUgR Can you update Matthias?

MaEtUgR:
> `const bool ardupilotPrearm = apmFirmware() && ...`

Ha interesting! I might suggest this fix separately then. I'm not sure my fix does not have side effects, that's also why it's a draft pr atm. In fact I think I found a good reason to leave this condition as is today: There are certain event based failures that are important enough to read out loud and have pop up in a yellow box and this additional condition basically allows to do that from PX4 by sending an event with high severity.

DonLakeFlyer:
So close this then?

Issue #10046, opened 2021-11-28
Camera button is off-center


HTRamsey:
Critical program breaking bug! The camera button is not centered.

![Screenshot from 2021-11-28 16-01-49](https://user-images.githubusercontent.com/68555040/143785860-1fdf22c7-29f6-4c6e-97fd-f7c285a408cc.png)

PR #12179, opened 2024-12-09
Fix centering of inner circle in PhotoVideoControl

rubenp02:
# Fix centering of inner circle in PhotoVideoControl

Description
-----------
The inner circle in the PhotoVideoControl widget was not always perfectly centered due to a bug in Qt's anchors.centerIn property (QTBUG-95224) which causes it to return integer positions instead of subpixel values.

Implemented a workaround by manually calculating the center position, ensuring exact centering regardless of parent and child dimensions and DPI scaling.

This fix should be reverted once QGC moves to Qt 6.8.1 or higher, where the underlying Qt bug has been resolved.

Related Qt bug: https://bugreports.qt.io/browse/QTBUG-95224

Qt fix commit: https://github.com/qt/qtdeclarative/commit/fd23a222efe189607eebd5c6782ca73eafa7080c

Checklist:
----------

Related Issue
-----------
#10046

### DIFF
--- a/src/FlightMap/Widgets/PhotoVideoControl.qml
+++ b/src/FlightMap/Widgets/PhotoVideoControl.qml
@@ -165,7 +165,13 @@ Rectangle {
                     border.width:       3
 
                     Rectangle {
-                        anchors.centerIn:   parent
+                        // anchors.centerIn snaps to integer coordinates, which
+                        // depending on DPI can throw the centering off.
+                        // Setting alignWhenCentered to false avoids this issue.
+                        anchors {
+                            centerIn:           parent
+                            alignWhenCentered:  false
+                        }
                         width:              parent.width * (_isShootingInCurrentMode ? 0.5 : 0.75)
                         height:             width
                         radius:             _isShootingInCurrentMode ? 0 : width * 0.5


DonLakeFlyer:
> due to a mismatch in the parity of the parent and child widths

Not quite sure what you mean by parity? Is this some sort of sub-pixel positioning problem?

DonLakeFlyer:
@HTRamsey What do you think the likelihood is that we should move to 6.8.1 before the next Stable? Is there anything holding that up?

HTRamsey:
We can pretty much do it quickly today if you want. I think it builds as-is but there was a small change in linking QtMultimedia for Windows, the cmake translations command changes a bit, and I think we get a warning about the QSqlDatabase that needs a small fix. I had started to test Qt6.7 before doing Qt6.8 in #12145

DonLakeFlyer:
The problem is this centerIn bug is pretty major. This specific example here is just the tip of the iceberg with respect to QGC and use of centerIn with real number pixel positions.. That type of anchoring is used all over the place. Doesn't make sense to fix this one place when there is another 50 points of use. So I think moving to a new Qt would be worthwhile.

HTRamsey:
@rubenp02 sounds like maybe we should just skip this and jump straight to 6.8.1. Does that sound alright in your opinion?

rubenp02:
Since 6.6.3 still has official support and we have issue [#12977](https://github.com/mavlink/qgroundcontrol/issues/12977) to track any changes that need to be reverted once support is dropped, maybe this should be merged and added to that issue?

rubenp02:
For some reason this is still not centered even in 6.8.3, so I've reopened the PR.

DonLakeFlyer:
Yeah, I'll merge this in tomorrow when I have time to cherry pick it to Stable as well.

DonLakeFlyer:
@rubenp02 Isn't the fix for this to use: `anchors.alignWhenCentered: false`? According to bug that should work on 6.6 and up.

DonLakeFlyer:
@rubenp02 ?

rubenp02:
@DonLakeFlyer Sorry for the delay, I was out of town. I checked that what you said fixes the issue (it does, thanks!) and force pushed the updated solution. I'm still not sure as to why the bug is still happening on 6.8.3 though, the QTBUG I mention on the commit message has been merged since 6.8.1.

DonLakeFlyer:
Isn't the bug that `anchors.alignWhenCentered: false` wasn't working correctly and then with the fix it does. As far as I can tell anchors.centerIn is supposed to use integer positioning by default.

PR #12183, opened 2024-12-09
GuidedValueSlider: Rework to fix margin issues and other things

DonLakeFlyer:
Replacement for #12165

While looking at #12165 I discovered other issues and also wanted to rework it so that the code was clearer. So this pull is based on that pull with some additional changes.

Changes:
* The slider is now parented right up to the top/bottom/right edges of window. Previously there was a small margin around it. This could cause problems where you would mistakenly zoom the map when you really wanted to scroll the slider if you were off the edge.
* Mouse events were bleeding through to the map below. So if you happens to scroll over the slider left/right using a track pad the map would zoom below it. There is not a dead mouse area below the control to prevent mouse events from bleeding through controls below it.
* There were various problems with the margins between the ticks and both negative and large positive values. Sometimes the values would overlap with the major ticks. To fix this the margins have been reworked so that there should always be room for a max of three digits (with/without negative)without bumping into the ticks.
* The indicator was not drawing correctly in that the border was not equally sized around the indicator.
* Since the margins are now larger I made the tick marks slightly smaller so that the overall width of the control didn't grow.
* Reworked a number of property names/usage to make the code easier to understand.

### DIFF
--- a/src/FlightDisplay/FlyView.qml
+++ b/src/FlightDisplay/FlyView.qml
@@ -178,7 +178,6 @@ Item {
         //-- Guided value slider (e.g. altitude)
         GuidedValueSlider {
             id:                 guidedValueSlider
-            anchors.margins:    _toolsMargin
             anchors.right:      parent.right
             anchors.top:        parent.top
             anchors.bottom:     parent.bottom
--- a/src/FlightDisplay/GuidedValueSlider.qml
+++ b/src/FlightDisplay/GuidedValueSlider.qml
@@ -45,7 +45,7 @@ Item {
 
     property real   _majorTickWidth:        ScreenTools.largeFontPixelWidth * 2
     property real   _majorTickPixelHeight:  ScreenTools.largeFontPixelHeight * 2
-    property real   _majorTickValueMargin:  ScreenTools.defaultFontPixelWidth / 2
+    property real   _tickValueRightMargin:  ScreenTools.defaultFontPixelWidth / 2
     property real   _minorTickWidth:        _majorTickWidth / 2
     property real   _sliderValuePerPixel:   _majorTickValueStep / _majorTickPixelHeight
 
@@ -156,7 +156,7 @@ Item {
                         }
 
                         QGCLabel {
-                            anchors.margins:        _majorTickValueMargin
+                            anchors.margins:        _tickValueRightMargin
                             anchors.right:          parent.right
                             anchors.verticalCenter: majorTick.verticalCenter
                             text:                   parent.tickValue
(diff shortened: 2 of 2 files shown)


DonLakeFlyer:
@DieBorr Can you review this as well as run it to make sure it works correctly on you system as well?

HTRamsey:
This might just be me but it would be kinda nice to be able to quickly see Max/Min values which were easier to tell on the older slider since you could see how far up the slider you were.

DieBorr:
Hi @DonLakeFlyer,

I’ve checked your rework, and while the logic of the object is still a bit complicated, it’s definitely easier to read now. However, I did notice an issue that was the main reason why I initially made the slider width bigger. Here’s a screenshot:

image

As you can see, on a monitor that’s not very big (26/27"), there’s some visible overlap. Also, the white rectangle of the vertical flickable feels a bit too close to the tick numbers. Since this widget is pretty important, I think adding a bit more width wouldn’t be a big problem. You can take my fix where I just take the rectangle near to the slider but not inside.

From my point of view, the workaround in #12165 about dynamically changing the slider width feels less user-friendly, I mean, the workaround itself is nice, but I found it too noticeable.  Here’s a video to show what I mean:

https://github.com/user-attachments/assets/04379574-b629-4f28-b48f-abc171e1dc07

I find this width change too rough because it moves the widgets next to it in a aggressive way, with ft as metric system I find it smother, this should be a minor fix and is just a different point of view I think  that @HTRamsey  can give his opinion.

This are just little tweaks around the slider, spending time coding the base of the slider and not being able to leave it as refined as possible would be a shame. 
Thank you both for your time.

DonLakeFlyer:
Hmm, strange. I tested out to "-100" and don't see the overlap. Not sure why I see different things. Let me look again.

> Also, the white rectangle of the vertical flickable feels a bit too close to the tick numbers

Agree, I'll take a look at that as well.

DonLakeFlyer:
@DieBorr I'm hoping my last change will fix the spacing between tick marks and values. Can you try it again?

DieBorr:
> @DieBorr I'm hoping my last change will fix the spacing between tick marks and values. Can you try it again?

Yes! Will try in this day as soon as I can

DieBorr:
Congrats @DonLakeFlyer, it's a very clever solution. I have tested it with different UI scale values and also different font scale factors of the OS (ubuntu 22.04), and now it works perfectly fine. Thanks for the time you spent on it.

I'm planning to add a new feature to the slider. It consists of having a popup appear if the user holds the text value label for a specific time, allowing the user to select a specific value. Do you think this makes sense, or would it not be useful?

DonLakeFlyer:
> now it works

Quite a twisted bit of spacing/margin logic to get right. Kind of makes my head hurt to think about it.

> Do you think this makes sense, or would it not be useful?

Yeah, I think that would be super nice.

PR #12191, opened 2024-12-10
Fix case-sensitive flight mode comparisons

### LINKED ISSUE #12128: Standard Mode issue with ArduCopter 4.6
Standard Modes are all caps and when comparing to a flight mode name like "Guided" trying to set the flight mode ends up failing

rubenp02:
# Fix case-sensitive flight mode comparisons

Description
-----------
Fixed a bug where the "Start mission" action failed on ArduPilot, displaying an error popup with the message "Unable to start mission: Vehicle failed to change to Auto mode." The issue was caused by case-sensitive comparisons between flight mode strings from the vehicle and the reference strings. Moving away from using strings entirely should be considered.

Replaced direct string comparisons with case-insensitive comparisons using QString::compare() with Qt::CaseInsensitive flag. This fix resolves the immediate issue and also corrects a few other incorrect mode comparisons unrelated to the "Start mission" problem.

Closes #12128 

Checklist:
----------

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSubMotorComponentController.cc
+++ b/src/AutoPilotPlugins/APM/APMSubMotorComponentController.cc
@@ -22,7 +22,7 @@ void APMSubMotorComponentController::handleNewMessages(int uasid, int componenti
     Q_UNUSED(uasid);
     Q_UNUSED(componentid);
     Q_UNUSED(severity);
-    if (_vehicle->flightMode() == "Motor Detection"
+    if (QString::compare(_vehicle->flightMode(), "Motor Detection", Qt::CaseInsensitive) == 0
         && (text.toLower().contains("thruster") || text.toLower().contains("motor"))) {
         _motorDetectionMessages += text + QStringLiteral("\n");
         emit motorDetectionMessagesChanged();
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -670,7 +670,7 @@ const QVariantList& APMFirmwarePlugin::toolIndicators(const Vehicle* vehicle)
 
 bool APMFirmwarePlugin::isGuidedMode(const Vehicle* vehicle) const
 {
-    return vehicle->flightMode() == "Guided";
+    return QString::compare(vehicle->flightMode(), "Guided", Qt::CaseInsensitive) == 0;
 }
 
 void APMFirmwarePlugin::_soloVideoHandshake(void)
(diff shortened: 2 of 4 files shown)


HTRamsey:
Do you think this resolves #12128?

DonLakeFlyer:
I don't quite see where these all cap flight mode names end up coming into QGC. How does that happen?

DonLakeFlyer:
If these mode names now come from the mavlink protocol instead of being hardwired there is more stuff broken then this. For example, here: https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc#L35. The px4 flight mode names are translated. If the comparison flight mode string comes from the vehicle then it most likely is not translated. Or maybe it is but who knows. @bkueng?

HTRamsey:
@DonLakeFlyer if you have a master branch up for ArduPilot SITL you should immediately notice the issue

bkueng:
PX4 doesn't use all caps for modes. The change is fine with me, but I think it's worth considering adjusting ardupilot.
It was recently added here: https://github.com/ArduPilot/ardupilot/pull/28566

> If these mode names now come from the mavlink protocol instead of being hardwired there is more stuff broken then this. For example, here: https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc#L35. The px4 flight mode names are translated. If the comparison flight mode string comes from the vehicle then it most likely is not translated. Or maybe it is but who knows. @bkueng?

It should work as this list of modes isn't used outside of the firmware plugin when the modes come from mavlink. And the guided actions (e.g. [takeoff](https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc#L392)) either directly use the mavlink command, or the mode object (`_rtlFlightMode`).
Translation isn't supported yet when using the standard modes protocol.

HTRamsey:
So where do we stand with this then? At a minimum it won't break anything

DonLakeFlyer:
I haven't had time to look at the new mavlink flight mode spec. I want to see if with that we can switch to not using strings internally any longer at all. This whole case sensitive problem, string compare stuff makes my skin crawl from a standpoint of being crappy.

DonLakeFlyer:
Ok, so this is fine for now. But @bytesByHarsh is going to help with doing a full review of this stuff. At a minimum we should talk with ArduPilot folks about why flight modes names are all upper cases. Given we transitioned to the 21st century century a while back. I think we can move away from a stuck caps lock key for users. I'm going to write up an Issue with things I am concerned about and investigation as to what we can do about it.

DonLakeFlyer:
> It shouldn't matter if ArduPilot sends the mode names in whatever case

They are shown to the user so to me it does matter.

bkueng:
There's also a set of standard modes that can be identified by enum: https://mavlink.io/en/messages/development.html#MAV_STANDARD_MODE

DonLakeFlyer:
FYI: @bytesByHarsh Is working on a full solution to this so we don't have to always remember to do case sensitive compares. We've been discussion it for a few days now and he has already started on the the work. Should hopefully come in soon. So with that pull this won't be needed.

DonLakeFlyer:
Replaced by #12254

Issue #11620, opened 2024-06-17
ardupilot ALT_HOLD_RTL


tbo47:
## Current Behavior

ALT_HOLD_RTL disappeared from ardupilot from 4.4 to 4.5

https://ardupilot.org/plane/docs/parameters-Plane-stable-V4.5.0.html
https://ardupilot.org/plane/docs/parameters-Plane-stable-V4.4.4.html

and qgroundcontrol shows a popup saying it is missing :
src/AutoPilotPlugins/APM/APMSafetyComponentSummary.qml:149

## Steps to Reproduce:
1. qgroundcontrol 4.3 and ardupilot 4.5.4 on Pixhawk 6c
2.  in qgroundcontrol go to the Vehicle setup
3. the error message will appear after a couple of seconds

## System Information
- Operating System: debian 12
- QGC Version: 4.3.0
- QGC build: flatpak build
- Flight Controller: Pixhawk 6c
- Autopilot: Ardupilot 4.5.4

PR #12195, opened 2024-12-11
Update RTL altitude parameter for ArduPlane 4.5+

rubenp02:
# Update RTL altitude parameter for ArduPlane 4.5+

Description
-----------
The ALT_HOLD_RTL parameter was removed in ArduPlane 4.5 and replaced with RTL_ALTITUDE. This change caused the Safety Component to display error messages when connecting to vehicles running newer firmware versions.

Added logic to use 'RTL_ALTITUDE' for ArduPilot versions 4.5 and above, while maintaining 'ALT_HOLD_RTL' for older versions.

Checklist:
----------

Related Issue
-----------
#11620

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSafetyComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMSafetyComponent.qml
@@ -660,7 +660,13 @@ SetupPage {
                 Column {
                     spacing: _margins / 2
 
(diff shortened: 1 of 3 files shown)


HTRamsey:
Ideally we should eventually somehow make some kinda lookup function that takes in the current firmware version instead of having to do verbose checks like this everywhere.

PR #12205, opened 2024-12-14
Rework Scroll indicator

DonLakeFlyer:
* No longer a margin on edge anchoring
* Slightly larger, based on font widths
* Single control which cleans up duplication
* Related to discussion in #12183


DonLakeFlyer:
@DieBorr Can you look at this one as well to see what you think?

DieBorr:
It's fine, the only downside is what we discussed earlier, in objects that have elements near the right side the vertical flickable can overlap a litle bit, for example in the guided value slider this is what happens:
Screenshot from 2024-12-14 23-50-55

I have also made a pull request related with indentation.

DonLakeFlyer:
That's an image using the old slider code. The new slider code has a margin on the right.

DieBorr:
> That's an image using the old slider code. The new slider code has a margin on the right.

Sorry if there is a mistake, will take a look tomorrow!

PR #12210, opened 2024-12-15
Comms: Speed up UDPLink Reception

HTRamsey:
Small changes to reduce time spent in reception socket


menschel:
Causes 
```
qgc.comms.udplink: Failed to join multicast group "224.0.0.1"
qgc.comms.udplink: Communication error: "Failed to join multicast group"
```

HTRamsey:
Oh that was silently ignored previously, I will remove the error signal

PR #12238, opened 2024-12-19
Various Landing Pattern fixes and add final approach speed

### LINKED ISSUE #12266: VTOL Landing Pattern gets split into multiple waypoints
## Expected Behavior
VTOL Landing Pattern should always display as one waypoint called "Landing Pattern" in the mission planning editor.

## Current Behavior
VTOL Landing Pattern only displays as one waypoint when you create a new mission and upload it.
If you later download that same mission, it displays as "Land start", "Camera trigger distance", "Stop image capture", "Stop video capture", "Loiter (altitude)", and "VTOL land".

Before:
Screenshot from 2024-12-30 15-03-05

After:
Screenshot from 2024-12-30 15-03-42

This is very confusing to the user. In contrast, the Fixed-Wing Landing Pattern always displays as "Landing Pattern".
## Steps to Reproduce:
1. Upload a VTOL mission with a landing pattern
2. Download that mission by using File > Download, or by reopening QGC

## System Information
- Operating System: Windows, Linux, Android
- QGC Version: 4.4.3
- QGC build: Stable
- Flight Controller: SITL, Cubepilot Cube Orange +
- Autopilot (with version): PX4 v1.15

## Detailed Description
Provide further details about your issue/bug.

## Log Files and Screenshots
- [QGC Console Logs](https://docs.qgroundcontrol.com/en/settings_view/console_logging.html)
- Autopilot logs when available (post a link).
- Screenshots of QGC to help identify the current issue/bug behavior.

rubenp02:
# Various Landing Pattern fixes and add final approach speed

Description
-----------
### [Fix landing item scanning for ArduPilot fixed wing](https://github.com/mavlink/qgroundcontrol/commit/10374feaf0549b187a78d7754530b6b4acd0b909):

Addressed issues with scanning and interpreting landing sequences for fixed wing ArduPilot aircraft. Previously, landing sequences uploaded to ArduPilot fixed wing aircraft and then downloaded back weren't being reinterpreted as QGC LandingComplexItems. The system to detect a landing sequence and translate it back to a LandingComplexItem wasn't working correctly due to multiple assumptions about mission item parameters that weren't accurate for ArduPilot.

The changes include adjusting parameter checks for NAV_LAND, NAV_LOITER_TO_ALT, NAV_WAYPOINT, and DO_LAND_START commands to align with ArduPilot's implementation, without modifying the existing logic for other firmwares.

### [Fix waypoint glide slope heading calculations](https://github.com/mavlink/qgroundcontrol/commit/a7d3240d47bf28e557ddf0a3664893c49c3b82fa):

When not using the "Use loiter to altitude" option in the Landing Pattern mission item, the heading of the glide slope still took into account the old loiter radius. Adjusted the QML code to change the loiter radius to 0 in landing patterns that use a waypoint as the approach start, which in turns lead to correct heading calculations.

Ideally, the _recalcFromHeadingAndDistanceChange, _recalcFromRadiusChange and _recalcFromCoordinateChange should be refactored to account for the fact that a landing pattern can use a waypoint instead of a loiter. However, this fix works just as well and is much simpler.

### [Fix boolean Fact interpretation in QGCCheckBox](https://github.com/mavlink/qgroundcontrol/commit/27d1ad0c20500ebb33c04d615c768d61666013bb):

Addressed an issue where the values of boolean Facts where the QVariant was of type int were not being correctly interpreted to drive the checked status of the QGCCheckBox. This was a problem with boolean Facts loaded from JSON where the default was "false", in which case the checkbox didn't work correctly.

### [Add final approach speed to the Landing Pattern](https://github.com/mavlink/qgroundcontrol/commit/cda7b95dd601acbd98a9091178c72fb1eaf93c0a):

Introduced an option for setting an airspeed for the final approach in the Landing Pattern editor, almost identical to the one found in standard waypoints, for the purposes of energy management before the glide slope.

It works by inserting a DO_CHANGE_SPEED mission element right after the DO_LAND_START, in such a way that the speed is also changed in the event of an RTL. The option is unchecked by default.

### [Fix VTOL landing item scanning for PX4](https://github.com/mavlink/qgroundcontrol/commit/a8f6c0080604107b8109f64948f5b0f35ebd0d57):

Fixes #12266 (fix by @anaam-wingxpand).

Checklist:
----------

### DIFF
--- a/src/FactSystem/FactControls/FactCheckBox.qml
+++ b/src/FactSystem/FactControls/FactCheckBox.qml
@@ -14,8 +14,8 @@ QGCCheckBox {
     Binding on checkState {
         value: fact ?
                    (fact.typeIsBool ?
-                        (fact.value === false ? Qt.Unchecked : Qt.Checked) :
-                        (fact.value === 0 ? Qt.Unchecked : Qt.Checked)) :
+                        (fact.value ? Qt.Checked : Qt.Unchecked) :
+                        (fact.value !== 0 ? Qt.Checked : Qt.Unchecked)) :
                    Qt.Unchecked
     }
 
--- a/src/MissionManager/LandingComplexItem.cc
+++ b/src/MissionManager/LandingComplexItem.cc
@@ -388,14 +388,19 @@ bool LandingComplexItem::_scanForItem(QmlObjectListModel* visualItems, bool flyV
     MissionItem& missionItemFinalApproach = item->missionItem();
     if (missionItemFinalApproach.command() == MAV_CMD_NAV_LOITER_TO_ALT) {
         if (missionItemFinalApproach.frame() != landPointFrame ||
-                missionItemFinalApproach.param1() != 1.0 || missionItemFinalApproach.param3() != 0 || missionItemFinalApproach.param4() != 1.0) {
+            (masterController->managerVehicle()->apmFirmware()
+             // APM automatically changes the value of param1 to 1, so when sending a plan it will
+             // be 0, and when downloading it, the value will be 1
+             ? missionItemFinalApproach.param1() != 0.0 && missionItemFinalApproach.param1() != 1.0
+             : missionItemFinalApproach.param1() != 1.0) ||
+            missionItemFinalApproach.param3() != 0 || missionItemFinalApproach.param4() != 1.0) {
             return false;
         }
     } else if (missionItemFinalApproach.command() == MAV_CMD_NAV_WAYPOINT) {
         if (missionItemFinalApproach.frame() != landPointFrame ||
-                missionItemFinalApproach.param1() != 0 || missionItemFinalApproach.param2() != 0 || missionItemFinalApproach.param3() != 0 ||
-                !qIsNaN(missionItemFinalApproach.param4()) ||
-                qIsNaN(missionItemFinalApproach.param5()) || qIsNaN(missionItemFinalApproach.param6()) || qIsNaN(missionItemFinalApproach.param6())) {
+            missionItemFinalApproach.param1() != 0 || missionItemFinalApproach.param2() != 0 || missionItemFinalApproach.param3() != 0 ||
+            (!masterController->managerVehicle()->apmFirmware() && !qIsNaN(missionItemFinalApproach.param4())) ||
+            qIsNaN(missionItemFinalApproach.param5()) || qIsNaN(missionItemFinalApproach.param6()) || qIsNaN(missionItemFinalApproach.param6())) {
             return false;
         }
         useLoiterToAlt = false;
@@ -425,9 +430,8 @@ bool LandingComplexItem::_scanForItem(QmlObjectListModel* visualItems, bool flyV
     }
     MissionItem& missionItemDoLandStart = item->missionItem();
     if (missionItemDoLandStart.command() != MAV_CMD_DO_LAND_START ||
-            missionItemDoLandStart.frame() != MAV_FRAME_MISSION ||
-            missionItemDoLandStart.param1() != 0 || missionItemDoLandStart.param2() != 0 || missionItemDoLandStart.param3() != 0 || missionItemDoLandStart.param4() != 0 ||
-            missionItemDoLandStart.param5() != 0 || missionItemDoLandStart.param6() != 0 || missionItemDoLandStart.param7() != 0) {
+        missionItemDoLandStart.param1() != 0 || missionItemDoLandStart.param2() != 0 || missionItemDoLandStart.param3() != 0 || missionItemDoLandStart.param4() != 0 ||
+        missionItemDoLandStart.param5() != 0 || missionItemDoLandStart.param6() != 0 || missionItemDoLandStart.param7() != 0) {
         return false;
     }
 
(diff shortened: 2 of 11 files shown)


src/MissionManager/LandingComplexItem.cc
```diff
@@ -282,6 +284,10 @@ void LandingComplexItem::appendMissionItems(QList<MissionItem*>& items, QObject*
     MissionItem* item = _createDoLandStartItem(seqNum++, missionItemParent);
     items.append(item);
 
+    if (useDoChangeSpeed()->rawValue().toBool()) {
+        item = _createDoChangeSpeedItem(0, finalApproachSpeed()->rawValue().toDouble(), 0, seqNum++, missionItemParent);
```
DonLakeFlyer, line 288: Should use SPEED_TYPE_AIRSPEED instead of 0. Also I think throttle should be -1. That's how the normal change speed item works: SpeedSection.cc

src/MissionManager/LandingComplexItem.cc
```diff
@@ -388,13 +407,15 @@ bool LandingComplexItem::_scanForItem(QmlObjectListModel* visualItems, bool flyV
     MissionItem& missionItemFinalApproach = item->missionItem();
     if (missionItemFinalApproach.command() == MAV_CMD_NAV_LOITER_TO_ALT) {
         if (missionItemFinalApproach.frame() != landPointFrame ||
-                missionItemFinalApproach.param1() != 1.0 || missionItemFinalApproach.param3() != 0 || missionItemFinalApproach.param4() != 1.0) {
+            (masterController->managerVehicle()->firmwareType() == MAV_AUTOPILOT_ARDUPILOTMEGA ?
```
DonLakeFlyer, line 410: Should use `masterController->managerVehicle()->apmFirmware`. Same in other places

src/FactSystem/FactControls/FactCheckBox.qml
```diff
@@ -14,8 +14,8 @@ QGCCheckBox {
     Binding on checkState {
         value: fact ?
                    (fact.typeIsBool ?
-                        (fact.value === false ? Qt.Unchecked : Qt.Checked) :
-                        (fact.value === 0 ? Qt.Unchecked : Qt.Checked)) :
+                        (fact.value ? Qt.Checked : Qt.Unchecked) :
```
DonLakeFlyer, line 17: Was this change needed to fix some problem?

DonLakeFlyer (changes requested):
Looks good in general. I made some comments which need looking at. I also want to take a second look more in depth once those get addressed

DonLakeFlyer:
@rubenp02 Can you make sure #12266 is taken care of as well?

src/MissionManager/LandingComplexItem.cc
```diff
@@ -425,7 +467,9 @@ bool LandingComplexItem::_scanForItem(QmlObjectListModel* visualItems, bool flyV
     }
     MissionItem& missionItemDoLandStart = item->missionItem();
     if (missionItemDoLandStart.command() != MAV_CMD_DO_LAND_START ||
-            missionItemDoLandStart.frame() != MAV_FRAME_MISSION ||
+            (masterController->managerVehicle()->firmwareType() == MAV_AUTOPILOT_ARDUPILOTMEGA ?
```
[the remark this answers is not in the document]
DonLakeFlyer: > APM automatically changes it back to a MAV_FRAME_GLOBAL

That's quite odd. Oh well.

> I think these checks are a bit too strict

I think you are correct. For ones which don't specify values which are affected by frame then I think the best thing to do is remove the frame checks from those.

anaam-wingxpand:
i fixed it by making the VTOL isValidLandItem function check for MAV_CMD_NAV_VTOL_LAND, instead of MAV_CMD_NAV_LAND.

`bool VTOLLandingComplexItem::_isValidLandItem(const MissionItem& missionItem)
{
    if (missionItem.command() != MAV_CMD_NAV_VTOL_LAND ||`

src/MissionManager/LandingComplexItem.cc
```diff
@@ -415,6 +436,27 @@ bool LandingComplexItem::_scanForItem(QmlObjectListModel* visualItems, bool flyV
...
 
+    scanIndex--;
+    bool useDoChangeSpeed = false;
+    double finalApproachSpeed = 0;
+    if (scanIndex >= 0 && scanIndex < visualItems->count()) {
+        item = visualItems->value<SimpleMissionItem*>(scanIndex);
+        if (item) {
+            MissionItem& missionItemChangeSpeed = item->missionItem();
+            if (missionItemChangeSpeed.command() == MAV_CMD_DO_CHANGE_SPEED &&
+                missionItemChangeSpeed.param1() >= 0 && missionItemChangeSpeed.param1() <= 3 &&
```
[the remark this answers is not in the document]
DonLakeFlyer: > But that would mean that a mission sequence with a manually configured DO_CHANGE_SPEED item wouldn't draw a glide slope and such.

It wouldn't be a manually configured change speed, it would be a fully manually configured landing pattern. The change speed is after the DO_LAND_START. If you use the pattern, there is not way to manually get something in the middle of those items.

You are slurping a change speed command off the stack which you can't show the correct ui for in the pattern editor since the pattern editor only supports airspeed. The user won't really know what is happening.

DonLakeFlyer (approved):

DonLakeFlyer:
Thanks so much for this one @rubenp02

PR #12254, opened 2024-12-23
Standard methods for parsing available flight mode for PX4 and APM

bytesByHarsh:
Update APM and PX4 flight mode parsing methods

Description
-----------
- Removed use of direct strings to change flight mode.
- Case-sensitive issues in flight mode in APM.
- Internal Mapping to flight mode names.
- QGC Custom Build support has also been added.
- Add New Modes discovered via Available Mode packets received from the flight controller.

Related Issue
-----------
https://github.com/mavlink/qgroundcontrol/pull/12191
https://github.com/mavlink/qgroundcontrol/issues/12128

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -553,4 +553,12 @@ elseif(MACOS)
     install(SCRIPT "${CMAKE_SOURCE_DIR}/cmake/CreateMacDMG.cmake")
 endif()
 
+## Enforce Handling of All Enum Cases
+
+if(WIN32)
+    add_compile_options(/w44265 /we44265)
+else()
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wswitch -Werror=switch")
+endif()
+
 include(printSummary)
--- a/src/FirmwarePlugin/FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/FirmwarePlugin.cc
@@ -72,7 +72,7 @@ QString FirmwarePlugin::flightMode(uint8_t base_mode, uint32_t custom_mode) cons
     if (base_mode == 0) {
         flightMode = "PreFlight";
     } else if (base_mode & MAV_MODE_FLAG_CUSTOM_MODE_ENABLED) {
-        flightMode = QString("Custom:0x%1").arg(custom_mode, 0, 16);
+        flightMode = _modeEnumToString.value(custom_mode, QString("Custom:0x%1").arg(custom_mode, 0, 16));
     } else {
         for (size_t i=0; i<sizeof(rgBit2Name)/sizeof(rgBit2Name[0]); i++) {
             if (base_mode & rgBit2Name[i].baseModeBit) {
@@ -570,11 +570,7 @@ Autotune* FirmwarePlugin::createAutotune(Vehicle *vehicle)
 
 void FirmwarePlugin::updateAvailableFlightModes(FlightModeList modeList)
 {
-    _availableFlightModeList.clear();
-
-    for(auto mode: modeList){
-        _updateModeMappings(mode);
-    }
+    _updateModeMappings(modeList);
 }
 
 void FirmwarePlugin::_setModeEnumToModeStringMapping(FlightModeCustomModeMap enumToString)
@@ -582,17 +578,22 @@ void FirmwarePlugin::_setModeEnumToModeStringMapping(FlightModeCustomModeMap enu
     _modeEnumToString = enumToString;
 }
 
-void FirmwarePlugin::_updateModeMappings(FirmwareFlightMode &mode){
-    // Get Presaved Mode String, if mode is not present then take Custom Mode Name
+void FirmwarePlugin::_updateModeMappings(FlightModeList &modeList){
 
-    QString nModeName = mode.mode_name;
-    if(_modeEnumToString.contains(mode.custom_mode)){
-        nModeName = _modeEnumToString[mode.custom_mode];
+    _availableFlightModeList.clear();
+
+    for(auto &mode:modeList){
+        // Get Presaved Mode String, if mode is not present then take Custom Mode Name
+
+        QString nModeName = mode.mode_name;
+        if(_modeEnumToString.contains(mode.custom_mode)){
+            nModeName = _modeEnumToString[mode.custom_mode];
+        }
+        else{
+            // Mode Is New for QGC
+            _modeEnumToString[mode.custom_mode] = nModeName;
+        }
+        mode.mode_name = nModeName;
+        _availableFlightModeList += mode;
     }
-    else{
-        // Mode Is New for QGC
-        _modeEnumToString[mode.custom_mode] = nModeName;
-    }
-    mode.mode_name = nModeName;
-    _availableFlightModeList += mode;
 }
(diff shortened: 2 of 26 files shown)


DonLakeFlyer:
Working my way through this. Found some raw flight mode strings here: https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L1267. Not sure what that is used for.

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
```diff
@@ -789,3 +742,35 @@ const QVariantList& PX4FirmwarePlugin::toolIndicators(const Vehicle* vehicle)
...
     return _toolIndicatorList;
 }
+
+void PX4FirmwarePlugin::updateAvailableFlightModes(FlightModeList modeList)
+{
+    _availableFlightModeList.clear();
+    for(auto mode: modeList){
+
+        // Update Multi Rotor
+        switch (mode.custom_mode) {
```
DonLakeFlyer, line 752: Both of these switches would be better without a default. By listing out every mode if/when a new mode gets added the compiler will complain if these switches are not updated. With the default, a new mode could be added and if updating these switches are forgotten they will fall to default and possible do the wrong thing.

src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -567,3 +567,32 @@ Autotune* FirmwarePlugin::createAutotune(Vehicle *vehicle)
...
+        _updateModeMappings(mode);
+    }
+}
+
+void FirmwarePlugin::_setModeEnumToModeStringMapping(FlightModeCustomModeMap enumToString)
+{
+    _modeEnumToString = enumToString;
+}
+
+void FirmwarePlugin::_updateModeMappings(FirmwareFlightMode &mode){
```
DonLakeFlyer, line 585: I would make this method take in a modeList not a single mode. The reason is that this method expects the list to be empty when it is called. Both those clear calls are in multiple places in other parts of the codebase outside of this method. It would be super easy to forget to call clear before calling this method. By making the method take a list you can clear the list as needed directly in the _updateModeMappings method and outside callers don't need to know to do that themselves.

src/FirmwarePlugin/FirmwarePlugin.cc
```diff
@@ -567,3 +567,32 @@ Autotune* FirmwarePlugin::createAutotune(Vehicle *vehicle)
...
+    // Get Presaved Mode String, if mode is not present then take Custom Mode Name
+
+    QString nModeName = mode.mode_name;
```
DonLakeFlyer, line 588: I wonder if this should be the other way around. Naming coming from standard modes takes precedence. That naming is likely what is going to be in the docs for the firmware. If we use the internal QGC naming that make differ.

The benefit of using the internal QGC naming is that it will end up being translated.

Not quite sure what is good here. Did you have specific reason to do it this way?

DonLakeFlyer:
Shouldn't this method: https://github.com/mavlink/qgroundcontrol/blob/master/src/FirmwarePlugin/FirmwarePlugin.cc#L54 be updated. A non-PX4/ArduPilot vehicle could still support the new standard modes protocol. In that case the real names should be available. Seems like it should check to see if the list is populated and use that if available, otherwise do it the old way.

DonLakeFlyer:
Seems odd that standard modes don't come across with names: https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/StandardModes.cc#L47. @bkueng That said shouldn't these be translated?

DonLakeFlyer:
@bkueng This is interesting as well: https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/StandardModes.cc#L75. Not sure it makes sense for these to not be available for usage from the flight modes menu. They don't cause a problem if used from there do they?

DonLakeFlyer:
@bytesByHarsh If you search for `.flightMode` in `.qml` files you'll find a couple places which are still using raw flight mode strings. Those need to be cleaned up as well.

DonLakeFlyer (approved):
Mainly looks great. Just a couple questions.

src/FirmwarePlugin/FirmwarePlugin.cc
bytesByHarsh: So if we take the flight mode name from the standard modes then the issue is first of casing, ie arducopter sending all caps. Even if I add `Qt:CaseInsensitive` in the string comparison, in the case of ArduRover instead of GUIDED it is sending GUID, basically just 4 characters of the flight mode. So to avoid confusion in modes I decided to take them from internal mode strings.

src/FirmwarePlugin/FirmwarePlugin.cc
bytesByHarsh: Sure will update this

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
bytesByHarsh: So in the default case should we give any pop up message to UI?

src/FirmwarePlugin/FirmwarePlugin.cc
DonLakeFlyer: > in the case of ArduRover instead of GUIDED it is sending GUID

I would file a bug on that.

> So to avoid confusion in modes I decided to take them from internal mode strings.

Ok, let's go with that way then and see if it causes any confusion.

src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
DonLakeFlyer: There shouldn't be a default case at all. The compiler won't allow builds with missing cases.

CMakeLists.txt
```diff
@@ -553,4 +553,12 @@ elseif(MACOS)
     install(SCRIPT "${CMAKE_SOURCE_DIR}/cmake/CreateMacDMG.cmake")
 endif()
 
+## Enforce Handling of All Enum Cases
+
+if(WIN32)
+    add_compile_options(/w44265 /we44265)
+else()
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wswitch -Werror=switch")
```
DonLakeFlyer, line 561: @HTRamsey This won't be needed with the switch to `-Wall` right?

bkueng:
> Seems odd that standard modes don't come across with names: https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/StandardModes.cc#L47. @bkueng That said shouldn't these be translated?

I think I just did it that way to ensure consistent naming. We can also change it.
EDIT: it comes from the mavlink spec: ~https://mavlink.io/en/messages/development.html#AVAILABLE_MODES~ https://mavlink.io/en/messages/common.html#AVAILABLE_MODES
Right, I forgot about the translation.

> @bkueng This is interesting as well: https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/StandardModes.cc#L75. Not sure it makes sense for these to not be available for usage from the flight modes menu. They don't cause a problem if used from there do they?

I did it to preserve the previous behavior. There's no harm in adding them, but e.g. orbit requires a location/radius for activation.

PR #12263, opened 2024-12-30
GuidedValueSlier: Click on value to type into text field

DonLakeFlyer:
https://github.com/user-attachments/assets/8a90ecf8-74d0-4f68-893f-c6f2e7048fc4

### DIFF
--- a/src/FlightDisplay/GuidedValueSlider.qml
+++ b/src/FlightDisplay/GuidedValueSlider.qml
@@ -77,16 +77,26 @@ Item {
 
     property var _qgcPal: QGroundControl.globalPalette
 
+    function setCurrentValue(currentValue, animate = true) {
+        // Position the slider such that the indicator is pointing to the current value
+        var contentY = (_firstPixelValue - currentValue) / _sliderValuePerPixel - _indicatorCenterPos
+        if (animate) {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@DieBorr Can you look through this?

DieBorr:
Ok, I will check it right now.

DieBorr:
Okay, the textField label itself is very well integrated, however I can see a glitch while moving the slider with or without the textField enabled, here is a video:

https://github.com/user-attachments/assets/1b297eec-4784-4b94-9179-0d010863ef73

The problem is in :
`return Math.min(Math.max(value.toFixed(decimalPlaces), _sliderMinVal), _sliderMaxVal).toFixed(decimalPlaces)` 
at _clampedSliderValue function, math.min returns an int when you send a float number with non significant decimal numbers so I suggest to replace that line with:

`Math.min(Math.max(value.toFixed(decimalPlaces), _sliderMinVal), _sliderMaxVal).toFixed(decimalPlaces)`

I could make a pull request but with just this little fix it does not make sense
Thank you !

DieBorr:
Finally I have made a pull request because there was some more things that I think that will improve the user experience

DonLakeFlyer:
Yup, that is screwed up. Fixed.

PR #12264, opened 2024-12-30
Rework for sliderValue TextField

DieBorr:
Now there is a validator that can only accept numbers and a maximum of 1 decimal. Also if the textField is visible while there is a sliderValue change it will hide. @DonLakeFlyer  feel free to take whatever you want from the commit !

Related with #12263

### DIFF
--- a/src/FlightDisplay/GuidedValueSlider.qml
+++ b/src/FlightDisplay/GuidedValueSlider.qml
@@ -77,16 +77,26 @@ Item {
 
     property var _qgcPal: QGroundControl.globalPalette
 
+    function setCurrentValue(currentValue, animate = true) {
+        // Position the slider such that the indicator is pointing to the current value
+        var contentY = (_firstPixelValue - currentValue) / _sliderValuePerPixel - _indicatorCenterPos
+        if (animate) {
+            flickableAnimation.from = sliderFlickable.contentY
+            flickableAnimation.to = contentY
+            flickableAnimation.start()
+        } else {
+            sliderFlickable.contentY = contentY
+        }
+    }
+
     /// Slider values should be in converted app units.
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Yeah, I've been debating using those validators for a while now. I"m kind of 50/50. It's a bit mysterious as to why you can't type things. That said I actually took a look at implementing product wide in all fields a few days back. It's a bit tricky. At this point if we are going to use them I want to use them everywhere, not just one place which I think will be more confusing. I'll probably get back to trying to get them to work again at some point.

DieBorr:
You are completely right, I didn't think about it until you have said, the truth is that a user that is not used to work with UAV's at all can find this completely tricky especially with the points and commas, at least in Spain most people is used to write decimal number with commas and this behavior can cause a frustration when writing it and not seeing any alert message. We could also just do a rewrite of a comma to a dot or even just parse it at the moment we send the value to the vehicle. However I find the validator of just number completely intuitive, why would someone send a letter instead of a number for height or speed values?

PD: What do you think about the hide logic that I included in the commit too? I found it intuitive but share me your thoughts !

DonLakeFlyer:
> Also if the textField is visible while there is a sliderValue change it will hide

Thanks. I'll bring this across to my pull.

src/FlightDisplay/GuidedValueSlider.qml
```diff
@@ -239,6 +261,41 @@ Item {
...
+                focus = false
+                setCurrentValue(_clampedSliderValue(parseFloat(text)))
+            }
+
+            Connections {
+                target: control
+                on_sliderValueChanged: sliderValueTextField.visible = false
+            }
+
+            Connections {
```
DonLakeFlyer, line 294: I didn't plan to pull this one across since it didn't seem to be needed. I think hiding on slide is better.

PR #12270, opened 2025-01-01
CMake: Turn On Warnings as Errors

HTRamsey:
Best way to do this for now until the individual QGC source libraries are removed. Then we can just set this per target or something


DonLakeFlyer:
Can we turn on '-Wall' as well?

HTRamsey:
Yup, I'll add that to this (and fix windows redefinition warning)

PR #12274, opened 2025-01-03
Added horizontal value slider support

DonLakeFlyer:
This control works exactly like the vertical guided value slider you see during things like a Takeoff action. You can flick it left/right to change the value. And you can click on the value to type in a new value. The goal is to make QGC more mobile friendly, where typing in values can be a real pain with the virtual keyboard. That said it still supports clicking in the value and typing in a value directly if you want to. So hopefully the best of both worlds.

I've converted previous UI which was using the old FactSlider control to use the new control. Here is an example with the battery toolbar indicator:

New
Screenshot 2025-01-02 at 4 42 26 PM

Old
Screenshot 2025-01-02 at 5 29 12 PM


DonLakeFlyer:
@DieBorr This new control is basically a rewritten version of the vertical one. But I wrote it to be much closer to the code needed to make a single control which can do both vertical and horizontal orientations. Once this goes through I'm going to take a look at making that single control. Not quite sure if it will end up being so complex/obfuscated to be worth a single control. But I'm going to try.

DieBorr:
Hi @DonLakeFlyer sorry for not answering earlier.
Right now I have tested them, and they look so clean with the UI. This type of changes is what the average user values the most when deciding which applications to use. The fact is easier to change the values to more exact points while using the fingers makes it very useful.

P.S.: I'm planning to ask people who usually use other ground stations about which UI features they value the most when deciding to use a ground station. Based on their feedback, I'll try to rework those aspects within QGC. One example is the sliders, which now have a more 'solid/premium' style.

Thanks !

PR #12276, opened 2025-01-03
Update mavlink 

dakejahl:
Fixes PX4 broken mode lists from https://github.com/mavlink/mavlink/pull/2191

Do we still need the mavlink submodule?


DonLakeFlyer:
@HTRamsey How come the mavlink submodule is still here? We don't use that anymore do we?

HTRamsey:
Just answered via discord, but basically it and several other submodules were left from when we were deciding git submodules vs cmake dependency handling

PR #12277, opened 2025-01-03
remove FTDI from USB pid/vid list for SiK Radio

dakejahl:
Useful to debug an MCU while also using QGC. I don't think we should identify FTDI devices as SiK Radios and attempt to autoconnect as such.

### DIFF
--- a/src/Comms/USBBoardInfo.json
+++ b/src/Comms/USBBoardInfo.json
@@ -54,7 +54,6 @@
         { "vendorID": 9900, "productID": 4132,      "boardClass": "Pixhawk",    "name": "mRo Control Zero H7 OEM" },
 
         { "vendorID": 1027, "productID": 24597,     "boardClass": "SiK Radio",  "name": "SiK Radio",            "comment": "3DR Radio" },
-        { "vendorID": 1027, "productID": 24577,     "boardClass": "SiK Radio",  "name": "SiK Radio",            "comment": "3DR Radio on FTDI" },
         { "vendorID": 4292, "productID": 60000,     "boardClass": "SiK Radio",  "name": "SiK Radio",            "comment": "SILabs Radio" },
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
The question is what radio setup is this exactly. Both 3DR radio entires are identified as 1027 which is FTDI manufacturer. It's all kind of ancient history. Is it really some obscure setup which is rarely used?

dakejahl:
The Holybro SiK radios I have here identify as 0403:6015 which corresponds to that other 3DR Radio entry.
```
0403:6015 Future Technology Devices International, Ltd Bridge(I2C/SPI/UART/FIFO)
```
```
{ "vendorID": 1027, "productID": 24597,     "boardClass": "SiK Radio",  "name": "SiK Radio",            "comment": "3DR Radio" },
```

Whereas the one I removed corresponds to the basic USB UART converter. One other thing I noticed a while back is that if you have SiK Radio autoconnect enabled you can't create a mavlink connection using the Comm Links page on that interface since it is trying to autoconnect to it as a SiK Radio which uses a default 57600 baudrate. If your link (let's say a pixhawk telem port) isn't at 57600 it won't autoconnect and you can't connect with the Comm Links, you have to know to go and untick the sik radio autoconnect box which is default enabled.

Anyway, I guess what I'm getting at it can be a little confusing depending on what you're trying to do if you're not aware of this. IMO if someone has a SiK radio that identifies as an generic USB UART converter I think they should just manually create a comm link.

DonLakeFlyer:
Yeah, I think I can live with that.

DonLakeFlyer:
> One other thing I noticed a while back is that if you have SiK Radio autoconnect enabled you can't create a mavlink connection using the Comm Links page on that interface since it is trying to autoconnect to it as a SiK Radio which uses a default 57600 baudrate.

I think this could possibly be fixed at least to the point of letting the user know that they need to turn off autoconnect. You should be able to detect trying to connect to something which is already auto-connected to. Shouldn't be too hard...

PR #12259, opened 2024-12-28
GuidedActions: fix incorrect slider value conversion

or-rhssk:
Description
-----------
The floating point value got converted to a string, which in turn broke arithmetic in GuidedActionsController (e.g. `Vehicle::guidedModeOrbit` always received NAN as altitude).

Checklist:
----------
- [x] [Review Contribution Guidelines](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CONTRIBUTING.md).
- [x] [Review Code of Conduct](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CODE_OF_CONDUCT.md).
- [x] I have tested my changes.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #12278, opened 2025-01-03
Orbit fixes

DonLakeFlyer:
* Based on / replacement for #12259
* Make guided slider return floats not strings
* Fix orbit command to send altitude in AMSL frame


DonLakeFlyer:
@or-rhssk Could you help test this once the artifacts are done building?

PR #12280, opened 2025-01-04
Logging fixes

DonLakeFlyer:
Problems fixed:
* Joystick code spitting out pile of this `qgc.joystick.joysticksdl: Couldn't add GameController mapping: Couldn't parse GUID from # Game Controller DB for SDL in 2.0.16 format`
* Joystick logging getting output even when not turned on
* PITA to turn off logging categories which are turned on and/or see which ones are on. Have to scroll through a big list

New logging dialog:
Screenshot 2025-01-04 at 9 26 56 AM

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -23,7 +23,7 @@
 #include <QtCore/QSettings>
 #include <QtCore/QThread>
 
-QGC_LOGGING_CATEGORY(JoystickLog, "Joystick")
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
@HTRamsey I'm pretty sure the gamecontroller reading stuff is right. But you should check.

HTRamsey:
I think there's an SDL command to read a whole file instead of line by line

PR #12281, opened 2025-01-04
Mac dependency script

DonLakeFlyer:


HTRamsey:
should we call this from the workflow macos.yml?

DonLakeFlyer:
I guess, it depends. Which path is more important to verify in CI? With or without pre-installed libs?

DonLakeFlyer:
I guess the docs say to install the libs so that's the would be the one to replicate in CI.

HTRamsey:
I think you'll still want the `continue-on-error: true` because it'll pop an error if one of the downloads fails or something, at which case we can rely on cmake to just build the dependency

DonLakeFlyer:
I specifically left that out so that we will know if availability of something from brew changes. When that happens we will need to update the dependencies script. That seems helpful. But I get the download thing. If that seems to crap out a lot abd become and annoyance then we could add it back?

HTRamsey:
I think the task still actually shows as failing it just continues anyways, so it should still be obvious that something needs to be done but at least you'll know the build is okay

DonLakeFlyer:
> I think the task still actually shows as failing it just continues anyways

I just tried it and unless you specifically open up the depedencies task you'd have no idea it failed. I'm gone it leave it as failing if brew screws up.

PR #12285, opened 2025-01-05
Use MAV_CMD_DO_SET_MISSION_CURRENT for APMFirmwarePlugin

peterbarker:
Send the mavlink command MAV_CMD_DO_SET_MISSION_CURRENT in place of the old mavlink message MISSION_SET_CURRENT.  If "UNSUPPORTED" is received from the autopilot then send the old mavlink message.

The mavlink command has existed since 2019.

The mavlink message has been deprecated since Aug 2022.

ArduPilot has supported the command since 4.1.0 (2021)

Adds infrastructure which should be reused for (eg.) the DO_REPOSITION code.

### DIFF
--- a/src/Vehicle/Vehicle.h
+++ b/src/Vehicle/Vehicle.h
@@ -994,6 +994,9 @@ private:
 
     static void _rebootCommandResultHandler(void* resultHandlerData, int compId, const mavlink_command_ack_t& ack, MavCmdResultFailureCode_t failureCode);
 
+    // send mavlink command (deprecated since Aug 2022)
+    Q_INVOKABLE void setCurrentMissionSequenceWithMissionSetCurrent(int seq);
+
     int     _id;                    ///< Mavlink system id
     int     _defaultComponentId;
     bool    _offlineEditingVehicle = false; ///< true: This Vehicle is a "disconnected" vehicle for ui use while offline editing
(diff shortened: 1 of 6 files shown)


src/Vehicle/Vehicle.h
```diff
@@ -986,14 +994,17 @@ private slots:
 
     static void _rebootCommandResultHandler(void* resultHandlerData, int compId, const mavlink_command_ack_t& ack, MavCmdResultFailureCode_t failureCode);
 
+    // send mavlink command (deprecated since Aug 2022)
+    Q_INVOKABLE void setCurrentMissionSequenceWithMissionSetCurrent(int seq);
```
DonLakeFlyer, line 998: I would make this non-invokable and private. I don't think we want anyone using this except for the lambda fallback case. I might put lambdaFallback in the name to steer people away from using it in other Vehicle code.

DonLakeFlyer: Does that make sense?

peterbarker: I've pushed up a patch on top which simply inlines the function into the lambda itself.  I was avoiding that to reduce the patch size - but it does neaten things up a bit.

Thoughts?

DonLakeFlyer: Yeah, thanks Peter. I like that much better.

PR #12295, opened 2025-01-07
Disabled video support on MAC - plus other fixes

DonLakeFlyer:
* Temporarily disable video support on MAC till we can figure out how to make it work
* Fix crash in chunked status text handler
* Fix incorrect call to old popup code


HTRamsey:
Fixing the cmake configure now. Do you know why the keyValueRange thing fails in statusTextHandler?

DonLakeFlyer:
> Do you know why the keyValueRange thing fails in statusTextHandler?

Nope. Debugged it and it crashing deep in the Qt iterator code.

HTRamsey (approved):

PR #12298, opened 2025-01-07
[Backport v4.4] Added board ids

julianoes:
This copies the board IDs from the master branch.


julianoes:
@DonLakeFlyer and macOS CI fails again... see what I mean.

DonLakeFlyer:
Yeah, I've fought with that for ages. Seems way worse now than it used to be though. It's one of those "wave around some chicken blood" and it goes away kind of things.

PR #12307, opened 2025-01-08
Fix flight modes being announced multiple times, plus others

DonLakeFlyer:
Commit text says it all...

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -283,7 +283,6 @@ void Vehicle::_commonInit()
     _vehicleLinkManager             = new VehicleLinkManager            (this);
 
     connect(_standardModes, &StandardModes::modesUpdated, this, &Vehicle::flightModesChanged);
-    connect(_standardModes, &StandardModes::modesUpdated, this, [this](){ Vehicle::flightModeChanged(flightMode()); });
 
     _parameterManager = new ParameterManager(this);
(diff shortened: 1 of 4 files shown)


src/Vehicle/Vehicle.cc
```diff
@@ -283,7 +283,6 @@ void Vehicle::_commonInit()
     _vehicleLinkManager             = new VehicleLinkManager            (this);
 
     connect(_standardModes, &StandardModes::modesUpdated, this, &Vehicle::flightModesChanged);
-    connect(_standardModes, &StandardModes::modesUpdated, this, [this](){ Vehicle::flightModeChanged(flightMode()); });
```
DonLakeFlyer, line 286: @bytesByHarsh This was causing the current flight mode when this happens to be announced a second time. I think it's ok to remove this since at worse there could have been a name change on the flight mode. The actual flight mode would not have changed. So the user would have been previously notified of the current flight mode. I don't think there is a need to announce if just the name changed which would be odd anyway.

HTRamsey:
I guess this fixes that reset model warning?

DonLakeFlyer:
Right

HTRamsey (approved):

src/Vehicle/Vehicle.cc
bytesByHarsh: Yes true, the flight mode change we have shifted to the firmware plugin class so it should be ok to remove from here.

bytesByHarsh (approved):

PR #12311, opened 2025-01-09
Add Unique Flight Modes Only

bytesByHarsh:
Resolved multiple flight modes in ArduPlane: https://github.com/mavlink/qgroundcontrol/pull/12254#issuecomment-2571379641

Arduplane is sending standard modes twice, thus adding them to the list multiple times.

Also changed from `final` to `override` for better integration in Custom GCS


hamishwillee:
If you are requesting once and they are sending any mode more than once, that is a bug in their implementation.

PR #12312, opened 2025-01-09
Motor Setup

Godeffroy:
Motor Setup
-----------

Change the motor setup page : 
Instead of a slider for each motor with the difficulty to know how many power you put in, i made a slider to choose the power with fine tuning, a button for each motor to test them, a button to start all motor and a button to stop them all at once.

-----------

Tested on an hexa copter with Ardupilot Copter 4.5

Checklist:
----------

Related Issue
-----------


Godeffroy:
I commited all translation files because of the text changes. I don't know if i need to remove the *.ts files ? or do something to synchronize the translation ?

I also change the *.md file for the documentation with a new picture.

Godeffroy:
> It all seems quite a bit more complicated than it was before. Is there a reason you need to know how much exact power you are putting?

Yes, i'm working with QGC on a tablet size screen and with a big finger, you can put too much power by mistake ! This is also veryusefull when you want to compare each motors rotation (detection of noise during rotation for example) : you need to have the same input power to be able to compare.

Godeffroy:
It is how it's done in Mission Planner, there even is the rotation time available to change the default 3s.

DonLakeFlyer:
Hmm, ok. Can you use the new ValueSlider control then. This allows for everything you want and more. Including typing in a specific value. Example usage from battery toolbar indicator:
Screenshot 2025-01-14 at 8 36 14 AM

Godeffroy:
The ValueSlider is so cool, thanks for the tip !

Here is the new visual : 
image

About the documentation, i'm not able to translate the english version into Chinese, Korean and Turkish (with docs\en\qgc-user-guide\setup_view\motors.md)

DonLakeFlyer:
> About the documentation, i'm not able to translate the english version into Chinese, Korean and Turkish (with docs\en\qgc-user-guide\setup_view\motors.md)

Oh, sorry. Forgot to talk about that. You should only modify the english version as needed. You don't need to do the other languages.

DonLakeFlyer:
Also I don't think you don't need the plus/minus buttons. That sort of thing isn't anywhere else in the UI for these types of sliders. If you want that much direct control of motor value then just click number indicator and type in a value.

Godeffroy:
Ok, thanks for the feedback, i'm on it 👍

Godeffroy:
Should i remove all the modification of *.ts files in the dir ..\translations\ ?

Godeffroy:
Looks like this now
image

DonLakeFlyer:
You should only update the english version and leave the rest alone. They will get updated through an automated process

DonLakeFlyer:
I didn't realize this not only a ui change but also changes the way the slider/motor selection works. Now the motor is only commanded to a speed when you press a button. Whereas before the slider would adjust the motor speed. There was also a timeout of 3 seconds which would turn off the motors as a safety feature. I'm not convinced the new mechanism is better with respect to having to press the button to get the motor to spin. What's the reasoning behind changing the user model behind how it work. I thought all you wanted was more tunability to the exact motor percentages? Want to understand why this is better.

A second thing is that I hadn't realized that there was a ArduPilot variant of the control as well as a common variant of this control. The code for both is 99% exactly the same. Do you think you can help out and clean this up as well? The only different between the two is motor numbers version motor letters. Could you create a base class which holds all the guts of the control and calls up to a motorIndexToString function.  Then two classes CommonMotorComponent and APMMotoComponent which use CommonMotorComponent as base class and  provide that function.

DonLakeFlyer:
I'm thinking more along the lines that the motor buttons which be like checkboxes. Pick which motors you want to test then slide the slider to crank them up.

DonLakeFlyer:
@bkueng Can you comment on this? Wondering what you think about the click the button to make the motor go model, versus the slider the slider way it used to work. Would be good to make this consistent with the Actuators page as well. I'm assuming you did a pile of real world testing with motors/actuators page so can comment on what you think is the better way.

Godeffroy:
> A second thing is that I hadn't realized that there was a ArduPilot variant of the control as well as a common variant of this control. The code for both is 99% exactly the same. Do you think you can help out and clean this up as well? The only different between the two is motor numbers version motor letters. Could you create a base class which holds all the guts of the control and calls up to a motorIndexToString function. Then two classes CommonMotorComponent and APMMotoComponent which use CommonMotorComponent as base class and provide that function.

It is possible. It's not a big thing 👍

Godeffroy:
> I'm thinking more along the lines that the motor buttons which be like checkboxes. Pick which motors you want to test then slide the slider to crank them up.

That won't do : we can always slip by mistake on the engine power.

Godeffroy:
> A second thing is that I hadn't realized that there was a ArduPilot variant of the control as well as a common variant of this control. The code for both is 99% exactly the same. Do you think you can help out and clean this up as well? The only different between the two is motor numbers version motor letters. Could you create a base class which holds all the guts of the control and calls up to a motorIndexToString function. Then two classes CommonMotorComponent and APMMotoComponent which use CommonMotorComponent as base class and provide that function.

Wow, i was not aware that Ardupilot use letters for motors instead of numbers like PX4 !
It could be good indeed to clean up a little bit.

I can try.
If you know another component which do this, show me.

DonLakeFlyer:
> If you know another component which do this, show me.

Don't worry about it. I'll clean it up. I'll merge this in now.

PR #12313, opened 2025-01-09
Add support for multiple landing sequences

rubenp02:
# Add support for multiple landing sequences

Description
-----------
### [Change plan drawing in Fly View to match Plan View](https://github.com/mavlink/qgroundcontrol/commit/e2d7ca727a947b39993808301930b0033f5fedc8):

Updated the Fly View's flight plan drawing to match the Plan View, enabling support for discontinuities in missions (unconnected flight legs) in the former.

Key changes:
- Fly View now uses simpleFlightPathSegments, the same model used by Plan View, to draw flight paths. This enables proper handling of disconnected mission legs, which the previous _waypointPath approach did not support.
- Flight plan arrows now appear in the same places between the Fly and the Plan Views.
- Removed the old Fly View drawing logic and related members and properties.

### [Add support for multiple landing sequences](https://github.com/mavlink/qgroundcontrol/commit/2647246fa536963545029e910a1f9eb039d0f157):

Added an option to configure multiple landing sequences in a mission. This feature allows users to define alternative landing sites, which can be useful for contingency planning. When multiple landing sequences are present, the first one is used for the primary mission, but in the event of an RTL, the closest landing sequence will be used, as outlined in the MAVLink spec: https://mavlink.io/en/messages/common.html#MAV_CMD_DO_LAND_START.

Key changes:
- Disabled the restriction to only one landing sequence.
- Modified LandingComplexItem to support scanning for multiple complex items, and configuring the DO_LAND_START mission items to contain the coordinates of the first element in the landing sequence.
- Updated MissionController to handle multiple landing sequences, including proper flight path segment calculation and handling of discontinuities after landing commands.
- The TerrainProfile displays the entire mission with any possible alternative landing sequences, but the mission stats (distance and time) represent the primary mission, up to the default landing.
- Added an option in the Plan View settings to control this feature, defaulting to match the previous behavior.

### [Add support to set Land Start coords to next item](https://github.com/mavlink/qgroundcontrol/commit/ce1f246172f83ee37c1318274cdcee3ae53dfdfa):

Added an option in the Plan View settings to automatically set the coordinates of DO_LAND_START mission items to match the next waypoint in the mission. This is used to help find the closest landing sequence in the event of an RTL. This option is disabled by default to match the previous behavior.

Checklist:
----------


DonLakeFlyer:
> Added an option in the Plan View settings to automatically set the coordinates of DO_LAND_START mission items to match the next waypoint in the mission. This is used to help find the closest landing sequence in the event of an RTL. This option is disabled by default to match the previous behavior.

So it's the lat/lon in the DO_LAND_START which is used to find the nearest landing pattern sequence for an RTL if there are multiple patterns. Is that correct? If that's the case, is there a reason why setting this with lat/lon/alt should not always be done?

> Added an option to configure multiple landing sequences in a mission.

I can see how maybe a commercial vehicle creator may want to create a custom version of QGC where this is not allowed. But for general use I wonder if you should be able to turn off this behavior. Any reason why this shouldn't be allowed by default? Maybe a popup, with a "Don't show again" type thing which tells the user what will happen if they add a second pattern. Then they can Yes/No it if they did it by accident?

DonLakeFlyer:
> Added an option in the Plan View settings to automatically set the coordinates of DO_LAND_START mission items to match the next waypoint in the mission. This is used to help find the closest landing sequence in the event of an RTL. This option is disabled by default to match the previous behavior.

I wonder if an option is needed for this as well? I would change the handling of this to tweak DO_LAND_START in the scan code for the complex item on the way in. This way it only takes effect for landing patterns which were created using the Plan/Pattern support. It specifically does not touch DO_LAND_STARTs which may have been manually added to by the user in which case they control what they want themselves.

DonLakeFlyer:
> At least in ArduPilot, its not necessary. The DO_LAND_START seems to get picked based on the coords of the next item, at least when no DO_LAND_START items specify any coords. However the MAVLink spec seems to suggest having it, so I assume other firmwares might need it, which is why I'm adding it as an option.

Since PX4 doesn't support multiple landing points it doesn't look at the coordinate in here. To me the mavlink spec is bad. The spec should say that closest landing points are determined by the first coordinate item found after the DO_LAND_START. Otherwise the work a GCS has to go through to keep this up to date is crazy. If you create a manual landing path and then move the item after the DO_LAND_START then you need to go backwards and update the coordinate in the DO_LAND_START to keep things working. Makes no sense.

My suggestion is to remove the DO_LAND_START futzing code. And then I'll get a discussion started on changing the spec to be something useful.

DonLakeFlyer:
I've just been testing this with PX4 SITL and it fails upload of fixed wing landing sequences. Not sure why yet. I verified that Stable still works fine.

DonLakeFlyer:
I also did the following:
* Use PX4 SITL and the cessna for testing
* Use Stable to upload a simple mission with a landing sequence to the vehicle
* Quit QGC and start the version from this pull
* Go to Plan. The landing sequence degenerates into individual items instead of being sucked back into a landing pattern complex item

@rubenp02 Can you test this with PX4 SITL and figure out what is going on? I'm guess this sequence I've detailed here in this comment would break with ArduPilot as well.

DonLakeFlyer:
Might be easier to just flash a board with PX4 than setup SITL: https://docs.px4.io/main/en/dev_setup/dev_env.html.

> PX4 doesn’t support multiple landings?

I don't think so. Hoping to get some verification in that issue I created for DO_LAND_START

DonLakeFlyer:
Sure this will be in 5.0. 5.0 will be whatever daily is in a couple months.

src/FirmwarePlugin/PX4/PX4-MavCmdInfoCommon.json
```diff
@@ -14,6 +14,12 @@
             "comment":      "MAV_CMD_NAV_LAND",
             "paramRemove":  "1"
         },
+        {
+            "id":           189,
+            "comment":      "MAV_CMD_NAV_WAYPOINT",
+            "specifiesCoordinate":  false,
+            "standaloneCoordinate": false
+        },
```
DonLakeFlyer, line 22: Don't understand what this is? This is wrong as far as I can see.

DonLakeFlyer (changes requested):

DonLakeFlyer:
Also the discussion going on in the other issue about coords in DO_LAND_START is still clear as mud to me. Need to wait until that settles out as well.

src/FirmwarePlugin/PX4/PX4-MavCmdInfoCommon.json
DonLakeFlyer: Ah, ok. So just comment is wrong.

DonLakeFlyer:
> Also the discussion going on in the other issue about coords in DO_LAND_START is still clear as mud to me. Need to wait until that settles out as well.

I can live with this the code stands now so we can merge this in after the json fix. Then depending on what happens on the conclusion of that discussion in the other issue can you then come back and remove the coords in DO_LAND_START after the fact if that is where that discussion ends up. That way we don't wait on that to get this in.

peterbarker:
On Sun, 2 Mar 2025, Don Gagne wrote:

>       Have you tested this after the changes and does it look good to you now?
> 
> I'm trying to test in PX4 SITL but the current SITL plane won't land because it doesn't have lidar. Trying to get a version that will work for me. As it stands now I can't test with
> ArduPilot SITL. I can't figure out how to get SITL running in a VM to communicate over UDP to QGC on my main OS.

Adding a udp output is "output add 17.18.19.20:9876" for a UDP stream.. 
But with a more likely IP address.  There obviously has to be a route from 
the VM to the main OS for that to work.

Happy to do a VN / screen share to help you out with that.

DonLakeFlyer:
@peterbarker Yeah I've done the `output add` thing many times before. What happens is that the QGC to ArduPilot connection seems to only pass traffic in the ArduPilot->QGC direction. QGC connects, but then when QGC asks for things like the parameter list it gets no response back and times out.

peterbarker:
On Tue, 4 Mar 2025, Don Gagne wrote:

> @peterbarker Yeah I've done the output add thing many times before. What happens is that the QGC to ArduPilot connection seems to only pass traffic in the ArduPilot->QGC direction. QGC
> connects, but then when QGC asks for things like the parameter list it gets no response back and times out.

Probably worth running tcpdump on the VM to see if ArduPilot is sending 
back the relevant response - and which address it is sending to.

DonLakeFlyer:
I don't remember quite how Terrain Profile works. But here is an example with an alternate landing:
Screenshot 2025-03-05 at 8 29 13 AM

Could you see if it's possible to make the second landing pattern say Alternate FW Landing or something?

As it stands now I'm good to merge this. If you want I can merge and you can look at this in a separate pull?

DonLakeFlyer (approved):

PR #12330, opened 2025-01-16
[V4.4] CI: try to fix macos paths one more time

julianoes:


HTRamsey:
Can you explain the need for this?

julianoes:
macos-12 runners were deprecated, so I had to switch to macos-13 which broke. So, I had to figure out why it broke. Since I don't have a Mac, don't really understand the problem, I threw everything at claud.ai to fix it. I hope this time I got it.

The need for it is that I might publish V4.4.5 with a [board ID backport](https://github.com/mavlink/qgroundcontrol/pull/12298), and potential other fixes if any come up before I hit release.

Makes sense? I'm assuming V5 is still a bit out, or is it about to be dropped?

julianoes:
Not sure why it fails at "generating the rebase commit". I assume that's a github issue? :shrug:

HTRamsey:
I was asking because I was curious if I had to do it for master too. I'm having trouble with the macos GStreamer stuff in master too.

HTRamsey:
In QGC 4.X CMake build uses prepare_gstreamer_framework.sh but qmake doesn't. Not really sure why the difference

julianoes:
Every time I re-commit this and force push it, it seems to fail again. Then I fix it and next time it fails again.

julianoes:
Well, I might actually have to think this through since throwing it at Claude.ai doesn't seem to solve it for me.

HTRamsey:
https://github.com/OpenKJ/OpenKJ/blob/f116291e50f1fa46acf05135e164abd81d013ee0/CMakeLists.txt#L671

HTRamsey:
It's basically impossible if you don't have a mac to test on.

DonLakeFlyer:
@julianoes This is passing CI...

julianoes:
@DonLakeFlyer every second try... Let's see if I can amend, and force push to tr again.

julianoes:
Actually, it looks like the branch Stable_V4.4 is passing anyway, so I can close this.

julianoes:
Let's try again.

PR #12339, opened 2025-01-18
Add "Request Gimbal Control" popup

### LINKED ISSUE #12328: Missing AcquireGimbalControlPopup
In this PR (https://github.com/mavlink/qgroundcontrol/pull/11264), there was some logic added where certain gimbal actions will check if the Gimbal is acquired by someone else before commanding the gimbal action. If the gimbal is currently acquired by someone else, a `showAcquireGimbalControlPopup` signal gets emited. A comment says this is expected to be handled by QML. The problem is there is no such QML handling the signal

@Davidsastresas looks like you added that code. I assume you have a fork with the QML handler in it. Can you bring that to upstream?

gillamkid:
fixes #12328 

If another user was in control of the gimbal, the current logic would stop gimbal commands from being sent out, but not show a "Request Gimbal Control" popup. There was a signal being emitted suggesting  a popup was intended to be shown, but wasn't. Now a popup is indeed shown.

Screenshot 2025-01-17 200010

## Sponsor

### DIFF
--- a/src/QmlControls/QGCSimpleMessageDialog.qml
+++ b/src/QmlControls/QGCSimpleMessageDialog.qml
@@ -16,10 +16,17 @@ import QGroundControl.ScreenTools
 QGCPopupDialog {
     property alias  text:           label.text
     property var    acceptFunction: null        // Mainly used by MainRootWindow.showMessage to specify accept function in call
+    property var    closeFunction:  null
 
     onAccepted: {
         if (acceptFunction) {
             acceptFunction()
+        }
+    }
+
+    onClosed: {
+        if (closeFunction) {
+            closeFunction()
         }
     }
 
--- a/src/UI/MainRootWindow.qml
+++ b/src/UI/MainRootWindow.qml
@@ -160,8 +160,8 @@ ApplicationWindow {
     //-------------------------------------------------------------------------
     //-- Global simple message dialog
 
-    function showMessageDialog(dialogTitle, dialogText, buttons = Dialog.Ok, acceptFunction = null) {
-        simpleMessageDialogComponent.createObject(mainWindow, { title: dialogTitle, text: dialogText, buttons: buttons, acceptFunction: acceptFunction }).open()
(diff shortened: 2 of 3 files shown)


HTRamsey:
Ignore the macos failure, working on codesign right now

Davidsastresas:
Thanks for the PR.

I think this is the first time we are using this way of declaring UI from C++. I am not opposed to it, it is cool, but on this particular situation I don't think it makes sense, because we have pretty much the same functionality already in showMessageDialog() from mainrootwindow.qml. I would approach it this way:

- Declare that qml code for that component in GimbalIndicator.qml instead, in an standalone Component
- On the same GimbalIndicator.qml, connect the signal from GimbalController to call Mainwindow to create this new item when the signal is emitted from C++.

I think that way everything would be better placed, easier to figure out for other developers that might peek on those files, and easier to maintain.

Thanks!

gillamkid:
> Thanks for the PR.
> 
> I think this is the first time we are using this way of declaring UI from C++. I am not opposed to it, it is cool, but on this particular situation I don't think it makes sense, because we have pretty much the same functionality already in showMessageDialog() from mainrootwindow.qml, it uses the same QGCSimpleMessageDialog you used on your logic, so I think we could use that instead. I would approach it this way:
> 
> * Declare that qml code for that component in GimbalIndicator.qml instead of GimbalController cpp files
> * On the same GimbalIndicator.qml, connect the signal from GimbalController to call Mainwindow to create this new item when the signal is emitted from C++.
> * If you need any additional action, It would make more sense to expand it in showMessageDialog() rather than duplicating the functionality. I am thinking about your custom "onRejected" handling. But for the nature of your handling, we probably don't even need to expand showMessageDialog() at all I think, as onRejected is only closing the popup on this case.
> 
> I think that way everything would be better placed, easier to figure out for other developers that might peek on those files, and more compact and easier to maintain.
> 
> Thanks!

@Davidsastresas thanks for the well written feedback. I made the suggested changes :)

Davidsastresas (approved):

Davidsastresas:
@gillamkid That looks great now, thank you!

PR #12342, opened 2025-01-18
MAVLink Inspector, show all selected fields in the chart

mykhailopylyp:
Description
-----------
Currently, if I select `ATTITUDE.roll`, the data is displayed on the chart. However, if I switch to another message (e.g., `RAW_IMU`), the `ATTITUDE.roll` data stops rendering. It would be more intuitive if all selected fields (even from different messages) were shown on the chart simultaneously.

Screenshot from 2025-01-18 12-40-02

For example, I want to investigate the correlation between `ATTITUDE.roll` and `RAW_IMU.xacc`.

I investigated why this approach was implemented and couldn't find a specific reason preventing us from displaying multiple fields from different messages simultaneously.

Test Steps
-----------
- Make sure QGC is connected to a drone that is  in flight
- Go to MAVLink Inspector
- Select ATTITUDE.roll 
- Switch to another message (For example RAW_IMU)
- Observer that ATTITUDE.roll is still rendered on the chart

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/src/AnalyzeView/MAVLinkMessage.cc
+++ b/src/AnalyzeView/MAVLinkMessage.cc
@@ -108,7 +108,7 @@ void QGCMAVLinkMessage::update(const mavlink_message_t &message)
     _count++;
     _message = message;
 
-    if (_selected) {
+    if (_selected || _fieldSelected) {
         // Don't update field info unless selected to reduce perf hit of message processing
         _updateFields();
     }


src/AnalyzeView/MAVLinkMessage.cc
```diff
@@ -108,7 +108,7 @@ void QGCMAVLinkMessage::update(const mavlink_message_t &message)
     _count++;
     _message = message;
 
-    if (_selected) {
+    if (_selected || _fieldSelected) {
```
DonLakeFlyer, line 111: Hmm, interesting. Shouldn't this just be _fieldSelected? Since the message might be selected but that doesn't mean that a field is selected. Otherwise you end up doing a bunch of work in updateField every time you select a new top level message. Or am I reading the code wrong?

mykhailopylyp: You're right, `_selected` is not needed. It's not a big overhead though because there's only one message selected at the time. 

Do you want me to remove `_selected` from the if statement?

DonLakeFlyer: Yes, you should remove it. Make the code clearer to others looking at it later.

DonLakeFlyer: Any updates on this?

PR #12350, opened 2025-01-20
Rework the guided "Action" button

DonLakeFlyer:
* "Actions" button now shows a drop panel of additional action buttons instead of nasty previous ui
* Custom builds can add additional actions to the "Actions" panel by overriding FlyViewAdditionalCustomActionsList.qml
* Actions drop panel has three sections: Guided Actions, Custom Build Actions, Mavlink Actions
* Fixes a number of problems with Mavlink Actions

Example screenshot with first two standard Guided Actions, and second two user-defined Mavlink Actions:
Screenshot 2025-01-19 at 7 19 11 PM


DonLakeFlyer:
This is mostly a re-arranging of the existing deck chairs on qml files to allow for custom builds to take part.

PR #12352, opened 2025-01-20
Rework motor component into common base class

DonLakeFlyer:
@Godeffroy Can you test this on ArduPilot?


Godeffroy:
i can test it tomorrow yes.

Godeffroy:
This what i have for an Hexa copter With ArduCopter V4.6.0
image
image
image

But i have a problem with a fixed wing VTOL Quadrotor,  ArduPlane V4.5.1, i should have only 4 motors, but i have the default 8
image
image
Because there is no case for MAV_TYPE_FIXED_WING in the QGCMAVLink::motorCount function.

Godeffroy:
This what i have with PX4
image
image

PR #12353, opened 2025-01-20
OSX: Install GStreamer using dependency script

DonLakeFlyer:


HTRamsey:
There's also https://formulae.brew.sh/formula/gstreamer#default
Plus you should probably move the pkgconfig command into the script too right?

DonLakeFlyer:
@HTRamsey Do you see anything wrong here? My local builds are working fine on OSX.

DonLakeFlyer:
> Plus you should probably move the pkgconfig command into the script too right?

For me that isn't needed for it to build fine locally. I originally removed it. But ended up with framework linker error. So I put it back hoping it would maybe fix that which it didn't.

DonLakeFlyer:
If you install gstreamer from brew you don't get a framework you just get individual libs which creates a whole other can of worms.

PR #12359, opened 2025-01-21
Joystick:cc: Fix gimbal rate behaviour after commit e38fd4da1b4e54ec1…

Davidsastresas:
After the refactor done on this commit https://github.com/mavlink/qgroundcontrol/commit/e38fd4da1b4e54ec12ee233d2f4e86ef4c002fa9 we stopped sending the stop commands to gimbal control when releasing joystick buttons, so the functionality was totally broken.

For awareness @HTRamsey

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -1053,18 +1053,26 @@ void Joystick::_executeButtonAction(const QString &action, bool buttonDown)
     } else if (action == _buttonActionGimbalUp) {
         if (buttonDown) {
             emit gimbalPitchStart(1);
+        } else {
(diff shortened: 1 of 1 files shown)


AdamBucko1:
I have tested the implementation with a PS4 controller and a physical gimbal device, and I can confirm that the issue has been resolved. The gimbal now stops rotating as expected when a button of the respective direction is released.

HTRamsey:
Nice catch, sorry about that

Davidsastresas:
@HTRamsey no need to be sorry, you did a great and needed rework. I was just pinging you for awareness. Thanks!

Issue #12372, opened 2025-01-30
Daily build fails with "exiv2.dll was not found" error


rmackay9:
On my Windows11 machine, I uninstalled the stable QGC (which works fine) and re-installed the Daily build ([linked from this wiki page](https://docs.qgroundcontrol.com/master/en/qgc-user-guide/releases/daily_builds.html)) but sadly QGC fails to start properly.  It complains that, "exiv2.dll was not found"

![Image](https://github.com/user-attachments/assets/c47157f8-e681-4edc-9a4f-1cc9a49313cb)

hyeonseung89:
I'm experiencing the same problem. I reverted the commit ' fix window debug build '(c71b0d3) and it works. Can you confirm it?

rmackay9:
Hi there, thanks very much for fixing the above issue.  There's sadly another issue that is stopping the Daily build from running on Windows which I've added as a new issue here https://github.com/mavlink/qgroundcontrol/issues/12389

Happy to test any fixes that might help, sorry I'm not really too qualified to dig into it myself

HTRamsey:
@hyeonseung89 can you verify?

PR #12375, opened 2025-01-30
Add simple sanity check on generated release build

DonLakeFlyer:
Should prevent #12372 from sneaking through again


DonLakeFlyer:
This should be good to go once #12372 is fixed

PR #12376, opened 2025-01-30
Build Exiv2 as static lib

### LINKED ISSUE #12372: Daily build fails with "exiv2.dll was not found" error

DonLakeFlyer:
Fix #12372

### DIFF
--- a/src/AnalyzeView/CMakeLists.txt
+++ b/src/AnalyzeView/CMakeLists.txt
@@ -105,9 +105,7 @@ endif()
 set(EXIV2_VERSION 0.28.3)
 
 if(NOT QGC_BUILD_DEPENDENCIES)
-    if(MSVC)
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@HTRamsey I tried this change but it didn't work. Maybe you have some other ideas on how to use this as a static lib instead of dll.

HTRamsey:
It's because I changed it to Find instead of Add, it might find an installed lib, but it's not copying the DLL.

DonLakeFlyer:
I've verified this is working on windows now

rmackay9:
Thanks very much for this!

PR #12381, opened 2025-02-01
Fix Mac gstreamer support

DonLakeFlyer:
* Local builds now work
* Building a DMG you install from and run still doesn't work


s-lisovenko:
Can we statically link gstreamer for Mac OS? I think we can avoid a lot of linking problems, what do you think?

DonLakeFlyer:
I've tried using a brew installed gstreamer and I just ran into a larger pile of other problems. I don't think there are static libs available. Maybe. I'll look. Certainly would be easier if possible.

s-lisovenko:
> I've tried using a brew installed gstreamer and I just ran into a larger pile of other problems. I don't think there are static libs available. Maybe. I'll look. Certainly would be easier if possible.

I successfully linked it statically, but I used not brew version but gstreamer framework.

DonLakeFlyer:
> but I used not brew version but gstreamer framework.

Ah! I didn't of doing that?

s-lisovenko:
> > I successfully linked it statically
> 
> Were you able to play video through the build you made?

Yes, I tested it on local gst test pattern as Mac OS not a my daily driven OS.

PR #12384, opened 2025-02-02
Mac install: DMG creation and app inside DMG not work correctly

DonLakeFlyer:
* Turned off code signing in mac ci since it's not working yet
* Switched exiv2 on osx to static lic. Dynamic lib wasn't being copied into bundle causing runtime failure
* Ad-hoc codesign bundle after macdeployqt step
* Fix macdeployqt args. This was the main problem in that the plugins were not being copied to the bundle. And then when you ran QGC it just seemed to hang with no ui and no helpful qt output as to why. Took forever to figure out what was going on.


HTRamsey:
Looks good if it works. One of those 'verify bundle' commands from bundle utilities may be helpful, but if we run the executable test in CI I suppose it shouldn't really be necessary

HTRamsey (approved):

PR #12388, opened 2025-02-03
Always build exiv2 from source

DonLakeFlyer:


HTRamsey:
you still need EXIV2_VERSION

PR #12393, opened 2025-02-05
qt6: cmake: add export QT_SELECT=qt6

dakejahl:
If you have multiple versions of Qt installed cmake you need to set this, cmake will find Qt5 first.

### DIFF
--- a/docs/en/qgc-dev-guide/getting_started/index.md
+++ b/docs/en/qgc-dev-guide/getting_started/index.md
@@ -133,6 +133,11 @@ Example commands to build a default QGC and run it afterwards:
    cd qgroundcontrol
    ```
 
+1. If you have multiple versions of Qt installed, make sure you select Qt6
+   ```sh
+   export QT_SELECT=qt6
+   ```
+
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
But what version of Qt6 does that pick? For me I need `CMAKE_PREFIX_PATH` to point right to the one I want to use.

dakejahl:
so what's the best way to do this? Can we add the default path to cmake and error if it doesn't find the exact version we want? And a user could override with an env var if they want to specify a different version

DonLakeFlyer:
If you want to use cmake directly instead of say just using Qt Creator which will get you the right thing, then I think you are kinda on your own to set up an alias to qt-cmake using whatever OS mechanism you have available. Gonna be too long winded to explain something for every OS. But I'd hope if you want to go this route you kinda know what you are doing already and can handle that. Otherwise you should be using Qt Creator.

HTRamsey:
I think it will be better to just force this in cmake or get them to use the qt-cmake rather than rely on people to have to do this manually

PR #12399, opened 2025-02-05
param: debounce search with timer to reduce UI lag

### LINKED ISSUE #12394: Param search is laggy
The param search is pretty laggy, but only on the first few characters. I am thinking there is some kind of list traversal happening across the param list on every keystroke. It's not a huge deal but would be nice to fix. I'm looking into it but can't yet figure out why.

dakejahl:
Fixes #12394

### DIFF
--- a/src/QmlControls/ParameterEditorController.h
+++ b/src/QmlControls/ParameterEditorController.h
@@ -146,7 +146,7 @@ private:
 private:
     ParameterManager*           _parameterMgr           = nullptr;
     QString                     _searchText;
-    QTimer*                     _searchTimer;
+    QTimer                      _searchTimer;
     ParameterEditorCategory*    _currentCategory        = nullptr;
     ParameterEditorGroup*       _currentGroup           = nullptr;
     bool                        _showModifiedOnly       = false;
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Ah, I thought something like that was already there.

src/QmlControls/ParameterEditorController.h
```diff
@@ -141,9 +141,12 @@ private slots:
 private:
     bool _shouldShow(Fact *fact) const;
 
+    void _performSearch();
+
 private:
     ParameterManager*           _parameterMgr           = nullptr;
     QString                     _searchText;
+    QTimer*                     _searchTimer;
```
DonLakeFlyer, line 149: Why does this need to be allocated?

DonLakeFlyer: Simpler if just a regular member.

dakejahl: Okay changed. I was just following the usage pattern I saw in other files.

dakejahl:
The lag still occurs when the search is invoked but at least now it doesn't occur on every keystroke. I think the issue is the search function is expensive and running it in a connect callback blocks the event loop. The "right" way would be to thread it, but I don't think it matters that much

DonLakeFlyer (approved):
Perfect, thanks Jake

PR #12431, opened 2025-02-12
Vehicle: Fix sendMavCommandWithLambdaFallback showError

### LINKED ISSUE #12429: Missing showError param causes error
Missing showError parameter for sendMavCommand function causes error in function sendMavCommandWithLambdaFallback
https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.cc#L2471

HTRamsey:
Fixes #12429


HTRamsey:
@DonLakeFlyer Just saw you were going to do this, if this is right then we can go with this otherwise I'll just close this

DonLakeFlyer:
It's a lot more complicated than this

PR #12446, opened 2025-02-13
Add collapse arrow animation

kaydikat:
# Description

I noticed that this collapse/expand arrow dissappeared when I clicked on it to expand the section. I feel it is more typical of UI for the arrow to point left when collapsed and down when expanded. 

| **Before** | **After** |
|------------|----------|
| beforeUIfix | afterUIfix |

# Test Steps
1. Open **QGroundControl**.
2. Click **Q** in top left corner.
3. Click **Plan Flight**.
4. Open **Mission Start** on right side of screen.
5. Click the **Vehicle Info** section header to expand it. Collapse/expand arrow should stay visible and rotate to point downwards.
6. Click the **Vehicle Info** section header again to collapse it. Collapse/expand arrow should stay visible and rotate to point left.

### DIFF
--- a/src/QmlControls/SectionHeader.qml
+++ b/src/QmlControls/SectionHeader.qml
@@ -43,7 +43,8 @@ CheckBox {
                 height:                 width
                 source:                 "/qmlimages/arrow-down.png"
                 color:                  qgcPal.text
-                visible:                !control.checked
+                rotation:               control.checked ? 0 : 90
+                Behavior on rotation { NumberAnimation { duration: 200 } }
             }
         }
 


DonLakeFlyer:
The fact that un-collapsed sections do not show the arrow was done on purpose. The reason being that all those arrows tend to increase the UX clutter without adding much value. The collapsed arrows were left in there to show the user that collapsing is possible. The ability to expand/collapse section was added as a "nice to have". It's not a critical feature for everyone to know about. Give that I'd rather have a cleaner ui as a trade off for maybe a small set of users not realizing they can expand/collapse. If a section is expanded it pretty obvious it's expanded. 

I could certainly have my mind changed on all of that if it causes significant confusion. But I've never heard that in the past.

Here you can see an example in one of the more complex Plan items where it comes into play:
Screenshot 2025-02-16 at 10 39 52 AMScreenshot 2025-02-16 at 10 38 51 AM

I know it's subtle but I think it's a lot cleaner. Also with this pull it reverses the direction of the arrow indicator from previous version. which I don't think is a good thing.

PR #12450, opened 2025-02-13
CMake: Fix Android OpenSSL Dependency

HTRamsey:
Was temporarily unavailable using the other link but seems fine now so this may not be needed...
Also testing the NDK version changing the APK size???

### DIFF
--- a/.github/actions/qt-android/action.yml
+++ b/.github/actions/qt-android/action.yml
@@ -42,7 +42,7 @@ runs:
       uses: nttld/setup-ndk@v1
       id: setup-ndk
       with:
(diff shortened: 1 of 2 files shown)


HTRamsey:
@DonLakeFlyer do you have any idea why switching NDK versions like halves the android APK size? My first guess was something wrong with GStreamer libs linking, but it seems to be linked fine although I can't verify video easily until tomorrow when I have access to a camera.

DonLakeFlyer:
Not really. Most size stuff in the past has been cutting down gstreamer crap. Not sure if like on macdeployqt there is a strip symbols things.

PR #12457, opened 2025-02-14
Restore support for Qt 6.6.3 and Android 7.1

rubenp02:
# Restore support for Qt 6.6.3 and Android 7.1

Description
-----------
Restored support for Qt 6.6.3 while preserving the existing build configuration for the primary supported Qt version (6.8.2). Changed the min. Android SDK version to 25 when the Qt version supports it. This allows targeting Android 7.1-based GCS such as the CubePilot Herelink or the SkyDroid H16.

Key changes:
- Lowered the QGC_QT_MINIMUM_VERSION CMake variable to 6.6.3, the latest Qt version compatible with Android 7.1.
- Disabled the CMake features introduced in Qt versions 6.7 and 6.8 when configuring the project for Qt 6.6.3.
- Updated build.gradle to utilize qtGradlePluginType only when available.
- Reintroduced the "package" attribute in AndroidManifest.xml, which is necessary for Qt 6.6.3 but non-disruptive for the main supported version.

Checklist:
----------

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,10 +198,16 @@ if(LINUX)
     find_package(Qt6 ${QT_MINIMUM_VERSION} COMPONENTS WaylandClient)
 endif()
 
+# Set extra standard project setup options for Qt 6.7.0 and above
+set(EXTRA_STANDARD_PROJECT_SETUP_OPTIONS)
+if(QT_VERSION VERSION_GREATER_EQUAL 6.7.0)
+    list(APPEND EXTRA_STANDARD_PROJECT_SETUP_OPTIONS I18N_SOURCE_LANGUAGE en)
+endif()
+
 qt_standard_project_setup(
     REQUIRES ${QGC_QT_MINIMUM_VERSION}
     SUPPORTS_UP_TO ${QGC_QT_MAXIMUM_VERSION}
-    I18N_SOURCE_LANGUAGE en
+    ${EXTRA_STANDARD_PROJECT_SETUP_OPTIONS}
 )
 
 qt_policy(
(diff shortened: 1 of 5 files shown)


HTRamsey:
Yeah I think it's fine until we have to start checking Qt versions in the code all over the place. But we aren't using anything introduced in Qt6.8 yet, except for maybe QtMultimedia related stuff

DonLakeFlyer:
For my current version of Herelink I just downgraded the min sdk version and still use 6.8: https://github.com/DonLakeFlyer/Herelink-QGC/blob/master/custom/cmake/CustomOverrides.cmake

Why not do that instead?

DonLakeFlyer:
Sorry, I'm confused as to where we ended up. I can't remember the driver to move from 6.6.3 to 6.8. I think because 6.8 was LTS maybe? But LTS doesn't help us that much since open source doesn't get LTS updates anyway?

In general I'd rather pick a Qt version and use that for everything. Is there good reason to move to 6.8 from 6.6?

HTRamsey:
I recall one reason was supporting Arm based desktop builds

DonLakeFlyer:
Hmm, yeah. Useful. Maybe I'm overreacting to supporting multiple Qt. In reality the downgrade is just for things like Herelink/Skydroid. If it works it works. How about if we only allowed downgrade for android builds? That limits exposure on desktop.

DonLakeFlyer:
> support 6.6.3 until the next QGC stable to give these Android GCS one last big update

Yeah, kinda what I was thinking as well

DonLakeFlyer:
So I'm gonna merge this since it lets me make forward progress again on Herelink version.

PR #12464, opened 2025-02-15
Prevent taps from bleeding through to lower level controls

### LINKED ISSUE #12453: Clicking the mouse button causes the action to pass through the Value Slider control to the map
## Expected Behavior
Map actions should not be available when clicking on the widget over the map.

## System Information
- Operating System: macOS 15, Ubuntu 22
- QGC Version: latest
- QGC build: Daily

## Detailed Description
Image

DonLakeFlyer:
Fix #12453

### DIFF
--- a/src/QmlControls/ValueSlider.qml
+++ b/src/QmlControls/ValueSlider.qml
@@ -148,8 +148,9 @@ Control {
             }
 
             DragHandler {
-                id:             dragHandler
-                yAxis.enabled:  false
+                id:                 dragHandler
+                grabPermissions:    PointerHandler.TakeOverForbidden
+                yAxis.enabled:      false
             }
 
             BoundaryRule on x {


DonLakeFlyer:
@s-lisovenko Can you give this a try?

s-lisovenko:
> @s-lisovenko Can you give this a try?

I'll try on Monday. I don't remember exactly, but it seems like this solution didn't work in my case, the signal leaked even if the draghandler was removed. but, I tried it on qt 6.8.0. In my case, adding TapHandler helped. I'll look at this fix again, if not, then you can use the solution with taphandler.

DonLakeFlyer:
> the signal leaked even if the draghandler was removed

Hmm, interesting. I found the exact opposite. This whole bleed through by default in qml seems ass-backwards to me.

s-lisovenko:
Didn't help me. I don't understand how it was fixed for you? I checked on both Ubuntu 22 and MacOS 15 apple silicon, tеhe only thing that helps in this case is add TapHandler to whole control.

s-lisovenko:
@DonLakeFlyer I have prepared a PR that fix this issue.

PR #12468, opened 2025-02-16
Fix: Prevent events from passing through ValueSlider to the map

s-lisovenko:
Fix for https://github.com/mavlink/qgroundcontrol/issues/12453
The previous fix mentioned in this PR https://github.com/mavlink/qgroundcontrol/pull/12464 does not work as expected

### DIFF
--- a/src/QmlControls/ValueSlider.qml
+++ b/src/QmlControls/ValueSlider.qml
@@ -128,6 +128,14 @@ Control {
         colorGroupEnabled:  control.enabled
     }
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Thanks, not sure what I did here. I tried so many different things. And I think I was bouncing between 6.8.1 and 6.8.2 as well. Must have screwed something up along the way. I still need to look more at all of this because it's not just taps that bleed through. You can see Drag and Pinch as well bleeds though.

PR #12478, opened 2025-02-19
Android 6.6.3 verification

DonLakeFlyer:


HTRamsey:
@DonLakeFlyer you can't install PR artifacts because they can't be signed

s-lisovenko:
> @s-lisovenko Are you able to install the artifact from here onto a Skydroid? It's not working for me on a Herelink.

Hi, first need to sign the artifact. But I can compile it for local qt6.6.3. I think everything should be ok because the problem is only with qt 6.8.2 (maybe 6.8.1). At the moment qt support does not know if it is a bug or not. They will investigate this issue.

DonLakeFlyer:
Carp, forgot about secrets and PR. I can't seem to get it work when building locally either. But I'll try to get this working first though and then get back to my local problems.

DonLakeFlyer:
Since we are only debug signing I'm gonna move the keystore password to the clear which makes installing PRs for testing possible

PR #12481, opened 2025-02-20
Fix crowdin for docs - download workflow

hamishwillee:
Crowdin currently works for getting QGC strings but not for docs. This is because the default integration only expects one integration and looks for the definition of that file in the root as `crowdin.yml`. 

This uses the GitHub crowdin action to separately manage the docs: https://github.com/crowdin/github-action/tree/master?tab=readme-ov-file#github-crowdin-action-

This is the first part, a test to download any translations into a docs PR. Initially this is an action that has to be manually triggered. We might add support for sync job after.

Note, this uses three access tokens: PX4BUILDBOT_ACCESSTOKEN (already in repo, might not work), CROWDIN_DOCS_PROJECT_ID , PX4BUILDBOT_CROWDIN_PERSONAL_TOKEN

### DIFF
--- a/.github/workflows/crowdin_download.yml
+++ b/.github/workflows/crowdin_download.yml
@@ -1,0 +1,40 @@
+name: Crowdin - Download Translations
+
+# https://github.com/crowdin/github-action/tree/master
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  synchronize-with-crowdin:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: crowdin action
+        uses: crowdin/github-action@v2
+        with:
+          config: 'docs/crowdin.yml'
+          upload_sources: false
+          upload_translations: false
+          download_translations: true
+          localization_branch_name: l10n_crowdin_docs_translations
+          create_pull_request: true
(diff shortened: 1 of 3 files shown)


docs/crowdin.yml
```diff
@@ -0,0 +1,12 @@
+"project_id_env": "CROWDIN_PROJECT_ID"
+"api_token_env": "CROWDIN_PERSONAL_TOKEN"
+"base_path": "../"
```
hamishwillee, line 3: Note, location of project base relative to crowdin.yml (this file)

.github/workflows/crowdin_download.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: crowdin action
+        uses: crowdin/github-action@v2
+        with:
+          config: 'docs/crowdin.yml'
```
hamishwillee, line 23: Note, points to new yaml file for this particular workflow.

docs/crowdin.yml
```diff
@@ -0,0 +1,12 @@
+"project_id_env": "CROWDIN_PROJECT_ID"
```
hamishwillee, line 1: This is the project ID for the docs project. Should have given it that name.

hamishwillee:
Self  merging. Shouldn't impact anyone/thing else.

PR #12517, opened 2025-02-26
Fix map panning yet again

DonLakeFlyer:
Simplest explanation is that using DragHandler's don't work if you ever overlay any other Qml over it in the hierarchy. Qt bug report list is littered with bug reports of people having similar problems. With only clear as mud responses/excuses/explanations of how to work around it. My take is DragHandler's are broken and shouldn't be used.


DonLakeFlyer:
@HTRamsey This is you flickable problem on tablets

PR #12518, opened 2025-02-27
Fix android version code generation

DonLakeFlyer:
Android version code generation must be consistent release to release such that is always and increasing number. Without that you can install over other versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -407,17 +407,36 @@ elseif(ANDROID)
     include(${android_openssl_SOURCE_DIR}/android_openssl.cmake)
     add_android_openssl_libraries(${CMAKE_PROJECT_NAME})
 
-    set(ANDROID_PLATFORM_ARCHITECTURE_CODE)
-    if(${ANDROID_ABI} STREQUAL "armeabi-v7a")
-        set(ANDROID_PLATFORM_ARCHITECTURE_CODE "032")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -407,17 +407,36 @@ elseif(ANDROID)
...
+    endif()
+
+    set(ANDROID_PATCH_VERSION ${PROJECT_VERSION_PATCH})
+    if(${PROJECT_VERSION_PATCH} LESS 10)
+        set(ANDROID_PATCH_VERSION "0${PROJECT_VERSION_PATCH}")
+    endif()
+
+    # Version code format: BBMIPPDDD (B=Bitness, M=Major, I=Minor, P=Patch, D=Dev) - Dev not currently supported and always 000
+    set(ANDROID_VERSION_CODE "${ANDROID_BITNESS_CODE}${PROJECT_VERSION_MAJOR}${PROJECT_VERSION_MINOR}${ANDROID_PATCH_VERSION}000")
+    message("Android version code: ${ANDROID_VERSION_CODE}")
```
HTRamsey, line 439: Maybe this should be a DEBUG message type or something

DonLakeFlyer: I guess, but this version code stuff leads to all sorts of strange problems. That's why it was logged in previous stable as well. It's a pretty important piece of information for android builds. Stray tags cause it to puke.

PR #12524, opened 2025-02-28
PlanManager: prevent memory leak when writing mission items to vehicle

AsensioL:
# Prevent MissionIteam leakage when uploading mission to vehicle

Description
-----------
When uploading a mission to a vehicle, the `MissionController` takes its `VisualMissionItem`s and copies them into a temporary `QList<MissionItem*>`.

Detailed explanation:
1. This starts at `MissionController::sendToVehicle`, which calls `MissionController::sendItemsToVehicle` where `MissionItem`s are allocated by ...
2. calling `MissionController::_convertToMissionItems`, which itself ...
3. defers to the *pure virtual function* `VisualMissionItem::appendMissionItems`, where the implementation depends on the type of `VisualMissionItem`, but they all
4. use `new`, and set the `VisualMissionItem`'s parent to the `Vehicle*` instance.

When the connected vehicle is one that skips the first `MissionItem` (see implementation of `PlanManager::writeMissionItems`) then, the first `MissionItem*` is never moved to the `PlanManager::_writeMissionItems` and is leaked until the `Vehicle*` instance is destroyed (which kicks in Qt's cleaning broom).

Checklist:
----------


DonLakeFlyer:
Yup, thanks for the fix.

PR #12544, opened 2025-03-08
Fix radius reset when editing FW Landing Patterns

rubenp02:
# Fix radius reset when editing FW Landing Patterns

Description
-----------
Resolved an issue where the loiter radius of a fixed-wing Landing Pattern would revert back to its default value when starting to edit the mission item.

Checklist:
----------

### DIFF
--- a/src/PlanView/FWLandingPatternEditor.qml
+++ b/src/PlanView/FWLandingPatternEditor.qml
@@ -41,7 +41,7 @@ Rectangle {
     property string _setToVehicleLocationStr:   qsTr("Set to vehicle location")
     property bool   _showCameraSection:         !_missionVehicle.apmFirmware
     property int    _altitudeMode:              missionItem.altitudesAreRelative ? QGroundControl.AltitudeModeRelative : QGroundControl.AltitudeModeAbsolute
-    property real   _previousLoiterRadius:      0
+    property real   _previousLoiterRadius:      missionItem.loiterRadius.rawValue
 
     Column {
         id:                 editorColumn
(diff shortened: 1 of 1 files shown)


src/PlanView/FWLandingPatternEditor.qml
```diff
@@ -74,12 +74,8 @@ Rectangle {
                 // glide slope heading correctly
                 onCheckedChanged: {
                     if (checked) {
-                        // Restore the previous loiter radius or set the default value
-                        if (_previousLoiterRadius > 0) {
-                            missionItem.loiterRadius.rawValue = _previousLoiterRadius
-                        } else {
-                            missionItem.loiterRadius.rawValue = missionItem.loiterRadius.defaultValue
-                        }
+                        // Restore the previous loiter radius
```
DonLakeFlyer, line 77: This is incorrect. If the item starts without loiter to altitude off. Then the previous loiter radius is 0. Then the user decides they want to use loiter to alt. The old code would set the radius to the default when they turned it on. Now it stays at 0. What's the exact sequence of events which cause the problem?

DonLakeFlyer: Ah, ok. The way I've tended to handle stuff like in the past is like this:
```
property bool _initialLoadComplete: false
Component.onCompleted: _initialLoadComplete = true
...
if (_initialLoadComplete) {
   // do stuff which should only happen after the initial loads finishes all the signalling it might cause
}
'''

DonLakeFlyer: It's a pretty bad user experience to end up with a 0 in there. You want a reasonable value to go in. I'd much rather have it like it was before.

DonLakeFlyer:
Yeah, that should do it. Thanks.

PR #12545, opened 2025-03-08
Fix Loiter (time) mission command for ArduPlane

rubenp02:
# Fix Loiter (time) mission command for ArduPlane

Description
-----------
ArduPlane does not support the radius or heading required parameters in the MAV_CMD_NAV_LOITER_TIME command, [as seen in the source code](https://github.com/ArduPilot/ardupilot/blob/1f51cb395e0453b87c2f94a807b848ca2bda9718/libraries/AP_Mission/AP_Mission.cpp#L1129-L1133).

Marked those parameters as removed in APM-MavCmdInfoFixedWing.json and updated the loiter radius visibility to not show for ArduPlane Loiter (time) mission items.

Checklist:
----------

### DIFF
--- a/src/MissionManager/SimpleMissionItem.cc
+++ b/src/MissionManager/SimpleMissionItem.cc
@@ -537,6 +537,18 @@ bool SimpleMissionItem::isLoiterItem() const
 
 bool SimpleMissionItem::showLoiterRadius() const
 {
+    const MissionCommandUIInfo *uiInfo =
+        MissionCommandTree::instance()->getUIInfo(
+            _controllerVehicle, _previousVTOLMode, MAV_CMD_NAV_LOITER_TIME);
+    bool showUI;
+    uiInfo->getParamInfo(3, showUI);
+
+    if (isLoiterItem() && command() == MAV_CMD_NAV_LOITER_TIME && !showUI) {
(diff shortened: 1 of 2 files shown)


src/MissionManager/SimpleMissionItem.cc
```diff
@@ -537,6 +537,14 @@ bool SimpleMissionItem::isLoiterItem() const
 
 bool SimpleMissionItem::showLoiterRadius() const
 {
+    // Special case: APM fixed-wing doesn't support specifying a radius with Loiter (time)
+    if (_controllerVehicle->apmFirmware() &&
```
DonLakeFlyer, line 541: Ok, so give no radius is correct. It would be much better to look up in the mission command tree whether the current firmware has param3 available or not. You can get to the root of the tree with this method::
```
    const MissionCommandUIInfo *getUIInfo(Vehicle *vehicle, QGCMAVLink::VehicleClass_t vtolMode, MAV_CMD command);
```

Then you can figure out whether param3 is there or not. This way this code only needs to change once to this and then it will track whatever changes happen in the mission json across firmwares and changes which happen in taht json as newer firmwares support new things.

DonLakeFlyer:
Perfect, thanks

PR #12562, opened 2025-03-14
Rework into new MavlinkSettings

DonLakeFlyer:
* Some moved from AppSettings
* Some came from ancient code in MavlinkProtocol which did not use new Settings support
* Some came from MultiVehicleManager which did not use new Settings support
* Deprecated settings are moved to new settings
* Fixes problem with old MavlinkProtocol settings not being saved on android


HTRamsey (commented):
We should try to standardize using either 'MAVLink' or 'Mavlink'

DonLakeFlyer:
> We should try to standardize using either 'MAVLink' or 'Mavlink'

I don't disagree. I'll get to a some point when I'm bored...

PR #12565, opened 2025-03-15
Add fallback takeoff support to the Takeoff action

rubenp02:
# Add fallback takeoff support to the Takeoff action

Description
-----------
Implemented a fallback for the Takeoff action for vehicles that can take off but lack guided takeoff support. The fallback action switches to the vehicle's takeoff mode and then arms it, similar to how the Start Mission action works.

Key changes:
- Introduced a new GuidedTakeoffCapability flag to distinguish between takeoff capability and guided takeoff support.
- Added startTakeoff method in FirmwarePlugin to switch to takeoff mode and arm the vehicle, similar to startMission.
- Defined the takeoff flight mode for ArduPlane.
- Updated GuidedActionsController to use the original guided takeoff action or this fallback, as appropriate.

Checklist:
----------

### DIFF
--- a/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
+++ b/src/FirmwarePlugin/APM/APMFirmwarePlugin.cc
@@ -81,12 +81,15 @@ AutoPilotPlugin* APMFirmwarePlugin::autopilotPlugin(Vehicle *vehicle) const
 bool APMFirmwarePlugin::isCapable(const Vehicle* vehicle, FirmwareCapabilities capabilities) const
 {
     uint32_t available = SetFlightModeCapability | PauseVehicleCapability | GuidedModeCapability | ROIModeCapability;
-    if (vehicle->multiRotor()) {
+    if (vehicle->fixedWing()) {
         available |= TakeoffVehicleCapability;
-        available |= ROIModeCapability;
+    } else if (vehicle->multiRotor()) {
+        available |= TakeoffVehicleCapability;
+        available |= GuidedTakeoffCapability;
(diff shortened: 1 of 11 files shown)


DonLakeFlyer:
> Implemented a fallback for the Takeoff action for vehicles that can take off but lack guided takeoff support.

Which are what exactly?

DonLakeFlyer:
Ok, so the takeoff capability bit was essentially created to know the fact that currently takeoff is not supported by fixed wing (or rover, ...). With your changes it seems like takeoff is now supported by fixed wing as well. Given that, why not just set the takeoff capability bit for fixed wings as well now and then head two different directions in the guidedModeTakeoff implementation based on vehicle type. That seems much simpler when you look in from the outside. It's all just takeoff so no need for two capability bits and two methods which both do the same.

Unless I'm not understanding something...

DonLakeFlyer:
> and there's also the issue of the handler method having "guided" in the name

The "Guided" naming just indicates this is a toolstrip action. For example, PX4 has no concept of a guided mode. Only ArduPilot has the whole "Guided" mode things. I plan to rework toolstrip action support in 5.1 to be way more generic. I'll clean this up then. So for now leave it as is.

Just implement this new stuff in the existing methods instead of adding new ones.

DonLakeFlyer:
> I'll make this use the same handlers as the guided takeoff stuff so it can be merged as is

Can you finish this up and then I'll merge

DonLakeFlyer:
All good. We can revisit when I take a big swing through guided commands rework in 5.1.

PR #12578, opened 2025-03-18
[v4.4] remove CI cache for macOS

julianoes:
Given it only fails after the first build it might have something to do with the caching, so let's just try to remove it for now.


julianoes:
Running the same macOS commit again to check whether it fails on the second as previously.

PR #12580, opened 2025-03-18
Prep for Qt6 QML Registering Method

HTRamsey:
QML_ELEMENT will be used with qt_add_qml_module for each class, comment these out for now so they can be tested and switched over slowly


DonLakeFlyer:
> qt_add_qml_module for each class

How is qt_add_qml_module going to work with custom build resource overrides? I still don't have an answer to that.

HTRamsey:
@DonLakeFlyer that's a good question, but this part refers to cpp classes that don't use resource overrides so it's not immediately relevant to this

DonLakeFlyer:
Ah, ok

Issue #12549, opened 2025-03-11
Custom modes not shown


dakejahl:
PX4 allows users to define "custom modes" from a companion computer according to the specification here:
https://docs.px4.io/main/en/ros2/px4_ros2_control_interface.html

This worked up until maybe 6 months ago, at which some point something changed in QGC and this no longer works. I spent maybe an hour looking into this and this is what I discovered.

**Output from QGC console**
```
StandardModesLog: Completed, num modes: 25
StandardModesLog: Available modes changed, re-requesting
StandardModesLog: Requesting available modes
StandardModesLog: Got mode: "Manual" , idx: 1 , custom_mode 65536
StandardModesLog: Got mode: "Altitude" , idx: 2 , custom_mode 131072
StandardModesLog: Got mode: "Position" , idx: 3 , custom_mode 196608
StandardModesLog: Got mode: "Mission" , idx: 4 , custom_mode 67371008
StandardModesLog: Got mode: "Hold" , idx: 5 , custom_mode 50593792
StandardModesLog: Got mode: "Safe Recovery" , idx: 6 , custom_mode 84148224
StandardModesLog: Got mode: "Position Slow" , idx: 7 , custom_mode 33751040
StandardModesLog: Got mode: "Acro" , idx: 8 , custom_mode 327680
StandardModesLog: Got mode: "Termination" , idx: 9 , custom_mode 655360
StandardModesLog: Got mode: "Offboard" , idx: 10 , custom_mode 393216
StandardModesLog: Got mode: "Stabilized" , idx: 11 , custom_mode 458752
StandardModesLog: Got mode: "Takeoff" , idx: 12 , custom_mode 33816576
StandardModesLog: Got mode: "Land" , idx: 13 , custom_mode 100925440
StandardModesLog: Got mode: "Follow Target" , idx: 14 , custom_mode 134479872
StandardModesLog: Got mode: "Precision Landing" , idx: 15 , custom_mode 151257088
StandardModesLog: Got mode: "Orbit" , idx: 16 , custom_mode 16973824
StandardModesLog: Got mode: "VTOL Takeoff" , idx: 17 , custom_mode 168034304
StandardModesLog: Got mode: "CustomMode" , idx: 18 , custom_mode 184811520
StandardModesLog: Got mode: "(Mode not available)" , idx: 19 , custom_mode 201588736
StandardModesLog: Got mode: "(Mode not available)" , idx: 20 , custom_mode 218365952
[...]

PR #12581, opened 2025-03-18
Cleanup Mavlink Standard Modes Protocol support

DonLakeFlyer:
* Fixes signalling error on flight mode update
* Adds better logging
* Related to #12549

### DIFF
--- a/custom-example/src/FirmwarePlugin/CustomFirmwarePlugin.cc
+++ b/custom-example/src/FirmwarePlugin/CustomFirmwarePlugin.cc
@@ -19,7 +19,7 @@
 //-----------------------------------------------------------------------------
 CustomFirmwarePlugin::CustomFirmwarePlugin()
 {
-    for (auto &mode: _availableFlightModeList){
+    for (auto &mode: _flightModeList){
         //-- Narrow the flight mode options to only these
         if(mode.mode_name != _holdFlightMode && mode.mode_name != _rtlFlightMode && mode.mode_name != _missionFlightMode){
             // No other flight modes can be set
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
@dakejahl Can you give this a try when CI is done. The flight mode list update was being signalled before the list was updated. That said I think the dropdown still should have shown the correct list. We'll see. If it still doesn't work this also has better logging which should tell me what's going on.

dakejahl:
I tried it out, see the attached txt file. I am still not seeing the **CustomMode** in the list
image

[qgc_output_modes.txt](https://github.com/user-attachments/files/19333021/qgc_output_modes.txt)

DonLakeFlyer:
I added more logging, can you try again?

dakejahl:
I renamed the mode to "JakesMode" 
[output.txt](https://github.com/user-attachments/files/19357759/output.txt)

dakejahl (approved):
It works! Thanks Don!

DonLakeFlyer:
Finally. Ok, I need to do some more cleanup here to remove extra logging. Internally although it works it still isn't really correct in exactly how it works. I'll try to find some time to make this functionally correct as well.

fsoest:
@dakejahl Is the custom flight mode added dynamically on your end? I have to restart QGC for the mode to show up if the mode was not started first.

dakejahl:
> @dakejahl Is the custom flight mode added dynamically on your end? I have to restart QGC for the mode to show up if the mode was not started first.

Yes this was also an issue, the mode list wasn't getting update. I think we need to handle [AVAILABLE_MODES_MONITOR](https://mavlink.io/en/messages/common.html#AVAILABLE_MODES_MONITOR)

fsoest:
Yes, using this pull.

DonLakeFlyer:
Ok, can you enter a new issue for this with repro steps

PR #12584, opened 2025-03-18
Fix so plan is shown as synced if just downloaded from vehicle

gillamkid:
If a vehicle with a loaded mission was connected to QGC, when QGC would download the mission, the mission would be marked as "dirty" (aka Plan View thinks it has unsaved changes and prompts the operator to upload the changes). The plan should be marked as clean if it was just downloaded. This commit has a fix so download results in the plan being marked clean (aka has no unsaved changes) as  one would expect.

## Step to Reproduce issue 
1. Start a simulator
```
# run this command in the PX4-Autopilot folder
make px4_sitl gazebo
```
2. Download and extract sample plan 
Here is the exact plan I used. There is nothing special about the plan other than it successfully loads to the vehicle. The `.zip` file will need to be extracted for its inner `simpleMission.plan` file to be selected later.
[simpleMission.zip](https://github.com/user-attachments/files/19328121/simpleMission.zip)
3. Open the plan and upload it to the vehicle
Screenshot from 2025-03-18 14-00-42
Screenshot from 2025-03-18 14-02-11
Screenshot from 2025-03-18 13-49-26
4. After pressing upload, you will see that Plan View considers itself to be in sync with the vehicle. QGC indicates the plan is in sync by showing the `Upload` button (instead of the `Upload Required` button) and making the left-toolbar button icon hollow (instead of containing an explamation mark).
noUnsaved
5. Download the plan just uploaded to the vehicle
Screenshot from 2025-03-18 12-56-03
6. After download you will see either the reported issue if you are using QGC master or you will see the fixed behavior if you built QGC from this PR. The thing to note is that in this PR doing a download sets the the unsaved changes flag (aka the dirty flag) to false and does not prompt the user to save changes.
badMaster
goodPr

## Test Environment
I mostly tested using my local Debug build in Ubuntu 22.04 with a gazebo sim. Also tested a release build of Windows with a WSL Ubuntu 20.04 gazebo sim. Behavior was the same in both.

## Sponsor

### DIFF
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -61,8 +61,8 @@ MissionController::MissionController(PlanMasterController* masterController, QOb
     connect(this,                                               &MissionController::missionPlannedDistanceChanged,      this, &MissionController::recalcTerrainProfile);
 
     // The follow is used to compress multiple recalc calls in a row to into a single call.
-    connect(this, &MissionController::_recalcMissionFlightStatusSignal, this, &MissionController::_recalcMissionFlightStatus,   Qt::QueuedConnection);
-    connect(this, &MissionController::_recalcFlightPathSegmentsSignal,  this, &MissionController::_recalcFlightPathSegments,    Qt::QueuedConnection);
+    connect(this, &MissionController::_recalcMissionFlightStatusSignal, this, &MissionController::_recalcMissionFlightStatus);
+    connect(this, &MissionController::_recalcFlightPathSegmentsSignal,  this, &MissionController::_recalcFlightPathSegments);
     qgcApp()->addCompressedSignal(QMetaMethod::fromSignal(&MissionController::_recalcMissionFlightStatusSignal));
     qgcApp()->addCompressedSignal(QMetaMethod::fromSignal(&MissionController::_recalcFlightPathSegmentsSignal));
     qgcApp()->addCompressedSignal(QMetaMethod::fromSignal(&MissionController::recalcTerrainProfile));


src/MissionManager/MissionController.cc
```diff
@@ -61,8 +61,8 @@ MissionController::MissionController(PlanMasterController* masterController, QOb
     connect(this,                                               &MissionController::missionPlannedDistanceChanged,      this, &MissionController::recalcTerrainProfile);
 
     // The follow is used to compress multiple recalc calls in a row to into a single call.
-    connect(this, &MissionController::_recalcMissionFlightStatusSignal, this, &MissionController::_recalcMissionFlightStatus,   Qt::QueuedConnection);
-    connect(this, &MissionController::_recalcFlightPathSegmentsSignal,  this, &MissionController::_recalcFlightPathSegments,    Qt::QueuedConnection);
+    connect(this, &MissionController::_recalcMissionFlightStatusSignal, this, &MissionController::_recalcMissionFlightStatus);
```
DonLakeFlyer, line 64: This might fix the symptom but it will break other things. The compressing of multiple signals into a signals is there for performance reasons.

DonLakeFlyer:
What are the exact repro steps for this? Also a plan file. It doesn't repro for me.

gillamkid:
> What are the exact repro steps for this? Also a plan file. It doesn't repro for me.

I edited the PR description to have more detailed instructions for reproducing the issue.

src/MissionManager/MissionController.cc
gillamkid: I did a test on how often `_recalcMissionFlightStatus` gets called every time I click `Download` to re-download the plan from the vehicle.

When connect uses `Qt::QueuedConnection`: 5 calls per click
When connect does NOT use  `Qt::QueuedConnection`: 27 calls per click

So there is indeed some resource saving when using the `Qt::QueuedConnection` flag.

DonLakeFlyer:
> I edited the PR description to have more detailed instructions for reproducing the issue.

Fantastic thanks,. I should be able to figure out easily where the missing setDirty is

PR #12590, opened 2025-03-19
Plan: Fix dirty signal handling

DonLakeFlyer:
Replacement for #12584. Repro steps there. Mission recalc routines were incorrectly causing the dirty bit on one of the visual items to change. Hence the whole plan was marked dirty.


DonLakeFlyer:
@gillamkid Can you play around with this a bit once CI finishes?

DonLakeFlyer:
It was a pretty obscure problem deep in the mission code.

PR #12628, opened 2025-04-02
Android CI fix: Prevent future android SDK releases from messing up CI

gillamkid:
## Reason Fix is Needed
"android-35-ext15" was just added and it broke Android CI. Instead of adding a new uninstall command for "android-35-ext15" I made a command which uninstalls anything following the pattern android-* (except for android-33)

## Sponsor

### DIFF
--- a/.github/workflows/android_release.yml
+++ b/.github/workflows/android_release.yml
@@ -74,17 +74,7 @@ jobs:
           add-to-path: false
 
       - name: Remove Android SDKs to force usage of android-33 only
-        run: |
-            ${ANDROID_SDK_ROOT}/cmdline-tools/latest/bin/sdkmanager --uninstall "platforms;android-33-ext5"
-            ${ANDROID_SDK_ROOT}/cmdline-tools/latest/bin/sdkmanager --uninstall "platforms;android-33-ext4"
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Doesn't this need to be in master as well?

gillamkid:
> Doesn't this need to be in master as well?

I don't think so. On master the CI `.yml` files for Android have changed quite a bit and master does not appear to run `uninstall` to remove unused Android SDKs. Here is the equivalent section of code on `mater` https://github.com/mavlink/qgroundcontrol/blob/bafa7320bba6a2a229bc99ac77cf9837521c62df/.github/actions/qt-android/action.yml#L41-L57 
I'm not sure what `master`is doing to avoid the need to fun the `uninstall` command (maybe it's setting the right set of environment variables??) but I would say the change I've suggested in this PR is prolly the smallest/easiest/least-intrusive change that will keep the Android CI build of `Stable_V4.4` working (whereas now it repeatedly breaks as Android releases new SDKs)

HTRamsey:
> Doesn't this need to be in master as well?

No, I've got it set up so this no longer matters

julianoes (approved):
Thanks, that makes a lot more sense!

PR #12634, opened 2025-04-03
Vehicle Card Telem Data

gillamkid:
## Feature Description
Operators are now given a customizable Telemetry data table in each vehicle card!

## Importance
This was a request from a flight test team that was using QGC to fly multiple prototype autonomous fixed wing vehicles. When flying the vehicles, altitude/air-speed needs to be monitored for all vehicles at once to avoid stalls/crashes.

## Screenshots
t1
t2
t3
t4

## Steps to Reproduce
I used PX4's mutli vehicle sim while developing/testing
```
# From PX4-Autopilot directory 
./Tools/simulation/gazebo-classic/sitl_multiple_run.sh -m plane -n 2
```

## Sponsor

### DIFF
--- a/src/API/QGCCorePlugin.cc
+++ b/src/API/QGCCorePlugin.cc
@@ -154,83 +154,121 @@ QString QGCCorePlugin::showAdvancedUIMessage() const
               "Are you sure you want to enable Advanced Mode?");
 }
 
-void QGCCorePlugin::factValueGridCreateDefaultSettings(const QString &defaultSettingsGroup)
+void QGCCorePlugin::factValueGridCreateDefaultSettings(FactValueGrid* factValueGrid)
 {
-    HorizontalFactValueGrid *const factValueGrid = new HorizontalFactValueGrid(defaultSettingsGroup);
+    if(factValueGrid->userSettingsGroup() == HorizontalFactValueGrid::vehicleCardUserSettingsGroup){
+        bool includeFWValues = factValueGrid->vehicleClass() == QGCMAVLink::VehicleClassFixedWing || factValueGrid->vehicleClass() == QGCMAVLink::VehicleClassVTOL || factValueGrid->vehicleClass() == QGCMAVLink::VehicleClassAirship;
 
-    const bool includeFWValues = ((factValueGrid->vehicleClass() == QGCMAVLink::VehicleClassFixedWing) || (factValueGrid->vehicleClass() == QGCMAVLink::VehicleClassVTOL) || (factValueGrid->vehicleClass() == QGCMAVLink::VehicleClassAirship));
+        factValueGrid->setFontSize(FactValueGrid::LargeFontSize);
+        factValueGrid->appendColumn();
+        factValueGrid->appendColumn();
 
-    factValueGrid->setFontSize(FactValueGrid::LargeFontSize);
+        int rowIndex = 0;
+        int colIndex = 0;
 
-    (void) factValueGrid->appendColumn();
-    (void) factValueGrid->appendColumn();
-    (void) factValueGrid->appendColumn();
-    if (includeFWValues) {
+        // first cell
+        QmlObjectListModel* column      = factValueGrid->columns()->value<QmlObjectListModel*>(colIndex++);
+        InstrumentValueData* value = column->value<InstrumentValueData*>(rowIndex);
+        value->setFact("Vehicle", "AltitudeRelative");
+        value->setIcon("arrow-thick-up.svg");
+        value->setText(value->fact()->shortDescription());
+        value->setShowUnits(true);
+
+        // second cell
+        column = factValueGrid->columns()->value<QmlObjectListModel*>(colIndex++);
+        value = column->value<InstrumentValueData*>(rowIndex);
+        if (includeFWValues) {
+            value->setFact("Vehicle", "AirSpeed");
+            value->setText("AirSpd");
+            value->setShowUnits(true);
+        }
+        else {
+            value->setFact("Vehicle", "GroundSpeed");
+            value->setIcon("arrow-simple-right.svg");
+            value->setText(value->fact()->shortDescription());
+            value->setShowUnits(true);
+        }
+
+        factValueGrid->saveSettingsForced();
+    }
+    else if(factValueGrid->userSettingsGroup() == HorizontalFactValueGrid::telemetryBarUserSettingsGroup){
+        const bool includeFWValues = ((factValueGrid->vehicleClass() == QGCMAVLink::VehicleClassFixedWing) || (factValueGrid->vehicleClass() == QGCMAVLink::VehicleClassVTOL) || (factValueGrid->vehicleClass() == QGCMAVLink::VehicleClassAirship));
+
+        factValueGrid->setFontSize(FactValueGrid::LargeFontSize);
+
         (void) factValueGrid->appendColumn();
-    }
-    factValueGrid->appendRow();
+        (void) factValueGrid->appendColumn();
+        (void) factValueGrid->appendColumn();
+        if (includeFWValues) {
+            (void) factValueGrid->appendColumn();
+        }
+        factValueGrid->appendRow();
 
-    int rowIndex = 0;
-    QmlObjectListModel *column = factValueGrid->columns()->value<QmlObjectListModel*>(0);
+        int rowIndex = 0;
+        QmlObjectListModel *column = factValueGrid->columns()->value<QmlObjectListModel*>(0);
 
-    InstrumentValueData *value = column->value<InstrumentValueData*>(rowIndex++);
-    value->setFact(QStringLiteral("Vehicle"), QStringLiteral("AltitudeRelative"));
-    value->setIcon(QStringLiteral("arrow-thick-up.svg"));
-    value->setText(value->fact()->shortDescription());
-    value->setShowUnits(true);
+        InstrumentValueData *value = column->value<InstrumentValueData*>(rowIndex++);
+        value->setFact(QStringLiteral("Vehicle"), QStringLiteral("AltitudeRelative"));
+        value->setIcon(QStringLiteral("arrow-thick-up.svg"));
+        value->setText(value->fact()->shortDescription());
+        value->setShowUnits(true);
 
-    value = column->value<InstrumentValueData*>(rowIndex++);
-    value->setFact(QStringLiteral("Vehicle"), QStringLiteral("DistanceToHome"));
-    value->setIcon(QStringLiteral("bookmark copy 3.svg"));
-    value->setText(value->fact()->shortDescription());
-    value->setShowUnits(true);
+        value = column->value<InstrumentValueData*>(rowIndex++);
+        value->setFact(QStringLiteral("Vehicle"), QStringLiteral("DistanceToHome"));
+        value->setIcon(QStringLiteral("bookmark copy 3.svg"));
+        value->setText(value->fact()->shortDescription());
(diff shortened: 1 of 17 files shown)


DonLakeFlyer:
Wow, cool stuff. Will take me a bit to get to this since I'm out of town now.

gillamkid:
@DonLakeFlyer Back in town yet? :)

DonLakeFlyer:
Yeah, just got back. I'll start digging through stuff soon.

DonLakeFlyer:
There's some console logging that was left in somewhere:
```
this:  HorizontalFactValueGrid_QMLTYPE_385(0x12a45ac20, id="valueArea", parent=0x12a453d80, geometry=0,0 283.875x56) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  MultiVehicleFactValueGrid_QMLTYPE_311(0x1695c7ae0, id="valueGrid", parent=0x16ffd9590, geometry=0,0 169x28) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  HorizontalFactValueGrid_QMLTYPE_385(0x12a45ac20, id="valueArea", parent=0x12a453d80, geometry=0,0 283.875x56) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  MultiVehicleFactValueGrid_QMLTYPE_311(0x12ed28b20, id="valueGrid", parent=0x12ed20330, geometry=0,0 169x28) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  HorizontalFactValueGrid_QMLTYPE_385(0x12a45ac20, id="valueArea", parent=0x12a453d80, geometry=0,0 283.875x56) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  MultiVehicleFactValueGrid_QMLTYPE_311(0x1695c7ae0, id="valueGrid", parent=0x16ffd9590, geometry=0,0 169x28) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  HorizontalFactValueGrid_QMLTYPE_385(0x12a45ac20, id="valueArea", parent=0x12a453d80, geometry=0,0 283.875x56) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  MultiVehicleFactValueGrid_QMLTYPE_311(0x12ed28b20, id="valueGrid", parent=0x12ed20330, geometry=0,0 169x28) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  HorizontalFactValueGrid_QMLTYPE_385(0x12a45ac20, id="valueArea", parent=0x12a453d80, geometry=0,0 283.875x56) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  MultiVehicleFactValueGrid_QMLTYPE_311(0x1695c7ae0, id="valueGrid", parent=0x16ffd9590, geometry=0,0 169x28) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  HorizontalFactValueGrid_QMLTYPE_385(0x12a45ac20, id="valueArea", parent=0x12a453d80, geometry=0,0 283.875x56) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  MultiVehicleFactValueGrid_QMLTYPE_311(0x12ed28b20, id="valueGrid", parent=0x12ed20330, geometry=0,0 169x28) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
this:  HorizontalFactValueGrid_QMLTYPE_385(0x12a45ac20, id="valueArea", parent=0x12a453d80, geometry=0,0 283.875x56) spath:  "/Users/don/.config/org.qgroundcontrol"  exists:  true
```

Or maybe that was already there from before. If so could you remove that, or maybe comment it out if it's useful.

DonLakeFlyer:
Did you look through the second half of:
```
void InstrumentValueData::_activeVehicleChanged(Vehicle* activeVehicle)
{
    if (_vehicle) {
        disconnect(_vehicle, &Vehicle::factGroupNamesChanged, this, &InstrumentValueData::_lookForMissingFact);
    }

    if (!activeVehicle) {
        activeVehicle = MultiVehicleManager::instance()->offlineEditingVehicle();
    }

    _vehicle = activeVehicle;
    connect(_vehicle, &Vehicle::factGroupNamesChanged, this, &InstrumentValueData::_lookForMissingFact);

    emit factGroupNamesChanged();

    if (!_factGroupName.isEmpty() && !_factName.isEmpty()) {
        _setFactWorker();
    }
}
```

That is not being called in the multi-vehicle case. Don't quite remember what this all does but it seems fishy that it can be skipped?

src/QmlControls/InstrumentValueData.cc
```diff
@@ -23,13 +23,26 @@ const QStringList InstrumentValueData::_rangeTypeNames = {
     QT_TRANSLATE_NOOP("InstrumentValue", "Icon"),
 };
 
+InstrumentValueData::InstrumentValueData(FactValueGrid* factValueGrid, QObject* parent, Vehicle* vehicle)
```
DonLakeFlyer, line 26: Could you comment the two constructors to explain the difference?

src/QmlControls/FactValueGrid.h
```diff
@@ -50,6 +52,7 @@ class FactValueGrid : public QQuickItem
     Q_PROPERTY(QStringList          iconNames                       READ iconNames                                          CONSTANT)
     Q_PROPERTY(FontSize             fontSize                        READ fontSize                       WRITE setFontSize   NOTIFY fontSizeChanged)
     Q_PROPERTY(QStringList          fontSizeNames                   MEMBER _fontSizeNames                                   CONSTANT)
+    Q_PROPERTY(Vehicle *            vehicle                         READ vehicle                        WRITE setVehicle)
```
DonLakeFlyer, line 55: Missing a NOTIFY here on vehicle change.

DonLakeFlyer: Actually I think this property should be `REQUIRED`.

src/QmlControls/FactValueGrid.cc
```diff
@@ -96,6 +102,14 @@ void FactValueGrid::setFontSize(FontSize fontSize)
     }
 }
 
+void FactValueGrid::setVehicle(Vehicle * vehicle)
+{
+    _vehicle = vehicle;
```
DonLakeFlyer, line 107: And then given the `REQUIRED` property I would change this to two things:
* qCritical if incoming vehicle is null
* And then qCritical if called when _vehicle is not null which indicates multiple calls. And then you don't need the _componentCompleted flag.

src/QmlControls/FactValueGrid.cc
```diff
@@ -273,6 +297,28 @@ void FactValueGrid::_saveSettings(void)
         settings.endArray();
     }
     settings.endArray();
+
+    // If this this settings change was set from a Vehicle card, this makes so the changes are
```
DonLakeFlyer, line 301: "this this" typo

src/QmlControls/MultiVehicleFactValueGrid.qml
```diff
@@ -0,0 +1,201 @@
...
+            }
+
+            ColumnLayout {
+                spacing: 1
+                visible: settingsUnlocked
+
+                QGCButton {
+                    Layout.preferredWidth:  ScreenTools.minTouchPixels
+                    Layout.fillHeight:      true
+                    topPadding:             0
```
DonLakeFlyer, line 120: I'm assuming `padding: 0` doesn't work because QGCButton farts around with the individual paddings?

DonLakeFlyer:
I also noticed that the mouse wheel bleeds through the multi-vehicle control to the map. You should be a DeadMouseArea as the lowest child control filling the whole things. THis will prevent bleed-through.

src/QmlControls/MultiVehicleFactValueGrid.qml
gillamkid: this entire file was copied, so I copied the way it was done there. This comment is no longer relevant because I got rid of MulitVehicleFactValueGrid entirely (found a way to reuse the existing HorizontalFactValueGrid)

gillamkid:
@DonLakeFlyer Thanks for your feedback. I agree with your comments and have addressed.

note: Upon taking a second look, I found a couple ways to refactor which simplified the changes made in this PR
1. re-use `HorizontalFactValueGrid` instead of creating new `MultiVehicleFactValueGrid` type (no longer need to duplicate as much code!)
2. I also made `InstrumentValueData` have a single constructor and not have any code in it specific to handling the `activeVehilce`, whereas before it had a couple constructors, one for when its `vehicle` arg wouldn't change and one for when `activeVehilce` was the vehicle of interest

DonLakeFlyer:
> I found a couple ways to refactor which simplified the changes made in this PR

Yeah, all nice stuff. I debated asking for the single constructor thing as well.

DonLakeFlyer:
My only other comment would be that I think you might be able to use a smaller variant of the compass edge pitch/roll indicators in the multi-view. It might at least give you an indication of one of the vehicles is tumbling out of the sky :)

gillamkid:
> My only other comment would be that I think you might be able to use a smaller variant of the compass edge pitch/roll indicators in the multi-view. It might at least give you an indication of one of the vehicles is tumbling out of the sky :)

Sure. Done!
beforeAfter2

DonLakeFlyer:
So this should be all ready to go now right?

gillamkid:
> So this should be all ready to go now right?

Yep, I'm done making changes unless required for approval

src/API/QGCCorePlugin.cc
```diff
@@ -154,83 +154,122 @@ QString QGCCorePlugin::showAdvancedUIMessage() const
...
+        value->setFact(QStringLiteral("Vehicle"), QStringLiteral("FlightDistance"));
+        value->setIcon(QStringLiteral("travel-walk.svg"));
+        value->setText(value->fact()->shortDescription());
+        value->setShowUnits(true);
+
+        factValueGrid->saveSettingsForced();
+    }
+    else {
+        qCritical() << "factValueGridCreateDefaultSettings: Unexpected userSettingsGroup: " << factValueGrid->userSettingsGroup();
+        exit(1);
```
DonLakeFlyer, line 271: This should be removed. You never want to crash out of QGC. You always want to try to keep it running at all costs since there may be a vehicle connected on the other side. Better to have strange behavior than lose full control.

DonLakeFlyer (changes requested):

src/API/QGCCorePlugin.cc
gillamkid: done

DonLakeFlyer (approved):

src/QmlControls/FactValueGrid.cc
```diff
@@ -96,6 +101,15 @@ void FactValueGrid::setFontSize(FontSize fontSize)
     }
 }
 
+void FactValueGrid::setVehicle(Vehicle * vehicle)
+{
+    if(vehicle == nullptr){
+        qCritical() << "FactValueGrid assumes vehicle is not NULL)";
```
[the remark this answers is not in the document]
DonLakeFlyer: One last type here...,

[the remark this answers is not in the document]
gillamkid: done

HTRamsey:
Does this accidentally change the default Telemetry bar to look like this? This is what I get by default after clearing cache/config
image

PR #12646, opened 2025-04-08
CMake: GStreamer Linking Updates

HTRamsey:
Minor updates to CMake for GStreamer, and prep for using GStreamer provided FindGStreamer.cmake.
Also fix FindGStreamer for MacOS


HTRamsey:
@DonLakeFlyer If you are able to get a chance at some point, can you double check that this works for MacOS correctly?

PR #12649, opened 2025-04-08
Utilities: Improve Debug Logging Output

HTRamsey:
image
@DonLakeFlyer thoughts on the debug output formatting? This will remove the need to add the function to the log message every time too


DonLakeFlyer:
Yeah, I like it

PR #12652, opened 2025-04-09
test: see if Stable_V4.4 is failing CI

gillamkid:


gillamkid:
A fix for the broken tool is currently in progress. Progress on the fix and a temporary way to fix can be seen here: https://github.com/jurplel/install-qt-action/pull/282

gillamkid:
looks like the issue has resolved itself. The issue may popup again in the future. See https://github.com/jurplel/install-qt-action/issues/281#issuecomment-2789853555 for explanation

PR #12658, opened 2025-04-09
Vehicle: Facts Formatting

HTRamsey:
Cleanup Vehicle Facts


HTRamsey:
I don't disagree, especially with Q_PROPERTY. I'll adjust spacing back

PR #12686, opened 2025-04-18
GNSS base station menu updates

Louis-max-H:
This PR aim to develop base station menu and depend on this PR for [GPS Driver](https://github.com/PX4/PX4-GPSDrivers/pull/178)

Description
-----------

UBlox menu | Septentrio menu             |  Manufacturer list
:-------------------------:|:-------------------------: | :---: 
 |   |  
 | Ps: Protocols displayed in Septentrio menu will be part of another PR |

1) Proposing different settings according to receiver manufacturer
2) Base Station mode change from bool to int. That will allow a future PR that have base station (for example, moving base station on boat)
3) Implement base station menu for Septentrio (base infos, base config)

### Explanation

1) Manufacturer allow the hide and show of some settings to keep only theses that are implemented.
We use bit mask defined like this :
```
    readonly property var    _standard:           0b00001
    readonly property var    _trimble:            0b00010
    readonly property var    _septentrio:         0b00100
    readonly property var    _femtomes:           0b01000
    readonly property var    _ublox:              0b10000
```

To restrict a parameters' visibility to U-Blox + Septentrio : (You can use either | or +)
```
visible:    manufacturer & (_ublox | _septentrio)
```
If implemented by all manufacturer :
```
visible:    manufacturer & _standard
```
This parameter is always visible because it is implemented by everyone, but I think that specifying it each time avoids uncertainties such as, how compatible is this parameter with the receiver? Has the developer thought this through?
Don't hesitate to tell me what you think :)

To show parameters available for U-blox, set manufacturer to `_standard + _ublox`
Manufacturer is set automatically when plugin a receiver.

2) Base station mode is now :
0 : Survey-in
1 : Fixed
This change is compatible with the 4 different drivers that use BaseSettingsType (base_station.h).

3)
a) Base stations info is display by using PosGeodetic information.
b) Some parameters like Ublox min precision don't have an equivalent in Septentrio, here is a quick recap of with parameters is available or not :

| Parameters​       |                   | Ublox​ | Septentrio​ | Femtomes​ | Trimble​ |
| -----------------|------------------ | :----------:| :-----------: | :---------: | :--------: |
| Survey-in (auto)​ | Min duration​      | ✅          | ​        | ✅​       | ✅​ |
| ​                 | Min precision​     | ✅​          |         | ​       | ​ |
| Specify location​ | Long, Lat, Alt​    | ✅​          | ✅​        | ✅​       | ✅​ |
| ​                 | Base precision​    | ✅​          | ​        | ​       | ​ |
| ​                 | Set from position​ | ✅​          | ✅​        | ✅​       | ✅​ |

Survey-in options don't have equivalents for Septentrio receivers.

Test Steps
-----------
I have followed this procedure and another intern (@Duck) have done the same on another computer.
```bash
git clone https://github.com/Louis-max-H/qgroundcontrol
cd qgroundcontrol
git checkout pr-base-qgc
git submodule update --init --force --recursive
chmod +x tools/setup/*.sh
docker build --file ./deploy/docker/Dockerfile-build-ubuntu -t qgc-linux-docker .
```
Edit the file src/GPS/CMakeLists.txt
To have
```
CPMAddPackage(
    NAME px4-gpsdrivers
    GITHUB_REPOSITORY Louis-max-H/PX4-GPSDrivers
    GIT_TAG pr-base-station-px4
    SOURCE_SUBDIR src
)
```
Instead of :
```
CPMAddPackage(
    NAME px4-gpsdrivers
    GITHUB_REPOSITORY PX4/PX4-GPSDrivers
    GIT_TAG main
    SOURCE_SUBDIR src
)
```
You can now build QGC using :
```
docker run --privileged --rm -v ${PWD}:/project/source -v ${PWD}/build:/project/build qgc-linux-docker
./build/AppDir/AppRun
```

- Unplug all
- Plug CubeOrangePlus
- Select a different manufacturer than the board tested
- Select the config you want to test
- Plug base station
- See if :
    - Precision is improved, position is RTK
    - Information are displayed
    - Manufacturer is autautomaticallyt

Do these steps for Survey-In, Set location
Test have been made on Ublox and Septentrio, I don't have the equipment to test other GPS systems but they should work has Ublox does.

Checklist:
----------
- [x ] [Review Contribution Guidelines](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CONTRIBUTING.md).
- [ x] [Review Code of Conduct](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CODE_OF_CONDUCT.md).
- [ x] I have tested my changes.

Please let me know if I need to make any changes to this RP so that it can be accepted by the community, this would be my first PR :partying_face: 

Ps: The automatic tests could fail because of the need to modify the Makefile, so I'm putting the PR in draft form and I'll come back to you once the PR for the GPS driver has been accepted.


src/GPS/GPSProvider.h
```diff
@@ -42,7 +42,7 @@ class GPSProvider : public QThread
     struct rtk_data_s {
         double surveyInAccMeters = 0;
         int surveyInDurationSecs = 0;
-        bool useFixedBaseLoction = false;
+        int baseMode = 0;
```
bkueng, line 45: Use an enum?

src/Settings/RTK.SettingsGroup.json
```diff
@@ -3,9 +3,17 @@
     "fileType":     "FactMetaData",
     "QGC.MetaData.Facts":
 [
+{
+    "name":                 "baseReceiverManufacturers",
+    "shortDesc":            "GPS manufacturers for settings",
+    "type":                 "uint8",
+    "enumStrings":          "Standard,All,Trimble,Septentrio,Femtomes,UBlox",
```
bkueng, line 10: What is the meaning of `Standard` (I see what it means in the code, but from a user perspective)?
And shouldn't this be determined automatically based on the connected gps type?

src/UI/toolbar/GPSIndicatorPage.qml
```diff
@@ -27,7 +27,33 @@ ToolIndicatorPage {
...
-    property bool   useFixedPosition:   rtkSettings.useFixedBasePosition.rawValue
+    property var    baseMode:           rtkSettings.baseMode.rawValue
+    property var    manufacturer:       rtkSettings.baseReceiverManufacturers.rawValue
+    readonly property var    _standard:           0b00001
+    readonly property var    _trimble:            0b00010
+    readonly property var    _septentrio:         0b00100
+    readonly property var    _femtomes:           0b01000
+    readonly property var    _ublox:              0b10000
+    
+    /* Manufacturer is used to determine witch parameters to displays
```
bkueng, line 38:
```suggestion
    /* Manufacturer is used to determine which parameters to displays
```

bkueng (commented):
Nice, having type-dependent options makes sense.
But ideally the manufacturer is determined automatically, as we already know it when connecting the device.

src/Settings/RTK.SettingsGroup.json
DonLakeFlyer: Are these values "1,31,3,5,9,17" somewhere in a driver code enum somewhere? If so, can that be used on the C++ side instead of all the magic numbers in the code? If not, can an enum be added to RTKSettings such that we don't use magic numbers on the c++ side?

DonLakeFlyer:
FYI: I'm going to branch for QGC 5.0 RC next week. So if this wants to go into 5.0 it needs to hurry.

Louis-max-H (commented):
Thanks for the info!
I think everything's already good, I've added a version with enumeration for the baseMode and I tested this morning that the menu works well.

Here's a quick recap of the latest changes:
- baseMode use enum as suggested by bkueng
- Bitmask calculation moved in c++ file to avoid magic constant

DonLakeFlyer:
> But ideally the manufacturer is determined automatically, as we already know it when connecting the device.

Is this supposed to be working? I have a Ublox RTK which I plugged in. When I went to the page the manufacturer was set to all.

Louis-max-H:
That's very strange ...

I have test again with an Ublox ZED F9P an the menu was detected.
Detection is made in src/GPS/GPSRtk.cc, but the if/else tests should only set manufacturer to Ublox/Septentrio/Fentom/Trimble or Standard, but not All.

I've also checked that there's no inconsistency with regard to enumerations (0 = Standart, 1 = All).

I have made a little improvement to the print pattern, with the latest changes can you :
- Check the message send by qCDebug(GPSRtkLog)
    1) Connecting U-blox device -> Manufacturer to Ublox
    2) Connecting device has U-blox by default" -> Manufacturer to default
    
- Close menu and open again
     1) Settings are correctly updated even with the menu open, only the manufacturer is not changed when menu open. Cf gif.
demo

- We are actully checking for "blox" in the name of the receiver, maybe we can change it to something else ?
     1) Can you check for the name of your receiver has detected by QGC ?
     2)  qCDebug(GPSRtkLog) << gps_type; to add in GPSrtk.cc
     
I hope it will help to understand what is happening !

Louis-max-H:
After discussion with another QGC user, the "manufacturer" reference may be misleading.

He thought that this choice had an impact on the protocol used, whereas it's just a graphics option to hide some settings, so I changed it to "Settings displayed".

I have tested this change.

Louis-max-H:
Do you have any news @DonLakeFlyer ?
Feel free to tel me if I can help you somehow !

DonLakeFlyer:
I'll get back to this once we get 5.0 out the door

Louis-max-H:
FYI, the pull request on PX4 have been merged :smiley: 

My internship will end in 3 weeks, feel free to ask me questions, I could respond even after, but I will be easier for me to finish it before,

DonLakeFlyer:
What is the "Standard" selection?

Louis-max-H:
Standard is for settings implemented by all receivers.

DonLakeFlyer:
Ok, so what is the difference between Standard and All then? Why are both needed?

DonLakeFlyer:
Yeah, I think that makes more sense.

Louis-max-H:
@DonLakeFlyer Do you know if you can merge this PR ?

Tory9 will be taking over these changes since it will be the end of my internship !
I'll continue to follow progress, but I'll have limited access to the hardware,

Tory9:
@DonLakeFlyer hi, can you pls let me know if we can merge this request

src/Settings/RTK.SettingsGroup.json
```diff
@@ -30,11 +38,11 @@
     "qgcRebootRequired":    true
 },
 {
-    "name":                 "useFixedBasePosition",
-    "shortDesc":     "Use specified base position",
-    "longDesc":      "Specify the values for the RTK base position without having to do a survey in.",
-    "type":                 "bool",
-    "default":         false,
+    "name":                 "baseMode",
```
DonLakeFlyer: I don't get why this changed? It's still basically a boolean value with only 0 or 1 as valid. What was wrong with the way it was before. The problem with changing settings like this is that all user who have set this value will lose their specific setting when they upgrade without really knowing it.

Tory9: The idea was to change from book to enum to update in future code to add moving base

DonLakeFlyer: OK, that's fine. But it should be done in a way that doesn't affect existing settings which may have been saved:
* Leave the `baseMode` name the same so you don't trash peoples existing settings
* Change the `baseMode` json to use `enumStrings/Values` and encode the values such that 0 and 1 mean the same thing as they did when the setting was a bool.

This way for existing users they are not affected.

Tory9: okay, thank you.  i will implement the changes and tag you when its done (next week most probably)

Tory9:
@DonLakeFlyer i have implemented the changes you mentioned. Could you please review if it is okay now?

Tory9:
Hello @DonLakeFlyer, could you pls let me know if there is anything you need me to address before the dev call tomorrow? We're hoping to get this ready for final review.

DonLakeFlyer:
I did quick look today and I'm pretty certain it's all good now. I'm going to have a bit more time tomorrow just to make sure.

mrpollo:
Nice updates, thanks @Tory9. Hopefully, we can get these into a release soon.

Tory9:
Thanks so much for the final review and merge, @DonLakeFlyer. Since the PX4 part of this feature will be in its next release, we're hopeful this can make it into the next QGC release as well to tie it all together. Looking forward to any updates on release planning!

PR #12709, opened 2025-04-21
VideoManager: Verify GStreamer Plugins Initialized

HTRamsey:
Normally you would just get the error about tee failing. This makes it more obvious to users what is happening


DonLakeFlyer:
Funny, I was going to ask you if there was a way to detect plugin screw up in a less roundabout way.

DonLakeFlyer (approved):

PR #12721, opened 2025-04-23
Add motor interlock aux function to joystick button options

bnsgeyer:
Add motor interlock aux function to joystick button options

Description
This PR adds the motor interlock feature of Ardupilot to the options for button actions in the joystick. It uses the mavlink aux function command ardupilot to enable and disable the motor interlock.

Test Steps
-----------
Start an helicopter simulation in ardupilot. Add motor interlock enable as an action on one button and motor interlock disable as an action on a separate button.  Press the motor interlock disable button to allow arming. Arm the heli and use the motor interlock enable button to spool up the rotor. 

Checklist:
----------
- [x] [Review Contribution Guidelines](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CONTRIBUTING.md).
- [x] [Review Code of Conduct](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CODE_OF_CONDUCT.md).
- [ ] I have tested my changes.

Related Issue
-----------
None 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #12722, opened 2025-04-23
Vehicle: Only Allow Motor Interlock Aux Function w/ APM

HTRamsey:
Follow up to #12721 that ensures usage of AUX_FUNCTION only w/ APM.
The separation of this kind of stuff really needs to be improved at some point

### DIFF
--- a/src/FirmwarePlugin/APM/APM.h
+++ b/src/FirmwarePlugin/APM/APM.h
@@ -1,0 +1,214 @@
+/****************************************************************************
+ *
+ * (c) 2009-2024 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ *
+ * QGroundControl is licensed according to the terms in the file
+ * COPYING.md in the root of the source code directory.
+ *
+ ****************************************************************************/
+
+#pragma once
+
+namespace APM
+{
+
+enum AUX_FUNC {
+    DO_NOTHING =           0, // aux switch disabled
+    FLIP =                 2, // flip
+    SIMPLE_MODE =          3, // change to simple mode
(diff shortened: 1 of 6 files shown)


bnsgeyer:
@HTRamsey Thanks for doing this.  Since this was an ArduPilot specific aux function, I was wondering how that would be segregated to be only active for Ardupilot FC.

HTRamsey:
No problem, not a fan of using QGC_NO_ARDUPILOT_DIALECT so much everywhere, so I'll have to come back and make a cleaner approach that works for all other APM stuff too

DonLakeFlyer:
So the goal of `QGC_NO_ARDUPILOT_DIALECT` was two-fold:
* Allow custom builds to work only with a specific firmware type. This results in a slight reduction in UI complexity. Specifically because there are some parts of the QGC ui which end up being complicated by not knowing the firmware type. A lot of this shows up in creating Plans while not connected to a vehicle.
* Reduce the code size of a custom build by removing the other firmware support you don't need.

In reality the first goal of UI simplification is the way more important one. I have a feeling the second size reduction goal doesn't really accomplish much from a size standpoint and just adds complexity to the code with the 'ifdef' crap. I'd be fine with removing the second thing as a goal for QGC keeping only the first thing.

HTRamsey:
@DonLakeFlyer So in this case should I just remove all usage of QGC_NO_ARDUPILOT_DIALECT except for the one place where it creates a list of actions that show up in the qml list?

DonLakeFlyer:
Why don't you enter an Issue against me to clean this up. I need to remember what should go and what should stay.

PR #12735, opened 2025-04-25
Simple boot test now verifies gstreamer

DonLakeFlyer:


DonLakeFlyer:
@HTRamsey Can you take a look at how I'm sanity checking the gstreamer builds. You can see from this now that both Windows and Mac fail to run gstreamer from a dev build. Note that this does not currently verify the builds which people install from.

Still left to do:
* Fix the dev build problem
* Add an additional simple boot test run on the "installed" version of QGC
* And the above will likely show that Mac gstreamer isn't working from DMG install so then figure out how to fix that

HTRamsey:
Yeah I think adding a step to run the windows installer & verify the installation should help eliminate any future issues.

PR #12751, opened 2025-04-29
Prevent extra values from showing up in bar

### LINKED ISSUE #12748: Initial Telemetry Grid row/column size is off
### Description

From #12634
@gillamkid 

### System Information

Daily v5.0

### Log Files and Screenshots

Image

DonLakeFlyer:
Fix #12748


HTRamsey:
I think there's still one more issue. I believe you have to swap around the order of _vehicleChanged otherwise I get a crash

DonLakeFlyer:
Yeah, that's incorrect as well.

PR #12775, opened 2025-05-04
Add KakuteH7-Wing board ID

julianoes:
This goes together with https://github.com/PX4/PX4-Autopilot/pull/24669.

FYI @vincentpoont2


dakejahl (approved):

HTRamsey:
should probably make that mapping an override-able resource like USBBoardInfo.json at some point

PR #12777, opened 2025-05-05
[V4.4] Enable standard modes

julianoes:
This was already working in the Daily Build, so - I think - it makes sense to enable this for v4.4.5.


julianoes:
@farhangnaderi any idea why macOS CI is failing here?

farhangnaderi:
> @farhangnaderi any idea why macOS CI is failing here?

I did not have any idea about it. But seems fixed. Sorry I was at Xponential and missed this. 
Thank you.

PR #12781, opened 2025-05-05
Follow Me: Don't trust update interval from device

### LINKED ISSUE #11245: GQC Stable 4.3 crashes if "stream GCS position" option is set on "Always"
## Current Behavior
When using the QGC stable 4.3.0 app (not the herelink app version!!) with Herelink remote controller and "stream GCS position" option is set on "Always", if the herelink remote controller has gps signal, the QGC app starts to break up leading to intense cpu load. The connection bar to vehicle starts to take longer to load and if using video stream to rtsp I notice the video starts to break up as well. This sometimes leads to the app crashing or the remote controller reboot.

## Steps to Reproduce:
1. Start QGC app
2. Turn option "Stream GCS position" to "Always"
3. Turn on vehicle
4. **Make sure GPS on the remote controller (GCS running QGC) is turned on and has satellites!**
5. Restart QGC app
6. Initial connection to vehicle starts to break up
7. Video stream from rtsp also starts to break up

## System Information
- Operating System: Android 7.1
- QGC Version: 4.3.0
- QGC build: Stable
- Flight Controller: Matek H743 Mini
- Autopilot (with version): Ardupilot Copter 4.4.4

### LINKED ISSUE #12564: FOLLOW_TARGET message sent at extreme rate
## Expected Behavior
When enabling the "Stream GCS Position" function using a NMEA GPS in "Always" mode, QGC should send the FOLLOW_TARGET message perhaps a few times per second, or at the rate of the data coming from the GPS.

## Current Behavior

When enabling the "Stream GCS Position" function using a NMEA GPS in "Always" mode, QGC tries to send the FOLLOW_TARGET message tens of thousands of times per second. This saturates the vehicle comm link and eventually fills up the hard disk by creating MAVLink telemetry logs at a dozen megabytes a second.

## Steps to Reproduce:
1. Set up NMEA GPS device in the AutoConnect settings
2. Set Stream GCS Position mode to Always
3. Connect to vehicle
4. Connect GPS to computer
5. Observe connection becomes sluggish and telemetry log grows at an extreme rate.

## System Information
- Operating System: Xubuntu 22.04
- QGC Version: 4.4.3
- QGC build: stable
- Flight Controller: Cube Orange Plus
- Autopilot (with version): ArduCopter 4.5.4

## Detailed Description
The NMEA GPS is a Holybro UM982 module and is configured to send data at 1Hz.

I think the root cause of the bug is this function:

https://github.com/mavlink/qgroundcontrol/blob/0243afc521428855b842df9e4fff1b7af8ed4b7e/src/FollowMe/FollowMe.cc#L79-L85

The `updateInterval()` defaults to 0, so the `qMin` will set the timer for 0 milliseconds, evidently causing the function to send the data to be called as fast as possible.

I think the `qMin` should instead be `qMax`. A 250 millisecond (value of `kMotionUpdateInterval`) lower bound is quite reasonable. I am happy to send a PR to fix this if my assessment is accurate.

## Log Files and Screenshots

From the ArduPilot stream stats web tool:

Image

DonLakeFlyer:
* Fix #11245
* Fix #12564

### DIFF
--- a/src/FollowMe/FollowMe.cc
+++ b/src/FollowMe/FollowMe.cc
@@ -29,6 +29,10 @@ FollowMe::FollowMe(QObject *parent)
     // qCDebug(FollowMeLog) << Q_FUNC_INFO << this;
 
     _gcsMotionReportTimer->setSingleShot(false);
+    // We set the update interval to a fixed amount of time. 
+    // Even if the device can't update this quickly we'll pick up a new value on the next time around.
+    // We can't trust the value for update interval which comes from the actual device.
+    _gcsMotionReportTimer->setInterval(kMotionUpdateInterval);
 }
 
 FollowMe::~FollowMe()
@@ -79,7 +83,6 @@ void FollowMe::_settingsChanged(QVariant value)
 void FollowMe::_enableFollowSend()
 {
     if (!_gcsMotionReportTimer->isActive()) {
-        _gcsMotionReportTimer->setInterval(qMin(QGCPositionManager::instance()->updateInterval(), kMotionUpdateInterval));
         _gcsMotionReportTimer->start();
     }
 }


DonLakeFlyer:
Have you ever done any Follow Me ArduPilot stuff? I'm not set up to test this.

HTRamsey:
Yeah, I don't recall having any particular issues like this but I'll double check

HTRamsey:
@DonLakeFlyer I just checked and some of the positioning plugins in Qt have update intervals like 1ms which is far beyond the needs of FollowMe, so I think doing this is good.

HTRamsey (approved):

PR #12787, opened 2025-05-06
gz: rate limit pipeline retries on failure

dakejahl:
Currently when pipeline construction fails in `makeSource` the worker thread immediately retries. This causes the retries to occur extremely rapidly, like thousands of times per second. 

I spent a bunch of time trying to pass a delay_ms variable into the thread dispatcher from `_initVideoReceiver`, which is the thing that kicks everything off from the `onStartComplete` signal, but it ended up being way too complicated. This implementation just sleeps for a second after failure before emitting the signal, which is much simpler and accomplishes the goal.


HTRamsey:
I think this also caused an insane amount of debug output, so hopefully that's not as bad now either

PR #12789, opened 2025-05-06
Hardwire MAC GST to not use find_package which does not work

DonLakeFlyer:
This should hopefully get mac video at least working. It's still laggy which is a separate issues.


DonLakeFlyer:
Verified laggy video at least works now.

PR #12791, opened 2025-05-07
PX4FirmwarePlugin: change yaw unit for DO_REPOSITION

julianoes:
PX4 actually implements this in radians, and so does MAVSDK.

My hunch is that this was not fully tested for PX4.

Further discussion in:
https://github.com/mavlink/mavlink/pull/2271


julianoes:
@dakejahl and @DonLakeFlyer, yes it's not required because it's just meta information. That being said updating MAVLink won't hurt.

PR #12798, opened 2025-05-08
MissionManager: increment take photo sequence

julianoes:
For a big survey mission it is nice if the images have sequence numbers. Also, this can prevent double captures when a command is re-transmitted somwhere in the chain.

### DIFF
--- a/src/MissionManager/CameraSection.cc
+++ b/src/MissionManager/CameraSection.cc
@@ -31,6 +31,7 @@ CameraSection::CameraSection(PlanMasterController* masterController, QObject* pa
     , _cameraPhotoIntervalDistanceFact  (0, _cameraPhotoIntervalDistanceName,   FactMetaData::valueTypeDouble)
     , _cameraPhotoIntervalTimeFact      (0, _cameraPhotoIntervalTimeName,       FactMetaData::valueTypeUint32)
     , _cameraModeFact                   (0, _cameraModeName,                    FactMetaData::valueTypeUint32)
+    , _takePhotoSequence                (0)
     , _dirty                            (false)
 {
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Can you just double-check with the ardupilot folks that they won't puke on this being non-zero?

julianoes:
It doesn't look like they are using it:
https://github.com/ArduPilot/ardupilot/blob/d952f8f261e870b48779795e2ace13cf594a263f/libraries/AP_Mission/AP_Mission_Commands.cpp#L182-L204

@peterbarker do you know if my change here which is adding the seq number to the IMAGE_START_CAPTURE command breaks ArduPilot?

peterbarker:
> For a big survey mission it is nice if the images have sequence numbers. Also, this can prevent double captures when a command is re-transmitted somwhere in the chain.

The retransmission thing is a little problematic.  An autopilot is unlikely to remember a whole bunch of sequence numbers, most likely just the last one received and not take another picture based on that.  Similarly for a camera taking pictures.

But what I'm most confused about is how that's relevant - if you are repeating this message on-board your aircraft somehow... well, you've got issues :-)

As is pointed out, we don't use this value from the mission at the moment.  We do go to the trouble of storing/retrieving it from mission storage, but then just don't use it :-)  I guess we really should make an effort at least to pass it onto mavlink cameras...

This does make sense to me, but please understand that @rmackay9 looks after our camera library.

julianoes:
I agree there are edge cases with retransmission, so maybe that should not be the main argument.

In terms of how this is relevant: in the current setup I'm working with, these sequence numbers are plugged all the way through and so we end up with a camera that gets sequence 0 over and over which doesn't help when you try to check whether all images have been taken.

PR #12833, opened 2025-05-13
QML: Fix Viewer3D Closing

HTRamsey:
Viewer3DWindow is otherwise undefined and the close function doesn't get called


DonLakeFlyer:
When I run this I get a full blank white screen for fly view. No toolstrip or instruments. I'm guessing zorder problem.

PR #12836, opened 2025-05-13
Utilities: Add Logging Filter

HTRamsey:
Adds a custom filter function for the logging categories which can allow easier/finer control of the enabled categories.
Also could collect the registered categories here and remove the custom QGCLoggingCategory class.

Fixes a debug output issue too.

### DIFF
--- a/src/Utilities/QGCLogging.cc
+++ b/src/Utilities/QGCLogging.cc
@@ -18,6 +18,8 @@
 #include <QtCore/QStringListModel>
 #include <QtCore/QTextStream>
 
+QGC_LOGGING_CATEGORY(QGCLoggingLog, "qgc.utilities.qgclogging")
+
 #ifdef Q_OS_WIN
 
 #include <crtdbg.h>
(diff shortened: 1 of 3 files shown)


dakejahl:
On the topic of logging, can we move the log namespace as a prefix? 

**current**
```
82.241 - debug: UDP State Changed: QAbstractSocket::ClosingState (qgc.comms.udplink:UDPWorker::setupSocket()::<lambda(QAbstractSocket::SocketState)>:302)
```
**suggested**
```
qgc.comms.udplink:: 82.241 - debug: UDP State Changed: QAbstractSocket::ClosingState (UDPWorker::setupSocket()::<lambda(QAbstractSocket::SocketState)>:302)
```

HTRamsey:
Yup, it's just changing the order [here](https://github.com/mavlink/qgroundcontrol/blob/cd7128cc31e112b5b02cac347f6a228d261997b2/src/Utilities/QGCLogging.cc#L75) if you want to make a PR

PR #12848, opened 2025-05-14
Utilities: Add GDAL

HTRamsey:
Just a test


Ryanf55:
Compare it to my attempt which compiles if you want: https://github.com/mavlink/qgroundcontrol/compare/master...Ryanf55:qgroundcontrol:add-gdal-map-provider

PR #12860, opened 2025-05-16
QML: Fix Accidental Hiding of Main Window

### LINKED ISSUE #12858: QGC crashes after disconnecting USB [Release V5.0]
I can't add the tag - please mark this as `Release V5.0`.

**Clearly mention the version you're testing and your operating system.**
QGC https://github.com/mavlink/qgroundcontrol/commit/93ee76f13cb3054c5692751165e4b3fe704ade35
Windows 11 24H2

**Describe your hardware setup (telemetry, flight controller, firmware).**
Holybro Pixhawk 6X connected by USB cable.
Firmware: PX4 1.15.4

**Include detailed steps to reproduce the issue.**
After physically disconnecting the USB cable, while the connection in QGC is still open, I get an error message about access to the COM port.
Connection is still open (but with state `disconnected`), so I want to terminate it by clicking `Disconnect`. After clicking the button, the QGC window disappears, but the process continues working in the background. To launch QGC again, it needs to be manually closed via Task Manager as there can't be two instances running simultaneously. 

On the sidenote, it takes quite some time to launch every time.

**Add screenshots or screen recordings where possible — they really help!**

https://github.com/user-attachments/assets/286d0e7c-a8b9-45af-86fd-3c29c35a0f88

HTRamsey:
Fix #12858

### DIFF
--- a/src/FlightDisplay/FlyViewMap.qml
+++ b/src/FlightDisplay/FlyViewMap.qml
@@ -388,15 +388,6 @@ FlightMap {
             globals.guidedControllerFlyView.fwdFlightGotoMapCircle = this
         }
 
-        Connections {
-            target: QGroundControl.multiVehicleManager
-            function onActiveVehicleChanged(activeVehicle) {
-                if (!activeVehicle) {
-                    visible = false
-                }
-            }
-        }
-
         Binding {
             target: _fwdFlightGotoMapCircle
             property: "center"
@@ -476,7 +467,7 @@ FlightMap {
             target: QGroundControl.multiVehicleManager
             function onActiveVehicleChanged(activeVehicle) {
                 if (!activeVehicle) {
-                    hide()
+                    gotoLocationItem.hide()
                 }
             }
         }


HTRamsey:
@rubenp02 @DonLakeFlyer FYI

HTRamsey:
This one was tough because it wasn't what I was looking for at all. I'm not sure I actually understand how a Connections object changes the namespace to the mainwindow

DonLakeFlyer:
Hah, I was just hitting this debugging the usb pull crash. Wondering what the hell was going on!

DonLakeFlyer:
I deal with this in my usb disconnect fix

DonLakeFlyer:
For me, removing this:
```
       Connections {
            target: QGroundControl.multiVehicleManager
            function onActiveVehicleChanged(activeVehicle) {
                if (!activeVehicle) {
                    visible = false
                }
            }
        }
```

Isn't fixing the main window hide problem. No idea why...

HTRamsey:
@DonLakeFlyer that one hides the flyview or something, the one with gotoLocationItem.hide() was hiding the whole window

DonLakeFlyer:
@rubenp02 Not quite sure about whether both of the Connections blocks for null activeVehicle were added by you. If so, was there a reason why? Do you find a case which didn't work. There is already other code which tracks the _activeVehicle going away which works fine. I removed both blocks and tested that the goto location item goes away correctly. Did you see something different?

PR #12864, opened 2025-05-16
Fix crash on usb disconnect, fix hiding of main window

### LINKED ISSUE #12858: QGC crashes after disconnecting USB [Release V5.0]

DonLakeFlyer:
* Fix #12858
* Fixes crash on usb cable disconnect
* Fixes main window disappearing on active vehicle going away
* Fixes link errors showing message in title and title in message area
* Stops popping link error dialog on usb cable pull
* Replacement for #12860

### DIFF
--- a/src/Comms/LinkManager.cc
+++ b/src/Comms/LinkManager.cc
@@ -171,7 +171,7 @@ bool LinkManager::createConnectedLink(SharedLinkConfigurationPtr &config)
     _rgLinks.append(link);
     config->setLink(link);
 
-    (void) connect(link.get(), &LinkInterface::communicationError, qgcApp(), &QGCApplication::showAppMessage);
+    (void) connect(link.get(), &LinkInterface::communicationError, this, &LinkManager::_communicationError);
     (void) connect(link.get(), &LinkInterface::bytesReceived, MAVLinkProtocol::instance(), &MAVLinkProtocol::receiveBytes);
     (void) connect(link.get(), &LinkInterface::bytesSent, MAVLinkProtocol::instance(), &MAVLinkProtocol::logSentBytes);
     (void) connect(link.get(), &LinkInterface::disconnected, this, &LinkManager::_linkDisconnected);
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
@rubenp02 I could see why those two Connections blocks were needed. There was already activeVehicle checking in other places on the goto location item. I also tested the vehicle going away while a goto location was displayed and it correctly went away.

DonLakeFlyer:
> * Stops popping link error dialog on usb cable pull

@HTRamsey This didn't happen in pre-5.0 builds. It's kinda annoying. Not sure what you think about this change.

HTRamsey:
@DonLakeFlyer yeah that's fine. I kinda wish we had a separate notification message that just kinda pops up and fades out after a few seconds, particularly for things like this that the user should notice but doesn't require actual input to dismiss it.

HTRamsey:
I believe that change in FactValueGrid should fix these warnings too right? It'll be the offline vehicle instead of null now I think
image

DonLakeFlyer:
> I believe that change in FactValueGrid should fix these warnings too right?

The second one. But not GuidedActionsController

I'm going to merge this because without it the main window close thing makes testing really hard!

PR #12866, opened 2025-05-16
Threaded parameter search

DonLakeFlyer:
@HTRamsey Can you take a look at this? I'm not super great at threads and even worse with Qt and threads. I'm doing something wrong with respect to getting the created search list from one thread to the main thread. If I delete the previous parameter list is crashes with memory corruption.

Basically I created the new QmlObjectListMode on the thread. But then I need to get it over to the main thread so I can replace the search results with it.

Still missing code to cancel thread on controller destruct as well.

### DIFF
--- a/src/QmlControls/ParameterEditorController.cc
+++ b/src/QmlControls/ParameterEditorController.cc
@@ -24,14 +24,11 @@ ParameterEditorController::ParameterEditorController(QObject *parent)
 
     _buildLists();
 
-    _searchTimer.setSingleShot(true);
-    _searchTimer.setInterval(300);
-
     connect(this, &ParameterEditorController::currentCategoryChanged,   this, &ParameterEditorController::_currentCategoryChanged);
     connect(this, &ParameterEditorController::currentGroupChanged,      this, &ParameterEditorController::_currentGroupChanged);
     connect(this, &ParameterEditorController::searchTextChanged,        this, &ParameterEditorController::_searchTextChanged);
     connect(this, &ParameterEditorController::showModifiedOnlyChanged,  this, &ParameterEditorController::_searchTextChanged);
-    connect(&_searchTimer, &QTimer::timeout,                            this, &ParameterEditorController::_performSearch);
+    connect(this, &ParameterEditorController::searchResultsReady,       this, &ParameterEditorController::_swapSearchResults, Qt::QueuedConnection);
     connect(_parameterMgr, &ParameterManager::factAdded,                this, &ParameterEditorController::_factAdded);
 
     ParameterEditorCategory* category = _categories.count() ? _categories.value<ParameterEditorCategory*>(0) : nullptr;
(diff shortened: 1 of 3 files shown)


src/QmlControls/ParameterEditorController.cc
```diff
@@ -374,61 +371,88 @@ bool ParameterEditorController::_shouldShow(Fact* fact) const
...
+    emit searchResultsReady();
+}
 
-            _currentCategory    = nullptr;
-            _currentGroup       = nullptr;
-        }
+void ParameterEditorController::_swapSearchResults()
+{
+    if (_searchResults) {
+        //delete _parameters;
```
DonLakeFlyer, line 451: @HTRamsey If I comment this out it all works, but leaks. If I put it in it will crash with memory corruption.

HTRamsey: This might be one of those tasks that's easier to do through QtConcurrent, but yeah I'll take a look at it later.

DonLakeFlyer:
I figured out how to re-write so that I don't have to allocate anything on the thread side.

DonLakeFlyer:
It's all working now with the last set of changes. @HTRamsey Can you still review?

DonLakeFlyer:
@rubenp02 Can you give this a try?

HTRamsey:
If you are interested in how it performs for me, it definitely lags a bit before the changes but it's actually quite a bit worse after the threading.

Before: 

https://github.com/user-attachments/assets/07d54b8e-7092-42f7-9851-8e8277d3fcb3

After: 

https://github.com/user-attachments/assets/dcf2963b-814e-45a2-b04c-1f3df66a5c6b

HTRamsey:
That's weird that it was worse in release for that guy because debug disables QT_ENABLE_REGEXP_JIT which disables the optimization of the QRegularExpression so it should actually be slower in debug. In a few hours I'll test your PR against the base code using a QElapsedTimer right before the search to see if the mean time changes any

HTRamsey:
Gotcha, either way we'll find out. If this doesn't help much then I think I've found the best possible solution anyways so I'll implement it tonight and see what you guys think

DonLakeFlyer:
I also wonder about the regular expression. There are probably like two people in the world that know you can type regular expressions in there. Straight up contains is gonna be like 99.9% use case. I had no idea this was change the regex somewhere along the way.

DonLakeFlyer:
@rubenp02 Can you try this when build finished. This is basically a combination of everything:
* Timer for keystrokes
* Search on thread
* Fix regex stupidity

DonLakeFlyer:
This should theoretically be as good as it gets. If not, we'll just go with the other pull

dakejahl:
The regular expression can be really useful when you don't know the name of the parameter you're looking for. But has the side effect of returning way too many matches if you are not verbose. For context the PR that introduced the feature is here
https://github.com/mavlink/qgroundcontrol/pull/10479

DonLakeFlyer:
I guess maybe I could see people expecting a wildcard "*" to work.

HTRamsey:
> Out of curiosity, what's your idea?

There's a few optimizations to be made in c++ but that's not where the issue is. It's in QML, the best thing to do is to go back and fix the ListView, or at a minimum remove the 3 repeaters with just 1

HTRamsey:
Here are some results on Linux on an Intel NUC, duration of the _performSearch function is displayed in debug output:
1. Original:

https://github.com/user-attachments/assets/75462654-0ec2-425f-96cf-1e3642cfb93b

2. Switch to 1 repeater instead of 3

https://github.com/user-attachments/assets/77f42bcd-2a4e-45b8-988c-b75c11a5ad1e

3. Using a ListView instead of Repeaters

https://github.com/user-attachments/assets/5b0d5162-2f6d-4998-b7bb-98b6f8f51e09

I don't have a video of this PR but basically the threaded search function takes <10ms but starting the thread is very slow. So the visual results are slow if you are typing slower than the period of the _searchTimer (300ms, typical typing pace on desktop is about 1 character every 160ms but mobile is going to be a lot slower).

DonLakeFlyer:
I have a reworked TableVIew/QAbstractTableModel version I'm playing around with. We'll see...

PR #12867, opened 2025-05-16
APM: fix uavcan compass name

dakejahl:
`busType` is a string, so the existing logic has never worked. This bug resulted in the compass being reported as
```
HMC5883_OLD(UAVCAN0)
```
With the fixed logic it is now
```
UAVCAN(UAVCAN0)
```
The DroneCAN MagneticFieldStrength2 message only has a `uint8 id` which is used for sensor instance reporting. There is no mechanism to report device type of the magnetometer sensor (eg ii2smdc, hmc5883, etc)

### DIFF
--- a/src/FirmwarePlugin/APM/APMSensorIdDecoder.qml
+++ b/src/FirmwarePlugin/APM/APMSensorIdDecoder.qml
@@ -148,9 +148,9 @@ QGCLabel {
             return ""
         }
         if (deviceName.startsWith('COMPASS')) {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Hmm, never seen this code before. I assume if you tested then it should be all good now.

dakejahl:
Yup I tested. It's only a thing with ardupilot but shows you on the vehicle page which sensors are on the board. Kinda cool, I was thinking about adding something like it for PX4

Issue #12236, opened 2024-12-19
Very poor performance when moderately large missions are loaded


rubenp02:
## Current Behavior
When working with moderately large missions (of maybe around 250+ simple waypoints, probably more than the typical use case but still much less than what is currently supported in the main autopilot firmwares), QGroundControl becomes very unresponsive to the point of being unusable, particularly in the Plan View and in low end devices such as Android GCS. In the latter case it can and does crash at any moment, making it unsuitable for operations where the GCS is critical.

## Expected Behavior
Adequate performance and stability, at least up to 600 waypoints, which is what ArduPilot supports without an external SD card. Or at the very least the performance of the current QGC stable, where this issue isn't nearly as bad. The model size of even a huge mission is very manageable by any system, so this seems to be due to very innefficient rendering or unnecessary view updates.

## Steps to Reproduce:
1. Create and load a mission with a large number of simple waypoints
2. Pan the view, scroll down the waypoint list, etc.
3. The performance will be abysmal at this point. If using an Android GCS, the app will crash within a few minutes.

## System Information
- QGC Version: current master

## Sample mission
[test-250-simple-wp-mission.zip](https://github.com/user-attachments/files/18195955/test-250-simple-wp-mission.zip)

HTRamsey:
Are you able to run the QML Profiler? If not I will do so at some point. I'm not really aware of anything that's changed with any of the mission/plan view stuff so I'll have to look into it further

HTRamsey:
Haven't tried on Android yet but at least on a high end Linux it works fine.
![image](https://github.com/user-attachments/assets/b06cc88f-49b2-4fec-bda8-507f46db4143)

DonLakeFlyer:
Unless there is come specific control that has gone whacky there isn't much different in the Plan view Stable -> Daily.

DonLakeFlyer:
Tried OSX, don't see anything out of the ordinary

DonLakeFlyer:
Windows is working fine for me as well with the example plan from above

DonLakeFlyer:
Herelink seems fine as well. Scrolling the waypoint list up/down is a little sluggish but still workable.

DonLakeFlyer:
Still odd you would see it on multiple OS

DonLakeFlyer:
[...]

PR #12870, opened 2025-05-17
Map: Fix various zoomLevel problems

DonLakeFlyer:
Discovered while testing #12236

### DIFF
--- a/src/FlightMap/FlightMap.qml
+++ b/src/FlightMap/FlightMap.qml
@@ -12,6 +12,7 @@ import QtQuick.Controls
 import QtLocation
 import QtPositioning
 import QtQuick.Dialogs
+import Qt.labs.animation
 
 import QGroundControl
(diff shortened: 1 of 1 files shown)


HTRamsey:
I think 15x might not be quite enough, at least in my case we move around within a radius of like 15m. My original plan was to limit zoom to the QGeoCameraCapabilities.maximumZoomLevel

DonLakeFlyer:
The 15 is just when loading a plan file. You can still manually zoom in all the way to 20. Problem is not all providers go up that far.

DonLakeFlyer:
I'm going to look at plumbing the max zoom level up through the providers so that it's real/variable as opposed to hardcoded.

DonLakeFlyer:
Turns out tiles vary by location as far as zoom level availability. So the problem I was seeing is location specific. I'll change this to no go past the global max of 20.

PR #12875, opened 2025-05-18
CMake: Fix Overriding of CPM_SOURCE_CACHE

### LINKED ISSUE #12734: Figure out CPM_SOURCE_CACHE usage
As it currently stands CPM_SOURCE_CACHE environment variable doesn't work because CustomOptions.cmake is stomping over it. I tried remove the settings from CustomOptions.cmake but that didn't seem to really do the trick. Still looking...

HTRamsey:
Fix #12734

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,12 @@ endif()
 include(GNUInstallDirs)
 include(FetchContent)
 include(CMakePrintHelpers)
+
+if(EXISTS "${QGC_CPM_SOURCE_CACHE}")
+    set(ENV{CPM_SOURCE_CACHE} "${QGC_CPM_SOURCE_CACHE}")
+else()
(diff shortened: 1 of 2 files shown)


HTRamsey:
@DonLakeFlyer thoughts on this? I originally didn't accommodate for the fact that CPM_SOURCE_CACHE was supposed to be an Env variable instead of a CMake variable

DonLakeFlyer:
Why not just let the normal CPM stuff handle this and remove all the special stuff in CMakeLists and CustomOptions? If folks want to override they can just set the regular CPM_SOURCE_CACHE in their custom build. Is the default provided by CPM for this not good so we need to specify our own?

DonLakeFlyer:
Makes sense then

DonLakeFlyer (approved):

PR #12876, opened 2025-05-19
YAPSP: TableView/QAbstractTableModel based parameter list

DonLakeFlyer:
Yes Another Parameter Search Pull. TableView/QAbstractItemModel basically achieves the best of both worlds with respect to ListView caching and only created visible items as well as more control over column visuals. This is still a major WIP but there should be enough there to verify list perf.


DonLakeFlyer:
@rubenp02, @HTRamsey FYI so you guys can try the perf. 

Also my take on mobile is that it should not be typedown style search but have a real "Search" button to trigger a search.

HTRamsey (approved):

PR #12887, opened 2025-05-21
CMake: Implement qt_add_qml_module

HTRamsey:
It's a bit difficult to do the whole app at one time, so this is a start with the AnalyzeView classes being automatically registered. Everything will be able to be cleaned up nicely after custom build design is restructured after this

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -102,7 +102,7 @@ jobs:
 
       - name: Sanity check dev build excecutable
         if: matrix.BuildType == 'Release'
-        working-directory: ${{ runner.temp }}/shadow_build_dir/Release/QGroundControl.app/Contents/MacOS
+        working-directory: ${{ runner.temp }}/shadow_build_dir/QGroundControl.app/Contents/MacOS
         run: ./QGroundControl --simple-boot-test
 
       - name: Create DMG
@@ -118,7 +118,7 @@ jobs:
         if: matrix.BuildType == 'Release'
         working-directory: /Volumes/QGroundControl/QGroundControl.app/Contents/MacOS
         run: ./QGroundControl --simple-boot-test
-  
+
       - name: Upload Build File
         if: matrix.BuildType == 'Release'
         uses: ./.github/actions/upload
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -120,7 +120,7 @@ jobs:
 
       - name: Sanity check release excecutable
         if: matrix.BuildType == 'Release'
-        working-directory: ${{ runner.temp }}\shadow_build_dir\Release
+        working-directory: ${{ runner.temp }}\shadow_build_dir
         run: QGroundControl.exe --simple-boot-test
 
       - name: Install
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,4 @@ docs/.vitepress/dist/
 
 CMakeUserPresets.json
 cpm_modules
+**/.qmlls.ini
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,9 +145,9 @@ if(CMAKE_CROSSCOMPILING)
     endif()
 endif()
 
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<CONFIGURATION>/lib)
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<CONFIGURATION>)
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/$<CONFIGURATION>)
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib) # $<CONFIGURATION>/
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 
 # https://cmake.org/cmake/help/latest/policy/CMP0168.html#policy:CMP0168
 if(POLICY CMP0168)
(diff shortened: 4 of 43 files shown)


DonLakeFlyer:
Is it required that all the "Controller" stuff needs to be changed in order to support qml compiling? Seems pretty invasive.

HTRamsey:
No not really I just wanted to make sure the type compiling from cpp to qml would work too. The bare minimum to get it to work is probably to just move the ${QGC_RESOURCES} from the qt_add_executable to qt_add_qml_module QML_FILES.

DonLakeFlyer:
I'd rather have a solution which doesn't require modifications to the existing qml source. Which the controller stuff seems to require? It's so late in the game to be futzing in there.

HTRamsey:
No QML changes are necessary, except adding import QGC, which I did to put it in a different namespace than the existing QGroundControl namespace for now

DonLakeFlyer:
I guess I'm confused then. The C++ controllers are already in the QGroundControl.Controllers module. Why is a new one needed? Or are you talking about something else?

HTRamsey:
So you know how we have
`qmlRegisterType<GeoTagController>("QGroundControl.Controllers", 1, 0, "GeoTagController");`?
Adding QML_ELEMENT to the header of GeoTagController automatically creates
```
Component {
        file: "GeoTagController.h"
        name: "GeoTagController"
        accessSemantics: "reference"
        prototype: "QObject"
        exports: ["QGC/GeoTagController 254.0"]
        exportMetaObjectRevisions: [65024]
        Property {
            name: "logFile"
            type: "QString"
            read: "logFile"
            write: "setLogFile"
            notify: "logFileChanged"
            index: 0
        }
        Property {
            name: "imageDirectory"
            type: "QString"
            read: "imageDirectory"
            write: "setImageDirectory"
            notify: "imageDirectoryChanged"
            index: 1
        }
        Property {
            name: "saveDirectory"
            type: "QString"
            read: "saveDirectory"
            write: "setSaveDirectory"
            notify: "saveDirectoryChanged"
            index: 2
        }
        Property {
            name: "errorMessage"
            type: "QString"
            read: "errorMessage"
            notify: "errorMessageChanged"
            index: 3
            isReadonly: true
        }
        Property {
            name: "progress"
            type: "double"
            read: "progress"
            notify: "progressChanged"
            index: 4
            isReadonly: true
        }
        Property {
            name: "inProgress"
            type: "bool"
            read: "inProgress"
            notify: "inProgressChanged"
            index: 5
            isReadonly: true
        }
        Signal {
            name: "logFileChanged"
            Parameter { name: "logFile"; type: "QString" }
        }
        Signal {
            name: "imageDirectoryChanged"
            Parameter { name: "imageDirectory"; type: "QString" }
        }
        Signal {
            name: "saveDirectoryChanged"
            Parameter { name: "saveDirectory"; type: "QString" }
        }
        Signal {
            name: "progressChanged"
            Parameter { name: "progress"; type: "double" }
        }
        Signal { name: "inProgressChanged" }
        Signal {
            name: "errorMessageChanged"
            Parameter { name: "errorMessage"; type: "QString" }
        }
        Method {
            name: "_workerProgressChanged"
            Parameter { name: "progress"; type: "double" }
        }
        Method {
            name: "_setErrorMessage"
            Parameter { name: "errorMsg"; type: "QString" }
        }
        Method {
            name: "_workerError"
            Parameter { name: "errorMsg"; type: "QString" }
        }
        Method { name: "startTagging" }
        Method { name: "cancelTagging" }
    }
```
which then enables compiling when you add ENABLE_TYPE_COMPILER to qt_add_qml_module. This is a different kind of compiling than the qmlcachegen, and I don't intend on enabling it for everything right now I just wanted to make sure it was set up to work correctly. Enabling qmlcachegen (the caching that v4.4 did) just requires add the qml files in cmake to the qt_add_qml_module command instead of the qt_add_executable command.

HTRamsey:
And then a file is auto-created that has this:
```
Q_QMLTYPE_EXPORT void qml_register_types_QGC()
{
    qmlRegisterModule("QGC", 254, 0);
    QT_WARNING_PUSH QT_WARNING_DISABLE_DEPRECATED
    qmlRegisterTypesAndRevisions<GeoTagController>("QGC", 254);
    qmlRegisterTypesAndRevisions<LogDownloadController>("QGC", 254);
    qmlRegisterTypesAndRevisions<MAVLinkChartController>("QGC", 254);
    qmlRegisterTypesAndRevisions<MAVLinkConsoleController>("QGC", 254);
    qmlRegisterAnonymousType<QAbstractItemModel, 254>("QGC", 254);
    qmlRegisterTypesAndRevisions<MAVLinkInspectorController>("QGC", 254);
    qmlRegisterTypesAndRevisions<QGCLogEntry>("QGC", 254);
    qmlRegisterTypesAndRevisions<QGCMAVLinkMessage>("QGC", 254);
    qmlRegisterTypesAndRevisions<QGCMAVLinkMessageField>("QGC", 254);
    qmlRegisterTypesAndRevisions<QGCMAVLinkSystem>("QGC", 254);
    QMetaType::fromType<QStringListModel *>().id();
    QT_WARNING_POP
    qmlRegisterModule("QGC", 254, 254);
}
```

HTRamsey:
so it enables auto registering of types and type compiling, neither of which v4.4 had

DonLakeFlyer:
So I totally missed that fact that Qt has two types of QML "compiling". One compiles to bytecode and the other compiles to C++.  I think I now understand that previous stable basically compiled to bytecode only using qmlcachegen. Now I get what this new stuff does.

My opinion is that it's too late in the game to take any change beyond bytecode compiling for 5.0. We theoretically know what worked in previous QGC versions. I'm concerned if we go any further that we can totally change the code path to use something new which has never been used before in QGC a few days before we plan to cut a new stable. Seems majorly dangerous to me. For example if we are not compiling to C++ then the current way controller code is built and exposed to Qml stays the same. So I don't get why we would switch to using QML_ELEMENT right now?

HTRamsey:
That's why I said I didn't intend to switch everything over right now. But I needed to make sure the way we set up the module will actually work for type compiling, so I tested it on the AnalyzeView classes. Otherwise you risk setting something up that has to be completely changed later

DonLakeFlyer:
Ok, understood. Thanks for taking the time to explain it all to me.

PR #12894, opened 2025-05-21
Only add true mavlink port from composite device

### LINKED ISSUE #12893: Deal with multiple composite ports on same device
ArduPilot with a CubeOrange will expose two ports over a single usb connection. Only one of these is the mavlink connection. In the past, Stable 4.4 timeframe at least on Mac the first port was always the mavlink one. Hence Stable 4.4 would disregard the second port from the same device. That fix no longer works at least in OSX where it's the second port listed which is the mavlink connection.

Current 5.0 no longer does the secondary port connections and will try to connect to both ports. This works fine on Mac, but throws device errors on Linux on the non-mavlink port.

I thinking we are going to have to keep track of the composite ports associated with the same device. Connect to both, but don't show errors until we attempt the multiple connects. And if one of them connects we are good and eat the error on the other one. If both connect without errors and only one has mavlink then that is fine as well. Any maybe we then disconnect the other one.

DonLakeFlyer:
* This code assumes the mavlink port name is the lower numbered one (Windows: `COM5/COM6`, Linux: `ttyACM0/ttyACM1`, Mac: `cu.usbmodemB41101/cu.usbmodemB41103`) if there is a composite usb device. Only that port is added to the list of available ports.
* Fix #12893

### DIFF
--- a/src/Comms/QGCSerialPortInfo.cc
+++ b/src/Comms/QGCSerialPortInfo.cc
@@ -17,6 +17,9 @@
 #include <QtCore/QJsonDocument>
 #include <QtCore/QJsonObject>
 
+
+#include <tuple>
+
 QGC_LOGGING_CATEGORY(QGCSerialPortInfoLog, "qgc.comms.qgcserialportinfo")
 
 bool QGCSerialPortInfo::_jsonLoaded = false;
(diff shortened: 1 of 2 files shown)


HTRamsey:
I'm not a fan of making assumptions about composite devices like this. We have an exception for NMEA but there's an unknown amount of arbitrary devices that don't conform to this same pattern of first port is the mavlink port and we can ignore all the rest. It tends to be correct for popular devices like the Cubes but won't work for everything

DonLakeFlyer:
I'd rather get it working for the majority use case than have it break in the majority use case. Happy to take other suggestions as to how to solve this. But I don't think making CubePilot stuff not autoconnect is a good idea. Also if this works I can modify so that manual connect gets the full list. Wanted to understand whether this works or now across OS.

Who's in charge of setting up the usb ports? Is that a hardware determined thing or does the firmware set that up?

DonLakeFlyer:
@dakejahl Can you try this on Linux?

dakejahl:
> @dakejahl Can you try this on Linux?

It's still intermittent. Sometimes it connects, sometimes it doesn't and there's no message, and sometimes it doesn't and there's an error message
image

DonLakeFlyer:
Ok, thanks. I have a feeling the maavlink/slcan port associations are moving around which means this approach no longer work. Gonna have to fix this the hard PITA way unless I can figure out how to get that interface descriptor information.

DonLakeFlyer:
@dakejahl When you run QGC 4.4 against this same setup do you get consistent results?

dakejahl:
It's still janky with v4.4 but does always eventually connect. Unplugging and replugging the cube it will never reconnect though. 

I asked in Ardupilot discord:
https://discord.com/channels/674039678562861068/719662818898739260/1374893983712411778
> Peter Barker — 5/21/25, 4:23 PM
One thing to note is that very few of our boards continue to do SLCAN on that second interface.  We heavily discourage it!

But I'm not sure if they're discouraging SLCAN or composite ports in general.

dakejahl:
> @dakejahl What channel did you ask on in ArduPilot discord? That link isn't working for me.

Ahh whoops yeah it's a private channel, the ardupilot partners channel. I can take a stab at this next week too if you need some help.

DonLakeFlyer:
Ok, no problem.  I'm working out some other way to fix this.

PR #12901, opened 2025-05-23
Plan Mode: Improve Corridor Scan `Rotate Entry Point`

mrschick:
Description
-----------
1. ~~Tweaks the survey/scan presets so that they will auto-select the generated complex waypoint first. This cuts down on clicks in most situations where tweaking the takeoff waypoint's position or altitude is not a priority, i.e: because it's already set to the connected drone's position and already uses an acceptable altitude.~~ Will be revised and moved to a different PR
2. Tweaks the "Rotate Entry Point" logic of the Corridor Scan waypoint to be more user friendly, since before for scans narrow enough to cover with a single pass, the button would have to be clicked twice to get the expected result of inverting the flight's direction over the defined corridor.

Test Steps
-----------
1. ~~Under Plan Mode > File > Create Plan, try selecting any of the survey/scan presets under different conditions;~~
2. Create a Corridor Scan and tweak its waypoints and "corridor width" to test the behavior of the "Rotate Entry Point" button under different conditions;

Checklist:
----------

Related Issue
-----------

None

### DIFF
--- a/src/MissionManager/CorridorScanComplexItem.cc
+++ b/src/MissionManager/CorridorScanComplexItem.cc
@@ -178,7 +178,13 @@ void CorridorScanComplexItem::_polylineDirtyChanged(bool dirty)
 
 void CorridorScanComplexItem::rotateEntryPoint(void)
 {
-    _entryPoint++;
+    if (_calcTransectCount() < 2) {
+        // Cycle by 2 indexes when the path only has a single "lane", to prevent having to double click the "rotate entry point" button to invert the scan's direction along the path
+        _entryPoint += 2;
+    } else {
+        _entryPoint++;
+    }
+
     if (_entryPoint > 3) {
         _entryPoint = 0;
     }


DonLakeFlyer:
So I like the fix for 2. 

But I don't like the change for 1:
* This only sort of works if you are connected to a multi-rotor and it is already in its correct takeoff location. That doesn't seem like a majority case to me.
* The concept of the takeoff item not possibly needing tweaking also only works for multi-rotor take off. Fixed wing must set a climb out position.
* Given the way the ui works even if you don't change anything on takeoff you still need to go back and click done.
* Also the way the mavlink takeoff command works differs from PX4 and ArduPilot. PX4: the coordinate matters, ArduPilot: there is not coordinate it just goes straight up
* If you are creating a Plan offline or where the vehicle home position is not currently correct you need to set the correct position for Launch/Takeoff since it will now be known

This could theoretically be cleaned up for say multi-rotor and ardupilot and a known home position but that's all to complicated to deal with.

mrschick:
Dunno, the way I see it is that it makes planning of already connected multicopters quicker, since the takeoff item is auto-set and one doesn't need to touch it. While in any other mode it also doesn't have a negative impact, since one would have to set either item anyway, and the order in which they do doesn't impact the time spent.

DonLakeFlyer:
> since the takeoff item is auto-set and one doesn't need to touch it.

It's not though. You still need to click "Done" on the Takeoff item.

mrschick:
Odd.
While testing this on my Linux machine and emulating gazebo, creating the plan with the Multicopter already connected would auto-set the takeoff item to the drone's position, without having to press "done".

DonLakeFlyer:
Hmm

> While testing this on my Linux machine and emulating gazebo, creating the plan with the Multicopter already connected would auto-set the takeoff item to the drone's position, without having to press "done".

Let me check again. I may have created the plan with SITL before the home position became available.

DonLakeFlyer:
> I may have created the plan with SITL before the home position became available.

Yeah, that was it. You still run into the problem if you are planning offline though. Because then you need to set the takeoff item position.

The better way to make this work for complex items is to walk the list after everything is set up and fine the first "not done" item and select that. Instead of hardcoding to the polygon. That will work in all cases.

mrschick:
True, that would be even better 👌🏻

I did all of this for an internship that has since finished, there is a possibility that I will be employed longer-term in the near future and can improve this further.
Until then, I could revert the first change in the current PR and eventually make a new PR for its improved version.

DonLakeFlyer:
> Until then, I could revert the first change in the current PR and eventually make a new PR for its improved version.

Yeah, that would be great

DonLakeFlyer:
@mrschick As I was code reviewing this I realized the usage of _entryPoint and the magic values where clear as mud when you go back and look at it years later. I took your changes and cleaned up to code to be more easily understood in a new pull.

DonLakeFlyer:
Replace by my pull

PR #12911, opened 2025-05-24
Comms: Fix Mavlink Loss Percent

HTRamsey:


HTRamsey:
Originally there was a weird mix of logic where it seemed like they were tracking sysid/compid in one place but then use channels for the same variables in another place. Not 100% sure this completely resolves that but it is a step in that direction

HTRamsey:
There's also the issue of how UDP packets can come in out of order, so probably I'll come back and take a deeper look at this again soon

PR #12912, opened 2025-05-24
CI: Fix Android Windows Signing

HTRamsey:
Attempting to fix windows reading keystore


HTRamsey:
If this doesn't work we'll just temporarily disable it again since it's mainly meant as just a build check, and signing works locally.

DonLakeFlyer:
So that fixed it? The binary thing?

PR #12914, opened 2025-05-24
Fix Delete All Settings Option

HTRamsey:
You could enable it but then you couldn't disable it
also make use of QVersionNumber for simpler version comparison


DonLakeFlyer:
Can you hold the cleanup oriented stuff till after we branch and just keep bug fixes only? All of this is good but it's easy to screw things up by mistake by rearranging the deck chairs on things that aren't broken. To late in the game on 5.0 for that sort of stuff.

PR #12921, opened 2025-05-25
Autoconnect: Filter out secondary composite ports from autoconnect

DonLakeFlyer:
* Fix for #12893
* These changes should bring the basically code back to how 4.4 worked
* The secondary ports from a composite device are not made available to autoconnect
* The list of ports available to manual connections is the full un-filtered list. This differs from 4.4 which filtered the manual connect list as well.
* Fixed SerialLink to not pop PermissionError errors on autoconnect links

So far I've tested Mac and Windows for correct behavior on:
* Eating PermissionError's
* USB cable pull and reconnect. Note: On Windows, same as before, usb cable pull doesn't automatically close link,. You need to disconnect manually.
* Autoconnect to a composite device like Cube Orange running Ardupilot works

This can still cause failures on autoconnect to composite devices line Cube Orange. But it is basically the same failures that occurred with 4.4. Linux seems to be the OS which causes the most grief. In those cases you will need to turn off Autoconnect and manual connect to the right port.

### DIFF
--- a/src/Comms/LinkManager.cc
+++ b/src/Comms/LinkManager.cc
@@ -785,6 +785,32 @@ void LinkManager::resetMavlinkSigning()
     }
 }
 
+void LinkManager::_filterCompositePorts(QList<QGCSerialPortInfo> &portList)
+{
+    typedef QPair<quint16, quint16> VidPidPair_t;
+
+    QList<QGCSerialPortInfo>        list;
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
@HTRamsey Can you give Linux and Android a try with this?

HTRamsey:
Linux is good, will try Android tonight

PR #12926, opened 2025-05-26
FactControl/FlightMap: Convert to qt_add_qml_module

DonLakeFlyer:
This format seems to work nicely. Kinda simple.


DonLakeFlyer:
@HTRamsey So at least for me, doing it this way was easy and it worked. Have to be super careful with testing and qml files loading from resource in loaders though.

PR #12928, opened 2025-05-27
Vehicle: Report armed status from HIGH_LATENCY2 messages for ArduPilot

stephendade:
In High Latency mode (https://mavlink.io/en/services/high_latency.html), ArduPilot uses the ``custom0`` field of the ``HIGH_LATENCY2`` message for the base_mode (arming status). See https://github.com/ArduPilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS_Common.cpp#L7407

This patch allows QGC to display the correct arming state of the vehicle, based on the above field.

Tested in ArduPilot SITL.


DonLakeFlyer:
Looks good. Thanks.

PR #12936, opened 2025-05-28
Windows: Remove UAVDrivers

HTRamsey:
In the event that we determine the windows UAV drivers are not needed
Also remove the 32bit check from the nsis file


HTRamsey:
@DonLakeFlyer From what I can tell, this is still needed

PR #12957, opened 2025-06-03
CMake: Fix QML module conversion for Qt 6.6.3

rubenp02:
# Fix QML module conversion for Qt 6.6.3

Description
-----------
Fixed issues with the runtime discovery of modules defined with qt_add_qml_module in 6.6.3, which were causing the app to crash on startup.

Key changes:
- Explicitly linked all QML modules defined with qt_add_qml_module.
- Added the NO_PLUGIN option to all qt_add_qml_module calls, as unlike in Qt 6.8.X, it is always needed in 6.6.3 for this style of module.

Note: The changes above should be reverted once Qt 6.6.3 support is dropped.

In addition to this, the plugin workaround used in some of these modules has been replaced with the NO_PLUGIN option in qt_add_qml_module, which is the supported way to disable plugin generation for static QML modules.

Checklist:
----------

### DIFF
--- a/src/AutoPilotPlugins/APM/CMakeLists.txt
+++ b/src/AutoPilotPlugins/APM/CMakeLists.txt
@@ -88,4 +88,5 @@ qt_add_qml_module(AutoPilotPluginsAPMModule
         APMSubMotorComponent.qml
         APMTuningComponentCopter.qml
         APMTuningComponentSub.qml
+    NO_PLUGIN # Required by Qt 6.6.3. TODO: Remove when no longer supported.
 )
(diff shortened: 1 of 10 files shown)


DonLakeFlyer:
Ok, so then #12974 isn't needed right?

DonLakeFlyer:
@rubenp02 Can you create an Issue which details what should be changed back when we bail on 6.6.3 support and assign to me?

DonLakeFlyer:
Changed look good.

rubenp02:
> @rubenp02 Can you create an Issue which details what should be changed back when we bail on 6.6.3 support and assign to me?

Done at #12977. I have no permissions to assign issues though.

PR #12974, opened 2025-06-06
Add target_link_libraries for all Qml Modules

### LINKED ISSUE #12942: Add target_link_libraries entry for all qml module
Right now only some of them are linked in manually.

DonLakeFlyer:
Fix #12942


DonLakeFlyer:
I believe this is not needed given #12957

PR #12991, opened 2025-06-09
QmlControls: add getter for resetActive in QmlObjectListModel

HTRamsey:
Use the list model resetActive getter rather than a local resetActive variable

### DIFF
--- a/src/QmlControls/QGCMapPolygon.cc
+++ b/src/QmlControls/QGCMapPolygon.cc
@@ -23,7 +23,6 @@ QGCMapPolygon::QGCMapPolygon(QObject* parent)
     , _centerDrag           (false)
     , _ignoreCenterUpdates  (false)
     , _interactive          (false)
-    , _resetActive          (false)
 {
     _init();
 }
@@ -34,7 +33,6 @@ QGCMapPolygon::QGCMapPolygon(const QGCMapPolygon& other, QObject* parent)
     , _centerDrag           (false)
     , _ignoreCenterUpdates  (false)
     , _interactive          (false)
-    , _resetActive          (false)
 {
     *this = other;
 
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
Just trying to wrap my head around this again. Making notes to remember...

This example is exactly the case of preventing double reset stacking:
```
    /// Reset polygon back to initial default
    function _resetPolygon() {
        mapPolygon.beginReset()
        mapPolygon.clear()
        mapPolygon.appendVertices(defaultPolygonVertices())
        mapPolygon.endReset()
        _circleMode = false
    }
```

Where the mapPolygon.clear() also wants to do begin/endReset internal to that call. But not if it was already done externally.

DonLakeFlyer:
And then a case which stacks up resets two times:
```
void QGCMapPolygon::offset(double distance)
{
    ...

    // Update internals
    _beginResetIfNotActive();
    clear();
    appendVertices(rgNewPolygon);
    _endResetIfNotActive();
}
```

The clear tries to begin/end as does the appendVertices.

HTRamsey:
It looks like it's okay to beginReset multiple times but not endReset.
image

DonLakeFlyer:
So from staring at this more I believe the reason that Polygon for example tracked it's own resetActive state was because it can nest two times within it's own code. For example new code in polygon offset method:
```
    _beginResetIfNotActive();
    clear();
    appendVertices(rgNewPolygon);
    _endResetIfActive();
```

In this case the first call to _beginResetIfNotActive will set it. But then internal to appendVertices there is also an _endResetIfActive which will turn it off prematurely. Am I reading that right?

To truly make this bulletproof to stacking of begin/resets I think QmlObjectListModel needs to basically reference count the begin/end calls. Only doing a real begin on the first one, and a real end on the last one to pop off.

That said I could easily be wrong. This whole thing makes my head hurt.

DonLakeFlyer:
If you're tired of futzing with this I can make a pull with ref counting thingy so it will work in all cases. Let me know.

DonLakeFlyer:
This is replaced by my new pull

PR #12994, opened 2025-06-09
CMake: Use NO_PLUGIN for CustomModule

HTRamsey:
@rubenp02 is this all I have to do?


rubenp02:
Yes, and I think it should also be possible to remove [these lines](https://github.com/mavlink/qgroundcontrol/blob/855b41e90f281c7d6e3b74e8b428fb0e5f67a133/src/CMakeLists.txt#L64-L69) now, this is much cleaner.

PR #13000, opened 2025-06-10
Fix nesting of begin/endResetModel

DonLakeFlyer:
Only outermost nested calls to begin/end pairs perform the actual reset

Replacement for #12991

### DIFF
--- a/src/QmlControls/QmlObjectListModel.h
+++ b/src/QmlControls/QmlObjectListModel.h
@@ -60,8 +60,10 @@ public:
     /// Clears the list and calls deleteLater on each entry
     void clearAndDeleteContents     ();
 
-    void beginReset                 ();
-    void endReset                   ();
+    /// These methods handling nesting a begin/end pairs. Such that only the outermost
+    /// beginResetModel/endResetModel pair will emit modelReset.
+    void beginResetModel            ();
+    void endResetModel              ();
 
 signals:
     void countChanged               (int count);
@@ -84,7 +86,7 @@ private:
     
     bool _dirty;
     bool _skipDirtyFirstItem;
-    bool _externalBeginResetModel;
+    uint _resetModelNestingCount = 0;
         
     static constexpr int ObjectRole = Qt::UserRole;
     static constexpr int TextRole = Qt::UserRole + 1;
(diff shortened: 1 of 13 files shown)


src/QmlControls/QmlObjectListModel.h
```diff
@@ -60,8 +60,10 @@ class QmlObjectListModel : public QAbstractListModel
     /// Clears the list and calls deleteLater on each entry
     void clearAndDeleteContents     ();
 
-    void beginReset                 ();
-    void endReset                   ();
+    /// These methods handling nesting a begin/end pairs. Such that only the outermost
+    /// beginResetModel/endResetModel pair will emit modelReset.
+    void beginResetModel            ();
```
HTRamsey, line 68: These shadow the methods from QAbstractItemModel

DonLakeFlyer: Yeah, I did that on purpose since I was expecting folks who understand QAbstractItemModel to use those methods names. Also we don't want people directly calling the QAbstractItemModel versions of begin/end since that would break the system.

Happy to take better ideas on how to do this. Or I could also add more commenting about it.

DonLakeFlyer:
I'd like to merge this so I can make a new RC

HTRamsey:
Only other comments I have would be to use qCDebug/qCWarning instead of qDebug/qWarning, and maybe to add that same resetActive() getter from my PR which I think could be useful (or maybe not, just an idea)

DonLakeFlyer:
Right, I'll fix up the qWarnings which are still left in there. The qDebug's are all commented out. I left them in as a debugging aid. But they are too noisy to leave in all the time.

DonLakeFlyer:
> But they are too noisy to leave in all the time.

Or maybe not. QmlObjectListModel doesn't have logging category stuff yet. So I can add a category. And make the qDebugs use it since it won't add noise to other non-existent logging.

DonLakeFlyer:
> and maybe to add that same resetActive() getter from my PR which I think could be useful (or maybe not, just an idea)

I'd rather not have people poking around behind the scenes on this. If it doesn't just work without understanding what is going on something is wrong.

HTRamsey:
Also keep in mind you shouldn't really need stuff like `QAbstractListModel::beginResetModel" << Q_FUNC_INFO` any more because it should be included in all messages due to [this log format](https://github.com/mavlink/qgroundcontrol/blob/7e44af13f7f326b17890805e7ebc9c3841195a3f/src/Utilities/QGCLogging.cc#L75)

DonLakeFlyer:
Ah, ok

PR #13009, opened 2025-06-11
GNSS Resilience with Septentrio

Louis-max-H:
Add support for GNSS receiver resilience information

Depend on this PX4-Autopilot PR : https://github.com/PX4/PX4-Autopilot/pull/25012
Original PR : https://github.com/mavlink/qgroundcontrol/pull/11781

Description
-----------
This PR is mostly taken from @chiara-septentrio and @flyingthingsintothings PR.
The code is mostly the same, I have improved a little code structure and move some files to update code structure 

If you want to have resilience information, you should use this PR of PX4 : https://github.com/PX4/PX4-Autopilot/pull/25012
And set `CONFIG_MAVLINK_DIALECT="development"` to your `default.px4board`

Test Steps
-----------
I have used a Septentrio emulator that emulate Jamming, Spoofing and Authentication attacks which was impossible to test before.
The functionality run smoothly in all cases, tell me if you need a video of demonstration or something's else to merge the PR :smiley:  

Checklist:
----------
- [x ] [Review Contribution Guidelines](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CONTRIBUTING.md).
- [ x] [Review Code of Conduct](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CODE_OF_CONDUCT.md).
- [x ] I have tested my changes.

Related Issue
-----------

### DIFF
--- a/src/Vehicle/FactGroups/VehicleGPSFactGroup.cc
+++ b/src/Vehicle/FactGroups/VehicleGPSFactGroup.cc
@@ -35,21 +35,6 @@ VehicleGPSFactGroup::VehicleGPSFactGroup(QObject *parent)
     _addFact(&_gnssSignalQualityFact);
     _addFact(&_postProcessingQualityFact);
 
-    _spoofingStateFact.metaData()->setEnumInfo(
-        QStringList() << "Disconnected" << "OK" << "Mitigated" << "Ongoing",
-        QVariantList() << 0 << 1 << 2 << 3
-    );
-
-    _jammingStateFact.metaData()->setEnumInfo(
-        QStringList() << "Disconnected" << "OK" << "Mitigated" << "Ongoing",
-        QVariantList() << 0 << 1 << 2 << 3
-    );
-
-    _authenticationStateFact.metaData()->setEnumInfo(
-        QStringList() << "Unknown" << "Initializing..." << "Failed" << "OK" << "Disabled",
-        QVariantList() << 0 << 1 << 2 << 3 << 4
-    );
-
     _latFact.setRawValue(std::numeric_limits<float>::quiet_NaN());
     _lonFact.setRawValue(std::numeric_limits<float>::quiet_NaN());
     _mgrsFact.setRawValue("");
--- a/src/QmlControls/GPSIndicatorPage.qml
+++ b/src/QmlControls/GPSIndicatorPage.qml
@@ -65,24 +65,28 @@ ToolIndicatorPage {
 
 
     function errorText() {
-        if(!_activeVehicle){
-            return qsTr("Disconnected")
-        } else if (_activeVehicle.gps.systemErrors.value === 1) {
-            return qsTr("Incoming correction")
-        } else if (_activeVehicle.gps.systemErrors.value === 2) {
-            return qsTr("Configuration")
-        } else if (_activeVehicle.gps.systemErrors.value === 4) {
-            return qsTr("Software")
-        } else if (_activeVehicle.gps.systemErrors.value === 8) {
-            return qsTr("Antenna")
-        } else if (_activeVehicle.gps.systemErrors.value === 16) {
-            return qsTr("Event congestion")
-        } else if (_activeVehicle.gps.systemErrors.value === 32) {
-            return qsTr("CPU overload")
-        } else if (_activeVehicle.gps.systemErrors.value === 64) {
-            return qsTr("Output congestion")
+        if (!_activeVehicle) {
+            return qsTr("Disconnected");
         }
-        return qsTr("Multiple errors")
+
+        switch (_activeVehicle.gps.systemErrors.value) {
+            case 1:
+                return qsTr("Incoming correction");
+            case 2:
+                return qsTr("Configuration");
+            case 4:
+                return qsTr("Software");
+            case 8:
+                return qsTr("Antenna");
+            case 16:
+                return qsTr("Event congestion");
+            case 32:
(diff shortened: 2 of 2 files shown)


HTRamsey:
Let me know when that PX4 PR gets merged

Tory9:
@HTRamsey, hello. The PX4 side was merged this week. Can this pr be reviewed again ?

Tory9:
@HTRamsey thank you for the review, I have applied the suggested changes

HTRamsey:
Looks good to me but I'll get Don to double check. Only comment I have (and it's not required but would be nice) would be to use an enum and switch in the QML files kinda similar to how this is done in RemoteIDSettings.h/RemoteIDSettings.qml

HTRamsey:
Actually can't you just return the enum string from the fact in qml for spoofingState, jammingState, and authenticationState?

Tory9:
Thank you, I applied the suggestion. Let me know if it looks better now. I have limited access to hardware to test these new changes, but if you confirm the code is fine, I’ll test the final version on hardware and let you know that everything works as expected

src/Vehicle/FactGroups/VehicleGPSFactGroup.cc
```diff
@@ -25,13 +26,43 @@ VehicleGPSFactGroup::VehicleGPSFactGroup(QObject *parent)
...
+    _addFact(&_systemErrorsFact);
+    _addFact(&_spoofingStateFact);
+    _addFact(&_jammingStateFact);
+    _addFact(&_authenticationStateFact);
+    _addFact(&_correctionsQualityFact);
+    _addFact(&_systemQualityFact);
+    _addFact(&_gnssSignalQualityFact);
+    _addFact(&_postProcessingQualityFact);
+
+    _spoofingStateFact.metaData()->setEnumInfo(
```
DonLakeFlyer: Why is this metadata being set manually given the fact that it is in the json already?

src/Comms/MockLink/MockLink.cc
```diff
@@ -1357,23 +1357,22 @@ void MockLink::_sendGpsRawInt()
         _vehicleComponentId,
         mavlinkChannel(),
         &msg,
-        timeTick++,                             // time since boot
+        timeTick++,                           // time since boot
```
DonLakeFlyer, line 1360: Why the changes to MockLink?

DonLakeFlyer:
@Tory9 Not really interested in adding two more slots into the toolbar for GPS. That's a lot of real estate. Can you provide screenshots of what this looks like with all GPS stuff in the toolbar explaining what it all means. Including the UI changes to the existing gps indicator. Then I'll think about how to organize UI.

Tory9:
@DonLakeFlyer thanks for the review, I get your point. I’ve attached a short PDF explaining the current approach. I would appreciate your suggestions for a more acceptable alternative
[QGC resilience feature explained.pdf](https://github.com/user-attachments/files/22866388/QGC.resilience.feature.explained.pdf)

DonLakeFlyer:
Perfect @Tory9. Give me a bit to think about this.

DonLakeFlyer:
So from looking at the mavlink spec for GNSS_IDENTITY it seems like that messages is not a "fused" message like the other GPS messages which combine multiple GPS into one. It seems like in the cases of multiple GNSS on a vehicle you will get multiple GNSS_IDENTITY telemetry each with a unique id. If so, then stuff these values into the the GPS fact group (which are fused values) seems wrong. Each unique GNSS telemetry is going to stomp over each other in the one gps fact group. Am I reading that wrong?

Tory9:
You’re right, if there are multiple receivers supporting this message, each will indeed send its own separate GNSS_IDENTITY telemetry. I also find it a bit confusing how potentially different resilience data from multiple GPS instances could be interpreted as a single

DonLakeFlyer:
So what I would do is fuse the telemetry from multiple GNSS together on the QGC side. Keep an array of GNSS_INTEGRITY messages by id in the gps fact group. Update the array when the data comes in. Then walk the array taking the "worst" value across all values. Stuff that "worst" value in the Fact if it differs from current value. 

My other question would be from a vehicle side standpoint how does the firmware use this information to make decisions? The idea would be to replicate how the vehicle side deals with multiple sets of this information.

Tory9:
For now, the firmware doesn’t really use this information for decision-making. It’s mainly passed to the GCS to provide environmental awareness and because it can be useful or interesting for the operator. I suppose this data could eventually be used to adjust the reliability weighting during sensor fusion, but that part isn’t implemented yet

DonLakeFlyer:
Ok, so then I would do the QGC side fuse thing. 

I'm also waiting to hear back from the PX4 Firmware folks about how they handle Mavlink Development dialect stuff. In QGC we've put this stuff behind an #ifdef keeping it only in debug builds. Given the fact that development dialect message can change out from under you. Not sure if we should change that or not.

Tory9:
Got it, thanks! Yeah, with this message still being in development there will definitely be some complications. I believe a PR to move it into the common dialect should be opened soon. The next release isn’t coming soon, so hopefully it can make it into common before then

HTRamsey:
@Louis-max-H what's the status of this?

Tory9:
@HTRamsey hi! Since multiple changes were requested, I decided to re-implement this feature in a new PR. I just need to do some final testing next week before opening it, but otherwise it’s already complete. I’ll open the new PR next week

HTRamsey:
Superceded

Issue #12972, opened 2025-06-06
Add Crowdin validation to enumStrings/Values comma separator


DonLakeFlyer:
The translations for the comma separator can be problematic. See #12966

DonLakeFlyer:
Validation scripts are only supported in enterprise addition. Instead I added translatorcomment entries to enumString values to hopefully guide the translator in the right direction.

PR #13013, opened 2025-06-11
Add translator comment for enumString translation to .ts files

DonLakeFlyer:
Trying to prevent conversion of string lists using comma separator to translated comma


DonLakeFlyer:
Related to #12972

PR #13019, opened 2025-06-13
Add yaw in GPS fact

Louis-max-H:
Add yaw in GPS fact and displayed it under GPS Status menu

Description
-----------
When using GPS heading along Gyroscope, QGrondControl will only show gyroscope compass, user can't know easily if heading is working correctly or what is his value.
I have added yaw to gps fact and make it displayed under Vehicle GPS Status menu.

The only way before was to open MAVLink inspector and check the value of yaw.
0 will be displayed if GPS do not provide heading and N/A when heading is not available.

image

Test Steps
-----------
Use GPS heading, 

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/src/Vehicle/FactGroups/VehicleGPSFactGroup.cc
+++ b/src/Vehicle/FactGroups/VehicleGPSFactGroup.cc
@@ -22,6 +22,7 @@ VehicleGPSFactGroup::VehicleGPSFactGroup(QObject *parent)
     _addFact(&_hdopFact);
     _addFact(&_vdopFact);
     _addFact(&_courseOverGroundFact);
+    _addFact(&_yawFact);
     _addFact(&_lockFact);
     _addFact(&_countFact);
 
@@ -31,6 +32,7 @@ VehicleGPSFactGroup::VehicleGPSFactGroup(QObject *parent)
     _hdopFact.setRawValue(std::numeric_limits<float>::quiet_NaN());
     _vdopFact.setRawValue(std::numeric_limits<float>::quiet_NaN());
     _courseOverGroundFact.setRawValue(std::numeric_limits<float>::quiet_NaN());
+    _yawFact.setRawValue(std::numeric_limits<int16_t>::quiet_NaN());
 }
 
 void VehicleGPSFactGroup::handleMessage(Vehicle *vehicle, const mavlink_message_t &message)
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
The toolbar indicator dropdown isn't meant to include stuff like this. It's for general status. This is a pretty low level detail. With the yaw adding to the fact group you can just include the values you want in the instrument panel to view.

Louis-max-H:
Thanks for the review !
I didn't know it was possible to display information like that, I have updated the PR to keep only yaw in fact settings groups,

image

dakejahl:
I like it. What about GPS_RAW_INT? We should include the yaw from that message as well. I'm not sure we can assume that the RTK Rover is always GPS2?

DonLakeFlyer:
When updating translations you only updated the english one and leave the translated ones alone. Can you update the pull to only include qgc-json.ts and fix the merge conflicts if any once you do that?

DonLakeFlyer:
> What about GPS_RAW_INT? We should include the yaw from that message as well. I'm not sure we can assume that the RTK Rover is always GPS2?

Also need this

Louis-max-H (commented):
I have tested this PR, here his what is changing : 
- Keep only qgc-json.ts
- Take Yaw from GPS2 (VehicleGPS2FactGroup.cc) and from GPS1 (VehicleGPSFactGroup.cc)

src/Vehicle/FactGroups/VehicleGPSFactGroup.cc
```diff
@@ -64,6 +66,7 @@ void VehicleGPSFactGroup::_handleGpsRawInt(const mavlink_message_t &message)
     hdop()->setRawValue((gpsRawInt.eph == UINT16_MAX) ? qQNaN() : (gpsRawInt.eph / 100.0));
     vdop()->setRawValue((gpsRawInt.epv == UINT16_MAX) ? qQNaN() : (gpsRawInt.epv / 100.0));
     courseOverGround()->setRawValue((gpsRawInt.cog == UINT16_MAX) ? qQNaN() : (gpsRawInt.cog / 100.0));
+    yaw()->setRawValue((gpsRawInt.yaw == UINT16_MAX) ? qQNaN() : (gpsRawInt.yaw / 100.0));
```
dakejahl, line 69: From the message definition.
> Use 0 if this GPS does not provide yaw. Use UINT16_MAX if this GPS is configured to provide yaw and is currently unable to provide it

With your current implementation the GPS without yaw (value 0) will be constantly overwriting the valid value.

dakejahl (changes requested):

src/Vehicle/FactGroups/VehicleGPSFactGroup.cc
Louis-max-H: humm ... I see multiples ways of fixing that, 

Juste before, I want to precise that we have one yaw for each GPS, so if one is set to 0, it will not affect the other one. (I am open to discuss the fact to have 2 yaw for one vehicle)

To solve that we can :
1) Not update value when it's 0 : Vehicle will still have the same yaw even if it's not accurate, can be misleading
2) Set to NaN : Can't make the difference with a GPS unable to provide heading, will also overwrite the value
3) Set to 0 (actual solution) : User is informed that he have lost heading because of the 0 "default" value

Do you want me to make the 1) ?

dakejahl: sorry my mistake, there are separate Facts for each GPS

dakejahl (approved):

src/Vehicle/FactGroups/VehicleGPSFactGroup.cc
Louis-max-H: No problem ! Could have been a good catch !

PR #13024, opened 2025-06-13
Improve flight path calculation with loiter radius support

Eureak-zon:
This commit implements the adjustCoordinatesForLoiterRadius() method to correctly calculate tangent points for flight paths connecting loiter circles. The implementation:

- Handles connections between two loiter circles using proper tangent calculation
- Supports connections between a point and a loiter circle
- Adjusts display coordinates while preserving the original waypoint positions
- Uses geometric calculations based on azimuth and distance between waypoints

image

### DIFF
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -1183,8 +1183,44 @@ FlightPathSegment* MissionController::_createFlightPathSegmentWorker(VisualItemP
     } else if (pair.second->isLandCommand()) {
         segmentType = FlightPathSegment::SegmentTypeLand;
     }
+    SimpleMissionItem* firstItem = nullptr;
+    SimpleMissionItem* secondItem = nullptr;
+    LandingComplexItem* secondLandingItem = nullptr;
+    if (pair.first->isSimpleItem()) {
+        firstItem = qobject_cast<SimpleMissionItem*>(pair.first);
+    }
+    if (pair.second->isSimpleItem()) {
+        secondItem = qobject_cast<SimpleMissionItem*>(pair.second);
+    } else if (pair.second->isLandCommand()){
+        secondLandingItem = qobject_cast<LandingComplexItem*>(pair.second);
+    }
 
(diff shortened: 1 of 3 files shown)


Eureak-zon:
This change is based on my current understanding of QGC, though I realize it may not be an ideal solution.

rubenp02:
Haven't looked at the implementation yet but this was much needed!

DonLakeFlyer:
@rubenp02 Can you double-check this?

rubenp02:
The main issue I see with this conceptually is that at least in ArduPilot, navigation to a loiter isn't done straight to its tangent but instead towards its center, and when the aircraft gets close enough to the orbit path it gets "captured" by it. Not sure how it works in PX4 but I assume its the same. The exit depends on whether the loiter is configured to exit normal or tangent to the loiter trajectory, but in most cases, it's normal. This is a good start but needs to take this, and maybe other things I'm missing right now, into account.

DonLakeFlyer:
@rubenp02 So should we take this as is since it's better than before? The flight path calcs are all approximations in reality. And then try to get even better in another pull? Or is this problematic as is?

rubenp02:
> @rubenp02 So should we take this as is since it's better than before? The flight path calcs are all approximations in reality. And then try to get even better in another pull? Or is this problematic as is?

I think it's problematic because the path visualizations don't match what the aircraft will fly which I think is very critical for obstacle avoidance and such. Maybe keep the improved calculations and revert the map line drawing to how it was before, or adjust the improved one to be correct.

DonLakeFlyer:
> I think it's problematic because the path visualizations don't match what the aircraft will fly which I think is very critical for obstacle avoidance and such

Duh, I totally missed the fact that this was changing the flight path visualization.

> Maybe keep the improved calculations and revert the map line drawing to how it was before, or adjust the improved one to be correct.

I agree

DonLakeFlyer:
@Eureak-zon Can you update this?

Eureak-zon:
What should I do?

Eureak-zon:
@DonLakeFlyer Could you help me review the code? I don't think this implementation is the best at present.

PR #13034, opened 2025-06-17
MissionManager: Fix landing approach item handling

rubenp02:
# MissionManager: Fix landing approach item handling

Description
-----------
Updated the LandingComplexItem recalculation logic to properly handle waypoint approach items, fixing several issues with heading calculations and when loading plans, such as the loiter radius being taken into account despite the "Use loiter to altitude" option being disabled. Refactored the relevant areas to unify the naming and logic between waypoint and loiter-based approaches.

This replaces previous GUI-level workarounds (https://github.com/mavlink/qgroundcontrol/commit/ddd97adb7c4485a78faca0e4d4313c58e88dd366, https://github.com/mavlink/qgroundcontrol/commit/e0f6a0ad46a5f6f6e665cdac33fdcaa7bbfef2d7) that partially fixed these heading calculation problems by manipulating the loiter radius. This commit addresses the root cause instead, so it also fixes the issues for VTOL landing sequences, where the GUI workarounds hadn't yet been applied.

Key changes:
- Reverted the now obsolete GUI hacks.
- Updated the coordinate recalculation logic to handle waypoint and loiter approaches as distinct modes, rather than forcing the waypoint approach through the loiter-based architecture.
- Added _recalcFromApproachModeChange method to properly handle switching between loiter and waypoint approach items, instead of using the generic coordinate change handler.
- Renamed "loiter tangent" to "slope start".
- Updated all UI components to remove some explicit approach item-dependant logic which is no longer needed, and to use the new naming.

Checklist:
----------

### DIFF
--- a/src/MissionManager/FixedWingLandingComplexItem.cc
+++ b/src/MissionManager/FixedWingLandingComplexItem.cc
@@ -173,8 +173,8 @@ void FixedWingLandingComplexItem::_updateFlightPathSegmentsDontCallDirectly(void
     _flightPathSegments.beginResetModel();
     _flightPathSegments.clearAndDeleteContents();
     if (useLoiterToAlt()->rawValue().toBool()) {
-        _appendFlightPathSegment(FlightPathSegment::SegmentTypeGeneric, finalApproachCoordinate(), amslEntryAlt(), loiterTangentCoordinate(),  amslEntryAlt()); // Best we can do to simulate loiter circle terrain profile
-        _appendFlightPathSegment(FlightPathSegment::SegmentTypeLand, loiterTangentCoordinate(), amslEntryAlt(), landingCoordinate(),        amslExitAlt());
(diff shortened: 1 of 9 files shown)


DonLakeFlyer:
So without this what goes wrong with VTOL in 5.0?

rubenp02:
IIRC (can't test it right now), when you don't use a loiter to alt the heading is off because it accounts for the radius of the non existent loiter.

Edit: Just checked, that's exactly it.

DonLakeFlyer:
Hmm, can you sign up for trying to help with quick turn around fixes if I merge this into 5.0 and it blows something up on stable?

rubenp02:
Sure!

DonLakeFlyer:
Ok, it would be tricky to cherry pick this to Stable. Could you instead submit a separate pull for this against Stable as well and then I'll merge both in.

rubenp02:
Done @ #13072

Issue #13023, opened 2025-06-13
Daily QGC, Memory Usage Keeps Increasing During Long-Term Connection to Emulator


Eureak-zon:
### Have you checked the latest release for fixes?

- [x] Yes, I’ve tested on the latest release

### Description

When connecting to the emulator via UDP, memory usage keeps increasing over time. Once the emulator is disconnected, QGC releases some of the memory, but a portion still remains.

It seems that QGC retains some data related to the interaction with the emulator. Is it possible that QGC holds connection-related information even after the link is closed?

If you reconnect to the emulator and then disconnect again, memory usage continues to grow.

Using Visual Leak Detector, we found that QGC has memory leaks in some places.

![Image](https://github.com/user-attachments/assets/2a204848-4a35-4c93-8ef0-c993b272c077)

![Image](https://github.com/user-attachments/assets/e68b686f-ed79-42c0-8425-59df6816b925)

![Image](https://github.com/user-attachments/assets/424e1bca-9c8e-4082-8e1b-1032f3850c3b)

### System Information

Daily QGC

### Log Files and Screenshots

[1.log](https://github.com/user-attachments/files/20731380/1.log)

DonLakeFlyer:
Don't know what QGC version you are running but that isn't a 5.0 build. The instrument panel is wrong.

DonLakeFlyer:
I tested 5.0 on MacOS and memory stayed constant while sitting in SITL emulator connection.

DonLakeFlyer:
Tried connect/disconnect/connect/disconnect. Might see some small increases.

Eureak-zon:
> Don't know what QGC version you are running but that isn't a 5.0 build. The instrument panel is wrong.

The instrument panel is custom

DonLakeFlyer:
I found a decent sized leak on vehicle connect/disconnect/connect/...

But so far nothing significant with staying connected to Vehicle. Still look there. Some small leaks from the guts of Qt and gstreamer but all measured in small number of bytes.

Eureak-zon:
When the simulator continuously sends `_handleCameraFeedback`, a large number of `QGCQGeoCoordinate` instances are created, leading to a significant memory increase. This depends on how many times `_handleCameraFeedback` is processed.

Eureak-zon:
Suggested fix: limit the number of `_cameraTriggerPoints`, or perform a comparison before creating a `QGCQGeoCoordinate`.

aXynt:
[...]

PR #13050, opened 2025-06-19
GStreamer: Fix leak of static pad on shutdown

DonLakeFlyer:
Related to discussion in #13023


DonLakeFlyer:
This needs to be cherry-picked to master as well.

PR #13054, opened 2025-06-20
Restart video delay

### LINKED ISSUE #13051: GStreamer video timeout causes main thread to burn CPU
* Enable RTSP video stream: rtsp://127.0.0.1:8554/test with no video on that IP
* Video stream will timeout on new video
* Main thread will immediately start the same stream again
* Video stream will timeout again
* Repeat until the battery on your laptop is fully consumed by useless quick churn

DonLakeFlyer:
Only restart video ofter a 1 second delay. This way we don't hammer the main thread when the video server isn't there yet on the other end of the connection.

Fix #13051

### DIFF
--- a/src/VideoManager/VideoReceiver/GStreamer/GstVideoReceiver.cc
+++ b/src/VideoManager/VideoReceiver/GStreamer/GstVideoReceiver.cc
@@ -70,7 +70,7 @@ void GstVideoReceiver::start(uint32_t timeout)
     _timeout = timeout;
     _buffer = lowLatency() ? -1 : 0;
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
This need to go over to master as well

dakejahl:
I thought I fixed it here
https://github.com/mavlink/qgroundcontrol/pull/12787/files

but I guess timeout must be a different kind of failure

DonLakeFlyer:
Nope, that's only if there is a failure on stream setup. If the stream setup succeeds but then can't connect then the stream gets stopped and restarted. Different path.

HTRamsey (approved):

PR #13060, opened 2025-06-21
Vehicle: Fix TextMessage Slot Parameters

HTRamsey:
I though enforcing QT_DISABLE_DEPRECATED_UP_TO=0x060800 was supposed to cause the build to fail if there were any of these...


DonLakeFlyer:
What's the repercussions of this? Something that needs to go into 5.0?

HTRamsey:
I think (hope) nothing

PR #13100, opened 2025-06-28
Camera: try to make discovery more robust

julianoes:
This is an attempt to fix some of the corner cases trying to discover cameras, namely:

- Implement missing retries for CAMERA_INFORMATION, us old specific commands and REQUEST_MESSAGE.
- Try to simplify CameraControl::_initWhenReady a bit.
- Fix _requestStreamInfo() and _requestStreamStatus() using wrong retry variable.
- Add a few missing timers.

I haven't been able to properly test each and every retry mechanism used but overall this consistently initializes correctly for my case when it previously does not.

### DIFF
--- a/src/Camera/QGCCameraManager.cc
+++ b/src/Camera/QGCCameraManager.cc
@@ -35,6 +35,8 @@ QGCCameraManager::CameraStruct::CameraStruct(QObject* parent, uint8_t compID_, V
     , compID    (compID_)
     , vehicle   (vehicle_)
 {
+    backoffTimer = new QTimer(this);
+    backoffTimer->setSingleShot(true);
 }
 
 //-----------------------------------------------------------------------------
@@ -162,6 +164,7 @@ void QGCCameraManager::_handleHeartbeat(const mavlink_message_t &message)
                 if (pInfo->lastHeartbeat.elapsed() > 5000) {
                     qCDebug(CameraManagerLog) << "Camera" << message.compid << "reappeared after being silent. Resetting retry count and requesting info.";
                     pInfo->retryCount = 0;  // Reset retry count for fresh attempts
+                    pInfo->backoffTimer->stop();  // Stop any pending backoff timer
                     pInfo->lastHeartbeat.start();
                     _requestCameraInfo(pInfo);
                 } else {
@@ -251,6 +254,7 @@ QGCCameraManager::_handleCameraInfo(const mavlink_message_t& message)
         //-- Flag it as done
         _cameraInfoRequest[sCompID]->infoReceived = true;
         _cameraInfoRequest[sCompID]->retryCount = 0;  // Reset retry counter on success
+        _cameraInfoRequest[sCompID]->backoffTimer->stop();  // Stop any pending backoff timer
         qCDebug(CameraManagerLog) << "_handleCameraInfo: Success for compId" << message.compid << "- reset retry counter";
         mavlink_camera_information_t info;
         mavlink_msg_camera_information_decode(&message, &info);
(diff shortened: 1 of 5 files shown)


julianoes:
This does improve camera discovery for me but I want to walk through it once more and verify what is going on.

julianoes:
I tested this again and went through it once again. I believe it's in a better state overall but if you have concerns, let me know.

DonLakeFlyer:
I don't know if this would make your code more complicated with all the timers and whatnot but there is a method for the "try this and if that doesn't work do this instead" pattern:
```
    /// Sends the command and calls the fallback lambda function in
    /// case the command is MAV_RESULT_UNSUPPORTED
    void sendMavCommandWithLambdaFallback(
        std::function<void()> lambda,
        int compId, MAV_CMD command,
        bool showError,
        float param1 = 0.0f, float param2 = 0.0f, float param3 = 0.0f, float param4 = 0.0f, float param5 = 0.0f, float param6 = 0.0f, float param7 = 0.0f);
```


Not sure if that would simplify or complicate your code?

julianoes:
@DonLakeFlyer yes that would probably make sense to use. Does this command do internal retries or not? If it does, it wouldn't work because these need to happen outside where we can try both new and legacy commands.

DonLakeFlyer:
Only these commands will retry:
```
bool Vehicle::_sendMavCommandShouldRetry(MAV_CMD command)
{
    switch (command) {
#ifdef QT_DEBUG
    // These MockLink command should be retried so we can create unit tests to test retry code
    case MockLink::MAV_CMD_MOCKLINK_ALWAYS_RESULT_ACCEPTED:
    case MockLink::MAV_CMD_MOCKLINK_ALWAYS_RESULT_FAILED:
    case MockLink::MAV_CMD_MOCKLINK_SECOND_ATTEMPT_RESULT_ACCEPTED:
    case MockLink::MAV_CMD_MOCKLINK_SECOND_ATTEMPT_RESULT_FAILED:
    case MockLink::MAV_CMD_MOCKLINK_NO_RESPONSE:
        return true;
#endif

        // In general we should not retry any commands. This is for safety reasons. For example you don't want an ARM command
        // to timeout with no response over a noisy link twice and then suddenly have the third try work 6 seconds later. At that
        // point the user could have walked up to the vehicle to see what is going wrong.
        //
        // We do retry commands which are part of the initial vehicle connect sequence. This makes this process work better over noisy
        // links where commands could be lost. Also these commands tend to just be requesting status so if they end up being delayed
        // there are no safety concerns that could occur.
    case MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES:
    case MAV_CMD_REQUEST_PROTOCOL_VERSION:
    case MAV_CMD_REQUEST_MESSAGE:
    case MAV_CMD_PREFLIGHT_STORAGE:
    case MAV_CMD_RUN_PREARM_CHECKS:
        return true;

    default:
        return false;
    }
}
```

julianoes:
So this would retry the MAV_CMD_REQUEST_MESSAGE command and that means we can't use it. Unless we add an argument that let's us specify whether to retry.

DonLakeFlyer:
How come you don't want to retry the request message?

julianoes:
I am retrying it with the logic around it but I'm retrying like with the legacy command interleaved, e.g.:
1. REQUEST_MESSAGE (CAMERA_INFORMATION)
2. MAV_CMD_REQUEST_CAMERA_INFORMATION (deprecated)
3. REQUEST_MESSAGE
etc.

5 times each with exponential back off for slow cameras.

DonLakeFlyer:
OK, understood on retries.

@julianoes I poked at this a bit and the crashes in unit tests are real. They are crashing in the camera retry logic. Run with --unittest and a debugger.

julianoes:
Ok, I will look into that. Thanks!

julianoes:
I've added a commit to fix the unit tests crashing on destruction.

DonLakeFlyer:
You should bring this across to Stable as well

julianoes:
@DonLakeFlyer I tried but it's a bit tricky with quite a few conflicts, mostly due to the changes in https://github.com/mavlink/qgroundcontrol/pull/11178.

Edit: oh, you mean Stable_V5.0, not V4.4!

julianoes:
@DonLakeFlyer backport: https://github.com/mavlink/qgroundcontrol/pull/13342

PR #13120, opened 2025-07-01
QML: Fix Deprecated Signal Functions

HTRamsey:
still some of these left


DonLakeFlyer:
Are you finding these with qmllint? Or just by global search?

HTRamsey:
qmllint

PR #13144, opened 2025-07-07
Docs: Updates for 5.0

DonLakeFlyer:


docs/.vitepress/config.mjs
```diff
@@ -158,6 +158,10 @@ export default defineConfig({
             text: "master",
             link: "https://docs.qgroundcontrol.com/master/en/",
           },
+          {
```
hamishwillee, line 161: FYI, this is right, but if you also want the links in other versions you'll have to also update this in those branches.

DonLakeFlyer: So this is in 5.0 and master. When you do PX4, how far back do you go?

hamishwillee: > So this is in 5.0 and master. When you do PX4, how far back do you go?

We have people still working on PX4 v1.12. I go back that far because it is irritatingly popular. I hope to drop that when v1.16 is released.

However for QGC it is really is a matter of your support model and compatibility. If it is impossible to get an old version of QGC then there is no point documenting it for fliers, but it might be worth doing limited support for people who have forked.

My instinct is that for QGC you don't need to go very far back. What you do need to do is make it clear what versions of PX4/ArduPilot are supported by each version of QGC.

PR #13159, opened 2025-07-10
Joystick: Update to SDL3

HTRamsey:
This may improve joystick compatibility & functionality
Also update thread safety in JoystickManager and cleanup vehicle joystick handling

### DIFF
--- a/cmake/CustomOptions.cmake
+++ b/cmake/CustomOptions.cmake
@@ -34,6 +34,9 @@ option(QGC_ENABLE_GST_VIDEOSTREAMING "Enable GStreamer Video Backend" ON)
 cmake_dependent_option(QGC_CUSTOM_GST_PACKAGE "Enable Using QGC Provided Custom GStreamer Packages" OFF "QGC_ENABLE_GST_VIDEOSTREAMING" OFF)
 option(QGC_ENABLE_QT_VIDEOSTREAMING "Enable QtMultimedia Video Backend" OFF) # Qt6Multimedia_FOUND
 
+# Joystick
+set(SDL_GAMECONTROLLERCONFIG "0300000009120000544f000011010000,OpenTX Radiomaster TX16S Joystick,leftx:a3,lefty:a2,rightx:a0,righty:a1,platform:Linux" CACHE STRING "Custom SDL Joystick Mappings")
+
 # MAVLink
 set(QGC_MAVLINK_GIT_REPO "https://github.com/mavlink/c_library_v2.git" CACHE STRING "URL to MAVLink Git Repo")
 set(QGC_MAVLINK_GIT_TAG "19f9955598af9a9181064619bd2e3c04bd2d848a" CACHE STRING "Tag of MAVLink Git Repo")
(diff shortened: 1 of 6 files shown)


src/Joystick/JoystickSDL.cc
```diff
@@ -14,119 +14,118 @@
...
-    // qCDebug(JoystickSDLLog) << Q_FUNC_INFO << this;
+    qCDebug(JoystickSDLLog) << this;
 }
 
 bool JoystickSDL::init()
 {
-    SDL_SetMainReady();
-    if (SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER | SDL_INIT_JOYSTICK) < 0) {
-        (void) SDL_JoystickEventState(SDL_DISABLE);
+    if (!SDL_InitSubSystem(SDL_INIT_GAMEPAD | SDL_INIT_JOYSTICK)) {
```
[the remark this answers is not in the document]
DonLakeFlyer: The `< 0` does seem clearer given the return values

cmake/CustomOptions.cmake
```diff
@@ -34,6 +34,9 @@ option(QGC_ENABLE_GST_VIDEOSTREAMING "Enable GStreamer Video Backend" ON)
 cmake_dependent_option(QGC_CUSTOM_GST_PACKAGE "Enable Using QGC Provided Custom GStreamer Packages" OFF "QGC_ENABLE_GST_VIDEOSTREAMING" OFF)
 option(QGC_ENABLE_QT_VIDEOSTREAMING "Enable QtMultimedia Video Backend" OFF) # Qt6Multimedia_FOUND
 
+# Joystick
+set(QGC_SDL_GAMECONTROLLERCONFIG "0300000009120000544f000011010000,OpenTX Radiomaster TX16S Joystick,leftx:a3,lefty:a2,rightx:a0,righty:a1,platform:Linux" CACHE STRING "Custom SDL Joystick Mappings")
```
DonLakeFlyer: The db entry is wrong or missing in the one we use? Or is this just an example for the override possibility?

src/Joystick/JoystickSDL.cc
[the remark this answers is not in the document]
HTRamsey: Copilot is wrong about this, it's using the SDL2 return value. SDL3 just returns a bool https://wiki.libsdl.org/SDL3/SDL_InitSubSystem

cmake/CustomOptions.cmake
HTRamsey: There have been a few issues related to this joystick, and somebody commented that this config fixes it. Haven't verified it myself tho

DonLakeFlyer: Should we have our own override file we read in to tweak some set over the top of the gamecontrollerdb file? Seems kinda specific to override a single joystick which a compiler option.

HTRamsey: we can, although I'd be surprised if we ever actually have to do this again for another joystick. (And maybe SDL3 fixes it by itself anyways)

DonLakeFlyer: I guess in reality if some commercial company hits a problem with there systems it's gonna be a specific single joystick that needs tweaking. So this way is probably fine then. No need to get fancier.

PR #13161, opened 2025-07-10
Viewer3D: Use geographiclib

HTRamsey:
Just use geographiclib rather than manually writing conversions


HTRamsey:
@omid-esrafilian thoughts?

DonLakeFlyer (approved):

PR #13184, opened 2025-07-17
docs: update download instructions

mrpollo:
Fixes issue with info box, and updates linux instructions

### DIFF
--- a/docs/en/qgc-user-guide/getting_started/download_and_install.md
+++ b/docs/en/qgc-user-guide/getting_started/download_and_install.md
@@ -39,7 +39,7 @@ _QGroundControl_ can be installed on macOS 12 (Monterey) or later:
 
 ::: info
 QGroundControl continues to not be signed. You will not to allow permission for it to install based on you macOS version.
-::
+:::
 
 ## Ubuntu Linux {#ubuntu}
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
QGC has this code checked at boot:
```
    #ifdef Q_OS_LINUX
    #ifndef Q_OS_ANDROID
    #ifndef QGC_NO_SERIAL_LINK
        if (!_runningUnitTests) {
            // Determine if we have the correct permissions to access USB serial devices
            QFile permFile("/etc/group");
            if(permFile.open(QIODevice::ReadOnly)) {
                while(!permFile.atEnd()) {
                    const QString line = permFile.readLine();
                    if (line.contains("dialout") && !line.contains(getenv("USER"))) {
                        permFile.close();
                        showAppMessage(tr(
                            "The current user does not have the correct permissions to access serial devices. "
                            "You should also remove modemmanager since it also interferes.<br/><br/>"
                            "If you are using Ubuntu, execute the following commands to fix these issues:<br/>"
                            "<pre>sudo usermod -a -G dialout $USER<br/>"
                            "sudo apt-get remove modemmanager</pre>"));
                        break;
                    }
                }
                permFile.close();
            }
        }
    #endif
    #endif
    #endif
```

Is this no longer correct?

mrpollo:
It's still valid. The diff might be throwing you off. I didn't remove anything, only updated and expanded. I'm only adding extra info to clarify why each step is needed, specifically calling out that the workarounds aren't a QGC problem.

The current documentation is copy-paste reliant and assumes the user doesn't fully understand what's happening. This update explains each line and its purpose.

DonLakeFlyer:
Ok, yeah sorry I see it now

Issue #12947, opened 2025-05-29
ESC health, RPM, temperature reporting


rmackay9:
### Feature Description

ESCs that provide health, RPM and temperature feedback are becoming increasingly common so it would be good to report this to the pilot in a clear way.  I think the most intuitive display would be similar to the existing GPS or battery displays which would be a coloured icon showing the health (white or green for OK, orange or red for unhealthy) and then if the user hovers over the icon or clicks it  more information is displayed including the per-ESC RPM, temperature and errors (so users can see **which** ESC is struggling).

As an example of where this would have helped, we have [this users report from the ArduPilot 4.6 forums](https://discuss.ardupilot.org/t/copter-flipped-and-crashed-ekf3-innovations-high-seeking-help-for-analysis/134912/7) where AM32 bi-directional DShot ESCs started reporting errors 4 minutes before one of them conked out and the vehicle crashed.  With better warning it is quite likely that the pilot would have landed safely

I suspect someone will read this and immediately tell me that QGC already includes this and all that ArduPilot needs to do is send the WIP ESC_INFO and ESC_STATUS messages but that would also be a great result.

### Flight Stacks

ArduPilot

### Vehicle Types

Multirotor

HTRamsey:
I see esc_status is read but nothing is done with it I don't think. But I agree this is important and I'd like it for v5.1

dakejahl:
@rmackay9 What do you think?
![Image](https://github.com/user-attachments/assets/8fcdc387-1269-4474-9322-99787959cfa4)

If you add sending of ESC_STATUS and ESC_INFO this will populate. Also working on adding the RPM to the motor test page. For example with PX4 

![Image](https://github.com/user-attachments/assets/73e69d01-21c8-4b2a-ab17-e38830681497)

I can do the same with the APMMotorComponent.qml

DonLakeFlyer:
Can you make the look match Battery Status where each Motor is in it's own section. The format you have with the Green lines looks nice but it's inconsistent with the rest of the UI.

DonLakeFlyer:
Or does the green line indicate health?

dakejahl:
Yeah green lines are the health. I'll take a look at how the battery looks and try to match style. I have to try and keep it somewhat compact so that we can show potentially 8 motors

dakejahl:
A better motor icon would be nice. I'll see if I can find/make something

DonLakeFlyer:
[...]

PR #13191, opened 2025-07-20
ESC Status Toolbar

dakejahl:
Toolbar icon for ESC status. Requires reception of ESC_INFO and ESC_STATUS mavlink messages.

**TODO**
- clean up messy claude logic in qml code
- any way to have a vector/array of facts instead of first/second/third/etc naming?
- test with 8 motors
- add RPM to Ardupilot MotorComponent
- test with Ardupilot

Closes https://github.com/mavlink/qgroundcontrol/issues/12947

image
image


Davidsastresas:
Thanks Jake, I will try to take a look at it this week

Davidsastresas:
About the vector thing, I think if you check how batteryfactgroup is done we could do something similar maybe here

DonLakeFlyer:
> About the vector thing, I think if you check how batteryfactgroup is done we could do something similar maybe here

Yup

DonLakeFlyer:
@dakejahl Can you add screenshots?

dakejahl:
Added screenshots to the PR description

dakejahl:
The "max temp" and "max current" are just what QGC has seen... I'm not sure if it's that useful. If we cared about max values the FC should report that explicitly.

DonLakeFlyer:
@dakejahl Do you mind if I take over the ESC_STATUS and ESC_INFO telemetry handling part of this? I'd like to create a generic FactGroup list mechanism which will handle multiples. The battery FactGroup code already has most of what is needed, it's just not done in a fully generic way. If so, just sit on this for a bit and I'll create the telemetry backend for all of it. Then you can adjust the rest to that.

dakejahl:
I would love if you did that! Thank you! I will rework this as needed once you have the backend implemented

dakejahl:
I've also got a PR here for adding a new mavlink message ESC_EEPROM. This would just be an array of uint8.
https://github.com/mavlink/mavlink/pull/2324/files

Similar to BatteryFactGroup should I just use a uint8 pointer Fact and allocate an array?

DonLakeFlyer:
Sorry for the delay, I was out of town. I'll get this done by end of day today.

PR #13192, opened 2025-07-20
Vehicle: Minor Telemetry Fixes

HTRamsey:


DonLakeFlyer:
This is really odd. I swear I've looked at voltage in my own flights and it seemed fine. But it sure seems wrong in the code.

DonLakeFlyer:
> But it sure seems wrong in the code.

Sorry, looked at change wrong. This doesn't really change the voltage values does it? Not sure where the problem is and how it fixes it. I get voltage values fine with my vehicle.

HTRamsey:
Yeah I don't have any issues either. I was going to look into it more later. But I did find that minDistance of VehicleDistanceSensorFactGroup was never set so I added this anyways

DonLakeFlyer:
Any reason for this to be Draft? Seems like a good fix that should go in to master and Stable

HTRamsey:
Was trying to figure out if I could reproduce the original issue too

PR #13202, opened 2025-07-22
fix/telemetry-color-to-hex

new-tonAA:
### Problem

In the Telemetry Display -> Value Range UI, when setting a custom color for a specific value range, the selected color always reverts to white. This makes the value-based coloring feature unusable.

### Root Cause

The QML code attempted to format the selected color using `Qt.format(...)`, which does not exist in QML/JavaScript. This caused a runtime error and the color assignment silently failed, resulting in the default fallback color `#ffffff`.

### Fix

- Replaced the use of `Qt.format()` with a JavaScript `colorToHex()` function.
- `colorToHex()` converts the selectedColor from the `ColorDialog` to a valid hex string (`#rrggbb`), which QGroundControl stores correctly.

### Result

- Color selection now works as expected.
- The configured color is saved and restored properly.
- No side effects or regressions expected, since the change is scoped to `InstrumentValueEditDialog.qml`.

Fix tested on desktop build (Qt 6.8.3), QGC v5.0.4 source.

### DIFF
--- a/src/QmlControls/InstrumentValueEditDialog.qml
+++ b/src/QmlControls/InstrumentValueEditDialog.qml
@@ -279,7 +279,16 @@ QGCPopupDialog {
                 id:             colorPickerDialog
                 modality:       Qt.ApplicationModal
                 selectedColor:  instrumentValueData.rangeColors.length ? instrumentValueData.rangeColors[colorIndex] : "white"
-                onAccepted:     updateColorValue(colorIndex, color)
+                onAccepted: {
+                    function colorToHex(c) {
+                        function toHex(d) {
+                            var s = Math.round(d * 255).toString(16);
+                            return s.length === 1 ? "0" + s : s;
+                        }
+                        return "#" + toHex(c.r) + toHex(c.g) + toHex(c.b);
+                    }
+                    updateColorValue(colorIndex, colorToHex(colorPickerDialog.selectedColor))
+                }
 
                 property int colorIndex: 0
             }


src/QmlControls/InstrumentValueEditDialog.qml
```diff
@@ -279,7 +279,16 @@ QGCPopupDialog {
                 id:             colorPickerDialog
                 modality:       Qt.ApplicationModal
                 selectedColor:  instrumentValueData.rangeColors.length ? instrumentValueData.rangeColors[colorIndex] : "white"
-                onAccepted:     updateColorValue(colorIndex, color)
+                onAccepted: {
```
DonLakeFlyer: I don't think all of this is needed. As far as I can tell the actual bug is the fact the it should be using 'selectedColor' in the `updateColorValue` call.

It should just be: `onAccepted:     updateColorValue(colorIndex, selectedColor)`. Which works for me in my testing. Am I missing something here?

new-tonAA: Thanks a lot for the review!

You're absolutely right — I re-tested with the simpler version:
onAccepted: updateColorValue(colorIndex, selectedColor)

It works exactly as expected, and there's no need for the manual hex conversion. I’ll updated the PR accordingly to reflect your suggestion.

new-tonAA: Thanks a lot for the review!

You're absolutely right — I re-tested with the simpler version: onAccepted: updateColorValue(colorIndex, selectedColor)

 It works exactly as expected, and there's no need for the manual hex conversion. I’ll updated the PR accordingly to reflect your suggestion.

PR #13204, opened 2025-07-23
GimbalIndicator: prevent text from jittering around

julianoes:
This fixes the issue where the gimbal indicators Azimuth and Yaw keep jumping around when the value is around 0, going from -0 to 0.

The fix (from what I understand):
- Makes the boxes have a fixed width,
- And round to full degrees.

### DIFF
--- a/src/UI/toolbar/GimbalIndicator.qml
+++ b/src/UI/toolbar/GimbalIndicator.qml
@@ -21,7 +21,7 @@ import QGroundControl.FactControls
 
 Item {
     id:             control
-    width:          gimbalIndicatorIcon.width * 1.1 + gimbalTelemetryLayout.childrenRect.width + margins
+    width:          gimbalIndicatorIcon.width * 1.1 + ScreenTools.defaultFontPixelWidth * 12 + margins
     anchors.top:    parent.top
     anchors.bottom: parent.bottom
 
@@ -376,6 +376,7 @@ Item {
         anchors.left:              gimbalIndicatorIcon.right
         anchors.leftMargin:        margins
         anchors.verticalCenter:    parent.verticalCenter
+        width:                     ScreenTools.defaultFontPixelWidth * 12
         columns:                   2
         rows:                      2
         rowSpacing:                0
@@ -389,17 +390,20 @@ Item {
                                         (activeGimbal && activeGimbal.yawLock ? qsTr("Yaw locked") : qsTr("Yaw follow"))
             Layout.columnSpan:      2
             Layout.alignment:       Qt.AlignHCenter
+            Layout.preferredWidth:  ScreenTools.defaultFontPixelWidth * 10
         }
         QGCLabel {
             id:                     pitchLabel
-            text:                   activeGimbal ? qsTr("P: ") + activeGimbal.absolutePitch.rawValue.toFixed(1) : ""
+            text:                   activeGimbal ? qsTr("P: ") + Math.round(activeGimbal.absolutePitch.rawValue) : ""
+            Layout.preferredWidth:  ScreenTools.defaultFontPixelWidth * 5
         }
         QGCLabel {
             id:                     panLabel
             text:                   activeGimbal ? 
-                                        gimbalTelemetryLayout.showAzimuth ? (qsTr("Az: ") + activeGimbal.absoluteYaw.rawValue.toFixed(1)) :
-                                            (qsTr("Y: ") + activeGimbal.bodyYaw.rawValue.toFixed(1)) :
+                                        gimbalTelemetryLayout.showAzimuth ? (qsTr("Az: ") + Math.round(activeGimbal.absoluteYaw.rawValue)) :
+                                            (qsTr("Y: ") + Math.round(activeGimbal.bodyYaw.rawValue)) :
                                                 ""
+            Layout.preferredWidth:  ScreenTools.defaultFontPixelWidth * 6
         }
     }
 


Davidsastresas:
Hi Julian, it is true that around 0 that behavior was annoying, thanks for addressing it.

Your approach seems fine to me. However, I am not sure if I like to remove completely floating points. Some users might like to have that decimal point there.

If you think it makes sense, I think we could expose a setting in gimbalcontrollersettings, a quick checkbox, and add it to GimbalIndicator.qml, we have the gimbal settings there too.

If you already tried your approach with floating points and the extra width made this all look weird, it isn't a big deal, I am fine too with no floating points.

Thanks!

julianoes:
Thanks for the review @Davidsastresas. I guess I don't understand why you would want the floating point. It's not like you do something with that exact number, it's more of an indicator. Or what would be the purpose of knowing 0.1 degrees?

Davidsastresas:
I agree most users won't mind.

I have in mind some niche users like survey people doing mapping of solar panels. For some reason it is really important for the sensor to be parallel to the panel's angle, and they could appreciate the extra floating point to be extra sure everything is fine during mission.

In any case, those niche users will probably have a QGC custom fork anyway, so if you think it is a good idea to remove the floating point I am fine with it too.

Thanks!

DonLakeFlyer:
> I have in mind some niche users like survey people doing mapping of solar panels. For some reason it is really important for the sensor to be parallel to the panel's angle, and they could appreciate the extra floating point to be extra sure everything is fine during mission.

They could add the value to telemetry if they want more detailed?

julianoes:
Can they? Or do we need to add that somehow?

DonLakeFlyer:
> Or do we need to add that somehow?

The Gimbal values could be set up as a FactGroup then they are available to the telemetry display.

DonLakeFlyer:
> This fixes the issue where the gimbal indicators Azimuth and Yaw keep jumping around when the value is around 0, going from -0 to 0.

Is this just a problem with 0s? Would a gimbal angle bounce around between say 9 and 10 causing the digit count to change constantly? Reason I ask is that I'd rather not have this be fixed width. Toolbar space is at a premium. You want to take as little as possible.

julianoes:
> Would a gimbal angle bounce around between say 9 and 10 causing the digit count to change constantly?

Yes, it also changes from 9 to 10, and any change really as it's not a monospace font.

If you ask me, the gimbal angle probably shouldn't be up there like that in the first place and could be a more intuitive indicator, but it's better than nothing.

DonLakeFlyer:
> Yes, it also changes from 9 to 10, and any change really as it's not a monospace font.

Ok, let me play around with this a bit

DonLakeFlyer:
> Yes, it also changes from 9 to 10

But if the gimbal position is static would it be possible for the gimbal angle reported to bounce between 9.5 and 9.4 causing the rounded value to bounce between 9 and 10 even though the gimbal is not moving.

Or to put the question another way if we just round the gimbal values and don't do fix width. Will the only time the gimbal indicator change size happen when the gimbal is actually being moved.

Davidsastresas:
@DonLakeFlyer That's correct, with rounding that should only change when gimbal moves.

However keep in mind we have the option of showing yaw as azimuth or local yaw. So depending on this option and gimbal being on locked or follow mode, yaw/azimuth could change with vehicle yaw.

About having values there at all, I understand it is a larger indicator than usual. My initial idea for it was an additional button/indicator in flyview, something like this:
Screenshot from 2025-07-28 19-43-45

and after click expands to the whole controls panel. But I understand this could be too invasive for fly view as well.

So I am really okay with whatever you guys decide. We always have time to keep iterating if we miss anything in the future.

julianoes:
@Davidsastresas your solution looks nice. I'll close my PR then.

Davidsastresas:
Thanks @julianoes. That was the initial approach, but @DonLakeFlyer suggested to based it instead as a top toolbar indicator, so it is like that now.

I would need confirmation from Don's side that he is fine with it to make a PR of it.

I am happy with either choice.

DonLakeFlyer:
My opinion hasn't changed. I'm still not super interested in creating yet another new place for UI stuff. If anything, I would say that the more likely place to move this would be within the camera control on the right. If toolbar is that bad.

julianoes:
I really don't care, as long as it doesn't jitter around like it does right now.

Davidsastresas:
I understand @DonLakeFlyer position. However I don't think it will be better than it is now in the camera widget. I think they kind of widget this is is definetely better in a corner or side. So I would say to leave it where it is.

DonLakeFlyer:
> However I don't think it will be better than it is now in the camera widget. I think they kind of widget this is is definetely better in a corner or side.

Still don't understand why. The camera widget is in the upper right corner?

Davidsastresas:
> Still don't understand why. The camera widget is in the upper right corner?

Sorry I didn't check this until now. You are right, I was still thinking in QGC 4 where it was in the middle of the right. In that case I might give it a go and try to integrate it there. Thanks!

DonLakeFlyer:
> In that case I might give it a go and try to integrate it there.

I'm going to get to this as some point. I've got a vehicle now I can put together with video and gimbal. Just haven't had time to do it yet. A focus for me for 5.1 will be camera/video/gimbal.

DonLakeFlyer:
I'll take care of this with the other gimbal stuff I'm doing in 5.1.

Issue #5511, opened 2017-07-30
Daily Build / Beta Icon for all OS


dagar:
In the Play store the QGroundControl and QGC (Daily Test Build) have slightly different icons, but on the active device they're using the same icon.

![image](https://user-images.githubusercontent.com/84712/28755018-d887e050-751f-11e7-9312-85cd363070cb.png)

We need an a better beta/test icon generated for android in all sizes, then build scripting to change the icon here. https://github.com/mavlink/qgroundcontrol/blob/master/android/AndroidManifest.xml#L3
https://github.com/mavlink/qgroundcontrol/blob/master/tools/update_android_version.sh#L16

hamishwillee:
The UX team at Intel are going to love the opportunity to help with this. BUT if we want help we need to provide very clear direction. So, list of what you want ...

dagar:
Something similar to what Google did with Android Chrome?

![image](https://user-images.githubusercontent.com/84712/28758155-66264320-7561-11e7-8dfa-60dee6bf7886.png) https://play.google.com/store/search?q=chrome

hamishwillee:
Also, are we happy with the current iconography, or tempted by elements of the proposed DC branding: https://www.youtube.com/watch?v=i4kblPyLdBw&feature=youtu.be

W.r.t. the above @dagar I like that, but this exact approach might cut the bottom off the Q. All we want is something that is recognizably QGroundControl and says "Beta" right?

anassinator:
@dagar, @hamishwillee:

From my experience dogfooding, I've seen two ways of distinguishing the beta/nightly builds from prod:
- Adding a banner as Chrome does above
- Changing the colors or background somehow to make it look like a WIP or different enough from prod such that they can't be confused

What do you guys think of changing the background instead? This is a quick and rough example of what I mean:

![android_dev_512x512](https://user-images.githubusercontent.com/723610/28808287-6dd78728-7683-11e7-8f7a-62766994bb99.png)

Original for comparison:

![android_512x512](https://user-images.githubusercontent.com/723610/28808130-6b6edcb2-7682-11e7-9dd9-83808cfd85e2.png)

On a side note, I'd love to see a more DC-inspired icon.

dagar:
I'm not sure if the default QGC coloring is sufficiently recognizable at this point. Can you still slap a beta label across it?

DonLakeFlyer:
like the graph paper style thing. But agree that some sort of Beta label is likely still needed.

[...]

PR #13205, opened 2025-07-23
Allow custom build to override application window icon on Linux

Maarrk:
Allow custom build to override application window icon on Linux

Description
-----------

Moved the call to `setWindowIcon` from constructor to `_initForNormallAppBoot`, so the icon URL can be modified by `CustomPlugin` the same way as all imports from QML are (prefix with `/Custom` if the result of that is defined).

This way the icon resource defined here is used correctly:

https://github.com/mavlink/qgroundcontrol/blob/c5c0fc5bb1e0f76d3ff06f0f4ded173c53c5817b/custom-example/custom.qrc#L36-L38

Test Steps
-----------

Do custom build according to the instructions:
https://github.com/mavlink/qgroundcontrol/blob/c5c0fc5bb1e0f76d3ff06f0f4ded173c53c5817b/custom-example/README.md?plain=1#L3-L10

**Note:** On one of my setups (Ubuntu 24.04 noble) QGroundControl window always has the application icon unset, including the official v5.0 AppImage. However when tested on other machines (eg. Ubuntu 22.04 jammy) where the stable build has an icon, this change works as expected.

Checklist:
----------

Related Issue
-----------

Didn't find any issues for that specifically, but it is probably because of relatively recent switch to CMake. This may possibly help with #5511.

### DIFF
--- a/src/QGCApplication.cc
+++ b/src/QGCApplication.cc
@@ -104,9 +104,6 @@ QGCApplication::QGCApplication(int &argc, char *argv[], bool unitTesting, bool s
     setOrganizationName(QGC_ORG_NAME);
     setOrganizationDomain(QGC_ORG_DOMAIN);
     setApplicationVersion(QString(QGC_APP_VERSION_STR));
-#ifdef Q_OS_LINUX
-    setWindowIcon(QIcon(":/res/qgroundcontrol.ico"));
-#endif
 
     // Set settings format
     QSettings::setDefaultFormat(QSettings::IniFormat);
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@HTRamsey I think I may have missed something with the url interceptor. Are they only intercepted in Qml code but not C++ code?

HTRamsey:
@DonLakeFlyer I think the format ":/my/file" looks on file rather than going through the resource system, so the intercept doesn't get used unless you use the qrc:/my/file format. That's what it looks like this PR is changing

Maarrk:
@DonLakeFlyer as far as I understand the interceptor is part of Qml engine, so it does not work for all C++ code automatically.

@HTRamsey I did try the `QIcon` constructor with `qrc:/path` but it didn't work. I believe `:/path` does use the resource system, I'm surprised that the `qrc` prefix isn't supported, but that's how all examples from Qt documentation look like

DonLakeFlyer:
Does it need to be: `setWindowIcon(QIcon(QUrl::fromUserInput("qrc::/res/qgroundcontrol.ico")));`?

That's how it's used everywhere else in c++.

Maarrk:
> Does it need to be: `setWindowIcon(QIcon(QUrl::fromUserInput("qrc::/res/qgroundcontrol.ico")));`?

Sorry, there is no `QIcon` constructor that accepts `QUrl`.

If I just use `QUrl::fromUserInput` for the initial value, and don't pass it to `QQmlEngine::interceptUrl` the default QGC icon is used.

I did find a way to write the URL to String line in a way that's a bit simpler to read though. Removing the leading `":"` also breaks the icon.

---

For completeness, this is my result now:

obraz

PR #13229, opened 2025-07-29
fix: replace Chinese comma with English comma in translation to fix mission display issue

new-tonAA:
### Summary
This PR fixes a bug in the Chinese translation where the use of a Chinese comma (`，`) in the `Indoor,Outdoor` enumeration causes mission waypoints to fail displaying properly.

### Details
- Changed the delimiter in the translation string from the Chinese comma to an English comma (`,`) as required by the QGroundControl parsing logic.
- This prevents mission load errors caused by improper string splitting.

### Impact
- Only affects the Chinese translation file.
- No functional change in the source code.
- Fixes the issue where missions with certain translations fail to display correctly in the UI.

### References
- Original issue discovered during testing of mission display on Chinese localization.
- Related discussion: QGroundControl translation guidelines.

Please review and merge if appropriate. Thanks!

### DIFF
--- a/translations/qgc_json_zh_CN.ts
+++ b/translations/qgc_json_zh_CN.ts
@@ -71,7 +71,7 @@
       <translatorcomment>Only use english comma &apos;,&apos; to separate strings</translatorcomment>
       <location filename="../src/FirmwarePlugin/APM/APM-MavCmdInfoCommon.json"/>
       <source>On,Off</source>
-      <translation>开启，关闭</translation>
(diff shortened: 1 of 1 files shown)


HTRamsey:
I think this needs to be done from Crowdin otherwise it will get overwritten

new-tonAA:
Thanks for the reminder. I’ll make the same change directly via Crowdin to avoid it being overwritten.

PR #13230, opened 2025-07-29
More QML integration

s-lisovenko:
Description
-----------
More QML cmake integration: Resources integration.

Checklist:
----------

Related Issue
-----------


HTRamsey:
Using multiple qt_add_resources breaks 6.6.3, which is why I held off on this. @DonLakeFlyer thoughts on 6.6.3 support? Leave it in v5.0 branch but kill it in master?

s-lisovenko:
> Using multiple qt_add_resources breaks 6.6.3, which is why I held off on this. @DonLakeFlyer thoughts on 6.6.3 support? Leave it in v5.0 branch but kill it in master?

Hmm, it's strange, I'm pretty sure this has worked for me in the past on different project.

s-lisovenko:
> > Using multiple qt_add_resources breaks 6.6.3, which is why I held off on this. @DonLakeFlyer thoughts on 6.6.3 support? Leave it in v5.0 branch but kill it in master?
> 
> Hmm, it's strange, I'm pretty sure this has worked for me in the past on different project.

You are right, since we use one target name for all resources, this leads to errors in Qt 6.6. Looks like we can’t merge this PR because qt 6.6.3. I leave it if you decide stops 6.6.3 support i can do rebasing and we can merge it.

HTRamsey:
I want to abandon 6.6 in master anyways, so I'll make a PR to do so, then this can be merged right after whenever that one gets merged

HTRamsey:
@s-lisovenko could you fix this up now?

src/CMakeLists.txt
```diff
@@ -97,6 +97,7 @@ target_link_libraries(${CMAKE_PROJECT_NAME}
         ToolbarModule
         UTMSPModule
         VehicleSetupModule
+        PX4FirmwareModule
```
HTRamsey: Shouldn't this be guarded by an if statement like AutoPilotPluginsPX4Module is?

s-lisovenko: Yes, thanks, I added the link directly to the module cmake

PR #13240, opened 2025-07-30
Show airframe ID as vehicle name if vehicle name is not known

gillamkid:
Description
-----------
If the vehicle airframe details aren't stored in QGC's `AirframeFactMetaData.xml` file, the Vehicle name is left blank. It would be more informative to show the airframe ID when the vehicle name isn't known.

ex. 
Without the changes in this PR, If you use the PX4 SIH sim
```
~/PX4-Autopilot$ make px4_sitl sihsim_quadx
```
you see
before

with the changes in this PR, you will now see `10040` as the vehicle name. This is the airframe ID. A developer who knows about airframe IDs will know the meta data for the vehicle can be manually viewed in the airframe file, in this case `PX4-Autopilot/ROMFS/px4fmu_common/init.d-posix/airframes/10040_sihsim_quadx`
after

The vehicle name still shows the normal stuff if the vehicle has a known airframe id. For example, if you run
```
~/PX4-Autopilot$ make px4_sitl gz_x500
```
The changes in this PR do not change how it displays the recognized vehicle's metadata
same

### DIFF
--- a/src/AutoPilotPlugins/PX4/AirframeComponentController.cc
+++ b/src/AutoPilotPlugins/PX4/AirframeComponentController.cc
@@ -41,7 +41,7 @@ AirframeComponentController::AirframeComponentController(void) :
     
     bool autostartFound = false;
     _autostartId = getParameterFact(ParameterManager::defaultComponentId, "SYS_AUTOSTART")->rawValue().toInt();
-
+    _currentVehicleName = QString::number(_autostartId); // Temp val. Replaced with actual vehicle name if found
     
     for (int tindex = 0; tindex < AirframeComponentAirframes::get().count(); tindex++) {
 


DonLakeFlyer:
To me it would be better to show airframe id in Airframe type. And then "Unknown" in Vehicle.

gillamkid:
> To me it would be better to show airframe id in Airframe type. And then "Unknown" in Vehicle.

I looked again at various vehicle definitions in `PX4-Autopilot/ROMFS/px4fmu_common/init.d-posix/airframes`

Vehicle name seems it is  meant to be a unique identifier of a vehicle model the same way airframe id is meant to be a unique identifier of a vehicle model
ex. `Aion Robotics R1 Rover`, `PX4 Vision Dev Kit v1.5`

Vehicle type seems to be more of a categorization and refers to the general type of vehicle this vehicle model can be represented as
ex. `Quadrotor`, `Plane`, `Balloon`, `Airship`, `Rover`

I feel putting airframe ID as the vehicle name as I currently have it is most appropriate. 

note: I gave a pretty bad example screenshot photo of the current behavior (picture with yellow marker on it) I would say many of the simulation airframes were given names that are too-generic. For example in my screen shot I connected to the airframe `4001_quad_x` which has the vehicle name `Generic Quadcopter`. A better vehicle name would be `Sample Gazebo Quadcopter Sim`

DonLakeFlyer:
Yup, I agree with your reasoning

PR #13254, opened 2025-08-04
Add generic support for dynamic id based list of FactGroups

DonLakeFlyer:
* FactGroupListModel supports dynamic creation of FactGroup lists from a single mavlink message which represents multiple items. Example: multiple batteries each with their own unique id
* Convert Battery fact group to use this new model
* ToDo in further pull: Convert other fact groups to use the model like ESC status


DonLakeFlyer:
@HTRamsey @dakejahl Can you guys look at this. I only converted battery over to keep the pull simple. I'll do the others once the review is done and I merge.

dakejahl (approved):
Looks good to me. I noticed we allocate but never clean up, is there a memory leak if vehicles come and go?

HTRamsey (approved):

DonLakeFlyer:
> I noticed we allocate but never clean up, is there a memory leak if vehicles come and go?

QObject parenting should clean up when vehicle goes away. I'll double check I got that all right.

dakejahl:
> > I noticed we allocate but never clean up, is there a memory leak if vehicles come and go?
> 
> QObject parenting should clean up when vehicle goes away. I'll double check I got that all right.

Are you saying that QObject overrides `new` and allocates from a memory pool that it manages?

DonLakeFlyer:
If you set a parent on a QObject it will get deleted when the parent gets deleted.

DonLakeFlyer:
And I did screw up the parenting, which I am fixing now

PR #13256, opened 2025-08-05
fix: Gstream RTSP streaming delay

Eureak-zon:
I tried the latest version of QGC, but there is still a delay in RTSP streaming. After my verification, the problem lies in the CMake configuration.


Eureak-zon:
The current problem: After entering the url address, the video stream will get stuck. The program needs to be restarted for the video stream to function properly

PR #13259, opened 2025-08-06
Esc status indicator

DonLakeFlyer:
* Based off of #13257
* Partial replacement for #13191. Does not include Actuator page changes
* Added support to SettingsGroupLayout for specifying heading font size
* Differences from #13191 code:
  * Temperature field uses UINT16_MAX per spec for not specified, whereas previous code used < 0 which seemed incorrect
  * Online/Offline is shown with border colors. Offline text is only show next to motor number. Offline text is never shown.

Screenshot 2025-08-06 at 12 15 37 PM


DonLakeFlyer:
@dakejahl Should be easier to test with this

DonLakeFlyer:
The UI is slightly different. It was a little tricky doing things from an Esc list

dakejahl:
Thanks a ton! I will test this out today

dakejahl:
#### Testing Feedback
- Toolbar icon does not appear if ESC_STATUS / ESC_INFO mavlink messages are not received when QGC connects, even if they are received later. For example: I plug in my FC over USB and open QGC, I don't see the icon (because ESCs are not powered). I then turn on my power supply and the ESCs fire up. I see the mavlink messages coming in, but the icon never appears. If I restart QGC it's 50/50 on if the icon appears
- Float precision needs to be rounded down, maybe 2 digits
- Units should be added to the dropdown
- When a motor goes offline it always show as Motor3. This could be an issue on PX4 side, will need to check there.
- Motor names are the index, but we should do index + 1 to follow motor1-8 convention

image
image

DonLakeFlyer:
All fixed except for "When a motor goes offline it always show as Motor3. ". I hacked in 1-based motor 2 going offline after a delay and it showed correctly. So this is likely firmware side problem.

dakejahl:
It was indeed a PX4 firmware side problem. Basically every edge case on PX4 was not handled properly -- I've got a PX4 branch and I will PRing soon to fix a ton of dshot and esc related issues.

#### What I tested
- All 4 motors connected
- All 4 motors connected but any 1 of them disabled (unset actuator mapping in Actuators tab). Shows green/OK because only 3 motors are configured, but shows the offline motor in the UI since it had previously received data about that motor.
image
image

- Arbitrary ActuatorIndex --> MotorNumber mapping (ie pwm output mapping to motor number)
image


#### What I need to test
- All 4 motors connected -- physically disconnect one of the Motor wires 
- Airframe with 8 motors and all of the same tests as before

dakejahl:
@DonLakeFlyer we could merge this now or wait until I've finished the PX4 work. I am also working on adding AM32 parameters to PX4 to allow configuring ESC settings from the QGC UI via parameters. This will be facilitated on the FC side via "DShot programming mode". If you want to know more about that you can read my mavlink PR here, I've since decided to use params instead of adding to mavlink https://github.com/mavlink/mavlink/pull/2324

DonLakeFlyer:
Ok, I'll merge this is and you can report any additional bugs you find

PR #13261, opened 2025-08-06
Send buttons over 16 via buttons2 in MANUAL_CONTROL

ikalnytskyi:
Description
-----------

When a joystick has more than 16 buttons, the state of the buttons beyond the first 16 is now sent via the `MANUAL_CONTROL.buttons2` field. This ensures full button state is transmitted for joysticks with extended input capabilities.

Test Steps
-----------

Inspect MAVLink `MANUAL_CONTROL` sent from QGC or received on FC and ensure `buttons2` is properly filled.

Note: The `buttons` field is always `0` in current `master` due to custom joystick config below:

https://github.com/mavlink/qgroundcontrol/blob/f2d676939d79a66dd8081d1cb6023bea39d3e85a/cmake/CustomOptions.cmake#L38

One may want to remove the value before testing it. 

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -672,8 +672,9 @@ void Joystick::_handleAxis()
 
     emit axisValues(roll, pitch, yaw, throttle);
 
-    const uint16_t shortButtons = static_cast<uint16_t>(buttonPressedBits & 0xFFFF);
-    _activeVehicle->sendJoystickDataThreadSafe(roll, pitch, yaw, throttle, shortButtons);
+    const uint16_t lowButtons = static_cast<uint16_t>(buttonPressedBits & 0xFFFF);
+    const uint16_t highButtons = static_cast<uint16_t>((buttonPressedBits >> 16) & 0xFFFF);
+    _activeVehicle->sendJoystickDataThreadSafe(roll, pitch, yaw, throttle, lowButtons, highButtons);
 }
 
 void Joystick::startPolling(Vehicle* vehicle)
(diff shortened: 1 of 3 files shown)


HTRamsey:
Replied to your comment in the SDL3 commit, but better to mention it here for visibility. Is the SDL_GAMECONTROLLERCONFIG thing just breaking that one joystick on Linux? Somebody had provided that config saying it fixed that joystick for them, but I have no way to verify. If it's a problem can you just comment out the set(SDL_GAMECONTROLLERCONFIG... in CustomOptions.cmake in this PR too?

ikalnytskyi:
Hey @HTRamsey,

> Is the SDL_GAMECONTROLLERCONFIG thing just breaking that one joystick on Linux?

I'm using an EdgeTX Radiomaster TX12 MK II, not TX16S. However, here's the behavior I'm observing:

* When `SDL_GAMECONTROLLERCONFIG` is set to the value from the `master` branch, only the stick inputs are recognized. When I press or toggle any buttons on the remote, the corresponding button numbers are not even highlighted in QGC's Joystick UI.

* WWhen I comment out the `SDL_GAMECONTROLLERCONFIG` option, both stick and button inputs are recognized.

I'm not entirely sure why this is happening. At first glance, it seems like `SDL_GAMECONTROLLERCONFIG` is intended to override settings for a TX16S model only.

> If it's a problem can you just comment out the set(SDL_GAMECONTROLLERCONFIG... in CustomOptions.cmake in this PR too?

I'm generally not very comfortable with changes I don't fully understand. Would it be possible to merge this "as is" for now, and I can take a deeper look into the `SDL_GAMECONTROLLERCONFIG` option later today? At the very least, I'd really like to understand why my joystick was affected, even though it seemingly shouldn't have been.

HTRamsey:
I'd guess your TX12 GUID matches `0300000009120000544f000011010000`, which makes me wonder how that GUID was supposedly for the TX16S too. But yeah the stick input working makes sense because that custom controller config only provided input values for the stick input and none for buttons. I'm hoping the update to SDL3 might have fixed the issues with that joystick without needing a custom config

PR #13263, opened 2025-08-07
Unset SDL_GAMECONTROLLERCONFIG value

ikalnytskyi:
Description
-----------

The current `SDL_GAMECONTROLLERCONFIG` parameter lacks button mappings, resulting in buttons not being detected on any joystick matching its GUID. This GUID corresponds to OpenTX Radiomaster TX12 MKII, despite the joystick name in the configuration indicating it should apply to TX16S.

The origin of this GUID is unclear, but since it incorrectly matches a different device and causes button input failures that were previously absent, it is preferable to comment out this value.

This change restores expected behavior and aligns with the stable branch, where the parameter was also disabled.

Test Steps
-----------

- Run QGC.
- Navigate to Joystick setup screen.
- Click on buttons and switch toggles.
- See buttons and toggles are picked up (i.e., pressed buttons are highlighted).

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/cmake/CustomOptions.cmake
+++ b/cmake/CustomOptions.cmake
@@ -35,7 +35,7 @@ cmake_dependent_option(QGC_CUSTOM_GST_PACKAGE "Enable Using QGC Provided Custom 
 option(QGC_ENABLE_QT_VIDEOSTREAMING "Enable QtMultimedia Video Backend" OFF) # Qt6Multimedia_FOUND
 
 # Joystick
-set(SDL_GAMECONTROLLERCONFIG "0300000009120000544f000011010000,OpenTX Radiomaster TX16S Joystick,leftx:a3,lefty:a2,rightx:a0,righty:a1,platform:Linux" CACHE STRING "Custom SDL Joystick Mappings")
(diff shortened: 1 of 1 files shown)


ikalnytskyi:
Hey @HTRamsey,

This is a follow-up to what we discussed in https://github.com/mavlink/qgroundcontrol/pull/13261.

I investigated the issue further and found that my joystick's GUID is a bit different:

```
0300e14e09120000544f000011010000       -- OpenTX Radiomaster TX12 MKII
0300000009120000544f000011010000       -- SDL_GAMECONTROLLERCONFIG
```

Turns out these two bytes are the CRC checksum, which SDL ignores ([link](https://github.com/libsdl-org/SDL/blob/248bcf6b29de94e9f1d6e82e75613ec672fe851c/src/joystick/SDL_gamepad.c#L986)).

So, I don't know whether the TX16S uses the same GUID or not, but this mapping lacks buttons and clearly affects devices it never meant to.

HTRamsey:
Interesting, thank you for finding this and letting me know. At least the option can be there commented out as an example for other people to use.

PR #13278, opened 2025-08-09
Camera: Cleanup CameraManager

HTRamsey:
This should fix a use after free issue
Just stop requesting messages during the RequestMessageTest rather than deleting the manager itself.
Formatting fixes


DonLakeFlyer:
Should the prepareDelete do a camera manager stop now instead of nothing. Just to be safe?

HTRamsey:
@DonLakeFlyer so one question I have is the original code only removes one stale camera at a time in _checkForLostCameras, any reason not to iterate all of them? I guess it's just deferring the rest of the cleanup to 500ms later but I don't see why you wouldn't just perform the full loop.

PR #13302, opened 2025-08-14
PhotoVideoControl: Fix null pointer errors on camera initialization

Eureak-zon:
Add comprehensive null checks for _camera and _cameraManager to prevent
TypeError exceptions during component initialization. Guard all property
bindings and method calls with null checks and provide appropriate
fallback values.


HTRamsey:
@DonLakeFlyer should this whole element just be under a loader enabled by an existing camera?

DonLakeFlyer:
Ah, thanks. Been meaning to get to that myself. It's been making me a bit crazy.

DonLakeFlyer:
I need to noodle on this a bit. If you look at FlyViewTopRightColumnLayout.qml there is Loader code in there meant to prevent these null refs from happening. I think it used to work, but now it doesn't. Want to understand why first.

DonLakeFlyer:
So the reason the Loader isn't working is because of this:
```
void Vehicle::prepareDelete()
{
    // Clean up camera manager to stop all timers and prevent crashes during destruction
    if(_cameraManager) {
        // because of _cameraManager QML bindings check for nullptr won't work in the binding pipeline
        // the dangling pointer access will cause a runtime fault
        auto tmpCameras = _cameraManager;
        _cameraManager = nullptr;
        delete tmpCameras;
        emit cameraManagerChanged();
        // Note: Removed qApp->processEvents() to prevent MAVLink crashes during destruction
    }
}
```

If still thinking about whether there is a way to deal with this without needing to put null checks everywhere which just makes the code grungy.

DonLakeFlyer:
@julianoes Hey Julian, do you have any idea why Qml binding would be causing crashed in CameraManager code. If I remove this prepareDelete things seems to work fine for me and we no longer need nulls checked everywhere since the Loader which controls PhotoVideoControl from coming going on active vehicle gets rid of it before the camera manager on the vehicle goes away.

@HTRamsey Not sure if you have any thoughts here as well?

HTRamsey:
That's something I kinda address in https://github.com/mavlink/qgroundcontrol/pull/13278. I don't personally think the prepare delete is necessary, but I made adjustments to prevent issues in unit tests in that PR

julianoes:
I think you two @DonLakeFlyer and @HTRamsey have a much better grasp on this than me. Sorry for use after free.

I remember the issue I was trying to fix was disconnecting and reconnecting cameras. I can re-test after the fixes have gone in.

DonLakeFlyer:
Replaced by my pull.

PR #13309, opened 2025-08-15
Fix mavlink inspector crash

vryabokon1705:
Fix MAVLink Inspector crash

Description
-----------
Fixes unexpected crash when charts plotting selected in Mavlink Inspector and user selected some other "Analize Tools" page

Test Steps
-----------
- Connect to vehicle
- Open MAVLink Inspector
- Select MAVLink message ( fe HIGHRES_IMU)
- Plot values in different plots
- Click MAVLink console
- Repeat muiltiple times

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/src/AnalyzeView/MAVLinkInspectorController.cc
+++ b/src/AnalyzeView/MAVLinkInspectorController.cc
@@ -7,310 +7,315 @@
  *
  ****************************************************************************/
 
-#include "MAVLinkInspectorController.h"
-#include "MAVLinkChartController.h"
-#include "MAVLinkMessage.h"
-#include "MAVLinkProtocol.h"
-#include "MAVLinkSystem.h"
-#include "MultiVehicleManager.h"
-#include "QGCApplication.h"
-#include "QGCLoggingCategory.h"
-#include "QmlObjectListModel.h"
-#include "Vehicle.h"
+ #include "MAVLinkInspectorController.h"
+ #include "MAVLinkChartController.h"
+ #include "MAVLinkMessage.h"
+ #include "MAVLinkProtocol.h"
+ #include "MAVLinkSystem.h"
+ #include "MultiVehicleManager.h"
+ #include "QGCApplication.h"
+ #include "QGCLoggingCategory.h"
+ #include "QmlObjectListModel.h"
+ #include "Vehicle.h"
 
-#include <QtQml/QQmlEngine>
+ #include <QtQml/QQmlEngine>
 
-QGC_LOGGING_CATEGORY(MAVLinkInspectorControllerLog, "qgc.analyzeview.mavlinkinspectorcontroller")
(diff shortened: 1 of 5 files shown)


src/AnalyzeView/MAVLinkInspectorController.cc
```diff
@@ -84,7 +84,12 @@ MAVLinkInspectorController::~MAVLinkInspectorController()
 {
     qDeleteAll(_timeScaleSt);
     qDeleteAll(_rangeSt);
-    _charts->clearAndDeleteContents();
+
+    for (int i = _charts->count() - 1; i >= 0; i--) {
+        auto chartObject = _charts->get(i);
+        MAVLinkChartController* chart = dynamic_cast<MAVLinkChartController*>(chartObject);
+        deleteChart(chart);
```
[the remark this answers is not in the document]
vryabokon1705: nullpointer checked in the deleteChart method:

```
void MAVLinkInspectorController::deleteChart(MAVLinkChartController *chart)
{
    if (!chart) {
        return;
    }
```
So I do not think we need this one another check

src/QmlControls/MAVLinkChart.qml
```diff
@@ -23,9 +23,9 @@ ChartView {
     property var chartController:   null
     property var seriesColors:      ["#00E04B","#DE8500","#F32836","#BFBFBF","#536DFF","#EECC44"]
```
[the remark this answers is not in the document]
vryabokon1705: MAVLinkChart.qml does not have documentation for other methods. As well as other qml files

src/AnalyzeView/MAVLinkInspectorController.cc
```diff
@@ -84,7 +84,12 @@ MAVLinkInspectorController::~MAVLinkInspectorController()
...
+    }
```
DonLakeFlyer, line 92: I don't understand how this fixes a problem? As far as I can tell it ends up being functionally the same thing. The only diiference is that chartsChanged is not signalled. But given the fact this is happening during destruction that should be needed.

vryabokon1705: It helps to prevent trying to draw the chart, which was already deleted, but the list has not been clear yet

DonLakeFlyer: How, given the fact that it is functionally equivalent to the previous code? Is it related to the chartsChanged signalling?

vryabokon1705:
```
void QmlObjectListModel::clearAndDeleteContents()
{
    for (int i=0; i<_objectList.count(); i++) {
        _objectList[i]->deleteLater();
    }
    clear();
}
```
In this method, we first iterate over the list and delete objects, but pointers still exist. And during this process,  before _objectList is cleared, we can try to draw the chart, which can lead to a null pointer exception

PR #13319, opened 2025-08-20
Fix entry point rotation when single transect

DonLakeFlyer:
* Also cleaned entry point code to be clearer as to how it works
* Replacement for #12901


DonLakeFlyer:
@mrschick Can you review this?

PR #13328, opened 2025-08-22
Improve flight path calculation with loiter radius support

Eureak-zon:
@DonLakeFlyer  Could you help me review the code? I don't think this implementation is the best at present.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,59 @@
+{
+    "version": 6,
+    "cmakeMinimumRequired": {
+        "major": 3,
+        "minor": 21,
+        "patch": 3
+    },
+    "configurePresets": [
+        {
+            "name": "Windows",
+            "hidden": true,
+            "displayName": "Windows configuration",
+            "generator": "Ninja",
+            "binaryDir": "${sourceDir}/build/qt6-Windows",
+            "architecture": {
+                "value": "x64",
+                "strategy": "external"
+            },
+            "cacheVariables": {
+                "CMAKE_C_COMPILER": "cl",
+                "CMAKE_CXX_COMPILER": "cl",
+                "QT_ROOT_DIR": "D:/Devoloper/Qt/6.8.3/msvc2022_64"
+            },
+            "condition": {
+                "type": "equals",
+                "lhs": "${hostSystemName}",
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
I assume the cmake file in the pull is a mistake?

I've been playing around with this using PX4 SITL. My concern is that this kinda just draws flight paths which differ from  reality in a different wrong way. 

For example with PX4:
* When flying to a loiter point the initial flight path is not to a tangent to a loiter but to the center of the loiter. Only when it gets close does it diverge towards the tangent. This new change doesn't show this correctly. It draws from the previous waypoint to the tangent.
* When flying away from a loiter point I think it always flies away from the tangent direct to the next waypoint which differs from entry. This change does draw this correctly.

Given that, I'm not sure this is really better. Just differently wrong as far as I can tell.

I'm also unsure whether ArduPilot actual flight path differs from how PX4 does it? The real flight path might be firmware specific. @rubenp02?

DonLakeFlyer:
So I guess after playing around with ti I agree with what @rubenp02 said in the original pull:
"I think it's problematic because the path visualizations don't match what the aircraft will fly which I think is very critical for obstacle avoidance and such. Maybe keep the improved calculations and revert the map line drawing to how it was before, or adjust the improved one to be correct."

Eureak-zon:
@DonLakeFlyer CMakePresets.json is a bug commit, please ignore it

Eureak-zon:
My idea is that in the circling mode, the ideal flight path should be accurately plotted, while the actual flight path is determined by the flight control system itself. After all, the algorithms of each simulator are different.

DonLakeFlyer:
What is the actual flight path for ArduPilot when coming in to and exiting a Loiter?

rubenp02:
> What is the actual flight path for ArduPilot when coming in to and exiting a Loiter?

Entry is straight towards the center until it gets "captured" and starts loitering. Exit can be perpendicular or tangent to the loiter path.

DonLakeFlyer:
Ok, so as can be seen from how entry/exit is firmware specific and also not always predictable I don't think this is possible to make any better than it currently is. Yes it's not correct. But it doesn't really try to be correct in a wrong way which I think is better.

PR #13330, opened 2025-08-22
Comms: Fix TLOG Permissions

### LINKED ISSUE #13250: Cannot share telemetry files after installing QGC 5.0 for Android

### Description

I have recently installed the latest QGC 5.0 stable version on a Moto Neo 50 running Android 15, but I cannot share telemetry (*.tlog) files. I tried several exploreres with  no success, USB over pc, update do drive and adb.

I can share files at the same folder (android/data) of other apps (e.g Mission Planner), but not with QGC 5.0 stable.

### System Information

Latest 5.0 stable

### Log Files and Screenshots

_No response_

HTRamsey:
Fixes #13250
Uses a QSaveFile and sets file permissions
Also need to look into Android Permissions changes...


DonLakeFlyer:
Why is tlog different than any other file we save? Don't see why it doesn't get default permissions.

HTRamsey:
It's due to the QGCTemporaryFile I think

PR #13354, opened 2025-09-04
ArduPilot: Remove hardcoded flight mode names

### LINKED ISSUE #13224: Vehicle mode checking issues

### Description

When running APM master with QGC master, the mode names are received in all caps, breaking any mode checking logic that relies on strict string comparisons. This seems to depend on the firmware version used and it's not the first time it breaks, so a more robust way to compare modes should probably be implemented throughout QGC.

### System Information

QGC and APM master (SITL) at the time of creating this issue.

### Log Files and Screenshots

_No response_

DonLakeFlyer:
Fix #13224 

@rubenp02 Can you verify this works for you?


rubenp02:
This fixes the issue! Thank you.

PR #13359, opened 2025-09-05
Delete all current translations file so we can take a reset on them

DonLakeFlyer:


HTRamsey:
FYI there's a runnable workflow that creates a PR with the lupdate/lupgrade stuff

PR #13369, opened 2025-09-09
Fix CI build for macos by removing cmake installation from brew

ddatsko:
Fix CI build for MacOS by removing cmake installation from brew

Description
-----------
I noticed that recently all the github actions building for MacOS are failing at the moment of `.dmg` file creation. By deeper inspection I have found that in the `Install Dependencies` stage we run `brew install cmake ... create-dmg`. This command fails (see the screenshot from the latest MacOS build action from master attached below) as cmake is already installed from `local/pinned tap`. 
image

Thus `create-dmg` (as well as some other packages) is not installed leading to cmake trying to install the `create-dmg` in `CreateMacDMG.cmake`, which fails. I haven't checked why the `CPMAddPackage` fails there, but I guess, this is an issue with CMake version 4.4.1.

By removing the `cmake` from `brew install` command, rest of the listed packages get installed, and workflow finishes successfully.

Test Steps
-----------
Trigger MacOS build github action in master branch

Checklist:
----------

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -16,6 +16,7 @@ on:
       - 'src/**'
       - 'CMakeLists.txt'
       - 'cmake/**'
+      - 'tools/setup/install-dependencies-osx.sh'
 
 # concurrency:
 #   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


HTRamsey:
Yeah I believe this started with cmake 4.1.1 but I'd rather not remove cmake from the build dependencies script. 

Actually figured it out, don't use FetchContent in the install phase for macos

Issue #13376, opened 2025-09-12
Possible Bug in RemoteIDManager relating to GCS GPS being invalid with negative altitude


jnomikos:
### Have you checked the latest release for fixes?

- [x] Yes, I’ve tested on the latest release

### Description

**Expected Behavior**
On RemoteIDManager, it has been noticed on an android GCS that GCS GPS occasionally flashed between Green and Red. I believe the reason is because sometimes the GPS on the Ground Control Station is reporting a negative altitude, and based on the code [here](https://github.com/mavlink/qgroundcontrol/blob/2bd777f266a1bd3d494e2d73de2dba9f43971730/src/Vehicle/RemoteIDManager.cc#L299C1-L306C14), a negative altitude is determined to be invalid. However, I suspect that negative altitude can normally happen and doesn't necessarily mean the altitude is invalid. What if you are at an altitude below sea level?

**Current Behavior**
GCS GPS goes red if altitude is negative

**Steps To Reproduce**
1. Enable Remote ID
2. Have an altitude below zero

### System Information

QGC 4.4.4
Android
Cube Blue
ArduCopter 4.5.6

jnomikos:
I think GCS GPS altitude would be NAN if unset

jnomikos:
Resolved here: #13378

PR #13378, opened 2025-09-12
Remote ID: Fix zero and negative GCS GPS altitude being invalid

jnomikos:
When the GCS GPS altitude is 0 or negative, GCS GPS validity will flash between valid and invalid. This is not correct. Altitude can be 0 or negative since it is measured as altitude above sea level (I think). ~It should be a NAN check instead.
Made GCS GPS validity not check if altitude is <= 0. Instead, check if it's not NAN.~

**Edit: Ensures that GCS GPS position data is 3D rather than doing incorrect altitude check which marks valid altitudes of 0 and negatives as invalid.

Also, now tracks last GCS GPS position error and marks GCS GPS data as invalid if there's a positioning error.**

Test Steps
-----------
Run normal QGC with GCS that has negative altitude or 0 and witness GCS GPS validity flash between valid and invalid. With this PR, witness that not happen.

Checklist:
----------

Related Issue
-----------
https://github.com/mavlink/qgroundcontrol/issues/13376

-- 

Note that on my PC which doesn't have GPS, I had lat, lon, and alt show up as zero and GPS was still valid, so we might want to add some other way of detecting if alt is emitted or not.

### DIFF
--- a/src/Vehicle/RemoteIDManager.cc
+++ b/src/Vehicle/RemoteIDManager.cc
@@ -258,13 +258,16 @@ void RemoteIDManager::_sendOperatorID()
 
 void RemoteIDManager::_updateGcsGpsStatus(bool gpsGood, const QString& error)
 {
+    if (!error.isEmpty() && _gcsGPSError != error) {
+        _gcsGPSError = error;
+        qCWarning(RemoteIDManagerLog) << "GCS GPS error:" << error;
+    }
     if (_gcsGPSGood != gpsGood) {
         _gcsGPSGood = gpsGood;
-        emit gcsGPSGoodChanged();
-        if (!error.isEmpty() && _gcsGPSError != error) {
-            _gcsGPSError = error;
-            qCWarning(RemoteIDManagerLog) << "GCS GPS error:" << error;
+        if (gpsGood) {
+            _gcsGPSError.clear();
         }
+        emit gcsGPSGoodChanged();
     }
 }
 
--- a/src/PositionManager/PositionManager.cpp
+++ b/src/PositionManager/PositionManager.cpp
@@ -207,7 +207,6 @@ void QGCPositionManager::_setPositionSource(QGCPositionSource source)
             _currentSource->setUpdateInterval(_updateInterval);
         #endif
 
-
         (void) connect(_currentSource, &QGeoPositionInfoSource::positionUpdated, this, &QGCPositionManager::_positionUpdated);
         (void) connect(_currentSource, &QGeoPositionInfoSource::errorOccurred, this, &QGCPositionManager::_positionError);
 


jnomikos:
Note: I still see GPS GCS flashing between invalid and valid, will investigate further

Edit: was due to logic issue fixed below

jnomikos:
There was a logic issue in my fix. qIsNan logic was reversed, oops.

src/Vehicle/RemoteIDManager.cc
```diff
@@ -298,7 +298,7 @@ void RemoteIDManager::_sendSystem()
         // GPS position needs to be valid before checking other stuff
         if (geoPositionInfo.isValid()) {
             // If we dont have altitude for FAA then the GPS data is no good
-            if ((_settings->region()->rawValue().toInt() == Region::FAA) && !(gcsPosition.altitude() >= 0) && _gcsGPSGood) {
+            if ((_settings->region()->rawValue().toInt() == Region::FAA) && qIsNaN(gcsPosition.altitude()) && _gcsGPSGood) {
```
DonLakeFlyer: The NaN check makes sense except for the fact that PositionManager will never set altitude to NaN. The reason your code remove the flicker is because of this fact. Or in other words `qIsNaN(gcsPosition.altitude()` will always be false.

The PositionManager code needs to be updated to use NaN for value not available for lat/lon/altitude/heading. And then all the code that references PositionManager gcs position/heading needs to be checked to handle unavailable values.

Can you take a look at doing that as well?

jnomikos: Thanks for the info, will look into it. Might be in a few weeks because I have a lot on my plate and will be moving all next week

Davidsastresas:
Thanks for the review Don. Maybe we are better using a feasible range for GCS altitude? like (-1000) - 10000 or something like that to be on the safe side, in case for some reason we were receiving odd data from the GPS?

jnomikos:
> Thanks for the review Don. Maybe we are better using a feasible range for GCS altitude? like (-1000) - 10000 or something like that to be on the safe side, in case for some reason we were receiving odd data from the GPS?

Yeah, right now [we simply debug when there's an error from QGeoPositionInfoSource](https://github.com/mavlink/qgroundcontrol/blob/ebc3cd27564318a69dce76975e1587471d09f276/src/PositionManager/PositionManager.cpp#L214-L216). I think we should instead handle the error and base validity on that.

DonLakeFlyer:
Updates on this?

jnomikos:
> Updates on this?

Working on it now.

src/Vehicle/RemoteIDManager.cc
jnomikos: I ended up writing it so that, rather than utilizing NaN, it simply ensures that the position data is 3D in the validity check. I also refactored some of the functions, and added an additional check that tracks positioning errors

Issue #13377, opened 2025-09-12
RemoteIDManager: Bug where Operator ID becomes invalid after vehicle reboot


jnomikos:
### Have you checked the latest release for fixes?

- [x] Yes, I’ve tested on the latest release

### Description

**Expected Behavior**
I expect Operator ID to stay valid when rebooting the vehicle.

**Current Behavior**
On vehicle reboot, for a split second when the vehicle reconnects, onTextChanged is detected with empty text "" which causes invalid operator id to be set.

**Steps To Reproduce**
1. Have vehicle with RID module
2. Enable Remote ID
3. Set an operator ID and have it broadcast
4. Reboot the vehicle in param window
5. Witness Operator ID become invalid on reconnect

**Additional Details**
Add any extra info (errors, warnings, etc.) here.

### System Information

Latest Stable 4.5 (also witnessed it on 4.4
Stable
ArduCopter 4.5.6

jnomikos:
It isn't just a QML bug. There's a validity check in the constructor which controls if operator id is good. If the operator ID is invalid in EU jurisdiction but we're FAA, it'll turn invalid on vehicle reboot or app start.

I will make a PR for this

PR #13379, opened 2025-09-12
Remote ID FAA: Operator ID invalid on reboot or app restart

jnomikos:
Description
-----------
1. If in FAA jurisdiction, operator ID will be invalidated on app restart or vehicle reboot (if not correct under EU jurisdiction) because there's a validity check in the constructor which checks if the operator ID is valid under EU jurisdiction even if in FAA.

2. If vehicle is rebooting and you are in the Remote ID app settings page, the quick refresh of the page is detected as a text field change and can be detected as operator ID being changed. Make it so it tracks if the text field is being focused.

Test Steps
-----------
1. Have remote ID enabled
2. Make sure FAA is the jurisdiction and assign an operator ID that would be invalid in EU jurisdiction like "John"
3. Either reboot the vehicle in param page or close app and reopen
4. Witness operator ID stay valid
5. Next, reboot vehicle in param page and go to remote id page as vehicle loads back up
6. Ensure operator id doesn't change

Checklist:
----------

Related Issue
-----------
https://github.com/mavlink/qgroundcontrol/issues/13377

### DIFF
--- a/src/UI/AppSettings/RemoteIDSettings.qml
+++ b/src/UI/AppSettings/RemoteIDSettings.qml
@@ -375,6 +375,10 @@ SettingsPage {
                         }
                         
                         onTextChanged: {
+                            if (!activeFocus) {
+                                return;
+                            }
+
                             operatorIDFact.value = text
                             if (_activeVehicle) {
                                 _remoteIDManager.checkOperatorID(text)
--- a/src/Vehicle/RemoteIDManager.cc
+++ b/src/Vehicle/RemoteIDManager.cc
@@ -66,11 +66,11 @@ RemoteIDManager::RemoteIDManager(Vehicle* vehicle)
     _targetSystem = _vehicle->id();
     _targetComponent = _vehicle->compId();
 
-    if (_settings->operatorIDValid()->rawValue() == true) {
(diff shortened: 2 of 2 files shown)


src/UI/AppSettings/RemoteIDSettings.qml
```diff
@@ -375,6 +375,10 @@ SettingsPage {
                         }
                         
                         onTextChanged: {
+                            if (!activeFocus) {
```
DonLakeFlyer: So I don't quite get why this is needed. If you look at `checkOperatorId`:
```
void RemoteIDManager::checkOperatorID(const QString& operatorID)
{
    // We overwrite the fact that is also set by the text input but we want to update
    // after every letter rather than when editing is done.
    // We check whether it actually changed to avoid triggering this on startup.
    if (operatorID != _settings->operatorID()->rawValueString()) {
        _settings->operatorIDValid()->setRawValue(_isEUOperatorIDValid(operatorID));
    }
}
```

It seems like the `!=` check would prevent anything from happening?

jnomikos: Good catch. You are correct. I'll remove that change

DonLakeFlyer:
@jnomikos Is this important enough to cherry pick to stable as well?

jnomikos:
> @jnomikos Is this important enough to cherry pick to stable as well?

In my opinion, it isn't important enough to cherry pick since operator ID is technically not required by FAA.

DonLakeFlyer:
OK, thanks

PR #13382, opened 2025-09-14
CMake: Fix MacOS Create DMG

HTRamsey:
Don't use FetchContent in install mode


HTRamsey:
@DonLakeFlyer issue with AGL on latest MacOS: https://developer.apple.com/documentation/macos-release-notes/macos-26-release-notes#Deprecations
I've had to force a maximum XCode version

PR #13383, opened 2025-09-14
CMake: Adjust MacOS DMG Install

HTRamsey:
accidentally was overriding the cpm package

### DIFF
--- a/cmake/install/CreateMacDMG.cmake
+++ b/cmake/install/CreateMacDMG.cmake
@@ -1,9 +1,8 @@
 # -- Variables ----------------------------------------
 # QGC_STAGING_BUNDLE_PATH => full path to MyApp.app
+# CREATE_DMG_PROGRAM => full path to create-dmg program
+# ---------------------------------------------------------------------------
 
-# ---------------------------------------------------------------------------
-# 1. Build the DMG with a nice drag-and-drop layout
-# ---------------------------------------------------------------------------
 set(QGC_DMG_PATH "${CMAKE_BINARY_DIR}/package")
 file(REMOVE_RECURSE "${QGC_DMG_PATH}")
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
You know my opinion: Doing this multiple way just leads to more support nightmares and optional things that break since they are not using in CI. create-dmg is installed by brew in the osx install script. Which in turn puts it on the path. Why do we need any more than what stable had:

```
message(STATUS "Creating DMG: ${TARGET_APP_NAME}.dmg")
execute_process(
    COMMAND create-dmg --volname "${TARGET_APP_NAME}" --filesystem "APFS" "${TARGET_APP_NAME}.dmg" "${CMAKE_BINARY_DIR}/package/"
    COMMAND_ERROR_IS_FATAL ANY
)
```

cmake/install/CreateMacDMG.cmake
```diff
@@ -13,7 +12,6 @@ cmake_path(GET QGC_STAGING_BUNDLE_PATH STEM QGC_TARGET_APP_NAME)
 
 set(QGC_DMG_NAME "${QGC_TARGET_APP_NAME}.dmg")
 
-find_program(CREATE_DMG_PROGRAM create-dmg REQUIRED)
```
DonLakeFlyer, line 16: Also this change means that CREATE_DMG_PROGRAM is never set even though the one installed by brew should be there

HTRamsey: it's set in Install.cmake, where it determines whether it can find create-dmg from brew, or has to download it.

HTRamsey:
Nothing has changed except for moving the part that finds create-dmg, which was required in CMake 4.1.x, so there aren't any new options or anything.

PR #13400, opened 2025-09-17
Sign, Notarize and Staple Mac app bundle

DonLakeFlyer:
Jumping through the Mac security hoops to use QGC is no longer needed


DonLakeFlyer:
The signing is working now. But now I need add a notarize step.

PR #13413, opened 2025-09-21
Clean up mavlink inspector code to prevent crashing

DonLakeFlyer:
* Remove dangling references to chart controllers
* Also modified to improve readability
* Possible replacement for #13309

### DIFF
--- a/src/AnalyzeView/MAVLinkChartController.cc
+++ b/src/AnalyzeView/MAVLinkChartController.cc
@@ -18,12 +18,10 @@
 
 Q_DECLARE_METATYPE(QAbstractSeries*)
 
-QGC_LOGGING_CATEGORY(MAVLinkChartControllerLog, "qgc.analyzeview.mavlinkchartcontroller")
+QGC_LOGGING_CATEGORY(MAVLinkChartControllerLog, "MAVLinkChartControllerLog")
 
-MAVLinkChartController::MAVLinkChartController(MAVLinkInspectorController *controller, int index, QObject *parent)
+MAVLinkChartController::MAVLinkChartController(QObject *parent)
     : QObject(parent)
-    , _index(index)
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
@vryabokon1705 Can you take a look at this? Both from a review standpoint as well as a testing and trying to repro the crash standpoint? The previous existing code for chart controllers was just a huge mess. There was no reason I could find the need to manage these dynamically. They should just work like other c++ controllers and be directly associated/created with the control. I believe this removes the possibility of crashing on in correct chart controller referencing. But I was not able to reproduce the crash you were seeing although I could understand how it could happen.

That said this charting code is still a big mess and this only fixes one aspect of it. The fact that it tries to support multi-vehicle and doesn't correctly deal with vehicles going away will cause other crashes in those cases. This code needs further modification to remove support for multi-vehicle since it's hopeless to try to fix that given all the reach-around to other objects it has embedded in it. I'm going to deal with that in another pull.

DonLakeFlyer:
@vryabokon1705 When you saw it crashing, was it referencing a chart controller or was it crashing on a reference to a message field?

vryabokon1705:
> When you saw it crashing, was it referencing a chart controller or was it crashing on a reference to a message field?

Hi. Looks like it referencing a chart controller

vryabokon1705:
> Can you take a look at this

Ok, let I try it

vryabokon1705:
Hi. I am not able to reproduce the crash issue with this branch. It is Ok from my point of view.

I uploaded video to show how you can reproduce it on the master branch.

https://github.com/user-attachments/assets/7915d177-277e-4f91-8a9e-47f627b9248e

DonLakeFlyer:
@vryabokon1705 Thanks so much for the help on this

PR #13443, opened 2025-09-28
Rework logging categories to double level of hierarchy

DonLakeFlyer:
* Support wildcarding of parent hierarchies. Example: Vehicle.*
* Major rework on category naming to better fit new double level model
* Use class names as base category as much as possible to simplify going from source to logging category
* Rework category setting ui dialog for new model

UI samples:
Screenshot 2025-09-28 at 12 15 05 PM
Screenshot 2025-09-28 at 12 15 20 PM
Screenshot 2025-09-28 at 12 15 29 PM

### DIFF
--- a/custom-example/src/CustomPlugin.cc
+++ b/custom-example/src/CustomPlugin.cc
@@ -19,7 +19,7 @@
 #include <QtQml/QQmlApplicationEngine>
 #include <QtQml/QQmlFile>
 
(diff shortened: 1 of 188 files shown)


HTRamsey:
Very nice. Might also be useful to add some of the built in Qt categories to the toggle options in the UI. I forget what they are off the top of my head but there's one for like QML, audio, video, etc

PR #13476, opened 2025-10-03
remove warning about calibration over wifi

dakejahl:
I'm not sure if this warning is relevant anymore. It's very common to connect to QGC over the network with mavlink-router running onboard a companion computer. I've never had an issue with calibration over wifi and this warning has caused confusion for some users.

### DIFF
--- a/custom-example/src/CustomPlugin.h
+++ b/custom-example/src/CustomPlugin.h
@@ -49,10 +49,6 @@ public:
 
     // Overrides from QGCOptions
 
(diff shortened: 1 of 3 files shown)


HTRamsey:
Seems like this was specifically for stuff like the esp 8266, which I can see having unreliable wifi but I've never tried it. At a minimum the default could be true with the ability to change it to false in a custom build. @DonLakeFlyer thoughts?

DonLakeFlyer:
Yup, ancient history

PR #13488, opened 2025-10-08
Show joystick calibration

Niki-dev12:
Show calibrated joystick axis output in calibration screen
-----------

Description
-----------
This update chnages the joystick calibration UI by displaying real-time calibrated axis output values for Roll, Pitch, Yaw, and Throttle during and after calibration.
It helps verify that calibration is applied correctly.

Key changes
-----------

- Added dynamic axis visualization (AxisMonitor) for each joystick channel.
- Implemented helper functions (joyToAxis, joyToThrottleAxis) to apply clamping and mapping to QGC’s internal range
- Integrated calibration reversal logic so UI reflects post-calibration orientation.
- Updated Connections.onAxisValues to respond immediately to calibration changes.

Test Steps
-----------

- Open Joystick Setup Calibration in QGroundControl.
- Move Roll, Pitch, Yaw, and Throttle sticks after calibration.
- Confirm that each Axis Monitor bar responds smoothly and in the correct direction.
- Toggle Reverse on any axis and ensure the visualization flips 
- Verify that throttle range maps correctly from 0–1 to –32768..32768.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/src/Vehicle/VehicleSetup/JoystickConfigGeneral.qml
+++ b/src/Vehicle/VehicleSetup/JoystickConfigGeneral.qml
@@ -13,11 +13,11 @@ import QtQuick.Dialogs
 import QtQuick.Layouts
 
 import QGroundControl
-
+import QGroundControl.Palette
 import QGroundControl.Controls
-
-
-
+import QGroundControl.ScreenTools
+import QGroundControl.Controllers
+import QGroundControl.FactSystem
 import QGroundControl.FactControls
 
 Item {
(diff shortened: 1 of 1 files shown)


HTRamsey:
Looks like you removed the 0:1 range for thrust?

DonLakeFlyer:
@Niki-dev12 Just making sure I"m not missing something here. The only change here is that the channel display bars respect the reversed attribute right? Other than that they work exactly the same?

Niki-dev12:
> @Niki-dev12 Just making sure I"m not missing something here. The only change here is that the channel display bars respect the reversed attribute right? Other than that they work exactly the same?

Before QML was showing **raw joystick signal** with a fixed formula, so any **calibration changes (invert, min/max, center, deadzone, remap)** applied in the controller layer did not affect what the QML drew. QGC was visualizing the _pre-calibration_ values.

Niki-dev12:
> Looks like you removed the 0:1 range for thrust?

The input range (0..1) is still assumed, but the polarity flipped compared to the old non negative thrust path.

PR #13498, opened 2025-10-09
Fix: Windows executable missing application icon (Resolves #13474)

### LINKED ISSUE #13474: Windows: Application icon is missing on builds from recent master branch

### Description

Expected Behavior
The application executable (.exe) should display the default QGroundControl icon after being built from source.

Current Behavior
The application icon is blank on the final .exe file. This affects the icon in Windows Explorer, the taskbar, and the window's title bar.

Steps To Reproduce

Clone the master branch of QGroundControl.

Do not modify any source code.

Configure and build the project using the toolchain specified in the System Information section below.

Check the icon of the resulting qgroundcontrol.exe file.

Additional Details

As a point of comparison, I tried building the `v5.0.7` stable tag using the exact same build environment. That build was successful, and the application icon displayed correctly. **Furthermore, I have confirmed that this issue with the `master` branch is reproducible on a second, separate computer.** These facts lead me to believe the issue is not with my PC or a specific environment setup.

My personal suspicion is that the problem lies within the CMake build scripts, which seem to have undergone significant changes between v5.0.7 and the current master. I am not an expert in CMake structure, so I was unable to pinpoint the exact source of the regression.

One significant behavioral difference I observed is that in the working v5.0.7 build, a deploy/windows directory is created inside the build directory. This directory is missing when building the master branch. I also attempted to rule out a corrupted icon file by replacing the .ico file with different ones (including the original QGC icon), but this had no effect.

### System Information

QGC Version: master branch at commit [c8c27880ea97a031fd6953ac0112c116e97df286]

QGC build: daily

Operating System: Windows 11 Home 24H2

Flight Controller: N/A (This is a build system issue, not related to flight hardware)

Autopilot Software (with version): N/A

### Log Files and Screenshots

Image

Image

DRT-Jin:
Fix: Windows executable missing application icon (Resolves #13474)

Description
-----------
This PR resolves the regression where the application icon is blank on Windows executables built from the master branch, as reported in issue #13474.

**Background:**
The issue occurred despite `QT_TARGET_RC_ICONS` being set. The core problem appears to be that the logic explicitly linking the Windows resource (`.rc`) file was removed in a previous CMake refactoring. The Qt 6 CMake module **may not be reliably triggering** the compilation and linking of the resource file in the current project structure, **which seems to result** in the icon being omitted from the final EXE.

**Solution:**
The fix is to explicitly force the compilation and linking of the resource file in `cmake/platform/Windows.cmake`.
1. Added `target_sources(${CMAKE_PROJECT_NAME} PRIVATE "${QGC_WINDOWS_RESOURCE_FILE_PATH}")` to directly include the RC file in the target's sources.
2. Ensured `QT_TARGET_WINDOWS_RC_FILE` is set to maintain compatibility with other resource-based metadata handling.

This approach robustly ensures the icon is embedded by the Resource Compiler (rc.exe).

Test Steps
-----------
1. Clone and configure the latest master branch.
2. Build the project for Windows (e.g., using Ninja or Visual Studio).
3. Check the resulting `qgroundcontrol.exe` file in Windows Explorer, the taskbar, or the window title.
    * **Before Fix:** The application icon is blank.
    * **After Fix:** The correct QGroundControl application icon is displayed.
    
Checklist:
----------

Related Issue
-----------
#13474


HTRamsey:
Doesn't setting QT_TARGET_WINDOWS_RC_FILE override all the other variables? Looking in the Qt CMake code for this part, I think it's better to use the individual QT_TARGET variables. Just need to fix the QT_TARGET_RC_ICONS part I guess

DRT-Jin:
Agreed. The proper solution is to find and fix the root cause of the QT_TARGET_RC_ICONS failure.

For now, I've implemented a reliable workaround by manually including the .rc file to ensure everything functions correctly. I plan to keep investigating the root cause within the Qt CMake logic.

If you have any specific ideas or insights regarding this failure, please share them so I can prioritize my investigation.guess

HTRamsey:
Looking at the [v5.0 Stable part](https://github.com/mavlink/qgroundcontrol/blob/fc5461b828b648135fdf9e98fa52b1ff3a9045bb/CMakeLists.txt#L333), you probably need to just do what you did for the RC file to the Icon File instead. So adding the icon file to target_sources or something like that

HTRamsey:
Figured it out. Qt wanted to use CMake's RC generation so now you have to manually call _qt_internal_generate_win32_rc_file

PR #13507, opened 2025-10-10
CMake: Fix Windows Icon

### LINKED ISSUE #13474: Windows: Application icon is missing on builds from recent master branch

HTRamsey:
Apparently _qt_internal_generate_win32_rc_file has to be called manually. Qt wanted to use CMake's upstream RC generation but it doesn't exist

Closes #13474

### DIFF
--- a/cmake/platform/Windows.cmake
+++ b/cmake/platform/Windows.cmake
@@ -3,19 +3,35 @@ if(NOT WIN32)
     return()
 endif()
 
-set_target_properties(${CMAKE_PROJECT_NAME}
-    PROPERTIES
-        WIN32_EXECUTABLE TRUE
-        # QT_TARGET_WINDOWS_RC_FILE "${QGC_WINDOWS_RESOURCE_FILE_PATH}"
-        QT_TARGET_COMPANY_NAME "${QGC_ORG_NAME}"
-        QT_TARGET_DESCRIPTION "${CMAKE_PROJECT_DESCRIPTION}"
-        QT_TARGET_VERSION "${CMAKE_PROJECT_VERSION}"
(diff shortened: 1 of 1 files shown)


DRT-Jin:
That's a perfect solution. Thank you! I've learned a lot thanks to you

PR #13512, opened 2025-10-11
PX4: Remove RC Throttle Trim Hotfix

HTRamsey:
Replaces #10497
See: https://github.com/PX4/PX4-Autopilot/pull/15949


HTRamsey:
Nevermind, looks like both ArduPilot and PX4 handle this correctly now

PR #13517, opened 2025-10-12
Add gimbal support to joystick interface

ikalnytskyi:
Description
-----------

Extend joystick interface to support gimbal pitch and yaw controls on axes 4 and 5. Gimbal values are transmitted via `MANUAL_CONTROL` extension fields with flags indicating axis presence. Gimbal calibration steps are automatically skipped if the axes are not detected on the connected joystick.

Test Steps
-----------

1. Connect a joystick with 4 axes.
2. Open the joystick calibration screen.
3. Verify that only 4 axes are calibrated and that the calibration completes successfully.
4. Connect a joystick with 6 axes.
5. Open the joystick calibration screen again.
6. Verify that, in addition to the 4 main axes, 2 optional axes are also calibrated.
7. Inspect the transmitted MAVLink `MANUAL_CONTROL` message:
8. The optional axes must be sent via the `s` and `t` fields.
9. Their presence must be indicated in the `enabled_extensions` field.

Checklist:
----------

Related Issue
-----------

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -608,13 +608,13 @@ void Joystick::_handleAxis()
     axis = _rgFunctionAxis[throttleFunction];
     float throttle = _adjustRange(_rgAxisValues[axis], _rgCalibration[axis], (_throttleMode == ThrottleModeDownZero) ? false :_deadband);
 
-    float gimbalPitch = 0.0f;
+    float gimbalPitch = NAN;
     if (_axisCount > 4) {
         axis = _rgFunctionAxis[gimbalPitchFunction];
         gimbalPitch = _adjustRange(_rgAxisValues[axis], _rgCalibration[axis],_deadband);
     }
 
-    float gimbalYaw = 0.0f;
+    float gimbalYaw = NAN;
     if (_axisCount > 5) {
         axis = _rgFunctionAxis[gimbalYawFunction];
         gimbalYaw = _adjustRange(_rgAxisValues[axis],   _rgCalibration[axis],_deadband);
@@ -674,7 +674,7 @@ void Joystick::_handleAxis()
 
     const uint16_t lowButtons = static_cast<uint16_t>(buttonPressedBits & 0xFFFF);
     const uint16_t highButtons = static_cast<uint16_t>((buttonPressedBits >> 16) & 0xFFFF);
-    _activeVehicle->sendJoystickDataThreadSafe(roll, pitch, yaw, throttle, lowButtons, highButtons);
+    _activeVehicle->sendJoystickDataThreadSafe(roll, pitch, yaw, throttle, lowButtons, highButtons, gimbalPitch, gimbalYaw);
 }
 
 void Joystick::startPolling(Vehicle* vehicle)
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -1902,7 +1902,7 @@ void Vehicle::virtualTabletJoystickValue(double roll, double pitch, double yaw, 
                     static_cast<float>(pitch),
                     static_cast<float>(yaw),
                     static_cast<float>(thrust),
-                    0, 0);
+                    0, 0, NAN, NAN);
     }
 }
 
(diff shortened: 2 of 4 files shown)


src/Joystick/Joystick.cc
```diff
@@ -608,13 +608,13 @@ void Joystick::_handleAxis()
     axis = _rgFunctionAxis[throttleFunction];
     float throttle = _adjustRange(_rgAxisValues[axis], _rgCalibration[axis], (_throttleMode == ThrottleModeDownZero) ? false :_deadband);
 
-    float gimbalPitch = 0.0f;
+    float gimbalPitch = NAN;
```
ikalnytskyi, line 611: **n.b.** It seems both `gimbalPitch` and `gimbalYaw` functionality has never been finished. Their values were only logged but never used. I switched the default value to `NAN`, so we can distinguish an axis detected vs not detected later on without passing extra parameters to `sendJoystickDataThreadSafe()`.

HTRamsey:
Nice. I got a couple things I will follow this up with

HTRamsey:
Looks like ArduSub uses indices 6 & 7 instead of 4 and 5 for the camera. Probably need a way to add firmware defined axis mapping

DonLakeFlyer:
This change breaks the ability to calibrate a gamepad controller. I don't think just because you have more than 4 axes means that the additional axes can/should be used for a gimbal. On a gamepad the triggers are axes but with the current code they are impossible to calibrate and you just get stuck on the gimbal axis cal step. I'll see if I can figure out what is going on here.

DonLakeFlyer:
The other problem is the use of enabled_extensions bits. This all seems like magic to me. Spec says:

Pitch-only-axis, normalized to the range [-1000,1000]. Generally corresponds to pitch on vehicles with additional degrees of freedom. Valid if bit 0 of enabled_extensions field is set. Set to 0 if invalid.

No mention of gimbal. So automatically sending this out without some flag to turn it on seems like magic setup to me. It could be used for something other than a gimbal depending on the vehicle. Why is sort of what Holden is saying about ArduSub.

If there are additional axes available for use on the joystick then the user should be given the option to specify whether to use them or not in manual control. And then in what part of manual control.

DonLakeFlyer:
I'm in the middle of all the gimbal stuff right now so I'll clean this up. And @ikalnytskyi, thanks so much for this pull. Good to get all of this going in the right direction.

ikalnytskyi:
@DonLakeFlyer Thank you! It's a pleasure contributing to this project — such a welcoming community! ❤️

DonLakeFlyer:
@ikalnytskyi When you are using gimbal through MANUAL_CONTROL is this on custom firmware/vehicle or is there some default gimbal control built into firmwares via MANUAL_CONTROL?

ikalnytskyi:
@DonLakeFlyer I'm sorry, missed your question. It's a custom firmware/vehicle for a rover type drone. Not px4/ardupilot, if that's what you asked.

DonLakeFlyer:
OK, thanks

Merlin-ua:
I have a strange issue with this. I wrote a simple script in python that listens to MANUAL_CONTROL and connected and calibrated a joystick. It asked me to calibrate gimbal pitch and yaw, so I selected two channels for that. Immediately after QGC starts I receive s and t values, but at some point they just disappear, the log of my Python script looks like:

> ...
> MANUAL_CONTROL {target : 1, x : 0, y : 0, z : 0, r : 0, buttons : 0, buttons2 : 0, enabled_extensions : 3, s : 1000, t : -1000, aux1 : 0, aux2 : 0, aux3 : 0, aux4 : 0, aux5 : 0, aux6 : 0}
> MANUAL_CONTROL {target : 1, x : 0, y : 0, z : 0, r : 0, buttons : 0, buttons2 : 0, enabled_extensions : 0, s : 0, t : 0, aux1 : 0, aux2 : 0, aux3 : 0, aux4 : 0, aux5 : 0, aux6 : 0}
> ...

It never comes back to non-zero values. This happens about 5 seconds after QGC starts. I looked through the code but I don't see a possible reason for such a change in behaviour.

HTRamsey:
Can you make that into a separate issue and I'll look into it

Merlin-ua:
@HTRamsey sorry, I actually figured it out on myself. The problem was that QGC was switching into MavLink1 mode after timeout. My test script identified itself as "generic", not as "ardupilot". For generic flight controllers PROTOCOL_VERSION mavlink message is required, if version is not identified in time protocol reverts to MavLink1. All the extensions are not part of MavLink1, which is why I saw them disappear.

PR #13534, opened 2025-10-15
Mac: Fix android GST discovery

DonLakeFlyer:
@HTRamsey This is the only way I can get local android builds to work for me on Mac. Not sure why CI works, but local doesn't for me. But this fixes it. I've been having to keep this change locally hanging around so I can build android. Is it ok to merge this in?

### DIFF
--- a/cmake/find-modules/FindGStreamer.cmake
+++ b/cmake/find-modules/FindGStreamer.cmake
@@ -168,6 +168,21 @@ elseif(ANDROID)
         set(ENV{PKG_CONFIG_LIBDIR} "${GSTREAMER_LIB_PATH}/pkgconfig;${GSTREAMER_PLUGIN_PATH}/pkgconfig")
         list(APPEND PKG_CONFIG_ARGN --dont-define-prefix)
     elseif(CMAKE_HOST_UNIX)
+        if(CMAKE_HOST_APPLE)
+            # Try to find pkg-config in common Homebrew locations and in PATH
+            find_program(PKG_CONFIG_EXECUTABLE
+                NAMES pkg-config
+                PATHS /opt/homebrew/bin /usr/local/bin
+                NO_DEFAULT_PATH
+            )
+            if(NOT PKG_CONFIG_EXECUTABLE)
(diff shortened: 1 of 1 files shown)


HTRamsey:
Yup fine with me. I think somewhere in the GStreamer docs I had found it said not to use homebrew's pkgconfig but if it works it works

highfreq:
Would this apply to Mac m3 building 4.4.5? Because no matter what I can't seem to have the project find my gstreamer.

DonLakeFlyer:
No. Look in CI workflow and make sure you are replicating how it installs gst on mac.

highfreq:
I think I don't understand what I have to look at :)

DonLakeFlyer:
https://github.com/mavlink/qgroundcontrol/blob/Stable_V4.4/.github/workflows/macos_release.yml

highfreq:
i am cross compiling on Mac for an android device, does it still apply?

DonLakeFlyer:
No. 4.4 builds did not verify cross compile of android on MacOs in CI whereas master does. It should be possible to build but it could be any sort of problem. You need to install the gst framework in the default location which is Library/Frameworks I believe. Don't install gst with brew. If that still doesn't work you are going to need to track through the build system to figure out why.

highfreq:
Ok, thank you very much.

highfreq:
One last thing, what framework release do you think I should install to compile with QGC 4.4.5 ??

DonLakeFlyer:
Wait this is android, forgot. I don't think I've ever gotten cross compile on mac to work with video. You should match the android version used by linux android in CI. That said not sure if this is even possible in 4.4. So you are kinda on your own. You are better off trying to do it in a Linux VM which matches how CI works.

highfreq:
Ok thank you very much.

highfreq:
Will install linux on parallels and start from there, maybe I'll have better luck :)

highfreq:
Q5.15.2 is not available for linux anymore.........looks like I am stuck. Skydroid H16 doesn't run QGC5..0.

PR #13547, opened 2025-10-16
GitHub Actions: Store all version info in single action

DonLakeFlyer:


DonLakeFlyer:
@HTRamsey I figure out how to get version info in a single file. But I still think it ends up being too messy.

Issue #13551, opened 2025-10-17
Android, turning on radio after app is launched causes request for AUTOPILOT_VERSION to fail and version to be Unknown because of initial radio lag


jnomikos:
### Have you checked the latest release for fixes?

- [x] Yes, I’ve tested on the latest release

### Description

**Expected Behavior**
I would expect QGC to need autopilot version information in order to connect, but it seems like if it fails to get it, some generic default for autopilot is used.

I know my issue is very likely specific to my radio lagging when it first turns on, but I wonder if it would be best to require successful autopilot version in order to connect.

**Current Behavior**
Note that I've only been able to reproduce this on Android QGC on my device which has a taisync radio specifically when I have QGC already launched and turn on the radio. When it auto connects, the request to get AUTOPILOT_VERSION gets no ack each time it requests it and then QGC uses some generic default instead of trying some more.

**Steps To Reproduce**
1. Launch QGC android with radio turned off
2. Turn on radio and have device autoconnect
3. Witness this happen (might be specific to the hardware you use)

**Additional Details**
From my testing, the failure to get AUTOPILOT_VERSION is that the REQUEST_MESSAGE for it never gets an ack. 

Note:
 [I looked into the code and know that QGC never keeps trying if the initial requests for AUTOPILOT_VERSION fails and it instead just keeps connecting and assumes some generic vehicle. ](https://github.com/mavlink/qgroundcontrol/blob/0e8f0ecd8c0fd9f897c2235704431f6d9dea06dc/src/Vehicle/InitialConnectStateMachine.cc#L175-L190)

Are there any systems which necessitate this? In my opinion, maybe we shouldn't advance the initial connect state machine if we can't get AUTOPILOT_VERSION.

### System Information

QGC Versions: Stable 5.0.8, Stable 4.4.4
QGC build: Stables
Operating System: Android
Flight Controller: CubePilot Cube Blue
AutoPilot Software: ArduCopter 4.5.6

### Log Files and Screenshots

https://drive.google.com/file/d/1k68-qBZSycj-dsne8UJ6Yi9SyL3LdZ_w/view?usp=sharing

^ Video of the issue. Note in the video I say "Request for autopilot type" but I misspoke. I meant "Request for autopilot version"

DonLakeFlyer:
What needs to happen here:
[...]

PR #13552, opened 2025-10-17
InitialConnectStateMachine: Don't assume capabilities on AUTOPILOT_VERSION if failure requesting

jnomikos:
Description
-----------
I was running into an issue on my android system where if I launch the radio while QGC is already running, that initial boot of the radio will cause lag in the connection, and so the initial AUTOPILOT_VERSION request would fail.

QGC will assume some generic default autopilot version if the initial request fails rather than continually retrying. So, that launch will have the version as "Unknown".

Unless there are specific autopilots that do not use AUTOPILOT_VERSION, I think that QGC should probably require it in order to have a successful connection. This PR will continually request it during the connection phase and will not advance the state machine until it successfully gets it, which avoids chance of radio lag causing the connection to not get AUTOPILOT_VERSION successfully. 

Note: This PR is created under the assumption that there are no systems which don't use AUTOPILOT_VERSION. If it turns out that there are systems which don't provide it, we might want to maybe just increase the retries of this specific step or something

Test Steps
-----------
1. Launch QGC on device with radio turned off
2. Turn on radio
3. That initial radio lag of the radio turning on will now simply halt the connection states until we can successfully get AUTOPILOT_VERSION

Checklist:
----------

Related Issue
-----------
https://github.com/mavlink/qgroundcontrol/issues/13551

### DIFF
--- a/src/Vehicle/InitialConnectStateMachine.cc
+++ b/src/Vehicle/InitialConnectStateMachine.cc
@@ -173,17 +173,9 @@ void InitialConnectStateMachine::_autopilotVersionRequestMessageHandler(void* re
     }
 
     if (failureCode != Vehicle::RequestMessageNoFailure) {
-        qCDebug(InitialConnectStateMachineLog) << "REQUEST_MESSAGE:AUTOPILOT_VERSION failed. Setting no capabilities";
-        uint64_t assumedCapabilities = 0;
-        if (vehicle->_mavlinkProtocolRequestMaxProtoVersion >= 200) {
-            // Link already running mavlink 2
-            assumedCapabilities |= MAV_PROTOCOL_CAPABILITY_MAVLINK2;
-        }
-        if (vehicle->px4Firmware() || vehicle->apmFirmware()) {
-            // We make some assumptions for known firmware
-            assumedCapabilities |= MAV_PROTOCOL_CAPABILITY_MISSION_INT | MAV_PROTOCOL_CAPABILITY_COMMAND_INT | MAV_PROTOCOL_CAPABILITY_MISSION_FENCE | MAV_PROTOCOL_CAPABILITY_MISSION_RALLY;
-        }
-        vehicle->_setCapabilities(assumedCapabilities);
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
> This PR is created under the assumption that there are no systems which don't use AUTOPILOT_VERSION

Requiring AUTOPILOT_VERSION is not part of the mavlink spec. This will break support for generic firmwares. Also the concept that QGC could just randomly hang if there is a problem with gettting AUTOPILOT_VERSION is pretty wrong.

My suggestions for this problem:
* Vehicle::requestMessage support should be upgraded to support multiple retries like the rest of the methods like that already do. I was surprised when I went looking for requestMessage retry to find it wasn't there. I thought I had added that ages ago.
* Code which references firmware version numbers to make decisions should take into account the fact that the version might not be set. It should default to assume newer firmwares. It "mostly" does, but a full review is warranted.
* If QGC fails to get version info it should warn the user about it noting that usability might be flaky. Given them an indication that they should figure out what is wrong with their comm link.

jnomikos:
> > This PR is created under the assumption that there are no systems which don't use AUTOPILOT_VERSION
> 
> Requiring AUTOPILOT_VERSION is not part of the mavlink spec. This will break support for generic firmwares. Also the concept that QGC could just randomly hang if there is a problem with gettting AUTOPILOT_VERSION is pretty wrong.
> 
> My suggestions for this problem:
> 
>     * Vehicle::requestMessage support should be upgraded to support multiple retries like the rest of the methods like that already do. I was surprised when I went looking for requestMessage retry to find it wasn't there. I thought I had added that ages ago.
> 
>     * Code which references firmware version numbers to make decisions should take into account the fact that the version might not be set. It should default to assume newer firmwares. It "mostly" does, but a full review is warranted.
> 
>     * If QGC fails to get version info it should warn the user about it noting that usability might be flaky. Given them an indication that they should figure out what is wrong with their comm link.

Thx for the info. Given this will break support for general firmware, I'll close the PR.

PR #13564, opened 2025-10-20
Fix gamepad joystick calibration

### LINKED ISSUE #13558: Joystick Cal: Unable to calibrate gamepad controller
It's impossible to move past the gimbal axis calibration step because gamepad triggers only emit positive values. Hence the step to move the axis to min value can never be completed.

Related to #13517

DonLakeFlyer:
* Added setting on advanced page for allowing gimbal values in manual control. Just a hack at this point to allow gamepad calibration. Will be putting much more work in here in future pulls.
* Fix #13558


DonLakeFlyer:
@ikalnytskyi Can you test this with your setup?

ikalnytskyi:
> Can you test this with your setup?

Works fine. Thank you!

PR #13570, opened 2025-10-21
Standard modes: Don't filter mode list

DonLakeFlyer:
This makes all flight modes available for Joystick button actions.

This needs to be cherry picked to master as well.


DonLakeFlyer:
I'll squash this to a single commit

PR #13578, opened 2025-10-21
DelayButton: Gradient to indicate button type, auto help text

DonLakeFlyer:
* There is a gradient on the left to visual indicate this is a delay button
* If you click and release before activating it will show "Hold to Confirm" help text within the button
* The "Hold to Confirm" help text will disappear after a timeout

Gradient visual indication:
Screenshot 2025-10-21 at 12 19 21 PM

Hold to Confirm:
Screenshot 2025-10-21 at 12 19 25 PM


DonLakeFlyer:
Probably not the last iteration, but heading in the right direction...

PR #13592, opened 2025-10-25
fix typo: assigment -> assignment

julianoes:
Just came across this by accident.

### DIFF
--- a/src/Vehicle/VehicleSetup/JoystickConfig.qml
+++ b/src/Vehicle/VehicleSetup/JoystickConfig.qml
@@ -80,7 +80,7 @@ SetupPage {
                     visible:    _allowJoystickSelection
                 }
                 QGCTabButton {
-                    text:       qsTr("Button Assigment")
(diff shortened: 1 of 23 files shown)


DonLakeFlyer:
@hamishwillee I'm always confused by the right way to handle these string changes. Is it ok to update all the languages? Or should you just update english and then let crowdin flow the change through the other language?

julianoes:
No idea. @hamishwillee ?

hamishwillee:
@DonLakeFlyer @julianoes Always update just English, and crowdin will round trip the change.

It probably doesn't matter **in this case** because the strings are untranslated. However in some cases crowdin may end up clashing and you'll have to do some manual integration/fix conflicts. That is a pain in the bum.

Let's merge this one and see if it breaks anything. If it does, a learning experience for all :-)

hamishwillee:
PS Thanks for taking the time to fix this @julianoes !

PR #13545, opened 2025-10-16
Video: Added ability for force hardware decoder usage

DonLakeFlyer:
* Added setting to force hardware decoder usage
* Changed forced ranking of software/hardware decoders to walk installed plugins instead of use hardcoded list
* Added Video.GStreamerDecoderRanks category which will log the default decoder priority ranking in rank order
* Added much better logging to various places

Thanks to our AI overlords for dealing with most of this :)

PR #13617, opened 2025-10-31
fix: Android video hardware decoder selection issue

hywelstar:
Exclude Android software decoders (OMXGoogle / C2Android) and set software decoder rank to GST_RANK_NONE. Prioritize actual hardware decoders so decodebin3 automatically selects them on Android devices.

Description
-----------

Since QGroundControl v5.0, GStreamer 1.22.12 is used.  
Simply increasing the priority of hardware decoders is not enough to ensure their selection. It is also necessary to set the software decoders' rank to GST_RANK_NONE and fix the decoder detection logic.  
With these changes, the system can correctly select and use hardware decoders for video playback.

Previously, in version v4.4.5 (using GStreamer 1.18.1), simply raising the hardware decoder priority was sufficient for the system to choose hardware decoders.

Key changes in this PR:
- Exclude Android software decoders (OMXGoogle / C2Android)
- Set software decoder ranks (avdec_*, vp8dec, vp9dec, etc.) to GST_RANK_NONE
- Prioritize actual hardware decoders so that decodebin3 automatically selects them

Test Steps
-----------

1. Run QGroundControl on an Android device.
2. Play a video stream and select:
   `Video -> Settings -> Force video decoder priority -> Force hardware decoder`.
3. Enable logging to capture video-related logs.
4. Restart QGroundControl and load the video content.
5. Save the logs and check the following key entries to verify that a hardware decoder is selected.  
   > Note: The example below is for reference only; actual log entries may vary depending on the device.
```yaml
9.617 Decodebin3 selected codec:rank - "androidmedia" / amcviddec-c2imxavcdecoder - Codec/Decoder/Video/Hardware (HW) : 257 - Video.GstVideoReceiver - (GstVideoReceiver::_logDecodebin3SelectedCodec:946)
9.692 Video "videoContent" resized. New resolution: 1280 x 720 - Video.VideoManager - (VideoManager::_initVideoReceiver(VideoReceiver *, QQuickWindow *)::(anonymous class)::operator():771)
```
6. Confirm from the logs that the correct hardware decoder is selected.
7. Verify that video playback works correctly and that performance is improved.

Checklist:
----------

Related Issue
-----------

Fixes or continues work from #13545 (Android hardware decoder selection was not resolved)

### DIFF
--- a/src/VideoManager/VideoReceiver/GStreamer/GStreamer.cc
+++ b/src/VideoManager/VideoReceiver/GStreamer/GStreamer.cc
@@ -343,6 +343,29 @@ void _logDecoderRanks()
     gst_plugin_feature_list_free(decoderFactories);
 }
 
+void _lowerSoftwareDecoderRanks(GstRegistry *registry)
+{
+    static constexpr uint16_t NewRank  = GST_RANK_NONE;
+    if (!registry) {
+        qCCritical(GStreamerLog) << "Invalid registry!";
+        return;
+    }
+
+    const char* softDecoders[] = {"avdec_h264", "avdec_h265", "avdec_mjpeg", "avdec_mpeg2video", "avdec_mpeg4",
+                                  "avdec_vp8", "avdec_vp9", "dav1ddec", "vp8dec", "vp9dec"};
+
+    for (const char *name : softDecoders) {
+        GstPluginFeature *feature = gst_registry_lookup_feature(registry, name);
+        if (feature) {
+            qCDebug(GStreamerLog) << "Setting software decoder rank low:" << name << " rank:" << NewRank;
+            gst_plugin_feature_set_rank(feature, NewRank);
+            gst_object_unref(feature);
+        } else {
+            qCDebug(GStreamerLog) << "Software decoder not found:" << name;
+        }
+    }
+}
+
 void _changeFeatureRank(GstRegistry *registry, const char *featureName, uint16_t rank)
 {
     if (!registry || !featureName) {
@@ -379,7 +402,6 @@ void _prioritizeByHardwareClass(GstRegistry *registry, uint16_t prioritizedRank,
 
     qCDebug(GStreamerLog) << "Prioritizing" << (requireHardware ? "hardware" : "software")
                            << "video decoders with rank:" << prioritizedRank;
-
     int matchedFactories = 0;
     for (GList *node = decoderFactories; node != nullptr; node = node->next) {
         GstElementFactory *factory = GST_ELEMENT_FACTORY(node->data);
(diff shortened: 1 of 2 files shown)


HTRamsey:
So do you know if this issue was just caused by GStreamer 1.22.X or do all newer GStreamer versions act like this?

hywelstar:
Thanks, I’m not sure yet. I’ll compare it with the newer GStreamer versions and run some tests. If there’s any progress, I’ll post an update here.

doquangquan93:
Hello there

I installed and tested the latest Daily Build, but after switching the video setting to “Force hardware decoder”, the video does not display.
It works with the default/auto selection, but when forcing the hardware decoder, the video feed shows nothing.

Are you having the same problem?

DonLakeFlyer:
@doquangquan93 If you are having problems create a new Issue with exact details of the problem as well as exact details of the hardware/android environment.

@hywelstar This is nice, but isn't it possible to walk the software decoders like _prioritizeByHardwareClass does without hardcoding a list of decoders? Or does that not work for some reason? I removed the hardcoded lists to future proof the code.

DonLakeFlyer:
Given this change worked better, then to me the change should be use GST_RANK_NONE in both the software/hardware forcing code to lower the ranks instead of bumping up the ranks. So for hardware force, lower the rank of software decoders to GST_RANK_NONE. And for software force, lower the rank of hardware to GST_RANK_NONE. Then we have a consistent way of forcing.

DonLakeFlyer:
If folks agree (@hywelstar, @HTRamsey) I'll submit a pull with this change and we can test that. And then if this works given how android 5.0 is screwed up, I'd suggest moving up the window of 5.1 stable to get this out the door sooner rather than later.

hywelstar:
Hi, DonLakeFlyer, 

GStreamer 1.22.12, the FFmpeg-related plugins are given very high default priority:
https://gitlab.com/gstreamer/gstreamer/-/blob/1.22/subprojects/gst-libav/ext/libav/gstavviddec.c#L2700

On Android, both hardware and software decoders have lower priority. I also found in  1.22.12 that simply changing the plugin rank doesn’t solve this issue, so I think it’s a version-specific problem. That’s why I set the FFmpeg video plugins to GST_RANK_NONE here.

If we force hardware decoding, it’s possible that the chip doesn’t have a hardware decoder. In that case, no video may appear, which I think is normal behavior.

The question now is: if the device doesn’t support hardware decoding, should we fall back to software decoding? Also, I’d like to know what kind of device this is.

hywelstar:
> If folks agree (@hywelstar, @HTRamsey) I'll submit a pull with this change and we can test that. And then if this works given how android 5.0 is screwed up, I'd suggest moving up the window of 5.1 stable to get this out the door sooner rather than later.

Agreed.

DonLakeFlyer:
So I tried forcing hardware decoding on my Siyi UNIRC7 and it fails to show video in that case (using a daily build). Not sure why yet.

hywelstar:
> So I tried forcing hardware decoding on my Siyi UNIRC7 and it fails to show video in that case (using a daily build). Not sure why yet.

Hi, Could you provide the running logs so I will check what’s going on? Also, what is the current Android version, and which chip does it use?
Thanks.

PR #13634, opened 2025-11-13
NTRIP Implementation

### LINKED ISSUE #7133: RTK: NTRIP Corrections
The following template is for Issue/Bug reporting only. https://docs.qgroundcontrol.com/en/Support/Support.html#github-issues

For questions about how to use or build QGC see: http://qgroundcontrol.com/#resources

If you are using an old stable release please upgrade first and verify your issue before reporting.
----
## Expected Behavior
I want to be able to access RTK corrections from my NTRIP provider. Mission Planner allows me to do this, however I see nothing about NTRIP in QGC.

## Current Behavior
Unavailable

## Steps to Reproduce:
N/A

## Detailed Description
I believe Mission Planner's code utilizing NTRIP is here: https://github.com/ArduPilot/MissionPlanner/blob/b26324bdc7e9abd87eb7e9c02bfc11f7184a9a51/ExtLibs/Comms/CommsNTRIP.cs

olmpya12:
Closes #7133

This PR rebases and fixes the NTRIP refactor originally implemented by @KnightHawk06.

After rebasing onto current master, the NTRIP code started hitting runtime Qt asserts and thread-affinity errors related to NTRIPTCPLink and its use of sockets inside a QThread. The logic itself was mostly correct, but the object/thread ownership model no longer matched Qt’s expectations.

This PR completes the transition to a proper QObject worker running in a dedicated QThread, keeping the behavior of the original implementation but making it thread-safe.

**Problem**
After the rebase, running QGroundControl with NTRIP enabled produced runtime errors such as:
- ASSERT failure in QCoreApplication::sendEvent: "Cannot send events to objects owned by a different thread..."
- QSocketNotifier: Socket notifiers cannot be enabled or disabled from another thread
- Deadlock warnings around BlockingQueuedConnection

The root cause was that NTRIPTCPLink was being used as both:
- a worker doing socket I/O, and
- a QThread object,

which led to mismatched thread affinity:

- Sockets and notifiers were created in one thread but manipulated from another.
- BlockingQueuedConnection calls into the same thread caused deadlock warnings.
- Qt’s event delivery complained about sending events to objects owned by a different thread.

The code worked before the rebase, but the updated Qt / refactored startup logic made this misuse of QThread visible as runtime errors.

**What this PR changes**
This PR finishes the refactor by making NTRIPTCPLink a pure QObject worker and having NTRIPManager own a separate QThread to run it.
**Key changes**

- Change NTRIPTCPLink from a QThread subclass to a QObject worker.
- Add a start() slot in NTRIPTCPLink to perform _hardwareConnect() when the thread starts.
- Ensure sockets, timers, and the RTCM parser are created and used only in the worker thread.
- Replace direct thread control (isRunning(), quit(), wait(), run(), exec()) with:
    - requestStop() in the worker, which:
    - stops and deletes the socket
    - emits finished
  - NTRIPManager listening to finished and stopping the owning QThread.
- Update NTRIPManager::startNTRIP() to:
  - create a QThread
  - move the NTRIPTCPLink instance to that thread
  - connect QThread::started → NTRIPTCPLink::start()
- Update NTRIPManager::stopNTRIP() to:
  - invoke requestStop() via Qt::QueuedConnection
  - quit() and wait() on the QThread
  - let the worker and thread clean up via deleteLater().

**Result**
- No more cross-thread socket access.
- No more QSocketNotifier warnings.
- No more sendEvent/thread-affinity asserts.
- Clean start/stop behavior and predictable teardown.

**Testing**
✅ Windows (VS2022 + Qt 6.10): app starts, NTRIP manager works without runtime asserts.
✅ NTRIP connectivity:
- Tested with RTK2go using the nearest base station in simulation (SITL).
- Verified that NTRIP connects successfully and RTCM data flows.

⚠️ Not tested on a real vehicle yet:
- I currently don’t have a hardware setup ready to verify end-to-end RTK injection into an actual autopilot.
- In the past, NTRIP worked with the Turkish caster TUSAGA-Aktif, and the new threading model is expected to preserve that behavior once hardware is available again.

**Credits**
The initial NTRIP refactor and SPARTN/NTRIP work was originally implemented by @KnightHawk06.
This PR rebases that work onto current master and fixes the runtime threading issues exposed after the rebase.


olmpya12:
do I need to update everything regarding that comments? @HTRamsey

HTRamsey:
Nah, only if it's like a logic issue

HTRamsey:
@DonLakeFlyer when you see this, this code is pretty well isolated so I will be merging so that I can make some general updates and you can change the UI whenever you want

DonLakeFlyer:
Can someone also update the docs for this new support?

PR #13639, opened 2025-11-14
Derive additional action availability from model

rubenp02:
# Derive additional action availability from model

Description
-----------
Replace hardcoded expression with iteration over the model to determine whether any entry is visible, in both the AdditionalActionsList and the AdditionalCustomActionsList.

This removes the maintenance burden of updating the expression whenever actions are added, removed, or a special condition is added to their availability (e.g., only appearing in Advanced Mode).

Checklist:
----------


HTRamsey:
Not sure if those copilot suggestions matter but if so maybe something [like this](https://doc.qt.io/qt-6/qml-qt-labs-synchronizer-synchronizer.html) could help. I've been wanting to try it out.

rubenp02:
I think they're nonsense (the suggested changes sure are) since I tested it and it was reactive. But I'll test it out a bit more thoroughly when I can.

rubenp02:
The reactivity bug Copilot mentions doesn't actually exist because the model is recomputed whenever a visibility flag changes. But if the model is mutated in place rather than reassigned then yes, anyActionAvailable may be out of date. But of course, the same was true with the old code. Neither a Syncronizer nor a Binding alone would fix this, since we're using a plain JS array of vars for the model: I think that would also require using a Qt model like a ListModel or to keep the JS array but to make its elements QtObjects.

The changes work but I guess the implementation isn't the best, that much is true.

DonLakeFlyer:
I think the discussion above is about what happens when 'visible' on an item changes with respect `anyActionAvailable`. I still don't see how the Additional Actions buttons will show/hide when either all actions go invisible or it transitions from all actions invisible to one or more visible.

rubenp02:
> I think the discussion above is about what happens when 'visible' on an item changes with respect `anyActionAvailable`. I still don't see how the Additional Actions buttons will show/hide when either all actions go invisible or it transitions from all actions invisible to one or more visible.

Yes, this change is non-functional. It's just meant to avoid having to change anyActionAvailable whenever you add an action, since it can be easily derived from the model itself. In our QGC custom build we have a couple custom actions that are nonetheless implemented as normal AdditionalActions (rather than AdditionalCustomActions) and everytime something is added or removed upstream it causes merge conflicts on that line.

DonLakeFlyer:
Not sure what to do with this since it doesn't work?

rubenp02:
> Not sure what to do with this since it doesn't work?

What do you mean? I think it works.

Edit: By non-functional I mean that it doesn't intend to change any of the functionality. It's just a simple change from a hardcoded visibility expression to a generic one.

src/FlyView/FlyViewAdditionalActionsList.qml
```diff
@@ -12,8 +12,15 @@ import QtQml
...
-    property bool anyActionAvailable: guidedController.showStartMission || guidedController.showContinueMission || guidedController.showChangeAlt ||
-                                      guidedController.showChangeLoiterRadius ||  guidedController.showLandAbort || guidedController.showChangeSpeed
+    property bool anyActionAvailable: {
+        for (var i = 0; i < model.length; i++) {
+            if (model[i].visible)
+                return true
+        }
+
+        return false
+    }
```
DonLakeFlyer, line 22: I don't see how this ends up being re-triggered when a visible property on one of the model items changes value?

rubenp02: Because when one of the visible properties changes the entire model changes (a modelChanged signal is emitted) and since this reads it, it gets rerun. Note: If you were to manually set model[0].visible = false it wouldn't count as a model change because we're bypassing the binding chain, and this would fail. But we don't do that.

Basically a change of any visible property doesn't mutate the model, it actually rebuilds it instead.

DonLakeFlyer: Wow, this must have been added to Qt somewhere along the way. I'd fought with Qt specifically "not" doing that for ages. Not sure when that changed, but I've continued to assume it didn't work. Good to know.

DonLakeFlyer:
@HTRamsey Shouldn't close then reopen on this pull the new base master which doesn't have the PX4 gps driver failures. Doesn't seem to be working for me

HTRamsey:
Nah it's still 20 commits behind master. Need to specifically rebase branch

rubenp02:
Rebased. Let’s see

DonLakeFlyer:
@HTRamsey Docker Android is still 100% fail. I'll merge with that.

HTRamsey:
Yeah it's just running out of space scanning entire docker image instead of just the APK. Fixing that and a couple of other things when I get the chance

PR #13651, opened 2025-11-18
Update Mavlink submodule to fix camera settings

gillamkid:
Issue Overview
-----------
A few weeks ago, QGC updated its Mavlink git-hash, creating a bug with message encoding (this bug is not present in the 5.0.8 release, but is present in the daily build)

One way the issues manifests is that when I use QGC daily with a PX4 and an Airpixel TAG-E camera, the camera settings page shows "Unknown: 3" after I try changing the value of the TG_EXPMODE parameter
bug1

This PR fixes the issue

Issue Further details
-----------
When debugging, I added the following temporary prints to `src/Camera/QGCCameraIO.cc`
Screenshot from 2025-11-18 12-48-19

This showed that there was a bug in the Mavlink library making so the `param_value` property was not correctly being encoded.
fix1

Behavior after fix
----------
After the fix made in this PR, the camera settings page successfully can successfully change the camera mode to "A"
fix3

This is because updating the Mavlink git-hash makes so the encode function works properly
fix4

### DIFF
--- a/cmake/CustomOptions.cmake
+++ b/cmake/CustomOptions.cmake
@@ -73,7 +73,7 @@ option(QGC_ENABLE_QT_VIDEOSTREAMING "Enable QtMultimedia video backend" OFF)
 # ============================================================================
 
 set(QGC_MAVLINK_GIT_REPO "https://github.com/mavlink/c_library_v2.git" CACHE STRING "MAVLink repository URL")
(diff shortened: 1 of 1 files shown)


gillamkid:
@DonLakeFlyer you were the last one to change the Mavlink version. Please review.

HTRamsey:
I guess this is a github issue. I'll close & reopen to re-rerun all the workflows later

HTRamsey:
I will take care of it in #13653, thanks

PR #13670, opened 2025-11-24
Remove support for MAVLink v1 message protocol

### LINKED ISSUE #13625: Deprecate PROTOCOL_VERSION
### Feature Description

MAVLink is proposing to deprecate PROTOCOL_VERSION MAVLink message, and PX4 is looking to remove it - see https://github.com/PX4/PX4-Autopilot/pull/25836

The action here is to look to finding a backwards compatible way of handling its removal. Right now if the code is removed then QGC displays

Context is in linked issue.

Note, ArduPilot does not use this message.

### Flight Stacks

PX4

### Vehicle Types

_No response_

DonLakeFlyer:
* Notify user if v1 traffic is detected on link
* FIX #13625

### DIFF
--- a/src/Comms/LinkInterface.cc
+++ b/src/Comms/LinkInterface.cc
@@ -88,6 +88,7 @@ bool LinkInterface::_allocateMavlinkChannel()
 
     qCDebug(LinkInterfaceLog) << "_allocateMavlinkChannel" << _mavlinkChannel;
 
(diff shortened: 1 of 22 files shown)


DonLakeFlyer:
Gonna sit on this for a couple days to see if anyone comes up with some nasty reason to not remove v1 support

hamishwillee:
I may be the minority, but I think this is a good step forward. Thank you.

PR #13671, opened 2025-11-25
Add CUAV x25-evo board ID

cuav-zhang3:


DonLakeFlyer:
Doesn't USBBoardInfo.json need an update as well?

cuav-zhang3:
> Doesn't USBBoardInfo.json need an update as well?

Thank you for your reply. There is no need to update USBBoardInfo.json. Please help me merge. Thank you

PR #13679, opened 2025-11-26
Fix gimbal indicator pitch value jitter

DonLakeFlyer:
Values would bounce between 0.0 and -0.0 and -90 and nan.

### DIFF
--- a/src/Gimbal/GimbalController.cc
+++ b/src/Gimbal/GimbalController.cc
@@ -16,6 +16,8 @@
 #include "SettingsManager.h"
 #include "Vehicle.h"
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@Davidsastresas Can you test this out?

Davidsastresas:
I just tested, it looks good to me. Thank you for keeping me in the loop.

I had a weird crash going to the joystick menu and disconnecting the joystick, in Linux. However I wasn't able to reproduce it after several tries.

DonLakeFlyer:
Thanks for testing David

PR #13628, opened 2025-11-09
Android: Fix white bar by enabling immersive fullscreen (API 30+)

olmpya12:
Android: Fix white bar by enabling immersive fullscreen (API 30+)

After upgrading to Qt 6.10 / NDK 27, Android stopped applying immersive
flags, causing visible white bars (system nav/status areas) below the Qt window on tablets.

image

This patch restores true fullscreen behavior on Android 11+ using the
WindowInsetsController API.

Changes:
- Added QGCActivity.goImmersive() with WindowInsetsController.hide()
- Called goImmersive() from onCreate() and onWindowFocusChanged()
- Applies only on API >= 30

Result: QGroundControl launches in proper immersive fullscreen mode on
Android.
WhatsApp Image 2025-11-09 at 12 21 17_76fb24df

Test: Samsung s7 fe and samsung a63
Further: on Samsung a63 test there is a white gap in the camera area I dont know its on purpose or not but this commit fix the issue on the tablet 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #13690, opened 2025-12-02
Android: Fix white bars showing up on edge of screen

DonLakeFlyer:
Android specific replacement for #13628

@olmpya12 Can you verify this works for you?


olmpya12:
@DonLakeFlyer I will check in weekend thank you

PR #13693, opened 2025-12-03
Correct OS conditions for video decoder options

igoutta:
Description
-----------
I was wandering around the settings and then I found this list without proper order. I was like: Vulkan for WIndows? WTF, so I swaped the OS valitador.


HTRamsey:
Wow I don't how I managed but I messed up that entire list actually

HTRamsey:
Oh wait nevermind, those are the options that get removed from the total list of available options, and actually windows can use vulkan

igoutta:
> Oh wait nevermind, those are the options that get removed from the total list of available options, and actually windows can use vulkan

oh okey.

PR #13696, opened 2025-12-04
Add script to build and run Docker container with Podman for Ubuntu and Debian setups. Spanish translation accidentally added.

igoutta:
Description
-----------
Since I discovered podman as alternative of docker and love small details like the use of `--jobs` to build images, I have modified the original script to deploy the Ubuntu container for podman. I use it everyday with my Nixos workstation so I named as debian-based setup. I could be wrong in there since everything is inside the container, but I hope you could guide or accept the name as it is.

Test Steps
-----------
Install beforehand podman and setup your computer to properly use containers.

From the root folder, you could run the script with a command like as `sh deploy/docker/run-podman-debian-based.sh`.

PS. By the way, now creates a build folder if detects there is no one.


igoutta:
Oh my, I'm kinda new. There has been a slight slip, and I pushed my spanish translation done at least 66%.

igoutta:
I'm going to close the pull request and fix separate everything into branches. :3

Issue #13698, opened 2025-12-04
Fix the `--.-` display for no value


DonLakeFlyer:
It seems visually odd with the misbalanced dashes. Much better to always be '-.-'

@rubenp02 FYI: I've been seeing what this looks like in the UI and it's quite strange looking when it shows up.

rubenp02:
You mean the whole invalid strings stuff or using two dashes? I did think it looked weird as well which is why I initially used a single dash for the integer part. But as you pointed out that does make Facts with no decimals look like a minus sign.

DonLakeFlyer:
No the whole thing just the dashes display. `--` for no decimal is fine but I was seeing `--.-` in some places for decimals which just looks strange. That should just be `-.-`.

rubenp02:
I'd find that fairly inconsistent. Maybe we should use number dashes ("–")? In the QGC font they are clearly different from dashes and the minus sign, but they're still not as wide as em dashes which would look quite weird. And they're supposedly meant to be the same width as numbers. Might be exactly what we're looking for.

rubenp02:
Just tried it out and I think it's perfect. Doesn't look like a minus sign when it's on its own whatsoever:

<img width="3200" height="1904" alt="Image" src="https://github.com/user-attachments/assets/db776cf0-a54d-44e9-b82d-3dcfb7de693e" />

I left a couple normal dashes in the seconds part of the Hobbs meter just for comparison.

LMK if I should open a PR with this.

PR #13700, opened 2025-12-05
Use number dashes for invalid value placeholders

rubenp02:
# Use number dashes for invalid value placeholders

Description
-----------
Replaced hyphens with number/en dashes in Fact::invalidValueString, as well as in all invalid-value placeholder strings. Also changed instances of 2 hyphens into a single en dash. En dashes are visually wider and more uniform with numeric characters, improving readability and preventing confusion with minus signs.

Checklist:
----------

Related Issue
-----------
#13698

### DIFF
--- a/src/FactSystem/Fact.cc
+++ b/src/FactSystem/Fact.cc
@@ -386,14 +386,14 @@ QString Fact::invalidValueString(int decimalPlaces) const {
     case FactMetaData::valueTypeFloat:
     case FactMetaData::valueTypeDouble:
         if (decimalPlaces <= 0) {
-            return QStringLiteral("--");
+            return QStringLiteral("–");
         }
-        return QStringLiteral("--.") +
-               QString(decimalPlaces, QLatin1Char('-'));
+        return QStringLiteral("–.") +
+               QString(decimalPlaces, QChar(u'–'));
     case FactMetaData::valueTypeElapsedTimeInSeconds:
-        return QStringLiteral("--:--:--");
+        return QStringLiteral("––:––:––");
     default:
-        return QStringLiteral("--");
+        return QStringLiteral("–");
     }
 }
 
--- a/src/FlyView/ProximityRadarValues.qml
+++ b/src/FlyView/ProximityRadarValues.qml
@@ -40,7 +40,7 @@ QtObject {
     property var    rgRotationValueStrings:     [ rotationNoneValueString, rotationYaw45ValueString, rotationYaw90ValueString, rotationYaw135ValueString, rotationYaw180ValueString, rotationYaw225ValueString, rotationYaw270ValueString, rotationYaw315ValueString ]
 
     property var    _distanceSensors:       vehicle ? vehicle.distanceSensors : null
-    property string _noValueStr:            qsTr("--.--")
+    property string _noValueStr:            qsTr("–.––")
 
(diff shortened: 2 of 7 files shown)


rubenp02:
By the way, I've also tried using em dashes for the non-numeric types as well as the integer part of numeric types (basically the places where we are currently using 2 hyphens), and it doesn't look too bad at all:

imagen

DonLakeFlyer:
I don't think this looks better. I like the existing look better.

rubenp02:
> I don't think this looks better. I like the existing look better.

That's fair. You mean the em dashes, or using number dashes in general? That last screenshot is with the further change to em dashes, this PR just does the switch to number dashes. You can see how this one looks in the screenshot in issue #13698.

DonLakeFlyer:
> You mean the em dashes

Correct

src/FactSystem/Fact.cc
```diff
@@ -386,14 +386,14 @@ QString Fact::invalidValueString(int decimalPlaces) const {
...
-            return QStringLiteral("--");
+            return QStringLiteral("–");
         }
-        return QStringLiteral("--.") +
-               QString(decimalPlaces, QLatin1Char('-'));
+        return QStringLiteral("–.") +
+               QString(decimalPlaces, QChar(u'–'));
     case FactMetaData::valueTypeElapsedTimeInSeconds:
-        return QStringLiteral("--:--:--");
+        return QStringLiteral("––:––:––");
```
DonLakeFlyer, line 394: Just to be clear, using this as an example. Since I think I said it wrong. I like the original version better where  the indicator is made up of dashes or negative signs which show up as individual digits and don't blend together as a single bloc.

rubenp02: I think they just blend together because of the font, they look separate both in QtCreator and in QGC. They're just like hyphens, but wider (average digit width). Check the screenshot in #13698 to see how they look. To me it's much better than master, makes the fields jump less in size when they finally become valid. Also removes the inconsistency of using 2 dashes since these don't really look like a minus sign.

DonLakeFlyer: > I think they just blend together because of the font,

Yeah, seem right. I just built it and it looks like this:
Screenshot 2025-12-08 at 8 16 05 AM
which seems fine to me.

rubenp02: Wait how come the vertical speed and ground speed are 0? Should be invalid and using the placeholder string.

DonLakeFlyer: Good question. Didn't notice that!

DonLakeFlyer:
I can merge once conflicts are fixed and you've figure out what's going on with the missing vertical/ground speed thing.

rubenp02:
> I can merge once conflicts are fixed and you've figure out what's going on with the missing vertical/ground speed thing.

I've rebased this, but I can't replicate the vertical/ground speed stuff, it works as expected on my end everytime I launch a new instance with default settings.

rubenp02:
By the way, something's up with the CI, seems like it's getting a captcha on the GStreamer download step.

DonLakeFlyer:
> By the way, something's up with the CI, seems like it's getting a captcha on the GStreamer download step.

Fixed

rubenp02:
CI passed, should be ready to merge

PR #13728, opened 2025-12-10
Add video-fit and video-grid-line settings to Video Settings page

gillamkid:
Before, these 2 settings could only be changed when a MAVLink Camera was connected (editors for the settings exist in the Camera settings page). This was problematic because in some setups there is a video stream but no MAVLink Camera. Adding the settings to the Video Settings page makes so the settings are editable when there is no MAVLink camera.

VideoFitSettings

### DIFF
--- a/src/UI/AppSettings/VideoSettings.qml
+++ b/src/UI/AppSettings/VideoSettings.qml
@@ -141,4 +141,23 @@ SettingsPage {
             enabled:            _videoSettings.enableStorageLimit.rawValue
         }
     }
+
+    SettingsGroupLayout {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
> Before, these 2 settings could only be changed when a MAVLink Camera was connected

That's a bug. It was just fixed here: https://github.com/mavlink/qgroundcontrol/pull/13737

DonLakeFlyer:
If you find other problems with naked video streams (no camera v2 protocol) make sure to let me know. I've been cleaning my way through that in the last couple days. In general it should all work similar to a mavlink v2 camera.

PR #13758, opened 2025-12-17
Revert "Fix bug associated with value type null validation"

rubenp02:
# Revert "Fix bug associated with value type null validation"

Description
-----------
This reverts commit f3f7162d2eac5dbb3d1e79eb30a091111c8fa657.

The bugfix in that commit introduced a regression where the JSON parser failed on null values in double fields, breaking core features such as .plan mission loading. The original logic is correct for what the code comment states: accept null values to represent "no value" even when the JSON field is expected to be double.

Test Steps
-----------
Load pretty much any .plan file.

Checklist:
----------

### DIFF
--- a/src/MissionManager/MissionCommandUIInfo.cc
+++ b/src/MissionManager/MissionCommandUIInfo.cc
@@ -364,7 +364,7 @@ bool MissionCommandUIInfo::loadJsonInfo(const QJsonObject& jsonObject, bool requ
 
             // Validate key types
             QList<QJsonValue::Type> types;
-            types << QJsonValue::Null << QJsonValue::Double << QJsonValue::String << QJsonValue::String
+            types << QJsonValue::Double << QJsonValue::Double << QJsonValue::String << QJsonValue::String
                   << QJsonValue::String << QJsonValue::String << QJsonValue::Bool
                   << QJsonValue::Double << QJsonValue::Double << QJsonValue::Double << QJsonValue::Double;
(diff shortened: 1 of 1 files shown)


rubenp02:
I realized that a bugfix in https://github.com/mavlink/qgroundcontrol/commit/62296617026c9bbdef7e1db6930d18cf52ce3220 finally turned on JSON validation which in turn exposed another bug where the default value was expected to be null (to represent optional double) rather than double which is allowed to be null. So I've fixed that to support the revert.

This fixes the plan loading and everything else I've tested so far. Probably enough for a hotfix. However, this changes 8 y.o. code in the JSON parser. Might be wise to look into it more thoroughly than I've been able to.

src/MissionManager/MissionCommandUIInfo.cc
```diff
@@ -364,7 +364,7 @@ bool MissionCommandUIInfo::loadJsonInfo(const QJsonObject& jsonObject, bool requ
 
             // Validate key types
             QList<QJsonValue::Type> types;
-            types << QJsonValue::Null << QJsonValue::Double << QJsonValue::String << QJsonValue::String
+            types << QJsonValue::Double << QJsonValue::Double << QJsonValue::String << QJsonValue::String
```
DonLakeFlyer: The default value can be either a Double or a NaN. Which is signified by the `QJsonValue::Null` type. So this is correct as is.

PR #13767, opened 2025-12-18
Remove PX4 Obstacle Avoidance (deprecated)

dakejahl:
Remove COM_OBS_AVOID parameter and associated menu item since PX4 no longer supports Obstacle Avoidance. PX4 still supports Collision Prevention, which is supported in QGC via the (misnamed) `VehicleObjectAvoidance` class.

I thought that **src/FirmwarePlugin/PX4/PX4ParameterFactMetaData.xml** was autogenerated? I'm not sure where or how, if so, we should add a header to the file which says something like `DO NOT MANUALLY EDIT, AUTOGENERATED BY...`

### DIFF
--- a/src/AutoPilotPlugins/PX4/SafetyComponent.qml
+++ b/src/AutoPilotPlugins/PX4/SafetyComponent.qml
@@ -48,7 +48,6 @@ SetupPage {
             property Fact _dlLossAction:        controller.getParameterFact(-1, "NAV_DLL_ACT")
             property Fact _disarmLandDelay:     controller.getParameterFact(-1, "COM_DISARM_LAND")
             property Fact _collisionPrevention: controller.getParameterFact(-1, "CP_DIST")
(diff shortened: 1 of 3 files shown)


HTRamsey:
There used to be a bot that updated the xml. I think I asked @DonLakeFlyer about it before but forgot the answer

dakejahl:
@copilot is PX4ParameterFactMetaData.xml autogenerated? Where?

HTRamsey:
@copilot are you stupid

HTRamsey:
Copilot has achieved true AGI

dakejahl:
lol yeah okay it's not quite there yet.

> Note on PX4ParameterFactMetaData.xml:
This file is autogenerated by PX4's xmlout.py script but manually copied into QGC. A bot previously automated this sync but is no longer active. Manual edits for deprecated parameter removal are appropriate - updated metadata is fetched via MAVLink when users flash newer firmware.

PR #13812, opened 2025-12-25
Add GPS resilience monitoring with GNSS_INTEGRITY message

Tory9:
### Description
Adds GPS resilience monitoring to display jamming, spoofing, and authentication status from MAVLink GNSS_INTEGRITY messages.

Based on: PR #13009 

### Key Features:

- Dual-GPS aggregation: merges integrity data from GPS1 and GPS2 using worst-case logic to always display the most critical state
- Layered toolbar indicator: shield icon representing authentication status with overlaid interference waves for jamming/spoofing alerts
- Per-GPS drill-down view: status panel showing individual GPS receiver resilience metrics
- GPS error field enhancement: extends existing GPS status QML with error reporting

Screenshot 2025-12-23 171843

### Type of Change
 New feature (non-breaking change that adds functionality)

### Related Work
MAVLink Message Definition (development. xml): mavlink/mavlink#2110
Message value naming updated: mavlink/mavlink#2385
PX4 implementation using GNSS_INTEGRITY: PX4/PX4-Autopilot#25012 (merged)
ArduPilot implementation using GNSS_INTEGRITY: ArduPilot/ardupilot#31040 (in review)

### Hardware Tested:
Flight Controller: CubeOrange+
Flight Stacks: ArduPilot & PX4 implementations
GNSS Receiver: Septentrio Mosaic X5
Platform: Linux

### DIFF
--- a/src/UI/toolbar/GPSResilienceIndicator.qml
+++ b/src/UI/toolbar/GPSResilienceIndicator.qml
@@ -127,25 +127,25 @@ Item {
                 heading: qsTr("GPS 1 Details")
                 showDividers: true
                 visible: _activeVehicle && _activeVehicle.gps && (
-                    _activeVehicle.gps.jammingState.value < 255 ||
-                    _activeVehicle.gps.spoofingState.value < 255 ||
-                    _activeVehicle.gps.authenticationState.value < 255
+                    (_activeVehicle.gps.jammingState.value > 0 && _activeVehicle.gps.jammingState.value < 255) ||
+                    (_activeVehicle.gps.spoofingState.value > 0 && _activeVehicle.gps.spoofingState.value < 255) ||
+                    (_activeVehicle.gps.authenticationState.value > 0 && _activeVehicle.gps.authenticationState.value < 255)
                 )
 
                 LabelledLabel {
                     label: qsTr("Jamming")
                     labelText: (_activeVehicle && _activeVehicle.gps) ? (_activeVehicle.gps.jammingState.enumStringValue || qsTr("n/a")) : qsTr("n/a")
-                    visible: _activeVehicle.gps.jammingState.value < 255
+                    visible: _activeVehicle.gps.jammingState.value > 0 && _activeVehicle.gps.jammingState.value < 255
                 }
                 LabelledLabel {
                     label: qsTr("Spoofing")
                     labelText: (_activeVehicle && _activeVehicle.gps) ? (_activeVehicle.gps.spoofingState.enumStringValue || qsTr("n/a")) : qsTr("n/a")
-                    visible: _activeVehicle.gps.spoofingState.value < 255
+                    visible: _activeVehicle.gps.spoofingState.value > 0 && _activeVehicle.gps.spoofingState.value < 255
                 }
                 LabelledLabel {
                     label: qsTr("Authentication")
                     labelText: (_activeVehicle && _activeVehicle.gps) ? (_activeVehicle.gps.authenticationState.enumStringValue || qsTr("n/a")) : qsTr("n/a")
-                    visible: _activeVehicle.gps.authenticationState.value < 255
+                    visible: _activeVehicle.gps.authenticationState.value > 0 && _activeVehicle.gps.authenticationState.value < 255
                 }
             }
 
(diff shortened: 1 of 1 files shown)


HTRamsey:
you can ignore the pre-commit for now. Is this ready?

Tory9:
Hi @HTRamsey. Yes, I think it is ready for review.
A few design decisions are worth discussing in case the community sees a better or more suitable approach, such as the addition of VehicleGPSAggregateFactGroup, authentication state merge priority, and info displaying in the GUI

HTRamsey:
Gotcha, in general it look good to me. We'll wait for @DonLakeFlyer to check out the QML and your attached picture

src/QmlControls/GPSIndicatorPage.qml
```diff
@@ -63,6 +63,30 @@ ToolIndicatorPage {
...
+                return qsTr("Antenna");
+            case 16:
+                return qsTr("Event congestion");
+            case 32:
+                return qsTr("CPU overload");
+            case 64:
+                return qsTr("Output congestion");
+            default:
+                return qsTr("Multiple errors");
+        }
```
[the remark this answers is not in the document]
dakejahl: This does need to be addressed

src/Vehicle/FactGroups/VehicleGPSAggregateFactGroup.cc
```diff
@@ -0,0 +1,164 @@
...
+    //Priority: 0 (Unknown) < 4 (Disabled) < 1 (Initializing) < 3 (OK) < 2 (Error)
+    auto getWeight = [](int val) {
+        switch (val) {
+        case -1: return -1;
+        case 0: return 0;
+        case 4: return 1;
+        case 1: return 2;
+        case 3: return 3;
+        case 2: return 4;
+        default: return -1;
```
dakejahl: these should be defined somewhere, not magic numbers

src/Vehicle/FactGroups/VehicleGPSAggregateFactGroup.cc
```diff
@@ -0,0 +1,164 @@
...
+
+VehicleGPSAggregateFactGroup::VehicleGPSAggregateFactGroup(QObject *parent)
+    : FactGroup(0, parent)
+    , _spoofingStateFact      (0, "spoofingState",       FactMetaData::valueTypeUint8, this)
+    , _jammingStateFact       (0, "jammingState",        FactMetaData::valueTypeUint8, this)
+    , _authenticationStateFact(0, "authenticationState", FactMetaData::valueTypeUint8, this)
+    , _isStaleFact            (0, "isStale",             FactMetaData::valueTypeBool, this)
+{
+    _spoofingStateFact.setEnumInfo(
+        QStringList({ "Unknown", "Not spoofed", "Mitigated", "Ongoing" }),
```
dakejahl: This is already done during the metadata parsing, why do it here?

src/Vehicle/FactGroups/VehicleGPSFactGroup.cc
```diff
@@ -10,6 +10,8 @@
 #include "VehicleGPSFactGroup.h"
 #include "Vehicle.h"
 #include "QGCGeo.h"
+#include "QGCLoggingCategory.h"
+#include "development/mavlink_msg_gnss_integrity.h"
```
dakejahl, line 14: So QGC is indeed building and shipping development messages? I thought recently we decided not to do that anymore due to the standard modes fiasco? Should we put the mavlink development message features behind a preprocessor guard for debug builds only?

src/Vehicle/FactGroups/VehicleGPSFactGroup.h
```diff
@@ -54,4 +75,14 @@ class VehicleGPSFactGroup : public FactGroup
...
+    Fact _systemErrorsFact = Fact(0, QStringLiteral("systemErrors"), FactMetaData::valueTypeUint32);
+    Fact _spoofingStateFact = Fact(0, QStringLiteral("spoofingState"), FactMetaData::valueTypeUint8);
+    Fact _jammingStateFact = Fact(0, QStringLiteral("jammingState"), FactMetaData::valueTypeUint8);
+    Fact _authenticationStateFact = Fact(0, QStringLiteral("authenticationState"), FactMetaData::valueTypeUint8);
+    Fact _correctionsQualityFact = Fact(0, QStringLiteral("correctionsQuality"), FactMetaData::valueTypeUint8);
+    Fact _systemQualityFact = Fact(0, QStringLiteral("systemQuality"), FactMetaData::valueTypeUint8);
+    Fact _gnssSignalQualityFact = Fact(0, QStringLiteral("gnssSignalQuality"), FactMetaData::valueTypeUint8);
+    Fact _postProcessingQualityFact = Fact(0, QStringLiteral("postProcessingQuality"), FactMetaData::valueTypeUint8);
+
+    uint8_t _gnssIntegrityId;
```
dakejahl:
```suggestion
    uint8_t _gnssIntegrityId {};
```

src/Vehicle/FactGroups/VehicleGPS2FactGroup.h
```diff
@@ -17,7 +17,10 @@ class VehicleGPS2FactGroup : public VehicleGPSFactGroup
 
 public:
     explicit VehicleGPS2FactGroup(QObject *parent = nullptr)
-        : VehicleGPSFactGroup(parent) {}
+        : VehicleGPSFactGroup(parent)
+    {
+        _setGnssIntegrityContext(1);
```
dakejahl: what's this supposed to do?

dakejahl (commented):
Do we really want to aggregate the data? I would probably try and use the FactGroupListModel so that you can show the data from both GPS if available rather than aggregating and only showing the worst.

src/Vehicle/FactGroups/VehicleGPSFactGroup.cc
HTRamsey: I think it was like that, and then we needed development for some particular reason that I can't recall right now

src/QmlControls/GPSIndicatorPage.qml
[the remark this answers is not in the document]
Tory9: I intentionally grouped multiple errors into a single "Multiple errors" message to keep the UI clean, as there isn't much space to list several error strings like "Incoming correction, Configuration" without cluttering the display. Since these are exceptional situations, and having multiple at once is even rarer, a simple summary seemed like the best trade-off, especially since investigation would be required anyway. However, if you think seeing every individual error is critical, I can add the bit-flag checking

Tory9:
> Do we really want to aggregate the data? I would probably try and use the FactGroupListModel so that you can show the data from both GPS if available rather than aggregating and only showing the worst.

This aggregation strategy was actually recommended in earlier discussions (see PR #13009). The idea is to use the toolbar icon for a "worst-case" summary—if either GPS has an issue, the pilot sees it immediately without having to mentally parse two separate statuses. Since we still show the full breakdown for GPS 1 and GPS 2 in the expanded details view, we get the immediate warning on the main screen with the specific details just a click away

src/Vehicle/FactGroups/VehicleGPSAggregateFactGroup.cc
Tory9: Thanks for the catch, addressed this in commit 507556e8a19addfbf74ba6582de79a2c723e0061

src/Vehicle/FactGroups/VehicleGPSAggregateFactGroup.cc
Tory9: When I added it initially I wasn’t loading the parent metadata, but it makes more sense to rely on it. I dropped the redundant setEnumInfo in commit 34a7f52e9e7c9c7b84a4c6bf4ce57e44fc7e1ed3

src/Vehicle/FactGroups/VehicleGPS2FactGroup.h
Tory9: I wanted each GPS fact group to filter GNSS_INTEGRITY msgs only for its corresponding receiver (gps1/gps2). That’s why I added gnssIntegrityId, which is checked in _handleGnssIntegrity before setting values. Earlier I used a helper _setGnssIntegrityContext(), but it was simpler to set the ID directly in the constructor. This is in commit b9c531e1958fb49f3c922f125f250dee0b7756a5

HTRamsey:
@Tory9 Since the c++ code is valuable and well isolated I'm gonna go ahead and merge this with the understanding that @DonLakeFlyer may want to make changes to the indicator.

PR #13280, opened 2025-08-10
Feature: Add QML-configurable gimbal max speed and deadzones

Niki-dev12:
**Title:**
----------
Add QML-configurable gimbal max speed and yaw/pitch deadzone settings

**Description**
----------
This PR adds new joystick configuration features to allow runtime adjustment of:
    Gimbal pitch and yaw max speed, using a single gimbalMaxSpeed parameter exposed to QML
    Deadzone control for pitch and yaw gimbal axes via gimbalPitchDeadzone and gimbalYawDeadzone
    Enable/disable axis-based gimbal control (gimbalAxisEnabled)

This allows users to fine-tune gimbal behavior without recompiling, which is especially useful when mapping joystick axes for dynamic camera control.

**Changes include:**
----------
    Q_PROPERTY bindings for the three settings
    Persistent saving and loading through _saveSettings() and _loadSettings()
    Updated _handleAxis() to apply speed scaling and deadzone filtering
    Optional offset for gimbal pitch and yaw to avoid zero drift

**Test Steps**
----------
    Launch QGC with a joystick connected
    Use a QML debug view or a settings panel to:
        Set gimbalMaxSpeed (e.g., 30–160)
        Adjust gimbalPitchDeadzone or gimbalYawDeadzone
        Toggle gimbalAxisEnabled
    Observe how gimbal responsiveness changes accordingly
    Confirm values persist between sessions

Tested in:
----------
    Real drone running ardupilot with gimbal
    Ubuntu 22 and Zorro 

Checklist:
----------

- [ X] [Review Contribution Guidelines](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CONTRIBUTING.md).
- [ X] [Review Code of Conduct](https://github.com/mavlink/qgroundcontrol/blob/master/.github/CODE_OF_CONDUCT.md).
- [ X] I have tested my changes.

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #13818, opened 2025-12-30
Complete rewrite of Joystick/RC Cal and friends code

DonLakeFlyer:
* Both Joystick and RC calibration share the same base class code and ui
* The changes should be functionally equivalent to previous support
* Joystick has a new settings format. Legacy settings are migrated one time
* One small functional difference is that Joystick always shows up in Vehicle Setup even when there are no joysticks

Coming in a later pull:
* I'm not super happy with settings ui. But it's functional for now. I'll continue to fix it up.
* ArduSub JS button support has not been converted over yet.
* I want to add gimbal axis display just like roll/pitch/yaw/throttle
* I'm considering adding a joystick indicator to the toolbar to indicate readiness
* Gimbal pitch/roll is hardwired to axis 4/5 like it was before. This should be configured through stick detection

Screenshot 2025-12-30 at 11 24 46 AM


DonLakeFlyer:
@Niki-dev12, @ikalnytskyi Hey can you guys help test this out? After this merges in then I'll move over @Niki-dev12 other gimbal work into this: [#13280](https://github.com/mavlink/qgroundcontrol/pull/13280)

ikalnytskyi:
@DonLakeFlyer I'll have some time to test it tomorrow. Thank you for letting me know!

ikalnytskyi:
@DonLakeFlyer I finally had a chance to test this. The extra gimbal axes (pitch/roll) are detected separately (axes 5 and 6), but only one axis is actually sent in the `MANUAL_CONTROL` message. Its value ends up in both `MANUAL_CONTROL.s` and `MANUAL_CONTROL.t`, which isn't expected. Could this be because we don't prompt the user to assign a function to them anymore?

DonLakeFlyer:
> The extra gimbal axes (pitch/roll) are detected separately (axes 5 and 6), but only one axis is actually sent in the `MANUAL_CONTROL` message. Its value ends up in both `MANUAL_CONTROL.s` and `MANUAL_CONTROL.t`, which isn't expected.

Fixed

> Could this be because we don't prompt the user to assign a function to them anymore?

Not sure what you mean here? In the old code I thought there was just a manual control extensions setting. That still exists in the new ui. What prompt are you talking about? Maybe I missed something.

ikalnytskyi:
> Fixed

Thanks! It works now.

> Not sure what you mean here? In the old code I thought there was just a manual control extensions setting. That still exists in the new ui. What prompt are you talking about? Maybe I missed something.

I was referring to a calibration step, not a prompt. In the old flow, the user was asked to move the gimbal pitch stick first (up/down), and then the gimbal roll stick. That step is what determined which physical stick was assigned to which function. There's also a chance this was changed later, I was still using an early Daily Build. So I’m wondering how the function is chosen now.

DonLakeFlyer:
> In the old flow, the user was asked to move the gimbal pitch stick first (up/down), and then the gimbal roll stick.

You're right. I screwed that up. The gimbal axes are hardwired in this version. I'll fix that up in another pull to get that functionality back.

ikalnytskyi:
> The gimbal axes are hardwired in this version. I'll fix that up in another pull to get that functionality back.

@DonLakeFlyer I think the worst part is that since there's no `min`/`max` range detected for those axis, they have broken behaviour: either -1000 or 1000, nothing in between.

DonLakeFlyer:
The joystick cal has a final step which should capture min/max for all additional axes. It's the last step in the cal process. It should say something like "Move all stick to extreme positions" or something like that. That said, there might be a bug there.

Niki-dev12:
We have verified the PR with a **Radiomaster Zorro** on **Windows and Ubuntu**.

-   **Windows:**
    
    -   Joystick calibration works and the throttle axis is correct.
    -   Gimbal axes cannot be configured or calibrated.
        
-   **Ubuntu:**
    
    -   Zorro calibration cannot be completed.
    -   The throttle axis is offset and cannot move into the negative range.
    -   Some axes appear to have a halved/asymmetric range.
    -   Axis order/enumeration differs compared to Windows.
    -   SDL test apps report correct ranges, so this seems QGC side.
        
We understand this PR focuses on restructuring and that some joystick/gimbal work is planned later, but in its current state the Zorro is not usable on Linux.

**Question:**  
Do you expect the Linux axis range/enumeration issues and gimbal axis selection/calibration to be addressed in a follow-up PR in the near future?

For now, we will leave this as public feedback only. We are happy to retest once changes land.

Thanks again for the work and responsiveness.

PR #13828, opened 2026-01-02
Handles USB disconnect by closing port on resource error

AlexKlimaj:
Closes the serial port when a resource error occurs, typically due to a USB disconnect, to facilitate automatic reconnection. Improves error logging by making it more context-specific and ensures errors are only emitted after handling port closure.


HTRamsey:
Is this related to [this](https://github.com/qt/qtserialport/commit/a9206ddefa5650846484b506c863288e08d70fbc) newish Qt commit?

PR #13839, opened 2026-01-05
Vagrant: use Ubuntu Questing for base image

peterbarker:
Jammy doesn't have a recent-enough version of cmake to compile qgc

## Description
`vagrant up` doesn't give you a QGC binary at all as it is outdated

## Type of Change
- [ ] Documentation update
- [x] CI/Build changes
- [ ] Other

## Testing

- [x] Tested locally
- [x] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [x] ArduPilot

### DIFF
--- a/deploy/vagrant/Vagrantfile
+++ b/deploy/vagrant/Vagrantfile
@@ -32,7 +32,7 @@ Vagrant.configure(2) do |config|
   config.vm.provider "virtualbox"
   config.vm.provider "vmware_fusion"
 
-  config.vm.box = "ubuntu/jammy64"
+  config.vm.box = "alvistack/ubuntu-25.10"
   config.vm.provider :docker do |docker, override|
     override.vm.box = "tknerr/baseimage-ubuntu-16.04"
   end
@@ -80,6 +80,8 @@ Vagrant.configure(2) do |config|
 
      sudo DEBIAN_FRONTEND=noninteractive apt-get -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install %{apt_pkgs}
 
+     sudo apt-get install -y python3-pip python3-venv python3-dev
+
      # use common package installation:
      /usr/bin/env bash /vagrant/tools/setup/install-dependencies-debian.sh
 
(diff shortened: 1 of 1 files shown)


HTRamsey:
Is there value in adding this as a workflow test? I figure most people just use multipass now instead of Vagrant though. @DonLakeFlyer Ramon had mentioned get some AWS runners for us so we can cut CI time down significantly to add other stuff.

DonLakeFlyer:
> Is there value in adding this as a workflow test?

If we support it, then yeah CI validation is needed. If we can get away with not supporting it you know I love reducing the maintenance burden. Then we should remove it, and point people somewhere else. @peterbarker would have to comment on that.

peterbarker:
> Is there value in adding this as a workflow test? I figure most people just use multipass now instead of Vagrant though. @DonLakeFlyer Ramon had mentioned get some AWS runners for us so we can cut CI time down significantly to add other stuff.

I doubt a workflow test will work well for this; it takes a *long* time to run, chewing up resources better used elsewhere.

multipass/docker isn't really the same thing as a Vagrant-created virtualbox image here - they're used differently.

peterbarker:
> > Is there value in adding this as a workflow test?
> 
> If we support it, then yeah CI validation is needed. If we can get away with not supporting it you know I love reducing the maintenance burden. Then we should remove it, and point people somewhere else. @peterbarker would have to comment on that.

The maintenance burden isn't really high here.

Just did a quick seartch... `qgroundcontrol multipass` didn't seem to find any documentation.  I've tried docker for QGC before... from recollection that wasn't pretty, but the tree may not have been in a good state at the time.

ATM this is the only way for me to build QGC, so from my own perspective I'd like to keep it around.  I can just keep it in a branch I rebase on QGC master, 'though, so if you're of the opinion the presence of the Vagrant support clutters up the tree and offers no value then remove it.  Best check with the other people that have poked it over the years, they may still have interest, 'though.

HTRamsey:
Docker builds are verified in CI now so I'd say you probably tried before the QGCv5 migration was in a decent state. Multipass isn't verified and probably needs to be updated to get it to work. Also documentation on the various deployment methods tends to lag behind actual state a fair bit. In my personal opinion it's not like leaving it involves any real clutter, but neither Don nor I are likely to test or maintain it. So the only real risk is having a broken Vagrant script and getting complaints in the issues if anybody tried to use it

PR #13844, opened 2026-01-06
Fix path ghosting when dragging takeoff item

DonLakeFlyer:
* When the launch position changes (settings item coordinate) only recalc flight status
* Did some cleanup on TakeoffMissionItem, no functionality differences
* Replacement for #13810


DonLakeFlyer:
@Eureak-zon Can you give this a try?

Eureak-zon:
yes

Eureak-zon:
@DonLakeFlyer  I’ve tested it, and this submission is fine.

PR #13865, opened 2026-01-14
flight modes: px4: remove RC_DZ params

dakejahl:
I noticed while on the Flight Modes page the console was spamming a warning
```
   122.770 Warning: Missing parameter: 1 "RC6_DZ" - FactSystem.FactPanelController - (FactPanelController::_reportMissingParameter:40)
   122.770 Critical: "RC6_DZ"  Fact is NULL in channelValuesChanged /home/jake/code/jake/qgroundcontrol/src/AutoPilotPlugins/PX4/PX4SimpleFlightModesController.cc 108 - default - (PX4SimpleFlightModesController::channelValuesChanged:108)
   122.825 Warning: Missing parameter: 1 "RC6_DZ" - FactSystem.FactPanelController - (FactPanelController::_reportMissingParameter:40)
   122.825 Critical: "RC6_DZ"  Fact is NULL in channelValuesChanged /home/jake/code/jake/qgroundcontrol/src/AutoPilotPlugins/PX4/PX4SimpleFlightModesController.cc 108 - default - (PX4SimpleFlightModesController::channelValuesChanged:108)
   122.871 Warning: Missing parameter: 1 "RC6_DZ" - FactSystem.FactPanelController - (FactPanelController::_reportMissingParameter:40)
   122.871 Critical: "RC6_DZ"  Fact is NULL in channelValuesChanged /home/jake/code/jake/qgroundcontrol/src/AutoPilotPlugins/PX4/PX4SimpleFlightModesController.cc 108 - default - (PX4SimpleFlightModesController::channelValuesChanged:108)

```

This also fixes the mode/channel highlighting due to the early return from this error.

RC_DZ params have been removed since https://github.com/PX4/PX4-Autopilot/pull/25502

### DIFF
--- a/src/AutoPilotPlugins/PX4/PX4SimpleFlightModesController.cc
+++ b/src/AutoPilotPlugins/PX4/PX4SimpleFlightModesController.cc
@@ -135,7 +135,7 @@ void PX4SimpleFlightModesController::channelValuesChanged(QVector<int> pwmValues
             calibrated_value = (channelValue - pwmTrim) / (float)(pwmTrim - pwmMin);
 
         } else {
-            /* in the configured dead zone, output zero */
+            /* at the trim position, output zero */
             calibrated_value = 0.0f;
         }
 
--- a/src/AutoPilotPlugins/PX4/PX4SimpleFlightModesController.cc
+++ b/src/AutoPilotPlugins/PX4/PX4SimpleFlightModesController.cc
@@ -100,18 +100,6 @@ void PX4SimpleFlightModesController::channelValuesChanged(QVector<int> pwmValues
 
     int pwmTrim = pFact->rawValue().toInt();
 
-    pFact = getParameterFact(-1, QString("RC%1_DZ").arg(flightModeChannel + 1));
-    if(!pFact) {
-#if defined _MSC_VER
-        qCritical() << QString("RC%1_DZ").arg(flightModeChannel + 1) << "Fact is NULL in" << __FILE__ << __LINE__;
-#else
-        qCritical() << QString("RC%1_DZ").arg(flightModeChannel + 1) << " Fact is NULL in" << __func__ << __FILE__ << __LINE__;
-#endif
-        return;
-    }
-
-    int pwmDz = pFact->rawValue().toInt();
-
     if (flightModeChannel < 0 || flightModeChannel > channelCount) {
         return;
     }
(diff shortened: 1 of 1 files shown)


HTRamsey:
Is this something that should handle backwards compat with those params?

dakejahl:
> Is this something that should handle backwards compat with those params?

Hmm possibly yes, I'll take a look again

DonLakeFlyer:
How far back is the version that removed these? Wondering about back compat.

dakejahl:
> How far back is the version that removed these? 

August 2025. 

> Wondering about back compat.

I gave Claude the context and asked it to look at the code, here's what it says (it's right)

---

**Old PX4 (still has RC_DZ) with new QGC:** The old firmware still applies the deadzone internally when normalizing RC values, so there's a slight mismatch in the flight mode indicator calculation. In practice this doesn't matter much - the default deadzone is 10us out of a ~1000us range, which translates to at most ~2% error in the normalized value. Since flight mode channels are typically switches snapping between discrete positions (not proportional inputs), and each slot is ~0.33 wide, you'd have to be right at a boundary to see a difference. And even then, this is just a cosmetic indicator in QGC - it doesn't affect what PX4 actually does.                                                                                                                                                           
                                                                                                                                                                                                            
**New PX4 (RC_DZ removed) with old QGC:** This is the more important case. The old QGC code tries to fetch `RC{n}_DZ` via `getParameterFact()`, gets null, logs an error, and bails out. The flight mode indicator stops working entirely.                                                                                                                                                                         
                                                                                                                                                                                                            
So removing the RC_DZ handling is the right call - keeping it would break on new firmware, while removing it only causes a negligible display difference on old firmware

dakejahl:
@DonLakeFlyer we should get this merged, the flight modes page is broken pending this

mengchaoheng:
> **New PX4 (RC_DZ removed) with old QGC:** This is the more important case. The old QGC code tries to fetch `RC{n}_DZ` via `getParameterFact()`, gets null, logs an error, and bails out. The flight mode indicator stops working entirely.

It's surprising that the latest release version also belongs to "Old QGC". I'm confused as to why the stable version of QGC is still 5.0.8. I personally am not used to daily build versions because I find they differ significantly from the existing release versions. There is no smooth transition between them. After these parameters are deleted, the QGC flight mode configuration page can no longer highlight the current switch corresponding channel.

dakejahl:
Hence this fix

PR #13867, opened 2026-01-14
Fix QML undefined property assignment warnings

dakejahl:
Replace ternary null-guards with optional chaining for nested Fact property access

Follow up from https://github.com/mavlink/qgroundcontrol/pull/13866

Needs to be fully tested


DonLakeFlyer:
This is being caused by the recent change to initialize Facts to an invalid variant. My plan is to fix this in a generic way so that we don't need to be do all this insanity on the Qml side. I did an initial try which didn't work. I think I know how to do it now though. Let me try to take that approach first. Leave this around for now till I get that done.

PR #13871, opened 2026-01-19
SettingsGroupLayout: Fix separator display when using a Repeater

### LINKED ISSUE #13729: SettingsGroupLayout: Separators still not working correctly with invisible items
There is a hidden UDP autoconnect link which is causing this problem:
Image

DonLakeFlyer:
Fix #13729


DonLakeFlyer:
Still not thrilled with where this is at. But it gets rid of the bugs. I'll come back to it a some point.

PR #13890, opened 2026-01-23
Revert "fix: `Unable to assign [undefined]` errors"

DonLakeFlyer:
Reverts mavlink/qgroundcontrol#13866

I'm reverting this since I have a different global fix for it which doesn't require Qml modification


DonLakeFlyer:
@dakejahl FYI

PR #13891, opened 2026-01-23
Fact: Add support for fact not set by default

DonLakeFlyer:
* Fact defaults rawValue to 0 like before
* Whether Fact has been set or not is tracked in _rawValueIsNotSet bool
* _rawValueIsNotSet is used to determined when to show invalid value string
* This fixes problem where Qml code which directly reference a raw value (not by string) was generating undefined errors. But if you explicitly got after valueString you will get invalid value string support.
* Related to #13890, #13866, #13867

### DIFF
--- a/src/FactSystem/Fact.cc
+++ b/src/FactSystem/Fact.cc
@@ -50,6 +50,7 @@ Fact::Fact(const QString& settingsGroup, FactMetaData *metaData, QObject *parent
             const QVariant defaultValue = metaData->rawDefaultValue();
             QMutexLocker<QRecursiveMutex> locker(&_rawValueMutex);
             _rawValue = defaultValue;
+            _rawValueIsNotSet = false;
         }
     }
 
@@ -88,6 +89,7 @@ const Fact &Fact::operator=(const Fact& other)
     _name = other._name;
     _componentId = other._componentId;
     _rawValue = other._rawValue;
+    _rawValueIsNotSet = other._rawValueIsNotSet;
     _type = other._type;
     _sendValueChangedSignals = other._sendValueChangedSignals;
     _deferredValueChangeSignal = other._deferredValueChangeSignal;
@@ -136,6 +138,7 @@ void Fact::setRawValue(const QVariant &value)
                 QMutexLocker<QRecursiveMutex> locker(&_rawValueMutex);
                 if (typedValue != _rawValue) {
                     _rawValue = typedValue;
+                    _rawValueIsNotSet = false;
                     changed = true;
                 }
             }
@@ -197,6 +200,7 @@ void Fact::containerSetRawValue(const QVariant &value)
         QMutexLocker<QRecursiveMutex> locker(&_rawValueMutex);
         if (_rawValue != value) {
             _rawValue = value;
+            _rawValueIsNotSet = false;
             changed = true;
         }
         currentRaw = _rawValue;
@@ -352,9 +356,11 @@ QStringList Fact::selectedBitmaskStrings() const
 
 QString Fact::_variantToString(const QVariant &variant, int decimalPlaces) const
 {
-    if (!variant.isValid()) {
+    QMutexLocker<QRecursiveMutex> locker(&_rawValueMutex);
+    if (_rawValueIsNotSet) {
         return invalidValueString(decimalPlaces);
     }
+    locker.unlock();
 
     QString valueString;
 
--- a/src/FactSystem/Fact.h
+++ b/src/FactSystem/Fact.h
@@ -201,7 +201,8 @@ protected:
 
     QString _name;
     int _componentId = -1;
-    QVariant _rawValue; // QVariant::Invalid
+    QVariant _rawValue{0};
+    bool _rawValueIsNotSet = true;
     mutable QRecursiveMutex _rawValueMutex;
     FactMetaData::ValueType_t _type = FactMetaData::valueTypeInt32;
     FactMetaData *_metaData = nullptr;


DonLakeFlyer:
I will rebase once #13890 is finished.

@dakejahl, @rubenp02 Can you guys review and test? Probably wait on testing I rebase it to #13890.

dakejahl:
Sure thing. Ping me again when it's ready to test

rubenp02:
Same here!

DonLakeFlyer:
This is rebased now

dakejahl:
Looks good to me!
```
jake@pop-os:~/code/jake/qgroundcontrol$ qgc
QML debugging is enabled. Only use this in a safe environment.
     0.007 Using Qt multimedia with FFmpeg version 7.1.2 LGPL version 2.1 or later - qt.multimedia.ffmpeg - (unknown:0)
     0.934 Warning: Neither desktopId plugin parameter nor QGuiApplication::desktopFileName has been set. Please consider setting one of the two. - qt.positioning.geoclue2 - (unknown:0)
     1.283 Critical: Unable to start the client: "org.freedesktop.DBus.Error.AccessDenied" "Geolocation disabled for UID 1000" - qt.positioning.geoclue2 - (unknown:0)
     1.283 Warning: QGCPositionManager::_setPositionSource(QGCPositionManager::QGCPositionSource)::<lambda(QGeoPositionInfoSource::Error)> QGeoPositionInfoSource::AccessError - PositionManager.QGCPositionManager - (QGCPositionManager::_setPositionSource(QGCPositionManager::QGCPositionSource)::<lambda(QGeoPositionInfoSource::Error)>:206)
     3.172 Warning: QObject::disconnect: wildcard call disconnects from destroyed signal of MissionSettingsItem::unnamed - default - (unknown:0)
     3.399 Warning: Locale  "en_US"  not found in json - ComponentInformation.ComponentInformationTranslation - (ComponentInformationTranslation::getUrlFromSummaryJson:55)
     3.428 Warning: Download already in progress - Utilities.QGCCachedFileDownload - (QGCCachedFileDownload::download:141)
     3.428 Warning: RequestMetaDataTypeStateMachine::_requestFile QGCCachedFileDownload::download returned failure - Vehicle.ComponentInformationManager - (RequestMetaDataTypeStateMachine::_requestFile:413)
     8.434 Warning: Unknown 'function': "center" - Vehicle.ActuatorsConfig - (Actuators::parseJson:496)
     8.434 Warning: Unknown 'function': "center" - Vehicle.ActuatorsConfig - (Actuators::parseJson:496)
     8.434 Warning: Unknown 'function': "center" - Vehicle.ActuatorsConfig - (Actuators::parseJson:496)
     8.484 Warning: QObject::disconnect: wildcard call disconnects from destroyed signal of MissionSettingsItem::unnamed - default - (unknown:0)
     8.484 Warning: QObject::disconnect: wildcard call disconnects from destroyed signal of MissionSettingsItem::unnamed - default - (unknown:0)
     ```

rubenp02:
Might be missing some context here, but I'd say this is a bit of a workaround rather than the intended architectural pattern within Qt. Any reason why we can't change all the usages to agree with the QVariant::isValid policy instead?

Besides being the "canonical" Qt way and semantically proper, I'd say that not having every Fact start its life as a 0 int probably saves us from some bugs. It also makes it automatically clear from the outside whether a Fact has a value yet or not, which this doesn't as far as I can see. With this being the Fact System, semantic conflation should be avoided. And even if _rawValueIsNotSet were to be made public, new devs, specially those familiar with Qt, would have to go looking into the Fact API to see it, rather than just using the usual Qt idiom.

Maybe do something with the rawValue getter so that it remains compatible with the old standard of returning 0 for no value to the QML side, while the QVariant remains invalid? I'd still prefer changing all the usages and make this fully canonical with Qt standard practices though, but maybe this could be done in such a way that it keeps the old code working while allowing for a transition into QVariant::isValid.

rubenp02:
Tested and it builds and runs on Windows 11 with no issues (CMake Configure was extremely slow though? Did something change there recently?).

DonLakeFlyer:
> (CMake Configure was extremely slow though? Did something change there recently?).

@HTRamsey Would know if that were the case

HTRamsey:
Some of the dependencies were updated so maybe they were just slow to download. Could also be I did something to cause that when I created the build-config.json thing. I'll try it out soon and see

DonLakeFlyer:
A Fact never had the concept of the value not being set. From the start it was always defaulted to 0. Hence there is a massive amount of code that assumes that. Trying to clean all that up doesn't really create any value. As can be seen from the change to default to invalid, it creates a bug farm in 100s of places in the code that require cleanup. The concept of a Fact not being set is just not super useful in all cases. Having to constantly check for the value being invalid in both Qml and C++ code in any places that access just makes working with it annoyingly complex. Hence the concept of not set only comes to front when it's displayed to the end user. 

The internal code which uses Facts has always had existing mechanisms to indicate not set as needed in the places where it matters. In most places this is done by initializing to NaN. Also the reason for the FactGroup to have the 'telemetryAvailable' bool to indicate this group has not telemetry yet for it.

I know it's not super clean, but it's way simpler.

rubenp02:
> A Fact never had the concept of the value not being set. From the start it was always defaulted to 0. Hence there is a massive amount of code that assumes that. Trying to clean all that up doesn't really create any value. As can be seen from the change to default to invalid, it creates a bug farm in 100s of places in the code that require cleanup. The concept of a Fact not being set is just not super useful in all cases. Having to constantly check for the value being invalid in both Qml and C++ code in any places that access just makes working with it annoyingly complex. Hence the concept of not set only comes to front when it's displayed to the end user.
> 
> The internal code which uses Facts has always had existing mechanisms to indicate not set as needed in the places where it matters. In most places this is done by initializing to NaN. Also the reason for the FactGroup to have the 'telemetryAvailable' bool to indicate this group has not telemetry yet for it.
> 
> I know it's not super clean, but it's way simpler.

I get the pragmatism but I do honestly see it as a bit of technical debt, because 1) 0 is a legitimate (and very common) actual value, and 2) having QVariants always need to change type on runtime by design is a bit of a smell. In the QML side, undefined is the idiomatic value it makes sense for an an unset Fact to have. If it's truly annoying to always have to guard it, then maybe have a thin wrapper, but in my opinion it should never come as an incorrect value of possibly the wrong type directly from the backend.

> A Fact never had the concept of the value not being set.

But I think this shows the need for that. The errors caused by switching to invalid are all spots where a Fact existed but wasn't set. I stand by my opinion that all systems that store real world data, and one called Fact System probably more so, need a way clear-cut to tell if each datapoint contains legitimate data, even if they're guaranteed to under normal circumstances.

DonLakeFlyer:
> The errors caused by switching to invalid are all spots where a Fact existed but wasn't set.

You are absolutely correct. But you are missing the point that for majority of those spots it doesn't matter. And having to fix up big piles of code for the sake of internal purity and in the process making that code more complex for no end-user benefit doesn't seem worthwhile.

PR #13892, opened 2026-01-24
AutoPilotPlugins: RadioComponent: add Payload Power Switch RC channel

dakejahl:
Added RC payload power switch param [RC_MAP_PAY_SW](https://docs.px4.io/main/en/advanced_config/parameter_reference#RC_MAP_PAY_SW) to the Radio page


DonLakeFlyer:
Is this a new parameter or has it been around for a while. Thinking about back compat problems. If older firmwares don't have it this will cause a missing param error to pop.

dakejahl:
It's been around since Oct 2024
https://github.com/PX4/PX4-Autopilot/pull/23830

DonLakeFlyer:
All good then

PR #13896, opened 2026-01-24
Fix Multi-vehicle section background 

gillamkid:
# Description
fixes https://github.com/mavlink/qgroundcontrol/issues/13919
fixes https://github.com/mavlink/qgroundcontrol/issues/13921
Made the Fly View items for Multi-vehicle control look a bit better.

# Before/After
before
after


gillamkid:
@DonLakeFlyer , you were the one that changed the background recently to be transparent. Thoughts on this PR?

DonLakeFlyer:
A

> you were the one that changed the background recently to be transparent

I think that was a screw up.

I reviewed and it all looks good.

PR #13897, opened 2026-01-24
actuators: add center to Actuators::parseJson

dakejahl:
Noticed warnings in the console
```
     8.434 Warning: Unknown 'function': "center" - Vehicle.ActuatorsConfig - (Actuators::parseJson:496)
     8.434 Warning: Unknown 'function': "center" - Vehicle.ActuatorsConfig - (Actuators::parseJson:496)
     8.434 Warning: Unknown 'function': "center" - Vehicle.ActuatorsConfig - (Actuators::parseJson:496)
```

The PWM_AUX_CENT${i} parameters came in with PX4 in https://github.com/PX4/PX4-Autopilot/pull/25897.

What is odd is we get the warning in QGC, but the UI still works properly. Then what is the purpose of this code?

### DIFF
--- a/src/Vehicle/Actuators/ActuatorOutputs.h
+++ b/src/Vehicle/Actuators/ActuatorOutputs.h
@@ -55,6 +55,7 @@ public:
         Disarmed,
         Minimum,
         Maximum,
+        Center,
         Failsafe
     };
 
--- a/src/Vehicle/Actuators/Actuators.cc
+++ b/src/Vehicle/Actuators/Actuators.cc
@@ -490,6 +490,8 @@ bool Actuators::parseJson(const QJsonDocument &json)
                     function = ChannelConfig::Function::Minimum;
                 } else if (functionStr == "max") {
                     function = ChannelConfig::Function::Maximum;
+                } else if (functionStr == "center") {
+                    function = ChannelConfig::Function::Center;
                 } else if (functionStr == "failsafe") {
                     function = ChannelConfig::Function::Failsafe;
                 } else if (functionStr != "") {


DonLakeFlyer:
ASking @bkueng to take a look since he wrote all this

bkueng:
`center` is not valid according to https://github.com/mavlink/mavlink/blob/master/component_metadata/actuators.schema.json#L65, and should rather be removed from PX4 (https://github.com/PX4/PX4-Autopilot/pull/25897/changes#diff-cee35bfc5ae2fbfa111d93a705d66e7422acbf017cd97c44c234e1898da354fbR241)
@ttechnick fyi

ttechnick:
I`m not fully following you @bkueng.

> `center` is not valid according to https://github.com/mavlink/mavlink/blob/master/component_metadata/actuators.schema.json#L65,

So we first have to adjust it in mavlink, then in QGC?

---
> and should rather be removed from PX4 (https://github.com/PX4/PX4-Autopilot/pull/25897/changes#diff-cee35bfc5ae2fbfa111d93a705d66e7422acbf017cd97c44c234e1898da354fbR241) 

As in remove the feature from PX4 completely? Not really an option for me, as many different people and companies need this feature.

bkueng:
What I meant was to remove it from the metadata only (the linked source) not the complete feature.

ttechnick:
Oh I see. When I remove it from the metadata you linked, the graphical interface to change the values does not work anymore. But it fixes the bug. 
However I would prefer to invest the time to fix this properly and have the GUI, rather then having a buried param somewhere.  Or do you think this is unrealistic/impossible @bkueng ?

dakejahl:
Following up -- I want to add DShot Pole count per channel and have it appear in the actuators page in QGC. @bkueng looking for an answer here, how do we extend the UI for options like `center` and `pole count` without changing the schema?

bkueng:
I think you'll need to adjust the script to not include the 'function' here: https://github.com/ttechnick/PX4-Autopilot/blob/f92937aecd9da15cf47d1bdb8738bf91133538a0/Tools/module_config/generate_actuators_metadata.py#L253

bkueng:
If it's a specific driver only you can also do it like this: https://github.com/PX4/PX4-Autopilot/blob/main/src/drivers/pca9685_pwm_out/module.yaml#L11-L16

dakejahl:
PX4 PR for the fix is here
https://github.com/PX4/PX4-Autopilot/pull/26503

PR #13900, opened 2026-01-28
QGCTabBar: Fix for Repeater children in RowLayout

dakejahl:
### Problem
The custom QGCTabBar (a18778ae2) assumes all RowLayout children are buttons. When a Repeater is used, the Repeater itself is a child of the layout, causing `Cannot assign to non-existent property "checked"` and `Cannot append QQuickRepeater to a QML list of QQuickAbstractButton` warnings.                                                      
                                                                                                                                                                                                           
### Solution
Filter control.children to only operate on AbstractButton instances, and populate ButtonGroup via onChildrenChanged instead of binding directly to control.children.

### DIFF
--- a/src/QmlControls/QGCTabBar.qml
+++ b/src/QmlControls/QGCTabBar.qml
@@ -16,23 +16,33 @@ RowLayout {
 
     property bool _preventCurrentIndexBindingLoop: false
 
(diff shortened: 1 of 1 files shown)


dakejahl:
This isn't right

PR #13901, opened 2026-01-28
Bump mavlink to latest

dakejahl:
Bumps mavlink to latest from master. I am opening a PR soon which depends on the new ESC_EEEPROM message


dakejahl:
Gripper naming got reverted in https://github.com/mavlink/mavlink/pull/2393

PR #13902, opened 2026-01-28
Fix parameter search when "Show modified only" is checked

dakejahl:
### Problem
Parameter search not filtering results when typing with the "Show modified only" checkbox enabled

### Solution
Added the same _externalBeginResetModel guard that clear() already uses

### DIFF
--- a/src/QmlControls/ParameterEditorController.cc
+++ b/src/QmlControls/ParameterEditorController.cc
@@ -97,7 +97,7 @@ void ParameterTableModel::beginReset()
 {
     _resetNestingCount++;
 
-    if (_resetNestingCount == 0) {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I haven't dug deeply into there. But does this need nested reset tracking like ObjectListModelBase.cc has so that this will happen automatically instead of whack-a-mole on places where bool checks are missing in other calls? It all familiar to the problems I have with QmlObjectListModel and solved once and for all with the changes there.

dakejahl:
Implemented the nesting count as suggested. Tested functional.

DonLakeFlyer:
Thanks Jake

PR #13904, opened 2026-01-28
LogDownloadController: reverse log list

dakejahl:
### Problem

When you have a lot of logs you have to scroll down to select the latest log to download

### Solution

Present the log list in reverse order


DonLakeFlyer:
CoPilot is angry at various things related to the reversal? Wouldn't it be possible to manage the list in reverse order from the start? Add them in reverse order as they come?

dakejahl:
> Wouldn't it be possible to manage the list in reverse order from the start? Add them in reverse order as they come?

I did try that initially, but co-pilot complained about some things... I'll come back to this, thought this would be an easy one :sweat_smile:

PR #13915, opened 2026-01-31
Joystick: Full support for manual control extensions

DonLakeFlyer:
Also:
* Major cleanup on logging
* Much better settings read/save with validation
* Changes were to comprehensive to maintain conversion of legacy joystick settings. Joysticks will need re-calibration and enabling
* Settings UI still sucks. I will work on that in another pull

Screenshot 2026-01-31 at 3 28 45 PM


DonLakeFlyer:
FYI: Not ready for testing yet. Waiting on Copilot review and then my own re-testing after I fix up whatever it finds.

DonLakeFlyer:
FYI: @Niki-dev12, @ikalnytskyi, @HTRamsey. Stay tuned for when it's ready for testing.

src/Joystick/Joystick.h
```diff
@@ -115,16 +115,30 @@ class Joystick : public QThread
         int center = 0;
         int deadband = 0;
         bool reversed = false;
+
+        void reset() {
+            min = -32767;
```
HTRamsey: AxisMin, AxisMax

DonLakeFlyer:
@HTRamsey I ended up with a nasty merge conflict with your `movedSignificantly` change. For now I've removed it. I see the problem you are trying to fix. But it's not quite right. It should be in both stick detect and stick min and should use _calValidMin/MaxValue to determine whether the stick has moved for enough away from center to consider it's position as a min/max value. I'll create an Issue for this once this pull goes in and fix it up in a separate pull

DonLakeFlyer:
Important things to check:
* After calibration that the mapped indicators show that the values are mapped to the right channels are going in the right direction up/down, left/right as appropriate for function.
* Quit QGC, restart and recheck mapped indicators to make sure load/save is working correctly.
* Changing settings like enabling/disabling a new manual control extension should require you to re-calibrate.
* Make sure you don't see any warnings in the logs

HTRamsey:
Oh yeah that's fine. I was having trouble figuring out if my calibration problem was SDL related or something else. Also FYI for bottom LS and bottom RS, you seem to have to go the other way from neutral a bit and then go to the target extreme to get it to register correctly.

DonLakeFlyer:
> Also FYI for bottom LS and bottom RS, you seem to have to go the other way from neutral a bit and then go to the target extreme to get it to register correctly.

Hmm, I never noticed that in the like 1000 joystick cals I went through to test this.

src/Vehicle/VehicleSetup/RemoteControlCalibration.qml
```diff
@@ -41,83 +41,71 @@ ColumnLayout {
...
+                        { name: qsTr("Aux 3"),  extensionEnabled: controller.aux3ExtensionEnabled,  mapped: controller.aux3ExtensionChannelMapped,  value: controller.adjustedAux3ExtensionChannelValue,    deadband: controller.aux3ExtensionDeadband },
+                        { name: qsTr("Aux 4"),  extensionEnabled: controller.aux4ExtensionEnabled,  mapped: controller.aux4ExtensionChannelMapped,  value: controller.adjustedAux4ExtensionChannelValue,    deadband: controller.aux4ExtensionDeadband },
+                        { name: qsTr("Aux 5"),  extensionEnabled: controller.aux5ExtensionEnabled,  mapped: controller.aux5ExtensionChannelMapped,  value: controller.adjustedAux5ExtensionChannelValue,    deadband: controller.aux5ExtensionDeadband },
+                        { name: qsTr("Aux 6"),  extensionEnabled: controller.aux6ExtensionEnabled,  mapped: controller.aux6ExtensionChannelMapped,  value: controller.adjustedAux6ExtensionChannelValue,    deadband: controller.aux6ExtensionDeadband }
+                    ]
```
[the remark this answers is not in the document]
DonLakeFlyer: Tracked in #13917 and will be fixed in alter pull

src/Vehicle/VehicleSetup/RemoteControlCalibration.qml
```diff
@@ -41,83 +41,71 @@ ColumnLayout {
...
 
-                RowLayout {
-                    Layout.fillWidth: true
+                Repeater {
+                    model: [
+                        { name: qsTr("Pitch"),      mapped: controller.pitchChannelMapped,      value: controller.adjustedPitchChannelValue,      deadband: controller.pitchDeadband },
+                        { name: qsTr("Roll"),       mapped: controller.rollChannelMapped,       value: controller.adjustedRollChannelValue,       deadband: controller.rollDeadband },
+                        { name: qsTr("Yaw"),        mapped: controller.yawChannelMapped,        value: controller.adjustedYawChannelValue,        deadband: controller.yawDeadband },
+                        { name: qsTr("Throttle"),   mapped: controller.throttleChannelMapped,   value: controller.adjustedThrottleChannelValue,   deadband: controller.throttleDeadband }
+                    ]
```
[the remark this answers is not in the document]
DonLakeFlyer: Tracked in #13917 and will be fixed in alter pull

DonLakeFlyer:
This is ready for testing now. @HTRamsey Can you help test RC transmitter.

src/Vehicle/Vehicle.cc
```diff
@@ -3797,24 +3753,26 @@ void Vehicle::sendJoystickDataThreadSafe(float roll, float pitch, float yaw, flo
...
-    }
-
-    if (std::isfinite(gimbalYaw)) {
-        extensions |= 2;
+    // Scale and set extension bits/values
+    float incomingExtensionValues[] = { pitchExtension, rollExtension, aux1, aux2, aux3, aux4, aux5, aux6 };
+    uint16_t outgoingExtensionValues[std::size(incomingExtensionValues)];
+    for (size_t i = 0; i < std::size(incomingExtensionValues); i++) {
+        int16_t scaledValue = 0;
+        if (qIsNaN(incomingExtensionValues[i])) {
```
ikalnytskyi: @DonLakeFlyer It should be inverted condition: you want to set an extension bit if an incoming value **is not** `NaN`.

DonLakeFlyer: Ah, last minute CoPilot code review thing I screwed up.

DonLakeFlyer: Fixed. I managed to screw it up from review without re-testing.

ikalnytskyi:
@DonLakeFlyer Just tried this patch, and it seems like extensions are not properly saved if they are randomly enabled. E.g., if you just enable `aux1`, but nothing else. If you restart a QGC, it will start complaining about missing mapping and the joystick will be unresponsive even though the Joystick panel will keep saying it's enabled and calibrated (should probably got reset to uncalibrated?). 

  I also have a question, what is this blue bar? It doesn't seem to be shown after every calibration I performed.
  image

DonLakeFlyer:
> what is this blue bar?

Deadband

> it seems like extensions are not properly saved if they are randomly enabled. E.g., if you just enable `aux1`, but nothing else. If you restart a QGC, it will start complaining about missing mapping and the joystick will be unresponsive even though the Joystick panel will keep saying it's enabled and calibrated (should probably got reset to uncalibrated?).

Crap I tested all that. I must have screwed other stuff up in response to CoPilot review. You are correct that the joystick cal should have reset and I thought I had that all working. I'll fix it. Thanks for testing.

DonLakeFlyer:
> if you just enable `aux1`, but nothing else. If you restart a QGC, it will start complaining about missing mapping and the joystick will be unresponsive even though the Joystick panel will keep saying it's enabled and calibrated (should probably got reset to uncalibrated?).

@ikalnytskyi Hmm, this doesn't repro for me. I did this:
* My current setting had no extensions enabled
* I enabled Aux 1
* Recalibrated
* All working normally
* Reboot
* Still all working normally

I also tried reproing it after I did a clear settings. Still no luck.

There must be something you are doing differently. Also can you let me know exactly what the warning output is. That will at least help me to fix the problem where I should be catching a bit settings load.

DonLakeFlyer:
I found the problem where the extensions are not validated to be mapped at load time. I only verified attitude control mappings. Stupid. I'll fix that after we figure out how to get the mapping bug reproed and fixed.

ikalnytskyi:
@DonLakeFlyer I'm having the following warning, and it's looping. Again, after calibration -- everything works fine. Once I restart QGC, it's what I get:

```
Warning: Internal Error: Missing aux1 extension axis function mapping! - Joystick.Joystick - (Joystick
::_handleAxis:864)
```

However, once I manually removed _everything_ related to joystick from `~/.config/QGroundControl/QGroundControl Daily.ini`, it started to work great. So it may be related to some settings format migration. :thinking:

DonLakeFlyer:
Hmm, there shouldn't have been any overlap between old settings to new. There is no migration. Still worried there is something lurking, but not 100% sure of that any more. I'll fix the missing mappings should reset thing and then it should be good from the Joystick side.

Just need to hear back from the RC transmitter side and we'll be good to go on this pull.

DonLakeFlyer:
> I'll fix the missing mappings should reset thing and then it should be good from the Joystick side.

Done.

@HTRamsey Can you help with RC cal testing?

DonLakeFlyer:
Need to fix Center stick is zero throttle. Should be rover only.

jnomikos:
@DonLakeFlyer One issue I notice when testing on android is that when you have two joysticks and switch between one and the other, the switching doesn't seem to work on second attempt. 


https://github.com/user-attachments/assets/649e59a7-daab-4209-a08b-928681fb32b7

Issue #13927, opened 2026-02-04
ArduCopter, confusing UI interaction when on last waypoint and adjusting current waypoint


jnomikos:
### Have you checked the latest release for fixes?

- [x] Yes, I’ve tested on the latest release

### Description

**Expected Behavior**

Some users do not pause the mission before adjusting the current waypoint and expect that to always work. When a user is flying a mission and clicks on a waypoint to "Adjust current waypoint to X", they expect the vehicle to perform that action without any further intervention, as long as the vehicle is flying.

**Current Behavior**
Normally, when flying, adjusting the current waypoint without pausing the mission works. But if you are on the very last waypoint in a mission which does not have a RTL, adjusting the current waypoint does not make the vehicle start moving to that waypoint. Only if you pause the mission before adjusting the current waypoint does it work.

**Steps To Reproduce**
1. Plan a SITL mission which does not have a RTL
2. Have the vehicle go to the last waypoint
3. Adjust the current waypoint without pausing the mission
4. Vehicle will not start moving towards the adjusted waypoint

**Additional Details**
I am testing on ArduCopter 4.5.6 which is somewhat old. It might be different on newer versions

**Proposed Solution**
Perhaps streamline things by automatically sending a pause command in the current waypoint adjustment (maybe only when on the last waypoint)

### Platform

Linux

### Flight Stack

ArduPilot

### System Information

QGC Version: Master, Stable 4.4
QGC Build: Daily, Stable 4.4
Operating System: Ubuntu 24, Android
Flight Controller: Cube Orange, SITL
Autopilot Software: ArduCopter 4.5.6

### Log Files and Screenshots

https://github.com/user-attachments/assets/3fdb4d33-4473-4f90-8083-050a408b7c0e

jnomikos:
@DonLakeFlyer FYI I am working on potential solution right now

jnomikos:
Draft PR is here: https://github.com/mavlink/qgroundcontrol/pull/13928

I need to tweak how the logic is handled so it doesn't pause after vehicle has landed in a normal mission with an RTL. 

Edit: tweaked

DonLakeFlyer:
If QGC is doing the same thing every time when modifying a single waypoint position isn't this a firmware bug that it all works correctly except for on the last one?

jnomikos:
> If QGC is doing the same thing every time when modifying a single waypoint position isn't this a firmware bug that it all works correctly except for on the last one?

[...]

PR #13934, opened 2026-02-06
setMissionCurrentSequence: Set "Reset Mission" flag to true

jnomikos:
## Description
When changing the current mission waypoint, the "Reset Mission" item on the MAVLink command is now set to true. 

This will reset all jump counters and changes mission state "completed" to be "active" or "paused". Having "reset" set to true fixes the problem on ArduCopter when the final mission item is completed in a mission without RTL.

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [ ] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [x] PX4
- [x] ArduPilot

Tested SITL on PX4 v1.17.0-beta1, ArduCopter master, and ArduCopter 4.5.6.

My test consisted of plotting a simple mission and changing current waypoint in various stages of the mission

## Checklist

## Related Issues
#13927

---


DonLakeFlyer:
Doesn't this mean that if I have paused a mission and then change the active waypoint the mission it suddenly going to start going on it's own. I'm not sure that is a good thing?

jnomikos:
@DonLakeFlyer 
> Doesn't this mean that if I have paused a mission and then change the active waypoint the mission it suddenly going to start going on it's own. I'm not sure that is a good thing?

Definitely a good point and important for the vehicle to stay paused even if you change current waypoint.

From my investigation, on ArduCopter 4.6.3 SITL and PX4 v1.17.0-beta1 SITL, the vehicle does **not** unpause if you change waypoint with "Reset Mission" set to true.

Nothing in the MAVLink spec for that command mentions that it would unpause the mission if that "Reset Mission" param is set true either.

According to the "Reset Mission" description in [MAV_CMD_DO_SET_MISSION_CURRENT](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_MISSION_CURRENT), "Resets jump counters to initial values and changes mission state "completed" to be "active" or "paused".".

The only area that I can see (based on the MAVLink description) that is up to the choice of the Autopilot (if they act in accordance with the MAVLink description) is how to handle if the autopilot is in a completed state (not paused or active). 

Copter 4.6.3 SITL test:
https://github.com/user-attachments/assets/9d744dd9-de45-464a-bf6e-3ab55c727e5b


On PX4 v1.17.0-beta1:
https://github.com/user-attachments/assets/839d2617-c08d-4f60-80a0-df557af1b9ef

DonLakeFlyer:
> Resets jump counters to initial values

I don't quite get what this really means?

jnomikos:
> > Resets jump counters to initial values
> 
> I don't quite get what this really means?

(This is my interpretation of it, but it could be wrong)

I believe it is referring to the [MAV_CMD_DO_JUMP](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_JUMP) mission item. MAV_CMD_DO_JUMP can be set with a repeat count to repeat a portion of a mission a specified number of times. I think you can have a mission item do MAV_CMD_DO_JUMP.

If "Reset Mission" is set to true when sending [MAV_CMD_DO_SET_MISSION_CURRENT](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_MISSION_CURRENT), I think the jump counter in MAV_CMD_DO_JUMP will be reset. 

So, if your mission includes a MAV_CMD_DO_JUMP near the end to repeat part of the mission a certain number of times, enabling Reset Mission will reset that command’s internal repeat counter back to its original value. This means the repeated section will run again the full number of times specified in the mission, even if it had already partially completed those repetitions before the reset.

DonLakeFlyer:
Here is what Claude has to say about this:
# Code Review: Branch `jnomikos-dev-do-set-mission-set-current`

**Commit:** `b982d1ee86` — *setMissionCurrentSequence: Set "Reset Mission" flag to true*

**Change:** In `src/Vehicle/Vehicle.cc` (lines 2364–2365), `param2` of `MAV_CMD_DO_SET_MISSION_CURRENT` is set to `true` (i.e., `1.0f`), enabling the "reset mission" flag.

## What the change does

Per the [MAVLink spec](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_MISSION_CURRENT), setting param2 to `1` (MAV_BOOL_TRUE) when calling `MAV_CMD_DO_SET_MISSION_CURRENT`:

1. Resets DO_JUMP repeat counters to their initial values.
2. Changes mission state from `MISSION_STATE_COMPLETE` to `MISSION_STATE_ACTIVE` or `MISSION_STATE_PAUSED`, allowing the mission to resume.

Previously, param2 defaulted to `0.0f`, meaning jump counters were not reset and a completed mission could not be restarted by setting the current sequence.

## Issues / Concerns

### 1. Implicit `bool` → `float` conversion

The function signature is `float param2`. Passing `true` works (implicit conversion to `1.0f`), but it's inconsistent with param1 which explicitly uses `static_cast<uint16_t>(seq)`. For clarity and consistency, consider passing `1.0f` explicitly or using a named constant. The current form obscures the actual numeric value being sent.

### 2. Always resetting may be too aggressive

This unconditionally resets jump counters every time `setCurrentMissionSequence` is called. If a user is simply skipping to a different waypoint mid-mission (without wanting to reset DO_JUMP counters), this behavior change could be surprising. Consider whether this should be:

- A parameter to `setCurrentMissionSequence` so callers can control it
- Only applied when the intent is truly to "restart" the mission

### 3. PX4 does not support this command

The PX4FirmwarePlugin (`src/FirmwarePlugin/PX4/PX4FirmwarePlugin.h`, line 100) marks `MAV_CMD_DO_SET_MISSION_CURRENT` as `UNSUPPORTED`, so on PX4 the lambda fallback (`mavlink_msg_mission_set_current_pack_chan`) is used instead. That deprecated message has no "reset" param, so **this change has no effect on PX4 vehicles**. This isn't a bug per se, but the behavior will differ between firmware types — worth documenting or noting in the PR.

### 4. Comment is overly long for a single line

The inline comment duplicates the MAVLink spec verbatim and extends well past any reasonable line width. A shorter comment like `// reset jump counters and mission state` with a reference to the MAVLink docs would be more readable and consistent with the project's `CODING_STYLE.md`.

### 5. No corresponding change to the deprecated fallback path

The lambda fallback (lines 2351–2359) using `mavlink_msg_mission_set_current_pack_chan` doesn't support a reset flag. This is inherent to the deprecated message format, but it creates a silent behavioral inconsistency — worth a brief comment in the fallback lambda noting that reset is not supported via this path.

## Summary

The intent is sound — resetting jump counters and mission state when setting a mission current item is useful for restarting missions. The main concerns are:

1. Consider whether unconditional reset is always desired
2. Use an explicit numeric value instead of `bool` → `float` implicit conversion
3. Trim the comment length

DonLakeFlyer:
@jnomikos Can you comment on what Claude says?

2 - Seems like an interesting thing with regards to not using it all the time. This was to handle a specific problem which only occured at the end of mission wasn't it. Maybe limiting to that case would be a safe change?

3 - That's a strange on. I've never seen that code before. I need to figure out what it does.

DonLakeFlyer:
@rubenp02 You've been deep into missions lately. What do you think?

rubenp02:
> @rubenp02 You've been deep into missions lately. What do you think?

For ArduPilot at least (but I suspect other firmwares will be the same), this isn't right. We 100% need an easy way to reset the mission (I say this because I don't think this option currently exists, and if it does, it's buried deep enough I've never seen it), and perhaps to make it an user-configurable trigger upon completing the mission. But changing waypoints is just changing waypoints. I do it constantly not just when I've landed and want to fly the mission again, but also when I want to take a shortcut or repeat some part of the mission. I fully expect jump counters to stay as they are when doing this.

jnomikos:
These are very good points. 

I also am thinking that maybe making it a conditional thing that only happens if we have reached the very last waypoint in a non-rtl mission would be good enough. That way, if you are in the middle of a mission and change waypoints, jump counters don't get reset. 


> @jnomikos Can you comment on what Claude says?

Sure.

> 1. Implicit bool → float conversion
> The function signature is float param2. Passing true works (implicit conversion to 1.0f), but it's inconsistent with param1 which explicitly uses static_cast<uint16_t>(seq). For clarity and consistency, consider passing 1.0f explicitly or using a named constant. The current form obscures the actual numeric value being sent.


I think the implicit conversion is actually clearer because the command param is a true or false, but we can remove it if needed.

> 2. Always resetting may be too aggressive
> 
> This unconditionally resets jump counters every time setCurrentMissionSequence is called. If a user is simply skipping to a different waypoint mid-mission (without wanting to reset DO_JUMP counters), this behavior change could be surprising. Consider whether this should be:
> 
>     A parameter to setCurrentMissionSequence so callers can control it
>     Only applied when the intent is truly to "restart" the mission

This one I agree with. We can make it conditional to when the drone has reached the final waypoint and there's no RTL in the mission. That way, jump counters never get reset.


> 3. PX4 does not support this command
> 
> The PX4FirmwarePlugin (src/FirmwarePlugin/PX4/PX4FirmwarePlugin.h, line 100) marks MAV_CMD_DO_SET_MISSION_CURRENT as UNSUPPORTED, so on PX4 the lambda fallback (mavlink_msg_mission_set_current_pack_chan) is used instead. That deprecated message has no "reset" param, so this change has no effect on PX4 vehicles. This isn't a bug per se, but the behavior will differ between firmware types — worth documenting or noting in the PR.

I was not aware of this. PX4FirmwarePlugin's anyVersionSupportsCommand function determines MAV_CMD_DO_SET_MISSION_CURRENT as unsupported. So, this code change shouldn't affect PX4. 

> 5. No corresponding change to the deprecated fallback path
> 
> The lambda fallback (lines 2351–2359) using mavlink_msg_mission_set_current_pack_chan doesn't support a reset flag. This is inherent to the deprecated message format, but it creates a silent behavioral inconsistency — worth a brief comment in the fallback lambda noting that reset is not supported via this path.

I think this is probably fine, because behavioral inconsistency already exists regardless. PX4 (which doesn't support the MAV_CMD_DO_SET_MISSION_CURRENT message) behaves completely different to ArduPilot after completing the last waypoint in a mission without an RTL.

DonLakeFlyer:
> PX4FirmwarePlugin's anyVersionSupportsCommand function determines MAV_CMD_DO_SET_MISSION_CURRENT as unsupported.

I remember what this is for now. This is for determining whether to use MISSION_SET_CURRENT or MAV_CMD_DO_SET_MISSION_CURRENT.

rubenp02:
> I also am thinking that maybe making it a conditional thing that only happens if we have reached the very last waypoint in a non-rtl mission would be good enough. That way, if you are in the middle of a mission and change waypoints, jump counters don't get reset.

And only if you're changing to the 1st waypoint, no? Or am I misunderstanding something? Either way, I don't think making this implicit is the right UX call. Just have a reset mission option! Let changing waypoints be just that.

jnomikos:
> > I also am thinking that maybe making it a conditional thing that only happens if we have reached the very last waypoint in a non-rtl mission would be good enough. That way, if you are in the middle of a mission and change waypoints, jump counters don't get reset.
> 
> And only if you're changing to the 1st waypoint, no? Or am I misunderstanding something? Either way, I don't think making this implicit is the right UX call. Just have a reset mission option! Let changing waypoints be just that.

I like this idea. @DonLakeFlyer What do you think?

DonLakeFlyer:
I think that makes sense if both you and Ruben are happy with that

jnomikos:
Closing this PR since the resolution is that we should probably add a "Reset Mission" option instead. I’ll open a new PR for that change once it’s implemented.

Issue #13960, opened 2026-02-10
USB Video does not size according to videoFit setting


gillamkid:
### Have you checked the latest release for fixes?

- [x] Yes, I’ve tested on the latest release

### Description

**Issue**
In the video settings there is a `Video Screen Fit` option that can be set to `Fit Width`, `Fit Height`, `Fill`, or `No Crop`. When the video is a gstreamer type (ex. MPEG-TS or RTSP) the video sizes itself as appropriately. When the video type is a web cam or USB camera, the video sizing has no change when the `Video Screen Fit` option is changed.

**Steps To Reproduce**
1. go to video settings and set `Video Screen Fit` to `Fit Height`
2. set video source to a webcam or USB camera
3. resize the QGC window so it is short and very wide
4. instead of shrinking the video and having black bars on the right and left of it, the video is stretched so it takes up the whole screen
grid lines do not appear

### Platform

Linux

### Flight Stack

None

### System Information

I tested on QGC Daily

Issue #13959, opened 2026-02-10
USB Video does not show Grid Lines


gillamkid:
### Have you checked the latest release for fixes?

- [x] Yes, I’ve tested on the latest release

### Description

**Issue**
When `QGroundControl.settingsManager.videoSettings.gridLines` is `true`, grid lines will be shown when the video is a gstreamer type (ex. MPEG-TS or RTSP) but not when the video type is a web cam or USB camera.

**Steps To Reproduce**
1. go to video settings and set `Video Grid Lines` to `On`
2. set video source to a webcam or USB camera
3. grid lines do not appear

### Platform

Linux

### System Information

Tested on Daily build

PR #13961, opened 2026-02-10
Usb video fixes

gillamkid:
## Description
fixes https://github.com/mavlink/qgroundcontrol/issues/13959
and fixes https://github.com/mavlink/qgroundcontrol/issues/13960

---

### DIFF
--- a/src/FlyView/FlightDisplayViewUVC.qml
+++ b/src/FlyView/FlightDisplayViewUVC.qml
@@ -7,6 +7,8 @@ Rectangle {
     id:                 _root
     width:              parent.width
     height:             parent.height
+    implicitWidth:      videoOutput.implicitWidth
+    implicitHeight:     videoOutput.implicitHeight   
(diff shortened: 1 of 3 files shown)


DonLakeFlyer:
How come this requires so much shuffling around of code for what would seem like a simple change?

gillamkid:
@DonLakeFlyer I moved the USB video QML to the same place where the GStreamer video QML is. This makes so the video sizing/overlay QML used for the GStreamer video item can also be applied when using the USB video item. As far as the question "How come this requires so much shuffling around of code", I feel the shuffling I did was both a bug-fixing improvement as well as organization improvement. I tried to keep the diff minimal and not add much excessive re-organizing beyond what was being touched to fix the issue.

DonLakeFlyer:
> I moved the USB video QML to the same place where the GStreamer video QML is

Ah, ok. Now I get it

PR #13965, opened 2026-02-10
Fix launch location indicator position for Horizontal Compass layout

DieBorr:
When using the "Horizontal Compass & Attitude" instrument panel, the launch location indicator ("L") was rendered using parent.width / 2 as its translation radius, which caused it to appear outside, overlapping the attitude widget.

image
image

## Type of Change

- [ X] Bug fix (non-breaking change that fixes an issue)

## Testing

- [ X] Tested locally
- [X ] Tested with simulator (SITL)

### Platforms Tested

- [x ] Linux


DieBorr:
There is a much better approach

Issue #13952, opened 2026-02-09
ArduPilot sensor calibration popups still show on disconnect which causes segfault


jnomikos:
### Have you checked the latest release for fixes?

- [x] Yes, I’ve tested on the latest release

### Description

**Expected Behavior**
If disconnect occurs, sensor calibration screen and open popups related to it should close

**Current Behavior**
Popups stay open on disconnect which triggers crash

**Steps To Reproduce**
1. Go to Sensors page
2. Click on any of the sensors which show a popup
3. Disconnect vehicle and notice the popups stay which triggers a crash when interacting with them

**Additional Details**
Add any extra info (errors, warnings, etc.) here.

### Platform

Linux

### Flight Stack

ArduPilot

### System Information

    QGC Version: Master
    QGC build: Built from mastere
    Operating System: Ubuntu 24.04
    Flight Controller: Cube Orange Plus
    Autopilot Software (with version): ArduCopter 4.5.6

### Log Files and Screenshots

https://github.com/user-attachments/assets/edc72cfa-dcc5-44f9-be9f-f178b4a58e0e

DonLakeFlyer:
I see what's going on now and this is a really bad bug caused by a bad interface into how you use QGCPopupDialog. It's across the board throughout the product.

DonLakeFlyer:
@jnomikos Can you test out the pull?

jnomikos:
@DonLakeFlyer I just tested it and it fixed the sensor popup issue.

jnomikos:
@DonLakeFlyer 
I tested some more on other areas in the sensors screen and there are still popups that appear on disconnect, but none of the remaining ones cause crashes.

Areas in sensor popup I still see popups remaining on disconnect:

"Level horizon:" (does not cause crash but remains on disconnect)
"Calibrate gyro: " (does not cause crash but remains on disconnect)
"Pressure:"  (does not cause crash but remains on disconnect)
"CompassMot:"  (does not cause crash but remains on disconnect)

DonLakeFlyer:
Yup those are the mainWindow.showSimpleMessageDialog calls I mentioned in the pull. I edited the text on the pull so this doesn't get closed until those are fixed as well.

PR #13977, opened 2026-02-13
Fix QGCPopupDialog usage pattern such that it closes when owner is gone

DonLakeFlyer:
* You now show dialogs using the QGCPopupDialogFactory pattern
* This way if the owner of the dialog goes away while the dialog is being displayed it will close the dialog
* The prevents crashes from orphaned dialogs
* Related to #13952

There are still other cases that can cause similar problems which I will deal with in another pull:
* Maybe DropPanels
* Maybe other stuff calling into MainWindow which uses mainWindow as parent

### DIFF
--- a/custom-example/src/CustomPlugin.cc
+++ b/custom-example/src/CustomPlugin.cc
@@ -287,9 +287,8 @@ CustomOverrideInterceptor::CustomOverrideInterceptor()
 QUrl CustomOverrideInterceptor::intercept(const QUrl &url, QQmlAbstractUrlInterceptor::DataType type)
 {
     switch (type) {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@jnomikos I added the fixes for the showMessageDialog calls. Can you try this again. I don't have the ability to test the specific examples you gave right now. I'll test again tomorrow.

jnomikos:
> @jnomikos I added the fixes for the showMessageDialog calls. Can you try this again. I don't have the ability to test the specific examples you gave right now. I'll test again tomorrow.

Just tested. The new changes fixed all of the examples I gave. Thanks!

DonLakeFlyer:
@jnomikos Thanks for testing

Issue #668, opened 2014-05-22
Info View cannot be shown in the Mission perspective


Susurrus:
I don't know what's happened within the last few months, but the whole widgets and perspectives set up has been pretty messed up for regular UASes. There are a host of issues with this.

This issue pertains to using the Info View widget in the Mission perspective. It cannot currently be done. If you detach the Info View widget from the Flight perspective (where it can appear), and switch to the Mission perspective, you can't dock it anywhere.

Susurrus:
It can now.

PR #13996, opened 2026-02-17
feat(AnalyzeView): add MAVLink FTP log download for PX4 .ulg log files

dakejahl:
## Feature Description

Adds MAVLink FTP-based log download for PX4 vehicles, providing faster and more
reliable log transfers than the legacy LOG_REQUEST_DATA protocol. When a PX4
vehicle reports MAV_PROTOCOL_CAPABILITY_FTP, the log download page automatically
uses MAVFTP to list and download .ulg files from `/fs/microsd/log`.

The legacy LOG_REQUEST protocol remains fully functional for ArduPilot and
non-FTP-capable vehicles.

## Implementation Details

- Auto-detects FTP capability on PX4 vehicles via `MAV_PROTOCOL_CAPABILITY_FTP`
- Walks PX4 log directory tree via `FTPManager::listDirectory` (date subdirs → .ulg files)
- Downloads selected logs sequentially via `FTPManager::download` with progress/rate display
- All MAVFTP code guarded by `QGC_ENABLE_MAVFTP_LOG_DOWNLOAD` cmake option (ON by default)
  so the legacy protocol can be isolated or deprecated via build flag in the future
- Explicit `MavftpListState` enum tracks listing phase (root vs subdirectory)
- Proper state cleanup on vehicle switch and cancellation

## Testing

- [ ] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [x] PX4
- [ ] ArduPilot (legacy path unchanged)

## Checklist

- [ ] I have added tests that prove my feature works

## Known Limitations

- `eraseAll()` still uses legacy `LOG_ERASE` MAVLink message when MAVFTP is active
  (PX4 should still honor this, but FTP-based deletion could be a follow-up)
- Unit tests cover only the legacy path; MAVFTP test would require MockLink FTP
  directory listing support (follow-up)

---
By submitting this pull request, I confirm that my contribution is made under
the terms of the project's dual license (Apache 2.0 and GPL v3).

### DIFF
--- a/src/AnalyzeView/LogDownloadController.cc
+++ b/src/AnalyzeView/LogDownloadController.cc
@@ -1,6 +1,10 @@
 #include "LogDownloadController.h"
 #include "AppSettings.h"
 #include "LogEntry.h"
+
+#ifdef QGC_MAVFTP_LOG_DOWNLOAD
+#include "FTPManager.h"
+#endif
 #include "MAVLinkProtocol.h"
 #include "MultiVehicleManager.h"
 #include "ParameterManager.h"
@@ -11,6 +15,7 @@
 #include "Vehicle.h"
 
 #include <QtCore/QApplicationStatic>
+#include <QtCore/QDir>
 #include <QtCore/QTimer>
 
 QGC_LOGGING_CATEGORY(LogDownloadControllerLog, "AnalyzeView.LogDownloadController")
@@ -55,6 +60,34 @@ void LogDownloadController::_downloadToDirectory(const QString &dir)
     if (!_downloadPath.endsWith(QDir::separator())) {
         _downloadPath += QDir::separator();
     }
+
+#ifdef QGC_MAVFTP_LOG_DOWNLOAD
+    if (_useMavftp) {
+        // Build queue of selected entries that have FTP paths
+        _ftpDownloadQueue.clear();
+        const int numLogs = _logEntriesModel->count();
+        for (int i = 0; i < numLogs; i++) {
+            QGCLogEntry *const entry = _logEntriesModel->value<QGCLogEntry*>(i);
+            if (entry && entry->selected() && !entry->ftpPath().isEmpty()) {
+                entry->setStatus(tr("Waiting"));
+                _ftpDownloadQueue.enqueue(entry);
+            }
+        }
+
+        if (_ftpDownloadQueue.isEmpty()) {
+            qCWarning(LogDownloadControllerLog) << "MAVFTP: no selected logs have FTP paths for download";
+            return;
+        }
+
+        qCDebug(LogDownloadControllerLog) << "MAVFTP: queued" << _ftpDownloadQueue.size() << "logs for download to" << _downloadPath;
+        _setDownloading(true);
+
+        // Start the first download
+        QGCLogEntry *first = _ftpDownloadQueue.dequeue();
+        _downloadLogMavftp(first);
+        return;
+    }
+#endif
 
     QGCLogEntry *const log = _getNextSelected();
     if (log) {
@@ -134,6 +167,20 @@ void LogDownloadController::_setActiveVehicle(Vehicle *vehicle)
         _logEntriesModel->clearAndDeleteContents();
         (void) disconnect(_vehicle, &Vehicle::logEntry, this, &LogDownloadController::_logEntry);
         (void) disconnect(_vehicle, &Vehicle::logData,  this, &LogDownloadController::_logData);
+
+#ifdef QGC_MAVFTP_LOG_DOWNLOAD
+        // Disconnect MAVFTP signals and reset state
+        FTPManager *ftp = _vehicle->ftpManager();
+        (void) disconnect(ftp, &FTPManager::listDirectoryComplete, this, &LogDownloadController::_mavftpListDirComplete);
+        (void) disconnect(ftp, &FTPManager::downloadComplete, this, &LogDownloadController::_ftpDownloadComplete);
+        (void) disconnect(ftp, &FTPManager::commandProgress, this, &LogDownloadController::_ftpDownloadProgress);
+        _useMavftp = false;
+        _mavftpListState = MavftpIdle;
+        _mavftpDirsToList.clear();
+        _mavftpLogIdCounter = 0;
+        _ftpDownloadQueue.clear();
+        _ftpCurrentDownloadEntry = nullptr;
+#endif
     }
 
     _vehicle = vehicle;
@@ -427,6 +474,23 @@ bool LogDownloadController::_prepareLogDownload()
 void LogDownloadController::refresh()
 {
     _logEntriesModel->clearAndDeleteContents();
+
+#ifdef QGC_MAVFTP_LOG_DOWNLOAD
+    if (_vehicle
+        && _vehicle->px4Firmware()
+        && _vehicle->capabilitiesKnown()
+        && (_vehicle->capabilityBits() & MAV_PROTOCOL_CAPABILITY_FTP)) {
+        _useMavftp = true;
+        qCDebug(LogDownloadControllerLog) << "refresh: using MAVFTP (px4:" << _vehicle->px4Firmware()
+            << "capsKnown:" << _vehicle->capabilitiesKnown()
+            << "ftpCap:" << bool(_vehicle->capabilityBits() & MAV_PROTOCOL_CAPABILITY_FTP) << ")";
+        _requestLogListMavftp();
+        return;
+    }
+    _useMavftp = false;
+#endif
+
+    qCDebug(LogDownloadControllerLog) << "refresh: using legacy LOG_REQUEST protocol";
     _requestLogList(0, 0xffff);
 }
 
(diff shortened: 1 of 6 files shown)


src/AnalyzeView/LogDownloadController.cc
```diff
@@ -631,6 +733,255 @@ void LogDownloadController::_setListing(bool active)
     }
 }
 
+#ifdef QGC_MAVFTP_LOG_DOWNLOAD
+
+// --- MAVFTP log listing ---
+
+static constexpr const char *kPx4LogRoot = "/fs/microsd/log";
```
dakejahl, line 740: This needs to be something more generic that the FC can reinterpret as the root log dir
```suggestion
static constexpr const char *kPx4LogRoot = "@LOGS_DIR";
```

src/AnalyzeView/LogDownloadController.cc
```diff
@@ -492,6 +587,13 @@ void LogDownloadController::eraseAll()
         return;
     }
 
+#ifdef QGC_MAVFTP_LOG_DOWNLOAD
+    if (_useMavftp) {
+        // TODO: Implement MAVFTP-based log erasure (FTP file removal)
```
dakejahl, line 592: Maybe just continue to use the LOG_ERASE for mass erase. Individual erase is too slow and I don't see the use.

dakejahl (commented):
I need to review this more thoroughly. I tested it and it does work. Over USB I am seeing 350KB/s.

julianoes:
I would make it a separate tab with general file access rather than replacing the existing log file tab, as that would generally be good to have.

HTRamsey:
Is there any value in having that cmake option or just enable it always when PX4 is detected?

dakejahl:
> I would make it a separate tab with general file access rather than replacing the existing log file tab, as that would generally be good to have.

I did think about that. I agree another tab for general file access would be super nice. My thinking was that by replacing the Log Download page with FTP is we could eventually deprecate and remove the non-FTP impl of log download in PX4.

> Is there any value in having that cmake option or just enable it always when PX4 is detected?

Yeah probably not much value. And yes, this may also work for Ardupilot too and we should ask them if we should prefer it. 

The primary motivating factor for this PR is to avoid link spam during log download. The LOG_DATA doesn't have target sys/comp ID fields so during bulk download all endpoints on a mavlink-router system get flooded.

julianoes:
@dakejahl are you aware of the fact that we are just discussing this over here?
https://github.com/mavlink/mavlink-devguide/pull/668

dakejahl:
Thanks I was not aware. Then maybe @peterbarker can comment: would Ardupilot prefer FTP over LOG_DATA as the default mechanism for log download in QGC?

peterbarker:
> Thanks I was not aware. Then maybe @peterbarker can comment: would Ardupilot prefer FTP over LOG_DATA as the default mechanism for log download in QGC?

Transfer via the traditional mechanism is actually marginally faster on ArduPilot - 800kB/s or there-abouts vs 750kB/s on USB.

I don't think this really needs to be any form of "default".  Ask the vehicle what file it should look at to get a list of logs per #668.  In ArduPilot we may synthesise that listing entirely, but at the very least it will allow us to abstract out the actual path to the log files.  On stm32-based boards these are in APM/LOGS on every in-tree board.  But on Linux the FTP client gets free access to the root of the SD card and the logs actually appear in several different places, depending on the specific Linux board.  The contents of the directory are always the same, 'though there's `LASTLOG.txt` in there to gum up the words for automation.

I'm pretty sure the string `px4` shouldn't be appearing in the files you're modifying here; resource discovery should make that not an issue.  If the resource discovery fails then the existing interface can be used.

DonLakeFlyer:
> Transfer via the traditional mechanism is actually marginally faster on ArduPilot - 800kB/s or there-abouts vs 750kB/s on USB.

That would indicate something is wrong somewhere wouldn't it. The FTP packets have more room for data than the LOG ones.

peterbarker:
On Tue, 17 Feb 2026, Don Gagne wrote:

>       Transfer via the traditional mechanism is actually marginally faster on ArduPilot - 800kB/s or there-abouts vs 750kB/s on USB.
> 
> That would indicate something is wrong somewhere wouldn't it. The FTP packets have more room for data than the LOG ones.

I'm not sure what the limits on the whole burst-read in FTP thing are. 
But you can just ask the autopilot to send a log file using the existing 
protocol and you just get a stream of messages past that point, no acks 
required, so there's never a round-trip overhead for the initial fetch.

In ArduPilot we also make the traditional log transfer from the main 
thread (something I've been trying to change for a long time - I have a 
branch), so *that* will also give the old mechanism a bit of a leg-up vs 
FTP (which is on a thread)

Note that packing FTP packets is a little rather more expensive CPU-wise 
than the old messages, too.

In terms of resource allocation it would be far preferable to concentrate 
on making FTP faster, for sure!

hamishwillee:
A couple of things to note from https://github.com/mavlink/qgroundcontrol/pull/13996 and its discussion:
- We're trying to define `@MAV_LOG` as a virtual drive prefix so that the location of the files is standardized and abstracted from flight stacks. 
- Some flight stacks might store logs on file systems that don't support per-file removal. This means that deleting individual files will fail. However you can still implement the same behaviour as the current UI's LOG_ERASE by deleting a whole folder.
- If we can't agree a standard virtual drive, then we should add a log location to COMPONENT_METADATA such that a GCS can fetch the generic location of files. This might be on a server somewhere (suggestion courtesy of @peterbarker)

PS FWIW Burst mode should be lots faster.

peterbarker:
On Tue, 17 Feb 2026, Hamish Willee wrote:

>  *  If we can't agree a standard virtual drive, then we should add a log location to COMPONENT_METADATA such that a GCS can fetch the generic location of files. This might be on a server
>     somewhere (suggestion courtesy of @peterbarker)

I think the virtual drive stuff is probably going to work.

... also, it crossed my mind that it's not just "cloud" that means there 
might be better ways to get the vehicle than mavftp.  You can run a web 
server on ArduPilot and apparently it's *much* faster to get the logs over 
http than either of ftp or the old protocol; I'll be playing with that 
this-afternoon.  I think this simply shows that the COMPONENT_METADATA 
thing is *complicated* and shouldn't block the virtual filesystem 
suggestion here.

hamishwillee:
> ...  I think this simply shows that the COMPONENT_METADATA thing is *complicated* and shouldn't block the virtual filesystem suggestion here.

IMO COMPONENT_METADATA is simpler 
- because the device says where it stores its logs and a GCS reads that and uses whatever URL it is given - HTTP or MAVFTP or whatever. 
- If you already have COMPONENT_METADATA it is much easier because you don't need to change MAVFTP at all.

But I still want to use virtual drives such as `@MAVLINK_LOG` because 
- COMPONENT_METADATA has been around for ever and despite obvious benefits still not adopted by ArduPilot. 
- Virtual drives in MAVFTP likely to be easier to standardize on, and in no way block COMPONENT_METADATA based approach.

@dakejahl I'd be interested on your thoughts. Would PX4 be interested in providing an abstraction to a log virtual drive?

julianoes:
> @dakejahl I'd be interested on your thoughts. Would PX4 be interested in providing an abstraction to a log virtual drive?

I would think so.

peterbarker:
> > Transfer via the traditional mechanism is actually marginally faster on ArduPilot - 800kB/s or there-abouts vs 750kB/s on USB.
> 
> That would indicate something is wrong somewhere wouldn't it. The FTP packets have more room for data than the LOG ones.

A little more info.  In addition to the above speeds (CubeRedPrimary), I get about the same log-download rates over a 100Mbit TCP connection.  

OTOH, over http (using a http server written in Lua running on CubeRedPrimary):
```
$ wget http://172.16.3.68:8080/mnt/APM/LOGS/00000001.BIN
--2026-02-18 17:37:32--  http://172.16.3.68:8080/mnt/APM/LOGS/00000001.BIN
Connecting to 172.16.3.68:8080... connected.
HTTP request sent, awaiting response... 200 OK
Length: 38366640 (37M) [application/octet-stream]
Saving to: ‘00000001.BIN’

00000001.BIN        100%[===================>]  36.59M  4.18MB/s    in 8.8s    

2026-02-18 17:37:41 (4.15 MB/s) - ‘00000001.BIN’ saved [38366640/38366640]

$ 
```

I've varying reports on how far you can push that up towards the limits implied by the ethernet bitrate.

peterbarker:
tridge suggests that giving the client a selection of URLs to download the logs from would be a good idea.  e.g. http will likely be fastest but the client may not have a route.

dakejahl:
Virtual drive for logs makes sense. I like the idea of using a file to provide the log list and associated URLs. I'd be happy to help work on this both in PX4 and QGC. Can we put together a spec into a doc somewhere? 

This is my understanding
- GCS requests logs lists (via FTP, asking for file @LOG_LIST which is an alias for the actual location?)
- FC sends down the file which contains logs metadata: index, datetime, URL
- GCS FTPs from the URL (which could be onboard the FC or in the cloud somewhere)

It's that simple?

peterbarker:
On Wed, 25 Feb 2026, Jacob Dahl wrote:

> Virtual drive for logs makes sense. I like the idea of using a file to provide the log list and associated URLs. I'd be happy to help work on this both in PX4 and QGC. Can we put
> together a spec into a doc somewhere?
> 
> This is my understanding
>
>  *  GCS requests logs lists (via FTP, asking for file @LOG_LIST which is an alias for the actual location?)
>  *  FC sends down the file which contains logs metadata: index, datetime, URL
>  *  GCS FTPs from the URL (which could be onboard the FC or in the cloud somewhere)
> 
> It's that simple?

That all sounds about right - but I'd suggest a bog-standard directory 
listing of ***@***.***_LOGS is also a reasonable thing, and standard FTP 
download of ***@***.***_LOGS/log-20240224-1.BIN would be a thing.

Doc would be good - I'm assuming it would be JSON - and let's make sure to 
include a damned json version number in there because it won't be right 
the first time :-)

hamishwillee:
@dakejahl For the MAVFTP part, we're suggesting `@MAV_LOG` for the virtual directory name. I think we have broad agreement on this https://github.com/mavlink/mavlink-devguide/pull/668 PR - needs prototype if you're up for it?

This doesn't suggest anything about log lists or whatever, but there is no reason this could not be standardized.

Component metadata might further specify the log location for logs that might be exposed via HTTPS - of which this could be one option.

peterbarker:
On Thu, 26 Feb 2026, Hamish Willee wrote:

> @dakejahl For the MAVFTP part, we're suggesting @MAV_LOG for the virtual directory name. I think we have broad agreement on this mavlink/mavlink-devguide#668 PR - needs prototype if
> you're up for it?

For retrieval of logs via FTP, sure.

> This doesn't suggest anything about log lists or whatever, but there is no reason this could not be standardized.
> 
> Component metadata might further specify the log location for logs that might be exposed via HTTPS - of which this could be one option.

tridge separately suggested something to what @dakejahl was suggesting - 
providing a list of alternative fetch mechanisms for each log as it comes 
up.

dakejahl:
Closing in favor of https://github.com/mavlink/qgroundcontrol/pull/14290

PR #14013, opened 2026-02-21
[PX4] AM32 EEPROM settings support

dakejahl:
## Summary                                                
  - Add AM32 ESC EEPROM read/write settings UI via MAVLink [ESC_EEPROM](https://mavlink.io/en/messages/development.html#ESC_EEPROM) message
  - Fetch and cache AM32 JSON schema from [am32.ca](https://am32.ca/eeprom) with fallback to bundled schema
  - Implement per-ESC selection, bulk read/write, majority-match detection, and version-aware setting visibility
  - Refactor ESC status indicators from bitmask-based to per-ESC online status, with grid layout for >4 ESCs

This is effectively the [AM32 Configurator](https://am32.ca/configurator) built into QGC. This enables a wireless method to check and edit ESC settings.

The accompanying PX4 PR is here https://github.com/PX4/PX4-Autopilot/pull/26263

  ## Details
  New files:
  - `AM32Component` — PX4 auto-pilot plugin component for the AM32 settings page
  - `AM32SettingsComponent.qml` — Full settings UI with read/write controls, unsaved change indicators, and setting sliders
  - `AM32EepromFactGroupListModel` — MAVLink message handling for ESC EEPROM read/write, including write masks and majority matching across selected ESCs
  - `AM32EepromSchema` — Remote schema fetching with `CachedFileDownload`, JSON parsing, version overrides, and conversion factor support

  Modified:
  - `EscIndicator.qml` / `EscIndicatorPage.qml` — Per-ESC online status instead of shared bitmask; grid layout for motor count >4
  - `EscStatusFactGroupListModel.cc` — Fix index-to-bitmask mapping for 4-per-message groups
  - `Vehicle.cc/h` — Integrate AM32 EEPROM model

  ## Test plan
  - [x] Build succeeds with no warnings
  - [ ] Connect to vehicle with AM32 ESCs and verify schema downloads (check logs for "Fetching AM32 schema from")
  - [x] Read/write EEPROM settings through the AM32 settings page
  - [ ] Verify ESC indicator shows correct online/offline status for >4 ESCs
  
  ## Screenshots
  
  All settings match
image

One ESC beep volume is different
image

ESC indicator grid layout
image

### DIFF
--- a/src/AutoPilotPlugins/PX4/AM32Component.cc
+++ b/src/AutoPilotPlugins/PX4/AM32Component.cc
@@ -1,0 +1,57 @@
+/****************************************************************************
+ *
+ * (c) 2009-2024 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ *
+ * QGroundControl is licensed according to the terms in the file
+ * COPYING.md in the root of the source code directory.
+ *
+ ****************************************************************************/
+
+#include "AM32Component.h"
+#include "Vehicle.h"
+
+AM32Component::AM32Component(Vehicle* vehicle, AutoPilotPlugin* autopilot, QObject* parent)
+    : VehicleComponent(vehicle, autopilot, AutoPilotPlugin::UnknownVehicleComponent, parent)
+    , _name(tr("AM32 ESC"))
+{
+}
+
+QString AM32Component::name(void) const
+{
+    return _name;
+}
+
+QString AM32Component::description(void) const
+{
+    return tr("AM32 ESC Setup is used to configure AM32 ESC settings.");
+}
+
+QString AM32Component::iconResource(void) const
+{
+    return "/qmlimages/EscIndicator.svg";
+}
+
+bool AM32Component::requiresSetup(void) const
+{
+    return false;
+}
+
+bool AM32Component::setupComplete(void) const
+{
+    return true;
+}
+
+QStringList AM32Component::setupCompleteChangedTriggerList(void) const
+{
+    return QStringList();
+}
+
+QUrl AM32Component::setupSource(void) const
+{
+    return QUrl::fromUserInput("qrc:/qml/QGroundControl/AutoPilotPlugins/PX4/AM32Component.qml");
+}
+
+QUrl AM32Component::summaryQmlSource(void) const
+{
+    return QUrl::fromUserInput("qrc:/qml/QGroundControl/AutoPilotPlugins/PX4/AM32ComponentSummary.qml");
+}
--- a/src/AutoPilotPlugins/PX4/AM32Component.h
+++ b/src/AutoPilotPlugins/PX4/AM32Component.h
@@ -1,0 +1,36 @@
+/****************************************************************************
+ *
+ * (c) 2009-2024 QGROUNDCONTROL PROJECT <http://www.qgroundcontrol.org>
+ *
+ * QGroundControl is licensed according to the terms in the file
+ * COPYING.md in the root of the source code directory.
+ *
+ ****************************************************************************/
+
+#pragma once
+
+#include "VehicleComponent.h"
+
+class AM32Component : public VehicleComponent
+{
+    Q_OBJECT
+    
+public:
+    AM32Component(Vehicle* vehicle, AutoPilotPlugin* autopilot, QObject* parent = nullptr);
+    
+    // Overrides from VehicleComponent
+    QStringList setupCompleteChangedTriggerList(void) const override;
+    
+    // Overrides from VehicleComponent
+    QString name                    (void) const override;
+    QString description             (void) const override;
+    QString iconResource            (void) const override;
+    bool    requiresSetup           (void) const override;
+    bool    setupComplete           (void) const override;
+    QUrl    setupSource             (void) const override;
+    QUrl    summaryQmlSource        (void) const override;
+    bool    allowSetupWhileArmed    (void) const override { return true; }
+
+private:
+    const QString   _name;
+};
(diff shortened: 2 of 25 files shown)


dakejahl:
@DonLakeFlyer do you have any feedback on the design?

DonLakeFlyer:
I'll try to to get to this. I'm swamped with my own wildlife drone testing right now.

dakejahl:
> I'll try to to get to this. I'm swamped with my own wildlife drone testing right now.

Okay no rush, I just wasn't sure if you saw it. Let me know if I can help.

DonLakeFlyer:
First quick comment: With regards to the new `ESC_EEPROM` message which sets params. Wouldn't it be better if this where a MAV_CMD? This way you can get an Ack result as well as automatic retry support.

DonLakeFlyer:
For the places you are using sliders, what about using FactTextFieldSlider. Then you get both slider and type in support:
Screenshot 2026-03-03 at 9 11 15 AM

dakejahl:
> First quick comment: With regards to the new `ESC_EEPROM` message which sets params. Wouldn't it be better if this where a MAV_CMD? This way you can get an Ack result as well as automatic retry support.

There's a couple of reasons I went with the message "package" structure
- Atomic updates. The FC writes the entire settings struct to the ESC in one go. This is required by the ESC due to how the comm protocol works. When the FC writes settings it performs a command sequence to write the EEPROM, we need to know when to start/stop. It also allows use to show the current state of the ESC settings and the pending changes from the GCS.
- Simplicity. MAV_CMDs would spam the link and the 1:1 setting-to-command is the wrong paradigm. The FC responds with its own ESC_EEPROM message which serves as an ACK.

There's a lot more context in the original RFC https://github.com/mavlink/rfcs/pull/27

DonLakeFlyer:
> * The FC responds with its own ESC_EEPROM message which serves as an ACK.

I'm all good as long as there is some way to determine whether it worked or not.

DonLakeFlyer:
@dakejahl Can you squash to a reasonable set of commits. Rebase and then I'll submit it for another CoPilot review.

dakejahl:
I'm also going to run some Claude /review on this. Let's hold off on copilot since it can really thrash the PR discussion with noise.

DonLakeFlyer:
I guess the thing I wonder about with this is why it's valuable to increase the surface area of QGC for something which is already supported by an external tool. Why not go the same route as ArduPilot which has a mode to make the regular configurator? Also means you need to chase changes made to the original tool.

I always worry about biggish new things since I tend to get left holding the bag on them for years to come...

dakejahl:
> Why not go the same route as ArduPilot which has a mode to make the regular configurator?

AM32 configurator **requires** a physical USB connection. 
- This is impossible on our [ARK Jetson Carrier](https://arkelectron.com/product/ark-jetson-pab-carrier/) boards since the FC USB is not exposed. 
- Many builds bury the FC inside the airframe or otherwise block the USB port. 

> Also means you need to chase changes made to the original tool.

The eeprom data has a `version` field and the schema https://am32.ca/eeprom is also version aware. The result should be that this QGC tool will simply leave out new fields that it doesn't know about.  

> I always worry about biggish new things since I tend to get left holding the bag on them for years to come...

Totally understand that. I can always make a separate Qt tool for AM32 over MAVLink, but it's nice to have everything in one tool. The impact to other modules is low, so can be easily removed if it becomes a burden.

We get this complaint very often:
> "So I need to buy a separate FC and flash Betaflight just to update these ESCs???"

And PX4 doesn't support "ESC passthrough" (the USB protocol) so indeed they need to use an FC that runs betaflight. So they have to disassemble their airframe just configure settings or flash fw or they need an FC that supports both PX4 and Betaflight.

And the counter argument is 
> "users should know they need to configure/flash their ESCs before they build their drone!"

But a lot of people don't know this and end up frustrated. All of this work is coming from real world user complaints around setting up and updating AM32 firmware.

DonLakeFlyer:
Good reasons to add this then...

dakejahl:
> Good reasons to add this then...

Yup. I'll work with Claude today/tomorrow and try to get this as good as possible before you review so that you don't have to spend too much time on it :smiley:

PetervdPerk-NXP:
@dakejahl Is this working and what's blocking this?
I've got on my list to test the bidrectional dshot aspect with config functionality as well.

dakejahl:
@PetervdPerk-NXP it does indeed work! There is one issue which needs to be fixed, we need to create a whitelist for mismatched settings and add `directionReversed`, since we expect CW and CCW combinations. The red/green highlighting is intended to let the user know when critical settings such as Pole Count or KV are mismatched. I would also appreciate another set of eyes to review the code, as a large chunk of it was written with AI assistance.

DonLakeFlyer:
So what's the state of this?

dakejahl:
I wanted to wait until the AM32 PR is in https://github.com/am32-firmware/AM32/pull/295 and get the metadata hosted so that we're not relying on the built-in version of the schema. Just waiting on the maintainers over there to do something.

DonLakeFlyer:
Ok, this is going to miss next stable then. I'm wrapping up now.

HTRamsey:
Should we have a blocked label or something that shows a PR is waiting on something else first

DonLakeFlyer:
Update on this?

dakejahl:
No motion from AM32 maintainers. We've since forked AM32 into [ARK32](https://github.com/ARK-Electronics/ARK32) and will be maintaining that as our officially supported ESC firmware instead. I am considering adding mavlink support directly into the ark32-configurator, so this PR is likely moot

DonLakeFlyer:
Ok, closing for now then

PR #14045, opened 2026-02-27
Pre-compile QRegularExpression in BoardRegExpFallback_t

DieBorr:
QGCSerialPortInfo::getBoardInfo() was creating QRegularExpression objects inside a loop on every call, triggering PCRE2 JIT compilation on each iteration. Since LinkManager calls this function for every serial port every second from the main thread, this was blocking the Qt event loop, causing cyclical stutters in MAVLink data reception.

This behavior can be clearly observed on the Analyze page: telemetry values periodically freeze and then update abruptly.
I alredy checked if this was a mavproxy problem, it is not, as the time_boot value looks has a almost constant slope.

## Description
For some time I had been noticing that the UI felt laggy. When I started investigating, I realized it wasn't a rendering or UI issue. It was a stutter that occurred semi-periodically and affected all vehicle-related values simultaneously (telemetry, instruments, everything). After running some profiling and tracing, I found periodic blocks on the main thread. Eventually I traced the root cause back to LinkManager, specifically to QGCSerialPortInfo methods that were being called by it on every serial port scan cycle.

## Testing

- [x] Tested locally
- [ ] Tested with hardware (Please check)

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [x] ArduPilot

## Related Issues

---

### DIFF
--- a/src/Comms/QGCSerialPortInfo.cc
+++ b/src/Comms/QGCSerialPortInfo.cc
@@ -110,8 +110,15 @@ bool QGCSerialPortInfo::_loadJsonData()
             return false;
         }
 
+        const QRegularExpression regExp(fallbackObject[_jsonRegExpKey].toString(), QRegularExpression::CaseInsensitiveOption);
+        if (!regExp.isValid()) {
+            qCWarning(QGCSerialPortInfoLog) << "Invalid regular expression in board description fallback:"
+                                             << regExp.errorString()
+                                             << "pattern:" << fallbackObject[_jsonRegExpKey].toString();
+            return false;
+        }
         const BoardRegExpFallback_t boardFallback = {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
The CoPilot review comments make sense. They should spit out a qCrCritical to indicate an internal error.

DieBorr:
> The CoPilot review comments make sense. They should spit out a qCrCritical to indicate an internal error.

Okay, as it was not already implemented I decided to not do it in this commit, I will take care of the validation.

DieBorr:
@DonLakeFlyer how can I trigger Copilot to do another review? This PR is already completed, but I’m asking for future PRs.

DieBorr:
Okay, found how :)

DonLakeFlyer:
And then on the older review you can mark the review items as resolved. That way I know you took care of them.

PR #14046, opened 2026-02-27
Back out concept of raw value not being set

DonLakeFlyer:
* Was causing too many problems in random places which would only be identified if you stumbled across a dusty corner of the QGC qml ui and saw it failing.
* I've already gone through multiple iterations of changing it try to make it work without problems but I just discovered again that it wasn't working in all cases.
* The FactSystem and the way UI is written just isn't set up to handle this.
* Can't afford to play whack-a-mole with it any more. The UI impact on the change isn't worth the impact on having to manually test the entire QGC ui looking for screws-ups.

@rubenp02 FYI

### DIFF
--- a/src/FactSystem/Fact.cc
+++ b/src/FactSystem/Fact.cc
@@ -50,7 +50,6 @@ Fact::Fact(const QString& settingsGroup, FactMetaData *metaData, QObject *parent
             const QVariant defaultValue = metaData->rawDefaultValue();
             QMutexLocker<QRecursiveMutex> locker(&_rawValueMutex);
             _rawValue = defaultValue;
-            _rawValueIsNotSet = false;
         }
     }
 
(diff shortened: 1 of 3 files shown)


rubenp02:
Sorry, a bit OOTL on this. What are the functional implications of this? Placeholder hypens on indicators are no more and we're back to showing 0s, or is it something else?

DonLakeFlyer:
Yeah, that's it. Default rawValue is 0 if not set.

rubenp02:
> Yeah, that's it. Default rawValue is 0 if not set.

Okay. Certainly better than random stuff being broken. However I think having stuff show 0 when it's actually unset or unavailable is extremely bad for an application like this. Isn't agentic AI supposed to be really good at exactly the kind of work that would be to fix (or at least detect) all these bugs? I.e., very simple but high volume work across the codebase.

PR #14111, opened 2026-03-12
GStreamer: macOS auto-download install, CI verify-executable, and init tests

HTRamsey:
## Summary

- Add `GStreamer_AUTO_DOWNLOADED` macOS install path: dylibs to `Contents/Frameworks`, plugins/GIO modules to `Contents/lib`, rpath fixups for codesigning compatibility, and helper binary (`gst-plugin-scanner`, `gst-ptp-helper`) bundling
- Add `build-dir` input to `verify-executable` action and Python script to extract `GStreamer_ROOT_DIR` from `CMakeCache.txt` and set plugin env vars for dev-build boot tests (macOS/Windows)
- Add `VideoManagerInitTest` covering both QML-before-GStreamer and GStreamer-before-QML init orderings with idempotency checks
- Fix `GStreamerTest` ASan hang: restrict plugin scan under sanitizer builds to avoid `__bindtextdomain` stall in shadow memory

## Test plan

- [ ] macOS CI: verify dev-build boot test passes with `build-dir` set
- [ ] Windows CI: verify dev-build boot test passes with `build-dir` set
- [ ] Linux sanitizer CI: verify `GStreamerTest` no longer times out
- [ ] `VideoManagerInitTest` passes in both GStreamer-enabled and disabled builds
- [ ] macOS release build with auto-downloaded SDK: verify codesigning succeeds


DonLakeFlyer:
So I tried building this and it failed because I had a gstreamer 1.24 already installed via brew on my machine for other uses. Since this downloads to the build directory I wouldn't go looking for it in other places. Also wonder about downloading dependencies to the build dir. That means deleting build dir to get a clean build needs internet connection which isn't a good thing. Maybe put it in CPM_SOURCE_CACHE location or something like that for users who can't afford to need internet to do a clean build (like me in Africa).

DonLakeFlyer:
Or maybe you have it the right way. If you don't want to download during build. You preinstall with brew and then you are good to go...

DonLakeFlyer:
> You preinstall with brew and then you are good to go...

But that doesn't work:
```
[cmake] -- CPM: Adding package mavlink@0 (b1fb5a1a32c41c6e46fea70600d626a0b5a8edbe at /Users/don/repos/cpm_source_cache/qgroundcontrol/mavlink/fa55)
[cmake] -- Not enabling format target (clang-format not found)
[cmake] -- CPM: Adding package libevents@ (main at /Users/don/repos/cpm_source_cache/qgroundcontrol/libevents/1ea2)
[cmake] -- Checking for module 'gstreamer-1.0=1.28.1'
[cmake] -- Package 'glib-2.0', required by 'gstreamer-1.0', not found
[cmake] Package 'gobject-2.0', required by 'gstreamer-1.0', not found
[cmake] CMake Error at /opt/homebrew/share/cmake/Modules/FindPkgConfig.cmake:1093 (message):
[cmake] The following required packages were not found:
[cmake]
[cmake] - gstreamer-1.0=1.28.1
[cmake]
[cmake] Call Stack (most recent call first):
[cmake] /opt/homebrew/share/cmake/Modules/FindPkgConfig.cmake:1166 (_pkg_check_modules_internal)
[cmake] cmake/find-modules/FindQGCGStreamer.cmake:697 (pkg_check_modules)
[cmake] src/VideoManager/VideoReceiver/GStreamer/CMakeLists.txt:5 (find_package)
[cmake]
[cmake]
[cmake] -- Configuring incomplete, errors occurred!
```

DonLakeFlyer:
When I remove all the gstreamer versions on my Mac I can build. But when I get no window. QGC is running though since I can hear it connecting to SITL. And the terminal output is splatting out random chars. Very strange.

I tried installing the version from this pull artifact. Same problem, no window.

PR #14118, opened 2026-03-12
Gimbal: Fix rate control dropping stop command on simultaneous axis release

### LINKED ISSUE #13580: Gimbal rate controll send infinte rate control in one axes, after both axis were held on joystick

## Expected Behavior
Holding both the **Roll Rate** and **Pitch Rate** buttons on the gimbal should spin the gimbal diagonally.  
When both buttons are released, the gimbal should stop spinning.

## Current Behavior
When both axes are released simultaneously, the gimbal continues spinning on one axis.

This is due to the MAVLink commands being sent from QGC:

Example:
Holding both buttons:
```
MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
params3 = +40 (deg/s)
params4 = +40 (deg/s)
```
Letting go of buttons:
```
MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
params3 = 0 (deg/s)
params4 = +40 (deg/s)
```

Causing the gimbal to keep spining

## Steps to Reproduce
1. I used a PS4 joystick with gimbal control mapped to the arrow buttons.  
2. Run **QGroundControl v5 (stable)** inside an **Ubuntu 24 Docker** container.  
   - The issue was also reproduced on **Ubuntu 22** system with QGC v5 (stable).

### System Information

Ubuntu 20 running Ubuntu 24 docker with the QGC v5 stable, tested on native ubuntu 22 as well. 
Controlled via PS4 joystick

Adding photos from wireshark with caught messages from QGC

Image

Image

### Log Files and Screenshots

_No response_

DonLakeFlyer:
When both joystick gimbal axes are released at the same time, `gimbalPitchStop()` and `gimbalYawStop()` each call `sendRate()`, which uses `sendMavCommand(MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW)`. Vehicle's duplicate command detection drops the second call while the first is still in-flight, leaving one axis spinning indefinitely.

**Fix:** Switch `sendRate()` to use `_sendGimbalAttitudeRates()` which sends a raw `gimbal_manager_set_attitude` mavlink message, bypassing Vehicle command tracking. This is the same approach already used by `sendGimbalRate()`.

Only the joystick rate path is changed. All other gimbal commands (`sendPitchBodyYaw`, `sendPitchAbsoluteYaw`, `sendPitchYawFlags`, etc.) continue to use `sendMavCommand` with their existing error reporting.

Fixes #13580

### DIFF
--- a/src/Gimbal/GimbalController.cc
+++ b/src/Gimbal/GimbalController.cc
@@ -563,25 +563,10 @@ void GimbalController::sendRate()
         return;
     }
 
-    unsigned flags = GIMBAL_MANAGER_FLAGS_ROLL_LOCK | GIMBAL_MANAGER_FLAGS_PITCH_LOCK;
-
-    if (_activeGimbal->yawLock()) {
-        flags |= GIMBAL_MANAGER_FLAGS_YAW_LOCK;
-    }
-
-    _vehicle->sendMavCommand(
-        _activeGimbal->managerCompid()->rawValue().toUInt(),
-        MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW,
-        false,
-        NAN,
-        NAN,
-        _activeGimbal->pitchRate(),
-        _activeGimbal->yawRate(),
-        flags,
-        0,
-        _activeGimbal->deviceId()->rawValue().toUInt());
-
-    qCDebug(GimbalControllerLog) << "Gimbal rate sent!";
+    // We send raw mavlink instead of using MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW because
+    // when both pitch and yaw stop simultaneously, Vehicle's duplicate command detection
+    // drops the second sendMavCommand call, leaving one axis spinning indefinitely.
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
FYI: I haven't been able to test this myself yet. Waiting for a new joystick to show up...

@AdamBucko1 Can you test this?

@Davidsastresas Can you review and test as well?

Davidsastresas:
I wasn't aware of this issue, but I was able to reproduce it.

The change  you did here is good I think, but unfortunately I was able to reproduce the issue with your commit just the same.

I went up on the architecture of the whole thing, and I found https://github.com/mavlink/qgroundcontrol/pull/14118/changes/7c4763b4a4559f9c7e53fe46585b4b44329a3fd1. We were only processing the first button action release.

I think your change is good too, so I left it, rebased to master, and added the joystick fix.

Thanks for taking a look at it!

Davidsastresas:
Actually I realized the same happens on button press. If several buttons pushed at the same time we were returning after the first action executed. The last commit addresses that.

The QSet approach is to handle properly multi-button actions.

PR #14159, opened 2026-03-18
PlanView: Fix 'Cannot read property of null' warnings on vehicle connect

dakejahl:
## Summary
- Adds null guards to declarative QML bindings in `MissionItemEditor`, `MissionSettingsEditor`, `CameraSection`, and `FactTextFieldSlider`
- When a vehicle connects the mission tree model rebuilds, destroying backing C++ objects while QML delegates still briefly exist — causing cascading "Cannot read property X of null" warnings
- Completes the fix started in 19d3c042 which handled the *creation* path via `setSource()` but missed the *teardown* path

## Test plan
- [ ] Connect a vehicle (real or SITL) and verify no "Cannot read property X of null" warnings in console
- [ ] Open Plan View and verify mission item editors still function correctly
- [ ] Verify camera section controls (gimbal, photo interval, mode) work in mission settings
- [ ] Verify FactTextFieldSlider displays and operates correctly for gimbal pitch/yaw

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/FactSystem/FactControls/FactTextFieldSlider.qml
+++ b/src/FactSystem/FactControls/FactTextFieldSlider.qml
@@ -27,7 +27,7 @@ Rectangle {
     radius:         ScreenTools.defaultBorderRadius
 
     property bool _loadComplete:            false
-    property bool _showSlider:              fact.userMin !== undefined && fact.userMax !== undefined
+    property bool _showSlider:              fact && fact.userMin !== undefined && fact.userMax !== undefined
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
This fix is wrong. The fact that these are popping out indicates a bug in the PlanTreeView generation (which is all new). You should never get to these controls with nulls. This fixes the symptom, not the actual problem. I've been working on this, just haven't found all the problems yet. I have an Issue somewhere to track it.

DonLakeFlyer:
Related to #14156. The bug in the new tree view object list model lifecycle management code.

PR #14168, opened 2026-03-19
APM: add Servo Outputs setup page with live output bars

### LINKED ISSUE #10199: Servo Output Functions Page (w/ Control?)
Would be nice to have a page where you can set the servo outputs to different functions. That way when I am setting up my planes in ArduPlane I don’t have to mess with the parameter list for basic setup. Don’t know if this should be its own page - or merged into the Motors page and renamed to “Actuators” or something like that.

AhmWael:
## Description
Adds a new Servo Outputs setup page for ArduPilot (APM) vehicles.

This page provides both real-time monitoring and configuration of servo outputs, including:

- Live visualization of PWM outputs using dynamic bars
- Ability to change the function assigned to each servo channel
- Option to reverse servo direction
- Editing of min, max, and trim PWM values

This significantly improves the user experience by consolidating servo configuration and feedback into a single, intuitive interface, making setup and debugging easier.

## Type of Change
- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing
- [x] Tested locally
- [ ] Tested with hardware

### Platforms Tested
- [ ] Linux
- [x] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested
- [ ] PX4
- [x] ArduPilot

## Screenshots
image

## Checklist

## Related Issues
Closes #10199

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -586,24 +586,31 @@ void Vehicle::_mavlinkMessageReceived(LinkInterface* link, mavlink_message_t mes
         mavlink_servo_output_raw_t servoOutputRaw;
         mavlink_msg_servo_output_raw_decode(&message, &servoOutputRaw);
 
-        const int baseIndex = static_cast<int>(servoOutputRaw.port) * 4;
-        if (baseIndex >= 0 && baseIndex < _servoOutputRawValues.size()) {
-            const uint16_t rawValues[4] = {
-                servoOutputRaw.servo1_raw,
-                servoOutputRaw.servo2_raw,
-                servoOutputRaw.servo3_raw,
-                servoOutputRaw.servo4_raw
-            };
+        // ArduPilot commonly publishes servo1_raw..servo16_raw in a single packet (port may remain 0).
+        const uint16_t rawValues[16] = {
+            servoOutputRaw.servo1_raw,
+            servoOutputRaw.servo2_raw,
+            servoOutputRaw.servo3_raw,
+            servoOutputRaw.servo4_raw,
+            servoOutputRaw.servo5_raw,
+            servoOutputRaw.servo6_raw,
+            servoOutputRaw.servo7_raw,
+            servoOutputRaw.servo8_raw,
+            servoOutputRaw.servo9_raw,
+            servoOutputRaw.servo10_raw,
+            servoOutputRaw.servo11_raw,
+            servoOutputRaw.servo12_raw,
+            servoOutputRaw.servo13_raw,
+            servoOutputRaw.servo14_raw,
+            servoOutputRaw.servo15_raw,
+            servoOutputRaw.servo16_raw
+        };
 
-            for (int i = 0; i < 4; i++) {
-                const int servoIndex = baseIndex + i;
-                if (servoIndex >= 0 && servoIndex < _servoOutputRawValues.size() && rawValues[i] != UINT16_MAX) {
-                    _servoOutputRawValues[servoIndex] = static_cast<int>(rawValues[i]);
-                }
-            }
-
-            emit servoOutputsChanged(_servoOutputRawValues);
+        for (int servoIndex = 0; servoIndex < _servoOutputRawValues.size() && servoIndex < 16; servoIndex++) {
+            _servoOutputRawValues[servoIndex] = (rawValues[servoIndex] == UINT16_MAX) ? -1 : static_cast<int>(rawValues[servoIndex]);
         }
+
+        emit servoOutputsChanged(_servoOutputRawValues);
     }
         break;
     case MAVLINK_MSG_ID_BATTERY_STATUS:


DonLakeFlyer:
Fantastic new support. Thanks. 

One question: Is this servo support global to all vehicle types in ArduPilot?

DonLakeFlyer:
Also if you could comment on other high level stuff like this for ArduPilot which is missing that would be helpful. I've been putting in a bunch of effort to make this next stable better with ArduPilot. But I tend to fly PX4 so I don't know what should be added and what is more edge case from how much it gets used.

AhmWael:
> Fantastic new support. Thanks.
> 
> One question: Is this servo support global to all vehicle types in ArduPilot?

Thanks for your comment!
It should work for all ArduPilot vehicles that expose SERVOx_* parameters.

Also, I just switched to QGC so I’m still exploring the full ArduPilot workflow. I’ll keep an eye on missing or high-impact features and share feedback as I come across them.

DonLakeFlyer:
This is good to go. Huge thanks for help with this.

DonLakeFlyer:
I have a question on the typical usage pattern for servo setup. Do they tend to be in a row, say you need to configure funtions on servo 3-7. Or sparse, you typically set up say 5, 10, 14.

AhmWael:
Typically for planes, servos are assigned in a sequential order, for example: Aileron, Elevator, Throttle, Rudder, etc., usually mapped to adjacent output channels.

Sparse assignments can also happen, but they are less common and usually due to specific hardware layouts, existing wiring, or incremental configuration changes.

DonLakeFlyer:
Ok, thanks

DonLakeFlyer:
@AhmWael What do you think about using a FactTextFieldSlider control for PWM futzing. That's a more common UI control for value adjustment. I think it depends on whether it more common to tweak a PWM up/down by just 1. If so, the plus/minus works better. If you usually tweak PWM by more than 1 increment then I think the slider might be better. Also only show adjustment controls for rows which are enabled:

Screenshot 2026-03-21 at 12 32 22 PM

AhmWael:
@DonLakeFlyer From my experience, PWM adjustments, especially for trim values, are usually done in very small increments, so the +/- buttons make fine tuning easier and more controlled.

Also, with press-and-hold behavior, the values can be adjusted more quickly when needed, so they still allow for reasonably fast changes without losing precision.

I do agree on only showing adjustment controls for enabled rows, that would make the UI cleaner and avoid confusion. I’ll update that.

DonLakeFlyer:
> I do agree on only showing adjustment controls for enabled rows, that would make the UI cleaner and avoid confusion. I’ll update that.

No need. I already have the change locally.

> From my experience, PWM adjustments, especially for trim values, are usually done in very small increments, so the +/- buttons make fine tuning easier and more controlled.

Ok, exactly what I was wondering about. So the +/- is better.

BOLYFLY:
Great work, it's really impressive. Can't wait to try it out.

AhmWael:
> 
> No need. I already have the change locally.
> 

Ok, If you have the changes locally for hiding controls for disabled servos, you can push them from your end.

BOLYFLY:
**_I am not a developer, so apologies if these modifications are not perfectly aligned with your coding standards._**

Problem: The current SERVO_OUTPUT_RAW decoding in Vehicle.cc is limited to 4 or 8 channels. This ignores ArduPilot extensions (16-channel single packets) and uses an incorrect baseIndex multiplier (port * 4), which causes data overlap or missing channels for users with advanced flight controllers like the Cube Orange.

Proposed Changes in Vehicle.cc:

Extended Decoding: Updated the mavlink_servo_output_raw_t decoding to read all 16 available fields (servo1_raw through servo16_raw) provided by the ArduPilot MAVLink dialect.

Corrected Indexing: Updated the baseIndex logic to use a multiplier of 16 (instead of 4). This ensures that if a system uses the multi-port standard (Port 0: 1-8, Port 1: 9-16), the data is stored in the correct vector indices without overwriting.

Vector Resizing: Added a safety check to ensure _servoOutputRawValues is resized to 16 elements upon receiving the first message to prevent data loss.

Updated Code Block in vehicle.cc:
```
case MAVLINK_MSG_ID_SERVO_OUTPUT_RAW:
{
    mavlink_servo_output_raw_t servoOutputRaw;
    mavlink_msg_servo_output_raw_decode(&message, &servoOutputRaw);

    // Ensure the vector supports 16 channels
    if (_servoOutputRawValues.size() < 16) {
        _servoOutputRawValues.fill(-1, 16);
    }

    // Port multiplier set to 16 to prevent index overlap
    const int baseIndex = static_cast<int>(servoOutputRaw.port) * 16;

    if (baseIndex >= 0 && baseIndex < 16) {
        const uint16_t rawValues[16] = {
            servoOutputRaw.servo1_raw,  servoOutputRaw.servo2_raw,  servoOutputRaw.servo3_raw,  servoOutputRaw.servo4_raw,
            servoOutputRaw.servo5_raw,  servoOutputRaw.servo6_raw,  servoOutputRaw.servo7_raw,  servoOutputRaw.servo8_raw,
            servoOutputRaw.servo9_raw,  servoOutputRaw.servo10_raw, servoOutputRaw.servo11_raw, servoOutputRaw.servo12_raw,
            servoOutputRaw.servo13_raw, servoOutputRaw.servo14_raw, servoOutputRaw.servo15_raw, servoOutputRaw.servo16_raw
        };

        for (int i = 0; i < 16; i++) {
            const int servoIndex = baseIndex + i;
            if (servoIndex < _servoOutputRawValues.size() && rawValues[i] != UINT16_MAX) {
                _servoOutputRawValues[servoIndex] = static_cast<int>(rawValues[i]);
            }
        }
        emit servoOutputsChanged(_servoOutputRawValues);
    }
}
break;
```

Impact: Enables full monitoring of AUX/MAIN outputs (9-16) for complex airframes (VTOL, Octocopters, etc.), which was previously broken or truncated in the UI.

AhmWael:
Thanks for the detailed suggestion, really appreciate it!

From what I observed while testing with ArduPilot SITL, it already sends servo1_raw through servo16_raw in a single SERVO_OUTPUT_RAW message (typically with port = 0), which is why my current implementation reads all 16 values directly without using the port offset.

So in practice this already handles AUX outputs (9–16) correctly for ArduPilot.

That said, your point about handling multi-port setups is valid for a more generic MAVLink implementation. If needed, we could extend this later to use port * 8 indexing to fully align with the MAVLink spec and support systems that split outputs across multiple messages.

Let me know what you think!

BOLYFLY:
To be honest, the number of ports seemed more important to “Claude”...  For me, using 16 or 0 for baseIndex didn’t change anything with the Cube orange with Arduplane.

AhmWael:
Got it, that makes sense 👍

On ArduPilot (including Cube Orange / ArduPlane), "SERVO_OUTPUT_RAW" typically comes as a single message with "port = 0" and all 16 channels populated, so changing the baseIndex doesn’t really have any visible effect in that setup.

I also checked the ArduPilot implementation in the GCS code, and it explicitly sends "servo1_raw" through "servo16_raw" in a single message for "port = 0", which matches what I’m seeing in SITL.

The difference would only show up in cases where outputs are split across multiple ports (e.g. systems using >16 channels or implementations strictly following the MAVLink spec with 8 channels per port). In that case, indexing based on "port" would matter.

For now, since ArduPilot sends all 16 values in one message, the current implementation already covers the common case correctly. We can always extend it later if we want to fully support multi-port decoding.

BOLYFLY:
I'm currently learning how to customize QGC. My goal was to create a simple ON (1000µs) / OFF (2000µs) button in the UI to control a strobe on my drone.

I wasn't sure about the "official" way to do it, but I eventually succeeded by modifying Vehicle.h to expose the servo PWM values directly to QML using a Q_PROPERTY:


```
Q_PROPERTY(QVariantList servoOutputRawValues READ servoOutputRawValues NOTIFY servoOutputsChanged)

// Getter to convert the internal QVector to a QML-friendly list
QVariantList servoOutputRawValues() { 
    QVariantList list;
    for (int value : _servoOutputRawValues) list << value; 
    return list; 
}
```

Now, in my FlyViewCustomLayer.qml, I have direct access to the real-time state of my outputs:

`activeVehicle.servoOutputRawValues[9]` servo 10 output en us

Happy to hear any feedback or if there's a better way to achieve this!

PR #14169, opened 2026-03-19
Add MAVLink 2 message signing support

### LINKED ISSUE #14160: PX4 Mavlink Signing Issues
PX4 added support https://github.com/PX4/PX4-Autopilot/pull/25284

Notes from @hamishwillee testing:

Did some testing. It seems to work but is suboptimal. 

The UI allows you to set a password and send it to the device in the https://mavlink.io/en/messages/common.html#SETUP_SIGNING message. This appears to work.

There are a few problems. 
* If you're setting up signing, it also needs to be enabled. On PX4 that requires setting a parameter. It would be better if the options for setting a password and enabling signing were co-located.
* When you set a password, my assumption is that QGC stores this password associated with the particular vehicle. That way if you exit and return it knows what signature to use. What isn't clear here is whether you HAVE to send the password to set it. If you do this means that every GCS that you want to have on the network must set the password, and must set the same password. You might instead be safer allowing the UI to save the password to use without setting it.
* It isn't clear if signing is working/enabled. I think this should probably be inferred and clearly published based on the packets themselves - if they have a signature.

Solving 1 is annoying because it is flight-stack specific. We should perhaps define a standard way to enable the feature. If it were up to me this would be a component metadata definition that points at the parameter that controls the feature (like the camera defininitions) but ArduPIlot doesn't support that yet, so open to suggestions.

DonLakeFlyer:
Fixes #14160

## Summary

Implements MAVLink 2 message signing support with named key management, per-vehicle auto-detection, and a toolbar indicator.

## Key Features

- **Named signing key storage** - Passphrases are SHA-256 hashed to 32-byte keys, stored via QSettings. Custom builds can seed default keys via QSettings.
- **Auto-detection** - When QGC receives signed packets from a vehicle, it automatically tries each stored key. On match, signing is configured on the link with no manual action needed.
- **Enable/Disable per vehicle** - Send SETUP_SIGNING to enable a key on a vehicle, or send an all-zero key to disable. Messages are sent twice for reliability.
- **Toolbar indicator** - Lock icon (green when signed) in the vehicle toolbar with full key management in expanded view.
- **Shared UI** - SigningKeyManager.qml used by both toolbar indicator and Telemetry settings.
- **Active key tracking** - Vehicle tracks which key is active with proper signal emission for QML bindings.
- **Key deletion protection** - Cannot delete a key in use by any connected vehicle.

## Tests

- SigningTest (5 unit tests): initSigning, checkSigningLinkId, createSetupSigning, verifySignature, tryDetectKey
- MockLinkSigningTest (3 integration tests): sendSetupSigning, sendDisableSigning, signingKeysAddRemove with QSignalSpy

## Documentation

- New MAVLink Signing section in Fly View toolbar docs
- New MAVLink 2 Signing section in Telemetry settings docs
- Renamed mavlink.md to telemetry.md to match current UI naming

Settings:
Screenshot 2026-03-19 at 2 35 16 PM

Toolbar:
Screenshot 2026-03-19 at 2 35 37 PM


DonLakeFlyer:
@hamishwillee FYI. I believe this covers your complaints.

hamishwillee:
> @hamishwillee FYI. I believe this covers your complaints.

@DonLakeFlyer Very nice.

PR #14178, opened 2026-03-22
AudioOutput: Add volume slider & stop currently playing audio on mute

jnomikos:
## Description
1. Added the ability to adjust volume from 0-100% in General Settings
2. Added "Test" button next to new slider which does an audio test
3. Fix minor annoyance where if you mute audio output while sound is playing / queued, you must wait until all the sound finishes playing before the mute takes effect.

Sometimes I've found the audio is way too loud compared to the rest of my system and so I reflexively mute it. The volume slider fixes that problem.

Note: I made the slider only visible if audio output isn't muted.

## Type of Change

- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [x] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [ ] ArduPilot

## Screenshots

image
image
image

## Checklist

## Related Issues

---


DonLakeFlyer:
I would use FactTextFieldSlider instead which is a simpler version. Since you are just adjusting audio output, not need to such detail.

jnomikos:
> I would use FactTextFieldSlider instead which is a simpler version. Since you are just adjusting audio output, not need to such detail.

Good point, I switched to FactTextFieldSlider as suggested.

I also simplified things further by removing the audioMuted fact and instead basing muted state on the audioVolume fact. Since muting was effectively just setting volume, this avoids duplication and I make use of the enableCheckbox on the FactTextFieldSlider by doing so.

image
image

DonLakeFlyer:
Thanks John

PR #14251, opened 2026-03-31
fix(signing): handle SETUP_SIGNING rejection to prevent communication loss

### LINKED ISSUE #14250: fix(signing): SETUP_SIGNING rejection causes communication loss
When QGC sends `SETUP_SIGNING` and the vehicle rejects it, QGC enables local signing anyway. Since the vehicle never accepted the key, both sides talk past each other and communication is lost with no recovery path except restarting QGC.

### How to reproduce

1. Connect QGC to PX4 (SITL or hardware)
2. Arm the vehicle
3. Go to Settings > Telemetry > Signing Key, enable a key
4. PX4 rejects: `MAVLink signing: rejected while armed`
5. QGC shows the rejection banner but has already enabled local signing
6. Communication is lost

The same problem occurs with `sendDisableSigning()` if PX4 rejects the disable.

### Root cause

In `Vehicle::sendSetupSigning()`, QGC calls `MAVLinkSigning::initSigning()` (enables local signing) **before** sending the `SETUP_SIGNING` message. There is no confirmation step. If the vehicle rejects, QGC is stuck signing outgoing messages with a key the vehicle doesn't have.

Additionally, `insecureConnectionAcceptUnsignedCallback` drops all unsigned incoming messages except `RADIO_STATUS`. This means the vehicle's unsigned STATUSTEXT rejection message ("rejected while armed") is silently dropped by QGC's signing layer and never reaches the user.

The result is a dead-lock: QGC signs outgoing (vehicle can't verify), QGC drops unsigned incoming (vehicle isn't signing). Neither side can hear the other.

### Suggested fix

Send `SETUP_SIGNING` first, then wait for confirmation before changing local signing state:

1. Send `SETUP_SIGNING` (still unsigned for enable, still signed for disable)
2. Start a confirmation timer (~3s)
3. If a heartbeat arrives (vehicle is still talking to us), confirm and enable/disable local signing
4. If a STATUSTEXT containing "rejected" arrives, cancel and notify the user
5. On timeout, roll back and notify

This also applies to `sendDisableSigning()`: don't clear local signing until confirmation.

### Observed during

Testing PX4 PR [PX4-Autopilot#26894](https://github.com/PX4/PX4-Autopilot/pull/26894) which adds an armed guard that rejects `SETUP_SIGNING` while the vehicle is armed.

mrpollo:
## Summary

- Defers local signing activation until the vehicle confirms it accepted the SETUP_SIGNING message
- Sends SETUP_SIGNING unsigned first, then waits up to 3 seconds for a heartbeat (confirmation) or STATUSTEXT (rejection)
- On timeout or rejection: cancels the pending signing change and notifies the user via app message
- Same confirmation pattern applied to disable signing: keeps local signing active until vehicle confirms the disable
- Previously, QGC enabled local signing before sending SETUP_SIGNING, so if the vehicle rejected the key, QGC would be stuck signing with a key the vehicle does not have, causing total communication loss

## Test plan

- [ ] Enable signing with a key the vehicle accepts: verify QGC enables local signing after heartbeat is received
- [ ] Enable signing with a key the vehicle rejects: verify QGC shows rejection message and does not enable local signing
- [ ] Enable signing when vehicle is unreachable (no heartbeat): verify 3-second timeout fires and user is notified
- [ ] Disable signing on a vehicle that accepts: verify local signing is disabled after heartbeat
- [ ] Disable signing on a vehicle that rejects: verify local signing remains active (rollback)
- [ ] Verify normal heartbeat processing is unaffected when no signing operation is pending

Fixes #14250

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -142,6 +142,11 @@ Vehicle::Vehicle(LinkInterface*             link,
     _prearmErrorTimer.setInterval(_prearmErrorTimeoutMSecs);
     _prearmErrorTimer.setSingleShot(true);
 
+    // Signing confirmation timer
+    _signingConfirmationTimer.setSingleShot(true);
+    _signingConfirmationTimer.setInterval(kSigningConfirmationTimeoutMs);
+    connect(&_signingConfirmationTimer, &QTimer::timeout, this, &Vehicle::_signingConfirmationTimeout);
+
     // Send MAV_CMD ack timer
     _mavCommandResponseCheckTimer.setSingleShot(false);
     _mavCommandResponseCheckTimer.setInterval(_mavCommandResponseCheckTimeoutMSecs());
@@ -1431,6 +1436,11 @@ void Vehicle::_handleHeartbeat(mavlink_message_t& message)
         if (previousFlightMode != flightMode()) {
             emit flightModeChanged(flightMode());
         }
+    }
+
+    // A heartbeat from the vehicle while signing setup is pending means the vehicle accepted
+    if (_isSigningPending()) {
+        _confirmSigningSetup();
     }
 }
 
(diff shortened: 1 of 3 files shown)


HTRamsey:
Replaced by #14268

PR #14261, opened 2026-04-03
Add SITL integration tests with PX4 SIH container

mrpollo:
Adds a new SITL integration test framework that runs QGC against a real PX4 SITL instance via Docker, validating MAVLink protocol behavior that MockLink cannot exercise: real UDP packet handling, parameter sync with retries over lossy transport, mission protocol round-trips with retransmission, standard modes negotiation via `AVAILABLE_MODES`, and command ACK semantics including `TEMPORARILY_REJECTED`.

Tests are gated behind `-DQGC_SITL_TESTS=ON` (off by default) and require Docker with the [`px4io/px4-sitl-sih`](https://hub.docker.com/r/px4io/px4-sitl-sih) container image. Each test method gets a fresh container started in `init()` and stopped in `cleanup()` for complete isolation. The image is pinned by digest in `.github/px4-sitl-digest.txt`.

The test framework splits into two layers:

- **MAVLink protocol tests** (`test/SITL/MAVLink/`) validate QGC's protocol implementation against any MAVLink-speaking autopilot. PX4 SITL is the concrete vehicle, but the assertions are protocol-level: heartbeat detection and communication loss, parameter sync completeness and modification round-trip, mission upload/download, command ACK handling, and `AVAILABLE_MODES` population. These would work against any conformant autopilot.

- **PX4-specific tests** (`test/SITL/PX4/`) validate PX4 firmware plugin behavior: full flight lifecycle (arm → takeoff → hold → land → auto-disarm), mode transition display fidelity, and firmware version identification from real `AUTOPILOT_VERSION`.

All 7 test classes (13 test methods) pass locally on macOS with Docker Desktop in ~124 seconds total. The lifecycle test exercises real SIH physics — the simulated quadcopter actually takes off to 10m, holds for 5 seconds, lands, and auto-disarms.

Also includes a fix for a null `QQmlEngine` crash in `ActuatorComponent` that surfaces when connecting to real PX4 SITL in test mode (MockLink never hits this path because it doesn't provide actuator metadata).

CI integration adds a non-blocking `px4-sitl-test` job to `linux.yml` with `continue-on-error: true` while the suite stabilizes.


HTRamsey:
Love it. This is something I had been wanting to do for a while. I'll check it out when I get back in town this weekend

mrpollo:
I'm going to try and get these to run faster than realtime to speed things up, hopefully it works!

DonLakeFlyer:
Keeping this hanging around for now as an example of how to do this

DonLakeFlyer:
Closing for now. We'll pick this back up after 5.1 goes out the door...

PR #14262, opened 2026-04-03
RemoteID: restore operator ID validation and sanitizing

### LINKED ISSUE #14204: Remote ID Operator ID not validated
The Remote ID Operator ID atleast [used to be validated](https://github.com/mavlink/qgroundcontrol/pull/10970). This is [currently broken](https://github.com/mavlink/qgroundcontrol/issues/13893) in the latest release. There was a fix [#13817](https://github.com/mavlink/qgroundcontrol/pull/13817) but it has not been released yet. 

I compiled from master and I can now successfully enter an operator id. However it now just uses whatever user gives as an input. I can give `DEADBEEFDEADBEEF` and it is happily accepted. Seems there is no validation anymore.

Image

```
$ tshark -q -l -i bluetooth-monitor -Y \
  "btcommon.eir_ad.entry.uuid_16 == 0xfffa" -T fields \
  -e btcommon.eir_ad.entry.service_data | rid

Capturing on 'bluetooth-monitor'
{"counter": 64, "protocol_version": 2, "message_type": 5, "id_type": 0, "operator_id": "DEADBEEFDEADBEEF"}
{"counter": 75, "protocol_version": 2, "message_type": 5, "id_type": 0, "operator_id": "DEADBEEFDEADBEEF"}
{"counter": 85, "protocol_version": 2, "message_type": 5, "id_type": 0, "operator_id": "DEADBEEFDEADBEEF"}
{"counter": 96, "protocol_version": 2, "message_type": 5, "id_type": 0, "operator_id": "DEADBEEFDEADBEEF"}
```

Also if user gives the full Operator ID with the secret part (xyz in the example) it is not removed but instead broadcasted to the world. The secret part is supposed not to be stored nor broadcasted. It is used only for validation.

Image

The `FIN87astrdge12k8-xyz` below is not real, it is an example operator id.

```
$ tshark -q -l -i bluetooth-monitor -Y \
  "btcommon.eir_ad.entry.uuid_16 == 0xfffa" -T fields \
  -e btcommon.eir_ad.entry.service_data | rid

Capturing on 'bluetooth-monitor'
{"counter": 234, "protocol_version": 2, "message_type": 5, "id_type": 0, "operator_id": "FIN87astrdge12k8-xyz"}
{"counter": 247, "protocol_version": 2, "message_type": 5, "id_type": 0, "operator_id": "FIN87astrdge12k8-xyz"}
{"counter": 255, "protocol_version": 2, "message_type": 5, "id_type": 0, "operator_id": "FIN87astrdge12k8-xyz"}
{"counter": 10, "protocol_version": 2, "message_type": 5, "id_type": 0, "operator_id": "FIN87astrdge12k8-xyz"}
{"counter": 21, "protocol_version": 2, "message_type": 0, "id_type": 2, "ua_type": 2, "uas_id": "ABCDEFGHIJKLMN0123"}
{"counter": 33, "protocol_version": 2, "message_type": 5, "id_type": 0, "operator_id": "FIN87astrdge12k8-xyz"}
{"counter": 42, "protocol_version": 2, "message_type": 5, "id_type": 0, "operator_id": "FIN87astrdge12k8-xyz"}
```

Junior00619:
Fixes #14204.

## Summary
- restore Operator ID validation inside `RemoteIDManager` so the generated settings UI no longer bypasses the EU checksum path
- sanitize validated EU Operator IDs back down to the public 16-character form before they are stored or reused
- add `RemoteIDManagerTest` coverage for valid EU IDs, invalid EU IDs, and switching from FAA to EU after entering a validated full ID

## Verification
- configured QGroundControl in the repo's Ubuntu/Qt Docker build environment with `-DQGC_BUILD_TESTING=ON`
- generated MAVLink headers and `MAVLinkEnums.h` in the build tree
- compiled the touched translation units directly from `compile_commands.json` inside the container:
  - `src/Vehicle/RemoteIDManager.cc`
  - `test/Vehicle/RemoteIDManagerTest.cc`

I did not finish a full `QGroundControl --unittest:RemoteIDManagerTest` run locally because the complete application build is much larger than the changed scope here, so the remaining end-to-end validation should come from CI.


DonLakeFlyer:
@Davidsastresas Can you look at this and verify this is a correct thing to do?

Davidsastresas:
Thanks for tackling this, wiring validation to the Fact signals is the right approach, since the generated settings UI writes the `operatorID` Fact directly and never went through the old `checkOperatorID()`/`setOperatorID()` path. The reentrancy guard around the sanitizing write-back looks correct, and `_setOperatorIDGood` only emitting on change is a nice touch.

A couple of things before merge:

**1. The PR description doesn't match the tests that shipped.** The summary lists coverage for "switching from FAA to EU after entering a validated full ID," but that case isn't in the diff, there are only `_validEUOperatorIDIsSanitized` and `_invalidEUOperatorIDClearsTrustedState`, and neither switches region. Looks like `narrow validation coverage to issue path` dropped it. Since validation is region-independent (a valid full ID entered under FAA sets `operatorIDValid=true`, then switching to EU strips it), that transition is probably the most valuable case to keep, could we restore it, or at least update the description?

**2. `checkOperatorID()` and `setOperatorID()` are now dead code.** They're `Q_INVOKABLE` but have no callers anywhere in the repo after this change. `checkOperatorID()` now duplicates `_handleOperatorIDChanged`'s validation, and `setOperatorID()` is just `_refreshOperatorIDState()`. Could we delete them (or have them delegate) rather than leave two parallel validation paths?

Minor:
- `length() > 16` is redundant with `_isEUOperatorIDValid`, which already requires exactly 19/20 chars. A named constant for the 16-char public length would also read better than the repeated literal.
- `operatorIDType >= 0` is always true (`uint8`, default 0). Pre-existing, just noting it's carried forward.

DonLakeFlyer:
This is still out of whack. This may validate the id internally and not send crap, but it doesn't tell the user through the ui that the id is bad like it used to. The other part of this got screwed up with the change the json settings page definitions. I'll take a look at taking this and also in combination fixing up the UI in a replacement pull that is a superset of this.

DonLakeFlyer:
@Davidsastresas Questions on stored settings: The current/original code allows storage of invalid operator ids into settings. Is there some specific spec/usage need for that? If instead it showed you the validation error and didn't persist bad values would that be a bad thing?

DonLakeFlyer:
@Junior00619 @Davidsastresas Can you try a new daily sometime after today?

Davidsastresas:
Sorry for the late response @DonLakeFlyer. I don't have anymore access to remote ID hardware so I can not help with testing at the moment.
 
On the stored-settings question — no, there's no spec or usage requirement to persist invalid operator IDs. Looking back at the original code it's really just a side effect of two things:

- The text field writes the fact on every keystroke (onTextChanged → checkOperatorID() + operatorIDFact.value = text) so the validation could run live as you type.
- For EU we need the full 19/20-char string (public part + the 3 secret chars) around just long enough to compute the Luhn checksum, and only once it validates do we strip it down to the 16-char public portion (sliced(0,16) in setOperatorID()).

So an invalid value only lingers in settings because nothing overwrites it until you type a valid one. In practice it's harmless — the send path is already gated on _operatorIDGood, so an invalid ID is never broadcast — but there's no reason it has to persist.

So showing the validation error and not persisting bad values wouldn't break anything, and it's cleaner. The only constraint to keep is that the field still needs the full value (incl. the secret) transiently to validate, while only the 16-char public part is what actually gets stored/sent. Your superset approach sounds right to me.

DonLakeFlyer:
Great, thanks for the info @Davidsastresas

PR #14268, opened 2026-04-08
fix(signing): defer local signing until vehicle confirms via signature verification

### LINKED ISSUE #14250: fix(signing): SETUP_SIGNING rejection causes communication loss

HTRamsey:
## Summary

- Replaces fire-and-forget SETUP_SIGNING with cryptographic confirmation: enable waits for a signed message verifiable against the pending key, disable waits for an unsigned heartbeat
- Adds 5-second timeout with user notification on rejection/no-response
- Temporarily broadens accept-unsigned callback during pending disable so unsigned heartbeats reach QGC
- Fixes pre-existing bug in `MAVLinkSigning::verifySignature()` that read CRC from wrong location in parsed messages
- Makes MockLink actually sign outgoing messages when signing is enabled (needed for confirmation flow)

## Context

Alternative to #14251. That PR used "any heartbeat = accepted" which false-positives because heartbeats prove the vehicle is alive, not that it accepted the key. This implementation uses actual signature verification as proof.

Fixes #14250

## Test plan

- [x] Enable signing with accepted key: QGC confirms via signed heartbeat, enables local signing
- [x] Disable signing on accepting vehicle: QGC confirms via unsigned heartbeat, disables local signing
- [x] Enable signing with unreachable vehicle: 5-second timeout fires, signing stays off, user notified
- [ ] Manual test with real vehicle


mrpollo:
@HTRamsey let me know when this is ready for testing, and I can run a few tests with hardware

HTRamsey:
Okay will do. FYI this is based on yours but just made a separate PR with some changes to make things quick and simple

mrpollo:
Cool, yeah, the biggest problem is that QGC could get out of sync with PX4. Once we figure that out, we should be good!

Additionally, at some point, it would be nice to add the option of loading a custom private key instead of generating one, and give users both options.

HTRamsey:
@mrpollo should be good to go

Issue #14306, opened 2026-04-28
ArduCopter duplicate trigger points in mission with MAVLink v2 Camera Server


jnomikos:
### Have you checked the latest release for fixes?

- [x] Yes, I’ve tested on the latest release

### Description

**Expected Behavior**
One trigger point per photo.

**Current Behavior**
When running a mission with ArduCopter 4.5.6 with a MAVLink v2 camera server, whenever a photo is taken, QGroundControl both gets `CAMERA_FEEDBACK` and `CAMERA_IMAGE_CAPTURED`. As a result, two photo points appear on the map per photo taken. 

I am trying to think of the best solution for this problem. Based on what I understand, relying on `CAMERA_IMAGE_CAPTURED` is usually better if it's available because you can get information on if a photo failed to be taken. `CAMERA_IMAGE_CAPTURED` is sent by the MAVLink camera server too. On the other hand `CAMERA_FEEDBACK` is sent by the flight controller itself.

Because of this info, I think a time or interval based deduplication would not be good enough. Because if you can get `CAMERA_IMAGE_CAPTURED`, you probably wouldn't want to use `CAMERA_FEEDBACK`.

But `CAMERA_FEEDBACK` can be useful for other camera types.

Perhaps the solution could be:
If QGC detects a heartbeat from a component with a Camera capability, it should ignore CAMERA_FEEDBACK from the Autopilot for map-plotting purposes, treating the Camera Component as the primary source of truth.

**Steps To Reproduce**
1. Fly survey in QGC with ArduCopter with camera server connected
2. Notice that trigger points are duplicated

**Additional Details**
1. I tested on both ArduCopter 4.5.6 and 4.6.3 SITL and same issue occurs
2. I am using MAVSDK for my camera server

Add any extra info (errors, warnings, etc.) here.

SITL ArduCopter 4.5.6
https://github.com/user-attachments/assets/1d413389-919b-43b9-a86c-afa226f3279b

SITL ArduCopter  4.6.3

https://github.com/user-attachments/assets/47ea3ee2-51ef-4910-a69f-5b9e72d32947

### Platform

Linux

### Flight Stack

ArduPilot

### System Information

QGC Version: Master
QGC Build: Daily
Operating System: Linux
Flight Controller: Cube Orange, SITL
AutoPilot Software: ArduCopter 4.5.6, ArduCopter 4.6.3

PR #14307, opened 2026-04-28
fix: APM duplicate trigger points bug fix

jnomikos:
Fixed bug where QGroundControl counts trigger points from both CAMERA_FEEDBACK and CAMERA_IMAGE_CAPTURED at the same time. Now, if a MAVLink Camera is detected, we ignore CAMERA_FEEDBACK messages, as CAMERA_IMAGE_CAPTURED from the MAVLink camera server should be the authoritative source of camera feedback.

## Description
On ArduPilot, when running a mission with a MAVLink camera server running, the autopilot will send CAMERA_FEEDBACK on photo taken and the MAVLink camera server will send CAMERA_IMAGE_CAPTURED. QGroundControl indiscriminately counts both as photo points. So, duplication happens.

## Type of Change

- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [x] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [x] ArduPilot

## Checklist

## Related Issues
#14306

---

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -767,7 +767,9 @@ void Vehicle::_handleCameraFeedback(const mavlink_message_t& message)
 {
     // If CAMERA_IMAGE_CAPTURED is supported, then CAMERA_FEEDBACK is redundant and should be ignored
     // to avoid duplicate points.
-    if(_cameraImageCapturedMessageAvailable) return;
+    if (_cameraImageCapturedMessageAvailable) {
+        return;
+    }
 
     mavlink_camera_feedback_t feedback;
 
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
This seems a bit fragile. Does it really a guarantee that we'll get that specific messages if its a v2 camera. The way we handle this in other cases of dups is keep a flag to determine if we've seen one or the other telemetry with one taking precedence over the other if we see both. Does that make sense for this case as well?

jnomikos:
> This seems a bit fragile. Does it really a guarantee that we'll get that specific messages if its a v2 camera. The way we handle this in other cases of dups is keep a flag to determine if we've seen one or the other telemetry with one taking precedence over the other if we see both. Does that make sense for this case as well?

I agree that makes sense for this case and would lead to something way less fragile.

jnomikos:
@DonLakeFlyer I just updated the implementation and ran an ArduCopter SITL mission piped to a real camera to verify it works

DonLakeFlyer:
Can you clean up the coding style in here? 

```
    // Space between if and paran condition
    if (foo) {
       // Always a braces body even for single line. Prevent later bugs when a single line turns into multi-line
    }
```

Normally I would't complain and just take it through, but you've been doing a bunch on contributions (thank you!) so good to get style right?

HTRamsey:
pre-commit will hopefully take care of that automatically eventually, or we add a bot command to format

jnomikos:
> pre-commit will hopefully take care of that automatically eventually, or we add a bot command to format

I know some other repos use clang-tidy successfully (e.g PX4). Has that been considered in the past?

If clang-tidy is implemented, a format check workflow could be added which automatically verifies if all code passes the format rules. And then a script could exist which uses clang-tidy to format the code automatically.

HTRamsey:
yeah that's already a part of pre-commit but I only have it detect issues, not fix them

PR #14322, opened 2026-05-01
Joystick: replace polling type state machine with PollingFlags bitmask

DonLakeFlyer:
## Summary

Replaces the confusing `_currentPollingType`/`_previousPollingType` save/restore pattern in `Joystick` with a single `PollingFlags` bitmask. This makes it possible for vehicle polling and configuration polling to coexist explicitly, and eliminates the fragile "delayed start" logic where a vehicle poll request had to be deferred if configuration polling was active.

Also fixed a bug related to stopping polling and the starting polling. The second start after stop would fail because of a stale flag.

## Changes

- Add `PollingFlag` enum (`PollingNone=0x0`, `PollingForVehicle=0x1`, `PollingForConfiguration=0x2`) with `using PollingFlags = QFlags<PollingFlag>`
- Remove `_currentPollingType`, `_previousPollingType`, and `_pollingTypeToString()`
- Add `_pollingFlags` bitmask member; all four polling methods updated to use `|=` / `&= ~PollingFlags(...)` / `.testFlag()`
- Extract `_startPollingThread()` / `_stopPollingThread()` private helpers with consistent `qCDebug` logging
- `_startPollingThread()` cancels a pending stop if the thread is still running, fixing a subtle race where a rapid stop+start sequence could leave the thread permanently dead
- Rename `_exitThread` → `_exitPollingThread` for clarity
- Add `_pollingFlagsToString()` that handles all flag combinations including both flags set simultaneously

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -160,7 +160,7 @@ Joystick::Joystick(const QString &name, int axisCount, int buttonCount, int hatC
 
 Joystick::~Joystick()
 {
-    _exitThread = true;
+    _exitPollingThread = true;
     if (isRunning()) {
         if (QThread::currentThread() == this) {
             qCWarning(JoystickLog) << "Skipping wait() on joystick thread";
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
@AndKe I'm hoping this fixes the last set of bugs where the Joystick isn't polling when it should be! Famous last words...

AndKe:
@DonLakeFlyer It looks good, and works.  :)  (tested in SITL, tested calibration too)

DonLakeFlyer:
> and works

Kinda shocking, given how many iterations I had to go through :)

PR #14331, opened 2026-05-05
AnalyzeView: unify onboard-log download into single page with transport facade

HTRamsey:
## Summary

Collapses the parallel **Onboard Logs** and **Onboard Logs (FTP)** AnalyzePages and their controllers into one. A new `OnboardLogController` facade picks the LOG protocol or MAVLink-FTP transport per active vehicle based on `MAV_PROTOCOL_CAPABILITY_FTP`, eliminating duplicate UI, controllers, and entry types.

## Why

The two pages did the same job — list and download flight logs from the vehicle — over different transports. Users had to know which page their firmware supported. The QML pages and entry classes were ~95% duplicate; the controllers shared an identical state machine (refresh → list → download → cancel) and diverged only in transport.

### New layout (everything under `src/AnalyzeView/OnboardLogs/`)

```
OnboardLogController.{h,cc}    # QML_SINGLETON facade (chooses + forwards)
OnboardLogTransport.{h,cc}     # abstract base — model/state/signals contract
LogProtocolTransport.{h,cc}    # MAVLink LOG_REQUEST_LIST/LOG_DATA implementation
FtpTransport.{h,cc}            # MAVLink FTP implementation
OnboardLogEntry.{h,cc}         # shared row item (id/time/size/status/selected/ftpPath)
OnboardLogDownloadData.{h,cc}  # LOG-protocol-only chunk-tracking helper
OnboardLogPage.qml             # single AnalyzePage entry
OnboardLogTable.qml            # shared row grid, parameterized on `controller`
OnboardLogIcon.svg
```

### Architecture

- `OnboardLogController` is now a thin `QML_SINGLETON` facade. It owns one `LogProtocolTransport` and one `FtpTransport` as children, picks one as `_active`, and forwards `refresh/download/cancel/eraseAll` plus the QML-facing properties (`model`, `requestingList`, `downloadingLogs`, `canErase`, `transportName`).
- Transport selection runs on `MultiVehicleManager::activeVehicleChanged`, on `Vehicle::capabilityBitsChanged` (handles late-arriving caps), and when the user toggles the override Fact at runtime.
- `LogProtocolTransport` (was `OnboardLogController`) and `FtpTransport` (was `OnboardLogFtpController`) are demoted from QML singletons. Their bodies are unchanged except for inheriting from `OnboardLogTransport` instead of `QObject`.
- `OnboardLogEntry` (was `QGCOnboardLogEntry`) is the shared model item; FTP path stays empty for LOG-protocol entries.
- `OnboardLogPage.qml` shows **Erase All** only when the active transport advertises `canErase`. Description text switches between LOG-protocol and MAVLink-FTP framings via `transportName`.
- The second `analyzePages()` entry in `QGCCorePlugin.cc` and the `OnboardLogsFtp/` subdirectory are removed.

### FTP erase parity

`FtpTransport` overrides `canErase()=true` and implements `eraseAll()` via `FTPManager::deleteFile()`, processing the model sequentially. Successful deletes drop the entry from the model; failures leave it with an Error status. `cancel()` also tears down in-flight delete state.

### Settings override

New `MavlinkSettings.onboardLogTransport` enum Fact (`Auto` / `LOG protocol` / `MAVLink FTP`, default `Auto`). The facade re-evaluates on Fact change so users can force a transport when capability detection is unreliable.

### Folder cleanup

- `FtpTransport` logging-category declaration moved out of the header (no external consumer).
- Stale `class QThread;` forward decl removed from `LogProtocolTransport.h`.
- Drive-by null-check fix in `LogProtocolTransport`'s PX4 `SYS_LOGGER` parameter lookup (was already unsafe before this PR).

## Compatibility

- **QML public API:** `OnboardLogController` retains its singleton name; existing QML bindings continue to work, now resolved against the facade.
- **Custom builds:** Downstream `QGC_CUSTOM_BUILD` plugins that listed both AnalyzePages or referenced `OnboardLogFtpController` will need a one-line update — the FTP class is now `FtpTransport` and is no longer a QML singleton.

## Test plan

- [x] Builds clean (Linux, Ninja, Debug + Release)
- [x] All 14 `AnalyzeView` tests pass (`ctest -L AnalyzeView`)
- [x] `OnboardLogDownloadTest` integration test exercises the facade → LOG-transport path on a non-FTP mock vehicle
- [ ] Tested manually against PX4 SITL (FTP transport, list + download + erase)
- [ ] Tested manually against ArduPilot SITL (LOG transport, list + download + erase)
- [ ] Tested the override Fact swap at runtime (Auto ↔ Log ↔ FTP) with a connected vehicle
- [ ] Verified the **Erase All** button hides on a non-FTP vehicle when override = Auto and shows on a connected FTP vehicle
- [ ] Verified `pageDescription` swaps text on transport change

## Follow-ups (intentionally out of scope)

1. Translation pass — both old `tr()` strings ("Log Refresh" / "Onboard Log Refresh", etc.) need consolidation in `qgc_*.ts`.
2. Compression UI — `LogProtocolTransport` still exposes the `compressLogs`/`compressing`/`compressionProgress` Q_PROPERTYs, but the facade doesn't forward them. Wire them up if/when the UI lands.
3. ArduPilot FTP RemoveFile robustness check — older ArduPilot firmware may not honor the delete request; consider falling back to LOG-protocol erase even when FTP is the listing transport.


HTRamsey:
@julianoes just merging your FTP log stuff in to one tab

julianoes (changes requested):
> picks the LOG protocol or MAVLink-FTP transport per active vehicle based on MAV_PROTOCOL_CAPABILITY_FTP

You can't necessarily do that. I was told that ArduPilot boards without SD card but logs in flash, won't support this. We have to check whether they are setting MAV_PROTOCOL_CAPABILITY_FTP or not!

Also, download via FTP does not have date and time for ArduPilot (see discussion in https://github.com/ArduPilot/ardupilot/pull/32860), so this feature needs to be opt-in or at least manually selectable which is the reason why I made it as a separate tab.

You can of course move it to the same tab and show a selector instead to select the backend.

PR #14332, opened 2026-05-05
Video: unify rendering on videoconvert→appsink + cross-platform zero-copy

HTRamsey:
## Summary

Replaces the per-platform custom GStreamer sinks (`qml6glsink`, `qml6d3d11sink`) with a single `qgcvideosinkbin` that pushes frames to a `QVideoSink` via `appsink` + a thin adapter (`GstAppSinkAdapter`). Adds optional zero-copy GPU paths (DMABuf/EGLImage on Linux, D3D11 on Windows, IOSurface on macOS, AHardwareBuffer on Android) with automatic per-frame fallback to a CPU copy if a GPU import fails.

The Linux GPU path routes through `glupload` so va's `DMA_DRM` DMABuf output is imported into GL textures (still zero-copy) and consumed as `GLMemory` by appsink — works on GStreamer 1.24+ with the Mesa va plugin without needing 1.26's `format=DMA_DRM` appsink negotiation.

## Why

- Single rendering pipeline shrinks the per-platform surface area (no more custom QML sink plugins to maintain per Qt version, no `gstqml6gl/`, no `gstqml6d3d11/`).
- Consolidates the video → Qt boundary in `GstAppSinkAdapter`, which makes telemetry, fallback, and lifecycle uniform across receivers.
- Zero-copy is now automatic when the decoder + driver supports it; users no longer pick between sinks.

### Pipeline
- New element `qgcvideosinkbin` (LGPL, ships in-tree): one bin with two construction modes selected by a construct-only `gpu-zerocopy` GObject prop.
  - **CPU path** (`gpu-zerocopy=FALSE`): `videoconvert → appsink` accepting widely-supported formats. Always works.
  - **GPU path** (`gpu-zerocopy=TRUE`): platform-specific zero-copy. On Linux: `glupload → appsink` with `memory:GLMemory` caps. On other platforms: bare `appsink` with the platform's native memory feature in caps (`memory:D3D11Memory`, `memory:AHardwareBuffer`, etc.).
- Old custom sinks (`qml6glsink`, `qml6d3d11sink`) removed.

### Adapter (`GstAppSinkAdapter`)
- Pulls samples from appsink, parses caps once per caps-change (cached), dispatches to the matching `QHwVideoBuffer` subclass via `makeHwVideoBuffer()` factory.
- Per-frame factory returns `nullptr` ⇒ adapter copies to system memory (CPU fallback path) instead of crashing.
- `Q_PROPERTY` telemetry: `gpuFrameCount`, `cpuFrameCount`, `gpuFallbackCount` for runtime introspection (1 Hz throttled signal).
- DMA_DRM caps (GStreamer 1.24+) parsed via `gst_video_info_dma_drm_from_caps` so future decoders that emit them work without further changes.
- Reads `gpu-zerocopy` back from the bin via `g_object_get` after construction so the adapter's path mode can never disagree with the bin's actual pipeline (eliminates a class of fact-vs-bin desync bugs).

### `QHwVideoBuffer` implementations
- `GstDmaBufVideoBuffer` — Linux DMABuf via EGLImage import, modifier-aware (`EGL_EXT_image_dma_buf_import_modifiers`).
- `GstGlVideoBuffer` — generic GLMemory wrap with proper `GstGLSyncMeta` flush before texture access.
- `GstD3D11VideoBuffer` — Windows D3D11Memory share with QRhi's D3D11 backend; ref-leak-safe on early-exit.
- `GstIOSurfaceVideoBuffer` — macOS CoreVideo CVPixelBuffer via `gst_apple_core_video_memory_get_pixel_buffer()` (replaces a UB pointer-cast heuristic).
- `GstAHardwareBufferVideoBuffer` — Android AHardwareBuffer via `eglGetNativeClientBufferANDROID` (gst-plugins-bad ≥ 1.28).
- `GstHwVideoBuffer` base + `GstHwVideoBufferFactory` consolidate the previously copy-pasted GstSample/info/format ownership and dispatch.

### Context bridges
- `GstGlContextBridge` — Linux GL context bridge for glupload's `NEED_CONTEXT` bus message. Tries Qt's EGL native interface first (Wayland/eglfs/xcb_egl), falls back to GLX (the xcb default on Qt 6 desktop) wrapping `GstGLDisplayX11` + `GST_GL_PLATFORM_GLX`.
- `GstD3D11ContextBridge` — Windows D3D11 device wrap via `gst_d3d11_device_wrap`.
- `GstContextBridgeRegistry` — replaces a hardcoded `#if` chain in `GstVideoReceiver` with a registry-based dispatcher.

### Settings
- `gpuZeroCopyEnabled` (default off, "experimental") replaced with **`forceCpuVideoPath`** (default off, no "experimental" framing). Default behavior is now: use GPU when the decoder supports it, fall back to CPU per-frame on import failure. The new flag is a debug escape hatch that forces the CPU path.
- `forceCpuVideoPath=true` and `forceVideoDecoder=Software` both resolve to the CPU pipeline. Software decoder + GPU path is structurally impossible (SW decoders output system memory) — handled silently in `GStreamer::createVideoSink`, no warning needed.

### Tests
- 30 GStreamer unit tests; new ones cover: CPU/GPU bin construction, GLMemory dispatch, DMABuf fallback (synthesized via factory mock), caps cache invalidation, teardown under load, context bridge registry fanout, Q_PROPERTY telemetry signal, `QGCRhiCapture` cache lifecycle, CPU memcpy active-row stride correctness.

## Verified on Linux desktop (xcb, GStreamer 1.24.2)

RTSP H.265 4K @ 10 fps from a Dahua camera, vah265dec selected by decodebin3:

| Setting | Bin pipeline | Allocator | Frame stats (1 min) |
|---|---|---|---|
| Default | `glupload → appsink` | `GLMemoryPBO` | `GL:627/0` |
| `forceCpuVideoPath=true` | `videoconvert → appsink` | `SystemMemory` | `CPU:387` |
| `forceVideoDecoder=Software` | `videoconvert → appsink` | `SystemMemory` | `CPU:n` |

Start/stop exercised across multiple process cycles; clean teardown (`Adapter teardown — CPU:0 DMABuf:0 GL:627 GL-failures:0`), no leaks reported.

## Performance

Same throughput across all three modes when the camera is the bottleneck. Differences scale with resolution and stream count:

- **Default GPU path** — 0 CPU memcpy per frame, ~0 PCIe round-trip bandwidth.
- **Force CPU** — 1× GPU→CPU download (videoconvert) + 1× CPU→GPU upload (QRhi) per frame. ~750 MB/s memory bandwidth at 4K@30, ~1.5 GB/s at 4K@60.
- **Force SW decoder** — adds CPU decode cost (4K HEVC ~150–300% of one core); the only path with a hard real-time ceiling.

The GPU path is what makes multi-stream viable; per-stream marginal cost on the GPU path is essentially decode + render budget, with no cross-bus traffic.

## Test plan

- [x] Linux desktop (xcb, GStreamer 1.24.2): RTSP H.265 4K, default + `forceCpuVideoPath=true` + `forceVideoDecoder=Software`
- [x] All 30 GStreamer unit tests pass
- [ ] Wayland session
- [ ] Windows + D3D11 zero-copy
- [ ] macOS + IOSurface zero-copy
- [ ] Android + AHardwareBuffer zero-copy
- [ ] GStreamer 1.28 SDK (CI)
- [ ] Multi-stream concurrency (videoContent + thermalVideo)
- [ ] Recording branch (filesink) under both CPU and GPU modes


HTRamsey:
It's basically an adaptation of this for each platform: https://github.com/qt/qtmultimedia/blob/dev/src/plugins/multimedia/gstreamer/common/qgstvideobuffer.cpp

jnomikos:
I tested on both Ubuntu 24.04 and Android and it seems to be working well on my end.

Ryanf55:
Can we find a way to not rely on the `#include <private/qvideotexturehelper_p.h>`? This PR has broken compilation on Ubuntu 22, even though I am using QGC 6.10.1.

PR #14341, opened 2026-05-08
feat(analyze): add PX4 ULog support and refactor log viewer parser architecture

DonLakeFlyer:
## Summary

Adds PX4 ULog (**.ulg**) file support to the Log Viewer and refactors the DataFlash parser out of the monolithic `LogFileParser.cc` into separate, independently testable sub-parsers.

### Architecture
- **`LogFileParser`** (QML_ELEMENT) is now a thin dispatcher: reads the file extension and delegates to `DataFlashParser::parseFile()` or `ULogParser::parseFile()`
- **`LogParseResultPrivate.h`** defines the shared `LogParseResult` struct used by all sub-parsers
- **`LogViewerDataFlashParser`** — renamed from the old inline code to avoid an include-path collision with `GeoTag/DataFlashParser.h`
- **`LogViewerULogParser`** — new parser using `ulog_cpp::Reader` + `ULogFullHandler` to extract plottable fields, parameters, events (warning/error/info), mode segments, dropouts, and messages from ULog files
- Sub-parsers no longer include `LogFileParser.h`; use `QCoreApplication::translate()` instead

### QML
- `LogViewerPage.qml`: removed dead `_axisXToPixel` function; series creation uses explicit `axisX`/`axisY` bindings required by Qt Graphs 2D; incremental add/remove pattern for field series

### Tests
- `LogFileParserTest`: 8 unit tests covering ULog numeric fields, parameters, warning events, mode segments, dropouts, invalid files, and unsupported extensions; DataFlash regression guard
- `APMDataFlashLogParserTest`: fixed temp file paths to use `QDir::tempPath()` instead of hardcoded `/tmp`

## Test Plan
```
cd build && ctest --output-on-failure -L Unit -R LogFileParser
```

### DIFF
--- a/src/AnalyzeView/LogViewer/APMDataFlash/APMDataFlashLogParser.cc
+++ b/src/AnalyzeView/LogViewer/APMDataFlash/APMDataFlashLogParser.cc
@@ -19,13 +19,13 @@ namespace {
 struct ParseResult {
     bool ok = false;
     QString errorMessage;
-    QStringList availableSignals;
-    QStringList plottableSignals;
+    QStringList availableFields;
+    QStringList plottableFields;
     QVariantList parameters;
     QVariantList events;
     QVariantList messages;
     QVariantList modeSegments;
-    QHash<QString, QVector<QPointF>> signalSamples;
+    QHash<QString, QVector<QPointF>> fieldSamples;
     double minTimestamp = -1.0;
     double maxTimestamp = -1.0;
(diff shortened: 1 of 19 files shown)


DonLakeFlyer:
I'm not super happy with the factoring of the base canonical log support interface versus the px4/ardupilot specific case. But it works now so I'll come back to that later.

PR #14347, opened 2026-05-09
Fix parameter file diff loader and compass-cal refresh popup spam

### LINKED ISSUE #14346: Parameter file import incorrectly reports “There are no differences”

### Description

QGroundControl incorrectly reports:
"There are no differences between the file loaded and the current settings on the Vehicle."

even when the parameter values in the file are different from the current values on the flight controller.

Example:
Current FC value:
SERVO_BLH_AUTO = 0

Parameter file contains:
1 1 SERVO_BLH_AUTO 1 2

or:
SERVO_BLH_AUTO,1

QGC still reports no differences and does not apply the parameter.

Notes:

Parameters can be changed manually in QGC.
The issue only affects loading from file.
Older QGC versions appear to work correctly.
Running under XCB instead of Wayland does not fix the issue.

Additional observations:
QGC AppImage on Fedora 44 also shows multiple QML binding loop warnings in console output.

### Platform

Linux

### Flight Stack

ArduPilot

### System Information

Fedora 44
GNOME Wayland
QGroundControl AppImage (latest version)
ArduPilot 4.6.3
SpeedyBee F405 V3

### Log Files and Screenshots

QGroundControl incorrectly reports:
"There are no differences between the file loaded and the current settings on the Vehicle."

even when the parameter values in the file are different from the current values on the flight controller.

Example:
Current FC value:
SERVO_BLH_AUTO = 0

Parameter file contains:
1 1 SERVO_BLH_AUTO 1 2

or:
SERVO_BLH_AUTO,1

QGC still reports no differences and does not apply the parameter.

Notes:

Parameters can be changed manually in QGC.
The issue only affects loading from file.
Older QGC versions appear to work correctly.
Running under XCB instead of Wayland does not fix the issue.

Additional observations:
QGC AppImage on Fedora 44 also shows multiple QML binding loop warnings in console output.

Image
Image
Image

HTRamsey:
## Summary

Two parameter-system UX bugs surfaced together in #14346.

- **Parameter diff loader**: `ParameterEditorController::buildDiffFromFile` only split on `\t`, so files with whitespace-aligned columns (Mission Planner / ArduPilot tooling, e.g. `1 1 SERVO_BLH_AUTO 1 2`) and simple `name,value` CSVs produced zero parsed entries and the dialog reported "There are no differences between the file loaded and the current settings on the Vehicle" even when values clearly differed. Now splits on `[\s,]+` and accepts both the 5-column and a new 2-column format.
- **Post-calibration refresh popup spam**: after a compass cal stops (especially cancel) `_refreshParams` requests a bulk refresh of `COMPASS_*` / `INS_*` (APM) or `CAL_*` / `SENS_*` (PX4). The FC is often slow to answer mid-cancel, every timeout fired a modal "Parameter read failed" via `QGC::showAppMessage`, and dozens of modals stacked on the QML root and froze the UI. Added a `notifyFailure` flag (default true, preserves existing call sites) to `ParameterManager::refreshParameter` / `refreshParametersPrefix` and pass `false` from the post-cal refresh paths in both sensors controllers. The state machine still logs and emits `_paramRequestReadFailure`; only the modal is suppressed for the batch refresh.

Fixes #14346.

## Test plan

- [ ] Load a Mission Planner whitespace-separated `.param` file with at least one differing value → diff dialog lists differences and applies them on OK.
- [ ] Load a `name,value` CSV → same.
- [ ] Load a tab-separated QGC-exported file with all matching values → dialog still says "no differences" (regression check).
- [ ] Start a compass calibration on ArduPilot, cancel it → no flood of "Parameter read failed" modals; UI stays responsive.
- [ ] Run a normal compass cal to completion on ArduPilot and PX4 → params repopulate, no popup spam.
- [ ] `ctest -L Unit -R Parameter` passes.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSensorsComponentController.cc
+++ b/src/AutoPilotPlugins/APM/APMSensorsComponentController.cc
@@ -381,13 +381,16 @@ void APMSensorsComponentController::_refreshParams()
         QStringLiteral("INS_ACCOFFS_X"), QStringLiteral("INS_ACCOFFS_Y"), QStringLiteral("INS_ACCOFFS_Z")
     };
 
+    // After a calibration finishes (success or cancel) the FC can be slow to answer the
+    // bulk refresh, so suppress the per-parameter "read failed" popups — otherwise dozens
+    // of modals stack up and freeze the UI (e.g. cancelling compass cal on ArduPilot).
     for (const QString &paramName : fastRefreshList) {
-        _vehicle->parameterManager()->refreshParameter(ParameterManager::defaultComponentId, paramName);
+        _vehicle->parameterManager()->refreshParameter(ParameterManager::defaultComponentId, paramName, false /* notifyFailure */);
     }
 
     // Now ask for all to refresh
-    _vehicle->parameterManager()->refreshParametersPrefix(ParameterManager::defaultComponentId, QStringLiteral("COMPASS_"));
-    _vehicle->parameterManager()->refreshParametersPrefix(ParameterManager::defaultComponentId, QStringLiteral("INS_"));
+    _vehicle->parameterManager()->refreshParametersPrefix(ParameterManager::defaultComponentId, QStringLiteral("COMPASS_"), false /* notifyFailure */);
+    _vehicle->parameterManager()->refreshParametersPrefix(ParameterManager::defaultComponentId, QStringLiteral("INS_"), false /* notifyFailure */);
 }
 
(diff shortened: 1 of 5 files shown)


HTRamsey:
@DonLakeFlyer this is a quick sample to show you what fixed some of the issues, in case you want to make your own PR based on this

DonLakeFlyer:
Ok, thanks. I'll take a look...

DonLakeFlyer:
Both problems dealt with in Pulls

DonLakeFlyer:
I did not add support for .csv

PR #14351, opened 2026-05-10
APM: Always fetch latest ArduPilot ParameterRepository metadata

DonLakeFlyer:
## Summary

Replace the pinned `CPMAddPackage` `GIT_TAG` for `ArduPilot/ParameterRepository` with a `git clone`/`fetch` approach that always tracks HEAD.

## Problem

CPM caches packages by a hash of the tag string and never re-fetches once the directory exists. Using a pinned commit hash means:
- New firmware version directories (e.g. `Copter-4.8`) are never picked up
- Updated parameter descriptions, ranges, and units are never picked up
- Someone must manually bump the hash and rebuild to get any improvements

## Solution

- **First configure**: `git clone --depth=1 --progress` into `${CPM_SOURCE_CACHE}/ardupilotparams/HEAD` so the clone survives `rm -rf build/` and is shared across Debug/Release/etc. build trees
- **Subsequent configures**: `git fetch --depth=1 origin` + `reset --hard origin/HEAD`, which is robust to force-pushes (unlike `pull --ff-only` on a shallow clone)
- **Network failure**: emits a `WARNING` and continues with cached metadata so offline builds still work
- **Clone failure**: `FATAL_ERROR` with the full git error message for easy diagnosis

### DIFF
--- a/src/FirmwarePlugin/APM/CMakeLists.txt
+++ b/src/FirmwarePlugin/APM/CMakeLists.txt
@@ -45,11 +45,48 @@ qt_add_resources(${CMAKE_PROJECT_NAME} firmaware_plugin_apm_resource
 
 #===========================================================================#
 
-CPMAddPackage(
-    NAME ArduPilotParams
-    GITHUB_REPOSITORY ArduPilot/ParameterRepository
-    GIT_TAG 037ff6bcfc080608544a4526f2c22406f2cce60f
-)
+# Always track HEAD of the ParameterRepository.
+# CPM cannot do this (it caches by tag string and never re-fetches a branch).
+# Instead, clone on first configure and pull on every subsequent configure so
+# the embedded metadata is always up to date with new firmware versions and
+# updated parameter descriptions.
+# Store alongside other CPM packages so the clone survives build-dir wipes and
+# is shared across Debug/Release/etc. build trees.
+find_package(Git REQUIRED)
+set(_ardupilot_params_dir "${CPM_SOURCE_CACHE}/ardupilotparams/HEAD")
+if(NOT EXISTS "${_ardupilot_params_dir}/.git")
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@HTRamsey Can you look at this? having to continuously bump the previous tag is bad. We pull ardupilot dev build metadat from here so it is constantly changing.

HTRamsey:
@DonLakeFlyer I thought we wanted static tags so that every build is guaranteed to be the same? It's harder to diagnose issues when two people could be on different tags

DonLakeFlyer:
Not for parameter metadata. Since ArduPilot doesn't support COMPONENT_INFORMATION for parameter metadata which allows metadata to change firmware to firmware version we need to do it this way manually. Otherwise we will always be out of date on dev builds from ArduPilot. This is similar to PX4 build bot jamming in PX4 metadata on a schedule. The only metadata which would be changing over time is the last dev build version. The older ones should be static.

HTRamsey:
wait why aren't we caching this with CPM?

DonLakeFlyer:
It's not possible to have CPM track a HEAD commit

HTRamsey:
What do you think about doing something like this instead?
[apm-track-master.patch](https://github.com/user-attachments/files/27610664/apm-track-master.patch)
[cpm-update-strategy.patch](https://github.com/user-attachments/files/27610665/cpm-update-strategy.patch)

DonLakeFlyer:
I'm not sure what all that does other than it looks like more code then my pull. That said, I don't really care how it's accomplished. You know this stuff better than me. All I care about is the end result :)

HTRamsey:
It's just add a command in to CPM for these kinda scenarios in the future. (refreshing cached git when HEAD is updated). So yeah it's more code but it's also a template for a permanent solution

DonLakeFlyer:
@HTRamsey I'll close this since you are going to do it in a better way...

PR #14363, opened 2026-05-11
LogViewer: GPS map, Parameters tab, performance, and bug fixes

DonLakeFlyer:
## Summary

Extends the Log Viewer with a GPS flight path map, a full Parameters tab, performance improvements, and several bug fixes.

### Map tab
- Displays GPS flight path overlaid on a `FlightMap` for both PX4 ULog and APM DataFlash logs
- APM DataFlash: uses `GPS.Lat`/`GPS.Lng` with `Status >= 3` filter (3D fix required) as primary source; falls back to `POS.Lat`/`POS.Lng` (EKF-fused)
- PX4 ULog: uses `vehicle_global_position.lat`/`lon`, falling back through `vehicle_gps_position` / `sensor_gps` field name variants
- Altitude mini-chart (`LogViewerAltChart`) below the map with zoom and a min/max popup
- Clicking or dragging the altitude chart moves a yellow position dot on the map
- Cursor is linked between the altitude chart and the main Charting tab

### Parameters tab
- "Changed only" toggle (hides parameters matching their system default)
- Search over name, value, and short description
- Per-row: value formatted with units and decimal places from FactMetaData; enum values shown as label + number
- Orange indicator dot for parameters that differ from their system default
- Default value shown alongside for changed parameters

### Messages tab
- Timestamped rows with alternating background

## Performance improvements

- `LogFileParser::fieldSamplesFiltered()`: bucketed min/max downsampling — one pixel column per bucket, up to 4 representative points per bucket (first, min-y, max-y, last), returned in time order
- `LogFileParser::fieldMinMax()`: compute full-dataset min/max in C++ and return a two-value map, eliminating the need to marshal all `QPointF` samples across the C++/QML bridge just to find the range
- `LogViewerChart` and `LogViewerAltChart` both updated to use `fieldMinMax()` and `fieldSamplesFiltered()`

## Bug fixes

- `ULogFullHandler`: `qFuzzyCompare(0.0, 0.0)` returns `false` (Qt documented limitation — relative epsilon is undefined at zero); replaced with exact equality check for the zero case
- `LogViewerPage`: scroll position of the field list is now preserved when a field is toggled on/off
- `LogViewerController`: removed unused `statusText` property

## New files

| File | Purpose |
|------|---------|
| `LogViewerAltChart.qml` | Altitude mini-chart with zoom, cursor, and min/max popup |
| `LogViewerParamMetaData.cc/.h` | Enriches parameter rows from bundled PX4/APM FactMetaData JSON |

## Tests added

- `_fieldSamplesFilteredComprehensiveTest`: 500-sample synthetic dataset; verifies bucketing output count and landmark preservation under full-range, passthrough, zoom, and out-of-range conditions
- `_gpsPathULogVehicleGlobalPositionTest`
- `_gpsPathULogVehicleGpsPositionLatDegTest`
- `_gpsPathAPMDataFlashPOSTest`

### DIFF
--- a/docs/en/SUMMARY.md
+++ b/docs/en/SUMMARY.md
@@ -58,6 +58,7 @@
   - [Console Logging](qgc-user-guide/settings_view/console_logging.md)
   - [Virtual Joystick (PX4)](qgc-user-guide/settings_view/virtual_joystick.md)
 - [Analyze](qgc-user-guide/analyze_view/index.md)
+  - [Log Viewer](qgc-user-guide/analyze_view/log_viewer.md)
   - [Log Download](qgc-user-guide/analyze_view/log_download.md)
   - [GeoTag Images (PX4)](qgc-user-guide/analyze_view/geotag_images.md)
   - [MAVLink Console (PX4)](qgc-user-guide/analyze_view/mavlink_console.md)
--- a/docs/en/qgc-user-guide/analyze_view/log_viewer.md
+++ b/docs/en/qgc-user-guide/analyze_view/log_viewer.md
@@ -1,20 +1,79 @@
 # Log Viewer (Analyze View)
 
-The _Log Viewer_ screen (**Analyze > Log Viewer**) provides a unified workflow for post-flight analysis:
+The _Log Viewer_ screen (**Analyze > Log Viewer**) is a unified post-flight analysis tool that supports three log formats:
 
-- Open ArduPilot DataFlash logs (`.bin`/`.log`) for parameter and event inspection.
-- Open telemetry logs (`.tlog`) and replay them with timeline controls.
-- Reuse existing MAVLink inspection and charting tools while replaying telemetry.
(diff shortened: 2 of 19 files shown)


codeceptsDE:
Thanks for making the 'What's New Since Stable V5.0' section available, @DonLakeFlyer! I was looking for something like this!

It does, however, step on on the toes of the (outdated) [daily_build_new_features.md](https://github.com/mavlink/qgroundcontrol/blob/master/docs/en/qgc-user-guide/releases/daily_build_new_features.md) a little bit. And [CHANGELOG.md](https://github.com/mavlink/qgroundcontrol/blob/master/CHANGELOG.md), for that matter. 
Not sure which one is the better place to put it, but maybe they should be unified :)

DonLakeFlyer:
Yeah, I need to clean that up. Those other ones should go away.

PR #14370, opened 2026-05-13
Remove bug label from issue template in favor of issue type

DonLakeFlyer:
Remove the `labels: ["bug"]` from the bug report issue template since the `type: bug` field provides the same categorization through GitHub's newer issue types system.

This avoids redundant tagging—the issue type alone is sufficient for filtering and organization.


DonLakeFlyer:
After this goes through I'll remove Report: Bug label. I've already made sure all items that have that label have Type set to Bug. This way there is only only way to note an Issue is a Bug

PR #14388, opened 2026-05-15
AnalyzeView: Preserve page state when popping out to window

### LINKED ISSUE #14373: Graphing issues:
1: 
@DonLakeFlyer  this is somewhat related to the recent graph changes.
Please note how only one of these values is properly graphed:

Image

2: the time/length on the X axis is now gone, after recent update to new graphing widget.(I guess you are aware of that?)
3: After selecting some data for graphing, and clicking the 

Image
...pop-out icon, selection is lost.  (it would be very nice to have it pop out with actual selection - because the workflow is often : "find_value, graph it, popout for monitoring")

-
4: a confusing bug:  please note how only one value is selected for plot 1, and yet we see a line at 32767, one below zero, and two colors/two lines at 12600   (at the times moves, the orange/red part keeps position.)
To reproduce, just select/check/uncheck graphs.

Image

DonLakeFlyer:
Fixes #14373

## Problem
When popping out the MAVLink Inspector (or any Analyze page) to a separate window, all configured chart state (plotted fields, series, scale selections) was lost. The popout created a fresh instance via a Loader, discarding the live QML item.

## Solution
Replace Loader-based page management in AnalyzeView with manual `Qt.createComponent`/`createObject` lifecycle management. When popout is triggered:

1. The existing page item's ownership is released (not destroyed)
2. The live item is reparented directly into the popout Window
3. A fresh instance is created in-place for the main AnalyzeView
4. The popout Window starts hidden until reparenting is complete, avoiding `qt.graphs2d` "window doesn't exist" errors

This preserves all page state (charts, fields, scrollback) in the popout window.

### DIFF
--- a/src/AnalyzeView/AnalyzeView.qml
+++ b/src/AnalyzeView/AnalyzeView.qml
@@ -18,20 +18,34 @@ Rectangle {
 
     property var  _activeVehicle: QGroundControl.multiVehicleManager.activeVehicle
     property var  _currentPage:   null
+    property var  _currentItem:   null
+
+    function _loadPage(source) {
+        if (_currentItem) {
+            _currentItem.destroy()
+            _currentItem = null
+        }
+        if (source !== "") {
+            var component = Qt.createComponent(source)
+            if (component.status === Component.Ready) {
+                _currentItem = component.createObject(panelContainer)
+            }
+        }
+    }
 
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Fixed in daily

PR #14389, opened 2026-05-15
MAVLink Inspector: instance awareness, array expansion, debug separation, chart improvements

### LINKED ISSUE #7313: Improvement: Mavlink Inspector: Don't use scientific notation on bitfields
Moved from: bluerobotics/qgroundcontrol#174
Original author: @jaxxzer

### Issue description
:roll_eyes: This is not helpful:
image

### LINKED ISSUE #8454: Analize plugin missing feature
Hi,
I have just upgrade to version 4.0.0 of QGCS and notice that MAVLink Inspector and Analize plugin get merged. There is a problem however. You can plot only 4 items at once and can't plot specific item from table variable (like field `voltages' from message `BATTERY_STATUS`). I am developing autopilot software and these features are very helpful.

Any chance to bring at least analize tool back?

### LINKED ISSUE #9578: viewing multiple debug values in Mavlink Inspector
Hello,

in qgroundcontrol v3.5.1 I am able to graph multiple debug values in mavlink inspector , however with the same working code, if I use qgroundcontrol v4.1.2 I can not graph multiple values.
I can find the values under Named_Value_float, but they alternate under the same name and can not be separated. 
could not find any reference here or elsewhere to solve this issue. 

I'm using the qgroundcontrol on a windows machine, pixhawk 1 and pixhawk 4 controllers runing px4 v1.9.4

### LINKED ISSUE #9581: Allow more than 60s of plots in MAVLink Inspector
I used to be able to plot up to 10 minutes of MAVLink data, but now the options only go up to "60 Sec". Any reason for removing the longer options? It was useful for having real time and long lasting visualizations and seems easy enough to allow. Was it removed for a particular reason?

### LINKED ISSUE #12234: MAVLink Inspector: Add back Instance Awareness
MAVLink messages that have an "instance field" should support separate plotting and reporting for each value of the instance field.

As an example consider the `BATTERY_STATUS` message. This has the `id` field with the `instance="true"` attribute as shown below:

```
<field type="uint8_t" name="id" instance="true">Battery ID</field>
```

If a vehicle has multiple batteries then each of these should have a different value for id and should be plotted separately. As it is now a plot of `BATTERY_STATUS` is useless for multi-battery systems, because the message plot shows traces that switch between different sources.

 Currently this affects batteries, ESC, IMUs, but there are others, and more to come.

Note that this was previously supported, but was lost when transitioning to QML controls.

---- 
- What problem does it solve?

  > Allows instance data to be plotted and inspected.

- What does it do?

  > Allows data to be presented usefully
  > All
  > All
    > Mission Planner.
  > MAVLink

DonLakeFlyer:
## Summary

Enhance the MAVLink Inspector with instance-aware message separation, array field expansion, debug message handling, and efficient chart data management.

## Changes

- **Instance awareness**: Separate messages by instance field (e.g. `sensor_id`, `target_system`) using a build-time generated lookup table from MAVLink XML (`instance="true"` attribute)
- **Array field expansion**: Numeric array fields are expanded into individual elements (e.g. `covariance[0]`, `covariance[1]`, ...) that can be individually charted
- **Debug message separation**: `NAMED_VALUE_FLOAT`, `NAMED_VALUE_INT`, `DEBUG_VECT`, and `DEBUG` messages are separated by their name/ind field as implicit instance discriminators
- **Longer time scales**: Added "2 Min" and "5 Min" options
- **Bucketed ring buffer**: Replaced unbounded data storage with a pixel-width-proportional ring buffer using min/max bucketing (2 points per pixel column, ~64KB per plotted field)
- **Chart lag fix**: In-progress bucket is rendered at current time so the series line always extends to the chart's right edge
- **Scientific notation fix**: Float/double values now use higher precision formatting to avoid scientific notation for large integer-like values (bitfields)
- **MockLink test output**: Added sin/cos wave `NAMED_VALUE_FLOAT` messages for manual testing

## Testing

- All existing unit tests pass (updated `_timeScalesCountTest` for new count)
- New `test_mavlink_instance_fields.py` (12 tests) validates the generator
- Manual testing with MockLink confirms instance separation and charting

Fixes #12234
Fixes #8454
Fixes #9578
Fixes #9581
Fixes #7313

### DIFF
--- a/src/AnalyzeView/MAVLinkInspector/MAVLinkChart.qml
+++ b/src/AnalyzeView/MAVLinkInspector/MAVLinkChart.qml
@@ -55,6 +55,7 @@ ColumnLayout {
         id:                     chartController
         inspectorController:    chartView.inspectorController
         chartIndex:             chartView.chartIndex
+        plotPixelWidth:         Math.max(1, Math.floor(_graphsView.plotArea.width))
     }
 
     // -------------------------------------------------------------------------
--- a/src/AnalyzeView/MAVLinkInspector/MAVLinkChartController.cc
+++ b/src/AnalyzeView/MAVLinkInspector/MAVLinkChartController.cc
@@ -7,6 +7,8 @@
 #include <QtGraphs/QAbstractSeries>
 #include <QtCore/QTimer>
 
+static constexpr qreal kMinDelta = 1e-6;
+
 QGC_LOGGING_CATEGORY(MAVLinkChartControllerLog, "AnalyzeView.MAVLinkChartController")
 
 MAVLinkChartController::MAVLinkChartController(QObject *parent)
@@ -78,6 +80,7 @@ void MAVLinkChartController::setRangeXIndex(quint32 index)
     emit rangeXIndexChanged();
 
     updateXRange();
+    _resetFieldBucketing();
 }
 
 void MAVLinkChartController::updateXRange()
@@ -119,7 +122,7 @@ void MAVLinkChartController::updateYRange()
         return; // No field has received data yet (sentinel values)
     }
 
-    if (qAbs(vmax - vmin) < 0.000001) {
+    if (qAbs(vmax - vmin) < kMinDelta) {
         vmin -= 1.0;
         vmax += 1.0;
     } else {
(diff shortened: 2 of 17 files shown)


hamishwillee:
Fantastic. Thank you.

PR #14400, opened 2026-05-18
FirmwareUpgrade: add Gear Up AirBrainH743 board

julianoes:
Maps PX4 board id `1209` to the build target `gearup_airbrainh743_default` so QGC's firmware selector offers the hosted PX4 firmware for the Gear Up AirBrainH743.

The board was added to PX4-Autopilot (`boards/gearup/airbrainh743`) and is built/published by PX4 CI automatically. With this entry, the `master` (developer) firmware is reachable from QGC.

Note: tagged releases upload only to versioned S3 directories, which `px4_board_name_map` does not point at, so those builds are not yet reachable via QGC for any board. PX4/PX4-Autopilot#25414 (unified runtime manifest) is the longer-term replacement for this hardcoded map.


julianoes:
Duplicate, closing.

PR #14407, opened 2026-05-18
fix(LogViewer): restore hover-driven chart cursor after extraction

AhmWael:
## Description

When `LogViewerChart` was extracted from `LogViewerPage` in #14354, the chart `MouseArea` lost `hoverEnabled` and the logic that updates the cursor while the pointer moves. The vertical marker and value popup only updated on a short click-release, which regressed inspection of DataFlash (`.bin`) and ULog charts.

This PR restores the pre-extraction desktop behavior:

- **LogViewerChart:** `hoverEnabled` on desktop, cursor updates on press and while moving (when not drag-zooming), marker hidden when the pointer leaves the chart, marker X refreshed when `plotArea` layout changes
- **LogViewerAltChart:** same hover/press/move/exit behavior for the map altitude chart and map marker sync

Mobile remains click-only (`hoverEnabled` disabled via `ScreenTools.isMobile`). 

## Type of Change

- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [x] Tested locally
- [ ] Tested with hardware

Manual verification:

1. Open **Analyze → Log Viewer** and load a `.bin` (or `.ulg`) log.
2. On the **Charting** tab, select one or more fields.
3. Move the mouse over the chart — the vertical marker and popup should follow the pointer and show time, mode, and field values (Current / Min / Max).
4. Click-drag to zoom; right-click to reset zoom.
5. On the **Map** tab (if GPS/altitude data exists), confirm the altitude chart cursor behaves the same and the map position dot tracks the cursor.

### Platforms Tested

- [ ] Linux
- [x] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [x] ArduPilot

## Checklist

---


AhmWael:
@DonLakeFlyer Can you take a look at this PR?

DonLakeFlyer:
I changed this because of mobile. I'd rather keep them the same but if you think this hover feature is critical for analysis I can live with the difference. Do you think it's important enough for feature set to differ?

AhmWael:
> I changed this because of mobile. I'd rather keep them the same but if you think this hover feature is critical for analysis I can live with the difference. Do you think it's important enough for feature set to differ?

I find it really useful for log analysis on desktop. Being able to just hover through the trend and inspect values feels much more natural than having to click every point to update the readings.

DonLakeFlyer:
Thanks, makes total sense. Sorry I didn't take something like that into account.

AhmWael:
It should be ready for review now @DonLakeFlyer

AhmWael:
@DonLakeFlyer Can you review this when you have a chance.

DonLakeFlyer:
I've played around with this and I'm not super happy with this. The reason being I think the concept of locking a cursor position is valuable. Especially with multiple charts and then you can add/remove fields to the chart and still see the values at the locked position you are trying to focus on. I think the solution is to support both hover and locked positions. I'm working on something that tries that out.

AhmWael:
@DonLakeFlyer 
I've updated the implementation to support both needs:
- Click anywhere to fixate the cursor (so it stays in place when you move the mouse to add/remove fields).
- Drag the cursor line directly to slide it horizontally and scan values to see trends.
- Drag-to-zoom remains fully functional when dragging anywhere else on the chart.

Let me know if this approach works for you!

AhmWael:
@DonLakeFlyer 
I also added a nice feature for APM .bin files: they now display absolute UTC time (deduced from the GPS messages, adjusting for leap seconds) in both the chart popups and message logs.

For PX4 logs, it falls back to displaying time in seconds since boot. I don't have any .ulg test files on hand to verify absolute time extraction for PX4, so I left it on the relative fallback for now.

DonLakeFlyer:
@AhmWael Can you take a look?

AhmWael:
@DonLakeFlyer 
I was just testing it now. I like the new approach.

Can you also add the absolute UTC time feature as I find it very useful, I added it in my last commit.

DonLakeFlyer:
Why UTC as opposed to local time?

AhmWael:
We calculate it in UTC because GPS time is inherently UTC-relative (with a leap second correction) and doesn't store timezone info. However, when we display it in QML, we use Qt.formatDateTime which automatically converts it and shows it in the local time of your computer.

DonLakeFlyer:
Ok, I'll do that in a separate pull

PR #14414, opened 2026-05-20
fix(Terrain): back off after failed tile fetches to stop warning spam

dakejahl:
## Summary

Eliminate the warning spam (and unnecessary network traffic) when the elevation server cannot return data for a tile, by caching recent failures in `TerrainTileManager` and not looping on a known-failed tile in `TerrainProtocolHandler`.

## Problem

When a vehicle sends `TERRAIN_REQUEST` with uninitialized lat=0/lon=0, the request maps to elevation tile (0, 0) — "Null Island". The Auterion terrain server returns HTTP 500 for that tile, but `TerrainTileManager` had no memory of recent failures, so every `TerrainProtocolHandler` timer tick (~12 Hz) dispatched a fresh HTTP request. Observed log shows continuous warnings from both `QtLocationPlugin.QGeoTiledMapReplyQGC` and `Terrain.TerrainTileManager::_terrainDone` at ~80–150 ms intervals:

```
Warning: 1 "Error transferring https://terrain-ce.suite.auterion.com/api/v1/carpet?points=0,0,0.01,0.01 - server replied with status code 500"
Warning: Elevation tile fetching returned error: "Error transferring https://terrain-ce.suite.auterion.com/api/v1/carpet?points=0,0,0.01,0.01 - server replied with status code 500"
```

The handler also intentionally never cleared the mask bit on a non-success result, so even with backoff added, it would just transfer the spam to its own log line.

## Solution

`src/Terrain/TerrainTileManager.{h,cc}`

- Add `_failedTiles` (tile hash → ms timestamp) with a 5 s backoff.
- While a tile is in the failed set, `getAltitudesForCoordinates` returns NaN + `error=true` without dispatching a network request.
- `_terrainDone` records the hash on failure, removes it on success, and rate-limits the warning (first failure at warning level; repeats inside the window at debug level).

`src/Vehicle/TerrainProtocolHandler.cc`

- `_sendTerrainData` now clears the mask bit on `error` as well as on success, so the handler stops looping on a known-failed tile. The vehicle re-issues `TERRAIN_REQUEST` if it still needs the data, which then gets a fresh fetch once the backoff window expires.
- Demote the per-call warning to debug since the underlying cause is already logged by `TerrainTileManager`.

### DIFF
--- a/src/Terrain/TerrainTileManager.cc
+++ b/src/Terrain/TerrainTileManager.cc
@@ -10,6 +10,7 @@
 #include "QGCLoggingCategory.h"
 #include "QGCGeo.h"
 
+#include <QtCore/QDateTime>
 #include <QtLocation/private/qgeotilespec_p.h>
 #include <QtNetwork/QNetworkAccessManager>
 #include <QtNetwork/QNetworkRequest>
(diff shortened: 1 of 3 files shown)


dakejahl:
Corresponding Ardupilot PR which fixes the root cause of the issue
https://github.com/ArduPilot/ardupilot/pull/33135

PR #14415, opened 2026-05-20
feat(ParameterEditor): allow Advanced UI override of read-only params

dakejahl:
## Summary

Lets Advanced-UI users override the read-only lock on a parameter, mirroring the existing "Force save" escape hatch for min/max bounds.

## Problem

QGC respects the `readOnly` metadata flag (#14007) by hiding the input field entirely, with no way to override it — even for diagnostic or recovery cases where the lock is too strict, such as unjamming ArduPilot (ArduPilot/ardupilot#33136).

## Solution

In `ParameterEditorDialog`, read-only params get a "Force edit read-only param" checkbox nested under the existing "Advanced settings" checkbox — hidden by default, requiring a deliberate opt-in. Ticking it switches to the normal edit view (input + Save); unticking "Advanced settings" re-locks the param. UI-only change, no firmware/C++ gating.

### DIFF
--- a/src/QmlControls/ParameterEditorDialog.qml
+++ b/src/QmlControls/ParameterEditorDialog.qml
@@ -26,6 +26,7 @@ QGCPopupDialog {
     property bool   _allowDefaultReset:         fact.defaultValueAvailable
     property bool   _showCombo:                 fact.enumStrings.length !== 0 && fact.bitmaskStrings.length === 0 && !validate
     property bool   _readOnlyDisplay:           fact.readOnly && !forceEdit.checked
+    property bool   _allowForceEdit:            fact.readOnly && _allowForceSave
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
This seems very strange. Is it read only or isn't it? The firmware will allow this?

DonLakeFlyer:
Can you provide some screenshots?

dakejahl:
This is the only way to unjam Ardupilot when this issue occurs https://github.com/ArduPilot/ardupilot/pull/33136

image
image

DonLakeFlyer:
It's not hidden behind the advanced checkbox? If not, this can't just be out in the open.

dakejahl:
@DonLakeFlyer is this what you're suggesting?
image

PR #14430, opened 2026-05-22
fix(MAVLink/Signing): wall-clock timestamp refresh, re-sign at send, drop qtkeychain keystore

### LINKED ISSUE #14375: MAVLink signing: GCS fails to connect when SITL is already running

## Expected behaviour

QGC should successfully establish a signed MAVLink link regardless of whether the SITL was started before or after QGC. Signed message timestamps must reflect the current time at the moment of signing.

## Current behaviour

All outbound commands are silently rejected by the SITL due to stale signing timestamps. The log shows repeated `Giving up sending command after max retries` for every command in the initialisation sequence, and the state machine eventually gives up.

## Steps to reproduce

1. Start ArduPilot SITL with MAVLink signing enabled and a key configured.
2. Wait at least 2–3 minutes.
3. Start QGroundControl with the matching signing key.
4. Observe: QGC detects the vehicle but all initialisation commands time out. No telemetry stream is established.
5. Capture UDP traffic with Wireshark — signed packets from QGC show a timestamp ~3 minutes behind wall-clock time.

## Describe the bug

When QGroundControl starts *after* an ArduPilot SITL that is already configured with a MAVLink signing key, no link is established. The vehicle appears connected at the UDP/link level but the initialisation state machine exhausts all retries for every command (`MAV_CMD_SET_MESSAGE_INTERVAL`, `MAV_CMD_REQUEST_MESSAGE` for `AUTOPILOT_VERSION`, `AVAILABLE_MODES`, `COMPONENT_METADATA`, etc.) without receiving any acknowledgement.

## Root cause (observed)

Wireshark capture of the UDP traffic shows that `REQUEST_DATA_STREAM` messages signed by QGC carry a MAVLink timestamp approximately **3 minutes in the past**. The SITL's replay-protection window rejects any signed packet whose timestamp is not within a small delta of its own clock. Because every outbound message is rejected, the vehicle never ACKs any command and the initialisation state machine times out.

## Likely cause

The signing timestamp appears to be initialised once at application start (or from a persisted/cached value) and is not refreshed when a link is actually established. The ~3-minute offset closely matches the time elapsed between QGC startup and first successful UDP contact with an already-running SITL. The timestamp used for signing should be taken from the current monotonic or wall-clock time at the moment each message is signed, not from a value frozen at startup.

### Platform

Linux

### Flight Stack

ArduPilot

### System Information

| | |
|---|---|
| QGC build | Debug (custom build from source) |
| Qt / FFmpeg | Qt multimedia with FFmpeg 7.1.3 |
| SITL | ArduPilot SITL with signing key pre-configured |
| Transport | UDP |
| OS | Linux (Ubuntu) |

## Log excerpt

```
4.672  Warning: Giving up sending command after max retries: MAV_CMD_SET_MESSAGE_INTERVAL
5.672  Warning: Giving up sending command after max retries: MAV_CMD_REQUEST_MESSAGE requesting: AUTOPILOT_VERSION
8.172  Warning: InitialConnectStateMachine:RequestAutopilotVersion — Max retries exhausted after timeout
10.672 Warning: Giving up sending command after max retries: MAV_CMD_REQUEST_MESSAGE requesting: AVAILABLE_MODES
13.244 Warning: Giving up sending command after max retries: MAV_CMD_REQUEST_MESSAGE requesting: COMPONENT_METADATA
15.673 Warning: COMP_METADATA_TYPE_GENERAL: failed to load metadata (primary and fallback)
```

Pcap captured
[sitl-signed.zip](https://github.com/user-attachments/files/27754112/sitl-signed.zip)

HTRamsey:
## Summary

Hardens MAVLink signing so QGC reliably connects to an already-running, signing-enabled vehicle, and simplifies key storage. Fixes #14375 (GCS fails to connect when SITL is already running — `OLD_TIMESTAMP` rejection).

## Root cause

`SigningChannel::init()` seeds `_signing.timestamp` from wall clock once. After that, libmavlink's `mavlink_sign_packet` only advances the value by `+1` per outbound packet — it does not track real time. Two paths then drift behind wall clock and get rejected by peers that pin `signing.timestamp` to wall clock (ArduPilot's `GCS_Signing::update_signing_timestamp`):

1. **Idle outbound** — QGC starts, sits idle detecting the vehicle, then sends its first command minutes later. The wire timestamp is still near the `init()` value (~3-minute offset in #14375).
2. **Cached resend** — `Vehicle::sendMessageMultiple` caches the *already-signed* bytes and re-emits them across retries, so the frozen signature timestamp drifts behind real time.

## Fix

- **Wall-clock refresh** — `SigningChannel::refreshOutgoingTimestamp()` bumps `_signing.timestamp` up to current wall clock under the write lock (no-op when disabled or already ahead). `SigningController` drives it from a 1 Hz `QTimer`, well under the receiver's 6 s `MAVLINK_SIGNING_TIMESTAMP_LIMIT`. The timer is now **gated on channel-signing state** — it runs only while a channel is actually signing, not for every link's lifetime.
- **Re-sign at send** — `SigningChannel::signOutgoing()` re-signs a cached outgoing message in place with a current, monotonic timestamp immediately before it hits the wire, fixing the cached-resend drift. The secret key never leaves the signing layer.
- **Keystore simplification** — drops `QGCKeychain` (qtkeychain) and the `QLockFile` cross-process guard; keys persist via `QSettings` with an explicit `sync()`. Removes the qtkeychain dependency and its install plumbing.
- **FSM refactor** — `VehicleSigningController` splits handler wiring from transmit (`_wireConfirmHandlers` once per op, `_sendAndStartRetransmit` for the wire), so the retransmit path can't double-deliver the one-shot confirm/fail connections.

## Test plan

- [x] `SigningTest` — new regressions: `_testRefreshOutgoingTimestamp` (3-minute stall catch-up, no-op when ahead/disabled), `_testSignOutgoingRefreshesCachedTimestamp` (cached-resend re-sign stays monotonic and still verifies), `_testKeyStorePersistRoundTrip` (QSettings persist/reload/remove).
- [x] `SigningControllerTest` — new: `_testWallClockTimerRefreshesAfterEnable` / `_testWallClockTimerStoppedWhenIdle` (timer gating). Burst-alert / auto-detect / state-machine unchanged.
- [x] `MockLinkSigningTest` — new: `_testEnableDisableReEnableCycle` (handlers re-wire per op; no spurious failure). Enable/disable FSM unchanged.
- Added a test-only `SigningController::setTimeoutForTesting()` seam so timeout-path tests no longer pay the production 5 s wait (suite runtime 35 s → 17 s).
- [ ] Manual: ArduPilot SITL with signing key, start SITL → wait 3 min → start QGC → confirm initialisation commands succeed.


TomZanna:
Hi, thanks for working on this! I've been testing this draft for a week, but the bug is still present.
I ran some Wireshark captures and noticed that connection works fine for the first few seconds of startup. After that, a timestamp drift appears between the SITL's HEARTBEAT and QGC's REQUEST_DATA_STREAM.
Since the fix introduces a 1 Hz timer and locks, could this be a race condition? For example, two different threads communicating with the same socket but not sharing or updating the timestamp correctly?

Tested on build ID 81416f0 timestamp 2026-05-22T16:29:33-04:00 (artifact of this PR's action)

TomZanna:
Just tested the new commit. The bug seems fixed!

Issue #10, opened 2012-01-27
Horizontal Situation Indicator plots local waypoints


Susurrus:
I would expect that the HSI would plot the remote waypoints. I first noticed that it didn't when it wasn't highlighting the current waypoint according to my MAV, but according to what was first retrieved from the controller. Then I noticed as I deleted missions locally they were removed from the HSI. Is there a reason why this widget doesn't follow the remote waypoints and instead follows the local ones? It also contrasts with how waypoints can be created by looking on the map as any locally-reference missions are not displayed on the HSI or the map if they're created manually only if they're pulled from the remote MAV.

PR #14433, opened 2026-05-24
FlyView: display Catmull-Rom curve for spline waypoints

leonfliesthings-dev:
## Summary

QGC draws straight lines between `NAV_SPLINE_WAYPOINT` (cmd 82) mission items, even though ArduPilot flies smooth Hermite/Catmull-Rom curves through them. This makes it difficult for operators to visualize the actual flight path when using spline waypoints.

This PR adds a cyan Catmull-Rom interpolation overlay on the fly map that shows the approximate spline curve between consecutive spline waypoints.

## How it works

- A `MapPolyline` in `FlyViewMap.qml` polls the mission controller's `visualItems` every 3 seconds
- When spline waypoints (cmd 82) are detected, it computes Catmull-Rom interpolation between consecutive waypoints
- The curve is drawn as a cyan overlay on the map (3px wide, 85% opacity)
- Only renders when spline waypoints are present — no effect on regular waypoint missions
- Minimal performance impact: computation only runs when the mission item count changes

## Screenshots

The cyan line shows the approximate spline curve ArduPilot will fly, overlaid on the straight-line mission path display.

## Testing

- Loaded mission with `NAV_SPLINE_WAYPOINT` items in SITL
- Cyan curve appears on the fly map showing smooth paths between waypoints
- No curve shown for missions using only `NAV_WAYPOINT` (cmd 16)
- No performance impact observed

### DIFF
--- a/src/FlyView/FlyViewMap.qml
+++ b/src/FlyView/FlyViewMap.qml
@@ -259,6 +259,73 @@ FlightMap {
         }
     }
 
+    // ── Spline Waypoint Curve Display ─────────────────────────────────
+    // ArduPilot's NAV_SPLINE_WAYPOINT (cmd 82) flies smooth Hermite curves
+    // between waypoints, but QGC only draws straight lines. This adds a
+    // Catmull-Rom interpolation overlay showing the approximate spline curve.
+    MapPolyline {
+        id:         splineCurve
+        line.width: 3
+        line.color: "cyan"
+        opacity:    0.85
+        z:          QGroundControl.zOrderTrajectoryLines + 1
+        visible:    !pipMode && path.length > 0
+    }
+
+    property var _splineMissionController: _planMasterController ? _planMasterController.missionController : null
+
+    Timer {
+        interval: 3000; running: _root.visible; repeat: true
+        property int _prevCount: 0
+        onTriggered: {
+            try {
+                if (!_splineMissionController) return
+                var items = _splineMissionController.visualItems
+                if (!items || items.count < 2) return
+                if (items.count === _prevCount) return
+                _prevCount = items.count
+
+                // Extract waypoints that have coordinates, check for cmd 82
+                var wps = []
+                var hasSpline = false
+                for (var i = 0; i < items.count; i++) {
+                    var item = items.get(i)
+                    if (!item) continue
+                    if (typeof item.specifiesCoordinate === "undefined" || !item.specifiesCoordinate) continue
+                    var coord = item.coordinate
+                    if (!coord || !coord.isValid) continue
+                    var cmd = typeof item.command !== "undefined" ? item.command : -1
+                    wps.push({lat: coord.latitude, lon: coord.longitude, cmd: cmd})
+                    if (cmd === 82) hasSpline = true
+                }
+
+                if (!hasSpline || wps.length < 2) { splineCurve.path = []; return }
+
+                // Catmull-Rom interpolation between consecutive waypoints
+                var curve = []
+                var N = 10  // samples per segment
+                for (var s = 0; s < wps.length - 1; s++) {
+                    var p0 = wps[Math.max(0, s - 1)]
+                    var p1 = wps[s]
+                    var p2 = wps[s + 1]
+                    var p3 = wps[Math.min(wps.length - 1, s + 2)]
+                    for (var k = 0; k < N; k++) {
+                        var t = k / N
+                        var t2 = t * t, t3 = t2 * t
+                        var lat = 0.5 * ((2*p1.lat) + (-p0.lat+p2.lat)*t + (2*p0.lat-5*p1.lat+4*p2.lat-p3.lat)*t2 + (-p0.lat+3*p1.lat-3*p2.lat+p3.lat)*t3)
+                        var lon = 0.5 * ((2*p1.lon) + (-p0.lon+p2.lon)*t + (2*p0.lon-5*p1.lon+4*p2.lon-p3.lon)*t2 + (-p0.lon+3*p1.lon-3*p2.lon+p3.lon)*t3)
+                        curve.push(QtPositioning.coordinate(lat, lon))
+                    }
+                }
+                var last = wps[wps.length - 1]
+                curve.push(QtPositioning.coordinate(last.lat, last.lon))
+                splineCurve.path = curve
+            } catch(e) {
+                splineCurve.path = []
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I'll try to get this into 5.1 but if it is too invasive it will need to wait till 5.2 so it has some bake time

leonfliesthings-dev:
> I'll try to get this into 5.1 but if it is too invasive it will need to wait till 5.2 so it has some bake time

Thanks, DonLakeFlyer. If it's too much, let me know; I can make some adjustments. I have a new version that colour-ramps altitude on the splines, which gives good alt context to the flight paths.

DonLakeFlyer:
Here is what Claude-Fable has to say about this change. For now I'm moving this to 5.2 since it requires some significant cleanup I think:
# Code Review: `leonfliesthings-dev-feature/spline-curve-display`

**Branch:** `leonfliesthings-dev-feature/spline-curve-display`
**Changes:** One feature commit (`0946f8b37`) plus a merge commit (`df900d155`) — 67 lines added to `src/FlyView/FlyViewMap.qml`
**Feature:** Catmull-Rom curve overlay in Fly view approximating ArduPilot `NAV_SPLINE_WAYPOINT` (MAV_CMD 82) flight paths.

## Correctness

1. **Curve is drawn through every waypoint if *any* item is a spline waypoint** — `FlyViewMap.qml#L299` sets `hasSpline` globally, then the loop at `FlyViewMap.qml#L307-L320` splines through *all* coordinate items. ArduPilot only flies a spline between `NAV_SPLINE_WAYPOINT` items; regular waypoints in the same mission are flown as straight legs. In a mixed mission the overlay draws a curved path over segments the vehicle will fly straight — actively misleading. The interpolation must be gated per-segment (segment `p1→p2` is curved only when the relevant items are cmd 82).

2. **Stale curve never cleared when the mission shrinks or is removed** — the early return at `FlyViewMap.qml#L284` (`items.count < 2`) exits *without* clearing `splineCurve.path`. Remove the mission (or switch vehicles) and the old cyan curve stays on the map indefinitely.

3. **Change detection by item count only** — `FlyViewMap.qml#L285-L286`. Dragging a waypoint, editing its coordinate, or loading a *different* plan with the same number of items leaves the curve stale/wrong. Count is not a valid proxy for "mission changed".

4. **Planned-home is included in the curve** — `visualItems` item 0 (MissionSettingsItem/home) reports `specifiesCoordinate == true` once home is set, so home becomes a Catmull-Rom control point and distorts the first segment's tangent. The vehicle never splines through home.

5. **Longitude interpolated as a raw scalar** — `FlyViewMap.qml#L317`. No antimeridian handling: waypoints straddling ±180° produce a curve wrapping across the entire globe. (High-latitude lat/lon-space interpolation inaccuracy is probably acceptable for a visual aid, but the wrap is a real bug.)

6. **Silent `try/catch` swallowing everything** — `FlyViewMap.qml#L324-L326`. Any logic error in this block is invisible forever. QGC code doesn't use blanket catch; remove it and let qmllint/console surface problems.

## Performance (map display)

7. **Polling timer instead of signals** — `FlyViewMap.qml#L277-L278`. A 3-second repeating timer wakes the UI event loop for the life of the Fly view even when nothing changes, and adds up to 3s of latency after a mission load. `MissionController` already emits `visualItemsReset` (see `src/MissionManager/MissionController.h#L52`) — use a `Connections` on the controller instead. This would also fix issues 2 and 3.

8. **The PIP map instance runs the timer too** — `running: _root.visible` is true for the pip-mode map instance, so a second copy of this computation runs every 3s and builds a path that `visible: !pipMode` then hides. Gate with `running: _root.visible && !pipMode`.

9. **Full path reassignment with dense sampling** — 10 samples/segment for *all* segments means a large mission (hundreds of items) produces thousands of `QGeoCoordinate` JS objects plus a full `MapPolyline` geometry re-tessellation on every update, and every subsequent pan/zoom pays per-point re-projection cost for the polyline. Fixing issue 1 (spline segments only) shrinks this dramatically; also consider building the points as a plain array once and skipping regeneration when the plan is unchanged (signal-driven, per #7).

10. **This belongs in C++, not a QML overlay** — the mission line geometry is already computed centrally (`simpleFlightPathSegments` / `FlightPathSegment`, `src/MissionManager/MissionController.h#L60`). Doing spline rendering there would: apply to Plan view too (currently the Plan view still shows straight lines for the same mission — inconsistent UX), reuse existing change notification, and keep the interpolation out of the JS engine. If a QML-side prototype is intended as a first step, that tradeoff should at least be stated in the PR.

## Style / conventions

11. Multiple statements per line with semicolons (`interval: 3000; running: ...; repeat: true`, and `{ splineCurve.path = []; return }`) — `FlyViewMap.qml#L278` and `FlyViewMap.qml#L302`. One property/statement per line, and no brace-less/single-line `if`s (`FlyViewMap.qml#L282-L285`).

12. `property var _splineMissionController` declared mid-file between child items — `FlyViewMap.qml#L275`. Private properties belong at the top of the root item with the other `_`-prefixed properties.

13. Magic number `82` — `FlyViewMap.qml#L297-L299`. Should reference the MAV_CMD enum value symbolically (or at minimum a named `readonly property int _mavCmdNavSplineWaypoint`).

14. Hardcoded `"cyan"` — `FlyViewMap.qml#L269`. There's precedent for hardcoded map-line colors (trajectory red), but cyan on satellite imagery vs. the existing mission-line color scheme deserves a deliberate choice; consider matching the mission line palette with a dashed/derived style instead.

15. Branch history contains a merge commit (`df900d155`) on top of the single feature commit — should be rebased flat before PR.

## Summary

The idea is genuinely useful (QGC has never visualized `NAV_SPLINE_WAYPOINT` paths), but as written the overlay draws incorrect curves for mixed missions (#1), leaves stale geometry on screen (#2, #3), and uses a polling architecture that costs CPU and adds latency when signal-driven updates are readily available (#7). Items #1, #2, and #7 are the minimum fixes; the C++ `FlightPathSegment` approach (#10) is the proper home for this so Plan view benefits too.

DonLakeFlyer:
@leonfliesthings-dev Have you looked at the code review comments?

leonfliesthings-dev:
> @leonfliesthings-dev Have you looked at the code review comments?

I have, and they are on my list of things to do ...

DonLakeFlyer:
Update?

PR #14438, opened 2026-05-27
Fly View: add altitude support for Click to ROI

### LINKED ISSUE #14394: Fly View: Click to ROI should support altitude
Related to #11559

Daksh7112003:
## Description
Adds altitude support for Fly View "Click to ROI".

Previously, ROI commands only sent latitude and longitude coordinates,
causing the altitude field to remain unset/zero. This update includes
altitude in the ROI coordinate sent through MAVLink.

This improves camera/gimbal targeting accuracy when using Click to ROI.

Fixes #14394.

## Type of Change

## Testing
- [x] Tested locally

### Platforms Tested
- [x] Windows

### Flight Stacks Tested
- [x] ArduPilot

## Screenshots
N/A

## Checklist
- [x] I have read the Contribution Guidelines
- [x] I have read the Code of Conduct

## Related Issues
#14394

### DIFF
--- a/src/FlyView/FlyViewMap.qml
+++ b/src/FlyView/FlyViewMap.qml
@@ -709,7 +709,7 @@ FlightMap {
                         visible:            globals.guidedControllerFlyView.showROI
                         onClicked: {
                             mapClickDropPanel.close()
-                            globals.guidedControllerFlyView.executeAction(globals.guidedControllerFlyView.actionROI, mapClickCoord, 0, false)
+                            globals.guidedControllerFlyView.confirmAction(globals.guidedControllerFlyView.actionROI, mapClickCoord)
                         }
                     }
 
--- a/src/FlyView/GuidedActionsController.qml
+++ b/src/FlyView/GuidedActionsController.qml
@@ -225,7 +225,7 @@ Item {
             } else {
                 console.error("setupSlider called for inapproproate change speed action", _vehicleInFwdFlight, _activeVehicle.haveMRSpeedLimits)
             }
-        } else if (actionCode === actionChangeAlt || actionCode === actionOrbit || actionCode === actionGoto || actionCode === actionPause) {
+        } else if (actionCode === actionChangeAlt || actionCode === actionOrbit || actionCode === actionROI || actionCode === actionGoto || actionCode === actionPause) {
             guidedValueSlider.setupSlider(
                 GuidedValueSlider.SliderType.Altitude,
                 _flyViewSettings.guidedMinimumAltitude.value,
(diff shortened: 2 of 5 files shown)


DonLakeFlyer:
Due to Claude-Fable review needing to be addressed with multiple problems this is going to have to get cleaned up and wait till we are ready to start taking merges for 5.2

DonLakeFlyer:
> Due to Claude-Fable review needing to be addressed with multiple problems this is going to have to get cleaned up and wait till we are ready to start taking merges for 5.2

This comment was wrong. Looking at the change now.

DonLakeFlyer:
Thanks for taking this on — click-to-ROI altitude support is a real gap (#14394) and the direction here is right. I tested the branch end to end and reviewed the code, and unfortunately there are a few structural problems that keep the feature from working:

1. **The altitude slider never appears.** `setupSlider()` only configures the slider; each action case in `confirmAction()` must also explicitly set `guidedValueSlider.visible = true` (compare `actionChangeAlt`, `actionOrbit`, `actionPause`). The `actionROI` case doesn't, so the slider stays hidden. `GuidedActionConfirm.qml` then passes `sliderOutputValue = 0` when the slider isn't visible, which means `valueInMeters` is always `0` — the `isNaN` fallback is never taken, and the vehicle always receives altitude 0. Net behavior is identical to master, but with an extra confirmation step added for the user.

2. **Overloading `coordinate.altitude()` as "relative altitude" breaks other callers.** `Vehicle::guidedModeROI()` is also called from `GimbalIndicator.qml` with `activeVehicle.homePosition` (whose altitude is **AMSL**) and from the Edit ROI Position dialog. With the new `roiWithTerrain(centerCoord, centerCoord.altitude())`, PX4 now computes `terrain + homeAMSL` — e.g. terrain + 1400 m in mountainous areas. The relative altitude needs to be an explicit parameter rather than smuggled inside the coordinate, so "no altitude", "AMSL", and "relative" can't be confused.

3. **The altitude reference differs per firmware.** With the slider working, the same value would mean "above terrain at the ROI point" on PX4 (terrain query + offset) but "above home" on ArduPilot (`MAV_FRAME_GLOBAL_RELATIVE_ALT`). In steep terrain these diverge significantly, and the "Alt (rel)" label doesn't say which reference applies. The slider default (`altitudeRelative.value`, the vehicle's current altitude) is also an odd starting point for a ground-target ROI.

4. Minor: QGC QML style requires `let`/`const` instead of `var` for new code.

Given that the fixes needed are mostly design-level (explicit altitude parameter, defined altitude semantics across firmwares, slider wiring), I'm going to put together a replacement PR that implements the feature on top of a cleaned-up `guidedModeROI` API — likely defining the slider value as "meters above home" for both firmwares, which also avoids the terrain-query latency on PX4. I'll credit this PR and #14394 there. Thanks again for surfacing the issue and the groundwork here.

Issue #12881, opened 2025-05-20
Please offer compatibility with classic .params files, and ability to SAVE as classic file by default


AndKe:
### Feature Description

the QGC format for .params is not compatible with other GCS's , and hard to compare(diff)
readability is not helped by lines like this:

1	1	ATC_RAT_YAW_FLTD	0.000000000000000000	9
1	1	ATC_RAT_YAW_FLTE	1.000000000000000000	9
1	1	ATC_RAT_YAW_FLTT	20.000000000000000000	9
1	1	ATC_RAT_YAW_I	0.091200001537799835	9
1	1	ATC_RAT_YAW_IMAX	0.500000000000000000	9
1	1	ATC_RAT_YAW_P	0.912000000476837158	9
1	1	AUTO_OPTIONS	0	6

This many decimals is .. not needed and just bad for comparing/diff/version tracking
for exaxmple is clearly set to be: 
ATC_RAT_YAW_I	0.0912

### Flight Stacks

ArduPilot, PX4

### Vehicle Types

Multirotor, Fixed-wing, VTOL, Submarine

DonLakeFlyer:
So this was changed 9 years ago: https://github.com/mavlink/qgroundcontrol/issues/2951. The rational behind it which is to save in the file the actual value with full precision I think still makes sense.

AndKe:
The full precision is an illusion :) - as you see/know that the last digits were not actually entered, so in most cases the later decimals is just an error caused by the binary approximation due to limited mantissa bits of the fraction.  If we had actual interest in precision that far out we would store the values as decimal.Decimal - so I do understand the idea of storing "everything" - but there are no parameters that I can think of where this is useful.

There are additional reasons for supporting the classic .param (regardless of number of digits):

APMPlanner 2 and mavproxy can compare .param with actual settings /  this is useful for configuration check, or copying parts of the configuration between vehicles, or checking configuration against earlier "known good" state.
- for that one needs to have a compatible format.
It would be nice is QGC was one of the tools that could at least export to such file - and preferably read such files.

The simpler .param format is also the one "universal" format across GCS software.  It is odd that QGC is the only GCS that is not compatible with it. :)

DonLakeFlyer:
I thought it was just decimal places were the problem. Not sure where QGC's param format differs. Can you detail? 

Also FYI: QGC 5.0 supports showing you the diff between vehicle and param files when you load and allows you to selectively include/exclude.

[...]

PR #14457, opened 2026-06-03
ParameterEditor: fix MP .param import, fix Fact float precision formatting

DonLakeFlyer:
Fixes #14451, related to #12881

## ParameterEditorController

- Support Mission Planner 2-column `.param` file format (`Name Value`) in `buildDiffFromFile` in addition to the existing QGC 5-column format
- Fix delimiter splitting to handle tab, space, and comma separators so MP files parse correctly
- Track parameters from MP files that are not present on the vehicle and emit `missingParamsFromFile` signal so the UI can inform the user
- Clear `_diffMissingParams` in `clearDiff()`

## ParameterEditor.qml

- Fix `onMissingParamsFromFile` handler: replace invalid QML syntax `QGC::showAppMessage(...)` with `QGroundControl.showMessageDialog(...)`
- Display each missing parameter name on its own line instead of a single comma-separated line that does not wrap

## Fact.cc

- Replace Qt float formatting loop (`QString::number` / `.toFloat()`) with std equivalents: `snprintf` / `strtof` / `strcmp`
- Normalise `-0` output via `strcmp` instead of `QString` comparison
- Increase `snprintf` buffer from 16 to 32 bytes (defensive against 3-digit exponent renderers on older MSVC runtimes)
- Move std system headers before project headers per convention

## Tests

- Add unit tests for `rawValueStringFullPrecision()` (float and double) using `strtof`/`strtod` round-trip assertions
- Add unit tests for `ParameterEditorController` MP format parsing and `missingParamsFromFile` signal

### DIFF
--- a/src/FactSystem/Fact.cc
+++ b/src/FactSystem/Fact.cc
@@ -1,3 +1,9 @@
+#include <cmath>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <limits>
+
 #include "Fact.h"
 #include "FactValueSliderListModel.h"
 #include "AppMessages.h"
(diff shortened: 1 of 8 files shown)


DonLakeFlyer:
@AndKe Can you give this a try?

AndKe:
It works perfect, I tested with both apmplanner2 style, and mavproxy style files.
image

- please remove trailing zeros. 
- please enable both .param and .params by default
I see the option "
image
"
IDK what mission planner files are, AFAIK MP never had it's own standard, I would suggest selection between ".param*" and "all".

Thank you!

PR #14465, opened 2026-06-04
cmake: auto-detect/create workspace .venv for Python generators

### LINKED ISSUE #14449: build: Python build-time deps (scripts group) not documented or auto-installed
## Problem

Two related gaps in the build setup tooling:

### 1. Missing Python modules break the build silently

CMake runs Python code generators during the build (MAVLink enum headers, QML type registrations, etc.) that `import` packages from the `scripts` extras group in `tools/pyproject.toml` (`defusedxml`, `httpx`, `jinja2`).

If a developer runs `install_dependencies` (the documented setup step) and nothing else, these packages are **not** installed. The build then fails with a `ModuleNotFoundError` at configure/generate time, with no actionable guidance in the error message pointing back to a missing setup step.

### 2. Dev guide does not mention Python build-time dependencies

`docs/en/qgc-dev-guide/getting_started/index.md` lists `install_dependencies` as the only setup command. It makes no mention of `install_python.py` or the `scripts` group. A developer following the guide exactly will hit the above failure.

`tools/README.md` does reference `install_python.py`, but that file is developer-facing and not linked from the public dev guide.

## Suggested fixes

- **Option A (preferred):** Add a step to the dev guide to run `python3 tools/setup/install_python.py scripts` after `install_dependencies`, with a brief explanation that it installs Python packages required by the CMake build generators.
- **Option B:** Have `install_dependencies` automatically invoke `install_python.py scripts` as a final step, and document that this is what it does.
- In either case, improve the `ModuleNotFoundError` failure mode with a helpful message pointing to the correct setup step.

## Affected files

- `docs/en/qgc-dev-guide/getting_started/index.md`
- `tools/setup/install_dependencies/_cli.py`
- `tools/setup/install_python.py`
- `tools/pyproject.toml`

### LINKED ISSUE #14463: Dependencies:  defusedxml not included
please note that the need for "pip install defusedxml "  is not covered by dependency installing script.
```
......[1/705] Generating MAVLinkInstanceFields.h
FAILED: [code=1] MAVLinkInstanceFields.h /home/andre/prog/qgroundcontrol/build/MAVLinkInstanceFields.h 
cd /home/andre/prog/qgroundcontrol/build/src/MAVLink && /usr/bin/python3 /home/andre/prog/qgroundcontrol/tools/generators/mavlink_instance_fields.py /home/andre/prog/qgroundcontrol/.cache/CPM/mavlink/fa55/message_definitions/v1.0 all /home/andre/prog/qgroundcontrol/build/MAVLinkInstanceFields.h
Traceback (most recent call last):
  File "/home/andre/prog/qgroundcontrol/tools/generators/mavlink_instance_fields.py", line 13, in <module>
    import defusedxml.ElementTree as ET
ModuleNotFoundError: No module named 'defusedxml'
[9/705] Building CXX object src/Utilities/Geo/Formats/CMakeFiles/QGCGeoFormats.dir/GeoJsonHelper.cc.o
```

A decision is needed if is best to tell people top use --break-system-packages or add a venv for this.

HTRamsey:
Fixes #14449
Fixes #14463

## Problem

The MAVLink, AppSettings, and AutoPilotPlugins (PX4/APM) code generators run under `find_package(Python3 REQUIRED COMPONENTS Interpreter)`. With no hint, CMake resolves that to the **system** interpreter (`/usr/bin/python3`), which lacks the generator dependencies (`defusedxml`, `jinja2`, `httpx`) declared in `tools/pyproject.toml` under the `scripts` extra.

Only the `justfile` pins the right interpreter (`-DPython3_EXECUTABLE=$PWD/.venv/bin/python`). Configuring from an IDE (e.g. Qt Creator) or a raw `cmake` invocation skips that, so the build fails:

```
Traceback (most recent call last):
  File ".../tools/generators/mavlink_instance_fields.py", line 13, in <module>
    import defusedxml.ElementTree as ET
ModuleNotFoundError: No module named 'defusedxml'
```

## Fix

Add `cmake/modules/PythonVenv.cmake`, included right after `include(Toolchain)` and before `add_subdirectory(src)` so all five `find_package(Python3)` sites resolve consistently. It:

1. Respects an already-pinned `Python3_EXECUTABLE` (justfile/CI) — no-op.
2. Pins `<repo>/.venv` when it exists.
3. Otherwise, with `QGC_AUTO_PYTHON_VENV=ON` (default), bootstraps the venv via `tools/setup/install_python.py scripts`, then pins it.
4. With auto-create disabled and no venv, warns and falls back to system python.

This makes every configure path (IDE, raw cmake, `just`, CI) resolve the generator interpreter without requiring the `-DPython3_EXECUTABLE` flag. Opt out of auto-create with `-DQGC_AUTO_PYTHON_VENV=OFF`.

## Testing

- `cmake -P` parse-check of the module passes.
- Verified the `scripts` extra in `tools/pyproject.toml` provides `defusedxml`/`jinja2`/`httpx`.

### DIFF
--- a/.github/actions/cache-cleanup/action.yml
+++ b/.github/actions/cache-cleanup/action.yml
@@ -14,6 +14,18 @@ inputs:
     description: 'Write a markdown table of caches to $GITHUB_STEP_SUMMARY'
     required: false
     default: 'false'
+  prune:
+    description: 'GC mode: keep build caches, evict the rest when the pool nears the 10 GiB cap'
+    required: false
+    default: 'false'
+  high-water-mb:
+    description: 'Prune trigger: only evict when total cache size exceeds this'
+    required: false
+    default: '9000'
+  keep-mb:
+    description: 'Prune target: evict down to roughly this size, leaving headroom'
+    required: false
+    default: '6500'
 
 outputs:
   count:
@@ -45,4 +57,7 @@ runs:
         ARGS=(--branch "${{ inputs.branch }}")
         [[ "${{ inputs.dry-run }}" != "true" ]] && ARGS+=(--delete)
         [[ "${{ inputs.summary }}" == "true" ]] && ARGS+=(--summary)
+        if [[ "${{ inputs.prune }}" == "true" ]]; then
+          ARGS+=(--prune --high-water-mb "${{ inputs.high-water-mb }}" --keep-mb "${{ inputs.keep-mb }}")
+        fi
         python3 "${GITHUB_WORKSPACE}/.github/scripts/gh_cache_cleanup.py" "${ARGS[@]}"
--- a/.github/scripts/gh_cache_cleanup.py
+++ b/.github/scripts/gh_cache_cleanup.py
@@ -12,7 +12,9 @@ or workflow) is responsible for installation.
 from __future__ import annotations
 
 import argparse
+import json
 import os
+import re
 import sys
 from dataclasses import dataclass
 
@@ -20,7 +22,12 @@ from ci_bootstrap import ensure_tools_dir
 
 ensure_tools_dir(__file__)
 
-from common.gh_actions import gh, write_github_output, write_step_summary  # noqa: E402
+from common.gh_actions import gh, write_github_output, write_step_summary
+
+# Build caches are the expensive-to-rebuild data the GC must never evict; the
+# 10 GiB/repo pool is reclaimed from everything else first.
+DEFAULT_PROTECT = r"^(ccache|cpm-modules)-"
+_MIB = 1024 * 1024
 
 
 @dataclass(frozen=True)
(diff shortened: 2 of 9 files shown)


HTRamsey:
Addressed all three points from the Copilot review in `c8044703b`:

1. **Bootstrap interpreter version** — after `find_program(python3 python)`, the module now runs `python -c "sys.exit(0 if sys.version_info >= (3,12) else 1)"` and fails fast with an actionable message if the interpreter is older than 3.12 (matching `tools/pyproject.toml`'s `requires-python`). Kept `find_program` over `find_package(Python3)` deliberately so the bootstrap doesn't pollute the `Python3_*` cache the real generator `find_package` calls rely on.
2. **Surfaced install output** — `execute_process` now captures `OUTPUT_VARIABLE`/`ERROR_VARIABLE` and includes the uv/pip output in the `FATAL_ERROR`, matching the pattern in `GenerateSettingsQml.cmake`.
3. **Corrupt `.venv`** — added an explicit guard: if `<repo>/.venv` exists but the interpreter is missing, configure fails immediately with `rm -rf .venv && python tools/setup/install_python.py scripts`, instead of letting `install_python.py` silently early-return on the existing directory.

PR #14468, opened 2026-06-05
feat(LogViewer): UI polish, mobile guard, HH:MM:SS x-axis, start time / vehicle display

DonLakeFlyer:
## Summary

A batch of UI polish improvements to the Log Viewer, plus a mobile guard to prevent OOM crashes.

### Mobile guard
- Log Viewer is excluded on Android and iOS via `#if !defined(Q_OS_ANDROID) && !defined(Q_OS_IOS)` in `analyzePages()`. Parsing large log files (e.g. 900 MB ULog with 1000+ fields) exhausts the mobile heap. Proper mobile support requires time-bucketed downsampling and is deferred to a future major release.

### Toolbar improvements
- Filename only (not full path) is displayed in the top toolbar
- Log start time is shown using the system locale's short datetime format (correct date order, AM/PM, 12h/24h)
- Detected vehicle type (`ArduCopter`, `ArduPlane`, `PX4`, etc.) is shown in the toolbar
- Detected vehicle type removed from the fields panel (now lives in the toolbar)

### X-axis tick labels
- Tick labels now show `HH:MM:SS` / `M:SS` / `SSs` for elapsed time, or locale-formatted local wall-clock time (`h:mm:ss AP` / `HH:mm:ss`) when local time is selected
- Implemented via `AbstractAxis.labelDelegate` (a proper QML Component), replacing the `printf`-only `labelFormat`
- **Elapsed / Local time** radio buttons added as the second row in the fields panel; only visible when the log has a valid start time (hidden for `.tlog` and logs without timestamps)

### Bug fixes
- **Cursor drag**: cursor position now updates while dragging (non-shift mouse press was not moving the cursor on `onPositionChanged`)
- Removed mobile-only `PinchHandler` dead code from `LogViewerChart`

### DIFF
--- a/src/API/QGCCorePlugin.cc
+++ b/src/API/QGCCorePlugin.cc
@@ -72,11 +72,19 @@ QGCCorePlugin *QGCCorePlugin::instance()
 
 const QVariantList &QGCCorePlugin::analyzePages()
 {
+    // Log Viewer is excluded on mobile (Android/iOS) because parsing large log files
+    // (e.g. 900 MB ULog files with 1000+ fields) exhausts the mobile heap, causing
+    // OOM crashes. Proper mobile support requires time-bucketed downsampling and will
+    // be addressed in a future major release.
+#if defined(Q_OS_ANDROID) || defined(Q_OS_IOS)
+    static const QVariantList analyzeList = {
+#else
     static const QVariantList analyzeList = {
         QVariant::fromValue(new QmlComponentInfo(
             tr("Log Viewer"),
             QUrl::fromUserInput(QStringLiteral("qrc:/qml/QGroundControl/AnalyzeView/LogViewer/LogViewerPage.qml")),
             QUrl::fromUserInput(QStringLiteral("qrc:/qmlimages/MAVLinkInspector.svg")))),
+#endif
         QVariant::fromValue(new QmlComponentInfo(
             tr("Onboard Logs"),
             QUrl::fromUserInput(QStringLiteral("qrc:/qml/QGroundControl/AnalyzeView/OnboardLogs/OnboardLogPage.qml")),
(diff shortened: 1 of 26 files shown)


DonLakeFlyer:
@AhmWael I'm not done with this but give this a try:
* I removed hover cursor stuff
  * It was just too chaotic with respect to going back/forth when selecting fields
  * Also not enough room for two cursors worth of data on smaller screens
* Add support for both elapsed and local time as x axis. Defatuls to local. I'll add a setting to persist selection in ongoing work
* Cursor info show both local and elapsed time at position
  *  Seconds display of elapsed time goes into decimal places depending on resolution of data being shown

Probably other things I've forgotten about. I did a lot of reorganizing.

I've tested this on .ulg files to verify that start time detection from gps works.

AhmWael:
@DonLakeFlyer 
Thanks for the great work on this. Here are a few thoughts and suggestions from reviewing the changes:

- Overall Look & Feel: First of all, I really like the cleaner look! It makes the interface feel much better.
- Loading Bar: For the new loading bar, would it be possible to make the filled part of the progress bar blue? I think that would look cleaner and offer better visual feedback.
- Flight Mode Colors: I noticed that sometimes multiple modes are assigned the same color. This can get a bit confusing when analyzing logs; could we make sure each mode gets a distinct color?
- Map View Improvements:
    - Could we add a Reset Zoom button to the map view (similar to other views)?
    - Can we make the cursor draggable in the map view just like it is in the normal chart?
- Zoom/Interaction UX (Nice-to-have):
    - Using Shift + Click/Drag to zoom feels a little clunky.
    - As an alternative, maybe we could make it so that hovering in the vicinity of the cursor allows you to drag it, while clicking anywhere else moves the cursor to that spot. Then, clicking and dragging in an empty area could zoom in. I think this would be much more intuitive and user-friendly.

PR #14502, opened 2026-06-08
feat(FirmwareUpgrade): improve port selector UX

DonLakeFlyer:
## Summary

Improves the port selector UX on the Firmware Upgrade page. The previous UI showed a plain combo box with raw port paths; this change makes the selection smarter and the labels more readable.

The main goal of this PR is to not expose users to port names unless truly needed. For Mac/Windows users who fly a single vehicle on a SiK Radio or WiFi a port name is not something they would normally deal with or ever see.

Related to #14308

## Changes

- **Auto-preselect recognized board** — the first connected Pixhawk or SiK Radio is automatically selected in the combo on page open and whenever the port list changes.
- **Short label in closed combo** — when exactly one recognized device is connected, the combo header shows `Description [BoardName]` (e.g. `Cube Orange [Cube Pilot]`) without the port path. The full `port — description [name]` still appears in the open dropdown list.
- **Multiple devices** — when more than one recognized device is connected the full `displayName` (including port path) is shown so the user can distinguish them. A one-time warning is appended to the status log on the transition from ≤1 to >1 recognized devices.
- **Selection stability** — the current selection is only preserved if it is itself a recognized device. If an unrecognized port is selected and a Pixhawk is plugged in, the combo switches to the Pixhawk automatically.
- **Popup stability** — the port list model is not rebuilt while the combo dropdown is open, preventing it from collapsing mid-browse. A refresh runs when the popup closes.
- **Found device message** includes the port name, e.g. `Found device: Pixhawk 5X (ttyACM0)`.
- **`_lastKnownCount` reset** on `setupPageCompleted()` so the multi-device warning fires correctly on each page visit.
- **`_preselectIndex()` returns `-1`** when no recognized device is present instead of silently selecting the first unrecognized port.

### DIFF
--- a/src/Vehicle/VehicleSetup/FirmwareUpgrade.qml
+++ b/src/Vehicle/VehicleSetup/FirmwareUpgrade.qml
@@ -59,6 +59,20 @@ SetupPage {
             function setupPageCompleted() {
                 controller.startBoardSearch()
                 _defaultFirmwareIsPX4 = _defaultFirmwareFact.rawValue === _defaultFimwareTypePX4 // we don't want this to be bound and change as radios are selected
+                _lastKnownCount = 0
+            }
+
+            property int _lastKnownCount: 0
+
+            function _recognizedBoardCount() {
+                var ports = controller.availablePorts
+                var count = 0
+                for (var i = 0; i < ports.length; i++) {
+                    if (ports[i].boardType === _boardTypePixhawk || ports[i].boardType === _boardTypeSiKRadio) {
+                        count++
+                    }
+                }
+                return count
             }
 
             function _preselectIndex() {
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@julianoes Can you take a look at this

DonLakeFlyer:
It's not radically different in any way. Just enough to make me happy for stable. Then I'll take a larger whack at this in 5.2

julianoes:
I have this to try for tomorrow.

julianoes:
Works still for me, apart from the fact that the aws download times out, but that seems to be unrelated to QGC.

julianoes (approved):

DonLakeFlyer:
Thanks for testing Julian

PR #14519, opened 2026-06-11
fix: Request CAMERA_SETTINGS on zoom and focus

jnomikos:
## Description

The changes from PR #13612 do not actually end up requesting `CAMERA_SETTINGS`  whenever zoom level or focus changes. Thus, QGC may request `CAMERA_SETTINGS` (and thus CAMERA_FOV_STATUS) once, but it won't necessarily do so after that. This could result in the FOV stored on QGC not actually reflecting the FOV of the camera after zooming.

## Type of Change

- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [x] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [ ] ArduPilot

## Checklist

## Related Issues

---

### DIFF
--- a/src/Camera/VehicleCameraControl.cc
+++ b/src/Camera/VehicleCameraControl.cc
@@ -173,6 +173,7 @@ VehicleCameraControl::~VehicleCameraControl()
     _streamInfoTimer.stop();
     _streamStatusTimer.stop();
     _cameraSettingsTimer.stop();
+    _cameraSettingsRetryTimer.stop();
     _storageInfoTimer.stop();
 
     delete _netManager;
@@ -191,8 +192,11 @@ void VehicleCameraControl::_initWhenReady()
         _requestAllParameters();
     }
 
+    connect(&_cameraSettingsTimer, &QTimer::timeout, this, &VehicleCameraControl::_requestCameraSettings);
+    _cameraSettingsTimer.setSingleShot(true);
+
     QTimer::singleShot(500, this, &VehicleCameraControl::_requestCameraSettings);
-    connect(&_cameraSettingsTimer, &QTimer::timeout, this, &VehicleCameraControl::_cameraSettingsTimeout);
+    connect(&_cameraSettingsRetryTimer, &QTimer::timeout, this, &VehicleCameraControl::_cameraSettingsTimeout);
 
     QTimer::singleShot(1000, this, &VehicleCameraControl::_checkForVideoStreams);
 
@@ -591,6 +595,7 @@ void VehicleCameraControl::setZoomLevel(qreal level)
                 ZOOM_TYPE_RANGE,                        // Zoom type
                 static_cast<float>(level));             // Level
         }
+        _cameraSettingsTimer.start(1000);
     }
 }
 
@@ -608,6 +613,7 @@ void VehicleCameraControl::setFocusLevel(qreal level)
                 FOCUS_TYPE_RANGE,                       // Focus type
                 static_cast<float>(level));             // Level
         }
+        _cameraSettingsTimer.start(1000);
     }
 }
 
@@ -649,6 +655,8 @@ void VehicleCameraControl::stepZoom(int direction)
             false,                                  // ShowError
             ZOOM_TYPE_STEP,                         // Zoom type
             direction);                             // Direction (-1 wide, 1 tele)
+
+        _cameraSettingsTimer.start(1000);
     }
 }
 
@@ -675,6 +683,8 @@ void VehicleCameraControl::stopZoom()
             false,                                  // ShowError
             ZOOM_TYPE_CONTINUOUS,                   // Zoom type
             0);                                     // Direction (-1 wide, 1 tele)
+
+        _cameraSettingsTimer.start(1000);
     }
 }
 
@@ -688,6 +698,8 @@ void VehicleCameraControl::stepFocus(int direction)
             false,                                  // ShowError
             FOCUS_TYPE_STEP,                        // Focus type
             direction);                             // Direction (-1 in, 1 out)
+
+        _cameraSettingsTimer.start(1000);
     }
 }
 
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
Claude-Fable review. I believe most of this is correct and needs fixing:
# Code Review: PR #14519 — fix: Request CAMERA_SETTINGS on zoom and focus

**PR:** https://github.com/mavlink/qgroundcontrol/pull/14519
**Author:** jnomikos (John Nomikos)
**Files changed:** `src/Camera/VehicleCameraControl.cc` (+20/−6), `src/Camera/VehicleCameraControl.h` (+1/−0)

## Does it fix a real problem? Yes.

The FOV-update chain from #13612 is: receive `CAMERA_SETTINGS` → `QGCCameraManager::_handleCameraSettings()` → `requestCameraFovForComp()`. But nothing on master ever *re-requests* `CAMERA_SETTINGS` after a zoom/focus command — `setZoomLevel()`, `stepZoom()`, etc. just send `MAV_CMD_SET_CAMERA_ZOOM` and stop. The only request happens 500 ms after init (`_initWhenReady()`). The MAVLink camera protocol doesn't require cameras to broadcast `CAMERA_SETTINGS` on change, so with a camera that doesn't, QGC's zoom level / FOV goes stale after the first zoom. The gap is real and the general approach (debounced re-request after zoom/focus commands) is reasonable.

## Review Items

1. **Confusing timer repurposing.** The PR renames the *retry* role of the existing `_cameraSettingsTimer` to a new `_cameraSettingsRetryTimer`, then reuses the old name `_cameraSettingsTimer` for a brand-new debounce/request-delay timer. Anyone diffing history or reading logs will misread this. The new timer should get the new name — e.g. `_cameraSettingsRequestDebounceTimer` — and `_cameraSettingsTimer` should keep its existing retry meaning. This also keeps the sibling patterns (`_storageInfoTimer`, `_streamStatusTimer` = retry timers) consistent.

2. **Retry counter is never reset for command-triggered requests.** `_cameraSettingsRetries` only resets in `handleCameraSettings()`. If the initial connect-time request cycle exhausts its 6 attempts (camera doesn't answer), every later zoom-triggered request gets zero retries — `_cameraSettingsTimeout()` sees `retries > 5` and bails on the first timeout. Also, a stale odd counter flips the `% 2` alternation in `_requestCameraSettings()` so zoom-triggered requests may lead with the deprecated legacy command. Consider resetting the counter when a fresh request cycle starts from a zoom/focus command (bounded, so no infinite spam).

3. **`startZoom()` / `startFocus()` are not covered.** The PR adds the timer to `stepZoom`, `stopZoom`, `stepFocus`, `stopFocus`, `setZoomLevel`, `setFocusLevel`, but not `startZoom()` / `startFocus()`. Relying on the matching `stop*` is usually fine, but if the stop command is lost or the camera hits its zoom limit and auto-stops, no refresh happens. Adding the same call there is cheap and makes the coverage complete.

4. **Cleaner single-point alternative.** Instead of six scattered `_cameraSettingsTimer.start(1000)` calls, the debounce could be started in one place — `_mavCommandResult()` on `MAV_RESULT_ACCEPTED` for `MAV_CMD_SET_CAMERA_ZOOM` / `MAV_CMD_SET_CAMERA_FOCUS`. That also avoids requesting settings when the camera NAKs the command. Worth suggesting; not a blocker.

5. **Single-shot config belongs in the constructor.** `setSingleShot(true)` is done in `_initWhenReady()`. If any zoom path runs before `_initWhenReady()` (or if a subclass skips it), the timer starts in repeating mode and — once connected — would fire `_requestCameraSettings` every second forever. Setting single-shot where the other timers are configured (constructor) removes the ordering hazard. Note the retry timer is also left repeating (default), which is pre-existing behavior and is masked by stop/restart in the handlers, but worth being explicit about.

6. **Inconsistent guard placement.** In `setZoomLevel`/`setFocusLevel` the `start(1000)` is placed outside the `if(_vehicle)` guard (inside `hasZoom()`/`hasFocus()`), so the timer starts even when no command was sent. Harmless since `_requestCameraSettings()` null-checks `_vehicle`, but it should sit next to the `sendMavCommand` it belongs to for consistency with the `stepZoom`/`stopZoom` placement.

7. **Magic number repeated six times.** The `1000` ms debounce is duplicated at every call site. A named constant (e.g. `static constexpr int kCameraSettingsRequestDelayMs = 1000`) would make the intent and tuning point obvious.

8. **Stale debug text.** The log in `_requestCameraSettings()` now prints `"timer active:" << _cameraSettingsRetryTimer.isActive()` while the message text still just says "timer" — fine, but with two similarly-named timers the messages should say *which* timer ("retry timer") to keep the logs debuggable.

9. **No tests.** The PR checklist admits no unit tests were added. This class isn't currently well covered, so not a blocker, but a test asserting "zoom command ⇒ CAMERA_SETTINGS request goes out within the debounce window" would lock in the fix that #13612 was missing.

10. **Destructor update is correct** — the new `_cameraSettingsRetryTimer.stop()` was properly added alongside the existing timer stops.

## Verdict

Real bug, right general idea; main asks are the timer naming swap (item 1), the retry-counter interaction (item 2), and coverage of `startZoom`/`startFocus` (item 3).

DonLakeFlyer:
@jnomikos If you want I can deal with all of this and create a replacement pull for this

jnomikos:
> @jnomikos If you want I can deal with all of this and create a replacement pull for this

Yep, I am kind of swamped. If you could, that would be excellent.

DonLakeFlyer:
No problem at all.

DonLakeFlyer:
Closing in favor of #14614, which is a superset/replacement of this fix (as discussed above). It keeps your ack-driven re-request of `CAMERA_SETTINGS` after zoom/focus changes and additionally addresses the review items (timer naming, retry-counter reset, `startZoom()`/`startFocus()` coverage, no refresh on NAK, single-shot config in the constructor) plus adds a unit test that locks the behavior in. Thanks again for the diagnosis and original fix @jnomikos!

PR #13966, opened 2026-02-10
Replace hardcoded PX4 firmware lookup with manifest-based discovery

mrpollo:
## Summary

- Replace the static `px4_board_name_map` (60+ hardcoded board ID → firmware name entries) and the GitHub Releases API version detection with dynamic firmware discovery from the PX4 manifest
- Add a user-selectable firmware build list for PX4 in the upgrade dialog, mirroring the existing ArduPilot combo box pattern
- Bootloader and canbootloader builds are filtered out from the user-facing list
- SHA-256 hash verification from the manifest is wired into firmware downloads

## Dependency

> **This PR depends on [PX4/PX4-Autopilot#25414](https://github.com/PX4/PX4-Autopilot/pull/25414)** which adds firmware manifest generation to PX4's build system and CI pipeline. That PR creates the manifest infrastructure (per-board metadata, per-release `release_manifest.json`, and a top-level `index.json`) that QGC consumes here. The PX4 PR must be merged and at least one tagged release published with the new manifest before this QGC change can be merged.

Currently this PR points at the existing manifest URL (`https://px4-travis.s3.amazonaws.com/Firmware/manifest.json`) which is already live with format v2 data. The manifest format and URL may be updated as the PX4 PR evolves.

### Commit 1: Manifest parsing backend
- Remove hardcoded `px4_board_name_map` and `_determinePX4StableVersion()` (GitHub Releases API)
- Add manifest download (`_downloadPX4Manifest`), parse (`_parsePX4Manifest`), and data model structs (`PX4ManifestBuildInfo_t`, `PX4ManifestReleaseInfo_t`)
- Add `_buildPX4FirmwareHashFromManifest()` for firmware URL resolution per board/channel
- Expose version properties (`px4DevVersion`, `px4ManifestLoaded`, `px4ManifestDownloading`, `px4AvailableVersions`) to QML
- Wire SHA-256 hash verification from manifest into `_downloadFirmware()`
- Add 5 unit tests covering manifest parsing, version extraction, firmware hash building, missing board handling, and invalid manifest rejection
- Add test fixture `px4_manifest_v2.json` with multi-board, multi-channel, bootloader/variant build coverage

### Commit 2: User-selectable build list UI
- Add `_buildPX4FirmwareNames()` — filters bootloader builds, builds friendly display names, pre-selects `_default` build
- Add `setSelectedPX4Version()` for version picker support in advanced mode
[...]

PR #14530, opened 2026-06-15
feat(FirmwareUpgrade): manifest-based PX4 firmware discovery and selection

dakejahl:
## Summary
Supersedes #13966, rebased onto current master.

Replaces the hardcoded PX4 board→firmware table with PX4 manifest-based discovery and adds user-selectable firmware version and build dropdowns to the upgrade dialog.

## Problem
PX4 firmware selection relied on a hardcoded board-id→firmware-name map that needed a manual entry for every new flight controller, and the upgrade dialog exposed only a single stable/beta/dev option with no insight into the build variants actually available for the detected board.

## Solution
Discover firmware from the PX4 manifest (board id, channel, version, build variants) and drive the upgrade dialog from it, with version and build selection that mirrors the existing ArduPilot flow and a redesigned advanced mode. Firmware downloads move to the CloudFront CDN, and the decompressed image's SHA-256 is verified against the manifest before flashing.


mrpollo:
Thanks for the rebase, I think I'm still missing a few things regarding functionality that need to be addressed, I'm already working on this, and I'll get it updated soon. 

The other thing to keep in mind, is that we can't merge this change until the PX4 infra side is ready, since its going depending on servers and files being available with the right data, right now its pointing to a demo file for testing.

I'm going to close so it doesnt get accidentally merged, but will pull your rebased branch and carry on with this one.

PR #14532, opened 2026-06-15
style(FirmwareUpgrade): clang-format FirmwareImage and FirmwareUpgradeController

dakejahl:
## Summary
Apply clang-format (the repo's pinned 21.1.7 config) to FirmwareImage and FirmwareUpgradeController with no functional changes.

## Problem
These files predate clang-format enforcement and are not `.clang-format`-compliant. Since the pre-commit/CI check formats whole files, any PR that touches them is forced to reformat the entire file, burying the real change under hundreds of lines of pointer-alignment, include-sorting, brace-style and comment-reflow churn. #14530 is the immediate case.

## Solution
Reformat the files in isolation so the churn lands as a standalone, trivially-verifiable commit. The diff is pure formatting — identical include set, and code token-identical to master ignoring whitespace — and #14530 rebases on top to show a clean, formatting-free diff.

### DIFF
--- a/src/Vehicle/VehicleSetup/FirmwareImage.cc
+++ b/src/Vehicle/VehicleSetup/FirmwareImage.cc
@@ -1,12 +1,13 @@
 #include "FirmwareImage.h"
-#include "JsonParsing.h"
-#include "QGCApplication.h"
-#include "FirmwarePlugin.h"
-#include "FirmwarePluginManager.h"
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
We aren't doing a code base wide reformat yet. We'll probbaly do this in 5.2. So closing for now t reduce churn in 5.1

PR #14556, opened 2026-06-18
QtLocationPlugin: tile cache, fetch, and provider overhaul

HTRamsey:
## Summary

Overhaul of the QtLocation plugin's tile cache, fetch pipeline, and providers.

### Providers
- HTTPS tile URLs; `mapTypeId`/server-count/format-detection correctness
- LINZ Basemaps token via settings

### Tile fetch / network
- Enforce adaptive concurrency: gate concurrent leader GETs at the cellular/wifi budget (previously computed but never applied)
- Single transfer-timeout authority (20s leader, `chrono` overload)
- `must-revalidate`/`no-cache`: cache then lazily revalidate
- Viewport-aware prioritization: drop pending off-screen tiles, keep in-flight visible tiles ahead of prefetch backlog
- 7-day default tile expiry; identifiable QGroundControl UA for OSM
- Drop dead `QNetworkRequest` cache/HTTP2 attributes and unused include

### Tile cache (SQLite)
- True LRU eviction via `accessed`-timestamp column (was FIFO by insert date)
- `mustRevalidate` column; in-place schema migrations to `user_version` 4
- Fix WAL read-lock that blocked legacy-DB reset on `DROP TABLE`
- Schema DDL/migrate/reset split into a standalone `QGCTileDatabaseSchema` namespace, independently unit-testable
- `QGeoFileTileCacheQGC` now subclasses `QAbstractGeoTileCache` directly so the SQLite store is the sole authoritative cache (no shadow flat-file/QCache3Q layers)
- `QGCCachedTileSet` download state is now main-thread-only (queued task-result delivery), removing the per-tile mutex

### Import
- MBTiles (TMS→XYZ) and PMTiles v3 (Hilbert dir, gzip) import into the tile cache as offline sets

### Tests
- Added across providers, fetch, cache, schema, importer, and circuit-breaker.

## Notes
- Legacy (pre-versioned) caches are discarded and rebuilt on first open; versioned caches migrate in place.


DonLakeFlyer:
@HTRamsey Should this wait for 5.2? Concerned such a major change will get no bake time before 5.1 goes out?

HTRamsey:
Yup that's fine

DonLakeFlyer:
@HTRamsey This is really old and out of date now? Close?

PR #14587, opened 2026-07-06
RemoteID/PositionManager: fix GCS Live GNSS position

julianoes:
Two fixes so the **Live GNSS** Remote ID GCS position works on Android / in FAA regions.

- **Freshness from local arrival time:** `_updateLastGCSPositionInfo()` stored `update.timestamp()` (the source's clock), which on Android is offset from the system clock the freshness check compares against — making `gcsGPSGood` flap at 1 Hz near the 5 s threshold. Now stamps local arrival time.
- **Keep altitude on poor vertical accuracy:** altitude was dropped above a strict 10 m, leaving a non-3D coordinate that the FAA path rejects. Now accept altitude on any 3D fix within threshold (or with no reported accuracy), and raise the threshold to 100 m to match the horizontal one.

### DIFF
--- a/src/PositionManager/PositionManager.h
+++ b/src/PositionManager/PositionManager.h
@@ -1,5 +1,6 @@
 ﻿#pragma once
 
+#include <QtCore/QDateTime>
 #include <QtCore/QObject>
 #include <QtPositioning/QGeoCoordinate>
 #include <QtPositioning/QGeoPositionInfo>
@@ -33,6 +34,13 @@ public:
     qreal gcsPositionHorizontalAccuracy() const { return _gcsPositionHorizontalAccuracy; }
     QGeoPositionInfo geoPositionInfo() const { return _geoPositionInfo; }
     QGeoPositionInfoSource::Error gcsPositioningError() const { return _gcsPositioningError; }
+
+    /// Local arrival time of the last position update which passed the accuracy gates and was
+    /// copied into gcsPosition. Invalid until the first such update arrives. This is the local
+    /// clock rather than the position source's own timestamp, which on some platforms (e.g.
+    /// Android) is offset from the system clock.
+    ///     @return Arrival time, in UTC, of the last position update applied to gcsPosition.
+    QDateTime gcsPositionTimestamp() const { return _gcsPositionTimestamp; }
 
     int updateInterval() const { return _updateInterval; }
 
@@ -74,6 +82,7 @@ private:
     QGeoPositionInfoSource::Error  _gcsPositioningError = QGeoPositionInfoSource::NoError;
 
     QGeoCoordinate _gcsPosition;
+    QDateTime _gcsPositionTimestamp;
     qreal _gcsHeading = qQNaN();
     qreal _gcsPositionHorizontalAccuracy = std::numeric_limits<qreal>::infinity();
     qreal _gcsPositionVerticalAccuracy = std::numeric_limits<qreal>::infinity();
--- a/src/PositionManager/PositionManager.cpp
+++ b/src/PositionManager/PositionManager.cpp
@@ -134,6 +134,10 @@ void QGCPositionManager::_positionUpdated(const QGeoPositionInfo &update)
             if (_gcsPositionHorizontalAccuracy <= kMinHorizonalAccuracyMeters) {
                 newGCSPosition.setLatitude(update.coordinate().latitude());
                 newGCSPosition.setLongitude(update.coordinate().longitude());
+                // Stamp the local arrival time so consumers can tell how fresh gcsPosition is.
+                // Updates rejected by the accuracy gate leave the stamp alone, since they leave
+                // the previous coordinate in place as well.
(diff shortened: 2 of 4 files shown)


src/PositionManager/PositionManager.h
```diff
@@ -85,6 +85,6 @@ private slots:
     QGCCompass *_compass = nullptr;
 
     static constexpr qreal kMinHorizonalAccuracyMeters = 100.;
-    static constexpr qreal kMinVerticalAccuracyMeters = 10.;
+    static constexpr qreal kMinVerticalAccuracyMeters = 100.;
```
julianoes: This might need discussion. It solved the problem for me with Herelink but might not be generic.

DonLakeFlyer:
@julianoes Can you look at the review?

DonLakeFlyer:
# PR #14587 — Implications of the `kMinVerticalAccuracyMeters` Change (10 m → 100 m)

PR: [RemoteID/PositionManager: fix GCS Live GNSS position](https://github.com/mavlink/qgroundcontrol/pull/14587)

The 10 m → 100 m change in `PositionManager.h` (together with the accept-when-unreported logic in `PositionManager.cpp` `_positionUpdated()`) controls when the GCS device's altitude is copied into `gcsPosition`. Implications, by consumer:

## 1. Remote ID (the intended beneficiary)

`RemoteIDManager::_sendSystem()` broadcasts `gcsPosition.altitude()` as the operator geodetic altitude in `OPEN_DRONE_ID_SYSTEM`, and the FAA-region check hard-fails if the coordinate isn't 3D. Phones commonly report vertical accuracy in the 15–50 m range, so the old 10 m gate silently stripped altitude and made `gcsGPSGood` red in FAA regions. Raising to 100 m fixes that. The message carries no operator-altitude accuracy field, so broadcasting a ±100 m altitude is regulatorily better than broadcasting "unknown" — this is the right trade for Remote ID.

## 2. Follow Me — the real safety-relevant implication

`FollowMe.cc` sends `gcsCoordinate.altitude()` as `altMetersAMSL` in `FOLLOW_TARGET`. If the firmware is configured to track target altitude, the vehicle could now be commanded to follow an altitude that's up to ~100 m wrong (and accuracy figures are 1-σ, so occasionally worse), where previously it would have gotten the stale/last-good altitude instead. Mitigating factor: FollowMe forwards the vertical std-dev in `pos_std_dev[2]`, so firmware *can* de-weight it — but PX4's follow-me altitude modes don't necessarily do so.

## 3. Set-home-to-GCS

`Vehicle::_updateHomepoint()` sends `MAV_CMD_DO_SET_HOME` with `gcsPosition.altitude()` when "update home position" is enabled. Home altitude is the reference for relative-altitude display and RTL behavior; a 100 m error there is operationally significant. Previously the loose altitude simply wasn't there (the coordinate kept the old/NaN altitude).

## 4. NTRIP GGA

`NTRIPGgaProvider.cc` sends the position to the caster for VRS selection; a 100 m altitude error is harmless here.

## 5. The "no reported accuracy → accept" part is the bigger loosening

Independent of the 10→100 constant, the new `verticalAccuracyOk = true` default means any 3D fix from a source that doesn't report `VerticalAccuracy` (e.g. plain NMEA without GST sentences) now passes its altitude straight through, with *unknown* quality. That's arguably correct for Remote ID, but it applies to Follow Me and set-home too.

## 6. Minor pre-existing quirk it doesn't change

When the gate *does* reject, `newGCSPosition` keeps the previous altitude while lat/lon update — so consumers can get a fresh 2D position paired with a stale altitude. Not made worse by this PR, just noting the gate never clears altitude.

## Summary

The constant change is consistent (the horizontal gate is already 100 m via `kMinHorizonalAccuracyMeters`) and is what makes FAA Remote ID usable on phones. The cost is that Follow Me and update-home-position now consume altitude of much lower guaranteed quality. If that's a concern, the gate could be kept per-consumer (e.g. Remote ID reads altitude regardless, Follow Me keeps a stricter check) rather than loosened globally.

DonLakeFlyer:
2 and 3 are the main problems. 3 was already a bit whacky, but 2 will lead to real problems.

julianoes:
Sorry, missed this. I'll revisit it.

julianoes:
Reworked: the vertical accuracy gate in `QGCPositionManager` is untouched, so Follow Me, update-home-position and NTRIP GGA behave exactly as before (points 2, 3, 5).

- `RemoteIDManager::_sendSystem()` now takes the altitude straight from the fix for `OPEN_DRONE_ID_SYSTEM` only — that message has no operator-altitude accuracy field, so a loose altitude beats none, and nothing else consumes it.
- Freshness now comes from a new `QGCPositionManager::gcsPositionTimestamp()`, the local arrival time of updates that actually passed the gate. Fixes the Android clock-skew flapping and Copilot's point that a rejected fix used to look fresh while `gcsPosition` was stale.

DonLakeFlyer:
Thanks @julianoes

PR #14588, opened 2026-07-06
PositionManager/LinkManager: fall back to internal GPS when NMEA disabled

julianoes:
Switching NMEA off (e.g. **Disabled** in Comm Links) closed the UDP socket but left `QGCPositionManager` on the dead NMEA source, freezing the GCS position with no way back to the built-in GPS.

Add `QGCPositionManager::resetNmeaSourceDevice()` to tear down the NMEA source and, if active, revert to the platform default (e.g. integrated Android GPS). `LinkManager` calls it when NMEA is no longer configured, and also tears down a lingering serial NMEA port there.

### DIFF
--- a/src/AppSettings/NmeaGpsSettings.qml
+++ b/src/AppSettings/NmeaGpsSettings.qml
@@ -15,29 +15,49 @@ SettingsGroupLayout {
 
         model: ListModel {}
 
+        // Untranslated value stored in the setting for each entry, parallel to the display model.
+        // LinkManager compares these against fixed strings, so they must not be translated.
+        property var _values: []
+
         onActivated: (index) => {
             if (index !== -1) {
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
@julianoes Can you address the review. I believe both of these are correct from a bug standpoint

julianoes:
@DonLakeFlyer done.

DonLakeFlyer:
@julianoes Can you look at new code review. It tends to be whack-a-mole unfortunately but it also tends to be correct.

julianoes:
Ha ok.

DonLakeFlyer:
If you've dealt with the code review then hide it marked as resolved. Then I'll run another code review

DonLakeFlyer:
Closing due to replacement

PR #14365, opened 2026-05-12
VideoManager: Remove unused/incorrect override

Ryanf55:
## Description
Fixes a compilation error on Ubuntu 22 introduced by https://github.com/mavlink/qgroundcontrol/pull/14332.
This function is not used in the codebase, so it can just be removed.

```
/home/ryan/Dev/qgroundcontrol/src/VideoManager/VideoReceiver/GStreamer/HwBuffers/GstDmaBufVideoBuffer.h:27:10: error: ‘bool GstDmaBufVideoBuffer::isDmaBuf() const’ marked ‘override’, but does not override
   27 |     bool isDmaBuf() const override { return true; }
```
   
   

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] CI/Build changes
- [ ] Other

## Testing

- [x] Tested locally
- [ ] Added/updated unit tests
- [ ] Tested with simulator (SITL)
- [ ] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [ ] ArduPilot

## Checklist

- [x] I have read the [Contribution Guidelines](CONTRIBUTING.md)
- [x] I have read the [Code of Conduct](CODE_OF_CONDUCT.md)
- [x] My code follows the project's coding standards
- [ ] I have added tests that prove my fix/feature works
- [x] New and existing unit tests pass locally

## Related Issues

---
By submitting this pull request, I confirm that my contribution is made under the terms of the project's dual license (Apache 2.0 and GPL v3).

PR #14589, opened 2026-07-06
VideoManager: Remove unused/incorrect override

DonLakeFlyer:
Removes the unused/incorrect `isDmaBuf() const override` from `GstDmaBufVideoBuffer`. The method has no callers anywhere in the codebase.

This replaces #14365, which had become unmergeable after the GStreamer backend overhaul (#... "Video: overhaul GStreamer backend and packaging") relocated `GstDmaBufVideoBuffer.h` into the new `HwBuffers/dmabuf/` directory. This branch re-applies the same one-line removal at the file's new location on top of current `master`.

Original author: @Ryanf55

### DIFF
--- a/src/VideoManager/VideoReceiver/GStreamer/HwBuffers/dmabuf/GstDmaBufVideoBuffer.h
+++ b/src/VideoManager/VideoReceiver/GStreamer/HwBuffers/dmabuf/GstDmaBufVideoBuffer.h
@@ -19,8 +19,6 @@ public:
     GstDmaBufVideoBuffer(GstSample* sample, const GstVideoInfo& videoInfo, const QVideoFrameFormat& format,
                          EGLDisplay eglDisplay);
 
-    bool isDmaBuf() const override { return true; }
-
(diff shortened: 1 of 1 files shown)


HTRamsey:
So that function does actually get used internally in Qt but at least in Qt 6.10 it was only used in QtWidgets. Something to keep an eye on if we remove this

DonLakeFlyer:
If it ever comes back it will fail compile right?

HTRamsey:
Nah there's a default virtual that return false. It's used to determine video texture fallback policy in qvideotexturehelper
image

image

DonLakeFlyer:
Oh, ick. So this is a GST version problem in that Ubuntu 22 has an older GST? I originally thought this was just a compile time thing on an override that wasn't actually implemented. What's the way to deal with this? How does Qt run on Ubuntu 22 if this virtual is missing?

HTRamsey:
This is entirely qt based and unrelated to gstreamer, since it applies to all video backends that qt supports. I guess the best solution would be to handle the error in cmake or with compile defs

PR #14593, opened 2026-07-09
feat(RemoteID): per-region operator IDs with EU validity enforcement

DonLakeFlyer:
## Overview

Replaces the single `operatorID` / `operatorIDValid` Remote ID settings pair with per-region facts (`operatorIDEU`, `operatorIDFAA`), with legacy stored values migrated automatically.

## EU operator ID: valid by construction

- Writes to `operatorIDEU` are gated by an EN 4709-002 validator (format, Luhn mod-36 checksum, ASCII country code).
- The full 19/20 character ID (including the 3 secret characters) must be entered, but only the 16-character public part is ever stored or broadcast — the secret characters never persist.
- Region changes enforce regulatory requirements: EU forces operator ID broadcast on, FAA forces live operator location.

## RemoteIDManager

- Broadcasts the region-appropriate operator ID, gated on validity.
- Tracks the vehicle's basic-ID-missing arm status error as current state instead of latching it forever.
- Bounded UTF-8 decode of arm status error text.

## Supporting infrastructure (separate commits)

- `FactMetaData`: new `maxStringLength` key, enforced during validation and wired into `FactTextField.maximumLength`; strict unknown-key rejection for fact metadata JSON.
- New `JsonResourceAuditTest` validates every bundled JSON resource against its parser — and caught latent errors fixed here (`defaultValue` vs `default`, `minValue`/`maxValue` vs `min`/`max` which had left the Remote ID fixed-position facts without limits).
- QML settings/config generators: strict schema validation with contextual errors, stable `objectName`s for UI tests.
- MockLink: arm-status override and last-received-message test APIs.
- Categorized-logging pre-commit hook plus qCWarning conversions.

## Testing

New suites: `RemoteIDSettingsTest` (unit), `RemoteIDManagerTest` (integration), `RemoteIDSettingsUITest` (QML UI), plus expanded `FactMetaDataTest`, `JsonResourceAuditTest`, and ~130 new generator pytest cases. Full unit suite and generator tests pass.


DonLakeFlyer:
This is a replacement for #14262. Along the way I found a number of directly and indirectly related boat anchors I took care of. This will now show operator id validation errors using the standard validation error ui.

The main problem was that I screwed up the remote id settings ui when I converted it to json. This pull fixes all that. It use the pre-existing custom fact validation support to hook in the operator Id validation. In the end it accomplishes the same validation that original pull did. 

Also cleaned up a bunch of complex code path by storing EU and FAA operator id formats as two separate settings.

PR #14596, opened 2026-07-10
feat(ParameterEditor): allow Advanced UI override of read-only params

DonLakeFlyer:
Replacement for #14415.

## Summary

Lets Advanced-UI users override the read-only lock on a parameter, mirroring the existing "Force save" escape hatch for min/max bounds.

## Problem

QGC respects the `readOnly` metadata flag (#14007) by hiding the input field entirely, with no way to override it — even for diagnostic or recovery cases where the lock is too strict, such as unjamming ArduPilot (ArduPilot/ardupilot#33136).

## Solution

In `ParameterEditorDialog`, read-only params get a "Force edit read-only param" checkbox nested under the existing "Advanced settings" checkbox — hidden by default, requiring a deliberate opt-in. Ticking it switches to the normal edit view (input + Save); unticking "Advanced settings" re-locks the param. UI-only change, no firmware/C++ gating.

`QGCPopupDialog` now reconfigures its buttons whenever `buttons` changes, so the dialog's Save/Cancel vs Close state tracks the force-edit toggle. Also removes the unused `validate`/`validateValue` code path from `ParameterEditorDialog`.

### DIFF
--- a/src/QmlControls/ParameterEditorDialog.qml
+++ b/src/QmlControls/ParameterEditorDialog.qml
@@ -11,12 +11,10 @@ QGCPopupDialog {
     id:         root
     title:      fact.componentId > 0 ? fact.name : qsTr("Value Editor")
 
-    buttons:    fact.readOnly ? Dialog.Close : (Dialog.Save | (validate ? 0 : Dialog.Cancel))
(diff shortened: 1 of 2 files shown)


dakejahl:
Thanks for picking this up!

PR #14597, opened 2026-07-12
fix(QGCFileDialog): avoid macOS native dialog crash on Plan open/save

datlt4:
## Description

Clicking Open/Save in Plan view crashed on macOS 13 with EXC_BAD_ACCESS
image

Force Qt's non-native FileDialog on macOS via DontUseNativeDialog so the
buggy NSOpenPanel path is bypassed. Other platforms keep the native
dialog.

## Type of Change

- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [x] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [ ] Linux
- [ ] Windows
- [x] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [x] PX4
- [ ] ArduPilot

## Checklist

## Related Issues

---

### DIFF
--- a/src/QmlControls/QGCFileDialog.qml
+++ b/src/QmlControls/QGCFileDialog.qml
@@ -128,6 +128,9 @@ Item {
         nameFilters:    _root.nameFilters ? _root.nameFilters : []
         title:          _root.title
         defaultSuffix:  _root.defaultSuffix
+        // macOS: the native NSOpenPanel crashes inside Apple's ViewBridge/ImageIO when it
+        // snapshots the accessory (name-filter) view over XPC. Use Qt's non-native dialog there.
+        options:        Qt.platform.os === "osx" ? FileDialog.DontUseNativeDialog : 0
 
         onAccepted: {
             var fullPath = QGCFileDialogController.urlToLocalFile(selectedFile)


DonLakeFlyer:
Not sure why we would not use the native dialog? Why is it crashing? This seems to fix the symptom as opposed to the problem.

DonLakeFlyer:
@datlt4 Can you get me the stack trace for the crash? This is likely a mac sdk problem solely on 13 and below, whereas 14+ has already fixed it.

datlt4:
> @datlt4 Can you get me the stack trace for the crash? This is likely a mac sdk problem solely on 13 and below, whereas 14+ has already fixed it.

Thanks for looking into this. You're right that I'm on macOS 13.

I investigated further, and this isn't a QGC bug. The crash only happens when launching from Qt Creator's Run button — running the app bundle directly, or under lldb, never crashes.

The cause is Qt Creator's disclaim launch wrapper, visible in the log:

The command "/Users/.../Qt Creator.app/Contents/Resources/libexec/disclaim /Users/.../QGroundControl" terminated abnormally.
The crash report confirms the parent: parentProc: "Qt Creator", coalitionName: "org.qt-project.qtcreator". The crashing thread is entirely inside Apple frameworks (native NSOpenPanel / ViewBridge), with no QGC or Qt frame involved — so it doesn't affect users running a normally-launched app bundle.

Given that, I'll close this PR and just keep the patch locally for my own dev setup. Thanks for taking the time to review!

DonLakeFlyer:
Wow, that's really odd. Will have to remember this if someone else hits it so they know what they can do.

PR #14605, opened 2026-07-15
PX4FirmwarePlugin: use correct unset param1

julianoes:
## Description

NAV_TAKEOFF param1 allows to set the minimum pitch. Generally, when not setting a param it should be set to NAN or 0. The sentinel value -1 is not in the spec.

Note that this value is not used in PX4 anyway.

Also, unused param2 can be NAN instead of 0 which is generally preferred.

## Type of Change

- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [ ] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [x] PX4
- [ ] ArduPilot

## Checklist

## Related Issues

https://github.com/PX4/PX4-Autopilot/pull/27899

---

### DIFF
--- a/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
+++ b/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc
@@ -330,8 +330,9 @@ void PX4FirmwarePlugin::guidedModeTakeoff(Vehicle* vehicle, double takeoffAltRel
         vehicle->defaultComponentId(),
         MAV_CMD_NAV_TAKEOFF,
         true,                                   // show error is fails
-        -1,                                     // No pitch requested
-        0, 0,                                   // param 2-4 unused
+        NAN,                                    // No pitch requested
+        NAN,                                    // param 2 unused
+        0,                                      // param 3 takeoff flags
         NAN, NAN, NAN,                          // No yaw, lat, lon
         static_cast<float>(takeoffAltAMSL));    // AMSL altitude
 }


julianoes:
Testing pending Qt download...

julianoes:
PX4 SITL tested.

DonLakeFlyer:
@hamishwillee Mavlink spec says nothing about how to deal with not passing various MAV_CMD_NAV_TAKEOFF params. Is there really a spec that says NAN is always not used. If so, I believe ardupilot tends to puke on NAN in various commands.

@julianoes Is this fact in px4 firmware side version agnostic? Not something that change somewhere along the way?

julianoes:
> I believe ardupilot tends to puke on NAN in various commands.

It used to but that's mostly fixed up by now. In this case that doesn't matter as this is the PX4 plugin.

We started rejecting invalid commands on the PX4 side which triggered QGC takeoff to fail. So obviously we had to revert the strict check on the PX4 side. However, instead of just ignoring that QGC sends something non-compliant, I opted to fix it.

DonLakeFlyer:
Ok, so I think that means that new QGC with this change in it isn't going to cause old PX4 to fail on the NANs.

hamishwillee:
@DonLakeFlyer @julianoes Late to the party sorry.

> @hamishwillee Mavlink spec says nothing about how to deal with not passing various MAV_CMD_NAV_TAKEOFF params. Is there really a spec that says NAN is always not used. 

Yes. Its is defined in the encapsulating messages: [COMMAND_INT](https://mavlink.io/en/messages/common.html#COMMAND_INT) and [COMMAND_LONG](https://mavlink.io/en/messages/common.html#COMMAND_LONG). Note the `invalid` values listed for floats is `NaN` and for int32s in `INT32MAX`. So the value should be `INT32MAX` for unused if sent in a COMMAND_INT for param5,6 and NaN for other cases. 

[MISSION_ITEM_INT](https://mavlink.io/en/messages/common.html#MISSION_ITEM_INT) doesn't say this, but should.

It is desirable we move towards that understanding and set the correct invalid values. Right now QGC tends to send 0 more often than not.

The reason is that otherwise it is hard to safely start using an unused parameter if it was silently accepting any old crap. It also means that a system can infer when a particular param is not supported, because it will always get a NACK.

>  If so, I believe ardupilot tends to puke on NAN in various commands.

This is my understanding too - @peterbarker has told me that historically ArduPilot couldn't handle NaN and while it may have been improved, it is not guaranteed to be robust.

So this change is desirable, but we needs to maintain current behaviour for ArduPilot.

PX4 should not fail with any change to an unused parameter on old builds: in most cases it simply ignores them.
My intent/hope is that we will log errors in this respect for a few releases, fix QGC, and eventually stop accepting the old values.

hamishwillee:
PS @DonLakeFlyer Note that the change is contentious (in PX4) in that we don't want to arbitrarily break consumers. Discussion ongoing in https://github.com/PX4/PX4-Autopilot/pull/27899. This case probably OK to merge, but we probably shouldn't be rushing off to do bigger changes.

PR #14634, opened 2026-07-18
fix(facts): use RPM shortDesc for ESC rotation rate

Bartok9:
## What
Replace EscStatus `rpm` shortDesc "Rotation Per Minute" with the standard abbreviation **RPM**.

## Motivation
Telemetry panels are dense; RPM is universal and matches generator fact style (`units: rpm`).

## Verification
JSON-only metadata; no unit conversion change.

Reviewed by hand on master tip.


Bartok9:
Consolidating per @DonLakeFlyer's request into a single PR: #14651. Closing this one. Thanks!

PR #14637, opened 2026-07-18
fix(facts): use SI capital V for battery voltage units

Bartok9:
Battery voltage fact used units `"v"` while Esc/Generator voltage facts use `"V"`. Align battery metadata unit casing to SI capital V.

No conversion table change expected; Fact units string only.

AI-assisted; human-reviewed. Path-orthogonal to other fact PRs tonight.


Bartok9:
Consolidating per @DonLakeFlyer's request into a single PR: #14651. Closing this one. Thanks!

PR #14671, opened 2026-07-21
fix(math): reject opposite-signed infinities in fuzzyCompare

Bartok9:
## Summary
- `QGC::fuzzyCompare` (double + float defaults) now treats infinity via `qIsInf` + `==`, so `+inf` vs `-inf` is **not** equal. Qt’s `qFuzzyCompare(+inf, -inf)` returning true is a known pitfall.
- Unit coverage: same/opposite signed inf, inf/NaN (double + float), plus existing `limitAngleToPMPId(3π)` wrap check.

## Test plan
- [x] Diff on `src/Utilities/Math/QGCMath.cc` + `test/Utilities/QGCMathTest.*`
- [ ] CI Linux unit matrix green

### DIFF
--- a/src/Utilities/Math/QGCMath.cc
+++ b/src/Utilities/Math/QGCMath.cc
@@ -112,6 +112,9 @@ bool fuzzyCompare(double value1, double value2)
         return true;
     } else if (qIsNaN(value1) || qIsNaN(value2)) {
         return false;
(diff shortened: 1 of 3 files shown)


Bartok9:
Closing remaining open work — standing down all automated coding on this repo until further notice per maintainer feedback. Thanks for your time.

PR #14673, opened 2026-07-21
test(json): external header wrong type and version range

Bartok9:
## Summary
External QGC JSON header validation now has reject tests parallel to the internal-file suite:
- wrong `fileType`
- version above allowed max

Guards plan/mission-style loaders. AI-assisted; human-reviewed.

## Test plan
- [x] JsonHelperTest only
- [ ] CI


DonLakeFlyer:

PR #14674, opened 2026-07-21
test(comms): DataRateTracker zero-byte records stay idle

Bartok9:
## Summary
`recordBytes(0)` must keep `totalBytes` at 0, `bytesPerSec` at 0, and leave `rateUpdated` false inside a fresh window so telemetry plugins cannot invent a spike from empty ticks.

## Test plan
- [x] DataRateTrackerTest only
- [ ] CI


DonLakeFlyer:

PR #14678, opened 2026-07-21
test(QmlUITests): instrument PlanViewLayerUITest to localize CI-only QML warnings

DonLakeFlyer:
## Purpose

Diagnostic-only PR — **do not merge**. `PlanViewLayerUITest::_testEditingLayerSwitching` intermittently fails CI on the strict-mode log check with:

- `QQmlContext: Cannot set context object on invalid context.`
- `QQmlComponent: Cannot create a component in an invalid context`

These are classic symptoms of async component creation racing a context being destroyed, but the CI log doesn't show *where* in the test they fire.

## Changes

- `PlanViewLayerUITest.cc`: `[LayerUITest]` phase markers before/after every interaction in sections 2.1–2.7, plus startUI/stopUI boundaries.
- `QmlUITestBase.cc`: `[QmlUITeardown]` markers covering each teardown sub-phase (window close, settle drain, incubator drain + remaining count, GC settle, window delete, engine delete) — teardown is the prime suspect.
- Markers use `qInfo()` (the test env sets `QT_LOGGING_RULES=*.debug=false`, so `qDebug()` is invisible) and are strict-mode ignored via a pattern registered in `startUI()`.

## How to use

When the failure reproduces, find the two warnings in the CI log and read the nearest preceding marker to identify the emitting phase.

## Verification

- Local build + `ctest -R PlanViewLayerUITest` passes; all markers verified in `-V` output.

To be reverted once the culprit is identified.

### DIFF
--- a/src/PlanView/MissionItemEditor.qml
+++ b/src/PlanView/MissionItemEditor.qml
@@ -349,8 +349,10 @@ Rectangle {
         anchors.margins:    _innerMargin
         anchors.left:       parent.left
         anchors.top:        topRowLayout.bottom
-        asynchronous:       true
 
+        // Load synchronously: an async incubation can still be in flight when the
+        // TreeView collapses the mission group and invalidates this delegate's
+        // context, emitting "Cannot create a component in an invalid context".
         Component.onCompleted: _root._loadEditor()
     }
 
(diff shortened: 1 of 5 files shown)


DonLakeFlyer:
Superseded by #14690 — the instrumentation here localized the failure; the actual fix (synchronous editor loading in MissionItemEditor.qml) is in that PR.

Issue #14470, opened 2026-06-05
Potential memory leak and UI freezes during long flights


s-lisovenko:
### Have you checked the latest release for fixes?

- [x] Yes, I’ve tested on the latest release

### Description

**Expected Behavior**
The application should remain responsive and stable during flight operations, with behavior comparable to the Qt5 version.

**Current Behavior**
The application gradually consumes more memory over time, which appears to lead to UI freezes and noticeable lag. The issue is especially visible during long-duration flights.

**Steps To Reproduce**
The issue shown in the attached video is the easiest way to reproduce it, but similar behavior also occurs intermittently during long flights.

https://github.com/user-attachments/assets/60e1a224-cb17-4dcf-8800-7175fee09671

### Platform

macOS

### Flight Stack

PX4

### System Information

The issue may be related to rendering or processing a large number of flight map elements, which can potentially freeze the main UI.

DonLakeFlyer:
I assume this repros on SITL for you? Can you give me simple steps to reproduce using just a flight? Not by creating massive survey areas.

s-lisovenko:
> I assume this repros on SITL for you? Can you give me simple steps to reproduce using just a flight? Not by creating massive survey areas.

Hello, this issue was discovered on a real drone during a very long flight. A quick field analysis shows it may be related to the large number of map elements and the flight path. The main thread, responsible for the user interface, is completely overloaded and cannot correctly handle the GUI. I don't know how this was solved in Qt5, but this issue doesn't exist there.

Additionally, in the scenario where a huge flight plan is created, the memory does not seem to be properly freed after the plan is deleted.

DonLakeFlyer:
Can you give me the plan file?

s-lisovenko:
Okay, first of all, thanks for trying to fix this.

I probably didn't describe the problem correctly. There are actually two issues here.
1. Interface lag due to the large number of elements on the map (even when they are deleted, the main thread remains 100% CPU-bound).
2. Potential memory leak (I noticed that when a plan is created and then deleted, the memory isn't freed up).

Regarding interface overload (1.) , I've attached a simple default ADB-S generator with 3000 vehicles (can be configured in a script) (we use an AIS with a much larger number of boats). You will see noticeable UI lags.
[adbs_generator.py](https://github.com/user-attachments/files/28995848/adbs_generator.py)

I assume that flight path drawing can also lead to lags in the interface during long flights and QGC operation (for example, 5 hours or more)


[...]

PR #14679, opened 2026-07-22
fix(ADSB): eliminate UI freezes and memory growth with high traffic volumes

DonLakeFlyer:
## Summary

Fixes the ADS-B portion of #14470. With ~3000 targets streaming from an SBS-1 server, the UI thread saturated (map froze) and memory grew without bound — a regression relative to Qt5 builds.

## Root causes and fixes

- **Main-thread parsing with a throughput cap**: `ADSBTCPLink` parsed SBS-1 lines on the UI thread, capped at 2000 lines/sec (50 ms timer × 100 lines). A 3000 msg/sec feed outran the cap, so the line buffer grew forever while the UI churned. The link now runs on a worker thread (canonical QThread worker pattern); lines are parsed inline as they arrive and the cap/buffer are gone.
- **Per-message property updates**: every message triggered property signals → `MapQuickItem` re-projections across thousands of map items. `ADSBVehicle` now throttles property updates to 1 Hz. Alert transitions bypass the throttle; the expiration keep-alive still updates on every message.
- **Unbounded vehicle accumulation**: the stale-vehicle cleanup timer only ran while a TCP link was active, so MAVLink-sourced ADS-B vehicles never expired. The timer now always runs.
- **Heavy map delegate**: ADS-B targets shared `VehicleMapItem`, instantiating a `MultiEffect` shader (which was 0×0 and never rendered) plus vehicle-only bindings per target. New lightweight `ADSBVehicleMapItem` delegate; `VehicleMapItem` is now vehicle-only.

Also restores hostname support for the ADS-B server address (Qt5 behavior lost in the Qt6 port) by deferring resolution to `QTcpSocket::connectToHost`, and bounds the worker-thread shutdown wait with a warning.

## Testing

- New unit tests: vehicle expiration/cleanup via the manager (regression test for the MAVLink-source leak), TCP link start/stop/restart lifecycle through the settings facts, and property update throttling.
- `ADSBTest` 8/8 passing; `ToolbarIndicatorUITest` passing; qmllint/clang hooks clean.
- Manually verified with a 3000-target SBS-1 generator feeding localhost: map stays responsive, memory stable, targets expire correctly.

### DIFF
--- a/src/ADSB/ADSBTCPLink.cc
+++ b/src/ADSB/ADSBTCPLink.cc
@@ -2,17 +2,15 @@
 // #include "QGCSensors.h"
 #include "QGCLoggingCategory.h"
 
-#include <QtCore/QTimer>
 #include <QtNetwork/QTcpSocket>
 
 QGC_LOGGING_CATEGORY(ADSBTCPLinkLog, "ADSB.ADSBTCPLink")
 
-ADSBTCPLink::ADSBTCPLink(const QHostAddress &hostAddress, quint16 port, QObject *parent)
+ADSBTCPLink::ADSBTCPLink(const QString &hostAddress, quint16 port, QObject *parent)
     : QObject(parent)
     , _hostAddress(hostAddress)
     , _port(port)
     , _socket(new QTcpSocket(this))
-    , _processTimer(new QTimer(this))
 {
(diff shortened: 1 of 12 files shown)


DonLakeFlyer:
@s-lisovenko Can you you try latest daily to see how this works with your ADSB stuff?

PR #14681, opened 2026-07-22
test(math): pin crc32 single-byte and all-zero buffer paths

Bartok9:
## Summary
Extends `QGCMathTest` with the ISO-HDLC single-byte `0x00` vector and an eight-zero incremental vs single-pass equality check.

## Test plan
- [x] CRC vectors match zlib.crc32 / known constants
- [ ] CI

Claim: bartok
Operator: bartok
Campaign: aerial-drone


DonLakeFlyer:

PR #14682, opened 2026-07-22
test(geo): pin zero-distance destination and self geodesic length

Bartok9:
## Summary
Asserts `geodesicDestination(..., 0)` returns the start point and `geodesicDistance` to self is zero (including elevated alt that must not affect horizontal geodesics).

AI-assisted; human-reviewed. Tests only. No flight safety path edits.

## Test plan
- [x] GeoTest slots only
- [ ] CI

Claim: bartok
Operator: bartok
Campaign: aerial-drone


DonLakeFlyer:

PR #14683, opened 2026-07-22
test(memory): cover secureZero one-byte wipe and cleared size

Bartok9:
## Summary
Pins one-byte `secureZero` and confirms `QByteArray` wipe ends at `size() == 0` so callers cannot reuse a cleared buffer as still sized.

## Test plan
- [x] SecureMemoryTest only
- [ ] CI

Claim: bartok
Operator: bartok
Campaign: aerial-drone


DonLakeFlyer:

PR #14689, opened 2026-07-22
test(cli): pin normalizeArgs empty-colon and mixed options

Bartok9:
## Summary

- Pin `normalizeArgs` behavior for empty `--option:` values (drop blank token) and mixed colon + plain flags.

## Motivation

The unittest-build normalizer intentionally injects an empty arg for bare `--unittest`, but general `--logging:` style empty values must **not** inject a blank token. Locking both alone and in a mixedargv keeps boot option parsing stable.

## Verification

- Extends `QGCCommandLineParserTest` pure-function layer (no process exit paths).
- Tests only; no product code changes.

---

**Agent-Owner:** `sera` · **Platform:** hermes · **Claim-TTL:** 24h


DonLakeFlyer:

Bartok9:
I understand, not trying to create churn. But I totally get it. I'll stop my agent from anything automatically here, without my full review first. All good. Thanks! (Hope you're having a great day)

PR #14705, opened 2026-07-23
fix(params): improve feedback when loading parameter files

### LINKED ISSUE #14704: Parameter file load reports "no differences" for dynamically generated ArduPilot helicopter parameters

- [x] Yes, I've tested on the latest release

### Description

When loading a .params file for an ArduPilot helicopter (FRAME_CLASS=11, Heli_Dual), QGC reports "There are no differences between the file loaded and the current settings on the Vehicle" but parameters are NOT actually updated.

This appears related to #14346 but that issue was closed as not reproducible. The difference here is that **helicopter parameters (H_*) are dynamically generated** by ArduPilot only after FRAME_CLASS is set to 6 (Heli) or 11 (Heli_Dual) and the vehicle reboots. This creates a specific failure scenario:

1. QGC may cache the parameter list from initial connection (before H_* params exist)
2. After reboot, if QGC begins parameter comparison before the full parameter list is retrieved, H_* parameters may be skipped
3. The comparison logic may not handle parameters that exist on the vehicle but were loaded after initial connection

### Steps to Reproduce

1. Flash ArduCopter 4.7.0 SimOnHardWare firmware to Cube Orange+
2. Connect via USB to QGC 5.0
3. Set FRAME_CLASS = 11 (Heli_Dual)
4. Reboot vehicle
5. Verify H_* parameters are visible (H_DUAL_MODE, H_RSC_MODE, etc.)
6. Go to Vehicle Setup > Parameters > Tools > Load from file
7. Select a .params file containing H_* parameters with values different from defaults
8. Observe "no differences" message
9. Check individual parameters, they have NOT been updated

### Evidence from Log Analysis

| Parameter | Value on Cube | Value in File |
|-----------|---------------|---------------|
| H_RSC_MODE | 1 | 3 |
| ARMING_SKIPCHK | 64 | 80 |
| FS_THR_ENABLE | 1 | 0 |
| FS_GCS_ENABLE | 1 | 0 |
| EK3_SRC1_YAW | 0 | 8 |
| H_CYC_MAX | 2500 | 7.0 |

All parameters had to be set manually one by one.

### Expected Behavior

Parameters in the file should be compared against vehicle values and updated where different. A summary should show how many parameters were updated vs skipped.

### Actual Behavior

"No differences" message appears, but parameters remain at default values.

### Platform

macOS

### Flight Stack

ArduPilot

### System Information

- QGC Version: 5.0 (macOS)
- OS: macOS 15.x
- Firmware: ArduCopter V4.7.0 CubeOrange+SimOnHW
- Frame: Heli_Dual (FRAME_CLASS=11)

### Root Cause Hypothesis

The `loadFromFile` function likely compares parameter names against the vehicle's current parameter set. If QGC caches the parameter list from initial connection (before FRAME_CLASS change and reboot), it may not see the newly available H_* parameters even after they become visible in the parameter editor.

Alternatively, there may be a race condition where after reboot, QGC begins parameter comparison before the full parameter list is retrieved from the vehicle.

### Proposed Fix

1. Before comparing, trigger a full parameter refresh from vehicle
2. Track count of parameters in file vs parameters successfully matched
3. Show message: "Loaded X of Y parameters. Z parameters not found on vehicle."
4. List skipped parameters in a dialog or log

### Related Issues

- #14346 (same symptom, closed as not reproducible, tested with non dynamic params)
- #4154 (partial load status feedback)
- #10438 (unable to write params to new controller)

thehalvo:
## Summary

- Improve user feedback when loading parameter files to clearly indicate what happened
- Add debug logging for parameter comparison to aid troubleshooting

## Problem

When loading a .params file, QGC reports "no differences" even when:
- Parameters exist in file but not on vehicle
- All parameters already match vehicle settings

This is particularly problematic for ArduPilot helicopters where H_* parameters only exist after setting FRAME_CLASS and rebooting. Users see "no differences" but have no indication of why parameters weren't loaded.

## Solution

- Track count of parsed, changed, unchanged, and missing parameters
- Show specific feedback message based on what happened:
  - "The following parameters differ..." when changes are found
  - "All X parameters already match vehicle settings" when no changes needed
  - "No differences found. X parameters not found on vehicle" when params missing
- Add debug logging showing each parameter comparison for troubleshooting

## Testing

- [x] Tested with ArduPilot helicopter params after FRAME_CLASS change
- [x] Verified feedback messages are accurate
- [x] Build succeeds on macOS

## Related Issues

Fixes #14704

### DIFF
--- a/src/QmlControls/ParameterDiffDialog.qml
+++ b/src/QmlControls/ParameterDiffDialog.qml
@@ -23,9 +23,17 @@ QGCPopupDialog {
         QGCLabel {
             Layout.preferredWidth:  mainGrid.visible ? mainGrid.width : ScreenTools.defaultFontPixelWidth * 40
             wrapMode:               Text.WordWrap
-            text:                   paramController.diffList.count ?
-                                        qsTr("The following parameters from the loaded file differ from what is currently set on the Vehicle. Click 'Ok' to update them on the Vehicle.") :
-                                        qsTr("There are no differences between the file loaded and the current settings on the Vehicle.")
+            text:                   {
+                if (paramController.diffList.count > 0) {
+                    return qsTr("The following parameters from the loaded file differ from what is currently set on the Vehicle. Click 'Ok' to update them on the Vehicle.")
+                } else if (paramController.diffMissingCount > 0) {
+                    return qsTr("No differences found. %1 parameters from the file were not found on the vehicle (see console log for details).").arg(paramController.diffMissingCount)
+                } else if (paramController.diffUnchangedCount > 0) {
+                    return qsTr("All %1 parameters in the file already match the vehicle settings.").arg(paramController.diffUnchangedCount)
+                } else {
+                    return qsTr("There are no differences between the file loaded and the current settings on the Vehicle.")
+                }
+            }
         }
 
         GridLayout {
--- a/src/QmlControls/ParameterEditorController.cc
+++ b/src/QmlControls/ParameterEditorController.cc
@@ -363,10 +363,15 @@ void ParameterEditorController::clearDiff(void)
     _diffList.clearAndDeleteContents();
     _diffOtherVehicle = false;
     _diffMultipleComponents = false;
+    _diffUnchangedCount = 0;
+    _diffParsedCount = 0;
     _diffMissingParams.clear();
 
     emit diffOtherVehicleChanged(_diffOtherVehicle);
     emit diffMultipleComponentsChanged(_diffMultipleComponents);
+    emit diffUnchangedCountChanged(_diffUnchangedCount);
+    emit diffMissingCountChanged(0);
+    emit diffParsedCountChanged(_diffParsedCount);
 }
 
 void ParameterEditorController::sendDiff(void)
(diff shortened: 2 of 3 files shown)


DonLakeFlyer:
Does this handle the case of some params differ and are on vehicle and some params are not currently seen?

Also what happens when it gets params which it doesn't currently have differs by file format:
* QGC format - the param updates are still sent to the vehicle
* MP format - the param updates are not sent to the vehicle

For the QGC format case I think the user should still have the option to de-select them from being updated

DonLakeFlyer:
> Does this handle the case of some params differ and are on vehicle and some params are not currently seen?
> 
> Also what happens when it gets params which it doesn't currently have differs by file format:
> 
> * QGC format - the param updates are still sent to the vehicle
> * MP format - the param updates are not sent to the vehicle
> 
> For the QGC format case I think the user should still have the option to de-select them from being updated

All of that said, if you want I can take over the pull from here if you like. I want to also add comprehensive unit testing for all these cases as well which can be a PITA.

Thanks so much for your help

DonLakeFlyer:
@thehalvo FYI: Thanks for this. I'm going to take this over so I can rework the UI and UTs to match what I'm looking for.

DonLakeFlyer:
Thanks @thehalvo for kicking off the work on improving parameter file load feedback! This has been superseded by #14717, which builds on the same idea with a full diff/review dialog. Closing in favor of that PR.

PR #14725, opened 2026-07-28
fix(CorridorScanComplexItem): Fix the signal emission logic for

cuav-zhang3:
coordinate specification state changes

Add a cache variable to record the previous coordinate specification status, detect status changes during corridor polygon reconstruction, and emit corresponding signals to ensure the task controller correctly updates the waypoint connection lines.

## Type of Change

- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [x] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [ ] Linux
- [x] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [x] ArduPilot

## Screenshots

test1
test2

## Checklist

## Related Issues

---

### DIFF
--- a/src/MissionManager/CorridorScanComplexItem.cc
+++ b/src/MissionManager/CorridorScanComplexItem.cc
@@ -19,6 +19,10 @@ CorridorScanComplexItem::CorridorScanComplexItem(PlanMasterController* masterCon
 {
     _editorQml = "qrc:/qml/QGroundControl/PlanView/CorridorScanEditor.qml";
 
+    // specifiesCoordinate() depends on _corridorPolyline.count(), so cache the
+    // initial value before any signal connections fire to keep the cache in sync.
+    _specifiesCoordinate = specifiesCoordinate();
+
     // We override the altitude to the mission default
     if (_cameraCalc.isManualCamera() || !_cameraCalc.valueSetIsDistance()->rawValue().toBool()) {
         _cameraCalc.distanceToSurface()->setRawValue(SettingsManager::instance()->appSettings()->defaultMissionItemAltitude()->rawValue());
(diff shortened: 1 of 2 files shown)


DonLakeFlyer:
The repro for this is:
* Set home
* Add takeoff
* Add corridor scan
* Trace - add two points
* Done tracing

Now the corridor scan doesn't connect to the takeoff item

DonLakeFlyer:
@cuav-zhang3 Thanks for the pull. I'm going to reorganize this a bit to make the fix clearer.

DonLakeFlyer:
Thanks for tracking this down — the diagnosis is correct: `specifiesCoordinate()` is dynamic but `specifiesCoordinateChanged` was never emitted, so `MissionController` never re-evaluated the waypoint connector lines.

I'm closing this in favor of #14749, which fixes the same bug with a slightly different structure:

- **Emission tied to the actual state change.** `specifiesCoordinate()` depends only on the polyline vertex count, so the emit lives in a dedicated `_updateSpecifiesCoordinate()` slot connected to `QGCMapPolyline::countChanged`, rather than inside `_rebuildCorridorPolygon()`. That slot is also triggered by corridor width changes (which can't affect `specifiesCoordinate()`), and a future refactor of the polygon rebuild couldn't silently drop the signal emission.
- **`countChanged` covers every mutation path** — `appendVertex`, `clear()`, and `setPath`/JSON load all route through it, so both the false→true and true→false transitions are always signaled.
- **No constructor cache seeding needed.** The `_specifiesCoordinate = specifiesCoordinate()` line in this PR runs at a point where the polyline is guaranteed empty (the KML/SHP load happens after the connects), so it was a no-op; #14749 instead anchors the cache once after the signal connections are made.
- **Adds a regression unit test** (`_testSpecifiesCoordinateChanged`) that fails on master and passes with the fix.

Appreciate the report and the analysis — the screenshots made the repro easy to confirm.

PR #14753, opened 2026-07-31
fix(pre-commit): install vale-sync from the cmd/vale package

TSC21:
[RIIS](https://riis.com)

<br clear="all">

---

## CI/Build Changes

Two commits to the pre-commit configuration:

1. The `vale-sync` hook installs its Go dependency from `github.com/errata-ai/vale/v3/cmd/vale` instead of the module root. The module root does not contain a Go package (the vale binary lives under `cmd/vale`), so `go install` fails and pre-commit aborts the entire run before a single hook fires.
2. Three inline bash hook entries (`check-categorized-logging`, `check-no-qtest-ignore-message`, `check-no-fixed-qwait`) are folded into YAML block scalars because they run past the 200 column limit that yamllint enforces, so the config was failing its own gate as soon as the file gets linted. YAML folds the lines with single spaces, so the entries parse to exactly the same command as before.

## Reason

Right now the whole pre-commit gate is silently dead: the broken `vale-sync` environment install makes pre-commit exit with code 3 and zero hook verdicts, everywhere this config runs, CI included. The error is:

```
go: module github.com/errata-ai/vale/v3@v3.14.2 found, but does not contain package github.com/errata-ai/vale/v3
```

So clang-format, typos, qmllint, actionlint and everything else in the config have not actually been running. I found this while chasing a pre-commit results artifact that reported `exit_code: 3, passed: 0, failed: 0` on a fork. There is a companion PR that makes the workflow fail loudly when this class of breakage happens again.

## Testing

- [x] Tested workflow locally (ran the pre-commit runner itself with this config: it installs the environment and produces hook verdicts again)
- [x] Verified YAML syntax (check-yaml and yamllint pass on the file)
- [x] Tested on fork before submitting (the same fix is running on my fork, and each folded hook was checked to still exit 1 on a synthetic violation and 0 on a clean file)

## Impact

The pre-commit workflow on every PR, and every local `pre-commit run`. No workflow files change in this PR.

## Checklist

- [x] Workflow permissions follow least-privilege principle (no workflow changes here)
- [x] No secrets are exposed in logs

---

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -152,8 +152,11 @@ repos:
       - id: vale-sync
         name: vale sync (download styles)
         language: golang
+        # The vale module root carries no Go package (the binary lives under cmd/vale), so a go
+        # install of the bare module fails and aborts the whole pre-commit run before any hook
+        # fires.
         additional_dependencies:
-          - 'github.com/errata-ai/vale/v3@v3.14.2'
+          - 'github.com/errata-ai/vale/v3/cmd/vale@v3.14.2'
         entry: vale sync
         files: ^(docs/|\.vale\.ini$)
         pass_filenames: false
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -227,14 +227,20 @@ repos:
 
       - id: check-categorized-logging
         name: Check for category function passed to uncategorized logging macro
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
@TSC21 Yikes, this was nasty. Thanks so much for finding/fixing.

DonLakeFlyer:
One request before merge: please align the vale version pins while you're in here. `vale-sync` installs `github.com/errata-ai/vale/v3/cmd/vale@v3.14.2`, but the `vale` lint hook below it uses repo rev `v3.15.1`, so the styles are synced with a different vale version than the one that lints. It's pre-existing drift rather than something this PR introduced, but since this PR is already touching the `vale-sync` dependency line, bumping it to `@v3.15.1` keeps the two in lockstep.

DonLakeFlyer:
> One request before merge: please align the vale version pins while you're in here.

@TSC21 Don't worry about this, I'll create a separate pull for that.

PR #14755, opened 2026-07-31
fix(CustomPlugin): guard the QML engine pointer through shutdown

TSC21:
[RIIS](https://riis.com)

<br clear="all">

---

## Bug Description

Every exit of a build carrying the custom plugin (`QGC_CUSTOM_DIR=custom-example`, or any downstream custom build based on the template) ends in a SIGSEGV during shutdown. The app's work completes normally and the process then crashes on the way out. It also fails core tests that drive the full application lifecycle when they run against a custom build: MissionCommandTreeEditorTest and VehicleLinkManagerTest pass all their cases and then die in the shutdown crash, which CTest reports as a failure.

## Root Cause

The application owns the QML engine and destroys it before the plugin, so shutdown reaches `CustomPlugin::cleanup()` with `_qmlEngine` pointing at a destroyed engine, and `removeUrlInterceptor()` is called through a raw pointer into freed memory. The existing null check does not help because the pointer never becomes null. gdb backtrace:

```
#0 QQmlEngine::removeUrlInterceptor(QQmlAbstractUrlInterceptor*)
#1 CustomPlugin::cleanup() at custom-example/src/CustomPlugin.cc:60
#2 QGCApplication::shutdown()
#3 main()
```

## Solution

Hold the engine as a `QPointer<QQmlApplicationEngine>`. It nulls itself when the engine is destroyed, so the existing guard in `cleanup()` finally answers the question it asks. No behaviour change on any path where the engine is still alive.

## Testing

- [x] Tested locally (custom build exits cleanly; MissionCommandTreeEditorTest and VehicleLinkManagerTest go from shutdown-crash to passing)
- [ ] Added regression test (the two existing full-lifecycle tests above already catch this whenever they run against a custom build)
- [ ] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [ ] ArduPilot
- [x] N/A

## Checklist

- [x] I have added a test that reproduces the bug (it crashes in removeUrlInterceptor without the fix)

---

### DIFF
--- a/custom-example/CMakeLists.txt
+++ b/custom-example/CMakeLists.txt
@@ -152,3 +152,10 @@ set(CUSTOM_QT_COMPONENTS
     Core
     CACHE INTERNAL "" FORCE
 )
+
+# Compiled into the application target and registered with CTest, but only when testing is
+# enabled, so a production overlay build carries no test code. The subdirectory appends to the
+# same CUSTOM_SOURCES cache src/CMakeLists.txt hands to the application target.
+if(QGC_BUILD_TESTING AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/test/CMakeLists.txt")
+    add_subdirectory(test)
+endif()
--- a/custom-example/src/CustomPlugin.cc
+++ b/custom-example/src/CustomPlugin.cc
@@ -52,7 +52,10 @@ void CustomPlugin::cleanup()
         _qmlEngine->removeUrlInterceptor(_selector);
     }
 
+    // Nulled so a second cleanup is a no-op: shutdown owns one call, and the regression test that
+    // pins the engine guard above exercises the call twice on purpose.
     delete _selector;
+    _selector = nullptr;
 }
 
 void CustomPlugin::_advancedChanged(bool changed)
--- a/custom-example/src/CustomPlugin.h
+++ b/custom-example/src/CustomPlugin.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <QtCore/QPointer>
 #include <QtCore/QTranslator>
 #include <QtQml/QQmlAbstractUrlInterceptor>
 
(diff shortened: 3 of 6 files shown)


TSC21:
Added a regression test on top of the fix: CustomPluginTest drives the real engine creation path, deletes the engine the way shutdown ordering does, and calls cleanup() twice. Without the QPointer guard the first cleanup dies inside QQmlEngine::removeUrlInterceptor, so the case reproduces the crash exactly; the second call pins cleanup as re-entrant, with the interceptor handle nulled after delete. The test rides the custom-example CUSTOM_SOURCES hook and only compiles under QGC_BUILD_TESTING. Verified green on a Linux custom build (Qt 6.11.1): the case passes in 0.5 s and the crash returns if the guard is reverted.

DonLakeFlyer:
> Every exit of a build carrying the custom plugin (`QGC_CUSTOM_DIR=custom-example`, or any downstream custom build based on the template) ends in a SIGSEGV during shutdown.

Are you sure about that. This seems to be a unit test only problem?

DonLakeFlyer:
So the fundamental problem is the fact that QGCCorePlugin::createQmlApplicationEngine is not symmetric. There needs to be a corresponding destroy:
```
    /// Allows the plugin to override or get access to the QmlApplicationEngine to do things like add import
    /// path or stuff things into the context prior to window creation.
    virtual QQmlApplicationEngine *createQmlApplicationEngine(QObject *parent);

    /// Symmetric counterpart to createQmlApplicationEngine. Engines obtained from the create hook
    /// must be destroyed through this hook so the plugin can release any per-engine state it
    /// attached at creation (url interceptors, etc) before the engine goes away.
    virtual void destroyQmlApplicationEngine(QQmlApplicationEngine *qmlEngine);
```

I'm working on a replacement pull for this problem which fixes this up as well as tries to get UTs to run on a custom build since that exercises the problem. Whereas the UT in this pull is never run in CI.

DonLakeFlyer:
Replaced by a number of other reference pulls. @TSC21 Thanks so much for identifying this problem and providing a possible fix.

PR #14767, opened 2026-08-03
Implement pip bootstrap in Python virtual environment

bytesByHarsh:
Added a function to ensure pip is installed in a Python virtual environment

Primarily if the virtual environment is managed via `uv`.

### DIFF
--- a/cmake/modules/PythonVenv.cmake
+++ b/cmake/modules/PythonVenv.cmake
@@ -6,6 +6,31 @@ else()
     set(_qgc_venv_python "${CMAKE_SOURCE_DIR}/.venv/bin/python")
 endif()
 
+
+function(_qgc_ensure_pip _py)
+    execute_process(
+        COMMAND "${_py}" -m pip -V
+        RESULT_VARIABLE _pip_result
+        OUTPUT_QUIET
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I'm not sure I want some sort of hybrid support for an externally created environment. I'd rather just have a clear failure when it doesn't work. To me it just creates additional maintenance surface area.

DonLakeFlyer:
Thanks @bytesByHarsh for identifying this problem — the failure mode is real and worth fixing.

We've decided to take a different approach in #14784: since QGC's own tooling always creates `.venv` with pip seeded (`uv venv --seed`), a pip-less venv can only come from outside QGC's bootstrap. Rather than having CMake silently install pip into a venv it didn't create, configure now fails fast with a clear, actionable message. Closing this in favor of that replacement.

If you have strong disagreement with this direction, please comment here on why — happy to discuss.

DonLakeFlyer:
In reality the idea what QGC would automatically fix an environment which it didn't create itself doesn't seem like the greatest idea.

bytesByHarsh:
Yes, agreed! #14784 is better.

PR #14774, opened 2026-08-07
fix(Vehicle): use standard PX4 airspeed calibration value

msli-dev:
## Description

Fix PX4 airspeed calibration to use the MAVLink-defined value `2` for parameter 6 of `MAV_CMD_PREFLIGHT_CALIBRATION`.

The previous value, `1`, is deprecated for PX4 airspeed calibration. A regression test was added to verify the complete command payload.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] CI/Build changes
- [ ] Other

## Testing

- [x] Added/updated unit tests
- [x] Verified the modified source and test files compile
- [x] Ran `git diff --check`
- [ ] Tested with PX4 SITL
- [ ] Tested with hardware

The final executable link could not complete because a running `MGC.exe` instance held the output file open.

### Platforms Tested

- [x] Windows
- [ ] Linux
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [x] PX4
- [ ] ArduPilot

## Checklist

- [x] The change follows the project coding standards
- [x] A regression test verifies `param6` is set to `2`
- [x] The change is limited to PX4 airspeed calibration
- [ ] Full unit test suite passes locally

## Related Issues

None.

Release-Note: Fix PX4 airspeed calibration command compatibility.

### DIFF
--- a/src/Vehicle/Vehicle.cc
+++ b/src/Vehicle/Vehicle.cc
@@ -2364,7 +2364,7 @@ void Vehicle::startCalibration(QGCMAVLink::CalibrationType calType)
         param7 = 1;
         break;
     case QGCMAVLink::CalibrationPX4Airspeed:
-        param6 = 1;
+        param6 = 2;
         break;
     case QGCMAVLink::CalibrationPX4Pressure:
         param3 = 1;
--- a/test/Vehicle/SendMavCommandWithHandlerTest.cc
+++ b/test/Vehicle/SendMavCommandWithHandlerTest.cc
@@ -136,4 +136,28 @@ void SendMavCommandWithHandlerTest::_compIdAllFailure()
     QCOMPARE(_mockLink->receivedMavCommandCount(testCase.command), testCase.expectedSendCount);
 }
 
+void SendMavCommandWithHandlerTest::_px4AirspeedCalibrationCommand()
+{
+    _mockLink->clearReceivedMavCommandCounts();
+    _mockLink->clearReceivedMavlinkMessageCounts();
+
+    _vehicle->startCalibration(QGCMAVLink::CalibrationPX4Airspeed);
+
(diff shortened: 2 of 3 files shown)


DonLakeFlyer:
Deprecated when? What version of firmware?

msli-dev:
> Deprecated when? What version of firmware?

https://mavlink.io/en/messages/common.html#MAV_CMD_PREFLIGHT_CALIBRATION

DonLakeFlyer:
Looked into whether PX4 changed behavior here. Findings from current PX4 main:

- PX4's `Commander::handle_command()` accepts **both** values for airspeed calibration: `else if ((int)(cmd.param6) == 1 || (int)(cmd.param6) == 2)` — with the note `// TODO: param6 == 1 is deprecated, but we still accept it for a while (feb 2017)` ([Commander.cpp](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/Commander.cpp)).
- PX4's own `commander calibrate airspeed` CLI sends `param6 = 2`.

So nothing has changed in PX4 along the way — `param6 = 1` has worked (and still works) against every PX4 release since the deprecation notice in Feb 2017. This PR is correct per the MAVLink spec (`MAV_CMD_PREFLIGHT_CALIBRATION` defines 2 for airspeed) and aligns QGC with PX4's own usage, but it is a spec-compliance cleanup rather than a fix for an actual breakage. It also removes the risk if PX4 ever follows through on dropping the deprecated value.

Compatibility note: since PX4 has accepted `2` alongside `1` since 2017, this change is safe against any remotely recent PX4 firmware.

DonLakeFlyer:
Thanks @tustin-dev for finding this and submitting the fix! We've picked it up in replacement PRs #14780 (master) and #14781 (Stable_V5.1) so it can land on both branches — closing this one in favor of those. Your contribution is much appreciated.

PR #14797, opened 2026-08-09
fix(Joystick): match actual update rate to joystick setting

bosism:
## Bug Description

The joystick's actual axis update rate can be lower than the value configured in **Axis frequency**. With the default 25 Hz setting, the configured interval is 40 ms, but an update at exactly 40 ms was rejected.

## Root Cause

The polling thread checks joystick state approximately every 20 ms. `_handleAxis()` returned while the elapsed time was *less than or equal to* the configured interval. A poll at exactly 40 ms was therefore skipped and the update could be delayed until the next poll, around 60 ms.

## Solution

Treat the configured interval as inclusive: an axis update is due when `elapsed >= delay`. A deterministic regression test covers the boundary at 39, 40, and 41 ms.

## Testing

- [x] Tested locally
- [x] Added regression test
- [ ] Tested with hardware

Validation performed:

- Debug build completed successfully.
- All joystick-labelled unit tests pass (4/4).
- Relevant repository policy hooks and changed-line clang-format checks pass.
- The full Unit label passes 193/194 tests. The unrelated `BluetoothWorkerTest` fails because the host BlueZ backend emits an unexpected warning for the test's invalid Bluetooth address; the failure reproduces when that test is run alone.

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [ ] ArduPilot
- [x] N/A

## Checklist

- [x] I have added a test that reproduces the bug
- [ ] New and existing unit tests pass locally (193/194; unrelated Bluetooth failure described above)

---

### DIFF
--- a/src/Joystick/Joystick.cc
+++ b/src/Joystick/Joystick.cc
@@ -935,7 +935,7 @@ uint16_t Joystick::_adjustRangeToRcOverridePwm(int value, const AxisCalibration_
 void Joystick::_handleAxis()
 {
     const int axisDelay = static_cast<int>(1000.0 / _joystickSettings.axisFrequencyHz()->rawValue().toDouble());
-    if (_axisElapsedTimer.elapsed() <= axisDelay) {
+    if (!_axisUpdateDue(_axisElapsedTimer.elapsed(), axisDelay)) {
         return;
     }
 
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
@bosism Is there are reason this is marked as Draft?

PR #14824, opened 2026-08-11
fix(i18n): make APM sensor calibration strings translatable

coolebyak:
## Summary

- Wrap the `Simple Accelerometer Calibration` label in `qsTr()`.
- Replace the dynamically generated `qsTr()` input for latitude and longitude
  with a stable source string using `%1` and `%2` placeholders.
- Add translator context for the coordinate placeholders.

## Problem

The calibration label is currently hard-coded in English.

The coordinate string is interpolated before it is passed to `qsTr()`, so the
runtime value cannot match a stable key in the translation catalog.

## Test plan

- Verified that only `APMSensorsComponent.qml` is modified.
- Ran `git diff --check`.
- Verified by code inspection that both coordinate values are preserved.

## Additional context

Both issues were confirmed against the current `master` branch.

### DIFF
--- a/src/AutoPilotPlugins/APM/APMSensorsComponent.qml
+++ b/src/AutoPilotPlugins/APM/APMSensorsComponent.qml
@@ -506,7 +506,7 @@ SetupPage {
                             }
 
                             QGCCheckBox {
-                                text: "Simple Accelerometer Calibration"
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:

PR #14826, opened 2026-08-11
fix(i18n): make custom ArduSub frame names translatable

coolebyak:
## Summary

- Make the custom ArduSub frame name translatable.
- Use a separate translatable string for an unknown frame configuration.

## Problem

`Custom` and `Unknown` are returned as plain strings from `frameName()`, so
they cannot be localized.

A separate `Unknown frame` source string avoids sharing the generic `Unknown`
translation already used for firmware versions and Git hashes in this file.

## Test plan

- Ran `git diff --check`.
- Verified that only `APMSubFrameComponentSummary.qml` is modified.
- Verified the `FRAME_CONFIG = 7` and fallback branches by code inspection.


DonLakeFlyer:

PR #14827, opened 2026-08-11
fix(i18n): make APM goto location error translatable

coolebyak:
## Summary

- Make the APM guided-mode goto-location error message translatable.

## Problem

When the vehicle position is unknown, the user-facing error message is created
with `QStringLiteral()` and therefore bypasses Qt translation.

## Test plan

- Ran `git diff --check`.
- Verified that only `APMFirmwarePlugin.cc` is modified.
- Verified by code inspection that the error condition and displayed English
  fallback text are unchanged.


DonLakeFlyer:

PR #14832, opened 2026-08-11
fix(Settings): correct descent speed description

coolebyak:
## Summary

- Correct the description of the offline editing descent speed setting.

## Problem

The `offlineEditingDescentSpeed` setting currently describes its value as a
`cruising speed`, although it is used as the multi-rotor descent speed when
calculating mission duration.

## Test plan

- Validated `App.SettingsGroup.json`.
- Ran `git diff --check`.
- Verified that only `App.SettingsGroup.json` is modified.
- Verified that the setting name, type, limits, and default value are unchanged.


DonLakeFlyer:

PR #14834, opened 2026-08-11
fix(i18n): make transect editor fallback translatable

coolebyak:
## Summary

- Make the transect editor's `Internal Error` fallback text translatable.

## Problem

The fallback value is displayed to users when a derived transect editor does
not provide its help text or values header. It is currently a plain QML string
and bypasses Qt translation.

## Test plan

- Ran `git diff --check`.
- Verified that only `TransectStyleComplexItemEditor.qml` is modified.
- Verified that the fallback English text and property behavior are unchanged.


DonLakeFlyer:

PR #14835, opened 2026-08-11
fix(i18n): make unsupported vehicle command errors translatable

coolebyak:
## Summary

- Make unsupported Orbit, ROI, and Pause command messages translatable.

## Problem

These user-facing errors are currently created with `QStringLiteral()` and
therefore bypass Qt translation.

The three unique messages occur at four call sites because the ROI error is
used in two code paths.

## Test plan

- Ran `git diff --check`.
- Verified that only `Vehicle.cc` is modified.
- Verified that all four affected call sites now use `Vehicle::tr()`.
- Verified that the message text and command handling logic are unchanged.


DonLakeFlyer:

PR #14836, opened 2026-08-11
fix(i18n): make ADSB server errors translatable

coolebyak:
## Summary

- Make ADSB server error messages translatable.
- Keep the newline separator outside the translated text.

## Problem

The ADSB server error and disabled notification are user-facing strings, but
they are currently created without `tr()` and therefore bypass Qt translation.

## Test plan

- Ran `git diff --check`.
- Verified that only `ADSBVehicleManager.cc` is modified.
- Verified that the `%1` placeholder and error text are preserved.
- Verified that the disabled notification remains separated by a newline.


DonLakeFlyer:

PR #14837, opened 2026-08-11
fix(i18n): make custom flight mode fallback translatable

coolebyak:
## Summary

- Make the fallback name for an unknown custom flight mode translatable.

## Problem

When a custom flight mode value is not present in the mode map, QGroundControl
displays `Custom:0x%1`. This user-facing fallback is currently created with
`QStringLiteral()` and bypasses Qt translation.

## Test plan

- Ran `git diff --check`.
- Verified that only `FirmwarePlugin.cc` is modified.
- Verified that the custom mode value remains formatted as hexadecimal.
- Verified that the fallback English text and lookup behavior are unchanged.


DonLakeFlyer:

PR #14838, opened 2026-08-11
fix(i18n): make firmware upgrade errors translatable

coolebyak:
## Summary

- Make two firmware upgrade error messages translatable.

## Problem

The bootloader synchronization failure and generic upgrade error are shown in
the firmware upgrade status log, but are currently created without `tr()` and
therefore bypass Qt translation.

## Test plan

- Ran `git diff --check`.
- Verified that only `FirmwareUpgradeController.cc` is modified.
- Verified that the `%1` placeholder and supplied error text are preserved.
- Verified that firmware upgrade error handling is unchanged.


DonLakeFlyer:

PR #14839, opened 2026-08-11
fix(i18n): make Survey-In duration unit translatable

coolebyak:
## Summary

- Make the Survey-In duration value and unit translatable.
- Add context for the `%1` duration placeholder.

## Problem

The RTK GPS status page currently appends the seconds unit as a plain QML
literal. Languages that use a different seconds abbreviation or value/unit
ordering cannot localize the complete label.

## Test plan

- Ran `git diff --check`.
- Verified that only `GPSIndicatorPage.qml` is modified.
- Verified that `%1` is replaced with the current Survey-In duration.
- Verified that the English output remains `<duration> s`.


DonLakeFlyer:

PR #14840, opened 2026-08-11
fix(i18n): make gimbal control dialog translatable

coolebyak:
## Summary

- Make the gimbal control request dialog translatable.

## Problem

The dialog title and message are user-facing QML strings, but they are passed
to `showMessageDialog()` without `qsTr()` and therefore bypass Qt translation.

## Test plan

- Ran `git diff --check`.
- Verified that only `GimbalIndicator.qml` is modified.
- Verified that the dialog buttons and callbacks are unchanged.
- Verified that the English fallback text remains unchanged.


DonLakeFlyer:

PR #14895, opened 2026-08-19
fix(PlanManager): resend last mission waypoint in case of final ack timeout

MaEtUgR:
## Problem
If the final mission ack is lost on the link the user gets the message "Mission write failed, vehicle failed to send final ack." even though the mission was uploaded succesfully. Based on the message he'll manually reupload the entire mission again.

## Solution
[PX4 does resend the final ack upon receiving the last waypoint again](https://github.com/PX4/PX4-Autopilot/blob/cda3c3603f0b0c52e878ed1852c1a80131bddf01/src/modules/mavlink/mavlink_mission.cpp#L1164). So in case the last mission ack times out, we should not immediately give up but retransmit the last waypoint like we do if a waypoint or an ack in the middle of the mission transfer goes missing.

## Type of Change
- Bug fix (non-breaking change that fixes an issue)

## Testing

- [ ] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [ ] ArduPilot

## Checklist

- [x] I have added tests that prove my fix/feature works

---

### DIFF
--- a/src/Comms/MockLink/MockLinkMissionItemHandler.cc
+++ b/src/Comms/MockLink/MockLinkMissionItemHandler.cc
@@ -415,6 +415,12 @@ void MockLinkMissionItemHandler::_handleMissionItem(const mavlink_message_t &msg
         return;
     }
 
+    if (_failureMode == FailWriteFinalAckFirstResponse && _failWriteFinalAckFirstResponse) {
+        _failWriteFinalAckFirstResponse = false;
+        qCDebug(MockLinkMissionItemHandlerLog) << "not sending final ack due to failure mode FailWriteFinalAckFirstResponse, first response";
+        return;
+    }
+
     if (_failureMode != FailWriteFinalAckNoResponse) {
         MAV_MISSION_RESULT ack = MAV_MISSION_ACCEPTED;
 
--- a/src/Comms/MockLink/MockLinkMissionItemHandler.h
+++ b/src/Comms/MockLink/MockLinkMissionItemHandler.h
@@ -43,6 +43,7 @@ public:
         FailWriteRequest0ErrorAck,          // Instead of sending MISSION_REQUEST 0, send MISSION_ACK error
         FailWriteRequest1ErrorAck,          // Instead of sending MISSION_REQUEST 1, send MISSION_ACK error
         FailWriteFinalAckNoResponse,        // Don't send the final MISSION_ACK
+        FailWriteFinalAckFirstResponse,     // Don't send the final MISSION_ACK the first time, send it if the last item is sent again
         FailWriteFinalAckErrorAck,          // Send an error as the final MISSION_ACK
         FailWriteFinalAckMissingRequests,   // Send the MISSION_ACK before all items have been requested
     };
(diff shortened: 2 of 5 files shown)


DonLakeFlyer:
FYI: I traced how ArduPilot reacts to this retry mechanism, since the recovery relies on the vehicle re-acking a duplicate final item:

**ArduPilot does not resend the final ack.** When it sends the final `MISSION_ACK`, `MissionItemProtocol::transfer_is_complete()` immediately tears down the upload session (`receiving = false`, `link = nullptr`). When the retransmitted last item then arrives, the dispatch layer in `GCS_MAVLINK::handle_mission_item()` hits:

```cpp
if (!prot->receiving) {
    send_mission_ack(msg, type, MAV_MISSION_ERROR);
    return;
}
```

So ArduPilot replies with a `MISSION_ACK` of `MAV_MISSION_ERROR` (not `MAV_MISSION_INVALID_SEQUENCE`, so the existing APM ignore special-case in `_handleMissionAck` doesn't apply). QGC then fails the transaction with a `VehicleAckError`.

**Net effect on ArduPilot:** same false "write failed" outcome as today when the final ack is lost (the mission is actually stored fine), just reached via one retransmit and an error ack instead of a timeout. No regression and no added delay — but the recovery in this PR only benefits PX4. Per the MAVLink mission protocol spec the vehicle is supposed to resend the ack in this situation, so this is arguably an ArduPilot compliance gap worth an upstream issue.

DonLakeFlyer:
@hamishwillee The spec for mission upload seems silent on what to do if the GCS doesn't receive the final item ACK. With this change PX4 kinda makes an assumption on that which ArduPilot doesn't also assume.

DonLakeFlyer:
> like we do if a waypoint or an ack in the middle of the mission transfer goes missing.

But that isn't really how upload works. The vehicle side is in control. The GCS side does no retry if it's in the middle I believe. The protocol doesn't really work that way.

a-bernasconi:
### Test in SIH

In order to test if the implementation fixes the issue I ran 3 tests using SIH.

## 1. AMC 2.3.4 Dropping 1 ACK

https://github.com/user-attachments/assets/9c8dea9c-0874-49a3-9b82-fd1ed76164b2


## 2. AMC 2.3.4 Dropping 0 ACK (check if the script is not causing the issue)

https://github.com/user-attachments/assets/a5dde484-0d2e-4f72-a294-75b7faf9382f



## 3. FIX dropping 2 ACK and 99 ACK

https://github.com/user-attachments/assets/5cda25b9-3248-4812-be9b-ca31aebab9f6

### Results

As expected the AMC 2.3.4 returns a n error even if just one ACK is lost. With the fix AMC works even if ACK gets lost and returns an appropriate error message when too many ACK are lost

In order to block the ACK message the following file was used (Generated by Claude Code)

[drop_final_ack.py](https://github.com/user-attachments/files/32464456/drop_final_ack.py)

PR #14964, opened 2026-08-26
Complete Simplified Chinese (zh_CN) translation

shengdeng2016:
This PR completes the Simplified Chinese (zh_CN) localization for the v5.1.3 translation set (qgc_source_zh_CN.ts and qgc_json_zh_CN.ts). All entries translated; terminology normalized; Chinese punctuation replaced with ASCII to avoid known zh-locale parsing bugs (e.g. mission planning). Note: the translation work was performed by an AI assistant (Qwen) with human review of key terms.


DonLakeFlyer:
Thanks, I was going to try something similar myself. It does require round tripping translations back up to crowdin though which is currently not supported. I'm going to need to figure out how to turn that one before this can be merged. Otherwise the next crowdin dump will just trash this.

hamishwillee:
> Thanks, I was going to try something similar myself. It does require round tripping translations back up to crowdin though which is currently not supported. I'm going to need to figure out how to turn that one before this can be merged. Otherwise the next crowdin dump will just trash this.

Crowdin can import ts files, so you should be able to do to import this, if you trust this translation. The problem is that it is hard to undo once you have the strings in crowdin.

PR #14976, opened 2026-08-27
fix(dialog): label vehicle reboot prompt Reboot/Continue

dakejahl:
## Summary
Rename the vehicle-reboot-required dialog buttons from Ok/Cancel to Reboot/Continue.

## Problem
Ok is the highlighted action and reboots the vehicle. Configuring several reboot-required parameters makes it easy to reboot after each change.

## Solution
Keep the Ok/Cancel roles, override the labels to Reboot and Continue, and match the prompt text. Continue dismisses; Reboot still reboots.

### DIFF
--- a/src/MainWindow/MainWindow.qml
+++ b/src/MainWindow/MainWindow.qml
@@ -206,17 +206,22 @@ ApplicationWindow {
         _showMessageDialogWorker(mainWindow, dialogTitle, dialogText)
     }
 
-    // This variant is only meant to be called by QGCApplication. Ok reboots the active vehicle.
+    // This variant is only meant to be called by QGCApplication. Reboot reboots the active vehicle; Continue dismisses.
     function _showRebootVehicleDialog(dialogTitle, dialogText) {
-        _showMessageDialogWorker(mainWindow, dialogTitle,
-                                 dialogText + " " + qsTr("Click Ok to reboot the vehicle now."),
-                                 Dialog.Ok | Dialog.Cancel,
-                                 function() {
-                                     const activeVehicle = QGroundControl.multiVehicleManager.activeVehicle
-                                     if (activeVehicle) {
-                                         activeVehicle.rebootVehicle()
-                                     }
-                                 })
+        let dialog = simpleMessageDialogComponent.createObject(mainWindow, {
+            title: dialogTitle,
+            text: dialogText + " " + qsTr("Click Reboot to reboot the vehicle now."),
+            buttons: Dialog.Ok | Dialog.Cancel,
+            acceptButtonText: qsTr("Reboot"),
+            rejectButtonText: qsTr("Continue"),
+            acceptFunction: function() {
+                const activeVehicle = QGroundControl.multiVehicleManager.activeVehicle
+                if (activeVehicle) {
+                    activeVehicle.rebootVehicle()
+                }
+            }
+        })
+        dialog.open()
     }
 
     Connections {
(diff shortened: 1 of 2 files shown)


src/MainWindow/MainWindow.qml
```diff
@@ -206,17 +206,22 @@ ApplicationWindow {
...
-                                     const activeVehicle = QGroundControl.multiVehicleManager.activeVehicle
-                                     if (activeVehicle) {
-                                         activeVehicle.rebootVehicle()
-                                     }
-                                 })
+        let dialog = simpleMessageDialogComponent.createObject(mainWindow, {
+            title: dialogTitle,
+            text: dialogText + " " + qsTr("Click Reboot to reboot the vehicle now."),
+            buttons: Dialog.Ok | Dialog.Cancel,
+            acceptButtonText: qsTr("Reboot"),
```
DonLakeFlyer, line 215: Huh! I never thought of just doing that. I think it works...

DonLakeFlyer: Oh, wait you modded the dialog code as well. I see.

fallenmi (approved):
Reviewed exact head `169e4c3204284421eedc94dcc6d83f17978b9798`.

The custom labels preserve the existing dialog semantics: the reboot prompt still uses `Dialog.Ok | Dialog.Cancel`, `Reboot` remains the accepted action, and `Continue` remains the rejected/dismiss action. The accepted path still null-checks the active vehicle before calling `rebootVehicle()`, while the Cancel role keeps Escape dismissal enabled.

Verification:
- A focused exact-source QML/JavaScript oracle passed on the submitted head and a conflict-free integration tree against live `master`: custom labels, accept/reject routing, close behavior, and both active-vehicle/null-vehicle callback paths were exercised for 20 iterations across 10 repeated runs.
- `git diff --check` passed.
- Exact-head CI reports 47 successful checks (plus one skipped documentation deploy), including 447 passing tests, Linux ASan+UBSan, platform builds, CodeQL, and grype.

No blocking issues found.

Disclosure: I used OpenAI Codex to assist with inspecting the exact diff, repository instructions and interaction history, checking CI, and constructing the focused oracle. I verified the evidence and own this review.

dakejahl (commented):
What's the verdict @DonLakeFlyer ? Shall we merge it?

DonLakeFlyer:
Yeah, I've been waffling on this. I would much rather have standard buttons names instead of a free for all.

> Ok is the highlighted action and reboots the vehicle. Configuring several reboot-required parameters makes it easy to reboot after each change.

I don't quite see how this change fixes this problem?

DonLakeFlyer:
I think the real problem is that having a vehicle reboot tied to just a dialog no matter what is not a good thing. You could reboot something while it's flying fairly easy I think. This should likely have more scaffolding around it for flying and so forth and then probably use a confirm slider thingy. That way you really have to work hard to do it by accident.

dakejahl:
What if when a param is changed that requires a reboot we flag a "reboot required" somewhere that flashes on the screen or otherwise visually indicates to the user that they need to reboot for their settings to take effect.

And then you could click on this indicator to be presented with a reboot/cancel dialogue

DonLakeFlyer:
Yeah, maybe something like that. I'll create a separate Issue to track this and fix it up in 5.2. I need to noodle on it a bit to come up with something that makes me happy.

DonLakeFlyer:
So I think this should be closed then till I figure something out. I'll try to get to it soon. I'm super mired in keeping on top of 5.1 fixes right now as well as a ton of my own wildlife work taking all my time. Thanks for raising the problem, it's pretty bad/dangerous usability now.

dakejahl:
Sounds good, no rush. Thanks!

PR #14977, opened 2026-08-27
feat(Actuators): per-channel protocol column on PWM outputs

dakejahl:
## Summary
Flatten PWM actuator outputs into one table with protocol as a column.

## Problem
QGC grouped channels by timer and put the protocol dropdown on the group. That matches STM32 (several pins per timer) but not iMXRT, where each FlexPWM submodule is its own timer, so the page became one block per pin.

## Solution
Render every PWM channel as a row and put protocol in a column. Channels that share a timer still share one TIM parameter, so changing any dropdown in the group updates the rest. Timer N headers appear only when a timer has multiple channels. UAVCAN and other non-timer outputs keep the old subgroup layout.
Screenshot from 2026-08-27 11-57-28
Screenshot from 2026-08-27 11-56-44

### DIFF
--- a/src/AutoPilotPlugins/PX4/ActuatorComponent.qml
+++ b/src/AutoPilotPlugins/PX4/ActuatorComponent.qml
@@ -436,12 +436,101 @@ SetupPage {
                             }
 
 
+                            // PWM/timer outputs: one row per channel, protocol as a column.
+                            // Channels that share a timer bind the same primary Fact, so changing
+                            // any protocol dropdown in the group updates the rest.
+                            GridLayout {
+                                id: protocolGrid
+                                visible: selActuatorOutput.actuatorOutput.hasPrimaryProtocol && selActuatorOutput.actuatorOutput.groupsVisible
+                                property var output:            selActuatorOutput.actuatorOutput
+                                property var channelConfigs:    output.tableChannelConfigs
+                                property real groupGap:         ScreenTools.defaultFontPixelHeight / 2
+
+                                rows:       output.tableRowCount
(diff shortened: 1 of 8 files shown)


fallenmi (changes requested):
AI disclosure: I used OpenAI Codex to assist this review; I verified the result against exact head `fce2a675d94040654ee7a1497c8e2fd3b107bf30`.

Requesting changes for one actuator-layout regression: `ActuatorOutput::tableChannelConfigs()` exposes only subgroup 0, and the new flattened grid uses that model for its headers and column count. Each row still emits the full `configInstances` model from its own subgroup, though. The existing actuator fixture demonstrates the mismatch: AUX 1-4 has five per-channel parameters, while AUX 5-6 and AUX 7-8 each have a sixth `DSHOT_FMU_MIN${i}` parameter. Those DShot rows therefore create the visible Min editor in a sixth parameter column for which the grid has no header. The previous per-subgroup grid used each subgroup's own header model and did not have this mismatch.

Please align the flattened schema across all subgroups, for example with a union/keyed mapping, or preserve the per-subgroup layout when schemas differ, and add a fixture-backed test covering AUX 5-8. My exact-source oracle reported no mismatches for the base per-subgroup layout and two for the head flattened layout: `AUX/AUX 5-6: headers=5, cells=6` and `AUX/AUX 7-8: headers=5, cells=6`. Exact-head GitHub checks report 50 successes and one skip.

dakejahl:
I don't think the juice is worth the squeeze here. We're just adding a ton of code for slightly better formatting.

PR #14978, opened 2026-08-27
fix(APMFirmwarePlugin): start mission correctly for already-armed vehicles

DonLakeFlyer:
Replaces #14973 — thanks to @eris-ismajli for finding and diagnosing the underlying bug.

## Problem

An ArduPilot vehicle that was already armed but not flying fell straight through `APMFirmwarePlugin::startMission()` without any action: it skipped the `vehicle->flying()` branch and also skipped the `!vehicle->armed()` block, so neither a mode change nor `MAV_CMD_MISSION_START` was ever issued. This commonly shows up in multi-vehicle mission start, but the bug is per-vehicle.

## How this differs from #14973

#14973 changed the initial condition to `flying() || armed()`, which switches an armed grounded vehicle to Auto mode. That works for Copter (Auto with AUTO_OPTIONS set appropriately), but it isn't what ArduPilot semantics call for on the ground: an armed, grounded copter should receive `MAV_CMD_MISSION_START`, which causes the firmware itself to enter Auto and begin the mission. This PR instead:

- **Flying vehicles** (any type): switch to Auto — same as before.
- **Fixed wing**: set Auto first, then arm if needed. Mode is set before arming because arming a tilt-rotor in Guided would arm it in the dangerous forward-flight position. No `MISSION_START` needed.
- **All other vehicle types**: arm in Guided if disarmed, then send `MAV_CMD_MISSION_START` — including when the vehicle was already armed, which is the case #14973 targeted.

Each failure path now reports a specific error message.

## Test support

To make this testable without SITL, MockLink now simulates ArduPilot more realistically instead of relying on test-only hooks:

- Handles `MAV_CMD_DO_SET_MODE` (preserving the armed flag like real firmware) and `MAV_CMD_MISSION_START` in both COMMAND_LONG and COMMAND_INT paths.
- Derives heartbeat `MAV_STATE` (ACTIVE/STANDBY) from vehicle altitude like real firmware, which drives ArduPilot flying detection in QGC.
- Serves firmware-correct AVAILABLE_MODES lists per APM vehicle type (Copter/Plane/Rover/Sub), mirroring the plugin flight-mode tables so StandardModes streaming yields plugin-identical mode names.

## Testing

New `APMStartMissionTest` (Integration/Vehicle labels) covers the state matrix:

- Flying copter → switches to Auto, no MISSION_START
- Disarmed copter → arms in Guided, then MISSION_START
- Armed grounded copter → MISSION_START sent, mode untouched (the #14973 scenario)
- Disarmed plane → Auto then arm, no MISSION_START
- Armed grounded plane → switches to Auto (fails without the fix — verified TDD-style)

Full `-L Vehicle` ctest sweep passes.

### DIFF
--- a/src/Comms/MockLink/MockLink.cc
+++ b/src/Comms/MockLink/MockLink.cc
@@ -59,6 +59,103 @@ QList<MockLink::FlightMode_t> MockLink::_availableFlightModes = {
     { "MockLink Mode (delayed)",0,                          PX4CustomMode::AUTO_FOLLOW_TARGET,  true,       false},
 };
 
+// The ArduPilot lists below mirror the availableFlightModes lists in the corresponding
+// firmware plugins (names and custom mode values must match) so that StandardModes streaming
+// results in the same mode names the plugins use.
+QList<MockLink::FlightMode_t> MockLink::_apmCopterAvailableFlightModes = {
+    // Mode Name          Standard Mode   Custom Mode     CanBeSet    adv
+    { "Stabilize",        0,              0,              true,       true },
+    { "Acro",             0,              1,              true,       true },
+    { "Altitude Hold",    0,              2,              true,       true },
+    { "Auto",             0,              3,              true,       true },
+    { "Guided",           0,              4,              true,       true },
+    { "Loiter",           0,              5,              true,       true },
+    { "RTL",              0,              6,              true,       true },
+    { "Circle",           0,              7,              true,       true },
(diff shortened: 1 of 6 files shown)


DonLakeFlyer:
Also APMFirmwareMission::startMission was getting really messy with the hacked up if structure. I cleaned that up to improve readability.

DonLakeFlyer:
Tested Start Mission against ArduPilot SITL on all vehicle types.

Issue #14982, opened 2026-08-28
feat(params): key PX4 parameter cache by hash instead of sysid


dakejahl:
## Summary

PX4's `_HASH_CHECK` cache is stored as a single `ParamCache/{sysid}_{compid}.v2` file. Almost every vehicle is sysid 1, so connecting a second airframe overwrites the first. Reconnecting the original vehicle then always hash-misses and does a full download.

## Problem

`ParameterManager::parameterCacheFile()` keys the on-disk cache by MAVLink system and component id. The hash itself already identifies the param set (CRC32 over used non-volatile names and values). Two vehicles with slightly different firmware or tunings, or the same vehicle after a one-param change followed by connecting a different copter, thrash that one slot.

The hash cannot tell QGC *which* params changed. It can tell QGC *whether this exact set has been seen before*. That is enough to skip the download for "I have several drones, I have talked to each of them before."

Component-info files already cache this way (`ComponentInformationManager::_getFileCacheTag` is `{crc}_{type}_{translation}`).

## Solution

Store param-value caches as `ParamCache/{hash}.v2` (small LRU, not unbounded). On `_HASH_CHECK`:

- lookup by the vehicle's hash
- hit: load that blob, `PARAM_SET _HASH_CHECK` to stop any stream, done
- miss: existing full download (FTP `@PARAM/param.pck` once PX4 serves it, else `PARAM_REQUEST_LIST`), then write the new hash entry

Vehicle identity is optional as a secondary index for a stale "last known for this UUID" fallback. It should not be the lookup that decides whether to skip the download.

This is independent of making the miss path faster. A two-param change is still a miss; the miss should be a packed FTP transfer, not a PARAM_VALUE firehose.

DonLakeFlyer:
Ah, ok. Now I see.

PR #14983, opened 2026-08-28
feat(params): download PX4 parameters via MAVFTP after hash miss

dakejahl:
## Summary

On PX4, keep `_HASH_CHECK` first and download `@PARAM/param.pck` on a miss instead of `PARAM_REQUEST_LIST`.

Companion: https://github.com/PX4/PX4-Autopilot/pull/28435. Cache-key follow-up: #14982

## Problem

A hash miss starts the PARAM_VALUE stream. On SiK that transfer often stalls and never completes, so Vehicle Setup stays gated. QGC already has the packed-file client and parser for ArduPilot.

## Solution

Enable the FTP param path for PX4 as well, after the existing hash check. `_HASH_CHECK` is still accepted while FTP is armed so the cache can skip the download. Firmware without the virtual file NAKs `kErrFailErrno`/ENOENT rather than `kErrFailFileNotFound`; FTPManager now reports that as File Not Found so the fallback to `PARAM_REQUEST_LIST` stays immediate instead of arriving two retry timeouts later. A successful packed parse writes the PX4 param cache so the next hash can hit. A retry while the file is still transferring no longer discards the in-flight download. Companion/DroneCAN components are still not in the packed file (same as ArduPilot today).

### DIFF
--- a/src/Comms/MockLink/MockLinkFTP.cc
+++ b/src/Comms/MockLink/MockLinkFTP.cc
@@ -2,6 +2,8 @@
 #include "MockLink.h"
 #include "QGCLoggingCategory.h"
 
+#include <cerrno>
+
 #include <QtCore/QDataStream>
 #include <QtCore/QDir>
 #include <QtCore/QTemporaryFile>
@@ -153,6 +155,10 @@ void MockLinkFTP::_openCommand(uint8_t senderSystemId, uint8_t senderComponentId
     } else if (path == "/parameter.json.xz") {
         tmpFilename = QStringLiteral(":MockLink/Parameter.MetaData.json.xz");
     } else if (path == "@PARAM/param.pck" || path.startsWith("@PARAM/param.pck?")) {
+        if (!_paramPckEnabled) {
+            _sendNakErrno(senderSystemId, senderComponentId, ENOENT, outgoingSeqNumber, MavlinkFTP::kCmdOpenFileRO);
+            return;
+        }
         const bool withDefaults = path.contains(QStringLiteral("withdefaults=1"));
         tmpFilename = _generateParamPck(withDefaults);
     } else if (path.startsWith(QStringLiteral("@MAV_LOG/"))) {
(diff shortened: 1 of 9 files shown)


dakejahl:
Tested on an ARKV6X with PX4 https://github.com/PX4/PX4-Autopilot/pull/28435. SiK is a stock 57600 pair on TELEM1, `SYS_USB_AUTO 0`.

**PX4 main**
- USB: cache miss still starts with `_HASH_CHECK`. Open of `@PARAM/param.pck` NAKs File Not Found; this PR reports that as File Not Found and falls back to `PARAM_REQUEST_LIST` on the first attempt, then loads 1238 parameters from the stream.
- SiK: fallback is `PARAM_REQUEST_LIST`. That stream reached 1000 of 1238 parameters in 90 s and did not finish.

**PX4 #28435**
- USB, cache miss (`--clear-cache`): `_HASH_CHECK`, then FTP of the full set. Open reports 13770 bytes; parse finishes 60 ms after Open.
- USB, second connect (cache left in place): `_HASH_CHECK` hits the PX4 param cache this PR writes from that file. No FTP.
- SiK, pymavlink: packed FTP of all 1238 parameters in 9.4 s (not via QGC).
- SiK, QGC cache miss: `_HASH_CHECK`, then FTP (Open reports 13753 bytes); parse finishes 12 s after Open.

fallenmi (approved):
AI disclosure: I used OpenAI Codex to assist this review; I verified the results locally against exact head `5bf8fe9aa694c386b2b921a70984a37d217f8809`.

Looks good. The PX4 path remains hash-first: a cache hit stops before FTP, a miss downloads the packed file and primes the existing cache format, and ENOENT from firmware without the virtual file disables FTP before immediately returning to `PARAM_REQUEST_LIST`. The in-flight guard also preserves the active transfer and its completion handler when the outer retry timer fires.

I built the exact head with the repository's Qt 6.11.1 / GStreamer 1.28.4 macOS configuration, serial enabled and warnings-as-errors, then ran `HashCheckTest`, `ParameterManagerTest`, `FTPManagerTest`, `InitialConnectTest`, and `InitialConnectPeripheralStartupTest` twice each. All 10 executions passed, including cache hit/miss, hash timeout, packed-file success, ENOENT fallback, manual refresh, and timer-exhaustion paths. Exact-head CI is also green across all 47 reported checks.

PR #14985, opened 2026-08-29
feat(qgc4qgis): add QGIS plugin for QGroundControl mission planning

d-camargo:
QGC4QGIS is a self-contained QGIS plugin that brings QGroundControl's photogrammetry mission planning into QGIS. It lives entirely in the new qgc4qgis/ subfolder and does not touch any C++/QML source, CMake target or build preset - the QGroundControl build is unaffected.

What it does:

- Survey grid generation over a polygon AOI, following the same camera model and grid math as QGC's SurveyComplexItem (camera library, GSD/altitude, front/side overlap, transect angle, turnaround).
- Photo centers and footprints as QGIS layers, plus mission statistics (distance, duration, photo count).
- Export to QGroundControl's native .plan, and to Litchi (.csv, .kml), DJI Fly (WPML .kmz) and KML.
- Automatic terrain: fetches the DEM from the same terrain server QGC queries (terrain-ce.suite.auterion.com, Copernicus DEM), writes it as a GeoTIFF, loads it as a project raster and samples terrain-following altitudes from it.
- QGIS Processing provider so every step is also scriptable/batchable.

Source strings are in English, with a Brazilian Portuguese Qt translation shipped in qgc4qgis/i18n/ (loaded at startup for pt-BR locales) and a Portuguese README (README.pt-BR.md). XML output (.kml/.kmz) is written by a minimal write-only serializer (qgc4qgis/core/xmlwrite.py) instead of the standard library XML parser - kml.py and wpml.py never parse XML, so the parser attack surface is not shipped at all; output bytes are identical. Export, survey grid and DEM download failures are logged to the QGIS log instead of being silenced.

Infrastructure changes are scoped to the plugin: pytest.ini, ruff isort and per-file-ignores entries, ruff-format restricted to qgc4qgis/ and tests/ (so upstream-identical files under tools/ and .github/scripts/ are never reformatted), qgis-install/uninstall/test recipes in the justfile, export-ignore rules for the plugin zip, a .qgis-plugin-ci config, and a .gitignore exception so the compiled .qm is tracked.

Tested with 215 pytest tests covering the geometry, camera, survey and terrain math, the plan/WPML/Litchi/KML writers, the XML serializer, the dock widget and the Processing algorithms; ruff clean.

## Type of Change

- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [ ] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [ ] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [ ] PX4
- [ ] ArduPilot

## Checklist

## Related Issues

---


DonLakeFlyer:
> It lives entirely in the new qgc4qgis/ subfolder and does not touch any C++/QML source, CMake target or build preset

Which confuses me as to why this would be in the repo then? Heck of a lot of stuff to throw over the wall with no explanation of how it works, how you use it, what it does, so many questions! Seems like a complete python rewrite of a pile of internal QGC code. Why should I be interested in taking this as a pull?

d-camargo:
Fair question, let me answer it directly, then propose what I think is the right outcome.

**What it is:** a QGIS (https://github.com/qgis/QGIS) plugin that generates photogrammetry survey grids over a polygon AOI drawn in QGIS, previews photo centers and footprints as layers, and exports QGC's native `.plan` (plus Litchi, DJI WPML and KML). The audience is people whose AOIs, DEMs and cadastral layers already live in QGIS and who today redraw that polygon by hand in QGC.

**And yes, it is a Python reimplementation** of the SurveyComplexItem camera/grid math. That isn't accidental, and it isn't something I would ask you to maintain: a QGIS plugin runs inside QGIS's own Python, so it can't link QGC's C++. Matching the math was the only way to make the exported `.plan` behave in QGC the way the preview does in QGIS.

**Which leaves the real risk you're pointing at: silent drift.** A reimplementation that quietly falls behind QGC produces plans that look right and fly wrong, and you'd get the bug reports. That is a tractable problem, and it's mine to solve, with a scheduled workflow in my repo that watches this one:
    - `src/Camera/CameraMetaData.json`: the plugin ships a byte-identical copy of it today   (`cmp` is clean). The workflow opens a PR on my side whenever yours changes, so the   camera list can never diverge.
    - The accepted survey version range in `SurveyComplexItem.cc` (`version < 2 || version > 5`   today; the plugin writes 5) and `kPlanFileVersion` in `PlanMasterController.h` (1, which the plugin mirrors). My CI fails the moment the plugin writes a version QGC no longer accepts.
    - Any commit touching `SurveyComplexItem.cc` or `CameraCalc.cc` opens an issue on my side to re-check the grid math against the golden `.plan` fixtures.
All of it runs in my repository against your public history: no workflow added here, no secrets, no notifications, nothing for you to review or maintain.

**As for why this should be in the repo: it shouldn't.**

**Licensing.** COPYING.md is unambiguous — contributions must be usable under **both** licenses, and "GPL-only code cannot be accepted". This plugin is GPL-3.0, and that wasn't a preference. Its only entry points are `qgis.core`, `qgis.gui` and `qgis.PyQt.*`: QGIS is GPL-2.0-or-later, and `qgis.PyQt` is QGIS's shim over PyQt5/PyQt6, which Riverbank licenses as GPL-3.0-or-later or commercially. The plugin is not a standalone program, it is loaded into the QGIS process and only ever exists as a combined work with those, which lands on GPLv3. plugins.qgis.org requires GPLv2+ from hosted plugins for the same reason.

I could relicense my own source files as Apache-2.0 (the survey math derives from your dual-licensed SurveyComplexItem, so that direction is open), but it would buy QGC nothing. Your Apache-2.0 half exists so QGC can ship in proprietary and app-store builds and, as COPYING.md notes, that path already requires a commercial Qt license. The equivalent path for this plugin would require a commercial PyQt license, which no one distributing through the QGIS plugin ecosystem has. An Apache-2.0 header on these files would be a promise the code cannot keep.

**Release cadence.** It tracks QGIS and Qt versions through the QGIS Plugin Repository, not QGC's build and release cycle. Vendoring it here couples two schedules that have no reason to be coupled.

Staying out of tree is also what makes the sync above honest: I track your master and adapt, rather than freezing a copy of your internals inside your own repository.

**On the missing explanation, sorry, that one is on me.** There is a README (`qgc4qgis/README.md`: install, usage, exporters, terrain), but I buried it in a 71-file diff and left the PR template's Description empty. Nothing about how I presented this made any of it findable, and "thrown over the wall" is a fair description of the result.

So I'm closing this PR. The plugin stays in my own repo and ships through the QGIS Plugin Repository, with the drift checks above wired up. If a pointer from the QGC docs ever seems useful to your users, I'm glad to open a one-line docs PR, but only if you'd want it; nothing needed from you otherwise.

Thanks for taking the time to look at it.

PR #14986, opened 2026-08-29
feat(PlanView): choose PX4 VTOL takeoff mode

alireza787b:
## Feature Description

QGroundControl currently inserts `MAV_CMD_NAV_VTOL_TAKEOFF` whenever the plan
vehicle is a VTOL. PX4 also supports an ordinary `MAV_CMD_NAV_TAKEOFF` for a
VTOL that should start and continue the mission in multicopter mode.

For PX4 VTOL plans, the Takeoff tool now offers two choices:

- **VTOL takeoff** inserts `MAV_CMD_NAV_VTOL_TAKEOFF`. It remains the first and
  primary choice, so the existing workflow does not change.
- **Multicopter takeoff** inserts `MAV_CMD_NAV_TAKEOFF` and keeps the mission in
  multicopter mode until an explicit transition item is reached.

ArduPilot and non-VTOL vehicles keep the existing single-action Takeoff tool.
This PR is limited to QGroundControl mission planning. It does not add a
MAVLink command, change the mission format, require a PX4 change, or include
any video/networking work.

## Implementation Details

- The PX4 firmware plugin advertises the additional VTOL capability. Plan View
  does not contain a firmware-name check.
- The selection belongs to the inserted mission item; there is no global or
  persistent takeoff-mode setting.
- Mission geometry and VTOL mode calculations follow the command that is
  actually stored in the plan.
- The actual multicopter takeoff item gets a distinct name and description;
  ordinary PX4 multicopter takeoff metadata remains unchanged.
- The chooser refreshes when an offline planning vehicle changes firmware or
  vehicle type.
- `MAV_CMD_DO_VTOL_TRANSITION` is unchanged. Its existing target-state selector
  remains the way to transition later in the mission.

## Testing

- [x] Tested locally (Linux, offline PX4 VTOL Plan View)
- [x] Added/updated unit tests
- [ ] Tested with hardware

Manual check:

- Checked the offline PX4 VTOL flow in Plan View on Linux.
- Confirmed both choices are shown, VTOL takeoff is primary, and each choice
  inserts the expected takeoff mission item.

Automated coverage includes:

- default and explicit command insertion;
- unsupported-command rejection without changing the plan;
- QGC JSON and WPL load/save behavior;
- mission geometry and following-waypoint VTOL mode, including recovery from a
  preceding fixed-wing transition;
- unchanged ArduPilot, PX4 multicopter, and PX4 fixed-wing behavior;
- offline vehicle-type changes after Plan View is instantiated;
- item-specific multicopter takeoff name and description.

The full feasible suite is run by the QGroundControl GitHub Actions matrix for
this commit. No SITL or hardware evidence is claimed here.

### Platforms Tested

- [x] Linux
- [x] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [x] PX4 (offline planning)
- [ ] ArduPilot

## Screenshots / Demo

PX4 VTOL takeoff choices

## Compatibility

- Existing PX4 VTOL plans and the default insertion path are unchanged.
- Saved missions use standard MAVLink command IDs and need no migration.
- Firmware that does not advertise the capability does not show the additional
  choice.
- ArduPilot behavior is covered by a focused compatibility test.

## Checklist

- [x] I have added tests that prove my feature works
- [x] Focused and full feasible unit tests pass in GitHub Actions

The full QGC build/test toolchain was not available in the local checkout; the
local validation was limited to static checks and the manual Linux UI flow.

---
By submitting this pull request, I confirm that my contribution is made under
the terms of the project's dual license (Apache 2.0 and GPL v3).

### DIFF
--- a/src/FirmwarePlugin/PX4/PX4-MavCmdInfoVTOL.json
+++ b/src/FirmwarePlugin/PX4/PX4-MavCmdInfoVTOL.json
@@ -4,5 +4,10 @@
     "fileType":  "MavCmdInfo",
 
     "mavCmdInfo": [
+        {
+            "id":           22,
+            "comment":      "MAV_CMD_NAV_TAKEOFF",
+            "description":  "Take off vertically and continue the mission in multicopter mode."
+        }
     ]
 }
--- a/src/FirmwarePlugin/PX4/PX4-MavCmdInfoVTOL.json
+++ b/src/FirmwarePlugin/PX4/PX4-MavCmdInfoVTOL.json
@@ -4,10 +4,5 @@
     "fileType":  "MavCmdInfo",
 
     "mavCmdInfo": [
-        {
-            "id":           22,
-            "comment":      "MAV_CMD_NAV_TAKEOFF",
-            "description":  "Take off vertically and continue the mission in multicopter mode."
-        }
     ]
 }
--- a/src/MissionManager/MissionController.cc
+++ b/src/MissionManager/MissionController.cc
@@ -24,6 +24,7 @@
 #include "TakeoffMissionItem.h"
 #include "PlanViewSettings.h"
 #include "MissionCommandTree.h"
+#include "AppMessages.h"
 #include "QGCMath.h"
 #include "QGCLoggingCategory.h"
 
@@ -71,7 +72,7 @@ MissionController::~MissionController()
 
 void MissionController::_resetMissionFlightStatus(void)
 {
-    _flightStatusCalc.reset(_controllerVehicle, _managerVehicle, _missionContainsVTOLTakeoff);
+    _flightStatusCalc.reset(_controllerVehicle, _managerVehicle, _missionStartsInVTOLMulticopterMode);
     _missionFlightStatus = _flightStatusCalc.status();
 
     emit missionPlannedDistanceChanged(_missionFlightStatus.plannedDistance);
@@ -314,8 +315,32 @@ VisualMissionItem* MissionController::insertSimpleMissionItem(QGeoCoordinate coo
 
 VisualMissionItem* MissionController::insertTakeoffItem(QGeoCoordinate /*coordinate*/, int visualItemIndex, bool makeCurrentItem)
 {
+    return _insertTakeoffItemWorker(
+        _controllerVehicle->vtol() ? MAV_CMD_NAV_VTOL_TAKEOFF : MAV_CMD_NAV_TAKEOFF,
+        visualItemIndex,
+        makeCurrentItem);
+}
+
+VisualMissionItem* MissionController::insertVTOLMulticopterTakeoffItem(
+    QGeoCoordinate /*coordinate*/, int visualItemIndex, bool makeCurrentItem)
+{
+    const FirmwarePlugin* firmwarePlugin = _controllerVehicle->firmwarePlugin();
+    const QList<MAV_CMD> supportedCommands = firmwarePlugin->supportedMissionCommands(_controllerVehicle->vehicleClass());
+    const bool commandSupported = supportedCommands.isEmpty() || supportedCommands.contains(MAV_CMD_NAV_TAKEOFF);
+    if (!_controllerVehicle->vtol()
+        || !firmwarePlugin->isCapable(_controllerVehicle, FirmwarePlugin::VTOLMulticopterTakeoffCapability)
+        || !commandSupported) {
+        QGC::showAppMessage(tr("Multicopter takeoff is not supported for this vehicle and firmware."));
+        return nullptr;
+    }
+
+    return _insertTakeoffItemWorker(MAV_CMD_NAV_TAKEOFF, visualItemIndex, makeCurrentItem);
+}
+
+VisualMissionItem* MissionController::_insertTakeoffItemWorker(MAV_CMD command, int visualItemIndex, bool makeCurrentItem)
+{
     int sequenceNumber = _nextSequenceNumber();
-    _takeoffMissionItem = new TakeoffMissionItem(_controllerVehicle->vtol() ? MAV_CMD_NAV_VTOL_TAKEOFF : MAV_CMD_NAV_TAKEOFF, _masterController, _flyView, _settingsItem, false /* forLoad */);
+    _takeoffMissionItem = new TakeoffMissionItem(command, _masterController, _flyView, _settingsItem, false /* forLoad */);
     _takeoffMissionItem->setSequenceNumber(sequenceNumber);
     _initVisualItem(_takeoffMissionItem);
 
@@ -996,7 +1021,7 @@ void MissionController::_recalcFlightPathSegments(void)
 
     FlightPathSegmentHashTable oldSegmentTable = _flightPathSegmentHashTable;
 
-    _missionContainsVTOLTakeoff = false;
+    _missionStartsInVTOLMulticopterMode = false;
     _flightPathSegmentHashTable.clear();
 
     _simpleFlightPathSegments.beginResetModel();
@@ -1036,14 +1061,19 @@ void MissionController::_recalcFlightPathSegments(void)
             MAV_CMD command = simpleItem->mavCommand();
             switch (command) {
             case MAV_CMD_NAV_TAKEOFF:
+                if (firstCoordinateNotFound
+                    && _controllerVehicle->firmwarePlugin()->isCapable(
+                        _controllerVehicle, FirmwarePlugin::VTOLMulticopterTakeoffCapability)) {
+                    _missionStartsInVTOLMulticopterMode = true;
+                }
+                if (!linkEndToHome && firstCoordinateNotFound) {
+                    linkStartToHome = true;
+                }
+                break;
             case MAV_CMD_NAV_VTOL_TAKEOFF:
-                _missionContainsVTOLTakeoff = command == MAV_CMD_NAV_VTOL_TAKEOFF;
-                if (!linkEndToHome) {
-                    // If we still haven't found the first coordinate item and we hit a takeoff command this means the mission starts from the ground.
-                    // Link the first item back to home to show that.
-                    if (firstCoordinateNotFound) {
-                        linkStartToHome = true;
-                    }
+                if (!linkEndToHome && firstCoordinateNotFound) {
+                    _missionStartsInVTOLMulticopterMode = true;
+                    linkStartToHome = true;
                 }
                 break;
             case MAV_CMD_NAV_RETURN_TO_LAUNCH:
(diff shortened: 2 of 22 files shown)


DonLakeFlyer:
Great stuff. Looking at it now...

DonLakeFlyer (commented):
Checked the premise against PX4 v1.18 navigator/feasibility code: `NAV_CMD_TAKEOFF` on a landed MC-mode VTOL goes CLIMB → waypoint with no transition, and `FeasibilityChecker` accepts it and sets `_has_takeoff`, so the feature is sound. Core logic looks correct; a few items inline.

src/MissionManager/MissionController.cc
```diff
@@ -1036,14 +1061,19 @@ void MissionController::_recalcFlightPathSegments(void)
...
             case MAV_CMD_NAV_VTOL_TAKEOFF:
-                _missionContainsVTOLTakeoff = command == MAV_CMD_NAV_VTOL_TAKEOFF;
-                if (!linkEndToHome) {
-                    // If we still haven't found the first coordinate item and we hit a takeoff command this means the mission starts from the ground.
-                    // Link the first item back to home to show that.
-                    if (firstCoordinateNotFound) {
-                        linkStartToHome = true;
-                    }
+                if (!linkEndToHome && firstCoordinateNotFound) {
+                    _missionStartsInVTOLMulticopterMode = true;
```
DonLakeFlyer, line 1075: Behavior change: `_missionContainsVTOLTakeoff` used to be set for a `NAV_VTOL_TAKEOFF` anywhere in the mission; `_missionStartsInVTOLMulticopterMode` is now only set when `!linkEndToHome && firstCoordinateNotFound`. PX4 rejects takeoff-not-first missions anyway so this is practically harmless, but the two branches are also asymmetric — the `NAV_TAKEOFF` branch above sets the flag regardless of `linkEndToHome`. Please make both branches use the same condition (or note the intended difference).

src/MissionManager/MissionController.cc
```diff
@@ -313,9 +314,33 @@ VisualMissionItem* MissionController::insertSimpleMissionItem(QGeoCoordinate coo
...
+
+VisualMissionItem* MissionController::insertVTOLMulticopterTakeoffItem(
+    QGeoCoordinate /*coordinate*/, int visualItemIndex, bool makeCurrentItem)
+{
+    const FirmwarePlugin* firmwarePlugin = _controllerVehicle->firmwarePlugin();
+    const QList<MAV_CMD> supportedCommands = firmwarePlugin->supportedMissionCommands(_controllerVehicle->vehicleClass());
+    const bool commandSupported = supportedCommands.isEmpty() || supportedCommands.contains(MAV_CMD_NAV_TAKEOFF);
+    if (!_controllerVehicle->vtol()
+        || !firmwarePlugin->isCapable(_controllerVehicle, FirmwarePlugin::VTOLMulticopterTakeoffCapability)
+        || !commandSupported) {
```
DonLakeFlyer, line 332: The QML only exposes the multicopter choice when `supports.vtolMulticopterTakeoff` is true, so reaching this branch is a caller-contract violation, not a user-facing condition. That should be a `qCWarning(MissionControllerLog)` + `return nullptr`, not `QGC::showAppMessage`.

The extra checks are also redundant with `isCapable(VTOLMulticopterTakeoffCapability)`: PX4 only grants the capability when `vehicle->vtol()`, and `PX4FirmwarePlugin::supportedMissionCommands` always includes `MAV_CMD_NAV_TAKEOFF` for `VehicleClassVTOL`. A single `isCapable` check is enough; the `AppMessages.h` include can go and `_testUnsupportedVTOLMulticopterTakeoff` would switch to `expectLogMessage`.

src/MissionManager/MissionController.cc
```diff
@@ -1036,14 +1061,19 @@ void MissionController::_recalcFlightPathSegments(void)
             MAV_CMD command = simpleItem->mavCommand();
             switch (command) {
             case MAV_CMD_NAV_TAKEOFF:
+                if (firstCoordinateNotFound
+                    && _controllerVehicle->firmwarePlugin()->isCapable(
+                        _controllerVehicle, FirmwarePlugin::VTOLMulticopterTakeoffCapability)) {
```
DonLakeFlyer, line 1066: This predicate (VTOL + `MAV_CMD_NAV_TAKEOFF` + `VTOLMulticopterTakeoffCapability`) is hand-rolled three times: here, `MissionFlightStatusCalculator::recalc`, and `TakeoffMissionItem::_initLaunchTakeoffAtSameLocation`. Please pull it into a single helper (e.g. on `TakeoffMissionItem` or `FirmwarePlugin`) so the rule can't drift between the geometry, flight-status and item code.

src/FirmwarePlugin/PX4/PX4-MavCmdInfoVTOL.json
```diff
@@ -4,5 +4,10 @@
     "fileType":  "MavCmdInfo",
 
     "mavCmdInfo": [
+        {
+            "id":           22,
+            "comment":      "MAV_CMD_NAV_TAKEOFF",
+            "description":  "Take off vertically and continue the mission in multicopter mode."
```
DonLakeFlyer, line 10: Only `description` is overridden here, so in the mission item list a PX4 VTOL user sees plain "Takeoff" for the multicopter variant and can't tell it from the fixed-wing meaning until they open the item. Consider also overriding `friendlyName` (e.g. "Multicopter takeoff") so the two takeoff types are distinguishable at a glance.

test/MissionManager/MissionCommandTreeTest.cc
```diff
@@ -176,6 +176,16 @@ void MissionCommandTreeTest::testOverride()
...
 }
 
+void MissionCommandTreeTest::testPX4VTOLTakeoffOverride()
+{
+    MissionCommandList commandList(QStringLiteral(":/json/PX4-MavCmdInfoVTOL.json"), false /* baseCommandList */);
+    const MissionCommandUIInfo* const takeoffInfo = commandList.getUIInfo(MAV_CMD_NAV_TAKEOFF);
+
+    QVERIFY(takeoffInfo);
+    QCOMPARE(takeoffInfo->description(),
+             QStringLiteral("Take off vertically and continue the mission in multicopter mode."));
```
DonLakeFlyer, line 186: This asserts a JSON string literal equals a copy of itself — it exercises no behavior beyond JSON parsing that `testAllTrees` already covers. Suggest dropping it.

src/Vehicle/VehicleSupports.h
```diff
@@ -30,6 +30,7 @@ class VehicleSupports : public QObject
     Q_PROPERTY(bool guidedTakeoffWithAltitude       READ guidedTakeoffWithAltitude      CONSTANT)
     Q_PROPERTY(bool guidedTakeoffWithoutAltitude    READ guidedTakeoffWithoutAltitude   CONSTANT)
     Q_PROPERTY(bool changeHeading                   READ changeHeading                  CONSTANT)
+    Q_PROPERTY(bool vtolMulticopterTakeoff          READ vtolMulticopterTakeoff         CONSTANT)
```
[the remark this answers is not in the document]
DonLakeFlyer: Confirmed this is reachable even though the offline vehicle-type combos are hidden once the plan has items (`PlanInfoEditor._allowFWVehicleTypeSelection` = `_noMissionItemsAdded && !activeVehicle`). The binding only needs one type change after Plan View is instantiated, and that's allowed on an empty plan:

1. Start QGC with offline vehicle class = Multicopter, open Plan View. `dropPanelComponent` evaluates `supports.vtolMulticopterTakeoff` → `false` and, being `CONSTANT`, never re-evaluates.
2. In Plan Info (plan is empty, so the combo is shown) switch vehicle class to VTOL. `_controllerVehicle` is updated in place (`Vehicle.cc` connects the offline facts to `_offlineVehicleTypeSettingChanged`), so `vtol`/`rover` bindings refresh via `vehicleTypeChanged` and the Takeoff button behaves as VTOL.
3. Click Takeoff → `dropPanelComponent` is still `null`, a plain `MAV_CMD_NAV_VTOL_TAKEOFF` is inserted, no chooser.

Reverse order (start VTOL, switch to Multicopter) leaves the panel attached, so a multicopter gets the "VTOL takeoff / Multicopter takeoff" chooser. Loading a `.plan` whose firmware/vehicle fields differ from the current offline settings drives the same in-place update.

`VehicleSupports` already has `terrainFrameChanged` for exactly this reason, so a `NOTIFY vtolMulticopterTakeoffChanged` emitted from `Vehicle::vehicleTypeChanged`/`firmwareTypeChanged` is the consistent fix. The UI test sets the offline facts before `startUI()`, so it won't catch this without a type switch mid-test.

src/FirmwarePlugin/PX4/PX4-MavCmdInfoVTOL.json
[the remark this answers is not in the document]
alireza787b: Addressed in `08338dd03`. I removed the unused VTOL metadata override and now provide the specific name and description from the actual `TakeoffMissionItem` when it represents a supported PX4 VTOL `MAV_CMD_NAV_TAKEOFF`. Putting this in the multirotor metadata would also rename normal PX4 multicopter takeoff items. The focused test now checks the real item rather than the JSON file.

src/Vehicle/VehicleSupports.h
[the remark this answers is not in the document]
alireza787b: Addressed in `08338dd03`. `vtolMulticopterTakeoff` now has a notify signal wired to both firmware and vehicle type changes. The Plan View UI test starts as PX4 Multicopter, changes to VTOL after the view is instantiated, verifies the notification, and then exercises both choices.

src/MissionManager/MissionController.cc
alireza787b: Addressed in `08338dd03`. The ordinary takeoff branch now updates `_missionStartsInVTOLMulticopterMode` inside the same `!linkEndToHome && firstCoordinateNotFound` condition used by the VTOL takeoff branch.

src/MissionManager/MissionController.cc
alireza787b: Addressed in `08338dd03`. Unsupported insertion now uses a categorized warning and returns `nullptr`; the redundant vehicle and supported-command checks and `AppMessages.h` include are gone. The test now uses `expectLogMessage`.

src/MissionManager/MissionController.cc
alireza787b: Addressed in `08338dd03`. The rule is centralized in `TakeoffMissionItem::isVTOLMulticopterTakeoff` and reused by insertion validation, geometry handling, and mission flight-status calculation.

src/FirmwarePlugin/PX4/PX4-MavCmdInfoVTOL.json
alireza787b: Addressed in `08338dd03`. The actual item is now labeled `Multicopter takeoff` and carries the multicopter-mode description. This is item-specific so regular PX4 multicopter takeoff keeps its existing label.

test/MissionManager/MissionCommandTreeTest.cc
alireza787b: Removed in `08338dd03`. Coverage now goes through the actual `TakeoffMissionItem` path.

src/Vehicle/VehicleSupports.h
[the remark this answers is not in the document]
alireza787b: Addressed in `08338dd03`. The property now emits on both `Vehicle::vehicleTypeChanged` and `Vehicle::firmwareTypeChanged`. The QML test changes the offline class from Multicopter to VTOL after Plan View has been created, so it covers the stale-binding case directly.

alireza787b:
I addressed the review in `08338dd03` and the macOS const-correctness follow-up in `9b7da08e2`. The existing VTOL takeoff stays first and primary, while multicopter takeoff is the secondary per-item choice. The capability binding now follows offline vehicle changes, the mode calculation handles an ordinary takeoff after an FW transition, and the shared predicate is used across insertion and mission calculations.

All current Actions checks pass (48 successful, no failures), and GitHub reports the branch mergeable. Ready for another look.

Issue #11017, opened 2024-02-03
Throttle with joystick is only 50% so makes it unable to calibrate


flyasky:
## Expected Behavior
Able to calibrate joystick.

## Current Behavior
Unable to proceed with calibration because QGC sees just 50% of the axis when I move the throttle fully down.

## Steps to Reproduce:
Please provide an unambiguous set of steps to reproduce the current behavior
1. joystick is Radiomaster zorro controller with EdgeTX latest, model with classic joystick and AETR mixes applied. (it could be calibrated in DRL, Lift-Off with this setup).
2. no settings in ~/.mavlink/joysticks
3. run SITL + QGC
4. load module joystick in mavproxy
5. open QGC > Vehicle Setup/ Joystict Setup, try to calibrate (see attached images)

## System Information
When posting bug reports, include the following information
- Operating System: Ubuntu 22.04
- QGC Version: 4.3.0
- QGC build: stable, appimage
- Flight Controller: just simulation
- Autopilot (with version): Ardupilot Copter 4.5

## Detailed Description

pic1 (maximum, left stick is fully up, see axis2/4)
![image](https://github.com/mavlink/qgroundcontrol/assets/213282/122afcd3-a1dd-44ac-bcc4-2eca13951c82)

pic2 (minimum, left stick is fully down, see axis2/4)
![image](https://github.com/mavlink/qgroundcontrol/assets/213282/b3ef6474-a5ca-41e7-ba9e-5afa79ba5e8b)

mtzm:
I experience the same issue with the FrSky Taranis.

After finding this and this [issue](https://github.com/mavlink/qgroundcontrol/issues/10104), the clues lead to the Game Controller interface of SDL2. After running [jstest-sdl](https://github.com/meleu/jstest-sdl.git) i noticed that the joystick inputs are correct when running `jstest-sdl --test 0`, but wrong when running `jstest-sdl --gamecontroller 0`.
Running `jstest-sdl -l` showed following game controler mapping:
`Mapping: '030068cd09120000544f000011010000,OpenTX FrSky Taranis Joystick,a:b0,b:b1,x:b3,y:b4,back:b10,guide:b12,start:b11,leftstick:b13,rightstick:b14,leftshoulder:b6,rightshoulder:b7,leftx:a0,lefty:a1,rightx:**a3**,righty:**a4**,lefttrigger:**a2**,righttrigger:a5,crc:cd68,platform:Linux'`

[...]

PR #15032, opened 2026-09-03
fix(Joystick): fix 0-to-MAX trigger axis calibration and UI preview

M4rkoHR:
- Lower movement delta thresholds during Joystick mode so half-axis (0-to-MAX) triggers auto-detect.
- Update _validateAndAdjustCalibrationValues to prevent resetting valid trigger spans that do not cross 0.
- Stretch half-axis values in _adjustChannelRawValue to display full -MAX to +MAX travel in the QML Attitude Controls preview.

## Description
When calibrating a gamepad trigger or other half-axis control (which physically outputs `0` to `32767` rather than `-32768` to `32767`) mapped to a primary attitude control like Throttle, the calibration workflow previously suffered from three key issues:

1. **Auto-Detect Failure:** Triggers resting at `0` or `16383` failed to trigger automatic stick movement detection during `_inputStickDetect` and `_inputStickMin` because the logic required crossing a hardcoded negative center threshold (`0`). Users were forced to manually click the "Next" button to proceed.
2. **Validation Reset:** At the end of calibration, `_validateAndAdjustCalibrationValues()` reset any non-extension channel whose `channelMin` did not cross a negative lower bound (`_calValidMinValue`). This wiped valid trigger calibrations back to default `-32768` bounds.
3. **UI Preview Clipping:** Because the QML Attitude Controls preview expects a normalized `-32768` to `32767` range, applying non-negative trigger inputs against default bounds caused the preview dot to sit at the middle (`0`) at rest and only move through the upper half of the display.

### Changes Made:
- **`_inputStickDetect` & `_inputStickMin`:** Lowered the movement delta requirement in Joystick mode to `8000` and updated the `_inputStickMin` check to compare movement relative to the captured center value rather than hardcoded `0`.
- **`_validateAndAdjustCalibrationValues`:** Added a `validSpan` check (`channelMax - channelMin > 10000`) so valid half-axis calibrations are retained rather than reset to defaults, and automatically sets `channelTrim` to the true midpoint when a trigger edge-resting trim is detected.
- **`_adjustChannelRawValue`:** Added linear range normalization in Joystick mode to scale physical `0 -> 32767` inputs across the full `-32768 -> 32767` range for proper QML Attitude Controls preview rendering.

## Type of Change

- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [x] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested
- [ ] PX4
- [x] ArduPilot

## Screenshots
Throttle at minimum (channel 5):
image

Throttle at maximum:
image

## Checklist

## Related Issues

#11017 
---

### DIFF
--- a/src/Vehicle/VehicleSetup/RemoteControlCalibrationController.cc
+++ b/src/Vehicle/VehicleSetup/RemoteControlCalibrationController.cc
@@ -389,7 +389,9 @@ void RemoteControlCalibrationController::_setupCurrentState()
 
     _saveCurrentRawValues();
 
-    _nextButton->setEnabled(state.nextButtonFn != nullptr);
+    _nextButton->setEnabled(state.nextButtonFn != nullptr ||
+                        state.channelInputFn == &RemoteControlCalibrationController::_inputStickDetect ||
+                        state.channelInputFn == &RemoteControlCalibrationController::_inputStickMin);
     emit oneSidedButtonVisibleChanged(oneSidedButtonVisible());
 }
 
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
I wonder how this fix interacts with #11017 style problems? Does it just cover over the fact that SDL doesn't work correctly with some joysticks causing other problems. Or do you think it's fully safe to fix it globally this way? My concern is that making this change will gloss over other possible problems.

DonLakeFlyer:
@M4rkoHR Can you look at review and question?

DonLakeFlyer:
I have to look a little more but I think that will allow things like spring loaded triggers to be mapped to attitude controls which is a bad thing.

M4rkoHR:
> I wonder how this fix interacts with #11017 style problems? Does it just cover over the fact that SDL doesn't work correctly with some joysticks causing other problems. Or do you think it's fully safe to fix it globally this way? My concern is that making this change will gloss over other possible problems.

Issue #11017 is actually the exact problem I'm trying to solve here. SDL frequently misidentifies standard sticks on EdgeTX and older OpenTX radios as 0-to-MAX trigger channels over USB. Even switching to Advanced USB mode in EdgeTX doesn't always fix it.
So rather than glossing over the SDL bug, this fix gracefully handles the reality of how SDL exposes these legitimate HID RC radios. Without it, users are forced to write custom SDL config overrides just to get a standard radio to work out of the box. I believe it's safe to handle this globally because we are simply allowing QGC's calibration state machine to accept the valid 0-to-MAX span that the hardware is physically sending.

> I have to look a little more but I think that will allow things like spring loaded triggers to be mapped to attitude controls which is a bad thing.

You are completely right about the danger there. If someone actually maps a spring-loaded Xbox trigger to Roll, letting go of it drops the value to 0 and outputs full-left roll. There's no software fix for a piece of hardware that lacks a physical center-return spring.
However, I think it's much more useful to have standard RC radios work correctly out of the box instead of breaking them globally just to prevent a user who doesn't know what they're doing from intentionally mapping a literal gamepad trigger to an attitude control.
To make the misidentified RC sticks work properly, the PR ensures that if a 0-to-MAX channel gets bound to Roll, Pitch, or Yaw, the validation step forces the channelTrim to the exact mathematical midpoint. Because real RC sticks physically spring back to the middle of their travel, this midpoint calculation allows them to function perfectly.

In two radios I used with QGroundControl, SDL mapped throttle on my Radiomaster Pocket and Yaw on my Taranis X9D to these trigger channels.

PR #15128, opened 2026-09-15
fix(MainWindow): stop critical vehicle message popup stealing keyboard focus

Claudio-Chies:
## Summary

The critical vehicle message toast no longer takes keyboard focus, so an incoming vehicle error can't interrupt what the user is typing in.
## Problem

`criticalVehicleMessagePopup` opened with `focus: true`, so every critical STATUSTEXT yanked active focus away from whatever had it — most visibly the MAVLink Console, whose `TextArea` claims focus once in `Component.onCompleted` and never re-acquires it. Keystrokes were then dropped entirely. The toast has no auto-close timer, so it held focus until the user clicked it.
The popup only had focus so `Popup.CloseOnEscape` could see the key — `QQuickPopup::keyPressEvent` early-returns unless the popup has active focus.

## Solution

Drop `focus` and close the toast from a `Shortcut` instead, enabled only while it is shown.
The `Shortcut` is declared inside the `Popup` rather than at window level. `QQuickShortcutContext::matcher` gates every `Qt.WindowShortcut` on `!isBlockedByPopup(item)`, and a `Shortcut` under `ApplicationWindow` is parented to `contentItem` — an ancestor of the popup item, never a descendant — so it is blocked and silently never fires while any `CloseOnEscape` popup is open. Anchoring it inside the popup also arbitrates correctly on its own: a dialog stacked above blocks it and handles Escape itself. The `enabled` gate is required rather than defensive — a closed popup's content item has no window, which leaves the shortcut unblocked and swallowing Escape application-wide.
Escape and the click gesture now share `acknowledge()`, which also closes a pre-existing gap where Escape bypassed `resetErrorLevelMessages()` / `dropMainStatusIndicatorTool()`. An incidental press outside the toast still just closes it, unchanged.

## Type of Change

- [ ] Documentation update
- [ ] CI/Build changes
- [ ] Other

## Testing

- [x] Tested locally
- [ ] Tested with hardware

### Platforms Tested

- [x] Linux
- [ ] Windows
- [ ] macOS
- [ ] Android
- [ ] iOS

### Flight Stacks Tested

- [x] PX4
- [x] ArduPilot (not tested)

### Before
A warning takes the keyboard focus away, interrupting you when typing commands or parameter searches
https://github.com/user-attachments/assets/565cb4f4-9c74-482e-96bf-24e70758ee44
 ### After
https://github.com/user-attachments/assets/b46c2557-fbe7-4b00-870a-c53b3723b758

---


Claudio-Chies:
@DonLakeFlyer This would be ready for review. the CI seems to be unrelated

PR #15129, opened 2026-09-15
ci: run upstream fork PRs on RunsOn

mrpollo:
Run upstream Linux, Windows x64/cross, Android, Docker, custom, and VM PR jobs on RunsOn, including PRs from forks. Enable their Magic Cache and Windows toolchain setup.

Keep GitHub-hosted fallbacks for independent forks and unsupported platforms. Publishing restrictions are unchanged.


mrpollo:
Running fork PRs on RunsOn is intentional. Cache misses and cold rebuilds are expected operational outcomes and do not, by themselves, establish a security defect.
 
This review identifies shared-cache permissions, but has not established that they lead to compromised trusted build output. I’m treating that as one infrastructure-level observation, not a reason to restore fork-origin restrictions across these workflows. This PR will retain RunsOn routing for fork PRs.

PR #15131, opened 2026-09-15
ci: add a 16-vCPU analysis runner

mrpollo:
Add an on-demand 16-vCPU/64-GiB Linux pool for single-job compiler analysis. Existing jobs are unchanged.

This definition must reach master before the analysis workflow can select the named pool.


mrpollo:
Superseded by the direct-runner trial; no named pool is needed for this experiment.

mrpollo:
Superseded by the inline on-demand 16-vCPU runner configuration in #15132. Consolidating the analysis feature there; a separate named-pool prerequisite is no longer needed.

PR #15132, opened 2026-09-15
ci: run incremental analysis with inline PR reviews

mrpollo:
Replace four clang-tidy shards with one on-demand 16-vCPU runner using all available workers and RunsOn S3 caching. Keep Clazy separate.

PR analysis selects affected translation units and reports findings on changed lines. Both tools export inline COMMENT reviews, including for fork PRs, without repeating analysis. Compiler errors and timeouts fail the job; manual full scans remain available.

The trusted review poster activates after its workflow and scripts reach master. Container changes remain separate in #15133.


.github/workflows/analysis.yml
```diff
@@ -57,34 +57,48 @@ permissions:
...
-            shard: 4
 
     steps:
       - name: Harden Runner
         uses: step-security/harden-runner@v2
         with:
           egress-policy: audit
 
+      - name: Enable RunsOn magic cache
+        if: github.repository_owner == 'mavlink'
```
mrpollo, line 86: Useless the package already works transparently and doesn't pollute non runs on runners

PR #15144, opened 2026-09-16
fix(FTP): fit radio read chunks to a SiK radio with error correction

julianoes:
## Description
Follow-up to 3187e9dd5 (@DonLakeFlyer). A SiK radio fragments any packet longer than its `max_data_packet_length` (see `ATI6`): 250 bytes with error correction (ECC) off, but only 118 with ECC on. The GCS cannot tell which is configured. The current 110 byte chunk is 137 bytes on the wire, so with ECC on every FTP packet is split, which costs throughput and loses the whole packet if any fragment drops.

This PR sets the radio chunk to 91 bytes, which is exactly 118 on the wire, and to 78 on a signed link to leave room for the 13 byte signature. A link that is switching signing on or off counts as signed.

This goes together with the PX4 side, https://github.com/PX4/PX4-Autopilot/pull/28714, which clamps its burst and read replies the same way. On a bench 57600 SiK pair with ECC on, bursts of up to three 118 byte packets arrived intact, while larger packets were fragmented.

## Type of Change

## Testing
- [x] Added/updated unit tests (new `signed_radio_link` row in `_testReadChunkSizeFollowsLinkType`; FTPManagerTest, FTPControllerTest and OnboardLogFtpDownloadTest pass)
- [ ] Tested with hardware: not yet with this QGC change. The 118 byte limit was measured with PX4 and MAVSDK over the SiK pair.

### Platforms Tested
- [x] Linux

Note: the commit skipped the local pre-commit hooks, which flag existing issues across `FTPManager.cc` (include order, a `Q_ASSERT`). This change does not touch those lines.


julianoes:
@DonLakeFlyer I think this one is worth having, for anyone that is using ECC=1 and has not updated the radio firmware.

See updates coming here: https://github.com/ArduPilot/SiK/pulls

